PR #8189, opened 2022-06-03
Merge Asset IO functionality with Library IO managers

clairelin135:


sryza (commented):
Good good good. Are you able to update the docs for the IO managers to clarify their behavior in the asset case?

clairelin135:
@sryza Yes, I can update the docs. I'd prefer to do it in a separate branch so that I can pass to Erin for feedback too.

sryza (approved):
Sounds good

PR #8192, opened 2022-06-04
[docs] add multi_assets docs page [CON-33]

OwenKephart:
### Summary & Motivation

Document multi assets! This is a fairly niche feature, and should live at the same place in the hierarchy as graph-backed assets (in my opinion). I just stashed it next to the SDA page as a temporary spot to plop it. It is assumed that anyone interacting with this page should have an understanding of what SDAs are.

still need to add tests to the snippets

### DIFF
--- a/docs/content/concepts/assets/multi-assets.mdx
+++ b/docs/content/concepts/assets/multi-assets.mdx
@@ -15,11 +15,13 @@ A multi-asset represents a set of assets that are all updated by the same operat
 
 ## Overview
 
-When working with software-defined assets, it's sometimes inconvenient or impossible to map each persisted asset to a unique operation. A multi-asset is a way to define a single operation that will produce the contents of multiple data assets at once. Multi-assets may be useful in the following scenarios:
+When working with [software-defined assets](/concepts/assets/software-defined-assets), it's sometimes inconvenient or impossible to map each persisted asset to a unique operation. A multi-asset is a way to define a single operation that will produce the contents of multiple data assets at once.
+
+Multi-assets may be useful in the following scenarios:
 
 - A single call to an API results in multiple tables being updated (e.g. Airbyte, Fivetran, dbt)
-- A DataFrame is transformed into multiple different assets, and serializing it between separate ops would be prohibitively inefficient
-- The same in-memory object is used to compute multiple different assets
+- A DataFrame is transformed into multiple assets, and serializing it between separate ops would be prohibitively inefficient
+- The same in-memory object is used to compute multiple assets
 
 ## Defining multi-assets
 
@@ -27,7 +29,7 @@ The function responsible for computing the contents of any software-defined asse
 
 ### A basic multi-asset
 
-The easiest way to create a multi-asset is with the <PyObject object="multi_asset" decorator /> decorator. This decorator functions similarly to the <PyObject object="asset" decorator /> decorator, but requires an `outs` parameter specifying each output asset of this function.
+The easiest way to create a multi-asset is with the <PyObject object="multi_asset" decorator /> decorator. This decorator functions similarly to the <PyObject object="asset" decorator /> decorator, but requires an `outs` parameter specifying each output asset of the function.
 
 ```python file=/concepts/assets/multi_assets.py startafter=start_basic_multi_asset endbefore=end_basic_multi_asset
 from dagster import multi_asset, Out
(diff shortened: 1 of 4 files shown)


docs/content/concepts/assets/multi-assets.mdx
```diff
@@ -0,0 +1,114 @@
...
+    outs={
+        "my_string_asset": Out(),
+        "my_int_asset": Out(),
+    }
+)
+def my_function():
+    return "abc", 123
+```
+
+By default, the names of the outputs will be used to form the asset keys of this multi-asset. The decorated function will be used to create the op for these assets, and must emit an output for each of them. In this case, we can emit multiple outputs by returning a tuple of values, one for each asset.
```
erinkcochran87, line 46: > By default, the names of the outputs will be used to form the asset keys of `this` multi-asset.

Same here as my other comment. I feel like this should be `the` instead, since it sounds like it applies generally to all multi-assets and not a specific one.

erinkcochran87, line 46: > The decorated function will be used to create the op for these assets, and must emit an output for each of them.

Let's remove the comma here since it isn't necessary.

docs/content/concepts/assets/multi-assets.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+### Customizing how assets are materialized with IO managers
+
+As with regular assets, you can customize how each asset is materialized with [IO managers](/concepts/io-management/io-managers). You may specify an `io_manager_key` on each output of the multi-asset.
```
erinkcochran87, line 50: > You may specify an `io_manager_key` on each output of the multi-asset.

Is this required? If it's not, we should reword this, as the `may` makes it sound optional.

docs/content/concepts/assets/multi-assets.mdx
```diff
@@ -0,0 +1,114 @@
...
+        "adls_asset": Out(io_manager_key="adls2_io_manager"),
+    },
+)
+def my_assets():
+    return "store_me_on_s3", "store_me_on_adls2"
+```
+
+### Subsetting multi-assets
+
+By default, it is assumed that the computation inside of a multi-asset will always produce the contents all of the associated assets. This means that attempting to execute a set of assets that produces some, but not all of the assets defined by a given multi-asset will result in an error.
```
erinkcochran87, line 68:
```suggestion
By default, it is assumed that the computation inside of a multi-asset will always produce the contents all of the associated assets. This means that attempting to execute a set of assets that produces some, but not all, of the assets defined by a given multi-asset will result in an error.
```

docs/content/concepts/assets/multi-assets.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+Sometimes, the underlying computation is sufficiently flexible to allow for computing an arbitrary subset of the assets that it is associate with. For these cases, you may set the `is_required` attribute of the outputs to `False`, and set the `can_subset` parameter of the decorator to `True`.
```
erinkcochran87, line 70:
```suggestion
Sometimes, the underlying computation is sufficiently flexible to allow for computing an arbitrary subset of the assets associated with it. In these cases, you may set the `is_required` attribute of the outputs to `False`, and set the `can_subset` parameter of the decorator to `True`.
```

erinkcochran87: > you may set the `is_required` attribute of the outputs to `False`, and set the `can_subset` parameter of the decorator to `True`.

It sounds like doing this is required for this to work correctly. If so, let's reword this (cut the `may`) so it's clear that it's not optional.

docs/content/concepts/assets/multi-assets.mdx
```diff
@@ -0,0 +1,114 @@
...
+        yield Output(value=123, output_name="a")
+    if "b" in context.selected_output_names:
+        yield Output(value=456, output_name="b")
+```
+
+Because our outputs are now optional, we can no longer rely on returning a tuple of values, as we don't know in advance which values will be computed. Instead, we explicitly `yield` each output that we're expected to create.
+
+### Dependencies inside multi-assets
+
+When a multi-asset is created, it is assumed that each output asset depends on all of the input assets. However, this may not always be the case. In these situations, you may optionally provide a mapping from each output asset to the set of <PyObject object="AssetKey" />s that it depends on. This information will be used when displaying lineage information in Dagit and when parsing selections over your asset graph.
```
erinkcochran87, line 96:
```suggestion
When a multi-asset is created, it is assumed that each output asset depends on all of the input assets. This may not always be the case, however.

In these situations, you may optionally provide a mapping from each output asset to the set of <PyObject object="AssetKey" />s that it depends on. This information is used to display lineage information in Dagit and for parsing selections over your asset graph.
```

This paragraph is on the longer side and is also informationally dense - let's break it up a bit. I also changed some of this to use active voice. Feel free to re-write!

erinkcochran87 (changes requested):

OwenKephart:
@erinkcochran87 thanks for the review, really appreciate all the comments 🙏 -- addressed everything (I believe) + added tests

PR #8197, opened 2022-06-05
asset defs directly on repository

sryza:


python_modules/dagster/dagster/core/definitions/repository_definition.py
```diff
@@ -1232,5 +1248,12 @@ def _process_and_validate_target(
         pipelines_or_jobs[target.name] = target
 
 
+def _validate_asset_resources(
+    assets_defs: Sequence["AssetsDefinition"], source_assets: Sequence[SourceAsset]
+):
+    """Validate that all asset resource requirements are satisfied and that no two assets have
+    conflicting definitions for the same resource"""
```
sryza, line 1255: @dpeng817 is there some place in the codebase where similar logic is implemented that I could steal from?

sryza:
> We'll want to update that logic so that we don't leave out assets that are not in any jobs.

With this approach, all assets are ultimately still going to end up inside jobs on the repository (using the same auto-generated jobs that we had with asset groups).

OwenKephart:
> 

good point

sryza:
Previously, this PR was blocked on https://github.com/dagster-io/dagster/pull/8223, because this PR was doing resource validation on the assets passed to the repo. That PR is taking longer than originally expected, so I updated this PR to omit resource validation, and we can add that later in a separate PR.

OwenKephart (approved):

PR #8200, opened 2022-06-06
[dagit] Update MenuLink to support a disabled state

bengotow:
### Summary & Motivation

MenuLink inherits props from both Menu and Link, but then applies them to a Link element. The link doesn't do anything with the `disabled={true}` value (it's still clickable and looks clickable). Rather than rewrite all the CSS, I just made it render a proper MenuItem when it's disabled so all the adjustments to the icons, etc. are consistent.

### DIFF
--- a/js_modules/dagit/packages/core/src/ui/MenuLink.tsx
+++ b/js_modules/dagit/packages/core/src/ui/MenuLink.tsx
@@ -20,9 +20,9 @@ interface MenuLinkProps
  * If you want to use a menu item as a link, use `MenuLink` and provide a `to` prop.
  */
 export const MenuLink: React.FC<MenuLinkProps> = (props) => {
-  const {icon, intent, text, ...rest} = props;
+  const {icon, intent, text, disabled, ...rest} = props;
 
-  if (rest.disabled) {
+  if (disabled) {
     return <MenuItem disabled icon={icon} intent={intent} text={text} />;
   }
   return (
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/core/src/ui/MenuLink.tsx
```diff
@@ -16,6 +22,9 @@ interface MenuLinkProps
 export const MenuLink: React.FC<MenuLinkProps> = (props) => {
   const {icon, intent, text, ...rest} = props;
 
+  if (rest.disabled) {
```
hellendag, line 25: May as well extract `disabled` from props instead of leaving it in `rest`.

js_modules/dagit/packages/core/src/ui/MenuLink.tsx
```diff
@@ -16,6 +22,9 @@ interface MenuLinkProps
...
+    return <MenuItem disabled icon={icon} intent={intent} text={text} />;
```
hellendag, line 26: Alternatively, should this just use `...rest`?

hellendag (approved):

js_modules/dagit/packages/core/src/ui/MenuLink.tsx
bengotow: Ahh I tried, but it gets angry about onClick and a few others having the wrong type (which is ok really, since it's disabled). Let me see what I can do!

PR #8202, opened 2022-06-06
AssetSelection

smackesey:
### Summary & Motivation

- `AssetSelection` initial implementation. Provides a Python API to define boolean filters over sets of assets, then convert them to jobs.
- Adds type annotations and makes some additions to `subset_selector.py`
- Updates the `get_user_string` method on `AssetKey` to use `/` instead of `>` as separator.

New unit tests

### DIFF
--- a/python_modules/dagster/dagster/core/asset_defs/asset_selection.py
+++ b/python_modules/dagster/dagster/core/asset_defs/asset_selection.py
@@ -1,10 +1,10 @@
-import re
 from abc import ABC, abstractmethod
 from functools import reduce
-from typing import AbstractSet, Dict, MutableSet, Pattern, Set, TypeVar
+from typing import AbstractSet, Dict, Literal, MutableSet, Optional, Pattern, TypeVar
 
 from typing_extensions import TypeAlias
 
+import dagster._check as check
 from dagster.core.asset_defs.assets import AssetsDefinition
 
 T = TypeVar("T")
@@ -20,8 +20,11 @@ class AssetSelection(ABC):
     def groups(*groups) -> "GroupsAssetSelection":
         return GroupsAssetSelection(*groups)
 
-    def downstream(self) -> "DownstreamAssetSelection":
-        return DownstreamAssetSelection(self)
+    def downstream(self, depth: Optional[int] = None) -> "DownstreamAssetSelection":
+        return DownstreamAssetSelection(self, depth=depth)
+
+    def upstream(self, depth: Optional[int] = None) -> "UpstreamAssetSelection":
+        return UpstreamAssetSelection(self, depth=depth)
 
     def __or__(self, other: "AssetSelection") -> "OrAssetSelection":
         return OrAssetSelection(self, other)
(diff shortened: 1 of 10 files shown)


smackesey:
* master
  * **PR #8202** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8202" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8202?utm_source=stack-comment).

sryza (commented):
I think it's important to include `AssetSelection.all()` as well.

smackesey:
>I think it's important to include AssetSelection.all() as well.

Implemented in the latest.

python_modules/dagster/dagster/core/asset_defs/asset_selection.py
```diff
@@ -0,0 +1,142 @@
...
+        self.children = (child,)
+        self.depth = depth
+
+
+# ########################
+# ##### RESOLUTION
+# ########################
+
+
+class Resolver:
```
sryza, line 90: Thoughts on making this a function instead of a class? That would make it crystal clear what data `resolve` operates on (because it would be in the signature).

python_modules/dagster/dagster/core/asset_defs/asset_selection.py
```diff
@@ -0,0 +1,142 @@
...
+    def __init__(self, all_assets: AssetSet):
+        self.all_assets = all_assets
+        self.asset_dep_graph = generate_asset_dep_graph(list(all_assets))
+        self.all_assets_by_name = generate_asset_name_to_definition_map(all_assets)
+
+    def resolve(self, node: AssetSelection) -> AssetSet:
+        if isinstance(node, AllAssetSelection):
```
sryza, line 97: Did you investigate whether it would make sense to push this down into an abstract method of `AssetSelection` instead of doing these `isinstance` checks?

python_modules/dagster/dagster/core/asset_defs/asset_selection.py
```diff
@@ -0,0 +1,142 @@
...
+        elif isinstance(node, UpstreamAssetSelection):
+            child = self.resolve(node.children[0])
+            return reduce(
+                operator.or_,
+                [self.gather_connected_assets(asset, "upstream", node.depth) for asset in child],
+            )
+        else:
+            check.failed(f"Unknown node type: {type(node)}")
+
+    def gather_connected_assets(
```
sryza, line 128: If this is only used by `resolve`, its name should start with an underscore.

python_modules/dagster/dagster/core/asset_defs/asset_selection.py
smackesey: It's a class because it needs to essentially cache some intermediate data structures (dependency graph and asset lookup table) that are used for the duration of the computation, which includes multiple recursive node resolutions.

python_modules/dagster/dagster/core/asset_defs/asset_selection.py
smackesey: That was the original implementation, but then I converted it to a class for the reason above-- because intermediate data structures are needed over the course of a resolution. It is cleaner to retain these as instance variables of a class then it is to recursively pass them down when for many individual node resolutions they are not needed.

python_modules/dagster/dagster/core/asset_defs/asset_selection.py
smackesey: good point, fixed

PR #8203, opened 2022-06-06
[dagit] Add top level Asset Group pages, Asset Groups in left nav

bengotow:
### Summary & Motivation

This PR adds asset groups to the sectioned left nav and dedicated pages for viewing asset groups! The new asset group pages offer a more structured replacement for the global asset graph.

**GraphQL changes:**

- The `assetNodes` resolver now takes an optional `group: GrapheneAssetGroupSelector` filter so that the UI can easily grab a group. (This is slightly preferable to going `group() { assetNodes { ... } }` because it's challenging to conditionally switch queries and easy to pass different params to the same query.

**Asset Groups Pages:**

Screen Shot 2022-06-03 at 1 46 06 PM

Screen Shot 2022-06-03 at 1 46 23 PM

**Updated Left Nav:**

- There are now sections for Jobs / Asset Groups (the section titles disappear when only showing the match job)
- The header now says "Workspace" instead of "Jobs"
- There are now icons to the left of each item within the selection highlight

Screen Shot 2022-06-03 at 1 40 27 PM
Screen Shot 2022-06-03 at 1 44 15 PM
Screen Shot 2022-06-03 at 1 43 29 PM


js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
```diff
@@ -401,52 +398,45 @@ export const AssetGraphExplorerWithData: React.FC<
...
+              placeholder="Type an asset subset…"
+              onChange={(opsQuery) => onChangeExplorerPath({...explorerPath, opsQuery}, 'replace')}
+              popoverPosition="bottom-left"
+            />
+          </QueryOverlay>
         </>
       }
       second={
-        !options.enableSidebar ? null : selectedGraphNodes.length === 1 && selectedGraphNodes[0] ? (
+        selectedGraphNodes.length === 1 && selectedGraphNodes[0] ? (
```
bengotow, line 439: Removing `options.enableSidebar` because it was never set to false anywhere. The "Asset Details Lineage" tab which this option was built for no longer uses this component

js_modules/dagit/packages/core/src/assets/AssetsCatalogRoot.tsx
```diff
@@ -1,25 +1,71 @@
...
+  useDocumentTitle(
+    currentPath && currentPath.length
+      ? `Assets: ${displayNameForAssetKey({path: currentPath})}`
+      : 'Assets',
+  );
+
+  return queryResult.loading ? (
+    <Page>
+      <AssetPageHeader assetKey={{path: currentPath}} />
+      <Loading queryResult={queryResult}>{() => null}</Loading>
```
hellendag, line 43: Just render a `Spinner` or something?

js_modules/dagit/packages/core/src/assets/AssetsCatalogTable.tsx
```diff
@@ -10,48 +10,95 @@ import {
...
+    if (groupSelector) {
+      return {
+        query: groupQuery,
+        error: undefined,
+        assets: groupQuery.data?.assetNodes.map<AssetTableFragment>((definition) => ({
+          __typename: 'Asset',
+          id: definition.id,
+          key: definition.assetKey,
+          definition,
+        })),
```
hellendag, line 70: Seems a little strange to reconstruct fragment objects from a GraphQL response. Why does this need to be remapped?

hellendag (approved):

js_modules/dagit/packages/core/src/assets/AssetsCatalogRoot.tsx
bengotow: Oh good find, this is a bit odd, will fix!

js_modules/dagit/packages/core/src/assets/AssetsCatalogTable.tsx
bengotow: Ahh this is a bit weird. I pulled this mapping out into a helper method and put a comment on it for now - I think that the core problem is that one query gets us `assets` (with an optional definition) and the other gets us `assetNodes` (that are a definition). 

I think the best path might be to make a separate interface for the AssetCatalogTable and convert both into that type, or change the backend so that we can fetch `assets(groupName: )` instead of `assetNodes(groupName:)`. Either way will revisit in a bit because I think it could be a bit messy!

js_modules/dagit/packages/core/src/assets/AssetGroupRoot.tsx
```diff
@@ -0,0 +1,76 @@
...
+            Asset Group in <RepositoryLink repoAddress={repoAddress} />
+          </Tag>
+        }
+        tabs={
+          <Box flex={{direction: 'row', justifyContent: 'space-between', alignItems: 'flex-end'}}>
+            <Tabs selectedTabId={tab}>
+              <TabLink
+                id="lineage"
+                title="Lineage"
+                to={workspacePathFromAddress(repoAddress, `/asset-groups/${groupName}?tab=lineage`)}
```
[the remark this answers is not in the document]
bengotow: Oh interesting you're definitely right about this. I was originally trying to push it into a query param so I could use path for asset selection etc in the lineage tab but we ended up not doing that. Will change this to match the other tabbed UIs!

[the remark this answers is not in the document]
bengotow: Edit: it looks like I need to address this later in the PR stack because it changed a bit, will move this feedback there!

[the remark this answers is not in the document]
bengotow: Edit: it looks like I need to address this later in the PR stack because it changed a bit, will move this feedback there!

PR #8204, opened 2022-06-06
[dagit] Add asset group filter to the Asset Catalog

bengotow:

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/AssetsCatalogTable.tsx
+++ b/js_modules/dagit/packages/core/src/assets/AssetsCatalogTable.tsx
@@ -1,8 +1,18 @@
 import {gql, QueryResult, useQuery} from '@apollo/client';
+import {
+  Box,
+  CursorPaginationControls,
+  CursorPaginationProps,
+  TextInput,
+  Suggest,
+  MenuItem,
+  Icon,
+} from '@dagster-io/ui';
+import isEqual from 'lodash/isEqual';
+import uniqBy from 'lodash/uniqBy';
 import * as React from 'react';
 import styled from 'styled-components/macro';
 
-import {Box, CursorPaginationControls, CursorPaginationProps, TextInput} from '../../../ui/src';
 import {PythonErrorInfo, PYTHON_ERROR_FRAGMENT} from '../app/PythonErrorInfo';
 import {
   FIFTEEN_SECONDS,
@@ -15,8 +25,10 @@ import {toGraphId, tokenForAssetKey} from '../asset-graph/Utils';
 import {useLiveDataForAssetKeys} from '../asset-graph/useLiveDataForAssetKeys';
 import {useQueryPersistedState} from '../hooks/useQueryPersistedState';
 import {AssetGroupSelector} from '../types/globalTypes';
+import {ClearButton} from '../ui/ClearButton';
 import {LoadingSpinner} from '../ui/Loading';
 import {StickyTableContainer} from '../ui/StickyTableContainer';
+import {buildRepoPath} from '../workspace/buildRepoAddress';
 
 import {AssetTable, ASSET_TABLE_DEFINITION_FRAGMENT, ASSET_TABLE_FRAGMENT} from './AssetTable';
 import {AssetViewModeSwitch} from './AssetViewModeSwitch';
@@ -86,11 +98,16 @@ export const AssetsCatalogTable: React.FC<AssetCatalogTableProps> = ({
   prefixPath = [],
   groupSelector,
 }) => {
+  const [view, setView] = useAssetView();
   const [cursor, setCursor] = useQueryPersistedState<string | undefined>({queryKey: 'cursor'});
   const [search, setSearch] = useQueryPersistedState<string | undefined>({queryKey: 'q'});
-  const [view, _setView] = useAssetView();
+  const [searchGroup, setSearchGroup] = useQueryPersistedState<AssetGroupSelector | null>({
+    queryKey: 'g',
+    decode: (qs) => (qs.group ? JSON.parse(qs.group) : null),
+    encode: (group) => ({group: group ? JSON.stringify(group) : undefined}),
+  });
 
-  const searchSeparatorAgnostic = (search || '')
+  const searchPath = (search || '')
     .replace(/(( ?> ?)|\.|\/)/g, '/')
     .toLowerCase()
     .trim();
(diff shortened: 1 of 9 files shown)


js_modules/dagit/packages/core/src/assets/AssetsCatalogTable.tsx
```diff
@@ -89,9 +101,14 @@ export const AssetsCatalogTable: React.FC<AssetCatalogTableProps> = ({
   prefixPath = [],
   groupSelector,
 }) => {
+  const [view, _setView] = useAssetView();
```
hellendag, line 104: If `_setView` won't be used, don't assign it?

js_modules/dagit/packages/ui/src/components/Suggest.tsx
```diff
@@ -55,7 +55,7 @@ export const Suggest = <T,>(props: React.PropsWithChildren<SuggestProps<T>>) =>
     ...props.popoverProps,
     minimal: true,
     modifiers: deepmerge(
-      {offset: {enabled: true, offset: '0, 8px'}},
```
bengotow, line 58: I baked this into our core component because we only have two Suggests and both of them were passing this override so the dropdown aligns with the bottom of the input better.

js_modules/dagit/packages/core/src/assets/AssetsCatalogTable.tsx
```diff
@@ -101,12 +118,14 @@ export const AssetsCatalogTable: React.FC<AssetCatalogTableProps> = ({
...
-          tokenForAssetKey(a.key).toLowerCase().includes(searchSeparatorAgnostic),
-      ),
-    [assets, searchSeparatorAgnostic],
+      (assets || [])
+        .filter((a) => !searchGroup || isEqual(buildAssetGroupSelector(a), searchGroup))
+        .filter(
+          (a) =>
+            !searchSeparatorAgnostic ||
+            tokenForAssetKey(a.key).toLowerCase().includes(searchSeparatorAgnostic),
+        ),
```
hellendag, line 127: Combine into a single `filter`?

js_modules/dagit/packages/core/src/assets/AssetsCatalogTable.tsx
```diff
@@ -203,6 +226,68 @@ export const AssetsCatalogTable: React.FC<AssetCatalogTableProps> = ({
...
+        (assets || []).map(buildAssetGroupSelector).filter((a) => !!a) as AssetGroupSelector[],
+        (a) => JSON.stringify(a),
+      ).sort((a, b) => a.groupName.localeCompare(b.groupName)),
+    [assets],
+  );
+
+  const repoContextNeeded = React.useMemo(() => {
+    const result: {[groupName: string]: boolean} = {};
+    assetGroups.forEach(
+      (group) => (result[group.groupName] = group.groupName in result ? true : false),
```
hellendag, line 246: `result.hasOwnProperty(group.groupName)` just in case there's a `groupName` that matches a property in the prototype chain?

js_modules/dagit/packages/core/src/assets/AssetsCatalogTable.tsx
```diff
@@ -203,6 +226,68 @@ export const AssetsCatalogTable: React.FC<AssetCatalogTableProps> = ({
...
+              {repoContextNeeded[assetGroup.groupName] ? (
+                <span style={{opacity: 0.5, paddingLeft: 4}}>
+                  {buildRepoPath(assetGroup.repositoryName, assetGroup.repositoryLocationName)}
+                </span>
+              ) : undefined}
+            </>
+          }
+        />
+      )}
+      noResults={<MenuItem disabled={true} text="No asset groups." />}
```
hellendag, line 286: imo `No asset groups` without a period

js_modules/dagit/packages/core/src/assets/AssetsCatalogTable.tsx
```diff
@@ -244,6 +329,16 @@ const ASSET_CATALOG_GROUP_TABLE_QUERY = gql`
...
 `;
 
+function buildAssetGroupSelector(a: Asset) {
+  return a.definition && a.definition.groupName
+    ? {
+        groupName: a.definition.groupName,
+        repositoryName: a.definition?.repository.name,
+        repositoryLocationName: a.definition?.repository.location.name,
+      }
+    : false;
```
hellendag, line 339: `null` instead?

js_modules/dagit/packages/core/src/assets/AssetsCatalogTable.tsx
```diff
@@ -244,6 +329,16 @@ const ASSET_CATALOG_GROUP_TABLE_QUERY = gql`
   ${ASSET_TABLE_DEFINITION_FRAGMENT}
 `;
 
+function buildAssetGroupSelector(a: Asset) {
+  return a.definition && a.definition.groupName
+    ? {
+        groupName: a.definition.groupName,
+        repositoryName: a.definition?.repository.name,
+        repositoryLocationName: a.definition?.repository.location.name,
```
hellendag, line 337: Are the `?` operators necessary here, since you've already refined it?

js_modules/dagit/packages/ui/src/components/Suggest.tsx
```diff
@@ -55,7 +55,7 @@ export const Suggest = <T,>(props: React.PropsWithChildren<SuggestProps<T>>) =>
...
+      {offset: {enabled: true, offset: '-5px 8px'}},
```
hellendag, line 58: Is this always correct for all `Suggest`?

js_modules/dagit/packages/core/src/assets/AssetsCatalogTable.tsx
```diff
@@ -203,6 +226,68 @@ export const AssetsCatalogTable: React.FC<AssetCatalogTableProps> = ({
...
+    <Suggest<AssetGroupSelector>
+      selectedItem={value}
+      items={assetGroups}
+      inputProps={{
+        style: {width: 220},
+        placeholder: 'Filter asset groups…',
+        rightElement: value ? (
+          <div style={{marginTop: 6.5, marginRight: 6}} onClick={() => onChange(null)}>
+            <Icon name="cancel" color={Colors.Gray400} />
+          </div>
```
hellendag, line 261: Make this a styled `button` instead of a div?

hellendag (approved):

js_modules/dagit/packages/core/src/assets/AssetsCatalogTable.tsx
bengotow: Mm good call, can't be too safe.. I have a bad tendency of using `in` in all these scenarios...

js_modules/dagit/packages/ui/src/components/Suggest.tsx
bengotow: Hmm, actually I compared this to a few of our other dropdowns like the "runs filter", and it looks like the -5px is aligning the titles of the dropdown options with the text in the box but is not consistent with what we do in the run filter... I'll go ahead and revert this change for now and we can assess whether everything should be bumped out 5px so the text aligns.

js_modules/dagit/packages/core/src/assets/AssetsCatalogTable.tsx
bengotow: Ahh good call, it looks like we have one on the logs filter we can reuse for this actually 🙌

PR #8207, opened 2022-06-06
[assets] UnresolvedAssetJobDefinition

OwenKephart:
### Summary & Motivation

In order for this to work completely, we need to store AssetDefinitions (and SourceAssets) directly on the RepositoryDefinition. For now, I grab them from the AssetGroup when we're building from a list, but this won't work for dictionary-based repos.


python_modules/dagster/dagster/core/definitions/repository_definition.py
```diff
@@ -603,6 +605,12 @@ def from_dict(repository_definitions: Dict[str, Dict[str, Any]]) -> "CachingRepo
 
             if isinstance(job, GraphDefinition):
                 repository_definitions["jobs"][key] = job.coerce_to_job()
+            elif isinstance(job, UnresolvedAssetJobDefinition):
+                repository_definitions["jobs"][key] = job.resolve(
+                    # TODO: once AssetDefs are on the repository, do this
```
dpeng817, line 610: nit bc I know this is going to be resolved shortly: but github issue is more trackable than TODO

dpeng817 (approved):
LGTM!

python_modules/dagster/dagster/core/definitions/unresolved_asset_job_definition.py
```diff
@@ -0,0 +1,59 @@
...
+            ("executor_def", Optional["ExecutorDefinition"]),
+            ("description", Optional[str]),
+            ("tags", Optional[Dict[str, Any]]),
+        ],
+    )
+):
+    def __new__(
+        cls,
+        name: str,
+        selection: str,
```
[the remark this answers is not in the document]
OwenKephart: yeah the string is just a placeholder. I don't care either way re: who does the merging, and yeah I'm imagining we'd automatically convert a string-based selection to an object

python_modules/dagster/dagster/core/definitions/unresolved_asset_job_definition.py
```diff
@@ -0,0 +1,111 @@
...
+    if u:
+        selection = selection.upstream(u)
+    if d:
+        selection = selection.downstream(d)
+    return selection
+
+
+def define_asset_job(
+    name: str,
+    selection: Union[str, Sequence[str], "AssetSelection"] = None,
```
sryza, line 85: Needs to be Optional, right?

python_modules/dagster/dagster/core/definitions/unresolved_asset_job_definition.py
```diff
@@ -0,0 +1,111 @@
...
+    executor_def: "ExecutorDefinition" = None,
+    description: str = None,
+    tags: Dict[str, Any] = None,
```
sryza, line 88: Mapping?

python_modules/dagster/dagster/core/definitions/unresolved_asset_job_definition.py
```diff
@@ -0,0 +1,111 @@
...
+) -> UnresolvedAssetJobDefinition:
+    """Creates a definition of a job which will update a selection of assets. This will only be
```
sryza, line 90: "update" -> "materialize"

sryza: Would be helpful to include docstring for arguments.

python_modules/dagster/dagster/core/definitions/unresolved_asset_job_definition.py
```diff
@@ -0,0 +1,111 @@
...
+        selection = selection.upstream(u)
+    if d:
+        selection = selection.downstream(d)
+    return selection
+
+
+def define_asset_job(
+    name: str,
+    selection: Union[str, Sequence[str], "AssetSelection"] = None,
+    executor_def: "ExecutorDefinition" = None,
```
sryza, line 86: Putting an executor on an asset job feels scary to me, because it means that executing the job inside vs. outside of the job has subtle differences that are difficult to detect ahead of time. I wonder if we should punt on this or mark it as experimental until we can do a little more diligence.

sryza (approved):

PR #8212, opened 2022-06-06
Add docs for top level job inputs

dpeng817:
Added as an advanced example since I think the use case is potentially niche.

### DIFF
--- a/docs/content/concepts/ops-jobs-graphs/jobs-graphs.mdx
+++ b/docs/content/concepts/ops-jobs-graphs/jobs-graphs.mdx
@@ -398,6 +398,40 @@ def alias():
     add_one.alias("second_addition")(add_one(return_one()))
 ```
 
+### Providing Top Level Inputs to a Job
+
+You can wire inputs from the top-level of a job to the constituent ops. Consider the following op and graph:
+
+```python file=/concepts/ops_jobs_graphs/jobs.py startafter=start_top_level_input_graph endbefore=end_top_level_input_graph
+from dagster import graph, op
+
+
+@op
+def op_with_input(x):
+    return do_something(x)
+
+
+@graph
+def wires_input(x):
+    op_with_input(x)
+```
+
+I can turn the graph into a job by calling <PyObject object="GraphDefinition" method="to_job" />, and can provide a value to the input `x` using the `input_values` argument.
+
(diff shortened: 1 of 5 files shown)


sryza (commented):
Great to have this documented. This page doesn't introduce `execute_in_process`, so it might be confusing to introduce it here? Maybe makes more sense on the Job Execution page? Or on the testing page?

docs/content/concepts/ops-jobs-graphs/jobs-graphs.mdx
```diff
@@ -398,6 +398,40 @@ def alias():
...
+def op_with_input(x):
+    return do_something(x)
+
+
+@graph
+def wires_input(x):
+    op_with_input(x)
+```
+
+I can turn the graph into a job by calling <PyObject object="GraphDefinition" method="to_job" />, and can provide a value to the input `x` using the `input_values` argument.
```
erinkcochran87, line 419:
```suggestion
Turn the graph into a job by calling <PyObject object="GraphDefinition" method="to_job" /> and providing a value to the input `x` using the `input_values` argument:
```

Second person is okay in the docs, but let's refrain from first. I changed this to use an understood subject (`you`) and make it second person.

docs/content/concepts/ops-jobs-graphs/jobs-graphs.mdx
```diff
@@ -398,6 +398,40 @@ def alias():
...
+
+```python file=/concepts/ops_jobs_graphs/jobs.py startafter=start_top_level_input_job endbefore=end_top_level_input_job
+the_job = wires_input.to_job(input_values={"x": 5})
+```
+
+I can also provide input values using <PyObject object="JobDefinition" method="execute_in_process" /> or <PyObject object="GraphDefinition" method="execute_in_process" />.
```
erinkcochran87, line 425:
```suggestion
You can also provide input values using <PyObject object="JobDefinition" method="execute_in_process" /> or <PyObject object="GraphDefinition" method="execute_in_process" />:
```

Same thing here, re: first vs second person.

dpeng817:
> Great to have this documented. This page doesn't introduce `execute_in_process`, so it might be confusing to introduce it here? Maybe makes more sense on the Job Execution page? Or on the testing page?

Will link the docs to execute_in_process, but I think this should probably

erinkcochran87 (approved):
:shipit: Looks good!

PR #8213, opened 2022-06-06
Add kvs table to oss

johannkm:
KVS storage for bringing automatic re-execution daemon to OSS. Will be a bit odd- we need to stick this in run storage in oss, while have it in cloud storage in cloud. Couple ways to handle that, I think we can just make the kvs methods on cloud run storage error


johannkm (commented):
Shall I add pg and mysql tests?

prha (approved):
Yeah, I think it's good practice to add the backcompat tests.

PR #8217, opened 2022-06-06
Override resource defs when invoking assets

dpeng817:


python_modules/dagster/dagster/core/asset_defs/assets.py
```diff
@@ -72,7 +75,32 @@ def __init__(
         self._can_subset = can_subset
 
     def __call__(self, *args, **kwargs):
-        return self._node_def(*args, **kwargs)
+        from dagster.core.definitions.decorators.solid_decorator import DecoratedSolidFunction
+
+        if isinstance(self.node_def, GraphDefinition):
+            return self._node_def(*args, **kwargs)
+        solid_def = self.node_def.ensure_solid_def()
```
OwenKephart, line 82: can this just be op_def = self.op_def?

OwenKephart (approved):
nicenicenice

python_modules/dagster/dagster/core/asset_defs/assets.py
dpeng817: yea that works too

PR #8220, opened 2022-06-06
Fix upgrade story for new scheme field in postgres

gibsondan:
Summary:
By changing the default to always include this, we made it so that old user code running against a new helm upgrade errors out. Instead, only set this in the instance config if they manually set it in the Helm chart.

Test Plan: BK


rexledesma (approved):
The implementation from @peay originally had this, but I removed it 😢 sorry!

PR #8222, opened 2022-06-06
[assets] cycle resolution

OwenKephart:
### Summary & Motivation

In the face of an absurd number of merge conflicts, I took the coward's way out and created a new PR.

I wanted to be pretty conservative here, just because this is a pretty spicy change this close to the release. In particular, I only attempt to do this weird cycle resolution stuff in the case that we experience an error without doing it. Basically, you only hit this code path if you mean to hit it.

In the future, this can be factored into something a little more pleasant, but for now this should do.

unit tests


python_modules/dagster/dagster/core/asset_defs/assets_job.py
```diff
@@ -84,10 +85,23 @@ def asset2(asset1):
...
     source_assets_by_key = build_source_assets_by_key(source_assets)
 
-    deps, assets_defs_by_node_handle = build_deps(assets, source_assets_by_key.keys())
     partitioned_config = build_job_partitions_from_assets(assets, source_assets or [])
     resource_defs = check.opt_mapping_param(resource_defs, "resource_defs")
 
+    deps, assets_defs_by_node_handle = build_deps(assets, source_assets_by_key.keys())
+
+    try:
+        node_deps = {}
```
sryza, line 94: Could be nice to put this inside a utility function called `has_cycles` or something.

sryza (approved):

PR #8223, opened 2022-06-06
Add resource defs to source asset, handle transitive dependencies

dpeng817:
This adds resource_defs to source assets, which allows for transitive resource dependencies to be satisfied. Also added general checking of transitive resource deps for asset jobs.

We'd like to error out when users don't satisfy resource dependencies on the asset itself at job construction time, but we can't do that until we remove the ability to specify resources externally from the asset (outside of with_resources, which actually constructs a new asset).

### DIFF
--- a/python_modules/dagster/dagster/core/asset_defs/assets_job.py
+++ b/python_modules/dagster/dagster/core/asset_defs/assets_job.py
@@ -29,7 +29,7 @@ from dagster.core.definitions.dependency import (
 )
 from dagster.core.definitions.events import AssetKey
 from dagster.core.definitions.executor_definition import ExecutorDefinition
-from dagster.core.definitions.graph_definition import GraphDefinition
+from dagster.core.definitions.graph_definition import GraphDefinition, default_job_io_manager
 from dagster.core.definitions.job_definition import JobDefinition
 from dagster.core.definitions.output import OutputDefinition
 from dagster.core.definitions.partition import PartitionedConfig, PartitionsDefinition
@@ -38,6 +38,7 @@ from dagster.core.definitions.resource_definition import ResourceDefinition
 from dagster.core.errors import DagsterInvalidDefinitionError
 from dagster.core.execution.with_resources import with_resources
 from dagster.core.selector.subset_selector import AssetSelectionData
+from dagster.utils import merge_dicts
 from dagster.utils.backcompat import experimental
 
 from .asset_partitions import get_upstream_partitions_for_partition_range
@@ -99,6 +100,7 @@ def build_assets_job(
     check.opt_str_param(description, "description")
     check.opt_inst_param(_asset_selection_data, "_asset_selection_data", AssetSelectionData)
     resource_defs = check.opt_mapping_param(resource_defs, "resource_defs")
+    resource_defs = merge_dicts({"io_manager": default_job_io_manager}, resource_defs)
 
     assets = with_resources(assets, resource_defs)
     source_assets = with_resources(source_assets, resource_defs)
--- a/python_modules/dagster/dagster/core/asset_defs/assets_job.py
+++ b/python_modules/dagster/dagster/core/asset_defs/assets_job.py
@@ -35,6 +35,9 @@ from dagster.core.definitions.output import OutputDefinition
 from dagster.core.definitions.partition import PartitionedConfig, PartitionsDefinition
 from dagster.core.definitions.partition_key_range import PartitionKeyRange
 from dagster.core.definitions.resource_definition import ResourceDefinition
+from dagster.core.definitions.resource_requirement import (
+    get_and_validate_transitive_resource_dependencies,
+)
 from dagster.core.errors import DagsterInvalidDefinitionError
 from dagster.core.selector.subset_selector import AssetSelectionData
 from dagster.utils.backcompat import experimental
@@ -131,14 +134,20 @@ def build_assets_job(
     )
 
     all_resource_defs = dict(resource_defs)
-    for asset_def in assets:
-        for resource_key, resource_def in asset_def.resource_defs.items():
+    for assets_def in assets:
+        # Until we remove the ability to specify resource defs externally here, cannot check all asset defs for satisfying their resource requirements.
+        if assets_def.resource_defs:
+            get_and_validate_transitive_resource_dependencies(
+                assets_def.resource_defs, assets_def.required_resource_keys
+            )
+        for resource_key, resource_def in assets_def.resource_defs.items():
             if (
                 resource_key in all_resource_defs
                 and all_resource_defs[resource_key] != resource_def
             ):
+                asset_keys_str = ", ".join([key for key in assets_def.asset_keys])
                 raise DagsterInvalidDefinitionError(
-                    f"When attempting to build job, asset {asset_def.asset_key} had a conflicting version of the same resource key {resource_key}. Please resolve this conflict by giving different keys to each resource definition."
+                    f"When attempting to build job, asset keys {asset_keys_str} had a conflicting version of the same resource key {resource_key}. Please resolve this conflict by giving different keys to each resource definition."
                 )
             all_resource_defs[resource_key] = resource_def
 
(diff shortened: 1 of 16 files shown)


python_modules/dagster/dagster/core/definitions/resource_requirement.py
```diff
@@ -197,3 +199,30 @@ def ensure_requirements_satisfied(
             raise DagsterInvalidDefinitionError(
                 f"{requirement.describe_requirement()} was not provided{mode_descriptor}. Please provide a {str(requirement.expected_type)} to key '{requirement.key}', or change the required key to one of the following keys which points to an {str(requirement.expected_type)}: {requirement.keys_of_expected_type(resource_defs)}"
             )
+
+
+def get_and_validate_transitive_resource_dependencies(
+    resource_defs: Dict[str, "ResourceDefinition"],
```
sryza, line 205: Nitpick: prefer Mapping to Dict

python_modules/dagster/dagster/core/asset_defs/assets_job.py
```diff
@@ -111,14 +114,20 @@ def asset2(asset1):
...
+            )
+        for resource_key, resource_def in assets_def.resource_defs.items():
             if (
                 resource_key in all_resource_defs
                 and all_resource_defs[resource_key] != resource_def
             ):
+                asset_keys_str = ", ".join([key for key in assets_def.asset_keys])
                 raise DagsterInvalidDefinitionError(
-                    f"When attempting to build job, asset {asset_def.asset_key} had a conflicting version of the same resource key {resource_key}. Please resolve this conflict by giving different keys to each resource definition."
+                    f"When attempting to build job, asset keys {asset_keys_str} had a conflicting version of the same resource key {resource_key}. Please resolve this conflict by giving different keys to each resource definition."
```
sryza, line 130: I think it's a little bit inaccurate to say that the "asset keys" have conflicting versions of the same resource key. It's the assets that have resources - not the keys. Separately, if I read that "asset1, asset2" had a conflicting version of the same resource key, I would think that I needed to fix the problem by using the same resource key for asset1 and for asset2, which I don't think is the case, right?

python_modules/dagster/dagster/core/asset_defs/assets_job.py
```diff
@@ -111,14 +114,20 @@ def asset2(asset1):
     )
 
     all_resource_defs = dict(resource_defs)
-    for asset_def in assets:
-        for resource_key, resource_def in asset_def.resource_defs.items():
+    for assets_def in assets:
+        # Until we remove the ability to specify resource defs externally here, cannot check all asset defs for satisfying their resource requirements.
```
sryza, line 118: Thoughts on using "`with_resources`" here on the assets so that the code paths afterwards don't need to account for the resource_defs that are passed in?

python_modules/dagster/dagster/core/definitions/resource_requirement.py
```diff
@@ -197,3 +199,30 @@ def ensure_requirements_satisfied(
             raise DagsterInvalidDefinitionError(
                 f"{requirement.describe_requirement()} was not provided{mode_descriptor}. Please provide a {str(requirement.expected_type)} to key '{requirement.key}', or change the required key to one of the following keys which points to an {str(requirement.expected_type)}: {requirement.keys_of_expected_type(resource_defs)}"
             )
+
+
+def get_and_validate_transitive_resource_dependencies(
```
sryza, line 204: I find this function a little bit confusing to grok, because it's kind of doing two things at the same time. Would it make sense to split them out into two steps?

python_modules/dagster/dagster/core/asset_defs/assets_job.py
```diff
@@ -127,7 +136,26 @@ def asset2(asset1):
...
+                source_asset.resource_defs, source_asset.io_manager_def.required_resource_keys
+            )
             all_resource_defs[source_asset.get_io_manager_key()] = source_asset.io_manager_def
+        for resource_key, resource_def in source_asset.resource_defs.items():
+            if (
+                resource_key in all_resource_defs
+                and all_resource_defs[resource_key] != resource_def
+            ):
+                raise DagsterInvalidDefinitionError(
+                    f"When attempting to build job, asset {source_asset.key} had a conflicting version of the same resource key {resource_key}. Please resolve this conflict by giving different keys to each resource definition."
```
sryza, line 156: Line length

python_modules/dagster/dagster/core/asset_defs/assets_job.py
```diff
@@ -127,7 +136,26 @@ def asset2(asset1):
         if not source_asset.io_manager_def:
             required_io_manager_keys.add(source_asset.get_io_manager_key())
         else:
+            # Explicitly check top-level dependencies so we can give error
+            # message in terms of asset key, and not system-provided resource key.
+            for requirement in source_asset.io_manager_def.get_resource_requirements():
+                if not requirement.resources_contain_key(source_asset.resource_defs):
+                    raise DagsterInvalidDefinitionError(
+                        f"IO manager for source asset with key {source_asset.key} requires resource with key {requirement.key}, but none provided."
```
sryza, line 144: Line length.

python_modules/dagster/dagster/core/asset_defs/assets_job.py
```diff
@@ -127,7 +136,26 @@ def asset2(asset1):
         if not source_asset.io_manager_def:
             required_io_manager_keys.add(source_asset.get_io_manager_key())
         else:
+            # Explicitly check top-level dependencies so we can give error
```
sryza, line 139: This strikes me as a little awkward. Would we not want to give the further transitive errors in terms of the asset key as well?

python_modules/dagster/dagster/core/definitions/resource_requirement.py
dpeng817: I think at this point we can verify that it's dict. Agreed that we prefer that we use Mapping when ingesting resource defs, but internally we should transform to dict so that stuff like `merge_dicts` works properly.

python_modules/dagster/dagster/core/asset_defs/assets_job.py
dpeng817: That's correct - good call out. Will change the wording here.

python_modules/dagster/dagster/core/asset_defs/assets_job.py
dpeng817: cool idea. `with_resources` just landed so will add that.

python_modules/dagster/dagster/core/definitions/resource_requirement.py
dpeng817: this is a case where we cannot do that unless we want to create a secondary error state, as in order to get the next set of dependencies we need to validate the previous set.

python_modules/dagster/dagster/core/asset_defs/assets_job.py
dpeng817: Was thinking about that. I think the thing we don't like about the top-layer error messages is that it references a system-provided key. I see your point about it being a bit awkward though

python_modules/dagster/dagster/core/definitions/resource_requirement.py
sryza: The point of using `Mapping` instead of `Dict` is that it communicates that the function will not mutate the object that's passed to it, which is a helpful invariant for reasoning about what code does. `merge_dicts` actually accepts `Mapping` objects.

dpeng817: hmm I thought I had gotten errors from using `merge_dicts` with mapping in the past, but perhaps that's since been rectified. Will change.

python_modules/dagster/dagster/core/definitions/resource_requirement.py
sryza: Got it 👍

dpeng817: hmm upon further thinking I don't think what I say here is true ^. Working to split these out.

sryza (approved):
This looks good to me aside from the last couple comments I left

python_modules/dagster/dagster/core/asset_defs/assets_job.py
```diff
@@ -130,30 +136,18 @@ def asset2(asset1):
...
-            ):
+    all_resource_defs = {}
+    for asset in assets + source_assets:
+        for resource_key, resource_def in asset.resource_defs.items():
+            if resource_key not in all_resource_defs:
+                all_resource_defs[resource_key] = resource_def
+            if all_resource_defs[resource_key] != resource_def:
                 raise DagsterInvalidDefinitionError(
-                    f"When attempting to build job, asset {asset_def.asset_key} had a conflicting version of the same resource key {resource_key}. Please resolve this conflict by giving different keys to each resource definition."
+                    f"Conflicting versions of resource with key '{resource_key}' "
```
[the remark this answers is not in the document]
dpeng817: agreed that's ideal, agreed we should kick that.

python_modules/dagster/dagster/core/asset_defs/source_asset.py
```diff
@@ -81,26 +97,44 @@ def metadata(self) -> MetadataMapping:
         return {entry.label: entry.entry_data for entry in self.metadata_entries}  # type: ignore
 
     def get_io_manager_key(self) -> str:
```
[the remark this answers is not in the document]
dpeng817: Originally the idea was that it was calculated and changeable by calls to with_resources, but fine to make it a property too.

python_modules/dagster/dagster/core/execution/resources_init.py
```diff
@@ -89,12 +82,32 @@ def _get_deps_helper(resource_key):
...
+                    f"Resource with key '{reqd_resource_key}' required by resource with key '{resource_key}', but not provided."
+                )
+            _helper(reqd_resource_key)
         path.remove(resource_key)
+
+    for resource_key in sorted(list(resource_deps.keys())):
+        _helper(resource_key)
+
+
+def get_dependencies(resource_name, resource_deps):
```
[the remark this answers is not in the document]
dpeng817: Will follow up on that because typing the code path here is a whole different can of worms I think

PR #8225, opened 2022-06-06
`assets_from_module` methods [1/2]

clairelin135:
This PR creates utility methods of the following static initializers from `AssetGroup`:
- `assets_from_modules`
- `assets_from_current_modules`
- `assets_from_package_module`
- `assets_from_package_name`

These methods will return a combined list that contains both the assets and source assets in the modules/packages.


sryza (commented):
This looks great. Will these show up in API docs?

clairelin135:
@sryza now they will show up in API docs!

sryza (approved):

PR #8232, opened 2022-06-07
unique name for assets in adls2 tests

jamiedemaria:
### Summary & Motivation
the assets in the azure io manager tests need to have unique names so that the python version tests dont try to access the same asset at the same time


prha (approved):

python_modules/libraries/dagster-azure/dagster_azure_tests/adls2_tests/test_io_manager.py
```diff
@@ -226,13 +226,13 @@ def second_op(op_1):
         assert op_1 == 5
         return op_1 + 1
 
-    @graph(name=f"graph_asset_{_id}", out={"asset3": GraphOut()})
+    @graph(name=f"graph_asset_{_id}", out={f"asset3_{_id}": GraphOut()})
```
prha, line 229: maybe add a comment here about unique names so we don't hit this again?

PR #8235, opened 2022-06-07
Renamed the auto created job prefix from '__ASSET_GROUP' to '__ASSET_JOB'

shalabhc:
### Summary & Motivation

The old name is confusing because we now have asset groups and these jobs are not related to asset groups.

Ran adjusted tests.


sryza (approved):
Lgtm. Did you try clicking around Dagit to make sure it still works?

bengotow (approved):
This looks good to me! I think those are all the front-end occurrences 👍

shalabhc:
@sryza yes clicking around and materializing assets without jobs works, but I just noticed that the UI shows '__ASSET_JOB' in some places where it doesn't on main.

Eg new:
image

vs old:
image

So something is amiss. Looking into it.

shalabhc:
> So something is amiss. Looking into it.

Running 'make rebuild_dagit' fixed it.

PR #8236, opened 2022-06-07
`assets_from_module` methods [2/2]

clairelin135:
Adds `group_name` and `asset_key_prefix` fields on `assets_from_module` methods. Refactors existing `AssetGroup.prefixed` to be in a helper function. For now, group names will be added to all `AssetsDefinition` instances (so will not apply to source assets).

Followups:
- Since this allows us to add group names fields on multi-assets, we should add `group_name` as a field on `multi_asset`
- `group_name` can also exist on `SourceAsset`?


python_modules/dagster/dagster/core/asset_defs/assets_from_modules.py
```diff
@@ -71,7 +71,10 @@ def assets_and_source_assets_from_modules(
 
 
 def assets_from_modules(
-    modules: Iterable[ModuleType], extra_source_assets: Optional[Sequence[SourceAsset]] = None
+    modules: Iterable[ModuleType],
+    group_name: Optional[str] = None,
+    asset_key_prefix: Optional[str] = None,
+    extra_source_assets: Optional[Sequence[SourceAsset]] = None,
```
sryza, line 77: I should have caught this the first time around: now that this just returns a list, I don't think there's any need for this `extra_source_assets` argument. The user could just concatenate this list to the list afterwards.

python_modules/dagster/dagster/core/asset_defs/assets_from_modules.py
```diff
@@ -71,7 +71,10 @@ def assets_and_source_assets_from_modules(
 
 
 def assets_from_modules(
-    modules: Iterable[ModuleType], extra_source_assets: Optional[Sequence[SourceAsset]] = None
+    modules: Iterable[ModuleType],
+    group_name: Optional[str] = None,
+    asset_key_prefix: Optional[str] = None,
```
sryza, line 76: Thoughts on just calling this `key_prefix` for brevity? Also, these need documentation.

sryza (changes requested):
Looking good

python_modules/dagster/dagster/core/asset_defs/assets_from_modules.py
clairelin135: Got it--removed these.

python_modules/dagster/dagster/core/asset_defs/assets_from_modules.py
clairelin135: Changed this to `key_prefix` and added docstrings!

sryza (commented):
Had a few more comments.

Also, while reviewing this PR, it occurred to me that `assets_from_package_module` should probably be `load_assets_from_package_module` , to be consistent with `load_assets_from_dbt_project` and to follow the general guideline of functions as verbs instead of nouns. My bad for recommending the wrong thing in the first place.

sryza (approved):
Nice!

PR #8237, opened 2022-06-07
Move config editor to core so we can reuse in admin portal

salazarm:
### Summary & Motivation

This PR moves the config editor dialog into the dagit core library so that other packages depending on it can reuse this dialog

Replaced the cloud version with this version and everything still worked


bengotow (approved):
This looks good! I guess the corresponding deletion is in the other repo?

PR #8242, opened 2022-06-07
[dagit] Don't block scroll when mouse hits custom tooltip

hellendag:
### Summary & Motivation

Because the BaseTag uses `pointer-events: none` for the custom tooltip overlay, scrolling is blocked when the mouseover happens to be on an overflowing `BaseTag`, e.g. on the Runs page.

I'm not sure we actually need this rule. Removing it resolves the issue.

View Runs page with a long tag. Verify that I can scroll past it, and that interaction with the tag behaves normally.

### DIFF
--- a/js_modules/dagit/packages/ui/src/components/BaseTag.tsx
+++ b/js_modules/dagit/packages/ui/src/components/BaseTag.tsx
@@ -19,7 +19,7 @@ const BaseTagTooltipStyle: React.CSSProperties = {
   alignItems: 'center',
   padding: '4px 8px',
   userSelect: 'text',
-  pointerEvents: 'all',
+  pointerEvents: 'none',
   borderRadius: 8,
   border: 'none',
   top: -10,


hellendag:
* master
  * **PR #8242** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8242" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8242?utm_source=stack-comment).

hellendag:
@bengotow Do you remember why this rule is in place on `BaseTag`?

bengotow:
Hey hey! Ahh yeah this was originally `pointer-events: all` so that you could select text inside the tooltip + copy/paste it. I think it was mostly useful for tags that contained large JSON blobs and stuff like that. Now that we have the "Copy" affordance above the tag I think we can undo it 👍 

(This was the scenario I think)
image

bengotow (approved):

PR #8243, opened 2022-06-07
add an "env_vars" key to EcsRunLauncher / EcsContainerContext

gibsondan:
Summary:
Lets you pass in a set of key-value pairs to be injected into the container.

### DIFF
--- a/python_modules/dagster/dagster/core/utils.py
+++ b/python_modules/dagster/dagster/core/utils.py
@@ -1,9 +1,10 @@
+import os
 import random
 import string
 import uuid
 import warnings
 from collections import OrderedDict
-from typing import Union
+from typing import Tuple, Union, cast
 
 import toposort as toposort_
 
@@ -80,3 +81,14 @@ def check_dagster_package_version(library_name, library_version):
             __version__, library_name, library_version
         )
         warnings.warn(message)
+
+
+def parse_env_var(env_var_str: str) -> Tuple[str, str]:
+    if "=" in env_var_str:
+        split = env_var_str.split("=", maxsplit=1)
+        return (split[0], split[1])
+    else:
+        env_var_value = os.getenv(env_var_str)
+        if env_var_value == None:
+            raise Exception(f"Tried to load environment variable {env_var_str}, but it was not set")
+        return (env_var_str, cast(str, env_var_value))
--- a/python_modules/libraries/dagster-aws/dagster_aws/ecs/container_context.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/ecs/container_context.py
@@ -1,10 +1,11 @@
-from typing import TYPE_CHECKING, Any, List, Mapping, NamedTuple, Optional
+from typing import TYPE_CHECKING, Any, Dict, List, Mapping, NamedTuple, Optional
 
 from dagster import Array, Field, Noneable, Shape, StringSource
 from dagster import _check as check
 from dagster.config.validate import process_config
 from dagster.core.errors import DagsterInvalidConfigError
 from dagster.core.storage.pipeline_run import PipelineRun
+from dagster.core.utils import parse_env_var
 from dagster.utils import merge_dicts
 
 from ..secretsmanager import get_tagged_secrets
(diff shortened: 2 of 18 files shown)


python_modules/libraries/dagster-aws/dagster_aws/ecs/container_context.py
```diff
@@ -1,25 +1,36 @@
...
 
 from ..secretsmanager import get_tagged_secrets
 
 if TYPE_CHECKING:
     from . import EcsRunLauncher
 
+# Config shared between EcsRunLauncher and EcsContainerContext
+SHARED_ECS_SCHEMA = {
+    "env_vars": Field(
+        [StringSource],
```
jmsanders, line 19: What's the current environment? Like from the container of the task that's launching it?

Is this to maintain parity with how we set env vars in other launchers? Otherwise, we might want to more directly mirror the standard ECS configuration since we've been bitten a bit in the past (secrets) with confused users when we tried to add clever abstractions on top of the raw config:

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/taskdef-envfiles.html

python_modules/libraries/dagster-aws/dagster_aws/ecs/container_context.py
```diff
@@ -65,6 +84,10 @@ def get_secrets_dict(self, secrets_manager) -> Mapping[str, str]:
             {secret["name"]: secret["valueFrom"] for secret in self.secrets},
         )
 
+    def get_environment_dict(self) -> Dict[str, str]:
```
jmsanders, line 87: Related to my other comment but I see transformation of our env var representation (`List[str]` -> `Dict[str, str]` -> `List[Dict[str,str]]`) to be a bit of an unfortunate complexity caused by our original decision to represent secrets differently in our config than AWS expects them in theirs. And perhaps something we could avoid entirely with env vars.

python_modules/libraries/dagster-aws/dagster_aws_tests/ecs_tests/launcher_tests/test_launching.py
```diff
@@ -151,12 +151,19 @@ def test_task_definition_registration(
 def test_reuse_task_definition(instance):
     image = "image"
     secrets = []
+    environment = [
+        {
+            "name": "MY_SECRET",
```
jmsanders, line 156: Nit: let's called this `MY_ENV_VAR` (or just start with an empty environment)

python_modules/dagster/dagster/core/utils.py
```diff
@@ -80,3 +81,14 @@ def check_dagster_package_version(library_name, library_version):
             __version__, library_name, library_version
         )
         warnings.warn(message)
+
+
+def parse_env_var(env_var_str: str) -> Tuple[str, str]:
+    if "=" in env_var_str:
+        split = env_var_str.split("=")
```
jmsanders, line 88: This will cause issues if you have `=` in your environment variable (very likely for anything base64 encoded). Something like:

```suggestion
        split = env_var_str.split("=", maxsplit=1)
```

will probably be sufficient because I don't think you can have `=` in the name of an envvar, but we should add a few unit tests of this function to assert against various edge cases we can imagine (empty list, list with only keys, list with names and key/value pairs, values with an equal sign, values with multiple equal signs, etc.)

jmsanders (changes requested):

python_modules/libraries/dagster-aws/dagster_aws/ecs/container_context.py
gibsondan: Yeah, it's to have a consistent way to set env vars across all three environments

gibsondan: in this case, i think its syntactic sugar on top of the exact same functionality (a name and a value). I actually originally had "environment" mapping directly onto the field in the container definition, but decided against it for consistency across the agents and because the raw AWS format didn't offer anything new, unlike secrets

we could additionally offer a separate environment_files key if we wanted (and I would still like to offer a raw JSON config option as an escape hatch)

gibsondan:
(also added the k8s version - now docker, k8s, and ecs all have the same way to specify env vars, which I think is nice)

PR #8229, opened 2022-06-07
docs: improve search 1/

yuhan:
### Summary & Motivation
What went wrong
- Changelog and migration shouldn't be included when indexing - this PR explicitly stops indexing these in [1].
- Changelog and migration and cloud site shouldn't be included in sitemap (docs.dagster.io/sitemap.xml) which is the starting point of the indexing script too - this PR removes these entries from the xml generation script in [2].
- Apidocs weren't indexed recently because the html is now using "section" id instead of "section" class - this PR changes how the search indexing script finds api docs in [3].
- Section title wasn't properly marked as class "DocSearch-lvl1" (which is what the search indexing script identifies as the section name to fill) so missing that class resulted in search results falling back to the default value "section" (see second bullet on issue DREL-296 and screenshot below) - this PR adds that class back in [4]
  image
  * **once we land this, reindexing should fill out all the "section" in the screenshot with correct title names**

### How I Tested These Changes
- Reran the search indexing script locally which 1) stopped crawling changelog and migration (changelog and migration entries no longer show up in the docs search) and 2) started to crawl api docs (apidocs show up in the search).
- Wait to land this to prod so we can rerun the script again to pick up the section title fix.

PR #8244, opened 2022-06-07
docs: improve search 2/

yuhan:
### Summary & Motivation

**This needs to get in before the release, because every release would reindex the search.**

In https://github.com/dagster-io/dagster/pull/8229, we added the class "DocSearch-lvl1" (which is what the search indexing script identifies as the section name to fill) to the left nav. This PR moves those "DocSearch-lvlx" classes from left nav to `BreadcrumbNav`.

Why?
- The change is because the left nav is client-side rendered and when crawling client-side rendered information it requires a lot longer time to crawl (see details in [algolia docs](https://docsearch.algolia.com/docs/legacy/config-file#js_wait-optional)).
  * js_render=true took ~60 mins while the indexing with js_render=false took ~5 mins.

How?
- Make BreadcrumbNav component (with navigation data) statically generated which means the page HTML is generated at build time (see details in [nextjs docs](https://nextjs.org/docs/basic-features/pages#static-generation-recommended)) -- once the HTML is generated at build time, the indexing script won't need to render js in order to crawl it.
- Move the [custom selectors](https://docsearch.algolia.com/docs/legacy/config-file#using-custom-selectors-per-page) set on left nav to breadcrumb so the indexing script would be looking for breadcrumb in HTML and therefore improve its scraping performance.

- manually update config.json to crawl preview and update the index to a test index
- see expected result in algolia: 
  * Screen Shot 2022-06-07 at 4 40 24 PM

  * before, without js_render it was: Screen Shot 2022-06-07 at 4 23 45 PM
- script finishes within 1 min.


erinkcochran87 (approved):
Oh goodness yes! 🎉

docs/.gitignore
```diff
@@ -1,2 +1,3 @@
 sphinx/_build/
 .env
+next/.env
```
sryza, line 3: nitpick: newline at end of file

sryza (approved):

PR #8246, opened 2022-06-08
[dagit] Asset group styling adjustments, polish

bengotow:
### Summary & Motivation

This PR addresses feedback Shalabh called out in https://elementl.quip.com/JV67AfmatdZh/Dagit-UI-Assets-not-highlighted and a few more misc things I noticed during testing today:

- The "Graphs" tab in the repository view now has an empty state
- Assets appear first in all views where we list Assets, Jobs, Ops, Graphs, Sensors, Schedules
- Assets appear in the placeholder for search
- The query refresh indicator is now on the right hand side of the asset catalog list view
- The spacing of controls that float on top of the DAG have been tweaked so that they 1) align with our left margin rule for tables, and 2) align vertically with the toolbar above our tables. The end result is that tabbing from List to Lineage in the Asset Group view barely move the Materialize button etc under the tab bar!
- Apply a negative margin to the ReloadAllButton on asset group pages so the header is the same height as the job header. (Noticeable when toggling between a job and an asset group)
- Make the Latest Run column wide enough to hold the live "run failed” warnings
- Pad the bottom pagination of the asset catalog to match the bottom pagination of "runs"
- Make “No assets to display” span the asset table properly (colspan=6)

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/AssetGroupRoot.tsx
+++ b/js_modules/dagit/packages/core/src/assets/AssetGroupRoot.tsx
@@ -42,7 +42,11 @@ export const AssetGroupRoot: React.FC<{repoAddress: RepoAddress; tab: 'lineage' 
     <Page style={{display: 'flex', flexDirection: 'column', paddingBottom: 0}}>
       <PageHeader
         title={<Heading>{groupName}</Heading>}
-        right={<ReloadAllButton label="Reload definitions" />}
+        right={
+          <div style={{marginBottom: -8}}>
+            <ReloadAllButton label="Reload definitions" />
+          </div>
+        }
         tags={
           <Tag icon="asset_group">
             Asset Group in <RepositoryLink repoAddress={repoAddress} />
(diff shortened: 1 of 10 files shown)


js_modules/dagit/packages/core/src/assets/AssetGroupRoot.tsx
```diff
@@ -36,7 +36,11 @@ export const AssetGroupRoot: React.FC<{repoAddress: RepoAddress}> = ({repoAddres
     <Page style={{display: 'flex', flexDirection: 'column', paddingBottom: 0}}>
       <PageHeader
         title={<Heading>{groupName}</Heading>}
-        right={<ReloadAllButton label="Reload definitions" />}
+        right={
+          <div style={{marginBottom: -8}}>
```
hellendag, line 40: Seems slightly fragile but I guess we'll notice it if anything changes.

hellendag (approved):

PR #8251, opened 2022-06-08
[dagit] Hold the lineage graph zoom level constant as you navigate

bengotow:
### Summary & Motivation

Linear issue: https://linear.app/elementl/issue/DAGIT-69/persist-the-zoom-level-between-asset-lineage-graphs

Quick demo: https://www.loom.com/share/59617d0609ca4ffe8a2225543a0f6c91

This PR stores the zoom level of the asset lineage graph in local storage, so that you can click through nodes, toggle from upstream / downstream, etc. without the boxes resizing. (They still recenter, just not resize).

I'd like for this to not be in local storage - if you reload the page on an asset node, it should really reset the zoom and it should only be held constant as you navigate on that single page.   This is really tricky though because navigating to another node in the graph hits a new asset page (eg: `assets/foo/bar`), and we make a GraphQL query to decide whether to show the asset list or asset detail. The loading state blows away the entire AssetView and there's not a great place to  store the zoom level temporarily... 

Definitely open to other ideas for this one. I think if we can get rid of the AssetRoot query and it's loading state, that might make cleaner solutions possible (and would get rid of that loading flicker...). We should have enough context to know that the asset is an asset and not a prefix... somehow :-/

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/AssetNodeLineageGraph.tsx
+++ b/js_modules/dagit/packages/core/src/assets/AssetNodeLineageGraph.tsx
@@ -4,16 +4,18 @@ import {useHistory} from 'react-router-dom';
 import styled from 'styled-components/macro';
 
 import {AssetConnectedEdges} from '../asset-graph/AssetEdges';
+import {EXPERIMENTAL_MINI_SCALE} from '../asset-graph/AssetGraphExplorer';
 import {AssetNodeMinimal, AssetNode} from '../asset-graph/AssetNode';
 import {ForeignNode} from '../asset-graph/ForeignNode';
 import {buildComputeStatusData, GraphData, LiveData, toGraphId} from '../asset-graph/Utils';
 import {SVGViewport} from '../graph/SVGViewport';
 import {useAssetLayout} from '../graph/asyncGraphLayout';
+import {getJSONForKey} from '../hooks/useStateWithStorage';
 
 import {AssetKey} from './types';
 import {AssetNodeDefinitionFragment} from './types/AssetNodeDefinitionFragment';
 
-const EXPERIMENTAL_MINI_SCALE = 0.5;
+const LINEAGE_GRAPH_ZOOM_LEVEL = 'lineageGraphZoomLevel';
 
 export type AssetLineageScope = 'neighbors' | 'upstream' | 'downstream';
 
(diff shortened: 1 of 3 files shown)


braunjj:
Love this!

js_modules/dagit/packages/core/src/assets/AssetNodeLineageGraph.tsx
```diff
@@ -36,7 +37,8 @@ export const AssetNodeLineageGraph: React.FC<{
 
   React.useEffect(() => {
     if (viewportEl.current && layout) {
-      viewportEl.current.autocenter(false);
+      const lastZoomLevel = Number(window.localStorage.getItem(LINEAGE_GRAPH_ZOOM_LEVEL));
```
hellendag, line 40: Use `useStateWithStorage`?

js_modules/dagit/packages/core/src/graph/SVGViewport.tsx
```diff
@@ -273,17 +273,16 @@ export class SVGViewport extends React.Component<SVGViewportProps, SVGViewportSt
...
     const dw = ownerRect.width / this.props.graphWidth;
     const dh = ownerRect.height / this.props.graphHeight;
     const desiredScale = Math.min(dw, dh);
-    const boundedScale = Math.max(
-      Math.min(desiredScale, this.props.maxAutocenterZoom),
-      MIN_AUTOCENTER_ZOOM,
-    );
+    const boundedScale = scale
+      ? scale
+      : Math.max(Math.min(desiredScale, this.props.maxAutocenterZoom), MIN_AUTOCENTER_ZOOM);
```
hellendag, line 285:
```
const boundedScale = scale || ...
```

hellendag (approved):

js_modules/dagit/packages/core/src/assets/AssetNodeLineageGraph.tsx
bengotow: Hmm I tried to do this just now, but this is a bit of a weird case where we don't want the setter to trigger a re-render, because the scale is actually not a controlled prop at this level. I realized we do need to try catch this though so i'll make a few edits and re-use the helpers in useStateWithStorage

PR #8264, opened 2022-06-08
with_resources top level export

dpeng817:


sryza (approved):
Is this included in apidocs?

PR #8265, opened 2022-06-08
WIP: Add a threadpool to the sensor daemon

gibsondan:
Summary:
This adds a ThreadPoolExecutor to the daemon loop, to allow sensors to run in parallel instead of sequentially. We now make a copy of the workspace on each sensor iteration, since the workspace might get cleaned up while the sensor is in the middle of executing.

Test Plan: BK


python_modules/dagster/dagster/core/test_utils.py
```diff
@@ -562,3 +563,30 @@ async def run():
     counts = counter.counts()
     assert counts["foo"] == 20
     assert counts["bar"] == 10
+
+
+class MockThreadPoolExecutor:
```
rexledesma, line 568: nit: subclass `concurrent.futures.Executor` to be explicit?

python_modules/dagster/dagster/core/test_utils.py
```diff
@@ -562,3 +563,30 @@ async def run():
...
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        pass
+
+    def submit(self, fn, *args, **kwargs):
+        # execute functions in series without creating threads
+        # for easier unit testing
```
rexledesma, line 580: nit: would prefer this to be as a top-level docstring on the class, since this basically summarizes the class

python_modules/dagster/dagster/daemon/sensor.py
```diff
@@ -169,6 +170,56 @@ def _check_for_debug_crash(debug_crash_flags, key):
 RELOAD_WORKSPACE = 60
 
 
+class SensorFutureContext(
```
rexledesma, line 173: image

python_modules/dagster/dagster/daemon/sensor.py
```diff
@@ -169,6 +170,56 @@ def _check_for_debug_crash(debug_crash_flags, key):
...
+                "timeout",
+                default=pendulum.now("UTC").add(seconds=cls.DEFAULT_TIMEOUT_SECONDS).timestamp(),
+            ),
+        )
+
+    def timed_out(self):
+        return pendulum.now("UTC").timestamp() > self.timeout
+
+
+def check_sensor_futures(logger, future_contexts):
```
rexledesma, line 203: nit: types

python_modules/dagster/dagster/daemon/sensor.py
```diff
@@ -178,32 +229,56 @@ def execute_sensor_iteration_loop(instance, workspace, logger, until=None):
...
+    max_workers = int(os.getenv("DAGSTER_SENSOR_DAEMON_THREADPOOL_WORKERS", "3"))
 
-        if start_time - workspace_loaded_time > RELOAD_WORKSPACE:
-            workspace.cleanup()
-            workspace_loaded_time = pendulum.now("UTC").timestamp()
-            workspace_iteration = 0
+    with concurrent.futures.ThreadPoolExecutor(
+        max_workers=max_workers,
+        thread_name_prefix="sensor_daemon_worker",
+    ) as executor:
```
rexledesma, line 237: We should only instantiate this once, probably during daemon initialization? Would need to thread it down to here.

python_modules/dagster/dagster/daemon/workspace.py
```diff
@@ -68,15 +71,33 @@ def get_repository_location(self, location_name: str) -> RepositoryLocation:
...
+
+
+# A copy of the main workspace's locations that can be called from a background thread
+# in a daemon without worrying that the main thread will clean up the locations underneath us.
+# Analagous to WorkspaceRequestContext in Dagit.
+#
+# Daemons that call this should be careful to set cleanup_locations=False when calling cleanup
+# on the parent workspace that get_workspace_copy_for_iteration() was called on to create
+# this workspace.
+class DaemonIterationWorkspace(BaseDaemonWorkspace):
```
rexledesma, line 94: nit: prefer this to be a docstring so that editors can see the description on hover/on typeahead

rexledesma (changes requested):
Looks fine if it passes tests, though left a comment about initializing the thread pool only once (as well as some nits).

python_modules/dagster/dagster/daemon/sensor.py
gibsondan: execute_sensor_iteration_loop is called exactly once in the sensor daemon (SensorDaemon is not an IntervalDaemon). So this is already getting instantiated exactly once in the sensor daemon

unless you mean sharing the executor across multiple daemons (scheduler/sensor/run queue/backfill/etc.), which is an interesting idea with some tradeoffs.

The pros are that we could have one executor with the ideal number of threads for the process, and when one daemon isn't doing much the other daemons would be able to get full utilization out of the threadpool

The cons are that there would be a risk of all other daemons falling behind when one daemon kicks off a bunch of threads (i'm imagining the run queue daemon will get a threadpool soon too - so imagining a bunch of runs getting launched and the sensor daemons falling behind and timing out)

python_modules/dagster/dagster/daemon/sensor.py
```diff
@@ -169,6 +170,56 @@ def _check_for_debug_crash(debug_crash_flags, key):
...
+        future: concurrent.futures.Future,
+        timeout: Optional[float] = None,
+    ):
+        return super(cls, SensorFutureContext).__new__(
+            cls,
+            check.inst_param(future, "future", concurrent.futures.Future),
+            check.opt_float_param(
+                timeout,
+                "timeout",
+                default=pendulum.now("UTC").add(seconds=cls.DEFAULT_TIMEOUT_SECONDS).timestamp(),
```
gibsondan, line 195: realizing that this needs some more thought when used in this context... if we're not careful and things fall behind, you could get to a point where every future is timing out

python_modules/dagster/dagster/daemon/sensor.py
```diff
@@ -184,39 +185,50 @@ def execute_sensor_iteration_loop(instance, workspace, logger, until=None):
...
-        start_time = pendulum.now("UTC").timestamp()
-        if until and start_time >= until:
-            # provide a way of organically ending the loop to support test environment
-            break
-
-        if start_time - workspace_loaded_time > RELOAD_WORKSPACE:
-            workspace.cleanup()
-            workspace_loaded_time = pendulum.now("UTC").timestamp()
-            workspace_iteration = 0
+    max_workers = int(os.getenv("DAGSTER_SENSOR_DAEMON_THREADPOOL_WORKERS", "3"))
```
gibsondan, line 188: Questions in my mind about this:
- do we want to make this opt in while its new? If it's opt-in, how should it be configured? Do we need a helm chart hook?
- 3 was not chosen here in the WIP with any rigor at all - how should we go about setting a default here?

I had been considering the idea of making it opt-in by using the sychronous executor that was there in the previous rev of this PR

python_modules/dagster/dagster/daemon/workspace.py
```diff
@@ -68,15 +71,36 @@ def get_repository_location(self, location_name: str) -> RepositoryLocation:
 
         return location_entry.repository_location
 
-    def cleanup(self) -> None:
+    # change this default param
```
gibsondan, line 74: i think i had been considering taking out the default param and making callsites set this explicitly

gibsondan (commented):
thanks for picking this back up - I can't request changes on my own PR, maybe best to make a new branch/PR and link this one too it?

I think the biggest remaining question is what the rollout/configuration of this looks like and how we should choose the number of threads - should the default be 1 and people can opt-in to adding more threads? Should we try to figure it out based on the number of CPUs on the machine?

gibsondan (commented):
i had also been hoping to run some kind of benchmark to verify that this actually helps - like 3 sensors that run slowly but have a low min_interval_seconds, verifying that they actually are able to execute much closer to that min_interval_seconds now

python_modules/dagster/dagster/daemon/sensor.py
prha: Yeah, I think adding instance settings might make sense here, but that probably means hooks in the helm chart.

The synchronous option as default makes sense here... 

I'd also consider keeping it asynchronous (with a default worker count of 1) and add a `wait_for_*` call at the end of the sensor iteration?  That might mitigate the concerns around falling behind, but you'd still mitigate the IO-bound gaps.

python_modules/dagster/dagster/daemon/workspace.py
prha: Resolved in https://github.com/dagster-io/dagster/pull/8642

python_modules/dagster/dagster/daemon/sensor.py
prha: Resolved in https://github.com/dagster-io/dagster/pull/8642...

PR #8268, opened 2022-06-08
Standalone materialize method

dpeng817:
barebones standalone materialize method. Lmk if there are any test cases worth addressing here that I missed due to my lack of asset edge cases knowledge.

### DIFF
--- a/python_modules/dagster/dagster/core/asset_defs/materialize.py
+++ b/python_modules/dagster/dagster/core/asset_defs/materialize.py
@@ -1,8 +1,7 @@
-from typing import Any, List, Mapping, Optional, Sequence, Union
+from typing import Any, Optional, Sequence, Union
 
 import dagster._check as check
 
-from ..definitions import ResourceDefinition
 from ..execution.execute_in_process_result import ExecuteInProcessResult
 from ..execution.with_resources import with_resources
 from ..instance import DagsterInstance
--- a/python_modules/dagster/dagster/core/asset_defs/materialize.py
+++ b/python_modules/dagster/dagster/core/asset_defs/materialize.py
@@ -1,0 +1,49 @@
+from .source_asset import SourceAsset
+from .assets import AssetsDefinition
+from ..definitions.asset_layer import build_asset_selection_job
+from typing import List, Optional, Any, Mapping, Sequence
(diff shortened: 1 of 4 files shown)


python_modules/dagster/dagster/core/asset_defs/materialize.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def materialize(
+    assets: Sequence[AssetsDefinition],
+    source_assets: Optional[Sequence[SourceAsset]] = None,
+    run_config: Any = None,
+    instance: Optional[DagsterInstance] = None,
+) -> ExecuteInProcessResult:
+    """
+    Executes an in-process run which materializes assets that fulfill selection.
```
clairelin135, line 22: I think it would be good to mention in this docstring that this method will write asset outputs to the filesystem (via `fs_io_manager`).

clairelin135 (approved):
Looking good to me after we add the additional test cases owen mentioned

python_modules/dagster/dagster/core/asset_defs/materialize.py
sryza: I think also worth mentioning that the materializations will happen serially in a single thread.

sryza:
`load_assets_from_package_module` returns a list of AssetsDefinitions and SourceAssets, but `materialize` in the form it's in here accepts these in separate lists. Which means that `materialize(load_assets_from_package_module(...))` won't always work. I wonder if we should have `materialize` accept a list that can have both types? There might be downsides I'm not thinking of though?

dpeng817:
> `load_assets_from_package_module` returns a list of AssetsDefinitions and SourceAssets, but `materialize` in the form it's in here accepts these in separate lists. Which means that `materialize(load_assets_from_package_module(...))` won't always work. I wonder if we should have `materialize` accept a list that can have both types? There might be downsides I'm not thinking of though?

I think this is potentially problematic when you want to use assets as source assets but not materialize them? There's no way to do that if you put everything as one list right?

sryza:
> I think this is potentially problematic when you want to use assets as source assets but not materialize them? There's no way to do that if you put everything as one list right?

`SourceAsset` objects can't be materialized no matter what though, right?

PR #8271, opened 2022-06-08
[assets] Update modern-data-stack assets

OwenKephart:
### Summary & Motivation

I found a bug when running through this demo repo. Basically, we were passing in lists that contained multiple copies of each AssetsDefinition object whenever we had multi assets in the original job. This would screw up the logic in subset_asset_defs.

I added a test and changed the input types to make it clear that sets are required.

### DIFF
--- a/python_modules/dagster/dagster/core/asset_defs/asset_group.py
+++ b/python_modules/dagster/dagster/core/asset_defs/asset_group.py
@@ -109,8 +109,8 @@ class AssetGroup:
         assets = with_resources(assets, resource_defs)
         source_assets = with_resources(source_assets, resource_defs)
 
-        self._assets = assets
-        self._source_assets = source_assets
+        self._assets = set(assets)
+        self._source_assets = set(source_assets)
         self._resource_defs = resource_defs
         self._executor_def = executor_def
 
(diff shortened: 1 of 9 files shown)


python_modules/dagster/dagster/core/asset_defs/asset_group.py
```diff
@@ -109,8 +109,8 @@ def __init__(
         assets = with_resources(assets, resource_defs)
         source_assets = with_resources(source_assets, resource_defs)
 
-        self._assets = assets
-        self._source_assets = source_assets
+        self._assets = set(assets)
```
clairelin135, line 112: Since we are making this change, we should cast the `AssetGroup.assets` and `AssetGroup.source_assets` properties to be lists.

Docstrings (and maybe examples?) reference these as lists

clairelin135 (approved):
Thanks for making this fix!

python_modules/dagster/dagster/core/asset_defs/asset_group.py
OwenKephart: yeah you're right I refactored this to not do it that way

PR #8278, opened 2022-06-08
Add typecheck for resolving node config

clairelin135:
Currently, displaying a graph-backed asset on Dagit errors with the following:

> Operation name: AssetGraphQuery
> Message: 'GrapheneCompositeSolidDefinition' object has no attribute 'resolve_config_field'

A config schema can only exist on an op. This PR adds a check to prevent resolving config schema for composite solids (graphs).

Tested in Dagit to confirm that providing launchpad config works and that providing config on `@graph` will propagate to inner ops.


smackesey:
I'm still kind of fuzzy on this and how `config_schema` _should_ work with graph-backed assets, but let's get this in so it acn be picked for release and discuss a deeper fix/improvement later.

smackesey (approved):

PR #8279, opened 2022-06-08
fix group_name in to_source_assets

sryza:
I came across this because it was causing some strange behavior in how assets displayed in Dagit.


dpeng817 (approved):
should probably test but crunch time

PR #8287, opened 2022-06-09
fix bug where issubclass(list[str], DagsterType) throws a surprising exception

gibsondan:
This issue was causing some user jobs to fail that weren't failing before, when we inadvertently made DagsterType an ABC subclass, triggering one of the cpython issues linked in the implementation of is_subclass. 

Create a new implementation that doesn't run afoul of that problem.

### DIFF
--- a/python_modules/dagster/dagster/seven/__init__.py
+++ b/python_modules/dagster/dagster/seven/__init__.py
@@ -168,3 +168,19 @@ def get_import_error_message(import_error):
 @contextmanager
 def nullcontext():
     yield
+
+
+def is_subclass(child_type, parent_type):
+    """Due to some pathological interactions betwen bugs in the Python typing library
+    (https://github.com/python/cpython/issues/88459 and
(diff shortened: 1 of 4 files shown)


python_modules/dagster/dagster/seven/__init__.py
```diff
@@ -168,3 +168,18 @@ def get_import_error_message(import_error):
...
+        return False
+
+    try:
+        return issubclass(child_type, parent_type)
+    except TypeError:
+        # Due to some pathological interactions betwen bugs in the Python typing library
+        # (https://github.com/python/cpython/issues/88459 and
+        # https://github.com/python/cpython/issues/89010)
+        # some types (list[str] in Python 3.9, for example) pass the inspect.isclass check
+        # above and then raise an exception here
```
rexledesma, line 184: nit: make this into the docstring

rexledesma (approved):
pretty gross -

PR #8290, opened 2022-06-09
AssetSelection tweaks

smackesey:
### Summary & Motivation

Tweaks `AssetSelection`:

- `AssetSelection`-creating methods are now runtime-type checked
- `AssetSelection.keys` now takes `CoerceableToAssetKey` instead of just string

Unit tests

### DIFF
--- a/python_modules/dagster/dagster/core/asset_defs/asset_selection.py
+++ b/python_modules/dagster/dagster/core/asset_defs/asset_selection.py
@@ -5,7 +5,7 @@ from typing import AbstractSet, FrozenSet, Optional, Sequence
 
 import dagster._check as check
 from dagster.core.asset_defs.assets import AssetsDefinition
-from dagster.core.definitions.events import AssetKey
+from dagster.core.definitions.events import AssetKey, CoerceableToAssetKey
 from dagster.core.selector.subset_selector import (
     fetch_connected,
     generate_asset_dep_graph,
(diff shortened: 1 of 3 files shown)


smackesey:
* master
  * **PR #8290** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8290" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8290?utm_source=stack-comment).

python_modules/dagster/dagster/core/asset_defs/asset_selection.py
```diff
@@ -19,26 +19,33 @@ def all() -> "AllAssetSelection":
         return AllAssetSelection()
 
     @staticmethod
-    def keys(*key_strs: str) -> "KeysAssetSelection":
-        return KeysAssetSelection(*key_strs)
+    def keys(*key_coercables: CoerceableToAssetKey) -> "KeysAssetSelection":
```
sryza, line 22: Elsewhere, when we have args of type `CoerceableToAssetKey`, we just call them `key` or `asset_key` - doing that here would be more consistent and avoid introducing a term that users are otherwise unfamiliar with.

sryza (approved):
I left one comment. Otherwise, this looks great.

PR #8295, opened 2022-06-09
Add a dagster_task_definition_dict to dagster-aws (not yet used)

gibsondan:
Summary:
Would like to add a path to the EcsRunLauncher creating a task definition from scratch. Pulled this code from the ECS agent for that purpose, will start using it next.

Test Plan: BK


gibsondan:
@jmsanders scoped this down quite a bit compared to the previous rev at your suggestion

jmsanders (approved):

PR #8297, opened 2022-06-09
[dagit] Fix JS error on repo reload

hellendag:
### Summary & Motivation

There is currently a JS error that occurs when reloading a single repo from the left nav while viewing the Instance Overview page. It appears to be due to an Apollo bug: when the store is emptied after reloading the repo, the queries in `InstanceOverviewPage` resolve to `{}` instead of `undefined` while data is being repopulated.

Because our code currently checks whether `data` is truthy in order to determine whether there are values to extract from it, this results in a JS error because fields expected to be there (and typechecked as such) are missing.

To resolve this, use the `loading` values instead. As far as I know, we shouldn't really have to do that, but I don't know if this has been fixed in recent versions of `@apollo/client`.

I also fixed some spacing and truncation issues on the repo name at the foot of the left nav.

View Dagit, reload the current repo via left nav while viewing Instance Overview. Verify that the page reloads and populates properly, with no JS errors.

### DIFF
--- a/js_modules/dagit/packages/core/src/instance/InstanceOverviewPage.tsx
+++ b/js_modules/dagit/packages/core/src/instance/InstanceOverviewPage.tsx
@@ -102,7 +102,7 @@ export const InstanceOverviewPage = () => {
     fetchPolicy: 'network-only',
     notifyOnNetworkStatusChange: true,
   });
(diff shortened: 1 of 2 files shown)


bengotow (approved):
This looks like a good fix for now, I suppose everywhere we check for !data, we could also check `loading`, but what a pain...

PR #8303, opened 2022-06-09
[dagster-dbt] group names on dbt assets

OwenKephart:
### Summary & Motivation

UPDATE:

for now, just support one level of grouping, as this seems the most backwards-compatible.

I also noticed a bug w/ how the group stuff works, where if you remapped the asset keys, the group info would be left behind. This fixes that.

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
@@ -115,12 +115,12 @@ def _get_node_asset_key(node_info: Mapping[str, Any]) -> AssetKey:
 
 
 def _get_node_group_name(node_info: Mapping[str, Any]) -> Optional[str]:
-    """A node's group name is obtained by concatenating the subfolders it resides in (if any)"""
+    """A node's group name is subdirectory that it resides in"""
     fqn = node_info.get("fqn", [])
     # the first component is the package name, and the last component is the model name
     if len(fqn) < 3:
         return None
-    return "_".join(fqn[1:-1])
+    return fqn[1]
 
 
 def _get_node_description(node_info):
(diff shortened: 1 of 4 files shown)


python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
```diff
@@ -62,11 +62,11 @@ def _select_unique_ids_from_manifest_json(
         import dbt.graph.selector as graph_selector
         from dbt.contracts.graph.manifest import Manifest
         from networkx import DiGraph
-    except ImportError:
-        check.failed(
+    except ImportError as e:
+        raise check.CheckError(
```
OwenKephart, line 66: unrelated, I'm terrible

shalabhc (approved):

python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
```diff
@@ -114,6 +114,15 @@ def _get_node_asset_key(node_info):
     return AssetKey(components)
 
 
+def _get_node_group_name(node_info: Mapping[str, Any]) -> Optional[str]:
+    """A node's group name is obtained by concatenating the subfolders it resides in (if any)"""
+    fqn = node_info.get("fqn", [])
+    # the first component is the package name, and the last component is the model name
+    if len(fqn) < 3:
+        return None
+    return "_".join(fqn[1:-1])
```
sryza, line 123: Often subfolders will have underscores, so this naming could be confusing in some cases. Thoughts on what the ideal endstate is here? Hierarchical groups? If we wanted to do that later, would that be a backwards-incompatible change?

OwenKephart: hierarchical groups would be the ideal (I would probably do "/".join(...)).

I suppose it would be an incompatible change in the case that people were doing selections based on this group name.

I believe "/" is not currently supported but maybe we should allow it, and then optionally later on we can give the "/" a special meaning?

sryza: A situation that could get weird:
* We allow "/".
* Someone makes a group called "a/b/c" and also makes a group called "a/b".
* They build a job that selects group="a/b".
* We add hierarchical groups.
* Now that job includes the assets in the "a/b/c" group as well.

OwenKephart: I guess the backwards-compatible way to make the change would be to have the updated AssetSelection.from_group() thing take an "include subgroups" argument (default False). Seems mostly ok to me. The other option would just be to only include the top level group name when we're generating the dbt group names, which also seems fine to me.

shalabhc (approved):
Using the top level group name seems reasonable since we don't really know how people would want their dbt assets grouped .

PR #8307, opened 2022-06-09
[dagit] Small left-nav design adjustments

bengotow:
### Summary & Motivation

This PR makes several small tweaks to the left nav:

- If you only have asset groups or jobs (not both), we don't show the subheadings. In scenarios where you had a lot of repos with just asset groups / jobs, they got kind of repetitive. 

- If you have so many repos that the left nav scrolls, scroll the "active" item offscreen and click a section to expand it, the render would cause the nav to scroll back to the active item. This is annoying because you had to scroll back to the section you opened to click a job / asset group. I memoized the sections so the active section isn't re-rendering and moving you back.

- The width of the nav items was fixed (230px or 250px without a right hand icon), but the width of the nav itself varies based on whether there's a scrollbar. This caused the "Legacy" tag and the "Sensor" icon to overlap if the scrollbar was visible (at least in Firefox for me). To fix this I made the label a flex child with a min-width so it uses the available space, and accounted for the right icon with a margin.

I also fixed a small bug where going to the WorkspaceRoot would jump you to Jobs, which is now Tab #2. To make sure this doesn't happen in the future I replaced the fallthrough path with a reference to the tab path.

Screen Shot 2022-06-09 at 3 00 20 PM

### DIFF
--- a/js_modules/dagit/packages/core/src/ui/SectionedLeftNav.tsx
+++ b/js_modules/dagit/packages/core/src/ui/SectionedLeftNav.tsx
@@ -116,6 +116,11 @@ export const SectionedLeftNav = () => {
 const HEADER_HEIGHT = 48;
 const HEADER_HEIGHT_WITH_LOCATION = 64;
 
+// Note: This component uses React.memo so that it only re-renders when it's props change.
+// This means opening/closing a collapsed section doesn't re-render other sections (a nice
+// perf win) but more importantly opening a section doesn't cause the view to scroll back
+// to the selected item, which could be offscreen.
+//
 interface SectionProps {
   expanded: boolean;
   onToggle: (repoAddress: RepoAddress) => void;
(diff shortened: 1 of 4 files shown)


js_modules/dagit/packages/core/src/ui/SectionedLeftNav.tsx
```diff
@@ -208,47 +213,49 @@ export const Section: React.FC<SectionProps> = (props) => {
...
-      pipelinePath: string;
-    }
-  | {
-      repoPath: string;
-      groupName: string;
-    };
+type PathMatch = {
+  repoPath: string;
+  pipelinePath?: string;
+  groupName?: string;
```
bengotow, line 221: Less technically correct but also less annoying to work with 🙈

js_modules/dagit/packages/core/src/ui/SectionedLeftNav.tsx
```diff
@@ -208,47 +213,49 @@ export const Section: React.FC<SectionProps> = (props) => {
...
-        repoAddress,
-        itemName: explorerPathFromString(match.params.pipelinePath).pipelineName,
-        itemType: 'job' as const,
-      }
-    : {
-        repoAddress,
-        itemName: match.params.groupName,
-        itemType: 'asset-group' as const,
-      };
+  return React.useMemo(() => {
```
bengotow, line 231: Adding this useMemo was the reason for the changes here. Unfortunately the `match` object changes even if all three items inside it are identical strings, so the useMemo couldn't use `match` as it's dep.

PR #8315, opened 2022-06-10
reorder concept super-sections in left nav

sryza:
### Summary & Motivation

This reorders the concept sections based on somewhat-subjective opinions of importance. Some impressions:
* Almost every Dagster user uses Schedules, Sensors, and Partitions.
* Resources are a big source of confusion for early users and it's possible to use Dagster without them, so I moved them down.
* Many users don't use Dagster types.

yarn dev

### DIFF
--- a/docs/content/_navigation.json
+++ b/docs/content/_navigation.json
@@ -135,20 +135,36 @@
         ]
       },
       {
-        "title": "Resources",
+        "title": "Schedules, Sensors, & Partitions",
         "children": [
           {
-            "title": "Resources",
-            "path": "/concepts/resources"
+            "title": "Schedules",
+            "path": "/concepts/partitions-schedules-sensors/schedules"
+          },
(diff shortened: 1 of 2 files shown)


erinkcochran87 (commented):
Thoughts about having the last part of the nav be in this order?

```
- Dagster types
- Logging
- Testing
```

I see logging and testing as related, and could see going from logging to testing as a natural progression.

sryza:
@erinkcochran87 👍, I made the change you suggested.

erinkcochran87 (changes requested):
Just remembered that the sidenav order is also reflected in the **Guides** page. Could you update that to match and then this'll be good to go?

sryza:
@erinkcochran87 I updated the PR to reorder the concept page as well.

erinkcochran87 (approved):

PR #8316, opened 2022-06-10
AssetSelection.assets

sryza:
### Summary & Motivation

I was writing an example for partitioned asset jobs and found it too awkward without this.

```
@asset(partitions_def=hourly_partitions_def)
def asset1():
    ...


@asset(partitions_def=hourly_partitions_def)
def asset2():
    ...


partitioned_asset_job = define_asset_job(
    name="partitioned_job",
    selection=AssetSelection.assets(asset1, asset2),
    partitions_def=hourly_partitions_def,
)
```

Controversial, I know. I think that having the words `AssetSelection` there will help make it clear that we're not actually directly referencing the assets.

Added a test.

### DIFF
--- a/python_modules/dagster/dagster/core/asset_defs/asset_selection.py
+++ b/python_modules/dagster/dagster/core/asset_defs/asset_selection.py
@@ -19,6 +19,12 @@ class AssetSelection(ABC):
         return AllAssetSelection()
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


OwenKephart (approved):
I think this is functionality that we should support, happy with this layer of indirection

PR #8317, opened 2022-06-10
AssetsDefinition property name changes

sryza:
### Summary & Motivation

* group_names -> group_names_by_key
* asset_keys_by_input_name -> keys_by_input_name
* asset_keys_by_output_name -> keys_by_output_name

bk


OwenKephart (commented):
I think this makes sense. I wonder if we could make our lives easier by changing all these args to "..._by_key" (in the same vein as the other pr)

just seems so wordy if you want to set all these parameters manually

shalabhc:
This makes sense. From buildkite, there's a couple of fixes needed in asset_layer and test_external_data

shalabhc (commented):
Added a comment.

OwenKephart (approved):
nice

PR #8325, opened 2022-06-10
AssetsDefinition.asset_keys -> keys and similar

sryza:
### Summary & Motivation

Renames the following properties on `AssetsDefinition`:
- `asset_keys` -> `keys`
- `asset_key` -> `key`
- `dependency_asset_keys` -> `dependency_keys`

I am not 100% confident that this is the right thing to do. The reasoning is basically that `OutputDefinition` has `name`, not `output_name`, `OpDefinition` has `name`, not `op_name`, etc.

bk

### DIFF
--- a/examples/docs_snippets/docs_snippets_tests/concepts_tests/assets_tests/test_multi_assets.py
+++ b/examples/docs_snippets/docs_snippets_tests/concepts_tests/assets_tests/test_multi_assets.py
@@ -7,16 +7,16 @@ from docs_snippets.concepts.assets.multi_assets import (
 
 
 def test_basic():
-    assert len(my_function.asset_keys) == 2
+    assert len(my_function.keys) == 2
 
 
 def test_io_manager():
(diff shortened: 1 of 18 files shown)


clairelin135 (commented):
I think we should choose between `AssetKey` and `Key` and standardize across `.asset_keys`, `SourceAsset.key`, `.dependency_asset_keys`, and `AssetKey()`.

I think it's fine to convert all of these to `key` because it will be less verbose for users to specify, but then I think we should also modify `AssetKey`. I also like that converting to `key` will unify this with `SourceAsset.key`.

OwenKephart (approved):
I like this one. We also have "key_prefix", not "asset_key_prefix", so this lines up with that pattern. it's exhausting typing "asset" a billion times.

sryza:
@clairelin135 good point about SourceAsset - I had forgotten about that.

I think the ship has sailed on `AssetKey` - as that's not experimental like the rest of the SDA APIs.

PR #8328, opened 2022-06-10
[docs] - Clarify explanation of graph-backed assets

sryza:


docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -198,9 +198,7 @@ In this example, Dagster doesn’t need to load data from `upstream_asset` to su
 
 ### Graph-backed assets
 
-[Basic software-defined assets](#a-basic-software-defined-asset) can only produce one data artifact. If generating an asset involves multiple discrete computations, you can use graph-backed assets by separating each computation into an op and building a graph to combine your computations. This way, each discrete computation can be reused in other assets and jobs.
-
-This approach to asset creation allows you to define a graph of ops which can produce one or multiple assets. **Note**: To use graphs to create an asset, the graph **must return an output**.
```
yuhan, line 203: I liked this phrase - it made sense to me that graph-backed assets are assets that generated from graphs.

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -198,9 +198,7 @@ In this example, Dagster doesn’t need to load data from `upstream_asset` to su
...
+[Basic software-defined assets](#a-basic-software-defined-asset) are computing a single op. If generating an asset involves multiple discrete computations, you can use graph-backed assets by separating each computation into an op and building a graph to combine your computations. This allows you can launch re-executions of runs at the op boundaries, but doesn't require you to link each intermediate value to an asset in persistent storage.
```
yuhan, line 201: I think this makes sense for folks who are primarily using assets. 

However, my confusion was coming as an existing user who is familiar with graph/job/op but trying to connect the dots between graph and assets, which i think represents many of our current users. how about also including explanation for these users. 

For example, [here](https://dagster.slack.com/archives/C01U954MEER/p1652107591985779) is a similar situation from dagster-support and how the user asked for this feature:
> I wrote the pipeline in ops/graphs first. Then I realize that asset helps me to keep track some of the external entities. However, not all ops in my graph has to be assets. So I want to migrate only several ops to assets, the others still the same ops.

I feel this API could be a main path for folks to incrementally migrate from graphs to assets.

Maybe something like:
"If you have an existing graph and you'd like to connect it with other assets in a job, you can generate assets from the graph and ..."
- another thing unclear to me is the `...` part: after defining the asset from `graph_asset = AssetsDefinition.from_graph(store_slack_files_in_sql)`, how do i link it with other assets? (e.g. is it via `AssetIn` and where do i specify it? can it be downstream of other asset and how?) including answers to these questions in the code example would be great.

docs/content/concepts/assets/software-defined-assets.mdx
sryza: I had taken it out because it felt redundant with this sentence: "If generating an asset involves multiple discrete computations, you can use graph-backed assets by separating each computation into an op and building a graph to combine your computations." Is there information in the latter sentence that's missing in the former?

docs/content/concepts/assets/software-defined-assets.mdx
sryza: I strongly agree that we need to help users make this connection.

Question that I could be convinced either way on: is this the right place to make that point? We have an example in the upgrade guide that shows how to do this: https://docs.dagster.io/master/guides/dagster/enriching-with-software-defined-assets#not-all-ops-produce-assets. Though maybe that's too buried? We could also discuss it on the graph page?

sryza: > another thing unclear to me is the ... part: after defining the asset from graph_asset = AssetsDefinition.from_graph(store_slack_files_in_sql), how do i link it with other assets? (e.g. is it via AssetIn and where do i specify it? can it be downstream of other asset and how?) including answers to these questions in the code example would be great.

@clairelin135 would you be able to work with @erinkcochran87  to address this?

yuhan: > Question that I could be convinced either way on: is this the right place to make that point? We have an example in the upgrade guide that shows how to do this: https://docs.dagster.io/master/guides/dagster/enriching-with-software-defined-assets#not-all-ops-produce-assets. Though maybe that's too buried? We could also discuss it on the graph page?

Yea that feels a bit buried inside the upgrade guide. While the upgrade guide does need to detail about this, I also think we should cover this here as well, because 1) it'll be not only a migration story but also a long-living op asset interplay use case 2) it feels natural to learn assets on the asset page (aka here) and try to wrap head around with users current mental model with graph while reading about assets. So, I think we should have it in both pages but slightly differ the perspectives - upgrade guide leans towards "how can i migrate graphs to assets" and asset concept leans towards "how can i connect graphs where parts don't unnecessarily in assets with other assets, e.g. reusability"

clairelin135: Thanks @yuhan for voicing the perspective of someone who's not super familiar with graph-backed assets.

> another thing unclear to me is the ... part: after defining the asset from graph_asset = AssetsDefinition.from_graph(store_slack_files_in_sql), how do i link it with other assets? (e.g. is it via AssetIn and where do i specify it? can it be downstream of other asset and how?) including answers to these questions in the code example would be great.

I'm going to file a PR to add more code examples here that document how to specify upstream/downstream dependencies. I'll add sections that include basic dependencies and explicit dependencies.

I agree that this is valuable information to detail on the software-defined assets page.

docs/content/concepts/assets/software-defined-assets.mdx
yuhan: After the discussion we had on the other comment, I think this comment was more related to the graph to asset path, as opposed to separating one asset to multiple ops. The sentence you quoted is sufficient and the other discussion should be able to address my confusion.

docs/content/concepts/assets/software-defined-assets.mdx
sryza: > Yea that feels a bit buried inside the upgrade guide.

@yuhan that makes sense to me. I'll update this to add verbiage similar to what you suggested.

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -198,9 +198,9 @@ In this example, Dagster doesn’t need to load data from `upstream_asset` to su
 
 ### Graph-backed assets
 
-[Basic software-defined assets](#a-basic-software-defined-asset) can only produce one data artifact. If generating an asset involves multiple discrete computations, you can use graph-backed assets by separating each computation into an op and building a graph to combine your computations. This way, each discrete computation can be reused in other assets and jobs.
+[Basic software-defined assets](#a-basic-software-defined-asset) are computed using a single op. If generating an asset involves multiple discrete computations, you can use graph-backed assets by separating each computation into an op and building a graph to combine your computations. This allows you can launch re-executions of runs at the op boundaries, but doesn't require you to link each intermediate value to an asset in persistent storage.
```
erinkcochran87, line 201:
```suggestion
[Basic software-defined assets](#a-basic-software-defined-asset) are computed using a single op. If generating an asset involves multiple discrete computations, you can use graph-backed assets by separating each computation into an op and building a graph to combine your computations. This allows you to launch re-executions of runs at the op boundaries but doesn't require you to link each intermediate value to an asset in persistent storage.
```

Fixing a typo, removing unneeded comma.

erinkcochran87 (changes requested):
I think this, in combination with Claire's PR (#8339), will address Yuhan's excellent point about this section. I just had a comment about some grammatical stuff.

yuhan (approved):
i like it 👍

PR #8329, opened 2022-06-10
Attempt to speed up query for existing run keys for a given sensor

gibsondan:
Summary:
DB profiling suggests that the existing query here is quite expensive. Instead of doing a join across sensor name, filter by run key and then filter down by sensor name

The way that this could go wrong is if a user is sharing run key formats across lots of sensors - but I think that is less likely to happen than the current situation, and there are mitigations we can apply if we observe that.

### DIFF
--- a/python_modules/dagster/dagster/core/storage/runs/sql_run_storage.py
+++ b/python_modules/dagster/dagster/core/storage/runs/sql_run_storage.py
@@ -230,7 +230,14 @@ class SqlRunStorage(RunStorage):  # pylint: disable=no-init
             query = query.where(
                 db.or_(
                     *(
-                        db.and_(RunTagsTable.c.key == key, RunTagsTable.c.value == value)
+                        db.and_(
+                            RunTagsTable.c.key == key,
+                            (
+                                RunTagsTable.c.value == value
+                                if isinstance(value, str)
+                                else RunTagsTable.c.value.in_(value)
+                            ),
+                        )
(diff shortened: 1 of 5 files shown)


python_modules/dagster/dagster/core/storage/runs/sql_run_storage.py
```diff
@@ -230,7 +230,14 @@ def _add_filters_to_query(self, query, filters: RunsFilter):
             query = query.where(
                 db.or_(
                     *(
-                        db.and_(RunTagsTable.c.key == key, RunTagsTable.c.value == value)
+                        db.and_(
+                            RunTagsTable.c.key == key,
+                            (
+                                RunTagsTable.c.value == value
+                                if isinstance(value, str)
+                                else RunTagsTable.c.value.in_(value)
```
prha, line 238: this is better yeah...  not sure why I didn't think about this before.

python_modules/dagster/dagster/daemon/sensor.py
```diff
@@ -512,39 +512,11 @@ def _fetch_existing_runs(instance, external_sensor, run_requests):
     if not run_keys:
         return {}
 
-    existing_runs = {}
-
-    if instance.supports_bucket_queries:
```
prha, line 517: yeah, I guess if we're doing an existence check, we don't need to bucket...

prha (approved):

PR #8330, opened 2022-06-10
[with-resources changes 2/n] If resources collide when using with_resources, error.

dpeng817:
Previously, with_resources allowed for resource _overrides_. That is, if a resource key was present on an asset definition or source asset already, and the same resource key was provided to with_resources, then using with_resources would passively bring the key provided to the asset. This PR changes that behavior to error if a key collision occurs when using with_resources.

### DIFF
--- a/python_modules/dagster/dagster/core/asset_defs/assets.py
+++ b/python_modules/dagster/dagster/core/asset_defs/assets.py
@@ -418,7 +418,7 @@ class AssetsDefinition(ResourceAddable):
     def required_resource_keys(self) -> Set[str]:
         return {requirement.key for requirement in self.get_resource_requirements()}
 
-    def to_string(self):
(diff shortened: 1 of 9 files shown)


python_modules/dagster/dagster/core/asset_defs/assets.py
```diff
@@ -416,9 +418,29 @@ def get_resource_requirements(self) -> Iterator[ResourceRequirement]:
     def required_resource_keys(self) -> Set[str]:
         return {requirement.key for requirement in self.get_resource_requirements()}
 
+    def to_string(self):
```
sryza, line 421: `__str__` is preferred over `to_string`

sryza (approved):
Looks good aside from the comment I left

PR #8332, opened 2022-06-10
uniform prefix parsing

OwenKephart:
### Summary & Motivation

We were sort of all over the place on this. Now we have a single code path / type for things that we want to coerce to prefixes.

Right now it's just string and list of strings.

Also I fixed the spelling on the existing CoerceableToAssetKey to CoercibleToAssetKey

unit tests


sryza (approved):
Nice

PR #8333, opened 2022-06-10
in load_assets_from_... functions, make key_prefix docstring consiste…

sryza:
…nt with group


clairelin135 (approved):
Thanks for making this change!

PR #8336, opened 2022-06-10
[with-resources changes 3/n] without_resources

dpeng817:
Adds a method `without_resources` which removes all resource definitions attached to an AssetsDefinition or SourceAsset. There is some intricacy in the implementation, because in order to support directly providing io manager definitions on assets, we set a system-generated io manager key on asset definitions. When removing resources, we would want to remove that key and use the default key `io_manager` instead, since the user is never exposed to the original io manager key.

### DIFF
--- a/python_modules/dagster/dagster/core/asset_defs/assets.py
+++ b/python_modules/dagster/dagster/core/asset_defs/assets.py
@@ -581,22 +581,14 @@ class AssetsDefinition(ResourceAddable):
         if isinstance(self.node_def, OpDefinition):
 
             output_defs = []
+            replaced_io_manager_keys = set()
             for output_def in self.op.output_defs:
                 asset_key = self.keys_by_output_name.get(output_def.name)
                 if output_def.io_manager_key == io_manager_key_for_asset_key(asset_key):
                     io_manager_key = DEFAULT_IO_MANAGER_KEY
-                    new_required_keys = {
-                        io_manager_key,
-                        *{
-                            key
-                            for key in self.op.required_resource_keys
-                            if key != io_manager_key_for_asset_key(self.asset_key)
-                        },
-                    }
-
+                    replaced_io_manager_keys.add(io_manager_key_for_asset_key(asset_key))
                 else:
                     io_manager_key = output_def.io_manager_key
-                    new_required_keys = self.op.required_resource_keys
 
                 output_def = OutputDefinition(
                     dagster_type=output_def.dagster_type,
@@ -611,6 +603,18 @@ class AssetsDefinition(ResourceAddable):
                 )
                 output_defs.append(output_def)
 
+            if replaced_io_manager_keys:
+                required_resource_keys = {
+                    DEFAULT_IO_MANAGER_KEY,
+                    *{
+                        key
+                        for key in self.op.required_resource_keys
+                        if key not in replaced_io_manager_keys
+                    },
+                }
+            else:
+                required_resource_keys = self.op.required_resource_keys
+
             node_def = OpDefinition(
                 name=self.op.name,
                 input_defs=self.op.input_defs,
@@ -619,7 +623,7 @@ class AssetsDefinition(ResourceAddable):
                 config_schema=self.op.config_schema,
                 description=self.op.description,
                 tags=self.op.tags,
-                required_resource_keys=new_required_keys,
+                required_resource_keys=required_resource_keys,
                 version=self.op.version,
                 retry_policy=self.op.retry_policy,
             )
--- a/python_modules/dagster/dagster/core/asset_defs/assets.py
+++ b/python_modules/dagster/dagster/core/asset_defs/assets.py
@@ -21,6 +21,7 @@ from dagster.core.definitions import (
     ResourceDefinition,
 )
 from dagster.core.definitions.events import AssetKey
+from dagster.core.definitions.output import OutputDefinition
 from dagster.core.definitions.partition import PartitionsDefinition
 from dagster.core.definitions.utils import DEFAULT_GROUP_NAME, validate_group_name
 from dagster.core.errors import DagsterInvalidInvocationError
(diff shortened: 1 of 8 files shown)


python_modules/dagster/dagster/core/asset_defs/assets.py
```diff
@@ -396,9 +399,29 @@ def get_resource_requirements(self) -> Iterator[ResourceRequirement]:
     def required_resource_keys(self) -> Set[str]:
         return {requirement.key for requirement in self.get_resource_requirements()}
 
+    def to_string(self):
+        if len(self.asset_keys) == 1:
+            return f"AssetsDefinition with key {self.asset_key.to_string()}"
+        else:
+            asset_keys = ", ".join(
```
OwenKephart, line 406: tiny nit: for dbt assets (and, presumably, others) the list of asset keys can get really large (>100). Maybe just show the first 5 and then "..."?

Doesn't need to be in this pr.

python_modules/dagster/dagster/core/asset_defs/assets.py
```diff
@@ -428,6 +451,69 @@ def with_resources(self, resource_defs: Mapping[str, ResourceDefinition]) -> "As
...
+                is_required=output_def.is_required,
+                io_manager_key=io_manager_key,
+                metadata=output_def.metadata,
+                asset_key=output_def._asset_key,
+                asset_partitions=output_def._asset_partitions_fn,
+                asset_partitions_def=output_def._asset_partitions_def,
+            )
+
+            node_def = OpDefinition(
+                name=op_def.name,
```
OwenKephart, line 491: In practice, I think this will be ok, but I'm a little wary of this sort of implementation (because when you do this, you end up with two conflicting OpDefinitions with the same name).

This only mattes once you put this asset on a repository, which you should never be doing if I'm understanding this correctly, but still.

Also seems like a pain / fairly brittle to do all this copying.

Can we do something fancier here, where we keep around a list of keys for the io manager defs that were removed? something like self._removed_io_manager_keys

Then when we call with_resources, we can map whatever resource_def is provided for key "io_manager" to all of the "self._removed_io_manager_keys".

kinda ugly but maybe there's a way to riff on that which feels more elegant. just seems like in the future if a field is added to any of OpDefinition, AssetsDefinition, or OutputDefinition, this code could potentially break

OwenKephart: i guess this would only work if we expect the user to invoke with_resources to add resources back on (which I'm not sure is the case)

dpeng817: While I think it's probably fine that the op has the same name, and we could add an optional name parameter as an escape hatch, totally agreed that this implementation is pretty jank. I like your idea of keeping around some sort of reference to the old key, it feels a bit less invasive.

python_modules/dagster/dagster/core/asset_defs/assets.py
dpeng817: Ugh yea good point.

python_modules/dagster/dagster/core/asset_defs/assets.py
```diff
@@ -396,9 +399,29 @@ def get_resource_requirements(self) -> Iterator[ResourceRequirement]:
     def required_resource_keys(self) -> Set[str]:
         return {requirement.key for requirement in self.get_resource_requirements()}
 
+    def to_string(self):
```
sryza, line 402: Any particular reason to call this `to_string` instead of `__str__`. Even though we use `to_string` in some places in our codebase, I think it's generally considered better practice to use `__str__`.

dpeng817: potential user impact from users calling str(assets_def) is what I was thinking, but if that's not a concern I can change.

python_modules/dagster/dagster/core/asset_defs/assets.py
```diff
@@ -570,6 +575,69 @@ def with_resources(self, resource_defs: Mapping[str, ResourceDefinition]) -> "As
...
+                        *{
+                            key
+                            for key in self.op.required_resource_keys
+                            if key != io_manager_key_for_asset_key(self.asset_key)
+                        },
+                    }
+
+                else:
+                    io_manager_key = output_def.io_manager_key
+                    new_required_keys = self.op.required_resource_keys
```
OwenKephart, line 599: this seems like the wrong place to put this logic, as new_requried_keys is dependent on whichever output definition is the last in the list. I don't think required_resource_keys will contain the keys of io managers in the general case, so it's probably fine to omit this part.

python_modules/dagster/dagster/core/asset_defs/assets.py
```diff
@@ -570,6 +575,69 @@ def with_resources(self, resource_defs: Mapping[str, ResourceDefinition]) -> "As
...
+
+            node_def = OpDefinition(
+                name=self.op.name,
+                input_defs=self.op.input_defs,
+                output_defs=output_defs,
+                compute_fn=self.op.compute_fn,
+                config_schema=self.op.config_schema,
+                description=self.op.description,
+                tags=self.op.tags,
+                required_resource_keys=new_required_keys,
```
OwenKephart, line 622: see above

python_modules/dagster/dagster/core/asset_defs/assets.py
```diff
@@ -570,6 +575,69 @@ def with_resources(self, resource_defs: Mapping[str, ResourceDefinition]) -> "As
...
+            for output_def in self.op.output_defs:
+                asset_key = self.keys_by_output_name.get(output_def.name)
+                if output_def.io_manager_key == io_manager_key_for_asset_key(asset_key):
+                    io_manager_key = DEFAULT_IO_MANAGER_KEY
+                    new_required_keys = {
+                        io_manager_key,
+                        *{
+                            key
+                            for key in self.op.required_resource_keys
+                            if key != io_manager_key_for_asset_key(self.asset_key)
```
OwenKephart, line 593: self.asset_key will error for multi assets

OwenKephart (changes requested):
I feel like the dagit thing should not be a concern (as users shouldn't be pulling up things that have without_resources() called on them in dagit -- this feature is for unit tests if I'm understanding correctly), but im ok with the copy approach for now (in the future, we should probably have a standard way to copy classes like OpDef and JobDef).

had some unrelated comments

python_modules/dagster/dagster_tests/core_tests/resource_tests/test_without_resources.py
```diff
@@ -0,0 +1,198 @@
...
+    def the_asset():
+        pass
+
+    assert the_asset.required_resource_keys == {
+        "foo",
+        io_manager_key_for_asset_key(the_asset.asset_key),
+    }
+    transformed_asset = without_resources([the_asset])[0]
+    # Transitive resource dependencies are kept around, because there are is no way to specify resources as optional.
+    assert transformed_asset.required_resource_keys == {"foo", DEFAULT_IO_MANAGER_KEY}
```
dpeng817, line 198: this is an unfortunate little tidbit. When a user provides a resource def directly to an asset, we can't know their intent. In order to make it possible to use provided resources directly within the op, we make them required on the op itself. This means that if I provide a resource required by my io manager, and then remove all resources, I still require the key transitively required by the io manager. If we had a notion of optional resources, we could default to making provided resource defs optional resource keys, so that it wouldn't be necessary to re-provide that value.

OwenKephart: ah I see, that makes sense

OwenKephart (approved):

PR #8344, opened 2022-06-12
add tags in run_request_for_partition

jburnich:
### Summary & Motivation

I use a sensor that triggers partitioned queries. I need to have additional tags as it is already possible to add when running a job. Here, the update function allows to overwrite the tags that have the same key as the partition_tags, which guarantees to always have the authentic partition tags

```
@sensor(job=job)
def my_sensor(context):
    since_key = context.cursor or None
    new_s3_keys = get_s3_keys(
        "my_bucket", prefix="my_prefix/", since_key=since_key, s3_session=S3_session
    )
    if not new_s3_keys:
        return SkipReason("No new s3 files found for bucket my_s3_bucket.")
    run_requests = [
        job.run_request_for_partition(
            partition_key="2022-06-12", run_key=s3_key, tags={"my_tag": "my_value"}
        )
        for s3_key in new_s3_keys
    ]
    context.update_cursor(new_s3_keys[-1])
    return run_requests

```

### DIFF
--- a/python_modules/dagster/dagster/core/definitions/job_definition.py
+++ b/python_modules/dagster/dagster/core/definitions/job_definition.py
@@ -431,8 +431,6 @@ class JobDefinition(PipelineDefinition):
         run_key: Optional[str],
         tags: Optional[Dict[str, str]] = None,
     ) -> RunRequest:
-        print(tags)
-        print(run_key)
         partition_set = self.get_partition_set_def()
         if not partition_set:
             check.failed("Called run_request_for_partition on a non-partitioned job")
--- a/python_modules/dagster/dagster/core/definitions/job_definition.py
+++ b/python_modules/dagster/dagster/core/definitions/job_definition.py
@@ -425,14 +425,17 @@ class JobDefinition(PipelineDefinition):
 
         return mode.partitioned_config.partitions_def
 
-    def run_request_for_partition(self, partition_key: str, run_key: Optional[str]) -> RunRequest:
+    def run_request_for_partition(
+        self, partition_key: str, run_key: Optional[str], tags: Optional[dict]
+    ) -> RunRequest:
         partition_set = self.get_partition_set_def()
         if not partition_set:
             check.failed("Called run_request_for_partition on a non-partitioned job")
 
         partition = partition_set.get_partition(partition_key)
         run_config = partition_set.run_config_for_partition(partition)
-        tags = partition_set.tags_for_partition(partition)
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/core/definitions/job_definition.py
```diff
@@ -425,14 +425,23 @@ def partitions_def(self) -> Optional[PartitionsDefinition]:
...
+    ) -> RunRequest:
         partition_set = self.get_partition_set_def()
         if not partition_set:
             check.failed("Called run_request_for_partition on a non-partitioned job")
 
         partition = partition_set.get_partition(partition_key)
         run_config = partition_set.run_config_for_partition(partition)
-        tags = partition_set.tags_for_partition(partition)
+        if tags:
+            tags.update(partition_set.tags_for_partition(partition))
```
prha, line 441: I think that this will mutate the passed in dictionary, which might have unexpected consequences for the caller.   Could be safer to use this pattern instead:
```
run_request_tags = {**partition_set.tags_for_partition(partition), **tags}
```

python_modules/dagster/dagster_tests/execution_tests/test_job.py
```diff
@@ -164,6 +164,13 @@ def my_job():
         assert run_request.tags
         assert run_request.tags.get(PARTITION_NAME_TAG) == partition_key
 
+        run_request_with_tags = my_job.run_request_for_partition(
+            partition_key=partition_key, run_key=None, tags={"foo": "bar"}
+        )
+        assert run_request_with_tags.run_config == partition_fn(partition_key)
+        assert run_request_with_tags.tags
+        assert run_request_with_tags.tags.get(PARTITION_NAME_TAG) == partition_key
```
prha, line 172: maybe we should add a check for the user-provided tags?

```
assert run_request_sith_tags.tags.get("foo") == "bar"
```

prha (changes requested):
I think this looks good overall, some small changes req'd

python_modules/dagster/dagster/core/definitions/job_definition.py
prha: I think, in practice, it doesn't matter that much.  People don't tend to pass in / override dagster tags.

As a pattern, we tend to put the user-provided tags last, since those are the most confounding in terms of wondering why the thing you provided is not available as a tag.

prha (approved):

PR #8352, opened 2022-06-13
Allow number type for id in ButtonGroup

salazarm:
### Summary & Motivation

Update ButtonGroup to allow a number type (for typescript Enums which by default are number based if no value is specified/required) since a Key can be either a string or a number


hellendag:
See the repository filter dialog (via left nav) as an example, or lists of schedules, sensors, or assets.

hellendag (approved):

PR #8356, opened 2022-06-13
Assorted type annotations

smackesey:
### Summary & Motivation

This is a large PR with a lot of type annotation improvements and a few other minor changes. Unfortunately it is difficult to break up into anything smaller because typing changes cause cascading mypy errors-- adding annotations in one place causes errors to surface in another place, which then needs type annotations added, etc. However review shouldn't be too difficult since most of the changes are pretty trivial:

- Added declarations of instance attributes to several classes. In some cases corresponding type declarations at the point of assignment of instance variables were removed.
- Many type annotations using Dict/List/Set replaced with Mapping/Sequence/AbstractSet respectively. Sometimes
  corresponding runtime type checks were also swapped (e.g. `check.list_param` > `check.sequence_param`)
- Added many missing argument and return type annotations
- Factored out a few complex Callable or Union types into more readable type aliases
- Fixed various incorrect annotations
- Added some type-ignores where added annotations triggered mypy bugs

In several other places `check.not_none` was inserted where there was an implicit assumption that a value was non-null, but this wasn't inferrable by the type-checker.

There are also a few minor runtime changes:

- dagster.config.traversal_context: config schema args should be non-optional
- dagster.core.definitions.configurable: use isinstance (legible to type-checkers) instead of custom method that calls
  isinstance

Finally, internal requires updates to match these, here's the sister PR: https://github.com/dagster-io/internal/pull/2498. That should be merged after this, which will need to be merged with the internal compatibility check still red.

Existing test suite.

### DIFF
--- a/python_modules/dagster/dagster/core/definitions/asset_layer.py
+++ b/python_modules/dagster/dagster/core/definitions/asset_layer.py
@@ -404,7 +404,7 @@ class AssetLayer:
     _dependency_node_handles_by_asset_key: Mapping[AssetKey, Set[NodeHandle]]
     _source_assets_by_key: Mapping[AssetKey, "SourceAsset"]
     _asset_defs_by_key: Mapping[AssetKey, "AssetsDefinition"]
-    _asset_defs_by_node_handle: Mapping[NodeHandle, Set[AssetsDefinition]]
+    _asset_defs_by_node_handle: Mapping[NodeHandle, Set["AssetsDefinition"]]
     _io_manager_keys_by_asset_key: Mapping[AssetKey, str]
 
     def __init__(
--- a/python_modules/dagster/dagster/core/definitions/asset_layer.py
+++ b/python_modules/dagster/dagster/core/definitions/asset_layer.py
@@ -532,7 +532,7 @@ class AssetLayer:
                 inner_output_def, inner_node_handle = assets_def.node_def.resolve_output_to_origin(
                     output_name, handle=node_handle
                 )
-                node_output_handle = NodeOutputHandle(inner_node_handle, inner_output_def.name)
+                node_output_handle = NodeOutputHandle(check.not_none(inner_node_handle), inner_output_def.name)
                 partition_fn = lambda context: {context.partition_key}
                 asset_info_by_output[node_output_handle] = AssetOutputInfo(
                     asset_key,
--- a/python_modules/dagster/dagster/core/definitions/asset_layer.py
+++ b/python_modules/dagster/dagster/core/definitions/asset_layer.py
@@ -19,6 +19,8 @@ from typing import (
 )
 
 import dagster._check as check
+from dagster.core.definitions.events import AssetKey
+from dagster.core.definitions.metadata import MetadataUserInput, RawMetadataValue
 from dagster.core.selector.subset_selector import AssetSelectionData
 from dagster.utils.backcompat import ExperimentalWarning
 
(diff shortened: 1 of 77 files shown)


sryza:
Is this meant to be marked as a draft?

smackesey:
>Whoops forgot to unmark it

python_modules/dagster/dagster_tests/core_tests/definitions_tests/test_schedule_invocation.py
```diff
@@ -15,12 +13,6 @@
 from dagster.core.test_utils import instance_for_test
 
 
-def test_schedule_context_backcompat():
-    # If an instance of ScheduleEvaluationContext is a ScheduleExecutionContext, then annotating as
-    # ScheduleExecutionContext and passing in a ScheduleEvaluationContext should pass mypy
-    assert isinstance(ScheduleEvaluationContext(None, None), ScheduleExecutionContext)
```
alangenfeld, line 21: it would be surprising to accidentally delete since theres a comment next to it - but you could preserve a test here that just annotates some functions as am additional guard

alangenfeld (approved):

python_modules/dagster/dagster/core/definitions/asset_layer.py
```diff
@@ -397,6 +398,15 @@ class AssetLayer:
...
     """
 
+    _asset_keys_by_node_input_handle: Mapping[NodeInputHandle, AssetKey]
+    _asset_info_by_node_output_handle: Mapping[NodeOutputHandle, AssetOutputInfo]
+    _asset_deps: Mapping[AssetKey, AbstractSet[AssetKey]]
+    _dependency_node_handles_by_asset_key: Mapping[AssetKey, Set[NodeHandle]]
+    _source_assets_by_key: Mapping[AssetKey, "SourceAsset"]
+    _asset_defs_by_key: Mapping[AssetKey, "AssetsDefinition"]
+    _asset_defs_by_node_handle: Mapping[NodeHandle, Set["AssetsDefinition"]]
+    _io_manager_keys_by_asset_key: Mapping[AssetKey, str]
```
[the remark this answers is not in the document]
smackesey: A few reasons:

- This is the standard way to annotate instance attrs in Python-- see [PEP 526](https://peps.python.org/pep-0526/#class-and-instance-variable-annotations). There is a special `ClassVar` annotation you use for class variables.
- IMO it is a lot more readable than annotations in `__init__.

> I'm surprised the types didnt propagate from __init__ correctly though, i guess issues witch check calls?

- Yes, the types are often not inferred correctly in `__init__`. I _think_ it would work the same as the above style if we explicitly annotated the type for every assignment in `__init__`, but I'm not 100% sure. Regardless, it's a lot harder to read/find the declared type when its embedded in often lengthy `__init__` logic.
- The above style matches the style used by dataclasses and pydantic, which I still hope to transition to

python_modules/dagster/dagster_tests/core_tests/definitions_tests/test_schedule_invocation.py
smackesey: I actually realize now that I made a mistake in making `ScheduleExecutionContext` a type alias, because it's a literal alias of `ScheduleEvaluationContext` that's exported top-level. I guess we'll remove with 1.0 but for now I'll restore it (and the test), though the comment I think is wrong, we're not testing if mypy passes here but rather python `isinstance`

PR #8357, opened 2022-06-13
refactor stephandler API to support the same yield+write pattern as other executors

gibsondan:
This makes stephandlers follow the same write-and-yield pattern that the rest of our event write paths do. I don't actually love that pattern, but I think there is value in consistency there. This will allow us to use a consistent event write API in all our executors as we are adding new structured events like step process start / step process started / etc.

### DIFF
--- a/python_modules/dagster/dagster/core/events/__init__.py
+++ b/python_modules/dagster/dagster/core/events/__init__.py
@@ -668,13 +668,18 @@ class DagsterEvent(
 
     @staticmethod
     def step_failure_event(
-        step_context: IStepContext, step_failure_data: "StepFailureData"
+        step_context: IStepContext,
+        step_failure_data: "StepFailureData",
+        message=None,
     ) -> "DagsterEvent":
         return DagsterEvent.from_step(
             event_type=DagsterEventType.STEP_FAILURE,
             step_context=step_context,
             event_specific_data=step_failure_data,
-            message='Execution of step "{step_key}" failed.'.format(step_key=step_context.step.key),
+            message=(
+                message
+                or 'Execution of step "{step_key}" failed.'.format(step_key=step_context.step.key)
+            ),
         )
 
     @staticmethod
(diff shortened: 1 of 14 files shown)


gibsondan:
@johannkm I have some tests to fix still and likely some bugs to fix too but lmk if you have any thoughts on the overall API here

python_modules/libraries/dagster-k8s/dagster_k8s/executor.py
```diff
@@ -220,26 +220,20 @@ def launch_step(self, step_handler_context: StepHandlerContext):
...
-                step_key=step_key,
-                message=f"Executing step {step_key} in Kubernetes job {job_name}",
-                event_specific_data=EngineEventData(
-                    [
-                        MetadataEntry("Step key", value=step_key),
-                        MetadataEntry("Kubernetes Job name", value=job_name),
-                    ],
-                ),
-            )
+        yield DagsterEvent.engine_event(
```
johannkm, line 223: definitely an improvement to log these events immediately, instead of once the fn returned like before

python_modules/dagster/dagster/core/events/__init__.py
```diff
@@ -668,13 +668,18 @@ def step_output_event(
...
+        message=None,
     ) -> "DagsterEvent":
         return DagsterEvent.from_step(
             event_type=DagsterEventType.STEP_FAILURE,
             step_context=step_context,
             event_specific_data=step_failure_data,
-            message='Execution of step "{step_key}" failed.'.format(step_key=step_context.step.key),
+            message=(
+                message
+                or 'Execution of step "{step_key}" failed.'.format(step_key=step_context.step.key)
```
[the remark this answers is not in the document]
gibsondan: Added the step key to the callsite where we override this (it's also included in the row in the first column in the event log)

johannkm (approved):

python_modules/dagster/dagster/core/executor/step_delegating/step_handler/base.py
```diff
@@ -52,13 +81,13 @@ def name(self) -> str:
         pass
 
     @abstractmethod
-    def launch_step(self, step_handler_context: StepHandlerContext) -> List[DagsterEvent]:
+    def launch_step(self, step_handler_context: StepHandlerContext) -> Iterator[DagsterEvent]:
         pass
 
     @abstractmethod
-    def check_step_health(self, step_handler_context: StepHandlerContext) -> List[DagsterEvent]:
+    def check_step_health(self, step_handler_context: StepHandlerContext) -> CheckStepHealthResult:
```
johannkm, line 88: much better api

PR #8358, opened 2022-06-13
[docs] - Move assets into intro tutorial + up in sidenav [CON-18] (#8…

yuhan:
test to see if cherry-picking https://github.com/dagster-io/dagster/commit/02e9d20aee1ddcdc69bdcce2f530dae1abb828aa would fix the latest version of the docs

if so, we would do a hot fix which would include:
- manually backfill 0.14.20 entry with the navigation here in https://github.com/dagster-io/dagster/blob/master/docs/next/.versioned_content/_versioned_navigation.json
- manually update the _navigation in 0.14.20 versioned s3 folder


yuhan:
actually not this route - need to update https://github.com/dagster-io/dagster/blob/master/docs/next/.versioned_content/_versioned_navigation.json first

PR #8364, opened 2022-06-13
materialize_in_process method

dpeng817:
Adds a materialize_in_process method which uses the in memory io manager and in process executor to load assets. For use with testing.

### DIFF
--- a/python_modules/dagster/dagster/core/asset_defs/materialize.py
+++ b/python_modules/dagster/dagster/core/asset_defs/materialize.py
@@ -1,7 +1,8 @@
-from typing import Any, Optional, Sequence, Union
+from typing import Any, Mapping, Optional, Sequence, Union
 
 import dagster._check as check
 
+from ..execution.build_resources import wrap_resources_for_execution
 from ..execution.execute_in_process_result import ExecuteInProcessResult
 from ..execution.with_resources import with_resources
 from ..instance import DagsterInstance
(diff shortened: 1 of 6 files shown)


clairelin135 (approved):
LGTM!

python_modules/dagster/dagster/core/asset_defs/materialize.py
```diff
@@ -40,3 +42,43 @@ def materialize(
         assets=assets_defs,
         source_assets=source_assets,
     ).execute_in_process(run_config=run_config, instance=instance)
+
+
+def materialize_in_process(
```
sryza, line 47: Even though I think I might have been the one who proposed this name, I'm not 100% sure it's the right one. In the context of `execute_in_process`, "in-process" mostly means the in-process executor. But `materialize` also uses the in-process executor. The main difference here is that we use the in-memory IO manager. So maybe we should call this `materialize_to_memory`?

python_modules/dagster/dagster/core/asset_defs/materialize.py
```diff
@@ -40,3 +42,43 @@ def materialize(
...
+    run_config: Any = None,
+    instance: Optional[DagsterInstance] = None,
+    resources: Optional[Mapping[str, object]] = None,
+) -> ExecuteInProcessResult:
+    """
+    Executes a single-threaded, in-process run which materializes provided assets.
+
+    By default, will materialize assets to memory, meaning results will not be
+    persisted. This behavior can be changed by overriding the default io
+    manager key "io_manager", or providing custom io manager keys to assets.
```
sryza, line 58: If someone provides an io_manager_key but doesn't provide any resources for it, will this throw raise an error? I think ideally we'd use the in-memory IO manager.

PR #6784, opened 2022-02-24
fix: use run status to determine success for an in-process result

rexledesma:
## Summary
Ran into this in internal - I expected the run from the execution result to have the correct run status, but it was returning `NOT_STARTED`. 

Just for consistency, put the correct status on the run.

## Test Plan
pytest

PR #8365, opened 2022-06-13
Simplify execute_in_process result

gibsondan:
Summary:
I think the only part of https://github.com/dagster-io/dagster/pull/6784/files that we needed was the part where the run is reloaded from the DB - the new test added in that PR still passes.

Test Plan: BK


rexledesma (approved):
nice

PR #8368, opened 2022-06-13
new typed events for step worker start/end and resource init/teardown

gibsondan:
Summary:
This gives us more consistent/structured logging of step lifecycle instead of needing to parse through engine events to figure out how long the various parts of the step lifecycle took.


gibsondan:
* master
  * **PR #8357** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8357" target="_blank">Graphite</a> 
    * **PR #8368** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8368" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8368?utm_source=stack-comment).

python_modules/dagster/dagster/cli/api.py
```diff
@@ -361,6 +363,22 @@ def _execute_step_command_body(
...
         )
 
+        log_manager = create_context_free_log_manager(instance, pipeline_run)
+
+        yield DagsterEvent.step_process_started(
+            log_manager,
+            pipeline_run.pipeline_name,
+            message="Started process for step"
+            + ("s" if not single_step_key else "")
+            + ((": " + ",".join(args.step_keys_to_execute)) if args.step_keys_to_execute else ""),
```
rexledesma, line 373: Could this just be a metadata entry?

python_modules/dagster/dagster/core/storage/event_log/sql_event_log.py
```diff
@@ -404,8 +404,8 @@ def get_step_stats_for_run(self, run_id, step_keys=None):
                         DagsterEventType.STEP_EXPECTATION_RESULT.value,
                         DagsterEventType.STEP_RESTARTED.value,
                         DagsterEventType.STEP_UP_FOR_RETRY.value,
-                        DagsterEventType.ENGINE_EVENT.value,
                     ]
+                    + list(marker_event.value for marker_event in MARKER_EVENTS)
```
rexledesma, line 408:
```suggestion
                    + [marker_event.value for marker_event in MARKER_EVENTS]
```

rexledesma (commented):
impl looks fine

prha:
Yeah, this is good....

we just need to be mindful of the fact that some of those timing calculations won't be backwards compatible.

python_modules/dagster/dagster/core/events/__init__.py
```diff
@@ -882,6 +922,46 @@ def pipeline_canceled(
...
+        message: str,
+        metadata_entries: List[MetadataEntry],
+    ) -> "DagsterEvent":
+        return DagsterEvent.from_step(
+            DagsterEventType.STEP_WORKER_STARTING,
+            step_context,
+            message=message,
+            event_specific_data=EngineEventData(
+                metadata_entries=metadata_entries, marker_start="step_process_start"
+            ),
```
gibsondan, line 937: one thing to double-check is that we're ok having this share EngineEventData - I think this would be the first case of two different event types sharing an event_specific_data class (but I would argue that we could benefit from a bit more consolidation in our event path in general...)

gibsondan:
@rexledesma believe all feedback here is addressed now

rexledesma (approved):

PR #8369, opened 2022-06-13
Run retries helm config

johannkm:

### DIFF
--- a/helm/dagster/schema/schema/charts/dagster/subschema/daemon.py
+++ b/helm/dagster/schema/schema/charts/dagster/subschema/daemon.py
@@ -81,6 +81,7 @@ class Daemon(BaseModel):
     startupProbe: kubernetes.StartupProbe
     annotations: kubernetes.Annotations
     runMonitoring: Dict[str, Any]
+    runRetries: Dict[str, Any]
 
     class Config:
         extra = Extra.forbid
--- a/helm/dagster/values.yaml
+++ b/helm/dagster/values.yaml
@@ -999,6 +999,10 @@ dagsterDaemon:
     # run launcher supports resuming runs, otherwise defaults to 0.
     maxResumeRunAttempts: ~
 
+  runRetries:
+    enabled: false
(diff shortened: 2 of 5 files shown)


johannkm:
* master
  * **PR #8369** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8369" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8369?utm_source=stack-comment).

helm/dagster/values.yaml
```diff
@@ -999,6 +999,10 @@ dagsterDaemon:
     # run launcher supports resuming runs, otherwise defaults to 0.
     maxResumeRunAttempts: ~
 
+  runRetries:
```
johannkm, line 1002: will link to docs once https://app.graphite.dev/github/pr/dagster-io/dagster/8367 lands

helm/dagster/schema/schema/charts/dagster/subschema/daemon.py
```diff
@@ -81,6 +81,7 @@ class Daemon(BaseModel):
     startupProbe: kubernetes.StartupProbe
     annotations: kubernetes.Annotations
     runMonitoring: Dict[str, Any]
+    runRetries: Dict[str, Any]
```
rexledesma, line 84: Could make this into a typed object - just need to assert it against the config type. See how we did it with the compute logs manager for example: https://sourcegraph.com/github.com/dagster-io/dagster/-/blob/helm/dagster/schema/schema_tests/test_instance.py?L508%3A93=

rexledesma (approved):
You can have finer types here, up to you

helm/dagster/schema/schema/charts/dagster/subschema/daemon.py
johannkm: Ah nice, will follow up

PR #8370, opened 2022-06-13
[docs] - Update config docs with assets

smackesey:
### Summary & Motivation

This updates the Config concept page to work in assets. It also ended up being a more general restructuring:

- Moved config schema to the top and added warning that using config without schema is not recommended
- Worked in resources to the main flow instead of being ghettoized in example

I think it still needs a bit of work but IMO is good enough for release.

Browsed docs.


smackesey:
* master
  * **PR #8370** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8370" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8370?utm_source=stack-comment).

docs/content/concepts/configuration/config-schema.mdx
```diff
@@ -15,69 +15,75 @@ Run configuration allows providing parameters to jobs at the time they're execut
...
 
 A common use of configuration is for a [schedule](/concepts/partitions-schedules-sensors/schedules) or [sensor](/concepts/partitions-schedules-sensors/schedules) to provide configuration to the job run it is launching. For example, a daily schedule might provide the day it's running on to one of the ops as a config value, and that op might use that config value to decide what day's data to read.
 
 Dagster includes a system for gradually-typed configuration schemas. These make it easy to catch configuration errors before job execution, as well as to learn what configuration is required to execute a job.
 
-## Using Configuration Inside an Op
+## Defining and Accessing Configuration for an Op, Asset, or Resource
```
erinkcochran87, line 26:
```suggestion
## Defining and accessing configuration for an op, asset, or resource
```

Suggesting we make this use sentence case. Heading case has been inconsistent until now, but we're going with H1 as camel and all others as sentence.

docs/content/concepts/configuration/config-schema.mdx
```diff
@@ -15,69 +15,75 @@ Run configuration allows providing parameters to jobs at the time they're execut
...
 
-This example shows how to write an op whose behavior is based on values that are passed in via configuration:
+Configurable parameters accepted by an op, asset, or resource are specified by providing a `config_schema` to the corresponding decorator. The structure of a `config_schema` is very flexible and [fully documented in the API Reference](/\_apidocs/config). However, most of the time you will want to provide a Python dictionary, with keys the names of parameters and values the types of those parameters.
```
erinkcochran87, line 28: Nit - can we lose the qualifier?

> The structure of a `config_schema` is ~very~ flexible

erinkcochran87, line 28: This read like some words were left out (see below) - or was this intentional?

> However, most of the time you will want to provide a Python dictionary, with keys **as** the names of parameters and values **as** the types of those parameters.

docs/content/concepts/configuration/config-schema.mdx
```diff
@@ -15,69 +15,75 @@ Run configuration allows providing parameters to jobs at the time they're execut
...
-## Providing Run Configuration
 
-How you specify config values depends on how you're running your job:
+@resource(config_schema={"url": str})
+def resource_using_config(context):
+    return MyDatabaseConnection(context.resource_config["url"])
+```
 
-### Python API
+<Note>
```
erinkcochran87, line 51: Markdown isn't usable inside these callouts, unfortunately. Could you change the backticks here to `<code></code>` so it displays correctly?

docs/content/concepts/configuration/config-schema.mdx
```diff
@@ -15,69 +15,75 @@ Run configuration allows providing parameters to jobs at the time they're execut
...
+  It is technically possible to access `context.op_config` inside ops (not
+  assets) without defining a `config_schema`. However, this is not recommended
+  and will likely be deprecated in the future.
+</Note>
 
-When executing a job with <PyObject object="JobDefinition" method="execute_in_process"/>, you can specify the config values through `run_config` argument:
+You can also build config into jobs, as described in [this section of the Jobs concept page](/concepts/ops-jobs-graphs/jobs-graphs#advanced-job-configuration).
```
erinkcochran87, line 57:
```suggestion
You can also build config into jobs, as described in [the Jobs documentation](/concepts/ops-jobs-graphs/jobs-graphs#advanced-job-configuration).
```

Sorry, nit here too. I'm trying to make internal references a bit more consistent and concise.

docs/content/concepts/configuration/config-schema.mdx
```diff
@@ -15,69 +15,75 @@ Run configuration allows providing parameters to jobs at the time they're execut
...
 
-```python file=/concepts/configuration/execute_with_config.py startafter=start_execute_with_config endbefore=end_execute_with_config
-result = config_example.execute_in_process(
-        run_config={"ops": {"uses_config": {"config": {"iterations": 1}}}}
-    )
-```
+## Specifying Runtime Configuration
```
erinkcochran87, line 59:
```suggestion
## Specifying runtime configuration
```

docs/content/concepts/configuration/config-schema.mdx
```diff
@@ -15,69 +15,75 @@ Run configuration allows providing parameters to jobs at the time they're execut
...
 
-You can also build config into jobs, as described in [this section of the Jobs concept page](/concepts/ops-jobs-graphs/jobs-graphs#advanced-job-configuration).
+If we want to execute `op_using_config` or materialize `asset_using_config`, we'll need to provide values for the parameters specified in `config_schema`. How we provide these values depends on the interface we are using.
```
erinkcochran87, line 61:
```suggestion
If we want to execute `op_using_config` or materialize `asset_using_config`, we'll need to provide values for the parameters specified in `config_schema`. How we provide these values depends on the interface we are using:

- [Python](#python)
- [Dagit](#dagit)
- [Command line](#command-line)
```

It's helpful to have inline subsection headings in top-level sections. It keeps the reader focused on the copy and not having to look off to the side at the on-page nav to know what's in the section.

docs/content/concepts/configuration/config-schema.mdx
```diff
@@ -15,69 +15,75 @@ Run configuration allows providing parameters to jobs at the time they're execut
...
 
-### Dagster CLI
+### Python
 
-When executing a job from the command line, the easiest way to provide config is to put it into a YAML file, like:
+From the Python API, we can use the `run_config` argument for <PyObject object="JobDefinition" method="execute_in_process"/> or <PyObject module="materialize"/>. This takes a dictionary where configuration values for ops/assets are specified under `ops.<op_or_asset_name>.config` (for resources under `resources.<resource_name>.config`):
```
erinkcochran87, line 65: The `materialize` PyObject isn't showing up correctly in the preview.

erinkcochran87 (commented):
This is great - thanks Sean! Most of my comments are around grammar and formatting, but overall this looks very shippable. We should get someone to do a pass for technical accuracy before releasing.

docs/content/concepts/configuration/config-schema.mdx
```diff
@@ -15,69 +15,75 @@ Run configuration allows providing parameters to jobs at the time they're execut
 
 ## Overview
 
-It's often useful to configure jobs at run time. For example, you might want someone to manually operate a deployed job and choose what dataset it operates on when they run it. In general, you should use Dagster's config system when you want the person or software that is executing a job to be able to make choices about what the job does, without needing to modify the job definition.
+It's often useful to configure Dagster jobs at runtime. For example, you might want to choose what dataset an op runs against, or provide a connection URL for a database resource. In general, you should use Dagster's config system when you want to modulate a computation at runtime without needing to modify the definition.
```
sryza, line 18: I think it's worth mentioning software-defined assets here, because you can supply config when launching a materialization run even if you haven't defined any jobs.

docs/content/concepts/configuration/config-schema.mdx
sryza: Asset -> Software-defined asset

docs/content/concepts/configuration/config-schema.mdx
```diff
@@ -15,69 +15,75 @@ Run configuration allows providing parameters to jobs at the time they're execut
...
 
-The objects that compose a job - ops and resources - are each individually configurable. When executing a job, you can supply "run configuration" that specifies the configuration for each of the objects in the job. When you execute a job with Dagster's Python API, you supply run configuration as a Python dictionary. When you execute a job from Dagit or the CLI, you can provide config in a YAML document.
+Various Dagster entities (ops, assets, resources) can have an attached _configuration schema_ that specifies the configurable parameters for that entity. When launching a job that executes (ops), materializes (assets), or instantiates (resources) a configurable entity, you can provide values for the parameters specified in the schema. Dagster validates this _run configuration_ against the schema and proceeds only if validation is successful. Within the function that defines the entity, you can access the passed-in configuration off of the `context`.
```
sryza, line 20: Not leading with config schema was a conscious choice from a prior revision - we were concerned that asking readers to immediately understand the notion of config schema would make the system less approachable.  I don't think that we should try to bury config schema, but I think the notion that ops, software-defined assets, and resources can be configured should be introduced before introducing schema.

docs/content/concepts/configuration/config-schema.mdx
```diff
@@ -15,69 +15,75 @@ Run configuration allows providing parameters to jobs at the time they're execut
...
+@resource(config_schema={"url": str})
+def resource_using_config(context):
+    return MyDatabaseConnection(context.resource_config["url"])
+```
 
-### Python API
+<Note>
+  It is technically possible to access `context.op_config` inside ops (not
+  assets) without defining a `config_schema`. However, this is not recommended
+  and will likely be deprecated in the future.
```
sryza, line 54: We should avoid making claims about what changes we'll make in the future until we have definite plans. Better to just say that using config schema is recommended.

docs/content/concepts/configuration/config-schema.mdx
sryza, line 61: We prefer second-person over first-person.

sryza (commented):
Looking good! Left a few comments.

docs/content/concepts/configuration/config-schema.mdx
smackesey: The impetus to put schema first came from the fact that you cannot use config in assets without schema (this is a consequence of there already being an implicit schema with the `assets` key for assets). If we want to present assets and ops together (which IMO we should for simplicity's sake), the old ordering becomes very awkward since the first section (accessing config inside compute fn _without_ schema) does not work for assets.

FWIW however, I also find the new ordering more clear generally, especially if we are discouraging config without schema. IMO 99% of the time, people are going to think: "Configuration? Cool, first, where do I define the accepted parameters? Second, how do I access them?" Not: "First, how do I access the parameters? Second, how do I specify them?" It's like you'd typically write a function signature before writing the body.

>we were concerned that asking readers to immediately understand the notion of config schema would make the system less approachable.

I agree that attempting a full presentation of config schema would be undesirable, but the common case (which is used in the example) is simple and virtually self-explanatory: key-value pairs with keys parameter names and values simple types. I mentioned that I have some other improvements for the page in mind, one of these is a greater-depth explanation of schema further down.

>I think the notion that ops, software-defined assets, and resources can be configured should be introduced before introducing schema.

Is this not already introduced beforehand the Overview section?

docs/content/concepts/configuration/config-schema.mdx
smackesey: Yeah I was uncertain about this. I wanted to avoid having to list assets and ops side-by-side everytime I say anything general about config because in reality, they are using the same system. So I left it at "jobs" since asset materializations technically run through a job. But if this is hidden from the user, I see why it shouldn't be the case.

What do you think of saying "configure Dagster runs at launch"? Since runs are the public-facing common connection between jobs and asset materializations.

docs/content/concepts/configuration/config-schema.mdx
smackesey: Hmm actually now I see that you are commenting on the Overview, I thought your comment was about the flow of the lower sections. It's an easy fix in the Overview.

docs/content/concepts/configuration/config-schema.mdx
```diff
@@ -15,69 +15,77 @@ Run configuration allows providing parameters to jobs at the time they're execut
...
 
-The objects that compose a job - ops and resources - are each individually configurable. When executing a job, you can supply "run configuration" that specifies the configuration for each of the objects in the job. When you execute a job with Dagster's Python API, you supply run configuration as a Python dictionary. When you execute a job from Dagit or the CLI, you can provide config in a YAML document.
+Various Dagster entities (ops, assets, resources) can be individually configured. When launching a job that executes (ops), materializes (assets), or instantiates (resources) a configurable entity, you can provide _run configuration_ for each entity. Within the function that defines the entity, you can access the passed-in configuration off of the `context`. Typically, the provided run configuration values correspond to a _configuration schema_ attached to the op/asset/resource definition. Dagster validates the run configuration against the schema and proceeds only if validation is successful.
```
sryza, line 20: * I found the executes / materializes / instantiates with the parentheses a little hard to parse. Maybe just say "uses"?
* "launching a job" -> "launching a run".

docs/content/concepts/configuration/config-schema.mdx
```diff
@@ -15,69 +15,77 @@ Run configuration allows providing parameters to jobs at the time they're execut
...
 
 A common use of configuration is for a [schedule](/concepts/partitions-schedules-sensors/schedules) or [sensor](/concepts/partitions-schedules-sensors/schedules) to provide configuration to the job run it is launching. For example, a daily schedule might provide the day it's running on to one of the ops as a config value, and that op might use that config value to decide what day's data to read.
 
-Dagster includes a system for gradually-typed configuration schemas. These make it easy to catch configuration errors before job execution, as well as to learn what configuration is required to execute a job.
+## Defining and accessing configuration for an op, asset, or resource
 
-## Using Configuration Inside an Op
+Configurable parameters accepted by an op, asset, or resource are specified by providing a `config_schema` to the corresponding decorator. The structure of a `config_schema` is flexible and [fully documented in the API Reference](/\_apidocs/config). However, most of the time you will want to provide a Python dictionary, with keys the names of parameters and values the types of those parameters.
```
sryza, line 26: This is pure style feedback, so feel free to ignore, but I think the first sentence would be more readable in the active voice. E.g. "You can which configuration parameters your op, asset, or resource accepts by providing a `config_schema` to the corresponding decorator".

docs/content/concepts/assets/asset-materializations.mdx
```diff
@@ -40,32 +40,13 @@ One way of recording materialization events is to log <PyObject module="dagster"
...
-def my_materialization_op(context):
-    df = read_df()
-    remote_storage_path = persist_to_storage(df)
-    context.log_event(
-        AssetMaterialization(
-            asset_key="my_dataset", description="Persisted result to storage"
-        )
-    )
-    return remote_storage_path
+ENOENT: no such file or directory, open '/Users/smackesey/stm/code/elementl/oss/examples/docs_snippets/docs_snippets/concepts/assets/materialization_ops.py'
```
sryza, line 49: Not sure what's going on here. Maybe a race condition with my change? Maybe related to the lint error you mentioned?

sryza (approved):
Plz incorporate the changes at least from my first comment. Otherwise looks good to go!

PR #8372, opened 2022-06-13
[assets] assorted bugfixes

OwenKephart:
### Summary & Motivation

This is the branch that I stacked my new demo code off of. I noticed several small issues while doing this, so I fixed them all in this branch. They are:

1. The error detection logic I used to determine if a selection of asset keys was invalid was (still) faulty. The logic said "if an asset is selected that is not provided by any AssetsDefinitions or SourceAssets, then it's an error". This is wrong because if the asset key is used as an input with type Nothing, then this is still ok.
2. key_prefix and namespace were set to function identically, but we actually don't want the key prefix to apply to the inputs of the decorated function (because this is not how the key prefix argument works in other scenarios)
3. dbt: added database, schema, and table to metadata for each asset. this makes it easier to write io managers
4. dbt: added the source schema name to the key asset key for dbt sources (previously, it was just source table name). this is necessary because source table names may overlap.
5. dbt: added a source_key_prefix argument to load_assets_from_dbt*. This is necessary because the source keys may need a different prefix that the model keys.

units

### DIFF
--- a/examples/modern_data_stack_assets/modern_data_stack_assets/assets.py
+++ b/examples/modern_data_stack_assets/modern_data_stack_assets/assets.py
@@ -13,7 +13,9 @@ from .constants import *  # pylint: disable=wildcard-import,unused-wildcard-impo
 from .pandas_io_manager import pandas_io_manager
 
 airbyte_assets = build_airbyte_assets(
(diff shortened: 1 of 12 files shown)


sryza (approved):
Discussed offline that we'll take out the db/schema-on-dbt-model changes. Otherwise, this looks great.

PR #8377, opened 2022-06-14
[dagit] Update the linking between asset groups + jobs

bengotow:
### Summary & Motivation

This PR moves the "what if this asset isn't in the displayed graph?" handling out of `AssetGraphExplorer` into a prop, so that the group and job graphs can link to appropriate places.

On an asset group page, clicking an asset node outside the group takes you to that assets group page or to the definition if it is a source asset.

On an asset job page, clicking an asset node outside the job takes you to that asset's job page or to the definition if it is a source asset.

This addresses feedback that clicking around between groups would land you on definition pages instead of on the other groups.

To make things a bit simpler I also created a helper for generating the asset details page path. I wanted the params to be type checked, and wanted to make sure that the path components are always URL-encoded, in case someone is crazy enough to put a ? or #, etc.  in an asset key path. I realized I'd been doing it kind of inconsistently.

- Clicked "foreign" links on asset jobs and asset groups, in cases where resolution succeeds and where resolution fails (source asset).

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
@@ -3,7 +3,6 @@ import pickBy from 'lodash/pickBy';
 import uniq from 'lodash/uniq';
 import without from 'lodash/without';
 import React from 'react';
-import {useHistory} from 'react-router-dom';
 import styled from 'styled-components/macro';
 
 import {GraphQueryItem} from '../app/GraphQueryImpl';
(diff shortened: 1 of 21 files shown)


js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
```diff
@@ -171,33 +171,10 @@ export const AssetGraphExplorerWithData: React.FC<
...
-
-      // This asset is in different job (and we're in the job explorer),
-      // go to the other job.
-      if (clicked.jobName !== explorerPath.pipelineName) {
-        onChangeExplorerPath(
-          {...explorerPath, opNames: [token], opsQuery: '', pipelineName: clicked.jobName!},
-          'replace',
-        );
-        return;
+        return onNavigateToForeignNode(await findAssetLocation(assetKey));
```
hellendag, line 177: Does this need any kind of loading state while the asset location is queried?

hellendag (approved):

PR #8380, opened 2022-06-14
RFC: Screw around with check

johannkm:
I wonder how bad an idea this is

check becomes O(how many locals are in your frame)

### DIFF
--- a/python_modules/dagster/dagster_tests/general_tests/check_tests/test_check.py
+++ b/python_modules/dagster/dagster_tests/general_tests/check_tests/test_check.py
@@ -52,6 +52,14 @@ def test_bool_param():
         check.bool_param("val", "b")
 
 
+def test_bool_param_no_param_name():
+    my_value = 42
+    with pytest.raises(ParameterCheckError) as e:
+        check.bool_param(my_value)
+
+    assert str(e.value) == "Param \"my_value\" is not a bool. Got 42 which is type <class 'int'>."
+
+
 def test_opt_bool_param():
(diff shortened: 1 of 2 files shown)


johannkm:
* master
  * **PR #8380** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8380" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8380?utm_source=stack-comment).

python_modules/dagster/dagster_tests/general_tests/check_tests/test_check.py
```diff
@@ -52,6 +52,14 @@ def test_bool_param():
         check.bool_param("val", "b")
 
 
+def test_bool_param_no_param_name():
+    my_value = 42
+    with pytest.raises(ParameterCheckError) as e:
+        check.bool_param(my_value)
+
+    assert str(e.value) == "Param \"my_value\" is not a bool. Got 42 which is type <class 'int'>."
```
prha, line 60: what's the pathological case here?  something like this?

```
def foo(a, b):
    check.opt_bool_param(a)
    check.bool_param(b)
    
foo(None, None)
```

Seems like this would be too easy to get wrong.

prha (changes requested):

johannkm (commented):
Yeah that's a good callout. I wonder if this would be possible using https://docs.python.org/3/library/traceback.html instead. Putting on ice for now

PR #8384, opened 2022-06-14
Patch op definition resolution for assets in GQL

smackesey:
### Summary & Motivation

This is to work around a current incompatibility between graph-backed assets and the op definition resolver for assets in GQL . Currently `ExternalAssetNode` stores the op handle instead of definition name for `op_name`. If the same op is invoked twice, the op handle is `<op_def_name>_n` instead of the plain op def name. This breaks GQL resolution which attempts to find an op definition with this name.

The workaround is a new `node_definition_name` property to `ExternalAssetNode` that contains the node definition name, which is used in GQL resolution.

- BK and playing with asset config in dagit (examples/bollinger)


smackesey:
* master
  * **PR #8384** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8384" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8384?utm_source=stack-comment).

OwenKephart (approved):

PR #8385, opened 2022-06-14
Fix for blocks_fn option

solarisa21:
If user provides blocks, do not append default mrkdwn block - allow users to define block via their blocks_fn

### DIFF
--- a/python_modules/libraries/dagster-slack/dagster_slack/sensors.py
+++ b/python_modules/libraries/dagster-slack/dagster_slack/sensors.py
@@ -33,7 +33,7 @@ def _build_slack_blocks_and_text(
     main_body_text = text_fn(context)
 
     if blocks_fn:
-        blocks.extend(blocks_fn(context))
+        blocks_fn(context)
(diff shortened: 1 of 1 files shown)


yuhan:
This makes sense to me - kicking off tests. 
@solarisa21 Thanks for the contribution!

solarisa21:
> Hi @solarisa21. Seems like there is a minor mypy error, would you mind fixing it before we merge this? https://buildkite.com/dagster/dagster/builds/31459#018168ca-93b0-4dfd-bbec-5372f7f8660f/302-640
> 
> I think you'll also need to pull from master to resolve some of the existing buildkite errors.
I'm getting a "Page not found" error from your buildkite link.  Can you guys can help out?

yuhan:
I believe this was landed in https://github.com/dagster-io/dagster/pull/8448 and released in [0.15.5](https://github.com/dagster-io/dagster/releases/tag/0.15.5) last week. Thanks for the contribution!

PR #8396, opened 2022-06-14
docs: default to master version when vercel env is not prod

yuhan:
### Summary & Motivation
make vercel previews default to the master branch

preview defaults to the master branch

### DIFF
--- a/docs/next/util/useVersion.ts
+++ b/docs/next/util/useVersion.ts
@@ -3,11 +3,17 @@ import { useEffect, useState } from "react";
 import ALL_VERSIONS from "../.versioned_content/_versions.json";
 import { useRouter } from "next/router";
 
-const defaultVersion =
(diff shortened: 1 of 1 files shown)


rexledesma (approved):
image

rexledesma:
We should make sure that release builds (e.g. 0.15.1) don't default to the main branch, but I don't think we can simulate that until we actually release

yuhan:
I don't think this affects the release builds because how doc release works is the release pipeline merges the automation: `versioned docs ...` commit ([this](https://github.com/dagster-io/dagster/commit/41ea0be6e9f1a3d4f37f60fd5a3f775281442116)) back to the main branch and vercel builds it as a "production" build, i.e. there will be a preview for the commit but on the release branch and will be another production build for the same commit that gets merged into the main branch:
image

PR #8399, opened 2022-06-14
purge ticks older than 30 days

prha:

### DIFF
--- a/python_modules/dagster/dagster/core/instance/__init__.py
+++ b/python_modules/dagster/dagster/core/instance/__init__.py
@@ -1970,8 +1970,8 @@ class DagsterInstance:
     def update_tick(self, tick):
         return self._schedule_storage.update_tick(tick)
 
-    def purge_ticks(self, origin_id, selector_id, tick_status, before):
-        self._schedule_storage.purge_ticks(origin_id, selector_id, tick_status, before)
+    def purge_ticks(self, origin_id, selector_id, before, tick_statuses=None):
+        self._schedule_storage.purge_ticks(origin_id, selector_id, before, tick_statuses)
 
     def wipe_all_schedules(self):
         if self._scheduler:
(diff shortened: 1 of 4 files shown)


prha:
* master
  * **PR #8399** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8399" target="_blank">Graphite</a>  👈
    * **PR #8400** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8400" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8399?utm_source=stack-comment).

gibsondan (changes requested):
I think this is very reasonable but also think that we should allow users to tune if it they want? (even if via environment variable or something)

If we imagine allowing users to set a retention policy on runs in the future too we could stick that retention information in instance settings somewhere (or we could just use an env var and default it to 30)

prha:
Yeah, I like the idea of putting retention policy on the instance... What's the equivalent place that users can configure this in cloud?

python_modules/dagster/dagster/core/instance/config.py
```diff
@@ -177,4 +253,5 @@ def dagster_instance_config_schema():
         "code_servers": Field(
             {"local_startup_timeout": Field(int, is_required=False)}, is_required=False
         ),
+        "tick_retention": tick_retention_config_schema(),
```
gibsondan, line 256: one thing that we are going to want to figure out at some point - every time we add one of these fields and somebody uses it, they have to upgrade their user code as well, or it will error out when it tries to load the instance when launching a run and this new unexpected key is there (even though a setting like this is totally irrelevant in user code)

gibsondan, line 256: do you think something like this could make sense to colocate with event log retention or run retention once that is available? Maybe this should just be 'retention' for future retention-related settings?

gibsondan (approved):
would suggest changing the "tick_retention" key to "retention" for the reasons inline, otherwise i think this looks good

PR #8401, opened 2022-06-14
GrapheneAsset -> GrapheneMaterializedKey

clairelin135:
`GrapheneAsset` was the old graphql asset class used for non-SDA assets (assets without definitions). This PR renames `GrapheneAsset` to `GrapheneMaterializedKey` to reduce confusion.

This PR also updates `GrapheneMaterializedKey.definition` to `GrapheneMaterializedKey.assetNode`. Previously, `definition` existed to differentiate between defined asset nodes and the definition-less `Asset` object, and now we can update this naming to be more consistent.


js_modules/dagit/packages/core/src/assets/AssetsCatalogTable.tsx
```diff
@@ -74,11 +74,14 @@ function useAllAssets(
         assets: assetNodes?.map(definitionToAssetTableFragment),
       };
     } else {
-      const assetsOrError = assetsQuery.data?.assetsOrError;
+      const assetsOrError = assetsQuery.data?.materializedKeysOrError;
```
bengotow, line 77: We can revisit a few of these and push the change deeper into Dagit 🙌 

I think MaterializedKey is a big improvement on Asset and all the confusion around Asset vs. AssetNode.  It does make me wonder about these two:

`MaterializedKey => assetMaterializations`
`MaterializedKey => definition`

Could assetMaterializations just be `materializations`? (Since this isn't an AssetMaterializationkey, just a MaterializationKey, and there aren't other kinds of materializations)

Definition could stay the same, but I could also see renaming it `assetNode`? I think we just called it definition because we needed to clarify the relationship between aset and assetNode somehow. 

Will defer to the other reviewers but I'm super excited to get this naming improved!

clairelin135: Definitely agree with you about renaming `MaterializedKey.definition` to `MaterializedKey.assetNode`. I've gone ahead and added a commit to change this.

I'm a little more hesitant about renaming `assetMaterializations` to just `materializations`, purely to maintain consistency between `assetMaterialization` and the python `AssetMaterialization` object. We'd also have to rename `assetObservations` to `observations`. In any case, I'd prefer to do this in a separate PR to break out the changes a little bit more.

sryza:
I don't have any strong opinions on what our internal APIs should look like, but from the user perspective, I'd rather distinguish between "Software-defined assets" and "Assets" than between "Assets" and "MaterializedKeys".  With that lens in mind, would it possibly make sense to call this "AssetEventHistory" or something?

sryza:
Also worth thinking about: AssetObservations.

clairelin135:
@sryza That makes sense. If we want to make the differentiation extra clear, I think we can rename `AssetNode` to `SdaAsset`
I feel like "AssetEventHistory" still feels ambiguous what this object is. I'd rather name it `UndefinedAsset` or `NoDefinitionAsset` instead.

PR #7217, opened 2022-03-28
fix: prevent runs filter from accepting an empty list of run ids

rexledesma:
## Summary
Currently, applying a run filter with `run_ids` as `None` or `[]` has the same result of fetching all the runs. This is because both values are falsey.

We could change the logic to treat `[]` as a short-circuit and skip fetching any runs, but then the API consumer would need to be wary that passing in `[]` has this particular behavior.

Instead, if `[]` is passed in, throw an error. The API consumer should explicitly guard for this case in their business logic, rather than relying on the underlying API to skip the fetching of any runs.

## Test Plan
pytest

run dagit locally, fill out `id:` field with random values

https://user-images.githubusercontent.com/16431325/160492995-11d5aaff-48f9-4594-8302-07a309d2cbea.mov

PR #8415, opened 2022-06-15
Fix deserialization of RunsFilter

johannkm:
A side effect of https://github.com/dagster-io/dagster/pull/7217 is that you can't serialize and deserialize runsfilters. I'm not sure where else we do that but I ran in to it for bulk action stuff

Is there a better way to hack this?

### DIFF
--- a/python_modules/dagster/dagster/core/storage/pipeline_run.py
+++ b/python_modules/dagster/dagster/core/storage/pipeline_run.py
@@ -1,6 +1,7 @@
 import warnings
 from datetime import datetime
 from enum import Enum
+from inspect import Parameter
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/core/storage/pipeline_run.py
```diff
@@ -529,6 +530,23 @@ def value_to_storage_dict(
         # job_name as pipeline_name
         return replace_storage_keys(storage, {"job_name": "pipeline_name"})
 
+    @classmethod
+    def value_from_storage_dict(
```
prha, line 534: where do we actually store `RunsFilter`?

prha: I guess we could just have it deserialized in cloud from a client running a previous version of dagster...

might be worth adding a comment that this is specifically for that case, and note that we could drop this if we can guarantee that all clients are running version `0.14.7`

PR #8432, opened 2022-06-16
Update runtime type checks in resource definition invocation

gibsondan:
Summary:
Right now the callsite here allows a bound resource context but then fails when it gets to the interior that only allows an unbound one. Update the runtime type checks and throw some mypy checks in there too.

Test Plan: Run a pipeline that was relying on direct resource invocation within another resource, it now fails earlier


dpeng817 (approved):
Thanks!

PR #8435, opened 2022-06-16
clean up some language in the content docs

prha:
### Summary & Motivation
Just noticed some language that felt awk when reading through the docs.

Inspection


sryza (approved):
Thanks for fixing

PR #8443, opened 2022-06-16
[dagit] Send YAML insead of parsed JSON

hellendag:
### Summary & Motivation

Depends on https://github.com/dagster-io/dagster/pull/8436.

When submitting run config for validation or launch, just send yaml instead of parsing it into an object first. This will allow us to handle yaml syntax properly, instead of incorrectly converting certain values (e.g. `.NAN`, strings that are numeric values).

View Launchpad.

- Add some invalid yaml, verify correct linting behavior.
- Inspect validation query variables, verify that we're sending a yaml string instead of a JS object.
- Add a `.NAN`, verify that it is sent and parsed correctly -- if used in a place where a string should be, there is an error saying that it shouldn't be a float value
- Launch a run, verify that yaml string is sent and handled properly.

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
+++ b/js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
@@ -13,7 +13,6 @@ import {
 import reject from 'lodash/reject';
 import React from 'react';
 import {useHistory} from 'react-router-dom';
-import * as yaml from 'yaml';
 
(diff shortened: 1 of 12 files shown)


hellendag:
* master
  * **PR #8443** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8443" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8443?utm_source=stack-comment).

gibsondan (approved):

PR #8444, opened 2022-06-16
[ui] Add ability to supply TokenizingField with raw tokens

benpankow:
## Summary

Adds the option to supply a `tokens` param to the `TokenizingField` input in place of or in addition to the `suggestionProviders` param.

This allows us to create multi-select inputs with autocompletes, either for raw tokens or key-value pairs.

Also adds the ability to add a custom rendering function to suggestions.

## Test Plan

Storybook

### DIFF
--- a/js_modules/dagit/packages/ui/src/components/TokenizingField.stories.tsx
+++ b/js_modules/dagit/packages/ui/src/components/TokenizingField.stories.tsx
@@ -1,0 +1,100 @@
+import {Meta} from '@storybook/react/types-6-0';
+import * as React from 'react';
+import styled from 'styled-components/macro';
+
+import {Colors} from './Colors';
+import {Group} from './Group';
+import {TokenizingField, TokenizingFieldValue} from './TokenizingField';
+
+// eslint-disable-next-line import/no-default-export
+export default {
+  title: 'TokenizingField',
+  component: TokenizingField,
+} as Meta;
+
+export const Default = () => {
+  const [value, setValue] = React.useState<TokenizingFieldValue[]>([]);
+
+  const suggestions = [
+    {
+      token: 'pipeline',
+      values: () => ['airline_demo_ingest', 'airline_demo_warehouse', 'composition'],
+    },
+    {
+      token: 'status',
+      values: () => ['QUEUED', 'NOT_STARTED', 'STARTED', 'SUCCESS', 'FAILURE', 'MANAGED'],
+    },
+  ];
+
+  return (
+    <TokenizingField
+      values={value}
+      onChange={(values) => setValue(values)}
+      suggestionProviders={suggestions}
+    />
+  );
+};
+
+export const TokenProvider = () => {
+  const [value, setValue] = React.useState<TokenizingFieldValue[]>([]);
+
+  return (
+    <TokenizingField
+      values={value}
+      onChange={(values) => setValue(values)}
+      tokens={['ben@elementl.com', 'dish@elementl.com', 'marco@elementl.com']}
+    />
+  );
+};
+
+export const TokenAndSuggestionProviders = () => {
+  const [value, setValue] = React.useState<TokenizingFieldValue[]>([]);
+
+  const suggestions = [
+    {
+      token: 'group',
+      values: () => ['core@elementl.com', 'cloud@elementl.com'],
+    },
+  ];
+
+  return (
+    <TokenizingField
+      values={value}
+      onChange={(values) => setValue(values)}
+      tokens={['ben@elementl.com', 'dish@elementl.com', 'marco@elementl.com']}
+      suggestionProviders={suggestions}
+    />
+  );
+};
+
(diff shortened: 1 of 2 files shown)


benpankow:
* master
  * **PR #8444** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8444" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8444?utm_source=stack-comment).

js_modules/dagit/packages/ui/src/components/TokenizingField.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+      values: () => ['core@elementl.com', 'cloud@elementl.com'],
+    },
+  ];
+
+  return (
+    <TokenizingField
+      values={value}
+      onChange={(values) => setValue(values)}
+      tokens={['ben@elementl.com', 'dish@elementl.com', 'marco@elementl.com']}
+      suggestionProviders={suggestions}
```
bengotow, line 65: I think this is nice but it's a little unclear what the behavior is when tokens / providers are passed together -- I wonder if it'd be a bit clearer if TokenizingField exposed a convenience method for building a provider in a really shorthand way for the basic tokens case? (I don't feel strongly about it though!)

Maybe this could be something like:

```

    <TokenizingField
      values={value}
      onChange={(values) => setValue(values)}
      suggestionProviders={[providerForValues(['ben@elementl.com', 'dish@elementl.com', 'marco@elementl.com'])}
/>
```

I'm a little wary of building more complexity into the TokenizingField itself because we've had a couple discussions about adding other features to it like "double-click edit an item" etc., and it could end up being a beast if it supports a lot of use cases. 

 --

Maybe we ship this as implemented for now, but if we end up making TokenizingField do more powerful stuff for the run filter bar, etc., we consider splitting this "basic non-key-value text tokens" use case out into a separate component? Curious to get your thoughts @hellendag

js_modules/dagit/packages/ui/src/components/TokenizingField.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+    Blue: Colors.Blue500,
+    Yellow: Colors.Yellow500,
+  };
+
+  return (
+    <TokenizingField
+      values={value}
+      onChange={(values) => setValue(values)}
+      tokens={Object.keys(colors)}
+      suggestionRenderer={(suggestion) => (
```
bengotow, line 84: This is a great API addition and I like that it mirrors the `itemRenderer` naming convention that blueprint uses 👍

js_modules/dagit/packages/ui/src/components/TokenizingField.tsx
```diff
@@ -96,12 +102,18 @@ export const stringFromValue = (value: TokenizingFieldValue[]) =>
...
+ *  sourced from the `suggestionProviders` param. In the latter case, the
+ *  key is one of a known set of "suggestion provider tokens".
+ *
+ *  Provide one or more SuggestionProviders or a list of tokens
+ *  to build the tree of autocompletions.
+ *
+ *  The input also allows for freeform typing (`value` items with no token value) */
 export const TokenizingField: React.FC<TokenizingFieldProps> = ({
-  suggestionProviders,
+  suggestionProviders: rawSuggesetionProviders,
```
bengotow, line 116: Ahh tiny sp here

bengotow (approved):
This looks great! Would be nice to get @hellendag's thoughts too because we were talking about adding features to this input soonish

js_modules/dagit/packages/ui/src/components/TokenizingField.tsx
```diff
@@ -36,14 +36,21 @@ interface TokenizingFieldProps {
   maxValues?: number;
   onChange: (values: TokenizingFieldValue[]) => void;
   onChangeBeforeCommit?: boolean;
+  addOnBlur?: boolean;
```
salazarm, line 39: also `addOnPaste` would  be nice

js_modules/dagit/packages/ui/src/components/TokenizingField.tsx
```diff
@@ -384,8 +427,16 @@ export const TokenizingField: React.FC<TokenizingFieldProps> = ({
             setOpen(true);
             onFocus && onFocus();
           },
-          onBlur: () => setOpen(false),
+          onBlur: () => {
```
salazarm, line 430: There should already be an `addOnBlur` prop on `StledTagInput`

js_modules/dagit/packages/ui/src/components/TokenizingField.tsx
benpankow: See below - I can't seem to get it to work as expected, I think because we override some of the input/change methods.

hellendag (approved):

PR #8445, opened 2022-06-16
[dagster-io/ui] Add canShow to Tooltip

hellendag:
### Summary & Motivation

We have a common issue in Dagit where we sometimes want to show a disabled control (e.g. a `Button`) with some information in a tooltip, for instance to indicate why the button is disabled.

Toggling tooltip-ness on an element isn't exactly supported behavior on the Blueprint tooltip component, however. If you leave the tooltip content empty, a warning appears. If you force `isOpen` to be undefined or false, it's kind of a weird hack.

This PR aims to resolve this with `MaybeTooltip`, which uses a `hasTooltip` prop to determine whether to render a tooltip wrapper on a given child component. If false, just render the child. If true, render the tooltip with the supplied props.

In this way, we can toggle tooltip-ness on `hasTooltip`, and not need to worry about sidestepping the `Tooltip` component API, rendering unwanted wrapper nodes, or triggering React warnings.

Storybook example, Jest.

### DIFF
--- a/js_modules/dagit/packages/ui/src/components/Tooltip.stories.tsx
+++ b/js_modules/dagit/packages/ui/src/components/Tooltip.stories.tsx
@@ -1,6 +1,9 @@
 import {Meta} from '@storybook/react/types-6-0';
 import * as React from 'react';
 
+import {Box} from './Box';
+import {Button} from './Button';
+import {Checkbox} from './Checkbox';
 import {Colors} from './Colors';
 import {CustomTooltipProvider} from './CustomTooltipProvider';
 import {Group} from './Group';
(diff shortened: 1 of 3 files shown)


hellendag:
* master
  * **PR #8445** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8445" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8445?utm_source=stack-comment).

bengotow (approved):
This looks great! I do wonder if we could build this directly into `Tooltip`? It seems being able to pass no tooltip content and get no tooltip would be nice pretty much all the time.

hellendag:
> This looks great! I do wonder if we could build this directly into Tooltip? It seems being able to pass no tooltip content and get no tooltip would be nice pretty much all the time.

Actually not a bad idea. Might be nice to have that for `Popover` too.

benpankow (approved):
sweet

PR #8446, opened 2022-06-16
Fold asset_defs submodule into main dagster structure

smackesey:
### Summary & Motivation

Given that assets are no longer an "experiment", it makes sense for them to be harmoniously integrated into the structure of the codebase instead of siloed in a separate module. This PR folds all assets definitions into the `dagster.core.definitions` package, excepting the decorators which are now under `dagster.core.definitions.decorators`.

This should be a purely internal change since assets are imported by users from the top level `dagster`.

Existing BK tests.


smackesey:
* master
  * **PR #8446** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8446" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8446?utm_source=stack-comment).

sryza (approved):
Nice. Before merging this, I think it would be good to do a PSA in #project-software-defined-assets, because it might cause a bunch of merge conflicts for in-flight PRs.

smackesey:
>I just noticed that the tests are still in the old structure - it would be good to move them as well so that they remain easy to find.

OK, on my radar

PR #8450, opened 2022-06-16
[dagit] Repair loading state on Instance Overview

hellendag:
### Summary & Motivation

There is currently a bug in Instance Overview where the loading state clears the entire page briefly. This is because we're now checking the `loading` values on the queries to avoid the weird Apollo caching bug where reloading a repo puts us in an invalid cached state that leads to a JS error.

To resolve this, check `Object.keys()` on the query data to ensure that we're not looking at empty objects. We really shouldn't have to do this, and it may be fixed in more recent versions of Apollo, but it does the trick here.

View Instance Overview.

- Refresh data, verify that there is no flash of a blank screen.
- Reload a repo, verify that it works correctly, with no JS errors.

### DIFF
--- a/js_modules/dagit/packages/core/src/instance/InstanceOverviewPage.tsx
+++ b/js_modules/dagit/packages/core/src/instance/InstanceOverviewPage.tsx
@@ -102,7 +102,7 @@ export const InstanceOverviewPage = () => {
     fetchPolicy: 'network-only',
     notifyOnNetworkStatusChange: true,
   });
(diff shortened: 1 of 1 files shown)


hellendag:
* master
  * **PR #8450** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8450" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8450?utm_source=stack-comment).

bengotow (approved):
Ahh yeah this makes sense 👍  good catch

PR #8455, opened 2022-06-17
[dagit] “Canvas” dots on the DAGs 🎨

bengotow:
### Summary & Motivation

This PR brings "canvas dots" to the DAG visualizations via an SVG css background + position + scale properties.

These stick to the viewport so they get smaller as you zoom and move with the view as you pan around. Props to @braunjj for the idea, I think it makes the viz look more polished and I never would have thought to add it.

Note: We tried using Gray100 and KeylineGray for the dot color but didn't /quite/ like either of them, so this is breaking our color palette slightly. I think it's more important that the dots are visible but not distracting than that they match a color constant.

image
image


hellendag (approved):
Love this. My only note is that the dots look *very* low contrast in the screenshots -- they're pretty hard to see. (My eyes aren't great.)

PR #8456, opened 2022-06-17
BulkAction storage

johannkm:

### DIFF
--- a/python_modules/dagster/dagster/core/execution/backfill.py
+++ b/python_modules/dagster/dagster/core/execution/backfill.py
@@ -29,17 +29,7 @@ from dagster.serdes import whitelist_for_serdes
 from dagster.utils import merge_dicts
 from dagster.utils.error import SerializableErrorInfo
 
-
-@whitelist_for_serdes
-class BulkActionStatus(Enum):
-    REQUESTED = "REQUESTED"
-    COMPLETED = "COMPLETED"
-    FAILED = "FAILED"
-    CANCELED = "CANCELED"
-
-    @staticmethod
-    def from_graphql_input(graphql_str):
-        return BulkActionStatus(graphql_str)
+from .bulk_actions import BulkActionStatus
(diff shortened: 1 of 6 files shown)


prha (commented):
Is there an advantage to having all the stored bulk action payloads using the same namedtuple?

Or should we have different payloads for backfill, bulk deletes, bulk cancels, etc.

Is the reason for keeping backfill and bulk action namedtuple separate because of the difficulty in deserializing the historical fields?

johannkm:
I think this makes sense given what I think the eventual front end will look like. This new class of bulk actions (termination, deletion, re-exectution) all start from the drop down on the runs page. If we had a status page for the progress on bulk actions it would make sense to have these all in a table together. Having one namedtuple with the schema for timestamp, status, etc. makes that easier. Backfills are different, they operate on partitions instead of runs, so it's less likely we'd want to display them all in a table together

johannkm (commented):
BulkRunActions makes sense to me, not sure that it should include all run ids, maybe just a filter?

prha:
Yeah, the only question I have about the filter is it strong enough to accurately describe the set of runs at the time of execution?  Do you support retries, or is it immutable?

johannkm (commented):
Updated to BulkRunAction. Will add the run ids field in https://app.graphite.dev/github/pr/dagster-io/dagster/8457

johannkm (commented):
Should storage methods become `add_bulk_run_action`?

johannkm (commented):
Tests will fail until cloud equivalent lands

prha (approved):
I still think this is going to need the `run_ids` field in it... 

It might also need a checkpointing field so that we can keep track of progress within the daemon (to recover from crashes).  We might want that to happen in a separate DB column, but it's currently implemented in the JSON body for backfills.

Feel free to req another review if you want me to look again.

PR #8458, opened 2022-06-17
[dagit] Fix UI issues around graph backed assets

bengotow:
### Summary & Motivation

This PR makes several small changes:

- When showing a link to a graph-backed SDA's graph, we pull it's actual job name rather than the job name prefix. In my testing, the generated job was __ASSET_JOB_0, not __ASSET_JOB, and the link was broken.

- I tried showing the graph in a modal but it's nice to have the sidebar, ability to go into nested composites etc, so the I made the link still take you to the graph page and:
   + Hid the breadcrumb that would allow you to reach the "root" of the hidden asset graph. At this point, the only place the asset job is exposed is in the URL.
   + Changed the page title to match the header shown on the page, which is the op name when you're within a composite op, and not the asset job name.

- I fixed a small bug where going into a composite op still showed you the "Expand graph" button in the right sidebar. When you're viewing the contents of the parent op, you shouldn't have the option to go into it a second time.

image

  
I clicked around a test asset group containing a graph-backed SDA, and also changed the __ASSET_JOB prefix in Dagit (to make it think it was a normal graph with a normal composite solid) and checked that my changes didn't break the standard ops use case

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/AssetNodeDefinition.tsx
+++ b/js_modules/dagit/packages/core/src/assets/AssetNodeDefinition.tsx
@@ -209,7 +209,7 @@ const OpNamesDisplay = (props: {
   repoAddress: RepoAddress;
 }) => {
   const {assetNode, repoAddress} = props;
-  const {assetKey, graphName, opNames} = assetNode;
+  const {assetKey, graphName, opNames, jobNames} = assetNode;
   const opCount = opNames.length;
 
   if (!opCount) {
@@ -232,16 +232,18 @@ const OpNamesDisplay = (props: {
     );
   }
 
-  const graphPath = workspacePathFromAddress(
-    repoAddress,
-    `/graphs/${__ASSET_JOB_PREFIX}/${graphName}/`,
-  );
+  if (!jobNames.length) {
+    return null;
+  }
 
   return (
     <Box flex={{gap: 4, alignItems: 'center'}}>
       <Icon name="schema" size={16} />
       <Mono>
-        <Link to={graphPath}>{graphName}</Link> ({opCount === 1 ? '1 op' : `${opCount} ops`})
+        <Link to={workspacePathFromAddress(repoAddress, `/graphs/${jobNames[0]}/${graphName}/`)}>
+          {graphName}
+        </Link>
+        {` (${opCount === 1 ? '1 op' : `${opCount} ops`})`}
       </Mono>
     </Box>
   );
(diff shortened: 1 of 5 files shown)


js_modules/dagit/packages/core/src/assets/AssetNodeDefinition.tsx
```diff
@@ -234,7 +234,7 @@ const OpNamesDisplay = (props: {
 
   const graphPath = workspacePathFromAddress(
     repoAddress,
-    `/graphs/${__ASSET_JOB_PREFIX}/${graphName}/`,
+    `/graphs/${assetNode.jobNames[0]}/${graphName}/`,
```
hellendag, line 237: Is `jobNames` guaranteed to have a 0th item?

js_modules/dagit/packages/core/src/pipelines/GraphExplorer.tsx
```diff
@@ -175,6 +176,22 @@ export const GraphExplorer: React.FC<GraphExplorerProps> = (props) => {
...
+    const breadcrumbs = opNames.map((name, idx) => ({
+      text: name,
+      onClick: () => {
+        onChangeExplorerPath({...explorerPath, opNames: opNames.slice(0, idx + 1)}, 'push');
+      },
+    }));
+    // If you're viewing a graph that is part of an asset job, we don't want to let you view
+    // the "root" graph becacuse it's not something you defined explicitly. Remove the first item.
+    if (isHiddenAssetGroupJob(explorerPath.pipelineName)) {
+      breadcrumbs.splice(0, 1);
```
hellendag, line 190: `foo.shift()`

js_modules/dagit/packages/core/src/pipelines/SidebarTabbedContainer.tsx
```diff
@@ -80,7 +80,7 @@ export const SidebarTabbedContainer: React.FC<ISidebarTabbedContainerProps> = (p
             handleID={parentOpHandleID}
             showingSubgraph={true}
             getInvocations={getInvocations}
-            onEnterSubgraph={onEnterSubgraph}
+            onEnterSubgraph={undefined}
```
hellendag, line 83: Unfortunate. Can the prop be omitted?

hellendag (approved):

js_modules/dagit/packages/core/src/pipelines/SidebarTabbedContainer.tsx
bengotow: Oh derp, yes it can 🤦‍♂️ good call

PR #8477, opened 2022-06-19
[dagster-hashicorp] Hashicorp Vault Integration Library

silentsokolov:
### Summary & Motivation

Adding a basic module for [HashiCorp](https://www.vaultproject.io/) which would all for read secret from Vault. Overall this adds:

- A resource for Vault


sryza:
Hey @silentsokolov - thanks for putting this together! I Very cool.  We've actually been chatting recently about best practices for managing secrets with Dagster, and I want to make sure we're giving users a consistent way to think about secrets.

One question I've been thinking about have is how we'd expect a resource like this to work with other library resources. E.g. suppose someone wants to make a resource that connects to snowflake, and they need a secret to be able to do that - how would we expect their resource implementation to access secrets from the vault? Would every resource that uses secrets be expect to have a resource dependency on a secretmanager resource?

silentsokolov:
Hey @sryza I think dagster library must provide the different bricks (resources or ops, etc) for the building of pipelines. For example, `vault_resource` or `s3_resource`. I think it is the right way.

These the bricks also allow to create high-level abstractions. For example we could create a new type StringSecret as [StringSource](https://docs.dagster.io/_apidocs/config#dagster.StringSource) but it gets value from a secret store against from an environment variable. The StringSecret could be part of package a `dagster-secret`, users would need to set the environment variables (`VAULT_ADDR`, `VAULT_TOKEN`, etc - for use vault or `AWS_SECRET_ACCESS_KEY`, etc - for aws ... `GCP_SOMETHING` - for gcp).

In our team we did something similar.

```python
from dagster import job, op

@op(config_schema=StringSecret)
def secret_op(context) -> str:
    return context.op_config

@job
def secret_job():
    secret_op()

secret_job.execute_in_process(
    run_config={
        'ops': {'secret_op': {'config': 'my_super_password'}}  # in local develop and test
    }
)

secret_job.execute_in_process(
    run_config={
        'ops': {'secret_op': {'config': '~vault:path/to/secret#name'}}  # if a string start with "~vault:" we create request to vault
    }
)

secret_job.execute_in_process(
    run_config={
        'ops': {'secret_op': {'config': '~aws:secret_name'}}  # if a string start with "~aws:" we create request to aws secret manager
    }
)
```

I know prefixes `~aws:` or  `~vault:` are not the best practice, but they solve many problems for us.

silentsokolov:
@sryza Could you please review this RP?

sryza:
Apologies for the delay @silentsokolov.

For an orchestrator like Dagster, a big part of the value we can provide with secrets support will come from standardizing how secrets are accessed.  Before introducing libraries for accessing secrets, we'd like to do some design research to develop an opinion on this.

I started a [Github discussion](https://github.com/dagster-io/dagster/discussions/8644) to hold this discussion.

In the mean time, if you would be interested in self-hosting this library, we'd love to publicize it and link to it from our docs.

We really appreciate this contribution.  However, we've gotten a lot of feedback that there are too many different ways to do things in Dagster, so we want to put extra effort into standardized recommendations before adding new functionality.

sryza:
Also @silentsokolov - I'd love to get your thoughts on Dagster & secrets in general if you'd be interested in doing a call at some point. If so, DM me in the Dagster Slack.

silentsokolov:
Ok, I see!
Thanks you!

silentsokolov:
For history: https://github.com/silentsokolov/dagster-hashicorp

sryza:
Thanks for the link @silentsokolov. We referenced it on our library integrations page: https://github.com/dagster-io/dagster/pull/9368.

PR #8479, opened 2022-06-19
[dagit] Collapsed DAG rendering of multiple edges between the same ops

bengotow:
### Summary & Motivation

This PR aims to address the bad op DAG I/O rendering shown here:
Screen Shot 2022-06-19 at 3 13 35 PM

Changes: 

- If there are too many I/Os on any given op, we switch to a "dots-only" rendering, but we weren't turning off the labels until you zoomed out. This PR fixes this so the text isn't overflowing all over in the screenshot.

- If there are multiple I/Os connecting the same two ops, we sort them to avoid the unnecessary crossed lines and also compact them into a bundle.
 
- If there are a /LOT/ of inputs and outputs and even the compact rendering causes the dots to overflow (bottom of the screenshot above), we draw them closer together.

After:

image

- Tested using `elementl.dagster.cloud` example
- Tested composite op and regular op rendering to make sure good rendering cases weren't impacted

### DIFF
--- a/js_modules/dagit/packages/core/src/graph/OpEdges.tsx
+++ b/js_modules/dagit/packages/core/src/graph/OpEdges.tsx
@@ -21,9 +21,18 @@ type Path = {
 const buildSVGPaths = weakmapMemoize((edges: OpLayoutEdge[], nodes: {[name: string]: OpLayout}) =>
   edges
     .map(({from, to}) => {
-      const sourceOutput = nodes[from.opName].outputs[from.edgeName];
-      const targetInput = nodes[to.opName].inputs[to.edgeName];
+      const source = nodes[from.opName];
+      const sourceOutput =
+        source.outputs[from.edgeName] ||
+        Object.values(source.outputs).find((o) => o.collapsed.includes(from.edgeName));
+
+      const target = nodes[to.opName];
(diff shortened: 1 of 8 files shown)


rexledesma:
Rather than compacting lines into a bundle, could they just merge into one line? Then on hover, you could see the specific connections, as a list? 

Same thing for dots overflow - could it be compressed into one dot? Or maybe compressed to a single, clickable label called "outputs"?

hellendag (approved):

bengotow:
Hey @rexledesma! Hmm, I think we can do it that way too - do you know if it's useful to know at a glance that there are multiple bound inputs / outputs? I thought it might be useful to see that there are 7 separate bound outputs but maybe that's not important...

bengotow:
Hey gang! I've tweaked this PR to collapse the inputs / outputs instead of just packing them together. I also created a new job in our toys repo for exploring edge cases around this, since it wasn't covered by any of our existing ones.

Here's what it looks like now:

image

If an op has more than 5 inputs / outputs, we switch away from the detailed rendering with labels and render just the dots. If some of the inputs or outputs have the same upstream or downstream op, they're compacted into a single dot with "+3" indicating there are collapsed edges!

PR #8483, opened 2022-06-20
[dagit] Repair empty config case in Launchpad

hellendag:
### Summary & Motivation

The recent change to send config yaml instead of JSON led to breakages with empty configs:

- The backend doesn't like the empty string. Send an empty object instead.
- The `responseToYamlValidationResult` function tries `yaml.parse` an empty string and evaluate the resulting `Object.keys`, leading to a JS error.

Repair these by using empty objects, as-is or stringified as needed.

View a job with no config required. Verify that config is considered valid by the backend.

Launch a run, verify that the run is created successfully with the empty config.


bengotow (approved):
This looks good to me!

PR #8490, opened 2022-06-20
Make default_executor_def on repo work on jobs

dpeng817:
Makes default_executor_def on the repository overwrite the executor def provided to jobs. Unfortunately, requires the use of a big copy constructor. I thought about using an in-place method to replace the executor def, but that could have ramifications for providing jobs to multiple repos and expecting different default executor defs to be used in each. So I think this is the best we can do.


python_modules/dagster/dagster/core/definitions/repository_definition.py
```diff
@@ -788,6 +789,11 @@ def from_list(
             else:
                 pipelines[name] = pipeline_or_job
 
+        if default_executor_def:
+            for name, job_def in jobs.items():
+                if job_def.executor_def == multi_or_in_process_executor:
```
sryza, line 794: What if someone sets a default executor def and then also sets `multi_or_in_process_executor` on one of their jobs?

dpeng817: In this world, that executor would get overridden. I guess we could protect against that?

Nintorac:
I am using `0.15.2` and the `default_executor_def` is not working for jobs in my repo. Was this included in `0.15.2`?

Cannot find anyway to set executor for jobs created from assets i.e with `define_asset_job`

Current workaround that I know of is to use `build_asset_selection_job` which as I understand is an internal API and subject to change

PR #8506, opened 2022-06-20
Add constant for default io manager key

dpeng817:
We use the default io manager key all over the place as a plain string, which gives me major heebie jeebies. Changes to use an importable constant.

I would feel even better about being able to annotate it in docstrings using this constant as well. Was thinking maybe we could interpolate it, or better yet just reference it using sphinx machinery and export it.


sryza (approved):
great call

PR #8508, opened 2022-06-20
AssetsDefinition.from_op

clairelin135:
We've gotten requests from users to build an asset from an op (e.g. for dagstermill ops to be converted to assets). This utility method makes it possible to construct an `AssetsDefinition` from an op in the same way that `AssetsDefinition.from_graph` infers input/output assets.


sryza (approved):
Looks great!

PR #8512, opened 2022-06-20
Add default_logger_defs arg to repository

dpeng817:
Adds a `default_logger_defs` argument to the repository. Default logger defs are applied to asset jobs, and jobs where no loggers are explicitly specified.

### DIFF
--- a/docs/content/concepts/logging/loggers.mdx
+++ b/docs/content/concepts/logging/loggers.mdx
@@ -193,6 +193,31 @@ width={2048}
 height={804}
 />
 
+### Specifying Default Loggers
+
+Default loggers can be specified on a <PyObject object="repository" decorator /> using the `default_logger_defs` argument. These loggers will be added to every job specified on the repository, and will be used for any materializatons of software-defined assets provided to the repository.
+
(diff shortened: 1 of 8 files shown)


sryza:
>  and jobs where no loggers are explicitly specified.

Loggers are additive, so I wonder if we should just combine them? I don't have particular opinions here, but some rationale in either direction would be worthwhile.

dpeng817:
> > and jobs where no loggers are explicitly specified.
> 
> Loggers are additive, so I wonder if we should just combine them? I don't have particular opinions here, but some rationale in either direction would be worthwhile.

Making it additive just introduces edge cases and intricacies regarding collisions. But I'm not against it.

docs/content/concepts/logging/loggers.mdx
```diff
@@ -193,6 +193,31 @@ width={2048}
 height={804}
 />
 
+### Specifying Default Loggers
```
sryza, line 196: I think it's worth mentioning this is at the repository level in the heading.

sryza (approved):

docs/content/concepts/logging/loggers.mdx
dpeng817:
```suggestion
### Specifying Default Loggers on Repository
```

PR #8517, opened 2022-06-21
group-based asset dependency resolution

sryza:
### Summary & Motivation

When asset keys are long, getting them to match up when specifying dependencies can be hard and fiddly.  Owen and I have both personally experienced this when dogfooding, and I've observed users report it as well.

This PR proposes falling back to matching based on the last segment of the asset key, if both assets are in the same group.

For now, this is marked experimental. The big drawback here is that asset dependency resolution becomes a more complicated decision tree, and thus can be more difficult to debug.  I think we should feel this out for a few releases before marking it stable. 

A couple interesting edge cases to point out:
* Assets within a multi-asset can belong to different groups. In this case, we turn don't do group-based dependency resolution.
* Two assets within the same group can have the same name. In this case, we turn don't do group-based dependency resolution.

Added tests. Observed experimental warning.

### DIFF
--- a/python_modules/dagster/dagster/core/asset_defs/resolved_asset_deps.py
+++ b/python_modules/dagster/dagster/core/asset_defs/resolved_asset_deps.py
@@ -1,0 +1,75 @@
+from typing import AbstractSet, Dict, Iterable, List, Mapping, Sequence, Set, Tuple
+
+from dagster.core.definitions.events import AssetKey
+from dagster.core.errors import DagsterInvalidDefinitionError
+
+from .assets import AssetsDefinition
+from .source_asset import SourceAsset
+
+
+class ResolvedAssetDependencies:
+    def __init__(
+        self, assets_defs: Iterable[AssetsDefinition], source_assets: Iterable[SourceAsset]
+    ):
+        self._deps_by_assets_def_id = resolve_assets_def_deps(assets_defs, source_assets)
+
+    def get_resolved_upstream_asset_keys(
+        self, assets_def: AssetsDefinition, asset_key: AssetKey
+    ) -> AbstractSet[AssetKey]:
+        return self._deps_by_assets_def_id[id(assets_def)].values()
+
+    def get_resolved_asset_key_for_input(
+        self, assets_def: AssetsDefinition, input_name: str
+    ) -> AssetKey:
+        return self._deps_by_assets_def_id[id(assets_def)][input_name]
+
+
+def resolve_assets_def_deps(
+    assets_defs: Iterable[AssetsDefinition], source_assets: Iterable[SourceAsset]
+) -> Mapping[int, Mapping[str, AssetKey]]:
+    """
+    For each AssetsDefinition, resolves its inputs to upstream asset keys. Matches based on either
+    of two criteria:
+    - The input asset key exactly matches an asset key.
+    - The input asset key has one component, that component matches the final component of an asset
+        key, and they're both in the same asset group.
+    """
+    asset_keys_by_group_and_name: Dict[Tuple[str, str], AssetKey] = {}
+    for assets_def in assets_defs:
+        for key in assets_def.keys:
+            asset_keys_by_group_and_name[(assets_def.group_names_by_key[key], key.path[-1])] = key
+    for source_asset in source_assets:
+        asset_keys_by_group_and_name[
+            (source_asset.group_name, source_asset.key.path[-1])
+        ] = source_asset.key
+
+    asset_keys = set(asset_keys_by_group_and_name.values())
+
+    result: Mapping[int, Mapping[str, AssetKey]] = {}
+    for assets_def in assets_defs:
+        group = (
+            next(iter(assets_def.group_names_by_key.values()))
+            if len(assets_def.group_names_by_key) == 1
+            else None
+        )
+
+        dep_keys_by_input_name: Dict[str, AssetKey] = {}
+        for input_name, upstream_asset_key in assets_def.keys_by_input_name.items():
+            group_and_upstream_name = (group, upstream_asset_key.path[-1])
+            if upstream_asset_key in asset_keys:
+                dep_keys_by_input_name[input_name] = upstream_asset_key
+            elif group is not None and group_and_upstream_name in asset_keys_by_group_and_name:
+                dep_keys_by_input_name[input_name] = asset_keys_by_group_and_name[
+                    group_and_upstream_name
+                ]
+            else:
+                raise DagsterInvalidDefinitionError(
+                    f"Input asset '{upstream_asset_key.to_string()}' for asset "
+                    f"'{next(iter(assets_def.keys)).to_string()}' is not "
+                    "produced by any of the provided asset ops and is not one of the provided "
+                    "sources"
+                )
+
+        result[id(assets_def)] = dep_keys_by_input_name
+
+    return result
--- a/python_modules/dagster/dagster/core/asset_defs/resolved_asset_deps.py
+++ b/python_modules/dagster/dagster/core/asset_defs/resolved_asset_deps.py
@@ -1,13 +1,21 @@
-from typing import AbstractSet, Dict, Iterable, List, Mapping, Sequence, Set, Tuple
+from typing import AbstractSet, Dict, Iterable, Mapping, Tuple, cast
 
 from dagster.core.definitions.events import AssetKey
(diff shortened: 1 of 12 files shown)


python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets_job.py
```diff
@@ -1649,3 +1619,57 @@ def my_derived_asset(my_source_asset):  # pylint: disable=unused-argument
...
+
+
+def test_resolve_dependency_in_group():
+    @asset(key_prefix="abc")
+    def asset1():
+        ...
+
+    @asset
+    def asset2(context, asset1):
+        del asset1
```
dpeng817, line 1631: why is `del` necessary here?

python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets_job.py
```diff
@@ -1649,3 +1619,57 @@ def my_derived_asset(my_source_asset):  # pylint: disable=unused-argument
...
+        assert context.asset_key_for_input("asset1") == AssetKey(["abc", "asset1"])
+
+    the_job = build_assets_job(name="test", assets=[asset1, asset2])
+    assert the_job.execute_in_process().success
```
dpeng817, line 1635: use `materialize_to_memory` for tests pls

dpeng817:
`Two assets within the same group can have the same name` can you explain this a bit more? Don't assets not have names, and asset keys have to be unique, yes?

dpeng817 (commented):
Why does this fuzzy resolution work differently than exact resolution across asset groups exactly? It feels like a weird dichotomy to me that exact resolution works across asset groups, but prefix resolution does not, and the error message does not make that very clear. I feel like this is going to throw someone for a loop. 
Otherwise, would be nice to see the multi asset with different groups case represented under test.

python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets_job.py
sryza: It's an unused argument - following the top-upvoted answer on this stackoverflow question: https://stackoverflow.com/questions/10025680/how-can-i-denote-unused-function-arguments

python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets_job.py
sryza: oo hell yes

python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets_job.py
```diff
@@ -1649,3 +1619,57 @@ def my_derived_asset(my_source_asset):  # pylint: disable=unused-argument
...
+
+def test_resolve_dependency_fail_across_groups():
+    @asset(key_prefix="abc", group_name="other")
+    def asset1():
+        ...
+
+    @asset
+    def asset2(context, asset1):
+        del asset1
+        assert context.asset_key_for_input("asset1") == AssetKey(["abc", "asset1"])
```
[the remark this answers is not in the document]
sryza: ah not the latter one, will remove

clairelin135:
I like this functionality a lot and I think it will greatly simplify our users' lives. However, the only part that I still feel weird about is that I think an `AssetKey` should be an absolute path to the asset and that we should not infer dependency relationships when an `AssetKey` has been explicitly passed unless if it matches exactly.

I'd almost prefer passing an `AssetName` object that implies that we can do suffix matching.

python_modules/dagster/dagster/core/asset_defs/resolved_asset_deps.py
```diff
@@ -0,0 +1,108 @@
...
+            if upstream_key not in asset_keys and len(upstream_key) == 1:
+                group_and_upstream_name = (group, upstream_key.path[-1])
+                if group is not None and group_and_upstream_name in asset_keys_by_group_and_name:
+                    resolved_key = asset_keys_by_group_and_name[
+                        cast(Tuple[str, str], group_and_upstream_name)
+                    ]
+                    resolved_keys_by_unresolved_key[upstream_key] = resolved_key
+
+                    if not warned:
+                        experimental_warning(
```
clairelin135, line 88: I think we should update this warning message to contain the full list of inferences

sryza: I had originally considered that, but there could be hundreds, which would be overwhelming.

sryza:
> I like this functionality a lot and I think it will greatly simplify our users' lives. However, the only part that I still feel weird about is that I think an AssetKey should be an absolute path to the asset and that we should not infer dependency relationships when an AssetKey has been explicitly passed unless if it matches exactly.
> I'd almost prefer passing an AssetName object that implies that we can do suffix matching.

@clairelin135 I hear you on this. Do you have specific thoughts on what this would look like from a user perspective?

clairelin135:
@sryza Something like this:
```
@asset(group_name="foo", key_prefix=["prefix_1, prefix_2"])
def upstream():
   ...

@asset(ins={"in_1": AssetName("upstream")}, group_name="foo")
def downstream(in_1):
   ...
```
The exact ergonomics of `AssetIn`/`ins` are still up for discussion, but I would feel better about the fuzzy matching if we had some asset suffix entity that didn't imply an absolute path like an asset key.

sryza:
@clairelin135 - IMO, ideally, we'd allow people to avoid specifying `ins` in as many situations as possible.  E.g. for the example you posted above, writing it like this: 

```
@asset(group_name="foo")
def downstream(upstream):
   ...
```

Thoughts on allowing people to write it like that? Is there a particular way that we would _not_ want users to specify dependencies, that, with the current implementation in this PR, they'd be allowed to?

sryza:
> It feels weird to me that you can specify dependencies this way with an explicit asset key and dagster will infer a dependency relationship here.

👍👍 - I'll investigate changing this.

sryza:
I took a stab at implementing the change you brought up @clairelin135, and it raised a set of thorny issues:
* Can an asset's name be different than the last segment of its key?
* How do we represent this in the AssetsDefinition? A set alongside `keys_by_input_name` that's called something like `explicitly_passed_keys`. Thread a new `AssetName` class throughout the codebase? Do we expose that to users?

@clairelin135 what would you think about merging this today (as experimental) and then investigating the asset name change next week? I'm worried that, with all our marketing and public content around assets, people will try them, but then get discouraged by key prefixing difficulties, so I want to try to get this change in ASAP. I think that specifying an AssetKey with a single segment on an AssetIn will be a relatively rare thing to do, and making a breaking change to how dependencies are handled in that case wouldn't affect many users.

clairelin135:
Got it, that makes sense. Approving now

shalabhc (approved):
(I see this just got merged but wanted to leave the comment anyway for future discussion)

I think it's ok for now to have this convenient group based aliasing to make asset references easier to use. Users might want aliases for cross group references as well, but limiting this to just the current group and marking it experimental makes sense.

For a longer term solution, we should think about refactoring some code to separate the concept of an asset reference from an asset key. The main difference is that a `reference` will be resolved to a `key` at some point later in time. At asset definition time we should only store a reference (not a key, like we do currently) and define a specific point in time where all references are resolved to keys. I think this resolution logic should be scoped to the repository and unrelated to jobs.

sryza:
@shalabhc - that sounds sensible to me

python_modules/dagster/dagster/core/asset_defs/resolved_asset_deps.py
```diff
@@ -0,0 +1,111 @@
...
+    - The input asset key has one component, that component matches the final component of an asset
+        key, and they're both in the same asset group.
+
+    The returned dictionary only contains entries for assets definitions with group-resolved asset
+    dependencies.
+    """
+    asset_keys_by_group_and_name: Dict[Tuple[str, str], AssetKey] = {}
+    for assets_def in assets_defs:
+        for key in assets_def.keys:
+            asset_keys_by_group_and_name[(assets_def.group_names_by_key[key], key.path[-1])] = key
```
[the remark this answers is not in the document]
sryza: oo this seems like a bug. I filed an issue to track: https://github.com/pulls

PR #8523, opened 2022-06-21
Add community meeting graph-backed asset example

clairelin135:


sryza:
@clairelin135 I see that this is marked as a Draft - are there particular changes you're planning to make?

clairelin135:
@sryza nope--mistakenly marked this as a draft. It's ready for review

sryza (changes requested):
Currently, for historical reasons, the "software_defined_assets" example folder refers to this particular example: https://docs.dagster.io/guides/dagster/software-defined-assets. We should probably rename that folder to something like "software_defined_assets_pandas_pyspark", though you don't need to do that unless you feel motivated.  A new example should get its own folder though.

sryza (approved):
This looks great

PR #8528, opened 2022-06-21
Use status column for reading runs from sql storage

johannkm:
Use the status in the status column of run rows, rather than whatever is serialized in the run body

This makes it harder for concurrent writes to a run- e.g. updating status and updating tags- to overwrite eachother.

This has test coverage by the handle_run_event tests, a little tricky to write more

Cloud will follow

### DIFF
--- a/python_modules/dagster/dagster/core/storage/runs/sql_run_storage.py
+++ b/python_modules/dagster/dagster/core/storage/runs/sql_run_storage.py
@@ -174,15 +174,20 @@ class SqlRunStorage(RunStorage):  # pylint: disable=no-init
                 RunsTable.update()  # pylint: disable=no-value-for-parameter
                 .where(RunsTable.c.run_id == run_id)
                 .values(
-                    status=new_pipeline_status.value,
                     run_body=serialize_dagster_namedtuple(run.with_status(new_pipeline_status)),
+                    status=new_pipeline_status.value,
                     update_timestamp=now,
                     **kwargs,
                 )
             )
 
-    def _row_to_run(self, row: Tuple) -> PipelineRun:
-        return deserialize_as(row[0], PipelineRun)
+    def _row_to_run(self, row) -> PipelineRun:
+        run = deserialize_as(row["run_body"], PipelineRun)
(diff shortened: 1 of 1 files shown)


johannkm:
* master
  * **PR #8528** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8528" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8528?utm_source=stack-comment).

python_modules/dagster/dagster/core/storage/runs/sql_run_storage.py
```diff
@@ -143,11 +145,6 @@ def handle_run_event(self, run_id: str, event: DagsterEvent):
         if event.event_type not in EVENT_TYPE_TO_PIPELINE_RUN_STATUS:
             return
 
-        run = self.get_run_by_id(run_id)
```
johannkm, line 146: I wonder how many times we fetch the run from the db per event write haha. Now that -1

python_modules/dagster/dagster/core/storage/runs/sql_run_storage.py
```diff
@@ -175,14 +170,15 @@ def handle_run_event(self, run_id: str, event: DagsterEvent):
...
                     **kwargs,
                 )
             )
 
-    def _row_to_run(self, row: Tuple) -> PipelineRun:
-        return deserialize_as(row[0], PipelineRun)
+    def _row_to_run(self, row) -> PipelineRun:
+        run = deserialize_as(row["run_body"], PipelineRun)
+        status = DagsterRunStatus(row["status"])
+        return run.with_status(status)
```
gibsondan, line 181: maybe worth a commnent here explaining why we can't trust what's in run_body

gibsondan (approved):
nice, are you going to do an internal one too?

I assume there's a bunch of implicit test coverage here since if a run stayed at its initial status forever lots of things would break.. could imagine an explicit test that calls handle_run_event and verifies that the resulting run has the correct status when pulled from the DB, if that doesn't already exist

PR #8529, opened 2022-06-21
Add resource_defs arg to multi_assets

dpeng817:
Adds resource_defs as an arg to multi_assets. Along the way, made some small tweaks to the way required_resource_keys were generated for regular SDAs, namely that io manager defs provided do not end up in the required resource keys of an asset. Also added documentation for the resource_defs arg of @asset

### DIFF
--- a/python_modules/dagster/dagster/core/asset_defs/decorators.py
+++ b/python_modules/dagster/dagster/core/asset_defs/decorators.py
@@ -259,6 +259,10 @@ class _Asset:
         with warnings.catch_warnings():
             warnings.simplefilter("ignore", category=ExperimentalWarning)
 
+            required_resource_keys = set(self.required_resource_keys).union(
+                set(self.resource_defs.keys())
+            )
+
             if isinstance(self.io_manager, str):
                 io_manager_key = cast(str, self.io_manager)
             elif self.io_manager is not None:
@@ -278,13 +282,6 @@ class _Asset:
                 description=self.description,
             )
 
-            required_resource_keys = set()
-            for key in self.required_resource_keys:
-                required_resource_keys.add(key)
-            for key in self.resource_defs.keys():
-                # Ensure that io managers not available from context on op.
-                if not isinstance(self.resource_defs[key], IOManagerDefinition):
-                    required_resource_keys.add(key)
             op = _Op(
                 name="__".join(out_asset_key.path).replace("-", "_"),
                 description=self.description,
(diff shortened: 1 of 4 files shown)


python_modules/dagster/dagster/core/asset_defs/decorators.py
```diff
@@ -145,6 +145,8 @@ def asset(
             `json.loads(json.dumps(value)) == value`.
         group_name (Optional[str]): A string name used to organize multiple assets into groups. If not provided,
             the name "default" is used.
+        resource_defs (Optional[Mapping[str, ResourceDefinition]]):
+            A mapping of resource keys to resource definitions. These resources will be initialized during execution, and can be accessed from the context within the body of the function.
```
sryza, line 149: Line length

python_modules/dagster/dagster/core/asset_defs/decorators.py
```diff
@@ -278,6 +280,8 @@ def __call__(self, fn: Callable) -> AssetsDefinition:
             for key in self.required_resource_keys:
                 required_resource_keys.add(key)
             for key in self.resource_defs.keys():
+                if isinstance(self.resource_defs[key], IOManagerDefinition):
```
sryza, line 283: What's the rationale for this check? Also, best to avoid `continue` when not needed. Could just do `if isinstance(self.resource_defs[key], IOManagerDefinition):`?

dpeng817: This check is to ensure we don't accidentally make io managers available to the body of an op via the resources object, which we don't do elsewhere.

python_modules/dagster/dagster/core/asset_defs/decorators.py
```diff
@@ -278,7 +282,8 @@ def __call__(self, fn: Callable) -> AssetsDefinition:
             for key in self.required_resource_keys:
                 required_resource_keys.add(key)
             for key in self.resource_defs.keys():
-                required_resource_keys.add(key)
+                if not isinstance(self.resource_defs[key], IOManagerDefinition):
```
clairelin135, line 285: got confused here by this check as well. Would be great to add a comment here to explain

clairelin135 (approved):

python_modules/dagster/dagster/core/asset_defs/decorators.py
dpeng817: As it turns out, I'm wrong. IO manager defs _are_ available from within the body of an op.

PR #8530, opened 2022-06-21
Specifying executor docs examples

dpeng817:
As the title.


docs/content/deployment/executors.mdx
```diff
@@ -22,6 +22,24 @@ Every job has an executor. The default executor is the <PyObject object="multi_o
...
+@graph
+def the_graph():
+    ...
+
+
+# Providing an executor using graph_def.to_job(...)
+other_job = the_graph.to_job(executor_def=multiprocess_executor)
+```
+
 A default executor can be specified for all jobs and assets provided to a repository using the `default_executor_def` argument of <PyObject object="repository" decorator />. All jobs that don't specify an executor will use this default executor, but if a job explicitly specifies an executor, then the default provided to the repository will not be used.
```
jamiedemaria, line 43: while you're adding code snippets, an example of this could also be helpful

dpeng817: whoops I did, just forgot to reference it in the markdown.

sryza (approved):

PR #8531, opened 2022-06-21
[dagit] Add control for graph depth on the Asset Lineage page, default to 5

bengotow:
### Summary & Motivation
Jun-21-2022 14-58-19

This PR adds the depth control above to the lineage views and makes the upstream / downstream perspectives default to 5 layers.

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/useAssetGraphData.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/useAssetGraphData.tsx
@@ -1,7 +1,9 @@
 import {gql, useQuery} from '@apollo/client';
+import keyBy from 'lodash/keyBy';
 import React from 'react';
 
 import {filterByQuery, GraphQueryItem} from '../app/GraphQueryImpl';
+import {AssetKey} from '../assets/types';
 import {AssetGroupSelector, PipelineSelector} from '../types/globalTypes';
 
 import {ASSET_NODE_FRAGMENT} from './AssetNode';
@@ -17,6 +19,11 @@ export interface AssetGraphFetchScope {
   pipelineSelector?: PipelineSelector;
   groupSelector?: AssetGroupSelector;
 }
+
+export type AssetGraphQueryItem = GraphQueryItem & {
+  node: AssetNode;
+};
+
 /** Fetches data for rendering an asset graph:
  *
  * @param pipelineSelector: Optionally scope to an asset job, or pass null for the global graph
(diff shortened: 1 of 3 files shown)


js_modules/dagit/packages/core/src/asset-graph/useAssetGraphData.tsx
```diff
@@ -130,6 +133,34 @@ const removeEdgesToHiddenAssets = (graphData: GraphData) => {
   }
 };
 
+export const calculateGraphDistances = (items: AssetGraphQueryItem[], assetKey: AssetKey) => {
```
hellendag, line 136: Would you mind writing a test for this? Also, is there any possibility that it could have some super pathological outcome on extremely wide/deep graphs?

js_modules/dagit/packages/core/src/assets/AssetNodeLineage.tsx
```diff
@@ -65,3 +101,57 @@ export const AssetNodeLineage: React.FC<{
...
+  onChange: (v: number) => void;
+}> = ({value, max, onChange}) => {
+  const commitValue = (target: HTMLInputElement) => {
+    const next = Number(target.value) ? Math.min(max, Number(target.value)) : value;
+    onChange(next);
+    target.value = `${next}`;
+  };
+
+  return (
+    <Box flex={{gap: 8, alignItems: 'center'}}>
```
hellendag, line 117: Is this Box wrapper needed?

js_modules/dagit/packages/core/src/assets/AssetNodeLineage.tsx
```diff
@@ -65,3 +101,57 @@ export const AssetNodeLineage: React.FC<{
...
+
+const LineageDepthControl: React.FC<{
+  value: number;
+  max: number;
+  onChange: (v: number) => void;
+}> = ({value, max, onChange}) => {
+  const commitValue = (target: HTMLInputElement) => {
+    const next = Number(target.value) ? Math.min(max, Number(target.value)) : value;
+    onChange(next);
+    target.value = `${next}`;
```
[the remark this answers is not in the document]
bengotow: Ahh man... I debated making it a controlled input, but we'd need a separate piece of state for the value because I want it to be ok to erase the value (to "") and then type in another value without it inserting a `0`. Committing on blur / enter and normalizing the value is definitely a bit of a hack though... let me try the controlled version again and see if I can make it nice.

js_modules/dagit/packages/core/src/asset-graph/useAssetGraphData.tsx
bengotow: Yep will do 🙌

Issue #7434, opened 2022-04-13
asset says "upstream changed" even when materialized after partitioned upstream asset


sryza:
Reported here: https://dagster.slack.com/archives/C01U954MEER/p1649863788378579

![image](https://user-images.githubusercontent.com/654855/163242159-a5e917ab-c7fa-44a5-b6ac-0d45987a7ce9.png)

To reproduce:
* Dagit -f this:
```
from dagster import asset, AssetGroup, repository, DailyPartitionsDefinition


@asset(partitions_def=DailyPartitionsDefinition(start_date="2020-02-02"))
def asset1():
    pass


@asset
def asset2(asset1):
    pass


group = AssetGroup.from_current_module()


@repository
def my_job():
    return [group]
```
* Materialize one of the partitions of `asset1`
* Materialize one of the partitions of `asset2`
* Observe that `asset2` says "Upstream changed"

hellendag:
Let's discuss at the next SDA UI meeting.

PR #8535, opened 2022-06-21
[dagit] Move Asset “Upstream Changed” computation to Python

bengotow:
### Summary & Motivation

Hey folks! This PR is my attempt to extend @clairelin135's excellent work on the `assetsLatestInfo` resolver to calculate the "upstream changed" status of each asset in Python instead of in Dagit. There are several reasons for this change:

- The "upstream changed" state of assets cascade. Given `A => B => C`, if B is older than A, both B and C display "upstream changed". This means that calculating this value for a given asset requires the latest materialization timestamps of ALL upstream assets across repos, asset groups, etc. Dagit increasingly does not have this info loaded (asset group pages, etc.). We're computing "upstream changed" with the subset of data it *does* have, and it produces inconsistent results based on what subgraph of assets you're looking at.

- The "upstream changed" calculation currently ignores partitioned assets and source assets, and Dagit needs to request partition definitions, etc. just for the purposes of running this algorithm.

- Having this computed server side allows us to render the status in more places more easily!

The backend implementation of this is a basic translation of the JS version, so it may not be the most performant / pythonic approach! Super open to changes.

**Specific Behaviors:**

- Assets have a "NONE" status if they have never been materialized.

- Assets have an "OUT_OF_DATE" status if:
   + One of their dependencies has a newer materialization date
   + One of their dependencies has a NONE or OUT_OF_DATE status.

- Assets have an "UP_TO_DATE" status if:
    + They are partitioned assets (not supported)
    + They are source assets (not tracked)
    + They do not match the NONE or OUT_OF_DATE conditions
    
If you have a partitioned asset (always UP_TO_DATE) which "sinks" into an unpartitioned asset, the unpartitioned asset essentially does not consider the partitioned asset, but will still consider other inputs. (Addresses https://github.com/dagster-io/dagster/issues/7434) I /think/ this is the best way to do this but I'm open to thoughts!

New UI:

In the asset catalog!
image

In the top nav on asset details! (clicking this takes you to upstream lineage)
image

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/Utils.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/Utils.tsx
@@ -1,5 +1,7 @@
 import {pathVerticalDiagonal} from '@vx/shape';
 
+import {AssetComputeStatus} from '../types/globalTypes';
+
 import {
   AssetGraphLiveQuery_assetsLatestInfo_latestRun,
   AssetGraphLiveQuery_assetNodes_assetMaterializations,
@@ -108,14 +110,13 @@ export const buildSVGPath = pathVerticalDiagonal({
   y: (s: any) => s.y,
 });
 
-export type ComputeStatus = 'good' | 'old' | 'none' | 'unknown';
-
 export interface LiveDataForNode {
   stepKey: string;
   unstartedRunIds: string[]; // run in progress and step not started
   inProgressRunIds: string[]; // run in progress and step in progress
   runWhichFailedToMaterialize: AssetGraphLiveQuery_assetsLatestInfo_latestRun | null;
   lastMaterialization: AssetGraphLiveQuery_assetNodes_assetMaterializations | null;
+  computeStatus: AssetComputeStatus;
 }
 export interface LiveData {
   [assetId: GraphId]: LiveDataForNode;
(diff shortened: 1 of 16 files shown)


bengotow:
Related:
https://github.com/dagster-io/dagster/issues/7434
https://github.com/dagster-io/dagster/issues/6728

python_modules/dagster-graphql/dagster_graphql/implementation/fetch_runs.py
```diff
@@ -122,21 +125,89 @@ def get_runs(graphene_info, filters, cursor=None, limit=None):
 ]
 
 
+class AssetComputeStatus(Enum):
+    OLD = "OLD"
```
prha, line 129: maybe `OUT_OF_DATE`, `NONE`, `UP_TO_DATE`?

js_modules/dagit/packages/core/src/asset-graph/Utils.tsx
```diff
@@ -202,8 +154,9 @@ export const buildLiveData = ({assetNodes, assetsLatestInfo}: AssetGraphLiveQuer
     data[graphId] = {
       lastMaterialization,
       stepKey: liveNode.opNames[0],
-      inProgressRunIds: assetLiveRuns?.inProgressRunIds || [],
-      unstartedRunIds: assetLiveRuns?.unstartedRunIds || [],
+      inProgressRunIds: assetLatestInfo?.inProgressRunIds || [],
+      unstartedRunIds: assetLatestInfo?.unstartedRunIds || [],
+      computeStatus: assetLatestInfo?.computeStatus || AssetComputeStatus.UNKNOWN,
```
prha, line 159: do we need this unknown status default?  when is this used?

clairelin135 (commented):
Really excited about this! Agree with phil on the comments, I added one more about a minor refactor.

python_modules/dagster-graphql/dagster_graphql/implementation/fetch_runs.py
```diff
@@ -122,21 +125,89 @@ def get_runs(graphene_info, filters, cursor=None, limit=None):
...
+        asset_node = all_asset_nodes.get(asset_key)
+        if not asset_node:
+            continue
+        last_changed[asset_key] = asset_record.asset_entry.last_materialization.timestamp
+        if not asset_node.op_names:
+            status[asset_key] = AssetComputeStatus.GOOD  # source assets are always "up to date"
+        elif asset_node.partitions_def_data:
+            status[asset_key] = AssetComputeStatus.NONE  # partitioned assets are not supported
+        elif last_changed[asset_key]:
+            status[asset_key] = AssetComputeStatus.UNKNOWN
```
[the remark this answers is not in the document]
bengotow: Ahh you're right! It's just meant to track that the value still needs to be determined in this code and we can't early-exit. I like the idea of using an optional, will fix this!

python_modules/dagster-graphql/dagster_graphql/implementation/fetch_runs.py
```diff
@@ -122,21 +125,89 @@ def get_runs(graphene_info, filters, cursor=None, limit=None):
...
 
 
+class AssetComputeStatus(Enum):
+    OLD = "OLD"
+    GOOD = "GOOD"
+    NONE = "NONE"
+    UNKNOWN = "UNKNOWN"
+
+
+def get_all_asset_nodes(graphene_info):
```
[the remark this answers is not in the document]
bengotow: Ahh you're right @clairelin135 - I'll swap this out for that one! It looks like it returns GrapheneAssetNodes instead of just the external_asset_node objects, but I think that's ok? I'm not sure how much overhead there is in building them.

python_modules/dagster-graphql/dagster_graphql_test_results.xml
```diff
@@ -0,0 +1 @@
+<?xml version="1.0" encoding="utf-8"?><testsuites><testsuite name="pytest" errors="0" failures="0" skipped="0" tests="1" time="8.429" timestamp="2022-06-23T09:01:25.236867" hostname="Bens-M1-MBP.attlocal.net"><testcase classname="dagster_graphql_tests.graphql.test_assets.TestAssetAwareEventLog" name="test_latest_run_by_asset[sqlite_with_default_run_launcher_managed_grpc_env]" time="6.343" /></testsuite></testsuites>
```
prha, line 1: what's this?  should we exclude?

python_modules/dagster-graphql/dagster_graphql/implementation/fetch_runs.py
```diff
@@ -122,21 +128,93 @@ def get_runs(graphene_info, filters, cursor=None, limit=None):
...
+            return AssetComputeStatus.UP_TO_DATE  # partitioned assets are not supported
+        elif not last_changed.get(key):
+            return AssetComputeStatus.NONE  # never materialized assets cannot be stale
+
+        # mark us out-of-date if upstream nodes have newer materialization times
+        for dep in asset_node.dependencies:
+            dep_key = dep.upstream_asset_key
+            dep_node = all_asset_nodes[dep_key].external_asset_node
+            if (
+                last_changed.get(dep_key, 0) > last_changed.get(key)
```
prha, line 174: looks like mypy is complaining about this line... might need to do `> last_changed.get(key, 0)`

prha (approved):
I think this is looking pretty good.

clairelin135 (approved):
Looking great to me, excited about this!

python_modules/dagster-graphql/dagster_graphql/implementation/fetch_runs.py
bengotow: Thanks! I think I need to get my editor configured to do this typechecking so it's not caught by Buildkite 🙈

PR #8538, opened 2022-06-21
Add tests for dataframe annotations

dpeng817:
Don't think this was directly under test


sryza (commented):
Are dataframes special in some way as a type?

sryza (approved):
ok doesn't hurt - in general I don't think that we need to test that every type works with annotations

dpeng817:
Generally agreed. Probably overkill, but dataframes seem important enough.

PR #8542, opened 2022-06-22
Fallback to a placeholder event on a new event type that we don't recognize instead of raising an exception

gibsondan:
Summary:
This should make us a bit more fault-tolerant when introducing new events.


python_modules/dagster/dagster_tests/core_tests/test_event_logging.py
```diff
@@ -129,3 +137,43 @@ def simple():
         yes()
 
     return simple
+
+
+SERIALIZED_EVENT_FROM_THE_FUTURE_WITH_EVENT_SPECIFIC_DATA = '{"__class__": "DagsterEvent", "event_specific_data": {"__class__": "FutureEventData", "foo": null, "bar": null, "baz": null, "metadata_entries": [{"__class__": "EventMetadataEntry", "description": null, "entry_data": {"__class__": "TextMetadataEntryData", "text": "999"}, "label": "pid"}]}, "event_type_value": "EVENT_TYPE_FROM_THE_FUTURE", "logging_tags": {}, "message": "howdy", "pid": null, "pipeline_name": "nonce", "solid_handle": null, "step_handle": null, "step_key": "future_step", "step_kind_value": null}'
```
rexledesma, line 142: might be nice to include instructions on how you generated this (I assume copy/paste from ipython REPL)

rexledesma (approved):

Issue #8314, opened 2022-06-10
kubernetes executor: python3: argument list too long for complex dependency graphs


kervel:
## Summary

I have a re-execution of a job with a complex dynamic DAG (> 1K ops, of which i 600 ran during the first try)
i'm using the kubernetes executor to launch the steps. the kubernetes executor adds a "known_state" to the list of python arguments to launch the pod:

this "known_state" object seems to be huge (3K lines) and the pod fails with:
```
exec /usr/bin/python3: argument list too long
```

## Reproduction
any job with a large enough DAG will trigger this (at least in its re-execution)

## Dagit UI/UX Issue Screenshots

![image](https://user-images.githubusercontent.com/154455/173083984-747ecb71-a5b0-4782-a467-a8473a98c7fb.png)

![image](https://user-images.githubusercontent.com/154455/173083454-ee88b4aa-e784-44f5-896e-55cf8bdfc1a8.png)

## Additional Info about Your Environment

---
#### Message from the maintainers:
Impacted by this bug? Give it a 👍. We factor engagement into prioritization.

johannkm:
Apologies for the delay here, this should be fixed this week

johannkm:
Closed by https://github.com/dagster-io/dagster/pull/8564?no-redirect=1, will go out in the Thursday release

sinnfashen:
Hi team! I initially thought I met the same problem but now I'm not that sure. The situation we were facing is pretty similar: we had a job run with 600 dynamic output ops. We saw the error message to be `exec /opt/conda/envs/akuna/bin/dagster: argument list too long`. And none of those ops can run. (One difference here might be that we don't need any re-execution to hit the error). After seeing this thread we updated our dagster to the latest one (1.0.13). Still we see the same error while I can see that `DAGSTER_EXECUTE_STEP_ARGS` is properly set as an env in the container section. 

<img width="1042" alt="Screen Shot 2022-10-17 at 9 44 49 AM" src="https://user-images.githubusercontent.com/22710431/196207999-796660a3-405a-4a0b-94f5-042d0f67b301.png">

<img width="329" alt="Screen Shot 2022-10-17 at 9 44 12 AM" src="https://user-images.githubusercontent.com/22710431/196207834-02e3cc32-1918-4e6f-9049-4021351e8eb4.png">

sinnfashen:
[...]

PR #8564, opened 2022-06-22
k8s executor pass args via env var

johannkm:
Update:

changed to using env var for the full argument using the click feature. Not currently zipping- I think this buys us the time to make the proper change of filtering known state down to what's required.

---

For https://github.com/dagster-io/dagster/issues/8314

From https://elementl-workspace.slack.com/archives/C03D094F23G/p1654870219039639, could also
- gzip
- pass the whole input json as an env

Now that I type this out maybe that second option makes the most sense?

### DIFF
--- a/python_modules/dagster/dagster/_cli/api.py
+++ b/python_modules/dagster/dagster/_cli/api.py
@@ -315,7 +315,7 @@ def verify_step(instance, pipeline_run, retry_state, step_keys_to_execute):
         "interactively."
     ),
 )
-@click.argument("input_json", type=click.STRING)
+@click.argument("input_json", type=click.STRING, envvar="DAGSTER_EXECUTE_STEP_ARGS")
 def execute_step_command(input_json):
     with capture_interrupts():
 
--- a/python_modules/dagster/dagster/_grpc/types.py
+++ b/python_modules/dagster/dagster/_grpc/types.py
@@ -236,12 +236,23 @@ class ExecuteStepArgs(
             ),
         )
 
-    def get_command_args(self) -> List[str]:
-        return _get_entry_point(self.pipeline_origin) + [
-            "api",
-            "execute_step",
-            serialize_dagster_namedtuple(self),
-        ]
+    def get_command_args(self, skip_serialized_namedtuple=False) -> List[str]:
+        """
+        Get the command args to run this step. If skip_serialized_namedtuple is True, then get_command_env should
+        be used to pass the args to Click using an env var.
+        """
+        return (
+            _get_entry_point(self.pipeline_origin)
+            + ["api", "execute_step"]
+            + ([serialize_dagster_namedtuple(self)] if not skip_serialized_namedtuple else [])
+        )
+
+    def get_command_env(self) -> List[Dict[str, str]]:
+        """
+        Get the env vars for overriding the Click args of this step. Used in conjuction with
+        get_command_args(skip_serialized_namedtuple=True).
+        """
+        return [{"name": "DAGSTER_EXECUTE_STEP_ARGS", "value": serialize_dagster_namedtuple(self)}]
 
 
 @whitelist_for_serdes
--- a/python_modules/libraries/dagster-k8s/dagster_k8s/executor.py
+++ b/python_modules/libraries/dagster-k8s/dagster_k8s/executor.py
@@ -113,7 +113,7 @@ def k8s_job_executor(init_context: InitExecutorContext) -> Executor:
             load_incluster_config=run_launcher.load_incluster_config,
             kubeconfig_file=run_launcher.kubeconfig_file,
         ),
-        retries=RetryMode.from_config(init_context.executor_config["retries"]),  # type: ignore
+        retries=RetryMode.from_config(exc_cfg["retries"]),  # type: ignore
         max_concurrent=check.opt_int_elem(exc_cfg, "max_concurrent"),
         should_verify_step=True,
     )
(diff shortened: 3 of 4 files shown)


johannkm:
* master
  * **PR #8564** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8564" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8564?utm_source=stack-comment).

gibsondan (dismissed):
i think both of your bullet points are strong contenders if this is a size concern - gzip should eat this for breakfast since it has so much repetition

python_modules/dagster/dagster/cli/api.py
```diff
@@ -323,6 +326,12 @@ def execute_step_command(input_json):
```
gibsondan, line 326: if you do input_json there's actually an existing env var you can use? believe it's DAGSTER_CLI_API_EXECUTE_STEP_INPUT_JSON

(but that won't work if you gzip it so maybe not)

johannkm (commented):
This also won't be backcompat if the Daemon or user code are on different versions...
could possibly cook something up so that a new user code run could still accept it via CLI. not sure about compat in the other direction

johannkm (commented):
Unfortunately the env vars only exist by default for options, not arguments. However the backcompat concerns aren't as bad as I thought- I forgot this is all happening in run workers and step workers, so they'll both be the same version in the vast majority of cases.

Add an env var for the arg, and provide an option to opt out just in case this brakes for users who override the image for specific steps

python_modules/libraries/dagster-k8s/dagster_k8s/executor.py
```diff
@@ -43,6 +44,14 @@
                 description="Limit on the number of pods that will run concurrently within the scope "
                 "of a Dagster run. Note that this limit is per run, not global.",
             ),
+            "send_args_via_cli": Field(
```
johannkm, line 47: you might be able to convince me we don't need this... we don't test version differences between run worker and step worker, and I don't think we've ever claimed to support it?

python_modules/dagster/dagster/grpc/types.py
```diff
@@ -236,12 +236,24 @@ def __new__(
...
+        """
+        Get the command args to run this step. If with_args is False, then get_command_env should
+        be used to pass the args over env vars.
+        """
+        return (
+            _get_entry_point(self.pipeline_origin)
+            + ["api", "execute_step"]
+            + [serialize_dagster_namedtuple(self)]
+            if with_args
+            else []
```
alangenfeld, line 249: i find this hard to read what exactly the if is gating, maybe extra parens or be more explicit

alangenfeld (commented):
what are the length limitations for passing by env in k8s? 

looking at the screenshot in https://github.com/dagster-io/dagster/issues/8314 it appears theres a bug in whats getting put in the `dynamic_mappings` data structure

johannkm (commented):
1,048,576 characters with caveats for env vs 65,535 for CLI. 

> it appears theres a bug

because the result: null? I was assuming that was for steps that haven't finished yet but haven't really looked at it

alangenfeld (commented):
https://github.com/dagster-io/dagster/pull/8975?no-redirect=1

alangenfeld:
whats up with the test failures? I think my targeted bug fix may take the heat of rushing this in if we want to consider approaches without time pressure.

python_modules/libraries/dagster-k8s/dagster_k8s/executor.py
gibsondan: yeah i don't think supporting back-compat there is essential at this stage

gibsondan (changes requested):
yeah i don't think you need send_args_via_cli, otherwise this looks good

Milias:
Hello, I am also affected by this issue.

Is there any update on the pull request? :smile:

python_modules/dagster/dagster/_cli/api.py
```diff
@@ -315,7 +315,7 @@ def verify_step(instance, pipeline_run, retry_state, step_keys_to_execute):
         "interactively."
     ),
 )
-@click.argument("input_json", type=click.STRING)
+@click.argument("input_json", type=click.STRING, envvar="DAGSTER_EXECUTE_STEP_ARGS")
```
alangenfeld, line 318: im a little nervous about using a new manually set env var rather than the one implicitly provided by click - i guess we should never be out of version sync on the two sides of this exchange but still spooks me

python_modules/dagster/dagster/_grpc/types.py
```diff
@@ -236,12 +236,22 @@ def __new__(
             ),
         )
 
-    def get_command_args(self) -> List[str]:
-        return _get_entry_point(self.pipeline_origin) + [
-            "api",
-            "execute_step",
-            serialize_dagster_namedtuple(self),
-        ]
+    def get_command_args(self, with_args=True) -> List[str]:
```
alangenfeld, line 239: get args with args feels a bit obtuse - think a more verbose name would be nice 

would also consider not defaulting the arg and updating callsites

alangenfeld (approved):

python_modules/dagster/dagster/_cli/api.py
johannkm: Click only implicilty provides env for options, not args https://click.palletsprojects.com/en/7.x/arguments/#environment-variables

In terms of backcompat... yeah originally I made this opt in but it seemed rare enough that run and step workers are different images (certainly doesn't happen by default)

PR #8566, opened 2022-06-22
fix executor test

gibsondan:
Summary:
Not totally clear to me why this was passing before, but is passing now.


rexledesma (approved):

gibsondan:
ah it was doing an exact match on the field in the namedtuple i think, and i added a period

PR #8569, opened 2022-06-22
GraphQL Fix indexing on deleted run

clairelin135:
A rare error occurs when an asset run is deleted at the same time that Dagit attempts to load the latest run by asset:
https://dagster.slack.com/archives/C01U954MEER/p1655917412099299

I wasn't able to replicate the error (I think the timing has to be exactly right) but I think this fix will prevent this error from occurring again.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/implementation/fetch_runs.py
+++ b/python_modules/dagster-graphql/dagster_graphql/implementation/fetch_runs.py
@@ -170,7 +170,10 @@ def get_assets_live_info(graphene_info, step_keys_by_asset: Mapping[AssetKey, Li
             list(unstarted_run_ids_by_asset.get(asset_key, [])),
             list(in_progress_run_ids_by_asset.get(asset_key, [])),
             GrapheneRun(run_records_by_run_id[latest_run_ids_by_asset[asset_key]])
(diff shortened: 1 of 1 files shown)


python_modules/dagster-graphql/dagster_graphql/implementation/fetch_runs.py
```diff
@@ -170,7 +170,10 @@ def get_assets_live_info(graphene_info, step_keys_by_asset: Mapping[AssetKey, Li
             list(unstarted_run_ids_by_asset.get(asset_key, [])),
             list(in_progress_run_ids_by_asset.get(asset_key, [])),
             GrapheneRun(run_records_by_run_id[latest_run_ids_by_asset[asset_key]])
+            # Dagit error occurs if a run is terminated at the same time that this endpoint is called
```
prha, line 173: this doesn't quite make sense to me... 

run_records_by_run_id is calculated by fetching all the run_records based on the provided run_ids.  Why does run termination affect anything?

clairelin135: Ah, instead of terminated I meant deleted. The user stated that the run disappeared from the runs page, which implies the run was deleted.

prha (approved):

PR #8570, opened 2022-06-22
[dagit] Repair dialog ordering for confirmation/alert dialogs

### LINKED ISSUE #8423: Multiple config modals display for graph-backed assets
### Issue from the [Dagster Slack](https://join.slack.com/t/dagster/shared_invite/enQtNjEyNjkzNTA2OTkzLTI0MzdlNjU0ODVhZjQyOTMyMGM1ZDUwZDQ1YjJmYjI3YzExZGViMDI1ZDlkNTY5OThmYWVlOWM1MWVjN2I3NjU)

This issue was generated from the slack conversation at: https://dagster.slack.com/archives/C01U954MEER/p1655334943671729?thread_ts=1655334943.671729&cid=C01U954MEER

---

### Conversation excerpt

U02545XK221: For a graph-backed asset, are you able to see the visual representation for the underlying graph in dagit?
U028M11QNDD: Hi Dmitry, yes. There is a toggle button at the bottom labeled with "View as Asset Graph" and turning it off will show the underlying op/graph structure. You can then also toggle the "Explode graphs" to view all of the ops underlying a graph-backed asset:
U02545XK221: Thanks.  Just a piece of general feedback, dagit behaviour seems a little wonky in this release. For example, if I click on Materialise button for an asset that requires configuration, I get this window
U02545XK221: If I click on Scaffold Missing Config, the subsequent dialogue appears *under* this popup panel and is inaccessible until after I close it
U02545XK221: this doesn't happenevery time.
U02545XK221: but fairly often
U02545XK221: This is the dialogue that I don't see until after I close the previous window:
U028M11QNDD: Got it. Thanks for reporting this bug, I'll file an issue
U028M11QNDD: <@U018K0G2Y85> issue Multiple config modals display for graph-backed assets

---

Do you care about this too? Give it a :thumbsup:. We factor engagement into prioritization.

hellendag:
### Summary & Motivation

Resolves #8423.

The Dialog components used for `CustomConfirmationProvider` and `CustomAlertProvider` are reused singleton components that, once rendered initially, remain in the DOM forever.

The problem is that when one of these is opened from another Dialog after already having been rendered before (and therefore already existing in the DOM), the DOM order causes the confirmation or alert Dialog to sit behind the Dialog that opened it.

We could strap on our helmets and start some z-index wars, but I think I prefer this instead: don't leave the confirmation or alert Dialog sitting in the DOM once we're done with it. Unmount it after a delay, then remount it when a new confirmation/alert arrives. The delay allows the fade-out animation to be visible to the user before the unmount occurs.

Open to thoughts on this.

View asset config dialog that requires some scaffolding. Open scaffold Dialog (confirmation), then close both dialogs. Repeat, and verify that the confirmation dialog now properly sits on top of the config dialog.

I also added a jest test that includes testing the unmount behavior.

### DIFF
--- a/js_modules/dagit/packages/core/src/app/CustomAlertProvider.tsx
+++ b/js_modules/dagit/packages/core/src/app/CustomAlertProvider.tsx
@@ -23,7 +23,10 @@ export const showCustomAlert = (opts: Partial<ICustomAlert>) => {
   setCustomAlert(Object.assign({body: '', title: 'Error'}, opts));
 };
 
+const REMOVAL_DELAY = 500;
+
 export const CustomAlertProvider = () => {
+  const [mounted, setMounted] = React.useState(false);
   const [alert, setAlert] = React.useState(() => CurrentAlert);
   const body = React.useRef<HTMLDivElement>(null);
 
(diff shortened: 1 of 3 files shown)


hellendag:
* master
  * **PR #8570** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8570" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8570?utm_source=stack-comment).

salazarm (approved):

bengotow (approved):
This looks good to me!

PR #8580, opened 2022-06-23
fix repo in SDA pandas pyspark  guide

sryza:


sryza:
testing

sryza:
This won't be necessary after jamie's change to accept lists of assets on repos.

PR #8587, opened 2022-06-23
Update text in asset catalog filter

sryza:


bengotow (approved):
Yep all good! Not sure why that underscore was in there...

PR #8592, opened 2022-06-23
update hackernews tests to use asset invocation and materialize_to_memory

dpeng817:
title


examples/hacker_news_assets/hacker_news_assets_tests/test_resources/test_parquet_io_manager.py
```diff
@@ -26,21 +26,20 @@ def spark_input_asset(pandas_df_asset: SparkDF):
...
             },
+            partition_key="2022-01-01-16:00",
         )
 
         expected_path = os.path.join(
             temp_dir, "pandas_df_asset", "20220101160000_20220101170000.pq"
         )
-        res = io_manager_test_job.execute_in_process(partition_key="2022-01-01-16:00")
         assert res.success
         assert os.path.exists(expected_path)
```
sryza, line 44: Does this work? If we're invoking `materialize_to_memory` here, I would think that would either ignore the IO manager or error?

sryza (commented):
I think you missed test_core

PR #8594, opened 2022-06-23
Allow op def / resource def invocation when resources are required, but no used in body of op/resource.

dpeng817:
It's possible to specify an op definition or resource definition that requires a resource to be initialized, but doesn't explicitly use it within the body of the op. We don't handle this case gracefully when invoking, because we expect all ops/resources that have resource requirements to have a context argument. This fixes the case for both ops and resources.


prha:
Should we warn in this case?  They'll be incurring some sort of resource init time unnecessarily for this op...

dpeng817:
> Should we warn in this case? They'll be incurring some sort of resource init time unnecessarily for this op...
cc @prha 
Do you mean should we warn when someone creates an op that does this? I don't think so, there are valid reasons someone might have a resource without a context argument. IE step launchers, context manager resources.

sryza (approved):

PR #8613, opened 2022-06-24
[dagit] If you have no jobs, prefer routing to asset group over asset catalog

bengotow:
### Summary & Motivation

Our FallthroughRoot is getting fairly nuanced, but the updated ordering is this:

- If you have locations but no repos (most likely due to an error), show workspace.
- If you have no asset jobs or op jobs and have asset group(s), show the first asset group
- If you have exactly one repo with one job, show the job
- If you have more jobs/repos, show the instance factory floor.
- If you have no jobs, show docs on how to add one.
- If you have no repos, show docs on how to add one.

### DIFF
--- a/js_modules/dagit/packages/core/src/app/FallthroughRoot.tsx
+++ b/js_modules/dagit/packages/core/src/app/FallthroughRoot.tsx
@@ -1,10 +1,10 @@
-import {Box, ExternalAnchorButton, Colors, NonIdealState, Spinner} from '@dagster-io/ui';
+import {Box, Colors, ExternalAnchorButton, NonIdealState, Spinner} from '@dagster-io/ui';
 import * as React from 'react';
 import {Redirect, Route, Switch, useLocation} from 'react-router-dom';
 
 import {isHiddenAssetGroupJob} from '../asset-graph/Utils';
-import {WorkspaceContext} from '../workspace/WorkspaceContext';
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/core/src/app/FallthroughRoot.tsx
```diff
@@ -49,21 +52,28 @@ const FinalRedirectOrLoadingRoot = () => {
     return <Redirect to="/workspace" />;
   }
 
-  const reposWithAJob = allRepos.filter((r) => r.repository.pipelines.length > 0);
+  const reposWithVisibleJobs = allRepos.filter((r) => getVisibleJobs(r).length > 0);
```
hellendag, line 55: Could use something like this for the filter, since you're just checking `length > 0`:

```
const hasVisibleJobs = r.repository.pipelines.some((j) => !isHiddenAssetGroupJob(j.name));

// then
const reposWithVisibleJobs = allRepos.filter(hasVisibleJobs);
```

hellendag (approved):

js_modules/dagit/packages/core/src/app/FallthroughRoot.tsx
bengotow: Ahh this makes sense! let me see what I can do 🙌

PR #8621, opened 2022-06-26
use a custom yaml loader/dumper for loading and dumping run config, without surprising datetime parsing or octal string behavior

gibsondan:
Summary:
Switching dagit to send and receive data in YAML wasn't quite sufficient to handle weird edge cases that pyyaml handles strangely:
- datetime-like strings in config are parsed as datetimes, even though that's not a valid dagster config type
- Octal strings without any 8s or higher in them are bizarrely parsed as integers

This adds a custom yaml loader and dumper that we can customize to change behaviors like this. It digs a little bit into pyyaml internals to do so, but I haven't been able to find another way to customize the default tag behaviors of Pyyaml.


prha (approved):
This is a good find.

PR #8626, opened 2022-06-27
[dagit] Make "Add to filter" an append action

### LINKED ISSUE #8025: Add "Append tag to filter" hover action
``Copy tag`` and ``Add tag to filter`` actions have been added in https://github.com/dagster-io/dagster/pull/7751 (great feature, thanks a lot).

But ``Add tag to filter`` overwrites previously added tags.

How about adding ``Append tag to filter``?

hellendag:
### Summary & Motivation

Resolves #8025.

When adding a tag to the filter in a Runs view, make it an "append" action instead of a replacement action. This allows the user to add multiple tags through the tooltip action.

If the appended tag already exists in the list, no-op.

View Runs page. Append some tags, verify that they are appended and that the query string updates properly. Try to add a tag that's already there, verify no-op.

### DIFF
--- a/js_modules/dagit/packages/ui/src/components/TokenizingField.tsx
+++ b/js_modules/dagit/packages/ui/src/components/TokenizingField.tsx
@@ -92,8 +92,11 @@ export function tokenizedValueFromString(
   return {value: str};
 }
 
+export const tokenToString = (v: TokenizingFieldValue) =>
+  v.token ? `${v.token}:${v.value}` : v.value;
+
 export const tokensAsStringArray = (value: TokenizingFieldValue[]) =>
-  value.filter((v) => v.value !== '').map((v) => (v.token ? `${v.token}:${v.value}` : v.value));
+  value.filter((v) => v.value !== '').map(tokenToString);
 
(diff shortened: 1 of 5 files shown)


hellendag:
* master
  * **PR #8626** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8626" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8626?utm_source=stack-comment).

js_modules/dagit/packages/ui/src/components/TokenizingField.tsx
```diff
@@ -429,7 +432,6 @@ export const TokenizingField: React.FC<TokenizingFieldProps> = ({
           ) : undefined
         }
       />
-      <StyledTagInput />
```
hellendag, line 432: Not sure how long this has been sitting here, but it's not supposed to be here and it causes `Popover2` warnings in console. 😆

prha (approved):

PR #8637, opened 2022-06-27
force partition status to return values for missing partitions

prha:
### Summary & Motivation
'Show coverage' returns incorrect results.

Loaded partitioned job that has never run


rexledesma (approved):
adding a graphql test for partition sets would be nice

PR #8641, opened 2022-06-27
Explicitly disallow assets with resources in materialize_to_memory, use mem_io_manager as every io manager

dpeng817:
As the title. I think there are a few worlds that make sense for materialize_to_memory:
1. explicitly disallow any resource defs to be provided on assets that are passed in to materialize_to_memory
    - this makes testing assets more difficult, as `without_resources` will always have to be used whenever `resource_defs` are used.
2. explicitly disallow only io manager defs to be provided on assets that are passed in to materialize_to_memory
    - this makes the particular case of removing io manager defs more confusing, as your escape hatch is using `without_resources`, which will remove all resources from your asset, not just io manager defs. (maybe we want just a `without_io_managers`?)
 3. We just ignore whichever io managers are on your assets, and provide mem_io_manager. We still use your provided resources that aren't io managers.
      - slightly magical but also probably the most useful of the bunch.
 5. We ignore whichever io managers are on your assets and ignore the resources you provide (this one doesn't really make sense).

### DIFF
--- a/python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets.py
+++ b/python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets.py
@@ -18,6 +18,7 @@ from dagster import (
     resource,
 )
 from dagster._check import CheckError
+from dagster.core.test_utils import instance_for_test
 from dagster.core.asset_defs import AssetGroup, AssetIn, SourceAsset, asset, multi_asset
 from dagster.core.errors import DagsterInvalidDefinitionError, DagsterInvalidInvocationError
 from dagster.core.storage.mem_io_manager import InMemoryIOManager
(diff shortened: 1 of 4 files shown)


python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets.py
```diff
@@ -458,7 +459,9 @@ def my_asset(context):
         yield Output(1, "key1")
         yield Output(2, "key2")
 
-    materialize_to_memory([my_asset])
+    with instance_for_test() as instance:
+        result = materialize([my_asset], instance=instance)
+        assert result.success
```
dpeng817, line 464: this feels unfortunate to me, and is a result of us explicitly erroring if resource defs are directly provided.

PR #8642, opened 2022-06-28
Add a threadpool to the sensor daemon

prha:
### Summary & Motivation
Commandeered version of https://github.com/dagster-io/dagster/pull/8265

BK

### DIFF
--- a/python_modules/dagster/dagster/daemon/sensor.py
+++ b/python_modules/dagster/dagster/daemon/sensor.py
@@ -1,4 +1,5 @@
 import concurrent.futures
+import multiprocessing
 import os
 import sys
 import time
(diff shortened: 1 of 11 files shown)


python_modules/dagster/dagster/daemon/sensor.py
```diff
@@ -175,6 +176,46 @@ def _check_for_debug_crash(debug_crash_flags, key):
...
+    def shutdown(self, wait=True):
+        pass
+
+
+def sensor_executor(instance):
+    settings = instance.get_settings("sensors")
+
+    if settings.get("use_threads"):
+        return concurrent.futures.ThreadPoolExecutor(
+            max_workers=settings.get("num_workers"),
```
gibsondan, line 213: See https://github.com/dagster-io/internal/pull/76 for some context on how we picked the default value for the cloud agent (not clear if that's totally applicable here - plus in the future we might have multiple threadpools for different daemons, we'll have to sort that out)

gibsondan (approved):
thanks for pushing this forward - on to the run queue daemon!

PR #8643, opened 2022-06-28
Expose TimeWindow in dagster package

nvinhphuc:
### Summary & Motivation
Allow to import TimeWindow from dagster package for usability.

Manual testing


python_modules/dagster/dagster/core/definitions/__init__.py
```diff
@@ -140,6 +140,7 @@
 from .solid_container import create_execution_structure
 from .solid_definition import CompositeSolidDefinition, NodeDefinition, SolidDefinition
 from .time_window_partitions import (
+    TimeWindow,
```
sryza, line 143: Hey @nvinhphuc - it looks like the import ordering is off. You can fix this by running "make isort", or by moving this to its alphabetical position.

nvinhphuc: Hi @sryza , I sorted the import, please check again if anything is wrong.

sryza: Looks like that fixed it.

PR #8646, opened 2022-06-28
[dagster-io/ui] Repair ConfigEditor typeahead for nullables

### LINKED ISSUE #8476: List typeaheads not working in yaml editor
## Summary

I can't get the typeahead to fire with list of key value pairs

See example here from the add code location dialog

https://user-images.githubusercontent.com/28738937/174449362-a54463a5-c23d-45bc-aee4-f7b158909908.mov

It passes config validation but the typeahead (according to the UI) but the typeahead is not firing.

---
Impacted by this bug? Give it a 👍. We factor engagement into prioritization.

hellendag:
### Summary & Motivation

Resolves #8476.

The ConfigEditor yaml mode typeahead does not currently appear to regard the `NullableConfigType` as a value that it can handle. As a result, no typeahead suggestions are offered when the cursor is within a nullable type, e.g. a nullable array.

To resolve this, determine whether we are within a `NullableConfigType`, and if so, extract its inner type so that we can appropriately determine suggestions for it. Additionally, determine whether the nullable value is a composite/list/map type, to properly add indentation.

View code location config editor in Cloud. Repeat the steps taken in the original issue, verify that `name` is shown as a typeahead option, and that indentation and typeahead insertion occur properly.

### DIFF
--- a/js_modules/dagit/packages/ui/src/components/configeditor/codemirror-yaml/mode.tsx
+++ b/js_modules/dagit/packages/ui/src/components/configeditor/codemirror-yaml/mode.tsx
@@ -400,10 +400,20 @@ CodeMirror.registerHelper(
       }
       // Using a lookup table here seems like a good idea
       // https://github.com/dagster-io/dagster/issues/1966
-      const type = options.schema.allConfigTypes.find((t) => t.key === key);
+      let type = options.schema.allConfigTypes.find((t) => t.key === key);
       if (!type) {
         return false;
       }
+
+      // If nullable, extract the inner type.
+      if (type.__typename === 'NullableConfigType') {
+        const innerKey = type.typeParamKeys[0];
+        type = options.schema.allConfigTypes.find((t) => t.key === innerKey);
+        if (!type) {
+          return false;
+        }
+      }
+
       return (
         type.__typename === 'ArrayConfigType' ||
         type.__typename === 'CompositeConfigType' ||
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/ui/src/components/configeditor/codemirror-yaml/mode.tsx
```diff
@@ -609,7 +619,14 @@ function findAutocompletionContext(
...
 
-      const parentConfigType = schema.allConfigTypes.find((t) => t.key === typeKey)!;
+      let parentConfigType = schema.allConfigTypes.find((t) => t.key === typeKey)!;
+
+      // If nullable, extract the inner type.
+      if (parentConfigType.__typename === 'NullableConfigType') {
+        const innerType = parentConfigType.typeParamKeys[0];
+        parentConfigType = schema.allConfigTypes.find((t) => t.key === innerType)!;
+      }
+
```
bengotow, line 629: This seems like a safe change to me! Before we merge though, I'm curious if this blows up downstream if `parentConfigType!` is actually null. It seems like it was unsafe before and it's unsafe now, so probably fine 😅

bengotow (approved):

js_modules/dagit/packages/ui/src/components/configeditor/codemirror-yaml/mode.tsx
hellendag: It seems possible, but yeah, it's already unsafe so I think this shouldn't cause new problems. I think if it does we'll spot it pretty quickly.

PR #8649, opened 2022-06-28
Support nested lists in repository def

### LINKED ISSUE #8563: Accept list of AssetsDefinitions directly on repo
I've seen a few users get tripped up on this (including myself)

Right now you can do this:
```
assets: Sequence[AssetsDefinition, SourceAsset] = ...

@repository
def repo():
    return [*assets]
```

but this fails:

```
assets: Sequence[AssetsDefinition, SourceAsset] = ...

@repository
def repo():
    return [assets]
```

jamiedemaria:
### Summary & Motivation

Started just adding support for lists of assets/source assets in repository, but then remembered this handy function. so now it supports flattening any list of definitions

resolves #8563

bk + new unit tests


python_modules/dagster/dagster/core/definitions/decorators/repository_decorator.py
```diff
@@ -19,6 +20,30 @@
 from ..unresolved_asset_job_definition import UnresolvedAssetJobDefinition
 
 
+def _flatten(items):
+    from dagster.core.asset_defs import AssetGroup, AssetsDefinition, SourceAsset
+
+    BaseTypes = (
```
sryza, line 26: It might be cumbersome to remember to include new definitions here when they're added. Thoughts on doing it the other way around? I.e. checking to see if the item is _not_ a list?

sryza (commented):
For the sake of being conservative about how much API surface area we expose, I think for now we should just allow a single level of nesting. I could be convinced otherwise though.

python_modules/dagster/dagster/core/definitions/decorators/repository_decorator.py
jamiedemaria: yeah, could be more strict about it and not allow tuples or whatever else people come up with

sryza:
> @sryza can you give an example of how multiple layers of nesting exposes more API? i'm not sure i'm following

I mean it just in the general sense that it allows more freedom in how to do things, which we'll need to support in the future.

PR #8656, opened 2022-06-28
switch unloadable states to dedupe by selector id

prha:
### Summary & Motivation
If you switched pyenv such that the origin changed and the selector did not, you would see your schedules/sensors duplicated in the unloadable section.

Ran bk, loaded sensors / schedules in different virtualenv, saw empty unloadable schedules


prha:
The origin changes when the hash of the repository location changes, but that includes things like python environment, etc.  

The selector just consists of the location name / repository name / instigator name

PR #8657, opened 2022-06-28
add helm values for configuring instance sensor config

prha:
### Summary & Motivation
We now let users configure the sensor daemon to evaluate using a thread pool

BK, inspection

### DIFF
--- a/helm/dagster/schema/schema/charts/dagster/subschema/daemon.py
+++ b/helm/dagster/schema/schema/charts/dagster/subschema/daemon.py
@@ -61,7 +61,7 @@ class RunCoordinator(BaseModel):
 
 
 class Sensors(BaseModel):
-    useThreads: Optional[bool]
+    useThreads: bool
     numWorkers: Optional[int]
 
 
@@ -87,7 +87,7 @@ class Daemon(BaseModel):
     annotations: kubernetes.Annotations
     runMonitoring: Dict[str, Any]
     runRetries: Dict[str, Any]
-    sensors: Optional[Sensors]
+    sensors: Sensors
 
     class Config:
         extra = Extra.forbid
--- a/helm/dagster/schema/schema/charts/dagster/subschema/daemon.py
+++ b/helm/dagster/schema/schema/charts/dagster/subschema/daemon.py
@@ -59,9 +59,11 @@ class RunCoordinator(BaseModel):
             )
         }
 
+
 class Sensors(BaseModel):
     useThreads: Optional[bool]
     numWorkers: Optional[int]
+
 
 class Daemon(BaseModel):
     enabled: bool
--- a/helm/dagster/schema/schema/charts/dagster/subschema/daemon.py
+++ b/helm/dagster/schema/schema/charts/dagster/subschema/daemon.py
@@ -59,6 +59,9 @@ class RunCoordinator(BaseModel):
             )
         }
 
+class Sensors(BaseModel):
+    useThreads: Optional[bool]
+    numWorkers: Optional[int]
 
 class Daemon(BaseModel):
     enabled: bool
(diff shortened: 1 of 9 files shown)


helm/dagster/values.yaml
```diff
@@ -1003,6 +1003,13 @@ dagsterDaemon:
     enabled: false
     maxRetries: 0
 
+  sensors:
+    # Whether to evaluate sensors using an asynchronous thread pool.  Defaults to false
+    use_threads: false
+    # The max number of worker threads to use when asynchronously evaluating sensors. Will use the
+    # default value used by Python, which depends on the number of cores available.
+    num_workers: ~
```
rexledesma, line 1011: camel case

helm/dagster/schema/schema_tests/test_dagster_daemon.py
```diff
@@ -247,3 +248,20 @@ def test_daemon_labels(template: HelmTemplate):
 
     assert set(deployment_labels.items()).issubset(daemon_deployment.metadata.labels.items())
     assert set(pod_labels.items()).issubset(daemon_deployment.spec.template.metadata.labels.items())
+
+
+def test_sensor_threading(instance_template: HelmTemplate):
```
rexledesma, line 253: Can also test if the configuration has all the fields in the configurable class, similar to:

https://github.com/dagster-io/dagster/blob/b1aa9599f40b8655fc389996f42018ce7ac75e5a/helm/dagster/schema/schema_tests/test_instance.py#L471-L472

helm/dagster/schema/schema/charts/dagster/subschema/daemon.py
```diff
@@ -82,6 +87,7 @@ class Daemon(BaseModel):
     annotations: kubernetes.Annotations
     runMonitoring: Dict[str, Any]
     runRetries: Dict[str, Any]
+    sensors: Optional[Sensors]
```
rexledesma, line 90: Technically it's not optional, as it has a default specified in `values.yaml`. 

```suggestion
    sensors: Sensors
```

rexledesma (approved):
couple of small typing/style changes, looks good otherwise. can also add an assertion on the configurable class fields.

helm/dagster/schema/schema/charts/dagster/subschema/daemon.py
prha: Well, I want to avoid the backcompat breakage (old user code against new helm upgrade), so should I remove the settings in values.yaml?

helm/dagster/templates/configmap-instance.yaml
```diff
@@ -105,5 +105,16 @@ data:
...
+    {{- if and (.Values.dagsterDaemon.enabled) (.Values.dagsterDaemon.sensors) }}
+    {{- $sensors := .Values.dagsterDaemon.sensors }}
+    {{- if $sensors.useThreads }}
+    sensors:
+      use_threads: {{ $sensors.useThreads }}
+      {{- if $sensors.numWorkers }}
+      num_workers: {{ $sensors.numWorkers }}
+      {{- end }}
+    {{- end }}
+    {{- end }}
```
rexledesma, line 117: We can simplify this because we know `.Values.dagsterDaemon.sensors` will always be defined, because of our schema.

```suggestion
    {{- $sensors := .Values.dagsterDaemon.sensors }}
    {{- if and (.Values.dagsterDaemon.enabled) ($sensors.useThreads) }}
    sensors:
      use_threads: {{ $sensors.useThreads }}
      {{- if $sensors.numWorkers }}
      num_workers: {{ $sensors.numWorkers }}
      {{- end }}
    {{- end }}
```

PR #8666, opened 2022-06-29
fix(dagster-dbt): add check for dbt-core before executing CLI commands

rexledesma:
### Summary & Motivation
Add a check that `dbt-core` is installed before running CLI commands.

Alternatively, we can specify `dbt-core` and `dbt-rpc` as dependencies on the python package, although they are not explicitly imported for their Python library.

Do both. In the rare case that somehow `dbt-core` is uninstalled while `dagster-dbt` is, we can still receive a helpful error message.

pytest


rexledesma:
* master
  * **PR #8666** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8666" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8666?utm_source=stack-comment).

PR #8668, opened 2022-06-29
hold shift to force asset config modal

### LINKED ISSUE #8524: Make it possible to specify config when materializing assets with optional config
And when resources that the assets depend on have optional config

https://dagster.slack.com/archives/C01U954MEER/p1655801664267859

smackesey:
### Summary & Motivation

Changes behavior of the asset materialize button so that, when you hold "shift", the asset config modal always pops up. This change applies everywhere the asset materialize button appears (graph explorer, asset details page, etc). Also adds a tooltip explaining this behavior.

Note that "shift" was chosen over alt/ctrl because: ctrl-click is already taken by the browser (brings up context menu, like right-click), and both alt and ctrl bring up shortcuts. But I'm definitely open to reconsidering the modifier. 

Note also that ultimately we want to add an ellipsis to the button text when holding down the modifier, but that's not in this  PR because it's a bit tricky and we want this in tmrw's release.

Resolves #8524.

Manually in dagit.

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
+++ b/js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
@@ -115,7 +115,7 @@ export const LaunchAssetExecutionButton: React.FC<{
 
   return (
     <>
-      <Tooltip content="shift-click to open the config editor">
+      <Tooltip content="Shift+click to add configuration">
         <Button
           intent={intent}
           onClick={onClick}
--- a/js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
+++ b/js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
@@ -96,7 +96,8 @@ export const LaunchAssetExecutionButton: React.FC<{
       variables: {assetKeys: assetKeys.map(({path}) => ({path}))},
     });
     const assets = result.data.assetNodes;
-    const next = stateForLaunchingAssets(assets, e, preferredJobName);
+    const forceLaunchpad = e.shiftKey;
+    const next = stateForLaunchingAssets(assets, forceLaunchpad, preferredJobName);
 
     if (next.type === 'error') {
       showCustomAlert({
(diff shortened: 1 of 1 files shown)


smackesey:
* master
  * **PR #8668** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8668" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8668?utm_source=stack-comment).

js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
```diff
@@ -152,6 +154,7 @@ export const LaunchAssetExecutionButton: React.FC<{
 
 function stateForLaunchingAssets(
   assets: LaunchAssetExecutionAssetNodeFragment[],
+  mouseEvent: React.MouseEvent<any>,
```
hellendag, line 157: I'd avoid passing the mouse event any further than the immediate handler function, and instead use a `forceLaunchpad` argument here or something similar based on the `shiftKey` value determined in the handler.

js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
```diff
@@ -114,19 +114,21 @@ export const LaunchAssetExecutionButton: React.FC<{
...
-          state.type === 'loading' ? (
-            <Spinner purpose="body-text" />
-          ) : (
-            <Icon name="materialization" />
-          )
-        }
-      >
-        {label}
-      </Button>
+      <Tooltip content="shift-click to open the config editor">
```
hellendag, line 117: `Shift+click to add configuration`? I'm not sure we refer to it as the "config editor" anywhere in the product, or just in code.

js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
smackesey: Good call, done

PR #8670, opened 2022-06-29
Option to skip daemon heartbeats with no errors

johannkm:
Disabled by default

Option to only send heartbeats when they contain errors. Useful in cloud where we don't need to rely on them to know if the daemons are alive. See https://github.com/dagster-io/internal/pull/2450?no-redirect=1 for more context

Under test in https://app.graphite.dev/github/pr/dagster-io/internal/2450

### DIFF
--- a/python_modules/dagster/dagster/daemon/controller.py
+++ b/python_modules/dagster/dagster/daemon/controller.py
@@ -253,11 +253,14 @@ class DagsterDaemonController:
         start_time = time.time()
         last_heartbeat_check_time = start_time
         while True:
-            # Wait until a daemon has been unhealthy for a long period of time
-            # before potentially restarting it due to a hanging or failed daemon
             with raise_interrupts_as(KeyboardInterrupt):
                 time.sleep(THREAD_CHECK_INTERVAL)
                 self.check_daemon_threads()
+
+                if self._instance.daemon_skip_heartbeats_without_errors:
+                    # If we're skipping heartbeats without errors, we just check the threads.
+                    # If there's no errors, the daemons won't be writing heartbeats.
+                    continue
 
                 now = time.time()
                 # Give the daemon enough time to send an initial heartbeat before checking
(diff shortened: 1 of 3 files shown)


johannkm:
* master
  * **PR #8670** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8670" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8670?utm_source=stack-comment).

python_modules/dagster/dagster/daemon/controller.py
```diff
@@ -249,19 +249,16 @@ def check_daemon_heartbeats(self):
...
-        last_heartbeat_check_time = start_time if check_heartbeats else None
+        last_heartbeat_check_time = start_time
         while True:
             # Wait until a daemon has been unhealthy for a long period of time
             # before potentially restarting it due to a hanging or failed daemon
             with raise_interrupts_as(KeyboardInterrupt):
                 time.sleep(THREAD_CHECK_INTERVAL)
                 self.check_daemon_threads()
 
-                if not check_heartbeats:
```
prha, line 262: will we not start raising exceptions due to the healthy heartbeat check?

alangenfeld (commented):
how do we catch stalled threads in this model? 

i'm a little weary managing such a fundamental behavior difference with a bool we poke at from random places

johannkm (commented):
Good callout. My take is similar to the justification of removing the livenessProbes from the helm chart: I don't think have evidence of stalled threads, vs. I think there's been a few cases of the daemon getting in a crashloop unnecessarily due to the heartbeat check (https://elementl.quip.com/69TdA05tpocT/Incident-Sensor-Daemon-Heartbeats isn't the best example, but the recent one)

In terms of toggling it- agreed that these instance injections never feel great, but I definitely think I can justify forking the behavior the cloud, where we have K8s and Datadog watching the daemon

prha (approved):

PR #8677, opened 2022-06-29
Load asset launchpad for assets with configurable resources

### LINKED ISSUE #8469: Config on asset job doesn't populate in materialize launchpad modal
### Issue from the [Dagster Slack](https://join.slack.com/t/dagster/shared_invite/enQtNjEyNjkzNTA2OTkzLTI0MzdlNjU0ODVhZjQyOTMyMGM1ZDUwZDQ1YjJmYjI3YzExZGViMDI1ZDlkNTY5OThmYWVlOWM1MWVjN2I3NjU)

This issue was generated from the slack conversation at: https://dagster.slack.com/archives/C01U954MEER/p1655490666755289?thread_ts=1655490666.755289&cid=C01U954MEER

---

### Conversation excerpt

U02QUNEGP0V: Hey dagster team - I'm trying the following code after upgrading to 0.15.0:
```# Run config no longer works from passing it in. Must use launch pad
run_config_path = dagster.file_relative_path(
    __file__, os.path.join(".", "run_config", "vocational_assets_job.yaml")
)
config = dagster.config_from_files([run_config_path])
vocational_assets_job = dagster.build_assets_job(
    assets=[*extract_assets, *load_assets, *transform_assets],
    name="vocational_assets_job",
    config=config,
)```
However, the config doesn't work and it still requires me to pass in via launchpad. Whats the best way to set a default config?
U02QUNEGP0V: Just to clarify using build_assets_job worked fine before on 0.14.3 and i'm passing it in the same way ~
U028M11QNDD: Hi Charles. In 0.15.0, we've switched to defining asset jobs via `define_asset_job` instead of `build_assets_job` in order to allow asset definitions to live directly on repositories. You can rewrite your job definition as follows:
```vocational_assets_job = define_asset_job("vocational_assets_job", config=config).resolve(
    [my_asset], source_assets=[]
)```
U02QUNEGP0V: Ah thank you! i'll give that a shot
U028M11QNDD: If you add your asset job to a repository, you can directly just write
```vocational_assets_job = define_asset_job("vocational_assets_job", config=config)```
and dagster will make the resolve call for you
U02QUNEGP0V: :smile: appreciate the fast turnaround
U02QUNEGP0V: Hey <@U028M11QNDD>! I gave it a shot but unfortunately i'm still getting the same error when i click materialize:
```vocational_assets_job cannot be executed with the provided config. Please fix the following errors:

Missing required config entry "resources" at the root. Sample config for missing entry: {'resources': {'bigquery': {'config': {'dataset': '...'}}, 'gcs': {'config': {'gcs_bucket': '...'}}}}```
Even though it's configured as you listed - config is a dictionary like this:
```{'resources': {'gcs': {'config': {'gcs_bucket': 'my-bucket', 'gcs_prefix': 'my-prefix'}}, 'bigquery': {'config': {'project': {'env': 'GCP_PROJECT'}, 'dataset': 'my-dataset'}}}}```

U028M11QNDD: Oh I see, you're seeing this error from the materialize button. It seems there is a bug in passing predefined config on your job to display in Dagit in the launchpad modal. I'll file an issue, for now you you can execute from the launchpad (and this will prepopulate the config from your yaml correctly).
U02QUNEGP0V: Thank you so much!
U028M11QNDD: <@U018K0G2Y85> issue Config on asset job doesn't populate in materialize launchpad modal

---

Do you care about this too? Give it a :thumbsup:. We factor engagement into prioritization.

### LINKED ISSUE #8629: Kicking off runs via materialize button gives no way to set resource config
The resources provided to assets may require config, but when kicking off runs via materialize, there is no way to edit the config that the materialization run will kick off with.

### PR DESCRIPTION
**Pull request by smackesey**
### Summary & Motivation

Makes sure the asset launchpad loads when materializing a set of assets where one or more has configurable resources. Fixes #8469. Fixes #8629.

This required modifying the graphql representation of asset nodes with an added `required_resources` resolver. Along the way I also made some tweaks to the way info off the backing op/graph is pulled in general (continuation of work done for adding asset config). @prha for graphql review

On the frontend, the current behavior is just to get the list of required resources, and if _any_ are present, to load the asset config modal. That is, I don't presently check if the resources require config (which would require another query), which is undesirable. @bengotow do you know if there's an existing query I can use for this purpose?

Modified `graphs_jobs_ops_toys/resources.py` and ensured that the launchpad loads for the sample asset (which has a resource) in that repo.

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
+++ b/js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
@@ -208,9 +208,13 @@ function stateForLaunchingAssets(
     };
   }
 
+  const requiredResources = assets.flatMap((a) => a.requiredResources.map((r) => r.resourceKey));
+  // temporary until a cleaner way to query the config requirements of resources is available
+  const anyResourcesHaveConfig = requiredResources.length >= 1;
+
   // Ok! Assertions met, how do we launch this run
 
-  if (anyAssetsHaveConfig || forceLaunchpad) {
+  if (anyAssetsHaveConfig || anyResourcesHaveConfig || forceLaunchpad) {
     const assetOpNames = assets.flatMap((a) => a.opNames || []);
     return {
       type: 'launchpad',
@@ -293,6 +297,9 @@ export const LAUNCH_ASSET_EXECUTION_ASSET_NODE_FRAGMENT = gql`
         id
         name
       }
+    }
+    requiredResources {
+      resourceKey
     }
     ...AssetNodeConfigFragment
   }
--- a/python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, List, Optional, Sequence, Union
+from typing import TYPE_CHECKING, List, Optional, Sequence, Union, cast
 
 import graphene
 from dagster_graphql.implementation.events import iterate_metadata_entries
@@ -6,6 +6,7 @@ from dagster_graphql.schema.config_types import GrapheneConfigTypeField
 from dagster_graphql.schema.metadata import GrapheneMetadataEntry
 from dagster_graphql.schema.solids import (
     GrapheneCompositeSolidDefinition,
+    GrapheneResourceRequirement,
     GrapheneSolidDefinition,
     build_solid_definition,
 )
@@ -13,11 +14,13 @@ from dagster_graphql.schema.solids import (
 from dagster import AssetKey
 from dagster import _check as check
 from dagster.core.host_representation import ExternalRepository, RepositoryLocation
+from dagster.core.host_representation.external import ExternalPipeline
 from dagster.core.host_representation.external_data import (
     ExternalAssetNode,
     ExternalStaticPartitionsDefinitionData,
     ExternalTimeWindowPartitionsDefinitionData,
 )
+from dagster.core.snap.solid import CompositeSolidDefSnap, SolidDefSnap
 
 from ..implementation.fetch_runs import AssetComputeStatus
 from ..implementation.loader import BatchMaterializationLoader, CrossRepoAssetDependedByLoader
@@ -91,6 +94,13 @@ class GrapheneAssetLatestInfo(graphene.ObjectType):
 
 class GrapheneAssetNode(graphene.ObjectType):
 
+    _depended_by_loader: Optional[CrossRepoAssetDependedByLoader]
+    _external_asset_node: ExternalAssetNode
+    _node_definition_snap: Optional[Union[CompositeSolidDefSnap, SolidDefSnap]]
+    _external_pipeline: Optional[ExternalPipeline]
+    _external_repository: ExternalRepository
+    _latest_materialization_loader: Optional[BatchMaterializationLoader]
+
     # NOTE: properties/resolvers are listed alphabetically
     assetKey = graphene.NonNull(GrapheneAssetKey)
     assetMaterializations = graphene.Field(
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by smackesey** [comment]
* master
  * **PR #8668** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8668" target="_blank">Graphite</a> 
    * **PR #8677** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8677" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8677?utm_source=stack-comment).

File: python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py [hunk 1]
```diff
@@ -199,6 +222,32 @@ def get_partition_keys(self) -> Sequence[str]:
...
+        self, node_def_snap: Union[CompositeSolidDefSnap, SolidDefSnap]
+    ) -> Sequence[str]:
+        if isinstance(node_def_snap, CompositeSolidDefSnap):
+            constituent_node_names = [
+                inv.solid_def_name
+                for inv in node_def_snap.dep_structure_snapshot.solid_invocation_snaps
+            ]
+            external_pipeline = self.get_external_pipeline()
+            constituent_resource_key_sets = [
+                self.get_required_resources(external_pipeline.get_node_def_snap(name))
```
bengotow, line 241: The Dagit updates look good to me! Will defer to @prha on the GraphQL changes.

I think one solution to the "does this required resource require config" challenge might be to expose more than the resource key here? If we were to have `GrapheneResourceRequirement(resource)` here instead of just the key set, we might be able to pull other fields off the resource requirement and get it's required info? It looks like this is calling through into code outside this PR though, so that may not be possible.

I think you're right that without that, we'd need two a follow-up query for more information about the resources.

python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
```diff
@@ -425,6 +485,11 @@ def resolve_repository(self, graphene_info) -> "GrapheneRepository":
             graphene_info.context.instance, self._external_repository, self._repository_location
         )
 
+    def resolve_required_resources(self, _graphene_info) -> Sequence[GrapheneResourceRequirement]:
+        node_def_snap = self.get_node_definition_snap()
```
bengotow, line 489: One more possible solution - we could extend this resolver to fetch the resources and return them rather than just the `all_unique_keys`. It might be somewhat slow, but if Dagit is going to turn around and make a second graphQL query for the resources we may as well do it on the backend in one request?

js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
```diff
@@ -294,6 +298,9 @@ export const LAUNCH_ASSET_EXECUTION_ASSET_NODE_FRAGMENT = gql`
         name
       }
     }
+    requiredResources {
+      resourceKey
+    }
```
bengotow, line 303: Thanks for putting this in here! Nice to defer fetching this info until you click the button 🙌

bengotow (approved):

smackesey:
@bengotow

>I think one solution to the "does this required resource require config" challenge might be to expose more than the resource key here? If we were to have GrapheneResourceRequirement(resource) here instead of just the key set, we might be able to pull other fields off the resource requirement and get it's required info?

Unfortunately `GrapheneResourceRequirement` only has a single field `resourceKey`, so there's no other information.

>One more possible solution - we could extend this resolver to fetch the resources and return them rather than just the all_unique_keys. It might be somewhat slow, but if Dagit is going to turn around and make a second graphQL query for the resources we may as well do it on the backend in one request?

I'm looking into doing this.

prha (approved):
Looks good... the memoization across resolvers is a great bonus.

lilinxi:
I am using the 1.0.12 dagit but it seems that I meet this issue again, when I trigger the `Materialize all` button, it tells me that there is a resource config miss but do not have place to add them.

PR #8682, opened 2022-06-29
[assets] Fix issue with graph backed assets + partitions

OwenKephart:
### Summary & Motivation

report: https://dagster.slack.com/archives/C01U954MEER/p1656497541526939

This issue arises because we use the `asset_keys_for_node()` function to determine which asset keys (and therefore which AssetsDefinition object) maps to a the node we are currently executing. With graph-backed assets, it may be the case that the node we're currently executing does not produce any assets itself (instead, a downstream node may produce the asset).

This PR makes it so that any op on the critical path of the creation of a given asset is mapped to that asset in the asset layer.

More simply, if you have a graph backed asset `some_asset`:

graph: (op1 -> op2)

And op2 creates the asset (op1 is just a helper), then `asset_keys_for_node(op1)` will return `some_asset` instead of an empty set.

Unit test.

### DIFF
--- a/python_modules/dagster/dagster/core/definitions/asset_layer.py
+++ b/python_modules/dagster/dagster/core/definitions/asset_layer.py
@@ -446,12 +446,13 @@ class AssetLayer:
 
         # keep an index from node handle to all keys expected to be generated using that node
         self._asset_keys_by_node_handle: Dict[NodeHandle, Set[AssetKey]] = defaultdict(set)
-        _asset_info_by_key = {
-            asset_info.key: asset_info
+        _required_asset_keys = {
+            asset_info.key
             for asset_info in self._asset_info_by_node_output_handle.values()
+            if asset_info.is_required
         }
         for asset_key, node_handles in self._dependency_node_handles_by_asset_key.items():
-            if _asset_info_by_key[asset_key].is_required:
+            if asset_key in _required_asset_keys:
                 for node_handle in node_handles:
                     self._asset_keys_by_node_handle[node_handle].add(asset_key)
 
--- a/python_modules/dagster/dagster/core/definitions/asset_layer.py
+++ b/python_modules/dagster/dagster/core/definitions/asset_layer.py
@@ -444,11 +444,16 @@ class AssetLayer:
             for key in assets_def.keys
         }
 
-        # keep an index from node handle to all keys expected to be generated in that node
+        # keep an index from node handle to all keys expected to be generated using that node
         self._asset_keys_by_node_handle: Dict[NodeHandle, Set[AssetKey]] = defaultdict(set)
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/core/definitions/asset_layer.py
```diff
@@ -444,11 +444,17 @@ def __init__(
             for key in assets_def.keys
         }
 
-        # keep an index from node handle to all keys expected to be generated in that node
+        # keep an index from node handle to all keys expected to be generated using that node
         self._asset_keys_by_node_handle: Dict[NodeHandle, Set[AssetKey]] = defaultdict(set)
-        for node_output_handle, asset_info in self._asset_info_by_node_output_handle.items():
-            if asset_info.is_required:
-                self._asset_keys_by_node_handle[node_output_handle.node_handle].add(asset_info.key)
+        _required_asset_keys = {
```
sryza, line 450: why use an underscore here?

python_modules/dagster/dagster/core/definitions/asset_layer.py
```diff
@@ -294,10 +294,11 @@ def _asset_key_to_dep_node_handles(
                 )
 
     # handle internal_asset_deps
-    for node_handle, assets_defs in assets_defs_by_node_handle.items():
-        all_output_asset_keys = assets_defs.keys
-        for asset_key, dep_asset_keys in assets_defs.asset_deps.items():
-            for dep_asset_key in [key for key in dep_asset_keys if key in all_output_asset_keys]:
+    for node_handle, assets_def in assets_defs_by_node_handle.items():
+        for asset_key, dep_asset_keys in assets_def.asset_deps.items():
```
OwenKephart, line 298: what's going on here is that this change uncovered an unrelated bug with how we were calculating these dependency keys after doing the cycle resolution stuff.

basically, we were using the internal_asset_deps as the source of truth for which assets were produced by which node, rather that assets_def.keys.

sryza (commented):
I think this makes the contents of `asset_keys_for_node` a little counterintuitive. Or at least I don't know if I would expect this behavior from looking at the variable name.  I wonder if we could sidestep this confusion by somehow eliminating `asset_keys_for_node` and instead exposing a method called `assets_def_for_node`?  I think the behavior of that method would be unambiguous.

OwenKephart:
@sryza yeah you're right it's confusing and we only ever use this function to lookup AssetsDefinitions, so cutting out the middleman makes sense, just replaced the function and it feels a bit better

At some point we should do an actual audit of all these random indexes we're creating over these definition objects and see what we can do to simplify this, but for now this works alright.

sryza (approved):
Nice nice

Issue #8603, opened 2022-06-24
dagster_databricks - support aws_attributes


Phazure:
## Use Case

Often for ops which leverage the `databricks_pyspark_step_launcher`, we execute the op logic within a new databricks job cluster. We would need the job cluster to have specific AWS permissions, and one way to achieve this is by using the `aws_attributes/instance_profile_arn` field as referenced in this [docs page](https://docs.databricks.com/dev-tools/api/latest/clusters.html).

![image](https://user-images.githubusercontent.com/17210429/175577764-f5392cf5-e04b-403f-8102-0673b27307e2.png)

It would be great if we could also pass in `aws_attributes` to our new cluster config when defining the `databricks_pyspark_step_launcher` resource.

## Ideas of Implementation

Referencing this [function](https://github.com/dagster-io/dagster/blob/cca45616bbc99372a21a0f1394f90fdb49a9fa06/python_modules/libraries/dagster-databricks/dagster_databricks/configs.py#L197), we have solved this with the following:
```
def define_new_cluster():
    spark_version = Field(
        String,
        description="The Spark version of the cluster. "
        "A list of available Spark versions can be retrieved by using the "
        "Runtime versions API call. This field is required.",
        is_required=True,
    )

    spark_conf = Field(
        Permissive(),
        description="An object containing a set of optional, user-specified Spark configuration "
        "key-value pairs. You can also pass in a string of extra JVM options to the driver and the "
        "executors via spark.driver.extraJavaOptions and spark.executor.extraJavaOptions "
        "respectively. "
        "Example Spark confs: "
        '{"spark.speculation": true, "spark.streaming.ui.retainedBatches": 5} or '
        '{"spark.driver.extraJavaOptions": "-verbose:gc -XX:+PrintGCDetails"}',
        is_required=False,
    )

    aws_attributes = Field(Permissive(), is_required=False)  # NEW FIELD ADDED HERE; permissive could ofc be replaced with a more explicit schema

    ssh_public_keys = Field(
        [String],
        description="SSH public key contents that will be added to each Spark node in this "
        "cluster. The corresponding private keys can be used to login with the user name ubuntu "
        "on port 2200. Up to 10 keys can be specified.",
        is_required=False,
    )

    init_scripts = Field(
        [ddc._define_init_script()],
        description="The configuration for storing init scripts. Any number of scripts can be "
        "specified. The scripts are executed sequentially in the order provided. "
        "If cluster_log_conf is specified, init script logs are sent to "
[...]

PR #8684, opened 2022-06-30
Feature/phazure/8603 dagster_databricks - support aws_attributes

Phazure:
### Summary & Motivation
Original issue: https://github.com/dagster-io/dagster/issues/8603

Often for ops which leverage the databricks_pyspark_step_launcher, we execute the op logic within a new databricks job cluster. We would need the job cluster to have specific AWS permissions, and one way to achieve this is by using the aws_attributes/instance_profile_arn field as referenced in this [docs page](https://docs.databricks.com/dev-tools/api/latest/clusters.html).

Tested by building / installing a new wheel for dagster_databricks.


python_modules/libraries/dagster-databricks/dagster_databricks/configs.py
```diff
@@ -138,6 +138,83 @@ def _define_s3_storage_info():
     )
 
 
+def _define_aws_attributes_conf():
+    return Field(
+        Shape(
```
OwenKephart, line 143: this looks great, only recommendation is that it might be better to wrap this in Permissive rather than Shape, to make it possible to support new aws attributes without modifying this code

Phazure: Ha, that's actually how I had it originally, but thought we'd prefer to protect against misconfigurations and make any errors as explicit as possible. I'll change this back to `Permissive()`.

python_modules/libraries/dagster-databricks/dagster_databricks/configs.py
```diff
@@ -138,6 +138,16 @@ def _define_s3_storage_info():
     )
 
 
+def _define_aws_attributes_conf():
+    return Field(
+        Permissive(),
```
OwenKephart, line 143: ah for my original comment, I meant just substituting Shape(fields=...) for Permissive(fields=...). The Permissive object allows you to define some subset of possible fields, while allowing users to also include fields that are not included in that set.

this way, you get a bit of the best of both worlds -- you have stricter checking for the fields you know about, but in the future (if new attributes are added) you don't lock people out of using them. sorry for the confusion!

OwenKephart (commented):
Once that small comment is addressed, I'd be happy to merge this in, and it should be able to go out in this week's release -- thanks for the contribution!

python_modules/libraries/dagster-databricks/dagster_databricks/configs.py
Phazure: Gotcha - updated again!

OwenKephart (approved):
thanks for the contribution! this should be in this week's release

Issue #8606, opened 2022-06-24
dagster_databricks - support arbitrary env variables


Phazure:
## Use Case

It would be nice to be able to pass into Databricks cluster configurations arbitrary environment variables. I believe env variables can be provided using the `secrets_to_env_variables` key [here](https://github.com/dagster-io/dagster/blob/master/python_modules/libraries/dagster-databricks/dagster_databricks/databricks_pyspark_step_launcher.py#L52), but sometimes these variables aren't so secretive that they need to be actual secrets.

## Ideas of Implementation
We've gotten this into a working state with the following:
```
class CustomDatabricksPySparkStepLauncher(DatabricksPySparkStepLauncher):

    DATABRICKS_RUN_RETRIES = 3

    def __init__(
        self,
        run_config,
        databricks_host,
        databricks_token,
        secrets_to_env_variables,
        storage,
        env_variables,  # NEW FIELD HERE
        staging_prefix,
        wait_for_logs,
        max_completion_wait_time_seconds,
        poll_interval_sec=5,
        local_pipeline_package_path=None,
        local_dagster_job_package_path=None,
    ):
        super().__init__(
            run_config,
            databricks_host,
            databricks_token,
            secrets_to_env_variables,
            storage,
            staging_prefix,
            wait_for_logs,
            max_completion_wait_time_seconds,
            poll_interval_sec=poll_interval_sec,
            local_pipeline_package_path=local_pipeline_package_path,
            local_dagster_job_package_path=local_dagster_job_package_path,
        )
        self.env_variables = check.dict_param(env_variables, "env_variables")  # NEW FIELD HERE

    def _upload_artifacts(self, log, step_run_ref, run_id, step_key):
        """Upload the step run ref and pyspark code to DBFS to run as a job."""

        log.info("Uploading main file to DBFS")
        main_local_path = self._main_file_local_path()
        with open(main_local_path, "rb") as infile:
            self.databricks_runner.client.put_file(
                infile, self._dbfs_path(run_id, step_key, self._main_file_name()), overwrite=True
            )

        log.info("Uploading dagster job to DBFS")
        with tempfile.TemporaryDirectory() as temp_dir:
            # Zip and upload package containing dagster job
            zip_local_path = os.path.join(temp_dir, CODE_ZIP_NAME)
            build_pyspark_zip(zip_local_path, self.local_dagster_job_package_path)
            with open(zip_local_path, "rb") as infile:
                self.databricks_runner.client.put_file(
[...]

PR #8685, opened 2022-06-30
feature/Phazure/8606 dagster_databricks - support arbitrary env variables

Phazure:
### Summary & Motivation
Original issue: https://github.com/dagster-io/dagster/issues/8606

It would be nice to be able to pass into Databricks cluster configurations arbitrary environment variables. I believe env variables can be provided using the secrets_to_env_variables key [here](https://github.com/dagster-io/dagster/blob/master/python_modules/libraries/dagster-databricks/dagster_databricks/databricks_pyspark_step_launcher.py#L52), but sometimes these variables aren't so secretive that they need to be actual secrets.

Tested by building / installing a new wheel for dagster_databricks.


python_modules/libraries/dagster-databricks/dagster_databricks/configs.py
```diff
@@ -589,16 +589,14 @@ def _define_adls2_storage_credentials():
...
-
-
 def define_databricks_storage_config():
     return Field(
         Selector(
-            {"s3": _define_s3_storage_credentials(), "adls2": _define_adls2_storage_credentials()}
+            {
+                "s3": _define_s3_storage_credentials(),
+                "adls2": _define_adls2_storage_credentials(),
+                "N/A": Field(Permissive(), description="Use if no storage configuration necessary"),
```
OwenKephart, line 598: would you ever actually configure something within this "N/A" field? What's the difference between specifying the N/A field and just not configuring storage at all? If there's no difference, I'd recommend just removing the N/A field

python_modules/libraries/dagster-databricks/dagster_databricks/configs.py
```diff
@@ -607,6 +605,14 @@ def define_databricks_storage_config():
     )
 
 
+def define_databricks_env_variables():
+    return Field(
+        Permissive(),
+        description="Dictionary of arbitrary environment variables",
```
OwenKephart, line 611: This description could be a bit more specific, maybe "Dictionary of arbitrary environment variables to be set on the databricks cluster"?

OwenKephart (changes requested):
Thanks for the contribution! On the whole, this looks good, but I had some small comments to address.

python_modules/libraries/dagster-databricks/dagster_databricks/configs.py
Phazure: Gave it a shot locally but unfortunately doesn't do the trick (same `TypeError: __init__() missing 1 required positional argument: 'storage'` error).. Any other suggestions come to mind?
```
def __init__(
        self,
        run_config,
        permissions,
        databricks_host,
        databricks_token,
        secrets_to_env_variables,
        staging_prefix,
        wait_for_logs,
        max_completion_wait_time_seconds,
        env_variables={},
        storage={},
        poll_interval_sec=5,
        local_pipeline_package_path=None,
        local_dagster_job_package_path=None,
    ):
```

OwenKephart: hi @Phazure I can't see why you'd be hitting that error -- do you mind showing a bit more of the stack trace? Is this coming from the constructor of the DatabricksPysparkStepLauncher class or somewhere else? I'm wondering if that's an issue with the code being out of sync with wherever you're running it from (i.e. the file hasn't saved or something), just because in the code you have, `storage` is definitely not a required positional argument to that class.

Phazure: @OwenKephart Ah yes I rescind my statement - this does indeed work. Forgot to include `--force-reinstall` in the `pip install` command after making the new changes. Sorry for the confusion!

python_modules/libraries/dagster-databricks/dagster_databricks/databricks_pyspark_step_launcher.py
```diff
@@ -130,10 +132,11 @@ def __init__(
         databricks_host,
         databricks_token,
         secrets_to_env_variables,
-        storage,
         staging_prefix,
         wait_for_logs,
         max_completion_wait_time_seconds,
+        env_variables={},
```
OwenKephart, line 138: ok looks like there's a small linting issue with this setup (my bad). If you change this to env_variables=None, storage=None, and then change the check.dict_param calls to check.opt_dict_param() calls, then everything should sort itself out

Phazure: Updated.

OwenKephart (approved):
alright, this is good to go! thanks once more for all these contributions, we appreciate it!

Issue #8565, opened 2022-06-22
Returning an empty list raises InvariantViolationError when part of multiple outputs


abkfenris:
## Summary

When an op with multiple outputs returns an empty list Dagster raises an `DagsterInvariantViolationError`, where as the same op will work correctly when the list contains something.

## Reproduction

```py
from dagster import op, job, Out


@op(
    out={"list_out": Out(list[str]), "str_out": Out(str)}, config_schema={"error": bool}
)
def test_op(context):
    if context.op_config["error"]:
        return [], "error"

    return ["works"], "ok"


error_op = test_op.configured({"error": True}, name="error_op")
ok_op = test_op.configured({"error": False}, name="ok_op")


@job
def test_output_job():
    error_list, error_str = error_op()

    ok_list, ok_str = ok_op()
```

```py
dagster.core.errors.DagsterInvariantViolationError: Received a list of DynamicOutputs for output named 'list_out', but output is not dynamic.
  File "/usr/local/miniconda3/envs/dagster-bugs/lib/python3.9/site-packages/dagster/core/execution/plan/execute_plan.py", line 226, in dagster_event_sequence_for_step
    for step_event in check.generator(step_events):
  File "/usr/local/miniconda3/envs/dagster-bugs/lib/python3.9/site-packages/dagster/core/execution/plan/execute_step.py", line 353, in core_dagster_event_sequence_for_step
    for user_event in check.generator(
  File "/usr/local/miniconda3/envs/dagster-bugs/lib/python3.9/site-packages/dagster/core/execution/plan/execute_step.py", line 69, in _step_output_error_checked_user_event_sequence
    for user_event in user_event_sequence:
  File "/usr/local/miniconda3/envs/dagster-bugs/lib/python3.9/site-packages/dagster/core/execution/plan/compute.py", line 174, in execute_core_compute
    for step_output in _yield_compute_results(step_context, inputs, compute_fn):
  File "/usr/local/miniconda3/envs/dagster-bugs/lib/python3.9/site-packages/dagster/core/execution/plan/compute.py", line 142, in _yield_compute_results
    for event in iterate_with_context(
  File "/usr/local/miniconda3/envs/dagster-bugs/lib/python3.9/site-packages/dagster/utils/__init__.py", line 406, in iterate_with_context
    next_output = next(iterator)
[...]

PR #8686, opened 2022-06-30
Fix error with list as output type for ops

dpeng817:
Fixes https://github.com/dagster-io/dagster/issues/8565. `all(...)` evaluates to true on an empty list.


sryza (approved):

dpeng817:
more comprehensive fix in another PR

PR #8693, opened 2022-06-30
docs: improve left nav for items with children DREL-359

yuhan:
### Summary & Motivation
This PR:
- allows to open and close items in the sidebar without leaving the current page
- allows to open multiple items in the sidebar at a time
- allows to set default open section via `"isDefaultOpen": true` in `_navigation.json` [OPEN: which ones do we want to open by default]
 
preview

### DIFF
--- a/docs/content/_navigation.json
+++ b/docs/content/_navigation.json
@@ -71,6 +71,7 @@
     "title": "Concepts",
     "icon": "Menu",
     "path": "/concepts",
+    "isDefaultOpen": true,
     "children": [
       {
         "title": "Assets",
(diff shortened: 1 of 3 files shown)


braunjj (commented):
This is GLORIOUS! 😍

docs/content/_navigation.json
```diff
@@ -71,6 +71,7 @@
     "title": "Concepts",
     "icon": "Menu",
     "path": "/concepts",
+    "isDefaultOpen": true,
```
erinkcochran87, line 74: Having both of these open by default is a bit much - combo of too long and a bit cluttered, due to list items in the **Guides** section having lengthy names. Maybe we just go with **Concepts** for the time being? We can always revisit as part of the consolidation work we're doing.

yuhan: sure - will update. we'll change the nav in the consolidation work anyways and we can keep "the default open section" as a bullet point to decide in nav revamp :)

erinkcochran87 (approved):

PR #8697, opened 2022-06-30
docs: fix prev/next pagination DREL-345

yuhan:
### Summary & Motivation
the problem was bc we didn't use the normalized (remove <version> from the path) `asPath` for pulling the right `currentIndex` for the page.
before:
Screen Shot 2022-06-30 at 3 58 57 PM

after:
Screen Shot 2022-06-30 at 3 59 01 PM


erinkcochran87 (approved):
🔥

PR #8703, opened 2022-07-01
order the backfill partitions before creating

prha:
Selected a bunch of random partitions, saw the backfill partitions were ordered.

### DIFF
--- a/js_modules/dagit/packages/core/src/partitions/BackfillSelector.tsx
+++ b/js_modules/dagit/packages/core/src/partitions/BackfillSelector.tsx
@@ -154,7 +154,9 @@ export const BackfillPartitionSelector: React.FC<{
     (name: string) => partitionData[name] === RunStatus.SUCCESS,
   );
   const setSelected = (selection: string[]) => {
-    _setSelected(selection);
+    const selectionSet = new Set(selection);
+    // first order the partition names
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/core/src/partitions/BackfillSelector.tsx
```diff
@@ -154,7 +154,9 @@ export const BackfillPartitionSelector: React.FC<{
     (name: string) => partitionData[name] === RunStatus.SUCCESS,
   );
   const setSelected = (selection: string[]) => {
-    _setSelected(selection);
+    const selectionSet = new Set(selection);
+    // first order the partition names
+    _setSelected(partitionNames.filter((name) => selectionSet.has(name)));
```
rexledesma, line 159: It's a little unintuitive how this is sorting the selection - maybe also include a note that `partitionNames` is already sorted?

rexledesma (approved):

js_modules/dagit/packages/core/src/partitions/BackfillSelector.tsx
prha: hmm... yeah.  It's not sorted by any sort of sort order.   Partition names are given in a particular order by user code.  So we have to maintain the order that they're given to us... but yeah, can add a comment about it.

PR #8709, opened 2022-07-01
[dagit] Bring back the global asset graph as an “all asset groups” view

bengotow:
### Summary & Motivation

This PR adds a new `/instance/asset-groups` page (currently not linked to from anywhere) that restores the "Global Asset Graph" as an "All Asset Groups" page. The plan is to share this with customers that ask for it, while also collecting some feedback about what they're using it for.

Eventually, this will probably get a link / placement within Josh's new designs.

Features:

- A repository filter lets you filter down the groups that are displayed.

- The group outlines are back! They appear whenever a DAG is rendering content that spans more than one asset group, which includes the lineage views. I /think/ this is nice, but we should give it a spin and see. We could alternatively only show the grouping on this global graph. 

- I fixed a few size calculations and z-indexes that were causing slight rendering issues.

Caveats / Questions:

- We currently show just the asset group name, not it's repo + location, but we do show a separate box for each repo's group (eg: in my screenshots, there are 6 "default" groups and not just one.) I think this is clean and probably clear enough for people who are familiar with the assets, but I'm not 100% sure.

image
image

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
@@ -1,8 +1,17 @@
-import {Box, Checkbox, NonIdealState, SplitPanelContainer} from '@dagster-io/ui';
+import {
+  Box,
+  Checkbox,
+  Colors,
+  Icon,
+  Mono,
+  NonIdealState,
+  SplitPanelContainer,
+} from '@dagster-io/ui';
 import pickBy from 'lodash/pickBy';
 import uniq from 'lodash/uniq';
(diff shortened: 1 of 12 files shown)


sryza:
Awesome! I loaded this up locally and it all works out. I like the group background color better than the previous one.

Small thing - when clicking around, I hoped that clicking on the name of a group would take me to the page for that group.

> We currently show just the asset group name, not it's repo + location, but we do show a separate box for each repo's group (eg: in my screenshots, there are 6 "default" groups and not just one.) I think this is clean and probably clear enough for people who are familiar with the assets, but I'm not 100% sure.

Would it make sense to just show repo + location in cases where the group names collide?

PR #8714, opened 2022-07-01
[WIP] [docs] Dagster + dbt guide

OwenKephart:
### Summary & Motivation

We want to SDA-ify the existing guide.

Beyond being very op-focused, I feel like the existing version of the guide had way too much introduction stuff. For example, it doesn't seem necessary to explain what dbt is to someone who is on the page for this guide -- if they don't already know, then they are probably not the target audience for this guide.

My feeling is that there should be a short blurb somewhere at the top mentioning that if you just want to run a dbt command as part of a traditional job, that is possible, but beyond that we don't need to mention the oppy stuff, as this is all covered in the API docs.


docs/content/integrations/dbt.mdx
```diff
@@ -1,380 +1,159 @@
...
 ---
 
-# Using dbt with Dagster
+## Using dbt with Dagster
 
-This guide explains how you can run a [dbt](https://docs.getdbt.com/docs/introduction) project as part of a Dagster job. This guide focuses on dbt Core, the open-source dbt offering. Dagster also offers an integration with the hosted version of dbt, [dbt Cloud](https://docs.getdbt.com/docs/dbt-cloud/cloud-overview). Check out the [API docs](/\_apidocs/libraries/dagster-dbt#dagster_dbt.dbt_cloud_resource) to learn more.
+Dagster orchestrates dbt alongside _other technologies_, so you can combine dbt with Spark, Python, etc. in a single workflow. Dagster's software-defined asset abstractions make it simple to define data assets that depend on specific dbt models, or to define the computation required to compute the sources that your dbt models depend on.
 
-## What is dbt?
+This guide focuses on how to work with `dbt` models through the SDA framework, which we recommend for most use cases. Dagster's SDAs allow you to map multiple
```
erinkcochran87, line 10: Unfinished sentence?

docs/content/integrations/dbt.mdx
```diff
@@ -1,380 +1,159 @@
...
+## dbt Models and Software-Defined Assets
 
-dbt not only runs your data transformations, but also can create data quality tests and generate documentation for your data, right out of the box. To learn more about dbt, visit [the official dbt documentation website](https://docs.getdbt.com/docs/introduction).
+A software-defined asset contains an asset key, a set of upstream asset keys, and an operation that is responsible for computing the asset from its upstream dependencies. Models defined in a `dbt` project are conceptually similar to Dagster's software-defined assets.
 
-## How does dbt work with Dagster?
+- The asset key for a dbt model is (by default) the name of the model.
+- The upstream dependencies of a dbt model are defined with `ref` or `source` calls within the model's definition.
+- The computation required to compute the asset from its upstream dependencies is the sql within the model's definition.
```
erinkcochran87, line 18:
```suggestion
- The computation required to compute the asset from its upstream dependencies is the SQL within the model's definition.
```

docs/content/integrations/dbt.mdx
```diff
@@ -1,380 +1,159 @@
...
 
-Dagster orchestrates dbt alongside _other technologies_, so you can combine dbt with Spark, Python, etc. in a single job. Dagster also provides built-in operational and data observability capabilities, like storing dbt run results longitudinally and sending alerts when a dbt run fails.
+These similarities make it natural to want to interact with dbt models as SDAs, making it possible to
 
-`dagster-dbt` is an integration library that provides pre-built resources for using dbt together with Dagster. These are all designed to be configurable for any dbt project.
+### Loading `dbt` models from a dbt project
 
-The resources that dagster_dbt provides are
+For smaller dbt projects, where compilation time is not a concern, the simplest way to load your dbt assets into Dagster is the following:
```
erinkcochran87, line 24:
```suggestion
For smaller dbt projects where compilation time is not a concern, the simplest way to load your dbt assets into Dagster is the following:
```

Removing an unneeded comma

docs/content/integrations/dbt.mdx
```diff
@@ -1,380 +1,159 @@
...
-
-@job(resource_defs={"dbt": my_dbt_resource})
-def my_dbt_job():
-    dbt_run_op()
 ```
 
-#### Using dbt_cli_resource to run a specific set of models
-
-Sometimes, you just want to run a select set of models in your dbt project, rather than the entire thing. The below examples show two ways of doing this, depending on your use case.
+The `load_assets_from_dbt_project` function will compile your dbt project, parse the metadata that dbt provides, and generate a set of software-defined assets that reflect the models in the project. These assets will all share the same underlying operation, which will invoke dbt to run the models represented by the loaded assets.
```
erinkcochran87, line 34: Suggesting an alternative formatting approach to break this up and make it more readable:

```suggestion
The `load_assets_from_dbt_project` function:

1. Compiles your dbt project,
2. Parses the metadata that dbt provides, and
3. Generates a set of software-defined assets that reflect the models in the project. These assets will share the same underlying operation, which will invoke dbt to run the models represented by the loaded assets.
```

docs/content/integrations/dbt.mdx
```diff
@@ -1,380 +1,159 @@
...
 
-Note that in both cases, the `models` option takes in a list of strings. The string `"tag:staging"` uses [dbt's node selection syntax](https://docs.getdbt.com/reference/node-selection/syntax) to filter models with the tag `"staging"`. For more details, [visit the official dbt documentation on the node selection syntax](https://docs.getdbt.com/reference/node-selection/syntax).
+### Loading `dbt` models from a dbt manifest.json file
```
erinkcochran87, line 36:
```suggestion
### Loading dbt models from a dbt manifest.json file
```

Let's keep additional formatting out of headings :)

docs/content/integrations/dbt.mdx
```diff
@@ -1,380 +1,159 @@
...
 
-##### **... supplying an argument**
+Note: if you make any changes to your dbt project that change the structure of the project (such as changing the dependencies of a model or adding a new one), you'll need to regenerate your manifest file for those changes to be reflected in Dagster.
 
-If you want to change which models you select depending on what happens during execution, you should create your own op that invokes the dbt resource (rather than using a prebuilt op). This allows you to explicitly set the `models` parameter at runtime. This would look similar to the following:
+## Adding a resource
 
-```python file=/integrations/dbt.py startafter=start_marker_dbt_cli_run_specific_models_runtime endbefore=end_marker_dbt_cli_run_specific_models_runtime dedent=4
-from dagster import op
+In order to function, these assets require a "dbt" resource, which will be responsible for firing off dbt cli commands. The `dagster-dbt` integration provides the `dbt_cli_resource` to make this easy. This resource can be configured with CLI flags that will be passed i
```
erinkcochran87, line 53: > In order to function, these assets require a "dbt" resource, which will be responsible for firing off dbt cli commands.

1. Are the quotes for `"dbt"` necessary? What's the purpose of having them here?
2. As CLI is an acronym, I feel like it should be in all caps. Is there a reason it wouldn't be?

erinkcochran87, line 53: > This resource can be configured with CLI flags that will be passed i

Unfinished sentence?

docs/content/integrations/dbt.mdx
```diff
@@ -1,380 +1,159 @@
...
-        context.resources.dbt.run(models=["tag:other"])
-```
-
-#### Using a different dbt profile for different dagster modes
-
-Dagster supports creating [multiple jobs from the same graph](/concepts/ops-jobs-graphs/jobs#from-a-graph). dbt has a similar concept, [profiles](https://docs.getdbt.com/dbt-cli/configure-your-profile). You might want to run a dev version of your graph that targets the development-specific dbt profile, and then have a prod version that runs using the prod dbt profile. This example shows how to accomplish this.
-
-```python file=/integrations/dbt.py startafter=start_marker_dbt_cli_profile_modes endbefore=end_marker_dbt_cli_profile_modes dedent=4
-from dagster_dbt import dbt_cli_resource, dbt_run_op
+You can configure this resource and add it to your dbt assets:
```
erinkcochran87, line 55: This comes across as an incomplete sentence - maybe just adding `by` to the end before the colon would help. That links the copy to the example that follows.

docs/content/integrations/dbt.mdx
```diff
@@ -1,380 +1,159 @@
...
-@job(resource_defs={"dbt": my_dbt_resource})
-def my_dbt_job():
-    dbt_test_op(start_after=dbt_run_op())
-```
-
-## Using a dbt RPC server in a Dagster job
+## Scheduling dbt jobs
 
-This integration provides two separate resources to help run commands against a [dbt RPC server](https://docs.getdbt.com/reference/commands/rpc). The `dbt_rpc_resource` will fire off requests against the server and return immediately, while the `dbt_rpc_sync_resource` will poll the server until each request completes.
+Once you have your dbt assets, it's simple to define a job that runs some or all of these assets on a schedule
```
erinkcochran87, line 73: I'm sorta against including words like 'simple' or 'easy' in documentation because for some folks, it *isn't* easy and I don't want them to be frustrated. It also tends to make copy wordier than it needs to be.

What about something like:

> Once your dbt assets are ready, you can define a job that runs some or all of the assets on a schedule:

docs/content/integrations/dbt.mdx
```diff
@@ -1,380 +1,159 @@
...
+See (docs for running stuff on a schedule) to learn more.
 
-```python file=/integrations/dbt.py startafter=start_marker_dbt_cli_config_profile_and_target endbefore=end_marker_dbt_cli_config_profile_and_target dedent=4
-from dagster_dbt import dbt_cli_resource
-
-from dagster import job
+## Defining Downstream Dependencies
 
-config = {"profile": PROFILE_NAME, "target": TARGET_NAME}
+Dagster allows you to define assets that are downstream of specific dbt models. One property of dbt based assets is that the external tool (in this case, dbt), handles the loading/storing of those models internally, rather than Dagster directly storing/loading the tables that are updated.
```
erinkcochran87, line 102:
```suggestion
Dagster allows you to define assets that are downstream of specific dbt models. One property of dbt-based assets is that the external tool - in this case, dbt - handles the loading/storing of those models internally, rather than Dagster directly storing/loading the tables that are updated.
```

docs/content/integrations/dbt.mdx
```diff
@@ -1,380 +1,159 @@
...
 
-@job(resource_defs={"dbt": dbt_cli_resource.configured(config)})
-def my_job():
-    # ...
-```
+This means that there's a range of ways that we might want to load a dbt model as input to a python function. For example, you might want to load the contents of the table as a Pandas dataframe, or you might want to load the contents into a PySpark session.
```
erinkcochran87, line 104:
```suggestion
This means that there's a range of ways to load a dbt model as input to a Python function. For example, you might want to load the contents as a Pandas dataframe or into a PySpark session.
```

I wanted to tighten this up a bit. I'm not sure if my removing `... of the table` changes the meaning or accuracy of this sentence, but given there wasn't a mention of `tables` before, I thought it was okay. Feel free to put it back/tinker if I'm wrong.

docs/content/integrations/dbt.mdx
```diff
@@ -1,380 +1,159 @@
...
-    # ...
+@asset(
+    ins={"model_name": AssetIn(input_manager_key="pandas_dbt_input_manager")}
+)
+def my_downstream_asset(model_name):
+    return
 ```
 
-**dbt CLI: Select specific models to run**
+## Customizing AssetKeys
```
erinkcochran87, line 118: This comment is applicable to the mention of AssetKeys in this section as a whole.

What is/are AssetKey(s) referring to, here? Is it the concept of an asset key *or* a literal object or value you'd see in code? Ex:

- Speaking generally/conceptually: "Each asset has an asset key..."
- Referring to a literal object or value in code: "The `AssetKey`..."

docs/content/integrations/dbt.mdx
```diff
@@ -1,380 +1,159 @@
...
 
-```python file=/integrations/dbt.py startafter=start_marker_dbt_cli_config_select_models endbefore=end_marker_dbt_cli_config_select_models dedent=4
-from dagster_dbt import dbt_cli_resource
+Dagster automatically generates AssetKeys for each dbt model in the project, as well as any of the sources used by these models.
```
erinkcochran87, line 120:
```suggestion
Dagster automatically generates AssetKeys for each dbt model in the project and the sources used by the models.
```

docs/content/integrations/dbt.mdx
```diff
@@ -1,380 +1,159 @@
...
 
-from dagster import job
+The default asset key for a model will be the model name, and the default name for a source in the project will be the name of the source alongside the name of the source table.
 
-config = {"models": ["my_dbt_model+", "path.to.models", "tag:nightly"]}
+For example, a model defined in `my_model.sql` will be given the asset key `my_model`, while the source table defined in the following `sources.yaml` will be `jaffle_shop/orders`.
```
erinkcochran87, line 124:
```suggestion
For example, a model defined in `my_model.sql` will be given the asset key `my_model`, while the source table defined in the following `sources.yaml` will be `jaffle_shop/orders`:
```

Replaced the period with a colon to connect the copy to the example that follows.

erinkcochran87 (changes requested):
Back to you!

docs/content/integrations/dbt.mdx
OwenKephart: I put the quotes there to hint that the resource itself needs to be named "dbt", but I don't think it's necessary so I'll remove them. No real reason to make cli lowercase either

docs/content/integrations/dbt.mdx
OwenKephart: It's referring to both but I think speaking conceptually is better in this case

docs/content/integrations/dbt.mdx
```diff
@@ -1,380 +1,159 @@
...
+### Fully customizing AssetKeys
 
-**dbt RPC: Configure a remote dbt RPC resource**
-
-```python file=/integrations/dbt.py startafter=start_marker_dbt_rpc_resource_example endbefore=end_marker_dbt_rpc_resource_example dedent=4
-from dagster_dbt import dbt_rpc_resource
-
-custom_resource = dbt_rpc_resource.configured({"host": HOST, "post": PORT})
-```
+Sometimes, you want to have full control over what asset key you generate. You can do this by setting the `node_info_to_asset_key` parameter to a function that takes a dictionary of dbt node info (parsed from the dbt manifest) and returns an asset key. This dictionary contains (link to doc).
```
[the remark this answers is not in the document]
OwenKephart: I decided to remove this section because it's a bit of a niche feature and takes up a lot of space compared to other parts

OwenKephart:
@erinkcochran87 thanks once again for the feedback -- I reworked my vision for this guide a bit and put more focus on the "defining downstream assets" section, which I think is pretty critical for people that want to use this in the wild (and can be hard to figure out for yourself). It might be a bit much to include an IOManager implementation directly in the docs, but I think it might be useful to give people something they can copy paste and expand on.

I'm also feeling like the "understanding asset keys" section might be misplaced a bit.

docs/content/integrations/dbt.mdx
```diff
@@ -1,379 +1,209 @@
 ---
 title: Dagster with dbt
```
erinkcochran87, line 2:
```suggestion
title: "Using Dagster with dbt"
```

Suggesting we make the page title match the H1/title on the page itself.

docs/content/integrations/dbt.mdx
```diff
@@ -1,379 +1,209 @@
...
+sources:
+  - name: jaffle_shop
+    tables:
+      - name: orders
 ```
 
-For more details, [visit the official documentation on dbt's node selection syntax](https://docs.getdbt.com/reference/node-selection/syntax).
-
-**dbt CLI: Set key-values for dbt vars**
+### Adding a prefix to AssetKeys
```
erinkcochran87, line 190:
```suggestion
### Adding a prefix to asset keys
```

erinkcochran87: Just saw your note about this section feeling misplaced - I agree.

What if we moved this up to come after the **dbt models and software-defined assets** section? This section relates to assets themselves, so it feels like that might be a good place for it instead of at the end. Especially with the title `Understanding asset keys`.

Thoughts?

docs/content/integrations/dbt.mdx
```diff
@@ -66,9 +52,11 @@ Note: if you make any changes to your dbt project that change the structure of t
 
 ## Adding a resource
 
-In order to function, these assets require a "dbt" resource, which will be responsible for firing off dbt cli commands. The `dagster-dbt` integration provides the `dbt_cli_resource` to make this easy. This resource can be configured with CLI flags that will be passed i
+In order to be executed, these assets require a dbt resource, which will be responsible for firing off dbt cli commands. The `dagster-dbt` integration provides the `dbt_cli_resource` for this purpose. This resource can be configured with CLI flags that will be passed into every dbt invocation.
```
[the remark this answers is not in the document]
erinkcochran87: Also, what assets is `these assets` referring to? Should this be something like `Dagster assets` to be more specific? Or should we just drop `these`?

docs/content/integrations/dbt.mdx
```diff
@@ -1,379 +1,209 @@
...
+        ),
+        ScheduleDefinition(
+            job=run_everything,
+            cron_schedule="@weekly",
+        )
+    ]
 ```
 
-Note that the job above will NOT wait until the dbt RPC server has finished executing your request. Instead, it will return immediately with a request token from the dbt RPC server. If you want the op to wait until execution is finished, instead supply a `dbt_rpc_sync_resource` (seen below).
+See the docs on [running jobs on a schedule](https://docs.dagster.io/concepts/partitions-schedules-sensors/schedules#running-the-scheduler) to learn more.
```
erinkcochran87, line 102:
```suggestion
Refer to the [Schedule documentation](/concepts/partitions-schedules-sensors/schedules#running-the-scheduler) for more info on running jobs on a schedule.
```

This is also a nit - sorry.

docs/content/integrations/dbt.mdx
```diff
@@ -111,33 +99,86 @@ def my_repo():
...
+```python
+@asset(
+    ins={"my_dbt_model": AssetIn(input_manager_key="pandas_df_manager")}
+)
+def my_downstream_asset(my_dbt_model):
+    # my_dbt_model is a Pandas DataFrame
+    return my_dbt_model.where(foo="bar")
+```
+
+### Defining your IOManager
```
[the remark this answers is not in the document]
erinkcochran87: I also just noticed this is an H3, so it sits under the **Defining downstream dependencies** section. Is that correct or should this be its own section?

docs/content/integrations/dbt.mdx
```diff
@@ -1,380 +1,159 @@
...
+```python
+dbt_assets = load_assets_from_dbt_project(
+    ..., key_prefix="snowflake"
+)
 ```
 
-For more details, [visit the official documentation on dbt's node selection syntax](https://docs.getdbt.com/reference/node-selection/syntax).
-
-**dbt CLI: Set key-values for dbt vars**
+This prefix will apply only to the models. If you want to apply a prefix to the source keys that dagster generates:
```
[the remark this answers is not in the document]
erinkcochran87: It looks like this didn't get applied.

docs/content/integrations/dbt.mdx
```diff
@@ -1,379 +1,209 @@
...
+Dagster orchestrates dbt alongside _other technologies_, so you can combine dbt with Spark, Python, etc. in a single workflow. Dagster's software-defined asset (SDA) abstractions make it simple to define data assets that depend on specific dbt models, or to define the computation required to compute the sources that your dbt models depend on.
 
-## What is dbt?
+This guide focuses on how to work with `dbt` models through the SDA framework, which we recommend for most use cases.
 
-[dbt](https://docs.getdbt.com/docs/introduction) (data build tool) helps engineers transform data in their warehouses by simply writing `SELECT` statements. dbt automatically builds a dependency graph for your transformations and turns these `SELECT` statements into tables and views in your data warehouse.
+## dbt Models and Software-Defined Assets
```
erinkcochran87, line 12:
```suggestion
## dbt models and software-defined assets
```

Making the casing consistent.

erinkcochran87 (changes requested):
This is looking great! Two things:

1. Some of the code snippets look to be hard-coded - we should probably put them in the `doc_snippets` folder, yeah?
2. I wonder if we should pass this over to another engineer for a completeness/technical accuracy check - thoughts?

docs/content/integrations/dbt.mdx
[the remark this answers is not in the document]
erinkcochran87: Gotcha - thanks for clarifying!

What about something like this, which specifies the assets being discussed?: `Assets loaded from dbt require a dbt resource, which is responsible for firing off dbt CLI commands.`

docs/content/integrations/dbt.mdx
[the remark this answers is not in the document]
OwenKephart: this is correct, as you generally need to define an IOManager if you want to load the contents of a dbt asset in a downstream dependency

and no worries on the nitpicks, appreciate the attention to detail :)

docs/content/integrations/dbt.mdx
```diff
@@ -111,33 +99,86 @@ def my_repo():
...
+    ins={"my_dbt_model": AssetIn(input_manager_key="pandas_df_manager")}
+)
+def my_downstream_asset(my_dbt_model):
+    # my_dbt_model is a Pandas DataFrame
+    return my_dbt_model.where(foo="bar")
+```
+
+### Defining your IOManager
+
+The implementation of your IOManager will depend on what python object you want to use to represent your table, as well as what database you're loading from. A simple IOManager implementation that loads data from a dbt-managed table into a Pandas DataFrame would look somthing like the following:
```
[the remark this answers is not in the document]
OwenKephart: added some clarification re: the database question

erinkcochran87 (approved):

PR #8717, opened 2022-07-01
skip threadpool sensor daemon tests

prha:
### Summary & Motivation
Disable threaded tests for now.

BK


python_modules/dagster/dagster_tests/daemon_tests/test_sensor_run.py
```diff
@@ -383,10 +382,7 @@ def workspace_load_target(attribute="the_repo"):
 
 
 def get_sensor_executors():
-    return [
-        SingleThreadPoolExecutor(),
-        SynchronousExecutor(),
-    ]
```
rexledesma, line 389: We can wrap `SingleThreadPoolExecutor` in `pytest.param` with a skipped mark to explicitly know that it's being skipped: https://stackoverflow.com/a/30575822

rexledesma (approved):

PR #8720, opened 2022-07-01
prevent multiple sensor evaluations from multithreaded race conditions

prha:
### Summary & Motivation
Seeing more than expected runs generated from the stress test.

Typically we can hit these race conditions if the evaluation takes longer than the minimum interval (30 seconds).

BK

### DIFF
--- a/python_modules/dagster/dagster_tests/daemon_tests/test_sensor_failure_recovery.py
+++ b/python_modules/dagster/dagster_tests/daemon_tests/test_sensor_failure_recovery.py
@@ -198,7 +198,7 @@ def test_failure_after_run_created_before_run_launched(crash_location, crash_sig
 
             launch_process = spawn_ctx.Process(
                 target=_test_launch_sensor_runs_in_subprocess,
-                args=[instance.get_ref(), frozen_datetime.add(seconds=1), None],
+                args=[instance.get_ref(), frozen_datetime.add(seconds=31), None],
             )
             launch_process.start()
             launch_process.join(timeout=60)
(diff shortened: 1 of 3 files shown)


alangenfeld (commented):
can you expand on what exactly the error is that we are seeing ? looks like we are adding a lock and a second write which i don't find immediately obvious

prha:
Yeah, we're updating the timestamp of when we've last touched the tick at the end of the sensor evaluation, but we're using the timestamp of when the tick was queued.  This means that under high load, we could put a sensor evaluation on the queue, it could take some amount of time to actually evaluate, and if that takes longer than 30 seconds since the last time it evaluated, another evaluation could be queued.  

This change makes sure that the timestamp is updated immediately, so it isn't assigned to another thread.

python_modules/dagster/dagster_tests/daemon_tests/test_sensor_failure_recovery.py
```diff
@@ -57,7 +57,8 @@ def _test_launch_sensor_runs_in_subprocess(instance_ref, execution_datetime, deb
 @pytest.mark.skipif(
     IS_WINDOWS, reason="Windows keeps resources open after termination in a flaky way"
 )
-@pytest.mark.parametrize("crash_location", ["TICK_CREATED", "TICK_HELD"])
+@pytest.mark.parametrize("crash_location", ["TICK_CREATED"])
+# @pytest.mark.parametrize("crash_location", ["TICK_CREATED", "TICK_HELD"])
```
alangenfeld, line 61: ?

alangenfeld (approved):

PR #8724, opened 2022-07-01
add retention helm values

prha:
### Summary & Motivation
We added instance configuration for tick retention, but didn't expose it in the helm chart.

BK

### DIFF
--- a/helm/dagster/schema/schema/charts/dagster/subschema/retention.py
+++ b/helm/dagster/schema/schema/charts/dagster/subschema/retention.py
@@ -4,10 +4,10 @@ from pydantic import BaseModel, Extra
 
 
 class TickRetentionByType(BaseModel):
-    skipped: Optional[int]
-    success: Optional[int]
-    failure: Optional[int]
-    started: Optional[int]
+    skipped: int
+    success: int
+    failure: int
+    started: int
 
     class Config:
         extra = Extra.forbid
@@ -18,6 +18,7 @@ class TickRetention(BaseModel):
 
 
 class Retention(BaseModel):
+    enabled: bool
     sensor: TickRetention
     schedule: TickRetention
 
--- a/helm/dagster/schema/schema/charts/dagster/subschema/retention.py
+++ b/helm/dagster/schema/schema/charts/dagster/subschema/retention.py
@@ -1,0 +1,22 @@
+from pydantic import BaseModel, Extra
+from typing import Any, Dict, List, Optional, Union
+
+
+class TickRetentionByType(BaseModel):
+    skipped: Optional[int]
+    success: Optional[int]
+    failure: Optional[int]
+    started: Optional[int]
+
+    class Config:
+        extra = Extra.forbid
+
+class TickRetention(BaseModel):
+    purgeAfterDays: Union[int, TickRetentionByType]
+
+class Retention(BaseModel):
+    sensor: TickRetention
+    schedule: TickRetention
+
+    class Config:
+        extra = Extra.forbid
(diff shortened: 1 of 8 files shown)


helm/dagster/values.yaml
```diff
@@ -1142,3 +1142,13 @@ serviceAccount:
...
+
+retention:
+  schedule:
+    purgeAfterDays: -1
+  sensor:
+    purgeAfterDays:
+      failure: -1
+      skipped: 7
+      started: -1
+      success: -1
```
rexledesma, line 1154: should probably be nested under `dagsterDaemon`. documentation for each field would be nice too. Adding a link to https://docs.dagster.io/deployment/dagster-instance#data-retention would be great as well

helm/dagster/schema/schema/charts/dagster/subschema/retention.py
```diff
@@ -0,0 +1,25 @@
+from typing import Any, Dict, List, Optional, Union
+
+from pydantic import BaseModel, Extra
+
+
+class TickRetentionByType(BaseModel):
+    skipped: Optional[int]
+    success: Optional[int]
+    failure: Optional[int]
+    started: Optional[int]
```
rexledesma, line 10: Technically, these are not optional, since you've put default values in `values.yaml`.
```suggestion
    skipped: int
    success: int
    failure: int
    started: int
```

helm/dagster/templates/configmap-instance.yaml
```diff
@@ -116,3 +116,15 @@ data:
...
+    retention:
+      {{- if .Values.retention.sensor }}
+      sensor:
+        purge_after_days: {{ .Values.retention.sensor.purgeAfterDays | toYaml | nindent 12 }}
+      {{- end }}
+      {{- if .Values.retention.schedule }}
+      schedule:
+        purge_after_days: {{ .Values.retention.schedule.purgeAfterDays | toYaml | nindent 12 }}
+      {{- end }}
+    {{- end }}
```
rexledesma, line 130: for backcompat, should include a `.Values.retention.enabled` field. As it stands, this field will always be included.

helm/dagster/schema/schema_tests/test_instance.py
```diff
@@ -663,3 +665,30 @@ def test_run_coordinator_has_schema(json_schema_model, run_coordinator_class):
     run_coordinator_fields = set(map(to_camel_case, run_coordinator_class.config_type().keys()))
 
     assert json_schema_fields == run_coordinator_fields
+
+
+def test_retention(template: HelmTemplate):
```
rexledesma, line 670: add test for backcompat (e.g. with default Helm values, this field is not included in `instance.yaml`)

helm/dagster/schema/schema/charts/dagster/subschema/retention.py
```diff
@@ -0,0 +1,25 @@
...
+
+    class Config:
+        extra = Extra.forbid
+
+
+class TickRetention(BaseModel):
+    purgeAfterDays: Union[int, TickRetentionByType]
```
rexledesma, line 17: It's not intuitive that this is a union. I don't think it's clear on https://docs.dagster.io/deployment/dagster-instance#data-retention either.

rexledesma (dismissed):
need to enforce backcompat + documentation

helm/dagster/values.yaml
prha: Added some documentation, but resisting nesting under dagsterDaemon, which feels more like an implementation detail.  Also, we might introduce new fields with data retention policies that don't have to do with schedule/sensor ticks.

rexledesma (approved):

PR #8729, opened 2022-07-03
Improve race condition in run monitor

brad-alexander:
### Summary & Motivation
The run monitor seems to rely on the results of the initial `get_runs` results, not taking into account the status of the run may have changed since that scan.  This can sometimes result in the run being marked as failed after it has succeeded.

`resume_run` already checks the status of the run before modification.

Added some additional logic in `monitor_started_run` to trust `check_run_worker_health` if it reports the run as succeeded, checks that the run status has not changed, and fixed a small bug with f-string.

bk

### DIFF
--- a/python_modules/dagster/dagster/core/instance/__init__.py
+++ b/python_modules/dagster/dagster/core/instance/__init__.py
@@ -1611,6 +1611,7 @@ class DagsterInstance:
         from dagster.core.events import DagsterEvent, DagsterEventType
 
         check.inst_param(pipeline_run, "pipeline_run", PipelineRun)
+
         message = check.opt_str_param(
             message,
             "message",
--- a/python_modules/dagster/dagster/core/instance/__init__.py
+++ b/python_modules/dagster/dagster/core/instance/__init__.py
@@ -1611,7 +1611,6 @@ class DagsterInstance:
         from dagster.core.events import DagsterEvent, DagsterEventType
 
         check.inst_param(pipeline_run, "pipeline_run", PipelineRun)
-        
         message = check.opt_str_param(
             message,
             "message",
--- a/python_modules/dagster/dagster/core/instance/__init__.py
+++ b/python_modules/dagster/dagster/core/instance/__init__.py
@@ -1611,12 +1611,22 @@ class DagsterInstance:
         from dagster.core.events import DagsterEvent, DagsterEventType
 
         check.inst_param(pipeline_run, "pipeline_run", PipelineRun)
-
+        
         message = check.opt_str_param(
             message,
             "message",
             "This run has been marked as failed from outside the execution context.",
         )
+
+        run = self.get_run_by_id(pipeline_run.run_id)
+        if run is None:
+            raise DagsterInvariantViolationError(
+                f"Could not load run {pipeline_run.run_id} that was passed to report_run_failed"
+            )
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/core/instance/__init__.py
```diff
@@ -1611,13 +1611,23 @@ def report_run_failed(self, pipeline_run, message=None):
         from dagster.core.events import DagsterEvent, DagsterEventType
 
         check.inst_param(pipeline_run, "pipeline_run", PipelineRun)
-
+
```
brad-alexander, line 1614:
```suggestion

```
unintentional whitespace change

python_modules/dagster/dagster/core/instance/__init__.py
```diff
@@ -1618,6 +1618,16 @@ def report_run_failed(self, pipeline_run, message=None):
             "This run has been marked as failed from outside the execution context.",
         )
 
+        run = self.get_run_by_id(pipeline_run.run_id)
+        if run is None:
+            raise DagsterInvariantViolationError(
+                f"Could not load run {pipeline_run.run_id} that was passed to report_run_failed"
+            )
+        if run.status not in IN_PROGRESS_RUN_STATUSES:
```
brad-alexander, line 1626: I'm not confident that this same check is appropriate for guarding state transition to failed.  I could see this just checking for success if transitions from other statuses should be allowed.

johannkm: Runs can fail from `NOT_STARTED`, `QUEUED`, and `MANAGED`

python_modules/dagster/dagster/daemon/monitoring/monitoring_daemon.py
```diff
@@ -61,7 +61,7 @@ def monitor_started_run(instance: DagsterInstance, workspace, run, logger):
             if instance.run_launcher.supports_resume_run:
                 msg = (
                     f"Detected run worker status {check_health_result}. Marking run {run.run_id} as "
-                    "failed, because it has surpassed the configured maximum attempts to resume the run: {max_resume_run_attempts}."
```
johannkm, line 64: oops...

johannkm (changes requested):
Thanks for jumping on this! Should we reload the run to check status before reporting a failure? I think there still could be a race condition with reporting a failed run as failure twice (daemon collects in progress runs -> run process fails, but is able to log the failure -> daemon checks on run worker, reports it as failed). Not as bad as the condition this fixes though

python_modules/dagster/dagster/daemon/monitoring/monitoring_daemon.py
```diff
@@ -61,7 +70,8 @@ def monitor_started_run(instance: DagsterInstance, workspace, run, logger):
...
+                    f"failed, because it has surpassed the configured maximum attempts to resume the "
```
brad-alexander, line 73:
```suggestion
                    "failed, because it has surpassed the configured maximum attempts to resume the "
```

python_modules/dagster/dagster/core/instance/__init__.py
brad-alexander: changed to just recheck during monitoring loop

johannkm (approved):

PR #8732, opened 2022-07-04
Add MetaDataEntry.path to PickledObjectS3IOManager

danielgafni:
### Summary & Motivation
I noticed this was missing and fixed this little issue.
The uri now can be observed in the Asset view, too. 

Running Dagster locally with these modifications produces the desired result:

image

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
@@ -2,7 +2,7 @@ import io
 import pickle
 from typing import Union
 
-from dagster import Field, InputContext, MemoizableIOManager, MetadataEntry, OutputContext, StringSource
+from dagster import Field, InputContext, MemoizableIOManager, MetadataValue, OutputContext, StringSource
 from dagster import _check as check
 from dagster import io_manager
 from dagster.utils import PICKLE_PROTOCOL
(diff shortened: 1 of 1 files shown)


danielgafni:
Or maybe this should be a `MetadataEntry.url` ?

prha:
Yeah, I think `MetadataEntry.url` would be preferred here.

cc @smackesey / @sryza  since I see yielding `MetadataEntry` is marked for deprecation but we will probably want a carve-out for the IO manager use case.

sryza:
@prha @danielgafni - the preferred invocation would be
```
context.add_output_metadata("metadata_label": MetadataValue.url(...))
```

danielgafni:
I see, thanks. I didn't find anything about the deprecation in the docs (although it was obvious from the code). The docs have a section about using MetaDataEntry. Probably this should be fixed?

danielgafni:
I'm thinking this should be a `path`, not a `url`. A url is clickable, while a path can be copied to clipboard on click. This (copying) is exactly what we need for an s3 path.

sryza:
> I'm thinking this should be a path, not a url. A url is clickable, while a path can be copied to clipboard on click. This (copying) is exactly what we need for an s3 path.

That reasoning makes sense to me.

> I see, thanks. I didn't find anything about the deprecation in the docs (although it was obvious from the code). The docs have a section about using MetaDataEntry. Probably this should be fixed?

Thanks for reporting this. I'll file an issue.

sryza:
@danielgafni it looks like all the tests passed, but the build is complaining about some code style and import ordering. If you run `make isort` and `make black` from the repo root, it should fix these.

sryza (changes requested):
Marking this "Changes requested" a la my previous review.

danielgafni:
@sryza sorry, done

PR #8734, opened 2022-07-04
[Run Job Perf] Receive logs in webworker and send to main thread using Transferable

salazarm:
### Summary & Motivation

On heavy jobs, receiving logs for the `PipelineRunLogsSubscription` ends up taking a lot of CPU cycles on the main thread leading to poor user experience due to lack of idle time. To combat this we move the websocket to a webworker so that the data is received in a separate thread. We then transfer the data to the main-thread using an `ArrayBuffer` and parse the JSON on the main-thread.

**Benefits:**
- ~33% Decrease in CPU use on the main thread.
- Websocket messages can be received in parallel with UI rendering. While loading the [500mb debug file here](https://linear.app/elementl/issue/DAGIT-25/run-page-performance-w-large-jobs) the worker version was able to finish loading all of the logs in about 40% less time than the non worker version (~90 seconds vs ~150 seconds)

Implementation:
It's a relative simple implementation in that the webworker just passes the messages straight to the main thread as soon as it receives them. If the UI thread is too busy to process them immediately it leads to bigger batches being processed at a time by the main thread as the messages accumulate.

1. Loaded the [500mb debug file here](https://linear.app/elementl/issue/DAGIT-25/run-page-performance-w-large-jobs)
2. Modified many_events_job to materialize 10,000 files and validate 10,000 tables and profiled with and without the worker 10 times each and saw consistently a 33% decrease in the main thread being busy.

worker:
Screen Shot 2022-07-04 at 3 33 13 PM

No worker:
Screen Shot 2022-07-04 at 3 31 28 PM


js_modules/dagit/packages/core/src/workers/util.ts
```diff
@@ -0,0 +1,25 @@
...
+export function arrayBufferToString(buf: ArrayBuffer): string {
+  let binary = '';
+  const bytes = new Uint16Array(buf);
+  const len = bytes.byteLength;
+  for (let i = 0; i < len; i++) {
+    const char = String.fromCharCode(bytes[i]);
+    if (char !== '\x00') {
+      binary += char;
+    }
+  }
```
bengotow, line 12: Hmm this seems like it could be a pretty big bottleneck, is there a particular field in our JSON that contains null characters? It seems like removing them during both serialization and deserialization is unnecessary. (Nice for completeness, but if they're always used together it'd be safe to clean only the serializer?)

If we could reduce this entire function to this line, I think it'd be a big win because we can avoid having to concatenate onto a string over and over:

```
String.fromCharCode.apply(null, new Uint16Array(buf))
```

bengotow (approved):
This looks great so far! I added one inline comment, overall it's super cool this is working! I'm not 100% up-to-date on our Apollo Client configuration, but it's possible that to deploy this we'd also need to copy some base path / auth token type of stuff that is applied in cloud? @hellendag might know more about that.

Maybe we should try to layer on one more phase of this with a perf win before merging it?

PR #8735, opened 2022-07-05
[Job Log perf] Use rAF to call throttleSetNodes

salazarm:
### Summary & Motivation

There is a performance edge cases where `throttleSetNodes`'  does not end up batching logs. This happens when react rendering takes as long or longer than the `BATCH_INTERVAL` of 100ms. In that case, by the time we're done rendering, the BATCH_INTERVAL time has elapsed, and the next call to `throttleSetNodes` ends up causing another re-render before any other calls can be batched with it. To get around this problem we're using `requestAnimationFrame` so that at least all calls within a single animation frame end up getting batched together. This ends up being important for jobs with extremely high parallelism as they can have multiple messages arriving at the same time for each thread the job is executing on. 

I also looked into setting `{leading: false}` as an option to `throttle` so that it fires after `100ms` and not at the start but that doesn't solve the problem and we end up in the same situation.

I modified many_events to have a lot of parallelism and set `max_concurrent = 36`

I took a bunch of chrome performance profiles and added console.log's to verify that websocket messages within the same animation frame are actually getting batched together.

This is an example of receiving 23 messages within 100ms:
Screen Shot 2022-07-05 at 6 11 36 AM

Before this change those 23 messages would lead to 23 consecutive react renders and cause the frames to be much longer: (Note all of the red corners indicating long frames)
Screen Shot 2022-07-05 at 6 00 23 AM

After- Very few frames have red corners:
Screen Shot 2022-07-05 at 5 59 43 AM

In my profiling I found that what  pushes rendering over the 100ms mark is the Gantt chart. If we could virtualize it to minimize its rendering cost that would help out a lot.

### DIFF
--- a/js_modules/dagit/packages/core/src/runs/LogsProvider.tsx
+++ b/js_modules/dagit/packages/core/src/runs/LogsProvider.tsx
@@ -168,15 +168,13 @@ const useLogsProviderWithSubscription = (runId: string) => {
       }
       const queuedLogs = [...queue.current];
       queue.current = [];
-      const queuedMessages = ([] as RunDagsterRunEventFragment[]).concat(
-        ...queuedLogs.map((log) => log.messages),
-      );
+      const queuedMessages = queuedLogs.flatMap((log) => log.messages);
       const lastLog = queuedLogs[queuedLogs.length - 1];
       const hasMore = lastLog.hasMorePastEvents;
       const cursor = lastLog.cursor;
 
       dispatch({type: 'append', queued: queuedMessages, hasMore, cursor});
-      const nextPipelineStatus = pipelineStatusFromMessages(lastLog.messages);
+      const nextPipelineStatus = pipelineStatusFromMessages(queuedMessages);
 
       // If we're still loading past events, don't sync to the cache -- event chunks could
       // give us `status` values that don't match the actual state of the run.
--- a/js_modules/dagit/packages/core/src/runs/LogsProvider.tsx
+++ b/js_modules/dagit/packages/core/src/runs/LogsProvider.tsx
@@ -13,6 +13,7 @@ import {LogNode} from './types';
 import {
   PipelineRunLogsSubscription,
   PipelineRunLogsSubscriptionVariables,
+  PipelineRunLogsSubscription_pipelineRunLogs_PipelineRunLogsSubscriptionSuccess,
 } from './types/PipelineRunLogsSubscription';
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/core/src/runs/LogsProvider.tsx
```diff
@@ -159,12 +162,29 @@ const useLogsProviderWithSubscription = (runId: string) => {
...
+    return throttle(() => {
+      if (!queue.current.length) {
+        return;
+      }
+      const queuedLogs = [...queue.current];
       queue.current = [];
-      dispatch({type: 'append', queued, hasMore, cursor});
+      const queuedMessages = ([] as RunDagsterRunEventFragment[]).concat(
+        ...queuedLogs.map((log) => log.messages),
+      );
```
bengotow, line 173: Minor nit but I think this line is just doing `queuedLogs.flatMap(log => log.messages)` and we have that now that we're targeting ES2019?

js_modules/dagit/packages/core/src/runs/LogsProvider.tsx
```diff
@@ -159,12 +162,29 @@ const useLogsProviderWithSubscription = (runId: string) => {
...
+      const cursor = lastLog.cursor;
+
+      dispatch({type: 'append', queued: queuedMessages, hasMore, cursor});
+      const nextPipelineStatus = pipelineStatusFromMessages(lastLog.messages);
+
+      // If we're still loading past events, don't sync to the cache -- event chunks could
+      // give us `status` values that don't match the actual state of the run.
+      if (nextPipelineStatus && !hasMore) {
+        syncPipelineStatusToApolloCache(nextPipelineStatus);
+      }
```
bengotow, line 185: This may have been a bug before, but I feel like there's a weird edge case here in which the "last pipeline status event" is NOT in the last "batch" seen by `throttle`, which means these two arguments are never `true` at the same time and the latest pipeline status is not set?

It might be better to say `if (!hasMore) { syncPipelineStatusToApolloCache(<<all messages>>) }` to set the pipeline status instead of just considering the last batch.

bengotow (approved):
This looks great to me!

js_modules/dagit/packages/core/src/runs/LogsProvider.tsx
salazarm: Yeah it is, good call!

Issue #8707, opened 2022-07-01
Dagster should error when non-asset-producing ops are leaf nodes in a graph-backed asset


OwenKephart:
### Dagster version

0.15.3

### What's the issue?

Definitely a mouthful, but the only leaf nodes of a graph-backed asset should be ops that produce assets. If we detect that this is not the case, we should error early, rather than surface confusing behavior in the UI.

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

PR #8754, opened 2022-07-05
[assets] add validation for graph backed asset graphs

OwenKephart:
### Summary & Motivation

fixes: https://github.com/dagster-io/dagster/issues/8707

It is invalid for a graph-backed asset to have leaf nodes that are ops. Before this PR, it was possible to do this without getting an error at definition time. If you tried to materialize an asset in dagit, only the ops that were on the critical path of updating that asset would be executed, meaning those leaf ops would be silently excluded, which could cause confusion.

pytest

### DIFF
--- a/python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets_job.py
+++ b/python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets_job.py
@@ -1080,27 +1080,35 @@ def test_twice_nested_graph():
     def transformer(string):
         return string + "qux"
 
-    @graph(out={"n1": GraphOut(), "n2": GraphOut()})
+    @op
+    def combiner(s1, s2):
+        return s1 + s2
+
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets_job.py
```diff
@@ -1081,27 +1081,35 @@ def innermost_thing():
...
-        return {"n1": n1, "n2": n2}
+        unused_output = combiner(unused_output, transformer(foo_asset))
+        return {"n1": n1, "n2": n2, "unused": unused_output}
 
     @asset
     def foo_asset():
         return "foo"
 
-    thing_asset = AssetsDefinition(
+    thing_asset = AssetsDefinition.from_graph(
```
OwenKephart, line 1105: the graph created in this test was invalid (because it had unused outputs), so this needed to updated.

clairelin135 (approved):
This looks great to me! Thanks for making this change

PR #8755, opened 2022-07-06
Refactor op return checking code

dpeng817:
This PR refactors the checks around returns in ops. There was a whole host of things going weird with this stuff; 
- some of the if statements were impossible to enter
- the new returnable output / dynamic output stuff had a bunch of weird corner cases (being able to return the underlying type without returning the Output object itself)
- https://github.com/dagster-io/dagster/issues/8565
- Invocation pathway had a completely different series of checks for returns
- The way we expect an output to look often is dependent upon what the return annotation of the decorated function is. But the logic checking the output was not referring back to the annotation.

This unifies the code checking pathway between invocation and real execution (at least for returns) and simplifies the return logic checking. The result is less weird corner cases slipping through the cracks, and hopefully less confused users with error messages unrelated to their problem.
**Important to note** that this PR involves a lot of small breaking changes of undocumented behavior:
- a dynamic op with no return or yield will now error
- returning a single `Output` object without annotation will now error
- Returning the underlying type not wrapped in an output object will now error if you used generic output annotation


python_modules/dagster/dagster/core/definitions/decorators/solid_decorator.py
```diff
@@ -56,6 +56,11 @@ def has_var_kwargs(self) -> bool:
         # var keyword arg has to be the last argument
         return len(params) > 0 and param_is_var_keyword(params[-1])
 
+    def get_output_annotation(self) -> Any:
+        from ..inference import infer_output_props
+
+        return infer_output_props(self.decorated_fn).annotation
+
```
dpeng817, line 63: This allows us to refer back to the annotation of the underlying function from the definition itself.

python_modules/dagster/dagster/core/definitions/solid_definition.py
```diff
@@ -213,6 +213,19 @@ def is_graph_job_op_node(self) -> bool:
...
+
+        return isinstance(self._compute_fn, DecoratedSolidFunction)
+
+    def get_output_annotation(self) -> Any:
+        if not self.is_from_decorator():
+            raise DagsterInvalidInvocationError(
+                f"Attempted to get output annotation for {self.describe_op()}, "
+                "which was not constructed from a decorated function."
+            )
+        return self.compute_fn.get_output_annotation()
```
dpeng817, line 227: found myself often repeating the pattern of cast(DecoratedSolidFunction, solid_def.compute_fn). This is just a clearer way of doing that, and can probably be used in a bunch more places (grep `cast(DecoratedSolidFunction` in the codebase and you'll see what I mean)

python_modules/dagster/dagster/core/definitions/solid_invocation.py
```diff
@@ -289,59 +289,14 @@ def type_check_gen(gen):
...
-    else:
-        _type_check_output(solid_def.output_defs[0], result, context)
-        received_output = solid_def.output_defs[0].name
-
-    for output_def in solid_def.output_defs:
-        if output_def.is_required and not output_def.name == received_output:
-            raise DagsterInvariantViolationError(
-                f"Invocation of {solid_def.node_type_str} '{solid_def.name}' did not return an output for non-optional "
-                f"output '{output_def.name}'."
-            )
```
dpeng817, line 344: bye bye

python_modules/dagster/dagster/core/execution/plan/compute_generator.py
```diff
@@ -58,151 +70,151 @@ def compute(context, input_defs) -> Generator[Output, None, None]:
...
+            raise DagsterInvariantViolationError(
+                f"{context.describe_op()} has multiple outputs, but only one "
+                f"output was returned of type {type(result)}. When using "
+                "multiple outputs, either yield each output, or return a tuple "
+                "containing a value for each output. Check out the "
+                "documentation on outputs for more: "
+                "https://docs.dagster.io/concepts/ops-jobs-graphs/ops#outputs."
+            )
+    else:
+        yield 0, output_defs[0], result
```
dpeng817, line 104: zip up the outputs with their accompanying output defs so that output checking logic can be reused across the single and multi output case (before they were split).

python_modules/dagster/dagster/core/execution/plan/compute_generator.py
```diff
@@ -58,151 +70,151 @@ def compute(context, input_defs) -> Generator[Output, None, None]:
...
+
+def _get_annotation_for_output_position(
+    position: int, solid_def: SolidDefinition, output_defs: Sequence[OutputDefinition]
+) -> Any:
+    if solid_def.is_from_decorator():
+        if len(output_defs) > 1:
+            return get_args(solid_def.get_output_annotation())[position]
+        else:
+            return solid_def.get_output_annotation()
+    return NoAnnotationSentinel()
```
dpeng817, line 115: this check kept getting re-used, so I split it out

python_modules/dagster/dagster_tests/core_tests/definitions_tests/decorators_tests/test_op.py
```diff
@@ -1137,11 +1136,18 @@ def basic() -> List[DynamicOutput]:
     def basic_yield():
         pass
 
-    result = execute_op_in_graph(basic_yield)
-    assert result.success
+    with pytest.raises(
+        DagsterInvariantViolationError,
+        match=r"dynamic output 'result' expected a list of DynamicOutput objects, but instead received instead an object of type \<class 'NoneType'\>\.",
+    ):
+        execute_op_in_graph(basic_yield)
```
dpeng817, line 1143: new error state: previously, an op with a non-optional dynamic output and no return would silently complete. This now results in a failure.

python_modules/dagster/dagster_tests/core_tests/definitions_tests/decorators_tests/test_op.py
```diff
@@ -1137,11 +1136,18 @@ def basic() -> List[DynamicOutput]:
...
 
     # Ensure that invocation behavior matches
-    basic_yield()
+    with pytest.raises(
+        DagsterInvariantViolationError,
+        match=r"dynamic output 'result' expected a list of DynamicOutput objects, but instead received instead an object of type \<class 'NoneType'\>\.",
+    ):
+        basic_yield()
```
dpeng817, line 1150: invocation behavior same as [1]

python_modules/dagster/dagster_tests/core_tests/definitions_tests/decorators_tests/test_op.py
```diff
@@ -1198,3 +1212,52 @@ def the_job():
...
+        "has generic output annotation, but did not receive an Output object "
+        "for this output. Received instead an object of type <class 'int'>.",
+    ):
+        basic_multi_output()
+
+
+def test_list_out_op():
+    @op(out={"list_out": Out(List[str]), "other_out": Out(int)})
+    def test_op() -> Tuple[List[str], int]:
+        return ([], 5)
```
dpeng817, line 1258: list case from https://github.com/dagster-io/dagster/issues/8565

python_modules/dagster/dagster_tests/core_tests/test_solid_invocation.py
```diff
@@ -496,7 +490,17 @@ def test_optional_output_return():
...
+        DagsterInvariantViolationError,
+        match="solid \"solid_multiple_outputs_not_sent\" has multiple outputs, but only one output was returned of type <class 'dagster.core.definitions.events.Output'>.",
+    ):
+        solid_multiple_outputs_not_sent()
+
+    with pytest.raises(
+        DagsterInvariantViolationError,
+        match="solid \"solid_multiple_outputs_not_sent\" has multiple outputs, but only one output was returned of type <class 'dagster.core.definitions.events.Output'>.",
+    ):
+        execute_solid(solid_multiple_outputs_not_sent)
```
dpeng817, line 503: similar to [2]

python_modules/dagster/dagster_tests/core_tests/definitions_tests/decorators_tests/test_op.py
```diff
@@ -841,55 +841,38 @@ def test_output_generic_correct_inner_type():
...
-
-    result = the_op_annotation_not_using_output()
-    assert isinstance(result, Output)
-    assert result.value == 42
-
-
-def test_output_generic_type_mismatches():
-    @op
-    def the_op_annotation_type_mismatch() -> int:
-        return Output("foo")
```
alangenfeld, line 864: why were these tests removed?

alangenfeld (commented):
looks like tests are failing? 

consider being less aggressive on the pytest match text content, its nice to not have to update a bunch of tests if for example we moved some code.

python_modules/dagster/dagster_tests/core_tests/definitions_tests/decorators_tests/test_op.py
dpeng817: They were testing a bunch of type mismatch behavior that is now irrelevant since we now error in a much broader case.

dpeng817:
Turns out we use the `return Output(...)` pattern a bunch in tests, which is giving me pause about making it not work in all cases. Perhaps there are still scenarios it's reasonable to keep it working:
- No type annotation is provided, you are using a single output, and you return an Output object. This should probably still work.
- No type annotation is provided, you are using a single output, and you return a list of DynamicOutput objects. This should also probably still work.

sryza:
> No type annotation is provided, you are using a single output, and you return an Output object. This should probably still work.

I agree that this should still work

alangenfeld (approved):
just want to make sure the empty list / empty iterator still works for dynamic outputs, also would be nice to clean up `basic_yield` since thats named inaccurately

dpeng817:
> is a dynamic out that is actually an iterator (has yield keyword) but returns no results still work? I think that should still work
Do you mean things other than `DynamicOutput` get yielded, but no DynamicOutputs get yielded? If so, that still works. Adding a test for that.

> empty list still works too right?
Yes that's under test

PR #8756, opened 2022-07-06
Restore sensor daemon yielding when evaluating sensors synchronously

prha:
### Summary & Motivation
Large sensors yielding a lot of run requests can cause the sensor daemon to stop heartbeating.  This PR restores the yielding behavior for the synchronous evaluation, but perform the full evaluation when delegating to a thread pool.

BK

### DIFF
--- a/python_modules/dagster/dagster/daemon/sensor.py
+++ b/python_modules/dagster/dagster/daemon/sensor.py
@@ -359,19 +359,32 @@ def execute_sensor_iteration(
         elif _is_under_min_interval(sensor_state, external_sensor, now):
             continue
 
-        future = executor.submit(
-            _process_tick,
-            logger,
-            instance,
-            workspace,
-            now,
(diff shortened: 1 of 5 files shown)


python_modules/dagster/dagster/daemon/sensor.py
```diff
@@ -359,8 +359,46 @@ def execute_sensor_iteration(
...
+                workspace,
+                now,
+                external_sensor,
+                sensor_state,
+                sensor_debug_crash_flags,
+                tick_retention_settings,
+            )
+            yield
+            if debug_futures is not None:
+                debug_futures[external_sensor.selector_id] = future
```
alangenfeld, line 387: as per DMs - with this "unified interface" breaking down just delete sync executor and embrace the fork assuming thats reasonable

alangenfeld (approved):
nod

Issue #8705, opened 2022-07-01
Cross-repo asset dependencies surface SourceAssets, confuse Dagit


bengotow:
### Dagster version

015.3

### What's the issue?

hey gang! I’m seeing some undesirable behavior in our root workspace where there are cross-repo asset graphs and I’d be interested thoughts about a potential fix. The problem is that calls to asset_node.dependencies { } in graphQL still return the source asset (in the repo) and not the real asset (in a different repo)

https://www.loom.com/share/004a470fad734fe79e73ced950eeb11a

This breaks some GraphQL assumptions because the real asset node and the source asset node both have the same “id” field value and overwrite each other in the Apollo cache (but one has op_names, job_names, etc and the other doesn’t.)

### What did you expect to happen?

I think we could:

- Change the dependencies / dependedBy resolvers to fully resolve source asset nodes (so source asset nodes would not reach Dagit unless the defining repository was not loaded)
- Change the asset_node.id so that it includes the repo + repo location, avoiding cache collisions between two copies of the same asset node

### How to reproduce?

This is reproducible in our root workspace, see attached video!

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

clairelin135:
Thanks for sharing this @bengotow! I think that I'd prefer to change the python implementation to fully resolve source asset nodes rather than changing asset_node.id because `GrapheneAsset`s also contain an ID but may not be defined in a repo. To keep things consistent between `GrapheneAssetNode` and `GrapheneAsset` it's probably easiest to fully resolve the asset.

I'll put out a diff to fix this!

clairelin135:
Actually, on second thought, fully resolving source assets would force unnecessarily loading all repository locations on cloud upon page load. It's probably best to avoid this--I'll put out a diff to modify the IDs instead.

PR #8762, opened 2022-07-06
Update asset ID to contain repository location and name

clairelin135:
### Summary & Motivation

Resolving issue https://github.com/dagster-io/dagster/issues/8705

This PR modifies the `id` field of `GrapheneAsset` and `GrapheneAssetNode` to contain repository location name and repository name (if exists). This ensures that there will not be collisions when an asset is loaded as a derived asset in one repository and a source asset in another repository.

I searched through the instances where the `id` field is used in the front end and found one spot where the implementation needed to be changed, let me know how my changes look!

Locally in Dagit

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/AssetDefinedInMultipleReposNotice.tsx
+++ b/js_modules/dagit/packages/core/src/assets/AssetDefinedInMultipleReposNotice.tsx
@@ -11,6 +11,10 @@ export const AssetDefinedInMultipleReposNotice: React.FC<{
   assetId: string;
   loadedFromRepo: RepoAddress;
 }> = ({assetId, loadedFromRepo}) => {
+  function get_asset_key_from_id(id: string) {
+    return id.split('.').pop();
+  }
+
   const {data} = useQuery<AssetIdScanQuery>(ASSET_ID_SCAN_QUERY);
   const otherRepos =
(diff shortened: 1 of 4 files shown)


js_modules/dagit/packages/core/src/assets/AssetDefinedInMultipleReposNotice.tsx
```diff
@@ -11,16 +11,21 @@ export const AssetDefinedInMultipleReposNotice: React.FC<{
   assetId: string;
   loadedFromRepo: RepoAddress;
 }> = ({assetId, loadedFromRepo}) => {
+  function get_asset_key_from_id(id: string) {
```
prha, line 14: do we need to do this?

My preference is just have id be some unique string and not try to parse it (it's used internally for caching in the apollo cache).  Instead, we just add `assetKey` the asset node field and use that instead.

prha (changes requested):

js_modules/dagit/packages/core/src/assets/AssetDefinedInMultipleReposNotice.tsx
clairelin135: That makes a lot of sense, I'll update this to use asset key instead 👍

prha (approved):

PR #8765, opened 2022-07-06
Document valid names for asset keys

### LINKED ISSUE #7847: in SDA keys, only allow characters that can be included in definition names
I.e. no slashes, etc.

We already effectively do this, because we convert the keys to an op name. However, we should document it and raise a clearer error.

jamiedemaria:
### Summary & Motivation
adds docs for valid asset key names

resolves #7847


jamiedemaria:
@sryza the original issue says to raise a clearer error. This is the current error 
```
UserWarning: Error loading repository location dagster_test.graph_job_op_toys.repo:dagster.core.errors.DagsterInvalidDefinitionError: "this/is/invalid__daily_temperature_highs" is not a valid name in Dagster. Names must be in regex ^[A-Za-z0-9_]+$.
```

Were you thinking that the wording of this should be changed (would affect all ops since this is raised in node definition) or that we should check name validity earlier in the stack (like in _Asset) and raise an asset specific error message?

python_modules/dagster/dagster/core/asset_defs/decorators.py
```diff
@@ -106,12 +106,16 @@ def asset(
 
     Args:
         name (Optional[str]): The name of the asset.  If not provided, defaults to the name of the
-            decorated function.
+            decorated function. The asset's name must be a valid name in dagster (ie only contains
+            letters, numbers, and _) and may not contain python reserved keywords.
         namespace (Optional[Sequence[str]]): **Deprecated (use `key_prefix`)**. The namespace that
-            the asset resides in.  The namespace + the name forms the asset key.
+            the asset resides in.  The namespace + the name forms the asset key. The asset key must
```
sryza, line 112: This can be a sequence, right? I think this description is a little bit confusing with respect to that.

sryza (changes requested):

sryza (approved):
Nice

PR #8768, opened 2022-07-06
partition mappings on graph-backed assets

sryza:
### Motivation

https://dagster.slack.com/archives/C01U954MEER/p1657107439141239


python_modules/dagster/dagster/core/asset_defs/assets.py
```diff
@@ -395,6 +417,8 @@ def get_partition_mapping(self, in_asset_key: AssetKey) -> PartitionMapping:
         if self._partitions_def is None:
             check.failed("Asset is not partitioned")
 
+        print(self._partition_mappings)
```
clairelin135, line 420: need to remove these

clairelin135 (approved):
LGTM

PR #8774, opened 2022-07-06
Add partition key to build_output_context, add documentation for partition_key on build_op_context

dpeng817:


python_modules/dagster/dagster/core/execution/context/output.py
```diff
@@ -307,7 +312,7 @@ def partition_key(self) -> str:
                 "For more details: https://github.com/dagster-io/dagster/issues/7900"
             )
 
-        return self.step_context.partition_key
+        return self._partition_key
```
sryza, line 315: In the main execution case (not `build_output_context`), I'm not seeing how we still get the partition key. Maybe I'm missing something?

dpeng817: arf I meant to make this a conditional

python_modules/dagster/dagster/core/execution/context/output.py
```diff
@@ -291,7 +293,10 @@ def has_partition_key(self) -> bool:
                 "For more details: https://github.com/dagster-io/dagster/issues/7900"
             )
 
-        return self.step_context.has_partition_key
+        if self._step_context:
```
sryza, line 296: Just a thought: might it be better to do this branching inside the constructor? Then it would only need to happen once, instead of every time these methods are called, and it would only need to be coded in one place?

sryza (approved):

python_modules/dagster/dagster/core/execution/context/output.py
dpeng817: good idea

dpeng817: hm on second thought - early access of partition key methods leads to partition errors when there are not partitions on the given run.

dpeng817:
cleaning my queue for now

python_modules/dagster/dagster/core/execution/context/output.py
dpeng817: hrm I guess. it checks tags, not key, which seems weird.

PR #8800, opened 2022-07-08
Top-level export adjustments: drop deprecations, fix lint errors, replace __all__ 

smackesey:
### Summary & Motivation

This PR does several things. They are related but can be separated if necessary:

- Fixes user lint errors for dagster imports
- Removes __all__ in favor of redundant alias imports
- Drops pep562 dep
- Adds `ExperimentalWarning` as a public export

## Fixes user lint errors for dagster imports

Discovered during dogfooding: Pyright/pylance currently raise warnings for a good deal of our top-level imports. The reason for this is that the way they are imported in `dagster/__init__.py` registers them as private-- e.g. the decorator `op` is imported from `dagster.core.definitions`, but isn't part of that submodules "internal public" API, because imported symbols (it's imported into `dagster.core.definitions` from `dagster.core.definitions.decorators`)  aren't part of the "internal public" API by default. If the chain of imports leading to a definition has a single private link, it causes an import error for the user.

This PR fixes that by ensuring that there are no private links in the chains leading to our top-level imports. This was done by making imports in top-level `dagster` come directly from the module where the imported symbol is defined.

## Removes top-level `__all__` in favor of redundant alias imports

This PR gets rid of `__all__` in favor of redundant aliased imports. `__all__` is confusing because many people think it defines a module's public API, but it does not-- it only restricts what is exposed by `from X import Y`. It is also used by some type checkers to override privacy rules for `_` symbols.

Redundant aliased imports have basically the same effect as `__all__`,  while cutting the length of `dagster/__init__.py` in half and making the list of top-level exports easier to manage by co-locating the duplicate name with the original import (there have been several instances where something intended to be exported was accidentally left out of `__all__`. The change should be invisible to users as well-- anyone doing `from dagster import *` (discouraged anyway, but this is what `__all__` affects) will still get all of dagster's symbols.

To make the redundant alias imports more manageable, I ran `isort` with a short line length on `dagster/__init__.py` to get one import per-line, then used `# isort: off/on` comments to turn off isort for the main import block in `dagster/__init__.py`.

## Drops pep562 dep

This is needed to override `__getattr__` and `__dir__` (used for dynamic imports) on 3.6, but we're dropping 3.6 so we don't need it anymore.

## Adds `ExperimentalWarning` as a public export

I believe `ExperimentalWarning` needs to be public since users need to use it to filter out Dagster's experimental warnings.

(1) BK
(2) I pulled the `__all__` list off of master, cleaned/sorted it, and compared it to the cleaned/sorted list of exports from this module to ensure the lists were identical (with the exception of `ExperimentalWarning`)


smackesey:
* master
  * **PR #8800** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8800" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8800?utm_source=stack-comment).

sryza:
This looks good aside from the comment I left

sryza (approved):

PR #8817, opened 2022-07-11
[hacker news] add missing key prefix to activity analytics python assets

sryza:
I missed these when I did my post-0.15.0 hacker news refactor


OwenKephart (approved):
lgtm

PR #8820, opened 2022-07-11
Deprecate job_selection and pipeline_run_status from run status sensors

jamiedemaria:
### Summary & Motivation
removes pipeline_failure_sensor and the following params to run status sensors that have been renamed:
`pipeline_selection` -> removed
`pipeline_run_status` -> `run_status`
`job_selection` -> `monitored_jobs`

has the happy side effect of getting the `dagster-msteams` library up to post-crag standards


sryza (commented):
I've been waffling on this one.  I think we should definitely get rid of pipeline_failure_sensor and all the arguments that have pipeline in them.  However, the `job_selection` change is going to break a bunch of people, and we only deprecated it pretty recently.  I'm open to convincing, but my instinct would be to keep it around (as deprecated) through 1.x.  The cost of keeping it is low, and it could save us from a sizable chunk of ill will.

jamiedemaria:
@sryza right, I spaced that we talked about that. I'm totally fine keeping it, just went on autopilot removing stuff. I can add it back in, but do you think there's a target version I should update the deprecation message to (1.2, 1.3, etc)?

sryza:
> I can add it back in, but do you think there's a target version I should update the deprecation message to (1.2, 1.3, etc)

We won't be able to remove deprecated APIs until 2.0, which I think is OK in this case.  Pandas kept some deprecated APIs around for several years.  I believe Windows has done it for decades.

sryza (approved):
Nice

PR #8824, opened 2022-07-11
provide description for MAP ConfigType

Jiafi:
### Summary & Motivation
Was getting an error trying to build dagster-docs with a `Map` type in my `config_schema`.  Realized it is because there is no `type_repr` for `MAP`.  

Not sure if this is the proper `type_repr`.  Gonna leave that up to the dagster team
Properly builds docs with the changes.


yuhan:
kicking off test.

this looks reasonable to me. cc @smackesey do you happen to know if there is a better `type_repr` to use in this case?

prha (approved):

PR #8825, opened 2022-07-11
fix gql resolver for graph-backed assets resources

smackesey:
### Summary & Motivation

Fixes a bug in the resolver for required resource keys of graph-backed assets. The issue was calling a misspelled method that does not exist. Trying to figure out why `mypy` did not catch this.

- Reproed issue, eliminated issue after change

Since we are doing a patch release today or tomorrow, putting this out without an additional test, but some tests are coming in a follow-up PR.


clairelin135 (approved):

prha:
@smackesey can we add a graphql test that requests this field?

PR #8829, opened 2022-07-11
mark dagster.{api,cli,config} private

smackesey:
### Summary & Motivation

Marks `dagster.api`, `dagster.cli`, and `dagster.config` private. Part of incremental adoption of privacy for all submodules. Adds sys.meta_path mappings so that imports of e.g. `dagster.api` are mapped to `dagster._api`.

Changes were done by:

- Using pycharm refactor to perform automated module rename
- Doing associated manual renames in the docs

I also created an "internal public" API for `dagster._config` by exposing every symbol imported from `dagster._config` in `dagster._config.__init__` and converting all imports of those symbols to come from `dagster._config` instead of a submodule. Moving towards this kind of internal organization should keep our imports more comprehensible and reduce the likelihood of changes triggering circular imports.

As part of this transition, I needed to factor the config-specific portions of `dagster.primitive_mapping` into a new _config submodule-- without this change, one ends up with a circular import that starts with `dagster.dagster_type` (also accessed from `dagster.primitive_mapping`

Requires a sister PR in internal: https://github.com/dagster-io/internal/pull/2583

BK.


alangenfeld (approved):
I assume this as an automated change? Would be good to describe how it was done if so in summary. 

Were there any manual fix-ups? Skimming through this looks reasonable but I did not go line by line through the whole thing.

dpeng817 (approved):
Looks good! Nice one on catching the docs changes here.

PR #8835, opened 2022-07-11
Retrieve minimal set of asset records for assetsLatestInfo

bengotow:
### Summary & Motivation

This PR makes two small changes to the GraphQL resolver for assetsLatestInfo to resolve performance issues:

- Rather than retrieve the latest materialization of all asset nodes, we traverse the asset node graph and load only the materializations that we will need to compute upstream changed.

- If we fetch the list of asset nodes required and it's empty (either because there are no SDAs or because the step_keys_by_asset array passed in was empty), exit early to avoid calling `get_asset_records` with `[]`, which it interprets as "All"

I tested these changes manually by logging the asset records that were being fetched as I loaded different views in Dagit, confirming that it was just the upstream assets.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/implementation/fetch_runs.py
+++ b/python_modules/dagster-graphql/dagster_graphql/implementation/fetch_runs.py
@@ -134,6 +134,26 @@ class AssetComputeStatus(Enum):
     OUT_OF_DATE = "OUT_OF_DATE"
 
 
+def add_all_upstream_keys(
+    all_asset_nodes: Mapping[AssetKey, "GrapheneAssetNode"],
+    requested_asset_keys: KeysView[AssetKey],
+):
+    required: Dict[AssetKey, bool] = {}
+
+    def append_key_and_upstream(key: AssetKey):
+        if required.get(key):
+            return
+        required[key] = True
+        asset_node = all_asset_nodes[key].external_asset_node
+        for dep in asset_node.dependencies:
+            append_key_and_upstream(dep.upstream_asset_key)
+
+    for asset_key in requested_asset_keys:
+        append_key_and_upstream(asset_key)
+
+    return required.keys()
+
+
 def get_upstream_changed_by_asset(
     all_asset_nodes: Mapping[AssetKey, "GrapheneAssetNode"],
     all_asset_records: Iterable[AssetRecord],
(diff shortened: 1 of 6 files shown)


python_modules/dagster-graphql/dagster_graphql/implementation/fetch_runs.py
```diff
@@ -196,15 +216,19 @@ def fill_or_return_status_for(key: AssetKey):
...
     from ..schema.pipelines.pipeline import GrapheneRun
 
     instance = graphene_info.context.instance
 
     asset_nodes = get_asset_nodes_by_asset_key(graphene_info)
-    asset_records = instance.get_asset_records(asset_nodes.keys())
+    asset_record_keys_needed = add_all_upstream_keys(asset_nodes, step_keys_by_asset.keys())
+    if not asset_record_keys_needed:
+        return []
+
```
gibsondan, line 230: could this be the logic in get_asset_records instead / as well? Seems like a bit of a landmine to have passing in the empty list be treated identical as passing in None

bengotow: Ahh yeah, I was afraid there might be implications of changing the behavior there but I'll review the callsites and see. Would definitely be nice to do an audit of places we say `if not array:` in these resolvers

bengotow:
Ok! I also changed the behavior a bit lower down and updated a few tests per @gibsondan's comment. `event_storage.get_asset_records` discriminates between None and `[]` - the former returns records for all assets and the latter returns nothing, which I think is expected. 

This changes the behavior of in_memory `get_asset_records`, and sql event log `get_asset_records` and `get_latest_materialization_events`. I looked through all the callsites and it seems like this is a safe change, we weren't explicitly passing `asset_keys=[]` anywhere, and places we were passing an array should be better off getting nothing if the array is empty.

clairelin135 (approved):
This is looking great to me. Thanks for making these changes!

bengotow:
It looks like the Buildkite failure is real and these changes require a corresponding update to `CloudEventLogStorage` (because I wrote tests for the behavior I changed in SqlEventLogStorage and InMemoryLogStorage.) Will get a PR up for that tonight.

clairelin135:
@bengotow I think we just need to make the same switch from `not asset_keys` to `asset_keys is None`. I wrote a test PR to confirm this: https://github.com/dagster-io/internal/pull/2621

Feel free to take over the PR!

bengotow:
Thanks @clairelin135! 🙌

PR #8837, opened 2022-07-11
Fix issue where partitioned assets with resources fail materialization in dagit

### LINKED ISSUE #8810: after clicking Materialize button, config modal shows up even when there's no required config
And the partition selector does not show up

To reproduce:
* go here: https://demo.elementl.show/workspace/hacker_news_assets_prod@hacker-news-assets/asset-groups/core
* click "Materialize All"

Perhaps relevant: these assets are partitioned

smackesey:
### Summary & Motivation

This PR adjusts the behavior of the asset materialization button (LaunchAssetExecutionButton). Fixes #8810 

Old behavior:

If an asset had any associated resources, the launchpad modal would show. This included partitioned assets. Showing the launchpad modal would prevent the partition dialog from showing.

New behavior:

If any asset has a partition definition:

- It is an error if any asset has config OR a required resource with required config
- Otherwise, the partition window will show instead of the launchpad. This means you can't add optional config to resources at runtime

If no asset has a partition definition:

- The launchpad modal will show if any asset or resource has a config schema (regardless of whether it is required)

---

This fixes two problems: (1) launchpad modal is overly aggressive for assets with resources that don't require configuration; (2) partitioned assets with (non-configurable) resources were being blocked from materialization in dagit.

The implementation required running a second query in LaunchAssetExecutionButton to retrieve resource info off the pipeline. This is not ideal and we'll probably want to change it later.

- Tried "Materialize All" on the `core` asset group in `examples/hacker_news_assets` => partition dialog
- Tried "Materialize" on the `resource_asset` for `dagster_test/graph_job_op_toys/resources.py`. It has a config schema so it shows the asset modal.

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/AssetConfig.tsx
+++ b/js_modules/dagit/packages/core/src/assets/AssetConfig.tsx
@@ -2,7 +2,7 @@ import {gql} from '@apollo/client';
 
 import {
   SidebarAssetQuery_assetNodeOrError_AssetNode_configField as ConfigField,
-  SidebarAssetQuery_assetNodeOrError_AssetNode_configField_configType_CompositeConfigType as AssetConfigSchema,
+  SidebarAssetQuery_assetNodeOrError_AssetNode_configField_configType_CompositeConfigType as ConfigSchema,
 } from '../asset-graph/types/SidebarAssetQuery';
 import {CONFIG_TYPE_SCHEMA_FRAGMENT} from '../typeexplorer/ConfigTypeSchema';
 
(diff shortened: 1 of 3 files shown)


smackesey:
* master
  * **PR #8837** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8837" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8837?utm_source=stack-comment).

hellendag (approved):

PR #8838, opened 2022-07-12
distinguish between [] and None for asset queries

prha:


gibsondan (approved):
sweet, does this need an internal version too

prha:
The two implementations have sufficiently diverged that this doesn't really have the same issue in OSS.  I think https://github.com/dagster-io/internal/pull/2578 sufficiently helps with this query in internal.

PR #8840, opened 2022-07-12
Remove unused backfill stats graphql field

prha:
### Summary & Motivation
The backfill / partition status refactor made this run-based change obsolete.

BK


gibsondan (approved):
if we wanted to be maximally safe, we would wait to land this until the next release goes out - to avoid cross-push issues with old JS hitting new python

prha:
Yeah, I'm comfortable sitting on this until after this week's release.

PR #8842, opened 2022-07-12
Extract ECS task overrides

jmsanders:
This makes it easier to extend in the future and also prevents an
additional storage call to look up the run's tags when we already have
the run.


python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -286,6 +282,18 @@ def launch_run(self, context: LaunchRunContext) -> None:
             cls=self.__class__,
         )
 
+    def get_overrides(self, run: PipelineRun) -> Dict[str, str]:
```
gibsondan, line 285: _get_overrides? this is private/protected, no?

gibsondan (approved):

PR #8843, opened 2022-07-12
[1.0] move solid to dagster.legacy

dpeng817:
Big Code Mod

Before landing, will re-add solid to the top level API so as to not cause breakages. Just making sure that tests pass without solid being imported from top level, so that at 1.0 we can just flip the switch.


sryza (approved):
Nice

PR #8847, opened 2022-07-12
[asset-defs] allow multi assets to have group names

OwenKephart:
### Summary & Motivation

Addresses: https://github.com/dagster-io/dagster/issues/8787

For this implementation, it seemed more natural to include this field as a part of the AssetOut class, rather than maintain this as a separate mapping.

Unit tests


OwenKephart:
@sryza I can add a group_name parameter to the multi_asset which is mutually exclusive with this behavior

sryza (approved):

Issue #8819, opened 2022-07-11
Add metadata to AssetsDefinition.from_graph


zareami10:
### Dagster version

15.3

### What's the issue?

Unless I'm missing something, there currently doesn't seem to be an easy way to set meatada for graph-backed assets (with different input/outputs).

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

OwenKephart:
hi @zzztimbo ! Currently, the metadata shown for each asset is sourced from the underlying OutputDefinition associated with that asset. This doesn't quite qualify as an "easy way" to set metadata, but if you add metadata to the relevant outputs of the op that produces the asset, then this will be reflected in Dagit, i.e.

```python
@asset(
    out=Out(metadata={"something": 1, "foo": "bar"})
)
def my_asset_producing_op(foo):
    # ...

@graph
def my_gba():
    return my_asset_producing_op(some_op())
```

In an ideal world, are you imagining setting the metadata when doing the `AssetsDefinition.from_graph()` call? Something like:

```python
my_assets = AssetsDefinition.from_graph(
    my_gba,
    metadata_by_key={
        AssetKey("foo"): {"metadata": 1, "other_metadata": "bar"}
    }
)
```

zareami10:
> hi @zzztimbo ! Currently, the metadata shown for each asset is sourced from the underlying OutputDefinition associated with that asset. This doesn't quite qualify as an "easy way" to set metadata, but if you add metadata to the relevant outputs of the op that produces the asset, then this will be reflected in Dagit, i.e.
> 
> ```python
> @asset(
>     out=Out(metadata={"something": 1, "foo": "bar"})
> )
> def my_asset_producing_op(foo):
>     # ...
> 
> @graph
> def my_gba():
>     return my_asset_producing_op(some_op())
> ```
> 
> In an ideal world, are you imagining setting the metadata when doing the `AssetsDefinition.from_graph()` call? Something like:
> 
> ```python
> my_assets = AssetsDefinition.from_graph(
>     my_gba,
>     metadata_by_key={
>         AssetKey("foo"): {"metadata": 1, "other_metadata": "bar"}
>     }
> )
> ```

Not sure if the mention was mistaken, but anyways.

[...]

PR #8861, opened 2022-07-13
add metadata_by_output_name

OwenKephart:
### Summary & Motivation

resolves: https://github.com/dagster-io/dagster/issues/8819

adds together explicit metadata added in the from_graph/from_op invocation with metadata sourced from the output definition

### DIFF
--- a/python_modules/dagster/dagster/core/definitions/asset_layer.py
+++ b/python_modules/dagster/dagster/core/definitions/asset_layer.py
@@ -618,7 +618,7 @@ class AssetLayer:
                 else None
             )
         elif asset_key in self._assets_defs_by_key:
-            return self._assets_defs_by_key[asset_key].metadata_by_asset_key[asset_key]
+            return self._assets_defs_by_key[asset_key].metadata_by_key[asset_key]
         else:
             check.failed(f"Couldn't find key {asset_key}")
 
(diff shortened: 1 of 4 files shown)


erinov1:
@OwenKephart I'm having trouble accessing metadata set using `metadata_by_output_name` in an io_manager, e.g., the following does not work:

```
from dagster import (
    AssetsDefinition,
    IOManager,
    io_manager,
    op,
    repository,
    with_resources,
)


class MyIOManager(IOManager):
        assert context.metadata.get("metadata") == "my_value"

        pass


@io_manager
def my_io_manager():
    return MyIOManager()


@op
def my_op():
    pass


my_asset = AssetsDefinition.from_op(
    op_def=my_op, metadata_by_output_name={"result": {"metadata": "my_value"}}
)

my_asset_with_io_manager = with_resources(
    [my_asset], resource_defs={"io_manager": my_io_manager}
)


@repository
def my_repo():
    return [my_asset_with_io_manager]
```
On the other hand, if I explicitly set the metadata in the op then this works fine, e.g.,

```
@op(out=Out(metadata={"metadata": "my_value"}))
def my_op():
    ...
```
Is this expected? It would nice to be able to access parametrized metadata in the io_manager in this way as well.

PR #8865, opened 2022-07-13
[rfc] execute_job

dpeng817:
Apologies in advance for the essay.

Adds an execute_job function. While based on the old `execute_pipeline` API, this attempts to reinvent some aspects of the return object to fit better with the new APIs. While we _could_ add an execute_job method using the old return method, we would incur a bunch of tech debt that would be nice to avoid (and delete once we remove solid/pipeline from the top level API).

Some problems with the original `execute_pipeline` API:
- steep difficulty increase once using an executor that isn't in_process. Made worse by the new APIs, which default to multi-process.
- requires construction of a reconstructable, which might require users to reconfigure how they write their jobs (or might not work at all with certain job creation pathways, such as factories).
- retrieves results by opening a context manager within a function, which can result in weird interactions for certain types.
The `execute_in_process` API solves most of the test use cases that `execute_pipeline` used to fill, without many of the problems. But there's still a distinct use case of wanting to execute a job from python using a non-in-process executor. This is a heavier weight task, and I've attempted to design the new `execute_job` around that a bit better:

- The new return object shares a unified interface across execute_in_process.
- The result object is a context manager. Since `execute_job` operates in multiple processes, we need to capture outputs via the io managers. Certain types of objects can only be accessed when resource context is open though (iirc some objects in spark work this way). Retrieving objects under the guise of a context manager helps us ensure that all objects will be accessible.
- The function only accepts a reconstructable, while execute_pipeline accepted both reconstructable and a PipelineDefinition. I've made this change because it forces users who want to use this function to incur the complexity of multi-process execution immediately, which is the intended use case (if they want to test in-process, they can use execute_in_process).
- The function requires an instance. Similarly to how we only accept a reconstructable, forces a user to incur the complexity of using the instance instead of setting up an ephemeral one (which will not work on most executors).

One issue that persists is that the context we spin up for output retrieval re-ups the resources, instead of "keeping them warm", so to speak. This means that any outputs whose values are only stored on the resource will be lost (affects mem_io_manager for example, although we can mitigate that case in the short term using output_capture). If we take the approach of `execute_job` being a context manager however, then we can fix this internally without any breaking changes / API changes for users.

Another issue that persists is that we are still using reconstructable. Given the time-frame to 1.0, it is unlikely that we will come to consensus on a replacement. Out of the scope of this PR, but I think if we include some reconstructable-creation utility methods on the repository, we can at least handle the factory case. A la `repo_def.get_reconstructable_job(job_name)`.

### DIFF
--- a/examples/docs_snippets/docs_snippets/guides/dagster/reexecution/unreliable_job.py
+++ b/examples/docs_snippets/docs_snippets/guides/dagster/reexecution/unreliable_job.py
@@ -1,5 +1,12 @@
-from random import random
+RESULTS = [1, 1, 0.4]
 
+
+def some_random_result():
+    return RESULTS.pop()
+
+
+# isort: split
+# start_random_job
 from dagster import in_process_executor, job, op
 
 
@@ -11,7 +18,7 @@ def start():
 @op
 def unreliable(num: int) -> int:
     failure_rate = 0.5
-    if random() < failure_rate:
+    if some_random_result() < failure_rate:
         raise Exception("blah")
 
     return num
@@ -25,3 +32,6 @@ def end(_num: int):
 @job(executor_def=in_process_executor)
 def unreliable_job():
     end(unreliable(start()))
+
+
+# end_random_job
--- a/python_modules/dagster/dagster/_core/execution/api.py
+++ b/python_modules/dagster/dagster/_core/execution/api.py
@@ -169,8 +169,7 @@ def execute_run(
     pipeline_run: PipelineRun,
     instance: DagsterInstance,
     raise_on_error: bool = False,
-    _from_execute_job: bool = False,
-) -> Union[PipelineExecutionResult, ExecuteJobResult]:
+) -> PipelineExecutionResult:
     """Executes an existing pipeline run synchronously.
 
     Synchronous version of execute_run_iterator.
@@ -260,19 +259,7 @@ def execute_run(
     event_list = list(_execute_run_iterable)
 
     pipeline_def = pipeline.get_definition()
-    if _from_execute_job:
-        return ExecuteJobResult(
-            job_def=pipeline.get_definition(),
-            reconstruct_context=scoped_pipeline_context(
-                execution_plan,
-                pipeline,
-                pipeline_run.run_config,
-                pipeline_run,
-                instance,
-            ),
-            event_list=event_list,
-            dagster_run=instance.get_run_by_id(pipeline_run.run_id),
-        )
+
     return PipelineExecutionResult(
         pipeline.get_definition(),
         pipeline_run.run_id,
@@ -380,34 +367,27 @@ def execute_job(
     raise_on_error: bool = True,
     op_selection: Optional[List[str]] = None,
 ) -> ExecuteJobResult:
-    """Execute a pipeline synchronously.
+    """Execute a job synchronously.
 
-    Users will typically call this API when testing pipeline execution, or running standalone
+    Users will typically call this API when testing job execution, or running standalone
     scripts.
 
     Parameters:
-        pipeline (Union[IPipeline, PipelineDefinition]): The pipeline to execute.
-        run_config (Optional[dict]): The configuration that parametrizes this run,
-            as a dict.
-        mode (Optional[str]): The name of the pipeline mode to use. You may not set both ``mode``
-            and ``preset``.
-        preset (Optional[str]): The name of the pipeline preset to use. You may not set both
-            ``mode`` and ``preset``.
-        tags (Optional[Dict[str, Any]]): Arbitrary key-value pairs that will be added to pipeline
-            logs.
-        instance (Optional[DagsterInstance]): The instance to execute against. If this is ``None``,
-            an ephemeral instance will be used, and no artifacts will be persisted from the run.
+        job (ReconstructableJob): The job to execute.
+        instance (DagsterInstance): The instance to execute against.
+        run_config (Optional[dict]): The configuration that parametrizes this run, as a dict.
+        tags (Optional[Dict[str, Any]]): Arbitrary key-value pairs that will be added to run logs.
         raise_on_error (Optional[bool]): Whether or not to raise exceptions when they occur.
             Defaults to ``True``, since this is the most useful behavior in test.
-        solid_selection (Optional[List[str]]): A list of solid selection queries (including single
-            solid names) to execute. For example:
+        op_selection (Optional[List[str]]): A list of op selection queries (including single
+            op names) to execute. For example:
 
-            - ``['some_solid']``: selects ``some_solid`` itself.
-            - ``['*some_solid']``: select ``some_solid`` and all its ancestors (upstream dependencies).
-            - ``['*some_solid+++']``: select ``some_solid``, all its ancestors, and its descendants
+            - ``['some_op']``: selects ``some_op`` itself.
+            - ``['*some_op']``: select ``some_op`` and all its ancestors (upstream dependencies).
+            - ``['*some_op+++']``: select ``some_op``, all its ancestors, and its descendants
               (downstream dependencies) within 3 levels down.
-            - ``['*some_solid', 'other_solid_a', 'other_solid_b+']``: select ``some_solid`` and all its
-              ancestors, ``other_solid_a`` itself, and ``other_solid_b`` and its direct child solids.
+            - ``['*some_op', 'other_op_a', 'other_op_b+']``: select ``some_op`` and all its
+              ancestors, ``other_op_a`` itself, and ``other_op_b`` and its direct child ops.
 
     Returns:
       :py:class:`JobExecutionResult`: The result of job execution.
@@ -423,9 +403,15 @@ def execute_job(
         tags=tags,
         solid_selection=op_selection,
         raise_on_error=raise_on_error,
-        _from_execute_job=True,
     )
-    return cast(ExecuteJobResult, result)
+
+    # We use PipelineExecutionResult to construct the JobExecutionResult.
+    return ExecuteJobResult(
+        job_def=job.get_definition(),
+        reconstruct_context=result.reconstruct_context(),
+        event_list=result.event_list,
+        dagster_run=instance.get_run_by_id(result.run_id),
+    )
 
 
 def execute_pipeline(
@@ -496,8 +482,7 @@ def _logged_execute_pipeline(
     tags: Optional[Dict[str, Any]] = None,
     solid_selection: Optional[List[str]] = None,
     raise_on_error: bool = True,
-    _from_execute_job: bool = False,
-) -> Union[PipelineExecutionResult, ExecuteJobResult]:
+) -> PipelineExecutionResult:
     check.inst_param(instance, "instance", DagsterInstance)
     (
         pipeline,
@@ -534,7 +519,6 @@ def _logged_execute_pipeline(
         pipeline_run,
         instance,
         raise_on_error=raise_on_error,
-        _from_execute_job=_from_execute_job,
     )
 
 
--- a/python_modules/dagster/dagster/_core/execution/api.py
+++ b/python_modules/dagster/dagster/_core/execution/api.py
@@ -1,12 +1,12 @@
 import sys
 from contextlib import contextmanager
-from typing import Any, Dict, FrozenSet, Iterator, List, Mapping, Optional, Tuple, Union
+from typing import Any, Dict, FrozenSet, Iterator, List, Mapping, Optional, Tuple, Union, cast
 
 import dagster._check as check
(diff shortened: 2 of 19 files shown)


sryza:
With this PR, it would be good to include the MIGRATION.md entry for this.  It could go in a 1.0 section marked "in-progress".

> The new return object shares a unified interface across execute_in_process.

Makes sense

> The function only accepts a reconstructable, while execute_pipeline accepted both reconstructable and a PipelineDefinition. I've made this change because it forces users who want to use this function to incur the complexity of multi-process execution immediately, which is the intended use case (if they want to test in-process, they can use execute_in_process).

This makes complete sense. 

> The function requires an instance. Similarly to how we only accept a reconstructable, forces a user to incur the complexity of using the instance instead of setting up an ephemeral one (which will not work on most executors).

Will an ephemeral instance work with the multiprocess executor? If no, then I think requiring an instance is an easy choice. If yes, then might require more deliberation.

> The result object is a context manager. Since execute_job operates in multiple processes, we need to capture outputs via the io managers. Certain types of objects can only be accessed when resource context is open though (iirc some objects in spark work this way). Retrieving objects under the guise of a context manager helps us ensure that all objects will be accessible.

Does this mean we'll boot up resources immediately, when we might have been lazy otherwise?

dpeng817:
To answer your questions:
1. An ephemeral instance will _not_ work with the multiprocess executor; which is the major reason for this change.
2. We will only boot up resources if you explicitly open the context manager; you can still access `result.success` and events without the context manager, and you can open it explicitly when you want to access outputs.

sryza:
All sounds good to me in that case. Anything I didn't think of @alangenfeld ?

python_modules/dagster/dagster/core/execution/execute_job_result.py
```diff
@@ -0,0 +1,118 @@
...
+            event
+            for event in self._filter_events_by_handle(handle)
+            if event.step_kind == StepKind.COMPUTE
+        ]
+
+    def _get_output_for_handle(self, handle: NodeHandle, output_name: str) -> Any:
+        if not self._context:
+            raise DagsterInvariantViolationError(
+                "In order to access output objects, the result of `execute_job` must be opened as a context manager: 'with execute_job(...) as result:"
+            )
```
alangenfeld, line 65: This pattern is a little tricksy, where we have an optionally context manger object and some methods throw if you didn't enter. The alternative is introducing an different object that you have to context manager enter to get that has the result methods. Trade off from my view is api surface area vs static analyzability.

I'm fine with either approach, just want to make sure we make this trade off decision explicitly.

alangenfeld: The main benefit of the tighter surface area appears to be `ExecutionResult` base where output access is common between `execute_job` and `execute_in_process`. That is nice, though does sort of hide the fundamental change in behavior underlying the output fetching between the two approaches. I think `ExecutionResult` not having output fetching as common methods and having the divergent behavior modeled as different apis has its merits.

python_modules/dagster/dagster/core/execution/api.py
```diff
@@ -356,6 +372,62 @@ def ephemeral_instance_if_missing(
...
+
+            - ``['some_solid']``: selects ``some_solid`` itself.
+            - ``['*some_solid']``: select ``some_solid`` and all its ancestors (upstream dependencies).
+            - ``['*some_solid+++']``: select ``some_solid``, all its ancestors, and its descendants
+              (downstream dependencies) within 3 levels down.
+            - ``['*some_solid', 'other_solid_a', 'other_solid_b+']``: select ``some_solid`` and all its
+              ancestors, ``other_solid_a`` itself, and ``other_solid_b`` and its direct child solids.
+
+    Returns:
+      :py:class:`JobExecutionResult`: The result of job execution.
```
alangenfeld, line 413: `pipeline` `solid` still present here - do another pass

python_modules/dagster/dagster/core/execution/api.py
```diff
@@ -257,6 +259,20 @@ def execute_run(
...
+            job_def=pipeline.get_definition(),
+            reconstruct_context=scoped_pipeline_context(
+                execution_plan,
+                pipeline,
+                pipeline_run.run_config,
+                pipeline_run,
+                instance,
+            ),
+            event_list=event_list,
+            dagster_run=instance.get_run_by_id(pipeline_run.run_id),
```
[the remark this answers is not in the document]
dpeng817: The pipeline run in scope isn't updated with correct run status, since it's created before the execution actually occurs

python_modules/dagster/dagster/core/execution/execute_job_result.py
dpeng817: I feel like if communicated effectively in docs, then it doesn't make a huge difference. I've generally found that when there are two separate feature stacks that are parallel, there will be feature slippage, and I think having a shared interface will hopefully help to prevent that.

python_modules/dagster/dagster/core/execution/api.py
```diff
@@ -356,6 +359,61 @@ def ephemeral_instance_if_missing(
...
+    raise_on_error: bool = True,
+    op_selection: Optional[List[str]] = None,
+) -> ExecuteJobResult:
+    """Execute a job synchronously.
+
+    Users will typically call this API when testing job execution, or running standalone
+    scripts.
+
+    Parameters:
+        job (ReconstructableJob): The job to execute.
```
alangenfeld, line 376: probably want to communicate `reconstructable` here

python_modules/dagster/dagster/core/execution/api.py
```diff
@@ -356,6 +359,61 @@ def ephemeral_instance_if_missing(
...
+        instance (DagsterInstance): The instance to execute against.
+        run_config (Optional[dict]): The configuration that parametrizes this run, as a dict.
+        tags (Optional[Dict[str, Any]]): Arbitrary key-value pairs that will be added to run logs.
+        raise_on_error (Optional[bool]): Whether or not to raise exceptions when they occur.
+            Defaults to ``True``, since this is the most useful behavior in test.
```
alangenfeld, line 381: I'm not sure we want this default here? `execute_in_process` is test focused now so this is more likely script / "real" usage where we may want to default the other way

python_modules/dagster/dagster/core/execution/api.py
```diff
@@ -564,12 +622,13 @@ def reexecute_pipeline(
...
         )
 
-        return execute_run(
+        result = execute_run(
             pipeline,
             pipeline_run,
             execute_instance,
             raise_on_error=raise_on_error,
         )
+        return cast(PipelineExecutionResult, result)
```
alangenfeld, line 631: you can clean this up right?

python_modules/dagster/dagster/core/execution/execute_job_result.py
alangenfeld: > communicated effectively in docs

[1]

are there going to be additional docs? should those just be in this PR too?

python_modules/dagster/dagster/core/execution/execution_result.py
```diff
@@ -0,0 +1,192 @@
...
+
+        If the top-level job has no output, calling this method will result in a
+        DagsterInvariantViolationError.
+
+        Args:
+            output_name (Optional[str]): The name of the output to retrieve. Defaults to `result`,
+                the default output name in dagster.
+
+        Returns:
+            Any: The value of the retrieved output.
```
alangenfeld, line 88: [1] the shared impl API docs don't telegraph having to open the context manager. Lil gross for the base class docs to have to know about / refer to its sub classes. Could override the method and just call the parent to provide different doc strings.

python_modules/dagster/dagster/core/execution/execution_result.py
```diff
@@ -0,0 +1,192 @@
...
+        origin_output_def, origin_handle = mapped_node.definition.resolve_output_to_origin(
+            output_mapping.maps_from.output_name,
+            NodeHandle(mapped_node.name, None),
+        )
+
+        # Get output from origin node
+        return self._get_output_for_handle(check.not_none(origin_handle), origin_output_def.name)
+
+    def output_for_node(self, node_str: str, output_name: str = DEFAULT_OUTPUT) -> Any:
+        """Retrieves output value with a particular name from the in-process run of the job.
```
alangenfeld, line 114: in-process not necessarily the case anymore

python_modules/dagster/dagster/core/execution/execution_result.py
```diff
@@ -0,0 +1,192 @@
...
+        raise NotImplementedError()
+
+    @property
+    def success(self) -> bool:
+        """bool: Whether execution was successful."""
+        return self.dagster_run.is_success
+
+    @property
+    def all_node_events(self) -> Sequence[DagsterEvent]:
+        """List[DagsterEvent]: All dagster events from the in-process execution."""
```
alangenfeld, line 54: this is base class so in-process not always true

python_modules/dagster/dagster/core/execution/api.py
```diff
@@ -257,6 +258,8 @@ def execute_run(
     )
     event_list = list(_execute_run_iterable)
 
+    pipeline_def = pipeline.get_definition()
```
alangenfeld, line 261: cleanup

python_modules/dagster/dagster/core/execution/execute_job_result.py
dpeng817: will add

python_modules/dagster/dagster/core/execution/api.py
dpeng817: that makes sense to me

examples/docs_snippets/docs_snippets/guides/dagster/reexecution/unreliable_job.py
```diff
@@ -1,5 +1,12 @@
-from random import random
+RESULTS = [1, 1, 0.4]
```
dpeng817, line 1: deterministic results so that we can actually put this under test.

python_modules/dagster/dagster/_core/execution/api.py
```diff
@@ -356,6 +376,177 @@ def ephemeral_instance_if_missing(
             yield ephemeral_instance
 
 
+class ReexecutionOptions(NamedTuple):
+    """Reexecution configuration for python-based execution in Dagster.
+
+    Args:
+        parent_run_id (str): The run_id of the run to reexecute.
+        step_selection (Sequence[str]): The set of steps to reexecute. Must be a subset or match of the set of steps executed in the original run.
+    """
```
dpeng817, line 385: reexecution machinery

python_modules/dagster/dagster/_core/execution/api.py
```diff
@@ -356,6 +376,195 @@ def ephemeral_instance_if_missing(
...
+    ``execute_job`` can also be used to reexecute a run, by providing a :py:class:`ReexecutionOptions`.
+
+    .. code-block:: python
+
+        from dagster import ReexecutionOptions, execute_job
+
+        instance = DagsterInstance.get()
+
+        options = ReexecutionOptions.from_failure(run_id=failed_run_id, instance)
+        execute_job(reconstructable(job), instance, reexecution_options=options)
```
dpeng817, line 498: code blocks for various pieces of functionality.

yuhan (commented):
I only looked the re-execution part and it looks pretty reasonable.

python_modules/dagster/dagster/_core/execution/api.py
```diff
@@ -356,6 +376,195 @@ def ephemeral_instance_if_missing(
             yield ephemeral_instance
 
 
+class ReexecutionOptions(NamedTuple):
+    """Reexecution configuration for python-based execution in Dagster.
```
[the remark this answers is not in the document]
dpeng817: Places where the execution is triggered by you calling a python function, as opposed to dagit or the CLI. FWIW this is consistent with how we message in the docs iirc.

python_modules/dagster/dagster_tests/execution_tests/test_execute_job.py
```diff
@@ -0,0 +1,257 @@
...
+def test_reexecution_with_steps(instance):
+    with execute_job(
+        reconstructable(emit_based_on_config),
+        instance,
+        run_config={"ops": {"the_op": {"config": "blah"}}},
+    ) as result:
+        assert result.success
+        assert result.output_for_node("ingest") == "blah"
+
+    reexecution_options = ReexecutionOptions(parent_run_id=result.run_id, step_selection=["ingest"])
```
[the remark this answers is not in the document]
dpeng817: good idea

alangenfeld (approved):

PR #8866, opened 2022-07-13
PartitionMappings in public API and when non-partitioned assets depend on partitioned assets

sryza:
Fixes https://github.com/dagster-io/dagster/issues/8867


OwenKephart (approved):
nice

PR #8869, opened 2022-07-13
[graphql] status code based on errors

alangenfeld:
### Summary & Motivation

* instead of assuming errors returned by the `ExecutionResult` are bad queries (`400`), attempt to discern resolver errors which we now classify as `500` from syntax/query errors which should continue to `400`
* when `ErrorCapture` occurs and produces a `PythonError`, classify those requests as `500` since ideally that condition indicates an unexpected server error has occurred 
* update the Apollo `ErrorLink` to resolve requests even on `500` if graphql data is present 

* also includes a small fix to the sensor stop mutation which now producing `500` s was causing issues in tests / dagit. 

BK here and internal
trigger ErrorCaptrue to generate `PythonError` payloads and ensure the front end continues to display the modal showing the error.

### DIFF
--- a/python_modules/dagit/dagit/graphql.py
+++ b/python_modules/dagit/dagit/graphql.py
@@ -4,8 +4,9 @@ from enum import Enum
 from typing import Any, AsyncGenerator, Dict, List, Optional, Tuple, Union, cast
 
 from dagit.templates.playground import TEMPLATE
+from dagster_graphql.implementation.utils import ErrorCapture
 from graphene import Schema
-from graphql.error import GraphQLError
+from graphql.error import GraphQLError, GraphQLLocatedError
 from graphql.error import format_error as format_graphql_error
 from graphql.execution import ExecutionResult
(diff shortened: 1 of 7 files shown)


hellendag (commented):
If everything seems to be working well, then the JS part seems like a fairly clean solution to me. Will allow others to take a look at the Python.

python_modules/dagit/dagit/graphql.py
```diff
@@ -256,6 +263,33 @@ def create_asgi_app(
...
+        server_error = False
+        user_error = False
+
+        if resolver_errors:
+            for error in resolver_errors:
+                if isinstance(error, GraphQLLocatedError):
+                    server_error = True
+                else:
+                    # syntax error, invalid query, etc
+                    user_error = True
```
gibsondan, line 280: maybe input_error or request_error?

gibsondan (approved):

PR #8883, opened 2022-07-14
remove the non-asset version of the hacker news example

sryza:
### Summary & Motivation

 - Because the aim of the hacker news jobs is to produce and maintain data assets, we wouldn't recommend using the non-asset APIs.
 - Showing multiple ways of doing the same thing can be confusing.
 - The example was getting out of date and is costly to maintain.


OwenKephart (approved):
beautiful

docs/content/concepts/ops-jobs-graphs/ops.mdx
```diff
@@ -225,11 +225,3 @@ def my_op_factory(
...
     return my_inner_op
 ```
-
-## See it in action
-
-For more examples of ops, check out the following in our [Hacker News example](https://github.com/dagster-io/dagster/tree/master/examples/hacker_news):
-
-- [Specifying Ins and Outs](https://github.com/dagster-io/dagster/blob/master/examples/hacker_news/hacker_news/ops/comment_stories.py)
-- [Using resources](https://github.com/dagster-io/dagster/blob/master/examples/hacker_news/hacker_news/ops/download_items.py)
-- [Per-Output IO Manager](https://github.com/dagster-io/dagster/blob/master/examples/hacker_news/hacker_news/ops/user_top_recommended_stories.py)
```
yuhan, line 235: [1]

yuhan (commented):
Makes total sense to do so.

My only concern is that we might have been pointing users to this example for certain features like [1] (perfect if we believe it's a rare case), so removing this example right away may cause some ttv issue for users learning. thoughts on waiting for the TTV comprehensive example and then drop this one?

sryza:
@yuhan a big part of my motivation for this change was a conversation I had with a prospective user who came across this example.  My impression was that it left them more confused than if they hadn't found it at all.  So my thinking here was that pointing users to a stale example that doesn't match current recommendations can be less helpful than having no example.

For the op features in [1], I just pushed a change that backfills some of them using the NYT example.

Thoughts?

yuhan (approved):
i'm sold - lets get rid of this one

jamiedemaria (approved):

PR #8885, opened 2022-07-14
[bug] fix input loading regression

OwenKephart:
### Summary & Motivation

If the following were both true:

- An `In()` of an op had an `asset_key` argument set
- That `In()` was unconnected, and intended to be loaded using a DagsterTypeLoader, e.g.:

```python
run_config={"ops": {"foo": {"inputs": {"bar": 1}}}}
```

Then you would experience an error. This is because of this line: https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/core/definitions/run_config.py#L211.

Because this input was technically mapped to an asset key, this line would evaluate to True, causing the input config schema to be None, rather than the config schema specified by the type loader.

The feature that line corresponds to is purely designed for software defined assets, which cannot be used with dagster type loaders. This change breaks any mapping of InputDefinition to AssetKey in the AssetLayer for non-SDA use cases. All instances of this mapping being used (as far as I could tell) are similarly designed around SDAs, and so this seems to be a safe change.

The equivalent asset_info_for_output_handle does not have any similar issues (as far as I can tell), so I left it in, alongside all of the other asset information we scrape out.

New test failed before, succeeds now.

### DIFF
--- a/python_modules/dagster/dagster/core/definitions/run_config.py
+++ b/python_modules/dagster/dagster/core/definitions/run_config.py
@@ -208,7 +208,11 @@ def get_inputs_field(
         has_upstream = input_has_upstream(dependency_structure, inp_handle, solid, name)
         if inp.input_manager_key:
             input_field = get_input_manager_input_field(solid, inp, resource_defs)
-        elif asset_layer.asset_key_for_input(handle, name) and not has_upstream:
+        elif (
+            asset_layer.has_assets_defs
+            and asset_layer.asset_key_for_input(handle, name)
+            and not has_upstream
+        ):
             input_field = None
         elif name in direct_inputs and not has_upstream:
             input_field = None
--- a/python_modules/dagster/dagster/core/definitions/run_config.py
+++ b/python_modules/dagster/dagster/core/definitions/run_config.py
@@ -209,6 +209,7 @@ def get_inputs_field(
         if inp.input_manager_key:
             input_field = get_input_manager_input_field(solid, inp, resource_defs)
         elif (
+            # if you have asset definitions, input will be loaded from the source asset
             asset_layer.has_assets_defs
             and asset_layer.asset_key_for_input(handle, name)
             and not has_upstream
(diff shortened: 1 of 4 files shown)


OwenKephart:
> looks like some real test are failing? will wait to review incase those tests cause code significant changes

A wise decision -- I changed the approach to be a bit more surgical, and the tests seem happier for it. 

Now, I only ignore asset key <> input mapping information in the exact code paths related to input loading / input config. We determine whether an asset layer was generated from SDAs or InputDef(asset_key=...) by seeing if the AssetLayer has any AssetsDefinitions in it.

python_modules/dagster/dagster/core/definitions/run_config.py
```diff
@@ -208,7 +208,11 @@ def get_inputs_field(
         has_upstream = input_has_upstream(dependency_structure, inp_handle, solid, name)
         if inp.input_manager_key:
             input_field = get_input_manager_input_field(solid, inp, resource_defs)
-        elif asset_layer.asset_key_for_input(handle, name) and not has_upstream:
+        elif (
+            asset_layer.has_assets_defs
+            and asset_layer.asset_key_for_input(handle, name)
+            and not has_upstream
+        ):
```
alangenfeld, line 215: worth a comment

python_modules/dagster/dagster/core/execution/plan/plan.py
```diff
@@ -482,7 +482,9 @@ def get_step_input_source(
     ):
         if input_def.root_manager_key or input_def.input_manager_key:
             return FromRootInputManager(solid_handle=handle, input_name=input_name)
-        elif asset_layer.asset_key_for_input(handle, input_handle.input_name):
+        elif asset_layer.has_assets_defs and asset_layer.asset_key_for_input(
+            handle, input_handle.input_name
+        ):
```
alangenfeld, line 487: comment also ?

alangenfeld (approved):
dope

PR #8888, opened 2022-07-15
Move pipeline to dagster.legacy

dpeng817:
Title.


sryza (approved):
Should we do `_legacy`?

PR #8889, opened 2022-07-15
AssetSensorDefinition arg deprecations

dpeng817:
Deprecate arguments on AssetSensorDefinition


smackesey:
What do you think about putting these warnings in the constructor instead of on property access? Property access is IMO likely to generate more warning spam.

PR #8890, opened 2022-07-15
[1.0] DagsterPipelineRunMetadataValue -> DagsterRunMetadataValue

dpeng817:
Add DagsterRunMetadataValue, throw 1.1.0 deprecation warning for DagsterPipelineRunMetadata.

Also follows through on removal of DagsterPipelineRunMetadataEntry, which is telegraphed for 1.0. Therefore, this cannot land until 1.0.

### DIFF
--- a/python_modules/dagster/dagster/__init__.py
+++ b/python_modules/dagster/dagster/__init__.py
@@ -356,7 +356,7 @@ _DEPRECATED_RENAMED = {
     "FloatMetadataEntryData": (FloatMetadataValue, "1.0.0"),
     "IntMetadataEntryData": (IntMetadataValue, "1.0.0"),
     "DagsterPipelineRunMetadataEntryData": (
-        DagsterPipelineRunMetadataValue,
+        DagsterRunMetadataValue,
         "1.0.0",
     ),
     "DagsterAssetMetadataEntryData": (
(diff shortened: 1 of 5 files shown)


smackesey (commented):
If I understand this correctly, this is basically the same situation I encountered with renaming metadata classes a few months ago-- it's purely a class name change. I think this will be simpler if we use the import deprecation layer I put up for the metadata classes, you'd just:

- change the class name to `DagsterRunMetadataValue`
- use the old name for storage name on the serializer (as you've done)
- add an entry in `dagster/__init__.py` under `_DEPRECATED_RENAMED`

No new class required. What do you think?

dpeng817:
@smackesey seems reasonable to me. Will do

sryza (approved):

python_modules/dagster/dagster/__init__.py
```diff
@@ -33,7 +33,8 @@
     CompositeSolidDefinition,
     ConfigMapping,
     DagsterAssetMetadataValue,
-    DagsterPipelineRunMetadataValue,
+    DagsterRunMetadataValue,
+    DagsterRunMetadataValue,
```
OwenKephart, line 37: dupe?

PR #8891, opened 2022-07-15
Throw warning if using solid arg on DependencyDefinition

dpeng817:
We didn't explicitly do this.


python_modules/dagster/dagster/core/definitions/dependency.py
```diff
@@ -637,6 +638,10 @@ def __new__(
         description: Optional[str] = None,
         node: Optional[str] = None,
     ):
+        if solid:
```
smackesey, line 641: We should use `canonicalize_backcompat_args` in `dagster.utils.backcompat` instead, it serves exactly this purpose and will keep error messages consistent.

smackesey (changes requested):

dpeng817:
We're removing arguments

PR #8893, opened 2022-07-15
[docs] - Clean up graph-backed asset example, put under test

dpeng817:
Going through AssetGroup removal, noticed this


sryza (approved):
You probably need to run `make snapshot` to get this to be reflected in the mdx, right?

PR #8902, opened 2022-07-15
Change base image for OSS release builds

gibsondan:
Summary:
Hypothesis that a newer version has the right pip that matches what we are using for testing elsewhere


alangenfeld (approved):
cc @johannkm can we clean up a bunch of these?

would be great to get single source of truth between test image builds and these builds to prevent repeated drift

PR #8904, opened 2022-07-15
update README for hacker news assets example

sryza:


examples/hacker_news_assets/README.md
```diff
@@ -1,109 +1,46 @@
...
+# Hacker News Demo
 
-This repo used Dagster's experimental software-defined asset APIs to set up a "realistic" example of using dagster in a production world, integrating with
-many of its features, including:
+This is meant to be a full "realistic" demo of Dagster, that takes advantage of many of its features, including:
 
-- Schedules / Sensors
-- Asset Materializations
+- Software-defined assets
+- Schedules
```
jamiedemaria, line 6: is there a schedule in the example? the schedules folder is empty

jamiedemaria (approved):
the readme is much better now!

Another thing i noticed when poking around is that there are a lot of unused folders in the example (jobs, hooks, schedules, maybe a couple others) that i think are just adding noise and could be removed. would probably be quick to add into this pr

examples/hacker_news_assets/README.md
```diff
@@ -1,109 +1,46 @@
...
+- Sensors
 - IOManagers
 - Resources
-- Unit Tests
-- dbt, snowflake, s3, and pyspark integrations
+- dbt, Snowflake, S3, and PySpark integrations
```
jamiedemaria, line 10: the example still uses a snowflake io manager that's written in `resources/snowflake_io_manager.py` not our built in io manager

sryza: good point

examples/hacker_news_assets/README.md
sryza: there actually is - it's in core/__init__.py

sryza:
@jamiedemaria I updated the diff to take Snowflake out of the integrations section and remove those unused folders (great call).

Issue #8786, opened 2022-07-07
hide mode name in resources section of sidebar for jobs


sryza:
In this screenshot, "default" refers to the mode, which is a legacy concept. It will always say "default" for every job, so there's no reason to keep it.
<img width="430" alt="image" src="https://user-images.githubusercontent.com/654855/177812482-0740c250-a9af-48ff-a013-264d3b842f16.png">

PR #8907, opened 2022-07-15
[dagit] Remove “mode” label and description in explorer sidebar

bengotow:
### Summary & Motivation

This was a super straightforward ask in #8786. If you're viewing a Job, the right sidebar which shows resources no longer shows the mode name or description, because it's always default.

Todo: 

This code is pretty old and needs to be cleaned up I think - we have some stray Pipeline / Solid naming to clean up, and "isGraph" actually means "is a graph or job", which isn't clear. I'm going to push a follow up to this that renames some of the components, and maybe another follow up that changes the way the "root" of the sidebar is rendered with more clear cases for `snapshot`, `graph`, `job` and `pipeline`, which are sort of mangled together currently.


js_modules/dagit/packages/core/src/pipelines/GraphExplorer.tsx
```diff
@@ -291,8 +288,8 @@ export const GraphExplorer: React.FC<GraphExplorerProps> = (props) => {
           <Route
             // eslint-disable-next-line react/no-children-prop
             children={({location}: {location: any}) => (
-              <SidebarTabbedContainer
```
bengotow, line 294: I renamed this because 1) it's the thing rendered by a Route so it's sort of a root, and I think we should avoid calling in the sidebar "Container" because "SolidContainer" is actually an interface exposed by the GraphQL api

hellendag (approved):

Issue #8849, opened 2022-07-12
error when @repository-decorated function has arguments


sryza:
This should error:
```
@repository
def repo(default_executor_def=in_process_executor):
    ...
```

PR #8913, opened 2022-07-15
Error when @repository-decorated function has arguments

roeij:
### Summary & Motivation
Resolving https://github.com/dagster-io/dagster/issues/8849

- Debugged and tested locally
- Unit test for `CheckError` trigger on invalid decorated func


sryza (changes requested):
Thanks for the contribution @roeij !

I ran tests on it and discovered a couple small issues:
* The linter complained about `[2022-07-16T00:08:38Z] dagster_tests/general_tests/test_repository.py:127:31: W0613: Unused argument 'invalid_arg'`. You should be able to fix that by adding a leading underscore to `invalid_arg`.
* isort complained about import ordering. You should be able to fix that by running `make isort` or by moving the import you added down a line.

sryza (approved):
It looks like the update fixed the lint and isort issues. Now merging!

PR #8914, opened 2022-07-15
[dagit] Clean up code around the graph sidebar

bengotow:
### Summary & Motivation

This PR simplifies the right sidebar of Dagit's  DAG explorer and modernizes the naming conventions. We render the same "SidebarContainerOverview" regardless of what the "container" is (pipeline / graph / job or snapshot) and the component inspects the object given to it to make subtle changes to the panel:

- Pipelines get "Modes" instead of "Resources"
- Snapshots have a "snapshot" label if they are not current
- Graphs have no resources

I also removed a whacky context I'd added that was feeding this sidebar a React component. I don't quite remember why we did that but I think things have changed such that it's not necessary anymore.

Here's a quick video of what these different states look like: 
https://www.loom.com/share/5020a2929114448ab06e3c70c2fe1906


hellendag (approved):
Seems reasonable!

PR #8927, opened 2022-07-18
[docs] - graph backed assets doc fix

jamiedemaria:
### Summary & Motivation
first code snippet for graph backed assets doesn't return a value from the graph, which will throw an error if you copy paste it


erinkcochran87 (approved):
Good catch!

PR #8928, opened 2022-07-18
add a job_name property to PipelineRun

sryza:
### Summary & Motivation

This will allow transitioning away from legacy concepts in usages like these: https://docs.dagster.io/concepts/partitions-schedules-sensors/sensors#run-status-sensors

I did not

### DIFF
--- a/python_modules/dagster/dagster/core/storage/pipeline_run.py
+++ b/python_modules/dagster/dagster/core/storage/pipeline_run.py
@@ -484,6 +484,10 @@ class PipelineRun(
         # Compat
         return self.parent_run_id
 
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster/core/storage/pipeline_run.py
```diff
@@ -484,6 +484,10 @@ def previous_run_id(self):
         # Compat
         return self.parent_run_id
 
+    @property
+    def job_name(self):
```
alangenfeld, line 488: return type?

alangenfeld (approved):

PR #8933, opened 2022-07-18
PartitionsDefinition.get_partition_keys_in_range

sryza:
### Summary & Motivation

Requested by a user


python_modules/dagster/dagster/core/definitions/partition.py
```diff
@@ -238,6 +247,32 @@ def get_partitions(
     ) -> Sequence[Partition[str]]:
         return self._partitions
 
+    def get_partition_keys_in_range(self, start: str, end: str) -> Sequence[str]:
```
OwenKephart, line 250: not blocking, and very much a silly quibble, but I'm not sure if I would expect the partitions defined by a static partitions definition to have a defined order. They implicitly do, because we take in a List[str] not Set[str], I just felt vaguely weird seeing this, as I would have intuitively expected StaticPartitionsDefinition(["car", "truck"]) == StaticPartitionsDefinition(["truck", "car"]).

People that want that to be the case don't have to use this function, so it's probably fine, just thought I'd flag.

python_modules/dagster/dagster/core/definitions/time_window_partitions.py
```diff
@@ -141,17 +148,14 @@ def time_window_for_partition_key(self, partition_key: str) -> TimeWindow:
...
-
-    def start_time_for_partition_key(self, partition_key: str) -> datetime:
-        return pendulum.instance(datetime.strptime(partition_key, self.fmt), tz=self.timezone)
-
-    def get_default_partition_mapping(self):
-        from dagster.core.definitions.time_window_partition_mapping import (
-            TimeWindowPartitionMapping,
-        )
+        prev_time = next(iterator)
+        while prev_time.timestamp() < start_timestamp:
```
OwenKephart, line 152: don't fully understand why this is necessary (would have expected the first value of iterator to always be start_timestamp), but I accept that time iterator functions are complicated and that there's probably a good reason for this.

OwenKephart (approved):

python_modules/dagster/dagster/core/definitions/partition.py
sryza: Hmm yeah - thoughts on whether there are any `StaticPartitionsDefinitions` that make sense to be ordered?

sryza: I'll just put this on TimeWindowPartitionsDefinition for now

sryza:
@OwenKephart I shifted this so that only `TimeWindowPartitionsDefinition` has the method for now, and it now accepts a `PartitionKeyRange` instead of start and end arguments. Mind taking another look?

OwenKephart (approved):
looks good!

PR #8934, opened 2022-07-18
ttv: remove undocumented/legacy example - user in loop

yuhan:
### Summary & Motivation
remove this undocumented and legacy example - i don't think the content is very relevant and it's also using all legacy apis.


sryza (approved):
nice

PR #8942, opened 2022-07-19
[mock] - New landing page

erinkcochran87:
Mock up for a new landing page. This uses components from Yuhan's https://github.com/dagster-io/dagster/pull/8797 PR to mock things up. If we like the examples gallery approach (I do!) and this approach, we can use the same styling (or with minimal adjustments) to make things work.


docs/content/getting-started.mdx
```diff
@@ -1,145 +1,141 @@
...
-To install Dagster into an existing Python environment, run:
-
-```bash
-pip install dagster
-```
-
-This installs the latest stable version of the core Dagster packages in your current Python environment.
+<div className="inline-flex flex-row space-x-4">
+    <button className="py-1 px-4 rounded-lg bg-gable-green text-white hover:bg-transparent hover:text-gable-green hover: border hover:border-gable-green">
+        Set up Dagster
```
sryza: My impulse would be to leave "Set up Dagster" out. I think there's high cost to every choice that we make the reader make, and I think few people who are new to Dagster are going to want to immediately start just installing it - what would they do after that?  If we do want to include guidance here on how to install Dagster, I think it would be better to include what we have on the current "Getting Started" page - i.e. the "pip install", which immediately shows off how simple it is to install Dagster, and avoids the reader needing to make a choice.

docs/content/_navigation.json
```diff
@@ -5,439 +5,195 @@
...
     "children": [
       {
-        "title": "Assets",
+        "title": "Understanding Dagster",
         "children": [
           {
-            "title": "Software-Defined Assets",
-            "path": "/concepts/assets/software-defined-assets"
+            "title": "Designing DAGs",
+            "path": "/building#designing-dags"
```
sryza, line 40: I find it a little confusing to have some left nav items take you to subsections and others take you to standalone pages.

PR #8949, opened 2022-07-19
Snowflake IO manager fix

jamiedemaria:
### Summary & Motivation
the io manager tries to delete the table on every handle output, but when an asset is materialized the first time the table doesnt exist. this causes the io manager to fail unless you make the table manually first. I assume this isn't the behavior we want so i added a check to only delete the table if it exists

also fixes a bug i introduced when i added schema to the config (schema was getting passed as a param twice to some functions)


sryza (commented):
An option that would avoid needing to fire off an additional query could be to catch the error if delete fails and just ignore it if it's because the table doesn't exist? Thoughts?

jamiedemaria:
@sryza that makes a lot more sense. finally got around to implementing it

python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
```diff
@@ -52,10 +54,19 @@ def snowflake_io_manager():
...
+            else {}
+        )
         with SnowflakeConnection(
-            dict(schema=table_slice.schema, **(context.resource_config or {})), context.log  # type: ignore
+            dict(schema=table_slice.schema, **no_schema_config), context.log  # type: ignore
         ).get_connection() as con:
-            con.execute_string(_get_cleanup_statement(table_slice))
+            try:
+                con.execute_string(_get_cleanup_statement(table_slice))
+            except ProgrammingError:
```
sryza, line 67: Do we want to be specific here about the text we expect to distinguish from other things that could cause an error? Might not matter, as it's hard to imagine a situation beyond "table doesn't exist" where we'd run into an error here but not when trying to write the data.

sryza (approved):

python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
jamiedemaria: yeah i didn't have any ideas of errors that wouldn't be caught at the write step. The error message is super simple though, so it'd be easy to match on (thinking "does not exist" would do it)

```
 002003 (42S02): SQL compilation error:
Object 'SANDBOX.JAMIE.STARTER' does not exist or not authorized.
```

PR #8951, opened 2022-07-19
remove partition entries from asset op config schema

sryza:
### Summary & Motivation

I had meant to remove this in the PR that made it so that we no longer rely on these entries, but neglected to.


OwenKephart (approved):
I also thought these were already gone!

OwenKephart:
@sryza looks like snapshots will need to be updated (makes sense). I was trying to be extra cautious when doing a similar thing awhile back, and `git diff` on the updated snapshot files is really awful (makes it look like basically everything is changing). I found success using `git diff --diff-algorithm=histogram --color-moved=dimmed-zebra` for a somewhat more comprehensible diff that I could manually scan through and make sure nothing unexpected was happening.

sryza:
> I found success using git diff --diff-algorithm=histogram --color-moved=dimmed-zebra for a somewhat more comprehensible diff that I could manually scan through and make sure nothing unexpected was happening.

Oo nice

PR #8953, opened 2022-07-19
[1.0] Get rid of pipeline/solid in backcompat tests

dpeng817:
Given that they are no longer supported in 1.0

### DIFF
--- a/.buildkite/dagster-buildkite/dagster_buildkite/steps/integration.py
+++ b/.buildkite/dagster-buildkite/dagster_buildkite/steps/integration.py
@@ -85,7 +85,7 @@ def backcompat_extra_cmds(_, factor: str) -> List[str]:
     return [
         f"export EARLIEST_TESTED_RELEASE={EARLIEST_TESTED_RELEASE}",
         "pushd integration_tests/test_suites/backcompat-test-suite/dagit_service",
-        f"./build.sh {dagit_version} {user_code_version}",
+        f"./build.sh {dagit_version} {user_code_version} {_extract_major_version(user_code_version)}",
         "docker-compose up -d --remove-orphans",  # clean up in hooks/pre-exit
         *network_buildkite_container("dagit_service_network"),
         *connect_sibling_docker_container(
@@ -95,6 +95,13 @@ def backcompat_extra_cmds(_, factor: str) -> List[str]:
         ),
         "popd",
     ]
+
+
+def _extract_major_version(release):
+    """Returns major version if 0.x.x release, returns 'current_branch' if master."""
+    if release == "current_branch":
+        return release
+    return release.split(".")[0]
 
 
 # ########################
--- a/integration_tests/test_suites/backcompat-test-suite/dagit_service/Dockerfile_user_code_legacy_release
+++ b/integration_tests/test_suites/backcompat-test-suite/dagit_service/Dockerfile_user_code_legacy_release
@@ -1,0 +1,23 @@
+FROM python:3.8-slim
+
+ARG USER_CODE_BACKCOMPAT_VERSION
+
+COPY pins.txt pins.txt
+
+RUN pip install \
+    -r pins.txt \
+    dagster=="${USER_CODE_BACKCOMPAT_VERSION}" \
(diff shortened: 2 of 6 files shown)


alangenfeld (changes requested):
i assume this is automated? I think its useful to include in the pr summary how you performed it if thats the case. Sending back to you since theres merge conflicts.

dpeng817:
cc @alangenfeld massive file list was from a different PR.

smackesey:
`pipeline`/`solid` are unsupported in 1.0 but aren't these backcompat tests about cross-process communication between different `dagit` and `dagster` versions? In the test matrix there are four combos:

```
RELEASE_TEST_MAP = {
    "dagit-earliest-release": [EARLIEST_TESTED_RELEASE, DAGSTER_CURRENT_BRANCH],
    "user-code-earliest-release": [DAGSTER_CURRENT_BRANCH, EARLIEST_TESTED_RELEASE],
    "dagit-latest-release": [MOST_RECENT_RELEASE_PLACEHOLDER, DAGSTER_CURRENT_BRANCH],
    "user-code-latest-release": [DAGSTER_CURRENT_BRANCH, MOST_RECENT_RELEASE_PLACEHOLDER],
}
```

Looks like `repo.py` is part of the user code, which sometimes in "user-code-earliest-release" is running older dagster versions that still support pipeline.

IIUC seems like the solution here is to actually use a different repo.py for user code earliest release (which should still be pipeline-based) and user cade latest release (which would contain your changes).

dpeng817:
> `pipeline`/`solid` are unsupported in 1.0 but aren't these backcompat tests about cross-process communication between different `dagit` and `dagster` versions? In the test matrix there are four combos:
> 
> ```
> RELEASE_TEST_MAP = {
>     "dagit-earliest-release": [EARLIEST_TESTED_RELEASE, DAGSTER_CURRENT_BRANCH],
>     "user-code-earliest-release": [DAGSTER_CURRENT_BRANCH, EARLIEST_TESTED_RELEASE],
>     "dagit-latest-release": [MOST_RECENT_RELEASE_PLACEHOLDER, DAGSTER_CURRENT_BRANCH],
>     "user-code-latest-release": [DAGSTER_CURRENT_BRANCH, MOST_RECENT_RELEASE_PLACEHOLDER],
> }
> ```
> 
> Looks like `repo.py` is part of the user code, which sometimes in "user-code-earliest-release" is running older dagster versions that still support pipeline.
> 
> IIUC seems like the solution here is to actually use a different repo.py for user code earliest release (which should still be pipeline-based) and user cade latest release (which would contain your changes).

cc @alangenfeld for opinions here, but my take was that we would stop explicitly supporting using pipeline/solid and having that being interpretable by the front end post-1.0. As in, if you upgrade your user code deployment and you can no longer use pipeline/solid, it seems silly that the front-end has to explicitly support functionality which is gone in the current version. But maybe it's always necessary for the purpose of interpreting old snapshots, so willing to be overridden here.

alangenfeld:
ya i think we should continue to back compat test pipeline/solid in the "earliest version" and explicitly change that if we hit a breaking change we feel we need to make.

One argument is that if `dagster._legacy` is in `1.0` and has `pipeline` and `solid` in it, and we want to say we support user code from any `1.0` release then they should continue to work in dagit. 

We should probably add `1.0.0` entries to the test matrix too around launch. 

I doubt we will want to break these cross process loading structures anytime soon, but I do think itll be nice for the tests to make that decision explicit.

sryza:
> One argument is that if dagster._legacy is in 1.0 and has pipeline and solid in it, and we want to say we support user code from any 1.0 release then they should continue to work in dagit.

We don't want to support users importing from dagster._legacy, full stop. The only reason for it is to save us from migrating all our tests immediately.

alangenfeld:
Got it - my misunderstanding. Then I suppose the utility of continuing to test that we don't break legacy user code in current host processes is catching it earlier than internal variants of the test, until we ensure all of our cloud customers are  running >= 1.0 (or at least have no more active legacy definitions).

alangenfeld (approved):

PR #8960, opened 2022-07-19
Docs build overhaul

smackesey:
### Summary & Motivation

This PR updates and cleans up our docs code and README with a particular focus on Sphinx and the Screenshot management code. It attempts to remove superfluous or outdated config/extensions, ease the workflow for iterating on API docs, update to newer versions of sphinx and deps, and thoroughly explain how everything fits together in the README.

There are a lot of changes listed, and it's possible I didn't understand everything, so anything here is of course open to reversion.

- README
    - Reorganized and added a lot of explanatory content.
- API docs (sphinx) build environment:
    - Removed `py38` factor from `sphinx` tox environment. This means sphinx will attempt to use whatever Python version you're using (as of now it works on 3.8, 3.9, and 3.10). This was mainly for consistency with other tox configurations throughout the dagster codebase.
    - Bumped sphinx version from 3.5.2 to 5.0.2
    - Bumped sphinx-click version from 2.5.0 to 4.3.0
    - Removed nbsphinx package/extension. We do not use it.
    - Removed recommonmark extension. This is used to parse `.md` files alongside `.rst` files, but we don't actually have any `.md` files in our API docs.
    - Delete `docs-requirements.txt` and no longer install all target modules (i.e. dagster libs) into the environment with `pip`. Instead, place them on `sys.path` in `conf.py`. This is sphinx recommended way of making them available to autodoc. It also allows us to build the sphinx environment faster and more predictably (with a single `pip install` command instead of the previously required two due to use of `--no-deps`).
    - Delete `docs-build-requirements.txt`, fold its content into the `sphinx` tox env dependencies, and edit the list. In particular I removed many one-off dependencies where it was unclear why they were there, and replaced them with editable installs of the dagster package they service. There are two reasons it is preferable to keep the list in the tox env deps instead of a requirements file: (1) it allows tox to track changes to the dependencies and only rebuild the sphinx env when needed; (2) it's simpler.
- Sphinx config:
    - Remove `nbsphinx`, `recommonmark`, `intersphinx`, `ifconfig` extensions.
    - Remove a large amount of boilerplate config related to HTML and latex output. Pretty sure this stuff was added by the config generator for an old version of sphinx, but we don't use any of it since we only build JSON.
    - Removed a bunch of unnecessary entries from `exclude`. The patterns here matched various JS items but `exclude_patterns` is relative to the API doc source root, which does not have any JS stuff.
    - Removed HTML templates in `_templates`, since we don't use them.
- Next
[...]

PR #8961, opened 2022-07-19
python_modules/tools and tool environment changes

smackesey:
### Summary & Motivation

This PR does several things related to tooling and Python environments. The overarching goals are to:

- Close the gap between local dev and CI lint/test output 
- Reduce the probability of randomly occurring dependency collisions and build errors
- Make dagster's extensions to dev tools easier to discover, manage, comprehend, and
  modify

Changes:

- Breaks up `dagster[test]` into tool-specific sets of dependencies. This was already partially
  done, but the process is now complete. This enables each tool to run in an environment with only
  the dependencies it needs, rather than being subject to the dependency constraints of every
  package in our monorepo. In place of `dagster[test]` there is now:
    - `dagster[black]` (was already there)
    - `dagster[isort]` (was already there)
    - `dagster[mypy]` (was already there)
    - `dagster[pylint]` uses `pylint` and `astroid`
    - `dagster[pytest]` takes bulk of packages from `dagster[test]`
    - 
- Breaks `pylint` and `astroid` out of `dagster[test]` and puts them into a new `extra` pylint. This
  matches other tools and shrinks `[test]`.
- Creates a new directory `python_modules/tools` and places two new packages under it:
    - `dagster-pylint` includes Dagster's pylint extension, which was previously encoded within
      `dagster` itself under `dagster.utils`. There was no reason to include it there because this
      package is intended for checking Dagster's own code, not user code.
    - `dagster-sphinx` is a Sphinx extension that we can use to customize our Sphinx builds. It is
      seeded with the `autodoc-configurable` extension written by @owenkephart for rendering config
      schemas (previously directly included in `docs/sphinx`), but I am adding additional
      functionality upstack.
- Updates toxfiles to use only tool-scoped environments. Previously `dagster[test]` and
  `dagster[mypy]` were installed into a common testenv for each project. Now `mypy`, `pylint`, and
  `pytest` will run in environments with only the packages they need. This speeds environment
  creation and lessens the probability of weird hard-to-debug dependency collisions.
- To support the toxfile change, the default tox environments no longer invoke pytest-- you now run
  `tox -e pytest`.
- `dagster-buildkite` updated for compatibility with the toxfile changes.

- Added commands to the root `Makefile` to easily execute tools in the same environment that
  buildkite runs them in. This means running a tool in its specified tox environment instead of the
  "common" virtualenv built with `scripts/install_dev_python_modules.py`. This should reduce
  mismatch between local tool results and BK tool results.
    - `make tox <package> <env>` will run the specified tox env for the specified package. Packages
      can be referred to by by path, basename, or basename minus a `dagster-` prefix:

```python
# These three forms are equivalent
make tox python_modules/libraries/dagster-snowflake pylint
make tox dagster-snowflake pylint
make tox snowflake pylint
```

Note that the first time a tox environment is used, it has to be built, which can take a minute or
so. So the first run of a given tool/library combo will be slow. Subsequent runs will reuse the tox env
(unless changes occur to its dependencies) and be only imperceptibly slower than a direct execution.

BK


smackesey:
* master
  * **PR #8960** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8960" target="_blank">Graphite</a> 
    * **PR #8961** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8961" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8961?utm_source=stack-comment).

cmpadden:
Closing due to inactivity; please re-open if necessary, thank you

PR #8963, opened 2022-07-19
[dagit] Fix “Related Assets” tag for asset jobs, hide Launchpad

bengotow:
### Summary & Motivation

This PR hides the "Launchpad" tab for asset jobs, and switches the "Related Assets" tag shown in the job / pipeline nav to reflect the actual assetNodes of the job, rather than the assets run in the last 5 jobs (the implementation when it appears on op jobs.)

This required a bit of a refactor because the app actually doesn't /know/ if the job is an asset job until it loads the ops and sees that they have assetNodes defined. I don't think we want to load `isAssetJob` for every job instance-wide (the way we do `isJob`) because the graphql resolver seems expensive.

The downside of hiding this tab based on async-loaded data is that the Launchpad tab appears a moment after the others do for normal op jobs. I'm not sure of a good workaround for this. We could make it less noticeable by putting the Launchpad tab at the end of the tab list. 

Video of the change here: https://www.loom.com/share/52115dbb64b54a88a67f3c665040209e

### DIFF
--- a/js_modules/dagit/packages/core/src/nav/PipelineNav.tsx
+++ b/js_modules/dagit/packages/core/src/nav/PipelineNav.tsx
@@ -13,7 +13,7 @@ import {isThisThingAJob, useRepository} from '../workspace/WorkspaceContext';
 import {RepoAddress} from '../workspace/types';
 import {workspacePathFromAddress} from '../workspace/workspacePath';
 
-import {JobMetadata} from './JobMetadata';
+import {JobMetadata, useJobNavMetadata} from './JobMetadata';
 import {RepositoryLink} from './RepositoryLink';
 
 interface TabConfig {
@@ -78,6 +78,7 @@ export const PipelineNav: React.FC<Props> = (props) => {
   const {repoAddress} = props;
   const permissions = usePermissions();
   const repo = useRepository(repoAddress);
+
   const match = useRouteMatch<{tab?: string; selector: string}>([
     '/workspace/:repoPath/pipelines/:selector/:tab?',
     '/workspace/:repoPath/jobs/:selector/:tab?',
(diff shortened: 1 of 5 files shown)


js_modules/dagit/packages/core/src/nav/PipelineNav.tsx
```diff
@@ -88,15 +89,19 @@ export const PipelineNav: React.FC<Props> = (props) => {
...
   const hasPartitionSet = partitionSets.some(
     (partitionSet) => partitionSet.pipelineName === pipelineName,
   );
 
+  const navMetadata = useJobNavMetadata(repoAddress, pipelineName);
+  const hasLaunchpad = navMetadata.assetNodes && navMetadata.assetNodes.length === 0;
+
   const tabs = currentOrder
+    .filter((key) => hasLaunchpad || key !== 'playground')
     .filter((key) => hasPartitionSet || key !== 'partitions')
```
hellendag, line 105: Combine?

hellendag (approved):

peay:
@bengotow we've just run into this change after an upgrade (which, to be fair, is clearly listed in the CHANGELOG), and I have been wondering, is there a way to get the Launchpad back? 😁 

The issue is that the new "Materialize" workflow does not yet support everything. For instance, [when assets are partitioned, it does not allow specifying any configuration](https://github.com/dagster-io/dagster/issues/9467). When mixing a non-partitioned and a partitioned asset in a job, there are also some limitations.

The Launchpad was a fine workaround for all of these until Dagit catches up, but its removal is a blocker to upgrade when sometimes using custom configuration.

Note that the Overview tab still allows toggling between asset graph, and operation graph - maybe that's another argument for keeping the Launchpad, since one can still decide to work with ops?

bengotow:
Hey thanks for the feedback @peay! I think you're right that partitions + config is the loophole in the current Materialize button. Let me check with @sryza and see what we can do. (I think we could restore the launchpad, but we could also fast-track work to fix the materialize button.) I do like the idea of tying it to the assets vs ops selection in the DAG UI.

peay:
@bengotow sounds good.

Another remark is that for an asset job (e.g., after clicking on Materialize on an asset), there is still the button "Open in Launchpad" in the Run view. Although it opens the launchpad, it seems to be partly broken as for a time partitioned asset, it doesn't seem to set the partition from the run back in the launchpad. It also shows the Launchpad, although there is no "Launchpad" tab in the tab bar.

PR #8976, opened 2022-07-20
Keyword-only args for decorators and associated defs

smackesey:
### Summary & Motivation

This PR adds bare asterisks to enforce keyword-only args for decorators and their associated definitions.

This is a pattern that we may want to apply in many places in the codebase in the for 1.0. The main reason is that it provides us greater flexibility to change/add to function signatures going forward. Without enforcing keyword-only arguments, it is technically possible for a user to call e.g. `@op` like this:

```python
@op(
  "my_op",
  op_description,
  op_ins,
  op_outs,
  op_config_schema,
  ...
)
def my_op():
    ...
```

This means adding a new argument anywhere but the end of the arglist is technically a breaking change. With keyword-only args order no longer matters and we have more flexibility in adding to the signature.

A bonus is that our API docs become a bit less confusing since the first arg of a decorator no longer needs to do double-duty as a decorated function and a decorator parameter.

There are likely several places where the first one or two decorator params are frequently used positionally-- in those cases we should consider moving the bare asterisk behind these. We probably also need to look at every public API signature.

However, to get started here I went with a high-priority subset of the API (decorators in `dagster.core.definitions.decorators` except solid/pipeline ones that are going away) and their associated definition classes. 

I also used the simplest possible strategy of:

- Making all decorator params keyword only.
- Making all underlying `XDefinition` constructor params keyword-only.

We can discuss individual cases where we may want to deviate from this in the comments.

Note also that this PR just changes the signatures, not all the internal callsites (so there will be test failures). When we finalize our decision I can create another PR to change the internal callsites and stack this PR on top of that one.

Existing BK suite, nothing changed yet.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/assets.py
@@ -78,6 +78,7 @@ class AssetsDefinition(ResourceAddable):
 
     def __init__(
         self,
+        *,
         keys_by_input_name: Mapping[str, AssetKey],
         keys_by_output_name: Mapping[str, AssetKey],
         node_def: NodeDefinition,
@@ -187,6 +188,7 @@ class AssetsDefinition(ResourceAddable):
     @staticmethod
     def from_graph(
         graph_def: "GraphDefinition",
+        *,
         keys_by_input_name: Optional[Mapping[str, AssetKey]] = None,
         keys_by_output_name: Optional[Mapping[str, AssetKey]] = None,
         key_prefix: Optional[CoercibleToAssetKeyPrefix] = None,
@@ -255,6 +257,7 @@ class AssetsDefinition(ResourceAddable):
     @staticmethod
     def from_op(
         op_def: OpDefinition,
+        *,
         keys_by_input_name: Optional[Mapping[str, AssetKey]] = None,
         keys_by_output_name: Optional[Mapping[str, AssetKey]] = None,
         key_prefix: Optional[CoercibleToAssetKeyPrefix] = None,
@@ -314,6 +317,7 @@ class AssetsDefinition(ResourceAddable):
     @staticmethod
     def _from_node(
         node_def: Union[OpDefinition, "GraphDefinition"],
+        *,
         keys_by_input_name: Optional[Mapping[str, AssetKey]] = None,
         keys_by_output_name: Optional[Mapping[str, AssetKey]] = None,
         internal_asset_deps: Optional[Mapping[str, Set[AssetKey]]] = None,
--- a/python_modules/dagster/dagster/_core/definitions/assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/assets.py
@@ -241,15 +241,15 @@ class AssetsDefinition(ResourceAddable):
         if resource_defs is not None:
             experimental_arg_warning("resource_defs", "AssetsDefinition.from_graph")
         return AssetsDefinition._from_node(
-            graph_def,
-            keys_by_input_name,
-            keys_by_output_name,
-            internal_asset_deps,
-            partitions_def,
-            group_name,
-            resource_defs,
-            partition_mappings,
-            metadata_by_output_name,
+            node_def=graph_def,
+            keys_by_input_name=keys_by_input_name,
+            keys_by_output_name=keys_by_output_name,
+            internal_asset_deps=internal_asset_deps,
+            partitions_def=partitions_def,
+            group_name=group_name,
+            resource_defs=resource_defs,
+            partition_mappings=partition_mappings,
+            metadata_by_output_name=metadata_by_output_name,
             key_prefix=key_prefix,
         )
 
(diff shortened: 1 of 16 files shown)


smackesey:
* master
  * **PR #8976** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8976" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8976?utm_source=stack-comment).

dpeng817:
A few things:
- how does this manifest in docs? IE, when I look at a method whose constructor uses this, will I see an asterisk?
- could we throw a useful error if args are non-empty? (Assuming this would require actually having a named argument instead of just *)

Personally, I think that the typing clarity alone is worth it to users. I think the main questions for me are 
1. can we get error messages s.t. it's clear to users who are upgrading what to fix
2. are we actually okay with making a breakage this large scale.

gibsondan (commented):
Is this likely to break a lot of users' existing code when they upgrade? Or is it more of an edge-case feature that most users won't notice

python_modules/dagster/dagster/core/definitions/decorators/schedule_decorator.py
```diff
@@ -48,6 +48,7 @@
 
 
 def schedule(
+    *,
```
sryza, line 51: I think we should move this wildcard one argument down.

python_modules/dagster/dagster/core/definitions/decorators/schedule_decorator.py
```diff
@@ -195,6 +196,7 @@ def _wrapped_fn(context: ScheduleEvaluationContext) -> RunRequestIterator:
 
 
 def monthly_schedule(
```
sryza, line 198: I think we should move this wildcard one argument down. That said, we'll probably remove these because they're pre-job/op.

python_modules/dagster/dagster/core/definitions/decorators/sensor_decorator.py
```diff
@@ -97,6 +98,7 @@ def inner(fn: RawSensorEvaluationFunction) -> SensorDefinition:
 
 
 def asset_sensor(
+    *,
```
sryza, line 101: I think we should move this one argument down

python_modules/dagster/dagster/core/definitions/graph_definition.py
```diff
@@ -190,6 +190,7 @@ def add_one(num):
 
     def __init__(
         self,
+        *,
```
sryza, line 193: One down

python_modules/dagster/dagster/core/definitions/repository_definition.py
```diff
@@ -1110,6 +1110,7 @@ class RepositoryDefinition:
 
     def __init__(
         self,
+        *,
```
sryza, line 1113: One down

python_modules/dagster/dagster/core/definitions/schedule_definition.py
```diff
@@ -230,6 +230,7 @@ class ScheduleDefinition:
 
     def __init__(
         self,
+        *,
```
sryza, line 233: One down

python_modules/dagster/dagster/core/definitions/sensor_definition.py
```diff
@@ -191,6 +191,7 @@ class SensorDefinition:
 
     def __init__(
         self,
+        *,
```
sryza, line 194: One down

smackesey:
@dpeng817 

>how does this manifest in docs? IE, when I look at a method whose constructor uses this, will I see an asterisk?

Built the docs locally on this branch, the signature includes the asterisk (which is good IMO):

``
@dagster.op(compute_fn=None, *, name=None, description=None, ins=None, out=None, config_schema=None, required_resource_keys=None, tags=None, version=None, retry_policy=None, input_defs=None, output_defs=None)[[source]](http://localhost:3001/_modules/dagster/core/definitions/decorators/op_decorator#op)
```

The param list does not.

>could we throw a useful error if args are non-empty? (Assuming this would require actually having a named argument instead of just *)

We could do something like this:

```python
def reject_decorator_pos_args(fn: T_Callable) -> T_Callable:

    argspec = inspect.getfullargspec(fn)

    @wraps(fn)
    def _(*args, **kwargs):
        if len(args) > len(argspec.args):
            raise InformativeErrorMessage()
        return fn(*args, **kwargs)

    return _

# ...

@reject_decorator_pos_args
def op(...):
    pass
```

Although I'm not sure this is better, bc I think by gaining this nice runtime error we _lose_ linter/typechecker flagging of too many pos args-- though maybe not in all cases if we can fool the linter/typechecker.

@gibsondan 

>Is this likely to break a lot of users' existing code when they upgrade? Or is it more of an edge-case feature that most users won't notice

That depends on the case-by-case decisions. With the crude approach currently in this PR, yes we would probably break a lot of users. If we judiciously move the asterisks back a few args when necessary, then we'll prob break very few.

dpeng817:
Agreed that including the asterisk is good - I think it makes it clear where the kwargs begin and why someone might be hitting an error. Regarding error messages, I see your point regarding losing static type checking capabilities, which is probably more important. I'm good with this, will leave final sign off for now just in case we want to change position of asterisks any further.

sryza (commented):
To be conservative, I lean towarsd putting name and description in most of these above the wildcard. Does that cause problems?

smackesey:
>To be conservative, I lean towarsd putting name and description in most of these above the wildcard. Does that cause problems?

For decorators, it does, as that requires us to recombine `compute_fn` and `name` in the first positional argument, which IMO is bad for docs clarity. For definition classes, it's not a problem.

sryza (approved):
Ok, I think it's fine. 🤞 this doesn't cause too much trouble.

PR #8978, opened 2022-07-20
Thread config during asset selection

clairelin135:
Reported by a user ([link](https://elementl-workspace.slack.com/archives/C03CCE471E0/p1658179015802579))

Default config that is provided in `define_asset_job` is not threaded through when an asset selection occurs. This PR amends this behavior by passing in this default config.

Other fields were also missing (e.g. metadata, loggers, etc.) which don't currently affect users because there's currently no way to specify these fields on asset jobs. This PR threads these fields to the subsetted job so that these fields will be respected should they exist in the future.

Additionally, if default config is provided for an asset that is not part of the asset selection, this PR ignores the config for these de-selected assets. This is existing behavior for op/solid selection.


sryza:
@OwenKephart mind reviewing this? I probably won't be able to get to it today. Seems like this should go into today's release?

OwenKephart (approved):
looks good (sorry my email checking frequency is low, didn't see this until just now)

PR #8982, opened 2022-07-21
Bug fix with run status sensors that return run requests

jamiedemaria:
### Summary & Motivation
User reported bug when a run status sensor returned a PipelineRunReaction causes a check failure. This happened with returning RunRequests as well because instead of returning a RunRequest (or pipeline run reaction, etc), the individual attributes of the run request named tuple were returned. 

I have no idea how this got through testing (i tested the exact case that replicates the bug a bunch of times when originally developing this feature) so, if possible, i'd like to add some kind of integration test that can check for this in the future. However, i'm not sure where such tests exist in the code base if they do exist. lmk if there's somewhere i can look

### DIFF
--- a/python_modules/dagster-test/dagster_test/graph_job_op_toys/run_status_sensors.py
+++ b/python_modules/dagster-test/dagster_test/graph_job_op_toys/run_status_sensors.py
@@ -43,9 +43,31 @@ def status_job():
 
 
 @run_status_sensor(pipeline_run_status=DagsterRunStatus.SUCCESS, request_job=status_job)
-def succeeds_sensor(context):
+def yield_run_request_succeeds_sensor(context):
+    """
+    We recommend returning RunRequests, but it's possible to yield, so this is here to test it
+    """
     if context.dagster_run.pipeline_name != status_job.name:
         yield RunRequest(
+            run_key=None,
+            run_config={
+                "ops": {
+                    "status_printer": {
+                        "config": {
+                            "message": f"{context.dagster_run.pipeline_name} job succeeded!!!"
+                        }
+                    }
+                }
+            },
+        )
+    else:
+        yield SkipReason("Don't report status of status_job.")
+
+
+@run_status_sensor(pipeline_run_status=DagsterRunStatus.SUCCESS, request_job=status_job)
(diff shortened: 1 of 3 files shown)


python_modules/dagster-test/dagster_test/graph_job_op_toys/run_status_sensors.py
```diff
@@ -42,7 +45,7 @@ def status_job():
 @run_status_sensor(pipeline_run_status=DagsterRunStatus.SUCCESS, request_job=status_job)
 def succeeds_sensor(context):
     if context.dagster_run.pipeline_name != status_job.name:
-        return RunRequest(
+        yield RunRequest(
```
prha, line 48: ah... I think we should add a comment here to the effect of:

`keeping this as yield even though we encourage return values so that we test the thingggg`

And maybe split this out as a separate sensor called `yields_run_request_sensor`

python_modules/dagster/dagster/core/definitions/run_status_sensor_definition.py
```diff
@@ -546,7 +546,13 @@ def _wrapped_fn(context: SensorEvaluationContext):
                                     update_timestamp=update_timestamp.isoformat(),
                                 ).to_json()
                             )
-                            yield from sensor_return
+
+                            if isinstance(
+                                sensor_return, (RunRequest, SkipReason, PipelineRunReaction)
```
prha, line 551: can they also return a list of `RunRequest`?

prha (approved):
oh wow, this is a really good catch...

python_modules/dagster/dagster/core/definitions/run_status_sensor_definition.py
jamiedemaria: i think the `else: yield from sensor_return` should take care of that but i will confirm that constructing a list of run requests and returning the list (rather than yielding multiple run requests) works as expected

jamiedemaria: added a couple more types of run status sensors and confirmed they all work:
* yielding a single run request
* returning a single run request
* yield multiple run requests
* returning a list of run requests

PR #8983, opened 2022-07-21
Fix handling of launch path when a repository location is still loading

gibsondan:
Summary:
Adding some defensive coding here when you try to programatically launch a run for a repo location that is still loading.

- has_external_pipeline should return False when a repo location is still loading, not throw an exception
- If you do manage to reach get_repository_location somehow while the location is still loading, it should raise an exception explaining that this is why rather than an issue from trying to call to_string() on a load error that doesn't exist

Will add a test in an environment where it's more feasible to have a location in a LOADING state

### DIFF
--- a/python_modules/dagster/dagster/_core/workspace/context.py
+++ b/python_modules/dagster/dagster/_core/workspace/context.py
@@ -112,10 +112,15 @@ class BaseWorkspaceRequestContext(IWorkspace):
         if location_entry.repository_location:
             return location_entry.repository_location
 
-        error_info = cast(SerializableErrorInfo, location_entry.load_error)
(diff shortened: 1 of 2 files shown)


alangenfeld (dismissed):

rexledesma (changes requested):
we should ensure that

- the old location is still used for launching runs during update
- submitting runs do not fail during the update step

rexledesma (changes requested):
a test for the loading code location entry should be pretty easy with mocks.

Also maybe we should raise a catch-all exception in the case that none of the conditional branches in this method execute? Right now it could technically return `None`

gibsondan:
@rexledesma test added!

rexledesma (approved):

PR #8990, opened 2022-07-21
Add public/deprecated annotation decorators

smackesey:
### Summary & Motivation

This adds `@public` and `@deprecated` annotations to a new submodule. We can use these for marking the public API.

I also moved the `@experimental` decorator in here, added the same kind of tag as used for public/deprecated, and deleted `@experimental_decorator` (folded its functionality into `@experimental`), and deleted some outdated experimental tests (we should be using the decorator and not calling the class/fn warning methods manually).

I paired the decorators with `is_public`, `is_deprecated`, and `is_experimental` predicates, which should give us freedom to easily change the implementation if we want to later.

Added unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_annotations.py
+++ b/python_modules/dagster/dagster/_annotations.py
@@ -1,21 +1,113 @@
-from typing import Callable, TypeVar
+import inspect
+from functools import wraps
+from typing import Callable, Type, TypeVar, cast
+
+import dagster._check as check
+from dagster.utils.backcompat import (
+    experimental_class_warning,
+    experimental_decorator_warning,
+    experimental_fn_warning,
+)
 
 T_Callable = TypeVar("T_Callable", bound=Callable)
 
+##### PUBLIC
 
-def public(fn: T_Callable) -> T_Callable:
-    setattr(fn, "_public", True)
-    return fn
+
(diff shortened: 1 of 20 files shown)


smackesey:
* master
  * **PR #8990** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8990" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8990?utm_source=stack-comment).

python_modules/dagster/dagster/_annotations.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def public(obj: T_Callable) -> T_Callable:
+    """
+    Mark a method on a public class as public. This distinguishes the method from "internal"
+    methods, which are methods that are public in the Python sense of being non-underscored, but
+    not intended for user access. Only `public` methods of a class are rendered in the docs.
+    """
+    check.callable_param(obj, "obj")
+    setattr(obj, "_public", True)
```
sryza, line 24: `_is_x` is generally preferred for booleans

sryza (approved):
Had one comment - otherwise looks good

PR #8992, opened 2022-07-21
create a top-level "env_vars" key in container context

gibsondan:
Summary;
Allows you to have a top-level set of contaienr context config that isn't specific to a particular environment. env_vars is the first entry.


benpankow (approved):
thank you for this 🙏

Issue #8532, opened 2022-06-21
In asset node, latest run shouldn't point to a run that targeted the multi-asset but didn't target the asset


sryza:
I navigated to an asset group with a set of dbt assets. I kicked off a materialization of one of them, and it failed. All of the dbt assets got their "latest run" updated to that run, and had the red warning box.

sryza:
@clairelin135 @bengotow do you know what would be required to fix this? Seems like it could be really easy or also could be really hard?

clairelin135:
@sryza Can you share a code example for this? I think I know the root cause but need to replicate

sryza:
Hmm, I tried to reproduce the error with this example, but was not successful:
```
from dagster import multi_asset, AssetOut, Output


@multi_asset(
    outs={"a": AssetOut(is_required=False), "b": AssetOut(is_required=False)}, can_subset=True
)
def multi(context):
    raise ValueError("fdsjk")
    for output_name in context.selected_output_names:
        yield Output(None, output_name)
```

The I tried it again on Purina and got the same problem. I tried to launch a materialization of just one of the dbt assets, but it appears that the run is actually trying to target all of the assets, based on the ops that are included in the run and the assets displayed at the top of the page.

clairelin135:
Both @OwenKephart and I have taken time to diagnose and replicate this issue. We've tried a lot of things, including:
- creating assets with the same structure as the purina DAG
- binding those assets to resources
- executing with docker run launcher and k8s run launcher
We've spent a significant amount of time to look into this and have still been unable to replicate this issue outside of the original purina DAG, it seems like there are probably rare conditions required for the bug to occur. 

I'll take time in the future to try and resolve this, but due to more urgent things on my plate I'll need to bump on this.

clairelin135:
@bengotow I was finally able to diagnose this error. Turns out that in Dagit, when an asset job requires config and the job is executed via the launchpad, [we supply a `SolidSelection`](https://github.com/dagster-io/dagster/blob/1961e51615540111d9cb16e540485bc598ceae13/js_modules/dagit/packages/core/src/launchpad/LaunchpadSession.tsx#L184) rather than an `AssetSelection`. 

We'd want to use a [pipeline selector like this one instead](https://github.com/dagster-io/dagster/blob/1961e51615540111d9cb16e540485bc598ceae13/js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx#L299), that provides an asset selection.

[...]

PR #8994, opened 2022-07-21
Fix ReconstructablePipeline asset subselection bug

clairelin135:
I dug back into https://github.com/dagster-io/dagster/issues/8532 by looking back on the debug logs of this [run](https://elementl.dagster.cloud/prod/instance/runs/6d1aa926-8927-42f5-aa98-db122ff7c3c6) and we indeed did provide an `asset_selection` instead of a `solid_selection`. Turns out that the config launchpad providing `solid_selection` was a related bug but not the root cause of this issue. 

The root cause was that some run launchers such as the `K8sRunLauncher` that Cloud uses by default, we first create the run (with an asset selection) and then reconstruct the job via an execution snapshot created without the asset selection. This causes a mismatch between the selected steps/resources between the run and the execution plan. We needed to instead recreate the execution plan from the subsetted job. This PR makes this fix and adds a test for it.

This PR should resolve https://github.com/dagster-io/dagster/issues/8532.


OwenKephart (approved):
nice 🥇

PR #8998, opened 2022-07-22
Fix OpDefinition constructor

dpeng817:
Adds ins/outs arguments to OpDefinition constructor.

A bunch of gross logic for resolving between using input_defs/output_defs and ins/outs, which will go away in a 1.0 PR. 

I intentionally named the parameter on OpDefinition `outs` rather than `out`, to match with the outs property. Given that constructing OpDefinition directly is pretty advanced, I think it's fine for us to not include the convenience stuff that we have on the decorator, since it makes the internal logic simpler.


smackesey (approved):
bootiful

PR #8999, opened 2022-07-22
Remove submodule imports from doc snippets

smackesey:
### Summary & Motivation

This removes all submodule imports from docs snippets. In order to do this, I had to move imports for `script_relative_path` and `load_yaml_from_path` out of the imports actually shown to the user (these come from `dagster.utils`).

While this means the snippets that use these functions are not 100% self-contained, I think it is the best option-- we don't want to export these kinds of util functions from top-level dagster and it would clutter the docs to define them inline.

BK


smackesey:
* master
  * **PR #8999** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8999" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/8999?utm_source=stack-comment).

dpeng817 (approved):
Nice - I think this is the right move. Having to support `script_relative_path` at the top level had me rolling my eyes

PR #9002, opened 2022-07-22
add ability to customize disabled reason for permissions from server

gibsondan:

### DIFF
--- a/js_modules/dagit/packages/core/src/graphql/schema.graphql
+++ b/js_modules/dagit/packages/core/src/graphql/schema.graphql
@@ -2243,6 +2243,7 @@ type PartitionBackfills {
 type Permission {
   permission: String!
   value: Boolean!
+  disabledReason: String
 }
 
 type AssetLatestInfo {
--- a/python_modules/dagster-graphql/dagster_graphql/schema/permissions.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/permissions.py
@@ -1,6 +1,7 @@
 import graphene
 
 import dagster._check as check
+from dagster._core.workspace.permissions import PermissionResult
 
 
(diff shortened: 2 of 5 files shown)


gibsondan:
* master
  * **PR #9002** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9002" target="_blank">Graphite</a>  👈
    * **PR #9009** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9009" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9002?utm_source=stack-comment).

python_modules/dagster/dagster/core/workspace/context.py
```diff
@@ -92,7 +92,7 @@ def version(self) -> Optional[str]:
 
     @property
     @abstractmethod
-    def permissions(self) -> Dict[str, bool]:
+    def permissions(self) -> Dict[str, PermissionResult]:
```
alangenfeld, line 95: some concern for this api change - i think mypy won't care if `PermissionResult` gets checked against directly as a bool which will always be True if its a tuple instance 

maybe `PermissionResult` should be a class impl that overrides `__bool__` / `__nonzero__` to handle checks against the object instead of the `enabled` property

gibsondan: @alangenfeld does the fact that there are only 2 callsites dull your concern at all? one in this file, the other in graphql for generating the perms for dagit

alangenfeld: that does help, though even just raising on `__bool__` / `__nonzero__` defensively is not a big code change so i think its worth considering

alangenfeld (approved):

PR #9005, opened 2022-07-22
[dagit] Improve asset lineage copy, default to hiding “secondary", show groups

bengotow:
### Summary & Motivation

Before:
image

After:
image

The only "link"-style assets you'll see on the lineage graph are source assets, which are not defined in any repo.


hellendag (approved):

sryza:
I think this is a big improvement:

A few thoughts:
- What would you think about "+" and "-" instead of arrows for incrementing/decrementing the depth?
- It looks weird to me that "Graph depth" is to the right of the depth chooser.
- When we're not displaying full depth, would there be some way to indicate that there are assets in the lineage that we're not showing? E.g. some sort of persistent toast inside at the top/bottom of the asset graph that says "Not all upstream/downstream assets show. Increase the depth to show more.".  Right now, if you know what you're doing, you can figure it out from the graying out of the increment/decrement buttons, but I don't think that would be obvious to newcomers to the UI.

bengotow:
Sorry for the delay on this one, I've applied the changes you suggested and I think it looks great. The -/+ buttons definitely make more sense than the arrows 👍 

image

sryza:
I like it!

PR #9006, opened 2022-07-22
[dagit] Switch to updated “segmented control” styles for ButtonGroup

bengotow:
### Summary & Motivation

This is a small PR that updates ButtonGroup to reflect the styling we discussed on Slack last month: 
https://elementl-workspace.slack.com/archives/C03CCE471E0/p1656010228830629

The goal of this change is to bring the border styling in line with our other components and make the button group look better when it's displayed in a toolbar with other inputs / buttons.

image

It also updates "JoinedButtons" so that the buttons are touching and there is no hairline gap between them.

Before:
image

Before:
image

After:
image

After:
image


hellendag (approved):
Sounds good to me!

PR #9008, opened 2022-07-22
consistent sorting for container context values

gibsondan:
Summary:
Noticed a test flake that blamed to this - we should probably make this conssitent since they end up serialized in runs

### DIFF
--- a/python_modules/libraries/dagster-k8s/dagster_k8s/container_context.py
+++ b/python_modules/libraries/dagster-k8s/dagster_k8s/container_context.py
@@ -18,7 +18,7 @@ from .models import k8s_snake_case_dict
 
 
 def _dedupe_list(values):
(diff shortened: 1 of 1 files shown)


alangenfeld (approved):

python_modules/libraries/dagster-k8s/dagster_k8s/container_context.py
```diff
@@ -18,7 +18,9 @@
 
 
 def _dedupe_list(values):
-    return list(set([make_readonly_value(value) for value in values]))
+    return sorted(
+        list(set([make_readonly_value(value) for value in values])), key=lambda val: hash(val)
+    )
```
alangenfeld, line 23: can just use an `OrderedSet` too

PR #9011, opened 2022-07-22
modern_data_stack_assets example updates

sryza:
I'm working on including code and a screenshot from the MDS example on dagster.io.

This PR makes a set of changes to the MDS example that I think make it more exemplary:
- IO managers should be named for the storage systems they encapsulate, not the types they handle.
- Took out the input manager. IMO this the current example is not a realistic situation where it makes sense to define a separate input manager, because it's easy enough to just call `to_numpy` on the dataframe.
- Avoided non-top-level Dagster imports.
- Moved resources into a separate module, so human eyes don't need to process as much at once.

I loaded this up in dagit, but shamefully have not tried to run it because of all the setup steps involved. I could be convinced to do so.

### DIFF
--- a/examples/modern_data_stack_assets/modern_data_stack_assets/assets.py
+++ b/examples/modern_data_stack_assets/modern_data_stack_assets/assets.py
@@ -1,16 +1,13 @@
-from typing import Any, Tuple
-
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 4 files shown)


OwenKephart (approved):
looks much better 👍

PR #9012, opened 2022-07-23
Providing an empty-string group_name for assets should be avoided

### LINKED ISSUE #8667: we should error when an asset has an empty-string group name
right now, we silently convert to "default"

```
from dagster import asset


@asset(group_name="")
def asset():
    ...
```

HassenIO:
### Summary & Motivation

tl;dr: Providing an explicit empty-string asset `group_name` should result in an error instead of silently using the `default` group name.

More details can be found in this issue -> https://github.com/dagster-io/dagster/issues/8667

The logic behind this is that providing `group_name=""` is confusing, and probably an issue with the logic building the group name. We should explicitly provide notification to the user about that, inviting him to provide a valid group_name of use `None` for `default` value.

### Remarks

- This is going to break old assets where users were using empty-string group name values
  - Actually we silently default the name

- [x] I added a new unit test to check valid group name
- [x] Added this test to the `_Asset` class

### Todo

- [x] Create a check for asset group_name not empty
- [x] Test the newly created check
- [x] Add the check in the creation of the asset (ie `_Asset` class)
- [x] Document the new condition on asset group_name
- [ ] Maybe add a more global check on the creation of the asset with empty group_name

Closes #8667

### DIFF
--- a/docs/content/concepts/assets/software-defined-assets.mdx
+++ b/docs/content/concepts/assets/software-defined-assets.mdx
@@ -462,6 +462,8 @@ In Dagster, there are two ways to assign assets to groups:
 
 By default, assets that aren't assigned to a group will be placed in a group named `default`. Use Dagit to [view these assets](#viewing-asset-groups-in-dagit).
 
+Note that `group_name` if provided, should not be a blank value. Doing so will result in an error being raised.
+
 #### From assets in a sub-module
 
 **This recommended approach** constructs a group of assets from a specified module in your project. Using the `load_assets_from_package_module` function, you can import all assets in a module and apply a grouping:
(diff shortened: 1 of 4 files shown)


sryza (changes requested):
Thanks for submitting this PR @hassen-io!

A couple comments:
- The asset decorators aren't the only place that `AssetsDefinitions` can be created. If you put the check inside the constructor of `AssetsDefinition` (and `SourceAsset`), that would allow us to catch the problem in more situations.
- We generally keep "application-specific` logic out of the check module - i.e. it just checks basic types and doesn't know about Dagster concepts like groups. It would be better to define any helper function for checking groups outside of that module and closer to where the check occurs.
- Regarding `strip`, I don't think we want to allow any leading/trailing whitespace in group names  (but I think this is already checked somewhere).

PedramNavid:
closed by #17221

PR #9016, opened 2022-07-25
move test_sensor_run and test_scheduler_run off legacy APIs

sryza:
I had some time on a flight

### DIFF
--- a/python_modules/dagster/dagster_tests/daemon_tests/test_sensor_run.py
+++ b/python_modules/dagster/dagster_tests/daemon_tests/test_sensor_run.py
@@ -17,7 +17,10 @@ from dagster import (
     Field,
     Output,
     graph,
+    job,
+    op,
     pipeline_failure_sensor,
+    reconstructable,
     repository,
     run_failure_sensor,
 )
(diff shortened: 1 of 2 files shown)


gibsondan (approved):
thanks - seems clearly better, might we still want one test case that hits a pipeline though, as long as it's something we intend to not be broken?

sryza:
> might we still want one test case that hits a pipeline though, as long as it's something we intend to not be broken?

we're removing it for 1.0

gibsondan:
But we could still be executing code in the scheduler in cloud that was
generated using pipelines

sryza:
> But we could still be executing code in the scheduler in cloud that was generated using pipelines

By that logic though, we'd never be able to move our internals off of PipelineDefinitions and daily_schedules, right? The way we've been looking at it, the only reason we still have it around inside `dagster._legacy` is because it's too onerous to remove it from all the tests at once.

sryza:
I.e. I thought our philosophy was basically that backcompat tests should ensure that the interface between user code and host code is stable, so that we don't need to actually test every old version of user code against our host code?

gibsondan:
That sounds like a reasonable philosophy but I can't think of any existing tests today that actually do that. There's a test suite that runs a fixed set of code against various releases of dagster, but I don't think there's anything that would keep on testing that our system machinery still works with pipelines after we remove them from the codebase

gibsondan:
I guess the problem i'm describing is still a problem independent from this PR

gibsondan:
(the problem being that we wouldn't have a great way of knowing if we introduced a change in dagit that broke old users using @pipeline and @solid)

cmpadden:
Closing due to lack of activity.

PR #8975, opened 2022-07-20
[known state] fix build_for_reexecution bug

alangenfeld:
pretty nasty bug caused us to stuff a lot of garbage in to `dynamic_mappings` 

### How I Tested These Changes

updated runtime checks on passed in values

PR #9022, opened 2022-07-25
Add known_state bug fix to changelog

johannkm:
will cherry pick to release branch


johannkm (commented):
for https://github.com/dagster-io/dagster/pull/8975

alangenfeld (commented):
re: https://github.com/dagster-io/dagster/pull/8564#issuecomment-1190783302

I don't think my fix is enough to stop people from hitting the limit, so I didn't bother adding a changelog entry for it

PR #9031, opened 2022-07-25
Remove graph/job/op guide from docs site, link to old docs to reference

dpeng817:
In order to remove pipeline/solid APIs from the top-level API, we need to be able to get rid of these code examples. But especially for 1.0, the migration guide is actually quite important. As a result, we link to a previous version of the docs that still has the graph/job/op APIs referenced.


sryza:
Nice! I'm going to defer to @yuhan on this one, because she's likely to have a better understanding of any potential consequences of this witchcraft.

yuhan (approved):
verified the link works in preview: https://dagster-git-dpeng817-project10removegraphjobopexamples-elementl.vercel.app/guides

PR #9032, opened 2022-07-26
RFC: A way to launch runs without hitting the gRPC server, by delaying execution plan creation until the start of the run worker

gibsondan:
Summary:
This is a quick-and-dirty end-to-end prototype of launching runs without going through the gRPC server first. This could be useful to make dagster more serverless in various scenarios - you could spin up a server to generate metadata, then spin it down until you need to create a run, possibly much later.

There are some downsides/caveats/tradeoffs though:
- It only works if you don't have any ops or assets selected - if you do, they have to go through the gRPC server first to get the right pipeline snapshot
- You lose some pre-run launch validation - for example, beofre, if you referenced an env var that didn't actually exist in the container where the run is happening, this grpc call would fail and you'd get an error before any run objects were created. There are actually some nice things about that though - now, the failed run would trigger alerts and I think appear in the factory floor, whereas before the failure would be pretty silent and stay as a failed scheduler tick that was otherwise not surfaced anywhere.

There are a couple other smaller things we would need to figure out that i put inline. Back-compat could also be a bit tricky, you could make this a setting or something that people opt into, in both OSS and cloud.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/schema/pipelines/pipeline.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/pipelines/pipeline.py
@@ -485,7 +485,7 @@ class GrapheneRun(graphene.ObjectType):
         return from_captured_log_data(log_data)
 
     def resolve_executionPlan(self, graphene_info: ResolveInfo):
-        if not (self.dagster_run.execution_plan_snapshot_id and self.dagster_run.job_snapshot_id):
+        if not self.dagster_run.execution_plan_snapshot_id:
             return None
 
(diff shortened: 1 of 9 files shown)


johannkm (commented):
Not much to comment here except I'm a fan.

+1 on having a failed run when missing an env var etc. We might want to remove that validation even for op selections, for more consistency

alangenfeld (changes requested):
definitely tricky on the back-compat front

shelving in your queue for now

prha (changes requested):
back to your queue

PR #8161, opened 2022-06-02
Op that runs a kubernetes job

gibsondan:
Summary:
This introduces a more direct analogue to a 'pod that manages a k8s container' - to handle cases where you want to stitch together a mixture of ops, or orchestrate non-Python languages.

A nice thing is that we can use container_context to still pass config through to ops like this (so you can set up a default set of secrets to include in all ops even when you're not launching the run in k8s, for example). And we can expose raw job_config on the op as well to give you access to the full k8s api for the job.

PR #9037, opened 2022-07-26
Fix overriding the command to use in container_config

gibsondan:
Summary:
Overriding the command in container_config changed in https://github.com/dagster-io/dagster/pull/8161 - we override the passed in command to the function, even if it is empty. Instead, fall back to the user-provided container_config if it is set.

Test Plan: BK

### DIFF
--- a/python_modules/libraries/dagster-k8s/dagster_k8s/job.py
+++ b/python_modules/libraries/dagster-k8s/dagster_k8s/job.py
@@ -528,8 +528,6 @@ def construct_dagster_k8s_job(
     pod_name=None,
     component=None,
     labels=None,
-    env_vars=None,
-    command=None,
 ):
     """Constructs a Kubernetes Job object for a dagster-graphql invocation.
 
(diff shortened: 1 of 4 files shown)


python_modules/libraries/dagster-k8s/dagster_k8s/job.py
```diff
@@ -635,6 +635,10 @@ def construct_dagster_k8s_job(
 
     resources = user_defined_resources if user_defined_resources else job_config.resources
 
+    user_defined_command = container_config.pop("command", None)
+    if not command:
+        command = user_defined_command
```
alangenfeld, line 640: `command` and `user_defined_k8s_config` args are not in the doc block for this method

probably be good to add them and explicitly call out the precedence order, maybe a comment here as well

alangenfeld (approved):

gibsondan:
new take based on alex's feedback, we didn't actually need that separate arg in the first place really

python_modules/libraries/dagster-k8s/dagster_k8s/job.py
```diff
@@ -545,7 +543,6 @@ def construct_dagster_k8s_job(
             app.kubernetes.io/component. Defaults to None.
         labels(Dict[str, str]): Additional labels to be attached to k8s jobs and pod templates.
             Long label values are may be truncated.
-        env_vars(Dict[str, str]): Additional environment variables to add to the K8s Container.
```
alangenfeld, line 548: user_defined_k8s_config still should get added I think

PR #9042, opened 2022-07-26
[dagster-dbt] fix bk

OwenKephart:
### Summary & Motivation

https://buildkite.com/dagster/dagster/builds/33162#01823bb0-a99c-46ca-9476-e1241b89341d

mysterious mypy / json decoding errors occur in python 3.7. The cast has no behavior change, and now if a string is returned when decoding from json (instead of the expected dictionary), the entire string will be logged (rather than trying to parse out a specific field from the blob).

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cli/utils.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cli/utils.py
@@ -119,8 +119,10 @@ def execute_cli(
             except json.JSONDecodeError:
                 pass
             else:
-                message = json_line.get("message", json_line.get("msg", message))
-                log_level = json_line.get("levelname", json_line.get("level", "debug"))
+                # in rare cases, the loaded json line may be a string rather than a dictionary
+                if isinstance(message, dict):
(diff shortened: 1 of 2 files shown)


python_modules/libraries/dagster-dbt/dagster_dbt/cli/utils.py
```diff
@@ -119,8 +119,10 @@ def execute_cli(
             except json.JSONDecodeError:
                 pass
             else:
-                message = json_line.get("message", json_line.get("msg", message))
-                log_level = json_line.get("levelname", json_line.get("level", "debug"))
+                # in rare cases, the loaded json line may be a string rather than a dictionary
+                if isinstance(message, dict):
```
dpeng817, line 123: what happens if it _is_ a string? should we parse it out?

OwenKephart: by default, message is set to the bare string (and log level to "info"). so if it's a string, the raw json line will be emitted (seems ok to me).

alangenfeld (approved):

PR #9043, opened 2022-07-26
Add @public annotations to methods of SolidExecutionContext, OutputContext, and InputContext

sryza:


python_modules/dagster/dagster/_core/execution/context/compute.py
```diff
@@ -233,6 +244,7 @@ def solid_handle(self) -> NodeHandle:
         return self._step_execution_context.solid_handle
 
     @property
+    @public
     def op_handle(self) -> NodeHandle:
         """NodeHandle: The current op's handle.
```
dpeng817, line 249: why do we want to make this public?

python_modules/dagster/dagster/_core/execution/context/compute.py
```diff
@@ -470,6 +499,7 @@ def log_materialization(context):
         else:
             check.failed("Unexpected event {event}".format(event=event))
 
+    @public
     def add_output_metadata(
         self,
         metadata: Mapping[str, Any],
```
dpeng817, line 505: I don't think we want to make this public on `OpExecutionContext`. In fact, I think we probably want to get rid of it, since you can use `Output` objects to log metadata.

dpeng817 (approved):
Generally good with this except for the few comments I left; I do wonder if we're being a bit too permissive here about what we include in the top level API for these objects; ie all these asset methods on `OpExecutionContext`. Do we expect all these methods to be necessary for users to invoke?

python_modules/dagster/dagster/_core/execution/context/compute.py
sryza: I think you are right

python_modules/dagster/dagster/_core/execution/context/compute.py
sryza: Hmm I think we do not. Thanks for catching

sryza:
@dpeng817 alas I do think we need to keep all the asset methods

dpeng817 (changes requested):
Coming back to this, I think the set of things exposed here is too broad - some of the things exposed here are not APIs exposed at the top-level.

python_modules/dagster/dagster/_core/execution/context/compute.py
```diff
@@ -249,6 +260,7 @@ def solid(self) -> Node:
         """
         return self._step_execution_context.pipeline_def.get_solid(self.solid_handle)
 
+    @public
     @property
     def op(self) -> Node:
```
dpeng817, line 265: not a public API - I think `node_name` is fine, but we should discuss if we think `Node` should be exported top level.

dpeng817 (approved):

PR #9047, opened 2022-07-26
[1.0] Remove legacy APIs from dagster_airflow

dpeng817:
Felt unnecessary to add a legacy import here, was far easier to just remove the top-level import.


sryza (approved):
Do we need to remove these from the api docs? Otherwise, looks good to me.

PR #9048, opened 2022-07-26
[1.0] remove legacy dagster_databricks imports

dpeng817:
as the title. `create_databricks_job_solid` is the only legacy import here.


sryza (approved):
Nice

PR #9049, opened 2022-07-26
[1.0] remove legacy APIs from top level of dagster-dbt

dpeng817:
as the title. Basically gets rid of all the top-level exported solids, nothing else.


OwenKephart (approved):
awesome -- imports w/in the tests will need to be updated as well, but this looks good otherwise.

PR #9050, opened 2022-07-26
[1.0] remove legacy APIs in dagster-gcp

dpeng817:
As the title. solids-related APIs mostly.


python_modules/libraries/dagster-gcp/dagster_gcp/bigquery/solids.py
```diff
@@ -1,4 +1,4 @@
 # pylint: disable=unused-import
 
 # Keep module for legacy backcompat
```
sryza, line 3: Remove this comment?

sryza (approved):

PR #9053, opened 2022-07-26
[1.0] remove legacy APIs snowflake/spark integrations

dpeng817:
combining so fewer reviews.


sryza (approved):
Should we delete `snowflake_solid_for_query`?

dpeng817:
yes, nice catch

PR #9055, opened 2022-07-27
remove legacy dbt docs example, fix graphql client example

dpeng817:
As the title. Removes dbt docs example that was only using legacy APIs, and fixes a graphql client example that referenced `PipelineRunStatus`.


OwenKephart (approved):
my only concern here is if something in the docs happened to reference this dbt example, but I guess if tests pass, then this should be fine.

PR #9063, opened 2022-07-27
Make op decorator usable as function

bcvanmeurs:
### Summary & Motivation

I was trying to use existing functions with the `op` decorator, in this way:

```python
my_op = op(
    my_fun, description="my_op", out={"a": Out(metadata={"x": 1}), "b": Out(metadata={"y": 2})}
)
```

However none of the arguments are passed into the instantiation of the `_Op` object, which I changed.

This is my first PR in this codebase, and 
 - I don't know if it is a deliberate choice to not do this
 - I don't fully understand the reasoning behind the original `check.invariant()` for everything to be `None`. Please help me understand if it is okay to remove these.

Ran all tests and added test case to use the decorator as a function.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/op_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/op_decorator.py
@@ -222,17 +222,20 @@ def op(
                 return 'cool', 4
     """
 
-    # This case is for when decorator is used bare, without arguments. e.g. @op versus @op()
     if callable(name):
-        check.invariant(input_defs is None)
-        check.invariant(output_defs is None)
-        check.invariant(description is None)
-        check.invariant(config_schema is None)
-        check.invariant(required_resource_keys is None)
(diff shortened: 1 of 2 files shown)


smackesey:
Hi @bcvanmeurs, thanks for the PR and your interest in Dagster!

As you've noticed, in the current design we don't support this-- in _any_ of our decorators. So, if we were to merge this PR, we'd have to change our decorator style in many other places to this style, which we're not going to do right now (we already have a planned change for upcoming 1.0 to make params keyword-only args).

But, there are two ways to accomplish what you want:

- You can construct an `OpDefinition` directly. This is the recommended pattern when you want to construct ops by some method other than a decorated function.
- Alternatively, `op(description=...)(my_fun)` will do it.

Thanks again for the PR, this is useful feedback that we're not clear enough in our docs on alternative methods of op construction.

bcvanmeurs:
Hi @smackesey, thanks for the feedback!

PR #9065, opened 2022-07-27
Add PUBLIC constant for NamedTuple attrs, add property/staticmethod support for @public

smackesey:
### Summary & Motivation

This adds a `PUBLIC` constant that we can use to mark whether `NamedTuple` attrs are public:

```
# Use `PUBLIC` with `typing_extensions.Annotated to annotate public attributes on `NamedTuple`:
#
# from typing_extensions import Annotated
# from dagster._annotations import PUBLIC
#
# class Foo(NamedTuple("_Foo", [("bar", Annotated[int, PUBLIC])])):
#     ...
```

See [thread](https://elementl-workspace.slack.com/archives/C03MMSAQKU7/p1658861914489869) for context.

Unit test.

### DIFF
--- a/python_modules/dagster/dagster/_annotations.py
+++ b/python_modules/dagster/dagster/_annotations.py
@@ -1,6 +1,6 @@
 import inspect
 from functools import wraps
-from typing import Callable, Type, TypeVar, cast
+from typing import Callable, Optional, Type, TypeVar, Union, cast
 
 from typing_extensions import Annotated, Final, TypeAlias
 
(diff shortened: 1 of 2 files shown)


smackesey:
* master
  * **PR #9065** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9065" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9065?utm_source=stack-comment).

smackesey:
@sryza I think that would be possible but I'm leery because it's an extra layer on top of plain Python's purpose-built abstraction for this (`Annotated`), with (IMO) minimal benefit, especially since the vast majority of our `NamedTuple` classes have one attr per line.

sryza:
A couple thoughts:
* Even if doesn't affect the # of lines, it's still annoying to type and read more characters. Not the end of the world though.
* If I'm understanding `Annotated` correctly, this is a bit of an abuse of what it's for.  I.e. `Annotated` is meant to encode metadata about the data type, but we're using it to encode metadata about the attribute.  Using the Annotated machinery as a hack for this behind the scenes seems fine, but I think encapsulating it would help make the mental model less confusing to readers of our NamedTuple classes.

Separately, throwing another idea out there that doesn't necessarily address these issues, just for the sake of exploring the space of options:
```
class Foo(NamedTuple("_Foo", [("bar", PublicAttr[int])])):
    ...
```

smackesey:
>If I'm understanding Annotated correctly, this is a bit of an abuse of what it's for. I.e. Annotated is meant to encode metadata about the data type, but we're using it to encode metadata about the attribute.

This is a very good point.

```
class Foo(NamedTuple("_Foo", [("bar", PublicAttr[int])])):
    ...
```

I like this a lot because I just discovered this pattern is [in the PEP](https://peps.python.org/pep-0593/#aliases-concerns-over-verbosity), which gives me confidence we're unlikey to have weird incompatibility with tooling. I'm sold.

sryza:
> I like this a lot because I just discovered this pattern is [in the PEP](https://peps.python.org/pep-0593/#aliases-concerns-over-verbosity),

Oh cool - let's do it.

sryza (approved):
Do we need to do any validation to make sure Sphinx will be able to access this?

smackesey:
>Do we need to do any validation to make sure Sphinx will be able to access this?

Sphinx literally imports your Python objects (it's running in a regular Python env) so it won't be a problem.

PR #9066, opened 2022-07-27
add @public decorator to AssetsDefinition and SourceAsset methods

sryza:

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/source_asset.py
+++ b/python_modules/dagster/dagster/_core/definitions/source_asset.py
@@ -1,6 +1,7 @@
 from typing import Dict, Iterator, Mapping, NamedTuple, Optional, Sequence, Union, cast
 
 import dagster._check as check
+from dagster._annotations import PublicAttr, public
 from dagster._core.definitions.events import AssetKey, CoercibleToAssetKey
 from dagster._core.definitions.metadata import (
     MetadataEntry,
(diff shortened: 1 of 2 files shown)


OwenKephart (approved):
seems like the correct set of things to me

sryza:
@OwenKephart re-requesting review because I added in SourceAsset

OwenKephart (approved):
was it a deliberate choice to keep SourceAsset.with_resources private? I'm fine with that, just double checking

dpeng817 (changes requested):
Gonna block this - I think that the set of things here is too broad. I don't think there is user utility for a lot of stuff like `can_subset` for example, and casting too wide a net on what properties/methods we directly communicate might make life difficult in the future. It also opens up a can of worms around APIs that aren't part of the public API but might be exposed through excessive property visibility. For example, if `AssetsDefinition.can_subset` is public, should `OpDefinition.compute_fn` be public (which returns a currently private class)? I think drawing that distinction is hard, and fwiw our docs don't have this level of detail right now. Look at JobDefinition's API docs https://docs.dagster.io/_apidocs/jobs#dagster.JobDefinition, which only expose `execute_in_process` and `with_hooks`.

sryza:
@dpeng817 we caught up on this in Slack, but repeating the high-level points here: I think that `can_subset` and the compute function are both part of the AssetsDefinition data model, and that data model should be visible to users.  I think `can_subset` is easy to expose, because the internal property is identical to the value that gets passed in, and thus we should make it public.  I think the compute function is more difficult to expose because of the way we wrap it, and, it's not extremely high priority, so we can punt on it for now.

python_modules/dagster/dagster/_core/definitions/source_asset.py
```diff
@@ -31,13 +32,13 @@ class SourceAsset(
     NamedTuple(
         "_SourceAsset",
         [
-            ("key", AssetKey),
+            ("key", PublicAttr[AssetKey]),
             ("metadata_entries", Sequence[Union[MetadataEntry, PartitionMetadataEntry]]),
```
dpeng817, line 36: arguably, shouldn't this be public for people directly utilizing the `SourceAsset` object in some way?

dpeng817 (approved):

PR #9069, opened 2022-07-28
fix minute/hour/day offsets in external asset partitions

sryza:
### Summary & Motivation

Without this change, trying to materialize a partition of the following asset from Dagit will fail:
```
from dagster import WeeklyPartitionsDefinition, asset

partitions_def = WeeklyPartitionsDefinition(start_date="2022-01-01", day_offset=2)


@asset(partitions_def=partitions_def)
def asset1():
    ...
```

This was reported by @geoHeil here: https://dagster.slack.com/archives/C01U954MEER/p1658410348160939.

Made sure the above code snippet passed.

### DIFF
--- a/python_modules/dagster/dagster/_core/host_representation/external_data.py
+++ b/python_modules/dagster/dagster/_core/host_representation/external_data.py
@@ -469,6 +469,9 @@ class ExternalTimeWindowPartitionsDefinitionData(
             ("timezone", Optional[str]),
             ("fmt", str),
             ("end_offset", int),
+            ("minute_offset", int),
+            ("hour_offset", int),
+            ("day_offset", Optional[int]),
         ],
     ),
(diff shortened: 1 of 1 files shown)


jamiedemaria (approved):
seems good to me

python_modules/dagster/dagster/_core/host_representation/external_data.py
```diff
@@ -479,6 +482,9 @@ def __new__(
         timezone: Optional[str],
         fmt: str,
         end_offset: int,
+        minute_offset: int,
+        hour_offset: int,
+        day_offset: Optional[int],
```
alangenfeld, line 487: adding non optional entries to existing `@whitelist_for_serdes` causes loading old entries to break

PR #9071, opened 2022-07-28
dagster project scaffold 1/n

yuhan:
### Summary & Motivation
The start of `dagster project` CLI stack:

```
$ dagster project --help
Usage: dagster project [OPTIONS] COMMAND [ARGS]...

  Commands for scaffolding new Dagster projects.

Options:
  -h, --help  Show this message and exit.

Commands:
  from-example         Create a new Dagster project using one of the official Dagster examples.
  scaffold             Create a new Dagster project using the default template
  scaffold-repository  Scaffold a new Dagster repository in an existing Dagster project
```
This PR only implements `dagster project scaffold` and `dagster project scaffold-repository`. 

end-to-end recording:
https://www.loom.com/share/a4ac71ddf78240dd89ce71acf5336673

```
# Activate pyenv 
pyenv install 3.8.6 # if you don't have pyenv installed
pyenv virtualenv 3.8.6 <my_pyenv_name>
pyenv activate <my_pyenv_name>
pip install --upgrade pip

# Run scaffold CLI
$ pip install dagster
$ dagster project scaffold --name my-project-1

Creating a Dagster project at my-project-1.
Creating a Dagster repository at my-project-1.
Generated files for Dagster repository in my-project-1.
Generated files for Dagster project in my-project-1.
Success! Created my-project-1 at /Users/yuhan/Desktop/test-dagster-project/my-project-1.
```

then, run:
```python
pip install -e .
dagit
```
opens dagit with an empty repo. 

newly defined asset shows up in the repo.


yuhan:
* master
  * **PR #9071** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9071" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9071?utm_source=stack-comment).

python_modules/dagster/dagster/_generate/templates/PROJECT_NAME_PLACEHOLDER/PROJECT_NAME_PLACEHOLDER/repository.py.tmpl
```diff
@@ -0,0 +1,12 @@
...
+from {{ repo_name }}.assets import cereals, nabisco_cereals
+
+asset_job = define_asset_job("asset_job")
+
+daily_cereals_schedule = ScheduleDefinition(job=asset_job, cron_schedule="0 0 * * *")
+
+
+@repository
+def {{ repo_name }}():
+    return [cereals, nabisco_cereals, daily_cereals_schedule]
```
sryza, line 12: We should use `load_assets_from_modules` or `load_assets_from_package_module` here so they don't need to add assets here every time they define one.

python_modules/dagster/dagster/_generate/templates/PROJECT_NAME_PLACEHOLDER/requirements.txt.tmpl
```diff
@@ -0,0 +1,3 @@
+-e .
+dagster=={{ dagster_version }}
```
sryza, line 2: What's the thinking behind doing this pin?

python_modules/dagster/dagster/_cli/__init__.py
```diff
@@ -29,6 +30,7 @@ def create_dagster_cli():
         "asset": asset_cli,
         "debug": debug_cli,
         "new-project": new_project_cli,
+        "project": project_cli
```
sryza, line 33: we should remove "new-project" soon after adding this, right?

sryza (commented):
Looking good!

Aside from the comments I left in the PR, my only high-level comment is that it strikes me as a weird that we have a `from-example` command but that the `scaffold` command also includes an example. My leaning would be to take the example out of the `scaffold` template and allow people to use `from-example` if they want an example.  That would make life easier for someone who understands Dagster already and wants to create multiple projects and doesn't want to have to clear out the example assets each time.

python_modules/dagster/dagster/_generate/templates/PROJECT_NAME_PLACEHOLDER/cloud_workspace.yaml.tmpl
```diff
@@ -0,0 +1,4 @@
+locations:
```
clairelin135, line 1: Do we need the `cloud_workspace.yaml` for the default project? Or should it just be for `scaffold-deployment` project?

python_modules/dagster/dagster/_generate/templates/PROJECT_NAME_PLACEHOLDER/requirements.txt.tmpl
```diff
@@ -0,0 +1,3 @@
+-e .
```
clairelin135, line 1: Do we need `requirements.txt` and `setup.py` for the default project? Not super clear on why someone would need both

python_modules/dagster/dagster/_cli/project.py
```diff
@@ -0,0 +1,80 @@
...
+        raise FileExistsError(
+            f"""
+            The directory {project_dir_abspath} already exists. Please delete the contents of this path
+            or choose another project location.
+            """
+        )
+
+    click.echo(f"Creating a new Dagster repository at {project_dir_abspath}...")
+    generate_project(name)
+    click.echo("Done.")
```
clairelin135, line 46: I think it would be nice here to include some starter commands (e.g. "You can run `dagit -w ....` to locally run your jobs")

clairelin135:
This is looking great to me! I agree with you on the defaults selected in this PR for cli naming, pinning dagster/dagit, and schedules

Had some questions about adding `requirements.txt` and `cloud_workspace.yaml` for the default project, but other than this is looking good

python_modules/dagster/dagster/_cli/__init__.py
yuhan: yes! i also plan to talk to 1.0 folks about deprecation vs removal

python_modules/dagster/dagster/_generate/templates/PROJECT_NAME_PLACEHOLDER/requirements.txt.tmpl
yuhan: for development, users will `pip install -e .` to use the latest version.
but when things are running in deployment, users will be using requirements.txt so if they'd *maybe* want to pin dagster, thats the place for them to pin.

but i think for scaffolding minimum, im leaning towards unpinning everything now.

python_modules/dagster/dagster/_generate/templates/PROJECT_NAME_PLACEHOLDER/cloud_workspace.yaml.tmpl
yuhan: was thinking that `dagster project scaffold` would be a wrapper of `dagster project scaffold-repository` (which creates the first code location / dagster repo) and `dagster project scaffold-deployment` (which scaffolds the deployment for cloud)


maybe - a better way around is to have dagster cloud cli that does the cloud deployment scaffold?

python_modules/dagster/dagster/_generate/templates/PROJECT_NAME_PLACEHOLDER/requirements.txt.tmpl
yuhan: requirements.txt will be used for deployment which is more tied to a concrete environment. here's a convo in slack [here](https://elementl-workspace.slack.com/archives/C03A0D72A6T/p1658515403719279)

jamiedemaria: maybe some in file comments about why both are included?

yuhan:
@sryza to clarify, are you saying `dagster project scaffold` should include an empty repository.py?

i think taking the example out of the scaffold template makes sense but my only reservation is when a new user opens dagit, they would see an empty repo which feels off and delays them seeing UI values. how about having a hello_world asset in repository.py (no assets.py, no schedule)?

sryza:
> @sryza to clarify, are you saying dagster project scaffold should include an empty repository.py?

I was imagining something like
- repository.py:

```
from dagster import load_assets_from_package_module, repository

from {{ repo_name }} import assets

@repository
def {{ repo_name }}():
    return [load_assets_from_package_module(assets)]
```

- and there would be an empty assets.py. 

> my only reservation is when a new user opens dagit, they would see an empty repo which feels off and delays them seeing UI values. how about having a hello_world asset in repository.py (no assets.py, no schedule)?

If a user wants to see stuff in Dagit before writing any code, `dagster project from-example` makes this very easy, right?

yuhan:
makes sense!

python_modules/dagster/dagster/_cli/project.py
yuhan: im including the next-step-ish explainer in README for now

yuhan:
The only thing im unsure about this dir structure is the ability to expand to multiple repositories. 

in terms of dagster repo, here's what this pr has now:
```
.
├── cloud_workspace.yaml
├── my_dagster_project_2
│   ├── __init__.py
│   ├── assets
│   │   └── __init__.py
│   └── repository.py
├── my_dagster_project_2_tests
│   ├── __init__.py
│   └── test_assets.py
├── setup.py
└── workspace.yaml
```

i feel for multiple repos, we may want to recommend this, where repo_1 and repo_2 are two submodules inside my_dagster_project:
```
.
├── README.md
├── cloud_workspace.yaml
├── my_dagster_project_multiple_locations
│   ├── __init__.py
│   ├── repo_1
│   │   ├── __init__.py
│   │   ├── assets
│   │   │   └── __init__.py
│   │   └── repository.py
│   └── repo_2
│       ├── __init__.py
│       ├── assets
│       │   └── __init__.py
│       └── repository.py
├── my_dagster_project_multiple_locations_tests
│   ├── __init__.py
│   ├── repo_1
│   │   ├── __init__.py
│   │   └── test_assets.py
│   └── repo_2
│       ├── __init__.py
│       └── test_assets.py
├── pyproject.toml
├── requirements.txt
├── setup.cfg
├── setup.py
└── workspace.yaml
```

so the minimal scaffold case might need to be, which adds an additional unnecessary "my_repo/" layer for the single-repo case (which could be majority):
```
.
├── README.md
├── cloud_workspace.yaml
├── my_dagster_project_multiple_locations
│   ├── __init__.py
│   ├── my_repo # HERE
│   │   ├── __init__.py
│   │   ├── assets
│   │   │   └── __init__.py
│   │   └── repository.py
├── my_dagster_project_multiple_locations_tests
│   ├── __init__.py
│   ├── my_repo # HERE
│   │   ├── __init__.py
│   │   └── test_assets.py
├── pyproject.toml
├── requirements.txt
├── setup.cfg
├── setup.py
└── workspace.yaml
```

@sryza @clairelin135 @jamiedemaria thoughts?

python_modules/dagster/dagster/_cli/project.py
```diff
@@ -0,0 +1,80 @@
...
+@click.group(name="project")
+def project_cli():
+    """
+    Commands for scaffolding new Dagster projects.
+    """
+
+
+@project_cli.command(
+    name="scaffold-repository",
+    help="Create a new Dagster repository using the default template",
```
sryza, line 18: I think it would be more accurate to say something like "create a folder structure with a single Dagster repository, in the current directory".

python_modules/dagster/dagster/_cli/project.py
```diff
@@ -0,0 +1,80 @@
...
+            """
+        )
+
+    generate_repository(name)
+    click.echo(_styled_success_statement(name, dir_abspath))
+
+
+@project_cli.command(
+    name="scaffold",
+    help="Create a new Dagster project using the default template",
```
sryza, line 46: Same as above, I think we should be crystal clear that this is creating files and folders. Otherwise people might think it's some sort of action that mutates the Dagster instance or something.

python_modules/dagster/dagster/_generate/templates/PROJECT_NAME_PLACEHOLDER/requirements.txt.tmpl
```diff
@@ -0,0 +1,3 @@
+-e {{ repo_name }}
+# It's recommended to include environment-specific requirements in this file.
+# For example, pinning a package's version for deployment.
```
sryza, line 3: Nitpick: newline at end of file

python_modules/dagster/dagster/_generate/templates/PROJECT_NAME_PLACEHOLDER/requirements.txt.tmpl
```diff
@@ -0,0 +1,3 @@
+-e {{ repo_name }}
```
sryza, line 1: I haven't thought about this deeply, but if we're including setup.py at the repository level, I would expect requirements.txt to be at that level as well?

python_modules/dagster/dagster/_generate/generate.py
```diff
@@ -3,36 +3,86 @@
 
 import jinja2
 
+from dagster._utils import file_relative_path
 from dagster.version import __version__ as dagster_version
 
-NEW_PROJECT_PLACEHOLDER = "new_project"
-NEW_PROJECT_PATH = os.path.join(os.path.dirname(__file__), NEW_PROJECT_PLACEHOLDER)
+
+def generate_repository(path: str):
```
sryza, line 10: Not a blocker or anything, but it would be cool for this to add an entry to the workspace.yaml

sryza (commented):
Looking at the text of these commands, it's a little difficult for me to know what to expect the difference is between `scaffold` and `scaffold-repository`.  I think it's important for the help text to disambiguate.  Maybe `scaffold-repository` should say something like "create a project for a repository within a folder structure that includes multiple repositories" or something like that.

Other than that, I think this looks great.

python_modules/dagster/dagster/_generate/templates/PROJECT_NAME_PLACEHOLDER/requirements.txt.tmpl
yuhan: for the minimal case, they are at the same level, and it doesn't matter.

the reason why requirements.txt is in project root while setup.py is at the repo level is for the multi repo case (example: https://github.com/yuhan/my-dagster-project-multiple-locations), we'll have the requirements.txt at the root which installs all the dagster repos / code locations as separate python packages: 
```
.
├── README.md
├── cloud_workspace.yaml
├── my_repo_1
│   ├── my_repo_1
│   │   ├── __init__.py
│   │   ├── assets
│   │   │   └── __init__.py
│   │   └── repository.py
│   ...
│   ├── pyproject.toml
│   ├── setup.cfg
│   └── setup.py
├── my_repo_2
│   ├── my_repo_2
│   │   ├── __init__.py
│   │   ├── assets
│   │   │   └── __init__.py
│   │   └── repository.py
│   ...
│   ├── pyproject.toml
│   ├── setup.cfg
│   └── setup.py
├── requirements.txt
```

Currently, we want the requirements.txt in the git repo root because that's how the cloud cli assumes. 
I don't think we have a final decision yet. We're still discussing that with cloud:
https://elementl-workspace.slack.com/archives/C03MCLC0L83/p1659142575405289?thread_ts=1659137503.267199&cid=C03MCLC0L83

python_modules/dagster/dagster/_generate/generate.py
yuhan: yup and add an entry to the requirements.txt in the project root as well.
in a later diff, i plan to a cli prompt to ask for paths to workspace.yaml and requirements.txt - it could be tricky to assume the file path or name.

python_modules/dagster/dagster/_generate/templates/PROJECT_NAME_PLACEHOLDER/cloud_workspace.yaml.tmpl
yuhan: ok, i decided to take the cloud yaml out of the default project template. and i opened a thread to discuss the deployment cli.

python_modules/dagster/dagster/_cli/project.py
yuhan: image

python_modules/dagster/dagster/_generate/templates/PROJECT_NAME_PLACEHOLDER/requirements.txt.tmpl
sryza: in the multi-repo case, each entry in the requirements.txt would be one of the code locations?

yuhan: yes, it'd be
```
my_repo_1
my_repo_2
# and other reqs for deployment
```

sryza: In many cases, they'll have conflicting dependencies, so they can't be installed together into the same python environment. But the point of a requirements.txt is a set of libraries that go into the same python environment, right?

yuhan: ah that's right - confirmed w/ cloud that each code location and have its own requirements.txt for serverless. 
moved requirements.txt to repo level

sryza (approved):
Awesome!

Issue #9014, opened 2022-07-24
[K8S] The example k8s-example-user-code template not ready because exceeded readiness probe checking


leehuwuj:
### Dagster version

0.15.6

### What's the issue?

The example template for user deployment: k8s-example-user-code always not ready because exceeded deadline in readiness probe checking.
```
Events:
  Type     Reason     Age               From               Message
  ----     ------     ----              ----               -------
  Normal   Scheduled  31s               default-scheduler  Successfully assigned dagster/dagster-dagster-user-deployments-k8s-example-user-code-1-5blhx2 to docker-desktop
  Normal   Pulling    30s               kubelet            Pulling image "docker.io/dagster/user-code-example:0.15.6"
  Normal   Pulled     27s               kubelet            Successfully pulled image "docker.io/dagster/user-code-example:0.15.6" in 3.637367626s
  Normal   Created    26s               kubelet            Created container dagster-user-deployments
  Normal   Started    26s               kubelet            Started container dagster-user-deployments
  Warning  Unhealthy  8s (x2 over 17s)  kubelet            Readiness probe errored: rpc error: code = Unknown desc = deadline exceeded ("DeadlineExceeded"): context deadline exceeded
```

### What did you expect to happen?

I think the readinessProbe default config in deployment-user.yaml is set too strict:
```
readinessProbe:
  exec:
    command: ["dagster", "api", "grpc-health-check", "-p", "3030"]
  periodSeconds:
    20
  timeoutSeconds:
    3
  successThreshold:
    1
  failureThreshold:
    3
```

When i run command manually in the pod, it's always return after 5-8 seconds with my local machine (Macbook Pro M1 Pro) so i think the timeoutSeconds was set too strict (only 3 seconds) which always returns failed.

I think this issue will make new users confuse when try Dagster k8s deployment for the fist time and it would be better if we increase the `timeoutSeconds` to `10` seconds.

### How to reproduce?

I render the k8s template from Helm chart by this command, my values.yaml almost the same with the default one!
```
helm template dagster ./dagster --version 0.15.6 --values ./dagster/values.yaml --output-dir ./rendered
```

### Deployment type

Dagster Helm chart

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

johannkm:
Sounds reasonable to me, there's not too high a cost of having a higher timeout I think

wannesghielens:
[...]

PR #9073, opened 2022-07-28
#9014 Increase user-deployments readinessProbe timeout

leehuwuj:
### Summary & Motivation
As discussion at https://github.com/dagster-io/dagster/issues/9014 .
Increase the `timeoutSeconds` of readinessProbe for user-deployment example to make sure the deployment will be checked by the right way.


smackesey:
cc @johannkm

emgeee:
Just to jump in -- I was having the issue described in https://github.com/dagster-io/dagster/issues/9014 and this timeout increase fixed it!

catrider:
ditto - ran into this with the helm deployment tutorial and the timeout increase fixed it

johannkm (approved):

PR #9077, opened 2022-07-28
remove deprecated arguments on @asset, @multi_asset, and AssetIn

sryza:

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
@@ -6,7 +6,6 @@ from typing import (
     Dict,
     Mapping,
     Optional,
-    Sequence,
     Set,
     Tuple,
     Union,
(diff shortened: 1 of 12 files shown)


python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -107,16 +103,11 @@ def asset(
...
-            the asset resides in.  The concatenation of namespace and name forms the asset key. Each
-            item in namespace must be a valid name in dagster (ie only contains letters, numbers,
-            and _) and may not contain python reserved keywords.
         key_prefix (Optional[Union[str, Sequence[str]]]): If provided, the asset's key is the
             concatenation of the key_prefix and the asset's name, which defaults to the name of
             the decorated function. Each item in key_prefix must be a valid name in dagster (ie only
             contains letters, numbers, and _) and may not contain python reserved keywords.
-        ins (Optional[Mapping[str, AssetIn]]): A dictionary that maps input names to their metadata
-            and namespaces.
+        ins (Optional[Mapping[str, AssetIn]]): A dictionary that maps input names to their metadata.
```
OwenKephart, line 110: a nitpick, because this wasn't introduced in this commit, but  "A dictionary that maps input names to their metadata" is not exactly accurate. If we steal the description of the op's "ins" parameter, that would look like "Information about the inputs to the asset. " (not very descriptive, but at least it's symmetrical with the op name)

feel free to ignore

OwenKephart (approved):
awesome

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
sryza: 👍

PR #9082, opened 2022-07-28
respect repository namespace in schedule & sensor idempotence checks

alangenfeld:
if a user has multiple copies of the same schedule running in different repos at the same time, a run in one will prevent the other from firing. 

added tests that previously failed

### DIFF
--- a/python_modules/dagster/dagster/_daemon/sensor.py
+++ b/python_modules/dagster/dagster/_daemon/sensor.py
@@ -5,17 +5,18 @@ import threading
 import time
 from collections import defaultdict
 from contextlib import ExitStack
-from typing import Dict, NamedTuple, Optional
+from typing import Dict, NamedTuple, Optional, Sequence
 
 import pendulum
 
 import dagster._check as check
 import dagster._seven as seven
-from dagster._core.definitions.run_request import InstigatorType
+from dagster._core.definitions.run_request import InstigatorType, RunRequest
 from dagster._core.definitions.sensor_definition import DefaultSensorStatus, SensorExecutionData
 from dagster._core.definitions.utils import validate_tags
 from dagster._core.errors import DagsterError
 from dagster._core.host_representation import PipelineSelector
+from dagster._core.host_representation.external import ExternalSensor
 from dagster._core.instance import DagsterInstance
 from dagster._core.scheduler.instigation import (
     InstigatorState,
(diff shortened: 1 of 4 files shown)


alangenfeld:
* master
  * **PR #9082** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9082" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9082?utm_source=stack-comment).

python_modules/dagster/dagster/_scheduler/scheduler.py
```diff
@@ -519,11 +523,17 @@ def _schedule_runs_at_time(
...
+    instance,
+    external_schedule: ExternalSchedule,
+    schedule_time,
+    run_request,
+):
     tags = merge_dicts(
         PipelineRun.tags_for_schedule(external_schedule),
         {
             SCHEDULED_EXECUTION_TIME_TAG: to_timezone(schedule_time, "UTC").isoformat(),
+            REPOSITORY_LABEL_TAG: external_schedule.get_external_origin().external_repository_origin.get_label(),
```
prha, line 536: Ah, I more meant that we should put this tag inside of the `PipelineRun.tags_for_schedule` call, but this works if we want to limit the scope and change all the other repository-scoped run fetching calls in one fell swoop.

I imagine we'd need to apply the identical fix for sensors, also.

alangenfeld (commented):
updated to include sensors - i saw https://github.com/dagster-io/dagster/pull/8329 but couldn't discern why mutiple tags were not submitted so went that direction

prha:
I'd feel a whole lot better about this change (especially with the run key tag change) if we added some tests.

prha:
Specifically, using list values for tags in combination with non-list values for tags.

gibsondan (commented):
While we are messing with this function (or separately) could we also do the optimization here? https://elementl-workspace.slack.com/archives/C03MCLC0L83/p1659215257740189?thread_ts=1659193150.860529&channel=C03MCLC0L83&message_ts=1659215257.740189

Seems to make a big difference

python_modules/dagster/dagster/_scheduler/scheduler.py
prha: Ignore my comment on `tags_for_schedule`... there's some complication there in terms of extracting the repository label.

python_modules/dagster/dagster/_daemon/sensor.py
```diff
@@ -659,20 +660,28 @@ def _is_under_min_interval(state, external_sensor):
...
 
-    runs_with_run_keys = instance.get_runs(filters=RunsFilter(tags={RUN_KEY_TAG: run_keys}))
-
-    valid_runs = [
-        run for run in runs_with_run_keys if run.tags.get(SENSOR_NAME_TAG) == external_sensor.name
-    ]
+    runs_with_run_keys = instance.get_runs(
+        filters=RunsFilter(
+            tags={
+                RUN_KEY_TAG: run_keys,
```
gibsondan, line 676: Worried about perf implications of adding additional tags as per data from the thread I posted earlier

also worried that this could filter out old runs from before the repo label tag existed, and suddenly cause a bunch of new runs to be created that has run keys already 


To fix both worries, could we do this filtering in memory instead using the external_pipeline_origin field on the candidate run? That has the nice property of still always being set for runs from before the repo label tag was created. Would still need to check the sensor name tag most likely (but doing that in memory as well perhaps?)

python_modules/dagster/dagster/_daemon/sensor.py
```diff
@@ -659,16 +660,29 @@ def _is_under_min_interval(state, external_sensor):
...
+    # filter down to runs with run_key that match the sensor name and its namespace (repository)
     valid_runs = [
-        run for run in runs_with_run_keys if run.tags.get(SENSOR_NAME_TAG) == external_sensor.name
+        run
+        for run in runs_with_run_keys
+        if run.tags.get(SENSOR_NAME_TAG) == external_sensor.name
+        and (
+            external_sensor.get_external_origin().external_repository_origin.get_label()
+            == run.external_pipeline_origin.external_repository_origin.get_label()
+        )
```
alangenfeld, line 685: i went to investigate the linked threads about DB perf and observed different query plans executed. For whatever reason the query planner got smarter.

Despite that, I preserved the in memory filtering to reduce deviation of this change.

python_modules/dagster/dagster/_daemon/sensor.py
```diff
@@ -659,16 +660,29 @@ def _is_under_min_interval(state, external_sensor):
...
 
+    # filter down to runs with run_key that match the sensor name and its namespace (repository)
     valid_runs = [
-        run for run in runs_with_run_keys if run.tags.get(SENSOR_NAME_TAG) == external_sensor.name
+        run
+        for run in runs_with_run_keys
+        if run.tags.get(SENSOR_NAME_TAG) == external_sensor.name
+        and (
+            external_sensor.get_external_origin().external_repository_origin.get_label()
+            == run.external_pipeline_origin.external_repository_origin.get_label()
```
gibsondan, line 684: get_id() over get_label() ? I could imagine get_label() truncating in the future or somethign for example, get_id() will always be unique

gibsondan (approved):

python_modules/dagster/dagster/_daemon/sensor.py
alangenfeld: do you mean `selector_id` ? we don't want code loading details which effect the origin id to get in the mix here do we?

gibsondan: yeah, get_selector_id, my mistake

PR #9090, opened 2022-07-28
Fix docstrings for context objects

dpeng817:
Adds examples, and cautions to not instantiate directly.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/sensor_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/sensor_definition.py
@@ -49,10 +49,11 @@ DEFAULT_SENSOR_DAEMON_INTERVAL = 30
 
 
 class SensorEvaluationContext:
-    """Sensor execution context.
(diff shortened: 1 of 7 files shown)


python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -49,10 +49,9 @@ class DefaultSensorStatus(Enum):
 
 
 class SensorEvaluationContext:
-    """Sensor execution context.
+    """The context object available as the argument to the evaluation function of a :py:class:`dagster.SensorDefinition`.
 
-    An instance of this class is made available as the first argument to the evaluation function
-    on SensorDefinition.
+    Users should not instantiate this object directly. To construct a `SensorEvaluationContext` for testing purposes, use :py:func:`dagster.build_sensor_context`.
```
sryza, line 54: Keep length under 100 chars

sryza (approved):
Nice

PR #9091, opened 2022-07-28
final round of API moves to legacy

dpeng817:
As the title.


sryza (commented):
image


I wonder if we should make `dagster_type_materializer` experimental instead, given that we never called it out as part of our legacy APIs? Otherwise, LGTM.

dpeng817:
> image
> 
> I wonder if we should make `dagster_type_materializer` experimental instead, given that we never called it out as part of our legacy APIs? Otherwise, LGTM.

fine with that lol

sryza (approved):

PR #9094, opened 2022-07-29
in fs_io_manager avoid storing and loading for None/Nothing

sryza:


python_modules/dagster/dagster/_core/storage/fs_io_manager.py
```diff
@@ -131,6 +131,9 @@ def handle_output(self, context, obj):
 
         filepath = self._get_path(context)
 
+        if context.dagster_type.is_nothing or context.dagster_type.typing_type is None:
+            return
```
alangenfeld, line 135: invariant/assert `obj` is None?

python_modules/dagster/dagster/_core/definitions/output.py
```diff
@@ -272,7 +273,13 @@ def combine_with_inferred(self: TOut, inferred: InferredOutputProps) -> TOut:
 
 def _checked_inferred_type(inferred: Any) -> DagsterType:
     try:
-        return resolve_dagster_type(inferred)
+        if inferred == inspect.Parameter.empty:
+            return resolve_dagster_type(None)
+        elif inferred is None:
+            return resolve_dagster_type(type(None))
```
alangenfeld, line 279: worth adding comments - very subtle

python_modules/dagster/dagster_tests/core_tests/storage_tests/test_fs_io_manager.py
```diff
@@ -360,3 +365,57 @@ def four(inp):
...
+        assert not os.path.exists(os.path.join(tmpdir_path, "asset2"))
+        handled_output_events = list(
+            filter(lambda evt: evt.is_handled_output, result.all_node_events)
+        )
+        assert len(handled_output_events) == 2
+
+        for event in handled_output_events:
+            assert len(event.event_specific_data.metadata_entries) == 0
+
+
```
alangenfeld, line 396: should we add a test to protect / demonstrate the behavior around being able to handle `None` in a custom io manager?

alangenfeld (commented):
I imagined us handling this at the "dont call handle_output layer" but I'm cool with this approach too. 

If we do it this way - we should update all the cloud bucket io managers too right? 

To clarify, the current behavior is that `Nothing` causes us to bypass `handle_output` though right? Do we want to change that?

sryza:
> To clarify, the current behavior is that Nothing causes us to bypass handle_output though right? Do we want to change that?

@alangenfeld I was surprised the discover that is actually not the current behavior.  I.e. this test fails:

```
def test_fs_io_manager_ops_none():
    from dagster import Nothing, In, job

    with tempfile.TemporaryDirectory() as tmpdir_path:
        io_manager_def = fs_io_manager.configured({"base_dir": tmpdir_path})

        @op(out=Out(Nothing))
        def op1():
            pass

        @op(ins={"abc": In(Nothing)}, out=Out(Nothing))
        def op2():
            pass

        @job(resource_defs={"io_manager": io_manager_def})
        def job1():
            op2(op1())

        result = job1.execute_in_process()

        handled_output_events = list(
            filter(lambda evt: evt.is_handled_output, result.all_node_events)
        )
        assert len(handled_output_events) == 2

        for event in handled_output_events:
            assert len(event.event_specific_data.metadata_entries) == 0
```

python_modules/dagster/dagster_tests/core_tests/storage_tests/test_fs_io_manager.py
dpeng817: agreed - there should be a test in the generalized io manager tests that documents the behavior of `Nothing` with `handle_output`

sryza: added this

python_modules/dagster/dagster_tests/core_tests/storage_tests/test_fs_io_manager.py
```diff
@@ -361,3 +366,57 @@ def four(inp):
...
+        result = job1.execute_in_process()
+
+        handled_output_events = list(
+            filter(lambda evt: evt.is_handled_output, result.all_node_events)
+        )
+        assert len(handled_output_events) == 2
+
+        for event in handled_output_events:
+            assert len(event.event_specific_data.metadata_entries) == 0
```
dpeng817, line 422: I know that the implementation bypasses this, but for posterity's sake it might be good to have a test that optional types still get handled properly vs None/Nothing types.

python_modules/dagster/dagster_tests/core_tests/storage_tests/test_io_manager.py
```diff
@@ -919,3 +920,65 @@ def the_graph():
...
+            self.handle_output_calls += 1
+
+    my_io_manager = MyIOManager()
+
+    @op(out=Out(Nothing))
+    def op1():
+        ...
+
+    @op
+    def op2(_input1):
```
dpeng817, line 974: nice

dpeng817 (approved):
One comment regarding tests for your consideration, otherwise this looks like the right approach. Should copy this behavior over for s3_pickle, adls2_pickle, and gcs_pickle (and the snowflake one idr it's name)

sryza:
@dpeng817 I just pushed up a change that adds the gcs, s3, and adls2 io managers. I think we can wait on the snowflake one.

PR #9013, opened 2022-07-23
Typing improvements

smackesey:
### Summary & Motivation

More type annotations.

### How I Tested These Changes

BK

PR #8733, opened 2022-07-04
more types

smackesey:
Assorted type annotations

PR #9095, opened 2022-07-29
sequence and mapping mass convert

smackesey:
### Summary & Motivation

This is a large PR that mass converts List/Dict to Sequence/Mapping throughout Dagster core. This was done with a supervised automated tool that let me rapidly cycle through every instance of List/Dict in a function param, return type, and NamedTuple spec, allowing me to check them off or skip them one by one. There was then a process of refinement and reversing some of the changes until mypy/pylint passed.

The rule used here was: if the value is not intended to be mutated, it is Sequence/Mapping rather than List/Dict. This applied to the vast majority of cases identified by the tool.

This should reduce annoying cascanding mypy errors in the future when adding new type annotations, as sometimes adding a new function with a Sequnece annotation would cause errors due to calling older APIs that (unnecessarily) expected Lists.

There are also a small number of non-sequence/mapping-specific changes that needed to be included due to mypy errors triggered as a result of the sequence/mapping changes. These are concentrated in the second, much smaller commit (reviewers look here).

Existing BK suite.


smackesey:
* master
  * **PR #9095** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9095" target="_blank">Graphite</a>  👈
    * **PR #8733** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/8733" target="_blank">Graphite</a> 
      * **PR #9013** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9013" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9095?utm_source=stack-comment).

dpeng817 (approved):

PR #9096, opened 2022-07-29
Be more defensive in PythonErrorInfo when rendering causes

gibsondan:
Summary:
Got an error in dagit that maps to this not always being set to an empty list. Be more defensive

Test Plan:
Push this change to dogfood and hit redeploy on the workspace tab, get a clearer error


hellendag:
https://github.com/dagster-io/dagster/pull/9103

gibsondan:
deferring to dish's superior PR

PR #9097, opened 2022-07-29
Add missing `job_name` argument for creating run request for partioned jobs

metinsenturk:
### Summary & Motivation

Creating a `RunRequest` for a partitioned job is missing `job_name` argument. This causes errors when using a sensor that triggers multiple jobs.

When a sensor defined to trigger multiple jobs, it asks for each RunRequest to have a `job_name` argument, therefore we need to have `job_name`, like below.

``` py
@sensor(jobs=[job1_with_no_partition, job2_with_partition], minimum_interval_seconds=60 * 20)
def my_cool_sensor(context: SensorEvaluationContext):
    """
    Checks a process and starts multiple jobs.
    """
    yield RunRequest(
        run_key=None, run_config={...}, job_name='job1_with_no_partition')
    yield job2_with_partition.run_request_for_partition(
        partition_key="my_partition_key", run_key="my_partition_key", job_name='job2_with_partition')

    context.update_cursor(...)

```

I can add if needed.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/job_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/job_definition.py
@@ -550,6 +550,7 @@ class JobDefinition(PipelineDefinition):
         partition_key: str,
         run_key: Optional[str],
         tags: Optional[Mapping[str, str]] = None,
+        job_name: Optional[str] = None,        
     ) -> RunRequest:
         partition_set = self.get_partition_set_def()
         if not partition_set:
@@ -563,7 +564,7 @@ class JobDefinition(PipelineDefinition):
             else partition_set.tags_for_partition(partition)
         )
 
-        return RunRequest(run_key=run_key, run_config=run_config, tags=run_request_tags)
+        return RunRequest(run_key=run_key, run_config=run_config, tags=run_request_tags, job_name=job_name)
 
     def with_hooks(self, hook_defs: AbstractSet[HookDefinition]) -> "JobDefinition":
         """Apply a set of hooks to all op instances within the job."""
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -550,6 +550,7 @@ def run_request_for_partition(
         partition_key: str,
         run_key: Optional[str],
         tags: Optional[Mapping[str, str]] = None,
+        job_name: Optional[str] = None,
```
dpeng817, line 553: I think we don't want to make this passable like so - I think we would just want to pass the current job's name to the run request.

python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -563,7 +564,7 @@ def run_request_for_partition(
             else partition_set.tags_for_partition(partition)
         )
 
-        return RunRequest(run_key=run_key, run_config=run_config, tags=run_request_tags)
+        return RunRequest(run_key=run_key, run_config=run_config, tags=run_request_tags, job_name=job_name)
```
dpeng817, line 567:
```suggestion
        return RunRequest(run_key=run_key, run_config=run_config, tags=run_request_tags, job_name=self.name)
```

python_modules/dagster/dagster/_core/definitions/job_definition.py
dpeng817, line 553:
```suggestion
```

dpeng817 (changes requested):
A few comments - but great catch.

metinsenturk:
> A few comments - but great catch.

Oh i didn't see the `self.job_name`. Great that it helps!

dpeng817 (approved):

PR #9098, opened 2022-07-29
add public decorator to executor def methods

dpeng817:
Didn't include config_schema bc return type is `IDefinitionConfigSchema`, which is not in the public API (or usable for that matter). Only other one I would consider is `get_requirements`.


sryza (approved):
nice

PR #9103, opened 2022-07-29
[dagit] Fix generic error in PythonErrorInfo

hellendag:
### Summary & Motivation

`PythonErrorInfo` is able to render generic `{message: string}` errors that include no `stack` and no `causes` values.

Unfortunately, we are currently casting the object to `PythonErrorFragment` and referencing `stack` and `causes`, which leads to runtime errors when trying to map over the `causes` array.

Eliminate the cast and instead define the generic error as an object that may optionally have `stack` and `causes`, even though it never will. This allows us to reliably reference these keys and set them as empty arrays if they are not supplied, thus avoiding the runtime error of calling `map` on an undefined value.

I included a test to verify that the generic error type renders correctly.

`yarn jest PythonErrorInfo`

### DIFF
--- a/js_modules/dagit/packages/core/src/app/PythonErrorInfo.tsx
+++ b/js_modules/dagit/packages/core/src/app/PythonErrorInfo.tsx
@@ -7,20 +7,27 @@ import {MetadataEntries} from '../metadata/MetadataEntry';
 import {MetadataEntryFragment} from '../metadata/types/MetadataEntryFragment';
 import {ErrorSource} from '../types/globalTypes';
 
-import {PythonErrorFragment} from './types/PythonErrorFragment';
+import {
+  PythonErrorFragment,
+  PythonErrorFragment_causes as Cause,
+} from './types/PythonErrorFragment';
+
+export type GenericError = {
+  message: string;
+  stack?: string[];
+  causes?: Cause[];
+};
 
 interface IPythonErrorInfoProps {
   showReload?: boolean;
   centered?: boolean;
-  error: {message: string} | PythonErrorFragment;
+  error: GenericError | PythonErrorFragment;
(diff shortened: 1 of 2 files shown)


hellendag:
* master
  * **PR #9103** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9103" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9103?utm_source=stack-comment).

js_modules/dagit/packages/core/src/app/PythonErrorInfo.tsx
```diff
@@ -7,20 +7,27 @@ import {MetadataEntries} from '../metadata/MetadataEntry';
...
+  error: GenericError | PythonErrorFragment;
   failureMetadata?: {metadataEntries: MetadataEntryFragment[]} | null;
   errorSource?: ErrorSource | null;
 }
 
 export const PythonErrorInfo: React.FC<IPythonErrorInfoProps> = (props) => {
-  const message = props.error.message;
-  const stack = (props.error as PythonErrorFragment).stack;
-  const causes = (props.error as PythonErrorFragment).causes;
+  const {message, stack = [], causes = []} = props.error;
```
gibsondan, line 30: this is ultimately the only non-typing change right? None will be cast to empty list now

gibsondan (approved):

js_modules/dagit/packages/core/src/app/PythonErrorInfo.tsx
hellendag: No casting anywhere now. `causes` can now be an array or `undefined` (because of the `GenericError | PythonErrorFragment` union, so in this case we're just setting the undefined value (a `GenericError`) to an empty array so that `map` is always called on an array.

PR #9106, opened 2022-07-29
[docs] - Reformat category pages

erinkcochran87:
This PR reformats the following category pages in the docs:

- [Guides](https://dagster-git-erin-category-pages-elementl.vercel.app/guides)
- [Concepts](https://dagster-git-erin-category-pages-elementl.vercel.app/concepts)
- [Integrations](https://dagster-git-erin-category-pages-elementl.vercel.app/integrations)

### DIFF
--- a/docs/content/concepts.mdx
+++ b/docs/content/concepts.mdx
@@ -1,89 +1,241 @@
+---
+title: Concepts | Dagster Docs
+---
+
 # Concepts
 
-Dagster is a data orchestrator. It lets you define jobs in terms of the data flow between logical components called ops. These jobs can be developed locally and run anywhere.
+TODO
 
-## How to use the Concepts section
+---
 
-Each page in this section contains:
+## Software-defined assets
 
-- **Overview**: Description of the concept.
-- **Relevant APIs**: Index of top-level Dagster APIs that are relevant to the concept.
-- **Examples**: Easy-to-copy code snippets for the concept.
-- **Patterns**: A list of advanced patterns to use and anti-patterns to avoid with the concept.
+An asset is an object in persistent storage, such as a table, file, or persisted machine learning model. A software-defined asset is a Dagster object that couples an asset to the function and upstream assets that are used to produce its contents.
 
-## Sections
+<ArticleList>
+  <ArticleListItem
+    title="Software-defined assets"
+    href="/concepts/assets/software-defined-assets"
+  ></ArticleListItem>
+  <ArticleListItem
+    title="Asset materializations"
+    href="/concepts/assets/sasset-materializations"
+  ></ArticleListItem>
+  <ArticleListItem
+    title="Asset observations"
+    href="/concepts/assets/asset-observations"
+  ></ArticleListItem>
+  <ArticleListItem
+    title="Multi-assets"
+    href="/concepts/assets/multi-assets"
+  ></ArticleListItem>
+</ArticleList>
 
-<LinkGrid>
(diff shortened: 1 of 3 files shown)


docs/content/concepts.mdx
```diff
@@ -1,89 +1,241 @@
+---
+title: Concepts | Dagster Docs
+---
+
 # Concepts
 
-Dagster is a data orchestrator. It lets you define jobs in terms of the data flow between logical components called ops. These jobs can be developed locally and run anywhere.
+TODO
```
erinkcochran87, line 7: TODO

docs/content/concepts.mdx
```diff
@@ -1,89 +1,241 @@
+---
```
erinkcochran87, line 1: The goal of this page is to act as a jumping-off point, so the copy should be kept high-level and brief.

sryza, line 1: Huge improvement

docs/content/guides.mdx
```diff
@@ -1,16 +1,49 @@
...
+  <ArticleListItem
+    title="Enriching with Software-defined Assets"
+    href="/guides/dagster/enriching-with-software-defined-assets"
+  ></ArticleListItem>
+  <ArticleListItem
+    title="Using the Hacker News example project"
+    href="/guides/dagster/example_project"
+  ></ArticleListItem>
+  <ArticleListItem
+    title="Re-excuting Dagster jobs"
```
sryza, line 19: Re-executing

docs/content/guides.mdx
```diff
@@ -1,16 +1,49 @@
+---
+title: Guides | Dagster Docs
+---
+
 # Guides
 
-## Dagster Guides
+Learn to accomplish common tasks in Dagster.
```
sryza, line 7: I think losing the descriptions on this page might sacrifice more context than is worth it for the streamlining - each guide is a bird of such a different feather that seeing them in a list without additional context makes it difficult to get a handle on what's going on.

E.g. it's hard for me to imagine that someone coming to Dagster and seeing "Validating data with Dagster Type factories" would have any idea about what to do with it.

I also think several (most?) of these guides don't really fit the description "accomplish common tasks in Dagster":
- Enriching with Software-defined Assets - this is a guide on how to upgrade, not a guide on how to accomplish a task using Dagster
- Migrating to graphs, jobs, and ops - same as above
- Using the Hacker News example project - this is an overview of an example, not a task you'd accomplish

docs/content/integrations.mdx
```diff
@@ -5,59 +5,173 @@ description: Dagster is flexible and allows incremental adoption. It provides ad
 
 # Integrations
 
-Dagster is flexible and allows incremental adoption. It provides add-on libraries to integrate with your existing tools and infrastructure.
```
sryza, line 8: I think this is an improvement

docs/content/guides.mdx
erinkcochran87: Good catch, thanks!

docs/content/guides.mdx
erinkcochran87: That's a good point. This page was really tricky - I definitely didn't love it when I pushed this up the first time.

I took another shot at this. It doesn't use the two-column list like the other pages do, however - I opted for a single so summaries could be added without everything running together. At least with this approach (versus the original table), the elements are consistent from page to page, even if the columns are different.

sryza (approved):
I like it!

PR #9110, opened 2022-07-29
Publicity Audit for definition-related classes

dpeng817:
- Adds `PendingNodeInvocation` to top level API
- Adds public decorator to relevant methods on `ScheduleDefinition`, `SensorDefinition`, `OpDefinition`, `ResourceDefinition`, `GraphDefinition`, `JobDefinition`. 
- In some cases, pulls attributes off of superclass for public marking purposes. There's definitely a better way to handle this, but crunch is crunch.
- Removes unnecessary args/attributes from `InitResourceContext` because it's easy. Doesn't do this for assorted definition objects because it is not easy.


python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -451,10 +462,20 @@ def check_valid_run_requests(self, run_requests: Sequence[RunRequest]):
...
+    def job_name(self) -> Optional[str]:
+        return self.pipeline_name
+
     @property
     def pipeline_name(self) -> Optional[str]:
         return self._target.pipeline_name if self._target else None
 
+    @public  # type: ignore
+    @property
+    def op_selection(self) -> Optional[Sequence[str]]:
```
sryza, line 476: Are we going to add an op_selection argument to `@sensor`? In the past I think we've considered just telling people to do the selection on the job instead. If that's the case, we shouldn't expose this.  Same for schedule.

sryza (approved):
lgtm except for comment I had on SensorDefinition.op_selection

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
dpeng817: gotcha. sounds reasonable.

PR #7248, opened 2022-03-31
interpret None type annotations as Nothing dagster type

sryza:
## Summary

If a parameter or return value's value will always be None, there's no point saving / loading it with an IO manager.

Not sure whether we should do this on the input side, at least for now - I think it will require a few more changes to input loading.

## Checklist

- [ ] My change requires a change to the documentation and I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.

PR #9112, opened 2022-07-30
make None annotations result in Nothing type

sryza:
Prior to this change, the following test would fail:
```
    @asset
    def my_asset() -> None:
        pass

    assert my_asset.op.outs["result"].dagster_type.typing_type is None
```

`my_asset.op.outs["result"].dagster_type.typing_type` would come out as `Any`

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/input.py
+++ b/python_modules/dagster/dagster/_core/definitions/input.py
@@ -1,3 +1,4 @@
+import inspect
 from types import FunctionType
 from typing import (
     TYPE_CHECKING,
@@ -326,7 +327,15 @@ class InputDefinition:
 
 def _checked_inferred_type(inferred: InferredInputProps, decorator_name: str) -> DagsterType:
     try:
-        resolved_type = resolve_dagster_type(inferred.annotation)
+        if inferred.annotation == inspect.Parameter.empty:
+            resolved_type = resolve_dagster_type(None)
+        elif inferred.annotation is None:
+            # When inferred.annotation is None, it means someone explicitly put "None" as the
+            # annotation, so want to map it to a DagsterType that checks for the None type
+            resolved_type = resolve_dagster_type(type(None))
+        else:
+            resolved_type = resolve_dagster_type(inferred.annotation)
+
     except DagsterError as e:
         raise DagsterInvalidDefinitionError(
             f"Problem using type '{inferred.annotation}' from type annotation for argument "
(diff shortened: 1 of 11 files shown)


python_modules/dagster/dagster/_core/definitions/input.py
```diff
@@ -326,7 +327,13 @@ def combine_with_inferred(
 
 def _checked_inferred_type(inferred: InferredInputProps, decorator_name: str) -> DagsterType:
     try:
-        resolved_type = resolve_dagster_type(inferred.annotation)
+        if inferred.annotation == inspect.Parameter.empty:
+            resolved_type = resolve_dagster_type(None)
+        elif inferred.annotation is None:
+            resolved_type = resolve_dagster_type(type(None))
```
alangenfeld, line 333: worth some quick comments to make the intention here clear since its subtle

python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_decorators.py
```diff
@@ -443,6 +444,34 @@ def my_asset(_input1: str):
         pass
 
     assert my_asset.op.input_defs[0].dagster_type.display_name == "String"
+    assert my_asset.op.input_defs[0].dagster_type.typing_type == str
+
+
```
alangenfeld, line 449: [1] we should add some tests to clarify the choice, something like
```
@op
def produce() -> None:
    pass


@op
def consume(x):
    return x


@graph
def ok_or_invalid():
    consume(produce())
```

and the equivalent in `@asset` - assuming that `-> None` becoming `Nothing` would cause you to change how your asset deps are defined.

alangenfeld (commented):
this has some overlap with https://github.com/dagster-io/dagster/pull/9094 and is related to https://github.com/dagster-io/dagster/pull/7248

with the issue I raised in #7248 - should we relax the `Nothing` constraints to allow inputs to be present for `Nothing` values or do we want to have users add `ins={}` arguments for any thing they take a dep on that `-> None` ?  [1]

alangenfeld (approved):
cool I think this looks right

python_modules/dagster/dagster_tests/core_tests/test_multi_dependency.py
```diff
@@ -225,24 +225,18 @@ def _target_composite_dsl(str_in, none_in):
...
-                        [
-                            DependencyDefinition("emit_num"),
-                            DependencyDefinition("emit_nothing"),
-                            DependencyDefinition("emit_str"),
-                        ]
-                    )
-                }
-            },
-        )
+    PipelineDefinition(
```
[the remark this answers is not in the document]
sryza: eh I was tempted, but thought maybe that kind of change should live in a separate PR

PR #9115, opened 2022-07-30
publicity for partitions stuff

sryza:
PartitionsDefinition and subclasses, PartitionedConfig, PartitionKeyRange, TimeWindow


dpeng817 (approved):
Looks reasonable

PR #9129, opened 2022-08-01
Mark public methods on a bunch of classes

sryza:
Field, AssetSelection, RunRequest, SkipReason, ExpectationResult, OutputMapping, InputMapping, DagsterType, IOManager, IOManagerDefinition, TypeCheck, TypeCheckContext, InitLoggerContext, HookContext, EventLogEntry, RunStatusSensorContext

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_selection.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_selection.py
@@ -4,6 +4,7 @@ from functools import reduce
 from typing import AbstractSet, FrozenSet, Optional, Sequence, Union
 
 import dagster._check as check
+from dagster._annotations import public
 from dagster._core.errors import DagsterInvalidSubsetError
 from dagster._core.selector.subset_selector import (
     fetch_connected,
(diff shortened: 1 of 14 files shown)


python_modules/dagster/dagster/_core/definitions/asset_selection.py
```diff
@@ -39,28 +40,33 @@ class AssetSelection(ABC):
             AssetSelection.groups("marketing") & AssetSelection.keys("leads").downstream()
     """
 
+    @public  # type: ignore
     @staticmethod
     def all() -> "AllAssetSelection":
```
dpeng817, line 45: these types should be added to the top level API perhaps a la `AllAssetSelection`

python_modules/dagster/dagster/_core/definitions/input.py
```diff
@@ -372,13 +372,7 @@ class InputMapping(
...
     )
 ):
-    """Defines an input mapping for a composite solid.
-
-    Args:
-        definition (InputDefinition): Defines the input to the composite solid.
-        solid_name (str): The name of the child solid onto which to map the input.
-        input_name (str): The name of the input to the child solid onto which to map the input.
-    """
+    """Defines an input mapping for a graph."""
```
dpeng817, line 375: is args removal intentional here?

python_modules/dagster/dagster/_core/definitions/output.py
```diff
@@ -339,15 +339,9 @@ def node_name(self):
 class OutputMapping(
     NamedTuple("_OutputMapping", [("definition", OutputDefinition), ("maps_from", OutputPointer)])
 ):
-    """Defines an output mapping for a composite solid.
+    """Defines an output mapping for a graph."""
```
dpeng817, line 342: args removal intentional here?

python_modules/dagster/dagster/_core/definitions/run_config.py
```diff
@@ -106,30 +106,37 @@ def define_execution_field(executor_defs: Sequence[ExecutorDefinition]) -> Field
...
+        return Field(Permissive(), is_required=False, default_value={}, description=description)
 
-    return Field(selector)
+    return Field(selector, description=description)
 
 
-def define_single_execution_field(executor_def: ExecutorDefinition) -> Field:
-    return def_config_field(executor_def)
+def define_single_execution_field(executor_def: ExecutorDefinition, description: str) -> Field:
+    return def_config_field(executor_def, description=description)
```
dpeng817, line 125: yoikes. nice catch

python_modules/dagster/dagster/_core/types/dagster_type.py
```diff
@@ -190,15 +192,18 @@ def from_builtin_enum(builtin_enum) -> "DagsterType":
     def metadata_entries(self) -> t.List[MetadataEntry]:
         return self._metadata_entries  # type: ignore
 
+    @public  # type: ignore
     @property
     def required_resource_keys(self) -> TypingAbstractSet[str]:
```
dpeng817, line 197: naming here is rough

dpeng817 (approved):
A few questions but not enough to block

python_modules/dagster/dagster/_core/definitions/input.py
sryza: Yeah - two reasons:
- The documented args aren't actually accurate
- We don't want people to be constructing these, because the only acceptable values are a legacy type (InputDefinition) and a non-public type

python_modules/dagster/dagster/_core/definitions/output.py
sryza: Yeah - same reasoning as with InputMapping

python_modules/dagster/dagster/_core/definitions/asset_selection.py
sryza: my impulse would be to keep them private - they're kind of an implementation detail. though maybe by that reasoning the return types of these functions should just be AssetSelection?

python_modules/dagster/dagster/_core/types/dagster_type.py
sryza: yeah I agree, though alas I think the ship has sailed on this one

PR #9137, opened 2022-08-01
remove legacy arguments from api docs

sryza:
Unclear if we'll be able to remove the actual arguments in time for 1.0, but at the very least we can remove them from the docs.


dpeng817 (approved):
Nice

PR #9138, opened 2022-08-01
fix group_name for graph-backed asset with key prefix

### LINKED ISSUE #9114: incorrect dagit display if both `key_prefix` and `group_name` are set using AssetsDefinition.from_graph()
### Dagster version

0.15.7

### What's the issue?

Suppose an asset is defined using `AssetsDefinition.from_graph()` (or `AssetsDefinition.from_op()`). 
- If `key_prefix` is specified, then the asset appears in the "default" group in dagit, and the key prefix is displayed correctly.
- If `group_name` is specified, then the asset appears in the specified group in dagit, and there is no key prefix.
- If both `key_prefix` and `group_name` are specified, then the asset appears in the "default" group (instead of the specified group), and the key prefix is displayed correctly.

If both `key_prefix` and `group_name` are specified, then the asset should appear in the specified group (instead of the "default" group), and the key prefix should be displayed correctly.

### How to reproduce?

Define the following three assets and place each one in a repo:
```
@op()
def my_op():
    return True

@graph
def my_graph():
    return my_op()

# displays correctly
asset_with_group = AssetsDefinition.from_graph(graph_def=my_graph, group_name="my_group")

# displays correctly
asset_with_prefix = AssetsDefinition.from_graph(graph_def=my_graph, key_prefix="my_prefix")

# is placed in the "default" group instead of "my_group"
asset_with_group_and_prefix = AssetsDefinition.from_graph(
    graph_def=my_graph, 
    key_prefix="my_prefix",
    group_name="my_group"
)
```

Incidentally, if all three are placed in the same repo, e.g.,
```
@repository
def my_repo():
    return [asset_with_group, asset_with_prefix, asset_with_group_and_prefix]
```
then dagit spits out GraphQL error like

```
Operation name: SidebarAssetQuery

Message: 'my_graph_3'

Path: ["assetNodeOrError","op"]

Locations: [{"line":25,"column":3},{"line":113,"column":3}]
```
when trying to click on the assets in the "default" group.

### Deployment type

_No response_

### Deployment details

_No response_

### Additional information

_No response_

sryza:
### Summary & Motivation

Fixes #9114

Added tests and also loaded up the example from the issue in local dagit.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/assets.py
@@ -360,7 +360,7 @@ class AssetsDefinition(ResourceAddable):
         # For graph backed assets, we assign all assets to the same group_name, if specified.
         # To assign to different groups, use .with_prefix_or_groups.
         group_names_by_key = (
-            {asset_key: group_name for asset_key in keys_by_output_name.values()}
+            {asset_key: group_name for asset_key in keys_by_output_name_with_prefix.values()}
             if group_name
             else None
         )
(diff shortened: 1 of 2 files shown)


sryza:
ping @jamiedemaria @shalabhc - I'd like to get this in to tomorrow's release because it's been reported a couple times

shalabhc (approved):

PR #9140, opened 2022-08-01
Publicity audit for dagster core

dpeng817:
Adds publicity attribution for remaining public classes in `dagster._core`. Too many to effectively list; but some notable inclusions:
- `DagsterRun`, had to change the docs for this to get inherited members.
- `ExecuteInProcessResult`, also had to change the docs here
- `DagsterInstance`, which follows the audit that @alangenfeld did a while back for BITL.
- Notes that users should not instantiate `EventLogEntry`
- `InitExecutorContext` exposes `IPipeline`, which is a private, somewhat legacy API. Since `InitExecutorContext` is experimental (only used when someone is writing their own executor, I feel like it's probably fine to not do anything about this in the short term, but thought I would flag).

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/events.py
+++ b/python_modules/dagster/dagster/_core/definitions/events.py
@@ -230,10 +230,12 @@ class Output(Generic[T]):
     def metadata_entries(self) -> Sequence[Union[PartitionMetadataEntry, MetadataEntry]]:
         return self._metadata_entries
 
+    @public  # type: ignore
     @property
     def value(self) -> Any:
         return self._value
 
+    @public  # type: ignore
     @property
     def output_name(self) -> str:
         return self._output_name
@@ -296,14 +298,17 @@ class DynamicOutput(Generic[T]):
     def metadata_entries(self) -> Sequence[Union[PartitionMetadataEntry, MetadataEntry]]:
         return self._metadata_entries
 
+    @public  # type: ignore
     @property
     def mapping_key(self) -> str:
         return self._mapping_key
 
+    @public  # type: ignore
     @property
     def value(self) -> T:
         return self._value
 
+    @public  # type: ignore
     @property
     def output_name(self) -> str:
         return self._output_name
@@ -638,9 +643,9 @@ class TypeCheck(
     NamedTuple(
         "_TypeCheck",
         [
-            ("success", bool),
-            ("description", Optional[str]),
-            ("metadata_entries", List[MetadataEntry]),
+            ("success", PublicAttr[bool]),
+            ("description", PublicAttr[Optional[str]]),
+            ("metadata_entries", PublicAttr[List[MetadataEntry]]),
         ],
     )
 ):
--- a/python_modules/dagster/dagster/_core/instance/__init__.py
+++ b/python_modules/dagster/dagster/_core/instance/__init__.py
@@ -750,8 +750,8 @@ class DagsterInstance:
     # run storage
     @public
     @traced
-    def get_run_by_id(self, run_id: str) -> Optional[PipelineRun]:
-        return self._run_storage.get_run_by_id(run_id)
+    def get_run_by_id(self, run_id: str) -> Optional[DagsterRun]:
+        return cast(DagsterRun, self._run_storage.get_run_by_id(run_id))
 
     @traced
     def get_pipeline_snapshot(self, snapshot_id: str) -> "PipelineSnapshot":
@@ -1380,7 +1380,6 @@ class DagsterInstance:
 
     # asset storage
 
-    @public
     @traced
     def all_asset_keys(self):
         return self._event_storage.all_asset_keys()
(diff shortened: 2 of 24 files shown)


python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -423,6 +426,7 @@ def get() -> "DagsterInstance":
 
         return DagsterInstance.from_config(dagster_home_path)
 
+    @public
```
sryza, line 429: I think it would be good for @prha to weigh in on which of these DagsterInstance methods we should mark public

sryza (commented):
Mostly looks good - a couple comments

python_modules/dagster/dagster/_core/definitions/events.py
```diff
@@ -225,14 +226,17 @@ def __init__(
         self._output_name = check.str_param(output_name, "output_name")
         self._metadata_entries = normalize_metadata(metadata, metadata_entries)
 
+    @public  # type: ignore
```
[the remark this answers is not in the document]
dpeng817: ack

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -744,6 +748,7 @@ def dispose(self):
         self._compute_log_manager.dispose()
 
     # run storage
+    @public
```
alangenfeld, line 751: is there a run record variant of this we would like to promote instead? I dunno, the cost on this one is very low and pretty high utility so im fine if we end up supporting both

alangenfeld:
For `DagsterInstance` stuff - I think we may want to be more conservative on the `asset` related methods that we expose, theres quite a few `@public` and I'm not sure we want to commit to supporting them in their current form. We can always add @public , but we can't remove it so I would bias towards taking it off anything we're not certain on. Would love to hear from @prha for the remaining asset instance methods.

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1373,18 +1380,22 @@ def end_watch_event_logs(self, run_id, cb):
 
     # asset storage
 
+    @public
```
prha, line 1383: I think this should not be public... I think you can get the same effect by calling `get_asset_keys`

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1373,18 +1380,22 @@ def end_watch_event_logs(self, run_id, cb):
...
     @traced
     def get_asset_keys(self, prefix=None, limit=None, cursor=None):
         return self._event_storage.get_asset_keys(prefix=prefix, limit=limit, cursor=cursor)
 
+    @public
     @traced
     def has_asset_key(self, asset_key: AssetKey) -> bool:
         return self._event_storage.has_asset_key(asset_key)
 
+    @public
```
prha, line 1398: I think should probably not be public

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1412,17 +1423,20 @@ def get_event_records(
...
         return self._event_storage.get_event_records(event_records_filter, limit, ascending)
 
+    @public
     @traced
     def get_asset_records(
         self, asset_keys: Optional[Sequence[AssetKey]] = None
     ) -> Iterable["AssetRecord"]:
         return self._event_storage.get_asset_records(asset_keys)
 
+    @public
```
prha, line 1433: I think not public.

prha, line 1433: Let's make this not public.

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1412,17 +1423,20 @@ def get_event_records(
         """
         return self._event_storage.get_event_records(event_records_filter, limit, ascending)
 
+    @public
```
prha, line 1426: It's surprising to me that `get_event_records` is not marked as public.

I know that people are already using `get_event_records`, but I suspect that people are not really using `get_asset_records`.

I think we probably want to move in the direction of `get_asset_records` (and deprecating `get_event_records`), but I don't think that's realistic in terms of how people are using OSS right now.  

cc @alangenfeld @sryza

python_modules/dagster/dagster/_core/instance/__init__.py
prha, line 1398: I think this should probably not be public.

python_modules/dagster/dagster/_core/instance/__init__.py
dpeng817: this seems reasonable to me - if we're going to be marking `get_event_records` public, should we also make `RunRecord` available in the top-level API? At the very least, we should probably put it in the API docs.

sryza (approved):

python_modules/dagster/dagster/_core/instance/__init__.py
prha: I think that should be safe, yeah

PR #9143, opened 2022-08-01
migrate test_nothing_dependencies to non-legacy APIs

sryza:
### Summary & Motivation

I was working on adding a test to #9112 that would go in this file and found it difficult to parse what the existing tests were doing to figure out whether the cases I wanted to add were already there. So I migrated them.


dpeng817 (approved):
Good samaritan behavior

PR #9146, opened 2022-08-02
[1.0] Mark a bunch of stuff experimental

dpeng817:
Marks FileManager APIs experimental, custom executors experimental, and adds some memoization imports / docstrings. 

One interesting thing is the blurb at the top of `internals.rst` - `internals.rst` in general has a bunch of imports from... internal modules. Do we want to keep that as the case? If so, how do we communicate against it? Took a stab, but would be good to have a once over on this stuff.

### DIFF
--- a/docs/sphinx/sections/api/apidocs/internals.rst
+++ b/docs/sphinx/sections/api/apidocs/internals.rst
@@ -14,7 +14,7 @@ customize it.
 Executors (Experimental)
 ------------------------
 
-APIs for constructing custom executors. While this is considered advanced experimental usage, significant advanced warning will be given before any breakages occur to this set of APIs. Please note that using dagster-provided executors is considered stable, common usage.
+APIs for constructing custom executors. This is considered advanced experimental usage. Please note that using Dagster-provided executors is considered stable, common usage.
 
 .. autodecorator:: executor
 
--- a/docs/sphinx/sections/api/apidocs/internals.rst
+++ b/docs/sphinx/sections/api/apidocs/internals.rst
@@ -3,8 +3,7 @@ Internals
 
 .. currentmodule:: dagster
 
-Please note that internal APIs are likely to be in much greater flux pre-1.0 than user-facing APIs,
-particularly if not exported in the top level ``dagster`` module.
+Note that APIs imported from Dagster submodules are not considered stable, and are potentially subject to change in the future.
 
 If you find yourself consulting these docs because you are writing custom components and plug-ins,
 please get in touch with the core team `on our Slack <https://join.slack.com/t/dagster/shared_invite/enQtNjEyNjkzNTA2OTkzLTI0MzdlNjU0ODVhZjQyOTMyMGM1ZDUwZDQ1YjJmYjI3YzExZGViMDI1ZDlkNTY5OThmYWVlOWM1MWVjN2I3NjU>`_.
@@ -12,8 +11,11 @@ We're curious what you're up to, happy to help, excited for new community contri
 to make the system as easy to work with as possible -- including for teams who are looking to
 customize it.
 
-Executors
----------
+Executors (Experimental)
+------------------------
+
+APIs for constructing custom executors. While this is considered advanced experimental usage, significant advanced warning will be given before any breakages occur to this set of APIs. Please note that using dagster-provided executors is considered stable, common usage.
+
 .. autodecorator:: executor
 
 .. autoclass:: ExecutorDefinition
@@ -27,8 +29,8 @@ Executors
 
 ----
 
-File Manager
---------------
+File Manager (Experimental)
+---------------------------
 
 .. currentmodule:: dagster._core.storage.file_manager
 
@@ -39,6 +41,11 @@ File Manager
 
 .. autodata:: local_file_manager
    :annotation: ResourceDefinition
+
+.. autoclass:: FileHandle
+   :members:
+
+.. autoclass:: LocalFileHandle
 
 ----
 
--- a/docs/sphinx/sections/api/apidocs/io-managers.rst
+++ b/docs/sphinx/sections/api/apidocs/io-managers.rst
@@ -46,6 +46,16 @@ Built-in IO Managers
   :annotation: IOManagerDefinition
 
 
+Input Managers (Experimental)
+----------------------------------
+
+Input managers load inputs from either upstream outputs or from provided default values.
+
+.. autodecorator:: input_manager
+
+.. autoclass:: InputManager
+    :members:
+
 Root Input Managers (Experimental)
 ----------------------------------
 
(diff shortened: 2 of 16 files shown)


docs/sphinx/sections/api/apidocs/internals.rst
```diff
@@ -3,17 +3,19 @@ Internals
 
 .. currentmodule:: dagster
 
-Please note that internal APIs are likely to be in much greater flux pre-1.0 than user-facing APIs,
-particularly if not exported in the top level ``dagster`` module.
+Note that APIs imported from Dagster submodules are not considered stable, and are potentially subject to change in the future.
```
sryza, line 6: 👍

docs/sphinx/sections/api/apidocs/internals.rst
```diff
@@ -3,17 +3,19 @@ Internals
...
 We're curious what you're up to, happy to help, excited for new community contributions, and eager
 to make the system as easy to work with as possible -- including for teams who are looking to
 customize it.
 
-Executors
----------
+Executors (Experimental)
+------------------------
+
+APIs for constructing custom executors. While this is considered advanced experimental usage, significant advanced warning will be given before any breakages occur to this set of APIs. Please note that using dagster-provided executors is considered stable, common usage.
```
sryza, line 17:
```suggestion
APIs for constructing custom executors. While this is considered advanced experimental usage, significant advanced warning will be given before any breakages occur to this set of APIs. Please note that using Dagster-provided executors is considered stable, common usage.
```

sryza: I don't think we necessarily need to promise "significant advanced warning". Otherwise, this text looks good.

docs/sphinx/sections/api/apidocs/io-managers.rst
```diff
@@ -46,6 +46,16 @@ Built-in IO Managers
   :annotation: IOManagerDefinition
 
 
+Input Managers (Experimental)
```
sryza, line 49: I think these are stable, not experimental. @jamiedemaria and/or @OwenKephart can confirm.

docs/sphinx/sections/api/apidocs/libraries/dagster-aws.rst
```diff
@@ -38,6 +32,17 @@ S3
 .. autoconfigurable:: dagster_aws.s3.s3_pickle_io_manager
   :annotation: IOManagerDefinition
 
+File Manager (Experimental)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
sryza, line 36: We normally use dashes, not carets

docs/sphinx/sections/api/apidocs/libraries/dagster-aws.rst
```diff
@@ -38,6 +32,17 @@ S3
...
+
+File Manager APIs will be removed from the main dagster library in version ``1.1.0``, and these APIs will be removed in version ``0.17.0`` of ``dagster-aws``.
```
sryza, line 38: I'm not confident that we should remove file managers.  I think it's worth thinking about them a little more before making this promise.

python_modules/dagster/dagster/_core/definitions/version_strategy.py
```diff
@@ -68,6 +67,11 @@ def get_resource_version(
 
 
 class SourceHashVersionStrategy(VersionStrategy):
+    """VersionStrategy that checks for changes to the source code of ops and resources.
+
+    Only checks for changes within the immediate body of the op/resource's decorated function (or compute function, if the op/resource was constructed directly from a definition).
```
sryza, line 72: Line length

docs/sphinx/sections/api/apidocs/libraries/dagster-aws.rst
dpeng817: i believe carets are for subheadings, but I could be wrong

docs/sphinx/sections/api/apidocs/libraries/dagster-aws.rst
dpeng817: Roger. Right now, FileManager is only accessible via a submodule import. Should we change that?

docs/sphinx/sections/api/apidocs/libraries/dagster-aws.rst
sryza: Ah oops ok I believe you

docs/sphinx/sections/api/apidocs/libraries/dagster-aws.rst
sryza: Interesting - eh no strong opinion either way, but if you left it up to me, I'd say leave it how it is

docs/sphinx/sections/api/apidocs/io-managers.rst
jamiedemaria: I haven't had much time to look into this, but i recently ran across  a funky edge case where I'm concerned that input managers don't cover the same functionality as root input managers. here's a (somewhat) trivial run down:

if we have op_a return 2.  op_b takes it as input, adds 1 and returns 
if we use a root input manager to provide 3 as input to op_b and just execute op_b we get 4 as expected. 

the typical case for using an input manager is to load the input to an op in a special way (ie output from an op is a dataframe, but we want it in a different format for the next op). 
if we have an input manager on op_b that multiplies the output by 2, we execute the job and get 5 as expected. 

the other use case for input managers is to replace root input managers
if we have a new input manager that provides 3 as input to op_b and we execute op_b we get 4 as expected. BUT if we execute the full job, the input manager will still provide 3 to op_b. this is because there isn't a good way to determine which type of input manager (the first version or the root manager replacing version) is intended

the way we get around this is by writing logic in the input manager to check if there is an upstream output, and if so we handle it differently (the example in the docs is to make the input manager extend from an IO manager and then call the IO manager's load_input function) 

this works pretty well except for the in-memory io manager because we don't have a way to access the instantiation of the in-mem io manager that is storing the output


I haven't had time to dig into this yet, but on the chance that addressing it causes the user facing API for input managers to shift a little bit, I think it should stay experimental.

That was a very long comment and a bit of a half baked example of the issue. I can write up an working code snippet tomorrow or hop on a call to explain if that would be helpful

dpeng817: If we're unsure, best to keep it experimental now IMO, especially since we aren't getting rid of type loaders or materializers yet. We can always un-experimentalize later with little to no cost.

python_modules/dagster/dagster/__init__.py
```diff
@@ -307,6 +312,21 @@
...
+        "dagster._core.storage.file_manager",
+        "1.1.0",
+        "It is recommended to handle I/O to a filesystem within an IO manager.",
+    ),
+    "LocalFileHandle": (
+        "dagster._core.storage.file_manager",
+        "1.1.0",
+        "Local file I/O can be handled by the fs_io_manager.",
+    ),
+    "local_file_manager": (
```
sryza, line 325: Assuming we're not targeting a removal date for file manager, we should take out these deprecations.

sryza (approved):
LGTM aside from one comment

python_modules/dagster/dagster/__init__.py
dpeng817: roger

PR #9157, opened 2022-08-02
Make direct resources experimental

dpeng817:
as the title. Applied to `asset`, `multi_asset`, `SourceAsset` and `AssetsDefinition.from_graph`.


sryza (approved):
awesome

PR #9160, opened 2022-08-02
[docs] - Implement new landing page design

erinkcochran87:
This PR implements the new landing page design from #8942 

---

To do:

- [x] Finalize examples
- [x] Update 'elevator pitch' copy
- [x] Add descriptions to examples


erinkcochran87:
@sryza @yuhan I temporarily removed the `**New to Dagster?**` line and buttons in the intro. I'm not sure that we can get the overview out before Tuesday (but I'm certainly trying) and didn't want that to block the rest of these improvements. Everything else is the same as the original design.

sryza:
Without the "New to Dagster" section, this makes me think that the first thing we want people to do is start with one of the examples. I don't think those examples are the best way to learn for most people because they don't really explain what's going on in a way that's targeted at people who are new to Dagster.

What do you think about having a "poor-man's version" of the "New to Dagster" section that prompts people to either take the (Asset) Tutorial or go to the Concepts category page?

yuhan:
I left some comments in the preview.

and i also agree with sandy's point about missing new to dagster doesn't feel right to a new user

erinkcochran87:
@sryza @yuhan That's fine by me - I'll add it in

erinkcochran87:
@sryza @yuhan Back to you folks. Sorry about the ping; I can't re-request since reviews weren't actually completed.

yuhan:
FYI - @erinkcochran87 i noticed broken external links in the preview so i pushed a fix directly to your branch

yuhan:
looking great. a few comments
- maybe rename templates -> examples
- a few broken/empty links

erinkcochran87:
Addressed. `snapshot` didn't catch that broken Helm link (weird). I opted to temporarily remove the link to the Dagster Cloud page, as I realized not all of the content on that specific page won't be ready.

I have to add the Cloud pages to the sidneav anyway, so I can add it in then.

yuhan (approved):

sryza (approved):
I love it!

PR #9163, opened 2022-08-02
[docs] - Sidenav cleanup

erinkcochran87:
This PR cleans up the sidenav in the docs:

- Consolidates some top-level items
- Removes a section from **Integrations**
- Adds **About** section
- Adds links to **Community**

### DIFF
--- a/docs/content/_navigation.json
+++ b/docs/content/_navigation.json
@@ -15,10 +15,6 @@
       {
         "title": "Telemetry",
         "path": "/getting-started/telemetry"
-      },
-      {
-        "title": "Releases and Changes",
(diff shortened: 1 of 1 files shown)


sryza (commented):
What would you think about consolidating "About" and "Community" into a single section - maybe called "Project"?

erinkcochran87:
@sryza I like keeping Community separate. I don't think I'd look in a category called *Project* or *About* to find support. I'd like to make this visible and easy to access.

If you're concerned about the number of items in the sidenav right now, remember that Cloud is going away soon :)

sryza (approved):
👍

PR #9165, opened 2022-08-02
remove dagster cli api subcommand from docs

jamiedemaria:
### Summary & Motivation

@sryza I think this addresses `Remove api subcommands of CLI from docs site` in the quip doc, but lmk if I misunderstood

Currently having issues getting the docs to build (i think pip library inconsistencies causing errors), so need to get that fixed before this is ready to merge


sryza:
I think I still see them here: https://dagster-git-jamie-cli-api-docs-elementl.vercel.app/_apidocs/cli#dagster-api-grpc

jamiedemaria:
@sryza yeah I'm having issues rebuilding the docs locally so the json isn't getting updated. Just wanted to confirm that removing that line from the api docs was all i needed to do. still debugging why i can't build docs

sryza:
Gotcha - yeah, if we can get that line to disappear on the site, I think we're good.

yuhan:
per offline discussion, i think we can discard the json changes here - 1) it's one line removal which is safe enough to merge 2) im also worried that checking in json files and the objects.inv at this point may cause merge conflicts with other api related diffs

sryza (approved):

PR #9169, opened 2022-08-02
fix master: downgrade pinned virtualenv pip in tox

yuhan:
### Summary & Motivation
unbreak master for now.

i thought we dropped py36 but it seems like it's still testing against 36


alangenfeld (approved):
accepting to fast fix master
but i think you can put out a PR to drop 3.6 that re-updates this

PR #9171, opened 2022-08-02
Docs search use window.location.href instead

clairelin135:
[Linear](https://linear.app/elementl/issue/DREL-328/%5Bdocs%5D-nextjs-redirect-routing-doesnt-work-with-anchors#comment-c97d097e)

Navigating via search to pages that contain redirects do not navigate to the correct section. The current implementation is to use `NextRouter` to avoid reloading all contents of the page, but the caching/redirecting is potentially interfering with navigation to the appropriate section.

In the future, we can revert back to this behavior, but for now, this PR amends the navigation behavior to do a direct reload on search so that the appropriate section will be loaded. I think the added load time is worth the search working correctly.

Test this here: https://dagster-git-claire-search-elementl.vercel.app/getting-started

### DIFF
--- a/docs/next/components/Search.tsx
+++ b/docs/next/components/Search.tsx
@@ -2,9 +2,7 @@ import { DocSearchModal, useDocSearchKeyboardEvents } from "@docsearch/react";
 import { useCallback, useEffect, useRef, useState } from "react";
 
 import Head from "next/head";
-import Link from "next/link";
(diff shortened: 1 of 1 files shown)


yuhan (approved):
i think this looks great - our site is fast enough and a full load wont be too bad for user experience. plus we can always revert this.

PR #9175, opened 2022-08-02
Remove Legacy section from API reference page

sryza:
### Summary & Motivation

I missed this in my earlier change that ripped out legacy API docs


erinkcochran87 (approved):
This looks safe. I'm not sure if we need to add any redirects for these?

PR #9183, opened 2022-08-02
remove op/graph input_defs/output_defs

dpeng817:
as the title.

This brings to light a bug with graph inputs; graph ins have no way of specifying dagster type, so you cannot specify a nothing input on a graph. Since the underlying input definition has an all dagster type, config is expected for unconnected graph inputs. To be clear, this isn't a _new_ bug, this PR just exposes it.


sryza (approved):
Awesome

PR #9188, opened 2022-08-03
Remove requirements.txt from scaffold template and examples

yuhan:
### Summary & Motivation
per [slack discussion](https://elementl-workspace.slack.com/archives/C03A0D72A6T/p1658515403719279), we decided to leave requirements out and only rely on setup.py for dependency.


sryza (approved):

jamiedemaria (commented):
can you update from latest and remove `requirements.txt` from `development_to_production`? thanks!

PR #9191, opened 2022-08-03
3.14/n example overhaul: hacker_news_assets -> project_fully_featured  + recommended project layout

yuhan:
### Summary & Motivation

Project layout change:
- Put all assets under `assets/` dir, separated by business domain inside the assets/ dir
It was separate in activity_analytics/assets, core/assets, recommender/assets. Having the business domains to be top-level dirs in a dagster project module makes it hard to discover assets in a project for new authors/collaborators.
- Consolidate where we define schedules and sensors.
	* Moves all asset schedules and sensors to repository.py; they were separate in activity_analytics/__init__.py, core/__init__.py, recommender/__init__.py. I think it was hard to find and get an overview of all sensor/schedule defs in a repo.
- `partitions/__init__.py` -> `partitions.py`
- Include imports in each submodule's root __init__.py, such as `resources/__init__.py` and `assets/__init__.py`. it helps collaborators to find the definitions that are available to use, sometimes encapsulated with certain config via configured like the ones in `resources/__init__.py`

Semantic renaming:
- `dbt` -> `dbt_project` to make it clear it's a dbt project and also consistent with other examples.

check out https://github.com/dagster-io/dagster/tree/yuhan/3.14-hacker_news_assets/examples/hacker_news_assets for easier review.

~**[🙏 Help needed]** I don't have a good intuitive name for this example - trying to convey "complex project", "fully featured", "project layout".~
=> project_fully_featured

unit, and internal


examples/hacker_news_assets/dbt_project/.gitignore
```diff
@@ -1,3 +1,2 @@
 dbt_modules/
 logs/
-target/
```
sryza, line 3: Why take this out?

examples/hacker_news_assets/hacker_news_assets/repository.py
```diff
@@ -0,0 +1,77 @@
+import os
```
sryza, line 1: If an organization had 100 asset groups and 1.5 jobs per asset group, how would we recommend they structure things?  Asking because the HN example is meant to demonstrate how to structure large projects, and, even we don't fully need it for our three asset groups, I still think it's worth using that structure.

examples/hacker_news_assets/tox.ini
```diff
@@ -8,13 +8,14 @@ setenv =
 passenv = CI_* COVERALLS_REPO_TOKEN BUILDKITE SNOWFLAKE_ACCOUNT SNOWFLAKE_USER SNOWFLAKE_PASSWORD
 deps =
   -e ../../python_modules/dagster[mypy,test]
+  -e ../../python_modules/dagit
+  -e ../../python_modules/dagster-graphql
```
sryza, line 12: Why dagster-graphql?

examples/hacker_news_assets/tox.ini
```diff
@@ -8,13 +8,14 @@ setenv =
...
   -e ../../python_modules/libraries/dagster-pandas/
   -e ../../python_modules/libraries/dagster-spark/
   -e ../../python_modules/libraries/dagster-pyspark/
   -e ../../python_modules/libraries/dagster-dbt/
   -e ../../python_modules/libraries/dagster-slack/
   -e ../../python_modules/libraries/dagster-aws/
-  -e ../../python_modules/libraries/dagster-gcp/
```
sryza, line 17: Nice

examples/hacker_news_assets/tox.ini
```diff
@@ -8,13 +8,14 @@ setenv =
 passenv = CI_* COVERALLS_REPO_TOKEN BUILDKITE SNOWFLAKE_ACCOUNT SNOWFLAKE_USER SNOWFLAKE_PASSWORD
 deps =
   -e ../../python_modules/dagster[mypy,test]
+  -e ../../python_modules/dagit
+  -e ../../python_modules/dagster-graphql
   -e ../../python_modules/libraries/dagster-pandas/
```
sryza, line 13: Feel free to disregard, but I don't think we need dagster-pandas or dagster-spark anymore here.

examples/hacker_news_assets/tox.ini
yuhan: some of the dependencies here install `dagster-graphql` if we don't include it as an editable here, tox would install a released version which caused version mismatch in tests.

can audit if that's real necessary but don't think it's blocking the release

examples/hacker_news_assets/dbt_project/.gitignore
yuhan: talked to @OwenKephart, and confirmed that we're using it as the metadata that we load assets from. so it should be checked in.

examples/hacker_news_assets/tox.ini
sryza: gotcha sounds fine

examples/hacker_news_assets/hacker_news_assets/repository.py
sryza: If we have a file like that, I think we should call it jobs.py instead of schedules.py, because not every job will have a schedule, but every schedule will have a job.

yuhan: how's this looking? https://github.com/dagster-io/dagster/tree/yuhan/3.14-hacker_news_assets/examples/hacker_news_assets/hacker_news_assets

sryza (approved):
I like it!

PR #9196, opened 2022-08-03
[docs] - Open source vs Cloud

erinkcochran87:
This PR adds an [**Open source vs Cloud** guide](https://dagster-git-erin-core-vs-cloud-elementl.vercel.app/getting-started/open-source-vs-cloud) to the docs.


elementljarredc (changes requested):
I think this is an OK start. I think it would be greatly improved on if we had a chart which showed the OS architecture and compared it to the Cloud Architecture... 

i.e. here is what a hosted OSS implementation looks like, here is what a hybrid cloud looks like, and here is what Serverless looks like. 

I also wonder if we need to have such an extensive list of things if they're all the same... 

I think we may also need to add UI-based env vars management here (cloud only)

erinkcochran87:
@elementljarredc 

> I think this is an OK start. I think it would be greatly improved on if we had a chart which showed the OS architecture and compared it to the Cloud Architecture...
> i.e. here is what a hosted OSS implementation looks like, here is what a hybrid cloud looks like, and here is what Serverless looks like.

I definitely agree here. This was a rushed first pass at trying to get something in before Dagster Day, so it didn't go too in-depth. I agree that we should add content that outlines Cloud architecture a bit better.

> I also wonder if we need to have such an extensive list of things if they're all the same...

I'm open to changing this, but the motivation at the time was users being confused by Cloud and thinking it was a totally different form of Dagster. I feel like this comes up a lot less often now, but it could still be helpful for brand-new users coming to Dagster.

> I think we may also need to add UI-based env vars management here (cloud only)

We do! This list is definitely outdated - we also don't have `Repositories` anymore.

erinkcochran87:
Closing - I don't think the confusion we once saw around OSS/Cloud is the same as when this was written, so it's no longer needed. We can always revive it if we think it's a good idea.

PR #9202, opened 2022-08-03
Hardcode version numbers in published packages

smackesey:
### Summary & Motivation

This hardcodes the version numbers and dagster pins for published packages. Necessary for new release process.


smackesey:
* master
  * **PR #9202** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9202" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9202?utm_source=stack-comment).

gibsondan (commented):
how does a new person making changes to a package know when to update the string in __version__? Should we have a test that fails if you modify a package and don't change the version?

gibsondan:
oh i see, this is handled during the release via the bump_versions call

smackesey:
>how does a new person making changes to a package know when to update the string in version? Should we have a test that fails if you modify a package and don't change the version?

I have that code in a separate branch but it's not included in the stack for 1.0.0 since all packages are still being bumped every time.

>oh i see, this is handled during the release via the bump_versions call

`dagster-release bump-versions` isn't called as part of the release, it's intended to be called _before_ the release with the results possibly manually tweaked. It also needs some work since it doesn't currently handle updating the pins. Can do this ASAP after thurs release, but `bump-versions` doesn't matter for the release itself since the versions have already been bumped (in this PR-- I used `sed`/`rpl`).

PR #9204, opened 2022-08-03
docs: revamp create new project page 4/n

yuhan:
### Summary & Motivation
Update Create New Project to reflect changes in `dagster project` CLI.

plz feel free to use the new "preview with comments" functionality to leave comments/feedback :)

preview

### DIFF
--- a/docs/content/getting-started/create-new-project.mdx
+++ b/docs/content/getting-started/create-new-project.mdx
@@ -5,79 +5,86 @@ description: Dagster comes with a convenient CLI command for creating a new proj
 
 # Create a New Project
 
-This section will show you how to create a new Dagster project and organize your files as you build larger and larger jobs. Dagster comes with a convenient CLI command for generating a project skeleton, but you can also choose to organize your files differently as your project evolves.
+The easiest way to start building a Dagster project is by using the `dagster project` CLI. This CLI tool helps generates files and folder structures that enables you to quickly get started with Dagster. You can scaffold a new project using the default project skeleton, or start with using one of the official Dagster examples.
 
-If you're completely new to Dagster, we recommend that you visit our [Tutorial](/tutorial) to learn all the basic concepts of Dagster.
+## Option 1: Starting with an empty project skeleton
 
-## Generating a Project Skeleton
-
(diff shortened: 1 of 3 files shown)


erinkcochran87 (commented):
Added some comments on the Vercel preview :D

I like this - I feel like I understand more clearly how to start a project and then what to do next with it. This is great!

yuhan:
@erinkcochran87 updated! thanks for all the feedback!

sryza:
I left a few comments in Vercel but overall this is looking quite good

yuhan:
@sryza updated the docs to address your feedback.

i also included a few code changes on this pr for fastening review (for sure not a good practice):
- list-example -> list-examples
- unexperiment clis
image

docs/content/getting-started/create-new-project.mdx
```diff
@@ -5,79 +5,117 @@ description: Dagster comes with a convenient CLI command for creating a new proj
...
+
+Tests can be added in the `my_dagster_project_tests` directory and you can run tests using `pytest`:
 
 ```bash
-dagster-daemon run
+pytest my_dagster_project_tests_tests
 ```
 
-Once your Dagster Daemon process is running, you should be able to enable schedules and sensors for your Dagster jobs.
+### Schedules and sensors
```
jamiedemaria, line 100: do we want to recommend where users put the code for their schedules/sensors? like recommend putting them in repository.py when they are first getting started, and then refactor when they start writing more schedules?

erinkcochran87 (approved):
Just a few small formatting things, but overall this looks great.

Approved from my end, but I'd feel comfortable having @sryza also give a :+1: before merging

sryza (approved):
I had one small comment, otherwise this looks great.

Doesn't need to be in this PR, but I do think we generally need better recommendations on where to put op-based jobs than repository.py - that will get unwieldy quickly.

yuhan:
For schedules, sensors, and jobs, I think we can say they can start with repository.py and direct users to the fully featured project for best practices (speaking of which, as I was drafting the fully featured project walkthrough, I think we should include a op-based job in that example)

sryza:
> For schedules, sensors, and jobs, I think we can say they can start with repository.py and direct users to the fully featured project for best practices (speaking of which, as I was drafting the fully featured project walkthrough, I think we should include a op-based job in that example)

That sounds right to me. And yeah, the fully-featured project currently doesn't demo best practices for organizing op-based jobs.  Also, graph-backed assets - should the ops for them go inside the assets/ dir?

PR #9211, opened 2022-08-04
[dagit] Expose more context values for Instance page

hellendag:
### Summary & Motivation

Cloud uses slightly different strings in a few places on the Instance section of Dagit. Expose a bit more via context.

Load Dagit, view Instance pages. Verify correct strings. Load Cloud in its current master, verify that it still works correctly. Load Cloud with new context provider in place, verify overridden strings.

### DIFF
--- a/js_modules/dagit/packages/core/src/app/AppProvider.tsx
+++ b/js_modules/dagit/packages/core/src/app/AppProvider.tsx
@@ -25,6 +25,7 @@ import {BrowserRouter} from 'react-router-dom';
 import {createGlobalStyle} from 'styled-components/macro';
 import {SubscriptionClient} from 'subscriptions-transport-ws';
 
+import {InstancePageContext} from '../instance/InstancePageContext';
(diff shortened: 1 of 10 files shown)


hellendag:
* master
  * **PR #9211** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9211" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9211?utm_source=stack-comment).

salazarm (approved):
Approved but I think I would like to see just one Context provider that we add more context too for cloud specific overrides like this, otherwise I can imagine we'd have a lot of different context providers. No change needed in this PR, we can let this evolve naturally and see.

hellendag:
> Approved but I think I would like to see just one Context provider that we add more context too for cloud specific overrides like this, otherwise I can imagine we'd have a lot of different context providers. No change needed in this PR, we can let this evolve naturally and see.

Yeah, I think that's probably the right direction long term.

PR #9212, opened 2022-08-04
Add `custom_path_fs_io_manager` to legacy API removals in migration.md

dpeng817:
TItle


sryza (approved):
We should make sure this gets synced to the changelog as well

PR #9217, opened 2022-08-04
Update Config Examples

clairelin135:
Updates the config documentation examples:
- adds config via env vars
- configuring per-environment resources
- updates existing examples to contain real examples


yuhan (approved):
looks great!

PR #9219, opened 2022-08-04
docs infra: always open the current section in nav

yuhan:


erinkcochran87 (commented):
Added a comment + GIF of some odd behavior on the Vercel preview.

yuhan:
i think a better path is to separate the click on down arrow from clicking the whole menu item. as in clicking 
image should only open/close the section and clicking on the text should actually navigate to the page.

salazarm (approved):

yuhan:
too old to pick up. will cut a fresh branch

PR #9222, opened 2022-08-04
[draft] remove all usages of solid decorator/class in code

dpeng817:
Codemod: https://github.com/dpeng817/codemods/blob/main/src/codemods/codemod_solid.py

Potentially painful bits mostly revolve around dbt solids; I just removed them.


examples/assets_dbt_python/assets_dbt_python/__init__.py
```diff
@@ -1 +0,0 @@
-from .repository import assets_dbt_python
```
dpeng817, line 1: ugh these all got picked up by accident

PR #9223, opened 2022-08-04
docs: recommended project structure

yuhan:
### Summary & Motivation

https://dagster-git-yuhan-docs-project-layout-elementl.vercel.app/guides/dagster/recommended-project-structure


docs/content/guides/dagster/recommended-project-structure.mdx
```diff
@@ -0,0 +1,44 @@
...
+
+  - This patten also helps you to easily execute jobs in different environment without code changes. In this example, we dynamically define repository based on the deployment in [`repository.py`](https://github.com/dagster-io/dagster/blob/master/examples/project_fully_featured/project_fully_featured/repository.py) and are able to keep all code the same across testing, local development, staging, and production. Read more about our recommendations in [Transitioning Data Pipelines from Development to Production](/guides/dagster/transitioning-data-pipelines-from-development-to-production)
+
+- TODO: What to do with ops/graphs/jobs???
```
clairelin135, line 36: I think we decided from the recommended project structure to have an `ops.py` and a `jobs.py`. I can see graphs fitting in with either ops or jobs--we don't necessarily need to address graphs though, we could leave it up to the user. 

Definitely makes sense to me to recommend having all of the ops exist in one file though, for ease of importing and reusing across multiple jobs.

clairelin135 (commented):
Added some preliminary comments, but generally this is looking great to me!

erinkcochran87 (commented):
This seems to still be in fairly early stages, but I like where it's going. I left some comments on the Vercel preview, but didn't look too closely at grammar for this pass. I think the content itself is looking good, but it could use some structuring to make this easier to parse - I'd love to see it again after that's been added in.

docs/content/guides/dagster/recommended-project-structure.mdx
sryza: A medium-sized organization could easily have hundreds of ops, which could mean thousands of lines of code, and I think that would be unwieldy to have in one file.

yuhan:
close as https://github.com/dagster-io/dagster/pull/12656 is merged

PR #9235, opened 2022-08-05
feat(helm): add job_namespace config to CeleryK8sExecutor

gmontanola:
### Summary & Motivation

This can be defined in K8sRunLauncher parameters and setting this via Helm chart for Celery runs
would be great.

Were there any issues not allowing this to happen before?

I haven't for now but I can test this with minikube locally. I followed the YAML trail and looks like this can work.


gmontanola:
Oh well, minutes before 1.0.0.

If it's not possible to merge this, i
I think this will do. 


> A default_executor_def argument has been added to the @repository decorator. If specified, this will be used for any jobs (asset or op) which do not explicitly set an executor_def

Milias:
Hello, is someone willing to take a look at this PR? It is really annoying that new users need to add this setting to the configuration every time, causing unnecessary failed runs.

Thanks!

gibsondan:
Sorry for the long delay here - there are some more substantial changes that would need to be made to support this, which are in https://github.com/dagster-io/dagster/pull/12911 - thanks for sending this out nonetheless!

PR #9247, opened 2022-08-05
[docs] - Clean up Guides sidenav + category page

erinkcochran87:
This PR:

- Removes child items from the **Guides** section of the sidenav. This was cluttering the sidenav. Now, when a user clicks **Guides** they'll be taken to the **Guides** category page, where they can view all guides at once.
   This does make it the only top-level item in the sidenav without children, but I don't think that's really a big deal.
- Adds a **Best practices** section to the **Guides** page


docs/content/_navigation.json
```diff
@@ -477,50 +477,7 @@
   {
     "title": "Guides",
     "icon": "Search",
-    "path": "/guides",
```
yuhan, line 480: removing all child items feels odd that i can no longer find these individual guides in the left nav or can't see the item show up on the left when im inside a particular guide.

in order to organize the nav better, could we instead introduce extra layers:
```
Guides
    - General
    - Best practices
    - Experimental features
```
like this: image

yuhan (commented):
content looks good

docs/content/_navigation.json
erinkcochran87: I had this thought too - do we want to add additional pages for this? Sandy mentioned it was odd to have the sidenav link to subheadings in a page.

docs/content/_navigation.json
```diff
@@ -484,45 +484,16 @@
...
-      },
-      {
-        "title": "Fully-Featured Example Project",
-        "path": "/guides/dagster/example_project"
-      },
-      {
-        "title": "Validating Data with Dagster Type Factories",
-        "path": "/guides/dagster/dagster_type_factories"
+        "title": "Experimental features",
+        "path": "/guides/experimental-features"
```
yuhan, line 496:
```suggestion
        "title": "Experimental features",
        "path": "/guides/experimental-features"
        "children": [
        	{
          		"title": ...,
          		"path": "/guides/dagster/memoization"
          	},
          	{
          		"title": ...,
          		"path": "/guides/dagster/run-attribution"
          	}
        ]
```

should we also include the subpages as `"children"` in each section here like the suggestion^? so the guides can show up in the left nav

erinkcochran87: Yes!

sryza (approved):
This looks good to me

PR #9253, opened 2022-08-05
Add DagsterTypeMaterializerContext and DagsterTypeLoaderContext APIs

dpeng817:
@dagster_type_loader and @dagster_type_materializer both take in a `StepExecutionContext` to their functions, which in 1.0 is in a private submodule. Rather than expose `StepExecutionContext`, this PR adds `DagsterTypeMaterializerContext` and `DagsterTypeLoaderContext` as subclasses of `StepExecutionContext` (necessary so as to not break people upgrading to 1.0). Also adds docs.


smackesey (approved):
lgtm

PR #9268, opened 2022-08-07
ecs run launcher add hooks to change run tags

prha:
### Summary & Motivation
Enables subclasses to change the format of the ECS info stored in tags.

BK

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
@@ -24,7 +24,7 @@ from .container_context import SHARED_ECS_SCHEMA, EcsContainerContext
 from .tasks import default_ecs_task_definition, default_ecs_task_metadata
 from .utils import sanitize_family
 
-Tags = namedtuple("Tags", ["arn", "cluster", "cpu", "memory"])
+EcsInfo = namedtuple("EcsInfo", ["arn", "cluster", "cpu", "memory"])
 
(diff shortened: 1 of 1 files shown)


gibsondan (changes requested):
we ended up doing this differently i think?

prha:
This is separate.... in case we want to change the stored run tags to remove the serverless account id from the arn/cluster.

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -164,20 +164,21 @@ def _set_ecs_tags(self, run_id, task_arn):
...
 
-    def _set_run_tags(self, run_id, task_arn):
+    def set_run_tags(self, run_id, task_arn):
+        # Extracted method to allow for subclasses to customize the ECS tag storage implementation
         cluster = self._task_metadata().cluster
         tags = {"ecs/task_arn": task_arn, "ecs/cluster": cluster}
         self._instance.add_run_tags(run_id, tags)
 
-    def _get_run_tags(self, run_id):
+    def get_ecs_info(self, run_id) -> EcsInfo:
```
gibsondan, line 173: bit confused why we took out the underscore - them being overridden by a subclass doesn't necc make them public I think?

gibsondan: def _get_ecs_info_from_run_tags?

gibsondan (approved):
better late than never

PR #9280, opened 2022-08-08
[dagit] Allow searching by op_selection

hellendag:
### Summary & Motivation

Restore `solid_selection` tag rendering, update Launchpad to use `op_selection` for future runs.

Thus, `solid_selection` and `op_selection` may appear side by side in the run list and the filter input, and @dpeng817 will look at changing the backend to handle filtering.

### How I Tested These Changes

Perform a run with the new `op_selection` key, see it on the Run page and the Runs list. Verify that I can add the tag and filter by it. Verify that `solid_selection` tags also behave as before, addable and filterable.

PR #9284, opened 2022-08-08
interpret solid selection as op selection for querying runs.

dpeng817:
Until https://github.com/dagster-io/dagster/pull/9280 lands, we have been writing selection tags as `solid_selection` instead of `op_selection`. We want to be able to filter runs s.t. when searching for the op_selection tag, historic runs which still stored using the solid_selection tag, show up.

In order to test this; I created some fresh runs on current-master in my local dagster installation using selection; which persisted those runs with the `solid_selection` tag clearly visible. I then rebased Dish's change from https://github.com/dagster-io/dagster/pull/9280 on top of this PR, and re-launched another run, which launched with the `op_selection` tag. Searching by the op_selection tag surfaces the runs with a matching solid selection:
Screen Shot 2022-08-10 at 10 41 25 AM

Searching by solid_selection only surfaces items with the solid_selection tag, which I think is fine.

Requires a companion PR for dagster-cloud; https://github.com/dagster-io/internal/pull/3317

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/runs/in_memory.py
+++ b/python_modules/dagster/dagster/_core/storage/runs/in_memory.py
@@ -1,5 +1,5 @@
 from collections import OrderedDict, defaultdict
-from typing import Callable, Dict, Iterable, List, Optional, Set, Tuple, Union, cast
+from typing import Any, Callable, Dict, Iterable, List, Optional, Set, Tuple, Union, cast
 
 import dagster._check as check
 from dagster._core.errors import (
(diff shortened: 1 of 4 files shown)


prha (commented):
Have we considered other options, like doing a data migration over the solid selection run tags?  Or having the op selection tag permanently be represented in the DB as solid selection and transforming in the storage layer?

dpeng817:
@prha I assumed we wouldn't be comfortable with inducing a data migration, but I guess since the functionality is pretty niche, it's probably OK? How would we surface that a migration is necessary you think?

prha:
We have data migrations which will run when users run `dagster instance migrate`.  We can just put something in the release notes saying that if they see something odd when they search for runs with op selections (e.g. missing runs), they can run `dagster instance migrate`.

The functionality degrades nicely, in that if they do nothing, maybe some run views look wrong, but it'll only be old runs (new runs should have the proper tags).

dpeng817:
Right - was just wondering if we'd want to do anything more in terms of telling users "hey you have runs whose tags might need to be migrated - or something but that makes sense to me.

dpeng817 (commented):
Closed in favor of https://app.graphite.dev/github/pr/dagster-io/dagster/9284

Issue #9277, opened 2022-08-08
PartitionedParquetIOManager is unable to load partitioned input in `project_fully_featured` example project


kien-truong:
In the example project `project_fully_featured`, we have 2 core assets `items`, `stories`. 

If you materialize `items` and `stories` in the same run, everything run correctly. 

However, if they are materialized in 2 separate runs, the first run to materialize `items` will finish but the second run to materialize `stories` will fail. It will be unable to load the `items` assets, because the upstream partition info is no longer available.

I'm testing with Dagster `1.0.1`

sryza:
Thanks for reporting this @kien-truong . I was able to reproduce this and am working on a fix.

PR #9290, opened 2022-08-08
avoid using InputContext.upstream_output in asset examples

sryza:
### Summary & Motivation

"Upstream output" doesn't always make sense in the context of assets, and the implementation also doesn't work well with partitions.

With https://github.com/dagster-io/dagster/pull/9288, this fixes https://github.com/dagster-io/dagster/issues/9277.

Ran the HN demo locally.

### DIFF
--- a/docs/content/concepts/io-management/io-managers.mdx
+++ b/docs/content/concepts/io-management/io-managers.mdx
@@ -544,14 +544,14 @@ class PandasAssetIOManager(IOManager):
         file_path = self._get_path(context)
         store_pandas_dataframe(name=file_path, table=obj)
 
-    def _get_path(self, output_context):
+    def _get_path(self, context):
         return os.path.join(
             "storage",
-            f"{output_context.name}.csv",
+            f"{context.asset_key.path[-1]}.csv",
         )
 
     def load_input(self, context):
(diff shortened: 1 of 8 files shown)


docs/content/concepts/io-management/io-managers.mdx
```diff
@@ -544,14 +544,14 @@ class PandasAssetIOManager(IOManager):
         file_path = self._get_path(context)
         store_pandas_dataframe(name=file_path, table=obj)
 
-    def _get_path(self, output_context):
+    def _get_path(self, context):
         return os.path.join(
             "storage",
-            f"{output_context.name}.csv",
+            f"{context.name}.csv",
```
OwenKephart, line 550: Isn't `context.name` generally `"result"` on the OutputContext (and something else on InputContext)? This wasn't great in the original example either (everything would get stored to `result.csv`) but this impl seems objectively broken.

This should probably be something like "_".join(context.asset_key.path), or asset_key.path[-1]

OwenKephart (approved):
Other than comment, lgtm

docs/content/concepts/io-management/io-managers.mdx
sryza: Mm good point

PR #9311, opened 2022-08-09
[dagit] Guard against missing workspace on repo reload

hellendag:
### Summary & Motivation

It is possible to get into a state where a repo reload will lead to a JS error because the Apollo cache will be reset, leading to an unexpected empty `data` object in the query result. We don't expect to guard against this kind of object in TypeScript, because the `data` object is never expected to empty in this way. Definitely an Apollo bug, but we need to cover it here.

Reload a repo from the Workspace page. Verify that the page doesn't crash.


hellendag:
* master
  * **PR #9311** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9311" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9311?utm_source=stack-comment).

gibsondan (approved):

PR #9318, opened 2022-08-09
make it easy to build custom schedules from partitioned jobs

sryza:
### Summary & Motivation

Reading https://github.com/dagster-io/dagster/pull/9105/files made me think about "what's our replacement for custom partition selectors in a post-legacy world?"

This PR posits that we should basically encourage users to use `@schedule` instead of `build_schedule_from_partitioned_job` when they want to implement custom partition selection logic.

To make this possible / prove that this is possible, it implements `build_schedule_from_partitioned_job` on top of `@schedule` and adds methods to JobDefinition and PartitionsDefinition to make the implementation simple.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/partitioned_schedule.py
+++ b/python_modules/dagster/dagster/_core/definitions/partitioned_schedule.py
@@ -1,23 +1,12 @@
-from datetime import time
 from typing import Optional, Union, cast
 
 import dagster._check as check
 
+from .decorators.schedule_decorator import schedule
 from .job_definition import JobDefinition
-from .partition import (
-    Partition,
-    PartitionSetDefinition,
-    PartitionedConfig,
-    ScheduleType,
-    get_cron_schedule,
-)
 from .run_request import SkipReason
-from .schedule_definition import (
-    DefaultScheduleStatus,
-    ScheduleDefinition,
-    ScheduleEvaluationContext,
-)
-from .time_window_partitions import TimeWindow, TimeWindowPartitionsDefinition
+from .schedule_definition import DefaultScheduleStatus, ScheduleDefinition
+from .time_window_partitions import TimeWindowPartitionsDefinition
 from .unresolved_asset_job_definition import UnresolvedAssetJobDefinition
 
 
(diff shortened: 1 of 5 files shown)


python_modules/dagster/dagster/_core/definitions/partitioned_schedule.py
```diff
@@ -40,78 +29,35 @@ def build_schedule_from_partitioned_job(
...
         job=job,
         default_status=default_status,
+        execution_timezone=partitions_def.timezone,
+        name=check.opt_str_param(name, "name", f"{job.name}_schedule"),
+        description=check.opt_str_param(description, "description"),
     )
+    def schedule_def(context):
+        partition_keys = partitions_def.get_partition_keys(context.scheduled_execution_time)
+        if len(partition_keys) == 0:
+            return SkipReason()
```
dpeng817, line 54: I feel like it's worth including a message here regarding why the skip is occurring.

python_modules/dagster/dagster/_core/definitions/partitioned_schedule.py
```diff
@@ -40,78 +29,35 @@ def build_schedule_from_partitioned_job(
...
 
-    return schedule_def
+        partition_key = partition_keys[-1]
 
+        yield job.run_request_for_partition(partition_key=partition_key, run_key=partition_key)
```
dpeng817, line 58: is there guaranteed to be just one partition key that we want to execute over? Should we be yielding run requests for each partition?

dpeng817 (approved):
Some comments and questions; assuming they don't bring anything super new to light fine with landing. I wonder if we should follow this up with some docs for more customized partition selection use cases / using the schedule decorator?

python_modules/dagster/dagster/_core/definitions/partitioned_schedule.py
dpeng817: either way good to include a comment here as it's a bit confusing

sryza: added

python_modules/dagster/dagster/_core/definitions/partitioned_schedule.py
sryza: that does seem like a nice thing to add

sryza: added

sryza:
> I wonder if we should follow this up with some docs for more customized partition selection use cases / using the schedule decorator?

I think that would be useful but doesn't feel like the highest priority thing for me right now

clairelin135 (approved):
This looks good to me. Definitely feels much more natural to fetch partitions definitions directly via `schedule.job.partitions_def` rather than from the schedule directly

PR #9322, opened 2022-08-09
Fix hydration mismatch due to invalid dom nesting

salazarm:
### Summary & Motivation

`p` tag is an invalid child of an `a` tag and this ends up tripping up react hydration when it calls `validateDOMNesting`. Changing to a `span` fixes the issue

Loaded the docs site and saw the big scary error dialog go away


erinkcochran87 (approved):
🙌

PR #9324, opened 2022-08-09
RFC: Allow the use of env vars to override the default io manager

gibsondan:
Summary:
Could imagine using this in different execution environments where filesystem as the default doesn't make sense. The calling environment sets the env var like in the test.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/job_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/job_definition.py
@@ -1,3 +1,6 @@
+import importlib
+import json
+import os
 from functools import update_wrapper
 from typing import (
     TYPE_CHECKING,
@@ -47,7 +50,7 @@ from dagster._core.selector.subset_selector import (
     OpSelectionData,
     parse_op_selection,
 )
-from dagster._core.storage.io_manager import io_manager
+from dagster._core.storage.io_manager import IOManagerDefinition, io_manager
 from dagster._core.utils import str_format_set
 from dagster._utils import merge_dicts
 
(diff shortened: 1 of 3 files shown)


sryza (changes requested):
Requesting changes for q management

mingfang:
+1
Env vars enables us to run the same code in different environments(dev,qa,prod, etc) without making any code changes.
We use env vars in all of our in house apps, so it would be great if we can manage dagster the same way.

sryza:
@mingfang have you seen the pattern in use in the fully-featured example: https://github.com/dagster-io/dagster/blob/master/examples/project_fully_featured/project_fully_featured/repository.py#L27? Would this work for you?

mingfang:
@sryza 
That's better than nothing I guess. 
But I much rather using something like `dotenv` and 
then move all configuration to either `.env` files and/or just-in-time env vars as part of my Kubernetes deployments.

python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -800,13 +803,47 @@ def get_direct_input_values_from_job(target: PipelineDefinition) -> Mapping[str,
...
-def default_job_io_manager(init_context: "InitResourceContext"):
+def _default_job_io_manager(init_context: "InitResourceContext"):
     from dagster._core.storage.fs_io_manager import PickledObjectFilesystemIOManager
 
     instance = check.not_none(init_context.instance)
     return PickledObjectFilesystemIOManager(base_dir=instance.storage_directory())
 
 
+def get_default_job_io_manager():
+    io_manager_env_var = os.getenv("DEFAULT_IO_MANAGER_DEFINITION_ATTRIBUTE")
```
jmsanders, line 814: I put together a kind of fun proof-of-concept for this type of approach while out:

https://github.com/jmsanders/depenv

The ergonomics potentially leave something to be desired although I like your idea in this PR of using a json string.

But in general, I agree with @sryza re: the slippery slope. It feels odd to me that you could set IO managers this way but not other resources. And I'm not quite sure I'm ready to fully advocate for setting everything this way without dogfooding the idea a bit more.

PR #9325, opened 2022-08-10
remove solid-related toys, rename graph-job-op toys to toys

dpeng817:
Title. Old toys are on old APIs, not surfaced anywhere.


python_modules/libraries/dagster-dask/dagster_dask_tests/hammer_job.py
```diff
@@ -0,0 +1,9 @@
+from dagster_dask import dask_executor
```
dpeng817, line 1: Dagster-dask test was using a pipeline created within the toys module; moved that here so that we don't have some wacky dagster-dask import within our toys repo.

sryza (approved):
+10000000

sryza:
I've been wanting to do this for so long

PR #9327, opened 2022-08-10
always_materialize on asset decorator

### LINKED ISSUE #8011: is_required on @asset

jamiedemaria:
### Summary & Motivation

adds `always_materialize` to `@asset` so that you can halt downstream execution/ only materialize an asset conditionally

TODO 
- [x] docs section 
- [x] test execution with sensors (ie asset sensor shouldn't trigger when no materialization)

resolves #8011

BK; added an asset to toys repo that sometimes materializes. Confirmed in dagit that a downstream asset doesn't run and the sensor isn't triggered when the asset doesn't materialize


docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -195,6 +195,25 @@ def downstream_asset():
 
 In this example, Dagster doesn’t need to load data from `upstream_asset` to successfully compute the `downstream_asset`. While `downstream_asset` does depend on `upstream_asset`, the key difference with `non_argument_deps` is that data isn’t being passed between the functions. Specifically, the data from the `sugary_cereals` table isn't being passed as an argument to `downstream_asset`.
 
+#### Conditional materialization
+
+In some cases, an asset may not need to be updated in storage each time the decorated function is executed. In this case you can use the `always_materialize` parameter along with `yield` syntax to only materialize an asset to storage in some cases. If an asset is not materialized to storage, downstream assets will not be materialized, and asset sensors monitoring the asset will not trigger
```
OwenKephart, line 200: I think that this can be more direct about what's happening under the hood. Something like "If the `always_materialize` parameter is set to False, you may indicate to the Dagster framework that no data should be persisted to storage by not yielding an output from your computation function. If an output is not emitted during computation, no asset materialization event will be created, the io manager will not be invoked, downstream assets will not be materialized, and asset sensors monitoring the asset will not trigger.

OwenKephart (commented):
Nice! Glad to have this feature in. Had some brief comments on the docs bit, and I also wanted to propose `output_required` as an alternate name for this parameter.

`always_materialize` seems slightly confusing, just because it isn't immediately clear to me what scope of situations "always" applies to. I also think `output_required` implies a bit more of what the user's workflow will look like (optionally emitting an output).

Open to other suggestions as well.

jamiedemaria:
@owen that's a good name idea! some others that were tossed around were:
is_conditional
always_output

 @sryza what do you think about `output_required`?

sryza:
@jamiedemaria I would defer to you on "always_materialize" vs. "output_required".  They've both got their pros and cons and one doesn't strike me as obviously better than the other.

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -195,6 +195,25 @@ def downstream_asset():
 
 In this example, Dagster doesn’t need to load data from `upstream_asset` to successfully compute the `downstream_asset`. While `downstream_asset` does depend on `upstream_asset`, the key difference with `non_argument_deps` is that data isn’t being passed between the functions. Specifically, the data from the `sugary_cereals` table isn't being passed as an argument to `downstream_asset`.
 
+#### Conditional materialization
```
sryza, line 198: Nitpick: I think this should go at the bottom of this section (i.e. after "graph-backed assets" etc., because I think it's likely to be used less often than the other functionality)

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -195,6 +195,25 @@ def downstream_asset():
...
+
+In some cases, an asset may not need to be updated in storage each time the decorated function is executed. In this case you can use the `always_materialize` parameter along with `yield` syntax to invoke this behavior. If the `always_materialize` parameter is set to `False`, you may indicate to the Dagster framework that no data should be persisted to storage by not yielding an output from your computation function. If an output is not emitted during computation, no asset materialization event will be created, the I/O manager will not be invoked, downstream assets will not be materialized, and asset sensors monitoring the asset will not trigger.
```
sryza, line 200: This is a great explanation

sryza:
Here's a thread where there was some confusion about the difference between `is_required` and `can_subset`: https://dagster.slack.com/archives/C01U954MEER/p1660126092122879.

If one of the names reduces the ambiguity here, that could be a point in its favor.

sryza (approved):

PR #9328, opened 2022-08-10
Incompatible variable type fixed

luca-digrazia:
**"filename"**: "python_modules/dagster/dagster/_core/definitions/decorators/op_decorator.py"
**"warning_type"**: "Incompatible variable type [9]"
**"warning_message"**: " outs is declared to have type `Optional[Mapping[str, Out]]` but is used as type `Dict[str, Union[Mapping[str, Out], Out]]`."
**"warning_line"**:63
**"fix"**: self.out is None

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/op_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/op_decorator.py
@@ -61,7 +61,7 @@ class _Op:
         )
 
         outs: Optional[Mapping[str, Out]] = None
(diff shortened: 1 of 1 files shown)


clairelin135:
Hi @luca-digrazia. I think we might actually want to keep the `not None` check as this ensures that directly providing an `Out` object will propagate this to the underlying `OpDefinition`.

Where were you seeing this warning?

cmpadden:
Closing due to inactivity; please re-open if necessary, thank you

PR #9329, opened 2022-08-10
move healthcheck before step event processing loop

MannySchneck:
### Summary & Motivation

On intial pull of master, the test
`python_modules/dagster/dagster_tests/execution_tests/engine_tests/test_step_delegating_executor.py::test_execute_intervals`
was failing.

Debugging showed that this test failed because we were processing the steps before
the health checks had a chance to run. This change moves the health check before the
step status processing loop, so that every step is health-checked at least
once, even if it has submitted a completion event.

This change repairs a test failure on the current `master` branch.

`pytest
python_modules/dagster/dagster_tests/execution_tests/engine_tests/test_step_delegating_executor.py`
succeeds. I have not re-run the full test suite.

### DIFF
--- a/python_modules/dagster/dagster/_core/executor/step_delegating/step_delegating_executor.py
+++ b/python_modules/dagster/dagster/_core/executor/step_delegating/step_delegating_executor.py
@@ -204,28 +204,6 @@ class StepDelegatingExecutor(Executor):
 
                     return
 
-                for dagster_event in self._pop_events(
-                    plan_context.instance,
-                    plan_context.run_id,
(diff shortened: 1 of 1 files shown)


MannySchneck:
The remaining failing test looks like a curation/decision issue about what belongs in the examples folder:
```
python_modules/dagster/dagster_tests/cli_tests/test_project_commands.py::test_available_examples_in_sync_with_example_folder
```

clairelin135:
@johannkm looks like you might have done some work on this in the past. Mind taking a look at this?

PR #9339, opened 2022-08-11
multi asset sensor

jamiedemaria:
introduces `multi_asset_sensor` (100% open to name changes) that does the following:
* Can monitor multiple assets 
* provides a context with functions that can fetch the materialization events for the asset, allowing the sensor to yield run requests based on the materializations (ie all monitored assets have materialized, a subset have materialized and others have not etc)

TODO:
- [x] update docs once API is locked down

BK and new assets and sensors in toys repo


jamiedemaria:
@sryza i'm a bit iffy on some of the names I've chosen in this PR. let me know what you think about:
`asset_status_sensor` - I'm worried this doesn't capture the multi asset capabilities of the sensor
`trigger_fn` - this name isn't very descriptive of what this param does (determine if the sensor should run based on the events)

sryza (commented):
Very excited for this! I've got thoughts on a few different pieces:

### Asset status

I think the "asset status" idea is interesting, and I could see it working well once we add some additional concepts around asset status, but I don't think a dagster event type is closely analogous enough to a dagster run status for this to make sense right now:
- An event type says "this thing happened"
- A run status says "this run is now in this particular state"

Let me know if there are particular use cases you have in mind that we'd miss out on without the status functionality?

### User-provided functions

Thoughts on combining the trigger_fn and the evaluation function? I.e. instead of this:

```
@asset_status_sensor(
    asset_keys=[AssetKey("asset_c"), AssetKey("asset_d")],
    asset_status=DagsterEventType.ASSET_MATERIALIZATION,
    trigger_fn=lambda x: all(x.values()),
    job=my_job,
)
def asset_c_or_d_sensor(context, _asset_events):
    return RunRequest(run_key=context.cursor)
```

this: 

```
@asset_status_sensor(
    asset_keys=[AssetKey("asset_c"), AssetKey("asset_d")],
    asset_status=DagsterEventType.ASSET_MATERIALIZATION,
    trigger_fn=lambda x: all(x.values()),
    job=my_job,
)
def asset_c_or_d_sensor(context, _asset_events):
    if all(context.events_by_monitored_asset_key.values()):
        return RunRequest(run_key=context.cursor)
```

A few reasons this could be beneficial:
- Supplying functions to other functions make code kind of awkward.  E.g. if the triggering logic is complex, it might not fit into a lambda, which means the triggering code needs to be located in a module-scope block that's separate from the rest of the sensor definition.
- The RunRequests might depend on the same data as the trigger function.
- The decision of whether to trigger might depend on other data inside the context, like the cursor or other events on the instance.

jamiedemaria:
@sryza 

Asset status - I agree that it seems unlikely that statuses other than `asset materialization` would be used. When i was thinking about implementation i had it in my head that we needed to incorporate the status, but while writing code it seemed less useful. if more statuses get added though, i could see it coming in handy later (and could be added then) 

incorporating trigger function - I like that as well, it also allows you to switch on more complex groupings (if A and B materialize do X, and if B and C do Y, if A and C do Z) rather than having three sensors

jamiedemaria:
spent some time removing the trigger function param and ran into a question:

What is the correct behavior when a sensor yields a SkipReason? Looking at this sensor:
```
@multi_asset_sensor(
    asset_keys=[AssetKey("asset_a"), AssetKey("asset_b")],
    job=log_asset_sensor_job,
)
def asset_a_and_b_sensor(context, asset_events):
    if all(asset_events.values()):
        return RunRequest(
            run_key=context.cursor,
        )
    else:
        return None
```
I think there's a reasonable argument that it should be
```
@multi_asset_sensor(
    asset_keys=[AssetKey("asset_a"), AssetKey("asset_b")],
    job=log_asset_sensor_job,
)
def asset_a_and_b_sensor(context, asset_events):
    if all(asset_events.values()):
        return RunRequest(
            run_key=context.cursor,
        )
    else:
        return SkipReason("condition not met")
```

option 1: returning RunRequest, SkipReason, and None do three different things. RunRequest when the condition is met, None when the condition isn't met, and SkipReason for when the condition is met, but we don't want to run a job for some other reason AND want to update the cursor. 

option 2: returning SkipReason and None do the exact same thing and mean that the condition hasn't been met (therefore cursor will not be updated)

python_modules/dagster-test/dagster_test_tests/test_generators.py
```diff
@@ -0,0 +1,23 @@
+import inspect
```
jamiedemaria, line 1: delete this file before merge

sryza:
>What is the correct behavior when a sensor yields a SkipReason? Looking at this sensor:

Let me know if this isn't a helpful answer, but I think one should basically yield a SkipReason whenever they want a message to show up explaining why the ship happened. Is the issue here that part of the reason for skipping might end up coming from the code we write and another part of it might come from the user?

jamiedemaria:
i don't think there's an issue of skip reasons from our code vs user code. I think the main thing is just determining which return types cause a cursor update. But sounds like you lean that None and SkipReason should be the same

sryza:
> But sounds like you lean that None and SkipReason should be the same. 

Yeah

jamiedemaria:
I'm back to thinking that `SkipRequest` might need to mean something other than `None`.


consider the following sensor 
```
@multi_asset_sensor(
    asset_keys=[AssetKey("asset_c"), AssetKey("asset_d")],
    job=my_job,
)
def asset_c_or_d_sensor(context, asset_events):
    # this sensor will run when only asset_c and not asset_d have been materialized
    if asset_events["asset_c"] is not None and asset_events["asset_d"] is None:
        return RunRequest(run_key=context.cursor)
    else:
        # you can optionally return a SkipReason
        return SkipReason("condition not met")
```
before we run anything our cursor is `{"asset_c": None, "asset_d": None}`
if we materialize `asset_c` then the sensor runs and the cursor gets updated to `{"asset_c": 1, "asset_d": None}`
Now we materialize `asset_d` and then materialize `asset_c` again. In the sensor, there will be event logs for both `asset_c` and `asset_d` so we will skip execution and the cursor will stay `{"asset_c": 1, "asset_d": None}`

No matter what we do from now, the sensor will never launch a RunRequest because every time the sensor executes there will be events for `asset_d` (since the cursor stays at None)

So now I'm back to thinking that maybe `SkipReason` should update the cursor and `None` should not update the cursor (this is needed for the AND case to work), or we should have some other way to specify that the cursor should be updated (like an arg on SkipReason)

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -661,3 +668,115 @@ def _fn(context):
...
+        jobs (Optional[Sequence[Union[GraphDefinition, JobDefinition, UnresolvedAssetJobDefinition]]]):
+            (experimental) A list of jobs to be executed when the sensor fires.
+        default_status (DefaultSensorStatus): Whether the sensor starts as running or not. The default
+            status can be overridden from Dagit or via the GraphQL API.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        asset_keys: List[AssetKey],
```
sryza, line 701: Nitpick: prefer Sequence and Mapping over List and Dict

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -661,3 +668,115 @@ def _fn(context):
...
+                        cursor_update_dict[a.to_user_string()] = cursor_dict.get(a.to_user_string())
+
+                result = materialization_fn(context, asset_event_records)
+                if result is None:
+                    return
+
+                # because the materialization_fn can yield results (see _wrapped_fn in multi_asset_sensor decorator),
+                # even if you return None in a sensor, it will still cause in inspect.isgenerator(result) == True.
+                # So start a yield_count to see if we actually return any values and should update the cursor
+                yield_count = 0
```
sryza, line 749: nitpick: it might be clearer to call this run_request_count? also, if we don't use the number for anything, it could just be  boolean?

sryza (commented):
This is looking close.

@prha brought up that fetching all the asset events could be expensive in some situations. E.g. imagine an AND sensor that depends on two assets: an asset that gets updated every minute and an asset that gets updated every month. By the end of the month, every tick would be pulling out 24 * 30 * 60 materializations from the DB.

A way to avoid this issue would be to create a MultiAssetSensorContext and replace the "asset_events" argument with methods on the context. Different methods could return the latest materialization or all materializations, depending on what the user cares about.  Thoughts?

Lastly, a few cosmetic things to think about with the "asset_events" name:
- If feasible, we should find a way to name things so that it's clear to users whether the dictionary includes keys for assets with no events
- Should it be `asset_materializations`?

jamiedemaria (commented):
@sryza @prha, I did a very rough pass at making a context specific for multi asset sensors that would allow users to fetch events in the sensor body more easily. Not sure it fully works yet, but I want to get some initial feedback on the user facing API to decide if this makes sense as the way to go. If we do go this direction, i'll update the rest of the code and clean it up. 

the sensors in `toys/asset_sensors.py` have been updated to use the new version so that's a good file to look at

i left a couple comments in the PR on specific places where I have questions/thoughts

python_modules/dagster-test/dagster_test/toys/asset_sensors.py
```diff
@@ -0,0 +1,119 @@
...
+        return RunRequest(
+            run_key=context.cursor,
+            run_config={
+                "ops": {
+                    "log_asset_sensor": {"config": {"message": f"Asset events dict {asset_events}"}}
+                }
+            },
+        )
+    else:
+        return None
```
prha, line 70: Can we discuss a little more around having `return SkipReason(...)` vs `return None` be significant for cursor updating logic?

I have some reservations (especially since this will be new behavior that only applies to this `@multi_asset_sensor`) so want to make sure we explore the full set of potential solutions.

For user-defined sensors, the cursor management logic is maintained by the user themselves, so it is reasonable that they call `context.update_cursor`.  For `@asset_sensor`, we fully manage the cursor management, so the user code doesn't have to do anything.  For these `@multi_asset_sensor`, I can completely see wanting to have some simplified cursor management, but maybe with a cleaner API.

Because we're creating a new context type, I can see us having some custom cursor management.  What do you think about something like `context.advance_cursor(AssetKey("asset_a"))` and `context.advance_all_cursors()` or something similar?  

They might need to work in concert with the event methods (e.g. `context.get_latest_event` / `context.get_all_events`).  For example, let's say that there have been 1000 materializations since the last cursor for a given sensor.  I could see wanting to do some amount of work for every single materialization (maybe there's a backlog), but I am fine with only processing 25 in this sensor evaluation.   Can I just get the earliest 25 materializations (without fetching all from the DB), advance the cursor for just that amount, and wait for the next sensor tick to process the remaining 975 materializations?

I think in order for this to work, `get_all_events` might need to return a generator so that under the hood we  can do batch fetching from the DB.

prha (commented):
This is amazing work... Really excited for this and I'm glad you're thinking through these issues.

Mostly, I have questions around the cursor management and the implications of the current API and the types of DB query load we're going to enable (the two are very interrelated).  Just want to make sure that the first time we turn on a multi-asset-sensor for a long-existing asset key, we don't cause timeouts, have common gotchas for users.

python_modules/dagster-test/dagster_test/toys/asset_sensors.py
sryza: I thought where we ended up was that `return SkipReason(...)` vs `return None` would not be significant for cursor-updating logic, but that we'd punt on adding something like `context.advance_cursor` until later.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -153,6 +155,107 @@ def update_cursor(self, cursor: Optional[str]) -> None:
...
+
+        from dagster import sensor, SensorEvaluationContext
+
+        @sensor
+        def the_sensor(context: SensorEvaluationContext):
+            ...
+
+    """
+
+    def get_latest_events(self, asset_keys: Sequence[AssetKey]):
```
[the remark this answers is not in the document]
sryza: I think a name like `latest_materializations_by_key` would better help users understand what this is going to return. Also, I would expect the keys of the dictionary to be AssetKeys instead of strings.

python_modules/dagster-test/dagster_test/toys/asset_sensors.py
jamiedemaria: sandy's correct, the current implementation treats `SkipReason` and `None` the same (and doesn't update the cursor). The idea is to add the base functionality now, and then a follow up PR for allowing users to update the cursor themselves like you mention.

maybe for `advance_cursor` you can pass the asset key and the event record and that will update the cursor to the storage_id in the event record. so then you could process the 25 runs and update the cursor on the last one without needing to know to dig into the event record and get `storage_id`

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -153,6 +155,107 @@ def update_cursor(self, cursor: Optional[str]) -> None:
...
+        cursor_dict = json.loads(self.cursor) if self.cursor else {}
+        asset_event_records = {}
+        for a in asset_keys:
+            event_records = self.instance.get_event_records(
+                EventRecordsFilter(
+                    event_type=DagsterEventType.ASSET_MATERIALIZATION,
+                    asset_key=a,
+                    after_cursor=cursor_dict.get(a.to_user_string()),
+                ),
+                ascending=False,
```
jamiedemaria, line 225: flip to True so that the latest events are first (to support if we want to process for each event)

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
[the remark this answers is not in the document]
sryza: Also, can we use the asset keys from the sensor for this instead of asking the user to provide them? Requiring users to provide the asset keys would solve the ambiguity I brought up in my last review about which keys are included, but IMO it's not worth the additional boilerplate for users.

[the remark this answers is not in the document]
jamiedemaria: i like that name, and yeah the keys being asset keys makes more sense.  

I'm thinking the function could take an optional list of keys and if none are provided it'll default to the list provided to the sensor (just didn't want to implement rn in case we scrapped the new approach)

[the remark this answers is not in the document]
sryza: Accepting an optional list of keys sounds sensible to me

python_modules/dagster-test/dagster_test/toys/asset_sensors.py
jamiedemaria: thinking about cursors strikes again.... In the case phil brought up where we have a sensor that is working through a backlog of 1000 asset materializations and launching jobs for 25 materializations per evaluation, we only want to update the cursor to those 25 materializations. But right now if you return any RunRequest the cursor gets updated to the latest materialization. So I think the `advance_cursor` functions need to be included in this PR, and we'll need to make it so that every sensor must manually handle the cursor updating logic.

alternately we could have a bool attribute on the context like `cursor_has_been_updated` that gets flipped to false at the beginning of the evaluation fn. any call to `advance_cursor` flips `cursor_has_been_updated` to true. Then after we have run the user's decorated function, we can check if a run request has been returned AND `cursor_has_been_updated` is false, and if so we can advance the cursor all the way to the latest materializations

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
sryza: I think I would expect the latest events to come last

python_modules/dagster-test/dagster_test/toys/asset_sensors.py
sryza: Mmm - interesting - that makes sense.
 
I don't think it's terrible to expect users to call `advance_cursor` all the time- although I think we should provide some default or other method which makes it really easy to advance the cursor past all the materializations.

If we go this direction, I think maybe we should raise an error if the cursor doesn't get advanced at all, to avoid a that common failure mode.

Also, I like the idea of passing the event record so that users don't need to think about storage IDs.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
jamiedemaria: yep, typo in my comment oops

python_modules/dagster/dagster/_core/definitions/decorators/sensor_decorator.py
```diff
@@ -156,3 +158,83 @@ def _wrapped_fn(context, event):
...
+            if inspect.isgenerator(result) or isinstance(result, list):
+                for item in result:
+                    yield item
+            elif isinstance(result, (RunRequest, SkipReason)):
+                yield result
+
+            elif result is not None:
+                raise DagsterInvariantViolationError(
+                    (
+                        "Error in sensor {sensor_name}: Sensor unexpectedly returned output "
```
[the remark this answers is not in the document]
jamiedemaria: maybe it will? this is in the asset_sensor too so i just copied it. I'll check if it's necessary

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -154,6 +159,120 @@ def repository_name(self) -> Optional[str]:
         return self._repository_name
 
 
+class MultiAssetSensorEvaluationContext(SensorEvaluationContext):
+    """The context object available as the argument to the evaluation function of a :py:class:`dagster.MultiAssetSensorDefinition`.
```
jamiedemaria, line 163: An issue I'm having with this new context is that the `SensorEvaluationContext` expects the cursor to be a string, but in this multi asset case, the cursor is a dict. it's screwing up a bunch of mypy checks and our internal `check` calls. I \~think\~ it makes sense to keep `MultiAssetSensorEvaluationContext` as a subclass of `SensorEvaluationContext` so that changes to the shared functions only need to be made in one place. One option would be to make the type for cursor in `SensorEvaluationContext` a `Union[str, Mapping[AssetKey, str]]` but i'm not sure if that would screw up backcompat or something. thoughts?

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -154,6 +159,120 @@ def repository_name(self) -> Optional[str]:
...
+                event_type=DagsterEventType.ASSET_MATERIALIZATION,
+                asset_key=asset_key,
+                after_cursor=cursor_dict.get(asset_key.to_user_string()),
+            ),
+            ascending=True,
+            limit=limit,
+        )
+
+    def advance_cursor(
+        self, materializations_by_key
```
jamiedemaria, line 261: When I try to add type annotations here, it should be `Mapping[AssetKey, EventLogRecord]` but that causes an import error because importing `EventLogRecord` creates a circular import. What are the options for getting around that?

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -477,7 +596,7 @@ def __new__(
     ):
         check.opt_list_param(run_requests, "run_requests", RunRequest)
         check.opt_str_param(skip_message, "skip_message")
-        check.opt_str_param(cursor, "cursor")
+        # check.opt_str_param(cursor, "cursor")
```
jamiedemaria, line 599: add this check back once i figure out how to handle the `mapping[assetkey, str]` type for the new context

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -557,14 +676,61 @@ def build_sensor_context(
...
+        asset_keys (Sequence[AssetKey]): The list of asset keys monitored by the sensor
+        instance (Optional[DagsterInstance]): The dagster instance configured to run the sensor.
+        cursor (Optional[Mapping[AssetKey, str]]): A dictionary of cursor values to provide to the evaluation of the sensor.
+        repository_name (Optional[str]): The name of the repository that the sensor belongs to.
+
+    Examples:
+
+        .. code-block:: python
+
+            context = build_multi_asset_sensor_context()
```
jamiedemaria, line 701: update this example to pass instance and asset keys since those are required for this to work

python_modules/dagster/dagster_tests/core_tests/definitions_tests/test_sensor_invocation.py
```diff
@@ -253,3 +258,35 @@ def basic_sensor_w_arg(context):
         return RunRequest(run_key=None, run_config={}, tags={})
 
     assert basic_sensor_w_arg(context).run_config == {}
+
+
+def test_multi_asset_sensor_no_cursor_update():
```
jamiedemaria, line 263: rename

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -154,6 +159,120 @@ def repository_name(self) -> Optional[str]:
...
+            sensor. Use the preferred `cursor` attribute instead.
+        repository_name (Optional[str]): The name of the repository that the sensor belongs to.
+        instance (Optional[DagsterInstance]): The deserialized instance can also be passed in
+            directly (primarily useful in testing contexts).
+
+    Example:
+
+    .. code-block:: python
+
+        from dagster import sensor, SensorEvaluationContext
```
jamiedemaria, line 184: update this example with the correct types

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -154,6 +159,120 @@ def repository_name(self) -> Optional[str]:
...
+        last_run_key: Optional[str],
+        cursor: Optional[Mapping[AssetKey, str]],
+        repository_name: Optional[str],
+        asset_keys: Sequence[AssetKey],
+        instance: Optional[DagsterInstance] = None,
+    ):
+        self._asset_keys = asset_keys
+        self.cursor_has_been_updated = False
+
+        super(MultiAssetSensorEvaluationContext, self).__init__(None, None, None, None, None, None)
```
jamiedemaria, line 205: This is a bit hacky right now to avoid the error with the cursor typing. Once that gets figured out, this will be cleaned up

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -154,6 +159,120 @@ def repository_name(self) -> Optional[str]:
...
+            asset_keys = self._asset_keys
+
+        cursor_dict = json.loads(self.cursor) if self.cursor else {}
+        asset_event_records = {}
+        for a in asset_keys:
+            event_records = self.instance.get_event_records(
+                EventRecordsFilter(
+                    event_type=DagsterEventType.ASSET_MATERIALIZATION,
+                    asset_key=a,
+                    after_cursor=cursor_dict.get(a.to_user_string()),
```
jamiedemaria, line 231: change all cursor logic to use AssetKey as keys

jamiedemaria:
@sryza @prha i think this is getting close. I have a couple annoying type checking things i could use some help figuring out (see comments i recently left), but otherwise i think this is good for another review round. After we feel good about the new API setup, i'll take another pass at the docs and update them to talk about cursor updating etc.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -154,6 +159,120 @@ def repository_name(self) -> Optional[str]:
...
+        self._instance_ref = check.opt_inst_param(instance_ref, "instance_ref", InstanceRef)
+        self._last_completion_time = check.opt_float_param(
+            last_completion_time, "last_completion_time"
+        )
+        self._last_run_key = check.opt_str_param(last_run_key, "last_run_key")
+        self._cursor = check.opt_dict_param(cursor, "cursor", key_type=AssetKey, value_type=str)
+        self._repository_name = check.opt_str_param(repository_name, "repository_name")
+        self._instance = check.opt_inst_param(instance, "instance", DagsterInstance)
+
+    def latest_materializations_by_key(self, asset_keys: Optional[Sequence[AssetKey]] = None):
```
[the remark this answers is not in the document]
sryza: When I suggested the name `latest_materializations_by_key`, I was thinking that the values of the returned dictionary would be `AssetMaterialization` objects. I see the wisdom in returning `EventLogRecord` objects though, because then users can get the storage ID and timestamp.  Maybe a name like `latest_materialization_records_by_key` would be clearer in that case?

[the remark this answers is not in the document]
sryza: Analogous naming feedback applies to the `materializations_for_key` method and the `materializations_by_key` arg to `advance_cursor`.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
sryza: Given that the cursors function differently, I think making it so that `MultiAssetSensorEvaluationContext` doesn't inherit from `SensorEvaluationContext` might be the most sensible path.

sryza (commented):
Looking very close! Had a couple nitpicks on naming.

examples/docs_snippets/docs_snippets/concepts/partitions_schedules_sensors/sensors/sensors.py
```diff
@@ -239,8 +247,49 @@ def multi_asset_sensor(context):
...
+# start_multi_asset_sensor_custom_fn_marker
+
+
+@multi_asset_sensor(
+    asset_keys=[AssetKey("asset_c"), AssetKey("asset_d")],
+    job=my_job,
+)
+def asset_c_or_d_sensor(context, asset_events):
+    # this sensor will run when either asset_c or asset_d materialize
+    if any(asset_events.value()):
```
prha, line 284: should this be `.values`?  maybe put these under test?

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -154,6 +159,120 @@ def repository_name(self) -> Optional[str]:
...
+            )
+
+            if event_records:
+                asset_event_records[a] = event_records[0]
+            else:
+                asset_event_records[a] = None
+
+        return asset_event_records
+
+    def materializations_for_key(self, asset_key: AssetKey, limit: int):
```
prha, line 244: Doesn't need to be this diff, but we potentially will want to expose `ascending` as an argument.  This might affect the `advance_cursor` API a little so at least worth thinking through a couple cases.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
jamiedemaria: oh cool, i didn't realize those two things went together, was trying to figure out why the string type annotations didn't work since i see them everywhere. thanks!

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
jamiedemaria: I don't think exposing `ascending` would cause any problems because the user has to provide a single EventLogRecord to `advance_cursor`, so they will have to make a dictionary where they pick the record they want the cursor to advance to. I adding a param for `ascending` because we might as well do it now

examples/docs_snippets/docs_snippets/concepts/partitions_schedules_sensors/sensors/sensors.py
jamiedemaria: yep good catch. i'll put them all under test after i revamp the docs with the new changes

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -154,6 +160,205 @@ def repository_name(self) -> Optional[str]:
...
+    @public
+    def materialization_records_for_key(
+        self, asset_key: AssetKey, limit: int, ascending: bool = True
+    ) -> Iterable["EventLogRecord"]:
+        """Fetches asset materialization event records for asset_key
+
+        Args:
+            asset_key (AssetKey): The asset to fetch materialization events for
+            limit (int): The number of events to fetch
+            ascending (bool): If True, materialization events will be returned with the earliest event first. If False, the most recent
```
prha, line 319: I think the cursor logic might behave in a surprising way if we allow for `ascending=False`, because when we pass the event into `advance_cursor`, it will filter to events AFTER that cursor, which we've already examined.  I'm actually fine removing the `ascending` parameter so we don't have to deal with it, but it's worth thinking through if our cursor API has the right set of parameters also.

examples/docs_snippets/docs_snippets/concepts/partitions_schedules_sensors/sensors/sensors.py
```diff
@@ -190,17 +197,18 @@ def my_asset_sensor(context, asset_event):
...
 
-# start_multi_asset_sensor_marker
+# start_custom_multi_asset_sensor_marker
 import json
 from dagster import EventRecordsFilter, DagsterEventType
 
 
 @sensor(job=my_job)
-def multi_asset_sensor(context):
+def custom_multi_asset_sensor(context):
```
sryza, line 206: I think we can remove this

python_modules/dagster/dagster/__init__.py
```diff
@@ -296,10 +297,13 @@
 )
 from dagster._core.definitions.sensor_definition import (
     AssetSensorDefinition as AssetSensorDefinition,
+    MultiAssetSensorDefinition as MultiAssetSensorDefinition,
```
sryza, line 300: alphabetical order plz

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -154,6 +160,205 @@ def repository_name(self) -> Optional[str]:
...
+        return self._last_run_key
+
+    @public  # type: ignore
+    @property
+    def cursor(self) -> Optional[Mapping[AssetKey, Optional[int]]]:
+        """The cursor value for this sensor, which was set in an earlier sensor evaluation."""
+        return self._cursor
+
+    @public
+    def update_cursor(self, cursor: Optional[Mapping[AssetKey, Optional[int]]]) -> None:
```
sryza, line 254: It seems a little odd to have both this and `advance_cursor`, given that they work very similarly and their names don't reflect the differences in how they work

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
jamiedemaria: oh i see what you mean. I think we'd need to expose the `before_cursor` arg on the `EventsFilter` too and keep track of two cursors per assetkey. I think that would be a case where it would be reasonable to tell the user to write a plain `sensor` that does the event fetching manually

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
jamiedemaria: unresolving this because the cursor in this context is back to being stored as a string and it made more sense to have the context inherit from `SensorEvaluationContext` because of mypy. I was thinking i could update the two cursor functions to `update_cursor` and `update_all_cursors`, but overriding `update_cursor` with a new param name caused other type checking errors (and i thought was frowned upon in python but i could def be wrong about that)

So right now the state of the MultiAssetSensorEvaluationContext is:
`update_cursor`
`advance_cursor`
`advance_all_cursors`

I could override the body of `update_cursor` to throw an error, but that doesn't feel like a great solution either

python_modules/dagster/dagster/_grpc/impl.py
```diff
@@ -283,24 +287,45 @@ def get_external_sensor_execution(
...
-                lambda: "Error occurred during the execution of evaluation_fn for sensor "
-                "{sensor_name}".format(sensor_name=sensor_def.name),
-            ):
-                return sensor_def.evaluate_tick(sensor_context)
-        except Exception:
-            return ExternalSensorExecutionErrorData(
-                serializable_error_info_from_exc_info(sys.exc_info())
-            )
+    if isinstance(sensor_def, MultiAssetSensorDefinition):
+        with MultiAssetSensorEvaluationContext(
```
prha, line 291: nit: could get rid of the code dupe by using exitstack to handle the conditional contexts...

prha (commented):
This looks good to me!

If we wanted to be more conservative (we think that the API might change in a breaking way), we might want to consider marking experimental while we get some users to try out?

python_modules/dagster/dagster/_grpc/impl.py
jamiedemaria: could you explain a bit more? i'm not quite following

prha: You could do something like this:

```
with ExitStack() as stack:
    if isinstance(sensor_def, MultiAssetSensorDefinition):
        sensor_context = stack.enter_context(MultiAssetSensorEvaluationContext(...))
    else:
        sensor_context = stack.enter_context(SensorEvaluationContext(...))
    try:
        with user_code_error_boundary(...):
            return sensor_def.evaluate_tick(sensor_context)
    except Exception:
        return ExternalSensorExecutionErrorData(...)
```

jamiedemaria: oh cool! i've never seen that before, that's super useful

sryza (approved):
Woohoo!

PR #9341, opened 2022-08-11
Provide install_default_libraries for databricks_pyspark_step_launcher run_config

nvinhphuc:
### Summary & Motivation
We are using dagster_databricks. When dagster upgrades to a new version, the job submit to Databricks will fail because Databricks cluster cannot automatically install new version without delete older version.

In production environment, we cannot just go to Databricks cluster and delete the old libraries. The default libraries in this case make this difficult to handle the situation. Thus, this PR is to provide the options to run Databricks jobs without libraries by adding a install_default_libraries params in run_config.

Unit test + Local test

### DIFF
--- a/python_modules/libraries/dagster-databricks/dagster_databricks/configs.py
+++ b/python_modules/libraries/dagster-databricks/dagster_databricks/configs.py
@@ -502,10 +502,16 @@ def _define_submit_run_fields():
         "This token should have at most 64 characters.",
         is_required=False,
     )
+    install_default_libraries = Field(
+        Bool,
+        description="To install default libraries or not.",
+        is_required=False,
+    )
     return {
         "cluster": _define_cluster(),
         "run_name": run_name,
         "libraries": _define_libraries(),
+        "install_default_libraries": install_default_libraries,
         "timeout_seconds": timeout_seconds,
         "idempotency_token": idempotency_token,
     }
--- a/python_modules/libraries/dagster-databricks/dagster_databricks/configs.py
+++ b/python_modules/libraries/dagster-databricks/dagster_databricks/configs.py
@@ -504,7 +504,9 @@ def _define_submit_run_fields():
     )
     install_default_libraries = Field(
         Bool,
-        description="To install default libraries or not.",
+        description="By default, Dagster installs a version of dagster, dagster-databricks, and "
+        "dagster-pyspark matching the locally-installed versions of those libraries. If you would "
+        "like to disable this behavior, this value can be set to False.",
         is_required=False,
     )
     return {
--- a/python_modules/libraries/dagster-databricks/dagster_databricks/databricks.py
+++ b/python_modules/libraries/dagster-databricks/dagster_databricks/databricks.py
@@ -173,14 +173,16 @@ class DatabricksJobRunner:
         # since they're imported by our scripts.
         # Add them if they're not already added by users in config.
         libraries = list(run_config.get("libraries", []))
-        python_libraries = {
-            x["pypi"]["package"].split("==")[0].replace("_", "-") for x in libraries if "pypi" in x
-        }
-        for library in ["dagster", "dagster-databricks", "dagster-pyspark"]:
-            if library not in python_libraries:
-                libraries.append(
-                    {"pypi": {"package": "{}=={}".format(library, dagster.__version__)}}
-                )
+        install_default_libraries = run_config.get("install_default_libraries", True)
+        if install_default_libraries == True:
+            python_libraries = {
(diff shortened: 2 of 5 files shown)


OwenKephart:
@nvinhphuc Thanks for the contribution! Having these libraries included by default helps a lot for the onboarding experience (as the listed libraries are essential to have installed), so I'd prefer not to remove them if possible.

When does this failure occur, exactly? Is it after you upgrade your local version of Dagster, at which point launching a step attempts to install that newer version of Dagster on the Databricks cluster? If so, could you resolve this by including config that pins those three libraries to the older version of Dagster?

nvinhphuc:
Hi @OwenKephart , thank you for reviewing my PR. By pinning those libraries can make it works temporarily, but in the perspective of building a CI/CD around that, it's not a very good way, because after all you have to delete the older libraries and install new libraries. We manage our Databricks cluster in Terraform, so having to install dependencies of Databricks in another source is not a good choice.

I appreciate the thought of you to support beginners for a more comfortable onboard experiences, if the file doesn't contain any other libraries, but in my opinion, it will add more complexity later. Because you have to include all the libraries in the file containing Databricks ops, not just those three libraries. So sooner or later users will encounter missing library errors and realize they have to install other libraries as well. I think this may lead to more confusion. In fact, I have to dive in the code to see what happened.

In the other hands, I think it is reasonable having an option to executing Databricks ops without any libraries. If you don't agree with my approach in this PR. Maybe you can suggest another way to support this feature? I'm willing to support you to do that.

OwenKephart:
Ah I see, that makes sense (sorry for the slow response). Maybe an alternative solution here would be to have a separate config option for "install_default_libraries", which defaults to True, but if set to False, would disable this behavior. This way, for cases like yours, you have that escape hatch, but we retain the same default behavior.

You do raise a great point about people running into more complexity down the line (maybe this could be alleviated with better documentation -- the docs for this library are pretty non-existent), but I wouldn't feel comfortable changing the current behavior until this library gets bumped to 1.x, because there are plenty of people currently relying on the current behavior, so this would be a pretty rough breaking change.

Another thing I will note is that there's no guarantee of version compatibility between these libraries (even between patch releases), so it is somewhat risky to allow the dagster version on the databricks cluster to get out of sync with the version of dagster on the host process. In most releases, this won't be an issue, but it's something to keep in mind.

nvinhphuc:
Hi @OwenKephart , I update the PR using install_default_libraries config. Could you review it for me? Thank you.

python_modules/libraries/dagster-databricks/dagster_databricks/configs.py
```diff
@@ -502,10 +502,16 @@ def _define_submit_run_fields():
         "This token should have at most 64 characters.",
         is_required=False,
     )
+    install_default_libraries = Field(
+        Bool,
+        description="To install default libraries or not.",
```
OwenKephart, line 507: "By default, Dagster installs a version of dagster, dagster-databricks, and dagster-pyspark matching the locally-installed versions of those libraries. If you would like to disable this behavior, this value can be set to False."

python_modules/libraries/dagster-databricks/dagster_databricks/databricks.py
```diff
@@ -173,14 +173,18 @@ def submit_run(self, run_config, task):
...
+        if install_default_libraries:
+            python_libraries = {
+                x["pypi"]["package"].split("==")[0].replace("_", "-")
+                for x in libraries
+                if "pypi" in x
+            }
+            for library in ["dagster", "dagster-databricks", "dagster-pyspark"]:
+                if library not in python_libraries:
+                    libraries.append(
+                        {"pypi": {"package": "{}=={}".format(library, dagster.__version__)}}
```
OwenKephart, line 186: I just realized that, as of dagster 1.0.x, this will not work, as the versions of `dagster-databricks` and `dagster-pyspark` are version 0.16.x, which is not the same as the dagster version.

I can handle this in a follow-up PR, but I just wanted to make sure that this wasn't causing issues for you currently.

OwenKephart (commented):
Just got back from vacation -- this looks good! I left a comment for updating the wording on the config, and had a question lower down, but I think this is basically ready to be merged in. Thanks!

nvinhphuc:
Hi @OwenKephart , I fixed the description. Please check again when it is convenient. Thank you.

nvinhphuc:
Hi @OwenKephart , just for your visibility if you missed, we need this for our CI/CD to run smoothly so please help to review this. Thank you.

OwenKephart (approved):

OwenKephart:
hey! sorry for the delay and thanks for the reminder. I just merged this in and it will go out as part of tomorrow's release

PR #9343, opened 2022-08-11
include ResourceClosedError in retry_mysql_connection_fn

alangenfeld:
It seems that the `mysql` `dbapi` under `sqlalchemy` implicitly holds connections, so we get `ResourceClosedError` on long lived processes even with no explicit pooling turned on.

It should be fine to retry the connection in this condition. Worst case scenario is that it repeats this failure until the retries are exhausted. Ideally will cause the second attempt to connect. 

existing tests


alangenfeld:
* master
  * **PR #9343** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9343" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9343?utm_source=stack-comment).

OwenKephart (approved):

PR #9347, opened 2022-08-11
remove legacy dbt solids

dpeng817:
These had already been removed from the top level of the module and all docs, actually cleans them up from the codebase. Companion internal PR: https://github.com/dagster-io/internal/pull/3324


OwenKephart (approved):
nice

PR #9352, opened 2022-08-11
Add mapping_to and mapping_from methods for graphIn and graphOut

dpeng817:
In order to create graphs with inputs/outputs without the decorator, it's necessary to define input/output mappings that map a graph's outer inputs/outputs to those of inner nodes. Previously, this was only possible using InputDefinition / OutputDefinition, but this adds the capability to `GraphIn`/`GraphOut`. I opted for instance methods , which may be a bit more future-proof if we end up adding further stuff to `GraphIn`/`GraphOut` in the future.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/input.py
+++ b/python_modules/dagster/dagster/_core/definitions/input.py
@@ -15,7 +15,7 @@ from typing import (
 )
 
 import dagster._check as check
-from dagster._annotations import PublicAttr
+from dagster._annotations import PublicAttr, public
 from dagster._core.definitions.events import AssetKey
 from dagster._core.definitions.metadata import (
     MetadataEntry,
@@ -382,7 +382,9 @@ class InputMapping(
         [("definition", InputDefinition), ("maps_to", Union[InputPointer, FanInInputPointer])],
     )
 ):
-    """Defines an input mapping for a graph."""
+    """Defines an input mapping for a graph.
+
+    Users should not construct objects of this class directly. Instead, they should construct using :py:meth:`dagster.GraphIn.mapping_to`."""
 
     def __new__(cls, definition: InputDefinition, maps_to: Union[InputPointer, FanInInputPointer]):
         return super(InputMapping, cls).__new__(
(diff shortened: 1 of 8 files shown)


python_modules/dagster/dagster/_core/definitions/input.py
```diff
@@ -382,7 +382,9 @@ class InputMapping(
         [("definition", InputDefinition), ("maps_to", Union[InputPointer, FanInInputPointer])],
     )
 ):
-    """Defines an input mapping for a graph."""
+    """Defines an input mapping for a graph.
+
+    Users should not construct objects of this class directly. Instead, they should construct using :py:meth:`dagster.GraphIn.mapping_to`."""
```
sryza, line 387: Line length

sryza (commented):
Why not just enable people to directly construct InputMappings and OutputMappings?

dpeng817:
cc @sryza under the hood, Input/OutputMapping use deprecated APIs I think we'd rather not expose. We can change this but it's a non trivial amount of work, and a breaking change. Given that in the past they were constructed via an InputDefinition, I think this is a natural extension.

dpeng817:
* master
  * **PR #9352** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9352" target="_blank">Graphite</a>  👈
    * **PR #9355** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9355" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9352?utm_source=stack-comment).

sryza:
I think the `mapping_to` method made more sense in the world of `InputDefinition`s than it does in the world of `In`s, because, in the former, the object with the `mapping_to` method knew the outer name.  I.e. it saved you the redundancy of needing to write the outer name twice.  But now that `GraphOut` doesn't hold the outer name, it's not really relevant to the mapping.

dpeng817:
@sryza  Hmm I guess since users would previously only construct `InputMapping`/`OutputMapping` from an `InputDefinition`/`OutputDefinition`, it should be okay to change the base class directly..

dpeng817:
cc @sryza I think changing the base class is probably the right move, but doing so is blocked on the removal of composite solid from the codebase (input definitions are pulled off of the input mapping for checking inputs). I think it's fine to keep this PR waiting until that is the case.

sryza (changes requested):

PR #9353, opened 2022-08-11
make op-based tutorial non-redundant with asset tutorial

sryza:
### Summary & Motivation

This makes the op-based tutorial non-redundant with the asset-based tutorial, so users aren't immediately confronted with two ways of doing the same thing.

The use case - reporting basic statistics about files on the local FS - is a tiny bit lame, because it had to fit these constraints:
- Doesn't interact with any external services
- Doesn't create any data assets
- Platform-independent
- Doesn't use any advanced Dagster functionality

I also took this as an opportunity to remove some legacy concepts from the file paths of the supporting code.

Reviewed the new pages locally and updated the tests.


erinkcochran87 (approved):
Approving to unblock. Overall, this looks good to me with the exception of some small failures in Buildkite.

PR #9361, opened 2022-08-12
add a section to the asset tutorial about assets without args and return values

sryza:
https://dagster-git-tutorial-nothing-elementl.vercel.app/tutorial/assets/non-argument-deps

### Summary & Motivation

The asset tutorial currently presents a constrained programming model where all business logic and I/O are separate. I suspect this leads a substantial set of people to read it and conclude that SDAs aren't a good fit for them because they don't have Python objects that correspond to their assets.

This adds a section to the tutorial that shows how to write assets without args and return values. It's perhaps somewhat controversial whether this functionality is important enough to be a section of the intro tutorial.  I believe that it is, based on some of the questions we've gotten in Slack and my gut about how we come off.

Loaded the pages, added a test


docs/content/tutorial/assets/non-argument-deps.mdx
```diff
@@ -0,0 +1,107 @@
+# Assets without Arguments and Return Values
```
erinkcochran87, line 1: This page needs to have Frontmatter, including the `title` and `description`.

erinkcochran87 (approved):
Approved to unblock. I left some final comments on the Vercel preview about some small grammar issues, but overall this looks good.

Issue #9287, opened 2022-08-08
PartitionBackfills for SDAs execute all assets under the multi_asset


OwenKephart:
### Dagster version

1.0.1

### What's the issue?

Currently, the backfills that are created for partitioned asset jobs do not have asset selection data passed in to them. The subselection is handled purely through step_keys_to_execute, which is not fine-grained enough in cases where a single step may produce multiple assets (e.g. dbt multi assets).

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

OwenKephart:
Stems from: https://sourcegraph.com/github.com/dagster-io/dagster/-/blob/js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx?L229 using op names (which get translated to step names).

PR #9369, opened 2022-08-12
Fix asset backfill issue

OwenKephart:
### Summary & Motivation

Resolves: https://github.com/dagster-io/dagster/issues/9287

Basically, when you launched a backfill for assets selected from the AssetGraph, this used to be translated into a set of step_keys_to_execute. There's no precise mapping between step keys and asset keys, so this would fail in more complex situations, such as with multi assets or with reused node definitions.

This change adds in an asset_selection argument to the backfill, essentially allowing you to backfill a set of assets out of a job, instead of a set of steps.

The new test failed before these changes (the part where we check the number of run ids for asset keys would fail, because all the runs would materialize non-selected asset keys). Now it passes


python_modules/dagster/dagster/_core/definitions/assets_job.py
```diff
@@ -231,12 +231,18 @@ def build_node_deps(
 def _has_cycles(deps: Dict[Union[str, NodeInvocation], Dict[str, IDependencyDefinition]]) -> bool:
     """Detect if there are cycles in a dependency dictionary."""
     try:
-        node_deps: Dict[Union[str, NodeInvocation], Set[str]] = {}
+        node_deps: Dict[str, Set[str]] = {}
```
OwenKephart, line 234: this change is unrelated -- it fixes a bug that I found when creating the example for the test.

Basically you would get the wrong error type when checking for cycles sometimes because toposort doesn't play nicely with keys that can be multiple different types.

sryza:
I'm going to defer to you on this one @clairelin135

python_modules/dagster-graphql/dagster_graphql/implementation/execution/backfill.py
```diff
@@ -67,6 +68,12 @@ def create_and_launch_partition_backfill(graphene_info, backfill_params):
         reexecution_steps=backfill_params.get("reexecutionSteps"),
         tags={t["key"]: t["value"] for t in backfill_params.get("tags", [])},
         backfill_timestamp=pendulum.now("UTC").timestamp(),
+        asset_selection=[
+            AssetKey.from_graphql_input(asset_key)
+            for asset_key in backfill_params.get("assetSelection")
+        ]
+        if backfill_params.get("assetSelection")
```
clairelin135, line 75: I think it shouldn't be possible for `assetSelection` to have type None if the input param type is
```
graphene.List(graphene.NonNull(GrapheneAssetKeyInput))
```
Would it make sense to make assetSelection take in `None` to signify an asset selection wasn't provided?

clairelin135 (commented):
The asset selection stuff looks great to me. Had a couple questions in inline comments

python_modules/dagster/dagster/_core/execution/backfill.py
```diff
@@ -72,6 +75,7 @@ def __new__(
         backfill_timestamp: float,
         last_submitted_partition_name: Optional[str] = None,
         error: Optional[SerializableErrorInfo] = None,
+        asset_selection: Optional[FrozenSet[AssetKey]] = None,
```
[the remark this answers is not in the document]
clairelin135: Think this might have to be a list instead?

clairelin135 (approved):
This is looking great to me! Added one small question, but otherwise 🚢

PR #9376, opened 2022-08-15
add cli command to migrate job runs from one repo to another

prha:
### Summary & Motivation
Context: https://github.com/dagster-io/dagster/issues/9317

Enables refactoring of repo code, while preserving history

BK

### DIFF
--- a/python_modules/dagster/dagster/_cli/run.py
+++ b/python_modules/dagster/dagster/_cli/run.py
@@ -1,6 +1,11 @@
 import click
 from tqdm import tqdm
 
+from dagster import __version__ as dagster_version
+from dagster._cli.workspace.cli_target import (
+    get_external_pipeline_or_job_from_kwargs,
+    job_target_argument,
+)
 from dagster._core.instance import DagsterInstance
 
 
@@ -73,7 +78,10 @@ def run_wipe_command(force):
         raise click.ClickException("Exiting without deleting all run history and event logs.")
 
 
-@run_cli.command(name="migrate", help="Migrate the run history from one repository to another.")
+@run_cli.command(
+    name="migrate-repository",
+    help="Migrate the run history for a job from a historic repository to its current repository.",
+)
 @click.option(
     "--from",
     "-f",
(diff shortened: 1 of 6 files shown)


python_modules/dagster/dagster/_cli/run.py
```diff
@@ -70,3 +71,70 @@ def run_wipe_command(force):
         click.echo("Deleted all run history and event logs.")
     else:
         raise click.ClickException("Exiting without deleting all run history and event logs.")
+
+
+@run_cli.command(name="migrate", help="Migrate the run history from one repository to another.")
```
sryza, line 76: What do you think about calling it "migrate-repository" in case there are other kinds of migrations we want to enable in the future?

sryza (commented):
This seems reasonable, though I'm not super well versed in the part of the code, so I think it could be worth it for @alangenfeld to take a look as well.

python_modules/dagster/dagster/_cli/run.py
```diff
@@ -70,3 +71,70 @@ def run_wipe_command(force):
...
+
+        count = len(records)
+        confirmation = click.prompt(
+            f"Are you sure you want to migrate the run history for {job_name} ({count} runs)? Type MIGRATE"
+        )
+        should_migrate = confirmation == "MIGRATE"
+
+        if should_migrate:
+            for record in tqdm(records):
+                with instance.run_storage.connect() as conn:
```
alangenfeld, line 126: keep connection through loop - no need to reconnect each iteration

alangenfeld: little spicy depending on `run_storage.connect` here like this but i suppose the test helps protect (at least the one impl that the test uses)

alangenfeld (commented):
the origin in the serialized `DagsterRun` will be out of sync, we use that as source of truth for idempotence checks.

So if you try to move a job that is hooked up to a sensor with run_key idempotence, the old run key runs will get discarded when we see a non matching origin id

python_modules/dagster/dagster/_cli/run.py
```diff
@@ -70,3 +76,72 @@ def run_wipe_command(force):
...
+        records = instance.get_run_records(
+            filters=RunsFilter(job_name=job_name, tags={REPOSITORY_LABEL_TAG: from_label})
+        )
+
+        if not records:
+            click.echo(f"No runs found for {job_name} in {from_label}.")
+            return
+
+        if not isinstance(instance.run_storage, SqlRunStorage):
+            return
```
gibsondan, line 129: raise?

gibsondan (commented):
latest version here seems reasonable to me, any lingering concerns alex?

alangenfeld (approved):
nope

PR #9400, opened 2022-08-16
remove legacy concepts from CLI help text

sryza:

### DIFF
--- a/python_modules/dagster/dagster/_cli/job.py
+++ b/python_modules/dagster/dagster/_cli/job.py
@@ -16,7 +16,7 @@ from dagster._cli.workspace.cli_target import (
     get_external_pipeline_or_job_from_kwargs,
     get_external_repository_from_kwargs,
     get_external_repository_from_repo_location,
-    get_pipeline_or_job_python_origin_from_kwargs,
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_cli/job.py
```diff
@@ -109,11 +100,7 @@ def execute_list_command(cli_args, print_fn, using_job_op_graph_apis=False):
                 if pipeline.description:
                     print_fn("Description:")
                     print_fn(format_description(pipeline.description, indent=" " * 4))
-                print_fn(
-                    "{solid_or_op}: (Execution Order)".format(
-                        solid_or_op="Ops" if using_job_op_graph_apis else "Solids"
-                    )
-                )
+                print_fn("{solid_or_op}: (Execution Order)".format(solid_or_op="Ops"))
```
clairelin135, line 103: I think we can just do 
```
print_fn("Ops: (Execution Order)")
```
here

clairelin135 (commented):
Looks so much cleaner now! Yay!

Left one comment, but otherwise should be good after fixing up bk

PR #9405, opened 2022-08-17
Cross repo run status sensor

### LINKED ISSUE #9394: run status sensors monitoring jobs in other repos
allow run status sensors to monitor jobs in other repos 

https://dagster.slack.com/archives/C01U5LFUZJS/p1660586409688809

jamiedemaria:
### Summary & Motivation

Adds a `JobAddress` type that allows run status sensors to monitor jobs in other repositories

resolves  #9394

The implementation involves switching the conditional in the run status sensor evaluation function, so the diff is pretty large. Basically we switch from "if the run doesn't meet our condition, skip the sensor" to "if the run meets condition A or condition B continue on, otherwise skip the sensor" 

Questions:
* What do we think about `JobAddress` as the class name? See [this slack thread](https://elementl-workspace.slack.com/archives/C03AFEEDDNX/p1660601283800969) for other proposals 
* Is `repository_location` or `code_location` the correct term?

Todo:
- [ ] docs
new daemon test; added a cross repo run status sensor to toys and confirmed that it works


python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -185,15 +185,35 @@ def build_run_status_sensor_context(
     )
 
 
+class JobAddress(
+    NamedTuple(
+        "_JobAddress",
+        [("repository_location", str), ("repository", str), ("job_name", str)],
```
prha, line 191: should this be location_name, repository_name, job_name?

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -185,15 +185,35 @@ def build_run_status_sensor_context(
     )
 
 
+class JobAddress(
```
prha, line 188: What's the difference between `JobAddress` and `PipelineSelector` (maybe renamed to `JobSelector`)?

prha (commented):
Currently, the `monitored_jobs` argument for run status sensors is optional, and it filters by the current repository.   Is there a way to target a remote repository, but not specify a job name?

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
jamiedemaria: yeah, one of the questions i have is on the naming and terminology. I'm still a bit unclear on the specific differences between `code_location` `repository_location` or just `location`

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
jamiedemaria: ah didn't know `PipelineSelector` existed! they look pretty similar to me. I think this still needs a new type (PipelineSelector references solids) but I can make a `JobSelector` alongside `PipelineSelector`

jamiedemaria:
> Is there a way to target a remote repository, but not specify a job name?
@prha that's a good idea. One way this could work is we add a param `monitored_repositories` that takes a list of `RepositorySelector`. if given it will monitor all jobs in the repositories. if both `monitored_repositories` and `monitored_jobs` are given then we monitor all jobs in the `RepositorySelectors`, monitor the jobs passed by definition in the current repository, and monitor jobs passed by `JobSelector` in their repository

sryza:
> if both monitored_repositories and monitored_jobs are given then we monitor all jobs in the RepositorySelectors, monitor the jobs passed by definition in the current repository, and monitor jobs passed by JobSelector in their repository

I wonder if it would be simplest to just make these two parameters mutually exclusive?

Also, what if someone wants to monitor all jobs in all repositories?

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
prha: they're all the same.

sryza: I think we're trying to move away from `repository_location`

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -14,6 +14,8 @@
 from dagster._core.events import PIPELINE_RUN_STATUS_TO_EVENT_TYPE, DagsterEvent
 from dagster._core.instance import DagsterInstance
 from dagster._core.storage.pipeline_run import DagsterRun, DagsterRunStatus, PipelineRun, RunsFilter
+
+# from dagster._core.host_representation import JobSelector, RepositorySelector
```
jamiedemaria, line 18: I was thinking it would make sense to use the existing RepositorySelector and make a new JobSelector, but the classes in host_representation aren't made public and attempting to import them here causes a circular import

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -195,6 +224,7 @@ def run_failure_sensor(
     job_selection: Optional[
         List[Union[PipelineDefinition, GraphDefinition, UnresolvedAssetJobDefinition]]
     ] = None,
+    monitored: Optional[List[Union[RepositorySelector, JobSelector]]] = None,
```
jamiedemaria, line 227: todo - switch to sequence

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -552,12 +617,14 @@ def run_status_sensor(
         minimum_interval_seconds (Optional[int]): The minimum number of seconds that will elapse
             between sensor evaluations.
         description (Optional[str]): A human-readable description of the sensor.
-        monitored_jobs (Optional[List[Union[PipelineDefinition, GraphDefinition, UnresolvedAssetJobDefinition]]]):
+        monitored_jobs (Optional[List[Union[PipelineDefinition, GraphDefinition, UnresolvedAssetJobDefinition, JobAddress]]]):
```
jamiedemaria, line 620: todo - remove JobAddress from docs

python_modules/dagster/dagster/_utils/alert.py
```diff
@@ -85,11 +89,24 @@ def make_email_on_run_failure_sensor(
     name: Optional[str] = None,
     dagit_base_url: Optional[str] = None,
     monitored_jobs: Optional[
-        List[Union["PipelineDefinition", "GraphDefinition", "UnresolvedAssetJobDefinition"]]
+        List[
```
jamiedemaria, line 92: todo - update to sequence

python_modules/dagster/dagster/_utils/alert.py
```diff
@@ -10,7 +10,11 @@
 if TYPE_CHECKING:
     from dagster._core.definitions.graph_definition import GraphDefinition
     from dagster._core.definitions.pipeline_definition import PipelineDefinition
-    from dagster._core.definitions.run_status_sensor_definition import RunFailureSensorContext
+    from dagster._core.definitions.run_status_sensor_definition import (
```
jamiedemaria, line 13: todo - update slack alert and msteams alerts to take `monitored`

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -220,6 +250,8 @@ def run_failure_sensor(
         job_selection (Optional[List[Union[JobDefinition, GraphDefinition]]]): (deprecated in favor of monitored_jobs)
             The jobs that will be monitored by this failure sensor. Defaults to None, which means
             the alert will be sent when any job in the repository fails.
+        monitored (Optional[List[Union[RepositorySelector, JobSelector]]]): Other repositories or jobs in other repositories that this
```
jamiedemaria, line 253: todo - update doc string for `monitored_jobs` and `job_selection` to say that it's for jobs in the same repo as the sensor

jamiedemaria (commented):
updated this to do the following:
* run_status_sensors now take a `monitored` arg that can be a `RepositorySelector` or `JobSelector` and is meant to cover monitoring jobs/repos that are not in the current repo
* reverts `monitored_jobs` to only taking `JobDefinitions` and is meant to monitor jobs in the current repo
* Makes a `JobSelector` type, but having some issues with where this class should live (see the comment i left)

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
prha: You can tackle this with the same mypy workaround, by using strings instead of type objects, and then importing the class in a `TYPE_CHECKING` conditional block (as opposed to on import).

jamiedemaria:
added some docs and got CI green so should be almost set. will take a final pass myself when i’m back on wifi

python_modules/libraries/dagster-msteams/dagster_msteams/sensors.py
```diff
@@ -47,6 +63,11 @@ def make_teams_on_run_failure_sensor(
             messages to include deeplinks to the failed run.
         default_status (DefaultSensorStatus): Whether the sensor starts as running or not. The default
             status can be overridden from Dagit or via the GraphQL API.
+        monitored_jobs (Optional[List[Union[PipelineDefinition, GraphDefinition, UnresolvedAssetJobDefinition, RepositorySelector, JobSelector]]]):
+            Jobs in the current repository that will be monitored by this sensor. Defaults to None, which means the alert will
+            be sent when any job in the repository matches the requested run_status. To monitor jobs in external repositories, use RepositorySelector and JobSelector
+        monitored (Optional[Sequence[Union[RepositorySelector, JobSelector]]]): Other repositories or jobs in other repositories that this
```
prha, line 69: let's remove

prha (approved):
I think this looks good!

One thing I'd want to advocate for (maybe in a follow-up diff before `1.1.0`) is to change the default when `monitored_jobs` is empty.  Maybe we can warn and ask them to specify `RepositorySelector` for the current repo?  It feels more explicit to me (current behavior seems somewhat magical).

PR #9406, opened 2022-08-17
[dagit] Refetch queries instead of dumping Apollo cache

hellendag:
### Summary & Motivation

Now that we have `refetchQueries` available in Apollo, use it instead of resetting the entire Apollo store. This will hopefully make it a bit less jarring when the user reloads a repository or entire workspace.

Load Dagit. Reload a repo after making changes to its Python, verify with console logging that queries are refetched and that the UI refreshes correctly. Reload the entire Workspace, verify same.

### DIFF
--- a/js_modules/dagit/packages/core/src/nav/RepositoryLocationStateObserver.tsx
+++ b/js_modules/dagit/packages/core/src/nav/RepositoryLocationStateObserver.tsx
@@ -78,8 +78,7 @@ export const RepositoryLocationStateObserver = () => {
               underline="always"
               onClick={() => {
                 setUpdatedLocations([]);
-                refetch();
-                client.resetStore();
+                client.refetchQueries({include: 'active'});
               }}
             >
               Update data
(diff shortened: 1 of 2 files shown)


hellendag:
* master
  * **PR #9406** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9406" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9406?utm_source=stack-comment).

js_modules/dagit/packages/core/src/nav/RepositoryLocationStateObserver.tsx
```diff
@@ -78,8 +78,7 @@ export const RepositoryLocationStateObserver = () => {
               underline="always"
               onClick={() => {
                 setUpdatedLocations([]);
-                refetch();
-                client.resetStore();
+                client.refetchQueries({include: 'active'});
```
salazarm, line 81: why do we need `refetchQueries` here? Can we narrow down the queries being fetched somehow? My concern is that  we might refetch some expensive query added later that doesn't need to be refetched

salazarm (approved):

PR #9407, opened 2022-08-17
[dagster-dbt] load_assets_from_dbt_* should support an exclude parameter, in addition to select

### LINKED ISSUE #8772: [dagster-dbt] load_assets_from_dbt_* should support an exclude parameter, in addition to select
### What's the use case?

Both of these parameters are useful for describing subsets of a dbt project, so it'd be nice if both were supported

### Ideas of implementation

_No response_

### Additional information

_No response_

flvndh:
Resolves #8772

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
@@ -343,8 +343,8 @@ def _dbt_nodes_to_assets(
 
     # prevent op name collisions between multiple dbt multi-assets
     op_name = f"run_dbt_{package_name}"
-    if select != "*":
-        op_name += "_" + hashlib.md5(select.encode()).hexdigest()[-5:]
+    if select != "*" or exclude:
+        op_name += "_" + hashlib.md5(select.encode() + exclude.encode()).hexdigest()[-5:]
 
     dbt_op = _get_dbt_op(
         op_name=op_name,
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
@@ -532,8 +532,12 @@ def load_assets_from_dbt_manifest(
     dbt_nodes = {**manifest_json["nodes"], **manifest_json["sources"]}
 
     if selected_unique_ids:
-        select = " ".join(".".join(dbt_nodes[uid]["fqn"]) for uid in selected_unique_ids)
-        exclude = ""
+        select = (
+            " ".join(".".join(dbt_nodes[uid]["fqn"]) for uid in selected_unique_ids)
+            if select is None
+            else select
+        )
+        exclude = "" if exclude is None else exclude
     else:
         select = select if select is not None else "*"
         exclude = exclude if exclude is not None else ""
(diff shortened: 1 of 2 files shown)


sryza:
@OwenKephart mind taking a look?

OwenKephart:
Hi @flvndh ! Thanks so much for the contribution, this looks great -- exactly what I was imagining in terms of implementation :)

One thing to note -- I ran the test suite on this, and ran into some errors with the `load_assets_from_dbt_manifest` function. Basically, the `parse_difference` function fails with the following stack trace:

```
    selected_unique_ids = _select_unique_ids_from_manifest_json(manifest_json, select, exclude)                                                                                 [1120/1986]
../dagster_dbt/asset_defs.py:103: in _select_unique_ids_from_manifest_json
    parsed_spec = graph_cli.parse_difference(
/Users/owen/.virtualenvs/dagster/lib/python3.8/site-packages/dbt/graph/cli.py:78: in parse_difference
    include, DEFAULT_INCLUDES, indirect_selection=IndirectSelection(flags.INDIRECT_SELECTION)
/Users/owen/.pyenv/versions/3.8.7/lib/python3.8/enum.py:339: in __call__
    return cls.__new__(cls, value)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cls = <enum 'IndirectSelection'>, value = None

    def __new__(cls, value):
        # all enum instances are actually created during class construction
        # without calling this method; this method is called by the metaclass'
        # __call__ (i.e. Color(3) ), and by pickle
        if type(value) is cls:
            # For lookups like Color(Color.RED)
            return value
        # by-value search for a matching enum member
        # see if it's in the reverse mapping (for hashable values)
        try:
            return cls._value2member_map_[value]
        except KeyError:
            # Not found, no need to do long O(n) search
            pass
        except TypeError:
            # not there, now do long search -- O(n) behavior
            for member in cls._member_map_.values():
                if member._value_ == value:
                    return member
        # still not found -- try _missing_ hook
        try:
            exc = None
            result = cls._missing_(value)
        except Exception as e:
            exc = e
            result = None
        if isinstance(result, cls):
            return result
        else:
            ve_exc = ValueError("%r is not a valid %s" % (value, cls.__name__))
            if result is None and exc is None:
>               raise ve_exc
E               ValueError: None is not a valid IndirectSelection

/Users/owen/.pyenv/versions/3.8.7/lib/python3.8/enum.py:662: ValueError
```

Basically, the default dbt `flags.IndirectSelection` is None, which is invalid, so this call fails. I was wondering if you experienced this error on your end as well (i'm on dbt 1.2.0).

I tried swapping out the offending line with:

```
# create a parsed selection from the select string
parsed_spec = graph_cli.parse_union([select], True)
if exclude:
    parsed_spec = graph_cli.SelectionDifference(
        components=[parsed_spec, graph_cli.parse_union([exclude], True)]
    )
```

and it seemed to pass the tests locally for me.

python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
```diff
@@ -338,6 +346,7 @@ def _dbt_nodes_to_assets(
         ins=dict(asset_ins.values()),
```
OwenKephart, line 346: git won't let me comment on the correct line, but I think technically the `op_name` parameter should be a hash of both the select AND the exclude parameters. I think in practice, this will not cause problems (because you'd need to have multiple invocations of load_assets...() with the same select parameter but different exclude parameters, which would be hard to do in any valid workflow, but I thought I'd mention.

OwenKephart (approved):

OwenKephart:
hi @flvndh ! Thanks again for the contribution, I just merged this in :)

PR #9415, opened 2022-08-17
event watcher thread only join if alive

alangenfeld:
In some rare cases (in tests) we may quickly shutdown before the thread has had a chance to start - only join if its alive. 

ran test that sporadically failed  100 times


shalabhc (approved):
There may be a race condition in `watch_run` as well with a non atomic test and set here:
```
if not self._watcher_thread:
    ....
    self._watcher_thread = ...
```
Not sure if this will actually cause problems.

PR #9417, opened 2022-08-17
[dagit] Repair Asset table

hellendag:
### Summary & Motivation

An odd bug appears in circumstances where there are lots of namespaces in an asset table. We're slicing the namespaces for the entire set of assets before comparing the set to the prefix path, which means there's a good chance that we end up missing prefix path matches in the asset set.

Instead, slice the list of displayed assets *after* filter matches are found.

Also fix a bug where an empty asset table has the check-all checkbox set to `checked` because we're not checking how many checked items we have.

View a very large asset table at a point where nested assets would incorrectly fail to show any filtered assets. Verify that they now appear correctly, and can be navigated as normal.

Verify also that an empty asset table will not have the check-all box checked.

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/AssetTable.tsx
+++ b/js_modules/dagit/packages/core/src/assets/AssetTable.tsx
@@ -120,7 +120,10 @@ export const AssetTable = ({
                   checkedPathsOnscreen.length > 0 &&
                   checkedPathsOnscreen.length !== pageDisplayPathKeys.length
                 }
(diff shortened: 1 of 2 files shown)


hellendag:
* master
  * **PR #9417** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9417" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9417?utm_source=stack-comment).

bengotow (approved):
Oh man this is subtle - good catch 🙌

Issue #9337, opened 2022-08-11
Does dagster-dbt support dbt metrics? Because I get an error when using a dbt project that has metrics


ericwijaya001:
### Dagster version

0.15.8

### What's the issue?

Does dagster-dbt support dbt metrics? Because I get an error when using a dbt project that has metrics

### What did you expect to happen?

I don't know why even though I've defined metrics correctly in schema.yml and I've called it in one of the dbt model files

### How to reproduce?

schema.yml file :
metrics:
  - name: avg_buy_in_usd
    label: Avg buy in usd
    model: ref('converted_to_usd')
    description: "Avg buy in usd "

    type: average
    sql: idr_product_price 

    timestamp: created_at
    time_grains: [day, week, month]

    dimensions:
      - user_id

metrics_avg_buy_per_user_id.sql dbt model:
{{ config(materialized='table') }}
select * 
from {{metrics.calculate(metric('avg_buy_in_usd'), 
        grain='month', 
    )}}

Then I run dagster (dagit) and get an error like this:
C:\users\eric\appdata\local\programs\python\python37\lib\site-packages\dagster\_core\workspace\context.py:557: UserWarning: Error loading repository location coba_dagster_dbt_gg:KeyError: 'metric.coba.avg_buy_in_usd'

Stack Trace:
  File "C:\users\eric\appdata\local\programs\python\python37\lib\site-packages\dagster\_grpc\server.py", line 230, in __init__
ython_package
    package_name, working_directory, remove_from_path_fn=remove_from_path_fn
  File "C:\users\eric\appdata\local\programs\python\python37\lib\site-packages\dagster\_core\code_pointer.py", line 136, in load_python_module
    return importlib.import_module(module_name)
  File "C:\users\eric\appdata\local\programs\python\python37\lib\importlib\__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
[...]

Issue #8771, opened 2022-07-06
[dagster-dbt] Do not represent ephemeral dbt models as assets


OwenKephart:
### What's the use case?

Ephemeral dbt models are not materialized in the database -- they're inlined into downstream models instead. This means that they should not be represented as assets, as they are right now.

### Ideas of implementation

This is not super trivial, as we'll need to reconstruct the lineage graph once we prune out the ephemeral models, but on the whole it seems straightforward enough.

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

PR #9419, opened 2022-08-17
[dagster-dbt] allow dbt projects including metrics to be loaded

OwenKephart:
### Summary & Motivation

Fixes: https://github.com/dagster-io/dagster/issues/9337 and https://github.com/dagster-io/dagster/issues/8771

This adds special handling for two special types of dbt nodes: metrics, and ephemeral models. These both share the property of being nodes that can be either upstream or downstream of regular dbt models, but are not assets themselves (because they basically are just sql that is in-lined into actual model queries).

When such a node is a parent of a regular node, instead of listing its node_id as the parent_unique_id, we do a BFS up the dependency graph to search for valid parent nodes. Basically, if we have:

A -> (Ephemeral1) -> (Ephemeral2) -> B

, then we replace it with:

A -> B

Added two new ephemeral models to the test project, and all the tests still passed (so the lineage was unbroken and the ephemeral models were not registered as assets)


sryza:
Will this represent metrics as assets?

sryza:
Unless we're pretty confident that making assets from the metrics is a good idea, I think it would be better to just silently ignore the metrics, because it will be easier to add them in later than to remove them.

OwenKephart:
Note: will rework implementation to not include metrics as assets

clairelin135 (approved):

PR #9421, opened 2022-08-17
RFC: Search Page

clairelin135:
Adds a full-page search with paginated results. The DocSearch modal currently has a limitation where only the top 5 results per section (concepts, guides, etc.) are displayed. This page will enable seeing the full results in the case that the top 5 results don't have the desired content.

[Preview Link](https://dagster-9d5xgfqh2-elementl.vercel.app/searchpage?query=hello)

How to view:
Click on "see all search results" at bottom of modal"
image

Full page search result:
image

I'm not super familiar with front end, so curious how this code is looking! I also styled it myself (with pretty rudimentary styling) so welcoming feedback there too

A future step for this functionality is to completely sync the URL (so that when you have a new search, the page URL will change too). This will be nice so that users can share paginated search results via URL.


yuhan (approved):
this is great! another future improvement could be to improve the UX to e.g. make the section / type of the doc more prominent. but this is a great start!

PR #9427, opened 2022-08-18
[dagster-io/ui] Support spinner in Tag

hellendag:
### Summary & Motivation

We often want to put spinners in `Tag` components, so let's just support that without requiring anyone to use `BaseTag`.

Screen Shot 2022-08-18 at 1 14 41 PM

Storybook examples.


bengotow (approved):
Ahh man good call - I always feel bad importing BaseTag...

PR #9429, opened 2022-08-18
captured log manager

prha:
### Summary & Motivation
Creates a new, simplified API for capturing logs.

Also, this converts the existing local compute log manager to implement the CapturedLogManager interface, and implements the local compute log manager interface on top of the more general captured log manager implementation.

Also added a generic test class for captured logs to minimize copy/paste testing for each implementation.

Not included in this diff:
* Hooking up the compute log subscription to this interface.
* Changing executors to capture by process rather than by step.
* Changing the compute logs UI in any way

BK

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/captured_log_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/captured_log_manager.py
@@ -1,0 +1,217 @@
+# pylint: disable=unused-argument
+
+from abc import ABC, abstractmethod
+from contextlib import contextmanager
+from typing import List, NamedTuple, Optional
+
+MAX_BYTES_CHUNK_READ = 4194304  # 4 MB
+
+
+class CapturedLogData(
+    NamedTuple(
+        "_CapturedLogData",
+        [
+            ("log_key", List[str]),
+            ("stdout", Optional[bytes]),
+            ("stderr", Optional[bytes]),
+            ("cursor", Optional[str]),
+        ],
+    )
+):
+    """
+    Object representing captured log data, either a partial chunk of the log data or the full
+    capture.  Contains the raw bytes and optionally the cursor offset for the partial chunk.
+    """
+
+    def __new__(
+        cls,
+        log_key: List[str],
+        stdout: Optional[bytes] = None,
+        stderr: Optional[bytes] = None,
+        cursor: Optional[str] = None,
+    ):
+        return super(CapturedLogData, cls).__new__(cls, log_key, stdout, stderr, cursor)
+
+
+class CapturedLogMetadata(
+    NamedTuple(
+        "_CapturedLogMetadata",
+        [
+            ("external_url", Optional[str]),
+            ("stdout_location", Optional[str]),
+            ("stderr_location", Optional[str]),
+            ("stdout_download_url", Optional[str]),
+            ("stderr_download_url", Optional[str]),
+        ],
+    )
+):
+    """
+    Object representing metadata info for the captured log data, containing a display string for
+    the location of the log data and a URL for direct download of the captured log data.
+    """
+
+    def __new__(
+        cls,
+        external_url: Optional[str] = None,
+        stdout_location: Optional[str] = None,
+        stderr_location: Optional[str] = None,
+        stdout_download_url: Optional[str] = None,
+        stderr_download_url: Optional[str] = None,
+    ):
+        return super(CapturedLogMetadata, cls).__new__(
+            cls,
+            external_url=external_url,
+            stdout_location=stdout_location,
+            stderr_location=stderr_location,
+            stdout_download_url=stdout_download_url,
+            stderr_download_url=stderr_download_url,
+        )
+
+
+class CapturedLogSubscription:
+    def __init__(self, manager, log_key, cursor):
+        self._manager = manager
+        self._log_key = log_key
+        self._cursor = cursor
+        self._observer = None
+
+    def __call__(self, observer):
+        self._observer = observer
+        self.fetch()
+        if self._manager.is_capture_complete(self._log_key):
+            self.complete()
+        return self
+
+    @property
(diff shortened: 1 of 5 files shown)


python_modules/dagster/dagster/_core/storage/local_compute_log_manager.py
```diff
@@ -71,21 +62,116 @@ def config_type(cls):
...
+        check.inst_param(pipeline_run, "pipeline_run", PipelineRun)
+        check.opt_str_param(step_key, "step_key")
+
+        log_key = self.build_log_key(pipeline_run.run_id, step_key or pipeline_run.pipeline_name)
+        with self.capture_logs(log_key):
+            yield
+
+    def build_log_key(self, run_id, key):
+        """Legacy adapter to translate run_id/key to captured log manager-based log_key"""
+        return [run_id, "compute_logs", key]
```
alangenfeld, line 168: nit: sub_key

python_modules/dagster/dagster/_core/storage/captured_log_manager.py
```diff
@@ -0,0 +1,122 @@
...
+
+        Args:
+            log_key (List[String]): The log key identifying the captured logs
+
+        Returns:
+            Boolean
+        """
+
+    @abstractmethod
+    def get_stdout(
```
alangenfeld, line 65: nit: think a suffix would be nice, `_data`, `_bytes`

python_modules/dagster/dagster/_core/storage/captured_log_manager.py
```diff
@@ -0,0 +1,122 @@
...
+        [("location", Optional[str]), ("download_url", Optional[str])],
+    )
+):
+    """
+    Object representing metadata info for the captured log data, containing a display string for
+    the location of the log data and a URL for direct download of the captured log data.
+    """
+
+    def __new__(cls, location: Optional[str] = None, download_url: Optional[str] = None):
+        return super(CapturedLogMetadata, cls).__new__(cls, location, download_url)
```
alangenfeld, line 35: [1]

alangenfeld (commented):
I think we should consider adding a layer of indirection (leaning in to / fleshing out [1])  to leave us room to support some cases we've repeatedly discussed:
* sending signed s3 urls to browser and removing webservers as middle man for bytes
* having "captured logs" just generate urls to existing capture systems 
* streaming (maybe already covered)

just riffing - some tagged type w/ enum or union of supported types, something like:
*  "direct" and gives back the `key` that you can pass back to dagit to get bytes directly
* "indirect_url" where a url is returned that we can download to in browser
* "indirect_external" where a url or just display string is given to show in dagit to help the user get to existing logs but we dont deal with bytes directiy 

dont love those names, but i think as long as we leave a layer of indirection via tagged types and make the APIs somewhat coherent against that we will be setup

prha:
I like the breakdown of the different types, at least for differentiating the behavior on the read side, but on the capture side, I think this is probably best configured on a different part of the system.  I was thinking that because captured logs are configured on the instance and compute logs are generated (currently) during run execution, this type of external capture is best configured on the run launcher.  You would then have two different configurations for capturing run logs, vs dagster-controlled captured sensor logs on the grpc server, for example.

> sending signed s3 urls to browser and removing webservers as middle man for bytes
I think we already do this for download urls, but we haven't built out the logic in the client to read from the s3 url for the custom UI.

> having "captured logs" just generate urls to existing capture systems
I think this is probably a different logging system, that gets configured at the run launcher / executor level.

> streaming (maybe already covered)
I don't think this API precludes this (it's just a contextmanager).

python_modules/dagster/dagster/_core/storage/captured_log_manager.py
```diff
@@ -0,0 +1,217 @@
...
+    """
+
+    def __new__(
+        cls,
+        log_key: List[str],
+        stdout: Optional[bytes] = None,
+        stderr: Optional[bytes] = None,
+        cursor: Optional[str] = None,
+    ):
+        return super(CapturedLogData, cls).__new__(cls, log_key, stdout, stderr, cursor)
```
alangenfeld, line 33: having both out and err streams in the same payload seems like it might be awkward. I didn't think much of it until i saw the combined cursor.

what led you to this set-up?

prha: I first had them split out, but it just seemed easier to group them together.  I don't have super strong feelings here...  The frontend subscribes/unsubscribes to both at the same time, each with its own subscription.

prha:
* master
  * **PR #9429** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9429" target="_blank">Graphite</a>  👈
    * **PR #9956** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9956" target="_blank">Graphite</a> 
      * **PR #9957** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9957" target="_blank">Graphite</a> 
        * **PR #9958** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9958" target="_blank">Graphite</a> 
          * **PR #9959** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9959" target="_blank">Graphite</a> 
            * **PR #9960** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9960" target="_blank">Graphite</a> 
              * **PR #9961** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9961" target="_blank">Graphite</a> 
                * **PR #9962** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9962" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9429?utm_source=stack-comment).

Issue #9270, opened 2022-08-07
CVE-2022-34749 through nbconvert


nico525:
Currently, there is CVE-2022-34749 which is popping up in my trivy check on the image running dagit 1.0.1.
This is due to dagit requiring nbconvert which imports an old mistune version:
https://github.com/dagster-io/dagster/blob/d8612490271eac50d1fd7d494b6d95727186e8f6/python_modules/dagit/setup.py#L54

Can we make the notebook download endpoint and subsequently the nbconvert requirement an optional/extra requirement?

clairelin135:
Ideally nbconvert updates to a newer version of mistune, but until then making nbconvert an extra requirement sounds reasonable to me.

Thoughts @alangenfeld ?

alangenfeld:
It could be done. We would need to make sure that the webapp degrades gracefully in its absence and ideally points a user in the right direction if the feature is engaged with without the extra installed.

nico525:
I fully agree with @clairelin135 but it appears the next release of nbconvert which bumps mistune will also be a new major release and therefore might be breaking some of the code logic.
If it wasn't critical, I wouldn't be too concerned, but most enterprise pipelines are configured to stop if a critical CVE is found

sryza:
A separate reason to make nbconvert and extras requirement: the latest major version requires a jinja version that's incompatible with dbt.

https://dagster.slack.com/archives/C01U5LFUZJS/p1660724879367519

clairelin135:
I will take on this issue

PR #9432, opened 2022-08-18
Make nbconvert an extra install in Dagit

clairelin135:
In response to https://github.com/dagster-io/dagster/issues/9270

This PR makes nbconvert an extra install in Dagit. When nbconvert is not installed, Dagit will display with the following error:
image

Tested by running:
`uvicorn dagit.app:default_app`
and viewed `hello_world_notebook_pipeline` notebook in Dagit


python_modules/dagit/dagit/webserver.py
```diff
@@ -124,6 +123,15 @@ async def download_debug_file_endpoint(self, request: Request):
         return StreamingResponse(result, media_type="application/gzip")
 
     async def download_notebook(self, request: Request):
+        try:
+            import nbconvert
+        except ImportError:
+            return HTMLResponse(
+                "Notebook support requires nbconvert, which is not installed. You can install nbconvert using Dagit's 'notebook' extra via <code>pip install dagit[notebook]</code>"
```
sryza, line 130: Nitpick: line length

sryza (commented):
This seems like a net benefit to me. Curious on @alangenfeld 's thoughts, as he raised some initial objections.  I think we'd want dagster cloud to use the nbconvert extra?

alangenfeld (approved):
this looks great to me - just double check the internal repo dep set-up like sandy mentioned

python_modules/dagit/dagit/webserver.py
```diff
@@ -124,6 +123,15 @@ async def download_debug_file_endpoint(self, request: Request):
...
+            )
+
+        from nbconvert import HTMLExporter
```
alangenfeld, line 133: should probably just do the useful import in the try block instead of this - i feel like this double import set-up is more fragile

alangenfeld (commented):
would be great to protect this with a test but i understand thats a lot since we need to create separate env

clairelin135:
Thanks! I updated according to feedback and filed an internal diff. I probably won't add a separate venv for the test because it's a lot for a feature that is used pretty infrequently

PR #9433, opened 2022-08-18
RunRequests support asset_selection

OwenKephart:
### Summary & Motivation

It's nice to be able to kick off just a selection of assets on a job rather than the whole thing

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/run_request.py
+++ b/python_modules/dagster/dagster/_core/definitions/run_request.py
@@ -84,7 +84,6 @@ class RunRequest(
             run_config=check.opt_dict_param(run_config, "run_config", key_type=str),
             tags=check.opt_dict_param(tags, "tags", key_type=str, value_type=str),
             job_name=check.opt_str_param(job_name, "job_name"),
-            # convert all sequence inputs to lists
             asset_selection=check.opt_list_param(
                 asset_selection, "asset_selection", of_type=AssetKey
             ),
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster/_core/definitions/run_request.py
```diff
@@ -72,13 +76,17 @@ def __new__(
         run_config: Optional[Mapping[str, Any]] = None,
         tags: Optional[Mapping[str, str]] = None,
         job_name: Optional[str] = None,
+        asset_selection: Optional[List[AssetKey]] = None,
```
sryza, line 79: Nitpick: Sequence

sryza (approved):
Oh wow very simple. Is there any reason we might want this to be experimental?

OwenKephart:
@sryza One thing to think about is if we should allow the `asset_selection` argument to take in an `AssetSelection` object (instead of a list of asset keys). Also there's a bit of inconsistency around the codebase on if asset_selection is a Set[AssetKey] or List[AssetKey], so we might want to give ourself some leeway there (although realistically we should probably just accept either).

cc @clairelin135 if you have thoughts

clairelin135:
I think we should definitely allow `AssetSelection` objects in asset_selection arguments. In a sensor, you would ideally want to kick off materializations for an asset an all of its downstream assets. It's way more ergonomic to be able to say `AssetSelection("a").downstream()` than to have to manually specify all the asset keys.

sryza:
@clairelin135 good point

PR #9435, opened 2022-08-18
walk back automatic secret passing into serverless image

prha:
### Summary & Motivation
Not sure we want to automatically bake all GH secrets into the serverless image.  This should only affect folks who clone *after* we push https://github.com/dagster-io/dagster-cloud-serverless-quickstart/pull/13

BK


benpankow (approved):
do we want to add docs somewhere?

Issue #8925, opened 2022-07-18
from Launchpad, it's too easy to kick off a non-partitioned run of a partitioned job


sryza:
A partitioned job will usually fail if not launched with a partition.  However, if you go to the Launchpad and launch a run and forget to select a partition, this failure will often not occur until runtime.  We should force you or heavily encourage you to select a partition before launching a run.

It's possible that there are some edge cases, with non-asset jobs, where this is a desirable thing to do - e.g. if you want to manually provide all the config that you'd otherwise provide through the partition, and you don't use `context.partition_key` or similar anywhere. So I'm not sure we should 100% outlaw it in all cases.  In the asset case, we should 100% outlaw it.

bengotow:
It looks like this is still the case and can be repro'd with the `many_events` job, which runs (and succeeds) without a partition specified but maybe shouldn't.

I'll add a warning of some kind that you have to click through to start the job without the partition filled in if a partition set is present.

PR #9442, opened 2022-08-22
[dagit] Warn before launching a partitioned job without a partition from the launchpad

bengotow:
### Summary & Motivation

This is a small fix to an issue reported here: https://github.com/dagster-io/dagster/issues/8925

### DIFF
--- a/js_modules/dagit/packages/core/src/launchpad/LaunchButton.tsx
+++ b/js_modules/dagit/packages/core/src/launchpad/LaunchButton.tsx
@@ -12,11 +12,13 @@ import {
 import * as React from 'react';
 import styled from 'styled-components/macro';
 
+import {useConfirmation} from '../app/CustomConfirmationProvider';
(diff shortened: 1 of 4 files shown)


js_modules/dagit/packages/core/src/launchpad/LaunchButton.tsx
```diff
@@ -173,6 +175,7 @@ export const LaunchButtonDropdown = ({
 
 interface ButtonWithConfigurationProps {
   title: string;
+  warning?: string;
```
hellendag, line 178: Maybe a `React.ReactNode` for this?

hellendag (approved):

js_modules/dagit/packages/core/src/launchpad/LaunchButton.tsx
bengotow: Ahh yeah that should work and would be nice futureproofing 🙌

PR #9448, opened 2022-08-22
[dagit] When launching assets, warn if upstreams have never been materialized

bengotow:
### Summary & Motivation

This PR adds another pre-flight check to the Materialize button in Dagit. We fetch the last materialization of the assets immediately upstream of the ones you're launching and warn if any of them have never been materialized. This check also loads enough data to see / skip upstream assets that are source assets. 

For partitioned assets, we just naively check to see if ANY materialization exists because we can't know for sure that the upstream asset is partitioned in the same way as the selected asset.

image

I also refactored a few things into helper methods because this code is getting pretty intense.

Tested with:
- No upstream assets
- An upstream source asset
- An upstream asset that has been materialized
- An upstream asset that has never been materialized

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
+++ b/js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
@@ -5,9 +5,10 @@ import uniq from 'lodash/uniq';
 import React from 'react';
 
 import {showCustomAlert} from '../app/CustomAlertProvider';
+import {useConfirmation} from '../app/CustomConfirmationProvider';
 import {IExecutionSession} from '../app/ExecutionSessionStorage';
 import {usePermissions} from '../app/Permissions';
-import {isSourceAsset} from '../asset-graph/Utils';
+import {displayNameForAssetKey, isSourceAsset} from '../asset-graph/Utils';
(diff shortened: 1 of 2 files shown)


js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
```diff
@@ -274,6 +301,41 @@ async function stateForLaunchingAssets(
   };
 }
 
+function getCommonJob(assets: LaunchAssetExecutionAssetNodeFragment[], preferredJobName?: string) {
+  const everyAssetHasJob = (jobName: string) => assets.every((a) => a.jobNames.includes(jobName));
+  const jobsInCommon = assets[0] ? assets[0].jobNames.filter(everyAssetHasJob) : [];
+  return jobsInCommon.find((name) => name === preferredJobName) || jobsInCommon[0];
```
hellendag, line 307: Makes me itchy to return the `[0]` item on an array that may intentionally be empty.

hellendag (approved):

js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
bengotow: Hmm let me make this a bit cleaner before I merge 👍

PR #9451, opened 2022-08-22
[dagit] Implement repo-bucketed Run timeline

hellendag:
### Summary & Motivation

Implement repo-level bucketing on the run timeline component, with collapsed/expanded state preserved in localStorage. This is behind a client-side feature flag for now.

The idea here is to make the run timeline a little easier to visually parse and consume, and allow users to focus on the specific jobs they care about.

Screen Shot 2022-08-22 at 4 43 51 PM

Storybook example. Toggle bucketing on and off, verify correct rendering in each case. Verify that expand/collapse works as expected, and that identically-named repos display the repo location name as well.

### DIFF
--- a/js_modules/dagit/packages/core/src/runs/QueryfulRunTimeline.tsx
+++ b/js_modules/dagit/packages/core/src/runs/QueryfulRunTimeline.tsx
@@ -1,0 +1,241 @@
+import {gql, useQuery} from '@apollo/client';
+import * as React from 'react';
+
+import {useFeatureFlags} from '../app/Flags';
+import {SCHEDULE_FUTURE_TICKS_FRAGMENT} from '../instance/NextTick';
+import {InstigationStatus, RunStatus} from '../types/globalTypes';
+import {LoadingSpinner} from '../ui/Loading';
+import {buildRepoAddress} from '../workspace/buildRepoAddress';
+import {repoAddressAsString} from '../workspace/repoAddressAsString';
+import {RepoAddress} from '../workspace/types';
+import {workspacePipelinePath} from '../workspace/workspacePath';
+
+import {doneStatuses} from './RunStatuses';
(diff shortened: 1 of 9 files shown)


hellendag:
* master
  * **PR #9451** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9451" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9451?utm_source=stack-comment).

js_modules/dagit/packages/core/src/runs/QueryfulRunTimeline.tsx
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    const earliest = jobs.reduce((accum, job) => {
+      const startTimes = job.runs.map((job) => job.startTime);
+      return {...accum, [job.key]: Math.min(...startTimes)};
+    }, {} as {[jobKey: string]: number});
+
+    setJobsWithRuns(jobs.sort((a, b) => earliest[a.key] - earliest[b.key]));
+    setLoadedWindow(hourWindow);
+  }, [data, start, end, loading, jobsByKey, hourWindow]);
```
bengotow, line 171: This looks good to me, but technically this is more of a useMemo than a useEffect because these setters are only called here. I think making it a useEffect just means there's one extra render after the params change and before the state is updated (so it doesn't matter much)...

bengotow (approved):
This looks good to me!

js_modules/dagit/packages/core/src/runs/QueryfulRunTimeline.tsx
hellendag: I think you're right -- this is just copied from existing code, so I didn't do much with this, but we can come back and fix it up a bit more.

PR #9460, opened 2022-08-23
remove graph-backed asset note

sryza:
### Summary & Motivation

I wasn't able to figure out the meaning of this note, and I think it might have more to do with multi-assets than graph-backed assets? I'm also very open to rewording it if it's important.

-

### DIFF
--- a/docs/content/concepts/assets/software-defined-assets.mdx
+++ b/docs/content/concepts/assets/software-defined-assets.mdx
@@ -237,8 +237,6 @@ def store_slack_files_in_sql():
 graph_asset = AssetsDefinition.from_graph(store_slack_files_in_sql)
 ```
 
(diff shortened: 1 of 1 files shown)


clairelin135:
This message is here because we don't technically allows graph-backed asset subsetting yet. So, when a graph-backed asset outputs multiple assets, if you do an asset selection in Dagit's UI (or via python) we force you to select all of the outputted assets from the graph.

That being said, I'm putting out a PR today that will enable graph-backed asset subsetting. I think we can hold off on deleting this message until that PR is merged.

erinkcochran87:
Related to https://github.com/dagster-io/dagster/pull/9466

sryza:
@clairelin135 would it make sense to remove this note now?

clairelin135:
@sryza lets do it!

clairelin135 (approved):

PR #9464, opened 2022-08-23
include AssetOut in apidoc

sryza:


erinkcochran87:
@sryza What exactly should I be looking for in the Vercel preview? Specifically, what's changed? The API docs for assets look the same in the preview as they do for the live 1.0.4 docs, and I'm not sure if that's expected.

sryza:
I've actually wondered about how this works. @yuhan and/or @smackesey - would you be able to weigh in? Should I be running `make apidoc-build` every time I post a PR that affects API docs?

yuhan:
> Should I be running make apidoc-build every time I post a PR that affects API docs?
yes

sryza:
@yuhan what if I update the docstring of a public function/method or add a public function/method?

yuhan:
if you want it to show up in "master" version of the docs, then yes. if not, you don't have to because the weekly release will build the apidoc as part of the docs release process, i.e. the released docs will have up-to-date api docs data.

sryza:
The preview should now reflect the changes!

erinkcochran87 (approved):

PR #9465, opened 2022-08-23
[dagit] Fix asset graph rendering in Firefox, Safari

bengotow:
### Summary & Motivation

This PR fixes the issue shown in the screenshot below, and makes some tiny adjustments to the heights of nodes to fit them in the group boxes better.

This was happening because 1) Safari does not allow a `foreignObject` to be the relative container for `absolute` positioning, and 2) it does not honor "margin" properties on the top level of elements within a `foreignObject`. This unfortunately means that we needed to wrap AssetNode in one more `<div>` to get the correct rendering.

image

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
@@ -44,88 +44,90 @@ export const AssetNode: React.FC<{
   const {lastMaterialization, computeStatus} = liveData || MISSING_LIVE_DATA;
 
   return (
-    <AssetNodeContainer $selected={selected}>
-      <AssetNodeBox $selected={selected}>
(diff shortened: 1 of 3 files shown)


js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
```diff
@@ -200,17 +204,20 @@ const BoxColors = {
   Stats: 'rgba(236, 236, 248, 1)',
 };
 
+const AssetInsetForHoverEffect = styled.div`
+  padding-top: 10px;
+  padding-right: 4px;
+  padding-left: 4px;
+  padding-bottom: 2px;
```
hellendag, line 211: Combine?

hellendag (approved):

PR #9466, opened 2022-08-23
Subset Graph-Backed Asset

clairelin135:
Enables graph-backed asset subsetting.

When an asset selection is provided that is a partial selection of an `AssetsDefinition`, use the `dependency_node_handles_by_asset_key` field from the asset layer of the original job to determine the op selection necessary to materialize the selected assets.

Then, subset the job to contain only the op selection necessary. It is necessary to pass the whole job into this function so we can differentiate between instances of a graph that is reused to build multiple assets within the same asset job.

A follow up here is to add documentation for this feature.


python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets.py
```diff
@@ -673,3 +679,158 @@ def foo():
...
+            asset_selection=[AssetKey("one"), AssetKey("upstream_1"), AssetKey("upstream_2")],
+        )
+        materialization_planned = list(
+            instance.get_event_records(
+                EventRecordsFilter(DagsterEventType.ASSET_MATERIALIZATION_PLANNED)
+            )
+        )
+        assert len(materialization_planned) == 3
+        step_keys = get_step_keys_from_run(instance)
+        assert set(step_keys) == set(["my_graph.bar_1", "upstream_1", "upstream_2"])
```
OwenKephart, line 806: For this test, I'm curious about what would happen if the asset selection was just "one" and "upstream_1". I think that would better match the name of "test_input_subsetting", as in this case only one of the input to the graph would be used.

Fine with this test sticking around, just would also like to see that other behavior under test.

python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets.py
```diff
@@ -673,3 +679,158 @@ def foo():
...
+    )
+
+    with instance_for_test() as instance:
+        asset_job.execute_in_process(instance=instance, asset_selection=[AssetKey("one")])
+        materialization_planned = list(
+            instance.get_event_records(
+                EventRecordsFilter(DagsterEventType.ASSET_MATERIALIZATION_PLANNED)
+            )
+        )
+        assert len(materialization_planned) == 1
```
OwenKephart, line 754: bit of a nitpick, but in all of these tests, do you mind adding a check for the actual number of asset materialization events (in addition to the number of planned events)?

This one is a bit of a special case, as it will emit more materializations than planned, but the others should match up I believe.

OwenKephart (changes requested):
Very cool stuff! I think it should be possible to rework this so that asset_selection_data does not need to be wired through, which would make this feel a bit cleaner (see comments for more details), but overall I'm excited about the change :)

clairelin135:
Thanks @OwenKephart for the feedback! I've addressed your comments and updated the tests, so I think it's ready for you to take another look.

I was able to remove threading the `_asset_selection_data` argument. But getting the mapping of asset key to dependent node handles (via `asset_key_to_dep_node_handles`) ended up being pretty tricky. I played around with it for a couple hours, and I ultimately found that I would need to thread an additional arg through that code denoting that the graph_def was an asset node versus the graph underlying the job that contained top level asset nodes. The threading ended up being pretty complicated and convoluted the code implementation.

I found that it was way easier to add a helper function that wrapped the graph node with a top-level graph wrapper, then passed that graph wrapper into `asset_key_to_dep_node_handles`. Let me know how this looks!

OwenKephart (approved):
much better! definitely a complicated feature but happy to see it implemented w/o any extra top level changes :)

PR #9475, opened 2022-08-24
[Docs] Use `asPath` from pageProps if available instead of `router`

salazarm:
### Summary & Motivation

Currently we're setting the canonical URL using the `asPath` from the `router`. The problem with this is that the router is not available during Server Side Rendering (SSR) so initially the canonical url is always "/" and only after a second render pass on the client does the canonical URL get updated correctly. 

To fix the issue this PR grabs `asPath` from `pageProps` if its available and falls back to the router if not.

Screen Shot 2022-08-24 at 2 15 06 PM

view-source:https://dagster-ml54zkbn1-elementl.vercel.app/getting-started


yuhan (approved):
🙏

PR #9479, opened 2022-08-24
[dagit] Add elapsed time tooltips to log row timestamps

### LINKED ISSUE #9209: Display relative time in event logs
**User reported in slack:**

> Just want to suggest that it would be a nice feature if the user could switch from timestamp to relative time in structured event logs in run view, so here in the screenshot below first event time would be 00:00:00.000 and other will be related to it.
It would simplify estimating time delta between some events

>  often look to the timestamp to check where are slow parts of a job (relative time use case) and sometimes when some event occurs to match some other events in external services (timestamp use case) so timestamp on hover is looking more preferable option for me

image

hellendag:
### Summary & Motivation

Resolves #9209.

Add a tooltip to log timestamps to show the elapsed time since the start of the run, and if applicable, the elapsed time since the start of the relevant step.

I did some refactoring of the time utilities here. Tests are included.

Screen Shot 2022-08-24 at 2 00 27 PM

View a run, hover over timestamps. Verify that the tooltips render correctly, including "negative" times that are for logs that occur prior to the start of the run.

### DIFF
--- a/.buildkite/dagster-buildkite/dagster_buildkite/steps/dagit_ui.py
+++ b/.buildkite/dagster-buildkite/dagster_buildkite/steps/dagit_ui.py
@@ -11,6 +11,10 @@ def build_dagit_ui_steps() -> List[CommandStep]:
         .run(
             "cd js_modules/dagit",
             "pip install -U virtualenv",
+            # Explicitly install Node 16.x because BK is otherwise running 12.x.
+            # Todo: Fix BK images to use newer Node versions, remove this.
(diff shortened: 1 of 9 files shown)


hellendag:
* master
  * **PR #9479** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9479" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9479?utm_source=stack-comment).

prha (approved):

Issue #7109, opened 2022-03-17
Graph In/Out Mapping Type Issues


alangenfeld:
`CompositeSolid` enforced that types had to be aligned along a mapping. Things like execution plan input resolution were set-up against this assumption. 

https://github.com/dagster-io/dagster/pull/4876 dropped `dagster_type`, effectively removing the enforcement that implicit assumptions were coded against.

This leads to bugs when users happen to violate the implicit assumptions. 

@dpeng817 and I chatted about it a bit today, and believe we need:
* validation that all destinations for input mapping are of the same type
  * special consideration for `Nothing` / `Any` 
* validation when type-hint information is folded in (https://github.com/dagster-io/dagster/issues/6499)

sryza:
Why do all destinations for input mapping need to be the same type? I'm imagining a situation where the same outputs gets routed to inputs of two different ops. The output has 20 columns, but each of the ops only needs 5 columns to be present, and those 5 columns are different between the two ops.

alangenfeld:
> Why do all destinations for input mapping need to be the same type?  ...   output has 20 columns, but each of the ops only needs 5 columns to be present, and those 5 columns are different between the two ops

Thats a good point. Probably makes sense for the mappings to follow similar enforcement to our input/output type alignment rules aka basically nothing. 

Will update the issue.

alangenfeld:
> GraphIn does not support dagster_type but you can set one via type hints
>  I believe only allowing the information via typehints was unlikely an explicit design choice.

Some discussion on this from https://github.com/dagster-io/dagster/pull/7741, summarized:
* we should remove it, but need to be careful not to regress type loader based config value setting
* could try to follow the mappings and infer the type from the destination
* tricky if mixed destination types, especially on accident, maybe want to to be able to explicit set (so maybe don't remove?)

PR #9489, opened 2022-08-25
allow Nothing annotations on @graph

alangenfeld:
as shown in added test case - the specific case cant upgrade to 1.X since we have not yet gotten to a sane end state with graph mappings as discussed in https://github.com/dagster-io/dagster/issues/7109

This stop gap allows users actively trying to upgrade to 1.X who are hitting this issue to proceed while we work towards a better final state. 

added test

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/solid_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/solid_decorator.py
@@ -395,21 +395,28 @@ def resolve_checked_solid_fn_inputs(
     for input_def in explicit_input_defs:
         if input_def.name in inferred_props:
             # combine any information missing on the explicit def that can be inferred
-            input_defs.append(
-                input_def.combine_with_inferred(
-                    inferred_props[input_def.name], decorator_name=decorator_name
-                )
-            )
+            input_defs.append(input_def.combine_with_inferred(inferred_props[input_def.name]))
         else:
             # pass through those that don't have any inference info, such as Nothing type inputs
             input_defs.append(input_def)
 
     # build defs from the inferred props for those without explicit entries
-    input_defs.extend(
-        InputDefinition.create_from_inferred(inferred, decorator_name=decorator_name)
+    inferred_input_defs = [
+        InputDefinition.create_from_inferred(inferred)
         for inferred in inferred_props.values()
         if inferred.name in inputs_to_infer
-    )
+    ]
+
+    if exclude_nothing:
+        for in_def in inferred_input_defs:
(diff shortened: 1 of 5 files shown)


alangenfeld:
* master
  * **PR #9489** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9489" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9489?utm_source=stack-comment).

sryza (commented):
Will mypy not complain on this?

dpeng817 (commented):
@sryza where exactly do you think mypy would complain? As you pass the parameter of the graph to the op?

dpeng817 (approved):

sryza:
Goals-wise, the way I've been thinking about this is that we should escape a world where GraphIns and GraphOuts have Dagster Types, via type annotation or implicit construction.

If I understand correctly, the issue is that this causes issues with type-based input-loading.  Without yet having thought about all the edge cases, the way I should think this should work is basically that:
- If all the leaf inputs that the graph input is mapped to have the same type, then the graph input automatically gets that type.
- If all the leaf inputs that the graph input aren't mapped to have the same type, then we have to pick one of:
  - Default to Any
  - Default to None
  - Error at graph-construction time

Does this line up with the way you're thinking about it?  Is the idea here that the above is a pain to implement and the implementation here is a stopgap?

sryza:
> @sryza where exactly do you think mypy would complain? As you pass the parameter of the graph to the op?

Annotating the input param with `Nothing`, which isn't a type.

alangenfeld:
> Does this line up with the way you're thinking about it? 
 
 ya i think so - thats inline with whats discussed in the linked issue in the summary.

 > Is the idea here that the above is a pain to implement and the implementation here is a stopgap?
 
 Yea this is a stopgap to allow users actively trying to upgrade to 1.0 who are blocked by this problem.

alangenfeld:
> Will mypy not complain on this?

Yep updated test case to show that `None` can be used if passing `mypy` is desired

sryza:
Got it. I'm not in love with this approach because it endorses type annotation on graph params, which is something that we'll be stuck with for a while. But will trust your judgment if you think the alternative is too heavy a lift and we need to get something working ASAP?

alangenfeld:
> endorses type annotation on graph params, which is something that we'll be stuck with for a while.

Since we already read and respect the type annotations on `@graph` and it effects config schema, I don't think this PR makes the situation worse. Correcting the existing behavior will have to be done carefully.

Issue #9360, opened 2022-08-12
threaded sensor daemon run key race condition


alangenfeld:
```
dagster._check.CheckError: Invariant failed.

Stack Trace:
  File "/app/.runtimes/python-3.10.0-2/lib/python3.10/site-packages/dagster/_core/instance/__init__.py", line 1682, in submit_run
    submitted_run = self._run_coordinator.submit_run(
,  File "/app/.runtimes/python-3.10.0-2/lib/python3.10/site-packages/dagster/_core/run_coordinator/queued_run_coordinator.py", line 121, in submit_run
    check.invariant(pipeline_run.status == PipelineRunStatus.NOT_STARTED)
,  File "/app/.runtimes/python-3.10.0-2/lib/python3.10/site-packages/dagster/_check/__init__.py", line 1472, in invariant
    raise CheckError("Invariant failed.")
```

> And indeed, looking at my sensor ticks, I can see that the same job is being dispatched from two subsequent ticks.

* `QueuedRunCoordinator` `dequeue_interval_seconds` `1`s
* sensor with `minimal_interval_tick` `1`s
* threading enabled, `8` workers

https://dagster.slack.com/archives/C01U954MEER/p1660105991117999

alangenfeld:
How this I think this happened:
* two threads overlap for sensor with such low `minimal_interval_tick`
* the same `run_key` is derived in both
* `_get_or_create_sensor_run` in one thread finds the already created run and uses it instead
  - (I believe this logic that was created to handle recovering from failure mid tick evaluation)
*  both threads submit the same run_Id
* second one fails invariant check on submit

https://github.com/dagster-io/dagster/blame/master/python_modules/dagster/dagster/_daemon/sensor.py#L703-L737

alangenfeld:
some options:
* just not invariant in the coordinator and treat duplicate submissions as acceptable, second submit no-ops. 
* use a locked datastructure to coordinate between threads that have sensor ticks that evaluate to same `run_key`

@prha

prha:
If we have plans to horizontally scale daemons, we should lean towards the former?

johannkm:
Yeah I guess I'm a fan of the former. #2 sounds like trouble

PR #9490, opened 2022-08-25
relax run coordinator invariant to handle threaded sensor race conditions

prha:
### Summary & Motivation
https://github.com/dagster-io/dagster/issues/9360

BK


alangenfeld (approved):
nit:  which is a no-op -> "skipping launch" / "skipping enqueue"

PR #9497, opened 2022-08-25
[dagit] Refactor Runs list tabs

hellendag:
### Summary & Motivation

Refactor the Runs page a bit in preparation for adding some more features to it. The visuals and interactions are unchanged.

- Move `RunListTabs` out into its own component.
  - Build URLs for the individual tabs instead of tracking state for `status` values.
- Make "Scheduled" its own page that consumes `RunListTabs`. It was basically already this, but we were jamming it into the existing `RunsRoot` and just treating it separately there.
  - Add a route for this, `/instance/runs/scheduled`

View Runs page.

- Click on each of the tabs, verify correct navigation behavior with status values applied to filter.
- Verify that the statuses are shown as static, as before.
- Click "Scheduled", verify rendering and behavior.
- While viewing "Done" runs, add a tag to the filter. Click Queued, All, In Progress, and verify that the tag remains in the filter during navigation.


hellendag:
* master
  * **PR #9497** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9497" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9497?utm_source=stack-comment).

js_modules/dagit/packages/core/src/hooks/useQueryPersistedState.tsx
```diff
@@ -97,7 +97,7 @@ export function useQueryPersistedState<T extends QueryPersistedDataType>(
       }
 
       currentQueryString = next;
-      history.replace(`${location.pathname}?${qs.stringify(next, {arrayFormat: 'brackets'})}`);
+      history.push(`${location.pathname}?${qs.stringify(next, {arrayFormat: 'brackets'})}`);
```
hellendag, line 100: Not sure why this is currently `replace` instead of `push`. I think it would be kind of nice to be able to navigate back to previous filter states via the Back button.

js_modules/dagit/packages/core/src/runs/RunListTabs.tsx
```diff
@@ -0,0 +1,102 @@
...
+export const RunListTabs = React.memo(() => {
+  const [filterTokens] = useQueryPersistedRunFilters();
+  const runsFilter = runsFilterForSearchTokens(filterTokens);
+
+  const {data} = useQuery<RunTabsCountQuery, RunTabsCountQueryVariables>(RUN_TABS_COUNT_QUERY, {
+    variables: {
+      queuedFilter: {...runsFilter, statuses: Array.from(queuedStatuses)},
+      inProgressFilter: {...runsFilter, statuses: Array.from(inProgressStatuses)},
+    },
+  });
```
hellendag, line 27: Tab counts are now maintained in the tabs component, instead of in `RunsRoot`.

js_modules/dagit/packages/core/src/hooks/useQueryPersistedState.tsx
bengotow: Oh hmm I'm not sure about this being the default - I think we have a lot of places where an input field reads/writes the state, so this would yield a history.push on every keystroke. For non-text inputs this seems like a reasonable choice though, I wonder if we could get fancy and stack the changes within a certain timeframe into one history item...

bengotow (approved):
This looks great! Wrapping all the tabs related stuff into a component really cleans up RunsRoot...

js_modules/dagit/packages/core/src/hooks/useQueryPersistedState.tsx
hellendag: Good point. I'll change it back for now and we can investigate further.

PR #9498, opened 2022-08-25
Improve docs for asset sensors

sryza:
- Add an example to the decorator docstring
- Include type annotations in the snippet on the concepts page


prha (approved):

python_modules/dagster/dagster/_core/definitions/decorators/sensor_decorator.py
```diff
@@ -120,6 +120,29 @@ def asset_sensor(
...
+    Example:
+
+        .. code-block:: python
+
+            from dagster import AssetKey, EventLogEntry, SensorEvaluationContext, asset_sensor
+
+
+            @asset_sensor(asset_key=AssetKey("my_table"), job=my_job)
+            def my_asset_sensor(context: SensorEvaluationContext, asset_event: EventLogEntry):
+                yield RunRequest(
```
prha, line 134: what do you think about returning instead of yielding?

Hoping to move away from the generator syntax since it misleads folks in terms of the execution pattern.

sryza: sounds fine to me

PR #9502, opened 2022-08-26
Update changelog version

johannkm:
Will just need to cherry pick in to release branch


johannkm:
We could mention 1.0.5 - skipped, but it looks like we haven't in the past

PR #9503, opened 2022-08-26
Add 0.16.x to the changelog

johannkm:
Since we lump core and libs into the same changelog, I think we need to include both versions

Will need to modify each release branch :/


sryza (approved):
Good thinking

erinkcochran87 (approved):

PR #9511, opened 2022-08-26
CompositePartitionsDefinition

### LINKED ISSUE #4591: multi-dimensional partitioning
Consider a job that, every day, processes data from N different data sources and updates partitions in incremental tables that correspond to each of those data sources.  So there are N data sources, N target tables, and N partitions for each date (one in each target table), and N runs for each date.

One way to implement this would be to have N jobs - one for each data source.  However, if there are a lot of data sources, this might make Dagit crowded.  And in some situations, you'd want to view a run history for all the runs across all jobs, together.

In many situations, it would be desirable to operate over a single data source.  I.e. if you only care about a particular data source, you'd want to view a partition-step matrix that just includes the jobs for that data source, and effectively includes a partition for each date.

You might also want to launch a backfill over all the dates for a single data source.

One way for us to accomplish this would be to explicitly model multi-dimensional partition sets.  I.e. the partition set could have two dimensions: source name and date.  Then the UI could support slicing by one of the dimensions - i.e. filtering to a single source name or date.

This would be similar to @kinghuang's proposal here: https://github.com/dagster-io/dagster/discussions/3904.

User requests:
- https://dagster.slack.com/archives/C01U954MEER/p1658157470946549
- https://dagster.slack.com/archives/C023MS97478/p1658173927153479
- https://dagster.slack.com/archives/C01U954MEER/p1661331179772159?thread_ts=1660740883.950979&cid=C01U954MEER
- https://dagster.slack.com/archives/C02T7SJQ849/p1661952155521029
- https://github.com/dagster-io/dagster/discussions/9703
- https://dagster.slack.com/archives/C01U954MEER/p1664815078042949?thread_ts=1664792151.002079&cid=C01U954MEER

sryza:
### Summary & Motivation

Fixes #4591.

Open questions and reservations:
- Is this the right name? What if we want to add something in the future that's union of two PartitionsDefinitions, instead of the cross-product?
- How do we feel this approach where we keep partition keys as strings and stuff multiple dimensions inside them with a delimiter? An alternative would be to turn partition keys into tuples, although that would require touching lots and lots of code.
- Assuming we do want to go with the string approach, this would break if someone already has the delimiter inside their partition key.

Future work:
- Make it possible to slice along a particular axis when viewing partition status or launching backfills.

Loaded this in Dagit:
```
from dagster import (
    StaticPartitionsDefinition,
    DailyPartitionsDefinition,
    asset,
)
from dagster._core.definitions.composite_partitions import CompositePartitionsDefinition


partitions_def = CompositePartitionsDefinition(
    [
        StaticPartitionsDefinition(["a", "b", "c"]),
        DailyPartitionsDefinition(start_date="2022-08-01"),
    ]
)


@asset(partitions_def=partitions_def)
def asset1():
    ...
```

and launched a materialization.

image

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
@@ -16,6 +16,7 @@ from dagster._core.host_representation import ExternalRepository, RepositoryLoca
 from dagster._core.host_representation.external import ExternalPipeline
 from dagster._core.host_representation.external_data import (
     ExternalAssetNode,
+    ExternalCompositePartitionsDefinitionData,
     ExternalStaticPartitionsDefinitionData,
     ExternalTimeWindowPartitionsDefinitionData,
 )
@@ -229,8 +230,14 @@ class GrapheneAssetNode(graphene.ObjectType):
         partitions_def_data = self._external_asset_node.partitions_def_data
         if partitions_def_data:
             if isinstance(
-                partitions_def_data, ExternalStaticPartitionsDefinitionData
-            ) or isinstance(partitions_def_data, ExternalTimeWindowPartitionsDefinitionData):
+                partitions_def_data,
+                (
+                    ExternalStaticPartitionsDefinitionData,
+                    ExternalTimeWindowPartitionsDefinitionData,
+                    ExternalCompositePartitionsDefinitionData,
+                    ExternalCompositePartitionsDefinitionData,
+                ),
+            ):
                 return [
                     partition.name
                     for partition in partitions_def_data.get_partitions_definition().get_partitions()
--- a/python_modules/dagster/dagster/_core/definitions/composite_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/composite_partitions.py
@@ -1,0 +1,39 @@
+import itertools
+from datetime import datetime
+from typing import Optional, Sequence
+
+from .partition import Partition, PartitionsDefinition
+
+
+class CompositePartitionsDefinition(PartitionsDefinition):
+    """The set of partitions is the cross product of partitions in the inner partitions
+    definitions"""
+
+    def __init__(self, partitions_defs: Sequence[PartitionsDefinition]):
+        self._partitions_defs = partitions_defs
+
+    @property
+    def partitions_defs(self):
+        return self._partitions_defs
+
+    def get_partitions(self, current_time: Optional[datetime] = None) -> Sequence[Partition]:
+        partition_sequences = [
+            partitions_def.get_partitions(current_time=current_time)
+            for partitions_def in self._partitions_defs
+        ]
+        return [
+            Partition(
+                value=tuple(partition.value for partition in partitions_tuple),
+                name="|".join(partition.name for partition in partitions_tuple),
+            )
+            for partitions_tuple in itertools.product(*partition_sequences)
+        ]
+
+    def __eq__(self, other):
+        return (
+            isinstance(other, CompositePartitionsDefinition)
+            and self.partitions_defs == other.partitions_defs
+        )
+
+    def __hash__(self):
+        return hash(tuple(self.partitions_defs))
--- a/python_modules/dagster/dagster_tests/core_tests/definitions_tests/test_composite_partitions.py
+++ b/python_modules/dagster/dagster_tests/core_tests/definitions_tests/test_composite_partitions.py
@@ -1,0 +1,39 @@
+from datetime import datetime
+
+from dagster import DailyPartitionsDefinition, StaticPartitionsDefinition
+from dagster._core.definitions.composite_partitions import CompositePartitionsDefinition
+
+DATE_FORMAT = "%Y-%m-%d"
+
+
+def test_composite_static_partitions():
+    partitions1 = StaticPartitionsDefinition(["a", "b", "c"])
+    partitions2 = StaticPartitionsDefinition(["x", "y", "z"])
+    composite = CompositePartitionsDefinition([partitions1, partitions2])
+    assert composite.get_partition_keys() == [
+        "a|x",
+        "a|y",
+        "a|z",
+        "b|x",
+        "b|y",
+        "b|z",
+        "c|x",
+        "c|y",
+        "c|z",
+    ]
+
+
+def test_composite_time_window_static_partitions():
+    time_window_partitions = DailyPartitionsDefinition(start_date="2021-05-05")
+    static_partitions = StaticPartitionsDefinition(["a", "b", "c"])
+    composite = CompositePartitionsDefinition([time_window_partitions, static_partitions])
+    assert composite.get_partition_keys(
+        current_time=datetime.strptime("2021-05-07", DATE_FORMAT)
+    ) == [
+        "2021-05-05|a",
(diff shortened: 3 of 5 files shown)


python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
```diff
@@ -229,8 +230,14 @@ def get_partition_keys(self) -> Sequence[str]:
...
         if partitions_def_data:
             if isinstance(
-                partitions_def_data, ExternalStaticPartitionsDefinitionData
-            ) or isinstance(partitions_def_data, ExternalTimeWindowPartitionsDefinitionData):
+                partitions_def_data,
+                (
+                    ExternalStaticPartitionsDefinitionData,
+                    ExternalTimeWindowPartitionsDefinitionData,
+                    ExternalCompositePartitionsDefinitionData,
+                    ExternalCompositePartitionsDefinitionData,
```
clairelin135, line 238: nit: `ExternalCompositePartitionsDefinitionData` listed twice

python_modules/dagster/dagster_tests/core_tests/definitions_tests/test_composite_partitions.py
```diff
@@ -0,0 +1,39 @@
+from datetime import datetime
+
+from dagster import DailyPartitionsDefinition, StaticPartitionsDefinition
+from dagster._core.definitions.composite_partitions import CompositePartitionsDefinition
+
+DATE_FORMAT = "%Y-%m-%d"
+
+
+def test_composite_static_partitions():
```
clairelin135, line 9: Seems like its possible in this implementation to create a composite partitions definition with other composite partitions definitions. Probably worth putting this under test

python_modules/dagster/dagster/_core/definitions/composite_partitions.py
```diff
@@ -0,0 +1,39 @@
...
+    definitions"""
+
+    def __init__(self, partitions_defs: Sequence[PartitionsDefinition]):
+        self._partitions_defs = partitions_defs
+
+    @property
+    def partitions_defs(self):
+        return self._partitions_defs
+
+    def get_partitions(self, current_time: Optional[datetime] = None) -> Sequence[Partition]:
```
clairelin135, line 19: I can't think of a reason why someone would want to do this, but do you think it's possible people might want to specify a different time for each partitions definition?

clairelin135 (commented):
If we go down this route, people might ask for other composite partition types like:
- the sum of two partitions definitions
- all matching keys among two partitions

which makes me wonder if we should think of a more specific name, like `CrossProductPartitionsDefinition`? And if we want to allow users to have greater granularity in which composite partitions are created from a set of partitions definitions, should we just improve support on dynamic partitions definitions to allow supplying a set of partitions?

If we did change the partition to being represented as a tuple, in order to filter by one axis of partitions, I think we'd have to deserialize all the relevant event logs to filter them. Not sure if that's any better than using something like `db.contains_`

sryza:
@clairelin135 

> which makes me wonder if we should think of a more specific name, like CrossProductPartitionsDefinition?

I think you're right - will change.

> if we want to allow users to have greater granularity in which composite partitions are created from a set of partitions definitions, should we just improve support on dynamic partitions definitions to allow supplying a set of partitions?

The issue with dynamic partitions definitions is that they require an extra round trip to user code when loading a page, which can be slow.

> If we did change the partition to being represented as a tuple, in order to filter by one axis of partitions, I think we'd have to deserialize all the relevant event logs to filter them. Not sure if that's any better than using something like db.contains_

Mmm, good point, so this a strike in favor of keeping them as strings. If we wanted to mitigate this, we have a few options, not all of which are necessarily good ideas:
- They could be strings in storage, but tuples in Python / GraphQL / frontend
- We could serialize the tuple we're looking for and then scan the DB for partitions that have that serialized value as their key

clairelin135:
@sryza I think displaying and exposing the partition keys as tuples will make them more readable and tractable. Though, I don't think we can completely hide that the partition keys have an underlying string implementation. I wouldn't be surprised if some users directly fetched event log records and found that the multi-dimensional partition was a string.

We could serialize the tuple, but then we wouldn't be able to search for a specific dimension of the partition key. I think this is another point in favor of storing as a plaintext string so we can search for a specific dimension (e.g. for `CompositePartitionsDefinition(StaticPartitionsDefinition([...]), MonthlyPartitionsDefinition(...)` getting all partitions with monthly partition December)

prha:
I actually prefer `Composite` to `CrossProduct` (which maybe feels too mathy).

If we wanted to distinguish between the cross-product and the sum (which I think would be pretty rare), I think we could offer `UnionPartitionsDef` and `DifferencePartitionsDef`.  But also, I'm fine making folks eject and write their own anyway.  The Union/Difference defs are all single-dimension anyway.... you're just applying an operator to filter.

The tricky stuff is making sure we have the right API for dealing with multiple dimensions. I think this implementation which just flattens everything out is just solving the one of the problems with these multi-dimensional partitions, which is defining the full global set of partitions (arguably the smallest).  

I'm with @clairelin135 in that I think the trickier problem is coordinating with the UI / storage to display things like status (e.g. how are we selecting / fetching / communicating the status of the `2022-08-29` slice of the partition set?). How would we schedule a backfill of just the `a` partitions?  To launch a single run, do I still have to select from an exploded list of partitions?   Or can I do it by dimension?

Random implementation thought: w.r.t searching for particular dimensions, I wonder if we can start using dynamic tag keys (which incorporate the dimension name) for doing some sort of search.

sryza:
I just had a chat with @prha about this.  Here were my main takeaways:
- These are the special things you might want to be able to do in Dagit with multi-dimensional partitions.
  - When viewing partition status, fix the value for some of the partition dimensions.
  - When launching a backfill, fix the value for some of the partition dimensions.
  - When launching a run from the launchpad, fix the value for all of the partition dimensions.
- Giving users this capability might encourage them to have much larger partition sets than they currently do.  Up to now, we've done tests on the order of 10k partitions = a partition for every hour for a few years.  If someone throws in another dimension, this could easily balloon to hundreds of thousands or millions of partitions.  We should make sure we can handle this.

prha:
some naming alternatives:
* MultiPartitionsDef
* MultidimensionalPartitionsDef
* CompoundPartitionsDef

sryza:
> But also, I'm fine making folks eject and write their own anyway.

I think this isn't really an option because we need all PartitionsDefinition to be able to run in the host process so we're not calling out to user code on page load etc.

jtv8:
Happy to see this work happening - this would definitely solve a current need for us!

> Make it possible to slice along a particular axis when viewing partition status or launching backfills.

I can imagine you might want to be able to do this in a `PartitionMapping` as part of an asset definition as well. For example, if you have an asset that creates aggregates over one of the dimensions of an upstream asset it's important to know which partitions in the downstream asset to recalculate.

A few ideas (these are all equivalent definitions that mean "roll up the first dimension"):

```
mapping_def = CompositePartitionMapping(
    [
        AllPartitionMapping(),
        TimeWindowPartitionMapping(),
    ]
)
```

```
mapping_def = AggregatePartitionMapping(axes=[0])
```

```
mapping_def = GroupByPartitionMapping(axes=[1])
```

PR #9518, opened 2022-08-28
accrete metadata with multiple calls to add_output_metadata

### LINKED ISSUE #7104: enable multiple calls to `add_output_metadata`  inside an IOManager to combine metadata
Currently, successive calls to `add_output_metadata` will overwrite each other. It would be helpful to merge them.

sryza:
### Summary & Motivation

Fixes #7104

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/context/output.py
+++ b/python_modules/dagster/dagster/_core/execution/context/output.py
@@ -26,7 +26,7 @@ from dagster._core.definitions.events import (
 from dagster._core.definitions.metadata import RawMetadataValue
 from dagster._core.definitions.partition_key_range import PartitionKeyRange
 from dagster._core.definitions.time_window_partitions import TimeWindow
-from dagster._core.errors import DagsterInvariantViolationError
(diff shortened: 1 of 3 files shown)


dpeng817 (commented):
What happens when the metadata entries overlap for a given key? For example, I specify `"foo": "bar` and then `"foo": "baz`. I believe it will allow that to pass through, but what happens if you try to view that metadata later as a dictionary of metadata? Will one be overwritten? In general, we want the metadata exposed to users to be available in dictionary form I think, not as MetadataEntry, so I think that matters

python_modules/dagster/dagster/_core/execution/context/output.py
```diff
@@ -665,7 +665,10 @@ def handle_output(self, context, obj):
         """
         from dagster._core.definitions.metadata import normalize_metadata
 
-        self._metadata_entries = normalize_metadata(metadata, [])
+        self._metadata_entries = [
```
schrockn, line 668: ah i was wondering about this.

schrockn (approved):
@dpeng817 great catch

Issue #9230, opened 2022-08-05
If you click on a SourceAsset in Dagit, opening the sidebar, then click "Materialize", you'll get a GQL error.


OwenKephart:
### Dagster version

0.15.8

### What's the issue?

Slack: https://dagster.slack.com/archives/C01U954MEER/p1659658372336479?thread_ts=1659655298.487849&cid=C01U954MEER

Basically the title. A source asset can be selected, but this does not disable the "Materialize" button. When you click that button, the target asset that Dagit tries to materialize is the source asset, which is not possible to materialize (as it's a source asset).

The materialize button should probably be greyed out when a source asset is targeted.

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

Issue #9520, opened 2022-08-28
"Asset must be part of at least one job" error in dagit when selecting a Dbt source


mingfang:
### Dagster version

dagit, version 1.0.6

### What's the issue?

"Asset must be part of at least one job" error in dagit when selecting a Dbt source.

### What did you expect to happen?

I should see details of the dbt sources.
Sources in dbt should behave like `SourceAssets` and definition they are not materialized by any jobs.

### How to reproduce?

To recreate the problem, change this line
https://github.com/dagster-io/dagster/blob/master/examples/assets_dbt_python/assets_dbt_python/repository.py#L51
to remove `raw_data_assets` like this 
`dbt_assets + forecasting_assets,`

Then in dagit `Global Asset Lineage` graph, try clicking `duckdb / raw_data / orders` and it results in this error
```
Operation name: SidebarAssetQuery

Message: Invariant failed. Description: Asset must be part of at least one job

Path: ["assetNodeOrError","type"]

Locations: [{"line":114,"column":3}]
```

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

sryza:
@OwenKephart mind taking a look?

mingfang:
@sryza @OwenKephart 
Thanks for looking into this issue.
I think this is a general issue with all `SourceAssets`.
You can see the problem by running this example
https://github.com/dagster-io/dagster/blob/master/docs/content/concepts/assets/software-defined-assets.mdx?plain=1#L126

mingfang:
confirm fixed in v1.0.7

PR #9526, opened 2022-08-29
Allow fetching fields of a Source Asset without Dagit throwing exceptions

bengotow:
### Summary & Motivation
This resolves the issue described in https://github.com/dagster-io/dagster/issues/9230

I'm not wild about these extra checks - it seems like a lot of the AssetNode resolvers call `get_external_pipeline` and those raise exceptions when called on a source asset. This fixes the two callsites that were in use / causing problems in Dagit and I didn't see more, but there could be others.


clairelin135 (approved):
Thanks for making this change! I feel like we've run into similar issues in the past where errors occur when nonexistent properties are fetched from source assets. Ideally we can refactor the code to avoid this--I'll think about this some more

mingfang:
Hope this PR will also fix this issue https://github.com/dagster-io/dagster/issues/9520

Issue #9224, opened 2022-08-04
Wonky op graph displays when creating AssetsDefinitions from graphs


OwenKephart:
### Dagster version

0.15.8

### What's the issue?

When you create an `AssetsDefinition.from_graph`, and view the op graph in Dagit, the connections are incorrectly displayed. Take this example repository:

```
Users/owen/.zshrc:ulimit:73: setrlimit failed: invalid argument                            │from dagster import run_status_sensor, DagsterRunStatus, bu
# Introduction
from dagster import AssetsDefinition, op, graph, asset, repository, define_asset_job

@op
def bar(inp):
    return inp

@op
def combine(x):
    return x

@graph
def ag(inp1):
    a = bar(inp1)
    b = bar(inp1)
    return combine([a, b])

ag_asset = AssetsDefinition.from_graph(ag)

@asset
def inp1():
    return 1


@repository
def repo():
    return [ag_asset, inp1, define_asset_job("all")]
```

When viewing the op graph, we'd expect both `bar` and `bar_2` to be connected to the graph input `inp1`. However, we only see a single connection here:
![Screen Shot 2022-08-04 at 2 57 04 PM](https://user-images.githubusercontent.com/22457492/182959858-4c914693-7eb7-4ec8-8811-d581ce74e7d7.png)

The view when you click "Explode graphs" is even more inaccurate:
![Screen Shot 2022-08-04 at 2 58 32 PM](https://user-images.githubusercontent.com/22457492/182959994-d1bc2f57-b636-496e-bbdc-e1be5da1dda7.png)

However, the underlying ops appear to be connected as expected, and if you execute this graph you'll see an accurate step dependency graph:

![Screen Shot 2022-08-04 at 2 53 58 PM](https://user-images.githubusercontent.com/22457492/182960062-30cf0d46-3530-4396-8046-82bb44cf912d.png)

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

bengotow:
I was able to repro this using your example, thanks for filing it! This is... quite busted, and it does seem like its running into limitations of our current rendering logic.

PR #9527, opened 2022-08-29
Fix issue with one-to-many input/output mappings in GraphQL responses

bengotow:
### Summary & Motivation
This is a fix for the issue described in https://github.com/dagster-io/dagster/issues/9224.

Given the example in that issue (a two op graph, where the second op is a composite graph and both inner ops depend on the first), the following GraphQL query returned the wrong data. Rather than two input mappings, we only see one. It appears this is due to the fact that the input_mappings resolver is iterating over input_defs and then retrieving the input mapping for each def. There can be more than one, but it just returns the first hit.

We have input_mapping_defs now (maybe we didn't initially?) and iterating over those fixes this. 

I also removed the "input def => mapping" lookup functions because the code was not used elsewhere and doesn't capture the "more than one mapping" case correctly.

Before:
image

After:

image


clairelin135 (approved):
LGTM, thanks Ben!

PR #9529, opened 2022-08-29
[dagit] Run timeline fixes

hellendag:
### Summary & Motivation

A couple of run timeline fixes. I noticed these things were broken (or not ideal) while looking at some heavier timelines.

- Very large batches need the hover content broken up, otherwise we try to render a ton of rows. Slice at 50 and link to the Runs tab for the job.
- Do not allow runs to batch if the batch would overlap the "now" line. The goal here is to keep scheduled ticks separate from past or ongoing runs, at all times.

Jest. Loaded timeline affected by these issues, verified that the hover content is sane and that scheduled ticks are no longer batched with previous/ongoing runs.

### DIFF
--- a/js_modules/dagit/packages/core/src/runs/RunTimeline.tsx
+++ b/js_modules/dagit/packages/core/src/runs/RunTimeline.tsx
@@ -414,7 +414,7 @@ const RunTimelineRow = ({
               }}
             >
               <Popover
-                content={<RunHoverContent jobKey={job.key} batch={batch} />}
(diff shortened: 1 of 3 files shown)


hellendag:
* master
  * **PR #9529** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9529" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9529?utm_source=stack-comment).

prha (approved):

Issue #7580, opened 2022-04-26
partitioned asset sensors


sryza:
When an asset sensor observes a materialization that corresponds to a particular partition, it should kick off runs that are downstream of that partition.

PR #9533, opened 2022-08-29
Partitioned Asset Sensor

clairelin135:
Adds a `@partitioned_asset_sensor` with a `PartitionedAssetSensorContext` with methods to:
- fetch the latest materialization per asset, and its partition
- fetch the latest materialization per partition per asset
- infer the corresponding partition in a downstream asset given the upstream partition and upstream asset
- get current cursor partition

Currently only allows inferring partition mapping between time window partitioned assets. Functions similarly to @jamiedemaria's `@multi_asset_sensor`

Next steps include:
- Allowing partition mapping between non-time window partitioned assets (e.g. unpartitioned -> time window, static -> static, etc.) 
- Building automatic sensor that detects upstream materializations and kicks off corresponding runs

Issue: https://github.com/dagster-io/dagster/issues/7580

### DIFF
--- a/python_modules/dagster/dagster/__init__.py
+++ b/python_modules/dagster/dagster/__init__.py
@@ -305,6 +305,7 @@ from dagster._core.definitions.sensor_definition import (
     SensorEvaluationContext as SensorEvaluationContext,
     build_sensor_context as build_sensor_context,
     build_multi_asset_sensor_context as build_multi_asset_sensor_context,
+    build_partitioned_asset_sensor_context as build_partitioned_asset_sensor_context,
 )
 from dagster._core.definitions.source_asset import (
     SourceAsset as SourceAsset,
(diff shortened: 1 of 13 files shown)


sryza (commented):
Would it make sense to avoid introducing a new class and instead add methods / attributes to `MultiAssetSensorContext`?

clairelin135:
@sryza Yep, agree that we should do this. Originally I created a new class in order to accept the `AssetsDefinitions`, but I think I can use @jamiedemaria's `UnresolvedAssetSensorDefinition` to modify `MultiAssetSensorContext` to accept the resolved assets

clairelin135:
Closing in favor of merging with `MultiAssetSensor` on top of https://github.com/dagster-io/dagster/pull/9566/files

PR #9535, opened 2022-08-29
Consolidate resource / instance logic for invocation

dpeng817:
### Summary & Motivation
Consolidates the logic used to open resource-based context managers for direct invocation of dagster definitions. Also fixes https://github.com/dagster-io/dagster/issues/9362

Existing unit tests + ensuring that the repro in linked issue does not result in an error.

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/context/hook.py
+++ b/python_modules/dagster/dagster/_core/execution/context/hook.py
@@ -9,10 +9,11 @@ from ...definitions.decorators.graph_decorator import graph
 from ...definitions.dependency import Node
 from ...definitions.hook_definition import HookDefinition
 from ...definitions.op_definition import OpDefinition
-from ...definitions.resource_definition import IContainsGenerator, Resources
-from ...errors import DagsterInvalidPropertyError, DagsterInvariantViolationError
+from ...definitions.resource_definition import Resources
+from ...errors import DagsterInvalidPropertyError
 from ...log_manager import DagsterLogManager
 from ..plan.step import ExecutionStep
+from .resource_invocation import UsesResourcesContext
 from .system import StepExecutionContext
 
 
@@ -184,7 +185,7 @@ class HookContext:
         return self.solid_output_values
 
 
-class UnboundHookContext(HookContext):
+class UnboundHookContext(HookContext, UsesResourcesContext):
     def __init__(
         self,
         resources: Mapping[str, Any],
@@ -193,7 +194,6 @@ class UnboundHookContext(HookContext):
         job_name: Optional[str],
         op_exception: Optional[Exception],
     ):  # pylint: disable=super-init-not-called
-        from ..build_resources import build_resources, wrap_resources_for_execution
         from ..context_creation_pipeline import initialize_console_manager
 
         self._op = None
@@ -205,11 +205,11 @@ class UnboundHookContext(HookContext):
 
             self._op = temp_graph.solids[0]
 
-        # Open resource context manager
-        self._resource_defs = wrap_resources_for_execution(resources)
-        self._resources_cm = build_resources(self._resource_defs)
-        self._resources = self._resources_cm.__enter__()  # pylint: disable=no-member
-        self._resources_contain_cm = isinstance(self._resources, IContainsGenerator)
+        UsesResourcesContext.__init__(
+            self,
+            resources=resources,
+            context_str="hook",
+        )
 
         self._run_id = run_id
         self._job_name = job_name
@@ -218,17 +218,6 @@ class UnboundHookContext(HookContext):
         self._log = initialize_console_manager(None)
 
         self._cm_scope_entered = False
-
-    def __enter__(self):
-        self._cm_scope_entered = True
-        return self
-
-    def __exit__(self, *exc):
-        self._resources_cm.__exit__(*exc)  # pylint: disable=no-member
-
-    def __del__(self):
-        if self._resources_contain_cm and not self._cm_scope_entered:
-            self._resources_cm.__exit__(None, None, None)  # pylint: disable=no-member
 
     @property
     def job_name(self) -> str:
@@ -263,14 +252,8 @@ class UnboundHookContext(HookContext):
         raise DagsterInvalidPropertyError(_property_msg("hook_def", "property"))
 
     @property
-    def resources(self) -> "Resources":
-        if self._resources_contain_cm and not self._cm_scope_entered:
-            raise DagsterInvariantViolationError(
-                "At least one provided resource is a generator, but attempting to access "
-                "resources outside of context manager scope. You can use the following syntax to "
-                "open a context manager: `with build_hook_context(...) as context:`"
-            )
-        return self._resources
+    def resources(self) -> Resources:
+        return UsesResourcesContext.get_resources(self)
 
     @property
     def solid_config(self) -> Any:
--- a/python_modules/dagster/dagster/_core/execution/context/init.py
+++ b/python_modules/dagster/dagster/_core/execution/context/init.py
@@ -1,16 +1,14 @@
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Mapping, Optional, Union
 
 import dagster._check as check
 from dagster._annotations import public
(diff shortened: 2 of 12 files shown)


dpeng817:
* master
  * **PR #9535** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9535" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9535?utm_source=stack-comment).

python_modules/dagster/dagster/_core/execution/context/hook.py
```diff
@@ -205,11 +205,11 @@ def temp_graph():
...
-        # Open resource context manager
-        self._resource_defs = wrap_resources_for_execution(resources)
-        self._resources_cm = build_resources(self._resource_defs)
-        self._resources = self._resources_cm.__enter__()  # pylint: disable=no-member
-        self._resources_contain_cm = isinstance(self._resources, IContainsGenerator)
+        UsesResourcesContext.__init__(
+            self,
+            resources=resources,
+            context_str="hook",
+        )
```
alangenfeld, line 212: what is going on here? why cant we just use `super()` ?

python_modules/dagster/dagster/_core/execution/context/hook.py
```diff
@@ -263,14 +252,8 @@ def required_resource_keys(self) -> Set[str]:
...
-    def resources(self) -> "Resources":
-        if self._resources_contain_cm and not self._cm_scope_entered:
-            raise DagsterInvariantViolationError(
-                "At least one provided resource is a generator, but attempting to access "
-                "resources outside of context manager scope. You can use the following syntax to "
-                "open a context manager: `with build_hook_context(...) as context:`"
-            )
-        return self._resources
+    def resources(self) -> Resources:
+        return UsesResourcesContext.get_resources(self)
```
alangenfeld, line 256: [1]

python_modules/dagster/dagster/_core/execution/context/init.py
```diff
@@ -129,56 +130,23 @@ def __init__(
...
-        if self._resources_cm and self._resources_contain_cm and not self._cm_scope_entered:
-            self._resources_cm.__exit__(None, None, None)  # pylint: disable=no-member
-        if self._instance_provided and not self._cm_scope_entered:
-            self._instance_cm.__exit__(None, None, None)  # pylint: disable=no-member
+        UsesResourcesContext.__init__(
+            self,
+            resources=resources,
+            instance=instance,
+            context_str="init_resource",
+        )
```
alangenfeld, line 149: <i class='graphite__hidden'>[Re: lines 135 to 149]<br/></i>

[1]<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9535?utm_source=unchanged-line-comment">Graphite</a>.</span>

python_modules/dagster/dagster/_core/execution/context/init.py
```diff
@@ -190,18 +158,18 @@ def resource_def(self) -> Optional[ResourceDefinition]:
...
-    def resources(self) -> Resources:
-        if self._resources_cm and self._resources_contain_cm and not self._cm_scope_entered:
-            raise DagsterInvariantViolationError(
-                "At least one provided resource is a generator, but attempting to access "
-                "resources outside of context manager scope. You can use the following syntax to "
-                "open a context manager: `with build_init_resource_context(...) as context:`"
-            )
-        return self._resources
+    def resources(self) -> Any:
+        return UsesResourcesContext.get_resources(self)
```
alangenfeld, line 164: [1]

python_modules/dagster/dagster/_core/execution/context/input.py
```diff
@@ -91,33 +102,15 @@ def __init__(
...
-            self._resources_cm.__exit__(*exc)  # pylint: disable=no-member
-
-    def __del__(self):
-        if self._resources_cm and self._resources_contain_cm and not self._cm_scope_entered:
-            self._resources_cm.__exit__(None, None, None)  # pylint: disable=no-member
+        UsesResourcesContext.__init__(
+            self,
+            resources=resources,
+            context_str="input",
+        )
```
alangenfeld, line 113: [1]

python_modules/dagster/dagster/_core/execution/context/output.py
```diff
@@ -149,38 +136,15 @@ def __init__(
...
-            and self._resources_cm
-            and self._resources_contain_cm
-            and not self._cm_scope_entered
-        ):
-            self._resources_cm.__exit__(None, None, None)  # pylint: disable=no-member
+        UsesResourcesContext.__init__(
+            self,
+            resources=resources,
+            context_str="output",
+        )
```
alangenfeld, line 147: [1]

python_modules/dagster/dagster/_core/execution/context/resource_invocation.py
```diff
@@ -0,0 +1,112 @@
...
+        self._resources = self._resources_cm.__enter__()
+        return self
+
+    def __exit__(self, *exc):
+        if self._instance_cm:
+            self._instance = None
+            self._instance_cm.__exit__(*exc)
+        if self._resources_cm:
+            self._resources = None
+            self._resources_cm.__exit__(*exc)
```
alangenfeld, line 104: when do we still need exit clean up?

alangenfeld (commented):
confused by this weird sub/super class setup [1]

python_modules/dagster/dagster/_core/execution/context/hook.py
dpeng817: We can't use super because the two constructors have different arguments. It's a case of diamond inheritance: https://stackoverflow.com/questions/46199975/resolving-diamond-inheritance-within-python-classes

python_modules/dagster/dagster/_core/execution/context/resource_invocation.py
dpeng817: Some resources will require cleanup (eg a finally statement). For resources that have this, we need to make sure that cleanup doesn't occur until after the context manager ie `with build_x_context` is exited. And we only want to call exit if the cm actually exists.

Hmm I guess I'm thinking now that we also need to check whether instance_cm / resources_cm has already been exited, so maybe we set those to none after exiting as well

python_modules/dagster/dagster/_core/execution/context/hook.py
alangenfeld: i dont think its a diamond since `UsesResourcesContext` doesn't have a parent class - look in to the module resolution order and maybe switch up the order of inheritance - the `mro()` method can be useful for debugging. If that proves fruitless these deserve a comment at least 

https://www.python.org/download/releases/2.3/mro/

alangenfeld (changes requested):
to you for another spin

sryza (changes requested):

dpeng817 (commented):
@alangenfeld you are correct, it's not diamond inheritance. It's just a case where the two superclasses each have different arguments to the __init__ function, making it impossible for super to properly resolve. It's not about ordering, it's just that the arguments don't match at all. I guess one thing we could do is move to a model where each of the superclasses uses `args` and `kwargs`, and thus match, but handle arguments differently but I honestly think that's more confusing than just calling init methods directly. Thoughts?

dpeng817 (commented):
Re-requesting review since tests are fixed, and comment is addressed

dpeng817 (commented):
Found a different solution here which I like: https://stackoverflow.com/questions/18006310/calling-superclass-constructors-in-python-with-different-arguments - instead of making `UsesResourcesContext` a superclass, we can have the invocation contexts contain an instance of a class that metes out resource access. Also, the pattern I use here for `.__init__` calling directly can apparently lead to problems where not all superclasses are initialized: https://fuhm.net/super-harmful/

dpeng817 (commented):
@schrockn no review needed - it's in draft mode so I thought it would remove itself from your review queue, but that might only be a graphite thing

PR #9537, opened 2022-08-30
[dagit] Support composite input/output mappings with a name change

bengotow:
### Summary & Motivation

This is the second (Dagit) fix for https://github.com/dagster-io/dagster/issues/9224. The other portion of the fix is a backend change in https://github.com/dagster-io/dagster/pull/9527.

To test this code, I expanded on the example in #9224 to create a worst case scenario where composite input mappings map external ops like `inp1`, `inp2` to inputs with a different name on the interior ops (just `inp`). It turns out we didn't support this in Dagit because of a minor error in this algorithm.

I also created a scenario where a composite has two inputs, and only some of the interior ops use each input. This was also not reflected correctly, because our algorithm applied each composite input mapping to each interior op, not just the mapped op.

Test case:

```
from dagster import AssetsDefinition, op, graph, asset, repository, define_asset_job


@op
def bar(inp):
    return inp


@op
def combine(x):
    return x


@op
def yoyoyo(zztop):
    return 5


@graph
def ag(inp1, inp2):
    a = bar(inp1)
    b = bar(inp1)
    c = yoyoyo(inp2)
    return combine([a, b, c]), b


ag_asset = AssetsDefinition.from_graph(ag)


@asset
def inp1():
    return 1


@asset
def inp2():
    return 2


@asset
def finalizer(ag):
    return 1


@asset
def hobbyproject(ag):
    return 2


@repository
def repo():
    return [ag_asset, finalizer, inp1, inp2, hobbyproject, define_asset_job("all")]
```

Graph:
Screen Shot 2022-08-30 at 8 21 09 AM
Screen Shot 2022-08-30 at 8 21 12 AM

### DIFF
--- a/js_modules/dagit/packages/core/src/pipelines/CompositeSupport.ts
+++ b/js_modules/dagit/packages/core/src/pipelines/CompositeSupport.ts
@@ -2,49 +2,11 @@ import {GraphExplorerSolidHandleFragment} from './types/GraphExplorerSolidHandle
 
 function explodeComposite(
   handles: GraphExplorerSolidHandleFragment[],
-  handle: GraphExplorerSolidHandleFragment,
+  composite: GraphExplorerSolidHandleFragment,
 ) {
-  if (handle.solid.definition.__typename !== 'CompositeSolidDefinition') {
+  if (composite.solid.definition.__typename !== 'CompositeSolidDefinition') {
     throw new Error('explodeComposite takes a composite handle.');
   }
-
-  // Replace all references to this composite's inputs in other solid defitions
-  // with the interior target of the input mappings
-  handle.solid.definition.inputMappings.forEach((inmap) => {
-    const solidName = `${handle.solid.name}.${inmap.mappedInput.solid.name}`;
-    handles.forEach((h) =>
-      h.solid.outputs.forEach((i) => {
-        i.dependedBy.forEach((dep) => {
-          if (
-            dep.definition.name === inmap.definition.name &&
-            dep.solid.name === handle.solid.name
-          ) {
-            dep.solid.name = solidName;
-            dep.definition.name = inmap.mappedInput.definition.name;
-          }
-        });
-      }),
-    );
-  });
-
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/core/src/pipelines/CompositeSupport.ts
```diff
@@ -2,107 +2,130 @@ import {GraphExplorerSolidHandleFragment} from './types/GraphExplorerSolidHandle
...
+  composite: GraphExplorerSolidHandleFragment,
 ) {
-  if (handle.solid.definition.__typename !== 'CompositeSolidDefinition') {
+  if (composite.solid.definition.__typename !== 'CompositeSolidDefinition') {
     throw new Error('explodeComposite takes a composite handle.');
   }
 
-  // Replace all references to this composite's inputs in other solid defitions
-  // with the interior target of the input mappings
-  handle.solid.definition.inputMappings.forEach((inmap) => {
```
bengotow, line 13: I left this code unchanged but moved it down in the function (inside the other inputMapping iterator) so it could re-use the same local variables

js_modules/dagit/packages/core/src/pipelines/CompositeSupport.ts
```diff
@@ -2,107 +2,130 @@ import {GraphExplorerSolidHandleFragment} from './types/GraphExplorerSolidHandle
...
-  // target solids
-  handle.solid.definition.inputMappings.forEach((inmap) => {
-    handle.solid.inputs.forEach((input) => {
-      const solidName = `${handle.solid.name}.${inmap.mappedInput.solid.name}`;
-      const [interiorTarget] = handles.filter((h) => h.handleID === solidName);
-      if (!interiorTarget) {
-        return;
-      }
-      const matchingInput = interiorTarget.solid.inputs.find(
-        (i) => i.definition.name === input.definition.name,
```
bengotow, line 80: Main fix: This should have been `== inmap.mappedInput.definition.name`

hellendag (approved):

PR #9538, opened 2022-08-30
Allow scheduling of union of cron strings

erinov1:
### Summary & Motivation
Allow users to specify a schedule as the union of cron strings, in which case the next execution datetime is obtained by computing the next cron datetime after the current execution datetime for each cron string in the list, and then choosing the earliest among them.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/schedule_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/schedule_decorator.py
@@ -2,7 +2,7 @@ import copy
 import datetime
 import warnings
 from functools import update_wrapper
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, cast
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, cast, Union, Sequence
 
 import dagster._check as check
 from dagster._core.definitions.partition import (
@@ -47,7 +47,7 @@ if TYPE_CHECKING:
 
 
 def schedule(
-    cron_schedule: str,
+    cron_schedule: Union[str, Sequence[str]],
     *,
     job_name: Optional[str] = None,
     name: Optional[str] = None,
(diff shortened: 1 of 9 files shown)


erinov1:
@sryza These are just changes to the execution time iterator (and a unit test). If this looks reasonable then I will work on threading this into the ScheduleDefinition and all the other tests. A few questions/comments:

1. There doesn't seem to be a `check` for `Union[type, List[type]]`. Is the best way to do this just
```python
check.inst_param(cron_schedule, "cron_schedule", (str, List))
if not instance(cron_schedule, str):
    check.list_param(cron_schedule, "cron_schedule", of_type=str)
```
2. I'm not familiar with the dagit side of things. There appears to be code for displaying the schedule that I don't know how to address.

sryza (commented):
@erinov1 this looks right to me.

On typing:
- We don't need checks inside the `schedule_execution_time_iterator` and `cron_string_iterator`, because they're not public-facing. If they had them in the past, it's because they were written before we started mypy-typing our codebase.
- For the public-facing arguments of `ScheduleDefinition` etc., we will need checks, and I'm not aware of a better way of doing it then the pattern you mentioned.
- We typically prefer `Sequence` to `List`, because it indicates that the function won't mutate the value, and it's [covariant](https://blog.magrathealabs.com/pythons-covariance-and-contravariance-b422c63f57ac).

erinov1:
I think these are most of the pure python + dagster changes needed. Let me know if there are other tests that should be updated — my impression is that the remaining schedule tests do not need to test for unions of cron strings.

Am I also correct in thinking that the `cron_schedule` argument in `PartitionScheduleDefinition` does not need to be updated, since it is only used for the monthly/weekly/daily/hourly schedules?

sryza:
@erinov1 this looks great. buildkite is complaining about import ordering and code formatting in a few places.  Would you be able to run `make isort` and `make black` from the repo root?

Also, it's outputting a few mypy and lint issues. 

mypy output:
```
[2022-08-31T18:43:49Z] dagster/_core/definitions/schedule_definition.py:266: error: Argument 1 to "sequence_param" has incompatible type "Optional[Sequence[str]]"; expected "Sequence[str]"  [arg-type]
--
  | [2022-08-31T18:43:49Z] dagster/_core/definitions/schedule_definition.py:466: error: Incompatible return value type (got "Sequence[str]", expected "str")  [return-value]
```

lint output:
```
[2022-08-31T18:43:57Z] ************* Module dagster._utils.schedules
--
  | [2022-08-31T18:43:57Z] dagster/_utils/schedules.py:2:0: W0611: Unused List imported from typing (unused-import)
```

erinov1:
@sryza I reran black + isort. I was unable to fix the mypy issue in `schedule_definition.py`, since the typehint for `cron_schedule` is `Optional[Union[str, Sequence[str]]`, and the `check` functions respect this typehint. I just added `type: ignore` comments for now, but can implement a different fix if you have an idea (maybe `assert`ing` after the checks will remove the `Optional` type).

I also added a test to check for an empty union that I forgot before.

sryza (approved):
Nice! I'm going to merge this.

erinov1:
Thanks! (I do still think that supporting more general user-defined execution time iterators could be useful.)

PR #9541, opened 2022-08-30
[dagit] Fix zero-index reference on potentially empty array

hellendag:
### Summary & Motivation

Based on an error that appeared in Cloud.

We aren't checking whether `assetMaterializations` actually has a zero element, just that `asset` is not nullish. The result is that checking `timestamp` on the zero-index element can lead to a runtime error.

Ensure that the array has a length before looking at its first element.

Buildkite. Sanity check that an asset view loads.


hellendag:
* master
  * **PR #9541** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9541" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9541?utm_source=stack-comment).

prha (approved):

PR #9546, opened 2022-08-30
Asset sensor utility

### LINKED ISSUE #9000: utility to build an asset sensor from a software-defined asset
* By default, it would trigger whenever the upstream assets are updated. Will need to figure out how this works when there are multiple upstream assets - AND or OR?
* If the assets are partitioned, it would trigger the corresponding partition in the downstream assets when the corresponding partitions in the upstream assets are updated.

Relevant requests:
- https://dagster.slack.com/archives/C01U5LFUZJS/p1659067359346519

jamiedemaria:
### Summary & Motivation
rough first pass implementation of a sensor utility that will monitor the parents of all the provided assets and kick off a materialization of an asset when all of its parents have materialized. 

Fixes #9000 

TODO:
- [ ] ~~docs~~ - punting to a follow up pr after this one goes in
- [x] clean up code
- [ ] come up with a way for users to unit test?
- [ ] ~~logging in sensor fn to document which assets are being skipped and why~~ - punting as there's no way to log in sensors right now

new daemon tests


python_modules/dagster/dagster/_core/definitions/unresolved_asset_sensor_definition.py
```diff
@@ -0,0 +1,266 @@
...
+        """Computes a mapping of assets in self._selection to their parents in the asset graph"""
+        upstream = defaultdict(set)
+        selection_resolved = list(self._selection.resolve([*assets, *source_assets]))
+        for a in selection_resolved:
+            a_parents = list(AssetSelection.keys(a).upstream().resolve([*assets, *source_assets]))
+            # filter out a because upstream() includes the assets in the original AssetSelection
+            upstream[a] = {p for p in a_parents if p != a}
+        return upstream
+
+    def _get_parent_materializations(self, context, asset_key, parent_assets, cursor):
```
jamiedemaria, line 74: I'm going to rename this to be more generic `_get_parent_updates` or something along those lines so that if/when we incorporate source asset versioning we can support source assets

python_modules/dagster/dagster/_core/definitions/unresolved_asset_sensor_definition.py
```diff
@@ -0,0 +1,266 @@
...
+#         selection_resolved = list(self._selection.resolve([*assets, *source_assets]))
+#         for a in selection_resolved:
+#             a_parents = list(AssetSelection.keys(a).upstream().resolve([*assets, *source_assets]))
+#             # filter out a because upstream() includes the assets in the original AssetSelection
+#             upstream[a] = {p for p in a_parents if p != a}
+#             for p in upstream[a]:
+#                 downstream[p].add(a)
+#         return {"upstream": upstream, "downstream": downstream}
+
+#     def make_sensor(self, dependency_mapping: Mapping[AssetKey, Sequence[AssetKey]]):
```
jamiedemaria, line 177: alternate implementation that uses a combo of cursor and materialization time to determine if an asset should be materialized

jamiedemaria (commented):
@sryza @prha @clairelin135 looking for a quick gut check review for any glaring issues with the way this is implemented or the API for specifying the sensor. I'm going to dive into clean up and unit tests next, and after that I'll make this a real PR and we can do more thorough review. Right now I have a super simple use case working in dagit that you can see in the toys dir 

Based on some of the conversations we've had with Owen and Sean I want to add the functionality to have the sensor materialize an asset if any of its parents have updated. We could do this in a couple of ways:
1. User makes two sensors, one that monitors assets that should only materialize when all parents are materialized (AND case), and one that monitors assets that should materialize when any parent updates (OR case)
2. Users make a single sensor, but pass in two AssetSelectors, one for the AND case and one for the OR case

any opinions on which direction? I think 1 makes more sense to me, but i don't feel super strong that that's the for sure best way to go.

sryza:
This AND/OR problem is tricky. In theory, you could imagine a situation where you want to update asset X when you've received updates to ((A and B) or C).

Thinking through it, my instinct is to try to back up and think about _why_ a user might choose AND vs. OR.  Some considerations:

1. If you've got a diamond-shaped asset graph like this:

```
A -> B -> D
 \-> C ->/ 
```
You likely want to update D only when both B AND C have been updated.  Because updates to B and C are downstream of updates to A, B and C will be in a kind of inconsistent state if only one of them incorporates an update to A.

2. It depends on when the other asset is likely to be updated.  E.g. imagine asset C depends on asset A and B.  If A is updated yearly, and B is updated hourly, you probably want to update C as soon as B is updated.  But if A and B are both updated hourly but land at slightly different times within the hour, you likely want to update C just once per hour.

Alas, neither of these give a clear algorithm for what to do in all cases, so I'm a tiny bit stumped.  Tossing out some heuristics to consider:
- If there are active runs that are on track to materialize an asset, wait for that asset to be materialized.
- Wait some amount of time for an asset to be materialized before launching a run without it.

sryza:
Another option could just be to start with AND and then see what users ask for

danielgafni:
I suggest to use AND by default and allow to pass a user-defined decision function. Something like this:

```python
def my_sensor_condition(upstream_statuses) -> bool:
    # user logic

build_asset_sensor(asset, sensor_condition=my_sensor_condition)

```

python_modules/dagster/dagster/_core/definitions/unresolved_asset_sensor_definition.py
```diff
@@ -0,0 +1,266 @@
...
+            should_materialize = []
+            cursor_update_dict = {}
+            for a in upstream.keys():
+                a_cursor = cursor_dict.get(str(a)) if cursor_dict.get(str(a)) else {}
+                (
+                    parent_asset_event_records,
+                    a_cursor_update_dict,
+                ) = self._get_parent_materializations(context, a, upstream[a], a_cursor)
+
+                if all(
```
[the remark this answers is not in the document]
jamiedemaria: it currently doesn't do that, but it's been floating around my head as a good thing to add in

jamiedemaria:
@clairelin135 @prha @sryza I took another pass at this, here's what has and hasn't changed:

Changes:
- assets downstream of assets that will be materialized by the sensor will now be materialized. For example if we have the graph 
```
A -> B -> D
 \-> C ->/ 
```
then when A is materialized, the sensor will know that B and C should materialize and therefore D can also materialize
- assets with a parent currently being materialized will NOT be materialized. For example, if we have the graph `A->B` and during sensor execution, we see that A has been materialized, but there is also a run in progress that will materialize A, we will NOT materialize B (once the new materialization of A is complete, the next sensor tick will materialize B)

Things I have not yet addressed:
* the AND vs OR vs custom function question. I am quite hesitant to allow custom functions (at least for the first iteration of this sensor). Allowing the user to specify a custom function also means that there will need to be custom cursor updating logic (see the discussion in PR #9339). Keeping track of the cursor in this sensor is already quite challenging, and putting the burden to grok the cursor on users feels unnecessary, at least at this point. As it stands, i feel pretty confident that AND and OR cases can be supported, so I'm up to allowing users to specify AND or OR (via a bool param or something) in the first version. 
* if creating an UnresolvedAssetSensor class is the right approach. I know in Claire's PR the choice was made to move to having functions on the context that could fetch the repository. I'd be up to change this PR to do that as well, but i need to think through any implications of that (i dont think there are any, but i'm not 100% sure on that)
* the name of the sensor. As Sandy mentioned `build_asset_sensor` doesn't quite capture what this does. I haven't thought too much about better names, but i'll be brainstorming, and i'm open to suggestions
* complexity/readability of the code. right now the code feels quite messy and hard to understand. I want to take some time to try and refactor it to make it easier to read + maintain. I'm interested to know what sections you find challenging to understand so i can incorporate that into any refactoring work

jamiedemaria:
brainstorming for what to name this thing

theme words:
- automatic
- reconciliation
- up to date/updated/updating


name ideas:
- asset_reconciliation_sensor
-

sryza (commented):
I like `build_asset_reconciliation_sensor`!

python_modules/dagster/dagster/_core/definitions/unresolved_asset_sensor_definition.py
```diff
@@ -0,0 +1,279 @@
...
+    def _get_upstream_mapping(
+        self,
+        assets: Sequence["AssetsDefinition"],
+        source_assets: Sequence["SourceAsset"],
+    ) -> Mapping[AssetKey, Set[AssetKey]]:
+        """Computes a mapping of assets in self._selection to their parents in the asset graph"""
+        upstream = defaultdict(set)
+        selection_resolved = list(self._selection.resolve([*assets, *source_assets]))
+        for a in selection_resolved:
+            a_parents = list(AssetSelection.keys(a).upstream().resolve([*assets, *source_assets]))
```
clairelin135, line 53: It's possible if we had assets A -> B -> C and we only selected asset C that this would fetch all upstreams (including both A and B) right?

Should we only be selecting `upstream(depth=1)`? As in only the immediate upstream parents of the selected assets

python_modules/dagster/dagster/_core/definitions/unresolved_asset_sensor_definition.py
```diff
@@ -0,0 +1,279 @@
...
+                print(f"will materialize record {materialization_planned_event_records}")
+
+                if event_records and materialization_planned_event_records:
+                    # planned materialization is from a different run than the most recent
+                    # completed materialization
+                    if (
+                        event_records[0].event_log_entry.run_id
+                        != materialization_planned_event_records[0].event_log_entry.run_id
+                    ):
+                        # we don't want to materialize the child asset because one of its parents is
```
clairelin135, line 147: I think it would be helpful to signify to the user somehow when an asset materialization is in progress, which is why a certain asset was skipped in execution.

Throwing out some ideas:
- Log some sort of message
- Maybe it's possible yield a skip reason when certain assets in the sensor selection are skipped?

python_modules/dagster/dagster/_core/definitions/unresolved_asset_sensor_definition.py
```diff
@@ -0,0 +1,279 @@
...
+                    limit=1,
+                )
+
+                print(f"will materialize record {materialization_planned_event_records}")
+
+                if event_records and materialization_planned_event_records:
+                    # planned materialization is from a different run than the most recent
+                    # completed materialization
+                    if (
+                        event_records[0].event_log_entry.run_id
```
clairelin135, line 144: I think it's possible that a run will yield an `ASSET_MATERIALIZATION_PLANNED` event and then fail before the asset is actually materialized. I think we should instead test if the run of the `ASSET_MATERIALIZATION_PLANNED` event is in progress instead of testing equivalence by run id:
```
instance.get_runs(filters=RunsFilter(run_ids=[...], statuses=IN_PROGRESS_RUN_STATUSES))
```

python_modules/dagster/dagster_tests/daemon_sensor_tests/test_asset_sensor_run.py
```diff
@@ -0,0 +1,660 @@
...
+                [r.asset_selection == {AssetKey("f"), AssetKey("d")} for r in run_request_runs]
+            )
+
+            freeze_datetime = freeze_datetime.add(seconds=60)
+
+
+@pytest.mark.parametrize("executor", get_sensor_executors())
+def test_layered_sensor(executor):
+    """Asset graph:
+        x       z       e
```
clairelin135, line 271: probably worth adding another test like this one, that only selects g, to confirm that materializations for x/z/e shouldn't affect g

python_modules/dagster/dagster_tests/daemon_sensor_tests/test_asset_sensor_run.py
```diff
@@ -0,0 +1,660 @@
...
+        instance,
+        workspace,
+        external_repo,
+    ):
+        with pendulum.test(freeze_datetime):
+            for _ in range(5):
+                materialize([x], instance=instance)
+                wait_for_all_runs_to_finish(instance)
+
+            # something is wrong with this test. The in progress asset materialization events are coming in
```
clairelin135, line 357: very odd... the asset materialization planned events should be created upon run creation

python_modules/dagster/dagster_tests/daemon_sensor_tests/test_asset_sensor_run.py
```diff
@@ -0,0 +1,660 @@
...
+
+            freeze_datetime = freeze_datetime.add(seconds=60)
+
+
+@pytest.mark.parametrize("executor", get_sensor_executors())
+def test_layered_sensor(executor):
+    """Asset graph:
+        x       z       e
+        \       /\      /
+            d       f
```
clairelin135, line 273: Can we also add another test where a materialization is triggered by a user for an asset that is monitored by the asset sensor? For example, with an asset selection of d, f, g where d is materialized so we can confirm that g gets materialized but d does not

python_modules/dagster/dagster/_core/definitions/unresolved_asset_sensor_definition.py
```diff
@@ -0,0 +1,279 @@
...
+        source_assets: Sequence["SourceAsset"],
+    ) -> SensorDefinition:
+        self._source_asset_keys = [asset.key for asset in source_assets]
+        return self._make_sensor(self._get_upstream_mapping(assets, source_assets))
+
+
+@experimental
+def build_asset_sensor(
+    selection: AssetSelection,
+    name: str,
```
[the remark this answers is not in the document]
jamiedemaria: maybe all of the assets?

python_modules/dagster/dagster/_core/definitions/unresolved_asset_sensor_definition.py
jamiedemaria: good catch! i misread the docstring on `upstream`

python_modules/dagster/dagster/_core/definitions/unresolved_asset_sensor_definition.py
jamiedemaria: yeah logging would be good. i don't think we can yield a SkipReason because there is a built in check in `SensorDefinition` that only allows one SkipReason to be returned (and doesn't allow returning both a run request and a skip reason).

python_modules/dagster/dagster/_core/definitions/unresolved_asset_sensor_definition.py
jamiedemaria: oh cool that's way better, didn't think/know to do that

python_modules/dagster/dagster/_core/definitions/unresolved_asset_sensor_definition.py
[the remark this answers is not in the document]
sryza: that could get big in some cases - maybe if it gets really big, truncate it and include a hash of the remaining assets?

python_modules/dagster/dagster/_core/definitions/unresolved_asset_sensor_definition.py
```diff
@@ -0,0 +1,279 @@
...
+                    cursor_update_dict[str(a)] = self._get_cursor(cursor_dict, a)
+
+            if len(should_materialize) > 0:
+                context.update_cursor(json.dumps(cursor_update_dict))
+                return RunRequest(run_key=f"{context.cursor}", asset_selection=should_materialize)
+
+        return SensorDefinition(
+            evaluation_fn=sensor_fn,
+            name=self.name,
+            job_name="__ASSET_JOB",
```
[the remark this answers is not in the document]
sryza: take a look at the logic here: https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_core/definitions/asset_group.py#L433

I can add more detail if the logic there isn't clear

jamiedemaria:
jotting down some notes on what would/might happen if we remove `UnresolvedAssetSensor` and directly construct a `MultiAssetSensor`:

* the user passes a selection of assets, but the sensor actually monitors the parents. We can't get the parents during sensor construction time since we don't have access to the repository, so we need to fetch them at execution time. but then when we go to get the events for a parent asset, the MultiAssetSensor will error since the parent asset isn't in the list of assets it's monitoring. This would be an easyish fix of just removing that check 
* a direct port over to `MultiAssetSensor` would still require writing new cursor updating functions since the cursor of this sensor is different than the one the `MultiAssetSensor` expects. all the functions could be private so that users writing a `@multi_asset_sensor` don't see them, but it could still cause confusion on our end

python_modules/dagster/dagster/_core/definitions/unresolved_asset_sensor_definition.py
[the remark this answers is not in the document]
jamiedemaria: oh interesting, so there's an `_ASSET_JOB` for each kind of partition (daily, yearly, etc)? 

thinking out loud here...
so we'll need a way to pass all of those jobs to the SensorDefinition, which will be complicated because we won't know how many partitioned jobs there are until the repository is loaded (and since we're likely moving away from the UnresolvedAssetSensor this means we won't know what jobs to pass to the sensor definition at construction time)

python_modules/dagster/dagster/_core/definitions/unresolved_asset_sensor_definition.py
jamiedemaria: this fixed the issue in the test suite too. thanks for the suggestion!

python_modules/dagster/dagster/_core/definitions/unresolved_asset_sensor_definition.py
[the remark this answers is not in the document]
jamiedemaria: i originally misread the comment, and thought you were suggesting a default for the `selection` param (hence, my original suggestion that the default be all assets). updated to reflect a default name for the sensor. do you think we should be generating a unique default name so that if someone creates multiple of these sensors they don't have to provide names for any of them?

clairelin135:
In general the implementation is looking good to me. I think the tricky part remaining is figuring out how to represent the cursor when the sensor itself is triggering a materialization.

Random ideas about how to handle the cursor when an asset will be materialized by the sensor, which may not be good:

- In the cursor, store a set of run IDs triggered by the sensor previously (can [update the sensor cursor after run launch](https://github.com/dagster-io/dagster/blob/abba86410fde90620d353706c01513d5faecc00c/python_modules/dagster/dagster/_daemon/sensor.py#L587)). Then, on the next sensor tick, iterate through each run that is completed (no longer in progress) and advance the cursor of each asset that was materialized in that run to its materialization event. With this approach, to ensure that the set of runs kicked off by the sensor doesn't grow infinitely large, we'd need to remove runs from the set after they complete and are evaluated in the next tick.
- Since we don't allow assets whose parents are being rematerialized in an in-progress run to be kicked off by this sensor, we could just update the cursor after a run kicked off by this sensor completes. In this case, we'd have to think of some way to poll for run completion, probably by checking for the run status right before the tick is evaluated and then passing the new cursor object to the sensor context for tick evaluation

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -0,0 +1,332 @@
...
+                        event_records[0].event_log_entry.timestamp,
+                    )
+            else:
+                # the parent has not been materialized and will not be materialized by the sensor
+                parent_asset_event_records[p] = (False, 0.0)
+
+    return parent_asset_event_records
+
+
+def _make_sensor(
```
sryza, line 175: As this isn't user-facing, and the only place it's called is with all its values specified, better not to give it default values. Synchronizing them with the default values in other methods creates room for error.

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -0,0 +1,332 @@
...
+            any of its parents have materialized.
+        parent_in_progress_stops_materialization (bool): If True, the sensor will not materialize an
+            asset there is an in-progress run that will materialize any of the asset's parents. Defaults to
+            False.
+        minimum_interval_seconds (Optional[int]): The minimum amount of time that should elapse between sensor invocations.
+        description (Optional[str]): A description for the sensor.
+        default_status (DefaultSensorStatus): Whether the sensor starts as running or not. The default
+            status can be overridden from Dagit or via the GraphQL API.
+
+    Returns: A MultiAssetSensor that will monitor the parents of the provided assets to determine when
```
sryza, line 320: MultiAssetSensor isn't the name of a class, right?

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -0,0 +1,332 @@
...
+        materialize d. Once a has completed materialization, the next sensor tick will launch a run to materialize d.
+
+    Other considerations:
+    * If an asset has a SourceAsset as a parent, the sensor will always consider the source asset to updated
+        since the previous sensor tick. If you have the asset graph my_source->my_asset where my_source is
+        a SourceAsset. The sensor created from build_asset_reconciliation_sensor(AssetSelection.assets(my_asset), name="my_sensor")
+        will materialize my_asset on every sensor tick.
+
+    Args:
+        selection (AssetSelection): The group of assets you want to monitor
```
sryza, line 307: This should be something like "keep up-to-date" instead of "monitor", right?

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -0,0 +1,332 @@
...
+        name (str): The name to give the sensor.
+        and_condition (bool): If True (the default) the sensor will only materialize an asset when
+            all of its parents have materialized. If False, the sensor will materialize an asset when
+            any of its parents have materialized.
+        parent_in_progress_stops_materialization (bool): If True, the sensor will not materialize an
+            asset there is an in-progress run that will materialize any of the asset's parents. Defaults to
```
sryza, line 313: "if there is"

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -0,0 +1,332 @@
...
+        a SourceAsset. The sensor created from build_asset_reconciliation_sensor(AssetSelection.assets(my_asset), name="my_sensor")
+        will materialize my_asset on every sensor tick.
+
+    Args:
+        selection (AssetSelection): The group of assets you want to monitor
+        name (str): The name to give the sensor.
+        and_condition (bool): If True (the default) the sensor will only materialize an asset when
+            all of its parents have materialized. If False, the sensor will materialize an asset when
+            any of its parents have materialized.
+        parent_in_progress_stops_materialization (bool): If True, the sensor will not materialize an
```
sryza, line 312: Maybe "wait_for_in_progress_runs"?

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -0,0 +1,332 @@
...
+        default_status=default_status,
+    )
+
+
+@experimental
+def build_asset_reconciliation_sensor(
+    selection: AssetSelection,
+    name: str,
+    and_condition: bool = True,  # TODO better name for this parameter
+    parent_in_progress_stops_materialization: bool = False,  # TODO - better name for this parameter
```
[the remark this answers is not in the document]
jamiedemaria: I was thinking the opposite because of the case like when we talked to yuhan and rex where they dont want to rerun the dbt models if the upstreams are in progress. But having the sensor get into a deadlock state would maybe be a worse experience for the user than having an asset materialized an extra time? Both choices for defaults have their downsides so I'm fine switching to True (or not providing a default and forcing the user to choose)

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
jamiedemaria: it's not - my bad. Should be MultiAssetSensorDefinition

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
jamiedemaria: yeah i like that a lot better

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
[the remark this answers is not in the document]
jamiedemaria: yeah sounds good

python_modules/dagster-test/dagster_test/toys/asset_sensors.py
```diff
@@ -132,6 +139,48 @@ def every_fifth_materialization_sensor(context):
         )
 
 
+@asset
```
jamiedemaria, line 142: TODO - fix and run this whole section

jamiedemaria (commented):
Getting closer!

Things I still need to do:
- [ ] add logging to the sensor describing why an asset isn't getting materialized. @sryza or @clairelin135 is there a standard way we're adding logging to sensors right now? last i knew there wasn't, but not sure if anything has changed...
- [x] fixing up the remaining TODOs

sryza:
>  @sryza or @clairelin135 is there a standard way we're adding logging to sensors right now?

Alas, there is not. So probably best to punt on logging for now.

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -0,0 +1,348 @@
...
+    its parents have materialized, but it can be set to materialize an asset when any of its
+    parents have materialized.
+
+    Example:
+        If you have the following asset graph:
+
+            .. code-block:: python
+
+                a       b       c
+                \       /\      /
```
sryza, line 270:
```suggestion
                \      / \      /
```

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -0,0 +1,348 @@
...
+        description=description,
+        default_status=default_status,
+    )
+
+
+@experimental
+def build_asset_reconciliation_sensor(
+    selection: AssetSelection,
+    name: str,
+    and_condition: bool = True,  # TODO better name for this parameter
```
sryza, line 252: With `wait_for_in_progress_runs` defaulting to True, I think we can default this to false.

Figuring out the naming for this one is giving me a headache.

Maybe `wait_for_all_upstream`?

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -0,0 +1,348 @@
...
+    wait_for_in_progress_runs: bool = True,
+    minimum_interval_seconds: Optional[int] = None,
+    description: Optional[str] = None,
+    default_status: DefaultSensorStatus = DefaultSensorStatus.STOPPED,
+) -> MultiAssetSensorDefinition:
+    """Constructs a sensor that will monitor the parents of the provided assets and materialize an asset
```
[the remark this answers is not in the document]
sryza: Ok actually I scrolled further down and it does get kinda gnarly.
- Any idea why the code samples are coming out italicized?
- If you take a look at the ConfigSchema class, you can see an example of nice list formatting

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -0,0 +1,370 @@
...
+    current_asset: AssetKey,
+    parent_assets: Set[AssetKey],
+    cursor_timestamp: float,
+    will_materialize_list: Sequence[AssetKey],
+    wait_for_in_progress_runs: bool,
+) -> Mapping[AssetKey, Tuple[bool, float]]:
+    """The bulk of the logic in the sensor is in this function. At the end of the function we return a
+    dictionary that maps each asset to a Tuple. The Tuple contains a boolean, indicating if the asset
+    has materialized or will materialize, and a float representing the timestamp the parent asset
+    would update the cursor too if it is the most recent materialization of a parent asset. In some cases
```
sryza, line 50:
```suggestion
    would update the cursor to if it is the most recent materialization of a parent asset. In some cases
```

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -0,0 +1,370 @@
...
+    wait_for_all_upstream: bool,
+    wait_for_in_progress_runs: bool,
+    minimum_interval_seconds: Optional[int] = None,
+    description: Optional[str] = None,
+    default_status: DefaultSensorStatus = DefaultSensorStatus.STOPPED,
+) -> MultiAssetSensorDefinition:
+    """Creates the sensor that will monitor the parents of all provided assets and determine
+    which assets should be materialized (ie their parents have been updated).
+
+    The cursor for this sensor is a dictionary mapping stringified AssetKeys to a timestamp (float). For each
```
sryza, line 179: Why timestamp instead of storage ID? Timestamps are less reliable for ordering events because of the way computer clocks work. And I think less efficient for us to query on?

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -0,0 +1,370 @@
...
+            parent_update_records = _get_parent_updates(
+                context,
+                current_asset=a,
+                parent_assets=upstream[a],
+                cursor_timestamp=a_cursor,
+                will_materialize_list=should_materialize,
+                wait_for_in_progress_runs=wait_for_in_progress_runs,
+            )
+
+            condition = all if wait_for_all_upstream else any
```
sryza, line 225: woah cool

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -0,0 +1,370 @@
...
+                0.0,
+            )
+        # TODO - when source asset versioning lands, add a check here that will see if the version has
+        # updated if p is a source asset
+        else:
+            if wait_for_in_progress_runs:
+                # if p is currently being materialized, then we don't want to materialize current_asset
+
+                # get the most recent planned materialization
+                materialization_planned_event_records = context.instance.get_event_records(
```
sryza, line 100: Will this result in sometimes querying for the same asset key multiple times. If so, could we cache instead?

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -0,0 +1,370 @@
...
+        upstream[a] = {p for p in a_parents if p != a}
+    return upstream
+
+
+def _get_parent_updates(
+    context,
+    current_asset: AssetKey,
+    parent_assets: Set[AssetKey],
+    cursor_timestamp: float,
+    will_materialize_list: Sequence[AssetKey],
```
sryza, line 44: Documenting the meaning of this argument would be helpful, because it took me a little while to figure it out.

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -0,0 +1,370 @@
...
+    We want to get the materialization information for all of the parents of an asset to determine
+    if we want to materialize the asset in this sensor tick. We also need determine the new cursor
+    value for the asset so that we don't process the same materialization events for the parent
+    assets again.
+
+    We iterate through each parent of the asset and determine its materialization info. The parent
+    asset's materialization status can be one of three options:
+    1. The parent has materialized since the last time the child was materialized (determined by comparing
+        the timestamp of the parent materialization to the cursor_timestamp).
+    2. The parent will be materialized as part of the materialization that will be kicked off by the
```
sryza, line 64: would it be accurate to say "The parent is already slated to be materialized (i.e. included in `will_materialize_list`"?

this tripped me up a little bit on first read, because I first interpreted it (I think erroneously?) as that this function was both finding out info about parents and also making decisions about would be materialized.

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -0,0 +1,370 @@
...
+            context._repository_def.source_assets_by_key  # pylint: disable=protected-access
+        )
+        upstream: Mapping[AssetKey, Set[AssetKey]] = _get_upstream_mapping(
+            selection=selection,
+            assets=asset_defs_by_key.values(),
+            source_assets=source_asset_defs_by_key.values(),
+        )
+
+        cursor_dict: Dict[str, float] = json.loads(context.cursor) if context.cursor else {}
+        should_materialize: List[AssetKey] = []
```
sryza, line 201: Maybe this should be a set because it's elements are supposed to be unique and we're often checking to see whether keys are included in it?

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -0,0 +1,370 @@
...
+        name=name,
+        job_name="__ASSET_JOB",
+        minimum_interval_seconds=minimum_interval_seconds,
+        description=description,
+        default_status=default_status,
+    )
+
+
+@experimental
+def build_asset_reconciliation_sensor(
```
sryza, line 254: This won't work for partitioned jobs, right? If so, I think we should caveat.

sryza (commented):
Looking close! The apidoc is really really good.

sryza:
Also, I'm excited to include this as part of the Dagster demo in my next talk

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
jamiedemaria: Since the query just gets the single latest asset materialization event, i dont need to use the timestamp as part of the query, and just use it to see if the returned asset materialization event is more recent. I could switch the cursor to the most recent storage id, that would work as well.

Question about that - lets say that assets A and B are parents to C and both A and B were materialized and B has the higher storage id (lets say 3). So then we set the cursor for C to 3. On the next tick when I query for asset materialization events for A can I pass in the `after_cursor` of 3 and it'll only return the events after storage id 3 even if that storage id doesn't correspond to a materialization of A? I would assume that should be fine but the event fetching is a bit of a black box to me so i dont know if there are any gotchas.

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
jamiedemaria: like if asset A and B both have a parent C? in that case yeah when we get the parent info for A we'll fetch the event for C and then when we get the parent info for B we'll make the same call for C. Definitely seems like a good place for caching, i'll add that

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
jamiedemaria: yeah that's clearer

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
sryza: right

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
sryza: It would definitely work on most of our event log storages. But it might not work with run-sharded event log storages. Thoughts @clairelin135 ?

jamiedemaria: actually, i think if i keep it so that i only fetch the latest materialization and then do a storage id comparison after (instead of passing `after_cursor`) i can cache all of these calls as well

python_modules/dagster/dagster/_core/storage/event_log/sqlite/sqlite_event_log.py
```diff
@@ -409,6 +409,10 @@ def alembic_version(self):
         with self.index_connection() as conn:
             return check_alembic_revision(alembic_config, conn)
 
+    @property
+    def is_sharded(self):
```
sryza, line 413: there are multiple ways it could theoretically be sharded - I think it would be clearer to call this `is_run_sharded`

sryza (approved):
Looks good aside from one comment I had!

PR #9568, opened 2022-08-31
UnresolvedMultiAssetSensor

clairelin135:
Built upon @jamiedemaria's work in creating an UnresolvedAssetSensor.

This PR enables merging partition asset functionality into the `MultiAssetSensor` by allowing the `MultiAssetSensor` to access assets from the repository via specified asset keys.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/sensor_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/sensor_decorator.py
@@ -3,15 +3,12 @@ from functools import update_wrapper
 from typing import TYPE_CHECKING, Callable, Optional, Sequence
 
 import dagster._check as check
-from dagster._core.definitions import AssetSelection
 from dagster._annotations import experimental
-<<<<<<< Updated upstream
-=======
+from dagster._core.definitions import AssetSelection
 from dagster._core.definitions.assets import AssetsDefinition
 from dagster._core.definitions.unresolved_asset_sensor_definition import (
     UnresolvedMultiAssetSensorDefinition,
 )
->>>>>>> Stashed changes
 from dagster._core.errors import DagsterInvariantViolationError
 
 from ...errors import DagsterInvariantViolationError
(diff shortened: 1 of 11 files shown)


python_modules/dagster/dagster/_core/definitions/decorators/sensor_decorator.py
```diff
@@ -249,9 +254,9 @@ def _wrapped_fn(context):
                     ).format(sensor_name=sensor_name, result=result, type_=type(result))
                 )
 
-        return MultiAssetSensorDefinition(
+        return UnresolvedMultiAssetSensorDefinition(
             name=sensor_name,
-            asset_keys=asset_keys,
+            asset_selection=AssetSelection.keys(*asset_keys),
```
clairelin135, line 259: probably worth considering if we should be accepting `AssetSelection` instead of `asset_keys` in `@multi_asset_sensor`

sryza:
@clairelin135 are you able to provide a little bit of background on the motivation for this change?

clairelin135:
@sryza Currently, in the existing `MultiAssetSensorDefinition` we only have access to the asset keys of the assets we want to monitor. If we want to merge in partition functionality (for example getting the latest materialization by partition), we need to have access to the `PartitionsDefinition` of an asset.

In the future, I think it might be cool to be able to do something like:
```
@multi_asset_sensor(asset_selection=AssetSelection.keys("asset_a").upstream(depth=1))
def my_sensor():
   pass
```
so that we can automatically infer the assets to monitor. In order to do this, we need to have the assets definitions available in the repository.

sryza:
Let me know if I'm thinking about this the wrong way, but an alternative could be to add an attribute an `SensorEvaluationContext` that returns the current `RepositoryDefinition`, and then have the body of the sensor use it to resolve the asset selection. Thoughts on how that compares with the approach here?

clairelin135:
@sryza how would the `SensorEvaluationContext` have a pointer to the current `RepositoryDefinition`? Would we ask the user to pass it in as a parameter, or would we add it to the `SensorEvaluationContext` at repository load time?

sryza:
I was imagining that we'd add it to the SensorEvaluationContext at repository load time?

clairelin135:
@sryza That makes sense. I reworked this to follow the approach of adding the repository definition pointer to the sensor context: https://github.com/dagster-io/dagster/pull/9596

Seems like there are net positives of going with this new approach (readability + ability to directly invoke sensors for testing)

PR #9577, opened 2022-09-01
Ignore warnings sent from the dagster module

dpeng817:
### Summary & Motivation
The gigantic slew of warnings at the end of our builds makes it super annoying to figure out what errors are. This attempts to significantly reduce the number that appear.

### Testing
You can see that it works by looking at the requisite build; the number of warnings is vastly reduced.

### DIFF
--- a/examples/assets_dbt_python/tox.ini
+++ b/examples/assets_dbt_python/tox.ini
@@ -16,7 +16,7 @@ allowlist_externals =
   /bin/bash
 commands =
   !windows: /bin/bash -c '! pip list --exclude-editable | grep -e dagster -e dagit'
-  pytest -vv
+  pytest -c ../../pyproject.toml -vv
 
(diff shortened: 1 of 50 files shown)


smackesey:
This is a good idea, but rather than cloning the config across all tox files, why not add a `tool.pytest.ini_options` [section](https://docs.pytest.org/en/6.2.x/customize.html#pyproject-toml) to `pyproject.toml` to match `pylint`/`mypy` and then point the tox invocation to that rather than copying the config across all tox files?

```
[testenv]
...
commands =
  ...
  pytest -c ../../../pyproject.toml ... # use the mypy/pylint to match the path
```

sryza:
This seems legit to me. Will defer to @smackesey for final approve.

smackesey (approved):

PR #9580, opened 2022-09-01
[dagit] Virtualized Jobs table

hellendag:
### Summary & Motivation

A proof of concept for a scalable virtualized table for showing a list of jobs, then lazy-loading the details when the job is scrolled into view. Storybook example contains 100 repos with 500 jobs each (50k jobs total), and a search input.

https://user-images.githubusercontent.com/2823852/188006648-7071999e-ae7f-4ddb-b4db-4287f6d767a2.mov

yarn storybook

### DIFF
--- a/js_modules/dagit/packages/core/package.json
+++ b/js_modules/dagit/packages/core/package.json
@@ -31,6 +31,7 @@
     "styled-components": "^5.3.3"
   },
   "dependencies": {
+    "@tanstack/react-virtual": "3.0.0-beta.18",
     "@vx/shape": "^0.0.192",
     "amator": "^1.1.0",
     "ansi-to-react": "^5.1.0",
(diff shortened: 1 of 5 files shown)


hellendag:
* master
  * **PR #9580** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9580" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9580?utm_source=stack-comment).

bengotow (approved):
This looks good to me - I know the LogsScrollingTable and the Suggest dropdown use a separate library called `react-virtualized`, but that one is pretty heavyweight (and in that table the rows are varying height which makes it a lot more complicated) 

I don't think it's bad to have both libraries, and I like that this one is only a hook and doesn't try to wrangle an entire Container component of it's own into the tree. React-virtualized has a lot of options and props, etc. that are really just letting you control things you could have controlled directly if you owned the container. 

We could probably also build this on top of our useViewport hook (that one powers the virtualization of the Gantt viz and PartitionStepStatus, etc.) That hook gives you the visible region of the element as state, but you'd need to measure the rows and slice down the array to the visible portion in another hook. Probably fine to just let this library do it!

PR #9585, opened 2022-09-02
Add list_objects to fake s3 resource

Vivanov98:
### Summary & Motivation

The Fake s3 resource currently only has list_objects_v2. The current s3 IO manager uses list_objects (and possibly other things do too) which means the fake resource can't be used everywhere.

This module doesn't have tests, but I confirmed locally that it works.

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws/s3/s3_fake_resource.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/s3/s3_fake_resource.py
@@ -28,13 +28,20 @@ class S3FakeSession:
         self.mock_extras.head_object(*args, **kwargs)
         return {"ContentLength": len(self.buckets.get(Bucket, {}).get(Key, b""))}
 
-    def list_objects_v2(self, Bucket, Prefix, *args, **kwargs):
-        self.mock_extras.list_objects_v2(*args, **kwargs)
+    def _list_objects(self, Bucket, Prefix):
         key = self.buckets.get(Bucket, {}).get(Prefix)
         if key:
             return {"KeyCount": 1, "Contents": [{"Key": key}], "IsTruncated": False}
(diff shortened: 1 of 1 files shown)


python_modules/libraries/dagster-aws/dagster_aws/s3/s3_fake_resource.py
```diff
@@ -36,6 +36,9 @@ def list_objects_v2(self, Bucket, Prefix, *args, **kwargs):
         else:
             return {"KeyCount": 0, "Contents": [], "IsTruncated": False}
 
+    def list_objects(self, Bucket, Prefix, *args, **kwargs):
```
jamiedemaria, line 39: we probably want to include a `self.mock_extras.list_objects(*args, **kwargs)` here. by doing that it means that it means that `self.mock_extras.list_objects(*args, **kwargs)` and `self.mock_extras.list_objects_v2(*args, **kwargs)` will both get called by this mock fn. To avoid that, could you factor out the body of `list_objects_v2` into a helper function that both `list_objects_v2` and `list_objects` call after the `mock_extras` call?

jamiedemaria (commented):
thanks for the contribution! had one small comment, but otherwise this is great

python_modules/libraries/dagster-aws/dagster_aws/s3/s3_fake_resource.py
Vivanov98: Ah good point! I missed that, yep updated that, let me know if it looks good

jamiedemaria (approved):
looks great! just need to wait for CI to pass, but if it doesn't reveal any errors, this is good to go!

Vivanov98:
Perfect, thanks a lot!

PR #8965, opened 2022-07-20
Remove some old stuff

smackesey:
### Summary & Motivation

This PR just removes some old no longer used stuff.

- Pipfile and Pipfile.lock in python_modules (not sure how this ended up there in the first place)
- `dagster-buildkite` integration pipeline, which we no longer use (was merged into the main pipeline months ago).
- `freezegun` in `dagster[test]` appears not to be used in our source or by CI commands
- `flake8` in `dagster[test]` was used in the `dagster` tox, but the check was very limited and redundant with mypy/pylint

### How I Tested These Changes

BK.

PR #9587, opened 2022-09-05
Convert singleton comparisons from equality to identity

chrisRedwine:
### Summary & Motivation

Hey there, and congrats on 1.0! 

I was diving into the codebase a bit and noticed several places (though not everywhere) in which comparisons to `None`, `True`, and `False` use equality `==` rather than identity `is`. Generally, this is ill-advised and can cause subtle bugs unless it's done intentionally/explicitly, which didn't appear to be the case from what I could tell (see [PEP8](https://peps.python.org/pep-0008/#programming-recommendations), flake8 [E711](https://www.flake8rules.com/rules/E711.html)/[E712](https://www.flake8rules.com/rules/E712.html), pylint [C0121](https://pylint.pycqa.org/en/latest/user_guide/messages/convention/singleton-comparison.html), etc.). Of course, if the team has already discussed and has a convention around this stuff, let me know.

For use of `== None` in the context of SQLAlchemy queries, I switched it to leverage the `is_(None)` column operator, which in addition to being consistent with the above also happens to be a bit faster since it doesn't need to go through the overloading of `__eq__` first just to call `is_` anyways. Similarly, I left the pandas examples that use `==` for boolean comparison alone, since that's a special case that actually does rely on overloading `__eq__`.

As a side note, I also noticed that sometimes tests assert truthiness/falsiness rather than testing against `True|False` specifically, though it's inconsistent. I'd be happy to change this so that boolean test assertions are all consistent one way or the other, but I figured it'd be best to ask first.

Thanks!

Tested against python 3.8 via `pytest python_modules/dagster/dagster_tests`

### DIFF
--- a/examples/docs_snippets/docs_snippets_tests/deploying_tests/test_job_retries.py
+++ b/examples/docs_snippets/docs_snippets_tests/deploying_tests/test_job_retries.py
@@ -20,5 +20,5 @@ def test_instance(docs_snippets_folder):
         os.path.join(docs_snippets_folder, "deploying", "dagster_instance")
     )
 
-    assert ref.settings["run_retries"]["enabled"] == True
+    assert ref.settings["run_retries"]["enabled"] is True
     assert ref.settings["run_retries"]["max_retries"] == 3
--- a/helm/dagster/schema/schema_tests/test_dagster_daemon.py
+++ b/helm/dagster/schema/schema_tests/test_dagster_daemon.py
@@ -217,7 +217,7 @@ def test_run_monitoring(
 
     instance = yaml.full_load(configmaps[0].data["dagster.yaml"])
 
-    assert instance["run_monitoring"]["enabled"] == True
+    assert instance["run_monitoring"]["enabled"] is True
 
 
 def test_run_retries(
@@ -233,7 +233,7 @@ def test_run_retries(
 
     instance = yaml.full_load(configmaps[0].data["dagster.yaml"])
 
-    assert instance["run_retries"]["enabled"] == True
+    assert instance["run_retries"]["enabled"] is True
 
 
 def test_daemon_labels(template: HelmTemplate):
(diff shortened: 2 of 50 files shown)


yuhan:
@chrisRedwine Thanks for the PR! Overall looking great. I'm kicking off tests.

Out of curiosity, what can it go wrong for `== None`? The pages linked say it's bad practice but don't explain why.

chrisRedwine:
@yuhan 

Well, in addition to being more explicit/direct, identity comparison is always going to be significantly faster since it's just a memory address lookup in C instead of an entire function call with comparison logic in python, etc. So, that gives it some initial weight as to why it should be preferred. 

But more importantly, when comparing against a singleton like `None` you typically (though not always) want to know if the result directly _is_ the object `None` and nothing else. However, if someone overrides `__eq__` for the object being compared against, then the results you get when comparing with `==` can be unexpected/unpredictable and could even break existing comparisons.

e.g., if someone decides to define some whacky `__eq__` on a class like this:
```python
class Wat:
     def __eq__(self, other):
         return other is None
```
then `Wat() == None` will be `True` even though it isn't actually `None`, whereas `Wat() is None` gives the expected result of `False`.

Or, what if at some point in the future in `python_modules/dagster-graphql/dagster_graphql/schema/backfill.py#L166-L172` the `graphene_info.context.instance.get_run_records` method somehow returned a numpy ndarray. The comparison using `if self._records == None:` on line 167 would actually raise a `ValueError`, due to the way that ndarrays define their `__eq__` method!

As a side note, I saw that flake8 checks were [removed](https://github.com/dagster-io/dagster/pull/8965) a little while ago with the reasoning that "the check was very limited and redundant with mypy/pylint". I think this was probably just because only a very limited subset of rules were being selected (E9,F63,F7,F82). I'd personally argue for a more robust set of flake8 checks that don't overlap with your pylint config, but that's just my $0.02 - obviously you all will have your own opinions 😄.

Hopefully that answers your question. Thanks!

chrisRedwine:
Hi @yuhan - no rush at all of course, I just wanted to double check that my response made sense? Thanks!

yuhan:
@chrisRedwine sorry for the late response, and thank you for the comprehensive answers. it makes total sense!
re-running the test suite now. i can take it from here.

yuhan (approved):

PR #9593, opened 2022-09-06
TimeWindowPartitionsDefinition from cron_schedule

sryza:
### Summary & Motivation

Fixes https://github.com/dagster-io/dagster/issues/7938.  This was requested on Slack twice in the last few days.

This allows you to use arbitrary cron expressions to define TimeWindowPartitionsDefinitions.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
@@ -1,4 +1,5 @@
-from datetime import datetime, time
+import re
+from datetime import datetime
 from typing import Any, Callable, Dict, Iterable, List, NamedTuple, Optional, Sequence, Union, cast
 
 import pendulum
@@ -6,7 +7,7 @@ import pendulum
 import dagster._check as check
 from dagster._annotations import PublicAttr, public
 from dagster._utils.partitions import DEFAULT_HOURLY_FORMAT_WITHOUT_TIMEZONE
-from dagster._utils.schedules import schedule_execution_time_iterator
+from dagster._utils.schedules import cron_string_iterator, schedule_execution_time_iterator
 
 from .partition import (
     DEFAULT_DATE_FORMAT,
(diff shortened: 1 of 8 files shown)


python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -29,45 +30,80 @@ class TimeWindow(NamedTuple):
 class TimeWindowPartitionsDefinition(
     PartitionsDefinition[TimeWindow],  # pylint: disable=unsubscriptable-object
     NamedTuple(
-        "_TimeWindowPartitions",
+        "_TimeWindowPartitionsDefinition",
```
gibsondan, line 33: looks like there are no backcompat concerns here since its not whitelisted for serdes, hooray

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -29,45 +30,80 @@ class TimeWindow(NamedTuple):
...
+            ("cron_schedule", PublicAttr[str]),
         ],
     ),
 ):
+    """
+    A set of partitions where each partitions corresponds to a time window.
+
+    The provided cron_schedule determines the bounds of the time windows. E.g. a cron_schedule of
+    "0 0 * * *" will result in daily partitions that start at midnight and end at midnight of the
+    following day.
```
gibsondan, line 48: the time window here is not necc. uniform here right? Like if you do 1AM and 3AM every Tuesday, you get some 2-hour long windows and some week-minus-two-hours long windows.

Would that be an argument for not including both start and end time and just including start time? Or is it still better to include both for consistency?

gibsondan (approved):

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
sryza: When you say "including", where do you mean? Do you mean like the general notion that this corresponds to a time window instead of a moment in time? Or arguments that are accepted somewhere?

PR #9596, opened 2022-09-06
Repository on Multi Asset Sensor Context

clairelin135:
Reworked https://github.com/dagster-io/dagster/pull/9568 to avoid creating an `UnresolvedMultiAssetSensor` and instead added a pointer to the repository definition from the `MultiAssetSensorEvaluationContext` allowing for resolving assets definitions from a set of selected asset keys to monitor.

Upsides of this approach:
- Sensors can be directly invoked by providing a context. With the previous approach, we had to add the sensor to a repository for resolution
- More readable than having an `UnresolvedMultiAssetSensor`

Downsides:
- Must pass `repository_def` into `build_multi_asset_sensor_context`

Seems like to me this is net positive and this approach is probably what we'd prefer.


python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -197,10 +200,22 @@ def __init__(
         last_run_key: Optional[str],
         cursor: Optional[str],
         repository_name: Optional[str],
+        repository_def: "RepositoryDefinition",
         asset_keys: Sequence[AssetKey],
         instance: Optional[DagsterInstance] = None,
     ):
+        self.repository_def = repository_def
```
sryza, line 207: Class attributes should generally be private (start with an underscore), so users can't overwrite them.

python_modules/dagster/dagster/_core/definitions/repository_definition.py
```diff
@@ -1272,6 +1289,10 @@ def has_sensor_def(self, name: str) -> bool:
     def source_assets_by_key(self) -> Dict[AssetKey, SourceAsset]:
         return self._repository_data.get_source_assets_by_key()
 
+    @property
+    def assets_by_key(self) -> Dict[AssetKey, "AssetsDefinition"]:
```
sryza, line 1293: Doesn't roll off the tongue quit as well, but I think it would be more clear to call this `_assets_defs_by_key`.  Also, prefer `Mapping` over `Dict`.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -301,6 +317,11 @@ def advance_all_cursors(self):
         materializations_by_key = self.latest_materialization_records_by_key()
         self.advance_cursor(materializations_by_key)
 
+    @public  # type: ignore
+    @property
+    def assets_by_key(self) -> Mapping[AssetKey, "AssetsDefinition"]:
```
sryza, line 322: I think this one is also worth changing. Alternatively, we could just expect users to do `context.repository_def.assets_defs_by_key`

sryza (approved):
I left a last comment that I think is worth addressing - otherwise LGTM.

clairelin135:
@sryza fixed, thanks!

PR #9603, opened 2022-09-06
Make daemon sensor test suite faster

gibsondan:
Summary:
use fixtures to not spin up a new instance/workspace for every single test - similar to the fixture setup in the scheduler tests.

### DIFF
--- a/python_modules/dagster/dagster_tests/daemon_sensor_tests/test_run_status_sensors.py
+++ b/python_modules/dagster/dagster_tests/daemon_sensor_tests/test_run_status_sensors.py
@@ -1,0 +1,812 @@
+import tempfile
+import time
+from contextlib import contextmanager
+
+import pendulum
+import pytest
+
+from dagster import DagsterRunStatus
+from dagster._core.events import DagsterEvent, DagsterEventType
+from dagster._core.events.log import EventLogEntry
+from dagster._core.scheduler.instigation import TickStatus
+from dagster._core.storage.event_log.base import EventRecordsFilter
+from dagster._core.test_utils import create_test_daemon_workspace, instance_for_test
+
+from .conftest import workspace_load_target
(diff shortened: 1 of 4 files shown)


python_modules/dagster/dagster_tests/daemon_sensor_tests/test_run_status_sensors.py
```diff
@@ -0,0 +1,812 @@
...
+
+@pytest.fixture(name="instance_module_scoped", scope="module")
+def instance_module_scoped_fixture():
+    # Overridden from conftest.py, uses DefaultRunLauncher since we care about
+    # runs actually completing for run status sensors
+    with instance_for_test(
+        overrides={},
+    ) as instance:
+        yield instance
+
```
gibsondan, line 37: this file just has all the run status sensor tests from the main file - both to keep the test from being unreadably massive, and also to override this fixture to actually launch the runs

python_modules/dagster/dagster_tests/daemon_sensor_tests/test_sensor_failure_recovery.py
```diff
@@ -59,175 +54,148 @@ def _test_launch_sensor_runs_in_subprocess(instance_ref, execution_datetime, deb
...
     frozen_datetime = to_timezone(
         create_pendulum_time(year=2019, month=2, day=28, hour=0, minute=0, second=1, tz="UTC"),
         "US/Central",
     )
 
-    with instance_with_sensors() as (
-        instance,
-        _grpc_server_registry,
-        external_repo,
-    ):
```
gibsondan, line 72: these changes look big but are just taking this out and using a shared fixture instead

gibsondan (commented):
appears to roughly halve test execution time

prha (approved):

PR #9605, opened 2022-09-06
Partition functionality on Multi Asset Sensor Context

clairelin135:
Augments `MultiAssetSensorEvaluationContext` to contain functionalities to work with asset partitions.

Adds a couple of methods:
- `latest_materialization_by_partition`: Given an asset key, returns a dict mapping partition key to the corresponding materialization
- `get_cursor_partition`: Gets the current partition the cursor is on
- `map_partition`: Given a partition key and the partitions definition it comes from, returns a list of corresponding partitions it maps to in a downstream partitions definition
- `get_partition_from_event_log_record`: Utility method to return the partition key (if exists) on an event log record

Updates the following existing methods to contain an `after_cursor_partition` flag that allows for retrieving only event log entries whose partitions are after the cursor partition. By default, `after_cursor_partition` is set to `True`.
- `materialization_records_for_key`
- `latest_materialization_records_by_key`


python_modules/dagster/dagster/_core/definitions/time_window_partition_mapping.py
```diff
@@ -101,7 +101,9 @@ def round_datetime_to_period(dt: datetime, period: ScheduleType) -> datetime:
     if period == ScheduleType.MONTHLY:
         return dt.replace(day=1, hour=0, minute=0, second=0, microsecond=0)
     elif period == ScheduleType.WEEKLY:
-        return (dt - timedelta(days=dt.weekday())).replace(
+        # isoweekday returns 1 for Monday, 7 for Sunday. Our weekly partitions start from every
```
clairelin135, line 104: Fixes a bug that I came across where mapping a partition key into its corresponding weekly partition key will result in a weekly partition beginning on Monday.

Our weekly partitions definitions start on Sunday. Updates this mapping to map to a weekly partition starting on Sunday.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -271,28 +313,222 @@ def latest_materialization_records_by_key(
...
+                # Only update if no previous materialization for this partition exists,
+                # to ensure we return the most recent materialization for each partition.
+                partition = self.get_partition_from_event_log_record(materialization)
+                if materialization_by_partition[partition] is None:
+                    materialization_by_partition[partition] = materialization
+
+        return materialization_by_partition
+
+    @public
+    def get_cursor_partition(self, asset_key):
```
sryza, line 427: type annotations plz

sryza (commented):
What would the code look like if a user wants to trigger runs that update partitions of asset X as soon as all of the upstream partitions are available?

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -219,6 +226,11 @@ def __init__(
                 )
             self._assets_by_key[asset_key] = assets_def
 
+        self._partitions_def_by_asset_key = {
+            asset_key: asset_def.partitions_def
```
jamiedemaria, line 230: if an asset isn't partitioned will `asset_def.partitions_def` be None?

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -248,15 +282,23 @@ def latest_materialization_records_by_key(
 
         if asset_keys is None:
             asset_keys = self._asset_keys
+        else:
+            asset_keys = check.opt_list_param(asset_keys, "asset_keys", of_type=AssetKey)
```
jamiedemaria, line 286: good catch

clairelin135:
@sryza Here are some code examples of what this would look like: https://elementl.quip.com/a8POAGGGLBSQ/Partitioned-Asset-Sensors

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -271,28 +313,222 @@ def latest_materialization_records_by_key(
...
+        return partition_key, cursor
+
+    @public
+    def latest_materialization_by_partition(
+        self,
+        asset_key: AssetKey,
+        after_cursor_partition: Optional[bool] = True,
+    ) -> Mapping[str, Optional["EventLogRecord"]]:
+        """
+        Fetches the most recent materialization event record for each partition of each asset.
```
jamiedemaria, line 363: it took me a minute to figure out why this fn was different than `latest_materiazations_by_key`, so we could probably make the doc string comments more explanatory. Maybe an example of what get returned, but not sure yet what that would look like

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -271,28 +313,222 @@ def latest_materialization_records_by_key(
...
                 asset_key=asset_key,
-                after_cursor=cursor_dict.get(str(asset_key)),
+                after_cursor=cursor,
+                asset_partitions=partitions_to_fetch,
             ),
             ascending=True,
             limit=limit,
         )
 
+    def _get_cursor(self, asset_key: AssetKey):
```
jamiedemaria, line 351: we could make `asset_key` optional (default to None) here and if it's None then we returned the `unpacked_cursor`. then we can just call this util everywhere we have the `cursor = json.loads(self.cursor) if self.cursor else {}` line which would be nice

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -306,7 +542,9 @@ def advance_cursor(
         """
         cursor_dict = json.loads(self.cursor) if self.cursor else {}
         update_dict = {
-            str(k): v.storage_id if v else cursor_dict.get(str(k))
+            str(k): (self.get_partition_from_event_log_record(v), v.storage_id)
```
jamiedemaria, line 545: one of the things i struggled with the most when i was writing this Context type was keeping track of what info the cursor contained and what the types of the keys/values were. now that the cursor is more complicated, i think it would be really helpful  for whoever works on this next to have a comment somewhere in the class that describes the format of the cursor and what info it contains. something like

```
The cursor maps stringified AssetKeys (str(asset_key)) to Tuples. The first value in the tuple is the partition of type X which means Y. The second value in the tuple is the storage_id (int) of the materialization event for the asset that has been most recently fetched from the DB
```

sryza:
If I'm understanding correctly, those code examples don't handle the case where we'd like to trigger downstream materializations of backfilled partitions. Is that right?  I.e. if you have daily partitions that you're filling in every day, but then someone notices a problem with a partition from a year ago and then materializes it.

clairelin135:
@sryza Yes, that's right. If you wanted to handle partitions being materialized out of order, you would do this:
```
context.latest_materialization_by_partition(asset_key, after_cursor_partition = False)
```
Which would search after the cursor, but not filter for partitions past the cursor (so older partitions would still show up). This would return a dictionary with all the partition keys mapped to the materialization, or None if the materialization doesn't exist.

I currently set the `after_cursor_partition` to be `True` by default, but wondering if we event want to be filtering for partitions past the cursor.

sryza:
> I currently set the after_cursor_partition to be True by default, but wondering if we even want to be filtering for partitions past the cursor.

If we do have after_cursor_partition, my instinct would be to default it to False. From the name of the method, I expected that it would return those materializations.

Do you have a code example for what it would look like for a user to trigger runs that update partitions of asset X as soon as all of the upstream partitions are available, including partitions that have already been processed in the past?

clairelin135:
Yep, that makes sense, I can change the default `after_cursor_partition` to be `False`. 

If you mean processing past partitions as in materializing the same partition multiple times, here's an example. If you kick off partition 2022-08-01 for `daily_asset`, the sensor will kick off that same partition for `daily_asset_2`. If on the next tick we materialize 2022-08-01 again for `daily_asset`, the sensor will kick off another run for `daily_asset_2`. 
```
@multi_asset_sensor(asset_keys=[daily_asset.key], job=daily_asset_2_job)
def trigger_daily_asset_from_daily_asset(context):
    materializations_by_partition = context.latest_materialization_by_partition(
        daily_asset.key, after_cursor_partition=False
    )

    # Get all corresponding daily partitions for any materialized weekly partitions
    for partition, materialization in materializations_by_partition.items():
        if materialization:
            daily_partitions = context.map_partition(
                partition,
                daily_asset_2_job.partitions_def,
            )
            for daily_partition in daily_partitions:
                yield daily_asset_job.run_request_for_partition(daily_partition, run_key=None)
                context.advance_cursor({daily_asset.key: materialization})
```
However, we don't allow for fetching events before the cursor. So lets say you wanted to kick off a weekly partition after all 7 upstream daily partitions are materialized. If you backfilled all 7 daily partitions and the sensor kicked off the weekly partition, rematerializing one daily partition would not kick off the weekly partition again.

sryza:
> However, we don't allow for fetching events before the cursor. So lets say you wanted to kick off a weekly partition after all 7 upstream daily partitions are materialized. If you backfilled all 7 daily partitions and the sensor kicked off the weekly partition, rematerializing one daily partition would not kick off the weekly partition again.

If we wanted to allow that, what would it take? A cursor component per partition or something?

clairelin135:
Hm.. seems like we need the ability to look behind a cursor and check for the existence of materializations.

We could add an additional helper method `all_partitions_materialized` that searches all events (including before the cursor) and returns if all partitions in a list have been materialized. So, for the example above, if a daily partition is returned from after the cursor, the user can invoke the `all_partitions_materialized` method to discern if all 7 daily partitions have existing materializations.

sryza:
@clairelin135 mind writing out what that would look like? I'm having a little bit of trouble visualizing it

clairelin135:
@sryza Yep, here's an example of what this could look like:
```
@partitioned_asset_sensor(assets=[daily_asset], job=weekly_asset_job)
def trigger_weekly_asset_from_daily_asset(context):
    # All materializations after cursor
    materializations_by_partition = context.latest_materialization_by_partition(daily_asset.key)

    # Get all materialized daily partitions after the cursor. Map each to the corresponding
    # weekly partition, and add all of these weekly partitions to a set to ensure uniqueness.
    evaluated_weekly_partitions = set()
    for (partition, materialization) in materializations_by_partition.items():
        if materialization:
            weekly_partition = context.map_partition(
                partition,
                weekly_asset_job.partitions_def,
            )[0]
            # We only want to evaluate each unique weekly partition once per tick
            if weekly_partition not in evaluated_weekly_partitions:
                evaluated_weekly_partitions.add(weekly_partition)

                daily_partitions_in_week = context.map_partition(
                    weekly_partition,
                    to_partitions=daily_asset.partitions_def,
                    from_partitions=weekly_asset.partitions_def,
                )
                if context.all_partitions_materialized(daily_partitions_in_week):
                    yield weekly_asset_job.run_request_for_partition(weekly_partition, run_key=None)
                    # Advance the cursor so we only check event log records past the cursor
                    context.advance_cursor({daily_asset.key: materialization})
```

sryza:
Some comments on that code example:
- When would `

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
sryza: Materializations should be plural to match our convention for naming dictionaries.  Also, if this is returning EventLogRecords, it should be `latest_materialization_records_by_partition`.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -271,28 +313,222 @@ def latest_materialization_records_by_key(
...
+    ) -> Mapping[str, Optional["EventLogRecord"]]:
+        """
+        Fetches the most recent materialization event record for each partition of each asset.
+
+        Args:
+            assets (Optional[Sequence[AssetsDefinition]]): list of assets to fetch events for.
+                If not specified, the latest partition materialization will be fetched for all
+                assets the partitioned_asset_sensor monitors.
+            after_cursor_partition (Optional[bool]): If True, only materializations with partitions
+                after the cursor's current partition will be returned. By default, set to True. For
```
sryza, line 370: For unpartitioned assets, would it ever make sense to call this function? Maybe we should error in that case?

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -271,28 +313,222 @@ def latest_materialization_records_by_key(
...
+    def latest_materialization_by_partition(
+        self,
+        asset_key: AssetKey,
+        after_cursor_partition: Optional[bool] = True,
+    ) -> Mapping[str, Optional["EventLogRecord"]]:
+        """
+        Fetches the most recent materialization event record for each partition of each asset.
+
+        Args:
+            assets (Optional[Sequence[AssetsDefinition]]): list of assets to fetch events for.
```
sryza, line 366: This documented argument doesn't appear to match the function signature.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -271,28 +313,222 @@ def latest_materialization_records_by_key(
...
+
+            downstream_partitions = partition_keys[
+                partition_keys.index(downstream_partition_key_range.start) : partition_keys.index(
+                    downstream_partition_key_range.end
+                )
+                + 1
+            ]
+            return downstream_partitions
+
+    @public
```
sryza, line 517: Does this need to be public?

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
sryza: Why does the cursor need to include the partition key? In theory we could get it by pulling the event using the storage ID? Or is it an optimization?

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -271,28 +313,222 @@ def latest_materialization_records_by_key(
...
+    def get_partition_from_event_log_record(
```
sryza, line 518: Nitpick: `get_partition_key...` would be clearer

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -271,28 +313,222 @@ def latest_materialization_records_by_key(
...
+
+    def _can_map(self, to_partitions: PartitionsDefinition, from_partitions: PartitionsDefinition):
+        if isinstance(to_partitions, TimeWindowPartitionsDefinition) and isinstance(
+            from_partitions, TimeWindowPartitionsDefinition
+        ):
+            return True
+        return False
+
+    @public
+    def map_partition(
```
sryza, line 453: A couple thoughts on the signature of this method:
- A name that indicates what it returns would be most clear. E.g. `get_downstream_partition_keys` or something?
- It might save the user some characters if it accepts an `AssetKey` instead of a `PartitionsDefinition`?

sryza (commented):
Thanks for bearing with me on all these comments - it's an inherently complex problem, so trying to figure out how we can maximally relieve the burden on the user. I'm excited about where this is going.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
clairelin135: Yep, we're raising an error on line 396

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
clairelin135: Yep, I can change the name of the method!

Agree that it is easier to accept an asset key. But there's a situation that this might be a little awkward. So a user might pass in the partitions definition of the job the sensor yields a run request for. For example:

```
@multi_asset_sensor(asset_keys=[AssetKey("a")], job_name="asset_b_and_c_job")
def my_sensor(context):
   ...
   context.map_partition(partition_key, to_partitions=asset_b_and_c_job.partitions_def)
```
We could also allow the sensor to have access to the partitions definition of the targetted job, but it seems a little odd to have to select one asset out of the job. I guess another option is to add an additional param `from_job_partitions`?

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
clairelin135: It's an optimization--if a user wants to see what partition key the cursor is on, storing the partition key of the cursor allows us to avoid refetching the event log record from storage.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
clairelin135: Yep, we raise an error in every time we try to fetch a partitions definition but it is None

clairelin135:
@sryza @jamiedemaria just updated this diff per the feedback here and on slack! Included changes:
- docstring / type annotation updates
- added `all_partitions_materialized` method
- function renames

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
sryza: I guess what I mean is: does it need to be user-facing"

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
sryza: Makes sense. Now I can't recall why I thought this didn't make sense.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
clairelin135: I think it would be a helpful utility method if users want to fetch the partition off an event log record

sryza: Got it. The reason I'm pushing back against this is that it feels awkward to me for this to be a method of the sensor context, given that it doesn't use `self` at all.  If anything, it seems like maybe it should be a method of the EventLogRecord?  Although, by that logic, we might end up adding a lot of other methods to EventLogRecord to get other assorted data within the event? Or we could provide a method that returns a mapping of (asset key, partition key) to EventLogRecord?  Or, if we don't think that users will need this functionality in common cases, we could keep it private to punt on figuring it out.

clairelin135: I think we should add these methods to `EventLogRecord` eventually. But in terms of the multi asset sensor use case, I think we can just make this method private. In most cases I think users will use `latest_materialization_records_by_partition`, which will return a dict of partition key -> materialization, so we don't need to make this utility available

clairelin135:
@sryza Just added more updates per our discussion on slack. Latest changes:
- `get_partition_key_from_event_log_record` is now a private method
- `map_partitions` is renamed to `get_downstream_partition_keys`, which accepts `to_asset_key` and `from_asset_key`.

`get_downstream_partition_keys` will use the partition mapping to the upstream asset if it exists. Some trickiness here is our execution APIs use the default partition mapping if the user does not specify a partition mapping.

I know you mentioned previously that we should not infer the partition mapping, but I think it would be awkward if users are relying on our inferred partition mappings for execution (e.g. a user has a weekly partitioned asset with an upstream daily partitioned asset dependency, but is using the default partition mapping) but find that in this sensor we require the downstream asset to have an explicit partition mapping in order to find mapped partition keys. So for now, I've kept the current version of the code to use the default partition mapping if one isn't explicitly defined.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -271,28 +327,282 @@ def latest_materialization_records_by_key(
...
+        for that partition. Fetches only materializations after the cursor.
+
+        Args:
+            asset_key (AssetKey): The asset to fetch events for.
+            after_cursor_partition (Optional[bool]): If True, only materializations with partitions
+                after the cursor's current partition will be returned. By default, set to False. For
+                unpartitioned assets, this parameter is ignored.
+
+        Returns:
+            Mapping[AssetKey, Mapping[str, EventLogRecord]]:
```
sryza, line 410: This should be Mapping[str, EventLogRecord], right?

sryza (approved):
This looks good! 

> I know you mentioned previously that we should not infer the partition mapping, but I think it would be awkward if users are relying on our inferred partition mappings for execution (e.g. a user has a weekly partitioned asset with an upstream daily partitioned asset dependency, but is using the default partition mapping) but find that in this sensor we require the downstream asset to have an explicit partition mapping in order to find mapped partition keys. So for now, I've kept the current version of the code to use the default partition mapping if one isn't explicitly defined.

If asset A depends on asset B, but an explicit partition mapping wasn't provided for that dependency, I think using the default is definitely the right the to do: because that's the partition mapping that would be used for that dependency elsewhere.  The thing that I think is a little weird is that if A doesn't depend on B, we'd still return a partition mapping for it - in that case, I think of the partition mapping as undefined.  Not a huge deal though.

sryza:
I think adding examples of this to the docs will be important.

clairelin135:
Yep! Will push out a PR next updating docs to contain examples for this

PR #9614, opened 2022-09-07
Return last run correclty in job table

gibsondan:
Summary:
runs[0] is the earliest run, not the latest run

Test Plan:
View the status overview page when there are more than 1 run, latest run tab is now correct


johannkm:
Should we sanity check other usages? Are these returned in the opposite order? https://github.com/dagster-io/dagster/blob/master/js_modules/dagit/packages/core/src/instigation/InstigationUtils.tsx#L22

alangenfeld (approved):
the one linked above is `runs(limit: 1)` so should be fine

prha (approved):

PR #9615, opened 2022-09-07
remove flaky grpc test

gibsondan:
Summary:
Unclear to me what this was intended to test - no guarantee that if a server shuts down that a long-running streaming request won't terminate, it just waits for any launched runs to finish (then stops the server with a grace period)

Test Plan: BK


alangenfeld:
https://dagster.phacility.com/D3946
seems like some ideas that were maybe never seen through

alangenfeld (approved):

PR #9626, opened 2022-09-07
Add scheduler_name to k8s container context

johannkm:


helm/dagster/charts/dagster-user-deployments/templates/helpers/_helpers.tpl
```diff
@@ -129,6 +129,9 @@ DAGSTER_K8S_PIPELINE_RUN_ENV_CONFIGMAP: "{{ template "dagster.fullname" . }}-pip
     {{- if .resources }}
     resources: {{- .resources | toYaml | nindent 6 }}
     {{- end }}
+    {{- if .schedulerName }}
+    scheduler_name: {{ .schedulerName }}
+    {{- end }}
```
gibsondan, line 134: this won't be set unless they explicitly override it right? the default case is for it to not be included?

python_modules/libraries/dagster-k8s/dagster_k8s/job.py
```diff
@@ -653,6 +663,11 @@ def construct_dagster_k8s_job(
         "service_account_name", job_config.service_account_name
     )
 
+    scheduler_name = (
+        pod_spec_config.pop("scheduler_name", job_config.scheduler_name)
+        or job_config.scheduler_name
```
gibsondan, line 668: we don't need both the second arg to pop and the or right?

python_modules/libraries/dagster-k8s/dagster_k8s/job.py
```diff
@@ -673,6 +688,7 @@ def construct_dagster_k8s_job(
                 "restart_policy": "Never",
                 "containers": [container_config] + user_defined_containers,
                 "volumes": volumes,
+                "scheduler_name": scheduler_name,
```
gibsondan, line 691: I assume nothing bad/different happens when this is set to None vs. the key being missing entirely? In both cases it falls back to the default? lots of stuff would break if it iddn't

gibsondan (approved):

helm/dagster/charts/dagster-user-deployments/templates/helpers/_helpers.tpl
johannkm: correct

python_modules/libraries/dagster-k8s/dagster_k8s/job.py
johannkm: good catch

python_modules/libraries/dagster-k8s/dagster_k8s/job.py
johannkm: yeah- I think it's None by default. I suppose to be extra safe we can put it behind a merge dict

PR #9613, opened 2022-09-07
ExternalRepository clean-up

alangenfeld:
Working from https://github.com/dagster-io/dagster/pull/9563/, this tries to get `ExternalRepository` cleaned up a bit before making changes. Starting from trying to remove the methods on `ExternalRepository` that distinguish `job` from `pipeline` in confusing ways, I ended up cleaning up a lot of CLI code branches that were no longer triggered due to https://github.com/dagster-io/dagster/pull/9170 . I did a set of renames to try to leave things in a reasonable spot. 

### How I Tested These Changes

bk

PR #9627, opened 2022-09-07
[grpc] use sync request for repository load

alangenfeld:
We set up this application level streaming before investigating the grpc settings we needed to be able to load large repositories.

The grpc streaming doesn't compress so we end up sending a much larger payload over the network in manual chunks. 

pytest python_modules/dagster/dagster_tests/api_tests/test_api_snapshot_repository.py::test_giant_external_repository_grpc

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql_tests/graphql/test_context.py
+++ b/python_modules/dagster-graphql/dagster_graphql_tests/graphql/test_context.py
@@ -42,7 +42,7 @@ def test_can_reload_on_external_repository_error():
                 # note it where the function is *used* that needs to mocked, not
                 # where it is defined.
                 # see https://docs.python.org/3/library/unittest.mock.html#where-to-patch
-                "dagster._core.host_representation.repository_location.sync_get_streaming_external_repositories_data_grpc"
+                "dagster._core.host_representation.repository_location.sync_get_external_repositories_data_grpc"
             ) as external_repository_mock:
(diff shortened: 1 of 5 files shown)


alangenfeld:
* master
  * **PR #9613** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9613" target="_blank">Graphite</a> 
    * **PR #9627** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9627" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9627?utm_source=stack-comment).

gibsondan (approved):

gibsondan (approved):
does this same logic apply (less urgently) to the other API calls that chunk in client.py? partition sets / schedules / sensors

alangenfeld:
> does this same logic apply (less urgently) to the other API calls that chunk in client.py? partition sets / schedules / sensors

ya but since those don't have a sync variant sitting around on old version servers - we would have to do add the sync variants and do some sort of versioning handshake dance to make sure we call the right one. Not worth for now but something we can consider once we have to do version tracking for a good reason.

alangenfeld (commented):
upon further testing - this will be a breaking change for users with location snapshots > 50MB uncompressed. 

There is an env var to override that limit - so its something we could do a on a 1.X release but not patch.

will update to only use sync on the deferred snapshot variant.

PR #9630, opened 2022-09-07
[grpc] deferred snapshot support

alangenfeld:
Adds a `defer_snapshots` arg to `ExternalRepository` and a new grpc method `ExternalJob` for loading the data/snapshot for single job.

origin prototype https://github.com/dagster-io/dagster/pull/9563

added test


alangenfeld:
* master
  * **PR #9613** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9613" target="_blank">Graphite</a> 
    * **PR #9627** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9627" target="_blank">Graphite</a> 
      * **PR #9630** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9630" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9630?utm_source=stack-comment).

python_modules/dagster/dagster/_core/host_representation/external.py
```diff
@@ -216,7 +272,12 @@ def solid_names_in_topological_order(self):
 
     @property
     def external_pipeline_data(self):
-        return self._external_pipeline_data
+        if self._data is None:
+            if self._ref is None or self._ref_to_data_fn is None:
+                check.failed("unexpected state - unable to load data from ref")
+            self._data = self._ref_to_data_fn(self._ref)
```
gibsondan, line 278: worried about thread safety here - is there anywhere where two different threads call this on the same object? Before it was immutable

python_modules/dagster/dagster/_grpc/client.py
```diff
@@ -252,10 +256,31 @@ def external_repository(self, external_repository_origin):
...
+            defer_snapshots=defer_snapshots,
         )
 
         return res.serialized_external_repository_data
 
+    def external_job(
+        self,
+        external_repository_origin: ExternalRepositoryOrigin,
+        job_name: str,
+    ) -> str:
```
gibsondan, line 268: what's the back-compat plan here in OSS - to make this feature opt-in and if you opt-in you need to make sure everything is upgraded? or do so some kind of fallback in the future if we add it?

gibsondan (commented):
this broadly makes sense to me other than the thread-safety q - do we possibly need a lock? i think the daemon at least passes a workspace around and expects to be able to call these methods on it

python_modules/dagster/dagster/_core/host_representation/external.py
gibsondan: maybe this won't cause issues other than possibly duplicated work if its called from two different threads? still spooky

python_modules/dagster/dagster/_grpc/client.py
alangenfeld: This grpc method is only called when the deferred snapshots responses comes back, so we only call it when we we know we are hitting a recent-enough grpc server. 

If an external_job call got added somewhere else that could be a problem.

python_modules/dagster/dagster/_core/host_representation/external.py
alangenfeld: ya agree it shouldn't break anything but could cause repeat fetches

python_modules/dagster/dagster/_core/host_representation/external_data.py
```diff
@@ -48,29 +48,28 @@ class ExternalRepositoryData(
         "_ExternalRepositoryData",
         [
             ("name", str),
-            ("external_pipeline_datas", Sequence["ExternalPipelineData"]),
             ("external_schedule_datas", Sequence["ExternalScheduleData"]),
             ("external_partition_set_datas", Sequence["ExternalPartitionSetData"]),
             ("external_sensor_datas", Sequence["ExternalSensorData"]),
             ("external_asset_graph_data", Sequence["ExternalAssetNode"]),
+            ("external_pipeline_datas", Optional[Sequence["ExternalPipelineData"]]),
```
gibsondan, line 55: re-ordering doesn't confuse serdes i assume

python_modules/dagster/dagster/_grpc/protos/api.proto
```diff
@@ -168,3 +170,12 @@ message StartRunReply {
...
 }
+
+message ExternalJobRequest {
+  string serialized_repository_origin = 1;
+  string job_name = 2;
+}
+
+message ExternalJobReply {
+  string serialized_response = 1;
+}
```
gibsondan, line 181: per feedback on other diff - could imagine having two fields, one for good, one for error, vs. parsing the string looking for clues that it's a serialized error - not the end of the world

python_modules/dagster/setup.py
```diff
@@ -87,7 +87,7 @@ def get_version() -> str:
             "test": [
                 "coverage==5.3",
                 "docker",
-                "grpcio-tools",
+                "grpcio-tools>=1.32.0,<1.44.0",  # related to above grpcio pins
```
gibsondan, line 90: related?

gibsondan (approved):

python_modules/dagster/setup.py
alangenfeld: ya needed this to run the proto gen script

PR #9631, opened 2022-09-07
Add script to pull pypi changes in the last N days

gibsondan:
Summary:
Quick and dirty way to identify upstream deps that might have broken us recently.


johannkm (approved):
lovely

PR #9635, opened 2022-09-08
[Tooltip] Make tooltip show/hide logic controlled with wrapper div

salazarm:
### Summary & Motivation

Issue: [Blueprint tooltips on disabled buttons stay open ](https://github.com/palantir/blueprint/issues/2779#issuecomment-411446961). 

Solution:
Make tooltip controlled and handle showing/hiding it ourselves. I used a wrapper div with `onMouseLeave` and `onMouseEnter` handlers and the behavior is the same as before (transitions in/out).

I put this behavior under a flag using a prop `useDisabledButtonTooltipFix` so that we can roll this out incrementally and test each call site. We have to do it this way because adding the wrapper div can break CSS so we'll need to test it everywhere that we enable it and potentially fix css

I used the attribute in cloud and saw it fixed the issue

### DIFF
--- a/js_modules/dagit/packages/ui/src/components/Tooltip.tsx
+++ b/js_modules/dagit/packages/ui/src/components/Tooltip.tsx
@@ -59,7 +59,7 @@ export const Tooltip: React.FC<Props> = (props) => {
 
   if (useDisabledButtonTooltipFix) {
     return (
-      <div onMouseOut={() => setIsOpen(false)} onMouseEnter={() => setIsOpen(true)}>
+      <div onMouseLeave={() => setIsOpen(false)} onMouseEnter={() => setIsOpen(true)}>
         {styledTooltip}
       </div>
     );
--- a/js_modules/dagit/packages/ui/src/components/Tooltip.tsx
+++ b/js_modules/dagit/packages/ui/src/components/Tooltip.tsx
@@ -28,17 +28,17 @@ const overwriteMerge = (destination: any[], source: any[]) => source;
 interface Props extends Tooltip2Props {
   display?: React.CSSProperties['display'];
   canShow?: boolean;
+  useDisabledButtonTooltipFix?: boolean;
 }
 
 export const Tooltip: React.FC<Props> = (props) => {
-  const {children, display, canShow = true, ...rest} = props;
+  const {useDisabledButtonTooltipFix = false, children, display, canShow = true, ...rest} = props;
 
-  if (!canShow) {
-    return <>{children}</>;
-  }
+  const [isOpen, setIsOpen] = React.useState<undefined | boolean>(undefined);
 
-  return (
+  const styledTooltip = (
     <StyledTooltip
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/ui/src/components/Tooltip.tsx
```diff
@@ -28,17 +28,21 @@ const overwriteMerge = (destination: any[], source: any[]) => source;
 interface Props extends Tooltip2Props {
   display?: React.CSSProperties['display'];
   canShow?: boolean;
+  useDisabledButtonTooltipFix?: boolean;
```
salazarm, line 31: We should use this anytime we wrap a button that can be disabled with a Tooltip

salazarm: (eventually we can get rid of this prop when we test every callsite with this prop set to `true`)

js_modules/dagit/packages/ui/src/components/Tooltip.tsx
```diff
@@ -28,17 +28,21 @@ const overwriteMerge = (destination: any[], source: any[]) => source;
...
     return <>{children}</>;
   }
 
-  return (
+  const styledTooltip = (
     <StyledTooltip
+      isOpen={isOpen}
       {...rest}
       minimal
       $display={display}
```
bengotow, line 48: I think this prop probably needs to be applied to the wrapper div!

bengotow (approved):
It's a bummer we have to fix this - it seems like a bug in Chrome.. but it seems like it's been an issue for a while and is unlikely to go away :-( 

It's interesting looking at the DOM tree for Tooltip, because Blueprint already wraps the target in a div: `<div class="bp3-popover-target"><button type="button" disabled="" /></div`

I wish we could jam these onMouseEnter / Leave listeners onto that div rather than creating another one, but it doesn't seem possible.

We might want to make the additional wrapper div `display: inline-block` by default to match blueprint's wrapper and increase the odds it plays nicely in our existing containers.

Good call on a step-by-step roll-out. 👍

js_modules/dagit/packages/ui/src/components/Tooltip.tsx
salazarm: `isOpen` gets passed to blueprint's Tooltip component. It switches the Tooltip to be a controlled rather than uncontrolled component

salazarm: oh you meant `$display`?

salazarm:
> I wish we could jam these onMouseEnter / Leave listeners onto that div rather than creating another one, but it doesn't seem possible.

Let me look into if its possible.

salazarm:
yeah doesn't seem possible.. :\ 
I kinda of want to put a patch up in Blueprint... like why are the events on the button instead of the wrapper 🤔

PR #9642, opened 2022-09-08
context.selected_output_names for graph-backed asset contexts

clairelin135:
Enables op execution contexts within graph-backed assets to access a `selected_output_names` property, allowing optional outputs to be yielded within ops when a subset of asset outputs are selected.

For example:
```
@op(out={"out_1": Out(is_required=False), "out_2": Out(is_required=False)})
def two_outputs_op(context):
    if "out_1" in context.selected_output_names:
        yield Output(1, output_name="out_1")
    if "out_2" in context.selected_output_names:
        yield Output(1, output_name="out_2")

@graph(out={"asset_one": GraphOut(), "asset_two": GraphOut()})
def two_assets():
    out_1, out_2 = two_outputs_op()
    return {"asset_one": out_1, "asset_two": out_2}

AssetsDefinition.from_graph(two_assets, can_subset=True)
```
Adds a `downstream_dep_assets(node_handle, output_name)` method to `AssetLayer` to allow fetching for downstream dependent assets for a given output. In the example above, this method allows `two_outputs_op` to know that `out_2` is an upstream dependency of asset_2 and that `out_1` is an upstream dependency of asset_1.

This PR does this by refactoring the existing `asset_key_to_dep_node_handles` method to also return the upstream node output handle dependencies for each node output handle. This addresses more advanced edge cases, for example:
- Many intermediate ops before yielding an asset
- Nested graph-backed assets
- Reusing an asset output as an input to another asset within the same graph-backed asset

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_layer.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_layer.py
@@ -1,4 +1,3 @@
-from platform import node
 import warnings
 from typing import (
     TYPE_CHECKING,
@@ -535,6 +534,8 @@ class AssetLayer:
             value_type=str,
         )
 
+        # Used to store the asset key dependencies of op node handles within graph backed assets
+        # See AssetLayer.downstream_dep_assets for more information
         self._node_output_handle_to_dep_asset_keys = check.opt_dict_param(
             node_output_handles_to_dep_asset_keys,
             "node_output_handles_to_dep_asset_keys",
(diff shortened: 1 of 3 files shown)


OwenKephart (approved):
this entire space twists my brain in knots but this looks great! yet another addition to the big bag of asset layer dictionaries lol

clairelin135:
@OwenKephart my brain is also twisted in knots 😂

PR #9645, opened 2022-09-09
add api doc for run_request_for_partition

sryza:

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/job_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/job_definition.py
@@ -574,7 +574,7 @@ class JobDefinition(PipelineDefinition):
         """
         Creates a RunRequest object for a run that processes the given partition.
 
-        Args
+        Args:
             partition_key: The key of the partition to request a run for.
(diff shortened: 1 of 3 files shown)


clairelin135:
Can we also make this show up in the docs? Noticed that it doesn't seem to appear in the run request api docs

sryza:
@clairelin135 when the release runs, it will automatically become part of the JobDefinition and UnresolvedAssetJobDefinition API docs (because it's a method of them). Is that what you're thinking about, or is there somewhere you think it would be useful for it to show up as well?

clairelin135:
@sryza I think it would be nice for it to show up here: https://docs.dagster.io/_apidocs/schedules-sensors# since we show `RunRequests`

sryza:
> @sryza I think it would be nice for it to show up here: https://docs.dagster.io/_apidocs/schedules-sensors# since we show RunRequests

That makes sense. I added a note to the docstring for RunRequest.

sryza:
@clairelin135 mind taking another look?

clairelin135 (approved):
Looking good!

PR #9658, opened 2022-09-11
infer graph input types from inner input types

sryza:
### Summary & Motivation

Makes this possible:
```
from dagster import op, graph, Nothing, In, AssetsDefinition, repository


@op(ins={"in1": In(Nothing)})
def op1():
    ...


@graph
def graph1(in1):
    return op1(in1)


@repository
def repo():
    return [AssetsDefinition.from_graph(graph1)]
```

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/graph_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/graph_definition.py
@@ -1,4 +1,4 @@
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 from typing import (
     TYPE_CHECKING,
     AbstractSet,
@@ -812,6 +812,8 @@ def _validate_in_mappings(
     input_defs_by_name: Dict[str, InputDefinition] = OrderedDict()
     mapping_keys = set()
 
+    target_input_defs_by_graph_input_name: Dict[str, List[InputDefinition]] = defaultdict(list)
+
     for mapping in input_mappings:
         # handle incorrect objects passed in as mappings
         if not isinstance(mapping, InputMapping):
@@ -842,6 +844,7 @@ def _validate_in_mappings(
             )
 
         target_input_def = target_node.input_def_named(mapping.maps_to.input_name)
+        target_input_defs_by_graph_input_name[mapping.graph_input_name].append(target_input_def)
         solid_input_handle = SolidInputHandle(target_node, target_input_def)
 
         if mapping.maps_to_fan_in:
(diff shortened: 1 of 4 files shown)


python_modules/dagster/dagster/_core/definitions/graph_definition.py
```diff
@@ -859,42 +856,45 @@ def _validate_in_mappings(
...
-                            )
+                            f"Unsatisfied MappedInputPlaceholder at index {idx} in "
+                            f"MultiDependencyDefinition for '{input_handle.node_name}.{input_handle.input_name}'"
                         )
 
-    return input_mappings, list(input_def_dict.values())
+    # if the dagster type on a graph input is None and all its target inputs have the
+    # same dagster type, then use that dagster type for the graph input
+    for graph_input_name, graph_input_def in input_defs_by_name.items():
+        if graph_input_def.dagster_type.kind == DagsterTypeKind.ANY:
```
alangenfeld, line 887: we have a `_type_not_set` bool on `InputDefinition` you can use to check for unset type. The case of concern would be an explicit `Any`

python_modules/dagster/dagster/_core/definitions/input.py
```diff
@@ -323,6 +323,19 @@ def combine_with_inferred(self, inferred: InferredInputProps) -> "InputDefinitio
...
+            name=self.name,
+            dagster_type=dagster_type,
+            description=self.description,
+            default_value=self.default_value if self.has_default_value else NoValueSentinel,
+            root_manager_key=self._root_manager_key,
+            metadata=self._metadata,
+            asset_key=self._asset_key,
+            asset_partitions=self._asset_partitions_fn,
+            input_manager_key=self._input_manager_key,
+        )
```
alangenfeld, line 337: did you consider re-using `combine_with_inferred` either as the call or in this impl? Main motivation would be to reduce spooky copy-all-the-properties-over call sites.

alangenfeld (changes requested):
add tests for 
* explicit type (not Any or Nothing) on graph 
* explicit Any on graph 

I know we want to remove the typing at that layer  - but I think its good to encode the expected behavior in test until we pull the trigger on removal.

python_modules/dagster/dagster/_core/definitions/graph_definition.py
sryza: I actually tried that first, but it didn't work for some reason. I think it's not making it through due to some sequence of `combine_with_inferred` calls or something, and I was hesitant to interrupt that delicate ecosystem.

My thinking was basically that it would be fine to override explicit Any. If I understand correctly, the only consequence ares that:
- If all the inner types are Nothing, then you don't need an upstream output, which I think should be true independent of the outer type.
- If all the inner types are the same Dagster type, then you can use the type loader on that type, which I think should be true independent of the outer type.

Anything I'm missing here?

alangenfeld: I cant think of anything, just update the comment `if the dagster type on a graph input is None` to match then.

sryza:
I updated the comment and added tests. I chose not to not combine with combine_with_inferred, because I think it makes the code net gnarlier

alangenfeld (approved):
Does this cause the types to show up correctly in dagit or is that another code path?

sryza:
It does appear to cause the types to show up correctly in Dagit

PR #9655, opened 2022-09-09
[dagster-airbyte] Support representing normalization-created tables as Assets

benpankow:
## Summary

Adds two new optional parameters to `build_airbyte_assets`.

- `upstream_assets` optionally allows a user to specify one or more upstream `AssetKey`s that the Airbyte assets should depend on, this can be used to e.g. represent the source data dependency.
- `normalization_tables` is a Mapping which maps a destination table to a set of one or more derived tables created by the [Airbyte basic normalization](https://docs.airbyte.com/understanding-airbyte/basic-normalization/#nesting) process. These normalization-generated tables will be shown as Assets downstream of their parent table, and materializations will be triggered by the parent table materializing.

Screen Shot 2022-09-09 at 4 14 47 PM

## Test Plan

Tested locally; new unit test.

PR #9649, opened 2022-09-09
[dagster-airbyte] Add config to avoid pulling/forwarding Airbyte logs to compute log

benpankow:
### Summary & Motivation

Currently, we poll the [`/jobs/get`](https://airbyte-public-api-docs.s3.us-east-2.amazonaws.com/rapidoc-api-docs.html#post-/v1/jobs/get) API endpoint to get the status of our Airbyte sync. Unfortunately, this endpoint pulls all the Airbyte logs, which can mean the API takes several seconds to return for long (hours) running syncs.

This PR introduces an optional config for the Airbyte resource that disables fetching logs. Under the hood, this instead hits the [`/jobs/list`](https://airbyte-public-api-docs.s3.us-east-2.amazonaws.com/rapidoc-api-docs.html#post-/v1/jobs/list) API endpoint, which lists recent syncs but provides us the same info minus logs. This will hopefully be faster for users hitting this limitation (though it's a bit hacky).

### How I Tested These Changes

Added params to existing unit tests, tested locally with both config options.

PR #9664, opened 2022-09-12
[dagster-airbyte] Add ability to generate Airbyte assets from YAML files

benpankow:
## Summary

Adds a new `@experimental` API that lets users load a set of Airbyte assets from an Airbyte project folder containing YAML files pulled using the [Octavia CLI](https://github.com/airbytehq/airbyte/tree/master/octavia-cli#octavia-import-all).

```python
airbyte_assets, source_assets = load_assets_from_airbyte_project(
    AIRBYTE_PROJECT_DIR, 
)
```

Screen Shot 2022-09-12 at 2 53 23 PM

Also provides a few optional config arguments to let users customize the output asset group names, whether or not to show normalization-created tables as separate assets etc.

## Test Plan

Tested locally; new unit test.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/load_assets_from_modules.py
+++ b/python_modules/dagster/dagster/_core/definitions/load_assets_from_modules.py
@@ -318,7 +318,7 @@ def with_group(
     """
     group_name = check.opt_str_param(group_name, "group_name")
     if not group_name:
-        return list(assets_defs)
+        return assets_defs
 
     return [
         assets_def.with_prefix_or_group(
--- a/python_modules/dagster/dagster/_core/definitions/load_assets_from_modules.py
+++ b/python_modules/dagster/dagster/_core/definitions/load_assets_from_modules.py
@@ -192,7 +192,7 @@ def load_assets_from_package_module(
     if key_prefix:
         assets = prefix_assets(assets, key_prefix)
     if group_name:
-        assets = with_group(assets, group_name)
+        assets = list(with_group(assets, group_name))
         source_assets = [asset.with_group_name(group_name) for asset in source_assets]
 
     return [*assets, *source_assets]
--- a/python_modules/dagster/dagster/_core/definitions/load_assets_from_modules.py
+++ b/python_modules/dagster/dagster/_core/definitions/load_assets_from_modules.py
@@ -192,7 +192,7 @@ def load_assets_from_package_module(
     if key_prefix:
         assets = prefix_assets(assets, key_prefix)
     if group_name:
-        assets = group_assets(assets, group_name)
+        assets = with_group(assets, group_name)
         source_assets = [asset.with_group_name(group_name) for asset in source_assets]
 
     return [*assets, *source_assets]
@@ -310,11 +310,11 @@ def prefix_assets(
     return result_assets
 
 
-def group_assets(
+def with_group(
     assets_defs: Sequence[AssetsDefinition], group_name: Optional[str]
-) -> List[AssetsDefinition]:
+) -> Sequence[AssetsDefinition]:
     """
-    Given a list of assets, group them under the group_name.
+    Given a list of assets, groups them under the group_name.
     """
     group_name = check.opt_str_param(group_name, "group_name")
     if not group_name:
--- a/python_modules/dagster/dagster/_core/definitions/load_assets_from_modules.py
+++ b/python_modules/dagster/dagster/_core/definitions/load_assets_from_modules.py
@@ -192,12 +192,7 @@ def load_assets_from_package_module(
     if key_prefix:
         assets = prefix_assets(assets, key_prefix)
     if group_name:
-        assets = [
-            asset.with_prefix_or_group(
-                group_names_by_key={asset_key: group_name for asset_key in asset.keys}
-            )
-            for asset in assets
-        ]
+        assets = group_assets(assets, group_name)
         source_assets = [asset.with_group_name(group_name) for asset in source_assets]
 
     return [*assets, *source_assets]
@@ -313,3 +308,21 @@ def prefix_assets(
             )
         )
     return result_assets
+
+
+def group_assets(
+    assets_defs: Sequence[AssetsDefinition], group_name: Optional[str]
+) -> List[AssetsDefinition]:
+    """
+    Given a list of assets, group them under the group_name.
+    """
+    group_name = check.opt_str_param(group_name, "group_name")
+    if not group_name:
+        return list(assets_defs)
+
+    return [
+        assets_def.with_prefix_or_group(
+            group_names_by_key={asset_key: group_name for asset_key in assets_def.keys}
+        )
+        for assets_def in assets_defs
+    ]
--- a/python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
+++ b/python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
@@ -10,7 +10,7 @@ from dagster import _check as check
 from dagster._annotations import experimental
 from dagster._core.definitions import AssetsDefinition, multi_asset
 from dagster._core.definitions.events import CoercibleToAssetKeyPrefix
-from dagster._core.definitions.load_assets_from_modules import group_assets
+from dagster._core.definitions.load_assets_from_modules import with_group
 from dagster._core.definitions.utils import VALID_NAME_REGEX
 
 
@@ -293,7 +293,7 @@ def load_assets_from_airbyte_project(
         )
 
         if connection_to_group_fn:
-            assets_for_connection = group_assets(
+            assets_for_connection = with_group(
                 assets_for_connection, connection_to_group_fn(connection_name)
             )
         assets.extend(assets_for_connection)
--- a/python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
+++ b/python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
@@ -1,12 +1,17 @@
+import os
 from itertools import chain
-from typing import List, Mapping, Optional, Set
+from typing import Any, Callable, List, Mapping, NamedTuple, Optional, Sequence, Set
 
+import yaml
 from dagster_airbyte.utils import generate_materializations
 
-from dagster import AssetKey, AssetOut, Output
+from dagster import AssetKey, AssetOut, Output, SourceAsset
 from dagster import _check as check
 from dagster._annotations import experimental
 from dagster._core.definitions import AssetsDefinition, multi_asset
+from dagster._core.definitions.events import CoercibleToAssetKeyPrefix
+from dagster._core.definitions.load_assets_from_modules import group_assets
+from dagster._core.definitions.utils import VALID_NAME_REGEX
 
 
 @experimental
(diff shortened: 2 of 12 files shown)


benpankow:
* master
  * **PR #9649** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9649" target="_blank">Graphite</a> 
    * **PR #9655** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9655" target="_blank">Graphite</a> 
      * **PR #9664** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9664" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9664?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/load_assets_from_modules.py
```diff
@@ -313,3 +308,21 @@ def asset2(asset1):
             )
         )
     return result_assets
+
+
+def group_assets(
+    assets_defs: Sequence[AssetsDefinition], group_name: Optional[str]
+) -> List[AssetsDefinition]:
```
sryza, line 315: Any reason not to return a `Sequence`?

python_modules/dagster/dagster/_core/definitions/load_assets_from_modules.py
```diff
@@ -313,3 +308,21 @@ def asset2(asset1):
             )
         )
     return result_assets
+
+
+def group_assets(
```
sryza, line 313: `with_group` might be more consistent with other functions and methods that do similar things

python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
```diff
@@ -88,3 +93,209 @@ def _assets(context):
...
+
+
+@experimental
+def load_assets_from_airbyte_project(
+    project_dir: str,
+    key_prefix: Optional[CoercibleToAssetKeyPrefix] = None,
+    source_key_prefix: Optional[CoercibleToAssetKeyPrefix] = None,
+    create_assets_for_normalization_tables: bool = True,
+    connection_to_group_fn: Optional[Callable[[str], Optional[str]]] = _clean_name,
+) -> Tuple[List[AssetsDefinition], List[SourceAsset]]:
```
sryza, line 229: Returning a `List[Union[AssetsDefinition, SourceAsset]]` would be more consistent with functions like `load_assets_from_package_module`.

python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
```diff
@@ -88,3 +93,209 @@ def _assets(context):
...
+    check.invariant(
+        config_contents is not None, "Airbyte connection config is missing 'configuration' key"
+    )
+
+    return AirbyteConnection(
+        name=contents["resource_name"],
+        source_config_path=contents["source_configuration_path"],
+        stream_prefix=config_contents.get("prefix", ""),
+        has_basic_normalization=any(
+            op.get("operator_configuration", {}).get("operator_type") == "normalization"
```
OwenKephart, line 211: will this catch custom normalization as well? or does that show up as a different operator type?

python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
```diff
@@ -88,3 +93,209 @@ def _assets(context):
...
+    project_dir: str,
+    key_prefix: Optional[CoercibleToAssetKeyPrefix] = None,
+    source_key_prefix: Optional[CoercibleToAssetKeyPrefix] = None,
+    create_assets_for_normalization_tables: bool = True,
+    connection_to_group_fn: Optional[Callable[[str], Optional[str]]] = _clean_name,
+) -> Tuple[List[AssetsDefinition], List[SourceAsset]]:
+    """
+    Loads an Airbyte project into a set of Dagster assets.
+
+    Point to the root folder of an Airbye project synced using the Octavia CLI. For
```
OwenKephart, line 233: nit: Airbye -> Airbyte

sryza (commented):
I left a few pedantic comments. This looks pretty cool. Leaving it up to the Airflow expert Owen to provide more useful feedback.

python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
benpankow: Makes sense - how do you pull it back out into assets and source assets for e.g. `build_assets_job` then?

python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
```diff
@@ -88,3 +93,209 @@ def _assets(context):
...
+    for connection_name in os.listdir(connections_dir):
+        connection_dir = os.path.join(connections_dir, connection_name)
+        with open(os.path.join(connection_dir, "configuration.yaml"), encoding="utf-8") as f:
+            connection = _airbyte_connection_from_config(yaml.safe_load(f.read()))
+
+        with open(os.path.join(project_dir, connection.source_config_path), encoding="utf-8") as f:
+            source = _airbyte_source_from_config(yaml.safe_load(f.read()))
+            if source.name not in source_assets:
+                source_asset_key = AssetKey(source_key_prefix + [_clean_name(source.name)])
+                source_assets[source.name] = SourceAsset(key=source_asset_key)
```
OwenKephart, line 270: I thought about this a bit more, and I'm not sure that we should create actual SourceAsset objects on behalf of the user. For the dbt case, the reason we don't create SourceAssets is that it's common for users to want to supply their own asset definitions for the data that dbt is consuming, so if we created SourceAssets on their behalf, then these definitions would conflict with the definitions that they wanted to define.

It's definitely less likely for Airbyte users to want to supply these definitions (airbyte is often the first step of a data pipeline), but it's unclear if it adds much value to provide them to the user, compared to the cost of making users understand that this function returns two separate lists.

python_modules/dagster/dagster/_core/definitions/load_assets_from_modules.py
benpankow: nope, will tweak

python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
sryza: Users don't call `build_assets_job` - normally they'd either call `materialize` or put them on their repo, both of which don't require separating them out.

python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
```diff
@@ -88,3 +93,209 @@ def _assets(context):
...
+    """
+
+
+def _airbyte_connection_from_config(contents: Mapping[str, Any]) -> AirbyteConnection:
+    config_contents = cast(Mapping[str, Any], contents.get("configuration"))
+    check.invariant(
+        config_contents is not None, "Airbyte connection config is missing 'configuration' key"
+    )
+
+    return AirbyteConnection(
```
OwenKephart, line 206: I think it'd be marginally cleaner to have an `AirbyteConnection.from_config` method, rather than having this helper function. It co-locates the logic for how each of these fields is populated with the object itself.

OwenKephart (changes requested):
overall looking solid, but had some comments

python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
benpankow: I do think there's some visual benefit to have a source asset present, in my opinion it makes it much more clear that the resulting assets are tied together. Without it the user is just presented with a bundle of free-floating assets that don't have as clear an indication that they come from the same origin.

Screen Shot 2022-09-12 at 3 34 04 PM

Maybe we could output a single, mixed list as Sandy suggested above?

python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
benpankow: will add a check to `option` field to make sure it only triggers on basic normalization

python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
sryza: In the case where each table in Snowflake is pulled from a table in the user's app DB, would we have a source asset per table in the user's app DB or a source asset for "app DB"?

benpankow: Updated to output a single list. Let me know if the new API feels ok.

It might make sense to either allow a user to suppress creating source assets or substitute fully-defined assets. I do think it is a nice default to have the source assets created, though.

OwenKephart: hm the presence / absence of the source asset should not change the visual representation here, to the best of my knowledge. I think the only change would be that if you clicked on the link w/o the source asset there'd be no page for it (not even confident that's true)

benpankow: > In the case where each table in Snowflake is pulled from a table in the user's app DB, would we have a source asset per table in the user's app DB or a source asset for "app DB"?

At present, the latter. It's one-source-asset-per-data-source (GitHub, a Postgres DB etc). We could instead have one-source-asset-per-"stream". That makes some more sense in the DB-to-DB sync case since the source tables would be individually identified. In the above case where we're syncing a bunch of data from GitHub, it might be a bit messier. What do you think?

benpankow: > hm the presence / absence of the source asset should not change the visual representation here

Sorry, I think i misunderstood due to a lack of understanding of the SourceAsset system. It sounds like we can have an upstream `AssetKey` specified without actually creating / returning the `SourceAsset` object? Things should still work OK then?

python_modules/dagster/dagster/_core/definitions/load_assets_from_modules.py
```diff
@@ -313,3 +308,21 @@ def asset2(asset1):
...
+
+def with_group(
+    assets_defs: Sequence[AssetsDefinition], group_name: Optional[str]
+) -> Sequence[AssetsDefinition]:
+    """
+    Given a list of assets, groups them under the group_name.
+    """
+    group_name = check.opt_str_param(group_name, "group_name")
+    if not group_name:
+        return list(assets_defs)
```
OwenKephart, line 321: nit: input type is sequence, as is output type, so can probably just return assets_defs here, no?

python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
```diff
@@ -88,3 +92,208 @@ def _assets(context):
...
+    for connection_name in os.listdir(connections_dir):
+        connection_dir = os.path.join(connections_dir, connection_name)
+        with open(os.path.join(connection_dir, "configuration.yaml"), encoding="utf-8") as f:
+            connection = AirbyteConnection.from_config(yaml.safe_load(f.read()))
+
+        with open(os.path.join(project_dir, connection.source_config_path), encoding="utf-8") as f:
+            source = AirbyteSource.from_config(yaml.safe_load(f.read()))
+            source_asset_key = AssetKey(source_key_prefix + [_clean_name(source.name)])
+
+        state_file = next(
```
OwenKephart, line 270: will there ever be multiple state files in the directory under normal circumstances? or does the cli always delete previous versions when writing a new one?

OwenKephart (approved):
looks great! excited about this feature -- had a couple remaining comments but this looks ready to merge for the most part

python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
sryza: > At present, the latter. It's one-source-asset-per-data-source (GitHub, a Postgres DB etc). We could instead have one-source-asset-per-"stream". That makes some more sense in the DB-to-DB sync case since the source tables would be individually identified. In the above case where we're syncing a bunch of data from GitHub, it might be a bit messier. What do you think?

I think I'd expect a source asset per stream.  If the main reason for the source assets is to group things visually, I think we should find a different way to do that.  My lean would be to punt on the source assets entirely.  Might be easiest to chat about this on zoom?

sryza (changes requested):
requesting changes based on my comment about source assets

python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
benpankow: I think there could be multiple if the user has multiple workspaces - I'll add an optional workspace ID argument which will disambiguate & error if a user doesn't provide it & there's more than one state file.

benpankow: did the above

python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
benpankow: Following offline discussion, removed upstream assets for now, with the option to add them down the line.

sryza (approved):
bombs away!

benpankow:
**Graphite Merge Job**

Current status: ⏳ Queued to merge

This pull request is currently queued to be merged as part of a stack.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9664?utm_source=stack-comment).


Job Reference: ykk4kBl6Nl5kRQqvJjms

PR #9666, opened 2022-09-12
run status sensor that can monitor the whole instance

### LINKED ISSUE #9499: allow run failure sensors to monitor all jobs within the instance
Feedback:
- i think generally i've found myself wanting broader scopes for sensors, especially with these "system-level sensors". long-term vision here is that we can grant autonomy to code owners while still having global visibility into the platform

jamiedemaria:
### Summary & Motivation
adda param `monitor_all_repositories` to run status sensors that will monitor all jobs in an instance. setting this param to True is incompatible with also passing jobs via `monitored_jobs` or `job_selection` and will raise an error

resolves #9499

daemon test


prha:
The `InstanceSelector` feels a bit off to me, since it's not something we can actually select.  

We'd be referencing the idea of an instance into the repository definition itself, which makes me a bit uncomfortable.  Are we feeling like the default behavior for `monitored_jobs` (e.g. `None`) should not be repository-scoped?

jamiedemaria:
introducing the `InstanceSelector` was one idea i had so that it could fit the pattern of the other cross repo selectors, but i agree that it's strange. other ideas i had could be to have a new run status sensor decorator for monitoring the instance (`@instance_run_status_sensor` or something), introducing a new param to `run_status_sensor` (something like `monitor_instance: bool = False`).

sryza:
How about some sort of ALL_RUNS_ON_INSTANCE constant that you can supply to the `monitored_jobs` argument?  I'm with @prha that it feels weird to "select" the instance, because there aren't other instances you can select, but I think it's fine for repository-level constructs to be aware of the idea of the instance - e.g. we have an `instance` param on most context objects.

I think the interaction of a `monitor_instance` bool param and supplying a `RepositorySelector` to `monitored_jobs` would be confusing.

prha:
prefer a new boolean flag param than having to export some new constant from the top-level, but don't have strong feelings about it.

We could mitigate the param interactions by raising some sort of error if you supply both.

Long term, we could also start generating warnings when you don't pass an explicit repository selector to monitored jobs, and eventually flip the behavior so that the default is instance-scoped?

sryza:
> prefer a new boolean flag param than having to export some new constant from the top-level, but don't have strong feelings about it.

I don't feel super strongly on this. If both of you prefer the bool param, you can override me.

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -235,10 +237,12 @@ def run_failure_sensor(
         description (Optional[str]): A human-readable description of the sensor.
         monitored_jobs (Optional[List[Union[JobDefinition, GraphDefinition, UnresolvedAssetJobDefinition, RepositorySelector, JobSelector]]]): The jobs
             in the current repository that will be monitored by this failure sensor. Defaults to None, which means the alert will
-            be sent when any job in the current repository fails. Jobs in external repositories can be specified using RepositorySelector and JobSelector.
+            be sent when any job in the current repository fails.
         job_selection (Optional[List[Union[JobDefinition, GraphDefinition, RepositorySelector, JobSelector]]]): (deprecated in favor of monitored_jobs)
             The jobs in the current repository that will be monitored by this failure sensor. Defaults to None, which means
             the alert will be sent when any job in the repository fails.
+        monitor_instance (bool): If set to True, the sensor will monitor all jobs in the Dagster instance. If used, you cannot also specify monitored_jobs
```
sryza, line 244: A few nitpicks:
- Worth pointing out that this is mutually exclusive with `monitored_jobs`.
- It would make sense to put this right beneath `monitored_jobs`.
- Better to say "runs" than "jobs", because it will also monitor asset materializations that aren't part of jobs.
- Exceeds the 100-character line length limit.
- What do you think about calling this `monitor_all_repositories` to avoid the "instance" vs. "deployment" confusion.

sryza (approved):
Nice

prha (approved):

PR #9678, opened 2022-09-13
Add option to run EcsRunLauncher without pulling network config from the current ECS task, and kwargs to customize the task config

gibsondan:
Right now the only way to use the EcsRunLauncher involves pulling permissions and other configuration from the task that is launching the run. This creates a problem in situations where you might want system code to have different IAM roles than user code, or even launch runs from outside of ECS. In Cloud, it creates an awkward situation where the grpc server tasks  are configured based on fields on the instance, but runs are configured by pulling from the current task.

This PR creates a configuration option that lets you pass in all the configuration you need when launching run, with nothing pulled from the current container. The old version is kept as well.

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
@@ -1,13 +1,12 @@
 import json
 import warnings
 from collections import namedtuple
-from contextlib import suppress
 from typing import Any, Dict, Optional
 
 import boto3
 from botocore.exceptions import ClientError
 
-from dagster import Array, Field, Noneable, ScalarUnion, StringSource
+from dagster import Array, Field, Noneable, Permissive, ScalarUnion, StringSource
 from dagster import _check as check
 from dagster._core.events import EngineEventData, MetadataEntry
 from dagster._core.launcher.base import (
@@ -22,7 +21,13 @@ from dagster._serdes import ConfigurableClass
 
 from ..secretsmanager import get_secrets_from_arns
 from .container_context import SHARED_ECS_SCHEMA, EcsContainerContext
-from .tasks import default_ecs_task_definition, default_ecs_task_metadata
+from .tasks import (
+    DagsterEcsTaskDefinitionConfig,
+    get_current_ecs_task,
+    get_current_ecs_task_metadata,
+    get_task_definition_dict_from_current_task,
+    get_task_kwargs_from_current_task,
+)
 from .utils import sanitize_family
 
 Tags = namedtuple("Tags", ["arn", "cluster", "cpu", "memory"])
@@ -51,6 +56,8 @@ class EcsRunLauncher(RunLauncher, ConfigurableClass):
         secrets_tag="dagster",
         env_vars=None,
         include_sidecars=False,
+        use_current_ecs_task_config: bool = True,
+        run_task_kwargs: Optional[Dict[str, Any]] = None,
     ):
         self._inst_data = inst_data
         self.ecs = boto3.client("ecs")
@@ -94,6 +101,33 @@ class EcsRunLauncher(RunLauncher, ConfigurableClass):
                 f"'{self.task_definition}' because the container is not defined.",
             )
             self.task_definition = task_definition["taskDefinition"]["taskDefinitionArn"]
+
+        self.use_current_ecs_task_config = check.opt_bool_param(
+            use_current_ecs_task_config, "use_current_ecs_task_config"
+        )
+
+        self.run_task_kwargs = check.opt_dict_param(run_task_kwargs, "run_task_kwargs")
+        if run_task_kwargs:
+            check.invariant(
+                "taskDefinition" not in run_task_kwargs,
+                "Use the `taskDefinition` config field to pass in a task definition to run.",
+            )
+            check.invariant(
+                "overrides" not in run_task_kwargs,
+                "Task overrides are set by the run launcher and cannot be set in run_task_kwargs.",
+            )
+
+            expected_keys = [
+                key for key in self.ecs.meta.service_model.shape_for("RunTaskRequest").members
+            ]
+
+            for key in run_task_kwargs:
+                check.invariant(
+                    key in expected_keys, f"Found an unexpected key {key} in run_task_kwargs"
+                )
+
+        self._current_task_metadata = None
+        self._current_task = None
 
     @property
     def inst_data(self):
@@ -151,6 +185,33 @@ class EcsRunLauncher(RunLauncher, ConfigurableClass):
                     "Defaults to False."
                 ),
             ),
+            "use_current_ecs_task_config": Field(
+                bool,
+                is_required=False,
+                default_value=True,
+                description=(
+                    "Whether to use the run launcher's current ECS task in order to determine "
+                    "the cluster and networking configuration for the launched task. Defaults to "
+                    "True. Should only be called if the run launcher is running within an ECS "
+                    "task."
+                ),
+            ),
+            "run_task_kwargs": Field(
+                Permissive(
+                    {
+                        "cluster": Field(
+                            StringSource,
+                            is_required=False,
+                            description="Name of the ECS cluster to launch ECS tasks in.",
+                        ),
+                    }
+                ),
+                is_required=False,
+                description="Additional arguments to include while running the task. See "
+                "https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs.html#ECS.Client.run_task "
+                "for the available parameters. The overrides and taskDefinition arguments will always "
+                "be set by the run launcher.",
+            ),
             **SHARED_ECS_SCHEMA,
         }
 
@@ -165,9 +226,11 @@ class EcsRunLauncher(RunLauncher, ConfigurableClass):
         except ClientError:
             pass
 
-    def _set_run_tags(self, run_id, task_arn):
-        cluster = self._task_metadata().cluster
-        tags = {"ecs/task_arn": task_arn, "ecs/cluster": cluster}
+    def _set_run_tags(self, run_id: str, cluster: str, task_arn: str):
+        tags = {
+            "ecs/task_arn": task_arn,
+            "ecs/cluster": cluster,
+        }
         self._instance.add_run_tags(run_id, tags)
 
     def _get_run_tags(self, run_id):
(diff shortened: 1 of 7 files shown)


gibsondan:
* master
  * **PR #9678** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9678" target="_blank">Graphite</a>  👈
    * **PR #9762** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9762" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9678?utm_source=stack-comment).

python_modules/libraries/dagster-aws/dagster_aws_tests/ecs_tests/launcher_tests/test_launching.py
```diff
@@ -100,6 +100,93 @@ def test_default_launcher(
...
+    # We get the family name from the location name. With the InProcessExecutor that we use in tests,
+    # the location name is always <<in_process>>. And we sanitize it so it's compatible with the ECS API.
+    assert task_definition["family"] == "in_process"
+    assert len(task_definition["containerDefinitions"]) == 1
+    container_definition = task_definition["containerDefinitions"][0]
+    assert container_definition["name"] == "run"
+    assert container_definition["image"] == image
+    assert not container_definition.get("entryPoint")
+    assert not container_definition.get("dependsOn")
+    # But other stuff is inherited from the parent task definition
```
jmsanders, line 148: I know this is largely copy-pasted from other tests, but isn't this the exact opposite of what we're trying to support here?

Can we maybe reduce the test assertions to just the behavior that diverges from the default?

python_modules/libraries/dagster-aws/dagster_aws_tests/ecs_tests/launcher_tests/test_launching.py
```diff
@@ -148,90 +235,65 @@ def test_task_definition_registration(
     assert len(ecs.list_task_definitions()["taskDefinitionArns"]) == len(task_definitions) + 1
 
 
-def test_reuse_task_definition(instance):
-    image = "image"
-    secrets = []
+def test_reuse_task_definition(instance, ecs):
```
jmsanders, line 238: It's a little unclear to me if this change is meant to:

- force registration of a new task definition revision for the same family
- use a task definition without any pre-populated defaults from the parent task

Which probably points to both some poor decisions with how I initially wrote this and also some naming ambiguities that could be cleaned up in this PR?

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -333,75 +422,141 @@ def terminate(self, run_id):
...
-            if secrets
-            else {}
+        family = sanitize_family(
+            run.external_pipeline_origin.external_repository_origin.repository_location_origin.location_name  # type: ignore
         )
 
-        task_definition = {}
-        with suppress(ClientError):
-            task_definition = self.ecs.describe_task_definition(taskDefinition=family)[
+        if self.use_current_task_definition:
```
jmsanders, line 453: What if we instead let people set an optional "TaskDefinition" dict in config - and we merged that dict with our existing defaults? That might help squash "impossible" states where you set `use_current_task_definition` to `False` but also don't provide everything needed to define a new task definition.

I'm also just having difficulty following what even happens differently if you set this to `True` - what do we still pull from the parent's task definition and not from the launcher config that couldn't be represented as a "default" value?

jmsanders (dismissed):

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
gibsondan: i'm excited to offer something like a task_definition_kwargs field or something that lets you specify arbitrary task definition config so that we don't have to re-implement the whole spec as Dagster config, but I don't think it would be sufficient here. The pieces config that we're offering are a mixture of things on the dagster container (log configuration / sidecars), the task definition (execution_role_arn, task_role_arn), and the task (cluster / security groups / subnets).

Generally I'm trying to follow the general playbook here (https://elementl.quip.com/9K0bAWhc6t9n/Agent-Configuration-Options) - where we have a relatively simple and blessed way to set the most common pieces of config we think people are likely to want(secrets etc.) and then eventually also an escape hatch where we expose the full raw config for the power users. I think there's value in not needing to send people to the boto spec if they want to do something like 'set env vars' or 'point at an existing cluster' or 'change the IAM role that's used'. But which things are considered 'the most common pieces of config' is very subjective.

python_modules/libraries/dagster-aws/dagster_aws/ecs/tasks.py
```diff
@@ -33,27 +135,49 @@ class EcsNoTasksFound(Exception):
...
     # (status, revision, etc.)
     expected_keys = [
         key for key in ecs.meta.service_model.shape_for("RegisterTaskDefinitionRequest").members
     ]
     task_definition = dict(
-        (key, metadata.task_definition[key])
+        (key, current_task_definition_dict[key])
         for key in expected_keys
-        if key in metadata.task_definition.keys()
+        if key in current_task_definition_dict.keys()
```
gibsondan, line 180: @jmsanders my understanding is that the answer to your question of what else can end up in the task definition if you set use_current_task_definition=True can be found here - it's the set of things that are a) params going into register_task_definition and b) found on the output of describe_task_definition

jmsanders: But then we merge a bunch of stuff onto it based on what's in your launcher config.

I guess my question is how much is even left that we actually inherit vs. override.

gibsondan: There are many random fields here on the taskDefinition arg to  that would no longer be copied over (placementConstraints etc.): https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs.html#ECS.Client.register_task_definition

Whether they are actually in use by anybody, I do not know

python_modules/libraries/dagster-aws/dagster_aws/ecs/tasks.py
```diff
@@ -1,21 +1,123 @@
...
+            command=container_definition.get("command"),
+            log_configuration=task_definition_dict.get("logConfiguration"),
+            secrets=container_definition.get("secrets"),
+            environment=container_definition.get("environment"),
+            execution_role_arn=container_definition.get("executionRoleArn"),
+            task_role_arn=container_definition.get("taskRoleArn"),
+            sidecars=sidecars,
+        )
+
+    def task_definition_dict(self):
```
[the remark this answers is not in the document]
jmsanders: I think the original motivations for the way we did things still hold true. Namely that we can't override the image and we want to enable the simple case where a user doesn't need to set up their own infrastructure (and can instead just use docker compose to launch):

https://dagster.phacility.com/D8404
https://dagster.phacility.com/D8486

But the more I've talked with Daniel, the more I think we were overzealous in how much information we copy forward from the original task definition.

gibsondan:
@jmsanders and @Ramshackle-Jamathon latest rev responds to your feedback about task definitions by changing the plan - instead of having a path where the run launcher constructs a new task def from scratch, the grpc server would specify the task definition arn to use via container_context (likely using its own task definition). That avoids a situation where we have two totally separate task defs to use for user code.

Still need to do a pass on the tests to reflect that new version, but curious for overall thoughts on that plan

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -151,6 +163,23 @@ def config_type(cls):
...
+                            is_required=False,
+                            description="Name of the ECS cluster to launch ECS tasks in.",
+                        ),
+                    }
+                ),
+                is_required=False,
+                description="Additional arguments to include while running the task. See "
+                "https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs.html#ECS.Client.run_task "
+                "for the available parameters. The overrides and taskDefinition arguments will always "
+                "be set by the tun launcher. If this field is not set, the arguments to run_task "
```
jmsanders, line 180:
```suggestion
                "be set by the run launcher. If this field is not set, the arguments to run_task "
```

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -333,75 +352,107 @@ def terminate(self, run_id):
...
-            else {}
-        )
+            if not self._reuse_task_definition(
+                task_definition_config,
+            ):
+                self.ecs.register_task_definition(**task_definition_dict)
+
+            task_definition = family
+
+        if self.run_task_kwargs != None:
```
jmsanders, line 406: 👨‍🍳 💋 

Must nicer if/else logic.

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
jmsanders: Also, let's make sure to add a test for this statement:

> The overrides and taskDefinition arguments will always be set by the tun launcher.

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -249,17 +274,10 @@ def launch_run(self, context: LaunchRunContext) -> None:
...
-                "awsvpcConfiguration": {
-                    "subnets": metadata.subnets,
-                    "assignPublicIp": metadata.assign_public_ip,
-                    "securityGroups": metadata.security_groups,
-                }
-            },
+            taskDefinition=task_config.task_definition,
             launchType="FARGATE",
+            overrides=overrides,
+            **task_config.run_task_kwargs,
```
jmsanders, line 280: Don't we need to munge `task_config.run_task_kwargs` to not include `taskDefinition`, `launchType`, and `overrides`? Otherwise, we can end up in a situation where the same kwarg is passed to the function twice.

```py
>>> def foo(**kwargs):
...   pass
... 
>>> kwargs = {"bar": 1}
>>> foo(**kwargs)
>>> foo(bar=2, **kwargs)
TypeError: foo() got multiple values for keyword argument 'bar'
```

jmsanders (changes requested):
Love this new direction.

I think we need to sanitize `run_task_kwargs` a bit though, don't we?

gibsondan:
Well this has been a journey! @jmsanders i think this is ready for perusal again

gibsondan:
Joe can the image be overridden? If not I think that’s the big blocker here
- although maybe the gRPC servers could expose a task definition to use
rather than an image to use

I’m not sure I want to block this change on that though since it’s fairly
substantial and there are some real wins here (we’re not doing any more
task definition registration after this PR than we were before)

PR #9679, opened 2022-09-13
Add asset selection flag to run_request_for_partition

clairelin135:
We already allow an `asset_selection` on a run request. This PR enables a partitioned run request to also contain an asset selection. Adding this in preparation to add docs examples for triggering partitioned runs via `multi_asset_sensor`.


sryza (approved):
Nice

PR #9680, opened 2022-09-13
[dagster-dbt] fix missing select when using dbt_cli_resource

OwenKephart:
### Summary & Motivation

I recently added select as named keyword argument to this method, which messed up this code path. this fixes it.


alangenfeld (approved):
add a test to prevent similar mistakes?

Issue #8959, opened 2022-07-19
Dagit UI is completely broken when accessing certain asset metadata 


danielgafni:
### Dagster version

0.15.6

### What's the issue?

When clicking on a certain asset, the Dagit UI turns completely blank. This error appears in the js console:

```
react-dom.production.min.js:216 SyntaxError: Unexpected token N in JSON at position 317
    at JSON.parse (<anonymous>)
    at Object.content (MetadataEntry.tsx:106:36)
...
```

### Deployment type

Dagster Helm chart

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

danielgafni:
The error disappeared after this asset was rematerialized (I used an old run to do this). Seems like something was broken for the latest partition metadata.

hellendag:
@danielgafni Is it possible that you have an `NaN` value in the metadata object being parsed here?

Do you have some more details on reproducing the issue? Is this occurring on the Run view only, or elsewhere for this asset as well?

danielgafni:
Hey! I can't reproduce it since it happened a long time ago.

Nan values are possible.

It happened both in the Run view and in the Asset Catalog when selecting the corrupted asset.

PR #9682, opened 2022-09-14
[dagit] Defensively pretty-print JSON in run logs

bengotow:
### Summary & Motivation

This resolves the issue reported here: https://github.com/dagster-io/dagster/issues/8959

I feel slightly gross exporting the existing helper for this from `@dagster/ui`, but I felt equally gross redefining it in the main app. Curious if maybe that'd be better? Two definitions might not kill us but I wanted to include the comment explaining why it was try-catched.


hellendag (approved):
I think it's okay. It's a useful general UI utility, so it's not so bad for it to be exported from `ui`.

PR #9685, opened 2022-09-14
Make confirmation provider more customizable

salazarm:
### Summary & Motivation

Allow changing the icon and the intent on the dialog

Using these props in cloud

### DIFF
--- a/js_modules/dagit/packages/core/src/app/CustomConfirmationProvider.tsx
+++ b/js_modules/dagit/packages/core/src/app/CustomConfirmationProvider.tsx
@@ -5,27 +5,33 @@ interface ConfirmationOptions {
   catchOnCancel?: boolean;
   title?: string;
   description?: React.ReactNode;
+  icon?: ConfirmationDialogProps['icon'];
+  intent?: ConfirmationDialogProps['intent'];
 }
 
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/core/src/app/CustomConfirmationProvider.tsx
```diff
@@ -5,27 +5,33 @@ interface ConfirmationOptions {
...
+  icon,
   title,
+  intent = 'danger',
   description,
   onSubmit,
   onClose,
 }) => {
   return (
-    <Dialog icon={title ? 'info' : undefined} onClose={onClose} title={title} isOpen={open}>
+    <Dialog icon={title ? icon ?? 'info' : icon} onClose={onClose} title={title} isOpen={open}>
```
hellendag, line 30: Hm, does `icon` need to care about the `title` value?

hellendag (approved):

js_modules/dagit/packages/core/src/app/CustomConfirmationProvider.tsx
salazarm: I don't know but I figured I should preserve the existing behavior in case we're relying on it somewhjere

PR #9692, opened 2022-09-14
graph_name parameter on define_asset_job

sryza:
### Summary & Motivation

I'm hoping to get this in for this week's release, to help out a user.

Without this change, this breaks:
```
from dagster import op, graph, repository, AssetsDefinition, define_asset_job


@op
def op1():
    ...


@graph
def graph1():
    ...


@repository
def repo():
    asset1 = AssetsDefinition.from_graph(graph1)
    return [asset1, define_asset_job("graph1")]
```

with 

```
dagster._core.errors.DagsterInvalidDefinitionError: Conflicting definitions found in repository with name 'graph1'. Op/Graph/Solid definition names must be unique within a repository. GraphDefinition is defined in job '__ASSET_JOB' and in job 'graph1'.
```

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_layer.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_layer.py
@@ -790,6 +790,7 @@ def build_asset_selection_job(
     tags: Optional[Mapping[str, Any]] = None,
     asset_selection: Optional[FrozenSet[AssetKey]] = None,
     asset_selection_data: Optional[AssetSelectionData] = None,
+    graph_name: Optional[str] = None,
 ) -> "JobDefinition":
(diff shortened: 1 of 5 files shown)


OwenKephart:
hm might be missing some context but is the solution of changing the job name to `run_graph1` or the graph name to `do_graph1` less desirable than `define_asset_job("graph1", graph_name="graph1_foo")`?

sryza:
> hm might be missing some context but is the solution of changing the job name to run_graph1 or the graph name to do_graph1 less desirable than define_asset_job("graph1", graph_name="graph1_foo")?

I think that would work in a lot of cases, but some people need to preserve the names of their jobs and graphs so they don't lose run history.

OwenKephart (approved):

PR #9696, opened 2022-09-14
[dagster-databricks] fix library install version

OwenKephart:
### Summary & Motivation

Fix library install version -- the dagster library version is no longer guaranteed to match the dagster-pyspark / dagster-databricks library versions.

### DIFF
--- a/python_modules/libraries/dagster-databricks/dagster_databricks/databricks.py
+++ b/python_modules/libraries/dagster-databricks/dagster_databricks/databricks.py
@@ -175,15 +175,23 @@ class DatabricksJobRunner:
         libraries = list(run_config.get("libraries", []))
         install_default_libraries = run_config.get("install_default_libraries", True)
         if install_default_libraries:
+            import dagster_databricks
+            import dagster_pyspark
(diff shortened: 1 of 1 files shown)


python_modules/libraries/dagster-databricks/dagster_databricks/databricks.py
```diff
@@ -175,15 +175,23 @@ def submit_run(self, run_config, task):
         libraries = list(run_config.get("libraries", []))
         install_default_libraries = run_config.get("install_default_libraries", True)
         if install_default_libraries:
+            import dagster_databricks
+            import dagster_pyspark
```
alangenfeld, line 179: do these have to be deferred? these sure to be present?

alangenfeld (approved):

python_modules/libraries/dagster-databricks/dagster_databricks/databricks.py
OwenKephart: fair enough yeah these can be imported at the top

PR #9702, opened 2022-09-14
[dagster-airbyte] Add ability to generate Airbyte assets by pulling config from Airbyte instance

benpankow:
## Summary

Adds a new `@experimental` API similar to #9664 that allows users to load Airbyte assets from an Airbyte instance, connecting at initialization-time and pulling connection definitions.

```python
airbyte_instance = airbyte_resource.configured(
    {
        "host": "localhost",
        "port": "8000",
        "forward_logs": False,
    }
)
airbyte_assets = load_assets_from_airbyte_instance(
    airbyte_instance,
)
```

Has the same config options as the 'from project' API.

## Test Plan

Tested locally; new unit tests

### DIFF
--- a/python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
+++ b/python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
@@ -314,9 +314,11 @@ def load_assets_from_airbyte_instance(
             )
         assets.extend(assets_for_connection)
 
-    return with_resources(
-        assets,
-        resource_defs={"airbyte": airbyte},
+    return list(
+        with_resources(
+            assets,
+            resource_defs={"airbyte": airbyte},
+        )
     )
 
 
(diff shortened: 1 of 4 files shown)


benpankow:
* master
  * **PR #9702** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9702" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9702?utm_source=stack-comment).

python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
```diff
@@ -207,12 +228,99 @@ def parse_stream_tables(
...
+            asset_key_prefix=key_prefix,
+        )
+
+        if connection_to_group_fn:
+            assets_for_connection = list(
+                with_group(assets_for_connection, connection_to_group_fn(connection.name))
+            )
+        assets.extend(assets_for_connection)
+
+    return assets
```
benpankow, line 316: I wonder, since the user is already specifying a resource at initialization time, should it be bound as a convenience? Then all a user has to do is

`airbyte_assets = load_assets_from_airbyte_instance(...)`

instead of

`airbyte_assets = with_resources(load_assets_from_airbyte_instance(...), ...)`

If a user wants to override the resource config for run-time, they could always just use `with_resource` to set the override.

sryza: that sounds right to me

PR #9707, opened 2022-09-15
Remove get_pod_names_in_job call from k8s_job_op

gibsondan:
Summary:
It's unclear why we are pulling the pod names from the job when we already know what the pod name is suppose to be - a user also reported a race condition when calling this method before the pods are actually ready.

Test Plan: Existing integration tests that covers k8s_job_op


alangenfeld (approved):
must be an artifact of code evolution? 

if the integration tests pass this looks good to me

gibsondan:
hmmmm, not that simple unfortunately I don't think...

gibsondan:
https://github.com/dagster-io/dagster/pull/9709 is a more plausible fix

PR #9709, opened 2022-09-15
Retry while waiting for the k8s_job_op job to have pods

gibsondan:
Summary:
User reports that the job being created does not always immediately have pods. Use a similar retry loop that we do throughout the rest of the op.

### DIFF
--- a/python_modules/libraries/dagster-k8s/dagster_k8s/ops/k8s_job_op.py
+++ b/python_modules/libraries/dagster-k8s/dagster_k8s/ops/k8s_job_op.py
@@ -15,8 +15,8 @@ from ..job import (
 )
 from ..launcher import K8sRunLauncher
 from ..utils import (
-    get_pod_names_in_job,
     wait_for_job,
+    wait_for_job_to_have_pods,
     wait_for_pod,
     wait_for_running_job_to_succeed,
 )
@@ -211,10 +211,18 @@ def k8s_job_op(context):
     wait_for_job(
         job_name=job_name,
         namespace=namespace,
-        wait_timeout=config.get("timeout", 0),
+        wait_timeout=timeout,
+        start_time=start_time,
     )
 
-    pod_names = get_pod_names_in_job(job_name, namespace=namespace)
+    pods = wait_for_job_to_have_pods(
+        job_name,
+        namespace,
+        wait_timeout=timeout,
+        start_time=start_time,
+    )
+
+    pod_names = [p.metadata.name for p in pods]
 
     if not pod_names:
         raise Exception("No pod names in job after it started")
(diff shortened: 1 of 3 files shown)


python_modules/dagster-test/Dockerfile
```diff
@@ -33,8 +33,10 @@ RUN pip install \
     -e modules/dagster-airflow \
     -e modules/dagstermill \
     -e . \
+    pyzmq==23.2.1 \
```
alangenfeld, line 36: cleanup - wheel is uploaded now

python_modules/libraries/dagster-k8s/dagster_k8s/utils.py
```diff
@@ -74,7 +92,7 @@ def wait_for_running_job_to_succeed(
     wait_timeout=DEFAULT_WAIT_TIMEOUT,
     wait_time_between_attempts=DEFAULT_WAIT_BETWEEN_ATTEMPTS,
     num_pods_to_wait_for=DEFAULT_JOB_POD_COUNT,
-    start_time=None,
+    start=None,
```
alangenfeld, line 95: why this rename?

python_modules/libraries/dagster-k8s/dagster_k8s/ops/k8s_job_op.py
```diff
@@ -211,18 +211,26 @@ def k8s_job_op(context):
...
+        wait_timeout=timeout,
+        start_time=start_time,
     )
 
-    pod_names = get_pod_names_in_job(job_name, namespace=namespace)
+    pods = wait_for_job_to_have_pods(
+        job_name,
+        namespace,
+        wait_timeout=timeout,
+        start_time=start_time,
```
alangenfeld, line 222: whats up with `start_time` here? we want to apply timeouts relative to the start of the overall process not each step? Felt confused reading the code figuring this out

alangenfeld (approved):

johannkm (approved):
Do we support parallel jobs? https://kubernetes.io/docs/concepts/workloads/controllers/job/#parallel-jobs

I think so right? We just only get the logs from one of them

python_modules/libraries/dagster-k8s/dagster_k8s/ops/k8s_job_op.py
gibsondan: That's what I would expect as a user, yeah - overall timeout for the op, not a timeout for whatever random individual steps make up the op

gibsondan:
@johannkm yeah, wouldn't be a huge lift to stream logs from each pod instead of just the first pod

PR #9710, opened 2022-09-15
[dagster-graphql] dont batch if only 1 target

alangenfeld:
noticed some traces where we were batching for 1 item which adds overhead we can bypass.

we could consider raising the batching threshold to a number higher than 1 but im not sure how we would pick that.

bk


alangenfeld:
* master
  * **PR #9710** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9710" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9710?utm_source=stack-comment).

prha (approved):

PR #9711, opened 2022-09-15
asset expectations

sryza:

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/schema/logs/events.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/logs/events.py
@@ -1,3 +1,5 @@
+from typing import Sequence
+
 import graphene
 
 import dagster._check as check
@@ -331,6 +333,9 @@ class AssetEventMixin:
         from ...implementation.events import _to_metadata_entries
 
         return _to_metadata_entries(self._metadata.metadata_entries)
+
+    def resolve_expectationResults(self, _graphene_info) -> Sequence[GrapheneExpectationResult]:
+        return "what goes here???"
 
     def resolve_partition(self, _graphene_info):
         return self._metadata.partition
--- a/python_modules/dagster/dagster/__init__.py
+++ b/python_modules/dagster/dagster/__init__.py
@@ -119,6 +119,9 @@ from dagster._core.definitions.composition import (
 from dagster._core.definitions.configurable import (
     configured as configured,
 )
+from dagster._core.definitions.decorators.asset_expectation_decorator import (
+    asset_expectation as asset_expectation,
+)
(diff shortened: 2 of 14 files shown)


python_modules/dagster-graphql/dagster_graphql/schema/logs/events.py
```diff
@@ -332,6 +334,9 @@ def resolve_metadataEntries(self, _graphene_info):
 
         return _to_metadata_entries(self._metadata.metadata_entries)
 
+    def resolve_expectationResults(self, _graphene_info) -> Sequence[GrapheneExpectationResult]:
+        return "what goes here???"
```
sryza, line 338: @clairelin135 are you able to help a graphql noob (me) with this? How do I turn `ExpectationResult` objects into `GrapheneExpectationResults`?

clairelin135, line 338: expectation results are only a field on `AssetObservations` and not `AssetMaterializations` right? If that's the case, you'll need to change `GrapheneAssetObservation` instead:

```
class GrapheneObservationEvent(graphene.ObjectType, AssetEventMixin):
    class Meta:
        interfaces = (GrapheneMessageEvent, GrapheneStepEvent, GrapheneDisplayableEvent)
        name = "ObservationEvent"

    expectationResult = non_null_list(GrapheneExpectationResult)

    def __init__(self, event):
        self._observation = event.dagster_event.asset_observation_data.asset_observation
        super().__init__(**_construct_asset_event_metadata_params(event, self._observation))
        AssetEventMixin.__init__(
            self,
            event=event,
            metadata=self._observation,
        )

    def resolve_expectationResult(self, _graphene_info):
        return self._observation.expectation_results
 ```
 
 I think doing this should work!

sryza: I think what's confusing me is that I expect the type of `self._observation.expectation_results` to be `ExpectationResult`, but I expect this to return a `GrapheneExpectationResult`.  What am I missing?

clairelin135: This part I don't completely understand either, but graphene has a ["default resolver"](https://docs.graphene-python.org/en/latest/types/objecttypes/#defaultresolver) functionality that will fetch the attribute from the provided object given the field name. So I think in this case, `GrapheneExpectationResult` has fields like 'success', 'label', etc. that are basically mapped to `ExpectationResult.success` and `ExpectationResult.label`

sryza: got it - that clears it up. thank you!

PR #9713, opened 2022-09-15
[WIP] [asset-slas] AssetSLAs

OwenKephart:
### Summary & Motivation

Add AssetSLAs + a way to make sensors to monitor if they're passing or not


examples/development_to_production/development_to_production/__init__.py
```diff
@@ -1 +1,6 @@
-from development_to_production.repository import repo
+import sys
```
OwenKephart, line 1: lolol oops

python_modules/dagster/dagster_tests/core_tests/definitions_tests/test_sensor_invocation.py
```diff
@@ -662,3 +669,192 @@ def asset_sensor(context):
             [july_daily_partitions.key], instance=instance, repository_def=my_repo
         )
         list(asset_sensor(ctx))
+
+
+@pytest.mark.parametrize(
```
OwenKephart, line 674: I kinda just put a bunch of random test cases I thought of together, this should be more structured

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -636,6 +637,201 @@ def assets_defs_by_key(self) -> Mapping[AssetKey, AssetsDefinition]:
...
+            assets_def = self._repository_def._assets_defs_by_key.get(asset_key)
+            if assets_def is None:
+                raise DagsterInvalidDefinitionError(
+                    f"No asset with {asset_key} found in repository"
+                )
+            if not assets_def.slas_by_key.get(asset_key):
+                continue
+            self._assets_by_key[asset_key] = assets_def
+
+        self._next_cursor = cursor
```
OwenKephart, line 676: I think this should potentially work a bit more like the MultiAssetSensor stuff, where the user has to manually update the cursor, using a dictionary of whatever statuses they used?

In my mind, the main use case for this will be to only alert when a status changes from the last tick (most often when it goes from passing to failing, but sometimes you want the other direction as well). This is because it would be extremely annoying to get (e.g.) a slack message every 30 seconds whenever an SLA was failing.

This use case necessitates a cursor to keep track of the previous status

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -636,6 +637,201 @@ def assets_defs_by_key(self) -> Mapping[AssetKey, AssetsDefinition]:
...
+        For each asset monitored by this sensor, calculate its "data time". The data time of an
+        asset is the time of the earliest data that could be incorporated into the current state
+        of the asset, calculated for each asset by comparing their latest timestamp to the latest
+        parent timestamp that happened before that materialization (as you cannot incorporate parent
+        data if that parent runs after you do).
+        """
+        from dagster._core.events import DagsterEventType
+        from dagster._core.storage.event_log.base import EventRecordsFilter
+
+        calculated_data_times: Dict[Tuple[AssetKey, Optional[float]], Optional[float]] = {}
```
OwenKephart, line 698: we're using a little bit of dynamic programming here -- this isn't just for show, as in a lot of normal scenarios there's a lot of overlap between query paths that need to be taken up the tree. might do some benchmarking on this to see if it helps in practice

python_modules/dagster/dagster/_core/definitions/asset_sla.py
```diff
@@ -0,0 +1,46 @@
...
+from croniter import croniter
+
+
+class AssetSLA:
+    def is_passing(self, data_timestamp: float):
+        pass
+
+
+class StalenessSLA(AssetSLA):
+    """This SLA specifies that an asset must not be more than `allowed_staleness` time out of date
```
sryza, line 13: It would be very helpful (for users, but also for review) to have a full, clear description here.

PR #9714, opened 2022-09-15
[docs] Add ability to include asset preview image next to code block

benpankow:
## Summary

Adapts the `RenderedDAG` component from the Dagster website to display alongside a code block in the docs.

Can be clicked to expand to the full width of the docs page. Hides automatically on mobile.

Right now, to add a DAG next to a code block, the image path must be included as a prop of the code block itself. A stacked diff w ill create a tool which will assist with generating these images.

See Vercel deployment: https://dagster-git-benpankow-display-assets-next-to-code-elementl.vercel.app/getting-started

```
`'`python dagimage=/images/asset-screenshots/my_assets.svg
import csv
import requests
from dagster import asset

...

`'`
```

## Future Work

It could be nice to by default zoom to a specific asset or set of assets, or highlight a specific asset.

## Demo

https://user-images.githubusercontent.com/10215173/190512004-d38bfe8d-43c9-4cb1-90b8-2d244ec5b0e5.mov

## Test Plan

Vercel, tested locally


benpankow:
* master
  * **PR #9714** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9714" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9714?utm_source=stack-comment).

docs/content/getting-started.mdx
```diff
@@ -4,6 +4,21 @@ title: "Welcome to Dagster! | Dagster Docs"
 
 # Welcome to Dagster!
 
+```python dagimage=/images/asset-screenshots/my_assets.svg
```
benpankow, line 7: reminder to remove before merge

docs/next/components/mdx/MDXComponents.tsx
```diff
@@ -338,6 +339,11 @@ const Experimental = () => {
 };
 
 const Pre = ({children, ...props}) => {
+  // Retrieve the DAG image to display next to the code block, if any
+  const childrenArr = React.Children.toArray(children);
+  const preComponent = childrenArr.length > 0 ? childrenArr[0] : null;
+  const dagImage = preComponent ? preComponent['props']['dagimage'] : null;
```
hellendag, line 345: It's pretty hacky to reach into the props of a child component -- could you instead provide `dagImage` as a separate prop, and leave `children` as is?

docs/next/components/mdx/MDXComponents.tsx
```diff
@@ -338,6 +339,11 @@ const Experimental = () => {
 };
 
 const Pre = ({children, ...props}) => {
+  // Retrieve the DAG image to display next to the code block, if any
```
hellendag, line 342: It would be nice to have real props defined for this component, even if it's just `React.HTMLProps<HTMLPreElement>`.

docs/next/components/mdx/MDXComponents.tsx
```diff
@@ -355,53 +361,66 @@ const Pre = ({children, ...props}) => {
...
+          <div className="absolute top-0 right-0 mt-1 mr-2">
+            <span className="inline-flex items-center px-2 rounded text-xs font-medium leading-4 bg-gray-100 text-gray-800">
+              Copied
+            </span>
+          </div>
+        </Transition>
+        <pre
+          ref={preRef}
+          style={{height: '100%', marginBottom: 0, marginTop: 0}}
+          {...(props as any)}
```
hellendag, line 412: With a props type defined for the component, the `any` here should be removable.

docs/next/components/mdx/RenderedDAG.tsx
```diff
@@ -0,0 +1,381 @@
...
+}
+
+interface Point {
+  x: number;
+  y: number;
+}
+
+const MIN_ZOOM = 0.17;
+
+export class SVGViewport extends React.Component<SVGViewportProps, SVGViewportState> {
```
hellendag, line 117: Is this a brand new component, or copied from somewhere? If it's new, we shouldn't really be using `React.Component` at this point, and all of this should be achievable with a function component.

docs/next/components/mdx/RenderedDAG.tsx
```diff
@@ -0,0 +1,381 @@
...
+        ></div>
+      </div>
+    );
+  }
+}
+
+/*
+  BG: Not using styled-components here because I need a `ref` to an actual DOM element.
+  Styled-component with a ref returns a React component we need to findDOMNode to use.
+  */
```
hellendag, line 371: Is styled-components used in the docs site? I don't see it in package.json.

docs/next/components/mdx/RenderedDAG.tsx
```diff
@@ -0,0 +1,381 @@
...
+      }}
+    >
+      <div
+        onMouseEnter={() => setHover(true)}
+        onMouseLeave={() => setHover(false)}
+        style={{
+          lineHeight: 1.1,
+          position: 'absolute',
+          bottom: 0, // 20
+          top: 0, // 20
```
hellendag, line 43: `// 20`?

hellendag (changes requested):
Requesting changes for the child access pattern, a couple questions inline otherwise.

erinkcochran87 (commented):
(I was testing this locally and pushed up a pull from master - sorry about that! I was trying to resolve an error.)

This looks great - I think it'll be really helpful for solidifying understanding of this concept. A few thoughts:

- It looks like this component doesn't pull in external snippets from `examples`, the way `python file=[file]` would - is that correct? If so, that means the code isn't under test the way other snippets are. I'll defer to @yuhan on this, but IIRC we're trying to keep all the snippets under test
- The code block currently has a horizontal scroll (as it should), but no vertical. Do we want to add a max height and enable vertical scrolling for this? I'm thinking of how this might look with longer code snippets
- I think adding some sort of default tooltip/caption/text to this component that explains the available actions (zoom, double-click to center, etc) would be helpful

I really love the side-by-side presentation of this and think there's more we can do with this approach to formatting in the future. For example, having copy and code side-by-side - especially in tutorials and guides - could be really helpful.

benpankow:
> (I was testing this locally and pushed up a pull from master - sorry about that! I was trying to resolve an error.)
> 
> This looks great - I think it'll be really helpful for solidifying understanding of this concept. A few thoughts:
> 
> * It looks like this component doesn't pull in external snippets from `examples`, the way `python file=[file]` would - is that correct? If so, that means the code isn't under test the way other snippets are. I'll defer to @yuhan on this, but IIRC we're trying to keep all the snippets under test

The subsequent PR #9764 adds a command which reads from a snippet file and generates an asset graph image from it. This way, we can update the images whenever we update the snippets (and the snippets themselves will stay under test).

> * The code block currently has a horizontal scroll (as it should), but no vertical. Do we want to add a max height and enable vertical scrolling for this? I'm thinking of how this might look with longer code snippets

Right now the component will just expand vertically the longer the code snippet is (the preview pane will grow correspondingly).

> * I think adding some sort of default tooltip/caption/text to this component that explains the available actions (zoom, double-click to center, etc) would be helpful

I think that could be good, will defer if others have a preference on how this should look.

docs/next/components/mdx/MDXComponents.tsx
benpankow: Yeah, I'd like to resolve this. I'm not confident in the process in which MDX is being converted into React components - somehow the code block is being transformed into a `pre` tag, which is then being wrapped into our `Pre` component. Ideally instead we would just convert the `pre` tag or be able to move some of its props to our `Pre` component.

docs/next/components/mdx/RenderedDAG.tsx
benpankow: Ah, yeah, it's copied from https://github.com/dagster-io/dagster-website/blob/master/components/RenderedDAG.tsx, as with most of this component.

docs/next/components/mdx/RenderedDAG.tsx
benpankow: ah no, this is a casualty of copypasta

docs/next/components/mdx/MDXComponents.tsx
benpankow: Alright... I found /a/ solution: https://github.com/dagster-io/dagster/pull/9714/commits/860d0228c72c46222fd13866e5f7023bd98d699d

It's a bit gross, but I think it's the nicest solution we'll get.

Right now, a code block in MDX (with three backticks) is converted into a <code> tag wrapped in a <pre> tag, which are nicely styled. However, the <code> tag contains all of the props (e.g. the dag image).

The change in the above commit takes the generated <pre> tag and no-ops the styling. Then, it creates a new <pre> tag in our <code> tag replacement, once we know whether the dag image is set or not.

yuhan (commented):
how do you feel about updating the docs lint to reduce the line length. on the mac screen, the code snippets are still being cut, i.e. people need to scroll to see half of the code. but we probably don't want to do that for the code snippets without dagimage, which is tricky for linter to tell.

alternatively, we could make the font even smaller -- the ones on the io site have much smaller fonts. we could specify font size based on "if dagimage"

docs/content/getting-started.mdx
bengotow, line 7: This is a really slick syntax 🙌

docs/next/components/mdx/RenderedDAG.tsx
```diff
@@ -0,0 +1,380 @@
...
+          width: focus ? '100%' : '40.5%',
+          maxWidth: focus ? '100%' : MAX_HOVER_WIDTH,
+          left: 'unset',
+          cursor: focus ? 'grab' : 'zoom-in',
+          zIndex: focus || hover ? 2 : 1,
+          transition:
+            'width 300ms ease-in-out, max-width 300ms ease-in-out, padding-left 300ms ease-in-out, left 300ms ease-in-out',
+        }}
+      >
+        {/* {isAssetGraph && <MaterializeButton />} */}
```
bengotow, line 58: I wonder if we could go ahead and remove the `isAssetGraph` prop? I could see wiring it up to whether or not the image URL contains `-assets` or something like that, but it looks like it's a no-op now that this is commented out (which makes sense, all the buttons could be confusing in the docs)

bengotow (approved):
This is super cool. I'm glad this could be adapted to the docs!

benpankow:
> how do you feel about updating the docs lint to reduce the line length

will do in follow-up!

PR #9717, opened 2022-09-16
api docs for snowflake resource

jamiedemaria:
### Summary & Motivation
Adds doc strings to the Snowflake resource


python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -139,7 +204,15 @@ def execute_queries(
 
         return results if fetch_results else None
 
-    def load_table_from_local_parquet(self, src, table):
+    @public
+    def load_table_from_local_parquet(self, src: str, table: str):
```
jamiedemaria, line 208: @gibsondan git blame says you were the last one to work on this fn, can you confirm that the doc string is accurate?

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -74,8 +79,20 @@ def __init__(self, config: Mapping[str, str], log):  # pylint: disable=too-many-
         self.autocommit = self.conn_args.get("autocommit", False)
         self.log = log
 
+    @public
```
jamiedemaria, line 82: @smackesey I'm still getting a bit used to what should/should not get teh `@public` decorator. This was the only way I got these functions to appear in the API docs, but want to confirm that it's ok to be marking these public

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -74,8 +79,20 @@ def __init__(self, config: Mapping[str, str], log):  # pylint: disable=too-many-
...
+        """Gets a connection to Snowflake as a context manager.
+
+        If using the execute_query, execute_queries, or load_table_from_local_parquet methods,
+        you do not need to create a
+        connection using this context manager.
+
+        Args:
+            raw_conn: If using the sqlalchemy connector, set raw_conn to True to create a raw
+                connection. Defaults to True.
+
```
jamiedemaria, line 94: TODO - add an example here, but need to find a fn on the connection that isn't executing a query

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -94,7 +112,36 @@ def get_connection(self, raw_conn=True):
                 conn.commit()
             conn.close()
 
-    def execute_query(self, sql, parameters=None, fetch_results=False, use_pandas_result=False):
+    @public
+    def execute_query(
+        self,
+        sql: str,
+        parameters: Optional[Mapping[Any, Any]] = None,
```
jamiedemaria, line 119: TODO - see if i can make this type annotation more specific - need to dig into snowflake docs

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -94,7 +123,37 @@ def get_connection(self, raw_conn=True):
                 conn.commit()
             conn.close()
 
-    def execute_query(self, sql, parameters=None, fetch_results=False, use_pandas_result=False):
+    @public
+    def execute_query(
+        self,
+        sql: str,
+        parameters: Optional[Union[Sequence[Any], Mapping[Any, Any]]] = None,
```
jamiedemaria, line 130: this type annotation is pulled from the snowflake type annotation https://github.com/snowflakedb/snowflake-connector-python/blob/0b24148ec0c947764c50f093f6bc1fe1638fb551/src/snowflake/connector/cursor.py#L572

jamiedemaria:
@gibsondan  and @smackesey  bump on this when you get a minute! thanks!

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
sryza: These are the rules that govern our guarantees for public APIs: https://docs.dagster.io/getting-started/releases

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
sryza: This looks right to me

sryza (approved):
Nice!

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -23,6 +24,11 @@
 
 
 class SnowflakeConnection:
+    """A connection to Snowflake that can provide a Snowflake connection and execute queries. In
```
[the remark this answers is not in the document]
jamiedemaria: yeah totally is

PR #9722, opened 2022-09-16
[docs] - Partitioned multi asset sensor examples

clairelin135:
Adds examples to `multi_asset_sensor` documentation to detail several use cases for monitoring asset partitions:
- Triggering a partitioned run after a corresponding upstream partition is materialized
- Updating a weekly asset partition when upstream daily partitions are materialized or replaced
- Materializing a daily asset partition if both upstream daily partitions are materialized


clairelin135:
Side note: these code examples are all of the `OR` case (e.g. if ANY upstream partition is replaced). We don't handle the `AND` case as well--for example, materializing a partition only if both upstreams have been replaced.

I think the majority of cases will be the `OR` case, though I wouldn't be surprised if users also wanted support for the `AND` case. It's a challenge because we can't evaluate materializations out of order due to the current cursor approach

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
```diff
@@ -144,23 +142,19 @@ def my_asset_sensor(context: SensorEvaluationContext, asset_event: EventLogEntry
...
-
-- <PyObject
-  object="MultiAssetSensorEvaluationContext"
-  method="latest_materialization_records_by_key"
-  />
-  :
+In the body of the sensor, you have access to the materialization event records for each asset via the `context`. The context provides methods for fetching the materialization event records, as well as methods to handle partitioned materializations:
 
-Returns a dictionary mapping the `AssetKey` for each monitored asset to the most recent materialization record. If there is no materialization event, the mapped value will be `None`
+| Method                                                                                                      | Description                                                                                                                                                                                                          |
```
sryza, line 147: This is duplicative of the API reference, and I think we should just point people there to avoid the concept page getting massive. Thoughts?

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
```diff
@@ -205,6 +199,52 @@ def every_fifth_asset_c_sensor(context):
...
+    for partition, materialization in materializations_by_partition.items():
+        downstream_partitions = context.get_downstream_partition_keys(
+            partition,
+            from_asset_key=AssetKey("upstream_daily_asset"),
+            to_asset_key=AssetKey("downstream_daily_asset"),
+        )
+        if downstream_partitions:  # Check that a downstream daily partition exists
+            # Upstream daily partition can only map to at most one downstream daily partition
+            yield downstream_daily_job.run_request_for_partition(
+                downstream_partitions[0], run_key=None
```
sryza, line 232: We can omit `run_key` now

sryza: I think we now prefer returning a list of RunRequests rather than yielding them one by one, because it makes it clear that runs aren't requested at yield time but rather once the function has returned.

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
```diff
@@ -205,6 +199,52 @@ def every_fifth_asset_c_sensor(context):
...
+    )
+
+    for partition, materialization in materializations_by_partition.items():
+        downstream_partitions = context.get_downstream_partition_keys(
+            partition,
+            from_asset_key=AssetKey("upstream_daily_asset"),
+            to_asset_key=AssetKey("downstream_daily_asset"),
+        )
+        if downstream_partitions:  # Check that a downstream daily partition exists
+            # Upstream daily partition can only map to at most one downstream daily partition
```
sryza, line 230: It's a little unclear to me what this comment means. Does this apply to all partitioned asset sensors, or just to this particular situation?

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
```diff
@@ -205,6 +199,52 @@ def every_fifth_asset_c_sensor(context):
...
+
+```python file=/concepts/partitions_schedules_sensors/sensors/sensors.py startafter=start_daily_asset_to_daily_asset endbefore=end_daily_asset_to_daily_asset
+downstream_daily_job = define_asset_job(
+    "downstream_daily_job",
+    AssetSelection.keys("downstream_daily_asset"),
+    partitions_def=daily_partitions_def,
+)
+
+
+@multi_asset_sensor(
```
sryza, line 214: Did you consider including multiple assets here? It would make the code more complex, but I think there's value in the general case. It's easy for users to adapt a multi-asset example to a single-asset example, but harder for them to go the other way around.

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
```diff
@@ -205,6 +199,52 @@ def every_fifth_asset_c_sensor(context):
...
+                downstream_partitions[0], run_key=None
+            )
+            context.advance_cursor({AssetKey("upstream_daily_asset"): materialization})
+```
+
+The <PyObject object="MultiAssetSensorEvaluationContext" method="get_downstream_partition_keys"/> accepts a partition key from the upstream asset and uses the existing <PyObject object="PartitionMapping"/> object on the downstream asset to fetch the corresponding partitions in the downstream asset.
+
+If a partition mapping is not defined, Dagster will use the default partition mapping, which is the <PyObject object="TimeWindowPartitionMapping"/> for time window partitions definitions and the <PyObject object="IdentityPartitionMapping"/> for other partitions definitions. The <PyObject object="TimeWindowPartitionMapping"/> will map an upstream partition to the downstream partitions that overlap with it.
+
+#### Additional examples
```
erinkcochran87, line 241: Thoughts on making this a callout instead of a section with a heading? I almost missed it when looking at the preview, and a callout could make it stand out.

Ex:

```
Looking for more? Check out the Examples section!

- [Updating a weekly asset partition when upstream daily partitions are materialized]()
- [Materializing a daily asset partition when both upstream daily partitions are materialized]()
```

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
```diff
@@ -721,6 +761,89 @@ def uses_db_connection():
 
 If a resource you want to initialize has dependencies on other resources, those can be included in the dictionary passed to <PyObject object="build_resources"/>. For more in-depth usage, check out the [Initializing Resources Outside of Execution](/concepts/resources#initializing-resources-outside-of-execution) section.
 
+### Monitoring asset partitions with sensors
+
```
erinkcochran87, line 765: Could you add a list of links to the subsections here? It makes it easy to see what's in this bit and jump around. Ex:

```
- [Updating a weekly asset partition when upstream daily partitions are materialized]()
- [Materializing a daily asset partition if both upstream daily partitions are materialized]()
```

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
```diff
@@ -721,6 +761,89 @@ def uses_db_connection():
...
+#### Updating a weekly asset partition when upstream daily partitions are materialized
```
erinkcochran87, line 766: This is probably a nit, but if there's a way to shorten this heading (and the other one in this section) I think it's worth a shot. It's getting squashed in the page nav and is pretty long for a heading in general.

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
```diff
@@ -721,6 +761,89 @@ def uses_db_connection():
...
+                )
+                # Advance the cursor so we only check event log records past the cursor
+                context.advance_cursor(
+                    {AssetKey("upstream_daily_asset"): materialization}
+                )
+```
+
+#### Materializing a daily asset partition if both upstream daily partitions are materialized
+
+The following example monitors two upstream daily-partitioned assets, kicking off a run in the downstream daily-partitioned asset if any upstream daily partition is replaced and the other upstream daily partition has an existing materialization.
```
erinkcochran87, line 813: I recommend breaking this sentence up into two, at a minimum. It's currently a bit hard to read + absorb due to its length.

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
clairelin135: got it--I will remove this

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
clairelin135: This applied to just this particular situation. In any case, I am removing this code example in favor of only displaying the example of monitoring multiple upstream assets.

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
clairelin135: That makes sense. I've updated this code example instead to be of the multi-asset case

clairelin135:
@sryza @erinkcochran87 I added updates to this page per your feedback, including:
- Featuring the multi-asset monitoring example at the forefront (materializing a daily asset when upstream partitions from 2 assets are materialized). Also deleted the original daily -> daily partition asset example for brevity.
- Formatting changes and text rewording
- Miscellaneous code changes

I think this is ready for you to take another look!

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
```diff
@@ -205,6 +191,58 @@ def every_fifth_asset_c_sensor(context):
...
+        ) in context.latest_materialization_records_by_partition(asset_key).items():
+            # Check that the partition is materialized in the other asset. If so,
+            # yield a run request for the downstream partition.
+            if all(
+                [
+                    context.all_partitions_materialized(other_key, [partition])
+                    for other_key in other_asset_keys
+                ]
+            ):
+                run_requests.append(
```
sryza, line 225: I might be reading this incorrectly, but it seems like we might end up adding `RunRequest`s twice for the same partition?  E.g. if we depend on "asset1" and "asset2" and they both have a materialization for partition "2020-01-02", we'll call `evaluate_latest_partitions` on "asset1", which will add a `RunRequest` and then on "asset2", which will add another `RunRequest`?

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
```diff
@@ -205,6 +191,58 @@ def every_fifth_asset_c_sensor(context):
...
+
+```python file=/concepts/partitions_schedules_sensors/sensors/sensors.py startafter=start_daily_partitioned_asset_with_two_upstreams endbefore=end_daily_partitioned_asset_with_two_upstreams
+@multi_asset_sensor(
+    asset_keys=[AssetKey("upstream_daily_1"), AssetKey("upstream_daily_2")],
+    job=downstream_daily_job,
+)
+def trigger_daily_asset_if_all_upstream_partitions_materialized(context):
+    run_requests = []
+
+    def evaluate_latest_partitions(asset_key):
```
sryza, line 208: Nesting functions inside of other functions is kind of a "fancy" Python feature that might throw some users off. I imagine the idea here is to avoid the text getting squished with all the layers of indenting, but I'm not sure it's worth it.

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
```diff
@@ -205,6 +191,58 @@ def every_fifth_asset_c_sensor(context):
...
+The <PyObject object="MultiAssetSensorEvaluationContext" method="latest_materialization_records_by_partition"/> method fetches the latest materialization records by partition for a given asset. Using this method, you can trigger downstream partitioned runs when the corresponding partitions of upstream assets are materialized.
+
+The following example monitors two upstream daily-partitioned assets, kicking off materializations of the corresponding partition in the downstream daily-partitioned asset. Whenever a partition is replaced in `upstream_daily_1` and the same partition has previously been materialized in `upstream_daily_2` (and vice versa), the sensor yields a run request for the downstream asset.
+
+```python file=/concepts/partitions_schedules_sensors/sensors/sensors.py startafter=start_daily_partitioned_asset_with_two_upstreams endbefore=end_daily_partitioned_asset_with_two_upstreams
+@multi_asset_sensor(
+    asset_keys=[AssetKey("upstream_daily_1"), AssetKey("upstream_daily_2")],
+    job=downstream_daily_job,
+)
+def trigger_daily_asset_if_all_upstream_partitions_materialized(context):
```
sryza, line 205: This is a pretty complex implementation for such a common use case. I wonder if there's additional utility methods we could supply to make it simpler?

sryza:
This is looking in the right direction! I like it how it is with 2 examples instead of 3.

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
clairelin135: Hm... taking into account your other feedback, here's an updated version of this sensor:
```
@multi_asset_sensor(
    asset_keys=[AssetKey("upstream_daily_1"), AssetKey("upstream_daily_2")],
    job=downstream_daily_job,
)
def trigger_daily_asset_if_all_upstream_partitions_materialized(context):
    run_requests_by_partition = {}

    for asset_key in context.asset_keys:
        for (
            partition,
            materialization,
        ) in context.latest_materialization_records_by_partition(asset_key).items():
            if (
                all(
                    [
                        context.all_partitions_materialized(other_key, [partition])
                        for other_key in context.asset_keys
                        if other_key != asset_key
                    ]
                )
                and partition not in run_requests_by_partition
            ):
                run_requests_by_partition[
                    partition
                ] = downstream_daily_job.run_request_for_partition(partition)
            context.advance_cursor({asset_key: materialization})

    return list(run_requests_by_partition.values())
```
One small optimization I can think of is to allow `all_partitions_materialized` to accept a list of asset keys, or use all monitored asset keys by default. But aside from that, it's hard to think of more because we are constrained to advancing the cursor for each monitored asset key sequentially

sryza: What if we had a method that returned a dictionary that mapped each partition key to a list or set of asset keys that had materializations with that partition? (We would probably want to exclude partitions with no materializations or something).  Then the user could just iterate through that list?

clairelin135: Yeah, we could do that. If we did that we'd have to advance all of the cursors in that call (unless if we plan on passing the materialization objects to the user within the call).

sryza: > (unless if we plan on passing the materialization objects to the user within the call).

Is there a downside to doing that? I think in general it's better to separate the methods that return information from the methods that advance the cursor.

clairelin135: I think the only downside is that the returned object gets a little bit more complicated, it would be `Mapping[str, Mapping[AssetKey, List[EventLogEntry]]]`

clairelin135: Agree that we want to avoid advancing the cursor in this call though

sryza: Could it just be `Mapping[str, Tuple[AssetKey, EventLogEntry]`? I.e. only return the latest entry for that asset-partition combo?

clairelin135: Yeah, we can do that.

One nuance with doing something like this is that the materializations will be returned out of order (from the order they occurred). So if we provided this method, users would have to use `context.advance_all_cursors` instead of `context.advance_cursor`.

For example, if asset A had partition A materialized and then partition B, and asset B had partition B materialized and then partition A, this method would return:
```
{
  partition A: [(Asset A, partition a materialization), (Asset B, partition B materialization)],
  partition B: [(asset A, partition b materialization), (Asset B, partition B materialization)]
}
And so calling `advance_cursor` (instead of advance_all_cursors) for partition B after partition A would backtrack the cursor for partition B

**Comment by clairelin135** [comment]
@sryza The code example of kicking off a downstream partition when either of the corresponding upstream partitions are materialized has been updated, following the implementation added in https://github.com/dagster-io/dagster/pull/9856

**Comment by clairelin135** [comment]
@sryza This PR is now updated with the `latest_materialization_by_partition_and_asset` method merged in #9856 

I think it's ready for you to take another look!

File: docs/content/concepts/partitions-schedules-sensors/sensors.mdx [hunk 12]
```diff
@@ -183,6 +183,49 @@ def asset_a_and_b_sensor_with_skip_reason(context):
...
+    run_requests_by_partition = {}
+    for (
+        partition,
+        materializations_by_asset,
+    ) in context.latest_materialization_records_by_partition_and_asset().items():
+        for asset_key, materialization in materializations_by_asset.items():
```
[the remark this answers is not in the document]
sryza: I see. Is there a reason we prefer that behavior over behavior that only rematerializes if all parents are rematerialized? I could see the case for either, but smaller simpler example code seems like a good tiebreaker.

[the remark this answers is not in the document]
clairelin135: I think the "and" case is actually more complex. So say:
- Downstream asset monitors upstreams A and B
- A materializes partitions 1 and 2, in that order
- B materializes partitions 3, 2, 1 in that order
The cursor can't actually be updated unless if the set of partitions in the first N materializations are the same for upstreams A and B. So in this case, until we get a partition 3 materialization in asset A, we can't advance the cursor for B (and thus can't advance the cursor for A)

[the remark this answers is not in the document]
sryza: Oh man, this is gnarly.  Thinking through all this stuff makes me think about how easy it would be for a user to mess this up.

If we had infinite resources, I think we'd ideally store a cursor component for every (asset key, partition) tuple instead of just for every asset key, right? Would that make programming against this significantly simpler? If so, I wonder if there's some data structure that we could use to make this possible.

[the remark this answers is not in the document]
sryza: For example, what if the cursor was a `Mapping[AssetKey, Sequence[StorageIdPartitionKeyRange]]`?  Where each `EventTimePartitionKeyRange` had `start_storage_id`, `end_storage_id`, `start_partition_key`, and `end_partition_key` fields.

`{AssetKey("asset1"): [EventTimePartitionKeyRange(1, 5, "2020-01-01", "2020-05-05")]}` would mean that our sensor has handled every event that has a storage_id between 1 and 5 that has a partition key between "2020-01-01" and "2020-05-05".

clairelin135:
@erinkcochran87 @sryza The implementation for multi assets is now merged! I've updated the docs to contain examples of the new functionality and broken out the asset sensor content into its own page to prevent the page from growing too unwieldy.

Let me know how this looks! https://dagster-hkdzmk6y2-elementl.vercel.app/concepts/partitions-schedules-sensors/asset-sensors

clairelin135:
@sryza oddly, I'm not seeing any new comments from you

sryza:
@clairelin135 they're on the Vercel preview

PR #9735, opened 2022-09-19
[dagit] Split asset graphs onto a new “Plots” tab

bengotow:
### Summary & Motivation

This PR splits the graphs off the Asset Events tab into a new "Plots" tab as shown here: https://www.figma.com/file/DJWVckz0oILsjoOzWuGpiu/Software-Defined-Assets-(1.1)?node-id=71%3A6578

The plots and the events table both rely on the same data, so I created / upgraded a few hooks so that this didn't duplicate a lot of code. I also split out the "sidebar" version of the events page because the new hooks made that cleaner than if/elsing the page component.

This does make the events page look a little plain, but I think we can make some headway there soon. May defer to @braunjj and @sryza  about whether this should merge now or later.

image
image
image


js_modules/dagit/packages/core/src/assets/AssetMaterializationGraphs.tsx
```diff
@@ -22,28 +27,17 @@ export const AssetMaterializationGraphs: React.FC<{
...
-    (label: string) => {
-      setCollapsedLabels((current = []) =>
-        current.includes(label) ? current.filter((c) => c !== label) : [...current, label],
-      );
-    },
-    [setCollapsedLabels],
-  );
+  if (process.env.NODE_ENV === 'test') {
+    return <span />; // chartjs and our useViewport hook don't play nicely with jest
+  }
```
hellendag, line 32: Could we have manual mocks for the stuff that breaks in Jest?

hellendag (approved):

bengotow:
Per discussion with Josh, updated this to put the new split UI behind a feature flag. The only change not feature flagged is the removal of the ability to "collapse" a graph, which we don't think people were actively using.

js_modules/dagit/packages/core/src/assets/AssetMaterializationGraphs.tsx
bengotow: Hmm I think we could, let me try in a follow-up today!

PR #9737, opened 2022-09-19
[dagit] Avoid showing hidden asset jobs on the backfill page

bengotow:
### Summary & Motivation

This PR adds a GraphQL resolver exposing `backfill.assetSelection` and then renders the asset selection in place of the hidden asset job name in the backfills table.

This uses the same "truncating set of asset tags" that we use for the run table, and offers to show a modal if there are more than three asset keys being backfilled.

I renamed this column in the Backfill table from "Partition Set" to "Backfill Target" because it actually shows the backfill repo, job, and partition set in the `op job` case, and showing a `repo + asset list` in a partition set column felt odd.

Before:

image

After: 

image

### DIFF
--- a/js_modules/dagit/packages/core/src/instance/BackfillTable.tsx
+++ b/js_modules/dagit/packages/core/src/instance/BackfillTable.tsx
@@ -19,8 +19,10 @@ import {showCustomAlert} from '../app/CustomAlertProvider';
 import {SharedToaster} from '../app/DomUtils';
 import {usePermissions} from '../app/Permissions';
 import {PythonErrorInfo, PYTHON_ERROR_FRAGMENT} from '../app/PythonErrorInfo';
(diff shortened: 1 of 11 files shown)


js_modules/dagit/packages/core/src/instance/BackfillTable.tsx
```diff
@@ -111,7 +110,7 @@ export const BackfillTable = ({
           <tr>
             <th style={{width: 120}}>Backfill ID</th>
             <th style={{width: 200}}>Created</th>
-            {showPartitionSet ? <th>Partition set</th> : null}
+            {showBackfillTarget ? <th>Backfill Target</th> : null}
```
hellendag, line 113: Backfill target

hellendag (approved):

PR #9740, opened 2022-09-20
Revert "remove dagster cli api subcommand from docs (#9165)"

gibsondan:
This reverts commit a4ffdc4c7ab38be32c04be8c71a960eb87815486. We reference this command in the docs when telling users how to run their own grpc server - i may be missing context on why it was removed originally though.


sryza:
@smackesey @dpeng817 I think this was part of the 1.0 shuffle. Thoughts?

gibsondan:
ping on this - should be straightforward?

PR #9744, opened 2022-09-20
[dagit] Workspace Overview page

hellendag:
### Summary & Motivation

Create a new Workspace overview page, currently behind a flag. The page shows all loaded repositories in a grid of cards, with a bit of metadata about each repo. Click through to a repo will lead to the new virtualized table views.

Because we will now show virtualized tables for single repos, the collapsible repo sections and repo filter button are removed.

Screen Shot 2022-09-20 at 2 15 40 PM

Screen Shot 2022-09-20 at 2 15 51 PM

Enable new workspace page flag, click "Workspace" nav item. Verify card rendering. Shrink viewport to verify card resizing and wrapping, hover on cards to verify styles. Click into a repo, then click Jobs tab. Verify that Jobs, Schedules, and Sensors all load correctly.


hellendag:
* master
  * **PR #9744** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9744" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9744?utm_source=stack-comment).

bengotow (approved):
Wow, lots of code removal here! The repo bucketing was pretty complex - nice that we're moving to single-repo views.

I think the changes look good, but there are two scenarios that worry me a bit:

- If someone only has one repo, is it annoying to see the new root page with one card? Should we auto-redirect into the workspace views?

- If you're viewing one of these repo pages, any buttons that let you reload the repo (is there one, actually? Maybe there should be...) might have an awkward new behavior where if the reload fails the repo location exists but the repo no longer does, you could be looking at a 404 page? We might need to test the reload flow a bit.

hellendag:
> If someone only has one repo, is it annoying to see the new root page with one card? Should we auto-redirect into the workspace views?

I think that would be sensible, though they would then miss out on the "Recently viewed" feature we're also planning to have here. What do you think @braunjj?

> If you're viewing one of these repo pages, any buttons that let you reload the repo (is there one, actually? Maybe there should be...) might have an awkward new behavior where if the reload fails the repo location exists but the repo no longer does, you could be looking at a 404 page? We might need to test the reload flow a bit.

I haven't put a button on there yet, but I think it would be a good idea. I think the failure case you describe might already be a problem on any repo permalink (for jobs, etc.) where you reload and the whole repo is gone...

braunjj:
I think it makes sense to redirect to the repo if there's only one for now at least. Once we start to implement more features or add more content at the root level we can always show it.

In the case of reloading the workspace and having the repository disappear, it might be cool to just redirect users to the workspace root page.

PR #9748, opened 2022-09-20
make run keys optional on run requests

sryza:


gibsondan (approved):
should we update any docs to take out run_key=None for brevity? 

I see a bunch of run_key=None if I grep for mdx

sryza:
> should we update any docs to take out run_key=None for brevity?

probably, but I'm going to leave that as future work for now

PR #9751, opened 2022-09-20
fix time window partition mapping with offset and lossy format

sryza:
### Summary & Motivation

Fixes a user-reported bug.  When a time window PartitionsDefinition has an offset that's not captured in its datetime format (e.g. an hourly offset with a format of YYYY-MM-DD), we were messing up partition mapping.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
@@ -148,11 +148,18 @@ class TimeWindowPartitionsDefinition(
         return partition_def_str
 
     def time_window_for_partition_key(self, partition_key: str) -> TimeWindow:
-        start = self.start_time_for_partition_key(partition_key)
-        return next(iter(self._iterate_time_windows(start)))
+        partition_key_dt = pendulum.instance(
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -148,11 +148,18 @@ def __str__(self) -> str:
...
+        )
+        return next(iter(self._iterate_time_windows(partition_key_dt)))
 
     def start_time_for_partition_key(self, partition_key: str) -> datetime:
-        return pendulum.instance(datetime.strptime(partition_key, self.fmt), tz=self.timezone)
+        partition_key_dt = pendulum.instance(
+            datetime.strptime(partition_key, self.fmt), tz=self.timezone
+        )
+        # the datetime format might not include granular components, so we need to recover them
+        # we make the assumption that the parsed partition key is <= the start datetime
```
gibsondan, line 161: seems like a reasonable assumption since all the offsets move forward? I can't think of any crazy daylight savings time stuff that would break that...

gibsondan (approved):
thanks!

PR #9752, opened 2022-09-21
use multi-asset sensor in hacker news example

sryza:


clairelin135 (commented):
Yay! This is looking good to me. Seems like there are test failures--I think we need to use `build_multi_asset_sensor_context` instead

yuhan (approved):
back to your queue. 
lg2m, lets fix the test and merge!

jamiedemaria (approved):

PR #9754, opened 2022-09-21
[docs] - simplify multi-asset sensor examples and fix formatting

sryza:
### Summary & Motivation

I was reading the multi-asset doc on the sensors concept page when working on using it inside the hacker news example, and I thought there was opportunity to simplify the first example and make the second example more realistic.

I also noticed this list formatting issue, which seems to get caused when PyObject tags are the first entry in a list:

image

Looked at them in the docs site.

### DIFF
--- a/docs/content/concepts/partitions-schedules-sensors/sensors.mdx
+++ b/docs/content/concepts/partitions-schedules-sensors/sensors.mdx
@@ -29,8 +29,6 @@ Sensors allow you to instigate runs based on any external state change.
 | <PyObject object="RunStatusSensorContext" />              | The context object passed to a run status sensor evaluation.                                                                                                                                                                                                                                                                                                            |
 | <PyObject object="build_run_status_sensor_context" />     | A function that constructs an instance of <PyObject object="RunStatusSensorContext" />. This is intended to be used to test a run status sensor.                                                                                                                                                                                                                        |
 
-                                                                                                                                                                                    |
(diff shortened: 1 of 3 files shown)


erinkcochran87:
@yuhan The `mdx-format` script strikes again -.- Perhaps it's worthwhile for us to spend some time hammering on this script so we're not always playing whack-a-mole with it?

There are a few things happening here - if a `PyObject` has more than one property, the `mdx-format` script will format the code on multiple lines only if it isn't preceded by other characters. `PyObjects` with a single property remain inline when the script is run, regardless of whether they're first in a list item.

Ex:
```
## This:

- <PyObject object="MultiAssetSensorEvaluationContext" method="latest_materialization_records_by_key" /> [text_here]

## Results in:

- <PyObject
  object="MultiAssetSensorEvaluationContext"
  method="latest_materialization_records_by_key"
  /> [text_here]
```

Which is causing the weird formatting / line splits reported by Sandy in this issue.

But the following won't be touched, remaining inline as intended:

```
- [text_here] <PyObject object="MultiAssetSensorEvaluationContext" method="latest_materialization_records_by_key" /> [text_here]
```

erinkcochran87 (approved):
@sryza This looks good to me. I prefer the list formatting, but given this will require an `mdx-format` scripting fix, I don't want to block it. We can always come back to re-format it later.

PR #9756, opened 2022-09-21
RFC (more like dump prototype) Typed resource ctor:

schrockn:
This is more of a pattern rather than real technology. I banged this out very quickly so it is not fleshed out. The idea here is to make it norm to provide typed constructors for resources as opposed to configured. The configurable resource is still available for advanced cases.

The big drawback here is that this breaks direct resource invocation. However I think for the spin-up/simple use case this is *far* more self-explantory.

Additionally a natural extension would be to generate config schema from the python types on the typed signature if that is desirable.

### DIFF
--- a/examples/docs_snippets/docs_snippets/guides/dagster/development_to_production/branch_deployments/repository_v2.py
+++ b/examples/docs_snippets/docs_snippets/guides/dagster/development_to_production/branch_deployments/repository_v2.py
@@ -26,11 +26,12 @@ resource_defs = {
                 "database": f"PRODUCTION_CLONE_{os.getenv('DAGSTER_CLOUD_PULL_REQUEST_ID')}",
             }
         ),
-        "snowflake": snowflake_resource.configured(
-            {
-                **snowflake_config,
-                "database": f"PRODUCTION_CLONE_{os.getenv('DAGSTER_CLOUD_PULL_REQUEST_ID')}",
-            }
+        "snowflake": snowflake_resource(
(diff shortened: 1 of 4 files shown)


examples/docs_snippets/docs_snippets/guides/dagster/development_to_production/branch_deployments/repository_v2.py
```diff
@@ -26,11 +26,12 @@
                 "database": f"PRODUCTION_CLONE_{os.getenv('DAGSTER_CLOUD_PULL_REQUEST_ID')}",
             }
         ),
-        "snowflake": snowflake_resource.configured(
-            {
-                **snowflake_config,
-                "database": f"PRODUCTION_CLONE_{os.getenv('DAGSTER_CLOUD_PULL_REQUEST_ID')}",
-            }
+        "snowflake": snowflake_resource(
```
schrockn, line 29: This IMO is the API we should shoot for

schrockn:
To point it out explictly, the level of indirection is so that existing callsites that call configured still work

sryza:
> The big drawback here is that this breaks direct resource invocation. However I think for the spin-up/simple use case this is far more self-explantory.

Is there a fundamental reason it needs to break direct resource invocation?

examples/docs_snippets/docs_snippets/guides/dagster/development_to_production/branch_deployments/repository_v2.py
petehunt: i kinda think this would be better as `SnowflakeResource(...)`

PR #9759, opened 2022-09-21
[dagit] Virtualized table for assets

hellendag:
### Summary & Motivation

Add a virtualized table for Assets within a repo. I'm trying to repurpose a bunch of existing Asset table code for this.

Asset groups are collapsible within the table. A bit of a yolo design here, so let's iterate further.

Screen Shot 2022-09-21 at 1 44 15 PM

View asset tab for a repo in the new Workspace section, verify proper rendering, fetching, collapsing.

### DIFF
--- a/js_modules/dagit/packages/core/src/workspace/VirtualizedAssetTable.tsx
+++ b/js_modules/dagit/packages/core/src/workspace/VirtualizedAssetTable.tsx
@@ -1,0 +1,377 @@
+import {gql, useLazyQuery} from '@apollo/client';
+import {Box, Caption, Colors, Icon, IconWrapper, Tag} from '@dagster-io/ui';
+import {useVirtualizer} from '@tanstack/react-virtual';
+import * as React from 'react';
+import {Link} from 'react-router-dom';
+import styled from 'styled-components/macro';
+
+import {AppContext} from '../app/AppContext';
+import {
+  AssetLatestRunWithNotices,
+  AssetRunLink,
+  ASSET_NODE_LIVE_FRAGMENT,
+  ComputeStatusNotice,
+} from '../asset-graph/AssetNode';
+import {buildLiveDataForNode} from '../asset-graph/Utils';
+import {ASSET_LATEST_INFO_FRAGMENT} from '../asset-graph/useLiveDataForAssetKeys';
+import {AssetActionMenu} from '../assets/AssetActionMenu';
+import {AssetLink} from '../assets/AssetLink';
+import {ASSET_TABLE_FRAGMENT} from '../assets/AssetTable';
+import {assetDetailsPathForKey} from '../assets/assetDetailsPathForKey';
+import {useStateWithStorage} from '../hooks/useStateWithStorage';
+import {TimestampDisplay} from '../schedules/TimestampDisplay';
+import {Container, HeaderCell, Inner, Row, RowCell} from '../ui/VirtualizedTable';
+
+import {LoadingOrNone, useDelayedRowQuery} from './VirtualizedWorkspaceTable';
+import {repoAddressAsString} from './repoAddressAsString';
+import {RepoAddress} from './types';
+import {SingleAssetQuery, SingleAssetQueryVariables} from './types/SingleAssetQuery';
+import {workspacePathFromAddress} from './workspacePath';
+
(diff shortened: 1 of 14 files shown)


hellendag:
* master
  * **PR #9744** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9744" target="_blank">Graphite</a> 
    * **PR #9759** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9759" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9759?utm_source=stack-comment).

js_modules/dagit/packages/core/src/workspace/VirtualizedAssetTable.tsx
```diff
@@ -0,0 +1,376 @@
...
+  const asset = React.useMemo(() => {
+    if (data?.assetOrError.__typename === 'Asset') {
+      return data.assetOrError;
+    }
+    return null;
+  }, [data]);
+
+  const assetKeyAsString = JSON.stringify(path);
+  const liveData = React.useMemo(() => {
+    if (asset?.definition && data?.assetsLatestInfo) {
```
bengotow, line 170: I might be misunderstanding some logic inside useDelayedRowQuery, but I think the .find in here could probably be a `[0]` since we are asking for each asset individually in it's own query now? This is still more robust but we could probably skip the path stringifying if we wanted to?

js_modules/dagit/packages/core/src/workspace/VirtualizedAssetTable.tsx
```diff
@@ -0,0 +1,376 @@
...
+          <Icon name="asset_group" />
+          <ExpandButton onClick={() => onToggle(groupName)} $open={expanded}>
+            <Icon name="arrow_drop_down" size={20} />
+          </ExpandButton>
+          {groupName === UNGROUPED_NAME ? (
+            <div>Ungrouped assets</div>
+          ) : (
+            <>
+              <div>Asset group</div>
+              <strong>{groupName}</strong>
```
bengotow, line 275: I'm not sure we need the "Asset group" label before the group name here, I think the asset group icon is probably enough. Can always come back to this UI polish though.

js_modules/dagit/packages/core/src/workspace/WorkspaceAssetsRoot.tsx
```diff
@@ -0,0 +1,132 @@
...
+    if (data?.repositoryOrError.__typename === 'Repository') {
+      return data.repositoryOrError.assetNodes;
+    }
+    return [];
+  }, [data]);
+
+  const filteredBySearch = React.useMemo(() => {
+    const searchToLower = sanitizedSearch.toLocaleLowerCase();
+    return assetNodes.filter(({assetKey}) =>
+      assetKey.path.some((elem) => elem.toLocaleLowerCase().includes(searchToLower)),
```
bengotow, line 43: Hmm, we don't have to change this now but I think this should match AssetsCatalogTable.tsx:117:

```
  const searchPath = (searchValue || '')
    .replace(/(( ?> ?)|\.|\/)/g, '/')
    .toLowerCase()
    .trim();


        const pathMatch = !searchPath || tokenForAssetKey(a.key).toLowerCase().includes(searchPath);
```

It'd actually be nice to move this "does this asset key match this search string" out to a helper method because it's been through a bunch of iterations to make it more permissive / less permissive and I think we landed with that somehow. (or at least, if we change it again it'd be nice for it to change on both pages)

bengotow (approved):
This looks good to me! Just left a few inline comments. I feel like the asset group rows look a bit tall, but I haven't seen all the other virtualized tables yet so that may be consistent with everywhere else!

bengotow:
image

Could we make it as flat as this maybe? And put the disclosure triangle over on the other side to match?

js_modules/dagit/packages/core/src/workspace/VirtualizedAssetTable.tsx
hellendag: Yeah, I just wasn't sure if it was obvious enough that that's what these rows were, especially if they haven't been expanded.

hellendag:
Updating to this:

image

PR #9760, opened 2022-09-21
Add way to build and deploy the deploy_ecs project locally

gibsondan:
Jordan I know you were/are not a fan of this pattern in the deploy_docker project... it was the only way that I could figure out to deploy a local ECS change to a real ECS cluster though. Open to feedback / alternatives.


examples/deploy_ecs/dagster.yaml
```diff
@@ -12,6 +12,7 @@ run_launcher:
   class: EcsRunLauncher
   config:
     include_sidecars: true
+    secrets_tag: ""
```
jmsanders, line 15: Maybe it's time to deprecate the ol' magic 😉 

And just make the default behavior do nothing. We could deprecate it with a warning over the next few releases if we want?

jmsanders (approved):

PR #9761, opened 2022-09-21
Enable efficient loading of AssetsDefinitions from slow or unreliable systems

OwenKephart:
### Summary & Motivation

There are many cases in which we would like to load AssetsDefinition objects from external services, rather than python code or other files that are checked in to the git repo like a manifest.json

In these cases, loading the definitions may take a significant amount of time, or involve a certain amount of "risk" (i.e. an external API may not respond). Incurring this startup time + risk on every single step launched from a repository containing such definitions is not an acceptable outcome in most scenarios.

This PR allows for a new way to create AssetsDefinitions (CacheableAssetsDefinition) that works as follows:

1. When at least one CacheableAssetsDefinition object is placed inside an @repository-decorated function, a PendingRepositoryDefinition object will be created when the module containing this repository is loaded, rather than a RepositoryDefinition.

2. The CacheableAssetsDefinition object contains two methods. First, a method for generating AssetsDefinitionMetadata. This method can contact external services, instigate long running processes, and what have you. The AssetsDefinitionMetadata is a serializable object which packages up this data into a convenient format. Second, a method for taking that AssetsDefinitionMetadata and returning AssetsDefinition objects. This method should generally not do any heavy processing, as it will need to be called in every subprocess.

3. When loading a repository from a code pointer, you may optionally supply a RepositoryMetadata object. This object will be passed into the PendingRepositoryDefinition's "resolve" function.  If you do not supply a RepositoryMetadata object, for each CacheableAssetsDefinition object, we will call the `get_metadata` function, then pass the resulting metadata to the `get_definition` function, to get AssetsDefinitions. This collection of AssetsDefinitions (plus whatever regular definitions existed in the repository-decorated function) will be used to create the actual RepositoryDefinition object. If you do supply a RepositoryMetadata object, we can skip calling the get_metadata function for each asset, and instead use the cached value.

4. When spinning up the grpc server for the first time, we will have no persistent metadata to fall back on, so all of the metadata will need to be refreshed. However, in processes that need to launch runs from this server, we serialize the already-calculated RepositoryMetadata, and attach it to the ExecutionPlanSnapshot. When creating the DagsterRun from this ExecutionPlanSnapshot, we add a flag on that object indicating if there is any repository metadata or not (to avoid having to fetch that information if it doesn't exist, a fairly small optimization).

5. Inside core_execute_run, when we first load the ReconstructablePipeline's definition, we check if there is any repository metadata available to us, and if there is, we grab the execution plan snapshot, and pull the repository metadata off of there. Then, we generate a new ReconstructablePipeline with this metadata baked into it, such that calling the recon_pipeline.get_definition() will in turn call recon_repository.get_definition(), which will pass in that RepositoryMetadata object when calling PendingRepositoryDefinition.resolve()

Unit tests in several places, as well as manually testing this workflow in dagit.

### Notes

This system could be useful for things other than AssetsDefinitions. Some users load definitions and configuration from internal databases, which could probably benefit from this sort of treatment. Not a current priority.


python_modules/dagster/dagster/_core/definitions/assets_lazy.py
```diff
@@ -0,0 +1,86 @@
...
+            # cache this generated metadata
+            # pylint: disable=protected-access
+            instance.run_storage._add_snapshot(
+                snapshot_id=self.unique_id,
+                snapshot_obj=metadata,
+                snapshot_type=SnapshotType.PIPELINE,
+            )
+        return self.generate_assets(metadata)
+
+    def generate_metadata(self) -> AssetsDefinitionMetadata:
```
benpankow, line 78: Some initial API feedback - been trying to port #9702 over to this system and one thing that's tricky is that my PR outputs a variable number of `AssetsDefinitions` depending on the number of connections. However, `generate_metadata` can only output a single `AssetsDefinitionMetadata`, so I have to merge all of my `AssetsDefinitions` into a single `Metadata` object and then break them out again in `generate_assets`.

This would be a lot nicer if instead `generate_metadata` could output a list.

benpankow:
* master
  * **PR #9761** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9761" target="_blank">Graphite</a>  👈
    * **PR #9766** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9766" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9761?utm_source=stack-comment).

python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_lazy_assets.py
```diff
@@ -0,0 +1,98 @@
+from typing import Sequence
```
OwenKephart, line 1: these tests need to be updated, pls ignore

python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
```diff
@@ -0,0 +1,78 @@
...
+@whitelist_for_serdes
+class AssetsDefinitionMetadata(
+    NamedTuple(
+        "_PipelineSnapshot",
+        [
+            ("input_keys", FrozenSet[AssetKey]),
+            ("output_keys", FrozenSet[AssetKey]),
+            ("asset_deps", Optional[Mapping[AssetKey, FrozenSet[AssetKey]]]),
+            ("group_names_by_key", Optional[Mapping[AssetKey, str]]),
+            ("metadata_by_key", Optional[Mapping[AssetKey, MetadataUserInput]]),
```
benpankow, line 20: running into a few issues with this - the `check` check seems to throw an error if the input is non-`None` & the `AssetKey`-keyed dict is throwing a JSON serialization error as part of GRPC communication

python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def __new__(
+        cls,
+        input_keys: AbstractSet[AssetKey],
+        output_keys: AbstractSet[AssetKey],
+        asset_deps: Optional[Dict[AssetKey, AbstractSet[AssetKey]]] = None,
+        group_names_by_key: Optional[Dict[AssetKey, str]] = None,
+        metadata_by_key: Optional[Dict[AssetKey, MetadataUserInput]] = None,
+    ):
+        asset_deps = check.opt_nullable_dict_param(
```
benpankow, line 36: the `value_type=set` doesn't seem to work properly here, gives me an error

python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+@whitelist_for_serdes
+class AssetsDefinitionMetadata(
+    NamedTuple(
+        "_PipelineSnapshot",
+        [
+            ("input_keys", FrozenSet[AssetKey]),
+            ("output_keys", FrozenSet[AssetKey]),
+            ("asset_deps", Optional[Mapping[AssetKey, FrozenSet[AssetKey]]]),
```
benpankow, line 18: as below, `AssetKey`-ed map doesn't seem to play nice with GRPC JSON marshalling

python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
```diff
@@ -0,0 +1,97 @@
...
+            ),
+            key_prefix=check.opt_inst_param(key_prefix, "key_prefix", (str, list)),
+            can_subset=check.opt_bool_param(can_subset, "can_subset", default=False),
+            extra_metadata=extra_metadata,
+        )
+
+    def __hash__(self):
+        """Some of the values of this NamedTuple may be recursive dictionaries which are not
+        hashable. Instead, we serialize this as a namedtuple and hash that string.
+        """
```
alangenfeld, line 73: recursive freeze the dicts? serializing is not cheap when this data structure gets big

python_modules/dagster/dagster/_core/definitions/repository_definition.py
```diff
@@ -1326,6 +1366,64 @@ def __call__(self, *args, **kwargs):
         return self
 
 
+class PendingRepositoryDefinition:
+    def __init__(
+        self,
+        name: str,
+        repository_definitions: Sequence[RepositoryDefinitionType],
```
alangenfeld, line 1373: confusing name

python_modules/dagster/dagster_tests/core_tests/snap_tests/snapshots/snap_test_execution_plan.py
```diff
@@ -22,6 +22,7 @@
     "step_output_versions": []
   },
   "pipeline_snapshot_id": "4e770878de7d4f59d0c4841924046e1310918f5b",
+  "repository_metadata": null,
```
alangenfeld, line 25: consider doing the serdes thing where we omit this new key on empty

python_modules/dagster/dagster/_core/definitions/decorators/repository_decorator.py
```diff
@@ -88,8 +99,17 @@ def __call__(self, fn: Callable[[], Any]) -> RepositoryDefinition:
...
                 )
+            if cacheable_defns:
+                return PendingRepositoryDefinition(
+                    self.name,
+                    repository_definitions=repository_defns,
+                    cacheable_definitions=cacheable_defns,
+                    description=self.description,
+                    default_executor_def=self.default_executor_def,
+                    default_logger_defs=self.default_logger_defs,
+                )
```
alangenfeld, line 110: bit worried about how this branching path will age - this code is already really gnarly having divergence like this makes it worse

alangenfeld (commented):
first quick pass

python_modules/dagster/dagster/_grpc/impl.py
```diff
@@ -86,6 +86,14 @@ def core_execute_run(
 
     # try to load the pipeline definition early
     try:
+        # add in cached metadata to load repository more efficiently
+        if pipeline_run.has_repository_metadata:
```
alangenfeld, line 90: skeptical on the stored bool - do we end up fetching this a plan anyway down the call stack?

OwenKephart: In cases where there we have solids_to_execute or asset_selection, we don't end up fetching the stored plan, and instead create a new one in this process (https://sourcegraph.com/github.com/dagster-io/dagster/-/blob/python_modules/dagster/dagster/_core/execution/api.py?L952:5)

So in cases where we have an asset selection and no repository load metadata, this check would save us from having to pull that from the instance. Sympathetic to the argument that this is not a particularly important optimization, so happy to take it out.

python_modules/dagster/dagster/_core/definitions/decorators/repository_decorator.py
OwenKephart: yeah that's legit -- refactored it a bit such that it doesn't need to branch at that point (although the logic is still fundamentally branching, so the amount of gnarl removed is pretty minimal)

alangenfeld (commented):
Did you test cloud yet as well?

Should we be concerned about someone who has a python test that interacts with a `@repository` -> `RepositoryDefinition` directly that ends up switching to `PendingRepositoryDefinition` when a definition thats pulled in changes? I guess they just call the "build from scratch" method.

I think this is getting close - by its nature deserves a lot of tests if we want to ensure this optimized load path is correctly triggered across the diversity of execution paths we support. 

I think it is worth dialing in the naming of this stuff a bit.`metadata` especially is very vague. `cacheable` is ok but I'm hopeful we can do better. Given the very specific intent of this feature i think more specific naming would be ideal.  One idea is `hydrate` / `rehydrate` coming from the React use of that language, though it is only vaguely like that.

python_modules/dagster/dagster/_core/definitions/reconstruct.py
```diff
@@ -668,13 +710,17 @@ def repository_def_from_target_def(target: object) -> Optional["RepositoryDefini
         )
     elif isinstance(target, RepositoryDefinition):
         return target
+    elif isinstance(target, PendingRepositoryDefinition):
+        return target.resolve(repository_metadata)
```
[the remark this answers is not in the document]
OwenKephart: yeah makes sense

python_modules/dagster/dagster/_cli/api.py
```diff
@@ -386,6 +386,13 @@ def _execute_step_command_body(
             if not success:
                 return
 
+        if pipeline_run.has_repository_metadata:
```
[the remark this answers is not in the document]
OwenKephart: those both go through core_execute_run, which has this same logic for fetching the metadata.

OwenKephart:
@alangenfeld yeah definitely interested in improving the naming. I made a couple of minor adjustments, so we now have:

- `CachedAssetsData`
- `CacheableAssetsDefinition`
- `RepositoryLoadData`

I feel pretty good about `RepositoryLoadData` (it's data that is used to load the repository, and we may add other sorts of thing to this object in the future that is not just related to cached assets, so it's fairly flexible).

As for the asset stuff, I do think "caching" is a pretty accurate description of what's going on here (although I'm open to other options). Rather than having to construct this data in every subprocess, we can just refer to the cached information we have. While hydrate/rehydrate is accurate(ish) as well, imo the point of this feature is its caching properties.

I still don't really like the first two names though. It's not really a "CacheableAssetsDefinition", it's really a "Cache-compatible AssetsDefinition Generator". Maybe "CacheableAssetsDefinitionsGenerator" is ok in the short term (we don't have to serialize this thing).

"CachedAssetsData" could also be:

- CachedAssetsDefinitionData
- CachedAssetsDefinition
- AssetsDefinitionData

python_modules/dagster/dagster/_core/definitions/reconstruct.py
```diff
@@ -668,13 +710,20 @@ def repository_def_from_target_def(target: object) -> Optional["RepositoryDefini
         )
     elif isinstance(target, RepositoryDefinition):
         return target
+    elif isinstance(target, PendingRepositoryDefinition):
+        # must regenerate metadata
+        if repository_load_data is None:
+            repository_load_data = target.get_repository_load_data()
```
alangenfeld, line 716: to get something different in the stack trace on cold load we need a different method to hit on `target` here that internally does the load -> resolve

python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
```diff
@@ -0,0 +1,106 @@
...
+
+class CacheableAssetsDefinition:
+    def __init__(self, unique_id: str):
+        self._unique_id = unique_id
+
+    @property
+    def unique_id(self) -> str:
+        return self._unique_id
+
+    def get_cached_data(self) -> Sequence[CachedAssetsData]:
```
alangenfeld, line 96: riffing
* `fetch_cachable_data`
* `compute_cachable_data`
* `compute_cachable_form`
* `fetch_cachable_form`
* `produce_intermediate_representation`

python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
```diff
@@ -0,0 +1,106 @@
...
+        """Returns an object representing cacheable information about assets which are not defined
+        in Python code.
+        """
+        raise NotImplementedError()
+
+    def get_definitions(
```
alangenfeld, line 102: more riffing:
* `build_definitions` 
* `build_definitions_from_data`
* `build_from_ir`

python_modules/dagster/dagster/_core/definitions/repository_definition.py
```diff
@@ -1379,6 +1416,78 @@ def __call__(self, *args, **kwargs):
...
+        self._name = name
+        self._description = description
+        self._default_logger_defs = default_logger_defs
+        self._default_executor_def = default_executor_def
+
+    @property
+    def name(self) -> str:
+        return self._name
+
+    def get_repository_load_data(self) -> RepositoryLoadData:
```
alangenfeld, line 1444: `build` / `fetch` / `compute` - to telegraph expected weight of execution, `get` implies cheap/free (at least to me)

python_modules/dagster/dagster/_core/definitions/repository_definition.py
```diff
@@ -1379,6 +1416,78 @@ def __call__(self, *args, **kwargs):
...
+
+        return RepositoryLoadData(
+            cached_data_by_key={
+                defn.unique_id: defn.get_cached_data()
+                for defn in self._repository_definitions
+                if isinstance(defn, CacheableAssetsDefinition)
+            }
+        )
+
+    def resolve(self, repository_load_data: RepositoryLoadData) -> RepositoryDefinition:
```
alangenfeld, line 1455: re good stacks would put the meat of resolution in a private `_` prefixed method and have 2 distinct entry points ie  `warm_load(data)`  / `cold_load()`

python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
```diff
@@ -0,0 +1,106 @@
...
+            ("keys_by_output_name", Optional[Mapping[str, AssetKey]]),
+            ("internal_asset_deps", Optional[Mapping[str, AbstractSet[AssetKey]]]),
+            ("group_name", Optional[str]),
+            ("metadata_by_output_name", Optional[Mapping[str, MetadataUserInput]]),
+            ("key_prefix", Optional[CoercibleToAssetKeyPrefix]),
+            ("can_subset", bool),
+            ("extra_metadata", Optional[Mapping[Any, Any]]),
+        ],
+    )
+):
```
alangenfeld, line 27: do we expect users to produce these or just us / library authors? Are you confident this data structure is right for this? Changing this can be a pain since we are persisting it. Having it be in the execution plan snapshots mean we will be loading this anytime some one looks at the associated run in dagit.

alangenfeld (commented):
"intermediate representation" comes to mind but is quite a mouthful and `AssetsIR` is probably too terse. 

I don't like "cached" since the tense isn't quite right to me. Changing the data piece to "cachable" i think would be better. `CachableData` `CachableForm` `CachableRepresntation` `IntermediateRepresentation` `IR`

python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
OwenKephart: I think there will be some advanced users that will want to use this behavior (essentially for home-grown libraries), but for the most part we'll be the ones creating these things

I'm confident that this is a reasonably accurate structure (it mirrors the arguments of the non-experimental AssetsDefinition.from_op), but I'm definitely not 100% confident in it.

python_modules/dagster/dagster/_core/definitions/repository_definition.py
OwenKephart: nice yeah

python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
OwenKephart: agree -- I like `compute` as the verb, as some cases the expensive process might not involve hitting (fetching from) an external API, although that will be the most common use case

alangenfeld (approved):
i remain mildly concerned about how this complexity will age - but the utility is pretty high here so hopefully it all works out. 

🙏

OwenKephart:
@alangenfeld 🙏

simonvanderveldt:
@OwenKephart @alangenfeld Small question about this change, I was wondering why this is enabled by default when probably most people just have their repository/Python modules locally and don't need to do any remote/network actions to instantiate their Jobs/Assets. This PR seems to add useful functionality, but I'd expect only for a minority of use-cases. Or were there performance issues before this PR even with only local Python modules?

Asking because we automatically generate our Assets (also only from local code/data) and this caching change prevented changes to be automatically be picked up (we've found a workaround in the meantime).

OwenKephart:
Hi @simonvanderveldt!

This is a purely additive feature, it shouldn't interfere with any of your custom code. If you don't manually create a CacheableAssetsDefinition, the repository loading behavior will be unchanged. Even if you do add a CacheableAssetsDefinition to your repository (e.g. if you are using the Airbyte integration), the loading behavior for the non-cacheable assets will be unchanged.

Just functionally, loading a repository in a separate process (i.e. to execute your code) still requires running all the relevant code to generate that repository, so any changes you make should be reflected in executed steps. Is that not the behavior you're seeing?

The one place where definitions stick around is the GRPC server, which serves Dagit representations of the assets you have in your repository (among other things). This will persist that representation until you hit the "reload definitions" button, or otherwise bring the the server down and back up. However, this works the same regardless of cacheable/non-cacheable asset stuff.

simonvanderveldt:
@OwenKephart Thanks for the additional explanation. The reason I posted the question was that we were using `@repository` before and that no longer worked/we no longer saw our changes reflected in Dagit. We managed to work around this/solve this by subclassing `RepositoryData` which isn't cached. I'm not sure if this was or wasn't the intention? Semantically we'd prefer to keep using `@repository` but there's not really any practical issue with using a `RepositoryData` subclass.
I believe this worked fine before this PR, i.e. the changes were visible without restarting/redeploying, but I might be incorrect given your third paragraph.

I'll provide a bit of context below, it touches on some other issues/things we're not sure will work for our usage, but I hope it'll be helpful nonetheless.

We ran into the caching because we tried syncing our Python modules into the repository (The repository/GRPC server is running as a Pod on Kubernetes with a sidecar that syncs the Python modules from a bucket). The syncing worked, but the changes weren't picked up/shown in Dagit. Once we switched from `@repository` to `RepositoryData` they did show up, so we perceived this as a breaking change (for our use-case).
But then the Runs would fail. They failed because apparently the Python modules need to be present on the filesystem of the Run process (another Pod/container in our case, we're using the K8sRunLauncher). We assumed that the Python modules would be serialized and stored only in the repository/GRPC server and both Dagit and the Runs would get them from the repository/GRPC server, ensuring what you see in Dagit is guaranteed to be exactly the same as the code that's being run. This is apparently not the case.

PR #9768, opened 2022-09-22
[fixit] Add library versions to docs

smackesey:
### Summary & Motivation

Shows library versions in addition to core version in docs.

- If core version is >= 1.0, then version dropdown will show e.g. "1.0.9 /0.16.9 (libs)"
- If core version is < 1.0, only one version is shown
- Changed display of `(latest)` for latest version to drop the parentheses and use a lavender background. This was necessary because of the new `(libs)` text.

Browsed dev site locally.

### DIFF
--- a/docs/next/components/VersionDropdown.tsx
+++ b/docs/next/components/VersionDropdown.tsx
@@ -70,14 +70,17 @@ export default function VersionDropdown() {
                 >
                   <div className="px-4 py-3">
                     <p className="text-sm leading-5">
-                      You are currently viewing the docs for Dagster{' '}
+                      You&apos;re currently viewing the docs for Dagster{' '}
                       <span className="text-sm font-medium leading-5 text-gray-900">
                         {currentVersion}
                       </span>
-                      . You can select a different version below. Note that prior to 1.0, all
-                      Dagster packages shared the same version. After 1.0, separate version tracks
-                      were adopted for mature Dagster packages and less mature integration
-                      libraries.
+                      . Select a different version below. Note that prior to 1.0, all Dagster
+                      packages shared the same version. After 1.0, we adopted separate version
+                      tracks to differentiate less mature integration libraries.{' '}
+                      <Link href="/getting-started/releases#dagster-integration-libraries">
+                        <span className="hover:underline cursor-pointer">Learn more</span>
+                      </Link>
+                      .
                     </p>
                   </div>
 
--- a/docs/next/components/VersionDropdown.tsx
+++ b/docs/next/components/VersionDropdown.tsx
@@ -11,14 +11,16 @@ function getLibraryVersion(coreVersion) {
   return parseInt(major) < 1 ? coreVersion : ['0', 16 + parseInt(minor), patch].join('.');
 }
 
+function getLibraryVersionText(coreVersion) {
+  const libraryVersion = coreVersion === 'master' ? 'master' : getLibraryVersion(coreVersion);
+  return libraryVersion === 'master' || coreVersion === libraryVersion
+    ? ''
+    : `/ ${libraryVersion} (libs)`;
+}
+
 export default function VersionDropdown() {
   const {latestVersion, version: currentVersion, versions, asPath} = useVersion();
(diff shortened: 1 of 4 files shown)


smackesey:
* master
  * **PR #9768** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9768" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9768?utm_source=stack-comment).

yuhan (commented):
looking good. should we change the ones in the dropdown as well?
image

docs/next/components/VersionDropdown.tsx
```diff
@@ -6,23 +6,38 @@ import {useVersion} from '../util/useVersion';
...
                     <span className="flex min-w-0 items-center justify-between space-x-3">
-                      <span className="flex-1 min-w-0">
-                        <span className="text-gray-900 dark:text-gray-300 text-xs lg:text-sm truncate">
-                          {currentVersion} {currentVersion === latestVersion && '(latest)'}
+                      <span className="flex-1 min-w-0 text-gray-900 dark:text-gray-300 text-xs lg:text-sm truncate space-x-1">
+                        <span>
+                          {currentVersion} {libraryVersionText}
                         </span>
+                        {currentVersion === latestVersion ? (
+                          <span className="bg-lavender rounded-full px-2 py-1">latest</span>
```
erinkcochran87, line 39: 🔥

docs/next/components/VersionDropdown.tsx
```diff
@@ -6,23 +6,38 @@ import {useVersion} from '../util/useVersion';
...
+function getLibraryVersion(coreVersion) {
+  const [major, minor, patch] = coreVersion.split('.');
+  return parseInt(major) < 1 ? coreVersion : ['0', 16 + parseInt(minor), patch].join('.');
+}
+
+function getLibraryVersionText(coreVersion) {
+  const libraryVersion = coreVersion === 'master' ? 'master' : getLibraryVersion(coreVersion);
+  return libraryVersion === 'master' || coreVersion === libraryVersion
+    ? ''
+    : `/ ${libraryVersion} (libs)`;
```
erinkcochran87, line 18: Is it safe to assume users will know what `libs` means?

docs/next/components/VersionDropdown.tsx
```diff
@@ -59,12 +74,16 @@ export default function VersionDropdown() {
                       <span className="text-sm font-medium leading-5 text-gray-900">
                         {currentVersion}
                       </span>
-                      . You can select a different version below.
+                      . You can select a different version below. Note that prior to 1.0, all
```
erinkcochran87, line 77: A few thoughts here:

- It feels a bit odd to have this copy display in its entirety for versions prior to 1.0 - to me it seems unnecessary
- It's also a bit long for its current location. Perhaps we could have an icon with a tooltip (or something) display when a version with diverging lib versions is selected
   We could link out to the page that explains our versioning methodology, too: `/getting-started/releases#dagster-integration-libraries`

erinkcochran87 (commented):
Thanks for taking this on, Sean! I think this could use a bit more refining, but overall it's a good step in the right direction.

yuhan:
looks look backend wise. defer to erin for how we present the info

docs/next/components/VersionDropdown.tsx
smackesey: - The copy only displays when you click the menu, which is the same regardless of the version you're browsing, right? It's meant to explain why some entries show two versions and some show one version. If you are on an older version and browsing the list of all versions (including the new ones), you want to know this information.
- I agree it's a little long, but I don't see how to compress it while still explaining what must be explained. I attempted a tooltip on hover over the dropdown, but could not find a pre-existing tooltip component in the docs site (or Headless UI framework) and didn't have the time to spend creating a custom one now (I don't think there are tooltips elsewhere in docs).

docs/next/components/VersionDropdown.tsx
smackesey: No, but that's what the explanatory text in the dropdown is for.

erinkcochran87 (approved):

PR #9770, opened 2022-09-22
[dagit] Fetch isAssetJob in WorkspaceContext, rm Launchpad tab flicker

bengotow:
### Summary & Motivation

This PR addresses a "flicker" users have reported, where the Launchpad tab appears briefly and then disappears when the asset node metadata loads and we realize it's an Asset job.

I originally avoided adding isAssetJob to the workspace context because I thought the implementation of the resolver looked expensive, but it's actually cheap because the data it references is already loaded. Pre-fetching that value lets us hide/show tabs correctly synchronously.

I think this would also let us use different icons for asset and op jobs, which might be a nice touch sometime soon.

### DIFF
--- a/js_modules/dagit/packages/core/src/nav/PipelineNav.tsx
+++ b/js_modules/dagit/packages/core/src/nav/PipelineNav.tsx
@@ -92,8 +92,8 @@ export const PipelineNav: React.FC<Props> = (props) => {
   const repoJobEntry = repo?.repository.pipelines.find(
     (pipelineOrJob) => pipelineOrJob.name === pipelineName,
   );
-  const isJob = repoJobEntry?.isJob || false;
-  const isAssetJob = repoJobEntry?.isAssetJob || false;
+  const isJob = !!repoJobEntry?.isJob;
+  const isAssetJob = !!repoJobEntry?.isAssetJob;
 
   // If using pipeline:mode tuple (crag flag), check for partition sets that are for this specific
   // pipeline:mode tuple. Otherwise, just check for a pipeline name match.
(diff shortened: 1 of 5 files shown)


js_modules/dagit/packages/core/src/nav/PipelineNav.tsx
```diff
@@ -88,18 +87,22 @@ export const PipelineNav: React.FC<Props> = (props) => {
...
   const explorerPath = explorerPathFromString(match!.params.selector);
   const {pipelineName, snapshotId} = explorerPath;
-  const isJob = isThisThingAJob(repo, pipelineName);
+
+  const repo = useRepository(repoAddress);
+  const repoJobEntry = repo?.repository.pipelines.find(
+    (pipelineOrJob) => pipelineOrJob.name === pipelineName,
+  );
+  const isJob = repoJobEntry?.isJob || false;
+  const isAssetJob = repoJobEntry?.isAssetJob || false;
```
hellendag, line 96:
```suggestion
  const isJob = !!repoJobEntry?.isJob;
  const isAssetJob = !!repoJobEntry?.isAssetJob;
```

hellendag (approved):

gibsondan:
"but it's actually cheap because the data it references is already loaded" - this came from me by inspecting the code, but we should verify it on a real beefy deployment if we haven't yet

bengotow:
I tested this on a dagster-cloud instance with ~50 assets and wasn't able to measure any perceivable difference so I think we're safe here. (Talked with Sandy about getting a much larger test asset graph so this is easier to validate in the future)

PR #9774, opened 2022-09-22
(feat) Optional Fivetran Historical Resync Parameters

dwallace0723:
### Summary & Motivation
This PR proposes making the `resync_parameters` configuration on the `fivetran_resync_op` Op optional. This is to allow for triggering a historical resync for an entire connector, which Fivetran recommends using a different endpoint for ([link](https://fivetran.com/docs/rest-api/connectors#resyncconnectordatahistoricalsync))
I wrote tests!


OwenKephart (approved):
Looks good! I was initially going to suggest just using Field(Permissive(), is_required=False), but in your version you're allowed to pass in {"resync_parameters": None} explicitly, which I can imagine might be nice sometimes (such as in tests!)

I'll merge this in and it'll be included in next week's release -- thanks for the contribution :)

PR #9775, opened 2022-09-22
[dagit] Multi-colored run batch backgrounds

hellendag:
### Summary & Motivation

Update run batch backgrounds for the run timeline, in situations where the batch contains runs of different statuses.

Specifically, show proportional backgrounds for the different statuses. For example, if a batch as two failed runs and three successful runs, show a background that is 40% red and 60% green, in that order. If a batch as a failed run, a successful run, an in-progress run, and a queued run, show each background color at 25%.

The goal here is to provide better insight into what the batches contain, instead of blowing away all context and just showing a single color.

Screen Shot 2022-09-22 at 2 10 05 PM
Screen Shot 2022-09-22 at 2 19 22 PM

Storybook examples, jest test. View a real run timeline with batches, verify that rendering looks correct.

### DIFF
--- a/js_modules/dagit/packages/core/src/runs/RunTimeline.tsx
+++ b/js_modules/dagit/packages/core/src/runs/RunTimeline.tsx
@@ -25,9 +25,10 @@ import {RepoAddress} from '../workspace/types';
 
 import {RepoSectionHeader, SECTION_HEADER_HEIGHT} from './RepoSectionHeader';
 import {RunStatusDot} from './RunStatusDots';
-import {failedStatuses, inProgressStatuses, queuedStatuses, successStatuses} from './RunStatuses';
+import {failedStatuses, inProgressStatuses, successStatuses} from './RunStatuses';
 import {TimeElapsed} from './TimeElapsed';
 import {batchRunsForTimeline, RunBatch} from './batchRunsForTimeline';
+import {mergeStatusToBackground} from './mergeStatusToBackground';
 
 const ROW_HEIGHT = 32;
 const TIME_HEADER_HEIGHT = 42;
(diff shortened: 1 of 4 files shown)


hellendag:
* master
  * **PR #9775** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9775" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9775?utm_source=stack-comment).

braunjj:
This is great!

js_modules/dagit/packages/core/src/runs/RunTimeline.tsx
```diff
@@ -481,7 +442,7 @@ const RunTimelineRow = ({
           return (
             <RunChunk
               key={batch.runs[0].id}
-              $color={mergeStatusToColor(batch.runs)}
+              $background={mergeStatusToBackground(batch.runs)}
```
bengotow, line 445: It looks like this `RunTimelineRow` component would memoize cleanly, maybe it's worth doing now that there's some pretty fancy logic happening in here? (Or we could pull background out to a useMemo, but I feel like memoizing the entire component might be better?)

bengotow (approved):

js_modules/dagit/packages/core/src/runs/RunTimeline.tsx
hellendag: Good thought. The `range` prop on `RunTimelineRow` would change pretty regularly, as the start/end window shifts with every poll result. I'd also worry slightly about memoizing the background value, since it would be memoizing based on an array of run objects. I think the array shouldn't change too often, but I also think most of the batches that get chunked together won't be prohibitively large, so I'm not sure it's worth memoizing.

Let's try it in a heavy timeline and see how the rendering costs look, and I'll add some memos in a followup if it looks worthwhile.

PR #9778, opened 2022-09-23
[dagit] Virtualized graphs table, reuse ops view

hellendag:
### Summary & Motivation

Create a virtualized Graph table and reuse the existing Ops view, in the new Workspace section behind the feature flag.

There's not much in the Graph table right now.

View a repo in the new Workspace section, verify correct rendering and behavior of Graphs and Ops tabs.

### DIFF
--- a/js_modules/dagit/packages/core/src/workspace/VirtualizedGraphTable.tsx
+++ b/js_modules/dagit/packages/core/src/workspace/VirtualizedGraphTable.tsx
@@ -1,0 +1,144 @@
+import {gql, useLazyQuery} from '@apollo/client';
+import {Box, Caption, Colors} from '@dagster-io/ui';
+import {useVirtualizer} from '@tanstack/react-virtual';
+import * as React from 'react';
+import {Link} from 'react-router-dom';
+import styled from 'styled-components/macro';
+
+import {Container, HeaderCell, Inner, Row, RowCell} from '../ui/VirtualizedTable';
+
+import {useDelayedRowQuery} from './VirtualizedWorkspaceTable';
+import {RepoAddress} from './types';
+import {SingleGraphQuery, SingleGraphQueryVariables} from './types/SingleGraphQuery';
+import {workspacePathFromAddress} from './workspacePath';
+
+export type Graph = {name: string; path: string; description: string | null};
+
+interface Props {
+  graphs: Graph[];
+  repoAddress: RepoAddress;
+}
+
+export const VirtualizedGraphTable: React.FC<Props> = ({repoAddress, graphs}) => {
+  const parentRef = React.useRef<HTMLDivElement | null>(null);
+
+  const rowVirtualizer = useVirtualizer({
+    count: graphs.length,
+    getScrollElement: () => parentRef.current,
+    estimateSize: () => 64,
+    overscan: 10,
+  });
+
+  const totalHeight = rowVirtualizer.getTotalSize();
+  const items = rowVirtualizer.getVirtualItems();
+
+  return (
+    <>
+      <Box
+        border={{side: 'horizontal', width: 1, color: Colors.KeylineGray}}
+        style={{
+          display: 'grid',
+          gridTemplateColumns: '100%',
+          height: '32px',
+          fontSize: '12px',
+          color: Colors.Gray600,
+        }}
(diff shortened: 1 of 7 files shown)


hellendag:
* master
  * **PR #9778** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9778" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9778?utm_source=stack-comment).

js_modules/dagit/packages/core/src/workspace/VirtualizedGraphTable.tsx
```diff
@@ -0,0 +1,144 @@
...
+          display: 'grid',
+          gridTemplateColumns: '100%',
+          height: '32px',
+          fontSize: '12px',
+          color: Colors.Gray600,
+        }}
+      >
+        <HeaderCell>Graph</HeaderCell>
+      </Box>
+      <div style={{overflow: 'hidden'}}>
```
bengotow, line 49: Hmm what impact does this `<div>` have?  I didn't think hidden would do anything on it's own -- does this do something similar to a `min-width:0`?

js_modules/dagit/packages/core/src/workspace/VirtualizedGraphTable.tsx
```diff
@@ -0,0 +1,144 @@
...
+          repositoryLocationName: repoAddress.location,
+          graphName: name,
+        },
+      },
+    },
+  );
+
+  useDelayedRowQuery(queryGraph);
+  const {data} = queryResult;
+
```
bengotow, line 101: I don't know that the lazy per-row query is super worth it for just the descriptions - as far as I've seen they're not usually present.  I guess it's good to have this stuff in place for future use, though.

js_modules/dagit/packages/core/src/workspace/WorkspaceGraphsRoot.tsx
```diff
@@ -0,0 +1,172 @@
...
+import {WorkspaceGraphsQuery, WorkspaceGraphsQueryVariables} from './types/WorkspaceGraphsQuery';
+
+export const WorkspaceGraphsRoot = ({repoAddress}: {repoAddress: RepoAddress}) => {
+  useTrackPageView();
+
+  const [searchValue, setSearchValue] = React.useState('');
+  const selector = repoAddressToSelector(repoAddress);
+
+  const queryResultOverview = useQuery<WorkspaceGraphsQuery, WorkspaceGraphsQueryVariables>(
+    WORKSPACE_SCHEDULES_QUERY,
```
bengotow, line 22: Hmmm might want to rename this var

bengotow (approved):
This looks good to me! 

Sidenote - I vaguely wish there was less boilerplate involved in making a new Workspace___Root.tsx, but it seems like there are just enough code changes in just enough places that it'd be difficult to make something re-usable? Having 5 simple files with similar code does seem better than one complicated one...

js_modules/dagit/packages/core/src/workspace/VirtualizedGraphTable.tsx
hellendag: Yeah it felt a little weird on this since there's nothing else there. But I figured there might be more we want to add that we've never bothered to add for Graphs, so I put it here. 🤷

hellendag:
> Sidenote - I vaguely wish there was less boilerplate involved in making a new Workspace___Root.tsx, but it seems like there are just enough code changes in just enough places that it'd be difficult to make something re-usable? Having 5 simple files with similar code does seem better than one complicated one...

There is definitely room for abstracting out shared pieces. With a fair amount still in flux I didn't want to abstract anything out too soon.

I am in favor of five similar files that reuse pieces in common ways instead of one file that tries to do everything. :)

js_modules/dagit/packages/core/src/workspace/VirtualizedGraphTable.tsx
hellendag: This restricts the height of the virtualized table so that we don't end up scrolling past it.

PR #9780, opened 2022-09-23
[dagster-airflow] Add DagsterOperator and associated airflow abstractions

Ramshackle-Jamathon:
### Summary & Motivation

This PR defines an airflow plugin(1.0.0+) and provider(2.0.0+ only) that provides the following to users building airflow dags

The plugin:
- provides a DagsterOperator and DagsterAirflowOperator that allow users to orchestrate dagster jobs from their airflow dag
- A DagsterLink that will direct link to the UI for each task instance of the operators
- A DagsterHook that can be used for defining other custom operators

The provider:
- defines a custom Dagster connection type that makes configuring (Organization ID, deployment name, user token, dagit url) much easier, if users are using airflow 1.0.0 they will need to pass these things to the operator instead.

This operator is mainly designed to work with dagster cloud but can work with OSS dagit so long as there is networking available between the dagit and airflow instances.

I've been running local airflow/dagit

### Todos

- [x] add a testing harness for running this on airflow 1 and 2, with and without an oss dagit

### DIFF
--- a/python_modules/libraries/dagster-airflow/dagster_airflow/hooks/dagster_hook.py
+++ b/python_modules/libraries/dagster-airflow/dagster_airflow/hooks/dagster_hook.py
@@ -1,0 +1,252 @@
+import json
+import time
+from typing import Any, Dict, Optional
+
+import requests
+from airflow.exceptions import AirflowException
+from airflow.hooks.base_hook import BaseHook
+from airflow.models import Connection
+
+from dagster._core.storage.pipeline_run import PipelineRunStatus
+
+
+class DagsterHook(BaseHook):
+
+    conn_name_attr = "dagster_conn_id"
+    default_conn_name = "dagster_default"
+    conn_type = "dagster"
+    hook_name = "Dagster"
+
+    @staticmethod
+    def get_ui_field_behaviour() -> Dict[str, Any]:
+        """Returns custom field behaviour"""
+        return {
+            "hidden_fields": ['port', 'schema', 'extra'],
+            "relabeling": {
+                'description': 'Dagster Cloud Organization ID',
+                'host': 'Dagster Cloud Deployment Name',
+                'login': 'Dagster URL',
+                'password': 'Dagster Cloud User Token',
+            },
+            "placeholders": {
+                'password': '',
+                'login': "https://dagster.cloud/",
+                'description': '',
+                'host': 'prod',
+            },
+        }
+
+    def __init__(self, dagster_conn_id: Optional[str] = "dagster_default", url: Optional[str] = None, user_token: Optional[str] = None) -> None:
+        super().__init__()
+        self.url = url
+        self.user_token = user_token
+        if dagster_conn_id is not None:
+            conn = self.get_connection(dagster_conn_id)
+            base_url = conn.login if conn.login else "https://dagster.cloud/"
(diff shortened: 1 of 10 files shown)


python_modules/libraries/dagster-airflow/dagster_airflow/hooks/dagster_hook.py
```diff
@@ -0,0 +1,252 @@
...
+                "selector":{
+                    "repositoryName": repository_name,
+                    "repositoryLocationName": repostitory_location_name,
+                    "jobName": job_name
+                },
+                "mode":"default",
+                "executionMetadata":{
+                    "tags":[
+                        {"key":"dagster/solid_selection","value":"*"}
+                    ]
```
yuhan, line 129: do we need this tag? -- not specifying anything results in "execute all" as well. if we do need it, i thought we've updated it to op_selection cc @dpeng817

python_modules/libraries/dagster-airflow/dagster_airflow/hooks/dagster_hook.py
```diff
@@ -0,0 +1,252 @@
...
+        response.raise_for_status()
+        response_json = response.json()
+
+        if response_json['data']['runOrError']['__typename'] == 'Run':
+            status = response_json['data']['runOrError']['status']
+        else :
+            raise AirflowException(f'Error fetching run status: {response_json["data"]["runOrError"]["message"]}')
+
+        if status == PipelineRunStatus.SUCCESS.value:
+          print(f"Run {run_id} completed successfully")
```
yuhan, line 192: out of curiosity, where would the prints go?

python_modules/libraries/dagster-airflow/dagster_airflow/links/dagster_link.py
```diff
@@ -0,0 +1,19 @@
...
+        run_id = ti.xcom_pull(task_ids=operator.task_id, key="run_id")
+        organization_id = ti.xcom_pull(task_ids=operator.task_id, key="organization_id")
+        deployment_name = ti.xcom_pull(task_ids=operator.task_id, key="deployment_name")
+
+        if run_id and organization_id and deployment_name:
+            return LINK_FMT.format(
+                organization_id=organization_id, deployment_name=deployment_name, run_id=run_id
+            )
+        else:
+            return ""
```
yuhan, line 19: is `return ""` a good practice in airflow's OperatorLink?

yuhan (commented):
is the plan to add tests before the release? overall the implementation looks pretty reasonable to me. but without test cases, it's a bit hard for me to imagine how the new operator will get used.

python_modules/libraries/dagster-airflow/dagster_airflow/hooks/dagster_hook.py
Ramshackle-Jamathon: they'll go into airflows log system

python_modules/libraries/dagster-airflow/dagster_airflow/links/dagster_link.py
Ramshackle-Jamathon: yep an empty link will be a non-clickable button

yuhan (approved):

PR #9781, opened 2022-09-23
Asset Selection on Multi Asset Sensor

clairelin135:
Modifies the multi asset sensor decorator to accept either a list of asset keys or an `AssetSelection`. This will enable use cases like:
```
@multi_asset_sensor(asset_selection=AssetSelection.groups("X").upstream(depth=1, include_self=False), job=all_assets_job)
def materialize_upon_upstreams_materialized(context):
    pass
```
This change also requires modifying the `AssetSelection.upstream` and `AssetSelection.downstream` methods to not include a selected asset in its own upstream/downstream results. I've added a flag called `include_self` on these methods.

There is one edge case here: if an upstream asset A and downstream asset B are both selected:
```
AssetSelection.keys("asset_a", "asset_b").downstream(include_self=False)
```
We will select all of asset b's downstreams as well as asset b itself because the function returns:
```
{asset_b, *asset_b_downstreams} # downstreams of asset a, excluding itself
.union(asset_b_downstreams) # downstreams of asset b, excluding itself
```
I added this in the docstrings to hopefully clarify this. If this isn't clear, we can consider other names for the `include_self` flag.


sryza (commented):
> We will select all of asset b's downstreams as well as asset b itself because the function returns:

I think this would make `include_self` not useful for the scenarios that I had imagined using it in, e.g. "launch a run to materialize all the assets in set X when all the assets upstream them have been materialized", where X might include assets that depend on other assets in X.

Could we make it work the other way? I.e. have `AssetSelection(x).upstream(include_self=False)` resolve to something like `AssetSelection(x).upstream().resolve(...).difference(AssetSelection(x).resolve(...))`?

That makes me think that we could actually offer `.difference` and/or the `-` operator on `AssetSelection`s.  E.g. support `AssetSelection(x).upstream() - AssetSelection(x)`. But I don't have a strong opinion on whether this is important to do or better than `include_self`.

clairelin135:
I wonder if we should also expose another sensor context method
`get_downstream_can_be_materialized_assets(materialized_assets)` that accepts a list of assets that have been materialized and returns a list of the downstream assets that now can be materialized?

sryza:
> I don't think theres a clear ergonomic way to get the current behavior we have with include_self

What if we had a method on AssetSelection that returned all the assets in the selection with no children in the selection? Then you could do `AssetSelection.group("x").upstream(depth=1) - AssetSelection.group("x").sinks()` or similar?

clairelin135:
@sryza I like that. I'll add the functionality to get `sinks` from within an asset selection

clairelin135:
@sryza I've updated the code with the following changes:
- Introduces a difference operator so you can do: `AssetSelection.groups("yay") - AssetSelection.keys("nay")`
- Introduces a sinks operator: `AssetSelection.groups("b").sinks()`
- Changed the `include_self` flag to do `AssetSelection(x).upstream().resolve(...).difference(AssetSelection(x).resolve(...))`
I think this is ready for you to take another look!

python_modules/dagster/dagster/_core/definitions/asset_selection.py
```diff
@@ -67,31 +68,55 @@ def groups(*group_strs) -> "GroupsAssetSelection":
...
         """
         Returns a selection that includes all assets that are downstream of any of the assets in
-        this selection, as well as all the assets in this selection.
+        this selection, selecting the assets in this selection by default. Iterates through each
+        asset in this selection and returns the union of all downstream assets.
 
         depth (Optional[int]): If provided, then only include assets to the given depth. A depth
             of 2 means all assets that are children or grandchildren of the assets in this
             selection.
+        include_self (bool): If True, then include the assets in this selection in the result.
```
sryza, line 82: Worth mentioning the default.

python_modules/dagster/dagster/_core/definitions/asset_selection.py
```diff
@@ -67,31 +68,55 @@ def groups(*group_strs) -> "GroupsAssetSelection":
         return GroupsAssetSelection(*group_strs)
 
     @public  # type: ignore
-    def downstream(self, depth: Optional[int] = None) -> "DownstreamAssetSelection":
+    def downstream(
+        self, depth: Optional[int] = None, include_self: Optional[bool] = True
```
sryza, line 72: No need to make `include_self` Optional if it defaults to `True`.

sryza (approved):
I had a few small comments. Otherwise, this looks great.

PR #9782, opened 2022-09-23
[perf] resolve lru_cache usage issues

alangenfeld:
`@lru_cache(maxsize=...)` does not work as one may intuitively expect for `NamedTuple`. Since there are no object instance to bind to for the tuples, the cache ends up as a process global with the tuple itself being used as the key.

* For the op signature ones i dropped the caches since those were basically not doing anything. The appropriate place to cache that work is probably as close the the fn inspect call as possible.

* For `ReconstructablePipeline` I added a clarifying comment for what the global cache of 1 actually helps with, and a test to go with it. 

* For `ReconstructableRepository` i removed the cache and added explicit holding in `LoadedRepositories`. To pass these already loaded definitions in to grpc server implementations involved cascading changes. I cleaned up the types while i was there.

added test

### DIFF
--- a/python_modules/dagster/dagster/_grpc/impl.py
+++ b/python_modules/dagster/dagster/_grpc/impl.py
@@ -3,14 +3,15 @@
 import os
 import sys
 from contextlib import ExitStack
-from typing import Generator, List, Optional
+from typing import Generator, List, Optional, Sequence
 
 import pendulum
 
 import dagster._check as check
 from dagster._core.definitions import ScheduleEvaluationContext
 from dagster._core.definitions.events import AssetKey
-from dagster._core.definitions.reconstruct import ReconstructablePipeline, ReconstructableRepository
+from dagster._core.definitions.reconstruct import ReconstructablePipeline
+from dagster._core.definitions.repository_definition import RepositoryDefinition
 from dagster._core.definitions.sensor_definition import (
     MultiAssetSensorDefinition,
     MultiAssetSensorEvaluationContext,
@@ -39,6 +40,7 @@ from dagster._core.host_representation.external_data import (
     ExternalSensorExecutionErrorData,
 )
 from dagster._core.instance import DagsterInstance
+from dagster._core.instance.ref import InstanceRef
 from dagster._core.snap.execution_plan_snapshot import (
     ExecutionPlanSnapshotErrorData,
     snapshot_from_execution_plan,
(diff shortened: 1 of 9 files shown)


alangenfeld:
* master
  * **PR #9782** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9782" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9782?utm_source=stack-comment).

python_modules/dagster/dagster/_grpc/impl.py
```diff
@@ -240,19 +242,13 @@ def get_external_pipeline_subset_result(
 
 
 def get_external_schedule_execution(
-    recon_repo,
-    instance_ref,
-    schedule_name,
-    scheduled_execution_timestamp,
-    scheduled_execution_timezone,
+    repo_def: RepositoryDefinition,
+    instance_ref: Optional[InstanceRef],
```
gibsondan, line 246: this seems clearer anyway even if there wasn't a perf benefit - unclear why we were doing this recon repo dance before

python_modules/dagster/dagster/_grpc/server.py
```diff
@@ -429,48 +427,43 @@ def ExternalPartitionSetExecutionParams(self, request, _context):
...
             PipelineSubsetSnapshotArgs,
         )
 
         return api_pb2.ExternalPipelineSubsetSnapshotReply(
             serialized_external_pipeline_subset_result=serialize_dagster_namedtuple(
                 get_external_pipeline_subset_result(
-                    self._recon_pipeline_from_origin(pipeline_subset_snapshot_args.pipeline_origin),
+                    self._get_recon_pipeline_for_origin(
+                        pipeline_subset_snapshot_args.pipeline_origin
+                    ),
```
gibsondan, line 466: so close to removing reconstructableXXX from the server altogether.. but I peeked into get_external_pipeline_subset_result and it does look dicey. Maybe not impossible though?

gibsondan (approved):
i love a good cascading change

python_modules/dagster/dagster/_grpc/server.py
alangenfeld: k i did this - re-requesting review

python_modules/dagster/dagster/_grpc/server.py
```diff
@@ -678,13 +657,18 @@ def StartRun(self, request, _context):
...
-                deserialize_json_to_dagster_namedtuple(request.serialized_execute_run_args),
+            execute_external_pipeline_args = deserialize_as(
+                request.serialized_execute_run_args,
                 ExecuteExternalPipelineArgs,
             )
             run_id = execute_external_pipeline_args.pipeline_run_id
-            recon_pipeline = self._recon_pipeline_from_origin(
-                execute_external_pipeline_args.pipeline_origin
+
+            # reconstructable required for handing execution off to subprocess
```
gibsondan, line 666: much more defensible

gibsondan (approved):
love it

PR #9784, opened 2022-09-23
Clarify `with_resources` error and update docs

clairelin135:
I've seen multiple users get confused by an "asset defined multiple times" error caused by having a call to `with_resources` in the outermost scope: https://dagster.slack.com/archives/C01U954MEER/p1660156122514989

This PR adjusts the code examples in the docs to place `with_resources` calls within repositories to prevent an error if the code snippet is directly copied. Also updates the error message to be more descriptive in situations where assets are defined multiple times due to `with_resources`, along with a disclaimer in the docs to alert users of this behavior.

Addresses Issue https://github.com/dagster-io/dagster/issues/9331


sryza (approved):
This is great

PR #9788, opened 2022-09-26
[k8s launcher] security context

alangenfeld:
to allow recitation of the The Sacred Rites Of Debugging

```
securityContext:
  capabilities:
    add:
    - SYS_PTRACE
```

or other security settings

added tests

### DIFF
--- a/python_modules/libraries/dagster-k8s/dagster_k8s/container_context.py
+++ b/python_modules/libraries/dagster-k8s/dagster_k8s/container_context.py
@@ -37,6 +37,7 @@ class K8sContainerContext(
             ("namespace", Optional[str]),
             ("resources", Mapping[str, Any]),
             ("scheduler_name", Optional[str]),
+            ("security_context", Mapping[str, Any]),
         ],
     )
 ):
@@ -59,6 +60,7 @@ class K8sContainerContext(
         namespace: Optional[str] = None,
         resources: Optional[Mapping[str, Any]] = None,
         scheduler_name: Optional[str] = None,
+        security_context: Optional[Mapping[str, Any]] = None,
     ):
         return super(K8sContainerContext, cls).__new__(
             cls,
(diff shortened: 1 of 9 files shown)


alangenfeld:
* master
  * **PR #9788** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9788" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9788?utm_source=stack-comment).

python_modules/libraries/dagster-k8s/dagster_k8s/container_context.py
```diff
@@ -37,6 +37,7 @@ class K8sContainerContext(
             ("namespace", Optional[str]),
             ("resources", Dict[str, Any]),
             ("scheduler_name", Optional[str]),
+            ("security_context", Dict[str, Any]),
```
gibsondan, line 40: as per offline convo I think future us might be thankful if this was container_config instead of security_context / future things that go on the container, and matched the format of the thing we allow you to put in tags on the job.

container_context : k8s : container_config is a real mouthful though :)

gibsondan (dismissed):
see optional suggestion inline

python_modules/libraries/dagster-k8s/dagster_k8s/container_context.py
alangenfeld: we are in sort of a weird situation where we've already hoisted several of the "container config" entries here on to `DagsterK8sJobConfig` / `K8sContainerContext` 
https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1Container.md

I understand the upside of getting future fields "for free" if we add this "other container settings"  grab bag but it also feels like doubling down on a confusing pattern. 

Theres also the other sibling fields to `container_config` in `UserDefinedDagsterK8sConfig` that im not sure about that have similar collisions with things that have been hoisted here. 

I haven't landed this despite the accept since it feeds bad but I spent a few hours trying `container_config` and it also feels bad for these reasons. Also getting the pydantic model set has been annoying since we can't just set it to the externally defined `Container` since we dont want to enforce that exact schema (and all my attempts in schema field removal in child class failed)

gibsondan (approved):
Seems reasonable - may need an internal PR to expose this in container_context as well

PR #9789, opened 2022-09-26
[dagit] Add reload button to WorkspaceHeader

hellendag:
### Summary & Motivation

Add a "Reload repository location" button to WorkspaceHeader, which is used on all the new Workspace virtualized table pages.

View a repo in the new sections, click to reload. Verify proper rendering, loading behavior, reload confirmation.

### DIFF
--- a/js_modules/dagit/packages/core/src/nav/ReloadRepositoryLocationButton.tsx
+++ b/js_modules/dagit/packages/core/src/nav/ReloadRepositoryLocationButton.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 
 import {AppContext} from '../app/AppContext';
+import {useFeatureFlags} from '../app/Flags';
 import {RepositoryLocationErrorDialog} from '../workspace/RepositoryLocationErrorDialog';
 
 import {buildReloadFnForLocation, useRepositoryLocationReload} from './useRepositoryLocationReload';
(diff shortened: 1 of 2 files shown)


hellendag:
* master
  * **PR #9789** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9789" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9789?utm_source=stack-comment).

bengotow (approved):
This looks good to me! We should probably check what happens if the repository fails to reload though - if we trigger a reload and the repo has an error + no longer exists, the route might no longer be valid and cause the error dialog to be dismissed right as it appears? (this also might not be an issue, but I think we ran into it once before)

hellendag:
> This looks good to me! We should probably check what happens if the repository fails to reload though - if we trigger a reload and the repo has an error + no longer exists, the route might no longer be valid and cause the error dialog to be dismissed right as it appears? (this also might not be an issue, but I think we ran into it once before)

Yeah, I think this is the case any time the user reloads a repo from a repo-specific permalink. When an error occurs, we show a dialog with the relevant information, and on dismissal, redirect to `/workspace`.

Since that URL is going to be a little different now behind the flag, I'll make it redirect to the code location list instead for flagged users.

PR #9792, opened 2022-09-26
load asset values outside of a run

### LINKED ISSUE #9524: API for loading an asset into a Python object
In a notebook, or in an op, you might want to load an asset to inspect it.  Imagine if you could do something like this:

```
@repo
def repo():
    ...

my_repo.load_asset(AssetKey("abc"))
```

It would call the IO manager associated with the asset and load it.  `load_asset` could accept an argument that's a Python type, for IO managers whose Python object depends on a Python type.

Right now, you'd need to instantiate the IO manager associated with the output, call `build_input_context` to create an input context with the relevant information, then call `load_input`.

sryza:
### Summary & Motivation

Fixes #9524.

In a notebook, or in an op, you might want to load an asset to inspect it. Imagine if you could do something like this:

```
from my_module import my_repo

my_repo.load_asset_value("asset_abc")
```

It would call the IO manager associated with the asset and load it.

added tests

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/repository_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/repository_definition.py
@@ -22,10 +22,11 @@ from typing import (
 import dagster._check as check
 from dagster._annotations import public
 from dagster._core.errors import DagsterInvalidDefinitionError, DagsterInvariantViolationError
+from dagster._core.instance import DagsterInstance
 from dagster._core.selector import parse_solid_selection
 from dagster._utils import merge_dicts
 
-from .events import AssetKey
+from .events import AssetKey, CoercibleToAssetKey
 from .executor_definition import ExecutorDefinition
 from .graph_definition import GraphDefinition, SubselectedGraphDefinition
 from .job_definition import JobDefinition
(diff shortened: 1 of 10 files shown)


python_modules/dagster/dagster/_core/definitions/repository_definition.py
```diff
@@ -1295,6 +1297,57 @@ def source_assets_by_key(self) -> Dict[AssetKey, SourceAsset]:
     def _assets_defs_by_key(self) -> Mapping[AssetKey, "AssetsDefinition"]:
         return self._repository_data.get_assets_defs_by_key()
 
+    @public
+    def load_asset_value(
+        self,
+        asset_key: CoercibleToAssetKey,
+        python_type: Optional[Type] = None,
+        instance: Optional[DagsterInstance] = None,
+    ) -> object:
```
OwenKephart, line 1306: should there be some way to apply resource config here? (not necessary for v0, but seems likely to be useful)

would also be nice to be able to specify partition key

OwenKephart (approved):
this is really cool!

I do have some worries about how it will interact with the whole PendingRepositoryDefinition stuff that's incoming, but I think it should be possible to hack around that when the time comes.

PR #9795, opened 2022-09-26
[1/n] Move subscription to web worker

salazarm:
### Summary & Motivation

This PR recreates https://github.com/dagster-io/dagster/pull/8734

Organizations have runs that can output thousands of logs and upwards of 500MB of logs. This causes extreme slowdown in the UI as the mainthread struggles to keep up with the giant spam of logs coming from the backend both when the run is active and especially when the run is completed and all of the logs are loaded upfront. The main source of lag in the UI is deserializing the JSON data. To reduce this time we're planning on only sending the logs/data for things that are actually visible in the UI. 

https://elementl.quip.com/To6cAXcqGIbi/rfc-Improving-Job-Run-Performance

Right now the worker will only be enabled if a query parameter worker=1 is in the URL

Loaded a huge run with dagit-debug and saw everything loaded fine.

Ran a few runs with and without the worker and the worker consistently beat the non worker version:

**many_events.py:**

No worker: 
* 115898.29999999981 ms
* 114627.20000000019 ms
* 115695.5 ms
* 116075.09999999963 ms
Average: 115.5 seconds

Worker: 
* 97091 ms
* 96878.70000000019 ms
* 97609.20000000019 ms
* 97558.39999999944 ms
Average: 97.2 seconds

 ~16% improvement

**Adtriba run debug file:**

No Worker:
* 332907.6000000015 ms
* 328559.40000000224
* 311353.5
* 324224.69999999925

Average: 324.26 seconds

Worker:
* 96900.20000000019
* 102775.20000000298
* 99352.19999999925
* 100879.39999999851

Average: 99.98 seconds

~71% improvement


js_modules/dagit/packages/core/src/runs/usePipelineRunLogsSubscription.tsx
```diff
@@ -0,0 +1,52 @@
...
+    worker.postMessage({
+      type: 'INITIALIZE',
+      runId,
+      staticPathRoot,
+      rootServerURI,
+    });
+    worker.addEventListener('message', (event) => {
+      try {
+        onLogs(JSON.parse(arrayBufferToString(event.data)));
+      } catch (_) {}
```
alangenfeld, line 43: we should make some noise somewhere on failure - this would be really gnarly to debug

alangenfeld (commented):
I see 1/N in the title but the PR summary does not make it clear what the expected scope of this is. From what i can tell this on its own doesn't improve anything, just adds a web worker hop to parsing the same volume of data.

salazarm:
@alangenfeld  Yeah this is just setup for what we want to do which is only send data that would be visible from the webworker to the mainthread so that the mainthread does less worker. 
The linked linear task has more information and heres this quip describing the plan: https://elementl.quip.com/To6cAXcqGIbi/rfc-Improving-Job-Run-Performance


Btw I did do 4 trial runs for each version with and without the worker and the worker version was consistently faster at loading a run with 200k events

js_modules/dagit/packages/core/src/runs/usePipelineRunLogsSubscription.tsx
salazarm: Yeah, I'll remove the try/catch

salazarm:
Just tested this on Adtriba's debug file from a while ago and the improvements are huge there:
No Worker:
* 332907.6000000015 ms
* 328559.40000000224

Worker:
* 96900.20000000019
* 102775.20000000298

About 70% faster

salazarm:
Just tested this on Adtriba's debug file from a while ago and the improvements are huge there:
No Worker:
* 332907.6000000015 ms
* 328559.40000000224
* 311353.5
* 324224.69999999925

Worker:
* 96900.20000000019
* 102775.20000000298
* 99352.19999999925
* 100879.39999999851

I don't see a lot more room for improvement on the frontend but my macbook's python server is pretty slow, on a faster machine its possible it could overwhelm the frontend. I'll see if I can setup dagster on my personal workstation pc

alangenfeld:
What exactly is that time measuring? May be useful to include profiler screenshots. 

Its not obvious to me why things are faster with this change if the volume of data parsed is the same in both approaches - unless 

> the main source of lag in the UI is deserializing the JSON data

isn't right or needs some more specific detail.

salazarm:
> What exactly is that time measuring? May be useful to include profiler screenshots.

I can put up a commit with the test setup. It just measures from pageload until the last log message from the webserver is handled (I used a `requestAnimationFrame` to get the time when rendering of the last message is done)

> Its not obvious to me why things are faster with this change if the volume of data parsed is the same in both approaches - unless
the main source of lag in the UI is deserializing the JSON data
isn't right or needs some more specific detail.

Yeah I'm going to be honest it did catch me by surprise. I can do some more digging to figure out exactly where the time is coming from. I setup my workstation so hopefully a beefier setup will make it easier to profile (my macbook hangs everytime I try to profile the non-worker version)

js_modules/dagit/packages/core/src/workers/util.ts
```diff
@@ -0,0 +1,24 @@
...
+    const index = i % MAX_BUFFER_SIZE;
+    if (index === 0) {
+      buf = new ArrayBuffer(Math.min(MAX_BUFFER_SIZE, str.length - i) * 2); // 2 bytes for each char\
+      view = new Uint16Array(buf);
+      buffers.push(buf);
+    }
+    view[index] = str.charCodeAt(i);
+  }
+  return buffers;
+}
```
hellendag, line 24: The commented URL looks fairly old and has an update note -- is there a more recent native API we can use for this?

js_modules/dagit/packages/core/src/workers/runLogs/runLogs.worker.ts
```diff
@@ -0,0 +1,42 @@
...
+              const buffers = stringToArrayBuffers(JSON.stringify(data));
+              // "When an ArrayBuffer is transferred between threads, the
+              // memory resource that it points to is literally moved between
+              // contexts in a fast and efficient zero-copy operation."
+              self.postMessage('startChunk');
+              buffers.forEach((buffer) => {
+                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                // @ts-ignore https://developer.mozilla.org/en-US/docs/Glossary/Transferable_objects
+                self.postMessage(buffer, [buffer]);
+              });
```
hellendag, line 34: Is it possible to type this and avoid the ts-ignore?

js_modules/dagit/packages/core/src/workers/runLogs/runLogs.tsx
```diff
@@ -0,0 +1,51 @@
...
+    case 'INITIALIZE':
+      initialize(data); // subscribes to the data using data argument
+      break;
+  }
+}
+
+const initialPostMessage = (_data: any): void => {
+  throw new Error('Worker not initialized');
+};
+let postMessage = initialPostMessage;
```
hellendag, line 27: Does this need to be tracked at the module level? I only see it referenced in `initialize`.

js_modules/dagit/packages/core/src/workers/runLogs/runLogs.tsx
```diff
@@ -0,0 +1,51 @@
...
+        if (!logs || logs.__typename === 'PipelineRunLogsSubscriptionFailure') {
+          console.error('PipelineRunLogsSubscriptionFailure', logs);
+          return;
+        }
+        postMessage(logs);
+      },
+      error(error: any) {
+        console.error(error);
+      },
+    });
```
hellendag, line 50: Do these need to be separate `subscribe` calls?

js_modules/dagit/packages/core/src/runs/usePipelineRunLogsSubscription.tsx
```diff
@@ -0,0 +1,59 @@
...
+      runId,
+      staticPathRoot,
+      rootServerURI,
+    });
+    let chunks: string[] = [];
+    worker.addEventListener('message', (event) => {
+      if (event.data === 'startChunk') {
+        chunks = [];
+      } else if (event.data === 'endChunk') {
+        const result = JSON.parse(chunks.join(''));
```
hellendag, line 45: Wrap in a try/catch just in case parsing fails?

js_modules/dagit/packages/core/src/runs/usePipelineRunLogsSubscription.tsx
```diff
@@ -0,0 +1,59 @@
...
+        if (!logs || logs.__typename === 'PipelineRunLogsSubscriptionFailure') {
+          return;
+        }
+        onLogs(logs);
+      },
+    },
+  );
+}
+
+export function usePipelineRunLogsSubscriptionWorker({runId}: any, onLogs: any) {
```
hellendag, line 30: Would be really nice to have the types tightened up on these.

js_modules/dagit/packages/core/src/workers/runLogs/runLogs.tsx
```diff
@@ -0,0 +1,51 @@
+import {ApolloClient} from '@apollo/client';
+
+import {PIPELINE_RUN_LOGS_SUBSCRIPTION} from '../../runs/PipelineRunLogsSubscription';
+import {PipelineRunLogsSubscription} from '../../runs/types/PipelineRunLogsSubscription';
+
+export type Message = INITIALIZE;
+type INITIALIZE = {
```
hellendag, line 7: `type InitializeMessage`? It's a little confusing on the eyes for a type to be `ALL_CAPS` since it looks like a constant value at a glance.

js_modules/dagit/packages/core/src/workers/runLogs/runLogs.worker.ts
```diff
@@ -0,0 +1,42 @@
+import type {Message} from './runLogs';
+
+type SHUTDOWN = {
+  type: 'SHUTDOWN';
+  staticPathRoot?: undefined;
+};
```
hellendag, line 6: Define `INITIALIZE` and `SHUTDOWN` types in the same location in a union so that you can have `data` be a single union type? Just to tidy things up a bit.

js_modules/dagit/packages/core/src/workers/runLogs/runLogs.worker.ts
```diff
@@ -0,0 +1,42 @@
...
+
+self.addEventListener('message', ({data}: {data: Message | SHUTDOWN}) => {
+  if (data.staticPathRoot) {
+    __webpack_public_path__ = data.staticPathRoot;
+  }
```
hellendag, line 11: Is it possible that there is no `staticPathRoot` defined? I feel like we probably always have something, even if it's just `/`, right?

js_modules/dagit/packages/core/src/workers/runLogs/apolloClient.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+  return apolloClient;
+}
+
+export function stop() {
+  if (apolloClient) {
+    apolloClient.stop();
+    apolloClient = undefined;
+  }
+}
```
hellendag, line 34: Nit: maybe call the setup/stop functions `startApolloClient` and `stopApolloClient`? That might help make the callsites a little clearer to follow.

hellendag (commented):
The perf benefit looks really cool. Thanks for putting this together!

salazarm:
I profiled both the worker and non-worker version. Getting a chrome profile is difficult because the buffer fills up pretty fast and the load takes a long time so instead I counted the number of GanttChart re-renders, the number of calls to `throttleSetNodes` and the number of times `throttleSetNodes` actually ran. This is with the 500mb debug file. I ran these runs multiple times and the results were very consistent.

nonWorkerVsWorker
You can see the worker version did more batching because it had less runs of `setNodes` than there were calls to `throttleSetNodes`. 
This is because the GanttChart re-render is slow so the worker ends up queuing multiple messages for the main thread to process. I'm not entirely sure why this behavior doesn't happen without the worker. This means the non-worker version scale linearly with the number of batches of messages while the worker  version takes as long as the webserver takes plus the time of a render for the last batch of messages.

<details><summary>Speculation</summary>
One possibility is that Chrome deserializes the messages lazily (the `event` and `event.data` objects are created lazily) and so in the non-worker version it doesn't queue messages up fast enough before the main thread is busy re-rendering (if deserializing 100k log events takes longer than 100ms then `throttleSetNodes` will fire before the next message meaning no messages are batched). The worker version sends the messages serialized using an `ArrayBuffer` so the messages are received very fast but then we need to deserialize the data on the main-thread.
</details>

In this profile I turned off the GanttChart which is the slowest part of the page to re-render. With the GanttChart disabled there were no batched `throttleSetNode` calls in either the worker or non-worker version because the rendering was done before the web-server sent the next message.

nonWorkerVsWorkerNoGantChart

You can see that the worker version took half as much time! This implies that there is less work being done on the main-thread to receive the message from the worker than without the worker. Unclear why but I think if we looked at Chrome internals we could get a better idea.

js_modules/dagit/packages/core/src/runs/LogsProvider.tsx
```diff
@@ -174,36 +174,39 @@ const useLogsProviderWithSubscription = (runId: string) => {
...
-      fetchPolicy: 'no-cache',
-      variables: {runId, cursor},
-      onSubscriptionData: ({subscriptionData}) => {
-        const logs = subscriptionData.data?.pipelineRunLogs;
-        if (!logs || logs.__typename === 'PipelineRunLogsSubscriptionFailure') {
-          return;
-        }
+  const useHook = new URLSearchParams(window.location.search).get('worker')
+    ? usePipelineRunLogsSubscriptionWorker
+    : usePipelineRunLogsSubscription;
```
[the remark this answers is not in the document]
salazarm: yeah. This makes it easier for me to test too.

js_modules/dagit/packages/core/src/app/PythonErrorFragment.tsx
```diff
@@ -0,0 +1,13 @@
...
+  fragment PythonErrorFragment on PythonError {
+    __typename
+    message
+    stack
+    cause {
+      message
+      stack
+    }
+  }
+`;
```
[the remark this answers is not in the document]
salazarm: yeah, I didn't want the initial PR to have a ton of changes so that its easier to review but I'll do what I did in the original PR where I moved the dependencies around and changed all the imports so that theres only 1 definition

salazarm:
I was able to take a full profile for both versions with the Gantt chart off. 

Non-worker:
Screen Shot 2022-09-28 at 4 27 48 AM
Screen Shot 2022-09-28 at 4 11 31 AM


Worker:
Screen Shot 2022-09-28 at 4 23 38 AM

Screen Shot 2022-09-28 at 4 11 09 AM

In both cases the mainthread spends a lot of time idle just waiting for the next message. In the non-worker case there is a lot more idle time though. Still unclear why... I'm going to ask around.

js_modules/dagit/packages/core/src/workers/runLogs/runLogs.worker.ts
salazarm: We always send this first so it will always be defined.

js_modules/dagit/packages/core/src/runs/PipelineRunLogsSubscription.tsx
```diff
@@ -0,0 +1,197 @@
...
+      }
+    }
+    ... on LogsCapturedEvent {
+      logKey
+      stepKeys
+      pid
+    }
+  }
+  ${METADATA_ENTRY_FRAGMENT}
+`;
```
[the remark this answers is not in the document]
salazarm: They exist elsewhere but I redefined them here so that they'd be packaged separately with the worker. Otherwise the worker bundle would likely include unused javascript from having a bigger bundle. I'll fix it like the original PR though so theres only 1 definition

js_modules/dagit/packages/core/src/workers/util.ts
salazarm: This is using native APIs underneath, it's just chunking to avoid going over the max number of arguments allowed for a function (Since we call `String.fromCharCode.apply` on the buffer, the number of arguments is the size of the array).

js_modules/dagit/packages/core/src/workers/runLogs/runLogs.worker.ts
salazarm: the postMessage api type doesnt match the spec 🤷🏾

js_modules/dagit/packages/core/src/workers/runLogs/runLogs.tsx
salazarm: yes, theyre different objects being subscribed to

js_modules/dagit/packages/core/src/runs/usePipelineRunLogsSubscription.tsx
salazarm: I think its okay to fail loudly here? This is within an event handler so its isolated

js_modules/dagit/packages/core/src/runs/usePipelineRunLogsSubscription.tsx
salazarm: will do

salazarm:
@hellendag I removed the branching based off the URL and made the worker the default since it performs better. Also thanks for the detailed feedback! I know the PR was really rough I was mostly trying to get something quick out there for feedback and testing

js_modules/dagit/packages/core/src/workers/runLogs/runLogs.worker.ts
hellendag: 🤦

alangenfeld (commented):
copying this over from slack DMs - context was investigating the mysterious overall speed-up of the worker version:

so i was playing around with this change
```
        if (!logs.hasMorePastEvents) {
          requestAnimationFrame(throttledSetNodes);
        }
```
just against master, and noticed in the react profiler that even though we try to don’t set nodes, the apollo subscription update still causes RunWithData to re-render every time we receive a websocket payload

Screen Shot 2022-09-29 at 3 28 56 PM

which upon googling appears to be a common source of performance issues

https://stackoverflow.com/questions/61876931/how-to-prevent-re-rendering-with-usesubscription
https://medium.com/@seniv/improve-performance-of-your-react-apollo-application-440692e37026

PR #9796, opened 2022-09-26
Add clearer logging when daemon is in the process of shutting down

gibsondan:
Summary:
Came up while debugging last week - we don't log anything on shutdown until the join finishes which can be quite slow. Log something as soon as we start shutting down instead.


alangenfeld (approved):
<3

PR #9797, opened 2022-09-26
[dagit] Add top-level polling to virtualized table pages

hellendag:
### Summary & Motivation

Add polling and refresh to virtualized Workspace list pages.

image

View each list page for a repo, verify that the polling/refresh behavior is correct.


js_modules/dagit/packages/core/src/workspace/WorkspaceAssetsRoot.tsx
```diff
@@ -27,6 +30,7 @@ export const WorkspaceAssetsRoot = ({repoAddress}: {repoAddress: RepoAddress}) =
     },
   );
   const {data, loading} = queryResultOverview;
+  const refreshState = useQueryRefreshAtInterval(queryResultOverview, POLL_INTERVAL);
```
bengotow, line 33: If you want, I think we have the standard 15s default defined inside `QueryRefresh.tsx` and we could import that one. I moved it there during a refactor because it seemed like it should be the same in most cases.

bengotow (approved):
This looks good! Just left one comment about maybe using the other 15s constant 🧑‍🤝‍🧑

PR #9803, opened 2022-09-27
Use /bin/bash -c in EcsRunLauncher command

johannkm:


alangenfeld (changes requested):
we found a different root cause for the associated issue right?

PR #9804, opened 2022-09-27
Enable Buildkite test analytics

jmsanders:
https://buildkite.com/docs/test-analytics/python-collectors

When run locally without an API token, this will no-op.

When run on buildkite, with an API token (and with the proper envvars
exposed), this will collect tests results and send them to Buildkite so
we can observe patterns in our test suite.


.buildkite/dagster-buildkite/tox.ini
```diff
@@ -6,7 +6,7 @@ skipsdist = true
 usedevelop = true
 setenv =
   VIRTUALENV_PIP=21.3.1
-passenv = CI_* COVERALLS_REPO_TOKEN BUILDKITE SNOWFLAKE_ACCOUNT SNOWFLAKE_USER SNOWFLAKE_PASSWORD
+passenv = CI_* COVERALLS_REPO_TOKEN BUILDKITE* SNOWFLAKE_ACCOUNT SNOWFLAKE_USER SNOWFLAKE_PASSWORD
```
jmsanders, line 9: Alternatively, we could allow only the envvars needed for the test collector to properly label things:

https://buildkite.com/docs/test-analytics/ci-environments#buildkite

Instead of every BUILDKITE prefixed envvar:

https://buildkite.com/docs/pipelines/environment-variables

I think the complete list is pretty innocuous though.

alangenfeld (approved):

PR #9805, opened 2022-09-27
[scheduler] dont clean auto running states for error locations

alangenfeld:
This can cause us to skip ticks for schedules when there is a transient error loading the workspace. 

WIP

### DIFF
--- a/python_modules/dagster/dagster/_scheduler/scheduler.py
+++ b/python_modules/dagster/dagster/_scheduler/scheduler.py
@@ -115,8 +115,8 @@ def execute_scheduler_iteration_loop(
 
 
 def launch_scheduled_runs(
-    instance,
-    workspace,
+    instance: DagsterInstance,
+    workspace: IWorkspace,
     logger,
     end_datetime_utc,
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_scheduler/scheduler.py
```diff
@@ -165,9 +168,19 @@ def launch_scheduled_runs(
...
         )
+        # don't clean up auto running state if its location is an error state
+        if location_name in error_locations:
+            break
+
+        if log_verbose_checks:
+            logger.info(
+                f"Removing state for automatically running schedule {state.instigator_name} "
+                f"that is no longer present in {location_name}"
+            )
```
gibsondan, line 182: I don't think we need to wrap this one in log_verbose_checks since its relatively rare? The logging stuff is admittedly pretty funky but i think this was designed to guard against logging the same thing over and over again when nothing really happened, this falls under 'something happening' i think and i'd want it to always show up in the logs if we do it?

gibsondan (approved):

PR #9814, opened 2022-09-28
Flag to set cursor to latest materializations on build_multi_asset_sensor_context

clairelin135:
This PR adds a flag to set the multi-asset sensor context to the latest materialization of each monitored asset. This enables more ergonomic testing.

For example, you might monitor asset A and asset B, kicking off asset C when both are materialized. With this flow, you can materialize asset A, set the sensor cursor to the latest materialization for A, and then kick off B. On the next tick, you can evaluate that a run of C is kicked off.

We previously discussed setting the cursor by accepting an `ExecuteInProcessResult` object, but this actually doesn't work as the events from an execute in process result are in the run-sharded event storage, whereas we evaluate asset events in the index event log storage. Given an event in run-sharded event storage, there's no built in way to determine the corresponding storage ID of the event in the index event log storage. I think this implementation feels cleaner and still achieves the same goals.


sryza:
Interesting - to help give a picture of what it would look like to use this, would you be able to post a PR showing off what it would look like on these tests? https://github.com/dagster-io/dagster/blob/master/examples/project_fully_featured/project_fully_featured_tests/test_sensors/test_hn_tables_updated_sensor.py

clairelin135:
@sryza Here's how this functionality looks on the hacker news assets: https://github.com/dagster-io/dagster/pull/9814/files

sryza:
@clairelin135 that link seems to link to this PR. Did you mean for it to point somewhere else?

clairelin135:
@sryza Ah, this is the actual link: https://github.com/dagster-io/dagster/pull/9842

sryza:
Having seen the how this looks on the hacker news assets, this looks good to me.  My main piece of feedback is that `set_cursor_to_latest_materializations` is a bit of a weird name, because it sounds like a method.  Maybe `cursor_from_latest_materializations`?

clairelin135:
@sryza thanks for giving this a look! `cursor_from_latest_materializations` sounds great to me. I will wait to update this PR until https://github.com/dagster-io/dagster/pull/9903 is nearing completion since we'll have incoming changes to the cursor structure

sryza (approved):

PR #9815, opened 2022-09-28
Specify platform to build in deploy_ecs example

gibsondan:
Summary:
So that users building on M1 macs will still build with the right architecture.

Test Plan:
I don't have an M1, but verify that the deploy_ecs example still deploys - check with somebody who has an M1 to try as well.


jmsanders (approved):
I also don't have an M1 so it'd be great if somebody who has one could try this - maybe @Ramshackle-Jamathon? - but this will hopefully prevent a lot of user confusion! Thanks!

PR #9818, opened 2022-09-28
handle None outputs in snowflake IO manager

sryza:

### DIFF
--- a/python_modules/libraries/dagster-snowflake/dagster_snowflake/db_io_manager.py
+++ b/python_modules/libraries/dagster-snowflake/dagster_snowflake/db_io_manager.py
@@ -84,17 +84,27 @@ class DbIOManager(IOManager):
 
     def handle_output(self, context: OutputContext, obj: object) -> None:
         table_slice = self._get_table_slice(context, context)
-        self._db_client.delete_table_slice(context, table_slice)
 
-        obj_type = type(obj)
-        check.invariant(
-            obj_type in self._handlers_by_type,
-            f"DbIOManager does not have a handler for type '{obj_type}'. Has handlers "
-            f"for types '{', '.join([str(handler_type) for handler_type in self._handlers_by_type.keys()])}'",
-        )
-        handler_metadata = (
-            self._handlers_by_type[obj_type].handle_output(context, table_slice, obj) or {}
-        )
+        if obj is not None:
+            obj_type = type(obj)
+            check.invariant(
+                obj_type in self._handlers_by_type,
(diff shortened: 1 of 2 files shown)


python_modules/libraries/dagster-snowflake/dagster_snowflake/db_io_manager.py
```diff
@@ -84,17 +84,27 @@ def __init__(self, type_handlers: Sequence[DbTypeHandler], db_client: DbClient):
...
+            )
+        else:
+            check.invariant(
+                context.dagster_type.is_nothing,
+                "Unexpected 'None' output value. If a 'None' value is intentional, set the output type to None.",
+            )
+            # if obj is None, assume that I/O was handled in the op body
+            handler_metadata = {}
 
         context.add_output_metadata(
```
jamiedemaria, line 109: i'm not 100% sure what this output metadata is used for but does it make sense to add the metadata about the query when the output is None?

jamiedemaria (approved):
looks good! one question about metadata, but otherwise 👍

sryza:
@OwenKephart - thoughts on whether we should instead have some dbt-specific handling?

python_modules/libraries/dagster-snowflake/dagster_snowflake/db_io_manager.py
sryza: This gives users a statement that they can copy/paste into the Snowflake interface to inspect the contents of their table.  How the table is created (by the I/O manager vs. in the op body) shouldn't affect the relevance of this query.

jamiedemaria: makes sense, thx for the clarification

jamiedemaria:
I coincidentally just started working on dbt tutorial stuff and am using the snowflake io manager becuase i was able to get it set up the quickest for dev work. I pulled the changes from this PR into my branch because I immediately ran into issue with the dbt_run op returning None outputs. With the changes brought in, the `check.invariant` call that confirms that the type annotation is also None fails. I assume that means that the op we generate for dbt run needs to have `None` type annotations added to it (or the invariant check could accept Any as a valid type?)

sryza:
@jamiedemaria ah yeah I made a separate PR for that: https://github.com/dagster-io/dagster/pull/9822

OwenKephart:
What would dbt-specific handling look like for this?

sryza:
> What would dbt-specific handling look like for this?

Maybe:
- The dbt op returns value of a `DbtOutput` type
- `build_snowflake_io_manager([

PR #9825, opened 2022-09-28
Add metadata hooks for making grpc client calls

gibsondan:
Summary:
One could imagine a multiplexing gRPC server passign through metadata for routing information to say which server the multiplex should route to. Also provide a utlity _get_response method that can be used to easily pass through requests.

### DIFF
--- a/python_modules/dagster/dagster/_grpc/client.py
+++ b/python_modules/dagster/dagster/_grpc/client.py
@@ -3,7 +3,7 @@ import subprocess
 import sys
 import warnings
 from contextlib import contextmanager
-from typing import Iterator, Optional
+from typing import Iterator, List, Optional, Tuple
 
 import grpc
 from grpc_health.v1 import health_pb2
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster/_grpc/client.py
```diff
@@ -105,21 +105,56 @@ def _channel(self):
...
+    def _get_response(
+        self,
+        method,
+        request,
+        metadata: Optional[List[Tuple[str, str]]] = None,
+        timeout=DEFAULT_GRPC_TIMEOUT,
+    ):
+        with self._channel() as channel:
+            stub = DagsterApiStub(channel)
+            return getattr(stub, method)(request, metadata=metadata, timeout=timeout)
```
alangenfeld, line 117: do you have any idea why we follow this pattern

alangenfeld (approved):

python_modules/dagster/dagster/_grpc/client.py
gibsondan: only max knows

Issue #7891, opened 2022-05-13
Unpin alembic


TomekTrzeciak:
## Use Case

After #4665 alembic got pinned below <1.7.0 and this causes problems with some other packages that pin it above that version.

---
#### Message from the maintainers:
Excited about this feature? Give it a :thumbsup:. We factor engagement into prioritization.

jamiedemaria:
We'll triage this and get it into our backlog!

laurentS:
As a bonus, 1.7.0 [adds typing support](https://github.com/sqlalchemy/alembic/commit/1977047a11124221461ad8daee502bc4c74d5f6c), which is nice for completion and stuff when writing migrations.

johnraz:
@jamiedemaria any updates on this?

This seems to have been implemented as a quick fix or kind of workaround here https://github.com/dagster-io/dagster/pull/4665 

Then this issue: https://github.com/dagster-io/dagster/issues/4667 was created to address it properly and for some reason got closed a year ago.

I'm willing to contribute and submit a patch but I'd like to have a confirmation that it's worth my time and there is a chance it would be accepted 🙂

jamiedemaria:
@gibsondan do you know what the status of alembic is?

jamesseth:
I must say this is just sad, Such an awesome orchestrator; I tried to intro it into a large global company... and boom flop fail... alembic versioning issues that cannot be resolved with out reworking most of their code base.  Or creating additional infrastructure to support just Dagster due to the alembic issue. 

Please fix it, Apache Airflow / Argo workflows will need to be the way the go for now I guess.

johnraz:
@jamesseth Just my 2 cents but being negative like this is not going to help people get engaged in what you are facing, just saying ✌️

jamesseth:
100% True @johnraz  apologies, on a side note i might have found a workaround...  I just need to test with some other alembic version to be sure.    again thx @johnraz  for calling me out we need it at times.

johnraz:
@jamesseth very welcome, we all have our grumpy moment , thanks for taking it on the constructive side as well 🚀

gibsondan:
@jamesseth appreciate the introspection, the underlying substance of the feedback about the pin being onerous is entirely valid though.

@johnraz I took a quick look and I think we can actually unpin this pretty easily, will get that out shortly.

johnraz:
@gibsondan  thanks again for the swift response and fix 👍

PR #9830, opened 2022-09-29
Make alembic pin much less restrictive

gibsondan:
Summary:
Took a look at the test that was failing in 1.7.0 and it looks like the underlying issue was fixed in the very next version, and the latest version passes tests too. Based on that, makes the pin less restrictive.

Resolves https://github.com/dagster-io/dagster/issues/7891.

Test Plan:
BK, verify that a migration from 0.15.x to 1.0.x still works with the latest version


alangenfeld (approved):
nice thanks for checking up on this - I feel like i want to do `!=` pins more often out the gate but its hard to discern temporary issues from fundamental breaks

Issue #9028, opened 2022-07-25
Allow partitions selection with cursor in the asset partitions view (like in the jobs partitions view)


danielgafni:
### What's the use case?

Currently to materialize an old asset partition you have to hover the cursor over the partition, view the date (partition key) and then type it manually. It would be nice to reuse the same UI elements the Job view has where you can just click on the desired partitions.

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

PR #9832, opened 2022-09-29
[dagit] Use a consistent partition selection UI for op / asset backfills

bengotow:
### Summary & Motivation

This PR fixes two issues with the "Materialize" button used to launch partitioned asset runs:

- This UI was slightly different from the version used by op jobs (and less nice, @prha's iteration on top of it added more features!) I refactored them to use the same component so you can click+drag to select partitions.  https://github.com/dagster-io/dagster/issues/9028

- You could attempt to submit the dialog with no partitions selected

Before:
Screen Shot 2022-09-23 at 4 00 00 PM

After:
Screen Shot 2022-09-23 at 3 59 14 PM

### DIFF
--- a/js_modules/dagit/packages/core/src/partitions/PartitionStatus.tsx
+++ b/js_modules/dagit/packages/core/src/partitions/PartitionStatus.tsx
@@ -13,9 +13,34 @@ type SelectionRange = {
 
 const MIN_SPAN_WIDTH = 8;
 
+export enum PartitionState {
+  MISSING = 'missing',
+  SUCCESS = 'success',
+  FAILURE = 'failure',
+  QUEUED = 'queued',
+  STARTED = 'started',
+}
+
+export const runStatusToPartitionState = (runStatus: RunStatus | null) => {
+  switch (runStatus) {
+    case RunStatus.CANCELED:
+    case RunStatus.CANCELING:
+    case RunStatus.FAILURE:
+      return PartitionState.FAILURE;
+    case RunStatus.STARTED:
+      return PartitionState.STARTED;
+    case RunStatus.SUCCESS:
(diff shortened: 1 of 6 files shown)


js_modules/dagit/packages/core/src/partitions/PartitionStatus.tsx
```diff
@@ -13,9 +13,34 @@ type SelectionRange = {
 
 const MIN_SPAN_WIDTH = 8;
 
+export enum PartitionState {
```
bengotow, line 16: Previously, this component used RunStatus and did a mapping to colors. I added this intermediate enum so that run states and asset states can be coerced into the same type and passed to this component.  Right now the asset version only uses missing and success, but with the new work Sean is doing I think we'll have failing / "out of date" soon.

js_modules/dagit/packages/core/src/partitions/PartitionStatus.tsx
```diff
@@ -164,9 +190,14 @@ export const PartitionStatus: React.FC<{
...
-              minWidth: s.status ? 2 : undefined,
+              minWidth: s.status && s.status !== PartitionState.MISSING ? 2 : undefined,
               position: 'absolute',
-              zIndex: s.startIdx === 0 || s.endIdx === highestIndex ? 3 : s.status ? 2 : 1, //End-caps, then statuses, then missing
+              zIndex:
+                s.startIdx === 0 || s.endIdx === highestIndex
+                  ? 3
+                  : s.status && s.status !== PartitionState.MISSING
+                  ? 2
+                  : 1, //End-caps, then statuses, then missing
```
hellendag, line 200: whoa

hellendag (approved):

PR #9845, opened 2022-09-30
fix api docs for python graphql client

gibsondan:
Summary:
We link to these methods in the docs but the links do not work. Now, they should.

Test Plan:
Build API docs locally, should see the linked methods on https://docs.dagster.io/concepts/dagit/graphql-client#overview

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/client/client.py
+++ b/python_modules/dagster-graphql/dagster_graphql/client/client.py
@@ -7,8 +7,9 @@ from gql.transport import Transport
 from gql.transport.requests import RequestsHTTPTransport
 
 import dagster._check as check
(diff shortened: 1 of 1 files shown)


python_modules/dagster-graphql/dagster_graphql/client/client.py
```diff
@@ -321,6 +323,7 @@ def submit_job_execution(
             is_using_job_op_graph_apis=True,
         )
 
+    @public
     def get_run_status(self, run_id: str) -> PipelineRunStatus:
```
alangenfeld, line 327: should we update this to `DagsterRunStatus` ?

alangenfeld (approved):

PR #9846, opened 2022-09-30
[dagit] Use a filterable Select for compute logs toolbar

hellendag:
### Summary & Motivation

Use a filterable Select instead of the current non-filterable one in the compute log toolbar step selection control.

image

Keyboard behavior with this thing seems pretty broken, so I'll need to figure out what's going on there. It's already broken, though, so this won't make it worse.

View a run, open compute log toolbar. Search for a step, verify that it filters and selects correctly.

### DIFF
--- a/js_modules/dagit/packages/core/src/runs/LogsToolbar.tsx
+++ b/js_modules/dagit/packages/core/src/runs/LogsToolbar.tsx
@@ -148,6 +148,7 @@ const ComputeLogToolbar = ({
   const logCaptureSteps =
     metadata.logCaptureSteps || extractLogCaptureStepsFromLegacySteps(Object.keys(metadata.steps));
   const isValidStepSelection = computeLogKey && logCaptureSteps[computeLogKey];
+
   const logKeyText = (logKey?: string) => {
     if (!logKey || !logCaptureSteps[logKey]) {
       return null;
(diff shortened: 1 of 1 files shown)


hellendag:
* master
  * **PR #9846** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9846" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9846?utm_source=stack-comment).

salazarm (approved):

js_modules/dagit/packages/core/src/runs/LogsToolbar.tsx
```diff
@@ -171,6 +172,9 @@ const ComputeLogToolbar = ({
         <Select
           disabled={!steps.length}
           items={Object.keys(logCaptureSteps)}
+          itemPredicate={(query, item) =>
+            item.toLocaleLowerCase().includes(query.toLocaleLowerCase())
```
salazarm, line 176: Should we use a Fuze?

hellendag: I think let's keep it simple for now, and maybe come back to it if it looks like fuzzy search would be a big win here.

PR #9855, opened 2022-09-30
[Performance] Stop recreating PipelineRunLogsSubscription on every render 

salazarm:
### Summary & Motivation

This PR fixes an issue arising due to the variables of our `useSubscription` changing on every render. By changing the variables we cause apollo-graphql to recreate the subscription from scratch on every render.  This is a problem because the python server fetches chunks for the subscription in a while loop and queues up all of the results but the client closes the subscription after it receives just one message. The result is that we end up overfetching and throwing away the work.

### Test Plan

Left side: Passing the cursor (start/stop multiple times) 
Right side: Without passing cursor (no start/stop multiple times)
image

### DIFF
--- a/js_modules/dagit/packages/core/src/runs/LogsProvider.tsx
+++ b/js_modules/dagit/packages/core/src/runs/LogsProvider.tsx
@@ -244,9 +244,10 @@ const SubscriptionComponent = ({
     return {
       runId,
       cursor,
-      lostWebsocket,
     };
     // Don't add cursor to memo otherwise subscription gets recreated
+    // Add lostWebSocket to dependency so that we record the cursor we're at
+    // for when the status comes back
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [runId, lostWebsocket]);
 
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/core/src/runs/LogsProvider.tsx
```diff
@@ -237,11 +237,25 @@ const SubscriptionComponent = ({
...
+  const variables = React.useMemo(() => {
+    return {
+      runId,
+      cursor,
+    };
+    // Don't add cursor to memo otherwise subscription gets recreated
+    // Add lostWebSocket to dependency so that we record the cursor we're at
+    // for when the status comes back
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [runId, lostWebsocket]);
```
hellendag, line 252: Seems like you don't need to track `runId` here, since it shouldn't update.

Can this websocket observation behavior go in the `useLogsProviderWithSubscription`? Since that's where the logic lives to track the cursor value, it seems like that could also be the place where you could decide whether or not to set a new cursor to pass to the `SubscriptionComponent` and update its variables. Then maybe you don't need to hack around the hook rule.

What do you think?

salazarm: > Seems like you don't need to track runId here, since it shouldn't update.

From this component's point of view since its a prop it could potentially change though so I don't think it would be wise to use knowledge that it won't update that isn't codified into this component

hellendag (approved):

PR #9856, opened 2022-09-30
MultiAssetSensorEvaluationContext.latest_materialization_by_asset_per_partition

clairelin135:
Adds a `latest_materialization_by_asset_per_partition` which finds all materializations after the cursor for each monitored asset. Builds a mapping of partition key to a list of `Tuple(AssetKey, EventLogEntry)` entries representing the materializations. Only adds the most recent materialization for a given asset partition to the mapping.

Also amends the `advance_cursor` function to only advance the cursor to the provided materialization if the provided materialization has a greater storage ID than the existing entry.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/sensor_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/sensor_definition.py
@@ -483,7 +483,18 @@ class MultiAssetSensorEvaluationContext(SensorEvaluationContext):
         self,
     ) -> Mapping[str, List[Tuple[AssetKey, "EventLogRecord"]]]:
         """
-        If all monitored assets are partitioned,
+        Finds the most recent materialization for each partition after the cursor for each asset
+        monitored by the sensor. Aggregates all materializations into a mapping of partition key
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -478,6 +478,45 @@ def my_sensor(context):
 
         return materialization_by_partition
 
+    @public
+    def latest_materialization_records_per_asset_by_partition(
+        self,
+    ) -> Mapping[str, List[Tuple[AssetKey, "EventLogRecord"]]]:
```
sryza, line 484: Thoughts on making this `-> Mapping[str, Mapping[AssetKey, EventLogRecord]]`? I think it's very rare that anyone would care about partition materializations before the latest one per partition. Sorry if I misled with my comment on the other PR.

clairelin135: No problem, this makes sense, I've updated the method to have the new function signature

sryza (approved):
Nice!

PR #9857, opened 2022-10-01
Fix bug with multiprocess execution, dynamic outputs, and retries

dpeng817:
### Summary & Motivation
Fixes a bug with out-of-process execution and dynamic execution where, if a dynamic step fails during initialization of resources, subsequent retries of the job will ignore the dynamic steps that did not execute. Now, resource initialization failures during step execution are treated as failures, and the re-execution machinery will pick up steps that failed during resource initialization.

### DIFF
--- a/python_modules/dagster/dagster/_core/executor/step_delegating/step_delegating_executor.py
+++ b/python_modules/dagster/dagster/_core/executor/step_delegating/step_delegating_executor.py
@@ -218,7 +218,11 @@ class StepDelegatingExecutor(Executor):
                         active_execution.verify_complete(plan_context, dagster_event.step_key)
                     else:
                         active_execution.handle_event(dagster_event)
-                        if dagster_event.is_step_success or dagster_event.is_step_failure:
+                        if (
+                            dagster_event.is_step_success
+                            or dagster_event.is_step_failure
+                            or dagster_event.is_resource_init_failure
+                        ):
                             assert isinstance(dagster_event.step_key, str)
                             del running_steps[dagster_event.step_key]
                             active_execution.verify_complete(plan_context, dagster_event.step_key)
--- a/python_modules/dagster/dagster/_core/events/__init__.py
+++ b/python_modules/dagster/dagster/_core/events/__init__.py
@@ -559,6 +559,11 @@ class DagsterEvent(
 
     @public  # type: ignore
     @property
+    def is_resource_init_failure(self) -> bool:
+        return self.event_type == DagsterEventType.RESOURCE_INIT_FAILURE
+
+    @public  # type: ignore
+    @property
     def is_step_skipped(self) -> bool:
         return self.event_type == DagsterEventType.STEP_SKIPPED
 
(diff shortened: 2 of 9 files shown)


dpeng817:
* master
  * **PR #9857** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9857" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9857?utm_source=stack-comment).

python_modules/dagster/dagster/_core/executor/step_delegating/step_delegating_executor.py
```diff
@@ -222,6 +222,21 @@ def execute(self, plan_context: PlanOrchestrationContext, execution_plan: Execut
...
+                                step_failure_data=StepFailureData(
+                                    error=dagster_event.event_specific_data.error,  # type: ignore
+                                    user_failure_data=None,
+                                ),
+                            )
+                            yield step_failure_event
+                            active_execution.handle_event(step_failure_event)
+                            assert isinstance(dagster_event.step_key, str)
+                            del running_steps[dagster_event.step_key]
+                            active_execution.verify_complete(plan_context, dagster_event.step_key)
```
alangenfeld, line 239: something seems off if we have to dupe the logic in active exec here as well

python_modules/dagster/dagster/_core/executor/multiprocess.py
```diff
@@ -213,6 +213,18 @@ def execute(self, plan_context, execution_plan):
...
+                                        step_context=plan_context.for_step(
+                                            active_execution.get_step_by_key(key)
+                                        ),
+                                        step_failure_data=StepFailureData(
+                                            error=event_or_none.event_specific_data.error,
+                                            user_failure_data=None,
+                                        ),
+                                    )
+                                    yield step_failure_event
+                                    active_execution.handle_event(step_failure_event)
```
alangenfeld, line 227: so the choice here is to emit step failure without step start for init failure? Did you consider other approaches? 

If we pursue this approach, should we respect the retry policy?

python_modules/dagster/dagster/_core/executor/step_delegating/step_delegating_executor.py
alangenfeld: oops misread that above was multiproc not active ex, but something still fishy about this

dpeng817: Agreed that it's not that clean. This seemed like the easiest solution; wanted to get takes on this approach first but another I have in mind would be having step start occur immediately after opening the process instead of waiting until initialization occurs first (I don't think that has any troublesome implications? And might just be cleaner).

python_modules/dagster/dagster/_core/executor/multiprocess.py
dpeng817: If we took this approach, then I would think no. The lack of step start is a good point, and I wonder if we could better solve this by just moving step start to occur before initialization occurs.

alangenfeld (approved):
nice i think this looks good

PR #9860, opened 2022-10-02
Reference correct method in "loading asset values outside of Dagster runs" example

paiforsyth:
### Summary & Motivation
In the section "Loading asset values outside of Dagster runs" of the concepts documentation for software-defined assets, two  examples of loading assets are presented, one of which uses `load_asset_value`, and one of which uses `get_asset_value_loader`.  However, the text describing both examples cites `load_asset_value`, which seems like a mistake.


sryza (approved):
Thanks for the fix @paiforsyth! Going to merge this.

PR #9863, opened 2022-10-03
Fix link in tutorial starting page

erinkcochran87:
This corrects a link in the tutorial starting page.


sryza (approved):
Thanks for fixing!

PR #9866, opened 2022-10-03
[docs] - Fix formatting in note

erinkcochran87:
This PR fixes some formatting in a Note component, to correctly use `pre` instead of `code`


erinkcochran87:
Scrapping this - this was done in error, and was actually caused by #9881 and fixed in #9881

PR #9885, opened 2022-10-05
Add threading to scheduler daemon

dpeng817:
### Summary & Motivation
Adds a threadpool to the scheduler daemon for launching off ticks of schedules. This was implemented mostly by cargo cult-ing the sensor daemon's thread pool, and making some modifications for schedule specific use cases.

Parameterized the scheduler test suite based on executor, so each test is run in both synchronous and multithreaded mode.

### DIFF
--- a/python_modules/dagster/dagster/_scheduler/scheduler.py
+++ b/python_modules/dagster/dagster/_scheduler/scheduler.py
@@ -2,8 +2,11 @@ import datetime
 import logging
 import os
 import sys
+import threading
 import time
 from collections import defaultdict
+from concurrent.futures import Future, ThreadPoolExecutor
+from contextlib import ExitStack
 from typing import Dict, List, Optional, cast
 
 import pendulum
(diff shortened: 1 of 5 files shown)


dpeng817:
* master
  * **PR #9885** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9885" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9885?utm_source=stack-comment).

python_modules/dagster/dagster/_scheduler/scheduler.py
```diff
@@ -281,6 +343,188 @@ def launch_scheduled_runs(
...
+            max_catchup_runs,
+            max_tick_retries,
+            tick_retention_settings,
+            schedule_debug_crash_flags,
+            log_verbose_checks,
+        )
+    )
+
+
+def _process_tick_generator(
```
alangenfeld, line 378: i think these names are worse than the previous ones - the code in question doesn't line up with the sensors behavior quite the same

python_modules/dagster/dagster/_scheduler/scheduler.py
```diff
@@ -281,6 +343,188 @@ def launch_scheduled_runs(
...
+                    error_data = serializable_error_info_from_exc_info(sys.exc_info())
+                    tick_context.update_state(
+                        TickStatus.FAILURE,
+                        error=error_data,
+                        failure_count=tick_context.failure_count + 1,
+                    )
+                    yield error_data
+
+
 def launch_scheduled_runs_for_schedule(
```
alangenfeld, line 284: still keeping all the code you duped - ideally try to minimize thrash to preserve blame

alangenfeld (changes requested):
back to you to clean-up the diff, mainly the duped code

python_modules/dagster/dagster/_scheduler/scheduler.py
```diff
@@ -416,7 +513,7 @@ def launch_scheduled_runs_for_schedule(
                             # or the schedule is turned off
                             failure_count=tick_context.failure_count,
                         )
-                        raise  # Raise the wrapped DagsterSchedulerError exception
+                        yield error_data
```
alangenfeld, line 516: i assume if the tests all pass this change is fine but its a little spooky to me - was this just cargo cult or did you have intention behind it?

alangenfeld (approved):

python_modules/dagster/dagster/_scheduler/scheduler.py
dpeng817: I had intention - certain tests were failing because errors were not propogating upward properly in the multithreaded case if they were just raised instead of the error data yielded - for example `test_bad_env_fn_with_retries`. The one thing that could be spooky is if more code got added later - after this except block, that code could potentially run. So maybe we want to return post-yield.

PR #9888, opened 2022-10-05
[dagit] Add a new GraphQL subscription for asset events, use for realtime asset DAG

bengotow:
### Summary & Motivation

Video here: https://www.loom.com/share/7278ffd529e04c5bb734de71a068a816

The goal of this work is to enable a more "realtime" Asset Graph view in Dagit without an obscene amount of polling.

I originally aimed to make the asset graph page hit `assetLatestInfo` and then subscribe to logs for all the inProgressRunIds it saw in the response. When certain logs went by, it would know to update the state of the asset graph. This seemed wasteful because 1) there could be many runs, 2) we don't care about past logs or non-asset logs (lots of ignored data being sent to the browser)

Instead, this PR creates a new GraphQL subscription type that follows the structure of the PollingEventWatcher used for streaming run logs, except that it watches for specific asset event types across all runs. (Using get_logs_for_all_runs_by_log_id, which unfortunately doesn't seem to be used much elsewhere).

Right now, when we see an asset materialization event go by, we trigger a refresh of the live query on the Asset Graph (at most every 2s). I would *love* to perform client-side state updates rather than re-requesting all the live data, but I tried this and it seems problematic. It's difficult and kinda fragile trying to update the data using a single materialization or observation event. Especially "upstream changed" (which cascades) and step failure.

Would love feedback about this. The end result is that we poll much LESS when nothing is happening and much MORE when assets are materializing, which feels much nicer, but will also create more spikey load on the system.

TODO:

- I implemented this new asset event watcher for Mysql and Postgres OSS event storage, but not the cloud Postgres storage, which seems to use a cool event driven mechanism I'm not familiar with instead of polling.


sryza:
This looks like an enormous user experience improvement! With this, I could seriously imagine using the asset graph for local development.

@prha - what do you think about the storage changes here?

python_modules/dagster/dagster/_core/storage/event_log/base.py
```diff
@@ -214,6 +214,15 @@ def watch(self, run_id: str, cursor: str, callback: Callable):
     def end_watch(self, run_id: str, handler: Callable):
         """Call this method to stop watching."""
 
+    def supports_watch_asset_events(self):
```
prha, line 217: do we need this?  It seems like the polling watcher would work for all sql based implementations?

python_modules/dagster-graphql/dagster_graphql/implementation/asset_subscription.py
```diff
@@ -0,0 +1,41 @@
...
+            self.observer.dispose()
+        self.observer = None
+
+    def handle_new_events(self, new_events: List[EventLogEntry]):
+        filtered = list()
+
+        for event in new_events:
+            d_event = event.dagster_event
+            step_hash = f"{d_event.pipeline_name}-{d_event.step_key}"
+            asset_hash = d_event.asset_key.to_string() if d_event.asset_key else None
```
prha, line 36: We should be able to assert that the event always has an asset key, right?

prha (commented):
I think this approach looks good.

python_modules/dagster/dagster/_core/storage/event_log/polling_asset_event_watcher.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def __init__(self, event_log_storage: SqlEventLogStorage):
+        super(SqlPollingAssetEventWatcherThread, self).__init__()
+        self._event_log_storage = check.inst_param(
+            event_log_storage, "event_log_storage", SqlEventLogStorage
+        )
+        self._callback_fn_list_lock: threading.Lock = threading.Lock()
+        self._callback_fn_list: List[Callable] = []
+        self._should_thread_exit = threading.Event()
+        self.name = "mysql-event-watch-assets"
```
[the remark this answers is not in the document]
bengotow: Ahh yep will fix this, good catch

python_modules/dagster/dagster/_core/storage/event_log/base.py
bengotow: Hmm this might be true - I was mostly planning of leaving this out of the LegacyEventLogStorage and InMemoryEventLogStorage, but maybe I can do a basic implementation for those and we'll have complete coverage.

The polling watcher does seem like it would work for all the SQL implementations, but it looks like each concrete class implements these methods and builds an instance of the watcher itself. It looks like the Postgres event log watcher uses a different implementation built on a postgres-specific notification system, so that may be why? I'll check with @gibsondan and see since it looks like he wrote the Postgres one!

python_modules/dagster-graphql/dagster_graphql/implementation/asset_subscription.py
bengotow: I think this is related to the comment above - the "match asset keys to op names OR asset names" complexity here is necessary because I wanted to catch step_start and step_failed events to update the progress spinners in the UI.

prha (changes requested):
Echoing our offline discussion, the default SQLite implementation is run-sharded, with only the asset events getting mirrored into the index-connection (from the run-specific connections).  This means the step start / step failed events will never get returned by this subscription.

We are probably better off limiting to subscribing to asset-only events, and then using the asset planned event to subsequently subscribing to all the run-specific events.

bengotow:
After playing around with this more and realizing that we need to subscribe to run logs to see asset materialization _failures_, I made a simpler implementation that 1) subscribes to run logs to see observation, materialization, and step start / failure events, and 2) refreshes manually after a run is launched to trigger the initial subscription.

I think this handles all but one case without any significant backend changes or new subscriptions. The one edge case is starting a run from one tab and then switching tabs to the asset graph. In that case, it takes the default polling interval before the view realizes a run is happening, and then everything updates rapidly.

I think it's worth exploring the simpler solution first to see if other parts of this work are actually workable. (I'm separately not 100% sure it's safe to be refreshing the live data so often). Going to close this in favor of #10028

PR #9894, opened 2022-10-05
View Notebook button opens notebook as link if it's a noteable notebook

jamiedemaria:
### Summary & Motivation
My frontend/JS knowledge is pretty poor so lmk if i'm doing anything silly here!

We'd like to make the View Notebook button open the linked notebook if the notebook is a noteable notebook. Ideally, I think it'd be nice to remove the check on if it's a noteable notebook and instead check if the `notebook_path` is a url (that way we could automatically support other hosted notebooks).  I'm not sure what the preferred way to see if a string is a url in JS. I saw some stuff about putting a try/catch around `URL(my_url_string)`, so if that's sufficient that'd be great!

it works in my example i'm running - I can write a test somewhere if you point me to the right file! i'm pretty lost in the `js_modules` folder


js_modules/dagit/packages/core/src/plugins/ipynb.tsx
```diff
@@ -16,42 +16,56 @@ export const SidebarComponent: React.FC<IPluginSidebarProps> = (props) => {
...
-        </DialogFooter>
-      </Dialog>
-    </div>
-  );
+  if (notebookKind && notebookKind.value === 'noteable') {
+    return (
+      <div>
+        <Button
+          icon={<Icon name="content_copy" />}
+          onClick={() => window.open(notebookPath.value, '_blank')}
```
hellendag, line 31: I'm not familiar with the different types of notebook possibilities here. Is it always preferable to open Noteable notebooks in a new tab instead of the iframe dialog (as done below)? If were were able to determine whether the `value` is a URL, would it be better to use the iframe or a new tab?

jamiedemaria: I think yes, it'd always be preferable to open a Noteable notebook (or other hosted notebook) in a new tab

jamiedemaria: @yuhan do you have any thoughts on this? I don't see a case where we wouldn't want to open a noteable notebook in a new tab (vs rendering in dagit), but I may be missing something. And I think it makes sense that we would generalize the behavior so that if `notebookPath` is a URL then we open it in a new tab

yuhan: imo for integration, it'd be better to open in a new tab (it feels to me more like an impression/experience feeling that, for example, if anything goes wrong on the third party end, when it's in a separate tab, the user would know immediately where to seek for help, whereas if we embed it in an iframe the user may not realize it was outside dagster). but for plain notebooks, we'd still want to show them in iframes.

> And I think it makes sense that we would generalize the behavior so that if notebookPath is a URL then we open it in a new tab
i like this idea.

yuhan (commented):
the implementation looks good to me! but im no frontend expert so i'll defer to dish to do a final sign off :)

js_modules/dagit/packages/core/src/plugins/ipynb.tsx
jamiedemaria: ok! @hellendag do you know of a way for us to check if a string is a URL?

hellendag: - What does `value` look like in the iframe case? Is it a path or something?
- Is the URL expected to be an external URL?

If you want to check whether the value is a valid full URL, you could do something like:

```
const value = notebookPath.value;
const url = React.useMemo(() => {
  let urlString = null;
  try {
    const url = new URL(value);
    return url.toString();
  } catch (e) {
    // Not a full valid URL
    return null;
  }
}, [value]);

if (url) {
  return <ExternalAnchorButton href={url}>...</ExternalAnchorButton>;
}

// Otherwise, it's presumably a path to open in an iframe Dialog?
return <...>...</...>;
```

jamiedemaria:
@hellendag  ready for more review! I added in your code snippet for checking is the string is a URL and it works great, but I get a little linting error in VScode for the line

```
const url = React.useMemo(() => {
``` 

saying `React Hook "React.useMemo" is called conditionally. React Hooks must be called in the exact same order in every component render.eslint[react-hooks/rules-of-hooks](https://reactjs.org/docs/hooks-rules.html)` is that something to be concerned about? how do i fix it since it doesn't look to me that the useMemo call is conditional?

js_modules/dagit/packages/core/src/plugins/ipynb.tsx
```diff
@@ -22,6 +22,20 @@ export const SidebarComponent: React.FC<IPluginSidebarProps> = (props) => {
...
+  const value = notebookPath.value;
+  const url = React.useMemo(() => {
+    try {
+      const url = new URL(value);
+      return url.toString();
+    } catch (e) {
+      // Not a full valid URL
+      return null;
+    }
+  }, [value]);
```
hellendag, line 34: Ah, yeah, the hook warning is valid.

Hooks cannot be called conditionally, which means this will need to go above the `!notebookPath` conditional. To accommodate the behavior of the hook, you can do:

```
const value = notebookPath?.value || '';
const url = React.useMemo(() => {
  // same behavior as before
}, [value]);

// This conditional has to be below the hook.
if (!notebookPath) {
  return <span />;
}

// Otherwise, do stuff with URL
```

jamiedemaria: fixed!

hellendag (approved):

PR #9897, opened 2022-10-05
Only add buildkite steps for affected changes

jmsanders:
This will benefit from some refactoring once I play around with it a bit
more.

We're still going to see a lot of tests run on a lot of PRs because
almost everything requires dagster.

Right now, this doesn't build the steps at all. We might also want to
experiment with somehow marking the steps as skipped so it's obvious
what did and didn't run on a PR.

### DIFF
--- a/.buildkite/dagster-buildkite/dagster_buildkite/utils.py
+++ b/.buildkite/dagster-buildkite/dagster_buildkite/utils.py
@@ -1,5 +1,6 @@
 import os
 import subprocess
+from pathlib import Path
 from typing import Dict, List, Optional, Union
 
 import packaging.version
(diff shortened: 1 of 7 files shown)


brad-alexander (approved):

.buildkite/dagster-buildkite/dagster_buildkite/utils.py
```diff
@@ -176,3 +177,13 @@ def parse_package_version(version_str: str) -> packaging.version.Version:
         parsed_version, packaging.version.Version
     ), f"Found LegacyVersion: {version_str}"
     return parsed_version
+
+
+def get_changed_files():
+    paths = (
+        subprocess.check_output(["git", "diff", "origin/master...HEAD", "--name-only"])
```
alangenfeld, line 184: are we certain about these commit targets of `origin/master` and `HEAD` ? I've seen issues with our current change detection acting out of line with expectations. I never had the chance to try to dig in and debug but by suspicion is that it has to do with this.

jmsanders: I just lifted the same incantation that we use elsewhere. Didn’t realize there were doubts about it. I’ll look into it in more detail tomorrow.

PR #9899, opened 2022-10-05
Observable source asset decorator

smackesey:
### Summary & Motivation

New `observable_source_asset` decorator.

Unit tests

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/source_asset_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/source_asset_decorator.py
@@ -1,5 +1,6 @@
 from typing import TYPE_CHECKING, Mapping, Optional, Sequence, Union, overload
 
+from dagster._annotations import experimental
 from dagster._core.definitions.events import AssetKey, CoercibleToAssetKeyPrefix
 from dagster._core.definitions.metadata import (
     MetadataEntry,
@@ -39,6 +40,7 @@ def observable_source_asset(
     ...
 
 
+@experimental
 def observable_source_asset(
     observe_fn: Optional[SourceAssetObserveFunction] = None,
     *,
@@ -53,7 +55,36 @@ def observable_source_asset(
     group_name: Optional[str] = None,
     resource_defs: Optional[Mapping[str, ResourceDefinition]] = None,
 ) -> Union[SourceAsset, "_ObservableSourceAsset"]:
+    """Create a `SourceAsset` with an associated observation function.
 
+    The observation function of a source asset is wrapped inside of an op and can be executed as
+    part of a job. Each execution generates an `AssetObservation` event associated with the source
+    asset. The source asset observation function should return a metadata dictionary that will be
+    attached to the `AssetObservation`.
+
+    Args:
+        name (Optional[str]): The name of the source asset.  If not provided, defaults to the name of the
+            decorated function. The asset's name must be a valid name in dagster (ie only contains
+            letters, numbers, and _) and may not contain python reserved keywords.
+        key_prefix (Optional[Union[str, Sequence[str]]]): If provided, the source asset's key is the
+            concatenation of the key_prefix and the asset's name, which defaults to the name of
+            the decorated function. Each item in key_prefix must be a valid name in dagster (ie only
+            contains letters, numbers, and _) and may not contain python reserved keywords.
+        metadata (Mapping[str, RawMetadataValue]): Metadata associated with the asset.
+        io_manager_key (Optional[str]): The key for the IOManager that will be used to load the contents of
+            the source asset when it's used as an input to other assets inside a job.
+        io_manager_def (Optional[IOManagerDefinition]): (Experimental) The definition of the IOManager that will be used to load the contents of
+            the source asset when it's used as an input to other assets inside a job.
+        description (Optional[str]): The description of the asset.
+        partitions_def (Optional[PartitionsDefinition]): Defines the set of partition keys that
+            compose the source asset.
+        group_name (Optional[str]): A string name used to organize multiple assets into groups. If not provided,
+            the name "default" is used.
+        resource_defs (Optional[Mapping[str, ResourceDefinition]]): (Experimental) resource
+            definitions that may be required by the :py:class:`dagster.IOManagerDefinition` provided in
+            the `io_manager_def` argument.
+        observe_fn (Optional[SourceAssetObserveFunction]) Observation function for the source asset.
+    """
     if observe_fn is not None:
         return _ObservableSourceAsset()(observe_fn)
 
--- a/python_modules/dagster/dagster/_core/definitions/decorators/source_asset_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/source_asset_decorator.py
@@ -1,9 +1,6 @@
 from typing import TYPE_CHECKING, Mapping, Optional, Sequence, Union, overload
 
-from typing_extensions import Protocol, TypeAlias
-
-from dagster._core.definitions.events import AssetKey, AssetObservation, CoercibleToAssetKeyPrefix
-from dagster._core.definitions.logical_version import LogicalVersion
+from dagster._core.definitions.events import AssetKey, CoercibleToAssetKeyPrefix
 from dagster._core.definitions.metadata import (
     MetadataEntry,
     MetadataUserInput,
@@ -12,8 +9,6 @@ from dagster._core.definitions.metadata import (
 from dagster._core.definitions.partition import PartitionsDefinition
 from dagster._core.definitions.resource_definition import ResourceDefinition
 from dagster._core.definitions.source_asset import SourceAsset, SourceAssetObserveFunction
-from dagster._core.errors import DagsterInvalidObservationError
-from dagster._core.execution.context.compute import OpExecutionContext
 from dagster._core.storage.io_manager import IOManagerDefinition
 
 from ..decorators.op_decorator import _Op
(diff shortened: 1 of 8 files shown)


smackesey:
* master
  * **PR #9899** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9899" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9899?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/decorators/source_asset_decorator.py
```diff
@@ -0,0 +1,155 @@
...
+    description: Optional[str] = None,
+    partitions_def: Optional[PartitionsDefinition] = None,
+    _metadata_entries: Optional[Sequence[Union[MetadataEntry, PartitionMetadataEntry]]] = None,
+    group_name: Optional[str] = None,
+    resource_defs: Optional[Mapping[str, ResourceDefinition]] = None,
+) -> "_SourceAsset":
+    ...
+
+
+def source_asset(
```
sryza, line 52: what do you think about calling this something like `observeable_source_asset`?   I've thought about the idea of a different source asset decorator that contains a function that loads the source asset (which could then maybe be called `loadable_source_asset`).

python_modules/dagster/dagster/_core/definitions/decorators/source_asset_decorator.py
```diff
@@ -0,0 +1,155 @@
...
+            key=source_asset_key,
+            metadata=self.metadata,
+            io_manager_key=self.io_manager_key,
+            io_manager_def=self.io_manager_def,
+            description=self.description,
+            partitions_def=self.partitions_def,
+            _metadata_entries=self._metadata_entries,
+            group_name=self.group_name,
+            resource_defs=self.resource_defs,
+            node_def=op,
```
sryza, line 143: I suspect that there are situations where it may be useful inside a sensor instead of inside a run.  E.g. perhaps the sensor produced by `build_asset_reconciliation_sensor` should invoke it as part of sensor execution.

1. Thoughts?
2. If so, then maybe the SourceAsset shouldn't have an op and should just have this function directly? Then we can convert it to an op later on.

python_modules/dagster/dagster/_core/definitions/logical_version.py
```diff
@@ -0,0 +1,93 @@
...
+from typing_extensions import Final
+
+from dagster import _check as check
+
+if TYPE_CHECKING:
+    from dagster._core.definitions.events import AssetKey
+    from dagster._core.events.log import EventLogEntry
+    from dagster._core.instance import DagsterInstance
+
+class LogicalVersion:
```
sryza, line 15: Do you imagine we may later want to add other attributes to this?

python_modules/dagster/dagster/_core/definitions/logical_version.py
```diff
@@ -0,0 +1,93 @@
...
+        else get_most_recent_logical_version(k, key_to_is_source_map[k], instance)
+        for k in ordered_dependency_keys
+    ]
+    all_inputs = [op_version, *dependency_logical_versions]
+    hash_sig = sha256()
+    hash_sig.update(bytearray("".join(all_inputs), "utf8"))
+    return LogicalVersion(hash_sig.hexdigest())
+
+
+def get_most_recent_logical_version(
```
sryza, line 55: I would expect this to live inside the instance.

sryza: also, can we do some caching or something to avoid fetching the same event multiple times within a single step?

python_modules/dagster/dagster/_core/definitions/source_asset.py
```diff
@@ -126,6 +161,46 @@ def io_manager_def(self) -> Optional[IOManagerDefinition]:
...
+    def is_versioned(self) -> bool:
+        return self.node_def is not None
+
+    @property
+    def node_def(self):
+        """Op that generates observation metadata for a source asset."""
+        if self.observe_fn is None:
+            return None
+        else:
+            observe_fn = self.observe_fn
```
sryza, line 184: Is this line necessary? Why not just use `self.observe_fn` below?

python_modules/dagster/dagster/_core/definitions/source_asset.py
```diff
@@ -126,6 +161,46 @@ def io_manager_def(self) -> Optional[IOManagerDefinition]:
...
+        )
+        return cast(OpDefinition, self.node_def)
+
+    @public
+    @property
+    def is_versioned(self) -> bool:
+        return self.node_def is not None
+
+    @property
+    def node_def(self):
```
sryza, line 179: Return type annotation would be helpful

python_modules/dagster/dagster/_core/definitions/source_asset.py
```diff
@@ -200,3 +277,15 @@ def get_resource_requirements(self) -> Iterator[ResourceRequirement]:
         )
         for source_key, resource_def in self.resource_defs.items():
             yield from resource_def.get_resource_requirements(outer_context=source_key)
+
+    def __hash__(self):
```
sryza, line 281: This strikes me as a tiny bit sketchy. Do we anticipate using source assets as keys in dictionaries?

python_modules/dagster/dagster/_core/definitions/decorators/source_asset_decorator.py
```diff
@@ -0,0 +1,118 @@
...
+    key_prefix: Optional[CoercibleToAssetKeyPrefix] = None,
+    metadata: Optional[MetadataUserInput] = None,
+    io_manager_key: Optional[str] = None,
+    io_manager_def: Optional[IOManagerDefinition] = None,
+    description: Optional[str] = None,
+    partitions_def: Optional[PartitionsDefinition] = None,
+    _metadata_entries: Optional[Sequence[Union[MetadataEntry, PartitionMetadataEntry]]] = None,
+    group_name: Optional[str] = None,
+    resource_defs: Optional[Mapping[str, ResourceDefinition]] = None,
+) -> Union[SourceAsset, "_ObservableSourceAsset"]:
```
sryza, line 55: This needs docstring

python_modules/dagster/dagster/_core/definitions/decorators/source_asset_decorator.py
```diff
@@ -0,0 +1,118 @@
...
+    description: Optional[str] = None,
+    partitions_def: Optional[PartitionsDefinition] = None,
+    _metadata_entries: Optional[Sequence[Union[MetadataEntry, PartitionMetadataEntry]]] = None,
+    group_name: Optional[str] = None,
+    resource_defs: Optional[Mapping[str, ResourceDefinition]] = None,
+) -> "_ObservableSourceAsset":
+    ...
+
+
+def observable_source_asset(
```
sryza, line 42: I think it would be safest to mark this as experimental for now, though open to otherwise if you disagree

python_modules/dagster/dagster/_core/definitions/source_asset.py
```diff
@@ -24,27 +27,44 @@
...
-            ("description", PublicAttr[Optional[str]]),
-            ("partitions_def", PublicAttr[Optional[PartitionsDefinition]]),
-            ("group_name", PublicAttr[str]),
-            ("resource_defs", PublicAttr[Dict[str, ResourceDefinition]]),
-        ],
-    ),
-    ResourceAddable,
-):
+
+class SourceAssetObserveFunctionWithContext(Protocol):
```
sryza, line 44: Why not use `Callable[[SourceAssetObserveContext], MetadataUserInput`?

python_modules/dagster/dagster/_core/definitions/source_asset.py
```diff
@@ -58,10 +78,22 @@ class SourceAsset(
...
+    key: PublicAttr[AssetKey]
+    metadata_entries: Sequence[Union[MetadataEntry, PartitionMetadataEntry]]
+    io_manager_key: PublicAttr[Optional[str]]
+    _io_manager_def: PublicAttr[Optional[IOManagerDefinition]]
+    description: PublicAttr[Optional[str]]
+    partitions_def: PublicAttr[Optional[PartitionsDefinition]]
+    group_name: PublicAttr[str]
+    resource_defs: PublicAttr[Dict[str, ResourceDefinition]]
+    observe_fn: PublicAttr[Optional[SourceAssetObserveFunction]]
+    _node_def: Optional[OpDefinition]
```
sryza, line 93: I think worth adding a comment saying that this is computed lazily

sryza (commented):
Left a few comments but looks pretty much there

python_modules/dagster/dagster/_core/definitions/source_asset.py
smackesey: This is required to create a closure for the observe function-- mypy flags it otherwise. While I don't _think_ it would break using `self.observe_fn` directly, the type-checker correctly observes that `self.observe_fn`, being a nullable property, might return `None` when it's called at some unpredictable point inside the `compute_fn`.

python_modules/dagster/dagster/_core/definitions/source_asset.py
smackesey: This was added to allow de-duping of source assets inside sets in a downstream PR. Might no longer be required, I'll remove it from this and add it back if it turns out to be needed downsteram.

python_modules/dagster/dagster/_core/definitions/source_asset.py
smackesey: Because the function only optionally takes a context function. This can't be represented with `Callable` unfortunately (that's why we don't do it for regular assets).

python_modules/dagster/dagster/_core/definitions/decorators/source_asset_decorator.py
smackesey: agreed

python_modules/dagster-graphql/dagster_graphql/cli.py
```diff
@@ -43,12 +43,15 @@ def execute_query(
 
     context = workspace_process_context.create_request_context()
 
-    result = cast(ExecutionResult, graphql(
-        request_string=query,
-        schema=create_schema(),
-        context_value=context,
-        variable_values=variables,
-    ))
+    result = cast(
```
sryza, line 46: This looks unrelated, right?

python_modules/dagster/dagster/_core/definitions/source_asset.py
sryza: I was imagining:
```
SourceAssetObserveFunction: TypeAlias = Union[
    Callable[[SourceAssetObserveContext], MetadataUserInput], Callable[[], MetadataUserInput]
]

```

sryza (approved):
A couple small last things. Otherwise looks great.

python_modules/dagster/dagster/_core/definitions/source_asset.py
smackesey: Yeah, this was the first thing I tried and for some reason it does not work (though logically it should).

sryza: got it - was not a blocking concern, just curious

PR #9900, opened 2022-10-05
Tweak and document Launch Type in existing VPC setup docs

gibsondan:
summary:
Missed a spot where we reference the new launch type.

Test Plan: view docs


jmsanders (approved):
The github .png diff is fun - didn't know it could do that.

PR #9901, opened 2022-10-05
Fix #5737

### LINKED ISSUE #5737: DagsterTranslator.codify is missing comment argument from superclass
## Summary
If you are importing dagstermill, but also call papermill directly on a Python notebook, you will get

```
TypeError: codify() takes 2 positional arguments but 3 were given
```

because dagstermill is [replacing](https://github.com/dagster-io/dagster/blob/b5713af752146674d1f96436d0a01363e6834cb7/python_modules/libraries/dagstermill/dagstermill/translator.py#L63) papermill's `PythonTranslator` with its own `DagsterTranslator`. However, the signature of `DagsterTranslator.codify` does not match that of the [superclass](
https://github.com/nteract/papermill/blob/aecd0ac16d34dd41a1a8e4ca3351844932af3d94/papermill/translators.py#L117), which does take an additional `comment` argument:
https://github.com/dagster-io/dagster/blob/b5713af752146674d1f96436d0a01363e6834cb7/python_modules/libraries/dagstermill/dagstermill/translator.py#L32

## Reproduction

```python
import dagstermill as dm
import papermill as pm

pm.execute_notebook(
       "notebook.ipynb",
       "notebook_output.ipynb",
       parameters={"my_param": "some value"},
)
```

## Additional Info about Your Environment

python                    3.7.10
dagstermill               0.13.4 
papermill                 2.3.3

tizz98:
### Summary & Motivation

When developing https://github.com/noteable-io/papermill-origami we ran into the same issue as described in #5737.
We fixed the issue by creating our own implementation of the `DagsterTranslator` https://github.com/noteable-io/papermill-origami/blob/main/papermill_origami/noteable_dagstermill/translator.py#L38

Our own implementation https://github.com/noteable-io/papermill-origami/blob/main/papermill_origami/noteable_dagstermill/translator.py#L38 works and by having this change upstream we won't need to maintain our own implementation.

### DIFF
--- a/python_modules/libraries/dagstermill/dagstermill/translator.py
+++ b/python_modules/libraries/dagstermill/dagstermill/translator.py
@@ -30,7 +30,8 @@ context = __dm_dagstermill._reconstitute_pipeline_context(
 
 class DagsterTranslator(papermill.translators.PythonTranslator):
     @classmethod
-    def codify(cls, parameters):  # pylint: disable=arguments-differ
+    def codify(cls, parameters, comment=None):
+        # comment is not used but is a required argument on newer versions of papermill
         assert "__dm_context" in parameters
         assert "__dm_executable_dict" in parameters
         assert "__dm_pipeline_run_dict" in parameters


sryza:
Thanks for submitting this @tizz98 . Do you know whether this will break with older versions of Papermill?

tizz98:
@sryza I think it should work given that `comment=None` has a default value, older versions of papermill that don't pass the parameter won't care

sryza:
@tizz98 makes sense. Kicking off a build.

sryza (approved):

tizz98:
Thanks @sryza

PR #9903, opened 2022-10-05
Add skipped events to multi asset sensor context

clairelin135:
This PR updates the multi asset sensor context to be able to handle "AND" cases by storing 25 skipped event IDs per asset in the cursor, alongside the latest evaluated materialization ID. Also breaks the cursor out into its own object for ease of updating. After evaluation, this cursor object is then serialized.

Example cases to demonstrate this functionality:
- Say partitions A and B are materialized in that order, and only B is marked as evaluated. After this tick, A is in the "skipped events". If A is then materialized again in the next tick, we remove A from the skipped events
- Say partition A is materialized twice in the same tick. Calling "advance_cursor" on second materialization will also mark the first materialization as evaluated
- We raise an error when the skipped events list for each asset is at the maximum size (25) and another event is attempted to be added
- Say we fetch the last 10 events, but we only call "advance_cursor" on the first. The cursor's pointer to the last evaluated materialization will be the first event, and the following 9 events will be after the cursor. Thus, they will show up in future evaluations

Open questions:
- Should we rename `advance_cursor` and `advance_all_cursors`? Now that we have a different meaning of "cursor", maybe we should rename `advance_cursor` -> `mark_materialization_as_evaluated` and `advance_all_cursors` -> `mark_all_materializations_as_evaluated`

Code sample of materializing a partition if the corresponding partitions of `upstream_daily_1` AND `upstream_daily_2` are newly materialized:
```
@multi_asset_sensor(
    asset_keys=[AssetKey("upstream_daily_1"), AssetKey("upstream_daily_2")],
    job=downstream_daily_job,
)
def trigger_daily_asset_if_both_upstream_partitions_materialized(context):
    run_requests = []
    for (
        partition,
        materializations_by_asset,
    ) in context.latest_materialization_records_by_partition_and_asset().items():
        if set(materializations_by_asset.keys()) == set(context.asset_keys):
            run_requests.append(downstream_daily_job.run_request_for_partition(partition))
            for asset_key, materialization in materializations_by_asset.items():
                context.advance_cursor({asset_key: materialization})
    return run_requests
```


python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -181,6 +181,136 @@ def _get_partition_key_from_event_log_record(event_log_record: "EventLogRecord")
...
+
+MAX_NUM_SKIPPED_EVENTS = 25
+
+
+class MultiAssetSensorContextCursor:
+    # Tracks the state of the cursor within the tick, created for utility purposes.
+    # Must call MultiAssetSensorEvaluationContext._update_cursor_after_evaluation at end of tick
+    # to serialize the cursor.
+    def __init__(self, cursor: Optional[str], context: "MultiAssetSensorEvaluationContext"):
+        self._unpacked_cursor: Dict[str, Tuple[Optional[str], Optional[int], List[int]]] = (
```
sryza, line 212: What do you think about defining a `NamedTuple` class called something like `CursorAssetComponent` to house this information and make it clear what's included? Also, a name for this dictionary that communicates what the keys are would be helpful.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -181,6 +181,136 @@ def _get_partition_key_from_event_log_record(event_log_record: "EventLogRecord")
...
+        )
+        check.dict_param(self._unpacked_cursor, "unpacked_cursor", key_type=str)
+        self._context = context
+        self._cursor_has_been_updated = False
+
+    def get_cursor_for_asset(self, asset_key: AssetKey):
+        partition_key, cursor, skipped_event_ids = self._unpacked_cursor.get(
+            str(asset_key), (None, None, [])
+        )
+        return partition_key, cursor, skipped_event_ids
```
sryza, line 223: Calling this `skipped_event_ids` feels a little off to me, for a few reasons:
* "skip" already has a meaning in the world of sensors, with `SkipReason`, which is independent of the meaning here.
* Partitions are heavily involved in the meaning of what's stored here. It feels like the name should have "partition" in it somewhere.

I'm trying to think through what the best term is for this and what the best way is to think about this.  A useful building block could be a term that means "an event that the user has advanced past".

Here's my understanding of what "skipped events" means: within a particular asset, what's currently called "skipped events" includes all events that occurred before the latest event that the user has advanced past, but that, for one of the asset's partitions, occurred after the latest event that the user has advanced past for that asset partition.

A bit of a mouthful, but maybe that gives us some fodder for a name?

sryza:
I love how simple this makes the code sample

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
sryza: Another way of putting it is that we're tracking the latest event that we've advanced passed for every asset partition, except that we're able to safely ignore asset partitions that don't have any events that we haven't advanced past that are before the latest event that we've advanced past for the asset.

(Also a mouthful, just spewing words as I deepen my understanding of what's going on here)

clairelin135: I was thinking `unevaluated_event_ids`?

clairelin135: > Here's my understanding of what "skipped events" means: within a particular asset, what's currently called "skipped events" includes all events that occurred before the latest event that the user has advanced past, but that, for one of the asset's partitions, occurred after the latest event that the user has advanced past for that asset partition.

Yes, that's right.

If we wanted to include "partition" in the name, we could say, "unevaluated_partition_event_ids"? Though it's possible unevaluated events could be unpartitioned. Can't name a good reason why someone would want to do that, but its still theoretically possible

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -278,46 +487,55 @@ def _get_partitions_after_cursor(self, asset_key: AssetKey) -> List[str]:
             ]
         return partitions_to_fetch
 
+    def update_cursor_after_evaluation(self) -> None:
+        """Updates the cursor after the sensor evaluation function has been called. This method
+        should be called at most once per evaluation.
+        """
+
+        for fetched_event_id, advanced in self._event_fetched_is_advanced.items():
```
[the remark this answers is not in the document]
clairelin135: Hmm, the current implementation of the cursor is to skip past unfetched events.

I also have a hard time thinking of why someone would want unfetched events to show up in their skipped events. It might also be annoying to have the "maximum skipped events" error thrown when those events were never requested.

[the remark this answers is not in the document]
clairelin135: Trying to think of when an event might be unfetched:
- Calling `latest_materialization_records_by_key`, which will only fetch the latest event per asset. Future calls won't even pull up the skipped event because the cursor points to the latest evaluated event and any skipped event will be before that
- `materialization_records_for_key` is called sporadically and there are unfetched events in between each call. These events are presumably unpartitioned and since the user fetched updated events later, can't think of a reason why they would want old unfetched events
- User accidentally creates unpartitioned materialization for a partitioned asset. This event wouldn't show up in `latest_materialization_records_by_partition`/`latest_materialization_records_by_partition_and_asset`, but we could remedy this by also including `None` partitions in return value

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -1332,6 +1596,7 @@ def __init__(
         job: Optional[ExecutableDefinition] = None,
         jobs: Optional[Sequence[ExecutableDefinition]] = None,
         default_status: DefaultSensorStatus = DefaultSensorStatus.STOPPED,
+        is_asset_reconciliation_sensor: bool = False,
```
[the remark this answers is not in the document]
clairelin135: yep -- I updated the `update_cursor_after_evaluation` function to only update the cursor if events were fetched through the context methods. This is cleaner now!

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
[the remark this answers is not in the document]
sryza: As general principle, I believe it's important to hard a strong separation between functions that mutate state that functions that retrieve it.  Without this separation, it makes it more difficult for users to reason about what's going on

> Hmm, the current implementation of the cursor is to skip past unfetched events.

Does that need to be the case? I believe that, from the user's perspective, we could make fetching and advancing independent.  Under the covers, `advance_cursor` (or `update_cursor_after_evaluation`) could fetch whatever data is needed in order to to do its thing, right? We could maintain a cache inside the context so that we don't fetch the same data twice?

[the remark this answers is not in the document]
clairelin135: I can see a situation, for example if a user updates partitions sequentially and thus only updates partitions after the cursor via the `after_cursor_partition` flag, where storing unfetched materializations of prior partitions in the cursor is undesirable

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
sryza: > Though it's possible unevaluated events could be unpartitioned.

IMO we shouldn't support this - there's not a clear criteria for when these should be included.

sryza: I.e. right now users aren't directly determining which events are included in this data structure, right? We're managing it.

clairelin135:
@sryza This PR now adds unfetched partitioned events to the `unevaluated_partitioned_event_ids` list for each asset cursor. I've also updated the other parts of the code per your feedback. I think it's ready for you to take another look!

sryza (commented):
Thanks for bearing with all the feedback on this - especially the nitpicky feedback on naming. This is fancy functionality (but very valuable IMO!), that's going to be a foundation for even fancier functionality, with partitioned asset reconciliation and SLA-based scheduling.  I think we'll be able to move a lot faster in the long run if we put in the effort now to make it as legible as possible.

As I review this PR, I'm finding it somewhat difficult to understand the dataflow. Is this an accurate abstraction of it?

image

If so, I wonder if it would be sensible to explicitly define some sort of object that contains just the data provided to `advance_cursor`?  That would help encapsulate the state mutation.

 Then you could define some function that gets invoked at the end of each sensor evaluation called `_next_cursor` that accepts the instance, data structure of cursor advances, and the previous cursor, and returns the next cursor. Similar to `update_cursor_after_evaluation`, but that makes the inputs and outputs very clear.

clairelin135:
@sryza Yep, your diagram is accurate, except that the state could be mutated multiple times within one tick. We do have the `get_unconsumed_events` method that fetches the skipped events at any given time. And calling `advance_cursor` could remove certain unconsumed events or move the latest materialized event id to a newer event, which will result in different results if context methods that retrieve materializations are invoked again.

I can add an object that does this, but I think we will need to ensure it can be invoked multiple times per tick

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -181,6 +180,204 @@ def _get_partition_key_from_event_log_record(event_log_record: "EventLogRecord")
     return None
 
 
+def _get_asset_key_from_event_log_record(event_log_record: "EventLogRecord") -> AssetKey:
```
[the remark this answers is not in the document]
clairelin135: Got it. I'd prefer to do this in a PR after this one is merged. We probably should also add the `get_partition_key_from_event_log_record` method as well

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -181,6 +180,207 @@ def _get_partition_key_from_event_log_record(event_log_record: "EventLogRecord")
...
+    )
+):
+    """A cursor component that is used to track the cursor for a particular asset in a multi-asset
+    sensor.
+
+    Attributes:
+        latest_evaluated_event_partition (Optional[str]): The partition of the latest evaluated event
+            for this asset.
+        latest_evaluated_event_id (Optional[int]): The event ID of the latest evaluated event for this asset.
+        unevaluated_partitioned_event_ids (List[int]): The storage IDs of the partitioned events that have not yet been evaluated.
```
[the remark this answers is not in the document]
clairelin135: I like "unconsumed". I can make these changes

sryza:
> And calling advance_cursor could remove certain unconsumed events or move the latest materialized event id to a newer event, which will result in different results if context methods that retrieve materializations are invoked again.

Ahhh I didn't realize we were supporting this.  If supporting this adds a lot of complexity, maybe we shouldn't support it?  It's hard for me to think of why users would need it, though let me know if you have something in mind.

clairelin135:
@sryza Since we are adding the functionality to have a set of "unconsumed events" for each asset that raises an error if it reaches it maximum size, I think we need to give users some way to resolve these unconsumed events. 

I was thinking they could use the `get_unconsumed_events` method to retrieve these events and selectively advance certain events via `advance_cursor`. Otherwise, the only other way to resolve these events is calling `advance_all_cursors` to clear all of the unconsumed events.

clairelin135:
@sryza Yeah, users ideally wouldn't be making multiple context calls to fetch materializations sequentially. I am okay with not supporting this

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -181,6 +180,207 @@ def _get_partition_key_from_event_log_record(event_log_record: "EventLogRecord")
...
+        [
+            ("latest_evaluated_event_partition", Optional[str]),
+            ("latest_evaluated_event_id", Optional[int]),
+            ("unevaluated_partitioned_event_ids", List[int]),
+        ],
+    )
+):
+    """A cursor component that is used to track the cursor for a particular asset in a multi-asset
+    sensor.
+
```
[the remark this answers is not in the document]
clairelin135: This is great, thanks!

clairelin135:
@sryza This PR is now updated with all of the changes mentioned above, as well as the single advance_cursor/advance_all_cursor mutation per tick. I think it's ready for you to take another look!

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -260,9 +448,61 @@ def __init__(
...
+    def _get_unconsumed_events_with_ids(self, event_ids: List[int]) -> Sequence["EventLogRecord"]:
+        self._cache_initial_unconsumed_events()
+        unconsumed_events = []
+        for event_id in event_ids:
+            event = self._initial_unconsumed_events_by_id.get(event_id)
+            unconsumed_events.extend([event] if event else [])
+
+        return unconsumed_events
+
+    @public
```
sryza, line 486: Is there a situation where we imagine that a user would need to use this? If not, could be best to hold it back? If so, I think we should make sure that the name makes it clear these are _trailing_ unconsumed events (if I'm understanding correctly that they're only the trailing ones).

sryza (commented):
This looks 98% ready to go!

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
clairelin135: Hmm, so I included this method so that users would have a way to mark the unconsumed events as consumed when they reach the max num allowed unconsumed events. Otherwise, there aren't any built in ways for them to fetch the unconsumed events. 

I can update this name to be trailing unconsumed events instead!

clairelin135:
@sryza Thanks for taking that last look! I've updated the method names and fixed the nits you mentioned in the comments above

sryza (approved):
Ship it!

PR #9905, opened 2022-10-06
[dagit] Bring back “groups only” zoom level for large asset graphs

bengotow:
### Summary & Motivation

Preview: https://www.loom.com/share/9954d7dd0f63484ebe20bcf2b6a33ef0

This makes a few changes for a dagster customer with 392 assets in their asset graph:

- A new feature flag allows you to enable "Experimental "groups-only" asset graph zoom level"

- This allows you to zoom farther out on the asset graph. As you zoom out, we show only asset group contents, enlarge the size of the asset group levels, and darken + enlarge the lines between the groups.

- If you double-click one of the groups, it zooms you to that group's extent [this is not feature flagged]

This graph might not look perfect because we don't resize the nodes or re-arrange them as you zoom out to compact them together, but it should allow customers with huge graphs to at least see everything at once in a reasonably useful way.


js_modules/dagit/packages/core/src/app/Flags.tsx
```diff
@@ -10,6 +10,7 @@ export enum FeatureFlag {
   flagDisableWebsockets = 'flagDisableWebsockets',
   flagNewWorkspace = 'flagNewWorkspace',
   flagRunBucketing = 'flagRunBucketing',
+  flagAssetGraphExperiments = 'flagAssetGraphExperiments',
```
hellendag, line 13: Intentionally a generic name related to any asset graph experiments? Or should this be more specifically named in reference to the zoom level changes?

hellendag (approved):
excellent loom demo!

js_modules/dagit/packages/core/src/app/Flags.tsx
bengotow: Ahh good call I'll change this, I think this came from the old diff I restored to get some of this work!

bengotow:
Buildkite seems to be upset about fetching it's secrets and the Dagit portions have succeeded, going to merge this so we can include it in today's release.

PR #9906, opened 2022-10-06
[Feature] Add possibility to change request timeout value when calling airbyte

FransDel:
### Summary & Motivation
In our setup calls to airbyte fail to create a connection sync because API takes between ~20 and ~30s to respond. When we change this timeout value it works fine. This will enable anyone to change its value and doesn't create breaking changes because it defaults to 15s (its old value).

> Not sure where to change the package's version, neither if its needed


sryza:
@OwenKephart or @benpankow - mind taking a look?

benpankow (approved):
Looks good to me. Thanks for contributing this!

PR #9912, opened 2022-10-06
[docs] asset reconciliation sensor concept page

jamiedemaria:
### Summary & Motivation
Adds concepts documentation for `build_asset_reconciliation_sensor`

In this first pass I avoided going through an example graph like in the API doc, but I can add that or something similar to the concepts page if it would be valuable. I'm just weighing the length of the page and the info already being in the API doc vs having the info all in one place.


docs/content/concepts/partitions-schedules-sensors/sensors.mdx
```diff
@@ -183,6 +183,62 @@ def asset_a_and_b_sensor_with_skip_reason(context):
...
+        asset_b,
+        asset_c,
+        build_asset_reconciliation_sensor(
+            asset_selection=AssetSelection.assets(asset_b, asset_c),
+            name="asset_reconciliation_sensor",
+        ),
+    ]
+```
+
+This sensor will materialize `asset_b` whenever any of its parents materialize, and it will materialize `asset_c` whenever any of its parents materialize.
```
jamiedemaria, line 204: I want to avoid implying that the sensor will update `asset_b` and `asset_c` even if only `asset_b`'s parents have updated, but that leaves me with a bit of wordy/repetitive sentence. It reads ok to me, but if you get tripped up on it, i'll start tweaking it some more

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
```diff
@@ -183,6 +183,62 @@ def asset_a_and_b_sensor_with_skip_reason(context):
...
+        asset_c,
+        build_asset_reconciliation_sensor(
+            asset_selection=AssetSelection.assets(asset_b, asset_c),
+            name="asset_reconciliation_sensor",
+            wait_for_in_progress_runs=False,
+        ),
+    ]
+```
+
+**Note:** If you have an asset whose parents update frequently, you should consider setting `wait_for_in_progress_runs=False` to avoid a deadlock where the reconciliation sensor cannot materialize an asset because each time the sensor evaluates, there is a parent asset materializing.
```
jamiedemaria, line 240: How does this explanation sound? I've taken a couple passes at making it clear, but i'm definitely struggling with it a bit. will continue to iterate if it's unclear

jamiedemaria (commented):
Left a couple comments in places I'd definitely like some feedback on wording. 

In general i tried to keep the explanation of the sensor a bit more high level, with the intention that the API docs have the more in-depth explanation. I could add a sentence to the end like "For more information, and an in-depth walk through of the sensor's behavior, see the API docs for build_asset_sensor". Or if that doesn't seem like enough, I'll add some more explanation to the concepts page

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
clairelin135: Maybe it's worth mentioning the default behavior? I.e.:

By default, the asset reconciliation sensor does not materialize any asset whose parent is currently materializing. To avoid a deadlock for an asset whose parents update frequently, you can set `wait_for_in_progress_runs=False` to kick off a materialization regardless of whether any run in progress exists for a parent asset.

clairelin135 (commented):
This is looking good to me! I think this is the right level of information. Agree with you that for more specifics, the user can reference the API docs.

For my current work with the multi asset sensor, I think it might be necessary to break out the asset sensor content in the docs into a separate page to avoid the sensors page becoming too unwieldy. So if you feel like it would be helpful to add more content to explain the functionality, maybe its worth breaking this out into a different page now

docs/content/concepts/partitions-schedules-sensors/sensors.mdx
jamiedemaria: yeah reiterating that would be good

sryza (approved):
This looks great

PR #9913, opened 2022-10-06
[dagit] Run timeline: reduce scheduled tick width, reduce chunk min width

hellendag:
### Summary & Motivation

In an effort to reduce how often high-firing scheduled ticks are batched together in the run timeline:

- Reduce the "duration" of scheduled ticks to 5 seconds (instead of 10) to make them a bit narrower
- Reduce the minimum chunk width to 12px, which should hopefully keep some runs/ticks separate that are currently being batched.

These are probably fairly subtle changes, but maybe they make some of the current batching slightly less confusing.

View storybook example with scheduled ticks shown, view a run timeline in Dagit with lots of batching and verify that the batching is slightly less aggressive.

### DIFF
--- a/js_modules/dagit/packages/core/src/runs/RunTimeline.stories.tsx
+++ b/js_modules/dagit/packages/core/src/runs/RunTimeline.stories.tsx
@@ -216,7 +216,7 @@ export const BucketByRepo = () => {
   );
 };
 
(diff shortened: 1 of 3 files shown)


prha (approved):
worth a shot

PR #9916, opened 2022-10-06
Show noteable logo on noteable backed assets/ops

jamiedemaria:
### Summary & Motivation
We talked about showing logos for assets backed by other tools (dbt, etc) so i took a stab at it for noteable. Not sure if `Tag` is the right component to use, but it looks pretty ok

Screen Shot 2022-10-06 at 12 51 06 PM


hellendag:
The icon looks cool! I'm not sure I see where the `<Tag>` is in the screenshot, though -- so I'm not sure it's necessary to use that component instead of just showing the icon. 🤔 (For reference, `<Tag>` will generally look like these: https://dagit-storybook.vercel.app/?path=/story/tag--basic)

jamiedemaria:
I think the `Tag` is just the logo. I have this `<Tag icon="noteable_logo" key={tag.label} />;` as the component and the screenshot is what that creates. I wasn't sure how to just show the icon, but when i searched for `icon=` in the code base i saw it used a lot in tags so that's what i did to just hack it together. Is there a correct way to just display the icon?

hellendag:
You can do `<Icon name="noteable_logo" />`. (This is what `<Tag>` does under the hood with the supplied `icon` prop.)

Since the DAG view has a white background, I think the transparency of the svg should end up looking fine without an additional background.

jamiedemaria:
@hellendag  updated to `Icon` lmk if there's anything else that should be done to get this looking good!

hellendag (approved):

PR #9931, opened 2022-10-07
Fix asset selection in asset launchpad

gibsondan:
Summary:
- Pass down assetSelection whenever it's set
- Hide asset ops that aren't in the selection, like we do for op selections

Test Plan:
View launchpad with an asset subset selected, it now opens with a subsetted config schema (ops that are invalid are indicated as such) and only the selected ops are selected.
Launchpad with no asset subset is unchanged

Is that a sufficient test plan / are there other weird subset cases to consider?


js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
```diff
@@ -254,6 +254,9 @@ async function stateForLaunchingAssets(
         pipelineName: jobName,
         repositoryName: assets[0].repository.name,
         repositoryLocationName: assets[0].repository.location.name,
+        assetSelection: assets.map((asset) => ({
+          path: asset.assetKey.path,
+        })),
```
gibsondan, line 259: this is the part I am the least confident in

python_modules/dagster-graphql/dagster_graphql/schema/inputs.py
```diff
@@ -80,6 +80,7 @@ class GraphenePipelineSelector(graphene.InputObjectType):
     repositoryName = graphene.NonNull(graphene.String)
     repositoryLocationName = graphene.NonNull(graphene.String)
     solidSelection = graphene.List(graphene.NonNull(graphene.String))
+    assetSelection = graphene.List(graphene.NonNull(GrapheneAssetKeyInput))
```
bengotow, line 83: This is cool! I can take a look through the other resolvers once this merges and make sure there aren't places where we're passing a PipelineSelector AND a list of AssetKeys as separate arguments, since they can now be sent together.

bengotow (approved):
This seems like a great improvement!

gibsondan:
somebody else seems to have fixed this, hooray

PR #9933, opened 2022-10-07
[dagster-io/ui] Middle truncation

hellendag:
### Summary & Motivation

Create a component that performs middle truncation on strings.

- Use canvas text measurement with the computed width and font style of the container.
- Use a resize observer to adjust widths as needed.
- Use a binary search to determine the appropriate truncation point.

I have included a Storybook example with a bunch of strings and a slider to adjust the container width, updated some existing middle-truncated strings in Dagit, and added some new middle truncations.

View Storybook example and Dagit use cases. Verify that when changing viewport size or adjusting the width slider in Storybook, the performance profile looks just fine. Verify that the middle truncation looks correct and fits within the allotted size.


hellendag:
* master
  * **PR #9933** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9933" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9933?utm_source=stack-comment).

js_modules/dagit/packages/ui/src/components/MiddleTruncate.tsx
```diff
@@ -0,0 +1,126 @@
+import useResizeObserver from '@react-hook/resize-observer';
```
salazarm, line 1: Should we use https://github.com/dagster-io/dagster/blob/master/js_modules/dagit/packages/core/src/gantt/useViewport.tsx instead?
We can add a flag to only enable the resize observer and disable monitoring scroll position/offset.

salazarm: Oh you can actually just not spread `onScroll` https://github.com/dagster-io/dagster/blob/master/js_modules/dagit/packages/core/src/gantt/useViewport.tsx#L159

js_modules/dagit/packages/ui/src/components/MiddleTruncate.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+    const targetWidth = width;
+    ctx.font = font;
+    body.appendChild(canvas);
+
+    // Binary search to find the maximum middle-truncated width that will fit within
+    // the container.
+    let start = 1;
+    const half = Math.floor(text.length / 2);
+    let end = half;
```
salazarm, line 105: `let end = text.length`

salazarm (approved):

js_modules/dagit/packages/ui/src/components/MiddleTruncate.tsx
hellendag: The lib I used here uses a single ResizeObserver for all observed elements, instead of a new ResizeObserver per element. Admittedly I don't know if the perf benefits are the same today as when the lib hook was written.

Perhaps `useViewport` could be updated to do the same, or to use the lib hook directly...

PR #9937, opened 2022-10-07
[dagster-airbyte] Support union types while generating normalization tables

benpankow:
## Summary

Allows our normalization-table-generating logic to correctly handle [union types](https://docs.airbyte.com/understanding-airbyte/supported-data-types/), which appear in some connection definitions.

Also added support for the `types` field, which I have seen appear (perhaps erroneously) instead of `type` in some connectors.

## Test Plan

Added unit test.

### DIFF
--- a/python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
+++ b/python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
@@ -13,7 +13,6 @@ from typing import (
     Optional,
     Sequence,
     Set,
-    Union,
     cast,
 )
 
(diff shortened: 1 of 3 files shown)


python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
```diff
@@ -237,19 +252,29 @@ def _get_normalization_tables_for_schema(
...
-    # Array types are also broken into a new table
-    elif schema["type"] == "array" or "array" in schema["type"]:
-        out.append(prefix + key)
-        for k, v in schema["items"]["properties"].items():
-            out += _get_normalization_tables_for_schema(k, v, f"{prefix}{key}_")
+
+    sub_schemas = _get_sub_schemas(schema)
+
+    for sub_schema in sub_schemas:
+        schema_type = _get_schema_types(sub_schema)
```
OwenKephart, line 259: this logic might be slightly simpler if _get_schema_types returns Optional[List[str]] (rather than either str, List[str], or None).

OwenKephart (approved):
looks good, had one small note

PR #9940, opened 2022-10-08
Add tolerations, nodeSelector, and podSecurityContext to job/pod context (fixes #9924)

### LINKED ISSUE #9924: K8sContainerContext is missing tolerations, nodeSelector, and podSecurityContext
### Dagster version

1.0.11

### What's the issue?

When a run job/pod is created, it doesn't include _all_ the config of the user repo deployment, even if `includeConfigInLaunchedRuns.enabled` is `true`. It is missing `tolerations`, `nodeSelector`, and `podSecurityContext`.

I expected all the config specified in the elements of the `deployments` list to be copied to the run jobs/pods, if `includeconfigInLaunchedRuns.enabled` is `true.

We'd need to add those missing properties here:
https://github.com/dagster-io/dagster/blob/f5ea4b7d861d4966166b54bc575b7a3ebbfe7851/python_modules/libraries/dagster-k8s/dagster_k8s/container_context.py#L24

and here:
https://github.com/dagster-io/dagster/blob/4754109a27fd0aedb107e4101818d8e4c174d492/helm/dagster/charts/dagster-user-deployments/templates/helpers/_helpers.tpl#L97

### How to reproduce?

_No response_

### Deployment type

Dagster Helm chart

### Deployment details

_No response_

### Additional information

_No response_

craustin:
Still getting environment setup. Have not run formatter, linter, or tests yet.


gibsondan (commented):
@craustin thanks so much for doing this.

I believe this PR would get the job done for your specific needs, but another possibility that we've been considering here is exposing the whole pod_spec_config as a dictionary on K8sContainerContext and swapping it in - that way as more of these configuration fields inevitably come up we could take them all out in one config field. Curious what you think about that direction - kind of generalization of this PR to one big dictionary field instead of three separate individual fields.

gibsondan:
(and sorry for the delay in getting to reviewing this PR)

gibsondan:
The trickiest bit I think would be the merge() implementation for pod_spec_config, there are a couple of options there but just doing a key-level merge, preferring the second one, may be the simplest place to start.

craustin:
Yes, that'd be great. 👍  It definitely felt like a more general solution would be best since that PR required so many similar updates in so many places.

gibsondan (changes requested):
Requesting changes for that suggestion as I think its the best way to keep from having to re-implement the whole k8s spec here :) Happy to work with you on that or I suspect we will add it ourselves fairly soon as well

ion-elgreco:
@gibsondan any plans to fix this?

cmpadden:
Closing due to inactivity and merge conflicts; please re-open if necessary, thank you

PR #9945, opened 2022-10-10
Data pipeline smoke test example

sryza:
### Summary & Motivation

This is the code that backs the upcoming data pipeline smoke test blog post


examples/assets_smoke_test/assets_smoke_test/pure_python_assets.py
```diff
@@ -0,0 +1,41 @@
...
+                "change": "string",
+            }
+        ),
+    },
+)
+
+
+@asset
+def country_populations(raw_country_populations) -> DataFrame:
+    country_populations = raw_country_populations.copy()
```
OwenKephart, line 24: I guess this is necessary when outputs are passed in-memory? or is there something more to this line?

python_modules/dagster/dagster/_core/definitions/metadata/table.py
```diff
@@ -121,6 +121,17 @@ def __new__(
             ),
         )
 
+    @public
+    @staticmethod
+    def from_name_type_dict(name_type_dict: Mapping[str, str]):
```
OwenKephart, line 126: this is clutch

examples/assets_smoke_test/assets_smoke_test_tests/smoke_test_python_and_dbt_assets.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def test_smoke_all():
+    snowflake_config = {
+        "user": os.environ["SNOWFLAKE_USER"],
+        "password": os.environ["SNOWFLAKE_PASSWORD"],
+        "account": os.environ["SNOWFLAKE_ACCOUNT"],
+    }
+
+    database_name = os.environ["SNOWFLAKE_DATABASE"]
```
OwenKephart, line 24: I see that database_name gets used later on, but I think it might be easier to read this test if the database is added to snowflake_config here, then pulled out of the dictionary on line 39.

examples/assets_smoke_test/dbt_project/config/profiles.yml
```diff
@@ -0,0 +1,28 @@
...
+      user: "{{ env_var('SNOWFLAKE_USER') }}"
+      password: "{{ env_var('SNOWFLAKE_PASSWORD') }}"
+
+      database: "{{ env_var('SNOWFLAKE_DATABASE') }}"
+      schema: PUBLIC
+      warehouse: ELEMENTL
+      client_session_keep_alive: False
+
+
+    smoke_test:
```
OwenKephart, line 18: why have two targets here if their config is identical? should we potentially write the smoke test outputs to a different schema or database?

examples/assets_smoke_test/assets_smoke_test/pure_python_assets.py
sryza: yeah, exactly

examples/assets_smoke_test/dbt_project/config/profiles.yml
sryza: the thinking was that in real life they'd end up varying. I'll vary the names of the env variables to make that more clear

PR #9950, opened 2022-10-10
Make the default_job_io_manager overridable via env

johannkm:
allow injecting a configurable class as the default io manager

Currently doesn't support injecting config- we don't need it, and yaml inside an env var would be gross

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/job_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/job_definition.py
@@ -1,3 +1,5 @@
+import importlib
+import os
 from functools import update_wrapper
 from typing import (
     TYPE_CHECKING,
@@ -47,7 +49,7 @@ from dagster._core.selector.subset_selector import (
     OpSelectionData,
     parse_op_selection,
 )
-from dagster._core.storage.io_manager import io_manager
+from dagster._core.storage.io_manager import IOManagerDefinition, io_manager
 from dagster._core.utils import str_format_set
 from dagster._utils import merge_dicts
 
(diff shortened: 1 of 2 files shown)


johannkm:
* master
  * **PR #9950** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9950" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9950?utm_source=stack-comment).

sryza (commented):
I started a thread with Nick to discuss the possibility of a more generic mechanism for the deployment to set resources.

Assuming we don't do that and we do go with this approach, main bits of feedback are:
- Pointing to a ResourceDefinition instead of a class would give us the flexibility to make the resource a context manager, and make it easier to provide config in the future if needed.
- I don't believe the implementation here will work for SDAs. I think that we provide the fs_io_manager to SDAs inside `with_resources`.  Worth testing.

python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -832,6 +836,19 @@ def get_direct_input_values_from_job(target: PipelineDefinition) -> Mapping[str,
...
+    module_name = os.getenv("DAGSTER_DEFAULT_IO_MANAGER_MODULE")
+    class_name = os.getenv("DAGSTER_DEFAULT_IO_MANAGER_CLASS")
+    if module_name and class_name:
+        from dagster._core.storage.io_manager import IOManager
+
+        io_manager_override = ConfigurableClassData(
+            module_name=module_name, class_name=class_name, config_yaml=yaml.dump({})
+        ).rehydrate()
+        check.invariant(isinstance(io_manager_override, IOManager))
+        return io_manager_override
```
shalabhc, line 849: Can we use a single env var - since you need both for this to work?

Also can you add a note re why we need the `ConfigurableClassData` and it is not just a simple `import()` like the default one.

johannkm: > Can we use a single env var
Not sure I follow. Like concatenate the strings?

johannkm:
@sryza 

For SDA's: I added a test that passes. Looks like it's injected to asset jobs here: 
https://github.com/dagster-io/dagster/blob/johann/10-10-Add_env_vars_to_override_default_job_io_manager/python_modules/dagster/dagster/_core/definitions/assets_job.py#L88
Any caveats with that though? I could imagine it won't work with load_asset_value?

> ResourceDefinition instead of a class would give us the flexibility to make the resource a context manager, and make it easier to provide config in the future if needed.

My take here is we can be very restrictive- we don't need to offer config of to be a context manager for our usecase. Is it worth being more general? What would the ResourceDefinition appraoch look like? A class that inherits from IOManagerDefinition and ConfigurableClass?

sryza:
> Any caveats with that though? I could imagine it won't work with load_asset_value?

I'm imagining the case where someone doesn't define a job at all - they just define some assets and then materialize them. Thinking about it more, that probably does already work, because everything is turned into a job behind the scenes before being shipped to the frontend.  But it might be worth testing.

> My take here is we can be very restrictive- we don't need to offer config of to be a context manager for our usecase.

We don't need to right now, but I think it's easy to imagine backing it with an implementation that needs to do some teardown at the end of a step.

> What would the ResourceDefinition appraoch look like? A class that inherits from IOManagerDefinition and ConfigurableClass?

I was imagining just pointing to an attribute that's defined at module scope and is an IOManagerDefinition.

johannkm:
Updated to point at an IOManagerDefinition

sryza (approved):
nice

python_modules/dagster/dagster/_core/definitions/job_definition.py
sryza: It can get a little confusing because `.` is also the delimiter for components within a Python package path

PR #9951, opened 2022-10-10
make it easier to fetch asset materialization metadata

sryza:
inspired by trying to write an answer to this StackOverflow question: https://stackoverflow.com/questions/73983827/how-can-i-do-an-incremental-load-based-on-record-id-in-dagster

### DIFF
--- a/python_modules/dagster/dagster_tests/core_tests/storage_tests/test_assets.py
+++ b/python_modules/dagster/dagster_tests/core_tests/storage_tests/test_assets.py
@@ -111,6 +111,9 @@ def test_backcompat_asset_materializations():
             b_mat = storage.get_latest_materialization_events([b]).get(b)
             _validate_materialization(b, b_mat)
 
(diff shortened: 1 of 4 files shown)


OwenKephart (approved):
Nice so users would be pointed towards doing something like `context.instance.get_latest_materialization_event(asset_key).metadata`?

python_modules/dagster/dagster_tests/core_tests/storage_tests/test_assets.py
```diff
@@ -111,6 +111,9 @@ def _validate_materialization(asset_key, event):
             b_mat = storage.get_latest_materialization_events([b]).get(b)
             _validate_materialization(b, b_mat)
 
+            b_mat = storage.get_latest_materialization_event(b)
```
OwenKephart, line 114: instance.get_latest_materialization_event

PR #9954, opened 2022-10-10
support asset keys and asset selection for multi asset sensors

jamiedemaria:
### Summary & Motivation
keeps asset_keys a list of keys rather than converting to an AssetSelection so that we can support source assets


sryza (commented):
Imagine that, in the future, we had a way for asset selections to include assets in repositories outside the repo that the sensor is inside. What would the ideal API look like in that case?

clairelin135 (approved):
This looks great to me, thanks for making this change!

jamiedemaria:
@sryza if the AssetSelection can handle other repos and source assets then i think eventually the end API would probably be only accepting an AssetSelection. Pros of that would be that you can programmatically create the sets of assets, but the con would be forcing the user to read about another concept/API before they can use the sensor.

python_modules/dagster/dagster/_core/definitions/decorators/sensor_decorator.py
```diff
@@ -214,7 +214,8 @@ def multi_asset_sensor(
 
     Args:
         asset_keys (Optional[Sequence[AssetKey]]): The asset keys this sensor monitors. If not
-            provided, asset_selection argument must be provided.
+            provided, asset_selection argument must be provided. To monitor SourceAssets, you must
```
sryza, line 217: @jamiedemaria sorry I had been meaning to comment on this. I think it would be more accurate to say "to monitor assets that aren't defined in the repo that this sensor is part of" or something like that. Not all of these will be SourceAssets.  I also think we should clarify that asset_selection only works within the same repo.

PR #9955, opened 2022-10-10
[dagster-io/ui] Test and comments for middle truncation search

hellendag:
### Summary & Motivation

Add a Jest test for the middle truncation binary search, along with a heavy dose of comments to explain what's going on.

Buildkite

### DIFF
--- a/js_modules/dagit/packages/ui/src/components/calculateMiddleTruncation.tsx
+++ b/js_modules/dagit/packages/ui/src/components/calculateMiddleTruncation.tsx
@@ -1,0 +1,31 @@
+/**
+ * Binary search to find the maximum middle-truncated text that will fit within the specified target width.
+ * The truncation will occur in the center of the string, with the same number of characters on either side.
+ */
+export const calculateMiddleTruncation = (
+  text: string,
+  targetWidth: number,
+  measure: (value: string) => number,
+): string => {
+  // Skip the search if the text will already fit as-is.
+  if (measure(text) <= targetWidth) {
+    return text;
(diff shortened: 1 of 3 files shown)


js_modules/dagit/packages/ui/src/components/calculateMiddleTruncation.tsx
```diff
@@ -0,0 +1,31 @@
...
+ */
+export const calculateMiddleTruncation = (
+  text: string,
+  targetWidth: number,
+  measure: (value: string) => number,
+): string => {
+  // Skip the search if the text will already fit as-is.
+  if (measure(text) <= targetWidth) {
+    return text;
+  }
```
hellendag, line 13: This is also new. Go ahead and measure the given text up front, and skip the truncation if we don't need it.

bengotow (approved):
This looks good to me!

One random drive-by thought -- I wonder if we could detect that the font is monospace and only measure once? I think for our monospace font, drawing any string onto the canvas and dividing length by character count would be a nice short-cut to avoid binary search.

That said we're not using the monospace font in too many places now and if this is primarily in virtualized tables it's probably not worth it!

I did finally remember why we wrote that split point logic for the Gantt charts - I think we were trying to avoid this scenario: `job_h...b_dev`, where the truncation leaves awkward bits of multiple name components instead of truncating just one of them eg  `job_hel...dev`. If it turns out that nobody notices the difference, it might actually be good to delete that logic since it's complicated and I think it was just a complaint on Slack once!

PR #9976, opened 2022-10-11
Add hook for launch pad root execution button

salazarm:
### Summary & Motivation

Adds a context that can be provided by cloud to override the default launch pad root execution button. 

Used this in cloud.
Screen Shot 2022-10-11 at 5 01 56 PM
Screen Shot 2022-10-11 at 5 02 01 PM

### DIFF
--- a/js_modules/dagit/packages/core/src/launchpad/LaunchpadHooksContext.tsx
+++ b/js_modules/dagit/packages/core/src/launchpad/LaunchpadHooksContext.tsx
@@ -1,0 +1,23 @@
+import React from 'react';
+
+import {LaunchRootExecutionButton} from './LaunchRootExecutionButton';
+
+type LaunchpadHooksContextValue = {
+  LaunchRootExecutionButton?: typeof LaunchRootExecutionButton;
+};
+
+export const LaunchpadHooksContext = React.createContext<LaunchpadHooksContextValue>({
+  LaunchRootExecutionButton: undefined,
+});
+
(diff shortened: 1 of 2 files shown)


hellendag (approved):

js_modules/dagit/packages/core/src/launchpad/LaunchpadHooksContext.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+type LaunchpadHooksContextValue = {
+  LaunchRootExecutionButton?: typeof LaunchRootExecutionButton;
+};
+
+export const LaunchpadHooksContext = React.createContext<LaunchpadHooksContextValue>({
+  LaunchRootExecutionButton: undefined,
+});
+
+export function useLaunchPadHooks(): {
```
bengotow, line 13: This looks good to me! But I'm not super clear on what this hook does other than hide `useContext(LaunchpadHooksContext)`?  Might be nice future proofing!

bengotow (approved):

js_modules/dagit/packages/core/src/launchpad/LaunchpadHooksContext.tsx
salazarm: Yeah it just does `useContext` and returns either the original or an override if provided. I thought I  might need to reuse this logic since there other execution buttons so I put it in this hook

PR #9978, opened 2022-10-11
[cacheable-assets] Enable use of with_resources, with_prefix_or_group on cacheable assets

benpankow:
## Summary

Follows up on the `CacheableAssetsDefinition` introduced in #9761 by allowing them to be bound resources at repo creation time. Also implements `with_prefix_or_group` as a utility fn on the `CacheableAssetsDefinition` object.

Implements `ResourceAddable` on `CacheableAssetsDefinition` so that it can be bound alongside other assets, e.g.

```python
my_assets = with_resources([
    my_asset_1,
    my_asset_2,
    my_cacheable_assets,
  ],
  {"foo": bar}
)
```

Both of these operations produce a `WrappedCacheableAssetsDefinition`, which simply holds the underlying `CacheableAssetsDefinition` alongside a wrapper fn which is applied to all generated assets, at definition build time.

## Test Plan

Introduced unit tests.


benpankow:
* master
  * **PR #9978** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9978" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9978?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
```diff
@@ -104,3 +106,56 @@ def build_definitions(
...
+    the assets themselves are created.
+    """
+
+    def __init__(
+        self,
+        unique_id: str,
+        wrapped: CacheableAssetsDefinition,
+        fn: Callable[[AssetsDefinition], AssetsDefinition],
+    ):
+        super().__init__(unique_id)
```
OwenKephart, line 151: This unique_id doesn't need to be passed in separately because it should always be a function of wrapped.unique_id (currently, we're just copying it in place but we could consider hashing in a constant value or something)

OwenKephart (dismissed):

OwenKephart:
pretty cool, this will be super helpful -- had some changes in mind, happy to chat offline as well

python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
```diff
@@ -104,3 +106,56 @@ def build_definitions(
...
+        self._wrapped = wrapped
+        self._fn = fn
+
+    def compute_cacheable_data(self) -> Sequence[AssetsDefinitionCacheableData]:
+        return self._wrapped.compute_cacheable_data()
+
+    def build_definitions(
```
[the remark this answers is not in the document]
benpankow: restructured using classes instead

python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
```diff
@@ -104,3 +111,161 @@ def build_definitions(
...
+        self, data: Sequence[AssetsDefinitionCacheableData]
+    ) -> Sequence[AssetsDefinition]:
+        return [self.modify_assets(assets) for assets in self._wrapped.build_definitions(data)]
+
+    @abstractmethod
+    def modify_assets(self, assets: AssetsDefinition) -> AssetsDefinition:
```
OwenKephart, line 161: maybe `def transformed_assets_def(self, assets_def: AssetsDefinition)`

two reasons, first assets -> assets_def to line up w/ other places where a singular AssetsDefinition parameter is referred to as "assets_def", second reason is that technically we're not really "modifying" the assets, as we're creating new copies of them.

OwenKephart: `get_transformed_assets_def` also works

OwenKephart (approved):
some small points, but overall this looks good

the hashing bit looks painful but better safe than sorry  🤷

PR #9979, opened 2022-10-11
Fix materialization count by partition

clairelin135:
Previously, the SQL event log storage `get_materialization_count_by_partition` method did not filter out asset observations. This PR amends this behavior to only search for asset materializations.

Internal PR: https://github.com/dagster-io/internal/pull/3754


prha (approved):

prha:
Although, looks like we're still getting a failure in the internal build?

clairelin135:
@prha I think the internal build should succeed once the internal PR is merged!

PR #9980, opened 2022-10-12
Fix incorrect secrets_tags docs

gibsondan:
Summary:
I incorrectly assumed that secrets_tag secrets were kv pairs. Update the agent docs and add a clarification in the OSS docs.


jmsanders (approved):
We could even put a few examples if we want. But I think this is fine.

PR #9985, opened 2022-10-12
do nothing on snapshot add conflict

alangenfeld:
In a condition where snapshot adds race - simply do nothing since they should be attempting to store the same data twice instead of raising.

added test


alangenfeld:
* master
  * **PR #9985** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9985" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/9985?utm_source=stack-comment).

PR #10001, opened 2022-10-13
[Multi-dimensional partitions 1/n] Add schema migration for asset event tags table

clairelin135:
In order to enable multi-dimensional partitioning (and other new asset features) this PR creates an asset event tags table. This table functions the same as the run tags table, by storing a foreign key to the asset event ID and key value string pairs for each tag.

For multi-dimensional partitioning, we plan to add a tag per dimension for each asset materialization. So a multi-dimensional partitions definition with a country dimension and date dimension could have a materialization with two tags `"dagster/partition/country": "US"` and `"dagster/partition/date": "10/13/2022"`. This will enable searching for materializations with just one axis set to a fixed value, for example, all of the materializations where the country is `"US"`.

Stacked diff containing read-write methods from asset event tags table
- https://github.com/dagster-io/dagster/pull/10056


python_modules/dagster/dagster/_core/storage/event_log/schema.py
```diff
@@ -52,6 +52,15 @@
     db.Column("create_timestamp", db.DateTime, server_default=get_current_timestamp()),
 )
 
+AssetEventTagsTable = db.Table(
+    "asset_event_tags",
+    SqlEventLogStorageMetadata,
+    db.Column("id", db.Integer, primary_key=True, autoincrement=True),
```
prha, line 58: probably this should be BigInteger?

python_modules/dagster/dagster/_core/storage/event_log/schema.py
```diff
@@ -52,6 +52,15 @@
...
+    db.Column("event_id", db.Integer, db.ForeignKey("event_logs.id", ondelete="CASCADE")),
+    db.Column("key", db.Text),
+    db.Column("value", db.Text),
```
prha, line 61: should we have asset_key in here?  Might make it easier to do stat queries without doing joins?

prha (commented):
Would love to see a stacked diff that actually adds the storage methods to read/write from this table.

clairelin135:
@prha Makes sense, I can add the stacked diff!

python_modules/dagster/dagster/_core/storage/event_log/schema.py
clairelin135: Makes sense. We'll be able to find out how many multi-dimensional partitions are filled by just querying this table, since we're only allowing tags on asset materializations at the moment

python_modules/dagster/dagster/_core/storage/event_log/schema.py
```diff
@@ -75,3 +91,6 @@
     SqlEventLogStorageTable.c.id,
     mysql_length={"dagster_event_type": 64},
 )
+db.Index(
+    "idx_asset_event_tags", AssetEventTagsTable.c.key, AssetEventTagsTable.c.value, mysql_length=64
+)
```
alangenfeld, line 96: confident this is what we need? i dont have context on expected access patterns

alangenfeld: is `64` the right length? Im not sure how those work or what restrictions we put on the contents of keys/values

alangenfeld (commented):
would be good to include a test in the cross impl test suite 
https://github.com/dagster-io/dagster/blame/master/python_modules/dagster/dagster_tests/core_tests/storage_tests/utils/event_log_storage.py
to ensure all the impls work as expected

python_modules/dagster/dagster/_core/storage/event_log/schema.py
```diff
@@ -52,6 +53,21 @@
...
+AssetEventTagsTable = db.Table(
+    "asset_event_tags",
+    SqlEventLogStorageMetadata,
+    db.Column(
+        "id",
+        db.BigInteger().with_variant(sqlite.INTEGER(), "sqlite"),
+        primary_key=True,
+        autoincrement=True,
+    ),
+    db.Column("event_id", db.Integer, db.ForeignKey("event_logs.id", ondelete="CASCADE")),
```
alangenfeld, line 65: hm, this may make future likely big int migration for event id a bit more challenging. Can we proactively make it bigint or does it have to match?

not having it covered by an index may make impact delete performance, especially as the table grows

alangenfeld (approved):
Given the limited context i have this seems likely fine. Storage changes are always scary as they are costly to correct.

This shouldn't land until all the internal changes are ready to land as well

python_modules/dagster/dagster/_core/storage/event_log/schema.py
clairelin135: I think in the primary views that we'd like to support are all in terms of a specific asset, i.e. showing how many partitions exist/are missing, so I think I should change this to index by asset key.

As of now, the tags are system populated. They'll be of the format `dagster/partition/dimension_name`. I'm not really sure how these index lengths work, but as as long as the user-specified dimension name doesn't exceed 46 characters we should be able to index properly

python_modules/dagster/dagster/_core/storage/event_log/schema.py
clairelin135: Hmm, I tried making this value a big int. It works for postgres, but MySQL requires the types to be the same. I'll keep it as a regular int for now, unless if you'd rather I make it a big int for postgres and not MySQL

What kind of index were you imagining? I noticed that we don't have one for the run tags table, perhaps we can wait until we need the index to add it?

prha (approved):

python_modules/dagster/dagster/_core/storage/alembic/versions/033_add_asset_event_tags_table.py
```diff
@@ -0,0 +1,64 @@
...
+            db.Column("asset_key", db.Text, nullable=False),
+            db.Column("key", db.Text, nullable=False),
+            db.Column("value", db.Text),
+        )
+
+    if not has_index("asset_event_tags", "idx_asset_event_tags"):
+        op.create_index(
+            "idx_asset_event_tags",
+            "asset_event_tags",
+            ["asset_key", "key", "value"],
```
prha, line 42: I wonder if we need `id` in this index, since we're usually ordering by id

prha: https://github.com/dagster-io/dagster/pull/10056

clairelin135: Hmm, so we should make this index `['asset_key', 'key', 'value', 'event_id']?

python_modules/dagster/dagster/_core/storage/alembic/versions/033_add_asset_event_tags_table.py
```diff
@@ -0,0 +1,63 @@
...
+            db.Column("asset_key", db.Text, nullable=False),
+            db.Column("key", db.Text, nullable=False),
+            db.Column("value", db.Text),
+        )
+
+    if not has_index("asset_event_tags", "idx_asset_event_tags"):
+        op.create_index(
+            "idx_asset_event_tags",
+            "asset_event_tags",
+            ["asset_key", "key", "value", "event_id"],
```
prha, line 41: I think you can probably drop`event_id` here (though you should confirm with an explain analyze)... at this point, you're kind of mirroring the whole table in the index.

clairelin135: Yeah, there's a pretty negligible difference with/without the `event_id` field. I'll drop it

PR #10003, opened 2022-10-13
Fix example scaffold

smackesey:
### Summary & Motivation

I generated a new example for asset versioning and discovered that the example scaffold code was outdated in several ways. This updates the code:

- Use job etc instead of legacy abstractions
- Don't try to add the example to a JSON index that no longer exists
- Use a Python package structure based on recent example overhaul by @yuhan 

Also, this PR deletes `pyproject.toml` from all examples. The reason for this is that it interferes with the `pyproject.toml` in the repo root (that's why our other packages don't have it), which means you have to duplicate black/isort config to get correct behavior.

Generated a new test example (1) verified it looked ok; (2) ran tests; (3) loaded it in dagit.


smackesey:
* master
  * **PR #10003** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10003" target="_blank">Graphite</a>  👈
    * **PR #10005** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10005" target="_blank">Graphite</a> 
      * **PR #9899** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9899" target="_blank">Graphite</a> 
    * **PR #10004** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10004" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10003?utm_source=stack-comment).

python_modules/automation/automation/scaffold/assets/dagster-example-tmpl/tox.ini.tmpl
```diff
@@ -1,13 +1,14 @@
...
 skipsdist = True
 
 [testenv]
 setenv =
-  VIRTUALENV_PIP=21.3.1
+  VIRTUALENV_PIP=22.1.2
 passenv = CI_* COVERALLS_REPO_TOKEN BUILDKITE
 deps =
   -e ../../python_modules/dagster[mypy,test]
+  -e ../../python_modules/dagit
```
yuhan, line 11: why do we need dagit as a test dep?

yuhan (changes requested):
the newest / up-to-date scaffold cmd is in https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_cli/project.py and https://github.com/dagster-io/dagster/tree/master/python_modules/dagster/dagster/_generate - (doc: https://docs.dagster.io/getting-started/create-new-project). 

i think what we should do are
- remove all the scaffold cmd in automation - they are super outdated and i don't think users are using it.
- update contributing doc if that scaffold command is mentioned anywhere, to use https://docs.dagster.io/getting-started/create-new-project instead
- make the pyproject.toml changes in https://github.com/dagster-io/dagster/tree/master/python_modules/dagster/dagster/_generate

smackesey:
Ah I'd forgotten about the new one. Is that the template used in the new examples after your overhaul? Agree we should remove the automation example scaffold code in that case.

>remove all the scaffold cmd in automation - they are super outdated and i don't think users are using it.

Users definitely aren't using any of it because it's not even published. I do wonder if we should keep library generation code around, but probably not since the scaffold gets outdated anyway.

python_modules/automation/automation/scaffold/assets/dagster-example-tmpl/tox.ini.tmpl
smackesey: Can't remember my reasoning but now it's a moot point if we're deleting this code.

yuhan:
> We just need to remember to remove the pyproject.toml when creating a new example. Are there internal instructions for creating a new example somewhere?

we'll use the `dagster project scaffold` to create new examples. there's no internal specific instructions.

yuhan (approved):

PR #10005, opened 2022-10-13
Type annotations in dagster-graphql

smackesey:
### Summary & Motivation

Adds type annotations for dagster-graphql implementation code. Most of it is trivial, but I also added a `WithContext` protocol that can be used to type the `graphql_info` first arg to resolvers. `WithContext` types the `context` property, which gives us type-checking for implementation logic (`graphql_info.context` is the entry point to interaction with the DB/instance).

mypy/pyright


alangenfeld (approved):
lgtm - just make sure its passing before land (internal sync too)

PR #9799, opened 2022-09-26
[managed stacks][wip] Prototype managed stack API

benpankow:

PR #10011, opened 2022-10-13
[dagster-managed-elements] Dagster-managed-elements CLI, APIs

benpankow:
## Summary

Introduces a small `dagster-managed-elements` package. This will serve as an optional dependency for integration libraries - if installed it can let users manage various 'elements' (external data tools, e.g. Airbyte, Fivetran). This enables a Pythonic way to configure these tools as code, which can neatly coexist with but is not tied to Dagster itself.

These APIs were separated into their own package (and refactored significantly from #9799) so that they can be iterated on (or removed) without touching core Dagster.

The package includes:

- `ManagedElementReconciler`, an abstract base class which has two methods that a user must implement: `check` and `apply`. The former returns a diff of config and actual state, e.g. what will happen if the config is applied. The latter actually attempts to reconcile the config with some external service (the 'element').
- A serializable `ManagedElementDiff` utility class, which tracks a diff between a config and actual state. Can be used in logic (e.g. is a diff empty?) and for displaying a human-readable diff to the user (e.g. what will change if I apply).
- A utility method to construct a `ManagedElementDiff` from two dicts.
- A `dagster-managed-elements` CLI, which contains two commands, `check` and `apply`. Each command requires the name of a Python module which acts as the entrypoint, which is searched for `ManagedElementReconciler` objects at the top level. 
  
  ```bash
  dagster-managed-elements check -m my_entrypoint_module
  ```
  
  Users can optionally specify a path to a directory to add to the module load path:
  
  ```bash
  dagster-managed-elements check -m my-my_entrypoint_module -d ~/Documents/working_directory_containing_module
  ```
  
  They can also explicitly specify one or more reconcilers by-attr to check, instead of the top-level search:
  
  ```bash
  dagster-managed-elements check -m my_entrypoint_module:my_reconcilers.foo,my_reconcilers.bar
  ```
  
  The `check`/`apply` operation is performed on each discovered reconciler, with output dumped to the console.

## Naming

The top-level interactions `check` and `apply`, with the underlying `diff` object referring to the difference between desired and actual state, were used to coincide with similar reconciliation tools. `apply` and `diff` are fairly standard verbiage across such tools. `check` is not standardized, and tends to vary across tools.

- Kubernetes uses `apply` to reconcile config with actual state, using `diff` or `dry-run` to see the changes that would be applied.
- Terraform uses `apply` to reconcile config with actual state and `plan` to see the diff that would be applied. The changes rendered by both commands are referred to as the `diff`.
- When React rerenders, it *diffs* elements to determine if they need to be rerendered or not.

Here, the motivation for using `check` over `plan` is in intention - while `terraform plan` is typically used to validate changes that a user plans to apply, `dagster-managed-elements check` is meant to be used in a wider array of contexts, including to check whether configuration is in sync before e.g. kicking off a run. The intent is not always to plan a change but rather to verify or validate that config has remained the same.

## Test Plan

Unit tests.


sryza:
I like this design.

Re: naming, I'm sorry to be such a debbie downer on this one, but I don't think we should call individual integrations "stacks". It's like if we started to call individual columns in a database "tables" - it adds unnecessary confusion by using a term that people already understand in a different way than they understand it.

Maybe `RemoteDefinitionsReconciler`? `ManagedDefinitionsReconciler`?  `ManagedServiceDefinitions`?  I'm also open to coining a new, sexy term - I just prefer one that doesn't already have an existing, conflicting meaning in the domain.

benpankow:
Updated the PR (and stacked PR) to use `element` instead of `stack`, for now.

python_modules/libraries/dagster-managed-elements/dagster_managed_elements/cli.py
```diff
@@ -0,0 +1,89 @@
...
+def get_reconcilable_objects(module: ModuleType) -> List[ManagedElementReconciler]:
+    """
+    Collect all ManagedElementReconciler-implementing objects in the root of the
+    module.
+    """
+    return [
+        getattr(module, obj)
+        for obj in dir(module)
+        if isinstance(getattr(module, obj), ManagedElementReconciler)
+    ]
```
alangenfeld, line 39: what led to this approach? Messing with imports can get spooky, might be nice to avoid if possible.

python_modules/libraries/dagster-managed-elements/dagster_managed_elements/types.py
```diff
@@ -0,0 +1,209 @@
...
+        Returns whether the user provided config for the managed element is in sync with the external resource.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def apply(self) -> ManagedElementCheckResult:
+        """
+        Reconciles the managed element with the external resource, returning the applied diff.
+        """
+        raise NotImplementedError()
```
alangenfeld, line 209: I think it would be good to cross reference whatever other reconciliation based systems you would compare this approach to in like a talk about it (terraform, react, k8s, etc) and include in the summary why you choose the language you did. I am not sure how all those systems handle their reconciliation APIs but if theres something common or a particular system this feels closest to, might be good to align against that language.

python_modules/libraries/dagster-managed-elements/dagster_managed_elements/cli.py
alangenfeld, line 39: what led to this set-up? Messing with imports can be spooky so might be nice to avoid if possible. Main spooky elements:
* working directory mattering a lot
* weird corner case behaviors related to how python import system works 

At minimum would be good to layer how things get built up so you don't have to import a whole file for any test or do other shenanigans for complex tests. More sophisticated users may also want to be able to use a cleaner entry point than crawling all module scope things and forcing things be defined at init time.

benpankow: Yeah, this setup is a bit rough at the moment. Originally, the CLI was meant to refer to/operate on a Dagster repository as an entry point, but there were concerns around tying the system to Dagster machinery, so it was suggested to use an adjacent file. This was my attempt to create a very barebones version of the repo loading logic.

I'll make the suggested refactor to make testing easier.

> More sophisticated users may also want to be able to use a cleaner entry point than crawling all module scope things and forcing things be defined at init time.

Do you have any thoughts on a different approach we could use? An entrypoint as sophisticated as the Dagster repository would be great, but is there a better middle-ground for the initial version that we dogfood/test with?

benpankow: I've updated the PR to path to reconcilers a bit differently:

```bash
dagster-managed-elements check -m my_entrypoint_module
```

Users can optionally specify a path to a directory to add to the module load path:

```bash
dagster-managed-elements check -m my-my_entrypoint_module -d ~/Documents/working_directory_containing_module
```

In the vein of uvicorn, they can also specify one or more reconcilers by-attr to check specifically:

```bash
dagster-managed-elements check -m my_entrypoint_module:my_reconcilers.foo,my_reconcilers.bar
```

benpankow: marking resolved, continuing discussion in thread below

python_modules/libraries/dagster-managed-elements/dagster_managed_elements/types.py
benpankow: I've added a bit more explanation to the PR summary above, and settled on `plan` and `apply` as top-level action verbs (each of which returns a `diff`).

benpankow: I've added a bit of a write-up above as to why I settled on the naming that's in the PR at present.

I don't have very strong feelings, so would be open to changing `check` towards `diff` or `plan`. 

My slight preference towards `check` is that it expresses the intent behind the action more clearly than `plan`, and to avoid overloading with the noun-form `diff`. This operation is less standard between other state reconciliation tools, so it feels ok to me.

python_modules/libraries/dagster-managed-elements/dagster_managed_elements/types.py
```diff
@@ -0,0 +1,209 @@
...
+        cls,
+        additions: Optional[List[Tuple[str, str]]] = None,
+        deletions: Optional[List[Tuple[str, str]]] = None,
+        modifications: Optional[List[Tuple[str, str, str]]] = None,
+    ):
+        return super().__new__(
+            cls,
+            additions or [],
+            deletions or [],
+            modifications or [],
```
yuhan, line 51: should the `Tuple[str, str]` be some type like `DiffData` to ensure it's robust? it seems lots of the diff check logic (and therefore the reconciliation as well) will depend on these lists of tuples.

user-facing wise, `str, str` is good, but in here, we probably want to 1) `dagster.check` the types at run time, 2) better type/schematize the user inputs.

benpankow: added a type alias, and a bunch of `check` checks 👍

alangenfeld (commented):
`dagster-managed-elements` cli entry is pretty long, tough to make that terse without getting clever. `dagter-man` ? `dagman` ? `dagster-me`? 

What are the API stability requirements out the gate here? Unless its surprisingly high or  were not happy with the stacked PRs that use this, i think this is in a good spot to land and iterate from.

python_modules/libraries/dagster-managed-elements/dagster_managed_elements/cli.py
```diff
@@ -0,0 +1,142 @@
...
+            click.echo(result)
+    return diff
+
+
+def apply(module_dir: str, import_str: str) -> ManagedElementDiff:
+    reconcilable_objects = get_reconcilable_objects_from_module(
+        module_dir=module_dir, import_str=import_str
+    )
+
+    click.echo(f"Found {len(reconcilable_objects)} stacks, applying...")
```
alangenfeld, line 86: outdated `stacks` language

benpankow:
> `dagster-managed-elements` cli entry is pretty long, tough to make that terse without getting clever. `dagter-man` ? `dagman` ? `dagster-me`?

I'm happy with `dagman` or `dagster-me` as a shortcut in addition to the long-form CLI name. I'll probably impl `dagster-me` unless folks have other ideas.

> What are the API stability requirements out the gate here? Unless its surprisingly high or were not happy with the stacked PRs that use this, i think this is in a good spot to land and iterate from.

My understanding is that the stability requirements are low for now - this is an experimental feature and one we plan on iterating on pretty heavily depending on user feedback.

python_modules/libraries/dagster-managed-elements/dagster_managed_elements/types.py
```diff
@@ -0,0 +1,237 @@
...
+        deletions: Optional[List[DiffData]] = None,
+        modifications: Optional[List[ModifiedDiffData]] = None,
+    ):
+        check.opt_list_param(additions, "additions", of_type=DiffData)
+        check.opt_list_param(deletions, "deletions", of_type=DiffData)
+        check.opt_list_param(modifications, "modifications", of_type=ModifiedDiffData)
+
+        return super().__new__(
+            cls,
+            additions or [],
```
yuhan, line 59:
```suggestion
            check.opt_list_param(additions, "additions", of_type=DiffData),
```
`opt_list_param` does this for you.
https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_check/__init__.py#L747

yuhan (commented):
left a comment about `check.opt_*` usage

yuhan (approved):
approved - i believe this will be super experimental, so i dont think the cli naming is blocking, but im supportive to come up with something shorter.

alangenfeld (approved):

PR #9998, opened 2022-10-12
[dagster-airbyte] Cache loading of Airbyte YAML files

benpankow:
## Summary

Some consolidation + cleanup of the `dagster-airbyte` impl, including moving `load_from_airbyte_project` to a `CacheableAssetsDefinition` so the YAML files are only read at initial repo load time.

This also lets us deduplicate a lot of the code between `load_from_airbyte_instance` and `load_from_airbyte_project`.

## Test Plan

Existing unit tests.

PR #9939, opened 2022-10-07
[dagster-airbyte] Generate schema metadata at load time when loading from project or instance

benpankow:
## Summary

Generates schema metadata for Airbyte asset definitions when loading from a project or instance, so that users can browse output tables & their schemas from Dagit without having to first materialize an asset.

## Test Plan

Added unit test, tested locally.

PR #10013, opened 2022-10-13
[dagster-airbyte] Airbyte managed elements impl

benpankow:
## Summary

Initial implementation of user-facing APIs and reconciliation logic for Airbyte managed config, based on the APIs in #10011. 

This PR allows users to specify `AirbyteConnection` objects to a reconciler, which creates/updates/deletes connections, sources, and destinations against a live Airbyte instance. This has no direct ties to Dagster itself (other than using the resource system), but the `AirbyteConnection` objects can be ingested to create assets via #10055

## Sample

```python

airbyte_instance = airbyte_resource.configured(
    {
        "host": os.getenv("AIRBYTE_HOSTNAME", "localhost"),
        "port": os.getenv("AIRBYTE_PORT", "80"),
    }
)

local_json_source = AirbyteSource(
    name="local-json-input",
    source_type="File",
    source_configuration={
        "url": "/local/sample_file.json",
        "format": "json",
        "provider": {"storage": "local"},
        "dataset_name": "my_data_stream",
    },
)

local_json_destination = AirbyteDestination(
    name="local-json-output",
    destination_type="Local JSON",
    destination_configuration={"destination_path": "/local/destination_file.json"},
)

local_json_conn = AirbyteConnection(
    name="local-json-conn",
    source=local_json_source,
    destination=local_json_destination,
    stream_config={
        "my_data_stream": AirbyteSyncMode.FULL_REFRESH_APPEND,
    },
    normalize_data=False,
)

reconciler = AirbyteManagedElementReconciler(
    airbyte=airbyte_instance,
    connections=[
        local_json_conn,
    ],
)
```

## Test Plan

New integration test suite which spins up Airbyte and tests reconciliation against the fresh instance.

Right now, the test suite is pretty simplistic, checking the diff from a fresh instance, creating resources, then tearing them down.

Subsequent PR will implement more sophisticated integration tests.


benpankow:
* master
  * **PR #9937** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9937" target="_blank">Graphite</a> 
    * **PR #9939** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9939" target="_blank">Graphite</a> 
      * **PR #9978** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9978" target="_blank">Graphite</a> 
        * **PR #9998** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9998" target="_blank">Graphite</a> 
          * **PR #10011** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10011" target="_blank">Graphite</a> 
            * **PR #10013** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10013" target="_blank">Graphite</a>  👈
      * **PR #9800** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9800" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10013?utm_source=stack-comment).

python_modules/libraries/dagster-airbyte/dagster_airbyte/resources.py
```diff
@@ -63,6 +69,33 @@ def api_base_url(self) -> str:
             + "/api/v1"
         )
 
+    @contextmanager
+    def cache_requests(self):
```
OwenKephart, line 73: what's the methodology for determining which requests to cache? Seems like this is currently only used when getting destination definition, but not when getting source definitions.

python_modules/libraries/dagster-airbyte/dagster_airbyte/managed/reconciliation.py
```diff
@@ -0,0 +1,510 @@
...
+
+        if configured_source:
+            defn_id = check.not_none(
+                res.get_source_definition_by_name(configured_source.source_type, workspace_id)
+            )
+            base_source_defn_dict = {
+                "name": configured_source.name,
+                "connectionConfiguration": configured_source.source_configuration,
+            }
+            source_id = "TBD"
```
OwenKephart, line 126: nit: maybe make this a bit more explicit, like "<NOT YET CREATED>", just to visually distinguish it (also applies to the destination_id in the other function)

python_modules/libraries/dagster-airbyte/dagster_airbyte/managed/reconciliation.py
```diff
@@ -0,0 +1,510 @@
...
+            diff_destinations(
+                configured_destination,
+                existing_destination.destination if existing_destination else None,
+            )
+        )
+
+        if not configured_destination and existing_destination:
+            initialized_destinations[destination_name] = existing_destination
+            if not dry_run:
+                res.make_request(
```
OwenKephart, line 189: this part is a bit scary -- are we confident that all users of this interface will want Dagster to manage all of their Airbyte connections? Worried about a poor user who hooks this stuff up to their prod airbyte instance to try it out and it deletes a bunch of stuff automatically.

I'd prefer to just leave the irrelevant connections alone or put this behavior behind a flag or something if possible. Willing to hear other opinions though

python_modules/libraries/dagster-airbyte/dagster_airbyte/managed/reconciliation.py
```diff
@@ -0,0 +1,510 @@
...
+                )["operationId"]
+        elif normalization_config is True:
+            raise Exception(
+                f"Destination {destination.destination.name} does not support normalization."
+            )
+
+    return None
+
+
+def reconcile_connections_pre(
```
OwenKephart, line 361: similar comment about deleting stuff here, but regardless of that discussion, I think the sequencing here is a bit confusing.

We use this function to generate a diff between the config state and the instance state (and delete unmanaged connections), but then a separate function (reconcile_connections_post), which also has to essentially calculate a diff itself, to apply non-deletion changes.

Would prefer if this functionality was combined into a single function (basically making a single call to /connections/list, then delegating out to functions like delete_connection or update_connection depending on the diff.

python_modules/libraries/dagster-airbyte/dagster_airbyte/managed/types.py
```diff
@@ -0,0 +1,182 @@
...
+    @classmethod
+    def from_api_json(cls, api_json: Dict[str, Any]):
+        return cls(
+            AirbyteDestination(
+                api_json["name"],
+                api_json["destinationName"],
+                api_json["connectionConfiguration"],
+            ),
+            api_json["destinationId"],
+            "",
```
OwenKephart, line 92: in what cases is the destination definition id available? also, any reason to have this be an empty string rather than None?

OwenKephart (dismissed):
this is really cool functionality -- had a bunch of smaller nits as well as some questions

python_modules/libraries/dagster-airbyte/dagster_airbyte/managed/reconciliation.py
benpankow: That seems reasonable to me. I'll add a flag to the reconciler object (defaults to off) that will tell it to delete sources/destinations/connections.

python_modules/libraries/dagster-airbyte/dagster_airbyte/managed/types.py
benpankow: This isn't available when we pull from the API directly, we'd have to query it separately (& we don't end up using it in this case). Will change to `None` since it's more explicit.

python_modules/libraries/dagster-airbyte/dagster_airbyte/resources.py
benpankow: Anything involving source/destination definitions should be very safe since they won't change during a reconciliation run, I've gone ahead and done a pass to update all of those.

We can probably expand this down the line, but some fields might change during reconciliation itself (e.g. sources/destinations getting updated).

python_modules/libraries/dagster-airbyte/dagster_airbyte/managed/reconciliation.py
benpankow: The ordering is pretty important here, basically we need to:

- `reconcile_connections_pre`: Delete all connections which depend on sources or destinations that are being recreated or deleted. The underlying sources/destinations can't be deleted unless this happens first.
- `reconcile_sources/reconcile_destinations`: Update all sources and destinations, including deletion/recreation.
- `reconcile_connections_post`: (Re)create connections based on new sources and destinations.

I've added a few more comments which hopefully make this a bit more clear. If you can think of a better way to structure this, happy to take another look.

OwenKephart: got it -- didn't realize that airbyte wouldn't let you delete a source with dependent connections (makes sense)

I think this is a reasonable way to structure things

OwenKephart (approved):
👍 🤠 👍

python_modules/libraries/dagster-airbyte/dagster_airbyte/managed/types.py
```diff
@@ -0,0 +1,189 @@
...
+
+class AirbyteDestination:
+    """
+    Represents a user-defined Airbyte destination.
+    """
+
+    def __init__(self, name: str, destination_type: str, destination_configuration: Dict[str, Any]):
+        self.name = name
+        self.destination_type = destination_type
+        self.destination_configuration = destination_configuration
```
yuhan, line 64: same comment about dagster.check

python_modules/libraries/dagster-airbyte/dagster_airbyte/managed/reconciliation.py
```diff
@@ -0,0 +1,589 @@
...
+        Args:
+            airbyte (ResourceDefinition): The Airbyte resource definition to reconcile against.
+            connections (Iterable[AirbyteConnection]): The Airbyte connection objects to reconcile.
+            delete_unmented_resources (bool): Whether to delete resources that are not mentioned in
+                the set of connections provided. When True, all Airbyte instance contents are effectively
+                managed by the reconciler. Defaults to False.
+        """
+        self._airbyte_instance: AirbyteResource = airbyte(build_init_resource_context())
+        self._connections = list(connections)
+        self._delete_unmentioned_resources = delete_unmented_resources
```
yuhan, line 571: same - run time checks

yuhan (commented):
i left some comments about checks to guard bad user inputs

python_modules/dagster-test/dagster_test/fixtures/docker_compose.py
```diff
@@ -12,15 +12,12 @@
 
 @contextmanager
 def docker_compose_cm(
-    docker_compose_yml,
-    network_name=None,
-    docker_context=None,
-    service=None,
+    docker_compose_yml, network_name=None, docker_context=None, service=None, env_file=None
```
jmsanders, line 15: Nice - so we no longer need to monkeypatch env in the tests?

jmsanders: Because the docker compose test fixtures are kind of...complicated...would you mind adding a unit test for this? Similar to the existing ones but passing an env file?

python_modules/libraries/dagster-airbyte/dagster_airbyte_tests/docker-compose.yml
```diff
@@ -0,0 +1,212 @@
+# Airbyte docker-compose file
+# Sourced from https://github.com/airbytehq/airbyte/blob/master/docker-compose.yml
```
jmsanders, line 2: Wowza - is this the minimal necessary config? Or just a direct copy? Non blocking, but it'd be nice if we were able to pare this down to make it more obvious which levers we're actually pulling and what we're relying on the defaults for.

jmsanders (approved):

python_modules/libraries/dagster-airbyte/dagster_airbyte_tests/docker-compose.yml
benpankow: We may be able to pare it down more, or to just directly pull this file from their GitHub repo instead. We could probably just pull this file and the env file in particular.

PR #9346, opened 2022-08-11
[buildkite] always run repo side steps

alangenfeld:
noticed when black / isort failures on docs only changes missed

### How I Tested These Changes

bk

PR #10016, opened 2022-10-13
Skip Python checks if no Python files change

jmsanders:
Incrementally skip more and more of our build when it isn't relevant.

This skips our three repo-wide Python checks (black, isort, and check-manifests) when there's no change to any python files.

I plan to continue to apply this pattern to more and more of margins of the build (various linters, one-off steps, etc.)

### DIFF
--- a/.buildkite/dagster-buildkite/dagster_buildkite/utils.py
+++ b/.buildkite/dagster-buildkite/dagster_buildkite/utils.py
@@ -186,7 +186,7 @@ def get_commit(rev):
 
 
 def get_changed_files():
-    subprocess.check(["git", "fetch", "origin", "master"])
+    subprocess.call(["git", "fetch", "origin", "master"])
     origin = get_commit("origin/master")
     head = get_commit("HEAD")
     logging.info(f"Changed files between origin/master ({origin}) and HEAD ({head}):")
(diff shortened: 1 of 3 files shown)


.buildkite/dagster-buildkite/dagster_buildkite/utils.py
```diff
@@ -186,4 +188,17 @@ def get_changed_files():
...
+    for path in paths:
+        logging.info(path)
     return [Path(path) for path in paths]
+
+
+def skip_if_no_python_changes():
+    if not is_feature_branch(os.getenv("BUILDKITE_BRANCH")):
+        return None
+
+    if not any(path.suffix == ".py" for path in get_changed_files()):
```
alangenfeld, line 201: we may consider memoizing get_changed_files if it gets called a lot and should be same for process life

alangenfeld (approved):
i did hit an issue with docs only changes a bit ago https://github.com/dagster-io/dagster/pull/9346

but i think that was due to directory based instead of checking for any py file changes

.buildkite/dagster-buildkite/dagster_buildkite/utils.py
jmsanders: The good thing is it only runs on the initial step. So at most a few times. But yeah, memoizing is a good idea.

PR #10024, opened 2022-10-13
[freshness-policy] [1/n] FreshnessPolicy object

OwenKephart:
### Summary & Motivation

This initial forray simply defines FreshnessPolicy objects which may be attached to AssetsDefinition objects. Methods for evaluating these freshness_policies and actually calculating upstream_materialization_timestamp will be added in later commits.

This was designed to enable (but not yet implement) two future directions. Right now, we assume that the upstream asset keys that you're measuring your Freshness in relation to will always be the root assets (i.e. the assets all the way at the top of the asset graph). It will be fairly straightforward to augment this policy object to accept explicit upstream keys that you want to measure your freshness relative to. We also do not make the assumption that your freshness will be always be determined with the same constraint per asset (although that is the most common case). Future work can enable setting a per-upstream asset freshness constraint.


python_modules/dagster/dagster/_core/definitions/asset_data_sla.py
```diff
@@ -0,0 +1,113 @@
...
+    """Specifies a constraint on the timestamp of the root data that was used for the most recent
+    asset materialization of a given asset.
+
+    The root data timestamp is calculated by traversing the history of asset materializations of
+    upstream assets which occured before the current materialization. This gives a lower bound on
+    the most recent records that could possibly be incorporated into the current state of the asset
+    to which this SLA is attached.
+    """
+
+    def is_passing(self, root_data_timestamp: Optional[float]) -> bool:
```
sryza, line 19: For purposes of testing and determining whether SLAs were met in the past, it might be helpful for this to accept a `current_time` argument, like `PartitionsDefinition.get_partitions` does.

python_modules/dagster/dagster/_core/definitions/asset_data_sla.py
```diff
@@ -0,0 +1,113 @@
+from typing import Mapping, Optional
+
+import pendulum
+from croniter import croniter
+
+from .events import AssetKey
+
+
+class RootDataTimeConstraint:
```
sryza, line 9: This is clean. I like it.

sryza: Thoughts on calling it `UpstreamMaterializationTimeConstraint`?
- It's conceivable that someone would want to set a constraint on an asset that's not a root asset.  E.g. if that asset depends on some upstream data but has an implementation that goes out and fetches some external data.
- "Materialization" has a clear meaning in the system, while "data" does not.

python_modules/dagster/dagster/_core/definitions/asset_data_sla.py
```diff
@@ -0,0 +1,113 @@
...
+    the most recent records that could possibly be incorporated into the current state of the asset
+    to which this SLA is attached.
+    """
+
+    def is_passing(self, root_data_timestamps: Mapping[AssetKey, float]) -> bool:
+        pass
+
+
+class UniformConstraintSLA(AssetRootDataSLA):
+    """This base class applies the same constraint to each root asset."""
```
sryza, line 83: We don't currently use the word "root" anywhere, so I think it's important to explain what it means - `SourceAsset` within the same repository of the asset, that's upstream of the asset?

python_modules/dagster/dagster/_core/definitions/asset_data_sla.py
```diff
@@ -0,0 +1,113 @@
...
+    upstream assets which occured before the current materialization. This gives a lower bound on
+    the most recent records that could possibly be incorporated into the current state of the asset
+    to which this SLA is attached.
+    """
+
+    def is_passing(self, root_data_timestamp: Optional[float]) -> bool:
+        pass
+
+
+class StalenessConstraint(RootDataTimeConstraint):
```
sryza, line 23: It might be most clear to call this `StalenessDurationConstraint`?

python_modules/dagster/dagster/_core/definitions/asset_data_sla.py
```diff
@@ -0,0 +1,113 @@
...
+    """This constraint specifies that the root data that was used for the most recent asset
+    materialization must not be more than `allowed_staleness_minutes` old, relative to the most
+    recent cron schedule tick.
+    """
+
+    def __init__(self, cron_schedule: str, allowed_staleness_minutes: float):
+        self._cron_schedule = cron_schedule
+        self._allowed_staleness = pendulum.duration(minutes=allowed_staleness_minutes)
+
+    def is_passing(self, root_data_timestamp: Optional[float]) -> bool:
```
sryza, line 51: Thoughts on floats vs. pendulum timestamps?

python_modules/dagster/dagster/_core/definitions/asset_data_sla.py
sryza: Two more thoughts:
* This should be an abstract base class, right?
* In many (all?) cases, it's possible to calculate how late you are, and I suspect users would be curious about that. It might be useful to return that as well.

OwenKephart: Re: the non-root data times, I think this becomes pretty difficult to calculate, as to track this in the way we're currently doing (which allows O(1) look ups after the fact), you'd need to tag each asset with the data timestamp of every single asset upstream of it (rather than just the roots). Maybe that's actually fine in practice, but it feels pretty sketch to be adding potentially thousands of tags to each asset event.

For that specific scenario, I'd probably model that external thing that's being fetched from as a source asset (but maybe there are other scenarios in which you'd want to depend on a non-root asset's data where that wouldn't work?)

Agree on all other points

sryza: > , as to track this in the way we're currently doing (which allows O(1) look ups after the fact), you'd need to tag each asset with the data timestamp of every single asset upstream of it (rather than just the roots)

I was imagining that the user would specify an SLA like "I don't want asset Y to be materialized later than 20 minutes after asset X", and then we would tag asset Y with the the timestamp of asset X, even if asset X was not a "root" asset.  Would that not work?

python_modules/dagster/dagster/_core/definitions/asset_data_sla.py
OwenKephart: decided to keep things as floats, just to not bake pendulum too deeply into the APIs here. Theoretically, you can use whatever time library you want to calculate these things (although I don't see a reason that would ever be anything other than pendulum).

don't feel strongly about this and can be swayed otherwise

sryza (commented):
I think it might be worth doing another iteration on the naming before we publicize, but that doesn't need to block this.  We should mark these experimental though.

The `current_time` argument that we provide to `get_partition_keys` is a Pendulum datetime, not a float. Thoughts on doing that here? An advantage is that, for someone writing their own FreshnessPolicy, they can look at the hour/minute/whatever without needing to convert it.  Same with the `upstream_materialization_timestamps`.

A couple freshness policies I brainstormed:
* The asset is in violation of policy if it hasn't been updated in the last 20 minutes (i.e. ancestors are not considered)
* The asset is in violation of policy if it hasn't been updated within 20 minutes of the most recent materialization of any of its parents.

Would those be easy to implement? Not requesting that you implement them in this PR, just want to make sure the API works for them.

python_modules/dagster/dagster/_core/definitions/assets.py
```diff
@@ -527,6 +539,24 @@ def get_partition_mapping(self, in_asset_key: AssetKey) -> PartitionMapping:
                 else AllPartitionMapping(),
             )
 
+    def with_freshness_policies(
```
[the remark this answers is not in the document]
OwenKephart: hm fair point -- in my opinion the main utility of this would come from calling this repeatedly on a list of assets (similarly to the with_resources() function vs the AssetsDefinition.with_resources method). So you could load a bunch of assets from_package_module and apply the same freshness policy to all of them. This doesn't seem particularly important though so I'll take this out for now.

OwenKephart:
@sryza good questions, I think both of your examples can be solved with a new (name tbd) "relative_to_assets" property of the FreshnessPolicy, which would be an enum (or whatever) with the following options:

    - SELF: relative to its own most recent materialization
    - PARENTS: relative to its immediate parents
    - ROOTS: relative to its most upstream assets in the repository
    - KEYS: relative to specific upstream asset keys in the repository

Currently, we're essentially assuming that all FreshnessPolicies have the ROOTS value for this field

- The asset is in violation of policy if it hasn't been updated in the last 20 minutes (i.e. ancestors are not considered)

    FreshnessPolicy.minimum_freshness(minutes=20, relative_to=SELF)

- The asset is in violation of policy if it hasn't been updated within 20 minutes of the most recent materialization of any of its parents.

    FreshnessPolicy.minimum_freshness(minutes=20, relative_to=PARENTS)

sryza:
Awesome. Thoughts about the datetime comment?

OwenKephart:
> Awesome. Thoughts about the datetime comment?

whoops sure I'll do that

sryza (approved):
Woohoo!

We should mark these as experimental before exposing them as top-level exports

PR #10028, opened 2022-10-14
[dagit] Subscribe to in progress asset runs, refresh on relevant events

bengotow:
### Summary & Motivation

Video: https://www.loom.com/share/7278ffd529e04c5bb734de71a068a816
This PR is my second attempt to make the asset graph, asset details page, and asset catalog feel more "real-time" when runs are launched.

My original approach (https://github.com/dagster-io/dagster/pull/9888) added a new GraphQL subscription for asset events, but it required a bunch of new code and the structure of event storage means it was possible to see asset materializations + observations, but not step failures, which could mean some updates not reaching the UI.

This solution is much simpler and strictly client-side:

1) Dagit polls `assetsLatestInfo` every 30 seconds

2) When a run is launched, it refreshes `assetsLatestInfo` immediately and discovers the new run.

3) Whenever `assetsLatestInfo` data contains `inProgressRunIds`  for the visible assets, it subscribes to the stream of run logs (up to 3 runs at a time). When asset or step start / failure events are seen and match the visible assets, it triggers an `assetsLatestInfo` refresh, up to every two seconds.

There's one major loophole in this approach, which is that if you start a run in one tab and switch open browser tabs to an asset graph view, it takes up to 30s for it to realize the run is happening and subscribe to it. I think this is not worth fixing now, and might be easier once we have more subscriptions capability.

There are two performance considerations here:

- This work causes Dagit to subscribe to run logs on more pages. It does not stream much of the data (just the asset keys + step keys), and does not backfill historical logs, but I think it'll cause a noticeable increase in run log queries. I think this is probably OK because we already fetch run logs to calculate stats, etc and it's super well optimized.

- This work causes `assetsLatestInfo` to be called less often when the instance is idle, and much more often when runs are in flight. We can still control the maximum refresh rate (and the UI ensures the requests never stack), so I think we can adjust these constants to keep things sane.

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/useLiveDataForAssetKeys.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/useLiveDataForAssetKeys.tsx
@@ -1,11 +1,18 @@
-import {gql, useQuery} from '@apollo/client';
+import {gql, NetworkStatus, useQuery, useSubscription} from '@apollo/client';
+import uniq from 'lodash/uniq';
 import React from 'react';
 
+import {useQueryRefreshAtInterval} from '../app/QueryRefresh';
+import {useDidLaunchEvent} from '../runs/RunUtils';
 import {AssetKeyInput} from '../types/globalTypes';
 
 import {ASSET_NODE_LIVE_FRAGMENT} from './AssetNode';
-import {buildLiveData} from './Utils';
+import {buildLiveData, tokenForAssetKey} from './Utils';
 import {AssetGraphLiveQuery, AssetGraphLiveQueryVariables} from './types/AssetGraphLiveQuery';
+import {AssetLiveRunLogsSubscription} from './types/AssetLiveRunLogsSubscription';
+
+const SUBSCRIPTION_IDLE_POLL_RATE = 30 * 1000;
+const SUBSCRIPTION_MAX_POLL_RATE = 2 * 1000;
 
 /** Fetches the last materialization, "upstream changed", and other live state
  * for the assets in the given pipeline or in the given set of asset keys (or both).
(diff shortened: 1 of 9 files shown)


sryza:
> There's one major loophole in this approach, which is that if you start a run in one tab and switch open browser tabs to an asset graph view, it takes up to 30s for it to realize the run is happening and subscribe to it. I think this is not worth fixing now, and might be easier once we have more subscriptions capability.

I think this is totally fine.

js_modules/dagit/packages/core/src/asset-graph/useLiveDataForAssetKeys.tsx
```diff
@@ -27,13 +34,126 @@ export function useLiveDataForAssetKeys(assetKeys: AssetKeyInput[]) {
...
+      if (timerRef.current) {
+        clearTimeout(timerRef.current);
+      }
+    };
+  }, [timerRef, liveResult.refetch]);
+
+  // If the event log storage does not support streaming us asset events, fall back to
+  // a polling approach and trigger a single refresh when a run is launched for immediate feedback
+  const liveDataRefreshState = useQueryRefreshAtInterval(liveResult, SUBSCRIPTION_IDLE_POLL_RATE);
+  useDidLaunchEvent(liveResult.refetch, 2000);
```
hellendag, line 71: Use the max poll rate const?

js_modules/dagit/packages/core/src/asset-graph/useLiveDataForAssetKeys.tsx
```diff
@@ -27,13 +34,126 @@ export function useLiveDataForAssetKeys(assetKeys: AssetKeyInput[]) {
...
+    };
+    if (!timerRef.current) {
+      timerRef.current = setTimeout(fire, SUBSCRIPTION_MAX_POLL_RATE);
+    }
+    return () => {
+      if (timerRef.current) {
+        clearTimeout(timerRef.current);
+      }
+    };
+  }, [timerRef, liveResult.refetch]);
```
hellendag, line 66: Slightly confused by this -- the `callback` prop appears to be `() => void`, but this `useCallback` returns a function.

js_modules/dagit/packages/core/src/asset-graph/useLiveDataForAssetKeys.tsx
```diff
@@ -27,13 +34,126 @@ export function useLiveDataForAssetKeys(assetKeys: AssetKeyInput[]) {
...
+  callback: () => void;
+}> = React.memo(({runId, assetKeyTokens, assetStepKeys, callback}) => {
+  // Useful for testing this component:
+  const counter = React.useRef(0);
+  React.useEffect(() => {
+    console.log(`Subscribed to ${runId}`);
+    return () => console.log(`Unsubscribed from ${runId} after ${counter.current} messages`);
+  }, [runId]);
+
+  useSubscription<AssetLiveRunLogsSubscription>(ASSET_LIVE_RUN_LOGS_SUBSCRIPTION, {
```
hellendag, line 120: Do we end up with any unwanted extra renders due to using `useSubscription`?

js_modules/dagit/packages/core/src/asset-graph/useLiveDataForAssetKeys.tsx
bengotow: Oh good catch, it looks like this is a mashup of a useEffect and a useCallback. Will fix!

js_modules/dagit/packages/core/src/asset-graph/useLiveDataForAssetKeys.tsx
bengotow: I checked and we're all good! The approach here is a bit odd (creating a leaf "component" for each subscription and using React keys to make sure the lifespan of the component matches the lifespan we want on the subscription), but when the useSubscription emits an event, only this tiny leaf node re-renders and it's just a `<span />` , so things are pretty quiet with a lot of events coming in!

hellendag (approved):

PR #10030, opened 2022-10-14
Skip docs changes

jmsanders:
There are two kinds of changes where we want to run our docs steps:

1. Changes to anything in the docs directory
2. ~Changes to python files that use the literalinclude docstring/~

The reason for the former is obvious. ~The latter is because it's used to generate our docs snippets.~

I'm going to add a few commits and revert them to see if it correctly skips/unskips the docs steps.

### DIFF
--- a/.buildkite/dagster-buildkite/dagster_buildkite/utils.py
+++ b/.buildkite/dagster-buildkite/dagster_buildkite/utils.py
@@ -300,10 +300,12 @@ def skip_if_no_docs_changes():
 
     # If anything changes in the docs directory
     if any(Path("docs") in path.parents for path in get_changed_files()):
+        logging.info("Run docs steps because files in the docs directory changed")
         return None
 
     # If anything changes that uses the literalinclude docstring
     if any("literalinclude" in path.read_text() for path in get_changed_files()):
+        logging.info("Run docs steps because files with the literalinclude docstring changed")
         return None
 
     return "No docs changes"
--- a/.buildkite/dagster-buildkite/dagster_buildkite/utils.py
+++ b/.buildkite/dagster-buildkite/dagster_buildkite/utils.py
@@ -204,6 +204,7 @@ def get_changed_files():
     return [Path(path) for path in paths]
 
 
+@functools.lru_cache(maxsize=None)
 def skip_if_no_python_changes():
     if not is_feature_branch():
         return None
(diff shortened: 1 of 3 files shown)


.buildkite/dagster-buildkite/dagster_buildkite/utils.py
```diff
@@ -220,3 +222,21 @@ def skip_if_no_helm_changes():
...
+
+    # If anything changes in the docs directory
+    if any(Path("docs") in path.parents for path in get_changed_files()):
+        logging.info("Run docs steps because files in the docs directory changed")
+        return None
+
+    # If anything changes that uses the literalinclude docstring
+    if any("literalinclude" in path.read_text() for path in get_changed_files()):
+        logging.info("Run docs steps because files with the literalinclude docstring changed")
+        return None
```
alangenfeld, line 240: im not sure this is right - the docs refer to python files using pointers and markers 
https://github.com/dagster-io/dagster/blame/master/examples/docs_snippets/docs_snippets/concepts/assets/materialization_io_managers.py#L38

and i think the docs tests need to re-run if the inlined snippets change 

defer to @yuhan on the best strategy here - i think most snippets are in `docs_examples` directory

jmsanders: I took the guidance from here:

https://github.com/dagster-io/dagster/blob/398ea993c5f443f606e6fb447eda6f631f4bc9b1/.buildkite/dagster-buildkite/dagster_buildkite/steps/docs.py#L15-L21

Maybe that's outdated?

I don't see a the `examples/docs_snippets` though. We ran it on changes to `examples` previously but I didn't realize we have that snippets directory hidden amongst the actual examples.

jmsanders: So the `literalinclude` no longer drives any functionality?

jmsanders: Also if "most" are in `examples/docs_snippets`, where are the rest? Or are all in there?

yuhan: alex's right. most are in examples/docs_snippets, the rest are in examples but outside the docs snippets (those are mostly standalone guides).

right we no longer use `literalinclude`.

alangenfeld (approved):

PR #10032, opened 2022-10-14
pass tags to asset reconciliation sensor

jamiedemaria:
### Summary & Motivation
Allows user to specify tags to have on the run request launched by the asset reconciliation sensor

new test


python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -207,9 +208,10 @@ def _make_sensor(
     name: str,
     wait_for_all_upstream: bool,
     wait_for_in_progress_runs: bool,
-    minimum_interval_seconds: Optional[int] = None,
-    description: Optional[str] = None,
-    default_status: DefaultSensorStatus = DefaultSensorStatus.STOPPED,
+    minimum_interval_seconds: Optional[int],
```
jamiedemaria, line 211: I forget if our best practice is to not have these params with defaults since they are passed through from `build_asset_reconciliation_sensor`. can add them back if they should be there

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -313,6 +317,7 @@ def build_asset_reconciliation_sensor(
     minimum_interval_seconds: Optional[int] = None,
     description: Optional[str] = None,
     default_status: DefaultSensorStatus = DefaultSensorStatus.STOPPED,
+    run_tags: Optional[Dict[str, str]] = None,
```
sryza, line 320: Nitpick: Mapping preferred over Dict

sryza (approved):
nice

PR #10037, opened 2022-10-14
Refactor build python package skipping

jmsanders:
Originally, this logic was implemented inside of a function that is responsible for building most - but not all - of our PackageSpec steps. This was convenient because we already had a list of most - but not all
- packages that we could iterate over.

But things like our helm steps and our integration steps still leverage PackageSpec but were not caught with the previous change. This refactor moves the skipping logic into PackageSpec itself.

We do this by using a glob statement to list all available packages (regardless of whether they have PackageSpecs defined or not). We use the existince of a setup.py file as evidence for a package - if we ever move to a different build process, we'll need to change this logic. One interesting check we could add to this is something to make sure every package actually has a PackageSpec - but that's work for another day.

The same general logic applies. Python package steps run on:

1. Every merge to main
2. If the package's implementation changes
3. If the package's tests change
4. If any of the package's requirements' implementations change

It's possible we don't have our full dependency tree spelled out - particularly for things like our integration tests - so I'll likely have a series of follow up PRs to patch holes.

### DIFF
--- a/.buildkite/dagster-buildkite/dagster_buildkite/package_spec.py
+++ b/.buildkite/dagster-buildkite/dagster_buildkite/package_spec.py
@@ -8,7 +8,7 @@ from setuptools.dist import Distribution
 from .python_version import AvailablePythonVersion
 from .step_builder import BuildkiteQueue
 from .steps.tox import build_tox_step
-from .utils import BuildkiteLeafStep, GroupStep
+from .utils import BuildkiteLeafStep, GroupStep, changed_python_package_names, is_feature_branch
 
 _CORE_PACKAGES = [
     "python_modules/dagster",
@@ -162,10 +162,7 @@ class PackageSpec(
             run_pylint,
         )
 
-    def build_skipped_steps(self, skip_reason: str) -> List[GroupStep]:
-        return self.build_steps(skip_reason=skip_reason)
-
-    def build_steps(self, skip_reason: Optional[str] = None) -> List[GroupStep]:
+    def build_steps(self) -> List[GroupStep]:
         base_name = self.name or os.path.basename(self.directory)
         steps: List[BuildkiteLeafStep] = []
 
@@ -237,7 +234,7 @@ class PackageSpec(
                             timeout_in_minutes=self.timeout_in_minutes,
                             queue=self.queue,
                             retries=self.retries,
-                            skip_reason=skip_reason,
+                            skip_reason=self.skip_reason,
                         )
                     )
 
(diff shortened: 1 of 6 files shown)


.buildkite/dagster-buildkite/dagster_buildkite/package_spec.py
```diff
@@ -291,3 +288,22 @@ def requirements(self):
         ]
         install = self.distribution.install_requires
         return extras + install
+
+    @property
+    def skip_reason(self):
```
alangenfeld, line 293: nit: this naming feels a little off given what actually happening, `should_skip`, `should_skip_with_reason`, `reason_if_should_skip` 
those names  aren't amazing either, maybe youll come up with something better. feedback applies equally to the arg on build_tox_step

.buildkite/dagster-buildkite/dagster_buildkite/steps/coverage.py
```diff
@@ -31,5 +31,6 @@ def build_coverage_step() -> CommandStep:
                 "CI_PULL_REQUEST",
             ],
         )
+        .with_skip("Skip coverage uploads until we're finished without Buildkite refactor")
```
alangenfeld, line 34: dont skip on main/full builds?

alangenfeld (approved):

.buildkite/dagster-buildkite/dagster_buildkite/package_spec.py
```diff
@@ -291,3 +288,22 @@ def requirements(self):
...
+        if self.name in changed_python_package_names().with_implementation_changes:
+            return None
+
+        if self.name in changed_python_package_names().with_test_changes:
+            return None
+
+        if any(
+            requirement in changed_python_package_names().with_implementation_changes
+            for requirement in self.requirements
+        ):
```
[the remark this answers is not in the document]
jmsanders: Ah yeah that's a good point. I don't know if realistically our tree even goes three deep, but it's definitely the right thing to do.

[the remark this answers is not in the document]
jmsanders: I'm going to handle this in a follow up PR. Right now, theres no global "registry" of PackageSpecs so even though we know the names of all the packages that changed, we don't have a way to get metadata about each. So it'll be a slightly larger refactor to properly walk the tree.

.buildkite/dagster-buildkite/dagster_buildkite/package_spec.py
jmsanders: I'm actually going to capture a lot of these renames in another PR to to save myself from needing to rebase a bunch of in flight work.

PR #10047, opened 2022-10-17
[dagit] Shift-clicking Materialize should show asset launchpad even if assets are partitioned

bengotow:
### Summary & Motivation

This PR makes a few small changes:

- The order of operations in the "Materialize" button has been flipped so that assets with both partitions and optional config get the launchpad rather than the "choose partitions" UI if you are holding the shift key.

- The launchpad `base` is specified so the default config of the partition set is shown and the partition picker is present.

- A few places show the real errors from the GraphQL response rather than saying "pipeline not found."

- A small CSS import ensures that the CodeMirror autocompletion dropdown is visible in the asset launchpad. It was previously under the dialog. 😬 
 
### Questions (cc @sryza):

- I'm not sure whether it's possible for a partitioned asset to have required resource config or asset config. In that case, this new code would result in the launchpad being shown all the time, even when you do not shift click. I think that may be preferable (since the config is required), but it would prevent the user from triggering a partition backfill via the Materialize button. We could also make that case still show the partitions UI.

- If you specify default config on a partitioned asset job, runs triggered from the config+partitions launchpad result in an error. This may be unrelated to this work but was tripping me up initially. Maybe this scenario is not valid at all?

```
from dagster import asset, Field, DailyPartitionsDefinition, define_asset_job, repository


@asset(
    config_schema={"a": Field(int, default_value=5)},
)
def asset1(context):
    context.log.info(context.op_config["a"])


@repository
def repo():
    return [asset1, define_asset_job("all_job", config={"ops": {"asset1": {"config": {"a": 10}}}})]
```

```dagster._check.CheckError: Invariant failed. Description: Can't supply a ConfigMapping for 'config' when 'partitions_def' is supplied.```

### DIFF
--- a/js_modules/dagit/packages/ui/src/components/ConfigEditorWithSchema.tsx
+++ b/js_modules/dagit/packages/ui/src/components/ConfigEditorWithSchema.tsx
@@ -18,8 +18,9 @@ interface Props {
 }
 
 // Force code editor hints to appear above the dialog modal
-const CodeMirrorShimStyle = createGlobalStyle`
-  .CodeMirror-hints {
+export const CodeMirrorInDialogStyle = createGlobalStyle`
+  .CodeMirror-hints,
+  .CodeMirror-hints.dagit {
     z-index: 100;
   }
 `;
@@ -39,7 +40,7 @@ export const ConfigEditorWithSchema: React.FC<Props> = ({
 
   return (
     <>
-      <CodeMirrorShimStyle />
+      <CodeMirrorInDialogStyle />
       <SplitPanelContainer
         ref={editorSplitPanelContainer}
         axis="horizontal"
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: js_modules/dagit/packages/ui/src/components/ConfigEditorWithSchema.tsx [hunk 1]
```diff
@@ -18,8 +18,9 @@ interface Props {
 }
 
 // Force code editor hints to appear above the dialog modal
-const CodeMirrorShimStyle = createGlobalStyle`
-  .CodeMirror-hints {
+export const CodeMirrorInDialogStyle = createGlobalStyle`
+  .CodeMirror-hints,
+  .CodeMirror-hints.dagit {
```
hellendag, line 23: Did this just need a specificity bump? If so, can it be `&.CodeMirror-hints` or something? It might be kind of nice to avoid needing to use the `dagit` class here specifically, since it's a `dagster-io/ui` component and we won't always have that classname on its use cases.

hellendag (approved):

sryza:
@bengotow @hellendag any chance we can get this in today?

PR #10052, opened 2022-10-17
[dagster-io/ui] Calculate available width for MiddleTruncate

hellendag:
### Summary & Motivation

The `MiddleTruncate` component is currently a bit fussy when rendered in a flex item, and also has some issues with the resize observer firing its callback when resizes have occurred due to the truncated text being rendered.

In order to alleviate these issues, I have modified `MiddleTruncate` to do the following:

1. Render a "fill" string into the container, and measure the container's width. Set the computed font style and available width in state.
2. Use the canvas 2D context as before, using the font style and available width to calculate the longest possible middle-truncated string that can fit within the container.
3. When the middle truncated string is available, render it into the container. Set a flag to denote that we have just determined the width and rendered the middle-truncated string, so the resize observer handler should be skipped.
4. After a brief delay, clear the flag. The delay is in place to ensure that we don't end up in an observe->measure->render loop.
5. When an element resize occurs, start at step 1 again.

Added a handful of Storybook examples to test flex item rendering.

Tested RepositoryLink components within Dagit to ensure that middle truncation looks appropriate, and that strings that are short enough to fit in the max-width of `RepositoryLink` are correctly *not* truncated.

Resize window when viewing virtualized jobs table in Overview, verify that job names are correctly middle truncated.

Tested in Chrome, FF, Safari.

### DIFF
--- a/js_modules/dagit/packages/ui/src/components/MiddleTruncate.tsx
+++ b/js_modules/dagit/packages/ui/src/components/MiddleTruncate.tsx
@@ -1,22 +1,46 @@
 import useResizeObserver from '@react-hook/resize-observer';
-import debounce from 'lodash/debounce';
 import * as React from 'react';
 import styled from 'styled-components/macro';
 
 import {calculateMiddleTruncation} from './calculateMiddleTruncation';
 
-interface Props {
-  text: string;
-}
+/**
+ * A component that performs middle truncation on a given string, based on the evaluated width
+ * of a container div.
+ *
+ * The component will render the provided string -- with no height/visibility -- into our target
+ * container to determine the maximum available width. This width value and the computed font
+ * style are then used to determine the longest middle-truncated string that can fit within the width.
+ *
+ * When the DOM element resizes, the measurement and calculation steps will occur again.
+ */
+export const MiddleTruncate: React.FC<{text: string}> = React.memo(({text}) => {
+  // Track the font style and target maximum width. `null` means no measurement has
+  // taken place yet.
+  const [truncatedText, setTruncatedText] = React.useState<string | null>(null);
 
-const getFont = (element: HTMLDivElement) => getComputedStyle(element).font;
+  // An element that renders the full text into the container, for the purpose of
+  // measuring the maximum available/necessary width for our truncated string.
+  const measure = React.useRef<HTMLDivElement>(null);
 
-export const MiddleTruncate: React.FC<Props> = (props) => {
-  const {text} = props;
-  const [targetStyle, setTargetStyle] = React.useState<TargetStyle | null>(null);
-  const ref = React.useRef<HTMLDivElement>(null);
+  // Given the target font style and allotted width, calculate the largest possible middle-
+  // truncated string.
+  const calculateTargetStyle = React.useCallback(() => {
+    if (measure.current) {
+      setTruncatedText(calculateMiddleTruncatedText(measure.current, text));
+    }
+  }, [text]);
 
-  const truncated = useFixedSpan(text, targetStyle);
+  // Use a layout effect to trigger the process of calculating the truncated text, for the
+  // initial render.
+  React.useLayoutEffect(() => {
+    calculateTargetStyle();
+  }, [calculateTargetStyle]);
+
+  // If the container has just been resized, recalculate.
+  useResizeObserver(measure.current, () => {
+    calculateTargetStyle();
+  });
 
   // Copy the full text, not just the truncated version shown in the DOM.
   const handleCopy = React.useCallback(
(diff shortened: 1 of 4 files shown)


hellendag:
* master
  * **PR #10052** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10052" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10052?utm_source=stack-comment).

js_modules/dagit/packages/ui/src/components/MiddleTruncate.tsx
```diff
@@ -9,14 +8,87 @@ interface Props {
   text: string;
 }
 
+const FILL = 'M'.repeat(200);
+
 const getFont = (element: HTMLDivElement) => getComputedStyle(element).font;
 
+/**
```
bengotow, line 15: +1 for all the new comments!

js_modules/dagit/packages/ui/src/components/MiddleTruncate.tsx
```diff
@@ -9,14 +8,87 @@ interface Props {
...
+  // Given the target font style and allotted width, calculate the largest possible middle-
+  // truncated string. This hook returns null if there is no style available yet.
+  const truncated = useMiddleTruncatedText(text, targetStyle);
+
+  // When we are rendering the "fill" string to determine the available width, use a layout
+  // effect to perform the measurement. Get the computed font style as well -- both of these
+  // will be used to draw the text string onto a canvas context to find the appropriate
+  // truncated string.
+  React.useLayoutEffect(() => {
+    if (container.current && determineAvailableWidth) {
```
bengotow, line 52: I think if you want, you could actually do this without the temporary state + re-render by directly assigning long content here, since the clientRect functions are synchronous:

```
      container.current.innerText = 'M'.repeat(200);
      setTargetStyle({
        width: container.current.getBoundingClientRect().width,
        font: getFont(container.current),
      });
      container.current.innerText = ''; // clear it, just to be sure nobody ever sees MMMMMM
      // innerText is restored by the re-render after setState
```

js_modules/dagit/packages/ui/src/components/MiddleTruncate.tsx
```diff
@@ -9,14 +8,87 @@ interface Props {
...
+
+  // If the container has just been resized -- but NOT due to a width difference between rendering "fill"
+  // text and our actual text -- then re-render the "fill" in order to perform a new measurement and
+  // recalculate the middle truncation.
+  const observerFn = React.useCallback(() => {
+    if (!justDetermined.current) {
+      setDetermineAvailableWidth(true);
+    }
+  }, []);
+
```
bengotow, line 90: This looks good to me! If the flag + setTimeout isn't sufficient and we keep seeing problems with ResizeObserver loops, I think we could also fix this by detatching + re-attaching the observer on each render (so  the render itself can't trigger it?) 

edit: It's actually kinda strange this is happening - I wonder if removing the React.useCallback would actually fix this, because it'd force useResizeObserver to detach + reattach on each render... the useResizeObserver hook might also need to mount in a useLayoutEffect instead of a useEffect :-/

bengotow (approved):
This looks great! I put a couple inline comments in but nothing major or blocking 🙌

salazarm (approved):
is the ResizeObserver being fussy because adding the text ends up resizing the flex item? If so would it be possible that by ignoring the immediate resize event after that we might actually be preventing the flex item from growing as much as it can?

hellendag:
> is the ResizeObserver being fussy because adding the text ends up resizing the flex item? If so would it be possible that by ignoring the immediate resize event after that we might actually be preventing the flex item from growing as much as it can?

Correct, the resize of the flex item is the problem, but it's kind of the opposite problem -- the flex item is shrinking to fit the text, which leads to resizing.

The element is measured with the initial width, and then we perform a truncation calculation. The truncated text is essentially guaranteed to have a smaller width than the default width, which means that when it renders, the flex item shrinks. If this triggers a resize observation, we can end up looping `resize -> truncate -> render -> resize...` in a _smaller_ direction until whatever point it happens to find a width where the text fits the container without being truncated again.

By ignoring the immediate resize event, I am aiming to stop it from looping to ever smaller widths. As soon as we have truncated text that fits the allotted width, it should stop trying to truncate further.

js_modules/dagit/packages/ui/src/components/MiddleTruncate.tsx
hellendag: Ah, that's a nice idea. I'll try it.

hellendag:
I think I can simplify this even further.

hellendag:
Thanks for your thoughts on this! I've got a simpler approach here (props @salazarm for the hidden element idea) that seems to work pretty well.

js_modules/dagit/packages/ui/src/components/MiddleTruncate.tsx
```diff
@@ -28,88 +52,65 @@ export const MiddleTruncate: React.FC<Props> = (props) => {
...
-      (value: string) => ctx.measureText(value).width,
-    );
+/**
+ * Compute the font style and maximum/necessary width for the measured container,
+ * for the specified string of text.
+ *
+ * Given those values, use a 2D canvas context to determine the longest possible
+ * middle-truncated string.
+ */
+const calculateMiddleTruncatedText = (container: HTMLDivElement, text: string) => {
```
hellendag, line 83: No longer a hook, since all state management now lives in the component instead.

js_modules/dagit/packages/ui/src/components/MiddleTruncate.tsx
```diff
@@ -1,22 +1,46 @@
...
+  // initial render.
+  React.useLayoutEffect(() => {
+    calculateTargetStyle();
+  }, [calculateTargetStyle]);
 
-  const truncated = useFixedSpan(text, targetStyle);
+  // If the container has just been resized, recalculate.
+  useResizeObserver(measure.current, () => {
+    calculateTargetStyle();
+  });
```
hellendag, line 43: Verified with logging in Storybook that this observer fires once per resize step, and not excessively in response to the truncation.

js_modules/dagit/packages/ui/src/components/MiddleTruncate.tsx
```diff
@@ -1,22 +1,46 @@
...
+ * A component that performs middle truncation on a given string, based on the evaluated width
+ * of a container div.
+ *
+ * The component will render the provided string -- with no height/visibility -- into our target
+ * container to determine the maximum available width. This width value and the computed font
+ * style are then used to determine the longest middle-truncated string that can fit within the width.
+ *
+ * When the DOM element resizes, the measurement and calculation steps will occur again.
+ */
+export const MiddleTruncate: React.FC<{text: string}> = React.memo(({text}) => {
```
hellendag, line 17: It occurred to me that this could be memoized.

js_modules/dagit/packages/ui/src/components/MiddleTruncate.tsx
```diff
@@ -1,22 +1,46 @@
...
+  // Track the font style and target maximum width. `null` means no measurement has
+  // taken place yet.
+  const [truncatedText, setTruncatedText] = React.useState<string | null>(null);
```
hellendag, line 20: Now only tracking truncated text in state, not the style/width values -- these are now only tracked within the calculation function, which is the only place we need them.

js_modules/dagit/packages/ui/src/components/MiddleTruncate.tsx
```diff
@@ -28,88 +52,65 @@ export const MiddleTruncate: React.FC<Props> = (props) => {
...
+    </Container>
   );
-};
+});
 
-const MiddleTruncateDiv = styled.div`
+// An invisible target element that contains the full, no-wrapped text. This is used
+// to measure the maximum available width for our truncated string.
+const MeasureWidth = styled.div`
+  height: 0;
```
hellendag, line 66: Zero height seems to work well enough for getting the necessary width, without a `visibility: hidden`.

js_modules/dagit/packages/ui/src/components/MiddleTruncate.tsx
```diff
@@ -1,22 +1,46 @@
...
+  // measuring the maximum available/necessary width for our truncated string.
+  const measure = React.useRef<HTMLDivElement>(null);
+
+  // Given the target font style and allotted width, calculate the largest possible middle-
+  // truncated string.
+  const calculateTargetStyle = React.useCallback(() => {
+    if (measure.current) {
+      setTruncatedText(calculateMiddleTruncatedText(measure.current, text));
+    }
+  }, [text]);
```
bengotow, line 32: Only playing devils advocate, but now that the component is memoized and it's only prop is `text`, I think any useEffect with just `text` in the deps can go!

It might actually be fine to `calculateMiddleTruncatedText` on every render since the entire component is memozied. (Not sure that this would be clearer though, I like being extra safe for future-proofing!)

js_modules/dagit/packages/ui/src/components/MiddleTruncate.tsx
bengotow: This is great! I'm a fan of this not being a hook because we could potentially use it in some DAG viz cases too where we probably want to avoid an extra render cycle for every label 🙌

bengotow (approved):
This looks great! The new approach seems so much simpler, really cool that the height=0 trick was enough to get the ideal bounds size.

salazarm (approved):
Sweet and +1 to logic outside of hooks!

PR #10056, opened 2022-10-17
[Multi-dimensional partitions 2/n] Read/write from asset event tags table

clairelin135:
Part 1: https://github.com/dagster-io/dagster/pull/10001

This PR introduces methods that read and write to the asset event tags table:
- Modifies the `AssetMaterialization` class to accept system-provided tags, writing these tags to the asset event tags table
- Adds a `tags` param to `EventRecordsFilter`, enabling filtering by tags
- Adds a `get_asset_event_tags` method, which filters for wiped materializations to ensure only valid event tags are returned

I've created a `MultiDimensionalPartitionKey` class that subclasses `str` to avoid modifying type signatures of methods that fetch the partition key.


python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
```diff
@@ -0,0 +1,76 @@
...
+        return super(PartitionDimensionKey, cls).__new__(
+            cls,
+            dimension_name=check.str_param(dimension_name, "dimension_name"),
+            partition_key=check.str_param(partition_key, "partition_key"),
+        )
+
+
+@whitelist_for_serdes
+class MultiDimensionalPartitionKey(
+    NamedTuple("_MultiDimensionalPartitionKey", [("dimension_keys", List[PartitionDimensionKey])])
```
sryza, line 32: Are there particular reasons not to use a `Mapping[str, str]`?

clairelin135: Currently we serialize this value (and mappings cannot be serialized).

But this also enforces that the dimension keys will be ordered by dimension name, to allow for equivalence checks and to ensure a consistent ordering of the stringified partition key (e.g. always displaying "US|2022-10-19" instead of "2022-10-19|US")

sryza: Got it - makes sense

python_modules/dagster/dagster/_core/definitions/events.py
```diff
@@ -443,17 +453,32 @@ def __new__(
             check.tuple_param(asset_key, "asset_key", of_type=str)
             asset_key = AssetKey(asset_key)
 
+        check.opt_mapping_param(tags, "tags", key_type=str, value_type=str)
+        if any([not tag.startswith(SYSTEM_TAG_PREFIX) for tag in tags or {}]):
+            check.failed(
+                "Users should not pass values into the tags argument for AssetMaterializations. The tags argument is reserved for system-populated tags."
```
sryza, line 459: Nice

sryza: Exceeds line length though

python_modules/dagster/dagster/_core/storage/event_log/base.py
```diff
@@ -292,6 +303,14 @@ def get_latest_materialization_events(
     ) -> Mapping[AssetKey, Optional[EventLogEntry]]:
         pass
 
+    @abstractmethod
+    def get_asset_event_tags(self) -> List[Tuple[str, Set[str]]]:
```
sryza, line 307: Do we only need this for tests? A method that returns all tags seems dangerously unperformant if a user gets a hold of it. Thoughts on scoping it to a particular asset key? Also, could use `Sequence` instead of `List` and `AbstractSet` instead of `Set`.

python_modules/dagster/dagster/_core/definitions/events.py
```diff
@@ -443,17 +453,32 @@ def __new__(
...
+            )
+
         metadata = check.opt_mapping_param(metadata, "metadata", key_type=str)
         metadata_entries = check.opt_sequence_param(
             metadata_entries, "metadata_entries", of_type=(MetadataEntry, PartitionMetadataEntry)
         )
 
+        serializable_partition: Optional[Union[str, MultiDimensionalPartitionKey]] = None
```
sryza, line 467: The name `serializable_partition` tripped me up a tiny bit here, because it implies that the `partition` argument supplied by the user isn't serializable. But both are serializable, right?

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+@whitelist_for_serdes
+class PartitionDimensionKey(
+    NamedTuple(
+        "_PartitionDimensionKey",
+        [
+            ("dimension_name", str),
+            ("partition_key", str),
+        ],
```
sryza, line 15: If you get rid of trailing commas, black might be able to format this more concisely.

sryza (commented):
This looks close to me.  The main question I have is whether, if we're going with `MultiPartitionsDefinition`, we should also go with `MultiPartitionsKey`.

python_modules/dagster/dagster/_core/storage/event_log/base.py
clairelin135: That sounds good--ive changed it now

clairelin135:
@sryza This PR now has been updated to make `MultiDimensionalPartitionKey` a subclass of string. The main difference here compared to the previous implementation is that the stringified representation is now serialized (instead of an object).

This means that we'll need to choose a consistent format for storage (currently separating by the pipe "|" character, ordering by dimension name). We determine from reading the asset materialization tags whether the partition is a multi-dimensional partition, so we don't necessarily need to disallow the delimiter character.

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -590,6 +647,28 @@ def enable_secondary_index(self, name):
                     .values(migration_completed=datetime.now())
                 )
 
+    def _apply_tags_table_joins(
```
sryza, line 650: There's only a single join going on here, right (with multiple join conditions)? If so, `_apply_tags_table_joins` would be more accurate.

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -202,6 +218,37 @@ def _get_asset_entry_values(self, event, has_asset_key_index_cols):
 
         return entry_values
 
+    def store_asset_event_tags(self, event: EventLogEntry, event_id: int):
```
sryza, line 221: A `None` return type would aid clarity

python_modules/dagster/dagster/_core/storage/event_log/in_memory.py
```diff
@@ -352,6 +373,29 @@ def get_asset_run_ids(self, asset_key):
...
+                    and record.dagster_event.asset_key == asset_key
+                    and record.dagster_event.event_type_value
+                    == DagsterEventType.ASSET_MATERIALIZATION.value
+                    and self._wiped_asset_keys.get(record.dagster_event.asset_key, 0)
+                    < record.timestamp
+                    and record.dagster_event.step_materialization_data.materialization.tags
+                ):
+                    for (
+                        key,
+                        value,
```
sryza, line 391: I think this could be laid out on fewer lines

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
```diff
@@ -0,0 +1,52 @@
...
+
+        str_key = super(MultiPartitionKey, cls).__new__(
+            cls, "|".join([dim_key.partition_key for dim_key in dimension_keys])
+        )
+
+        str_key.dimension_keys = dimension_keys
+
+        return str_key
+
+    def keys_by_dimension(self):
```
sryza, line 51: A return type would be helpful here

python_modules/dagster/dagster/_core/definitions/events.py
```diff
@@ -430,8 +435,13 @@ def __new__(
         description: Optional[str] = None,
         metadata_entries: Optional[Sequence[Union[MetadataEntry, PartitionMetadataEntry]]] = None,
         partition: Optional[str] = None,
+        tags: Optional[Mapping[str, str]] = None,
         metadata: Optional[Mapping[str, RawMetadataValue]] = None,
     ):
+        from dagster._core.definitions.multi_dimensional_partitions import (
```
sryza, line 441: Does this need to be on multiple lines?

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __new__(cls, dimension_name: str, partition_key: str):
+        return super(PartitionDimensionKey, cls).__new__(
+            cls,
+            dimension_name=check.str_param(dimension_name, "dimension_name"),
+            partition_key=check.str_param(partition_key, "partition_key"),
+        )
+
+
+class MultiPartitionKey(str):
```
sryza, line 21: Do we end up serializing this? How does that end up getting handled?

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
sryza: Also, this should be `@property`

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
```diff
@@ -0,0 +1,52 @@
...
+    Subclasses the string class to keep partition key type as a string.
+
+    Contains additional methods to access the partition key for each dimension.
+    Creates a string representation of the partition key for each dimension, separated by a pipe (|).
+    Orders the dimensions by name, to ensure consistent string representation.
+    """
+
+    dimension_keys: List[PartitionDimensionKey] = []
+
+    def __new__(cls, partition_dimension_mapping: Mapping[str, str]):
```
sryza, line 33: A name like `keys_by_dimension` would be more clear and more consistent with the property

sryza (commented):
This looks close. Curious on your thoughts about the suggestion from the user about using JSON serialization instead of the pipe character

clairelin135:
@sryza I commented back on the other PR, but on second thought I think we'll need to make the "|" character invalid in static partitions definitions. In Dagit, we'll have a syntax for selecting multiple multi-dimensional partitions by doing something like `"[US...Brazil]|2022-10-24"`, and having a static partition key containing the pipe character would cause breakages in evaluating this.

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
clairelin135: Hmm, did you mean `_apply_tags_table_join`?

sryza: Oops yes

python_modules/dagster/dagster/_core/storage/event_log/in_memory.py
clairelin135: black formatted this 🤷‍♀️

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
clairelin135: yes, this ends up getting serialized. So we directly serialize the stringified partition key e.g. `US|2022-10-10` into storage. This won't cause breakages for users though--we read from the tags to learn whether a string is a multi-partition key or not

python_modules/dagster/dagster/_core/storage/event_log/in_memory.py
sryza: if you take out the comma after value, it might compress them

clairelin135: still does not compress, not sure why

sryza: weird. ok no worries

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
sryza: Got it - makes sense

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1126,6 +1213,22 @@ def _add_assets_wipe_filter_to_query(
 
         return query
 
+    def get_asset_event_tags(self, asset_key: AssetKey) -> Sequence[Tuple[str, AbstractSet[str]]]:
+        tags = defaultdict(set)
+        query = (
+            db.select([AssetEventTagsTable.c.key, AssetEventTagsTable.c.value])
+            .distinct(AssetEventTagsTable.c.key, AssetEventTagsTable.c.value)
+            .where(AssetEventTagsTable.c.asset_key == asset_key.to_string())
+        )
```
prha, line 1222: Is this based off of the run tags table query logic?  This was changed in https://github.com/dagster-io/dagster/pull/10036 because we were getting bad performance with it.

python_modules/dagster/dagster/_core/storage/event_log/in_memory.py
```diff
@@ -236,6 +236,27 @@ def _apply_filters(record):
                 and record.timestamp >= event_records_filter.before_timestamp
             ):
                 return False
+
+            if (
```
prha, line 240: heads up, you might get merge conflicts with https://github.com/dagster-io/dagster/pull/10154, but it should be easy to resolve.

prha (changes requested):

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
clairelin135: Yep, it was. I updated this PR to follow the new changes in #10036!

prha: 🎉

python_modules/dagster/dagster/_core/definitions/events.py
```diff
@@ -459,17 +467,38 @@ def __new__(
             check.tuple_param(asset_key, "asset_key", of_type=str)
             asset_key = AssetKey(asset_key)
 
+        check.opt_mapping_param(tags, "tags", key_type=str, value_type=str)
+        if any([not tag.startswith(SYSTEM_TAG_PREFIX) for tag in tags or {}]):
+            check.failed(
+                "Users should not pass values into the tags argument for AssetMaterializations. "
```
prha, line 473: Should we even expose this as a top-level argument then?

Would we be better off exposing some sort of sequence of partition dimension keys then, that internally get stored as tags?

clairelin135: I thought about doing that too, but upon deserializing a materialization I found that there needed to be a tags argument since we're storing tags within a materialization.

We could decide not to serialize the tags within an asset materialization altogether, in which case we'd need to serialize the partition key as some sort of sequence of partition dimension keys, so that we can determine the partition dimensions without reading the tags. I felt that that made the APIs more awkward though

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1451,6 +1452,10 @@ def get_asset_records(
     ) -> Iterable["AssetRecord"]:
         return self._event_storage.get_asset_records(asset_keys)
 
+    @traced
+    def get_asset_event_tags(self, asset_key: AssetKey) -> Sequence[Tuple[str, AbstractSet[str]]]:
```
prha, line 1456: same comment as the cloud version: consider renaming to `get_all_event_tags_for_asset`?

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -657,6 +714,35 @@ def _apply_filter_to_query(
         if event_records_filter.storage_ids:
             query = query.where(SqlEventLogStorageTable.c.id.in_(event_records_filter.storage_ids))
 
+        if event_records_filter.tags:
+            # Also filter by asset key in asset key table to take advantage of asset key tags index
+            intersections = [
```
prha, line 719: same comment as in cloud: I think we should assert that asset_key is set...

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1126,6 +1213,22 @@ def _add_assets_wipe_filter_to_query(
...
+            .where(AssetEventTagsTable.c.asset_key == asset_key.to_string())
+        )
+
+        with self.index_connection() as conn:
+            results = conn.execute(query).fetchall()
+
+        for key, value in results:
+            tags[key].add(value)
+
+        return sorted(list([(k, v) for k, v in tags.items()]), key=lambda x: x[0])
```
prha, line 1230: Is this the right signature?  Will this power the status queries where one dimension is held fixed?

prha (changes requested):

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1453,8 +1453,8 @@ def get_asset_records(
         return self._event_storage.get_asset_records(asset_keys)
 
     @traced
-    def get_asset_event_tags(self, asset_key: AssetKey) -> Sequence[Tuple[str, AbstractSet[str]]]:
-        return self._event_storage.get_asset_event_tags(asset_key)
+    def get_all_event_tags_for_asset(self, asset_key: AssetKey) -> Sequence[Tuple[str, AbstractSet[str]]]:
```
prha, line 1456: might be okay for now, but we might want to add an optional filter arg here, so that we can fix one of the partition dimension values

clairelin135: Sounds good! This method is now updated to contain this filter arg and also filters out wiped materializations

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1451,6 +1452,12 @@ def get_asset_records(
     ) -> Iterable["AssetRecord"]:
         return self._event_storage.get_asset_records(asset_keys)
 
+    @traced
+    def get_all_event_tags_for_asset(
+        self, asset_key: AssetKey, key: Optional[str] = None
```
prha, line 1457: maybe rename to be `tag_key`?

and sorry for the back and forth, but maybe if we're applying a filter, it's confusing to have `get_all_` instead of just `get_`

python_modules/dagster/dagster_tests/core_tests/storage_tests/utils/event_log_storage.py
```diff
@@ -2330,3 +2331,271 @@ def test_get_maximum_record_id(self, storage):
...
+            )
+            assert len(materializations) == 1
+
+            asset_event_tags = storage.get_all_event_tags_for_asset(key)
+            assert asset_event_tags == [
+                ("dagster/partition/country", {"US"}),
+                ("dagster/partition/date", {"2022-10-13"}),
+            ]
+
+    def test_materialization_tag_on_wipe(self, storage, instance):
```
prha, line 2368: nice!

prha (approved):
Some naming nits, but I think this looks good.

clairelin135:
@prha I added the naming changes you suggested!

I did update the signature of the `get_event_tags_for_asset` method to group the tags by event, so we know which dimensions were materialized per event.

I also realized that instead of searching by dimension name (e.g. "country"), we actually want to search by dimension name and value ("country" == "US"). In this case, I've updated the query to search for all of the events containing the dimension name and value, then return the tags for all of those events. The biggest change here is that now the query contains a join, but I did confirm that we are using both indices (indexing by event id and by asset key/tag key/tag value).

python_modules/dagster/dagster/_core/storage/event_log/base.py
```diff
@@ -292,6 +304,20 @@ def get_latest_materialization_events(
     ) -> Mapping[AssetKey, Optional[EventLogEntry]]:
         pass
 
+    @abstractmethod
+    def get_event_tags_for_asset(
+        self, asset_key: AssetKey, tag_key: Optional[str] = None, tag_value: Optional[str] = None
```
prha, line 309: Instead of separate tag_key/tag_value arguments, maybe this should be `filter_tags` that takes a mapping?  If we ever want to support more than 2 dimensions.

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1278,30 +1264,27 @@ def get_event_tags_for_asset(
...
-        if not perform_tag_join:
+        if not filter_tags:
             with self.index_connection() as conn:
                 results = conn.execute(tags_query).fetchall()
         else:
-            tags_query = tags_query.where(
-                db.and_(
-                    AssetEventTagsTable.c.key == tag_key, AssetEventTagsTable.c.value == tag_value
+            intersections = [
+                tags_query.where(
```
prha, line 1272: i think this might cause some future confusion, in that sometimes, `tags_query` is the top-level query, and sometimes it's the subquery.

I'd prefer it if we just had the top-level query, and we applied the intersection and had the subquery not reuse any existing var.

So, something like this:
```
tags_query = db.select(key, value, event_id)
if filter_tags:
    intersections = [
        db.select(event_id).where(db.and_(asset_key == asset_key.to_string(), key == key, value = value))
        for key, value in filter_tags.items()
    ]
    tags_query = tags_query.where(event_id in db.intersect(*intersections))
```

PR #10059, opened 2022-10-17
Fix runs yielded error for multi asset sensor

clairelin135:
Currently, when any skip reasons are returned from the multi asset sensor, a `Asset materializations have been handled in this sensor, but the cursor was not updated` error was raised. This PR fixes the error. This occurs because all decorated multi asset sensor functions are wrapped with generator functions.

[Reported by a user](https://dagster.slack.com/archives/C01U954MEER/p1666036752225909)


adam-bloom (commented):
Tested this locally, confirmed that it fixes my issue, thanks @clairelin135! We also confirmed that it is the `multi_asset_sensor` decorator that is forcing all multi-asset sensors to be generators, so this fix benefits all users implementing sensors with that decorator.

PR #10061, opened 2022-10-18
[freshness-policy] [2/n] FreshnessPolicySensorDefinition

OwenKephart:
### Summary & Motivation

Depends on: https://github.com/dagster-io/dagster/pull/10024

This creates an FreshnessPolicySensorDefinition object (and decorator) which can be used to create a sensor which monitors the freshness status of assets in a repository.

The PR includes a utility to calculate the "upstream materialization times" of any asset in the repository, which is a mapping from each root asset key to the timestamp of the materialization of that asset which was consumed to produce the most recent materialization of that asset. To do this calculation, we just take the root data times of each parent asset materialization (the most recent asset materialization of the parent which happened before the current one) and combine them together (taking the maximum for each parent). We first check if the parent asset events already tagged with this information, but if not the process recurses up the tree. This means that the first call can be fairly expensive (as it may need to traverse the entire tree), but subsequent calls should be cheap, because that information is cached for later use.

This PR includes a hacked-in "asset tagging" system (using the key-value store on the instance), but this should be replaced with the real asset tags system before merging. (cc @clairelin135 )

I did some profiling, and on a large repository (1000 assets, 100 SLAs), calculating the SLA status of the entire repository using postgres took around 30 seconds on the first run (with the longest individual call taking ~3 seconds, but subsequent loads of the entire repository took < 5 seconds (as all the data was already cached), with each individual call taking a fraction of a second. This bit could almost certainly be sped up with a batch loader type of thing (like what we do for the latest materialization events). 

This makes me pretty confident that this process is fast enough to enable a "real-time" feel in the UI.

FUTURE IMPROVEMENTS:
- use real asset tags table (cc @clairelin135)
- instead of querying for "most recent asset materialization of parent before this", use the "upstream asset record ids" that will exist on the AssetMaterialization event (cc @smackesey ). currently, race conditions can cause this calculation to be inaccurate.
- allow for freshness policies which are not relative to the root assets (adds complexity, so left out for v0)


python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -1070,6 +1070,125 @@ def get_asset_cursor_with_advances(
         )
 
 
+@experimental
```
sryza, line 1073: This file is getting huge - I think better to put this in its own file. We probably should have done that for asset sensors and run status sensors too.

python_modules/dagster/dagster/_core/definitions/freshness_policy_sensor_definition.py
```diff
@@ -0,0 +1,376 @@
...
+
+    """
+
+    def __init__(
+        self,
+        instance_ref: Optional[InstanceRef],
+        cursor: Optional[str],
+        repository_def: "RepositoryDefinition",
+        asset_selection: Optional[AssetSelection],
+        asset_keys: Optional[Sequence[AssetKey]],
```
clairelin135, line 74: If i understand correctly, seems like if an asset key does not belong to an asset in the same repository as this sensor, these context methods won't really return anything useful (since there is no assets definition).

Thoughts on just throwing an error if the there is no assets definition associated with a provided asset key in the repo?

python_modules/dagster/dagster/_core/execution/calculate_data_time.py
```diff
@@ -0,0 +1,87 @@
...
+            upstream_key_strs = upstream_asset_key_mapping[key_str]
+            if not upstream_key_strs:
+                return {key_str}
+            return set().union(
+                *(_recursive(upstream_key_str) for upstream_key_str in upstream_key_strs)
+            )
+
+        return _recursive(asset_key.to_user_string())
+
+    def _get_root_data(record):
```
clairelin135, line 28: This part was trickier for me to understand, especially because it also writes into storage as well. Would it be possible to add a comment describing what this function does?

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1283,6 +1283,22 @@ def handle_run_event(self, run_id: str, event: "DagsterEvent"):
     def add_run_tags(self, run_id: str, new_tags: Dict[str, str]):
         return self._run_storage.add_run_tags(run_id, new_tags)
 
+    @traced
+    def add_asset_event_tags(self, asset_event_id: int, new_tags: Dict[str, str]):
```
clairelin135, line 1287: My current PRs only support adding tags when writing the initial event into storage. Just as a heads up, you'll need to make changes to allow updating tags for existing events after I merge

python_modules/dagster/dagster/_core/definitions/freshness_policy_sensor_definition.py
OwenKephart: good point, done

python_modules/dagster/dagster/_core/definitions/freshness_policy_sensor_definition.py
```diff
@@ -0,0 +1,381 @@
...
+            if minutes_late != 0
+        }
+
+
+@experimental
+class FreshnessPolicySensorDefinition(SensorDefinition):
+    """Define an asset sensor that takes some action based on the SLA status of selected assets.
+
+    Users should not instantiate this object directly. To construct a
+    `FreshnessPolicySensor`, use :py:func:`dagster.
```
sryza, line 189: Nitpick: `FreshnessPolicySensor` should either be "freshness policy sensor" or `FreshnessPolicySensorDefinition`.

python_modules/dagster/dagster/_core/definitions/freshness_policy_sensor_definition.py
```diff
@@ -0,0 +1,381 @@
...
+            status can be overridden from Dagit or via the GraphQL API.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        asset_keys: Optional[Sequence[AssetKey]],
+        asset_selection: Optional[AssetSelection],
+        job_name: Optional[str],
+        asset_materialization_fn: Callable[
```
sryza, line 218: This should have a different name than `asset_materialization_fn`, right?

python_modules/dagster/dagster/_grpc/impl.py
```diff
@@ -300,6 +304,16 @@ def get_external_sensor_execution(
                     asset_keys=sensor_def.asset_keys,
                 )
             )
+        elif isinstance(sensor_def, FreshnessPolicySensorDefinition):
```
sryza, line 307: I know we didn't start it here, but this is kind of weird that the grpc impl needs to know about the different kinds of sensors.  I would think that we could just pass the repository def to the SensorEvaluationContext in all cases and then have the user-code wrapper within the `FreshnessPolicySensorDefinition` construct the `FreshnessPolicySensorEvaluationContext`.

python_modules/dagster/dagster/_core/execution/calculate_data_time.py
```diff
@@ -0,0 +1,125 @@
...
+                            root_data[key] = (None, None)
+                        else:
+                            root_data[key] = max(root_data.get(key, tup), tup)
+
+        # now that we've done this calculation, store it in the database
+        instance.add_asset_event_tags(record.storage_id, {ROOT_DATA_TAG: root_data})
+        return root_data
+
+    # get most recent asset materialization event record for this asset
+    records = instance.get_event_records(
```
sryza, line 107: I think you can might be able to use `instance.get_latest_materialization_event`?

python_modules/dagster/dagster/_core/execution/calculate_data_time.py
```diff
@@ -0,0 +1,125 @@
...
+            if not upstream_key_strs:
+                return {key_str}
+            return set().union(
+                *(_recursive(upstream_key_str) for upstream_key_str in upstream_key_strs)
+            )
+
+        return _recursive(asset_key.to_user_string())
+
+    def _get_upstream_materialization_times(record):
+
```
sryza, line 53: nitpick: remove blank line at top of method. per: https://elementl-workspace.slack.com/archives/C03AFEEDDNX/p1666630503242289?thread_ts=1666629931.217449&cid=C03AFEEDDNX.

also, type annotations would be helpful here

python_modules/dagster/dagster/_core/execution/calculate_data_time.py
```diff
@@ -0,0 +1,125 @@
...
+        # check if we've already calculated data times for this record
+        cur_tags = instance.get_asset_event_tags(record.storage_id)
+        if ROOT_DATA_TAG in cur_tags:
+            return cur_tags[ROOT_DATA_TAG]
+
+        # find parent asset keys
+        cur_key = record.event_log_entry.dagster_event.event_specific_data.materialization.asset_key
```
sryza, line 60: now you can just use `record.asset_key`

python_modules/dagster/dagster/_core/execution/calculate_data_time.py
```diff
@@ -0,0 +1,125 @@
...
+    def _get_root_keys(asset_key):
+        def _recursive(key_str):
+            upstream_key_strs = upstream_asset_key_mapping[key_str]
+            if not upstream_key_strs:
+                return {key_str}
+            return set().union(
+                *(_recursive(upstream_key_str) for upstream_key_str in upstream_key_strs)
+            )
+
+        return _recursive(asset_key.to_user_string())
```
sryza, line 50: Why convert to user string?

python_modules/dagster/dagster/_core/execution/calculate_data_time.py
```diff
@@ -0,0 +1,125 @@
...
+from ..instance import DagsterInstance
+
+ROOT_DATA_TAG = ".dagster/root_data_ids"
+
+
+def get_upstream_materialization_times_for_key(
+    instance: DagsterInstance,
+    asset_key: AssetKey,
+    upstream_asset_key_mapping: Mapping[str, AbstractSet[str]],
+) -> Mapping[str, Tuple[Optional[int], Optional[datetime]]]:
```
sryza, line 14: Skimming this function definition and docstring, it's not immediately clear to me what entities the `str`, `Optional[int]`, and `Optional[datetime]` in the return value correspond to.

python_modules/dagster/dagster/_core/execution/calculate_data_time.py
```diff
@@ -0,0 +1,125 @@
+from datetime import datetime, timezone
+from typing import AbstractSet, Mapping, Optional, Tuple
+
+from ..definitions.events import AssetKey
+from ..instance import DagsterInstance
+
+ROOT_DATA_TAG = ".dagster/root_data_ids"
```
sryza, line 7: "root" and "data" aren't terms that are still used in the remainder of the freshness policy literature, right? Is there a name that would be more in-keeping?

Also, haven't thought this through at all, but did you consider having a tag per root asset? It could make it easier to find the downstream materializations associated with a materialization, if that's something we want to do in the future.

python_modules/dagster/dagster/_core/definitions/freshness_policy.py
```diff
@@ -0,0 +1,137 @@
...
+    This gives a lower bound on the most recent records that could possibly be incorporated into the
+    current state of the asset to which this policy is attached.
+    """
+
+    @abstractmethod
+    def minutes_late(
+        self,
+        evaluation_time: datetime,
+        upstream_materialization_times: Mapping[AssetKey, Optional[datetime]],
+    ) -> Optional[float]:
```
sryza, line 28: In what case would `None` get returned?

python_modules/dagster/dagster/_core/execution/calculate_data_time.py
OwenKephart: I wish :( -- we need the storage ID, and get_latest_materialization_event only returns the EventLogEntry, not the EventLogRecord (which contains the storage id)

python_modules/dagster/dagster/_core/definitions/freshness_policy.py
OwenKephart: None gets returned if the data time for any of its upstream keys is None. In these cases, you know something's wrong (you have no record of the materialization of your upstream asset which was used to create this one), but the amount wrong that you are is unknowable.

OwenKephart:
Some good conversation here. In the course of creating the graphql resolvers that I'd need later on, I realized that some significant refactors were needed to the calculate data time functionality, so I'm going to close this pr (deferring the calculation process to here: https://github.com/dagster-io/dagster/pull/10260), then open a separate PR that focuses purely on the sensor logic

PR #10063, opened 2022-10-18
Fix package skipping for dagster-test

jmsanders:
The guarded `setup()` call in dagster-test's setup.py was preventing us from properly parsing its dependency tree.


python_modules/dagster-test/setup.py
```diff
@@ -1,26 +1,25 @@
...
-            "Operating System :: OS Independent",
-        ],
-        packages=find_packages(exclude=["dagster_test_tests*"]),
-        install_requires=[
-            "dagster",
-            "pyspark",
-        ],
-        zip_safe=False,
-    )
+setup(
```
alangenfeld, line 3: how is this currently manifesting? should we log / fail somehow to prevent future similar mistakes?

alangenfeld (approved):

python_modules/dagster-test/setup.py
jmsanders: I think I'm going to add a check to make sure our setup.py files all conform to this (basically, just make sure there's no `__main__` string in any setup.py file).

Otherwise, when we go to parse the distribution, we get back an empty distribution so we don't know about the package's `install_requires` or `extras`.

jmsanders: I spent a little bit of time trying to monkeypatch our distribution parsing to make it look like we're running as `__main__` but decided it wasn't worth it - there are some more notes in https://github.com/dagster-io/dagster/pull/9889/commits/4b6abcfe30ef46831d7a60e74fe3469a5ed427c1

alangenfeld: > when we go to parse the distribution, we get back an empty distribution

could we raise in this condition?

jmsanders: Well, it's not really an _empty_ distribution. It's just whatever you get by default when you call `setup()`. I can look and see if there's any obvious tell that differentiates it from an otherwise vanilla package (like if we have a package with no dependencies).

PR #10064, opened 2022-10-18
fix graph type inference with fan-in

sryza:


python_modules/dagster/dagster/_core/types/dagster_type.py
```diff
@@ -183,6 +183,9 @@ def __eq__(self, other):
     def __ne__(self, other):
         return not self.__eq__(other)
 
+    def __hash__(self):
+        return hash(self.key)
```
alangenfeld, line 187: slight spook on this but i think its fine

PR #10067, opened 2022-10-18
Allow silencing failures with default io manager override

johannkm:
For backcompat in serverless. If the user code image doesn't have the serverless io manager we should fall back to the fs_io_manager. An alternative would be to not set these env vars on old code, but we don't know the version until we've launched it.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/job_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/job_definition.py
@@ -1,5 +1,6 @@
 import importlib
 import os
+import warnings
 from functools import update_wrapper
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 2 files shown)


johannkm:
* master
  * **PR #10067** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10067" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10067?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -837,17 +837,23 @@ def default_job_io_manager(init_context: "InitResourceContext"):
...
+            attr = getattr(module, attribute_name)
+            check.invariant(
+                isinstance(attr, IOManagerDefinition),
+                "DAGSTER_DEFAULT_IO_MANAGER_MODULE and DAGSTER_DEFAULT_IO_MANAGER_ATTRIBUTE must specify an IOManagerDefinition",
+            )
+            with build_resources({"io_manager": attr}) as resources:
+                return resources.io_manager
+        except (ModuleNotFoundError, AttributeError):
+            if not silence_failures:
+                raise
```
johannkm, line 857: is there a way to log a warning here

dpeng817: outside of using warnings.warn there's no context object to route with.

python_modules/dagster/dagster_tests/core_tests/definitions_tests/test_default_io_manager.py
```diff
@@ -85,3 +85,26 @@ def test_asset_override_default_io_manager(instance):
...
+
+def test_bad_override(instance):
+    with environ(
+        {
+            "DAGSTER_DEFAULT_IO_MANAGER_MODULE": "dagster_tests",
+            "DAGSTER_DEFAULT_IO_MANAGER_ATTRIBUTE": "foo_io_manager_def",
+        }
+    ):
+
+        result = execute_job(reconstructable(fs_io_manager_job), instance)
```
dpeng817, line 98: could use raise_on_error=False here and check actual exception a la `pytest.raises`

dpeng817 (approved):
LGTM w/ a few nits

PR #10071, opened 2022-10-18
[refactor] take logic for base asset jobs out of AssetGroup

sryza:
### Summary & Motivation

because AssetGroup is deprecated and will eventually go away


dpeng817 (approved):
Lovely

PR #10072, opened 2022-10-18
RepositoryDefinition.materialize

sryza:
### Summary & Motivation

This allows someone to kick off a materialization of an asset, using the resources and upstream asset definitions that they've included on their repository.

I'm imagining one of the main uses would be to call it from a Jupyter notebook.

It depends on the repository being defined at module scope.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/reconstruct.py
+++ b/python_modules/dagster/dagster/_core/definitions/reconstruct.py
@@ -434,29 +434,51 @@ def reconstructable(target):
             )
         )
 
-    try:
-        if (
-            hasattr(target, "__module__")
-            and hasattr(target, "__name__")
-            and inspect.getmodule(target).__name__ != "__main__"
-        ):
-            return ReconstructablePipeline.for_module(target.__module__, target.__name__)
-    except:
-        pass
-
-    python_file = get_python_file_from_target(target)
-    if not python_file:
+    code_pointer = _find_code_pointer(target)
+    if code_pointer:
+        return bootstrap_standalone_recon_pipeline(code_pointer)
+    else:
         raise DagsterInvariantViolationError(
             "reconstructable() can not reconstruct jobs or pipelines defined in interactive "
             "environments like <stdin>, IPython, or Jupyter notebooks. "
             "Use a pipeline defined in a module or file instead, or use build_reconstructable_job."
         )
 
-    pointer = FileCodePointer(
-        python_file=python_file, fn_name=target.__name__, working_directory=os.getcwd()
-    )
 
-    return bootstrap_standalone_recon_pipeline(pointer)
+def reconstructable_repository(target: "RepositoryDefinition") -> ReconstructableRepository:
+    code_pointer = _find_code_pointer(target)
+
+    if code_pointer:
+        return ReconstructableRepository(code_pointer)
+    else:
+        raise DagsterInvariantViolationError(
+            "Can only reconstruct repositories that are defined within a module, at module scope. "
+            "Cannot reconstruct repositories defined in interactive environments like <stdin>, "
+            "IPython, or Jupyter notebooks."
+        )
+
+
+def _find_code_pointer(target) -> Optional[CodePointer]:
+    try:
+        if (
+            hasattr(target, "__module__")
+            and hasattr(target, "__name__")
+            and inspect.getmodule(target).__name__ != "__main__"
+        ):
+            if getattr(inspect.getmodule(target), target.__name__, None) != target:
+                return None
+
+            return ModuleCodePointer(target.__module__, target.__name__, os.getcwd())
+    except:
+        pass
+
+    python_file = get_python_file_from_target(target)
+    if python_file:
+        return FileCodePointer(
+            python_file=python_file, fn_name=target.__name__, working_directory=os.getcwd()
+        )
+
+    return None
 
 
 @experimental
--- a/python_modules/dagster/dagster/_core/definitions/repository_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/repository_definition.py
@@ -1,8 +1,11 @@
+import itertools
 from abc import ABC, abstractmethod
+from contextlib import ExitStack
 from inspect import isfunction
 from types import FunctionType
 from typing import (
     TYPE_CHECKING,
+    AbstractSet,
     Any,
     Callable,
     Dict,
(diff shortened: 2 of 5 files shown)


python_modules/dagster/dagster/_core/definitions/repository_definition.py
```diff
@@ -1422,6 +1428,66 @@ def get_asset_value_loader(
...
+
+        with ExitStack() as exit_stack:
+            if not instance:
+                if is_dagster_home_set():
+                    instance = exit_stack.enter_context(DagsterInstance.get())
+                else:
+                    instance = exit_stack.enter_context(DagsterInstance.ephemeral())
+
+            return execute_job(
+                ReconstructableJob(reconstructable_repository(self), job_def.name),
```
dpeng817, line 1475: [1]

python_modules/dagster/dagster/_core/definitions/reconstruct.py
```diff
@@ -434,29 +434,48 @@ def make_bar_job():
...
+def reconstructable_repository(target: "RepositoryDefinition") -> ReconstructableRepository:
+    code_pointer = _find_code_pointer(target)
+
+    if code_pointer:
+        return ReconstructableRepository(code_pointer)
+    else:
+        raise DagsterInvariantViolationError(
+            "reconstructable_repository() can only reconstruct repositories that are defined "
+            "within a module, at module scope. It cannot not reconstruct repositories defined in "
+            "interactive environments like <stdin>, IPython, or Jupyter notebooks."
```
dpeng817, line 457: Given that the entrypoint for this error is [1], I don't think it makes sense to make the error in terms of `reconstructable_repository`, instead keeping things focused on the fact that the repo was not constructed at module scope.

python_modules/dagster/dagster/_core/definitions/repository_definition.py
```diff
@@ -1422,6 +1428,66 @@ def get_asset_value_loader(
...
+                instance=instance,
+            )
+
+    def _get_base_job_that_contains_all_asset_keys(
+        self, asset_keys: Sequence[AssetKey]
+    ) -> JobDefinition:
+        for job_name in self.job_names:
+            if is_base_asset_job_name(job_name):
+                job_def = self.get_job(job_name)
+                if ...:
```
dpeng817, line 1486: assuming this needs to be filled in

python_modules/dagster/dagster_tests/core_tests/definitions_tests/test_repository_definition.py
```diff
@@ -1428,3 +1432,32 @@ def the_repo():
         return [the_job]
 
     assert the_repo.get_job("the_job").loggers == {"foo": some_logger}
+
+
+@repository(default_executor_def=in_process_executor)
```
dpeng817, line 1437: would like to see a test with non-module-scoped repo.

dpeng817 (changes requested):
A few comments related to errors, more expansive tests. One if statement that needs to be filled in.

python_modules/dagster/dagster/_core/definitions/reconstruct.py
sryza: 👍

python_modules/dagster/dagster/_core/definitions/repository_definition.py
sryza: yikes - filled

python_modules/dagster/dagster/_core/definitions/reconstruct.py
```diff
@@ -434,29 +434,51 @@ def make_bar_job():
...
         )
 
+    code_pointer = _find_code_pointer(target)
+    if code_pointer:
+        return bootstrap_standalone_recon_pipeline(code_pointer)
+    else:
+        raise DagsterInvariantViolationError(
+            "reconstructable() can not reconstruct jobs or pipelines defined in interactive "
+            "environments like <stdin>, IPython, or Jupyter notebooks. "
+            "Use a pipeline defined in a module or file instead, or use build_reconstructable_job."
```
dpeng817, line 444: nit and I know it predates but you can probably just get rid of the pipeline verbiage here

dpeng817 (approved):
This is sick and makes me happy

python_modules/dagster/dagster/_core/definitions/repository_definition.py
```diff
@@ -1422,6 +1432,72 @@ def get_asset_value_loader(
...
+        selection: Union[AssetSelection, CoercibleToAssetKey, AbstractSet[CoercibleToAssetKey]],
+        instance: Optional[DagsterInstance] = None,
+    ) -> "ExecutionResult":
+        """
+        Executes a run that materializes all the selected assets.
+
+        The repository this is invoked on must be defined at module scope.
+
+        Args:
+            assets (Union[AssetSelection, str, AssetKey, AbstractSet[Union[str, AssetKey]]]):
```
sryza, line 1447: Fix this docstring

dpeng817 (approved):
Awesome - should this be added to the definitions API as well? Or does that auto-inherit from repository?

sryza:
> Awesome - should this be added to the definitions API as well? Or does that auto-inherit from repository?

@dpeng817 `Definitions` doesn't inherit from repository but has a `get_repository_def` method. Not having considered it deeply, I would think that `Definitions` should have the same method. @schrockn - thoughts?

schrockn:
I think the Definitions API is a good opportunity to rethink our strategy around "reconstructable" and the like. Given that we can assume the invariant of one Definitions object per module, we can offer an API that simply takes a module name as a parameter, which I think if far more intuitive and obvious.

```python
runner = DefinitionsRunner(module_name="where_my_code_lives") # CodeLocationRunner maybe?
runner.materialize(...)
```

We could even expose the same API to allow scripting of deployed code locations, which I think could be super interesting.

I also think this is pretty different use case than the DefinitionsRunner in https://github.com/dagster-io/dagster/pull/11315, which is decidedly geared toward in-process test cases.

schrockn (changes requested):
This is for queue management and pending result of discussion not sure we should go down this path.

sryza:
> I think the Definitions API is a good opportunity to rethink our strategy around "reconstructable" and the like.

What are the downsides of our current strategy that you have in mind? My thought had been that an error like "You can only call `Definitions.materialize` with an out-of-process executor if your `Definitions` object lives at module scope." is clear enough, given the rarity of users that are likely to encounter this situation.

Comparing 
```python
from my_module import defs

defs.materialize(...)
```

with
```python
runner = DefinitionsRunner(module_name="my_module")

runner.materialize(...)
```

A couple advantages of the former:
* Fewer classes / concepts.
* Static analysis tooling can tell you if typed the right module name and if that module has defs.


A potential middle ground could be to expose `defs.materialize_in_process`.

PR #10073, opened 2022-10-18
Use instance.python_log_level when determining default system logger level

gibsondan:
Summary:
This makes it so that if you set python_log_level to INFO in your dagster.yaml, it applies to the console output from dagit. I originally held off on this because setting this could affect the celery_executor and docker_celery_executor and celery_k8s_executor, which still read in system logs during the execution - but given that those are off the beaten path and python_log_level is an experimental feature, I think it's OK to set this? We could land it as part of 1.1 if we think there are people out there using this log level flag as well as one of those executors.


gibsondan:
See https://dagster.slack.com/archives/C01U5LFUZJS/p1656428446598809 for discussion / motivation

alangenfeld:
I think the fear that this effects celery executors isn't right, looks like we take the yielded events and write them out as json and thats the messages that get parsed 
https://github.com/dagster-io/dagster/blob/master/python_modules/libraries/dagster-celery/dagster_celery/tasks.py#L61-L72

python_modules/dagster/dagster/_loggers/__init__.py
```diff
@@ -111,13 +112,17 @@ def format(self, record):
     return logger_
 
 
-def default_system_loggers() -> Sequence[Tuple["LoggerDefinition", Mapping[str, object]]]:
+def default_system_loggers(
+    instance: Optional["DagsterInstance"],
```
alangenfeld, line 116: do we need to keep this optional ? worried about missing call sites over time. Maybe an explicit test using `instance_for_test` w/ overrides and `capfd`  ?

gibsondan:
@alangenfeld this is what I was thinking of:https://github.com/dagster-io/dagster/blob/master/python_modules/libraries/dagster-celery-docker/dagster_celery_docker/executor.py#L321-L325

I think its just used in celery docker and celery k8s though, not regular celery

gibsondan:
ahhh @johannkm informs me that those are written to stdout via a separate path

alangenfeld:
ya this is accurate link https://github.com/dagster-io/dagster/blame/master/python_modules/dagster/dagster/_cli/api.py#L318-L341

python_modules/dagster/dagster/_core/execution/context_creation_pipeline.py
```diff
@@ -63,7 +63,7 @@ def initialize_console_manager(
 ) -> DagsterLogManager:
```
gibsondan, line 63: @alangenfeld this is the sticking point for your 'can it be made non-optional' question, there's a long tail of callsites for this off the beaten path where there is no instance currently

gibsondan (commented):
Added a test case in celery-k8s to assuage my paranoia but you both appear to be correct

python_modules/dagster/dagster/_core/execution/context_creation_pipeline.py
alangenfeld: gotcha - maybe a test to ensure its respected in `execute_job` / `execute_in_process` then to protect over time

alangenfeld (approved):

gibsondan:
(added test)

PR #10082, opened 2022-10-18
[Multi-dimensional partitions 3/n] Output multi-dimensional materializations

clairelin135:
- https://github.com/dagster-io/dagster/pull/10001
- https://github.com/dagster-io/dagster/pull/10056

This PR:
- Renames the `CompositePartitionsDefinition` from https://github.com/dagster-io/dagster/pull/9511 to `MultiPartitionsDefinition`
- Adds multi-dimensional partition key tags to runs and asset materializations
- Threads the `MultiDimensionalPartitionKey` object to execution context objects (StepExecutionContext, InputContext/OutputContext) so users can access the partition key object. This will also enable future advanced use cases such as creating a partition mapping object that only depends on a certain upstream partition dimension.


python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
```diff
@@ -74,3 +79,158 @@ def deserialize_partition_from_db_string(
     if partition.startswith("["):
         return deserialize_as(partition, MultiDimensionalPartitionKey)
     return partition
+
+
+class MultiDimensionalPartition(Partition):
```
sryza, line 84: I'm not entirely convinced that this needs to be its own class. If I understand correctly, all it does is add an alias to the value called `partitions_by_dimension`. Thoughts on just using vanilla `Partition`s?

python_modules/dagster/dagster/_core/execution/plan/execute_step.py
```diff
@@ -427,19 +445,24 @@ def _type_check_and_store_output(
         yield evt
 
 
+def _get_output_asset_info(
```
sryza, line 448: This seems like it would make more sense as a method on `OutputContext`. If we don't want to expose `AssetOutputInfo` to users, we could avoid marking it `@public`.

clairelin135:
@sryza I don't have a strong opinion on the naming. I think the most accurate name would be `MultiDimensionPartitionsDefinition`. But that feels like such a mouthful.

To me, `CompositePartitionsDefinition` feels like more of a union operation than a cross-product operation. `MultiPartitionsDefinition` feels less like a union operation, and is a shorter name. `CrossProductPartitionsDefinition` is more accurate, but feels too on the nose and at that point I'd prefer `MultiDimensionalPartitionsDefinition`.

python_modules/dagster/dagster/_core/execution/context/input.py
```diff
@@ -282,16 +299,20 @@ def has_partition_key(self) -> bool:
 
     @public  # type: ignore
     @property
-    def partition_key(self) -> str:
+    def partition_key(self) -> Union[str, "MultiDimensionalPartitionKey"]:
```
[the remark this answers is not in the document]
clairelin135: Ooof, that is annoying.

My guess is that when an asset has multi-dimensional partitions, a user would probably want to be able to fetch partition values per-dimension from the step context. An alternative would be that we could add a `get_multi_dimensional_partition_key` method to the context objects, that will return the partition key object. And then the `partition_key` method will return the stringified multi-dimensional partition key.

I think ideally we would change the return type of the other context methods. If we didn't, we could just return the stringified key and force users to do some `MultiPartitionsDefinition.key_from_string(...)` operation if they wanted to get the object.

python_modules/dagster/dagster/_core/execution/plan/execute_step.py
clairelin135: makes sense! turns out there was an existing method on `OutputContext` that already returned this

python_modules/dagster/dagster/_core/execution/plan/execute_step.py
```diff
@@ -514,10 +540,20 @@ def _get_output_asset_materializations(
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore", category=DeprecationWarning)
 
+                tags = None
+                if isinstance(partition, MultiDimensionalPartitionKey):
```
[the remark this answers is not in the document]
clairelin135: yep, refactored this logic into a helper fxn

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
```diff
@@ -50,3 +60,123 @@ def __new__(cls, partition_dimension_mapping: Mapping[str, str]):
...
+    def __eq__(self, other):
+        return (
+            isinstance(other, MultiPartitionsDefinition)
+            and self.partitions_defs == other.partitions_defs
+        )
+
+    def __hash__(self):
+        return hash(tuple(self.partitions_defs))
+
+    def get_partition_key(self, partition_key_by_dimension: Mapping[str, str]) -> MultiPartitionKey:
```
sryza, line 153: I found this function a tiny bit confusing, because it led me to believe that there was some attribute of the `MultiPartitionsDefinition` that was required to build the `MultiPartitionKey`. If I understand correctly, the purpose of this function is just validation though? If that's the case, it might be preferable to change it into a function named something like `check_matches_partition_dimensions` that accepts a `MultiPartitionKey`? Or could consider leaving it out given that, in the only place this is used, it's us, not the user, who's supplying the arguments?

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
```diff
@@ -50,3 +60,123 @@ def __new__(cls, partition_dimension_mapping: Mapping[str, str]):
...
+            cls,
+            name=check.str_param(name, "name"),
+            partitions_def=check.inst_param(partitions_def, "partitions_def", PartitionsDefinition),
+        )
+
+
+@experimental
+class MultiPartitionsDefinition(PartitionsDefinition):
+    """The set of partitions is the cross product of partitions in the inner partitions
+    definitions"""
```
sryza, line 89: It would be helpful to add an attributes section of the docstring that explains what the keys and values of the  partitions_defs attr mean.

sryza (approved):
A few small comments. Otherwise, this looks good to go!

jtv8:
If you don't mind I have a small suggestion since I've been following this feature with interest!

I notice the plan is here to use a pipe e.g. `a|b` to stringify a multidimensional partition key.  I worry a bit about the edge case where (for whatever reason) one of the partition keys already has a pipe in it.

Would it make any sense to make use JSON serialization and deserialization for these so any special characters will automatically get appropriately escaped? Either in array format:

`["a", "b"]`

Or, for greater human readability (so you know which value corresponds to which dimension), object format:

`{"key1": "a", "key2": "b"}`

Obviously either option would increase the key length a bit - not sure if that has any performance implications?

clairelin135:
Hi @jtv8! Thanks for keeping up with this feature and sharing your suggestion. 

In Dagit, we'll allow for selecting multiple multidimensional partitions using a syntax similar to `"[US...Canada]|2022-10-24"`. Having a static partition containing a pipe character would cause errors in evaluating this, which makes me think that we should make the pipe character invalid altogether.

sryza:
@clairelin135 - in the UI, might it make sense to have two selection boxes instead?

clairelin135:
@sryza The view I'm referring to contained a picker where you could select multiple arbitrary subsets, i.e. `[US...Brazil]|[2022-10-12...2022-10-15], [UK...Canada]|2022-10-15` and this entire selection string would display in a text box

sryza:
@clairelin135 I see - yeah, thinking about it more, I don't think two text boxes would be enough.

That said, we must have some way of escaping characters in that selection syntax, right?  I.e. do we restrict partition keys from including square brackets, commas, and ellipses? If not, we could potentially handle the pipe in the same way in the selection syntax?

clairelin135:
@sryza We don't currently escape those characters. We do check to see if the syntax matches `[something....something]` or if there are commas, but I just tested and this currently breaks if a partition key includes a comma (because we split by commas and assume each split str is a partition). 

We could do a check to only split multidimensional partition keys by the pipe character, but that would still break if a static partition key include the "|" character.

I think we might need to consider making both the comma and pipe characters invalid.

sryza:
@bengotow - thoughts on what we should do here regarding special characters in the selection box backfill dialog? Prohibiting users from using characters in partition keys that they could use before would technically be a breaking change.  If it's just about selection syntax, it feels like there's a way we should be able to support this with proper escaping and stuff.

Looking at the Python code, it seems like we've been prohibiting ellipses but not square brackets.

That said, if we think there are a bunch of situations where special characters would cause pain, maybe it's worth the breakage.

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
clairelin135: That makes sense, I've removed this. If they wanted to, users can use the `MultiPartitionsKey` constructor instead to create a partition key

jtv8:
Given that what this is is essentially "selecting arbitrary partitions with a query", what about a query builder component like this? (Not this specific one necessarily, just so you have a visual example.)

https://www.syncfusion.com/blogs/post/new-react-query-builder-ui-component.aspx

Then any special characters in partition keys can be escaped/encoded appropriately by the UI when it compiles the user's query down to the machine-readable query syntax.

clairelin135:
Hi @jtv8. We discussed this today, and decided to keep the string serialized with the pipe delimiter for two major reasons:
- Better readability
- Displaying information in Dagit is more performant without having to deserialize JSON strings

In order to prevent the pipe character from breaking existing users, we decided to disallow these special characters when they are used within a multipartitions definition, and wait until a major release to make these special characters invalid for static partitions definitions.

The intent here is to allow for a multi-dimensional selection query syntax that can be copy-pasted and manually edited, which I think might be easier to work with than a query builder component.

clairelin135:
@bengotow I've updated the multi-partitions def to restrict a set of special characters for any static partitions dimensions within: `[` `]` `,` `|` 

Any other special characters you think we need to add? The ellipsis is already invalid for static partitions defs

PR #7920, opened 2022-05-16
[asset-resources 2/n][rfc] io manager defs directly on asset defs

dpeng817:
Allows io managers to be placed directly on assets definitions. `io_manager_key` becomes a backcompat argument, and there are 3 different ways that the argument can be specified: as a string key, as a dictionary mapping a key to a definition, or as a definition object only.

PR #10087, opened 2022-10-19
Fix assets with dashes in their path and `io_manager_def`

peay:
### Summary & Motivation

Dagster accepts dashes (`-`) in asset key path components. When automatically creating an operation for assets, dashes are replaced by underscores (`_`).

However, when defining an asset with either `@asset(..., io_manager_def=...)` or `SourceAsset(..., io_manager_def=...)`, it fails if the asset key contains a dash:

```
dagster._check.CheckError: Failure condition: Resource key 'my-repo_test1__io_manager' must be a valid Python identifier.
```

The reason is that the sanitization done for creating an operation is not done when creating an I/O manager resource. This PR adds a method to `AssetKey` to produce a sanitized name that can be used for the operation name as well as in `@asset` and `SourceAsset` for `io_manager_def`. This fixes the issue, and avoids hardcoding this rule in three places.

This does not produce an error anymore:

```py
test_asset = SourceAsset(
   ["some-prefix", "test-asset"],
   io_manager_def=custom_io_manager
)
```

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
@@ -250,8 +250,7 @@ class _Asset:
                 io_manager_def = check.inst_param(
                     self.io_manager, "io_manager", IOManagerDefinition
                 )
-                out_asset_resource_key = "__".join(out_asset_key.path)
-                io_manager_key = f"{out_asset_resource_key}__io_manager"
+                io_manager_key = out_asset_key.to_python_identifier("io_manager")
                 self.resource_defs[io_manager_key] = cast(ResourceDefinition, io_manager_def)
             else:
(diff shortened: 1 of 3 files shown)


peay:
@dpeng817 would you maybe be able to take a look at this one, given that you've worked on https://github.com/dagster-io/dagster/pull/7920? thanks! 😊

dpeng817:
Ahhh wow nice catch.

Build: https://buildkite.com/dagster/dagster/builds/37229 - let's see if any tests need to be fixed up, but this is great.

peay:
Thanks @dpeng817, could you share the test results? I don't have access to the BuildKite link. All was green when running `python -m pytest python_modules/dagster/dagster_tests` locally so maybe that's in one of libraries.

dpeng817 (approved):
@peay apologies, those build results are only accessible by the dagster core team. Relevant tests were green though, this is good to go!

PR #10089, opened 2022-10-19
Hook for loading secrets in grpc/run/step entry points

gibsondan:
Summary:
Adds a new instance component that lets user code load secrets and inject them into the environment (instead of our usual MO which is to rely on the surrounding compute environment to supply secrets, e.g. through kubernetes secrets specified in the Helm chart or injected in the run launcher).

The sketchiest part here is the addition of serialized instance ref as an argument to the grpc command - which would mean that this would only work in situations where some external Python process is managing the construction of grpc servers.

### DIFF
--- a/python_modules/dagster/dagster/_cli/api.py
+++ b/python_modules/dagster/dagster/_cli/api.py
@@ -4,6 +4,7 @@ import logging
 import os
 import sys
 import zlib
+from contextlib import ExitStack
 from typing import Any, Callable, Optional, cast
 
 import click
@@ -20,7 +21,7 @@ from dagster._core.events import DagsterEvent, DagsterEventType, EngineEventData
 from dagster._core.execution.api import create_execution_plan, execute_plan_iterator
 from dagster._core.execution.context_creation_pipeline import create_context_free_log_manager
 from dagster._core.execution.run_cancellation_thread import start_run_cancellation_thread
-from dagster._core.instance import DagsterInstance
+from dagster._core.instance import DagsterInstance, InstanceRef
 from dagster._core.origin import (
     DEFAULT_DAGSTER_ENTRY_POINT,
     PipelinePythonOrigin,
@@ -123,6 +124,7 @@ def _execute_run_command_body(
             recon_pipeline,
             pipeline_run,
             instance,
+            inject_env_vars=True,
         ):
             write_stream_fn(event)
             if event.event_type == DagsterEventType.PIPELINE_FAILURE:
@@ -226,6 +228,7 @@ def _resume_run_command_body(
             pipeline_run,
             instance,
             resume_from_failure=True,
+            inject_env_vars=True,
         ):
             write_stream_fn(event)
             if event.event_type == DagsterEventType.PIPELINE_FAILURE:
@@ -391,6 +394,14 @@ def _execute_step_command_body(
             ),
             step_key=single_step_key,
         )
+
+        location_name = (
+            pipeline_run.external_pipeline_origin.location_name
+            if pipeline_run.external_pipeline_origin
+            else None
+        )
+
+        instance.inject_env_vars(location_name)
 
         if args.should_verify_step:
             success = verify_step(
@@ -573,6 +584,28 @@ def _execute_step_command_body(
     "code from this server.",
     envvar="DAGSTER_CONTAINER_CONTEXT",
 )
+@click.option(
+    "--inject-env-vars-from-instance",
+    is_flag=True,
+    required=False,
+    default=False,
+    help="Whether to load env vars from the instance and inject them into the environment.",
+    envvar="DAGSTER_INJECT_ENV_VARS_FROM_INSTANCE",
+)
+@click.option(
+    "--location-name",
+    type=click.STRING,
+    required=False,
+    help="Name of the code location this server corresponds to.",
+    envvar="DAGSTER_LOCATION_NAME",
+)
+@click.option(
+    "--instance-ref",
+    type=click.STRING,
+    required=False,
+    help="[INTERNAL] Serialized InstanceRef to use for accessing the instance",
+    envvar="DAGSTER_INSTANCE_REF",
+)
 def grpc_command(
     port=None,
     socket=None,
@@ -588,6 +621,9 @@ def grpc_command(
     use_python_environment_entry_point=False,
     container_image=None,
     container_context=None,
+    location_name=None,
+    instance_ref=None,
+    inject_env_vars_from_instance=False,
     **kwargs,
 ):
     from dagster._core.test_utils import mock_system_timezone
@@ -629,11 +665,11 @@ def grpc_command(
             package_name=kwargs["package_name"],
         )
 
-    with (
-        mock_system_timezone(override_system_timezone)
-        if override_system_timezone
-        else seven.nullcontext()
-    ):
+    with ExitStack() as exit_stack:
+
+        if override_system_timezone:
+            exit_stack.enter_context(mock_system_timezone(override_system_timezone))
+
         server = DagsterGrpcServer(
             port=port,
             socket=socket,
@@ -652,6 +688,9 @@ def grpc_command(
             ),
             container_image=container_image,
             container_context=json.loads(container_context) if container_context != None else None,
+            inject_env_vars_from_instance=inject_env_vars_from_instance,
+            instance_ref=deserialize_as(instance_ref, InstanceRef) if instance_ref else None,
+            location_name=location_name,
         )
 
         code_desc = " "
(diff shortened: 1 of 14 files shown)


python_modules/dagster/dagster/_cli/api.py
```diff
@@ -219,6 +228,15 @@ def _resume_run_command_body(
     run_worker_failed = False
 
     try:
+
+        location_name = (
+            pipeline_run.external_pipeline_origin.location_name
+            if pipeline_run.external_pipeline_origin
```
johannkm, line 234: recap, when do runs not get an external_pipeline_origin?

johannkm (commented):
Looks great to me.

I've not been privy to all the discussions, I assume we've considered an option where user code explicitly fetches secrets rather than still getting them from env? Maybe that's something we'd add on top

python_modules/dagster/dagster/_cli/api.py
```diff
@@ -613,11 +664,18 @@ def grpc_command(
...
+        if override_system_timezone:
+            exit_stack.enter_context(mock_system_timezone(override_system_timezone))
+
+        instance = None
+        if load_secrets:
+            with DagsterInstance.from_ref(
+                instance_ref
+            ) if instance_ref else DagsterInstance.get() as instance:
+                instance.inject_secrets_into_environment(location_name)
+
```
alangenfeld, line 678: nit: this is easy to gloss over but pretty fundamental behavior divergence, might be good to add some comments and make it less compact

python_modules/dagster/dagster/_core/secrets/loader.py
```diff
@@ -0,0 +1,10 @@
+from abc import ABC, abstractmethod
+from typing import Dict, Optional
+
+from dagster._core.instance import MayHaveInstanceWeakref
+
+
+class SecretsLoader(ABC, MayHaveInstanceWeakref):
```
alangenfeld, line 7: is it right to name it secrets ? when we talked irl this is more generic stuff-things-in-to-environment loader. Not sure what the better name is but [1]

python_modules/dagster/dagster/_core/secrets/env_file.py
```diff
@@ -0,0 +1,57 @@
...
+
+from dagster._config import StringSource
+from dagster._core.utils import parse_env_var
+from dagster._serdes import ConfigurableClass
+from dagster._utils import merge_dicts
+
+from .loader import SecretsLoader
+
+
+class EnvFileSecretsLoader(SecretsLoader, ConfigurableClass):
```
alangenfeld, line 12: [1] i get bad vibes seeing "secrets" name for thing that loads plain text values from file.

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -2113,6 +2117,13 @@ def get_tick_retention_settings(
         default_tick_settings = get_default_tick_retention_settings(instigator_type)
         return get_tick_retention_settings(tick_settings, default_tick_settings)
 
+    def inject_secrets_into_environment(self, location_name: Optional[str]):
+        # Inject secrets into environment, precedence to first in the list
+        for secret_loader in reversed(self._secrets_loaders):
```
alangenfeld, line 2122: this seems unintuitive to me - i would assume last-write-wins. What led you to this approach?

python_modules/dagster/dagster/_cli/api.py
```diff
@@ -613,11 +664,18 @@ def grpc_command(
...
+        if override_system_timezone:
+            exit_stack.enter_context(mock_system_timezone(override_system_timezone))
+
+        instance = None
+        if load_secrets:
+            with DagsterInstance.from_ref(
+                instance_ref
+            ) if instance_ref else DagsterInstance.get() as instance:
+                instance.inject_secrets_into_environment(location_name)
```
alangenfeld, line 677: need to make sure the failure path here leads to a good user experience

alangenfeld (commented):
Thinking about abstracting this to its fundamental - in the end what we have is just a hook to have things run before loading user code in the user code server or execution code paths. Should we model the abstraction at that level? In that vein should they be configurable classes or just functions (with maybe some schema for values to pass in) since we just run one thing once? I would guess that these "pre code hooks" get used / abused for whatever whacky global poking someone needs even if we model them tightly expecting only environment variable manipulation, so maybe best to just embrace it out the gate.

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -2113,6 +2117,13 @@ def get_tick_retention_settings(
...
+            new_env = secret_loader.load_secrets(location_name)
+            for k, v in new_env.items():
+                os.environ[k] = v
```
alangenfeld, line 2125: would be good to output debug info about observed changes after invoking each hook - "these environment variables were added / changed (removed?)"

python_modules/dagster/dagster/_core/secrets/loader.py
gibsondan: EnvironmentLoader would also work

python_modules/dagster/dagster/_core/instance/__init__.py
gibsondan: I imagined you would enter them in the dagster.yaml in priority order, but either works fine for me

python_modules/dagster/dagster/_cli/api.py
```diff
@@ -573,6 +584,28 @@ def _execute_step_command_body(
...
     envvar="DAGSTER_CONTAINER_CONTEXT",
 )
+@click.option(
+    "--load-env-vars",
+    is_flag=True,
+    required=False,
+    default=False,
+    help="Whether to load env vars from the instance and inject them into the environment.",
+    envvar="DAGSTER_LOAD_ENV_VARS",
+)
```
alangenfeld, line 594: nit: should the name here align with the instance component more directly? `inject-secrets` ? Not public so less critical - maybe at least refer to `SecretsLoader` in the help text

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -2132,6 +2143,14 @@ def get_tick_retention_settings(
         default_tick_settings = get_default_tick_retention_settings(instigator_type)
         return get_tick_retention_settings(tick_settings, default_tick_settings)
 
+    def inject_env_vars(self, location_name: Optional[str]):
+        if not self._secrets_loader:
+            return
+
+        new_env = self._secrets_loader.load_env_vars(location_name)
+        for k, v in new_env.items():
+            os.environ[k] = v
```
alangenfeld, line 2152: should we log what keys are adding/changing values? engine event + metadata?

python_modules/dagster/dagster/_core/secrets/loader.py
```diff
@@ -0,0 +1,13 @@
...
+    @abstractmethod
+    def load_env_vars(self, location_name: Optional[str]) -> Dict[str, str]:
+        pass
```
alangenfeld, line 10: nit: `load_env_vars` naming feels a little off to me since this is loading a mapping of values to be injected in to the environment (probably)

python_modules/dagster/dagster/_core/secrets/loader.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC, abstractmethod
+from typing import Dict, Optional
+
+from dagster._core.instance import MayHaveInstanceWeakref
+
+
+class SecretsLoader(ABC, MayHaveInstanceWeakref):
+    @abstractmethod
+    def load_env_vars(self, location_name: Optional[str]) -> Dict[str, str]:
```
alangenfeld, line 9: is it worth doing a `context` argument here to set-up for future desired information?

python_modules/dagster/dagster/_grpc/server.py
```diff
@@ -227,6 +230,14 @@ def __init__(
         self._container_context = check.opt_dict_param(container_context, "container_context")
 
         try:
+            if load_env_vars:
+                # If arguments indicate it wants to load env vars, use the passed-in instance
+                # ref (or the dagster.yaml on the filesystem if no instance ref is provided)
+                with DagsterInstance.from_ref(
+                    instance_ref
+                ) if instance_ref else DagsterInstance.get() as instance:
+                    instance.inject_env_vars(location_name)
```
alangenfeld, line 239: one annoying failure mode this can introduce is other instance config env vars not being present on the user code server - the same type of problem people hit now when they first use the `instance` in sensor/schedule fns

alangenfeld (approved):
still worried about dealing with this surface area for back-compat over time - I guess well see how it plays out.

python_modules/dagster/dagster/_core/instance/__init__.py
gibsondan: interesting idea in the run case (there isn't an engine event in the grpc server case)

PR #10093, opened 2022-10-19
Expose permissions loading state

salazarm:
### Summary & Motivation

Using this in cloud dagit to figure out when we can decided if a user has permission to do certain things.

cloud

### DIFF
--- a/js_modules/dagit/packages/core/src/app/Permissions.tsx
+++ b/js_modules/dagit/packages/core/src/app/Permissions.tsx
@@ -74,17 +74,32 @@ const extractPermissions = (permissions: PermissionFragment[]) => {
 
 export type PermissionsMap = ReturnType<typeof extractPermissions>;
 
-export const PermissionsContext = React.createContext<PermissionFragment[]>([]);
+export const PermissionsContext = React.createContext<{
+  data: PermissionFragment[];
+  loading: boolean;
+}>({data: [], loading: true});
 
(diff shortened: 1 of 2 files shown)


js_modules/dagit/packages/core/src/app/Permissions.tsx
```diff
@@ -74,17 +74,32 @@ const extractPermissions = (permissions: PermissionFragment[]) => {
...
 };
 
 export const usePermissions = () => {
-  const rawPermissions = React.useContext(PermissionsContext);
-  return React.useMemo(() => extractPermissions(rawPermissions), [rawPermissions]);
+  const {data, loading} = React.useContext(PermissionsContext);
+  return React.useMemo(
+    () => ({
+      ...extractPermissions(data),
+      loading,
```
hellendag, line 99: Seems like `loading` would work better alongside the permissions object, rather than part of it.

```
{
  permissions: extractPermissions(data),
  loading,
}
```

I'm guessing that this would be annoying right now because all the callsites would need to be updated to extract from the `permissions` value? If so, it would be nice to do the refactor in a followup.

hellendag (approved):

js_modules/dagit/packages/core/src/app/Permissions.tsx
salazarm: Yeah it's annoying but its also not a problem, everything is well typed

PR #10099, opened 2022-10-19
Skip test-project image builds

jmsanders:
Our dependency chain here is...really hairy. I don't particularly like this solution but I'm prioritizing the immediate goal faster developer velocity over cleanliness/sanity of code.

Here's our problem: we only know if we need these images built and pushed once we know if we need to run certain Python package builds. Without a larger refactor to switch to a two-phase build of the pipeline (first generate a graph of our build, second compile it), it's really hard to decide if these image builds need to run.

While I hope we eventually move toward that two-phase approach, this takes a major shorcut generating the Python package build steps first and letting them set a marker value in `build_for` if we need the image built. It's bad. It's ugly. It's confusing. I think it might work?

Open to other (simple) suggestions. Otherwise, this is already a very complicated part of our build and I'm fine temporarily increasing its complexity if it unblocks moving onto other build improvements.


.buildkite/dagster-buildkite/dagster_buildkite/steps/test_project.py
```diff
@@ -8,15 +8,17 @@
 from ..step_builder import CommandStepBuilder
 from ..utils import BuildkiteLeafStep, GroupStep
 
+build_for: Set[AvailablePythonVersion] = set()
+build_core_for: Set[AvailablePythonVersion] = set()
```
alangenfeld, line 12: comment re-enforcing assumptions & more verbose name given the level of shenanigans at play

.buildkite/dagster-buildkite/dagster_buildkite/steps/test_project.py
```diff
@@ -113,6 +117,7 @@ def _test_project_step_key(version: AvailablePythonVersion) -> str:
 
 
 def test_project_depends_fn(version: AvailablePythonVersion, _) -> List[str]:
+    build_for.add(version)
```
alangenfeld, line 120: my understanding was that we build the step just with a skip reason attached, so I would assume this would still get invoked. What am I missing?

jmsanders: We're not doing that here - although we probably should for consistency and auditability.

Currently, we're not adding the dependency at all if we have a skip reason:

https://github.com/dagster-io/dagster/blob/46ba805d7ac0258585c65337b3ea2928573a6e09/.buildkite/dagster-buildkite/dagster_buildkite/package_spec.py#L218

alangenfeld (commented):
want to see the renames / comments connecting the dots on the fragile set-up this depends on

.buildkite/dagster-buildkite/dagster_buildkite/steps/test_project.py
jmsanders: yike indeed 😞

.buildkite/dagster-buildkite/dagster_buildkite/steps/test_project.py
```diff
@@ -8,15 +8,30 @@
...
+# somewhere, we use these sets to track internal state about whether or not to build the test- project and
+# test-project-core images.
+#
+# When we build the Buildkite steps for a python package that needs one or both of these images, add the
+# required versions to these sets. We'll otherwise skip building images for any versions not requested here.
+#
+# This means you need to call `build_test_project_steps()` after you've build the other Buildkite steps that
+# require the images.
+#
+# TODO: Don't do this :) More explicitly define the dependencies.
```
alangenfeld, line 22: maybe leave a comment on `package_spec.py#L218` as well since

alangenfeld (approved):
yuck

PR #10119, opened 2022-10-20
Don't gate on elementl emails

jmsanders:
This was originally added to prevent anybody from being able to kick off a build of our internal pipelines with a PR against Dagster. But we should relax this logic for two reasons:

1. We still want to know that a Dagster change is compatible with our Internal code even if we aren't the one making the change.
2. Only members of the Dagster organization in GitHub can kick off builds automatically - builds are off by default for pull requests from forked repositories.

This will unblock the ability for us to commit using our GitHub no-reply email address like https://buildkite.com/dagster/dagster/builds/37140#0183f725-5646-43ba-ac88-247f93e9f1d0


alangenfeld (approved):
> commit using our GitHub no-reply email address

ah interesting failure mode

PR #10121, opened 2022-10-21
Allow singular inputs to with_resources utility

benpankow:
## Summary

`CacheableAssetsDefinition` objects are singular, but might end up generating multiple assets. To make working with `CacheableAssetsDefinition`s feel natural & akin to the assets they represent, this PR allows `with_resources` to apply to singular objects as well as iterables, so users can write e.g.

```python
my_cached_assets = MyCacheableAssetsDefinition()

with_resources(
  my_cached_assets,
  {"foo": bar}
)
```

with the ergonomics identical to if they were working with the assets that will end up being produced

```python
my_assets = [ foo_asset, bar_asset ]

with_resources(
  my_assets,
  {"foo": bar}
)
```

## Test Plan

Unit test.

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/with_resources.py
+++ b/python_modules/dagster/dagster/_core/execution/with_resources.py
@@ -1,4 +1,4 @@
-import collections
+from collections.abc import Iterable as IterableABC
 from typing import Any, Iterable, List, Mapping, Optional, Sequence, TypeVar, Union, cast
 
 from dagster import _check as check
@@ -97,7 +97,7 @@ def with_resources(
                 )
             resource_defs[key] = resource_defs[key].configured(resource_config["config"])
 
-    if not isinstance(definitions, collections.Iterable):
+    if not isinstance(definitions, IterableABC):
         definitions = [definitions]
 
     transformed_defs: List[T] = []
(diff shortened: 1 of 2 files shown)


benpankow:
* master
  * **PR #10121** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10121" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10121?utm_source=stack-comment).

yuhan:
this makes sense to me and i like it. but i'll definitely let more core folks to weight in.

sryza:
I'm pretty open to this - I've been burned by it a few times when writing examples.

That said, maybe we should make `load_assets_from_airbyte_project` return a `Sequence[CacheableAssetsDefinition]` to make it consistent with `load_assets_from_dbt_project`?

python_modules/dagster/dagster/_core/execution/with_resources.py
```diff
@@ -96,6 +97,9 @@ def asset2(context):
                 )
             resource_defs[key] = resource_defs[key].configured(resource_config["config"])
 
+    if not isinstance(definitions, IterableABC):
+        definitions = [definitions]
```
dpeng817, line 101: Nit: prefer to check whether it is directly an instance of the classes we expect if possible, otherwise introduces a potential footgun with bad error message?

dpeng817 (approved):
I like it. Was thinking about it on the original feature but never got around to it, glad we're finally getting it now!

PR #10122, opened 2022-10-21
Fix schema_type of TypedDictTypeLoader

dpeng817:
Changes dictionary config type loader to use Map config type. This allows us to catch errors resulting from mismatches in the inner type at config validation time.

### DIFF
--- a/python_modules/dagster/dagster/_core/types/python_dict.py
+++ b/python_modules/dagster/dagster/_core/types/python_dict.py
@@ -17,19 +17,40 @@ PythonDict = PythonObjectDagsterType(
     dict,
     "PythonDict",
     loader=_dict_input,
-    description="""Represents a python dictionary to pass between solids""",
+    description="""Represents a python dictionary to pass between ops""",
 )
 
 
 class TypedDictLoader(DagsterTypeLoader):
-    def __init__(self, value_dagster_type):
+    def __init__(self, key_dagster_type, value_dagster_type):
+        from ..._config import ConfigTypeKind, Map, ScalarUnion
+
+        self._key_dagster_type = check.inst_param(key_dagster_type, "key_dagster_type", DagsterType)
+        check.param_invariant(self._key_dagster_type.loader, "key_dagster_type")
+
+        # Scalar union config types allow for passing scalar values either directly,
+        # or as a dictionary mapping set keys to a value. When a scalar union config type
+        # is used as the key type for a dictionary, we extract the scalar type
+        # only, so that we only permit passing the scalar value directly.
+        if self._key_dagster_type.loader.schema_type.kind == ConfigTypeKind.SCALAR_UNION:
+            key_schema_type = typing.cast(
+                ScalarUnion, self._key_dagster_type.loader.schema_type
+            ).scalar_type
+        else:
+            key_schema_type = self._key_dagster_type.loader.schema_type
+
         self._value_dagster_type = check.inst_param(
             value_dagster_type, "value_dagster_type", DagsterType
+        )
+        check.param_invariant(self._value_dagster_type.loader, "value_dagster_type")
+        self._schema_type = Map(
+            key_type=key_schema_type,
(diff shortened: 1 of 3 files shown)


dpeng817:
* master
  * **PR #10122** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10122" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10122?utm_source=stack-comment).

python_modules/dagster/dagster/_core/types/builtin_config_schemas.py
```diff
@@ -32,6 +33,15 @@ def define_typed_input_schema_dict(value_config_type):
 
 
 def load_type_input_schema_dict(value):
+    input_schema_dict_format_str = "{'value'|'json'|'pickle': <actual_value>}"
+    if not isinstance(value, collections.Mapping):
+        check.failed(
+            f"Expected input schema dict, but instead received a value of type {type(value)}. An input schema dict is of the format {input_schema_dict_format_str}"
+        )
+    if len(value) > 1 or ("value" not in value and "json" not in value and "pickle" not in value):
+        check.failed(
```
alangenfeld, line 42: is `check` the right type of error to throw here? How do these errors show up in `dagit`? I would expect a dedicated `DagsterInvalidConfigError` something - are these inner errors caught and turned in to that?

also how come this is not being caught at schema validation time? Doesn't the config schema know the valid keys that are possible?

alangenfeld (commented):
On board with the general solution here - this is a nuanced thing that a better error message will help people get through. We should probably make sure the error message points out its due to `Any` type and adding a better type hint would resolve.

alangenfeld (dismissed):
q mgmt

dpeng817 (commented):
How the error manifests in dagit:

python_modules/dagster/dagster/_core/types/builtin_config_schemas.py
alangenfeld: > schema type of a dict type loader is Permissive

ah - thats the root of the problem. We should change it to use the `Map` type that @benpankow  added 
https://github.com/dagster-io/dagster/blame/master/python_modules/dagster/dagster/_config/field_utils.py#L158-L180

alangenfeld (changes requested):

python_modules/dagster/dagster/_core/types/python_dict.py
```diff
@@ -17,19 +17,35 @@ def _dict_input(_context, value):
...
-    def __init__(self, value_dagster_type):
+    def __init__(self, key_dagster_type, value_dagster_type):
+        from ..._config import ConfigTypeKind, Map, ScalarUnion
+
+        self._key_dagster_type = check.inst_param(key_dagster_type, "key_dagster_type", DagsterType)
+        check.param_invariant(self._key_dagster_type.loader, "key_dagster_type")
+        if self._key_dagster_type.loader.schema_type.kind == ConfigTypeKind.SCALAR_UNION:
+            key_schema_type = typing.cast(
+                ScalarUnion, self._key_dagster_type.loader.schema_type
+            ).scalar_type
```
alangenfeld, line 33: whats this about? I think i probably get it but add a clarifying comment

python_modules/dagster/dagster_tests/core_tests/runtime_types_tests/test_python_dict.py
```diff
@@ -214,9 +215,7 @@ def test_dict_type_loader_inner_type_mismatch():
     def emit_dict(dict_input):
         return dict_input
 
-    # TODO: change this depending on the resolution of
-    # https://github.com/dagster-io/dagster/issues/3180
```
alangenfeld, line 218: omg a resolved TODO comment. its a miracle

alangenfeld (approved):
nice

PR #10127, opened 2022-10-21
Cache docstring check

schrockn:
### Summary & Motivation

To refamiliarize myself with the codebase I'm doing some profiling on different scenarios which allows for some quick wins and to see how everything is working end-to-end.

In case I saw that for larger graphs we were making huge numbers of expensive calls to `is_module_available("docstring_parser")` which can be easily cached. This diff assigns the results of that into a global constant.

On a test of constructing 10,000 assets this reduces execution time by 35%.

Ran this script:

```
assets = []

for i in range(0, 10000):
    @asset(name=f"asset_{i}")
    def foo():
        pass

    assets.append(foo)

@repository
def large_repo():
    return assets
```

Under a profiler.

Before:
    
Screen Shot 2022-10-21 at 10 56 08 AM

Screen Shot 2022-10-21 at 10 56 14 AM

After

Screen Shot 2022-10-21 at 10 57 35 AM

Screen Shot 2022-10-21 at 10 57 42 AM

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/inference.py
+++ b/python_modules/dagster/dagster/_core/definitions/inference.py
@@ -7,6 +7,7 @@ from .utils import NoValueSentinel
 
 IS_DOCSTRING_PARSER_AVAILABLE = is_module_available("docstring_parser")
 
+
 class InferredInputProps(NamedTuple):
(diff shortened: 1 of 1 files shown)


sryza (approved):
Nice

PR #10128, opened 2022-10-21
Refactor db io manager to core

jamiedemaria:
### Summary & Motivation
Moves the db io manager to core so it can be used by other db integrations


python_modules/dagster/dagster/_core/storage/db_io_manager.py
```diff
@@ -67,19 +66,24 @@ def get_select_statement(table_slice: TableSlice) -> str:
 
 
 class DbIOManager(IOManager):
-    def __init__(self, type_handlers: Sequence[DbTypeHandler], db_client: DbClient):
+    def __init__(
+        self,
+        type_handlers: Sequence[DbTypeHandler],
+        db_client: DbClient,
+        io_manager_name: str = "DbIOManager",
```
sryza, line 73: Thoughts on just using `self.__class__.__name__`?

python_modules/dagster/dagster/__init__.py
```diff
@@ -98,6 +98,13 @@
     IntSource as IntSource,
     StringSource as StringSource,
 )
+from dagster._core.storage.db_io_manager import (
```
sryza, line 101: If we're going to include these as part of the public API, I think we should mark them as experimental.

jamiedemaria: we could also not put them in the public api at all for now

python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
```diff
@@ -28,11 +34,86 @@ def build_snowflake_io_manager(type_handlers: Sequence[DbTypeHandler]) -> IOMana
...
             from dagster_snowflake_pandas import SnowflakePandasTypeHandler
 
+            @asset(
+                key_prefix=["my_schema"]  # will be used as the schema in snowflake
+            )
+            def my_table():  # the name of the asset will be the table name
+                ...
+
             snowflake_io_manager = build_snowflake_io_manager([SnowflakePandasTypeHandler()])
+            @repository
```
sryza, line 44: It looks like these changes are unrelated to this refactor? Would be preferable to include them separately.

sryza (approved):

PR #10132, opened 2022-10-21
[dagster-airbyte] Docs fix using with_resources with cacheable asset

benpankow:
### Summary & Motivation

The changes to loading Airbyte assets from YAML introduced in #9998 had the unfortunate side-effect of invalidating this docs example. (I'm a little confused as to why our snippet mypy tests didn't catch)

Since the `with_resources` utility only takes an iterable, and the load function now returns a singular `CacheableAssetsDefinition`, we need to wrap it in a list.

#10121 should make it possible to revert this docs change, once landed.

### DIFF
--- a/docs/content/integrations/airbyte.mdx
+++ b/docs/content/integrations/airbyte.mdx
@@ -66,7 +66,7 @@ from dagster_airbyte import airbyte_resource, load_assets_from_airbyte_project
 from dagster import with_resources
 
 airbyte_assets = with_resources(
-    load_assets_from_airbyte_project(project_dir="path/to/airbyte/project"),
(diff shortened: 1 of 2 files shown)


benpankow:
* master
  * **PR #10132** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10132" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10132?utm_source=stack-comment).

rexledesma (approved):
We should surface the mypy issue in Slack, that is concerning - outside the scope of the immediate fix in this PR, so approving

PR #10138, opened 2022-10-21
[docs formatter] Drop mypy ignore lines in docs

benpankow:
## Summary

Docs snippets aren't always totally mypy compliant, but we don't want to dillute them with `casts` everywhere. This lets us mark specific lines of docs snippets to be ignored by mypy & have the linter filter them out when populating the actual docs page, so users don't see them.

#10136 turns mypy bk tests back on.

Changes one file as a demo.

## Test Plan

Tested locally.

### DIFF
--- a/docs/next/util/limit.ts
+++ b/docs/next/util/limit.ts
@@ -1,6 +1,6 @@
 const SPLIT_PATTERN = /\r?\n/;
 const JOIN_PATTERN = '\n';
-
+const MYPY_IGNORE = '# type: ignore';
 /**
  * Use this method to apply a limit for file snippets generated
  * using the ::literalinclude directive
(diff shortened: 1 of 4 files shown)


benpankow:
* master
  * **PR #10138** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10138" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10138?utm_source=stack-comment).

docs/next/util/limit.ts
```diff
@@ -50,8 +50,9 @@ export const limitSnippetLines = (content, fromTo, dedent, startAfter, endBefore
   }
 
   const dedentedElements = elements.map((x) => x.substring(dedentLevel));
+  const filteredElements = dedentedElements.filter((x) => !x.trim().startsWith('# type: ignore'));
```
jmsanders, line 53: The annotations go at the end of the line, not on their own line. I think we need to take the substring up until we see `# type: ignore`

jmsanders (dismissed):

docs/next/util/limit.ts
benpankow: you're totally right - fixed

jmsanders (approved):

PR #10139, opened 2022-10-21
Make AssetKey.__eq__ more efficient

schrockn:
### Summary & Motivation

Noticed another wasteful codepath in asset construction path. In this case the equality operator for AssetKey did an expensive equality step that invoked serialization without caching. When building dictionaries and similar activities this was very wasteful.

Similar to previous change, did this on constructing 10000 asset.

```
from dagster import asset, repository


assets = []

for i in range(0, 10000):
    @asset(name=f"asset_{i}")
    def foo():
        pass

    assets.append(foo)

@repository
def large_repo():
    return assets
```

Before:

After:

Screen Shot 2022-10-21 at 2 58 24 PM

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/events.py
+++ b/python_modules/dagster/dagster/_core/definitions/events.py
@@ -103,7 +103,12 @@ class AssetKey(NamedTuple("_AssetKey", [("path", PublicAttr[List[str]])])):
     def __eq__(self, other):
         if not isinstance(other, AssetKey):
             return False
-        return self.to_string() == other.to_string()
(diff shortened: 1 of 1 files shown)


schrockn:
* master
  * **PR #10139** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10139" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10139?utm_source=stack-comment).

sryza (approved):

PR #10140, opened 2022-10-21
Clicking on top-level concept sections takes you to a page

sryza:
### Summary & Motivation

This also removes a level of hierarchy for top-level concept sections that only have one subpage, like resources and dagster types

### DIFF
--- a/docs/content/_navigation.json
+++ b/docs/content/_navigation.json
@@ -79,6 +79,7 @@
     "children": [
       {
         "title": "Assets",
+        "path": "/concepts/assets/software-defined-assets",
         "children": [
           {
             "title": "Software-Defined Assets",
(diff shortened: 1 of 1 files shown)


erinkcochran87:
One thing I'm not a fan of with an approach like this is that clicking an item with children navigates away from the current page. What if I just want to see what's in there and not actually navigate away from what I'm currently looking at?

I think this is good for items without children (ex: Resources), but perhaps not for items that do have them. Thoughts?

sryza:
@erinkcochran87 - can I ping you on this one?

sryza:
@erinkcochran87 - mind if ping you one more time on this one? Still a source of frustration for me when I'm navigating the docs.

erinkcochran87 (approved):

PR #10143, opened 2022-10-21
[dagit] Updated asset details event view

bengotow:
### Summary & Motivation

This PR implements the designs [here](https://www.figma.com/file/DJWVckz0oILsjoOzWuGpiu/Software-Defined-Assets-(1.1)?node-id=127%3A10708) for the new asset event viewer. At the moment it's functionally equivalent to the old one, but with a shiny new "list + detail" UI.

As we land Asset SLAs and other features, we'll add cards to the top of this overview page with information about partitions / next materialization / SLA, etc.

The designs for this page called for a new asset event search bar and I'm going to work on that in a follow-up because it'll involve some GraphQL and data fetching changes.

image

image

### Test Cases

[x] Asset with Partitions + Materializations
[x] Asset with Partitions + Materializations + Observations
[x] Asset with missing partitions
[x] Source asset with just observations
[] Asset with an "asset lineage" entry
[] Asset with materializations that have description


js_modules/dagit/packages/core/src/assets/AssetEventMetadataEntriesTable.tsx
```diff
@@ -0,0 +1,92 @@
...
+  tr td:first-child {
+    max-width: 300px;
+    width: 25%;
+  }
+  tr td {
+    border: 1px solid ${Colors.KeylineGray};
+    padding: 8px 12px;
+    font-size: 14px;
+  }
+`;
```
hellendag, line 89: - Can this table be full-bleed?
- The vertically-centered contents of each cell look a little off to me, can those be top-aligned with some top padding?

js_modules/dagit/packages/core/src/assets/AssetEventMetadataEntriesTable.tsx
```diff
@@ -0,0 +1,92 @@
...
+        {(observations || []).map((obs) => (
+          <React.Fragment key={obs.timestamp}>
+            {obs.metadataEntries.map((entry) => (
+              <tr key={`metadata-${obs.timestamp}-${entry.label}`}>
+                <td>{entry.label}</td>
+                <td>
+                  <MetadataEntry entry={entry} expandSmallValues={true} />
+                </td>
+                <td style={{opacity: 0.7}}>
+                  <Box flex={{gap: 8, alignItems: 'center'}}>
```
hellendag, line 53: I think a flex-start align will be more correct here, since the text contents are multiline. The icon might need a little top spacing to make it line up nicely with the line-height of the first line of text, though.

js_modules/dagit/packages/core/src/assets/AssetEventMetadataEntriesTable.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const {metadataEntries, timestamp} = event;
+
+  return (
+    <DetailsTable>
+      <tbody>
+        {metadataEntries.map((entry) => (
+          <tr key={`metadata-${entry.label}`}>
+            <td>
+              <Mono>{entry.label}</Mono>
```
hellendag, line 33: This row is mono but the others aren't. I think maybe let's just skip the mono font everywhere in the table?

js_modules/dagit/packages/core/src/assets/AssetPartitionDetail.tsx
```diff
@@ -0,0 +1,134 @@
...
+        border={{side: 'bottom', width: 1, color: Colors.KeylineGray}}
+        flex={{alignItems: 'center'}}
+      >
+        {partition ? (
+          <Box flex={{gap: 12}}>
+            <Heading>{partition}</Heading>
+            {latest ? <Tag intent="success">Materialized</Tag> : <Tag intent="none">Missing</Tag>}
+          </Box>
+        ) : (
+          <Heading>&nbsp;</Heading>
```
hellendag, line 50: An empty heading? Does this take up space? Is there a label we could give it instead?

js_modules/dagit/packages/core/src/assets/CurrentRunsBanner.tsx
```diff
@@ -1,59 +1,50 @@
...
+        title={
+          <div style={{fontWeight: 400}}>
+            {inProgressRunIds.length > 0 && (
+              <>
+                {inProgressRunIds.map((runId) => (
+                  <Link key={runId} to={`/instance/runs/${runId}`}>{`Run: ${titleForRun({
+                    runId,
+                  })}`}</Link>
+                ))}{' '}
+                {inProgressRunIds.length === 1 ? 'is' : 'are'} currently refreshing this asset.
```
hellendag, line 32: Could probably just remove the `is/are`, the text may just not need it

hellendag (approved):
Mostly comments from what I see in the screenshots, overall the code looks fairly straightforward to me. :)

js_modules/dagit/packages/core/src/assets/AssetEventMetadataEntriesTable.tsx
bengotow: Ahh I totally agree with the vertical centering, will fix!

I tried out the full bleed effect (Screenshot below) and I'm a little worried it "lifts" the table out of list + detail layout a bit. Going to leave that for now and we can see what Josh thinks! Just want to make sure people know it's inside the right panel and not a separate panel. 

image

js_modules/dagit/packages/core/src/assets/AssetEventMetadataEntriesTable.tsx
bengotow: Fixed! I put a bit of top spacing on to align with the first line of text 👍

js_modules/dagit/packages/core/src/assets/AssetEventMetadataEntriesTable.tsx
bengotow: Ahh good catch! I think for now I'll monospace the rest of these actually and ask Josh about it when he gets back. We also show these metadata tables inline in the logs and I think it's nice to have consistency between those two. It's arguable whether the log `info` column should be monospace though... maybe we could undo it in both places :-/

js_modules/dagit/packages/core/src/assets/AssetPartitionDetail.tsx
bengotow: Ahh this was lazy of me, let me see what I can do. I wanted the empty state to be identical to the filled state, and this was a cheap way to keep the layout height the same.

PR #10144, opened 2022-10-21
Ignore arg-type on dep dsl

jmsanders:
```
docs_snippets/concepts/ops_jobs_graphs/dep_dsl.py:31: error: Value of type "object" is not indexable  [index]
docs_snippets/concepts/ops_jobs_graphs/dep_dsl.py:42: error: Value of type "object" is not indexable  [index]
docs_snippets/concepts/ops_jobs_graphs/dep_dsl.py:43: error: "object" has no attribute "get"  [attr-defined]
docs_snippets/concepts/ops_jobs_graphs/dep_dsl.py:45: error: Argument "dependencies" to "GraphDefinition" has incompatible type "Dict[NodeInvocation, Dict[Any, DependencyDefinition]]"; expected "Optional[Mapping[
```

I don't think it's worth coercing the type on line 29 to satisfy mypy. Pulling this out into its own commit so I can verify that making the yaml loader's return annotation `dict` doesn't break other things. Otherwise, this is just a small fix related to #10136.


python_modules/dagster/dagster/_utils/yaml_utils.py
```diff
@@ -125,7 +125,7 @@ def merge_yaml_strings(yaml_strs, loader=DagsterRunConfigYamlLoader):
     return functools.reduce(deep_merge_dicts, yaml_dicts, {})
 
 
-def load_yaml_from_path(path: str, loader=DagsterRunConfigYamlLoader) -> object:
```
sryza, line 128: I think we should have this return `Any` instead of `dict`, because in theory yaml doesn't need to be a dictionary.

examples/docs_snippets/docs_snippets/concepts/ops_jobs_graphs/dep_dsl.py
```diff
@@ -42,7 +42,7 @@ def construct_graph_with_yaml(yaml_file, op_defs) -> GraphDefinition:
         name=yaml_data["name"],
         description=yaml_data.get("description"),
         node_defs=op_defs,
-        dependencies=deps,
+        dependencies=deps,  # type: ignore[arg-type]
```
sryza, line 45: It's preferable to avoid "type: ignore"s in user-facing docs snippets. Is there a way we can address this type issue?

sryza (changes requested):

examples/docs_snippets/docs_snippets/concepts/ops_jobs_graphs/dep_dsl.py
jmsanders: It won't be using facing - we strip it out as of #10138

python_modules/dagster/dagster/_utils/yaml_utils.py
jmsanders: That's a good point - maybe we should just go with https://github.com/dagster-io/dagster/pull/10145 and skip the mypy check instead of changing the type hint. It's probably not strictly correct that we're using index access without checking in our example.

PR #10145, opened 2022-10-21
[docs] Misc typing fixes/ignores

benpankow:

### DIFF
--- a/examples/docs_snippets/docs_snippets/concepts/assets/graph_backed_asset.py
+++ b/examples/docs_snippets/docs_snippets/concepts/assets/graph_backed_asset.py
@@ -84,7 +84,7 @@ def middle_asset(upstream_asset):
     return add_one(upstream_asset)
 
 
-middle_asset = AssetsDefinition.from_graph(middle_asset)
+middle_asset = AssetsDefinition.from_graph(middle_asset)  # type: ignore[assignment]
 
 
 @asset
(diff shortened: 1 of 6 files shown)


benpankow:
* master
  * **PR #10145** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10145" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10145?utm_source=stack-comment).

examples/docs_snippets/docs_snippets/concepts/assets/graph_backed_asset.py
```diff
@@ -84,7 +84,7 @@ def middle_asset(upstream_asset):
     return add_one(upstream_asset)
 
 
-middle_asset = AssetsDefinition.from_graph(middle_asset)
+middle_asset = AssetsDefinition.from_graph(middle_asset)  # type: ignore[assignment]
```
sryza, line 87: Not your fault, but we shouldn't be doing this kind of reassignment in a user-facing example. I think it would be best to give the graph a different name like `middle_asset_graph`.

examples/docs_snippets/docs_snippets/concepts/partitions_schedules_sensors/sensors/sensors.py
```diff
@@ -246,7 +246,7 @@ def send_slack_alert():
 # start_cross_repo_run_status_sensor
 
 
-@run_status_sensor(
+@run_status_sensor(  # type: ignore[arg-type]
```
sryza, line 249: I think users are likely to hit this too

I think the preferred fix would be to change the type annotation for `@run_status_sensor` from `Callable[[RunStatusSensorContext], RawSensorEvaluationFunctionReturn]` to `Callable[[...], RawSensorEvaluationFunctionReturn]`

(to reflect that fact that we don't actually require a context object)

benpankow:
- [ ]

PR #10149, opened 2022-10-21
[docs] Misc typing fixes 2

benpankow:
## Summary 

Fixes a few more miscellaneous mypy complaints, mostly through ignores, in our docs snippets.

### DIFF
--- a/examples/docs_snippets/docs_snippets/concepts/ops_jobs_graphs/dep_dsl.py
+++ b/examples/docs_snippets/docs_snippets/concepts/ops_jobs_graphs/dep_dsl.py
@@ -28,7 +28,7 @@ def construct_graph_with_yaml(yaml_file, op_defs) -> GraphDefinition:
 
     deps = {}
 
-    for op_yaml_data in yaml_data["ops"]:
+    for op_yaml_data in yaml_data["ops"]:  # type: ignore[index]
         def_name = op_yaml_data["def"]
         alias = op_yaml_data.get("alias", def_name)
         op_deps_entry = {}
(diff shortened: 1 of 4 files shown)


benpankow:
* master
  * **PR #10145** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10145" target="_blank">Graphite</a> 
    * **PR #10149** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10149" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10149?utm_source=stack-comment).

dpeng817 (changes requested):
As it stands, these type ignores will show up in code snippets, which I think we don't want. You can type ignore either at the top of the file or above the section to avoid.

benpankow:
> As it stands, these type ignores will show up in code snippets, which I think we don't want. You can type ignore either at the top of the file or above the section to avoid.

I merged #10138 which modifies the docs formatter to leave out these type ignores - as it stands I think the only way to type ignore is either the whole file (which might mask actual issues) or inline, so this seemed like the best solution

examples/docs_snippets/docs_snippets/concepts/ops_jobs_graphs/dep_dsl.py
```diff
@@ -28,7 +28,7 @@ def construct_graph_with_yaml(yaml_file, op_defs) -> GraphDefinition:
 
     deps = {}
 
-    for op_yaml_data in yaml_data["ops"]:
+    for op_yaml_data in yaml_data["ops"]:  # type: ignore[index]
```
jmsanders, line 31: I have a PR up specifically for this fix that I think might be a better approach - having `load_yaml_from_path` return a more specific type rather than returning `object`:

https://github.com/dagster-io/dagster/pull/10144

jmsanders (approved):
Definitely some overlap with https://github.com/dagster-io/dagster/pull/10136 but that's fine - I can just rebase if yours merges first.

PR #10151, opened 2022-10-21
[dagster-airbyte] allow filtering connections by directory names

adam-bloom:
### Summary & Motivation
@benpankow and I have been discussing making this change potentially, so thought I'd go ahead and put it in a PR so others can see. I'm using a custom version of this currently and it speeds up startup time by approximately 5x (and I only have 25 connections!).

Unit tests


benpankow (approved):
Looks good to me - thanks for contributing this!

PR #10152, opened 2022-10-21
graph_asset and graph_multi_asset decorators

sryza:
### Summary & Motivation

I was inspired to write this when I came across this example: https://docs.dagster.io/concepts/assets/graph-backed-assets#defining-basic-dependencies-for-graph-backed-assets.

I think `AssetsDefinition.from_graph` is useful in situations where someone has a general / reusable graph that they want to build assets from, but the decorators in this PR will be a lot more ergonomic in situations where someone has specific assets in mind when they define their graph.

This PR also updates the docs to focus primarily on the new decorators: [preview](https://dagster-git-graph-backed-asset-decorator-elementl.vercel.app/concepts/assets/graph-backed-assets).

Added tests

### DIFF
--- a/docs/content/concepts/assets/graph-backed-assets.mdx
+++ b/docs/content/concepts/assets/graph-backed-assets.mdx
@@ -5,29 +5,29 @@ description: Defining a software-defined asset with multiple discrete computatio
 
 # Graph-Backed Assets
 
-[Basic software-defined assets](/concepts/assets/software-defined-assets#a-basic-software-defined-asset) are computed using a single op. If generating an asset involves multiple discrete computations, you can use graph-backed assets by separating each computation into an op and building a graph to combine your computations. This allows you to launch re-executions of runs at the op boundaries but doesn't require you to link each intermediate value to an asset in persistent storage.
-
-Graph-backed assets are useful if you have an existing graph that produces and consumes assets. Wrapping your graph inside a software-defined asset gives you all the benefits of software-defined assets — like cross-job lineage — without requiring you to change the code inside your graph.
-
-Additionally, graph-backed assets allow you to reuse an existing op across other graph-backed assets and jobs, or within the same graph.
+[Basic software-defined assets](/concepts/assets/software-defined-assets#a-basic-software-defined-asset) are computed using a single op. If generating an asset involves multiple discrete computations, you can use graph-backed assets by separating each computation into an op and building a graph to combine your computations. This allows you to launch re-executions of runs at the op boundaries, but doesn't require you to link each intermediate value to an asset in persistent storage.
 
 ---
 
 ## Relevant APIs
 
-| Name                                                       | Description                                   |
-| ---------------------------------------------------------- | --------------------------------------------- |
-| <PyObject object="AssetsDefinition" method="from_graph" /> | Constructs an asset given a graph definition. |
+| Name                                                       | Description                                                                                                                                                              |
+| ---------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| <PyObject object="graph_asset" decorator />                | Decorator for defining an asset that's computed using a graph of ops. The dependencies between the ops are specified inside the body of the decorated function.          |
+| <PyObject object="graph_multi_asset" decorator />          | Decorator for defining a set of assets that are computed using a graph of ops. The dependencies between the ops are specified inside the body of the decorated function. |
+| <PyObject object="AssetsDefinition" method="from_graph" /> | Constructs an asset, given a graph definition. Useful if you have a single graph that you want to use to power multiple different assets.                                |
 
 ---
 
 ## Defining graph-backed assets
 
-To define a graph-backed asset, use the `from_graph` attribute on the <PyObject object="AssetsDefinition" /> object. The value returned from the graph that becomes the graph-backed asset will be stored in persistent storage as the asset:
+To define a graph-backed asset, use the <PyObject object="graph_asset" decorator /> decorator. The decorated function defines the dependencies between a set of ops, which are combined to compute the asset.
+
+In this case, when you tell Dagster to materialize the `slack_files_table` asset, Dagster will invoke `fetch_files_from_slack` and then invoke `store_files` after `fetch_files_from_slack` has completed.
 
 ```python file=/concepts/assets/graph_backed_asset.py startafter=start example endbefore=end example
 import pandas as pd
-from dagster import AssetsDefinition, graph, op
+from dagster import graph_asset, op
 
 
 @op(required_resource_keys={"slack"})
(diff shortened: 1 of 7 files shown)


clairelin135 (approved):
This looks great to me.

Thoughts on naming it `@graph_asset` instead of `@graph_backed_asset`?

sryza:
> Thoughts on naming it `@graph_asset` instead of `@graph_backed_asset`?

I waffled on this, didn't end up with a super strong opinion other than that `@graph_backed_asset` sounded a little less ambiguous. Do you prefer `@graph_asset`? @OwenKephart do you have thoughts?

clairelin135:
I personally prefer `@graph_asset`, because this function is creating a graph and then building assets from the graph. Both sound pretty unambiguous to me though, so generally prefer the shorter name

OwenKephart:
oops missed this ping -- I think `@graph_asset` is more aesthetic and easier to type.

sryza:
`@graph_asset` it is!

sryza:
@clairelin135 mind taking another look at this one? I added a `@graph_multi_asset` decorator and updated the docs. I also updated the PR description to reflect the changes.

docs/content/concepts/assets/graph-backed-assets.mdx
```diff
@@ -5,29 +5,29 @@ description: Defining a software-defined asset with multiple discrete computatio
...
+| <PyObject object="graph_multi_asset" decorator />          | Decorator for defining a set of assets that are computed using a graph of ops. The dependencies between the ops are specified inside the body of the decorated function. |
+| <PyObject object="AssetsDefinition" method="from_graph" /> | Constructs an asset, given a graph definition. Useful if you have a single graph that you want to use to power multiple different assets.                                |
 
 ---
 
 ## Defining graph-backed assets
 
-To define a graph-backed asset, use the `from_graph` attribute on the <PyObject object="AssetsDefinition" /> object. The value returned from the graph that becomes the graph-backed asset will be stored in persistent storage as the asset:
+To define a graph-backed asset, use the <PyObject object="graph_backed_asset" decorator /> decorator. The decorated function defines the dependencies between a set of ops, which are combined to compute the asset.
```
erinkcochran87, line 24: Is it `graph_backed_asset` as it is here, or `graph_asset` as in the examples and the relevant APIs table?

docs/content/concepts/assets/graph-backed-assets.mdx
```diff
@@ -51,79 +51,60 @@ def store_files(files):
...
 
 
 @asset
 def downstream_asset(middle_asset):
-    return middle_asset + 1
+    return middle_asset + 7
 ```
 
-When your graph returns multiple outputs, Dagster infers each output name to be the outputted asset key. In the below example, `two_assets_graph` accepts `upstream_asset` and outputs two assets, `first_asset` and `second_asset`:
+### Graph-backed multi-assets
```
erinkcochran87, line 94: This section should include the specific decorator the user needs to do this. Could you add it pls?

docs/content/concepts/assets/graph-backed-assets.mdx
```diff
@@ -51,79 +51,60 @@ def store_files(files):
...
     return store_files(fetch_files_from_slack())
-
-
-graph_asset = AssetsDefinition.from_graph(store_slack_files_in_sql)
 ```
 
 ### Defining basic dependencies for graph-backed assets
 
-The `from_graph` attribute on the `AssetsDefinition` object infers upstream and downstream asset dependencies from the graph definition provided. In the most simple case when the graph returns a singular output, Dagster infers the name of the graph to be the outputted asset key.
+Similar to with basic op-backed assets, Dagster infers the upstream assets from the names of the arguments to the decorated function.
```
erinkcochran87, line 61: > basic op-backed assets

This strikes me as potentially confusing, as we don't typically talk about assets in this way. Aren't all assets essentially backed by ops? Wouldn't `basic software-defined assets` be the same thing?

erinkcochran87 (commented):
Back to you!

docs/content/concepts/assets/graph-backed-assets.mdx
sryza: Yikes - thanks for catching

docs/content/concepts/assets/graph-backed-assets.mdx
sryza: 👍 👍

docs/content/concepts/assets/graph-backed-assets.mdx
sryza: Sure thing

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -617,6 +618,192 @@ def build_asset_ins(
...
+            partition_mappings=partition_mappings if partition_mappings else None,
+            group_name=self.group_name,
+        )
+
+
+def graph_multi_asset(
+    *,
+    outs: Mapping[str, AssetOut],
+    name: Optional[str] = None,
+    ins: Optional[Mapping[str, AssetIn]] = None,
```
clairelin135, line 752: Can happen in a follow up, but we likely want to support `internal_asset_deps` here

docs/content/concepts/assets/graph-backed-assets.mdx
```diff
@@ -51,79 +51,60 @@ def store_files(files):
...
+@graph_multi_asset(outs={"first_asset": AssetOut(), "second_asset": AssetOut()})
+def two_assets(upstream_asset):
     one, two = two_outputs(upstream_asset)
     return {"first_asset": one, "second_asset": two}
-
-
-two_assets = AssetsDefinition.from_graph(two_assets_graph)
-```
-
-### Defining explicit dependencies for graph-backed assets
```
clairelin135, line 105: I think it would be helpful to still include a brief example of how you can use the same `@graph` to create multiple `AssetsDefinition`s via `AssetsDefinition.from_graph`. This is a fairly common pattern and I think it'll help with understandability to at least briefly include how `keys_by_output_name` can be used

clairelin135 (approved):
Left two comments, but I'm ok with them happening in future follow ups.

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
sryza: Very true

docs/content/concepts/assets/graph-backed-assets.mdx
sryza: That is fair - I will try to take this on in a followup

simonvanderveldt:
@sryza Thanks for this! Makes it quiet a lot nicer to create graph backed Assets :)

Was just testing this and noticed that the description of an Asset created using `@graph_asset(description="foo")` isn't showing up in the Asset whereas for `@asset(description="foo")` it does work.

Example code:
```
from dagster import graph_asset, op

@op
def foo(context):
    return "foo"

@graph_asset(name="foo-bar", description="test")
def asset1():
    return foo()
```
Definition view in Dagit
image

Whereas with `@asset` it looks like this:
```
from dagster import asset

@asset(name="foo-bar", description="test")
def asset1():
    return [1, 2, 3]
```
image

sryza:
@simonvanderveldt - thanks for reporting the description issue. I filed an issue for this: https://github.com/dagster-io/dagster/issues/12569.

PR #10163, opened 2022-10-24
Allow overriding svg with color

salazarm:
### Summary & Motivation

We want to be able to override the color of icons that have default colors.

Screen Shot 2022-10-24 at 10 04 03 AM


hellendag (approved):
lgtm

PR #10164, opened 2022-10-24
Fix AssetsDefinition.with_prefix_or_group to update partition mappings

peay:
### Summary & Motivation

`AssetsDefinition.with_prefix_or_group` can be used to add a prefix or group to either the inputs or outputs of an assets definition.

Currently, partition mappings are not being updated when the inputs are renamed. Hence, after renaming an input from `A` to `B`, the partition mapping for A is lost.

This PR modifies `with_prefix_or_group` to rename the keys in `AssetsDefinition._partition_mappings`.

Notes:

- Should we also subset the partition mappings in [`AssetsDefinition.subset_for`](https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_core/definitions/assets.py#L717) a bit further below? I haven't used this particular API myself, but it looks like it might unnecessarily include partition mappings for inputs that have not been selected. cc @clairelin135 

- Added a unit test

cc @sryza

### DIFF
--- a/python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets.py
+++ b/python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets.py
@@ -1,4 +1,5 @@
 import ast
+import datetime
 import tempfile
 
 import pytest
@@ -8,10 +9,12 @@ from dagster import (
     AssetOut,
     AssetsDefinition,
     DagsterEventType,
+    DailyPartitionsDefinition,
     EventRecordsFilter,
     GraphOut,
     IOManager,
     IOManagerDefinition,
+    LastPartitionMapping,
     Out,
     Output,
     ResourceDefinition,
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets.py
```diff
@@ -31,6 +34,7 @@
 from dagster._core.errors import DagsterInvalidDefinitionError, DagsterInvalidInvocationError
 from dagster._core.storage.mem_io_manager import InMemoryIOManager
 from dagster._core.test_utils import instance_for_test
+from python_modules.dagster.dagster._core.definitions import partition_mapping
```
sryza, line 37: I don't think this import is needed

python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets.py
```diff
@@ -127,6 +131,27 @@ def ma_op():
     assert subset.group_names_by_key[AssetKey("b")] == "bar"
 
 
+def test_retain_partition_mappings():
+    @asset(
+        ins={"input_last": AssetIn(["input_last"], partition_mapping=LastPartitionMapping())},
+        partitions_def=DailyPartitionsDefinition(datetime.datetime(2022, 1, 1)),
+    )
+    def bar_(input_last):
+        pass
```
sryza, line 140: The linter complains that `input_last` is an unused import. You could fix this by replacing `pass` with `del input_last`.

sryza (changes requested):
Thanks for fixing this. Left a couple small comments, other than those it should be good to go.

python_modules/dagster/dagster_tests/core_tests/asset_defs_tests/test_assets.py
peay: Definitely a typo, fixed, thanks!

sryza (approved):
This looks great. Merging now.

PR #10006, opened 2022-10-13
Asset versioning UI

smackesey:
### Summary & Motivation

V1 of asset versioning UI code

### How I Tested These Changes

Manipulated the `asset_reconciliation` example in dagit with the new UI.

PR #10004, opened 2022-10-13
Asset versioning example

smackesey:
### Summary & Motivation

This adds an example for playing around with logical versioning
functionality.

### How I Tested These Changes

Loaded up the example in dagit with `dagit -w examples/assets_reconciliation/workspace.yaml`.

PR #10167, opened 2022-10-24
Add version to asset decorator

smackesey:
### Summary & Motivation

This PR adds a new `op_version` arg to the asset decorator that is passed directly to the underlying op. This will be improved in a future PR to work with a dictionary mapping outputs to versions.

New unit test.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
@@ -13,8 +13,6 @@ from typing import (
     overload,
 )
 
-from typing_extensions import Final
-
 import dagster._check as check
 from dagster._builtins import Nothing
 from dagster._config import UserConfigSchema
@@ -41,9 +39,6 @@ from ..partition import PartitionsDefinition
 from ..policy import RetryPolicy
 from ..resource_definition import ResourceDefinition
 from ..utils import DEFAULT_IO_MANAGER_KEY, NoValueSentinel
-
-DEFAULT_OP_VERSION: Final[str] = "DEFAULT"
-
 
 @overload
 def asset(
@@ -74,7 +69,7 @@ def asset(
     output_required: bool = ...,
     freshness_policy: Optional[FreshnessPolicy] = ...,
     retry_policy: Optional[RetryPolicy] = ...,
-    version: Union[bool, str] = ...,
+    op_version: Optional[str] = ...,
 ) -> Callable[[Callable[..., Any]], AssetsDefinition]:
     ...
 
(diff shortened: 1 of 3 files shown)


smackesey:
* master
  * **PR #10003** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10003" target="_blank">Graphite</a> 
    * **PR #10005** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10005" target="_blank">Graphite</a> 
      * **PR #9899** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/9899" target="_blank">Graphite</a> 
        * **PR #10166** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10166" target="_blank">Graphite</a> 
          * **PR #10167** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10167" target="_blank">Graphite</a>  👈
        * **PR #10006** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10006" target="_blank">Graphite</a> 
    * **PR #10004** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10004" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10167?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -144,6 +150,11 @@ def asset(
         output_required (bool): Whether the decorated function will always materialize an asset.
             Defaults to True. If False, the function can return None, which will not be materialized to
             storage and will halt execution of downstream assets.
+        version (Union[bool, str]): (Experimental) If `false` is passed, the asset is unversioned.
```
sryza, line 153: This explanation refers to concepts that might not have an obvious meaning for readers.

Thoughts trying to phrase it in terms of the specific impact that setting it will have?

E.g. something like:

> If an asset has an op version that's different than the op version used for its latest materialization, then it and all downstream assets will be considered stale. Conversely, if two materializations for the same asset used the same op version and had the same upstream inputs, then the second materialization won't mark any new downstream assets stale.

Also accepting boolean values seems unintuitive to me. Thoughts on accepting and defaulting to `None` instead? Apologies if we've already discussed this and I'm re-treading old ground.

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -208,6 +220,7 @@ def __init__(
         op_tags: Optional[Dict[str, Any]] = None,
         group_name: Optional[str] = None,
         output_required: bool = True,
+        version: Union[bool, str] = False,
```
sryza, line 223: What do you think about calling this "op_version" to make it clear that it's the version of the computation, not the version of the asset?

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
smackesey: This PR is prior to logical versions on materializations, which are upstack, so further explanation on this arg should be in one of those PRs.

sryza (approved):

PR #10172, opened 2022-10-24
Allow default io_manager load_input to support partitions of differing frequencies

### LINKED ISSUE #10120: in built-in object IO managers, handle loading objects for multiple partitions
If a daily asset depends on an hourly asset, then each partition of the daily asset will correspond to 24 partitions of the hourly asset.

When materializing the daily asset and `load_input` is called to load the contents of the hourly asset, we should return the contents of those 24 partitions.

For the Pandas type handler of the Snowflake IO manager, we handle this by returning a single dataframe that contains the concatenated contents of all the hourly partitions.

For the built-in object store IO managers, we could just return a list of the pickled objects. Or potentially a dictionary keyed by partition?

What we've heard:
- 2022-10-20: https://dagster.slack.com/archives/C01U954MEER/p1666175125167899

Here's a start on the implementation:
```python
from dagster import io_manager, IOManager, AssetKey


@io_manager
def my_io_manager():
    return MyIOManager()


class MyIOManager(IOManager):
        ...

    def _load(self, asset_key: AssetKey):
        ...

    def _load_partition(self, asset_key: AssetKey, partition_key: str):
        ...

        if context.has_asset_partitions:
            partition_key_range = context.asset_partition_key_range
            if partition_key_range.start == partition_key_range.end:
                return self._load_partition(context.asset_key, partition_key_range.start)
            else:
                return [
                    self._load_partition(context.asset_key, partition_key)
                    for partition_key in context.asset_partition_keys
                ]
        else:
            return self._load(context.asset_key)
```

andrewgryan:
### Summary & Motivation

The default `io_manager` raises an exception when a partitioned asset depends on a higher frequency partitioned asset, if the resulting `partition_key_range start !== end`. In the use case considered here a daily asset that depends on hourly assets would like to have access to all 24 hourly partitions. For example, to generate an asset that summarises the hourly values in some way, perhaps reports or statistics generation. 

Minor refactorings were carried out on the existing default IOManager to support partitioned assets explicitly, an improvement may be to continue the refactoring to abstract the path details away from the load/save logic.

Fixes #10120 

I wrote a unit test that simulates the above use case. It recreated the error reported in the slack channel and simulated the situation using real materialisations. The "unit" test itself is probably more in the category of integration test and should be optimised to run faster (since it takes 6 seconds on my macbook).

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/fs_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/fs_io_manager.py
@@ -145,7 +145,7 @@ class PickledObjectFilesystemIOManager(MemoizableIOManager):
 
         return os.path.join(self.base_dir, *path)
 
-    def _get_path_partition(self, asset_key: AssetKey, partition_key: str) -> str:
+    def _get_path_for_partition(self, asset_key: AssetKey, partition_key: str) -> str:
         """Construct filepath for a particular partition_key"""
         return os.path.join(self.base_dir, *asset_key.path, partition_key)
 
@@ -216,7 +216,7 @@ class PickledObjectFilesystemIOManager(MemoizableIOManager):
             # Multiple partition load
             partition_keys = context.asset_partition_keys
             paths = [
-                self._get_path_partition(context.asset_key, partition_key)
+                self._get_path_for_partition(context.asset_key, partition_key)
                 for partition_key in partition_keys
             ]
             return {key: self._load_pickle(path) for (key, path) in zip(partition_keys, paths)}
--- a/python_modules/dagster/dagster/_core/storage/fs_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/fs_io_manager.py
@@ -199,6 +199,24 @@ class PickledObjectFilesystemIOManager(MemoizableIOManager):
 
     def load_input(self, context):
         """Unpickle the file and Load it to a data object."""
+        if context.has_asset_partitions:
+            return [
+                self._load_pickle(self._partition_path(context.asset_key, partition_key))
+                for partition_key in context.asset_partition_keys
+            ]
+        else:
+            return self._load_input(context)
+
+    def _partition_path(self, asset_key, partition_key):
+        return os.path.join(self.base_dir, *asset_key.path, partition_key)
+
+    def _load_pickle(self, filepath):
+        with open(filepath, self.read_mode) as read_obj:
+            return pickle.load(read_obj)
+
+    def _load_input(self, context):
+        """Unpickle the file and Load it to a data object."""
+
         check.inst_param(context, "context", InputContext)
 
         if context.dagster_type.typing_type == type(None):
--- a/python_modules/dagster/dagster/_core/storage/fs_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/fs_io_manager.py
@@ -145,6 +145,10 @@ class PickledObjectFilesystemIOManager(MemoizableIOManager):
 
         return os.path.join(self.base_dir, *path)
 
+    def _get_path_partition(self, asset_key: AssetKey, partition_key: str) -> str:
+        """Construct filepath for a particular partition_key"""
+        return os.path.join(self.base_dir, *asset_key.path, partition_key)
+
     def has_output(self, context):
         filepath = self._get_path(context)
 
@@ -212,7 +216,7 @@ class PickledObjectFilesystemIOManager(MemoizableIOManager):
             # Multiple partition load
             partition_keys = context.asset_partition_keys
             paths = [
-                self._partition_path(context.asset_key, partition_key)
+                self._get_path_partition(context.asset_key, partition_key)
                 for partition_key in partition_keys
             ]
             return {key: self._load_pickle(path) for (key, path) in zip(partition_keys, paths)}
@@ -221,9 +225,6 @@ class PickledObjectFilesystemIOManager(MemoizableIOManager):
             filepath = self._get_path(context)
             context.add_input_metadata({"path": MetadataValue.path(os.path.abspath(filepath))})
             return self._load_pickle(filepath)
-
-    def _partition_path(self, asset_key: AssetKey, partition_key: str) -> str:
-        return os.path.join(self.base_dir, *asset_key.path, partition_key)
 
     def _load_pickle(self, filepath: str):
         """Unpickle the file and Load it to a data object."""
--- a/python_modules/dagster/dagster/_core/storage/fs_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/fs_io_manager.py
@@ -212,7 +212,11 @@ class PickledObjectFilesystemIOManager(MemoizableIOManager):
             key_range = context.asset_partition_key_range
             return key_range.start != key_range.end
 
-        if context.has_asset_partitions and has_multiple_partitions(context):
+        if (
+            context.has_input_name
+            and context.has_asset_partitions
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/storage/fs_io_manager.py
```diff
@@ -199,6 +199,24 @@ def handle_output(self, context, obj):
...
+                self._load_pickle(self._partition_path(context.asset_key, partition_key))
+                for partition_key in context.asset_partition_keys
+            ]
+        else:
+            return self._load_input(context)
+
+    def _partition_path(self, asset_key, partition_key):
+        return os.path.join(self.base_dir, *asset_key.path, partition_key)
+
+    def _load_pickle(self, filepath):
```
sryza, line 213: If we're going to put this in its own function, I think it's worth using in the other site inside the class where we `pickle.load`.

python_modules/dagster/dagster/_core/storage/fs_io_manager.py
```diff
@@ -199,6 +199,24 @@ def handle_output(self, context, obj):
 
     def load_input(self, context):
         """Unpickle the file and Load it to a data object."""
+        if context.has_asset_partitions:
+            return [
+                self._load_pickle(self._partition_path(context.asset_key, partition_key))
```
sryza, line 204: On the GH issue you files, you proposed including the partition keys in the result somehow. Any particular reason you moved away from this?

python_modules/dagster/dagster/_core/storage/fs_io_manager.py
```diff
@@ -199,6 +199,24 @@ def handle_output(self, context, obj):
 
     def load_input(self, context):
         """Unpickle the file and Load it to a data object."""
+        if context.has_asset_partitions:
```
sryza, line 202: For the single-partition case, this would be a behavior change, right?

If we were building this from the ground up, I could see the wisdom in handling partitions in a unified way, independent of whether there's a single one or plural.

However, given our compatibility guarantees, I think we need to maintain the existing behavior for single partitions.

python_modules/dagster/dagster/_core/storage/fs_io_manager.py
```diff
@@ -199,6 +199,24 @@ def handle_output(self, context, obj):
...
+            return [
+                self._load_pickle(self._partition_path(context.asset_key, partition_key))
+                for partition_key in context.asset_partition_keys
+            ]
+        else:
+            return self._load_input(context)
+
+    def _partition_path(self, asset_key, partition_key):
```
sryza, line 210: Not required, but if you feel up to adding type annotations to the argument types and return types, it would be helpful

python_modules/dagster/dagster/_core/storage/fs_io_manager.py
sryza: One more thing: I noticed that our current I/O manager has special handling for `if context.dagster_type.typing_type == type(None):`. I think we'd want to retain that even in the case of multiple asset partitions.

sryza (changes requested):
Thanks for submitting this! I left a few comments.

python_modules/dagster/dagster/_core/storage/fs_io_manager.py
andrewgryan: I agree with both comments, I added a unit test to check the backwards compatible use case and I reinstated the dagster_type check

python_modules/dagster/dagster/_core/storage/fs_io_manager.py
andrewgryan: I've added type annotations, did you want me to take this opportunity to add more type annotations to the signature of `load_input` or is it sufficient to just type annotate the methods I added?

python_modules/dagster/dagster/_core/storage/fs_io_manager.py
andrewgryan: No particular reason, I just went with the minimal implementation to begin with. I've added a key/value dict to make it easier for consuming code to look up fields. Users that want a list of values can use `.values()` method of dict.

python_modules/dagster/dagster/_core/storage/fs_io_manager.py
andrewgryan: I agree, to DRY it up I've also updated the existing load logic to call the `_load_pickle` private method

andrewgryan:
@sryza in the spirit of extending the existing code, what would be a natural `metadata` value to pass along? In the existing case `context.add_input_metadata({"path": MetadataValue.path(os.path.abspath(filepath))})` is forwarded to the user/dagit.

Would it make sense to pass `paths` with a {key: path} dict through the metadata so that the user can see in the browser what has been created? I don't know enough about the metadata mechanism in dagster to do the "right" thing.

python_modules/dagster/dagster/_core/storage/fs_io_manager.py
sryza: Annotating the methods you added is great!

sryza (commented):
This looks great.  A few last things:

It looks like there are a couple test failures:
```
dagster_tests/core_tests/test_multiple_outputs.py::test_multiple_outputs_only_emit_one_multiproc FAILED [  3%]
dagster_tests/core_tests/test_pipeline_execution.py::test_multiproc_reexecution_fs_storage_after_fail FAILED [  6%]
```

There are a few lint complaints:
```
[2022-10-25T15:26:55Z] ************* Module dagster_tests.core_tests.partition_tests.test_partitioned_io_manager
--
  | [2022-10-25T15:26:55Z] dagster_tests/core_tests/partition_tests/test_partitioned_io_manager.py:1:0: W0611: Unused AssetKey imported from dagster (unused-import)
  | [2022-10-25T15:26:55Z] dagster_tests/core_tests/partition_tests/test_partitioned_io_manager.py:1:0: W0611: Unused build_op_context imported from dagster (unused-import)
  | [2022-10-25T15:26:55Z] dagster_tests/core_tests/partition_tests/test_partitioned_io_manager.py:1:0: W0611: Unused build_input_context imported from dagster (unused-import)
  | [2022-10-25T15:26:55Z] dagster_tests/core_tests/partition_tests/test_partitioned_io_manager.py:1:0: W0611: Unused build_output_context imported from dagster (unused-import)
```

Last of all, running `make isort` will stop the isorter from complaining.

andrewgryan:
@sryza I fixed the test suite, ran make isort and renamed the partition name method to describe what it does more closely and to make it be as similar to `_get_path` so future devs can find the method easily by tab completing or grepping.

sryza:
It looks like `make isort` has one last complaint:
```
ERROR: /workdir/python_modules/dagster/dagster_tests/core_tests/partition_tests/test_partitioned_io_manager.py Imports are incorrectly sorted and/or formatted.
```

Otherwise, this looks great to me.

Also, I just noticed that I missed your earlier comment:

>  in the spirit of extending the existing code, what would be a natural metadata value to pass along

I think it's fine to leave it out for now, but one option would be to use a JSON metadata type to hold a dict of partition keys to paths. Or just a list of paths.

sryza (approved):
LGTM!

danielgafni:
@sryza any reason why `pre-commit` isn't used for formatting & linting on the user side automatically?

sryza:
@danielgafni no particular reason - just not something we've had the chance to build

danielgafni:
seems like it would dramatically reduce the amount of issues with failing linters in CI... the only problem I see is pylint as it's too slow to run for every commit

PR #10175, opened 2022-10-24
Snowflake IO Manager API docs

jamiedemaria:
### Summary & Motivation
adds more in-depth api docs for snowflake io manager


python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
```diff
@@ -34,11 +34,85 @@ def build_snowflake_io_manager(type_handlers: Sequence[DbTypeHandler]) -> IOMana
...
+        .. code-block:: YAML
+
+            database: my_database  # name of the database in snowflake
+            account: my_account  # snowflake account
+            user: user@email.com  # user account
+            password: abc123  # account password
+            warehouse: my_warehouse # snowflake warehouse
+            schema: my_schema  # name of the schema for the tables
+
+        You can also point all configuration to environment variables where dagster will find the values
```
sryza, line 67: Dagster should be capitalized

python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
```diff
@@ -34,11 +34,85 @@ def build_snowflake_io_manager(type_handlers: Sequence[DbTypeHandler]) -> IOMana
...
+                        "user: ...,
+                        ...
+                    })}
+                )
+
+        The returned Snowflake IOManager can be configured with the following values:
+
+        .. code-block:: YAML
+
+            database: my_database  # name of the database in snowflake
```
sryza, line 60: Do we have a standard way of documenting the config values that a resource accepts? If not, I think it's worth creating one. It feels a little awkward to have it inside the examples section.

python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
```diff
@@ -34,11 +34,85 @@ def build_snowflake_io_manager(type_handlers: Sequence[DbTypeHandler]) -> IOMana
...
-                    ...
+            @op(
+                out={"my_table": Out(metadata={"schema": "my_schema"})}
+            )
+            def make_my_table():
+                # the returned value will be stored at my_schema.my_table
+                ...
+
+            @asset(
+                key_prefix=["my_schema"]
```
sryza, line 99: This could go on the same line as `@asset`

python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
```diff
@@ -34,11 +34,85 @@ def build_snowflake_io_manager(type_handlers: Sequence[DbTypeHandler]) -> IOMana
...
+                # the returned value will be stored at my_schema.my_other_table
+                ...
+
+        To only use specific columns of a table as input to a downstream op or asset, add the metadata "columns" to the
+        In or AssetIn.
+
+        .. code-block:: python
+
+            @asset(
+                in={"my_table": AssetIn("my_table": metadata={"columns": ["a"]})}
```
sryza, line 111: `in` -> `ins`. also, looks like there's some sort of typo inside the `AssetIn` arguments.

python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
```diff
@@ -34,11 +34,85 @@ def build_snowflake_io_manager(type_handlers: Sequence[DbTypeHandler]) -> IOMana
...
+            )
+            def my_table_a(my_table):
```
sryza, line 113: I think the current snowflake IO manager only works if the input has a type annotation, right? If so, would be good to include one.

python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
sryza: The schema / key_prefix thing is already demonstrated in an example above, right?

python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
jamiedemaria: for resources that aren't in a wrapper function the docs site will automatically document the config ([example](https://docs.dagster.io/_apidocs/libraries/dagster-snowflake#dagster_snowflake.snowflake_resource)) but i don't know if there's a way to force that in the api docs for this io manager since it's wrapped in a function. I'll ask erin and see if she knows

sryza:
@jamiedemaria when you're ready for another round of review, re-request me

sryza (approved):
This looks great

PR #10179, opened 2022-10-24
Make handleLaunchResult agnostic to the query that returned the data

salazarm:
### Summary & Motivation

I want to reuse `handleLaunchResult` in cloud with a different query that returns data of the same type as `LaunchPipelineExecution`.

In order to reuse this in cloud without leaking the cloud mutation into OSS I need to make this function agnostic to the query that returned the data and instead make it take the actual inner data which is the same for the new query as the existing 2 launch run queries.

Testing in cloud


js_modules/dagit/packages/core/src/runs/RunUtils.tsx
```diff
@@ -65,24 +64,17 @@ export type LaunchBehavior = 'open' | 'open-in-new-tab' | 'toast';
 
 export function handleLaunchResult(
   pipelineName: string,
-  result: void | {data?: LaunchPipelineExecution | LaunchPipelineReexecution | null},
+  result: void | null | LaunchPipelineExecution_launchPipelineExecution,
```
hellendag, line 67: Might there be any problems with this type with regard to `LaunchPipelineReexecution`? E.g. is there a parent or root run ID that needs to be taken into account here? It looks like TS passes, so maybe it's fine?

hellendag (approved):

js_modules/dagit/packages/core/src/runs/RunUtils.tsx
salazarm: Yeah it's fine since we're not accessing any of the fields specific to the underlying types here

js_modules/dagit/packages/ui/src/components/MetadataTable.tsx
```diff
@@ -5,11 +5,11 @@ import {Box} from './Box';
 import {Colors} from './Colors';
 import {Table, TableProps} from './Table';
 
-type Row = {key: string; value: React.ReactNode};
+export type MetadataTableRow = {key: string; label?: React.ReactNode; value: React.ReactNode};
```
hellendag, line 8: Revert the changes in here?

hellendag (approved):

js_modules/dagit/packages/ui/src/components/MetadataTable.tsx
salazarm: I need this to allow rendering the Info icon. Current key can only be a string since it's used as the `key` on the react node.

PR #10180, opened 2022-10-25
EventLogRecord.partition_key and EventLogRecord.asset_key

sryza:
### Summary & Motivation

Reaching down deep into EventLogRecords creates a lot of boilerplate in our code base.

On the other hand, I marked it as an RFC, because this is a slippery slope, and as soon as we start exposing some stuff on EventLogRecord, what's to stop us from exposing everything??

### DIFF
--- a/python_modules/dagster/dagster/_core/event_api.py
+++ b/python_modules/dagster/dagster/_core/event_api.py
@@ -29,6 +29,22 @@ class EventLogRecord(NamedTuple):
     storage_id: PublicAttr[int]
     event_log_entry: PublicAttr[EventLogEntry]
 
+    @property
+    def asset_key(self) -> Optional[AssetKey]:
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/event_api.py
```diff
@@ -29,6 +29,22 @@ class EventLogRecord(NamedTuple):
...
+
+        return None
+
+    @property
+    def partition_key(self) -> Optional[str]:
+        dagster_event = self.event_log_entry.dagster_event
+        if dagster_event:
+            return dagster_event.partition
+
+        return None
```
alangenfeld, line 46: even if we expose more stuff, whats the cost? These are simple in concept and implementation so are near zero cost in my eyes

alangenfeld (approved):

sryza:
Ok great

PR #10181, opened 2022-10-25
Break up sensor_definition.py

sryza:
### Summary & Motivation

It has gotten quite big.

This PR moves asset sensors and multi-asset sensors into their own files.


jamiedemaria (approved):
nice! this will be so much better

PR #10184, opened 2022-10-25
Make PythonPackages.get() more flexible to _/-

jmsanders:
Basically the same thing we already do, but in reverse. Since we remain very inconsistent about when we refer to a package with its dashes and when we refer to it with its underscores because pip generally treats the two as synonyms.


alangenfeld (approved):
🤷

PR #10186, opened 2022-10-25
[dagit] Make 50-level colors opaque

hellendag:
### Summary & Motivation

The 50-level colors in `@dagster-io/ui` are all low-opactiy values, which means that we can end up with weird states where rendering these colors in layered situations. For instance, when hovering over a `Tag` with its `data-tooltip` visible, the translucency breaks the UI:

Screen Shot 2022-10-25 at 10 01 01 AM

I'm not really sure that we need these to be low-opacity colors. Our primary background colors are white and light gray, and these light colors should look roughly the same on these backgrounds irrespective of whether they're rendered as opaque or translucent. This could end up being different on a darker background...but I'm not sure we would want these 50-level colors to show up as translucent on dark backgrounds anyway.

This isn't fixing anything that hasn't been broken for a while, so I think this can wait until we can discuss the design ramificiations.

With this change:

Screen Shot 2022-10-25 at 11 08 23 AM

View tags throughout Dagit, verify that they look fine on white and gray wash backgrounds.


hellendag:
* master
  * **PR #10186** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10186" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10186?utm_source=stack-comment).

salazarm (approved):

PR #10187, opened 2022-10-25
[dagster-dbt] allow missing run_results.json

OwenKephart:
### Summary & Motivation

Some dbt cloud jobs (ones which do not run "dbt run" or "dbt build" etc.) will not produce a run_results.json artifact. This will cause the step to fail.

This PR will allow run_results.json artifacts to be missing without failing the step.

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/resources.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/resources.py
@@ -455,6 +455,7 @@ class DbtCloudResourceV2:
             run_results = self.get_run_results(run_id)
         # if you fail to get run_results for this job, just leave it empty
         except Failure as _:
+            self._log.info("run_results.json not available for this run")
             run_results = {}
         output = DbtCloudOutput(run_details=final_run_details, result=run_results)
         if output.docs_url:
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/resources.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/resources.py
@@ -451,7 +451,12 @@ class DbtCloudResourceV2:
         final_run_details = self.poll_run(
             run_id, poll_interval=poll_interval, poll_timeout=poll_timeout, href=href
         )
-        output = DbtCloudOutput(run_details=final_run_details, result=self.get_run_results(run_id))
+        try:
+            run_results = self.get_run_results(run_id)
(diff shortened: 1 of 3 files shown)


python_modules/libraries/dagster-dbt/dagster_dbt/cloud/resources.py
```diff
@@ -451,7 +451,12 @@ def run_job_and_poll(
         final_run_details = self.poll_run(
             run_id, poll_interval=poll_interval, poll_timeout=poll_timeout, href=href
         )
-        output = DbtCloudOutput(run_details=final_run_details, result=self.get_run_results(run_id))
+        try:
+            run_results = self.get_run_results(run_id)
+        # if you fail to get run_results for this job, just leave it empty
+        except Failure as _:
```
rexledesma, line 457: Could we use the more specific failure here, and maybe `self._log.info` that no run results exist?
```suggestion
        except DagsterDbtCliOutputsNotFoundError:
```

rexledesma (approved):
Looks good, but we should ensure that we properly log that we are handling failures in this manner.

python_modules/libraries/dagster-dbt/dagster_dbt/cloud/resources.py
OwenKephart: definitely agree on adding logging, but dagster.Failure is actually the most specific error we can get here (without a more involved change). This Failure exception comes from the underlying self.make_request() call, which will emit a Failure event if a request exception if encountered (after retrying a few times). We could pass in a flag through those calls telling make_request to propagate the original exception (rather than dagster.Failure), then just catch the specific RequestException here, but I figured that would be a pretty minor improvement vs. the added complexity.

python_modules/libraries/dagster-dbt/dagster_dbt/cloud/resources.py
```diff
@@ -451,7 +451,13 @@ def run_job_and_poll(
...
+            self._log.info("run_results.json not available for this run")
```
rexledesma, line 458:
```suggestion
            self._log.info("run_results.json not available for this run. Defaulting to empty result.")
```

Could we also fix the description in https://github.com/dagster-io/dagster/blob/2572d664b3c6c5389ca142f21b9b941d898d4d1c/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/types.py#L21-L24? Seems like we don't use `None`, we use `{}`.

PR #10190, opened 2022-10-25
add TimeIntervalPartitionMapping

danielgafni:
### Summary & Motivation

Right now Dagster doesn't have an official implementation of arbitrary PartitionMapping for TimeWIndow-based Partitionsdefinition. This is an attempt to add it.

Adding tests - in progress

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/partition_mapping.py
+++ b/python_modules/dagster/dagster/_core/definitions/partition_mapping.py
@@ -1,10 +1,10 @@
 from abc import ABC, abstractmethod
+from datetime import datetime, timedelta
 from typing import Optional
 
 import dagster._check as check
+from dagster import PartitionKeyRange, PartitionsDefinition, TimeWindowPartitionsDefinition
 from dagster._annotations import experimental, public
-from dagster._core.definitions.partition import PartitionsDefinition
-from dagster._core.definitions.partition_key_range import PartitionKeyRange
 
 
 @experimental
@@ -108,6 +108,53 @@ class LastPartitionMapping(PartitionMapping):
     ) -> PartitionKeyRange:
         last_partition_key = upstream_partitions_def.get_last_partition_key()
         return PartitionKeyRange(last_partition_key, last_partition_key)
+
+    def get_downstream_partitions_for_partition_range(
+        self,
+        upstream_partition_key_range: PartitionKeyRange,
+        downstream_partitions_def: Optional[PartitionsDefinition],
+        upstream_partitions_def: PartitionsDefinition,
+    ) -> PartitionKeyRange:
+        raise NotImplementedError()
+
+
+@experimental
+class TimeIntervalPartitionMapping(PartitionMapping):
+    """
+    Allows mapping arbitrary time intervals.
+    """
+
+    def __init__(self, interval: timedelta, offset: Optional[timedelta] = None):
+        super().__init__()
+
+        offset = offset or timedelta()  # zero timedelta
+
+        self.interval = interval
+        self.offset = offset
+
+    def get_upstream_partitions_for_partition_range(
+        self,
+        downstream_partition_key_range: Optional[PartitionKeyRange],
+        downstream_partitions_def: Optional[PartitionsDefinition],
+        upstream_partitions_def: PartitionsDefinition,
+    ) -> PartitionKeyRange:
+        assert isinstance(downstream_partitions_def, TimeWindowPartitionsDefinition)
+        assert isinstance(upstream_partitions_def, TimeWindowPartitionsDefinition)
+
+        mapped_range = PartitionKeyRange(
+            start=(
+                datetime.strptime(
+                    downstream_partition_key_range.start, downstream_partitions_def.fmt
+                )
+                - self.interval
+                - self.offset
+            ).strftime(upstream_partitions_def.fmt),
+            end=(
+                datetime.strptime(downstream_partition_key_range.end, downstream_partitions_def.fmt)
+                - self.offset
+            ).strftime(upstream_partitions_def.fmt),
+        )
+        return mapped_range
 
     def get_downstream_partitions_for_partition_range(
         self,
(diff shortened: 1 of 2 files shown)


danielgafni:
@sryza could you please take a look and tell me if I'm going in the right direction? the tests are not complete yet

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -116,3 +116,50 @@ def get_downstream_partitions_for_partition_range(
...
+@experimental
+class TimeIntervalPartitionMapping(PartitionMapping):
+    """
+    Allows mapping arbitrary time intervals.
+    """
+
+    def __init__(self, interval: timedelta, offset: Optional[timedelta] = None):
+        super().__init__()
+
+        offset = offset or timedelta()  # zero timedelta
```
sryza, line 130: Nitpick: could inline this with `self.offset = offset`

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -116,3 +116,50 @@ def get_downstream_partitions_for_partition_range(
...
+    def get_upstream_partitions_for_partition_range(
+        self,
+        downstream_partition_key_range: Optional[PartitionKeyRange],
+        downstream_partitions_def: Optional[PartitionsDefinition],
+        upstream_partitions_def: PartitionsDefinition,
+    ) -> PartitionKeyRange:
+        assert isinstance(downstream_partitions_def, TimeWindowPartitionsDefinition)
+        assert isinstance(upstream_partitions_def, TimeWindowPartitionsDefinition)
+
+        mapped_range = PartitionKeyRange(
```
sryza, line 144: If I understand correctly, this could return partition_key values that aren't part of the upstream `PartitionsDefinition`.  A couple cases where this could happen:
- Both upstream and downstream are daily-partitioned, and the time lag is 2 hours.
- Both upstream and downstream are daily-partitioned, starting on "2022-06-06", and we're looking for the upstream partitions for downstream partition "2022-06-06".

sryza (changes requested):
Thanks for posting this. I think a lot of people would benefit from it.

The general concept makes sense to me. Aside from the implementation comment I left inline, I think the biggest question is just what names to use. E.g. "interval" is kind of a synonym of "window", so it's a little bit difficult to tell just by looking at the name how this behaves differently from `TimeWindowPartitionsMapping`.  Writing docstring for these parameters might help us hone in on the right name.

What do you think about, instead of adding a new class, adding some parameters to the existing `TimeWindowPartitionsMapping`?

sryza:
@danielgafni - heads up that I ended up needing to work on this myself, because of an urgent request we needed to satisfy: https://github.com/dagster-io/dagster/pull/10979

danielgafni:
ok cool!

danielgafni:
Hey @sryza, do I understand correctly that the solution you mentioned above is still not general enough?

It can only be used with `TimeWindowPartitionsDefinitions`s with the same partition size?

sryza:
@danielgafni good point. I think that it could be worthwhile to also accept pendulum `Duration`s as values for the `start_offset` and `end_offset` argument, which would allow supporting different partition sizes. I probably won't get around to doing that soon and would be happy to review it if you wanted to.

danielgafni:
I'll reopen the PR then

danielgafni:
Hey @sryza, I started building something like this:


```python

class SlidingWindowPartitionMapping(PartitionMapping):
    def __init__(self, window: timedelta, offset: timedelta):
        self.window = window
        self.offset = offset

    def get_downstream_partitions_for_partition_range(
        self,
        upstream_partition_key_range: PartitionKeyRange,
        downstream_partitions_def: PartitionsDefinition | None,
        upstream_partitions_def: PartitionsDefinition,
    ) -> PartitionKeyRange:
        assert isinstance(downstream_partitions_def, TimeWindowPartitionsDefinition)
        assert isinstance(upstream_partitions_def, TimeWindowPartitionsDefinition)
        assert downstream_partitions_def is not None  # mypy

        upstream_start = datetime.strptime(upstream_partition_key_range.start, upstream_partitions_def.fmt)
        upstream_end = datetime.strptime(upstream_partition_key_range.end, upstream_partitions_def.fmt)

        downstream_window_start = upstream_start + self.window - self.offset

        downstream_window_end = upstream_end - self.offset
        start = downstream_partitions_def.get_last_partition_key(downstream_window_start)
        end = downstream_partitions_def.get_last_partition_key(downstream_window_end)

        return PartitionKeyRange(start, end)

    def get_upstream_partitions_for_partition_range(
        self,
        downstream_partition_key_range: PartitionKeyRange | None,
        downstream_partitions_def: PartitionsDefinition | None,
        upstream_partitions_def: PartitionsDefinition,
    ) -> PartitionKeyRange:
        assert isinstance(downstream_partitions_def, TimeWindowPartitionsDefinition)
        assert isinstance(upstream_partitions_def, TimeWindowPartitionsDefinition)
        assert downstream_partitions_def is not None  # mypy
        assert downstream_partition_key_range is not None  # mypy

        downstream_start = datetime.strptime(downstream_partition_key_range.start, downstream_partitions_def.fmt)
        downstream_end = datetime.strptime(downstream_partition_key_range.end, downstream_partitions_def.fmt)

        downstream_window_start = downstream_start - self.window + self.offset

        downstream_window_end = downstream_end + self.offset

        start = upstream_partitions_def.get_last_partition_key(downstream_window_start)
        end = upstream_partitions_def.get_last_partition_key(downstream_window_end)

        return PartitionKeyRange(start, end)
```

It's not tested yet (and not even used), but do you mind taking a look and confirming if I'm moving in the right direction?

sryza:
@danielgafni I think we should update the existing `TimeWindowPartitionMapping` to accept `Union[timedelta, int]`s for the `start_offset` and `end_offset` arguments.

danielgafni:
ok!

PR #10195, opened 2022-10-25
[dagster-airbyte][managed-elements] Explicit handling of reconciling secret values

benpankow:
## Summary

Right now, Airbyte's APIs return a censored value (`**********`) for secret values such as tokens or passwords. This means that we can't diff a user-provided value against this.

At present, we always show a diff with these secret values, which is a bad experience. This PR introduces:

- A custom comparison param to the `diff_dicts` managed elements utility
- The Airbyte managed element impl returns no diff between secrets if the state has the value `**********`
- The user can manually force the secret value to be updated by passing the ```--include-all-secrets` CLI argument, e.g.

  ```
  dagster-me apply --module my_module --include-all-secrets
  ```

## Test Plan

Unit/integration tests.


benpankow:
* master
  * **PR #10195** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10195" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10195?utm_source=stack-comment).

sryza:
@benpankow do you know how Terraform handles this?

benpankow:
> @benpankow do you know how Terraform handles this?

Terraform is able to handle this marginally better. In cases where they can't read the secret value from the infrastructure (e.g. https://github.com/hashicorp/terraform-provider-aws/issues/22524), the source-of-truth is the persistent Terraform state. If the secret changes outside of Terraform, there's no recourse, but it at least remembers what the last secret supplied is & reacts accordingly if the user config changes.

Our managed elements lacks this state store as a broker between user config and actual infrastructure - there's only one set of reconciliation between (user config <-> infrastructure) rather than (infrastructure <-> state store <-> user config). 

This makes our tool a lot more lightweight (no persistent infra to set up just for managed elements) but this is one of the drawbacks.

sryza (approved):
Looks good!

PR #10197, opened 2022-10-25
[dagster-airbyte] Fix recreating source/destination logic

benpankow:
## Summary

In the case that we have to recreate an Airbyte source or destination, not keeping around the old configuration would cause initializing our internal connection object later to error.

This PR fixes by keeping around that data, and introduces another set of integration tests to cover this case. It also refactors the integration tests a bit, reusing the same Airbyte instance for all tests.

## Test Plan

New tests.


benpankow:
* master
  * **PR #10197** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10197" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10197?utm_source=stack-comment).

OwenKephart (approved):

PR #10198, opened 2022-10-25
add cpu and memory to DagsterEcsTaskDefinitionConfig

gibsondan:
Summary:
This opens the door to being able to configure ecs task memory/cpu for services using DagsterEcsTaskDefinitionConfig.

Test Plan: BK

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws/ecs/tasks.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/ecs/tasks.py
@@ -23,6 +23,8 @@ class DagsterEcsTaskDefinitionConfig(
             ("task_role_arn", Optional[str]),
             ("sidecars", List[Dict[str, Any]]),
             ("requires_compatibilities", List[str]),
+            ("cpu", str),
+            ("memory", str),
         ],
     )
 ):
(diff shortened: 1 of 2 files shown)


python_modules/libraries/dagster-aws/dagster_aws/ecs/tasks.py
```diff
@@ -125,6 +130,8 @@ def from_task_definition_dict(task_definition_dict, container_name):
             task_role_arn=task_definition_dict.get("taskRoleArn"),
             sidecars=sidecars,
             requires_compatibilities=task_definition_dict.get("requiresCompatibilities"),
+            cpu=task_definition_dict.get("cpu"),
```
jmsanders, line 133: This will still be overridable via run tags - that feels like the correct behavior to me, but I wanted to call it out to make sure you agree.

jmsanders (approved):

python_modules/libraries/dagster-aws/dagster_aws/ecs/tasks.py
```diff
@@ -80,9 +86,8 @@ def task_definition_dict(self):
                 ),
                 *self.sidecars,
             ],
-            # https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-cpu-memory-error.html
-            cpu="256",
-            memory="512",
+            cpu=self.cpu,
+            memory=self.memory,
```
Ramshackle-Jamathon, line 90: not sure if we want validation for the cpu/mem combos https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-cpu-memory-error.html worst case it'll just fail upstream in ecs

Ramshackle-Jamathon (approved):

PR #10199, opened 2022-10-26
Fix stray references to define_assets_job

bengotow:
### Summary & Motivation

Was trying to update the hooli_data_eng example and noticed this was incorrect.


sryza (approved):
Oo - thanks for catching these. FYI it looks like a bunch of other formatting changes snuck in here, but seems like they're positive.

bengotow:
Agh crap you're right, looks like my autoformatter ran on them. Guess they do look ok. Thanks!

PR #10202, opened 2022-10-26
Override default io manager in more places

johannkm:
https://github.com/dagster-io/dagster/commit/064a376cde2017448e3e11f2352c0c510c3c9284

Doesn't seem to be applying to asset jobs. It was tested with `define_asset_job` but I guess these use different pathways?

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/job_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/job_definition.py
@@ -20,7 +20,7 @@ from typing import (
 
 import dagster._check as check
 from dagster._annotations import public
-from dagster._config import Field, Shape
+from dagster._config import Field, Shape, StringSource
 from dagster._config.config_type import ConfigType
 from dagster._config.validate import validate_config
 from dagster._core.definitions.composition import MappedInputPlaceholder
@@ -854,7 +854,7 @@ def default_job_io_manager(init_context: "InitResourceContext"):
                 isinstance(attr, IOManagerDefinition),
                 "DAGSTER_DEFAULT_IO_MANAGER_MODULE and DAGSTER_DEFAULT_IO_MANAGER_ATTRIBUTE must specify an IOManagerDefinition",
             )
-            with build_resources({"io_manager": attr}) as resources:
+            with build_resources({"io_manager": attr}, instance=init_context.instance) as resources:
                 return resources.io_manager
         except Exception as e:
             if not silence_failures:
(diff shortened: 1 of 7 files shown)


python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -872,6 +872,45 @@ def default_job_io_manager(init_context: "InitResourceContext"):
...
+            else:
+                warnings.warn(
+                    f"Failed to load io manager override with module: {module_name} attribute: {attribute_name}: {e}\n"
+                    "Falling back to default io manager."
+                )
+    from dagster._core.storage.fs_io_manager import PickledObjectFilesystemIOManager
+
+    # normally, default to the fs_io_manager
+    base_dir = init_context.resource_config.get(
+        "base_dir", init_context.instance.storage_directory()
```
johannkm, line 909: Some of the places where we were directly setting fs_io_manager have tests fail if we don't include this block. So I think it's a breaking change if we don't keep them separate, with this different schema and different base directory

sryza:
Ah yeah, this is what I was referring to here: https://github.com/dagster-io/dagster/pull/9950#issuecomment-1276768316.

I recall there was a particular reason that we used `fs_io_manager` in one place and `default_job_io_manager` in the other, but @dpeng817 might have more context. One thing we want to preserve is that, by default, the `materialize` function materializes assets to the file-system, not to memory.

johannkm (commented):
> One thing we want to preserve is that, by default, the materialize function materializes assets to the file-system, not to memory.

I believe this is preserved, the new function should exactly match the old call to fs_io_manager when the env vars aren't set

johannkm (commented):
test failures are unrelated stuff in internal

sryza:
> I believe this is preserved, the new function should exactly match the old call to fs_io_manager when the env vars aren't set

Can we make sure this is covered by tests?

dpeng817 (approved):
This seems fine and the additional io manager seems necessary to avoid switching to using in-memory when using the in-process executor. Ideally, some of the same improvements you've added to `default_job_io_manager_with_fs_io_manager_schema` could be added to the normal default job IO manager, but that can be a separate PR certainly. 

We originally kept using `default_job_io_manager` for the job case to easily power `materialize_to_memory` and testing IIRC, but I'm realizing that it also introduces a weird behavior disparity, which I think was an oversight, for jobs that are created using `define_asset_job` when assets are provided resource or not. 
- If I call define_asset_job and I used `with_resources` on my assets, the default IO manager is `default_job_io_manager_with_fs_io_manager_schema`.
- If I call define_asset_job and did not use `with_resources` on my assets, the default IO manager is `default_job_io_manager`. 
- Perhaps most pressingly (and maybe write a unit test to show this is the case), I think that if you call with_resources on only some of your assets and not others, and then create a job using all of them, it will simply fail to compile because of different resources for the io manager key.

Solving this is probably a breaking change, but should get something sensible here for 1.1. Fine with approving, since this behavior disparity predates this PR, but should consider ways to fix.

PR #10204, opened 2022-10-26
Mark celery-k8s-test-suite as flaky

jmsanders:
We already install https://github.com/pytest-dev/pytest-rerunfailures but we don't use it. I'm curious if rerunning a single test via PyTest (instead of rerunning the entire test suite via Buildkite) will make this fail builds less frequently until we manage to make it less flaky.

### DIFF
--- a/integration_tests/test_suites/celery-k8s-test-suite/tests/test_daemon_scheduler.py
+++ b/integration_tests/test_suites/celery-k8s-test-suite/tests/test_daemon_scheduler.py
@@ -1,5 +1,6 @@
 import time
 
+import pytest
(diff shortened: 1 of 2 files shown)


johannkm (changes requested):
Won't this guarantee a timeout?

johannkm (approved):
Hmm maybe some of the individual cases are small enough, and the slow fixtures are module scoped (and I assume will be reused). I guess if you want to see how it does go for it

jmsanders:
Yeah I'm not really sure - I consider this our least useful lever to pull right now.

PR #10207, opened 2022-10-26
Run core tests in parallel

jmsanders:
We frequently time out when running our core tests. I have low hopes of this working, but I'm curious if pytest-xdist can successfully parallelize the test suite without any additional work from us.


jmsanders:
Tests passed in 17 minutes on 3.9. That looks to be in the same ballpark that it takes for tests to pass without parallelization (give or take a minute or three). So even if we merge this, I think we should still do #10205.

PR #10208, opened 2022-10-26
Parse HEAD commit in addition to BUILDKITE_MESSAGE

jmsanders:
It looks like when you trigger a build by opening a PR, BUILDKITE_MESSAGE is the PR title and not the body of the latest commit.

Subsequent commits set BUILDKITE_MESSAGE to the commit message.

This means if you want to use any of our substring tooling to configure buildkite (NO_SKIP=true, etc.), you need to either include it in your title of your PR or you need to add a second commit.

Now, we'll parse both BUILDKITE_MESSAGE and check the HEAD commit message.

### DIFF
--- a/.buildkite/dagster-buildkite/dagster_buildkite/git.py
+++ b/.buildkite/dagster-buildkite/dagster_buildkite/git.py
@@ -10,6 +10,14 @@ def get_commit(rev):
     return subprocess.check_output(["git", "rev-parse", "--short", rev]).decode("utf-8").strip()
 
 
+def get_commit_message(rev):
+    return (
+        subprocess.check_output(["git", "rev-list", "--format=%B", "--max-count=1", rev])
+        .decode("utf-8")
+        .strip()
+    )
+
+
 @dataclass
 class GitInfo:
     directory: Path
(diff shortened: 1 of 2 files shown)


alangenfeld (approved):
> It looks like when you trigger a build by opening a PR, BUILDKITE_MESSAGE is the PR title and not the body of the latest commit.


I use the `[DAGSTER_BRANCH=...]` stuff by just amending the commit and submitting it - though always on the first line. Are we sure its the PR title and not just the first line of the commit? 

Scanning the full commit message seems net good still - just want to make sure we understand exactly whats going on.

jmsanders:
Yep - check out `BUILDKITE_MESSAGE` on the first build from this branch - it has only the first line despite being a multi-line commit:

https://buildkite.com/dagster/dagster/builds/37463#018414de-eb0d-4a4d-91b7-f2b448218fab

```
BUILDKITE_MESSAGE="Parse HEAD commit instead of BUILDKITE_MESSAGE"
```

Compared to the second:

https://buildkite.com/dagster/dagster/builds/37469#018414e5-24d5-4dda-b2d9-86eeaad7a34a

```
BUILDKITE_MESSAGE="Fixup list comprehension"
```

Which maybe isn't a great example because the second commit here doesn't have an additional line. But I'm pretty sure it's the distinction between being a pull request trigger vs. a push trigger.

PR #10212, opened 2022-10-26
[dagit] Invert stored state for expand/collapse in Overview pages

hellendag:
### Summary & Motivation

We currently track "expanded" repo keys in the new Overview pages, which means that when you first view the pages, everything is collapsed. This isn't a great experience. It also means that newly added repos are collapsed by default, which I think we don't want.

Instead, invert the tracked state so that we are storing *collapsed* repos. This means that all repos are expanded by default, and the user can clean things up by collapsing repos they don't need to see.

There is a bit of migration code to invert "expanded" state into "collapsed" state in the tracked localStorage values.

Jest test.

Verify expansion/collapse state manually in Dagit, keeping an eye on the localStorage values while doing so.


hellendag:
* master
  * **PR #10212** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10212" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10212?utm_source=stack-comment).

salazarm (approved):

PR #10203, opened 2022-10-26
rm all_types from config_context

alangenfeld:
not used

### How I Tested These Changes

bk

PR #10214, opened 2022-10-26
[config] avoid double init on field cache objects

alangenfeld:
prevents erasing snap memoization and redundant calls to `expand_fields_dict`

bk


alangenfeld:
* master
  * **PR #10203** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10203" target="_blank">Graphite</a> 
    * **PR #10210** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10210" target="_blank">Graphite</a> 
      * **PR #10214** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10214" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10214?utm_source=stack-comment).

python_modules/dagster/dagster_tests/core_tests/resource_tests/test_resource_definition.py
```diff
@@ -1102,21 +1103,21 @@ def configured_resource(num):
...
-            EnumValue("OTHER", TestPythonEnum.OTHER),
-        ],
-    )
+DagsterEnumType = Enum(
+    "ResourceTestEnum",
+    [
+        EnumValue("VALUE_ONE", TestPythonEnum.VALUE_ONE),
+        EnumValue("OTHER", TestPythonEnum.OTHER),
+    ],
+)
```
alangenfeld, line 1117: previously these enums were sharing an object instance but overwriting the inner properties on each `__init__` call

the resulting problem from not overwriting on `__init__` was that the enum `==` checks were failing since the exact enum subclasses are different, being scoped to the different test functions

gibsondan: to confirm, that problem is specific to the test setup right?

gibsondan (approved):

python_modules/dagster/dagster_tests/core_tests/resource_tests/test_resource_definition.py
alangenfeld: pretty much - any user who has duplicate `Enum` config type definitions that actually point at different in memory python enums would have experienced "last init wins" behavior which is busted just as much as "first init wins" so im not too concerned

PR #10215, opened 2022-10-26
[dagster-airbyte][docs] Reorganize Airbyte guide

benpankow:
## Summary

Iterates on the Airbyte getting started guide, by:

- Frontloading configuration/connection info, with a bit more detail on how to point at instances locally vs remotely.

- Using tabs to separate the three connection options

## Test Plan

Tested locally, vercel


benpankow:
* master
  * **PR #10215** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10215" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10215?utm_source=stack-comment).

docs/content/integrations/airbyte.mdx
```diff
@@ -31,14 +31,60 @@ An [Airbyte connection](https://docs.airbyte.com/understanding-airbyte/connectio
...
+- Load asset definitions from Airbyte YAML configuration files, generated by Airbyte's [Octavia CLI](https://github.com/airbytehq/airbyte/tree/master/octavia-cli#what-is-octavia-cli)
+
+You can also manually-build asset definitions on a per-connection basis.
+
+<TabGroup>
 
-You can also [manually-build asset definitions](#manually-building-airbyte-asset-definitions) on a per-connection basis.
+<TabItem name="Loading from Airbyte instance">
+
+### Loading Airbyte asset definitions from an Airbyte instance
```
rexledesma, line 71: These headers are displayed on the table of contents, even when the tab is collapsed, which is kind of awkward. Can we suppress them from showing, or remove them altogether?

rexledesma (commented):
This is wonderful - deferring @erinkcochran87 for comments on the structure of these pages.

One piece of feedback is that I envision the `Integrations > Airbyte` to be more of a landing page, quickly explaining the value of Airbyte, and then linking out to some guides on how to get started.

The content that you've laid out here (through Steps 1 -> 3) is basically a guide to schedule a sync in Airbyte. I think this should be separated out in its own page, and be linked out in the landing page. This is similar to how the `Dagster Cloud` page is structured: https://docs.dagster.io/dagster-cloud.

Another guide I think is valuable is a quick example of setting a downstream dependency of an airbyte asset (e.g. a dbt transformation).

docs/content/integrations/airbyte.mdx
benpankow: > more of a landing page, quickly explaining the value of Airbyte, and then linking out to some guides on how to get started

Yeah, I think this is the way to go eventually once we have more content. For now I bias to keeping it as a single page - I'm a little wary of contentless landing pages until we have a decent amount of sub-pages bc it pollutes docs search and adds clicks to find what a user is looking for.

> a quick example of setting a downstream dependency of an airbyte asset

#10217 adds a more simple example of that - dbt should probably go in its own guide page since it's more involved.

erinkcochran87: I agree with @rexledesma on this one:

> These headers are displayed on the table of contents, even when the tab is collapsed, which is kind of awkward. Can we suppress them from showing, or remove them altogether?

Additionally, the anchors in non-visible tabs don't work correctly if you try to navigate directly to them. I'd expect to be taken right to the heading and for the appropriate tab to be open, but instead I'm taken to the title of the page. Clicking the links in the sidenav also doesn't do anything, which makes things appear broken.

I'd opt for removing the headings from the tabs for now and deciding whether we want to fix this behavior. I do think there's value in it, but I'm not sure it's the best use of time atm.

I'm not sure we need headings in the tabs if the idea is to make the section topic obvious, as the topic is also in the tab name. We could instead opt for styling these in bold or creating a `<span>` tag that stands out, for use in cases like this one.

benpankow: https://github.com/dagster-io/dagster/pull/10231 should fix that behavior (links to headings in tabs will switch to displaying that tab), which should make it easier for us to link to particular tabs etc

not sure if we have an opinion on whether or not those headers should still show in the sidebar - https://github.com/dagster-io/dagster/pull/10231 leaves them

erinkcochran87 (commented):
This is a big improvement - thanks Ben! Overalll I think the structure is 👍 , but I did have some other thoughts. Also left some small comments on the [Vercel preview.](https://dagster-git-benpankow-airbyte-docs-tweaks-elementl.vercel.app/integrations/airbyte#using-airbyte-with-dagster)

- Add a **Prerequisites** section that outlines what users need to do before continuing with the rest of the guide. This includes things like installing libraries (in this case, `dagster-airbyte` and, as obvious as it is, `dagster`)

- I got turned around a lot between [Step 1](https://dagster-git-benpankow-airbyte-docs-tweaks-elementl.vercel.app/integrations/airbyte#step-1-connecting-to-airbyte) and [Step 2](https://dagster-git-benpankow-airbyte-docs-tweaks-elementl.vercel.app/integrations/airbyte#step-2-loading-airbyte-asset-definitions-into-dagster):

   - It took me a minute to realize that Step 2 was telling me to configure the resource I created in Step 1. I think it'd be helpful to specifically call this out - I initially wasn't sure if I had to create an **additional** resource or if I had to do something to what I created in Step 1.
   - Where should the code in Step 2 go? In the same file as the resource from Step 1? Or somewhere else? We've gotten a lot of feedback from users about being confused on where to put things, so I've been pushing to  have us include file names in examples, when appropriate. I think this is especially important in step-by-step guides like this guy.

rexledesma (approved):

erinkcochran87 (approved):
@benpankow this is awesome! thank you so much! :shipit:

PR #9571, opened 2022-09-01
Add docs for customizing the serverless base image

petehunt:

PR #10034, opened 2022-10-14
[docs] - Guides for configuring environment variables

erinkcochran87:
This PR:

- Adds documentation for configuring and using environment variables in Dagster OS and Cloud
- Renames `/dagster-cloud/developing-testing/environment-variables` > `/dagster-cloud/developing-testing/environment-variables-and-secrets`
- Addresses some small oversights in the sidenav and a few Dagster Cloud category pages
- Adds a configuration reference for Docker Hybrid agents. Necessary so the Cloud env vars guide can point somewhere that outlines how to do this.
- Adds env var + secret-specific sections to Hybrid agent configuration references
- Adds a `/dagster-cloud/deployment/kubernetes` page, which was causing a 404

While there are a bunch of small updates across a large number of files, the pages that need detailed review are linked here as previews:

- [Dagster Cloud environment variables and secrets](https://dagster-git-erin-env-vars-secrets-elementl.vercel.app/dagster-cloud/developing-testing/environment-variables-and-secrets)
- [Using environment variables and secrets](https://dagster-git-erin-env-vars-secrets-elementl.vercel.app/guides/dagster/using-environment-variables-and-secrets)

---

**TODOs:**

- [x] Complete OS version of guide
- [x] Add explainers + examples for accessing env vars in Dagster code
- [x] - Update links to Cloud env var (`/dagster-cloud/developing-testing/environment-variables`) to `/dagster-cloud/developing-testing/environment-variables-and-secrets`
- Add reserved variables to Cloud reference
- Update the following:
   - [x] `/dagster-cloud/deployment/serverless` - Add link + update **Secrets** section
   - [x] `/dagster-cloud/deployment/hybrid` - Update is TBD
   - [x] `/dagster-cloud/developing-testing/branch-deployments/using-branch-deployments-with-github` - Step 4.2 needs to be clarified that there's more to use custom env vars, specifically uncommenting some stuff in workflow files
   - [x] `/dagster-cloud/developing-testing/code-locations` - Setup steps discuss agent config - might be good to link out here
   - [x] `/dagster-cloud/deployment/agents` - This page should have a link somewhere, just not sure where
   - [x] `/dagster-cloud/deployment/agents/customizing-configuration` - Add section describing general approach to env vars?
   - [x] `/dagster-cloud/deployment/agents/*` - Add sections to agent config references; create if they don't exist. We can single-source this using the same content in the Hybrid setup guide.

PR #10220, opened 2022-10-27
document serverless pre and post install hooks

slopp:
### Summary & Motivation

~Customer found it difficult to identify how to include `dbt deps`  in a serverless deployment, this PR adds documentation for the pre and post install hooks that are available in the GHA. I also added a bit more detail to the section on adding secrets which is a common question.~

~PR reviewers: please confirm that the pre/post install hooks are not available through the CLI directly?~

Pete updated the docs in #9571 making this PR mostly redundant. There is one minor change remaining that adds more detail on how to edit secrets with the GitHub action.


docs/content/dagster-cloud/deployment/serverless.mdx
```diff
@@ -89,9 +89,36 @@ dagster-cloud serverless deploy \
 
 Often you'll need to securely access secrets from your jobs. With Dagster Cloud Serverless, secrets are added as environment variables that can be read using APIs like [StringSource](/guides/dagster/transitioning-data-pipelines-from-development-to-production#part-two-deployment).
 
-- If you're using our GitHub integration, you can pass in secrets as a JSON string input to the serverless deploy step in your repository's Github action script.
+- If you're using our GitHub integration, you can create secrets in GitHub and then pass those secrets to Dagster by modifying the serverless deploy step in your repository's GitHub action script.
```
gibsondan, line 92: We could link to https://docs.github.com/en/actions/security-guides/encrypted-secrets here

docs/content/dagster-cloud/deployment/serverless.mdx
```diff
@@ -89,9 +89,36 @@ dagster-cloud serverless deploy \
...
+            location: ${{ toJson(matrix.location) }}
+            # Uncomment to pass through Github Action secrets as a JSON string of key-value pairs
+            # env_vars: ${{ toJson(secrets) }}
+```
+
 - If you're using the CLI directly, you can pass `KEY=VALUE` pairs using the `--env` flag. Run `dagster-cloud serverless deploy --help` to learn more.
 
+### Adding Non-Python Dependencies
+
+While Python dependencies will be installed via `setup.py` or `requirements.txt`, occasionally a project will require additional dependencies. For example, dbt dependencies that are installed with `dbt deps`.
```
gibsondan, line 116: Believe we have taken out requirements.txt in the quickstart repos since having both was confusing, i think you can say setup.py here only?

docs/content/dagster-cloud/deployment/serverless.mdx
```diff
@@ -89,9 +89,36 @@ dagster-cloud serverless deploy \
...
+
+- If you're using our GitHub integration, you can create a shell script called `dagster_cloud_pre_install.sh` that will be run before the Python dependencies are installed or you can create a shell script called `dagster_cloud_post_install.sh` that will be run after the Python dependencies are installed. The scripts should be executable, e.g. run `chmod +x dagster_cloud_post_install.sh` prior to committing the file to your Git repo.
+
+- If you're using the CLI directly, you should create a custom base image with your dependencies and include it by specifying `--base-image` in the `deploy serverless` command.
```
gibsondan, line 120: s/`deploy serverless`/`dagster-cloud deploy serverless`

gibsondan (approved):
Thanks for adding this - the specific instructions about how to add secrets are really helpful.

erinkcochran87:
@slopp Just got caught up here - I have two thoughts:

- It looks like this PR has additional info about pre/post hooks that isn't covered in Pete's PR. Do you think the detail you have in here is worth incorporating?
- I'm inclined to drop the update about GitHub secrets, given that #10034 updates this page and will likely be out early next week. @gibsondan has also confirmed that we'll be discouraging this method of env var/secret configuration in the very near future, so we probably shouldn't add it and then have to remove it in a week or two. Thoughts?

slopp:
That sounds good to me Erin, and I don't think the additional hook language is critical, so I'm going to close this one out.

Issue #7384, opened 2022-04-11
[metadata] better access from events


alangenfeld:
> is there a way to get the metadata about an Asset via the EventLogEntry that gets passed to an asset sensor?

Current answer is `event_log_entry.dagster_event.step_materialization_data.materialization.metadata_entries` then dig through the list.

We should have a more convenient way to access metadata values, likely by label name / key.

sryza:
This is now a little bit easier: you can do `event_log_entry.dagster_event.step_materialization_data.materialization.metadata`, which returns a dictionary of metadata labels to metadata values.

PR #10230, opened 2022-10-27
add materialization property on dagster event

prha:
### Summary & Motivation
Add a convenience property to access materialization data off of the event.

Helps with https://github.com/dagster-io/dagster/issues/7384, such that you can now just do:
```event.dagster_event.materialization.metadata```

BK


sryza (approved):
FYI I just started adding attributes even higher up: https://github.com/dagster-io/dagster/pull/10180.  Taken to the limit, it adds a bunch of API surface area, but makes life way easier.

PR #10233, opened 2022-10-27
some refactorings to asset reconciliation sensor

sryza:
### Summary & Motivation

I'm getting started on adding handling for partitions to the asset reconciliation sensor. As I was digesting the code, there were a few refactorings that occurred to me that I thought could make it more performant and more straightforward to extend.

The changes include:
- Use storage_id instead of storage_id + timestamp for the cursor (based on discussion with Claire and Jamie, we don't need both).
- Move instance queries behind a `CachingInstanceQueryer` object. This allows caching results more aggressively and also condensing the code inside `_get_parent_updates`.
- Move all the core reconciliation functionality into a `reconcile` function, which can be tested independently from the rest of the sensor machinery.
- Move some data access code to `EventLogRecord`.

These should not change the behavior at all.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -1,17 +1,13 @@
 # pylint: disable=anomalous-backslash-in-string
 import json
 from collections import defaultdict
-from datetime import datetime
-from typing import TYPE_CHECKING, Dict, Mapping, Optional, Set, Tuple, cast
+from typing import TYPE_CHECKING, Dict, Mapping, Optional, Set, Tuple
 
-import pendulum
 import toposort
 
 import dagster._check as check
 from dagster._annotations import experimental
-from dagster._core.definitions.asset_selection import AssetSelection
 from dagster._core.storage.pipeline_run import IN_PROGRESS_RUN_STATUSES, RunsFilter
-from dagster._utils import utc_datetime_from_timestamp
 
 from .asset_selection import AssetSelection
 from .events import AssetKey
@@ -21,6 +17,8 @@ from .utils import check_valid_name
 
 if TYPE_CHECKING:
     from dagster._core.definitions import AssetsDefinition, SourceAsset
+    from dagster._core.event_api import RunShardedEventsCursor
+    from dagster._core.instance import DagsterInstance
     from dagster._core.storage.event_log.base import EventLogRecord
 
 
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -306,6 +200,192 @@ def sensor_fn(context):
...
+        )
+
+        condition = all if wait_for_all_upstream else any
+        if condition(
+            materialization_status for materialization_status, _ in parent_update_records.values()
+        ):
+            should_materialize.add(current_asset_key)
+
+            # get the cursor value by selecting the max of all the candidates. If we're using a
+            # sharded event log storage, compare timestamps, otherwise compare storage ids. See
```
clairelin135, line 257: comment about sharded event log storage can be removed

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -93,114 +90,56 @@ def _get_parent_updates(
...
-    from dagster._core.events import DagsterEventType
-    from dagster._core.storage.event_log.base import EventRecordsFilter
-
-    parent_asset_event_records: Dict[AssetKey, Tuple[bool, Tuple[float, int]]] = {}
+    parent_asset_event_records: Dict[AssetKey, Tuple[bool, Optional[int]]] = {}
 
     for p in parent_assets:
         if p in will_materialize_set:
             # if p will be materialized by this sensor, then we can also materialize current_asset
             # we don't know what time asset p will be materialized so we set the cursor val to (0.0, 0)
```
clairelin135, line 98: nit: update comment

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -306,6 +200,192 @@ def sensor_fn(context):
...
+
+def _serialize_cursor_dict(cursor_dict: Dict[str, int]) -> str:
+    return json.dumps(cursor_dict)
+
+
+def _deserialize_cursor_dict(cursor: str) -> Dict[str, int]:
+    return json.loads(cursor)
+
+
+class CachingInstanceQueryer:
```
clairelin135, line 282: can we rename this class to emphasize the queryer is instantiated once per tick? I was a little confused by the implementation until I realized this

Maybe `TickInstanceQueryer`?

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
sryza: Thanks for catching - fixed

clairelin135 (approved):
Looking good to me! I added a couple small comments

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -42,26 +40,25 @@ def _get_upstream_mapping(
...
     the parent asset would update the cursor to if it is the most recent materialization of a parent asset.
     In some cases we set the tuple to (0.0, 0) so that the tuples of other parent materializations will take precedent.
 
     Args:
         current_asset: We want to determine if this asset should materialize, so we gather information about
             if its parents have materialized.
         parent_assets: the parents of current_asset.
-        cursor_tuple: In the cursor for the sensor we store the timestamp and storage id of the most recent materialization
-            of current_asset's parents. This allows us to see if any of the parents have been materialized
+        cursor: In the cursor for the sensor, we store the timestamp and storage id of the most recent materialization
```
jamiedemaria, line 60: "we store the timestamp and storage id" isn't true anymore because `cursor` is an int right?

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -222,79 +161,32 @@ def _make_sensor(
     C at 5:16 w/ storage_id 3. When the sensor runs, the cursor for X will be set to (5:16, 3). This way, the next time
```
jamiedemaria, line 161: update comment to reflect that the cursor is an int now

jamiedemaria (approved):
noticed a couple comments that are out of date, but otherwise this looks good

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
sryza: 👍

Issue #10236, opened 2022-10-27
[perf] Poor performance when creating many small jobs


OwenKephart:
### Dagster version

1.0.15

### What's the issue?

Slack: https://dagster.slack.com/archives/C01U954MEER/p1666861854598629?thread_ts=1666020029.135299&cid=C01U954MEER

Take the example dagster repository here:

```
from dagster import repository, asset, define_asset_job

@repository
def experiment():
    elements = []
    for i in range(1,1000):
        @asset(name=f"a{i}")
        def myAsset():
            return 42
        asset_job = define_asset_job(name=f"a{i}_materialize", selection=myAsset.asset_key.to_user_string())
        elements += [myAsset, asset_job]
    return elements
```

This takes a surprising 36 seconds to load on my local machine. I did some profiling, and the vast majority of the time is spent in two functions: `to_source_assets` and `get_all_resource_defs`, which is pretty surprising!

![profile](https://user-images.githubusercontent.com/22457492/198402275-62a1a79d-11eb-4d99-8a79-2c3069a9c4fe.svg)

This is happening because each of these functions requires calling some semi-expensive methods (such as resolve_output_to_origin) for EVERY asset in the repository. Long story short, we can remedy this by not passing every asset in the repository into the build_assets_job function. We are currently doing this because we still need the definitions of some assets which are not selected in order to do things like load inputs (using the io manager defined on the upstream, excluded asset). However, if an asset is not upstream of any asset in the job, there is no reason for it to be present at this point in the call stack, so if we could avoid adding these to the source_assets list this would offer significant performance improvements. In this particular case, this reduces the load time back down to under 2 seconds.

This is a tractable problem, it's fairly straightforward to iterate through the list of included assets and pick out the excluded assets which are actually upstream of them, so we should consider prioritizing this work.

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

zyd14:
related - https://github.com/dagster-io/dagster/issues/22523

PR #10237, opened 2022-10-27
re-use AssetSelectionResolver when constructing multiple jobs

OwenKephart:
### Summary & Motivation

The following repository takes ~36 seconds to load on my laptop:

```
    @repository
    def experiment():
        elements = []
        for i in range(1, 1000):

            @asset(name=f"a{i}")
            def myAsset():
                return 42

            asset_job = define_asset_job(
                name=f"a{i}_materialize", selection=AssetSelection.keys(myAsset.key)
            )
            elements += [myAsset, asset_job]
        return elements
```

Taking this to py-spy, we get the following flame graph:

profile

There are two culprits that are slowing down this repo, the main one being discussed here: https://github.com/dagster-io/dagster/issues/10236, but this change was easier so I did this (it has a much smaller impact). Basically, we stop re-generating the asset dependency graph for every single asset selection

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/unresolved_asset_job_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/unresolved_asset_job_definition.py
@@ -21,6 +21,7 @@ if TYPE_CHECKING:
         PartitionsDefinition,
         SourceAsset,
     )
+    from dagster._core.defintions.asset_selection import AssetSelectionResolver
 
 
 class UnresolvedAssetJobDefinition(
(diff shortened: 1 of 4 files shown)


python_modules/dagster/dagster/_core/definitions/unresolved_asset_job_definition.py
```diff
@@ -138,18 +139,26 @@ def resolve(
         assets: Sequence["AssetsDefinition"],
         source_assets: Sequence["SourceAsset"],
         default_executor_def: Optional["ExecutorDefinition"] = None,
+        asset_selection_resolver: Optional["AssetSelectionResolver"] = None,
```
sryza, line 142: It's a little awkward to me that `asset_selection_resolver` contains kind of the same information as `assets` and `source_assets`.  E.g. it would be easy for them to get out of sync and us not notice it.

If we had the time to do a refactor, I think it might make sense to have some kind of internal `ClosedAssetGraph` class that represents an asset graph.  I.e. it would hold assets, all the source assets those assets depend on, and the . Then, instead of having `assets`, `source_assets`, and `asset_selection_resolver` arguments, it could just take a single argument.

And maybe we could use that inside `AssetLayer` as well?

That said, I realize this could be a big factor and not worth the time. Just wanted to call out that this is a little suboptimal.

python_modules/dagster/dagster/_core/definitions/unresolved_asset_job_definition.py
```diff
@@ -138,18 +139,26 @@ def resolve(
...
     ) -> "JobDefinition":
         """
         Resolve this UnresolvedAssetJobDefinition into a JobDefinition.
         """
+        # when resolving many AssetSelection objects at once (e.g. in a repository), more efficient
+        # to create the AssetSelectionResolver object a single time
+        if asset_selection_resolver is not None:
+            asset_selection = asset_selection_resolver.resolve(self.selection)
```
sryza, line 150: It's confusing that an `AssetSelection` gets resolved to an... `asset_selection`. What do you think about naming the variable `asset_keys`?

OwenKephart:
closing as the AssetGraph class solves this problem

PR #10238, opened 2022-10-27
fix UnresolvedAssetJobDefinition.run_request_for_partition when confi…

sryza:
…g is hardcoded


dpeng817 (approved):
Does a test already exist for the case where you provide a partitioned config directly to the job? What about when that partitioned config does not match the partitions def provided to the asset? If both of these cases are covered, LGTM

sryza:
> Does a test already exist for the case where you provide a partitioned config directly to the job? What about when that partitioned config does not match the partitions def provided to the asset? If both of these cases are covered, LGTM

The latter case didn't have a test. Just added one.

PR #10239, opened 2022-10-27
Highlight config entry when being hovered in yaml editor

salazarm:
### Summary & Motivation

This improves the affordance given to types in the config schema helper. This is to clearly delineate which comments belong to which field since they can be either above/below and the ambiguity leads to misinterpretation.

Screen Shot 2022-10-27 at 7 26 15 PM
Screen Shot 2022-10-27 at 7 26 12 PM
Screen Shot 2022-10-27 at 7 26 09 PM
Screen Shot 2022-10-27 at 7 26 06 PM
Screen Shot 2022-10-27 at 7 26 01 PM

https://www.loom.com/share/e170355e63be4b639331425d6d23b823

### DIFF
--- a/js_modules/dagit/packages/ui/src/components/ConfigTypeSchema.tsx
+++ b/js_modules/dagit/packages/ui/src/components/ConfigTypeSchema.tsx
@@ -223,6 +223,8 @@ const HoveredDictEntryContextProvider = React.memo(({children}: {children: React
       const self = React.useMemo(() => ({setHovered}), []);
       return {
         hovered,
+
+        // Unset the previous hovered target and set the current one
         onMouseEnter: React.useCallback(() => {
           const lastHovered = currentHoveredStack[currentHoveredStack.length - 1];
           if (lastHovered) {
@@ -233,24 +235,30 @@ const HoveredDictEntryContextProvider = React.memo(({children}: {children: React
           currentHoveredStack.push(self);
           setHovered(true);
         }, [self]),
+
+        // Unset the current hovered target and use its parent as the next hovered target if it has one
         onMouseLeave: React.useCallback(() => {
           const lastHovered = currentHoveredStack[currentHoveredStack.length - 1];
           if (!lastHovered) {
-            // This should never happen but if it does we're fine.
+            // This should never happen since we can't MouseLeave something we never MouseEnter'd
             return;
           }
           // We should be the last hovered element, if not lets proceed anyways because
           // we'll hover the correct element later.
           lastHovered.setHovered(false);
+
+          // Find the index of this element and remove it.
+          // There shouldn't be anything after it since MouseLeave events should bubble upwards
           const currentIndex = currentHoveredStack.indexOf(self);
           if (currentIndex !== -1) {
             // This should only remove 1 entry, the last hovered entry
             currentHoveredStack = currentHoveredStack.slice(0, currentIndex);
           }
+
+          // If something is still on the stack after this dict entry is no longer hovered then
+          // its a parent dict entry and should be hovered
           const nextLastHovered = currentHoveredStack[currentHoveredStack.length - 1];
           if (nextLastHovered) {
-            // If something is still on the stack after this dict entry is no longer hovered then
-            // its a parent dict entry and should be hovered
             nextLastHovered.setHovered(true);
           }
         }, [self]),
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/ui/src/components/ConfigTypeSchema.tsx
```diff
@@ -186,13 +186,123 @@ export const ConfigTypeSchema = React.memo((props: ConfigTypeSchemaProps) => {
...
+  const value = React.useMemo(() => {
+    // We need to keep a stack of the entries that are hovered because they are nested.
+    // The `MouseEnter` handler only fires when we first hover the entry, but it does not
+    // fire when exiting a nested dict entry because technically we never left.
+    // To handle that case whenever we `MouseLeave` fires we restore the last element in the
+    // stack before the leaving element as hovered
+
+    let currentHoveredStack: Array<{setHovered: (hovered: boolean) => void}> = [];
+    function useDictEntryHover() {
+      const [hovered, setHovered] = React.useState(false);
```
bengotow, line 222: Hmm I feel like this stack of context -> memo -> fake hook function -> state is quite a lot to grok. (eg: what exactly re-renders when this useState is altered? I think it's the react component invoking this function, but the component must call it on EVERY render or React will throw an exception because of a "conditional" `useState`?) 

It looks structurally sound though and is pretty cool! It's elegant hiding all the mess inside this context provider. I'd worry a bit about having to re-read this to make changes / fix bugs, but it seems unlikely we'll have to do much maintenance of a behavior like this 👍 

It is also tempting to fall back to a non-React solution here though. I think this could be implemented with a few imperative methods like this, where each new mouse enter removes any other hover state, and it's just tweaking class lists after the react render... (since it'd kinda be ok for the hover state to clear when the config editor content changes, anyway.)

```
onMouseEnter={e => {
   e.currentTarget.closest('.container').querySelector('.hovering')?.classList.remove('hovering')
   e.currentTarget.classList.add('hovering')
}
onMouseLeave={e => {
   e.currentTarget.classList.remove('hovering')
}
```

bengotow (approved):

js_modules/dagit/packages/ui/src/components/ConfigTypeSchema.tsx
salazarm: @bengotow I could use that logic where I use `setHovered(true);` but I'll still need the stack because when you exit a nested entry the parent won't get another `MouseEnter` event and then it won't look hovered even though it should.

js_modules/dagit/packages/ui/src/components/ConfigTypeSchema.tsx
```diff
@@ -186,13 +186,131 @@ export const ConfigTypeSchema = React.memo((props: ConfigTypeSchemaProps) => {
...
+          onMouseEnter={onMouseEnter}
+          onMouseLeave={onMouseLeave}
+          ref={ref}
+        />
+      </DictEntryDiv2>
+    );
+  },
+);
+
+const DictEntryDiv2 = styled.div``;
```
salazarm, line 297: I needed this second div because the `$hovered` branching in the css was causing the class name on the div to keep changing which meant using `&` would not reference the non-hovered versions since they have a different css class name.

js_modules/dagit/packages/ui/src/components/ConfigTypeSchema.tsx
salazarm: `context -> memo -> fake hook function -> state`
I agree its a lot but it seems necessary? 

Context -> Share state state between components

useMemo -> So we don't cause any of the components to re-render by creating a new hook every time the hover state changes

"Fake" Hook function -> Defining a hook function within a component/context is probably the easiest way to create a hook with shared state among multiple components. The alternative is to give all of the components a unique ID that they can use to make sure they all subscribe to the same state. That might actually look cleaner since now we wouldn't need to define a hook within a React component/context, however it means that we would need to use a WeakMap and ref objects that die with the component as keys to avoid keeping state around forever.

-> State -> We need the `hovered` state! 

I'm open to other approaches though this one just seemed to encapsulate everything most simply without needing to create an external store.

bengotow: Hey! Ahh I see, yeah my idea didn't address the nesting... that makes sense 👍  I think this works!

Maybe one other sorta-hacky solution -- if you put a mousemove on the sidebar container, I thikn the "event.currentTarget" would always be the container, but "event.target" would be the deepest div in the container that contained the mouse, which might be the one we want to highlight? Maaaybe 😅

Thanks for the explanation!

salazarm: @bengotow Yeah I considered that too! That would work I'd just need to check for the closest ancestor `DictEntry` and set that one as hovered. I can probably do that then not need the stack.

PR #10244, opened 2022-10-28
Compress execute_step args

johannkm:


python_modules/dagster/dagster/_cli/api.py
```diff
@@ -319,7 +320,8 @@ def verify_step(instance, pipeline_run, retry_state, step_keys_to_execute):
 def execute_step_command(input_json):
     with capture_interrupts():
 
-        args = deserialize_as(input_json, ExecuteStepArgs)
+        decompressed_input_json = zlib.decompress(input_json).decode()
```
prha, line 323: do we get compat issues here?  maybe we need to gracefully handle when we get uncompressed args?

johannkm: Got away without it last time and haven't heard any issues https://github.com/dagster-io/dagster/pull/8564

It would only come up with version skew between the step worker and run worker, which isn't something we explicitly support (only could happen in the very rare case where someone is overriding a step image, in which case it should be on the same version anyways bc if the user code is different things will be very odd).

prha (approved):

python_modules/dagster/dagster/_grpc/types.py
```diff
@@ -238,23 +240,28 @@ def __new__(
             ),
         )
 
+    def _get_compressed_args(self) -> str:
+        # Compress, then base64 encode so we can pass it around as a str
+        return base64.b64encode(zlib.compress(serialize_dagster_namedtuple(self).encode())).decode()
```
johannkm, line 245: Re-requesting review bc I had to add base64 encoding to be able to pass the compressed bytes as a string in the CLI command or env var

python_modules/dagster/dagster/_cli/api.py
```diff
@@ -319,7 +321,8 @@ def verify_step(instance, pipeline_run, retry_state, step_keys_to_execute):
 def execute_step_command(input_json):
     with capture_interrupts():
 
-        args = deserialize_as(input_json, ExecuteStepArgs)
+        decompressed_input_json = zlib.decompress(base64.b64decode(input_json.encode())).decode()
```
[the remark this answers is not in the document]
johannkm: Made the decompression opt in

python_modules/dagster/dagster/_cli/api.py
```diff
@@ -316,9 +318,21 @@ def verify_step(instance, pipeline_run, retry_state, step_keys_to_execute):
...
+    envvar="DAGSTER_CLI_API_EXECUTE_STEP_COMPRESSED",
+)
+def execute_step_command(input_json, commpressed):
     with capture_interrupts():
 
+        print("HERE", commpressed, input_json)
+
+        if commpressed:
+            input_json = zlib.decompress(base64.b64decode(input_json.encode())).decode()
+
```
alangenfeld, line 335: I think it would be clearer as two different inputs instead of one that changes based on another arg - adding an `xor` check isn't too bad

python_modules/dagster/dagster/_cli/api.py
```diff
@@ -316,9 +318,21 @@ def verify_step(instance, pipeline_run, retry_state, step_keys_to_execute):
...
+    "commpressed",
+    "--compressed",
+    is_flag=True,
+    default=False,
+    envvar="DAGSTER_CLI_API_EXECUTE_STEP_COMPRESSED",
+)
+def execute_step_command(input_json, commpressed):
     with capture_interrupts():
 
+        print("HERE", commpressed, input_json)
```
alangenfeld, line 331: cleanme

python_modules/dagster/dagster/_cli/api.py
johannkm: done

alangenfeld (approved):
what was the final byte count with compressed b64?

johannkm:
From 230,930 to 13,493 instead of 9,676

PR #10248, opened 2022-10-28
Fetch asset materialization planned event from index shard in Sqlite

clairelin135:
We're currently mirroring asset materialization planned events into the index shard of Sqlite event log storage.

However, `get_event_records` doesn't search for asset materialization planned events in the index shard--this PR amends this behavior. The query in the test failed before this PR and now succeeds.

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/event_log/sqlite/sqlite_event_log.py
+++ b/python_modules/dagster/dagster/_core/storage/event_log/sqlite/sqlite_event_log.py
@@ -262,10 +262,11 @@ class SqliteEventLogStorage(SqlEventLogStorage, ConfigurableClass):
         is_asset_query = event_records_filter and (
             event_records_filter.event_type == DagsterEventType.ASSET_MATERIALIZATION
             or event_records_filter.event_type == DagsterEventType.ASSET_OBSERVATION
+            or event_records_filter.event_type == DagsterEventType.ASSET_MATERIALIZATION_PLANNED
         )
         if is_asset_query:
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/storage/event_log/sqlite/sqlite_event_log.py
```diff
@@ -262,10 +262,11 @@ def get_event_records(
         is_asset_query = event_records_filter and (
             event_records_filter.event_type == DagsterEventType.ASSET_MATERIALIZATION
             or event_records_filter.event_type == DagsterEventType.ASSET_OBSERVATION
+            or event_records_filter.event_type == DagsterEventType.ASSET_MATERIALIZATION_PLANNED
```
prha, line 265: yikes... maybe we should have some sort of constant for the set of asset events?

and then we can just replace these types of things with:
```event_records_filter.event_type in ASSET_EVENT_TYPES```

prha (approved):
I guess this was causing some bug in the default sqlite storage?  How did it manifest?  Were you not getting in-progress run information for some asset graph views?

Might be that we need to also add some graphql test coverage for some of these asset graph queries?

clairelin135:
@prha There weren't any manifestations of the bug, I just noticed it while doing a pass through the code

python_modules/dagster/dagster/_core/storage/event_log/sqlite/sqlite_event_log.py
clairelin135: good call

PR #10251, opened 2022-10-28
[dagit] Updated Asset DAG styling

bengotow:
### Summary & Motivation

Loom video: https://www.loom.com/share/d4b422d4f60147cd8399a80fe4ab5cf2

This PR is based on Sean's excellent asset versioning work and updates the Asset DAG UI to reflect Josh's latest designs here: https://www.figma.com/file/vpKJztqdb5uOLsF2ipzjiK/Asset-Graph?node-id=372%3A1249

Screenshots:

image
image
image


sryza:
Beautiful. I do think it's a little misleading that we show a timestamp next to "Failed" that's not the timestamp that it failed. Maybe we can just omit that?

braunjj:
> Beautiful. I do think it's a little misleading that we show a timestamp next to "Failed" that's not the timestamp that it failed. Maybe we can just omit that?

I agree. My first impression of that timestamp is that it was the date of the failed run, not the last materialization. I think it's bet to just omit it for now. All these details are available if you click on the asset.

PR #10255, opened 2022-10-28
Add a utility method to execute op inside of graph

dpeng817:


dpeng817:
* master
  * **PR #10255** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10255" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10255?utm_source=stack-comment).

python_modules/dagster/dagster/_utils/test/__init__.py
```diff
@@ -198,6 +220,56 @@ def execute_solids_within_pipeline(
     return {sr.solid.name: sr for sr in result.solid_result_list}
 
 
+def execute_op_for_test(
```
sryza, line 223: It's a little strange to call a function like this `for_test`, because nothing about its implementation is test-specific.  Maybe `wrap_op_in_graph_and_execute`?

And a thought: if we changed this to be a function that accepts an op and returns a graph that wraps it, we could maybe expose it in the future as a useful utility.

sryza (changes requested):

python_modules/dagster/dagster/_utils/test/__init__.py
```diff
@@ -198,6 +220,60 @@ def execute_solids_within_pipeline(
     return {sr.solid.name: sr for sr in result.solid_result_list}
 
 
+def wrap_op_in_graph(op_def: OpDefinition, tags: Optional[Mapping[str, Any]] = None):
```
sryza, line 223: Nitpick: return type annotation would be helpful

sryza (approved):
Nice

PR #10258, opened 2022-10-28
disentangle asset reconciliation sensor from MultiAssetSensorDefinition

sryza:
### Summary & Motivation

Asset reconciliation sensors are multi-asset sensors, but they don't seem to use multi-asset sensor features.

When working on improvements to asset reconciliation sensors, I found that this was causing a bug where the multi-asset sensor code would do post-processing on the cursor object produced by the reconciliation sensor, even though it didn't match the shape expected by the multi-asset sensor code.

This makes it so that asset reconciliation sensors are no longer multi-asset sensors. I think it will make sense to make them multi-asset sensors in the future if we harmonize the way they view their cursors.

An effect of this PR is that now all sensors get the repository definition on them.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py
@@ -38,7 +38,6 @@ from .events import AssetKey
 from .run_request import RunRequest, SkipReason
 from .sensor_definition import (
     DefaultSensorStatus,
-    RawSensorEvaluationFunctionReturn,
     SensorDefinition,
     SensorEvaluationContext,
     is_context_provided,
@@ -266,6 +265,7 @@ class MultiAssetSensorEvaluationContext(SensorEvaluationContext):
             cursor=cursor,
             repository_name=repository_name,
             instance=instance,
+            repository_def=repository_def,
         )
 
     def _cache_initial_unconsumed_events(self) -> None:
@@ -343,7 +343,6 @@ class MultiAssetSensorEvaluationContext(SensorEvaluationContext):
         """Updates the cursor after the sensor evaluation function has been called. This method
         should be called at most once per evaluation.
         """
-
         new_cursor = self._cursor_advance_state_mutation.get_cursor_with_advances(
             self, self._unpacked_cursor
         )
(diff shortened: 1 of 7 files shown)


python_modules/dagster/dagster/_grpc/impl.py
```diff
@@ -286,30 +282,16 @@ def get_external_sensor_execution(
...
-        else:
-            sensor_context = stack.enter_context(
-                SensorEvaluationContext(
-                    instance_ref,
-                    last_completion_time=last_completion_timestamp,
-                    last_run_key=last_run_key,
-                    cursor=cursor,
-                    repository_name=repo_def.name,
-                )
+        sensor_context = stack.enter_context(
```
jamiedemaria, line 285: still wrapping me head around some of the impl in this PR, can you explain why we don't need to create the MultiAssetSensorEvalContext anymore?

sryza: It now gets created inside the evaluation function of `MultiAssetSensorDefinition`: https://github.com/dagster-io/dagster/pull/10258/files#diff-91398a61b68ddb85fac7e71bfb7d65188508c60968ea83a2f56c83d6e7791c6cR1120

python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py
```diff
@@ -339,21 +340,21 @@ def _get_partitions_after_cursor(self, asset_key: AssetKey) -> List[str]:
...
+        return json.dumps(
+            self._cursor_advance_state_mutation.get_cursor_with_advances(
+                self, self._unpacked_cursor
+            )
         )
 
-        if new_cursor != None:
-            # Cursor was not updated by this context object, so we do not need to update it
-            self._cursor = new_cursor
-            self._unpacked_cursor = MultiAssetSensorContextCursor(new_cursor, self)
```
clairelin135, line 354: I think we should still keep these lines (354 and 355). When users create a context via `build_multi_asset_sensor_context`, the context should be reset after each sensor invocation so the sensor can be invoked multiple times with the same testing context

sryza: Wrapping my head around this. Is there an example test that does this?

clairelin135: Here's one: https://github.com/dagster-io/dagster/blob/f466eb77e35459165865b781137e7e744b1862f3/python_modules/dagster/dagster_tests/core_tests/definitions_tests/test_sensor_invocation.py#L768

Updating the cursor object at the end of the invocation means that you can fetch the updated state of the cursor after invoking and test against it. I suspect that re-adding this will resolve most of the bk errors

sryza:
My latest update added an additional change that helps get the tests working: `@multi_asset_sensor` now behaves the way that regular `@sensor` behaves for direct invocation: when directly invoked, it returns exactly the same values that the decorated function rather than wrapping them and/or yielding.

sryza:
@clairelin135 @jamiedemaria mind taking another look at this one?

jamiedemaria (approved):
lgtm

PR #10260, opened 2022-10-29
[freshness-policy] Add utility method for calculating upstream data time

OwenKephart:
### Summary & Motivation

This separates out (and reworks) some functionality from a different PR, just to keep things a little cleaner (and multiple future PRs depend on this function, so I wanted to make my life a little easier).

Adds a utility function to recursively calculate the data times for a given asset materialization record. This is done by recursively examining the latest materialization times for each parent asset, filtering for materializations which happened before the given materialization record.

The format I used to express the expected results is pretty gnarly, but I think it makes a fair amount of sense once you wrap your mind around it. Basically, we can't check for explicit timestamps that will be returned by this function, because the time that each pipeline is run changes every time you run the test, and even if you did some time freeze stuff in pytest, you don't know how long each step will take once you start the run, so the materialization timestamps are pretty much unknowable beforehand.

Instead, I keep track of a mapping from AssetKey to an index of what time it was materialized in each run. This way, I can express things like "I expect AssetKey(b)'s root materialization times to be {AssetKey("a"): <timestamp of A from the first of these runs>}". This leads to a pretty terse (if a bit opaque) format, that is independent of the actual timestamps.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -1,5 +1,6 @@
 import warnings
-from typing import AbstractSet, Iterator, Optional, Sequence, Union
+from collections import deque
+from typing import AbstractSet, Iterator, Optional, Sequence, Set, Union
 
 import toposort
 
@@ -62,16 +63,21 @@ class AssetGraph:
 
     def get_roots(self, asset_key: AssetKey) -> AbstractSet[AssetKey]:
         """Returns all root assets that the given asset depends on"""
-        parents = self.get_parents(asset_key)
-        if not parents:
+        if not self.get_parents(asset_key):
             return {asset_key}
-        return set().union(*(self.get_roots(parent_key) for parent_key in parents))
+        return {key for key in self.upstream_key_iterator(asset_key) if not self.get_parents(key)}
 
     def upstream_key_iterator(self, asset_key: AssetKey) -> Iterator[AssetKey]:
         """Iterates through all asset keys which are upstream of the given key."""
-        for parent_key in self.get_parents(asset_key):
-            yield parent_key
-            yield from self.upstream_key_iterator(parent_key)
+        visited: Set[AssetKey] = set()
+        queue = deque([asset_key])
+        while queue:
+            current_key = queue.popleft()
+            for parent_key in self.get_parents(current_key):
+                if parent_key not in visited:
+                    yield parent_key
+                    queue.append(parent_key)
+                    visited.add(parent_key)
 
     def get_children_partitions(
         self, asset_key: AssetKey, partition_key: Optional[str] = None
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -1,5 +1,5 @@
 import warnings
-from typing import AbstractSet, Optional, Sequence, Union
+from typing import AbstractSet, Iterator, Optional, Sequence, Union
 
 import toposort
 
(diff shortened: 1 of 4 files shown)


clairelin135 (commented):
This is all looking good to me! Only thing is that the `get_event_tags_for_asset` method currently doesn't accept a storage ID: https://github.com/dagster-io/dagster/blob/a34ffc5c38ce87f2761c1464ebb9f0b6e7b814ff/python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py#L1218

You'll need to update this to accept event IDs to filter by and add these event ids to the `.where` clause

python_modules/dagster/dagster/_utils/calculate_data_time.py
```diff
@@ -0,0 +1,107 @@
...
+                    tup = tuple(tup)
+                    # if root data is missing, override other values
+                    if tup == (None, None) or known_times.get(key) == (None, None):
+                        known_times[key] = (None, None)
+                    else:
+                        known_times[key] = min(tuple(known_times.get(key, tup)), tup)
+
+        # cache asset event in the database
+        instance.add_asset_event_tags(
+            record.storage_id, {UPSTREAM_MATERIALIZATION_TIMES_TAG: known_times}
```
sryza, line 92: Have you considered having a tag per upstream asset? That would allow us to do queries like "what materializations are downstream of this materialization".  I know we wouldn't be able to do that immediately, because it won't always be cached, but I suspect there's a road to it.

python_modules/dagster/dagster/_utils/calculate_data_time.py
```diff
@@ -0,0 +1,107 @@
...
+from typing import AbstractSet, Mapping, Optional, cast
+
+from dagster import AssetKey, DagsterEventType, DagsterInstance, EventLogRecord, EventRecordsFilter
+
+UPSTREAM_MATERIALIZATION_TIMES_TAG = ".dagster/upstream_times"
+
+
+def get_upstream_materialization_times_for_record(
+    instance: DagsterInstance,
+    upstream_asset_key_mapping: Mapping[str, AbstractSet[str]],
```
sryza, line 11: With the changes in [this PR](https://github.com/dagster-io/dagster/pull/10402), can we operate on asset keys instead of strings?

sryza (commented):
Some of this will change after @smackesey's work gets merged, right?

python_modules/dagster/dagster/_utils/calculate_data_time.py
```diff
@@ -0,0 +1,126 @@
...
+
+    The heart of this functionality is a recursive method which takes a given asset materialization
+    and finds the most recent materialization of each of its parents which happened *before* that
+    given materialization event.
+    """
+
+    def _storage_key(storage_id: int) -> str:
+        return f".dagster-used_upstream_times-{storage_id}"
+
+    _local_cache: Dict[int, Dict[AssetKey, Tuple[Optional[int], Optional[float]]]] = {}
```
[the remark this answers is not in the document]
OwenKephart: mm yeah I can do that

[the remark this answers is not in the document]
OwenKephart: @sryza ok refactored this entire thing into a class. the benefit is that now we can keep around an instance of this when we plan on getting the data times of multiple assets in a row (which is very common). Previously, we ended up caching all "intermediate data" on the instance, which is no longer necessary as we can cache certain values in memory instead.

specifically if you had A -> B -> C -> D and only D had a freshness policy, we would store info on the instance for ALL of the assets. Now, we'll only cache the info for D on the instance and keep everything else in-memory.

python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -60,6 +60,19 @@ def get_parents(self, asset_key: AssetKey) -> AbstractSet[AssetKey]:
...
         return self.asset_dep_graph["upstream"][asset_key]
 
+    def get_roots(self, asset_key: AssetKey) -> AbstractSet[AssetKey]:
+        """Returns all root assets that the given asset depends on"""
+        parents = self.get_parents(asset_key)
+        if not parents:
+            return {asset_key}
+        return set().union(*(self.get_roots(parent_key) for parent_key in parents))
+
+    def upstream_key_iterator(self, asset_key: AssetKey) -> Iterator[AssetKey]:
```
sryza, line 70: See comment here: https://github.com/dagster-io/dagster/pull/10366/files#r1018266953

sryza (approved):
I think we should fix `get_roots` to only hit each node once, like I mentioned in a comment. Otherwise, ship it!

python_modules/dagster/dagster/_core/definitions/asset_graph.py
OwenKephart: 🤦 can't believe I didn't think of BFS for this

PR #10272, opened 2022-10-31
[dagster-airbyte] Generate src/dst Python classes for managed stacks

benpankow:

### DIFF
--- a/scripts/gen_airbyte_classes.py
+++ b/scripts/gen_airbyte_classes.py
@@ -1,0 +1,508 @@
+import importlib.util
+import json
+import os
+import re
+import subprocess
+import sys
+from abc import ABC, abstractmethod
+from contextlib import contextmanager
+from typing import Any, Dict, List, Optional, Sequence, Union
+
+import click
+import yaml
+
+import dagster._check as check
+from dagster._utils import file_relative_path, mkdir_p
+
+
+def _remove_invalid_chars(name: str) -> str:
+    return "".join([(x if x.isalnum() else "_") for x in name])
+
+
+# Any keywords we should avoid using as class names
+KEYWORDS = ["None", "Any", "Union", "List", "Dict", "Optional", "T"]
+
+
+def _to_class_name(name: str) -> str:
+    class_name = "".join([_remove_invalid_chars(x).capitalize() for x in re.split("\W+", name)])
+    if class_name in KEYWORDS:
+        class_name += "_"
+    return class_name
+
+
+class SchemaType(ABC):
+    @abstractmethod
+    def get_check(self, name: str, scope: Optional[str] = None):
+        pass
+
+    @abstractmethod
+    def annotation(self, scope: Optional[str] = None, quote: bool = False):
+        pass
+
+    @property
+    def const_value(self):
+        return None
+
+
+TYPE_MAPPING = {
+    "string": "str",
+    "integer": "int",
+    "number": "float",
+    "boolean": "bool",
+}
+
+
+CHECK_MAPPING = {
+    "str": "check.str_param({}, '{}')",
+    "int": "check.int_param({}, '{}')",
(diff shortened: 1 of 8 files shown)


benpankow:
* master
  * **PR #10272** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10272" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10272?utm_source=stack-comment).

scripts/gen_airbyte_classes.py
```diff
@@ -0,0 +1,579 @@
...
+    connector_name_human_readable: str,
+    cls_name: str,
+    cls_def: Dict[str, SchemaType],
+    nested: Optional[List[str]],
+    is_source: bool,
+    docs_url: str,
+):
+    nested_defs = ""
+    if nested:
+        nested_defs = "\n".join(
```
OwenKephart, line 322: https://docs.python.org/3/library/textwrap.html#textwrap.indent might be helpful here

scripts/gen_airbyte_classes.py
```diff
@@ -0,0 +1,579 @@
...
+        connector_name_human_readable,
+        connector_name,
+        cls_defs[connector_name],
+        defs,
+        is_source,
+        schema["documentationUrl"],
+    )
+
+
+SOURCE_OUT_FILE = "/Users/ben/Documents/repos/dagster/python_modules/libraries/dagster-airbyte/dagster_airbyte/managed/generated/sources.py"
```
OwenKephart, line 433: maybe a file_relative_path here :P

scripts/gen_airbyte_classes.py
```diff
@@ -0,0 +1,579 @@
...
+        cls_name=cls_name,
+        fields_in=fields_in,
+        self_fields=self_fields,
+        nested_defs=nested_defs,
+        human_readable_name=connector_name_human_readable,
+        docs_url=docs_url,
+    )
+
+
+class GithubSource:
```
OwenKephart, line 359: is this supposed to be in this file?

scripts/gen_airbyte_classes.py
```diff
@@ -0,0 +1,579 @@
...
+    def __init__(self, inner: SchemaType):
+        self.inner = inner
+
+    def __str__(self):
+        return f"Optional[{self.inner}]"
+
+    def annotation(self, scope: Optional[str] = None, quote: bool = False):
+        return f"Optional[{self.inner.annotation(scope, quote)}] = None"
+
+    def get_check(self, name: str, scope: Optional[str] = None):
```
OwenKephart, line 122: So my one main point of concern here (which is possibly fine?) is that `check.opt_list_param` will return an empty list when given `None` as an input, and I'm not sure we can be confident that all Airbyte sources / destinations will treat empty lists identically to `None` (I'd be surprised if they do)

Instead, we should use `check.opt_nullable_list_param` in those cases (mildly annoying because there's no analogous (e.g.) `check.opt_nullable_int_param`, so this logic will have to inspect the type of inner).

OwenKephart (changes requested):
This looks great! I had some small comments, but the code seems reasonable and a scan through the generated files didn't turn up anything too suspicious, other than the one bit about the check.opt_list_param call.

Happy to review as soon as that gets updated (or if it turns out not to be an issue)

scripts/gen_airbyte_classes.py
benpankow: no haha, let me remove

scripts/gen_airbyte_classes.py
benpankow: Good call, added some logic to inspect the inner type and return a nullable check instead.

OwenKephart (approved):
great!

PR #10279, opened 2022-10-31
run_config argument on run_request_for_partition

sryza:
### Summary & Motivation

This was recently requested by a user.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/job_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/job_definition.py
@@ -561,6 +561,7 @@ class JobDefinition(PipelineDefinition):
         run_key: Optional[str] = None,
         tags: Optional[Mapping[str, str]] = None,
         asset_selection: Optional[Sequence[AssetKey]] = None,
+        run_config: Optional[Mapping[str, Any]] = None,
     ) -> RunRequest:
         """
         Creates a RunRequest object for a run that processes the given partition.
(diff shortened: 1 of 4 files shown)


python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -573,6 +574,9 @@ def run_request_for_partition(
                 value means that a run will always be launched per evaluation.
             tags (Optional[Dict[str, str]]): A dictionary of tags (string key-value pairs) to attach
                 to the launched run.
+            run_config (Optional[Mapping[str, Any]]: Configuration for the run. If the job has
```
prha, line 577: This feels like a foot gun...  they want to do this rather than just create their own RunRequest and add the partition tags?

Granted, the current way of getting the partition tags is awkward, but if we made some small changes, we could enable something like:

```
yield RunRequest(run_key=None, run_config=my_custom_run_config, tags=job.tags_for_partition(my_partition_key)
```

sryza: When you say it's a footgun, are there situations you have in mind?

The way I've been looking at it is that tags are an implementation detail for assigning a partition to a run. I.e. most users who use partitions likely aren't aware that tags are the mechanism that we use for assigning partitions to runs.  In Dagit, the Launchpad allows you to launch a partitioned run and supply the configuration, without thinking about tags. As a user, I would expect to be able to perform that same operation using code.

prha: I was originally thinking that there'd be confusion about what config was used for what when trying to produce a partition run.  But the syntax I proposed doesn't really change that confusion.  Also, I suppose I've been stuck in the mode of equating a job's partition space to its run config, which we might be trying to move away from anyway.

Feel free to ignore.

prha (approved):

PR #10287, opened 2022-11-01
Display notebook metadata type in dagit

jamiedemaria:
### Summary & Motivation
There is a new NotebookMetadata type that can be used in dagster. This displays it in dagit as a "View Notebook" button. Some of the ways metadata is rendered (like in dagit logs) i couldn't find a way to get access to the repository location so i put in a condition that just renders the name of the file if there's no repo location.

Screen Shot 2022-11-01 at 9 39 51 AM

First "View Notebook" (red) is from metadata returned by the asset. Second "View Notebook" (blue) is from metadata on the asset decorator


js_modules/dagit/packages/core/src/metadata/MetadataEntry.tsx
```diff
@@ -171,6 +182,51 @@ export const MetadataEntry: React.FC<{
       return null;
     case 'TableSchemaMetadataEntry':
       return <TableSchema schema={entry.schema} />;
+    case 'NotebookMetadataEntry':
+      if (repoLocation) {
```
jamiedemaria, line 186: One thing i'd like to do is change this so that it supports URLs like in the ipynb plugin ([here](https://github.com/dagster-io/dagster/blob/master/js_modules/dagit/packages/core/src/plugins/ipynb.tsx#L22)) but i'm not sure how to do the conditional check (with `React.useMemo`). if i put it in this condition i get a linting error that `useMemo` can't be called conditionally. But putting it at the top of the function also seems wrong since we don't know what attribute of `entry` to access or what type it will be

python_modules/dagster-graphql/dagster_graphql/implementation/events.py
```diff
@@ -365,10 +373,8 @@ def from_dagster_event_record(event_record: EventLogEntry, pipeline_name: str) -
         )
     elif dagster_event.event_type == DagsterEventType.LOGS_CAPTURED:
         return GrapheneLogsCapturedEvent(
-            fileKey=dagster_event.logs_captured_data.file_key,
-            logKey=dagster_event.logs_captured_data.file_key,
```
bengotow, line 369: Hmm these changes look a bit unrelated?

js_modules/dagit/packages/core/src/metadata/MetadataEntry.tsx
bengotow: Hey @jamiedemaria! Ahh ok - what we should probably do here is take the content of `SidebarComponent.tsx` that handles the URL + rendering of the notebook and move it to a separate component. It looks like that SIdebarComponent does a /bit/ of unwrapping, etc. in the first few lines, but then it's mostly the same as what we want here?

You can only use React hooks like useEffect and useMemo before the first if/return statement of your component, so it'll complain about one being inside a switch here. Once we have that component that is shared between SidebarComponent and this MetadataEntry, we can say `return <NotebookButton repoLocation={repoLocation} path={entry.path} />` in both places and we should be set!

jamiedemaria: cool! that sounds good

bengotow (approved):
Looks awesome! 🙌

python_modules/dagster-graphql/dagster_graphql/implementation/events.py
jamiedemaria: yup - bad git merge. it's resolved now

PR #10313, opened 2022-11-02
Fix Helm package skipping

jmsanders:
There were two issues here:

1. The package spec name didn't match the package name so we weren't finding any dependency tree to walk.
2. Our extra steps bore no relation to our package spec steps. I think it's probably more accurate to run them any time the package spec steps run as well.

This should resolve the Helm steps not running on https://buildkite.com/dagster/dagster/builds/38057

### DIFF
--- a/.buildkite/dagster-buildkite/dagster_buildkite/steps/helm.py
+++ b/.buildkite/dagster-buildkite/dagster_buildkite/steps/helm.py
@@ -46,7 +46,7 @@ def _build_lint_steps(package_spec) -> List[CommandStep]:
             "pip install yamllint",
             "make yamllint",
         )
-        .with_skip(skip_if_no_helm_changes() or package_spec.skip_reason)
+        .with_skip(skip_if_no_helm_changes() and package_spec.skip_reason)
         .on_test_image(AvailablePythonVersion.get_default())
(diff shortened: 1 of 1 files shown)


rexledesma (approved):

.buildkite/dagster-buildkite/dagster_buildkite/steps/helm.py
```diff
@@ -38,14 +39,14 @@ def build_helm_steps() -> List[BuildkiteStep]:
...
-def _build_lint_steps() -> List[CommandStep]:
+def _build_lint_steps(package_spec) -> List[CommandStep]:
     return [
         CommandStepBuilder(":yaml: :lint-roller:")
         .run(
             "pip install yamllint",
             "make yamllint",
         )
-        .with_skip(skip_if_no_helm_changes())
+        .with_skip(skip_if_no_helm_changes() or package_spec.skip_reason)
```
prha, line 49: I think we want to run this if there are no helm changes and we're skipping the package_spec, right?

jmsanders: Oh yes you're absolutely right.

PR #10324, opened 2022-11-02
Read `account_id` from an environment variable in `dbt_cloud_resource`

sowusu-ba:
### Summary & Motivation
Whereas it is possible to read the `auth_token` field from an environment variable in  `dbt_cloud_resource`, it is not possible to do the same thing with the `account_id` field, which is configured as an `int`.  Doing something like 

```
dbt_cloud_resource.configured(
     {'account_id': {'env': 'ACCOUNT_ID'}
)
```
will not work. This PR adds the functionality

After making this change, I read the `account_id` from an environment variable
Screen Shot 2022-11-02 at 4 48 23 PM

then, I added a test for this to be an integer when read
Screen Shot 2022-11-02 at 4 41 55 PM

pytest results all passed
Screen Shot 2022-11-02 at 4 48 46 PM

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/resources.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/resources.py
@@ -8,7 +8,7 @@ from urllib.parse import urlencode, urljoin
 import requests
 from requests.exceptions import RequestException
 
-from dagster import Failure, Field, MetadataValue, StringSource, __version__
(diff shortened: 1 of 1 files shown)


alangenfeld:
awesome, thanks for the PR. Looks right to me, will merge after buildkite check.

alangenfeld:
one remaining formatting issue - can you run `make isort` (and `make black`  to be safe)

PR #10326, opened 2022-11-02
Bump dagster grpcio dependency to natively support M1 mac

pragbatra-mendel:


gibsondan (commented):
Native support for M1s sounds fantastic - I think we will want to test this very carefully though before landing, as we have had a lot of problems with the last couple of grpcio releases causing instability (crashes, hangs, etc.)

gibsondan:
Just looking through the release notes and I don’t see any mentions of M1s
https://github.com/grpc/grpc/releases/tag/v1.50.0 - is there more you can
share about how this release will help there?

gibsondan:
From slack it sounds like upgrading pip was able to fix this for you - let us know if there's any reason to believe that 1.50.0 is worth pinning.

joshuataylor:
1.50 doesn't have a version for wheels with arm64 (not sure why?), but it supports Python 3.11 wheels.

PR #10336, opened 2022-11-03
fix mysql heading level + typo

domsj:
### Summary & Motivation

noticed this while reading https://docs.dagster.io/deployment/dagster-instance#dagster-storage

not


smackesey (approved):
Thanks for the fix!

PR #10338, opened 2022-11-03
Better fix for tooltips around disabled buttons + gitlab icon

salazarm:
### Summary & Motivation

When a button is disabled I'm seeing that the `MouseLeave` event is fired inconsistently on the div wrapping the `button` causing the tooltip to still show in some cases. The same inconsistency was not seen with `MouseEnter`, so this approach keeps `MouseEnter` on the wrapper div, but adds a body listener when the tooltip is open that listens for `MouseMove` and checks if the target is within the wrapper div. If not then we close the tooltip.

Cloud


js_modules/dagit/packages/ui/src/components/Tooltip.tsx
```diff
@@ -36,6 +36,23 @@ export const Tooltip: React.FC<Props> = (props) => {
 
   const [isOpen, setIsOpen] = React.useState<undefined | boolean>(undefined);
 
+  const divRef = React.useRef<any>(null);
```
hellendag, line 39: Could be an `HTMLDivElement` instead of `any`, right?

js_modules/dagit/packages/ui/src/components/Tooltip.tsx
```diff
@@ -36,6 +36,23 @@ export const Tooltip: React.FC<Props> = (props) => {
...
+
+  React.useLayoutEffect(() => {
+    let listener: null | ((e: MouseEvent) => void) = null;
+    if (isOpen && useDisabledButtonTooltipFix) {
+      listener = (e: MouseEvent) => {
+        if (!divRef.current?.contains(e.target as HTMLDivElement)) {
+          setIsOpen(false);
+        }
```
hellendag, line 47: Look upward from `e.target` instead of downward from `divRef`? Then you're just looking up the ancestry list instead of through the entire tree below `divRef`? Maybe nbd since tooltip targets are generally small.

Or maybe `compareDocumentPosition`?

hellendag (approved):

js_modules/dagit/packages/ui/src/components/Tooltip.tsx
salazarm: Yeah I did consider that but the tree isn't that big so it seems like early optimization. I'm using `contains` though which theoretically probably works that way under the hood?

PR #10341, opened 2022-11-03
Back-compat fix for setting the task definition in the EcsRunLauncher to an env var

gibsondan:
Summary:
We recently moved task_definition from a StringSource to a scalar union, but some users were relying on setting the task definition to an environment variable. In hindsight we probably could have made it an entirely separate field instead of a scalar union, but for now to keep the current behavior without breaking old behavior, set an 'env' key to the dictionaryu that allows it to still be used to source the string from a environment variable.

Test Plan: BK

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
@@ -1,4 +1,5 @@
 import json
+import os
 import warnings
 from collections import namedtuple
 from typing import Any, Dict, Optional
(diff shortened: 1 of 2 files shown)


python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -65,10 +66,24 @@ def __init__(
...
-            task_definition if not isinstance(task_definition, str) else None
-        )
+        self.task_definition = None
+        self.task_definition_dict = None
+        if isinstance(task_definition, str):
+            self.task_definition = task_definition
+        elif task_definition and "env" in task_definition:
+            check.invariant(
+                len(task_definition) == 1,
+                "If `task_definition` is set to a dictionary with `env`, `env` must be the only key.",
```
jmsanders, line 76: Can you add a unit test for this too?

jmsanders (approved):

gibsondan:
@jmsanders i believe the test failures here are b/c something is not triggering the needed docker image to be built for the test suite

PR #10349, opened 2022-11-03
[dagit] Fix bug in asset live reloading after step failures, clean up AssetView

bengotow:
### Summary & Motivation

This fixes a subtle bug in the live reloading behavior of the asset graph that made it unable to reload after step failure events. (The spread operator was double-unwrapping an array of strings into an array of letters, which I didn't realize was possible!)

I also changed AssetView to pull the "last materialization" from this live data, and use it to trigger a reload of the events / partitions data after a run completes. AssetView was getting a bit complicated, so I pulled a few things into helper functions as well.


js_modules/dagit/packages/core/src/assets/AssetView.tsx
```diff
@@ -201,94 +156,98 @@ export const AssetView: React.FC<Props> = ({assetKey}) => {
...
+              assetLastMaterializedAt={lastMaterializedAt}
+              assetHasDefinedPartitions={!!definition?.partitionDefinition}
+              params={params}
+              paramsTimeWindowOnly={!!params.asOf}
+              setParams={setParams}
+              liveData={definition ? liveDataByNode[toGraphId(definition.assetKey)] : undefined}
+            />
+          )}
+        </>
+      )}
```
hellendag, line 250: Just one opinion here, but I find these nested ternaries pretty hard to follow.

hellendag (approved):

js_modules/dagit/packages/core/src/assets/AssetView.tsx
bengotow: Yeah... let me see if I can fix this before this merges 👍   I wish this used React Router sub-pages instead of a query string for each tab but that might make it difficult to feature flag...

PR #10355, opened 2022-11-03
jupyter logo

jamiedemaria:


js_modules/dagit/packages/core/src/graph/OpTags.tsx
```diff
@@ -39,7 +38,13 @@ function hueForTag(text = '') {
 
 function getTag(tag: IOpTag) {
   if (tag.label === 'noteable') {
-    return <Icon name="noteable_logo" />;
+    return <img src={noteable_logo} />;
```
bengotow, line 41: Maybe before we merge, want to add `alt="Noteable logo" role="img"` to each of these? Tthe linter isn't complaining but I thought that it would...

bengotow (approved):
This looks great! Excited to get these bigger logo tags in 🙌

braunjj:
Thanks for doing this @jamiedemaria! As discussed in Slack, Let's update these tags to look a bit more uniform across all integrations. Each one would get a solid bg color and a single color logo mark + the name written in our default font so it's more legible at such a small scale. This pattern should make it easy to add more of these in the future as well.

image

### **Colors**
**dbt**: bg = `#FF6B4C`, text = `#ffffff`
**Noteable**: bg = `#00D2D2`, text = `#171615`
**Airbyte**: bg = `#655CFC`, text = `#ffffff`
**jupyter**: bg = `#4E4E4E`, text = `#ffffff`
**Python**: bg = `#35668F`, text = `#ffffff`
**Snowflake**: bg = `#29B5E8`, text = `#ffffff`
**Fivetran**: bg = `#0073FF`, text = `#ffffff`

### **Custom Tags**
For all the other custom tags, can we update their style to match the other branded tags?
The bg and font color can be any color but the tag size should be 24px high with 8px border radius and the font should be `Inter semi-bold, 12px`
image

The Figma file for all these designs is available [here](https://www.figma.com/file/vpKJztqdb5uOLsF2ipzjiK/Asset-Graph?node-id=339%3A1227).

braunjj (approved):
Love it!

js_modules/dagit/packages/ui/src/components/Icon.tsx
```diff
@@ -330,15 +326,16 @@ export const IconWrapper = styled.div<WrapperProps>`
         // We could just use !important but specificity is a little more flexible
         `
         background: url(${p.$img});
+        background-size: cover;
```
salazarm, line 329: is this needed?

js_modules/dagit/packages/ui/src/components/Icon.tsx
```diff
@@ -330,15 +326,16 @@ export const IconWrapper = styled.div<WrapperProps>`
...
         &[role='img'][role='img'] {
           background-color: transparent;
         }
       `
       : `
         background: ${p.$color};
+        mask-size: cover;
         mask-image: url(${p.$img});
       `}
-  mask-size: cover;
```
salazarm, line 341: is this needed? You're not using <Icon/> anymore so I think not?

jmsanders:
This is so cool 🥳

js_modules/dagit/packages/ui/src/components/Icon.tsx
bengotow: I think when you specify a "fill color" we're using a solid color + mask image, so the mask-size is just necessary in that case. Jamie and I moved these rules up into the ternary above!

PR #10366, opened 2022-11-04
[freshness-policies] Add FreshnessPolicyInfo graphql endpoint

OwenKephart:
### Summary & Motivation

We want to be able to access info about how out of date an asset is with respect to its freshness policy in Dagit. We use a utility function to calculate this value. There are two values that we're interested in in particular (outside of a description of the policy). These are the number of minutes late a given asset currently is, and the number of minutes late that the most recent materialization of the asset was.

We currently can't take advantage of the fancy latestMaterialization batch loader thing, as we need the EventLogRecord (which has a storage id), not just the EventLogEntry. In the short term, this isn't a huge deal as we'll only need to make this query for each asset that has a freshness policy defined (rather than every single asset), but in the future, this should be optimized.

We need to be able to serialize the FreshnessPolicy objects because this will be evaluated in a process that does not have access to the source code, so we need to create these ExternalFreshnessPolicy things (we do a similar thing with PartitionSets). 

Hacked in a little bit of typescript and was able to add this info to the asset node live query / asset node in dagit. Worked as expected


python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
```diff
@@ -80,6 +81,15 @@ def resolve_asset(self, _graphene_info):
         )
 
 
+class GrapheneAssetFreshnessInfo(graphene.ObjectType):
+    currentMinutesLate = graphene.Field(graphene.Float)
```
clairelin135, line 85: I think these policyDescription field should be required, since it would be populated anytime the `FreshnessInfo` object is returned

so `graphene.Field(graphene.String)` -> `graphene.String`

python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
```diff
@@ -529,6 +540,74 @@ def resolve_type(self, _graphene_info) -> Optional[str]:
                     )
         return None
 
+    def resolve_freshnessInfo(self, graphene_info) -> Optional[GrapheneAssetFreshnessInfo]:
+        from dagster import DagsterEventType, EventRecordsFilter
+        from dagster._utils.calculate_data_time import get_upstream_materialization_times_for_record
+
+        # no FreshnessPolicy, so freshnessInfo does not apply
```
clairelin135, line 547: maybe we can break this out into a helper method in a separate file that returns `GrapheneAssetFreshnessInfo` objects to keep the `GrapheneAssetNode` object short?

clairelin135 (commented):
This is looking good to me. Can we also add a test for this? (The other asset graphql tests are located in this file: https://github.com/dagster-io/dagster/blob/master/python_modules/dagster-graphql/dagster_graphql_tests/graphql/test_assets.py)

python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
OwenKephart: good call (looks like I need it to be `graphene.NonNull(graphene.String)`)

python_modules/dagster/dagster/_core/host_representation/external_data.py
```diff
@@ -509,6 +515,99 @@ def __new__(cls, error: Optional[SerializableErrorInfo]):
         )
 
 
+class ExternalFreshnessPolicyData(ABC):
```
OwenKephart, line 518: A thought: how bad would it be to just `@whitelist_for_serdes` each of the actual FreshnessPolicy objects and make them NamedTuples?

python_modules/dagster-graphql/dagster_graphql/implementation/fetch_assets.py
```diff
@@ -243,3 +254,84 @@ def get_unique_asset_id(
...
     )
+
+
+def get_freshness_info(
+    instance: DagsterInstance,
+    asset_key: AssetKey,
+    asset_graph: AssetGraph,
+    freshness_policy: FreshnessPolicy,
+    batch_loader: BatchMaterializationLoader,
+):
```
sryza, line 265: A return type would be helpful here

python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -1,56 +1,122 @@
...
+
+    def get_partitions_def(self, asset_key: AssetKey) -> Optional[PartitionsDefinition]:
+        return self.partitions_defs_by_key.get(asset_key)
+
+    def get_partition_mapping(
+        self, asset_key: AssetKey, in_asset_key: AssetKey
+    ) -> PartitionMapping:
+        partitions_def = self.get_partitions_def(asset_key)
+        partition_mappings = self.partition_mappings_by_key.get(asset_key) or {}
+        return partition_mappings.get(
```
sryza, line 107: This is duplicating logic inside `AssetsDefinition.get_partition_mapping`, right? I think worth factoring it out into a common function so we don't accidentally change one without changing the other.

python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -1,56 +1,122 @@
...
+                if node.partitions_def_data
+                else None
+            )
+            group_names_by_key[node.asset_key] = node.group_name
+
+        return AssetGraph(
+            asset_dep_graph={"upstream": upstream, "downstream": downstream},
+            source_asset_keys=set(),  # TODO
+            partitions_defs_by_key=partitions_defs_by_key,
+            partition_mappings_by_key=dict(),  # TODO
```
sryza, line 91: Can we raise an error if any of the assets have partition mappings or if someone tries to use the partition mapping or something? Otherwise, I think it will be easy to write bugs that expect this to work.

python_modules/dagster/dagster/_core/host_representation/external_data.py
sryza: Hmm good question. I can't think of a problem.  Maybe @alangenfeld can?

python_modules/dagster-graphql/dagster_graphql/implementation/fetch_assets.py
```diff
@@ -243,3 +254,84 @@ def get_unique_asset_id(
...
+                batch_loader.get_latest_materialization_for_asset_key(upstream_key).timestamp
+            )
+            for upstream_key in relevant_upstream_keys
+        }
+
+        # same as above, but for the latest data that was available when we actually materialized
+        # the latest record
+        original_times = {
+            upstream_key: _convert_timestamp(
+                get_latest_materialization_record(
```
sryza, line 305: Would we be able to use the `TickInstanceQueryer` (maybe rename it) so we can cache these?

python_modules/dagster-graphql/dagster_graphql/implementation/fetch_assets.py
```diff
@@ -243,3 +254,84 @@ def get_unique_asset_id(
...
+
+    used_upstream_materialization_times = get_used_upstream_materialization_times_for_record(
+        instance=instance,
+        asset_graph=asset_graph,
+        upstream_keys=relevant_upstream_keys,
+        record=latest_record,
+    )
+
+    # not all FreshnessPolicies require this information, so don't fetch it if not necessary
+    if isinstance(freshness_policy, MaximumLatencyFreshnessPolicy):
```
sryza, line 290: Can we include a boolean property on the freshness policy instead of testing for particular subclasses?

python_modules/dagster/dagster/_core/host_representation/external_data.py
alangenfeld: yea i don't have full context but i assume these policy objects are declarative little data objects that would be fine as serdes tuples.

python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
clairelin135: yep, that should do it!

clairelin135 (approved):
Mostly looked at the graphql fields portions here, but those are looking good to me!

OwenKephart:
simplified this implementation a bunch based off of https://github.com/dagster-io/dagster/pull/10457, so would be great to get a once-over on this before I go ahead w/ merging it.

also split out GrapheneFreshnessPolicy from GrapheneAssetFreshnessInfo.

clairelin135:
@OwenKephart I think there might be some missing files, I don't see a `GrapheneFreshnessPolicy` object

sryza (approved):
All the non-GraphQL bits look good to me.

If we were less crunched for time, I'd say the AssetGraph changes should go into the own PR.

clairelin135 (approved):
GraphQL changes look good to me!

PR #10368, opened 2022-11-04
Mock each callsite of tqdm

gibsondan:
Summary:
Take 2 on mocking out tqdm - you have to list out each callsite for the patch to stick

Test Plan:
Run test_back_compat.py, no more progress bars


jmsanders (approved):
I think take 3 will be to rip it out entirely and take 4 will be to give up on this line of investigation.

PR #10372, opened 2022-11-04
Prevent segfaults

jmsanders:
We see frequent segfaults in our general test suite.

#10343  and #10368 proved to be ineffective. Even though we saw `tqdm` in the stack trace of each segfault, they continued happening after those changes were merged. We can probably revert these.

I'm still unable to reproduce this locally but the commits on this branch and associated builds show the segfaults reproduced on Buildkite.

Here's what I think is happening:

Dangling processes from previous tests are interacting with processes from the actual test. Initially, I thought we weren't cleaning these processes up at all. It turns out we're cleaning them up, but not fast enough - calling `process.terminate()` without `process.wait()` means that sometimes the process bleeds into the beginning of the next test. I verified this by printing in flight pids at the beginning of each test and seeing if the previous test's pid was in the set.

I've added the waits and I plan to re-run this build several times to see if I can reproduce the segfaults. If not, I say we merge.

We might still want to figure out why the particular interaction causes issues - there might be something about how we clean up a grpc server that interacts with the startup of a new one?


gibsondan (approved):
Here's hoping! This theory is very plausible to me. Do you want to revert the tqdm changes too while you're in here?

alangenfeld:
oooo nice 🙏

PR #10382, opened 2022-11-04
[dagit] Use a batch loader to load the partition health of several assets at once

bengotow:
### Summary & Motivation

This PR adds a new batch loader to our GraphQL layer that makes it efficient to ask for:

```
assetNodes($assetKeys: [AssetKey!]!) {
   id
   materializationCountByPartition {
        partition
        materializationCount
   }
}
```

Previously, the UI would make a separate GraphQL query for each asset node in your selection one by one. After seeing that these fall through to a `GROUP BY asset_key, partition` in run storage and the underlying API supports asking for multiple asset keys at once, I think this is likely faster?

The UI will now ask for up to 50 at once, and we can configure that count via a constant in the code.

I also removed an unused `loadMaterializations: true` optional param, since the batch loaders are now lazy.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/implementation/loader.py
+++ b/python_modules/dagster-graphql/dagster_graphql/implementation/loader.py
@@ -350,7 +350,7 @@ class CountByPartitionLoader:
             self._fetch()
 
         counts = self._counts_by_partition.get(asset_key)
-        if not counts:
+        if counts is None:
             check.failed(
                 f"Asset key {asset_key} not recognized for this loader.  Expected one of: {self._asset_keys}"
             )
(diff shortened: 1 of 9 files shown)


python_modules/dagster-graphql/dagster_graphql/implementation/loader.py
```diff
@@ -333,6 +333,37 @@ def _fetch(self) -> None:
         }
 
 
+class CountByPartitionLoader:
```
prha, line 336: Can we sanity check this against a user w/ a large number of assets?

bengotow: Hmm @prha  I'm not sure how best to check this without shipping it to cloud, but we could feature-flag the client-side part of this and that'd let us check the performance maybe? I did put a client-side cap of 50 asset keys in any given request for counts, and we could make that lower while we evaluate the performance.

Is this something that shadow dagit could do?

prha: Yeah, definitely can do it with shadow dagit... You will have to rebuild dagit in internal using a virtual env with your oss branch checked out.

PR #10383, opened 2022-11-04
[dagit] Create a new Asset version of the Job Partitions page

bengotow:
### Summary & Motivation

Summary of changes: https://www.loom.com/share/2e203e8dc38c4828a287b5b8cdf5cead

The partitions page of asset jobs has been very confusing to users because it still displays "op-style" partition health, where a "green" partition is one where the last job completed successfully, not one where a materialization exists.

This PR pulls common elements of the page out into components and creates two copies of the page -- one for assets and one for jobs. They will look the same after the 2 PRs in this stack land!


js_modules/dagit/packages/core/src/partitions/AssetJobPartitionsView.tsx
```diff
@@ -0,0 +1,139 @@
...
+              );
+              setOffset(nextOffset);
+              if (!showSteps) {
+                setShowSteps(true);
+              }
+            }}
+            tooltipMessage="Click to view per-step status"
+          />
+        </div>
+        {showSteps && <Box margin={{top: 16}} />}
```
prha, line 121: is the per-step status missing here?

How are we populating the per-asset state? I assume it's fetching the full asset status (materialization counts by partition) rather than filtering by last run?

bengotow: Hey hey! Ahh sorry I should have clarified, I split the per-asset status out into the second PR that is stacked on this one (https://github.com/dagster-io/dagster/pull/10385) because it required a lot of work.

I think the `showSteps` should be renamed to showAssets though.

We're populating the per-asset state using materializationCountByPartition + the batch loader I added in https://github.com/dagster-io/dagster/pull/10382. 

We're not filtering by last run but I think that's ok? The goal of this work is to remove run / op-centric partition state so that this page shows asset partition status that matches what you see when you view individual asset pages or go to launch asset backfills. (I think?)

js_modules/dagit/packages/core/src/assets/usePartitionNameForPipeline.tsx
```diff
@@ -0,0 +1,68 @@
...
+    partitionSet:
+      partitionSetsData?.partitionSetsOrError.__typename === 'PartitionSets'
+        ? partitionSetsData.partitionSetsOrError.results[0]
+        : undefined,
+    partitionSetError:
+      partitionSetsData?.partitionSetsOrError.__typename === 'PipelineNotFoundError' ||
+      partitionSetsData?.partitionSetsOrError.__typename === 'PythonError'
+        ? partitionSetsData.partitionSetsOrError
+        : undefined,
+  };
```
hellendag, line 33: Worth a `useMemo`?

js_modules/dagit/packages/core/src/partitions/AssetJobPartitionsView.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const [pageSize, setPageSize] = React.useState(60);
+  const [offset, setOffset] = React.useState<number>(0);
+  const [showSteps, setShowSteps] = React.useState(false);
+
+  React.useEffect(() => {
+    if (viewport.width && !showSteps) {
+      // magical numbers to approximate the size of the window, which is calculated in the step
+      // status component.  This approximation is to make sure that the window does not jump as
+      // the pageSize gets recalculated
+      const _approximatePageSize = Math.ceil((viewport.width - 330) / 32) - 3;
```
hellendag, line 57: Possible to clarify these values with some consts?

js_modules/dagit/packages/core/src/partitions/PipelinePartitionsRoot.tsx
```diff
@@ -30,80 +25,52 @@ export const PipelinePartitionsRoot: React.FC<Props> = (props) => {
...
-          );
-        }
+  if (!partitionSet) {
+    return (
+      <Box padding={{vertical: 64}}>
+        <NonIdealState
+          icon="error"
+          title="Partitions"
+          description={
+            <p>
```
hellendag, line 57: I'm not actually sure what style to expect from `p` tags in Dagit...

hellendag (approved):

js_modules/dagit/packages/core/src/partitions/AssetJobPartitionsView.tsx
bengotow: Oh man good call on this one...

PR #10389, opened 2022-11-04
Add support for loading cacheable assets from module

benpankow:
## Summary

The various 'load assets from module' functions currently ignore cacheable assets.

This PR allows them to be imported just like any other asset or source asset. It also adds some utility functions to mass-apply groups and prefixes to a cacheable asset definition's generated assets. Right now, doing so is tricky because these functions require an explicit mapping of asset key to group, but asset keys aren't known at initialization time for cacheable assets.

## Test Plan

New unit test.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
@@ -2,7 +2,7 @@ import hashlib
 import inspect
 import json
 from abc import ABC, abstractmethod
-from typing import AbstractSet, Any, List, Mapping, NamedTuple, Optional, Sequence, cast
+from typing import AbstractSet, Any, List, Mapping, NamedTuple, Optional, Sequence
 
 import dagster._check as check
 import dagster._seven as seven
@@ -15,7 +15,6 @@ from dagster._core.definitions.resource_definition import ResourceDefinition
 from dagster._core.definitions.resource_requirement import ResourceAddable
 from dagster._serdes import whitelist_for_serdes
 from dagster._utils import frozendict, frozenlist, make_readonly_value
-from dagster._utils.merger import deep_merge_dicts
 
 
 @whitelist_for_serdes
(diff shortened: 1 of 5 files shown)


benpankow:
* master
  * **PR #10389** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10389" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10389?utm_source=stack-comment).

sryza (commented):
I think there's a gnarly edge case here that isn't handled.  I'm not sure if there's a good way to handle it, but wanted to point it out:

If asset1 depends on asset2 and we supply a "prefix1" prefix while loading them, we'll update asset2's dependency to point to prefix1.asset1.

However, if asset1 is a cacheable asset, this PR doesn't do that. Right?

I think there's a good chance we're stuck with this behavior, but I haven't thought too much about whether there's a way to avoid it. At the least, we should document it somewhere.

python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
```diff
@@ -245,13 +271,58 @@ def _get_hash(self) -> str:
...
+                        else k.path
+                    )
+                    for k in assets_def.asset_keys
+                    if self._prefix_for_all_assets
+                },
+            ),
+        )
+        input_asset_key_replacements = cast(
+            Mapping[AssetKey, AssetKey],
+            deep_merge_dicts(
```
OwenKephart, line 309: it's not clear to me why this (or the equivalent thing for output_asset_key_replacements) is using deep_merge_dicts. Is there a case in which you would want to have both _prefix_for_all_assets and input_key_replacements set? If so, then those transformations should be done in sequence (right now _prefix_for_all_assets would override the input_key_replacements), and if not, then this can just be done in a conditional (i.e. if self._prefix_for_all_assets then input_key_replacements = that dictionary comprehension otherwise it = self.input_key_replacements)

python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
```diff
@@ -245,13 +271,58 @@ def _get_hash(self) -> str:
...
+                self._input_asset_key_replacements,
+                {
+                    k: AssetKey(
+                        path=self._prefix_for_all_assets + list(k.path)
+                        if self._prefix_for_all_assets
```
OwenKephart, line 314: no need to have both this conditional and the one below in this comprehension I think

OwenKephart (approved):
approving to unblock, but take a look at comments

python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
benpankow: Yeah, I think you're right to identify that they should be mutually exclusive. I will add a check to explicitly error if multiple are specified, and use a conditional.

PR #10393, opened 2022-11-04
[freshness-policy] add latency freshness policy

OwenKephart:
### Summary & Motivation

Add a new type of FreshnessPolicy which is relative to the latest available data for the upstream assets.

latest_upstream_materialization_times is going to be relative to the record that we're evaluating this minutes_late value for, so we'll probably end up storing this value in the asset event tags similar to the used_upstream_materialization_times, for efficiency. It corresponds to the newest asset materializations that were available at the time we produced this materialization.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/freshness_policy.py
+++ b/python_modules/dagster/dagster/_core/definitions/freshness_policy.py
@@ -24,7 +24,8 @@ class FreshnessPolicy(ABC):
     def minutes_late(
         self,
         evaluation_time: datetime,
-        upstream_materialization_times: Mapping[AssetKey, Optional[datetime]],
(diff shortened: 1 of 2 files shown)


sryza (commented):
> latest_upstream_materialization_times is going to be relative to the record that we're evaluating this minutes_late value for

I think this might be different than the way I was thinking about it.

Suppose we've got A -> B, with `FreshnessPolicy.maximum_latency(5)` on B.  A materializes at 9:00 am and B materializes at 9:04 am. Then A materializes again at 9:06 am, and now it's 10:00 am and we still haven't gotten any materializations for B.  I would say that B is 54 minutes late, even though the latest materialization for B was within the maximum latency of the latest materialization of A at the time that B materialized.

sryza (approved):

PR #10394, opened 2022-11-05
[dagit] Update partner compute_kind tags to latest designs

bengotow:
### Summary & Motivation

This is a small PR that replaces the artwork used in the op tags (which are linked to `compute_kind`). The new ones are brighter, slightly larger, and use the brand colors as backgrounds instead of text colors.

Here's the hardcoded set: 

image

And some random strings with autogenerated colors (which ideally will look good side by side):

image

I also removed the "larger size as you zoom out" -- I don't think these are important enough to justify getting big to remain legible!


braunjj (approved):
Wonderful, Thank you!

PR #10398, opened 2022-11-06
cached_method decorator

sryza:
### Motivation

I found myself wanting this a few times while working on partitioned asset reconciliation.

### Docstring

Caches the results of a method call.

Usage:

    .. code-block:: python

        class MyClass:
            @cached_method
            def fetch_value_from_database(self, key):
                ...

The main difference between this decorator and `functools.lru_cache(max_size=None)` is that each instance of the class whose method is decorated gets its own cache. This means that the cache can be garbage-collected when the object is garbage-collected.

A more subtle difference between this decorator and `functools.lru_cache` is that this one prioritizes preventing mistakes over ease-of-use and performance. With `functools.lru_cache`, these three invocations would all result in separate cache entries:

    .. code-block:: python

        class MyClass:
            @cached_method
            def a_method(self, arg1, arg2):
                ...

        obj = MyClass()
        obj.a_method(arg1="a", arg2=5)
        obj.a_method(arg2=5, arg1="a")
        obj.a_method("a", 5)

With this decorator, the first two would point to the same cache entry, and non-kwarg arguments are not allowed.

### DIFF
--- a/python_modules/dagster/dagster/_utils/cached_method.py
+++ b/python_modules/dagster/dagster/_utils/cached_method.py
@@ -1,0 +1,108 @@
+from functools import wraps
+
+from dagster import _check as check
+
+NO_VALUE_IN_CACHE_SENTINEL = object()
+
+
+def cached_method(method):
+    """
+    Caches the results of a method call.
+
+    Usage:
+
+        .. code-block:: python
+
+            class MyClass:
+                @cached_method
+                def fetch_value_from_database(self, key):
+                    ...
+
+    The main difference between this decorator and `functools.lru_cache(max_size=None)` is that each
+    instance of the class whose method is decorated gets its own cache. This means that the cache
+    can be garbage-collected when the object is garbage-collected.
+
+    A more subtle difference between this decorator and `functools.lru_cache` is that this one
+    prioritizes preventing mistakes over ease-of-use and performance. With `functools.lru_cache`,
+    these three invocations would all result in separate cache entries:
+
+        .. code-block:: python
+
+            class MyClass:
+                @cached_method
+                def a_method(self, arg1, arg2):
+                    ...
+
+            obj = MyClass()
+            obj.a_method(arg1="a", arg2=5)
+            obj.a_method(arg2=5, arg1="a")
+            obj.a_method("a", 5)
+
+    With this decorator, the first two would point to the same cache entry, and non-kwarg arguments
+    are not allowed.
+    """
+
+    @wraps(method)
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_utils/cached_method.py
```diff
@@ -0,0 +1,110 @@
...
+        key = _make_key(args, kwargs)
+        cached_result = cache.get(key, NO_VALUE_IN_CACHE_SENTINEL)
+        if cached_result is not NO_VALUE_IN_CACHE_SENTINEL:
+            return cached_result
+        else:
+            result = method(self, *args, **kwargs)
+            cache[key] = result
+            return result
+
+    return helper
```
alangenfeld, line 63: could be nice to have the `cache_info` setup from lru_cache as well for debugging hits/misses

alangenfeld (commented):
neato

>  This means that the cache can be garbage-collected when the object is garbage-collected.

add a test that demonstrates /  asserts this - can use `objgraph` 

> non-kwarg arguments are not allowed.

nit: add a test for this as well

python_modules/dagster/dagster_tests/general_tests/utils_tests/test_cached_method.py
```diff
@@ -0,0 +1,80 @@
...
+            return ValueClass("abc")
+
+    obj = MyClass()
+    assert objgraph.count("MyClass") == 1
+    assert objgraph.count("KeyClass") == 0
+    assert objgraph.count("ValueClass") == 0
+
+    obj.my_method(arg1=KeyClass("1234"))
+    assert objgraph.count("MyClass") == 1
+    assert objgraph.count("KeyClass") == 1
```
alangenfeld, line 73: is the `KeyClass` instance guaranteed to stick around? the cache doesn't hold a ref right, just uses it to calc a key? Think this might flake if gc runs at different times and clears that

alangenfeld (approved):
good stuff - didn't realize this problem with lru_cache & self was a thing

`CrossRepoAssetDependedByLoader` has a call site that looks like this should be used on

python_modules/dagster/dagster_tests/general_tests/utils_tests/test_cached_method.py
sryza: The cache key includes a ref. If we didn't hold a ref, we'd need some sort of scheme for generating a key that respects `__eq__` of the key class, which I don't think is possible in the general case.

PR #10401, opened 2022-11-06
Resolver -> AssetGraph

sryza:
### Summary & Motivation

This renames AssetGraph to Resolver, factors it out into its own module, and moves the resolution logic onto the subclasses of AssetSelection.

The idea is that we can then use AssetGraph in a bunch of other places that aren't centered on AssetSelection (e.g. partitioned reconciliation sensors).

This shouldn't change any user-facing behavior.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -1,0 +1,30 @@
+from typing import Sequence, Union
+
+import dagster._check as check
+from dagster._core.selector.subset_selector import (
+    generate_asset_dep_graph,
+    generate_asset_name_to_definition_map,
+)
+
+from .assets import AssetsDefinition
+from .source_asset import SourceAsset
+
+
+class AssetGraph:
+    def __init__(self, all_assets: Sequence[Union[AssetsDefinition, SourceAsset]]):
+        assets_defs = []
+        source_assets = []
+        for asset in all_assets:
+            if isinstance(asset, SourceAsset):
+                source_assets.append(asset)
+            elif isinstance(asset, AssetsDefinition):
+                assets_defs.append(asset)
+            else:
+                check.failed(f"Expected SourceAsset or AssetsDefinition, got {type(asset)}")
+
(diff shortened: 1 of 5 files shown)


python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -0,0 +1,30 @@
...
+        for asset in all_assets:
+            if isinstance(asset, SourceAsset):
+                source_assets.append(asset)
+            elif isinstance(asset, AssetsDefinition):
+                assets_defs.append(asset)
+            else:
+                check.failed(f"Expected SourceAsset or AssetsDefinition, got {type(asset)}")
+
+        self.assets_defs = assets_defs
+        self.asset_dep_graph = generate_asset_dep_graph(assets_defs, source_assets)
```
OwenKephart, line 26: to be honest, I don't remember why we keep this dependency graph in the language of asset key strings after we create it. is this necessary?

python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -0,0 +1,30 @@
...
+from dagster._core.selector.subset_selector import (
+    generate_asset_dep_graph,
+    generate_asset_name_to_definition_map,
+)
+
+from .assets import AssetsDefinition
+from .source_asset import SourceAsset
+
+
+class AssetGraph:
```
OwenKephart, line 13: would be nice to have some convenient properties like `upstream_mapping` and `downstream_mapping` (from the underlying asset_dep_graph, but that can come later

OwenKephart (approved):
I'm into this, but I do think we'll want to refactor the AssetGraph to move away from the string-based representation of asset keys -- it's a bit of a headache converting back and forth in different scenarios, so I think it'd be nice if we could just settle on AssetKeys instead of strings

sryza:
@OwenKephart check this out: https://github.com/dagster-io/dagster/pull/10402

OwenKephart:
> #10402

nice ok, that handles all my concerns

PR #10402, opened 2022-11-06
operate on keys instead of strings in AssetSelection and AssetGraph

sryza:
### Summary & Motivation

This enables `AssetGraph` and the operations within asset selection resolution to operate on asset keys instead of strings.  It changes the underlying implementation inside subset_selector to support arbitrary key types, not just strings. Advantages:
- Performance: less time spent converting to/from the string representations of asset keys
- Readability: less characters converting to/from the string representations of asset keys
- Wider use: `AssetGraph` will be able to be used in parts of the codebase that deal in asset keys

It also moves the selection code from `FrozenSet` to `AbstractSet`, trusting the type system to enforce that these return values don't get mutated instead of doing lots of copies.

It shouldn't change any user-facing behavior.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -1,10 +1,7 @@
 from typing import Sequence, Union
 
 import dagster._check as check
-from dagster._core.selector.subset_selector import (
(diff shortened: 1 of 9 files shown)


OwenKephart (approved):
<3

PR #10410, opened 2022-11-07
Support partitions in asset reconciliation sensor

sryza:
### Summary & Motivation

This extends the reconciliation behavior in https://github.com/dagster-io/dagster/pull/10435 from non-partitioned assets to asset partitions.

An asset partition is considered "unreconciled" if any of:

- This sensor has never tried to materialize it and it has never been materialized.
- Any of its parent asset partitions have been materialized more recently than it has (this will get more sophisticated with @smackesey's work).
- Any of its parents asset partitions are unreconciled.

The sensor materializes unreconciled asset partitions and avoids materializing asset partitions before their parent asset partitions are reconciled.

It supports all partition mappings, including fan-in and fan-out. It also supports dependencies between partitioned and non-partitioned assets and vice versa.


sryza:
> MySQL has a max row size of 66KB

If I'm reading this correctly, BLOB and TEXT values are stored separately from the rest of the row and can exceed 66KB: https://dev.mysql.com/doc/refman/8.0/en/column-count-limit.html#:~:text=The%20internal%20representation%20of%20a,the%20rest%20of%20the%20row.

That said, you're right that this can come out to be a lot. Let me take a stab and see if we can store partition ranges instead of all the partitions.

sryza:
@clairelin135 i pushed up a change that should address the cursor partitions issue

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -414,26 +539,49 @@ def reconcile(
...
+
+    run_requests = []
+
+    for (
+        _,
+        partition_key,
+    ), asset_keys in assets_to_reconcile_by_partitions_def_partition_key.items():
+        tags = {**(run_tags or {})}
+        if partition_key is not None:
+            tags[PARTITION_NAME_TAG] = partition_key
```
clairelin135, line 570: I think we'll need to assign different tags for multi-dimensional partitions.

I can add this in a future PR though--probably worth aggregating all of the partition tag logic into one function to avoid this logic being duplicated over many different call sites

clairelin135 (approved):
This is looking good to me 🚢 

I don't think the cursor will be too large, but if in the future it ends up causing some issues, we can change `ASSET_MATERIALIZATION_PLANNED` events to also store partition keys which should allow us to query for never materialized/requested partition keys from the instance.

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
sryza: Ah, makes sense. Sounds good.

PR #10413, opened 2022-11-07
Add methods to AssetGraph for traversing asset partitions graph

sryza:
### Summary & Motivation

Adapted from the `get_downstream_partition_keys` on `MultiAssetSensorEvaluationContext`.

This will make that functionality available to partitioned asset reconciliation sensors.


pyrightconfig.json
```diff
@@ -0,0 +1,22 @@
+{
```
clairelin135, line 1: do we need this file?

clairelin135 (approved):
The changes in this PR are looking good to me. Test failures seem real though--good to merge after those are resolved

Issue #8877, opened 2022-07-14
Provide a way to send extra arguments/flags to the database connection through sqlalchemy


RMHogervorst:
### What's the use case?

When you deploy dagster in a kubernetes cluster with a service mesh, or when you have an RDS store it is very useful to be able to tell dagster that it should retry connections, auto refresh db connections etc. By default the connections to postgres at least, think they live forever.

### Ideas of implementation

Ideally I would supply these arguments in the same place we supply config for schedule storage/ event_log_storage etc.  => dagster.yaml.

There is a way to pass arguments like `pool_pre_ping` or `pool_recycle` to sqlalchemy in one of the issues mentioned below someone pointed to this locaiton <https://github.com/dagster-io/dagster/blob/4d0b512b5f52c0dcb80c49994d8f65c4f122649a/python_modules/libraries/dagster-postgres/dagster_postgres/utils.py#L151-L153> 

So I guess you can have an argument like `extra_args` or something next to port, db_name, hostname, password, and username. 

## Situation
I have dagster deployed in an kubernetes cluster with istio as connection mesh. I have postgres as backend.
Standard timeouts on istio http connections is 15 minutes. 

## problem
Every morning when I browse to dagit to see the status of dagster I see these timout warning appear. Similar to 
https://github.com/dagster-io/dagster/issues/8462 and https://github.com/dagster-io/dagster/issues/4111, and issues described in the dagster slack channel.  https://github.com/dagster-io/dagster/issues/3719

My diagnosis is: when I make the first connection in the morning, dagit will try to use the old connection to the database. But that doesn't work, because postgres default timeouts are 4 hours. 

What I would like to tell the underlying sqlalchemy connection is:
* get a new connection if it is older than x
* ping intermittendly 

This is all possible, but not accessable in dagit.
I would like to supply extra arguments to the database connection. 

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

RMHogervorst:
Since this is merged, I think we can close this issue

PR #10416, opened 2022-11-07
Expose sqlalchemy pool recycle option for mysql/postgres

adam-bloom:
### Summary & Motivation
We experience intermittent dagit postgres connection issues (similar to #8462 and #8877).

This PR exposes a new parameter to the dagit invocation that is passed to sqlalchemy for mysql and postgres connections when in dagit mode for `pool_recycle`.

I put a little thought into making this just be `extra_kwargs` or something, but I couldn't think of a non-clunky way to pass that via a CLI arg (string encoded json?)

Updated unit tests.

### DIFF
--- a/python_modules/dagit/dagit/cli.py
+++ b/python_modules/dagit/dagit/cli.py
@@ -30,6 +30,7 @@ DEFAULT_DAGIT_HOST = "127.0.0.1"
 DEFAULT_DAGIT_PORT = 3000
 
 DEFAULT_DB_STATEMENT_TIMEOUT = 15000  # 15 sec
+DEFAULT_POOL_RECYCLE = 3600  # 1 hr
 
 
 @click.command(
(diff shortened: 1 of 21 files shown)


python_modules/dagit/dagit/cli.py
```diff
@@ -86,6 +87,13 @@ def create_dagit_cli():
     type=click.INT,
     show_default=True,
 )
+@click.option(
+    "--db-pool-recycle",
+    help="The maximum age of a connection to use from the sqlalchemy pool without connection recycling. Set to -1 to disable. Not respected in all configurations.",
```
alangenfeld, line 92: > Set to -1 to disable

is this how it works at the sqlalchemy level?

alangenfeld (commented):
awesome! thanks for taking the time to do this. Running CI now

python_modules/dagit/dagit/cli.py
adam-bloom: Yup.

> this setting causes the pool to recycle connections after the given number of seconds has passed. It defaults to -1, or no timeout.

alangenfeld:
seeing one issue in the helm json schema test - looks like casing got changed but the file not regenerated?
```
[2022-11-08T15:18:46Z]                  "dbPoolRecycle": {
--
  | [2022-11-08T15:18:46Z] -                    "title": "DbPoolRecycle",
  | [2022-11-08T15:18:46Z] +                    "title": "Dbpoolrecycle",

```

adam-bloom:
> think you just need to re-run `dagster-helm schema apply`

Yes indeed, fixed!

PR #10419, opened 2022-11-07
Install project-fully-featured

jmsanders:
I'm not sure how this was succeeding in the past. Maybe we were just lucky that everything we needed was installed elsewhere?

Now we're missing sklearn though:

https://buildkite.com/dagster/dagster/builds/38482#01845319-435b-44cc-83a7-d53fe6831c30

### DIFF
--- a/examples/project_fully_featured/setup.py
+++ b/examples/project_fully_featured/setup.py
@@ -37,7 +37,7 @@ setup(
         "fsspec",
         "s3fs",
         "scipy",
-        "sklearn",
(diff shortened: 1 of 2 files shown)


jmsanders:
I'm really not sure why this first broke https://buildkite.com/dagster/dagster/builds/38469

Maybe some upstream dependency used to include `sklearn` but dropped the dependency over the weekend?

sryza (approved):

PR #10423, opened 2022-11-07
More process cleanup in grpc tests

gibsondan:
Summary:
A few other files where we probably need to terminate() and/or wait() to ensure processes get cleaned up

### DIFF
--- a/python_modules/dagster/dagster_tests/general_tests/grpc_tests/test_ping.py
+++ b/python_modules/dagster/dagster_tests/general_tests/grpc_tests/test_ping.py
@@ -45,25 +45,30 @@ def test_server_socket():
             assert DagsterGrpcClient(socket=skt).ping("foobar") == "foobar"
         finally:
             interrupt_ipc_subprocess_pid(server_process.pid)
+            server_process.terminate()
+            server_process.wait()
 
 
 @pytest.mark.skipif(seven.IS_WINDOWS, reason="Unix-only test")
 def test_process_killed_after_client_finished():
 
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster_tests/general_tests/grpc_tests/test_watch_server.py
```diff
@@ -179,142 +198,6 @@ def should_not_be_called(*args, **kwargs):
     assert called["on_updated"] == new_server_id
 
 
-@pytest.mark.skip
-def test_grpc_watch_thread_server_complex_cycle():
```
gibsondan, line 183: these are long bit rotted

python_modules/dagster/dagster_tests/general_tests/grpc_tests/test_ping.py
```diff
@@ -45,25 +45,30 @@ def test_server_socket():
...
         finally:
             interrupt_ipc_subprocess_pid(server_process.pid)
+            server_process.terminate()
+            server_process.wait()
 
 
 @pytest.mark.skipif(seven.IS_WINDOWS, reason="Unix-only test")
 def test_process_killed_after_client_finished():
 
     server_process = GrpcServerProcess()
```
jmsanders, line 55: I wonder if we could use a context manager so this will always close itself.

jmsanders (approved):
I'm a little scared that we have no way of knowing if we have all of these or when we introduce new ones - but 🤞 this gets us fully green in this test suite.

Issue #10412, opened 2022-11-07
JobDefinition.execute_in_process not switching IO manager to mem_io_manager as per docs


northernSage:
### What's the issue or suggestion?

Dagster docs on https://docs.dagster.io/concepts/io-management/io-managers#built-in-io-managers state that:

> The default IO manager, [fs_io_manager](https://docs.dagster.io/_apidocs/io-managers#dagster.fs_io_manager), stores and retrieves values from pickle files in the local filesystem. If a job is invoked via [JobDefinition.execute_in_process](https://docs.dagster.io/_apidocs/jobs#dagster.JobDefinition.execute_in_process), **the default IO manager is switched to [mem_io_manager](https://docs.dagster.io/_apidocs/io-managers#dagster.mem_io_manager), which stores outputs in memory.**

However, this is not happening in practice and `fs_io_manager` is still being used with `execute_in_process`. We came across this issue after bumping dagster from 0.15.x to latest release, which caused one of our unit tests to blow up with the following:

```
test_regression_graph.py::test_download_and_train_no_time - TypeError: no default __reduce__ due to non-trivial __cinit__
...
pickle.dump(obj, write_obj, PICKLE_PROTOCOL)
File "stringsource", line 2, in _catboost._PoolBase.__reduce_cython__

```

The reason why pickling is breaking here is probably not relevant for this issue, but it happens when you try pickling a `Pool` object from `catboost` library (one can make it work by quantizing and using their provided `save` api). What's important here is that the `ops` in question is being executed with `execute_in_process` and it's clearly trying to pickle the object instead of storing it in memory.

To work around this, we had to:

```
       result: ExecuteInProcessResult = (
        train_and_save_traffic_shaping_server_classification_graph.execute_in_process(
            resources={
                "io_manager": mem_io_manager,   # added this line
                "clock": MockClock(),
            }
        )
    )
```
explicitly setting the `io_manager` to `mem_io_manager`

### Additional information

dagster version: latest
python release: 3.1
Execution environment: Debian latest

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

sryza:
@dpeng817 @johannkm mind taking a look?

PR #10431, opened 2022-11-08
Fix execute_in_process io manager behavior for graphs and aliased graphs

dpeng817:
### Summary & Motivation
https://github.com/dagster-io/dagster/issues/10412 - behavioral regression for graphs executed using execute_in_process. Haven't found the exact code change that is the culprit, but it's only a regression for graphs and aliased graphs. This PR fixes by having all flow through the job execute_in_process codepath, which explicitly swaps the default io manager. 
Added a test that ensures same default io manager behavior for all 3 cases (job, graph, and aliased graph)

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/composition.py
+++ b/python_modules/dagster/dagster/_core/definitions/composition.py
@@ -656,7 +656,6 @@ class PendingNodeInvocation:
             )
 
         from dagster._core.execution.build_resources import wrap_resources_for_execution
-        from dagster._core.execution.execute_in_process import core_execute_in_process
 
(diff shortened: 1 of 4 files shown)


dpeng817:
* master
  * **PR #10431** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10431" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10431?utm_source=stack-comment).

johannkm (commented):
I don't feel like I have enough context to review

PR #10424, opened 2022-11-07
docs(cacheable-assets): mark unimplemented methods as abstract

rexledesma:
### Summary & Motivation
As the title.

### How I Tested These Changes
existing pytest, local

PR #10434, opened 2022-11-08
Implement abstract methods

jmsanders:
Alternatively, we could revert:

https://github.com/dagster-io/dagster/pull/10424

I don't know enough about the API here - should this entire class be marked abstract (like I did here) or should we instead provide default implementations so we can initialize it directly?

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/cacheable_assets.py
@@ -1,7 +1,7 @@
 import hashlib
 import inspect
 import json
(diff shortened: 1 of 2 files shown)


rexledesma (approved):
This looks right. I don't think we need to mark `CacheableAssetDefinition` as an ABC since `ResourceAddable` is already an ABC... I'm not sure if python can understand that inheritance if there are two ABCs in the tree. 

I'm wondering how the initial PR passed CI checks - I only merged since it was green

rexledesma:
Eh I stand corrected since the build passed

Issue #10159, opened 2022-10-23
Partition matrix does not show correct partition state after partial backfills


aroig:
### Dagster version

commit ca87aa3c95204c7941e675318f714d24a4990132

### What's the issue?

On a repo with several partitioned assets, a backfill of only one asset would hide previous state of partitions in the asset x partition matrix on Job > Partitions > Show per-step status.

Here is  a picture of the partition matrix after a partial backfill. The dots for asset 1 which were previously green are now white. However, clicking one dot still shows the correct run history.
![after-partial-backfill](https://user-images.githubusercontent.com/1743143/197404078-8d26c052-a6b0-47ec-93fb-a30870445202.png)

### What did you expect to happen?

I would expect that after a partial backfill, the assets that were not touched, keep the state they had before (red or green color) on the partition matrix.

### How to reproduce?

With the following code

``` python
import dagster

partitions = dagster.StaticPartitionsDefinition(["A", "B", "C"])

@dagster.asset(
    partitions_def=partitions,
    group_name="all"
)
def asset_1():
    return 1

@dagster.asset(
    partitions_def=partitions,
    group_name="all"
)
def asset_2(asset_1: int):
    return asset_1 + 1

job_all = dagster.define_asset_job(
    name="all",
    selection=dagster.AssetSelection.groups("all")
)

@dagster.repository
def my_repository():
    return [
        asset_1,
        asset_2,
        job_all,
    ]
```
* Go to job "all" > Overview > Materialize button. Select all partitions, to materialize everything.
* Partitions page would show a green partition bar, and when expanding the asset x partition matrix, all dots are green.
* Go to job "all" > Overview. Select asset_2, and Materialize button. Then Select all partitions.
* After the backfill is complete, go to Partitions page. The partition bar shows all partitions green, as expected. However, if you expand the asset x partition matrix, all dots for asset_1 are blank now.

### Deployment type

Local

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

sryza:
@prha - does this look like the expected behavior to you?

I believe it may be the expected behavior, because the Partitions tab of the Job page takes a run-oriented view of partitions, rather than an asset-oriented view.  From the run-oriented perspective, the status of a Job partition is based on the most recent run that targeted that partition.

[...]

Issue #8706, opened 2022-07-01
enable backfilling multiple asset partitions within a single step (coalesced backfills)


sryza:
Here's a How-to guide to how to do this using Dagster's current somewhat-limited functionality: https://github.com/dagster-io/dagster/discussions/11653.

---
Requests:
* https://dagster.slack.com/archives/C01U954MEER/p1656662232023999?thread_ts=1656596814.202879&cid=C01U954MEER
* https://dagster.slack.com/archives/C01U954MEER/p1658781926863839
* https://dagster.slack.com/archives/C01U954MEER/p1660634859297209
* https://dagster.slack.com/archives/C01U954MEER/p1667932777213229?thread_ts=1667922601.771249&cid=C01U954MEER

Implementing this requires some way of allowing users to specify how to break up the partitions in a backfill into steps. This could be a function that accepts a list of partition keys and returns a list of PartitionKeyRanges. The default implementation of that function would return a PartitionKeyRange per partition key:
```
def step_partitions(context: LaunchBackfillContext) -> Sequence[PartitionKeyRange]:
    return [PartitionKeyRange(partition_key, partition_key) for partition_key in context.partition_keys]
```

but a custom implementation could return a single PartitionKeyRange:

```
def step_partitions(context: LaunchBackfillContext) -> Sequence[PartitionKeyRange]:
    return [context.partition_key_range]
```

There's the question of where this function gets specified:
* Is it on a `PartitionsDefinition`?
* Is it a separate attribute on each asset?
* Is it on the job?

There's also the question of PartitionKeyRanges vs. partition lists. In the case where you just want to fire off a SQL query that operates over a range, Dagster shouldn't have to do O(# partitions) work or pass around O(# partitions) data, but it might be too restrictive to require that steps always handle continuous sets of partitions.

zczhuohuo:
Any update or any workaround? Currently，It will be very slow when backfills a large set of partitions

Nintorac:
Just throwing in a usecase I've run into.

I have a large archive sitting on the net somewhere that contains many files named by their hash in a subfolders named for the initial char of the hash. eg `e/e990ff1.file` or `f/f556439485.file`. 

A single partition is pushing it for memory and it's not possible to seek a single partition efficiently.

For now my sol'n is to skip the IOManager for this asset, partition within the asset code itself and then on the consuming asset load the correct partition manually. 

It might be cool to be able to specify the partition key of a particular materialization  with something like
```
[...]

PR #10441, opened 2022-11-08
enable launching a run that targets a range of asset partitions

sryza:
### Summary & Motivation

With this change, you can set these two tags on a run instead of setting an individual partition tag:
- dagster/asset_partition_range_start
- dagster/asset_partition_range_end

The run will then expose that range within ops and I/O managers and generate asset materializations for all the partitions in the range.

Motivated by this request: https://dagster.slack.com/archives/C01U954MEER/p1667922601771249.

### Future work

This gets us part of the way to closing https://github.com/dagster-io/dagster/issues/8706. There are some gaps:
- This experience isn't yet integrated with the backfill dialog. You need to use the launchpad (shift+click "Materialize") and supply the tags using the "Edit tags" button.
- The partitions bar on the relevant asset details pages will be updated, but the Partitions tab on a job page won't reflect partitions that are backfilled this way. Relevant discussion here about making the Partitions tab on job pages more asset-oriented, which should address this: https://github.com/dagster-io/dagster/issues/10159.
- The built-in I/O managers don't yet support output partition key ranges.

- Added test
- Tried it out in Dagit and made sure that that the partitions show up correctly on the asset details page

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_layer.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_layer.py
@@ -612,11 +612,22 @@ class AssetLayer:
                 node_output_handle = NodeOutputHandle(
                     check.not_none(inner_node_handle), inner_output_def.name
                 )
-                partition_fn = lambda context: {context.partition_key}
+
+                def partitions_fn(context: "OutputContext") -> AbstractSet[str]:
+                    from dagster._core.definitions.partition import PartitionsDefinition
+
+                    if context.has_partition_key:
+                        return {context.partition_key}
+
+                    return set(
+                        cast(
+                            PartitionsDefinition, context.asset_partitions_def
+                        ).get_partition_keys_in_range(context.asset_partition_key_range)
+                    )
 
                 asset_info_by_output[node_output_handle] = AssetOutputInfo(
                     asset_key,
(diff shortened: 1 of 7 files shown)


sryza:
> I think if we started representing full partitions using event tags instead of just the partition column, we could fulfill the partition status queries using many to one partition => materializations.

Oh interesting - I like this.

Any objections to pushing that to a later PR? There's a lot of stuff that relies on the current scheme that would need to be updated.

prha:
I think that's fine, but cc @clairelin135  (this is an argument for mirroring, if we are eventually going to use tags for asset partition status)

clairelin135 (approved):
I like this, this change is looking good to me!

The many partitions => one materialization idea sounds good to me. I imagine we'd have a tag for each partition, and the partition column in event log storage would contain a range.

This means we would have to mirror the whole partition key into the new table, after disallowing invalid characters in static partition keys for 1.1.0. I can file a PR to add this this week

clairelin135:
@prha Yeah, I just wonder if it would be expensive to have to do that match against the partition column. If one materialization had multiple multidimensional partitions you wouldn't be able to learn which partitions were materialized per event with just per-dimension tags without comparing against the partition column

PR #10452, opened 2022-11-09
add cloud storage-based compute log managers

prha:
### Summary & Motivation

This PR creates an abstract implementation of the captured log manager (and corresponding compute log manager API) that uses the local compute log manager to capture logs and then uploads them to cloud-backed storage.

Also adds a cloud-based captured log manager test suite that tests the streaming features (which involves partial uploads at a polling interval).

BK


alangenfeld (approved):
did you do any manual testing of the polling and its UX? 

I think this looks good - great work here

prha:
Yeah, manually tested this using two local instances sharing an S3 bucket.

PR #10457, opened 2022-11-10
[freshness-policies] refactor FreshnessPolicy

OwenKephart:
### Summary & Motivation

Pares down the FreshnessPolicy class to a single entity, with some values renamed


sryza (approved):
Love to see lines of code go away

PR #10458, opened 2022-11-10
Fix step deletegating executor with delayed op retries

johannkm:

### DIFF
--- a/python_modules/dagster/dagster_tests/execution_tests/engine_tests/test_step_deletegating_executor_retries.py
+++ b/python_modules/dagster/dagster_tests/execution_tests/engine_tests/test_step_deletegating_executor_retries.py
@@ -1,0 +1,102 @@
+# pylint: disable=unused-argument
+import subprocess
+
+from mock import MagicMock
+
+from dagster import OpExecutionContext, RetryRequested, executor, job, op, reconstructable
+from dagster._config import Permissive
+from dagster._core.definitions.executor_definition import multiple_process_executor_requirements
+from dagster._core.execution.api import execute_pipeline
+from dagster._core.execution.retries import RetryMode
+from dagster._core.executor.step_delegating import (
+    CheckStepHealthResult,
+    StepDelegatingExecutor,
+    StepHandler,
+)
+from dagster._core.test_utils import instance_for_test
+from dagster._utils import merge_dicts
+
+
+class TestStepHandler(StepHandler):
+    launched_first_attempt = False
+    launched_second_attempt = False
+
+    @property
+    def name(self):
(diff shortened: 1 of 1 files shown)


johannkm:
* master
  * **PR #10458** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10458" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10458?utm_source=stack-comment).

python_modules/dagster/dagster_tests/execution_tests/engine_tests/test_step_deletegating_executor_retries.py
```diff
@@ -0,0 +1,145 @@
...
+            run_config={
+                "execution": {"config": {}},
+                "ops": {"retry_op": {"config": {"fails_before_pass": 1}}},
+            },
+        )
+        TestStepHandler.wait_for_processes()
+    assert result.success
+
+
+def test_retries_exhausted():
```
johannkm, line 127: not the best test but gives me some confidence

johannkm (commented):
The ideal will be to remove the running_steps dict and just rely on active execution. But that's a bigger refactor

python_modules/dagster/dagster_tests/execution_tests/engine_tests/test_step_deletegating_executor_retries.py
```diff
@@ -0,0 +1,145 @@
...
+        **(merge_dicts({"retries": RetryMode.ENABLED}, exc_init.executor_config)),
+        check_step_health_interval_seconds=0,
+    )
+
+
+@op(config_schema={"fails_before_pass": int})
+def retry_op(context: OpExecutionContext):
+    if context.retry_number < context.op_config["fails_before_pass"]:
+        # enough for check_step_health to be called, since we set check_step_health_interval_seconds=0
+        raise RetryRequested(seconds_to_wait=5)
```
alangenfeld, line 104: you can set `max_retries` on here and have it exhaust that way

alangenfeld (approved):
> After a step_up_for_retry can the next event be step failed? Might need to test this more

not immediate - it should `restart` before `fail` The step will fail when the limit is hit instead of going up for retry

python_modules/dagster/dagster_tests/execution_tests/engine_tests/test_step_deletegating_executor_retries.py
johannkm: true, I think the default of 1 is fine though

PR #10466, opened 2022-11-10
Make dagit selection strings invalid static partition substrings

clairelin135:
Since this is a breaking change, wanted to add it for 1.1.0. 

This will allow us to mirror partition keys into the asset event tags table.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
@@ -4,15 +4,13 @@ from typing import Dict, List, Mapping, NamedTuple, Optional, Sequence, Tuple
 
 import dagster._check as check
 from dagster._annotations import experimental
-from dagster._core.errors import DagsterInvalidDefinitionError, DagsterInvalidInvocationError
+from dagster._core.errors import DagsterInvalidInvocationError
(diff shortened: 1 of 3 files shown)


sryza:
Would we be able to instead require that users escape these characters when typing in selections?

clairelin135:
@sryza we could make users escape these characters in Dagit selection. But allowing certain characters also causes problems in storage--for example, a static partition key that is "a|1" will be equivalent to a multi-dimensional partition key "a|1". We could force users to escape these characters in their partitions definitions, but that is also a breaking change and seems gnarlier than simply disallowing them.

sryza:
> @sryza we could make users escape these characters in Dagit selection. But allowing certain characters also causes problems in storage--for example, a static partition key that is "a|1" will be equivalent to a multi-dimensional partition key "a|1". We could force users to escape these characters in their partitions definitions, but that is also a breaking change and seems gnarlier than simply disallowing them.

I thought we already forbid "|" in the case that multi-dimensional partitions definitions wrap static partitions? Is that not enough to address that issue?

clairelin135:
Yeah, I was concerned about future cases like the many -> multidimensional partitions to event case or about backcompat (e.g. going from multidimensional back to static). But I think those should be fine and we should be able to test which dimensions are populated

PR #10472, opened 2022-11-10
[job] drop default resolve config_mapping when op subsetting

alangenfeld:
### Summary & Motivation

When an explicit config dict is passed to a job, we create a preset to show the value in dagit and config_mapping that will resolve the values as defaults if nothing is passed in.

When op subelection occurs - we currently copy over the parent `config_mapping` directly, even though the schema has changed as a function of subsetting.

This change drops the default config mapping entirely when subsetting a job with explicit config. 

added test that previously failed

did subsetting in dagit - verified preset driven experience still positive

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/job_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/job_definition.py
@@ -174,6 +174,7 @@ class JobDefinition(PipelineDefinition):
         presets = []
         config_mapping = None
         partitioned_config = None
+        self._explicit_config = False
 
         if partitions_def:
             partitioned_config = PartitionedConfig.from_flexible_config(config, partitions_def)
@@ -201,6 +202,7 @@ class JobDefinition(PipelineDefinition):
                     config,
                     name,
                 )
+                self._explicit_config = True
             elif config is not None:
                 check.failed(
                     f"config param must be a ConfigMapping, a PartitionedConfig, or a dictionary, but "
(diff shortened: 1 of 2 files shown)


alangenfeld:
* master
  * **PR #10472** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10472" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10472?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -489,13 +491,20 @@ def _get_job_def_for_op_selection(
         try:
             sub_graph = get_subselected_graph_definition(self.graph, resolved_op_selection_dict)
 
+            # if explicit config was passed the config_mapping that resolves the defaults implicitly is
+            # very unlikely to work. The preset will still present the default config in dagit.
+            if self._explicit_config:
+                config_arg = None
+            else:
+                config_arg = self.config_mapping or self.partitioned_config
```
alangenfeld, line 499: trying to think of cases where people would actually get by with the existing behavior
* no dependencies in the graph such that subsetting doesn't change the schema
* ???

gibsondan:
I don't feel fully qualified to review this but it would be lovely to get in as part of 1.1

python_modules/dagster/dagster_tests/execution_tests/test_job.py
```diff
@@ -196,3 +196,35 @@ def test_job_input_values_out_of_process():
...
+    @job(config={"ops": {"emit": {"inputs": {"x": 1}}}})
+    def with_config():
+        echo(echo.alias("emit")())
+
+    result = with_config.execute_in_process()
+    assert result.success
+
+    subset_with_config = with_config.get_job_def_for_subset_selection(op_selection=["echo"])
+
+    result = subset_with_config.execute_in_process(
```
OwenKephart, line 227: Would this test have failed previously if run_config was left blank here? (my understanding is "no", but perhaps I'm missing something)

I think there is a fair scope of area where you'd be ok with the old behavior. Basically, in any case where you're passing in actual op config rather than input config, and you only have config defined for a small number of ops in your graph, most graph subselections that include all of the configured ops should work.

OwenKephart (approved):
I think this makes sense, and the fact that the default values will still get populated in Dagit means that this is almost certainly a strictly positive change.

python_modules/dagster/dagster_tests/execution_tests/test_job.py
alangenfeld: > Would this test have failed previously if run_config was left blank here?

it would fail because the subset `echo` requires input config now since its inputs have been disconnected

PR #10473, opened 2022-11-11
[dagit] Change backfill partition selection to a "range + status" style

bengotow:
### Summary & Motivation

This PR is "step 0" in moving to @braunjj's latest asset detail page designs and multidimensional partition support. We're slightly changing the way you select partition ranges in backfills:

Instead of creating a selection string containing lots of individual ranges, you select one or more "range of interest" and then use checkboxes to say "just the failed ones in this range".

This UI will be re-used on the upcoming "Asset > Partitions" page to specify a range you want to view materializations for, which is why we're starting here!

image

image

One nice side effect of this new approach is that when you choose "Re-execute from failure", we can auto-select failed runs and remove the other options:

image

x


js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
```diff
@@ -92,21 +107,36 @@ const LaunchAssetChoosePartitionsDialogBody: React.FC<Props> = ({
   upstreamAssetKeys,
 }) => {
   const partitionedAssets = assets.filter((a) => !!a.partitionDefinition);
-  const data = usePartitionHealthData(partitionedAssets.map((a) => a.assetKey));
-  const upstreamData = usePartitionHealthData(upstreamAssetKeys);
+  const assetHealth = usePartitionHealthData(partitionedAssets.map((a) => a.assetKey));
+  const upstreamAssetHealth = usePartitionHealthData(upstreamAssetKeys);
```
bengotow, line 111: Sorry for the thrashing here, I didn't like using "data" as the variable name...

js_modules/dagit/packages/core/src/assets/useRecentAssetEvents.tsx
```diff
@@ -79,41 +79,7 @@ export function useRecentAssetEvents(
...
-        }
-
-        definition {
-          id
-          partitionKeys
-        }
-      }
-    }
-  }
+export const ASSET_MATERIALIZATION_FRAGMENT = gql`
```
bengotow, line 82: Just breaking this query into sub-fragments for future use

js_modules/dagit/packages/core/src/partitions/PartitionStatus.tsx
```diff
@@ -350,7 +352,7 @@ export const PartitionStatus: React.FC<{
         <Box
           flex={{justifyContent: 'space-between'}}
           margin={{top: 4}}
-          style={{fontSize: '0.8rem', color: Colors.Gray500}}
+          style={{fontSize: '0.8rem', color: Colors.Gray500, minHeight: 17}}
```
bengotow, line 355: This prevents a slight jump when the data loads and the labels are populated

js_modules/dagit/packages/core/src/partitions/BackfillSelector.tsx
```diff
@@ -175,14 +207,15 @@ export const BackfillPartitionSelector: React.FC<{
                 checked={options.fromFailure}
                 disabled={!selected.every(isFailed)}
                 onChange={() => {
-                  if (!options.fromFailure) {
-                    setSelected(failedPartitions);
+                  const fromFailure = !options.fromFailure;
```
hellendag, line 210: This is a little confusing.

hellendag (approved):

js_modules/dagit/packages/core/src/partitions/BackfillSelector.tsx
bengotow: Good call will rephrase this 👀

prha (approved):
Fantastic

Issue #10148, opened 2022-10-21
Dagit terminate all runs backfill option doesn't work


adam-bloom:
### Dagster version

1.0.13

### What's the issue?

The backfills page presents an option to terminate all jobs for a pending or in-progress backfill. This option doesn't work.

### What did you expect to happen?

Terminate all jobs in the backfill.

### How to reproduce?

Create a large backfill (such that it will queue jobs), the backfills tab in "Overview" should show the backfill. The action menu on that backfill will show an option to terminate backfill runs - click it!

### Deployment type

Dagster Helm chart

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

PR #10475, opened 2022-11-11
fix backfill termination dialog

prha:
### Summary & Motivation
Stems from filter condition, checking for presence in a set using `in` instead of `has`.

Fixes https://github.com/dagster-io/dagster/issues/10148

https://user-images.githubusercontent.com/1040172/201246731-6e6ebcaa-f0ac-4adb-a182-0e2e4a08e337.mov

Created a backfill, canceled the backfill, terminated all the runs.

### DIFF
--- a/js_modules/dagit/packages/core/src/instance/BackfillTerminationDialog.tsx
+++ b/js_modules/dagit/packages/core/src/instance/BackfillTerminationDialog.tsx
@@ -29,7 +29,7 @@ export const BackfillTerminationDialog = ({backfill, onClose, onComplete}: Props
         partition.runStatus && partition.runId && cancelableStatuses.has(partition.runStatus),
     );
     return (
-      unfinishedPartitions?.reduce(
+      unfinishedPartitions.reduce(
         (accum, partition) =>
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/core/src/instance/BackfillTerminationDialog.tsx
```diff
@@ -20,7 +20,22 @@ export const BackfillTerminationDialog = ({backfill, onClose, onComplete}: Props
...
+  const unfinishedMap = React.useMemo(() => {
+    if (!backfill) {
+      return {};
+    }
+    const unfinishedPartitions = backfill.partitionStatuses.results.filter(
+      (partition) =>
+        partition.runStatus && partition.runId && cancelableStatuses.has(partition.runStatus),
+    );
+    return (
+      unfinishedPartitions?.reduce(
```
bengotow, line 32: If you want, I think you can remove this `?` since the line above will always return an array

bengotow (approved):
This looks good to me! Good catch on the Set vs Array issue 😬

PR #10488, opened 2022-11-11
[freshness-policy] Fix bug with latestMaterializationMinutesLate

OwenKephart:
### Summary & Motivation

Fix bug w/ how we were calculating latestMaterializationMinutesLate. To see how late you were when you emitted a given materialization, you need to see what the state of the asset was at the time of that materialization, which involves looking at the most recent materialization which happened before that.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/implementation/fetch_assets.py
+++ b/python_modules/dagster-graphql/dagster_graphql/implementation/fetch_assets.py
@@ -349,11 +349,15 @@ def get_freshness_info(
     current_time = datetime.datetime.now(tz=datetime.timezone.utc)
 
     latest_record = data_time_queryer.get_most_recent_materialization_record(asset_key=asset_key)
+    if latest_record is None:
+        return GrapheneAssetFreshnessInfo(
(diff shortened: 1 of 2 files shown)


python_modules/dagster-graphql/dagster_graphql/implementation/fetch_assets.py
```diff
@@ -368,9 +372,21 @@ def get_freshness_info(
...
+        )
+    previous_used_data_times = data_time_queryer.get_used_data_times_for_record(
+        record=previous_latest_record
+    )
+
     latest_materialization_minutes_late = freshness_policy.minutes_late(
         evaluation_time=latest_materialization_time,
-        used_data_times=used_data_times,
+        used_data_times=previous_used_data_times,
         available_data_times={key: latest_materialization_time for key in used_data_times.keys()},
```
bengotow, line 390: I'm not 100% sure what this parameter is providing but I wonder if this second reference to `used_data_times` should also be `prev_used_data_times`?

bengotow (approved):

PR #10491, opened 2022-11-11
Automatically load env vars from a local .env file when one is present during dagster local development

gibsondan:
Summary:
This makes the default dagit startup path check for a local .env file and pull secrets into it. The idea would be to make there be a dirt-simple way to get env vars into your local Dagster when getting started and doing local development.

It wouldn't work easily out of the box in docker/k8s/etc.

It would work well when doing local development with other services that export a .env file for you (like potentially Dagster Cloud!)

Will add tests and such if we think this implementation seems reasonable.

Also moved the default log level of local grpc servers to INFO so that there's some observability when it injects secrets from the file.


petehunt:
I like this since it matches what vercel/nextjs do

alangenfeld (commented):
A lot going on here.

Are we concerned at all about introducing this new default for users that have existing `.env` files that dagster will now start implicitly injecting?

gibsondan:
Moved the test changes / passing in instance ref changes into https://github.com/dagster-io/dagster/pull/10509 so that this PR could just be about the actual user-visible behavior changes / secrets bits

python_modules/dagster/dagster/_grpc/server.py
```diff
@@ -1009,16 +1009,18 @@ def wait_for_grpc_server(server_process, client, subprocess_args, timeout=60):
...
+    location_name=None,
     loadable_target_origin=None,
     max_workers=None,
     heartbeat=False,
     heartbeat_timeout=30,
     fixed_server_id=None,
     startup_timeout=20,
     cwd=None,
-    log_level="WARNING",  # don't log INFO messages for automatically spun up servers
+    log_level="INFO",
```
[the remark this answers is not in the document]
gibsondan: Old dagit startup output:
```
2022-11-14 10:47:23 -0600 - dagit - INFO - Serving dagit on http://127.0.0.1:3000 in process 21109
```

New dagit startup output:
```
2022-11-14 10:48:00 -0600 - dagster - INFO - Started Dagster code server for module dagster_test.toys.repo in process 21250
2022-11-14 10:48:02 -0600 - dagit - INFO - Serving dagit on http://127.0.0.1:3000 in process 21236
```

Seems fine / actually better to me?

gibsondan (commented):
"Are we concerned at all about introducing this new default for users that have existing .env files that dagster will now start implicitly injecting?"

Seems worth a Breaking Changes entry in 1.1 this week, but not overly concerned beyond that? It'd have to be in the same folder that you're running dagit from which is typically a Dagster-specific folder anyway

python_modules/dagster/dagster/_core/instance/ref.py
```diff
@@ -279,6 +279,9 @@ def config_defaults(base_dir):
                 "DefaultRunLauncher",
                 yaml.dump({}),
             ),
+            # For back-compat, the default is actually set in the secrets_loader property above,
+            # so that old clients loading new config don't try to load a class that they
+            # don't recognize
```
alangenfeld, line 284: when exactly do we hit this condition of old trying to load via this config? Is it just the release or 2 that we had the `secrets_loader_data` field on `InstanceRef` but the `EnvFileLoader` doesn't exist?

gibsondan: Yeah this is unfortunate planning on my part. If I had introduced this last week this in the same release that we added this secrets_loader field (like the original version of my PR even did!) then this wouldn't be a problem

alangenfeld: can we add a test for this specific thing? just to encode this known issue in the code base. Truly emulating in the test would be a lot but poking at a serialized instance ref should be fine. 

maybe update this comment with the specific problem versions as well .

alangenfeld:
> Seems worth a Breaking Changes entry in 1.1 this week, but not overly concerned beyond that? It'd have to be in the same folder that you're running dagit from which is typically a Dagster-specific folder anyway

sounds right to me - just wanted to think through it. 

Should we wait to hear from @yuhan or anyone else before proceeding? will leave it open to keep it in peoples queues for now.

python_modules/dagster/dagster/_core/instance/ref.py
```diff
@@ -463,7 +463,18 @@ def run_launcher(self):
...
+        # EnvFileLoader was not defined
+        return (
+            self.secrets_loader_data.rehydrate()
+            if self.secrets_loader_data
+            else ConfigurableClassData(
+                "dagster._core.secrets.env_file",
+                "EnvFileLoader",
+                yaml.dump({}),
+            ).rehydrate()
+        )
```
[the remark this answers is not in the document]
yuhan: should it be `env_loader` or `env_var_loader`?

python_modules/dagster/dagster/_core/secrets/env_file.py
```diff
@@ -0,0 +1,58 @@
...
+            key: check.not_none(val) for key, val in dotenv_values(file_path).items() if val != None
+        }
+
+        if len(env_var_dict):
+            logging.getLogger("dagster").info(
+                "Loaded environment variables from .env file: "
+                + ",".join([env_var for env_var in env_var_dict]),
+            )
+        else:
+            logging.getLogger("dagster").info("No environment variables in .env file")
```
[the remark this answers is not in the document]
petehunt: can we only show this message if a `.env` file is present?

python_modules/dagster/dagster/_core/secrets/env_file.py
```diff
@@ -0,0 +1,58 @@
...
+
+        return env_var_dict
+
+    def get_secrets_for_environment(self, location_name: Optional[str]) -> Dict[str, str]:
+        env_vars: Dict[str, str] = {}
+
+        env_file_path = os.path.join(self._base_dir, ".env")
+
+        if os.path.exists(env_file_path):
```
gibsondan, line 43: @petehunt we can and we do!

yuhan: oh i see! so it's not that easy to discover this feature from loading the code right? 
i think it's totally fine and in fact great as we're still early, but eventually, i'd like to hint the existence of this behavior by default (once we have the docs ready to link)

python_modules/dagster/dagster/_core/instance/ref.py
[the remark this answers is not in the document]
gibsondan: That ship has sailed for better or for worse - left it SecretsLoader to in the future potentially support loading a specific secret instead of auto-injecting it into the environment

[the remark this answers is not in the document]
yuhan: got it. makes sense.

python_modules/dagster/dagster/_core/secrets/env_file.py
```diff
@@ -0,0 +1,57 @@
...
+from dotenv import dotenv_values
+
+import dagster._check as check
+from dagster._config import Field, StringSource
+from dagster._serdes import ConfigurableClass
+
+from .loader import SecretsLoader
+
+
+class EnvFileLoader(SecretsLoader, ConfigurableClass):
```
jmsanders, line 14: I'm curious how this will interact with other popular env loaders. For example, when you cd into a directory with a .env file while using ZSH's dotenv plugin, you get prompted with:

```
dotenv: found '.env' file. Source it? ([Y]es/[n]o/[a]lways/n[e]ver)
```

Will we respect those decisions? If I told my shell not to source my .env file, I'd be surprised if dagster automatically sourced it anyway.

Should we prompt before loading? Should we include CLI flags to disable loading?

gibsondan: "If I told my shell not to source my .env file, I'd be surprised if dagster automatically sourced it anyway." - i'm not sure i agree with this, if you have a .env file in your dagster folder, it seems plausible to me that dagster would use it.

I'm not totally opposed to adding a flag to disable it but it seems like the kind of thing i'd wait for somebody to ask for - not hard to change it later if it comes up.

python_modules/dagster/dagster/_core/instance/ref.py
alangenfeld: which back-compat test fails? Just `N-1` release? 

i was thinking just verifying that the serialized instance ref doesn't contain `EnvFileLoader`

gibsondan:
Those are included in https://github.com/dagster-io/dagster/pull/10486 (alongside some other secrets-related changes)

alangenfeld (approved):
alright alright alright

PR #10492, opened 2022-11-11
docs: add test to verify repository loads code definitions

rexledesma:
### Summary & Motivation
This is a common pattern to verify that user defined code definitions are well formed.

Let's put it in the documentation.

vercel


rexledesma:
@erinkcochran87 will take this over!

PR #10495, opened 2022-11-11
[dagit] Asset catalog list UI for asset freshness

bengotow:
### Summary & Motivation

This PR adds asset freshness UI to the asset catalog! Here's a quick visual walk through of the changes. There is more work to do, specifically showing this info on the DAG itself and adjusting the appearance in the asset catalog, but I think it'll make more sense to do a final pass after asset versioning has landed and we can look at everything together.

cc @OwenKephart this no longer uses `latestMaterializationMinutesLate` so we can remove that if we want!

This PR doesn't change the UI for non-freshness-policy assets, EXCEPT that run status dots are now shown in the DAG and asset catalog per the latest mockups.

image
image
image
image

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
@@ -1,6 +1,17 @@
 import {gql} from '@apollo/client';
-import {Colors, Icon, Tooltip, FontFamily, Box, CaptionMono, Spinner} from '@dagster-io/ui';
+import {
+  Colors,
+  Icon,
+  Tooltip,
+  FontFamily,
+  Box,
+  CaptionMono,
+  Spinner,
+  Caption,
+  Tag,
+} from '@dagster-io/ui';
 import isEqual from 'lodash/isEqual';
+import moment from 'moment-timezone';
 import React from 'react';
 import {Link} from 'react-router-dom';
 import styled from 'styled-components/macro';
@@ -8,20 +19,26 @@ import styled from 'styled-components/macro';
 import {withMiddleTruncation} from '../app/Util';
 import {NodeHighlightColors} from '../graph/OpNode';
 import {OpTags} from '../graph/OpTags';
+import {RunStatusIndicator} from '../runs/RunStatusDots';
 import {linkToRunEvent, titleForRun} from '../runs/RunUtils';
 import {TimestampDisplay} from '../schedules/TimestampDisplay';
+import {humanCronString} from '../schedules/humanCronString';
 import {AssetComputeStatus} from '../types/globalTypes';
 import {markdownToPlaintext} from '../ui/markdownToPlaintext';
 
 import {LiveDataForNode} from './Utils';
 import {ASSET_NODE_ANNOTATIONS_MAX_WIDTH, ASSET_NODE_NAME_MAX_LENGTH} from './layout';
+import {AssetGraphLiveQuery_assetNodes_freshnessPolicy} from './types/AssetGraphLiveQuery';
 import {AssetNodeFragment} from './types/AssetNodeFragment';
 
-const MISSING_LIVE_DATA = {
+const MISSING_LIVE_DATA: LiveDataForNode = {
   unstartedRunIds: [],
   inProgressRunIds: [],
   runWhichFailedToMaterialize: null,
+  freshnessInfo: null,
+  freshnessPolicy: null,
   lastMaterialization: null,
+  lastMaterializationRunStatus: null,
   computeStatus: AssetComputeStatus.NONE,
   stepKey: '',
 };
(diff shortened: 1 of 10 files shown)


sryza:
Quick comment: "Materialized within 2 min" without other context sounds like we're saying that the asset was materialized within the last 2 minutes or something. I think we need to somehow communicate that this is a freshness policy that's set on the asset, not a description of what has occurred recently.

bengotow:
Hey @sryza! I talked with Josh and he was confused about the same thing, what do you think of the updated screenshots?

- We don't show the policy description in the asset catalog view except on hover
- We show it in the Definition tab and in the asset graph sidebar printed out in more detail

Hovering over "Fresh" now says:

image

Hovering over "Late" now says:
(maximally complicated example)

image

js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
```diff
@@ -318,46 +345,111 @@ export const ComputeStatusNotice: React.FC<{computeStatus: AssetComputeStatus}>
...
+
+export const CurrentMinutesLateTag: React.FC<{
+  liveData: LiveDataForNode;
+  policyOnHover?: boolean;
+}> = ({liveData, policyOnHover}) => {
+  const {freshnessInfo, freshnessPolicy} = liveData;
+  const description = policyOnHover ? freshnessPolicyDescription(freshnessPolicy) : '';
+
+  return freshnessInfo?.currentMinutesLate ? (
+    <Tooltip content={<div style={{maxWidth: 400}}>{`${OVERDUE_MESSAGE} ${description}`}</div>}>
```
OwenKephart, line 429: One thing to note is that if the currentMinutesLate is null, then this indicates that either this asset was never materialized, which might not make sense to indicate as "fresh"

OwenKephart:
Riffing a bit on the copy, maybe something along the lines of:

'<cron schedule or "at any point in time">, this asset should incorporate data up to <maximum_lag_minutes> before that time.'

cron_schedule defined: 
"By 9 AM every day, this asset should incorporate all data up to 9 hours before that time"
"By 30 minutes past the hour, this asset should incorporate all upstream data up to 15 minutes before that time"

no cron_schedule defined:
"At any point in time, this asset should incorporate all data up to 1 hour before that time"

When you're out of date, maybe something like "asset currently incorporates all data up to <...>, expected <expected data time>". This is something that a future gantt chart viewer would help elucidate for the user (as it's currently pretty hard for someone to figure out why they're out of date).

js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
bengotow: Ahh good catch, I can make it only show when there's a last materialization

sryza:
> Hey @sryza! I talked with Josh and he was confused about the same thing, what do you think of the updated screenshots?

This looks good.

bengotow:
Hey Owen, I updated the copy in this PR and I think this feels great:

image

image

image

I think we need a bit more info from GraphQL in order to do the `asset currently incorporates all data up to <...>`. We know how many minutes late it is, but we can't easily evaluate the cron string to the current target time...

js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
```diff
@@ -318,46 +345,126 @@ export const ComputeStatusNotice: React.FC<{computeStatus: AssetComputeStatus}>
...
+          <RunStatusIndicator status={lastMaterializationRunStatus} size={10} />
+        ) : (
+          <div style={{width: 10}} />
+        )}
+        <AssetRunLink
+          runId={lastMaterialization.runId}
+          event={{stepKey, timestamp: lastMaterialization.timestamp}}
+        />
+      </Box>
+    ) : undefined;
```
hellendag, line 397: I may have commented on this elsewhere, but I find these nested ternaries hard to follow. :(

js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
```diff
@@ -318,46 +345,126 @@ export const ComputeStatusNotice: React.FC<{computeStatus: AssetComputeStatus}>
...
+    minutes > 60 * 24 * 60
+      ? [Math.ceil(minutes / (30 * 24 * 60)), 'month']
+      : minutes > 3 * 24 * 60
+      ? [Math.ceil(minutes / (24 * 60)), 'day']
+      : minutes > 60
+      ? [Math.ceil(minutes / 60), 'hour']
+      : [Math.ceil(minutes), 'min'];
+
+  return val !== 1 && unit !== 'min' ? `${val} ${unit}s` : `${val} ${unit}`;
+};
```
hellendag, line 422: Can this use `moment` or something so that we don't need to implement it ourselves?

hellendag (approved):

js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
bengotow: Ahh let me see! There are a couple caveats to this one (showing up to 3 days as hours) but we can do away with that and use moment i think

js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
bengotow: Ahh will change this @hellendag! I need to find another pattern for doing this sort of thing, I guess a sub-render function with a series of if/returns? I think this one could be a whole sub-component maybe but it might take a bunch of props.

PR #10497, opened 2022-11-11
AssetsDefinition.configured (nonurgent)

clairelin135:
Updates `AssetsDefinition` to be configurable via `.configured`.

The intention is for an asset to be reusable multiple times within the same repository, e.g.:
```
@asset(config_schema={"hyperparameter": float})
def my_model(context):
    return context.op_config["hyperparameter"]

@repository
def my_repo():
    return [
        hyperparam_model,
        hyperparam_model.configured({"hyperparameter": 0.5}, "model_05")   # has asset key AssetKey("model_05")
        hyperparam_model.configured({"hyperparameter": 0.5}, "model_075")  # has asset key AssetKey("model_075")
   ]
```

This PR does a couple things:
- For assets generated with `@asset`:
    -  Accepts a new name. Since these assets use the name to generate the asset key, duplicates the op with the new name, populates config, and replaces the asset key with an asset key constructed from the new name
 - For `@multi_assets`:
    - Duplicates the inner op with the new name and populates config. If users want to change the output asset keys, they can override the keys with the `.with_prefix_or_group_name` method (which we should probably rename)
 - For graph-backed assets:
    - `AssetsDefinition.from_graph` does not accept config. So this method throws an error, instead asking users to do something like: `AssetsDefinition.from_graph(your_graph.configured(...))`

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/assets.py
@@ -2,6 +2,7 @@ import warnings
 from typing import (
     TYPE_CHECKING,
     AbstractSet,
+    Any,
     Dict,
     Iterable,
     Iterator,
@@ -17,8 +18,10 @@ import dagster._check as check
 from dagster._annotations import public
 from dagster._core.decorator_utils import get_function_params
 from dagster._core.definitions.asset_layer import get_dep_node_handles_of_graph_backed_asset
+from dagster._core.definitions.configurable import NamedConfigurableDefinition
 from dagster._core.definitions.events import AssetKey
 from dagster._core.definitions.freshness_policy import FreshnessPolicy
+from dagster._core.definitions.graph_definition import GraphDefinition
 from dagster._core.definitions.metadata import MetadataUserInput
 from dagster._core.definitions.partition import PartitionsDefinition
 from dagster._core.definitions.utils import DEFAULT_GROUP_NAME, validate_group_name
@@ -30,6 +33,7 @@ from dagster._utils.backcompat import (
     experimental_arg_warning,
 )
 
+from .definition_config_schema import IDefinitionConfigSchema
 from .dependency import NodeHandle
 from .events import AssetKey, CoercibleToAssetKeyPrefix
 from .node_definition import NodeDefinition
@@ -52,7 +56,7 @@ if TYPE_CHECKING:
     from .graph_definition import GraphDefinition
 
 
-class AssetsDefinition(ResourceAddable):
+class AssetsDefinition(ResourceAddable, NamedConfigurableDefinition):
     """
     Defines a set of assets that are produced by the same op or graph.
 
@@ -534,6 +538,10 @@ class AssetsDefinition(ResourceAddable):
     def metadata_by_key(self):
         return self._metadata_by_key
 
+    @property
+    def config_schema(self) -> Optional[IDefinitionConfigSchema]:
+        return self.node_def.config_schema
+
     @public
     def get_partition_mapping(self, in_asset_key: AssetKey) -> PartitionMapping:
         with warnings.catch_warnings():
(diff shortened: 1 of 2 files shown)


clairelin135:
In response to issue https://github.com/dagster-io/dagster/issues/9359

python_modules/dagster/dagster/_core/definitions/assets.py
```diff
@@ -855,6 +862,50 @@ def with_resources(self, resource_defs: Mapping[str, ResourceDefinition]) -> "As
...
+            partition_mappings=self._partition_mappings,
+            asset_deps=self._asset_deps,
+            selected_asset_keys=self._selected_asset_keys,
+            can_subset=self._can_subset,
+            resource_defs=self._resource_defs,
+            group_names_by_key=self.group_names_by_key,
+            metadata_by_key=self._metadata_by_key,
+            freshness_policies_by_key=self.freshness_policies_by_key,
+        )
+        if len(configured_asset.keys) == 1 and self.node_def.name.endswith(
```
sryza, line 894: Will we silently ignore the name parameter in situations where this doesn't hold true?

clairelin135: no, the name of the underlying op/graph within the asset will have the new name. We just won't replace the output asset keys since it's unclear what the new asset keys should be, or if they should be changed at all

dpeng817, line 894: does `self.node_def.name.endswith(configured_asset.key.path[-1])` always guarantee equality? It seems like there's some (admittedly pathological) cases where this might let through assets that were created by using `AssetsDefinition` manually rather than using the `@asset` decorator.

python_modules/dagster/dagster_tests/core_tests/test_configured.py
```diff
@@ -35,3 +55,156 @@ def mypipeline():
...
+        result.asset_materializations_for_node("prefix__my_asset")
+    ) == [AssetKey(["prefix", "my_asset"])]
+    assert get_asset_keys_from_materializations(
+        result.asset_materializations_for_node("configured_hello")
+    ) == [AssetKey(["prefix", "configured_hello"])]
+    assert get_asset_keys_from_materializations(
+        result.asset_materializations_for_node("configured_hi")
+    ) == [AssetKey(["prefix", "configured_hi"])]
+
+
```
dpeng817, line 101: what happens when an asset is constructed without a decorator (assuming that is a pattern we support)? As mentioned in another comment, I think you can construct some pathological cases here where it's not sufficient to assume that the name / asset key are equivalent.

dpeng817 (changes requested):
I'm wary about the `multi_asset` bit here - it seems that if a user calls configured on a multi_asset, they literally always need to call `with_prefix_or_group_name` in order to not create a footgun, where two assets are producing the same keys. I think we should strongly consider folding the output key overridding behavior into configured directly here, and force its usage for multi_assets. Requesting changes to spur discussion / move from queue.

sryza:
I share @dpeng817 's point that we should think deeply about the right behavior here for overriding asset keys. IMO it's undesirable that the asset keys get overridden when you have a single-asset but not a multi-asset. What about requiring that the user supplies overrides all the asset keys?

A couple additional thoughts:
- Almost nobody uses config-mapping and we could probably get away without supporting it if that makes things easier.
- I suspect the main use case for configuring assets at definition time does _not_ involve the same asset getting configured multiple times. People just want to be able to specify some of their asset's behavior in YAML and have it represented in the UI. So I think it's burdensome to require them to specify a new name every time they do this.

Here's an example user request: https://dagster.slack.com/archives/C01U5LFUZJS/p1670515667480569?thread_ts=1670403685.794129&cid=C01U5LFUZJS

> I remember a few months back when we floated the idea of a .configured method attached to AssetsDefinitions. This would definitely solve my use case, although for how I’ve developed thus far, I haven’t needed to keep different configurations of the same asset at the same time, so even just having a config flag on the decorators or .from_graph method (analog to @job(config={}) would already be enough.
> For more context, all the assets I’ve written thus far take some form of mandatory configuration. From fairly basic like the example with the compounds, to more complex in case they rely on the same underlying API and need to hit different endpoints and process the data in different ways.

Maybe we should just add a `config` parameter to `@asset` etc.?

cmpadden:
Hey @clairelin135 - slowly reviewing our backlog of pull requests.

Is this change still relevant, or can it be closed at this point? If it is, let me know, and I can assist in resolving conflicts.

PR #10499, opened 2022-11-12
[freshness-policy] assets_to_reconcile_for_freshness

OwenKephart:
### Summary & Motivation

Adds the ability for the reconciliation sensor to kick off runs to satisfy constraints derived from freshness policies defined on assets in the repository. If the reconcilliation planned on propagating a change to downstream assets, but a freshness policy plans on propogating that change at some point within the next 12 hours, we defer propogating that constraint to minimize unnecessary work.

Also combines the functionality of TickInstanceQueryer into DataTimeInstanceQueryer to keep things simpler

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -25,9 +25,9 @@ import pendulum
 import dagster._check as check
 from dagster._annotations import experimental
 from dagster._core.definitions.events import AssetKey, AssetKeyPartitionKey
-from dagster._core.definitions.freshness_policy import FreshnessConstraint, FreshnessPolicy
+from dagster._core.definitions.freshness_policy import FreshnessConstraint
 from dagster._core.storage.tags import PARTITION_NAME_TAG
-from dagster._utils.calculate_data_time import DataTimeInstanceQueryer
+from dagster._utils.caching_instance_queryer import CachingInstanceQueryer
 
 from .asset_selection import AssetGraph, AssetSelection
 from .partition import PartitionsDefinition, PartitionsSubset
@@ -217,7 +217,7 @@ class ToposortedPriorityQueue:
 
 
 def find_stale_candidates(
-    instance_queryer: DataTimeInstanceQueryer,
+    instance_queryer: CachingInstanceQueryer,
     cursor: AssetReconciliationCursor,
     target_asset_selection: AssetSelection,
     asset_graph: AssetGraph,
@@ -261,7 +261,7 @@ def find_stale_candidates(
 
 
 def find_never_materialized_or_requested_root_asset_partitions(
-    instance_queryer: DataTimeInstanceQueryer,
+    instance_queryer: CachingInstanceQueryer,
     cursor: AssetReconciliationCursor,
     target_asset_selection: AssetSelection,
     asset_graph: AssetGraph,
@@ -308,10 +308,11 @@ def find_never_materialized_or_requested_root_asset_partitions(
 
 
 def determine_asset_partitions_to_reconcile(
-    instance_queryer: DataTimeInstanceQueryer,
+    instance_queryer: CachingInstanceQueryer,
     cursor: AssetReconciliationCursor,
     target_asset_selection: AssetSelection,
     asset_graph: AssetGraph,
+    eventual_asset_partitions_to_reconcile_for_freshness: AbstractSet[AssetKeyPartitionKey],
 ) -> Tuple[
     AbstractSet[AssetKeyPartitionKey],
     AbstractSet[AssetKey],
@@ -345,6 +346,10 @@ def determine_asset_partitions_to_reconcile(
 
     while len(candidates_queue) > 0:
         candidate = candidates_queue.dequeue()
+
+        # no need to update this now, as it will be updated later
+        if candidate in eventual_asset_partitions_to_reconcile_for_freshness:
+            continue
 
         if (
             # all of its parents reconciled first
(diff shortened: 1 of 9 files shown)


OwenKephart:
need to fix some pendulum-version specific stuff

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -554,7 +601,12 @@ def reconcile(
         Tuple[Optional[PartitionsDefinition], Optional[str]], Set[AssetKey]
     ] = defaultdict(set)
 
-    for asset_partition in asset_partitions_to_reconcile:
+    for asset_partition in (
+        asset_partitions_to_reconcile | asset_partitions_to_reconcile_for_freshness
+    ):
+        # defer materializing the asset now if it will be materialized later
+        if asset_partition in eventual_asset_partitions_to_reconcile_for_freshness:
```
sryza, line 608: Imagine we have 

```
       /-> C
A -> B
       \-> D
```

C has a freshness policy and D does not. `asset_partitions_to_reconcile` includes B, C, and D, and `eventual_asset_partitions_to_reconcile_for_freshness ` includes B and C.  So we defer B and C, but we still execute D. Which is problematic, because D shouldn't execute until B executes.

I wonder if the freshness algorithm should be influencing what shows up in `to_reconcile` / `candidates_queue` in `determine_asset_partitions_to_reconcile`?

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -528,16 +393,198 @@ def determine_asset_partitions_to_reconcile(
     )
 
 
+def determine_asset_partitions_to_reconcile_for_freshness(
+    instance_queryer: DataTimeInstanceQueryer,
+    asset_graph: AssetGraph,
+    freshness_policies_by_key: Mapping[AssetKey, FreshnessPolicy],
```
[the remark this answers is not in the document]
OwenKephart: yeah this is probably for the best (just felt a little off to keep adding more asset attributes to this thing)

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
OwenKephart: good catch, I *think* this can be solved inside of determine_asset_partitions_to_reconcile. If we add an additional condition to allow an assetkeypartitionkey to be added to the `to_reconcile` set, which is that it cannot be part of the eventual_asset_partitions_to_reconcile_for_freshness set, then this should stop changes from propagating downstream (i.e. B would not be added to the set, so D would not have all of its parents in to_reconcile, so it would not be added to the set either)

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -0,0 +1,340 @@
...
+from dagster._core.storage.tags import PARTITION_NAME_TAG
+from dagster._utils.cached_method import cached_method
+from dagster._utils.merger import merge_dicts
+
+if TYPE_CHECKING:
+    from dagster import DagsterInstance
+
+
+class CachingInstanceQueryer:
+    """Allows caching queries to the instance while calculating data times."""
```
sryza, line 25: This docstring should be updated, right?

python_modules/dagster/dagster_tests/core_tests/definitions_tests/test_asset_reconciliation_sensor.py
```diff
@@ -529,10 +564,88 @@ def get_downstream_partitions_for_partition_range(
...
+        expected_run_requests=[],
+    ),
+    "freshness_overlapping_runs_half_stale": AssetReconciliationScenario(
+        assets=overlapping_freshness,
+        unevaluated_runs=[run(["asset1", "asset3", "asset5"]), run(["asset2", "asset4", "asset6"])],
+        # evaluate 35 minutes later, only need to refresh the assets on the shorter freshness policy
+        evaluation_delta=datetime.timedelta(minutes=35),
+        expected_run_requests=[run_request(asset_keys=["asset1", "asset3", "asset5"])],
+    ),
+    "freshness_overlapping_defer_propogate": AssetReconciliationScenario(
```
sryza, line 623:
```suggestion
    "freshness_overlapping_defer_propagate": AssetReconciliationScenario(
```

python_modules/dagster/dagster_tests/core_tests/definitions_tests/test_asset_reconciliation_sensor.py
```diff
@@ -529,10 +564,88 @@ def get_downstream_partitions_for_partition_range(
...
+            assets=overlapping_freshness_inf,
+            unevaluated_runs=[run(["asset1", "asset2", "asset3", "asset4", "asset5", "asset6"])],
+        ),
+        # change at the top, doesn't need to be propogated to 1, 3, 5 as freshness policy will
+        # handle it, but assets 2, 4, 6 will not recieve an update in the plan window. 2 can be
+        # updated immediately, but 4 and 6 depend on 3, so will be defered
+        unevaluated_runs=[run(["asset1"])],
+        expected_run_requests=[run_request(asset_keys=["asset2"])],
+    ),
+    "freshness_overlapping_defer_propogate2": AssetReconciliationScenario(
```
sryza, line 635:
```suggestion
    "freshness_overlapping_defer_propagate2": AssetReconciliationScenario(
```

sryza (approved):
Bam!

I love the way the algorithm is factored out- very easy to consume.

PR #10504, opened 2022-11-13
RFC: definitions replaces repository

schrockn:
### Summary & Motivation

As we all know, one of the biggest problems with dagster is the redundant, overlapping, and confusing concepts that makes the spin up more difficult and also incurs cognitive load even to existing users. While there are few different parts where this is true, one of the more painful parts is the workspace --> code location --> repository ontology. Also mix in the somewhat-but-not-really-orthogonal instance and deployment concepts, and you have quite the confusing stew.

This PR proposes adding a new entrypoint function `definitions` that would be used instead of `repository` going forward. It is typed and enforces that there is only one repository per python process, which enables the collapsing of code location and repository into a single concept for tooling purposes.

This injects a repository definition as an attribute into the entry point module using python internals, so all auto-discovery features and entry points should work as is.

**Resources become a top-level argument**

 In the short-term, this means improved ergonomics, as `with_resources` is automatically applied to included assets. I advocated for `with_resources` because I felt it was more composable. In retrospect I've come to believe that this was a mistake, complicating the common case for advanced cases that rarely, if ever, occur. If we feel the need to, we can introduce an advanced API later that allows the user to manage their own resource binding.

This also sets up a potential workstream of making resources more first-class citizens in operational tools as well as more ergonomic, typed APIs that pass resource instances directly to `asset` and similar functions rather than requiring contexts and magic strings keys. This is fairly speculative so I won't comment further here. Even if we do not move forward with what this paragraph alludes to, the ergonomic wins are sufficient enough to move forward with this.

**Resources in this API can be raw objects**

 This is a nice convenience and allows the developer to incrementally add resources without requiring all of that sugar

**A typed API make the system more documented and understandable**

At this point `repository` is a pretty rough API. The docs are outdated (they don't even acknowledge the existence of assets see https://github.com/dagster-io/dagster/issues/10505) in order to find what it actually accepts you find quite a rat's nest of deprecated concepts it accepts. Types and argument self-document to some degree, stay up to date, and improve discoverability.

**Non-invasive goodbye to repository**

This change can make repository "disappear" for new users.  The mental model here is that a call to `definitions` defines all the definitions for a _code location_, not a repository. This PR currently enforces this at the module level. We could make it more strict.

For existing users this doesn't really change much either. They will simply think "this is how I define the things in my repo". The function name `definitions` allows for both interpretations, which is why it was chosen.

If we move forward with this, we can mark code locations that only have a single repository and that were constructed that way as special, and proceed to fork our UI to make the entire concept of repository disappear. Only code location would remain.

For existing users that do have multiple repositories in a single code location, they would instead have to multiple code locations that happen to share the same python dependencies.  

* Included unit tests
* Ported assets_dbt_python to `definitions`


python_modules/dagster/dagster/_core/definitions/definitions_fn.py
```diff
@@ -0,0 +1,124 @@
...
+class DefinitionsAlreadyCalledError(Exception):
+    pass
+
+
+def get_dagster_definitions_in_module(mod: ModuleType):
+    return mod.__dict__[MAGIC_REPO_GLOBAL_KEY]
+
+
+# TODO: Add a new Definitions class to wrap RepositoryDefinition?
+def definitions(
```
petehunt, line 69: do we want to consider adding some sugar for `load_all_assets_from_package_name`? i.e. an arg `asset_package_names: List[str]`?

sryza:
I find this approach pretty appealing. The module-level function call is clever - aside from the benefits of making it singular per module and having clear arguments for the different definition types, I think getting rid of the decorator will reduce confusion for a lot of people who find decorators confusing.  The name "definitions" had actually also briefly occurred to me recently and very nicely sidesteps needing to add a new noun.

My main questions:
- How do we support users that want to supply different resources / config in different deployments / environments?  I.e. how would we migrate project_fully_featured? I think expecting a module per environment, where they would need to repeat the `assets=`, `jobs=`, etc. would be a poor experience.
- What happens to `RepositoryDefinition.load_asset_value`? ([docs](https://docs.dagster.io/concepts/assets/software-defined-assets#loading-asset-values-outside-of-dagster-runs)).  I had also been planning to add  https://github.com/dagster-io/dagster/pull/10072.
- What do we do about all the educational materials that we can't codemod? Sample here: https://www.notion.so/dagster/The-pain-of-replacing-repository-c80d89523fb44bcd926bc83aa611c92d. Many new users are likely to encounter these.
- If we decide to go this route, would we want to try to get it into 1.1?

elementljarredc:
should the 'definition' label be attached to some kind of noun? I think elsewhere we the concept of resourcedefinition, opdefinition, etc, do we need to attach something here to 
1. follow the previously established syntax
2. make sure it stays disambiguated from other uses of definition?

image

python_modules/dagster/dagster/_core/definitions/definitions_fn.py
schrockn: I believe we should leave that to user to compose in this API. Otherwise we will end up supporting a large surface area of potential ways to support asset loading in this top-level API.

schrockn:
* _How do we support users that want to supply different resources / config in different deployments / environments? I.e. how would we migrate project_fully_featured? I think expecting a module per environment, where they would need to repeat the assets=, jobs=, etc. would be a poor experience._

Ah so I should have made this clearer. In the final version of this, I believe this should be global to a code location. I can update the diff to make the code reflect that. The module variable trick exists solely so that we do not have to change all the code location pointer machine and to make all existing dagit and dagster CLI entry points work.

* _Would the provided resources apply to jobs eventually?_

Yes I think that is most natural thing to do. To sensors and schedules as well, once they support resources.

* _What happens to RepositoryDefinition.load_asset_value? ([docs](https://docs.dagster.io/concepts/assets/software-defined-assets#loading-asset-values-outside-of-dagster-runs)). I had also been planning to add https://github.com/dagster-io/dagster/pull/10072._

I think we should add a `get_dagster_context` (naming up for debate) that implements that function.

* _What do we do about all the educational materials that we can't codemod? Sample here: https://www.notion.so/dagster/The-pain-of-replacing-repository-c80d89523fb44bcd926bc83aa611c92d. Many new users are likely to encounter these. If we decide to go this route, would we want to try to get it into 1.1?_

Yes. I think it is work but tractable. It also depends on exact release timing.

sryza:
> Ah so I should have made this clearer. In the final version of this, I believe this should be global to a code location. I can update the diff to make the code reflect that. The module variable trick exists solely so that we do not have to change all the code location pointer machine and to make all existing dagit and dagster CLI entry points work.

@schrockn I don't entirely follow how this addresses my question. I'm imagining that multiple deployments might want to reference the same set of definitions (dev/staging/prod), but with different resources / config.

schrockn:
@sryza ah yes totally missed that I didn't answer the core of your question, rather than one of the assumptions in it.

* _How do we support users that want to supply different resources / config in different deployments / environments? I.e. how would we migrate project_fully_featured? I think expecting a module per environment, where they would need to repeat the assets=, jobs=, etc. would be a poor experience._


Was:

```
@repository
def hacker_news_repository():
    deployment_name = os.environ.get("DAGSTER_DEPLOYMENT", "local")
    resource_defs = resource_defs_by_deployment_name[deployment_name]

    definitions = [
        with_resources(all_assets, resource_defs),
        *all_jobs,
    ]
    if deployment_name in ["prod", "staging"]:
        definitions.append(make_slack_on_failure_sensor(base_url="my_dagit_url"))

    return definitions
```

Now:

```
deployment_name = os.environ.get("DAGSTER_DEPLOYMENT", "local")
resource_defs = resource_defs_by_deployment_name[deployment_name]
sensors = (
    [make_slack_on_failure_sensor(base_url="my_dagit_url")]
    if deployment_name in ["prod", "staging"]
    else []
)

definitions(assets=all_assets, jobs=all_jobs, sensors=sensors, resources=resource_defs)
```

Maybe I am missing something?

yuhan (commented):
(I haven't looked at feedback from others - so some may have already been covered by others - but i wanted to call them out so that we know which ones we feel the same about)

Things i like:
- Module level import is neat. I feel it could address the confusion around "when should i put things in __init__ and when not".
- Resources at the top-level. I found myself wanting top-level resource a lot more often as I'm writing analytical pipelines that share resources, esp when we are using a db as the single source of truth I feel top-level resources can save a lot of typing and make collab easier because configs can be shared by default.
- Typed API is amazing.

Things i have concerns about:
- Production education concerns. Would like to get  @slopp's to weight in.
  * Calling `definitions(..)` in `__init__.py` will need education. For example, just by looking at the code, I'm not sure when the definitions will be loaded.
  * Although I personally like the module-level function call, it may cause onboarding issue for users who are not very familiar with python, as in they may not understand the ` __init__.py` files and ignore it entirely.

Questions
- How would one do `RepositoryDefinition.load_asset_value` in the new world?
- v1.1 or any minor versions (by minor i mean 1.x) feels a right time to introduce changes like this. should we get this in v1.1 so we can get feedback earlier?
- Similar question^ how do we plan to roll this out? Although it's a non-invasive change, it'd involve a lot of docs/examples changes, which I'd also consider as breaking changes to content.

sryza:
@schrockn ah, makes sense.

slopp:
@yuhan I would think `definitions` would be similar from a user's perspective to `@repository`  in the sense that it could live anywhere, e.g. in a `definitions.py` file. It is up to us to determine the right locale for the occasion - e.g. a quick start that we want to be really simple would prefer `definitions` to be outside of `__init__.py`. While dagster projects are often python packages, they still shouldn't need to be

That said, I don't actually know if ^ is true? I'm not deep enough to know the implications of definitions being a function instead of a decorated function

schrockn:
We've reached consensus to move forward with this. I'm going to start a separate, clean stack of PRs to drive home implementation.

schrockn:
Closing in favor of https://github.com/dagster-io/dagster/pull/10526

PR #10509, opened 2022-11-14
Pass an instance ref and location into local grpc servers

gibsondan:
Summary:
This is to lay groundwork for https://github.com/dagster-io/dagster/pull/10491/files - fixes up a whole bunch of tests that were assuming that you could run a grpc server using an in memory instance (including a couple of whole graphql test suites), and pass in the local instance ref into the grpc server constructor. This isn't something that a user has actually been able to do for a while.

Test Plan: BK


alangenfeld (commented):
something about always requiring the instance ref makes me feel uneasy but i can't point to any specific problem that are that concerning. Theres like the theoretical case of loading different python environments with different dagster versions but thats very rare - and only a set-up for potential edge cases as opposed to a problem itself.

alangenfeld (approved):
👻

PR #10513, opened 2022-11-14
[dagit] Tweak GraphQL so single- and mulit- dim assets share resolvers

bengotow:
### Summary & Motivation

This PR makes a few small changes to the GraphQL schema around asset partitions so that we don't have to fork and make one query for 2d assets and a separate query for 1d assets.

1)  `partitionKeysByDimension` now works for a single-dimensional asset and returns the partition keys under the name "default"

2) `partitionMaterializationCounts` now returns the same lightweight array-of-numbers shape for one-dimensional assets. This isn't super important, but we'll always query `partitionKeysByDimension` alongside the counts, so we don't want them repeated in the result.

3) I had to rename `materializationCounts` to `materializationCountsGrouped` because GraphQL was upset about two different response types being available with the same field but different data types. (`[[Int!]]` vs `[Int!]`)

Sidenote: I think the GraphQL specification using nested integer arrays means that this only works for 2D assets and not 3D, 4D, etc assets? I think that's fine for now though.

Here's (roughly) the GraphQL the front-end will make which now works for both:

```
{
  assetNodeOrError(assetKey: {path: ["yoyoyoyoyo"]}) {
    __typename
    ... on AssetNode {
      partitionDefinition {
        type
        description
      }
      partitionKeysByDimension {
        name
        partitionKeys
      }
      partitionMaterializationCounts {
        ... on MaterializationCountGroupedByDimension {
          materializationCountsGrouped
        }
        ... on MaterializationCountSingleDimension {
          materializationCounts
        }
      }
    }
  }
}
```

Updated existing test coverage!


clairelin135 (approved):
Thanks for making this change and updating the tests! This all looks good to me

PR #10523, opened 2022-11-15
[dagit] Asset details support for mulit-partitioned assets in Dagit

bengotow:
### Summary & Motivation

Loom summary: https://www.loom.com/share/651b06f0cc64478d9d1f38451b728b96

This PR brings support for multidimensional assets to the "launch backfill" modal and the Asset Details page, which now features a top level partitions tab. *To try this out, you need to have enabled the "New asset detail pages" feature flag!*

There are a few GraphQL adjustments here:

- We choose the time series partition instead of the static partition as the "Primary" by default

- We return `dimensions: []` instead of `dimensions: [{name: default, partitionKeys:[]}]` for an unpartitioned asset

Notes:

- We're using a small hack (detecting 0000-00-00 prefix pattern) to decide if a partition set is a time series. This controls whether the timeline health bar is shown, descending sort order, and whether "Last 50", "Last 100", etc. are presented as options in the backfill modal.

- We only support 2D multipartitioned assets - this is largely because of the GraphQL schema I think. All of the data storage is contained within  `usePartitionHealthData.tsx` and we just need to change the code there for the UI to go full N-dimensional.

- We do not currently support notifying you of missing upstream materializations when you launch a backfill of a two-dimensional range. We could do this, but it can be challenging to express what's missing and our UI doesn't offer enough granularity to remove exact offending asset keys.

- The health bar shows light green instead of "dotted" or "striped" for partially missing.

- The Asset Job partitions page only understands the primary dimension and shows light-green if some second dimensions are present and others are missing, and there is no "click to view details" interaction.

image


clairelin135 (commented):
The backend changes in this PR look great to me!

I did play around in Dagit with the new pages, love the new folder layout of the partitions and the separation between events/partitions.

I did notice a potential bug--seems like the number of partitions calculated for a time dimension might be incorrect. For the assets def below:
```
time_window_partitions = DailyPartitionsDefinition(start_date="2021-05-05")
static_partitions = StaticPartitionsDefinition(["a", "b", "c"])
composite = MultiPartitionsDefinition({"date": time_window_partitions, "abc": static_partitions})


@asset(partitions_def=composite)
def asset1():
    return 1
```
I'm seeing only a subset of the expected dates. Also not sure if we should make the primary dimension in the partition folder view the dimension displayed in the partition bar?
image
This also occurs in the materialize modal on the asset graph view, and I can't launch a backfill:
image

js_modules/dagit/packages/core/src/assets/AssetEventsV2.tsx
```diff
@@ -47,7 +43,7 @@ export const AssetOverview: React.FC<Props> = ({
     loadedPartitionKeys,
     refetch,
     loading,
-  } = useRecentAssetEvents(assetKey, assetHasDefinedPartitions, params);
+  } = useRecentAssetEvents(assetKey, false, params);
```
hellendag, line 46: Un-named boolean args make me itchy -- what does `false` mean?

js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
```diff
@@ -319,50 +308,93 @@ const LaunchAssetChoosePartitionsDialogBody: React.FC<Props> = ({
...
+
+  if (upstreamUnavailableSpans.length === 0) {
+    return <span />;
+  }
+
+  return (
+    <Box margin={{top: 16}}>
+      <Alert
+        intent="warning"
+        title="Upstream Data Missing"
```
hellendag, line 384: Sentence case

js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
```diff
@@ -319,50 +308,93 @@ const LaunchAssetChoosePartitionsDialogBody: React.FC<Props> = ({
...
+        description={
+          <>
+            {upstreamUnavailableSpans
+              .map((span) => stringForSpan(span, ranges[0].selected))
+              .join(', ')}
+            {
+              ' cannot be materialized because upstream materializations are missing. Consider materializing upstream assets or '
+            }
+            <a onClick={onRemoveUpstreamUnavailable}>remove these partitions</a>
```
hellendag, line 393: Use a `ButtonLink` instead, we should avoid using href-less anchors.

hellendag (approved):
There's a bit too much going on here for me to review it very effectively, but I understand that we're aiming to have it in pretty soon, so I'll go ahead and give it a stamp.

My main request is that critical pieces of this get tests soon.

hellendag:
Btw, the green/gray dot pair looks a bit like a `Switch` to me, so they may need a little space between them.

js_modules/dagit/packages/core/src/assets/AssetEventsV2.tsx
bengotow: Good catch going to change this now 👍

bengotow:
Thanks for the quick feedback @hellendag! @braunjj  per our discussion I'm going to remove the feature flag around this before it merges. What do you think of not overlapping the dots or overlapping them less?

bengotow:
Talked with Claire about her bug report above and it'll be fixed in a separate PR in the graphql layer 🙌

PR #10526, opened 2022-11-15
Add experimental register_definitions

schrockn:
### Summary & Motivation

This is a refinement, ready to commit, of the API proposed in https://github.com/dagster-io/dagster/pull/10504.

As we all know, one of the biggest problems with dagster is the redundant, overlapping, and confusing concepts that makes the spin up more difficult and also incurs cognitive load even to existing users. While there are few different parts where this is true, one of the more painful parts is the workspace --> code location --> repository ontology. Also mix in the somewhat-but-not-really-orthogonal instance and deployment concepts, and you have quite the confusing stew.

This PR proposes adding a new entry point function `register_definitions` that would be used instead of `repository` going forward. It is typed and enforces that there is only one repository per python process, which enables the collapsing of code location and repository into a single concept for tooling purposes.

This injects a repository definition as an attribute into the entry point module using python internals, so all auto-discovery features and entry points should work as is.

The user-facing language around this focuses on making the current python module loadable by dagster tools, rather than introducing the code location concept. See the docblock on `register_definitions`

A few additional ergonomic here:

1) `resources` becomes a top-level argument to `register_definitions` and `with_resources` is automatically invoked on the user's behalf
2) `resources` also accepts any python object, not just resource definitions. All vanilla objects are coerced to resource definitions via `ResourceDefinition.hardcoded_resource`.
3) Typed api: At this point `repository` is a pretty rough and loosely defined API that can accept a dizzying array of object types. Types and argument self-document to some degree, stay up to date, and improve discoverability.

**Non-invasive goodbye to repository**

This change can make repository "disappear" for new users. The mental model here is that a call to definitions defines all the definitions for a code location, not a repository. This PR currently enforces this at the module level. We could make it more strict.

For existing users this doesn't really change much either. They will simply think "this is how I define the things in my repo". The function name definitions allows for both interpretations, which is why it was chosen.

If we move forward with this, we can mark code locations that only have a single repository and that were constructed that way as special, and proceed to fork our UI to make the entire concept of repository disappear. Only code location would remain.

For existing users that do have multiple repositories in a single code location, they would instead have to multiple code locations that happen to share the same python dependencies.

Included unit tests and converted examples in follow on PRs (to be actually committed later) once we are ready to flip over the tutorial and user-facing content to lead with this new API

### DIFF
--- a/python_modules/dagster/dagster/__init__.py
+++ b/python_modules/dagster/dagster/__init__.py
@@ -281,6 +281,9 @@ from dagster._core.definitions.reconstruct import (
     build_reconstructable_job as build_reconstructable_job,
     reconstructable as reconstructable,
 )
+from dagster._core.definitions.register_definitions import (
+    register_definitions as register_definitions,
+)
 from dagster._core.definitions.repository_definition import (
     RepositoryData as RepositoryData,
     RepositoryDefinition as RepositoryDefinition,
--- a/python_modules/dagster/dagster/_core/definitions/register_definitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/register_definitions.py
@@ -6,6 +6,7 @@ from types import ModuleType
 from typing import Any, Dict, Iterable, Mapping, Optional, Union
 
 import dagster._check as check
+from dagster._annotations import experimental
 from dagster._core.execution.with_resources import with_resources
 
 from .assets import AssetsDefinition, SourceAsset
@@ -43,6 +44,7 @@ def get_module_name_of_caller() -> str:
     return back_back_frame.f_globals["__name__"]
 
 
+# @experimental
 def register_definitions(
     *,
     assets: Optional[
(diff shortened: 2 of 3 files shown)


sryza (commented):
My main feedback here is that `register_definitions` seems fairly magical. As we briefly discussed, I think it would be easy for a user to guess that it writes the definitions to some sort of cloud service.

IMO the least magical thing would be to communicate to the user that they're setting a module-scoped variable.  Either with a name like `assign_definitions_to_module`. Or a reserved attr called something like `defs` and then they do
```

schrockn:
Oh I thought you were onboard with the name `register_definitions`, despite the possible confusion re: registration with a cloud service.

* re:`assign_definitions_to_module` is just a mouthful. I'm also concerned about introducing the term module in the API. A lot of people don't know the python module system at all, get packages and modules confused, and so for.

* re: `definitions` I am concerned about people NOT doing the assignment then either (1) not loading anything OR (2) it falling back to the autodiscovery behavior and only loading a subset of definitions, which would be super confusing behavior.

PR #10518, opened 2022-11-14
[dagster-airbyte] Add optional user-supplied function to customize asset keys

benpankow:
## Summary

Introduces an optional `connection_to_asset_key` function which can be supplied to `load_assets_from_airbyte_project` and `load_assets_from_airbyte_instance`. This allows a user to customize the asset keys associated with their Airbyte connections.

e.g.

```python
cached_assets = load_assets_from_airbyte_instance(
  airbyte_instance,
  connection_to_asset_key=lambda conn, table_name: f"{conn.name}_{table_name}"
)
```

## Test Plan

New unit tests.

PR #10217, opened 2022-10-26
[dagster-airbyte] Add ability to specify output IO managers

benpankow:
## Summary

Adds two new optional parameters to the Airbyte asset loading functions: `io_manager_key` and `connection_to_io_manager_key_fn`. The former defines an IO manager to apply to all Airbyte assets generated by the function, the latter lets the user customize on a per-connection basis (e.g. if they have many different destinations).

Also adds a new docs entry showing how to sequence downstream assets using this new parameter.

## Test Plan

Unit test.

PR #10529, opened 2022-11-15
[dagster-fivetran] Add optional user-supplied function to customize asset keys, io managers

benpankow:
## Summary

`dagster-fivetran` equivalent of #10518 and #10217, allows users employing `load_assets_from_fivetran_instance` to specify both custom asset keys and IO managers for each table/connector. New inputs:

- `io_manager_key`: defines an IO manager to apply to all Fivetran assets generated by the function.
- `connector_to_io_manager_key_fn`: defines a mapping function from connector to IO manager key for each Fivetran asset, useful if the user has many destinations.
- `connector_to_asset_key_fn`: defines a mapping function from connector & table name to output `AssetKey`.

```python
fivetran_cached_assets = load_assets_from_fivetran_instance(
    ft_instance,
    io_manager_key="snowflake_io_manager",
    connector_to_asset_key_fn=lambda conn, table: AssetKey([*conn.name.split("."), *table.split(".")])
)
```

## Test Plan

New unit tests.


sryza (dismissed):
We should be consistent on `_fn` at the end of these argument names

yuhan (approved):
let's update the summary to `*_fn` for future reference
lg2m!

PR #10532, opened 2022-11-15
Definitions class version

schrockn:
This is what this looks like in class-based version. This does not enforce one-repo-per-module.


schrockn:
going to different RFC

PR #10534, opened 2022-11-15
[freshness-policy] limited partition support

OwenKephart:
### Summary & Motivation

This adds limited time-windowed partition support for declarative scheduling. If an asset is partitioned, then it hands off control to the existing reconciliation logic, but in cases where you have a freshness policy defined downstream of a partitioned asset, we calculate the expected data time using the most recent partition key rather than the most recent materialization timestamp.

This is somewhat error prone, as really we should be using the the start timestamp of first missing time window, using something similar to get_never_requested_never_materialized_partitions() on the AssetReconciliation cursor, but this only contains information on root partitioned assets, and we need to know this for any asset in the graph. In the near future, we should be able to abuse the instance kvs stuff to stash the needed partition subset information for any asset in the graph, and switch over to using that instead.


sryza (commented):
It would be helpful to include a more granular unit test or few on `calculate_used_data` to verify / document its expected behavior for partitions.

PR #10535, opened 2022-11-15
use random string for compute log file key instead of pid which will collide

prha:
### Summary & Motivation
K8s cluster tests in cloud hit some test failures because the pids for recently spun up tasks were all colliding.

Ran cloud k8s test suite, where all runs will have `pid == 1`.

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/plan/execute_plan.py
+++ b/python_modules/dagster/dagster/_core/execution/plan/execute_plan.py
@@ -1,4 +1,3 @@
-import os
 import sys
 from contextlib import ExitStack
 from typing import Iterator, Sequence, cast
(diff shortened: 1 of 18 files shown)


python_modules/dagster/dagster/_core/execution/plan/execute_plan.py
```diff
@@ -42,8 +42,8 @@ def inner_plan_execution_iterator(
 
         # begin capturing logs for the whole process if this is a captured log manager
         if isinstance(compute_log_manager, CapturedLogManager):
-            pid = str(os.getpid())
-            log_key = compute_log_manager.build_log_key_for_run(pipeline_context.run_id, pid)
+            file_key = create_compute_log_file_key()
+            log_key = compute_log_manager.build_log_key_for_run(pipeline_context.run_id, file_key)
```
alangenfeld, line 46: should we include step key at all when we have a single step key? i guess no real reason to bother its not that useful

alangenfeld (approved):
good catch - i thought of this at somepoint but didn't raise it

PR #10541, opened 2022-11-16
resolve graphene error matching instigation enum

prha:
### Summary & Motivation
Started getting some graphene errors because of how python enums were getting resolved to graphene enums.

I think there might be some version differences.

Resolving this by forcing the graphene enum type to use the `from_enum` constructor, and regenerated the types from it (instead of coercing from a string repr).

BK, rendered schedule/sensor pages


hellendag (approved):
Does this need a `make generate-graphql`?

alangenfeld (approved):
yep - i had to make this change repeatedly in the graphene 3 conversion 

this must not have any test coverage at the graphql layer

PR #10544, opened 2022-11-16
Add method to add_asset_event_tags

OwenKephart:
### Summary & Motivation

Gives the ability to add tags post-hoc to a given record, as well as query for the tags of a given record id. If you add tags to a record after it's been stored, these will not be reflected on the EventLogRecord itself, so the tags table is the source of truth for the set of tags on an asset. This is pretty inconsequential, as the tags we want to add post-hoc are currently all for internal machinery.

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
+++ b/python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
@@ -217,9 +217,15 @@ class SqlEventLogStorage(EventLogStorage):
         return self.has_table(AssetEventTagsTable.name)
 
     def add_asset_event_tags(
-        self, event_record: EventLogRecord, new_tags: Mapping[str, str]
+        self,
+        event_id: int,
+        event_timestamp: float,
+        asset_key: AssetKey,
+        new_tags: Mapping[str, str],
     ) -> None:
-        check.inst_param(event_record, "event_record", EventLogRecord)
+        check.int_param(event_id, "event_id")
+        check.float_param(event_timestamp, "event_timestamp")
+        check.inst_param(asset_key, "asset_key", AssetKey)
         check.mapping_param(new_tags, "new_tags", key_type=str, value_type=str)
 
         if not self.has_asset_event_tags_table:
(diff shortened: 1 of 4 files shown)


python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1294,6 +1377,9 @@ def get_tag_filter_query(tag_key, tag_value):
                 )
             )
 
+        if filter_event_id is not None:
```
clairelin135, line 1380: Can we move these lines within the if-else block above? so
```
if not filter_tags:
            tags_query = (
               ...
            )
            if filter_event_id is not None:
                 tags_query = tags_query.where(AssetEventTagsTable.c.event_id == filter_event_id)
```
and 
```
def get_tag_filter_query(...)
     filter_query = ....
      if filter_event_id is not None:
             filter_query = filter_query.where(AssetEventTagsTable.c.event_id == filter_event_id)
```
Mostly just because the intersect query may be more expensive and I think it would be good to filter out undesired event IDs before the intersect

clairelin135 (commented):
Thanks for making this change and adding the checks to make sure the updated tags all have unique keys!

The one part i'm unsure about is that the new tags won't be part of the asset materialization object. I think this part should be fine though because none of these tags should be user-populated (and we can always fetch these tags from the tags table before returning, if desired). Thoughts @prha?

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
OwenKephart: I think the query optimizer inside of postgres will handle that sort of optimization for us 🤞

prha:
Hmm, it feels a little scary to not get the tags for a given materialization.

I imagine we'd want to display partition information at some point, when viewing materializations, rather than just filtering for status?

clairelin135 (approved):
Not having the tags exist inside the materialization is definitely a little odd, but I can't think of any reason why someone would want to display the freshness data. Going to go ahead and approve this so that this feature can make it into the release.

PR #10549, opened 2022-11-16
[freshness-policies] fix source asset freshness

OwenKephart:
### Summary & Motivation

When calculating how fresh a given asset is, this must be done relative to the most upstream AssetsDefinitions in the repository (SourceAsset updates are currently not factored into this calculation, but should be in the future). These updates do some refactoring, and ensure that when calculating which upstream assets to consider, we only factor in non source assets.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -110,11 +110,14 @@ class AssetGraph(
     def from_external_assets(external_asset_nodes: Sequence["ExternalAssetNode"]) -> "AssetGraph":
         upstream = {}
         downstream = {}
+        source_asset_keys = set()
         partitions_defs_by_key = {}
         group_names_by_key = {}
         freshness_policies_by_key = {}
 
         for node in external_asset_nodes:
+            if node.is_source:
+                source_asset_keys.add(node.asset_key)
             upstream[node.asset_key] = {dep.upstream_asset_key for dep in node.dependencies}
             downstream[node.asset_key] = {dep.downstream_asset_key for dep in node.depended_by}
             partitions_defs_by_key[node.asset_key] = (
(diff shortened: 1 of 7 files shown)


python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -310,18 +313,32 @@ def get_parent_partition_keys_for_child(
...
+            if not self.has_non_source_parents(key)
+        }
 
     def upstream_key_iterator(self, asset_key: AssetKey) -> Iterator[AssetKey]:
         """Iterates through all asset keys which are upstream of the given key."""
         visited: Set[AssetKey] = set()
         queue = deque([asset_key])
         while queue:
             current_key = queue.popleft()
+            if current_key in self.source_asset_keys:
```
sryza, line 340: Why is this needed? Won't the call to `self.get_parents` on the next line return an empty iterable?

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -443,6 +443,8 @@ def get_freshness_constraints_by_key(
     # upstream of another asset.
     for level in reversed(asset_graph.toposort_asset_keys()):
         for key in level:
+            if key in asset_graph.source_asset_keys:
```
sryza, line 446: Same question here as above

sryza (approved):

python_modules/dagster/dagster/_core/definitions/asset_graph.py
OwenKephart: the dependency dictionary doesn't contain a key for source assets, and get_parents doesn't use a get() call (it just accesses the dictionary directly). arguably, it should use .get(), was just trying not to disturb existing logic too much

PR #10552, opened 2022-11-16
remove IOManager.get_output_asset_key and related methods

sryza:
### Summary & Motivation

Before software-defined assets, we allowed users to associate asset keys with op outputs by defining a method on their IO manager.  This was always experimental, and we removed documentation for it in a recent major release.  This removes it entirely.


smackesey (approved):
lgtm once you remove the remaining tests using this functionality, which you should be able to determine from the BK failures

PR #10553, opened 2022-11-16
add cached_status_data field to AssetKeyTable

sryza:
### Summary & Motivation

This doesn't yet support actual functionality, but aiming to get it in for the release so it will get included as part of schema migrations that users run for 1.1.

The aim here is to support this proposal for cached/incremental asset partition summaries: https://www.notion.so/dagster/Cached-incremental-asset-partition-summaries-9afd9aa0b91940d49c3d5af5a832131a.

The cache would hold a serialized blob that represents the ranges of partitions that are materialized / reconciled / etc., along with the `latest_storage_id` and `partitions_def_id` that it's valid for.

I'm not confident on the name / scope at all. Other options include:
* partition_status_cache
* status_cache
* cached_status_info
etc.


clairelin135 (commented):
All of this makes sense to me and sounds good. I can approve this, though we still need to add the migration files and backcompat tests ([guide here](https://github.com/dagster-io/dagster/blob/5ab01a675801ee56fc34e90c16301f3d557d40d4/python_modules/dagster/dagster/_core/storage/alembic/README.md)). 

I think adding this column and being able to fetch the last materialization ID will help to speed up queries that require fetching asset event tags for the latest materialization (which affects asset versioning and freshness policy). I don't completely understand the presence of the `partitions_def_id`--how will know when the partitions definition has changed? But in any case that doesn't affect this immediate PR.

sryza (commented):
🙌 

@prha - do you want a chance to look at this before I approve it?

prha:
you'll need the corresponding alembic version bump in internal

prha:
(ideally by applying a similar migration in cloud, but understand if time is short)

clairelin135:
Added the corresponding internal PR! https://github.com/dagster-io/internal/pull/4107

Release is happening likely soon, so I just bumped the alembic version instead of adding a migration. I think this is likely fine though since we aren't writing to the column and can add the migration in next week's release

clairelin135 (approved):

PR #10556, opened 2022-11-16
Remove remaining references to “upstream changed” calculation + state

bengotow:
### Summary & Motivation

This PR doesn't need to ship today, but with the new asset versioning landed, Dagit is no longer using this resolver and I think we can safely remove this code!


sryza:
Awesome! Going to defer to @clairelin135 and @smackesey on this one.

smackesey (approved):
Generally LGTM but looks like this PR does considerably more than is listed in the summary?

bengotow:
Ahh sorry for the hassle @smackesey , it looks like when the stack merged this one was left in a weird state where it incorporated more changes than it should have...

PR #10559, opened 2022-11-16
Adjust GrapheneAssetNode multipartitions resolvers to return values in dimension definition order

clairelin135:
In `GrapheneAssetNode`, adjusts the multipartitions resolvers to return partitions in the same order as they exist on the backend (alphabetically by dimension name). 

Also adds a field to fetch the type of each dimension as well as a description.


clairelin135:
@bengotow I've adjusted this PR to return all of the contents in the order that the dimensions exist, and added a field to populate type per dimension. I think this is ready for you to take another look!

js_modules/dagit/packages/core/src/assets/PartitionHealthSummary.tsx
```diff
@@ -34,7 +34,7 @@ export const PartitionHealthSummary: React.FC<{
     return <div />;
   }
 
-  const keys = timeDimension.partitionKeys;
+  const keys = assetData.dimensions[0].partitionKeys;
```
bengotow, line 37: Hmm do you remember if this was broken before?

I think there are a couple other places in Dagit where we identify the "isTimeseries" partition and then use that for a timeline (it's kind of annoying but right now PartitionStatus.tsx is a separate version of a component very similar to this one, but it's interactive). If this works when it uses `timeDimension.partitionKeys`, it's probably preferable to continue doing that for now for consistency with those other places.

Josh and I were talking about revisiting this and making a variant of the timeline for static partition sets, possibly a list of boxes where there's a small whitespace between them marking that it's not really a "range".

python_modules/dagster-graphql/dagster_graphql/implementation/fetch_assets.py
```diff
@@ -311,17 +315,23 @@ def get_materialization_cts_grouped_by_dimension(
...
+    for (
+        primary_dim_key
+    ) in (
+        primary_dimension.external_partitions_def_data.get_partitions_definition().get_partition_keys()
+    ):
         materialization_counts_grouped_by_dimension.append(
             [
                 db_materialization_counts.get(primary_dim_key, {}).get(secondary_dim_key, 0)
-                for secondary_dim_key in partition_keys_by_dimension[secondary_dimension]
+                for secondary_dim_key in secondary_dimension.external_partitions_def_data.get_partitions_definition().get_partition_keys()
```
bengotow, line 334: This is a dumb nit, but I'm not sure how expensive this is:
```
secondary_dimension.external_partitions_def_data.get_partitions_definition().get_partition_keys()
```

It might be nice to prefetch it into a variable outside the inner loop so it's not repeated for each primary dimension key!

bengotow (approved):
This looks great! 🙌  Just left a few inline comments but nothing really blocking.

js_modules/dagit/packages/core/src/assets/PartitionHealthSummary.tsx
clairelin135: Yep, this was broken before. So this PR changes the default partition bar view and folder view to choose whichever dimension was ordered first. If the first dimension is not the time dimension, the partition view would break and the whole display would not show up before I made this change.

bengotow: Ok that's good to know! I'll do a bit more testing of this case today just to be safe

PR #10570, opened 2022-11-16
[dagit] Small Asset UI adjustments for the 1.1 release

bengotow:
### Summary & Motivation

This PR makes the following small changes:

- Asset nodes have a more visible selection highlight (in zoomed out mode as well)
image

- The Materialize button on the asset graph now only shows Observe source IF there are observable source assets, and is completely hidden otherwise.

- The Materialize button is split the same way the "Re-execute" button is. It shows a primary action and "Stale and missing" in the dropdown.
    + If you select nodes, it shows just the selected option
    + If you have no stale assets, the dropdown is still there and the option is grayed out
    + Shift-click to add config still works for all modes, including choosing from the dropdown

image

- The "Show notebook" button padding has been fixed here:

image

- The Stale tag placement has been changed here so it's right justified:
image

- Late assets are shown on the asset graph
image

- For large graphs where the live data takes time to load, there's now a loading state instead of "Never materialized"
image

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
@@ -278,7 +278,6 @@ export const AssetGraphExplorerWithData: React.FC<
 
   const allowExperimentalZoom =
     flags.flagAssetGraphExperimentalZoom && layout && Object.keys(layout.groups).length;
-  const selectionContext = selectedGraphNodes.length ? 'selected' : 'all';
 
   return (
     <SplitPanelContainer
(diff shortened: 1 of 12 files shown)


sryza:
Small thing - let me know if you disagree - I think it should be "Observe source**s**" instead of "Observe source".

braunjj:
Another small nit, but I think in the Materialize dropdown we should say `Materialize stale and missing` ... i know it's kinda long but I think we have room for it since it's in a dropdown anyway

smackesey:
>Another small nit, but I think in the Materialize dropdown we should say Materialize stale and missing ... i know it's kinda long but I think we have room for it since it's in a dropdown anyway

I'm on the fence about this-- a compromise might be if we could make the "All" disappear when you clicking the dropdown? So the upper button just says "Materialize"-- that would make it more clear that "Stale and Missing" is the object of materialization.

sryza:
> Another small nit, but I think in the Materialize dropdown we should say Materialize stale and missing ... i know it's kinda long but I think we have room for it since it's in a dropdown anyway

I had actually advocated for just "Stale and missing" to save space. I'm OK with "Materialize" though if @bengotow and @braunjj agree on it.

hellendag (approved):

bengotow:
Thanks folks! I've applied these suggestions (with fresh eyes I do think "Materialize stale and missing" is slightly better). Will merge this as soon as the build passes!

PR #10579, opened 2022-11-17
Correct typo in docs

C0DK:
### Summary & Motivation

There was a small typo in the documentation

As this is merely documentation, it's not really that testable, so io just looked at the before and after versions. Now it looks like valid python code.


sryza (approved):
This looks good. Thanks for the contribution!

PR #10589, opened 2022-11-17
allow op context to asset partition key

### LINKED ISSUE #10586: Allow partition asset tests by adding `asset_partition_key_for_output` to invocation context
### Dagster version

1.0.17

### What's the issue?

```
context = build_op_context(..., partition_key="2022-01-01")
my_partitioned_asset(context)
```

fails with 

```
File "/Users/lopp/Projects/internal/python_modules/purina/.venv/lib/python3.10/site-packages/dagster/_core/definitions/assets.py", line 180, in __call__
    return solid_def(*new_args, **kwargs)
  File "/Users/lopp/Projects/internal/python_modules/purina/.venv/lib/python3.10/site-packages/dagster/_core/definitions/solid_definition.py", line 183, in __call__
    return solid_invocation_result(self, context, *args[1:], **kwargs)
  File "/Users/lopp/Projects/internal/python_modules/purina/.venv/lib/python3.10/site-packages/dagster/_core/definitions/solid_invocation.py", line 61, in solid_invocation_result
    compute_fn.decorated_fn(context, **input_dict)
  File "/Users/lopp/Projects/internal/python_modules/purina/purina/assets/sales/__init__.py", line 72, in stale_sfdc_accounts
    current_partition = context.asset_partition_key_for_output()
  File "/Users/lopp/Projects/internal/python_modules/purina/.venv/lib/python3.10/site-packages/dagster/_core/execution/context/compute.py", line 382, in asset_partition_key_for_output
    return self._step_execution_context.asset_partition_key_for_output(output_name)
AttributeError: 'BoundSolidExecutionContext' object has no attribute '_step_execution_context'. Did you mean: 'get_step_execution_context'?
Exception ignored in: <generator object build_resources at 0x156f5a650>
```

I want to be able to run an asset for a manually specified partition key during tests by using `build_op_context` with a partition. 

### How to reproduce?

Create a partitioned asset that attempts to access the partition key, e.g. `current_partition = context.asset_partition_key_for_output()` 
Create an op context using `build_op_context`
Attempt to call the asset function 

### Deployment type

Local

### Deployment details

_No response_

### Additional information

_No response_

slopp:
### Summary & Motivation

Resolves #10586 

- added a test that directly invokes the asset with an op context

Question: Is it safe to ignore the output_name argument in this context?

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/context/invocation.py
+++ b/python_modules/dagster/dagster/_core/execution/context/invocation.py
@@ -261,7 +261,7 @@ class UnboundOpExecutionContext(OpExecutionContext):
             user_events=self._user_events,
             output_metadata=self._output_metadata,
             mapping_key=self._mapping_key,
-            partition_key=self._partition_key
+            partition_key=self._partition_key,
         )
 
     def get_events(self) -> Sequence[UserEvent]:
@@ -387,7 +387,7 @@ class BoundOpExecutionContext(OpExecutionContext):
         user_events: List[UserEvent],
         output_metadata: Dict[str, Any],
         mapping_key: Optional[str],
-        partition_key: Optional[str]
+        partition_key: Optional[str],
     ):
         self._op_def = op_def
         self._op_config = op_config
@@ -402,9 +402,8 @@ class BoundOpExecutionContext(OpExecutionContext):
         self._user_events = user_events
         self._seen_outputs = {}
         self._output_metadata = output_metadata
-        self._mapping_key = mapping_key,
+        self._mapping_key = mapping_key
         self._partition_key = partition_key
-        
 
     @property
     def solid_config(self) -> Any:
@@ -537,7 +536,7 @@ class BoundOpExecutionContext(OpExecutionContext):
                 output_name in self._seen_outputs and mapping_key in self._seen_outputs[output_name]
             )
         return output_name in self._seen_outputs
-    
+
     def asset_partition_key_for_output(self, output_name: str = "result") -> str:
         if self._partition_key is not None:
             return self._partition_key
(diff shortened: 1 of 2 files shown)


dpeng817 (commented):
deferring to @sryza on the multiple partition behavior which I do not have context on.

python_modules/dagster/dagster/_core/execution/context/invocation.py
```diff
@@ -533,6 +537,12 @@ def has_seen_output(self, output_name: str, mapping_key: Optional[str] = None) -
             )
         return output_name in self._seen_outputs
 
+    def asset_partition_key_for_output(self, output_name: str = "result") -> str:
```
OwenKephart, line 540: I think it might be better to override the `partition_key` property rather than this function

slopp: @OwenKephart  I attempted to do that and got an error 

```
 AttributeError: 'BoundOpExecutionContext' object has no attribute '_step_execution_context'. Did you mean: 'get_step_execution_context'?
```

Is the BoundOpExecutionContext supposed to have a step execution context?

OwenKephart: Hm I think this is a bit tricky, as I would expect users to more often call `context.partition_key` in order to get that information in the context of a run. But it looks like we support a few different ways for users to branch behavior inside the body of the asset based off of partition information.

Maybe both need to be overridden? @sryza

slopp: I added an update that implements the property and method in both the unbound and bound context plus test coverage.

OwenKephart (approved):

PR #10591, opened 2022-11-17
K8s tags are needed in ops to propagate to steps

nicholsn:
### Summary & Motivation

I raised a [support request](https://dagster.slack.com/archives/C01U954MEER/p1668643577955129) on slack where it wasn't clear to me why  k8s `tags` specified at the `@job` level weren't propagating to individual steps when using the `k8s_job_executor`.

This update to the docs includes a note that will hopefully help others understand that k8s `tags` need to be defined at the `op` level for a given `step` to use them.

My primary use case for this feature is to have ops run on specific node groups.

Local build of docs.


clairelin135 (approved):

clairelin135:
Thanks for making this change @nicholsn

PR #10592, opened 2022-11-17
Fix reference to S3 in GCSComputeLogManager docstring

reidab:
### Summary & Motivation

Noticed that the GCSComputeLogManager docstring referenced S3. Updated to reference GCS.

No testing required.


clairelin135 (approved):
Thanks for making this change!

PR #10593, opened 2022-11-17
[dagstermill] yield events from dagstermill assets

jamiedemaria:
### Summary & Motivation
Adds the capability to yield events from dagstermill assets.

Also adds a check that additional outputs are note yielded from notebooks (since they cannot be handled by dagster right now because there can only be one output per asset and that output is the notebook file)

bk + new unit tests


python_modules/libraries/dagstermill/dagstermill/manager.py
```diff
@@ -335,6 +335,13 @@ def yield_result(self, value, output_name="result"):
         # pass output value cross process boundary using io manager
         step_context = self.context._step_context  # pylint: disable=protected-access
         # Note: yield_result currently does not support DynamicOutput
+
+        # dagstermill assets do not support yielding additional results within the notebook:
+        if len(step_context.job_def.asset_layer.asset_keys) > 0:
```
jamiedemaria, line 340: this was the best way i could find to determine if the notebook was being run as a op vs as an asset. if there's a more elegant way lmk!

python_modules/libraries/dagstermill/dagstermill/manager.py
```diff
@@ -335,6 +335,13 @@ def yield_result(self, value, output_name="result"):
...
+            raise DagstermillError(
+                "dagstermill assets do not currently support dagstermill.yield_result"
```
jamiedemaria, line 342: add . to end of the sentence

sryza (commented):
Are you able to change `io_manager.handle_output(output_context, value)` to `list(io_manager.handle_output(output_context, value))` and make sure the tests still pass? That was the symptom that originally alerted me to the problem, and should be fixed anyway.

jamiedemaria:
hmm ok i think there are 2 things going on here:
1. `io_manager.handle_output` should never get called in `yield_result` if the notebook is an asset (until we support multi assets), that's where the new check comes in
2. when yield_result is called the io manager used shouldnt be the `local_output_notebook_io_manager`. The reason it is in this test is because the output `result` is the notebook which is given the io manager  `local_output_notebook_io_manager`. then in the notebook when we call `yield_result` it's given the default output name `result` so it tries to use the notebook io manager. 

When yield_result is supported for assets the outputs in the notebook will need to be named and then the correct io manager will be used for them. typically the `handle_output` method isn't a generator, so just calling the fn should be fine, but i guess someone could write a custom io manager that's a generator. I tried wrapping `handle_output` in `list` but then it fails when the io manager returns None. still figuring out what to do about that

the last thing: I don't know why the notebook io manager `handle_output` is a generator. is there a reason for that? if not, i'll make it a normal function

python_modules/libraries/dagstermill/dagstermill/manager.py
```diff
@@ -335,13 +335,21 @@ def yield_result(self, value, output_name="result"):
...
+            )
+
         step_output_handle = StepOutputHandle(
             step_key=step_context.step.key, output_name=output_name
         )
         output_context = step_context.get_output_context(step_output_handle)
         io_manager = step_context.get_io_manager(step_output_handle)
 
         # Note that we assume io manager is symmetric, i.e handle_input(handle_output(X)) == X
+        # use list() in case handle_output is a generator
```
sryza, line 352: did you mean to use `list` here?

jamiedemaria: forgot to delete the comment, oops. thanks for the catch

sryza (approved):

PR #10595, opened 2022-11-17
[dagit] Fix cron union string parsing

hellendag:
### Summary & Motivation

A recent change made it possible to define a union of cron strings for a schedule. We don't currently handle this in the UI, because in all scenarios we just expect a single cron string.

To support this, attempt to parse the supplied cron string as a JSON-stringified array. (It won't currently be, technically, because the string provided from GraphQL uses single quotes on the array elements.) Then, humanize the resulting elements of the parsed array, or of the single-element array that results from a single cron string.

The resulting combined string will have the humanized elements divided by a semicolon.

Added tests to cover the union case.

### DIFF
--- a/js_modules/dagit/packages/core/src/schedules/humanCronString.test.ts
+++ b/js_modules/dagit/packages/core/src/schedules/humanCronString.test.ts
@@ -98,6 +98,7 @@ describe('humanCronString', () => {
 
     afterAll(() => {
       jest.clearAllMocks();
+      jest.restoreAllMocks();
     });
 
     it('shows 24h format if locale uses it, and shows timezone if provided, if cron specifies a time', () => {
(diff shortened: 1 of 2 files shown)


hellendag:
* master
  * **PR #10595** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10595" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10595?utm_source=stack-comment).

js_modules/dagit/packages/core/src/schedules/humanCronString.test.ts
```diff
@@ -127,4 +128,36 @@ describe('humanCronString', () => {
...
+  describe('Cron union', () => {
+    it('handles a single-item union of cron strings, with timezone', () => {
+      const timezone = 'America/Phoenix';
+      expect(humanCronString("['2,4-5 1 * * *']", timezone)).toBe(
+        'At 2 and 4 through 5 minutes past the hour, at 01:00 AM MST',
+      );
+    });
+
+    it('handles a union of cron strings', () => {
+      expect(humanCronString("['0 2 * * FRI-SAT', '0 2,8 * * MON,FRI', '*/30 9 * * SUN']")).toBe(
```
bengotow, line 141: Good tests! 

It sounds like this is actually a python serialized object string and not a JSON string? I wonder if that could be changed on the GraphQL side...

bengotow (approved):

prha:
Yeah, this seems like it'd be easier to handle in the graphql layer, where we can actually do the type check against a string vs list of strings.

hellendag:
Would it be simplest just to make the field `[String]`? I don't mind updating the GraphQL usage to handle the array.

prha:
Yeah, I think that makes sense to me.

hellendag:
I'm fairly bogged down in try to get the Python side of this working and I don't want to spend all day hacking at it. I'm also not sure I'm even doing it thoroughly enough, given that there are a few places where we have `cronSchedule` fields, and I don't know which of them are supposed to support the union.

I'm going to just go ahead with this as is to get the bug fixed, and we can follow up with a GraphQL change.

PR #10599, opened 2022-11-17
[managed-elements][docs] apidoc for managed elements

benpankow:
## Summary

A bunch of APIdoc for managed elements, including for the generated classes.

Open question: should these APIdocs (especially for generated classes) live on another page? it might be too busy as-is?

## Test Plan

[vercel](https://dagster-55sd8v8dn-elementl.vercel.app/_apidocs/libraries/dagster-airbyte#managed-config)


benpankow:
* master
  * **PR #10599** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10599" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10599?utm_source=stack-comment).

erinkcochran87 (commented):
I added some comments on the Vercel preview for a few small things, but overall I think this looks good.

I don't mind having the generated classes on this page, tbh. It keeps everything in one place, and it being at the end doesn't make it feel like a slog to get through as all the important bits are at the top.

I noticed our API docs don't list the specific properties users need to provide to configure sources and destinations, ex: `destination_path` as used in the JSON example in your Getting Started guide. It took me a moment to realize that clicking **Source** in our docs would show me the source code where I could then look at the properties I'd need to provide.

I think this could be improved - we don't want users to go hunting for this info. Could we list the properties/definitions alongside the other reference info, or at least provide an example?

yuhan (dismissed):
@benpankow per Erin's comments in vercel, bump this back to your queue. would be great to get in this week's release

benpankow:
Updated to include the various nested classes for sources and destinations.

erinkcochran87:
@benpankow Thanks for figuring this out - I think this looks great and will be so helpful for users. I think we can address my comments about formatting URLs as links in a follow-up PR, but I would like the instances of `https://docsurl.com/` to be addressed - could you take care of that?

benpankow:
> @benpankow Thanks for figuring this out - I think this looks great and will be so helpful for users. I think we can address my comments about formatting URLs as links in a follow-up PR, but I would like the instances of `https://docsurl.com/` to be addressed - could you take care of that?

Yes, this is actually an issue with the underlying Airbyte definitions we're generating the docs from - they've left the docs URL as the default (docsurl.com). I can identify those and pull them out.

erinkcochran87 (approved):
Thanks Ben! :shipit:

PR #10600, opened 2022-11-17
add missing projected version loader when loading workspace/repo page

sryza:
### Summary & Motivation

I'm a bit out of my depth on this part of the codebase, but this appeared to fix the issue.

Loaded the repo page under the Workspace page and observed no graphql errors

```
from dagster import asset


@asset
def asset1():
    ...
```


prha (approved):
yeah, this looks right to me, though not familiar enough to know if we're overfetching now.

PR #10603, opened 2022-11-18
updated CHANGES.md to reflect that schema migration is required

sryza:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,9 +30,10 @@
 
 ### Database migration
 
-- Optional database schema migration, which can be run via `dagster instance migrate`:
-    - Improves Dagit performance by adding database indexes which should speed up the run view as well as a range of asset-based queries.
-    - Enables multi-dimensional asset partitions and asset versioning.
+- This release requires a database schema migration, which can be run via `dagster instance migrate`. The schema migration does the following:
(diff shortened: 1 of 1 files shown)


clairelin135 (approved):
Not sure if it's bad to do this, but 1.0.17 changelog mentions that the migration is optional: https://github.com/dagster-io/dagster/blob/0549db660d3a76dbb885c5a3d4fe3e81b847aab8/CHANGES.md?plain=1#L86

maybe we should change that part to mention required?

CHANGES.md
```diff
@@ -30,9 +30,7 @@
 
 ### Database migration
 
-- Optional database schema migration, which can be run via `dagster instance migrate`:
-    - Improves Dagit performance by adding database indexes which should speed up the run view as well as a range of asset-based queries.
-    - Enables multi-dimensional asset partitions and asset versioning.
+- This release requires a database schema migration, which can be run via `dagster instance migrate`.
```
prha, line 33: may want to still leave the reason why this is necessary.

prha (approved):

CHANGES.md
sryza: 👍

prha (approved):

PR #10611, opened 2022-11-18
Silent exit when asset tags table nonexistent 

clairelin135:
Now that logical versioning adds tags to each asset materialization, we want to ensure dagster still supports behavior regardless of whether the asset event tags table exists or not. This PR amends the existing behavior to not error and silently exit when the asset event tags table does not exist and dagster attempts to write tags. On read (which we do not currently do for logical versions) we continue to throw an error if we attempt to read from the asset event tags table and it does not exist.

Prior to this PR, the `has_table` method was broken and did not truly detect whether a table existed or not. This PR corrects this behavior and adds tests to confirm an informative error is raised.

Tested in bk to confirm asset materializations can be stored regardless of whether the tags table exists.


prha (approved):

prha (commented):
What's the error message if the table hasn't been created on read?

clairelin135:
@prha here it is: https://github.com/dagster-io/dagster/blob/b066f959150d66afeea7eece9f44f4bd85201a39/python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py#L1241

we have a similar one when you try to filter event log records by tags

PR #10612, opened 2022-11-18
import script_relative_path

alvinyeap:
script_relative_path needs to be imported in the example, otherwise raises an error.


clairelin135 (changes requested):
Hi Alvin, thanks for making this change! Would you actually mind editing the [example file](https://github.com/dagster-io/dagster/blob/fab86ed05b9b2102b2f933306ef5169644eaa1fb/examples/docs_snippets/docs_snippets/legacy/data_science/iris_classify_2.py#L1) directly?

You can rebuild the doc snippet by going to `dagster/docs/` and running `make mdx-format`

PR #10615, opened 2022-11-18
Add build_snowflake_pandas_io_manager wrapper to eliminate use of type handlers in the common case

schrockn:
### Summary & Motivation

To use our built in i/o managers for duckdb and snowflake, one has to import two libraries and be introduced to the concept of "type handlers", which are pretty arcane implementation detail that most people don't have to know about it. This PR introduces a function `build_snowflake_pandas_io_manager` that obviates the need to introduce SnowflakePandasTypeHandler in the common case. It also converts one example to this new function as a demo. If we move forward, we would also do this for duckdb and convert all examples.

Unit tests

### DIFF
--- a/python_modules/libraries/dagster-snowflake-pandas/dagster_snowflake_pandas/snowflake_pandas_type_handler.py
+++ b/python_modules/libraries/dagster-snowflake-pandas/dagster_snowflake_pandas/snowflake_pandas_type_handler.py
@@ -1,11 +1,19 @@
 from typing import Mapping, Union, cast
 
 import pandas as pd
+from dagster_snowflake import build_snowflake_io_manager
 from dagster_snowflake.resources import SnowflakeConnection
 from dagster_snowflake.snowflake_io_manager import SnowflakeDbClient
 from snowflake.connector.pandas_tools import pd_writer
 
-from dagster import InputContext, MetadataValue, OutputContext, TableColumn, TableSchema
+from dagster import (
+    IOManagerDefinition,
+    InputContext,
+    MetadataValue,
+    OutputContext,
+    TableColumn,
+    TableSchema,
+)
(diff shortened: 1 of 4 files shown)


schrockn (commented):
The most obvious downside here is replication of docs, but I think that is worth it in order to get the improved ergonomics in the common case.

python_modules/libraries/dagster-snowflake-pandas/dagster_snowflake_pandas/snowflake_pandas_type_handler.py
```diff
@@ -113,3 +121,65 @@ def load_input(self, context: InputContext, table_slice: TableSlice) -> pd.DataF
     @property
     def supported_types(self):
         return [pd.DataFrame]
+
+
+def build_snowflake_pandas_io_manager() -> IOManagerDefinition:
+    """
+    Builds an IO manager definition that reads inputs from and writes outputs to Snowflake.
```
sryza, line 128: I think worth mentioning Pandas in this docstring

sryza (approved):
I like it

jamiedemaria (commented):
i like it too!
An idea:
 since most of our io managers don't need to be "built" (ie you specify the object `fs_io_manager` not call a function `build_fs_io_manager()` ) could we do something similar here? so the user would just say `snowflake_pandas_io_manager` instead of `build_snowflake_pandas_io_manager()`? 

I'm not sure what the implications of this idea are since it would mean we instantiate something like
```
snowflake_pandas_io_manager = build_snowflake_io_manager([SnowflakePandasTypeHandler()])
```

in the file and then make `snowflake_pandas_io_manager` part of the public api. would that have perf affects?

jamiedemaria (approved):
so cool

PR #10617, opened 2022-11-18
remove make_read_only in AssetGraph constructor

sryza:
### Summary & Motivation

I think we can rely on immutable types to enforce this and save some cycles.


OwenKephart:
@sryza this was necessary to make it possible to pass an AssetGraph into a function decorated by @cached_property, as that decorator requires the input object to be hashable. I think this may be fixable w/o doing make_read_only_value(), but I'm not sure exactly how. Probably involves no longer depending on the NamedTuple subclass

sryza:
Ahh, makes sense. I think it's probably a waste to be iterating over the entire graph every time an AssetGraph is passed to a cached method.  I updated it to just override `__hash__` and `__eq__` to be based on reference equality.

PR #10620, opened 2022-11-18
WIP: Add support to runs that target a range of partitions to UPathIOManager 

### LINKED ISSUE #10490: Executing with an asset_partition_range throws error in `fs_io_manager`
### Dagster version

1.0.17

### What's the issue?

Adding the following tags to a partitioned asset job causes an error when using fs_io_manager:
```
ASSET_PARTITION_RANGE_START_TAG: "2020-01-01",
ASSET_PARTITION_RANGE_END_TAG: "2020-01-03"
```
```
dagster._core.errors.DagsterExecutionHandleOutputError: Error occurred while handling output "result" of step "first_asset1"::

dagster._check.CheckError: Failure condition: Tried to access partition key for output 'result' of step 'first_asset1', but the step output has a partition range: '2022-06-15' to '2022-06-20'.
```
The error is occurring because the run has a partition range and `fs_io_manager` attempts to store the step output in a file path containing the partition key.

_No response_

### How to reproduce?

_No response_

### Deployment type

_No response_

### Deployment details

_No response_

### Additional information

_No response_

_No response_

danielgafni:
### Summary & Motivation

resolve #10490

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/context/output.py
+++ b/python_modules/dagster/dagster/_core/execution/context/output.py
@@ -125,6 +125,7 @@ class OutputContext:
         asset_info: Optional[AssetOutputInfo] = None,
         warn_on_step_context_use: bool = False,
         partition_key: Optional[str] = None,
+        partition_key_range: Optional[PartitionKeyRange] = None,
     ):
         from dagster._core.definitions.resource_definition import IContainsGenerator, Resources
         from dagster._core.execution.build_resources import build_resources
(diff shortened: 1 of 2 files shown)


sryza (commented):
I would call this parameter `asset_partition_key_range` instead of `partition_key_range`, because it corresponds to tags that have assets in their names. Then you could do the same thing that's done with the `partition_key` parameter in the construct. Something like this:

```python
        if self._step_context and self._step_context.has_asset_partitions:
            self._asset_partition_key_range = self._step_context.asset_partition_key_range_for_output(name)
        else:
            self._asset_partition_key_range = asset_partition_key_range
```

PR #10623, opened 2022-11-18
fix lib minor version in changelog

yuhan:
### Summary & Motivation
spotted this while i was looking for entries in the changelog


sryza (approved):
Thanks for catching!

PR #10631, opened 2022-11-18
[freshness-policies] add freshness policy sensor

OwenKephart:


python_modules/dagster/dagster/_core/definitions/freshness_policy_sensor_definition.py
```diff
@@ -0,0 +1,366 @@
...
+def freshness_policy_sensor(
+    asset_selection: AssetSelection,
+    *,
+    name: Optional[str] = None,
+    minimum_interval_seconds: Optional[int] = None,
+    description: Optional[str] = None,
+    default_status: DefaultSensorStatus = DefaultSensorStatus.STOPPED,
+) -> Callable[[Callable[[FreshnessPolicySensorContext], None]], FreshnessPolicySensorDefinition,]:
+    """
+    Define a sensor that reacts to the status of a given set of asset freshness policies, where the
```
sryza, line 330: I kind of expected this to only get called when something became late. Are there non-late situations where you imagine this would get used? What would a canonical implementation look like?

OwenKephart: I was imagining a pattern like "alert a single time once I start being late, then alert a single time once I'm back to ok".

OwenKephart:
```
@freshness_policy_sensor(
    asset_selection=AssetSelection.all(),
)
def slack_on_freshness_policy_status(context: FreshnessPolicySensorContext):
    if context.minutes_late > 10 and context.previous_minutes_late < 10:
        # send failure slack message
        pass
    if context.minutes_late == 0 and context.previous_minutes_late != 0:
        # send success slack message
        pass
```

sryza (approved):

PR #10651, opened 2022-11-20
Propagate environment variables through container overrides on ECSRunLauncher

aroig:
### Summary & Motivation

When using ECSRunLauncher with a fixed task_definition via config, environment variables set through env_vars config are not propagated to worker ecs task. It behaves as expected if task definition is synthesized from launching task, as the variables are included only in the task definition.

This change sets the propagated environment variables also via containerOverrides on `run_task`.

Ran dag on our internal deployment with `dagster.yaml` containing something like this:

```yaml
run_launcher:
  module: dagster_aws.ecs
  class: EcsRunLauncher
  config:
    env_vars:
      - A_VERY_IMPORTANT_VARIABLE
    task_definition:
      env: WORKER_TASK_DEFINITION
```


sryza:
@johannkm @gibsondan - would one of you be the right person to look at this PR?

cmpadden:

PR #10655, opened 2022-11-20
[definitions-stack-1] Add Definitions class entry point

schrockn:
Summary & Motivation:

Repository and workspace are duplicate and confusing concepts in Dagster. This is the beginning of an effort to eliminate them and consolidate under the concept of a code location.

This introduces a new `Definitions` class that can be used an as entry point in dagster's tool. Dagit, the CLI, and any other tools discover in instance of this at the scope of the module they are loading. The variable must be named `defs`. There can only be one of these loaded, so the notion of having multiple repositories in a single code location goes away.

Importantly this also makes resources a top-level concept rather than something that must be threaded to all individual definitions. This diff does that for assets currently. Later we will apply that to jobs. In the short-term it eliminates the need for with_resources calls to assets.

This also makes the top-level API typed and discoverable, rather than the current `@repository` API which returns an array of heterogenous objects that are poorly documented.

TODO, link to public github discussion on this.

Test Plan: Included unit tests


python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -0,0 +1,92 @@
...
+from .resource_definition import ResourceDefinition
+from .schedule_definition import ScheduleDefinition
+from .sensor_definition import SensorDefinition
+
+
+@experimental
+class Definitions:
+    def __init__(
+        self,
+        assets: Optional[
```
OwenKephart, line 24: Historically, we've allowed for nested lists to be returned from the repository function, to make it more ergonomic to do things like

```
assets = [
    load_assets_from_dbt_project(...), # returns a list
    some_asset,
    another_asset,
]
```

without having to splat everything out.

if you ignore the type here, I think that will still work with this class, but it would probably be good to either take the hard-line stance that this list should be pre-flattened or update the type.

python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -0,0 +1,92 @@
...
+        Definitions provides a few conveniences for dealing with resources that do not apply to
+        vanilla dagster definitions:
+
+        (1) It takes a dictionary of top-level resources which are automatically bound
+        (via with_resources) to any asset passed to it. If you need to apply different
+        resources to different assets, use legacy @repository and use with_resources as before.
+
+        (2) The resources dictionary takes raw python objects, not just resource definitions.
+        """
+
```
OwenKephart, line 66: should we type-check immediately to avoid harder-to understand errors from inside created_repo (if you pass in a bad object type)?

OwenKephart (commented):
Had some small comments, but overall I'm a big fan of this idea. the jumble of random objects returned from a repository was always a bit of a tough sell in demos (I always just sort of avoided talking about it). this feels a lot cleaner.

python_modules/dagster/dagster/_core/definitions/definitions_class.py
schrockn: IMO we should just insist on pre-flattening to make the type signature reasonable

OwenKephart (approved):
🚢

PR #10656, opened 2022-11-20
[definitions-stack-2] Make CLI tools bootable from pyproject.toml

schrockn:
Summary & Motivation: This allows our CLI tools to load from a package specified in the pyproject.toml. This obviates the need for workspace.yaml in the common case where it's only purpose is to make dagster code loadable without having to specify command line arguments.

Test Plan:
Unit tests and manually loaded quickstart repo in downstream PR https://github.com/dagster-io/dagster/pull/10654


python_modules/dagster/dagster/_cli/workspace/cli_target.py
```diff
@@ -73,15 +73,43 @@ def are_all_keys_empty(kwargs: Mapping[str, str], keys: Iterable[str]) -> bool:
     "grpc_socket",
 )
 
+import tomli
```
OwenKephart, line 76: nit: any reason for this import to be down here?

OwenKephart (approved):
looks good, one nit

python_modules/dagster/dagster/_cli/workspace/cli_target.py
schrockn: nope. fixed. good catch.

PR #10660, opened 2022-11-21
Allow overriding Nav items

salazarm:
### Summary & Motivation

We want to override some of the nav-items in cloud.

Used this in cloud

Screen Shot 2022-11-21 at 7 52 56 AM


js_modules/dagit/packages/core/src/app/AppTopNav.tsx
```diff
@@ -11,89 +11,126 @@ import {LayoutContext} from './LayoutProvider';
...
   return (
     <AppTopNavContainer>
       <Box flex={{direction: 'row', alignItems: 'center', gap: 16}}>
         <AppTopNavLogo />
-        <Box margin={{left: 8}}>{navLinks()}</Box>
+        <Box margin={{left: 8}}>
+          <Box flex={{direction: 'row', alignItems: 'center', gap: 16}}>
+            {getNavLinks ? getNavLinks(navLinks()) : navLinks().map((link) => link.element)}
+          </Box>
+        </Box>
```
hellendag, line 133: Does this need to be nested boxes? Could it just be one that has both margin and flex?

js_modules/dagit/packages/core/src/app/AppTopNav.tsx
```diff
@@ -11,89 +11,126 @@ import {LayoutContext} from './LayoutProvider';
...
+            >
+              <Box flex={{direction: 'row', alignItems: 'center', gap: 6}}>
+                Deployment
+                <DeploymentStatusIcon />
+              </Box>
+            </TopNavLink>
+          </ShortcutHandler>
+        ),
+      },
+    ];
```
hellendag, line 122: Since the items are defined statically, another possibility might be to export the nav link objects and optionally specify them at the callsite as a prop, so that we don't need to define the list of items within `AppTopNav` and pass them upward to be manipulated.

```
// from AppTopNav (or adjacent)
export const items = {
  overview: {
    path: '/overview',
    isActive: (_, location) => ...,
    label: ...,
  },
  // etc.
};

// callsite with custom items
const itemsToUse = [items.overview, items.runs, items.assets, ...];
<AppTopNav items={itemsToUse} ... />

// callsite with default items
<AppTopNav ... />

// cloud callsite
const mySetupItem = {...};
const itemsToUse = [mySetupItem, items.runs, items.assets, ...];
<AppTopNav items={itemsToUse} ... />
```

This might make it a little easier to reorder items as needed, or swap one in for another. Then in `AppTopNav`, iterate over the items and render the `TopNavLink`, using `history` to set the shortcut behavior, setting the appropriate shortcut label/digit based on the index in the array, applying the data-cy className, etc.

hellendag (approved):

PR #10671, opened 2022-11-21
strip empty prefix parts for cloud-based compute log manager

prha:
### Summary & Motivation
Some cloud storage clients get confused with bonkers paths like `s3://bucket/foo//bar.err`.

This change sanitizes the custom prefix by stripping out trailing-slashes, or infix `//`.

BK

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws_tests/s3_tests/test_compute_log_manager.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws_tests/s3_tests/test_compute_log_manager.py
@@ -187,6 +187,22 @@ def test_blank_compute_logs(mock_s3_bucket):
         assert not stderr.data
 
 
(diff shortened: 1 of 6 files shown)


python_modules/libraries/dagster-aws/dagster_aws_tests/s3_tests/test_compute_log_manager.py
```diff
@@ -187,6 +187,22 @@ def test_blank_compute_logs(mock_s3_bucket):
         assert not stderr.data
 
 
+def test_prefix_filter(mock_s3_bucket):
+    s3_prefix = "foo/bar/"  # note the trailing slash
```
alangenfeld, line 191: worth adding any other cases ?

alangenfeld (approved):

PR #10672, opened 2022-11-21
[dagit] Show ellipsis on the Materialize button for partitioned assets

bengotow:
### Summary & Motivation

This PR adds ellipsis to the launch asset modal if one or more partitioned assets is in the selection. It also makes a few changes:

1) I cleaned up the props of `LaunchAssetMaterializationButton`

2) I made the button handle the `isSource` check and the determination of the "stale & missing", because I didn't like having "isSource" called in a lot of places.

3) I added an `AssetNode.isPartitioned` resolver because I didn't want to ask for `asset.partitionDefinition` for every asset in the graph. Just getting the typename wouldn't be bad, but on the backend it inflates a whole object and stringifies a description, etc. even if you just ask for the typename.

Results:

image
image

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
+++ b/js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
@@ -8,7 +8,7 @@ import {showCustomAlert} from '../app/CustomAlertProvider';
 import {useConfirmation} from '../app/CustomConfirmationProvider';
 import {IExecutionSession} from '../app/ExecutionSessionStorage';
 import {usePermissions} from '../app/Permissions';
-import {displayNameForAssetKey} from '../asset-graph/Utils';
+import {displayNameForAssetKey, LiveData, toGraphId} from '../asset-graph/Utils';
 import {useLaunchWithTelemetry} from '../launchpad/LaunchRootExecutionButton';
 import {AssetLaunchpad} from '../launchpad/LaunchpadRoot';
 import {DagsterTag} from '../runs/RunTag';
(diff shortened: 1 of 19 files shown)


js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
```diff
@@ -60,39 +61,72 @@ type LaunchAssetsState =
...
+    ];
+  }
+
+  const options: LaunchOption[] = [];
+  const assets = scope.all.filter((a) => !a.isSource);
+
+  options.push({
+    assetKeys: assets.map((a) => a.assetKey),
+    label:
+      assets.length > 1 && !scope.skipAllTerm
```
bengotow, line 98: `skipAllTerm` is a bit of a hack, but the problem is that on the Partitions view "All" seems like it's talking about all partitions, not all assets. In this /one/ case, Josh and I want to hide All 🫣

hellendag (approved):

js_modules/dagit/packages/core/src/assets/LaunchAssetObservationButton.tsx
```diff
@@ -112,9 +111,9 @@ export const LaunchAssetObservationButton: React.FC<{
 };
 
 async function stateForObservingAssets(
-  client: ApolloClient<any>,
+  _client: ApolloClient<any>,
   assets: LaunchAssetExecutionAssetNodeFragment[],
-  forceLaunchpad: boolean,
+  _forceLaunchpad: boolean,
```
bengotow, line 116: Flagging these for future cleanup bit didn't want to do it here

PR #10680, opened 2022-11-21
dagster-airbyte: make job cancellation on termination optional

adam-bloom:
### Summary & Motivation
Our dagster jobs run in an autoscaling node group in EKS. Occasionally, the dagster job pod will be killed while downscaling. Our airbyte workers run without autoscaling (to ensure syncs are not interrupted, among other reasons). Rather than killing the airbyte job when a dagster pod terminates mid-sync, we'd like the airbyte job to complete. This, of course, won't trigger asset materializations or other downstream jobs, but it will ensure we don't lose data and can easily pick up the next time that dagster triggers this job.

To accomplish this, I've added a new config parameter to the airbyte resource. It's definitely a long name, but I'm not sure how to shorten it without losing clarity. @benpankow - any suggestions?

Updated unit tests to validate that the airbyte API cancel route isn't called when a job abnormally terminates, if the option is disabled. Additionally added validation that it is called when enabled (default).

### DIFF
--- a/python_modules/libraries/dagster-airbyte/dagster_airbyte/resources.py
+++ b/python_modules/libraries/dagster-airbyte/dagster_airbyte/resources.py
@@ -45,6 +45,7 @@ class AirbyteResource:
         request_additional_params: Optional[Mapping[str, Any]] = None,
         log: logging.Logger = get_dagster_logger(),
         forward_logs: bool = True,
+        cancel_airbyte_sync_on_abnormal_termination: bool = True,
         username: Optional[str] = None,
         password: Optional[str] = None,
     ):
@@ -65,6 +66,10 @@ class AirbyteResource:
 
         self._username = username
         self._password = password
+
+        self._cancel_airbyte_sync_on_abnormal_termination = (
+            cancel_airbyte_sync_on_abnormal_termination
+        )
 
     @property
     def api_base_url(self) -> str:
@@ -334,7 +339,10 @@ class AirbyteResource:
         finally:
             # if Airbyte sync has not completed, make sure to cancel it so that it doesn't outlive
             # the python process
-            if state not in (AirbyteState.SUCCEEDED, AirbyteState.ERROR, AirbyteState.CANCELLED):
+            if (
+                state not in (AirbyteState.SUCCEEDED, AirbyteState.ERROR, AirbyteState.CANCELLED)
+                and self._cancel_airbyte_sync_on_abnormal_termination
+            ):
                 self.cancel_job(job_id)
 
         return AirbyteOutput(job_details=job_details, connection_details=connection_details)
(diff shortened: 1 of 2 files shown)


sryza:
@benpankow - mind taking a look?

python_modules/libraries/dagster-airbyte/dagster_airbyte/resources.py
```diff
@@ -45,6 +45,7 @@ def __init__(
         request_additional_params: Optional[Mapping[str, Any]] = None,
         log: logging.Logger = get_dagster_logger(),
         forward_logs: bool = True,
+        cancel_airbyte_sync_on_abnormal_termination: bool = True,
```
benpankow, line 48: I think we can remove `airbyte` from the param name since it's implicit from the resource, e.g. `cancel_sync_on_abnormal_termination`

benpankow: I believe this will also affect the behavior of manually cancelled runs via the UI, so maybe dropping the `abnormal` makes sense too? `cancel_sync_on_run_termination` seems reasonable

python_modules/libraries/dagster-airbyte/dagster_airbyte_tests/test_resources.py
```diff
@@ -396,3 +402,7 @@ def test_sync_and_poll_timeout(forward_logs):
     timeout = 1
     with pytest.raises(Failure, match="Timeout: Airbyte job"):
         ab_resource.sync_and_poll("some_connection", poll_wait_second, timeout)
+        if cancel_airbyte_sync_on_abnormal_termination:
```
benpankow, line 405: Might be worth adding a similar test which simulates other cancellation scenarios, e.g.

```python
...

# Simulate job interruption when we poll for job status
def callback(*_, **__):
    raise DagsterExecutionInterruptedError()

responses.add_callback(
    method=responses.POST, 
    url=ab_resource.api_base_url + "/jobs/get",
    callback=callback
)

responses.add(responses.POST, f"{ab_resource.api_base_url}/jobs/cancel", status=204)
poll_wait_second = 2
timeout = 1
with pytest.raises(DagsterExecutionInterruptedError):
    ab_resource.sync_and_poll("some_connection", poll_wait_second, timeout)
    if cancel_airbyte_sync_on_abnormal_termination:
        ...

```

benpankow (approved):
Looks good to me! Left a few comments inline re: naming and tests

python_modules/libraries/dagster-airbyte/dagster_airbyte/resources.py
adam-bloom: I like this name much better - addressed in 1560855

python_modules/libraries/dagster-airbyte/dagster_airbyte_tests/test_resources.py
adam-bloom: Added in 1560855

adam-bloom:
@benpankow anything else to address or is this good to go?

benpankow:
> @benpankow anything else to address or is this good to go?

Looks good to me! will kick off a buildkite test run and then will merge

Issue #10659, opened 2022-11-21
pass kwargs to multi_asset


lilinxi:
### Dagster version

1.0.15

### What's the issue?

i defined my asset like this. and the many_asset_list contains hundreds of parameters.
```
 @multi_asset( 
        ins={
            **{
                k: AssetIn() for k in many_asset_list
            }
        },
    )
    def my_asset(
            context: OpExecutionContext,
            **kwargs,
    ) -> Generator:
      pass
```

and write a test like this, 

```
output_generator = my_asset.op(
                context,
                k1=v1,
                k2=v2,
            )
```

and got this error

```
E           dagster._core.errors.DagsterInvalidInvocationError: Too many input arguments were provided for op 'swings_dataset_sehk'. This may be because an argument was provided for the context parameter, but no context parameter was defined for the solid.

.tox/dev/lib/python3.8/site-packages/dagster/_core/definitions/solid_invocation.py:141: DagsterInvalidInvocationError
```

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

sryza:
hey @lilinxi - can you confirm that "k1" and "k2" are among the inputs for the asset?

I just tried writing this test and it passed:

```python
def test_kwargs_multi_asset_with_context():
    @multi_asset(ins={"upstream": AssetIn()}, outs={"a": AssetOut()})
    def my_asset(context, **kwargs):
        return (7,)

    assert isinstance(my_asset, AssetsDefinition)
    assert len(my_asset.op.output_defs) == 1
    assert len(my_asset.op.input_defs) == 1
    assert AssetKey("upstream") in my_asset.keys_by_input_name.values()
    assert my_asset(build_op_context(), upstream=5) == (7,)
    assert my_asset.op(build_op_context(), upstream=5) == (7,)
```

sryza:
Going to close this because the tests pass, but feel free to reopen if you still think there's broken functionality.

PR #10682, opened 2022-11-22
add tests for asset and op decorators with kwargs and context

sryza:
### Summary & Motivation

Motivated by this issue: https://github.com/dagster-io/dagster/issues/10659


dpeng817 (changes requested):
Should add tests that actually materialize to ensure that behavior is the same across invocation + materialization

sryza:
@dpeng817 this PR is only adding tests - there are no behavior changes. I agree that those additional tests would be helpful, but I don't think we need to block new test coverage on even more test coverage.

dpeng817 (changes requested):
Apologies for the long cycle on this. I understand that the PR is just adding tests, I meant that within the context of each of these tests, it would be good to test feature congruency with actual execution. IE, testing that invocation works correctly is not enough, because there is a completely separate code path to normal execution.

dpeng817 (approved):
will accept bc more tests better than less tests

PR #10685, opened 2022-11-22
avoid setting the asset reconciliation cursor to none when there are …

sryza:
…no new events

### Summary & Motivation

We got a bug report that duplicate runs were being launched in some situations when a downstream asset has different partitions than the upstream asset.

#### Unit test
Without this PR, the second test I added fails. The first one succeeds either way, but added it as a sanity check.

#### Manual test
Ran the following and observed duplicate runs of asset1 get launched. Afterward the fix, only a single run would get launched per partition

```python
from dagster import (
    asset,
    repository,
    StaticPartitionsDefinition,
    AssetSelection,
    build_asset_reconciliation_sensor,
    load_assets_from_current_module,
)

import time


@asset
def upstream():
    ...


@asset(partitions_def=StaticPartitionsDefinition(["a", "b", "c"]))
def asset1(context, upstream):
    context.asset_partition_key_for_output()
    time.sleep(90)


@repository
def repo():
    return [
        load_assets_from_current_module(),
        build_asset_reconciliation_sensor(AssetSelection.all()),
    ]
```

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -122,8 +122,14 @@ class AssetReconciliationCursor(NamedTuple):
             | requested_non_partitioned_root_assets
         )
 
+        if latest_storage_id and self.latest_storage_id:
+            check.invariant(
+                latest_storage_id >= self.latest_storage_id,
+                "Latest storage ID should be >= previous latest storage ID",
+            )
+
         return AssetReconciliationCursor(
-            latest_storage_id=latest_storage_id,
+            latest_storage_id=latest_storage_id or self.latest_storage_id,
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -123,7 +123,7 @@ def with_updates(
         )
 
         return AssetReconciliationCursor(
-            latest_storage_id=latest_storage_id,
+            latest_storage_id=latest_storage_id or self.latest_storage_id,
```
clairelin135, line 126: I know the code currently returns `latest_storage_id=None` or `latest_storage_id >= self.latest_storage_id`, but thoughts on making this a `max` function to guard against future code changes/improve readability?

clairelin135 (approved):

PR #10690, opened 2022-11-22
Fix typing and missing asserts [fix-typing-in-test-autodiscovery]

schrockn:
Summary & Motivation: While using this file for another reason I noticed typing errors and missing asserts. This fixes that.

Test Plan: BK


OwenKephart (approved):
thank you 🙏

PR #10692, opened 2022-11-22
Don't use an inclusive BETWEEN when identifing partitions

jmsanders:
Snowflake's BETWEEN is inclusive meaning we're deleting the _next_ partition instead of everything up until the next partition.

https://docs.snowflake.com/en/sql-reference/functions/between.html


python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
```diff
@@ -154,4 +154,6 @@ def _time_window_where_clause(table_partition: TablePartition) -> str:
     start_dt, end_dt = table_partition.time_window
     start_dt_str = start_dt.strftime(SNOWFLAKE_DATETIME_FORMAT)
     end_dt_str = end_dt.strftime(SNOWFLAKE_DATETIME_FORMAT)
-    return f"""WHERE {table_partition.partition_expr} BETWEEN '{start_dt_str}' AND '{end_dt_str}'"""
+    # Snowflake BETWEEN is inclusive; start <= partition expr <= end. We don't want to remove the next partition so we instead
```
jmsanders, line 157: We could also maybe parametrize this to allow you to choose the exclusivity? But I think this more accurately captures the intended use case.

PR #10693, opened 2022-11-22
[dagit] Enable “groups-only” zoom level with auto-minimum zoom

bengotow:
### Summary & Motivation

This PR slightly adjusts the transition threshold between "full asset nodes" and "mini asset nodes" so the mini nodes appear more quickly (per https://linear.app/elementl/issue/DAGIT-203/flip-to-zoomed-out-asset-nodes-earlier)

I also removed the feature flag for the "experimental groups-only" zoom level and enabled it for everyone. The main blocker around this was that it let you zoom out /too far/, so I adjusted the SVGViewport behavior slightly so that passing a param allows you to enable minimum zoom out to "whatever is required to see the entire DAG in view". (We were always using the default min zoom before)

Sidenote:
After this is reviewed, I might rename a few references to minZoom / minScale / etc. to use the same term. I think we flip-flop between zoom and scale and they're the same. Just didn't want to clutter up review!

- Large and small op and asset DAGs


hellendag (approved):

bengotow:
Thanks for the inline edits!

PR #10699, opened 2022-11-22
fix logical versions for intrastep asset deps

smackesey:
### Summary & Motivation

Current behavior of logical versions does not properly handle assets with intrastep dependencies. This is most salient for `dagster-dbt`, where an entire graph of dbt-defined assets may be materialized in a single step. This adjusts the LV logic so that:

- individual assets from a single step properly store only their asset dependencies (as opposed to external asset deps for the step) in their provenance
- during generation of materialization events and recording of provenance, materialization records generated within the same step are properly retrieved and recorded in provenance

Note that for this to work, materializations from an asset graph within a single step need to be generated in topological order.

Manually + existing BK tests.

Manual tests were done to prevent regressions from previously fixed bugs. This is a stopgap until more automated tests can be added. Manual flow:

- Load `assets_reconciliation_example` in dagit (currently not in master-- you can find it in this PR: https://github.com/dagster-io/dagster/pull/10004. Perform following steps and ensure stale status updated as expected:
        - Materialize `baz`
        - Materialize `alpha`, `beta`, `delta`, `gamma`, `rho`, `epsilon`
        - Observe all sources
        - Materialize epsilon (still stale)
        - Materialize alpha, beta, delta (epsilon still stale)
        - Materialize epsilon (epsilon no longer stale)
- Load two repos with same name.
    - Navigate to global asset lineage.
    - Materialize some assets.
    - Observe no GQL errors
 - Load `assets_dbt_python` example
     - Materialize all
     - Observe stale status appropriately updated and no GQL errors.
     - Materialize `order_stats` (a dbt asset).
     - Observe staleness of downstream appropriately updated.


smackesey:
* master
  * **PR #10699** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10699" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10699?utm_source=stack-comment).

sryza (approved):
Are you able to include tests for this?

sryza (changes requested):
Accidentally hit "Approve", but I think we should include tests before merging

python_modules/dagster/dagster/_core/execution/context/system.py
```diff
@@ -806,7 +806,13 @@ def step_materializes_assets(self) -> bool:
...
         return self._input_asset_records
 
-    def fetch_input_asset_records(self) -> None:
+    def get_input_asset_record(self, key: AssetKey) -> Optional["EventLogRecord"]:
+        if not key in self._input_asset_records:
+            self._fetch_input_asset_record(key)
+        return self._input_asset_records[key]
+
+    # "external" refers to records for inputs generated outside of this step
+    def fetch_external_input_asset_records(self) -> None:
```
OwenKephart, line 815: I think the logic here is still slightly off, as it assumes that any asset upstream of an asset produced by this step is "external".

Take for example a multi-asset that has the following internal dependency structure:

```
    a
  /   \
 b     c
```

and let's imagine "`a`" depends on "`a'`" (produced by a different step).

When we run this function to materialize a, b, and c, upstream_assets_for_asset("b")/upstream_assets_for_asset("c") will both return AssetKey("a"). This means that self._input_asset_records[AssetKey("a")] will be populated when this function is called, and so it'll have an old version, rather than the version created when running this step.

however, if we run this function to materialize just b and c, then the behavior described above is actually correct, as you don't expect to get a new version of a, so you should query for whatever the old version was.

the asset_info returned by asset_info_for_output has an "is_required" property, which will tell you if, for the current execution, that asset key is required to be executed. (generally, if is_required is `False`, then we do not expect it to be generated, although we'll just emit a warning rather than an error if it is.

I think this information should be used to get a set of required_output_keys (rather than all output keys), but mostly I'm just advocating for tests of these sorts of scenarios

OwenKephart (approved):

sryza (approved):

PR #10706, opened 2022-11-23
output-scoped code versions

### LINKED ISSUE #10728: allow setting granular versions within subsettable multi-assets

### LINKED ISSUE #10729: rename op_version to code_version

smackesey:
### Summary & Motivation

- Resolves #10728 
- Resolves #10729

Update the code versioning system to scope code versions to individual outputs (as opposed to ops). This allows for ops that wrap external code (e.g. dbt models) to accurately represent versios in this external code. This in turn allows the logical versioning system to more accurately represent the staleness of assets defined in these external models. Changes include:

- Addition of `code_version` to `OutputDefinition` and associated abstractions
- Modification of `AssetLayer` to correctly map asset keys to output-scoped code versions
- Deprecation of the `version` property on `OpDefinition`-- a version passed to the constructor is merely a default for the code version set on `OutputDefinition`, so there's no need to store it on `OpDefinition` itself.
- Deprecation of the `version` arg wherever it is used to represent a code version, to be replaced with `code_version`.
- Remove old `is_versioned` property from `AssetsDefinition` (should've been removed in a previous PR)

#### Questions/Concerns

- There are a couple places that `AssetOutputInfo` is constructed to put on `OutputContext`. I did not modify these sites to pass a code version-- these `AssetOutputInfo` instances already seem incomplete, not sure what they're for.
- Doesn't yet include modifications to old memoization code that uses op versions (nonetheless, it passes tests-- I think coverage is weak there).

Added/modified unit tests.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/implementation/loader.py
+++ b/python_modules/dagster-graphql/dagster_graphql/implementation/loader.py
@@ -486,7 +486,7 @@ class ProjectedLogicalVersionLoader:
                 )
             else:
                 version = DEFAULT_LOGICAL_VERSION
-        elif node.op_version is not None:
+        elif node.code_version is not None:
             version = self._compute_projected_new_materialization_logical_version(node)
         else:
             materialization = self._instance.get_latest_materialization_event(key)
@@ -519,7 +519,7 @@ class ProjectedLogicalVersionLoader:
     ) -> LogicalVersion:
         dep_keys = {dep.upstream_asset_key for dep in node.dependencies}
         return compute_logical_version(
-            node.op_version or UNKNOWN_VALUE,
+            node.code_version or UNKNOWN_VALUE,
             {dep_key: self._get_version(key=dep_key) for dep_key in dep_keys},
         )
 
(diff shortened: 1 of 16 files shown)


smackesey:
* master
  * **PR #10699** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10699" target="_blank">Graphite</a> 
    * **PR #10706** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10706" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10706?utm_source=stack-comment).

OwenKephart (approved):
So I looked into the bit about manually constructing AssetOutputInfo, and it looks like it's just used when constructing an InputContext/OutputContext. These context objects are used within the IOManager's load_input/handle_output functions, respectively.

The AssetOutputInfo is used to supply values for properties such as `context.asset_key`. The OutputContext does have a `version` property, but it looks like it already has a property for this, and so does not need to get it from the AssetOutputInfo (although it seems like it'd be straightforward to add it in build_output_context, as the version parameter already exists for that function, I don't think it'd ever be accessed).

The input context part is slightly trickier, but ultimately still fine IMO, and not relevant to your changes. Each InputContext that we create is intended to have a `context.upstream_output` value, which should be populated with an `OutputContext`. When we're loading from a SourceAsset, there's no explicit upstream Output, but we still make a fake OutputContext to make it so that user's code works nicely without having to special case scenarios where there's actually an upstream source asset. In these cases, we don't supply a version to the upstream output context, but even if we wanted to do this (not sure if we would), this would be added in the version argument to the OutputContext, rather than in the AssetOutputInfo.

PR #10714, opened 2022-11-23
[dagstermill] save notebook on failure

jamiedemaria:
### Summary & Motivation
Adds the option to save the notebook to a storage location if the notebook fails during execution

new unit tests + manual test in the example repository (materialize the error notebook asset and then confirm that i can open the saved notebook)


python_modules/libraries/dagstermill/dagstermill/asset_factory.py
```diff
@@ -122,6 +135,7 @@ def define_dagstermill_asset(
     group_name: Optional[str] = None,
     io_manager_key: Optional[str] = None,
     retry_policy: Optional[RetryPolicy] = None,
+    save_notebook_on_failure: bool = False,
```
yuhan, line 138: should this be `Optional[bool]` instead?

python_modules/libraries/dagstermill/dagstermill/asset_factory.py
```diff
@@ -238,5 +255,6 @@ def iris_dataset():
         _dm_compute(
             name=name,
             notebook_path=notebook_path,
+            save_notebook_on_failure=save_notebook_on_failure,
```
yuhan, line 258: prob wanna capture the invalid type early on by `save_notebook_on_failure=check.opt_bool_param(save_notebook_on_failure)`

python_modules/libraries/dagstermill/dagstermill/asset_factory.py
jamiedemaria: what are you thinking would be the behavior if it's None?

yuhan: `opt_bool_param` should convert None to false by default

yuhan: so the idea is user-facing wise, they can omit this arg but once we validate user inputs, all the args to `_dm_compute` should be not null.

python_modules/libraries/dagstermill/dagstermill/asset_factory.py
```diff
@@ -205,6 +222,7 @@ def iris_dataset():
 
     check.str_param(name, "name")
     check.str_param(notebook_path, "notebook_path")
+    check.opt_bool_param(save_notebook_on_failure, "save_notebook_on_failure")
```
yuhan, line 225: can do `save_notebook_on_failure = check.opt_bool_param(save_notebook_on_failure, "save_notebook_on_failure")` so when `save_notebook_on_failure` is not provided by the user, we fill it in as `false`

python_modules/libraries/dagstermill/dagstermill/asset_factory.py
yuhan: hmm i was mistaken `bool = False` and `Optional[bool] = False` both would work.

yuhan (approved):

PR #10715, opened 2022-11-23
[freshness-policies] Improved docs coverage

OwenKephart:
### Summary & Motivation

Add some freshness-policy related stuff to API docs, add extra information on the freshness_policy_sensor to the sensors page, explain freshness-based scheduling.

### DIFF
--- a/docs/content/concepts/partitions-schedules-sensors/asset-sensors.mdx
+++ b/docs/content/concepts/partitions-schedules-sensors/asset-sensors.mdx
@@ -58,14 +58,14 @@ Multi-asset sensors, which can trigger job executions based on multiple asset ma
 In the body of the sensor, you have access to the materialization event records for each asset via the <PyObject object="MultiAssetSensorEvaluationContext" /> context object. Methods within the context object will search for events after the [cursor](/concepts/partitions-schedules-sensors/sensors#sensor-optimizations-using-cursors).
 
 - <PyObject
-  object="MultiAssetSensorEvaluationContext"
-  method="latest_materialization_records_by_key"
+    object="MultiAssetSensorEvaluationContext"
+    method="latest_materialization_records_by_key"
   /> returns a dictionary mapping the asset key for each monitored asset to the most
   recent materialization record. If there is no materialization event, the mapped
   value will be none.
 - <PyObject
-  object="MultiAssetSensorEvaluationContext"
-  method="materialization_records_for_key"
+    object="MultiAssetSensorEvaluationContext"
+    method="materialization_records_for_key"
   /> returns a list of materialization event records for a specified asset key.
 
 ```python file=/concepts/partitions_schedules_sensors/sensors/asset_sensors.py startafter=start_multi_asset_sensor_marker endbefore=end_multi_asset_sensor_marker
@@ -112,8 +112,8 @@ def asset_a_and_b_sensor_with_skip_reason(context):
 You can use the <PyObject object="MultiAssetSensorEvaluationContext" /> context object to fetch partitioned materializations for each monitored asset.
 
 - <PyObject
-  object="MultiAssetSensorEvaluationContext"
-  method="latest_materialization_records_by_partition"
+    object="MultiAssetSensorEvaluationContext"
+    method="latest_materialization_records_by_partition"
   /> returns an mapping of partition key to the most recent materialization event
   record for that partition. Returns records in order by event ID.
 - If all monitored assets have the same partitions definition, <PyObject
@@ -190,6 +190,8 @@ def trigger_daily_asset_when_any_upstream_partitions_have_new_materializations(c
   of updating a weekly asset when upstream daily assets are materialized.
 </Note>
 
+## Freshness-based scheduling <Experimental />
+
 ## Asset reconciliation sensor <Experimental />
 
 Dagster provides an out-of-box sensor that will monitor a provided set of assets and automatically materialize them if they're "unreconciled". An asset is considered "unreconciled" if any of:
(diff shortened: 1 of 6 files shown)


docs/content/concepts/partitions-schedules-sensors/asset-sensors.mdx
```diff
@@ -218,53 +242,18 @@ def repository_1():
 
 This sensor will materialize `asset_b` whenever any of its parents materialize, and it will materialize `asset_c` whenever any of its parents materialize.
 
-## Examples
+### Freshness-based scheduling <Experimental />
 
-### Monitoring daily assets to materialize a weekly asset
```
clairelin135, line 223: I'd like to keep this example since users have asked for it

clairelin135 (changes requested):
The content here you added looks good to me! I added a couple small nits in the vercel preview. Main change is that I'd prefer to keep the daily -> weekly example that has been deleted in this PR since users have asked for it

docs/content/concepts/partitions-schedules-sensors/asset-sensors.mdx
OwenKephart: Hm can they just use build_asset_reconciliation_sensor in that case? (that's why I removed it but I'll admit I'm not the most familiar w/ this part of the world)

clairelin135: Code snippet of what the "AND" case would look like (I tested to confirm this works):
```
@multi_asset_sensor(
    asset_keys=[AssetKey("upstream_daily_1")], job=weekly_asset_job, minimum_interval_seconds=5
)
def trigger_weekly_asset_from_daily_asset(context):
    run_requests = []
    materializations_by_partition = context.latest_materialization_records_by_partition(
        AssetKey("upstream_daily_1")
    )

    materializations_by_downstream_weekly_partition = defaultdict(list)
    # Get all corresponding weekly partitions for any materialized daily partitions
    for partition, materialization in materializations_by_partition.items():
        # Get the corresponding weekly partition, None if no weekly partition exists
        downstream_partitions = context.get_downstream_partition_keys(
            partition,
            from_asset_key=AssetKey("upstream_daily_1"),
            to_asset_key=AssetKey("downstream_weekly_asset"),
        )
        if downstream_partitions:
            # A daily partition can only have at most one corresponding weekly partition
            materializations_by_downstream_weekly_partition[downstream_partitions[0]].append(
                materialization
            )

    for (
        downstream_weekly_partition,
        daily_materializations,
    ) in materializations_by_downstream_weekly_partition.items():
        daily_partitions_in_week = context.get_downstream_partition_keys(
            downstream_weekly_partition,
            from_asset_key=AssetKey("downstream_weekly_asset"),
            to_asset_key=AssetKey("upstream_daily_1"),
        )
        if all(
            [
                daily_partition in materializations_by_partition
                for daily_partition in daily_partitions_in_week
            ]
        ):
            run_requests.append(
                weekly_asset_job.run_request_for_partition(downstream_weekly_partition)
            )
            for daily_materialization in daily_materializations:
                # Advance the cursor so we only check event log records past the cursor
                context.advance_cursor({AssetKey("upstream_daily_1"): daily_materialization})
    return run_requests
```

OwenKephart: ah perfect -- thanks for the code snippet. I'll add this back in and add a note about how this relates to the reconciliation sensor logic

OwenKephart: I actually ended up keeping the original snippet and just explaining that it did the same thing as the build_asset_reconciliation_sensor, and was more of a jumping-off point

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -752,7 +752,7 @@ def build_asset_reconciliation_sensor(
     - This sensor has never tried to materialize it and it has never been materialized.
     - Any of its parents have been materialized more recently than it has.
     - Any of its parents are unreconciled.
-    - It is not currently up to date with respect to its FreshnessPolicy
+    - It is not currently up to date with respect to its FreshnessPolicy.
```
erinkcochran87, line 755: Should this be `freshness policy`? The example asset graph uses this formatting. I'm not sure if this instance refers to the name of a specific property/attribute, or just a freshness policy in general.

erinkcochran87 (approved):
Two small comments from me, which might not be worth addressing atm. The new copy looks good tho!

clairelin135 (approved):
Included 2 small formatting nits in the vercel preview, but this is looking good to me!

docs/content/concepts/partitions-schedules-sensors/asset-sensors.mdx
```diff
@@ -218,9 +248,32 @@ def repository_1():
...
+
+- A timestamp of the upstream data that is incorporated into the current state of the asset. This is calculated by finding the most recent materialization of that asset, then recursively finding the most recent materializations of its parents that happened before that event. The minimum timestamp found this way will be used as the current data time.
+- Any in-progress runs that will update that asset in the near future. If an in-progress run will update an asset to an acceptable data time, then a new run will not be kicked off.
+- The constraints specifying what upstream data must be incorporated into the state of an asset by certain times. If waiting some amount of time to kick off a materialization of an asset will allow multiple constraints to be satisfied at once, then the asset will not be materialized on that tick.
+
+The sensor will only kick off runs for assets specified in the `asset_selection` argument of `build_asset_reconciliation_sensor`. If an asset in this selection is downstream of an asset outside of the selection, the sensor will not kick off a run of the selected asset until a materialization event for the upstream asset is created.
```
slopp, line 265: I don't follow this - how would the materialization event of the upstream asset be created? Would the freshness policy be violated if the upstream is never updated?

slopp:
This may not be the goal of this PR, but in addition to concept and API docs I think a guide with example code would be really helpful. After reading the "freshness scheduling" concept documentation I'm having a hard time mapping that to the code I would write. 

- It is easy to miss that I need to include an asset reconciliation sensor, and that sensor needs to be on! The asset selection piece of that sensor is still confusing to me, do I select all assets, assets with freshness policies, some combo?
- Its not clear to me how source assets are handled, and whether an observe_fn is needed?
- I'm not sure how freshness policies interact with asset versioning, and whether or not asset versioning is automatic or an additional concept that needs to be implemented?
- I'm not sure how these policies interact with assets loaded from other integrations

I think a really effective approach for a guide would be starting with the goal "the executive kpis table needs to be no more than 30 minutes out of date" and then showing all the steps necessary to do it including defining some simple assets. I'd be very happy to help flush out this type of guide but I am honestly not sure what the code would need to look like after skimming the API docs.

PR #10720, opened 2022-11-23
[dagit] Export WorkspaceOverviewGrid for Cloud

hellendag:
### Summary & Motivation

Export WorkspaceOverviewGrid to make it directly available for Cloud, to move the Workspace tab under Deployment. This is temporary until we have the new Code Locations table in place.

View Deployment in OSS Dagit, verify that Workspace renders correctly under Deployment. With this change in place, verify that I can do the same with a corresponding change in place in Cloud.

### DIFF
--- a/js_modules/dagit/packages/core/src/workspace/WorkspaceOverviewWithGrid.tsx
+++ b/js_modules/dagit/packages/core/src/workspace/WorkspaceOverviewWithGrid.tsx
@@ -22,62 +22,62 @@ import {repoAddressAsString} from './repoAddressAsString';
 import {workspacePath} from './workspacePath';
 
 export const WorkspaceOverviewWithGrid = () => {
-  const {loading, error, options} = useRepositoryOptions();
-
-  const content = () => {
-    if (loading) {
-      return (
-        <Box flex={{direction: 'row', justifyContent: 'center'}} style={{paddingTop: '100px'}}>
-          <Box flex={{direction: 'row', alignItems: 'center', gap: 16}}>
-            <Spinner purpose="section" />
-            <div style={{color: Colors.Gray600}}>Loading workspace…</div>
-          </Box>
-        </Box>
-      );
-    }
-
-    if (error) {
-      return (
-        <Box padding={{vertical: 64}}>
-          <NonIdealState
-            icon="error"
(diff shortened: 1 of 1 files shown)


hellendag:
* master
  * **PR #10720** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10720" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10720?utm_source=stack-comment).

js_modules/dagit/packages/core/src/workspace/WorkspaceOverviewWithGrid.tsx
```diff
@@ -22,62 +22,62 @@ import {repoAddressAsString} from './repoAddressAsString';
 import {workspacePath} from './workspacePath';
 
 export const WorkspaceOverviewWithGrid = () => {
+  return (
```
bengotow, line 25: Hmm, I wonder if it'd make sense to start calling things like this `WorkspaceOverviewWithGridOSS`  to make it more clear that there's an alternative version in cloud... I worry this is the sort of thing I will see in 6mo and think "this abstraction doesn't really seem necessary"  and fold the two components back together next time I need to do something structural to the page.

I wish we could have a "oss export" keyword or a `public` annotation or something :(  Not sure any such thing exists but I'd be nice if renaming or removing WorkspaceOverviewGrid would break the build locally.

Nothing immediately actionable though - will have to think about it a bit! :-/

bengotow (approved):

js_modules/dagit/packages/core/src/workspace/WorkspaceOverviewWithGrid.tsx
hellendag: Yeah, at least in this case if it were refactored away, it would break Cloud TS. But we do need to be careful to test Cloud when making changes to shared/forked parts of the app.

bengotow: Yeah I think that's true... just wish there was some indication of what was a `shared/forked part of the app`.  👍

PR #10721, opened 2022-11-23
[dagit] Rather than just showing groups, show tiny colorized assets when zoomed out

bengotow:
### Summary & Motivation

Instead of a "groups only" zoom level when you zoom really far out on the asset graph, we're exploring a new one that is just the colorized asset nodes. This mode is only available when you have enough assets to need it, and doesn't show any labels.

https://www.loom.com/share/ceeabe070f794e1b980a6154eb99f7fc


sryza:
This is very cool, but I like the simplicity of the pure-group-based zoomout. Just spitballing, but might it make sense to color the groups based on the statuses of the assets inside them?

hellendag (approved):

PR #10740, opened 2022-11-24
[convert-deployment-to-production-3] Convert @repository to Definitions

schrockn:
Summary & Motivation: Use Definitions instead of repository

Test Plan: BK. Load locally.


yuhan (approved):
so much better

PR #10742, opened 2022-11-24
[convert-development-to-production-5] Use ... instead of pass to avoid typing errors

schrockn:
Summary & Motivation: pyright was complaining about this in IDE. `...` tells it to avoid type checking whereas `pass` does not

Test Plan: BK


schrockn:
Abandoning this as this fixes a pyright error but introduces a mypy error

PR #10744, opened 2022-11-24
[convert-development-to-production-7] Convert guide to use Definitions

schrockn:
Summary & Motivation:

This changes the development to production guide to correspond to the new Definitions API
and its improvments. Overall I think this feels pretty good. Sometimes there is a bit of awkwardness
where it says "in the Definitions object" rather than "in the repository". Also we may want to move
Definitions objects to a `defs.py` file instead of a `__init__.py` for clarity.

This depends on a diff stack that changes the underlying code.

Test Plan: Read


erinkcochran87:
@schrockn Looking good! Left a few comments on the Vercel preview.

schrockn:
@erinkcochran87 per your comment about migrating over the env var guide I put up this PR so we can do it independently. https://github.com/dagster-io/dagster/pull/10774. Feel free to push back against that if you feel strongly about not having code duplication in this case

yuhan (commented):
overall lg2m!

erinkcochran87 (approved):
This looks good to me - thanks for taking this on @schrockn 🎉

jamiedemaria:
popping in to add more things to update! we have the reader download the corresponding example project as part of the guide https://github.com/dagster-io/dagster/tree/master/examples/development_to_production. unless changing the example project is part of the larger code change diff stack, it should also get updated so they don't end up with repositories in the example project and definitions in the guide

schrockn:
Yeah that is converted with a stack that ends with https://github.com/dagster-io/dagster/pull/10743/files

I'm still nailing my workflow to make my stacks more clear. Thanks for your patience here.

PR #10746, opened 2022-11-24
Support loading multiple modules in single command line invocation

schrockn:
Summary & Motivation: With workspace.yaml going away we need a way to easily load multiple code locations in a single invocation

Test Plan:
Included tests. Also manually tested dagit loading two examples installed as packages then  running `dagit -m project_fully_featured -m quickstart_etl`


python_modules/dagster/dagster/_cli/api.py
```diff
@@ -650,6 +650,18 @@ def grpc_command(
             "empty_working_directory",
         ]
     ):
+        # in the grpc api CLI we never load more than one module at a time
+        maybe_module = kwargs["module_name"]
```
gibsondan, line 654: You could also enforce this at the click level instead by passing a flag into python_target_click_options depending on the callsite, but that might be a pain

python_modules/dagster/dagster/_cli/workspace/cli_target.py
```diff
@@ -244,7 +270,10 @@ def python_target_click_options():
         click.option(
             "--module-name",
             "-m",
-            help="Specify module where repository or job function lives",
+            multiple=True,
```
gibsondan, line 273: I think we will want to apply this to package-name as well - they used to be (subtly) different but now are identical. I believe the quickstarts all use the equivalence of package-name currently: https://github.com/dagster-io/dagster-cloud-serverless-quickstart/blob/main/workspace.yaml#L2

And the docs recommend package-name as well and call module-name the deprecated one: https://docs.dagster.io/concepts/repositories-workspaces/workspaces#loading-via-python-package

python_modules/dagster/dagster/_cli/api.py
```diff
@@ -650,6 +650,18 @@ def grpc_command(
...
+        if maybe_module:
+            check.is_tuple(maybe_module, of_type=str)
+            check.invariant(
+                len(maybe_module) <= 1,
+                "The dagster grpc server cannot serve more than one code location at a time",
```
gibsondan, line 659: nit gRPC

gibsondan (dismissed):
this looks great to me - I think the main thing is you will want to add package_name as well? Or have some plan for which one is the 'blessed' one going forward

Long ago they were subtly different wrt how they handled the working directory, now they are 100% identical which is unfortunate - I thought we had gravitated more towards recommending package over module

python_modules/dagster/dagster/_cli/api.py
schrockn: Yeah I think this simpler

python_modules/dagster/dagster/_cli/workspace/cli_target.py
schrockn: Ah this is a good discussion to have then.

IMO we should actually standardize on module and then deprecated and eliminate package. I had also noticed that the code was identical was identical at this point.

Python's own documentation is inconsistent on this, but the explanation on the import statement page makes the most sense to me given the behavior we see. https://docs.python.org/3/reference/import.html

Pertinent quote:

_"It’s important to keep in mind that all packages are modules, but not all modules are packages. Or put another way, packages are just a special kind of module. Specifically, any module that contains a __path__ attribute is considered a package."_

Given this we can just say that when we say _module_ we mean the target of an import statement.  I think it is both the easiest to explain and the most flexible.

Issue #3608, opened 2021-01-27
Add private key authentication support to Snowflake


joshuataylor:
## Use Case
Snowflake allows using a private key authentication instead of passwords.
See: https://docs.snowflake.com/en/user-guide/key-pair-auth.html

## Ideas of Implementation
This should be fairly straight forward to implement, how I envision it is as follows:
1. Add the following options to the config (https://docs.dagster.io/_apidocs/libraries/dagster_snowflake)
```
'private_key_path': {'env': 'SNOWFLAKE_PRIVATE_KEY_PASS'},
'private_key_password': {'env': 'SNOWFLAKE_PASSWORD'},
```

OR
```
'private_key': {'env': 'SNOWFLAKE_PRIVATE_KEY'},
'private_key_password': {'env': 'SNOWFLAKE_PASSWORD'},
```

This allows you to either pass a path if it exists in the file system, or pass in the private key as an environment variable (when rotating keys this can be useful for deployments like docker.

## Additional Info
I'm happy to investigate adding this.

---
#### Message from the maintainers:
Excited about this feature? Give it a :thumbsup:. We factor engagement into prioritization.

sryza:
Hey @joshuataylor - agreed that this would be a helpful thing to have.  If you're interested in adding it as you mentioned, I'd be happy to review your PR.

kbrs-miaha:
I wanted to add my support for this feature. An alternative approach is to pass additional parameters to the Snowflake connector itself.

jayhale:
We're currently doing this by extending the Dagster-provided resource class.  Please add this since key-pair is a preferred auth approach over username/password.

PR #10750, opened 2022-11-26
Add private key authentication for Snowflake

joshuataylor:
### Summary & Motivation
Snowflake allows using a private key authentication instead of passwords.
See: https://docs.snowflake.com/en/user-guide/key-pair-auth.html

It would be great to support this, as private key authentication is a great feature.

Closes #https://github.com/dagster-io/dagster/issues/3608

I have tested this against our local dagster instance, using private key auth.


jamiedemaria:
hey @joshuataylor, i'm planning on taking a look at this tomorrow. thanks for the contribution!

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -36,6 +39,7 @@ def __init__(self, config: Mapping[str, str], log):  # pylint: disable=too-many-
         self.connector = config.get("connector", None)
```
jamiedemaria, line 39: We should add a check here that the user has provided one of (password) or (private_key, private_key_path, private_key_password) and throw an exception if neither group is provided. 

You could do this with something like

```
check.invariant(config.get("password", None) is not None or (config.get("private_key", None) is not None and config.get("private_key_path", None)  is not None and config.get("private_key_password", None) is not None), "Missing config: Password or private key authentication required")

```

jamiedemaria: we should probably also check that they dont provide both ways of authenticating

jamiedemaria (changes requested):
Overall this looks good! Left a comment on a part that was a bit confusing to me and to add a check that the user provides one way of authenticating.

I ran the PR in our CI and there are a couple formatting things:

MyPy errors

```


[2022-12-05T18:14:05Z] dagster_snowflake/resources.py:61: error: "Dict[str, Optional[str]]" has no attribute "connect_args"  [attr-defined]
--
  | [2022-12-05T18:14:05Z] dagster_snowflake/resources.py:88: error: Incompatible types in assignment (expression has type "bytes", target has type "Optional[str]")  [assignment]
```

you'll also need to run `make isort` from the top level dagster repo

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -34,6 +36,47 @@ def __init__(self, config: Mapping[str, str], log):  # pylint: disable=too-many-
         # snowflake.connector.connect() because they will override the default values set within the
         # connector; remove them from the conn_args dict.
         self.connector = config.get("connector", None)
+        self.sqlalchemy_engine_args = {}
+
+        # there are three different ways to authenticate with snowflake, we need to ensure that only
+        # one method is provided
+        auths_set = 0
```
jamiedemaria, line 43: Is there a better or more pythonic way to do a check that one and only one of three conditions is met?

jamiedemaria (approved):

jamiedemaria:
@sryza I've made some edits to this PR, so can you do an additional review pass?

PR #10757, opened 2022-11-28
[dagit] Remove tag fetch for RunsFilterInput typeahead

hellendag:
### Summary & Motivation

The query that retrieves the full list of key/value pairs that may be used for tag searching on Runs is potentially super expensive. Since the user can already use freeform typing to add a tag filter to the search, and because tags can be added to search via hover+click on tags shown in the Runs table, let's remove the tag search query altogether.

View Runs, type "tag:", verify that I can complete it with "tag:foo=bar" and successfully find matching runs. Use hover+click on rendered tags in the table, verify that search works with these as well.


hellendag:
* master
  * **PR #10757** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10757" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10757?utm_source=stack-comment).

prha (approved):

PR #10766, opened 2022-11-28
[useStateWithStorage] Remove `validate` as dependency to setter.

salazarm:
### Summary & Motivation

Removing `validate` as a dependency and instead relying on a ref to make sure we have the latest version of the callback.

This avoids a render loop when the `setState` function is a dependency of a `useEffect` that causes the component to rerender (and when `validate` callback reference changes on every render due to lack of `React.useCallback`)

Used this hook In cloud and saw infinite loop disappear.


bengotow (approved):
This seems safe to me!

PR #10775, opened 2022-11-28
handle captured log failures without failing runs

prha:
### Summary & Motivation
Runs are failing with compute log manager failures.

Added exclusive captured log manager implementation with the appropriate exception hooks.  The existing tests only implemented the legacy compute log manager interface

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/plan/execute_plan.py
+++ b/python_modules/dagster/dagster/_core/execution/plan/execute_plan.py
@@ -50,15 +50,8 @@ def inner_plan_execution_iterator(
                     yield DagsterEvent.capture_logs(
                         pipeline_context, step_keys, log_key, log_context
                     )
-                except Exception as e:
-                    yield DagsterEvent.engine_event(
-                        plan_context=pipeline_context,
-                        message="Exception while setting up compute log capture",
-                        event_specific_data=EngineEventData(
-                            error=serializable_error_info_from_exc_info(sys.exc_info())
-                        ),
-                    )
-                    log_capture_error = e
+                except Exception:
+                    yield from _handle_compute_log_setup_error(pipeline_context, sys.exc_info())
 
             # It would be good to implement a reference tracking algorithm here to
             # garbage collect results that are no longer needed by any steps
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/execution/plan/execute_plan.py
```diff
@@ -35,106 +35,132 @@ def inner_plan_execution_iterator(
...
+            try:
+                capture_stack.close()
+            except Exception:
+                yield DagsterEvent.engine_event(
+                    plan_context=pipeline_context,
+                    message="Exception while cleaning up compute log capture",
+                    event_specific_data=EngineEventData(
+                        error=serializable_error_info_from_exc_info(sys.exc_info())
+                    ),
+                )
```
alangenfeld, line 163: i dont know what it looks like - but there has to be a better way to structure this code this is really gnarly

python_modules/dagster/dagster_tests/core_tests/storage_tests/test_compute_log_manager.py
```diff
@@ -23,6 +34,80 @@ def test_compute_log_manager_instance():
...
+        self._fail_on_setup = check.opt_bool_param(fail_on_setup, "fail_on_setup")
+        self._fail_on_teardown = check.opt_bool_param(fail_on_teardown, "fail_on_teardown")
+
+    @contextmanager
+    def capture_logs(self, log_key: Sequence[str]) -> Generator[CapturedLogContext, None, None]:
+        if self._fail_on_setup:
+            raise Exception("wahhh")
+        yield CapturedLogContext(log_key=log_key)
+        if self._fail_on_teardown:
+            raise Exception("blahhh")
```
alangenfeld, line 48: nit: wahhh / blahhh -> "fail on setup" / "fail on teardown"

python_modules/dagster/dagster/_core/execution/plan/execute_plan.py
```diff
@@ -35,106 +35,132 @@ def inner_plan_execution_iterator(
...
 
-            with ExitStack() as step_stack:
-                if not isinstance(compute_log_manager, CapturedLogManager):
-                    # capture all of the logs for individual steps
-                    log_capture_error = None
-                    try:
-                        step_stack.enter_context(
-                            pipeline_context.instance.compute_log_manager.watch(
-                                step_context.pipeline_run, step_context.step.key
+                with ExitStack() as step_stack:
```
alangenfeld, line 89: do we need another whole stack here versus just popping one off of a single exit stack?

alangenfeld (approved):
take one more pass at cleaning up execute_plan.py insanity before landing but these tests give me confidence this is good

python_modules/dagster/dagster/_core/execution/plan/execute_plan.py
prha: yeah, this is because we can't pop a single CM off the stack... it's either close or pop_all.

PR #10777, opened 2022-11-29
feat(dbt-cloud): allow group name to be configurable

rexledesma:
### Summary & Motivation
Resolves https://github.com/dagster-io/dagster/issues/10662.

### How I Tested These Changes
pytest

PR #10776, opened 2022-11-29
test(dbt-cloud): move resources into fixtures

rexledesma:
### Summary & Motivation
As the title.

### How I Tested These Changes
existing pytest

PR #10778, opened 2022-11-29
feat(dbt-cloud): allow asset key to be configurable

rexledesma:
### Summary & Motivation
As the title.

pytest

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/asset_defs.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/asset_defs.py
@@ -309,6 +309,7 @@ class DbtCloudCacheableAssetsDefinition(CacheableAssetsDefinition):
 def load_assets_from_dbt_cloud_job(
     dbt_cloud: ResourceDefinition,
     job_id: int,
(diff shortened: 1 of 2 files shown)


rexledesma:
* master
  * **PR #10776** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10776" target="_blank">Graphite</a> 
    * **PR #10777** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10777" target="_blank">Graphite</a> 
      * **PR #10778** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10778" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10778?utm_source=stack-comment).

OwenKephart (commented):
this was my original sin, but I'd prefer this parameter to be called node_info_to_asset_key_fn. it's really a tradeoff between how much we care about the API being similar between the dbt Cloud / dbt Core integrations, vs how much we care about the API being similar within each integration (i.e. all the other function parameters end with `_fn`).

arguably, if and when we move to a node_info_to_asset_properties_fn/metadata_to_asset_properties_fn sort of vibe, then at that point we can deprecate all of these parameters at once, so we can just stick with this naming for now.

Issue #10705, opened 2022-11-23
Re-execute backfilled run starts other repo jobs


narfunikita:
### Dagster version

1.0.10

### What's the issue?

Re-execute backfilled run.
Instead of one job to be reruned,  Dagster starts all repo jobs without partition_def and all jobs with the same start_date in partition_def as the rerunned asset has

### What did you expect to happen?

I expect that re-execution of the backiffled asset starts only the same asset.

### First example: Two partitioned assets and unpartitioned assets

```python
@asset(
    key_prefix=["foo_prefix"],
)
def unpartitioned_foo_asset(context: OpExecutionContext):
    raise Exception("Foo asset 1 unpartitioned")

unpartitioned_foo_asset_job = define_asset_job(
    name=f"_job_for_{'_'.join(unpartitioned_foo_asset.key.path)}",
    selection=AssetSelection.assets(unpartitioned_foo_asset),
)

test_partition_def = DailyPartitionsDefinition(start_date="2022-06-26", timezone="Europe/Moscow")
@asset(
    key_prefix=["bar_prefix"],
    partitions_def=test_partition_def,
)
def partitioned_asset(context: OpExecutionContext):
    raise Exception(f"partitioned asset. {context.partition_key}")

partitioned_asset_job = define_asset_job(
    name=f"_job_for_{'_'.join(partitioned_asset.key.path)}",
    selection=AssetSelection.assets(partitioned_asset),
    partitions_def=test_partition_def,
)

barabar_partition_def = DailyPartitionsDefinition(start_date="2022-06-26", timezone="Europe/Moscow")
@asset(
    key_prefix=["barabar_prefix"],
    partitions_def=barabar_partition_def,
)
def partitioned_barabar_asset(context: OpExecutionContext):
    raise Exception(f"partitioned Second barabar asset. {context.partition_key}")

partitioned_barabar_asset_job = define_asset_job(
    name=f"_job_for_{'_'.join(partitioned_barabar_asset.key.path)}",
    selection=AssetSelection.assets(partitioned_barabar_asset),
    partitions_def=barabar_partition_def,
)
```

Repo 
```python
@repository
def my_repo() -> List[
    Union[JobDefinition, ScheduleDefinition, UnresolvedAssetJobDefinition, AssetsDefinition, SensorDefinition]
]:
    return [
        partitioned_asset,
        partitioned_asset_job, 
        unpartitioned_foo_asset,
        unpartitioned_foo_asset_job,
[...]

Issue #10707, opened 2022-11-23
Run backfill crashed if any another asset in repo has config_schema


narfunikita:
### Dagster version

1.0.10

### What's the issue?

Run backfill crashed if any another unpartitioned asset in repo has config_schema.
looks like it trying to check that parameters for this backfilled asset is appopriate for all unpartitioned assets in the repo

### What did you expect to happen?
Checks are called only for that asset for which the backfill run was launched.

### How to reproduce?

Unpartition Asset with custom config_schema
```python
@asset(
    key_prefix=["foo_prefix"],
    config_schema={"myparam": Field(str, description="YYYY-MM-DD")},
)
def unpartitioned_foo_asset(context: OpExecutionContext):
    raise Exception("Foo asset 1 unpartitioned")

unpartitioned_foo_asset_job = define_asset_job(
    name=f"_job_for_{'_'.join(unpartitioned_foo_asset.key.path)}",
    selection=AssetSelection.assets(unpartitioned_foo_asset),
)
```

Partitioned asset
```python
test_partition_def = DailyPartitionsDefinition(start_date="2022-06-26", timezone="Europe/Moscow")
@asset(
    key_prefix=["bar_prefix"],
    partitions_def=test_partition_def,
)
def partitioned_asset(context: OpExecutionContext):
    raise Exception(f"partitioned asset. {context.partition_key}")

partitioned_asset_job = define_asset_job(
    name=f"_job_for_{'_'.join(partitioned_asset.key.path)}",
    selection=AssetSelection.assets(partitioned_asset),
    partitions_def=test_partition_def,
)
```
repo 
```python
@repository
def my_repo() -> List[
    Union[JobDefinition, ScheduleDefinition, UnresolvedAssetJobDefinition, AssetsDefinition, SensorDefinition]
]:
    return [
        partitioned_asset,
        partitioned_asset_job, 
        unpartitioned_foo_asset,
        unpartitioned_foo_asset_job
    ]
```

if we try to run backfill for partitioned asset
![partition_asset_run_backfill](https://user-images.githubusercontent.com/9852326/203549885-a1088970-f737-4bc8-8329-c58872355a44.png)

We will get an error that config not setted for ANOTHER UNpartitioned asset
```
dagster._core.errors.DagsterInvalidConfigError: Error in config for job
[...]

PR #10782, opened 2022-11-29
fix partition backfill asset_selection

OwenKephart:
### Summary & Motivation

Fixes:

https://github.com/dagster-io/dagster/issues/10705
https://github.com/dagster-io/dagster/issues/10707

These two issues have the same root cause. Before this change, we were creating/serializing `DagsterRun` objects which had a completely accurate `ExternalExecutionPlan` object associated with them (via an execution plan snapshot), but the external execution plan is not directly used in the process that actually launches the run. Instead, it just uses the pipeline code origin that we serialize, along with any asset_selection / op_selection type information found on the `DagsterRun`, in order to get the actual JobDefinition to execute.

We implicitly create a set of base jobs on the repository, where each of these jobs contains all assets that share the same PartitionsDefinition + any unpartitioned assets, resulting in the behavior experienced in the reports. For the config issue, we were not properly subsetting the base job (to remove the unpartitioned asset) when validating the config, even though the execution plan (which is used to render that step diagram) was accurate. For the re-execution case, we rebuild the external execution plan from scratch, resulting in all asset selection information being dropped.

This change makes sure that the asset_selection is present not only on the execution plan, but also the `DagsterRun` itself.

Modified the existing test to be more realistic, failed before the change, succeeds now.

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/backfill.py
+++ b/python_modules/dagster/dagster/_core/execution/backfill.py
@@ -313,6 +313,9 @@ def create_backfill_run(
         external_pipeline_origin=external_pipeline.get_external_origin(),
         pipeline_code_origin=external_pipeline.get_python_origin(),
         solid_selection=solid_selection,
+        asset_selection=frozenset(backfill_job.asset_selection)
+        if backfill_job.asset_selection
+        else None,
     )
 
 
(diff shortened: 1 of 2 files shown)


clairelin135 (approved):
Thanks for taking this on!

PR #10783, opened 2022-11-29
[code-location-selector-stack Add CodeLocationSelector; Have run_status_sensor accept it

schrockn:
Summary & Motivation: Adds the ability to target a code location in the run_status_sensor rather than
a repository.

This is meant to be used in concert with the `Definitions` API that no longer exposes
the concept of a repository to a user. The new language is that the `Definitions` object determines the set
of definitions within a code location.

This diff hoists the computation of other_repos, other_repo_jobs, and current_repo_jobs out of the closure
so we get code coverage from the local unit test

Test Plan:
BK. Added additional testing in https://github.com/dagster-io/dagster/pull/10970


jamiedemaria:
This looks like a fine approach to me given the current pattern

I wonder if an overall better way to handle the different selectors `RepoSelector`, `CodeLocationSelector`, and `JobSelector` if we could remove them all and introduce a more general object that allows for selection at different levels. Can't think of a name for it rn, but let's say it's called the FooSelector. 

It'd be something like `FooSelector(deployment, code_location, job)` where all of the params are an `Optional[str] = None` and you could do something like `FooSelector(code_location="my_code_location)` to select all of the jobs in that code location, or you could do `FooSelector(code_location="my_code_location", job="my_job")` to select a single job in another code location. And then we could maybe support things like `FooSelector(code_location="*", job="my_job")` to monitor every job with that name regardless of code location.  The sensor would still take a list of these along with the plain job name strings. 

It's still pretty half baked in my head. i'm not sure if that set of parameters (esp including deployment) fully makes sense, or what the behavior specifics would be for default None values vs `*` inputs. But I think it could be nice to have a single object that can support this type of selection

schrockn:
@jamiedemaria completely agree and that's why I was alluding to with my comment "since I think we may be due for an even more aggressive revamp if we support cross-deployment/instance sensors."

My general thinking is that we should consider and design a successor more deliberately, and also do some deeper thinking around x-deployment references.

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -235,14 +239,14 @@ def run_failure_sensor(
...
-        monitored_jobs (Optional[List[Union[JobDefinition, GraphDefinition, UnresolvedAssetJobDefinition, RepositorySelector, JobSelector]]]):
+        monitored_jobs (Optional[List[Union[JobDefinition, GraphDefinition, UnresolvedAssetJobDefinition, RepositorySelector, JobSelector, CodeLocationSelector]]]):
             The jobs in the current repository that will be monitored by this failure sensor.
             Defaults to None, which means the alert will be sent when any job in the current
             repository fails.
         monitor_all_repositories (bool): If set to True, the sensor will monitor all runs in the
             Dagster instance. If set to True, an error will be raised if you also specify
             monitored_jobs or job_selection. Defaults to False.
-        job_selection (Optional[List[Union[JobDefinition, GraphDefinition, RepositorySelector, JobSelector]]]):
+        job_selection (Optional[List[Union[JobDefinition, GraphDefinition, RepositorySelector, JobSelector, CodeLocationSelector]]]):
```
jamiedemaria, line 249: need to add CodeLocationSelector to the parameter type on line 211 too, right?

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -360,15 +369,39 @@ def __init__(
...
+            for job in (monitored_jobs or [])
+        ]
+
         self._run_status_sensor_fn = check.callable_param(
             run_status_sensor_fn, "run_status_sensor_fn"
         )
         event_type = PIPELINE_RUN_STATUS_TO_EVENT_TYPE[run_status]
 
+        # split monitored_jobs into external repos, external jobs, and jobs in the current repo
+        other_repos = (
```
jamiedemaria, line 389: nice

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -628,6 +645,7 @@ def run_status_sensor(
                 UnresolvedAssetJobDefinition,
                 "RepositorySelector",
                 "JobSelector",
+                "CodeLocationSelector",
```
jamiedemaria, line 648: add to doc string

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -639,6 +657,7 @@ def run_status_sensor(
...
```
jamiedemaria, line 660: add to doc string

schrockn:
Tests added in https://github.com/dagster-io/dagster/pull/10970

schrockn:
Jamie I'm testing out a graphite replacement spr but I think it doesn't do the intermediate commit history in the way you expect it. If you check the files tab of the overall PR I believe you will see your feedback addresses

schrockn:
Nevermind this was my error, updating.

jamiedemaria (approved):

PR #10788, opened 2022-11-29
[code-location-concept-page-1] Concept page for code locations [WIP]

schrockn:
Summary & Motivation: This is a WIP, but I wanted to throw up the intro verbiage
to kickstart the conversation here.

The decision I want to discuss here is separating the notion of a python module the has a `Definitions` object defined at module-scope in the `defs` from an entity that is loadable by our tools. This is a subtle but important point. It means that a location doesn't really exist until it it loadable by our tools. So when you type `dagit -f foo.py` it is meant to mean "I am loading this code location in dagit with the current Python environment, communicating with that Python Environment locally, and loading the Python file foo.py as the entry point". This is different than saying that foo.py is a code location, even though undoubtedly it will be referred to it as that colloquially.

Test Plan: Read

### DIFF
--- a/docs/content/concepts/code-locations.mdx
+++ b/docs/content/concepts/code-locations.mdx
@@ -70,16 +70,32 @@ defs = Definitions(
 <TabGroup>
     <TabItem name="Local Development">
 
+### Load definitions in a file
+
+Example: `$ dagit -f foo.py`
+
 `dagit` and other Dagster tools can load a file directly as a code location. The command `dagit -f foo.py` is loading a code location: it will load the definitions in `foo.py` as a code location in the same python environment where Dagit resides.
 
-Installed python packages can also be loaded directly. For Dagster projects deployed to production we recommend this style of development, as it eliminates an entire class of python import errors. Create a Dagster project with the `dagster project scaffold` command or by using one of our templates. This project can then be installed using `pip install -e .` to make it an editable install for local development.
+### Load definitions in an installed package
+
+Example: `$ dagit -m your_package_name`
+
+Installed python packages can also be loaded by Dagster tools. It will load the definitions defined as the top-level of that package, in its root `__init__.py` file. For Dagster projects deployed to production we recommend this style of development, as it eliminates an entire class of python import errors. Create a Dagster project with the `dagster project scaffold` command or by using one of our templates. This project can then be installed using `pip install -e .` to make it an editable install for local development.
 
 Once installed, this package can be loaded typing `dagit -m your_package_name`. This should be read as "load definitions in the `defs` symbol in this package (defined as its root `__init__.py` file) in the same virtual environment as Dagit itself.
 
+### Load definitions without any command line arguments
+
+Example: `$ dagit`
+
 Our project templates also create a `pyproject.toml` file with a `tool.dagster` section an assignment to the `python_package` variable in that section. This allows users to type just `dagit` without any parameters and the package name is defined in that variable. This is very convenient for local development.
 
+```
     [tool.dagster]
     python_package = "your_package_name"
+```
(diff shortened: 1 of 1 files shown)


schrockn:
Here is the direct link to the preview. https://dagster-git-schrockn-concept-page-for-code-loca-5931d6-elementl.vercel.app/concepts/code-locations

@yuhan in particular interested in your thoughts on this notion of separating the definition of a "python module entrypoint" from the code location.

docs/content/concepts/code-locations.mdx
```diff
@@ -0,0 +1,147 @@
...
+# Code Locations
+
+A code location is set of dagster definitions loadable and accessible by Dagster's tools, such as the CLI, Dagit, and Dagster Cloud. A code location comprises:
+
+- A reference to a python module or package that has an instance of `Definitions` at the "defs" variable. We will refer to that as the "entry point".
+- A Python environment that can successfully load the entry point.
+
+Definitions within a code location have a common namespace and must have unique names. This allows them to be grouped and organized by code location in tools.
+
+Before the introduction of the <PyObject object="Definitions" /> API, definitions were grouped into repositories, and there could be many repostories in a particular code location. See the [guide](/concepts/repositories-workspaces/repositories) on Repositories for this previous API and mental model.
```
yuhan, line 15: We can wrap this in a note so it won't get in the way for users who don't care what a repository was

```suggestion
<Note>
Before the introduction of the <PyObject object="Definitions" /> API, definitions were grouped into repositories, and there could be many repostories in a particular code location. See the <a href="/concepts/repositories-workspaces/repositories">guide</a> on Repositories for this previous API and mental model.
</Note>
```

yuhan (commented):
I like the summary and like where the page is going. Left a few comments (mostly about formatting and structure).

Besides, should we talk about sequencing here? Shall me or @erinkcochran87 take over this while you're OOO?

docs/content/concepts/code-locations.mdx
schrockn: cool. i'll make this change myself as committing suggestions doesn't play nice with spr

gibsondan:
Separating out the concept of the "code pointer" and the environment where the code is running makes a lot of sense to me. Especially if we can use existing terms for these things rather than introducing new concepts that the user needs to grok, which I think that doc does successfully.

I'm not totally sure the "Python environment" is the right term though? Or at least it doesn't encompass everything that goes into a code location (like the image / container context / etc.) - there's almost yet a third level that defines the container / compute environment in which the Python environment exists. There's also metadata like a git hash / pointer to a git repo.

Like if you look on the Code locations tab in Dogfood there's a bunch of stuff there that's not yet explained on the Concepts page:

image (12)



Running your own grpc server also doesn't fit neatly into this split between code pointer and environment (since you just specify a host and port there, no code pointer required) but I think that's fine.

schrockn:
Yup this is great feedback and I think will be worth a call to talk through. I'm leaning towards pushing out this a week given unfortunate overlapping PTO with key people (you and Sandy) and Community Day this week.

Just to list out the properties of a code location that you list.

* A particular deployment (I'm hypothesizing a consolidation of instance and deployment language here btw) defines how a code location is loaded. 
* A loaded code location has a temporal component (e.g. the commit hash) as well as it's compute environment, as defined by the semantics of the deployment.
* So the language would be "a dagster cloud yaml defines how code locations are packaged into docker container, deployed, and loaded by the dagster cloud infrastructure" 

I think the cost of this language consolidation would be overlap between how people refer to the basket definitions as well as a deployed instance of a code location. The alternative would be introducing a term like "code location server" or something but I would like to avoid that.

PR #10803, opened 2022-11-30
Internal abstraction renames (solid -> node)

smackesey:
### Summary & Motivation

4 significant renames:

- SolidOutputHandle -> NodeOutput
- SolidInputHandle -> NodeInput
- InvokedSolidOutputHandle -> InvokedNodeOutputHandle
- InvakedSolidDynamicOutputWrapper -> InvokedNodeDynamicOutputWrapper

These abstractions don't get serialized so I don't think there are any complicating factors for renames.

Note that `Solid{Output,Input}Handle` were renamed to `Node{Input,Output}` _without_ "Handle' for a few reasons:

- `NodeInputHandle` and `NodeOutputHandle` were already taken!
- "Handle" was not a very accurate name. Elsewhere in the codebase (and intuitively IMO), a "Handle" is an _identifier_ for an object-- it shouldn't hold big data structures, but rather just keys allowing the abstraction holding the larger data structure to be looked up. But `Solid{Input,Output}Handle` actually contained the entire `{Input,Output}Definition`-- so it actually makes more sense to call them `Node{Input,Output}` (and these names weren't taken).

I also made an effort to rename local variables and properties wherever these classes were being used, e.g. a local variable called `output_handle` that contained a `SolidOutputHandle` is now called `node_output`.

Also contains a few minor typing additions and a local variable rename (for consistency) in asset_layer.py.

Existing BK.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/__init__.py
+++ b/python_modules/dagster/dagster/_core/definitions/__init__.py
@@ -5,10 +5,10 @@ from .dependency import (
     MultiDependencyDefinition,
     Node,
     NodeHandle,
+    NodeInput,
     NodeInvocation,
-    SolidInputHandle,
+    NodeOutput,
     SolidInvocation,
-    SolidOutputHandle,
 )
 from .events import (
     AssetKey,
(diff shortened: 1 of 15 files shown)


alangenfeld (approved):

alangenfeld:
nice thanks for the clean-up and thorough pr summary

PR #10806, opened 2022-11-30
Snowflake I/O manager guide

jamiedemaria:
### Summary & Motivation
This PR has been broken into a couple smaller PRs:
Quick Start - https://github.com/dagster-io/dagster/pull/10812

Adds a guide for the Snowflake I/O manager

Lots still to do in this, but here's my first pass at the content to start getting feedback! I erred on the side of longer code snippets, but if they are too long, i can condense. one of the pieces of feedback i hear sometimes is that a code snippet like

```
@asset
def my_asset():
    return [1, 2, 3]
```
isn't very helpful, so i tried to make the assets a bit more complex, which in turn makes them longer.

I also haven't even begun to address sections covering using the snowflake io manager with other tools like dbt. but I'm thinking that might fit better in its own guide. 

vercel preview - https://dagster-git-jamie-snowflake-guide-elementl.vercel.app/integrations/snowflake

Other TODOs:
- [ ] put code snippets under test
- [ ] formalize the structure into something that can be re-used for other io manager guides


rexledesma (changes requested):
I think this is a little text-heavy - and still a little too "choose your own adventure". We should wrangle the complexity by explicitly calling out the steps that people will always have to do (e.g. set up the io manager, how to configure it on an asset, how to load/save inputs/outputs from upstream assets).

I think we can split up this PR if that helps - part of the PR should lay out Steps 1-3 for setting up an IO manager + configuring it on a repository + asset.

Then in a separate PR, we can go over the customization route (step 4) which includes saving a subset of columns, storing data in multiple schemas, etc

Issue #10789, opened 2022-11-29
Add information in the docs regarding the partition capabilities of build_asset_reconciliation_sensor


OwenKephart:
### What's the issue or suggestion?

The docstring of that function has been updated to reference partitions, so those changes should be ported over, along with maybe a paragraph or so of extra explanation.

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

PR #10807, opened 2022-11-30
Add docs for partitioned asset reconciliation sensor

clairelin135:
https://github.com/dagster-io/dagster/issues/10789

Adds an explanation of how the partitioned asset reconciliation works to the documentation


erinkcochran87 (approved):
I think this looks good overall! I left a few small comments on the Vercel preview - nothing major, so I'm approving this. 

One thing that might be nice in a future PR is to include a diagram or something visual alongside the example you added. I understood the explanation just fine, but I think a visual explaining this would go along way in clarifying things.

PR #10780, opened 2022-11-29
Remove typing_compat dependency

smackesey:
### Summary & Motivation

This removes the `typing_compat` dependency. The same functionality is provided by the `typing_extensions` package, which we already use for other things.

### How I Tested These Changes

Existing tests.

PR #10818, opened 2022-11-30
add pep593 support for inputs

adriangb:
### Summary & Motivation

I've been mainly browsing the docs so far but I think there's an opportunity to reduce boilerplate by using PEP593 annotations to allow dagster metadata to flow alongside Python typing information.

I made a pretty crude POC. I'm honestly not sure what metadata should go there `a DagsterType` directly like I did here? Or should it be an `In` object? Maybe both? What about outputs?

This is a super crude POC. I'm not familiar with the codebase so I just made a `test.py` file and run it like `python test.py`:

```python
from typing import Annotated

from dagster import job, op, DagsterType

set_containing_1 = DagsterType(
    name="set_containing_1",
    description="A set containing the value 1. May contain any other values.",
    type_check_fn=lambda _context, obj: isinstance(obj, set) and 1 in obj,
)

@op
def get_data() -> set[int]:
    return {2}

SetContaining1 = Annotated[set[int], set_containing_1]

@op
def some_operation(foo: SetContaining1):
    pass

@op
def some_other_op(bar: SetContaining1):
    pass

@job
def file_sizes_job():
    data = get_data()
    some_operation(data)
    some_other_op(data)

if __name__ == "__main__":
    file_sizes_job.execute_in_process()
```

### DIFF
--- a/python_modules/dagster/dagster/_core/types/dagster_type.py
+++ b/python_modules/dagster/dagster/_core/types/dagster_type.py
@@ -9,6 +9,7 @@ from typing import Optional as TypingOptional
 from typing import Sequence
 from typing import Type as TypingType
 from typing import cast
+from dagster._utils.typing_api import unpack_if_type_is_pep593_type_annotation
 
 from typing_compat import get_args, get_origin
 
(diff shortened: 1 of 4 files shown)


adriangb:
Any interest in moving forward with this? With some initial feedback I'm happy to resolve conflicts, fix any failing tests, etc.

I'd also like to point out that regardless of this change I think you should ditch the `typing_compat` dependency to reduce your dependency graph since you already depend on `typing-extensions` which provides the same functionality and is maintained by Python core devs. `typing-compat` has 1 star on GitHub, it seems like a huge security risk for all Dagster users.

adriangb:
Looks like `typing-compat` was already removed in #10780, glad to see that!

smackesey:
Hi @adriangb, thanks for this proposal. I will give it a closer look this week, sorry for the late review here.

PR #10821, opened 2022-12-01
[RFC] Update cached_status_data column 1/2

clairelin135:
Following the [cached asset partition status](https://www.notion.so/dagster/Cached-incremental-asset-partition-summaries-9afd9aa0b91940d49c3d5af5a832131a) proposal, this PR creates an `update_asset_cached_status_data` method that writes to the new cached_status_data column of the AssetKeyTable. This column now contains:

- latest evaluated storage ID of the asset
- a serialized unique identifier to represent the partitions definition. If the partitions definition is changed, the mismatch between the serialized identifier and the new identifier triggers a recalculation and update to the cached status column.
- serialized representation of the materialized partition subset. This contains partition ranges for time window partitions definitions, and a set of materialized partition keys otherwise.

The `update_asset_cached_status_data` will check if the column exists, and if it does, it will:
- read the currently cached data per asset and
- if the partition definition unique identifier differs, recalculates the partition materialization status from event log storage
- otherwise, the unique identifier remains the same and the method reads the unevaluated events
- updates the `cached_status_data` column with the new partition materialization data

PR for part 2 updates Dagit to read from the cached data: https://github.com/dagster-io/dagster/pull/10822

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
@@ -1,4 +1,5 @@
 import itertools
+import json
 from datetime import datetime
 from typing import Dict, List, Mapping, NamedTuple, Optional, Sequence, Tuple
 
@@ -10,7 +11,12 @@ from dagster._core.storage.tags import (
     get_multidimensional_partition_tag,
 )
 
-from .partition import Partition, PartitionsDefinition, StaticPartitionsDefinition
+from .partition import (
+    DefaultPartitionsSubset,
+    Partition,
+    PartitionsDefinition,
+    StaticPartitionsDefinition,
+)
 
 INVALID_STATIC_PARTITIONS_KEY_CHARACTERS = set(["|", ",", "[", "]"])
 
@@ -160,6 +166,10 @@ class MultiPartitionsDefinition(PartitionsDefinition):
         )
 
     @property
+    def partition_dimension_names(self) -> List[str]:
+        return [dim_def.name for dim_def in self._partitions_defs]
+
+    @property
     def partitions_defs(self) -> Sequence[PartitionDimensionDefinition]:
         return self._partitions_defs
 
@@ -220,6 +230,57 @@ class MultiPartitionsDefinition(PartitionsDefinition):
 
     def __repr__(self) -> str:
         return f"{type(self).__name__}(dimensions={[str(dim) for dim in self.partitions_defs]}"
+
+    def get_multi_partition_key_from_str(self, partition_key_str: str) -> MultiPartitionKey:
+        """
+        Given a string representation of a partition key, returns a MultiPartitionKey object.
+        """
+        check.str_param(partition_key_str, "partition_key_str")
+
+        partition_key_strs = partition_key_str.split("|")
+        check.invariant(
+            len(partition_key_strs) == len(self.partitions_defs),
+            f"Expected {len(self.partitions_defs)} partition keys in partition key string {partition_key_str}, "
+            f"but got {len(partition_key_strs)}",
+        )
+        keys_per_dimension = [
+            (dim.name, dim.partitions_def.get_partition_keys()) for dim in self._partitions_defs
+        ]
+
+        partition_key_dims_by_idx = dict(enumerate([dim.name for dim in self._partitions_defs]))
+        for idx, key in enumerate(partition_key_strs):
+            check.invariant(
+                key in keys_per_dimension[idx][1],
+                f"Partition key {key} not found in dimension {partition_key_dims_by_idx[idx][0]}",
+            )
+
+        multi_partition_key = MultiPartitionKey(
+            {partition_key_dims_by_idx[idx]: key for idx, key in enumerate(partition_key_strs)}
+        )
+        return multi_partition_key
+
+    def deserialize_subset(self, serialized: str) -> "MultiPartitionsSubset":
+        return MultiPartitionsSubset.from_serialized(self, serialized)
+
+
+class MultiPartitionsSubset(DefaultPartitionsSubset):
+    @staticmethod
+    def from_serialized(
+        partitions_def: PartitionsDefinition, serialized: str
+    ) -> "MultiPartitionsSubset":
+        if not isinstance(partitions_def, MultiPartitionsDefinition):
+            check.failed(
+                "Must pass a MultiPartitionsDefinition object to deserialize MultiPartitionsSubset."
+            )
+        return MultiPartitionsSubset(
+            subset=set(
+                [
+                    partitions_def.get_multi_partition_key_from_str(key)
+                    for key in json.loads(serialized)
+                ]
+            ),
+            partitions_def=partitions_def,
+        )
 
 
 def get_tags_from_multi_partition_key(multi_partition_key: MultiPartitionKey) -> Mapping[str, str]:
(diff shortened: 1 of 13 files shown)


python_modules/dagster/dagster/_core/storage/partition_status_cache.py
```diff
@@ -0,0 +1,273 @@
...
+    def deserialize_materialized_partition_subsets(
+        self, partitions_def: PartitionsDefinition
+    ) -> PartitionsSubset:
+        if not self.materialized_partition_subsets:
+            return partitions_def.empty_subset()
+
+        return partitions_def.deserialize_subset(self.materialized_partition_subsets)
+
+
+def get_materialized_multipartitions_from_tags(
```
sryza, line 83: Nitpick: I think could leave `from_tags` out, because that's an implementation detail. This name led me to think that the tags would be one of the arguments.

python_modules/dagster/dagster/_core/storage/partition_status_cache.py
```diff
@@ -0,0 +1,273 @@
...
+    materialized_keys: List[MultiPartitionKey] = []
+    for event_tags in instance.get_event_tags_for_asset(asset_key):
+        event_partition_keys_by_dimension = {
+            get_dimension_from_partition_tag(key): value
+            for key, value in event_tags.items()
+            if key.startswith(MULTIDIMENSIONAL_PARTITION_PREFIX)
+        }
+
+        if all(
+            [
```
sryza, line 96: Can take out these square brackets - `all` can operate directly on iterables without wrapping them in lists.

python_modules/dagster/dagster/_core/storage/partition_status_cache.py
```diff
@@ -0,0 +1,273 @@
...
+                    {
+                        dimension_names[0]: event_partition_keys_by_dimension[dimension_names[0]],
+                        dimension_names[1]: event_partition_keys_by_dimension[dimension_names[1]],
+                    }
+                )
+            )
+    return materialized_keys
+
+
+def _rebuild_status_cache(
```
sryza, line 112: Nitpick: this won't always be a _re_-build, right? I.e. we'll call this the first time? If so, maybe a name like "build from scratch"?

python_modules/dagster/dagster/_core/storage/partition_status_cache.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+@whitelist_for_serdes
+class AssetStatusCacheValue(
+    NamedTuple(
+        "_AssetPartitionsStatusCacheValue",
+        [
+            ("latest_storage_id", int),
+            ("partitions_def_id", Optional[str]),
+            ("materialized_partition_subsets", Optional[str]),
```
sryza, line 27: My initial reaction was to think it's a little weird to represent this property as serialized instead of deserializing it and holding the `PartitionsSubset`. I suppose the advantage of having it serialized though is that we can avoid the deserialization cost in cases where only we need other entries from the cache?  If we do want to have it serialized, we should include that in the name to make it clear.

Also, subsets should be singular, not plural, right? I.e. it's just one subset, which includes multiple partitions?

python_modules/dagster/dagster/_core/storage/partition_status_cache.py
```diff
@@ -0,0 +1,273 @@
...
+    """
+    Set of asset fields that reflect partition materialization status. This is used to display
+    global partition status in the asset view.
+
+    Properties:
+        latest_storage_id (int): The latest evaluated storage id for the asset.
+        partitions_def_id (Optional(str)): The serializable unique identifier for the partitions
+            definition. When this value differs from the new partitions definition, this cache
+            value needs to be recalculated. None if the asset is unpartitioned.
+        materialized_partition_subsets (Optional(str)): The serializable representation of the
```
sryza, line 41:
```suggestion
        materialized_partition_subsets (Optional(str)): The serialized representation of the
```

python_modules/dagster/dagster/_core/storage/partition_status_cache.py
clairelin135: The problem is that we need the partitions definition that was used to serialize the partitions subset in order to know how to deserialize it. For example, given a serialized partitions subset, we don't know if it is a static partitions definition or a multi-partitions definition.

I figured that since we only need the cached value if the partitions definition remains the same, we can just deserialize it in those cases. I'll rename this to `serialized_materialized_partition_subset`.

python_modules/dagster/dagster/_core/storage/partition_status_cache.py
```diff
@@ -0,0 +1,273 @@
...
+    asset_key: AssetKey,
+    latest_storage_id: int,
+    partitions_def: Optional[PartitionsDefinition] = None,
+):
+    """
+    This method refreshes the asset status cache for a given asset key. It recalculates
+    the materialized partition subset for the asset key and updates the cache value.
+    """
+    if not partitions_def:
+        return AssetStatusCacheValue(latest_storage_id=latest_storage_id)
```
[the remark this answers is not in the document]
clairelin135: I think for un-partitioned assets we still want to store an asset status cache value. This will improve performance for things like fetching freshness data via graphql in batches since we currently don't have access to the latest storage id via the asset key table

clairelin135:
@sryza I've updated this PR per your feedback. I've also added a set of changes so this feature should be fully functional now:
- Wiping an asset will now also wipe its cached data
- Disallowing dynamic partitions since we don't display them in Dagit
- Improved test coverage covering each possible partition type

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
```diff
@@ -221,6 +231,53 @@ def __str__(self) -> str:
...
+        check.str_param(partition_key_str, "partition_key_str")
+
+        partition_key_strs = partition_key_str.split("|")
+        check.invariant(
+            len(partition_key_strs) == len(self.partitions_defs),
+            f"Expected {len(self.partitions_defs)} partition keys in partition key string {partition_key_str}, "
+            f"but got {len(partition_key_strs)}",
+        )
+
+        all_partition_keys = self.get_partition_keys()
```
sryza, line 247: I think it will be pretty expensive to get all the partition keys every time this method is called.

sryza (approved):
This looks great.

A couple small things before merging:
- Maybe it's already there and I didn't read closely enough, but can we make sure there's a test that asserts that the cache gets discarded if the PartitionsDefinition changes?
- I left a comment about perf

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1126,14 +1152,8 @@ def _fetch_raw_asset_rows(self, asset_keys=None, prefix=None, limit=None, cursor
...
         ]
-        if self.has_asset_key_index_cols():
-            columns.extend(
-                [
-                    AssetKeyTable.c.wipe_timestamp,
-                    AssetKeyTable.c.last_materialization_timestamp,
-                    AssetKeyTable.c.tags,
-                ]
-            )
+        if self.has_asset_key_col("cached_status_data"):
```
prha, line 1155: nit: should this use `can_cache_asset_status_data`?

prha (approved):
Nice job on this

PR #10822, opened 2022-12-01
[3/3 partition status cache] Update graphQL partition data

clairelin135:
This PR enables Dagit to display cached partition data, making the following changes:
- For time window partitions definitions:
  - Returns start and end partition key and start and end time in metadata object on the `partitionDefinition` resolver
  - Materialized time windows are now returned back in ranges, containing:
    - start partition key and end partition key
    - time window start and end
  - Adds some methods to time window partitions definitions (`get_num_partitions`, `get_partition_keys_between_idxs`) to improve performance, minimizing the number of string format operations necessary
  - `partitionKeysByDimension` resolver accepts optional start and end index params to fetch a subset of partition keys
- For multidimensional partitions definitions:
  - Now includes an `isPrimaryDimension` on the partitions definition dimension resolver
  - Encodes materialized partitions in a run length encoding, selecting the primary dimension to be the time dimension
- For all partitioned assets adds a `partitionStats` resolver that returns back the total partitions and # materialized partitions
 
Potential TODO: build in some sort of flag to allow Dagit to work without caching the data (in case if there are performance slowdowns)


sryza (commented):
The main question I have is whether we should be more "forward-looking" about what we want to support here. Thoughts on having the APIs cover ranges of partitions instead of individual partitions? And on returning a count of materialized partitions?

If we think it will be easy to evolve into that later, then it's fine to ship this first. But wondering whether this will be short-lived and it's worth jumping directly to the ideal state.

bengotow:
Hey @clairelin135  / @sryza! Hmm this is a good question. Right now the UI does build "spans" for display in the partition health bar - it's only for time dimensions right now, but it could be an efficient way of compacting all of this data for transfer.

There are a few other transforms that the UI does with this data that I think are relevant:

1) We look up the status of specific partitions in the 2D array (the health bar data also powers the status beside each partition in the list view).

2) We flatten 2D status along one of the two axes (eg: `[[1, 1], [1, 0]]` becomes `[1, partial]`)

3) We flatten the status of several assets together into one health bar. (sometimes after doing #1)

There are super fast algorithms to do those things with sets-of-ranges instead of matrices (and it does seem like it needs to happen eventually if we're going to have 10,000x10x000 multi-dimensional matrixes...) but it would be more complicated.

In thinking about making this all more efficient, I think we could:

A) Send down ranges instead of huge 2D arrays (this proposal) - would be a data transfer win, but with everything gzipped I'm not actually sure if it's 10x or 100x if it takes time to compute the spans. Having spans open up implementations of 1-3 above that are faster than O(n) and does generally seem better.

B) Allow asking for a subset of the data (eg: "last 1,000 on date axis"). I'm not sure how much users actually care about the partition health bar going back five years, etc. @sryza might know better! It's possible that at some point, we allow users to specify a "last 2 years rule" they want to apply to the UI, as the default for backfills, for "10/1000 missing" stats, etc.
   - On this same note, we currently ask for /all/ the partition keys on both dimensions in most places in the partitions UI. It could be nice to have a subset filter, or potentially send down the algorithm used to generate the keys if we think it'll remain a pretty discrete set of options (day, month, hour, static, etc)

C) Make a few more APIs that offer compacted / summary views so that we don't have to load the full matrix as often.

I think the vNext asset DAG is going to display summary partition info for every asset, but not a health bar, so a quick win might be adding a GraphQL API that flattens everything into just "# total", "# missing" (though future UI might be more ambitious!)

clairelin135:
@bengotow thanks for the detailed response! After chatting with Sandy it seems ideal to display the status of all partitions rather than a subset of partitions since forcing the user to specify how many partitions they'd like to see or paginating between multiple screens may introduce extra work for users. It seems ideal for performance to switch to assembling ranges instead.

Ranges get a bit hairy for multidimensionally partitioned assets though. Two options I thought of:

(1) Treat a multipartitioned asset like a static partitioned asset. For example, for an asset with partition dimensions: `{"ab": [a, b], "123": [1, 2, 3]}` would have a list of partition keys `['1|a', '1|b', '2|a', '2|b', '3|a', '3|b']`. If `2|a`, `2|b`, `3|a` were materialized, you would get a range `(2|a, 3|a)`. One consequence is that it might be kind of surprising that `2|b` is included in this range.

(2) Return ranges for each partition key of a dimension. So for the same example above where `2|a`, `2|b`, `3|a` were materialized, if we were grouping by dimension `123` we'd return `{'1': [] , '2': [('a', 'b')] '3': [('a', 'a')]}`

I'm curious how these options might perform with the transforms you mentioned. I think if we chose: 
- (1): each of the transforms (partition key lookup and flattening by dimension) you'd have to do an O(# partition keys) operation. 
- (2): looking up one partition key would be fast, but I think we might have to send a copy per dimension (i.e. send `{'1': [] , '2': [('a', 'b')] '3': [('a', 'a')]}` and `{'a': [(2, 3)], 'b': [(2, 2)]}` in order to optimize for the flattened views.

Not sure how the range algorithms you mentioned work, thoughts on which option might be better?

bengotow:
Hey @clairelin135! Hmm, I think it might make sense to optimize for the case where large "chunks" of the space have the same status (the `(2|a, 3|a)` idea), but my biggest concern is the second part you hit on. In order for this to be better than the status quo, we have to be able to encode and decode these _really_ fast, so fast it's not a problem doing it on 100k x 100k items. I did a bit of research last week, and in a 100k asset key test case most of the time was actually spent string formatting keys and the data transfer wasn't too bad.
 
Right now we have gzip compression turned on for the data transfer and long stretches of 1111111 or 00000 compress down nicely. The downside is that this compression is transport-only, and Dagit still inflates an array of 100k booleans... which will definitely break at some threshold of size X number of assets.

Any form of run-length encoding seems like it will switch look-ups from O(1) to a binary search of the range set, but I think that's the price we pay for the compression win. Only a few workflows in Dagit require the status of individual keys -- in the common case where we want to render the health bar or count items of a certain status, we could actually just render / operate on the run length encoded data, since we want "ranges" anyway!

  + This would require that in the 2D case, the outermost run-length encoded dataset is the time axis - right now we sometimes flip them locally so that the underlying data in Dagit is always the same as the backend's ordering I think.

Maybe we should try doing a simple run-length encoding and validate that we can do it fast enough in Python for it to be better than transmitting the giant array? For 2D assets, maybe we could start by doing RLE twice, first on each day's data, and then across days, (so if three Days RLE-encoded "dimension 2" data are identical, those become a range?)

sryza:
cc @bengotow 

For multi-partitions, do we necessarily need to fetch a data structure that represents all the individual partition statuses?

For the asset graph view, we just need the numbers of missing/filled/etc.

For the asset details page, there's the top bar, which can be collapsed / represented with a single 1-D range, and then there are the columns, which are essentially collapsible into 1-D ranges as well. Is that right?

clairelin135:
@bengotow I've updated this PR to change the materialization counts by partition gql endpoint to instead return a run-length encoded object containing just the subset of partition ranges that have been materialized. Following your suggestions:
- 1D partitions just contain a list of ranges that have been materialized. Each range is defined with a start and end partition value, both inclusive.
- 2D partitions encode by a primary dimension and a secondary dimension. The primary dimension is the time dimension if there is only one time dimension, otherwise whichever dimension is defined first.
  - For example, with a date dimension and a static dimension with partitions `[a, b, c]` you would get a range:
  - ```{primaryDimStart: "2022-01-01", primaryDimEnd: "2022-01-03", secondaryDimRanges: [{"start": "a", end: "b"}]}```
  - This range would represent that days Jan1-Jan3 have static partitions a and b materialized.
 
I did validate that these encode pretty quickly:
- 100K single-dimensional partitions will encode under 0.1 seconds
- 250K multidimensional partitions will encode under 0.4 seconds

For now the frontend will be broken until the client-side changes are updated. Let me know how this looks!

clairelin135:
@bengotow @sryza This PR now contains the changes we discussed yesterday! The description of the PR has been updated with the new changes contained.

bengotow:
Whoa awesome @clairelin135! That was fast :-) I'll try rebasing my work on this over the weekend 🙌

sryza:
> Materialized time windows are now returned back in ranges
I thought we decided not to do this?

clairelin135:
@sryza We are still returning back materialized time windows in ranges so we can quickly render the partition health bar.

The original problem with that was that with the ranges we'd still have to fetch all of the partition keys to display materialization per partition, but the solution we came up with was to still fetch the partition keys (as this is a quick enough operation) but be able to fetch just a subset of them based on the selection in the partition health bar.

sryza:
> [@sryza](https://github.com/sryza) We are still returning back materialized time windows in ranges so we can quickly render the partition health bar.

I thought we ascertained that rendering the partition health bar is fast enough without this change?  It seemed to take only a couple seconds or less when we were loading 26k partitions with that experiment on my computer.

clairelin135:
@sryza Depends on how many partitions we'd like to support. Fetching the partition keys of 100K time window partitions takes about 9 seconds (I'd assume 26K partitions is about 2s), and if we returned back all of materialized data by partition instead of ranges we'd have to incur the partition key fetch cost.

I figure that since we already have the data in terms of ranges we can return that content and be able to quickly render the time bar even for 100K partitions.

sryza:
@bengotow what do you think? I think 9 seconds for 100k partitions is fine if it's substantial work to switch the UI over the ranges (because 100k time partitions is pretty rare), but I have no opposition technical opposition to the ranges approach.

clairelin135:
@bengotow @sryza looks like the frontend changes in https://github.com/dagster-io/dagster/pull/11573 are close, please take a look at this PR again so we can merge the changes in together 🙌

python_modules/dagster-graphql/dagster_graphql/schema/pipelines/pipeline.py
```diff
@@ -87,27 +87,73 @@ def parse_time_range_args(args):
...
+    ranges = non_null_list(GrapheneTimePartitionRange)
 
     class Meta:
-        name = "MaterializationCountGroupedByDimension"
+        name = "TimePartitions"
 
 
-class GraphenePartitionMaterializationCounts(graphene.Union):
+class GrapheneDefaultPartitions(graphene.ObjectType):
+    materializedPartitions = non_null_list(graphene.String)
```
bengotow, line 108: I don't think it's worth changing now, but I think in a follow-up it might be nice to make this `materializedKeys` and `unmaterializedKeys` so it's more similar to `startKey` in the other types

python_modules/dagster-graphql/dagster_graphql/schema/pipelines/pipeline.py
```diff
@@ -87,27 +87,73 @@ def parse_time_range_args(args):
...
+    primaryDimEndTime = graphene.Field(graphene.Float)
+    secondaryDim = graphene.NonNull(GraphenePartitionStatus1D)
+
+    class Meta:
+        name = "MaterializedPartitionRange2D"
+
+
+class GrapheneMultiPartitions(graphene.ObjectType):
+    ranges = non_null_list(GrapheneMultiPartitionRange)
+
```
bengotow, line 140: Just wanted to call out that in my follow-up UI PR, I added another field here called "primaryDimensionName" so that I can tell from the UI-side whether the backend is re-ordering the dimensions to put time first.

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -128,15 +128,79 @@ def __new__(  # pylint: disable=arguments-differ
...
+
+                if time_window.end.timestamp() > current_timestamp:
+                    partitions_past_current_time += 1
+            else:
+                break
+
+        if self.end_offset < 0:
+            num_partitions += self.end_offset
+
+        return num_partitions
```
bengotow, line 163: Man it's really tough that we support full-complexity cron strings here 😅  This algorithm looks good to me knowing that we really have no choice but to walk through time.

In the UI we're going to request this for every asset in the asset graph along with other "live" asset info -- I think this is probably performant enough but if we find that it's not, I wonder if we could add some light caching around this function call. (maybe even cache the result in an instance var here?)

 eg: If I ask for the "num partitions" of 50 assets and all 50 of them reference the same partition definition, could we just count partitions in that definition once during the GraphQL request?

bengotow (approved):
This looks great @clairelin135 ! Took another read through -- I'm sure we'll make some tweaks as we continue to use more of these new features in the UI, but there's so much good stuff in here 👍

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -128,15 +128,79 @@ def __new__(  # pylint: disable=arguments-differ
             cls, start_dt, timezone or "UTC", fmt, end_offset, cron_schedule
         )
 
-    def get_partitions(
-        self, current_time: Optional[datetime] = None
-    ) -> Sequence[Partition[TimeWindow]]:
-        current_timestamp = (
+    def get_current_timestamp(self, current_time: Optional[datetime] = None) -> float:
+        # return pendulum.now(self.timezone).timestamp()
```
sryza, line 132: Remove this commented out line?

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -128,15 +128,79 @@ def __new__(  # pylint: disable=arguments-differ
...
+                    partitions_past_current_time += 1
+            else:
+                break
+
+        if self.end_offset < 0:
+            num_partitions += self.end_offset
+
+        return num_partitions
+
+    def get_partition_keys_between_idxs(
```
sryza, line 165: generally prefer to avoid abbreviations – idx -> index

sryza (commented):
I reviewed the parts in core. LGTM other than a couple small comments.

PR #10832, opened 2022-12-01
Add CLI command to delete runs older than date

joshuataylor:
### Summary & Motivation
Deleting older than a certain date is super useful in Dagster. We don't really care about jobs older than a certain date range (say 90 days), this just takes up storage space.

What would be useful is a command to delete older than a date.

See this issue https://github.com/dagster-io/dagster/issues/4497

I really like this idea: `U02B400J44U: Maybe `dagster run wipe` could take an argument like `--older-than=2w``, So I added it.

I really don't like adding this to wipe as wipe is an incredibly destructive action and I foresee users wiping their instances by mistake. Instead I've created a new command called `delete-range`. If someone has a better name for this, I'm open for suggestions as I don't like this name :).

Now you can do the following:
`dagster run delete-range 1h`

And it will ask you if you want to delete older than the date it found:
```bash
$ dagster run delete-range 1h
Are you sure you want to delete run history and event logs from 2022-12-01 18:58:57.162173? Type DELETE.
```

And after it will tell you how many deletions it made:
```bash
$ dagster run delete-range 1h
Are you sure you want to delete run history and event logs from 2022-12-01 18:58:57.162173? Type DELETE.: DELETE
Found 0 runs to delete.
0it [00:00, ?it/s]
Deleted run history and event logs older than 2022-12-01 18:58:57.162173.
```
It will also show you a progress bar which will update every run.

Questions:

1. Is there a better way to get Run IDs without having to load the entire run? I think this would be better for batching as you don't have to load more information to memory than you need.
2. How do I create runs with an arbitrary started time via tests?

Once I get #2 answered I'll finish the tests.

:thinking: Should there be a `delete_many` function added at some point to bulk delete runs? Right now you have to do run-by-run.

1. Manually on a local project we have to delete old runs
2. New tests

### DIFF
--- a/python_modules/dagster/dagster/_cli/run.py
+++ b/python_modules/dagster/dagster/_cli/run.py
@@ -1,7 +1,10 @@
+import datetime
+import re
+
 import click
 from tqdm import tqdm
 
-from dagster import __version__ as dagster_version
+from dagster import __version__ as dagster_version, RunsFilter
 from dagster._cli.workspace.cli_target import get_external_job_from_kwargs, job_target_argument
 from dagster._core.instance import DagsterInstance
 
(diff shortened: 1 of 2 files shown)


dpeng817:
buildkite build: https://buildkite.com/dagster/dagster/builds/40812. Will follow up with results

joshuataylor:
Thanks! If you also know the answers to my questions around setting the run start time, that'll be amazing as the tests should be fairly straight forward :-)

cmpadden:

Issue #10261, opened 2022-10-29
Intersect select queries to mysql storage broke jobs scheduling


paralyzah:
### Dagster version

1.0.15

### What's the issue?

```
sqlalchemy.exc.ProgrammingError: (mysql.connector.errors.ProgrammingError) 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'INTERSECT SELECT run_tags.run_id
FROM run_tags
WHERE run_tags.`key` = '.dagste' at line 5
[SQL: SELECT runs.run_body, runs.status
FROM runs
WHERE runs.run_id IN (SELECT run_tags.run_id
FROM run_tags
WHERE run_tags.`key` = %(key_1)s AND run_tags.value = %(value_1)s INTERSECT SELECT run_tags.run_id
FROM run_tags
WHERE run_tags.`key` = %(key_2)s AND run_tags.value = %(value_2)s) ORDER BY runs.id DESC]
```

INTERSECT clause became available at 2022-10-11 in MySQL 8.0.31 release. Only vanilla mysql have this feature at this moment, and only in 8.X.X releases.
How you can use this features, if you keeping backward compatibility with mysql 5.7?

### How to reproduce?
Updating to 1.0.14 or above

### Additional information

I guess https://github.com/dagster-io/dagster/pull/10036 is reason
Possible fix solution: INTERSECT clause emulation via OUTER JOIN's connected thru UNION clause

yuhan:
@prha might taking a look?

prha:
This is fixed as of `1.1.6`

PR #10838, opened 2022-12-01
add tag query support for older versions of mysql

prha:
### Summary & Motivation
Users reporting that older versions of mysql are raising errors, because support for the `intersect` command in mysql was only added in `8.0.31` (Oct 11, 2022).

We changed the default tag-based queries to use intersect in `1.0.14`, which was released Oct 20, 2022.  This is a pretty aggressive timeline to expect folks to support.

Our `mysql` test suite operates on the `mysql:8` tag, which is why our tests didn't catch this.

This PR adds a fixture to stand up both `mysql:8` and `mysql:8.0.30` (released July 26, 2022) instances to test against.

For https://github.com/dagster-io/dagster/issues/10261

BK


clairelin135 (approved):
LGTM! Thanks for adding this

Issue #3038, opened 2020-10-06
[dagster-k8s] remove utils.py ephemeral client instantiations


alangenfeld:
We shoudn't be newing up `DagsterKubernetesClient.production_client()` for single invocations

PR #10844, opened 2022-12-01
[k8s] use DagsterKubernetesClient directly

alangenfeld:
do this refactor that should have happened a couple years ago

We have this api client object set-up to facilitate testing and do retries on api methods, but we use it indirectly so its not clear there is a behavior difference unless you go digging a few layers.

resolves https://github.com/dagster-io/dagster/issues/3038

existing test coverage


alangenfeld:
* master
  * **PR #10844** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10844" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10844?utm_source=stack-comment).

johannkm (approved):

PR #10862, opened 2022-12-02
[dagstermill] render notebook in windows

jamiedemaria:
### Summary & Motivation
User reported an issue where notebooks were not rendered in windows. The issue was caused because the path to the notebook was completed with the `C:` prefix which the `URL` check in `NotebookButton.tsx` read as a valid URL. 

Fixing this by "cleaning" the notebook path set as the `notebook_path` tag and as metadata if the user is on Windows. If the user is on windows and the path has the `C:` prefix, we remove it. 

loaded dagit in a windows machine parallel on my computer and opened a notebook 

Screen Shot 2022-12-02 at 11 19 28 AM


python_modules/libraries/dagstermill/setup.py
```diff
@@ -47,11 +47,11 @@ def get_version() -> str:
         "packaging>=20.5",
         "papermill>=1.0.0",
         "scrapbook>=0.5.0",
+        "nbconvert",
```
jamiedemaria, line 50: rendering the notebook in dagit seems like a pretty nice feature to include by default, @yuhan what do you think of having nbconvert as a base dependency?

yuhan: there used to be some conflict dep. but if all the builds are running fine, i think it's fine

python_modules/libraries/dagstermill/dagstermill/factory.py
```diff
@@ -32,6 +32,15 @@
 from .errors import DagstermillError
 from .translator import DagsterTranslator
 
+IS_WINDOWS = os.name == "nt"
+
+
+def _clean_path_for_windows(notebook_path):
+    """In windows, the notebook cant render in dagit unless the C: prefix is removed"""
+    if IS_WINDOWS and ("c:" in notebook_path or "C:" in notebook_path):
```
yuhan, line 40: would it always be in C drive?

jamiedemaria: confirmed that on my windows machine i can have the notebook path without the `C:` prefix and the asset executes normally. So there could be a case when a user is on windows but specified the paths without `C:` so we wouldn't want to remove the first two characters of the path

yuhan: can we use [os.path](https://docs.python.org/3/library/os.path.html) to somehow trim wins drive prefix?

python_modules/libraries/dagstermill/dagstermill/factory.py
```diff
@@ -32,6 +32,15 @@
 from .errors import DagstermillError
 from .translator import DagsterTranslator
 
+IS_WINDOWS = os.name == "nt"
+
+
+def _clean_path_for_windows(notebook_path):
```
yuhan, line 38: worth adding type here since path can be in different types

python_modules/libraries/dagstermill/dagstermill/factory.py
jamiedemaria: yeah that's a much better idea

python_modules/libraries/dagstermill/dagstermill/factory.py
```diff
@@ -33,6 +33,13 @@
 from .translator import DagsterTranslator
 
 
+def _clean_path_for_windows(notebook_path: str) -> str:
+    """In windows, the notebook cant render in dagit unless the C: prefix is removed.
+    os.path.splitdrive will split the path into (drive, tail), so just return the tail
+    """
+    return os.path.splitdrive(notebook_path)[1]
```
yuhan, line 40: oh nice

yuhan (approved):

PR #10865, opened 2022-12-02
enforce multiasset topological order

smackesey:
### Summary & Motivation

This adds a check during step execution to ensure multiassets yield their constituent assets in topological order.

Without this check, the data version calculation is unreliable.

New unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/plan/execute_step.py
+++ b/python_modules/dagster/dagster/_core/execution/plan/execute_step.py
@@ -118,6 +118,39 @@ def _step_output_error_checked_user_event_sequence(
                     "must yield DynamicOutput, got Output."
                 )
 
+            # For any output associated with an asset, make sure that none of its dependent assets
+            # have already been yielded. If this condition (outputs yielded in topological order) is
+            # not satisfied, automatic data version computation can yield wrong results.
+            #
+            # We look for dependent keys that have already been yielded rather than dependency keys
+            # that have not yet been yielded. This is because we don't always know which
+            # dependencies will actually be computed within the step. If A depends on B, it is
+            # possible that a cached version of B will be used and B will never be yielded. In
+            # contrast, if both A and B are yielded, A should never precede B.
+            asset_layer = step_context.job_def.asset_layer
+            node_handle = step_context.node_handle
+            asset_info = asset_layer.asset_info_for_output(node_handle, output_def.name)
+            if (
+                asset_info is not None
+                and asset_info.is_required
+                and asset_layer.has_assets_def_for_asset(asset_info.key)
+            ):
+                assets_def = asset_layer.assets_def_for_asset(asset_info.key)
+                if assets_def is not None:
+                    all_dependent_keys = asset_layer.downstream_assets_for_asset(asset_info.key)
+                    step_local_asset_keys = step_context.get_output_asset_keys()
+                    step_local_dependent_keys = all_dependent_keys & step_local_asset_keys
+                    for dependent_key in step_local_dependent_keys:
+                        output_name = assets_def.get_output_name_for_asset_key(dependent_key)
+                        # Need to skip self-dependent assets (possible with partitions)
+                        if step_context.has_seen_output(output_name):
+                            raise DagsterInvariantViolationError(
+                                f'Asset "{dependent_key.to_user_string()}" was yielded before its'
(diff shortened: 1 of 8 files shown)


smackesey:
* master
  * **PR #10865** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10865" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10865?utm_source=stack-comment).

python_modules/dagster/dagster/_core/execution/plan/execute_step.py
```diff
@@ -122,6 +122,31 @@ def _step_output_error_checked_user_event_sequence(
...
+            if (
+                asset_info is not None
+                and asset_info.is_required
+                and asset_layer.has_assets_def_for_asset(asset_info.key)
+            ):
+                assets_def = asset_layer.assets_def_for_asset(asset_info.key)
+                if assets_def is not None:
+                    all_dep_keys = asset_layer.upstream_assets_for_asset(asset_info.key)
+                    step_local_asset_keys = assets_def.asset_keys
+                    step_local_dep_keys = all_dep_keys & step_local_asset_keys
```
OwenKephart, line 141: I think this logic breaks down for graph-backed assets, because not all of the asset keys in the definition may be produced in the same step.

might be misinterpreting, though, and I don't see an easy way to get the set of asset keys produced by a given step (rather than by a given node, where the node is the GraphDefinition).

Just for reference, if this is logic is incorrect, then https://sourcegraph.com/github.com/dagster-io/dagster/-/blob/python_modules/dagster/dagster/_core/execution/context/compute.py?L307&subtree=true is also incorrect, so I'm approving for now as this is not a net new issue, but I do think we should address the issue in a PR soon or make sure that it's not actually an issue.

cc @clairelin135 for your more seasoned opinion.

OwenKephart (approved):
heed comment, but this looks good.

python_modules/dagster/dagster/_core/execution/plan/execute_step.py
clairelin135: Yep, this logic would incorrectly throw an error when assets are yielded from different steps. For example, this code snippet below would throw an error. I can also imagine edge cases where an output name is shared by a graph and by an op.
```
@graph(out={"out": GraphOut(), "out_plus_one": GraphOut()})
def my_graph():
    x = upstream()
    return {"out_plus_one": downstream(x), "out": x}


AssetsDefinition.from_graph(
    my_graph, internal_asset_deps={"out_plus_one": {AssetKey("out")}, "out": set()}
)
```
I think what we can do instead is determine the `step_local_asset_keys` by finding all the assets outputted from the step from the [`asset_info_by_node_output_handle` field on the asset layer](https://github.com/dagster-io/dagster/blob/f37fd6cb2844689eea8b406a53254d5d96720ac0/python_modules/dagster/dagster/_core/definitions/asset_layer.py#L503). Presumably any asset dependencies have already outputted in upstream steps and thus have materializations.

OwenKephart (changes requested):
revoking the approve in light of claire's point

sryza (changes requested):
requesting changes for q management

python_modules/dagster/dagster/_core/execution/plan/execute_step.py
smackesey: After a long hiatus I've modified this PR to address the above feedback. But it looks like since I initially posted it, something has changed in `dagster-dbt`-- there is a test in there that is yielding outputs in non-topo-order. I don't know enough about `dagster-dbt` to quickly determine whether this is stemming from `dagster-dbt` itself or test design-- @OwenKephart mind having a look a the failing test?

OwenKephart:
@smackesey This is happening because that test applies config which further filters the set of assets from the multi asset (so fewer assets than expected will be materialized). The assets that are materialized will still be yielded in topological order. I think we should only error at the point in time that an asset key is produced that comes topologically before an asset key that was already yielded, although that's likely harder to implement

smackesey:
Thanks @OwenKephart -- suggestion makes sense. I implemented it and now we're green.

It required adding a new `downstream_assets_for_asset` method to `AssetLayer`. Do you think this should be cached (currently I construct it as needed by looking over `downstream_assets_for_asset`.

OwenKephart (approved):
nice

PR #10867, opened 2022-12-02
use logical version machinery when calculating data time

OwenKephart:
### Summary & Motivation

We have a more accurate way of calculating the upstream materialization for a given materialization, so let's use it!

If the data does not exist in the tags, defaults to old behavior. In theory, this should only happen when you're reading a record that was created pre-1.1.0.

The _extract_tags function is basically just there to make mypy happy, but helps to have a bit more invariant checking I suppose.

### DIFF
--- a/python_modules/dagster/dagster/_utils/caching_instance_queryer.py
+++ b/python_modules/dagster/dagster/_utils/caching_instance_queryer.py
@@ -5,7 +5,10 @@ from typing import TYPE_CHECKING, AbstractSet, Dict, Iterable, Mapping, Optional
 import dagster._check as check
 from dagster._core.definitions.asset_graph import AssetGraph
 from dagster._core.definitions.events import AssetKey, AssetKeyPartitionKey
+from dagster._core.definitions.logical_version import get_input_event_pointer_tag_key
 from dagster._core.errors import DagsterInvariantViolationError
+from dagster._core.event_api import EventRecordsFilter
+from dagster._core.events import DagsterEventType
 from dagster._core.storage.event_log import EventLogRecord, SqlEventLogStorage
 from dagster._core.storage.event_log.sql_event_log import AssetEventTagsTable
 from dagster._core.storage.pipeline_run import (
@@ -15,6 +18,7 @@ from dagster._core.storage.pipeline_run import (
     RunsFilter,
 )
 from dagster._core.storage.tags import PARTITION_NAME_TAG
+from dagster._core.utils import frozendict
 from dagster._utils.cached_method import cached_method
 from dagster._utils.merger import merge_dicts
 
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -383,3 +396,19 @@ def get_current_minutes_late_for_key(
             used_data_times=used_data_times,
             available_data_times={key: evaluation_time for key in used_data_times},
         )
+
+
+def _extract_tags(record: Optional[EventLogRecord]) -> Mapping[str, str]:
```
sryza, line 401: I think we should make this a `materialization_tags` property on `EventLogRecord`. Or add a `materialization` property to `EventLogRecord` and pull the tags off of it in the caller.

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -383,3 +396,19 @@ def get_current_minutes_late_for_key(
...
+    if dagster_event is None:
+        raise DagsterInvariantViolationError(
+            "Attempted to extract tags from record with no dagster event."
+        )
+    if dagster_event.event_type_value != DagsterEventType.ASSET_MATERIALIZATION:
+        raise DagsterInvariantViolationError(
+            "Attempted to extract tags from record with dagster event of type "
+            f"{dagster_event.event_type_value}."
+        )
+    return frozendict(dagster_event.step_materialization_data.materialization.tags or {})
```
sryza, line 414: Why use `frozendict`? Can we rely on the type system to enforce rather than doing a copy?

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -299,20 +300,31 @@ def calculate_used_data(
...
                         upstream_required_keys.add(upstream_key)
 
-                # get the most recent asset materialization for this parent which happened before
-                # the current record
-                latest_parent_record = self.get_latest_materialization_record(
-                    parent_key, before_cursor=record_id
-                )
+                input_event_pointer_tag = get_input_event_pointer_tag_key(parent_key)
+                if input_event_pointer_tag in record_tags:
+                    # get the asset materialization record that was calculated using the logical
```
sryza, line 305: nitpick: even though we added this tag in the same PRs that we added logical versioning, it stands on its own independent of logical versioning. I think this comment makes what's going on here sound fancier / more confusing than what actually is.

sryza (commented):
Is there a test that would make sense to add here? I think the behavior would be different in one of those race condition scenarios, right?

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
OwenKephart: this is because every argument passed into a cached_method needs to be hashable

sryza: Ahh, got it. Fine in that case.

sryza (approved):

PR #10873, opened 2022-12-02
Replace partition range methods with partition subset methods

clairelin135:
In this PR, I created a multidimensional -> singledimensional partition mapping, where the upstream asset partition definition is one dimension of the downstream multipartitioned asset.

In creating this, I realized that partition ranges are insufficient. For example:
```
MultiPartitionsDefinition(
)
```
has partition keys: `['1|a', '1|b', '2|a', '2|b']`

If this asset had an upstream asset with partitions definition `StaticPartitionsDefinition(["a", "b"])`, the downstream partition keys should be `['1|a', '2|a']` but this can't exist as a range of the partition key list.

This PR amends the existing methods that fetch upstream/downstream partitions ranges to instead work with subsets. This means that `PartitionMapping`s should return subsets instead of ranges. In order to not break users, this PR introduces the following methods:
- `get_upstream_partitions_for_partition_subset`, which reads from `get_upstream_partitions_for_partition_range`
- `get_downstream_partitions_for_partition_subset` which reads from `get_downstream_partitions_for_partition_range`

In the future, at the next major release, I think we should deprecate the `range` methods on `PartitionMappings` and replace these all with `subset` methods. This will enable future use cases like MultiPartition assets that depend on differently MultiPartitioned assets.


python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -224,27 +222,15 @@ def get_child_partition_keys_of_parent(
...
-            partition_mapping.get_downstream_partitions_for_partition_range(
+        downstream_partition_key_subset = (
+            partition_mapping.get_downstream_partitions_for_partition_subset(
                 PartitionKeyRange(parent_partition_key, parent_partition_key),
                 downstream_partitions_def=child_partitions_def,
                 upstream_partitions_def=parent_partitions_def,
             )
         )
 
-        partition_keys = child_partitions_def.get_partition_keys()
```
clairelin135, line 235: These checks will now occur when attempting to build a partitions subset from a partition key range: https://github.com/dagster-io/dagster/blob/305c6866ac00b23fae4646751eb4c03389af395d/python_modules/dagster/dagster/_core/definitions/partition.py#L228

PR #10878, opened 2022-12-02
fix subsettable multi asset case

OwenKephart:
### Summary & Motivation

Previously, you'd have the chance to encounter an error if you tried to use freshness-based scheduling when working with non-subsettable multi assets. For example, imagine you had a bunch of airbyte assets at the top of your graph, then an asset with a freshness policy that consumes just a single one of those tables. The existing logic would request a run of just that asset and its upstream dependency, which would be an invalid subset.

"required_neighbor_sets" isn't a great name, open to alternatives.

We'll need to serialize a bit of extra information on the ExternalAssetNode in order for the necessary information to cross the serialization boundary, but that shouldn't be a big deal. For now I left this out to keep the PR small.

Added two tests, the one where the multi asset was allowed to be subset passed before, but the other one failed. Now both pass.


python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -32,40 +33,14 @@
...
-        return super(AssetGraph, cls).__new__(
-            cls,
-            asset_dep_graph=asset_dep_graph,
-            source_asset_keys=source_asset_keys,
-            partitions_defs_by_key=partitions_defs_by_key,
-            partition_mappings_by_key=partition_mappings_by_key,
-            group_names_by_key=group_names_by_key,
-            freshness_policies_by_key=freshness_policies_by_key,
-        )
+class AssetGraph(NamedTuple):
```
sryza, line 36: <3 the condensed format

python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -346,6 +327,17 @@ def upstream_key_iterator(self, asset_key: AssetKey) -> Iterator[AssetKey]:
                     queue.append(parent_key)
                     visited.add(parent_key)
 
+    def get_required_neighbors(self, asset_key: AssetKey) -> AbstractSet[AssetKey]:
+        """For a given asset_key, return the set of asset keys that must be materialized at the same time."""
+        if self.required_neighbor_sets is None:
+            raise DagsterInvariantViolationError(
+                "Required neighbor information not set when creating this AssetGraph"
+            )
+        for required_assets in self.required_neighbor_sets:
```
sryza, line 336: If you had a bunch of multi-assets, this would be O(# assets), right? Thoughts on using a dict?

OwenKephart: mm yeah -- originally I was thinking we'd need to create some unique identifier for the keys of the dictionary, then some mapping from asset key to that unique identifier, but now I realize that's not necessary at all. will update.

sryza (commented):
This looks great. My final piece of feedback is that "required neighbors" feels kind like it could refer to a lot of different things. Maybe something like "atomic materialization units" would be more clear and specific? Or "non-subsettable multi-assets"?

sryza (approved):
Nice

PR #10882, opened 2022-12-05
[dagster-airflow] use named params and rename use_emphemeral_airflow_db

Ramshackle-Jamathon:

### DIFF
--- a/python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
+++ b/python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
@@ -53,7 +53,7 @@ def make_dagster_job_from_airflow_dag(
             framework authors to enforce unique op names within a repo.
         mock_xcom (bool): If True, dagster will mock out all calls made to xcom, features that
             depend on xcom may not work as expected. (default: False)
-        use_ephemeral_airflow_db (bool): If True, dagster will create an emphemeral sqlite airflow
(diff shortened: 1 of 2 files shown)


python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
```diff
@@ -47,26 +47,31 @@ def make_dagster_job_from_airflow_dag(
...
             Airflow TaskInstance model which requires and modifies the DagRun table. The use_airflow_template_context
-            setting is ignored if use_emphemeral_airflow_db is True.
+            setting is ignored if use_ephemeral_airflow_db is True.
             (default: False)
         unique_id (int): If not None, this id will be postpended to generated op names. Used by
             framework authors to enforce unique op names within a repo.
         mock_xcom (bool): If True, dagster will mock out all calls made to xcom, features that
             depend on xcom may not work as expected. (default: False)
-        use_emphemeral_airflow_db (bool): If True, dagster will create an emphemeral sqlite airflow
+        use_ephemeral_airflow_db (bool): If True, dagster will create an emphemeral sqlite airflow
```
rexledesma, line 56:
```suggestion
        use_ephemeral_airflow_db (bool): If True, dagster will create an ephemeral sqlite airflow
```

rexledesma (approved):
_technically_ should do a deprecation cycle, but I think this should be fine.

PR #10886, opened 2022-12-05
Check scheduler ticks right after each minute boundary instead of once every 5 seconds

gibsondan:
Summary:
A user reported that they were seeing between 1 and 6 seconds of latency when running the scheduler even when running a single simple schedule - this probably stems from our 5 second sleep interval. We have a simple tweak that we can do to try to maximize the chance that a tick will happen right after a 5-second boundary window.

The other change here is to include work that happens outside of the iterator in the timer (e.g. writing the heartbeat), which I did for sensors as well.

Test Plan:
Existing scheduler/sensor suite
Locally, run scheduler with a time.sleep(1) inside the heartbeat write function - see that ticks still happen every 5 seconds instead of every 6 seconds

### DIFF
--- a/python_modules/dagster/dagster/_daemon/sensor.py
+++ b/python_modules/dagster/dagster/_daemon/sensor.py
@@ -254,6 +254,10 @@ def execute_sensor_iteration_loop(
                 sensor_state_lock=sensor_state_lock,
                 log_verbose_checks=verbose_logs_iteration,
             )
+            # Yield to check for heartbeats in case there were no yields within
+            # execute_sensor_iteration
+            yield None
+
             end_time = pendulum.now("UTC").timestamp()
 
             if verbose_logs_iteration:
@@ -262,7 +266,6 @@ def execute_sensor_iteration_loop(
             loop_duration = end_time - start_time
             sleep_time = max(0, MIN_INTERVAL_LOOP_TIME - loop_duration)
             time.sleep(sleep_time)
-            yield None
 
 
 def execute_sensor_iteration(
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster/_daemon/sensor.py
```diff
@@ -254,6 +254,7 @@ def execute_sensor_iteration_loop(
                 sensor_state_lock=sensor_state_lock,
                 log_verbose_checks=verbose_logs_iteration,
             )
+            yield None
```
alangenfeld, line 257: comment why yield here

python_modules/dagster/dagster/_scheduler/scheduler.py
```diff
@@ -141,15 +141,20 @@ def execute_scheduler_iteration_loop(
...
-            time.sleep(sleep_time)
-            yield
+            # wait until at least the next time that's 0 mod MIN_INTERVAL_LOOP_TIME,
+            # which reduces latency since cron schedules will always happen on that boundary.
+            # if that time has already passed, then go again immediately.
+            next_time = (
+                start_time - (start_time % MIN_INTERVAL_LOOP_TIME)
+            ) + MIN_INTERVAL_LOOP_TIME
+            if next_time > end_time:
+                time.sleep(next_time - end_time)
```
alangenfeld, line 157: I stared at this for a while and do not feel like i am certain I get it - at minimum give the copy a second pass. Target audience: dummies like me. 

Maybe factor this out with and write some demonstrative test cases that show whats going on?

alangenfeld (approved):

gibsondan:
@alangenfeld new implementation that I think is both clearer and also potentially more performant? Shoots for the minute boundary since no schedule today will ever not be on a minute boundary anyway

alangenfeld (approved):
nice

PR #10892, opened 2022-12-05
[snowflake-pandas] show io manager in api docs

jamiedemaria:
### Summary & Motivation
I wanted to get the `snowflake_pandas_io_manager` to show in our API docs. It was weirder than expected, so here's a rundown of everything i tried so there's more context as to why I landed on this solution

**Attempt 1**: 
didn't change the code, just add to sphinx rst file 
the code looked like this
```
def wrap_io_manager(fn) -> IOManagerDefinition:
    return fn()


@wrap_io_manager
def snowflake_pandas_io_manager():
    """
    <doc string>
    """
    return build_snowflake_io_manager([SnowflakePandasTypeHandler()])
```

which resulted in API docs like this (no doc string, just the config dropdown)
Screen Shot 2022-12-05 at 3 07 17 PM

**Attempt 2**: 
I tried using the `functools.wraps` decorator because that will "promote" the doc string of the wrapped function to the wrapper

code snippet:
```
def wrap_io_manager(fn) -> IOManagerDefinition:
    @wraps(fn)
    def wrapper(*args, **kwargs):
        return fn()

    return wrapper

@wrap_io_manager
def snowflake_pandas_io_manager():
    """
    <doc string>
    """
    return build_snowflake_io_manager([SnowflakePandasTypeHandler()])

```

which resulted in API docs like this (how it should be!)

Screen Shot 2022-12-05 at 3 12 10 PM

However, this made it so you had to specify the io manager as `snowflake_pandas_io_manager()` in your code, which makes it out of sync with other io managers, which are specified `fs_io_manager` (no parentheses)

**Attempt 3**:
So then i tried something kinda wacky and returned the called wrapper from `wraps_io_manager`
```
def wrap_io_manager(fn) -> IOManagerDefinition:
    @wraps(fn)
    def wrapper(*args, **kwargs):
        return fn()

    return wrapper() # the change is here

@wrap_io_manager
def snowflake_pandas_io_manager():
    """
    <doc string>
    """
    return build_snowflake_io_manager([SnowflakePandasTypeHandler()])

```

But that resulted in the same issue as attempt 1

Screen Shot 2022-12-05 at 3 07 17 PM

**Attempt 4**:
so then i dug into the sphinx docs a bit. Our custom doc generator for IOManagers is based off of the sphinx DataDocumenter and it seemed like it should be capable of documenting a module scoped variable, so I tried 

```
snowflake_pandas_io_manager = build_snowflake_io_manager([SnowflakePandasTypeHandler()])
"""
<doc string>
"""
```

This also didn't work 
Screen Shot 2022-12-05 at 3 07 17 PM

**Attempt 5**:
But then i was curious if i could directly set the `__doc__` attribute of the variable, so i tried

```
snowflake_pandas_io_manager = build_snowflake_io_manager([SnowflakePandasTypeHandler()])
snowflake_pandas_io_manager.__doc__ = """<doc string>"""
```
And that worked! and the tests pass
Screen Shot 2022-12-05 at 3 12 10 PM

Let me know what you think of doing it this way. It's maybe a bit odd (i'm not sure what the standards are around directly setting the doc attr), but it seems to be working. I'm going to convert the duckdb one in the same way since the test coverage of that library is a bit more extensive, so it'll be more likely to catch issues


schrockn (approved):
Setting `__doc__` seems very explicit and straightforward to me

PR #10898, opened 2022-12-05
Also run tests when ipynb, yml files change

jmsanders:
We had a change to our dagstermill examples that didn't run our dagstermill tests.

This makes sure we run tests even if the only changes are to .ipynb files or to YAML files (like if a dagster.yaml or a docker-compose.yml) change.

- Add a commit that changes each of these file types, ensure the right dagstermill tests run, revert the change.

### DIFF
--- a/python_modules/libraries/dagstermill/dagstermill/examples/workspace.yaml
+++ b/python_modules/libraries/dagstermill/dagstermill/examples/workspace.yaml
@@ -1,4 +1,3 @@
----
 load_from:
   - python_module:
       module_name: dagstermill.examples.repository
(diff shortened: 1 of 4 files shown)


python_modules/libraries/dagstermill/dagstermill/examples/workspace.yaml
```diff
@@ -1,3 +1,4 @@
+---
```
jamiedemaria, line 1: is this from the testing?

jamiedemaria: oh yeah that's what's testing rn :)

jamiedemaria (approved):
yay!

PR #10900, opened 2022-12-05
update black 22.3.0 -> 22.12.0 and enable --preview

smackesey:
### Summary & Motivation

Updates `black` from 22.3.0 to 22.10.0 and enables `--preview` mode.

The version bump gives a small performance boost and bugfixes. Enabling `--preview` mode uses tweaks to the black style that are highly likely to be included in next major release, it's kind of like a conservative "enable-experimental-features". The rule changes that make up 99% of changes in this changeset include:

- Formatting of long string literals. There are a lot of these in the repo that had to be manually adjusted before, so this is a win. 
- Parens around return annotations now managed
- Deletion of blank lines at the start of a code block
- Several cases where extraneous parentheses are removed (awaited objects, with statements)

Comprehensive list is in [Black's changelog](https://black.readthedocs.io/en/stable/change_log.html).

### How I Tested These Changes

Existing BK tests.

Internal companion PR: https://github.com/dagster-io/internal/pull/4323

PR #10901, opened 2022-12-05
Adopt ruff as linter

smackesey:
### Summary & Motivation

Adopt [Ruff](https://github.com/charliermarsh/ruff) as our linter/import sorter.

Benefits. Ruff:

- Replaces pylint and isort.
- Can autofix many detected errors, functionality that `pylint` lacks completely.
- Seems 100-1000x faster than pylint (bc it's written in Rust instead of Python). This enables immediate in-editor feedback (there is a VSCode extension and you can also run in a generic language server like `null-ls`) and running on the whole repo as a pre-commit hook. `ruff \`git ls-files '*.py'\`` runs on my dev machine near-instantaneously.
- Implements rules not implemented in pylint (from flake8 and other Python linters).
- Is very actively developed. We can probably get items on our linter wishlist seriously considered and probably implemented on a relatively short timeline. I have already had success with this with Ruff's creator, @charliermarsh. In contrast stuff sits on isort/pylint's issue tracker for years.
- Allows us to run a single repo-wide lint step in BK instead of individual steps for every package (since ruff does not depend on Python env).
- Has rules to lint/format docstrings, something we have no tool for at present.

Current limitations. Ruff:

- Does not yet support all of pylint's rules, (but it is moving towards that goal).
- Does not yet support human-readable aliases for the numeric error flake8-style error codes it uses. These are much less readable than pylint codes.
- Does not yet support in-file file/block/function-scoped error-specific ignores (e.g. the way you can put `# pylint: disable=print-call` etc at the top of a file to disable that rule for the whole file. Instead there are just line-scoped error-specific ignores and file-scoped total opting out of linting. You can also use the config file to ignore specific errors for specific file patterns, but this is often a lot less convenient than in-file comments.
- Does not yet support all of `isort`'s action comments (like `# isort: split`, which we use in a few places.
- Does not yet support custom plugins. We have a custom pylint plugin with a few rules related to `pendulum` that can't be ported.

Notes:

- This PR changes a ton of files but most of these are corrections to newly detected lint errors and import sorting tweaks. Reviewers should look at the individual commits:
    - Updates to dagster dependencies, tool config in pyproject.toml, dev environment setup script, Makefile.
    - Lint error fixes (the vast bulk of changes)
    - Some questionable lint error fixes (removed some unused imports in integration tests that might have been there for a reason)
    - Toxfile changes dropping pylint
    - Buildkite changes dropping pylint/isort and introducing ruff.
- I changed the structure of all of our export-only `__init__.py` files to match the style used in `dagster/__init__.py`, where there is no `__all__` and exports are redundantly aliased. This is important to allow ruff to auto-detect and auto-delete unused imports (redundant alias marks the import as both public for type checkers and "used" for ruff).
- Because of the lack of support for file/block/function-scoped error-specific ignores, I've left our `# pylint` comments in the code until ruff supports this and we can bulk-convert them to ruff equivalents.
- You'll notice there is now a single ruff step in BK instead of the different `pylint` steps for each package. This makes immediate BK output a bit less informative, but saves a lot of computation. You can just run `ruff` locally to get immediate output if the Ruff step fails.
- Ruff step won't pass BK right now because I'm waiting on support for certain `# isort: off` comments that used in docs snippets. Will probably be supported in a few days.

BK.


smackesey:
* master
    * **PR #10901** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10901" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10901?utm_source=stack-comment).

charliermarsh:
Awesome, really cool PR summary :)

> Ruff step won't pass BK right now because I'm waiting on support for certain # isort: off comments that used in docs snippets. Will probably be supported in a few days.

Hopefully these are addressed in [v0.0.161](https://github.com/charliermarsh/ruff/releases/tag/v0.0.161) (building now), which includes the unterminated `# isort: off` comments, plus support for `# isort: split` and `# isort: skip_file`. I tried to test on the Dagster codebase where applicable, but please continue to file issues or even just ping me here if you run into other blockers.

examples/docs_snippets/docs_snippets_tests/concepts_tests/types_tests/test_types.py
```diff
@@ -1,7 +1,7 @@
 import pytest
+from docs_snippets.concepts.types.types import test_dagster_type
```
schrockn, line 2: Do we know why it is ordering dagster after docs_snippets

examples/project_fully_featured/project_fully_featured/__init__.py
```diff
@@ -1 +1 @@
-from .repository import hacker_news_repository
+from .repository import hacker_news_repository as hacker_news_repository
```
schrockn, line 1: Do we know why it is doing this? This might confuse users in public examples

schrockn (changes requested):
Super excited for this.

One suggestion here if it is a tractable amount of work to break this up.

1) Switch to ruff, but disable rules that are causing a lot of thrash in this PR/introduce new behavior.
2) Enable rules one by one. Then we can discuss implications like enabling the re-export rule on public examples.

If that process is too painful we can submit the whole thing, but if tractable I think that would be preferable.

examples/project_fully_featured/project_fully_featured/__init__.py
smackesey: `ruff` is not doing this-- I added the redundant aliases. Redundant aliasing is a more easily maintainable alternative to `__all__` that we already use it top level Dagster `__init__.py`-- there are some notes on it there: https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/__init__.py

Two main purposes here:

- mark the imports as public symbols for `pyright`
- prevent deletion of them as "unused" by `ruff`

smackesey:
@schrockn 

>One suggestion here if it is a tractable amount of work to break this up.

>Switch to ruff, but disable rules that are causing a lot of thrash in this PR/introduce new behavior.
>Enable rules one by one. Then we can discuss implications like enabling the re-export rule on public examples.

I can do it if you think it's necessary (this is what I did on the upstack docs PR, inidividual commits correspond to rules) but it would a lot of work, and I think the list I included in the description covers 95% of the changes (the bulk of which are import sorting changes)-- can we not just discuss those?

examples/docs_snippets/docs_snippets_tests/concepts_tests/types_tests/test_types.py
smackesey: Yes-- it is because `dagster` is right now recognized as the only "first-party" package. There is an extended discussion of this here with charlie here: https://github.com/charliermarsh/ruff/issues/1069#issuecomment-1339250929

Charlie is working on a solution that will get the current package treated as first-party. Until that lands (and some other relevant config tweaks) I decided to no try to force it with config, since previously isort was doing the same thing for all our other packages (if you look at e.g. `dagster-graphql` you'll see that `dagster` imports go below `dagster_graphql` imports).

examples/project_fully_featured/project_fully_featured/__init__.py
schrockn: Let's keep this out of the this diff then. These examples are not really meant for public exports, and I think this will just cause confusion

schrockn:
@smackesey got it re: the restack. 

At a minimum manual changes like adding the redundant aliases to public example should be removed from this and considered separately, since they affect content.

charliermarsh:
> Running ruff via your editor inside `docs_snippets` will probably return different results than running `make ruff` and `make check_ruff`.

This is fixed upstream (but not yet released). Ruff will now find the closest `pyproject.toml` file in the hierarchy, for every Python file: https://github.com/charliermarsh/ruff#pyprojecttoml-discovery.

schrockn (approved):
Incredibly. Per live conversation.

1) Most of these changes are quite machine-computable.
2) This will be announced to eng before it lands.

PR #10904, opened 2022-12-05
[dagit] Move op graphs behind “Show per-asset status”

bengotow:
### Summary & Motivation

This is a small PR that moves the "Run Duration" and "Step Duration" graphs on the Asset Partitions page behind the "Show per-asset status" button.

The rationale for this is that these graphs make expensive queries and slow down the overall page. On the corresponding op job page, only step duration is hidden until you "Show per-op status". The slight difference in behavior is due to the fact that the asset run data isn't loaded at all for the health bar.

Note: I'm not wild about the "Show per-asset status" button overall. When we get through the backlog the plan is to more broadly redesign this page!

### DIFF
--- a/js_modules/dagit/packages/core/src/partitions/AssetJobPartitionsView.tsx
+++ b/js_modules/dagit/packages/core/src/partitions/AssetJobPartitionsView.tsx
@@ -139,16 +139,18 @@ export const AssetJobPartitionsView: React.FC<{
           </Box>
         )}
       </Box>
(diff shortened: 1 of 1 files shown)


hellendag (approved):
Mind adding a test plan? I know it's a simple change, but it's worth a brief manual test plan. :)

bengotow:
You bet! Need to get in the habit of using the PR template... will make a plan and run through it real quick.

PR #10907, opened 2022-12-06
Add and apply ruff docstring rules

smackesey:
### Summary & Motivation

Enable Ruff's docstring formatting rules and apply them.

Only a subset of the total set of rules that apply to Google-style
docstrings are enabled-- for some, there were too many existing
violations and we should enable them when those violations are fixed.
For others, there are bugs in the detection and/or fix implementation,
so we'll leave them disabled until those are fixed.

BK.


smackesey:
* master
      * **PR #10907** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10907" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10907?utm_source=stack-comment).

schrockn:
forcing people to add periods in docblocks might feel a tad pedantic?

smackesey:
>forcing people to add periods in docblocks might feel a tad pedantic?

I would agree except for the fact that doc blocks are presented to our users in both our API docs and language servers. IMO it's important these appear polished and grammatically correct.  There is no easy way I'm aware of to get ruff to distinguish between docstrings that will be presented to users and those that won't. In any case ruff will automatically add missing periods so if you do format-on-save or have a format keybinding that combines black with ruff it's really no hassle.

schrockn:
Ok got it. I'd like to see that in a sep PR and then we can poll a few folks to take their temp on it.

schrockn (approved):

PR #10909, opened 2022-12-06
Replace partition ranges with subsets

clairelin135:
After multidimensional partition definitions were introduced, partition ranges are insufficient in many cases. For example:
```
MultiPartitionsDefinition(
)
```
has partition keys: ['1|a', '1|b', '2|a', '2|b']

If this asset had an upstream asset with partitions definition StaticPartitionsDefinition(["a", "b"]), the downstream partition keys should be ['1|a', '2|a'] but this can't exist as a range of the partition key list.

This PR amends the existing methods that fetch upstream/downstream partitions ranges to instead work with subsets. This means that PartitionMappings should return subsets instead of ranges. In order to not break users, this PR introduces the following methods:
- get_upstream_partitions_for_partition_subset, which reads from get_upstream_partitions_for_partition_range
- get_downstream_partitions_for_partition_subset which reads from get_downstream_partitions_for_partition_range

In the future, at the next major release, I think we should deprecate the range methods on PartitionMappings and replace these all with subset methods. This will enable future use cases like MultiPartition assets that depend on differently MultiPartitioned assets.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/partition.py
+++ b/python_modules/dagster/dagster/_core/definitions/partition.py
@@ -14,6 +14,7 @@ from typing import (
     NamedTuple,
     Optional,
     Sequence,
+    Set,
     TypeVar,
     Union,
     cast,
@@ -247,7 +248,7 @@ class PartitionsDefinition(ABC, Generic[T]):
         ]
 
     def empty_subset(self) -> "PartitionsSubset":
-        return DefaultPartitionsSubset(self, [])
+        return DefaultPartitionsSubset(self, set())
 
     def deserialize_subset(self, serialized: str) -> "PartitionsSubset":
         return DefaultPartitionsSubset.from_serialized(self, serialized)
@@ -1023,6 +1024,16 @@ class PartitionsSubset(ABC):
         raise NotImplementedError()
 
     @abstractmethod
+    def get_partition_keys(self, current_time: Optional[datetime] = None) -> Iterable[str]:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def get_partition_key_ranges(
+        self, current_time: Optional[datetime] = None
+    ) -> Sequence[PartitionKeyRange]:
+        raise NotImplementedError()
+
+    @abstractmethod
     def with_partition_keys(self, partition_keys: Iterable[str]) -> "PartitionsSubset":
         raise NotImplementedError()
 
@@ -1030,9 +1041,15 @@ class PartitionsSubset(ABC):
     def serialize(self) -> str:
         raise NotImplementedError()
 
+    @property
+    @abstractmethod
+    def partitions_def(self) -> PartitionsDefinition:
+        raise NotImplementedError()
+
 
 class DefaultPartitionsSubset(PartitionsSubset):
-    def __init__(self, partitions_def: PartitionsDefinition, subset=None):
+    def __init__(self, partitions_def: PartitionsDefinition, subset: Optional[Set[str]] = None):
+        check.opt_set_param(subset, "subset")
         self._partitions_def = partitions_def
         self._subset = subset or set()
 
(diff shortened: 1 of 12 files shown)


python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -54,6 +55,76 @@ def get_downstream_partitions_for_partition_range(
                 upstream asset.
         """
 
+    @public
+    def get_upstream_partitions_for_partition_subset(
+        self,
+        downstream_partition_key_subset: Optional[Union[PartitionKeyRange, PartitionsSubset]],
```
sryza, line 61: Nitpick: `partition_key_subset` -> `partitions_subset`

python_modules/dagster/dagster/_core/definitions/partition.py
```diff
@@ -1030,9 +1035,17 @@ def with_partition_keys(self, partition_keys: Iterable[str]) -> "PartitionsSubse
     def serialize(self) -> str:
         raise NotImplementedError()
 
+    def __eq__(self, other):
+        return (
+            isinstance(other, PartitionsSubset)
+            and self.get_partition_keys_not_in_subset == other.get_partition_keys_not_in_subset
```
sryza, line 1041: I think these need parentheses

python_modules/dagster/dagster/_core/definitions/partition.py
```diff
@@ -1041,12 +1054,23 @@ def get_partition_keys_not_in_subset(
...
 
     def serialize(self) -> str:
         return json.dumps(list(self._subset))
 
+    def __eq__(self, other):
+        return (
+            isinstance(other, DefaultPartitionsSubset)
+            and self._partitions_def == other._partitions_def
+            and self._subset == other._subset
+            and self._subset == other._subset
```
sryza, line 1071: the same condition shouldn't be repeated twice, right?

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -262,6 +263,18 @@ def get_partition_keys_in_range(self, partition_key_range: PartitionKeyRange) ->
         start_time = self.start_time_for_partition_key(partition_key_range.start)
         end_time = self.start_time_for_partition_key(partition_key_range.end)
 
+        partition_keys = self.get_partition_keys()
```
sryza, line 266: Calling `get_partition_keys` is expensive and makes an operation that's otherwise O(size of range), O(size of partition set). Is it possible to avoid it?

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -1004,6 +1017,14 @@ def get_partition_keys_not_in_subset(
 
         return result
 
+    def get_partition_keys(self, current_time: Optional[datetime] = None) -> Iterable[str]:
+        partition_keys_not_in_subset = set(self.get_partition_keys_not_in_subset())
+        return [
+            key
+            for key in self._partitions_def.get_partition_keys(current_time=current_time)
```
sryza, line 1024: Same comment as above about `get_partition_keys`. Could we instead iterate over the windows in the subset and return the partition keys in each window, without iterating over partition keys that are not in any of the windows?

sryza (commented):
This makes a lot of sense - I probably should have done this in the first place.

My main concerns here are about perf, and I left a few comments. I think it's important to make sure we're operating on ranges, not individual partitions, as much as possible.

To make sure we're not unpacking/repacking partitions unnecessarily, could we remove `TimeWindowPartitionMapping.get_upstream_partitions_for_partition_range` and add `TimeWindowPartitionMapping.get_upstream_partitions_for_partition_subset`.

Issue #10722, opened 2022-11-23
PartitionMapping between multi-dimensional partitions and single-dimensional partitions


sryza:
something like:
```python
class PartMap(PartitionMapping):
    def __init__(self, partition_dimension):
        self.partition_dimension = partition_dimension

    def get_downstream_partitions_for_partition_range(
        self,
        upstream_partition_key_range: PartitionKeyRange,
        downstream_partitions_def: Optional[PartitionsDefinition],
        upstream_partitions_def: PartitionsDefinition,
    ) -> PartitionKeyRange:
        raise NotImplementedError()
        return upstream_partition_key_range

    def get_upstream_partitions_for_partition_range(
        self,
        downstream_partition_key_range: Optional[PartitionKeyRange],
        downstream_partitions_def: Optional[PartitionsDefinition],
        upstream_partitions_def: PartitionsDefinition,
    ) -> PartitionKeyRange:

        # assuming only single value
        assert downstream_partition_key_range.start == downstream_partition_key_range.end
        multi_partition_key = downstream_partition_key_range.start
        upstream_partition_key = multi_partition_key.keys_by_dimension.get(self.partition_dimension)
        return PartitionKeyRange(upstream_partition_key, upstream_partition_key)
```

https://dagster.slack.com/archives/C01U954MEER/p1669227401243899?thread_ts=1669166628.531229&cid=C01U954MEER

bartek-hitachi:
Can we also get a mapping in the reverse direction to fully address this issue, please? By that I mean a: multidimension partition -> single dimension mapping. That would be very useful to have for something that I am currently working on :)

olexiyb:
I do not understand how this suppose to work, simple test
```
import dagster
from dagster import StaticPartitionsDefinition, MultiPartitionsDefinition, define_asset_job

upstream_partitions_def = StaticPartitionsDefinition(["up1", "up2"])
downstream_partitions_def = MultiPartitionsDefinition(
    {"part1": upstream_partitions_def, "part2": StaticPartitionsDefinition(["down1", "down2"])}
)


@dagster.asset(partitions_def=upstream_partitions_def)
def t01_part1_asset():
    pass


@dagster.asset(partitions_def=downstream_partitions_def)
def t01_part1_part2_asset():
    pass


t01_part1_part2_job = define_asset_job(
    name="t01_part1_part2_job",
    selection=["t01_part1_part2_asset", "t01_part1_asset"],
    partitions_def=downstream_partitions_def,
[...]

PR #10910, opened 2022-12-06
Add single dimension -> multidimension partition mapping

clairelin135:
Resolves https://github.com/dagster-io/dagster/issues/10722

Following https://github.com/dagster-io/dagster/pull/10909/files, this PR adds a partition mapping from an upstream single-dimensionally partitioned asset to a downstream multipartitioned asset. In order to use this class, the upstream partitions definition must be a dimension of the downstream asset.

The new `PartitionMapping` object also returns partition subsets instead of ranges, enabling use cases where a downstream subset is not a valid range e.g.:
```
MultiPartitionsDefinition(
)
```
has partition keys: ['1|a', '1|b', '2|a', '2|b']
If the upstream asset had partitions definition StaticPartitionsDefinition(["a", "b"]), the downstream partition keys subset should be {'1|a', '2|a'}, which can't be expressed as a range.


python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -193,6 +197,145 @@ def get_downstream_partitions_for_partition_range(
         raise NotImplementedError()
 
 
+@experimental
+class SingleDimensionToMultiPartitionMapping(PartitionMapping):
```
sryza, line 201: From the name, I wasn't able to tell which direction this goes - i.e. whether the upstream or the downstream partitions definition is expected to be multi-. Is there a way to make that more clear, or at least conform to a convention?

FWIW `AllPartitionMapping` and `LastPartitionMapping`, prefer to what partitions in the upstream asset correspond to a partition of the downstream asset.

clairelin135: That's a good point. Renaming this to `SingleDimensionDependencyMapping`, hopefully that is clearer.

Tossing out some possible naming ideas:
- SingleDimensionDependencyMapping
- MatchesDownstreamPartitionDimensionMapping
- PartitionDimensionMapping

clairelin135:
@sryza per your feedback I've renamed the class and switched to the new partition subset methods, raising `NotImplementedError`s for the range methods. I think this is ready for you to take another look!

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -215,6 +219,107 @@ def get_downstream_partitions_for_partition_range(
...
+        matching_dimensions = [
+            partitions_def
+            for partitions_def in downstream_partitions_def.partitions_defs
+            if partitions_def.name == self.partition_dimension_name
+        ]
+        if len(matching_dimensions) != 1:
+            check.failed(f"Partition dimension '{self.partition_dimension_name}' not found")
+        matching_dimension_def = next(iter(matching_dimensions))
+
+        if upstream_partitions_def != matching_dimension_def.partitions_def:
```
sryza, line 254: Something to keep in mind for future optimization is that this `!=` could be expensive for large `StaticPartitionsDefinition`s.

sryza (approved):
Nice

Issue #10453, opened 2022-11-09
Add more kind tags


braunjj:
[Jamie's PR](https://github.com/dagster-io/dagster/pull/10355) added a few new special kind tags to the asset graph last week and I'd like to support a few more + fix the spacing in the dbt tag svg

**SVGs are attached**:
[kind_tag_icons2.zip](https://github.com/dagster-io/dagster/files/9975835/kind_tag_icons2.zip)

**New icons we'll support**:
- duckdb
- pyspark
- pandas
- tensorflow
- dbt (replace the current one)
- pytorch
- sql
- googlesheets
- slack

PR #10931, opened 2022-12-06
[dagit] Updated asset DAG styles, added additional compute tags

bengotow:
### Summary & Motivation

This PR updates the Asset DAG UI to match this linear task: https://linear.app/elementl/issue/DAGIT-227/revise-asset-graph-node-statuses

I also addressed https://github.com/dagster-io/dagster/issues/10453 by adding the additional partner logos to the OpTags set, and updated a few elements of the Asset graph sidebar to match.

I tested this PR by creating a set of asset groups with assets in each of the available states and verified that the appearance matches the states described in the mockup:

- Loading
- Last Run Failed
- Freshness Policy + Late
- Stale
- Never Materialized
- Freshness Policy + On Time
- Materialized

I also zoomed out and verified that AssetNodeMinimal, which is unfortunately a slightly separate code path, uses the same colors.

image

image

image


braunjj (approved):
Looks great. Thank you!

PR #10934, opened 2022-12-06
[dagit] Fix “Job In” label regression in Chrome v109

bengotow:
### Summary & Motivation

This PR is a small patch to resolve an issue reported in Slack here:
https://elementl-workspace.slack.com/archives/C03CCE471E0/p1670355704131579

This rendering issue only appears in Chrome 109 and only impacts the "Job in repo", "Asset in repo" tags.  I added a test case to our MiddleTruncation storybook to verify the fix and make it easier to spot this in the future.

I believe the reason this was happening was because MiddleTruncation renders its text twice, and the default flex "align-items" value is "baseline", which aligns the children based on the location of the first text baseline inside each of them. Putting the measurement div below the rendered div fixes the issue for me.

Before:
image

After: 
image

### DIFF
--- a/js_modules/dagit/packages/ui/src/components/MiddleTruncate.stories.tsx
+++ b/js_modules/dagit/packages/ui/src/components/MiddleTruncate.stories.tsx
@@ -2,11 +2,13 @@ import {Meta} from '@storybook/react/types-6-0';
 import faker from 'faker';
 import * as React from 'react';
 
+import {BaseTag} from './BaseTag';
 import {Box} from './Box';
 import {Colors} from './Colors';
 import {Icon} from './Icon';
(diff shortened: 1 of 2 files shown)


js_modules/dagit/packages/ui/src/components/MiddleTruncate.stories.tsx
```diff
@@ -2,11 +2,13 @@ import {Meta} from '@storybook/react/types-6-0';
 import faker from 'faker';
 import * as React from 'react';
 
+import {BaseTag} from './BaseTag';
```
hellendag, line 5:
```suggestion
```

hellendag (approved):

PR #10936, opened 2022-12-06
[code-location-selector-stack] Fixing typing errors in in test_run_status_sensors (1/N)

schrockn:
Summary & Motivation: I need to add tests in this file and it is full of typing errors. Doing this in small chunks.

This imports check and applies not none to this quite involved access to repository location

Test Plan: BK

### DIFF
--- a/python_modules/dagster/dagster_tests/daemon_sensor_tests/test_run_status_sensors.py
+++ b/python_modules/dagster/dagster_tests/daemon_sensor_tests/test_run_status_sensors.py
@@ -6,6 +6,7 @@ import pendulum
 import pytest
 
 from dagster import DagsterRunStatus
+from dagster import _check as check
 from dagster._core.events import DagsterEvent, DagsterEventType
 from dagster._core.events.log import EventLogEntry
 from dagster._core.scheduler.instigation import TickStatus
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster_tests/daemon_sensor_tests/test_run_status_sensors.py
```diff
@@ -42,31 +43,37 @@ def instance_with_sensors(overrides=None, attribute="the_repo"):
...
+                repository_location.get_repository(attribute),
             )
 
 
 @contextmanager
 def instance_with_multiple_repos_with_sensors(overrides=None):
     with instance_for_test(overrides) as instance:
         with create_test_daemon_workspace_context(
-            workspace_load_target(None), instance=instance
+            workspace_load_target(), instance=instance
```
jamiedemaria, line 64: The `None` passed to `workspace_load_target` is required here. If no value is provided, the `workspace_load_target` tries to load a repo with the name `the_repo`. Since we're trying to load multiple repos here, we can't have a specific repo name specified. 

There's a couple of tests failing in BK right now because of it - test_cross_repo_run_status_sensor, test_instance_run_status_sensor , test_cross_repo_job_run_status_sensor,

jamiedemaria (commented):
looks good to me other than the one comment

PR #10949, opened 2022-12-07
[docs] - add a guide for scheduling assets

slopp:
### Summary & Motivation

Declarative scheduling is a new concept that is very powerful. This tutorial guides the reader through a series of code samples that build up to a fully declarative schedule using freshness policies, helping them gain intuition of how this new approach compares to traditional cron scheduling.

- code samples run
- tutorial renders
- links work


docs/content/_navigation.json
```diff
@@ -50,6 +50,10 @@
           {
             "title": "Testing Assets",
             "path": "/tutorial/assets/testing-assets"
+          },
+          {
+            "title": "Scheduling Assets",
+            "path": "/tutorial/assets/scheduling-assets"
```
yuhan, line 56: the other tutorials are more sequential, i.e. they are step-by-step instructions that build on top of the previous page, while this one feels like a guide. should we move this tutorial to guides here https://docs.dagster.io/guides/general, and add a link to the guide here: https://dagster-git-docs-lopp-freshness-tutorial-elementl.vercel.app/tutorial/assets/testing-assets#conclusion?

yuhan: when it's a guide, it's ok to use arbitrary a, b, c as asset names. if we decide to keep this as the last step in the tutorial, naming assets something more meaningful would make more sense.

erinkcochran87 (changes requested):
Back to you! Left comments on the Vercel preview. LMK if you'd like any help with this!

slopp:
@erinkcochran87  thanks for the thoughtful review and fixing all my embarrassing grammar mistakes. I believe I addressed all the comments. I think the only remaining items are re: images

I double checked the majority of the images have the correct asset highlighted so I don't think further highlighting is critical

You mentioned one screenshot that could use some re-formatting though, would you be able to assist with that one?

erinkcochran87 (approved):

PR #10950, opened 2022-12-07
[dagit] Use regular name@location for repo tag on runs

hellendag:
### Summary & Motivation

In cases where we are using the RepoAddress values to query for a run tag, we should continue using `repo@location`. This is not expected to be URI-encoded.

This should fix an issue where launching a run does not include the dunder repo in the created tag.

Run dagit with `-f` for a file that will use a dunder repo name. Launch a run. Verify that I see the (failed) run in the "Runs" tab on that job within Dagit, indicating that it can be queried successfully by tag.


hellendag:
* master
  * **PR #10950** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10950" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10950?utm_source=stack-comment).

PR #10967, opened 2022-12-08
Fix bug with re-execution snapshot ids

OwenKephart:
### Summary & Motivation

When you subset an asset job (i.e. most of the time), a new snapshot id is generated which will not match the snapshot id of the job found in the repository, which is what we check against when deciding if we'll show the warning message or not.

```
from dagster import asset


@asset
def a():
    return 1


@asset
def b(a):
    return a

```

Before these changes, the exclamation point would show up on the page for the run next to the Re-execute button, now it doesn't.

I have no idea of any better way of doing this.

### DIFF
--- a/js_modules/dagit/packages/core/src/workspace/useRepositoryForRun.ts
+++ b/js_modules/dagit/packages/core/src/workspace/useRepositoryForRun.ts
@@ -48,6 +48,9 @@ export const useRepositoryForRun = (
     }
 
     const pipelineName = run.pipelineName;
+    // When jobs are subsetted (with an opSelection or assetSelection), only their
+    // parentPipelineSnapshotId (the id of the pipelineSnapshot that they were subsetted from) will
+    // be found in the repository, so look for that instead.
     const snapshotId = run.parentPipelineSnapshotId ?? run.pipelineSnapshotId;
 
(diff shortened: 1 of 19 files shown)


js_modules/dagit/packages/core/src/workspace/useRepositoryForRun.ts
```diff
@@ -48,7 +48,7 @@ export const useRepositoryForRun = (
     }
 
     const pipelineName = run.pipelineName;
-    const snapshotId = run.pipelineSnapshotId;
+    const snapshotId = run.parentPipelineSnapshotId ?? run.pipelineSnapshotId;
```
hellendag, line 51: Is it always safe just to check one snapshot ID or the other, or is it possible we may want to check both?

OwenKephart: My understanding is that the pipelineSnapshotId found in the repository (i.e. what we're checking against) will always be the parentPipelineSnapshotId if there is a parent PipelineSnapshot. If there's no parent, that means the job wasn't subsetted, and so we'll compare against the regular pipelineSnapshotId

So I think this should be ok

alangenfeld (approved):
nice

PR #10969, opened 2022-12-08
add NullMetadataValue

smackesey:
### Summary & Motivation

This adds a new `NullMetadataValue`, which is a need that has come up from time to time.

Some of the existing scalar metadata types (float, int, bool) are treated as nullable inconsistently in our system-- for example, `MetadataValue.bool` is not supposed to accept `None` (type error) but in GQL dagit it is nullable. Then if a null value actually showed up in dagit it was just blank (see below).

I changed the representation to a string `null` and added an explicit `NullMetadataValue` type that is returned by `normalize_metadata_value` for `None`.

BK, looking at rendering in dagit.

Before: 
image

After:

image


smackesey:
* master
  * **PR #10969** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10969" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10969?utm_source=stack-comment).

sryza (commented):
This makes sense to me. Will defer to @bengotow on approval, because most of the changes are on the frontend.

js_modules/dagit/packages/core/src/metadata/MetadataEntry.tsx
```diff
@@ -158,7 +158,9 @@ export const MetadataEntry: React.FC<{
     case 'IntMetadataEntry':
       return <>{entry.intValue !== null ? entry.intValue : entry.intRepr}</>;
     case 'BoolMetadataEntry':
-      return entry.boolValue !== null ? <>{entry.boolValue.toString()}</> : null;
+      return <>{entry.boolValue !== null ? entry.boolValue.toString() : 'null'}</>;
```
bengotow, line 161: Just want to call out this line in case it was an earlier change / idea - it looks like the BoolMetadataEntry can also now be null? (looks like the previous version actually rendered nothing and not the string null)

bengotow (approved):
This looks great to me! Just left one inline comment about the BoolMetadataValue tweak 👍

js_modules/dagit/packages/core/src/metadata/MetadataEntry.tsx
smackesey: Pretty sure this should never actually become null, a null for bool metadata value will get caught on the backend. But leaving this in place just in case-- and IMO we should render null the same everywhere.

PR #10980, opened 2022-12-08
unexperimentalize PartitionMapping

sryza:
### Summary & Motivation

Writing custom PartitionMappings is deprecated, but I don't think we should warn anymore when someone uses one of the builtin PartitionMappings

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/partition_mapping.py
+++ b/python_modules/dagster/dagster/_core/definitions/partition_mapping.py
@@ -2,16 +2,18 @@ from abc import ABC, abstractmethod
 from typing import NamedTuple, Optional
 
 import dagster._check as check
-from dagster._annotations import experimental, public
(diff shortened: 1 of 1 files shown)


clairelin135:
In the next major release we should ideally deprecate the partition range methods and replace them with partition subset methods. Is it worth waiting until then to un-experimentalize partition mappings?

sryza:
@clairelin135 - because PartitionMappings are experimental, I don't think we need to wait until the next major release to call the partition range methods deprecated. I think we can remove them in the next major release. Implementing your own PartitionMapping is already deprecated, as of a change I merged last week. I can also add some comments calling out that making your own PartitionMapping is not stable.

clairelin135 (approved):
@sryza nice, I don't have any other concerns since we've deprecated implementing your own partition mapping 👍

PR #10982, opened 2022-12-08
sort task definition config secrets and env vars to ensure consistent ordering

gibsondan:
Summary:
The ordering of secrest and env vars that comes out of describe_task_definition doesn't always match what is passed into register_task_definition, so add a sort when comparing them.


python_modules/libraries/dagster-aws/dagster_aws/ecs/tasks.py
```diff
@@ -54,8 +54,8 @@ def __new__(
             check.str_param(container_name, "container_name"),
             check.opt_sequence_param(command, "command"),
             check.opt_mapping_param(log_configuration, "log_configuration"),
-            check.opt_sequence_param(secrets, "secrets"),
-            check.opt_sequence_param(environment, "environment"),
+            sorted(check.opt_sequence_param(secrets, "secrets"), key=lambda s: s["name"]),
+            sorted(check.opt_sequence_param(environment, "environment"), key=lambda e: e["name"]),
```
alangenfeld, line 58: do we maintain ordering of duplicate names? for cases where merging values from lets say the root task and container_context yield conflicts? I think `sorted` is good about this but maybe warrants a test case

alangenfeld (approved):

python_modules/libraries/dagster-aws/dagster_aws/ecs/tasks.py
gibsondan: This would be verifying that the sorted function is stable, yeah - which it is, will double-check

PR #10983, opened 2022-12-08
Adopt pyright for typechecking

smackesey:
### Summary & Motivation

- Drop mypy configuration and BK steps
- Add [Pyright](https://github.com/microsoft/pyright) configuration and a single BK step that checks the whole repo. BK step uses `make pyright`, which can also be run locally.

Benefits. Pyright:

- Is 4-5x faster than mypy (it runs on NodeJS and not Python)
- Is deeply integrated with Pylance, the state-of-the-art Python language server for VSCode. This allows rapid incremental type-checking during development, which makes it easy to achieve parity between editor diagnostics and CI.
- Is much better maintained than mypy. Bugs are fixed fast.
- Has a more powerful type-checking engine than mypy, allowing more expressive types to be used.
- Can be used to assess type-completeness of our public API in addition to type-correctness of annotated code.

Limitations:

- There are a few cases where third-party libraries use mypy-specific extensions to make non-standard APIs legible to mypy. Most of these do not yet offer similar "adapter" functionality for pyright. However, as Python's typing capabilities grow, the need for third-party extensions is diminishing.

BK

### DIFF
--- a/.buildkite/dagster-buildkite/dagster_buildkite/package_spec.py
+++ b/.buildkite/dagster-buildkite/dagster_buildkite/package_spec.py
@@ -12,7 +12,13 @@ from dagster_buildkite.python_packages import PythonPackages, changed_filetypes
 from .python_version import AvailablePythonVersion
 from .step_builder import BuildkiteQueue
 from .steps.tox import build_tox_step
-from .utils import BuildkiteLeafStep, GroupStep, is_feature_branch
+from .utils import (
+    BuildkiteLeafStep,
+    BuildkiteTopLevelStep,
+    GroupStep,
+    is_command_step,
+    is_feature_branch,
+)
 
 _CORE_PACKAGES = [
     "python_modules/dagster",
@@ -26,7 +32,6 @@ _INFRASTRUCTURE_PACKAGES = [
     "python_modules/automation",
     "python_modules/dagster-test",
     "python_modules/docs/dagit-screenshot",
-    "scripts",
 ]
 
 
@@ -98,8 +103,6 @@ class PackageSpec:
         timeout_in_minutes (int, optional): Fail after this many minutes.
         queue (BuildkiteQueue, optional): Schedule steps to this queue.
         run_pytest (bool, optional): Whether to run pytest. Enabled by default.
-        run_mypy (bool, optional): Whether to run mypy. Runs in the highest available supported
-            Python version. Enabled by default.
     """
 
     directory: str
@@ -115,7 +118,6 @@ class PackageSpec:
     timeout_in_minutes: Optional[int] = None
     queue: Optional[BuildkiteQueue] = None
     run_pytest: bool = True
-    run_mypy: bool = True
 
     def __post_init__(self):
         if not self.name:
@@ -127,7 +129,7 @@ class PackageSpec:
         self._should_skip = None
         self._skip_reason = None
 
-    def build_steps(self) -> List[GroupStep]:
+    def build_steps(self) -> List[BuildkiteTopLevelStep]:
         base_name = self.name or os.path.basename(self.directory)
         steps: List[BuildkiteLeafStep] = []
 
@@ -190,26 +192,22 @@ class PackageSpec:
                         )
                     )
 
-        if self.run_mypy:
-            steps.append(
-                build_tox_step(
-                    self.directory,
-                    "mypy",
-                    base_label=base_name,
-                    command_type="mypy",
-                    python_version=supported_python_versions[-1],
-                    skip_reason=self.skip_reason,
-                )
-            )
-
         emoji = _PACKAGE_TYPE_TO_EMOJI_MAP[self.package_type]  # type: ignore[index]
-        return [
-            GroupStep(
-                group=f"{emoji} {base_name}",
-                key=base_name,
-                steps=steps,
-            )
-        ]
+        if len(steps) >= 2:
+            return [
+                GroupStep(
+                    group=f"{emoji} {base_name}",
+                    key=base_name,
+                    steps=steps,
+                )
+            ]
+        elif len(steps) == 1:
+            only_step = steps[0]
+            if not is_command_step(only_step):
+                raise ValueError("Expected only step to be a CommandStep")
+            return [only_step]
+        else:
+            return []
 
     @property
     def requirements(self):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,42 +22,73 @@ required-version = "22.12.0"
 target-version = ['py37', 'py38', 'py39', 'py310', 'py311']
 
 # ########################
-# ##### MYPY
+# ##### PYRIGHT
 # ########################
 
 # [Docs root]
-#   https://mypy.readthedocs.io/en/stable/index.html
+#   https://github.com/microsoft/pyright/tree/main/docs
 # [Config option reference]
-#   https://mypy.readthedocs.io/en/stable/config_file.html#the-mypy-configuration-file
+#   https://github.com/microsoft/pyright/blob/main/docs/configuration.md
 
-[tool.mypy]
+# Pyright does not have a docs site, but the documentation (a collection of
+# markdown files in the GH repo, linked above) is pretty thorough.
 
-# Allow variables to be redefined with arbitrary types even if they already are typed.
-allow_redefinition = true
+[tool.pyright]
 
-# Allow for use of type annotations in local variables inside function bodies even when the
-# signature is not annotated.
-disable_error_code = ['annotation-unchecked']
+include = [
+  ".buildkite/dagster-buildkite",
+  "python_modules",
+  "examples",
+  "integration_tests",
+  "scripts",
+]
 
-# Ignore cases where mypy cannot resolve *the types* for a given import. Note that this is different
-# than not being able to resolve the import at runtime. A module can be installed, but mypy will
-# consider it unresolveable if it does not either (a) have a `py.typed` marker file; (b) have
-# a corresponding stub package (i.e. named `types-XXX`) available to mypy. Because Dagster uses
-# has several dependencies that do not satisfy either criterion, mypy will by default emit errors
-# when it encounters imports of those dependencies. Since there's no way to fix these, we silence
-# them.
-ignore_missing_imports = true
+# Unfortunately pyright does not offer a way to extend the default exclusions, so we have to
+# reiterate them here if we want to add anything else.
+exclude = [
+  "**/node_modules",
+  "**/__pycache__",
+  "**/__generated__",
+  "**/vendor",
+  "**/.tox",
+  ".git",
+  "**/.venv*",
+]
 
-# If a type stub package is missing from the environment but available on typeshed, automatically
-# install it. Note that without `--non-interactive` this will prompt the user.
-install_types = true
+# These two settings point pyright to a python environment to resolve imports against. This virtual
+# environment is defined in the `pyright` tox environment in the tox section below-- that
+# environment must be built before pyright can run correctly.
+venv = ".venv"
+venvPath = "pyright/envs/master"
 
-# Allow PEP-420-style namespace packages. Without this enabled, different parts of a namespace
-# package will trigger "Duplicate module" errors.
-namespace_packages = true
+# Set to false to help us during the transition from mypy to pyright. Mypy does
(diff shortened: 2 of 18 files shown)


smackesey:
* master
      * **PR #10907** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10907" target="_blank">Graphite</a> 
        * **PR #10983** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10983" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10983?utm_source=stack-comment).

schrockn:
I'm extremely excited to make this switch. but the approach here is simply too high risk and this PR is too large to be reviewable. We *have* to break this up into reviewable subcomponents. The only PRs that should be this large are purely mechanical (e.g. a rename of a commonly used utility function) to the point where they could be applied and replicated as a script.

This sequence of changes needs to end up in a state where, for example, your dagster-graphql changes are in 1 PR, the pendulum changes in another, and so forth.

We're chatting in Slack but wanted to make sure spectators know where this is at.

docs/dagit-screenshot/tox.ini
```diff
@@ -1,15 +0,0 @@
-[tox]
```
schrockn, line 1: is this deletion deliberate?

schrockn: this is only for mypy?

python_modules/dagster/dagster/_core/execution/job_backfill.py
```diff
@@ -280,7 +280,7 @@ def create_backfill_run(
             use_parent_run_tags=False,  # don't inherit tags from the previous run
         )
 
-    elif backfill_job.reexecution_steps:
+    else:  # backfill_job.reexecution_steps
```
schrockn, line 283: this looks very problematic

python_modules/dagster/dagster/_core/host_representation/repository_location.py
```diff
@@ -449,7 +449,7 @@ def get_external_schedule_execution_data(
         instance: DagsterInstance,
         repository_handle: RepositoryHandle,
         schedule_name: str,
-        scheduled_execution_time,
+        scheduled_execution_time: datetime.datetime,  # actually a pendulum datetime
```
schrockn, line 452: isn't it better to leave this untyped? why do this if you have the type: ignore below?

python_modules/dagster/dagster/_core/telemetry.py
```diff
@@ -571,11 +571,11 @@ def log_dagster_event(event: DagsterEvent, pipeline_context: PlanOrchestrationCo
 
     if event.is_step_start:
         action = STEP_START_EVENT
-    if event.is_step_success:
+    elif event.is_step_success:
         action = STEP_SUCCESS_EVENT
         if isinstance(event.event_specific_data, StepSuccessData):  # make mypy happy
             metadata["duration_ms"] = event.event_specific_data.duration_ms
-    if event.is_step_failure:
+    else:  # event.is_step_failure
```
schrockn, line 578: looks problematic

python_modules/dagster/dagster/_grpc/impl.py
```diff
@@ -265,14 +265,14 @@ def get_external_schedule_execution(
...
-    scheduled_execution_timezone: str,
+    scheduled_execution_timestamp: Optional[float],
+    scheduled_execution_timezone: Optional[str],
 ):
     schedule_def = repo_def.get_schedule_def(schedule_name)
     scheduled_execution_time = (
         pendulum.from_timestamp(
             scheduled_execution_timestamp,
-            tz=scheduled_execution_timezone,
+            tz=check.not_none(scheduled_execution_timezone),
```
schrockn, line 275: this looks scary. would prefer type: ignore for conservatism sake

scripts/check_schemas.py
```diff
@@ -53,9 +53,9 @@ def validate_column(column: Column):
         and str(column.server_default.arg) == "CURRENT_TIMESTAMP"
     ):
         raise Exception(
-            f"Column {column} has a server default of CURRENT_TIMESTAMP without precision specified. "
-            "To allow schema compatibility between MySQL, Postgres, and SQLite, "
-            "use dagster._core.storage.sql.py::get_current_timestamp() instead."
+            f"Column {column} has a server default of CURRENT_TIMESTAMP without precision"
```
schrockn, line 56: weird change to include but don't care

tox.ini
```diff
@@ -0,0 +1,108 @@
+[tox]
```
schrockn, line 1: This seems like a big change to our CI. Can you add a big comment block at the top explaining this?

schrockn (changes requested):
This is very epic. I have a few comments that I'd like address, but this is essentially gtg.

tox.ini
smackesey: I changed the approach, no longer have a root-level tox

docs/dagit-screenshot/tox.ini
smackesey: Yes only `mypy` was run on it. It's an internal utility (glorified script) for taking dagit screenshots that doesn't have any unit tests.

scripts/run-pyright.py
```diff
@@ -0,0 +1,302 @@
...
+    exclude_path = get_env_path(env, "exclude.txt")
+    config["include"] = load_path_file(include_path)
+    if os.path.exists(exclude_path):
+        config["exclude"] += load_path_file(exclude_path)
+    temp_config_path = f"pyrightconfig-{env}.json"
+    print("Creating temporary pyright config file at", temp_config_path)
+    with open(temp_config_path, "w", encoding="utf-8") as f:
+        json.dump(config, f)
+    yield temp_config_path
+    os.remove(temp_config_path)
```
schrockn, line 242: seems like this should be in a finally block

pyproject.toml
```diff
@@ -22,42 +22,73 @@ required-version = "22.12.0"
...
+useLibraryCodeForTypes = false
 
-# Print codes (e.g. "[arg-type]") for each error in output.
-show_error_codes = true
+# We use ruff for this.
+reportInvalidStringEscapeSequence = false
+
+# As of 2023-02-02, there are still many `py.typed` libs that are not compliant with the standards
+# for defining a public API.
+reportPrivateImportUsage = false
```
schrockn, line 87: wait so now we are turning this off?

pyright/alt-1/requirements.txt
```diff
@@ -0,0 +1,104 @@
...
+#
+# This file defines the dependencies for an alternative pyright environment. The packages listed in
+# this file will not be present in the environment that pyright/pylance will use by default when
+# running as a language server.
+#
+# NOTE: It is important that dagster-* dependencies of any package listed here are also listed.
+# e.g. if examples/foo depends on python_modules/libraries/dagster-foo, then dagster-foo must be
+# listed if examples/foo is listed.
+#
+# NOTE: Paths are relative to repo root.
```
schrockn, line 11: what is the plan for keeping this file up to date?

pyright/master/requirements.txt
```diff
@@ -0,0 +1,100 @@
+### Requirements: master
+#
+# This file defines the dependencies for the master pyright environment for the internal repo. This
+# is the environment that # pyright/pylance will use by default when running as a language server.
+#
+# Ideally, this would be the only pyright environment we need to define. This is not possible
+# because of dependency conflicts in between Dagster packages/examples. For ease of maintenance,
+# this file should maintain a complete list of Dagster packages, with packages excluded from the
+# enviroment commented out.
```
schrockn, line 9: Do we have a running list of known dependency conflicts? I'm sure that is in your head right now, and it would be useful to document.

pyright/alt-1/requirements.txt
```diff
@@ -0,0 +1,104 @@
...
+#     -e python_modules/libraries/dagster-k8s/
+#     -e python_modules/libraries/dagster-postgres/
+
+### examples/quickstart_gcp
+# -e examples/quickstart_gcp
+#   -e python_modules/dagster[pyright,test]
+#   -e python_modules/dagit
+#   -e python_modules/dagster-graphql
+#   -e python_modules/libraries/dagster-gcp
+#   -e ../internal/dagster-cloud/python_modules/dagster-cloud
```
schrockn, line 80: no references to internal in the OSS repo pleaase

schrockn (changes requested):
req'ing changes based on the reference to internal.

I'm definitely a bit concerned about the maintainability of the pyright merging logic over time.

Does that provide any value beyond the reduction in CI jobs?

pyproject.toml
smackesey: In that DM I said we'd turn it off (but that handles only _some_ cases)-- I edited it maybe an hour or so after first sent:

>Given this, the best we can do is turn off reportPrivateImportUsage, which will affect pendulum and datadog but not e.g. sqlalchemy and several other libs that required import changes. (edited)

pyright/master/requirements.txt
smackesey: I could figure it out with some digging but it's surprisingly hard to figure out what the actual conflicts are. The way I arrived at the current partition is by repeatedly `pip install --dry-run` on our whole package list and eliminating until I could get a successful install.

smackesey:
> I'm definitely a bit concerned about the maintainability of the pyright merging logic over time.

I can see why you might be worried, but I'm not particularly concerned-- it's just some JSON massaging. The default pyright output is terrible on a lot of files and JSON output of `pyright` is not an afterthought-- the maintainer is always telling anyone who doesn't like it to "use JSON output and write a script".

> Does that provide any value beyond the reduction in CI jobs?

The advantage is that it lets you run one command and see combined output from all pyright runs. Without some kind of similar merge post-processing, you'd have to scroll up a bunch in the terminal to see the output from successive runs.

pyright/alt-1/requirements.txt
smackesey: Yeah this was there as a placeholder I intend to remove before merge (they're gone now)-- I was looking for a solution for `quickstart-*` examples, since they depend on `dagster-cloud`.

But having looked at the code of these, they don't actually import from `dagster_cloud` (or most of their deps), so it looks like they can successfully be type-checked in the master env without needing to install their deps.

pyright/alt-1/requirements.txt
smackesey: There is no special plan-- I think it can more-or-less be treated like any one of our gazillion toxfiles, which also have deps that have to be kept in sync with corresponding setup.py files.

I would love it if we could get everything in one env or there were some kind of more automated system for partitioning our packages into envs, but I'm not aware of a better option.

scripts/run-pyright.py
smackesey: you're right

pyproject.toml
schrockn: Ah got thanks for clarification

pyright/master/requirements.txt
schrockn: Got it. Not asking you to dig just to document what is resident in your brain cache.

schrockn (approved):
Ok let's do this then.

I'd like @jmsanders or @alangenfeld  to give final blessing on the buildkite changes before you merge this sucker.

An exciting step forward for our codebase! Thank you for doing this I know it was an extraordinary amount of work.

.buildkite/dagster-buildkite/dagster_buildkite/package_spec.py
```diff
@@ -190,26 +192,22 @@ def build_steps(self) -> List[GroupStep]:
...
-        ]
+        if len(steps) >= 2:
+            return [
+                GroupStep(
+                    group=f"{emoji} {base_name}",
+                    key=base_name,
+                    steps=steps,
+                )
+            ]
+        elif len(steps) == 1:
```
jmsanders, line 204: What are we trying to guard against here?

jmsanders (approved):

.buildkite/dagster-buildkite/dagster_buildkite/package_spec.py
smackesey: We return a `CommandStep` instead of a `GroupStep` if there's only one step, to avoid having to unnecessarily click through a group layer in BK.

PR #10991, opened 2022-12-09
load asset value config

### LINKED ISSUE #10990: load_asset_value should accept config
### Dagster version

master

### What's the issue?

`load_asset_value` does not accept configuration, however it can be the case that an IOManager requires configuration

_No response_

### How to reproduce?

_No response_

### Deployment type

None

### Deployment details

_No response_

### Additional information

_No response_

Nintorac:
### Summary & Motivation

Allowed supply of configuration to `load_asset_value` and propagated information to where it's needed. 

fixes #10990

not as yet

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/repository_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/repository_definition.py
@@ -1423,6 +1423,7 @@ class RepositoryDefinition:
         python_type: Optional[Type] = None,
         instance: Optional[DagsterInstance] = None,
         partition_key: Optional[str] = None,
+        config: Optional[Any] = None
     ) -> object:
         """
         Load the contents of an asset as a Python object.
@@ -1446,7 +1447,7 @@ class RepositoryDefinition:
 
         with AssetValueLoader(self._assets_defs_by_key, instance=instance) as loader:
             return loader.load_asset_value(
-                asset_key, python_type=python_type, partition_key=partition_key
+                asset_key, python_type=python_type, partition_key=partition_key, config=config
             )
 
     @public
--- a/python_modules/dagster/dagster/_core/definitions/repository_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/repository_definition.py
@@ -1423,7 +1423,7 @@ class RepositoryDefinition:
         python_type: Optional[Type] = None,
         instance: Optional[DagsterInstance] = None,
         partition_key: Optional[str] = None,
-        config: Optional[Any] = None
+        resource_config: Optional[Any] = None,
     ) -> object:
         """
         Load the contents of an asset as a Python object.
@@ -1439,6 +1439,8 @@ class RepositoryDefinition:
             python_type (Optional[Type]): The python type to load the asset as. This is what will
                 be returned inside `load_input` by `context.dagster_type.typing_type`.
             partition_key (Optional[str]): The partition of the asset to load.
+            resource_config (Optional[Any]): A dictionary of resource configurations to be passed
+                to the :py:class:`IOManager`.
 
         Returns:
             The contents of an asset as a Python object.
@@ -1447,7 +1449,10 @@ class RepositoryDefinition:
 
         with AssetValueLoader(self._assets_defs_by_key, instance=instance) as loader:
             return loader.load_asset_value(
-                asset_key, python_type=python_type, partition_key=partition_key, config=config
+                asset_key,
+                python_type=python_type,
+                partition_key=partition_key,
+                resource_config=resource_config,
             )
 
     @public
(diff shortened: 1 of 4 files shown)


python_modules/dagster/dagster/_core/definitions/repository_definition.py
```diff
@@ -1403,6 +1403,7 @@ def load_asset_value(
         python_type: Optional[Type] = None,
         instance: Optional[DagsterInstance] = None,
         partition_key: Optional[str] = None,
+        config: Optional[Any] = None
```
dpeng817, line 1406: would want to see an entry in the doc string

python_modules/dagster/dagster/_core/storage/asset_value_loader.py
```diff
@@ -89,13 +91,15 @@ def load_asset_value(
         io_manager_key = assets_def.get_io_manager_key_for_asset_key(asset_key)
         io_manager_def = resource_defs[io_manager_key]
         required_resource_keys = io_manager_def.required_resource_keys | {io_manager_key}
-
+        config = config or {}
+        resource_config = config.get('resources', {})
```
dpeng817, line 95: it seems like the only thing we're pulling off of the config here is the resources config. I wonder if we would be better served by just making the parameter called `resources_config` directly, and allowing you to configure each resource by key without having to specify the top level `resources` key to the config dictionary.

dpeng817 (changes requested):
Some things to consider. Definitely also going to want to add tests; the common case where the io manager needs input config, but then also the case where a resource required by the IO manager needs config.

python_modules/dagster/dagster/_core/storage/asset_value_loader.py
Nintorac: I am passing the config into `build_input_context`, is it not required there?

Nintorac: As well, I tend to pass the same config around with config for the whole project, so it's a little less boiler plate and simplifies things a bit

dpeng817: The config argument on `build_input_context` is for input config specifically, so the config argument provided here wouldn't match up. `build_input_context` also has a `resource_config` argument (should probably be called resources_config) that allows for providing config for the resources.

Regarding passing the same config around, that's definitely reasonable and I imagine many users do the same thing, but it seems like the difference is just between passing `run_config` and `run_config.get("resources")`, no? In which case, I'm inclined to scope it down to just resources_config to keep the data model simpler.

Nintorac: OK sounds good, I have updated accordingly

Nintorac:
I think this is ready now, updates include:
- added a way to pass resource_config into asset value loader
- added tests on asset_value_loader for iomanagers with config
- added tests on asset_value_loader for iomanagers with resources with config
- added tests on repository load_asset_value for iomanagers with config
- added tests on repository load_asset_value for iomanagers with resources with config

python_modules/dagster/dagster/_core/storage/asset_value_loader.py
```diff
@@ -75,11 +81,14 @@ def load_asset_value(
             python_type (Optional[Type]): The python type to load the asset as. This is what will
                 be returned inside `load_input` by `context.dagster_type.typing_type`.
             partition_key (Optional[str]): The partition of the asset to load.
+            resource_config (Optional[Any]): A dictionary of resource configurations to be passed
+                to the :py:class:`IOManager`.
```
dpeng817, line 85: nit: would be more accurate to `resource configurations to be passed to resources required by the IOManager`

dpeng817 (commented):
https://buildkite.com/dagster/dagster/builds/43903 - this is looking super close - will follow up with results of this build!

python_modules/dagster/dagster/_core/storage/asset_value_loader.py
sryza: it's to the IOManager too though, right?

dpeng817: that is also true, yea

python_modules/dagster/dagster_tests/storage_tests/test_asset_value_loader.py
```diff
@@ -216,3 +216,66 @@ def repo():
...
+
+    class MyIOManager(IOManager):
+        def handle_output(self, context, obj):
+            assert False
+
+        def load_input(self, context):
+            return context.resources.io_resource
+
+    @io_manager(required_resource_keys={"io_resource"})
+    def my_io_manager(context):
```
dpeng817, line 264:
```suggestion
    def my_io_manager():
```
Linter complains about this

dpeng817 (approved):
Just that one lint error and this will be good to go

Nintorac:
Sorry for the delay -- Ruff looks cool. A fair bit faster!

This should be ready now :)

PR #10995, opened 2022-12-09
Ruff docstrings D212,D214 (MultilineSummaryFirstLine,SectionNotOverIndented)

smackesey:

PR #10994, opened 2022-12-09
Ruff docstrings D210 (NoSurrondingWhitespace)

smackesey:

PR #10992, opened 2022-12-09
Ruff docstrings D202 (NoBlankLineAfterFunction)

smackesey:

PR #10998, opened 2022-12-09
Ruff docstrings D411,D412 (BlankLineBeforeSection,NoBlankLinesBetweenHeaderAndContent)

smackesey:


smackesey:
* master
      * **PR #10907** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10907" target="_blank">Graphite</a> 
        * **PR #10992** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10992" target="_blank">Graphite</a> 
          * **PR #10993** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10993" target="_blank">Graphite</a> 
            * **PR #10994** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10994" target="_blank">Graphite</a> 
              * **PR #10995** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10995" target="_blank">Graphite</a> 
                * **PR #10996** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10996" target="_blank">Graphite</a> 
                  * **PR #10997** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10997" target="_blank">Graphite</a> 
                    * **PR #10998** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10998" target="_blank">Graphite</a>  👈
        * **PR #10983** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/10983" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/10998?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/decorators/repository_decorator.py
```diff
@@ -212,96 +212,95 @@ def repository(
         description (Optional[str]): A string description of the repository.
 
     Example:
+        .. code-block:: python
```
schrockn, line 215: Did ruff do this reformatting or is this you?

schrockn (changes requested):
Before approving I'd like an answer to my question about code indenting in docstring. What is going on there? Does ruff rewrite or is that you fixing stuff?

python_modules/dagster/dagster/_core/definitions/decorators/repository_decorator.py
smackesey: So I actually can't remember for that specific one-- at the time of these changes (there have prob been ~20 ruff releases since) ruff was flagging this and not autofixing.

I just experimented on the newest version and it accepts these:

```
Example:
.. code-block::

Example:
    .. code-block::

Example:

    .. code-block::
```

But not this:

```
Example:

.. code-block::
```

Which it autocorrects to:

```
Example:
.. code-block::
```

So probably we should go with that last form, since it has less indentation.

schrockn:
Why is this rule good then? Could you include in summary & motivation?

smackesey:
> Why is this rule good then? Could you include in summary & motivation?

In general I did not choose any of the docstring rules-- I took the preset config Charlie provided for Google-style docstrings (enable D category with a set of specific ignores), since that is what we are generally use throughout the codebase and the docstrings need to consistently meet this standard for tools that are expecting them in that format (e.g. certain sphinx plugins).

schrockn:
Does it autocorrect?

What I don't want to add is a rule that is a pure preference that forces engineers to correct things.

If 1) this prevents breakages in other tools (like Sphinx) OR 2) it just does it for people then I support it.

Is 1 of those 2 true?

smackesey:
Ruff now autocorrects this:

```
Example:

.. code-block::
```

To this:

```
Example:
.. code-block::
```

I believe, for the specific example you commented on, that at that time (there have been ~20 releases since) Ruff (incorrectly) flagged this code-block-immediately under example style as an error that could not be autofixed, so I added that indentation to fix the error. In general I only made manual docstring fixes to correct stuff that Ruff was flagging but wouldn't autofix.

>What I don't want to add is a rule that is a pure preference that forces engineers to correct things.

I also do not want to, and don't believe I added any of those-- only the stuff that's supposed to enforce Google-style docstrings. To the extent that any of the rules enabled currently lack autofix but are in principle auto-fixable, I'm sure Charlie will implement autofix (probably very fast).

schrockn:
So as of now this will surface errors that engineers have to fix manually, correct? Until Charlie adds support?

smackesey:
>So as of now this will surface errors that engineers have to fix manually, correct?

I don't think so. I think that _was_ true ~20 versions ago when I created this changeset, but since then the behavior seems to be that ruff will always just auto-remove a blank line between header and content.

schrockn (approved):

PR #11007, opened 2022-12-09
Temporarily disable some Azure test suites

jmsanders:
Something that landed between now and Wednesday broke a lot of Windows tests. Temporarily disabling so broken builds are less noisy but we'll need to look into fixing these before the next release:

https://dev.azure.com/elementl/dagster/_build/results?buildId=17542&view=logs&j=a542e6c6-b199-516e-0d52-86faaec15bd8&t=c277973c-ae99-5b1f-d347-23af2b7d8f82&l=655

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -13,12 +13,12 @@ parameters:
     type: object
     default:
       - api_tests
-      - cli_tests
+      # - cli_tests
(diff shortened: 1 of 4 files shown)


schrockn (changes requested):
Can you do this more selectively with pytest skips?

PR #11013, opened 2022-12-09
Pin tox on windows to < 4

jmsanders:
@alangenfeld has a hunch that the recent tox upgrade might be to blame for the regressions we're skipping in https://github.com/dagster-io/dagster/pull/11007


schrockn (changes requested):
Explain hunch and then explain how we will verify in test plan?

schrockn (approved):
great

alangenfeld:
make sure to update this with the revert of the skipped tests

PR #11019, opened 2022-12-09
[definitions-accessors] Add get_job_def to `Definitions`.

schrockn:
Summary & Motivation: We need to add accessors for testing/etc. This is an opportunity to clean up our interfaces as well.

This also renames get_inner_repository to get_inner_repository_for_loading_process to indicate that that method is supposed
to get used at a particular time.

Test Plan: BK

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/definitions_class.py
+++ b/python_modules/dagster/dagster/_core/definitions/definitions_class.py
@@ -3,6 +3,7 @@ from typing import Any, Dict, Iterable, Mapping, Optional, Union
 import dagster._check as check
 from dagster._annotations import experimental, public
 from dagster._core.execution.with_resources import with_resources
+from dagster._utils.cached_method import cached_method
 
 from .assets import AssetsDefinition, SourceAsset
 from .cacheable_assets import CacheableAssetsDefinition
@@ -95,7 +96,6 @@ class Definitions:
             ]
 
         self._created_pending_or_normal_repo = created_repo
-        self._resolved_repo: Optional[RepositoryDefinition] = None
 
     @public
     def get_job_def(self, name: str) -> JobDefinition:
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -96,6 +96,20 @@ def created_repo():
...
+    @public
+    def get_job_def(self, name: str) -> JobDefinition:
+        """Get a job definition by name. If you passed in a an UnresolvedAssetJobDefinition
+        (return value of define_asset_job) it will be resolved to a JobDefinition when returned
+        from this function."""
+
+        check.str_param(name, "name")
+        return self.get_resolved_repo().get_job(name)
+
+    def get_resolved_repo(self) -> RepositoryDefinition:
```
sryza, line 108: Might we want to cache this?

sryza (approved):

python_modules/dagster/dagster/_core/definitions/definitions_class.py
schrockn: `self._created_repo = self._created_repo.compute_repository_definition()`

Something else than this on line 110?

sryza: Oops, I read this too fast, this looks right. It strikes me as a little weird to overwrite the attribute, but seems fine.

python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -94,10 +94,37 @@ def created_repo():
...
+    @public
+    def get_job_def(self, name: str) -> JobDefinition:
+        """Get a job definition by name. If you passed in a an UnresolvedAssetJobDefinition
+        (return value of define_asset_job) it will be resolved to a JobDefinition when returned
+        from this function."""
+
+        check.str_param(name, "name")
+        return self.get_repository_def().get_job(name)
+
+    def get_repository_def(self) -> RepositoryDefinition:
```
sryza, line 109: if you used `@cached_method`, this could be:
```python
@cached_method
def get_repository_def(self) -> RepositoryDefinition:
    return (
        self._created_pending_or_normal_repo.compute_repository_definition()
        if isinstance(self._created_pending_or_normal_repo, PendingRepositoryDefinition)
        else self._created_pending_or_normal_repo
    )
```

PR #11025, opened 2022-12-09
Use buildkite_deps.txt to declare explicit buildkite deps

jmsanders:
Instead of co-opting the existing meaning of requirements.txt


gibsondan (approved):
this works in internal too?

PR #11031, opened 2022-12-09
Allow setting base_dir as a StringSource in sqlite storage

gibsondan:
Summary:
We say you can do this in the docs but it doesn't work - this makes it work.

There's a gotcha though, we can't set the legacy storage fields if we do this since they derive their values from the base_dir. I think it's OK for back-compat to not work in this scenario (it seems worse than parsing the env var and putting its value in plaintext) but open to other possibilities.

### DIFF
--- a/python_modules/dagster/dagster/_core/instance/ref.py
+++ b/python_modules/dagster/dagster/_core/instance/ref.py
@@ -56,7 +56,13 @@ def configurable_secrets_loader_data(
 
 def configurable_storage_data(
     config_field: Mapping[str, Any], defaults: Mapping[str, Optional[ConfigurableClassData]]
-) -> Sequence[ConfigurableClassData]:
+) -> Sequence[Optional[ConfigurableClassData]]:
+
+    storage_data: ConfigurableClassData
+    run_storage_data: Optional[ConfigurableClassData]
+    event_storage_data: Optional[ConfigurableClassData]
+    schedule_storage_data: Optional[ConfigurableClassData]
+
     if not config_field:
         storage_data = check.not_none(defaults.get("storage"))
         run_storage_data = check.not_none(defaults.get("run_storage"))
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/instance/ref.py
```diff
@@ -111,26 +117,38 @@ def configurable_storage_data(
...
-            "SqliteScheduleStorage",
-            yaml.dump({"base_dir": _schedule_directory(base_dir)}, default_flow_style=False),
+        schedule_storage_data = (
+            ConfigurableClassData(
+                "dagster._core.storage.schedules",
+                "SqliteScheduleStorage",
+                yaml.dump({"base_dir": _schedule_directory(base_dir)}, default_flow_style=False),
+            )
+            if isinstance(base_dir, str)
+            else None
```
alangenfeld, line 151: i don't quite follow why we have to do this - I see the type set to `StringSource` on these inner classes

python_modules/dagster/dagster/_core/instance/ref.py
```diff
@@ -111,26 +117,38 @@ def configurable_storage_data(
...
+        base_dir = config_field["sqlite"]["base_dir"]
         storage_data = ConfigurableClassData(
             "dagster._core.storage.sqlite_storage",
             "DagsterSqliteStorage",
             yaml.dump({"base_dir": base_dir}, default_flow_style=False),
         )
-        run_storage_data = ConfigurableClassData(
-            "dagster._core.storage.runs",
-            "SqliteRunStorage",
-            yaml.dump({"base_dir": _runs_directory(base_dir)}, default_flow_style=False),
```
gibsondan, line 123: this was the main thing that was breaking it - the base dirs are different for the different components so they can't all be sourced from the passed in env var directly

python_modules/dagster/dagster/_core/instance/ref.py
alangenfeld, line 151: maybe update this to a single if/else block with a nice long comment explaining these subtle shenanigans

alangenfeld (approved):
ok i think is reasonable but lets get this handling for certain scenario laid out more explicitly w/ comments in the code

PR #11032, opened 2022-12-09
Enable multiple files in CLI tools

schrockn:
Summary & Motivation: We also want to allow multiple ad hoc loading of files in separate code locations. This does this.

Test Plan:
Included tests and manually loaded two files in the same directory in dagit and verified they loaded.


python_modules/dagster/dagster/_cli/workspace/cli_target.py
```diff
@@ -133,13 +133,37 @@ def get_workspace_load_target(kwargs: Mapping[str, str]):
...
+            )
+        else:
+            # multiple files
+
+            if kwargs.get("attribute"):
+                raise UsageError(
+                    "If you are specifying multiple files you cannot specify an attribute."
+                )
+
+            return CompositeTarget(
```
gibsondan, line 155: You could alternately return a list of targets from this function vs. a new target type that wraps a list

gibsondan (approved):

python_modules/dagster/dagster/_cli/workspace/cli_target.py
schrockn: That would a non-trivial refactor, involving making WorkspaceProcessContext accept a list of targets, unless I am missing something

PR #11033, opened 2022-12-10
Replace usages of `nslookup` with `nc` for user deployments

michaeljguarino:
### Summary & Motivation
The effect of the two commands is basically the same, confirming the user deployments are functioning.  `nc` is slightly more robust as `nslookup` will fail if it fails on any of it's queries it takes, and apparently there are some cases where kubedns will be quirky and makes it spaz out.

I wouldn't mind help running the helm unit tests

### DIFF
--- a/helm/dagster/templates/deployment-daemon.yaml
+++ b/helm/dagster/templates/deployment-daemon.yaml
@@ -58,7 +58,7 @@ spec:
         {{- range $deployment := $userDeployments.deployments }}
         - name: "init-user-deployment-{{- $deployment.name -}}"
           image: {{ include "dagster.externalImage.name" $.Values.busybox.image | quote }}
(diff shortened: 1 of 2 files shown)


gmontanola:
**Anedoctal evidence** but our `busybox` tag was set to `stable` and the stable version of busybox (1.34) is facing some problems with `nslookup`. Our pods were stuck on init because of this. 

Reverting to 1.28 (helm chart tag) fixed the issue.

michaeljguarino:
nslookup definitely seems to have a lot of potential footguns in how it returns error codes, so moving away from it seems sound regardless if there's a fix to me

johannkm:
Sounds good to me, I'll run it through our integration tests and merge assuming everything looks good. Thanks for the PR!

johannkm (approved):

PR #11037, opened 2022-12-10
move ecs task tagging to run_task call

prha:
### Summary & Motivation

ECS run tasks are not getting tagged with the appropriate tags.  Instead of calling `ecs.tag_resource` on the newly created run task (which relies on the task to be up and running, which is almost surely not going to be true), this instead constructs the set of desired tags and passes it into the `ecs.run_task` call.

Manually tested against set of serverless run launches.  Also changed the ECS stubs so that tags passed in the `run_task` calls were tracked in the unit tests.


gibsondan (approved):
Fewer ECS calls seems nice

PR #11042, opened 2022-12-12
[dagit] Persist AssetPartitions state to URL, open Materialize panel with same selection

bengotow:
### Summary & Motivation

This PR moves the "Completed, Missing" checkbox state and the selected partition range to the URL. A new option on the hook allows you to choose whether the URL should be kept in sync or just used for initial state.

This means 1) you can share links to the partitions page with selections made, and 2) when you open the backfill modal, it's prefilled with the selection you already made on the partitions page.

I tested this by viewing the partitions page with all four styles of assets (multidimensional static, multidimensional time partitioned, static, and unpartitioned), reloading the page with various selections, and opening / editing in the "Materialize..." modal.

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/AssetPartitions.tsx
+++ b/js_modules/dagit/packages/core/src/assets/AssetPartitions.tsx
@@ -2,6 +2,7 @@ import {Box, Colors, Icon, Spinner, Subheading} from '@dagster-io/ui';
 import * as React from 'react';
 
 import {LiveDataForNode} from '../asset-graph/Utils';
+import {useQueryPersistedState} from '../hooks/useQueryPersistedState';
 import {PartitionRangeWizard} from '../partitions/PartitionRangeWizard';
 import {PartitionStateCheckboxes} from '../partitions/PartitionStateCheckboxes';
 import {PartitionState} from '../partitions/PartitionStatus';
@@ -19,7 +20,7 @@ import {PartitionHealthDimensionRange, usePartitionHealthData} from './usePartit
 
 interface Props {
   assetKey: AssetKey;
-  assetPartitionNames?: string[];
+  assetPartitionDimensions?: string[];
   liveData?: LiveDataForNode;
   params: AssetViewParams;
   paramsTimeWindowOnly: boolean;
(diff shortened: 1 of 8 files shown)


js_modules/dagit/packages/core/src/assets/AssetPartitions.tsx
```diff
@@ -33,21 +34,29 @@ interface Props {
...
   params,
   setParams,
   liveData,
 }) => {
   const [assetHealth] = usePartitionHealthData([assetKey], assetLastMaterializedAt);
-  const [ranges, setRanges] = usePartitionDimensionRanges(assetHealth, assetPartitionNames);
+  const [ranges, setRanges] = usePartitionDimensionRanges(
+    assetHealth,
+    assetPartitionDimensions,
+    true,
```
hellendag, line 51: Boolean function params are a mystery -- at a glance I can't tell what this is or is not going to do. :) Perhaps it would help to make the function parameter a config object?

js_modules/dagit/packages/core/src/assets/AssetPartitions.tsx
```diff
@@ -33,21 +34,29 @@ interface Props {
...
+  );
 
-  const [stateFilters, setStateFilters] = React.useState<PartitionState[]>([
-    PartitionState.MISSING,
-    PartitionState.SUCCESS,
-  ]);
+  const [stateFilters, setStateFilters] = useQueryPersistedState<PartitionState[]>({
+    defaults: {states: DISPLAYED_STATES.sort().join(',')},
```
hellendag, line 55: Doesn't look like it matters much on this array, but remember that `sort()` is mutative

js_modules/dagit/packages/core/src/partitions/SpanRepresentation.tsx
```diff
@@ -0,0 +1,62 @@
...
+  }
+  return result.sort((a, b) => all.indexOf(a) - all.indexOf(b));
+}
+
+export function partitionsToText(selected: string[], all: string[]) {
+  return assembleIntoSpans(all, (key) => selected.includes(key))
+    .filter((s) => s.status)
+    .map((s) => stringForSpan(s, all))
+    .join(', ');
+}
```
hellendag, line 62: These look like utilities that would be fairly straightforward to write tests for. Mind adding a few?

hellendag (approved):

js_modules/dagit/packages/core/src/assets/AssetPartitions.tsx
bengotow: Good cal! I think it's time to make this an options object 🙌

js_modules/dagit/packages/core/src/assets/AssetPartitions.tsx
bengotow: Mmm good catch this is not good

js_modules/dagit/packages/core/src/partitions/SpanRepresentation.tsx
bengotow: Yep! I think everything in this new file can have tests - will add some today and update this 🙌

PR #11043, opened 2022-12-12
[dagit] Render assets with self-dependencies in Dagit

bengotow:
### Summary & Motivation

This PR makes a few small changes to Dagit to support assets that rely on earlier partitions of themselves (pending backend changes):

- When an asset lists itself in it’s dependencyKeys, we ignore that edge and render the graph without it.

- When an asset lists itself in it’s dependencyKeys, we show a small banner in the sidebar and within the “Upstream Assets” section of the definition tab.

- The lineage graph also ignores self-dependencies when computing the “depth” of the graph. (I also renamed this function because it was doing depth-first search and not breadth-first search)

Note: This doesn't handle asset cycles, just assets with immediate self-dependencies. If you create a scenario where A relies on B, and B relies on yesterday's A, this notice is not displayed and Dagit will bail out of rendering the graph when it detects the cycle. I'm not sure if this scenario exists.

image
image

I setup the repro case described here: https://elementl-workspace.slack.com/archives/C03CA4TVCAW/p1670541710139359, and then visited all the asset views to confirm they no longer crash or end up in infinite recursion.

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
@@ -447,7 +447,7 @@ export const AssetGraphExplorerWithData: React.FC<
           <RightInfoPanel>
             <RightInfoPanelContent>
               <SidebarAssetInfo
-                assetKey={selectedGraphNodes[0].assetKey}
+                assetNode={selectedGraphNodes[0]}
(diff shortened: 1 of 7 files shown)


sryza (approved):
Great great.

Only piece of feedback: The "This asset depends on earlier partitions of itself" banner is the same colors as "stale", which I think implies it's kind of some sort of warning state. I think we should treat it as perfectly legitimate, i.e. not a warning state.

PR #11044, opened 2022-12-12
[dagit] Add empty value string for invalid tag input in Runs filter

hellendag:
### Summary & Motivation

Now that we don't show the full list of available tags to search Runs on, tags must be either entered with free text manually or via some other more structured means (e.g. clicking on the hover menu on a tag in the Runs list). The result is that it's possible to type an invalid tag pretty easily, which leads to a GraphQL error if the `value` field in the query variable is missing.

To resolve this, just tack on an empty string for the `value`. The query result will correctly come up empty (because no tags will match) and avoids the GraphQL error.

Screenshot 2022-12-12 at 10 22 57 AM

View Runs list. Type `tag:job_name`, verify that the GraphQL query is performed without any errors. Repeat search with a valid tag query by clicking a tag in the Run list, verify that it filters successfully.

### DIFF
--- a/js_modules/dagit/packages/core/src/runs/RunsFilterInput.tsx
+++ b/js_modules/dagit/packages/core/src/runs/RunsFilterInput.tsx
@@ -93,7 +93,7 @@ export function runsFilterForSearchTokens(search: TokenizingFieldValue[]) {
     } else if (item.token === 'snapshotId') {
       obj.snapshotId = item.value;
     } else if (item.token === 'tag') {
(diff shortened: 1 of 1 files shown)


hellendag:
* master
  * **PR #11044** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11044" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11044?utm_source=stack-comment).

PR #11054, opened 2022-12-12
[ui] Show toast for GraphQL timeouts in Dagit

hellendag:
### Summary & Motivation

When the `RootWorkspaceQuery` GraphQL query times out with a 504 gateway timeout, we currently do nothing to surface this to the user, so it looks like Dagit is just empty. I'm not sure there's a great way to surface these errors strictly for this operation, but we could surface a toast for any 504s to give the user a little bit of context about what's going on.

To accomplish this, I added a toast for 504/408 errors in the Apollo error link. I'm not sure this is really the correct thing to do in all cases, though. It might get pretty noisy in some circumstances.

This change also includes:

- Making the "warning" toast gray instead of red. It's weird that both "error" and "warning" toasts are currently red.
- Add a little more padding to the toast.
- Redirect to `/locations` in the fallthrough route if there are no code locations. This is a much better place to dump the user than the weird mystery empty state we currently show.

Return a 504 error for `RootWorkspaceQuery` with a special case in Python.

Load Dagit, verify that the gray toast appears to indicate the 504. Verify also that I am redirected to `/locations` from the `/` fallthrough when no code locations have successfully loaded.

### DIFF
--- a/js_modules/dagit/packages/core/src/app/AppError.tsx
+++ b/js_modules/dagit/packages/core/src/app/AppError.tsx
@@ -1,5 +1,5 @@
+import {ServerError} from '@apollo/client';
 import {ErrorResponse, onError} from '@apollo/client/link/error';
-import {ServerError} from '@apollo/client/link/utils';
 import {Observable} from '@apollo/client/utilities';
 import {FontFamily, Toaster} from '@dagster-io/ui';
 import {GraphQLError} from 'graphql';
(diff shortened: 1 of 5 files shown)


hellendag:
* master
  * **PR #11054** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11054" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11054?utm_source=stack-comment).

bengotow (approved):
This looks good to me -- I'm curious whether the user could trigger timeouts by closing their laptop / opening it later, but I think that would be a status code of `0` and not `408`

I think we could try this and if it's too noisy, we could try to add some mechanism for detecting whether a query is providing first-time data or is refreshing data already in the cache? I feel like the notice is most important for queries without previous data.

hellendag:
image

In dev, this will appear alongside the red console error toast.

PR #11064, opened 2022-12-13
Make JobSelector work on cross code location

schrockn:
Summary & Motivation: With repository going away we need the ability to select a single job in a different code location that does not have a named repository. This works by defaulting the repository_name on JobSelector to "__repository__" and adds a test to verify the behavior.

Because "repository_name" comes before "job" name we unfortunately have to make job_name optional whereas it is actually required.

Test Plan: BK


python_modules/dagster/dagster/_core/host_representation/selector.py
```diff
@@ -69,12 +69,21 @@ class JobSelector(
         "_JobSelector", [("location_name", str), ("repository_name", str), ("job_name", str)]
     )
 ):
-    def __new__(cls, location_name: str, repository_name: str, job_name: str):
+    def __new__(
+        cls,
+        location_name: str,
+        repository_name: Optional[str] = None,
+        job_name: Optional[str] = None,
```
sryza, line 76: Why make this optional if it's required?

python_modules/dagster/dagster/_core/host_representation/selector.py
```diff
@@ -69,12 +69,21 @@ class JobSelector(
...
             cls,
             location_name=check.str_param(location_name, "location_name"),
-            repository_name=check.str_param(repository_name, "repository_name"),
-            job_name=check.str_param(job_name, "job_name"),
+            repository_name=check.opt_str_param(
+                repository_name,
+                "repository_name",
+                default=SINGLETON_REPOSITORY_NAME,
+            ),
+            job_name=check.str_param(job_name, "job_name", "Must provide job_name argument."),
```
sryza, line 86: Won't `check` handle composing the appropriate error?

sryza (approved):
disregard my comments

PR #11065, opened 2022-12-13
in default IO manager, handle PartitionMappings that return 0 partitions

sryza:
### Summary & Motivation

This is a prerequisite for supporting assets with self partition deps.

It turned out that I missed some places where partition ranges should have been replaced with subsets in https://github.com/dagster-io/dagster/pull/10909, so this much of this PR is filling those in.


OwenKephart (approved):
great

PR #11066, opened 2022-12-13
Allow assets to depend on earlier partitions of themselves

sryza:
### Summary & Motivation

Support this:
```python
        @asset(
            ins={
                "a": AssetIn(
                    partition_mapping=TimeWindowPartitionMapping(start_offset=-1, end_offset=-1)
                )
            },
        )
        def a(a: Optional[int]) -> int:
            return 1 if a is None else a + 1
```


OwenKephart (approved):
well that's much less code than I'd have expected!

nice that toposort handles self-dependencies without complaining

PR #11083, opened 2022-12-13
Fix bug with logical versions for assets with changed deps

smackesey:
### Summary & Motivation

This fixes a bug with the resolution of the projected logical version for assets with changed deps. Prior to this fix, if a previous dependency of an asset (that it had been most recently materialized with) was deleted, dagit would spam GraphQL errors as the projected logical version resolver endlessly tried to look up the missing asset while calculating provenance staleness. This PR adds a check that the dependencies have not changed  before attempting to calculate provenance staleness.

- Unit test of the "deleted asset" scenario
- Manual dagit repro (h/t @owenkephart) for the repro

- Start with a file `my_defs.py`:

```
from dagster import asset


@asset
def repro_test_start():
    return 1


@asset
def repro_test_a(repro_test_start):
    return 1
```

- In shell, `$ dagit -f my_deps.py` and Materialize All
- Now change `my_defs.py` to:

```
from dagster import asset


@asset
def repro_test_a():
    return 1
```

- Reload definitions in dagit-- prior to fix produced GQL error, after fix, no longer does.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/implementation/loader.py
+++ b/python_modules/dagster-graphql/dagster_graphql/implementation/loader.py
@@ -500,7 +500,9 @@ class ProjectedLogicalVersionLoader:
     # Returns true if the current logical version of at least one input asset differs from the
     # recorded logical version for that asset in the provenance. This indicates that a new
     # materialization with up-to-date data would produce a different logical verson.
-    def _is_provenance_stale(self, node: ExternalAssetNode, provenance: LogicalVersionProvenance) -> bool:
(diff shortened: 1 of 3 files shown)


smackesey:
* master
  * **PR #11083** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11083" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11083?utm_source=stack-comment).

OwenKephart (approved):

PR #11085, opened 2022-12-13
handle self-dependencies in reconciliation sensor

sryza:
### Summary & Motivation

If asset1.partition2 depends on asset1.partition1, avoid launching runs for asset1.partition2 until asset1.partition1 is reconciled.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -34,6 +34,7 @@ from .partition import PartitionsDefinition, PartitionsSubset
 from .repository_definition import RepositoryDefinition
 from .run_request import RunRequest
 from .sensor_definition import DefaultSensorStatus, SensorDefinition
+from .time_window_partitions import TimeWindowPartitionsDefinition
 from .utils import check_valid_name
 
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -366,9 +388,7 @@ def determine_asset_partitions_to_reconcile(
                         # if they don't have the same partitioning, then we can't launch a run that
                         # targets both, so we need to wait until the parent is reconciled before
                         # launching a run for the child
-                        and asset_graph.have_same_partitioning(
-                            parent.asset_key, candidate.asset_key
-                        )
+                        and parent.partition_key == candidate.partition_key
```
OwenKephart, line 391: should we keep the original conditional as well? I forget how partition keys are generated but is it possible that a weekly and a daily partitions def both have a key of `2022-01-01`?

OwenKephart (approved):
see comment, otherwise looks good

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
sryza: I think you are right

PR #11097, opened 2022-12-13
Add a graphql field and API for location-scoped permissions

gibsondan:
Summary:
Adds a permissions field to GrapheneRepoLocation that can be used to scope permissions to particular code locations. The idea is that the frontend will, for these permissions, instead of sourcing them for a single static map, source them from a map for each code location.

A thing we will need to figure out is how to handle permissions checks during edge cases where the underlying location is no longer in the workspace (for example, terminating a run after its code location has been removed from the workspace). One possibility to handle that case would be to add canXXX fields on the relevant graphql object that is then mutated. Another is to fall back to the global permissions map to check if they are permissioned to do it for *all* locations (which is what the server currently does).

Added an example handler for launching and re-exeucting a run to demonstrate what the serverside checks would look like.

### DIFF
--- a/js_modules/dagit/packages/core/src/graphql/graphql.ts
+++ b/js_modules/dagit/packages/core/src/graphql/graphql.ts
@@ -3425,6 +3425,7 @@ export type WorkspaceLocationEntry = {
   loadStatus: RepositoryLocationLoadStatus;
   locationOrLoadError: Maybe<RepositoryLocationOrLoadError>;
   name: Scalars['String'];
+  permissions: Array<Permission>;
   updatedTimestamp: Scalars['Float'];
 };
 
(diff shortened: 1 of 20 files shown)


prha:
I think it's out of scope for this PR... but I'm mildly concerned that it's harder to see if we're missing a permission check (i.e. it's harder to see that the mutation field is not decorated by a permission check decorator).

I wonder if there's something that we can do to the request context on mutations to verify that SOME permission was checked.

gibsondan:
Yeah you could turn an UncheckedContext into a CheckedContext somehow? What's not clear to me is what exactly the boundary would be where it throws if the context is unchecked

alangenfeld (commented):
Bummer that the assert placement is inconsistent to get at the information it needs. 

Do we have integration level tests that ensure all these guarded calls fail when not permissioend? We probably should be somewhat thorough to protect asserts from future refactors.

alangenfeld (approved):

gibsondan:
I took a stab at making this stricter using a decorator that raises an Exception if a request succeeds but never checked the given permission - this would be to surface code errors rather than actually prevent a bad write (to do it before the write we'd need to somehow intercept it at the storage layer which seems quite tricky)

alangenfeld (approved):
nice seems clean to me

PR #11100, opened 2022-12-13
keep track of max timestamps client side for code location updates

prha:
### Summary & Motivation
Users with more than 1 dagit instance running, may get repeated 'Definitions updated' toasts because each dagit instance keeps a timestamp of when the locations were loaded into memory, and we show toasts based on the timestamp being greater than the last timestamp queried.

This diff keeps track of the largest timestamp seen per location, to guard against repeated requests.  You may still see N-1 number of toasts, where N is the number of dagit instances you have servicing graphql requests.

We could also try to implement this polling via a server-side subscription, but might be out of scope for now.

Changed the graphql resolver to randomly return one of two timestamps, saw the repeated toasts.  Checked that this solution kept track of the max timestamp.

### DIFF
--- a/js_modules/dagit/packages/core/src/nav/useCodeLocationsStatus.tsx
+++ b/js_modules/dagit/packages/core/src/nav/useCodeLocationsStatus.tsx
@@ -24,6 +24,7 @@ export const useCodeLocationsStatus = (skip = false): StatusAndMessage | null =>
   const history = useHistory();
 
   const [showSpinner, setShowSpinner] = React.useState(false);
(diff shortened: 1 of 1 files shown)


salazarm:
Is there a way we could use a hash of the workspace/locations to come up with some consistent versioning?

salazarm (approved):

prha:
Kind of hard, to be honest, since we want to avoid actually loading the locations in cloud for performance reasons.  The main motivation for the timestamp approach was to be able to detect changes very cheaply.  One thing I could see doing is to store a hash in a db column in cloud and just load that cheaply, but would require some schema migrations.

PR #11103, opened 2022-12-14
[dagster-dbt] Enable setting FreshnessPolicies on dbt assets

OwenKephart:
### Summary & Motivation

Adds a new `node_info_to_freshness_policy_fn` parameter to `load_assets_from_dbt*` (including the dbt cloud version). For dbt Cloud, had to get a little funky and went down a quick rabbit hole of adding a freshness_policies_by_output_name parameter to the AssetsDefinitionCacheableData class.

Arguably, I could have just passed that information through `extra_metadata`, as we do for group_names_by_key, but a) that would require custom serde logic and b) freshness policies are a product direction we're investing heavily in, so it felt natural to provide first-class support here.

Also had to add freshness_policy to the `AssetOut` class in order to enable setting FreshnessPolicies using the `@multi_asset` decorator, but again that seems like something we should have done already anyways.


python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
```diff
@@ -186,6 +187,17 @@ def _get_node_metadata(node_info: Mapping[str, Any]) -> Mapping[str, Any]:
     return metadata
 
 
+def _get_node_freshness_policy(node_info: Mapping[str, Any]) -> Optional[FreshnessPolicy]:
+
```
sryza, line 191: Nitpick: remove newline between function def and body

sryza (approved):
Nice

rexledesma (changes requested):
can we add test_custom_freshness_policy to cloud as well?

rexledesma (approved):

PR #11105, opened 2022-12-14
[dagster-aws] Extend fake S3 resource

asharov:
### Summary & Motivation

In our Dagster pipelines we are using more features of the S3 client than was provided by `FakeS3Session`. Since we use this class in our tests, we extended it to align it better with the [boto3 S3 client API](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html).

The changes:
- `list_objects` and `list_objects_v2` return all keys starting with the supplied prefix, not just one key matching the prefix
- `list_objects_v2` returns keys in sorted order (`list_objects` does too, but the actual ordering is not specified in the API documentation)
- `put_object` accepts a `bytes` object in addition to a file-like object
- `delete_object`, `upload_file`, and `download_fileobj` were implemented

These changes all follow the semantics of the boto3 API.

We have been using this version for quite some time already in our Dagster repositories as the S3 resource for tests.


jamiedemaria (approved):
this looks good to me! thanks for the contribution!

PR #11106, opened 2022-12-14
Add executor and loggers to Definitions

schrockn:
Summary & Motivation: This adds top-level executor and logger top-level arguments to Definitions.

A couple notes:

1) To be consistent with the renaming of "resource_defs" to "resources" not including defs suffix
2) Not calling it "default". This implies that in the definitions world the common case will be using this
API rather than specifying it on the job-level.
3) Marking experimental to allow for change when we make resources late-binding on jobs as well to provide option
value on this API.

Test Plan: Unit tests


sryza (commented):
> To be consistent with the renaming of "resource_defs" to "resources" not including defs suffix

My understanding of the reason we went with "resources" instead of "resource_defs" was that we actually aren't requiring `ResourceDefinition` objects as the values. Is the idea that we would do that here as well? Do we want to make a general shift in our public APIs away from the `_def` suffix?

> Not calling it "default". This implies that in the definitions world the common case will be using this API rather than specifying it on the job-level.

I don't have a strong opinion here in either direction, but wanted to call out that an advantage of `default_executor` over `executor` is that it makes it super clear what takes precedence if an executor is supplied both on the `Definitions` and on a job.

> Marking experimental to allow for change when we make resources late-binding on jobs as well to provide option value on this API.

What would we say to users who are using these arguments on `@repository` and want to stick with stable, non-deprecated APIs?

schrockn:
> My understanding of the reason we went with "resources" instead of "resource_defs" was that we actually aren't requiring `ResourceDefinition` objects as the values. Is the idea that we would do that here as well? Do we want to make a general shift in our public APIs away from the `_def` suffix?

Yes this gives us the option value to do that. I think we should be open to coercion in this case but it is not as critical.

> Not calling it "default". This implies that in the definitions world the common case will be using this API rather than specifying it on the job-level.

Agree that this is a downside. However, we are going to run into this same issue when we late bind resources to jobs.

> What would we say to users who are using these arguments on `@repository` and want to stick with stable, non-deprecated APIs?

Don't migrate for now

sryza (approved):

PR #11107, opened 2022-12-14
Coerce IOManager objects into IOManagerDefinitions in Definitions

schrockn:
Summary & Motivation: Right now if you pass in a raw I/O manager object into the resources argument of Definitions it counterintutively is wrapped in a ResourceDefinition rather than an IOManagerDefinition. This change detects if an incoming object is an instance of IOManager and coerces it.

Test Plan: BK


python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -69,7 +70,9 @@ def __init__(
         (via with_resources) to any asset passed to it. If you need to apply different
         resources to different assets, use legacy @repository and use with_resources as before.
 
-        (2) The resources dictionary takes raw python objects, not just resource definitions.
+        (2) The resources dictionary takes raw python objects, not just resource definitions. If
```
sryza, line 73:
```suggestion
        (2) The resources dictionary takes raw Python objects, not just ResourceDefinitions. If
```

python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -69,7 +70,9 @@ def __init__(
...
+        that raw object inherits from IOManager, it gets coerced at an IOManagerDefinition. Any other
```
sryza, line 74:
```suggestion
        that raw object inherits from IOManager, it gets coerced to an IOManagerDefinition. Any other
```

python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -69,7 +70,9 @@ def __init__(
...
+        object is coerced to a ResourceDefinition
```
sryza, line 75:
```suggestion
        object is coerced to a ResourceDefinition.
```

sryza (approved):

PR #11110, opened 2022-12-14
fix a small bug in UPathIOManager

danielgafni:
`self.get_metadata` was called with arguments in the wrong order

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/upath_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/upath_io_manager.py
@@ -45,7 +45,7 @@ class UPathIOManager(MemoizableIOManager):
 
     def get_metadata(
         self,
-        context: InputContext,
+        context: OutputContext,
         obj: Any,  # pylint: disable=unused-argument
     ) -> Dict[str, MetadataValue]:
         """Child classes should override this method to add custom metadata to the outputs."""
--- a/python_modules/dagster/dagster/_core/storage/upath_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/upath_io_manager.py
@@ -45,7 +45,7 @@ class UPathIOManager(MemoizableIOManager):
 
     def get_metadata(
         self,
-        context: OutputContext,
+        context: OutputContext,  # pylint: disable=unused-argument
         obj: Any,  # pylint: disable=unused-argument
     ) -> Dict[str, MetadataValue]:
         """Child classes should override this method to add custom metadata to the outputs."""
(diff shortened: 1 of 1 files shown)


danielgafni:
@sryza

danielgafni:
Also, do we need to log the metadata when loading input?

For assets, the exact same metadata is displayed in the logs as `ASSET_OBSERVATION` anyway. So right now it's being displayed twice.

What would happen when loading OP results?

sryza:
Mind running `make black`. A few formatting errors.

And a couple mypy complaints:
```
[2022-12-14T18:07:56Z] dagster_tests/core_tests/storage_tests/test_upath_io_manager.py:320: error: Argument 1 of "get_metadata" is incompatible with supertype "UPathIOManager"; supertype defines the argument type as "Union[InputContext, OutputContext]"  [override]
--
  | [2022-12-14T18:07:56Z] dagster_tests/core_tests/storage_tests/test_upath_io_manager.py:320: note: This violates the Liskov substitution
```

sryza:
> Also, do we need to log the metadata when loading input? We already do it when handling output.

What exactly do you mean by "log the metadata"? Create a log message that includes the path? I don't know if I would do that if I were starting from scratch, but would leave it for now because some people are probably relying on it.

danielgafni:
I mean these lines:

```python
metadata = {"path": MetadataValue.path(path)}
custom_metadata = self.get_metadata(obj, context)
metadata.update(custom_metadata)  # type: ignore
context.add_input_metadata(metadata)
```

we allow to define the `self.get_metadata` method do add additional metadata besides `path`.

sryza:
Ahh I see. I hadn't actually noticed that function. I would think it depends on the situation - some metadata might be expensive to compute and shouldn't happen on every load. Would it make sense for the developer subclassing UPathIOManager to just invoke `add_input_metadata` and `add_output_metadata`?

danielgafni:
not really, because the `load_from_path` method is sometimes used in a partitioned context and for some reason `add_output_metadata` fails in this case. Check [this](https://github.com/dagster-io/dagster/blob/8cbb4c68ded1872e826c39fa322237ea67e823b6/python_modules/dagster/dagster/_core/storage/upath_io_manager.py#L139) comment. 

Should we remove the `.get_metadata` call on loading for now?

danielgafni:
I don't remember the error, sorry, You should be able to reproduce it by just logging output metadata with a partitioned context.

sryza:
It looks like there's a failure that might be related:
```
[2022-12-15T17:47:28Z] dagster_tests/core_tests/storage_tests/test_fs_io_manager.py::test_fs_io_manager FAILED [ 87%]
```

And one lint complaint:
```
[2022-12-15T17:36:01Z] ************* Module dagster._core.storage.upath_io_manager
[2022-12-15T17:36:01Z] dagster/_core/storage/upath_io_manager.py:48:8: W0613: Unused argument 'context' (unused-argument)
```

danielgafni:
should be fixed

sryza (approved):
looks good!

PR #11114, opened 2022-12-14
[dagit] Add tests for partition health data parsing / accessors

bengotow:
### Summary & Motivation

There's a lot of business logic in `usePartitionHealthData` that transforms the GraphQL response and provides accessors that are independent of the number of dimensions. This PR adds some tests for the range access, etc.

New tests! 🎉

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/usePartitionHealthData.tsx
+++ b/js_modules/dagit/packages/core/src/assets/usePartitionHealthData.tsx
@@ -1,4 +1,4 @@
-import {ApolloClient, gql, useApolloClient} from '@apollo/client';
+import {gql, useApolloClient} from '@apollo/client';
 import isEqual from 'lodash/isEqual';
 import React from 'react';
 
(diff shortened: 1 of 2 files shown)


js_modules/dagit/packages/core/src/assets/usePartitionHealthData.tsx
```diff
@@ -39,7 +39,7 @@ export type PartitionHealthDimensionRange = {
   selected: string[];
 };
 
-async function loadPartitionHealthData(client: ApolloClient<any>, loadKey: AssetKey) {
+export async function loadPartitionHealthData(client: ApolloClient<any>, loadKey: AssetKey) {
```
hellendag, line 42: Since you're testing operations on the retrieved data, it might be simpler to export a function that encapsulates everything *after* the `client.query` call. Then you don't need to mock the `client.query` at all, you can just pass in your mocked objects.

hellendag (approved):

js_modules/dagit/packages/core/src/assets/usePartitionHealthData.tsx
bengotow: Ahh yeah let me switch to that, would be slightly cleaner than mocking the query function

PR #11118, opened 2022-12-14
[declarative-scheduling] Fix issue where in-progress runs were not properly handled

OwenKephart:
### Summary & Motivation

Previously, the way we calculated if an in-progress run would solve a given constraint was very naive, and only covered a small subset of the actual scenarios. Specifically, for a given asset key, it would check if the required upstream data key was in the most recent run of the given asset key.

However, this would not take into account situations where a currently-executing run of that asset key did not contain that upstream key, but would still pull in fresher data than the asset currently had (which is a pretty common scenario).

### DIFF
--- a/python_modules/dagster/dagster/_utils/caching_instance_queryer.py
+++ b/python_modules/dagster/dagster/_utils/caching_instance_queryer.py
@@ -1,6 +1,17 @@
 import datetime
 import json
-from typing import TYPE_CHECKING, AbstractSet, Dict, Iterable, Mapping, Optional, Tuple, Union, cast
+from typing import (
+    TYPE_CHECKING,
+    AbstractSet,
+    Dict,
+    Iterable,
+    Mapping,
+    Optional,
+    Sequence,
+    Tuple,
+    Union,
+    cast,
+)
 
 import dagster._check as check
 from dagster._core.definitions.asset_graph import AssetGraph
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -97,6 +117,26 @@ def _get_planned_materializations_for_run(self, run_id: str) -> AbstractSet[Asse
         ).records
         return set(cast(AssetKey, record.asset_key) for record in materializations_planned)
 
+    @cached_method
+    def _get_expected_materializations_for_run(self, run_id: str) -> AbstractSet[AssetKey]:
```
sryza, line 121: This makes me think "what's the difference between a 'planned materialization' and an 'expected materialization'?" What do you think about calling these something like `_get_assets_with_materialization_planned_events_for_run` and `_get_planned_materializations_for_run`?

Alternatively, we could just not make it so complicated (my fault) and always just look at the materialization_planned events.

sryza (approved):

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
OwenKephart: mm yeah good point -- I do think having the code path branch here is a good idea because a large percentage of jobs will go down the cheap path (basically any ad-hoc run and any run kicked off by this sensor), so skipping the expensive query as often as possible is nice

will rename though

PR #11128, opened 2022-12-14
Mark Definitions as not experimental

schrockn:
Summary & Motivation: This is front and center in our content and docs now, so remove experimental flag.

Test Plan: Load definitions. See no experimental warning


yuhan (approved):
:lfg

python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -29,7 +29,6 @@
     from dagster._core.storage.asset_value_loader import AssetValueLoader
 
 
-@experimental
```
sryza, line 32: Not critical for this PR, but a few comments on the docstring that I noticed while reviewing:
- "Create a set of definitions explicitly available and loadable by Dagster tools." - now that this is a class instead of a method, maybe take out the "Create". Also, this should probably be the first line?
- The bullet points blow past line width.
- It's not clear to me what's meant by "do not apply to vanilla Dagster definitions".

sryza (approved):

python_modules/dagster/dagster/_core/definitions/definitions_class.py
schrockn: The bullets need to be on a single line for them to be appropriately parsed by sphinx. Or at least that was the only way I was able to get sphinx compile.

schrockn: Sandy how about "Definitions provides a few conveniences for dealing with resources that do not apply to definitions groups with ``@repository` declarations."

schrockn: @sryza  we can discuss in https://github.com/dagster-io/dagster/pull/11134

PR #11134, opened 2022-12-14
Update Definitions docs

schrockn:
Summary & Motivation: Sandy had good feedback about some unclear language in Definitions docblock

Test Plan: Read


python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -56,8 +56,8 @@ class Definitions:
     A Python module is loadable by Dagster tools if there is a top-level variable
     that is an instance of Definitions.
 
-    Definitions provides a few conveniences for dealing with resources that do not apply to
-    vanilla Dagster definitions:
+    Definitions provides a few conveniences for dealing with resources that do
+    not apply to old-style `@repository` declarations:
```
sryza, line 60: "that do not apply to old-style `@repository` declarations:" could be a little jarring to a new user who has never encountered repositories. I would maybe preface this with something like "Before the introduction of the `Definitions` class, `@repository` was the API for bla blah. These are the main differences:".

python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -44,7 +44,7 @@ class Definitions:
             }
         )
 
-    Create a set of definitions explicitly available and loadable by Dagster tools.
+    A set of definitions to be explicitly available and loadable by Dagster tools.
```
sryza, line 47: [PEP](https://peps.python.org/pep-0257/#multi-line-docstrings) would recommend putting this at the top: 

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description. The summary line may be used by automatic indexing tools; it is important that it fits on one line and is separated from the rest of the docstring by a blank line. The summary line may be on the same line as the opening quotes or on the next line. The entire docstring is indented the same as the quotes at its first line (see example below).

sryza (commented):
Regarding multi-line docstrings, I believe there's way to do the indentation so that Sphinx renders them correctly. The `ConfigSchema` class is an example of maximally-complex sphinxdoc that renders nicely.

python_modules/dagster/dagster/_core/definitions/definitions_class.py
schrockn: 👍🏻

sryza (approved):
This looks great

PR #11139, opened 2022-12-15
[run coordinator] fix threaded tests

alangenfeld:
use set instead of list since order is not deterministic 

pytest --count 50, bk

### DIFF
--- a/python_modules/dagster/dagster_tests/daemon_tests/test_queued_run_coordinator_daemon.py
+++ b/python_modules/dagster/dagster_tests/daemon_tests/test_queued_run_coordinator_daemon.py
@@ -285,7 +285,8 @@ def test_tag_limits(use_threads, workspace_context, pipeline_handle, daemon):
 
         list(daemon.run_iteration(bounded_ctx))
 
-        assert get_run_ids(instance.run_launcher.queue()) == ["tiny-1", "large-1"]
+        # exact order non-deterministic due to threaded dequeue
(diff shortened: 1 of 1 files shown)


alangenfeld:
* master
  * **PR #11139** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11139" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11139?utm_source=stack-comment).

python_modules/dagster/dagster_tests/daemon_tests/test_queued_run_coordinator_daemon.py
```diff
@@ -479,7 +479,7 @@ def test_limits_per_unique_value(use_threads, workspace_context, pipeline_handle
 
         list(daemon.run_iteration(bounded_ctx))
 
-        assert get_run_ids(instance.run_launcher.queue()) == ["run-1", "run-3"]
+        assert set(get_run_ids(instance.run_launcher.queue())) == {"run-1", "run-3"}
```
schrockn, line 482: I think this is worth a comment. Seeing a "queue" and then adding sets around it to make order not matter is an eyebrow raiser. I imagine adding threads loosened the ordering guaranteed of this queue

schrockn (approved):
Approved but please see comment and add comment in test case

PR #11145, opened 2022-12-15
enable run status sensor logging

prha:
### Summary & Motivation
Propagates the sensor evaluation context to the run status sensor context, to give access to the context logger.

Passed the context instead of the logger to avoid the logger instantiation until it was actually used.  Required converting the context object from a namedtuple to an actual class (should be safe since it does not cross a serialization boundary).

BK

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
@@ -6,7 +6,7 @@ from typing import TYPE_CHECKING, Any, Callable, NamedTuple, Optional, Sequence,
 import pendulum
 
 import dagster._check as check
-from dagster._annotations import PublicAttr, public
+from dagster._annotations import public
 from dagster._core.definitions.instigation_logger import InstigationLogger
 from dagster._core.errors import (
     DagsterInvalidDefinitionError,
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -86,45 +88,66 @@ def from_json(json_str: str) -> tuple:
...
+        return self._dagster_event
+
+    @public  # type: ignore
+    @property
+    def instance(self) -> DagsterRun:
+        return self._instance
+
+    @public  # type: ignore
+    @property
+    def log(self) -> logging.Logger:
```
sryza, line 142: I think you could use `@cached_method` for this.

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -86,45 +88,66 @@ def from_json(json_str: str) -> tuple:
...
+        return self._sensor_name
+
+    @public  # type: ignore
+    @property
+    def dagster_run(self) -> DagsterRun:
+        return self._dagster_run
+
+    @public  # type: ignore
+    @property
+    def dagster_event(self) -> DagsterRun:
```
sryza, line 132: Looks like the wrong type annotation.

sryza (approved):

PR #11147, opened 2022-12-15
make schedules produced by build_schedule_from_partitioned_job more p…

sryza:
…erfomant

### Summary & Motivation

Instead of getting a list of all the partition keys, just ask the `PartitionsDefinition` for the last one.

Implementing this required making the return type of `PartitionsDefinition.get_last_partition_key` optional, so we can gracefully skip if there are no partitions yet.


OwenKephart (approved):
nice

PR #11169, opened 2022-12-15
fix precedence ordering when merging dictionaries in container context

gibsondan:
Summary:
Fixes an issue where we were being inconsistent with container_context with dict fields - with scalar fields we prefer the passed in container context (to support things like per-job config overriding global config), but for dictionaries we were doing the reverse.

Test Plan: BK

### DIFF
--- a/python_modules/libraries/dagster-docker/dagster_docker/container_context.py
+++ b/python_modules/libraries/dagster-docker/dagster_docker/container_context.py
@@ -6,7 +6,6 @@ from dagster._config import process_config
 from dagster._core.container_context import process_shared_container_context_config
 from dagster._core.errors import DagsterInvalidConfigError
 from dagster._core.storage.pipeline_run import DagsterRun
-from dagster._utils import merge_dicts
 
 if TYPE_CHECKING:
     from . import DockerRunLauncher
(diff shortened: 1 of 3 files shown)


python_modules/libraries/dagster-docker/dagster_docker/container_context.py
```diff
@@ -88,7 +88,7 @@ def merge(self, other: "DockerContainerContext"):
             registry=other.registry if other.registry != None else self.registry,
             env_vars=[*self.env_vars, *other.env_vars],
             networks=[*self.networks, *other.networks],
-            container_kwargs=merge_dicts(other.container_kwargs, self.container_kwargs),
+            container_kwargs=merge_dicts(self.container_kwargs, other.container_kwargs),
```
alangenfeld, line 91: nit: i've been switching away from our random `merge_dicts` util to language native `{**x, **y}`

alangenfeld (approved):

python_modules/libraries/dagster-docker/dagster_docker/container_context.py
jmsanders: Yes please ❤️

jmsanders (approved):

python_modules/libraries/dagster-docker/dagster_docker/container_context.py
gibsondan: done

PR #11173, opened 2022-12-16
Ignore stale timestamps from code location updates

prha:
### Summary & Motivation
Take 2 of https://github.com/dagster-io/dagster/pull/11100.

Rather than recalculate the current set of updates in a useEffect (prone to infinite render chains), this sets state within a query callback, but only if the payload has logically changed (by comparing timestamps).

Randomly cycled through a set of values for location timestamps, and saw that only values higher than the known max triggered toast/refreshes

### DIFF
--- a/js_modules/dagit/packages/core/src/nav/useCodeLocationsStatus.tsx
+++ b/js_modules/dagit/packages/core/src/nav/useCodeLocationsStatus.tsx
@@ -21,16 +21,11 @@ type LocationStatusEntry = {
 
 const POLL_INTERVAL = 5 * 1000;
 
-type State = {
-  entriesById: {[key: string]: LocationStatusEntry} | undefined;
-};
-const initialState: State = {
-  entriesById: undefined,
-};
+type EntriesById = Record<string, LocationStatusEntry>;
 
 export const useCodeLocationsStatus = (skip = false): StatusAndMessage | null => {
   const {locationEntries, refetch} = React.useContext(WorkspaceContext);
-  const [state, setState] = React.useState<State>(initialState);
+  const [previousEntriesById, setPreviousEntriesById] = React.useState<EntriesById | null>(null);
 
   const history = useHistory();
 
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/core/src/nav/useCodeLocationsStatus.tsx
```diff
@@ -29,12 +42,37 @@ export const useCodeLocationsStatus = (skip = false): StatusAndMessage | null =>
...
+          ? {
+              id: entry.id,
+              loadStatus: entry.loadStatus,
+              updateTimestamp: entry.updateTimestamp,
+            }
+          : previousEntry;
+      });
+      if (hasUpdates) {
+        setState({entriesById, previousEntriesById: state.entriesById});
+      }
```
hellendag, line 70: Since you're now using `onCompleted`, and since (I think) the effect is just responding to the state change, you might be able to put all of the effect behavior into here and remove the effect entirely.

Also, I think you could put this inside the setter so you don't risk ending up in some weird situation where `state` is some stale value that you don't expect. Not sure how likely that is, though.

```
setState((current) => {
  const {entriesById} = current;
  let hasUpdates = entries.length !== ...;
  // etc.

  return hasUpdates ? {entriesById: newEntriesById, previousEntriesById: entriesById} : current;
});
```

js_modules/dagit/packages/core/src/nav/useCodeLocationsStatus.tsx
```diff
@@ -85,37 +86,38 @@ export const useCodeLocationsStatus = (skip = false): StatusAndMessage | null =>
...
+            updateTimestamp: entry.updateTimestamp,
+          }
+        : previousEntry;
+    });
+    const previousEntriesById = state.entriesById;
+    const currentEntries = Object.values(entriesById || {});
+    const previousEntries = Object.values(previousEntriesById || {});
+    if (hasUpdatedEntries) {
+      setState({entriesById});
+    }
```
hellendag, line 120: This could probably be extracted out into a function (accepting the old and new entries as params) that you could write some tests against, with the `entriesById` being returned to be used in the setter.

hellendag (approved):
One suggestion for simplifying the values, otherwise this looks reasonable to me!

PR #11192, opened 2022-12-16
Retry registering task definitions

jmsanders:
When multiple runs try to register a task definition simultaneously, we can hit an ECS API error:

```
botocore.errorfactory.ClientException: An error occurred (ClientException)
when calling the RegisterTaskDefinition operation: Too many concurrent
attempts to create a new revision of the specified family.
```

Our task definition reuse strategy creates a race condition where:

- multiple simultaneous runs can't find a task defintion
- run A successfully registers the task definition, run B hits the error

Now run B will retry and see that it can reuse the task definition that run A created.

Testing this is proving to be tricky - I initially added a lock to our stub and ran man concurrent threads against it. But getting the botocore stubber interactions to work well with multiple threads was a challenge. I'm putting this up for review now and I'll follow up with a better test
- likely one that makes heavy use of mock.patch but I'm open to suggestions.

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
@@ -500,7 +500,7 @@ class EcsRunLauncher(RunLauncher, ConfigurableClass):
 
             backoff(
                 self._reuse_or_register_task_definition,
-                retry_on=(ClientError,),
+                retry_on=(Exception,),
                 kwargs={
                     "desired_task_definition_config": task_definition_config,
                     "container_name": container_name,
--- a/python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
@@ -19,6 +19,7 @@ from dagster._core.launcher.base import (
 from dagster._core.storage.pipeline_run import DagsterRun
 from dagster._grpc.types import ExecuteRunArgs
 from dagster._serdes import ConfigurableClass
+from dagster._utils.backoff import backoff
 
 from ..secretsmanager import get_secrets_from_arns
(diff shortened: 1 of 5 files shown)


python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -489,11 +490,18 @@ def _run_task_kwargs(self, run, image, container_context) -> Dict[str, Any]:
...
-            if not self._reuse_task_definition(
-                task_definition_config,
-                self._get_container_name(container_context),
-            ):
-                self.ecs.register_task_definition(**task_definition_dict)
+            container_name = self._get_container_name(container_context)
+
+            backoff(
+                self._reuse_or_register_task_definition,
+                retry_on=(ClientError,),
```
jmsanders, line 497: There might be a more narrow error we can catch on here.

gibsondan (changes requested):
as currently written this is going to call describe_task_definition on each iteration here: https://github.com/dagster-io/dagster/blob/master/python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py#L527 - I think we want to be tactical and only do it around the register call? At least for this specific error that we're retrying around. To reduce the risk of getting rate limited

gibsondan (approved):

jmsanders:
I think we want that behavior - otherwise if Run B requires a task definition that's different than the one Run A requested, it'd fail.

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
jmsanders: For now, I broadened this to `Exception`

jmsanders:
I went a little crazy with our StubbedEcs to allow us to test this 😅 

It was either that or a very verbose, confusing, brittle set of mocks. Or a very complicated integration test setup. Open to alternative suggestions.

Issue #11157, opened 2022-12-15
Unable to materialize any asset manually using the Materialize button


alexismanuel:
### Dagster version

dagster, version 1.1.6

### What's the issue?

When I try to materialize an asset selection of my job, the Materialize selected button does not create the popup to choose which partitions to run:
<img width="984" alt="image" src="https://user-images.githubusercontent.com/22666624/207909372-f31926d2-8b53-4ac3-a439-16e7da9f132f.png">

My assets are defined with the following partition:
```python
fifteen_minute_partitions = TimeWindowPartitionsDefinition(
    cron_schedule="*/15 * * * *",
    start=datetime(2022, 1, 1, 0, 0, 0),
    fmt="%Y-%m-%d %H:%M",
    timezone="Europe/Paris",
)
```

My guess is that the server encounters a timeout while trying to respond because of the large number of partitions.
Here is the http request which is in failure state (502 HTTP code):
<img width="667" alt="image" src="https://user-images.githubusercontent.com/22666624/207910404-5401c2ec-2890-437f-84de-ecba7feb16ce.png">

### What did you expect to happen?

I expected the run popup to appear

### How to reproduce?

Select assets with large number of partitions (over 100K in my case)

### Deployment type

Open Source

### Deployment details

K8S over GKE

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

sryza:
@bengotow - thoughts on whether there's a straightforward way to make this faster?

This is a scenario where operating on ranges instead of individual partitions might speed things up significantly.

bengotow:
Quick update - even on localhost a query for 4 assets with this partition definition takes 32s to respond, so I think we can rule out problems with the config / collision parts of the query.

I think we can restructure this to make it more usable -- this initial loader request just asks for the names of the partition dimensions so that the UI doesn't shift as the data loads in, and it looks like asking for the dimension names is causing the backend to build all the partition keys (which aren't actually returned in this query).

It looks like it takes ~3.1s to retrieve the `33573` partition keys of each of the four assets, and it feels like we could make that faster too, given that they are the same for all four assets. There's a good chance we're spending all this time string formatting the dates.

@sryza there may be broader improvements we need to make to support super-large partition key sets (eg: sending down a representation like "2022-01-01 , */15 * * * *" instead of sending down all the keys, etc.) but we can probably get 50% better performance with some adjustments to the current approach.

PR #11201, opened 2022-12-16
[dagit] Optimizations to backfill UI for large partition key sets

bengotow:
### Summary & Motivation

This PR addresses performance issues reported in https://github.com/dagster-io/dagster/issues/11157

With a large set of partition keys (35000 in this example), the Dagit backfill modal is unacceptably slow to load, and then also renders very slowly as you interact with it. These tweaks make things more reasonable: https://www.loom.com/share/c446d0f2123a45059b5182d533aca057

1) When you open the modal, we fetch the names of the partition dimensions via `dimensionTypes[].name` rather than `partitionKeysByDimension.name` which is much faster. I also extended this backend API to work in the single dimension case so that it matches the number of dimensions returned by partitionKeysByDimension (cc @clairelin135  on this bit!)

2) Add a few optimizations for the "select all" case - when we see the selected partition count and the all partition count are the same, we can show `[A...Z]` fast.

3) Add a few useMemo's that were missing

4) Stop unnecessarily sorting the selected keys. We actually don't rely on this anywhere as far as I can tell and it's very expensive.

5) For 35k items, it turns out it's quite a bit faster to convert the Array to a Set and then use JS's `Set.has()`. Ideally we would convert the selection state itself from an Array to a Set, but it would make this PR much larger and the Set constructor is very very fast. 

6) Update PartitionStatus.tsx to use a helper for building "spans" that creates fewer intermediate data structures.

Tested the backfill modal with existing assets in various states to make sure the spans, etc. still behave the same way. Tested with an asset with 35k partitions to see speed improvements.

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
+++ b/js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
@@ -42,10 +42,7 @@ import {RepoAddress} from '../workspace/types';
 import {executionParamsForAssetJob} from './LaunchAssetExecutionButton';
 import {explodePartitionKeysInRanges, mergedAssetHealth} from './MultipartitioningSupport';
 import {RunningBackfillsNotice} from './RunningBackfillsNotice';
-import {
-  LaunchAssetExecutionAssetNodeFragment_partitionDefinition,
-  LaunchAssetExecutionAssetNodeFragment_partitionKeysByDimension,
-} from './types/LaunchAssetExecutionAssetNodeFragment';
+import {LaunchAssetExecutionAssetNodeFragment_partitionDefinition} from './types/LaunchAssetExecutionAssetNodeFragment';
 import {usePartitionDimensionRanges} from './usePartitionDimensionRanges';
 import {PartitionHealthDimensionRange, usePartitionHealthData} from './usePartitionHealthData';
 import {usePartitionNameForPipeline} from './usePartitionNameForPipeline';
@@ -57,7 +54,6 @@ interface Props {
   assets: {
     assetKey: AssetKey;
     opNames: string[];
-    partitionKeysByDimension: LaunchAssetExecutionAssetNodeFragment_partitionKeysByDimension[];
     partitionDefinition: LaunchAssetExecutionAssetNodeFragment_partitionDefinition | null;
   }[];
   upstreamAssetKeys: AssetKey[]; // single layer of upstream dependencies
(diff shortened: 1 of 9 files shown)


js_modules/dagit/packages/core/src/partitions/SpanRepresentation.tsx
```diff
@@ -51,11 +52,15 @@ export function textToPartitions(selected: string, all: string[]) {
       result.push(term);
     }
   }
-  return result.sort((a, b) => all.indexOf(a) - all.indexOf(b));
+  return Array.from(new Set(result));
```
bengotow, line 55: Because I added a few optimizations based on selected.length === all.length, we need to make sure the selection doesn't have any duplicates. The order of the selected items never really mattered though as far as I can tell!

js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
```diff
@@ -108,11 +104,13 @@ const LaunchAssetChoosePartitionsDialogBody: React.FC<Props> = ({
...
   const mergedHealth = React.useMemo(() => mergedAssetHealth(assetHealth), [assetHealth]);
 
+  const knownDimensions = partitionedAssets[0].partitionDefinition?.dimensionTypes || [];
   const [ranges, setRanges] = usePartitionDimensionRanges({
-    knownDimensionNames: partitionedAssets[0].partitionKeysByDimension.map((d) => d.name),
+    knownDimensionNames: knownDimensions.map((d) => d.name),
     modifyQueryString: false,
     assetHealth: mergedHealth,
   });
+  console.log(ranges);
```
hellendag, line 113: rm

js_modules/dagit/packages/core/src/partitions/PartitionStatus.tsx
```diff
@@ -90,6 +90,8 @@ export const PartitionStatus: React.FC<{
     [partitionNames],
   );
 
+  const selectedSet = React.useMemo(() => new Set(selected), [selected]);
```
hellendag, line 93: Worth doing a `JSON.stringify()` on `selected` (maybe `JSON.stringify([...selected].sort())` or something) in case it's a new object on every render?

js_modules/dagit/packages/core/src/partitions/PartitionStatus.tsx
```diff
@@ -128,22 +130,27 @@ export const PartitionStatus: React.FC<{
...
-    ? assembleIntoSpans(partitionNames, (key) => selected.includes(key)).filter((s) => s.status)
-    : [];
+  const selectedSpans = React.useMemo(
+    () =>
+      selectedSet.size === 0
+        ? []
+        : selectedSet.size === partitionNames.length
+        ? [{startIdx: 0, endIdx: partitionNames.length - 1, status: true}]
+        : assembleIntoSpans(partitionNames, (key) => selectedSet.has(key)).filter((s) => s.status),
+    [selectedSet, partitionNames],
```
hellendag, line 140: Same question as above re: `partitionNames`

hellendag (approved):

js_modules/dagit/packages/core/src/partitions/PartitionStatus.tsx
bengotow: Ahh this is a good q - these are memoized further up in the state, so I think it's best not to re-memoify them here.  I was curious about the Set case above though and did some testing - 

it turns out it's actually faster for a large array to build a new Set every time than attempt to JSON.stringify to compare:

```
array = (new Array(1000000)).fill('1234') 
d = performance.now(); JSON.stringify(array); console.log(performance.now() - d);
61.32737499475479
d = performance.now(); new Set(array); console.log(performance.now() - d);
14.73987501859665
```

PR #11203, opened 2022-12-16
[dagit] Round middle truncate calculations for Firefox

bengotow:
### Summary & Motivation

We've seen rendering issues with MiddleTruncate choosing to truncate where it's not necessary in Firefox. This appears to be caused by a /tiny/ discrepancy between `container.getBoundingClientRect().width` (DOM method of measuring the space allocated to us based on the intrinsic size of the text) and `(ctx.measureText(value).width` (the canvas method for measuring text length)

In cases where a flexbox is sizing the box, the former is ~0.03-0.06px smaller than the latter, which makes MiddleTruncate chop one character out of the center of the string.

I added a Storybook that reproduces this flexbox-parent scenario (which is reproduced in Dagit in the left nav and the Asset Table), and then confirmed that rounding up the values fixes the error and does not produce regressions in the other storybook cases.

### DIFF
--- a/js_modules/dagit/packages/ui/src/components/MiddleTruncate.stories.tsx
+++ b/js_modules/dagit/packages/ui/src/components/MiddleTruncate.stories.tsx
@@ -45,6 +45,33 @@ export const Simple = () => {
   );
 };
 
+export const FlexboxContainerUsage = () => {
+  return (
+    <Box>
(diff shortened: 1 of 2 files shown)


hellendag (approved):
Thank you for tracking this down! Browsers, man...

PR #11214, opened 2022-12-19
[declarative-scheduling] Fix bug with declarative scheduling where repeated calls to get_latest_materialization_record could return incorrect results

OwenKephart:
### Summary & Motivation

TLDR; The test failed before this change, now it passes.

In cases where you call `get_latest_materialization_record(AssetKey("foo"))`, this will return the latest materialization record, and cache it in `self._latest_materialization_record_cache`. Let's say the latest record's id is `500`.

Subsequent calls (assuming no arguments are supplied) will hit that cache. However, if you call `get_latest_materialization_record(AssetKey("foo"), before_cursor=300)`, this will go into the code path that says that AssetKey("foo") is in the cache (well really asset partition, but that's not important), but then it will compare `before_cursor` to the latest record id, and see that it is not greater than the latest record id. This causes it to incorrectly return `None`.

So the bug is actually fairly obvious in hindsight, but what's not obvious is how it avoided detection, as it seems like this would cause problems in most normal test cases. The reason it doesn't is that this problem only crops up if `get_latest_materialization_record(AssetKey("foo"))` is called BEFORE `get_latest_materialization_record(..., before_cursor=...)`. When we do the first loop through the asset reconciliation sensor, we first calculate the used data for each asset with a freshness policy, which is a recursive function that calls that function with the `before_cursor` argument supplied. Because `_calculate_used_data()` is a `@cached_method`, if the asset with a freshness policy is at the bottom of the graph, then all of the `get_latest_materialization_record` calls with that argument supplied happen before later parts of the logic which call the method without that argument.

Even more so, this will only cause problems if an asset materialization consumes an upstream materialization record from a run that is not the most recent run. These two conditions combined make it pretty hard to reproduce in a unit test.

This specific added unit test could probably be minified, but you do need two separate freshness policies in order to force two separate calls to _calculate_used_data for the same asset key. These calls can't use the same upstream cached record, so you need to force a run where one freshness policy follows a different recursive path than the other.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -639,10 +639,19 @@ def determine_asset_partitions_to_reconcile_for_freshness(
                 & constraint_keys
             ) | {key}
 
+            print("-----------------------")
+            import os
+            if key in {
+                AssetKey(["ANALYTICS", "company_perf"]),
+                AssetKey(["ANALYTICS", "company_stats"]),
+            }:
+                os.environ["SHOULD_LOG"] = "yes"
             # figure out the current contents of this asset with respect to its constraints
             current_data_times = get_current_data_times_for_key(
                 instance_queryer, asset_graph, relevant_upstream_keys, key
             )
+            print(key, current_data_times)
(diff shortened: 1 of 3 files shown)


sryza (approved):
Nice catch. This could use a more descriptive title.

slopp (commented):
Confirmed fix on hooli!

PR #11216, opened 2022-12-19
Add test case for binding assets before passing to Definitions

schrockn:
### Summary & Motivation

I thought I was going to have to change code in order to support this use case (binding an asset to with_resources) before passing into Definitions but I was pleasantly surprised to see that this was already supported. Added test case I assumed would have failed.

BK

### DIFF
--- a/python_modules/dagster/dagster_tests/definitions_tests/test_definitions_class.py
+++ b/python_modules/dagster/dagster_tests/definitions_tests/test_definitions_class.py
@@ -13,6 +13,7 @@ from dagster import (
     op,
     repository,
     sensor,
+    with_resources,
 )
 from dagster._check import CheckError
 from dagster._core.definitions.cacheable_assets import (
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster_tests/definitions_tests/test_definitions_class.py
```diff
@@ -400,3 +401,37 @@ def a_logger(_):
...
+                        "passed-through-with-resources"
+                    )
+                },
+            ),
+            asset_two,
+        ],
+        resources={"b_resource": "passed-through-definitions"},
+    )
+
+    defs.get_job_def("__ASSET_JOB").execute_in_process()
```
sryza, line 434: `get_base_job_for_assets` would let you avoid using the magic string.

sryza (approved):

schrockn:
This ship has sailed a bit, but I do find the "base job" verbiage to be a bit odd. Base implies a relationship to the *other* jobs in the code location, which is not true. `get_implicit_asset_job` makes more sense to me

sryza:
I don't think the ship has sailed – this isn't a public API. Feel free to change.

schrockn:
will do so in followup

PR #11227, opened 2022-12-19
Product tour component

salazarm:
### Summary & Motivation
https://linear.app/elementl/issue/DSN-61/create-a-new-product-tour-component
https://www.figma.com/file/AhZwBSolIwuvHwg5gIRELn/Dagster-UI-Component-Library?node-id=1770%3A4556&t=CTeOXy6zgICatHRu-0

https://www.loom.com/share/c1085fa95d164dc6af73c1a2775cfd36


js_modules/dagit/packages/ui/src/components/ProductTour.tsx
```diff
@@ -0,0 +1,230 @@
...
+  BOTTOM_LEFT,
+  BOTTOM_CENTER,
+  BOTTOM_RIGHT,
+}
+
+type ObjectType =
+  | {img: string; video?: undefined; object?: undefined}
+  | {video: string; img?: undefined; object?: undefined}
+  | {object: React.ReactNode; video?: undefined; img?: undefined}
+  | {img?: undefined; video?: undefined; object?: undefined};
```
hellendag, line 23: This is kind of a confusing object type. Could it be a tagged union, and you refine on the `type` field?

```
type ObjectType =
| {type: 'image', img: string;}
| {type: 'video', video: string;}
| {type: 'object', object: React.ReactNode;}
| {type: 'empty'}
```

js_modules/dagit/packages/ui/src/components/ProductTour.tsx
```diff
@@ -0,0 +1,230 @@
...
+type Props = {
+  title: React.ReactNode;
+  description?: React.ReactNode;
+  actions?: {
+    next?: () => void;
+    dismiss?: () => void;
+  };
+  position: Position;
+  width?: CSSProperties['width'];
+} & ObjectType;
```
hellendag, line 34: Do the `ObjectType` fields need to be intersected onto `Props`? Or could the image/video/object be passed in as a single prop of `ObjectType`?

js_modules/dagit/packages/ui/src/components/ProductTour.tsx
```diff
@@ -0,0 +1,230 @@
...
+  | {img: string; video?: undefined; object?: undefined}
+  | {video: string; img?: undefined; object?: undefined}
+  | {object: React.ReactNode; video?: undefined; img?: undefined}
+  | {img?: undefined; video?: undefined; object?: undefined};
+
+type Props = {
+  title: React.ReactNode;
+  description?: React.ReactNode;
+  actions?: {
+    next?: () => void;
```
hellendag, line 29: Any need for `prev`?

hellendag (commented):
If the `ProductTour` component will wrap its target component, that looks pretty similar to the `Popover` (and siblings) API. Can this use a reskinned `Popover` (either from `@dagster-io/ui` or from `@blueprintjs`) so that you don't need to replicate positioning/portaling logic that it might handle instead?

Or alternatively, skip past `ui` and `blueprint` and use https://popper.js.org/react-popper/v2/react-portals/ directly -- blueprint uses Popper for this stuff, though I'm not sure if it uses `react-popper`.

js_modules/dagit/packages/ui/src/components/ProductTour.tsx
salazarm: You would have to pass type in as a prop and I think thats awkward from a user perspective.

From a user perspective this type definition is the friendliest imo since the call site doesnt have to do anything special.

js_modules/dagit/packages/ui/src/components/ProductTour.tsx
salazarm: The name of the field is how I determine which type it is image/video/object.

Otherwise you'd need to pass the type in some other way, I think using the name of the prop is the cleanest from an API perspective.

js_modules/dagit/packages/ui/src/components/ProductTour.tsx
salazarm: I don't think so right now, it's flexible to let you add that if you need it though.

salazarm:
> If the ProductTour component will wrap its target component, that looks pretty similar to the Popover (and siblings) API. Can this use a reskinned Popover (either from @dagster-io/ui or from @blueprintjs) so that you don't need to replicate positioning/portaling logic that it might handle instead?

> Or alternatively, skip past ui and blueprint and use https://popper.js.org/react-popper/v2/react-portals/ directly -- blueprint uses Popper for this stuff, though I'm not sure if it uses react-popper.

I don't mind switching to some other library. I actually wasn't sure if the existing popover supported the same functionality, but also we want to move away from blueprint and thats what it uses right? Popper seems like a good replacement.

js_modules/dagit/packages/ui/src/components/Popover.tsx
```diff
@@ -67,17 +70,18 @@ export const GlobalPopoverStyle = createGlobalStyle`
 const overwriteMerge = (destination: any[], source: any[]) => source;
 
 export const Popover: React.FC<Popover2Props> = (props) => {
+  const modifiers = deepmerge(
+    {offset: {enabled: true, options: {offset: [0, 8]}}},
+    props.modifiers || {},
+    {arrayMerge: overwriteMerge},
+  );
```
hellendag, line 77: Could probably put this back where it was, to limit the changed lines 🤷

js_modules/dagit/packages/ui/src/components/ProductTour.tsx
hellendag: Well, I was thinking it would be a prop like `thing={{type: 'image', img: 'https://...'}}`.

I don't feel that strongly about it, it just looked a little unusual to me. :)

js_modules/dagit/packages/ui/src/components/ProductTour.tsx
```diff
@@ -0,0 +1,128 @@
...
+      isOpen={true}
+      placement={position as Placement}
+      modifiers={{
+        arrow: {enabled: true},
+        preventOverflow: {enabled: true},
+      }}
+      minimal={false}
+      content={
+        <>
+          <div />
```
hellendag, line 79: Hm why the extra `<div />`s?

js_modules/dagit/packages/ui/src/components/Popover.stories.tsx
```diff
@@ -20,6 +20,7 @@ export const Default = () => {
       <Group direction="column" spacing={8} padding={8}>
         <Popover
           position="bottom-left"
+          modifiers={{arrow: {enabled: true}}}
```
hellendag, line 23: Remove?

hellendag (approved):
Thanks for changing to use `Popover`! 🎉

js_modules/dagit/packages/ui/src/components/ProductTour.tsx
salazarm: There was some css on :first-child and :last-child I was trying to avoid

js_modules/dagit/packages/ui/src/components/ProductTour.tsx
salazarm: Yeah I don't like having to pass a type explicitly. If we were to pass the type explicitly there would be no need to use different keys for image/video/object also. I'm essentially allowing the key you use to encode the type.

PR #11233, opened 2022-12-19
Use FromSourceAsset instead of FromRootInputManager when loading assets with input managers

jamiedemaria:
### Summary & Motivation

Flips the order with which we determine if an upstream dependency should be loaded via FromSourceAsset or FromRootInputManager so that if we are loading an upstream asset with an input manager we use a FromSourceAsset rather an a FromRootInputManager. This also makes FromSourceAsset support input managers.

should resolve https://elementl-workspace.slack.com/archives/C03VBNRRW4F/p1670935727868509


jamiedemaria:
* master
  * **PR #11233** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11233" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11233?utm_source=stack-comment).

python_modules/dagster/dagster_tests/storage_tests/test_input_manager.py
```diff
@@ -379,6 +383,43 @@ def check_input_managers():
...
+    class MyIOManager(IOManager):
+        def load_input(self, context):
+            assert context.upstream_output is not None
+
+            return 2
+
+        def handle_output(self, context, obj):
+            ...
+
+    group = AssetGroup(
```
sryza, line 404: AssetGroups are a deprecated concept. Instead, you can call `materialize` on a list of assets.

python_modules/dagster/dagster_tests/storage_tests/test_input_manager.py
```diff
@@ -379,6 +383,43 @@ def check_input_managers():
...
+    def upstream() -> int:
+        return 1
+
+    @asset(ins={"upstream": AssetIn(input_manager_key="special_io_manager")})
+    def downstream(upstream) -> int:
+        return upstream + 1
+
+    class MyIOManager(IOManager):
+        def load_input(self, context):
+            assert context.upstream_output is not None
```
sryza, line 397: I think it would also be worthwhile to assert that the right asset key is on the context

sryza (approved):
A couple small comments - otherwise, this looks great

python_modules/dagster/dagster_tests/storage_tests/test_input_manager.py
jamiedemaria: I tried that but it didn't work the way i needed it to. Here's the old code snippet 
```
def test_loading_already_materialized_asset():
    with instance_for_test() as instance:
        # materialize the upstream
        materialize([upstream], instance=instance)

        # materialize just the downstream
        materialize(
            [downstream],
            resources={
                "special_io_manager": IOManagerDefinition.hardcoded_io_manager(MyIOManager())
            },
            instance=instance,
        )
```

but that would give an error when materializing `downstream` that the dependency `upstream` didn't exist. Putting both assets in a group and then materializing via `selection` was the only way i could replicate the bug. if there's a better way to do this, i'm happy to change it

sryza: You can do 

```python
materialize(
            [*upstream_asset.to_source_assets(), downstream],
```

PR #11234, opened 2022-12-19
[dagit] Fresh + Stale should not show “Stale” on the Asset Graph

bengotow:
### Summary & Motivation

This PR makes a few small changes based on this week's UI sync:

- Assets that are Stale but are meeting their freshness policy do not show "Stale" on the graph, instead they show "Materialized"

- The word "On time" has been removed and we just show a freshness check icon you can see more info about on hover.

- I've added a storybook for seeing AssetNodes in all the various states because there are a lot of them and the ordering is fairly difficult to get right. 

New storybook!

image


hellendag (approved):
yay stories

PR #11252, opened 2022-12-20
refactor: sequester job backfill code

sryza:
branch-name: split-backfills

### Summary & Motivation

- Moves the code that's related to job backfills but not pure asset backfills into its own file.
- Rearranges the attributes in `PartitionBackfill` so that the ones only needed by job-based backfills are at the bottom.

This shouldn't change any behavior. It paves the way for adding pure asset backfills in https://github.com/dagster-io/dagster/pull/11099/files.

No behavior changes, so make sure existing tests pass. Also, with changes higher up in the stack, manually ran backfills.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/implementation/execution/backfill.py
+++ b/python_modules/dagster-graphql/dagster_graphql/implementation/execution/backfill.py
@@ -3,11 +3,8 @@ import pendulum
 import dagster._check as check
 from dagster._core.errors import DagsterError
 from dagster._core.events import AssetKey
-from dagster._core.execution.backfill import (
-    BulkActionStatus,
-    PartitionBackfill,
-    submit_backfill_runs,
-)
+from dagster._core.execution.backfill import BulkActionStatus, PartitionBackfill
+from dagster._core.execution.job_backfill import submit_backfill_runs
 from dagster._core.host_representation import RepositorySelector
 from dagster._core.utils import make_new_backfill_id
 
--- a/python_modules/dagster/dagster/_cli/job.py
+++ b/python_modules/dagster/dagster/_cli/job.py
@@ -27,11 +27,8 @@ from dagster._cli.workspace.cli_target import (
 from dagster._core.definitions.pipeline_base import IPipeline
 from dagster._core.errors import DagsterBackfillFailedError, DagsterInvariantViolationError
 from dagster._core.execution.api import create_execution_plan
-from dagster._core.execution.backfill import (
-    BulkActionStatus,
-    PartitionBackfill,
-    create_backfill_run,
-)
+from dagster._core.execution.backfill import BulkActionStatus, PartitionBackfill
+from dagster._core.execution.job_backfill import create_backfill_run
 from dagster._core.host_representation import (
     ExternalPipeline,
     ExternalRepository,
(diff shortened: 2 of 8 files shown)


prha:
Did you consider leaving `PartitionBackfill` alone and creating a new type of bulk object?  We could even rename `PartitionBackfill` to be `JobBasedPartitionBackfill` and create a new object type instead of trying to make it generically handle things.

prha:
More context: I think @johannkm was thinking about adding bulk deletes a while back, which we would also store in the bulk actions table.  It would have created a new `BulkActionType` and stored a new serialized named tuple in the `body` column.  I think the schema migrations are already done to support these new action types.

Not sure what you were thinking about keeping the daemons that actually perform the backfills separate or the same across the two backfill types, but this would allow you to query them separately, and then you're not burdened with maintaining the same generic data type.

sryza:
> Did you consider leaving PartitionBackfill alone and creating a new type of bulk object? We could even rename PartitionBackfill to be JobBasedPartitionBackfill and create a new object type instead of trying to make it generically handle things.

I went with this direction basically because it seemed like it would be net more lines of code to go the other direction: the asset backfills and the job backfills share a bunch of fields and operations, e.g. backfill_id, status, from_failure, tags, backfill_timestamp, error, asset_selection, with_status, with_error, get_backfills, add_backfill, update_backfill, etc.

My understanding coming out of the conversation we had last Thursday was that we would use a single daemon.

I had an earlier, cleaner version of the diff that had `JobBackfill` and `AssetBackfill` both extend from the same base class. But I had too much trouble getting inheritance to work with NamedTuples.

Another option could be to replace PartitionBackfill with a generic BulkAction type, which could have a bulk_action_data field that can have contents like JobBackfillData or AssetBackfillData. There would need to be some custom-loading code to handle backcompat.

The part of me that is eager to get asset backfills out the door wants to stick with this, but if you're saying that:
- Separating them out will make it easier to support directions that we want to go soon - e.g. multiple daemons or other bulk actions
- All the casts that this approach requires inside job_backfill.py are too ugly

I am open to changing. Thoughts?

prha:
If you think it's a strong possibility that we will want to split the work up so to be handled by two separate daemons, we should probably have two separate action types.  

If you want them to share the same data class, they can share the partition backfill namedtuple, but if you decide that they eventually diverge enough that you want the freedom to switch it out, we'll have to have some sort of backcompat logic.

The storage methods `add_backfill` / `update_backfill` will need to be supplemented (maybe replaced) with `add_bulk_action` / `update_bulk_action`, and would be used for all types of bulk actions we would take.  I think this would be orthogonal to the data class that's serialized into the body.

sryza:
> If you think it's a strong possibility that we will want to split the work up so to be handled by two separate daemons, we should probably have two separate action types.

👍 👍 . I added that to this downstream PR: https://github.com/dagster-io/dagster/pull/11378.

prha (approved):
This looks like it's mostly just 1) moving code to a different file, 2) rearranging namedtuple attrs 3) adding keyword args at the callsites to accommodate 2?

Might want to flesh out the commit message to talk through the overall plan, add a test 
section.

sryza:
@prha 👍 I made some updates to the PR description.

PR #11268, opened 2022-12-20
[structured config] Base structured config implementation

benpankow:
## Summary

Introduces the capability to specify config types for an Op or Asset by applying a type annotation to the new `config` keyword, e.g.

```python
@op
def greet_user(config: str):
    print(f"Hello, {config}")
```

For more complex config schema, users can build a Pydantic type ("structured config") extending the new `Config` base class. This provides type hinting when using config in an asset or op body:

```python
class GreetingConfig(Config):
    name: str
    should_print: bool

@op 
def greet_user(config: GreetingConfig):
    if config.should_print:
        print(f"Hello, {config.name}")
    else:
        ...
```

## Limitations

This PR only handles the case of structured config which has primitive types. More complex config (defaults, descriptions, Permissive, Selector/Unions, etc) will be handled in stacked diffs.

## Test Plan

Introduced a set of unit tests to cover basic config classes using primitives + nested structured config.

### DIFF
--- a/python_modules/dagster/dagster/_config/structured_config.py
+++ b/python_modules/dagster/dagster/_config/structured_config.py
@@ -1,4 +1,4 @@
-from typing import Type
+from typing import Any, Type
 
 from pydantic import BaseModel
 from pydantic.fields import SHAPE_SINGLETON, ModelField
@@ -29,7 +29,7 @@ def _convert_pydantic_field(pydantic_field: ModelField) -> Field:
     return convert_potential_field(dagster_type)
 
 
-def infer_schema_from_config_annotation(model_cls: Type) -> Field:
+def infer_schema_from_config_annotation(model_cls: Any) -> Field:
     """
     Parses a structured config class or primitive type and returns a corresponding Dagster config Field.
     """
@@ -40,12 +40,11 @@ def infer_schema_from_config_annotation(model_cls: Type) -> Field:
     return convert_potential_field(model_cls)
 
 
-def _safe_is_subclass(cls: Type, possible_parent_cls: Type) -> bool:
-    """Safe version of issubclass that returns False if cls is not a class."""
-    try:
-        return issubclass(cls, possible_parent_cls)
-    except TypeError:
+def _safe_is_subclass(cls: Any, possible_parent_cls: Type) -> bool:
+    """Version of issubclass that returns False if cls is not a Type."""
+    if not isinstance(cls, Type):
         return False
+    return issubclass(cls, possible_parent_cls)
 
 
 def infer_schema_from_config_class(model_cls: Type[Config]) -> Field:
--- a/python_modules/dagster/dagster/_config/structured_config.py
+++ b/python_modules/dagster/dagster/_config/structured_config.py
@@ -19,7 +19,7 @@ def _convert_pydantic_field(pydantic_field: ModelField) -> Field:
     Transforms a Pydantic field into a corresponding Dagster config field.
     """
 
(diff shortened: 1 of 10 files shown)


schrockn (dismissed):
See inline comments

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -0,0 +1,65 @@
...
+
+    return convert_potential_field(model_cls)
+
+
+def _safe_is_subclass(cls: Type, possible_parent_cls: Type) -> bool:
+    """Safe version of issubclass that returns False if cls is not a class."""
+    try:
+        return issubclass(cls, possible_parent_cls)
+    except TypeError:
+        return False
```
schrockn, line 48: This is pretty odd.

1) If the type annotations are to be trusted, issubclass should always succeed.
2) Even though some say it is "Pythonic" I think this is a dangerous coding pattern, as it is difficult to know the universe of reasons why something will throw an error.

I suggest instead checking `isinstance(cls, Type)` and returning False in that case.

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def infer_schema_from_config_class(model_cls: Type[Config]) -> Field:
+    """
+    Parses a structured config class and returns a corresponding Dagster config Field.
+    """
+
```
schrockn, line 55: you could also do an `inst` check here

python_modules/dagster/tox.ini
```diff
@@ -42,4 +42,4 @@ commands =
 
 [testenv:pylint]
 commands =
-  pylint -j0 --rcfile=../../pyproject.toml {posargs} dagster dagster_tests
+  pylint -j0 --rcfile=../../pyproject.toml {posargs} dagster dagster_tests --extension-pkg-whitelist='pydantic'
```
schrockn, line 45: can you comment as to why this is necessary

schrockn (changes requested):

python_modules/dagster/dagster/_config/structured_config.py
schrockn: `check.inst_param(cls, Type)` to be precise

python_modules/dagster/dagster/_config/structured_config.py
benpankow: Will gate with an `isinstance(cls, Type)` check, as suggested - "issubclass should always succeed" is true if `cls` is a class, but in this case we are occasionally dealing with non-class objects (e.g. `typing` annotations, `issubclass(List[str], Config)` raises a `TypeError`).

python_modules/dagster/tox.ini
benpankow: This is because of a pylint issue, this flag [allows pylint to recognize various top-level imports](https://github.com/pydantic/pydantic/issues/1961#issuecomment-729288794) from the `pylint` module. Added a comment in the `.ini` file as well.

schrockn (approved):
Awesome

benpankow:
**Graphite Merge Job**

Current status: 🔄 Waiting for checks to pass

This pull request is currently waiting for CI and mergeability checks to pass before being merged.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11268?utm_source=stack-comment).


Job Reference: TXP57Jx30eGCn07R3Kfh

PR #11279, opened 2022-12-20
Add get_implicit_global_asset_job on Definitions and RepositoryDefinition

schrockn:
### Summary & Motivation

It is often useful in test cases to easily access the singleton asset job object that exists when there is only one partitioning scheme in a code location.

This also implies a change from "base job" language to "implicit job" language. I think "implicit" job is a better name since "base" can imply a relationship between them and defined jobs, since it sounds like "base classes" in OOP-speak.

BK


schrockn:
* master
  * **PR #11279** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11279" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11279?utm_source=stack-comment).

python_modules/dagster/dagster_tests/definitions_tests/test_definitions_class.py
```diff
@@ -431,7 +431,28 @@ def asset_two(context):
...
+
+
+def test_implicit_global_job_with_job_defined():
+    @asset
+    def asset_one():
+        pass
+
+    defs = Definitions(assets=[asset_one], jobs=[define_asset_job("all_assets_job", selection="*")])
+
+    # I officially do not understand the semantics of the global implicit job at this point
```
schrockn, line 457: @sryza I officially don't understand the behavior of this. When is there no implicit global job?

sryza (commented):
"implicit" sounds like a good name to me. If we're going to make this change, I think we should remove "base" entirely rather than having two names coexist for the same thing

schrockn:
More than happy to codemod after this lands.

sryza (approved):

PR #11285, opened 2022-12-20
[dagit] Utility for timezone-aware date/time formatting

hellendag:
### Summary & Motivation

Create a utility hook for generating formatted date/time strings with awareness of the user's timezone.

The hook returns a memoized function that accepts a date and an `Intl.DateTimeOptions` object, and produces a formatted string with the user's timezone applied to it.

I wrote a test and updated `RunTimeline` to make use of it. Most of the use cases look like they'll be in Cloud.

Jest, load Dagit and view run timeline. Verify that the date and time labels look correct.

### DIFF
--- a/js_modules/dagit/packages/core/src/ui/useFormatDateTime.test.tsx
+++ b/js_modules/dagit/packages/core/src/ui/useFormatDateTime.test.tsx
@@ -1,0 +1,134 @@
+import {render, screen} from '@testing-library/react';
+import * as React from 'react';
+
+import {TimezoneProvider, TimezoneStorageKey} from '../app/time/TimezoneContext';
+
+import {useFormatDateTime} from './useFormatDateTime';
+
+jest.mock('../app/time/browserTimezone', () => ({
+  browserTimezone: jest.fn(() => 'Pacific/Honolulu'),
+}));
+
+const NYD_2023_CST = 1672552800000;
+
+describe('useFormatDateTime', () => {
+  interface Props {
+    date: Date;
+    options: Intl.DateTimeFormatOptions;
+  }
+
+  const Test: React.FC<Props> = ({date, options}) => {
+    const formatDateTime = useFormatDateTime();
+    return <div>{formatDateTime(date, options)}</div>;
+  };
+
+  afterEach(() => {
+    window.localStorage.removeItem(TimezoneStorageKey);
+  });
+
+  describe('Automatic timezone', () => {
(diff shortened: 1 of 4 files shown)


hellendag:
* master
  * **PR #11285** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11285" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11285?utm_source=stack-comment).

js_modules/dagit/packages/core/src/ui/useFormatDateTime.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+    const formatDateTime = useFormatDateTime();
+    return <div>{formatDateTime(date, options)}</div>;
+  };
+
+  afterEach(() => {
+    window.localStorage.removeItem(TimezoneStorageKey);
+  });
+
+  describe('Automatic timezone', () => {
+    // Timezone has not been set in localStorage, so it should be an `Automatic` timezone, and
```
salazarm, line 30: Why do we set the Timezone in localStorage? I'd think we should just always use automatic?

hellendag:
> Approved but unsure about saving timezone in localStorage. Won't this fall out of date easily? Like if I'm flying around from east/west coast will it keep using the east coast timezone when I'm on the west coast? How is this cache invalidated?

Fair questions! The user can set their timezone manually in User Settings.

- If the user keeps "Automatic" (default), then it will use the client runtime's default time zone, as evaluated by `Intl.DateTimeFormat().resolvedOptions().timeZone`. So if you change timezones flying between coasts, and your OS updates to the new time zone, this value will update accordingly and Dagit will show your new time zone.
  - I don't have data on this, but I assume this is what the vast majority of users have.
- The user can alternatively change to have Dagit show a different time zone, perhaps based on some static location, e.g. to keep it the same even when traveling, or to keep it on UTC/GMT. If they choose to change timezones away from "Automatic", we track the value in localStorage so that it remains static.
  - I don't have context on why this feature was initially added, but we can look back in time to find out.

Right now we don't actually do anything to delete the localStorage item. By default, it's undefined, which is evaluated as "Automatic".

If the user selects "Automatic" manually, then we track that string as the stored value, same as other timezone options. We could delete the item instead when choosing "Automatic" -- no objection to this from me -- but the end result is the same.

PR #11288, opened 2022-12-20
[dagit] Show observations about the latest materialization on Asset > Partitions

bengotow:
### Summary & Motivation

This PR addresses a regression in behavior during the asset redesign. Previously, viewing "asset events by partition" would collapse the last materialization and observations about that materialization, showing a nice summary of all the metadata.

This was lost in the new Asset Partitions tab, both because we weren't asking for the observations and because the partitions filtering had been commented out in the GraphQL resolver. (cc @smackesey on this bit, is it safe to uncomment this? https://github.com/dagster-io/dagster/commit/1f2d7465630c92c65b7be75441e4393d603abe67#diff-4d4316ce4b86a406d0b67adb85ef8cfdfc9d976d3d36887c68c2d29e89da8403R434)

I tested these changes using the test below, which materializes asset1, and then has asset2 leave an observation on asset1 with additional metadata. Viewing the Asset Partitions page should show the last materialization as well as the observation data for any observations after that materialization. I tested that only subsequent observations are rolled up into the table by changing the keys in the observation and re-materializing.

image

```
import time
from random import random

from dagster import (
    repository,
    AssetObservation,
    MetadataValue,
    Output,
    asset,
    MonthlyPartitionsDefinition,
)

monthly_partitions = MonthlyPartitionsDefinition(end_offset=1, start_date="2020-01-01")


@asset(
    partitions_def=monthly_partitions,
    group_name="observed",
    compute_kind="duckdb",
    op_tags={"compute_kind": "duckdb"},
)
def asset1():
    return 1


@asset(
    partitions_def=monthly_partitions, group_name="observed", description="This is my asset desc"
)
def asset2(context, asset1):
    time.sleep(25)
    yield AssetObservation(
        "asset1",
        partition=context.partition_key,
        metadata={
            "table name": "users_table",
            "row count 2": (random()),
            "custom value a": (random()),
            "anotha one": (random()),
            "data url": MetadataValue.url("http://mysite/users_table"),
        },
    )
    yield Output(value=1, output_name="result", metadata={"num_rows": 50})


@repository
def repo():
    return [asset1, asset2]
```


js_modules/dagit/packages/core/src/assets/AssetPartitionDetail.tsx
```diff
@@ -38,21 +38,31 @@ export const AssetPartitionDetailLoader: React.FC<{assetKey: AssetKey; partition
...
   }
 
-  const events =
+  const materializations =
     result.data?.assetNodeOrError?.__typename === 'AssetNode'
-      ? result.data.assetNodeOrError.assetMaterializations
+      ? [...result.data.assetNodeOrError.assetMaterializations].sort(
+          (a, b) => Number(b.timestamp) - Number(a.timestamp),
+        )
       : [];
```
hellendag, line 46: Worth doing `useMemo` on these?

js_modules/dagit/packages/core/src/assets/AssetPartitionDetail.tsx
```diff
@@ -97,6 +114,8 @@ export const AssetPartitionDetail: React.FC<{
         )
       : [];
 
+  console.log(observationsAboutLatest);
```
hellendag, line 117:
```suggestion
```

hellendag (approved):

smackesey:
>is it safe to uncomment this? https://github.com/dagster-io/dagster/commit/1f2d7465630c92c65b7be75441e4393d603abe67#diff-4d4316ce4b86a406d0b67adb85ef8cfdfc9d976d3d36887c68c2d29e89da8403R434

I think so

PR #11290, opened 2022-12-20
Make it more likely that we hit our lock

jmsanders:
We started seeing flakes on our race condition tests as soon as we merged. I wasn't able to reproduce locally but I suspect .01 doesn't give us enough margin for error to hit our lock.

I'm bumping this a bit to see if it helps.


jmsanders:
I'm going to retry this a few times before landing.

alangenfeld:
probably want to run this repeatedly on 3.7 specifically

jmsanders:
This succeeded 5 times in a row on 3.7.

gibsondan (approved):

PR #11291, opened 2022-12-20
[dagstermill] add retries to flaky tests

jamiedemaria:
### Summary & Motivation
The dagstermill tests suite is flaky due to the Jupyter kernel mysteriously dying. This PR makes a couple of changes that should make the test suite more resilient 

1. create the jupyter kernel for each test rather than the entire session. Running the kernel install command when the kernel already exists is a no-op, so we can rerun the command before each test to ensure that the kernel is alive when the test starts
2. add retries to the dagstermill test suite (and mark the dagstermill docs example tests as flaky so they get retried). if the kernel dies midway through a test, this should rerun the test and hopefully the kernel won't die again.

gonna keep an eye on the dagstermill tests for the next few weeks and see if this helps

### DIFF
--- a/examples/docs_snippets/docs_snippets_tests/integrations_tests/test_dagstermill.py
+++ b/examples/docs_snippets/docs_snippets_tests/integrations_tests/test_dagstermill.py
@@ -9,7 +9,6 @@ from dagster._legacy import execute_pipeline
 
 IS_BUILDKITE = os.getenv("BUILDKITE") is not None
 
-import re
 import subprocess
 import warnings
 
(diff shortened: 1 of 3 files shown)


jamiedemaria:
* master
  * **PR #11291** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11291" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11291?utm_source=stack-comment).

examples/docs_snippets/docs_snippets_tests/integrations_tests/test_dagstermill.py
```diff
@@ -63,6 +49,7 @@ def exec_for_test(module_name, fn_name, env=None, raise_on_error=True, **kwargs)
                 cleanup_result_notebook(result)
 
 
+@pytest.mark.flaky(reruns=1)
```
jamiedemaria, line 52: adding the flaky mark individually here so we don't have bk retry every test in the docs snippets suite

jmsanders (approved):

PR #11294, opened 2022-12-20
[declarative-scheduling] Update retry logic to attempt to retry failed materializations after some time has passed

OwenKephart:
### Summary & Motivation

Once an asset fails to be materialized in a given run, when do we try it again?

On the next tick after the failure would cause a never-ending spew of runs for assets if a bad code change was made, which is undesirable.

If we never tried failed assets again, manual intervention would be required to start things up again, which is also undesirable.

This PR updates the logic to ignore constraints that would have been solved had failed runs succeeded. 

This has the following arguably nice properties:

* If an asset that is upstream of a freshness policy with maximum_lag_minutes = 60 fails, then that asset will be retried approximately 60 minutes after the failure
* If the failed asset is upstream of two freshness policies, one 30 minutes and the other 60 minutes, then it will only wait around 30 minutes before retrying
* If a run fails, no more runs will be kicked off compared to if it had succeeded (although fewer runs may be kicked off as subsequent assets will realize that the data they need is not available, and therefore have no reason to run.


mjclarke94:
That logic seems like a nice way of getting the initial retry time, but still potentially has the issue you mentioned of firing off a large number of runs for broken code/trying to run whilst an external service is down.

Could it be worth tying this in with an (optional?) exponential backoff strategy so that the gaps between retries become successively larger with each failed run? That way you still don't need manual intervention to get a passing run for issues which will resolve themselves (long maintenance windows etc) but the amount of resources spent on broken code is minimised.

OwenKephart:
Hi @mjclarke94  ! Definitely a reasonable idea, and I appreciate the question! To pull back the curtain on some of my thinking not discussed in the description, I think there's an inherent tension in this feature space between what sort of retry logic should exist at this level vs the underlying operational level.

Ops support a [RetryPolicy](https://docs.dagster.io/_apidocs/ops#dagster.RetryPolicy), and so by extension assets, which contain ops, support them  (it's not part of the asset decorator at the moment, but it definitely should be). We also have [run retries](https://docs.dagster.io/deployment/run-retries#run-retries), which exists in this same sort of space.

Conceptually, complex retry logic might fit more naturally into that sort of level, where you can configure, per-op or per-run, if and how you'd like that logic to be retried if it happens to fail. That way, the retry policy is specific to the actual operation you might want to retry -- some things, if they fail, are very unlikely to succeed on subsequent attempts unless their code is changed (a bug in the logic), while others are extremely likely to succeed immediately after a failure (like a flaky API), and others (like the case you mention with planned downtime) will probably not succeed for an indeterminate amount of times, but then start working again. The place where you're most likely to be able to understand which of these categories you fit into is the op level in most cases.

However, there's also a case to be made that some generic backoff type of strategy is pretty universally applicable, and it's definitely worth considering. I don't plan to include that functionality in this PR but more customizability around retries here is definitely something we'd like to support.

sryza (commented):
If someone has `FreshnessPolicy(maximum_lag_minutes=5, cron_schedule="@daily")`, would this retry every five minutes? That might be undesirable?

sryza:
Or maybe that only applied to the only version and not the new one?

OwenKephart:
@sryza yeah that was only true for the old version -- just updated the description

mjclarke94:
@OwenKephart Thanks for clarifying. I agree, it makes a lot more sense at the op-scope in general.

A common pattern I used to use with another orchestrator (Aiida) was trying to spawn shells/fetch files from a HPC cluster. It was a pretty regular occurrence that the login nodes would be down, or a filesystem wouldn't be mounted during a maintenance window. That orchestrator had a per-task backoff approach, but that still meant you'd have thousands of tasks all trying to run with the same single point of failure.

Given the situation I described probably would kick in only when an external service was down, it might make sense to have some similar functionality at the resource scope as well? That way, after some critical mass of specific failure modes (connection refused, that sort of thing), you could delay running any jobs which rely on the flaky resource. In other words, you don't need each individual asset to try and fail to use the resource, the information that it was down would be passed through to it through the fact that the resource had been backed off.

That said, those sort of workflows were a lot more ad-hoc and used extremely janky HPC systems, so maybe I'm just projecting issues from another field on to this one!

sryza (approved):

PR #11300, opened 2022-12-21
Pass duckdb_path to __init__ rather than relying on context

schrockn:
### Summary & Motivation

In general we should not be relying on the context for values that are known, immutable, and typed at init time. Passing these directly to the initializer is more obvious and eases refactoring later down the line. More of the code ends up being not dependent on the context object and is more reusable.

BK


schrockn:
* master
  * **PR #11300** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11300" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11300?utm_source=stack-comment).

jamiedemaria (approved):

PR #11301, opened 2022-12-21
Hoist schema and database to DbIOManager constructor

schrockn:
### Summary & Motivation

Similar to https://github.com/dagster-io/dagster/pull/11300. We should be grabbing config as early as possible and putting it into vanilla python. This makes code more safe as well as easier to understand and refactor.

BK

`cd docs && make apidoc-build`

Screen Shot 2022-12-21 at 4 47 21 PM

Screen Shot 2022-12-21 at 4 47 32 PM

### DIFF
--- a/python_modules/libraries/dagster-duckdb/dagster_duckdb/io_manager.py
+++ b/python_modules/libraries/dagster-duckdb/dagster_duckdb/io_manager.py
@@ -83,7 +83,7 @@ def build_duckdb_io_manager(type_handlers: Sequence[DbTypeHandler]) -> IOManager
             ),
         }
     )
-    def duckdb_io_manager(_):
+    def duckdb_io_manager(init_context):
         """IO Manager for storing outputs in a DuckDB database
 
         Assets will be stored in the schema and table name specified by their AssetKey.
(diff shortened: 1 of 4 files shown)


schrockn:
* master
  * **PR #11300** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11300" target="_blank">Graphite</a> 
    * **PR #11301** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11301" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11301?utm_source=stack-comment).

python_modules/libraries/dagster-duckdb/dagster_duckdb/io_manager.py
```diff
@@ -83,7 +83,7 @@ def my_table_a(my_table: pd.DataFrame):
             ),
         }
     )
-    def duckdb_io_manager(_):
+    def duckdb_io_manager(init_context):
```
jamiedemaria, line 86: can you do a check that the api docs build still works with the new parameter? I think it should be fine, but iirc I put the _ in so that we could do the kinda hacky thing that lets the docstrings get picked up by sphinx. If it fails with the new parameter name, lmk and I can help debug, since sphinx is doing some kinda weird stuff here

jamiedemaria (commented):
just the one thing for docs but this looks good

schrockn:
Ran make apidoc-build and updated test plan

jamiedemaria (approved):
nice!!

PR #11306, opened 2022-12-21
Include parent exceptions in Dagster errors, even if they weren't explicitly raised

gibsondan:
Summary:
See https://learning-python.readthedocs.io/en/latest/chapter08/README.html - there are two types of chained exceptions, one explicit:

```
except Exception as e:
  raise ChildException() from e
```

and one implicit:
```
except:
  raise ChildException()
```

We were only handling and displaying hte former, which is in the __cause__ field - but the latter is stored in the __context__ field. Store and display both.

new bk test, new JS test - raise a nested exception in a job, see both types of error in Dagit error log


prha:
Are there scenarios where we explicitly wouldn't want to display the parent exception?

gibsondan:
I'm having trouble thinking of one other than it sometimes being more verbose - It's the default behavior when Python logs an uncaught exception

prha (approved):

PR #11307, opened 2022-12-21
chore: add method to strip error stack trace

rexledesma:
### Summary & Motivation
As the title.

N/A


rexledesma:
* master
  * **PR #11307** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11307" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11307?utm_source=stack-comment).

python_modules/dagster/dagster/_utils/error.py
```diff
@@ -38,6 +38,9 @@ def to_string(self):
         )
         return "{err.message}{stack}{cause}".format(err=self, stack=stack_str, cause=cause_str)
 
+    def strip_error_stack_trace(self) -> "SerializableErrorInfo":
+        return self._replace(stack=[], cause=None)
```
gibsondan, line 42: kind of a nit, but the cause=None part doesn't immediately follow for me from the name - is it more about just including the message?

Explicitly listing out the fields that should be included vs. removing the fields that shouldn't might be safer - e.g. I added the context field in https://github.com/dagster-io/dagster/pull/11306, would be easy to accidentally leave that included currently

```return SerializableErrorInfo(message=self.message, stack=[], cls_name=self.cls_name)?```

gibsondan (approved):

PR #11308, opened 2022-12-21
feat: add retry number and url to integration api call failure

rexledesma:
### Summary & Motivation
Something like this message is a little insufficient. Add the number of retries that occurred, as well as the specific url that was being requested.
```
dagster._core.errors.DagsterUserCodeUnreachableError: dagster._core.errors.DagsterUserCodeUnreachableError: Failure loading server endpoint for prod:purina:
dagster._core.definitions.events.Failure: Exceeded max number of retries.
```

TODO: pytest

### DIFF
--- a/python_modules/libraries/dagster-airbyte/dagster_airbyte/resources.py
+++ b/python_modules/libraries/dagster-airbyte/dagster_airbyte/resources.py
@@ -121,6 +121,7 @@ class AirbyteResource:
             Optional[Dict[str, Any]]: Parsed json data from the response to this request
         """
 
+        url = self.api_base_url + endpoint
         headers = {"accept": "application/json"}
 
         num_retries = 0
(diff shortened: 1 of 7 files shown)


rexledesma:
* master
  * **PR #11308** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11308" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11308?utm_source=stack-comment).

yuhan (commented):
code looks good. leaving it open as i'm curious how others think of the idea to consolidate them and making something like `DagsterResourceMaxRetriesExceededError(retry_num)` or something less mouthful?

OwenKephart:
Don't have much to add here that hasn't been said, but yeah I agree that it would be great to have some shared utilities for building integrations, as right now a ton of our libraries re-implement some basic `requests` functionality.

In general, we have a lot of code that's basically:

```
def do_something(self, param1, param2):
    self.make_request("some/endpoint", data={"param1": param1, "param2": param2})
```

at the very least, we could have a shared base class between these classes which implements `make_request`, including basic retry functionality

OwenKephart (approved):
approving because this is an improvement over status quo, and better abstractions shouldn't block this QoL upgrade

PR #11314, opened 2022-12-21
export product tour component

salazarm:


js_modules/dagit/packages/core/src/launchpad/LaunchpadHooksContext.tsx
```diff
@@ -14,16 +16,15 @@ export const LaunchpadHooksContext = React.createContext<LaunchpadHooksContextVa
...
 
 export function useLaunchPadHooks() {
-  const {LaunchRootExecutionButton: overrideLaunchRootExecutionButton} = React.useContext(
-    LaunchpadHooksContext,
-  );
-
-  const {useLaunchWithTelemetry: overrideUseLaunchWithTelemetry} = React.useContext(
-    LaunchpadHooksContext,
-  );
+  const {
```
salazarm, line 19: added this by accident

js_modules/dagit/packages/ui/CHANGES.md
```diff
@@ -1,3 +1,7 @@
+## 1.0.8 (December 21, 2022)
+
+- Added ProductTour component
```
hellendag, line 3: Can you see if there are any other notable changes since 1.0.7?

Also, do we need to publish a new version already? Cloud uses master `dagit-core` and `ui`, so we don't need a new version there.

js_modules/dagit/packages/ui/src/components/ProductTour.tsx
```diff
@@ -31,7 +31,7 @@ type Props = {
     next?: () => void;
     dismiss?: () => void;
   };
-  position: Position;
+  position: ProductTourPosition;
```
hellendag, line 34: Could `ProductTour` just use the same placement type/values as `Popover`?

salazarm: Yeah I was thinking about doing that but was a little torn about whether or not to use Popover because the arrow is always pointing at the center and we wanted to be able to have it point to the left/right as well as the center :\

salazarm: Lets make a follow up task to fix this

js_modules/dagit/packages/ui/CHANGES.md
salazarm: https://github.com/dagster-io/dagster/commits/31fe69fb2817cf50aa39e65bcc1fec0ac3a91cc1/js_modules/dagit/packages/ui

I don't think there's any other notable change.

salazarm: > Also, do we need to publish a new version already? Cloud uses master dagit-core and ui, so we don't need a new version there.

We don't. We should start the log now though so that when we do publish its easier for whoever publishes later. I can remove the date and put (TBD) instead? I didn't actually update the version in package.json

hellendag: Ah, got it. Yeah, I saw the date next to the version number and assumed you were intending to publish.

PR #11318, opened 2022-12-21
AssetGraph.from_external_assets().get_required_multi_asset_keys()

sryza:
### Summary & Motivation

Needed for partition-mapped asset backfills, which execute in a daemon.

### DIFF
--- a/python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_graph.py
+++ b/python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_graph.py
@@ -4,12 +4,20 @@ import pytest
 
 from dagster import (
     AssetIn,
+    AssetKey,
+    AssetOut,
+    AssetsDefinition,
     DailyPartitionsDefinition,
+    GraphOut,
     HourlyPartitionsDefinition,
     LastPartitionMapping,
+    Out,
     PartitionMapping,
     StaticPartitionsDefinition,
     asset,
+    graph,
+    multi_asset,
+    op,
     repository,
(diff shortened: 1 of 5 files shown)


python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_graph.py
```diff
@@ -218,3 +227,61 @@ def child(parent):
...
+    def op2():
+        return 4, 5
+
+    @graph(out={"a": GraphOut(), "b": GraphOut(), "c": GraphOut()})
+    def graph1():
+        a = op1()
+        b, c = op2()
+        return a, b, c
+
+    graph_backed_multi_asset = AssetsDefinition.from_graph(graph1)
```
OwenKephart, line 278: I think this logic might break down if you have two graph backed multi assets that derive from the same graph. It may all work out, but should be explicitly tested here.

In general, I'm not sure how much I trust those name fields to serve as a unique identifier for a node. You may get weird cases when you have multiple partitioned __ASSET_JOBS, which each reference copies of the same asset graph, and so therefore get the same graph_name or something like that.

An alternative, much safer, solution would be to include a "required_multi_asset_keys" field on the `ExternalAssetNode`. However, this is pretty inefficient storage-wise. An iteration on that would be to put an `md5hash(frozenset(required_multi_asset_keys))` as the unique identifier.

OwenKephart (changes requested):
see comment

OwenKephart (approved):
nice

PR #11321, opened 2022-12-21
[structured config] Structured-config-backed Resources

benpankow:
## Summary

Allows defining resources using structured config (#11268). This makes it easy to define a class-based resource which requires various configured inputs, with type hints etc.

```python
class WriterResource(StructuredConfigResource):
    prefix: str

    def output(self, text: str) -> None:
        output.append(f"{self.prefix}{text}")
```

By default, any resource extending `StructuredConfigResource` returns itself from the resource function, but this can be overwritten to wrap function-style resources:

```python
# function-style resource
@resource(config_schema={"connection": str})
def db_resource(init_context):
    ...

# structured config wrapper
class DbResource(StructuredConfigResource):
    connection: str

    def resource_fn(self, context) -> Any:
        return db_resource(context)
```

Along with #11322, this can make defining and using resources very lightweight:

```python
class WriterResource(StructuredConfigResource):
    prefix: str

    def output(self, text: str) -> None:
        output.append(f"{self.prefix}{text}")

@asset
def hello_world_asset(writer: WriterResource):
    writer.output("hello, world!")

defs = Definitions(
    assets=[hello_world_asset],
    resources={"writer": WriterResource(prefix="foo")}
)
```

## Test Plan

Small unit test, more forthcoming

### DIFF
--- a/python_modules/dagster/dagster/_config/structured_config.py
+++ b/python_modules/dagster/dagster/_config/structured_config.py
@@ -57,7 +57,9 @@ class Resource(
         ResourceDefinition.__init__(
             self,
             resource_fn=self.resource_function,
-            config_schema=configured_resource_def.config_schema,
+            # mypy worries that configured_resource_def could be self, which
+            # has not had config_schema initialized yet, but we know that's not the case
+            config_schema=configured_resource_def.config_schema,  # type: ignore[attr-defined]
             description=self.__doc__,
         )
 
--- a/python_modules/dagster/dagster/_config/structured_config.py
+++ b/python_modules/dagster/dagster/_config/structured_config.py
@@ -1,37 +1,18 @@
 import inspect
-from inspect import Parameter
-from typing import Any, Optional, Sequence, Type, TypeVar
+from typing import Any, Optional, Type
 
 from pydantic import BaseModel
 from pydantic.fields import SHAPE_SINGLETON, ModelField
-from typing_extensions import Annotated
 
 import dagster._check as check
 from dagster import Field, Shape
 from dagster._config.field_utils import FIELD_NO_DEFAULT_PROVIDED, convert_potential_field
-from dagster._core.decorator_utils import get_function_params
-from dagster._core.definitions.resource_definition import ResourceDefinition
 
 
 class Config(BaseModel):
     """
     Base class for Dagster configuration models.
     """
-
-
-def get_resource_args(fn) -> Sequence[Parameter]:
-
-    return [param for param in get_function_params(fn) if _is_resource_annotated(param)]
-
-
-def _is_resource_annotated(param: Parameter) -> bool:
-    return (
-        isinstance(param.annotation, type) and issubclass(param.annotation, ResourceDefinition)
-    ) or (hasattr(param.annotation, "__metadata__") and getattr(param.annotation, "__metadata__"))
-
-
-T = TypeVar("T")
-ResourceOutput = Annotated[T, "resource_output"]
 
 
 def _convert_pydantic_field(pydantic_field: ModelField) -> Field:
--- a/python_modules/dagster/dagster/_config/structured_config.py
+++ b/python_modules/dagster/dagster/_config/structured_config.py
@@ -1,5 +1,5 @@
 import inspect
-from functools import cached_property, lru_cache
+from functools import cached_property
 from typing import Any, Optional, Type
 
 from pydantic import BaseModel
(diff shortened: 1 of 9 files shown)


benpankow:
* master
  * **PR #11321** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11321" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11321?utm_source=stack-comment).

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -15,6 +20,46 @@ class Config(BaseModel):
     """
 
 
+class StructuredConfigResource(
```
schrockn, line 23: just call it `Resource`?

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -15,6 +20,46 @@ class Config(BaseModel):
...
+        # config schema. Pydantic will normally raise an error if you try to set an attribute
+        # that is not part of the schema.
+
+        if name.startswith("_"):
+            object.__setattr__(self, name, value)
+            return
+
+        return super().__setattr__(name, value)
+
+    def resource_fn(self, context) -> Any:  # pylint: disable=unused-argument
```
schrockn, line 57: rename `context` to `_` to avoid pylint pragma

benpankow: I was leaving it as `context` so the kwarg matches up in case a child class chooses to override `resource_fn`, e.g. a case where a user wants to wrap a function-style resource.

schrockn: ah I see. ok

schrockn: ah I see. ok

schrockn: ah I see. ok

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -15,6 +20,66 @@ class Config(BaseModel):
...
+        # This is a hack to allow us to set attributes on the class that are not part of the
+        # config schema. Pydantic will normally raise an error if you try to set an attribute
+        # that is not part of the schema.
+
+        if name.startswith("_"):
+            object.__setattr__(self, name, value)
+            return
+
+        return super().__setattr__(name, value)
+
```
schrockn, line 76: Can you add tests to see if `functools.cached_property` and our own `cached_method` works? I think that will be super useful pattern.

schrockn: Also add tests that test the core capability.

schrockn (changes requested):

python_modules/dagster/dagster/_config/structured_config.py
benpankow: Added some more tests, including testing the functionality of caching using `@cached_property` and `@cached_method`.

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -15,6 +21,88 @@ class Config(BaseModel):
     """
 
 
+class cached_method:
```
schrockn, line 24: was talking about re-using cached_method in python_modules/dagster/dagster/_utils/cached_method.py not a new one

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -15,6 +21,88 @@ class Config(BaseModel):
...
+
+            def output(self, text: str) -> None:
+                print(f"{self.prefix}{text}")
+
+    """
+
+    class Config:
+        arbitrary_types_allowed = True
+        keep_untouched = (cached_property, cached_method)
+        frozen = True
```
schrockn, line 66: is this supposed to be pydantic.Config?

If not this is super confusing and worth of comment.

schrockn (changes requested):

python_modules/dagster/dagster/_config/structured_config.py
schrockn: I see looks like Pydantic looks for a bare class called Config by convention and there is no way for us to avoid the name collison.

At a minimum there should be a comment describing this.

python_modules/dagster/dagster/_config/structured_config.py
benpankow: ah, yeah, realized this, the change was pushed into the IO manager PR by mistake. updated

schrockn (approved):

benpankow:
**Graphite Merge Job**

Current status: ⏳ Queued to merge

This pull request is currently queued to be merged as part of a stack.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11321?utm_source=stack-comment).


Job Reference: uHrT8DIrkchoQsyuEMKn

PR #11322, opened 2022-12-21
Allow specifying resources in Op/Asset params list

benpankow:
## Summary

Allows users to specify resource dependencies as parameters to an Asset or Op function.

Resource dependencies are identified by type annotations which either extend `ResourceDefinition` or are wrapped as `ResourceOutput[T]`. 

The first pattern is useful for "new"-style resources, which are Python classes that the user will interact with. In this case, the `ResourceDefinition` and the output from the resource function are identical. Here, a theoretical `S3Resource` implements `ResourceDefinition` and provides various interactions:

```python
@asset
def my_asset(s3: S3Resource):
    with s3.open_bucket(...):
        .... 
```

The second pattern is useful for function-style resources, which may output any arbitrary Python object. In this case, our `PostgresResource` outputs a `DatabaseConnection` object:

```python
@op
def my_op(postgres: ResourceOutput[DatabaseConnection]):
    postgres.query(...)
```

## Implementation

For the prototype, this approach was chosen over an approach which would use input names rather than annotations to identify which inputs are resources and which are other op/asset dependencies for a few reasons:

- Clarity: By using annotations, it is immediately clear to the reader which parameters are resources.
- Complexity: Asset dependencies are currently resolved immediately, when the decorator is applied. It would require a significant rewrite to delay this evaluation until the `Definitions` object containing the asset is evaluated.

## Test Plan

Unit tests

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
@@ -264,10 +264,13 @@ class _Asset:
         with warnings.catch_warnings():
             warnings.simplefilter("ignore", category=ExperimentalWarning)
 
-            required_resource_keys = (
-                set(self.required_resource_keys)
-                .union(set(self.resource_defs.keys()))
-                .union({arg.name for arg in get_resource_args(fn)})
+            arg_resource_keys = {arg.name for arg in get_resource_args(fn)}
+            decorator_resource_keys = set(self.required_resource_keys).union(
+                set(self.resource_defs.keys())
+            )
+            check.param_invariant(
+                len(decorator_resource_keys) == 0 or len(arg_resource_keys) == 0,
+                "Cannot specify resource requirements in both @asset decorator and as arguments to the decorated function",
             )
 
             if isinstance(self.io_manager, str):
@@ -295,7 +298,9 @@ class _Asset:
                 description=self.description,
                 ins=dict(asset_ins.values()),
                 out=out,
-                required_resource_keys=required_resource_keys,
+                # Any resource requirements specified as arguments will be identified as
+                # part of the Op definition instantiation
+                required_resource_keys=decorator_resource_keys,
                 tags={
                     **({"kind": self.compute_kind} if self.compute_kind else {}),
                     **(self.op_tags or {}),
@@ -408,6 +413,7 @@ def multi_asset(
     )
 
     required_resource_keys = set(required_resource_keys).union(set(resource_defs.keys()))
+
     for out in outs.values():
         if isinstance(out, Out) and not isinstance(out, AssetOut):
             deprecation_warning(
@@ -423,6 +429,12 @@ def multi_asset(
             fn, ins or {}, non_argument_deps=_make_asset_keys(non_argument_deps)
         )
         asset_outs = build_asset_outs(outs)
+
+        arg_resource_keys = {arg.name for arg in get_resource_args(fn)}
+        check.param_invariant(
+            len(required_resource_keys) == 0 or len(arg_resource_keys) == 0,
+            "Cannot specify resource requirements in both @multi_asset decorator and as arguments to the decorated function",
+        )
 
         # validate that the asset_deps make sense
         valid_asset_deps = set(asset_ins.keys()) | set(asset_outs.keys())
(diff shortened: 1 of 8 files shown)


benpankow:
* master
  * **PR #11322** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11322" target="_blank">Graphite</a>  👈
    * **PR #11321** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11321" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11322?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -263,8 +264,10 @@ def __call__(self, fn: Callable) -> AssetsDefinition:
         with warnings.catch_warnings():
             warnings.simplefilter("ignore", category=ExperimentalWarning)
 
-            required_resource_keys = set(self.required_resource_keys).union(
-                set(self.resource_defs.keys())
+            required_resource_keys = (
+                set(self.required_resource_keys)
+                .union(set(self.resource_defs.keys()))
+                .union({arg.name for arg in get_resource_args(fn)})
```
schrockn, line 270: I think we should strongly consider disallowing specifying `required_resource_keys`/`resource_defs` and the argument-based approach for now, similar to how we do not allow `config` argument and `config_schema`. We can always loosen this later, but I would prefer to be strict for now.

python_modules/dagster/dagster/_core/definitions/decorators/op_decorator.py
```diff
@@ -83,14 +83,18 @@ def __call__(self, fn: Callable[..., Any]) -> "OpDefinition":
         elif self.out is not None:
             outs = check.mapping_param(self.out, "out", key_type=str, value_type=Out)
 
+        resolved_resource_keys = set(self.required_resource_keys or []).union(
```
schrockn, line 86: same here

python_modules/dagster/dagster/_core/definitions/decorators/solid_decorator.py
```diff
@@ -148,6 +157,9 @@ def __call__(self, fn: Callable[..., Any]) -> OpDefinition:
             explicit_input_defs=self.input_defs,
             exclude_nothing=True,
         )
+        resolved_resource_keys = set(self.required_resource_keys or []).union(
```
schrockn, line 160: same

schrockn (changes requested):

schrockn (approved):
Looks great.

Given this is an exploration, I don't think we strictly need it, but make a note that we will need to test and support (or explicitly error) all different use cases where input names and upstream asset names might conflict with resource names.

schrockn (changes requested):
Actually we definitely need some test cases with ops with inputs and assets that have upstream assets. Please add.

benpankow:
Added some tests which go over upstream assets/ops. Doesn't handle the case in which names conflict, since we don't have a good way to resolve this at the moment without threading through resources to the execution planning logic (this may be what we want to do long-term).

python_modules/dagster/dagster/_core/definitions/decorators/solid_decorator.py
```diff
@@ -149,6 +158,14 @@ def __call__(self, fn: Callable[..., Any]) -> OpDefinition:
             exclude_nothing=True,
         )
 
+        arg_resource_keys = {arg.name for arg in compute_fn.get_resource_args()}
+        decorator_resource_keys = set(self.required_resource_keys or [])
+        check.param_invariant(
+            len(decorator_resource_keys) == 0 or len(arg_resource_keys) == 0,
+            "Cannot specify resource requirements in both @solid decorator and as arguments to the decorated function",
```
schrockn, line 165: s/solid/op

schrockn (approved):
👍🏻

benpankow:
**Graphite Merge Job**

Current status: 🔄 Waiting for checks to pass

This pull request is currently waiting for CI and mergeability checks to pass before being merged.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11322?utm_source=stack-comment).


Job Reference: uHrT8DIrkchoQsyuEMKn

PR #11323, opened 2022-12-21
[declarative-scheduling] In the presence of observable source assets, change how we calculate used_data_time

OwenKephart:
### Summary & Motivation

Previously, we completely ignored source assets (of any kind) when calculating data times. So if you had:

`observable_source -> a -> b`

then the data time of `b` would be entirely based on the timestamp of the materialization of `a` that was consumed by the latest materialization of `b`.

this is not a particularly accurate way to calculate how much of the data of the outside world that `b` has consumed at any given point in time. `observable_source` may represent some directory that rarely gets data added to it. So it may be completely valid for `a` not to execute for hours or days at a time, waiting until new data shows up. During that time, it's accurate to say that `a` has consumed all of the available data up until the current time.

Only once a new version of the source is observed, then `a` and `b` can begin to go out of date.

More specifically, it we look at a timeline of events:

```
[v1 observed]........[a materialized]...[v2 observed].....[a materialized]
```

After a is materialized the first time, up until v2 is observed, the data time of `a` will always be measured to be the current time. After v2 is observed, the data time of `a` will be equal to the timestamp of the observation event for `v2`. Once `a` is materialized again, the data time will go back to the current time.


OwenKephart:
@sryza yep this still needs review but it'll need to be reworked a bit in light of a bunch of the other changes in this area. I'll do that reworking once all the perf stuff is merged in, so I'll re-request you once that happens.

PR #11327, opened 2022-12-21
[structured config] Fix usage with Assets

benpankow:
## Summary

Since asset inputs flow through a separate code path, the `config` parameter wasn't being stripped out properly when using structured config with assets. Fixes + adds test.

### DIFF
--- a/python_modules/dagster/dagster_tests/core_tests/structured_config_tests/test_basic_structured_config.py
+++ b/python_modules/dagster/dagster_tests/core_tests/structured_config_tests/test_basic_structured_config.py
@@ -2,11 +2,12 @@ import pytest
 from pydantic import BaseModel
 
 from dagster import _check as check
-from dagster import job, op, validate_run_config
+from dagster import asset, job, op, validate_run_config
 from dagster._config.config_type import ConfigTypeKind
 from dagster._config.field_utils import convert_potential_field
 from dagster._config.structured_config import Config, infer_schema_from_config_class
 from dagster._config.type_printer import print_config_type_to_string
+from dagster._core.definitions.assets_job import build_assets_job
 from dagster._core.errors import DagsterInvalidConfigDefinitionError, DagsterInvalidConfigError
 from dagster._core.execution.context.invocation import build_op_context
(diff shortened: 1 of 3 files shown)


benpankow:
* master
  * **PR #11327** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11327" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11327?utm_source=stack-comment).

schrockn (approved):

python_modules/dagster/dagster_tests/core_tests/structured_config_tests/test_basic_structured_config.py
```diff
@@ -136,6 +137,28 @@ def my_asset(config: AnAssetConfig):
...
+        build_assets_job(
+            "blah",
+            [two_assets],
+            config={"ops": {"two_assets": {"config": {"a_string": "foo", "an_int": 2}}}},
+        )
+        .execute_in_process()
+        .success
+    )
+
+
```
schrockn, line 161: One thing I like to do in test cases like these is to absolutely ensure that the function body has been executed. 

Meaning before the function (in this case the asset):

`executed = {}`

Then in the function body:

`executed["yes"] = True`

Then after job execution:

`assert executed["yes"]`

When the next engineer comes along and they see that they have high confidence that the code is actually executed. Whenever I work on a framework that is calling back into user code that is doing asserts (and not verifying outputs) I usually do this. More than once I've been bitten debugging something only realizing after some time that some function was actually never called.

Don't feel super strongly so I'm not going to block commit, but I generally find it's a good best practice.

schrockn (approved):
Approved but consider final comment.

python_modules/dagster/dagster_tests/core_tests/structured_config_tests/test_basic_structured_config.py
benpankow: updated

PR #11332, opened 2022-12-22
store repository on external asset graph

sryza:
Adds a subclass of AssetGraph called ExternalAssetGraph which stores a RepositoryHandle for each asset

branch-name: external-asset-graph-repo


OwenKephart (approved):
this is great

PR #11333, opened 2022-12-22
Allow setting raw k8s config at the run launcher / container context level

gibsondan:
Summary:
This will allow users to set raw k8s config in the helm chart that applies to all launched runs - no more setting tags on every job just to e.g. set a TTL globally - and no more adding helm chart params one by one as people request them.

Also adds a container_context hook which will let us in the future vary raw config like this per-code location. The hook covers both runs and servers (currently unused in OSS since we don't have a way to programatically spin up gRPC servers in OSS).

Test Plan: New automated tests

### DIFF
--- a/helm/dagster/templates/helpers/instance/_run-launcher.tpl
+++ b/helm/dagster/templates/helpers/instance/_run-launcher.tpl
@@ -124,11 +124,30 @@ config:
   {{- end }}
 
   {{- if $k8sRunLauncherConfig.schedulerName }}
-  schedulerName: {{ $k8sRunLauncherConfig.schedulerName | quote }}
+  scheduler_name: {{ $k8sRunLauncherConfig.schedulerName | quote }}
   {{- end }}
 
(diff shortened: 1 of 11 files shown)


helm/dagster/templates/helpers/instance/_run-launcher.tpl
```diff
@@ -124,11 +124,30 @@ config:
   {{- end }}
 
   {{- if $k8sRunLauncherConfig.schedulerName }}
-  schedulerName: {{ $k8sRunLauncherConfig.schedulerName | quote }}
+  scheduler_name: {{ $k8sRunLauncherConfig.schedulerName | quote }}
```
johannkm, line 127: dang this was getting ignored before?

johannkm (approved):

helm/dagster/templates/helpers/instance/_run-launcher.tpl
gibsondan: i think it was most likely erroring when set

PR #11344, opened 2022-12-22
Rename "base asset job" to "implicit asset job" in RepositoryDefinition methods

schrockn:
### Summary & Motivation

The language of "implicit asset jobs" is more accurate and clear than "base asset jobs" so we are changing these method names.

BK

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/repository_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/repository_definition.py
@@ -1368,7 +1368,7 @@ class RepositoryDefinition:
 
         return self.get_job(ASSET_BASE_JOB_PREFIX)
 
-    def get_implicit_asset_job_def_names(self) -> Sequence[str]:
+    def get_implicit_asset_job_names(self) -> Sequence[str]:
         return [
             job_name for job_name in self.job_names if job_name.startswith(ASSET_BASE_JOB_PREFIX)
         ]
--- a/python_modules/dagster/dagster/_core/definitions/repository_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/repository_definition.py
@@ -1368,12 +1368,14 @@ class RepositoryDefinition:
 
         return self.get_job(ASSET_BASE_JOB_PREFIX)
 
-    def get_base_asset_job_names(self) -> Sequence[str]:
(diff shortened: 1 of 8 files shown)


schrockn:
* master
  * **PR #11279** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11279" target="_blank">Graphite</a> 
    * **PR #11344** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11344" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11344?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/repository_definition.py
```diff
@@ -1366,12 +1366,14 @@ def get_implicit_global_asset_job_def(self) -> JobDefinition:
 
         return self.get_job(ASSET_BASE_JOB_PREFIX)
 
-    def get_base_asset_job_names(self) -> Sequence[str]:
+    def get_implicit_asset_job_def_names(self) -> Sequence[str]:
```
sryza, line 1369: Do we need "def" in this name? I think we refer to "job names" in a bunch of other places, and IMO it's clear enough what that means.

schrockn: All the other method names refer to defs so this feels inconsistent. Seems like we need to come up with consistent guidance around this. The codebase is all over the place on this topic.

sryza: My reading of the larger codebase is that we've settled on job_name. I.e. I see job_name in a lot of places and not much job_def_name.  E.g. DagsterRun has a job_name, ScheduleDefinition/SensorDefinition/RunRequest/JobSelector all accept a job_name, the CLI talks about job_names.

I don't think I'd find it weird to come across both `get_job_defs` and `get_job_names` in `Definitions`.

sryza:
Was just noticing that we still have this divergence in naming ("base" in some parts of the codebase vs. "implicit" in others).  No crazy rush, but wanted to throw a vote in for getting these lined up.

sryza (approved):

PR #11372, opened 2022-12-23
Rename and refactor in structured_config.py for clarity

schrockn:
### Summary & Motivation

I was adding some functionality here and I thought a couple changes made this quite a bit clearer.

1) Consolidate the logic to create the configured config schema into a single function `_curry_config_schema`. This removes code duplication and makes it clear that the "inner resource" is solely for this purpose.  Passing `self.resource_function` to this placeholder made this particularly confusing.

2) Rename `resource_function` to make it distinct from `resource_fn` on the base class. I'm proposing the verbose but explicit name `create_object_to_pass_to_user_code`. Open to feedback.

### How I Tested These Changes

BK

PR #11373, opened 2022-12-24
Consolidate all logic and trickery to support private cached properties in pydantic in a single class

schrockn:
### Summary & Motivation

We are doing some real shenanigans to support cached_property and cached_method on these Config classes.
This PR consolidates them all into a single class that Config inherits from. This a few things things:

1) Vanilla Config classes now can do this
2) All hackery compartmentalized into a single class that can be removed cleanly.
3) Reduces the number of pyright errors about frozen classes to 1.

BK

### DIFF
--- a/python_modules/dagster/dagster/_config/structured_config.py
+++ b/python_modules/dagster/dagster/_config/structured_config.py
@@ -27,7 +27,34 @@ from dagster._core.definitions.resource_definition import ResourceDefinition, Re
 from dagster._core.storage.io_manager import IOManager, IOManagerDefinition
 
 
-class Config(BaseModel):
+class MakeConfigCacheable(
+    BaseModel,
+    # Various pydantic model config (https://docs.pydantic.dev/usage/model_config/)
+    # Necessary to allow for caching decorators
+    arbitrary_types_allowed=True,
+    # Avoid pydantic reading a cached property class as part of the schema
+    keep_untouched=(cached_property,),
+    # Ensure the class is serializable, for caching purposes
+    frozen=True,
+):
+    """This class centralizes and implements all the chicanery we need in order
+    to support caching decorators. If we decide this is a bad idea we can remove it
+    all in one go.
+    """
+
+    def __setattr__(self, name: str, value: Any):
+        # This is a hack to allow us to set attributes on the class that are not part of the
+        # config schema. Pydantic will normally raise an error if you try to set an attribute
+        # that is not part of the schema.
+
+        if name.startswith("_") or name.endswith("_cache"):
+            object.__setattr__(self, name, value)
+            return
+
+        return super().__setattr__(name, value)
+
+
+class Config(MakeConfigCacheable):
     """
     Base class for Dagster configuration models.
     """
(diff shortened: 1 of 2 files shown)


schrockn:
* master
  * **PR #11372** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11372" target="_blank">Graphite</a> 
    * **PR #11373** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11373" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11373?utm_source=stack-comment).

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -27,7 +27,34 @@ class cached_property:  # type: ignore[no-redef]
 from dagster._core.storage.io_manager import IOManager, IOManagerDefinition
 
 
-class Config(BaseModel):
+class MakeConfigCacheable(
```
benpankow, line 30: maybe `CacheableBaseModel`? or something like `ConfigBase`

python_modules/dagster/dagster_tests/core_tests/structured_config_tests/test_basic_structured_config.py
```diff
@@ -351,3 +354,88 @@ def pipeline_requires_config():
...
+    assert counts["plus"] == 1
+    assert counts["mult"] == 0
+
+    assert config.mult == 15
+
+    assert counts["plus"] == 1
+    assert counts["mult"] == 1
+
+
+def test_cached_method():
```
benpankow, line 403: Python 3.9 also introduces [`functools.cache`](https://docs.python.org/3/library/functools.html#functools.cache) - wonder if this would also work out of the box (seems functionally quite similar to `cached_method`)

benpankow (approved):
looks good, thanks for new tests

python_modules/dagster/dagster/_config/structured_config.py
schrockn: I am against a *Base naming convention as it is extremely unclear why the extra class exists.  `MakeConfigCacheable` is at a minimum very clear. I can change `CacheableBaseModel` but I actually don't like too much either as no one else is supposed to inherit form this. `MakeConfigCacheable` indicates the sole purpose of the class is to make the `Config` cache-able and that is all.

benpankow: keeping it as `MakeConfigCacheable` seems fine to me

python_modules/dagster/dagster_tests/core_tests/structured_config_tests/test_basic_structured_config.py
schrockn: It does not. 

```
    @pytest.mark.skipif(sys.version_info < (3, 9), reason="requires python3.9")
    def test_functools_cache():
        from functools import cache

        counts = {
            "plus": 0,
            "mult": 0,
        }

        class SomeConfig(Config):
            x: int
            y: int

            @cache
            def plus(self):
                counts["plus"] += 1
                return self.x + self.y

            @cache
            def mult(self):
                counts["mult"] += 1
                return self.x * self.y

        config = SomeConfig(x=3, y=5)

        assert counts["plus"] == 0
        assert counts["mult"] == 0

>       assert config.plus() == 8
E       TypeError: test_functools_cache.<locals>.SomeConfig.plus() missing 1 required positional argument: 'self'
```

PR #11378, opened 2022-12-25
handle pure asset backfills in backfill daemon

sryza:
branch-name: asset-backfill-daemon

### Summary & Motivation

Makes the backfill daemon handle pure asset backfills. Wraps the core asset backfill logic in that's implemented in https://github.com/dagster-io/dagster/pull/11377.

- Added tests
- With the graphql changes in https://github.com/dagster-io/dagster/pull/11377, launched backfills manually

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/asset_backfill.py
+++ b/python_modules/dagster/dagster/_core/execution/asset_backfill.py
@@ -1,6 +1,17 @@
 import json
-from typing import AbstractSet, Iterable, List, NamedTuple, Optional, Sequence, Set
+from typing import (
+    TYPE_CHECKING,
+    AbstractSet,
+    Iterable,
+    List,
+    NamedTuple,
+    Optional,
+    Sequence,
+    Set,
+    cast,
+)
 
+from dagster import _check as check
 from dagster._core.definitions.asset_graph import AssetGraph
 from dagster._core.definitions.asset_graph_subset import AssetGraphSubset
 from dagster._core.definitions.asset_reconciliation_sensor import (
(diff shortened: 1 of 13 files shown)


python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -93,6 +110,108 @@ def serialize(self) -> str:
...
+        asset_backfill_data=asset_backfill_data,
+        instance=instance,
+        asset_graph=asset_graph,
+    )
+
+    updated_backfill = backfill.with_asset_backfill_data(result.backfill_data)
+    if result.backfill_data.is_complete():
+        updated_backfill = updated_backfill.with_status(BulkActionStatus.COMPLETED)
+
+    for run_request in result.run_requests:
```
prha, line 141: you should probably be yielding `None` in loops, to give the daemon to heartbeat.  o.w. this can cause the daemon to get killed and restarted in a crash loop.

python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -93,6 +110,108 @@ def serialize(self) -> str:
...
+
+    asset_graph = ExternalAssetGraph.from_workspace_request_context(workspace)
+    if backfill.serialized_asset_backfill_data is None:
+        check.failed("Asset backfill missing serialized_asset_backfill_data")
+
+    asset_backfill_data = AssetBackfillData.from_serialized(
+        backfill.serialized_asset_backfill_data, asset_graph
+    )
+
+    result = execute_asset_backfill_iteration_inner(
```
prha, line 130: I think potentially you'd also want to change this to an iterator to allow for heartbeating, if this is potentially going to take a long time to execute.

prha (changes requested):

python_modules/dagster/dagster/_core/execution/asset_backfill.py
sryza: 👍

python_modules/dagster/dagster/_core/execution/asset_backfill.py
sryza: 👍

sryza:
@prha I updated the PR to turn `execute_asset_backfill_iteration` and `execute_asset_backfill_iteration_inner` into generators.   It makes the code a little uglier, but seems worth it.

prha:
I think probably the generator wrapping `execute_asset_backfill_iteration_inner` is unnecessary, if computing the results is not expensive.

sryza:
> I think probably the generator wrapping execute_asset_backfill_iteration_inner is unnecessary, if computing the results is not expensive.

Alas I think it probably can be expensive. It does O((# assets) + (# materialized asset partitions since last tick)) queries to the DB.

prha (approved):

prha (approved):

PR #11392, opened 2022-12-27
deprecate job-level memoization in docs

sryza:
### Summary & Motivation

This was inspired by this user question: https://dagster.slack.com/archives/C01U954MEER/p1672159464819809.

Right now, if you google / search the Dagster docs, you're likely to come under the impression that the 2021 job-level versioning system is the way we recommend doing versioning and memoization in Dagster. Instead, I think that the asset versioning system introduced in 1.1 is what we should point people to, and we should consider the earlier one deprecated.

I have some sad feelings about removing this, because I think it was quite cool for its time, but it ended up out-of-step with the direction of the project.


dpeng817 (approved):
Fine with this, having two parallel versioning systems for op land and asset land seems undesirable. I do wonder though, what should a user then do if they want memoization for their op job? Should they go back to providing versions on ops individually? We should probably have some recommendations for that

docs/sphinx/sections/api/apidocs/memoization.rst
```diff
@@ -1,8 +1,9 @@
-Versioning and Memoization (Experimental)
-=========================================
+Job-Level Versioning and Memoization (Deprecated)
```
erinkcochran87, line 1: I don't see this change reflected in the Vercel preview - was the build done correctly? https://dagster-git-memoization-docs-elementl.vercel.app/_apidocs/memoization

docs/sphinx/sections/api/apidocs/memoization.rst
```diff
@@ -1,8 +1,9 @@
...
+=================================================
 
-Dagster allows for code versioning and memoization of previous outputs based upon that versioning.
-Listed here are APIs related to versioning and memoization.
+Dagster has deprecated functionality that allows for job-level code versioning and memoization of previous op outputs based upon that versioning.
+
+This is currently deprecated in favor of asset versioning.
```
erinkcochran87, line 6: Add a link to the docs about asset versioning here?

docs/content/guides/experimental-features.mdx
```diff
@@ -4,6 +4,4 @@ title: Experimental Feature Guides | Dagster Docs
 
 # Experimental feature guides
 
-- [Using versioning and memoization](/guides/dagster/memoization) - Learn to use Dagster's versioning and memoization features in job re-execution
-
```
erinkcochran87, line 8: This block also needs to be removed from here: `/docs/content/guides.mdx`

erinkcochran87 (changes requested):
See inline comments

sryza:
> I do wonder though, what should a user then do if they want memoization for their op job?

My take would basically be that we don't support that (although they can still use the existing APIs if they dig deep enough). The argument would be that memoization only makes sense when you have persistent objects that outlive the scope of a run, i.e. assets.

docs/sphinx/sections/api/apidocs/memoization.rst
sryza: I just tried this, but the links in RST don't appear to be rendering correctly: https://dagster-git-memoization-docs-elementl.vercel.app/_apidocs/memoization. Any tips?

erinkcochran87: Yep! It took me a minute to figure this out, but here ya go:

```
# Syntax: `display text <https://somelink.com>`_

This is currently deprecated in favor of `asset versioning </concepts/assets/software-defined-assets#asset-code-versions>`_
```

It looks like reStructuredText doesn't use the flavor of Markdown we're used to, so styling is a bit different. I'm gonna add a formatting reference to the Dagster metadocs to make this info easily findable in the future.

dpeng817:
I see; essentially if you want memoization, you likely should be using assets.

erinkcochran87 (approved):
Didn't see you requested a review from me again - sorry!

I answered your ? about the syntax for links in API docs. If ya fix that up and it looks good, :shipit:

PR #11406, opened 2022-12-27
Be more resilient to user code errors when dequeuing runs

gibsondan:
Summary:
Right now if the user code server is unreachable while a run is being dequeued - e.g. you're using the default run launcher and the grpc server is down, or you're using cloud and the run is launched via the agent - the dequeue will immediately give up.

This is a work-in-progress solution that instead:
- Retries up to N times by re-enqueing the run
- Puts the last user code failure timestamp for that code location in a map to make it wait to try again for a while, so we don't race through the N retries right away

The other missing piece here would be to add a hook on each iteration to optionally check that all user code servers are unavailable - that would cover cases in cloud like the agent being in the middle of being upgraded for an extended period of time.

### DIFF
--- a/python_modules/dagster/dagster/_core/run_coordinator/queued_run_coordinator.py
+++ b/python_modules/dagster/dagster/_core/run_coordinator/queued_run_coordinator.py
@@ -17,10 +17,25 @@ class RunQueueConfig(
         [
             ("max_concurrent_runs", int),
             ("tag_concurrency_limits", Optional[Sequence[Mapping[str, Any]]]),
+            ("max_user_code_failure_retries", int),
+            ("user_code_failure_retry_delay", int),
         ],
     )
 ):
-    pass
+    def __new__(
+        cls,
+        max_concurrent_runs,
+        tag_concurrency_limits,
+        max_user_code_failure_retries=0,
+        user_code_failure_retry_delay=60,
+    ):
+        return super(RunQueueConfig, cls).__new__(
+            cls,
+            check.int_param(max_concurrent_runs, "max_concurrent_runs"),
+            check.opt_sequence_param(tag_concurrency_limits, "tag_concurrency_limits"),
+            check.int_param(max_user_code_failure_retries, "max_user_code_failure_retries"),
+            check.int_param(user_code_failure_retry_delay, "user_code_failure_retry_delay"),
+        )
 
 
 class QueuedRunCoordinator(RunCoordinator, ConfigurableClass):
@@ -36,6 +51,8 @@ class QueuedRunCoordinator(RunCoordinator, ConfigurableClass):
         dequeue_interval_seconds=None,
         dequeue_use_threads=None,
         dequeue_num_workers=None,
+        max_user_code_failure_retries=None,
+        user_code_failure_retry_delay=None,
         inst_data=None,
     ):
         self._inst_data = check.opt_inst_param(inst_data, "inst_data", ConfigurableClassData)
@@ -57,6 +74,12 @@ class QueuedRunCoordinator(RunCoordinator, ConfigurableClass):
             dequeue_use_threads, "dequeue_use_threads", False
         )
         self._dequeue_num_workers = check.opt_int_param(dequeue_num_workers, "dequeue_num_workers")
+        self._max_user_code_failure_retries = check.opt_int_param(
+            max_user_code_failure_retries, "max_user_code_failure_retries", 0
+        )
+        self._user_code_failure_retry_delay = check.opt_int_param(
+            user_code_failure_retry_delay, "user_code_failure_retry_delay", 60
+        )
         self._logger = logging.getLogger("dagster.run_coordinator.queued_run_coordinator")
         super().__init__()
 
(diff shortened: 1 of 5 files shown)


johannkm (commented):
Makes sense to me. Might have some challenges with reporting the situation. e.g. if I launch a bunch of runs for a code location, one will have the error reported but the rest will just sit in the queue without any message explaining. Could add something later

python_modules/dagster/dagster/_daemon/run_coordinator/queued_run_coordinator_daemon.py
```diff
@@ -133,50 +144,73 @@ def run_iteration(
...
 
+    def _dequeue_run_thread(
+        self,
+        workspace_process_context: IWorkspaceProcessContext,
+        run: DagsterRun,
+        run_queue_config: RunQueueConfig,
+    ) -> bool:
+        return self._dequeue_run(
+            workspace_process_context.instance,
+            workspace_process_context.create_request_context(),
```
gibsondan, line 188: curious how intentional it is that we make a new request context on each thread

alangenfeld: Im not sure how intentional it was. 

When the thread pool executor work queue is deeper than the number of workers, you can end up accumulating request contexts in memory which is a problem we have had before.

I don't believe creating request contexts has shown up as costly in profiling.

python_modules/dagster/dagster/_daemon/run_coordinator/queued_run_coordinator_daemon.py
```diff
@@ -270,48 +353,131 @@ def get_priority(run):
...
+                run.status,
             )
-            message_with_full_error = f"{message}: {error_info.to_string()}"
+            return False
 
-            self._logger.error(message_with_full_error)
-            instance.report_run_failed(run, message_with_full_error)
+        location_name = (
+            run.external_pipeline_origin.location_name if run.external_pipeline_origin else None
+        )
```
alangenfeld, line 387: in what conditions is this None?

python_modules/dagster/dagster/_daemon/run_coordinator/queued_run_coordinator_daemon.py
```diff
@@ -270,48 +353,131 @@ def get_priority(run):
...
+            elif run_queue_config.max_user_code_failure_retries and isinstance(
+                e, (DagsterUserCodeUnreachableError, DagsterRepositoryLocationLoadError)
+            ):
+
+                if location_name:
+                    with self._location_timeouts_lock:
+                        # Don't try to dequeue runs from this location for another N seconds
+                        self._location_timeouts[location_name] = (
+                            now + run_queue_config.user_code_failure_retry_delay
+                        )
```
alangenfeld, line 428: a comment about the expected behavior for the `else` case here

python_modules/dagster/dagster/_daemon/run_coordinator/queued_run_coordinator_daemon.py
```diff
@@ -270,48 +353,131 @@ def get_priority(run):
...
+                    )
+                    # Re-submit the run into the queue
+                    enqueued_event = DagsterEvent(
+                        event_type_value=DagsterEventType.PIPELINE_ENQUEUED.value,
+                        pipeline_name=run.pipeline_name,
+                    )
+                    instance.report_dagster_event(enqueued_event, run_id=run.run_id)
+                    return False
+            else:
+                message = "Caught an error while dequeuing the run. Marking the run as failed and dropping it from the queue"
```
alangenfeld, line 471: nit: "unexpected error"  "unrecoverable error" maybe

python_modules/dagster/dagster/_core/run_coordinator/queued_run_coordinator.py
```diff
@@ -133,6 +158,24 @@ def config_type(cls):
                 is_required=False,
                 description="If dequeue_use_threads is true, limit the number of concurrent worker threads.",
             ),
+            "max_user_code_failure_retries": Field(
+                config=IntSource,
+                is_required=False,
+                default_value=0,
```
alangenfeld, line 164: we expect to bump this up after we get some usage on this right?

gibsondan: We could - the thing about it is that the number of run launchers that go through user code in OSS is actually quite low by design (today it's only the default run launcher)

python_modules/dagster/dagster/_daemon/run_coordinator/queued_run_coordinator_daemon.py
gibsondan: Pretty much impossible now, maybe runs enqueued back on like 0.9.0

gibsondan: or maybe runs programatically submitted in python somehow

alangenfeld: think its worth leave a comment here

alangenfeld (approved):

PR #11407, opened 2022-12-27
Unexperimentalize LogicalVersion and quell warning messages whenever assets are materialized

sryza:
### Summary & Motivation

I observed that we were emitting warnings whenever assets were materialized. Our tests weren't caching these because https://github.com/dagster-io/dagster/pull/9577 was masking them.

Rather than hunt for every site where we instantiate `LogicalVersion`, I elected to take off the experimental marker. I think this is safe because the places it can be used (`observable_source_asset`) are experimental.

I also filtered out deprecation warnings that occur when deserializing JSON into named tuples, which we were hitting with MetadataEntries because of the deprecated `entry_data` field.

### DIFF
--- a/python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_group.py
+++ b/python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_group.py
@@ -43,6 +43,9 @@ def _all_asset_keys(result):
 @pytest.fixture(autouse=True)
 def check_experimental_warnings():
     with warnings.catch_warnings(record=True) as record:
+        # turn off any outer warnings filters
+        warnings.resetwarnings()
+
         yield
 
         raises_warning = False
--- a/python_modules/dagster/dagster/_core/definitions/logical_version.py
+++ b/python_modules/dagster/dagster/_core/definitions/logical_version.py
@@ -1,14 +1,11 @@
 from __future__ import annotations
 
-import warnings
 from hashlib import sha256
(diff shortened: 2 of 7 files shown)


python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_group.py
```diff
@@ -43,6 +43,9 @@ def _all_asset_keys(result):
 @pytest.fixture(autouse=True)
 def check_experimental_warnings():
     with warnings.catch_warnings(record=True) as record:
+        # turn off any outer warnings filters
+        warnings.resetwarnings()
```
smackesey, line 47: I might not be understanding, but does turning off external filters actually end up testing for a different warning profile than a user gets in a default environment?

smackesey (approved):
Looks good as a stopgap (I have a question about the tests though).

Probably a better long term solution is to (as previously discussed around time of 1.0) move experimental warnings out of class instantiation to the import layer, or otherwise figure out a way to limit redundant warnings, but not a priority.

python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_group.py
sryza: Yes, because external filters are set by our test infrastructure, which is not present in user environments.

https://github.com/dagster-io/dagster/blob/master/pyproject.toml#L165

sryza:
@smackesey +1 about that as a better long-term solution

PR #11411, opened 2022-12-28
[dagit] Begin adding new graphql-codegen

hellendag:
### Summary & Motivation

Apollo is deprecating its codegen CLI feature in favor of https://the-guild.dev/graphql/codegen. Begin swapping out our old codegen with the new.

- Use the recommended client preset to generate GraphQL types via `yarn graphql-codegen`. The produced output can be seen in the `src/graphql` directory. Namely, it produces:
  - `graphql` functions that are bound to query documents, which can be used at locations where queries are defined, instead of the `gql` template literal. This allows us to make `useQuery` calls that do not require annotation! Based on the supplied query, the variables and output are all completely typechecked. (See `VersionNumber` for an example.)
  - Types for all schema fields and query outputs. These are similar to our existing codegenned types, but they do *not* include the intermediate per-field types for each query. This means we'll need to do some refactoring in places where we reference nested types (`MyQuery_someField_subFieldFoo`) to use fragments or similar.
  - All types are located within `src/graphl`, so all GrahpQL type imports will be from the same place (src/graphql/graphql).

In the codegen configuration, I am disabling fragment masking at this point, since it adds more overhead to the refactor. We can add it later if it seems like something we'll want.

I have updated `VersionNumber` to serve as an example. I'll make a comment inline.

The recommendation (as noted in the codegen output file) is to use the preset's provided Babel plugin to reduce bundle size, since all GraphQL documents are output into the codegen file. We'll need to do this by updating our CRA fork, since that's where our Babel config lives. For now, our bundle size is a bit larger, like +20kb gzipped in the main bundle.

The `yarn generate-graphql` command will now produce *both* old and new types. When all of our GraphQL usage is moved over to the new API and types, we can remove the old types entirely.

Next steps:

- Begin converting existing `gql` TTLs to `graphql()` calls and fix types as we go.

TS, lint, jest. Run same on dagit-cloud to verify that everything still works.

Load Dagit, verify that the `VersionNumber` component behaves and renders correctly.

Verify that lint does not run on `src/graphql` output, since that will just be expensive and pointless. Verify that `__typename` values are present on all types, and that enum values preserve `ALL_CAPS` keys.

Produce a prod build with stats, verify that the bundle hasn't exploded in size due to the new codegen output.


hellendag:
* master
  * **PR #11411** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11411" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11411?utm_source=stack-comment).

js_modules/dagit/packages/core/.eslintrc.js
```diff
@@ -5,9 +5,8 @@ const schemaPath = path.resolve(path.join(__dirname, 'src', 'graphql', 'schema.g
 const schema = fs.readFileSync(schemaPath).toString();
 
 module.exports = {
-  extends: [
-    'plugin:storybook/recommended', '@dagster-io/eslint-config',
-  ],
+  extends: ['plugin:storybook/recommended', '@dagster-io/eslint-config'],
+  ignorePatterns: ['src/graphql/**'],
```
hellendag, line 9: Avoid linting codegen output.

js_modules/dagit/packages/core/codegen.ts
```diff
@@ -0,0 +1,25 @@
...
+  schema: './src/graphql/schema.graphql',
+  documents: ['src/**/*.tsx'],
+  ignoreNoDocuments: true, // for better experience with the watcher
+  generates: {
+    './src/graphql/': {
+      preset: 'client',
+      config: {
+        nonOptionalTypename: true,
+        namingConvention: {
+          enumValues: 'keep',
```
hellendag, line 13: Maintain `ALL_CAPS` enum keys for now. We can remove this in a later refactor step to simplify the codegen config, and update our enums to use `PascalCase`.

js_modules/dagit/packages/core/codegen.ts
```diff
@@ -0,0 +1,25 @@
...
+        },
+      },
+      presetConfig: {
+        fragmentMasking: false,
+      },
```
hellendag, line 18: We don't need fragment masking for now. Maybe later.

js_modules/dagit/packages/core/src/nav/VersionNumber.tsx
```diff
@@ -1,23 +1,20 @@
...
   font-size: 11px;
 `;
 
-const VERSION_NUMBER_QUERY = gql`
+const VERSION_NUMBER_QUERY = graphql(`
   query VersionNumberQuery {
     version
   }
-`;
+`);
```
hellendag, line 20: An example of what the change looks like. Note that we can remove the type annotations on the `useQuery` call above! 🎉

js_modules/dagit/packages/core/src/schedules/ScheduleMutations.tsx
```diff
@@ -4,11 +4,14 @@ import * as React from 'react';
...
-import {StartSchedule, StartSchedule_startSchedule_PythonError} from './types/StartSchedule';
+import {
+  StartThisSchedule,
+  StartThisSchedule_startSchedule_PythonError,
+} from './types/StartThisSchedule';
 import {StopSchedule, StopSchedule_stopRunningSchedule_PythonError} from './types/StopSchedule';
 
 export const START_SCHEDULE_MUTATION = gql`
-  mutation StartSchedule($scheduleSelector: ScheduleSelector!) {
+  mutation StartThisSchedule($scheduleSelector: ScheduleSelector!) {
```
hellendag, line 14: `StartScheduleMutation` is a generated schema type, and is also the name of the type generated for this specific mutation, which leads to a TypeScript error due to multiple types having the same name. To sidestep this, I just renamed the mutation.

js_modules/dagit/packages/core/src/sensors/SensorMutations.tsx
```diff
@@ -26,7 +29,7 @@ export const START_SENSOR_MUTATION = gql`
 `;
 
 export const STOP_SENSOR_MUTATION = gql`
-  mutation StopSensor($jobOriginId: String!, $jobSelectorId: String!) {
+  mutation StopRunningSensor($jobOriginId: String!, $jobSelectorId: String!) {
```
hellendag, line 32: Same issue as `StartSchedule`.

salazarm (approved):

PR #11421, opened 2022-12-29
[dagster-databricks] handle execution interrupts

OwenKephart:
### Summary & Motivation

When a user cancels execution from dagit, you'd expect computation to cease on the remote cluster. Instead, the op would be marked as failed within dagit, but no cancellation signal would be sent to the remote database cluster, so computation would continue, which is wasteful of compute resources.

Started a long-running step using the databricks_pyspark_step_launcher, then cancelled it from the UI, confirmed that the job was cancelled on the cluster. Did the same thing without cancelling, no cancellation signal sent. Also tried adding an error to the step so that it would fail on the remote cluster, and no cancellation signal was sent (as expected)

### DIFF
--- a/python_modules/libraries/dagster-databricks/dagster_databricks/databricks_pyspark_step_launcher.py
+++ b/python_modules/libraries/dagster-databricks/dagster_databricks/databricks_pyspark_step_launcher.py
@@ -222,12 +222,17 @@ class DatabricksPySparkStepLauncher(StepLauncher):
         if self.permissions:
             self._grant_permissions(log, databricks_run_id)
 
+        completed = False
         try:
             # If this is being called within a `capture_interrupts` context, allow interrupts while
             # waiting for the  execution to complete, so that we can terminate slow or hanging steps
             with raise_execution_interrupts():
                 yield from self.step_events_iterator(step_context, step_key, databricks_run_id)
+            completed = True
         finally:
+            # if execution is interrupted before the step is completed, cancel the run
+            if not completed:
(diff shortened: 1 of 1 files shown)


python_modules/libraries/dagster-databricks/dagster_databricks/databricks_pyspark_step_launcher.py
```diff
@@ -222,12 +222,17 @@ def launch_step(self, step_context):
...
         try:
             # If this is being called within a `capture_interrupts` context, allow interrupts while
             # waiting for the  execution to complete, so that we can terminate slow or hanging steps
             with raise_execution_interrupts():
                 yield from self.step_events_iterator(step_context, step_key, databricks_run_id)
+            completed = True
         finally:
+            # if execution is interrupted before the step is completed, cancel the run
+            if not completed:
+                self.databricks_runner.client.client.jobs.cancel_run(databricks_run_id)
```
gibsondan, line 235: is this meaningfully different from

```
except:
   self.databricks_runner.client.client.jobs.cancel_run(databricks_run_id)  
finally:
```

without the completed boolean?

gibsondan (approved):

python_modules/libraries/dagster-databricks/dagster_databricks/databricks_pyspark_step_launcher.py
OwenKephart: yeah that's a lot cleaner, so I switched to that (+ the raise so that the error isn't swallowed). Verified that cancelling a long-running step from the UI does indeed cancel the job on the databricks side.

PR #11425, opened 2022-12-29
BigQuery IO manager

jamiedemaria:
### Summary & Motivation
Adds a BigQuery IO manager following the DB IO manager pattern

To Do:
- [x] update API doc strings
- [x] Finalize how to handle the differing hierarchy of BigQuery (project, dataset, table for BQ and database, schema, table for most other dbs)
- [x] auth via env var support (store the var in a tamp file then set that file to GOOGLE_AUTH_CREDENTIALS)
- [x] convert all columns to upper case

@benpankow @sryza  some questions I'd appreciate your input on:
1. [resolved - moving to dagster_gcp_pandas] What should the name of the package currently named `dagster-bigquery-pandas` be called? 
Originally I named it this without thinking too much about it for the POC implementation. Now I think `dagster-gcp-pandas` would be a better name since it matches `dagster-gcp` and allows us to include more pandas type handlers if other GCP io managers need them
2. [resolved, keeping "schema" metadata for ops] The BigQuery table hierarchy is a bit different than other DBs. They have "datasets" instead of "schemas" and internally, I'm just mapping "dataset" to "schema" in the implementation. However, for ops, we can specify the schema/dataset using the "schema" metadata (see this [line](https://github.com/dagster-io/dagster/pull/11425/files#diff-0403f094881f04d0bed43a52525ac5b5b1f5c976b0d5ee81f685828f1c01888dR62)). I'm considering adding an additional metadata called "dataset" that we can check so that BigQuery users can use it. sound reasonable? 

new unit tests


jamiedemaria:
* master
  * **PR #11425** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11425" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11425?utm_source=stack-comment).

benpankow (commented):
first pass looks good to me - left a few small nits, but this feels good to go once the todos are finished

> What should the name of the package currently named dagster-bigquery-pandas be called?

`dagster-gcp-pandas` sounds good to me, since we already bundle our base GCP integrations it also makes sense to bundle the type handlers.


> I'm considering adding an additional metadata called "dataset" that we can check so that BigQuery users can use it. sound reasonable?

This makes sense to me (with `dataset` acting as an alias for `schema` and `schema` still working) but will defer to @sryza in case there's a reason we want to limit the number of metadata fields or avoid diverging.

.buildkite/dagster-buildkite/dagster_buildkite/steps/packages.py
```diff
@@ -448,6 +448,16 @@ def k8s_extra_cmds(version: str, _) -> List[str]:
         # Remove once https://github.com/dagster-io/dagster/issues/2511 is resolved
         retries=2,
     ),
+    PackageSpec(
+        "python_modules/libraries/dagster-bigquery-pandas",
```
sryza, line 452:
```suggestion
        "python_modules/libraries/dagster-gcp-pandas",
```

python_modules/libraries/dagster-gcp/dagster_gcp/bigquery/io_manager.py
```diff
@@ -0,0 +1,226 @@
...
+                os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = temp_file_name
+
+        yield DbIOManager(
+            type_handlers=type_handlers,
+            db_client=BigQueryClient(),
+            io_manager_name="BigQueryIOManager",
+            database=init_context.resource_config["project"],
+            schema=init_context.resource_config.get("dataset"),
+        )
+        if init_context.resource_config.get("gcp_credentials"):
```
sryza, line 147: I think this should be in a finally block, right?

python_modules/libraries/dagster-gcp/dagster_gcp/bigquery/io_manager.py
```diff
@@ -0,0 +1,226 @@
...
+            os.environ.pop("GOOGLE_APPLICATION_CREDENTIALS", None)
+            os.remove(temp_file_name)
```
sryza, line 149: Why use `delete=False` and remove it here instead of putting the `yield` inside the `with` statement?

sryza (commented):
This is really awesome. Looks very close.

I think it's OK to require "schema" instead of "dataset" for now on the `Out` metadata, because most uses of this will be assets anyway.

python_modules/libraries/dagster-gcp/dagster_gcp/bigquery/io_manager.py
jamiedemaria: In most cases the user should be authing with gcp via the CLI `gcloud auth application-default login` or by setting `GOOGLE_APPLICATION_CREDENTIALS` to a file containing the key. But in cases like serverless, there's no way for the user to upload the key file to the container, so we need to let them provide creds via config. The `yield` is outside with `with` because the `with` only happens in the case when we're authing with credentials in the config 

I suppose we could do this instead:

```
        mgr = DbIOManager(
            type_handlers=type_handlers,
            db_client=BigQueryClient(),
            io_manager_name="BigQueryIOManager",
            database=init_context.resource_config["project"],
            schema=init_context.resource_config.get("dataset"),
        )
        temp_file_name = ""
        if init_context.resource_config.get("gcp_credentials"):
            with tempfile.NamedTemporaryFile("w") as f:
                f.write(init_context.resource_config.get("gcp_credentials"))
                temp_file_name = f.name
                os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = temp_file_name

                try:
                    yield mgr
                finally:
                    os.environ.pop("GOOGLE_APPLICATION_CREDENTIALS", None)

        return mgr

File: python_modules/libraries/dagster-gcp/dagster_gcp/bigquery/io_manager.py (hunk 2 above)
Comment (jamiedemaria) [reply python_modules/libraries/dagster-gcp/dagster_gcp/bigquery/io_manager.py:147]: yeah i tried that at first and remember getting an error about not being a context manager, but i tried it again and now it works! so it must have been actually caused by something else. I'll make it a finally, thanks for making me try it again

File: python_modules/libraries/dagster-gcp/dagster_gcp/bigquery/io_manager.py (hunk 3 above)
Comment (jamiedemaria) [reply python_modules/libraries/dagster-gcp/dagster_gcp/bigquery/io_manager.py]: tried this and having the yield of the manager inside the with statement means that the contents of the key isn't fully written to the file. I don't quite understand why yet, but basically it seems like the key is only "saved" to the file once the file is closed. so i think we need to keep the `delete=False` and manually delete later

File: python_modules/libraries/dagster-gcp/dagster_gcp/bigquery/io_manager.py [hunk 4]
```diff
@@ -0,0 +1,239 @@
...
+            database=init_context.resource_config["project"],
+            schema=init_context.resource_config.get("dataset"),
+        )
+        if init_context.resource_config.get("gcp_credentials"):
+            if os.getenv("GOOGLE_APPLICATION_CREDENTIALS") is not None:
+                raise DagsterInvalidDefinitionError(
+                    "Resource config error: gcp_credentials config for BigQuery IO manager cannot"
+                    " be used if GOOGLE_APPLICATION_CREDENTIALS environment variable is set."
+                )
+            with tempfile.NamedTemporaryFile("w+", delete=False) as f:
```
sryza, line 150: Something that strikes me as risky here is that the temporary file could get deleted after this block exits and before we yield.

Could we put the `yield` within the `with` block and use `f.flush` to make sure the contents get written out?

jamiedemaria: yeah flush worked!

sryza (approved):
NICE!!!!

PR #11427, opened 2022-12-29
Export RunRecord in the public API

gibsondan:
Summary:
Since we point users at get_run_records (and RunRecord has information not available on DagsterRun) I think we should export it.


OwenKephart (approved):
honestly thought this was already exported

prha (approved):

PR #11433, opened 2022-12-29
Eliminate unused config_or_config_fn arg in copy_for_configured

schrockn:
### Summary & Motivation

I am doing some further exploration in the configured system and doing value-add refactors as I find them. In this case the `config_or_config_fn` argument to copy_for_configured is unused and unnecessary, so I am eliminating it.

### How I Tested These Changes

BK

PR #11434, opened 2022-12-29
RFC: Change .configured to leverage default values instead of config mapping when dealing with raw config

schrockn:
### Summary & Motivation

This is a proposal to change how the config system works, in particular when `.configured` receives raw values and not a config mapping.

There is a current usability issue in the system that when `.configured` is used, the values essentially "disappear" from our tools. They cannot be overridden in the launchpad. They are undocumented. This ends up being a sort of worst-of-both-worlds situation where you have both the awkward API ergonomics of our config system, combined with the lack of traceability and parameterization of using vanilla Python.

Instead of having a configured values encoded in the config mapping system, this PR proposes that we jam them into the "default values" of config. In effect when you write `some_definition.configured({"foo":"bar"})` with this approach you are saying "make bar the default value for foo in this config schema."

If we combine this with some UI changes––notably automatically expanding default values in the Launchpad and expanding all defaults prior to persisting them in runs––we will have a lot more visibility. Users will have the option to see the entire config slug that (I believe this should be the default behavior) run will accept, including the defaults. And then users will see historical config slugs as well. This is more obvious, and also covers cases like when default values change over time.

This will also make it so that users can change the values they have configured for deployed code, if they so choose.

Additionally, our proposed UIs for the resources system can be implemented leverage this scheme. We can introspect the default values for a particular resource, process them, and then render them. 

BK

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/configurable.py
+++ b/python_modules/dagster/dagster/_core/definitions/configurable.py
@@ -6,10 +6,14 @@ from typing_extensions import Self
 from dagster import Field
 from dagster import _check as check
 from dagster._config import EvaluateValueResult
+from dagster._config.post_process import resolve_defaults
+from dagster._config.validate import validate_config
+from dagster._core.errors import DagsterInvalidConfigError
 
 from .definition_config_schema import (
     CoercableToConfigSchema,
     ConfiguredDefinitionConfigSchema,
+    DefinitionConfigSchema,
     IDefinitionConfigSchema,
     convert_user_facing_definition_config_schema,
 )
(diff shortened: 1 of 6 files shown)


schrockn:
* master
  * **PR #11433** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11433" target="_blank">Graphite</a> 
    * **PR #11434** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11434" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11434?utm_source=stack-comment).

schrockn:
Note: this is more of an RFC to get feedback on the general _approach_ rather than for commit. I think we need to spec this out a little more and think about all the uses cases. However I think this is a promising angle and I want to make sure that I am not missing anything.

python_modules/dagster/dagster/_core/definitions/configurable.py
```diff
@@ -63,6 +67,74 @@ def apply_config_mapping(self, config: Any) -> EvaluateValueResult:
...
+    )
+
+
+def _apply_defaults_to_schema_field(field: Field, additional_default_values: Any) -> Field:
+    # This work by validating the top-level config and then
+    # just setting it at that top-level field. Config fields
+    # can actually take nested values so we only need to set it
+    # at a single level
+
+    evr = validate_config(field.config_type, additional_default_values)
```
dpeng817, line 85: another nice benefit of this is that we'll now see errors with configured at definition time rather than during runtime.

dpeng817 (commented):
I'm super bullish on this idea, just seems like a huge improvement for 90% of use cases. 
A few questions:
- Would the existing config mapping behavior still be something that we espouse, or would that be hidden backcompat functionality?
- Do we force people to provide "full" config when using `configured`? Seems like it would be relatively straightforward to just validate the individual defaults provided instead of the full schema for the definition, and then folks could provide partial defaults (while we're at it, this approach could also be used to allow partial defaults for job defs).

schrockn:
> Would the existing config mapping behavior still be something that we espouse, or would that be hidden backcompat functionality?

I think at a minimum we should de-emphasize. I'd want to have more clarity around our long-term vision for config before committing to it. But my thinking is certainly in the direction of shifting the entire config system to be pydantic-native in our next major release, in which case deprecating config mapping would make sense.

> Do we force people to provide "full" config when using configured? Seems like it would be relatively straightforward to just validate the individual defaults provided instead of the full schema for the definition, and then folks could provide partial defaults (while we're at it, this approach could also be used to allow partial defaults for job defs).

Ah that sounds like a nice idea. Agree seems totally tractable.

schrockn:
@gibsondan I'd like your thoughts here as well. In particular I'm envisioning that along with this change we would also change the daemons to apply defaults to config and then persist the entire config slug with every run, rather than relying on default evaluation at job execution time. Curious re:

1) How hard that would be to implement and 
2) Any bad behavior this would cause. Any obvious this is that if config schema changed between persisting the config and the actual runtime, but anything else beyond that?

gibsondan:
when you say "change the daemons to apply defaults to config" - we're talking about like the scheduler and sensor daemons? Or when you kick off an ad hoc run from dagit and then it gets stuck in the queue? So i think the relevant hook here is the execution plan generation grpc call. 

We already lock it to a specific image when we do this so I'm not super worried about the config schema changing between run submit time and run launch time

gibsondan:
the relevant part of the codebase is here for the scheduler: https://github.com/dagster-io/dagster/blob/ab6ab928e2cb447291ada416e391e94083929b9a/python_modules/dagster/dagster/_scheduler/scheduler.py#L705-L741 

it takes the passed in run config, uses it to make an execution plan snapshot, and then creates a run with the passed in run config and the snapshot.

Seems like as part of that grpc call to make the execution plan snapshot you could also pass back some kind of resolved run config and either stick that on the run, or persist it separately on the snapshot so that you have both the user-specified run config and the resolved run config as needed.

Alex would probably have good thoughts here too when he is back on Tuesday, he is more familiar with execution plan stuff than I

But overall this seems very tractable to me

schrockn:
Got it. Yeah it looks like we have passed back the fully resolved dictionary in `ResolvedRunConfig` when we build the execution plan and then jam that fully resolved dictionary into the `ExecutionPlanSnapshot` that gets passed back over the rpc layer. Thanks for the pointer.

sryza:
I am overall very positive on this direction though.

schrockn:
> With the struct-based config, are we going to essentially be deprecating configured? If so, might we want to leave the old behavior for configured to minimize risk and just add the new behavior to struct-based config?

Yup totally reasonable. We are going to press on with using this strategy for new resources temporarily to facilitate some prototyping: https://github.com/dagster-io/dagster/pull/11470

> It could potentially be useful to render some sort of "config value lineage" in Dagit, to help people understand where their config is being set. One of the things I find most difficult about dealing with config in software is understanding where it came from. How would this fit with that?

What do you mean by config value lineage in this context? Definitely one part of the vision here is to be able to display configured values of resources along with those resources. Is that what you mean?

> Do we want some config to be non-override-able at runtime? If so, how would someone specify that?

Right now they wouldn't be able to specify that but definitely something we should consider.

> Is "default" still the right word for how we want to message this for users? Would "definition-level" or something be more accurate?

Especially with the proposal to have non-override-able config, default definitely makes less sense as a name.

I have to step away from a few minutes but these are great questions

sryza:
> What do you mean by config value lineage in this context? Definitely one part of the vision here is to be able to display configured values of resources along with those resources. Is that what you mean?

I mean helping people understand where their config values are being set. Some of the places where config can be set:
1. As the default value provided by a library resource.
2. A definition-time override of a default value provided by a library resource.
3. When launching a run.

So I'm wondering if we'd want to help users understand the distinction between (1) and (2). This is obviously speculative, but just trying to understand the space of potential impacts of a change like this.

schrockn:
These exact lineage issues came up in our prep for the user council meetin tmrw. We are going to surface lineage + the risk of having configured values be editable.

sryza (changes requested):

PR #11436, opened 2022-12-30
Allow AssetIn(dagster_type=Nothing)

sryza:
### Summary & Motivation

Prior, there was no way to set a partition mapping on a non-argument dep.

This enables:
```python
@asset(ins={"upstream": AssetIn(dagster_type=Nothing, partition_mapping=TimeWindowPartitionMapping())})
def asset1():
    ...
```

Something awkward about this approach is that it gets kind of unnecessarily hairy if you want to specify a multi-component asset key:

```python
@asset(ins={"upstream": AssetIn(key_prefix=["other_component"], dagster_type=Nothing, partition_mapping=TimeWindowPartitionMapping())})
def asset1():
    ...
```

I considered a couple other options too:
```python
@asset(non_argument_deps={AssetIn("upstream", partition_mapping=TimeWindowPartitionMapping())})
def asset1():
    ...
```
I found this one a little weird because the `dagster_type`, `input_manager_key`, and `key_prefix` parameters on the `AssetIn` wouldn't be applicable.

```python
@asset(non_argument_deps={AssetDep("upstream", partition_mapping=TimeWindowPartitionMapping())})
def asset1():
    ...
```
I kind of like this one, but was hesitant to introduce a new class. I figured we can always add this later.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
@@ -559,8 +559,11 @@ def build_asset_ins(
     all_input_names = set(non_var_input_param_names) | asset_ins.keys()
 
     if not has_kwargs:
-        for in_key in asset_ins.keys():
-            if in_key not in non_var_input_param_names:
+        for in_key, asset_in in asset_ins.items():
+            if in_key not in non_var_input_param_names and (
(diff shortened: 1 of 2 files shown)


OwenKephart (approved):
Looks reasonable to me

clairelin135 (approved):
This looks good to me! Agree it's better than introducing a new class at the moment--we can always do that in the future

PR #11437, opened 2022-12-30
Fix issues with a SerializableErrorInfo being coerced to a GraphenePythonError

gibsondan:
Summary:
I stumbled on a bit of a landmine on a separate PR by adding logic to GraphenePythonError that assumes that it's a GraphenePythonError - normally a valid assumption, but if you pass a namedtuple where Graphene is expecting a GrapheneObjectType, it appears that it coerces it into the object without ever calling the constructor. So we were calling GraphenePythonError methods on a SerializableErrorInfo.

I'll audit other callsites of GraphenePythonError for places we might have been doing this, but this is the only one i've found so far.

Test Plan: New test that would fail in master

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/schema/errors.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/errors.py
@@ -68,15 +68,45 @@ class GraphenePythonError(graphene.ObjectType):
     def __init__(self, error_info):
         super().__init__()
         self._error_info = check.inst_param(error_info, "error_info", SerializableErrorInfo)
-        self.message = error_info.message
-        self.stack = error_info.stack
-        self.cause = error_info.cause
-        self.context = error_info.context
-        self.className = error_info.cls_name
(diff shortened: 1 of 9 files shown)


python_modules/dagster-graphql/dagster_graphql/schema/errors.py
```diff
@@ -68,15 +68,45 @@ class Meta:
...
+        self._className = error_info.cls_name
+        self._cause = error_info.cause
+        self._context = error_info.context
+
+    def resolve_message(self, _graphene_info):
+        check.invariant(
+            isinstance(self, GraphenePythonError),
+            f"GraphenePythonError methods called on a {type(self)} - this usually indicates that "
+            "a SerializableErrorInfo was passed in where a GraphenePythonError was expected",
+        )
```
gibsondan, line 82: you would think this could be a helper method, but when this happens you can't use other methods!

prha (approved):
nice find.

PR #11441, opened 2022-12-30
[dagster-airflow] add `make_dagster_definitions_from_airflow_dags_path` and `make_dagster_definitions_from_airflow_dag_bag` apis

Ramshackle-Jamathon:
### Summary & Motivation

This pr adds definition based equivalents of the current repo apis. and refactors to using a shared `_make_schedules_and_jobs_from_airflow_dag_bag` function for both repo/definition user facing apis

- added new airflow 1/2 unit tests that convert and execute dags.


python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
```diff
@@ -79,3 +85,124 @@ def make_dagster_job_from_airflow_dag(
...
+        use_airflow_template_context,
+        mock_xcom,
+        use_ephemeral_airflow_db,
+        connections,
+    )
+
+    return Definitions(
+        schedules=repo.schedule_defs,
+        jobs=repo.get_all_jobs(),
+    )
```
schrockn, line 208: I think this would feel a lot clearer and safer it `make_dagster_repo_from_airflow_dags_path` was actually `make_schedules_and_jobs_from_airflow_dags_path` as the core workhorse function. A few reasons:

* It's very surprising that only schedules and jobs get shuffled along here.
* And then inspecting the code it calls it actually sort of purports to create assets and doesn't. 
* Finally the way this all implemented we construct a repo, then pass to Definitions, which in turn does some processing, then constructs another repository which I think could cause some unexpected bugs long term.

So to summarize I recommend making a raw workhouse function that produces definitions "outside" a repository or a `Definitions` object, and then only pass those to the enclosing container (repository/Definitions) at the highest possible level in the callstack.

schrockn (changes requested):

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
Ramshackle-Jamathon: Fair points, I think keeping the declaration of `Definitions` in user-code will help make things a bit clearer? If theres a way to derive Definitions from a Repo that might also be interesting?

schrockn (approved):
Use but I think you should implement the existing function to create a repository using this function as its implementation and then add a definitions version that invokes this function. 

make_schedules_and_jobs_from_airflow_dags_path should be an implementation detail and I don't think it should be exported in __init__.py

schrockn (approved):

Issue #10840, opened 2022-12-01
Handling escaped characters in asset key names / elsewhere in the UI


trevorcampbell:
### Dagster version

1.1.3

### What's the issue?

Some escaped characters in asset key names (and maybe elsewhere too) are not shown in the UI. So this can lead to some very confusing situations -- e.g. where a static partition has 4 keys that look identical in the UI.

I ran into this bug when I had a string processing error in some other code that added a newline to the end of one of my keys. In the UI it looked like the same key, but the materialization seemed to "disappear" (because the key was actually new).

I think it was @clairelin135 who helped me track this down (assuming that's `@claire` on slack)

See below for the minimal reproducible example.

### What did you expect to happen?

Either an error should be thrown to prevent certain escaped special characters in key names, or the raw string should be displayed (showing the literal escape character).

### How to reproduce?

Create the tutorial scaffold project folder, and then create a single file `tests.py` in the assets folder:

```
from dagster import asset, StaticPartitionsDefinition
import os

@asset(
    partitions_def=StaticPartitionsDefinition(['a', 'a\n', 'a\n\n', 'a\n\n\n'])
)
def submissions(context):
    context.log.info(context.partition_key.keys_by_dimension)
```

Then the UI will display 4 asset keys that appear to be identical.

### Deployment type

Local

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

dpeng817:
we really ought to use the same rules for partition keys as we do for other keys in the system (op names, asset keys, resource keys, etc), and I'm honestly very surprised that we don't... cc @sryza is this intentional

sryza:
@dpeng817 I do think we probably should have done that, but I think it's probably too late at this point. I would support banning newlines from partition_keys though. I can't imagine that anyone is relying on that. Thoughts @clairelin135 ?

clairelin135:
Agree that it's probably too late to ban these characters.

I'm in favor of either:
- Updating Dagit to display the `repr` of partition keys, which will display escaped characters. If anybody happens to accidentally be using escaped characters this will prevent any breakages
- Removing all escape characters that don't display in Dagit (`\n`, `\r`, `\t`, `\f`, `\v`, `\0`) if we think nobody has any of these characters in their partition keys currently

sryza:
@clairelin135 let's go with the latter

PR #11443, opened 2022-12-30
Ban invalid escape chars

clairelin135:
Resolves https://github.com/dagster-io/dagster/issues/10840

In static partitions definitions, ban any escape characters that may not be displayed.


sryza (approved):

python_modules/dagster/dagster/_core/definitions/partition.py
```diff
@@ -278,10 +283,16 @@ class StaticPartitionsDefinition(
...
+        for partition_key in partition_keys:
+            found_invalid_substrs = [
+                invalid_substr
+                for invalid_substr in INVALID_PARTITION_SUBSTRINGS
+                if invalid_substr in partition_key
+            ]
+            if found_invalid_substrs:
+                raise DagsterInvalidDefinitionError(
+                    f"{found_invalid_substrs} are invalid substrings in a partition key"
+                )
```
gibsondan, line 295: commented in Slack but can we consider making this a warning until 1.2? Could we include the name of the offending key(s) in the error or warning as well so that people know where to look to resolve this?

PR #11446, opened 2022-12-31
Make all arguments on DagsterInstance.create_run keyword-only

schrockn:
### Summary & Motivation

In the course of doing some work to prototype a different treatment of `configured` I ran into an old friend, `DagsterInstance.create_run`.

`DagsterInstance.create_run` is in rough shape. It has built up over time with incremental features. Every time a feature is added that involves run creation, an optional parameter is added and then code has to be added at all callsites in an error prone fashion. If you look at the different production callsites of .create_run they all orchestrate similar interactions with code location:

* Create an external pipeline that corresponds to the relevant selection--determined by either solid_selection or asset_selection
* The creation of of an execution plan snapshot. If the call site supports re-execution, the computation of the step keys to execute and the creation of a `KnownExecutionState` object.
* The merging of tags specific to the callsite and the tags from the pipeline to create the tags for a run.  

And in the end they all have orchestrate these interactions in order to pass the correct 16 (!!!) arguments to create_run.

There is a ton of inconsistency between the call sites which is undoubtedly hiding bugs. There is also a completely different codepath for creating runs in an in-process context.

The plan:

* Tighten and harden the existing create_run method with type annotations, typehints, and invariants.
* Create a new function, persist_new_production_run (we can discuss name in the appropriate diff) that will consolidate all the interactions between the host process and then repository location and the associated logic therein.
* Use that new function in the major production callsites (CLI, graphql, sensor, schedule, backfills). This will typically involve refactoring the call site to have the same behavior as the new function, and then finally moving our to the new function.
* Change `create_run_for_pipeline` (the in-process cases) to use the production version to unify the codepaths.

The first step is requiring that create_run take keyword arguments. This will set up the elimination of default values, which will make the refactoring the callsites of create_run more straightforward.

BK


schrockn:
* master
  * **PR #11446** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11446" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11446?utm_source=stack-comment).

alangenfeld (approved):

PR #11467, opened 2023-01-03
Hook for Cloud dagit into PythonErrorInfo

salazarm:
### Summary & Motivation

Cloud dagit needs to augment PythonErrorInfo

Screen Shot 2023-01-03 at 9 56 03 AM


js_modules/dagit/packages/core/src/app/PythonErrorInfo.tsx
```diff
@@ -33,9 +34,15 @@ export const PythonErrorInfo: React.FC<IPythonErrorInfoProps> = (props) => {
...
+  const MaybeMissingEnvVarErrorHeader = useLaunchPadHooks().MaybeMissingEnvVarErrorHeader;
+
   return (
     <>
-      {context}
+      {MaybeMissingEnvVarErrorHeader ? (
+        <MaybeMissingEnvVarErrorHeader error={props.error} fallback={context} />
+      ) : (
+        context
+      )}
```
hellendag, line 45: I don't think we should refer to Cloud features in OSS. Maybe a new prop could be added to `PythonErrorInfo` that would allow passing this along to be rendered from the callsite?

hellendag (changes requested):

js_modules/dagit/packages/core/src/app/PythonErrorInfo.tsx
salazarm: I can generalize the name here like `CustomErrorHeader`

hellendag (approved):

Issue #9755, opened 2022-09-21
Allow multiprocess max_concurrent to be defined at an OP level


AlejandroUPC:
### What's the use case?

I have an @op that generates some dynamic output, the rest of the pipeline can be run at 15 ops at a time in parallel from here, but at some point in the pipeline there is an @op (which all the outputs must go through) that has a call to an API, which can't take x15 requests at a time.

It can handle a concurrency of 3, but anything beyond that kills the API.

I could set the config to run just on three multiprocesses but makes the rest of the job super slow, because a single-op bottle neck

```yaml
execution:
  config:
    multiprocess:
      max_concurrent: 3

```

### Ideas of implementation

Ideally, I would like to be able to define the max_concurrencies at an @op level, maybe something on the config_schema like

```python
@op(config_schema={'max_concurrent':8})
def very_easy_parallel_function():
    pass
 

@op(config_schema={'max_concurrent':2})
def this_function_must_run_slower_in_parallel():
    pass
```

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

OwenKephart:
Hi @AlejandroUPC ! Thanks for the suggestion -- I'm linking this issue to https://github.com/dagster-io/dagster/issues/8525, as I believe the functionality described in that issue would work for your desired use case. It's definitely on our radar as something we'd like to support :)

Issue #8525, opened 2022-06-21
Tag concurrency limits for multiprocess_executor (and other executors)


gibsondan:
## Use Case
Being able to configure an executor to run at most N ops with a given tag at once for a given run (much like how the run queue can be configured). Eventually being able to apply these rules globally across all runs, Celery-style, would be great - but in the short term being able to configure them within a single run could be built quickly and would satisfy a number of use cases.

## Additional Info

---
#### Message from the maintainers:
Excited about this feature? Give it a :thumbsup:. We factor engagement into prioritization.

gibsondan:
What this likely looks like in practice:

- Adding additional tag-based config fields to the executor definition https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_core/definitions/executor_definition.py#L336-L343 that are the same as the tag_concurrency_limits field here (ideally sharing the code for the field): https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_core/run_coordinator/queued_run_coordinator.py#L81-L104

- Mirroring the logic here from the run dequeuer: https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_daemon/run_coordinator/queued_run_coordinator_daemon.py#L148-L159

in the logic that picks which steps to execute: https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_core/execution/plan/active.py#L242-L256

Which is called from here: https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_core/executor/multiprocess.py#L195-L198

(that ActiveExecution class should already have access to tags for each step)

Even better if the logic for "take tagged thigns, apply rules to them, and return the correct tagged things to run" can be shared between runs and steps

PR #11472, opened 2023-01-03
Add tag_concurrency_limits config to executors

gibsondan:
Resolves https://github.com/dagster-io/dagster/issues/9755 and https://github.com/dagster-io/dagster/issues/8525


python_modules/dagster/dagster/_core/executor/multiprocess.py
```diff
@@ -196,10 +198,11 @@ def execute(
                         for key, event in term_events.items():
                             event.set()
 
-                    # start iterators
-                    while len(active_iters) < limit and not stopping:
+                    while not stopping:
                         steps = active_execution.get_steps_to_execute(
-                            limit=(limit - len(active_iters))
+                            max_num_to_return=limit - len(active_iters),
```
gibsondan, line 203: not positive we need to set this here

python_modules/dagster/dagster/_utils/tags.py
```diff
@@ -0,0 +1,84 @@
+from collections import defaultdict
+from typing import Dict, Tuple
+
+from dagster import _check as check
+
+
+class TagConcurrencyLimitsCounter:
```
gibsondan, line 7: unchanged other than method names to not mention runs

python_modules/dagster/dagster_tests/execution_tests/engine_tests/test_step_delegating_executor.py
```diff
@@ -258,6 +258,36 @@ def test_max_concurrent():
...
+    with instance_for_test() as instance:
+        result = execute_pipeline(
+            reconstructable(three_op_job),
+            instance=instance,
+            run_config={
+                "execution": {
+                    "config": {
+                        "tag_concurrency_limits": [{"key": "database", "value": "tiny", "limit": 1}]
+                    }
+                }
```
alangenfeld, line 272: nit: explicitly set max_concurrent to 2+

alangenfeld (approved):
nice, ez pz

python_modules/dagster/dagster/_utils/tags.py
alangenfeld: might be nice to add something like `ITagged` so this all can be typed more accurately

PR #11473, opened 2023-01-03
[dagster-dbt] add DbtManifestAssetSelection

OwenKephart:
### Summary & Motivation

dbt has a very expressive selection syntax, which people might want to use to describe sections of their overall asset graph. This PR adds the ability to do that.

This only works if you're using `load_assets_from_dbt_manifest` because a) if you don't have a compiled manifest available to you, you'll need to run `dbt ls` which can take a long time and b) if you have a large enough dbt graph to make you want to describe things in terms of dbt selections, you probably should be using `load_assets_from_dbt_manifest` anyway.


sryza (approved):
Nice

PR #11480, opened 2023-01-03
Adds OSS email collection dialog

salazarm:
### Summary & Motivation
https://linear.app/elementl/issue/CLOUD-2084/oss-email-collection-popup

Screen Shot 2023-01-03 at 4 48 03 PM


js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
```diff
@@ -0,0 +1,145 @@
...
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        email,
+        newsletter,
+      }),
+    });
+    dismiss();
+  };
```
hellendag, line 60: Does this POST to `dagster.cloud` actually work? I kind of feel like this should be a GraphQL mutation, instead of posting to another domain.

js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
```diff
@@ -0,0 +1,145 @@
...
+    />
+  );
+};
+
+const CommunityNuxImpl: React.FC<{dismiss: () => void}> = ({dismiss}) => {
+  const [shouldShowNux, setShouldShowNux] = React.useState(false);
+  React.useEffect(() => {
+    const timeout = setTimeout(() => {
+      setShouldShowNux(true);
+    }, 5000); // wait 5 seconds before showing Nux
```
hellendag, line 39: Five seconds seems almost too long, like it might be jarring to do stuff for a few seconds and only then see the dialog.

js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
```diff
@@ -0,0 +1,145 @@
...
+        <Box flex={{direction: 'column', justifyContent: 'stretch', gap: 12}}>
+          <div>Notify me about Dagster security updates</div>
+          <TextInput
+            value={email}
+            onChange={(e) => {
+              setEmail(e.target.value);
+              setEmailChanged(true);
+            }}
+            placeholder="hello@dagster.io"
+            strokeColor={!emailChanged || validEmail ? undefined : Colors.Red500}
```
hellendag, line 101: Am I understanding this correctly that while I'm typing my email address, it will appear to be in an error state until it is valid? If so, can we wait until submission to show the error state? It always feels a little weird to have an error state on something like this when I'm still typing.

js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
```diff
@@ -0,0 +1,145 @@
...
+            style={{width: '100%'}}
+          />
+          {emailChanged && !validEmail ? (
+            <div style={{paddingBottom: '12px', color: Colors.Red500, fontSize: '12px'}}>
+              Add your email to get updates from Dagster.
+            </div>
+          ) : null}
+          <Box flex={{direction: 'column', gap: 8}} padding={{bottom: 24}}>
+            <Box as="label" flex={{direction: 'row', gap: 8, alignItems: 'center'}}>
+              <Checkbox checked={true} disabled={true} onChange={() => {}} />{' '}
```
hellendag, line 111: - Impossible to uncheck?
- Shouldn't need the space, given the `gap`

js_modules/dagit/packages/app/src/index.tsx
```diff
@@ -70,6 +71,7 @@ ReactDOM.render(
     </AppTopNav>
     <App>
       <ContentRoot />
+      <CommunityNux />
```
hellendag, line 74: Nice

js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
salazarm: Yes, we want this data to go to the cloud endpoint so that we don't need to give postgress write access to another server.

js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
salazarm: Josh suggested 5 seconds. I think it's reasonable, though I do agree it can be jarring. The good news is that its not like a full page dialog or anything so it shouldn't interfere with you doing something

js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
salazarm: Yeah that's the correct behavior. I prefer this tbh, I think its a worse experience to click submit and then find out you made an error.

js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
salazarm: Yes its supposed to stay checked.
I'll remove the space it must have been added automatically or something idk

hellendag (approved):

js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
petehunt: this should be `https` right?

salazarm:
@hellendag Updated the PR to instead point to an iframe hosted on the dagit-website.

This PR adds the iframe and backend: https://github.com/dagster-io/dagster-website/pull/380

salazarm:
In a follow-up I'll just bypass blueprints dialog component and use a custom one so that we can preload the iframe in the correct spot

PR #11481, opened 2023-01-03
add run filter for updated before

prha:
### Summary & Motivation
For some internal stat keeping, we want to track runs stuck in a starting state that haven't been updated in the last X seconds.

BK


python_modules/dagster/dagster/_core/storage/runs/sql_run_storage.py
```diff
@@ -262,6 +262,9 @@ def _add_filters_to_query(self, query, filters: RunsFilter):
         if filters.updated_after:
             query = query.where(RunsTable.c.update_timestamp > filters.updated_after)
 
+        if filters.updated_before:
+            query = query.where(RunsTable.c.update_timestamp < filters.updated_before)
+
         if filters.created_before:
             query = query.where(RunsTable.c.create_timestamp < filters.created_before)
```
gibsondan, line 269: could throw in created_after while you're inhere

gibsondan (approved):

PR #11488, opened 2023-01-04
Update PartitionsSubset Serialization

clairelin135:
Following the decision to represent time window partition statuses in ranges in Dagit, this PR makes the following changes:
- For all partitions subsets:
   - Add `get_inverse_subset` methods to partitions subsets, allowing for performantly fetching the unmaterialized partition subsets from materialized subsets
   - Adds versions to the serializations. This allows for deserializing older versions of serialized subsets.
   - Adds a `can_deserialize` method, allowing for checking if a serialized subset can still be deserialized. For the partition status cache, rebuilds partition status from storage if the version is no longer supported.
- For `TimeWindowPartitionsSubset`s:
   - Restructure each stored time window to also store the number of partitions within, so Dagit can directly render the partition bar without searching for the length of each time window.
   - Performantly create `TimeWindowPartitionsSubset`s assuming that there are fewer than 10 ranges of materialized partitions
- Restructures the `MultiPartitionsSubset`. In Dagit, the "primary" key will be the static dimension (as many users want to view partition status by a static partition key). Accordingly, `MultiPartitionsSubset` now maps static partition keys to their corresponding `TimeWindowPartitionsSubset`s.

One decision in this PR is to not serialize the "unmaterialized" partition subsets. With the assumption that there will be no more than 10 ranges in a time window partitions definition, calculating the inverse should be fairly fast. I'm curious how this will perform for users, especially for the multidimensional case where the inverse must be calculated for each static partition. If users report this being slow, I'd advocate to switching to iterating through all time windows (versus partition keys) to speed up this behavior.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
@@ -48,6 +48,20 @@ class TimeWindow(NamedTuple):
     end: PublicAttr[datetime]
 
 
+class SubsetTimeWindow(NamedTuple):
+    """An interval that is closed at the start and open at the end.
+
+    Attributes:
+        start (datetime): A pendulum datetime that marks the start of the window.
+        end (datetime): A pendulum datetime that marks the end of the window.
+        length (int): The number of partitions included in the window.
+    """
+
+    start: PublicAttr[datetime]
+    end: PublicAttr[datetime]
+    length: int
+
+
 class TimeWindowPartitionsDefinition(
     PartitionsDefinition[TimeWindow],  # pylint: disable=unsubscriptable-object
     NamedTuple(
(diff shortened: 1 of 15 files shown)


python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -194,6 +196,45 @@ def start_time_for_partition_key(self, partition_key: str) -> datetime:
...
+            key=lambda dt: dt.timestamp(),
+        )
+        curr_key_i = 0
+        partition_key_time_windows = []
+        for time_window in (
+            self.iterate_valid_time_windows(self.start)
+            if not _valid_time_windows
+            else _valid_time_windows
+        ):
+            if time_window.start.timestamp() == sorted_key_datetimes[curr_key_i].timestamp():
```
sryza, line 227: I don't think doing this comparison will always work, because sometimes a partition named "2020-01-01" will correspond to a window timestamp of that day at 7 pm.

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -194,6 +196,45 @@ def start_time_for_partition_key(self, partition_key: str) -> datetime:
...
+                [
+                    pendulum.instance(datetime.strptime(partition_key, self.fmt), tz=self.timezone)
+                    for partition_key in partition_keys
+                ]
+            ),
+            key=lambda dt: dt.timestamp(),
+        )
+        curr_key_i = 0
+        partition_key_time_windows = []
+        for time_window in (
```
sryza, line 222: As mentioned on Slack, my concern here is that iterating over all the valid time windows will be O(total number of partitions). Ideally this function would only be O(`len(partition_keys`)).

Here's an idea of how to accomplish that without invoking `cron_string_iterator` on every partition key: aim to invoke `cron_string_iterator` once for every contiguous range of partition keys in `partition_keys`.

Something like:
```python
        cur_windows_iterator = windows_iterator_for_partition_key(sorted_partition_keys[0])
        for partition_key in sorted_partition_keys:
            next_window = next(cur_windows_iterator)
            if next_window.start.strftime(partitions_def.fmt) == partition_key:
                partition_key_time_windows.append(next_window)
            else:
                cur_windows_iterator = windows_iterator_for_partition_key(partition_key)
                partition_key_time_windows.append(next(cur_windows_iterator))
```

clairelin135: I tried out this approach.

On a hourly partitions definition with 100K keys, if i randomly select:
- 1/2  of the partitions to be in the subset it takes 60s to build 
- 7/8 of the partitions to be in the subset it takes 17s to build

With the existing implementation:
- 1/2 of the partitions to be in the subset it 4s to build
- 7/8 of the partitions it takes 6s to build

From the flame graph it seems like most of the time is spent creating the croniter object in the approach you mentioned. my hypothesis is that if a time window iterator has to be created more than ~500 times (e.g. if there are more than 500 ranges in the partition keys) then it will be more performant to do a single scan of all the partition keys rather than create a croniter object for each range.

Image attached is the flamegraph for 1/2 partitions materialized in the subset
1_in_2_partitions

sryza: This is a very cool investigation. It's really amazing how slow croniter is.

> my hypothesis is that if a time window iterator has to be created more than ~500 times (e.g. if there are more than 500 ranges in the partition keys) then it will be more performant to do a single scan of all the partition keys rather than create a croniter object for each range.

I believe that it's very rare to see a real-life partitions subset with this number of ranges. In most cases, there's just a single range, even when there are thousands of partitions. I think it's likely not worth the effort of maintaining code paths that are optimized for more than like 10 ranges.

clairelin135: got it--I also agree that in practice people will try to make sure all their partitions are materialized (or at least have large ranges). I figure if users complain that loading the subsets is really slow then we can add some perf improvements after.

sryza (commented):
Would it make sense to put the storage changes in a separate PR from the time window perf changes?

sryza (commented):
And potentially separate out the multi-partitions changes as well?

clairelin135:
@sryza that sounds good! I can break the changes out into multiple PRs

PR #11498, opened 2023-01-04
[dagster-airflow] warn on airflow dataset

Ramshackle-Jamathon:
### Summary & Motivation
dagster-airflow currently will not create SDAs from airflow datasets, it could in the future but there's a bit of complexity in implementing it (each dag can produce and depend on N number of datasets, hard to map cleanly) and for such a new feature its unclear that migrating teams will need it. So for now dagster-airflow is not able to migrate the data-aware schedules for datasets but it is able to migrate the actual dag that produces them. In the case a user is using an airflow dataset it would be possible for them to layer ontop their own schedule/sensor for running that (if we still haven't implemented a clean transform).

This pr adds a warning that will log whenever a airflow dataset is encountered.

### DIFF
--- a/python_modules/libraries/dagster-airflow/dagster_airflow/dagster_pipeline_factory.py
+++ b/python_modules/libraries/dagster-airflow/dagster_airflow/dagster_pipeline_factory.py
@@ -178,7 +178,6 @@ def make_dagster_repo_from_airflow_dag_bag(
 
     job_defs = []
     schedule_defs = []
-    asset_defs = []
     count = 0
     # To enforce predictable iteration order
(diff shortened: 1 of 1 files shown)


Ramshackle-Jamathon:
https://elementl-workspace.slack.com/archives/C03UTTB4TEJ/p1672851915521009 also came up that this looks like a public api, its not, and is not documented or exported  at the top module level atm but we might also want to `_` prefix it to make that clear?

currently its just used by `make_dagster_repo_from_airflow_dag_bag`

schrockn (changes requested):
Why even have this function or call it if it doesn't do anything?

schrockn (approved):

PR #11500, opened 2023-01-04
[dagit] RFC: Start moving GraphQL out to separate files

hellendag:
### Summary & Motivation

Move some GraphQL documents out to separate files, just to see how it looks/feels.

Benefits:

- No need for the `graphql()` function, which I'm a little worried will be an annoying import that will be easy to get wrong.
- More consistency around how fragments and query/mutation documents are defined. Now you definitely don't need to define a `const` for these, and you don't need to worry about how/whether to export them.
- Less reason to worry about weird import issues when queries/mutations are used in multiple places.
- To import the query/mutation doc, you import it from `../graphql/graphql`, same as type imports.
- Cloud will be less likely to accidentally pull in a GraphQL document from OSS.
- Subjectively feels a little cleaner overall?

Drawbacks:

- It's yet another file.
- The GraphQL documents no longer live exactly in the same implementation file as their usage.

I also installed the GraphQL Language Feature Support vscode extension to get syntax highlighting and autocomplete in queries/fragments, and it seems pretty nice. The `.graphqlrc.yml` is added to support that.

Thoughts on this?

Buildkite, sanity check Dagit.


hellendag:
* master
  * **PR #11500** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11500" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11500?utm_source=stack-comment).

salazarm (approved):

hellendag:
Uncovered a dealbreaker: jump-to-definition in vscode is useless, since we can't get to the query file itself from the callsite. This is going to be a serious problem when people want to edit or even just view queries.

Discussed a bit in Slack, going to go ahead and close this since it was speculative anyway.

PR #11506, opened 2023-01-04
only count materializations within backfill

sryza:
### Summary & Motivation

Asset backfills look at what asset partitions have already been materialized to decide what partitions need to be submitted.

Prior to this change, the backfill code would look at all materializations of the asset, independent of whether they were part of the backfill. This meant that any materialization prior to the backfill would stop the backfill from submitting runs.

This change filters to only materializations that were part of the backfill. It functions by adding a backfill tag to the relevant materializations

An alternative way to implement this would have been to look for the backfill tag on the runs, instead of on the asset materializations. This would have required joining the event log with the run tags table.

Added a test that fails before the change

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/asset_backfill.py
+++ b/python_modules/dagster/dagster/_core/execution/asset_backfill.py
@@ -28,6 +28,7 @@ from dagster._core.host_representation.selector import PipelineSelector
 from dagster._core.instance import DagsterInstance
 from dagster._core.storage.pipeline_run import DagsterRunStatus, RunsFilter
 from dagster._core.storage.tags import BACKFILL_ID_TAG, PARTITION_NAME_TAG
+from dagster._core.utils import frozendict
 from dagster._core.workspace.context import BaseWorkspaceRequestContext
 from dagster._utils.caching_instance_queryer import CachingInstanceQueryer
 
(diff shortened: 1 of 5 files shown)


prha:
Need to make sure this is guarded against the schema migrations req'd for asset tags.  Previously, this was gated on users who wanted to use asset partitions (new feature), while this will affect all materializations within backfilled jobs.

prha (changes requested):
This needs guards against various stages of migration state, and probably some tests for that.

sryza:
I added a PR that makes event tag filtering work both pre- and post- migration: https://github.com/dagster-io/dagster/pull/11509.

python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -315,7 +316,9 @@ def execute_asset_backfill_iteration_inner(
     recently_materialized_asset_partitions = AssetGraphSubset(asset_graph)
     for asset_key in asset_backfill_data.target_subset.asset_keys:
         records = instance_queryer.get_materialization_records(
-            asset_key=asset_key, after_cursor=asset_backfill_data.latest_storage_id
+            asset_key=asset_key,
+            after_cursor=asset_backfill_data.latest_storage_id,
+            tags=frozendict({BACKFILL_ID_TAG: backfill_id}),
```
sryza, line 321: Depends on https://github.com/dagster-io/dagster/pull/11509 to work in the unmigrated case

prha (approved):

PR #11509, opened 2023-01-04
enable filtering by asset tag when asset tags table is not present

sryza:
### Summary & Motivation

Prior to this change, you get a `DagsterInvalidInvocationError` if you try to pass `get_event_records` an `EventRecordsFilter` with the `tags` attr set.

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
+++ b/python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
@@ -803,13 +803,8 @@ class SqlEventLogStorage(EventLogStorage):
         if event_records_filter.storage_ids:
             query = query.where(SqlEventLogStorageTable.c.id.in_(event_records_filter.storage_ids))
 
-        if event_records_filter.tags:
-            if not self.has_table(AssetEventTagsTable.name):
-                raise DagsterInvalidInvocationError(
-                    "Cannot filter by asset event tags because AssetEventTags table does not "
-                    "exist. Run `dagster instance migrate` to create the table."
-                )
-
+        if event_records_filter.tags and self.has_table(AssetEventTagsTable.name):
+            # If we don't have the tags table, we'll filter the results after the query
             check.invariant(
                 isinstance(event_records_filter.asset_key, AssetKey),
                 "Asset key must be set in event records filter to filter by tags.",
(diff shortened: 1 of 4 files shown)


sryza:
@prha I updated the PR to `DagsterInvalidInvocationError` when a limit is supplied and the asset tags table is not present. I plan to fill that gap in a followup that won't make it into this week's release.

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -905,10 +900,24 @@ def get_event_records(
                         "Could not resolve event record as EventLogEntry for id `%s`.", row_id
                     )
                     continue
-                else:
-                    event_records.append(
-                        EventLogRecord(storage_id=row_id, event_log_entry=event_record)
-                    )
+
+                if event_records_filter.tags and not self.has_table(AssetEventTagsTable.name):
```
prha, line 904: is the indentation off on this block?

sryza: Mind elaborating? I don't think so, but maybe I wrote it in a more confusing way than I needed to?

prha: oh, you're right... didn't see this was part of a different function.

prha: I think you still need to add a check in line 867:

```
        if event_records_filter.tags and not self.supports_intersect and self.has_table(AssetEventTagsTable.name):
```

sryza: You are right. Thanks for catching.

prha (approved):

sryza:
Oops accidentally re-requested after Phil approved. No changes since his approval.

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -905,10 +904,24 @@ def get_event_records(
...
+                    # If we can't filter tags via the tags table, filter the returned records
```
clairelin135, line 909: Can we add comments to this function highlighting that any tags that do not exist on the materialization event object will not be returned?

I think in `add_asset_event_tags` it would also be helpful to comment that adding tags after the materialization already exists should be done with caution (as this `get_event_records` call won't find tags outside of the materialization event if the asset event tags table doesn't exist)

clairelin135 (approved):

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
sryza: Would it ever be possible to end up in that situation? It looks like `add_asset_event_tags` will fail when there's no asset event tags table.

clairelin135: ah gotcha, thought that it would silently exit in that case. this is looking fine to me then!

PR #11510, opened 2023-01-04
Consider the run worker unhealthy is the job has no active pods but the run is in a non-terminal state

gibsondan:
Summary:
A user reported an issue where run worker crash recovery didn't kick in for a STARTED run - there are cases in which there aren't and failed pods but the job is clearly no longer running, that should also trigger run monitoring

### DIFF
--- a/python_modules/libraries/dagster-k8s/dagster_k8s/launcher.py
+++ b/python_modules/libraries/dagster-k8s/dagster_k8s/launcher.py
@@ -367,12 +367,19 @@ class K8sRunLauncher(RunLauncher, ConfigurableClass):
     def supports_check_run_worker_health(self):
         return True
 
+    @property
+    def supports_run_worker_crash_recovery(self):
+        return True
+
     def check_run_worker_health(self, run: DagsterRun):
         container_context = self.get_container_context_for_run(run)
 
-        job_name = get_job_name_from_run_id(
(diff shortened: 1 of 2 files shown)


python_modules/libraries/dagster-k8s/dagster_k8s/launcher.py
```diff
@@ -379,6 +379,14 @@ def check_run_worker_health(self, run: DagsterRun):
             return CheckRunHealthResult(
                 WorkerStatus.UNKNOWN, str(serializable_error_info_from_exc_info(sys.exc_info()))
             )
+
+        # If the run has STARTED but there are no running pods, something went wrong
+        # Are there any other statuses with this property?
```
johannkm, line 384: maybe CANCELLING as well?

johannkm (changes requested):
I think it would be hard to make a non finicky e2e test for this since at least in the known cases, it's a race before k8s can create a new pod. A unit test seems adequate.

This seems like a good invariant to check on, but per the above it's not clear if it's adequate for detecting the node eviction behavior? I'm at a bit of a loss for how this was supposed to work, given the behavior we saw about the old pod just disappearing. I haven't come up with something bullet proof yet other than the non k8s job approach

johannkm (approved):

PR #11511, opened 2023-01-04
Fix maxResumeRunAttempts being explicitly enabled even if you set it to 0

gibsondan:
Summary:
This fix makes maxResumeRunAttempts equal to 0 when it is explicitly set to 0.

I feel a little weird about setting this experimental feature on by default when you turn on run monitoring in general, but changign that would be a breaking change that  can probably wait until 1.2 (where I think we should turn run monitoring on by default but this feature off by default until we have more confidence in it)

### DIFF
--- a/helm/dagster/templates/configmap-instance.yaml
+++ b/helm/dagster/templates/configmap-instance.yaml
@@ -90,7 +90,7 @@ data:
     run_monitoring:
       enabled: {{ $runMonitoring.enabled }}
       start_timeout_seconds:  {{ $runMonitoring.startTimeoutSeconds }}
-      {{- if $runMonitoring.maxResumeRunAttempts }}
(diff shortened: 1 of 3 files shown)


helm/dagster/templates/configmap-instance.yaml
```diff
@@ -90,7 +90,7 @@ data:
     run_monitoring:
       enabled: {{ $runMonitoring.enabled }}
       start_timeout_seconds:  {{ $runMonitoring.startTimeoutSeconds }}
-      {{- if $runMonitoring.maxResumeRunAttempts }}
+      {{- if not (kindIs "invalid" $runMonitoring.maxResumeRunAttempts) }}
```
johannkm, line 93: ah, helm

johannkm (approved):
thanks for the fix

PR #11512, opened 2023-01-04
[For 1.2] Change default run monitoring settings

gibsondan:
Summary:
Proposed run monitoring changes starting in 1.2
- on by default for starting runs and failing hanging started runs
- experimental run recovery off by default

Will wait to land this until 1.2 is cut in the distant future


johannkm (approved):
Great. I think we could also move `maxResumeRunAttempts` to deprecated, replaced by runRetries. The one missing feature would be to offer retries only on system failures and not user code exceptions.

PR #11516, opened 2023-01-05
stale asset materialization for schedules and sensors

smackesey:
### Summary & Motivation

Add the ability to materialize stale assets from a schedule or sensor. This is implemented via `RunRequest`.

- Add `stale_assets_only` flag to `RunRequest`
- Port projected logical version loader logic from dagster-graphql to core, where it is used in the `scheduler`

This approach was used because there is no way to access all external repositories in the workspace, which are required to resolve the staleness status of assets, from a repo process. Therefore we have to pass a flag (`stale_only`) back to the host process.

- Add unit tests alongside existing scheduler and sensor tests


smackesey:
* master
  * **PR #11516** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11516" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11516?utm_source=stack-comment).

python_modules/dagster/dagster/_scheduler/stale.py
```diff
@@ -0,0 +1,57 @@
...
+
+import dagster._check as check
+
+def resolve_asset_selection(context: WorkspaceProcessContext, run_request: RunRequest) -> Optional[Sequence[AssetKey]]:
+    if run_request.stale_only:
+        asset_selection = check.not_none(run_request.asset_selection)
+        stale_keys: List[AssetKey] = []
+        request_context = context.create_request_context()
+        repositories = get_repositories(request_context)
+        key_to_node_map = get_asset_nodes_by_asset_key(repositories)
```
sryza, line 18: I recently introduced `ExternalAssetGraph.from_workspace` for building a representation of the asset graph from a `WorkspaceRequestContext`. If we added code versions to `AssetGraph`, I think `CachingProjectedLogicalVersionResolver` might be able to use it instead of `key_to_node_map`.

An additional advantage of that is that, when we incorporate projected versions into asset reconciliation, we'd be able to pass in the `AssetGraph` that the reconciliation sensor has on hand.

python_modules/dagster/dagster/_core/definitions/run_request.py
```diff
@@ -49,6 +49,7 @@ class RunRequest(
             ("tags", PublicAttr[Mapping[str, str]]),
             ("job_name", PublicAttr[Optional[str]]),
             ("asset_selection", PublicAttr[Optional[Sequence[AssetKey]]]),
+            ("stale_only", PublicAttr[bool]),
```
sryza, line 52: I think this should be `stale_assets_only` for extra clarity.

python_modules/dagster/dagster/_scheduler/scheduler.py
```diff
@@ -599,12 +600,13 @@ def _schedule_runs_at_time(
         return
 
     for run_request in schedule_execution_data.run_requests:
+        asset_selection = resolve_asset_selection(workspace_process_context, run_request)  # type: ignore
```
sryza, line 603: we want this in sensors too, right?

python_modules/dagster-graphql/dagster_graphql/implementation/fetch_assets.py
```diff
@@ -396,22 +398,14 @@ def get_freshness_info(
         asset_graph=asset_graph, record=latest_record
     )
 
-    # in the future, if you have upstream source assets with versioning policies, available data
```
sryza, line 399: Are these changes related?

python_modules/dagster/dagster/_core/definitions/logical_version.py
```diff
@@ -193,3 +196,100 @@ def _extract_event_data_from_entry(
 
     assert isinstance(event_data, (AssetMaterialization, AssetObservation))
     return event_data
+
+# ########################
+# ##### PROJECTED LOGICAL VERSION LOADER
+# ########################
+
+class CachingProjectedLogicalVersionResolver:
```
sryza, line 204: We can delete this from the graphql code, right?

python_modules/dagster/dagster/_scheduler/stale.py
sryza: Not advocating for that refactor before merging this if it would risk making this slip from the next release.

sryza (commented):
The overall approach makes sense to me.

@OwenKephart @prha - curious if you have thoughts on this approach of adding a `stale_only` attribute to `RunRequest`, that gets resolved on the host side.

OwenKephart:
> The overall approach makes sense to me.
> 
> @OwenKephart @prha - curious if you have thoughts on this approach of adding a `stale_only` attribute to `RunRequest`, that gets resolved on the host side.

I agree that this makes sense, seems fairly intuitive to me (also agree that it should be `stale_assets_only`)

python_modules/dagster/dagster/_core/definitions/run_request.py
```diff
@@ -72,6 +73,9 @@ class RunRequest(
             Required for sensors that target multiple jobs.
         asset_selection (Optional[Sequence[AssetKey]]): A sequence of AssetKeys that should be
             launched with this run.
+        stale_assets_only (Optional[Sequence[AssetKey]]): Set to true to further narrow the asset
+            selection to stale assets. If passed without an asset selection, all stale assets in the
```
sryza, line 77: Responding to the second sentence, I would expect this parameter to only work in situations where either the sensor, the run request, or the job are already targeting an asset selection.  If none of those are targeting an asset selection, I think we should error.  I.e. I think supplying this parameter should only ever reduce the set of assets that get materialized.

python_modules/dagster/dagster/_core/definitions/run_request.py
```diff
@@ -91,22 +96,19 @@ def __new__(
...
-        self, job_name: Optional[str] = None, asset_selection: Optional[Sequence[AssetKey]] = None
-    ) -> "RunRequest":
-        return RunRequest(
-            run_key=self.run_key,
-            run_config=self.run_config,
-            tags=self.tags,
-            job_name=job_name or self.job_name,
-            asset_selection=asset_selection or self.asset_selection,
-        )
+    def with_replaced_attrs(self, **kwargs: Any) -> "RunRequest":
```
sryza, line 106: Nice.

python_modules/dagster/dagster/_scheduler/stale.py
```diff
@@ -0,0 +1,91 @@
...
+def resolve_asset_selection(
+    context: WorkspaceProcessContext,
+    run_request: RunRequest,
+    job_name: str,
+) -> Optional[Sequence[AssetKey]]:
+    if run_request.stale_assets_only:
+        request_context = context.create_request_context()
+        repositories = _get_repositories(request_context)
+        key_to_node_map = _get_asset_nodes_by_asset_key(repositories)
+        print(run_request)
```
sryza, line 26: prints

sryza (changes requested):
This looks close

sryza (approved):
Nice

As mentioned in off-PR conversation, I think we should make sure that users don't erroneously expect this to work for sensors too (either by making it work for sensors in a followup or making it clear that it doesn't),

PR #11522, opened 2023-01-05
get_event_records support limit + asset event tags + unmigrated state

sryza:


python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -925,6 +924,36 @@ def get_event_records(
...
+            and not self.has_table(AssetEventTagsTable.name)
+        ):
+            if ascending:
+                new_before_cursor = event_records_filter.before_cursor
+                new_after_cursor = row_id
+            else:
+                new_before_cursor = row_id
+                new_after_cursor = event_records_filter.after_cursor
+
+            event_records.extend(
```
prha, line 943: A little wary of exploding the stack by invoking this recursive call.

Also, if I pass in a limit of 100, and i get 99 back, then i'm resetting the call with a limit 1, which means i'll be recursing with a window size of 1.

prha (changes requested):

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
sryza: > Also, if I pass in a limit of 100, and i get 99 back, then i'm resetting the call with a limit 1, which means i'll be recursing with a window size of 1.

Do you mind spelling out the issue a little more? What's wrong with recursing with a window size of 1 in that case?

sryza: Oh I see, like then we'll have a stack frame for every 1 record

cmpadden:
Closing; the tag filtering functionality this PR enhances was completely removed in April 2024 (commit e90350889c).

PR #11527, opened 2023-01-05
[dagit] Track most recent preference for computed log type view

hellendag:
### Summary & Motivation

Use localStorage to track the preference between `stdout` and `stderr` for which tab the user will see first when switching to compute logs on a Run page.

- With no localStorage value set, `stdout` will be the default.
- If the user clicks on either `stdout` or `stderr` tab, track that value.
- If the user then navigates to a run and clicks the button to view compute logs, that tab will be the one that shows first.
- If the user changes tabs, update the tracked value.
- If the user navigates to a run with a `logType` query param (e.g. `logType=stderr`), show that compute log tab but do not set the user's preference locally.

Verify the bullet points above when viewing a run. Load a run, switch to compute logs. Verify that `stdout` is the default. Click `stderr`, reload the page with no `logType` query param, switch to compute logs. Verify that the `stderr` tab shows first.

### DIFF
--- a/js_modules/dagit/packages/core/src/runs/LogsToolbar.tsx
+++ b/js_modules/dagit/packages/core/src/runs/LogsToolbar.tsx
@@ -21,6 +21,7 @@ import styled from 'styled-components/macro';
 
 import {useCopyToClipboard} from '../app/browser';
 import {OptionsContainer, OptionsDivider} from '../gantt/VizComponents';
+import {useStateWithStorage} from '../hooks/useStateWithStorage';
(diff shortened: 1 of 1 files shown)


hellendag:
* master
  * **PR #11527** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11527" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11527?utm_source=stack-comment).

bengotow (approved):
This looks good!

PR #11528, opened 2023-01-05
[docs] - Move ops + jobs tutorial to Guides

erinkcochran87:
This PR moves the ops and jobs tutorial to the Guides category ([preview](https://dagster-git-erin-move-op-tutorial-elementl.vercel.app/guides/dagster/intro-to-ops-jobs)).

It also:

- Adds redirects to account for the re-homing
- Updates the sidenav
- Updates the Guides category page ([preview](https://dagster-git-erin-move-op-tutorial-elementl.vercel.app/guides#general))
- Makes the following changes to the assets tutorial:
  - Consolidates the `/tutorial/setup` page into `/tutorial` ([preview](https://dagster-git-erin-move-op-tutorial-elementl.vercel.app/tutorial))
  - Re-factors `/tutorial` to only focus on the assets tutorial ([preview](https://dagster-git-erin-move-op-tutorial-elementl.vercel.app/tutorial))
  - Moves the tutorial conclusion to its own page (`/tutorial/assets/next-steps`) and adds it to the sidenav ([preview](https://dagster-git-erin-move-op-tutorial-elementl.vercel.app/tutorial/assets/next-steps))

Note: The content is unchanged, with the exception of changing some headings, grammar stuff, etc. The goal of this update was not to refactor content.

---

## Testing

👀


docs/content/tutorial.mdx
```diff
@@ -1,35 +1,50 @@
 ---
-title: Tutorials | Dagster
-description: Getting familiar with Dagster's feature set and tooling through hands-on tutorials.
+title: Intro to Software-defined assets | Dagster Docs
```
tacastillo, line 2: 💭

When I was testing this, I was stuck for a quick second since the Pre-Reqs/Setup section on this page is below the fold on my laptop (13")

My journey was:
1. Click on tutorial
2. See some text that I _should_ read but isn't vital to me getting set up. Scene setting and the outline. All very important in the pedagogy of teaching, but it's like asking a student to read the syllabus. The great students will read it and learn better, but it's an unrealistic expectation for an 18yo to read a syllabus.
3. Skip immediately to the "Defining an Asset" section
4. Get confused that we're jumping into code already and I haven't installed Dagster yet or set up my environment yet because I missed the "Pre-Requisites" section.

As a nudge, maybe at the top of the "Defining an Asset" page, we can link to the "Installation" page (though we shouldn't miss that the Tutorial comes with an additional dependency of `requests`).

What's your take on this?

docs/content/tutorial/assets/next-steps.mdx
```diff
@@ -0,0 +1,15 @@
...
+description: What if you want to do more after completing the asset tutorial?
+---
+
+# Intro to assets: Next steps
+
+🎉 Congratulations! Having reached this far, you now have a working, tested set of software-defined assets.
+
+What if you want to do more?
+
+- **Automating asset materialization** - This tutorial covered how to manually materialize assets. Dagster can also kick off materializations automatically: on fixed [schedules](/concepts/partitions-schedules-sensors/schedules) or when your [sensor](/concepts/partitions-schedules-sensors/sensors) says to.
```
tacastillo, line 12: 🌱

More of just a reminder to myself to mention this in the rewrite, so pls don't action on this: declarative scheduling deserves a mention in the equivalent statement of the new version.

tacastillo (approved):
LGTM!

The [emojis on each comment](https://github.com/erikthedeveloper/code-review-emoji-guide) are to help you categorize them and make it easier to see where my head is at.

There are a couple nits, but none of them are needed on or worth blocking a merge on.

Also I'm trying to not be too critical because I know this is a stop gap.

docs/content/tutorial.mdx
erinkcochran87: 🌱 
This is an interesting point, and seems related to some other feedback I've seen about structuring tutorials. Some folks prefer a long, single page so they can skim and see everything in the tutorial at once, while others prefer the way this tutorial is set up - sections broken up into individual pages.

There are pros and cons to each approach:

- A single page puts everything in one place, allows users to use ctrl+f, and makes jumping around easier. On the other hand, the length can be intimidating.
- Separate pages could make it easier to define starting and stopping points, and also feel like you accomplished something. On the other hand, jumping around isn't as easy and you might miss something (like what happened here).

In the long term, I'd like to add a component to the docs that acts as a "you are here" indicator for tutorials/guides with multiple, lengthy steps. Prereqs/setup would always be included, so it would hopefully be easier to notice if something was skipped.

In the short term, we could:

- Move the Prereqs section to be first in the Defining an asset section
- Move the Prereqs section back to being its own page
- Keep the Prereqs section on this page, but move the syllabus/list down so setup shows above the fold
- Keep the Prereqs section on this page and roll the syllabus/list into it - ex:

   ```markdown
   In this tutorial, you'll:
   
   [list of things]
   
   To complete everything, you'll need:
   
   [setup list]
   ```

docs/content/tutorial/assets/next-steps.mdx
erinkcochran87: I'll add this to the Notion task related to this part of the tutorial so you can keep track of it!

docs/content/tutorial.mdx
tacastillo: I'll let you make the call on whether you want a short term change made, but if you're asking for my vote, I say 1 or 2!

PR #11535, opened 2023-01-06
fix docstring for static_partitioned_config

severo:


sryza (approved):
Looks great. Thanks @severo!

PR #11534, opened 2023-01-06
Add .git to excludes for ruff in pyproject.toml

schrockn:
### Summary & Motivation

Looks like ruff was parsing all the files in .git. Excluding it via pyproject.toml

### How I Tested These Changes

(dagster-x86-3.10.3) elementl-13in-m1-mbppro-2021:dagster schrockn$ make ruff

Before:

```
(dagster-x86-3.10.3) elementl-13in-m1-mbppro-2021:dagster schrockn$ make ruff
ruff --fix .
.git/logs/refs/remotes/origin/yuhan/12-12-convert-examples_4.2/_assets_pandas_type_metadata_repository.py_-___init__.py:1:43: E999 SyntaxError: invalid syntax. Got unexpected token 'a13b52de68dfa49916333202bf26d70ab50cb8db'
Found 1 error(s).
make: *** [ruff] Error 1
(dagster-x86-3.10.3) elementl-13in-m1-mbppro-2021:dagster schrockn$
```

After
```
(dagster-x86-3.10.3) elementl-13in-m1-mbppro-2021:dagster schrockn$ make ruff
ruff --fix .
(dagster-x86-3.10.3) elementl-13in-m1-mbppro-2021:dagster schrockn$
```

PR #11457, opened 2023-01-01
config refactor working branch

schrockn:

PR #7720, opened 2022-05-03
combine validate_tags and check_tags to JSONify nested tags

gibsondan:
Summary:
This will provide consistent Dict => str mapping for anywhere that users input tags.

Test Plan:
BK

PR #11536, opened 2023-01-06
Add persist_run_for_production; Use it in sensor run creation pathways

schrockn:
### Summary & Motivation

This adds a higher level function, persist_run_for_production, to encapsulate run creation pathways. As laid out in https://github.com/dagster-io/dagster/pull/11446 the run creation pathways are in a tough spot. Here we introduce a higher level API to persist new runs suitable for production. I will be adding call sites to this function and introducing new features incrementally as needed (e.g. setting root_run_id, parent_run_id, and explicit run ids)  

Note that I am changing validate_tags to allow for reserved tag names. At this layer in the stack it seems unnecessary and some legitimate tests rely on it (`test_launch_using_memoization`).

BK

### DIFF
--- a/python_modules/dagster/dagster/_core/instance/persist_run.py
+++ b/python_modules/dagster/dagster/_core/instance/persist_run.py
@@ -1,0 +1,68 @@
+from typing import Mapping, Optional
+
+from dagster._core.definitions.selector import PipelineSelector
+from dagster._core.definitions.utils import validate_tags
+from dagster._core.host_representation.code_location import CodeLocation
+from dagster._core.instance import DagsterInstance
+from dagster._core.storage.pipeline_run import DagsterRun, DagsterRunStatus
+from dagster._core.utils import make_new_run_id
+
+
+def persist_run(
+    *,
+    instance: DagsterInstance,
+    code_location: CodeLocation,
+    pipeline_selector: PipelineSelector,
+    run_config: Mapping[str, object],
+    run_tags: Mapping[str, str],
+    explicit_mode: Optional[str],
+) -> DagsterRun:
+    """Creates a run suitable for production using host process (i.e. External*) APIs.
+    This orchestrates necessary interactions with the user process (such as
+    fetching the external pipeline appropriate to the passed subset and
+    compiling the execution plan). This persists a run in instance with the
+    NOT_STARTED state.
+
+    Parameters:
+        instance (DagsterInstance): Instace to execute against
+        code_location (CodeLocation): RepositoryLocation corresponding to user code
+        pipeline_selector (PipelineSelector): Selector that encapsulates subset of pipeline that will be executed
+        run_config (Mapping[str, object]): Run configuration for this run
+        run_tags (Mapping[str, str]): Callsites typically have tags
+            specific to their context (e.g. users can specify ad hoc tags for runs in UI-driven cases).
+            This set of tags will be merged with the tags on the pipeline itself.
+        explicit_mode Optional[str]: Explicitly override the default mode for the pipeline.
+    """
+    external_pipeline = code_location.get_external_pipeline(pipeline_selector)
+    mode = explicit_mode or external_pipeline.get_default_mode_name()
+
+    external_execution_plan = code_location.get_external_execution_plan(
+        external_pipeline=external_pipeline,
+        run_config=run_config,
+        mode=mode,
+        step_keys_to_execute=None,
+        known_state=None,
+        instance=instance,
+    )
+
+    return instance.create_run(
+        pipeline_snapshot=external_pipeline.pipeline_snapshot,
+        parent_pipeline_snapshot=external_pipeline.parent_pipeline_snapshot,
+        execution_plan_snapshot=external_execution_plan.execution_plan_snapshot,
+        pipeline_name=external_pipeline.name,
+        run_id=make_new_run_id(),
+        asset_selection=frozenset(pipeline_selector.asset_selection)
+        if pipeline_selector.asset_selection
+        else None,
+        solid_selection=pipeline_selector.solid_selection,
+        solids_to_execute=external_pipeline.solids_to_execute,
+        run_config=run_config,
+        mode=mode,
+        step_keys_to_execute=external_execution_plan.step_keys_in_plan,
+        tags=validate_tags({**external_pipeline.tags, **run_tags}),
+        root_run_id=None,
+        parent_run_id=None,
+        status=DagsterRunStatus.NOT_STARTED,
+        external_pipeline_origin=external_pipeline.get_external_origin(),
+        pipeline_code_origin=external_pipeline.get_python_origin(),
+    )
--- a/python_modules/dagster/dagster/_daemon/sensor.py
+++ b/python_modules/dagster/dagster/_daemon/sensor.py
@@ -29,12 +29,12 @@ import dagster._seven as seven
 from dagster._core.definitions.run_request import InstigatorType, RunRequest
 from dagster._core.definitions.selector import PipelineSelector
 from dagster._core.definitions.sensor_definition import DefaultSensorStatus, SensorExecutionData
-from dagster._core.definitions.utils import validate_tags
 from dagster._core.errors import DagsterError
 from dagster._core.host_representation.code_location import CodeLocation
-from dagster._core.host_representation.external import ExternalPipeline, ExternalSensor
+from dagster._core.host_representation.external import ExternalSensor
 from dagster._core.host_representation.external_data import ExternalTargetData
 from dagster._core.instance import DagsterInstance
+from dagster._core.instance.persist_run import persist_run
 from dagster._core.scheduler.instigation import (
     InstigatorState,
     InstigatorStatus,
@@ -50,7 +50,6 @@ from dagster._core.workspace.context import IWorkspaceProcessContext
 from dagster._scheduler.stale import resolve_stale_or_missing_assets
 from dagster._utils import DebugCrashFlags, SingleInstigatorDebugCrashFlags
 from dagster._utils.error import SerializableErrorInfo, serializable_error_info_from_exc_info
-from dagster._utils.merger import merge_dicts
 
 if TYPE_CHECKING:
     from pendulum.datetime import DateTime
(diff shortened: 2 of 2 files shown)


schrockn:
* master
  * **PR #11534** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11534" target="_blank">Graphite</a> 
    * **PR #11457** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11457" target="_blank">Graphite</a> 
    * **PR #11536** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11536" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11536?utm_source=stack-comment).

schrockn:
@gibsondan added you around this allow this `validate_tags(external_pipeline.tags or {}, allow_reserved_tags=False)` issue as it blames to https://github.com/dagster-io/dagster/pull/7720

python_modules/dagster/dagster/_core/instance/persist_run_for_production.py
```diff
@@ -0,0 +1,71 @@
...
+
+from dagster._core.definitions.utils import validate_tags
+from dagster._core.host_representation.repository_location import RepositoryLocation
+from dagster._core.host_representation.selector import PipelineSelector
+from dagster._core.instance import DagsterInstance
+from dagster._core.storage.pipeline_run import DagsterRun, DagsterRunStatus
+from dagster._core.utils import make_new_run_id
+
+
+def persist_run_for_production(
```
sryza, line 11: The word "production" rubs me the wrong way a little it, because this will often be invoked within environments that users would not designate as "production".

Thoughts on just calling this "create_run" and renaming the other one to something like "create_run_inner"?

python_modules/dagster/dagster/_core/instance/persist_run_for_production.py
```diff
@@ -0,0 +1,71 @@
...
+
+    Parameters:
+        instance (DagsterInstance): Instace to execute against
+        repository_location (RepositoryLocation): RepositoryLocation corresponding to user code
+        pipeline_selector (PipelineSelector): Selector that encapsulates subset of pipeline that will be executed
+        run_config (Mapping[str, object]): Run configuration for this run
+        context_specific_tags (Mapping[str, str]): Callsites typically have tags
+            specific to their context (e.g. users can specify ad hoc tags for runs.
+            This set of tags will be merged with the tags on the pipeline itself)
+        explicit_mode Optional[str]: Explicitly override the default mode for the pipeline.
```
alangenfeld, line 35: missing () on the type

does it make senes to double encode types between doc string and fn signature? arguably just could drop them all

python_modules/dagster/dagster/_core/instance/persist_run_for_production.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def persist_run_for_production(
+    *,
+    instance: DagsterInstance,
+    repository_location: RepositoryLocation,
+    pipeline_selector: PipelineSelector,
+    run_config: Mapping[str, object],
+    context_specific_tags: Mapping[str, str],
+    explicit_mode: Optional[str],
```
sryza, line 18: Can this default to None?

python_modules/dagster/dagster/_core/instance/persist_run_for_production.py
alangenfeld, line 11: dont love `_for_production`, would personally lean just `persist_run` and use suffix differentiation on the non real shit entry points 

what brought you to this name? any others front runners?

python_modules/dagster/dagster/_core/instance/persist_run_for_production.py
sryza, line 35: Missing parentheses around type.

python_modules/dagster/dagster/_core/instance/persist_run_for_production.py
```diff
@@ -0,0 +1,71 @@
...
+        instance (DagsterInstance): Instace to execute against
+        repository_location (RepositoryLocation): RepositoryLocation corresponding to user code
+        pipeline_selector (PipelineSelector): Selector that encapsulates subset of pipeline that will be executed
+        run_config (Mapping[str, object]): Run configuration for this run
+        context_specific_tags (Mapping[str, str]): Callsites typically have tags
+            specific to their context (e.g. users can specify ad hoc tags for runs.
+            This set of tags will be merged with the tags on the pipeline itself)
+        explicit_mode Optional[str]: Explicitly override the default mode for the pipeline.
+    """
+    external_pipeline = repository_location.get_external_pipeline(pipeline_selector)
```
sryza, line 37: I believe this results in a gRPC call to the user code server.

In the backfill codepaths (maybe elsewhere too), we launch a set of runs that all use the same subset, and we call `get_external_pipeline` just once to avoid N gRPC calls.

Thoughts on how we could make this function usable in that situation without sacrificing perf? E.g. perhaps we could put an LRU cache on `get_external_pipeline`.

python_modules/dagster/dagster/_core/instance/persist_run_for_production.py
```diff
@@ -0,0 +1,71 @@
...
+) -> DagsterRun:
+    """
+    Creates a run suitable for production using host process (i.e. External*) APIs.
+    This orchestrates necessary interactions with the user process (such as
+    fetching the external pipeline appropriate to the passed subset and
+    compiling the execution plan). This persists a run in instance with the
+    NOT_STARTED state.
+
+    Parameters:
+        instance (DagsterInstance): Instace to execute against
```
sryza, line 28:
```suggestion
        instance (DagsterInstance): Instance to execute against
```

python_modules/dagster/dagster/_core/instance/persist_run_for_production.py
```diff
@@ -0,0 +1,71 @@
...
+        repository_location (RepositoryLocation): RepositoryLocation corresponding to user code
+        pipeline_selector (PipelineSelector): Selector that encapsulates subset of pipeline that will be executed
```
sryza, line 30: It's kind of wrinkly that the repository location needs to be specified both in the `pipeline_selector` and the `repository_location`.

python_modules/dagster/dagster/_core/instance/persist_run_for_production.py
```diff
@@ -0,0 +1,71 @@
...
+        run_config (Mapping[str, object]): Run configuration for this run
+        context_specific_tags (Mapping[str, str]): Callsites typically have tags
```
sryza, line 32: I don't love this name - arguably all arguments to all functions are "context-specific" in the sense that they vary based on the callsite. The reason not to just use "tags" or "run_tags" is to make it clear that the definition-level tags don't need to be provided here? Maybe "runtime_tags"?

python_modules/dagster/dagster/_core/instance/persist_run_for_production.py
sryza: We do it on public-facing functions because type annotations don't get picked up nicely by Sphinx. If this is internal, I don't think we need it.

schrockn: very supportive of getting rid of this

python_modules/dagster/dagster/_core/instance/persist_run_for_production.py
schrockn: Sandy I'm going to respond to alex's comment (similar feedback) and resolve this thread.

schrockn: Yeah it's not great. I didn't ruminate on it that much. I was seeing all the other usages of `create_run` in test scenarios so I was like "ok this one is for prod". But I agree with both you and Sandy that the name is pretty terrible.

@sryza  had same feedback:

```
The word "production" rubs me the wrong way a little it, because this will often be invoked within environments that users would not designate as "production".

Thoughts on just calling this "create_run" and renaming the other one to something like "create_run_inner"?
```

I prefer the word "persist" to "create" generally when talking about interaction with storage systems. I think one "creates" objects in memory and then "persists" that representation in storage.

Out of the proposed solutions I prefer calling this `persist_run` and then renaming `DagsterInstance.create_run` to `DagsterInstance.create_potentially_malformed_run` or something of the sort.

python_modules/dagster/dagster/_core/instance/persist_run_for_production.py
schrockn: So I can. However I do think we rely on optional arguments too much in the codebase and I think it was one of the reasons why this got so unruly. We just kept adding "optional" parameters that really weren't optional because we didn't want to update all call sites

python_modules/dagster/dagster/_core/instance/persist_run_for_production.py
schrockn: I'll just eliminate type

python_modules/dagster/dagster/_core/instance/persist_run_for_production.py
schrockn: lru cache seems like a good idea. I can also press ahead and refactor the backfill codepath before committing this to see what feels best.

python_modules/dagster/dagster/_core/instance/persist_run_for_production.py
schrockn: Yeah not totally ideal but I think wrangling PipelineSelector is a separate workstream

python_modules/dagster/dagster/_core/instance/persist_run_for_production.py
schrockn: can do run_tags

alangenfeld (changes requested):
to your queue

python_modules/dagster/dagster/_core/instance/persist_run_for_production.py
schrockn: @sryza  just to follow up here I don't want to further contribute to the pattern of threading both `solid_selection` and `asset_selection` everywhere. I think we should have a single object that represents a subselection of the job and thread that through the entire codebase. I find this sort of repetitive passing of dual parameters that have implicit invariants between them to be pretty confusing.

sryza: A single object that represents a subselection of a job sounds even better. Again not blocking feedback, but I think this could be a good place to introduce that - doesn't necessarily require wrangling the entire codebase.  Callers who want to use the new API and are currently working with a PipelineSelector could just do:

```python
persist_run(
    repository_name=my_pipeline_selector.repository_name,
    pipeline_name=my_pipeline_selector.pipeline_name,
    selection=my_pipeline_selector.selection,
    ...
)
```

Issue #11541, opened 2023-01-06
Regression in 1.1.8: _is_resource_annotated throws TypeError: issubclass() arg 1 must be a class


mikeatlas:
### Dagster version

1.1.8

### What's the issue?

Our job definitions test suite no longer passes with 1.1.8 - previously it worked with 1.1.7.

The code change in question appears to originate from https://github.com/dagster-io/dagster/pull/11322 - [Allow specifying resources in Op/Asset params list #11322](https://github.com/dagster-io/dagster/pull/11322)

Our failing tests stack trace looks like this:

```
___ ERROR collecting tests/my_test.py ___
 [ ... snip ... ]
    def do_something(context, messages: list[dict[str, Any]]):
dagster/_core/definitions/decorators/op_decorator.py:86: in __call__
    arg_resource_keys = {arg.name for arg in compute_fn.get_resource_args()}
dagster/_core/definitions/decorators/solid_decorator.py:65: in get_resource_args
    return get_resource_args(self.decorated_fn)
dagster/_core/definitions/resource_output.py:12: in get_resource_args
    return [param for param in get_function_params(fn) if _is_resource_annotated(param)]
dagster/_core/definitions/resource_output.py:12: in <listcomp>
    return [param for param in get_function_params(fn) if _is_resource_annotated(param)]
dagster/_core/definitions/resource_output.py:17: in _is_resource_annotated
    isinstance(param.annotation, type) and issubclass(param.annotation, ResourceDefinition)
python3.10/abc.py:123: in __subclasscheck__
    return _abc_subclasscheck(cls, subclass)
E   TypeError: issubclass() arg 1 must be a class
```

The op in question that this fails on is relatively benign looking:
```python
@op(
    required_resource_keys={"abc"},
    ins={"messages": In()},
)
def do_something(context, messages: list[dict[str, Any]]):
```

### What did you expect to happen?

Our tests do not even exercise our `op` function with this issue as it appears the test runner fails to load each unit test.

### How to reproduce?

Create a python test that invokes the op as normal function, and try to run with pytest.

```python
from my_job_defs import do_something

def test_do_something():
    assert do_something(null, []) == None
```
Run with pytest (note that the test itself does not even run the assertion line of code - it fails on module import).

### Deployment type

Other

### Deployment details

This can be surfaced without a deployment.

### Additional information

I am on the Dagster Slack and you can ping me with questions or ask here.

### Message from the maintainers

[...]

PR #11551, opened 2023-01-06
[fix] Fix builtin type error related to resource annotations

benpankow:
### Summary & Motivation

Addresses #11541, a bug introduced in #11322.

Users who are using Python >= 3.9 may use [builtin types](https://peps.python.org/pep-0585/) for generic type annotations, e.g. `list[dict[str, str]]` rather than `typing.List[typing.Dict[str, str]]`.

Unfortunately, these builtin generics manage to get around the safety checks in our subclass code - `isinstance(list[dict[str, str]], type) == True` and `inspect.isclass(list[dict[str, str]]) == True` but `issubclass(list[dict[str, str]], Resource)` raises an error that `list[dict[str, str]]` is not a class:

```python
TypeError: issubclass() arg 1 must be a class
```

This seems to be a specific issue related to `abc._abc_subclasscheck`, since `Resource` is an ABC. The CPython implementation of the subclass check seems to have different results than the python one, which uses `isinstance(..., type)` under the hood: https://bugs.python.org/issue44847

This PR adds a try/catch guard to areas where we make a subclass call to avoid this issue.

## Test Plan

Add a test suite which is run for Python >= 3.9 to ensure builtin generics can be used.

### DIFF
--- a/python_modules/dagster/dagster_tests/core_tests/resource_tests/test_resource_parameters.py
+++ b/python_modules/dagster/dagster_tests/core_tests/resource_tests/test_resource_parameters.py
@@ -306,6 +306,8 @@ def test_asset_with_structured_config():
     assert executed["the_asset"]
 
 
+# Disabled for Python versions <3.9 as builtin types do not support generics
+# until Python 3.9, https://peps.python.org/pep-0585/
 @pytest.mark.skipif(sys.version_info < (3, 9), reason="requires python3.9")
 def test_no_err_builtin_annotations():
     # Ensures that we can use Python builtin types without causing any issues, see
--- a/python_modules/dagster/dagster_tests/core_tests/resource_tests/test_resource_parameters.py
+++ b/python_modules/dagster/dagster_tests/core_tests/resource_tests/test_resource_parameters.py
@@ -1,3 +1,4 @@
+import sys
 from typing import Any
 
 import pytest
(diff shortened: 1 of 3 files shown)


benpankow:
* master
  * **PR #11551** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11551" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11551?utm_source=stack-comment).

mikeatlas:
Cheers

python_modules/dagster/dagster_tests/core_tests/resource_tests/test_resource_parameters.py
```diff
@@ -303,3 +304,40 @@ def the_asset(context, config: AnAssetConfig, foo: ResourceOutput[str]):
...
+
+@pytest.mark.skipif(sys.version_info < (3, 9), reason="requires python3.9")
+def test_no_err_builtin_annotations():
+    # Ensures that we can use Python builtin types without causing any issues, see
+    # https://github.com/dagster-io/dagster/issues/11541
+
+    # Use aliases for the builtin types so that this doesn't cause a load error on
+    # python < 3.9
+    _list = list
+    _dict = dict
```
schrockn, line 317: Why is this alias necessary for the this test to pass? I'm not understanding.

python_modules/dagster/dagster_tests/core_tests/resource_tests/test_resource_parameters.py
```diff
@@ -303,3 +304,40 @@ def the_asset(context, config: AnAssetConfig, foo: ResourceOutput[str]):
         .success
     )
     assert executed["the_asset"]
+
+
+@pytest.mark.skipif(sys.version_info < (3, 9), reason="requires python3.9")
+def test_no_err_builtin_annotations():
```
schrockn, line 310: Shouldn't we run this test on all versions

schrockn (changes requested):
Comments inline

python_modules/dagster/dagster_tests/core_tests/resource_tests/test_resource_parameters.py
benpankow: Ah yeah, I think it's actually not necessary, since the guard will prevent the function body & type annotations from being evaluated

python_modules/dagster/dagster_tests/core_tests/resource_tests/test_resource_parameters.py
benpankow: Builtin types cannot be used as generics in python < 3.9, e.g.

```python
Python 3.8.12 (default, Aug  5 2022, 12:15:35)
>>> list[dict[str, str]]
TypeError: 'type' object is not subscriptable
```

So this test only makes sense in >=3.9

python_modules/dagster/dagster_tests/core_tests/resource_tests/test_resource_parameters.py
schrockn: Great thanks

python_modules/dagster/dagster_tests/core_tests/resource_tests/test_resource_parameters.py
schrockn: Can you add "Builtin types cannot be used as generics in python < 3.9, e.g." as a comment then?

schrockn (approved):
Approved. Please add comment re: built-in types as generics.

PR #11552, opened 2023-01-06
fix remote captured log manager complete check

prha:
### Summary & Motivation
We should be checking to see if the capture is complete by checking remote storage, not just local storage so that the read-side compute log manager can report download links accurately.

This should ensure that download links can be constructed properly

BK

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/cloud_storage_compute_log_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/cloud_storage_compute_log_manager.py
@@ -112,7 +112,10 @@ class CloudStorageComputeLogManager(CapturedLogManager, ComputeLogManager):
         self.upload_to_cloud_storage(log_key, ComputeIOType.STDERR)
 
     def is_capture_complete(self, log_key: Sequence[str]) -> bool:
-        return self.local_manager.is_capture_complete(log_key)
+        if self.local_manager.is_capture_complete(log_key):
+            return True
+        # check remote storage
+        return self.cloud_storage_has_logs(log_key, ComputeIOType.STDERR)
 
     def _log_data_for_type(self, log_key, io_type, offset, max_bytes):
         if self._has_local_file(log_key, io_type):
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/storage/cloud_storage_compute_log_manager.py
```diff
@@ -112,7 +112,10 @@ def _on_capture_complete(self, log_key: Sequence[str]):
         self.upload_to_cloud_storage(log_key, ComputeIOType.STDERR)
 
     def is_capture_complete(self, log_key: Sequence[str]) -> bool:
-        return self.local_manager.is_capture_complete(log_key)
+        if self.local_manager.is_capture_complete(log_key):
+            return True
+        # check remote storage
+        return self.cloud_storage_has_logs(log_key, ComputeIOType.STDERR)
```
johannkm, line 118: why ERR? we can assume it also has out?

python_modules/dagster/dagster/_core/storage/cloud_storage_compute_log_manager.py
```diff
@@ -112,7 +112,10 @@ def _on_capture_complete(self, log_key: Sequence[str]):
         self.upload_to_cloud_storage(log_key, ComputeIOType.STDERR)
 
     def is_capture_complete(self, log_key: Sequence[str]) -> bool:
-        return self.local_manager.is_capture_complete(log_key)
+        if self.local_manager.is_capture_complete(log_key):
```
johannkm, line 115: If this is true, we have the files locally? Does that mean that they'd also be available in cloud storage (since that's where they'd have been pulled from?)

johannkm (approved):

python_modules/dagster/dagster/_core/storage/cloud_storage_compute_log_manager.py
prha: depends on whether this is the read side or write side.  On the write-side, we check for capture complete, then upload.  On the read-side, we check for the local files, then the remote files, then download from remote to local.

python_modules/dagster/dagster/_core/storage/cloud_storage_compute_log_manager.py
prha: well, for op-based log capture, err will always be populated, not necessarily out.  But you're right we should check both.

PR #11553, opened 2023-01-06
[2/n] Expose top-level resources via GraphQL

benpankow:
## Summary

Adds GraphQL endpoints necessary to query top-level resources from a repository. Includes an endpoint to fetch all top-level resources and an endpoint to fetch a particular resource by name.

## Test Plan

Unit tests of GraphQL endpoints.

PR #11529, opened 2023-01-05
[1/n] Serialize top-level resources into repository data

benpankow:
## Summary

Incorporates top-level resources into the Repository object, serializing them and sending them as part of external repository data so that they can be displayed in the UI (see stacked PRs).

Also incorporates the change from #11470 to copy values into default values when configuring structured resources.

## Test Plan

Unit tests.

PR #11554, opened 2023-01-06
[3/n] Display top-level resources in Dagit sidebar

benpankow:
## Summary

Adds a basic frontend layer to display top-level resources in the Dagit sidebar, and a resource page to display their details:

Screen Shot 2023-01-09 at 11 10 16 AM

This is currently gated behind a user-visible feature flag - by default does not affect what a user is presented with.

## Test Plan

tested locally

### DIFF
--- a/js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
+++ b/js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
@@ -62,8 +62,6 @@ export const ResourceRoot: React.FC<Props> = (props) => {
     notifyOnNetworkStatusChange: true,
   });
 
-  console.log(queryResult);
-
   return (
     <Page>
       <PageHeader
--- a/js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
+++ b/js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
@@ -30,11 +30,11 @@ interface Props {
 }
 
 const remapName = (inName: string): string => {
-  if (inName == 'StringSourceType') {
+  if (inName === 'StringSourceType') {
     return 'String';
-  } else if (inName == 'IntSourceType') {
+  } else if (inName === 'IntSourceType') {
     return 'Int';
-  } else if (inName == 'BoolSourceType') {
+  } else if (inName === 'BoolSourceType') {
     return 'Bool';
   }
   return inName;
@@ -103,7 +103,7 @@ export const ResourceRoot: React.FC<Props> = (props) => {
                       const defaultValue = field.defaultValueAsJson;
                       const actualValue = configuredValues[field.name];
 
-                      const isDefault = defaultValue == actualValue;
+                      const isDefault = defaultValue === actualValue;
 
                       return (
                         <tr key={field.name}>
--- a/js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
+++ b/js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
@@ -1,0 +1,207 @@
+import {gql, useQuery} from '@apollo/client';
+import {
+  Box,
+  Colors,
+  Heading,
+  Page,
+  PageHeader,
+  SplitPanelContainer,
+  Table,
+  Tag,
+  Tooltip,
+} from '@dagster-io/ui';
+import * as React from 'react';
+import {useParams} from 'react-router-dom';
+import styled from 'styled-components/macro';
+
+import {useTrackPageView} from '../app/analytics';
+import {ResourceRootQueryQuery} from '../graphql/graphql';
+import {useDocumentTitle} from '../hooks/useDocumentTitle';
+import {RepositoryLink} from '../nav/RepositoryLink';
+import {SidebarSection} from '../pipelines/SidebarComponents';
+import {Loading} from '../ui/Loading';
+import {repoAddressToSelector} from '../workspace/repoAddressToSelector';
+import {RepoAddress} from '../workspace/types';
+
+interface Props {
+  repoAddress: RepoAddress;
+}
+
+const remapName = (inName: string): string => {
+  if (inName == 'StringSourceType') {
+    return 'String';
+  } else if (inName == 'IntSourceType') {
+    return 'Int';
+  } else if (inName == 'BoolSourceType') {
+    return 'Bool';
+  }
+  return inName;
+};
+
+export const ResourceRoot: React.FC<Props> = (props) => {
+  useTrackPageView();
+
+  const {repoAddress} = props;
+
+  const {resourceName} = useParams<{resourceName: string}>();
+
+  useDocumentTitle(`Resource: ${resourceName}`);
+
+  const resourceSelector = {
+    ...repoAddressToSelector(repoAddress),
+    resourceName,
+  };
+  const queryResult = useQuery<ResourceRootQueryQuery>(RESOURCE_ROOT_QUERY, {
+    variables: {
+      resourceSelector,
+    },
+    fetchPolicy: 'cache-and-network',
+    partialRefetch: true,
+    notifyOnNetworkStatusChange: true,
+  });
+
+  console.log(queryResult);
+
+  return (
+    <Page>
+      <PageHeader
+        title={
+          <Box flex={{direction: 'row', alignItems: 'center', gap: 12}}>
+            <Heading>{resourceName}</Heading>
+          </Box>
+        }
+      />
+      <Loading queryResult={queryResult} allowStaleData={true}>
+        {({topLevelResourceOrError}) => {
+          if (topLevelResourceOrError.__typename !== 'TopLevelResource') {
+            return null;
+          }
+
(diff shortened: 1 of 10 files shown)


benpankow:
* master
  * **PR #11529** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11529" target="_blank">Graphite</a> 
    * **PR #11553** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11553" target="_blank">Graphite</a> 
      * **PR #11554** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11554" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11554?utm_source=stack-comment).

js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const queryResult = useQuery<ResourceRootQuery>(RESOURCE_ROOT_QUERY, {
+    variables: {
+      resourceSelector,
+    },
+    fetchPolicy: 'cache-and-network',
+    partialRefetch: true,
+    notifyOnNetworkStatusChange: true,
+  });
+
+  console.log(queryResult);
```
schrockn, line 65: 🤔

schrockn (approved):
So this approval is about my support for getting this in master in general. @hellendag should still do a tactical review of the JS

js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
benpankow: fixed

js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
```diff
@@ -0,0 +1,208 @@
...
+    notifyOnNetworkStatusChange: true,
+  });
+
+  return (
+    <Page>
+      <PageHeader
+        title={
+          <Box flex={{direction: 'row', alignItems: 'center', gap: 12}}>
+            <Heading>{resourceName}</Heading>
+          </Box>
```
hellendag, line 71: Doesn't seem like it needs to be wrapped in a `Box`.

js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
```diff
@@ -0,0 +1,208 @@
...
+          );
+
+          return (
+            <SplitPanelContainer
+              identifier="explorer"
+              firstInitialPercent={70}
+              firstMinSize={400}
+              height="calc(100% - 60px)"
+              first={
+                <Table>
```
hellendag, line 91: Let's add `$monospaceFont={false}` to this.

js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
```diff
@@ -0,0 +1,208 @@
...
+              }
+              second={
+                <RightInfoPanel>
+                  <RightInfoPanelContent>
+                    <Box
+                      flex={{gap: 4, direction: 'column'}}
+                      margin={{left: 24, right: 12, vertical: 16}}
+                    >
+                      <Heading>{resourceName}</Heading>
+                    </Box>
```
hellendag, line 139: Seems like it might be redundant and unnecessary since you've already got the page header.

js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
```diff
@@ -0,0 +1,208 @@
...
+        }}
+      </Loading>
+    </Page>
+  );
+};
+
+export const RightInfoPanel = styled.div`
+  // Fixes major perofmance hit. To reproduce, add enough content to
+  // the sidebar that it scrolls (via overflow-y below) and then try
+  // to pan the DAG.
```
hellendag, line 170: Relevant?

js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
```diff
@@ -0,0 +1,208 @@
...
+          const configuredValues = Object.fromEntries(
+            topLevelResourceOrError.configuredValues.map((cv) => [cv.key, cv.value]),
+          );
+
+          return (
+            <SplitPanelContainer
+              identifier="explorer"
+              firstInitialPercent={70}
+              firstMinSize={400}
+              height="calc(100% - 60px)"
```
hellendag, line 89: Not sure this new prop is necessary -- are there some other `SplitPanelContainer` uses that look similar? I think maybe a wrapper div to give the element height might do the job here. `OpsRoot` perhaps?

js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
```diff
@@ -0,0 +1,208 @@
...
+                              </div>
+                            </Box>
+                          </td>
+                          <td>{remapName(field.configTypeKey)}</td>
+                          <td>
+                            <Box flex={{direction: 'row', justifyContent: 'space-between'}}>
+                              <Tooltip content={<>Default: {defaultValue}</>} canShow={!isDefault}>
+                                {actualValue}
+                              </Tooltip>
+                              {isDefault && <Tag>Default</Tag>}
```
hellendag, line 122: Do we need both the tooltip and the tag?

js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+  const resourceSelector = {
+    ...repoAddressToSelector(repoAddress),
+    resourceName,
+  };
+  const queryResult = useQuery<ResourceRootQuery>(RESOURCE_ROOT_QUERY, {
+    variables: {
+      resourceSelector,
+    },
+    fetchPolicy: 'cache-and-network',
```
hellendag, line 60: Not necessary, this is already the default.

js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
```diff
@@ -0,0 +1,208 @@
...
+    partialRefetch: true,
+    notifyOnNetworkStatusChange: true,
```
hellendag, line 62: Are these needed?

js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
```diff
@@ -0,0 +1,208 @@
...
+        title={
+          <Box flex={{direction: 'row', alignItems: 'center', gap: 12}}>
+            <Heading>{resourceName}</Heading>
+          </Box>
+        }
+      />
+      <Loading queryResult={queryResult} allowStaleData={true}>
+        {({topLevelResourceOrError}) => {
+          if (topLevelResourceOrError.__typename !== 'TopLevelResource') {
+            return null;
```
hellendag, line 77: What should this case look like?

js_modules/dagit/packages/ui/src/components/SplitPanelContainer.tsx
```diff
@@ -15,6 +15,7 @@ interface SplitPanelContainerProps {
   firstInitialPercent: number;
   firstMinSize?: number;
   second: React.ReactNode | null; // Note: pass null to hide / animate away the second panel
+  height?: string;
```
hellendag, line 18: See comment on the `<SplitPanelContainer>` callsite, I'm not sure you need this.

hellendag (approved):

js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
benpankow: I do think the repetition is kind of nice personally, similar to the job page (this is a rough take based on Josh's [loom](https://www.loom.com/share/e4a3ad1f3dc04cd7b2331c645728c78b))

Screen Shot 2023-02-02 at 3 03 50 PM

js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
benpankow: Hmm, it looks like the split panel container doesn't want to fill the parent in that case. You can see that the parent div has 100% height but the container itself doesn't expand:

Screen Shot 2023-02-02 at 3 07 08 PM

js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
benpankow: The tooltip is useful to view the default value in the non-default case, the default tag is useful to tell without hovering that the value is a default.

We could potentially put the default in a tag (in the non-default case) but the size of the default value could be quite large.

js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
benpankow: Added error message display in the error case

PR #11555, opened 2023-01-06
use run tags instead of asset tags to filter materializations within backfill

sryza:
### Summary & Motivation

This is a reversal of the approach taken in https://github.com/dagster-io/dagster/pull/11506.

I realized a problem with that earlier approach: it depends on user code being upgraded.  I.e. if someone running a Dagster 1.1.8 daemon runs a backfill against user code that depends on Dagster 1.1.7 or earlier, then the asset materialization events that are created as part of that backfill won't have the expected backill tag on them.

As a consequence, if that backfill has some asset partitions that can’t be run until other materializations complete, then the backfill will fail to terminate.  This should only be a problem right now in very rare situations, but will be more of a problem when we remove restrictions that prevent users from launching backfills across different code locations and across different partitions definitions.

The expected behavior is covered by existing test coverage. In an ideal world, we would write a test that makes sure backfills kicked off on later versions of host code complete successfully on earlier versions of user code, but I can't think of a way to do that that avoids a bunch of very heavy lifting.

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/asset_backfill.py
+++ b/python_modules/dagster/dagster/_core/execution/asset_backfill.py
@@ -24,11 +24,11 @@ from dagster._core.definitions.events import AssetKey, AssetKeyPartitionKey
 from dagster._core.definitions.external_asset_graph import ExternalAssetGraph
 from dagster._core.definitions.mode import DEFAULT_MODE_NAME
 from dagster._core.definitions.run_request import RunRequest
+from dagster._core.events import DagsterEventType
 from dagster._core.host_representation.selector import PipelineSelector
 from dagster._core.instance import DagsterInstance
 from dagster._core.storage.pipeline_run import DagsterRunStatus, RunsFilter
 from dagster._core.storage.tags import BACKFILL_ID_TAG, PARTITION_NAME_TAG
-from dagster._core.utils import frozendict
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -302,55 +302,69 @@ def execute_asset_backfill_iteration_inner(
...
-    )
+        recently_materialized_asset_partitions = AssetGraphSubset(asset_graph)
+        for asset_key in asset_backfill_data.target_subset.asset_keys:
+            records = instance_queryer.get_materialization_records(
+                asset_key=asset_key, after_cursor=asset_backfill_data.latest_storage_id
+            )
+            records_in_backfill = [
+                record
+                for record in records
+                if instance_queryer.run_has_tag(
```
prha, line 334: Is it possible to pull this query out of the loop and batch?  Or limit the number of repeated calls to fetch run tags in case the run_id is repeated (e.g. materialize multiple assets within the same run)?

sryza: the instance_queryer handles caching (in `_get_run_record_by_id`)

prha: ah, missed the cached_method decorator.

prha (approved):

PR #11564, opened 2023-01-06
[dagit] Stop generating docblocks on schema.graphql and codegen

hellendag:
### Summary & Motivation

There is a discrepancy between `yarn generate-graphql` and `yarn graphql-codegen` (e.g. for `--watch`):

- For `generate-graphql`, we create `schema.graphql` in `generateGraphQLTypes.ts` twice: once without description docblocks for the generated type file, and once with descriptions to restore them for humans. The produced `graphql.ts` does not have docblocks.
- For `graphql-codegen`, which doesn't go through this script, only the version of `schema.graphql` *with* descriptions is created. The produced `graphql.ts` **does** have docblocks.

This is likely to cause confusion for users that use both commands (e.g. during development vs. when regenerating outside of immediate query development, regenerating permissions json, etc). In my opinion, we don't really need the docblocks on `schema.graphql` at all -- I've never even looked at them.

To simplify everything -- and speed up `generate-graphql` because we don't have to build the `schema.graphql` twice! -- let's just not ever include description docblocks.

`yarn generate-graphql` and `yarn graphql-codegen`, verify that they produce the same `graphql.ts` output.

### DIFF
--- a/js_modules/dagit/packages/core/src/graphql/schema.graphql
+++ b/js_modules/dagit/packages/core/src/graphql/schema.graphql
@@ -6,10 +6,6 @@ schema {
 
 type ComputeLogFile {
   path: String!
-
-  """
-  The data output captured from step computation at query time
-  """
   data: String
   cursor: Int!
   size: Int!
@@ -54,9 +50,6 @@ interface ErrorEvent {
   error: PythonError
 }
 
-"""
-The types of events that may be yielded by solid and pipeline execution.
-"""
 enum DagsterEventType {
   STEP_OUTPUT
   STEP_INPUT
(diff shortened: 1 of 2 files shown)


hellendag:
* master
  * **PR #11564** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11564" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11564?utm_source=stack-comment).

bengotow (approved):
This looks good to me but I'm curious if your other PR (https://github.com/dagster-io/dagster/pull/11565) actually gives us a use for the docblocks -- there's a chance that the plugin will pull these strings into the autocompletion tooltips the same way it does jsdoc annotations...

Either way, I usually pull up Graphiql to view these -- didn't even know they were in here -- so this seems ok!

Issue #11185, opened 2022-12-16
Dagit: show `compute_kind` badge in asset catalog and asset definition


peay:
### What's the use case?

`compute_kind` can be set on an asset to something like `dbt`, `spark` or others. This is shown as a badge in asset graphs, with a color, and is a useful visual cue to differentiate assets.

I would find it useful to also display the badge in the asset catalog, when browsing assets (under `/instance/assets/...`), and in the asset definition page (under `/instance/assets/asset-name?view=definition`).

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

peay:
Done in https://github.com/dagster-io/dagster/pull/11569, thanks!

PR #11569, opened 2023-01-07
[dagit] Add the Asset compute kind tag in more places #11185

bengotow:
### Summary & Motivation

Related: https://github.com/dagster-io/dagster/issues/11185

This PR rewrites the OpTags component a bit so that we render the text separately from the logo image and can invert the logo images to create a second "toned-down" version of the tag for display in other parts of Dagit.

I added the tags to the asset list view and also to the asset detail pages.

Viewed relevant pages of Dagit. I also confirmed that this change does not have any adverse impact on regular old op tags in op graps. (Examples attached)
 
image

image

image

Screen Shot 2023-01-16 at 11 29 22 AM
Screen Shot 2023-01-16 at 11 29 53 AM


peay:
Thanks @bengotow! For the asset details, would it make sense to have the tag on the top line next to `<Asset in repo> <GROUP2> <Stale>`, where we already have a bunch of tags? This would also keep it visible in tabs other than 'Definition'.

hellendag (approved):

PR #11573, opened 2023-01-07
[dagit] Support range-based partition health data

bengotow:
### Summary & Motivation

This PR makes Dagit compile again with the updates in #10822

This is phase 1 of two phases of work to use range-based partition health data. This PR updates our existing client-side health structures to read the new GraphQL endpoints which return the state of partition ranges rather than the state of each individual partition in a "matrix" of booleans.

Mapping the new APIs to our client-side requirements does require some transforms, and it increases code complexity in `usePartitionHealthData`. Namely, we need to add the start/end partition indexes of each range so we can efficiently look up the status of a particular partition within the range data. (eg: `given these ranges of [A->C], [E->F], is D materialized?`)

The second phase of work is to 1) make the "partition health bar" rely more directly on this data and not load the partition key space along with it, and 2) make Dagit query for the health of individual partitions individually via a new endpoint (so it doesn't have to answer the question above). This is our eventual goal, but will require a bit more work:

- In the static partitions case, `materializedPartitions` alone cannot tell us the correct display order for `materializedPartitions` / `unmaterializedParttions`. We still need to load all the keys in order in order to display the health bar.

- In the multi-dimensional health case, the returned structure contains subranges. We don't actually care at all about the subranges, but we need to know whether the second dimension is `"complete", "partial" or "missing"`. I think a backend resolver might need to be added for this that just returns these three states, or missingCount / completeCount.

We're planning to postpone the second chunk of work here, because the current implementation (which loads the partition keys) isn't actually that slow now that we've made other changes, and it's unclear if all the work will be worthwhile.

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/usePartitionHealthData.tsx
+++ b/js_modules/dagit/packages/core/src/assets/usePartitionHealthData.tsx
@@ -2,11 +2,13 @@ import {gql, useApolloClient} from '@apollo/client';
 import isEqual from 'lodash/isEqual';
 import React from 'react';
 
+import {assertUnreachable} from '../app/Util';
 import {PartitionState} from '../partitions/PartitionStatus';
 
 import {mergedStates} from './MultipartitioningSupport';
 import {AssetKey} from './types';
 import {
+  PartitionHealthMaterializedPartitionsFragment,
   PartitionHealthQuery,
   PartitionHealthQueryVariables,
 } from './types/usePartitionHealthData.types';
@@ -24,7 +26,6 @@ export interface PartitionHealthData {
   assetKey: AssetKey;
   dimensions: PartitionHealthDimension[];
   stateForKey: (dimensionKeys: string[]) => PartitionState;
-  stateForPartialKey: (dimensionKeys: string[]) => PartitionState;
   stateForSingleDimension: (
     dimensionIdx: number,
     dimensionKey: string,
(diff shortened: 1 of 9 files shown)


python_modules/dagster-graphql/dagster_graphql/schema/partition_sets.py
```diff
@@ -358,7 +358,7 @@ def _get_time_partitions_metadata(partition_def_data):
 
             partitions_def = partition_def_data.get_partitions_definition()
             return GrapheneTimePartitionsDefinitionMetadata(
-                startTime=partitions_def.start_time,
+                startTime=partitions_def.start,
```
bengotow, line 361: hey @clairelin135  I had to change this to get Dagit to load the asset graph for me - does this make sense?

clairelin135: Hi @bengotow ah that's my mistake! I think we actually want `partitions_def.start.timestamp()` as that will generate the float value instead of returning the datetime object.

But I've updated my PR to fix this so you should be able to rebase and get this new fix

clairelin135 (commented):
Thanks for putting this together! I noticed your comment below, I added a fix in my PR that should address this! Though you'll have to rebase.

I played around with this a bit and I think there's a small bug in how multidimensional partition statuses are rendered in the empty case. When they're unmaterialized, one status will display as fully present:
image

Otherwise, this looks good to me. I'll leave it to others to review the JS.

bengotow:
Ahh good catch on the empty state! Will take a look and see what's going on 👀

js_modules/dagit/packages/core/src/assets/usePartitionHealthData.tsx
```diff
@@ -49,78 +49,101 @@ export type PartitionDimensionSelection = {
...
-        ]),
-  );
+  // The backend re-orders the dimensions only for the materializedPartitions ranges so that
+  // the time partition is the "primary" one, even if it's dimension[1] elsewhere.
+  // This matches the way we display them in the UI and makes some common data retrieval faster,
+  // but Dagit's internals always use the REAL ordering of the partition keys, we need to flip
+  // everything in this function to match the range data.
+  const isRangeDataInverted =
+    materializedPartitions.__typename === 'MultiPartitions' &&
+    materializedPartitions.primaryDimensionName !== __dims[0].name;
```
hellendag, line 71: `__dims[0].name` looks a little risky for a type error, since it looks like there is a case where `__dims` can be an empty array.

hellendag (commented):
Left a comment inline, but I have some similar worries about the numeric index referencing throughout, which look a little worrying for type errors that TS won't catch by default. Would it be possible to add some guards to improve the soundness of those references?

hellendag (approved):
I'll go ahead and approve since the request above isn't blocking.

js_modules/dagit/packages/core/src/assets/usePartitionHealthData.tsx
bengotow: Hmm I think you're dwefinitely right about these index accesses - I'll add a few more defensive checks in here.

PR #11574, opened 2023-01-09
[dagit] Rename a few partition data structures in Dagit

bengotow:
### Summary & Motivation

This PR doesn't make any business logic changes - just renaming some things!

- The partition pages refer to your selection (eg `[2022-01-01...2023-01-01]`) as a partition "range". We're switching the partition health data structure to represent health as a series of ranges, so this is going to be super confusing. Changing it to `selections`

- Right now, all of Dagit operates on a selection of partition keys. When you select `[A...Z]`, we set the selection to `[A, B, C, D, ...]`.  This isn't ideal for gigantic partition key sets and our goal is to stop loading the entire partition key space into Dagit entirely, which will make this transform impossible.  This is going to be a fairly big change, and requires new GraphQL transforms and for the "launchBackfillMutation" to accept a range and not an array of keys, etc. But the first step is to start passing around both selectedKeys and selectedRanges, where the original user input `[A...Z]` is preserved in state and deprecating usage of the array of selected keys.

Used the relevant pages following these changes!


sryza:
I'm going to defer to @hellendag on this one

hellendag (approved):

PR #11576, opened 2023-01-09
Show clearer grpc errors, as well as a custom failure for sensor timeouts in particular since those are so common

gibsondan:
Summary:
- surface the error code instead of burying it in the nested cause error
- for timeouts, include the original timeout (and that it timed out)
- For sensors, include how you might fix it

### DIFF
--- a/python_modules/dagster/dagster/_grpc/client.py
+++ b/python_modules/dagster/dagster/_grpc/client.py
@@ -128,17 +128,34 @@ class DagsterGrpcClient:
             stub = DagsterApiStub(channel)
             return getattr(stub, method)(request, metadata=self._metadata, timeout=timeout)
 
+    def _raise_grpc_exception(self, e: Exception, timeout, custom_timeout_message=None):
+        if isinstance(e, grpc.RpcError):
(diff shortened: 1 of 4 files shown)


python_modules/dagster/dagster/_grpc/client.py
```diff
@@ -128,17 +128,34 @@ def _get_response(
             stub = DagsterApiStub(channel)
             return getattr(stub, method)(request, metadata=self._metadata, timeout=timeout)
 
+    def _raise_grpc_exception(self, e: Exception, timeout, custom_timeout_message=None):
+        if isinstance(e, grpc.RpcError):
+            if e.code() == grpc.StatusCode.DEADLINE_EXCEEDED:
+                raise DagsterUserCodeUnreachableError(
+                    custom_timeout_message
+                    or f"User code query timed out due to taking longer than {timeout} seconds to complete."
```
rexledesma, line 136: nit: user code -> code location server

```suggestion
                    or f"The query to the code location server timed out due to taking longer than {timeout} seconds to complete."
```

rexledesma (approved):

PR #11582, opened 2023-01-09
[asset-reconciliation] Limit the partitions that can be reconciled with the sensor

OwenKephart:
### Summary & Motivation

In general, it's undesirable to automatically kick off runs of old partitions, as this can inadvertently cause "backfills" to be initiated.

This change makes it so that, for time-window-partitioned assets, no runs will be kicked off for partitions whose windows end more than one day before the latest window for that asset. In essence, this gives a time bound of ~1 day for the sensor to materialize a partition of an asset once that partition comes into being. 

This required changing some tests.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -336,7 +336,7 @@ def determine_asset_partitions_to_reconcile(
                     candidate.partition_key
                 ).end
                 time_delta = latest_partition_end - candidate_partition_end
-                if time_delta > datetime.timedelta(days=1):
+                if time_delta >= datetime.timedelta(days=1):
                     return False
 
         if any(
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -17,6 +17,7 @@ from typing import (
     Tuple,
     cast,
 )
+from dagster._core.definitions.time_window_partitions import TimeWindowPartitionsDefinition
 
 import pendulum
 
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -321,6 +322,23 @@ def should_reconcile(
         candidates_unit: Iterable[AssetKeyPartitionKey],
         to_reconcile: AbstractSet[AssetKeyPartitionKey],
     ) -> bool:
+        # do not attempt to materialize partitions more than one day older than the latest partition
```
sryza, line 325: I think it's worth encapsulating this in a helper to like a lot of the other sub-logic in `should_reconcile` is

sryza (commented):
I am in favor of this - it's been a sticking point for a bunch of users adopting the reconciliation sensor. A question regarding perf - for root assets, I think we're going to end up iterating over (and discarding) all historical partitions on every tick - is there a straightforward way to avoid that?

OwenKephart:
@sryza re: avoiding iterating over those old empty partitions, I think this would only apply to the root asset partitions (as non-materializable partitions downstream of the root partitions will be filtered out during the bfs). I think we could fix this by modifying get_never_requested_never_materialized_partitions, but I don't think there currently exists a more efficient way of removing those old partitions from the subset than iterating through each of them, so not sure if there's a way to noticeably improve overall performance.

sryza:
> I think this would only apply to the root asset partitions (as non-materializable partitions downstream of the root partitions will be filtered out during the bfs). I think we could fix this by modifying get_never_requested_never_materialized_partitions

Yeah, that's what I was thinking

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -264,6 +265,30 @@ def find_never_materialized_or_requested_root_asset_partitions(
     )
 
 
+def can_materialize_candidates_unit(
```
sryza, line 268: Could this have a more specific name that refers to the fact that it's filtering based on time?

sryza (approved):
I'm a little conflicted on whether the root partitions issue is worth blocking this on. Seems like a user could easily get into a situation where they have a few 10k-partition root assets that they're never gonna care about backfilling the early partitions of. But the reconciliation sensor has to iterate over all those root partitions every time it runs, forever, which could be expensive.

PR #11584, opened 2023-01-09
RFC: Single "dagster start" command that spins up dagit and the daemon as subprocesses and requires no DAGSTER_HOME env var to be set

gibsondan:
Summary:
It is not uncommon for users to hit a roadblock when trying to get the daemon running for the first time - they need to set the DAGSTER_HOME env var, make sure its the same between two different services, keep them running in two separate windows, etc.

This proposes a single "dagster start" command that:
- spins up both services for you as subprocesses, and spins them down when the process is interrupted
- uses a subset of the workspace args available to both dagit and dagster-daemon currently
- requires no env vars: uses your current working directory as a home folder if none is set

As part of this I reduces the log level on the gRPC servers since they were a bit overwhelming out of the box (a nice future improvement would be to share gRPC servers between the two processes). We could also discuss what kind of log output we should give for people using this command for the first time to explain what's going on

Feedback on the direction welcome, if it seems promising I will clean it up and add tests.


python_modules/dagster/dagster/_cli/start.py
```diff
@@ -0,0 +1,124 @@
...
+@click.command(
+    name="start",
+    help=(
+        "Start a local deployment of Dagster, including dagit running on localhost and the"
+        " dagster-daemon running in the background"
+    ),
+)
+@start_command_options
+@click.option(
+    "--grpc-log-level",
```
sryza, line 38: Should this be user-code-server-log-level? We use grpc colloquially to refer to these, but it's just the transport protocol, right?

sryza (commented):
I would personally find this very useful.

Have you tested with Ctrl-C to make sure everything gets shut down?

I wonder whether "start" is the right name. Maybe `dagster deployment run`?

python_modules/dagster/dagster/_cli/start.py
gibsondan: we are inconsistent (the command is called dagster api grpc not dagster api code-server) but I agree with the feedback

schrockn:
Seems like a good idea to me. Also agree with @petehunt that this is a good way to start deprecating `dagit` as name.

gibsondan:
I like the simplicity of "dagster start" as the first thing you should try when running dagster but it is definitely a bit of a land grab. I feel like I would struggle to remember "dagster deployment run" but would not struggle to remember "dagster start" though. 

Alternatives: "dagster run-services", "dagster services run"?

sryza:
> I like the simplicity of "dagster start" as the first thing you should try when running dagster 

My pushback here would be that Dagster is not just an application - it's also a programming framework. Often the first thing you do when you "start" with Dagster is create a folder structure, then write some assets, then write some unit tests. Running dagit + daemon is very important, but not necessarily at the beginning of that process.

gibsondan:
Fair enough - I'd love to come up with a solution that addresses that that does not include "deployment" though - that is a very loaded and often ambiguous term

sryza:
> Fair enough - I'd love to come up with a solution that addresses that that does not include "deployment" though - that is a very loaded and often ambiguous term

Do you have thoughts on what the name of the thing is that is started / running after this command completes?

gibsondan:
services?

@petehunt also suggested “dagster dev” which I kind of like to make it clear that it’s for local / dev usage? For example I could imagine making modules hot reload by default when you use this entry point

petehunt:
I think we should go with `dagster dev` as it mirrors `next dev` (https://nextjs.org/docs/api-reference/cli)

sryza:
> services?

Are the daemon and the web server separate services in this nomenclature? Or it's one thing?

> also suggested “dagster dev” which I kind of like to make it clear that it’s for local / dev usage?

I think this arguably suffers from the same issue as `dagster start`. I.e. unlike with `next`, where everything happens with a running web server, there are ways to develop with Dagster that don't require booting up any long-running process.

Maybe I'm being overly pedantic here, but I think it's worth trying to come up with a satisfactory name (app?) for the thing that's being started before choosing a name for the command.

gibsondan:
I think maybe I see the disconnect, I’m viewing this as a replacement in the docs every time we mention dagit so I actually have almost the exact opposite goal as “identify the specific things that are being booted up and call it that”- I’d like to avoid having to identify a new concept or noun that a user has to understand and have it just be the happy path for “running dagster”, especially when people are first getting started.

While it’s true that there are other ways to run dagster code I think “load it up in dagit and view it and run it” is the most common one that we recommend no?

mjclarke94:
If you are spinning up and exiting regularly, it would nice to retain the ephemeral instance type behaviour you get when just running `dagit` without `DAGSTER_HOME` being set.

To avoid polluting our code location, my current workflow to achieve something similar is a wrapper script which: 
- Creates a directory in `/tmp`
- Copies a `dagster.yaml` file through/creates an empty one if none is provided
- Passes this path through to `dagit`/the daemon

I think I prefer the behaviour here of having the `DAGSTER_HOME` directory in the cwd, or preferably a newly created folder within it, as it's easier to explore whilst testing things out. That said, I'd definitely miss the cleanup afterwards. 

Could that be added as an option? Also, would this be a better default behaviour given the current working directory would probably be under VCS and not somewhere you'd want data accumulating?

sryza:
> I’d like to avoid having to identify a new concept or noun that a user has to understand and have it just be the happy path for “running dagster”, especially when people are first getting started.

I think what's problematic about "dagit" is that it's a non-descriptive name, not that it's its own noun. You see it and you have no idea what it's talking about unless you're familiar with the system.  I.e. if there were no daemon, I'd want to call this something like "dagster webserver start".

I think there's an unavoidable distinction between "dagster", which encompasses everything, and the specific thing which is the set of long-running processes that typically run in a Dagster deployment.  Even if we don't codify that distinction in this name, I think we need words to express it when answering user questions, in docs, etc.

gibsondan:
“dagster services start” would be my vote given those constraints. dagit
and the daemon are the two services that need to be running in order to use
the core features of dagster deployments, and services is an existing
well-known concept.

gibsondan:
@mjclarke94 a flag that puts the folder in a tempdir rather than the cwd seems reasonable, yeah - might do that in a followup but the use case makes a ton of sense to me

schrockn:
In my mind what you are doing here is running a local deployment for the purposes of development.  @gibsondan  is your objection to that nomenclature the existence of Instance in our ontology, or a more general English problem?

I oppose `dagster services start` as the command here (would prefer keeping dagit) as it feels _super_ heavyweight for cases like docs and lightweight local development. We want running the UI locally to feel lightweight and well-supported. That aspect of the name `dagit` was the best aspect IMO. The whimsical, specific name made running it locally seem natural and lightweight.

I do like `dagster dev`. It allows us to have a natural entry point to add features like live reloading, dev-only UI (NUX) and other features that _only_ make sense in a development context. I don't think we should use dev because next.js does it, but because of the reason why I listed.

@sryza w/r/t to "there are ways to develop with Dagster that don't require booting up any long-running process", that's interesting. I certainly don't interpret the existence of `dagster dev` as saying this is the _only_ way to do development in Dagster. 

My language in docs would be (post consolidating instance and deployment) would be that you "dagster dev runs a local deployment for you to develop on", then you push and run a branch deployment, and then merge to your production deployment.  IMO the language should center around deployment.

gibsondan:
@schrockn nicely articulated some of the nice things about 'dagit' that I was having trouble putting into words - it feels lightweight and quick, in a way that (to me) 'dagster deployment run' does not. I think the number of characters actually matters here, it's a bit silly but running 'dagit' feels fun and easy to me, even though it's a nonsense word - and running 'dagster deployment run' feels heavy and like work.

I feel like things might be coalescing somewhat around 'dagster dev'? That still feels pretty lightweight to me.

gibsondan:
Renamed to "dagster dev" for now (if we think there might be other commands under 'dev' we could consider a third word after dev to scope it - although as per above every additional character makes it feel more heavyweight)

sryza:
I agree with the importance of brevity.  I'm cool with `dagster dev`.

schrockn:
Time to resurrect live reloading when we push this out? @smackesey do you have that PR stashed away somewhere?

gibsondan:
The change I just made here makes the default behavior if DAGSTER_HOME is not set be ephemeral, just like dagit was before (i.e. it will open a tempdir and delete it when the process ends)

However, one crucial difference with dagit today is that it will still let you have a local dagster.yaml file in the same folder that you ran `dagster dev` in and load that config for the ephemeral instance.

smackesey:
>Time to resurrect live reloading when we push this out? @smackesey do you have that PR stashed away somewhere?

Yes I do: https://github.com/dagster-io/dagster/pull/5939

Before I update it (hasn't been touched in a year), would be good to get clarity on two issues that were debated at the time:

(1) Live-reload potentially has two parts: (a) automatically publishing definition changes to the client; (b) automatically reloading the UI representation of definitions. It is possible to do (a) but then prompt the user to click to refresh the UI. I personally prefer (though with more experience this might change) the fully automatic a-and-b experience, so long as a toaster is shown indicating a reload occurred when the UI representation of definitions updates.

(2) How do we configure this for individual code locations? Add a new param (`live-reload-definitions`?) you can use in `workspace.yaml` or `pyproject.toml` that normally defaults to `False` but defaults to `True` when running `dagster dev`?

gibsondan:
Live reloading will be sweet and this command will be a nice home for it - I think we can handle that separately/next though.

I think this is largely ready to go? I'd like to double-check that it works on Windows and then i think we are good. Any reviewers want to take a pas?

I also just want to make sure that reviewers are aware of the slightly different semantics about dagster.yaml after this (in keeping with the goal of 'people should be able to experiment with dagster features without having to think about env vars'):

- if you don't have DAGSTER_HOME set, dagster dev (and dagit now) will look for a dagster.yaml file in the local folder and apply that as the instance config (using a child tempdir as the base dir for any storage by default that is cleaned up afterwards)
- If you do have DAGSTER_HOME set it will work like it does today: storage and the dagster.yaml both need to be in that folder (I added a helpful warning if you have a local dagster.yaml and DAGSTER_HOME set that it's ignoring the local dagster.yaml file)

schrockn:
Yeah obviously not blocking on the live reloading. I have a hard stop rt now but can look at code this evening after 8 PM ET

python_modules/dagster/dagster/_core/instance/ref.py
```diff
@@ -310,10 +310,15 @@ def config_defaults(base_dir: str) -> Mapping[str, Optional[ConfigurableClassDat
         }
 
     @staticmethod
-    def from_dir(base_dir, config_filename=DAGSTER_CONFIG_YAML_FILENAME, overrides=None):
+    def from_dir(
+        base_dir, config_dir=None, config_filename=DAGSTER_CONFIG_YAML_FILENAME, overrides=None
```
schrockn, line 314: seems dangerous to add this in the middle here without requiring keyword args.

python_modules/dagster/dagster/_core/workspace/context.py
```diff
@@ -432,6 +432,7 @@ def __init__(
         version: str = "",
         read_only: bool = False,
         grpc_server_registry=None,
+        code_server_log_level="INFO",
```
schrockn, line 435: would be nice to have all these args typed, if nothing else aesthetically so the spacing around = is the same

python_modules/dagster/dagster/_core/instance/ref.py
schrockn: Looking at the callsites, requiring kwargs after base_dir also an option.

schrockn (approved):
Please heed final comments, especially around argument ordering of `from_dir`.

gibsondan:
> (2) How do we configure this for individual code locations? Add a new param (`live-reload-definitions`?) you can use in `workspace.yaml` or `pyproject.toml` that normally defaults to `False` but defaults to `True` when running `dagster dev`?

@smackesey I don't think it should be property of `pyproject.toml` or individual locations since those files are checked into source control. I think it should be a property of the CLI session, i.e. a flag that you pass into `dagster dev` (like `dagster dev --watch` or something) or even the default behavior when running the new command.

gibsondan:
> (1) Live-reload potentially has two parts: (a) automatically publishing definition changes to the client; (b) automatically reloading the UI representation of definitions. It is possible to do (a) but then prompt the user to click to refresh the UI. I personally prefer (though with more experience this might change) the fully automatic a-and-b experience, so long as a toaster is shown indicating a reload occurred when the UI representation of definitions updates.

I think we should start with a) which will be a huge step up and then investigate getting fully automatic working - i don't remember every detail but vaguely recall it required non-trivial JS work for fully automatic to work correctly with all the various frontend caches at play

python_modules/dagster/dagster/_cli/dev.py
```diff
@@ -0,0 +1,168 @@
...
+                        f" {daemon_process.returncode}"
+                    )
+
+        except BaseException as e:
+            logger.info("Shutting down Dagster services...")
+
+            # If it's a KeyboardInterrupt / SIGINT that probably means that the whole
+            # process group is being interrupted by a CTRL-C, so give the subprocesses a
+            # grace period to stop on their own. Otherwise, interrupt them immediately
+            grace_period = 30 if isinstance(e, KeyboardInterrupt) else 0
```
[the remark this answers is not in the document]
gibsondan: This whole gross thing goes away after https://github.com/dagster-io/dagster/pull/11737 (good riddance)

python_modules/dagster/dagster/_cli/dev.py
```diff
@@ -0,0 +1,168 @@
...
+    ),
+)
+@click.option("--dagit-port", help="Port to use for the Dagit UI.", required=False)
+def dev_command(code_server_log_level, dagit_port, **kwargs):
+    # check if dagit installed, crash if not
+    try:
+        import dagit  #  # noqa: F401
+    except ImportError:
+        raise click.UsageError(
+            "The dagit package must be installed in order to use the dagster dev command."
```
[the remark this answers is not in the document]
gibsondan: @schrockn do we want to assume pip though? what about poetry / conda / etc

PR #11587, opened 2023-01-09
[dagit] Add location-scoped permissions to Dagit

hellendag:
### Summary & Motivation

Begin adding location-scoped permissions to Dagit. Changes include:

- `usePermissionsForLocation()`, a hook that accepts a location name and returns its relevant permissions object.
  - Intended for use whenever we have a location-specific permission to look up.
- `useUnscopedPermissions()`, a hook that returns the "global" permission set. 
  - Intended for use on permissions that are never locally scoped, or for edge cases where we have no relevant location to associate with the UI.
  - Identical to the existing `usePermissions()`.
- `usePermissionsDEPRECATED()`, an alias of `useUnscopedPermissions()` that I put in place where I couldn't immediately apply a location for the permission lookup, but that it looks like we should have one.

In cases where we look up a permission with a location scope but none exists for that location (e.g. because it's not intended to be location-scoped, or for some reason we haven't defined it) it will fall back to the global unscoped permission set.

I also left `usePermissions()` temporarily in place for Cloud compatibility. Will remove after Cloud code is updated.

Once this is in, we should go through the `usePermissionsDEPRECATED()` callsites and either thread through the correct location to the callsite or change it to a `useUnscopedPermissions()` call.

jest Permissions to verify fallback behavior.

Load Dagit, verify that permissions appear to be working correctly throughout.


hellendag:
* master
  * **PR #11587** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11587" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11587?utm_source=stack-comment).

salazarm:
Any reason these need to be hooks? Seems like every hook has it's own React.memo caching the exact same data rather than there being one global cache? Or at least a cache at the Context level rather than having every component store it.

hellendag:
> Any reason these need to be hooks?

The permission data comes from GraphQL (`useQuery`) and is made available via a provider (`useContext`). So we're going  through a hook to get the information one way or another.

It would probably make sense to do the extraction at the provider level, though. I don't recall why it's not done there.

js_modules/dagit/packages/core/src/app/Permissions.tsx
```diff
@@ -55,60 +58,122 @@ const extractPermissions = (permissions: PermissionFragmentFragment[]) => {
...
+ * have a valid value.
+ */
+export const usePermissionsForLocation = (locationName: string | null | undefined) => {
+  const {fallbackPermissions, locationPermissions, loading} = React.useContext(PermissionsContext);
+  return React.useMemo(() => {
+    const targetPermissions = locationName
+      ? locationPermissions[locationName]
+      : fallbackPermissions;
+    return {
+      ...extractPermissions(targetPermissions || [], fallbackPermissions),
```
bengotow, line 149: This all looks good! It looks like until the data has loaded, the user has no permissions which seems good. That said I wonder if there's any chance the cache could include fallback permissions but an old version of the location data, resulting in a user briefly getting fallback permissions for a new location instead of the location permissions?

Sounds like it won't matter once the migration is complete and the hook usage is updated?

bengotow (approved):
This looks great! I like the migration approach you set up too, seems like this is a nice way to transition things 🙌

hellendag:
Test timeout in Cloud is unrelated.

PR #11588, opened 2023-01-09
[dagit] Stop foreign node links from creating sprawling asset graphs

bengotow:
### Summary & Motivation

This is a fix for one of the most pressing issues @braunjj identified in his asset graph explorations -- we weren't capping the length of external graph links on the Asset Graph, so they could easily result in horizontal expansion of the graph.

This PR limits them to 10 characters, which is enough that two can be shown side-by-side above an asset node without increasing the graph width. We could potentially just show the icon or redesign these futher, but this should be a good incremental improvement.

Before:
Screen Shot 2023-01-09 at 2 34 33 PM

After:
Screen Shot 2023-01-09 at 2 34 20 PM


braunjj:
This is great

js_modules/dagit/packages/core/src/asset-graph/layout.ts
```diff
@@ -221,13 +222,7 @@ export const extendBounds = (a: IBounds, b: IBounds) => {
...
 };
 
-export const ASSET_NODE_ICON_WIDTH = 20;
-export const ASSET_NODE_NAME_MAX_LENGTH = 32;
-const DISPLAY_NAME_PX_PER_CHAR = 8.0;
-
-export const assetNameMaxlengthForWidth = (width: number) => {
-  return (width - ASSET_NODE_ICON_WIDTH) / DISPLAY_NAME_PX_PER_CHAR;
-};
+export const ASSET_NODE_NAME_MAX_LENGTH = 28;
```
bengotow, line 225: This is based on the 255px width below, minus the box padding, outline, icon, margin from icon to text, etc. Thought about making const vars for all the values, but it's a bunch of layers...

hellendag (approved):

PR #11360, opened 2022-12-23
[structured config] Allow nesting structured config resources

benpankow:
## Summary

Allows structured-config-backed resources (#11321) to require other resources as input. By default, this doesn't work because the resource class is interpreted as nested config.

```python
class WriterResource(Resource):
    prefix: str

    def output(self, text: str) -> None:
        out_txt.append(f"{self.prefix}{text}")

class JsonWriterResource(Resource):
    writer: WriterResource
    indent: int

    def output(self, obj: Any) -> None:
        self.writer.output(json.dumps(obj, indent=self.indent))

writer_resource = WriterResource(prefix="greeting: ")
json_writer_resource = JsonWriterResource(writer=writer_resource, indent=2)
```

## Test Plan

Basic unit test.

PR #11589, opened 2023-01-09
[structured config] Allow specifying unconfigured struct config resources

benpankow:
## Summary

Allows using structured-config based resources and IO managers without binding config at defintiions-time, e.g.:

```python

class MyResource(Resource):
    a_string: str

@asset
def my_asset(res: MyResource):
    print(res.a_string)


defs = Definitions(
    assets=[my_asset],
    resources={"res": MyResource}
)
```
Screen Shot 2023-01-09 at 12 40 57 PM

## Test Plan

Add unit tests for resources, IO managers.

PR #11595, opened 2023-01-09
[structured config] Allow structured config resources, IO managers to depend on other resources

benpankow:
## Summary

Allows users of structured config resources/IO managers to specify resource-to-resource dependencies. These are introduced through ineriting the `RequiresResources` generic class:

```python
class AWSCredentials(Resource):
    username: str
    password: str

# Specifies a list of resource dependencies
class AWSDeps(Config):
    aws_creds: AWSCredentials

class S3Resource(Resource, RequiresResources[AWSDeps]):
    bucket: str

class EC2Resource(Resource, RequiresResources[AWSDeps]):
    pass

@asset
def my_asset(s3: S3Resource, ec2: EC2Resource):
    assert s3.required_resources.aws_creds.username == "foo"

defs = Definitions(
    assets=[my_asset],
    resources={
        "aws_creds": AWSCredentials(username=EnvVar("AWS_USER"), password=EnvVar("AWS_PASSWORD")),
        "s3": S3Resource(bucket="my_bucket"),
        "ec2": EC2Resource(),
    },
)
```

### Alternatives Considered

Another route, explored earlier in #11360, would specify resource-resource dependencies more simply as parameters, e.g.

```python
class AWSCredentials(Resource):
    username: str
		password: str

class S3Resource(Resource):
    creds: Credentials
    bucket: str

class EC2Resource(Resource):
    creds: Credentials

@asset
def my_asset(s3: S3Resource, ec2: EC2Resource):
    assert s3.creds.username == "foo"

aws_creds = AWSCredentials(
	username=EnvVar("AWS_USER"), 
	password=EnvVar("AWS_PASSWORD")
)

defs = Definitions(
	assets=[my_asset],
	resources={	
  	        "s3": S3Resource(creds=aws_creds, bucket="my_bucket"),
		"ec2": EC2Resource(creds=aws_creds),
	}
)
```

However, this approach runs into trouble when runtime config comes into play:
```python
# Unconfigured AWS credentials, should be provided in launchpad
aws_creds = AWSCredentials

defs = Definitions(
    assets=[my_asset],
    resources={
        "s3": S3Resource(creds=aws_creds, bucket="my_bucket"),
        "ec2": EC2Resource(creds=aws_creds),
    },
)

# What schema to display? Unclear where AWSCredentials config lives in hierarchy. 
# Can't be nested within one of the dependent resources because it affects both.
# In addition, resource is "anonymous"
# since it's not included in resource dict, it has no identifier associated with it.
```

I'm still open to this approach since the ergonomics seem better, but I'm not sure how to reconcile this issue. Since inner resources are by nature "anonymous", it's hard to keep track of all the places they are used.

One potential way to reconcile this would be to just require any resource which is not fully configured at definition-time to be specified at the top level.

## Test Plan

Unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_config/structured_config.py
+++ b/python_modules/dagster/dagster/_config/structured_config.py
@@ -14,7 +14,7 @@ except ImportError:
 
 
 from abc import ABC, abstractmethod
-from typing import Any, Optional, Type, cast
+from typing import Any, Generic, Optional, Set, Type, TypeVar, cast, get_args, get_origin
 
 from pydantic import BaseModel, Extra
 from pydantic.fields import SHAPE_SINGLETON, ModelField
@@ -94,6 +94,21 @@ def _curry_config_schema(schema_field: Field, data: Any) -> IDefinitionConfigSch
     return configured_resource_def.config_schema
 
 
+RequiredResources = TypeVar("RequiredResources")
+
+
+class RequiresResources(Generic[RequiredResources]):
+    _resources: Optional[RequiredResources] = None
+
+    @property
+    def required_resources(self) -> RequiredResources:
+        check.invariant(
+            self._resources is not None,
+            "Attempted to access required_resources before they were set.",
+        )
+        return self._resources
+
+
 class Resource(
     ResourceDefinition,
     Config,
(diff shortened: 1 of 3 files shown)


benpankow:
* master
  * **PR #11589** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11589" target="_blank">Graphite</a> 
    * **PR #11595** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11595" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11595?utm_source=stack-comment).

schrockn:
Consider the use case where a resource depends on two other resources with the same type but with different keys/names.

Should we just change our approach to fully unconfigured resources by just having it be an instantiation with no parameters as opposed to the class?

benpankow:
> Consider the use case where a resource depends on two other resources with the same type but with different keys/names.

That should work with this setup - I've added a test to show how:

```python
class MyProdIOManager(StructuredConfigIOManager):
    def handle_output(self, context: OutputContext, obj):
        prod_data_store[context.name] = obj

    def load_input(self, context: InputContext):
        return prod_data_store.get(context.name)

class MyStagingIOManager(StructuredConfigIOManager):
    def handle_output(self, context: OutputContext, obj):
        staging_data_store[context.name] = obj

    def load_input(self, context: InputContext):
        return staging_data_store.get(context.name)

class BranchingIOManagerDeps(Config):
    prod_io_manager: IOManager
    staging_io_manager: IOManager

class BranchingIOManager(StructuredConfigIOManager, RequiresResources[BranchingIOManagerDeps]):
    def handle_output(self, context: OutputContext, obj):
        self.required_resources.staging_io_manager.handle_output(context, obj)

    def load_input(self, context: InputContext):
        staging_value = self.required_resources.staging_io_manager.load_input(context)
        if staging_value is not None:
            return staging_value

        return self.required_resources.prod_io_manager.load_input(context)
```

> Should we just change our approach to fully unconfigured resources by just having it be an instantiation with no parameters as opposed to the class?

That feels nice but is hard in Python without overloaded constructors. One option could be to make all the fields on the Pydantic class optional, but that makes it hard for the user to distinguish which fields are required or not.

PR #11596, opened 2023-01-09
[dagster-dbt] parse log level when json_log_level is not set

OwenKephart:
### Summary & Motivation

We can make a pretty good guess about what log level certain lines should be, so let's do that.


sryza (commented):
The fuzziness of this sketches me out a bit. We think there's a big benefit?

sryza (approved):

PR #11599, opened 2023-01-10
Remove toposort dep for py>=3.9

adriangb:
### Summary & Motivation

This removes the `toposort` dependency for Python >=3.9 with no new dependencies. See https://github.com/dagster-io/dagster/discussions/11598 for reasoning.

Ran `python -m pytest python_modules/dagster/dagster_tests` with no errors.

I'm also curious about two things:
1. Could `TopologicalSorter`'s API be useful in places? Instead of grouping dependencies into "levels" it lets you dynamically mark dependencies as completed and get all newly unblocked dependencies. This seems like a great match for crunching through a DAG of computation given, but I didn't dig too deep into how Dagster is currently resolving the DAG or even if it is using the `toposort` module for that.
2. `toposort` could be removed altogether by using a graphlib backport (https://pypi.org/project/graphlib-backport/ or https://pypi.org/project/graphlib2/) which would also simplify things implementation wise.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -16,11 +16,10 @@ from typing import (
     cast,
 )
 
-import toposort
-
 import dagster._check as check
 from dagster._core.errors import DagsterInvalidInvocationError, DagsterInvariantViolationError
 from dagster._core.selector.subset_selector import DependencyGraph, generate_asset_dep_graph
+from dagster._core.utils import toposort
 
 from .assets import AssetsDefinition
 from .events import AssetKey, AssetKeyPartitionKey
@@ -305,7 +304,7 @@ class AssetGraph:
 
     def toposort_asset_keys(self) -> Sequence[AbstractSet[AssetKey]]:
         return [
-            {key for key in level} for level in toposort.toposort(self._asset_dep_graph["upstream"])
+            {key for key in level} for level in toposort(self._asset_dep_graph["upstream"])
         ]
 
     def has_self_dependency(self, asset_key: AssetKey) -> bool:
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -303,9 +303,7 @@ class AssetGraph:
         return set()
 
     def toposort_asset_keys(self) -> Sequence[AbstractSet[AssetKey]]:
-        return [
-            {key for key in level} for level in toposort(self._asset_dep_graph["upstream"])
-        ]
+        return [{key for key in level} for level in toposort(self._asset_dep_graph["upstream"])]
 
     def has_self_dependency(self, asset_key: AssetKey) -> bool:
         return asset_key in self.get_parents(asset_key)
--- a/python_modules/dagster/dagster/_core/utils.py
+++ b/python_modules/dagster/dagster/_core/utils.py
@@ -17,6 +17,7 @@ if sys.version_info >= (3, 9):
         ts: "TopologicalSorter[T]" = TopologicalSorter()
         for node, predecessors in dag.items():
             ts.add(node, *predecessors)
+        ts.prepare()
         res: List[Set[T]] = []
         while ts.is_active():
             ready = ts.get_ready()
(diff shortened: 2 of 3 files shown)


adriangb:
@dpeng817 mind considering this / tagging someone for review?

dpeng817:
https://buildkite.com/dagster/dagster/builds/45187 - build for this PR

adriangb:
Seems like I need some auth to view the build pipeline results?

dpeng817:
Especially given your edit, this does seem like a necessary change. Once tests pass I think this should be good to go, will loop in @sryza on whether something like a backport will be necessary given that it sounds like toposort might end up unmaintaned, and we do make consistent usage of it in important parts of the codebase (asset graph, resource to resource dependencies, core execution plan logic).

dpeng817:
> Seems like I need some auth to view the build pipeline results?

yup will reply with what (if anything) may need to be changed / build results

adriangb:
Thanks. From local testing it seemed like 1 test was failing because of that bug in toposort, but maybe I'm not running all of the tests.

sryza:
Thanks for your work on this @adriangb.

I just clicked into the Snyk report on toposort, and it's not clear to me that the reasons they bring up mean we should avoid this library. Their main strikes against it are that there are few maintainers and it hasn't received updates recently, but I think that this is a reasonable state to be in for small libraries that provide basic functionality - i.e. it's in a steady state and doesn't need updates.

We do our own cycle detection, so a toposort bug with cycles likely isn't problematic.

If we had the opportunity to entirely ditch toposort, I'd definitely be in favor of this change. But it's not clear to me that the issues with toposort merit the added complexity of maintaining two separate codepaths. Any time we have divergent implementations across Python versions, it makes it easier for bugs to slip through.

adriangb:
> We do our own cycle detection, so a toposort bug with cycles likely isn't problematic.

Well I found the bug in toposort by running your tests which pass with toposort but fail with graphlib, so at least in some places cycles aren't being detected correctly.

It's ultimately up to you as to what path you take, but IMO if the author says they plan to move off of their library because it is now part of the standard library I would strongly consider following their lead.

adriangb:
If you _do_ want to replace toposort with graphlib and backports (which would be a single codepath) I'm happy to help with that.

sryza:
> If you do want to replace toposort with graphlib and backports (which would be a single codepath) I'm happy to help with that.

Ah yes, if we could use a single codepath, that sounds great.

FYI, I think this current PR misses a few places where we're using toposort:

```
❯ gg toposort | grep import
python_modules/dagster/dagster/_core/definitions/asset_graph.py:import toposort
python_modules/dagster/dagster/_core/definitions/assets_job.py:from toposort import CircularDependencyError, toposort
python_modules/dagster/dagster/_core/definitions/graph_definition.py:from toposort import CircularDependencyError, toposort_flatten
python_modules/dagster/dagster/_core/execution/plan/plan.py:from dagster._core.utils import toposort
python_modules/dagster/dagster/_core/execution/resources_init.py:from dagster._core.utils import toposort
python_modules/dagster/dagster/_core/host_representation/external.py:from dagster._core.utils import toposort
python_modules/dagster/dagster/_core/snap/pipeline_snapshot.py:from dagster._core.utils import toposort_flatten
python_modules/dagster/dagster/_core/utils.py:import toposort as toposort_
python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_group.py:    from toposort import CircularDependencyError
```

> Well I found the bug in toposort by running your tests which pass with toposort but fail with graphlib, so at least in some places cycles aren't being detected correctly.

Out of curiosity, do you recall which test is failing?

adriangb:
> FYI, I think this current PR misses a few places where we're using toposort

I'll take a look at those spots. In the meantime before I invest more time into this it would be great to know if you are on board with replacing toposort and if so if you only want to do it for 3.9+ or if you want to use the graphlib backports as well. The implementation should be straightforward since there's nothing toposort does that TopologicalSorter can't do; it's just busywork.

> Out of curiosity, do you recall which test is failing?

It's `test_scenario_to_completion[one_asset_self_dependency-no_failures-all]`.  I set a breakpoint right above [this line](https://github.com/dagster-io/dagster/blob/458936bb848fd26902869a9e3912244ee0ba338a/python_modules/dagster/dagster_tests/core_tests/execution_tests/test_asset_backfill.py#L74) and inspected:

```python
asset_graph._asset_dep_graph["upstream"]
{AssetKey(['asset1']): {AssetKey(['asset1'])}}
```

Which is indeed a cycle.

adriangb:
> FYI, I think this current PR misses a few places where we're using toposort:

Okay I think I got all of these now. FYI some of them were false positives (e.g. `from dagster._core.utils import toposort_flatten`).

I also set the package up with the `graphlib2` backport for Python < 3.9. Happy to use `graphlib-backport` instead.

sryza:
Hey @adriangb - I chatted with the team about this, and we decided that we'd like to stick with the toposort package for now.  The reasoning is a combination of:
1. "If it ain't broke, don't fix it".  
2. For Python versions below 3.9, we'd be replacing toposort with other dependencies that don't have lots of maintenance, which defeats some of the advantages.

We really appreciate your work and concern here, and want to make sure that we don't lead you to spend time on changes that we don't plan to incorporate.

As an aside, with regards to that specific failing test, we actually do support some cycles in the Dagster asset graph: an asset can depend on earlier partitions.

adriangb:
Fair enough, thanks for explaining!

PR #11605, opened 2023-01-10
Fix graphene enum not being instantiated

gibsondan:
Summary:
I'm unsure when exactly this broke but I am suspicious of the graphql 3 migration? Unclear. At any rate, this resolves an error where eventType could not be instantiated b/c we were passing an instance of the Python enum into the identically typed Graphene class.

Test Plan:
Tried to write an automated graphql test and was unsuccessful at getting the subscription to actually return something, but:
- load a grpc server
- load dagit pointing at it
- shut down grpc server, no more error in dagit
- restart grpc server, code reloads in dagit


alangenfeld (approved):
bummer that its hard to get it under test

PR #11607, opened 2023-01-10
[freshness-policies] Add the ability to calculate the used_data_time of a TimeWindowPartitioned asset

OwenKephart:
### Summary & Motivation

This change solves two related problems.

1. Previously, we could not support FreshnessPolicies on time-based partitioned assets, because the way we calculated data times for assets (to see if they're fresh enough or not) involved recursively looking at the upstream materialization events that were consumed to produce that record. In general, this calculation would start at the most recent materialization record for that asset. For partitioned assets, the current state depends on all materializations of that asset, and in particular the data time is a function of which partitions have been materialized, not the time at which those materializations were created.
2. Similarly, if a non-partitioned asset was downstream of a partitioned asset, its data time would be calculated based on the most recent partition of that asset, which is a very incorrect method of calculation.

#### Algorithm

The method of calculation takes advantage of @clairelin135 's changes, which allow us to efficiently compute the currently materialized partitions of a given asset.

If we start from the (incorrect) assumption that all non-partitioned records for which we wish to calculate the used data time have consumed the latest state of all partitioned records, then everything is fairly simple. We just calculate which time windows of our partitioned asset have been materialized, and sort them.

If the first partition of the PartitionsDefinition has not been materialized, then we say we have a data time of `None`, otherwise, we take the end of the first included time window (i.e. if you have `2022-01-01 -> 2022-03-04` and `2022-05-05 -> 2022-07-01`, then we say that your data time is `2022-03-04`, as you have all available data up until that point)

However, what if new partitions have been materialized after the record? In this case, we figure out the set of net-new partitions that have been factored into the current cached state since this asset was consumed. We have a record id of the time at which this asset was consumed, so we can query for all partitions materialized after that point. We can count how many times each of these partitions was materialized in this time period, and compare it to the number of times that partition has been materialized overall. If those counts match, then that partition did not exist when this record was consumed, and we should exclude it from the set of partitions that make up those time windows.

This results in an accurate accounting of which partitions were available at any given point in time.

#### Assumptions

This algorithm requires linearly more data as the number of materializations of the partitioned asset since it was consumed by the non-partitioned asset grows. So if you materialize your partitioned asset once, then consume it with a non-partitioned asset, then materialize your non partitioned asset 10,000 times, then calculate the data time of your non-partitioned asset, then you'll need to query for all 10,000 of those new records in order to create an accurate diff between the current cached state and the cached state at the time of materialization.

In practice, this is unlikely. If you calculated the data time of the non-partitioned asset at any point before the end of that process, its value would be cached in the asset tags table, short-circuiting the need for the expensive calculation at the end.

Another point is that the `get_materialization_count_by_partition` call may be fairly slow (and this is necessary even if there is a single new materialization since the partitioned asset was last consumed). This can be made more efficient if this function were modified to accept an optional sequence of partition keys to filter the query for, as we do not use the counts of any partitions other than those materialized in the time between the record emitted and the current time.

#### Note

Support for freshness-based scheduling for partitioned assets will be added in a follow-up PR


python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -315,19 +321,123 @@ def _upstream_subset(
...
+            get_and_update_asset_status_cache_values,
+        )
+
+        # return the existing upstream data times already calculated for this record (if any)
+        known_data = self.get_known_used_data(asset_key, record_id)
+        if known_data:
+            return known_data
+
+        # this is the current state of the asset, not the state of the asset at the time of record_id
+        status_cache_value = get_and_update_asset_status_cache_values(
```
clairelin135, line 416: How will this work if the user hasn't run the migration to add the asset status cache data? I think we will need to add a check to `instance.can_cache_asset_status_data`. If it's not possible to cache, we'll just need to query for partition status via the counts.

clairelin135 (changes requested):
The method of calculating the partitions present at a current time makes sense to me.

I think we just need to guard against the possible states of migration--it's possible that the migration to add the cache status data column hasn't been run, or that neither the asset tags table or cache status data column migrations have been run. Though I'm sure optimized behavior with these migrations will motivate users to run them :)

To my knowledge the worst case scenario is if neither migration has been run, and dagster needs to generate the counts of all the partitions. We already do this on every partitions page load in Dagit, but how many records will this operation happen on?

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -315,19 +321,123 @@ def _upstream_subset(
...
+        # if the first partition has not been filled
+        if first_available_time_window.start < first_filled_time_window.start:
+            return None
+
+        # there are no events for this asset after the given record
+        if status_cache_value.latest_storage_id <= record_id:
+            return first_filled_time_window.end
+
+        # get the new records
+        new_records = list(
```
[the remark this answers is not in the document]
OwenKephart: that's a great idea -- I think we could take advantage of that elsewhere in the reconciliation code paths as well

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
OwenKephart: good call

OwenKephart:
@clairelin135 In lieu of automated regression tests, I did some pre-emptive perf profiling on this. With around 500 assets, daily partitioned, and with a bit of historical data, we can complete a tick in around 20 seconds. Definitely significantly slower than the non-partitioned case (which only takes around 3 seconds).

As you might expect, the the majority of the time is spent in _get_and_update_asset_status_cache values. I think the next logical step on that front would be to add a way to call that method for multiple assets at once, batching some of the queries. However, I think supporting up to 500 partitioned assets is ok for now, and we can push on the perf angle in a future PR.

OwenKephart:
also confirmed that the data times show up correctly in dagit for partitioned assets w/ freshness policies

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -399,12 +409,116 @@ def get_known_used_data(
             }
         return {}
 
+    def _calculate_partition_data_time(
```
sryza, line 412: the name makes me think this is calculating the data time for a single partition - maybe `_calculate_time_partitioned_asset_data_time`?

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -399,12 +409,116 @@ def get_known_used_data(
...
+        self,
+        asset_key: AssetKey,
+        record_id: int,
```
sryza, line 415: I think it's worth documenting the meaning of the record_id here

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -399,12 +409,116 @@ def get_known_used_data(
...
+        partitions_def: TimeWindowPartitionsDefinition,
+        status_cache_value: "AssetStatusCacheValue",
+    ) -> Optional[datetime.datetime]:
```
sryza, line 418: I think some documentation of what's going on here at a high level (like included in the PR description) would be helpful:
1. How the data time for a time-partitioned asset is defined.
2. How we use the status_cache_value to compute it

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -470,6 +584,35 @@ def _calculate_used_data(
...
+
+        partitions_def = asset_graph.get_partitions_def(asset_key)
+        if isinstance(partitions_def, TimeWindowPartitionsDefinition):
+            return self._calculate_used_data_partitioned(
+                asset_graph=asset_graph,
+                asset_key=asset_key,
+                record_id=record_id,
+                partitions_def=partitions_def,
+            )
+        else:
```
sryza, line 607: Should we explicitly handle the case where it's partitioned, but not time-partitioned? Raise an error?

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -470,6 +584,35 @@ def _calculate_used_data(
...
 
+    @cached_method
+    def _calculate_used_data(
+        self,
+        asset_graph: AssetGraph,
+        asset_key: AssetKey,
+        record_id: Optional[int],
+        record_timestamp: Optional[float],
+        record_tags: Mapping[str, str],
+    ) -> Dict[AssetKey, Tuple[Optional[int], Optional[float]]]:
```
sryza, line 595: Could this return a `Mapping`?

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -399,12 +409,116 @@ def get_known_used_data(
...
+        )[asset_key]
+
+        partition_data_time = self._calculate_partition_data_time(
+            asset_key=asset_key,
+            record_id=record_id,
+            partitions_def=partitions_def,
+            status_cache_value=status_cache_value,
+        )
+        partition_data_timestamp = partition_data_time.timestamp() if partition_data_time else None
+        from dagster import AssetSelection
```
sryza, line 511: we typically do these imports at the top of the function. also, since it's internal, shouldn't import from the top-level package

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -399,12 +409,116 @@ def get_known_used_data(
...
+            first_filled_time_window.end,
+        )
+
+    def _calculate_used_data_partitioned(
+        self,
+        asset_graph: AssetGraph,
+        asset_key: AssetKey,
+        record_id: int,
+        partitions_def: TimeWindowPartitionsDefinition,
+    ) -> Dict[AssetKey, Tuple[Optional[int], Optional[float]]]:
```
sryza, line 487: I think it's worth documenting what "used data" means for a partitioned asset, if it's different than for a non-partitioned asset.

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
OwenKephart: whoops this was me experimenting -- forgot to clean that up

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
OwenKephart: Definitely reasonable, I'm inclined to leave it as is for now to avoid changing existing behavior.

sryza: Can we address this in a followup then?

sryza: Or I guess - what is the existing behavior?

sryza: What is throwing me off here is that we're calling `_calculate_used_data_unpartitioned` on an asset that might be partitioned

sryza (approved):

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
OwenKephart: yeah definitely can address in a followup -- existing behavior was that we would treat a static-partitioned asset as an unpartitioned asset. This is not correct behavior, but throwing an exception may break existing workflows that "kinda work". Would rather have a smoother landing here than break it

clairelin135 (approved):
LGTM!

PR #11617, opened 2023-01-10
[dagster-dbt] allow for non-model nodes to be part of the DbtManifestAssetSelection

OwenKephart:
### Summary & Motivation

In the regular asset creation code, we mush together all of the dictionaries to make a single dictionary addressed by unique_id to select from. This does the same thing in the DbtManifestAssetSelection codepath.

If we don't do this, you can hit errors if you pass in a selection like `+foo`, where `foo` is downstream of a source, because that source's unique_id will not be a key of the `self.manifest_json["nodes"]` dictionary.


sryza (approved):
Is there a test we could include?

PR #11619, opened 2023-01-10
[freshness-policies] fix get_non_source_roots

OwenKephart:
### Summary & Motivation

Adds an extra guard for the case that a referenced asset is not part of self.all_asset_keys (which can happen with nothing-type dependencies that have no associated SourceAsset)

Test failed before the change, passes now.


sryza (approved):
Is it worth adding a reconciliation sensor test case that covers the case that was breaking here?

PR #11620, opened 2023-01-10
[airbyte] Consider keys containing credentials a secret

joel-olazagasti:
### Summary & Motivation

There are lots of keys which should be secret, such as any of google api's `credentials_json` fields that are currently not masked. The Airbyte api also does not return these fields, so they are always considered a diff, even if the credential is correctly applied

I made the requested change locally, and then tested the CLI `apply` and `check` commands. Fields with `credentials_json` were correctly masked and applied


benpankow (approved):
Looks good!

joel-olazagasti:
Hey @benpankow @sryza any further action needed from me here for a merge?

benpankow:
Nope, looks good to me. Will merge 👍

PR #11621, opened 2023-01-10
Fix typo in Dagster instance documentation

GrigoriiKushnir:
### Summary & Motivation
Noticed that the https://docs.dagster.io/deployment/dagster-instance has a wrong tab title:

dagster insTNace instead of dagster insTAnce


sryza (approved):
Thanks for the fix!

PR #11624, opened 2023-01-11
telemetry: add step start/success/failure event + client timestamp

yuhan:
### Summary & Motivation
add 3 additional data collection points:
- step_start_event
- step_success_event, which comes with duration_ms
- step_failure_event
with the `client_time` so we can reconstruct the duration in post-processing (based on the same `run_id_hash` and `step_key_hash` within the same `instance_id`)

code for testing:
```python
from dagster import op, job, DagsterInstance
import time


@op
def step_1():
    time.sleep(1)


@op
def step_2():
    time.sleep(2)


@op
def step_3():
    time.sleep(3)


@op
def step_4():
    time.sleep(4)
    raise Exception("fail at 4s")


@job
def my_job():
    step_1()
    step_2()
    step_3()
    step_4()

if __name__ == "__main__":
    my_job.execute_in_process(instance=DagsterInstance.get())
```

results:
- logs from running via api: https://gist.github.com/yuhan/cd9fa453683a6b7463f5651289b71ae0
- logs from running via dagit: https://gist.github.com/yuhan/d160d01cf12a67ceaa4c6234fa13a21a

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/api.py
+++ b/python_modules/dagster/dagster/_core/execution/api.py
@@ -35,7 +35,7 @@ from dagster._core.instance import DagsterInstance, InstanceRef
 from dagster._core.selector import parse_step_selection
 from dagster._core.storage.pipeline_run import DagsterRun, DagsterRunStatus
 from dagster._core.system_config.objects import ResolvedRunConfig
-from dagster._core.telemetry import log_repo_stats, telemetry_wrapper
+from dagster._core.telemetry import log_repo_stats, log_step_event, telemetry_wrapper
 from dagster._core.utils import str_format_set
 from dagster._utils.error import serializable_error_info_from_exc_info
 from dagster._utils.interrupts import capture_interrupts
@@ -1095,6 +1095,9 @@ def pipeline_execution_iterator(
                 failed_steps.append(event.step_key)
             elif event.is_resource_init_failure and event.step_key:
                 failed_steps.append(event.step_key)
+
+            # Telemetry
+            log_step_event(event, pipeline_context)
 
             yield event
     except GeneratorExit:
--- a/python_modules/dagster/dagster/_core/execution/api.py
+++ b/python_modules/dagster/dagster/_core/execution/api.py
@@ -35,7 +35,7 @@ from dagster._core.instance import DagsterInstance, InstanceRef
 from dagster._core.selector import parse_step_selection
 from dagster._core.storage.pipeline_run import DagsterRun, DagsterRunStatus
 from dagster._core.system_config.objects import ResolvedRunConfig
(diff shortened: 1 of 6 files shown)


yuhan:
* master
  * **PR #11624** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11624" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11624?utm_source=stack-comment).

python_modules/dagster/dagster/_core/telemetry.py
```diff
@@ -523,6 +528,40 @@ def log_action(
...
+    if event.is_step_start:
+        log_action(
+            instance=pipeline_context.instance,
+            action=STEP_START_EVENT,
+            client_time=datetime.datetime.now(),
+            metadata={
+                "run_id_hash": hash_name(pipeline_context.run_id),
+                "step_key_hash": hash_name(event.step_key),
+            },
+        )
```
rexledesma, line 541: nit: would probably be more terse to 

1/ have an early return for non-step events
2/ use the conditionals to determine the `action` and `metadata`, and have one `log_action` invocation once those are determined

```suggestion
def log_step_event(event: DagsterEvent, pipeline_context: PlanOrchestrationContext):
    if not any((event.is_step_start, event.is_step_success, event.is_step_failure)):
        return
    
    metadata = {
        "run_id_hash": hash_name(pipeline_context.run_id),
        "step_key_hash": hash_name(event.step_key),
    }
                
    if event.is_step_start:
        action = STEP_START_EVENT
```

rexledesma (approved):
some nits

python_modules/dagster/dagster/_core/execution/api.py
```diff
@@ -1096,6 +1096,9 @@ def pipeline_execution_iterator(
             elif event.is_resource_init_failure and event.step_key:
                 failed_steps.append(event.step_key)
 
+            # Telemetry
+            log_step_event(event, pipeline_context)
```
dpeng817, line 1100: `pipeline_execution_iterator` can execute in another process / potentially another pod, so I think this event is going to get lost when not using a local executor.

python_modules/dagster/dagster/_core/telemetry.py
```diff
@@ -523,6 +529,32 @@ def log_action(
         )
 
 
+def log_step_event(event: DagsterEvent, pipeline_context: PlanOrchestrationContext):
```
dpeng817, line 532: nit: make this `log_dagster_event` or something, potentially we might want to collect information from non-step events in the future (asset stuff for example) and this is a good centralized entrypoint

PR #11626, opened 2023-01-11
 Remove experimental flag on executor and loggers parameters of Definitions and add more detail to docblock

schrockn:
### Summary & Motivation

Remove the experimental flag on executor and loggers as we haven't come up with any better ideas here (like having special keys in the `resources` dictionary).

Also used this opportunity to write a more detailed docblock on the parameters to Definitions.

Read.

Screen Shot 2023-01-11 at 7 21 32 AM


python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -114,7 +113,48 @@ def created_repo():
...
+            :py:class:`IOManager`, it gets coerced to an :py:class:`IOManagerDefinition`.
+            Any other object is coerced to a :py:class:`ResourceDefinition`.
+            These resources will be automatically bound
+            to any assets passed to this Definitions instance using
+            :py:func:`with_resources <with_resources>`. Assets passed to Definitions with
+            resources already bound using :py:func:`with_resources <with_resources>` will
+            override this dictionary.
+
+        executor (Optional[ExecutorDefinition]):
+            Default executor for jobs. Individual jobs
```
sryza, line 150: I think it's worth mentioning that this will be used even outside the context of jobs - i.e. whenever an asset is materialized. While we technically represent that a jobs under the covers, that's not how it's exposed to users.

sryza (approved):

python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -114,7 +113,48 @@ def created_repo():
 
 class Definitions:
     """
-    A set of definitions to be explicitly available and loadable by Dagster tools.
+    A set of definitions explicitly available and loadable by Dagster tools.
+
+    Parameters:
+        assets (Optional[Iterable[Union[AssetsDefinition, SourceAsset, CacheableAssetsDefinition]]]):
+            A list of assets for this repository. Assets can be created by annotating
```
erinkcochran87, line 120: I feel like we should stay away from `repository` unless we're specifically referring to something like `a GitHub repository`. I've seen some folks in the Dagster Slack define `@repository` and then use `Definitions` alongside it, leading to unexpected results - I think this language could be confusing.

sryza: Oh yeah, +1

schrockn: Thanks great catch

PR #11640, opened 2023-01-11
[1/3 partition status cache] Time window subset perf improvements

clairelin135:
In order to avoid calling `croniter` on each partition key, this PR makes two changes:
- Fetches time windows for partition keys by using a shared time window iterator across continuous ranges
- Adds a helper method that finds time windows not part of the subset by iterating across contained time windows

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
@@ -191,6 +191,29 @@ class TimeWindowPartitionsDefinition(
         )
         return next(iter(self._iterate_time_windows(partition_key_dt)))
 
+    def time_windows_for_partition_keys(
+        self,
+        partition_keys: Sequence[str],
+    ) -> List[TimeWindow]:
+        if len(partition_keys) == 0:
+            return []
+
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -1034,29 +1057,59 @@ def __init__(
...
 
+    def _get_partition_time_windows_not_in_subset(
+        self,
+        current_time: Optional[datetime] = None,
+    ) -> Sequence[TimeWindow]:
+        """
+        Returns a list of partition time windows that are not in the subset.
+        Each time window is a single partition.
+        """
+        partitions_def = cast(TimeWindowPartitionsDefinition, self._partitions_def)
```
sryza, line 1068: I'm surprised this cast is needed

sryza (approved):
Nice!

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
clairelin135: you're right, looks like its not 👍

PR #11641, opened 2023-01-11
[2/4 partition status cache] Storing range length in Time Window Subset

clairelin135:
The key Dagit performance optimization we'd like to make is to be able to directly render the time window partition bar without recalculating the partitions in each range.

This PR enables this by:
- Adjusting `TimeWindowPartitionsSubset` to store a length for each range
- Creating a `get_inverse_subset` method that performantly builds a `TimeWindowPartitionsSubset` that contains all partitions not contained in the original subset. This enables fetching the unmaterialized partitions from a serialized representation of materialized partitions.
- Because the serialization representation has changed, this PR adds a serialization version to the serialization string. It also enables backcompat deserialization, allowing for deserializing serialized subsets created prior to this change.


clairelin135:
Closing since

PR #11645, opened 2023-01-11
[2/n][structured config] Enable struct config resources, IO managers to depend on other resources

benpankow:
## Summary

Allows resources to depend on other resources:

```python
class AWSCredentials(Resource):
    username: str
    password: str

class S3Resource(Resource):
    credentials: AWSCredentials
    region: str
    bucket: str

class EC2Resource(Resource):
    credentials: AWSCredentials

credentials = AWSCredentials(
    username=EnvVar("AWS_USERNAME"),
    password=EnvVar("AWS_PASSWORD"),
)

s3 = S3Resource(credentials=credentials)
ec2 = EC2Resource(credentials=credentials)

defs = Definitions(
    assets=[...],
    resources={
        "aws_credentials": credentials,
        "s3": s3,
        "ec2": snowflake,
    }
)
```

When using raw-Python outputs, these must be categorized as resources using the `ResourceDependency` annotation:

```python
@resource(config_schema={"username": str, "password": str})
def postgres_engine(context):
    return sqlalchemy.create_engine(...)

class DBResource:
    engine: ResourceOutput[sqlalchemy.engine.Engine]

engine = postgres_engine.configured({"username": "foo", "password": "bar"})

# engine input is typed to take any of
# sqlalchemy.engine.Engine, Resource[sqlalchemy.engine.Engine],
# PartialResource[sqlalchemy.engine.Engine],
# or ResourceDefintion (for old, function-based resources)
db = DBResource(engine=engine)
```

## Implementation

The goal of this PR is to allow for the following pattern:

```python
foo = FooResource.configure_at_runtime()
bar = BarResource(inner=foo)

defs = Definitions(
    ...
    resources={
        "my_foo": foo,
        "my_bar": bar,
    }
)
```

Here, `bar` has a pointer to a `ResourceDefinition` for `foo`, but has no knowledge that `foo` is tied to the resource key `"my_foo"` (this is only specified at `Definitions` time). However, any pipeline which depends on `"my_bar"` as a required resource key must also depend on `"my_foo"`, so `bar.required_resource_keys` must return `{"my_foo"}`.

Getting this information (that `foo` has `"my_foo"` as a resource key) to `required_resource_keys` is tricky. 

This PR adds a new wrapper `ResourceDefinition` called `ResourceWithKeyMapping` which pairs the wrapped resource with mapping that takes `ResourceDefinition` IDs to resource keys (e.g. allowing us to map `foo`->`"my_foo"`). It catches any calls to `required_resource_keys` and instead forwards them to `_resolve_required_resource_keys`, which takes the mapping.

See #12232 for more context.

## Test Plan

Added unit tests for I/O managers, resources.
## Test Plan

### DIFF
--- a/python_modules/dagster/dagster/_config/structured_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/structured_config/__init__.py
@@ -178,8 +178,8 @@ def _curry_config_schema(schema_field: Field, data: Any) -> IDefinitionConfigSch
     return DefinitionConfigSchema(_apply_defaults_to_schema_field(schema_field, data))
 
 
-ResValue = TypeVar("ResValue")
-IOManagerValue = TypeVar("IOManagerValue", bound=IOManager)
+TResValue = TypeVar("TResValue")
+TIOManagerValue = TypeVar("TIOManagerValue", bound=IOManager)
 
 ResourceId: TypeAlias = int
 
@@ -343,7 +343,7 @@ def attach_resource_id_to_key_mapping(
 
 
 class Resource(
-    Generic[ResValue],
+    Generic[TResValue],
     ResourceDefinition,
     Config,
     TypecheckAllowPartialResourceInitParams,
@@ -402,7 +402,7 @@ class Resource(
         """
         return PartialResource(cls, data=kwargs)
 
-    def initialize_and_run(self, context: InitResourceContext) -> ResValue:
+    def initialize_and_run(self, context: InitResourceContext) -> TResValue:
         # If we have any partially configured resources, we need to update them
         # with the fully configured resources from the context
 
@@ -440,12 +440,12 @@ class Resource(
 
         return self._create_object_fn(context)
 
-    def _create_object_fn(self, context: InitResourceContext) -> ResValue:
+    def _create_object_fn(self, context: InitResourceContext) -> TResValue:
         return self.create_object_to_pass_to_user_code(context)
 
     def create_object_to_pass_to_user_code(
         self, context: InitResourceContext
-    ) -> ResValue:  # pylint: disable=unused-argument
+    ) -> TResValue:  # pylint: disable=unused-argument
         """
         Returns the object that this resource hands to user code, accessible by ops or assets
         through the context or resource parameters. This works like the function decorated
@@ -454,7 +454,7 @@ class Resource(
         Default behavior for new class-based resources is to return itself, passing
         the actual resource object to user code.
         """
-        return cast(ResValue, self)
+        return cast(TResValue, self)
 
 
 def _is_fully_configured(resource: ResourceDefinition) -> bool:
@@ -473,12 +473,12 @@ def _is_fully_configured(resource: ResourceDefinition) -> bool:
 
 
 class PartialResource(
-    Generic[ResValue], ResourceDefinition, AllowDelayedDependencies, MakeConfigCacheable
+    Generic[TResValue], ResourceDefinition, AllowDelayedDependencies, MakeConfigCacheable
 ):
     data: Dict[str, Any]
-    resource_cls: Type[Resource[ResValue]]
+    resource_cls: Type[Resource[TResValue]]
 
-    def __init__(self, resource_cls: Type[Resource[ResValue]], data: Dict[str, Any]):
+    def __init__(self, resource_cls: Type[Resource[TResValue]], data: Dict[str, Any]):
         resource_pointers, data_without_resources = _separate_resource_params(data)
 
         MakeConfigCacheable.__init__(self, data=data, resource_cls=resource_cls)  # type: ignore  # extends BaseModel, takes kwargs
@@ -505,9 +505,9 @@ class PartialResource(
         )
 
 
-ResourceOrPartial: TypeAlias = Union[Resource[ResValue], PartialResource[ResValue]]
-ResourceOrPartialOrBase: TypeAlias = Union[
-    Resource[ResValue], PartialResource[ResValue], ResourceDefinition, ResValue
+ResourceOrPartial: TypeAlias = Union[Resource[TResValue], PartialResource[TResValue]]
+ResourceOrPartialOrValue: TypeAlias = Union[
+    Resource[TResValue], PartialResource[TResValue], ResourceDefinition, TResValue
 ]
 
 
@@ -521,7 +521,7 @@ class ResourceDependency(Generic[V]):
     def __get__(self, obj: "Resource", __owner: Any) -> V:
         return getattr(obj, self._name)
 
-    def __set__(self, obj: Optional[object], value: ResourceOrPartialOrBase[V]) -> None:
+    def __set__(self, obj: Optional[object], value: ResourceOrPartialOrValue[V]) -> None:
         setattr(obj, self._name, value)
 
 
@@ -566,7 +566,7 @@ class StructuredResourceAdapter(Resource, ABC):
         return self.wrapped_resource(*args, **kwargs)
 
 
-class StructuredConfigIOManagerBase(Resource[IOManagerValue], IOManagerDefinition):
+class StructuredConfigIOManagerBase(Resource[TIOManagerValue], IOManagerDefinition):
     """
     Base class for Dagster IO managers that utilize structured config. This base class
     is useful for cases in which the returned IO manager is not the same as the class itself
@@ -586,11 +586,11 @@ class StructuredConfigIOManagerBase(Resource[IOManagerValue], IOManagerDefinitio
             description=self.__doc__,
         )
 
-    def _create_object_fn(self, context: InitResourceContext) -> IOManagerValue:
+    def _create_object_fn(self, context: InitResourceContext) -> TIOManagerValue:
         return self.create_io_manager_to_pass_to_user_code(context)
 
     @abstractmethod
-    def create_io_manager_to_pass_to_user_code(self, context) -> IOManagerValue:
+    def create_io_manager_to_pass_to_user_code(self, context) -> TIOManagerValue:
         """Implement as one would implement a @io_manager decorator function"""
         raise NotImplementedError()
 
@@ -603,8 +603,8 @@ class StructuredConfigIOManagerBase(Resource[IOManagerValue], IOManagerDefinitio
         return PartialIOManager(cls, data=kwargs)
 
 
-class PartialIOManager(Generic[ResValue], PartialResource[ResValue], IOManagerDefinition):
-    def __init__(self, resource_cls: Type[Resource[ResValue]], data: Dict[str, Any]):
+class PartialIOManager(Generic[TResValue], PartialResource[TResValue], IOManagerDefinition):
+    def __init__(self, resource_cls: Type[Resource[TResValue]], data: Dict[str, Any]):
         PartialResource.__init__(self, resource_cls, data)
         IOManagerDefinition.__init__(
             self,
--- a/python_modules/dagster/dagster/_config/structured_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/structured_config/__init__.py
@@ -333,7 +333,7 @@ class IOManagerWithKeyMapping(ResourceWithKeyMapping, IOManagerDefinition):
 def attach_resource_id_to_key_mapping(
     resource_def: Any, resource_id_to_key_mapping: Dict[ResourceId, str]
 ) -> Any:
-    if isinstance(resource_def, ResourceDefinition):
+    if isinstance(resource_def, (Resource, PartialResource)):
         return (
             IOManagerWithKeyMapping(resource_def, resource_id_to_key_mapping)
             if isinstance(resource_def, IOManagerDefinition)
--- a/python_modules/dagster/dagster/_config/structured_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/structured_config/__init__.py
@@ -61,8 +61,7 @@ from dagster._core.definitions.resource_definition import (
 )
 from dagster._core.storage.io_manager import IOManager, IOManagerDefinition
 
-from . import typing_utils
-from .typing_utils import BaseResourceMeta
+from .typing_utils import BaseResourceMeta, LateBoundTypesForResourceTypeChecking
 from .utils import safe_is_subclass
 
 Self = TypeVar("Self", bound="Resource")
@@ -823,6 +822,9 @@ def _call_resource_fn_with_default(obj: ResourceDefinition, context: InitResourc
         return cast(ResourceFunctionWithoutContext, obj.resource_fn)()
 
 
-typing_utils._Resource = Resource
-typing_utils._PartialResource = PartialResource
-typing_utils._ResourceDep = ResourceDependency
+
+LateBoundTypesForResourceTypeChecking.set_actual_types_for_type_checking(
+    resource_dep_type=ResourceDependency,
+    resource_type=Resource,
+    partial_resource_type=PartialResource,
+)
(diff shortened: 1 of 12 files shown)


benpankow:
* master
  * **PR #11645** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11645" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11645?utm_source=stack-comment).

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -27,7 +31,26 @@ class cached_property:  # type: ignore[no-redef]
     convert_potential_field,
 )
 from dagster._core.definitions.resource_definition import ResourceDefinition, ResourceFunction
-from dagster._core.storage.io_manager import IOManager, IOManagerDefinition
+from dagster._core.storage.io_manager import (
+    IOManager,
+    IOManagerDefinition,
+)
+
+Self = Any
```
benpankow, line 39: keep mypy happy for now

schrockn (dismissed):
I'm definitely scared of this one. It is fiendishly clever, but I'm concerned about maintainability over time.

What alternatives did we consider? Is there an approach available to use where we use a type annotation for resource-to-resource dependencies.
```
class MyResource:
   inner_resource: DependsOnResource[resource_key, InnerResource]
```   
 
Stepping back I think this is type of situation where we should explore imposing more burden on users here in order to simplify the implementation and make the behavior more predictable.

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -114,11 +114,92 @@ def _curry_config_schema(schema_field: Field, data: Any) -> IDefinitionConfigSch
...
+            if not field.startswith("__"):
+                if get_origin(annotations[field]) == _ResourceDep:
+                    arg = get_args(annotations[field])[0]
+                    annotations[field] = Union[_PartialResource[arg], _Resource[arg]]
+                elif _safe_is_subclass(annotations[field], _Resource):
+                    base = annotations[field]
+                    annotations[field] = Union[_PartialResource[base], base]
+
+        namespaces["__annotations__"] = annotations
+        return super().__new__(self, name, bases, namespaces, **kwargs)
```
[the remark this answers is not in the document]
benpankow: When a user defines a class extending `Resource`, Pydantic will use the type annotations to build a schema of valid values for each field.

For example, if the user defines the resources:
```python
class MyCredentials(Resource):
    username: str
    password: str

class MyResource(Resource):
    creds: MyCredentials
```

Then the schema requires `creds` to be an instance `MyCredentials` when `MyResource` is constructed. However, we would like to enable a user to specify a runtime-configured resource, which is of the type `Partial[MyCredentials]`.

To do so, we want to redefine the user's type, behind the scenes, to:
```python
class MyResource(Resource):
    creds: MyCredentials | PartialResource[MyCredentials]
```

This is what this metaclass does - when the user defines a class extending `Resource`, it iterates over all the fields and remaps any `Resource`-annotated field with a union of `Resource` and `PartialResource`.


(I've added some more comments to describe this inline)

python_modules/dagster/dagster/_config/structured_config.py
benpankow: actually can be done using `TypeVar`, never mind

benpankow:
> I'm definitely scared of this one. It is fiendishly clever, but I'm concerned about maintainability over time.
> 
> What alternatives did we consider? Is there an approach available to use where we use a type annotation for resource-to-resource dependencies.
> 
> ```
> class MyResource:
>    inner_resource: DependsOnResource[resource_key, InnerResource]
> ```
> 
> Stepping back I think this is type of situation where we should explore imposing more burden on users here in order to simplify the implementation and make the behavior more predictable.

Ultimately I think most approaches will need to rely on similar hackery to what's encompassed in this PR + its parent.

The two somewhat scary pieces of metaprogramming are:
- The Resource metaclass
- The Resource descriptor (in the parent PR [here](https://github.com/dagster-io/dagster/pull/11773#discussion_r1088469963))

The Resource metaclass' purpose is to allow a Pydantic attribute to have a partial value which doe not match its type annotation. In effect, it quiets Pydantic's runtime checks for these particular fields. In any solution where resource-resource dependencies are defined as attributes, this sort of remapping is needed because the `Resource` class is constructed at load-time but the dependency is not resolved until runtime. [More explanation in my comment above](https://github.com/dagster-io/dagster/pull/11645#discussion_r1083133811). An `InnerResource` annotation is mapped to `Union[InnerResource, PartialResource[InnerResource]]`.

One option would be to make this remapping explicit, e.g.
```python
class MyResource:
   inner_resource: DependsOnResource[InnerResource]
```
Where `DependsOnResource` is a type alias for `Union[InnerResource, PartialResource[InnerResource]]`. Then we could drop the metaclass. 

I don't have a strong preference here - the metaclass complexity shouldn't bleed through to the user, and shouldn't grow in complexity over time, but if we prefer to avoid metaprogramming we could use a custom annotation.

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -113,12 +123,90 @@ def _curry_config_schema(schema_field: Field, data: Any) -> IDefinitionConfigSch
...
+
+
+class AllowDelayedDependencies:
+    _resource_pointers: Mapping[str, ResourceDefinition] = {}
+
+    def _resolve_required_resource_keys(self) -> AbstractSet[str]:
+        # All dependent resources which are not fully configured
+        # must be specified to the Definitions object so that the
+        # resource can be configured at runtime by the user
+        pointer_keys = {k: v.top_level_key for k, v in self._resource_pointers.items()}
```
schrockn, line 182: There's gotta be a way of doing this without the stateful ResourceDefinition object....

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -113,12 +123,90 @@ def _curry_config_schema(schema_field: Field, data: Any) -> IDefinitionConfigSch
...
+                    annotations[field] = Any
+                elif _safe_is_subclass(annotations[field], _Resource):
+                    # If the annotation is a Resource, we treat it as a Union of a PartialResource
+                    # and a Resource for Pydantic's sake, so that a user can pass in a partially
+                    # configured resource.
+                    base = annotations[field]
+                    annotations[field] = Union[_PartialResource[base], base]
+
+        namespaces["__annotations__"] = annotations
+        return super().__new__(self, name, bases, namespaces, **kwargs)
```
schrockn, line 172: We should have a separate file for all the code that effects only the typehinting this the Description mixin and this metaclass. The entire file will do nothing other than the annotation mangling. Again I'm also envisioning a future where rip out this stuff and want to be able to do it cleanly.

schrockn (changes requested):
I think we need another session to talk through this. The stateful `top_level_key` property in ResourceDefinition is quite the non-starter for me.

python_modules/dagster/dagster/_config/structured_config.py
benpankow: Moved metaclass to a standalone file

python_modules/dagster/dagster/_core/definitions/resource_definition.py
```diff
@@ -103,9 +103,25 @@ def __init__(
...
+        self._top_level_key = key
+
+    @property
+    def top_level_key(self) -> Optional[str]:
+        """
+        Gets the top-level resource key for this resource which was associated with it
+        in the Definitions object.
+        """
+        return self._top_level_key
+
```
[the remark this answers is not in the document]
benpankow: The `top_level_key` is used in the following scenario:

```python
foo = FooResource.configure_at_runtime()
bar = BarResource(inner=foo)

defs = Definitions(
    ...
    resources={
        "foo": foo,
        "bar": bar,
    }
)
```

Here, we need to affiliate the `inner` input to `bar` with the `foo`  top-level resource key, so we know how to populate the fully-configured resource at runtime.

The current approach is to:
- At Definitions time, associate all top-level resources’ `top_level_key` with the key they are given in the Definitions object
- If resource `A` depends on resource `B`, when we query `A`’s `required_resource_keys`, we look up `B’s` `top_level_key` and return it

An alternative which would remove statefulness is to
- At resource init time, generate a UUID (or perhaps a deterministic ID which remains the same if the repo does not change), `unique_id` for each resource.
- If resource `A` depends on resource `B`, when we query `A`’s `required_resource_keys`, we look up `B`’s `unique_id`. Then, we look at each of the top-level resources in the repo and find the resource key associated with that `unique_id`.

Here the unique_id is set when the resource is constructed and its only purpose is to uniquely identify a resource when figuring out the associated top-level key.

schrockn (changes requested):
We should go forward with your mapping solution. Let's implement that here. We shouldn't commit `top_level_resource` as an intermediate state since it makes the evolution of this confusing.

benpankow:
Merged #12232 into this PR, avoiding the `top_level_keys` approach.

python_modules/dagster/dagster/_config/structured_config/__init__.py
```diff
@@ -168,13 +180,154 @@ def _curry_config_schema(schema_field: Field, data: Any) -> IDefinitionConfigSch
...
+        # must be specified to the Definitions object so that the
+        # resource can be configured at runtime by the user
+        nested_partial_resource_keys = {
+            attr_name: resource_mapping.get(id(resource_def))
+            for attr_name, resource_def in self._nested_partial_resources.items()
+        }
+        check.invariant(
+            all(pointer_key is not None for pointer_key in nested_partial_resource_keys.values()),
+            (
+                "Any partially configured, nested resources must be specified to Definitions"
```
schrockn, line 217: specified a bit of an odd wording here

python_modules/dagster/dagster/_config/structured_config/__init__.py
```diff
@@ -168,13 +180,154 @@ def _curry_config_schema(schema_field: Field, data: Any) -> IDefinitionConfigSch
...
+    return resource_def
+
+
+def _get_resource_by_id(context: InitResourceContext, resource_id: ResourceId) -> Any:
+    """
+    Retrieves a resource's value from the resource context object from the ResourceDefinition's ID.
+    This is used to retrieve the value of nested resources which are not fully configured, since the
+    resource key is not known by the nested resource's parent.
+    """
+    return getattr(context.resources, f"id_{resource_id}", None)
```
schrockn, line 321: we should model this with a data structure somewhere, not mangle strings and monkeypatch

schrockn (changes requested):
This looks mostly good, except for the string mangling and jamming dynamically named properties on the resources object. Let's do something better than that.

python_modules/dagster/dagster/_config/structured_config/__init__.py
benpankow: Attached to a new `InitResourceContext` subclass, `InitResourceContextWithKeyMapping`, which includes the id-to-resource mapping.

python_modules/dagster/dagster/_config/structured_config/__init__.py
```diff
@@ -494,21 +767,62 @@ def infer_schema_from_config_annotation(model_cls: Any, config_arg_default: Any)
...
 
     shape_cls = Permissive if model_cls.__config__.extra == Extra.allow else Shape
 
     docstring = model_cls.__doc__.strip() if model_cls.__doc__ else None
+    return Field(config=shape_cls(fields), description=description or docstring)
+
+
+def _separate_resource_params(
+    data: Dict[str, Any]
+) -> Tuple[Dict[str, Union[Resource, PartialResource, ResourceDefinition]], Dict[str, Any]]:
```
schrockn, line 795: nit: NamedTuple with property names could make this data structure more self-documenting

python_modules/dagster/dagster/_config/structured_config/__init__.py
```diff
@@ -494,21 +767,62 @@ def infer_schema_from_config_annotation(model_cls: Any, config_arg_default: Any)
 def infer_schema_from_config_class(
     model_cls: Type[Config],
     description: Optional[str] = None,
+    ignore_resource_fields: Optional[Set[str]] = None,
```
schrockn, line 770: nit: `fields_to_omit` probably better name? `ignore_resource_fields` sounds like a bool

python_modules/dagster/dagster/_config/structured_config/__init__.py
```diff
@@ -494,21 +767,62 @@ def infer_schema_from_config_annotation(model_cls: Any, config_arg_default: Any)
...
+    if has_at_least_one_parameter(obj.resource_fn):
+        return cast(ResourceFunctionWithContext, obj.resource_fn)(context)
+    else:
+        return cast(ResourceFunctionWithoutContext, obj.resource_fn)()
+
 
-    return Field(shape_cls(fields), description=description or docstring)
+typing_utils._Resource = Resource
+typing_utils._PartialResource = PartialResource
+typing_utils._ResourceDep = ResourceDependency
```
schrockn, line 828: This is quite strange to set a type in _another_ module.

python_modules/dagster/dagster_tests/core_tests/resource_tests/test_struct_config_resources.py
```diff
@@ -430,3 +437,583 @@ def hello_world_asset(writer: WriterResource):
...
+            f.write(
+                """
+from dagster._config.structured_config import Resource, ResourceDependency
+
+class StringDependentResource(Resource):
+    a_string: ResourceDependency[str]
+
+reveal_type(StringDependentResource.__init__)
+
+my_str_resource = StringDependentResource(a_string="foo")
```
schrockn, line 1001: These tests are supppper slow. I think we should disable them by default locally and additionally, run them in their own test suite to keep the `core_test` time reasonable.

python_modules/dagster/dagster/_config/structured_config/typing_utils.py
```diff
@@ -1,8 +1,79 @@
-from typing import TYPE_CHECKING, Any, Optional, TypeVar, Union, cast
+from typing import TYPE_CHECKING, Any, Generic, Optional, Type, TypeVar, Union, cast
+
+import pydantic
+from typing_extensions import dataclass_transform, get_origin
+
+from .utils import safe_is_subclass
 
 if TYPE_CHECKING:
```
schrockn, line 8: Worthy of comment/explanation as to why this works while still having to do all the crazy shenanigans around temp classes.

schrockn (changes requested):
One last turn.

We definitely don't want to slow down the core tests case suite this much with those mypy and pyright tests.

schrockn:
Also I think you should do something like https://github.com/dagster-io/dagster/pull/12273 to hide all the temp class stuff in a single scope. Feels a little java-y, but those top-level symbols aren't great.

python_modules/dagster/dagster_tests/core_tests/resource_tests/test_struct_config_resources.py
benpankow: Gated behind `-m 'typesignature'` mark. They'll run separately in CI now & only run locally with that flag.

schrockn (approved):
ok lfg

python_modules/dagster/dagster/_config/structured_config/__init__.py
```diff
@@ -250,6 +506,23 @@ def resource_fn(context: InitResourceContext):
 
 
 ResourceOrPartial: TypeAlias = Union[Resource[ResValue], PartialResource[ResValue]]
+ResourceOrPartialOrBase: TypeAlias = Union[
```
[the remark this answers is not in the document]
benpankow: Changed to `OrValue`, since this can also hold the raw resource value (in the case of a bare Python object)

Issue #11622, opened 2023-01-11
An error occurred (ClientException) when calling the RegisterTaskDefinition operation


whillas-yabble:
### Dagster version

1.1.8 (core) / 0.17.8 (libraries)

### What's the issue?

Using ECS and seperate containers into 3 parts. Getting this error when launching a new task
```
botocore.errorfactory.ClientException: An error occurred (ClientException) when calling the RegisterTaskDefinition operation: The secret name must be unique and not shared with any new or existing environment variables set on the container, such as 'DAGSTER_POSTGRES_USER'.

  File "/usr/local/lib/python3.9/site-packages/dagster/_core/instance/__init__.py", line 1839, in launch_run
    self.run_launcher.launch_run(LaunchRunContext(pipeline_run=run, workspace=workspace))
  File "/usr/local/lib/python3.9/site-packages/dagster_aws/ecs/launcher.py", line 311, in launch_run
    run_task_kwargs = self._run_task_kwargs(run, image, container_context)
  File "/usr/local/lib/python3.9/site-packages/dagster_aws/ecs/launcher.py", line 497, in _run_task_kwargs
    self.ecs.register_task_definition(**task_definition_dict)
  File "/usr/local/lib/python3.9/site-packages/botocore/client.py", line 514, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/local/lib/python3.9/site-packages/botocore/client.py", line 934, in _make_api_call
    raise error_class(parsed_response, operation_name)
```
The DAGSTER_POSTGRES_USER is a secret tagged with `dagster` so it gets added to the container automatically on launch. The parent, user code, all 3 containers (user code, daemon & dagit) have this secret so they can access the DB.

This was working previously when we just set the DB credentials as env. vars (but this was obviously a security concern).

So ti looks like the parent, user code, container env vars are copied across (including the ones created from secrets) to the new task container and then the task container is trying to copy the secrets again over the top causing this error?

### What did you expect to happen?

No error

### How to reproduce?

Tag some secrets on a parent container and then launch a ECS task from this same container?

### Deployment type

Other Docker-based deployment

### Deployment details

ECS with the following config
```json
{
    "taskDefinitionArn": "arn:aws:ecs:ap-southeast-2:1234567890:task-definition/devau-somecompany-dagster-pipelines-task:5",
    "containerDefinitions": [
        {
            "name": "devau-somecompany-dagster-pipelines-container",
[...]

PR #11646, opened 2023-01-11
Prefer ECS secrets to environment variables of the same name when both are set

jmsanders:
https://github.com/dagster-io/dagster/issues/11622 describes a situation where we try to register a task definition with a secret and an environment variable with identical names.

This can happen because of something you do explicitly: you set both a secret and an env var in your container context.

It can also happen for more subtle and unexpected reasons: you set an environment variable but don't realize that you have a "tagged" secret with the same name, you set a secret but don't realize that an environment variable has been inherited from a "parent" task definition, etc.

I'm not entirely sure what the right fix is here. For now, I've opted to always prefer the secret to the env var if the two overlap. Of course, that raises the question of "what if I want to prefer the env var." Perhaps the best answer there for now is "rename one of the two."

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws/ecs/tasks.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/ecs/tasks.py
@@ -212,6 +212,11 @@ def get_task_definition_dict_from_current_task(
             *environment,
         ]
 
+    secret_names = set(secret.get("name") for secret in new_container_definition["secrets"])
+    for env in new_container_definition["environment"]:
+        if env.get("name") in secret_names:
+            new_container_definition["environment"].remove(env)
+
     if include_sidecars:
         container_definitions = current_task_definition_dict.get("containerDefinitions")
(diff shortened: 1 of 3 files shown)


gibsondan (commented):
now that we know the cause of the report i'm unsure about this one - it sounds like in the attached task that the error was actually identifying an invalid state (env vars specified in the dagster.yaml and also tagged dagster secrets) - it may be better to fail here with a reasonably clear message vs. overwrite the same environment variable?

jmsanders:
We can probably just rely on the AWS message for now - let ECS do the heavy lifting of validating the request.

I'm going to scale this back to just represent changes to our ECS stub since it brings it ever so slightly in line with actual ECS behavior.

gibsondan (changes requested):
finally catching back up on my queue - you were going to pull this back to just the stub yeah?

jmsanders:
Yep - just low priority.

Issue #11585, opened 2023-01-09
Adding tags to backfill runs from Dagit UI


Beaglefoot:
### What's the use case?

Adding tags allows to manage concurrency, restrict resources and lower priority of backfill runs.

### Ideas of implementation

It's already available with cli:
```
dagster job backfill --tags awesome_tag
```
So it's probably possible to copy/paste [Edit Tags UI from Launchpad](https://docs.dagster.io/concepts/ops-jobs-graphs/metadata-tags#specifying-tags) and somehow reuse API from the code snippet above.

### Additional information

I have a vague idea that this was available at some point in time previously.

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

sryza:
@bengotow mind taking a look at this one?

sryza:
Just came across another request for this one. @bengotow - would we be able to get this done in time for the next release?

bengotow:
Hey gang, yep there's a pull request up for this actually just forgot to tag it: https://github.com/dagster-io/dagster/pull/11656

PR #11656, opened 2023-01-12
[dagit] Add backfill warnings and run tags options to the Materialize modal

bengotow:
### Summary & Motivation

Related: https://github.com/dagster-io/dagster/issues/11585
A user recently flagged that our "Materialize" button for assets doesn't offer the ability to add tags to your run or backfill. I think this was just an oversight - these concepts are supported by the underlying APIs.

This PR moves more components of the job backfill modal into a shared set and re-uses them in the asset modal. In addition to the tag UI, I also added the two warnings that can appear in the job backfill modal: "You are using the default run launcher"  and "Your backfill daemon is not running". To make these cleaner, I also put the display conditionals inside them and gave them proper GraphQL fragments.

Here's what both modals look like now (in a worst-case scenario where all the warning states are present). In the asset modal, if you select only one partition the warnings disappear, since they're not relevant in the single job case, but you can add tags either way.

image

image

I verified that tags are applied for both single jobs and backfills of assets, and verified that the job backfill modal is not impacted. I also verified that turning on my dagster backfill daemon makes the warning disappear in both dialogs. Also verified that if you make the screen really small, the asset backfill modal scrolls, since it's getting a bit large.

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
+++ b/js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
@@ -1,21 +1,23 @@
-import {useApolloClient} from '@apollo/client';
+import {gql, useApolloClient, useQuery} from '@apollo/client';
 import {
-  Dialog,
-  DialogHeader,
-  DialogBody,
+  Alert,
   Box,
   Button,
   ButtonLink,
-  DialogFooter,
-  Tooltip,
   Colors,
-  Alert,
+  Dialog,
+  DialogBody,
+  DialogBodySection,
+  DialogFooter,
+  DialogHeader,
+  Tooltip,
 } from '@dagster-io/ui';
 import reject from 'lodash/reject';
 import React from 'react';
 import {useHistory} from 'react-router-dom';
 
 import {showCustomAlert} from '../app/CustomAlertProvider';
+import {PipelineRunTag} from '../app/ExecutionSessionStorage';
 import {usePermissionsForLocation} from '../app/Permissions';
 import {PythonErrorInfo} from '../app/PythonErrorInfo';
 import {displayNameForAssetKey, isHiddenAssetGroupJob} from '../asset-graph/Utils';
@@ -29,11 +31,17 @@ import {
 } from '../instance/types/BackfillUtils.types';
 import {CONFIG_PARTITION_SELECTION_QUERY} from '../launchpad/ConfigEditorConfigPicker';
 import {useLaunchPadHooks} from '../launchpad/LaunchpadHooksContext';
+import {TagContainer, TagEditor} from '../launchpad/TagEditor';
 import {
   ConfigPartitionSelectionQuery,
   ConfigPartitionSelectionQueryVariables,
 } from '../launchpad/types/ConfigEditorConfigPicker.types';
-import {showBackfillErrorToast, showBackfillSuccessToast} from '../partitions/BackfillSelector';
+import {
+  DaemonNotRunningAlert,
+  showBackfillErrorToast,
+  showBackfillSuccessToast,
+  UsingDefaultLauncherAlert,
+} from '../partitions/BackfillMessaging';
 import {DimensionRangeWizard} from '../partitions/DimensionRangeWizard';
 import {PartitionStateCheckboxes} from '../partitions/PartitionStateCheckboxes';
 import {PartitionState} from '../partitions/PartitionStatus';
@@ -47,6 +55,7 @@ import {PartitionDefinitionForLaunchAssetFragment} from './types/LaunchAssetExec
 import {usePartitionDimensionSelections} from './usePartitionDimensionSelections';
 import {PartitionDimensionSelection, usePartitionHealthData} from './usePartitionHealthData';
 import {usePartitionNameForPipeline} from './usePartitionNameForPipeline';
+
 interface Props {
   open: boolean;
   setOpen: (open: boolean) => void;
@@ -99,6 +108,10 @@ const LaunchAssetChoosePartitionsDialogBody: React.FC<Props> = ({
 
   const {canLaunchPartitionBackfill} = usePermissionsForLocation(repoAddress.location);
   const [launching, setLaunching] = React.useState(false);
+
+  const [tagEditorOpen, setTagEditorOpen] = React.useState<boolean>(false);
+  const [tags, setTags] = React.useState<PipelineRunTag[]>([]);
+
   const [previewCount, setPreviewCount] = React.useState(0);
   const morePreviewsCount = partitionedAssets.length - previewCount;
 
@@ -141,6 +154,9 @@ const LaunchAssetChoosePartitionsDialogBody: React.FC<Props> = ({
   const history = useHistory();
   const {useLaunchWithTelemetry} = useLaunchPadHooks();
   const launchWithTelemetry = useLaunchWithTelemetry();
+
+  const instanceResult = useQuery(LAUNCH_ASSET_CHOOSE_PARTITIONS_QUERY);
+  const instance = instanceResult.data?.instance;
 
   // Find the partition set name. This seems like a bit of a hack, unclear
   // how it would work if there were two different partition spaces in the asset job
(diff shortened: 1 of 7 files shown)


js_modules/dagit/packages/core/src/instance/InstanceBackfills.tsx
```diff
@@ -84,23 +84,7 @@ export const InstanceBackfills = () => {
...
-                          target="_blank"
-                          rel="noreferrer"
-                        >
-                          dagster-daemon documentation
-                        </a>{' '}
-                        for more information on how to deploy the dagster-daemon process.
-                      </div>
-                    }
-                  />
+                  <DaemonNotRunningAlertBody />
```
bengotow, line 87: I believe this page can / should use the `<DaemonNotRunningAlert />` and the new fragment,  but it's doing some other things and appears (maybe?) to support multiple backfill daemons? Will talk to some folks and see. For now, just re-using the body at least cleans things up a bit!

js_modules/dagit/packages/ui/src/components/Dialog.tsx
```diff
@@ -61,6 +61,26 @@ export const DialogBody: React.FC = (props) => {
   );
 };
 
+// The subheadings used in the backfill modals
```
hellendag, line 64: Does this need to be in `Dialog`? I'm not sure we have had anything quite like this elsewhere.

This also strikes me as not very dialog-specific -- it's a header and some children, with some spacing -- so I'm not sure this is really the right place for it.

js_modules/dagit/packages/ui/src/components/Dialog.tsx
```diff
@@ -61,6 +61,26 @@ export const DialogBody: React.FC = (props) => {
...
+export const DialogBodySection = ({
+  title,
+  children,
+}: {
+  title: string | React.ReactNode;
+  children: React.ReactNode;
+}) => (
+  <Box flex={{direction: 'column', gap: 8}}>
+    <strong style={{display: 'block'}}>{title}</strong>
```
hellendag, line 73: Possibly a styled `h3` or `h4` or something instead of a `strong`?

js_modules/dagit/packages/ui/src/components/Dialog.tsx
```diff
@@ -61,6 +61,26 @@ export const DialogBody: React.FC = (props) => {
...
+    <Box
+      flex={{direction: 'column', gap: 8}}
```
hellendag, line 75: Are you certain you always want the children to be flex-column items?

js_modules/dagit/packages/core/src/partitions/BackfillMessaging.tsx
```diff
@@ -0,0 +1,152 @@
...
+        >
+          dagster-daemon documentation
+        </a>{' '}
+        for more information on how to deploy the dagster-daemon process.
+      </div>
+    }
+  />
+);
+
+export const USING_DEFAULT_LAUNCH_ERALERT_INSTANCE_FRAGMENT = graphql(`
```
hellendag, line 117: Sorry to say that I'm about to change this back...

js_modules/dagit/packages/core/src/partitions/BackfillMessaging.tsx
```diff
@@ -0,0 +1,152 @@
...
+          <PythonErrorInfo error={error} key={idx} />
+        ))}
+      </>
+    );
+  }
+
+  return (
+    <Group direction="column" spacing={4}>
+      <div>An unexpected error occurred. This backfill was not launched.</div>
+      {errors ? (
```
hellendag, line 40: Looks like `errors` will always be truthy, since the default is a fragment.

js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
```diff
@@ -341,7 +358,34 @@ const LaunchAssetChoosePartitionsDialogBody: React.FC<Props> = ({
           ranges={ranges}
           setRanges={setRanges}
         />
+
+        <Box flex={{direction: 'column', gap: 32}} style={{marginTop: 32}}>
```
hellendag, line 362: 32 might be too much gap, the alerts look a little too far apart.

js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
```diff
@@ -341,7 +358,34 @@ const LaunchAssetChoosePartitionsDialogBody: React.FC<Props> = ({
...
+            <TagEditor
+              tagsFromSession={tags}
+              onChange={setTags}
+              open={tagEditorOpen}
+              onRequestClose={() => setTagEditorOpen(false)}
+            />
+            {tags.length ? (
+              <div style={{border: `1px solid ${Colors.Gray300}`, borderRadius: 8, padding: 3}}>
+                <TagContainer tagsFromSession={tags} onRequestEdit={() => setTagEditorOpen(true)} />
+              </div>
```
hellendag, line 373: Can't see what this looks like in the current screenshots, can you add one with the tag container visible?

js_modules/dagit/packages/core/src/partitions/BackfillMessaging.tsx
```diff
@@ -0,0 +1,152 @@
...
+        <div>
+          Check your instance configuration in <code>dagster.yaml</code> to either configure the{' '}
+          <a
+            href="https://docs.dagster.io/deployment/run-coordinator"
+            target="_blank"
+            rel="noreferrer"
+          >
+            queued run coordinator
+          </a>{' '}
+          or to configure a run launcher more appropriate for launching a large number of jobs.
```
hellendag, line 148: I think we can remove a few extra words in the copy:

Check your instance configuration in dagster.yaml to configure the queued run coordinator or a run launcher more appropriate for launching a large number of jobs.

js_modules/dagit/packages/core/src/partitions/BackfillMessaging.tsx
```diff
@@ -0,0 +1,152 @@
...
+export const UsingDefaultLauncherAlert: React.FC<{
+  instance: UsingDefaultLauncherAlertInstanceFragmentFragment;
+}> = ({instance}) =>
+  instance.runLauncher?.name === DEFAULT_RUN_LAUNCHER_NAME && !instance.runQueuingSupported ? (
+    <Alert
+      intent="warning"
+      title={
+        <div>
+          Using the default run launcher <code>{DEFAULT_RUN_LAUNCHER_NAME}</code> for launching
+          backfills without a queued run coordinator is not advised.
```
hellendag, line 135: s/for launching/to launch

js_modules/dagit/packages/ui/src/components/Dialog.tsx
bengotow: Hmm yeah I was kinda torn about this -- it renders these subheadings, and they appear in the backfill modals but not really anywhere else.

I guess for now I'll leave it in the backfill modal and export it into the other backfill modal and we can make it a global shared element if Josh uses this design pattern in more places. 

image

js_modules/dagit/packages/ui/src/components/Dialog.tsx
bengotow: Hmmm good call. It looks like this is equivalent to `Subheading` except with a slightly different line height, will use that for consistency

js_modules/dagit/packages/ui/src/components/Dialog.tsx
bengotow: Hmm will move this back to being in the materialize modal so we can keep it specific to that UI

js_modules/dagit/packages/core/src/partitions/BackfillMessaging.tsx
bengotow: Wow good catch, will change this! I wonder how long it's been like this, was just moving it in this PR

js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
bengotow: Yeah I agree actually, just didn't want to change the job backfill modal - will make this a bit more reasonable

js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
bengotow: Ahh yep it's this one - not super glamorous but the same as the op-job backfill UI 

image

js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
bengotow: Toned this down a bit!

image

image

bengotow (commented):
Ok should have this fixed up!

hellendag (approved):

PR #11657, opened 2023-01-12
Shut down the daemon faster when it is interrupted

gibsondan:
Summary:
we shut down the daemon by setting an event that each daemon thread checks to terminate cleanly. Sometimes a particular daemon is in the middle of a long sleep when this happens, making the daemon take forever to spin down. Pass the shut down event through to the daemons so that instead of calling time.sleep, they can call shutdown_event.wait


prha (approved):
oh nice, didn't realize there was a timeout for the event.wait

PR #11663, opened 2023-01-12
[dagster-airflow] pass airflow dag timezone to dagster schedule definition

Ramshackle-Jamathon:
### Summary & Motivation

This fixes whats currently converted airflow schedules that are always in UTC instead of whichever timezone was defined on the dag

### DIFF
--- a/python_modules/libraries/dagster-airflow/dagster_airflow/dagster_pipeline_factory.py
+++ b/python_modules/libraries/dagster-airflow/dagster_airflow/dagster_pipeline_factory.py
@@ -19,8 +19,6 @@ from airflow.models.dag import DAG
 from airflow.models.dagbag import DagBag
 from airflow.settings import LOG_FORMAT
 from airflow.utils import db
-from dagster_airflow.patch_airflow_example_dag import patch_airflow_example_dag
-
 from dagster import (
     Array,
     DagsterInvariantViolationError,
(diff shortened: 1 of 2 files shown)


python_modules/libraries/dagster-airflow/dagster_airflow/dagster_pipeline_factory.py
```diff
@@ -287,7 +287,10 @@ def make_dagster_schedule_from_airflow_dag(dag, job_def):
 
     if isinstance(dag.normalized_schedule_interval, str) and is_valid_cron_schedule(cron_schedule):
         return ScheduleDefinition(
-            job=job_def, cron_schedule=cron_schedule, description=schedule_description
+            job=job_def,
+            cron_schedule=cron_schedule,
+            description=schedule_description,
+            execution_timezone=dag.timezone.name,
```
gibsondan, line 293: are we confident these use the same format? not familiar with what airflow does here

gibsondan (approved):

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_pipeline_factory.py
Ramshackle-Jamathon: yep airflow uses pendulum and the `.name` property is the IANA time zone string

PR #11669, opened 2023-01-12
Added AWS S3 Glue Destination

shaidar:
### Summary & Motivation
This PR adds AWS S3 Glue as a destination which is not currently available and would allow us to manage our Airbyte connections through Dagster and thus materializing those assets in our pipelines.


sryza:
Mind taking a look @benpankow ?

cmpadden:

PR #11671, opened 2023-01-12
Revert "test fixes for python 3.7 (#11636)" and  "Fall back to a different port when 3000 is in use instead of failing (#11610)"

gibsondan:
This reverts commit ea523f5a3b400400f42181dab3497f607dafa3d1.

Summary: This is not firing the way I would expect and seems to still think 3000 is in use for a while after uvicorn shuts down.

Is there a way to shut down uvicorn more thoroughly on interrupt so that it relinquishes the port? I looked into a way to make it bubble up the "could not bind port" error the way that flask did but was not successful.

But for now lets just revert.


alangenfeld (approved):
excited about this feature but agree the current behavior is a bit rough so lets figure something out

PR #11675, opened 2023-01-12
[dagit] Restore some previous GraphQL codegen behavior

hellendag:
### Summary & Motivation

Okay, roll it mostly back. Keep the new codegen system, but customize it better for our needs.

- Types are back where they were, in types folders. Now with .types.ts suffix. Prettified after codegen.
- Back to using gql tagged template literals. Should allow validation in vscode.
- Back to interpolated fragment inclusion.
- Back to using type annotations on hooks.

graphql.ts and gql.ts are deleted, of course. I have a temporary `graphql.tsx` in place to support some Cloud imports, but will delete the file once those are updated to point at the new paths.

Reasons were noted in Slack, but:

- `gql.ts` and `graphql.ts` are highly susceptible to conflicts, since we check in our generated code and had no intention of changing this.
- The `../graphql/graphql` and `../graphql/gql` imports are confusing and non-customizable.
- The babel plugin to support the copied GraphQL documents is an annoying requirement, especially since we have a shared webpack config in our react-scripts fork.
- Code navigation and legibility took a pretty severe hit with `graphql.ts`. Jumping to a document or type meant jumping into a mess.
- We’d never be able to use vscode GraphQL validation. It’s broken with `graphql()` functions unless we force annotation through lint or something.
- If we do decide we want to move toward separate `.gql` files, this state of the world will be a lot easier. It just wouldn’t have worked with the `graphql()` function.
- We now have much better granular control over our codegen config. The client-preset had some opinions that differed from our needs, and very limited customizability.
- The main API benefit was Apollo hook usage without type annotations, which was nice. But the type annotations worked fine, and with a little more tweaking to our custom eslint rule, this could basically be fully automatable.
- Suffix duplication (`FooBarFragmentFragment`, `FooBarQueryQuery`) was janky and non-customizable.

I put fragment interpolation back in place manually and went through it twice, so I think I got everything. If any queries/mutations are missing fragments, we should catch them very quickly.

TS, lint, Jest. Load Dagit and click around. Sanity check that things are working as expected.

TS, lint, Jest in Cloud master with this changeset.


hellendag:
* master
  * **PR #11675** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11675" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11675?utm_source=stack-comment).

js_modules/dagit/packages/core/package.json
```diff
@@ -113,6 +114,7 @@
     "@types/styled-components": "^5.1.9",
     "@types/testing-library__jest-dom": "^5.14.2",
     "@types/ws": "^6.0.3",
+    "add": "^2.0.6",
```
hellendag, line 117: This shouldn't be here.

js_modules/dagit/packages/core/src/app/PythonErrorInfo.tsx
```diff
@@ -82,30 +80,11 @@ const ErrorContext: React.FC<{errorSource: ErrorSource}> = ({errorSource}) => {
...
-  }
-
-  fragment PythonErrorChain on ErrorChainLink {
-    isExplicitLink
-    error {
-      message
-      stack
-    }
-  }
-`);
```
hellendag, line 108: This has been moved to a separate file to avoid circular dependencies.

bengotow (approved):

PR #11688, opened 2023-01-12
Bring back dagit socket reuse

gibsondan:
Summary:
With one additional line in is_port_in_use that makes our socket check more closely mimic the socket check that uvicorn does:
sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)

Test Plan: Launch dagit, open browser
Close dagit process
Re-open dagit process
Repeat
Now reliably stays on 3000

### DIFF
--- a/python_modules/dagster/dagster/_utils/__init__.py
+++ b/python_modules/dagster/dagster/_utils/__init__.py
@@ -576,6 +576,20 @@ def find_free_port() -> int:
         return s.getsockname()[1]
 
 
+def is_port_in_use(host, port) -> bool:
+    # Similar to the socket options that uvicorn uses to bind ports:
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster/_utils/__init__.py
```diff
@@ -576,6 +576,18 @@ def find_free_port() -> int:
         return s.getsockname()[1]
 
 
+def is_port_in_use(host, port) -> bool:
+    sock = socket.socket(family=socket.AF_INET)
+    sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
```
gibsondan, line 581: only real change here

alangenfeld: maybe a comment "match the socket options uvicorn uses to bind"

alangenfeld (approved):
see you later `except SystemExit as e:`

PR #11693, opened 2023-01-13
[dagit] Fix Backfills page locking up when displaying large backfills, improve "target" column

bengotow:
### Summary & Motivation

I created a backfill of 180,000 partition keys and found that Dagit is unable to render the Overview > Backfills page. It seems this is due to several things:

- The API is returning a `PartitionStatuses` object for each backfill containing an object for each partition. It'd be great to move this to the ranges format that @clairelin135 is putting together which will pass straight through to the rendering!

- Apollo cache is trying to normalize all of these 160k objects into some sort of fancy cache, which it "deep partial updates" with the new data. We can disable this behavior for just `PartitionStatus` with an entry in AppCache.tsx, but I have mixed feelings about making global changes. Maybe we do it for now and remove it when we move this status data to a smaller / faster range format. It reduces the "page lock time" after data is received from 7s to about 2s on my machine.

- The Backfill page is using the `<PartitionStatus />` component in "separated bars" mode, which actually tries to render 180k divs for the 180k partitions.

This PR addresses the last two - if there are more than 1k partition keys in the backfill, we now just render a box that is the same shape of the PartitionStatus component but contains just the counts:

image

I also observed a lot of funkiness in the Backfill Target column because I have both asset and job backfills for both loaded and unloaded repositories in my test set. I changed the code so that it omits partition set names that are part of hidden asset jobs, and made it show asset keys even when the partition set isn't part of a loaded repo. 

Page loads!

For the target backfill changes, I attached before and after screenshots and I think my setup tests all five cases (assets / non-assets + jobs / asset graph / asset partition set).

image

### DIFF
--- a/js_modules/dagit/packages/core/src/app/AppCache.tsx
+++ b/js_modules/dagit/packages/core/src/app/AppCache.tsx
@@ -7,6 +7,11 @@ export const createAppCache = () =>
   new InMemoryCache({
     addTypename: true,
     possibleTypes,
+    typePolicies: {
+      PartitionStatus: {
+        keyFields: false,
+      },
+    },
     dataIdFromObject: (object: any) => {
       if (
         object.name &&
@@ -17,6 +22,8 @@ export const createAppCache = () =>
         return 'Instance';
       } else if (object.__typename === 'Workspace') {
         return 'Workspace';
+      } else if (object.__typename === 'PartitionBackfill') {
+        return object.backfillId;
       } else {
         return defaultDataIdFromObject(object);
       }
--- a/js_modules/dagit/packages/core/src/instance/BackfillRow.tsx
+++ b/js_modules/dagit/packages/core/src/instance/BackfillRow.tsx
@@ -13,6 +13,7 @@ import {RunStatus, BulkActionStatus} from '../graphql/types';
 import {
   PartitionState,
   PartitionStatus,
+  PartitionStatusCountsOnly,
   runStatusToPartitionState,
 } from '../partitions/PartitionStatus';
 import {PipelineReference} from '../pipelines/PipelineReference';
@@ -141,7 +142,6 @@ export const BackfillRow = ({
     </tr>
   );
 };
-
 const BackfillMenu = ({
   backfill,
   statusData,
@@ -223,6 +223,7 @@ const BackfillMenu = ({
     </Popover>
   );
 };
+const BACKFILL_PARTITIONS_COUNTS_THRESHOLD = 1000;
 
 const BackfillRunStatus = ({
   backfill,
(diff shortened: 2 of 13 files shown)


js_modules/dagit/packages/core/src/app/AppCache.tsx
```diff
@@ -17,6 +22,8 @@ export const createAppCache = () =>
         return 'Instance';
       } else if (object.__typename === 'Workspace') {
         return 'Workspace';
+      } else if (object.__typename === 'PartitionBackfill') {
```
salazarm, line 25: Oh wow didn't even realize we could do this. It would be nice if we had a helper or something so that we don't need to stuff all of the types in this file which is very general. (eg. `registerDataIDForObject(typename, fn)`)

bengotow: Yeah I definitely agree ... I think we should really have the GraphQL API serve "id" as the "id" field instead of runId, backfillId, etc. I know for runs we went back and added "id" so that we didn't have to do this, and that may be what we should do in the whole system :-/

js_modules/dagit/packages/core/src/partitions/PartitionStatus.tsx
```diff
@@ -336,6 +337,33 @@ export const PartitionStatus: React.FC<{
...
+  small?: boolean;
+  partitionNames: string[];
+  partitionStateForKey: (key: string) => PartitionState;
+}> = ({small, partitionNames, partitionStateForKey}) => {
+  const byState = countBy(partitionNames, partitionStateForKey);
+
+  const counts = Object.entries(byState)
+    .filter(([, count]) => count > 0)
+    .map(([state, count]) => `${count.toLocaleString()} ${state}`)
+    .join(', ');
```
hellendag, line 350: I hope this will fit. Any common circumstances where it might not?

hellendag (approved):

js_modules/dagit/packages/core/src/instance/BackfillRow.tsx
```diff
@@ -269,48 +275,47 @@ const BackfillTarget: React.FC<{
...
-          path: `/partitions?partitionSet=${encodeURIComponent(partitionSet.name)}`,
-        })}
-        style={{fontWeight: 500}}
-      >
-        {partitionSet.name}
-      </Link>
-      <Box flex={{direction: 'column', gap: 4}} style={{fontSize: '12px'}}>
-        {repoLink}
+  const buildPipelineOrAssets = () => {
+    if (assetSelection?.length) {
```
prha, line 314: This is great... 🙏

prha (approved):

js_modules/dagit/packages/core/src/partitions/PartitionStatus.tsx
bengotow: Hmm I will double check before I merge this, that's a good question actually. I think there can be three states that we need to display.

bengotow:
Hey gang, based on discussion with @prha  earlier today, I added a backend resolver for fetching just the partition status counts rather than filtering things on the front-end. This makes it considerably faster 🙌 

I also changed the UI to match what we show for a collection of runs on the timeline view because @hellendag  was correct that the text was going to be truncated in many cases.

I also made the "Requested" column appear in the Instance > Backfills page, just without the bar graph showing which partitions were launched. 

image

image

PR #11704, opened 2023-01-13
RFC - db io manager partition support 1/n

jamiedemaria:
### Summary & Motivation
The goal of this RFC is to prototype/demonstrate the method for supporting partitions for database io managers that requires the least invasive changes to dagster code. There are other, more invasive changes that could be made instead, see this [RFC](https://github.com/dagster-io/dagster/pull/10976) and this [notion page](https://www.notion.so/dagster/Proposed-Snowflake-integration-improvements-63fec38f93964dbaa179cf0b1796f0a2) for an overview of what those changes would look like. I'm interested to hear your thoughts on this minimal change version vs one of the more involved changes. 

--- 

Part 1 of the least invasive set of changes I think we need to make to support partitions in the db io managers. 

Problem: To specify which column of the table to partition on, the user has to specify `partition_expr` as metadata on the asset. This is confusing, as it's unclear what `partition_expr` is supposed to be.

Minimal solution: Change `partition_expr` to `partition_column` (in this PR)

A more dramatic change would be one of the solutions outlined in this [RFC](https://github.com/dagster-io/dagster/pull/10976). This would entail adding a new parameter to `asset` called `io_manager_metadata` that could take various classes with IO manager specific metadata (see Sandy's code snippet [here](https://github.com/dagster-io/dagster/pull/10976#issuecomment-1347467465))


sryza (commented):
Could the issue raised here potentially be resolved by documenting `partition_expr`?

Here's how I would explain `partition_expr`: "A SQL expression that allows tying records in the table to partitions in the asset's PartitionsDefinition. In many cases, this will just be the name of a column.  If the asset is partitioned by time window, the expression should return a timestamp, and the record will correspond to partition whose time window it's inside. Otherwise, the expression should return a value that corresponds to one of the asset's partition keys."

An advantage of using an expression vs. a column is that it allows partitions to correspond to composites of multiple columns.  E.g. if someone has separate columns for month and year.

I'm not tied to `partition_expr`, but I think that, if we're considering other options, it would be helpful to write up the documentation side by side and see if one comes out clearer.

sryza:
Another situation that might be tough to handle with partition_column + partition_conversion: a string column with values formatted like 9/12/2022.

jamiedemaria:
ah i guess i never put together that `partition_expr` could be anything but the name of a column. that's where i was thinking `partition_column` would be a better name because it's more descriptive of what we expect. but i just ran across this in the full feature project 

```
def _time_window_where_clause(self, time_window: Tuple[datetime, datetime]) -> str:
        start_dt, end_dt = time_window
        return f"""WHERE TO_TIMESTAMP(time::INT) BETWEEN '{start_dt.strftime(SNOWFLAKE_DATETIME_FORMAT)}' AND '{end_dt.strftime(SNOWFLAKE_DATETIME_FORMAT)}'"""
```
so `partition_expr` makes a lot more sense now since the user could provide `TO_TIMESTAMP(time::INT)`

PR #11705, opened 2023-01-13
RFC - db io manager partition support 2/n

jamiedemaria:
### Summary & Motivation
The goal of this RFC is to prototype/demonstrate the method for supporting partitions for database io managers that requires the least invasive changes to dagster code. There are other, more invasive changes that could be made instead, see this [RFC](https://github.com/dagster-io/dagster/pull/10976) and this [notion page](https://www.notion.so/dagster/Proposed-Snowflake-integration-improvements-63fec38f93964dbaa179cf0b1796f0a2) for an overview of what those changes would look like. I'm interested to hear your thoughts on this minimal change version vs one of the more involved changes. 

---
Part 2 of the set of changes I think we need to make to support partitions for the DB IO managers

Problem: The DB IO managers expect the time data in the partition column to be a string of the form `2023-01-13 12:45:30`. But some users may not want to store data that way. For example, in our full feature example, the time data is a posix time. 

Solution: Allow users to specify a conversion function that tells us how to convert a partition key to the format the data is stored as. This will allow the SELECT statement to work correctly.

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/db_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/db_io_manager.py
@@ -1,5 +1,8 @@
 from abc import ABC, abstractmethod
+from datetime import datetime
 from typing import (
+    Any,
+    Callable,
     Dict,
     Generic,
     Mapping,
@@ -27,6 +30,7 @@ T = TypeVar("T")
 class TablePartition(NamedTuple):
     time_window: TimeWindow
     partition_column: str
+    partition_conversion: Optional[Callable[[datetime], Any]]
 
 
 class TableSlice(NamedTuple):
@@ -203,9 +207,13 @@ class DbIOManager(IOManager):
                     f"Asset '{context.asset_key}' has partitions, but no 'partition_column'"
                     " metadata value, so we don't know what column to filter it on."
                 )
+            conversion = cast(
+                Callable[[datetime], Any], output_context_metadata.get("partition_key_conversion")
+            )
             partition = TablePartition(
                 time_window=time_window,
                 partition_column=partition_column,
+                partition_conversion=conversion,
             )
         else:
             partition = None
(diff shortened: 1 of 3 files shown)


jamiedemaria:
* master
  * **PR #11704** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11704" target="_blank">Graphite</a> 
    * **PR #11705** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11705" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11705?utm_source=stack-comment).

jamiedemaria:
One of my main concerns with supporting partitions via metadata on the asset is how it translates to multi partitions. For multi partitions we would need to specify N columns and N conversion functions. 

One option would be for it to look like this
```
@asset(
    partitions_def=MultiPartitionsDefinition(
        {
            "date": DailyPartitionsDefinition(start_date="2022-01-01"),
            "color": StaticPartitionsDefinition(["red", "yellow", "blue"]),
        }
    )
    metadata={
        "partition_column": ["date", "color"],
        "partition_key_conversion": {
            "date": lambda x: x.timestamp(),
            "color" lambda x: x.lower()
        }
    }
)
def multi_partitions_asset(context):
   ....
```

or like this
```
@asset(
    partitions_def=MultiPartitionsDefinition(
        {
            "date": DailyPartitionsDefinition(start_date="2022-01-01"),
            "color": StaticPartitionsDefinition(["red", "yellow", "blue"]),
        }
    )
    metadata={
        "partition_info": {
            "date": lambda x: x.timestamp(),
            "color" lambda x: x.lower()
        }
    }
)
def multi_partitions_asset(context):
   ....
```
and we get the columns from the keys in the metadata

PR #11713, opened 2023-01-13
fix issue with k8s_job_op when combining command and container_config

gibsondan:
Summary:
You can't add a key to an immutable dict.

Test Plan: New test case

### DIFF
--- a/python_modules/libraries/dagster-k8s/dagster_k8s/ops/k8s_job_op.py
+++ b/python_modules/libraries/dagster-k8s/dagster_k8s/ops/k8s_job_op.py
@@ -203,7 +203,7 @@ def execute_k8s_job(
         else None,
     )
 
-    container_config = container_config or {}
+    container_config = container_config.copy() if container_config else {}
     if command:
         container_config["command"] = command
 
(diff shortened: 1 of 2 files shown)


python_modules/libraries/dagster-k8s/dagster_k8s/ops/k8s_job_op.py
```diff
@@ -203,7 +203,7 @@ def execute_k8s_job(
         else None,
     )
 
-    container_config = container_config or {}
+    container_config = container_config.copy() or {}
```
alangenfeld, line 206: should we update the mypy type to `Mapping` to indicate it is / can be read only?

alangenfeld (approved):

python_modules/libraries/dagster-k8s/dagster_k8s/ops/k8s_job_op.py
gibsondan: I like this idea but it doesn't like calling .copy() on a Mapping

gibsondan: (mypy doesn't)

schrockn: To do this with Mapping you can use unpack it using `**`.

`new_dict = {**mapping}`

Issue #9671, opened 2022-09-13
ECS Task Executor


gibsondan:
### What's the use case?

Running each op of a Dagster job in its own ECS task (similar to the k8s_job_executor) to make it easier to horizontally scale jobs with many ops, and give better isolation between ops.

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

k1sauce:
Hi, I'd like to add my input. I am trying to build a bioinformatics pipeline in Dagster. A bit of background, most bioinformatics tools are command line tools; viz. BWA, samtools, bedtools, GATK, fasqc, etc. In dragster this means I need to have these tools available in my environment and run some kind of command shell op. For example I have an op that downloads a BAM file and computes statistics on it with samtools. I then read the output of the shell command and return it. 

To do this kind of thing with ECS it makes sense to have each op as a different ECS task. That way I can use different containers for each op, where the container has the bioinformatics command line utility required by the op. Also some of these ops use a large amount of resources (CPU/Memory) and other don't. I would like to scope the resources to the op based on the expected computational workload.

nicor88:
@gibsondan will this feature be implemented any time soon? Alternatively I could implement an `@op` where using boto3 I trigger the ecs task, watch for it, and then raise an error in case of failure.

Why I'm interested on this?
I generally like to have my orchestration separated from dbt, and I don't want to copy dbt models around, instead I bake a dbt image with all the models, and then trigger the image as container (either in ECS or K8S), I've working with this solution with airflow, but I would like to consider dagster instead of airflow for orchestration.

chasleslr:
Here's some additional input;
It's very typical for machine learning pipelines to include steps that need to execute on very different hardware -- data transformation may need lots of cpu/memory or very little (i.e. executed on a remote spark/dask cluster) and training may requires instances with GPU. A single job would often need to do all these steps, and running everything on a GPU instance may be cost prohibitive.

kaush-13:
any updates on if this will ever be supported?

yuhan:
Possible workaround is to the scaling issue: https://gist.github.com/yuhan/bd477c93cc36bc6528acc114a82578b9 which builds a run failure sensor to fetch info about the failed run due to limited resources, and then launch a new run with the updated tags and only run the selected assets.

PR #11735, opened 2023-01-16
RFC: Native multi-environment execution

schrockn:
### Summary & Motivation

We have some community members clamoring for more flexible execution models that (1) separate the orchestration environment from the compute/business logic environment (2) run steps in distinct python environments in Dagster-native way (meaning using `@asset` and `@op` in the remote process, not just the execution of raw code) and (3) provide a more lightweight of launching remote steps than the step launcher that integrates with existing deployment processes and existing code.

For example w/r/t to orchestration environment and compute environment separation, we have users who want to orchestrate on Dagster Cloud but have their steps execute in isolated environments withn AWS Lambda.

We have another user that wants to use the Dagster programming model within Databricks, but they want to use their existing mechanisms for deploying code, have their orchestration and transformation logic in separate github repos, and do not want to the launching environment to have a pyspark/jvm dependency. These requirements preclude the usage of existing pyspark step launchers.

This RFC contains a few changes that allow for these scenarios.

* `MultiEnvironmentExecutor`: This is an executor that allows each step of execution to delegate to a distinct environment. It leverages the existing `StepDelegatingExecutor`, but provides a `StepHandler` that in turn composes a new concept a `RemoteEnvironmentSingleStepHandler` that allows for remote execution on a per step basis. One can return different instances of `RemoteEnvironmentSingleStepHandler` for each step. One variant might launch a k8s pod; another might launch a spark job. This, in effect, enables native multi-container/environment execution  within the run worker.
* `execute_asset_within_inprogress_run` and `execute_op_within_inprogress_run`: These are new public APIs provided to users for usage in the worker processes. This allows users to control code deployment and the "main" function in the external process. Note how this constrasts with step launchers, that control the entry point in the target runtime. Instead here, we assume that the user will take responsibility for deploying code to the target runtime and has their own entry point, and they call back _into_ our library to participate in a Dagster run. 

These take an asset/op definitions respectively, a run_id, a step_key, and a Dagster Instance. Any further information needed for executed is stashed in the Instance's KV store and fetched in the worker process.

image

In terms of metadata, we need to represent ops and assets in both processes. The asset-based tests are how I imagine this working, where we generate stub `AssetsDefinitions` that do not implement an execution function. However they do have metadata for display, and it will be the responsibility of the `MultiEnvironmentStepHandler` to stash all the data needed for execution so that the worker process can retrieve it.

A big next step here is further fleshing out this metadata interchange. Partitions are a critical thing. Partitions will have to be represented in both the launching and the worker process, and information like the current partition key will have be passed to the worker process.

Before proceeding further I wanted to get feedback on the approach.
 
Included unit tests. Also used Dagit to run unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_core/executor/multi_environment/execute_out_of_process.py
+++ b/python_modules/dagster/dagster/_core/executor/multi_environment/execute_out_of_process.py
@@ -1,0 +1,117 @@
+from dagster import _check as check
+from dagster._core.code_pointer import CodePointer
+from dagster._core.definitions.assets import AssetsDefinition
+from dagster._core.definitions.definitions_class import Definitions
+from dagster._core.definitions.job_definition import JobDefinition
+from dagster._core.definitions.source_asset import SourceAsset
+from dagster._core.definitions.unresolved_asset_job_definition import define_asset_job
+from dagster._core.executor.multi_environment.remote_step_information import (
+    get_remote_step_info,
+)
+from dagster._core.instance import DagsterInstance
+from dagster._core.origin import PipelinePythonOrigin, RepositoryPythonOrigin
+from dagster._core.storage.pipeline_run import DagsterRun
+from dagster._grpc.types import ExecuteStepArgs
+
+
+# So that we can use PipelinePythonOrigin to wrap an ephemeral, in-process job
+class InMemoryCodePointer(CodePointer):
+    def __init__(self, target):
+        self.target = target
+
+    def load_target(self) -> object:
+        return self.target
+
+    def describe(self) -> str:
+        return "inmem"
+
+
+def pipeline_origin_for_op(job_def: JobDefinition):
+    return pipeline_origin_for_job_in_defs(Definitions(jobs=[job_def]), job_def.name)
+
+
+def pipeline_origin_for_job_in_defs(
+    definitions: Definitions, job_name: str
+) -> PipelinePythonOrigin:
+    repo = definitions.get_inner_repository_for_loading_process()
+    return PipelinePythonOrigin(
+        pipeline_name=job_name,
+        repository_origin=RepositoryPythonOrigin(
+            executable_path="invalid", code_pointer=InMemoryCodePointer(repo)
+        ),
+    )
+
+
+def execute_op_within_inprogress_run(
+    instance: DagsterInstance,
+    job_def: JobDefinition,
+    run_id: str,
+    step_key: str,
+):
+    fake_pipeline_origin = pipeline_origin_for_op(job_def)
+    dagster_run = check.not_none(instance.get_run_by_id(run_id))
+    dagster_run = dagster_run._replace(pipeline_code_origin=fake_pipeline_origin)
+    execute_step_in_out_of_proc_step(instance, dagster_run, step_key)
+
+
+def execute_step_in_out_of_proc_step(
+    instance: DagsterInstance,
+    dagster_run: DagsterRun,
+    step_key: str,
+):
+    run_id = dagster_run.run_id
+    remote_step_info = get_remote_step_info(instance=instance, run_id=run_id, step_key=step_key)
+
+    step_args = ExecuteStepArgs(
+        pipeline_origin=check.not_none(dagster_run.pipeline_code_origin),
+        pipeline_run_id=run_id,
+        step_keys_to_execute=[step_key],
+        instance_ref=None,
+        retry_mode=remote_step_info.retry_mode,
+        known_state=remote_step_info.known_state,
+        should_verify_step=False,
+    )
+
+    from dagster._cli.api import _execute_step_command_body
+
+    for event in _execute_step_command_body(step_args, instance, dagster_run):
+        print(f"processing event in worker {event}")
+
+
+def defs_for_single_assets_def(assets_def: AssetsDefinition, job_name: str):
+    source_assets = []
+    for _input_name, asset_keys_for_input in assets_def.keys_by_input_name.items():
+        for asset_key in asset_keys_for_input:
+            source_assets.append(SourceAsset(asset_key))
+
+    return Definitions(
+        jobs=[define_asset_job(job_name)],
+        assets=[assets_def, *source_assets],
+    )
+
+
+def execute_asset_within_inprogress_run(
+    instance: DagsterInstance,
+    assets_def: AssetsDefinition,
+    run_id: str,
+    step_key: str,
+):
+    dagster_run = check.not_none(instance.get_run_by_id(run_id), "Run must exist in instance")
+    job_name = dagster_run.pipeline_name
+
+    in_worker_process_origin = pipeline_origin_for_job_in_defs(
+        defs_for_single_assets_def(assets_def, job_name),
+        job_name,
+    )
+
+    dagster_run = dagster_run._replace(
+        pipeline_code_origin=in_worker_process_origin,
+        # right now if you do materialize all in dagit it actually fully
+        # expands the asset selection to be all the assets. when this
+        # gets passed through it fails a subsetting check because the
+        # source asset is not executable. So here we only execute the single
+        # asset key
+        asset_selection=frozenset([assets_def.key]),
+    )
+
+    execute_step_in_out_of_proc_step(instance, dagster_run, step_key)
--- a/python_modules/dagster/dagster/_core/executor/multi_environment/remote_step_information.py
+++ b/python_modules/dagster/dagster/_core/executor/multi_environment/remote_step_information.py
@@ -1,0 +1,39 @@
+from typing import NamedTuple, Optional
+
+from dagster._core.execution.plan.state import KnownExecutionState
+from dagster._core.execution.retries import RetryMode
+from dagster._core.instance import DagsterInstance
+from dagster._serdes.serdes import (
+    deserialize_as,
+    serialize_dagster_namedtuple,
+    whitelist_for_serdes,
+)
+
+
+@whitelist_for_serdes
+class RemoteStepInformation(NamedTuple):
+    retry_mode: Optional[RetryMode]
+    known_state: Optional[KnownExecutionState]
+
+
+def get_kv_key_for_step_args(run_id: str, step_key: str) -> str:
+    return f"STEPARGS:{run_id}:{step_key}"
+
+
+def set_remote_step_info(
+    instance: DagsterInstance, run_id: str, step_key: str, remote_step_info: RemoteStepInformation
+):
+    step_args_kv_key = get_kv_key_for_step_args(run_id, step_key)
+    return instance.kvs_set(
+        {
+            step_args_kv_key: serialize_dagster_namedtuple(remote_step_info),
+        }
+    )
+
+
+def get_remote_step_info(
+    instance: DagsterInstance, run_id: str, step_key: str
+) -> RemoteStepInformation:
+    kv_key = get_kv_key_for_step_args(run_id, step_key)
+    remote_step_info_json = instance.kvs_get({kv_key})[kv_key]
+    return deserialize_as(remote_step_info_json, RemoteStepInformation)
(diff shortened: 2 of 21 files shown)


schrockn:
* master
  * **PR #11735** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11735" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11735?utm_source=stack-comment).

python_modules/dagster/dagster/_core/executor/multi_environment/execute_out_of_process.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+
+    dagster_run = dagster_run._replace(
+        pipeline_code_origin=in_worker_process_origin,
+        # right now if you do materialize all in dagit it actually fully
+        # expands the asset selection to be all the assets. when this
+        # gets passed through it fails a subsetting check because the
+        # source asset is not executable. So here we only execute the single
+        # asset key
+        asset_selection=frozenset([assets_def.key]),
```
schrockn, line 114: This is definitely the most sketchy part of this. We fake out the dagster run in the worker process to make all of this work.

python_modules/dagster/dagster/_core/executor/multi_environment/remote_step_information.py
```diff
@@ -0,0 +1,39 @@
...
+from dagster._core.execution.retries import RetryMode
+from dagster._core.instance import DagsterInstance
+from dagster._serdes.serdes import (
+    deserialize_as,
+    serialize_dagster_namedtuple,
+    whitelist_for_serdes,
+)
+
+
+@whitelist_for_serdes
```
schrockn, line 13: Give a man a KV store and Lord knows what he'll stash in there!

python_modules/dagster/dagster_tests/execution_tests/engine_tests/multienvironment_tests/asset_variant/test_multi_environment_asset_execution.py
```diff
@@ -0,0 +1,143 @@
...
+def get_defs_for_two_assets():
+    # Imagine parsing a yaml file or other arbitrary metadata format and creating
+    # RemoteAsset instances
+    return build_definitions_for_remote_assets(
+        "two_assets_job",
+        [
+            remote_asset_in_script("one_asset", []),
+            remote_asset_in_script("add_one_to_one_asset", ["one_asset"]),
+        ],
+    )
```
schrockn, line 122: @sryza here is the part where we would process a yaml file in the orchestration/launching repo and construct a bunch of stub AssetsDefinitions

python_modules/dagster/dagster_tests/execution_tests/engine_tests/multienvironment_tests/asset_variant/test_multi_environment_asset_execution.py
```diff
@@ -0,0 +1,143 @@
...
+
+    assets_to_handlers = {
+        remote_asset.asset_key: remote_asset.single_step_handler for remote_asset in remote_assets
+    }
+
+    return Definitions(
+        assets=assets_defs,
+        jobs=[define_asset_job(asset_job_name, "*")],
+        executor=MultiEnvironmentExecutor(
+            lambda step_context: assets_to_handlers[step_context.step.key],
```
schrockn, line 84: and here it is the responsibility of the user to write a function that maps a step context in the launching process to a remove step handler

gibsondan: Is this something we would consider supporting via config or tags? it seems pretty powerful to be able to say something like "by default every step for this job should run in a separate k8s pod, except for this step which should run in a subprocess" without needing to write a lambda

And doing the mapping in user code locks us in to run workers being a user code process right? Which is probably fine at this point, but just flagging..

sryza:
`MultiEnvironmentExecutor` - sounds awesome.

`execute_asset_within_inprogress_run` - The cost of adding one or two experimental methods to the codebase is low, so I won't block that.  I do have doubts about whether this is a direction we want to ultimately support at all.  Even if we can come up with a good set of APIs for this, I have doubts about whether it's worth it to add them.  It would add a new dimension to the matrix of ways to use Dagster, which likely means significant maintenance burden for us and extra opportunities for confusion for our users.  To feel comfortable with this direction, IMO we need:
- A clearer picture of what the APIs would look like
- A case that this is something that a significant segment of users care about

A couple points on `import pyspark` FWIW:
- You don't need a JVM to `import pyspark` (just tested this out).
- If they want to interact with their data in their sensors, there's a good chance they'll need to `import pyspark` anyway.

python_modules/dagster/dagster_tests/execution_tests/engine_tests/multienvironment_tests/asset_variant/test_multi_environment_asset_execution.py
schrockn: I think it would be user code in the same way that DagsterInstance is "user code" because of the ConfigurableClass stuff.

schrockn: Re: supporting tags I think that is a great idea.

schrockn:
I for one definitely think is a direction we want to go and support in a first-class way over time horizon. It allows for greater flexibility and the reality is that lots of users want to separate their orchestration and compute environments and have good reasons for doing so. I actually think that the top-level API `execute_asset_within_inprogress_run` is a pretty reasonable API.

Re: sensors if someone wants to run sensors using the same environment as their spark cluster or whatever they can just set up a code location for that purpose expressly

schrockn:
Another important use case for this are people who are extensively using K8PodOperators in Airflow, want to use the Dagster programming model, but do not want to merge all their Python environments. I believe there are a large number of uses that fit that bill. Same with people migrating off of Argo/Kubeflow.

python_modules/dagster/dagster/_core/executor/multi_environment/execute_out_of_process.py
```diff
@@ -0,0 +1,117 @@
...
+        assets=[assets_def, *source_assets],
+    )
+
+
+def execute_asset_within_inprogress_run(
+    instance: DagsterInstance,
+    assets_def: AssetsDefinition,
+    run_id: str,
+    step_key: str,
+):
```
alangenfeld, line 98: [1] Seeing this makes me think a shared entry point would be tough, is that not desirable? The tests show a separate py file entry point for each asset, which seems cumbersome to me. Curious to hear more on the thinking here.

python_modules/dagster/dagster/_core/executor/multi_environment/execute_out_of_process.py
```diff
@@ -0,0 +1,117 @@
...
+        ),
+    )
+
+
+def execute_op_within_inprogress_run(
+    instance: DagsterInstance,
+    job_def: JobDefinition,
+    run_id: str,
+    step_key: str,
+):
```
alangenfeld, line 50: [1]

python_modules/dagster/dagster/_core/executor/multi_environment/remote_step_information.py
alangenfeld: definitely need to figure out some expirey/removal for this data if we are going to use the KV store for it so it doesn't pile up endlessly.

python_modules/dagster/dagster/_core/executor/multi_environment/execute_out_of_process.py
schrockn: I think for the intended design partner we would advise them to map step_key ==> assets_def.

In this case I did sep files to simulate multiple environments. E.g. imagine each of those with a different virtual env and deps

alangenfeld: step key -> asset def sketches me out in the general case, but i suppose the conceit here is that the user is setting up the custom step handler and entry point in lock step and nothing precludes them from passing additional information along for look-up or ensuring their definitions conform to rules that make step_key -> asset def resolution safe.

schrockn:
In terms of requests for multi-environment execution more are coming up today: https://github.com/dagster-io/dagster/issues/9671

Screen Shot 2023-01-20 at 1 22 45 PM


https://dagster.slack.com/archives/C02LJ7G0LAZ/p1674237818744389?thread_ts=1674236799.847639&cid=C02LJ7G0LAZ

python_modules/dagster/dagster/_core/executor/multi_environment/remote_step_information.py
schrockn: @alangenfeld I think it is pretty safe to clean up on run completion. however we need a way to index into that...

schrockn:
So reflecting on this @sryza I guess I'm a bit confused as to why you are excited about the multi-environment execution without the corresponding changes to add `execute_asset_within_inprogress_run` or something like it for invocation on the other side. The current execution system has fairly deep assumptions that identical python environments exist on both sides of the process boundary with the reconstructable machinery and how PipelinePythonOrigin works.

What the MultiEnvironmentExecutor as written alone would allow would be for one step to execute in a k8s container and another step to execute in a different substrate (a process on the same node as the run worker). But the would have the same Python environment. I don't think that is what users are asking for and it has limited utility.

schrockn:
@gibsondan In terms of path forward here to incrementally make progress I am going to plan on doing this per-step refactor with the express goal of making it more straightforward to write custom executors in terms of a per-step handler rather than writing an entirely new executor or an existing StepHandler. However this doesn't result in a deep execution model change.

sryza:
> So reflecting on this @sryza I guess I'm a bit confused as to why you are excited about the multi-environment execution without the corresponding changes to add execute_asset_within_inprogress_run or something like it for invocation on the other side. The current execution system has fairly deep assumptions that identical python environments exist on both sides of the process boundary with the reconstructable machinery and how PipelinePythonOrigin works.

I would have found this very useful at KeepTruckin.  Our situation:
- Single set of Python dependencies
- 1/3 of transformations were SQL on Snowflake - ideally in-process executor because all the compute is reote
- 1/3 of transformations were Pandas/ML/Python - needed beefy nodes
- 1/3 of transformations were Spark - needed to execute inside Qubole (same situation as Databricks)

schrockn:
Got it that makes sense. So with the scaled back per-step executor and no other architectural changes we would. end up with:

* k8 pod pre step  for the Python/Pandas nodes
* Modest subprocess steps for SQL in Snowflake
* Modest subprocess steps combined with customer step launcher for Qubole.

sryza:
> Modest subprocess steps combined with customer step launcher for Qubole.

If the executor is pluggable, the user could plug in something that launches the step directly on Qubole instead of using a step launcher, right?

schrockn:
@sryza What that leaves is the problem that `execute_asset_within_inprogress_run` is trying to solve, which is an ergonomic way to callback into a step in the external process in the cluster. 

In the qubole example that you note it would be left write the main function that ends up invoking `run_step_from_ref`, which is a huge part of the step launcher machinery, so I think this only solves "half" of the problem.

sryza:
> What that leaves is the problem that execute_asset_within_inprogress_run is trying to solve, which is an ergonomic way to callback into a step in the external process in the cluster.

If I understand correctly, the reason we can't just do what we do in K8s in Qubole is that in K8s we have the opportunity to launch a CLI command, but, in Qubole, we're already inside a Python process?

If so, then yes, I definitely agree that we want a utility like `execute_asset_within_inprogress_run` to help authors of `RemoteEnvironmentSingleStepHandler`.

To narrow my concern, it's about using a new model of loading user code.  I.e., if I understand correctly, the implementation of `execute_asset_within_inprogress_run` does not rely on reconstructable pipelines.  This implies enabling a divergence between the Python definition objects in the gRPC server and the Python definition objects in the step process. I suspect that divergence has wide-ranging consequences. My point in my first comment in this PR was that I think those consequences are worth understanding in more detail out before committing ourselves to this path.

sryza (changes requested):
Back to your queue

Issue #11174, opened 2022-12-16
Render materialization-level lineage in Dagit


sryza:
As of Dagster 1.1, each asset materialization records the event log storage_ids of the upstream materializations that it's based on.

When viewing a materialization, we could provide links to the upstream materializations.

bengotow:
We do this now!

PR #11736, opened 2023-01-16
[dagit] Display source data used in specific asset materializations

bengotow:
### Summary & Motivation

This PR addresses #11174 - a new section of Dagit's materialization event / partition detail view shows the upstream materializations that were "source data" for the displayed event.

This uses a new GraphQL resolver built on the underlying storage + fetching for asset versioning.

image

There's a new unit test, but I also tested this on a large set of existing materializations, verified it does not appear for observations, and checked that navigating the referenced assets produces internally consistent views. (eg: clicking B in A->B->C->D still shows the same C->D)

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/AssetEventDetail.tsx
+++ b/js_modules/dagit/packages/core/src/assets/AssetEventDetail.tsx
@@ -4,6 +4,7 @@ import {Link} from 'react-router-dom';
 
 import {Timestamp} from '../app/time/Timestamp';
 import {isHiddenAssetGroupJob} from '../asset-graph/Utils';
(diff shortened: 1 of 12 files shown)


sryza (commented):
Will leave it to others to review, but very excited about this. If we think this will be expensive to load, could we make it not block the rendering of other materialization info?

hellendag (approved):

clairelin135 (approved):
This looks good to me! Looks like we're fetching this data on a per-materialization basis. I think this will be expensive to calculate for many materializations/assets at once, so as long as we're not fetching this for each materialization or for multiple assets this is looking good to me.

PR #11737, opened 2023-01-16
open grpc servers in their own process group so they don't get interrupted along with whatever process opened them

gibsondan:
summary:
We were already doing this on windows - lets do it on linux too. Setting this function prevents the grpc servers from being CTRL-Ced whenever the process that opened them got CTRL-c - so we can use our fancier graceful termination logic.

BK, run dagit and CTRL-C it - no more gross "Aborted!" messages for the grpc servers

### DIFF
--- a/python_modules/dagster/dagster/_serdes/ipc.py
+++ b/python_modules/dagster/dagster/_serdes/ipc.py
@@ -188,13 +188,20 @@ def ipc_read_event_stream(file_path, timeout=30, ipc_process=None):
 
 
 def open_ipc_subprocess(parts, **kwargs):
-    """Sets new process group flags on Windows to support graceful termination."""
+    """Sets the correct flags to support graceful termination - the new subprocess will be in
+    its own process group so that CTRL-Cing the parent process doesn't immediately interrupt
+    the child process as well.
+    """
     check.list_param(parts, "parts", str)
 
     creationflags = 0
+    preexec_fn = None
     if sys.platform == "win32":
         creationflags = subprocess.CREATE_NEW_PROCESS_GROUP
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_serdes/ipc.py
```diff
@@ -188,13 +188,19 @@ def ipc_read_event_stream(file_path, timeout=30, ipc_process=None):
...
+    """
     check.list_param(parts, "parts", str)
 
     creationflags = 0
+    preexec_fn = None
     if sys.platform == "win32":
         creationflags = subprocess.CREATE_NEW_PROCESS_GROUP
-    return subprocess.Popen(parts, creationflags=creationflags, **kwargs)
+    else:
+        preexec_fn = os.setpgrp
```
schrockn, line 202: Does change what operating systems we can run on at all? Exclude any linux variants or anything? What about Mac?

gibsondan: https://docs.python.org/3/library/os.html#os.setpgrp says "Unix, not Emscripten, not WASI." Mac is fine, that's what i'm testing on

schrockn: Ah WebAssembly stuff. Ok.

schrockn: That is worth a comment in this code path.

schrockn (changes requested):
need rebase

schrockn (approved):
Lovely.

Please comment in that `os.setpgrp` for a future theoretical engineer trying to make this work in web assembly.

PR #11738, opened 2023-01-16
Make StaticPartitionMapping serializable. Add autodoc.

AlexanderVR:
### Summary & Motivation

Changes made by https://github.com/dagster-io/dagster/pull/11521 do not work in tandem with dagit due to lack of serialization.

This PR adds serde whitelisting similar to TimeWindowPartitionMapping.

Also adds basic class autodoc.

`dagit -f partitioned.py`

with the following as partitioned.py:
```python
from dagster import (
    AssetIn,
    Definitions,
    StaticPartitionMapping,
    StaticPartitionsDefinition,
    asset,
)

upstream_parts = StaticPartitionsDefinition(["p1", "p2", "p3", "q1", "q2", "r1"])
downstream_parts = StaticPartitionsDefinition(["p", "q", "r"])
mapping = StaticPartitionMapping({"p1": "p", "p2": "p", "p3": "p", "r1": "r"})


@asset(partitions_def=upstream_parts)
def a(context):
    return "key-" + context.asset_partition_key_for_output()


@asset(
    partitions_def=downstream_parts,
    ins={"up": AssetIn("a", partition_mapping=mapping)},
)
def b(up) -> dict:
    return up


defs = Definitions(
    assets=[a, b],
)
```

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/partition_mapping.py
+++ b/python_modules/dagster/dagster/_core/definitions/partition_mapping.py
@@ -332,7 +332,12 @@ class StaticPartitionMapping(
     PartitionMapping,
     NamedTuple(
         "_StaticPartitionMapping",
-        [("downstream_partition_keys_by_upstream_partition_key", PublicAttr[Mapping])],
+        [
+            (
+                "downstream_partition_keys_by_upstream_partition_key",
+                PublicAttr[Mapping[str, Union[str, Collection[str]]]],
+            )
+        ],
     ),
 ):
     """
--- a/python_modules/dagster/dagster/_core/definitions/partition_mapping.py
+++ b/python_modules/dagster/dagster/_core/definitions/partition_mapping.py
@@ -4,7 +4,7 @@ from collections import defaultdict
 from typing import Collection, Mapping, NamedTuple, Optional, Union, cast
 
 import dagster._check as check
-from dagster._annotations import experimental, public
+from dagster._annotations import PublicAttr, experimental, public
 from dagster._core.definitions.multi_dimensional_partitions import (
     MultiPartitionKey,
     MultiPartitionsDefinition,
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -327,7 +327,14 @@ def get_downstream_partitions_for_partitions(
         return downstream_partitions_def.empty_subset().with_partition_keys(set(matching_keys))
 
 
-class StaticPartitionMapping(PartitionMapping):
+@whitelist_for_serdes
+class StaticPartitionMapping(
+    PartitionMapping,
+    NamedTuple(
+        "_StaticPartitionMapping",
+        [("downstream_partition_keys_by_upstream_partition_key", PublicAttr[Mapping])],
```
sryza, line 335: Mind including the full parameterized type here?

sryza (changes requested):
Thanks for the PR @AlexanderVR. I think that, for this to work, you need to define `__new__` instead of `__init__`.  If you look at other classes in Dagster that extend `NamedTuple`, you'll  see what this looks like.

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
AlexanderVR: fixed

AlexanderVR:
> Thanks for the PR @AlexanderVR. I think that, for this to work, you need to define `__new__` instead of `__init__`. If you look at other classes in Dagster that extend `NamedTuple`, you'll see what this looks like.

So this actually works because classes inheriting from namedtuples will store private attributes -- here: _mapping and _inverse_mapping -- to the `__dict__` attribute of the class.
Since we are not adding custom logic to set the public attribute `downstream_partition_keys_by_upstream_partition_key` at class creation time I don't see a need for defining `__new__`, unless you just prefer that style.

I've added a test to show that dagster's namedtuple serde still works.

If this seems a little too magic, one idea is to use functools.cached_property to define these private attributes. But that will break compatibility with python 3.7.

sryza (approved):

PR #11739, opened 2023-01-16
Don't spew "Aborted!" to the command-line when the Daemon is cleanly interrupted

gibsondan:
Right now whenever you CTRL-C the daemon (or `dagster dev` terminates the daemon) we hit some generic click logic that prints out a scary "Aborted!" message, even though its the expected way to end the process. Instead of
bubbling the interrupt up to click, catch it in the CLI command and exit cleanly.

uvicorn in dagit has similar logic that causes dagit to exit with return code 0 when an interrupt happens.

Test Plan:
Run daemon locally, CTRL-C, no more scary "Aborted!" message, same with dagster dev

### DIFF
--- a/python_modules/dagster/dagster/_daemon/cli/__init__.py
+++ b/python_modules/dagster/dagster/_daemon/cli/__init__.py
@@ -48,11 +48,14 @@ def _get_heartbeat_tolerance():
 )
 @workspace_target_argument
 def run_command(code_server_log_level, instance_ref, **kwargs):
-    with capture_interrupts():
-        with DagsterInstance.from_ref(
-            deserialize_as(instance_ref, InstanceRef)
-        ) if instance_ref else DagsterInstance.get() as instance:
-            _daemon_run_command(instance, code_server_log_level, kwargs)
+    try:
+        with capture_interrupts():
+            with DagsterInstance.from_ref(
+                deserialize_as(instance_ref, InstanceRef)
+            ) if instance_ref else DagsterInstance.get() as instance:
+                _daemon_run_command(instance, code_server_log_level, kwargs)
+    except KeyboardInterrupt:
+        return  # Exit cleanly on interrupt
 
 
 @telemetry_wrapper(metadata={"DAEMON_SESSION_ID": get_telemetry_daemon_session_id()})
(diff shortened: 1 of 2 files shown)


schrockn (changes requested):
I think you need to rebase this

python_modules/dagster/dagster/_daemon/cli/__init__.py
```diff
@@ -48,11 +48,14 @@ def _get_heartbeat_tolerance():
...
-        ) if instance_ref else DagsterInstance.get() as instance:
-            _daemon_run_command(instance, code_server_log_level, kwargs)
+    try:
+        with capture_interrupts():
+            with DagsterInstance.from_ref(
+                deserialize_as(instance_ref, InstanceRef)
+            ) if instance_ref else DagsterInstance.get() as instance:
+                _daemon_run_command(instance, code_server_log_level, kwargs)
+    except KeyboardInterrupt:
+        return  # Exit cleanly on interrupt
```
schrockn, line 58: IMO we should log something here.

e.g. "Dagster daemon process cleaning up and exiting because it received a keyboard interrupt or something"

Essentially something that provides confidence to the user by showing them that we know exactly what is going on.

schrockn (approved):

gibsondan:
@schrockn we do log higher in the stack during cleanup, here is sample output after this change:

```
(dagster-3.8.9) dgibson@Daniels-MBP dagster % dagster-daemon run
2023-01-18 11:03:39 -0600 - dagster.daemon - INFO - Instance is configured with the following daemons: ['BackfillDaemon', 'SchedulerDaemon', 'SensorDaemon']
2023-01-18 11:03:39 -0600 - dagster.daemon.SensorDaemon - INFO - Not checking for any runs since no sensors have been started.
^C2023-01-18 11:03:42 -0600 - dagster.daemon - INFO - Shutting down daemon threads...
2023-01-18 11:03:42 -0600 - dagster.daemon - INFO - Daemon threads shut down.
```

schrockn:
One last line "Finished cleanup after keyboard interrupt instigated shutdown. Have a nice day!" or something would be nice. I like it when tools can express that they know what is going on precisely. Gives me a lot of confidence. Not a huge deal though so do what you feel is appropriate.

gibsondan:
Added something spiritually similar

PR #11741, opened 2023-01-17
New "running dagster locally" deployment guide that walks through `dagster dev` usage

gibsondan:
Summary:
Right now if you want to answer 'how do i run dagster' you have to kind of wade through the instance docs, and then the daemon docs if you want to use any schedules or sensors. This tries to streamline the question of 'how do i get dagster up and running locally' in a single guide using the new 'dagster dev' command.

Not sure about the exact information architecture here / if this even belongs under 'Deployment' per se - open to putting it elsewhere, but there was some existing stuff in Deployment, particularly in the Daemon section, that this somewhat replaces. The section labeled 'Open Source' is kind of alternately called 'Deploying to your own infra' elsewhere and this wasn't quite that, but then it's not under Open Source which is weird... I feel pretty good about the Content but am not quite sure where to fit it in.

### DIFF
--- a/docs/content/deployment/concepts.mdx
+++ b/docs/content/deployment/concepts.mdx
@@ -17,7 +17,7 @@ Learn about the concepts relevant to deploying Dagster. Refer to the [Core conce
 
 ## Dagster instance
 
-The `DagsterInstance` defines all of the configuration that Dagster needs for a single deployment - for example, where to store the history of past runs and their associated logs, where to stream the raw logs from op compute functions, and how to launch new runs.
+The Dagster instance defines all of the configuration that Dagster needs for a single deployment - for example, where to store the history of past runs and their associated logs, where to stream the raw logs from op compute functions, and how to launch new runs.
 
 [Learn more about setting up your Dagster instance](/deployment/dagster-instance).
 
--- a/docs/content/deployment/guides/running-locally.mdx
+++ b/docs/content/deployment/guides/running-locally.mdx
@@ -1,0 +1,107 @@
+---
+title: Running Dagster Locally | Dagster"
+description: How to run Dagster on your local machine.
+---
+
+# Running Dagster Locally
+
+The easiest way to start up Dagster services during local development is to run:
+
+```shell
+dagster dev
+```
+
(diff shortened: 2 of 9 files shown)


docs/content/deployment/concepts.mdx
```diff
@@ -17,7 +17,7 @@ Learn about the concepts relevant to deploying Dagster. Refer to the [Core conce
 
 ## Dagster instance
 
-The `DagsterInstance` defines all of the configuration that Dagster needs for a single deployment - for example, where to store the history of past runs and their associated logs, where to stream the raw logs from op compute functions, and how to launch new runs.
+The Dagster instance defines all of the configuration that Dagster needs for a single deployment - for example, where to store the history of past runs and their associated logs, where to stream the raw logs from op compute functions, and how to launch new runs.
```
schrockn, line 20: This paragraph makes me very motivated to consolidate instance and deployment

docs/content/deployment/guides/running-locally.mdx
```diff
@@ -0,0 +1,106 @@
...
+```shell
+dagster dev -f my_file.py -f my_second_file.py
+```
+
+---
+
+</TabItem>
+<TabItem name="From a module">
+
+Dagster can also load Python modules as code locations. When this approach is used, Dagster loads the definitions defined at the top-level of the module, in a variable containing the <PyObject object="Definitions" /> object of its root `__init__.py` file. As this style of development eliminates an entire class of Python import errors, we strongly recommend it for Dagster projects deployed to production.
```
schrockn, line 42: So this is an interesting case of verbiage. This paragraph says modules but is describing the behavior of installed modules, which is a subset of modules. `-m` can import vanilla files actually (just omit the .py)

schrockn (approved):
This is great. Approved but obviously @erinkcochran87 needs to take a pass.

docs/content/deployment/guides/running-locally.mdx
gibsondan: ah I just copied this over from the "deploying the daemon locally" docs, but can tweak

PR #11744, opened 2023-01-17
feat(dagster-dbt): use cached run id to fetch artifacts

rexledesma:
### Summary & Motivation
Start threading through the cached run id when generating the software defined assets in dbt Cloud.

pytest, local

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/asset_defs.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/asset_defs.py
@@ -49,6 +49,8 @@ from ..errors import DagsterDbtCloudJobInvariantViolationError
 from ..utils import ASSET_RESOURCE_TYPES, result_to_events
 from .resources import DbtCloudResource
 
+DAGSTER_DBT_COMPILE_RUN_ID_ENV_VAR = "DBT_DAGSTER_COMPILE_RUN_ID"
+
 
 class DbtCloudCacheableAssetsDefinition(CacheableAssetsDefinition):
     def __init__(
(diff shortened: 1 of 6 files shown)


python_modules/libraries/dagster-dbt/dagster_dbt/cloud/asset_defs.py
```diff
@@ -127,35 +129,19 @@ def get_compile_filters(parsed_args: Namespace) -> List[str]:
...
-                execute_steps=self._job_commands
-            )
+    def _get_cached_compile_dbt_cloud_job_run(self, compile_run_id: int) -> Tuple[int, int]:
+        compile_run = self._dbt_cloud.get_run(
+            run_id=compile_run_id, include_related=["trigger", "run_steps"]
         )
+        compile_run_has_generate_docs = compile_run["trigger"]["generate_docs_override"]
+
+        compile_job_materialization_command_step = len(compile_run["run_steps"])
+        if compile_run_has_generate_docs:
```
OwenKephart, line 139: why this method of figuring out the compile step rather than parsing the commands? fine for now, but it does seem a bit roundabout.

python_modules/libraries/dagster-dbt/dagster_dbt/cloud/asset_defs.py
```diff
@@ -213,8 +200,55 @@ def _get_dbt_nodes_and_dependencies(
...
+        )
+        compile_run_id = (
+            dbt_cloud_job_env_vars.get(DAGSTER_DBT_COMPILE_RUN_ID_ENV_VAR, {})
+            .get("job", {})
+            .get("value")
+        )
+
+        compile_run_id, compile_job_materialization_command_step = (
+            # If a compile run is cached, then use it.
+            self._get_cached_compile_dbt_cloud_job_run(compile_run_id=int(compile_run_id))
```
OwenKephart, line 247: not blocking, and mostly a matter of taste, but I think this might be easier to understand if there was a single self._get_compile_run_id_and_command_step (or self._get_manifest_and_run_results) function which branched its behavior internally

OwenKephart: as an aside, I wasn't aware that a compile command would generate a run_results.json file

OwenKephart (approved):

python_modules/libraries/dagster-dbt/dagster_dbt/cloud/asset_defs.py
rexledesma: We're going through dark magic here in the dbt Cloud APIs 🥲

python_modules/libraries/dagster-dbt/dagster_dbt/cloud/asset_defs.py
OwenKephart: hahaha ok fair enough

PR #11753, opened 2023-01-17
Include retry count in the job created by k8s_job_op

gibsondan:
Summary:
Right now if you have an op with a retry policy that calls execute_k8s_job, it will fail because it tries to create teh same job twice. Include the retry count in the job name so that it doesn't run into duplicate namespace issues.

Separately, we should change these cryptic names to be less cryptic (but not as part of this PR).

Test Plan: BK

### DIFF
--- a/python_modules/libraries/dagster-k8s/dagster_k8s/ops/k8s_job_op.py
+++ b/python_modules/libraries/dagster-k8s/dagster_k8s/ops/k8s_job_op.py
@@ -254,6 +254,10 @@ def execute_k8s_job(
 
     job_name = get_k8s_job_name(context.run_id, context.op.name)
 
+    retry_number = context.retry_number
+    if retry_number > 0:
+        job_name = f"{job_name}-{retry_number}"
+
     job = construct_dagster_k8s_job(
         job_config=k8s_job_config,
         args=args,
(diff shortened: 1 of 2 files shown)


python_modules/libraries/dagster-k8s/dagster_k8s/ops/k8s_job_op.py
```diff
@@ -254,6 +254,10 @@ def execute_k8s_job(
 
     job_name = get_k8s_job_name(context.run_id, context.op.name)
 
+    retry_number = context.retry_number
+    if retry_number > 0:
+        job_name = f"{job_name}-{retry_number}"
```
alangenfeld, line 259: does `get_k8s_job_name` truncate down to the limit such that adding `-N` would exceed or does it leave some buffer?

alangenfeld (approved):

python_modules/libraries/dagster-k8s/dagster_k8s/ops/k8s_job_op.py
gibsondan: the old name was always 33 characters (one of the reasons the name is so bad is that it just hexdigests it) - so as long as we don't have ~10^20 retries on a single op we're good

PR #11755, opened 2023-01-17
[asset-reconciliation] Improve performance for never materialized root partitions

OwenKephart:
### Summary & Motivation

A followup from the discussion here: https://github.com/dagster-io/dagster/pull/11582. This dramatically improves performance.

#### Test Setup

A 30-asset multi-asset, downstream of two 15-minute partitioned assets (start-date 2022-01-01). 

#### Benchmarks

I originally tested this before Claire's performance improvements to TimeWindowPartitionsSubset were merged in. In that state, it took 114 seconds (!!) to run my test, the majority of it spent iterating over time windows.

Claire's changes made a big difference, cutting down the total time to 28 seconds, although the performance scaled pretty poorly as the number of source assets increased (i.e. if you had 5 root assets instead of 2, it'd go back to taking over a minute).

After these changes, the original test completes in under 4 seconds, and with 100 source assets, the total time was around 17 seconds, with only around 25% of that time spent within this codepath.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -27,7 +27,6 @@ from dagster._core.definitions.freshness_policy import FreshnessConstraint
 from dagster._core.definitions.time_window_partitions import (
     TimeWindow,
     TimeWindowPartitionsDefinition,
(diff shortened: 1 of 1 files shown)


sryza (approved):
Nice! Seems like this should be merged in conjunction with https://github.com/dagster-io/dagster/pull/11582.

PR #11758, opened 2023-01-17
[OSS Nux] Add kill switch + persist Nux dismissal in temp folder

salazarm:
### Summary & Motivation

Couple of changes:

1. Only show the Nux if you're on localhost
2. Persist that we dismissed the nux in the directory we use for telemetry too
3. Add killswitch to completely disable this

For the kill switch in dagser.yaml we need to add
```
nux:
  enabled: False
```

Unit tests for the graphql

Made sure the file was persisted and stayed past restarts but wasn't sure how to effectively test this in python.

Screen Shot 2023-01-17 at 5 58 23 PM

### DIFF
--- a/js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
+++ b/js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
@@ -17,14 +17,15 @@ import React from 'react';
 import isEmail from 'validator/lib/isEmail';
 
 export const CommunityNux = () => {
-  const [didDismissCommunityNux, dismissCommunityNux] = useStateWithStorage(
+  const [didDismissCommunityNux, dissmissInBrowser] = useStateWithStorage(
     'communityNux',
     (data) => data,
   );
   const {data, loading} = useQuery(GET_SHOULD_SHOW_NUX_QUERY);
-  const [dismiss] = useMutation(SET_NUX_SEEN_MUTATION);
+  const [dismissOnServer] = useMutation(SET_NUX_SEEN_MUTATION);
 
   if (!isLocalhost()) {
+    // Yes, we only want to show this on localhost for now.
     return null;
   }
   if (didDismissCommunityNux || loading || (data && !data.shouldShowNux)) {
@@ -33,8 +34,8 @@ export const CommunityNux = () => {
   return (
     <CommunityNuxImpl
       dismiss={() => {
-        dismissCommunityNux('1');
-        dismiss();
+        dissmissInBrowser('1');
+        dismissOnServer();
       }}
     />
   );
--- a/js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
+++ b/js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
@@ -24,7 +24,7 @@ export const CommunityNux = () => {
   const {data, loading} = useQuery(GET_SHOULD_SHOW_NUX_QUERY);
   const [dismiss] = useMutation(SET_NUX_SEEN_MUTATION);
 
-  if (!window.location.origin.startsWith('http://localhost')) {
+  if (!isLocalhost()) {
     return null;
   }
   if (didDismissCommunityNux || loading || (data && !data.shouldShowNux)) {
@@ -232,3 +232,8 @@ const GET_SHOULD_SHOW_NUX_QUERY = gql`
     shouldShowNux
   }
 `;
+
+function isLocalhost() {
+  const origin = window.location.origin;
+  return origin.startsWith('http://127.0.0.1') || origin.startsWith('http://localhost');
+}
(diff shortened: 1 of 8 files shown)


python_modules/dagster/dagster/_core/nux.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+# Sets whether we've shown the Nux to any user on this instance
+def set_nux_seen():
+    try:  # In case we encounter an error while writing to user's file system
+        with open(nux_seen_filepath(), "w", encoding="utf8") as nux_seen_file:
+            # the contents of the file dont matter, we just check that it exists, but lets write seen: True here anyways.
+            yaml.dump({"seen": 1}, nux_seen_file, default_flow_style=False)
+    except Exception:
+        return "<<unable_to_write_instance_id>>"
```
gibsondan: this string seems not right

python_modules/dagster/dagster/_core/nux.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+# Gets whether we've shown the Nux to any user on this instance
+def get_has_seen_nux():
+    # We only care about the existence of the file
+    if not os.path.exists(nux_seen_filepath()):
+        return False
+    return True
```
gibsondan, line 30: return not os.path.exists(...)

js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
```diff
@@ -212,3 +220,20 @@ const RecaptchaIFrame: React.FC<{
...
+
+const GET_SHOULD_SHOW_NUX_QUERY = gql`
+  query ShouldShowNux {
+    shouldShowNux
+  }
+`;
+
+function isLocalhost() {
+  const origin = window.location.origin;
+  return origin.startsWith('http://127.0.0.1') || origin.startsWith('http://localhost');
```
gibsondan, line 238: i assume people using https here is not a thing

js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
```diff
@@ -20,13 +21,20 @@ export const CommunityNux = () => {
     'communityNux',
     (data) => data,
   );
-  if (didDismissCommunityNux) {
+  const {data, loading} = useQuery(GET_SHOULD_SHOW_NUX_QUERY);
+  const [dismiss] = useMutation(SET_NUX_SEEN_MUTATION);
+
+  if (!isLocalhost()) {
+    return null;
+  }
```
gibsondan, line 29: maybe worth a comment here since it might be the opposite of what people expect (only showing it on local dev)

js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
```diff
@@ -20,13 +21,20 @@ export const CommunityNux = () => {
...
+  if (didDismissCommunityNux || loading || (data && !data.shouldShowNux)) {
     return null;
   }
   return (
     <CommunityNuxImpl
       dismiss={() => {
         dismissCommunityNux('1');
+        dismiss();
```
gibsondan, line 37: names here are a bit confusing - this is dismissInBrowser and dismissOnServer maybe?

Do we still need the browser one once we have the server one and its only on for localhost? Certainly can't hurt...

gibsondan (approved):

js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
salazarm: Yeah we don't really need the browser one... but yeah figured it can't hurt

js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
salazarm: Screen Shot 2023-01-17 at 6 26 05 PM

Yea

js_modules/dagit/packages/app/src/NUX/CommunityNux.tsx
```diff
@@ -16,17 +17,25 @@ import React from 'react';
...
 
 export const CommunityNux = () => {
-  const [didDismissCommunityNux, dismissCommunityNux] = useStateWithStorage(
+  const [didDismissCommunityNux, dissmissInBrowser] = useStateWithStorage(
     'communityNux',
     (data) => data,
   );
-  if (didDismissCommunityNux) {
+  const {data, loading} = useQuery(GET_SHOULD_SHOW_NUX_QUERY);
+  const [dismissOnServer] = useMutation(SET_NUX_SEEN_MUTATION);
```
hellendag, line 25: These should be updated to use type annotations again. The lint rule might not be working right now.

salazarm: Interestingly... the codegen didn't generate any types in the folder... Could it be because there are no arguments or returned data from the mutation? And because the query just returns a primitive directly rather than an object?

PR #11759, opened 2023-01-17
add after_cursor to get_materialization_counts_by_partition

OwenKephart:
### Summary & Motivation

This is very useful in the reconciliation world, where we often want to produce a record of all things that have changed since a given cursor. Currently, this requires querying for all events since that cursor, which requires deserializing a potentially very large number of events.

In addition, the fact that this function already supports querying for multiple asset keys at once means that we can batch up a lot of work into a single DB query, rather than having to do N separate `get_event_log_records` queries (one for each asset).

unit

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/legacy_storage.py
+++ b/python_modules/dagster/dagster/_core/storage/legacy_storage.py
@@ -463,9 +463,11 @@ class LegacyEventLogStorage(EventLogStorage, ConfigurableClass):
         return self._storage.event_log_storage.wipe_asset(asset_key)
 
     def get_materialization_count_by_partition(
-        self, asset_keys: Sequence["AssetKey"]
+        self, asset_keys: Sequence["AssetKey"], after_cursor: Optional[int] = None
     ) -> Mapping["AssetKey", Mapping[str, int]]:
-        return self._storage.event_log_storage.get_materialization_count_by_partition(asset_keys)
+        return self._storage.event_log_storage.get_materialization_count_by_partition(
+            asset_keys, after_cursor
+        )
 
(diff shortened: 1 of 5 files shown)


python_modules/dagster/dagster/_core/storage/legacy_storage.py
```diff
@@ -463,9 +463,11 @@ def wipe_asset(self, asset_key: "AssetKey"):
         return self._storage.event_log_storage.wipe_asset(asset_key)
 
     def get_materialization_count_by_partition(
-        self, asset_keys: Sequence["AssetKey"]
+        self, asset_keys: Sequence["AssetKey"], after_cursor: Optional[int] = None
```
prha, line 466: Consider if we should just apply a generic `EventRecordsFilter` in here... might simplify the signature a bit, with the tradeoff that it could be confusing if some conflicting args were applied (e.g. non materialization event types).

prha (approved):

python_modules/dagster/dagster/_core/storage/legacy_storage.py
OwenKephart: I did think on this for awhile, as there really are plenty of useful fields on the EventRecordsFilter that you could hypothetically want to pass in here (`partition_keys`, `before_cursor`, and `tags` could all be useful in theory)  but I think the confusion of conflicting args might be too much -- EventRecordsFilter doesn't (yet?) support an `asset_keys` argument (just `asset_key`), and so there'd be two gotchas (you can't set the `asset_key` argument, and you need to set the `event_type` to `DagsterEventType.ASSET_MATERIALIZATION`).

PR #11764, opened 2023-01-18
[db io managers] create schema if not exist

jamiedemaria:
### Summary & Motivation
Adds new method to DB IO manager subclasses to create the schema if it doesn't exist. Calls this method within handle_output

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/db_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/db_io_manager.py
@@ -76,7 +76,7 @@ class DbClient:
         ...
 
     @staticmethod
-    def create_schema(context: OutputContext, table_slice: TableSlice) -> None:
+    def ensure_schema_exists(context: OutputContext, table_slice: TableSlice) -> None:
         ...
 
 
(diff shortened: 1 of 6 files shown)


jamiedemaria:
* master
  * **PR #11704** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11704" target="_blank">Graphite</a> 
    * **PR #11705** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11705" target="_blank">Graphite</a> 
      * **PR #11764** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11764" target="_blank">Graphite</a>  👈
        * **PR #11765** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11765" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11764?utm_source=stack-comment).

sryza (approved):

sryza:
just approved this, but, actually - should this be something we do inside the snowflake IO manager instead of specifically in the pandas type handler?

jamiedemaria:
@sryza  did a refactor to move schema creation to the db io manager. it means making the db connection a bunch of times now so i think a refactor could be in order. i'm going to try that in another pr

python_modules/dagster/dagster/_core/storage/db_io_manager.py
```diff
@@ -67,6 +67,10 @@ def delete_table_slice(context: OutputContext, table_slice: TableSlice) -> None:
     def get_select_statement(table_slice: TableSlice) -> str:
         ...
 
+    @staticmethod
+    def create_schema(context: OutputContext, table_slice: TableSlice) -> None:
```
sryza, line 71: Maybe we should call this something like `ensure_schema_exists`? Because we won't always be creating it?

PR #11767, opened 2023-01-18
Remove rosetta from contributing docs

gibsondan:
Summary:
No longer needed, particularly now that we supply a grpcio wheel.


schrockn (changes requested):
I think we should note the existence of the m1-specific wheel in case people run into that. Is this also the natural place to document the env var opt out?

gibsondan:
Added a little note in contributing as well as installing

gibsondan:
will be back with an option that incorporates plan B

PR #11770, opened 2023-01-18
[asset-reconciliation] pre-fetch the results of some queries

OwenKephart:
### Summary & Motivation

The two biggest time sucks in terms of querying are:

a) getting the latest materialization record per-asset
b) getting the AssetRecords per asset (slightly different from the above, these are the things containing AssetEntries, which are used to determine if a previous run failed to materialize a given asset)

This does a best-effort approach to batching together as much of that work into a couple of database queries, with some pretty large performance improvements. In particular, for a test case I cooked up, this took the execution time down from 75 seconds to 8.5(!!) seconds.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -21,6 +21,7 @@ import toposort
 import dagster._check as check
 from dagster._core.errors import DagsterInvalidInvocationError, DagsterInvariantViolationError
 from dagster._core.selector.subset_selector import DependencyGraph, generate_asset_dep_graph
+from dagster._utils.cached_method import cached_method
 
 from .assets import AssetsDefinition
 from .events import AssetKey, AssetKeyPartitionKey
@@ -311,6 +312,7 @@ class AssetGraph:
             return self._required_multi_asset_sets_by_key[asset_key]
         return set()
 
+    @cached_method
     def toposort_asset_keys(self) -> Sequence[AbstractSet[AssetKey]]:
         return [
             {key for key in level} for level in toposort.toposort(self._asset_dep_graph["upstream"])
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -223,9 +223,14 @@ def find_parent_materialized_asset_partitions(
     target_asset_keys = target_asset_selection.resolve(asset_graph)
 
     for asset_key in target_asset_selection.upstream(depth=1).resolve(asset_graph):
-        records = instance_queryer.get_materialization_records(
-            asset_key=asset_key, after_cursor=latest_storage_id
-        )
+        # if the asset is non-partitioned, we only need the latest materialization
+        if not asset_graph.is_partitioned(asset_key):
+            latest_record = instance_queryer.get_latest_materialization_record(asset_key)
+            records = [latest_record] if latest_record else []
+        else:
+            records = instance_queryer.get_materialization_records(
+                asset_key=asset_key, after_cursor=latest_storage_id
+            )
         for record in records:
             for child in asset_graph.get_children_partitions(asset_key, record.partition_key):
                 if child.asset_key in target_asset_keys and not instance_queryer.is_asset_in_run(
@@ -611,7 +616,7 @@ def determine_asset_partitions_to_reconcile_for_freshness(
 
             # the set of asset keys which are involved in some constraint and are actually upstream
             # of this asset
-            relevant_upstream_keys = frozenset(
+            relevant_upstream_keys = set(
                 set().union(
                     *(
                         expected_data_times_by_key[parent_key].keys()
@@ -699,6 +704,9 @@ def reconcile(
 ):
     instance_queryer = CachingInstanceQueryer(instance=instance)
     asset_graph = repository_def.asset_graph
+
+    # fetch some data in advance to batch together some queries
+    instance_queryer.prefetch_for_keys(list(asset_selection.resolve(asset_graph)))
 
     (
         asset_partitions_to_reconcile_for_freshness,
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -222,24 +222,32 @@ def find_parent_materialized_asset_partitions(
 
     target_asset_keys = target_asset_selection.resolve(asset_graph)
 
-    for asset_key in target_asset_selection.upstream(depth=1).resolve(asset_graph):
-        # if the asset is non-partitioned, we only need the latest materialization
-        if not asset_graph.is_partitioned(asset_key):
-            latest_record = instance_queryer.get_latest_materialization_record(asset_key)
-            records = [latest_record] if latest_record else []
-        else:
-            records = instance_queryer.get_materialization_records(
-                asset_key=asset_key, after_cursor=latest_storage_id
-            )
-        for record in records:
-            for child in asset_graph.get_children_partitions(asset_key, record.partition_key):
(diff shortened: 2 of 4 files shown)


python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -311,6 +312,7 @@ def get_required_multi_asset_keys(self, asset_key: AssetKey) -> AbstractSet[Asse
             return self._required_multi_asset_sets_by_key[asset_key]
         return set()
 
+    @cached_method
```
OwenKephart, line 315: for very big asset graphs, toposorting can take a few seconds, figured there wasn't much harm in caching this as we do call this method a few times

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -224,9 +224,14 @@ def find_parent_materialized_asset_partitions(
...
 
     for asset_key in target_asset_selection.upstream(depth=1).resolve(asset_graph):
-        records = instance_queryer.get_materialization_records(
-            asset_key=asset_key, after_cursor=latest_storage_id
-        )
+        # if the asset is non-partitioned, we only need the latest materialization
+        if not asset_graph.is_partitioned(asset_key):
+            latest_record = instance_queryer.get_latest_materialization_record(asset_key)
+            records = [latest_record] if latest_record else []
+        else:
```
OwenKephart, line 231: in a future PR (post-after_cursor being added to the materialization_count_by_partition function), this branch will get more efficient.

for now, if this is your first time turning on the sensor,  and you have partitioned assets, this will require querying for all historical materialization events for those assets (super expensive if you have a big history!).

however, at least for the non-partitioned assets, we'll only grab the first one, which provides a big performance boost for that (more common) case

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -612,7 +617,7 @@ def determine_asset_partitions_to_reconcile_for_freshness(
 
             # the set of asset keys which are involved in some constraint and are actually upstream
             # of this asset
-            relevant_upstream_keys = frozenset(
+            relevant_upstream_keys = set(
```
OwenKephart, line 620: frozenset gets surprisingly expensive, as hashing requires sorting the elements! this was just a tiny change, but we use frozenset elsewhere, so I may end up seeing if it's possible to remove those other uses in a future PR.

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -132,6 +133,7 @@ def prepare_insert_event(self, event):
             partition=partition,
         )
 
+    @functools.lru_cache(maxsize=None)
```
OwenKephart, line 136: this is a kinda funny one -- we call this a bunch in the get_asset_records code, and it requires querying the database each time. This added up to around 10 seconds of work in one test!

I figured the result of this wouldn't change from call to call, so why not cache it? maxsize=None because there's not a lot of distinct inputs to this function and it's just returning a bool.

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -54,10 +55,44 @@ def __init__(self, instance: "DagsterInstance"):
...
+        """For performance, batches together queries for selected assets"""
+        asset_records = self.instance.get_asset_records(asset_keys)
+        for asset_record in asset_records:
+            self._asset_record_cache[asset_record.asset_entry.asset_key] = asset_record
+
+        # get the latest N materialization records. in most cases, this will hit a large portion of
+        # the latest materialization records per asset key, although a more sophisticated method
+        # of bucketing queries could be used here to guarantee that we actually get the exact set
+        # we're aiming for here.
+        latest_records = self._instance.get_event_records(
```
OwenKephart, line 74: this method of getting relevant event records is a bit suspicious, but my guess is that it'll work out ok in practice. even if all of this work goes to waste (i.e. none of the most recent N asset events are relevant), this is still pretty cheap. When N=2000 (in my test), this query took less than .5s, with the entire prefetch method taking ~1s.

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -150,8 +185,11 @@ def get_latest_materialization_record(
         else:
             asset_partition = asset
 
-        # fancy caching only applies to after_cursor
         if before_cursor is not None:
+            latest_record = self._latest_materialization_record_cache.get(asset_partition)
```
OwenKephart, line 189: If the before_cursor is greater than the storage id of the latest record for an asset, then we know that the most recent record before that cursor is the same as the latest one.

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -224,9 +224,14 @@ def find_parent_materialized_asset_partitions(
...
     for asset_key in target_asset_selection.upstream(depth=1).resolve(asset_graph):
-        records = instance_queryer.get_materialization_records(
-            asset_key=asset_key, after_cursor=latest_storage_id
-        )
+        # if the asset is non-partitioned, we only need the latest materialization
+        if not asset_graph.is_partitioned(asset_key):
+            latest_record = instance_queryer.get_latest_materialization_record(asset_key)
+            records = [latest_record] if latest_record else []
+        else:
+            records = instance_queryer.get_materialization_records(
```
clairelin135, line 232: I think `get_event_records` has an `asset_partitions` argument to further filter these results. Hopefully this also helps with speeding up the query

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
clairelin135: Can we test that this call will still function correctly if someone run a migration between calls to this method?

We could make another call here to test that `instance.can_cache_asset_status_data()` will return `True` after the migration runs: https://github.com/dagster-io/dagster/blob/a54d17030f4cb9aab20b5498dc23f31913636b5c/python_modules/libraries/dagster-postgres/dagster_postgres_tests/compat_tests/test_back_compat.py#L776

If this test fails, we might need to come up with caching within `get_asset_records`

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -54,10 +55,44 @@ def __init__(self, instance: "DagsterInstance"):
...
+        # we're aiming for here.
+        latest_records = self._instance.get_event_records(
+            EventRecordsFilter(
+                event_type=DagsterEventType.ASSET_MATERIALIZATION,
+            ),
+            limit=len(asset_keys),
+            ascending=False,
+        )
+        for record in latest_records:
+            asset_partition = AssetKeyPartitionKey(
```
clairelin135, line 82: I think we might need a check downstream to ensure that these partitions are still valid partition keys of the asset

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
OwenKephart: that's a good point, I'd imagine it would fail -- i'll update get_asset_records directly instead

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
OwenKephart: that's fair, and that definitely would help, but a future change will switch to using get_materialization_count_by_partition anyway, so I think I'll keep it this way in the short term

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
OwenKephart: we never actually iterate through the cache, so this shouldn't change the functionality

clairelin135 (approved):

PR #11773, opened 2023-01-18
[1/n][structured config] Add ability to runtime-configure struct-config resources

benpankow:
## Summary

Allows using structured-config based resources and IO managers without binding config at defintiions-time, e.g.:

```python

class MyResource(Resource):
    a_string: str

@asset
def my_asset(res: MyResource):
    print(res.a_string)


defs = Definitions(
    assets=[my_asset],
    resources={"res": MyResource.partial()}
)
```
Screen Shot 2023-01-09 at 12 40 57 PM

## Test Plan

Add unit tests for resources, IO managers.

### DIFF
--- a/python_modules/dagster/dagster/_config/structured_config.py
+++ b/python_modules/dagster/dagster/_config/structured_config.py
@@ -199,7 +199,7 @@ class Resource(
         return cast(T, self)
 
     @classmethod
-    def partial(cls, **kwargs) -> "PartialResource[Self]":
+    def configure_at_launch(cls, **kwargs) -> "PartialResource[Self]":
         """
         Returns a partially initialized copy of the resource, with remaining config fields
         set at runtime.
@@ -223,8 +223,6 @@ class PartialResource(Generic[T], ResourceDefinition, MakeConfigCacheable):
 
     def __init__(self, resource_cls: Type[Resource[T]], data: Dict[str, Any]):
         check.invariant(data == {}, "PartialResource currently does not support config fields")
-
-        Config.__init__(self, data=data, resource_cls=resource_cls)
 
         MakeConfigCacheable.__init__(self, data=data, resource_cls=resource_cls)
 
@@ -317,7 +315,7 @@ class StructuredConfigIOManagerBase(Resource[IOManager], IOManagerDefinition):
         raise NotImplementedError()
 
     @classmethod
-    def partial(cls, **kwargs) -> "PartialIOManager":
+    def configure_at_launch(cls, **kwargs) -> "PartialIOManager":
         """
         Returns a partially initialized copy of the resource, with remaining config fields
         set at runtime.
--- a/python_modules/dagster/dagster/_config/structured_config.py
+++ b/python_modules/dagster/dagster/_config/structured_config.py
@@ -42,7 +42,7 @@ from dagster._config.field_utils import (
 from dagster._core.definitions.resource_definition import ResourceDefinition, ResourceFunction
 from dagster._core.storage.io_manager import IOManager, IOManagerDefinition
 
-Self = Any
+Self = TypeVar("Self", bound="Resource")
 
 
 def _safe_is_subclass(cls: Any, possible_parent_cls: Type) -> bool:
@@ -206,15 +206,23 @@ class Resource(
         """
         return PartialResource(cls, data=kwargs)
 
-    # Python descriptor
+    # The following methods are used to implement the descriptor protocol
     # https://docs.python.org/3/howto/descriptor.html
+    #
     # Used to adjust the types of resource inputs and outputs, e.g. resource dependencies can be passed in
     # as PartialResources or Resources, but will always be returned as Resources
-    def __get__(self, obj: "Resource", __owner: Any) -> Self:
-        ...
+    # Very similar to https://github.com/pydantic/pydantic/discussions/4262
 
-    def __set__(self, obj: Optional[object], value: Union[T, "PartialResource[T]"]) -> None:
-        ...
+    def __set_name__(self, _owner, name):
+        self._assigned_name = name
+
+    def __get__(self: Self, obj: Any, __owner: Any) -> Self:
+        return cast(Self, getattr(obj, self._assigned_name))
+
+    def __set__(
+        self: Self, obj: Optional[object], value: Union[Self, "PartialResource[Self]"]
+    ) -> None:
+        setattr(obj, self._assigned_name, value)
 
 
 class PartialResource(Generic[T], ResourceDefinition, MakeConfigCacheable):
--- a/python_modules/dagster/dagster/_config/structured_config.py
+++ b/python_modules/dagster/dagster/_config/structured_config.py
@@ -152,11 +152,11 @@ def _curry_config_schema(schema_field: Field, data: Any) -> IDefinitionConfigSch
     return DefinitionConfigSchema(_apply_defaults_to_schema_field(schema_field, data))
 
 
-T = TypeVar("T")
+ResValue = TypeVar("ResValue")
 
 
 class Resource(
-    Generic[T],
+    Generic[ResValue],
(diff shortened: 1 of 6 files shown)


benpankow:
* master
  * **PR #11773** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11773" target="_blank">Graphite</a>  👈
    * **PR #11645** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11645" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11773?utm_source=stack-comment).

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -132,7 +156,55 @@ def create_object_to_pass_to_user_code(self, context) -> Any:  # pylint: disable
...
+        return PartialResource(cls, data=kwargs)
+
+    # Python descriptor
+    # https://docs.python.org/3/howto/descriptor.html
+    # Used to adjust the types of resource inputs and outputs, e.g. resource dependencies can be passed in
+    # as PartialResources or Resources, but will always be returned as Resources
+    def __get__(self, obj: "Resource", __owner: Any) -> Self:
+        ...
+
+    def __set__(self, obj: Optional[object], value: Union[T, "PartialResource[T]"]) -> None:
```
schrockn, line 176: I'd like some additional explanation here. I'm not following exactly how this works.

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -132,7 +156,55 @@ def create_object_to_pass_to_user_code(self, context) -> Any:  # pylint: disable
...
+
+
+class PartialResource(Generic[T], ResourceDefinition, MakeConfigCacheable):
+    data: Dict[str, Any]
+    resource_cls: Type[Resource[T]]
+
+    def __init__(self, resource_cls: Type[Resource[T]], data: Dict[str, Any]):
+        check.invariant(data == {}, "PartialResource currently does not support config fields")
+
+        Config.__init__(self, data=data, resource_cls=resource_cls)
```
schrockn, line 187: It seems very strange to directly call the `__init__` for a class two notches up in the inheritance chain

schrockn (changes requested):
See two line questions.

Also should we call this `configure_at_launch` and only add partials when we actually support partials?

python_modules/dagster/dagster/_config/structured_config.py
benpankow: A Python class implementing the [descriptor protocol](https://realpython.com/python-descriptors) (the `__get__` and `__set__` methods here) has custom behavior when attributes of that type are get or set. In this case, the `__get__` and `__set__` implementations determine what happens when a `Resource`-typed attribute of another class is get or set.

Our implementations themselves don't do anything interesting, but the type annotations on the methods determine the type hints affiliated with the data class constructor + attribute accesses:

```python
class MyClass(Resource):
    inner: Resource

# Type determined by __set__ param type hint, in this case Union[T, PartialResource[T]]
foo = MyClass(inner=...)

# Type determined by __get__ return value type hint, in this case T
foo.inner
```

(This is the protocol used, for example, to implement @property - our use-case is a lot simpler since we only use it to suggest to mypy/Pyright the right type to use)

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -29,6 +33,22 @@ class cached_property:  # type: ignore[no-redef]
 from dagster._core.definitions.resource_definition import ResourceDefinition, ResourceFunction
 from dagster._core.storage.io_manager import IOManager, IOManagerDefinition
 
+Self = Any
```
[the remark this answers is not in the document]
benpankow: actually, we can use a typevar for this... never mind

python_modules/dagster/dagster/_config/structured_config.py
schrockn: So this has _zero) impact on runtime behavior at all and this is only for type-checking purposes? These are not invoked at all at runtime?

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -132,7 +156,61 @@ def create_object_to_pass_to_user_code(self, context) -> Any:  # pylint: disable
...
+    def __set_name__(self, _owner, name):
+        self._assigned_name = name
+
+    def __get__(self: Self, obj: Any, __owner: Any) -> Self:
+        return cast(Self, getattr(obj, self._assigned_name))
+
+    def __set__(
+        self: Self, obj: Optional[object], value: Union[Self, "PartialResource[Self]"]
+    ) -> None:
+        setattr(obj, self._assigned_name, value)
```
schrockn, line 185: If this is truly for type-checking only, I think this another case where putting this in a class and making it a mixin will be clearer and make this more "delete-able". So like a class called `AllowPartialResourceInitParams" or something.

schrockn (changes requested):
Just popping back in your queue for visibility. I want to make I understand what is going on completely. If that understanding is clear (that there is no runtime behavior change) let's plop that thing in its own class, mix it in, and overdocument/explain it.

python_modules/dagster/dagster/_config/structured_config.py
benpankow: No, not that I can see for this use-case - w/ print methods inserted in the various descriptor methods, none trigger:

```python
>>> from dagster._config.structured_config import Resource
>>> class FooResource(Resource):
...     a_str: str
...
>>> class MyClass():
...     res: Resource
...
>>> x = MyClass()
>>> x.res = FooResource(a_str="foo")
>>> print(x.res.a_str)
foo
>>> class MyOuterResource(Resource):
...     inner: Resource
...
>>> outer = MyOuterResource(inner=FooResource(a_str="foo"))
>>> print(outer.inner.a_str)
foo
```

They can be manually invoked though:
```python
>>> x.res.__set_name__(x, "res")
__set_name__ invoked
>>> x.res.__get__(x, x).a_str
__get__ invoked
'foo'
```

python_modules/dagster/dagster/_config/structured_config.py
benpankow: Updated to a standalone mixin class in separate file

python_modules/dagster/dagster/_config/structured_config/typing_utils.py
```diff
@@ -0,0 +1,49 @@
...
+    The following code will work:
+
+    .. code-block:: python
+
+        # Types as PartialResource[BarResource]
+        partial_bar = BarResource.configure_at_runtime()
+
+        # bar parameter takes BarResource | PartialResource[BarResource]
+        foo = FooResource(bar=partial_bar)
+
```
schrockn, line 31: One point of explanation here is that something has to happen between construction and the print state for `foo.bar` to return a `BarResource`. Some code has to go through and replace the partial with the full thing.

python_modules/dagster/dagster/_config/structured_config/typing_utils.py
```diff
@@ -0,0 +1,49 @@
...
+    from dagster._config.structured_config import PartialResource
+
+Self = TypeVar("Self", bound="AllowPartialResourceInitParams")
+
+
+class AllowPartialResourceInitParams:
+    """
+    Implementation of the Python descriptor protocol (https://docs.python.org/3/howto/descriptor.html)
+    to adjust the types of resource inputs and outputs, e.g. resource dependencies can be passed in
+    as PartialResources or Resources, but will always be returned as Resources.
```
schrockn, line 13: Please note here that this class exists _only_ to satisfy type-checking libraries. I would also rename the class to reflect that.

schrockn (approved):
Very cool. Please heed final comments.

PR #11777, opened 2023-01-18
Update docs to include wheel workarounds for m1 macs

gibsondan:
Summary:
Includes instructions for:
- installing on pip using our wheel
- installing on poetry using our wheel
- installing the dagster dev environment using our wheel

### DIFF
--- a/docs/content/community/contributing.mdx
+++ b/docs/content/community/contributing.mdx
@@ -7,58 +7,9 @@ description: Contributions are welcome and are greatly appreciated. Here's the c
 
 We love to see our community members get involved! If you are planning to contribute to Dagster, you will first need to set up a local development environment.
 
-## Environment Setup
+## Environment setup
 
-### Apple M1-specific setup
-
-If you have a Mac with an M1 chip, you will need to set up a separate environment to emulate an x86 architecture before following the remaining setup instructions. To find out if your Mac has an M1 chip, follow the instructions [here](https://www.howtogeek.com/706226/how-to-check-if-your-mac-is-using-an-intel-or-apple-silicon-processor/).
-
-If you don't have a Mac with an M1 chip, skip ahead to the next section - [Dagster development setup](#dagster-development-setup).
-
-1. Install rosetta. [About rosetta](https://developer.apple.com/documentation/apple-silicon/about-the-rosetta-translation-environment).
-
-   ```bash
-   softwareupdate --install-rosetta --agree-to-license
-   ```
-
-2. Create a duplicate terminal that will default to running the rosetta x86 emulator. [Source stack overflow post](https://apple.stackexchange.com/questions/428768/on-apple-m1-with-rosetta-how-to-open-entire-terminal-iterm-in-x86-64-architec).
-
-   - Go to `Finder` > `Applications` and find your terminal app.
-   - Right click the app and duplicate it.
-   - Right click the new terminal > `Get Info` > `Enable Open using Rosetta`.
-   - Click to open the terminal, type `arch` to verify it says `i386` now.
-
-3. In your rosetta terminal, install `homebrew`.
-
-   ```bash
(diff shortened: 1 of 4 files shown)


docs/content/community/contributing.mdx
```diff
@@ -7,58 +7,9 @@ description: Contributions are welcome and are greatly appreciated. Here's the c
...
-   export LDFLAGS="-L $(brew --prefix openssl)/lib"
-   export CFLAGS="-I $(brew --prefix openssl)/include"
-   ```
-
-After sourcing this file, you can ensure that the correct version of `homebrew` is being used by confirming the output of `which brew` is `/usr/local/bin/brew`.
-
-### Dagster development setup
-
-1. Install Python. Python 3.6 or above recommended, but our CI/CD pipeline currently tests against up-to-date patch versions of Python 3.6, 3.7 and 3.8.
+1. Install Python 3.6 or above.
```
alangenfeld, line 12: nit: 3.7 and above these days right

docs/content/getting-started/install.mdx
```diff
@@ -19,6 +19,12 @@ pip install dagster dagit
 
 This will install the latest stable version of the core Dagster packages in your current Python environment.
 
+**Note for Macs with an M1 or M2 chip**: Some users have reported installation problems due to missing wheels for arm64 Macs when installing the `grpcio` package. To install `dagster` using a pre-built wheel of the `grpcio` package for M1 and M2 machines, run:
+
+```bash
+pip install dagster dagit --find-links=https://github.com/dagster-io/build-grpcio/wiki/Wheels
```
schrockn, line 25: How does one encode this --find-links shenanigan in a setup.py or requirements.txt file? e.g. if someone downloads an example with a dagster dep, how do we make sure it gets installed with the proper wheels?

gibsondan: @schrockn it would be a CLI flag that you supply to the pip install in that case as well. There's not a way to encode it in the setup.py that we're aware of, doing that was Plan A but pypi didn't like it

shalabhc: Unfortunately PyPI seems to actively discourage/prevent silently using a wheel outside of PyPI. They used to have [dependency_links](https://setuptools.pypa.io/en/latest/deprecated/dependency_links.html) which did exactly this but has been removed in recent versions. Our attempt at just pointing directly to our wheel in setup.py also failed to publish.
I now assume their thinking is if you are using a non standard wheel repo, it should be specified by the user via --find-links or --extra-index-url.

smackesey: `setup.py` is for "abstract" dependencies and so it only supports version specifiers to my knowledge. `requirements.txt` on the other hand allows you to point to a specific "physical" package (the full spec is [requirement specifiers](https://pip.pypa.io/en/stable/reference/requirement-specifiers/#requirement-specifiers)), so you could do e.g. this:

```
### requirements.txt

grpcio @ https://github.com/dagster-io/build-grpcio/raw/main/wheels/grpcio-1.47.2-cp39-cp39-macosx_11_0_arm64.whl
```

shalabhc: @smackesey do you know if we can publish packages with `requirements.txt` like above to PyPI? `setup.py` supports [direct references](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#direct-url-dependencies) as well but PyPI [does not accept packages](https://github.com/dagster-io/dagster/pull/11760) that use them.

We want to use our wheel for m1 macs only and use the default for all other machines (intel macs, non macs). We also want a way to fallback to the old behavior (ie not use our wheel on m1s) if users already have `grpcio` installed from elsewhere. Doing all of these in `requirements.txt` seems hard.

shalabhc (approved):

PR #11778, opened 2023-01-18
[docs] [tutorial-overhaul] [1-3] Writing your pipeline with assets

tacastillo:
### Summary & Motivation

Here's the first actual tutorial page for the tutorial: writing your first asset. It goes over how the code that'll be used for reference, eases into the refactor from the script into writing your first asset, and then materializing it within Dagit.

[Link to the page](https://dagster-git-docs-tutorialbasicbuilding-assets-elementl.vercel.app/new-tutorial/your-first-asset)


docs/content/new-tutorial/your-first-asset.mdx
```diff
@@ -0,0 +1,187 @@
...
+By the end of this section, you will:
+
+- ingest data from the Hacker News API
+- turn that ingestion into an asset
+- explore the Dagit, the Dagster UI
+- launch your first Dagster run manually
+
+## Understanding the pipeline
+
+Hypothetically, you have a Python script called `analyze_hackernews.py`. The code for it is below. If you'd like to see this code work, you can run:
```
sryza, line 22: A few thoughts on this Python script approach:
- I think it needs a little more explanation / framing.  We should briefly motivate why we're showing them a Python script instead of jumping straight into Dagster.
- The script does not actually generate any assets - i.e. persistent objects in storage.  I think we need to add some `pickle.dumps` calls or similar to make it analogous to what we end up building in Dagster.
- The script is long and it's a lot to digest. What do you think about first just showing the part that generates the results DataFrame and adding the other part in a later section?

docs/content/new-tutorial/your-first-asset.mdx
```diff
@@ -0,0 +1,187 @@
...
+src="/images/tutorial/dagit-default.png"
+width={1392}
+height={1167}
+/>
+
+You can see that Dagit has detected your `hackernews_topstory_ids` asset. The UI isn't very useful yet because the asset hasn't been materialized yet. Let's do that next.
+
+---
+
+## Materializing your asset
```
sryza, line 148: I think it's important to communicate that Dagit isn't the only way to materialize your asset. Part of what's so cool about Dagster is that it has really lightweight ways of running stuff - e.g. on the CLI and with the Python APIs.

docs/content/new-tutorial/your-first-asset.mdx
```diff
@@ -0,0 +1,187 @@
...
+    return top_100_newstories
+```
+
+That's all it takes to get started. Dagster now knows that this is an asset. In future sections, you'll see how you can add more metadata, schedule when to refresh the asset, and more.
+
+And now you're done! Time to go into the Dagster UI and see what you've built.
+
+### Exploring the Dagster UI
+
+Dagster has a UI called Dagit. It's a great way to explore your data pipelines and see what's going on.
```
sryza, line 125: I think we can be more specific than "what's going on".  Maybe something like "explore your data assets, launch runs manually, observe what happened."

docs/content/new-tutorial/your-first-asset.mdx
```diff
@@ -0,0 +1,187 @@
+---
+title: "Intro to the Dagster Tutorial: Your First Asset | Dagster Docs"
+description: "Get set up for the intro to assets tutorial."
+noindex: True
+---
+
+# Writing Your Pipeline with Assets
+
+Now that you've installed Dagster and know what assets are, it's time for you to build parts of your first data pipeline!
```
sryza, line 9:
```suggestion
Now that you've installed Dagster and know what assets are, it's time for you to build your first data pipeline!
```

sryza (commented):
good stuff!

tacastillo:
Closing this for the combined #12509

PR #11782, opened 2023-01-18
Performance improvements for large multi assets

OwenKephart:
### Summary & Motivation

generate_asset_dep_graph for a multi asset with 5000 nodes took 30 *minutes* (!!)

This is because we were passing in 5000 copies of the same multi-asset when constructing our asset dependency graph. I also noticed that generating those error messages (regardless of if the error is relevant or not) took a significant amount of time for a graph of that size (around 20 seconds). It seems very unhelpful to list more than 20 or so keys, so I put that limit there, which brings the time back down to a couple of seconds.


sryza (approved):
Wow!

PR #11788, opened 2023-01-18
Fix multidimensional partition backfills

clairelin135:
This PR rolls up a couple of fixes into one.

- Currently, multidimensional asset backfills do not update partition status in Dagit. This is because recent changes to backfill logic did not add the corresponding per-dimension tags to the requested runs. This PR fixes this.
- The run status column of the backfill page displays as empty. This is because:
  - Multidimensional asset runs do not write a partition to the runs table
  - Minor JS bug (fix enclosed in PR)

This PR amends multidimensionally partitioned runs to also have the `partition` tag, which is the multidimensional partition key e.g. `a|2022-01-01`, subsequently writing this tag to the partition column of the runs table. Also updates the asset backfill logic to also add the multipartitions tags to asset backfill runs so multidimensional runs also contain per-dimension tags.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
@@ -293,9 +293,9 @@ class MultiPartitionsDefinition(PartitionsDefinition):
     def secondary_dimension(self) -> PartitionDimensionDefinition:
         return self._get_primary_and_secondary_dimension()[1]
 
-    def get_tags_from_partition_key(self, partition_key: str) -> Mapping[str, str]:
+    def get_tags_for_partition_key(self, partition_key: str) -> Mapping[str, str]:
         partition_key = cast(MultiPartitionKey, self.get_partition_key_from_str(partition_key))
-        tags = {**super().get_tags_from_partition_key(partition_key)}
+        tags = {**super().get_tags_for_partition_key(partition_key)}
         tags.update(get_tags_from_multi_partition_key(partition_key))
         return tags
 
--- a/python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
@@ -293,6 +293,12 @@ class MultiPartitionsDefinition(PartitionsDefinition):
     def secondary_dimension(self) -> PartitionDimensionDefinition:
         return self._get_primary_and_secondary_dimension()[1]
 
+    def get_tags_from_partition_key(self, partition_key: str) -> Mapping[str, str]:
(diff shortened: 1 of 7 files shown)


python_modules/dagster/dagster/_core/storage/pipeline_run.py
```diff
@@ -524,20 +529,40 @@ def tags_for_backfill_id(backfill_id: str) -> Mapping[str, str]:
...
+    )
+
+    tags = {PARTITION_NAME_TAG: partition_key}
+    if isinstance(partition_key, MultiPartitionKey):
+        tags.update(get_tags_from_multi_partition_key(partition_key))
+
+    return tags
+
+
+def get_partition_tags_from_partition_def(
```
sryza, line 550:
```suggestion
def get_partition_tags_from_partitions_def(
```

python_modules/dagster/dagster/_core/storage/pipeline_run.py
```diff
@@ -524,20 +529,40 @@ def tags_for_backfill_id(backfill_id: str) -> Mapping[str, str]:
...
-        if isinstance(partition.name, MultiPartitionKey):
-            tags.update(get_tags_from_multi_partition_key(partition.name))
-        else:
-            tags[PARTITION_NAME_TAG] = partition.name
-
+        tags.update(get_tags_from_partition_key(partition.name))
         return tags
 
 
+def get_tags_from_partition_key(partition_key: str) -> Mapping[str, str]:
```
sryza, line 537: What do you think about making this a method of `PartitionsDefinition` that gets overriden by `MultiPartitionsDefinition`?

python_modules/dagster/dagster/_core/storage/pipeline_run.py
clairelin135: 👍 👍

python_modules/dagster/dagster/_core/storage/pipeline_run.py
clairelin135: sounds good, good call!

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
```diff
@@ -293,6 +293,12 @@ def primary_dimension(self) -> PartitionDimensionDefinition:
     def secondary_dimension(self) -> PartitionDimensionDefinition:
         return self._get_primary_and_secondary_dimension()[1]
 
+    def get_tags_from_partition_key(self, partition_key: str) -> Mapping[str, str]:
```
sryza, line 296: Super nitpick: I'd call this "get_tags_for_partition_key", because the partition key doesn't contain the tags, we're just generating the corresponding set of tags.

sryza (approved):
Nice

PR #11795, opened 2023-01-19
Allow bare executor in Definitions

schrockn:
### Summary & Motivation

I found this very useful in a separate branch I was working on, and this is much more ergonomic in cases where all executor config is bound at definition time.

I also did a minor refactor to use opt* variants of check calls and do coercions closer to those calls because I think co-locating that makes it clearer.

BK.


schrockn:
* master
  * **PR #11795** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11795" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11795?utm_source=stack-comment).

sryza (approved):

PR #11800, opened 2023-01-19
changelog 1.1.11 pt 2

alangenfeld:


CHANGES.md
```diff
@@ -4,18 +4,26 @@
...
+nux:
+  enabled: false
+```
+
 - [dagit] Improved query performance on Scheduled Runs page.
 - [dagit] Added or improved document (browser tab) titles throughout Dagit.
+- [dagster-snowflake] Snowflake resource and IO manager now support private key authentication with unencrypted keys
 - [dagster-dbt] The op created when using `load_assets_from_dbt*` is now configurable, allowing you to pass in specific parameters to the underlying dbt command (such as `--full-refresh`). To do so, supply config of the form `{"ops": {"run_dbt_<abcde>": {"config": {"full_refresh": True}}}}` when defining your asset job, or in Dagit.
-- A `dagster dev` which spins up both dagit and the daemon, and requires no `$DAGSTER_HOME` env var to be set
-- `create_repository_using_definitions_args` has been added for users to backport their repository definitions to the new `Definitions` API
```
rexledesma, line 21: ### Experimental

- [dagster-dbt] Added a utility to cache compilations from dbt Cloud jobs, allowing software-defined assets to be loaded faster from dbt Cloud jobs.

rexledesma (approved):
probably should remove the formating fixes on the previous entries until we invest in a markdown linter in bk

PR #11810, opened 2023-01-19
[dagit] Fix support for long descriptions in the asset catalog table

bengotow:
### Summary & Motivation

This PR addresses an issue in the latest release related to my previous work in #10453.

Asset descriptions should appear on a line beneath the asset name and the kind tag, and should wrap accordingly. Here's the behavior after this fix:

image

image

To make this easier for us to catch in the future, I've also added a long description (containing markdown, another largely untested-in-toys feature) to one of the weather assets in our toys repo:

image
image

This PR also tweaks a Firefox constant we have in our middle truncation code, because I noticed one scenario where the DOM and Canvas reported widths of a string were off by 0.12, which caused the text to middle truncate unnecessarily.

Tested the asset catalog changes in Firefox and Chrome

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/AssetNodeList.tsx
+++ b/js_modules/dagit/packages/core/src/assets/AssetNodeList.tsx
@@ -53,6 +53,6 @@ const Container = styled(Box)`
 
 const AssetNodeWrapper = styled.div`
   cursor: pointer;
-  width: 240px;
+  width: 260px;
   flex-shrink: 0;
 `;
--- a/js_modules/dagit/packages/core/src/workspace/VirtualizedAssetRow.tsx
+++ b/js_modules/dagit/packages/core/src/workspace/VirtualizedAssetRow.tsx
@@ -107,8 +107,8 @@ export const VirtualizedAssetRow = (props: AssetRowProps) => {
           </RowCell>
         ) : null}
         <RowCell>
-          <Box flex={{justifyContent: 'space-between'}} style={{position: 'relative'}}>
-            <div style={{maxWidth: '100%', minWidth: 0}}>
+          <Box flex={{alignItems: 'center'}}>
+            <div style={{flex: 1, minWidth: 0}}>
               <AssetLink
                 path={type === 'folder' || view === 'directory' ? path.slice(-1) : path}
                 url={linkUrl}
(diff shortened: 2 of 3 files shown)


js_modules/dagit/packages/core/src/assets/AssetNodeList.tsx
```diff
@@ -53,6 +53,6 @@ const Container = styled(Box)`
 
 const AssetNodeWrapper = styled.div`
   cursor: pointer;
-  width: 240px;
+  width: 260px;
```
bengotow, line 56: This allows room for dates displaying the previous year here: 

image

js_modules/dagit/packages/ui/src/components/calculateMiddleTruncation.tsx
```diff
@@ -1,7 +1,7 @@
 // We've observed that Firefox's DOM APIs and Canvas APIs do not return
 // identical values given the same rendered text, in particular when the DOM
 // element is inside a flexbox. They're floating point numbers off by ~0.05.
-const FIREFOX_WIDTH_VARIANCE = 0.1;
+const FIREFOX_WIDTH_VARIANCE = 0.2;
```
alangenfeld, line 4: you high confidence on this?

js_modules/dagit/packages/core/src/assets/AssetNodeList.tsx
salazarm: I wonder if instead of the exact date we just show like "1 day ago" or something, and the on hover we show the full date? Might save us some space here

js_modules/dagit/packages/ui/src/components/calculateMiddleTruncation.tsx
bengotow: Hmm, I think this behavior is undesirable but it's not a recent regression -- let me undo this change here and I'll submit a separate PR for it. This is more "would be nice to fix for next release", and I'm actually not confident that 0.2 is high enough to fully eliminate the problem.

salazarm (approved):

js_modules/dagit/packages/core/src/assets/AssetNodeList.tsx
bengotow: Yeah I think that might be nice - now that we show the word "Materialized" we have had to make these bigger in a few places recently.

Issue #11787, opened 2023-01-18
backfills page shows error if asset targeted by pure asset backfill is no longer partitioned


sryza:
Error on the frontend:
```
Operation name: InstanceBackfillsQuery

Message: Internal Server Error (Trace ID: 5593880838688429486)

Path: ["partitionBackfillsOrError","results",1,"numPartitions"]

Locations: [{"line":7,"column":9},{"line":62,"column":3}]
```

Error inside the webserver:
```
builtins.AttributeError: 'NoneType' object has no attribute 'deserialize_subset'
Traceback (most recent call last):
  File "/dagster-graphql/dagster_graphql/schema/backfill.py", line 194, in resolve_numPartitions
    return self._backfill_job.get_num_partitions(_graphene_info.context)
  File "/dagster/dagster/_core/execution/backfill.py", line 121, in get_num_partitions
    asset_backfill_data = AssetBackfillData.from_serialized(
  File "/dagster/dagster/_core/execution/asset_backfill.py", line 124, in from_serialized
    target_subset=AssetGraphSubset.from_storage_dict(
  File "/dagster/dagster/_core/definitions/asset_graph_subset.py", line 162, in from_storage_dict
    partitions_subsets_by_asset_key[asset_key] = partitions_def.deserialize_subset(value)
AttributeError: 'NoneType' object has no attribute 'deserialize_subset'
```

PR #11812, opened 2023-01-20
When can't deserialize asset backfill because asset is no longer partitioned, return it as empty

sryza:
### Summary & Motivation

Addresses https://github.com/dagster-io/dagster/issues/11787

Prior to this change, the backfill UI would fail to load if any of the asset backfills had assets that were partitioned at the time of the backfill and are no longer partitioned.

A more principled solution would be to store some sort of snapshot of the backfill that's loadable even if the asset definitions change, but the approach in this PR assumes that, in most cases, users won't care about the partitions of a backfill that targeted an asset that's no longer partitioned.


alangenfeld (approved):
this looks clean to me

PR #11819, opened 2023-01-20
Fix 'dagster dev' command with workspace.yaml passed in

gibsondan:
Summary:
This is also a tuple - tested module/file/.toml, but not workspace.yaml. Now rectified.

Test Plan: New BK test

### DIFF
--- a/python_modules/dagster/dagster/_check/__init__.py
+++ b/python_modules/dagster/dagster/_check/__init__.py
@@ -1611,6 +1611,51 @@ def _check_tuple_items(
     return obj_tuple
 
 
+def tuple_elem(
+    ddict: Mapping,
+    key: str,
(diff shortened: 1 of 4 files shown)


python_modules/dagster/dagster/_check/__init__.py
```diff
@@ -1611,6 +1611,51 @@ def _check_tuple_items(
     return obj_tuple
 
 
+def tuple_elem(
+    ddict: Mapping,
+    key: str,
+    of_type: Optional[TypeOrTupleOfTypes] = None,
+    additional_message: Optional[str] = None,
+) -> List:
```
alangenfeld, line 1619: are these return types right?

alangenfeld (approved):
fix return types

PR #11821, opened 2023-01-20
Move pyproject.toml => origin generation to a WorkspaceLoadTarget class so that it can be reloaded in dagit

gibsondan:
Summary:
This gives us the nice property that you can change the contents of the pyproject.toml file and hit "Reload All" in dagit without needing to restart it to pick up the changes. Also clears the deck for returning multiple modules which I remember was requestsed.

Test Plan: BK, sanity check with a local dagit load

### DIFF
--- a/python_modules/dagster/dagster/_cli/workspace/cli_target.py
+++ b/python_modules/dagster/dagster/_cli/workspace/cli_target.py
@@ -1,7 +1,17 @@
 import os
 import sys
 from contextlib import contextmanager
-from typing import TYPE_CHECKING, Generator, Iterable, List, Mapping, Optional, Tuple, Union, cast
+from typing import (
+    TYPE_CHECKING,
+    Generator,
+    Iterable,
+    List,
+    Mapping,
+    Optional,
+    Tuple,
+    Union,
+    cast,
+)
 
 import click
 import tomli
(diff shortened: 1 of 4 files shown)


python_modules/dagster/dagster/_cli/workspace/cli_target.py
```diff
@@ -82,32 +93,28 @@ def are_all_keys_empty(kwargs: ClickArgMapping, keys: Iterable[str]) -> bool:
...
-                module_name=dagster_block["module_name"],
-                attribute=None,
-                working_directory=os.getcwd(),
-                location_name=None,
-            )
-        return None
+        return data.get("tool", {}).get("dagster")
+
+
+def _has_dagster_pyproject_toml():
```
schrockn, line 107: appears unused?

python_modules/dagster/dagster/_core/workspace/load_target.py
```diff
@@ -37,6 +40,28 @@ def create_origins(self):
...
+            return ModuleTarget(
+                module_name=dagster_block["module_name"],
+                attribute=None,
+                working_directory=os.getcwd(),
+                location_name=None,
+            ).create_origins()
+        return []
+
+
+class PyprojectFileTarget(NamedTuple("PyprojectFileTarget", []), WorkspaceLoadTarget):
```
schrockn, line 60: I think this should be "PyProject" rather than "Pyproject"

schrockn (approved):
Great thanks. A couple nits but gtg

PR #11824, opened 2023-01-20
[dagit] Add ErrorBoundary to Dagit to reduce severity of React errors

bengotow:
### Summary & Motivation

Corresponding Internal PR: https://github.com/dagster-io/internal/pull/4659

Hey folks, this PR wraps key components of Dagit in React "error boundaries", which give us the opportunity to present errors more gracefully and let users recover / continue to use other parts of the app. So far I've added them to:

- The top component that renders page content (everything below the toolbar)
- The details section of the asset partitions and events pages (which was a production issue yesterday)
- The asset graph and asset graph sidebar
- The op graph and op graph sidebar
- The run logs container and the run gannt chart
- The instance run timeline
- The `<Dialog />` component (errors in a dialog should never take down the rest of the page)

Anywhere else we should add these?

### Resetting after errors

The component I added takes a `resetErrorOnChange` prop which acts a bit like a useMemo dependency list -- if any of the values change and it's in an error state, it resets the error and tries to render it's subtree again. We need this for cases where the error boundary is not unmounted / remounted but it's content could meaningfully change. (eg: you click another asset or tab). I think this is better than putting a `key={}` on the Error Boundary because that would force a re-render on the whole subtree in the happy case.

### Cloud

In cloud dagit, we need to implement a top level context exposed by the new Error Boundary class in order to send errors (which are now caught) to DataDog. I think we also want to change the text to let users know that their errors are automatically submitted to us, so I made that text configurable. We can also disable the display of the error stack trace in cloud if we want to.

I tested these changes by adding `throw new Error()` to various React render methods and verifying that Dagit fails more gracefully. I also verified that this can catch the infinite recursion bug, which is a bit of an interesting one, by reverting to last night's code and running without the GraphQL patch.

image

image

Edit: Updated styling

image

### DIFF
--- a/js_modules/dagit/packages/core/src/overview/OverviewTimelineRoot.tsx
+++ b/js_modules/dagit/packages/core/src/overview/OverviewTimelineRoot.tsx
@@ -1,4 +1,13 @@
-import {Page, PageHeader, Heading, Box, TextInput, Button, ButtonGroup} from '@dagster-io/ui';
+import {
+  Page,
+  PageHeader,
+  Heading,
+  Box,
+  TextInput,
+  Button,
+  ButtonGroup,
+  ErrorBoundary,
+} from '@dagster-io/ui';
 import * as React from 'react';
 
 import {FIFTEEN_SECONDS, useQueryRefreshAtInterval} from '../app/QueryRefresh';
@@ -129,7 +138,9 @@ export const OverviewTimelineRoot = () => {
           </Box>
         </Box>
       </Box>
-      <RunTimeline loading={initialLoading} range={range} jobs={visibleJobs} />
+      <ErrorBoundary region="timeline">
+        <RunTimeline loading={initialLoading} range={range} jobs={visibleJobs} />
+      </ErrorBoundary>
     </Page>
   );
 };
--- a/js_modules/dagit/packages/ui/src/components/ErrorBoundary.tsx
+++ b/js_modules/dagit/packages/ui/src/components/ErrorBoundary.tsx
@@ -1,0 +1,84 @@
+import React from 'react';
+import styled from 'styled-components/macro';
+
+import {Box} from './Box';
+import {Colors} from './Colors';
+import {Body, Subheading} from './Text';
+import {FontFamily} from './styles';
+
+export type ErrorCollectionContextValue = {
+  errorStackIncluded: boolean;
+  errorCollectionMessage: string;
+  onReportError: (error: Error, context: Record<string, any>) => void;
+};
+
+export const ErrorCollectionContext = React.createContext<ErrorCollectionContextValue>({
+  errorStackIncluded: true,
+  errorCollectionMessage:
+    `Please report this error to the Dagster team via GitHub or Slack. ` +
+    `Refresh the page to try again.`,
+  onReportError: (error, context) => {
+    console.error(error, context);
+  },
+});
+
+interface ErrorBoundaryProps {
+  region: string;
+  resetErrorOnChange?: any[];
+}
+
+interface ErrorBoundaryState {
+  error: Error | null;
+  errorResetPropsValue: string | null;
+}
+
+export class ErrorBoundary extends React.Component<ErrorBoundaryProps, ErrorBoundaryState> {
+  state: ErrorBoundaryState = {error: null, errorResetPropsValue: null};
+
+  static contextType = ErrorCollectionContext;
+
+  componentDidUpdate() {
+    if (
+      this.state.error &&
+      this.state.errorResetPropsValue !== JSON.stringify(this.props.resetErrorOnChange)
+    ) {
+      this.setState({error: null, errorResetPropsValue: null});
+    }
+  }
+
+  componentDidCatch(error: Error, info: any) {
+    this.context.onReportError(error, {info, region: this.props.region});
+    this.setState({error, errorResetPropsValue: JSON.stringify(this.props.resetErrorOnChange)});
+  }
+
+  render() {
+    const {error} = this.state;
+    const {errorCollectionMessage, errorStackIncluded} = this
+      .context as ErrorCollectionContextValue;
+
+    if (error) {
+      return (
+        <Box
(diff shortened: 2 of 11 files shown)


js_modules/dagit/packages/core/src/overview/OverviewTimelineRoot.tsx
```diff
@@ -129,7 +138,9 @@ export const OverviewTimelineRoot = () => {
           </Box>
         </Box>
       </Box>
-      <RunTimeline loading={initialLoading} range={range} jobs={visibleJobs} />
+      <ErrorBoundary region="timeline" resetErrorOnChange={[range]}>
```
hellendag, line 141: `range` will change pretty often. Is that okay here? I could see it being an issue where an error is firing on every render, which would mean potentially re-triggering the error every time `range` changes, if I'm understanding this correctly.

js_modules/dagit/packages/ui/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,84 @@
...
+export type ErrorCollectionContextValue = {
+  errorStackIncluded: boolean;
+  errorCollectionMessage: string;
+  onReportError: (error: Error, context: Record<string, any>) => void;
+};
+
+export const ErrorCollectionContext = React.createContext<ErrorCollectionContextValue>({
+  errorStackIncluded: true,
+  errorCollectionMessage:
+    `Please report this error to the Elementl team via GitHub or Slack. ` +
```
hellendag, line 14: "the Dagster team" instead of "the Elementl team"?

js_modules/dagit/packages/ui/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,84 @@
...
+    const {errorCollectionMessage, errorStackIncluded} = this
+      .context as ErrorCollectionContextValue;
+
+    if (error) {
+      return (
+        <div
+          style={{
+            width: '100%',
+            height: '100%',
+            flex: 1,
```
hellendag, line 63: Does it need to be flex?

js_modules/dagit/packages/ui/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,84 @@
...
+            border: `1px solid ${Colors.HighlightRed}`,
+            background: 'white',
+            borderRadius: 3,
+            color: '#222',
```
hellendag, line 67: `Colors` values for these colors?

js_modules/dagit/packages/ui/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,84 @@
...
+    if (error) {
+      return (
+        <div
+          style={{
+            width: '100%',
+            height: '100%',
+            flex: 1,
+            border: `1px solid ${Colors.HighlightRed}`,
+            background: 'white',
+            borderRadius: 3,
```
hellendag, line 66: Border radius might be a little strange in places where the error display occupies a non-rounded container.

js_modules/dagit/packages/ui/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,84 @@
...
+            color: '#222',
+            padding: 15,
+          }}
+        >
+          <h3>Sorry, {this.props.region} can&apos;t be displayed.</h3>
+          <p style={{color: Colors.Gray300}}>{errorCollectionMessage}</p>
+          {errorStackIncluded && (
+            <code>
+              <pre>{`${error}`}</pre>
+            </code>
```
hellendag, line 76: Style the `<pre>` instead of wrapping it in `<code>`, since `<code>` is an inline element?

hellendag (commented):
Seems pretty good to me. Will leave it for @salazarm to take a look too.

js_modules/dagit/packages/ui/src/components/ErrorBoundary.tsx
bengotow: Ahh good call on these styles, I'll also swap this whole thing for a `<Box>`

js_modules/dagit/packages/core/src/overview/OverviewTimelineRoot.tsx
bengotow: Ahhh good call, I will remove this. Without it present, you have to move away from the timeline or reload the page to try again, but i think that's fine.

js_modules/dagit/packages/ui/src/components/ErrorBoundary.tsx
bengotow: I think so - my goal is to make the error box takes up the "whole space" it's replacing, regardless of how it's container is configured.

salazarm (approved):
lgtm

js_modules/dagit/packages/ui/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,86 @@
...
+      this.state.error &&
+      this.state.errorResetPropsValue !== JSON.stringify(this.props.resetErrorOnChange)
+    ) {
+      this.setState({error: null, errorResetPropsValue: null});
+    }
+  }
+
+  componentDidCatch(error: Error, info: any) {
+    // You can also log the error to an error reporting service
+    // Note: Sentry may be blocked by an ad-blocker, can't assume it's available
```
hellendag, line 51: Copy-pasted from somewhere?

hellendag (approved):
+1 🚢

js_modules/dagit/packages/ui/src/components/ErrorBoundary.tsx
salazarm: from here https://nextjs.org/docs/advanced-features/error-handling

bengotow: Ahh yes I will rm this before it merges - good catch

PR #11825, opened 2023-01-20
fix tz in time_windows_for_partition_keys

sryza:
### Summary & Motivation

This was causing test failures on my laptop that didn't replicate in buildkite, because of the different timezones.


clairelin135 (approved):
Thanks for catching this!

PR #11827, opened 2023-01-21
[graphql] launch backfills over assets with different partitionings, if all roots have same partitioning

sryza:
### Summary & Motivation

Recent changes enabled the backfill engine to include assets with different PartitionsDefinitions. However, there's still no way to launch such a backfill.  Based on some user discussions, we believe that the main scenario for launching this kind of value involves a single root asset and a set of assets downstream of it. I.e. "backfill X partitions of asset A and all the partitions of downstream assets B and C that are downstream of X".

This changes the launch backfill mutation to support asset selections that include assets with different partitionings, as long all of the assets are downstream of a set of root assets that are of the backfill and have the same PartitionsDefinition.  In that case, the `partition_names` will apply to the root assets, and partitions will be targeted for downstream assets if they descend from any of the partitions targeted by the root assets.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -2,6 +2,7 @@ import functools
 from collections import deque
 from heapq import heapify, heappop, heappush
 from typing import (
+    TYPE_CHECKING,
     AbstractSet,
     Callable,
     Dict,
@@ -28,10 +29,13 @@ from dagster._utils.cached_method import cached_method
 from .assets import AssetsDefinition
 from .events import AssetKey, AssetKeyPartitionKey
 from .freshness_policy import FreshnessPolicy
-from .partition import PartitionsDefinition
+from .partition import PartitionsDefinition, PartitionsSubset
 from .partition_mapping import PartitionMapping, infer_partition_mapping
 from .source_asset import SourceAsset
 from .time_window_partitions import TimeWindowPartitionsDefinition
+
+if TYPE_CHECKING:
+    from dagster._core.definitions.asset_graph_subset import AssetGraphSubset
 
 
 class AssetGraph:
@@ -147,6 +151,17 @@ class AssetGraph:
     def have_same_partitioning(self, asset_key1: AssetKey, asset_key2: AssetKey) -> bool:
         """Returns whether the given assets have the same partitions definition."""
         return self.get_partitions_def(asset_key1) == self.get_partitions_def(asset_key2)
+
+    def have_same_or_no_partitioning(self, asset_keys: Iterable[AssetKey]) -> bool:
+        partitions_defs = []
+        for asset_key in asset_keys:
+            partitions_def = self.get_partitions_def(asset_key)
+            if partitions_def:
+                partitions_defs.append(partitions_def)
+
+        return len(partitions_defs) <= 1 or all(
+            partitions_defs[i] == partitions_defs[0] for i in range(1, len(partitions_defs))
+        )
 
     def get_children(self, asset_key: AssetKey) -> AbstractSet[AssetKey]:
         """Returns all assets that depend on the given asset."""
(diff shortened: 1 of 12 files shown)


python_modules/dagster/dagster/_core/execution/backfill.py
```diff
@@ -245,20 +245,51 @@ def from_asset_partitions(
         backfill_id: str,
         asset_graph: AssetGraph,
         partition_names: Sequence[str],
+        anchor_asset: Optional[AssetKey],
         asset_selection: Sequence[AssetKey],
```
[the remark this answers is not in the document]
sryza: > asset selection is the overarching scope of assets that should be recalculated, and anchor assets provides the constraints of which partitions of which asset? and it's transitive?

Exactly.

Trying to answer your other questions has made me question whether we actually need this argument. It only makes sense to use when the anchor asset is the single root asset of an asset subselection.  In those cases, we can just infer that asset automatically by looking at the selection.

sryza:
@prha @bengotow I updated this PR to remove the anchorAsset param and instead infer it from the structure of the graph. I updated the title and description to reflect the changes.

gustavo-delfosim:
@sryza I believe you can support more than one inferred anchorAsset if all of the root assets share the same partition, and it would still be intuitive for the user.

One very common use-case would be having multiple hourly Assets with one daily Asset depending on them. It would be good to be able to select the multiple hourly Assets at the same time when calling for a backfill.

sryza:
@gustavo-delfosim good point. I updated the PR to allow multiple root assets that have the same partitioning.

python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -335,6 +350,75 @@ def toposort_asset_keys(self) -> Sequence[AbstractSet[AssetKey]]:
...
+                                )
+                            )
+                    else:
+                        child_partitions_subset = None
+
+                    if child not in all_assets:
+                        queue.append((child, child_partitions_subset))
+                        all_assets.add(child)
+
+        return result
```
bengotow, line 420: This looks good to me, but I think it'd be possible to compose a graph that fails this logic with something whacky like this:

```
A -> C (last 7 days) -> D
A -> B -> C (last 14 days) -> D
```

(Since it would choose only the last 7 days of C)

python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -335,6 +350,75 @@ def toposort_asset_keys(self) -> Sequence[AbstractSet[AssetKey]]:
...
+        - Asset matches the condition_fn
+        - Any of their ancestors >= initial_asset_partitions match the condition_fn
+
+        Visits parents before children.
+        """
+        from .asset_graph_subset import AssetGraphSubset
+
+        all_assets = set(initial_subset.asset_keys)
+        check.invariant(
+            len(initial_subset.asset_keys) == 1, "Multiple initial assets not yet supported"
```
bengotow, line 370: This might be unrelated but this seems like it still expects only a single root asset 👀

bengotow (approved):
This looks good @sryza! I added a couple comments -- I think we'd talked about exposing a "preview" method that runs the business logic in `from_asset_partitions` and reports back the number of partitions / runs that would be generated or an error. I can wire up the GraphQL if you want to break that out into a function 👋

python_modules/dagster/dagster/_core/definitions/asset_graph.py
sryza: @bengotow would you mind explaining your example in a little more detail? I'm having some trouble pinning down the notation.

sryza:
@bengotow I added the preview here: https://github.com/dagster-io/dagster/pull/12037

python_modules/dagster/dagster/_core/definitions/asset_graph.py
bengotow: Hey @sryza! Ahh sorry, in my example there are two assets in the graph that depend on different partitions of the same asset "C". The one that is visited first by this code is the one with the shortest path distance since this is a breadth-first-search.  We see A->C and put "child_partitions_subset = last 7 days" into the `queue` array on line 435. We never visit the B->C edge because `C` is already in `all_assets`, which means we don't visit the "last 14 days" dependency.

In general this algorithm does not support the same asset being referenced multiple times in the graph with different partition subsets. That might be totally fine, was mostly just calling out that it would generate an incorrect list of child partitions.

sryza: Ah gotcha - yeah, good point. Probably not super common, but I updated the PR to fix this for the sake of completeness

PR #11835, opened 2023-01-23
[EcsRunLauncher] Move env vars to container overrides instead of task definition config

gibsondan:
Summary:
These can be overridden on the container without needing to re-register a new task definition.

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws_tests/ecs_tests/launcher_tests/test_secrets.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws_tests/ecs_tests/launcher_tests/test_secrets.py
@@ -56,23 +56,24 @@ def test_environment(
     instance_cm,
     launch_run,
 ):
-    initial_task_definitions = ecs.list_task_definitions()["taskDefinitionArns"]
-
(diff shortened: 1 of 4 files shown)


python_modules/libraries/dagster-aws/dagster_aws_tests/ecs_tests/launcher_tests/test_secrets.py
```diff
@@ -148,23 +150,23 @@ def test_environment_with_container_context(
...
-    initial_task_definitions = ecs.list_task_definitions()["taskDefinitionArns"]
+    initial_tasks = ecs.list_tasks()["taskArns"]
 
     # Secrets config is pulled from container context on the run, rather than run launcher config
     config = {"env_vars": ["RUN_LAUNCHER_NAME=RUN_LAUNCHER_VALUE"]}
 
     with instance_cm(config) as instance:
         launch_run_with_container_context(instance)
 
-    # A new task definition is created
```
jmsanders, line 159: These new assertions don't actually prove that we're not registering a task definition. Since that's an important part of the expected behavior, I think we should assert against it to prevent regressions.

jmsanders (changes requested):

PR #11838, opened 2023-01-23
Simple dagster install script

shalabhc:
### Summary & Motivation

New users occasionally have issues installing `dagster` using `pip` following our instructions. An install script that checks the Python environment, handles known issues with installation, and reports failures with clear reasons can be very useful to new users, specially users new to Python as well.

For example due to missing `grpcio` wheels, some Mac M1 users are not able to install dagster. While we now have special [instructions](https://docs.dagster.io/getting-started/install#installing-dagster) for M1users, this script handles the M1 case automatically. It also prints details about the environment which can be useful if the install fails.

The `python` interpreter used to run the script should be from the same environment that you want to install `dagster` in.

An example of creating a new venv and running this script:
```bash
% python -m venv /tmp/install-test-venv

% /tmp/install-test-venv/bin/python scripts/install_dagster.py 

✨✨✨ Welcome to dagster! ✨✨✨

This will install the latest version of dagster and its dependencies.

Inspecting this system:
 ✅ Python executable: /private/tmp/install-test-venv/bin/python (version 3.10.9)
 ✅ Python environment: /private/tmp/install-test-venv
 ✅ Platform: macOS-12.6-arm64-arm-64bit

Installing dagster and dagit:
 ✅ successfully installed dagster and its dependencies
    ➤ you can now run 'dagster -h' or visit https://dagster.io to learn more

%
```

Eventually we can add docs for this script and also host it on a new domain so you could install `dagster` as

```
curl install.dagster.io | /path/to/python -
```

**NOTE: This script not intended to ever replace `pip`. If and when we add this to the instructions, the pip instructions should also be available next to it, eg in a tabbed section. Not everyone is comfortable running an install script and intermediate or advanced users should be able to use pip or a package installer of their choice.**

Manual testing on local machine.

### DIFF
--- a/scripts/install_dagster.py
+++ b/scripts/install_dagster.py
@@ -1,0 +1,97 @@
+"""
+This script installs dagster and its dependencies into the current Python environment.
+
+Run as `python3 install_dagster.py`.
+To install in a venv, run python from the venv: `/path/to/venv/bin/python3 install_dagster.py`
+"""
+
+import platform
+import subprocess
+import sys
+
+# pylint: disable=missing-function-docstring
+# pylint: disable=consider-using-f-string
+
+
+class InstallFailed(Exception):
+    """Unexpected installation failure"""
+
+
+def pip(*args):
+    return subprocess.run(
+        [sys.executable, "-m", "pip", *args], capture_output=True, check=False, encoding="utf-8"
+    )
+
+
+def has_pip():
+    proc = pip("-V")
+    return proc.returncode == 0
+
+
+def installed_packages():
+    proc = pip("freeze")
+    if proc.returncode:
+        raise InstallFailed("Could not determine installed packages", sys.stderr)
+    return dict(line.split("==", 1) for line in proc.stdout.splitlines(keepends=False) if "==" in line)
+
+
+def check_python():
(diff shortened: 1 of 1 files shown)


shalabhc:
* master
  * **PR #11838** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11838" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11838?utm_source=stack-comment).

gibsondan (commented):
Would we also do the whole curl from url | python thing?

It is nice that this handles the M1 stuff automatically under the hood..

smackesey:
This looks good.

One thought: if this is targeted to people uncomfortable with even `pip`, I would not expect them to be comfortable creating a virtual env. But the script seems to assume some knowledge of virtual envs:

```
To install in a venv, run python from the venv: `/path/to/venv/bin/python3 install_dagster.py`
```

Makes me think the script should handle venv creation as well.

shalabhc:
The grpcio issue seems resolved for now - not sure if we still want this.

gibsondan (changes requested):
i think we have this semi-tabled for now, but good to have it around if we decide we want it in the future

PR #11842, opened 2023-01-23
[asset-reconciliation] Fix case where the partitions definition does not have partitions for some subset of the past 24 hours

OwenKephart:
### Summary & Motivation

Fixes issue discussed in: https://github.com/dagster-io/dagster/discussions/11829#discussioncomment-4756197

Added test failed before this change, now passes. The allowable time window for a partitions def now takes into account the start date of that partitions definition, preventing this issue.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -301,10 +301,10 @@ def allowable_time_window_for_partitions_def(
         return None
 
     start = max(
-        # we use a greater than check later on, so we want to make sure we allow the earliest
-        # partition to be materialized
-        earliest_partition_window.start - datetime.timedelta.resolution,
-        latest_partition_window.start - datetime.timedelta(days=1),
+        earliest_partition_window.start,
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -295,8 +295,20 @@ def allowable_time_window_for_partitions_def(
...
         return None
+
+    earliest_partition_window = partitions_def.get_first_partition_window()
+    if earliest_partition_window is None:
+        return None
+
+    start = max(
+        # we use a greater than check later on, so we want to make sure we allow the earliest
+        # partition to be materialized
+        earliest_partition_window.start - datetime.timedelta.resolution,
```
sryza, line 306: Why would it ever be less than `earliest_partition_window.start`?

OwenKephart: the comment / logic was confusing, so I reorganized it (it still does fundamentally the same thing, just perhaps slightly more easy to understand)

sryza (approved):

PR #11843, opened 2023-01-23
fix(dbt-cloud): remove dbt selector when materializing subset

rexledesma:
### Summary & Motivation
According to https://docs.getdbt.com/reference/node-selection/syntax#examples, if a `--selector` flag is used for selection logic, other flags like `--select` and `--exclude` will be ignored.

> As your selection logic gets more complex, and becomes unwieldly to type out as command-line arguments, consider using a yaml selector. You can use a predefined definition with the --selector flag. Note that when you're using --selector, most other flags (namely --select and --exclude) will be ignored.

Since we append `--select` when we materialize subsets, our current subsetting logic is incompatible with the presence of a `--selector`. Fix this by removing the `--selector` once the subselection has already been determined.

pytest

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/asset_defs.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/asset_defs.py
@@ -438,6 +438,9 @@ class DbtCloudCacheableAssetsDefinition(CacheableAssetsDefinition):
 
                 dbt_options.append(f"--vars '{json.dumps(partition_var)}'")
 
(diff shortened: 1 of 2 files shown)


rexledesma:
* master
  * **PR #11843** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11843" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11843?utm_source=stack-comment).

OwenKephart (approved):

Issue #8821, opened 2022-07-11
footgun for not providing a partitions def to define_asset_job when targeting partitioned assets


sryza:
Example: https://dagster.slack.com/archives/C01U954MEER/p1657347125702809

A couple possible solutions:
- Raise an error at repo-definition time if partitioned assets are included in a non-partitioned asset job.
- Make it so that setting a partitions_def isn't required. This would mean that `build_schedule_from_partitioned_job` would need to defer building the schedule until repo-definition time.

PR #11844, opened 2023-01-23
remove need to provide `partitions_def` to asset job targeted by `build_schedule_from_partitioned_job`

sryza:
### Summary & Motivation

If you have a set of daily-partitioned assets and you want to build a daily schedule that targets them, you currently have to provide a `DailyPartitionsDefinition` to `define_asset_job`, even though you've already supplied that `DailyPartitionsDefinition` your assets. Doing so adds extra boilerplate and extra opportunity for error.  Users migrating from Airflow are especially likely to encounter this, because most Airflow DAGs are essentially partitioned schedules.

This PR removes that requirement - i.e. it infers the cadence of the schedule directly from the partitionings of assets targeted by the job.

It accomplishes this by adding the `UnresolvedPartitionedAssetScheduleDefinition` type and having `build_schedule_from_partitioned_job` return it. We need to do this because we can't get the `PartitionsDefinitions` for the assets targeted by the asset job until we resolve the asset selection, which happens at repository construction.

Fixes https://github.com/dagster-io/dagster/issues/8821

tests


python_modules/dagster/dagster/_core/definitions/repository_definition.py
```diff
@@ -850,19 +893,23 @@ def from_list(
...
                 )
 
-        # resolve all the UnresolvedAssetJobDefinitions using the full set of assets
-        for name, unresolved_job_def in unresolved_jobs.items():
-            if not combined_asset_group:
-                raise DagsterInvalidDefinitionError(
-                    f"UnresolvedAssetJobDefinition {name} specified, but no AssetsDefinitions exist"
-                    " on the repository."
+        if unresolved_jobs:
+            for (
```
clairelin135, line 897: why do we need to reevaluate unresolved asset schedules again when resolving unresolved jobs?

clairelin135 (approved):
All of these changes make sense to me. Had one open question commented below

python_modules/dagster/dagster/_core/definitions/repository_definition.py
sryza: OOPS

PR #11846, opened 2023-01-23
use an `AssetGraph` to resolve asset selections in asset jobs

sryza:
branch-name: internal-asset-graph

### Summary & Motivation

This should boost repository construction performance by avoiding asset graph construction every time we need to resolve an `UnresolvedAssetJobDefinition`.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -74,7 +74,9 @@ class AssetGraph:
         return self._freshness_policies_by_key
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -105,14 +107,16 @@ def from_assets(all_assets: Sequence[Union[AssetsDefinition, SourceAsset]]) -> "
 
             else:
                 check.failed(f"Expected SourceAsset or AssetsDefinition, got {type(asset)}")
-        return AssetGraph(
+        return InternalAssetGraph(
```
OwenKephart, line 110: this is a bit weird, as the ExternalAssetGraph contains its own static constructor, but the internal asset graph is constructed here, but it's not a huge deal

OwenKephart (approved):

PR #11847, opened 2023-01-23
[perf] Allow TimeWindowPartitionMapping to accept a DefaultPartitionsSubset

OwenKephart:
### Summary & Motivation

It's expensive to create a TimeWindowPartitionSubset, as this requires creating a cron string iterator. We create a lot of single-element TimeWindowPartitionSubsets inside the reconciliation sensor, purely for the purpose of figuring out what partitions are upstream/downstream of another partition. In most common cases, this is a 1-1 mapping (i.e. the partitions upstream of [2023-01-01] are [2023-01-01]). 

This PR enables passing in a DefaultPartitionsSubset to a TimeWindowPartitionMapping, only converting this into a TimeWindowPartitionsSubset if we need to (i.e. the mapping is non-trivial)

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -26,7 +26,7 @@ from dagster._utils.cached_method import cached_method
 from .assets import AssetsDefinition
 from .events import AssetKey, AssetKeyPartitionKey
 from .freshness_policy import FreshnessPolicy
-from .partition import PartitionsDefinition
+from .partition import DefaultPartitionsSubset, PartitionsDefinition
 from .partition_mapping import PartitionMapping, infer_partition_mapping
 from .source_asset import SourceAsset
 from .time_window_partitions import TimeWindowPartitionsDefinition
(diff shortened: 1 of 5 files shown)


python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -253,15 +257,20 @@ def get_parent_partition_keys_for_child(
...
 
         partition_mapping = self.get_partition_mapping(child_asset_key, parent_asset_key)
-        parent_partition_key_subset = partition_mapping.get_upstream_partitions_for_partitions(
-            cast(PartitionsDefinition, child_partitions_def)
-            .empty_subset()
-            .with_partition_keys([partition_key])
-            if partition_key
-            else None,
+
+        child_partitions_subset = (
```
sryza, line 261: I think that, ideally, the implementation behind the `PartitionsSubset` abstraction wouldn't leak to this part of the code. E.g. if we were able to move to a more performant implementation the cron string iterator, ideally it shouldn't affect the choices we make in this code.

What do you think about modifying `TimeWindowPartitionsSubset` itself to be more performant? E.g. enable it to be backed bye a list of partition keys instead of a list of time windows if the number of partition keys is small?

OwenKephart: yeah this is reasonable, and I started off with that sort of thing in mind, just didn't see a super easy way of making that work. I'll give it another run tho

OwenKephart:
closing this in favor of https://github.com/dagster-io/dagster/pull/11850

PR #11848, opened 2023-01-23
Fix for backoff logic incorrectly storing state

gibsondan:
Summary:
User reported an error with our backoff logic hitting a maximum python int limit. This is likely because we were initializing the default backoff generator once and keeping it running forever. This fixes that.

One question is whether we need to do anything to clean up the old generator as well.

### DIFF
--- a/python_modules/dagster/dagster_tests/general_tests/utils_tests/test_backoff.py
+++ b/python_modules/dagster/dagster_tests/general_tests/utils_tests/test_backoff.py
@@ -1,3 +1,5 @@
+import time
+
 import pytest
 from dagster._utils.backoff import backoff, backoff_delay_generator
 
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster_tests/general_tests/utils_tests/test_backoff.py
```diff
@@ -59,6 +73,8 @@ def test_backoff():
     assert backoff(fn, retry_on=(RetryableException,), args=[3, 2, 1], kwargs={"foo": "bar"})
     assert fn.call_count == 2
 
+    assert fake_sleep_times == [0.1, 0.2, 0.4, 0.8, 0.1]
```
gibsondan, line 76: before this was 1.6 not 0.1

prha (approved):

PR #11850, opened 2023-01-24
perf improvement for TimeWindowPartitionsSubset

OwenKephart:
### Summary & Motivation

In a lot of cases (i.e. when the number of individual partition keys is small), it's way more efficient for TimeWindowPartitionsSubset to just be a list of partition keys.

This PR makes that possible, only requiring the included_time_windows to be calculated when needed. These subsets will still always be serialized/deserialized with TimeWindows, this is mostly just for making stuff like `TimeWindowPartitionsDefinition(...).empty_subset().with_partition_keys([small, set, of, keys])` performant.

This results in a surprisingly dramatic performance improvement for the build_asset_reconciliation_sensor. For a toy example with 50 assets (all hourly partitioned):

**Before**: 70s
**After**: 2.5s

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
@@ -3,6 +3,7 @@ import json
 import re
 from datetime import datetime
 from typing import (
+    AbstractSet,
     Any,
     Callable,
     Iterable,
@@ -11,6 +12,7 @@ from typing import (
     NamedTuple,
     Optional,
     Sequence,
+    Tuple,
     Union,
     cast,
 )
@@ -543,7 +545,7 @@ class TimeWindowPartitionsDefinition(
         ) < self.start_time_for_partition_key(partition_key2)
 
     def empty_subset(self) -> "TimeWindowPartitionsSubset":
-        return TimeWindowPartitionsSubset(self, [], 0)
+        return TimeWindowPartitionsSubset(self, num_partitions=0, included_partition_keys=set())
 
     def deserialize_subset(self, serialized: str) -> "PartitionsSubset":
         return TimeWindowPartitionsSubset.from_serialized(self, serialized)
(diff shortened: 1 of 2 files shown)


sryza:
@OwenKephart let me know if this is ready for another round of review

OwenKephart:
@sryza this is ready for review

OwenKephart:
@sryza this is ready for review

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -1139,26 +1172,29 @@ def get_partition_keys_not_in_subset(
...
-    @property
-    def included_time_windows(self) -> Sequence[TimeWindow]:
-        return self._included_time_windows
-
-    def with_partition_keys(self, partition_keys: Iterable[str]) -> "TimeWindowPartitionsSubset":
-        result_windows = [*self._included_time_windows]
+    def _add_partitions_to_time_windows(
+        self, initial_windows: Sequence[TimeWindow], partition_keys: Sequence[str]
+    ) -> Tuple[Sequence[TimeWindow], int]:
+        """Merges a set of partition keys into an existing set of time windows, reuturning the
```
sryza, line 1194:
```suggestion
        """Merges a set of partition keys into an existing set of time windows, returning the
```

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -1202,8 +1238,26 @@ def with_partition_keys(self, partition_keys: Iterable[str]) -> "TimeWindowParti
 
                 num_added_partitions += 1
 
+        return result_windows, num_added_partitions
+
+    def with_partition_keys(self, partition_keys: Iterable[str]) -> "TimeWindowPartitionsSubset":
+        # if we are representing things as a static set of keys, continue doing so
```
sryza, line 1244: This comments implies that keys vs. windows is either/or, but invoking `get_partition_keys` allows getting into a state where both are present, which I found a little confusing.

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -1075,16 +1077,47 @@ class TimeWindowPartitionsSubset(PartitionsSubset):
...
+        )
+        self._included_partition_keys = (
+            sorted(
+                list(pkeys),
+                key=lambda pkey: self._partitions_def.start_time_for_partition_key(pkey),
+            )
+            if pkeys is not None
+            else None
+        )
+        print(self._included_partition_keys)
```
sryza, line 1109: print

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -1075,16 +1077,47 @@ class TimeWindowPartitionsSubset(PartitionsSubset):
...
+            included_time_windows, "included_time_windows", of_type=TimeWindow
+        )
+
+        pkeys = check.opt_nullable_set_param(
+            included_partition_keys, "included_partition_keys", of_type=str
+        )
+        self._included_partition_keys = (
+            sorted(
+                list(pkeys),
+                key=lambda pkey: self._partitions_def.start_time_for_partition_key(pkey),
```
sryza, line 1104: Is this not expensive? I don't actually know if this sorting matters.  When you asked me about the ordering of `get_partition_keys` on Slack, I thought you were asking about `PartitionsDefinition.get_partition_keys`.

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -1139,26 +1163,32 @@ def get_partition_keys_not_in_subset(
...
-        ]
+        if self._included_partition_keys is None:
+            return [
+                pk
+                for time_window in self.included_time_windows
+                for pk in self._partitions_def.get_partition_keys_in_time_window(time_window)
+            ]
+        return sorted(
+            list(self._included_partition_keys),
+            key=lambda pkey: self._partitions_def.start_time_for_partition_key(pkey),
```
sryza, line 1174: Is this not super unperformant? Same comment from earlier applies - do we need to sort?

sryza (approved):

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
OwenKephart: ah I misread your original comment -- we have tests that rely on this being in sorted order, but I can't see how a subset would require this. I'll remove this bit

PR #11851, opened 2023-01-24
remove support for PipelineDefinitions in RepositoryDefinition.from_list

sryza:
### Summary & Motivation

Small chip away at the legacy concepts inside the codebase


dpeng817 (changes requested):
Many legitimate test failures here

sryza:
Withdrawing this PR for now, because it seems like we need to migrate a bunch of tests first.

PR #11854, opened 2023-01-24
[dagit] Allow materializing “All” without requiring large graphs are rendered

bengotow:
### Summary & Motivation

Context:
https://elementl-workspace.slack.com/archives/C03CCE471E0/p1670944563171369

This change makes it possible to click "Materialize All" even if there are too many assets for the graph to render by default. We check for this case (`applyingEmptyDefault`) and set the selection for the purposes of both the Observe and Materialize buttons.

I also cleaned up a few stylistic things in this file because it's getting a bit large -- may try to do more cleanup soon.

I changed `MAX_RENDERED_FOR_EMPTY_QUERY` from 100 to 10 so I could easily test this "applyingEmptyDefault" case.

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
@@ -5,7 +5,6 @@ import without from 'lodash/without';
 import React from 'react';
 import styled from 'styled-components/macro';
 
-import {GraphQueryItem} from '../app/GraphQueryImpl';
 import {QueryRefreshCountdown, QueryRefreshState} from '../app/QueryRefresh';
 import {LaunchAssetExecutionButton} from '../assets/LaunchAssetExecutionButton';
 import {LaunchAssetObservationButton} from '../assets/LaunchAssetObservationButton';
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
```diff
@@ -112,31 +111,32 @@ export const AssetGraphExplorer: React.FC<Props> = (props) => {
...
-  } = props;
-
+type WithDataProps = {
+  allAssetKeys: AssetKey[];
+  assetGraphData: GraphData;
+  graphQueryItems: AssetGraphQueryItem[];
+  liveDataByNode: LiveData;
+  liveDataRefreshState: QueryRefreshState;
+  applyingEmptyDefault: boolean;
+} & Props;
```
hellendag, line 121: Is this more idiomatic for props?

```
interface WithDataProps extends Props {
  // etc.
}
```

hellendag (approved):

js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
bengotow: Hmm I think it's more consistent with what we do elsewhere, I'll change it in a follow-up 👍  From a technical standpoint I'm actually not sure what the differences are... 😳

Issue #11649, opened 2023-01-11
deep asset backfills can spend too much time in daemon sleep


sryza:
Problem I just discovered with partition-mapped asset backfills, initially reported by @tha23rd on Slack: when the chain of dependencies is very long (like in the case of partition self-deps), it's easy for the backfill daemon's sleep between iterations to dominate the length of the backfill.  I.e. fill partition 1, wait 29 seconds, fill partition 2, wait 29 seconds, etc.

A potential solution is to reduce the daemon interval time to a couple seconds, during periods that backfills are incomplete.  We'd want to make sure that the backfill daemon doesn't put too much load on the DB when nothing has changed.

PR #11860, opened 2023-01-25
[RFC] more frequent backfill iterations when there are active backfills

sryza:
### Summary & Motivation

We recently introduced the ability for asset backfills to wait until some assets are materialized to launch other runs. For example, if an asset's partitions depend on earlier partitions of the same asset, we essentially wait for each partition to complete before launching a run for the next one.

Because the default daemon sleep interval is 30 seconds, this means that the minimum time to complete such a backfill is 30 seconds * # partitions. If the tasks complete quickly, then much of the backfill will be spent in daemon sleep.

This PR proposes an approach to address this by reducing the sleep interval to 2 seconds when there are any active backfills.

Addresses https://github.com/dagster-io/dagster/issues/11649


python_modules/dagster/dagster/_daemon/backfill.py
```diff
@@ -1,19 +1,78 @@
...
+    while True:
+        start_time = pendulum.now("UTC").timestamp()
+        if until and start_time >= until:
+            # provide a way of organically ending the loop to support test environment
+            break
+
+        for backfill in execute_backfill_iteration(workspace_process_context, logger):
+            if backfill:
+                if backfill.status.is_active:
+                    active_backfills.add(backfill.backfill_id)
```
gibsondan, line 45: i am very down with the overall plan here - but I think given that the very first thing that execute_backfill_iteration does is fetch all requested backfills, we may be able to accomplish this goal without needing to pass anything back from the generator?

Could the logic be something like

- fetch all requested backfills in this function (instead of at the beginning of execute_backfill_iteration)
```
backfills = instance.get_backfills(status=BulkActionStatus.REQUESTED)
``` 
- pass those backfills into execute_backfill_iteration as an argument

- If there were any backfills in that list, wait ACTIVE_BACKFILLS_INTERVAL_SECONDS - if not wait DEFAULT_DAEMON_INTERVAL_SECONDS

The downside of that is you might do one last extra fast loop once there are no requested backfills, but that's not too expensive and it will short-circuit quickly

gibsondan: you could also consider only doing the ACTIVE_BACKFILLS_INTERVAL_SECONDS if there are specifically active asset backfills? from the description sounds like its only useful for those

python_modules/dagster/dagster/_daemon/backfill.py
```diff
@@ -1,19 +1,78 @@
...
 
 
 def execute_backfill_iteration(
     workspace_process_context: IWorkspaceProcessContext,
     logger: logging.Logger,
     debug_crash_flags: Optional[Mapping[str, int]] = None,
-) -> Iterable[Optional[SerializableErrorInfo]]:
+) -> Iterable[Optional[PartitionBackfill]]:
     instance = workspace_process_context.instance
     backfills = instance.get_backfills(status=BulkActionStatus.REQUESTED)
```
gibsondan, line 77: I think one question here is whether we actually need to re-fetch the backfill object itself from the DB every 2 seconds - if this PR caused problems I'd suspect hammering the backfill DB to be the most likely cause. If it doesn't need to fetch the updated backfill on each iteration and just needs to run other checkup logic more frequently, i'd suggest moving all logic that fetches the list of backfills to the outer function and running the refetch less frequently / only after a certain amount of time has passed. If it does need to re-fetch the backfill object on each 2 second tick to work, then it may be unavoidable

gibsondan (changes requested):
overall plan here makes sense to me (curious if you have any concerns @prha ) - i just have some thoughts on how to maybe simplify the interactions between the two functions - and wondering if we actually need to regenerate the backfill object DB every 2 seconds or if there's just some other logic using the backfill object we need to run more frequently

python_modules/dagster/dagster/_daemon/backfill.py
sryza: The main risk would be that someone cancels the backfill, but we go on submitting runs

python_modules/dagster/dagster/_daemon/backfill.py
sryza: I believe the majority of backfills (including asset backfills) will transition from REQUESTED to COMPLETED within the first `execute_backfill_iteration` that covers them. So we'd be doing roughly double the number of fast loops. But I agree that the loops will be very cheap when there are no requested backfills.

Is this suggestion about mainly code readability or performance?

gibsondan:
I guess the former but not strongly held

python_modules/dagster/dagster/_daemon/backfill.py
sryza: Response from Daniel that I received via email but somehow didn't make it on here: "I guess the former but not strongly held"

@prha - care to break the tie?

python_modules/dagster/dagster/_daemon/backfill.py
```diff
@@ -1,19 +1,78 @@
...
+
+
+def execute_backfill_iteration_loop(
+    workspace_process_context: IWorkspaceProcessContext,
+    logger: logging.Logger,
+    shutdown_event: threading.Event,
+    until=None,
+) -> "TDaemonGenerator":
+    """
+    Helper function that performs sensor evaluations on a tighter loop, while reusing grpc locations
```
prha, line 28: fix this comment (looks like copy/pasta from the sensor daemon)

python_modules/dagster/dagster/_daemon/backfill.py
prha: I think in the job backfill case, at least, we already pause every chunk of runs that we schedule to check for updates to the status.  And that would hammer the db every second?

python_modules/dagster/dagster/_daemon/backfill.py
prha: I do think that the loops of loops is a bit tough to reason about, especially since we have branched logic between assets/jobs, encompassed in the same job pass.  May be out of the scope of this particular PR, but I'd much rather see a split daemon between job backfills / asset backfills, since we've now broken the constraint that backfills are completed after a successful iteration.

I'm not that worried about the rate of DB fetches, since we have tight fetches in other cases (e.g. a single backfill scheduling a large number of runs will also hit the DB in a tight loop to look for backfill cancels).

cmpadden:

PR #11863, opened 2023-01-26
Upgrade base image on Dagster Dockerfiles to latest python version

gibsondan:
Summary:
Use python:3.x-slim instead of pinning to a specific patch release, so that we automatically move to the latest version on each push


.buildkite/dagster-buildkite/dagster_buildkite/utils.py
```diff
@@ -148,7 +148,8 @@ def connect_sibling_docker_container(
 
 
 def is_feature_branch(branch_name: str = safe_getenv("BUILDKITE_BRANCH")) -> bool:
-    return not (branch_name == "master" or branch_name.startswith("release"))
+    #    return not (branch_name == "master" or branch_name.startswith("release"))
+    return False
```
smackesey, line 152: Is this supposed to be here?

python_modules/automation/automation/docker/images/buildkite-test/Dockerfile
```diff
@@ -97,6 +96,15 @@ RUN curl -sL https://deb.nodesource.com/setup_14.x | bash - \
...
     && java -version
 
+# Install FOSSA cli for analyzing our open-source dependencies
+# https://github.com/fossas/fossa-cli
+RUN wget https://github.com/fossas/fossa-cli/releases/download/v${FOSSA_VERSION}/fossa-cli_${FOSSA_VERSION}_linux_amd64.tar.gz \
+   && tar xvzf fossa-cli_${FOSSA_VERSION}_linux_amd64.tar.gz fossa \
+   && chmod +x ./fossa \
+   && mv ./fossa /usr/local/bin/fossa \
+   && rm fossa-cli_${FOSSA_VERSION}_linux_amd64.tar.gz \
+   && fossa --version
```
smackesey, line 106: Why not use their recommended method of the [v1 install script](https://github.com/fossas/fossa-cli/blob/master/docs/walkthroughs/installing-fossa-cli.md#installing-cli-1x-using-installation-script)?

smackesey (approved):

Issue #11794, opened 2023-01-19
Unexpected GraphQL Error for Schedules


jerome-laurent-pro:
## Dagster version
1.1.10

## What's the issue?

Unexpected GraphQL Error when opening the Schedule screen or any asset on Dagit.

```
Operation name: LatestRunTagQuery

Message: Invalid version: '5.7.38-log'

Path: ["pipelineRunsOrError","results"]

Locations: [{"line":4,"column":7}]
```

## What did you expect to happen?

Schedules and assets are displayed correctly.
The schedules work.

## Additional information

This involves the use of [MySQL storage](https://docs.dagster.io/deployment/dagster-instance#mysql-storage) to persist the metadata.
For the local dev, I added a MySQL container to my `docker-compose`, and everything was running smoothly.
```
  db:
    profiles: [ "local" ]
    image: mysql:5.7
    container_name: dagster_mysql
    hostname: ${MYSQL_HOSTNAME}
    restart: always
    env_file:
      - .env
    environment:
      - MYSQL_DATABASE=${MYSQL_DATABASE}
      - MYSQL_ROOT_PASSWORD=${MYSQL_ROOT_PASSWORD}
      - MYSQL_USER=${MYSQL_USERNAME}
      - MYSQL_PASSWORD=${MYSQL_PASSWORD}
```

But for deployment, I used the MySQL instance of my company on AWS RDS.
It's working (the metadata is persisted in the MySQL database created for the occasion) BUT there is this specific error appearing on any asset page or on the schedule page. And the schedules don't work.

The MySQL version of the RDS is `5.7.38` so I don't know where the `log` (in `5.7.38-log`) is coming from.

## How to reproduce?

I tried to reproduce it locally but failed.

prha:
This should be resolved now with dagster release `1.1.14`. Feel free to open a new issue if you still see this.

If upgrading is not an option, you could also pin `packaging<22.0`

PR #11868, opened 2023-01-27
stop using python semver parsing for mysql versions

prha:
### Summary & Motivation
We were using `packaging.parse` to compare mysql server versions, which broke when `0.23.0` was released with dropped support for `LegacyVersion`.

This rolls our own custom version compares to find minimum supported versions, which is tolerant of non semver compliant version strings.

The regex parses all the numeric values, and does tuple int comparisons.

Reported in https://github.com/dagster-io/dagster/issues/11794

Reference links:
https://github.com/pypa/packaging/pull/407

Added some test cases that override the server version string to exercise the parsing logic.


python_modules/libraries/dagster-mysql/dagster_mysql/utils.py
```diff
@@ -46,6 +47,26 @@ def get_conn_string(username, password, hostname, db_name, port="3306"):
     )
 
 
+def parse_mysql_version(version: str) -> tuple:
```
alangenfeld, line 50: add some test cases that use this directly too

alangenfeld (approved):

PR #11873, opened 2023-01-27
Mutable Partitions Storage

clairelin135:
This PR implements the storage layer of the "runtime partitions definition" feature, renaming to "mutable partitions definition" as the partitions are generated via instance calls and not at runtime.

- Implements a new `PartitionsStorage` class alongside run storage and event log storage. Creates only a `PostgresPartitionsStorage` class as we currently only plan to support cloud use cases in the near term.
- Creates a `mutable_partitions_definition` table with columns:
  - `partitions_def_name`: the name of the mutable partitions definition
  - `partition_key`: a singular partition key that belongs to the partition def
- For the above table, creates an index with `[partitions_def_name, partition_key]` to allow for faster querying for presence of a certain partition key for a partitions def
- Creates a `MutablePartitionsDefinition` class, adding the following methods to the class and dagster instance:
  - Add partition keys
  - Delete partition keys
  - Check if a partition key exists
- Tests for postgres partitions storage

As all migrations are managed by alembic, this migration will automatically run upon `dagster instance migrate` for all users. We will gate this behavior by not creating a `PartitionsStorage` object to read/write from the table unless if the user is using unified postgres storage. Should future migrations occur, we will prevent reading/writing from the table unless if the user has run the latest migration.


clairelin135:
The following PR https://github.com/dagster-io/dagster/pull/11900 enables Dagit to load assets with mutable partitions definitions.

python_modules/dagster/dagster/_core/definitions/mutable_partitions_definition.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    A partitions definition whose partition keys can be dynamically added and removed.
+
+    This is useful for cases where the set of partitions is not known at definition time,
+    but is instead determined at runtime.
+
+    Partitions can be added and removed using the `add_partitions` and `remove_partitions` methods.
+    For example:
+
+    ```
```
sryza, line 26: We should use the standard sphinx examples syntax.  Look at `@asset` for an example.

python_modules/dagster/dagster/_core/definitions/mutable_partitions_definition.py
```diff
@@ -0,0 +1,86 @@
...
+    foo = MutablePartitionsDefinition("foo")
+
+    @sensor(job=my_job)
+    def my_sensor(context):
+        foo.add_partitions([partition_key])
+        return my_job.run_request_for_partition(partition_key)
+    ```
+    """
+
+    def __init__(self, name: str, instance: Optional[DagsterInstance] = None):
```
sryza, line 36: My instinct here is that, in general, you shouldn't need to instantiate a `DagsterInstance` in order to construct your `Definitions`. 
- Abstractly, the set of Definitions should be independent of the instance the user is running against.
- Concretely, constructing a `DagsterInstance` can has a lot of thorns. E.g. `DagsterInstance.get` raises an error if `DAGSTER_HOME` isn't set.

Instead, methods like `add_partitions` should accept an `instance` argument.  I know this potentially adds a lot of complexity / additional lifting if `get_partitions` needs to accept an instance argument, so not 100% sure on this, but that's where I would start.

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1682,6 +1709,49 @@ def get_materialization_count_by_partition(
...
+    def get_mutable_partitions(self, partitions_def_name: str) -> Sequence[str]:
+        check.str_param(partitions_def_name, "partitions_def_name")
+        if self._partitions_storage is None:
+            check.failed(
+                "Your storage configuration currently does not support the partitions storage."
+            )
+        return self._partitions_storage.get_partitions(partitions_def_name)
+
+    @traced
+    def add_mutable_partitions(
```
sryza, line 1724: I've internally struggled on this name a little. On one hand, having `mutable` in there makes it clear that this only applies to `MutablePartitionsDefinition`s. `mutable_partition` is weird because it's not the partition itself that's mutable, it's the set of partitions.  We could name it something like `add_partition_to_mutable_partitions`, but that's a little wordy.

Curious if @prha or @OwenKephart have thoughts.

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1682,6 +1709,49 @@ def get_materialization_count_by_partition(
...
+        check.str_param(partitions_def_name, "partitions_def_name")
+        check.sequence_param(partition_key, "partition_key", of_type=str)
+        if self._partitions_storage is None:
+            check.failed(
+                "Your storage configuration currently does not support the partitions storage."
+            )
+        self._partitions_storage.delete_partition(partitions_def_name, partition_key)
+
+    @traced
+    def has_mutable_partition(self, partitions_def_name: str, partition_key: str) -> bool:
```
sryza, line 1746: I suspect we'll eventually want to introduce a version of this that checks multiple partition keys. So I wonder if it's better to just do that now. I'm not super opinionated on this though.

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1682,6 +1709,49 @@ def get_materialization_count_by_partition(
...
+        check.str_param(partitions_def_name, "partitions_def_name")
+        check.sequence_param(partition_key, "partition_key", of_type=str)
```
sryza, line 1748: sequence?

python_modules/dagster/dagster/_core/storage/partitions/base.py
```diff
@@ -0,0 +1,44 @@
+from abc import ABC, abstractmethod
+from typing import Callable, Optional, Sequence
+
+from dagster._core.instance import MayHaveInstanceWeakref
+
+
+class PartitionsStorage(ABC, MayHaveInstanceWeakref):
```
sryza, line 7: Curious about @prha and/or @alangenfeld 's thoughts on adding `PartitionsStorage` vs. making this part of of `EventLogStorage`.

python_modules/dagster/dagster/_core/storage/partitions/sql_partitions_storage.py
```diff
@@ -0,0 +1,97 @@
...
+        return [row[1] for row in rows]
+
+    def get_partitions(self, partitions_def_name: str) -> Sequence[str]:
+        """Get the list of partition keys for a partition definition."""
+        self._check_partitions_table()
+        return self._fetch_partition_keys_for_partition_def(partitions_def_name)
+
+    def has_partition(self, partitions_def_name: str, partition_key: str) -> bool:
+        """Check if a partition exists."""
+        check.str_param(partitions_def_name, "partitions_def_name")
```
sryza, line 51: Because these aren't user-facing methods, I don't think we need checks here. We can rely on the type annotations to catch issues.

python_modules/dagster/dagster/_core/storage/alembic/versions/035_add_mutable_partitions_table.py
```diff
@@ -0,0 +1,49 @@
...
+revision = "e62c379ac8f4"
+down_revision = "6df03f4b1efb"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    if not has_table("mutable_partitions_definitions"):
+        op.create_table(
+            "mutable_partitions_definitions",
```
sryza, line 24: It doesn't quite roll of the tongue, but my instinct would be to name this something like `mutable_partitions_definitions_partitions`, because each element represents a (partitions_definition, partition_key) tuple, not just a partitions_definition. Would defer to @prha.

sryza (commented):
Awesome to see this so far along already!

python_modules/dagster/dagster/_core/storage/partitions/base.py
clairelin135: Yeah... I was unsure about this one too. It didn't seem to make sense as part of event log storage as the partitions aren't generated from events, but I don't feel strongly about this.

python_modules/dagster/dagster/_core/storage/alembic/versions/035_add_mutable_partitions_table.py
prha: `mutable_partitions_definitions_partitions` seems like too much... I prefer something more terse

python_modules/dagster/dagster/_core/storage/base_storage.py
```diff
@@ -31,3 +33,8 @@ def run_storage(self) -> RunStorage:
     @abstractmethod
     def schedule_storage(self) -> ScheduleStorage:
         raise NotImplementedError()
+
+    @property
+    @abstractmethod
+    def partitions_storage(self) -> Optional[PartitionsStorage]:
```
prha, line 39: I don't think we should implement a different storage abstraction... let's roll this into event log storage or run storage.

python_modules/dagster/dagster/_core/storage/partitions/base.py
prha: Let's just roll this into event log storage.  We're moving towards unifying all the storage abstractions anyway.

python_modules/dagster/dagster/_core/storage/alembic/versions/035_add_mutable_partitions_table.py
sryza: Let's go with `mutable_partitions`.

clairelin135:
Closing this out in favor of splitting into 2 prs, one with storage changes and the other with partitions definition changes

PR #11874, opened 2023-01-27
fix config issue when materializing a non-partitioned asset within a partitioned asset job

sryza:


cmpadden:

PR #11876, opened 2023-01-27
[dagster-airflow] `load_assets_from_airflow_dag` 

Ramshackle-Jamathon:
### Summary & Motivation
SDAs are the future. This PR adds an experimental `load_assets_from_airflow_dag` api to dagster-airflow that allows for converting an airflow DAG into a collection of partitioned, graph-backed dagster Assets.

Given the following airflow DAG:
Screen Shot 2023-01-26 at 5 08 04 PM

A user can make the following `task_ids_by_asset_key` mappings to create the following SDAs
Screen Shot 2023-01-26 at 5 04 09 PM

Which under the hood are backed by an 1:1 task:op mapped graph.
Screen Shot 2023-01-26 at 5 04 20 PM

- [x] unit tests are still needed


examples/with_airflow/setup.py
```diff
@@ -14,4 +14,4 @@
         "apache-airflow-providers-docker>=3.1.0",
     ],
     extras_require={"dev": ["dagit", "pytest"]},
-)
+)
```
sryza, line 17: Nitpick: keep newline at end of file

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
```diff
@@ -249,3 +253,106 @@ def make_airflow_example_dags():
     return make_dagster_definitions_from_airflow_dag_bag(
         dag_bag=dag_bag, use_ephemeral_airflow_db=use_ephemeral_airflow_db
     )
+
+def make_graph_backed_assets_from_airflow_dag(
```
sryza, line 257: Type annotations plz

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
```diff
@@ -249,3 +253,106 @@ def make_airflow_example_dags():
...
+    dag,
+    asset_key_to_task_ids=None,
+    connections=None,
+):
+    """Construct a Dagster Assets forr Airflow DAG.
```
sryza, line 262:
```suggestion
    """Construct a Dagster Assets for Airflow DAG.
```

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
```diff
@@ -249,3 +253,106 @@ def make_airflow_example_dags():
     return make_dagster_definitions_from_airflow_dag_bag(
         dag_bag=dag_bag, use_ephemeral_airflow_db=use_ephemeral_airflow_db
     )
+
+def make_graph_backed_assets_from_airflow_dag(
+    dag,
+    asset_key_to_task_ids=None,
```
sryza, line 259: Assuming this is a `Mapping[AssetKey, str]`, our typical naming scheme for this is like `task_ids_by_asset_key`

sryza: Separately, do we want to provide some way for people to say that some task inputs are provided by assets outside of the job?

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
```diff
@@ -249,3 +253,106 @@ def make_airflow_example_dags():
...
+        AssetDefinition
+    """
+    cron_schedule = dag.normalized_schedule_interval
+
+    job = make_dagster_job_from_airflow_dag(dag, use_ephemeral_airflow_db=True, connections=connections)
+    graph = job._graph_def
+    start_date = dag.start_date if dag.start_date else dag.default_args.get("start_date")
+
+    # if no asset_
+    if asset_key_to_task_ids is None or asset_key_to_task_ids == {}:
```
sryza, line 280: What if only some of the leaf task IDs are covered? Should we raise an error? Automatically fill in the remaining?

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
```diff
@@ -249,3 +253,106 @@ def make_airflow_example_dags():
...
+        # leaf nodes have no downstream nodes
+        forward_edges, _ = _create_adjacency_lists(graph.solids, graph.dependency_structure)
```
sryza, line 282: Ideally we'd avoid using the legacy `solids`.  We could do that either by adding a `nodes` property to `GraphDefinition` or using `graph.nodes_dict.values()`.

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
```diff
@@ -249,3 +253,106 @@ def make_airflow_example_dags():
...
+        }
+        asset_key_to_task_ids = {
+            dag.dag_id: leaf_nodes
+        }
+
+    output_mappings=set()
+    keys_by_output_name = {}
+    internal_asset_deps={}
+
+    visit_dict = {}
```
sryza, line 294: Could this just be a set called `visited_nodes`?

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
```diff
@@ -249,3 +253,106 @@ def make_airflow_example_dags():
...
+    upstream_deps = set()
+
+    def find_upstream_dependency(solid_name: str) -> None:
```
sryza, line 297:
```suggestion
    def find_upstream_dependency(node_name: str) -> None:
```

sryza: It's taking me a little while to grok what this function is doing. I think it's worth trying to give it a more specific name and/or add some docstring.

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
```diff
@@ -249,3 +253,106 @@ def make_airflow_example_dags():
...
+    )
+
+    asset_def = AssetsDefinition.from_graph(
+        graph_def=new_graph,
+        partitions_def=TimeWindowPartitionsDefinition(
+            cron_schedule=cron_schedule,
+            timezone=dag.timezone.name,
+            start=start_date.strftime("%Y-%m-%d"),
+            fmt="%Y-%m-%d"
+        ) if isinstance(cron_schedule, str) and is_valid_cron_schedule(cron_schedule) else None,
```
sryza, line 351: If it's not a valid cron schedule, I think we should probably avoid failing silently. Are there valid use cases for it not to be a valid cron schedule? Should we just error if it's not?

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
```diff
@@ -249,3 +253,106 @@ def make_airflow_example_dags():
...
+        resource_defs=job.resource_defs,
+        group_name=dag.dag_id,
+        keys_by_output_name=keys_by_output_name,
+        internal_asset_deps=internal_asset_deps,
```
sryza, line 355: Let me know if I'm misunderstanding, but I don't think we need to provide this argument. This argument is mainly for when there's structure within _a single op_.  If it's within a graph, then I believe we automatically infer the asset dependencies from the op dependencies within the graph.

sryza (changes requested):
Excited for this!  I haven't given this a detailed review, but wanted to leave some initial comments in the time I had.

I think this needs `make black` and `make ruff`.

petehunt:
Looks pretty sweet. A few thoughts:
* I think we can convert the examples to Definitions, right?
* Prefer to call it `load_assets_from_airflow_dag` so we don’t have to teach users about graph backed assets at this stage of onboarding. It also aligns with the naming from the dbt integration.

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
sryza: @OwenKephart can confirm or deny what I'm saying

Ramshackle-Jamathon: Without this arg the SDA graph will look like:
Screen Shot 2023-01-27 at 2 37 12 PM
But maybe there's another spot to declare the dependencies?

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
Ramshackle-Jamathon: if a user does that the following error currently appears
Screen Shot 2023-01-27 at 2 41 47 PM
do you think making a more specific one is worthwhile?

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
Ramshackle-Jamathon: > do we want to provide some way for people to say that some task inputs are provided by assets outside of the job?

good point, i think we do

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
Ramshackle-Jamathon: I'm going to auto fill them in, should avoid at least a confusing debug session for a new user

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
sryza: Ah OK - sounds like it is required. I think we should modify `AssetsDefinition.from_graph` to do this on its own, but not a blocker for this PR.

OwenKephart: I don't think it's intended that we'd have this behavior -- we do have [code in place](https://sourcegraph.com/github.com/dagster-io/dagster/-/blob/python_modules/dagster/dagster/_core/definitions/asset_layer.py?L386:5) to parse out the dependencies within graph-backed assets, which apparently I wrote but have no recollection of.

so this is probably a bug but a good workaround in the meantime is indeed just doing what you're doing

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
Ramshackle-Jamathon: added a `upstream_asset_keys_by_task_id` param to do this

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
```diff
@@ -249,3 +258,163 @@ def make_airflow_example_dags():
...
+    dag,
+    task_ids_by_asset_key: Optional[Mapping[AssetKey, Set[str]]] = None,
+    upstream_asset_keys_by_task_id: Optional[Mapping[str, Set[AssetKey]]] = None,
+    connections: List[Connection] = None,
+) -> AssetsDefinition:
+    """Construct Dagster Assets for a given Airflow DAG.
+
+    Args:
+        dag (DAG): The Airflow DAG to compile into a Dagster job
+        task_ids_by_asset_key (Optional[Mapping[AssetKey,Set[str]]]): A mapping from asset keys to task ids. Used break up the Airflow Dag into multiple SDAs
```
sryza, line 351: Keep under line limits, and wrapped text should start a tab ahead of the argument. Also, include spaces after commas in tyypes.

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
```diff
@@ -249,3 +258,163 @@ def make_airflow_example_dags():
...
+
+    return (output_mappings, keys_by_output_name, internal_asset_deps)
+
+
+def load_assets_from_airflow_dag(
+    dag,
+    task_ids_by_asset_key: Optional[Mapping[AssetKey, Set[str]]] = None,
+    upstream_asset_keys_by_task_id: Optional[Mapping[str, Set[AssetKey]]] = None,
+    connections: List[Connection] = None,
+) -> AssetsDefinition:
```
sryza, line 346: Wrapping this in a list would make it more consistent with `load_assets_from_dbt_project`

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
```diff
@@ -249,3 +258,163 @@ def make_airflow_example_dags():
...
+        # the tasks for a given asset should have the same internal deps
+        for task_id in task_ids_by_asset_key[asset_key]:
+            internal_asset_deps[f"result_airflow_{task_id}"] = asset_upstream_deps
+
+    return (output_mappings, keys_by_output_name, internal_asset_deps)
+
+
+def load_assets_from_airflow_dag(
+    dag,
+    task_ids_by_asset_key: Optional[Mapping[AssetKey, Set[str]]] = None,
```
sryza, line 343: `Set` -> `AbstractSet`

sryza (commented):
Looking close! Still needs tests and a few fixups

sryza (approved):
Awesome

PR #11877, opened 2023-01-27
fix: use new GitHub graphql resolvers for issue automation

rexledesma:
### Summary & Motivation
Basically follow https://docs.github.com/en/enterprise-cloud@latest/issues/planning-and-tracking-with-projects/automating-your-project/automating-projects-using-actions, but only use the fields that are in use by the automation.

create an issue after merging


yuhan (approved):
ty!

Issue #11882, opened 2023-01-27
Unpinned dependecies on dagster python module


menecio:
### Dagster version

latest

### What's the issue?

There are multiple dependencies that aren't pinned to specific versions or they don't have any limits, this causes issues like having SQLAlchemy==2.0.0 installed and breaking compatibility.

All python dependencies should be pinned at least defining a minimum and maximum versions to avoid unwanted versions.

### What did you expect to happen?

To have all dependencies versions pinned at setup.py

### How to reproduce?

Check the code

### Deployment type

None

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

AntonFriberg:
This breaks even the simple getting started guide with a confusing error message.

```
$ pip install dagster dagit
...
$ pip freeze | grep dag
dagit==1.1.13
dagster==1.1.13
dagster-graphql==1.1.13
$ dagster dev -m dagster_example
2023-01-27 16:33:14 +0100 - dagster - INFO - Using temporary directory /home/antonfr/Git/github/dagster-project-example/tmpw9kazjkq for storage. This will be removed when dagster dev exits.
2023-01-27 16:33:14 +0100 - dagster - INFO - To persist information across sessions, set the environment variable DAGSTER_HOME to a directory to use.
Traceback (most recent call last):
  File "/home/antonfr/Git/github/dagster-project-example/.venv/bin/dagster", line 8, in <module>
    sys.exit(main())
  File "/home/antonfr/Git/github/dagster-project-example/.venv/lib/python3.10/site-packages/dagster/_cli/__init__.py", line 46, in main
    cli(auto_envvar_prefix=ENV_PREFIX)  # pylint:disable=E1123
  File "/home/antonfr/Git/github/dagster-project-example/.venv/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/antonfr/Git/github/dagster-project-example/.venv/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/antonfr/Git/github/dagster-project-example/.venv/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/antonfr/Git/github/dagster-project-example/.venv/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
[...]

PR #11890, opened 2023-01-27
Cap some dagster dependencies to at most their current major version

gibsondan:
Summary:
The goal of this change would be to avoid automatically upgrade to new major versions, which often include breaking changes (e.g. sqlalchemy 2 which broke all dagster users yesterday who made a fresh install)
The downside of this change would be that on older versions, users could be potentially pinned to older versions of their dependencies.
Not strongly held as a good idea but wanted to put it out for discussion.

Resolves https://github.com/dagster-io/dagster/issues/11882

### DIFF
--- a/python_modules/dagster-graphql/setup.py
+++ b/python_modules/dagster-graphql/setup.py
@@ -35,8 +35,8 @@ setup(
     include_package_data=True,
     install_requires=[
         f"dagster{pin}",
-        "graphene>=3",
-        "gql[requests]>=3.0.0",
+        "graphene>=3,<4",
+        "gql[requests]>=3,<4",
         "requests",
         "starlette",  # used for run_in_threadpool utility fn
     ],
--- a/python_modules/dagster/setup.py
+++ b/python_modules/dagster/setup.py
@@ -76,7 +76,7 @@ setup(
     install_requires=[
         # cli
         "click>=5.0",
-        "coloredlogs>=6.1, <=14.0",
+        "coloredlogs>=6.1,<=14.0",
         "Jinja2",
         "PyYAML>=5.1",
         # core (not explicitly expressed atm)
@@ -87,26 +87,26 @@ setup(
         f"grpcio-health-checking>={GRPC_VERSION_FLOOR}",
         "packaging>=20.9",
         "pendulum>=0.7.0,<3",
-        "protobuf>=3.20.0",  # min protobuf version to be compatible with both protobuf 3 and 4
+        "protobuf>=3.20.0,<5",  # min protobuf version to be compatible with both protobuf 3 and 4
         "python-dateutil",
         "python-dotenv",
         "pytz",
         "requests",
         "setuptools",
         "tabulate",
-        "tomli",
-        "tqdm",
-        "typing_extensions>=4.4.0",
-        "sqlalchemy>=1.0",
+        "tomli<3",
+        "tqdm<5",
+        "typing_extensions>=4.4.0,<5",
+        "sqlalchemy>=1.0,<3",
         "toposort>=1.0",
         "watchdog>=0.8.3",
-        'psutil >= 1.0; platform_system=="Windows"',
+        'psutil>=1.0; platform_system=="Windows"',
         # https://github.com/mhammond/pywin32/issues/1439
-        'pywin32 != 226; platform_system=="Windows"',
+        'pywin32!=226; platform_system=="Windows"',
         "docstring-parser",
         "universal_pathlib",
         # https://github.com/pydantic/pydantic/issues/5821
-        "pydantic >1.10.0,!= 1.10.7",
+        "pydantic>1.10.0,!= 1.10.7,<3",
         f"dagster-pipes{pin}",
     ],
     extras_require={


schrockn:
I think this a fine temporary solution but has a couple downsides:

1) There are lots of breakages that happen where non-major releases are pushed.
2) This could atrophy without some of sort of automated process.

There is a fancier, semi-automated version of that that could be built that would handle minor release as well. . Essentially we would pin dependency to "last known good" and then update that on a weekly basis along with every release, thus avoiding outages between releases. That script is where we would track known problematic releases.

I think this change puts the project in a better place than it is now so I support it.

smackesey:
I understand the impetus behind this idea, but I think it would be a big mistake. It is almost guaranteed to cause many frustrating dependency headaches for users. We must understand: **Users can always add their own pin to fix an underconstraint, but have no way of removing an overaggressive pin**.

>The goal of this change would be to avoid automatically upgrade to new major versions, which often include breaking changes (e.g. sqlalchemy 2 which broke all dagster users yesterday who made a fresh install)

This _is_ a problem, but I strongly believe it is a smaller problem than what would be introduced by this change. Our CI catches these new versions fast and let's us add a pin when necessary.

A smart guy wrote a [small novel](https://iscinumpy.dev/post/bound-version-constraints/) on the pitfalls of this approach. Here's an excerpt:

>Capping dependencies has long term negative effects, especially for libraries, and should never be taken lightly. A library is not installed in isolation; it has to live with other libraries in a shared environment. Only add a cap if a dependency is known to be incompatible or there is a high (>75%) chance of it being incompatible in its next release. Do not cap by default - capping dependencies makes your software incompatible with other libraries that also have strict lower limits on dependencies, and limits future fixes. Anyone can fix a missing cap, but users cannot fix an over restrictive cap causing solver errors. It also encourages hiding issues until they become harder to fix, it does not scale to larger systems, it limits your ability to access security and bugfix updates, and some tools (Poetry) force these bad decisions on your downstream users if you make them. Never cap Python, it is fundamentally broken at the moment. Also, even packing capping has negative consequences that can produce unexpected solves.

> ...

>If you need a TL;DR for the TL;DR, I’ll just [quote](https://snarky.ca/why-i-dont-like-semver/) Python Steering Council Member and packaging expert Brett Cannon:

>Libraries/packages should be setting a floor, and if necessary excluding known buggy versions, but otherwise don’t cap the maximum version as you can’t predict future compatibility

---

Side note: I have actually been [trying](https://github.com/dbt-labs/dbt-core/issues/6461#issuecomment-1359353107) to get `dbt` to stop using preemptive pins because it has caused incompatibilities with various dagster packages. I'm not alone either, they regularly have issues around these conflicts. They have now begun reevaluating their philosophy:

- https://github.com/dbt-labs/dbt-core/discussions/6495

ei-grad:
Here is a useful piece of experience - https://github.com/apache/airflow#approach-for-dependencies-for-airflow-core (maybe sqlalchemy should not be <2.0 but <1.5).

Actually, Airflow team didn't follow their own rules having sqlalchemy>1.4 in the setup.cfg - https://github.com/apache/airflow/pull/28725. But they found the problem by CI failures before sqlalchemy-2.0 has been released. I don't know how exactly, maybe by using the --pre flag, but wouldn't it cause too many false positives?..

ei-grad:
> There is a fancier, semi-automated version of that that could be built that would handle minor release as well. . Essentially we would pin dependency to "last known good" and then update that on a weekly basis along with every release, thus avoiding outages between releases. That script is where we would track known problematic releases.

Do you mean it only applicable to an official distribution like docker images and the dagster cloud infrastructure / or as a recommended practice for OSS deployments? Having dependencies pinned in a public library would prevent users from being able to control the environment of their pipelines.

shalabhc:
I was on the fence but after reading @smackesey 's links I think we should not cap at all. Capping minor versions will also prevent key bugfixes and security patches from being immediately available, until we move our pins. Capping major versions is only useful for some libraries like sqlalchemy and seems over aggressive for compatible libraries.

I think locking and periodic upgrades is the right approach but has to be done by the end-user, not library authors.

schrockn:
I'm inclined to move forward with this change despite @smackesey well-articulated argument @shalabhc's concurrence.

Sean note that I think you can have this change in place and still philosophically be aligned with what you are saying. The only thing we need to make it consistent is some sort of process to ensure that we are not leave these pins in place permanently.

The only goal here is prevent breakages that we _know_ will happen and will break users. When a major release of a dependency happens, it should be our policy to _remove_ that pin as soon as feasibly possible.

I think these setup.py reflects the fact, accurately, that we are not pre-emptively testing again major upcoming releases of these libraries. At our scale I think that is the right call. This change is not a change of policy. It just accurately reflects that status quo, a status quo that I think we should maintain.

schrockn:
@smackesey in particular to drive home the point, merging this PR does not represent an embrace of dbt's extremely aggressive pinning policy, which I believe is actually quite a hostile stance to take for a commonly used library in the Python ecosystem. This is just reflecting the reality that we do not test against upcoming major versions of new dependencies.

smackesey:
Thanks for your reply @schrockn. Regardless of the final decision it will be useful to have the case for/against on the record, so below I go into more detail:

>Sean note that I think you can have this change in place and still philosophically be aligned with what you are saying. The only thing we need to make it consistent is some sort of process to ensure that we are not leave these pins in place permanently.

I think this is partially true, in that with perfect implementation of a bound-updating process, you can (mostly) eliminate negative impact for users of the latest version of Dagster. However, even a perfect process will cause a hassle for users of older Dagster versions (more on this below).

>The only goal here is prevent breakages that we know will happen and will break users. 

If that's the case, why not take the approach Schreiner suggests in the excerpt I posted: "Only add a cap if a dependency is known to be incompatible or there is a high (>75%) chance of it being incompatible in its next release." IIUC this PR is preemptively capping _every_ first-order dependency.

---

Here's an attempt at birds-eye cost-benefit analysis:

In the limit case where new dep versions are tested immediately on release, with bound bumped and a new dagster release immediately issued if needed-- then upper bounds could only be good, since they would only ever exclude nonexistent and truly incompatible versions.

The question is, how close can we realistically get to that limit case? 

Assume (a) we keep our weekly release cadence; (b) new dep versions are released at a random time in the week. Then if new dep releases are perfectly monitored, it will take on average 3.5 days for the latest Dagster release to reflect the result of monitoring for a given dep.

Scenario 1: We use preemptive upper bounds and have add new testing infrastructure to continually monitor PyPI for new dep releases. Immediately after a new dep release, it is tested and the upper bound bumped in master if tests pass.

Scenario 2: Status quo. No upper bound pins except for known incompatibilities. Our CI immediately catches most incompatibilities in new dep releases (due to highest-allowed-version policy in constraint solving). Upper bounds need to be added to get CI to pass any incompatibility.

In both scenarios, an erroneous bound (or lack thereof) will be present for 3.5 days on average. But Scenario 2 seems preferable because:

- During the 3.5 day error period, users can fix the problem themselves in Scenario 2 (add the bound themselves) but not Scenario 1 (can't remove a bound)
- Scenario 1 requires significant additional infrastructure (you have to continuously monitor PyPI for new releases and when a new one drops... raise the pin in a feature branch that runs on BK? There must be a better way, but I'm not sure). Scenario 2 does not.

That covers errors for users installing the latest version of Dagster. But what about users on older versions? A preemptive pin policy, even the perfect "limit case" above, still requires users on older Dagster versions to upgrade Dagster to access compatible upgrades from other libraries (this becomes a much bigger issue if you are also bounding minor/patch releases). In contrast, the status quo leaves users on older versions free to bump their deps as they please. IMO this is the single biggest reason not to preemptively pin.

schrockn:
> If that's the case, why not take the approach Schreiner suggests in the excerpt I posted: "Only add a cap if a dependency is known to be incompatible or there is a high (>75%) chance of it being incompatible in its next release." IIUC this PR is preemptively capping every first-order dependency.

This sounds like a great compromise/approach. @gibsondan want to apply that filter?

schrockn (dismissed):
Make this less aggressive per @smackesey 's feedback. Only cap things that we have high certainty will break us.

gibsondan:
In honor of pendulum upgrading to 3.0 over the weekend and breaking us, finally getting back around to this - did a much less aggressive pass, only on packages with known breaking changes in the past

schrockn (approved):
Although less pure, given what happened over the weekend, I think this is the right approach.

alangenfeld:
In the spirit of centralizing information around this decision in this PR given all the existing excellent discussion, linking a case where the defensive caps have caused some friction for users https://github.com/dagster-io/dagster/issues/25398

gibsondan:
one thing we could consider is adding more tests like this to proactively notify about new versions where we have pins to make sure they are still needed https://github.com/dagster-io/dagster/blob/master/python_modules/libraries/dagster-k8s/dagster_k8s_tests/unit_tests/test_version.py#L38-L45

PR #11900, opened 2023-01-27
[Dynamic Partitions 3] Display dynamic partitions in Dagit

clairelin135:
Enables Dagit to load assets with dynamic partitions definitions and their corresponding data. Dagit will be able to load dynamic partitions definitions with the `name` property, dynamic partitions definitions with the `partition_fn` param are not supported.

Creates a `CachingDynamicPartitionsLoader` object for a group of asset nodes to ensure that when an asset graph/job is loaded, we will fetch the list of partition keys for a given dynamic partitions definition from the instance a maximum of one time. Afterwards, results are cached and returned when desired.

Adds the dagster instance to several grpc endpoints. These endpoints are used when attempting to launch runs, e.g. fetching a dynamic partition's config, tags etc), which means that they should be called on a one-off basis.

Locally tested a set of views in dagit:
- Launching a singular run
- Launching a backfill
- Using the asset reconciliation sensor
- Viewing the partition health bar
- Asset details page
- Job partitions page

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/implementation/loader.py
+++ b/python_modules/dagster-graphql/dagster_graphql/implementation/loader.py
@@ -21,6 +21,7 @@ from dagster._core.scheduler.instigation import InstigatorType
 from dagster._core.storage.pipeline_run import JobBucket, RunRecord, RunsFilter, TagBucket
 from dagster._core.storage.tags import REPOSITORY_LABEL_TAG, SCHEDULE_NAME_TAG, SENSOR_NAME_TAG
 from dagster._core.workspace.context import WorkspaceRequestContext
+from dagster._utils.cached_method import cached_method
 
 
 class RepositoryDataType(Enum):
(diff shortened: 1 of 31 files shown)


python_modules/dagster-graphql/dagster_graphql/implementation/loader.py
```diff
@@ -329,6 +330,27 @@ def _fetch(self) -> None:
...
+    A batch loader that fetches partitions for a partition set.  This loader is expected to be
+    instantiated with a partition set name.
+    """
+
+    def __init__(self, instance: DagsterInstance):
+        self._instance = instance
+
+        self._partition_keys_by_mutable_partitions_def: Dict[str, Sequence[str]] = {}
+
+    def get_mutable_partitions(self, partitions_def_name: str) -> Sequence[str]:
```
sryza, line 344: Could we use `@cached_method` for this?

sryza (changes requested):

clairelin135:
@sryza makes sense, updated this PR to contain changes for your comments above

python_modules/dagster-graphql/dagster_graphql/implementation/loader.py
```diff
@@ -329,6 +330,27 @@ def _fetch(self) -> None:
         }
 
 
+class MutablePartitionsLoader(MutablePartitionsStore):
+    """
+    A batch loader that fetches partitions for a partition set.  This loader is expected to be
```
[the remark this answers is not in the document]
clairelin135: ah this comment was outdated

python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
```diff
@@ -327,6 +334,10 @@ def get_partition_keys(
         )
         check.opt_int_param(start_idx, "start_idx")
         check.opt_int_param(end_idx, "end_idx")
+
+        if not self._mutable_partitions_loader:
+            check.failed("mutable_partitions_loader must be provided to get partition keys")
```
sryza, line 339: Is this string right?

sryza (approved):
Nice

python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
clairelin135: Yes, the intent is so that when loading asset graphs / asset jobs any asset nodes all have a shared instance of the mutable partitions loader. I think I've added a mutable partitions loader to every asset node callsite, so nobody should run into this error

Issue #11867, opened 2023-01-27
Support dbt-core 1.4.* in dagster-dbt


simonloach:
### Dagster version

1.1.13

### What's the issue?

When I am trying to load my dbt manifest using `load_assets_from_dbt_manifest` into an asset I am getting an error in the dbt CLI imported by dagster. 

```
/home/jovyan/.cache/pypoetry/virtualenvs/dagster-demo-ahNgZajK-py3.8/lib/python3.8/site-packages/dagster/_core/workspace/context.py:588: UserWarning: Error loading repository location repo.py:ValueError: None is not a valid IndirectSelection

Stack Trace:
  File "/home/jovyan/.cache/pypoetry/virtualenvs/dagster-demo-ahNgZajK-py3.8/lib/python3.8/site-packages/dagster/_grpc/server.py", line 256, in __init__
    self._loaded_repositories: Optional[LoadedRepositories] = LoadedRepositories(
  File "/home/jovyan/.cache/pypoetry/virtualenvs/dagster-demo-ahNgZajK-py3.8/lib/python3.8/site-packages/dagster/_grpc/server.py", line 114, in __init__
    loadable_targets = get_loadable_targets(
  File "/home/jovyan/.cache/pypoetry/virtualenvs/dagster-demo-ahNgZajK-py3.8/lib/python3.8/site-packages/dagster/_grpc/utils.py", line 33, in get_loadable_targets
    else loadable_targets_from_python_file(python_file, working_directory)
  File "/home/jovyan/.cache/pypoetry/virtualenvs/dagster-demo-ahNgZajK-py3.8/lib/python3.8/site-packages/dagster/_core/workspace/autodiscovery.py", line 28, in loadable_targets_from_python_file
    loaded_module = load_python_file(python_file, working_directory)
  File "/home/jovyan/.cache/pypoetry/virtualenvs/dagster-demo-ahNgZajK-py3.8/lib/python3.8/site-packages/dagster/_core/code_pointer.py", line 86, in load_python_file
    return import_module_from_path(module_name, python_file)
  File "/home/jovyan/.cache/pypoetry/virtualenvs/dagster-demo-ahNgZajK-py3.8/lib/python3.8/site-packages/dagster/_seven/__init__.py", line 53, in import_module_from_path
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 848, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
[...]

PR #11902, opened 2023-01-27
feat(dagster-dbt): support dbt-core 1.4.x

rexledesma:
### Summary & Motivation
Resolves https://github.com/dagster-io/dagster/issues/11867.

Support both `v7/dbt-core==v1.3.*` and `v8/dbt-core==v1.4.*`.

Some things that have changed:
- The manifest artifact (empty fields are now none, so we should be more defensive when parsing these fields):
  - `v7`: https://schemas.getdbt.com/dbt/manifest/v7/index.html
  - `v8`: https://schemas.getdbt.com/dbt/manifest/v8/index.html
- The structured logging schema has different top-level keys https://docs.getdbt.com/reference/events-logging

bk


rexledesma:
* master
  * **PR #11902** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11902" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11902?utm_source=stack-comment).

examples/project_fully_featured/setup.py
```diff
@@ -23,10 +23,8 @@
         "dagster-pyspark",
         "dagster-slack",
         "dagster-postgres",
-        "dbt-core",
-        "dbt-duckdb",
+        "dbt-duckdb==1.3.1",  # dbt-duckdb is not yet compatible with dbt 1.4.*
```
sryza, line 26: Why are we pinning duckdb, not dbt?

python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
```diff
@@ -83,10 +83,12 @@ def _select_unique_ids_from_manifest_json(
 ) -> AbstractSet[str]:
     """Method to apply a selection string to an existing manifest.json file."""
     try:
+        import dbt.flags as flags
         import dbt.graph.cli as graph_cli
         import dbt.graph.selector as graph_selector
         from dbt.contracts.graph.manifest import Manifest
         from dbt.graph import SelectionSpec
+        from dbt.graph.selector_spec import IndirectSelection
```
sryza, line 91: Have we tested to make sure that this won't raise the CheckError when using dbt >=1.0.0, but < 1.4

sryza (commented):
How can we verify this works / make sure it's not breaking existing functionality

python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
OwenKephart: the dbt_13X tox environment installs dbt 1.3, and runs all the same tests

examples/project_fully_featured/setup.py
OwenKephart: once the dbt-duckdb adapter is updated we can remove the pin on dbt-duckdb, but if dbt-core comes first in this list then I think dbt-core will be pulled in as 1.4.x

so it's easier to just specify the pin on the package that's causing the problem

sryza (approved):

PR #11903, opened 2023-01-27
Enable use of arguments when using the `asset` function directly instead of as a decorator

nsfinkelstein:
### Summary & Motivation

I would like to be able to create assets by calling the `dagster.asset` function directly and passing in a callable, rather than using it is a decorator on a function.

I noticed that though this functionality is enabled, all the named arguments to the `dagster.asset` function get dropped.

I tested these changes locally by creating assets by calling the `dagster.asset` function directly. Before the changes, the assets were created, but named arguments to the `dagster.asset` function had no effect. In particular, dependencies specified through the `non_argument_dependencies` argument to the `dagster.asset` function were not registered. After the changes, the assets were created with the appropriate dependencies.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
@@ -170,23 +170,7 @@ def asset(
             def my_asset(my_upstream_asset: int) -> int:
                 return my_upstream_asset + 1
     """
-    if compute_fn is not None:
-        return _Asset()(compute_fn)
-
-    def inner(fn: Callable[..., Any]) -> AssetsDefinition:
-        check.invariant(
-            not (io_manager_key and io_manager_def),
-            (
-                "Both io_manager_key and io_manager_def were provided to `@asset` decorator. Please"
-                " provide one or the other. "
-            ),
-        )
-        if resource_defs is not None:
-            experimental_arg_warning("resource_defs", "asset")
-
-        if io_manager_def is not None:
-            experimental_arg_warning("io_manager_def", "asset")
-
+    def create_asset():
         return _Asset(
             name=cast(Optional[str], name),  # (mypy bug that it can't infer name is Optional[str])
             key_prefix=key_prefix,
@@ -207,7 +191,26 @@ def asset(
             freshness_policy=freshness_policy,
             retry_policy=retry_policy,
             code_version=code_version,
-        )(fn)
+        )
+
+    if compute_fn is not None:
+        return create_asset()(compute_fn)
+
+    def inner(fn: Callable[..., Any]) -> AssetsDefinition:
+        check.invariant(
+            not (io_manager_key and io_manager_def),
+            (
+                "Both io_manager_key and io_manager_def were provided to `@asset` decorator. Please"
+                " provide one or the other. "
+            ),
+        )
+        if resource_defs is not None:
+            experimental_arg_warning("resource_defs", "asset")
+
+        if io_manager_def is not None:
+            experimental_arg_warning("io_manager_def", "asset")
+
+        return create_asset()(fn)
 
     return inner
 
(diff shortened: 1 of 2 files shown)


sryza:
Hey @nsfinkelstein  - for people who want to use the `asset` function directly instead of a decorator, I've generally recommended:
```python
from dagster import asset

def my_function():
    ...

my_asset = asset(io_manager_key=..., ...)(my_function)
```

Would that work for you?

nsfinkelstein:
Thanks for the reply. Can I ask why that's your recommendation? Typically decorators can be called directly by passing in a callable. I think that's a well established idiom, and I'm worried breaking that idiom makes the code harder to read and reason about. In my case, I'm using `asset` primarily as a decorator, except for in one place where the asset function is itself computed dynamically, in which case `asset` cannot be used as a decorator. Because I'm using it in both ways, I'd prefer for the relationship between calling it directly and applying it as a decorator to be the standard pythonic relationship if at all possible.

If there is a reason for not permitting the `asset` function to be called directly with a callable, then I'd suggest issuing a warning or raising an error when it's attempted. At the moment, the code successfully returns an asset, but silently ignores all keyword arguments. In my case, this caused a relatively subtle bug that took a while to pin down.

sryza:
> Typically decorators can be called directly by passing in a callable. I think that's a well established idiom

Do you have examples of this idiom in Python core libraries or commonly used Python packages? Specifically where a decorator can receive both a function and other arguments at the same time?

I'm not necessarily opposed to this, but I'm a little hesitant about offering two ways of doing the same thing. I.e. after this change, both of these would work, right?
```python
my_asset = asset(io_manager_key=..., ...)(my_function)
my_asset = asset(io_manager_key=..., ..., compute_fn=my_function)
```

I definitely agree that we should raise an error if we choose not to support this.

nsfinkelstein:
@sryza I've just understood why this is set up the way it is.

With a typical decorator factory, the decorator factory has to be called, even when it is not given any arguments. For example, you'd need to write:

```py
@asset()
def myasset():
    ...
```

instead of


```py
@asset
def myasset():
    ...
```

with no parenthases.

So basically the current code is wanting to use the `asset` function as both a decorator factory as well as a decorator at the same time. The assumption is that when it is passed the `compute_fn` directly, then it is being used as a decorator and not a decorator factory, and then it is safe to ignore the keyword arguments.

As I've discovered, that assumption can be faulty. If it's called with the `compute_fn` directly as well as keyword args, then I have to think those keyword args should be used and not ignored.

With this new understanding, I'm reverting my most recent commit back to my original proposed solution. I agree with you that these would leave two ways of calling `dagster.asset` not as a decorator. However, the alternatives are to either disallow use of `asset` as a decorator (as opposed to decorator factory), which is not an option at this point, or to silently ignore keyword args with the `compute_fn` is passed directly. I'd argue that having the two options you outlined available is the better option, because silently ignoring arguments can lead to difficult bugs.

sryza:
> I would still think that's more efficient than having someone try to call dagster.asset directly with both the computer_fn and one or more keyword args, only to receive an error telling them they have to do it the other way.

I find that reasoning fairly convincing.

@schrockn - curious if you have thoughts, as we'd probably want to update a bunch of our decorators to conform to this pattern?  I.e. what do you think of allowing:

```python
from dagster

def my_func():
    ...

my_asset = asset(compute_fn=my_func, io_manager_key=...)
```

schrockn:
I think this is a great change.

Thanks @nsfinkelstein for  such a thoughtful explanation and PR.

I'd love to see an automated test that verifies this behavior.

nsfinkelstein:
That's very reasonable. I'd be happy to add tests, but it might not be for a week or two. My best guess is that the tests belong in [this file](https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster_tests/asset_defs_tests/test_assets.py); please let me know if there's somewhere else they should go.

sryza:
This one next to it would be the best place: https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster_tests/asset_defs_tests/test_decorators.py

nsfinkelstein:
I've added the relevant tests. The two that call `dagster.asset` directly without keyword args pass both before and after the change. The one that calls `dagster.asset` directly with a keyword arg fails before the change but passes after the change.

sryza (approved):
LGTM!

sryza:
Thanks for the contribution @nsfinkelstein !

PR #11908, opened 2023-01-27
[pyright] examples/project-fully-featured

smackesey:
### Summary & Motivation

Fix some type errors and remove `aiobotocore` pin.

### How I Tested These Changes

BK

PR #11909, opened 2023-01-27
[pyright] examples/quickstart-gcp

smackesey:
### Summary & Motivation

Fix type error in `quickstart-gcp` example

BK

### DIFF
--- a/examples/quickstart_gcp/quickstart_gcp/io_managers.py
+++ b/examples/quickstart_gcp/quickstart_gcp/io_managers.py
@@ -46,7 +46,7 @@ class BigQueryDataframeIOManager(IOManager):
 
     def load_input(self, context: InputContext):
         # upstream_output.asset_key is the asset key given to the Out that we're loading for
-        table_name = context.upstream_output.asset_key.to_python_identifier()
+        table_name = context.asset_key.to_python_identifier()
 
         df = pandas_gbq.read_gbq(
             f"SELECT * FROM `{self._dataset_id}.{table_name}`", credentials=self._credentials


smackesey:
* master
  * **PR #11908** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11908" target="_blank">Graphite</a> 
    * **PR #11909** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11909" target="_blank">Graphite</a>  👈
      * **PR #11910** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11910" target="_blank">Graphite</a> 
        * **PR #11911** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11911" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11909?utm_source=stack-comment).

examples/quickstart_gcp/quickstart_gcp/io_managers.py
```diff
@@ -46,7 +46,9 @@ def handle_output(self, context: OutputContext, obj: pd.DataFrame):
 
     def load_input(self, context: InputContext):
         # upstream_output.asset_key is the asset key given to the Out that we're loading for
-        table_name = context.upstream_output.asset_key.to_python_identifier()
+        output = context.upstream_output
+        assert output is not None
```
sryza, line 50: We generally avoid asserts in non-test code. This should be able to just use `context.asset_key` though.

sryza (changes requested):

examples/quickstart_gcp/quickstart_gcp/io_managers.py
smackesey: Unfortunately there is no good alternative to `assert` that I'm aware of for type-narrowing:

```
def some_fn() -> Union[Foo, Bar]:
    ...

x = some_fn()  # x has type Union[Foo, Bar]
assert isinstance(x, Foo)
x  # x now has type Foo
```

`check.inst` does not work here, and there is no way to make it work that I'm aware of (and I've tried quite a bit).

examples/quickstart_gcp/quickstart_gcp/io_managers.py
```diff
@@ -46,10 +46,11 @@ def handle_output(self, context: OutputContext, obj: pd.DataFrame):
 
     def load_input(self, context: InputContext):
         # upstream_output.asset_key is the asset key given to the Out that we're loading for
-        table_name = context.upstream_output.asset_key.to_python_identifier()
+        table_name = context.asset_key.to_python_identifier()
```
schrockn, line 49: is `context.upstream_output.asset_key` and `context.asset_key` guaranteed to be the same in all cases?

smackesey: I think so but inquiring with @sryza for confirmation on this, who suggested `context.asset_key`

sryza: Yes - it should be.

examples/quickstart_gcp/quickstart_gcp/io_managers.py
```diff
@@ -46,10 +46,11 @@ def handle_output(self, context: OutputContext, obj: pd.DataFrame):
...
 
         df = pandas_gbq.read_gbq(
-            f"SELECT * FROM `{self._dataset_id}.{table_name}`", credentials=self._credentials
```
sryza, line 52: These changes look irrelevant. Revert? Same with the other ones below.

sryza (approved):

examples/quickstart_gcp/quickstart_gcp/io_managers.py
smackesey: Ah, these weren't intended-- this is a weird effect of black pre-commit hook we should fix. Black resolves config  to the nearest common ancestor config of all formatted files. When all target files are in a single example, black will resolve to the `pyproject.toml` of the example, which doesn't have our line-length modification (found in our root pyproject.toml). So it shortens these lines. When black ran on pre-commit for this PR this is what happened, since all files are in this example.

PR #11914, opened 2023-01-27
[dagit] Show partition status on the asset graph, catalog pages, add tests

bengotow:
### Summary & Motivation

image

image

Looked at a wide range of assets in the asset graph, updated storybook

image

Also added storybooks for all of the partner tags 

image

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
@@ -2,11 +2,13 @@ import {gql} from '@apollo/client';
 import {Colors, Icon, FontFamily, Box, Caption, Spinner} from '@dagster-io/ui';
 import isEqual from 'lodash/isEqual';
 import React from 'react';
+import {Link} from 'react-router-dom';
 import styled from 'styled-components/macro';
 
 import {withMiddleTruncation} from '../app/Util';
 import {humanizedLateString, isAssetLate} from '../assets/CurrentMinutesLateTag';
 import {isAssetStale} from '../assets/StaleTag';
+import {assetDetailsPathForKey} from '../assets/assetDetailsPathForKey';
 import {AssetComputeKindTag} from '../graph/OpTags';
 import {TimestampDisplay} from '../schedules/TimestampDisplay';
 import {markdownToPlaintext} from '../ui/markdownToPlaintext';
@@ -21,12 +23,6 @@ export const AssetNode: React.FC<{
   liveData?: LiveDataForNode;
   selected: boolean;
 }> = React.memo(({definition, selected, liveData}) => {
-  const firstOp = definition.opNames.length ? definition.opNames[0] : null;
-
-  // Used for linking to the run with this step highlighted. We only support highlighting
-  // a single step, so just use the first one.
-  const stepKey = firstOp || '';
-
   const displayName = definition.assetKey.path[definition.assetKey.path.length - 1];
   const isSource = definition.isSource;
 
@@ -60,7 +56,10 @@ export const AssetNode: React.FC<{
                   <Caption style={{textAlign: 'right'}}>
                     <AssetRunLink
                       runId={liveData.lastObservation.runId}
-                      event={{stepKey, timestamp: liveData.lastObservation.timestamp}}
+                      event={{
+                        stepKey: getStepKey(definition),
+                        timestamp: liveData.lastObservation.timestamp,
+                      }}
                     >
                       <TimestampDisplay
                         timestamp={Number(liveData.lastObservation.timestamp) / 1000}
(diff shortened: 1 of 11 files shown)


bengotow:
Edit: per discussion with Sandy we're changing the UI a bit, back to the original version in which the data is combined on a single row. 

Partitioned assets cannot be meaningfully "failed" or "stale", so those cases are checked further down in the logic. Partitioned assets can be meaningfully "late", which introduces an interesting edge case. For that, we show them in yellow and the lateness replaces the number of missing partitions.

I also refactored this code so the minified asset node and the large asset node share all this business logic - it was getting out of control.

image
 

image
image

sryza:
@bengotow this looks good to me.

One thing - do we normally show Late as red? If so, we should do that too here, right?

clairelin135 (commented):
This also looks great to me! Will leave it to @hellendag to review the JS

bengotow:
Good call Sandy, changed the Late state to be red:

image

js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
```diff
@@ -131,22 +157,53 @@ export const AssetNodeStatusRow: React.FC<{
...
+  if (liveData.partitionStats) {
+    const {numPartitions, numMaterialized} = liveData.partitionStats;
+    const numMissing = numPartitions - numMaterialized;
+    return {
+      background: late ? Colors.Red50 : numMissing ? Colors.Yellow50 : Colors.Green50,
+      border: late ? Colors.Red500 : numMissing ? Colors.Yellow500 : Colors.Green500,
+      content: (
+        <>
+          <Caption color={late ? Colors.Red700 : numMissing ? Colors.Yellow700 : Colors.Green700}>
+            {`${numPartitions.toLocaleString()} Partitions`}
```
hellendag, line 184: - lowercase `partitions`
- Do we need a singular `1 partition` case?

js_modules/dagit/packages/core/src/workspace/VirtualizedAssetRow.tsx
```diff
@@ -205,6 +210,20 @@ export const VirtualizedAssetRow = (props: AssetRowProps) => {
...
+const AssetPartitionStatsText: React.FC<{
+  stats: {numMaterialized: number; numPartitions: number};
+}> = ({stats}) => {
+  const {numMaterialized, numPartitions} = stats;
+  const numMissing = numPartitions - numMaterialized;
+  return (
+    <span>
+      {numMissing > 0
+        ? `${numPartitions.toLocaleString()} Partitions (${numMissing.toLocaleString()} missing)`
+        : `${numPartitions.toLocaleString()} Partitions`}
```
hellendag, line 222: Same notes as above

js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
```diff
@@ -100,24 +99,51 @@ export const AssetNodeStatusBox: React.FC<{background: string}> = ({background,
...
+  // Used for linking to the run with this step highlighted. We only support highlighting
+  // a single step, so just use the first one.
+  const firstOp = definition.opNames.length ? definition.opNames[0] : null;
+  return firstOp || '';
+}
+
+export function buildAssetNodeStatusRow({
+  definition,
+  liveData,
+}: {
```
hellendag, line 122: Might be worth putting together a quick test for this, at least for any of the more complex conditions.

hellendag (approved):

js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
bengotow: Hmm good call re: 1 partition case -- it does feel a bit odd to uppercase it, I'll run this by josh

js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
bengotow: I'm a little torn about this -- there are storybooks for all of these cases and I feel like the errors are most likely going to be visual ones. Having tests eg: "expect it to say Materialized" would be a nice added layer but also give a false sense of security -- I wouldn't want someone to change this, see that the tests still pass, and merge it without actually looking at the storybooks for visual correctness...

I wonder if we could make some sort of PR review checkbox ala: "[] I have looked at the storybooks for this"?

bengotow: Update: I added tests for all the storybook cases. Just simple ones that verify the correct status text is displayed, but I think it'll keep bugs out of this.

bengotow:
PR failure is deep in some socket connection in python, going to ignore and merge.

PR #11919, opened 2023-01-28
Add `last_materialization_record` to AssetEntry.

OwenKephart:
### Summary & Motivation

This PR converts the AssetEntry class to contain a `last_materialization_record` property (which contains the storage_id of the latest materialization in addition to the the event_log_entry itself).

To do this, we begin to serialize the entire `EventLogRecord` in the `last_materialization` column of the `AssetKeys` table. Of course, rows added to this table before this change will still have an `EventLogEntry` (or even `AssetMaterialization`!) in that column. In these cases, we default to the backcompat query, which will manually query for events from the event log in those cases.


python_modules/dagster/dagster/_core/storage/event_log/base.py
```diff
@@ -112,13 +114,23 @@ def __new__(
...
+            last_storage_id=check.opt_int_param(last_storage_id, "last_storage_id"),
             last_run_id=check.opt_str_param(last_run_id, "last_run_id"),
             asset_details=check.opt_inst_param(asset_details, "asset_details", AssetDetails),
             cached_status=check.opt_inst_param(
                 cached_status, "cached_status", AssetStatusCacheValue
             ),
         )
 
+    @property
+    def last_materialization_record(self) -> Optional[EventLogRecord]:
```
OwenKephart, line 126: this whole thing is a little weird, and I could be convinced to do this the other way around (just pass in the last_materialization_record directly to this constructor).

The reason I didn't do it this way is that this would require a bit more messing around with the GraphQL Instance stuff for DagsterCloud, but that's not a great reason.

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1010,22 +1021,22 @@ def _construct_asset_record_from_row(
...
+        results: Dict[AssetKey, Optional[EventLogRecord]] = {}
         for row in raw_asset_rows:
             asset_key = AssetKey.from_db_string(row[1])
             if not asset_key:
                 continue
             event_or_materialization = (
                 deserialize_json_to_dagster_namedtuple(row[2]) if row[2] else None
             )
-            if isinstance(event_or_materialization, EventLogEntry):
+            if isinstance(event_or_materialization, EventLogRecord):
```
clairelin135, line 1039: I think if the object is an event log entry we don't need to do the backcompat fetch to refetch the event. I think we can just only fetch the storage ID and combine it with the event log entry from the asset row

clairelin135 (approved):
Had one suggestion, but otherwise this looks good to me.

Can we also add a backcompat test to confirm that if the entry written into the asset key table is a event log record we'll do a fetch to get the storage id? Similar to this: https://github.com/dagster-io/dagster/blob/d1aa8276eb865fee12e3c9f5bfa5a8cbe89812be/python_modules/dagster/dagster_tests/storage_tests/test_assets.py#L81

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
OwenKephart: This is definitely true, I just figured that this would add a lot more surface area to this change -- i'd have to keep track of two separate sets of things (the ones with just AssetMaterializations, and the ones with EventLogEntries), then have two separate backcompat queries. Just seemed like a lot of opportunity for error / more code to maintain, but interested if you have a different take

OwenKephart:
@clairelin135 good call on the backcompat test -- did the sequence of events described in one of the comments on a fresh environment with the newest version of dagster (1.1.14). Updated a couple tests to look at both the 0.11.0 and this version.

prha:
Should we consider doing a write on read for the backcompat fetch? 

Or add an optional data migration that OSS users can apply if perf looks really bad?

PR #11950, opened 2023-01-29
Add code versions to AssetGraph

smackesey:
### Summary & Motivation

Add code versions to `AssetGraph`.

### How I Tested These Changes

Modify exsiting unit test.

PR #11951, opened 2023-01-29
CachingProjectedLogicalVersionResolver -> CachingStaleStatusResolver

smackesey:
### Summary & Motivation

Significant refactor of `CachingProjectedLogicalVersionResolver` and its GQL proxy `ProjectedLogicalVersionLoader`:

- New `StaleStatus` enum to indicate stale status for an asset. Three values:
    - FRESH (currentLogicalVersion == projectedLogicalVersion)
    - STALE (currentLogicalVersion != projectedLogicalVersion)
    - UNKNOWN (projectedLogicalVersion is unknown)
- `CachingProjectedLogicalVersionResolver` is now `CachingStaleStatusResolver`. It allows retrieval of current logical version, projected logical version, or stale status for a key.
- `CachingStaleStatusResolver` uses `ExternalAssetGraph` instead of the previous `Mapping[AssetKey, ExternalAssetNode]`
- `ProjectedLogicalVersionLoader` is gone-- GQL resolvers now directly use `CachingStaleStatusResolver`

Anytime we want stale status without other, you can now construct a `CachingStaleStatusResolver` and just call `get_status` (cc @owenkephart). In this PR I updated the stale schedule/sensor logic to use this.

This doesn't hit any public APIs, so existing BK tests.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/implementation/fetch_assets.py
+++ b/python_modules/dagster-graphql/dagster_graphql/implementation/fetch_assets.py
@@ -10,7 +10,9 @@ from dagster import (
     _check as check,
 )
 from dagster._core.definitions.asset_graph import AssetGraph
+from dagster._core.definitions.external_asset_graph import ExternalAssetGraph
 from dagster._core.definitions.freshness_policy import FreshnessPolicy
+from dagster._core.definitions.logical_version import CachingStaleStatusResolver
 from dagster._core.events import ASSET_EVENTS
 from dagster._core.host_representation.external import ExternalRepository
 from dagster._core.host_representation.external_data import (
@@ -23,7 +25,6 @@ from dagster._utils.caching_instance_queryer import CachingInstanceQueryer
 
 from dagster_graphql.implementation.loader import (
     CrossRepoAssetDependedByLoader,
-    ProjectedLogicalVersionLoader,
 )
 
 if TYPE_CHECKING:
@@ -130,7 +131,9 @@ def get_asset_node_definition_collisions(
     return results
 
 
-def get_asset_nodes_by_asset_key(graphene_info) -> Mapping[AssetKey, "GrapheneAssetNode"]:
+def get_asset_nodes_by_asset_key(
+    graphene_info: "HasContext",
+) -> Mapping[AssetKey, "GrapheneAssetNode"]:
     """
     If multiple repositories have asset nodes for the same asset key, chooses the asset node that
     has an op.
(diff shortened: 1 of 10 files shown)


smackesey:
* master
  * **PR #11950** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11950" target="_blank">Graphite</a> 
    * **PR #11951** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11951" target="_blank">Graphite</a>  👈
      * **PR #11952** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11952" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11951?utm_source=stack-comment).

sryza (commented):
I love it. My only comment is: it seems like we use the term "loader" in some places and "resolver" in other. Should we standardize?

smackesey:
>My only comment is: it seems like we use the term "loader" in some places and "resolver" in other. Should we standardize?

"Loader" is an idiom used in the context of GQL resolver functions for an intermediate object that prevents the multiple database hits (n+1 problem). I changed it to "resolver" in core because (a) we're no longer necessarily in the GQL context; (b) "loader" is a bit misleading because we're actually computing new logical versions, not just hitting the DB. It is true that the resolver is _also_ used in GQL, but since it's defined in core I think resolver is better.

sryza:
In that case, can we change `def stale_status_loader` to `def stale_status_resolver`?

smackesey:
>In that case, can we change def stale_status_loader to def stale_status_resolver?

The problem with this is that in the GQL context "resolver" means something else, so that could be confusing.

I just pushed a commit that aliases `CachingStaleStatusResolver` to `StaleStatusLoader` where the other GQL loaders are defined. Then GQL code imports this alias instead of the `CachingStaleStatusResolver` directly.

sryza (approved):
The aliasing feels a little fishy to me as well - like is that something we would do if we had built CachingStaleStatusResolver first and then exposed a GraphQL API over it later?

Maybe CachingStateStatusQueryer.

That all said, not a big deal and easy to change later if we want - the current approach seems fine and happy to defer to you on this one.

smackesey:
>The aliasing feels a little fishy to me as well - like is that something we would do if we had built CachingStaleStatusResolver first and then exposed a GraphQL API over it later?

I actually think so-- it keeps it clear that within GQL that we're using a batch loader, with the `CachingStateStatusResolver` being an implementation detail.

PR #11952, opened 2023-01-29
Update GQL to expose StaleStatus and StaleStatusCause

smackesey:
### Summary & Motivation

This updates GQL to expose the new StaleStatus and StaleStatusCause data.

### How I Tested These Changes

Unit tests (in progress)

PR #11953, opened 2023-01-30
Add stale status causes

smackesey:
### Summary & Motivation

This adds a "stale status lineage" API to the`CachingStaleStatusResolver`. A stale status lineage is a sequence of "causes", where a cause is a NamedTuple having:

- status: the status the cause produces
- path: sequence of asset keys starting from the target node, proceeding to the terminal node of the cause
- reason: a string description of the reason for the status

Its possible the format of "cause" will change.

The lineage is intended for manipulation/display in dagit, where it can be used to provide a rich display of the reason for staleness. Alternative, the first cause can be taken for an abbreviated display.

Unit tests (in progress)

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/logical_version.py
+++ b/python_modules/dagster/dagster/_core/definitions/logical_version.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from enum import Enum
 from hashlib import sha256
-from typing import TYPE_CHECKING, Callable, Mapping, NamedTuple, Optional, Union
+from typing import TYPE_CHECKING, Callable, Iterator, Mapping, NamedTuple, Optional, Sequence, Union
 
 from typing_extensions import Final
 
@@ -10,19 +10,23 @@ from dagster import _check as check
 from dagster._utils.cached_method import cached_method
 
 if TYPE_CHECKING:
+    from dagster._core.definitions.asset_graph import AssetGraph
     from dagster._core.definitions.events import (
         AssetKey,
         AssetMaterialization,
         AssetObservation,
         Materialization,
     )
-    from dagster._core.definitions.external_asset_graph import ExternalAssetGraph
     from dagster._core.events.log import EventLogEntry
     from dagster._core.instance import DagsterInstance
 
 
 class UnknownValue:
     pass
+
+
+def foo(x):
+    return False
 
 
 UNKNOWN_VALUE: Final[UnknownValue] = UnknownValue()
@@ -194,7 +198,7 @@ def _extract_event_data_from_entry(
 
 
 # ########################
-# ##### CACHING STALE STATUS RESOLVER
+# ##### STALENESS OPERATIONS
 # ########################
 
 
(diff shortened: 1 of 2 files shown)


smackesey:
* master
  * **PR #11950** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11950" target="_blank">Graphite</a> 
    * **PR #11951** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11951" target="_blank">Graphite</a> 
      * **PR #11953** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11953" target="_blank">Graphite</a>  👈
        * **PR #11952** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11952" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11953?utm_source=stack-comment).

sryza (commented):
This is cool stuff. Could we just call it stale status causes to avoid the name collision between "lineage" and "asset lineage"?

python_modules/dagster/dagster/_core/definitions/logical_version.py
```diff
@@ -204,6 +204,15 @@ class StaleStatus(Enum):
     UNKNOWN = "UNKNOWN"
 
 
+class StaleStatusCause(NamedTuple):
+    status: StaleStatus
+    path: Sequence[AssetKey]
```
sryza, line 209: Why do we need the path vs. just the root asset? I'm thinking about cases where there's a diamond DAG and the asset at the root is stale and the leaf asset is stale for the same reason, by two different paths.

smackesey: >'m thinking about cases where there's a diamond DAG and the asset at the root is stale and the leaf asset is stale for the same reason, by two different paths.

That's a good point, I hadn't thought about duplicates.

>Why do we need the path vs. just the root asset?

I included it thinking it would make it easier to e.g. highlight the path on the dagit asset graph to the cause of a problem. The information is easy to get server-side and will probably need to be computed anyway in the client.

Come to think of it I'm not including multiple paths with the same root is a bad thing. I am imagining you are hovering over different causes in the upstream sidebar, with a level of processing that shows only each node, but when you hover it highlights all paths from the problem node to the target.

sryza: I agree that highlighting the path in Dagit would be a good experience. I think we might have enough information on the frontend though to render it.  Thoughts @bengotow ? 

My concern is that, if there's a lot of fan-out and fan-in in a data pipeline, there can be a combinatorial explosion of paths (e.g. 2^n_assets for a helix-shaped pipeline).

python_modules/dagster/dagster/_core/definitions/logical_version.py
```diff
@@ -204,6 +208,15 @@ class StaleStatus(Enum):
     UNKNOWN = "UNKNOWN"
 
 
+class StaleStatusCause(NamedTuple):
+    status: StaleStatus
+    key: AssetKey
+    reason: str
+
+    # def with_path_step(self, key: AssetKey) -> "StaleStatusCause":
```
sryza, line 216: Take this out?

sryza (approved):
Nice

PR #11955, opened 2023-01-30
Delete k8s job on failure (timout or otherwise) in k8s_job_op

gibsondan:


johannkm (changes requested):
Maybe I'm missing context- always deleting on failure seems like it would make it hard to track down errors for users who aren't exporting to datadog/etc.?

Can users use the native ttl setting for this?

PR #11972, opened 2023-01-30
[dagster-fivetran] Add option to force-create materializations for tables not in API response

benpankow:
## Summary

Occasionally Fivetran seems not to include tables associated with a sync in the API response if no rows are actually synced. This can cause Dagster to error because no asset materialization is output when one is expected. See https://dagster.slack.com/archives/C01U954MEER/p1674732294303889 for report.

Adds a new `materialize_missing_tables` parameter which creates asset materializations for all tables that a user expects to sync, even if Fivetran does not return API information on those tables after a sync completes. 

## Test Plan

Added a unit test of this behavior.

### DIFF
--- a/python_modules/libraries/dagster-fivetran/dagster_fivetran/asset_defs.py
+++ b/python_modules/libraries/dagster-fivetran/dagster_fivetran/asset_defs.py
@@ -42,6 +42,7 @@ def _build_fivetran_assets(
     table_to_asset_key_map: Optional[Mapping[str, AssetKey]] = None,
     resource_defs: Optional[Mapping[str, ResourceDefinition]] = None,
     group_name: Optional[str] = None,
+    materialize_missing_tables: bool = False,
 ) -> Sequence[AssetsDefinition]:
     asset_key_prefix = check.opt_sequence_param(asset_key_prefix, "asset_key_prefix", of_type=str)
 
@@ -75,6 +76,8 @@ def _build_fivetran_assets(
             poll_interval=poll_interval,
             poll_timeout=poll_timeout,
         )
+
+        materialized_asset_keys = set()
         for materialization in generate_materializations(
             fivetran_output, asset_key_prefix=asset_key_prefix
         ):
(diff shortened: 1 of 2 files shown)


benpankow:
* master
  * **PR #11972** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11972" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11972?utm_source=stack-comment).

bvujanic:
Note that I've reached out to fivetran around this particular bug on their end and right now their response is 
```
Hi Bojan,
 
I just wanted to followup on this. I've received a preliminary response from the engineering team. The had this to say:
 

    The metadata API only returns tables when some rows were synced from the source. This was desired behaviour originally.

 
This doesn't sound like what you were seeing though. This has been assigned to the engineer who made the change to see if there could have been an impact here. I'll let you know once they have had a chance to fully investigate.
 
Thanks,
 
Seamus
```
Which doesn't make sense to me as it's really hard to distinguish the no rows synced vs sync failed.

The fix above should help, but imo this ought to be patched in fivetrans api.

python_modules/libraries/dagster-fivetran/dagster_fivetran/asset_defs.py
```diff
@@ -129,6 +143,10 @@ def build_fivetran_assets(
             table name to user-supplied metadata that should be associated with the asset for that table.
         group_name (Optional[str]): A string name used to organize multiple assets into groups. This
             group name will be applied to all assets produced by this multi_asset.
+        materialize_missing_tables (bool): If True, will create asset materializations for tables specified
```
OwenKephart, line 146: kinda confusing name, as this option does not actually change what gets materialized. instead, it just creates events. maybe something like "infer_missing_tables" or something like that? hard to think of a not totally-verbose name

OwenKephart (approved):

PR #11974, opened 2023-01-30
remove protobuf pin in dagster[test]

yuhan:
### Summary & Motivation

this PR removes the unnecessary pin in `dagster[test]` since we are already pinning `protobuf>=3.13.0,<4` in `install_requires`

This PR also skips the last bit of `dagster_tests/execution_tests/engine_tests/test_step_delegating_executor.py::test_execute_intervals` because if step finishes fast enough before the final `check_step_health` is called, the count could be smaller than 3, so we may need to find a better way to test that. This test passed in local env.

https://buildkite.com/dagster/dagster/builds/47202#01860504-8355-49a0-877b-4dc9d0a08fb5/6-34

### DIFF
--- a/python_modules/dagster/dagster_tests/execution_tests/engine_tests/test_step_delegating_executor.py
+++ b/python_modules/dagster/dagster_tests/execution_tests/engine_tests/test_step_delegating_executor.py
@@ -68,12 +68,10 @@ class TestStepHandler(StepHandler):
 
     def check_step_health(self, step_handler_context) -> CheckStepHealthResult:
         TestStepHandler.check_step_health_count += 1
-        print("HERE!!!", "check_step_health")
         return CheckStepHealthResult.healthy()
 
     def terminate_step(self, step_handler_context):
         TestStepHandler.terminate_step_count += 1
-        print("HERE!!!", "terminate_step")
         raise NotImplementedError()
 
     @classmethod
@@ -193,7 +191,6 @@ def test_skipping():
 
 
 def test_execute_intervals():
-    print("??")
     TestStepHandler.reset()
     with instance_for_test() as instance:
         result = execute_pipeline(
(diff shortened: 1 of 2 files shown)


yuhan:
* master
  * **PR #11974** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11974" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11974?utm_source=stack-comment).

python_modules/dagster/dagster_tests/execution_tests/engine_tests/test_step_delegating_executor.py
```diff
@@ -220,6 +223,8 @@ def test_execute_intervals():
     assert TestStepHandler.terminate_step_count == 0
     # every step should get checked at least once
     assert TestStepHandler.check_step_health_count >= 3
```
yuhan, line 225: @jmsanders @johannkm i'm trying to remove the protobuf pin from `dagster[test]` but this assertion fails when i upgrade it to 3.20.3 ([bk log](https://buildkite.com/dagster/dagster/builds/47456#01860a41-3aa5-417c-bb26-6d5a3b78689a/6-34)). 

i can't repro this in my local, which passes all the tests. git blame led me to https://dagster.phacility.com/D8218 where you two discussed about this assertion. do either of you know of why that happens and if we could modify the test so we drop the pin?

python_modules/dagster/dagster_tests/execution_tests/engine_tests/test_step_delegating_executor.py
```diff
@@ -68,10 +68,12 @@ def launch_step(self, step_handler_context):
 
     def check_step_health(self, step_handler_context) -> CheckStepHealthResult:
         TestStepHandler.check_step_health_count += 1
+        print("HERE!!!", "check_step_health")
```
yuhan, line 71: this step only ran twice on bk while it was run three times in my local env

gibsondan:
what an odd bug... could relaxing the pin have caused some other library to change too?

gibsondan:
Nah its just protobuf that updated. My best guess would be that this new version made the step finish slightly faster so the last step actually completes before it does the final check_step_health call (i.e. there was an existing race condition in the test that this uncovered)

gibsondan:
i.e. I can see how this would happen if the step finishes fast enough that the success event gets popped off the queue here before the health check part below fires: https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_core/executor/step_delegating/step_delegating_executor.py#L225-L239

Given that skipping the test seems like an OK workaround to me, but will defer to @johannkm on that

gibsondan (approved):
accepting contingent on skipping test that's failing /removing comments

PR #11976, opened 2023-01-30
[dagster-dbt] Add a stream_events option to `load_assets_from_dbt*` 

OwenKephart:
### Summary & Motivation

We want the ability to emit events from the underlying dbt execution as they happen, rather than needing to wait until the entire command has finished. Ideally, this would become the default behavior (at least for users running dbt >= 1.4.0), as dbt has added a good amount of structured logging which makes it mostly useless to wait until the end of execution to emit events.


OwenKephart:
closing this in favor of cleaned up prs

PR #11982, opened 2023-01-31
Rename `_core.decorator_utils.is_context_provided`

### LINKED ISSUE #11980: Two copies of is_context_provided in the code base
### Dagster version

main

### What's the issue?

There are two copies of `is_context_provided` in the codebase. One is there focused on type correctness in python_modules/dagster/dagster/_core/decorator_utils.py. However there is another one which actually inspects the name of the parameters to ensure that they are valid context names in `python_modules/dagster/dagster/_core/definitions/decorators/solid_decorator.py` These should be reconciled and have standardized behavior. If two functions are different, they should be named differently and documented.

_No response_

### How to reproduce?

_No response_

### Deployment type

None

### Deployment details

_No response_

### Additional information

_No response_

smackesey:
### Summary & Motivation

Fixes #11980

From `_core.decorators_utils`: `is_context_provided` -> `has_at_least_one_parameter`

Existing BK tests

### DIFF
--- a/python_modules/dagster/dagster/_core/decorator_utils.py
+++ b/python_modules/dagster/dagster/_core/decorator_utils.py
@@ -87,6 +87,14 @@ def format_docstring_for_description(fn: Callable) -> Optional[str]:
 # When/if `StrictTypeGuard` is supported, we can drop `is_context_not_provided` since a False from
 # `is_context_provided` will be sufficient.
 def is_context_provided(
-    fn: Union[Callable[Concatenate[T, P], R], Callable[P, R]]
+    fn: Union[Callable[Concatenate[T, P], R], Callable[P, R]],
+    valid_names: Optional[Sequence[str]] = None,
 ) -> TypeGuard[Callable[Concatenate[T, P], R]]:
-    return len(get_function_params(fn)) >= 1
+    params = get_function_params(fn)
+    if valid_names is None:
+        return len(params) >= 1
+    else:
+        valid_name_permutations = [
+            x for name in valid_names for x in get_valid_name_permutations(name)
+        ]
+        return len(params) >= 1 and params[0].name in valid_name_permutations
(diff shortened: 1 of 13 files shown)


smackesey:
* master
  * **PR #11982** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11982" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11982?utm_source=stack-comment).

python_modules/dagster/dagster/_core/decorator_utils.py
```diff
@@ -87,6 +87,14 @@ def format_docstring_for_description(fn: Callable) -> Optional[str]:
...
 ) -> TypeGuard[Callable[Concatenate[T, P], R]]:
-    return len(get_function_params(fn)) >= 1
+    params = get_function_params(fn)
+    if valid_names is None:
+        return len(params) >= 1
+    else:
+        valid_name_permutations = [
+            x for name in valid_names for x in get_valid_name_permutations(name)
+        ]
+        return len(params) >= 1 and params[0].name in valid_name_permutations
```
schrockn, line 100: This doesn't really fix one of the problems here, which is that the function here is named inaccurately. What it means now is that "return true if these is more than one argument of any name. If valid_names is set the function then returns true if param name is raw underscore. or one of the valid names, or one the valid names preceded by a "_" or suffixed with "_".

This is an extremely complicated semantic and seems like a lot for a generic function. In particular that it takes multiple `valid_names` makes this particularly problematic and hard to understand.

I recommend that we instead change the name of this function from `is_context_provided` to `has_at_least_one_parameter` so that the name is _accurate_. Then all the callsites to this will make more sense where they have errors conditions that multiple arguments are illegal.

schrockn (changes requested):

python_modules/dagster/dagster/_core/decorator_utils.py
smackesey: I agree a generic function should have a different name. It's unclear to me from this though whether you're asking for:

- Two functions to be preserved with the `decorator_utils` one renamed
- The consolidated function to be renamed

schrockn: I think for now have two sep functions with two names for clarity. Changing the naming constraints is an entire other matter that would involve a deprecation cycle, etc. Let's just make the current state of the world more clear.

schrockn (approved):
👍🏻

PR #11984, opened 2023-01-31
[dagster-snowflake-pandas] snowflake-sqlalchemy pin

smackesey:
### Summary & Motivation

Replace upper bound pin on `snowflake-connector-python` with lower bound pin on `snowflake-sqlalchemy`-- a lower bound pin on an old version will cause fewer conflicts than capping a recent version.

This also lets us remove the guide pin on `assets_smoke_test`.

BK


alangenfeld (approved):
nice, thanks for following up and improving this

PR #11986, opened 2023-01-31
Fix doc op_version used instead of code_version

NicolasPA:
### Summary & Motivation

Fix a simple mistake in the documentation where the parameter name `op_version` is used when it should have been `code_version`.
Page: https://docs.dagster.io/guides/dagster/scheduling-assets#step-5-change-code-versions


smackesey (approved):

NicolasPA:
This is now fixed.

PR #11987, opened 2023-01-31
add .ruff_cache to gitignore

Ramshackle-Jamathon:
it would also be nice to get a vs-code config example of ruff+black format on save

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,6 +98,9 @@ Pipfile.lock
 # mkdocs documentation
 /site
 
+# ruff
+.ruff_cache/
+
 # mypy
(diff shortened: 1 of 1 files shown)


smackesey:
Hmm is `ruff` creating a cache in the repo root? I wonder why I'm not seeing it.

>it would also be nice to get a vs-code config example of ruff+black format on save

Maybe post in `#eng-help`, I'm sure many people have this set up (I'm not a VSCode user)

Ramshackle-Jamathon:
I've been running `ruff . --fix` from a variety of directories and the ruff_cache gets created

PR #11991, opened 2023-01-31
[asset-reconciliation][perf] Improve prefetch accuracy

OwenKephart:
### Summary & Motivation

base: https://github.com/dagster-io/dagster/pull/11919

Once last_materialization_record is a property of AssetEntry, it becomes possible to fetch EventLogRecords for a large set of asset keys at once, rather than resorting to many individual queries. This dramatically improves performance in cases where the old "hack" of fetching the last N asset materialization records did not do a good job of grabbing the records we actually needed. It's hard to get an accurate accounting here, as the impact of this can vary anywhere from "basically nothing" to "a lot", but in one test I did this took the execution time of the sensor down from 57s to 3s


clairelin135 (approved):
Glad to see this!

PR #11993, opened 2023-01-31
Add resource support to observable source assets

smackesey:
### Summary & Motivation

Allows provision of resources to observable source assets via either `required_resource_keys` (old-style) or Pythonic resource args.

Threading `required_resource_keys` through both `@observable_source_asset` and `SourceAssetDefinition` was awkward. If we allow config schema for observation functions (which we should) it will only get worse. The solution used here is to make `_required_resource_keys` a private arg on `SourceAsset` to avoid committing on some API questions discussed below.

Another possible approach would be to allow source asset to take a `node_def` (matching the `AssetsDefinition` API) instead of an `observe_fn`. That would technically be a break we wouldn't normally make, since `observe_fn` is not marked as an experimental arg on `SourceAsset`-- but I'd argue it is widely understood to be experimental since `@observable_source_asset` is, which is probably what anyone using this feature is using anyway.

New unit tests

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/source_asset.py
+++ b/python_modules/dagster/dagster/_core/definitions/source_asset.py
@@ -1,7 +1,16 @@
 from __future__ import annotations
 
 import warnings
-from typing import TYPE_CHECKING, Dict, Iterator, Mapping, Optional, Union, cast
+from typing import (
+    TYPE_CHECKING,
+    AbstractSet,
+    Dict,
+    Iterator,
+    Mapping,
+    Optional,
+    Union,
+    cast,
+)
 
 from typing_extensions import Protocol, TypeAlias
 
@@ -22,6 +31,7 @@ from dagster._core.definitions.resource_requirement import (
     ResourceAddable,
     ResourceRequirement,
     SourceAssetIOManagerRequirement,
+    ensure_requirements_satisfied,
     get_resource_key_conflicts,
 )
 from dagster._core.definitions.utils import (
@@ -74,6 +84,7 @@ class SourceAsset(ResourceAddable):
             the asset when it's used as an input to other assets inside a job.
         io_manager_def (Optional[IOManagerDefinition]): (Experimental) The definition of the IOManager that will be used to load the contents of
             the asset when it's used as an input to other assets inside a job.
+        required_resource_keys (Optional[Set[str]]): Set of resource handles required by the observe op.
         resource_defs (Optional[Mapping[str, ResourceDefinition]]): (Experimental) resource definitions that may be required by the :py:class:`dagster.IOManagerDefinition` provided in the `io_manager_def` argument.
         description (Optional[str]): The description of the asset.
         partitions_def (Optional[PartitionsDefinition]): Defines the set of partition keys that
@@ -104,6 +115,7 @@ class SourceAsset(ResourceAddable):
         group_name: Optional[str] = None,
         resource_defs: Optional[Mapping[str, ResourceDefinition]] = None,
         observe_fn: Optional[SourceAssetObserveFunction] = None,
+        required_resource_keys: Optional[AbstractSet[str]] = None,
         # Add additional fields to with_resources and with_group below
     ):
         if resource_defs is not None:
@@ -142,6 +154,9 @@ class SourceAsset(ResourceAddable):
         self.group_name = validate_group_name(group_name)
         self.description = check.opt_str_param(description, "description")
         self.observe_fn = check.opt_callable_param(observe_fn, "observe_fn")
+        self._required_resource_keys = check.opt_set_param(
+            required_resource_keys, "required_resource_keys", of_type=str
+        )
         self._node_def = None
 
     def get_io_manager_key(self) -> str:
@@ -192,6 +207,11 @@ class SourceAsset(ResourceAddable):
 
         return DecoratedOpFunction(fn)
 
+    @public
+    @property
+    def required_resource_keys(self) -> AbstractSet[str]:
+        return {requirement.key for requirement in self.get_resource_requirements()}
+
     @property
     def node_def(self) -> Optional[OpDefinition]:
         """Op that generates observation metadata for a source asset."""
@@ -203,6 +223,7 @@ class SourceAsset(ResourceAddable):
                 compute_fn=self._get_op_def_compute_fn(self.observe_fn),
                 name=self.key.to_python_identifier(),
                 description=self.description,
+                required_resource_keys=self._required_resource_keys,
             )
         return self._node_def
 
@@ -221,6 +242,10 @@ class SourceAsset(ResourceAddable):
 
         merged_resource_defs = merge_dicts(resource_defs, self.resource_defs)
 
+        # Ensure top-level resource requirements are met - except for
+        # io_manager, since that is a default it can be resolved later.
+        ensure_requirements_satisfied(merged_resource_defs, list(self.get_resource_requirements()))
+
         io_manager_def = merged_resource_defs.get(self.get_io_manager_key())
         if not io_manager_def and self.get_io_manager_key() != DEFAULT_IO_MANAGER_KEY:
             raise DagsterInvalidDefinitionError(
@@ -228,7 +253,7 @@ class SourceAsset(ResourceAddable):
                 f" '{self.get_io_manager_key()}', but none was provided."
             )
         relevant_keys = get_transitive_required_resource_keys(
-            {self.get_io_manager_key()}, merged_resource_defs
+            {*self._required_resource_keys, self.get_io_manager_key()}, merged_resource_defs
         )
 
         relevant_resource_defs = {
@@ -254,6 +279,7 @@ class SourceAsset(ResourceAddable):
                 resource_defs=relevant_resource_defs,
                 group_name=self.group_name,
                 observe_fn=self.observe_fn,
+                required_resource_keys=self._required_resource_keys,
             )
 
     def with_group_name(self, group_name: str) -> "SourceAsset":
@@ -276,9 +302,12 @@ class SourceAsset(ResourceAddable):
                 group_name=group_name,
                 resource_defs=self.resource_defs,
                 observe_fn=self.observe_fn,
+                required_resource_keys=self._required_resource_keys,
             )
 
     def get_resource_requirements(self) -> Iterator[ResourceRequirement]:
+        if self.node_def is not None:
+            yield from self.node_def.get_resource_requirements()
         yield SourceAssetIOManagerRequirement(
             key=self.get_io_manager_key(), asset_key=self.key.to_string()
         )
--- a/python_modules/dagster/dagster/_core/definitions/source_asset.py
+++ b/python_modules/dagster/dagster/_core/definitions/source_asset.py
@@ -4,19 +4,20 @@ import warnings
 from typing import (
     TYPE_CHECKING,
     AbstractSet,
+    Any,
+    Callable,
     Dict,
     Iterator,
     Mapping,
     Optional,
-    Union,
     cast,
 )
 
-from typing_extensions import Protocol, TypeAlias
+from typing_extensions import TypeAlias
 
 import dagster._check as check
 from dagster._annotations import PublicAttr, public
-from dagster._core.decorator_utils import has_at_least_one_parameter
+from dagster._core.decorator_utils import get_function_params
 from dagster._core.definitions.data_version import DATA_VERSION_TAG, DataVersion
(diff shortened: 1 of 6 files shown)


smackesey:
* master
  * **PR #11993** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11993" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11993?utm_source=stack-comment).

sryza (commented):
In the future we might want to introduce `@loadable_source_asset`:

```python
@loadable_source_asset
def my_csv_file():
    return pd.read_csv("my_csv_file.csv")
```

It's possible that a loadable source asset would also be observable.

The reason this is relevant to this PR: would we want the same set of required resource keys for both? If not, then maybe we should name this `observation_fn_required_resource_keys` or something?

On the other hand, it's likely that the set of required resource keys would be the same.  E.g. if your source asset is an S3 file, you likely need an S3 resource both to load it and to observe it.

smackesey:
@sryza 

>In the future we might want to introduce @loadable_source_asset

This is yet one more reason I think we should have `SourceAsset` take a `node_def` (`observation_node_def`) instead of observation fn-- complexity is going to multiply on the `SourceAsset` signature otherwise. What do you think of this idea?

sryza:
> This is yet one more reason I think we should have SourceAsset take a node_def (observation_node_def?) instead of observation_fn-- complexity is going to multiply on the SourceAsset signature otherwise. What do you think of this idea?

I definitely see the advantage here, but I think the reason that I had concerns about this originally is still relevant - we might want to invoke inside a sensor function. Also, it opens up a fairly wide API surface area - e.g. you'd now be able to add a `code_version` to your `observation_fn`.

Curious if @schrockn has thoughts, as he's been thinking about this area a bit recently.

schrockn:
I am still in the early stages of digging more deeply into observable source assets and I was not privy to all of the discussions around their design and implementation. But with these first impressions, my general thinking is that they feel and act much more like "normal" assets, rather than be a separate concept with separate language. Suggestions of naming things like `observation_fn_required_resource_keys` are definitely red flags to me that we are complecting two ideas and complicating things unnecessarily, rather than building up via composable parts. Another example red flag is the conversation in Slack where we have to contort `AssetSelection` to deal with observable source assets and normal assets in very different ways.

We have a few notions of assets.

SDA: A definition, in code, of an asset that _should_ exist. This traditional SDA definition also incorporates some very opinionated semantics in the I/O manager system.
Observable Source Asset: A definition, in code, of an asset that we _believe_ exists, but want to do computation verify to that it exists and validate/assert claims about it.
Source Asset: A definition in code of an asset that I claim exists.
Loadable Source Asset (Hypothesized above) A special sort of observable asset that automatically makes a bunch of claims given an in-memory representation of that asset returned to Dagster.

I don't think observable source assets are distinctly different enough from normal SDAs to warrant a different concept. I think they are much more similar to traditional assets rather than source assets, actually, despite their name and my definitions above. 

Instead, I think we should slightly loosen the constraints on the traditional SDAs to allow for this use case. Namely, you should be able to write an SDA where the op that SDA refers to _only_ emits an AssetObservation with attached logical version data. It, by convention, would return `Nothing` and have no associated I/O manager. (Note: I believe that currently we do emit an AssetMaterialization for Nothing outputs, but I don't think that makes sense and we should change that)

This is also connected to some thinking and exploration I've been doing to support external execution environments where the "contract" between the Dagster orchestrator and transformation is much more about the event stream, rather than our in-memory Python constructs. The top-of-mind use cases for me right now are Spark jobs and external K8s pods in different docker images. We we should be able to launch into external environments that in turn do the actual compute and read/write to storage, and then either (a) have those environments write to event log directly or (b) have the orchestrating process to do it on their behalf in an easy-to-use, well-supported way.

What connects these two use cases––source assets and external compute environments––is that it is no longer the I/O manager that is controlling writing asset materializations and observations. Instead compute is just emitting those events and taking over responsibility for I/O.

In this way of thinking I/O managers are very and ergonomic convenient ways to build a graph of assets in a single environment that abide by the event stream rules, but are not required.

sryza:
@schrockn do you have thoughts about the specific implications of your point for our Python APIs and/or Dagit?

I agree with you that there's a lot of thematic resemblance between these categories and opportunity to make the experience more consistent.

That said, I think there's a fundamental difference between:
1. an asset whose existence we know about at definition time
2. an asset whose existence we know about at definition time AND that we know how to materialize

The latter type supports specific, important operations in Python and in Dagit that the former type does not, and consequently I think that overly blurring the distinction between the two risks degrading the user experience.

sryza:
To add some specificity, I think it's very important that, before taking an action to "run an asset", a user can tell whether doing so is going to materialize it or just observe it.

sryza:
> @schrockn do you have thoughts about the specific implications of your point for our Python APIs and/or Dagit?

@schrockn sorry I realized I didn't read some of your points in enough detail and that you did include specific implications.

> Note: I believe that currently we do emit an AssetMaterialization for Nothing outputs, but I don't think that makes sense and we should change that

IMO the way this works right now makes a lot of sense.  I think that making this change would be a strong hit to user ergonomics.  It would effectively require them to redundantly specify their asset key in situations where they currently only need to specify them once, which adds boilerplate and opportunity for error.

schrockn:
> IMO the way this works right now makes a lot of sense. I think that making this change would be a strong hit to user e ergonomics. It would effectively require them to redundantly specify their asset key in situations where they currently only need to specify them once, which adds boilerplate and opportunity for error.

Note I was suggesting that _only_ for `Nothing`, not in the general case. I don't want to sidetrack the convo here. I agree that when you are fully in I/O manager land, automatically emitting AssetMaterializations with the right asset key makes sense for all the reasons you mention.

sryza:
> Note I was suggesting that only for Nothing, not in the general case. I don't want to sidetrack the convo here. I agree that when you are fully in I/O manager land, automatically emitting AssetMaterializations with the right asset key makes sense for all the reasons you mention.

I understand and that's what I am explicitly arguing against

schrockn:
> For example, if A -> B, and I'm launching a run that observes both A and B, there's no reason to wait until A has been observed before observing B. However, if I'm launching a run that materializes both A and B, then I should wait until A has materialized to materialize B.

I'm a bit confused by this example as it implies that B has an upstream dependency on A, and observable source assets don't have upstream dependencies. I suppose that would apply if we optionally added observation functions to all assets.

Re: "important operations in Python and in Dagit" do you have other specific ones that you think are of paramount importance?

sryza:
> I'm a bit confused by this example as it implies that B has an upstream dependency on A, and observable source assets don't have upstream dependencies. I suppose that would apply if we optionally added observation functions to all assets.

Ah right, yes, exactly.

> Re: "important operations in Python and in Dagit" do you have other specific ones that you think are of paramount importance?

The first one that comes to mind is clicking the Materialize button. I think it would be a loss if the button had to say something generic like "run" and then you don't get to find out whether the asset has been materialized or observed until the step completes.

sryza:
Here's what our concept page currently says:

> An asset is an object in persistent storage, such as a table, file, or persisted machine learning model. A software-defined asset is a Dagster object that couples an asset to the function and upstream assets that are used to produce its contents.

I would be open to changing this to something like "materializable software-defined asset" if we think it makes the ontology more clear.

smackesey:
@sryza Thoughts on naming the `required_resource_keys` arg here `_required_resource_keys`, so that we can get this feature in before resolving the broader questions of source asset ontology?

sryza:
Suppose we merged the PR as-is and we later wanted to allow source assets to be loadable - would that require a breaking change?

schrockn (dismissed):
What is the pressing need for this right now?

When it comes to this observables source asset duality issue, I'm certainly in "contain the damage" mode until we have more clarity here. Supporting this, for example, adds another decorator that we have know how to inject resources into, increasing the surface area of resources v2, just an example, and adds another concept to the list of concepts that we have to bind resources to.

I'd prefer we step back before proceeding here.

sryza:
I don't feel particular urgency on landing this, but FWIW I feel quite confident that the future will contain both:
- SourceAssets with a built-in observation function
- The ability to supply resources to that function

smackesey:
>Suppose we merged the PR as-is and we later wanted to allow source assets to be loadable - would that require a breaking change?

It depends on how we implemented loadable source assets, but if we merged it as is there would be `required_resource_keys` exposed directly on `SourceAsset`. If we wanted to have separate loadable and observable ops with separate `required_resource_keys`, then we'd end up with either a breaking change or a confusing API (where `required_resource_keys` applies only to observation). That's why I support making it `_required_resource_keys` on `SourceAsset`-- so that we have flexibility to change it later.

>What is the pressing need for this right now?

Just that people expect resources to work with source asset observation right now, and they don't. I see this more as fixing a bug than adding a feature.

smackesey:
> Which "people"? What is their exact use case?

I've had a support request for it at least once. I assume the use case is to query a database to compute a `LogicalVersion`.

smackesey:
>We should make sure to support new-style resource passing to this as well.

OK, I modified this and added a test to make sure it handles pythonic resources.

python_modules/dagster/dagster_tests/asset_defs_tests/test_source_asset_observation_job.py
```diff
@@ -105,3 +108,97 @@ def bar(context):
             assets=[foo, bar],
             jobs=[define_asset_job("mixed_job", [foo, bar])],
         )
+
+
+@pytest.mark.parametrize(
+    "is_valid,resource_defs",
```
sryza, line 114: The fact that most of the logic in the function depends on this boolean makes me think that it would be clearer to split it into two separate tests, and just define the asset at module level. Not critical feedback though.

python_modules/dagster/dagster/_core/definitions/source_asset.py
```diff
@@ -74,6 +85,7 @@ class SourceAsset(ResourceAddable):
             the asset when it's used as an input to other assets inside a job.
         io_manager_def (Optional[IOManagerDefinition]): (Experimental) The definition of the IOManager that will be used to load the contents of
             the asset when it's used as an input to other assets inside a job.
+        required_resource_keys (Optional[Set[str]]): Set of resource handles required by the observe op.
```
sryza, line 88: If this is private, we shouldn't document it publicly, right?

python_modules/dagster/dagster/_core/definitions/source_asset.py
```diff
@@ -59,9 +71,8 @@ def __call__(self) -> DataVersion:
         ...
 
 
-SourceAssetObserveFunction: TypeAlias = Union[
-    SourceAssetObserveFunctionWithContext, SourceAssetObserveFunctionNoContext
-]
+# Going with this catch-all for the time-being to permit pythonic resources
```
sryza, line 74: We don't need the protocols above anymore in this case, right?

python_modules/dagster/dagster/_core/definitions/source_asset.py
```diff
@@ -104,6 +116,7 @@ def __init__(
         group_name: Optional[str] = None,
         resource_defs: Optional[Mapping[str, ResourceDefinition]] = None,
         observe_fn: Optional[SourceAssetObserveFunction] = None,
+        _required_resource_keys: Optional[AbstractSet[str]] = None,
```
sryza, line 119: I think worth mentioning why this has an underscore - i.e. that it's used for the observation function, but we're not sure what we want to do when an asset is both observable and can also be subject to other operations.

python_modules/dagster/dagster/_core/definitions/source_asset.py
```diff
@@ -192,6 +217,11 @@ def fn(context: OpExecutionContext):
 
         return DecoratedOpFunction(fn)
 
+    @public
+    @property
+    def required_resource_keys(self) -> AbstractSet[str]:
```
sryza, line 222: I thought we said this would be private?

python_modules/dagster/dagster_tests/asset_defs_tests/test_source_asset_observation_job.py
smackesey: can't do that because of the closure over `executed`

sryza: ah right - could have a shared function though. or it's a small enough amount of code it could just be copy/pasted. the `is_valid=False` one doesn't even need an implementation

sryza (approved):

PR #11994, opened 2023-01-31
[Dynamic Partitions 1] Storage changes

clairelin135:
Creates a dynamic_partitions table with columns:
- partitions_def_name: the name of the mutable partitions definition
- partition_key: a singular partition key that belongs to the partition def
For the above table, creates an index with [partitions_def_name, partition_key] to allow for faster querying for presence of a certain partition key for a partitions def

Adds methods to event log storage to read/write from this new table.


python_modules/dagster/dagster/_core/storage/event_log/base.py
```diff
@@ -361,6 +361,26 @@ def get_materialization_count_by_partition(
...
+        """Get the list of partition keys for a partition definition."""
+        raise NotImplementedError()
+
+    @abstractmethod
+    def has_partition(self, partitions_def_name: str, partition_key: str) -> bool:
+        """Check if a partition exists."""
+        raise NotImplementedError()
+
+    @abstractmethod
+    def add_partitions(self, partitions_def_name: str, partition_keys: Sequence[str]) -> None:
```
sryza, line 375: I think we should be consistent on add_mutable_partitions vs. add_partitions at the different layers

python_modules/dagster/dagster/_core/storage/event_log/schema.py
```diff
@@ -70,6 +70,17 @@
     db.Column("event_timestamp", db.types.TIMESTAMP),
 )
 
+
+MutablePartitionsTable = db.Table(
+    "mutable_partitions",
+    SqlEventLogStorageMetadata,
+    db.Column("id", db.Integer, primary_key=True, autoincrement=True),
+    db.Column("partitions_def_name", db.Text, nullable=False),
+    db.Column("partition_key", db.Text, nullable=False),
```
sryza, line 79: if all our other tables use `partition` instead of `partition_key`, we should use `partition` here too. otherwise, either is fine

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1622,6 +1623,69 @@ def get_materialization_count_by_partition(
...
+        return partition_key in self._fetch_partition_keys_for_partition_def(partitions_def_name)
+
+    def add_partitions(self, partitions_def_name: str, partition_keys: Sequence[str]) -> None:
+        if isinstance(partition_keys, str):
+            # Guard against a single string being passed in `partition_keys`
+            raise DagsterInvalidInvocationError("partition_keys must be a sequence of strings")
+        check.sequence_param(partition_keys, "partition_keys")
+
+        self._check_partitions_table()
+        existing_partitions = set(self.get_partitions(partitions_def_name))
```
sryza, line 1664: I think we should find a way to avoid fetching the entire set of partitions when adding a partition.

clairelin135:
Thanks for the feedback @sryza, I've updated the naming as you suggested.

I did try to change the `add_mutable_partitions` methods to not check which partitions exist. I ran into a few sqlalchemy nuances:
- I tried to add a uniqueness constraint on the table, but unfortunately mysql doesn't currently support the uniqueness constraint for text objects. Instead I added the uniqueness constraint to the index.
- When doing a bulk insert (adding multiple rows) to mysql and sqlite, doing a `on_conflict_do_nothing` will break after finding conflicting rows, subsequently not inserting the new rows. Writing each row individually takes a very long time, so instead I changed these dialects to do an `on_conflict_do_update` and write the row again with the same contents. Oddly, `on_conflict_do_nothing` performs correctly for bulk inserts on psql. With these new changes mysql and sqlite use `on_conflict_do_update`, psql uses `on_conflict_do_nothing`.

Let me know any thoughts cc @prha

sryza:
@clairelin135 how about querying the database to see which of the partitions that are being added are already present and then inserting the ones that are not?

Alternatively, we might be able to use VARCHAR instead of TEXT for SqlAlchemy?

clairelin135:
@sryza We could do varchar and set the limit to something high like `VARCHAR(255)`, I'd be surprised if any users had partition keys / names longer than 255 chars.

I have a feeling that when bulk inserting many partitions doing a query per added partition probably performs worse than just fetching all partitions for a definition in one pass. I think our best options are probably the current approach in this PR or the VARCHAR approach, though @prha might have a more educated opinion.

sryza:
> I have a feeling that when bulk inserting many partitions doing a query per added partition

Would we need to do a query per added partition? I was imagining we would do a single query with all the partitions:

```sql
select * from mutable_partitions
where partitions_def_name = "abc"
    and partition not in ("partition1", "partition2")
```

sryza:
I'm also comfortable with a character limit though. Seems not too difficult to increase in the future, right? I do think we'd want to make the limit consistent across all storage systems for consistency.

clairelin135:
@sryza ah yep, that sounds fine to me too. I amended the PR to follow your suggestion of the query fetching existing partitions.

I think that the `add_mutable_partitions` method should function fairly performantly and shouldn't be called super frequently, so I think this approach should work fine.

python_modules/dagster/dagster/_core/storage/alembic/versions/035_add_mutable_partitions_table.py
```diff
@@ -0,0 +1,49 @@
+"""add runtime partitions table
```
sryza, line 1:
```suggestion
"""add mutable partitions table
```

sryza (approved):
ship it!

sryza (commented):
ctrl-f for "mutable" still yielded a few results in error messages and comments

PR #11996, opened 2023-01-31
Add observe function (source asset analogue to materialize)

smackesey:
### Summary & Motivation

This adds an `observe` function directly analogous to `materialize`. The added tests also cover source asset APIs more generally that were not under test before.

New unit tests.

### DIFF
--- a/python_modules/dagster/dagster/__init__.py
+++ b/python_modules/dagster/dagster/__init__.py
@@ -228,9 +228,6 @@ from dagster._core.definitions.multi_dimensional_partitions import (
     MultiPartitionKey as MultiPartitionKey,
     MultiPartitionsDefinition as MultiPartitionsDefinition,
 )
-from dagster._core.definitions.observe import (
-    observe as observe,
-)
 from dagster._core.definitions.op_definition import OpDefinition as OpDefinition
 from dagster._core.definitions.output import (
     DynamicOut as DynamicOut,
--- a/python_modules/dagster/dagster/__init__.py
+++ b/python_modules/dagster/dagster/__init__.py
@@ -228,6 +228,9 @@ from dagster._core.definitions.multi_dimensional_partitions import (
     MultiPartitionKey as MultiPartitionKey,
     MultiPartitionsDefinition as MultiPartitionsDefinition,
 )
+from dagster._core.definitions.observe import (
+    observe as observe,
+)
 from dagster._core.definitions.op_definition import OpDefinition as OpDefinition
 from dagster._core.definitions.output import (
     DynamicOut as DynamicOut,
--- a/python_modules/dagster/dagster/_core/definitions/observe.py
+++ b/python_modules/dagster/dagster/_core/definitions/observe.py
@@ -1,0 +1,78 @@
+import warnings
+from typing import TYPE_CHECKING, Any, Mapping, Optional, Sequence
+
+import dagster._check as check
+from dagster._utils.backcompat import ExperimentalWarning
+from dagster._utils.merger import merge_dicts
+
+from ..instance import DagsterInstance
+from .assets import AssetsDefinition
+from .assets_job import build_assets_job
+from .job_definition import default_job_io_manager_with_fs_io_manager_schema
+from .source_asset import SourceAsset
+from .utils import DEFAULT_IO_MANAGER_KEY
+
+if TYPE_CHECKING:
+    from ..execution.execute_in_process_result import ExecuteInProcessResult
+
+
+def observe(
+    source_assets: Sequence[SourceAsset],
+    run_config: Any = None,
+    instance: Optional[DagsterInstance] = None,
+    resources: Optional[Mapping[str, object]] = None,
+    partition_key: Optional[str] = None,
+    raise_on_error: bool = True,
+    tags: Optional[Mapping[str, str]] = None,
+) -> "ExecuteInProcessResult":
+    """
+    Executes a single-threaded, in-process run which observes provided source assets.
+
+    By default, will materialize assets to the local filesystem.
+
+    Args:
+        assets (Sequence[Union[AssetsDefinition, SourceAsset]]):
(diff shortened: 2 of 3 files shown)


smackesey:
* master
  * **PR #11993** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11993" target="_blank">Graphite</a> 
    * **PR #11996** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11996" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/11996?utm_source=stack-comment).

schrockn:
I have pretty mixed feelings about adding an entire parallel set of verbs for source assets. `observe` is also an extremely generic verb and gives more weight to this operation than it deserves, since this is pretty much a test utility.

I'd rather see what our well-supported execution pathway is when using `Definitions`. In general, I think we should standardizing more of our execution pathways to flow through a `Definitions` object so that we use uniform rules, especially when it comes to resource binding, which feels much more natural there.

In particular I think we should remove `build_source_asset_observation_job` and `build_assets_job` and instead make it more convenient to use Definitions.

schrockn (changes requested):
Popping back to your queue. In particular, I'd like to see what our well-supported, non-experimental method is for observing assets, and only then seeing if `observe` or something like it is necessary.

smackesey:
>I have pretty mixed feelings about adding an entire parallel set of verbs for source assets. observe is also an extremely generic verb and gives more weight to this operation than it deserves, since this is pretty much a test utility.

Part of the inspiration for this is your recommendation of an API for updating the logical version of regular assets via an `AssetObservation`. This function would be adapted for that case in the future. We'd need to add an `observe_fn` for regular assets, and this function would then handle either kind of asset.

That said that API of course doesn't exist yet, so I agree it probably doesn't make sense to make this public. But it is definitely a useful test utility-- can we merge it without public exposure (I removed the public exposure in last push).

>In particular I think we should remove build_source_asset_observation_job and build_assets_job and instead make it more convenient to use Definitions.

Agreed.

schrockn:
> Part of the inspiration for this is your recommendation of an API for updating the logical version of regular assets via an AssetObservation. This function would be adapted for that case in the future. We'd need to add an observe_fn for regular assets, and this function would then handle either kind of asset.

My suggestion was definitely _not_ to add an `observe_fn` to `AssetsDefinition`. I strongly oppose that expansion of surface area to the public API. We can discuss further about exactly what is needed there.

The next step is here to propose a first-class way of observing a source asset whilst using `define_asset_job` in my view. Then we can implement this utility function (renamed to be something more specific than observe) in terms of that API, rather than the API we both agree should be deleted.

smackesey:
>My suggestion was definitely not to add an observe_fn to AssetsDefinition. I strongly oppose that expansion of surface area to the public API. We can discuss further about exactly what is needed there.

I am curious what you have in mind then for the recommended API-- generating asset observations from an op?

>The next step is here to propose a first-class way of observing a source asset whilst using define_asset_job in my view. Then we can implement this utility function (renamed to be something more specific than observe) in terms of that API, rather than the API we both agree should be deleted.

Why block adding tests pending this new API? These test something that should go out in tmrw's release, which is use of `required_resource_keys` in a source asset: https://github.com/dagster-io/dagster/pull/11993

It will be easy to swap out the single `build_source_asset_observation_job` call when that API lands. Happy to rename this to `observe_source_asset` (or whatever else you want) and put it in `test_utils`.

sryza:
> The next step is here to propose a first-class way of observing a source asset whilst using define_asset_job in my view. Then we can implement this utility function (renamed to be something more specific than observe) in terms of that API, rather than the API we both agree should be deleted.

I'm a little bit confused about what `define_asset_job` has to do with this. Users should have a way of testing source asset observation functions without thinking about jobs at all, and without putting their assets into a `Definitions` object.  `define_asset_job` is not invoked by any of our asset testing APIs, and I don't think it should be.

schrockn:
I think the implementation of this function should create an ephemeral Definitions object, get the job out of of that, and return it. `build_source_asset_observation_job` is doing a bunch of duplicated work. From the user perspective they will have a utility function.

sryza:
> I think the implementation of this function should create an ephemeral Definitions object, get the job out of of that, and return it.

What's the advantage of that vs. directly invoking the API that `Definitions` will end up invoking?

schrockn:
I think there is pretty self-evident benefit in reducing the number of places where we do the machinery that Definitions does. 

The benefit is mostly around driving all public-facing user APIs through a single choke point where we can do all type-checking, good errors messaging, all usability affordances, everything. Right now it looks like -- from code inspection –– things like `build_source_asset_observation_job` are both "public" APIs both also used in the guts of `build_asset_selection_job` (with content managers to disable warnings and like, which is a pretty big red flag). This doesn't make too much sense to me. Having all the helpful functions goes through the same exact layer as production will reduce the distance between our testing APIs and our production APIs which is good.

We can work towards making it the responsibility Definitions to properly construct jobs given a set of assets, resources etc. It is the single "binding" layer in that way.  I think this is a fundamentally a better model than exposing internals publicly, which I how I view the current approach.

smackesey:
Just posted a fix for `define_assets_job` for source assets and adding more tests now: https://github.com/dagster-io/dagster/pull/12018

schrockn:
> > I think the implementation of this function should create an ephemeral Definitions object, get the job out of of that, and return it.
> 
> What's the advantage of that vs. directly invoking the function that `Definitions` will end up invoking?

Threw up https://github.com/dagster-io/dagster/pull/12038 to demonstrate this approach on materialize and show advantages.

sryza (approved):
This seems reasonable and low-stakes to me, given that we're not exposing it publicly.

schrockn (approved):
okie dokie

PR #12008, opened 2023-02-01
More helpful asset key mismatch errors

benpankow:
## Summary

A little QoL idea that came out of our onboarding breakout group last week. Adds a suggestion to the error message when an upstream asset dep is not found, listing any assets with similar keys:

```python
dagster._core.errors.DagsterInvalidDefinitionError: Input asset '["my", "asset1"]' for asset '["asset3"]' 
is not produced by any of the provided asset ops and is not one of the provided sources. 
Did you mean one of the following?
    ["my", "prefix", "asset1"]
```

Right now, uses three heuristics for finding potential candiates:
- Same asset name, similar prefix
- Similar asset name, same prefix
- Same asset name, prefix is off-by-one component (e.g. `["snowflake', "elementl", "prod", "my_asset"] vs ["snowflake", "elementl", "my_asset"]`

Currently the candidates are unordered and there's no limit to the number shown because this is a quick impl/rfc. Can flesh it out if we find it interesting.

## Test Plan

Added a few unit tests to showcase behavior.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/resolved_asset_deps.py
+++ b/python_modules/dagster/dagster/_core/definitions/resolved_asset_deps.py
@@ -1,6 +1,8 @@
 from collections import defaultdict
 from typing import AbstractSet, Dict, Iterable, List, Mapping, Tuple, cast
 
+from fuzzywuzzy import fuzz
+
 from dagster._core.definitions.events import AssetKey
 from dagster._core.errors import DagsterInvalidDefinitionError
 from dagster._utils.backcompat import experimental_warning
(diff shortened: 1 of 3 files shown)


benpankow:
* master
  * **PR #12008** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12008" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12008?utm_source=stack-comment).

sryza (commented):
I think this is a great idea. Can we factor it out into a helper function? Any concerns about adding fuzzywuzzy as a dep?

benpankow:
> Can we factor it out into a helper function? 

done 👍 

> Any concerns about adding fuzzywuzzy as a dep?

Switched over to [thefuzz](https://github.com/seatgeek/thefuzz) which is an actively maintained fork/successor w/ a corporate sponsor (seatgeek). I think it's a fairly innocuous dep & doesn't have any upstream deps.

python_modules/dagster/dagster/_core/definitions/resolved_asset_deps.py
```diff
@@ -41,6 +43,54 @@ def get_resolved_asset_key_for_input(
         )
 
 
+def _resolve_similar_asset_names(
+    target_asset_key: AssetKey,
+    assets_defs: Iterable[AssetsDefinition],
+) -> List[AssetKey]:
```
sryza, line 49: Nitpick: Sequence

sryza (commented):
Last thing: should we be worried about perf? Some users have repos with thousands of assets. Seems like at minimum we'd want to avoid printing that many out.

benpankow:
Added a small perf test to sanity check that the perf hit is minimal (~0.025s with 5k assets). Limited the number we print to 10 (arbitrarily) for now. Eventually might be nice to add some sorting.

sryza (approved):
Ship it!

PR #12009, opened 2023-02-01
[prototype] Link to asset definitions from local Dagit

benpankow:
****Superseded/fleshed out in follow-up stack: https://github.com/dagster-io/dagster/pull/12097****

## Summary

Based on an idea that came out of our onboarding breakout group last week. Very quick & rough implementation of using `inspect` to find out where a user is defining assets in their code and forwarding the codepath to Dagit via a special metadata field. Not very resilient but works well enough for simple cases.

https://user-images.githubusercontent.com/10215173/215924265-5b27b93a-5f2c-45df-9354-595ec42d0e75.mov

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/SidebarAssetInfo.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/SidebarAssetInfo.tsx
@@ -1,5 +1,5 @@
 import {gql, useQuery} from '@apollo/client';
-import {Box, Colors, ConfigTypeSchema, Icon, Spinner} from '@dagster-io/ui';
+import {Box, Colors, ConfigTypeSchema, ExternalAnchorButton, Icon, Spinner} from '@dagster-io/ui';
 import * as React from 'react';
 import {Link} from 'react-router-dom';
 import styled from 'styled-components/macro';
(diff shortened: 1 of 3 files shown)


benpankow:
* master
  * **PR #12009** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12009" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12009?utm_source=stack-comment).

petehunt:
This is sweet. Can we link to github from cloud/production dagit?

benpankow:
> Can we link to github from cloud/production dagit?

Yes, with a bit of extra work. With https://github.com/dagster-io/dagster-cloud-action/pull/90, we can let Cloud know where the user's code is located on GH (and where the package is nested). We also need to get the path from the module to the line-of-code, which is a little more work stacked on what's here - the result is pretty neat though:


https://user-images.githubusercontent.com/10215173/216190958-151b3beb-9a5f-4170-b8cf-3f66606f5c79.mov

This might not be super robust if users are doing weird things with their packages or repos, but should work for most use-cases out of the box.

salazarm:
[Developer developers developers](https://www.youtube.com/watch?v=Vhh_GeBPOhs)

This is great! Linking in VSCode would also be huge. 

We need a link like `vscode://file/{full path to project}/`, Maybe we can have a dialog to point us to where the project is on their machine and then remember it, one time setup for vscode deep linking.

salazarm:
This is great! I think we may want to model this as a property on the asset / run / job / schedule decorator and allow passing a dictionary in to allow for multiple types of links. We could make the dictionary keys like "vscode", "fivetran", "custom", so that we can show icons for the various link in the UI.

salazarm (approved):

Issue #2403, opened 2020-04-23
extract IS_AIRFLOW_INGEST_PIPELINE_STR and friends from core


mgasner:
cc @catherinewu

mgasner:
some options

1) move this out of tags completely (perhaps using a resource, but that would involve complex changes to resource lifecycle)
2) (re?)introduce a `tags_fn` on `PipelineDefinition`
3) add an `execution_time` field on `PipelineRun` that could be written when we change the state to `STARTED` (and that dagster-airflow and anyone else could use as they desired)

schrockn:
re: 3 that is not possible. this is more like a partition key and it can be very different from the actual execution time

schrockn:
1) is probably the most realistic option tbh. We decided to punt on that because want to make sure we understood requirements by getting into early user's hands

schrockn:
Based on meeting: Next steps are to explore a resource-based solution.

Could:

1) Require user to always provide this value in config
2) Have config system evaluate default value to get this value. Note: might require changes to the config system (to have default value be sourced from function)
3) Provide some for resources to stash values in tags (or elsewhere) at run init time.

schrockn:
Joe with https://github.com/dagster-io/dagster/pull/12023/ I think you can take this. Good riddance to this hackery.

schrockn:
Oh nevermind looks like we need it for ingestion too. Sigh.

PR #12023, opened 2023-02-01
[dagster-airflow] remove api's for 1.x.x release

Ramshackle-Jamathon:
### Summary & Motivation

This pr removes the following api's from dagster-airflow along with their tests and docs

- `make_airflow_dag`
- `make_airflow_dag_for_operator`
- `make_airflow_dag_containerized`
- `airflow_operator_to_op`

These are being remove because the original reasons for them being created hold true anymore; that dagster wants to be a software abstraction ontop of existing orchestration frameworks (for make_airflow_dag) and that we want to lift all the Operator/Integration functionality already existing in airflow into dagster(airflow_operator_to_op).

The challenge of running dagster on airflow is also something that we don't want to carry forward, already many dagster features don't work when transformed into airflow dags

In earlier versions of dagster it was harder for a user to get a "production" dagster infrastructure setup and being able to run on existing airflow infra made adoption easier. But thats not as true today, in OSS the local docker or helm setup works well and in cloud Serverless makes it really easy to get up and running.

For users who still want to keep everything captured in the same operational view (the airflow ui) they can use the DagsterOperator and DagsterCloudOperator which uses a more coherent interface for interacting with dagster from airflow.


docs/content/integrations/airflow.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 title: Dagster with Airflow | Dagster
-description: The dagster-airflow package allows you to export Dagster jobs as Airflow DAGs, as well as to import Airflow DAGs into Dagster jobs.
+description: The dagster-airflow package allows you to export Dagster jobs as Airflow DAGs, as well as trigger job runs from Airflow.
```
schrockn, line 3: I think this is wrong. We no longer export jobs as Airflow DAGs. But we can ingest Airflow DAGs and execute them natively in Dagster.

schrockn (approved):
cc: @sryza on this, especially airflow_operator_to_op

@Ramshackle-Jamathon also assigned https://github.com/dagster-io/dagster/issues/2403 to you, which you should be able to do with this deletion.

Definitely support this. Get nod from sandy before commiting

sryza:
Are you able to provide a bit of the context here? I had thought these were useful utilities.

Ramshackle-Jamathon:
@sryza Mostly Im in favor of removing these because Im don't think the original reasons for them being created hold true anymore; we don't want to be a platform agnostic orchestration framework (for `make_airflow_dag`) and that we want to lift all the Operator/Integrations already existing in airflow (`airflow_operator_to_op`).

I also think if there were significant usage of the `make_airflow_dag` apis we'd have more bug reports related to dagster features failing to port/map over onto their airflow counterparts. Im also aware of a few bugs ease of use issues that I would have expected to hear reports of (one the big one's being that in airflow 1 when you create a new DAG with `make_airflow_dag` its stuck in a deactivated state.

schrockn:
@Ramshackle-Jamathon Please update the commit summary with your explanation here.

@sryza effectively we've come to the determination that this adoption modality is too difficult to get right, and exposes us to a ton of surface area. Like Joe said, I don't believe it is being used much, and if so, would be a source of nearly impossible-to-debug issues. If we don't believe in this migration pathway, we should just delete it.

sryza:
Got it. My instinct would be to mark `make_airflow_dag` as experimental or deprecated, but will defer to you.

I think we should keep `airflow_operator_to_op`. Its use case is very different from `make_airflow_dag` and I think still valid.  It doesn't require any running Airflow infrastructure to be useful.

Ramshackle-Jamathon:
A big thing for me is the reliability/stability of these apis across the full range of airflow versions, they really should have always been marked as experimental, and for v1 I want to make sure that anything we expose works or is upfront about being experimental.

`airflow_operator_to_op` for example, I don't think is useful outside of a POC of running an airflow operator in dagster, the `execute` method has no way of passing config in (things that you _would_ want in a real world use case) and the way it creates connections only works with a small subset of connection types (those that _can_ be correctly serialized to uri) and airflow versions (the ones that support loading connections from envs).

If we want to re-add that type of thing im open to it but im very confident its not being used in a serious way atm

sryza:
> and airflow versions (the ones that support loading connections from envs).

Does this mean that `airflow_operator_to_op` does not work with the latest version of Airflow in any case?

Ramshackle-Jamathon:
sorry no it works with more than a small amount of versions, all 2.x.x airflow versions can load connections from envs (but support was added late-ish in 1.x.x) the uri serialization imo is more of an issue, it won't include description (which is often overrode for custom connection types, since you can't define unique fields) and extra when its not uri serializable json (fairly common)

sryza:
IMO if we think `airflow_operator_to_op` is useful in a bunch of common cases, it's worth keeping around as experimental.

A common objection from users is that the Airflow operator ecosystem is more developed than Dagster's, even though my understanding was that the majority of Airflow operators out there have little deep synergy with Airflow.   It's nice to be able to say in user chats that in most cases they can just use those operators (even if we can't guarantee it will work 100% of the time).

Ramshackle-Jamathon:
I would much rather revisit it than carry it forward, in its current state its not _really_ a way to use the airflow operator ecosystem in dagster. 

The pain of having to have airflow installed alongside dagster is not trivial, and using `airflow_operator_to_op` in its current state loses a bunch of the benefits of dagster (io_managers, resources). I think we can do better if we wanted to address porting airflow integrations to dagster.

sryza:
@clairelin135 and @OwenKephart were the ones that built this originally, so I'm curious if they have thoughts. 

> "it's not really a way to use the airflow operator ecosystem in dagster"

The way I'm inclined to think about this is that if if, in over 50% of situations that a person uses an Airflow operator, `airflow_operator_to_op` would work, then it's worth keeping. Do you have a guess on the % of situations where that would be the case?

OwenKephart:
I'll let @clairelin135 weigh in as well, but I would be very comfortable removing `airflow_operator_to_op`. I don't think the average quality of an airflow integration is high enough to warrant recommending this pattern, and I don't have high confidence that a high-quality integration would have a particularly high chance of working with this.

I also agree that we can revisit this at a later date if this functionality is something people are clamoring for, but my sense is that this is not something that has come up as a selling point.

clairelin135:
Agree with what @OwenKephart mentioned earlier--my feeling is also that this is a feature that is not being used in a serious way. We can always try to support it if users push for it

sryza (approved):
Alright sounds like I'm the sole defender. Let's kill it!

PR #12026, opened 2023-02-01
Add --working-directory argument to dagster dev

gibsondan:

### DIFF
--- a/python_modules/dagster/dagster/_cli/workspace/cli_target.py
+++ b/python_modules/dagster/dagster/_cli/workspace/cli_target.py
@@ -334,14 +334,18 @@ def python_module_option(allow_multiple):
     )
 
 
+def working_directory_option():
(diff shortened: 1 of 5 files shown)


python_modules/dagster/dagster/_cli/workspace/cli_target.py
```diff
@@ -334,14 +334,18 @@ def python_module_option(allow_multiple):
     )
 
 
+def working_directory_option():
+    return click.option(
+        "--working-directory",
+        "-d",
+        help="Specify working directory to use when loading the repository or job",
+        envvar="DAGSTER_WORKING_DIRECTORY",
+    )
```
alangenfeld, line 343: side-quest: i think there might be click arguments that validate the value is a directory

alangenfeld (approved):

PR #12033, opened 2023-02-02
add run job index

prha:
### Summary & Motivation
We're missing an OSS index that powers job-run views in dagit.

BK


python_modules/dagster/dagster/_core/storage/runs/schema.py
```diff
@@ -105,6 +105,14 @@
 
 db.Index("idx_run_tags", RunTagsTable.c.key, RunTagsTable.c.value, mysql_length=64)
 db.Index("idx_run_partitions", RunsTable.c.partition_set, RunsTable.c.partition, mysql_length=64)
+db.Index(
+    "idx_runs_by_job",
+    RunsTable.c.pipeline_name,
+    RunsTable.c.id,
+    mysql_length={
+        "pipeline_name": 255,
```
alangenfeld, line 113: confident in this 255? align with similar settings?

alangenfeld (approved):

python_modules/dagster/dagster/_core/storage/alembic/versions/035_add_run_job_index.py
```diff
@@ -0,0 +1,22 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    add_run_job_index()
+
+
+def downgrade():
+    pass
```
alangenfeld, line 22: we should probably drop on down grade right? unlikely path but probably best to follow the ideal

python_modules/dagster/dagster/_core/storage/runs/schema.py
prha: I'll switch this to 512... that's what we use in our mysql compatibility code to coerce text fields in other places.

PR #12034, opened 2023-02-02
[dagster-airflow] use full timestamp for partition name

Ramshackle-Jamathon:
### Summary & Motivation

the 1000s of partitions were coming from the combination of a 1min cron schedule and partition names that only used y/m/d

it might be worth making this configurable by users in the future.

### DIFF
--- a/python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
+++ b/python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
@@ -407,8 +407,8 @@ def load_assets_from_airflow_dag(
         partitions_def=TimeWindowPartitionsDefinition(
             cron_schedule=cron_schedule,
             timezone=dag.timezone.name,
-            start=start_date.strftime("%Y-%m-%d"),
-            fmt="%Y-%m-%d",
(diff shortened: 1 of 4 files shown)


python_modules/libraries/dagster-airflow/dagster_airflow/dagster_job_factory.py
```diff
@@ -407,8 +407,8 @@ def load_assets_from_airflow_dag(
         partitions_def=TimeWindowPartitionsDefinition(
             cron_schedule=cron_schedule,
             timezone=dag.timezone.name,
-            start=start_date.strftime("%Y-%m-%d"),
-            fmt="%Y-%m-%d",
+            start=start_date.strftime("%Y-%m-%dT%H:%M:%S"),
+            fmt="%Y-%m-%dT%H:%M:%S",
```
Ramshackle-Jamathon, line 411: its not really possible to set sub min crons so i could see dropping seconds

benpankow (approved):
Just to make sure - this will properly handle the sub-day cron case (down to minutely) but the example cron schedule was updated to be daily, right?

I assume this was manually tested?

PR #12037, opened 2023-02-02
[RFC] preview asset backfill

sryza:


ruizh22:
hey @sryza this seems like a really nice feature we could take advantage of, is there a plan to finish and release it? Happy to talk more about either here in the comment or on slack.

PR #12038, opened 2023-02-02
Use Definitions in the body of materialize functions (and other public-facing execution functions)

schrockn:
### Summary & Motivation

I had suggested that we use `Definitions` in the body of a proposed `observe` function in https://github.com/dagster-io/dagster/pull/11996 and I think the most efficient path forward was to discuss this in the form of an RFC diff.

This has two advantages:

1) Marginally less code
2) More importantly. this funnels more of our public-facing execution APIs through a single choke point––`Definitions`––rather than having multiple execution paths call into the lower-level APIs that `Definitions` itself ends up invoking. I think this is nearly an unalloyed good, making error messages more consistent and also forcing us to be more consistent on execution behavior.

Re: 2) this diff itself exposed a couple instances of this, where we don't allow empty asset jobs, wheres both `materialize` and our core `@job` APIs are fine with empty jobs. And then the different exception type.

BK

### DIFF
--- a/python_modules/dagster/dagster_tests/asset_defs_tests/test_materialize_to_memory.py
+++ b/python_modules/dagster/dagster_tests/asset_defs_tests/test_materialize_to_memory.py
@@ -21,6 +21,7 @@ from dagster import (
     op,
     with_resources,
 )
+from dagster._core.errors import DagsterInvalidInvocationError
 
 
 def test_basic_materialize_to_memory():
@@ -116,11 +117,11 @@ def test_materialize_conflicting_resources():
         materialize_to_memory([first, second])
 
     with pytest.raises(
-        DagsterInvalidDefinitionError,
+        DagsterInvalidInvocationError,
         match=(
-            "resource with key 'foo' provided to job conflicts with resource provided to assets."
-            " When constructing a job, all resource definitions provided must match by reference"
-            " equality for a given key."
+            r'AssetsDefinition with key \["first"\] has conflicting resource definitions with'
+            r" provided resources for the following keys: foo. Either remove the existing"
+            r" resources from the asset or change the resource keys"
         ),
     ):
         materialize_to_memory(
--- a/python_modules/libraries/dagster-airflow/dagster_airflow_tests/test_dagster_job_factory/test_load_assets_airflow_2.py
+++ b/python_modules/libraries/dagster-airflow/dagster_airflow_tests/test_dagster_job_factory/test_load_assets_airflow_2.py
@@ -56,10 +56,6 @@ def test_load_assets_from_airflow_dag():
         dag_bag = DagBag(dag_folder=tmpdir_path)
         asset_dag = dag_bag.get_dag(dag_id="asset_dag")
 
-        @asset
-        def new_upstream_asset():
-            return 1
-
         assets = load_assets_from_airflow_dag(
             dag=asset_dag,
             task_ids_by_asset_key={
@@ -76,6 +72,17 @@ def test_load_assets_from_airflow_dag():
         other_assets = load_assets_from_airflow_dag(
             dag=other_dag,
         )
+
+        resources = None
+
+        if assets:
+            first_asset = next(iter(assets))
+            if "io_manager" in first_asset.resource_defs:
+                resources = {"io_manager": first_asset.resource_defs["io_manager"]}
+
+        @asset(resource_defs=resources)
+        def new_upstream_asset():
+            return 1
 
         result = materialize(
             [*assets, new_upstream_asset, *other_assets],
--- a/python_modules/dagster/dagster/_core/definitions/materialize.py
+++ b/python_modules/dagster/dagster/_core/definitions/materialize.py
@@ -1,6 +1,8 @@
 from typing import TYPE_CHECKING, Any, Mapping, Optional, Sequence, Set, Union
 
 import dagster._check as check
+from dagster._core.definitions.events import AssetKey
+from dagster._core.definitions.unresolved_asset_job_definition import define_asset_job
 from dagster._utils.merger import merge_dicts
 
 from ..errors import DagsterInvariantViolationError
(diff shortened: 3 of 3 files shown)


schrockn:
* master
  * **PR #12038** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12038" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12038?utm_source=stack-comment).

python_modules/dagster/dagster_tests/asset_defs_tests/test_materialize.py
```diff
@@ -185,8 +185,11 @@ def the_asset():
 
 
 def test_materialize_no_assets():
-    with instance_for_test() as instance:
-        assert materialize([], instance=instance).success
+    # we need to decide if this is a valid use case given that it is inconsistent with Definitions
+    # rules
+    assert True
+    # with instance_for_test() as instance:
+    #     assert materialize([], instance=instance).success
```
dpeng817, line 192: what's the actual use case for this anyway - seems like a pythonic edge case rather than behavior users would actually want

python_modules/dagster/dagster_tests/asset_defs_tests/test_materialize_to_memory.py
```diff
@@ -115,12 +116,13 @@ def second():
...
 
     with pytest.raises(
-        DagsterInvalidDefinitionError,
-        match=(
-            "resource with key 'foo' provided to job conflicts with resource provided to assets."
-            " When constructing a job, all resource definitions provided must match by reference"
-            " equality for a given key."
-        ),
+        # TODO: which error is best this or DagsterInvalidDefinitionError?
+        DagsterInvalidInvocationError,
```
dpeng817, line 120: urgh yea I guess the definition error is a result of the internal processes which create the job, which might be confusing to users. invocation error seems more grokkable to a user encountering this imo

python_modules/dagster/dagster_tests/asset_defs_tests/test_materialize.py
schrockn: A case to consider if when you are producing the list of assets programmatically and don't want to special the case with no assets. I don't feel super strongly here on this particular issue. I think the interesting bit is the behavior is inconsitent.

python_modules/dagster/dagster_tests/asset_defs_tests/test_materialize_to_memory.py
```diff
@@ -158,7 +160,9 @@ def the_asset(the_source):
 
 
 def test_materialize_no_assets():
-    assert materialize_to_memory([]).success
+    assert True
+    # TODO do we support no asset jobs or not?
+    # assert materialize_to_memory([]).success
```
smackesey, line 165: Is there any case where a user would actually want to run an empty job? I can't think of any-- seems like it will always be unintended (could easily be wrong on this).

In that case, since this was working with the old implementation, seems like the ideal case would be to flag this at a lower level (so it works for both ops and assets) and throw an error?

schrockn: I'm totally fine allowing or disallowing this. The more interesting point to discuss in this PR is having all our top-level invocations go through the `Definitions` codepath good.

smackesey:
In general I am a fan of funneling public execution APIs through`Definitions` (for reasons of standardization as Nick stated), but I'm not sure I'm aware of potential costs. Maybe a small performance hit, if we're doing this when we could just call the lower level functions directly? But that seems negligible.

python_modules/dagster/dagster_tests/asset_defs_tests/test_materialize.py
sryza: > A case to consider if when you are producing the list of assets programmatically and don't want to special the case with no assets.

Yeah - I think it's worth making this complete successfully to support this use case.

sryza (commented):
This change seems net-positive to me if we don't think there are performance implications. I think we should eventually expose something like a `materialize` method on `Definitions`, which would allow this implementation to be even more condensed.

schrockn:
For a top level entry point like this that will be called once and then instigate expensive computations, I the one-time performance overhead and going through our Python layers is immaterial. If it is material either something has gone wrong and we need to fix it, or we are using something like caching asset definitions.

schrockn:
Ok my plan will be to submit a change to asset for empty asset jobs. Assuming that makes sense we will stack this on top of that.

schrockn:
Want to reiterate here that I think is quite important to consolidate the creation of our definition artifacts through the Definitions codepath as they have gotten more complicated as of late. Late-binding of resources is one such example, but the one that really sticks out is the "pending repository" concept used to implement caching asset definitions.

schrockn:
Ok I just "un-RFCed" this. I was able to retain previous behavior by writing this in terms of a defined job, rather than the "implicit" global job.

sryza (approved):

python_modules/libraries/dagster-airflow/dagster_airflow_tests/test_dagster_job_factory/test_load_assets_airflow_2.py
```diff
@@ -68,6 +64,17 @@ def new_upstream_asset():
...
+
+        if assets:
+            first_asset = next(iter(assets))
+            if "io_manager" in first_asset.resource_defs:
+                resources = {"io_manager": first_asset.resource_defs["io_manager"]}
+
+        @asset(resource_defs=resources)
+        def new_upstream_asset():
+            return 1
+
```
schrockn, line 77: Funny enough this appeared to have found a real issue in the dagster-airflow work where we could get in state of providing conflicting resource definitions. This code is structured a bit oddly, where the job is created in load_assets_from_airflow_dag, but then discarded after it's contents have been shuffled into a single graph-backed AssetsDefinition.

@Ramshackle-Jamathon, @sryza you two might want to talk about the best approach here.

python_modules/dagster/dagster_tests/asset_defs_tests/test_materialize.py
dpeng817: hard for me to imagine programmatic construction of assets ever giving no assets in a realistic scenario, seems more than likely that if no assets are produced, you're pointing at the wrong module, or something else is going wrong.

PR #12042, opened 2023-02-02
feat(dbt-cloud): compile run only if job has environment variable cache

rexledesma:
### Summary & Motivation
Some user feedback on the caching mechanism:
> Will this compile all jobs within that project ID - isn’t that a bit unnecessary since that means all of my staging, testing jobs will be compiled as well? Or will it reference my dagster repo code for only the jobs necessary to compile?

> we have jobs just hanging around from tests (branch tests for complex changes) , others that are ad hoc jobs (for people to manually update and run), jobs just for when I test dbt upgrades, etc etc :slightly_smiling_face:  None of those need to be run

> All jobs must comply to your "dagster" expectations meaning - I'm getting the below which as is, would be an issue unless I "fixed" and ran all jobs via Dagster

A dbt project could potentially have many jobs (for testing, managed by Dagster, manual use, etc). Rather than trying to manage all of this, only manage the jobs that are explicitly labeled by the user.

Here, we use the presence of the environment variable cache to determine whether or not we should populate its cache.

pytest, local

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/cli.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/cli.py
@@ -30,9 +30,24 @@ def cache_compile_references(
     # Compile each job with an override
     for dbt_cloud_job in dbt_cloud_jobs:
         job_id: int = dbt_cloud_job["id"]
-        job_commands: List[str] = dbt_cloud_job["execute_steps"]
+
+        # Only run on jobs with the Dagster dbt Cloud compile run id env var set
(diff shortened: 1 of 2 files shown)


rexledesma:
* master
  * **PR #12042** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12042" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12042?utm_source=stack-comment).

OwenKephart (approved):
nice, makes sense

PR #12068, opened 2023-02-03
Add `requests` dependency to hello-dagster

kevinschaich:
### Summary & Motivation

`requests` is required for the Hacker News example but not in the copy/paste command. Looking forward to trying this out, looks awesome! 🚀


jamiedemaria (approved):
good catch!

cmpadden:
Hey @kevinschaich & @jamiedemaria - thanks again for pointing this out.

It looks like `requests` is already installed as it's a sub-dependency of _dagster_:

```
 > pipdeptree -fl
 
dagit==1.6.4
    dagster-webserver==1.6.4
        ...
        dagster==1.6.4
            ...
            requests==2.31.0
```

Is it better to be explicit and include it in the `pip install` or is it fine to have the sub-dependency?

jamiedemaria:
I'd lean toward leaving it as a sub-dependency. On the off-chance we need to put a pin in the version of `requests` leaving it as a sub-dependency will ensure the user installs the correct version

PR #12074, opened 2023-02-03
Add 401 to the list of API codes that our k8s client retries on

gibsondan:
Summary:
401 would typically not be a retryable error, but a user reported hitting it when they scaled up their cluster, and https://github.com/aws/containers-roadmap/issues/1810 seems to suggest retrying as a workarounds. The downside of retrying on a 401 seems fairly low as well. Open to push-back on this though.

Test Plan: Existing BK test coverage of the k8s client


alangenfeld (approved):
Feels a little sketchy but as you called out the downside here is just a small amount of latency from the retries. If you feel good about this, go for it.

PR #12086, opened 2023-02-03
[dagster-dbt] [1/2] Add streaming entrypoint for dbt cli execution

OwenKephart:
### Summary & Motivation

We want to be able to process dbt output as we get it, so in order to do that, I'm refactoring the execute_cli method to call a new execute_cli_streaming method, which we'll consume from in a future PR.

For now, this should have no impact on functionality.


python_modules/libraries/dagster-dbt/dagster_dbt/cli/utils.py
```diff
@@ -48,60 +58,75 @@ def _create_command_list(
     return prefix + full_command
 
 
-def execute_cli(
-    executable: str,
-    command: str,
-    flags_dict: Mapping[str, Any],
-    log: Any,
-    warn_error: bool,
+def _parse_line(line: str, json_log_format: bool) -> DbtCliEvent:
```
rexledesma, line 61: this should just be a method on `DbtCliEvent`, just so we colocate this parsing logic

python_modules/libraries/dagster-dbt/dagster_dbt/cli/utils.py
```diff
@@ -111,74 +136,85 @@ def execute_cli(
...
+    lines, parsed_json_lines = [], []
+    for event in execute_cli_stream(
+        command_list=command_list,
+        json_log_format=json_log_format,
+        ignore_handled_error=ignore_handled_error,
+    ):
+        if event.return_code is not None:
+            return_code = event.return_code
+            log.info(f"dbt exited with return code {return_code}")
+            break
```
rexledesma, line 197: This feels like its own distinct event from `DbtCliOutput`, since it is a terminal event. I suggest that the `execute_cli_stream` be executed on the process, and then at the very end, the return code can be directly taken from the process, outside of the iteration.

rexledesma (changes requested):

python_modules/libraries/dagster-dbt/dagster_dbt/cli/utils.py
OwenKephart: agreeed

python_modules/libraries/dagster-dbt/dagster_dbt/cli/utils.py
OwenKephart: Agree that return_code is pretty distinct from the other DbtCliEvent stuff, so I'm down to remove it. However, the next PR in the stack adds another callsite for execute_cli_stream, which increases the need to have everything factored together. That is, we need to do a few things:

1. create a subprocess with some specific arguments / env vars
2. iterate through the stdout of that subprocess, yielding parsed events as we go
3. wait for the subprocess to complete, raising an error (in some cases) based off the return code
4. have the calling function aware of that return code value (so that it can construct a DbtCliOutput)

So we're stuck in this situation where we want the calling function to consume all the yielded events, but also know the return code. The only way we can do this without having some function which handles both yielding DbtCliEvents and return codes is if the calling process is given the responsibility for 1, 3, and 4. These could take the form of two functions (_get_dbt_subprocess() and _cleanup_dbt_subprocess()), but I figured that might be more confusing than anything. So I just updated the return type of `execute_cli_stream` to be Union[DbtCliEvent, int].

sryza (approved):

PR #12087, opened 2023-02-03
type annotations for input_bindings in composition.py

sryza:
branch-name: composition-types

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/composition.py
+++ b/python_modules/dagster/dagster/_core/definitions/composition.py
@@ -64,6 +64,91 @@ class MappedInputPlaceholder:
     """Marker for holding places in fan-in lists where input mappings will feed."""
 
 
+class InvokedNodeOutputHandle:
+    """The return value for an output when invoking a node in a composition function."""
+
+    solid_name: str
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster/_core/definitions/composition.py
```diff
@@ -64,6 +64,91 @@ class MappedInputPlaceholder:
...
+    """The return value for an output when invoking a node in a composition function."""
+
+    solid_name: str
+    output_name: str
+    node_type: str
+
+    def __init__(self, solid_name: str, output_name: str, node_type: str):
+        self.solid_name = check.str_param(solid_name, "solid_name")
+        self.output_name = check.str_param(output_name, "output_name")
+        self.node_type = check.str_param(node_type, "node_type")
```
alangenfeld, line 77: nit: this is internally constructed only right? could drop `check` calls and do condensed namedtuple or similar

alangenfeld: like the other entries in the source union

alangenfeld (approved):

Issue #10874, opened 2022-12-02
Enabling using assets as input for a job/graph


sryza:
_Originally posted by @PascalStehling in https://github.com/dagster-io/dagster/discussions/10802_

Hi all,

is there a way to have Assets as input for a "normal" Job. I have some Assets and based on them I want to run a Job, which does not create a new Asset. For the job I want to reuse the already existing structure with the Asset and its IO Manager and not load the Data directly from the source DB. Is this possible? And if so, how? 

Below is an example of what i am thinking about. Its not working as the AssetIn's are directly passed as Input and not the loaded Asset values.

Thank you for your help and keep on the good work :)

```py
from dagster import asset, repository, job, op, OpExecutionContext, AssetIn, fs_io_manager


@asset(io_manager_def=fs_io_manager)
def get_five() -> int:
    return 5


@asset(io_manager_def=fs_io_manager)
def get_ten() -> int:
    return 10


@op
def do_something_special_op(context: OpExecutionContext, five: int,
                            ten: int) -> None:
    context.log.info(five + ten)


@job(resource_defs={"io_manager": fs_io_manager},
     input_values={
         "get_five": AssetIn("get_five", input_manager_key="io_manager"),
         "get_ten": AssetIn("get_ten", input_manager_key="io_manager"),
     })
def asset_based_job(get_five: int, get_ten: int) -> None:
    do_something_special_op(get_five, get_ten)


@repository
def repo() -> list:
    return [get_five, get_ten, asset_based_job]
```
    

References:
- https://discuss.dagster.io/t/8204501/does-calling-an-asset-function-from-a-job-cause-the-asset-to
- https://discuss.dagster.io/t/8609918/how-can-i-access-a-materialized-asset-in-a-job

Bonnevie:
This is pretty under-documented, it seems like most places in the docs where this would naturally come up an ordinary python function is used to load some data from disk. But given that assets have internal ops, would it work to do
```python
do_something_special_op(get_five(), get_ten())
```
It's just not obvious to me that this would trigger the I/O manager given that it is an asset definition and not actually an op?

A related headache is if you can run jobs on an asset specified by configuration (i.e. asset key) - I assume the asset loader cannot be used inside a job? E.g. you might have a generic backup or observation routine.

jzgrzebnicki:
What if the asset is partitioned? Does the job will need to have the same partition?

sryza:
Potential API for this:

```python
@job(
     input_assets={
[...]

PR #12091, opened 2023-02-03
allow using SourceAssets to satisfy node inputs in non-asset jobs

sryza:
Addresses https://github.com/dagster-io/dagster/issues/10874

### Summary & Motivation

Several users have asked to be able create jobs that don't produce any assets but do consume assets that are produced elsewhere. This PR enables this use case by allowing users to directly pass SourceAssets as arguments to nodes, in composition. E.g.:

```python
emails_to_send = SourceAsset("emails_to_send", io_manager_key="snowflake")

@op
def send_emails(emails):
    ...


@job
def send_emails_job():
    send_emails(emails_to_send)
```

### Likely followups

- Automatically coerce non-source assets to source assets when used as inputs:
```python
@asset
def emails_to_send():
    ...

@op
def send_emails(emails):
    ...


@job
def send_emails_job():
    send_emails(emails_to_send)
```

- Utility to construct an asset sensor that launches the job wen the assets it reads from have been materialized:

```python
send_emails_after_emails_updated = build_asset_sensor_for_job(send_emails_job)
```
- On the job page in Dagit, include links to the assets it depends on
- On the asset details page in Dagit, include links to the jobs that depend on it

### Additional thoughts

- This is not meant to be the end-all-be-all of running ops after assets.  There's still room to introduce a new concept like "effects", which participate more natively in the asset graph. I see the change implemented here as filling in a clear blank space in our existing concepts - I've observed several users expect that this exists already.
- This would enable dropping support for root input managers.


alangenfeld (commented):
will have to take another pass over the implementation but I think this is pretty slick

schrockn:
This is cool. Seems like it solves an immediate user problem.

A few things.

* It would be great to see docs/README associated with this so we can kicktest how we would explain that to the user. I think that is especially important since we are intermixing concepts here.
* It is a source asset relative to the job, or relative to the code location? Put another way, if I want to depend on an asset that exists as an `AssetsDefinition` in that same code location, do I have to create a `SourceAsset` for the sole purpose of being an input into the job? Is your comment re "Automatically coerce non-source assets to source assets when used as inputs" referring to this?
* What does the API look like when you don't want to load the asset into memory? I can imagine use cases where you want to use this as a generalized "kick of this job when this asset key is materialized" rather than use it with i/o managers. Does `asset_sensor` and `multi_asset_sensor` cover this use case?
* Do you envision this rendering in a job graph view as an upstream node?

sryza:
> if I want to depend on an asset that exists as an AssetsDefinition in that same code location, do I have to create a SourceAsset for the sole purpose of being an input into the job? 

With what's provided in this PR, yes, you would need to do:
```python
my_op(my_asset.to_source_assets[0])
```

> Is your comment re "Automatically coerce non-source assets to source assets when used as inputs" referring to this?

Yes, exactly.

One question here is what the right ergonomics are for multi-assets. Potentially:

```python
my_op(my_multi_asset.to_source_asset("one_of_the_asset_keys"))
```

In this case, the word "source" would be relative to the job, not the code location.

> What does the API look like when you don't want to load the asset into memory? I can imagine use cases where you want to use this as a generalized "kick of this job when this asset key is materialized" rather than use it with i/o managers. Does asset_sensor and multi_asset_sensor cover this use case?

I think `asset_sensor` and `multi_asset_sensor` largely cover this use case, but I think there could be value in Dagster understanding the fact that an individual op within a job uses the data from a particular asset, even if the data isn't being loaded using an IO manager.

With the current PR, the following is valid:

```python
@op(ins={"input1": In(dagster_type=Nothing)})
def my_op():
    ...

@job
def my_job():
    my_op(my_source_asset)
```

The only effect would be that the graph definition knows about the dependency, but we could expose this in Dagit.

> Do you envision this rendering in a job graph view as an upstream node?

I hadn't thought in detail about how this would show up in the job graph view, but I like that idea. Potentially with the same visual presentation that we use for nodes in the asset graph that aren't part of the current selection but are upstream / downstream of the current selection.

schrockn:
I'm a big fan of this direction. 

My major question is around this job-scoped versus location-scoped asset representation issue, and that is why I think writing a README is important here. It is a bit of a conceptual shift since the current definition of SourceAsset in our docs is _"...an asset that will be loaded by (but not updated by) Dagster."_ 

Put another way, it seems a bit odd to have both a `SourceAsset` and an `AssetDefinitions` in memory at the same time that are modeling the same underlying thing. Makes me think that the conceptual model here should be the job relying on the _asset key_ and being agnostic whether it is source or not. 

Also  I/O manager that is the lynchpin here in terms of implementation details, since the pertinent information that SourceAsset adds beyond asset key is loading behavior and `io_manager_key`. But I do think the I/O manager modeling in source asset space is a bit awkward. Specifically if I am a user and I want to find the code that is loading the asset for this job I would need to:

1) Navigate to the the source asset OR asset definitions
2) Look up its `io_manager_key`
3) Then find the actual I/O manager implementation in the particular context.

This is a pre-existing phenomenon, but I think sticks out a lot in this case specifically.

So in summary:

1) It makes complete sense why this RFC is implemented in terms of SourceAssets given the way everything is implemented.
2) That doesn't align with the current definitions of our concepts I think it represents an implicit critique of our core ontology in this area.

sryza:
> That doesn't align with the current definitions of our concepts I think it represents an implicit critique of our core ontology in this area.

This statement feels viscerally correct to me - we mostly think about `SourceAsset`s as living on the boundaries of code locations.

That said, I think the way that we currently define `SourceAsset` in our docs might actually be flexible enough to accommodate the way it's being used in this PR:

image

After this change, I think we would amend the second sentence to: "`SourceAsset`s are used to represent assets that other assets _or jobs_ depend on, in settings where they can't be materialized themselves."

I just wrote up a guide on how software-defined assets relate to ops and graphs. I included a section on this functionality: https://www.notion.so/dagster/How-software-defined-assets-relate-to-ops-and-graphs-3548fa3c685a453c99388a7d7bd3d671#1193d1f743bf49b4a94394e3e622a158.

sryza:
> But I do think the I/O manager modeling in source asset space is a bit awkward

I think it depends on how heterogeneous the storage scheme is across the source assets. For source assets stored in Snowflake using a consistent scheme, I think `io_manager_key="snowflake"` is pretty informative. However, if every source asset requires its own special loading code, I agree that this is awkward.  I think something like a `@loadable_source_asset` decorator would be more ergonomic and clearer in the latter case.

schrockn:
Probably beating a dead horse, but I think further complicating the source asset ontology with "loadable", "observable" and vanilla source assets is quite a cognitive burden.

schrockn:
In terms of moving forward, I suggest:

that we

1. Change the definition of SourceAsset as you suggest. 
2. Figure out what the UI representation of this will be. I think is important when considering the design holistically.
3. Add a section to your notion doc that also hooks up a sensor to make the job actually execute and then consider an API to make that more ergonomic. One concern I have is that people will hook up the dependency and think they are "done" and adding a sensor will be like unnecessary boilerplate.

sryza:
> Probably beating a dead horse, but I think further complicating the source asset ontology with "loadable", "observable" and vanilla source assets is quite a cognitive burden.

Maybe there's a better name - `@source_asset_with_input_manager`?. I don't think of a loadable source asset as a fundamentally different kind of source asset. The idea is just a convenience decorator that condenses a few steps into one.

sryza:
I took off the RFC tag and did a few other things:
- Updated the PR to update the SDA concepts page with the wording above.
- Started a [Slack thread](https://elementl-workspace.slack.com/archives/C03CA4TVCAW/p1675795532133589) to discuss the UI implications. TLDR we'll include an asset node in the job's op graph. TBD whether this will be one of the "foreign asset links" that we show when you've selected a subset of the graph or whether it will be a normal asset node. I lean towards the former, because it makes it clear that the asset won't be materialized as part of the job.
- Added an explanation and some code at the [bottom of the Notion doc](https://www.notion.so/dagster/How-software-defined-assets-relate-to-ops-and-graphs-3548fa3c685a453c99388a7d7bd3d671#1193d1f743bf49b4a94394e3e622a158) about how to use this with a sensor.
- Added more tests and fixed bugs that they uncovered.

schrockn:
> Started a [Slack thread](https://elementl-workspace.slack.com/archives/C03CA4TVCAW/p1675795532133589) to discuss the UI implications. TLDR we'll include an asset node in the job's op graph. TBD whether this will be one of the "foreign asset links" that we show when you've selected a subset of the graph or whether it will be a normal asset node. I lean towards the former, because it makes it clear that the asset won't be materialized as part of the job.

Strongly agree with you here. I think it should be foreign asset link.

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -38,7 +38,7 @@ A software-defined asset includes the following:
...
 | ------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | <PyObject object="asset" decorator /> | A decorator used to define assets.                                                                                                                                                                                           |
-| <PyObject object="SourceAsset" />     | A class that describes an asset, but doesn't define how to compute it. <PyObject object="SourceAsset" />s are used to represent assets that other assets depend on, in settings where they can't be materialized themselves. |
+| <PyObject object="SourceAsset" />     | A class that describes an asset, but doesn't define how to compute it. <PyObject object="SourceAsset" />s are used to represent assets that other assets or jobs depend on, in settings where they can't be materialized themselves. |
```
schrockn, line 41: I think this problem might pre-exist this change, but is this definition in the context of the global asset lineage graph?

python_modules/dagster/dagster/_core/definitions/graph_definition.py
```diff
@@ -739,6 +755,25 @@ def with_hooks(self, hook_defs: AbstractSet[HookDefinition]) -> "PendingNodeInvo
...
+    def resolve_input_to_destinations(
+        self, input_handle: NodeInputHandle
+    ) -> Sequence[NodeInputHandle]:
+        all_destinations: List[NodeInputHandle] = []
+        for mapping in self.input_mappings:
+            if mapping.graph_input_name != input_handle.input_name:
+                continue
+            # recurse into graph structure
+            all_destinations += self.solid_named(
+                mapping.maps_to.solid_name
```
schrockn, line 767: urge to codemod, rising....

python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -1012,3 +1013,52 @@ def get_run_config_schema_for_job(
...
+    """
+    For non-asset jobs that have some inputs that are fed from SourceAssets, constructs an
+    AssetLayer that includes those SourceAssets.
+    """
+    asset_keys_by_node_input_handle: Dict[NodeInputHandle, AssetKey] = {}
+    source_assets_list = []
+    source_asset_keys_set = set()
+    io_manager_keys_by_asset_key: Mapping[AssetKey, str] = {}
+
+    stack: List[Tuple[GraphDefinition, Optional[NodeHandle]]] = [(job_graph_def, None)]
```
schrockn, line 1028: a comment here or a named tuple with named properties (just for documentation) saying that each stack entry is a graph_def and its parent node if it not the root. I searched and line 1031 answer the question for me because the variable names of the tuple unpacking. this is ok for now because it is so close, but code gets moved around etc.

python_modules/dagster/dagster/_core/definitions/run_config.py
```diff
@@ -448,6 +469,7 @@ def define_isolid_field(
     ignored: bool,
     is_using_graph_job_op_apis: bool,
     asset_layer: AssetLayer,
+    input_source_assets: Mapping[str, "SourceAsset"],
```
schrockn, line 472: would be nice to refactor this code to bunch of the "clumps" of parameters that are passed to every function a struct so a change like this isn't as noisy, but definitely don't need to change that for this PR

python_modules/dagster/dagster_tests/core_tests/graph_tests/test_graph.py
```diff
@@ -78,6 +81,103 @@ def the_graph(the_in):
...
+def test_source_asset_input_value():
+    asset1 = SourceAsset("asset1")
+
+    class MyIOManager(IOManager):
+        def handle_output(self, context, obj):
+            ...
+
+        def load_input(self, context):
+            assert context.asset_key == asset1.key
+            return 5
```
schrockn, line 93: as a matter of course when doing tests like these I really like to do a pattern like this:

```
    executed = {}

    class MyIOManager(IOManager):
            ...

            executed["yes"] = True
            assert context.asset_key == asset1.key
            return 5
      
      # later
      
      assert executed["yes"] 
```

Just to ensure that code the framework calls back into is actually called. I've caught more than one bug this way and it proves beyond all doubt to future code readers that the rest is working properly. More than once I've found test suites that asserted against success only and it turns out they have not configured things right to call back into the right user code.

schrockn (changes requested):
Implementation is solid. I'm req'ing changes because the test suite seems a little thin.

Some cases that appear untested:

* Op depends on two source assets
* Op depends on upstream op for one input, source asset for another input
* Two separate ops in the same graph depend on the same source asset

Those are just what immediately comes to mind. I bet there are more interesting ones to test.

python_modules/dagster/dagster/_core/definitions/graph_definition.py
sryza: lol same

python_modules/dagster/dagster_tests/core_tests/graph_tests/test_graph.py
sryza: In this case, I think the fact that `load_input` returns `5` and the op body asserts that the input value is 5 provides pretty strong guarantee. That said, updated the PR with a similar assert to what you described.

docs/content/concepts/assets/software-defined-assets.mdx
sryza: Let me know if this doesn't answer your question:

My thinking behind the definition here is that the "source asset" designation is setting-dependent. An asset isn't fundamentally a source asset (because all data is generated by code in some way). Rather, `SourceAsset` is a class that's used to represent an asset in settings where it's not necessary to know how the asset is computed:
- In a job that reads from the asset but doesn't materialize it
- In a `Definitions` object / code location
- There are no Python classes in play when looking at the global asset graph in Dagit, but I think it's reasonable to use the term "source asset" in that context for assets that don't have a materialization functions defined in any code location

sryza:
@schrockn back to your queue. I added the tests you requested, plus one additional test. And beefed up one of the existing tests a tiny bit.

docs/content/concepts/assets/software-defined-assets.mdx
schrockn: Great that all makes sense to me. 👍🏻

schrockn (approved):
Most excellent. Excited to get this in the hands of users.

PR #12095, opened 2023-02-03
[dagit] Mocks, storybooks and a test for BackfillTable

bengotow:
### Summary & Motivation

Hey gang, curious to get your thoughts about this. I added a storybook for BackfillTable that uses mock data to accurately test all of the different Backfill Target and Backfill Status states, along with the async loading of the per-row data.

This table isn't all that interactive - we could write jest tests that say "It renders 'complete' for a complete backfill", but I don't think they add much value over the storybook if we can wire up https://www.chromatic.com/features/test to put a snapshot of the Storybook file on this PR and highlight differences.

I wrote one test which tests that clicking a "Failed" status tag opens a modal showing the error, because that seemed like interactivity worthy of a test. SInce the mocks are broken out into a separate file, this test can re-use the test data from storybook which is kinda cool.

image

### DIFF
--- a/js_modules/dagit/packages/core/src/instance/BackfillTable.mocks.ts
+++ b/js_modules/dagit/packages/core/src/instance/BackfillTable.mocks.ts
@@ -1,0 +1,419 @@
+import {MockedResponse} from '@apollo/client/testing';
+
+import {BulkActionStatus, RunStatus} from '../graphql/types';
+
+import {
+  SINGLE_BACKFILL_STATUS_COUNTS_QUERY,
+  SINGLE_BACKFILL_STATUS_DETAILS_QUERY,
+} from './BackfillRow';
+import {SingleBackfillCountsQuery, SingleBackfillQuery} from './types/BackfillRow.types';
+import {BackfillTableFragment} from './types/BackfillTable.types';
+
+function buildTimePartitionNames(start: Date, count: number) {
+  const results: string[] = [];
+  for (let ii = 0; ii < count; ii++) {
+    start.setMinutes(start.getMinutes() + 15);
+    results.push(start.toISOString());
+  }
+  return results;
+}
+
+export const BackfillTableFragmentRequested2000AssetsPure: BackfillTableFragment = {
+  backfillId: 'qtpussca',
+  status: BulkActionStatus.REQUESTED,
+  numPartitions: 2000,
+  timestamp: 1675196684.587724,
+  partitionSetName: null,
+  partitionSet: null,
+  error: null,
+  numCancelable: 0,
+  partitionNames: buildTimePartitionNames(new Date('2020-01-01'), 2000),
+  assetSelection: [
+    {
+      path: ['global_graph_asset'],
+      __typename: 'AssetKey',
+    },
+  ],
+  __typename: 'PartitionBackfill',
+};
+
+export const BackfillTableFragmentRequested2000AssetsPureStatus: MockedResponse<SingleBackfillCountsQuery> = {
+  request: {
+    query: SINGLE_BACKFILL_STATUS_COUNTS_QUERY,
+    variables: {
+      backfillId: 'qtpussca',
+    },
+  },
+  result: {
+    data: {
+      __typename: 'DagitQuery',
+      partitionBackfillOrError: {
+        backfillId: 'qtpussca',
+        partitionStatusCounts: [
+          {
+            runStatus: RunStatus.NOT_STARTED,
+            count: 108088,
+            __typename: 'PartitionStatusCounts',
+          },
+          {
+            runStatus: RunStatus.SUCCESS,
+            count: 71,
+            __typename: 'PartitionStatusCounts',
+          },
+          {
+            runStatus: RunStatus.FAILURE,
+            count: 10,
+            __typename: 'PartitionStatusCounts',
+          },
+        ],
+        __typename: 'PartitionBackfill',
+      },
+    },
+  },
+};
+
+export const BackfillTableFragmentCancelledAssetsPartitionSet: BackfillTableFragment = {
+  backfillId: 'tclwoggv',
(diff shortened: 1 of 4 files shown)


js_modules/dagit/packages/core/src/instance/BackfillTable.mocks.ts
```diff
@@ -0,0 +1,419 @@
...
+  assetSelection: [
+    {
+      path: ['multipartitioned_asset'],
+      __typename: 'AssetKey',
+    },
+  ],
+  __typename: 'PartitionBackfill',
+};
+
+export const BackfillTableFragmentFailedErrorStatus: MockedResponse<SingleBackfillQuery> = {
```
bengotow, line 165: It took about half an hour to create these files by copy-pasting requests / responses out of the Chrome network tab and cutting them down to be representative examples only.

One annoying part is that TypeScript expects it's enums for things like RunStatus and won't accept the string "COMPLETED" instead of `RunStatus.Completed`, so it takes a bit of manual tweaking...

js_modules/dagit/packages/core/src/instance/BackfillTable.mocks.ts
```diff
@@ -0,0 +1,419 @@
...
+  assetSelection: [
+    {
+      path: ['global_graph_asset'],
+      __typename: 'AssetKey',
+    },
+  ],
+  __typename: 'PartitionBackfill',
+};
+
+export const BackfillTableFragmentRequested2000AssetsPureStatus: MockedResponse<SingleBackfillCountsQuery> = {
```
bengotow, line 40: Very longwinded name here, but was thinking these mocks should have some sort of consistent naming...  "fragment name + some descriptive info about this test case"?

salazarm:
> we could write jest tests that say "It renders 'complete' for a complete backfill", but I don't think they add much value over the storybook

I think they can still just confirm that these state transitions are connected. But agreed that a visual regression test would provide higher value

js_modules/dagit/packages/core/src/instance/BackfillTable.mocks.ts
salazarm: +1

salazarm (approved):
I actually had no idea about MockedResponse. This is great, I'm going to use this for all my tests

js_modules/dagit/packages/core/src/instance/BackfillTable.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+// This file must be mocked because Jest can't handle `import.meta.url`.
+jest.mock('../graph/asyncGraphLayout', () => ({}));
+
+describe('BackfillTable', () => {
+  it('allows you to click "Failed" backfills for error details', async () => {
+    jest.spyOn(Alerting, 'showCustomAlert');
+
+    render(
+      <TestProvider>
```
hellendag, line 22: Do we still need the full `TestProvider`? My hope was that we might not need it anymore for some/most/all cases that are using `MockedProvider` for Apollo needs.

js_modules/dagit/packages/core/src/instance/BackfillTable.mocks.ts
```diff
@@ -0,0 +1,419 @@
+import {MockedResponse} from '@apollo/client/testing';
```
hellendag, line 1: The `mocks.ts` extension seems like a good approach to me.

js_modules/dagit/packages/core/src/instance/BackfillTable.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+  BackfillTableFragmentRequested2000AssetsPureStatus,
+  BackfillTableFragments,
+} from './BackfillTable.mocks';
+
+// eslint-disable-next-line import/no-default-export
+export default {component: BackfillTable};
+
+export const GeneralStates = () => {
+  return (
+    <StorybookProvider>
```
hellendag, line 21: Similar comment to elsewhere -- do we need the full `StorybookProvider`, if we're not using the full mocked Apollo provider?

hellendag (approved):

js_modules/dagit/packages/core/src/instance/BackfillTable.mocks.ts
```diff
@@ -0,0 +1,419 @@
...
+  numPartitions: 100,
+  timestamp: 1674774274.343382,
+  partitionSetName: null,
+  partitionSet: null,
+  error: {
+    __typename: 'PythonError',
+    message:
+      'dagster._core.errors.DagsterLaunchFailedError: Tried to start a run on a server after telling it to shut down\n',
+    stack: [
+      '  File "/Users/bengotow/Work/F376/Projects/dagster/python_modules/dagster/dagster/_daemon/backfill.py", line 34, in execute_backfill_iteration\n    yield from execute_asset_backfill_iteration(backfill, workspace, instance)\n',
```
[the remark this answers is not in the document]
bengotow: Ahh good call can definitely do this

js_modules/dagit/packages/core/src/instance/BackfillTable.stories.tsx
bengotow: Hmm, it looks like the two elements that this still requires are the `CustomAlertProvider` and the `WorkspaceProvider` (to build some misc URLs in some backfill table states - not super important).

Maybe another option would be exporting a StorybookProvider that doesn't include the apollo part, and then having each page use the version of Apollo mocking it wants?

bengotow:
Hey gang thanks for the feedback on this! Going to apply PR comments but wait to try breaking apart the StorybookProvider / TestProvider until another PR. It loos like this still requires a few of the things there, but agree it's weird to stack two apollo providers...

PR #12032, opened 2023-02-01
[prototype] Link to asset definitions on GitHub from prod/Cloud Dagit

benpankow:
## Summary

Rough follow-up to #12009 which links out to asset definitions in GitHub for repositories which have `git` metadata configured (e.g. Cloud repos pushed to via GH action).

I think this could be especially neat for branch depls (to view changes in assets) and logged as part of history (e.g. being able to go back to an old materialization and jump to the exact version of code which it was built by)

https://user-images.githubusercontent.com/10215173/216195386-39fde623-0e69-4d2b-8117-6676ad3d6b08.mov

PR #12096, opened 2023-02-03
[1/n] Optionally attach definition file & line number information to assets & ops

benpankow:
## Summary

First PR in a stack reworking #12009 into something landable, at least as an experiment.

If enabled in dagster config, uses `inspect` to attach metadata to Assets and Ops identifying where they were defined (line # and absolute filepath).

The intention is for this to be used locally, PRs later in the stack will add [editor backlinks from Dagit](https://github.com/dagster-io/dagster/pull/12097) which will show up in a local dev environment.

## Test Plan

tested locally, introduced unit tests

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
@@ -21,12 +21,18 @@ import dagster._check as check
 from dagster._annotations import deprecated_param, experimental_param
 from dagster._builtins import Nothing
 from dagster._config import UserConfigSchema
-from dagster._core.decorator_utils import get_function_params, get_valid_name_permutations
+from dagster._core.decorator_utils import (
+    get_function_params,
+    get_valid_name_permutations,
+)
 from dagster._core.definitions.asset_dep import AssetDep, CoercibleToAssetDep
 from dagster._core.definitions.auto_materialize_policy import AutoMaterializePolicy
 from dagster._core.definitions.config import ConfigMapping
 from dagster._core.definitions.freshness_policy import FreshnessPolicy
-from dagster._core.definitions.metadata import ArbitraryMetadataMapping, MetadataUserInput
+from dagster._core.definitions.metadata import (
+    ArbitraryMetadataMapping,
+    MetadataUserInput,
+)
 from dagster._core.definitions.partition_mapping import PartitionMapping
 from dagster._core.definitions.resource_annotation import (
     get_resource_args,
--- a/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
@@ -21,18 +21,12 @@ import dagster._check as check
 from dagster._annotations import deprecated_param, experimental_param
 from dagster._builtins import Nothing
 from dagster._config import UserConfigSchema
-from dagster._core.decorator_utils import (
-    get_function_params,
-    get_valid_name_permutations,
-)
+from dagster._core.decorator_utils import get_function_params, get_valid_name_permutations
 from dagster._core.definitions.asset_dep import AssetDep, CoercibleToAssetDep
 from dagster._core.definitions.auto_materialize_policy import AutoMaterializePolicy
 from dagster._core.definitions.config import ConfigMapping
 from dagster._core.definitions.freshness_policy import FreshnessPolicy
-from dagster._core.definitions.metadata import (
-    ArbitraryMetadataMapping,
-    MetadataUserInput,
-)
+from dagster._core.definitions.metadata import ArbitraryMetadataMapping, MetadataUserInput
 from dagster._core.definitions.partition_mapping import PartitionMapping
 from dagster._core.definitions.resource_annotation import (
     get_resource_args,
--- a/python_modules/dagster/dagster/_core/definitions/decorators/op_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/op_decorator.py
@@ -1,10 +1,17 @@
+import inspect
+import os
+import sys
+from contextlib import contextmanager
 from functools import lru_cache, update_wrapper
 from inspect import Parameter
+from types import ModuleType
 from typing import (
     TYPE_CHECKING,
     AbstractSet,
     Any,
     Callable,
+    Dict,
+    Generator,
     List,
     Mapping,
     NamedTuple,
@@ -41,6 +48,38 @@ from ..utils import DEFAULT_OUTPUT
 if TYPE_CHECKING:
     from ..op_definition import OpDefinition
 
+CODE_ORIGIN_TAG_NAME = "__code_origin"
+
+
+# global var flag to enable/disable attaching code origin to ops and assets
+CODE_ORIGIN_ENABLED = [True]
+
+
+@contextmanager
+def do_not_attach_code_origin() -> Generator[None, None, None]:
+    """Disables attaching code origin to ops and assets. This is useful for testing, because code
+    origin can change from environment to environment and break snapshot tests.
+    """
+    CODE_ORIGIN_ENABLED[0] = False
+    try:
+        yield
+    finally:
+        CODE_ORIGIN_ENABLED[0] = True
+
+
+def is_code_origin_enabled():
+    return CODE_ORIGIN_ENABLED[0]
+
+
+def _get_root_module(module: ModuleType) -> ModuleType:
+    module_name_split = module.__name__.split(".")
+    for i in range(1, len(module_name_split)):
+        try:
+            return sys.modules[".".join(module_name_split[:i])]
+        except KeyError:
+            continue
+    return module
+
 
 class _Op:
     def __init__(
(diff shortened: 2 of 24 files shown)


benpankow:
* master
  * **PR #12096** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12096" target="_blank">Graphite</a>  👈
    * **PR #12097** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12097" target="_blank">Graphite</a> 
      * **PR #12098** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12098" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12096?utm_source=stack-comment).

sryza:
1. I think this would be very cool
2. Gating it on a special environment variable feels dirty
3. If we could provide the git hash as well, I think this would be very useful in production environments for linking to Github

benpankow:
> 2. Gating it on a special environment variable feels dirty
What sort of solution would feel better here? It could be attached to a dagster.yaml & toggled in the ephemeral case by a CLI flag?

> 3. If we could provide the git hash as well, I think this would be very useful in production environments for linking to Github

https://github.com/dagster-io/dagster/pull/12032 attempts to do something like this using code location metadata

sryza:
> What sort of solution would feel better here? It could be attached to a dagster.yaml & toggled in the ephemeral case by a CLI flag?

My impulse is to say that we should just eat the cost. Do we have a sense of how much this is likely to be in real time? I would guess that `@asset` build time is a pretty small % of overall time? Would it help if we deferred it to when we construct the `Definitions` object?

benpankow:
> My impulse is to say that we should just eat the cost. Do we have a sense of how much this is likely to be in real time? I would guess that `@asset` build time is a pretty small % of overall time? Would it help if we deferred it to when we construct the `Definitions` object?

Hmm, yeah it's probably worth running a few more benchmarks. The main motivation here was just to get something in, temporarily gated, which we could use to test/evaluate without having to commit to the inspect cost. Inspect metaprogramming seems scary in the critical path, so I'm a bit wary in general beyond just the perf impact.

sryza:
> Inspect metaprogramming seems scary in the critical path

FWIW if you grep for it, you can see we already use it in a bunch of different places. So I'm not super worried about it unless this usage is somehow qualitatively different?

benpankow:
Rebased + updated the PR, should be in a good state to review.

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -354,6 +385,7 @@ def __call__(self, fn: Callable) -> AssetsDefinition:
                 tags={
                     **({"kind": self.compute_kind} if self.compute_kind else {}),
                     **(self.op_tags or {}),
+                    **code_origin_tag,
```
sryza, line 388: to avoid this function getting huge, what do you think about making a helper function and just having this be `**get_code_origin_tags()`?

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -343,6 +364,16 @@ def __call__(self, fn: Callable) -> AssetsDefinition:
 
             op_required_resource_keys = decorator_resource_keys - arg_resource_keys
 
+            code_origin_tag = (
```
sryza, line 367: It's a little weird to have a variable that's a dictionary have a name that's a singular noun. Also, what are the tradeoffs of having a single tag with a JSON blob under it vs. separate tags for the file and line?

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -666,7 +698,11 @@ def build_asset_ins(
 
         ins_by_asset_key[asset_key] = (
             input_name.replace("-", "_"),
-            In(metadata=metadata, input_manager_key=input_manager_key, dagster_type=dagster_type),
+            In(
```
sryza, line 701: superfluous change?

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
benpankow: actually we don't even use this asset metadata in the UI, I'll just remove it

python_modules/dagster/dagster_tests/execution_tests/dynamic_tests/test_dynamic_execution.py
```diff
@@ -12,84 +12,77 @@
...
     reconstructable,
 )
+from dagster._core.definitions.decorators.op_decorator import do_not_attach_code_origin
 from dagster._core.errors import DagsterExecutionStepNotFoundError
 from dagster._core.execution.api import create_execution_plan
 from dagster._core.execution.plan.state import KnownExecutionState
 from dagster._core.test_utils import instance_for_test
 from dagster._utils.merger import merge_dicts
 
+with do_not_attach_code_origin():
```
yuhan, line 22: per offline convo, let's move this to a pytest.fixture so it's easier to use in test and change less test code.

python_modules/dagster/dagster/_core/definitions/decorators/op_decorator.py
```diff
@@ -40,6 +47,38 @@
...
 
+CODE_ORIGIN_TAG_NAME = "__code_origin"
+
+
+# global var flag to enable/disable attaching code origin to ops and assets
+CODE_ORIGIN_ENABLED = [True]
+
+
+@contextmanager
+def do_not_attach_code_origin() -> Generator[None, None, None]:
```
yuhan, line 58: maybe worth some comment to explain this type hint. (but if we move to pytest.fixture, we probably don't need this generator at all)

yuhan (commented):
overall looking close to land. can we also make sure the later stack is well feature gated?

python_modules/dagster/dagster/_core/definitions/decorators/op_decorator.py
```diff
@@ -41,6 +48,38 @@
 if TYPE_CHECKING:
     from ..op_definition import OpDefinition
 
+CODE_ORIGIN_TAG_NAME = "__code_origin"
```
sryza, line 51: We normally prefix system tags with "dagster/" - see storage/tags.py

python_modules/dagster/dagster/_core/definitions/decorators/op_decorator.py
```diff
@@ -75,6 +114,44 @@ def __init__(
...
+            origin_line = inspect.getsourcelines(fn)[1]
+
+            # Get the base module that the op function is defined in
+            # and find the filepath to that module
+            module = inspect.getmodule(fn)
+            root_module = _get_root_module(module) if module else None
+            path_to_module_root = (
+                os.path.abspath(os.path.dirname(os.path.dirname(root_module.__file__)))
+                if root_module and root_module.__file__
+                else "/"
```
sryza, line 143: This assumes unix, which won't always be the case, right?

python_modules/dagster/dagster/_core/definitions/decorators/op_decorator.py
```diff
@@ -75,6 +114,44 @@ def __init__(
...
+        """
+        if not is_code_origin_enabled():
+            return {}
+
+        # Attempt to fetch information about where the op is defined in code,
+        # which we'll attach as a tag to the op
+        cwd = os.getcwd()
+        origin_file: Optional[str] = None
+        origin_line = None
+        try:
```
sryza, line 131: To make it distinguish the Dagster-specific behavior from the underlying origin-scraping implementation, it would be helpful to factor out the latter into a utility function, perhaps in inference.py or some utils file.

sryza (commented):
Finally revisiting this at long last! Sorry for being so delinquent.

Is my understanding correct that we're exposing this as an instance-level config, and that it's off by default? If so, why are all the test changes and `do_not_attach_code_origin` required?

Separate thing: I'm trying to grok how this would work in a prod environment where we want to show links to a Github repo.  How would that work? Would we want to directly attach the Github URL? Attach the absolute path and do some transformation elsewhere? Are there tradeoffs to this two approaches?

Third thing: did you end up doing any investigation of the perf impact?

Fourth thing: how does someone customize this? E.g. for dbt we'd probably want links to the dbt sql files instead of the `@dbt_assets` invocation.

salazarm:
Thanks for getting this working. Re: customization I would love if we made it possible to support more than 1 type of link. I could imagine wanting both a link to vscode but also a link to dbt cloud or some other cloud provider.

Some of these customizations might be complex though, the most flexible approach I imagine would be to let the user run some python code that determines the link given the object and its metadata. That code could then either run at startup, or run lazily and be cached at runtime or just run every time we need the link. There are different tradeoffs for each approach and I'm not sure which would be the right tradeoff.

I'm envisioning it could be something like:

```
@link(name="vscode")
def vscode_link(object: Asset | Sensor | Op | Schedule | etc..., metadata: PythonInstrospectedMetadata):
  ....


@link(name="dbt_cloud")
def dbt_cloud_link(object: Asset | Sensor | Op | Schedule | etc..., metadata: PythonInstrospectedMetadata):
  ....

```

Then users could tag the asset/sensor/schedule/job with `CODE_ORIGIN_TAG_NAME: "vscode, dbt_cloud"`

sryza:
I drew out a diagram of my understanding of the dataflow for how we'd ultimately want this to work. Does this look accurate?

image

benpankow:
> Is my understanding correct that we're exposing this as an instance-level config, and that it's off by default? If so, why are all the test changes and do_not_attach_code_origin required?

Attaching the metadata is on by default, though I don't feel strongly about this behavior. The threading through of code links in UI is toggled by the `dagster.yaml` instance setting.

> Separate thing: I'm trying to grok how this would work in a prod environment where we want to show links to a Github repo. How would that work? Would we want to directly attach the Github URL? Attach the absolute path and do some transformation elsewhere? Are there tradeoffs to this two approaches?

See https://github.com/dagster-io/dagster/pull/14711 for the current impl for cloud - we get the GitHub URL off the repo settings and combine that with some path transformation logic. As we discussed offline

> Third thing: did you end up doing any investigation of the perf impact?

From initial PR desc from way backwhen, though I might do some more testing:

```
... a non-scientific pyspy investigation showed it added a solid 20-25% to @asset build time ...
```


> Fourth thing: how does someone customize this? E.g. for dbt we'd probably want links to the dbt sql files instead of the @dbt_assets invocation.

Right now there's no customization built in. I think your diagram shows the approach we'd want to take: Dagster will infer a filename and line number, and user code (or Dagster libraries) can override or provide additional filenames and line numbers.

For the DBT case, we could for example have 3 sets of files: the `@dbt_assets` decorator, dbt sql file, and the schema definition in the relevant `schema.yml` file.

We also likely want the ability for users to provide custom URLs. There's a broader question of how the APIs for this should look - just attaching string asset metadata, or something more complex.

sryza:
Thinking that something like [this](https://github.com/dagster-io/internal/discussions/8216) could help us make the jump from local paths to remote github links:

```python
from dagster_cloud import local_source_link_metadata_to_github_link

defs = Definitions(assets=...)

if not os.environ["IS_LOCAL"]:
    defs = defs.with_asset_metadata(local_source_link_to_github_link)
```

sryza:
> I still like the idea of letting the user run arbitrary python code to determine the link. It gives users the most freedom and they could do things that might require looking up an object in another catalog they own to link to it in that catalog or maybe link to a recent run in DBT, etc.

I definitely agree that we should enable running arbitrary Python code to determine the link.  In the proposal above, `with_asset_metadata` accepts an arbitrary Python function for determining the link.

But I think there's cost to adding a new abstraction which executes this code at some time other than normal import time:
1. More concepts for users to understand
2. Users need to understand when it actually gets executed

That said, `with_asset_metadata` could potentially return a lazily-instantiated `Definitions` object?

Issue #11752, opened 2023-01-17
[dagster-dbt] Emit Outputs/AssetMaterialization events as the dbt command executes, rather than waiting for run_results.json


OwenKephart:
### What's the use case?

Currently, we wait for the run_results.json file to become available before we emit any outputs. For very large dbt projects, this has two issues:

1. Large dbt projects (generally) take a while to run, meaning tables may be materialized well before their materialization event is created.
2. Each output requires three events to be emitted (OUTPUT, ASSET_MATERIALIZATION, STEP_OUTPUT). If the time it takes to write each event is high, then writing the (potentially thousands) of events at the end of dbt execution can take a non-negligible amount of time.

### Ideas of implementation

We'd need to add a new method, akin to `execute_cli`, that executes a given command/arguments in a streaming manner. We'd then consume from that stream within the body of the op. By default, dbt will provide us logs in a JSON format, so it likely would be possible to parse this json output to detect when models are materialized (or fail to be materialized).

This behavior should be made optional, as it would preclude us from being able to attach certain types of metadata (that only becomes available at the end of execution) to the outputs.

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

NiallRees:
Hi @OwenKephart, hope you don't mind me asking a question on this issue but it seemed the most appropriate place. We're now using this feature, and I'm wondering if it's possible to view each model's materialization in the overall run timing chart, even though it's technically a single dbt op. So rather than a single dbt bar per the screenshot, we'd see a separate bar for each model. dbt Cloud does this and it makes it really easy to identify bottlenecks in the DAG.

What we currently see
<img width="836" alt="image" src="https://github.com/dagster-io/dagster/assets/23722609/bdb64286-4cf3-4c8d-83f2-49a3b98b29d2">

What would be nice to see
![image](https://github.com/dagster-io/dagster/assets/23722609/b3d7b95e-9e1a-4d84-8e3b-29ba73073536)

sryza:
@NiallRees @karakanb - I filed an issue to track this request: https://github.com/dagster-io/dagster/issues/14559

NiallRees:
Thank you @sryza !

egeyumlu-cl:
[...]

PR #12100, opened 2023-02-03
[dagster-dbt] [2/2] Add ability to stream events while executing dbt assets

OwenKephart:
### Summary & Motivation

https://github.com/dagster-io/dagster/issues/11752


python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
```diff
@@ -603,6 +746,9 @@ def load_assets_from_dbt_project(
         display_raw_sql (Optional[bool]): [Experimental] A flag to indicate if the raw sql associated
             with each model should be included in the asset description. For large projects, setting
             this flag to False is advised to reduce the size of the resulting snapshot.
+        stream_events (Optional[bool]): [Experimental] A flag to indicate if events should be
```
sryza, line 749: Why would someone ever not want this?

OwenKephart: well dbt versions < 1.4 do not have sufficient metadata attached to their logs in order to produce these events, but I guess we should just do the swapping of behavior ourselves (checking the dbt version)

sryza (approved):
Does the modified test verify that we're streaming events?

Also, the title of this PR should change, right?

PR #12105, opened 2023-02-04
[pythonic config] Add disabled tests for various more complex config schemas

benpankow:
## Summary

Adds a series of pytest-skipped tests for patterns which the Pythonic config system doesn't yet support. Stacked PRs will start to tackle these.

PR #12106, opened 2023-02-04
[pythonic config] Correctly handle pydantic List types in config

benpankow:
## Summary

Implements support for lists in Pythonic config objects.

## Test Plan

Unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_config/structured_config.py
+++ b/python_modules/dagster/dagster/_config/structured_config.py
@@ -13,7 +13,7 @@ except ImportError:
 
 
 from abc import ABC, abstractmethod
-from typing import Any, Callable, Dict, Optional, Type, cast
+from typing import Any, Optional, Type, cast
 
 from pydantic import BaseModel, Extra
 from pydantic.fields import SHAPE_LIST, SHAPE_SINGLETON, ModelField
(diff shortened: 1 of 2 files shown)


benpankow:
* master
  * **PR #12105** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12105" target="_blank">Graphite</a> 
    * **PR #12106** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12106" target="_blank">Graphite</a>  👈
      * **PR #12107** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12107" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12106?utm_source=stack-comment).

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -216,20 +216,32 @@ def create_io_manager_to_pass_to_user_code(self, context) -> IOManager:
         return self
 
 
+WRAPPER_TYPES: Dict[int, Callable[[ConfigType], ConfigType]] = {
+    SHAPE_SINGLETON: lambda x: x,
+    SHAPE_LIST: Array,
+}
```
schrockn, line 222: It would be helpful to wrap this accessing this in a function and then explaining what is happening here. Questions I have:

1) Why is the key an `int`? I can infer that pydantic assigns int ids to things, but I'd rather see that laid out specifically
2) Why is this called "wrapped"? Doesn't seem like anything is being wrapped, but mapped. Let's work on a better name.

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -319,7 +331,9 @@ def _safe_is_subclass(cls: Any, possible_parent_cls: Type) -> bool:
 
 
 def infer_schema_from_config_class(
-    model_cls: Type[Config], description: Optional[str] = None
+    model_cls: Type[Config],
+    description: Optional[str] = None,
+    wrap_config_type_fn: Optional[Callable[[ConfigType], ConfigType]] = None,
```
schrockn, line 336: passing in a callback that gets invoked at the end of this function feels unnecessarily clever. would prefer to see an approach where callsites manage this, even if it involves a little code duplication

schrockn (changes requested):

python_modules/dagster/dagster/_config/structured_config.py
benpankow: Moved to a function instead of indexing into this map. Aliased the `int` type to something more descriptive as well.

Hopefully the `wrapped` name makes a bit more sense with the docstring, it's because it's wrapping the inferred config type in whichever data structure is inferred from the Pydantic shape (wrapped in a list, map, etc).

python_modules/dagster/dagster/_config/structured_config.py
```diff
@@ -216,26 +217,45 @@ def create_io_manager_to_pass_to_user_code(self, context) -> IOManager:
...
 
 
+PydanticShapeType: TypeAlias = int
+
+
+def _wrap_config_type(shape_type: PydanticShapeType, config_type: ConfigType) -> ConfigType:
+    """
+    Based on a Pydantic shape type, wraps a config type in the appropriate Dagster config wrapper.
+    For example, if the shape type is a Pydantic list, the config type will be wrapped in an Array.
+    """
```
schrockn, line 227: 👍🏻 this is much clearer

schrockn (approved):
excellent

PR #12113, opened 2023-02-05
Include logs in failure message for ECS monitoring failures

gibsondan:
Helps diagnose issues without needing to check ECS logs.

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws/ecs/utils.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/ecs/utils.py
@@ -56,3 +56,40 @@ def task_definitions_match(
         return False
 
     return True
+
+
+def get_task_logs(ecs, logs_client, cluster, task_arn, container_name, limit=10):
+    task = ecs.describe_tasks(cluster=cluster, tasks=[task_arn]).get("tasks")[0]
+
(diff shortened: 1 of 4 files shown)


python_modules/libraries/dagster-aws/dagster_aws/ecs/utils.py
```diff
@@ -56,3 +56,40 @@ def task_definitions_match(
...
+    container_name = container_definition.get("name")
+    task_id = task_arn.split("/")[-1]
+
+    log_stream = f"{log_stream_prefix}/{container_name}/{task_id}"
+
+    events = logs_client.get_log_events(
+        logGroupName=log_group, logStreamName=log_stream, limit=limit
+    ).get("events")
+
+    return [event.get("message") for event in events]
```
jmsanders, line 95: How constrained is this? Presumably `get_log_events` is already paginated but I just want to double check.

jmsanders (approved):

python_modules/libraries/dagster-aws/dagster_aws/ecs/utils.py
gibsondan: there's a limit arg that defaults to 10

PR #12116, opened 2023-02-05
[dagit] Add Storybook coverage of PartitionHealthSummary rendering

bengotow:
### Summary & Motivation

This PR adds a ton of mock data and a storybook that renders the PartitionHealthSummary component. I'm hoping storybook can be the primary mechanism for building and testing this UI because it takes a lot of time to get assets into the required states to really test it locally. 

I'm hoping I can re-use these same mocks to build tests for the materialize modal itself, since the status component / graphql request is the same  👀

The storybook shows two states for each asset - one with a selection made and one without. When there are two dimensions, making selections on one dimension narrows the states on the other dimension to the subset of keys you've selected. (eg: Given just "TN, VA", what's the health by date?) This feels really nice in the UI but is a bit confusing to look at statically.

The mock data in the storybook is pulled from the network tab, and I shortened it by writing a few helpers that generate date partition keys.

image

I'm really hoping we can wire up something like Chromatic (and it works as advertised) because I think looking for visual changes in this storybook would be an A++ way of automated testing partition health.


salazarm (approved):
i like

PR #12127, opened 2023-02-06
[dagster-airflow] drop unique_id param

Ramshackle-Jamathon:
### Summary & Motivation

this pr changes the naming scheme for converted jobs/ops.

Jobs are now just the name of the dag they are based on

Ops are now `${dag_id}__${task_id}`

no more magic int suffixes and `airflow_` prefix


schrockn (dismissed):
Makes sense to me. Can you add a couple unit tests? If nothing else than to document the naming scheme.

yuhan (approved):

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_asset_factory.py
```diff
@@ -47,7 +50,10 @@ def find_upstream_dependency(node_name: str) -> None:
             match = False
             # find any assets produced by upstream nodes and add them to the internal asset deps
             for asset_key in task_ids_by_asset_key:
-                if forward_node.replace("airflow_", "") in task_ids_by_asset_key[asset_key]:
+                if (
+                    forward_node.replace(f"{normalized_name(dag.dag_id)}__", "")
```
schrockn, line 54: we really don't have task ID in a proper data structure here anywhere?

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_asset_factory.py
```diff
@@ -60,25 +66,28 @@ def find_upstream_dependency(node_name: str) -> None:
...
+            if f"result_{normalized_name(dag.dag_id, task_id)}" in internal_asset_deps:
+                internal_asset_deps[f"result_{normalized_name(dag.dag_id, task_id)}"].update(
+                    asset_upstream_deps
+                )
             else:
-                internal_asset_deps[f"result_airflow_{task_id}"] = asset_upstream_deps
-            # internal_asset_deps[f"result_airflow_{task_id}"] = asset_upstream_deps
+                internal_asset_deps[
+                    f"result_{normalized_name(dag.dag_id, task_id)}"
+                ] = asset_upstream_deps
```
schrockn, line 90: I would feel marginally less terrible about this if we had a helper function to wrap all the identical f string generation

schrockn:
```
def output_name_for_task_in_dag(dag_id, task_id):
   return f"result_{normalized_name(dag.dag_id, task_id)}"
```

Actually making a more proper data structure for this would be a great follow up

schrockn: Also using `@cached_method` would make me feel less bad as this is doing a regex

python_modules/libraries/dagster-airflow/dagster_airflow/dagster_asset_factory.py
schrockn: nope we don't. just f strings all the way down instead of having dict lookups

schrockn (approved):
Would have definitely preferred using this as an excuse to have `_build_asset_dependencies` use a real data structure rather than continuous string mangling, but it was like that when you found it.

PR #12132, opened 2023-02-06
[docs] Add note about multi-dimensionally partitioned assets requiring database migration

NicolasPA:
### Summary & Motivation

If some pre 1.1.1 user wants to try multi-dimensionally partitioned assets, Dagit will fail displaying the description of the asset with a GraphQL error and the user won't be able to select partitions for materialization.
This is because it requires migrating the instance as stated in https://github.com/dagster-io/dagster/releases/tag/1.1.1.
The GraphQL error does mention the need to run the migration, but the reason is not clear.

I think it would be very helpful to add this information in the documentation as a note, hence this PR.

https://docs.dagster.io/concepts/partitions-schedules-sensors/partitions#multi-dimensionally-partitioned-assets

Please make sure the formatting is correctly displayed on the website, especially the backticks.

### DIFF
--- a/docs/content/concepts/partitions-schedules-sensors/partitions.mdx
+++ b/docs/content/concepts/partitions-schedules-sensors/partitions.mdx
@@ -94,6 +94,9 @@ If you're using the default I/O manager, materializing partition `2022-07-23` of
 <Note>
   This feature is <strong>experimental</strong>.
 </Note>
+<Note>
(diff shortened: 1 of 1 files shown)


NicolasPA:
This a 1.5 year old PR related to migration documentation, so it's probably not relevant anymore.

PR #12134, opened 2023-02-06
[dagster-airflow] use in_process execution by default

Ramshackle-Jamathon:
### Summary & Motivation

when doing multiprocess execution each op will perform a slow no-op of verifying that the airflow db is initialized, when doing all ops in the same process this is avoided and can dramatically speed up execution.

multiprocess: https://www.loom.com/share/4ac5eb3c3f604f00a017601bf6f9d994
inprocess: https://www.loom.com/share/2131eb722b124554ae396bee5a508ef5

~10 seconds vs ~50 seconds in whats a fairly contrived example


schrockn (changes requested):
I don't think this is a good idea. Doesn't it effectively hardcode the executor to be the in-process executor?

1) There is no parallelism
2) It reduces isolation

However what I do think this exposes is that it is nearly impossible for users to customize the behavior of the generated jobs.

One should be able to pass through metadata, op_retry_policy, version_strategy, config.

It might be sensible to add a job_kwargs argument for now and just splat all the arguments into the JobDefinition constructor

Ramshackle-Jamathon:
@schrockn pr for kwargs https://github.com/dagster-io/dagster/pull/12179

I don't think this prevents users from overriding, in testing i was still able to set multi-process execution from the launch pad.  I will try to get some unit tests to validate that though.

I do think that the current performance hit of the multi-process executor will be annoying for users as they iterate locally but there might be changes to the airflow_db resource itself that can resolve that. The classic analytics airflow mono-dag with 100s of tasks will be agonizing to run with the current performance of multi-process

PedramNavid:
Closing stale PR.

PR #12136, opened 2023-02-06
Revert "Test with a 64 bit after_cursor"

jmsanders:
Reverts dagster-io/dagster#12093

To troubleshoot if this is the cause of our recent Buildkite build failures


jmsanders:
There seems to be some belief that #11974 will fix the broken builds and that this is unrelated so I'm going to close this for now.

PR #12148, opened 2023-02-07
Sensor testing UI

salazarm:
### Summary & Motivation

This PR implements the designs from this figma: https://www.figma.com/file/DlSTpPlqaBcyUmxpQlCplM/Testing-Sensors-and-Schedules?node-id=142%3A1098 and this notion: https://www.notion.so/dagster/Testing-schedules-sensors-9f480eaf538048f9af6659832019dd05

I manually went through all of the states and also added jest tests.


js_modules/dagit/packages/core/src/sensors/SensorDetails.tsx
```diff
@@ -110,19 +114,40 @@ export const SensorDetails: React.FC<{
...
+                disabled={running}
+              >
+                Test Sensor
+              </Button>
+            ) : null}
           </Box>
         }
       />
+      <SensorDryRun
+        key={showTestTickDialog ? '1' : '0'} // This is to reset the dialog state when its opened/closed
```
hellendag, line 141: Does this affect the enter/exit Dialog animation?

js_modules/dagit/packages/core/src/ticks/SensorDryRun.tsx
```diff
@@ -0,0 +1,361 @@
...
+      style={{width: '70vw', display: 'flex'}}
+      title={
+        <Box flex={{direction: 'row', gap: 8, alignItems: 'center'}}>
+          <Icon name="sensors" />
+          <span>{name}</span>
+        </Box>
+      }
+    >
+      <DialogBody>
+        <div style={{minHeight: '300px'}}>{content}</div>
```
hellendag, line 310: You might consider moving the `content` part into its own component, that way you don't have to perform all of the state setup above unless the Dialog is actually opened. (The contents shouldn't render until the Dialog is open.)

js_modules/dagit/packages/core/src/ticks/SensorDryRun.tsx
```diff
@@ -0,0 +1,361 @@
...
+  return (
+    <Dialog
+      isOpen={isOpen}
+      onClose={onClose}
+      style={{width: '70vw', display: 'flex'}}
+      title={
+        <Box flex={{direction: 'row', gap: 8, alignItems: 'center'}}>
+          <Icon name="sensors" />
+          <span>{name}</span>
+        </Box>
```
hellendag, line 306: `Dialog` has an `icon` prop that you should be able to use here

js_modules/dagit/packages/core/src/ticks/SensorDryRun.tsx
```diff
@@ -0,0 +1,361 @@
...
+    sensorExecutionData,
+    error,
+    submitting,
+    currentCursor,
+    cursorState,
+    onPersistCursorValue,
+    name,
+    jobName,
+    repoAddress,
+    cursor,
```
hellendag, line 294: Is it worth memoizing here, given the number of deps?

js_modules/dagit/packages/core/src/ticks/SensorDryRun.tsx
```diff
@@ -0,0 +1,361 @@
...
+    () => ({
+      sensorName: name,
+      repositoryLocationName: repoAddress.location,
+      repositoryName: repoAddress.name,
+    }),
+    [repoAddress, name],
+  );
+
+  const submitTest = React.useCallback(async () => {
+    debugger;
```
hellendag, line 73: rm

js_modules/dagit/packages/core/src/ticks/SensorDryRun.tsx
```diff
@@ -0,0 +1,361 @@
...
+      } else {
+        setError(data);
+      }
+    } else {
+      assertUnreachable('sensorDryRun Mutation returned no data??' as never);
+    }
+    setSubmitting(false);
+  }, [sensorDryRun, sensorSelector, cursor, name]);
+
+  console.log(repoAddress);
```
hellendag, line 103: rm?

js_modules/dagit/packages/core/src/ticks/SensorDryRun.tsx
```diff
@@ -0,0 +1,361 @@
...
+          Close
+        </Button>
+      );
+    }
+    if (submitting) {
+      return (
+        <Box flex={{direction: 'row', gap: 8}}>
+          <Button onClick={onClose}>Cancel</Button>
+        </Box>
+      );
```
hellendag, line 118: Is it possible to cancel while it's submitting? Does the "Submit" button get hidden during the "submitting" phase?

js_modules/dagit/packages/core/src/ticks/SensorDryRun.tsx
```diff
@@ -0,0 +1,361 @@
...
+      return (
+        <Box flex={{direction: 'row', gap: 8}}>
+          <Button onClick={onClose}>Cancel</Button>
+          <Button onClick={submitTest} intent="primary" data-testid={testId('evaluate')}>
+            Evaluate
+          </Button>
+        </Box>
+      );
+    }
+  }, [sensorExecutionData, error, submitting, onClose, submitTest]);
```
hellendag, line 129: Similar question to below, is it worth memoizing these, given the number of deps?

js_modules/dagit/packages/core/src/ticks/SensorDryRun.tsx
```diff
@@ -0,0 +1,361 @@
...
+                {error ||
+                (currentCursor ?? '') ===
+                  (sensorExecutionData?.evaluationResult?.cursor ?? 'testing1234') ? null : (
+                  <Box flex={{direction: 'row', gap: 8, alignItems: 'center'}}>
+                    <Button
+                      disabled={['Persisting', 'Persisted'].includes(cursorState)}
+                      loading={cursorState === 'Persisting'}
+                      onClick={onPersistCursorValue}
+                    >
+                      <span data-testid={testId('persist-cursor')}>
```
hellendag, line 226: Why test ID on the span instead of the button itself? Just curious

js_modules/dagit/packages/core/src/ticks/SensorDryRun.tsx
```diff
@@ -0,0 +1,361 @@
...
+              </div>
+            </Grid>
+            {error ? (
+              <div>
+                <PythonErrorInfo error={error} />
+              </div>
+            ) : null}
+            {didSkip ? (
+              <div>
+                <Subheading>Skip Reason</Subheading>
```
hellendag, line 248: `Skip reason`

js_modules/dagit/packages/core/src/ticks/SensorDryRun.tsx
```diff
@@ -0,0 +1,361 @@
...
+          <Box>
+            <Grid>
+              <div>
+                <Subheading>Result</Subheading>
+                <Box flex={{grow: 1, alignItems: 'center'}}>
+                  <div>
+                    {error ? (
+                      <Tag intent="danger">Failed</Tag>
+                    ) : numRunRequests ? (
+                      <Tag intent="success">{numRunRequests} Run requests</Tag>
```
hellendag, line 197: `run requests`

js_modules/dagit/packages/core/src/ticks/tests/DryRunRequestTable.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+import {SensorDryRunMutationRunRequests} from './SensorDryRun.mocks';
+
+jest.mock('../../workspace/WorkspaceContext', () => ({useRepository: () => mockRepository}));
+
+// Not sure how to better handle this. Since we typed our mock with MockedResponse typescript thinks
+// these fields are nullable but they're not.
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+const runRequests = SensorDryRunMutationRunRequests!.result!.data.sensorDryRun.evaluationResult
+  .runRequests as RunRequest[];
```
hellendag, line 22: Can you define the `runRequests` values separately (typed as `RunRequest[]`) and then make your mock use that? Then you can just reference them directly here and they should have the correct types since they're independent of the full `MockedResponse`?

```
const runRequests: RunRequest[] = [
  // etc.
];

// later
...
evaluationResult: runRequests,
...
```

js_modules/dagit/packages/core/src/ticks/tests/SensorDryRun.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+  SensorDryRunMutationRunRequests,
+  PersistCursorValueMock,
+} from './SensorDryRun.mocks';
+
+// This component is unit tested separately so mocking it out
+jest.mock('../DryRunRequestTable', () => {
+  return {
+    RunRequestTable: () => <div />,
+  };
+});
```
hellendag, line 21: 👍

hellendag (changes requested):
I think you generated a `graphql/graphql.ts` that we don't actually want anymore. 🤔

js_modules/dagit/packages/core/src/ticks/SensorDryRun.tsx
salazarm: the button doesn't propagate the data attribute :(

js_modules/dagit/packages/core/src/ticks/tests/DryRunRequestTable.test.tsx
salazarm: yeah, thats what I ended up doing in schedules PR, I'll copy that here too

js_modules/dagit/packages/core/src/ticks/SensorDryRun.tsx
```diff
@@ -0,0 +1,361 @@
...
+
+import {RunRequestTable} from './DryRunRequestTable';
+import {SensorDryRunMutation, SensorDryRunMutationVariables} from './types/SensorDryRun.types';
+
+type DryRunInstigationTick = Extract<
+  SensorDryRunMutation['sensorDryRun'],
+  {__typename: 'DryRunInstigationTick'}
+>;
+
+export const SensorDryRun: React.FC<{
```
bengotow, line 42: It'd be nice to call this SensorDryRunDialog or otherwise indicate that it's primary rendered thing is a DIalog :-)

js_modules/dagit/packages/core/src/ticks/SensorDryRun.tsx
```diff
@@ -0,0 +1,361 @@
...
+                <pre>
+                  {sensorExecutionData?.evaluationResult?.cursor?.length
+                    ? sensorExecutionData?.evaluationResult.cursor
+                    : error
+                    ? 'Error'
+                    : 'None'}
+                </pre>
+                {error ||
+                (currentCursor ?? '') ===
+                  (sensorExecutionData?.evaluationResult?.cursor ?? 'testing1234') ? null : (
```
bengotow, line 219: I think this is whacky enough that it'd be nice to have a tiny helper method that gives this check a name. eg `isSameCursor(currnetCursor, sensorExecutionData)`?

Also maybe this `testing1234` value is actually for testing and this should be ??'ing both values with `''`?

js_modules/dagit/packages/core/src/ticks/SensorDryRun.tsx
bengotow: I mentioned this on the other PR but I'd agree this isn't worth it / is probably invalidating every render.

It seems like there are some useMemo's in here that are primarily functions for control flow (eg: early returns) and not because they're actually something that needs to be memoized. That might be fine but I think we could also break this component up into a few smaller components (eg: SensorDryRunDialogContent) to achieve the same thing.

bengotow (approved):

js_modules/dagit/packages/core/src/ticks/SensorDryRun.tsx
salazarm: Memoizing still has some value in case the parent component is needlessly re-rendering, but agreed in this case it doesn't look as good.

js_modules/dagit/packages/core/src/ticks/SensorDryRun.tsx
salazarm: yeah it was just for testing good catch

salazarm:
> It seems like there are some useMemo's in here that are primarily functions for control flow (eg: early returns) and not because they're actually something that needs to be memoized. That might be fine but I think we could also break this component up into a few smaller components (eg: SensorDryRunDialogContent) to achieve the same thing.

Yeah, the tricky thing is theres only 2 memos, content and buttons, and they both depend on the same state so it might not be worth breaking them up. In the schedules PR I tried breaking up the component by moving the mutation and the results of the mutation to a different component (with the useEffect to immediately fire off the mutation)

salazarm:
The failing alert test is unrelated and I'm investigating separately,
The pyright failure is also unrelated:

```


[2023-02-08T22:45:12Z] /workdir/python_modules/dagster/dagster/_core/execution/plan/inputs.py:
--
  | [2023-02-08T22:45:12Z]   471:26: Argument of type "str \| None" cannot be assigned to parameter "__key" of type "str" in function "__getitem__"
  | [2023-02-08T22:45:12Z]   Type "str \| None" cannot be assigned to type "str"
  | [2023-02-08T22:45:12Z]     Type "None" cannot be assigned to type "str" (reportGeneralTypeIssues)
  | [2023-02-08T22:45:12Z]
  | [2023-02-08T22:45:12Z]   472:48: Argument of type "str \| None" cannot be assigned to parameter "io_manager_key" of type "str" in function "build_resources_for_manager"
  | [2023-02-08T22:45:12Z]   Type "str \| None" cannot be assigned to type "str"
  | [2023-02-08T22:45:12Z]     Type "None" cannot be assigned to type "str" (reportGeneralTypeIssues)

```

PR #12160, opened 2023-02-07
Schedules Testing UI

salazarm:
### Summary & Motivation

Figma: https://www.figma.com/file/DlSTpPlqaBcyUmxpQlCplM/Testing-Sensors-and-Schedules?node-id=142%3A1379&t=Wvy1wm0hMk2Y8Gs6-0

Notion: https://www.notion.so/dagster/Testing-schedules-sensors-9f480eaf538048f9af6659832019dd05

Jest tests were added,
Manual tests were done for each of the states (Error, skipped, run requests returned)


js_modules/dagit/packages/core/src/schedules/ScheduleDetails.tsx
```diff
@@ -101,7 +105,29 @@ export const ScheduleDetails: React.FC<{
...
         }
-        right={<QueryRefreshCountdown refreshState={refreshState} />}
+        right={
+          <Box flex={{direction: 'row', alignItems: 'center', gap: 8}}>
+            <QueryRefreshCountdown refreshState={refreshState} />
+            <Button
+              onClick={() => {
+                setShowTestTickDialog(true);
+              }}
+              // disabled={running}
```
bengotow, line 115: Just flagging that this is commented out

js_modules/dagit/packages/core/src/ticks/EvaluateSchedule.tsx
```diff
@@ -0,0 +1,400 @@
...
+  const [_selectedTimestamp, setSelectedTimestamp] = React.useState<{ts: number; label: string}>();
+  const {data} = useQuery<GetScheduleQuery, GetScheduleQueryVariables>(GET_SCHEDULE_QUERY, {
+    variables: {
+      scheduleSelector: {
+        repositoryLocationName: repoAddress.location,
+        repositoryName: repoAddress.name,
+        scheduleName: name,
+      },
+    },
+  });
```
bengotow, line 58: Hmm this is a query within a dialog, but it's not lazy or using `skip: !isOpen` so I believe this is being called for every schedule row even if you're not opening the dialogs? 

To make this simpler (and to avoid the key={1 | 0} thing above), it might be nice to just render this content when the modal is open and sacrifice the brief animation. This is also my pitch for the imperative style `onClick={() => showCustomAlert(<EvaluateScheduleModal />)}` but I don't think we do that except for alerts.

js_modules/dagit/packages/core/src/ticks/EvaluateSchedule.tsx
```diff
@@ -0,0 +1,400 @@
...
+    _selectedTimestamp,
+    containerProps,
+    data,
+    isTickSelectionOpen,
+    jobName,
+    name,
+    repoAddress,
+    shouldEvaluate,
+    userTimezone,
+    viewport.width,
```
bengotow, line 144: I'm not sure this is worth memoizing -- with all these deps, it seems like this is probably being rebuilt on every render. What state isn't in this list? 😅

js_modules/dagit/packages/core/src/ticks/EvaluateSchedule.tsx
```diff
@@ -0,0 +1,400 @@
...
+        locale,
+        timezone: userTimezone,
+        timeFormat: {
+          showTimezone: true,
+        },
+      }),
+    }));
+    selectedTimestampRef.current = _selectedTimestamp || timestamps[0];
+    return (
+      <SelectWrapper>
```
bengotow, line 101: This seems like a standard component that should be part of our library - is this written from scratch to get the `data-testid` props in there? Seems like it'd be best to move it to it's own component so the `isTickSelectionOpen` state doesn't have to pollute the top-level modal.

js_modules/dagit/packages/core/src/ticks/EvaluateSchedule.tsx
```diff
@@ -0,0 +1,400 @@
...
+    }, [name, repoAddress, timestamp]),
+  );
+  const [result, setResult] = React.useState<Awaited<
+    ReturnType<typeof scheduleDryRunMutation>
+  > | null>(null);
+  React.useEffect(() => {
+    scheduleDryRunMutation().then((result) => {
+      setResult(() => result);
+    });
+  }, [scheduleDryRunMutation]);
```
bengotow, line 245: It's a little weird that something called `EvaluateScheduleContent` immediately invokes a mutation as soon as it's mounted. Maybe this should have a name that indicates it's the end of the workflow like `DryRunResultsContent` rather than just the name of the modal.

js_modules/dagit/packages/core/src/ticks/EvaluateSchedule.tsx
```diff
@@ -0,0 +1,400 @@
...
+      <Box padding={32} flex={{justifyContent: 'center', alignItems: 'center'}}>
+        <Spinner purpose="page" />
+      </Box>
+    );
+  }
+
+  const evaluationResult =
+    result?.data?.scheduleDryRun.__typename === 'DryRunInstigationTick'
+      ? result?.data?.scheduleDryRun.evaluationResult
+      : null;
```
bengotow, line 258: This might not be in-scope, but this seems like a place where we'd have `evaluationResult` or a `PythonError`. Any chance there's more info than "unknown error" we could display here?

js_modules/dagit/packages/core/src/ticks/EvaluateSchedule.tsx
```diff
@@ -0,0 +1,400 @@
...
+      return {
+        variables: {
+          selectorData: {
+            ...repositorySelector,
+            scheduleName: name,
+          },
+          timestamp,
+        },
+      };
+    }, [name, repoAddress, timestamp]),
```
bengotow, line 236: This is fine but it'd be kinda nice to pull this up to a top level `const scheduleDryRunVariables = React.useMemo(()=>` above the `useMutation` line to keep things standard looking.

js_modules/dagit/packages/core/src/ticks/EvaluateSchedule.tsx
```diff
@@ -0,0 +1,400 @@
...
+            <Box flex={{grow: 1, alignItems: 'center'}}>
+              <Mono>
+                {timestampToString({
+                  timestamp: {unix: timestamp},
+                  locale,
+                  timezone: userTimezone,
+                  timeFormat: {
+                    showTimezone: true,
+                  },
+                })}
```
bengotow, line 330: Could this be `<TimestampDisplay timestamp={timestamp} />`?  It seems like all the pieces of that component are being re-imported and implemented in this file and I _think_ they're the same?

js_modules/dagit/packages/core/src/ticks/tests/EvaluateSchedule.mocks.tsx
```diff
@@ -0,0 +1,170 @@
...
+      scheduleOrError: {
+        __typename: 'Schedule',
+        name: 'configurable_job_schedule',
+        potentialTickTimestamps: [1, 2, 4, 5, 6, 7, 8, 9, 10],
+      },
+    },
+  },
+};
+
+export const scheduleDryWithWithRunRequest = {
```
bengotow, line 29: I think if you put the `ScheduleDryRunMutation` type on this line you can avoid all the inline const annotations?

bengotow (approved):
This looks good! I added a bunch of inline comments but I think the only critical thing is that the dialog query appears to be running for each row without the dialogs being opened.

js_modules/dagit/packages/core/src/ticks/EvaluateSchedule.tsx
salazarm: Yeah, I mostly just didn't want the parent component to call the mutation and save the data and then pass it down to this component because the parent component already has lots of state. I was mostly just testing around to see, I can switch it back but really the difference between a query/mutation is... kind of arbitrary tbh. In this case we're not persisting any information, it's effectively a query, where as other queries do logging and other kinds of stuff which do persist data.

js_modules/dagit/packages/core/src/ticks/EvaluateSchedule.tsx
salazarm: the error is inside evaluationResult in this API. Not sure exactly why I think the underlying object is used in other APIs where there can be errors too

js_modules/dagit/packages/core/src/ticks/EvaluateSchedule.tsx
salazarm: I looked at TimestampDisplay and it has those extra icons to the left and right that we don't need. I could update it to make it possible to turn those off. It also has some other unnecessary stuff too I think though

js_modules/dagit/packages/core/src/ticks/tests/EvaluateSchedule.mocks.tsx
salazarm: If I do that then it's going to think the fields in the object are nullable since they are nullable in the `ScheduleDryRunMutation` type and I want to use this object in my test to iterate over the run requests

js_modules/dagit/packages/core/src/ticks/EvaluateSchedule.tsx
salazarm: This is just to add some custom CSS to override blueprint's CSS to make the Select button take the full width.

salazarm:
Build failure is unrelated and will be fixed by https://github.com/dagster-io/internal/pull/4822

PR #12163, opened 2023-02-07
[docs][cloud] Clarify that org. admin settings are needed to modify deployment settings

benpankow:
## Summary

This is currently an org. admin only setting, per https://dagster.slack.com/archives/C01U954MEER/p1675727780720119

### DIFF
--- a/docs/content/dagster-cloud/account/managing-users.mdx
+++ b/docs/content/dagster-cloud/account/managing-users.mdx
@@ -168,8 +168,8 @@ Dagster Cloud currently includes support for four levels of role-based access co
         settings
       </td>
       <td className="bg-red-50">N</td>
-      <td className="bg-green-50">Y</td>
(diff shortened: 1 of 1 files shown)


benpankow:
* master
  * **PR #12163** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12163" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12163?utm_source=stack-comment).

gibsondan (changes requested):
lets do the opposite and take out the incorrect claim about org admin that the user called out in the docs on deployment settings (and fix dagit separately to reflect that reality - that apart i can do)

erinkcochran87:
Can you ping me when this is updated again? Between this and Slack I'm not sure what the end state should actually be - there may be other docs that require updates depending on that.

PR #12129, opened 2023-02-06
[db io managers] support static partitions

jamiedemaria:
### Summary & Motivation
Adds static partition support for DBIOManager base class. Updates duckdb and snowflake io managers accordingly

docs plan - i'm going to add multi partition support, then do a pass at updating api docs and the guides to cover partitions

### How I Tested These Changes
bk - new unit tests

PR #12165, opened 2023-02-07
[db io manager] Support MultiPartitions

jamiedemaria:
### Summary & Motivation

Adds support for multi partitions for DB io managers (snowflake and duckdb). Additionally updates the snowflake and duckdb io managers to also handle multi partitions

Docs plan - will have a future PR updating API docs and guides

bk + new unit tests

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/db_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/db_io_manager.py
@@ -28,7 +28,7 @@ from dagster._core.storage.io_manager import IOManager
 T = TypeVar("T")
 
 
-class TablePartition(NamedTuple):
+class PartitionDimension(NamedTuple):
     partition_expr: str
     partition: Union[TimeWindow, str]
 
@@ -38,7 +38,7 @@ class TableSlice(NamedTuple):
     schema: str
     database: Optional[str] = None
     columns: Optional[Sequence[str]] = None
-    partition: Optional[Sequence[TablePartition]] = None
+    partition: Optional[Sequence[PartitionDimension]] = None
 
 
 class DbTypeHandler(ABC, Generic[T]):
@@ -150,7 +150,7 @@ class DbIOManager(IOManager):
         schema: str
         table: str
         partition_value: Optional[Union[TimeWindow, str]] = None
-        partitions: List[TablePartition] = []
+        partitions: List[PartitionDimension] = []
         if context.has_asset_key:
             asset_key_path = context.asset_key.path
             table = asset_key_path[-1]
@@ -195,7 +195,7 @@ class DbIOManager(IOManager):
                                 f" {part.name} partition."
                             )
                         partitions.append(
-                            TablePartition(
+                            PartitionDimension(
                                 partition_expr=partition_expr_str, partition=partition_value
                             )
                         )
@@ -207,7 +207,9 @@ class DbIOManager(IOManager):
                     )
 
                     partitions.append(
-                        TablePartition(partition_expr=partition_expr_str, partition=partition_value)
+                        PartitionDimension(
+                            partition_expr=partition_expr_str, partition=partition_value
+                        )
                     )
         else:
             table = output_context.name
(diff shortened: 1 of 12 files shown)


jamiedemaria:
* master
  * **PR #12129** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12129" target="_blank">Graphite</a> 
    * **PR #12165** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12165" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12165?utm_source=stack-comment).

python_modules/dagster/dagster/_core/storage/db_io_manager.py
```diff
@@ -139,6 +150,7 @@ def _get_table_slice(
         schema: str
         table: str
         partition_value: Optional[Union[TimeWindow, str]] = None
+        partitions: List[TablePartition] = []
```
sryza, line 153: It feels confusing to me to represent this as multiple `TablePartition` objects - logically, this slice is operating on a single partition, not multiple partitions.

It seems like we should either:
- Allow `TablePartition` to contain a multi-partition.
- Rename `TablePartition` to something that communicates that it's a particular partition dimension.

python_modules/libraries/dagster-duckdb/dagster_duckdb/io_manager.py
```diff
@@ -143,13 +149,20 @@ def _get_cleanup_statement(table_slice: TableSlice) -> str:
...
-    if table_slice.partition:
-        partition_where = (
-            _static_where_clause(table_slice.partition)
-            if isinstance(table_slice.partition.partition, str)
-            else _time_window_where_clause(table_slice.partition)
-        )
-        return f"DELETE FROM {table_slice.schema}.{table_slice.table}\n" + partition_where
+    if table_slice.partition and len(table_slice.partition) > 0:
+        query = f"DELETE FROM {table_slice.schema}.{table_slice.table} WHERE\n"
+        for i in range(len(table_slice.partition)):
```
jamiedemaria, line 154: building up the query as a bunch of string concatenations feels a bit fragile to me - are there better options? is that something sqlalchemy could be used for?

python_modules/dagster/dagster/_core/storage/db_io_manager.py
jamiedemaria: ah interesting. I prefer renaming to `PartitionDimension` vs nesting another object within TablePartition. will do

python_modules/dagster/dagster/_core/storage/db_io_manager.py
```diff
@@ -131,6 +133,15 @@ def load_input(self, context: InputContext) -> object:
             context, self._get_table_slice(context, cast(OutputContext, context.upstream_output))
         )
 
+    def _get_partition_value(
+        self, partition_def: PartitionsDefinition, partition_key: str
+    ) -> Union[TimeWindow, str]:
+        try:
```
clairelin135, line 139: Can we use an `isinstance(partitions_def, TimewindowPartitionsDefinition)` call instead of a try catch block?

python_modules/dagster/dagster/_core/storage/db_io_manager.py
clairelin135: Could we rename to `TablePartitionDimension`? `PartitionDimension` seems to conflict a bit with `PartitionDimensionDefinition` which is a known entity

clairelin135 (approved):
The partitioning parts of this look good to me pending a couple comments below

python_modules/dagster/dagster/_core/storage/db_io_manager.py
```diff
@@ -36,7 +44,7 @@ class TableSlice(NamedTuple):
     schema: str
     database: Optional[str] = None
     columns: Optional[Sequence[str]] = None
-    partition: Optional[TablePartition] = None
+    partition: Optional[Sequence[TablePartitionDimension]] = None
```
sryza, line 47: I think calling this something like `partition_dimensions` would make it more clear what it contains.

sryza (approved):
Two small remaining comments. Otherwise, looks great.

PR #12174, opened 2023-02-08
Keep dots in paths in UPathIOManager

### LINKED ISSUE #12173: UPathIOManager truncates paths when partition key has a "." in it
### Dagster version

dagster, version 1.1.17

### What's the issue?

I'm using a custom PartitionsDefinition which uses static decimal partitioning keys, which happen to make sense for my data. It provides partition keys which include a "." character. For example, `0.94-to-0.95` is a partition key I use.

I also use a UPathIOManager - specifically, one that works with Google Cloud Storage.

The IO manager is storing `asset` outputs in paths like `"gs://redacted/redacted/redacted/0.94-to-0"`

I would expect the UPathIOManager's generated paths to be unique per partition key, but since they truncate the partition key if it has a "." character, this is not the case. Data gets overwritten.

### How to reproduce?

Make a static partition definition using keys ["0.0", "1.0"]. Make assets using that partition definition, and use a UPathIOManager. Observe the paths where the assets are stored.

### Deployment type

Dagster Helm chart

### Deployment details

_No response_

### Additional information

_No response_

spenczar:
### Summary & Motivation

Fixes #12173 in the simplest way possible. But this doesn't really fix the underlying issue. If an extension is set, then partition keys will still be clobbered and overrun.

I think the use of `with_suffix` maybe needs to be removed entirely. It seems to me that the intended behavior is to tack on `self.extension`, not to replace any existing extension, which is what `with_suffix` does. But I don't know enough about the intended use of the `extension` class variable on `UPathIOManager`, so I held off.

I wrote a test. I ran `python -m pytest python_modules/dagster/dagster_tests/storage_tests/test_upath_io_manager.py`. The tests fails before my change. It passes after my change.

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/upath_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/upath_io_manager.py
@@ -62,9 +62,7 @@ class UPathIOManager(MemoizableIOManager):
         # Can't just call path.with_suffix(self.extension) because if
         # self.extension is "" then this trims off any extension that
         # was in the path previously.
-        if self.extension == "":
-            return path
-        return path.with_suffix(self.extension)
+        return path.parent / (path.name + self.extension)
 
     def _get_path_without_extension(self, context: Union[InputContext, OutputContext]) -> UPath:
         if context.has_asset_key:
--- a/python_modules/dagster/dagster/_core/storage/upath_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/upath_io_manager.py
@@ -58,6 +58,14 @@ class UPathIOManager(MemoizableIOManager):
     def has_output(self, context: OutputContext) -> bool:
         return self._get_path(context).exists()
 
+    def _with_extension(self, path: UPath) -> UPath:
+        # Can't just call path.with_suffix(self.extension) because if
+        # self.extension is "" then this trims off any extension that
+        # was in the path previously.
+        if self.extension == "":
+            return path
+        return path.with_suffix(self.extension)
+
     def _get_path_without_extension(self, context: Union[InputContext, OutputContext]) -> UPath:
         if context.has_asset_key:
             # we are dealing with an asset
@@ -75,7 +83,8 @@ class UPathIOManager(MemoizableIOManager):
         Returns the I/O path for a given context.
         Should not be used with partitions (use `_get_paths_for_partitions` instead).
         """
-        return self._get_path_without_extension(context).with_suffix(self.extension)
+        path = self._get_path_without_extension(context)
+        return self._with_extension(path)
 
     def _get_paths_for_partitions(
         self, context: Union[InputContext, OutputContext]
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/storage/upath_io_manager.py
```diff
@@ -58,6 +58,14 @@ def get_metadata(
     def has_output(self, context: OutputContext) -> bool:
         return self._get_path(context).exists()
 
+    def _with_extension(self, path: UPath) -> UPath:
+        # Can't just call path.with_suffix(self.extension) because if
```
sryza, line 62: In the subset of cases where `extension != ""`, I think that, even after this change, the problem you raised will still exist.

StackOverflow has a suggestion on how to fix this: https://stackoverflow.com/a/53092299/6258015. Thoughts?

sryza (commented):
Thanks for the fix @spenczar ! I left one question.

python_modules/dagster/dagster/_core/storage/upath_io_manager.py
spenczar: Agreed. I am fine with any of the methods in that SO link. I will pick one and push.

spenczar: Actually a further question @sryza. Suppose the path ends with "/path/to/file.pickle" and extension is "pickle". Should the result be "/path/to/file.pickle.pickle"?

sryza: > Actually a further question @sryza. Suppose the path ends with "/path/to/file.pickle" and extension is "pickle". Should the result be "/path/to/file.pickle.pickle"?

I think yes, just for consistency.

spenczar:
@sryza This is ready for a second look.

sryza:
@spenczar this looks very close. A couple things:

1. Mind running `make black` to use the autoformatter?
2. pyright is complaining about a few typing issues:
```
[2023-02-09T00:39:00Z] /workdir/python_modules/dagster/dagster_tests/storage_tests/test_upath_io_manager.py:
[2023-02-09T00:39:00Z]   221:72: Function with declared type of "List[Unknown]" must return value on all code paths
[2023-02-09T00:39:00Z]   Type "None" cannot be assigned to type "List[Unknown]" (reportGeneralTypeIssues)
[2023-02-09T00:39:00Z]
[2023-02-09T00:39:00Z]   227:80: "storage_directory" is not a known member of "None" (reportOptionalMemberAccess)
[2023-02-09T00:39:00Z]
[2023-02-09T00:39:00Z]   257:72: Function with declared type of "List[Unknown]" must return value on all code paths
[2023-02-09T00:39:00Z]   Type "None" cannot be assigned to type "List[Unknown]" (reportGeneralTypeIssues)
[2023-02-09T00:39:00Z]
[2023-02-09T00:39:00Z]   263:80: "storage_directory" is not a known member of "None" (reportOptionalMemberAccess)
[2023-02-09T00:39:00Z]
[2023-02-09T00:39:00Z]   238:62: Unnecessary "# type: ignore" comment
[2023-02-09T00:39:00Z]
[2023-02-09T00:39:00Z]   274:62: Unnecessary "# type: ignore" comment
[2023-02-09T00:39:00Z]
```

spenczar:
Sure, I think I've resolved all the black and pyright issues.

As an aside, https://docs.dagster.io/community/contributing#developing-dagster should be updated to mention `make pyright` and `make ruff` instead of `make pylint` and `make isort`.

sryza (approved):
This looks great.

PR #12177, opened 2023-02-08
[dagit] Storybooks for MetadataEntry rendering

bengotow:
### Summary & Motivation

I think this deserves a storybook preview because a few of the values are allowed to be null, which is confusing, and because the TableSchema type is difficult to repro locally without mocking.

Also: The "TableMetadataEntry" type doesn't render anything at all - I will verify separately that this is ok because it could be an oversight: https://github.com/dagster-io/dagster/blob/master/js_modules/dagit/packages/core/src/metadata/MetadataEntry.tsx#L172

I structured the test data in a bit of a strange way so that I could assert exhaustiveness - if you add a new MetadataEntry union member you have to add an example :-)  

This seems like another good candidate for some sort of visual regression testing because there's not much to assert against.

image

image

### DIFF
--- a/js_modules/dagit/packages/core/src/metadata/MetadataEntry.stories.tsx
+++ b/js_modules/dagit/packages/core/src/metadata/MetadataEntry.stories.tsx
@@ -1,0 +1,267 @@
+import {Box, Checkbox} from '@dagster-io/ui';
+import React from 'react';
+
+import {assertUnreachable} from '../app/Util';
+import {
+  BoolMetadataEntry,
+  IntMetadataEntry,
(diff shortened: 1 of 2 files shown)


hellendag (approved):
What counts as a "small value"?

bengotow:
Hmm @hellendag  the "small value" concept is sort of baked into this MetadataEntries component - right now the cutoff is 1000 characters of JSON, but it's somewhat arbitrary. I think the idea is that small things are displayed inline, and anything larger than that requires that you click `[Show Markdown]` or `[Show JSON]` etc

PR #12178, opened 2023-02-08
[dagit] Storybooks for asset partition and event details + dayJS fix

bengotow:
### Summary & Motivation

This adds storybooks for the asset partition details and asset event details UI in Dagit. These components mostly display static data, but there are sections that are only populated in certain scenarios (eg: when a materialization is followed by an observation event that contains additional metadata entries). It can be difficult to see the entire thing without a nice asset graph.

In the process of building these I discovered that the dayjs constructor only handles timestamps properly if they're passed as numbers, and fixed all the callsites that were passing them as strings.

I think that these should also get some jest tests that poke the rendering and verify text appears as expected using the same mock data as the storybooks, but it might be worth merging this now for the dayjs fix.

image

image

image

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/AllIndividualEventsLink.tsx
+++ b/js_modules/dagit/packages/core/src/assets/AllIndividualEventsLink.tsx
@@ -134,8 +134,8 @@ const MetadataEntriesRow: React.FC<{
                             <Link to={`/runs/${obs.runId}?timestamp=${obs.timestamp}`}>
                               <Mono>{titleForRun({runId: obs.runId})}</Mono>
                             </Link>
-                            {` (${dayjs(obs.timestamp).from(
-                              timestamp,
+                            {` (${dayjs(Number(obs.timestamp)).from(
+                              Number(timestamp),
                               true, // withoutSuffix
                             )} later)`}
                           </span>
--- a/js_modules/dagit/packages/core/src/assets/AssetEventDetail.stories.tsx
+++ b/js_modules/dagit/packages/core/src/assets/AssetEventDetail.stories.tsx
@@ -1,0 +1,61 @@
+import {MockedProvider} from '@apollo/client/testing';
+import {Box} from '@dagster-io/ui';
+import React from 'react';
+
+import {createAppCache} from '../app/AppCache';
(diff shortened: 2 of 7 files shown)


js_modules/dagit/packages/core/src/assets/AssetEventDetail.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+  MaterializationUpstreamDataEmptyMock,
+  MaterializationUpstreamDataFullMock,
+} from './AssetEventDetail.mocks';
+
+// eslint-disable-next-line import/no-default-export
+export default {component: AssetEventDetail};
+
+export const EmptyState = () => {
+  return (
+    <Box style={{width: '950px', display: 'flex', flexDirection: 'column'}}>
```
hellendag, line 23: Are the flex properties necessary on the wrapper in all cases? Should that style be pushed down into `AssetEventDetail` itself?

js_modules/dagit/packages/core/src/assets/AllIndividualEventsLink.tsx
```diff
@@ -134,8 +134,8 @@ const MetadataEntriesRow: React.FC<{
                             <Link to={`/runs/${obs.runId}?timestamp=${obs.timestamp}`}>
                               <Mono>{titleForRun({runId: obs.runId})}</Mono>
                             </Link>
-                            {` (${dayjs(obs.timestamp).from(
-                              timestamp,
+                            {` (${dayjs(Number(obs.timestamp)).from(
+                              Number(timestamp),
```
hellendag, line 138: 😭

hellendag (approved):

js_modules/dagit/packages/core/src/assets/AssetEventDetail.stories.tsx
bengotow: Ahh let me see if I can remove the display: flex on here, good call. I'd just added the width to better simulate these appearing in a list + detail view.

bengotow:
Failures in oss-internal-compatibility are related to AlertPolicyEditor dagit tests in cloud, going to go ahead and merge this.

PR #12179, opened 2023-02-08
[dagster-airflow] kwargs for flowing through configuration to dagster objects

Ramshackle-Jamathon:
### Summary & Motivation

The dagster-airflow api's currently don't allow users to configure the returns from its api's, resulting in users having to extract and redefine things if they want to modify configuration. This pr adds kwargs for the following api's:

- `make_dagster_definitions_from_airflow_dag_bag`
- `make_dagster_definitions_from_airflow_dags_path`
- `make_dagster_job_from_airflow_dag`
- `make_dagster_schedule_from_airflow_dag`

`load_assets_from_airflow_dag` and `make_schedules_and_jobs_from_airflow_dag_bag` are omitted since flowing kwargs is not as obvious with each. The work for annotating Assets backed by dags is probably more substantial than kwargs anyways...

- [ ] unit tests

### DIFF
--- a/python_modules/libraries/dagster-airflow/dagster_airflow/dagster_factory.py
+++ b/python_modules/libraries/dagster-airflow/dagster_airflow/dagster_factory.py
@@ -21,7 +21,7 @@ from dagster_airflow.utils import (
 def make_dagster_definitions_from_airflow_dag_bag(
     dag_bag: DagBag,
     connections: Optional[List[Connection]] = None,
-    kwargs: Optional[dict] = None,
+    definitions_kwargs: Optional[dict] = None,
 ) -> Definitions:
     """Construct a Dagster definition corresponding to Airflow DAGs in DagBag.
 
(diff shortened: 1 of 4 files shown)


python_modules/libraries/dagster-airflow/dagster_airflow/dagster_factory.py
```diff
@@ -38,16 +39,19 @@ def make_definition_from_dag_bag():
...
         Definitions
     """
+    definitions_kwargs = check.opt_dict_param(definitions_kwargs, "definitions_kwargs")
     schedules, jobs = make_schedules_and_jobs_from_airflow_dag_bag(
         dag_bag,
         connections,
     )
 
     return Definitions(
+        *definitions_kwargs if definitions_kwargs else (),
```
schrockn, line 54: Does this even work?

I've only ever seen `Definitions(arg1=arg1, arg2=arg2, **kwargs)` work. 

This needs unit tests either way.

schrockn (changes requested):
I'd love to be able to avoid having to do this. This is definitely a code smell where we haven't composed things property. Let's see where your refactor ends up post dealing with asset resource clash issue.

PR #12189, opened 2023-02-08
Allow setting logical version inside op

smackesey:
### Summary & Motivation

Allow setting the `logical_version` via `Output` object. This will override the auto-generated logical version.

New unit test

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/context/system.py
+++ b/python_modules/dagster/dagster/_core/execution/context/system.py
@@ -24,10 +24,6 @@ import dagster._check as check
 from dagster._annotations import public
 from dagster._core.definitions.events import AssetKey, AssetLineageInfo
 from dagster._core.definitions.hook_definition import HookDefinition
-from dagster._core.definitions.logical_version import (
-    LogicalVersion,
-    extract_logical_version_from_entry,
-)
 from dagster._core.definitions.mode import ModeDefinition
 from dagster._core.definitions.op_definition import OpDefinition
 from dagster._core.definitions.partition import PartitionsDefinition, PartitionsSubset
@@ -67,6 +63,9 @@ from .output import OutputContext, get_output_context
 if TYPE_CHECKING:
     from dagster._core.definitions.dependency import Node, NodeHandle
     from dagster._core.definitions.job_definition import JobDefinition
+    from dagster._core.definitions.logical_version import (
+        LogicalVersion,
+    )
     from dagster._core.definitions.resource_definition import Resources
     from dagster._core.event_api import EventLogRecord
     from dagster._core.execution.plan.plan import ExecutionPlan
@@ -488,7 +487,7 @@ class StepExecutionContext(PlanExecutionContext, IStepContext):
 
         self._input_asset_records: Dict[AssetKey, Optional["EventLogRecord"]] = {}
         self._is_external_input_asset_records_loaded = False
-        self._generated_logical_versions: Dict[AssetKey, LogicalVersion] = {}
+        self._logical_version_cache: Dict[AssetKey, "LogicalVersion"] = {}
 
     @property
     def step(self) -> ExecutionStep:
@@ -822,8 +821,14 @@ class StepExecutionContext(PlanExecutionContext, IStepContext):
             )
             return asset_info is not None
 
-    def record_logical_version(self, asset_key: AssetKey, logical_version: LogicalVersion) -> None:
-        self._generated_logical_versions[asset_key] = logical_version
+    def set_logical_version(self, asset_key: AssetKey, logical_version: "LogicalVersion") -> None:
+        self._logical_version_cache[asset_key] = logical_version
+
+    def has_logical_version(self, asset_key: AssetKey) -> bool:
+        return asset_key in self._logical_version_cache
+
+    def get_logical_version(self, asset_key: AssetKey) -> "LogicalVersion":
+        return self._logical_version_cache[asset_key]
 
     @property
     def input_asset_records(self) -> Optional[Mapping[AssetKey, Optional["EventLogRecord"]]]:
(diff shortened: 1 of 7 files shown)


smackesey:
* master
  * **PR #12189** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12189" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12189?utm_source=stack-comment).

sryza:
Our preferred way of logging op/asset output metadata is:

```python
@asset
def my_asset():
    return Output(..., metadata=...)
```

Thoughts on adding a version attribute to `Output` instead of adding a setter method?

sryza (commented):
The shape of this looks sensible to me.  A couple thoughts:
- Needs docstring and the experimental marker
- If we're going to change LogicalVersion to DataVersion, I think we should consider doing that first. Thoughts?
- I'm still trying to wrap my head around how this affects staleness calculations. Are you able to include some tests that define the expected behavior?  Random example: what happens if you set `code_version` on the same op? Does anything unexpected happen if the code version changes between two materializations but the data/logical version stays the same?

smackesey:
>If we're going to change LogicalVersion to DataVersion, I think we should consider doing that first. Thoughts?

I'm up for doing that for this week's release.

smackesey:
@sryza 

> Needs docstring and the experimental marker

done

>If we're going to change LogicalVersion to DataVersion, I think we should consider doing that first. Thoughts?

Would prefer to get this in this week, unadvertised, so our selected partners can at least experiment with it. `DataVersion` rename for next week.

>I'm still trying to wrap my head around how this affects staleness calculations. Are you able to include some tests that define the expected behavior? Random example: what happens if you set code_version on the same op? Does anything unexpected happen if the code version changes between two materializations but the data/logical version stays the same?

I have a plan for this but requires a bit more discussion, so not for this week.

sryza:
> Would prefer to get this in this week, unadvertised, so our selected partners can at least experiment with it. DataVersion rename for next week.

My concern about rushing this in before understanding the staleness semantics is that we'll be able to end up with permanently-stored values in our database. If we realize there are issues with the staleness semantics and revert this change, we'll still be stuck needing to interpret those values.

smackesey:
Here is how the staleness semantics work:

If the user is setting their own data versions (currently logical versions), then the current logical version will never equal the projected logical version. This means, with the current system in dagit, the asset will always show as stale if the last materialization used a manual logical version.

The solution to this is to add a flag on the asset, similar to code versioning, that indicates the data versioning strategy (manual or auto). If manual, then the projected logical version will always be treated as unknown, and therefore the stale status will always be unknown. This flag can be added next week, and we'll throw an error if the user tries to manually set the data version for an auto-versioned asset.

>If we realize there are issues with the staleness semantics and revert this change, we'll still be stuck needing to interpret those values.

I don't believe there are any issues with what I outlined above, the questions are just around the right API for setting the version mode.

sryza (approved):
I have some hesitations about this, because it seems like there are important API implications that we still need to work out.

I'm approving it because seems like it might unblock a user from trying out this functionality and giving feedback.  I think we should consider reverting it if it doesn't get traction soon, to avoid an undocumented loose end.

python_modules/dagster/dagster/_core/execution/context/system.py
```diff
@@ -822,8 +821,14 @@ def step_materializes_assets(self) -> bool:
             )
             return asset_info is not None
 
-    def record_logical_version(self, asset_key: AssetKey, logical_version: LogicalVersion) -> None:
-        self._generated_logical_versions[asset_key] = logical_version
+    def set_logical_version(self, asset_key: AssetKey, logical_version: "LogicalVersion") -> None:
```
schrockn, line 824: very strong aversion to making context mutable in this fashion

python_modules/dagster/dagster/_core/execution/plan/execute_step.py
```diff
@@ -474,9 +478,21 @@ def _get_output_asset_materializations(
...
+        logical_version = (
+            compute_logical_version(
+                code_version,
+                {k: meta["logical_version"] for k, meta in input_provenance_data.items()},
+            )
+            if output.logical_version is None
+            else output.logical_version
+        )
+        tags = _build_logical_version_tags(logical_version, code_version, input_provenance_data)
+        if not step_context.has_logical_version(asset_key):
```
schrockn, line 493: this is just asking for trouble. this has the potential to inject a silent failure condition where some code sets the logical version (e.g. the user does this in the user code themselves, which they can totally do) and breaks invariants in the system

python_modules/dagster/dagster_tests/execution_tests/test_logical_versions.py
```diff
@@ -29,19 +29,30 @@
...
 # ########################
 # ##### HELPERS
 # ########################
 
 
+class MaterializationTable:
+    def __init__(self, materializations: Mapping[AssetKey, AssetMaterialization]):
+        self.materializations = materializations
+
+    def __getitem__(self, key: Union[str, AssetKey]) -> AssetMaterialization:
```
schrockn, line 44: do we really need to overload an operator here? Seems unnecessarily clever/not amenable for grepping with limited upside

schrockn (commented):
This has already been submitted but I definitely have concerns about the way this PR is implemented and its implications. We'll have to discuss this. Unfortunately I will not be able to do so in depth synchronously until early next week.

python_modules/dagster/dagster/_core/execution/plan/execute_step.py
smackesey: >this has the potential to inject a silent failure condition where some code sets the logical version (e.g. the user does this in the user code themselves, which they can totally do)

Hmm... it's technically possible for the user to do this, but the `StepExecutionContext` isn't part of the public API-- not sure I understand what separates this case from any other case of internal book-keeping machinery.

python_modules/dagster/dagster/_core/execution/context/system.py
smackesey: I'm open-minded on other approaches but just want to point out that this PR does not introduce this mutability.

python_modules/dagster/dagster_tests/execution_tests/test_logical_versions.py
smackesey: It's just a convenience used in one test file, returned by a few helpers in that file, to avoid cluttering tests with `AssetKey(...)`. Necessary? No. Nicer to use when writing tests? IMO yes, but just a matter of taste. I don't have an aversion to using `__getitem__` for this kind of thing but I'm happy to change it.

PR #12195, opened 2023-02-08
validate time partition keys when adding to a TimeWindowPartitionsSubset

OwenKephart:
### Summary & Motivation

Align with this comment: https://sourcegraph.com/github.com/dagster-io/dagster/-/blob/python_modules/dagster/dagster/_core/storage/partition_status_cache.py?L157&subtree=true

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
@@ -1344,7 +1344,6 @@ class TimeWindowPartitionsSubset(PartitionsSubset):
         return result_windows, num_added_partitions
 
     def with_partition_keys(self, partition_keys: Iterable[str]) -> "TimeWindowPartitionsSubset":
-
         # if we are representing things as a static set of keys, continue doing so
         if self._included_partition_keys is not None:
             new_partitions = {*self._included_partition_keys, *partition_keys}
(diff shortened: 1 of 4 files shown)


sryza (commented):
Two thoughts:
- My instinct would be that `with_partition_keys` should error in this case.
- Do we know the perf impact of this?

OwenKephart:
perf impact should be fairly minimal, as this gets called on all partition keys in other parts of the code, and has never been an issue. agree that this feels a bit weird to allow -- I'd also be down to move. this logic to the spot indicated by the linked comment if you think that'd be better (probably adding an is_valid_partition_key to TimeWindowPartitionsDefinition)

sryza:
I think the spot with the comment is better, at the very least for consistency

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -1337,6 +1344,7 @@ def _add_partitions_to_time_windows(
         return result_windows, num_added_partitions
 
     def with_partition_keys(self, partition_keys: Iterable[str]) -> "TimeWindowPartitionsSubset":
+
```
sryza: Remove this

sryza (approved):
This looks good to me. I think it's worth @clairelin135 taking a look too.

PR #12196, opened 2023-02-08
Make Partition GRPC calls with `instance_ref` args optional

clairelin135:
In order to support older versions of Dagit with Dagster 1.1.18 and above, makes the `instance_ref` argument optional on GRPC endpoints for partitions.

When using Dagster 1.1.18 and an older version of Dagit with dynamic partitions, throws a more informative error telling the user to upgrade their version of Dagit.

### DIFF
--- a/python_modules/dagster/dagster/_grpc/impl.py
+++ b/python_modules/dagster/dagster/_grpc/impl.py
@@ -36,7 +36,7 @@ from dagster._core.host_representation.external_data import (
     ExternalScheduleExecutionErrorData,
     ExternalSensorExecutionErrorData,
 )
-from dagster._core.instance import DagsterInstance, DynamicPartitionsStore
+from dagster._core.instance import DagsterInstance
 from dagster._core.instance.ref import InstanceRef
 from dagster._core.snap.execution_plan_snapshot import (
     ExecutionPlanSnapshotErrorData,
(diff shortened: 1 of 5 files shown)


python_modules/dagster/dagster/_grpc/impl.py
```diff
@@ -53,6 +54,19 @@
 from .types import ExecuteExternalPipelineArgs
 
 
+class ErrorOnMissingInstance(DynamicPartitionsStore):
```
gibsondan, line 57: this kind of exploding class was my original proposal - but now that I know most of the code is able to handle an empty instance my proposal was to modify this error message instead to have this text https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_core/definitions/partition.py#L614-L618

python_modules/dagster/dagster/_grpc/server.py
```diff
@@ -437,43 +437,77 @@ def ExternalPartitionSetExecutionParams(self, request, _context):
...
+                        instance=instance,
+                    )
+                )
+        else:
+            serialized_data = serialize_dagster_namedtuple(
+                get_partition_set_execution_param_data(
+                    self._get_repo_for_origin(args.repository_origin),
+                    partition_set_name=args.partition_set_name,
+                    partition_names=args.partition_names,
+                )
```
gibsondan, line 456: you could do this fanciness

```
from contextlib import nullcontext
with (DagsterInstance.from_ref(args.instance_ref) if args.instance_ref else nullcontext()) as instance:
   ... # same body for both
```

After https://github.com/dagster-io/dagster/pull/12194 lands this can also try to pull an instance ref from self._instance_ref if there is no args.instance_ref

gibsondan (changes requested):

clairelin135:
Thanks @gibsondan! I followed your suggestions above and updated this PR

gibsondan (approved):

PR #12201, opened 2023-02-09
Rename "Loading multiple repositories" section in workspace files docs

gibsondan:
Summary:
User correctly pointed out that the name here is confusing (it's about different environments not different repositories) and applies as much to Definitions as it does to repositories.


erinkcochran87 (approved):
Ty!

schrockn (approved):

PR #12221, opened 2023-02-09
remove Out.asset_key

sryza:
### Summary & Motivation

This is experimental functionality that we no longer advertise in our docs and is subsumed by SDAs


OwenKephart (approved):
incredible 🙏

PR #12227, opened 2023-02-09
[Docs RFC] Dynamic Partitions

clairelin135:
Adds `DynamicPartitionsDefinition` to API docs and adds a docs snippet describing a use case for dynamic partitions.

One important thing to note is that this feature is experimental and subject to change. Future changes may result in losing partition history and errors of any sort. I attempted to communicate this through the info block, curious if there are better ways of making this information very clear.

Also not 100% sure we should make this part of docs, happy to move this to a github discussion if we feel that will better communicate the experimental nature of this feature.


docs/content/concepts/partitions-schedules-sensors/partitions.mdx
```diff
@@ -133,6 +134,49 @@ result = materialize(
 )
 ```
 
+#### Dynamically partitioned assets (Experimental)
+
+<Note>
+  This feature is <strong>experimental</strong>. Until this feature is marked as
+  stable, partition created prior may not be retained.
```
sryza, line 141: Is it the case in Cloud as well that partitions created may not be retained?

docs/content/concepts/partitions-schedules-sensors/partitions.mdx
```diff
@@ -133,6 +134,49 @@ result = materialize(
...
+    "images_job", AssetSelection.keys("images"), partitions_def=images_partitions_def
+)
+
+
+@sensor(job=images_job)
+def image_sensor(context):
+    new_images = [
+        img_filename
+        for img_filename in os.listdir(os.getenv("MY_DIRECTORY"))
+        if not images_partitions_def.has_partition(img_filename, context.instance)
```
sryza, line 168: This makes me think that we should definitely introduce a `has_partitions` method.

docs/content/concepts/partitions-schedules-sensors/partitions.mdx
```diff
@@ -133,6 +134,49 @@ result = materialize(
...
+```python file=/concepts/partitions_schedules_sensors/dynamic_partitioned_asset.py startafter=start_dynamic_partitions_marker endbefore=end_dynamic_partitions_marker
+images_partitions_def = DynamicPartitionsDefinition(name="images")
+
+
+@asset(partitions_def=images_partitions_def)
+def images(context):
+    ...
+```
+
+For a given dynamic partition set, partition keys can be added and removed. One common pattern is detecting the presence of a new partition through a [sensor](concepts/partitions-schedules-sensors/sensors#sensors), adding the partition and then triggering a run for that partition:
```
sryza, line 155:
```suggestion
For a given dynamic partition set, partition keys can be added and removed. One common pattern is detecting the presence of a new partition through a [sensor](concepts/partitions-schedules-sensors/sensors#sensors), adding the partition, and then triggering a run for that partition:
```

Oxford comma plz

docs/content/concepts/partitions-schedules-sensors/partitions.mdx
```diff
@@ -41,6 +41,7 @@ Having defined a partitioned job or asset, you can:
 | <PyObject object="StaticPartitionsDefinition" />  | A partitions definition with a fixed set of partitions.                           |
 | <PyObject object="MultiPartitionsDefinition" />   | A partitions definition with multiple dimensions.                                 |
 | <PyObject object="MultiPartitionKey" />           | A multi-dimensional partition key.                                                |
+| <PyObject object="DynamicPartitionsDefinition" /> | A partitions definition whose keys can be dynamically added and removed.          |
```
sryza, line 44:
```suggestion
| <PyObject object="DynamicPartitionsDefinition" /> | A partitions definition whose partitions can be dynamically added and removed.          |
```

docs/content/concepts/partitions-schedules-sensors/partitions.mdx
```diff
@@ -133,6 +134,49 @@ result = materialize(
...
 ```
 
+#### Dynamically partitioned assets (Experimental)
+
+<Note>
+  This feature is <strong>experimental</strong>. Until this feature is marked as
+  stable, partition created prior may not be retained.
+</Note>
+
+The <PyObject object="DynamicPartitionsDefinition" /> class accepts a name argument, representing the name of the partition set:
```
sryza, line 144: Before jumping into the details of the classes, I think it's worth providing a sentence of explanation for why someone would want to use this feature.  E.g. "Sometimes you don't know the set of partitions ahead of time when you're defining your assets. For example, maybe you want to add a new partition every time a new data file lands in a directory. In these cases, you can use a `DynamicPartitionsDefinition`."

docs/content/concepts/partitions-schedules-sensors/partitions.mdx
clairelin135: Depends on to what extent we want to support existing users using the feature. In Cloud we could always run a data migration to retain existing data

sryza: If we're writing a data migration, would it not work in open source as well?

clairelin135: I think it would work. We would have to make sure users run the data migration if they are affected, though I've never added a data migration in OSS so not super sure about the specifics

docs/content/concepts/partitions-schedules-sensors/partitions.mdx
clairelin135: 👍 👍

docs/content/concepts/partitions-schedules-sensors/partitions.mdx
clairelin135: this sounds great! thanks for the blurb

docs/content/concepts/partitions-schedules-sensors/partitions.mdx
clairelin135: We're just not sure if the schema that currently exists will require major changes in the future. Managing backcompat state / data migrations in open source feels tricky given that we can't force everyone to migrate, so I'd like to avoid a situation where we need to do this especially since this is an experimental feature

sryza: @prha - as someone who has done data migrations in open source, curious if you have thoughts here?

My instinct is that it's fine to say something like "Until this feature is marked stable, you maybe required to run migrations to ensure that it keeps functioning." But wouldn't be surprised if there are hidden complexities even with that?

prha: I don't think there are any hard and fast rules.  A lot of it depends on how much data there is and how bad of a product experience it would be to run the migration.  I have not ever pre-warned about migrations though.

Typically, for data migrations over the event log, I'll do a hard cutover to the new data format and ignore the old data format.   So then in the release notes, I'd say something like, changed experimental data format blahblahblah, and then instructions in case they want to run the data migration (using `dagster instance reindex`) so that their old data shows up.   If it's not experimental, I might do something like a backcompat query and lazily migrate on reads.

For data migrations over run storage, I'll add them to the list of data migrations that automatically run when you run `dagster instance migrate` (which typically is just schema migrations).

docs/content/concepts/partitions-schedules-sensors/partitions.mdx
```diff
@@ -133,6 +134,51 @@ result = materialize(
...
 ```
 
+#### Dynamically partitioned assets (Experimental)
+
+<Note>
+  This feature is <strong>experimental</strong>. Until this feature is marked as
+  stable, partition created prior may not be retained.
+</Note>
+
+Sometimes you don't know the set of partitions ahead of time when you're defining your assets. For example, maybe you want to add a new partition every time a new data file lands in a directory. Or every time you want to experiment with a new set of hyperparameters. In these cases, you can use a <PyObject object="DynamicPartitionsDefinition"/>.
```
erinkcochran87, line 144:
```suggestion
Sometimes you don't know the set of partitions ahead of time when you're defining your assets. For example, maybe you want to add a new partition every time a new data file lands in a directory, or every time you want to experiment with a new set of hyperparameters. In these cases, you can use a <PyObject object="DynamicPartitionsDefinition"/>.
```

Sort of a nit - `Or every time...` wasn't a complete sentence

docs/content/concepts/partitions-schedules-sensors/partitions.mdx
```diff
@@ -133,6 +134,51 @@ result = materialize(
 )
 ```
 
+#### Dynamically partitioned assets (Experimental)
```
erinkcochran87, line 137:
```suggestion
#### Dynamically partitioned assets
```

I think `(Experimental)` was breaking the on page nav, so I'm suggesting we remove it. 🤷‍♀️

examples/docs_snippets/docs_snippets_tests/concepts_tests/partitions_schedules_sensors_tests/test_dynamic_partitioned_asset.py
```diff
@@ -0,0 +1,42 @@
...
+
+            result = materialize(
+                [images], instance=instance, partition_key=partition_key
+            )
+            assert result.success
+            assert (
+                result.asset_materializations_for_node("images")[0].partition
+                == partition_key
+            )
+        mockenv.stop()
```
sryza, line 42: should this be in a `finally` block? i.e. is there a risk that `os.environ` will remain patched after this test exits?

sryza (approved):
Nice!

examples/docs_snippets/docs_snippets_tests/concepts_tests/partitions_schedules_sensors_tests/test_dynamic_partitioned_asset.py
clairelin135: good call

PR #12230, opened 2023-02-09
[asset-reconciliation][perf] Perf regression tests

OwenKephart:
### Summary & Motivation

We want to have an automated way of detecting when changes result in reduced performance for the reconciliation sensor. This does that. Note that these "target_execution_time" values represent an approximate current state of the world, and future performance improvements should be accompanied by updates to these values.

As a bonus, this gives quick ways to profile performance over a few different shapes of graphs, as the historical runs have already been created (this can be an extremely time-consuming process). All you have to do is kick off the pytest for your desired scenario and profile it.


python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/test_asset_reconciliation_sensor_perf.py
```diff
@@ -0,0 +1,366 @@
...
+            # write as a compressed file
+            with tarfile.open(self.path, "w:gz") as tf:
+                for fname in os.listdir(temp_dir):
+                    tf.add(os.path.join(temp_dir, fname), arcname=fname)
+
+
+class PerfScenario(NamedTuple):
+    snapshot: InstanceSnapshot
+    n_freshness_policies: int
+    target_execution_time: int
```
sryza, line 192: can we include unit on this?

sryza: and should we call it "max" instead of "target"?

sryza (commented):
1. Very cool
2. I'm interested in the implications of including these large files in our repo. How much does it increase the size of a Dagster checkout? Will they be included when someone does `pip install dagster`?

OwenKephart:
@sryza good point re: the large files -- I think the 5MB ones are fine, but 20+MB is a ton. One solution here is to not check those specific ones in, and skip them in pytest. I can replace those tests with ones that just have a single partition backfilled (which should result in a much smaller snapshot, without much change in their ability to discover issues)

sryza:
@alangenfeld - curious if you have thoughts on this PR? Particularly the part about adding big files to the git repo.

sryza:
`git rev-list --disk-usage --objects origin/master` spits out `503361265` (500 MB)

Which is more than I would have thought. Not 100% sure I'm using this command correctly.

alangenfeld:
We have other database snapshots in the repo for things like backcompat tests. The fact that we are zipping stuff up here goes further than we have in the past. I personally think this is fine. 

> Will they be included when someone does pip install dagster?
No, you can check `python_modules/dagster/MANIFEST.in` for what gets non py files get packaged

sryza:
Cool, I'm on board with including these. Can we include a README in the directory where they live explaining where they come from?

If we think we can discover most issues equally well using just the smaller ones, I think it's worth cutting out the bigger ones.

sryza (approved):
Nice!

PR #12235, opened 2023-02-09
[pythonic config] Rename pythonic config classes

benpankow:
## Summary

Renames some of our Pythonic config classes based on offline discussion.

Naming feedback being collected at https://github.com/dagster-io/internal/discussions/4862.

`Resource` -> `ConfigurableResource`
`StructuredConfigIOManager` -> `ConfigurableIOManager`
`StructuredConfigIOManagerBase` -> `ConfigurableIOManagerInjector`
`ResourceOutput` -> `Resource`

## Test Plan

Existing unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_config/structured_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/structured_config/__init__.py
@@ -64,7 +64,7 @@ from dagster._core.storage.io_manager import IOManager, IOManagerDefinition
 from .typing_utils import BaseResourceMeta, LateBoundTypesForResourceTypeChecking
 from .utils import safe_is_subclass
 
-Self = TypeVar("Self", bound="Resource")
+Self = TypeVar("Self", bound="ConfigurableResource")
 
 
 class MakeConfigCacheable(BaseModel):
@@ -333,7 +333,7 @@ class IOManagerWithKeyMapping(ResourceWithKeyMapping, IOManagerDefinition):
 def attach_resource_id_to_key_mapping(
     resource_def: Any, resource_id_to_key_mapping: Dict[ResourceId, str]
 ) -> Any:
-    if isinstance(resource_def, (Resource, PartialResource)):
+    if isinstance(resource_def, (ConfigurableResource, PartialResource)):
         return (
             IOManagerWithKeyMapping(resource_def, resource_id_to_key_mapping)
             if isinstance(resource_def, IOManagerDefinition)
@@ -342,7 +342,7 @@ def attach_resource_id_to_key_mapping(
     return resource_def
 
 
-class Resource(
+class ConfigurableResource(
     Generic[TResValue],
     ResourceDefinition,
     Config,
@@ -359,7 +359,7 @@ class Resource(
     Example:
     .. code-block:: python
 
-        class WriterResource(Resource):
+        class WriterResource(ConfigurableResource):
             prefix: str
 
             def output(self, text: str) -> None:
@@ -476,9 +476,9 @@ class PartialResource(
     Generic[TResValue], ResourceDefinition, AllowDelayedDependencies, MakeConfigCacheable
 ):
     data: Dict[str, Any]
-    resource_cls: Type[Resource[TResValue]]
+    resource_cls: Type[ConfigurableResource[TResValue]]
 
-    def __init__(self, resource_cls: Type[Resource[TResValue]], data: Dict[str, Any]):
+    def __init__(self, resource_cls: Type[ConfigurableResource[TResValue]], data: Dict[str, Any]):
         resource_pointers, data_without_resources = _separate_resource_params(data)
 
         MakeConfigCacheable.__init__(self, data=data, resource_cls=resource_cls)  # type: ignore  # extends BaseModel, takes kwargs
@@ -505,9 +505,9 @@ class PartialResource(
         )
 
 
-ResourceOrPartial: TypeAlias = Union[Resource[TResValue], PartialResource[TResValue]]
+ResourceOrPartial: TypeAlias = Union[ConfigurableResource[TResValue], PartialResource[TResValue]]
 ResourceOrPartialOrValue: TypeAlias = Union[
-    Resource[TResValue], PartialResource[TResValue], ResourceDefinition, TResValue
+    ConfigurableResource[TResValue], PartialResource[TResValue], ResourceDefinition, TResValue
 ]
 
 
@@ -518,14 +518,14 @@ class ResourceDependency(Generic[V]):
     def __set_name__(self, _owner, name):
         self._name = name
 
-    def __get__(self, obj: "Resource", __owner: Any) -> V:
+    def __get__(self, obj: "ConfigurableResource", __owner: Any) -> V:
         return getattr(obj, self._name)
 
     def __set__(self, obj: Optional[object], value: ResourceOrPartialOrValue[V]) -> None:
         setattr(obj, self._name, value)
 
 
-class StructuredResourceAdapter(Resource, ABC):
+class ConfigurableResourceAdapter(ConfigurableResource, ABC):
     """
     Adapter base class for wrapping a decorated, function-style resource
     with structured config.
@@ -545,7 +545,7 @@ class StructuredResourceAdapter(Resource, ABC):
 
             return output
 
-        class WriterResource(StructuredResourceAdapter):
+        class WriterResource(ConfigurableResourceAdapter):
             prefix: str
 
             @property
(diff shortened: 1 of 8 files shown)


benpankow:
* master
  * **PR #12235** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12235" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12235?utm_source=stack-comment).

schrockn:
Did you write a doc/README anywhere to run through how we would explain this to users?

benpankow:
> Did you write a doc/README anywhere to run through how we would explain this to users?

#12260 is a rewrite of the Resource docs using the new names

python_modules/dagster/dagster/_config/structured_config/__init__.py
```diff
@@ -379,7 +381,7 @@ def _wrap_config_type(
 
 
 class PartialIOManager(PartialResource[IOManager], IOManagerDefinition):
-    def __init__(self, resource_cls: Type[StructuredConfigIOManagerBase], data: Dict[str, Any]):
+    def __init__(self, resource_cls: Type[ConfigurableIOManagerBase], data: Dict[str, Any]):
```
schrockn, line 384: per discussion let's not say "ConfigurableIOManagerBase"

Proposals

`ConfigurableIOManagerFactory`
`ConfigurableIOManagerInjector`

I lean towards Injector since it is actually more specific than Factory. The Injector knows specially how to interact with our runtime. It "injects" a resource/IOManager into user-defined computations as runtime. `ConfigurableIOManagerFactory` sounds like it might create objects with type `ConfigurableIOManager` which is not true. It creates  objects with type`IOManager`.

schrockn (dismissed):
No base please

python_modules/dagster/dagster/_config/structured_config/__init__.py
schrockn: Cool. Given that recommend renaming `create_io_manager_to_pass_to_user_code` to `create_io_manager_to_inject_into_user_code`

benpankow: 👍 (updated `create_object_to_pass_to_user_code` to `create_resource_to_inject_into_user_code` as well)

schrockn (changes requested):
Minor point the method name, but I think this stuff matter a lot.

Also req'ing changes to see if @sryza has a moment of inspiration around an alternative name to `ConfigurableIOManagerInjector`

sryza:
ConfigurableIOManagerFactory?

sryza:
Oh, just saw that that was already suggested. Will think on this a little more.

sryza:
I think I would expect something called Injector to function like this:

```python
class MyInjector(Injector):
    def perform_injection(injectee: Injectee) -> None:
        injectee.inject_stuff(my_stuff)
```

Another option could be `IOManagerConfiguration`?

python_modules/dagster/dagster/_config/structured_config/__init__.py
```diff
@@ -306,19 +308,21 @@ class StructuredConfigIOManagerBase(Resource[IOManager], IOManagerDefinition):
...
     def __init__(self, **data: Any):
         schema = infer_schema_from_config_class(self.__class__)
-        Resource.__init__(self, **_process_config_values(schema, data, self.__class__.__name__))
+        ConfigurableResource.__init__(
+            self, **_process_config_values(schema, data, self.__class__.__name__)
+        )
         IOManagerDefinition.__init__(
             self,
-            resource_fn=self.create_io_manager_to_pass_to_user_code,
+            resource_fn=self.create_io_manager_to_inject_into_user_code,
```
schrockn, line 316: prefer `create_io_manager_to_inject`

schrockn (approved):
Ok let's move forward here.

One remaining question here that we should answer in docs is why we don't have a `ConfigurableResourceInjector` while do have a `ConfigurableIOManagerInjector`

sryza:
Based on my last comment and having thought about it a little more, I think I prefer "factory" to "injector", because:
- "Factory" has known semantics that seem to match the shape of what this does
- The semantics that I would guess for a name "Injector" don't seem to match what this does

benpankow:
> Based on my last comment and having thought about it a little more, I think I prefer "factory" to "injector", because:
> 
> * "Factory" has known semantics that seem to match the shape of what this does
> * The semantics that I would guess for a name "Injector" don't seem to match what this does

I'm happy to move over to `-Factory` if that feels better to you both. @schrockn what do you think?

sryza:
Btw I am ok with being overruled on this

schrockn:
Here is my case for injector. (This is fairly loosely held). I am imagining a guide or piece of content written about this system for an engineering audience and people who care about testing/dep injection and the design considerations of Dagster. The people who would who actually author resources for their stakeholders to use. Imagine this section of content:

---

### What is going on here?

Resources perform another function in Dagster: dependency injection. What is happening here is that you, as the programmer, are creating an object to pass to Definitions. The job of _that_ object in life is twofold: (1) to define configuration that Dagster users and practitioners can specify at runtime and 2) to create another object that the framework then passes to user code to be programmed against. We call this process _injection_.

In this framing, the `resources` parameter to `Definitions` should have been called `injectors` or `resource_injectors`. However our goal is that the majority of Dagster's users do _not_ have to deeply understand the mechanics of this unless they want to. By default an injector injects _itself_ into user code, and therefore the distinction is immaterial.

However, there are important use cases where the separation of concerns between the object that is configured and the object that is injected and programmed against is desirable. Consider the case where the practitioner wants to program directly against third-party library code (e.g. a connection object in an `acmedb` module), but wants runtime configuration. 

```
class AcmeConnResource(ResourceInjector):
   conn: str
   
   def create_object_to_inject(_):
      return acmedb.create_conn(self.conn)

@asset
def some_asset(acme_conn: Resource[AcmeConn]):
   acme_conn.execute("do something")
   
# in a test
def test_some_asset():
    # someone else on the internet wrote a library to test acmedb
    some_asset(acme_conn=acmedbmocker.create_mock_conn())
```

Another case is where the user wants to compose a resource object out of other Python objects, rather than just configuration, so they can use and test that object flexibly and reuse it directly outside of the Dagster context.

```
class PolicyEngineResource(ResourceInjector):
   policies: List[str]

   def create_object_to_inject(_):
       return PolicyEngine([create_policy(policy) for policy in self.policies])

def test_policy_engine():
    assert PolicyEngine([DummyPolicy()])
---

We think this provides a middle ground that is more approachable than traditional, more invasive dependency injection. In this approach, there is a single seam of testability in the program and that seam is specifically where business logic code invokes code that interacts with external services. The framework takes responsible for managing that object and injecting it into user code, rather than burdening the user with having to thread objects up and down the stack manually.

**Comment by sryza** [comment]
```python
class AcmeConnResource(ResourceInjector):
   conn: str
   
   def create_object_to_inject(_):
      return acmedb.create_conn(self.conn)

@asset
def some_asset(acme_conn: Resource[AcmeConn]):
   acme_conn.execute("do something")
   
# in a test
def test_some_asset():
    # someone else on the internet wrote a library to test acmedb
    some_asset(acme_conn=acmedbmocker.create_mock_conn())
```

@schrockn one of the things that made me excited about the resource revamp was that I thought we'd be able to push this complexity into user space and thus simplify the framework.  I.e. expect them to do this instead:

```python

**Comment by schrockn** [comment]
I think there is critical set of subusers in our persona that will want to plug in at this layer. I, personally, would be extremely annoyed (perhaps even offended or enraged?) if the framework forced me to write this:

```
# in a test
def test_some_asset():
    # someone else on the internet wrote a library to test acmedb
    some_asset(acme=MagicMock(create_conn=MagicMock(return_value=acmedbmocker.create_mock_conn())))
```

just because the framework wanted to hide something from me that I know _must_ exist given the behavior I am seeing elsewhere.

However I am very much for pushing it "down" the tech tree as much as possible, with the knowledge that it will leak into some examples.

benpankow:
Going to merge this to minimize thrash for stacked PRs. Happy to revisit the `injector` question in a follow-up, since it's used in far fewer places changing it again won't be a big lift.

sryza:
> just because the framework wanted to hide something from me that I know must exist given the behavior I am seeing elsewhere.

My (likely flawed, still ramping up on the internals of this) understanding here is that we're actually _still_ hiding what's going on, even with the injector.  I.e. if I understand correctly, the non-injector case looks like this under the covers:
- The user constructs their dataclass to pass it to the `resources` argument of `Definitions`.
- Within the step, we secretly construct a different instance of their dataclass, using the resolved config.

When an injector is used, we add a new third step after that second step, which takes the secret-constructed instance of the dataclass and invokes a method on it.  But the framework is still hiding that second step.

---

That all said, I agree that there are a set of situations where, practically, the injector provides superior economics. I won't push further against including it

---

I do think I've become a little more convicted in my concerns about the name. Basically: if a class has a name of the form "_verb_-er", I believe it's expected to have a method (or methods) that perform the verb.  E.g. a Visitor has visit methods that do visiting. A RunLauncher has a `launch_run` method. Sensors don't have an explicit "sense" method, but they are expected to do sensing inside their evaluation function.

schrockn:
@sryza quick update here we're not going to use the injector terminology and instead use your suggestion of "factory" when appropriate.

PR #12237, opened 2023-02-09
[asset-reconciliation][perf] Better caching of most recent materializations

OwenKephart:
### Summary & Motivation

Motivated by https://github.com/dagster-io/dagster/pull/12230/files, this takes some easy wins in the perf department to push down the execution time for the given test cases (note the lowered `target_execution_time` values)


sryza (approved):
Nice

Issue #12188, opened 2023-02-08
InstanceBackfillsQuery: 'NoneType' object is not iterable


spenczar:
### Dagster version

dagster, version 1.1.17

### What's the issue?

I was on the Assets page for a particular asset with many, many partitions - 10,000 partitions. I clicked "materialize asset" and selected about 350 partitions. I started the backfill job, and clicked the button which navigates me to `/overview/backfills`. I got the following error:

<img width="1640" alt="Screen Shot 2023-02-08 at 11 39 05 AM" src="https://user-images.githubusercontent.com/880413/217634137-1d42e2c5-c6df-4015-b144-9b07917aa163.png">

```
Operation name: InstanceBackfillsQuery

Message: 'NoneType' object is not iterable

Path: ["partitionBackfillsOrError","results",3,"numPartitions"]

Locations: [{"line":7,"column":9},{"line":62,"column":3}]
```

Every several seconds, "Unexpected GraphQL error" reappears on the top right.

### What did you expect to happen?

I expected to see the backfills.

### How to reproduce?

I don't know.

### Deployment type

Dagster Helm chart

### Deployment details

This dagit instance is running on a Kubernetes cluster, and I am using `kubectl port-forward` to be able to view it at a localhost address.

### Additional information

When I refresh the page, I still see the same error every time.

When I go to `/runs`, I see all 353 runs are in progress.

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

sryza:
Thanks for reporting this @spenczar - any chance there's a stack trace in your Dagit logs that correspond to this error?

spenczar:
Sorry to say there's nothing in the logs. I'm still able to reproduce this at will, though. Any time I go to /overview/backfills I get this error. If you'd like me to `kubectl exec` into the dagit container and do anything to poke around, I can.

sryza:
We think we've been able to reproduce this internally. @clairelin135 is investigating a fix.

PR #12238, opened 2023-02-10
Fix partitions backfill deserialization error

clairelin135:
Fixes https://github.com/dagster-io/dagster/issues/12188. This specific issue occurred when a backfill existed for a time-window partitions definition, and the partitions definition was changed to a static partitions definition.

This PR amends existing behavior to guard against partitions definition changes, instead displaying an informative message when the partitions definition has changed and the backfill data can no longer be deserialized:
image

This PR implements this by storing the `serializable_unique_ids` of each partitions definition in the backfill, as well as a type representing the type of partitions definition class (static, multipartitioned, etc.). Each partitions subset now has a `can_deserialize` method that accepts the serialized string, partitions definition, the serialized unique ID, and the type of the serialized subset:
- When the type of the serialized partitions def is not the type of the current partitions definition, `can_deserialize` returns False
- For deserializing time window partitions subsets, first checks the type. Then, checks that the serializable unique ID is the same (meaning that the time window partitions subset hasn't changed to another time interval)
- Default partitions subsets return `True` as long as the type is correct. This allows us to still display serialized partition keys even if the partition keys no longer exist, and display existing backfills even if the more partition keys are added
- In backcompat cases, loads the content of the serialized data to check validity


sryza:
This is awesome. I added @bengotow as a reviewer to have a look at the frontend parts.

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
```diff
@@ -325,7 +327,13 @@ def __init__(
     ):
         check.inst_param(partitions_def, "partitions_def", MultiPartitionsDefinition)
         subset = (
-            set(partitions_def.get_partition_key_from_str(key) for key in subset)
```
clairelin135, line 328: If a user has existing static partition backfills, changing the asset's partitions def to a multipartitioned asset would raise an error. This is because a multipartitions subset is serialized the same way as a default partitions subset, so we can't distinguish between the two using the serialized content. This line would then error because the static partition keys don't contain the "|" delimiter.

This change would enable existing static partition backfills to load even after the partitions def has been changed to a multipartitions def, by filtering out any invalid keys here

python_modules/dagster/dagster/_core/definitions/partition.py
```diff
@@ -1386,6 +1445,22 @@ def from_serialized(
...
+        cls,
+        partitions_def: PartitionsDefinition,
+        serialized: str,
+        serialized_partitions_def_unique_id: Optional[str],
+        subset_type: Optional[PartitionsSubsetType],
+    ) -> bool:
+        if subset_type is not None:
+            return subset_type == cls.subset_type()
+
+        data = json.loads(serialized)
```
sryza, line 1459: Does this mean that we'd now be deserializing the data twice when previously we were doing it once? I think this deserialization can be a pretty expensive operation in some situations, so ideally we'd just do it once.

python_modules/dagster/dagster/_core/definitions/partition.py
```diff
@@ -217,7 +217,22 @@ def __lt__(self, other: "ScheduleType") -> bool:
         return self.ordinal < other.ordinal
 
 
+@whitelist_for_serdes
+class PartitionsSubsetType(Enum):
```
sryza, line 221: I'm wondering if there's a way to avoid introducing this meta-concept and the associated complexity.

What if we serialized `partitions_def.__class__.__name__` as part of `partitions_def_class_names_by_asset_key` instead of  `subset_types_by_asset_key` and said `can_deserialize` is `False` if `partitions_def_class_names_by_asset_key[asset_key] != asset_graph.get_partitions_def(asset_key).__class__.__name__`?

python_modules/dagster/dagster/_core/definitions/partition.py
clairelin135: Yes, but in the future when all backfills have the subset type and serializable partitions def unique ID, we won't need to deserialize the data in order to check the validity. We only need to deserialize for the backcompat cases where we don't have the subset type / unique ID.

I think that the number of backcompat cases shouldn't be excessively high given that we only added serialized asset backfill data a month ago.

python_modules/dagster/dagster/_core/definitions/partition.py
clairelin135: I considered that too, though if the partitions subset class name changed then that would unexpectedly make all historic backfills display with the "partitions def has changed" error in the UI.

I figured that if we did this instead, if class names changed we would still have a backcompat route

sryza: I was thinking the PartitionsDefinition subclass - if those class names change, we'd have much bigger problems on our hands

python_modules/dagster/dagster/_core/definitions/partition.py
sryza: Ah I see - just the backcompat cases seems fine

clairelin135:
@sryza makes sense about serializing the partition definition class name instead--I've amended the code to do this instead.

I also tried adding the `empty_subset` abstract method you were mentioning above. Not entirely sure if what I added is what you meant, let me know what you think

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
sryza: FWIW, I think it's fine if you can't load a backfill after moving from static to multi-partitioned.

sryza (commented):
Left one last comment, otherwise this looks great

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
clairelin135: The struggle is that the whole backfill page won't load if a single backfill can't be deserialized, which is the reason I'm trying to support this situation. 

Maybe @bengotow knows how to change the backfill page to avoid breaking the whole page if a single row can't be deserialized?

python_modules/dagster/dagster/_core/definitions/partition.py
```diff
@@ -1386,6 +1431,22 @@ def from_serialized(
...
 
+    @classmethod
+    def can_deserialize(
+        cls,
+        partitions_def: PartitionsDefinition,
+        serialized: str,
+        serialized_partitions_def_unique_id: Optional[str],
+        serialized_partitions_def_class_name: Optional[str],
+    ) -> bool:
+        if serialized_partitions_def_class_name is not None:
```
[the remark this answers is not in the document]
clairelin135: We do do this check for each subset implementation. But the default subset validity logic looks different from the time window--the if the class is correct for the default case, we know the object can be deserialized. In the time window case, the class has to be correct and the serialized unique ID has to be correct too.

Since this logic can be a little more complicated, I'd prefer keeping this logic with the other deserialization logic

js_modules/dagit/packages/core/src/instance/BackfillRow.tsx
```diff
@@ -131,10 +131,14 @@ export const BackfillRow = ({
...
-          <BackfillRunStatus backfill={backfill} counts={counts} statuses={statuses} />
+        {backfill.isValidSerialization ? (
+          counts ? (
+            <BackfillRunStatus backfill={backfill} counts={counts} statuses={statuses} />
+          ) : (
+            <LoadingOrNone queryResult={queryResult} />
+          )
         ) : (
-          <LoadingOrNone queryResult={queryResult} />
+          <p>A partitions definition has changed since this backfill ran.</p>
```
bengotow, line 141: Not sure if this should say "partition definition" or "partitions definition"?

js_modules/dagit/packages/core/src/instance/BackfillTable.mocks.ts
```diff
@@ -336,6 +340,7 @@ export const BackfillTableFragmentCompletedAssetJobStatus: MockedResponse<Single
 export const BackfillTableFragmentCompletedOpJob: BackfillTableFragment = {
   backfillId: 'pqdiepuf',
   status: BulkActionStatus.COMPLETED,
+  isValidSerialization: true,
```
bengotow, line 343: Not a blocker, but it would be nice to add a mock for the the isValidSerialization:false case. It seems like it could be tricky for us front-end devs to see this case locally without spending a couple minutes setting up a backfill with an out-of-date def!

bengotow (approved):
The front-end part of this look good Claire! Left one note about adding a test but I can also throw one in my next storybook PR, doesn't need to block this from shipping 👋

python_modules/dagster/dagster/_core/definitions/partition.py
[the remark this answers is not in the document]
sryza: The class name piece is the same across all of them though, right?

Feel free to use your judgment on this one.

sryza (approved):

js_modules/dagit/packages/core/src/instance/BackfillTable.mocks.ts
clairelin135: sounds good, I added this test!

js_modules/dagit/packages/core/src/instance/BackfillRow.tsx
clairelin135: Hmm... our definitions objects are called `PartitionsDefinition` so I lean toward following this

PR #12239, opened 2023-02-10
treat pyright warnings as errors for buildkite purposes

gibsondan:
Summary:
We are at zero warnings now in dagster and 2 easy warnings in internal, may as well keep it that way?

Test Plan: BK, make pyright locally


smackesey (approved):
Thanks, I didn't realize it wasn't already doing this.

gibsondan:
we'll need to land https://github.com/dagster-io/internal/pull/4847 before I can land this

PR #12242, opened 2023-02-10
[typing/runtime] [gql] eliminate resolver **kwargs (3)

smackesey:
### Summary & Motivation

Replace resolver `**kwargs` with typed keyword args. Part of a collection of PRs that eliminates resolver `**kwargs` across all resolvers.

**NOTE**: The GQL schema in this PR is changed, but its a non-functional change-- the change arises from the parameter order of the signature of `resolve_isPipelineConfigValid`-- keyword args with defaults need to appear at the end of the param list in a python function. This changes the order in which the args are listed in the generated schema, but this doesn't matter because args are always passed by name in GQL.

BK. Also, correctness of added args was confirmed by manually inspecting the field definition in the parent object.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/schema/roots/query.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/roots/query.py
@@ -1,4 +1,4 @@
-from typing import Dict, Mapping, Optional, Sequence, cast
+from typing import Any, Dict, Mapping, Optional, Sequence, cast
 
 import dagster._check as check
 import graphene
@@ -634,12 +634,12 @@ class GrapheneDagitQuery(graphene.ObjectType):
         graphene_info: ResolveInfo,
         pipeline: GraphenePipelineSelector,
         mode: str,
-        runConfigData: Optional[GrapheneRunConfigData] = None,
+        runConfigData: Optional[Any] = None,  # custom scalar (GrapheneRunConfigData)
     ):
         return validate_pipeline_config(
             graphene_info,
             pipeline_selector_from_graphql(pipeline),
-            parse_run_config_input(runConfigData or {}, raise_on_error=False),  # type: ignore
+            parse_run_config_input(runConfigData or {}, raise_on_error=False),
             mode,
         )
 
@@ -648,12 +648,12 @@ class GrapheneDagitQuery(graphene.ObjectType):
         graphene_info: ResolveInfo,
         pipeline: GraphenePipelineSelector,
         mode: str,
-        runConfigData: Optional[GrapheneRunConfigData] = None,
+        runConfigData: Optional[Any] = None,  # custom scalar (GrapheneRunConfigData)
     ):
         return get_execution_plan(
             graphene_info,
             pipeline_selector_from_graphql(pipeline),
-            parse_run_config_input(runConfigData or {}, raise_on_error=True),  # type: ignore
+            parse_run_config_input(runConfigData or {}, raise_on_error=True),  # type: ignore  # (possible str)
             mode,
         )
 
@@ -681,8 +681,7 @@ class GrapheneDagitQuery(graphene.ObjectType):
         assetKeys: Optional[Sequence[GrapheneAssetKeyInput]] = None,
     ) -> Sequence[GrapheneAssetNode]:
         resolved_asset_keys = set(
-            AssetKey.from_graphql_input(cast(Mapping[str, Sequence[str]], asset_key_input))
-            for asset_key_input in assetKeys or []
+            AssetKey.from_graphql_input(asset_key_input) for asset_key_input in assetKeys or []
         )
         use_all_asset_keys = len(resolved_asset_keys) == 0
 
--- a/python_modules/dagster-graphql/dagster_graphql/schema/roots/query.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/roots/query.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Mapping, Sequence, cast
+from typing import Dict, Mapping, Optional, Sequence, cast
 
 import dagster._check as check
 import graphene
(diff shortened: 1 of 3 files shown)


smackesey:
* master
  * **PR #12242** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12242" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12242?utm_source=stack-comment).

python_modules/dagster-graphql/dagster_graphql/schema/roots/query.py
```diff
@@ -596,43 +629,69 @@ def resolve_pipelineRunTags(self, graphene_info: ResolveInfo):
     def resolve_runGroupOrError(self, graphene_info: ResolveInfo, runId):
         return get_run_group(graphene_info, runId)
 
-    def resolve_isPipelineConfigValid(self, graphene_info: ResolveInfo, pipeline, **kwargs):
+    def resolve_isPipelineConfigValid(
+        self,
+        graphene_info: ResolveInfo,
+        pipeline: GraphenePipelineSelector,
+        mode: str,
+        runConfigData: Optional[GrapheneRunConfigData] = None,
```
alangenfeld, line 637: [1] since GrapheneRunConfigData represents a "custom scalar", i don't think using that type here is right. Theoretically we could implement more on that class and align to clearer types that way. For now might be best to just do Any / Optional[object] / what we expect which is what `parse_run_config_input` is typed to.

https://docs.graphene-python.org/en/latest/types/scalars/#custom-scalars

python_modules/dagster-graphql/dagster_graphql/schema/roots/query.py
```diff
@@ -596,43 +629,69 @@ def resolve_pipelineRunTags(self, graphene_info: ResolveInfo):
...
+            mode,
         )
 
-    def resolve_executionPlanOrError(self, graphene_info: ResolveInfo, pipeline, **kwargs):
+    def resolve_executionPlanOrError(
+        self,
+        graphene_info: ResolveInfo,
+        pipeline: GraphenePipelineSelector,
+        mode: str,
+        runConfigData: Optional[GrapheneRunConfigData] = None,
```
alangenfeld, line 651: [1]

python_modules/dagster-graphql/dagster_graphql/schema/roots/query.py
```diff
@@ -596,43 +629,69 @@ def resolve_pipelineRunTags(self, graphene_info: ResolveInfo):
...
+    def resolve_assetNodes(
+        self,
+        graphene_info: ResolveInfo,
+        loadMaterializations: bool,
+        group: Optional[GrapheneAssetGroupSelector] = None,
+        pipeline: Optional[GraphenePipelineSelector] = None,
+        assetKeys: Optional[Sequence[GrapheneAssetKeyInput]] = None,
+    ) -> Sequence[GrapheneAssetNode]:
+        resolved_asset_keys = set(
+            AssetKey.from_graphql_input(cast(Mapping[str, Sequence[str]], asset_key_input))
```
alangenfeld, line 684: [2] this method exists for these call site, so it feels off that it requires casting like this. Seems likely worth updating the method definition.

python_modules/dagster-graphql/dagster_graphql/schema/roots/query.py
```diff
@@ -708,45 +770,62 @@ def load_asset_graph() -> ExternalAssetGraph:
             for node in results
         ]
 
-    def resolve_assetNodeOrError(self, graphene_info: ResolveInfo, **kwargs):
-        asset_key_input = cast(Mapping[str, Sequence[str]], kwargs.get("assetKey"))
+    def resolve_assetNodeOrError(self, graphene_info: ResolveInfo, assetKey: GrapheneAssetKeyInput):
+        asset_key_input = cast(Mapping[str, Sequence[str]], assetKey)
```
alangenfeld, line 774: [2]

python_modules/dagster-graphql/dagster_graphql/schema/roots/query.py
```diff
@@ -708,45 +770,62 @@ def load_asset_graph() -> ExternalAssetGraph:
...
-            limit=kwargs.get("limit"),
+            prefix=prefix,
+            cursor=cursor,
+            limit=limit,
         )
 
-    def resolve_assetOrError(self, graphene_info: ResolveInfo, **kwargs):
-        return get_asset(graphene_info, AssetKey.from_graphql_input(kwargs["assetKey"]))
+    def resolve_assetOrError(self, graphene_info: ResolveInfo, assetKey: GrapheneAssetKeyInput):
+        return get_asset(graphene_info, AssetKey.from_graphql_input(assetKey))
```
alangenfeld, line 792: [2] how is this one fine but the others needed casting?

python_modules/dagster-graphql/dagster_graphql/schema/roots/query.py
```diff
@@ -708,45 +770,62 @@ def load_asset_graph() -> ExternalAssetGraph:
...
 
-    def resolve_assetNodeDefinitionCollisions(self, graphene_info: ResolveInfo, **kwargs):
-        raw_asset_keys = cast(Sequence[Mapping[str, Sequence[str]]], kwargs.get("assetKeys"))
+    def resolve_assetNodeDefinitionCollisions(
+        self,
+        graphene_info: ResolveInfo,
+        assetKeys: Optional[Sequence[GrapheneAssetKeyInput]] = None,
+    ):
+        assert assetKeys is not None
+        raw_asset_keys = cast(Sequence[Mapping[str, Sequence[str]]], assetKeys)
```
alangenfeld, line 800: [2]

alangenfeld (changes requested):

python_modules/dagster-graphql/dagster_graphql/schema/roots/query.py
smackesey: Good catch, it actually doesn't require a `cast`, this was left over. `GrapheneAssetKeyInput`, like all Graphene input objects, functions as an implicit `Any`.

smackesey:
@alangenfeld Including fix for BK / addressing your comments on https://github.com/dagster-io/dagster/pull/12244 here since I accidentally merged that one before comment address.

alangenfeld (approved):

PR #12260, opened 2023-02-10
[draft][pythonic config][docs] Introduce intro to Resources doc utilizing Pythonic resources

benpankow:
## Summary

Introduces a new experimental guide, aiming to introduce the new Pythonic resource and IO manager APIs. This guide assumes some familiarity with Dagster resources, but introduces the new APIs from the ground-up.

todos:
 - expand IO manager content
 - more impl details

## Test Plan
https://dagster-git-benpankow-update-resources-docs-pyt-ef810b-elementl.vercel.app/guides/dagster/pythonic-resources

### DIFF
--- a/docs/content/guides/dagster/pythonic-resources.mdx
+++ b/docs/content/guides/dagster/pythonic-resources.mdx
@@ -246,6 +246,13 @@ You can test the initialization of a <PyObject object="ConfigurableResource"/> b
 ```python file=/guides/dagster/pythonic_resources/pythonic_resources.py startafter=start_new_resource_testing endbefore=end_new_resource_testing
 from dagster._config.structured_config import ConfigurableResource, ConfigurableResourceAdapter
 
+V = TypeVar("V")
+
+
+class ResourceDependency(Generic[V]):
+    def __get__(self, obj: "Resource", __owner: Any) -> V:
+        return getattr(obj, self._name)
+
 
 class MyResource(ConfigurableResource):
     value: str
@@ -340,5 +347,67 @@ def public_github_repos(github: Resource[GitHub]):
 defs = Definitions(
     assets=[public_github_repos],
     resources={"github": GitHub()},
+)
+```
+
+If you would like a Pythonic resource to depend on a bare Python object, you can use the <PyObject object="ResourceDependency"/> annotation to annotate the attribute in question. This indicates to Dagster that the attribute should be treated as a resource dependency.
+
+```python file=/guides/dagster/pythonic_resources/pythonic_resources.py startafter=start_raw_github_resource_dep endbefore=end_raw_github_resource_dep
+class DBResource(ConfigurableResource):
+    engine: ResourceDependency[Engine]
+
+    def query(self, query: str):
+        with self.engine.connect() as conn:
+            return conn.execute(query)
+
+
+engine = create_engine(...)
+defs = Definitions(
+    assets=...,
+    resources={"db": DBResource(engine=engine)},
+)
+```
+
+---
+
+## Implementation details
+
+### Resource input resolution
+
+Resources defined by subclassing <PyObject object="ConfigurableResource"/> can take various unresolved inputs, including `EnvVar`s instead of string values or unconfigured resources returned by `configure_at_launch()` in place of configured resources.
+
+Between construction time and runtime, the <PyObject object="ConfigurableResource"/> class resolves all of its attributes to their final values, meaning that when an attribute of a resource is accessed in the body of an asset or op, it will always be a fully resolved value.
+
+In the following example, the `FileStoreBucket` constructor will accept the still-unconfigured `unconfigured_credentials_resource`, which will be resolved once the resource is configured at runtime.
+
+```python file=/guides/dagster/pythonic_resources/pythonic_resources.py startafter=start_impl_details_resolve endbefore=end_impl_details_resolve
+class CredentialsResource(ConfigurableResource):
+    username: str
+    password: str
+
+
+class FileStoreBucket(ConfigurableResource):
+    credentials: CredentialsResource
+    region: str
+
+    def write(self, data: str):
+        # In this context, `self.credentials` is ensured to
+        # be a CredentialsResource
+
+        get_filestore_client(
+            username=self.credentials.username,
+            password=self.credentials.password,
+            region=self.region,
+        ).write(data)
+
+
+# unconfigured_credentials_resource is typed as PartialResource[CredentialsResource]
+unconfigured_credentials_resource = CredentialsResource.configure_at_launch()
+
+# FileStoreBucket constructor accepts either a CredentialsResource or a
+# PartialResource[CredentialsResource] for the `credentials` argument
+bucket = FileStoreBucket(
+    credentials=unconfigured_credentials_resource,
+    region="us-east-1",
 )
 ```
--- a/docs/content/guides/dagster/pythonic-resources.mdx
+++ b/docs/content/guides/dagster/pythonic-resources.mdx
@@ -346,7 +346,7 @@ from dagster import resource, Definitions
 from dagster._config.structured_config import ConfigurableResourceAdapter
 
 
-# Old code, cannot be changed for back-compat purposes
+# Old code, interface cannot be changed for back-compat purposes
 class Writer:
     def __init__(self, prefix: str):
         self._prefix = prefix
@@ -399,7 +399,7 @@ from dagster._config.structured_config import (
 import os
 
 
-# Old code, cannot be changed for back-compat purposes
+# Old code, interface cannot be changed for back-compat purposes
 class OldFileIOManager(IOManager):
     def __init__(self, base_path: str):
         self.base_path = base_path
@@ -411,7 +411,9 @@ class OldFileIOManager(IOManager):
     def load_input(self, context: InputContext):
         with open(
             os.path.join(
-                self.base_path, context.upstream_output.step_key, context.upstream_output.name
+                self.base_path,
+                context.upstream_output.step_key,
+                context.upstream_output.name,
             ),
             "r",
         ) as fd:
--- a/docs/content/guides/dagster/pythonic-resources.mdx
+++ b/docs/content/guides/dagster/pythonic-resources.mdx
@@ -24,7 +24,7 @@ So, why use resources?
 
 ---
 
-## Using resources
+## Using Pythonic resources
 
 ### Getting started using resources
 
@@ -285,6 +285,33 @@ def test_my_resource_with_context():
     resource = MyResourceRequiresAnother(foo=string_holder, bar="bar")
     assert resource.foo.value == "foo"
     assert resource.bar == "bar"
+```
+
+---
+
+## Defining Pythonic IO managers
+
+Pythonic [IO managers](/concepts/io-management/io-managers) are defined as subclasses of <PyObject object="ConfigurableIOManager"/>, and similarly to Pythonic resources specify any configuration fields as attributes. Each subclass must implement a `handle_output` and `load_input` method, which are called by the Dagster runtime to handle the storing and loading of data.
+
+```python file=/guides/dagster/pythonic_resources/pythonic_resources.py startafter=start_new_io_manager endbefore=end_new_io_manager
+class MyIOManager(ConfigurableIOManager):
+    root_path: str
+
+    def _get_path(self, asset_key: AssetKey) -> str:
+        return self.root_path + "/".join(asset_key.path)
+
+    def handle_output(self, context: OutputContext, obj):
+        write_csv(self._get_path(context.asset_key), obj)
+
+    def load_input(self, context: InputContext):
+        return read_csv(self._get_path(context.asset_key))
+
+defs = Definitions(
+    assets=...,
+    resources={
+        "io_manager": MyIOManager(root_path="/tmp/")
+    },
+)
 ```
 
 ---
--- a/docs/content/guides/dagster/pythonic-resources.mdx
+++ b/docs/content/guides/dagster/pythonic-resources.mdx
@@ -116,7 +116,6 @@ defs = Definitions(
 </TabItem>
 </TabGroup>
 
-
 ### Using environment variables with resources
 
 Resources can be configured using environment variables, which is useful for secrets or other environment-specific configuration. If you're using Dagster Cloud, environment variables can be [configured directly in the UI](/dagster-cloud/developing-testing/environment-variables-and-secrets).
@@ -144,7 +143,6 @@ defs = Definitions(
     },
 )
 ```
-
 
 ### Configuring resources at runtime
 
@@ -182,9 +180,7 @@ Then, configuration for the resource can be provided at runtime in the launchpad
 ```python file=/guides/dagster/pythonic_resources/pythonic_resources.py startafter=start_new_resource_runtime_launch endbefore=end_new_resource_runtime_launch
 from dagster import sensor, define_asset_job, RunRequest
 
-update_data_job = define_asset_job(
-    name="update_data_job", selection=[data_from_database]
-)
+update_data_job = define_asset_job(name="update_data_job", selection=[data_from_database])
 
 
 @sensor(job=update_data_job)
@@ -199,7 +195,6 @@ def table_update_sensor():
             ),
         )
 ```
-
 
 ### Resources which depend on other resources
 
@@ -261,7 +256,6 @@ defs = Definitions(
 )
 ```
 
-
 ### Testing configurable resources
 
 You can test the initialization of a <PyObject object="ConfigurableResource"/> by constructing it manually. In most cases, you can construct your resource directly:
@@ -307,7 +301,7 @@ def test_my_resource_with_context():
 
 ## Defining Pythonic I/O managers
 
-Pythonic [I/O managers](/concepts/io-management/io-managers) are defined as subclasses of <PyObject object="ConfigurableIOManager"/>, and similarly to Pythonic resources specify any configuration fields as attributes. Each subclass must implement a `handle_output` and `load_input` method, which are called  Dagster at runtime to handle the storing and loading of data.
+Pythonic [I/O managers](/concepts/io-management/io-managers) are defined as subclasses of <PyObject object="ConfigurableIOManager"/>, and similarly to Pythonic resources specify any configuration fields as attributes. Each subclass must implement a `handle_output` and `load_input` method, which are called Dagster at runtime to handle the storing and loading of data.
 
 ```python file=/guides/dagster/pythonic_resources/pythonic_resources.py startafter=start_new_io_manager endbefore=end_new_io_manager
 from dagster import Definitions, AssetKey, OutputContext, InputContext
@@ -332,6 +326,8 @@ defs = Definitions(
     resources={"io_manager": MyIOManager(root_path="/tmp/")},
 )
 ```
+
+###
 
 ---
 
@@ -379,8 +375,69 @@ def my_asset(writer: Writer):
     writer.output("hello, world!")
 
 
+defs = Definitions(assets=[my_asset], resources={"writer": WriterResource(prefix="greeting: ")})
+```
+
+### Adapting function-style I/O managers
+
+Similar to with resources, in codebases that utilize [function-style](https://docs.dagster.io/concepts/io-management/io-managers) I/O managers with the <PyObject object="@io_manager"/> decorator, you can use the <PyObject object="ConfigurableIOManagerAdapter"/> class to adapt them to the new Pythonic I/O manager pattern.
+
+The adapter subclass should list all config fields as attributes and implement a `wrapped_io_manager` property which returns the underlying I/O manager definition to adapt.
+
+```python file=/guides/dagster/pythonic_resources/pythonic_resources.py startafter=start_io_adapter endbefore=end_io_adapter
+from dagster import (
+    Definitions,
+    IOManagerDefinition,
+    io_manager,
+    IOManager,
+    InputContext,
+    OutputContext,
+)
+from dagster._config.structured_config import (
+    StructuredIOManagerAdapter as ConfigurableIOManagerAdapter,
+)
+import os
+
+
+# Old code, cannot be changed for back-compat purposes
+class OldFileIOManager(IOManager):
+    def __init__(self, base_path: str):
+        self.base_path = base_path
+
+    def handle_output(self, context: OutputContext, obj):
+        with open(os.path.join(self.base_path, context.step_key, context.name), "w") as fd:
+            fd.write(obj)
+
+    def load_input(self, context: InputContext):
+        with open(
+            os.path.join(
+                self.base_path, context.upstream_output.step_key, context.upstream_output.name
+            ),
+            "r",
+        ) as fd:
+            return fd.read()
+
+
+@io_manager(config_schema={"base_path": str})
+def old_file_io_manager(context):
+    base_path = context.resource_config["base_path"]
+    return OldFileIOManager(base_path)
+
+
+# New adapter layer
+class MyIOManager(ConfigurableIOManagerAdapter):
+    base_path: str
+
+    @property
+    def wrapped_io_manager(self) -> IOManagerDefinition:
+        return old_file_io_manager
+
+
 defs = Definitions(
-    assets=[my_asset], resources={"writer": WriterResource(prefix="greeting: ")}
+    assets=...,
+    resources={
+        "io_manager": MyIOManager(base_path="/tmp/"),
+    },
 )
 ```
 
--- a/docs/content/guides/dagster/pythonic-resources.mdx
+++ b/docs/content/guides/dagster/pythonic-resources.mdx
@@ -1,0 +1,337 @@
+---
+title: Pythonic resources | Dagster
+description: Pythonic resources are an experimental new feature that makes Dagster Resources easier to use and more powerful.
+---
+
+# Pythonic resources
+
+<Note>
+  This feature is considered <strong>experimental</strong>.
+</Note>
+
+This guide acts as an introduction to Dagster resources utilizing the new Pythonic resources API layer, which makes defining and using Dagster [resources](/concepts/resources) easier.
+
+Resources are objects that are shared across the implementations of multiple [software-defined assets](/concepts/assets/software-defined-assets) and [ops](/concepts/ops-jobs-graphs/ops) and that can be plugged in after defining those ops and assets.
+
+Resources typically model external components that assets and ops interact with. For example, a resource might be a connection to a data warehouse like Snowflake or a service like Slack.
+
+So, why use resources?
+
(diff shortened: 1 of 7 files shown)


benpankow:
* master
  * **PR #12235** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12235" target="_blank">Graphite</a> 
    * **PR #12260** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12260" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12260?utm_source=stack-comment).

schrockn (changes requested):
Can you post this as an entirely new file? I think that is how we will want to develop these docs anyways (in an experimental area). It will also be easier to review.

benpankow:
Starting to break out into its own guide: https://dagster-git-benpankow-update-resources-docs-pyt-ef810b-elementl.vercel.app/guides/dagster/pythonic-resources

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,337 @@
...
+
+Resources typically model external components that assets and ops interact with. For example, a resource might be a connection to a data warehouse like Snowflake or a service like Slack.
+
+So, why use resources?
+
+- **Plug in different implementations in different environments** - If you have a heavy external dependency that you want to use in production, but avoid using in testing, you can accomplish this by providing different resources in each environment. Check out [Separating Business Logic from Environments](/concepts/testing#separating-business-logic-from-environments) for more info about this capability.
+- **Share configuration across multiple ops or assets** - Resources are configurable and shared, so you can supply configuration in one place instead of configuring the ops and assets individually.
+- **Share implementations across multiple ops or assets** - When multiple ops access the same external services, resources provide a standard way to structure your code to share the implementations.
```
schrockn, line 22: We should add a top bullet that speaks to new Dagit features that display resources and their configuration. I think this observability is now the top reason to use them

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,337 @@
...
+
+---
+
+## Using resources
+
+### Getting started using resources
+
+A resource can be any Python object which you would like to share between software-defined assets or ops. Assets and ops specify resource dependencies by annotating the resource as a parameter to the asset or op function. The parameter type should be wrapped in a <PyObject object="Resource"/> annotation to indicate that it is a resource and not an upstream asset or op input.
```
schrockn, line 30: `The parameter type should be wrapped in a <PyObject object="Resource"/> annotation to indicate that it is a resource and not an upstream asset or op input.`

This is going to present a challenge for sure. I consider this an advanced capability but you are right that users will probably encounter it early in their journey. Still I would much prefer to try to introduce the `Resource` annotation later and instead assume they will annotate with their resource definition

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,337 @@
...
+
+To provide resource values to your assets and ops, attach them to your <PyObject object="Definitions" /> call. These resources are automatically passed to the function at runtime.
+
+<TabGroup persistentKey="assetsorops">
+<TabItem name="Using software-defined-assets">
+
+#### Using software-defined assets
+
+Here, we define a simple string resource representing a URL which we use in a software-defined asset. We specify a value for the URL in the <PyObject object="Definitions" /> call.
```
schrockn, line 39: The first example should be something more typically, like a database connection. Whatever it is it should be a class.

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,337 @@
...
+    resources={"data_url": "https://dagster.io"},
+)
+```
+
+</TabItem>
+</TabGroup>
+
+### Configurable resources
+
+In many cases, resources need to be [configurable](/concepts/configuration/config-schema). For example, a resource representing a connection to an external service may need a configurable username and password. Configurable resources can be defined by extending the <PyObject object="ConfigurableResource"/> class. The attributes on the class are used to define the configuration schema, which lets you customize behavior through run configuration.
```
schrockn, line 90: We should answer the question, why "configuration" rather than just plain ole python parameter passing here

docs/content/guides/dagster/pythonic-resources.mdx
schrockn: In fact, I don't think we should encourage the usage of resources like this at all.

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,337 @@
...
+    },
+)
+```
+
+</TabItem>
+</TabGroup>
+
+---
+
+## Using environment variables with resources
```
schrockn, line 182: Note somewhere that we have special features in Cloud to support in-UI env variables.

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,337 @@
...
+        )
+    },
+)
+```
+
+---
+
+## Configuring resources at runtime
+
+In some cases, you may want to specify configuration for a resource at runtime, in the launchpad. For example, you may want to specify a different target table in a database resource for each run.
```
schrockn, line 214: Other critical use case is RunRequest in schedules and sensors

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,337 @@
...
+    assets=...,
+    resources={"db_conn": DatabaseResource.configure_at_launch()},
+)
+```
+
+---
+
+## Resources which depend on other resources
+
+In some situations, you may want to define a resource which depends on other resources. For example, a resource representing a connection to an external service may rely on a resource representing credentials for that service.
```
schrockn, line 241: I think we need to flesh out the motivation here.

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,337 @@
...
+
+credentials = CredentialsResource(username="my_user", password="my_password")
+defs = Definitions(
+    assets=...,
+    resources={
+        "bucket": FileStoreBucket(
+            credentials=credentials,
+            region="us-east-1",
+        ),
+    },
```
schrockn, line 275: We should decide on best practice in terms of whether or not we encourage people to a set a top-level resource for something like credentials here, even if we do support it in this style.

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,337 @@
...
+)
+```
+
+If we instead would like the configuration for our credentials to be provided at runtime, we can use the `configure_at_launch()` method to defer the construction of the `CredentialsResource` until runtime.
+
+Because `credentials` requires runtime configuration through the launchpad, it must also be passed to the `Definitions` object, so that configuration can be provided at runtime. Nested resources only need to be passed to the `Definitions` object if they require runtime configuration.
```
schrockn, line 281: This is good example for motivation behind resource-to-resource specifies: canonicalization of runtime config to a single location.

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,337 @@
...
+class MyResourceRequiresAnother(ConfigurableResource):
+    foo: StringHolderResource
+    bar: str
+
+
+def test_my_resource_with_context():
+    resource = MyResourceRequiresAnother(foo=StringHolderResource("foo"), bar="bar")
+    assert resource.foo.value == "foo"
+    assert resource.bar == "bar"
+```
```
schrockn, line 337: Let's add a section describe typing and how we accept different types at the constructor than the output. E.g. you can pass `EnvVar` in but you always get `str` out.

schrockn (changes requested):
This is a good start. I think most importantly, I've been convinced that `ConfigurableResource` is a good name.

docs/content/guides/dagster/pythonic-resources.mdx
benpankow: moved out of critical path to an 'advanced usage' section

docs/content/guides/dagster/pythonic-resources.mdx
benpankow: moved out of critical path to an 'advanced usage' section

docs/content/guides/dagster/pythonic-resources.mdx
benpankow: added some more explanation

docs/content/guides/dagster/pythonic-resources.mdx
benpankow: Added in an `implementation details` section

benpankow:
Restructured doc a bit and included some more advanced functionality,  draft is ready for a second look

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+
+defs = Definitions(
+    assets=[data_from_service],
+    resources={
+        "my_conn": MyConnectionResource(username="my_user", password="my_password"),
+    },
+)
+```
+
+</TabItem>
```
schrockn, line 75: Here we should either (1) use a different example or (2) note that using a bare string for a password is a very bad idea. I bias towards (1) and then only introduce a password in the environment variable section.

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+
+
+defs = Definitions(
+    assets=...,
+    resources={"db_conn": DatabaseResource.configure_at_launch()},
+)
+```
+
+---
+
```
schrockn, line 179: Recommend showing a RunRequest construction with this config

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+### Resources which depend on other resources
+
+In some situations, you may want to define a resource which depends on other resources. This is useful for common configuration. For example, separate resources for a database and for a filestore may both depend on credentials for a particular cloud provider. Defining these credentials as a separate, nested resource allows you to specify configuration in a single place.
```
schrockn, line 182: Additional benefit is that this can assist in the testability of resources themselves.

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+```python file=/guides/dagster/pythonic_resources/pythonic_resources.py startafter=start_new_resource_testing endbefore=end_new_resource_testing
+from dagster._config.structured_config import ConfigurableResource, ConfigurableResourceAdapter
+
+V = TypeVar("V")
+
+
+class ResourceDependency(Generic[V]):
+    def __get__(self, obj: "Resource", __owner: Any) -> V:
+        return getattr(obj, self._name)
+
```
schrockn, line 255: I don't think we need this in the code example

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+    def load_input(self, context: InputContext):
+        return read_csv(self._get_path(context.asset_key))
+
+defs = Definitions(
+    assets=...,
+    resources={
+        "io_manager": MyIOManager(root_path="/tmp/")
+    },
+)
+```
```
schrockn, line 315: IMO after this section we should introduce the notion of the factory as I think it will be fairly common to do this for I/O managers *and* it is necessary to understand the factory in order to adapt the old-style resources

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+### Using bare Python objects as resources
+
+In some cases, you may want to use a bare Python object as a resource which is not a subclass of <PyObject object="ConfigurableResource"/>.
+
+For example, you may want to directly pass a third-party API client into your assets or ops. This follows a similar pattern to using a <PyObject object="ConfigurableResource"/> subclass, however assets and ops which use these resources must [annotate](https://docs.python.org/3/library/typing.html#typing.Annotated) them with <PyObject object="Resource"/>.
+
+```python file=/guides/dagster/pythonic_resources/pythonic_resources.py startafter=start_raw_github_resource endbefore=end_raw_github_resource
+@asset
+def public_github_repos(github: Resource[GitHub]):
+    return github.organization("dagster-io").repositories()
```
schrockn, line 371: Recommend adding comment here to emphasize how annotations work
e.g.

`# Resource[Github] is treated exactly like Github for type checking purposes and the runtime type of the github parameter is GitHub. The purpose of Resource is to tell dagster that github is a resource rather than an upstream asset.`

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+
+
+defs = Definitions(
+    assets=[public_github_repos],
+    resources={"github": GitHub()},
+)
+```
+
+If you would like a Pythonic resource to depend on a bare Python object, you can use the <PyObject object="ResourceDependency"/> annotation to annotate the attribute in question. This indicates to Dagster that the attribute should be treated as a resource dependency.
```
schrockn, line 380: When reading this in documentation form, it's a bit confusing. 

* What happens if you annotate it with just `Engine`. Is that an error? Why?
* What are the exact semantics of a resource dependency? Does it get an entry in the UI?

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+    username: str
+    password: str
+
+
+class FileStoreBucket(ConfigurableResource):
+    credentials: CredentialsResource
+    region: str
+
+    def write(self, data: str):
+        # In this context, `self.credentials` is ensured to
```
schrockn, line 421: "In this context, `self.credentials` is ensured to be a CredentialsResource with valid values for `username` and `password`"

schrockn (changes requested):
Great start. Left inline comments.

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+- **Plug in different implementations in different environments** - If you have a heavy external dependency that you want to use in production, but avoid using in testing, you can accomplish this by providing different resources in each environment. Check out [Separating Business Logic from Environments](/concepts/testing#separating-business-logic-from-environments) for more info about this capability.
+- **Surface configuration in the UI** - Resources and their configuration are surfaced in the Dagster UI, making it easy to see where your resources are used and how they are configured.
+- **Share configuration across multiple ops or assets** - Resources are configurable and shared, so you can supply configuration in one place instead of configuring the ops and assets individually.
+- **Share implementations across multiple ops or assets** - When multiple ops access the same external services, resources provide a standard way to structure your code to share the implementations.
+
+---
+
+## Using Pythonic resources
+
+### Getting started using resources
```
erinkcochran87, line 29: Nit - is `using resources` in this heading necessary? Given that the heading above it is `Using Pythonic resources`, it feels a bit redundant IMO. `Getting started` could be enough here.

docs/content/guides/dagster/pythonic-resources.mdx
erinkcochran87: +1

As the ops example is essentially the same thing, I assume this comment should extend to that example as well?

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+    resources={
+        "my_conn": MyConnectionResource(username="my_user", password="my_password"),
+    },
+)
+```
+
+</TabItem>
+</TabGroup>
+
+---
```
erinkcochran87, line 121: [1] Rules are only needed before H2 headings, so we can take this guy out

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+    resources={
+        "credentials": CredentialsResource(
+            username=EnvVar("MY_USERNAME"),
+            password=EnvVar("MY_PASSWORD"),
+        )
+    },
+)
+```
+
+---
```
erinkcochran87, line 151: [1]

docs/content/guides/dagster/pythonic-resources.mdx
erinkcochran87: +1

If we want to have two examples here, we could put them into tabs.

Ex: `Title for the basic example` | `Using a RunRequest`

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+
+defs = Definitions(
+    assets=...,
+    resources={"db_conn": DatabaseResource.configure_at_launch()},
+)
+```
+
+---
+
+### Resources which depend on other resources
```
erinkcochran87, line 180:
```suggestion
### Resources that depend on other resources
```

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+    return db_conn.read()
+
+
+defs = Definitions(
+    assets=...,
+    resources={"db_conn": DatabaseResource.configure_at_launch()},
+)
+```
+
+---
```
erinkcochran87, line 178: [1]

docs/content/guides/dagster/pythonic-resources.mdx
erinkcochran87, line 182:
```suggestion
In some situations, you may want to define a resource that depends on other resources. This is useful for common configuration. For example, separate resources for a database and for a filestore may both depend on credentials for a particular cloud provider. Defining these credentials as a separate, nested resource allows you to specify configuration in a single place.
```

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+
+### Resources which depend on other resources
+
+In some situations, you may want to define a resource which depends on other resources. This is useful for common configuration. For example, separate resources for a database and for a filestore may both depend on credentials for a particular cloud provider. Defining these credentials as a separate, nested resource allows you to specify configuration in a single place.
+
+In this case, you can list that nested resource as an attribute of your resource class.
```
erinkcochran87, line 184:
```suggestion
In this case, you can list the nested resource as an attribute of your resource class.
```

This feels like a nit (sorry!) but using `that` here makes me think we're referring to a specific nested resource when we're actually speaking about it in general

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+from dagster._config.structured_config import ConfigurableResource
+
+
+class CredentialsResource(ConfigurableResource):
+    username: str
+    password: str
+
+
+class FileStoreBucket(ConfigurableResource):
+    credentials: CredentialsResource
```
erinkcochran87, line 197: Thoughts on adding a code comment here to call this out? (cc @schrockn, if you have thoughts on this approach)

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+    resources={
+        "bucket": FileStoreBucket(
+            credentials=credentials,
+            region="us-east-1",
+        ),
+    },
+)
+```
+
+If we instead would like the configuration for our credentials to be provided at runtime, we can use the `configure_at_launch()` method to defer the construction of the `CredentialsResource` until runtime.
```
erinkcochran87, line 220:
```suggestion
If instead we want the configuration for our credentials to be provided at runtime, we can use the `configure_at_launch()` method to defer the construction of the `CredentialsResource` until runtime.
```

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+
+Because `credentials` requires runtime configuration through the launchpad, it must also be passed to the `Definitions` object, so that configuration can be provided at runtime. Nested resources only need to be passed to the `Definitions` object if they require runtime configuration.
```
erinkcochran87, line 222: I love that this is so straightforward!

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+        "credentials": credentials,
+        "bucket": FileStoreBucket(
+            credentials=credentials,
+            region="us-east-1",
+        ),
+    },
+)
+```
+
+---
```
erinkcochran87, line 240: [1]

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+def test_my_resource_with_context():
+    string_holder = StringHolderResource(value="foo")
+    resource = MyResourceRequiresAnother(foo=string_holder, bar="bar")
+    assert resource.foo.value == "foo"
+    assert resource.bar == "bar"
+```
+
+---
+
+## Defining Pythonic IO managers
```
erinkcochran87, line 292:
```suggestion
## Defining Pythonic I/O managers
```

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+
+Pythonic [IO managers](/concepts/io-management/io-managers) are defined as subclasses of <PyObject object="ConfigurableIOManager"/>, and similarly to Pythonic resources specify any configuration fields as attributes. Each subclass must implement a `handle_output` and `load_input` method, which are called by the Dagster runtime to handle the storing and loading of data.
```
erinkcochran87, line 294:
```suggestion
Pythonic [I/O managers](/concepts/io-management/io-managers) are defined as subclasses of <PyObject object="ConfigurableIOManager"/>, and similarly to Pythonic resources specify any configuration fields as attributes. Each subclass must implement a `handle_output` and `load_input` method, which are called by the Dagster runtime to handle the storing and loading of data.
```

Also:

> ... by the Dagster runtime to handle the storing and loading of data.

Is `the Dagster runtime` a thing? (I'm not trying to sound snarky, I honestly don't know - it read a bit like a typo)

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+
+## Advanced resource patterns
+
+The following sections describe more advanced patterns for using Pythonic resources.
+
+### Adapting function-style resources
+
+In codebases which utilize [function-style](https://docs.dagster.io/concepts/resources#resources) resources with the <PyObject object="@resource"/> decorator, you can use the <PyObject object="ConfigurableResourceAdapter"/> class to adapt these resources to the new Pythonic resource pattern.
+
+The adapter subclass should list all config fields as attributes, and should implement a `wrapped_resource` property which returns the underlying resource definition to adapt.
```
erinkcochran87, line 327:
```suggestion
The adapter subclass should list all config fields as attributes and implement a `wrapped_resource` property which returns the underlying resource definition to adapt.
```

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,440 @@
...
+
+---
+
+## Advanced resource patterns
+
+The following sections describe more advanced patterns for using Pythonic resources.
+
+### Adapting function-style resources
+
+In codebases which utilize [function-style](https://docs.dagster.io/concepts/resources#resources) resources with the <PyObject object="@resource"/> decorator, you can use the <PyObject object="ConfigurableResourceAdapter"/> class to adapt these resources to the new Pythonic resource pattern.
```
erinkcochran87, line 325:
```suggestion
In codebases that utilize [function-style](https://docs.dagster.io/concepts/resources#resources) resources with the <PyObject object="@resource"/> decorator, you can use the <PyObject object="ConfigurableResourceAdapter"/> class to adapt these resources to the new Pythonic resource pattern.
```

erinkcochran87 (commented):
This is looking great Ben! 🎉 

One thing I noticed is that some examples include imports while others don't. What's the thinking behind including them vs not? I think users generally find examples to be more helpful if they're immediately copy-and-pasteable, but I don't want to recommend doing that all the time without more info.

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -379,8 +375,69 @@ def my_asset(writer: Writer):
...
+    InputContext,
+    OutputContext,
+)
+from dagster._config.structured_config import (
+    StructuredIOManagerAdapter as ConfigurableIOManagerAdapter,
+)
+import os
+
+
+# Old code, cannot be changed for back-compat purposes
```
schrockn, line 402: I think one modification which should make here is that that we can't change the public interface for the code, rather than not change it at all.

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -379,8 +375,69 @@ def my_asset(writer: Writer):
...
+        with open(
+            os.path.join(
+                self.base_path, context.upstream_output.step_key, context.upstream_output.name
+            ),
+            "r",
+        ) as fd:
+            return fd.read()
+
+
+@io_manager(config_schema={"base_path": str})
```
schrockn, line 421: If we are allow to change code but keep interface, we can develop a pattern here where we can canonicalize the config schema. We define the config schema in the Pythonic style only and programmatically generate the old school schema. I think this is the pattern we should use when we start to do integration conversion. I don't think we need to block this guide on that but you should put it on your radar.

schrockn (approved):

docs/content/guides/dagster/pythonic-resources.mdx
```diff
@@ -0,0 +1,538 @@
...
+    tables = ...
+    for table_name in tables:
+        yield RunRequest(
+            run_config=RunConfig(
+                resources={
+                    "db_conn": DatabaseResource(table=table_name),
+                },
+            ),
+        )
+```
```
schrockn, line 197: Wanted to flag this this will potentially be confusing because the behavior here will be slightly different than passing resources to `Definitions`, in particular when raw resource objects are in play. I anticipate that people will copy and paste resource dictionaries around alternatively pass them to Definitions and this. There will be slight differences in behavior.

docs/content/guides/dagster/pythonic-resources.mdx
schrockn: To call what out? I might be missing some context here. In general a big fan in comments in code samples as I think a lot of readers skip prose and just look at examples.

schrockn (approved):
This is definitely read to ship. I think we can continue to iterate on the experimental guide as we get early usage.

Issue #12006, opened 2023-01-31
Allow specifying `timezone` on `FreshnessPolicy`


plaflamme:
### What's the use case?

We have a provider that makes a dataset available every day at 2AM **local time**. I’d like to make a `FreshnessPolicy` with `maximum_lag_minutes=24*60` and a `cron_schedule=0 2 * * *`  but the cron schedule is in UTC, so I have to adjust, e.g.: `cron_schedule=0 7 * * *`  but because of daylight savings in the local timezone, I actually have 2 different “UTC hour” to provide, (e.g.: 7 and 6 or similar) which is not possible (without restarting the server).

### Ideas of implementation

Add an `execution_timezone` attribute, similar to `ScheduleDefinition`.

### Additional information

Original slack thread: https://dagster.slack.com/archives/C01U954MEER/p1675119210899799

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

yuhan:
cc @OwenKephart / @sryza

PR #12263, opened 2023-02-10
[freshness-policies] add a cron_schedule_timezone argument to the FreshnessPolicy class

OwenKephart:
### Summary & Motivation

Resolves: https://github.com/dagster-io/dagster/issues/12006


sryza (commented):
Does it default to UTC? If so, worth documenting

sryza (approved):

PR #12264, opened 2023-02-10
in docs navigation, make partitions independent from schedules

sryza:
### Summary & Motivation

The fact that we bucket partitions with schedules makes users think that we discourage use of non-time-window partitions.

### DIFF
--- a/docs/content/_navigation.json
+++ b/docs/content/_navigation.json
@@ -144,7 +144,8 @@
         ]
       },
       {
-        "title": "Schedules, Sensors, & Partitions",
+        "title": "Schedules & Sensors",
+        "path": "/concepts/partitions-schedules-sensors/schedules",
         "children": [
           {
             "title": "Schedules",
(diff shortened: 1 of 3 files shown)


docs/content/_navigation.json
```diff
@@ -144,7 +144,8 @@
         ]
       },
       {
-        "title": "Schedules, Sensors, & Partitions",
+        "title": "Schedules & Sensors",
+        "path": "/concepts/partitions-schedules-sensors/schedules",
```
tacastillo, line 148: We should probably also split the directories in either your upstream PR or this PR, probs the other one to minimize conflicts.

tacastillo (approved):
I ❤️ this change.

I'd probably wait for Erin before merging since she owns information architecture.

docs/content/_navigation.json
erinkcochran87: I agree we need to do this eventually, but I recommend leaving it alone for now. Moving directories is a pretty involved process at the moment. The IA of the docs is likely to change fairly drastically in the next few months as we release the tutorial and new guides. I'm don't know that many people would clock this if we leave it, but we can re-assess if it does cause issues.

erinkcochran87 (approved):

docs/content/_navigation.json
sryza: 👍

PR #12266, opened 2023-02-10
DagsterLibraryRegistry

alangenfeld:
Add a means to track which dagster libraries have been loaded in a process. This can be used for example to show which libraries and their versions were used when loading definitions for a specific code location. 

added test, can add more

```
(py39) ~/dagster:al/02-10-DagsterLibraryRegistry$ ls -l python_modules/libraries/ | grep -v total | wc -l
      43
(py39) ~/dagster:al/02-10-DagsterLibraryRegistry$ grep "DagsterLibraryRegistry.register" python_modules/libraries/ -r | wc -l
      43
```

remove a register call and ensure new test fails

### DIFF
--- a/python_modules/automation/automation_tests/test_repo.py
+++ b/python_modules/automation/automation_tests/test_repo.py
@@ -1,0 +1,17 @@
+import os
+import subprocess
+from pathlib import Path
+
+
+def test_all_libraries_register() -> None:
+    # attempt to ensure all libraries in the repository register with DagsterLibraryRegistry
+    register_call = "DagsterLibraryRegistry.register"
+
+    library_dir = Path(__file__).parents[2] / "libraries"
+    assert str(library_dir).endswith("python_modules/libraries")
+
+    for library in os.listdir(library_dir):
+        result = subprocess.run(["grep", register_call, (library_dir / library), "-r"])
+        assert (
+            result.returncode == 0
+        ), f"Dagster library {library} is missing call to {register_call}."
--- a/python_modules/dagit/dagit/__init__.py
+++ b/python_modules/dagit/dagit/__init__.py
@@ -1,5 +1,5 @@
-from dagster._core.utils import check_dagster_package_version
+from dagster._core.libraries import DagsterLibraryRegistry
 
 from .version import __version__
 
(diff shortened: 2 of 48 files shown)


alangenfeld:
* master
  * **PR #12266** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12266" target="_blank">Graphite</a>  👈
    * **PR #12267** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12267" target="_blank">Graphite</a> 
      * **PR #12268** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12268" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12266?utm_source=stack-comment).

schrockn (dismissed):
Oh this is exciting. We can also add telemetry on libraries this way (cc: @petehunt) which I think would be great.

Is there a way we can add a test that would catch when someone adds a library without calling `DagsterLibraryRegistry.register`?

python_modules/automation/automation_tests/test_repo.py
```diff
@@ -0,0 +1,16 @@
...
+    # attempt to ensure all libraries in the repository register with DagsterLibraryRegistry
+    register_call = "DagsterLibraryRegistry.register"
+
+    library_dir = Path(__file__).parents[2] / "libraries"
+    assert str(library_dir).endswith("dagster/python_modules/libraries")
+
+    for library in os.listdir(library_dir):
+        result = subprocess.run(["grep", register_call, (library_dir / library), "-r"])
+        if result.returncode != 0:
+            assert False, f"Dagster library {library} is missing call to {register_call}."
```
alangenfeld, line 16: have a better idea than this? I put it in automation instead of dagster since its a meta repo level test but can move it where ever

schrockn: nope I must have missed that

schrockn (approved):

python_modules/automation/automation_tests/test_repo.py
alangenfeld: to be clear I added this in response to your review - was a good call

PR #12269, opened 2023-02-10
fix: only inspect schema when we may create tables

plaflamme:
### Summary & Motivation

Querying the database schema is only necessary when we're configured to possibly create missing tables. Given that this occurs in 3 different places when starting a dagster process, it becomes quite expensive.

`python -m pytest python_modules/libraries/dagster-postgres/dagster_postgres_tests` not sure if there's a better approach

### DIFF
--- a/python_modules/libraries/dagster-postgres/dagster_postgres/event_log/event_log.py
+++ b/python_modules/libraries/dagster-postgres/dagster_postgres/event_log/event_log.py
@@ -84,14 +84,14 @@ class PostgresEventLogStorage(SqlEventLogStorage, ConfigurableClass):
 
         self._secondary_index_cache = {}
 
-        table_names = retry_pg_connection_fn(lambda: db.inspect(self._engine).get_table_names())
-
         # Stamp and create tables if the main table does not exist (we can't check alembic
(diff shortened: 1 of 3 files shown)


alangenfeld:
thanks!

plaflamme:
@alangenfeld Thanks for merging this.

As a follow-up, I'd like to show an annotated screenshot of the "Query Insights" view of my Postgres instance in GCP. This view is used to analyze query plans and performance in general.

Screenshot from 2023-02-25 12-43-01

This screenshot shows the top of the list, ordered by descending frequency. There are many more queries below this threshold, most (all?) of which are "useful" application queries.

Even with this patch applied, the number of queries to extract schema information is massive compared to "useful work" queries hitting the database. These numbers are only a few minutes (in the tens of minutes) after starting an instance and having a few concurrent runs.

I believe this is caused by the lack of application-level connection pooling reported [here](https://github.com/dagster-io/dagster/issues/8466). Adding an external pool helps, but having an application-level pool would probably allow it to cache these schema-inspection queries (maybe?).

In any case, I think these numbers speak for themselves: it doesn't seem normal for schema inspection queries to be more frequent than the vast majority of "useful" application queries.

PR #12283, opened 2023-02-12
[typing/static] storage

smackesey:
### Summary & Motivation

Type annotations for storage layer. Annotations for many of the files in `dagster._core.storage`, `dagster-mysql` and `dagster-postgres`.

This is 99% a "pure typing PR". The only other change is some import changes (e.g. direct import `sqlalchemy.pool` and access as `db_pool` to avoid unloaded module access errors). **There are no changes to runtime type-checks**, which should make it much easier to review.

Added many type-ignores to suppress errors generated by the annotations. Fixed upstack.

BK

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/stats.py
+++ b/python_modules/dagster/dagster/_core/execution/stats.py
@@ -1,5 +1,3 @@
-# pyright: strict reportUnnecessaryIsInstance=false
-
 from collections import defaultdict
 from enum import Enum
 from typing import Any, Dict, Iterable, NamedTuple, Optional, Sequence, cast
--- a/python_modules/dagster/dagster/_core/execution/stats.py
+++ b/python_modules/dagster/dagster/_core/execution/stats.py
@@ -1,3 +1,5 @@
+# pyright: strict reportUnnecessaryIsInstance=false
+
 from collections import defaultdict
 from enum import Enum
 from typing import Any, Dict, Iterable, NamedTuple, Optional, Sequence, cast
@@ -11,7 +13,9 @@ from dagster._serdes import whitelist_for_serdes
 from dagster._utils import datetime_as_float
 
 
-def build_run_stats_from_events(run_id, records):
+def build_run_stats_from_events(
+    run_id: str, records: Iterable[EventLogEntry]
+) -> PipelineRunStatsSnapshot:
     try:
         iter(records)
     except TypeError as exc:
--- a/python_modules/dagster/dagster/_core/log_manager.py
+++ b/python_modules/dagster/dagster/_core/log_manager.py
@@ -24,7 +24,14 @@ class IDagsterMeta(Protocol):
         ...
 
 
-# We type-ignore because this is a stub class, we only ever cast log records to this.
+# The type-checker complains here that DagsterLogRecord does not implement the `dagster_meta`
+# property of `IDagsterMeta`. We ignore this error because we don't need to implement this method--
+# `DagsterLogRecord` is a stub class that is never instantiated. We only ever cast
+# `logging.LogRecord` objects to `DagsterLogRecord`, because it gives us typed access to the
+# `dagster_meta` property. `dagster_meta` itself is set on these `logging.LogRecord` objects via the
+# `extra` argument to `logging.Logger.log` (see `DagsterLogManager.log_dagster_event`), but
+# `logging.LogRecord` has no way of exposing to the type-checker the attributes that are dynamically
+# defined via `extra`.
 class DagsterLogRecord(logging.LogRecord, IDagsterMeta):  # type: ignore
     pass
 
--- a/python_modules/dagster/dagster/_core/log_manager.py
+++ b/python_modules/dagster/dagster/_core/log_manager.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 
 import datetime
 import logging
-from typing import TYPE_CHECKING, Any, Mapping, NamedTuple, Optional, Sequence, Union
+from typing import TYPE_CHECKING, Any, Mapping, NamedTuple, Optional, Sequence, Union, cast
+
+from typing_extensions import Protocol
 
 import dagster._check as check
 from dagster._core.utils import coerce_valid_log_level, make_new_run_id
@@ -14,6 +16,17 @@ if TYPE_CHECKING:
     from dagster._legacy import DagsterRun
 
 DAGSTER_META_KEY = "dagster_meta"
+
+
+class IDagsterMeta(Protocol):
+    @property
+    def dagster_meta(self) -> DagsterLoggingMetadata:
+        ...
+
+
+# We type-ignore because this is a stub class, we only ever cast log records to this.
+class DagsterLogRecord(logging.LogRecord, IDagsterMeta):  # type: ignore
+    pass
 
 
 class DagsterMessageProps(
@@ -123,7 +136,7 @@ class DagsterLoggingMetadata(
         )
 
     @property
-    def log_source(self):
+    def log_source(self) -> str:
         if self.resource_name is None:
             return self.pipeline_name or "system"
         return f"resource:{self.resource_name}"
@@ -160,7 +173,7 @@ def construct_log_string(
 
 def get_dagster_meta_dict(
     logging_metadata: DagsterLoggingMetadata, dagster_message_props: DagsterMessageProps
-) -> Mapping[str, Any]:
+) -> Mapping[str, object]:
     # combine all dagster meta information into a single dictionary
     meta_dict = {
         **logging_metadata._asdict(),
@@ -197,7 +210,7 @@ class DagsterLogHandler(logging.Handler):
         super().__init__()
 
     @property
-    def logging_metadata(self):
+    def logging_metadata(self) -> DagsterLoggingMetadata:
         return self._logging_metadata
 
     def with_tags(self, **new_tags: str) -> DagsterLogHandler:
@@ -219,7 +232,7 @@ class DagsterLogHandler(logging.Handler):
         ]
         return {k: v for k, v in record.__dict__.items() if k not in ref_attrs}
 
-    def _convert_record(self, record: logging.LogRecord) -> logging.LogRecord:
+    def _convert_record(self, record: logging.LogRecord) -> DagsterLogRecord:
         # we store the originating DagsterEvent in the DAGSTER_META_KEY field, if applicable
         dagster_meta = getattr(record, DAGSTER_META_KEY, None)
 
@@ -239,7 +252,8 @@ class DagsterLogHandler(logging.Handler):
         record.msg = construct_log_string(self._logging_metadata, dagster_message_props)
         record.args = ()
 
-        return record
+        # DagsterLogRecord is a LogRecord with a `dagster_meta` field
+        return cast(DagsterLogRecord, record)
 
     def filter(self, record: logging.LogRecord) -> bool:
         """If you list multiple levels of a python logging hierarchy as managed loggers, and do not
@@ -251,7 +265,7 @@ class DagsterLogHandler(logging.Handler):
             getattr(record, DAGSTER_META_KEY, None), dict
         )
 
-    def emit(self, record: logging.LogRecord):
+    def emit(self, record: logging.LogRecord) -> None:
         """For any received record, add Dagster metadata, and have handlers handle it."""
         try:
             # to prevent the potential for infinite loops in which a handler produces log messages
--- a/python_modules/dagster/dagster/_core/storage/alembic/env.py
+++ b/python_modules/dagster/dagster/_core/storage/alembic/env.py
@@ -12,6 +12,6 @@ config = context.config
 target_metadata = [SqlEventLogStorageMetadata, SqlRunStorage, SqlScheduleStorage]
 
 if context.is_offline_mode():
-    run_migrations_offline(context, config, target_metadata)
+    run_migrations_offline(context, config, target_metadata)  # type: ignore  # (bad stubs)
 else:
-    run_migrations_online(context, config, target_metadata)
+    run_migrations_online(context, config, target_metadata)  # type: ignore  # (bad stubs)
--- a/python_modules/dagster/dagster/_core/storage/captured_log_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/captured_log_manager.py
@@ -1,5 +1,3 @@
-# pyright: strict
-
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
--- a/python_modules/dagster/dagster/_core/storage/captured_log_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/captured_log_manager.py
@@ -1,14 +1,16 @@
-# pylint: disable=unused-argument
+# pyright: strict
 
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from contextlib import contextmanager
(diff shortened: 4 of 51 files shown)


smackesey:
* master
  * **PR #12283** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12283" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12283?utm_source=stack-comment).

python_modules/dagster/dagster/_core/execution/stats.py
```diff
@@ -1,3 +1,5 @@
+# pyright: strict reportUnnecessaryIsInstance=false
```
schrockn, line 1: what's going on here

python_modules/dagster/dagster/_core/log_manager.py
```diff
@@ -16,6 +18,17 @@
 DAGSTER_META_KEY = "dagster_meta"
 
 
+class IDagsterMeta(Protocol):
+    @property
+    def dagster_meta(self) -> DagsterLoggingMetadata:
+        ...
+
+
+# We type-ignore because this is a stub class, we only ever cast log records to this.
```
schrockn, line 27: not grokking this comment

python_modules/dagster/dagster/_core/storage/alembic/env.py
```diff
@@ -12,6 +12,6 @@
 target_metadata = [SqlEventLogStorageMetadata, SqlRunStorage, SqlScheduleStorage]
 
 if context.is_offline_mode():
-    run_migrations_offline(context, config, target_metadata)
+    run_migrations_offline(context, config, target_metadata)  # type: ignore  # (bad stubs)
```
schrockn, line 15: not grokking comment

python_modules/dagster/dagster/_core/storage/captured_log_manager.py
```diff
@@ -127,12 +129,12 @@ def fetch(self):
...
 
-    def complete(self):
+    def complete(self) -> None:
         self.is_complete = True
 
 
-def _has_max_data(chunk) -> bool:
-    return chunk and len(chunk) >= MAX_BYTES_CHUNK_READ
+def _has_max_data(chunk: Optional[bytes]) -> bool:
+    return chunk and len(chunk) >= MAX_BYTES_CHUNK_READ  # type: ignore
```
schrockn, line 137: why `type: ignore`? Seems like it should work

python_modules/dagster/dagster/_core/storage/compute_log_manager.py
```diff
@@ -200,13 +211,13 @@ def observable(self, run_id, key, io_type, cursor=None) -> "ComputeLogSubscripti
...
 
         if cursor:
-            cursor = int(cursor)
+            cursor = int(cursor)  # type: ignore
         else:
-            cursor = 0
+            cursor = 0  # type: ignore
 
-        subscription = ComputeLogSubscription(self, run_id, key, io_type, cursor)
+        subscription = ComputeLogSubscription(self, run_id, key, io_type, cursor)  # type: ignore
```
schrockn, line 218: brief comments about `type: ignore` plz

python_modules/dagster/dagster/_core/storage/config.py
```diff
@@ -1,7 +1,27 @@
+# pyright: strict
```
schrockn, line 1: what is our criteria for turning on strict in general? Should we be pushing everyone to do this?

python_modules/dagster/dagster/_core/storage/event_log/sqlite/sqlite_event_log.py
```diff
@@ -405,29 +415,36 @@ def watch(self, run_id, cursor, callback):
             self._obs.schedule(watchdog, self._base_dir, True),
         )
 
-    def end_watch(self, run_id, handler):
+    def end_watch(self, run_id: str, handler: EventHandlerFn) -> None:
         if handler in self._watchers[run_id]:
             event_handler, watch = self._watchers[run_id][handler]
-            self._obs.remove_handler_for_watch(event_handler, watch)
+            self._obs.remove_handler_for_watch(event_handler, watch)  # type: ignore
```
schrockn, line 421: comment plz

python_modules/dagster/dagster/_core/storage/file_manager.py
```diff
@@ -221,62 +223,64 @@ def files_pipeline():
...
-    def read(self, file_handle, mode="rb"):
+    def read(self, file_handle: LocalFileHandle, mode: str = "rb") -> Iterator[IOStream]:
         check.inst_param(file_handle, "file_handle", LocalFileHandle)
         check.str_param(mode, "mode")
         check.param_invariant(mode in {"r", "rb"}, "mode")
 
         encoding = None if mode == "rb" else "utf8"
         with open(file_handle.path, mode, encoding=encoding) as file_obj:
-            yield file_obj
+            yield file_obj  # type: ignore
```
schrockn, line 271: comment plz. for this file. I assume these are all related to the `self.read` calls?

python_modules/dagster/dagster/_core/storage/fs_io_manager.py
```diff
@@ -216,11 +217,11 @@ def handle_output(self, context, obj):
...
         )
 
-    def load_input(self, context):
+    def load_input(self, context: InputContext) -> object:
         """Unpickle the file from a given file path and Load it to a data object."""
         check.inst_param(context, "context", InputContext)
-        metadata = context.upstream_output.metadata
-        path = check.str_param(metadata.get("path"), "metadata.path")
+        metadata = context.upstream_output.metadata  # type: ignore
+        path = check.str_param(metadata.get("path"), "metadata.path")  # type: ignore
```
schrockn, line 224: explain in comments plz

python_modules/dagster/dagster/_serdes/config_class.py
```diff
@@ -3,13 +3,14 @@
...
 import dagster._check as check
+from dagster._config.config_schema import UserConfigSchema
 from dagster._utils import convert_dagster_submodule_name
 from dagster._utils.yaml_utils import load_run_config_yaml
 
 from .serdes import DefaultNamedTupleSerializer, WhitelistMap, whitelist_for_serdes
 
 if TYPE_CHECKING:
-    from dagster._config.config_type import ConfigType
+    pass
```
schrockn, line 13: rm?

python_modules/libraries/dagster-mysql/dagster_mysql/event_log/event_log.py
```diff
@@ -156,64 +168,64 @@ def store_asset_event(self, event, event_id):
             else:
                 try:
                     conn.execute(
-                        db.dialects.mysql.insert(AssetKeyTable).values(
-                            asset_key=event.dagster_event.asset_key.to_string(),
+                        db_dialects.mysql.insert(AssetKeyTable).values(
+                            asset_key=event.dagster_event.asset_key.to_string(),  # type: ignore
```
schrockn, line 172: explain in comment plz

python_modules/libraries/dagster-mysql/dagster_mysql/schedule_storage/schedule_storage.py
```diff
@@ -83,59 +90,61 @@ def optimize_for_dagit(self, statement_timeout, pool_recycle):
...
         if not rows:
             return None
 
         return rows[0][0]
 
-    def upgrade(self):
+    def upgrade(self) -> None:
         alembic_config = mysql_alembic_config(__file__)
-        run_alembic_upgrade(alembic_config, self._engine)
+        run_alembic_upgrade(alembic_config, self._engine)  # type: ignore  # shouldn't this be a connection
```
schrockn, line 142: who is this comment directed to?

schrockn (changes requested):
Looks good overall. I have a few questions inline. Overall commentary:

1) By default every `type: ignore` should have a comment. You have spent time figuring out what is going on and have a reason why you have to opt out of the type system in a particular. It's important to get that knowledge out of your head and into the codebase as a comment.
2) Seems like there are few places where you do repeated type ignores where you could consolidate to a single helper function with a single type ignore.  The `self.read` example comes to mind. Try to consolidate type: ignores where you can.
3) I'd like to know about your proposed policy around use of strict before start checking them in.

python_modules/dagster/dagster/_core/execution/stats.py
smackesey: `reportUnnecessaryIsIstance` is a pyright rule that is part of the `strict` ruleset. It flags when you perform an `isinstance` check that doesn't actually do any narrowing, because the type is already known by pyright.

To remove those unnecessary `isinstance` errors here requires runtime changes, so it's not done in this PR, but the file otherwise passes strict mode-- IMO this approach is more informative about the status of the file than leaving in type-ignores with or without a `# pyright: strict`.

python_modules/dagster/dagster/_core/log_manager.py
smackesey: updated comment

python_modules/dagster/dagster/_core/storage/alembic/env.py
smackesey: See my longer response to this review, `(bad stubs)` is an idiom used throughout codebase. Open to changing it.

python_modules/dagster/dagster/_core/storage/captured_log_manager.py
smackesey: Function is used as a predicate but does not actually return a `bool` (if `chunk` is truthy). Will be fixed in a `[typing/runtime]` upstack.

python_modules/dagster/dagster/_core/storage/config.py
smackesey: See my longer response to review.

python_modules/dagster/dagster/_core/storage/compute_log_manager.py
smackesey: ok

python_modules/dagster/dagster/_core/storage/fs_io_manager.py
smackesey: ok

python_modules/libraries/dagster-mysql/dagster_mysql/schedule_storage/schedule_storage.py
smackesey: Anyone who tries to fix the type error here-- reworded slightly.

smackesey:
>By default every type: ignore should have a comment. You have spent time figuring out what is going on and have a reason why you have to opt out of the type system in a particular. 

I can add more comments. In particular a huge number of type errors are caused by potential null references-- I can annotate those with `(possible none)`.

However, some of the time I _haven't_ figured out what is going on. There are many cases where I'm confident I've annotated a function correctly, but there are still type errors inside the function, usually resulting from other function calls. These type errors are ignored pre-annotation by CI pyright due to `analyzeUnannotatedFunctions=off` (and previously by mypy), but post-annotation they need to be fixed or ignored. It is better to get the annotations in place quickly than to do the detective work to figure out what's going on when we're not changing runtime behavior. We could adopt a standard `# type: ignore  # (??)` for unknown cases if you think that would be better than bare `# type: ignore`.

I have been putting comments in some cases when I have a clear idea of the issue, and there are a couple of idioms for common cases-- for instance, `(bad stubs)` means "the type info provided in this third-party lib or its stub package is likely wrong, causing this type error".

>Seems like there are few places where you do repeated type ignores where you could consolidate to a single helper function with a single type ignore. The self.read example comes to mind. Try to consolidate type: ignores where you can.

In this PR (and other `[typing/static]` PRs) I'd like to avoid introducing runtime changes like helper functions-- those can be introduced in an upstack `[typing/runtime]` PR, but more likely I'd just fix the underlying issue so the type-ignores can go away.

>I'd like to know about your proposed policy around use of strict before start checking them in.

My proposed policy is just: "if a file passes strict, put a `# pyright: strict` comment on it.

I usually add the comment when I start typing work in a file, since it will flag all type issues with editor diagnostics, allowing you to just cycle through the diagnostics in the file and fix stuff efficiently. If there is no sticking point I can't fix, I just leave it in-- IMO this is a good practice to prevent regression in typing quality.

schrockn:
> However, some of the time I haven't figured out what is going on. There are many cases where I'm confident I've annotated a function correctly, but there are still type errors inside the function, usually resulting from other function calls. These type errors are ignored pre-annotation by CI pyright due to analyzeUnannotatedFunctions=off (and previously by mypy), but post-annotation they need to be fixed or ignored. It is better to get the annotations in place quickly than to do the detective work to figure out what's going on when we're not changing runtime behavior. We could adopt a standard # type: ignore  # (??) for unknown cases if you think that would be better than bare # type: ignore.

Ok got it. I think the standard we want to hit here is an accurate reflection of your understanding. If the answer is "I don't know why this is necessary, but we didn't know before, and this just explicitly says that we don't know" I think that is fine. I just don't want to lose a bunch of knowledge and I had no way of knowing how much you knew.

schrockn:
> My proposed policy is just: "if a file passes strict, put a # pyright: strict comment on it. I usually add the comment when I start typing work in a file, since it will flag all type issues with editor diagnostics, allowing you to just cycle through the diagnostics in the file and fix stuff efficiently. If there is no sticking point I can't fix, I just leave it in-- IMO this is a good practice to prevent regression in typing quality.

We just need to get alignment with everyone that that is the goal. You just randomly splatting strict directives in the codebase without communicating anything is not that productive. 

We should do a post somewhere that says something like. "The ideal endstate of the codebase is the entire codebase passing strict mode. This is why. Are we going to do that en masse right now? No. In the meantime, this is what we are going to do"

Otherwise this is just another instance of Sean as a one-man army versus bringing in the whole team collaboratively.

python_modules/dagster/dagster/_core/storage/schedules/sql_schedule_storage.py
```diff
@@ -211,9 +225,16 @@ def _jobs_has_selector_state(self, conn, selector_id):
         result = conn.execute(query)
         row = result.fetchone()
         result.close()
-        return row[0] > 0
+        return row[0] > 0  # type: ignore
```
schrockn, line 228: how does this not return a bool?

python_modules/dagster/dagster/_core/storage/schedules/sql_schedule_storage.py
```diff
@@ -323,10 +352,10 @@ def get_ticks(self, origin_id, selector_id, before=None, after=None, limit=None,
 
         rows = self.execute(query)
         return list(
-            map(lambda r: InstigatorTick(r[0], deserialize_json_to_dagster_namedtuple(r[1])), rows)
+            map(lambda r: InstigatorTick(r[0], deserialize_json_to_dagster_namedtuple(r[1])), rows)  # type: ignore
```
schrockn, line 355: similarly confused at this one

schrockn (approved):
image

python_modules/dagster/dagster/_core/storage/schedules/sql_schedule_storage.py
smackesey: the issue there is that `row` is a possible none-- updated commetn

PR #12004, opened 2023-01-31
[typing/runtime] Refactor cached_method

smackesey:
### Summary & Motivation

Light refactor `cached_method` to fix type error.

### How I Tested These Changes

BK

PR #12003, opened 2023-01-31
[typing/static] assets

smackesey:
### Summary & Motivation

Pure type annotations diff for `AssetGraph` and friends.

### How I Tested These Changes

BK

PR #12284, opened 2023-02-12
[typing/static] PartitionsDefinition covariant type var

smackesey:
### Summary & Motivation

Improves type annotations of partition code with generics and makes the `TypeVar` used for partitions covariant. The lack of covariance was causing mypy errors for users in some contexts: https://dagster.slack.com/archives/C01U954MEER/p1677791924838639, and it makes sense given the immutability of a partition set.

I don't think covariance makes sense for dynamic partitions due to mutability, but currently `DynamicPartitionsDefinition` is not generic anyway (even though it inherits from `PartitionDefinition`, it binds `Any` to the typevar by default).

BK.

Before this PR, the `partitions_def` line in this snippet gives a type error, but not after this PR:

```
from dagster import AssetOut, MonthlyPartitionsDefinition, multi_asset

@multi_asset(
    partitions_def=MonthlyPartitionsDefinition(start_date='2022-01-01'),
    outs={'test': AssetOut()},
)
def test_multi_asset():
    pass
```

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/partition.py
+++ b/python_modules/dagster/dagster/_core/definitions/partition.py
@@ -70,20 +70,20 @@ from .utils import check_valid_name, validate_tags
 
 DEFAULT_DATE_FORMAT = "%Y-%m-%d"
 
-T = TypeVar("T")
+T_cov = TypeVar("T_cov", covariant=True)
 
 
 RawPartitionFunction: TypeAlias = Union[
-    Callable[[Optional[datetime]], Sequence[Union[str, "Partition[T]"]]],
-    Callable[[], Sequence[Union[str, "Partition[T]"]]],
+    Callable[[Optional[datetime]], Sequence[Union[str, "Partition[T_cov]"]]],
+    Callable[[], Sequence[Union[str, "Partition[T_cov]"]]],
 ]
 
 PartitionFunction: TypeAlias = Callable[[Optional[datetime]], Sequence["Partition[Any]"]]
 PartitionTagsFunction: TypeAlias = Callable[["Partition[object]"], Mapping[str, str]]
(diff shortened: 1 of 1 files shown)


smackesey:
* master
  * **PR #12003** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12003" target="_blank">Graphite</a> 
    * **PR #12284** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12284" target="_blank">Graphite</a>  👈
    * **PR #12004** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12004" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12284?utm_source=stack-comment).

sryza (commented):
Is there a test or something we can write to verify that this works?

sryza:
Or at least include a code snippet in the PR description that you tested this on?

sryza:
I don't see why `DynamicPartitionsDefinition` should be any different than `StaticPartitionsDefinition`? In both cases, all the partitions are strings.

smackesey:
>Or at least include a code snippet in the PR description that you tested this on?

I added the snippet

>I don't see why DynamicPartitionsDefinition should be any different than StaticPartitionsDefinition? In both cases, all the partitions are strings.

Hmm not that familiar with these APIs. The signature on `DynamicPartitionsDefinition` shows that `partition_fn` can return arbitrary `Partition` and not just string though?

sryza:
> Hmm not that familiar with these APIs. The signature on DynamicPartitionsDefinition shows that partition_fn can return arbitrary Partition and not just string though?

Ah right, the legacy version

sryza (approved):

PR #12302, opened 2023-02-13
[dagit] Expose range-based asset health, use it for partition status rendering

bengotow:
### Summary & Motivation

Earlier this month we started sending asset partition health to Dagit as "ranges" of partition keys. The goal of this was to more concisely express the typical case, in which large bands of a potentially enormous range are all in the same state.

This PR updates the partition health bars in Dagit to take advantage of this format, which more directly maps onto what we render (green / gray bars). This PR finally de-couples the performance of the partition health bars from the number of keys in the asset partition space.

This code is complicated because Dagit allows you to:

1) view partition health on both dimensions of a multi-partitioned asset, even though the data sent to Dagit is always ranges on one dimension with subranges on the other. This requires that dagit can effectively "pivot" ranges onto the other axis.

2) view partition health filtered to a selection of keys. ("Eg: If I look at just "Prod" on dimension 2, do I have all of 2022?" on dimension 1?) This is a super cool interaction in the backfill modal but requires that Dagit can slice and subset range-based data.

3) view the aggregate health of several assets at once, flattening their individual health ranges into a single bar that shows "partial" where some of the assets are unmaterialized. (in the backfill modal if you have multiple assets selected). This requires that we can "union" overlapping ranges and generate a new summary set. 

Next Steps:

- New tests covering all the utility methods with cases written to exercise all the branches in the logic.

- New tests covering the output of the range accessor on the health data object. I wrote these tests directly from the example data and then verified the output against the function (which I think is important because it'd be easy to copy-paste the output into the toEqual and not catch a bug in this...)

- I screenshotted the storybook coverage of PartitionHealthSummary and verified the rendering is the same for a variety of health bars. This was super useful in catching a bunch of bugs as I was writing the code.

### Follow-ups:

- There are only two callsites still using `stateForSingleDimension`. Most importantly, the AssetPartitions page uses this and introduces an O(length of partitions array) to implement the status dots next to the partition rows and the filtering for "Success / Missing" on that page. It'd be great to get rid of `stateForSingleDimension`, because it's largely just a way-less-performant version of `rangesForSingleDimension`.

- The PartitionState enum should really be called PartitionStatus to match RunStatus. Will fix this on a quiet Friday at some point soon.

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
+++ b/js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
@@ -329,13 +329,10 @@ const LaunchAssetChoosePartitionsDialogBody: React.FC<Props> = ({
             <DimensionRangeWizard
               key={range.dimension.name}
               partitionKeys={range.dimension.partitionKeys}
-              partitionStateForKey={(dimensionKey) =>
-                mergedHealth.stateForSingleDimension(
-                  idx,
-                  dimensionKey,
-                  selections.length === 2 ? selections[1 - idx].selectedKeys : undefined,
-                )
-              }
+              ranges={mergedHealth.rangesForSingleDimension(
+                idx,
+                selections.length === 2 ? selections[1 - idx].selectedRanges : undefined,
+              )}
               selected={range.selectedKeys}
               setSelected={(selectedKeys) =>
                 setSelections(
--- a/js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
+++ b/js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
@@ -329,10 +329,12 @@ const LaunchAssetChoosePartitionsDialogBody: React.FC<Props> = ({
             <DimensionRangeWizard
               key={range.dimension.name}
               partitionKeys={range.dimension.partitionKeys}
-              ranges={mergedHealth.rangesForSingleDimension(
-                idx,
-                selections.length === 2 ? selections[1 - idx].selectedRanges : undefined,
-              )}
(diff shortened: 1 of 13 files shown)


js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
```diff
@@ -329,13 +329,12 @@ const LaunchAssetChoosePartitionsDialogBody: React.FC<Props> = ({
             <DimensionRangeWizard
               key={range.dimension.name}
               partitionKeys={range.dimension.partitionKeys}
-              partitionStateForKey={(dimensionKey) =>
-                mergedHealth.stateForSingleDimension(
+              health={{
```
bengotow, line 332: Unfortunately we use the same UI for both asset backfills / partition display and for op job backfills / partition display, and the op GraphQL API still returns one status object for every partition. We still need to support health data being provided in this way, so I turned this into a little union type.

js_modules/dagit/packages/core/src/assets/MultipartitioningSupport.tsx
```diff
@@ -67,17 +78,110 @@ export function mergedAssetHealth(
...
+  return assembleRangesFromTransitions(allKeys, transitions, rangeSets.length);
+}
+
+export function assembleRangesFromTransitions(
+  allKeys: string[],
+  transitions: {idx: number; delta: number}[],
+  maxOverlap: number,
+) {
+  // sort the array
+  transitions.sort((a, b) => a.idx - b.idx);
```
hellendag, line 142: Probably want to sort a copy this since `sort` is mutative.

hellendag (approved):
I can't really follow much of this, so I'm going to trust that your tests cover the complexity. :) Thanks for the detailed comments in the code!

PR #12172, opened 2023-02-08
nux examples in oss 1/: add dagster_cloud.yaml to prep cloud nux onboarding

yuhan:
### Summary & Motivation
add `dagster_cloud.yaml` to 6 examples which will be used by the cloud nux "start from a template"
this also helps users to "import an existing project" smoothier as when they start from an oss example which already has `dagster_cloud.yaml`, they don't need to construct the cloud yaml themselves to deploy to cloud thru "import an existing project".

unblock https://github.com/dagster-io/internal/pull/4791

PR #12304, opened 2023-02-13
add dagster-cloud everywhere

slopp:
### Summary & Motivation

Customers who start with one of our older examples, or customers who start with the project scaffold, find it surprising/frustrating/confusing when deployments to cloud fail.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/policy.py
+++ b/python_modules/dagster/dagster/_core/definitions/policy.py
@@ -77,12 +77,12 @@ class RetryPolicy(
     ):
         if backoff is not None and delay is None:
             raise DagsterInvalidDefinitionError(
-                "Can not set backoff on RetryPolicy without also setting delay"
+                "Can not set jitter on RetryPolicy without also setting delay"
             )
 
         if jitter is not None and delay is None:
             raise DagsterInvalidDefinitionError(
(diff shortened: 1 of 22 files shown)


python_modules/dagster/dagster/_core/definitions/policy.py
```diff
@@ -53,9 +53,19 @@ class RetryPolicy(
             being started. This unit of time can be modulated as a function of attempt number
             with backoff and randomly with jitter.
         backoff (Optional[Backoff]):
-            A modifier for delay as a function of retry attempt number.
+            A modifier for delay as a function of retry attempt number, either dagster.Backoff.LINEAR or dagster.Backoff.EXPONENTIAL
```
yuhan, line 56: is it intentional to include changes in this file here?

yuhan (commented):
i lately added `dagster_cloud.yaml` to six examples used by nux (https://github.com/dagster-io/dagster/pull/12172). if we're adding dagster-cloud as a dep, maybe we can do that to everywhere too?

slopp:
Ugh... actually, this is a little tricky, because we'd have to pull dagster-cloud from internal master to match the dev version contraints in order to install the examples and pass the repo tests.

PR #12329, opened 2023-02-14
Fix multipartitions run length encoding error

clairelin135:
Seen here: https://app.datadoghq.com/apm/trace/3303427780901995569?spanID=13139862780265732778&spanViewType=errors

This PR is mostly just snapshot updates and is hopefully fairly fast to review!

The root cause is that the first dimension has no partitions. This can happen on an empty static partitions definition, or a time window partitions definition where the start time is in the future.

This PR adds a check for this case, exiting early if the primary dimension has no partition keys. The run length encoding will still succeed if the secondary dimension has no partition keys.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql_tests/graphql/repo.py
+++ b/python_modules/dagster-graphql/dagster_graphql_tests/graphql/repo.py
@@ -1857,6 +1857,19 @@ def multipartitions_2(multipartitions_1):
     return multipartitions_1
 
 
+no_partitions_multipartitions_def = MultiPartitionsDefinition(
+    {
+        "a": StaticPartitionsDefinition([]),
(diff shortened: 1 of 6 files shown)


python_modules/dagster-graphql/dagster_graphql_tests/graphql/repo.py
```diff
@@ -1857,6 +1857,19 @@ def multipartitions_2(multipartitions_1):
     return multipartitions_1
 
 
+no_partitions_multipartitions_def = MultiPartitionsDefinition(
+    {
+        "a": StaticPartitionsDefinition([]),
+        "b": StaticPartitionsDefinition([]),
```
bengotow, line 1863: Ahhh I see - glad you added a test case for this!

bengotow (approved):

PR #12331, opened 2023-02-14
[refactor] Renames of pipeline/solid symbols in execution APIs

smackesey:

PR #12330, opened 2023-02-14
[typing/static] Execution API types

smackesey:
### Summary & Motivation

Add type annotations for mostly execution-related modules. Pure typing PR-- no runtime changes.

### How I Tested These Changes

BK

PR #12332, opened 2023-02-14
Runtime staleness calculation

smackesey:
### Summary & Motivation

Rough implementation of runtime staleness calculation.

- Provenance for all assets in a pipeline is loaded during
`ExecutionPlan` construction and stored on `KnownExecutionState`.
Probably not the best place to put it, but currently following the
example of the older memoization implementation.
- Inside `ActiveExecution`, asset materializations are recorded and any
logical version associated with them is extracted.
- When a step is popped off the "pending" queue, we compare the
provenance loaded at plan construction time with the projected
provenance during this run. We skip if there is no change.

There are currently many todos but it passes basic tests.

New unit tests.


smackesey:
* master
  * **PR #12189** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12189" target="_blank">Graphite</a> 
    * **PR #12330** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12330" target="_blank">Graphite</a> 
      * **PR #12331** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12331" target="_blank">Graphite</a> 
        * **PR #12332** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12332" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12332?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/logical_version.py
```diff
@@ -55,6 +71,42 @@ def __new__(
 UNKNOWN_LOGICAL_VERSION: Final[LogicalVersion] = LogicalVersion("UNKNOWN")
 
 
+class LogicalVersionProvenanceSerializer(DefaultNamedTupleSerializer):
```
sryza, line 74: Why do we need a custom serializer?

python_modules/dagster/dagster/_core/execution/plan/active.py
```diff
@@ -222,10 +234,16 @@ def _update(self) -> None:
                             ]
                             break
 
+                if not self.is_provenance_changed(step):
+                    should_skip = True
+                    self._skipped_deps[step_key] = []
+
                 if should_skip:
+                    print(f"SKIPPING {step.key}")
```
sryza, line 242: print statements still in here just in case you're not aware

python_modules/dagster/dagster/_core/execution/plan/active.py
```diff
@@ -249,6 +267,32 @@ def _update(self) -> None:
             self._executable.append(key)
             del self._waiting_to_retry[key]
 
+    def is_provenance_changed(self, step: ExecutionStep) -> bool:
+        existing_combined_input_versions: Dict[AssetKey, LogicalVersion] = {}
```
sryza, line 271: It's worth adding some docstring here to explain what this function is intended to do. Is it something like "determine whether versions of any of the inputs of the assets targeted by the step have changed"?

python_modules/dagster/dagster/_core/execution/plan/state.py
```diff
@@ -29,6 +31,27 @@
     from dagster._core.execution.plan.plan import StepHandleUnion
 
 
+@whitelist_for_serdes
+class AssetProvenanceData(NamedTuple):
+    asset_key: AssetKey
+    logical_version_provenance: LogicalVersionProvenance
+
+    @staticmethod
+    def list_from_map(
```
sryza, line 40: Given that each of these static methods is just used in a single place and the functionality is fairly generic collections manipulation, could the implementations just be inlined?

python_modules/dagster/dagster/_core/execution/plan/state.py
```diff
@@ -97,6 +120,7 @@ class KnownExecutionState(
             ("step_output_versions", Sequence[StepOutputVersionData]),
             ("ready_outputs", Set[StepOutputHandle]),
             ("parent_state", Optional[PastExecutionState]),
+            ("asset_provenance", Sequence[AssetProvenanceData]),
```
sryza, line 123: Any particular reason not to use a `Mapping[AssetKey, LogicalVersionProvenance]` here? It would enforce one-provenance-per-asset-key, which seems desirable.

sryza (commented):
Mind explaining how `KnownExecutionState` is related? My (mostly uninformed) understanding of `KnownExecutionState` was that it's for data that's not known until runtime, that needs to accessed within the steps.  Is it accessed within the steps in this case?

One thing to think about is whether there's any risk in ballooning the `KnownExecutionState` in situations where a single dbt op corresponds to thousands of assets.

python_modules/dagster/dagster/_core/definitions/logical_version.py
smackesey: I may ultimately change the format of `LogicalVersionProvenance` to avoid this annoyance, but it requires a custom serializer because it has a dict with `AssetKey` as key (not allowed in JSON).

python_modules/dagster/dagster/_core/execution/plan/state.py
smackesey: LOL I went down this exact road. I don't use it because it requires _another_ custom serializer (`AssetKey` as dict key).

We need some sugar for serializers, it's absurd how much code you have to write to convert one filed that has a non-standard dict key.

python_modules/dagster/dagster/_core/execution/plan/state.py
smackesey: I would prefer to leave it as is because it directly mirrors the structure and handling of`StepOutputVersionData`. I would like to eliminate both of these in one stroke later.

smackesey:
>Mind explaining how KnownExecutionState is related? My (mostly uninformed) understanding of KnownExecutionState was that it's for data that's not known until runtime, that needs to accessed within the steps. Is it accessed within the steps in this case?

I am certainly not wedded to `KnownExecutionState` as the correct channel for the provenance information-- I don't think I've wrapped my head sufficiently around the whole architecture of the system to have a great sense of what the best channel would be.

For this prototype, in order to keep complexity down I decided to use the same channel as the existing memoization infrastructure, which passes `step_output_version_data` via `KnownExecutionState`. I also structured the asset provenance data similarly.

python_modules/dagster/dagster/_core/execution/plan/state.py
sryza: Oof, right. I ran into this same issue when working on backfills. Would be cool to upgrade our serialization machinery support this.

sryza:
IMO it's worth writing out the other possible options for how to do this and comparing with the option here. Even if we choose to go with this option for the sake of expediency, IMO it's important to understand the design tradeoffs.

It's worth chatting with @dpeng817 to understand the constraints that led to the original design. I think there's a good chance that KnownExecutionState is in fact the most principled way to implement this.

schrockn:
It appears that this requires computing all logical versions prior to the entire run. That is an interesting approach.

A critical question, however, is when we will launch the compute that will actually compute the logical hashes. That is unanswered by the PR. I assume this would involve having a job whose purpose is the compute the hashes and then have _that_ job kickoff the actual run. We can do this but we need to make sure to be intentional about this and think about the end-to-end experience.

I think we can implement this far more simply by pushing some complexity onto the user in the short term, in a way that is pretty intuitive, provides a slightly degraded experience while delivering the core value, and does so without permanently complicating the framework in a rushed design process.

* Provide an API to the user that allows them to fetch the latest version for an asset key.
* They write the logic to compare that version to a version that they compute. They control both sides of this (they have invoked a function to compute the logical version on materialization).
* If the versions do not match, they are responsible for instructing the framework to skip the current step.


```
@asset
def an_asset(context):
    recorded_data_version = context.data_version_for_asset_key(context.asset_key)
    computed_data_version = compute_data_version_for_asset_key(context.asset_key)
    if recorded_data_version != computed_data_version:
       context.log_skipped_event() # if we want to change the semantics of skip a allow user scope to tell use that the step has been skipped OR
       context.log("YAY WE CAN SKIP THIS TEST") # user just puts in the helpful message in event log and they can infer what is skipped by seeing that step is trivially short
       return
```

schrockn:
> A critical question, however, is when we will launch the compute that will actually compute the logical hashes. That is unanswered by the PR. I assume this would involve having a job whose purpose is the compute the hashes and then have that job kickoff the actual run. We can do this but we need to make sure to be intentional about this and think about the end-to-end experience.

Furthermore, if this is indeed the approach, we could just have the user passed the subselection to execute directly, and avoid changing the framework in its entirity.

dpeng817:
Had to dig up phabricator for this (rip). If anyone wants to see the original PR: https://dagster.phacility.com/D9220

Just to give some context on the original reasons we added versions to `KnownExecutionState` (although unfortunately many of these conversations are lost to the quip + slack migration); 
IIRC it was in large part due to subsetted execution + crossing process boundaries. `KnownExecutionState` is our espoused method of persisting dynamically computed information about the run across process boundaries, and version information fit that bill (at least at the time). 
Originally, we used the execution plan snapshot for storing versioning information, but when performing subsetted execution (eg retry from failure), we use `KnownExecutionState` to introspect information that was calculated during a previous run. If retrying from failure or even performing subsetted execution, you need to know which versions to use for the not-running upstream steps in order to load outputs. Just as we use `KnownExecutionState` to introspect which mapping keys were created during a previous run, we used it to introspect which versions upstream outputs were stored with in a prior run.

Haven't kept up with the asset versioning/memoization workstream, but I wonder if you could use the event log for a similar purpose now, ad-hoc retrieval of version information from a prior run (assuming events like `AssetMaterialization` and `AssetObservation` now have versioning information). Although I think that if you want to be able to memoize on a per-step basis, (ie different steps in a graph-backed asset, or an asset containing dynamic execution steps) `KnownExecutionState` may still be your best bet.

cmpadden:

PR #12333, opened 2023-02-14
Include partition in asset materialization planned event

johannkm:
Record which partition of an asset a run plans to materialize, so we can use it to say a materialization attempt failed.

Leaving out asset partition ranges for now. These will require resolving the range to actual partitions inside a host process.

### DIFF
--- a/python_modules/dagster/dagster/_core/events/__init__.py
+++ b/python_modules/dagster/dagster/_core/events/__init__.py
@@ -652,6 +652,8 @@ class DagsterEvent(
             return self.step_materialization_data.materialization.partition
         elif self.event_type == DagsterEventType.ASSET_OBSERVATION:
             return self.asset_observation_data.asset_observation.partition
+        elif self.event_type == DagsterEventType.ASSET_MATERIALIZATION_PLANNED:
+            return self.asset_materialization_planned_data.partition
         else:
             return None
 
@@ -1451,11 +1453,16 @@ class StepMaterializationData(
 
 @whitelist_for_serdes
 class AssetMaterializationPlannedData(
-    NamedTuple("_AssetMaterializationPlannedData", [("asset_key", AssetKey)])
+    NamedTuple(
+        "_AssetMaterializationPlannedData",
+        [("asset_key", AssetKey), ("partition", Optional[str])],
+    )
 ):
-    def __new__(cls, asset_key: AssetKey):
+    def __new__(cls, asset_key: AssetKey, partition: Optional[str] = None):
         return super(AssetMaterializationPlannedData, cls).__new__(
-            cls, asset_key=check.inst_param(asset_key, "asset_key", AssetKey)
+            cls,
+            asset_key=check.inst_param(asset_key, "asset_key", AssetKey),
+            partition=check.opt_str_param(partition, "partition"),
         )
 
 
--- a/python_modules/dagster/dagster/_core/instance/__init__.py
+++ b/python_modules/dagster/dagster/_core/instance/__init__.py
@@ -61,7 +61,14 @@ from dagster._core.storage.pipeline_run import (
     RunsFilter,
     TagBucket,
 )
-from dagster._core.storage.tags import PARENT_RUN_ID_TAG, RESUME_RETRY_TAG, ROOT_RUN_ID_TAG
+from dagster._core.storage.tags import (
+    ASSET_PARTITION_RANGE_END_TAG,
+    ASSET_PARTITION_RANGE_START_TAG,
+    PARENT_RUN_ID_TAG,
+    PARTITION_NAME_TAG,
+    RESUME_RETRY_TAG,
+    ROOT_RUN_ID_TAG,
+)
 from dagster._core.utils import str_format_list
 from dagster._serdes import ConfigurableClass
 from dagster._seven import get_current_datetime_in_utc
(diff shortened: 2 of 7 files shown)


johannkm:
* master
  * **PR #12333** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12333" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12333?utm_source=stack-comment).

python_modules/dagster/dagster/_core/events/__init__.py
```diff
@@ -1451,11 +1453,16 @@ def __new__(
 
 @whitelist_for_serdes
 class AssetMaterializationPlannedData(
-    NamedTuple("_AssetMaterializationPlannedData", [("asset_key", AssetKey)])
+    NamedTuple(
+        "_AssetMaterializationPlannedData",
+        [("asset_key", AssetKey), ("partition_key", Optional[str])],
```
sryza, line 1458: We probably should have used `partition_key` in `AssetMaterialization`, but, now that we've gone with `partition`, I think it's worth being consistent here.

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1122,31 +1129,56 @@ def _ensure_persisted_execution_plan_snapshot(
...
             DagsterEventType,
         )
 
-        pipeline_name = pipeline_run.pipeline_name
+        pipeline_name = run.pipeline_name
 
         for step in execution_plan_snapshot.steps:
             if step.key in execution_plan_snapshot.step_keys_to_execute:
                 for output in step.outputs:
+                    assert output.properties
```
sryza, line 1146: We generally avoid asserts in core Dagster and prefer something like:

```python
properties = check.not_none(output.properties)
```

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1122,31 +1129,56 @@ def _ensure_persisted_execution_plan_snapshot(
...
                         event = DagsterEvent(
                             event_type_value=DagsterEventType.ASSET_MATERIALIZATION_PLANNED.value,
                             pipeline_name=pipeline_name,
                             message=(
                                 f"{pipeline_name} intends to materialize asset"
                                 f" {asset_key.to_string()}"
                             ),
-                            event_specific_data=AssetMaterializationPlannedData(asset_key),
+                            event_specific_data=AssetMaterializationPlannedData(
+                                asset_key, partition_key=partition_key
```
sryza, line 1178: feel free to ignore, but could save a couple lines by doing `partition_key=partition_tag` here

python_modules/dagster/dagster_tests/asset_defs_tests/test_assets_job.py
```diff
@@ -1941,3 +1942,47 @@ def unpartitioned_asset():
...
+
+
+partitioned_asset_job = build_assets_job("static_partition_asset_job", [static_partition_asset])
+
+
+def reconstruct_partitioned_asset_job():
+    return partitioned_asset_job
+
+
+def test_asset_materialization_planned():
```
sryza, line 1959: I think `test_asset_events` might be a better place for this test. Also, I think we might e able to avoid building a job and instead just call `materialize_to_memory(instance=instance...)`.

sryza (changes requested):
Nice

python_modules/dagster/dagster_tests/core_tests/test_asset_events.py
```diff
@@ -87,3 +88,22 @@ def my_asset():
...
+
+def test_asset_partition_materialization_planned_events():
+    @asset(partitions_def=StaticPartitionsDefinition(["a", "b"]))
+    def my_asset():
+        pass
+
+    assets_job = build_assets_job("assets_job", [my_asset])
+
+    with instance_for_test() as instance:
+        result = assets_job.execute_in_process(instance=instance, partition_key="a")
```
sryza, line 101: IMO preferable to use `materialize`, but not a big deal

sryza (dismissed):
Nice

python_modules/dagster/dagster_tests/core_tests/test_asset_events.py
johannkm: swapped

clairelin135 (approved):
Can we also add a test for the case that an unpartitioned asset depends on a partitioned asset? Calling `execute_in_process(partition_key=...)` on this job should yield a materialization planned event for each asset, but the unpartitioned asset's event shouldn't have a partition key.

I think it would be good to add a test for this scenario.

johannkm:
Discussed offline, good call by Claire- we'd record a partition even for non partitioned assets. Figuring out how to tell- do we need to put `partitioned` in the execution plan?

clairelin135 (changes requested):
Discussed a bit with Johann offline about this, but the current approach will fail when an unpartitioned asset is part of an unpartitioned run (as we check the run tags for the partition and add it to each asset materialization planned event).

My thinking now is that we might have to build the execution plan knowing which partition each step will output. Since we'll likely support a future where a job can contain multiple different partitions defs, knowing whether the asset is partitioned or not is not enough. I think we can add logic to the step output properties that will get serialized into the execution plan that stores the partition key(s) of a given step: https://github.com/dagster-io/dagster/blob/85c1ac5b084e3032d85866e045fb4f7cc3fda02e/python_modules/dagster/dagster/_core/execution/plan/compute.py#L76

sryza:
> Since we'll likely support a future where a job can contain multiple different partitions defs

Right now, the assumption that each run targets the same partition (or partition range) for all the non-partitioned assets inside it is pretty deeply baked into our system.  E.g. our backfill system relies on it, as well as all of our context partition methods.

My instinct is that it's not so bad to rely on this assumption in the medium term. I'd rather rely on that assumption than, say, encode thousands of partition keys in the execution plan for each asset when someone executes a run with a partition range.

johannkm:
Ack on not worrying about runs with multiple partition defs. How should we solve knowing which steps are partitioned or not?

clairelin135:
In that case, I'm fine with storing whether the asset is partitioned or not on the execution plan.

johannkm:
I propose that instead of having to find and use the ExternalAssetGraph in the create run callsites, we add a field to StepOutputProperties. This object already has is_asset and asset_key fields that we use to emit the correct materialization planned events, so it seems like a natural place.

sryza:
> I propose that instead of having to find and use the ExternalAssetGraph in the create run callsites, we add a field to StepOutputProperties. This object already has is_asset and asset_key fields that we use to emit the correct materialization planned events, so it seems like a natural place.

As long as we're able to avoid the size of `StepOutputProperties` becoming O(# partitions) when we add support for partition ranges, this sounds good to me.

johannkm:
As long as 1 run can only materialize 1 partition or 1 partition range, we just have to know if the asset is partitioned or not. So we wouldn't need to include raw partition keys.

If we break that invariant in the future... maybe we'd expose which partition defs are on which assets?

sryza:
> As long as 1 run can only materialize 1 partition or 1 partition range, we just have to know if the asset is partitioned or not. So we wouldn't need to include raw partition keys.

I think that's a safe assumption for the medium term

PR #12345, opened 2023-02-14
[For 1.2] Don't include run/job tags in k8s_job_ops k8s config computations

gibsondan:
Summary:
This makes k8s_job_op consistent with k8s_job_exeuctor ops wrt how it handles bringing in job tags to the launched ops, following discussion on https://github.com/dagster-io/dagster/pull/12308/files - plan is to release this after the 1.2 branch cut since it's a breaking change and there's been a bit of thrash on k8s_job_op config during this release.

### DIFF
--- a/python_modules/libraries/dagster-k8s/dagster_k8s/ops/k8s_job_op.py
+++ b/python_modules/libraries/dagster-k8s/dagster_k8s/ops/k8s_job_op.py
@@ -202,7 +202,7 @@ def execute_k8s_job(
         context.instance.run_launcher
         if isinstance(context.instance.run_launcher, K8sRunLauncher)
         else None,
-        include_run_tags=True,
+        include_run_tags=False,
     )
 
     container_config = container_config.copy() if container_config else {}
--- a/integration_tests/test_suites/k8s-test-suite/tests/test_k8s_job_op.py
+++ b/integration_tests/test_suites/k8s-test-suite/tests/test_k8s_job_op.py
@@ -288,3 +288,53 @@ def test_k8s_job_op_retries(namespace, cluster_provider):
 
     assert "HERE IS RETRY NUMBER 0" in _get_pod_logs(cluster_provider, job_name, namespace)
     assert "HERE IS RETRY NUMBER 1" in _get_pod_logs(cluster_provider, job_name + "-1", namespace)
+
+
+@pytest.mark.default
+def test_k8s_job_op_ignore_job_tags(namespace, cluster_provider):
+    @op
+    def the_op(context):
+        execute_k8s_job(
+            context,
+            image="busybox",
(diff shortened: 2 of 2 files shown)


python_modules/dagster/dagster/_core/execution/context/invocation.py
```diff
@@ -607,6 +634,8 @@ def build_op_context(
     config: Any = None,
     partition_key: Optional[str] = None,
     mapping_key: Optional[str] = None,
+    run: Optional[DagsterRun] = None,
+    op_name: Optional[str] = None,
```
alangenfeld, line 638: I would expect to be able to derive `op_name` from the `op` once we bind the context to the invoked op

python_modules/dagster/dagster/_core/execution/context/invocation.py
```diff
@@ -607,6 +634,8 @@ def build_op_context(
     config: Any = None,
     partition_key: Optional[str] = None,
     mapping_key: Optional[str] = None,
+    run: Optional[DagsterRun] = None,
```
alangenfeld, line 637: cc @dpeng817 

I forget what our past discussions around adding run were - if we had a real reason to not or were just waiting for a good reason to

alangenfeld (changes requested):
thanks
shelve this in your queue for now

python_modules/dagster/dagster/_core/execution/context/invocation.py
gibsondan: I might just switch this to use a crazy magic mock instead

dpeng817: DagsterRun is an api that we expect users not to construct directly, only use for type annotations. Given that often in test scenarios they would be mocking this object, it seemed like a pretty weird API surface area to add

python_modules/libraries/dagster-k8s/dagster_k8s/ops/k8s_job_op.py
```diff
@@ -202,7 +202,7 @@ def execute_k8s_job(
         context.instance.run_launcher
         if isinstance(context.instance.run_launcher, K8sRunLauncher)
         else None,
-        include_run_tags=True,
+        include_run_tags=False,
```
alangenfeld, line 205: i forget can we clean up this bool then?

alangenfeld (approved):

python_modules/libraries/dagster-k8s/dagster_k8s/ops/k8s_job_op.py
gibsondan: nah its true in the run launcher

PR #12348, opened 2023-02-14
change storage signature for run tags

prha:
### Summary & Motivation
We want to enable some version of the tag autocomplete search on the runs page again.

Instead of fetching all key/value pairs of tags to power the search, I think we can do something where we first fetch all the possible keys, select a key (from a list) and then fetch all the values for the selected key.

For high cardinality keys, we can further do filtering by prefix and apply a limit.

BK


alangenfeld (approved):
nice

PR #9373, opened 2022-08-15
default_load_type argument to build_snowflake_io_manager

sryza:

PR #12356, opened 2023-02-15
[db io managers] add default_load_type

### LINKED ISSUE #12328: Snowflake IO Manager should not require type hints when set up to handle only one type
### What's the use case?

When opening a connection to Snowflake, it is possible to use `cursor.fetch_pandas.all()` to create a DataFrame without using type hints. However, when that DataFrame is passed to another asset as a parameter, the Snowflake IO Manager requires type hints to be able to read the Snowflake table into a DataFrame. This is inconvenient when working exclusively with Pandas or Spark. 

For instance, if I create an asset like this with the Snowflake IO Manager:

```
@asset
def my_asset(context) -> pd.DataFrame:
    with context.resources.snowflake.get_connection() as conn:
            with closing(conn.cursor()) as cursor:
                q2 = "SELECT * FROM SANDBOX.MY_SCHEMA.MY_TABLE"
                cursor.execute(q2)
                df = cursor.fetch_pandas_all()
                return df
```

I get an error if I try to load the same df in the next asset:

```
@asset
def my_next_asset(context, my_asset) -> pd.DataFrame:
    stuff...
```

The error is:

`dagster._check.CheckError: SnowflakeIOManager does not have a handler for type 'typing.Any'. Has handlers for types '<class 'pandas.core.frame.DataFrame'>'. Please add <class 'pandas.core.frame.DataFrame'> type hints to your assets and ops.`

### Ideas of implementation

@jamiedemaria says

> So the snowflake io manager can be set up to handle different types at once. for example you could make a snowflake io manager that can handle both pandas and pyspark dataframes. In that case we require the typing hints to know which type to load the input as. However, in the case when the io manager is only set up to handle one type, we should default to using that type if there are no hints. I’ve been mentally tracking this issue, but haven’t had a chance to fix it yet. if you want to open a github issue with your code sample and assign it to me we can officially track it

### Additional information

_No response_

jamiedemaria:
### Summary & Motivation
Adds `default_load_type` to DBIOManager so that they can load Any types as `default_load_type`

fixes #12328

replaces #9373 which did the same thing (but wasn't merged and now the merge conflicts are bad)

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/db_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/db_io_manager.py
@@ -1,3 +1,4 @@
+import warnings
 from abc import ABC, abstractmethod
 from contextlib import contextmanager
 from typing import (
@@ -27,7 +28,7 @@ from dagster._core.definitions.time_window_partitions import (
     TimeWindow,
     TimeWindowPartitionsDefinition,
 )
-from dagster._core.errors import DagsterInvalidDefinitionError
+from dagster._core.errors import DagsterInvalidDefinitionError, DagsterInvariantViolationError
 from dagster._core.execution.context.input import InputContext
 from dagster._core.execution.context.output import OutputContext
 from dagster._core.storage.io_manager import IOManager
@@ -157,6 +158,14 @@ class DbIOManager(IOManager):
     def load_input(self, context: InputContext) -> object:
         obj_type = context.dagster_type.typing_type
         if obj_type is Any and self._default_load_type is not None:
+            msg = (
+                f"Input {context.name} does not have a type hint, loading as default type"
+                f" {self._default_load_type}"
+            )
+            try:
+                context.log.warning(msg)
+            except DagsterInvariantViolationError:
+                warnings.warn(msg)
             load_type = self._default_load_type
         else:
             load_type = obj_type
--- a/python_modules/dagster/dagster/_core/storage/db_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/db_io_manager.py
@@ -96,6 +96,7 @@ class DbIOManager(IOManager):
         database: str,
         schema: Optional[str] = None,
         io_manager_name: Optional[str] = None,
+        default_load_type: Optional[Type] = None,
     ):
         self._handlers_by_type: Dict[Optional[Type], DbTypeHandler] = {}
         self._io_manager_name = io_manager_name or self.__class__.__name__
@@ -114,6 +115,14 @@ class DbIOManager(IOManager):
         self._db_client = db_client
         self._database = database
         self._schema = schema
+        if (
+            default_load_type is None
+            and len(type_handlers) == 1
+            and len(type_handlers[0].supported_types) == 1
+        ):
+            self._default_load_type = type_handlers[0].supported_types[0]
+        else:
+            self._default_load_type = default_load_type
 
     def handle_output(self, context: OutputContext, obj: object) -> None:
         table_slice = self._get_table_slice(context, context)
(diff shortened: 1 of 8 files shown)


jamiedemaria:
* master
  * **PR #12356** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12356" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12356?utm_source=stack-comment).

python_modules/dagster/dagster/_core/storage/db_io_manager.py
```diff
@@ -138,9 +147,14 @@ def handle_output(self, context: OutputContext, obj: object) -> None:
...
-        self._check_supported_type(obj_type)
+        if obj_type is Any and self._default_load_type is not None:
+            load_type = self._default_load_type
+        else:
+            load_type = obj_type
+
+        self._check_supported_type(load_type)
 
-        return self._handlers_by_type[obj_type].load_input(
+        return self._handlers_by_type[load_type].load_input(
```
benpankow, line 157: Is there a way that we could hint to the user that this is happening, in the case that it leads to unexpected behavior? I could see a situation where a user might be confused why their inputs are presented as a particular type.

I think this is a fairly minor issue especially since it only affects input loading, so not blocking (in theory there should be no issues with calling `load_input`, I could see default output types being more tricky)

benpankow (approved):
great, this is a big ergonomic win

python_modules/dagster/dagster/_core/storage/db_io_manager.py
```diff
@@ -138,9 +148,22 @@ def handle_output(self, context: OutputContext, obj: object) -> None:
 
     def load_input(self, context: InputContext) -> object:
         obj_type = context.dagster_type.typing_type
-        self._check_supported_type(obj_type)
+        if obj_type is Any and self._default_load_type is not None:
+            msg = (
+                f"Input {context.name} does not have a type hint, loading as default type"
```
sryza, line 153: I imagine this warning could get kind of annoying. Do we need it?

sryza: To expand on that a little bit, I think that in some situations, this will be a perfectly reasonable way to use these IO managers, and a logging a warning makes it seem like we're trying to discourage it.

jamiedemaria: hmm ben had a comment about indicating that we were loading in this way, so i experimented with this. but i dont see the comment here anymore. maybe mentioning the behavior in docs would be enough?

benpankow: 😅 I ended up thinking over the use-case where the default was applied and felt less strongly about my comment, so I deleted it. Sorry about any confusion! I think documenting the behavior might be sufficient, since users importing and using e.g. the preconstructed `snowflake_pandas_io_manager` would hit this warning otherwise which feels odd.

jamiedemaria: ah ok that makes a lot more sense now. I'm happy with a docs change!

jamiedemaria: API docs have been updated. I'm not sure it's worth a callout in the guide, but happy to be convinced otherwise

python_modules/libraries/dagster-duckdb/dagster_duckdb/io_manager.py
```diff
@@ -15,13 +15,17 @@
...
+    type_handlers: Sequence[DbTypeHandler], default_load_type: Optional[Type] = None
+) -> IOManagerDefinition:
     """
     Builds an IO manager definition that reads inputs from and writes outputs to DuckDB.
 
     Args:
         type_handlers (Sequence[DbTypeHandler]): Each handler defines how to translate between
-            DuckDB tables and an in-memory type - e.g. a Pandas DataFrame.
+            DuckDB tables and an in-memory type - e.g. a Pandas DataFrame. If only
+            one DbTypeHandler is provided, it will be used as teh default_load_type.
```
jamiedemaria, line 27:
```suggestion
            one DbTypeHandler is provided, it will be used as the default_load_type.
```

python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
```diff
@@ -16,13 +16,17 @@
...
+    type_handlers: Sequence[DbTypeHandler], default_load_type: Optional[Type] = None
+) -> IOManagerDefinition:
     """
     Builds an IO manager definition that reads inputs from and writes outputs to Snowflake.
 
     Args:
         type_handlers (Sequence[DbTypeHandler]): Each handler defines how to translate between
-            slices of Snowflake tables and an in-memory type - e.g. a Pandas DataFrame.
+            slices of Snowflake tables and an in-memory type - e.g. a Pandas DataFrame. If only
+            one DbTypeHandler is provided, it will be used as teh default_load_type.
```
jamiedemaria, line 28:
```suggestion
            one DbTypeHandler is provided, it will be used as the default_load_type.
```

benpankow (approved):

PR #12357, opened 2023-02-15
[freshness-policies] Allow setting freshness policies when using graph-backed assets

OwenKephart:
### Summary & Motivation

https://github.com/dagster-io/dagster/issues/11677

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/assets.py
@@ -482,7 +482,9 @@ class AssetsDefinition(ResourceAddable):
             freshness_policies_by_key={
                 keys_by_output_name[output_name]: freshness_policy
                 for output_name, freshness_policy in freshness_policies_by_output_name.items()
-            } if freshness_policies_by_output_name else None,
+            }
+            if freshness_policies_by_output_name
+            else None,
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster/_core/definitions/assets.py
```diff
@@ -298,6 +299,10 @@ def from_graph(
                 be associated with each of the output assets for this node. Keys are names of the
                 outputs, and values are dictionaries of metadata to be associated with the related
                 asset.
+            freshness_policy_by_ouptut_name (Optional[Mapping[str, FreshnessPolicy]]): Defines a
```
sryza, line 302:
```suggestion
            freshness_policy_by_output_name (Optional[Mapping[str, FreshnessPolicy]]): Defines a
```

sryza (commented):
Thoughts on specifying this by asset key? I feel like that might be the more natural way to think about it?

OTOH, I know that's a little bit inconsistent with metadata_by_output_name.

OwenKephart:
@sryza I think of this setup as analogous to the AssetOut stuff, where everything is keyed by output name (including asset keys, metadata, and freshness policies)

sryza (approved):
Ok, I buy it

PR #12363, opened 2023-02-15
[dagster-pandas][dagster-pandera] assign a typing_type for generated pandas dataframe DagsterTypes

OwenKephart:
### Summary & Motivation

We examine the typing_type in our DbIOManagers to determine if the IOManager can load a given input. Under the hood, we expect these types to be pd.DataFrame, so the IOManager should be totally fine to load them. However, we currently spit out an error because the underlying typing_type defaults to `Any`.


jamiedemaria (approved):
lgtm!

PR #12367, opened 2023-02-15
Fix intermittent dynamic partitions table SQLite concurrency error

clairelin135:
[BK fails intermittently](https://buildkite.com/dagster/dagster/builds/49143#018656f4-c887-4210-99a8-0f4835e83021/253-1210) when different processes contend to create tables.

This PR adds the dynamic partitions table to the error list to prevent this test from failing spuriously.

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/event_log/sqlite/sqlite_event_log.py
+++ b/python_modules/dagster/dagster/_core/storage/event_log/sqlite/sqlite_event_log.py
@@ -184,6 +184,7 @@ class SqliteEventLogStorage(SqlEventLogStorage, ConfigurableClass):
                     or "table event_logs already exists" in err_msg
                     or "table asset_event_tags already exists" in err_msg
                     or "table alembic_version already exists" in err_msg
+                    or "table dynamic_partitions already exists" in err_msg
                     or "database is locked" in err_msg
                     or "UNIQUE constraint failed: alembic_version.version_num" in err_msg
                 ):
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster/_core/storage/event_log/sqlite/sqlite_event_log.py
```diff
@@ -184,6 +184,7 @@ def _initdb(self, engine: Engine) -> None:
                     or "table event_logs already exists" in err_msg
                     or "table asset_event_tags already exists" in err_msg
                     or "table alembic_version already exists" in err_msg
+                    or "table dynamic_partitions already exists" in err_msg
```
alangenfeld, line 187: <i class='graphite__hidden'>[Re: lines 182 to 187]<br/></i>

i guess it would need to be regex to handle this better but we should maybe try to forward proof this better since its not great to have to come back and update this <span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12367?utm_source=unchanged-line-comment">Graphite</a>.</span>

alangenfeld (approved):

python_modules/dagster/dagster/_core/storage/event_log/sqlite/sqlite_event_log.py
clairelin135: Swapped this with regex!

PR #12379, opened 2023-02-15
Add text to timestamp dropdown

dpeng817:
Adds some text to the timestamp dropdown of the schedule testing button to just provide a little more context.
Screen Shot 2023-02-15 at 3 16 52 PM


salazarm (approved):

braunjj:
Thank you! Two small nits:
- can we add like 2px of space between the text and the input?
- remove the period in the label

PR #12389, opened 2023-02-16
Make a script to template out new dagster packages

jamiedemaria:
### Summary & Motivation
Every time we make a new integration package there is a lot of boilerplate to copy over. This process can get tedious, and can result in some errors if you miss a name that needs to be changed. This script will make all of the files for you and put the correct name in the correct place. 

ran the script to create `dagster-test-package` which you can see in the diff (I'll remove it before merging!)


jamiedemaria:
* master
  * **PR #12389** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12389" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12389?utm_source=stack-comment).

yuhan (commented):
+@smackesey for 👀 on set up like type checking, version, etc

scripts/templates_create_dagster_package/tox.ini.tmpl
```diff
@@ -0,0 +1,20 @@
...
+extras =
+  test
+setenv =
+  VIRTUALENV_PIP=21.3.1
+  !windows: COVERAGE_ARGS = --cov=dagster --cov-append --cov-report term:skip-covered --cov-report html --cov-report xml
+  windows: COVERAGE_ARGS =
+passenv = CI_* COVERALLS_REPO_TOKEN BUILDKITE*
+deps =
+  -e ../../dagster[test]
+  -e ../{{ hyphen_name }}
```
jamiedemaria, line 15:
```suggestion
  -e .
```

scripts/templates_create_dagster_package/tox.ini.tmpl
```diff
@@ -0,0 +1,20 @@
...
+
+[testenv]
+usedevelop = true
+extras =
+  test
+setenv =
+  VIRTUALENV_PIP=21.3.1
+  !windows: COVERAGE_ARGS = --cov=dagster --cov-append --cov-report term:skip-covered --cov-report html --cov-report xml
+  windows: COVERAGE_ARGS =
+passenv = CI_* COVERALLS_REPO_TOKEN BUILDKITE*
```
jamiedemaria, line 12: add TODO to fill out additional env var requirements and add them to buildkite/dagster-buildkite/dagster_buildkite/steps/packages.py

scripts/templates_create_dagster_package/MANIFEST.in.tmpl
```diff
@@ -0,0 +1,3 @@
+include README.md
+include LICENSE
+include {{ underscore_name }}/py.typed
```
jamiedemaria, line 3: @yuhan @smackesey  I'm not 100% sure what should go in the Manifest.in file. I based it on some other ones we have, but I also see some with `recursive-include`s. Is there a rule for what should go in here?

python_modules/libraries/dagster-test-package/tox.ini
```diff
@@ -0,0 +1,21 @@
+[tox]
+envlist = py{310,39,38,37}-{unix,windows}
+
+[testenv]
+usedevelop = true
+extras =
+  test
```
smackesey, line 7: I think this might cause an error since there is no defined `test` extra in `setup.py`--  I think you can just delete `extras` here.

scripts/templates_create_dagster_package/MANIFEST.in.tmpl
smackesey: Good question, I think what you have is fine. Most relevant files are included by default: https://packaging.python.org/en/latest/guides/using-manifest-in/#how-files-are-included-in-an-sdist

You only need more entries here if there's stuff like an embedded JS webapp or a directory full of data files.

smackesey (approved):

PR #12390, opened 2023-02-16
[RFC] Add an is_test_evaluation property for sensor contexts

dpeng817:
When evaluating a sensor, there's currently no way to avoid side effects. Adding this property would allow users to gate side effect behavior to not occur during test runs (namely test sensor / schedule button, direct invocation, sensor preview).

This is somewhat similar to just including mocked resources, but I think that there's a distinct set of scenarios where you might want to test run things on real data / real infrastructure, but not actually risk messing up anything in prod. 

Curious what folks think of this pattern. Could imagine including something similar for jobs / assets when using the launchpad for example, but some folks use the launchpad for prod use cases, and arguably config can serve the same purpose.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/schema/instigation.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/instigation.py
@@ -230,6 +230,7 @@ class GrapheneDryRunInstigationTick(graphene.ObjectType):
                     cursor=self._cursor,
                     last_completion_time=None,
                     last_run_key=None,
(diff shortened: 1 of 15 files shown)


alangenfeld (commented):
I think if we are going to thread a bit of information through this whole stack - it should be something more substantial / useful than a bool that we could still be used to derive a bool method. Something like `RequestOrigin = daemon | dagit | direct-invocation` but more likely fleshed out than that.

PR #12400, opened 2023-02-16
Auto infer multipartition <-> single dimension mapping

clairelin135:
The existing `SingleDimensionDependencyMapping` class handles the upstream single-dimension -> downstream multipartitioned dependency relationship. This PR refactors this class to also handle the opposite relationship (upstream multipartition -> downstream single dimension), renaming to `MultiToSingleDimensionMapping`.

This class previously required a `partition_dimension_name` argument, which this PR makes optional. This argument is required when the multipartitions definition has dimension partition defs that are equivalent, in which case the class throws an error.

The final change is that we can now auto-infer when we should use this partition mapping--when a single dimension partitions def matches a dimension of a multipartitions def, and the two assets have a dependency relationship. This PR adds tests to ensure that we choose this partition mapping in those cases, otherwise relying on the `IdentityPartitionMapping` or `AllPartitionMapping`.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/partition_mapping.py
+++ b/python_modules/dagster/dagster/_core/definitions/partition_mapping.py
@@ -1,7 +1,19 @@
 import collections.abc
+import warnings
 from abc import ABC, abstractmethod
 from collections import defaultdict
-from typing import Collection, Mapping, NamedTuple, Optional, Sequence, Tuple, Type, Union, cast
+from typing import (
+    Collection,
+    Mapping,
+    NamedTuple,
+    Optional,
+    Sequence,
+    Set,
+    Tuple,
+    Type,
+    Union,
+    cast,
+)
 
 import dagster._check as check
 from dagster._annotations import PublicAttr, experimental, public
@@ -18,6 +30,7 @@ from dagster._core.definitions.partition_key_range import PartitionKeyRange
 from dagster._core.definitions.time_window_partitions import TimeWindowPartitionsDefinition
 from dagster._core.instance import DynamicPartitionsStore
 from dagster._serdes import whitelist_for_serdes
+from dagster._utils.backcompat import ExperimentalWarning
 from dagster._utils.cached_method import cached_method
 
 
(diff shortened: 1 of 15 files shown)


clairelin135:
I can also break this PR up into parts if preferred.

clairelin135:
@sryza this pr is ready for another pass--I've changed the class name to `MultiToSingleDimensionPartitionMapping` and removed `get_default_partition_mapping`.

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -248,42 +249,112 @@ def get_downstream_partitions_for_partition_range(
...
+                    if isinstance(partitions_def, MultiPartitionsDefinition)
+                }
+            )
+        )
+
+        single_dimension_partitions_def = next(
+            iter(
+                {
+                    upstream_partitions_def,
+                    downstream_partitions_def,
```
sryza, line 311: Nitpick: take out comma so these can be on the same line?

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -248,42 +249,112 @@ def get_downstream_partitions_for_partition_range(
...
+        ):
             check.failed(
-                "The upstream partitions definition does not have the same partitions definition "
-                f"as dimension {matching_dimension_def.name}"
+                "This partition mapping defines a relationship between a multipartitioned and"
+                " single dimensional asset. The single dimensional partitions definition must be a"
+                " dimension of the MultiPartitionsDefinition."
+            )
+
+        multipartitions_def = next(
```
sryza, line 297: Feel free to ignore, but this would arguably a little more standard as an if/else.

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -301,55 +372,110 @@ def get_downstream_partitions_for_partition_range(
...
+                in partitions_subset.get_partition_keys()
+            ):
+                matching_keys.append(key)
+        return matching_keys
+
+    def _get_single_dim_keys_from_multipartitioned_subset(
+        self,
+        partitions_subset: PartitionsSubset,
+        partition_dimension_name: str,
+    ):
```
sryza, line 398: Return type plz

python_modules/dagster/dagster_tests/asset_defs_tests/partition_mapping_tests/test_asset_partition_mappings.py
```diff
@@ -586,3 +592,147 @@ def test_exported_partition_mappings_whitelisted():
...
+        def handle_output(self, context, obj):  # pylint: disable=unused-argument
+            if context.asset_key == AssetKey("upstream"):
+                assert context.has_asset_partitions
+                assert context.asset_partition_keys == ["a"]
+
+        def load_input(self, context):
+            assert context.has_asset_partitions
+            assert context.asset_partition_keys == ["a"]
+
+    upstream_job = build_assets_job(
```
sryza, line 669: Not blocking, but I think we should use `materialize` instead of `build_assets_job` in this kind of test

sryza (approved):
In what situations will users hit an experimental warning? I think it should only be when they manually instantiate a MultiToSingleDimensionPartitionMapping

PR #12401, opened 2023-02-16
Support requiring resources in sensors, experimentally

benpankow:
## Summary

Implements very basic resource support for sensors. Effectively just calls `build_resouces` and passes the results to user code, so we don't get any new observability from this implementation, this would be separately addressed by work that increases sensor observability in general.

Resources can be specified through `required_resource_keys` or through annoated params on the sensor fn. They can be accessed through context or through these params:

```python

    a_str: str
    
@sensor(job_name="the_job", required_resource_keys={"my_resource"})
def sensor_from_context(context: SensorEvaluationContext):
    return RunRequest(context.resources.my_resource.a_str, run_config={}, tags={})


@sensor(job_name="the_job")
def sensor_from_fn_arg(context: SensorEvaluationContext, my_resource: MyResource):
    return RunRequest(my_resource.a_str, run_config={}, tags={})

```

Directly invoking sensors requires that a user only provide the context (for simplicity's sake - alternatively, we could have users directly pass in `my_resource` here if that's preferred).

```python
# Here, `resource_defs` is exploded into the various resource args which `sensor_from_fn_arg` expects 
sensor_from_fn_arg(
    build_sensor_context(
        resource_defs={"my_resource": MyResource(a_str="foo")}
    )
)
```

This PR doesn't handle any of the fancy sensor types (e.g. `multi_asset_sensor`), bumped to a stacked diff so as to keep this diff fairly clean.

## Test Plan

unit test

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/sensor_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/sensor_definition.py
@@ -36,7 +36,9 @@ from dagster._core.instance import DagsterInstance
 from dagster._core.instance.ref import InstanceRef
 from dagster._serdes import whitelist_for_serdes
 
-from ..decorator_utils import get_function_params
+from ..decorator_utils import (
+    get_function_params,  # pylint: disable=unused-import
+)
 from .asset_selection import AssetSelection
 from .graph_definition import GraphDefinition
 from .mode import DEFAULT_MODE_NAME
--- a/python_modules/dagster/dagster/_core/definitions/sensor_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/sensor_definition.py
@@ -24,10 +24,10 @@ from typing_extensions import TypeAlias
 import dagster._check as check
 from dagster._annotations import public
 from dagster._core.definitions.instigation_logger import InstigationLogger
+from dagster._core.definitions.resource_annotation import get_resource_args
 from dagster._core.definitions.resource_definition import (
     Resources,
 )
-from dagster._core.definitions.resource_output import get_resource_args
 from dagster._core.errors import (
     DagsterInvalidDefinitionError,
     DagsterInvalidInvocationError,
--- a/python_modules/dagster/dagster/_core/definitions/sensor_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/sensor_definition.py
@@ -7,6 +7,7 @@ from typing import (
     Callable,
     Iterator,
     List,
+    Mapping,
     NamedTuple,
     Optional,
     Sequence,
@@ -25,7 +26,6 @@ from dagster._core.definitions.resource_definition import (
     Resources,
 )
 from dagster._core.definitions.resource_output import get_resource_args
-from dagster._core.definitions.scoped_resources_builder import ScopedResourcesBuilder
 from dagster._core.errors import (
     DagsterInvalidDefinitionError,
     DagsterInvalidInvocationError,
@@ -49,6 +49,7 @@ from .unresolved_asset_job_definition import UnresolvedAssetJobDefinition
 from .utils import check_valid_name
 
 if TYPE_CHECKING:
+    from dagster import ResourceDefinition
     from dagster._core.definitions.repository_definition import RepositoryDefinition
 
 
@@ -110,8 +111,13 @@ class SensorEvaluationContext:
         repository_def: Optional["RepositoryDefinition"] = None,
         instance: Optional[DagsterInstance] = None,
         sensor_name: Optional[str] = None,
-        resources: Optional[Resources] = None,
+        resource_defs: Optional[Mapping[str, "ResourceDefinition"]] = None,
     ):
+        from dagster._core.definitions.scoped_resources_builder import (
+            IContainsGenerator,
+        )
+        from dagster._core.execution.build_resources import build_resources
+
         self._exit_stack = ExitStack()
         self._instance_ref = check.opt_inst_param(instance_ref, "instance_ref", InstanceRef)
         self._last_completion_time = check.opt_float_param(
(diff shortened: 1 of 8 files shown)


benpankow:
* master
  * **PR #12401** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12401" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12401?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -338,9 +359,16 @@ def __init__(
         self._asset_selection = check.opt_inst_param(
             asset_selection, "asset_selection", AssetSelection
         )
+        resource_arg_names: Set[str] = {arg.name for arg in get_resource_args(self._raw_fn)}
+
+        self._required_resource_keys = (
+            check.opt_set_param(required_resource_keys, "required_resource_keys", of_type=str)
+            | resource_arg_names
+        )
```
schrockn, line 367: Is this the behavior in our other decorators? The union of the parameters and the `required_resourc_keys` arguments?

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -641,16 +680,32 @@ def build_sensor_context(
...
+            instance_ref=None,
+            last_completion_time=None,
+            last_run_key=None,
+            cursor=cursor,
+            repository_name=repository_name,
+            instance=instance,
+            repository_def=repository_def,
+            sensor_name=sensor_name,
+            resources=resources,
+        )
```
schrockn, line 708: at first look this is suspicious seems like `build_sensor_context` should itself be a context manager that is only closed after the tick evaluation is complete? If this is deliberate, at a minimum this is worthy of a comment

python_modules/dagster/dagster/_grpc/impl.py
```diff
@@ -306,17 +306,26 @@ def get_external_sensor_execution(
...
-                repository_def=repo_def,
-                sensor_name=sensor_name,
+        from dagster._core.execution.build_resources import build_resources
+
+        resources_to_build = {
+            k: v
+            for k, v in repo_def.get_top_level_resources().items()
+            if k in sensor_def.required_resource_keys
+        }
+        with build_resources(resources_to_build) as resources:
```
schrockn, line 316: similar comment here.

schrockn (dismissed):
questions around `build_resources` and when it should go out of scope

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
benpankow: 👍  migrated to using a pattern similar to `op_context` to initialize resources

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -366,17 +463,29 @@ def __call__(self, *args, **kwargs):
...
+                context = check.opt_inst_param(
+                    list(kwargs.values())[0], "context", SensorEvaluationContext
                 )
+            context = context if context else build_sensor_context()
 
-            return self._raw_fn()
+        resources = _validate_and_get_resource_dict(
+            context, self.name, self._required_resource_keys
+        )
+        return self._raw_fn(**context_param, **resources)
```
schrockn, line 483: This threw me for a loop a little bit. This relies knowing about an assignment 40 lines up to see that it will actually complete execution and that the context_param dictionary will only include a value if the `if context_param_name:` predicate passes.

I think you should be able to collapse this into a smaller chunk of code (with a single codepath for context creation) that simulateneously feels less clever.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -588,11 +697,18 @@ def wrap_sensor_evaluation(
...
+        resource_args_populated = _validate_and_get_resource_dict(
+            context, sensor_name, resource_arg_names
+        )
+
+        context_param_name = get_context_param_name(fn)
+        if context_param_name:
+            result = fn(**{context_param_name: context}, **resource_args_populated)
         else:
-            result = fn()  # type: ignore
+            result = fn(**resource_args_populated)
```
schrockn, line 711: this also looks like duplicated logic

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -618,6 +734,7 @@ def build_sensor_context(
     repository_name: Optional[str] = None,
     repository_def: Optional["RepositoryDefinition"] = None,
     sensor_name: Optional[str] = None,
+    resource_defs: Optional[Mapping[str, "ResourceDefinition"]] = None,
```
schrockn, line 737: document this in the docblock and specify whether it is this resource_defs or the Definitions resources that take precedences

schrockn (approved):
I definitely think we should enable direct passing of the resource instances, especially for cases where the context isn't required. We can do that in a follow up.

I'm approving this, but I did leave comments where I thought there could be substantial code cleanliness improvements. If you make changes that you think I might have another round of opinions on please reping me on this PR. But I don't want to block progress as I'll be a little slow on turnaround this week.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -618,6 +733,7 @@ def build_sensor_context(
...
```
dpeng817, line 736: Recommend naming this `resources` as opposed to `resource_defs` to match with `build_op_context`, `build_resource_context`, etc

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -231,13 +288,75 @@ def log_key(self) -> Optional[List[str]]:
...
+    fn: Callable, *args: Any, **kwargs: Any
+) -> SensorEvaluationContext:
+    """
+    Based on the passed resource function and the arguments passed to it, returns the
+    user-passed SensorEvaluationContext or creates one if it is not passed.
+
+    Raises an exception if the user passes more than one argument or if the user-provided
+    function requires a context parameter but none is passed.
+    """
+    context_param_name = context_param_name_if_present(fn)
```
schrockn, line 333: btw with the rename suggestion i was referring to the variable name, not the function name. it ends up making code further down make more sense:

`context_param_name = context_param_name or list(kwargs.keys())[0]`

becomes

`context_param_name = context_param_name_if_present or list(kwargs.keys())[0]`

as an example

schrockn (approved):
Please heed final comments but excited to this land.

PR #12403, opened 2023-02-16
Add valid start time check to materialized time partitions subsets

clairelin135:
User reported a negative number of missing partitions in dagit: https://dagster.slack.com/archives/C01U954MEER/p1676368552724739

We originally checked for partition key validity while iterating through time windows in time window subset construction. Recent changes made it possible to add partition keys without iterating through time windows. In backcompat cases for fetching materialized partitions, constructing the subset did not check for validity subsequently causing the materialized partitions subset to contain partition keys that were out of the valid time window range.

This PR makes a fix for this case and adds a test.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/implementation/fetch_assets.py
+++ b/python_modules/dagster-graphql/dagster_graphql/implementation/fetch_assets.py
@@ -335,7 +335,9 @@ def get_materialized_partitions_subset(
     if not partitions_def:
         return None
 
-    if instance.can_cache_asset_status_data() and partitions_def in CACHEABLE_PARTITION_TYPES:
+    if instance.can_cache_asset_status_data() and any(
+        isinstance(partitions_def, partition_type) for partition_type in CACHEABLE_PARTITION_TYPES
+    ):
         # When the "cached_status_data" column exists in storage, update the column to contain
         # the latest partition status values
         updated_cache_value = get_and_update_asset_status_cache_value(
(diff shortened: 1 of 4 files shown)


python_modules/dagster-graphql/dagster_graphql/implementation/fetch_assets.py
```diff
@@ -335,7 +335,9 @@ def get_materialized_partitions_subset(
     if not partitions_def:
         return None
 
-    if instance.can_cache_asset_status_data() and partitions_def in CACHEABLE_PARTITION_TYPES:
+    if instance.can_cache_asset_status_data() and any(
```
clairelin135, line 338: Another bug I discovered along the way: during the `partitions_def in CACHEABLE_PARTITION_TYPES` check, a daily partitions definition != time window partitions definition. This PR fixes this by using an `isinstance` check instead.

OwenKephart: nit: if CACHEABLE_PARTITIONS_TYPE is a tuple, then isinstance(partitions_def, CACHEABLE_PARTITION_TYPES) will do the same thing.

OwenKephart (approved):
looks good, feel free to disregard nit if you feel like it

python_modules/dagster-graphql/dagster_graphql/implementation/fetch_assets.py
clairelin135: ooo this is nice, thanks!

PR #12310, opened 2023-02-13
numFailed on asset partitions

johannkm:

PR #12404, opened 2023-02-16
Get partitions with planned but not completed materializations, the event log query

johannkm:
Fetch the latest materialzation and materialization planned events for each partition of the given asset.
Return the partitions that have a materialization planned event but no matching (same run) materialization event.
These materializations could be in progress, or they could have failed. 

Combined with a second query to check if these runs have failed, we can count how many partitions have have failed to materialize in the latest run.

Followup diff will include this (or more likely the result of checking this with the run status query) in the asset partition cache

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
+++ b/python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
@@ -1705,6 +1705,118 @@ class SqlEventLogStorage(EventLogStorage):
 
         return materialization_count_by_partition
 
+    def get_latest_asset_partition_materialization_attempts_without_materializations(
+        self, asset_key: AssetKey
+    ) -> Mapping[str, str]:
+        """
+        Fetch the latest materialzation and materialization planned events for each partition of the given asset.
+        Return the partitions that have a materialization planned event but no matching (same run) materialization event.
+        These materializations could be in progress, or they could have failed. A separate query checking the run status
+        is required to know.
+
+        Returns a mapping of partition to run_id.
+        """
+        check.inst_param(asset_key, "asset_key", AssetKey)
+
+        latest_event_ids_subquery = (
+            db.select(
+                [
+                    SqlEventLogStorageTable.c.dagster_event_type,
+                    SqlEventLogStorageTable.c.partition,
+                    db.func.max(SqlEventLogStorageTable.c.id).label("id"),
+                ]
+            )
+            .where(
+                db.and_(
+                    db.or_(
+                        SqlEventLogStorageTable.c.asset_key == asset_key.to_string(),
+                        SqlEventLogStorageTable.c.asset_key == asset_key.to_string(legacy=True),
+                    ),
+                    SqlEventLogStorageTable.c.partition != None,  # noqa: E711
+                )
+            )
+            .group_by(
+                SqlEventLogStorageTable.c.dagster_event_type, SqlEventLogStorageTable.c.partition
+            )
+        )
+
+        assets_details = self._get_assets_details([asset_key])
+        latest_event_ids_subquery = self._add_assets_wipe_filter_to_query(
+            latest_event_ids_subquery, assets_details, [asset_key]
+        ).alias("latest_materialization_event_ids")
+
+        latest_events_subquery = (
+            db.select(
+                [
+                    SqlEventLogStorageTable.c.dagster_event_type,
+                    SqlEventLogStorageTable.c.partition,
+                    SqlEventLogStorageTable.c.run_id,
+                ]
+            )
+            .select_from(
+                latest_event_ids_subquery.join(
+                    SqlEventLogStorageTable,
+                    SqlEventLogStorageTable.c.id == latest_event_ids_subquery.c.id,
+                ),
+            )
+            .alias("latest_materialization_events")
+        )
+
+        materialization_planned_events = (
+            db.select(
+                [
+                    latest_events_subquery.c.dagster_event_type,
+                    latest_events_subquery.c.partition,
+                    latest_events_subquery.c.run_id,
+                ]
+            )
+            .where(
+                latest_events_subquery.c.dagster_event_type
+                == DagsterEventType.ASSET_MATERIALIZATION_PLANNED.value
+            )
+            .alias("materialization_planned_events")
+        )
+
+        materialization_events = (
+            db.select(
+                [
+                    latest_events_subquery.c.dagster_event_type,
+                    latest_events_subquery.c.partition,
+                    latest_events_subquery.c.run_id,
+                ]
+            )
+            .where(
+                latest_events_subquery.c.dagster_event_type
+                == DagsterEventType.ASSET_MATERIALIZATION.value
+            )
+            .alias("materialization_events")
+        )
+
+        query = (
+            db.select(
+                [
+                    materialization_planned_events.c.partition,
+                    materialization_planned_events.c.run_id,
+                ]
+            )
+            .select_from(
+                materialization_planned_events.join(
+                    materialization_events,
+                    db.and_(
+                        materialization_planned_events.c.partition
+                        == materialization_events.c.partition,
+                        materialization_planned_events.c.run_id == materialization_events.c.run_id,
+                    ),
+                    isouter=True,
+                )
+            )
+            .where(materialization_events.c.run_id == None)  # noqa: E711
+        )
+
+        with self.index_connection() as conn:
+            rows = conn.execute(query).fetchall()
+            return {row["partition"]: row["run_id"] for row in rows}
+
     def _check_partitions_table(self) -> None:
         # Guards against cases where the user is not running the latest migration for
         # partitions storage. Should be updated when the partitions storage schema changes.
--- a/python_modules/dagster/dagster/_core/storage/legacy_storage.py
+++ b/python_modules/dagster/dagster/_core/storage/legacy_storage.py
@@ -488,6 +488,13 @@ class LegacyEventLogStorage(EventLogStorage, ConfigurableClass):
             asset_keys, after_cursor
         )
 
+    def get_latest_asset_partition_materialization_attempts_without_materializations(
+        self, asset_key: "AssetKey"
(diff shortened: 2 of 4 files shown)


johannkm:
* master
  * **PR #12333** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12333" target="_blank">Graphite</a> 
    * **PR #12404** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12404" target="_blank">Graphite</a>  👈
      * **PR #12405** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12405" target="_blank">Graphite</a> 
    * **PR #12310** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12310" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12404?utm_source=stack-comment).

johannkm:
* master
  * **PR #12333** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12333" target="_blank">Graphite</a> 
    * **PR #12404** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12404" target="_blank">Graphite</a>  👈
      * **PR #12405** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12405" target="_blank">Graphite</a> 
    * **PR #12310** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12310" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12404?utm_source=stack-comment).

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1669,6 +1669,60 @@ def get_materialization_count_by_partition(
 
         return materialization_count_by_partition
 
+    def get_latest_asset_materialization_planned_event_by_partition(
+        self, asset_key: AssetKey
+    ) -> Mapping[str, EventLogRecord]:
```
clairelin135, line 1674: Is there a reason why we need the full event log record? Generally fetching the full event log record is much more expensive than fetching just the asset key and partition key due to deserialization costs.

If we can display everything we need using just the asset key, partition, and storage ID from the row, I think we should try to do that

johannkm:
I propose that instead of having to find and use the ExternalAssetGraph in the create run callsites, we add a field to `StepOutputProperties`. This object already has `is_asset` and `asset_key`  fields that we use to emit the correct materialization planned events, so it seems like a natural place.

python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
```diff
@@ -791,11 +792,19 @@ def resolve_partitionStats(self, graphene_info) -> Optional[GraphenePartitionSta
...
+                self._dynamic_partitions_loader,
+                partitions_def_data.get_partitions_definition(),
+            )
+
             return GraphenePartitionStats(
                 numMaterialized=len(materialized_partition_subset),
                 numPartitions=partitions_def_data.get_partitions_definition().get_num_partitions(
                     dynamic_partitions_store=self._dynamic_partitions_loader
                 ),
+                numFailed=len(failed_partition_subset),
```
clairelin135, line 807: Thoughts on not counting a partition as materialized if it was failed more recently than materialized? I think it might be confusing if a partition can be counted both as materialized and failed. I think if it failed, it should not be counted as "materialized".

clairelin135: On `GrapheneAssetNode` theres also a `materializedPartitions` resolver that resolves the partition keys of the materialized partitions. I think we'll want that resolver to encompass failed partitions, though that can happen in a future PR. It might be easier to group those changes together with the `PartitionStats` changes in a separate PR.

python_modules/dagster-graphql/dagster_graphql_tests/graphql/test_assets.py
```diff
@@ -973,6 +974,42 @@ def test_materialized_default_partitions(self, graphql_context):
...
+        )
+        selector = infer_pipeline_selector(graphql_context, "fail_partition_materialization_job")
+
+        result = execute_dagster_graphql(
+            graphql_context,
+            GET_PARTITION_STATS,
+            variables={"pipelineSelector": selector},
+        )
+        assert result.data
+        print(result.data)
```
clairelin135, line 989: nit: remove print

one more below on line 1006

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1705,6 +1705,60 @@ def get_materialization_count_by_partition(
...
+        check.inst_param(asset_key, "asset_key", AssetKey)
+
+        latest_event_subquery = (
```
clairelin135, line 1713: nit: probs should rename this variable to something like `partitioned_materialization_planned_events` or something like that

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1705,6 +1705,60 @@ def get_materialization_count_by_partition(
...
+                SqlEventLogStorageTable, SqlEventLogStorageTable.c.id == latest_event_subquery.c.id
+            ),
+        )
+
+        with self.index_connection() as conn:
+            event_rows = conn.execute(query).fetchall()
+
+        results = {}
+        for row in event_rows:
+            results[row["partition"]] = EventLogRecord(
```
clairelin135, line 1754: I think we can get away with just representing the storage ID and not the full event log record. Then we can just have one query and not have deserialization costs

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1705,6 +1705,60 @@ def get_materialization_count_by_partition(
...
+                storage_id=row["id"],
+                event_log_entry=cast(
+                    EventLogEntry, deserialize_json_to_dagster_namedtuple(row["event"])
+                ),
+            )
+        return results
```
clairelin135, line 1760: One nuance here is that you can wipe assets--I'd recommend looking at `instance.get_materialization_count_by_partition` to see how it handles wipes. Because some events will be considered "wiped" and shouldn't pop up in future queries

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1705,6 +1705,60 @@ def get_materialization_count_by_partition(
 
         return materialization_count_by_partition
 
+    def get_latest_asset_materialization_planned_event_by_partition(
```
clairelin135, line 1708: This function looks a lot like the `get_materialization_count_by_partition`. I think it's worth refactoring the existing method to also contain functionality from this function.

We don't actually use counts, we just want to figure out if a partition is materialized. I think we'll want to refactor the method to return the mapping from `asset_key, partition_key -> latest storage ID`. And we can accept an `event_type` argument so we can get the latest storage ID per partition for materialization and materialization planned events, which should give us the info we need

python_modules/dagster/dagster/_core/storage/legacy_storage.py
```diff
@@ -488,6 +488,13 @@ def get_materialization_count_by_partition(
             asset_keys, after_cursor
         )
 
+    def get_latest_asset_materialization_planned_event_by_partition(
```
clairelin135, line 491: again would prefer to not have to add new storage method if possible, by refactoring the existing `get_materialization_count_by_partition`

clairelin135 (commented):
This is exciting to see, thanks for making these changes! The changes are generally looking good to me, I left some implementation comments

johannkm:
@clairelin135 updated to be pretty different. The present form doesn't seem to share enough with get_materialization_count_by_partition to be worth refactoring them both I think.

python_modules/dagster/dagster_tests/storage_tests/utils/event_log_storage.py
```diff
@@ -1997,6 +1998,98 @@ def _fetch_counts(storage, after_cursor=None):
                     )
                     assert _fetch_counts(storage, after_cursor=9999999999) == {c: {}, d: {}}
 
+    def test_get_latest_asset_partition_materialization_attempts_without_materializations(
+        self, storage, instance
+    ):
+        a = AssetKey(["a"])
```
clairelin135, line 2004: nit: can we add another test for that adds events for another asset key? so that we can confirm that events from other assets aren't returned

python_modules/dagster/dagster_tests/storage_tests/utils/event_log_storage.py
```diff
@@ -1997,6 +1998,98 @@ def _fetch_counts(storage, after_cursor=None):
...
+        run_id_1 = make_new_run_id()
+        run_id_2 = make_new_run_id()
+        run_id_3 = make_new_run_id()
+        with create_and_delete_test_runs(instance, [run_id_1, run_id_2, run_id_3]):
```
clairelin135, line 2008: can we also add tests for wiped assets and assets that don't have any events?

clairelin135 (approved):
LGTM

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1705,6 +1705,122 @@ def get_materialization_count_by_partition(
...
+                ]
+            )
+            .where(
+                db.and_(
+                    db.or_(
+                        SqlEventLogStorageTable.c.asset_key == asset_key.to_string(),
+                        SqlEventLogStorageTable.c.asset_key == asset_key.to_string(legacy=True),
+                    ),
+                    SqlEventLogStorageTable.c.partition != None,  # noqa: E711
+                    SqlEventLogStorageTable.c.dagster_event_type.in_(
```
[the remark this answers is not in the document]
johannkm: hmm probably not

PR #12407, opened 2023-02-16
RFC: add flag to zero out code location entry timestamps in OSS

prha:
### Summary & Motivation
Users running multiple replicas of dagit are getting annoyed with toasts popping up all the time, because OSS repos are marked with the timestamp at which dagit was loaded.  Depending on which dagit instance is hit on page load, we may end up with multiple `Definitions updated` toasts per page load.

This isn't an issue with Cloud, because the timestamps are loaded from the DB and are stable across dagit instances.

The solution here is to zero out the location status entry timestamps in OSS only, adding a hook to override with the db-read timestamp in Cloud.

Setup an nginx instance, with multiple upstream dagit servers running on different hosts.  Confirmed that the toasts did not show up on page loads.


alangenfeld:
So we have the code location "server id" based detection of change, and this timestamp based thing. Can we flatten the two down in to one and hide the difference in the implementation?

alangenfeld (approved):

PR #12412, opened 2023-02-17
[asset-reconciliation] Factor in more run statuses

OwenKephart:
### Summary & Motivation

There's a time period between a run being created and it starting that we should factor into our calculations. Without this change, runs that are queued by the reconciliation sensor but don't start until after the next tick will not be factored into calculations, resulting in duplicate runs.


sryza (approved):
oo

Issue #12236, opened 2023-02-09
MultiPartitioning Asset not working on Windows  


mananpmutinex:
### Dagster version

1.1.17

### What's the issue?

If an Asset has a MultiPartitionsDefinition, the materialization fails on Windows. 

This seems to be due to the '|' used by dagster for multipartition keys is invalid in Windows file paths.

### What did you expect to happen?

The asset should materialize successfully. 

### How to reproduce?

`from dagster import Definitions, asset, MultiPartitionsDefinition, StaticPartitionsDefinition

# Asset for external tables
@asset( 
    partitions_def=MultiPartitionsDefinition({
        "customer_code": StaticPartitionsDefinition(['abc','youi']),
        "datasource_name": StaticPartitionsDefinition(['tv','search']),
    })
)
def external_table(context):
    parition_keys = context.log.info(context.partition_key.keys_by_dimension)
    return_val = f"materialized external table for {parition_keys=}"
    context.log.info(return_val)
    return return_val

defs = Definitions(
    assets= [external_table]
)`

Trying to Materialize the 'abc|tv' partition for this asset through the UI will cause a failure. 

### Deployment type

Local

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

sryza:
@clairelin135 heads up I believe this is a duplicate of https://github.com/dagster-io/dagster/issues/11865

pdavidsonFIA:
multi partition runs fine on my windows pc (as well as my team) -> but we're using a custom IOManager so maybe it is due to the default fs io manager

Issue #11865, opened 2023-01-26
fs_io_manager doesn't work with multidimensional partitions on Windows


feljost:
### Dagster version

dagster, version 1.1.13

### What's the issue?

Using multidimensional partitions for assets and returning a value does not work as intended.

While the operations inside the asset can run correctly, adding a return to the end of the asset function leads to an error.

Take this example asset:
```
@asset(
    name="test_asset",
    partitions_def=MultiPartitionsDefinition(
        {
            "date": MonthlyPartitionsDefinition(start_date="2022-01-01"),
            "color": StaticPartitionsDefinition(["red", "blue"]),
        }
    ),
)
def test_asset(context: OpExecutionContext) -> int:
    context.log.info(context.partition_key)
    return 5
```

Trying to materialize the Asset in the Dagit UI will lead to the following error:

```
dagster._core.errors.DagsterExecutionHandleOutputError: Error occurred while handling output "result" of step "test_asset":
  File "d:\sandbox\felix\test-library\env\lib\site-packages\dagster\_core\execution\plan\execute_plan.py", line 265, in dagster_event_sequence_for_step
    for step_event in check.generator(step_events):
  File "d:\sandbox\felix\test-library\env\lib\site-packages\dagster\_core\execution\plan\execute_step.py", line 388, in core_dagster_event_sequence_for_step
    for evt in _type_check_and_store_output(step_context, user_event, input_lineage):
  File "d:\sandbox\felix\test-library\env\lib\site-packages\dagster\_core\execution\plan\execute_step.py", line 444, in _type_check_and_store_output
    for evt in _store_output(step_context, step_output_handle, output, input_lineage):
  File "d:\sandbox\felix\test-library\env\lib\site-packages\dagster\_core\execution\plan\execute_step.py", line 631, in _store_output
    for elt in iterate_with_context(
  File "d:\sandbox\felix\test-library\env\lib\site-packages\dagster\_utils\__init__.py", line 459, in iterate_with_context
    return
  File "D:\Program Files\Python38\lib\contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
  File "d:\sandbox\felix\test-library\env\lib\site-packages\dagster\_core\execution\plan\utils.py", line 85, in op_execution_error_boundary
    raise error_cls(

The above exception was caused by the following exception:

[...]

PR #12414, opened 2023-02-17
Fix multipartitions w/ fs_io_manager on windows

clairelin135:
Fixes an error where multipartitions cannot be used with the fs_io_manager on windows. This occurs because windows cannot store files containing the pipe (`|`) character which are used in multipartition keys:
- https://github.com/dagster-io/dagster/issues/11865
- https://github.com/dagster-io/dagster/issues/12236 (duplicate)

This PR resolves this by storing each dimension in its own directory level following the alphabetical order of the dimension names. For backcompatibility, if attempting to read from the new file directory format results in a file not found error, attempts to read from the old file directory format.


python_modules/dagster/dagster/_core/storage/upath_io_manager.py
```diff
@@ -96,16 +97,58 @@ def _get_paths_for_partitions(
...
+    def _load_single_input(
+        self, path: UPath, context: InputContext, backcompat_path: Optional[UPath] = None
+    ) -> Any:
         context.log.debug(f"Loading file from: {path}")
-        obj = self.load_from_path(context=context, path=path)
+        try:
+            obj = self.load_from_path(context=context, path=path)
+        except FileNotFoundError as e:
+            if backcompat_path is not None:
+                context.log.debug(
```
sryza, line 144: I think that we'd want to surface the backcompat path as little as possible in logs and exceptions, to avoid confusing people. Can we make it so that they only see a log if we successfully load from the backcompat path? And that, if we fail to load from either, the path that comes up in the stack trace is the non-backcompat path?

sryza (commented):
Nice

python_modules/dagster/dagster/_core/storage/upath_io_manager.py
clairelin135: Thanks for the feedback--this makes a lot of sense. This PR has now been updated to follow this behavior.

The error we now raise if no file exists at either path is the non-backcompat path. We only display the log message now if the file is found at the backcompat path.

sryza (approved):

PR #12418, opened 2023-02-17
[library template] add registry call

alangenfeld:
update to comply with new library registry 

```
python scripts/create_dagster_package.py -n dagster-testing-gen
pip install -e python_modules/libraries/dagster-testing-gen
python> import dagster_testing_gen
```

### DIFF
--- a/scripts/create_dagster_package.py
+++ b/scripts/create_dagster_package.py
@@ -83,6 +83,13 @@ def _make_dagster_package(package_name: str):
                 "underscore_name": package_name_underscore,
             },
         },
+        "__init__.py": {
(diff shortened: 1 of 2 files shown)


jamiedemaria (approved):
nice! thanks

alangenfeld:
is there a good way to test it? or are you confident in this from inspection

jamiedemaria:
not a great way to test rn other than making a fake package and looking that the files are correct. the template looks right to me though

do you have ideas on testing? i'm not sure what we'd do other than automating creating a fake package, and maybe running pytest on it?

alangenfeld:
test idea
* run it in to a temp dir 
* pip install -e it
* verify you can import it

PR #12419, opened 2023-02-17
fix: resolve correct legacy arguments for emr pyspark step launcher

rexledesma:
### Summary & Motivation
Compare `deploy_local_job_package` with `deploy_local_pipeline_package`, as intended.

pytest


rexledesma:
* master
  * **PR #12419** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12419" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12419?utm_source=stack-comment).

dpeng817 (approved):
Looks reasonable but I think at this point we should just completely get rid of the legacy `pipeline` args here, we declared EOL on all things pipeline in the 0.17 version of this package so I think it's reasonable to do so. Will leave that up to your discretion tho

Issue #12325, opened 2023-02-14
Dagster Helm chart should support the full set of things that you can configure as environment variables in Dagit/Daemon/other system pods


gibsondan:
### What's the use case?

The 'env' key here can only take in a dictionary of key-value env pairs: https://github.com/dagster-io/dagster/blob/master/helm/dagster/values.yaml#L84-L93

But k8s generally allows this key to be a list of env vars sourced from other places: https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1EnvVar.md

This issue is a request to be able to configure dagit (and other Dagster helm components like this)

```
dagit:
  env:
    - name: 'DD_ENTITY_ID',
       valueFrom:
         fieldRef:
           apiVersion: v1
           fieldPath: metadata.uid
```

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

PR #12425, opened 2023-02-17
Support env valueFrom in Helm chart

johannkm:
For https://github.com/dagster-io/dagster/issues/12325

The kubernetes spec for env vars is

```
    env:
    - name: DEMO_GREETING
      value: "Hello from the environment"
    - name: 'DD_ENTITY_ID',
       valueFrom:
         fieldRef:
           apiVersion: v1
           fieldPath: metadata.uid
  - name: S3_BUCKET
    valueFrom:
      configMapKeyRef:
        name: s3-settings
        key: bucket
```

we decided to simplify our helm chart to

```
env:
  ENV_ONE: one
  ENV_TWO: two
```

so we can't support the fancy `valueFrom`s.

**OLD**
~~This PR adds a new `envList` field for the rare users who need `valueFrom`~~. We could also try and fit this to our existing env field, but I think that's unnecessarily complicated. It'd need to accept Union[Dict,List] and fork based on the type: for lists we currently make a configmap instead of putting the env vars directly in the spec. `valueFrom` vars would need to be put directly on the spec to support the `fieldPath` args.

**NEW**
This PR makes `env` take either a list (the new format) or a dict (the old). If it's a dict, we add the env vars to the configmap like we have been. If it's a list, we add it directly to the container spec. This way we keep backcompat for current users with `env`, but we also support the full k8s schema for the field.


johannkm:
* master
  * **PR #12425** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12425" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12425?utm_source=stack-comment).

johannkm:
need to copy over to dagit and userDeployment but put up for review in case there's pushback about `envList`

helm/dagster/values.yaml
```diff
@@ -1124,6 +1124,8 @@ dagsterDaemon:
   #   - name: secret
   envSecrets: []
 
+  envList: []
```
johannkm, line 1127: could also be podEnv or containerEnv since it goes directly on the pod instead of configmap -> pod? But that seems more confusing

gibsondan:
"We could also try and fit this to our existing env field, but I think that's unnecessarily complicated"

Would it be a lot of work to make it behave differently if it's a list or a dict? It seems like a big gotcha that most of our fields behave just like the k8s spec but this one behaves subtly differently (for no particularly good reason that I can see, I guess it makes the syntax slightly less verbose)

Maybe we go with this for now but do a breaking change in 1.2 to make env more like the actual k8s spec?

gibsondan:
It's definitely a little strange but in my head it was deprecating the old strange dict way while leaving it around for back-compat (then we remove it eventually)

gibsondan (dismissed):
to your queue

helm/dagster/values.yaml
```diff
@@ -314,7 +314,7 @@ dagster-user-deployments:
       # env:
       #   ENV_ONE: one
       #   ENV_TWO: two
-      env: {}
```
gibsondan, line 317: gotta update the comment here too right?

gibsondan (approved):
Sweet, this works. The tricky part is I think the user who asked for this also wanted it to work for the user code deployments? Which would ideally mean the resulting config would also get passed through to any launched runs, which also likely means putting it on k8scontainercontext etc...

johannkm:
The specific request was for dagit and the daemon, but yeah we'll want it in runs soon enough

PR #11217, opened 2022-12-19
Dumb prototype of pending job

schrockn:

PR #12430, opened 2023-02-17
Bind top-level resources to jobs at Definitions-time

benpankow:
## Summary

Enables users to specify resources for jobs at the top-level in a `Definitions` object. Defers validation of a job's resources until repo initialization time. This is a pretty naive first-pass at doing so, there may be demons that this approach doesn't acknowledge.

Users may continue to specify resources at the job level. If they do so, they are validated immediately, as before.

```python
class WriterResource(ConfigurableResource):
    prefix: str

    def output(self, text: str) -> None:
        out_txt.append(f"{self.prefix}{text}")

@op
def hello_world_op(writer: WriterResource):
    writer.output("hello, world!")

# No resource binding needed here - if resource is not provided at `Definitions` time, error thrown
@job
def hello_world_job():
    hello_world_op()

defs = Definitions(
    jobs=[hello_world_job],
    resources={
        "writer": WriterResource(prefix="message: "),
    },
)
```

New unit tests & existing tests.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/definitions_class.py
+++ b/python_modules/dagster/dagster/_core/definitions/definitions_class.py
@@ -1,4 +1,14 @@
-from typing import TYPE_CHECKING, Any, Iterable, Mapping, Optional, Sequence, Type, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Iterable,
+    Mapping,
+    NamedTuple,
+    Optional,
+    Sequence,
+    Type,
+    Union,
+)
 
 import dagster._check as check
 from dagster._annotations import experimental, public
@@ -85,6 +95,65 @@ def create_repository_using_definitions_args(
     )
 
 
+class _AttachedObjects(NamedTuple):
+    jobs: Iterable[Union[JobDefinition, UnresolvedAssetJobDefinition]]
+    schedules: Iterable[Union[ScheduleDefinition, UnresolvedPartitionedAssetScheduleDefinition]]
+    sensors: Iterable[SensorDefinition]
+
+
+def _attach_resources_to_jobs(
+    jobs: Optional[Iterable[Union[JobDefinition, UnresolvedAssetJobDefinition]]],
+    schedules: Optional[
+        Iterable[Union[ScheduleDefinition, UnresolvedPartitionedAssetScheduleDefinition]]
+    ],
+    sensors: Optional[Iterable[SensorDefinition]],
+    resource_defs: Mapping[str, Any],
+) -> _AttachedObjects:
+    """
+    Given a list of jobs, schedules, and sensors along with top-level resource definitions,
+    attach the resource definitions to the jobs, schedules, and sensors which require them.
+    """
+    jobs = jobs or []
+    schedules = schedules or []
+    sensors = sensors or []
+
+    # Find unsatisfied jobs
+    unsatisfied_jobs = [
+        job
+        for job in jobs
+        if isinstance(job, JobDefinition) and job.is_missing_required_resources()
+    ]
+
+    # Create a mapping of job id to a version of the job with the resource defs bound
+    unsatisfied_job_to_resource_bound_job = {
+        id(job): job.with_top_level_resources(resource_defs)
+        for job in jobs
+        if job in unsatisfied_jobs
+    }
+
+    # Update all jobs to use the resource bound version
+    jobs_with_resources = [
+        unsatisfied_job_to_resource_bound_job[id(job)] if job in unsatisfied_jobs else job
+        for job in jobs
+    ]
+
+    # Update all schedules and sensors to use the resource bound version
+    updated_schedules = [
+        schedule.with_updated_job(unsatisfied_job_to_resource_bound_job[id(schedule.job)])
+        if (isinstance(schedule, ScheduleDefinition) and schedule.job in unsatisfied_jobs)
+        else schedule
+        for schedule in schedules
+    ]
+    updated_sensors = [
+        sensor.with_updated_job(unsatisfied_job_to_resource_bound_job[id(sensor.job)])
+        if sensor.job in unsatisfied_jobs
+        else sensor
+        for sensor in sensors
+    ]
+
+    return _AttachedObjects(jobs_with_resources, updated_schedules, updated_sensors)
+
+
 def _create_repository_using_definitions_args(
     name: str,
     assets: Optional[
@@ -133,14 +202,11 @@ def _create_repository_using_definitions_args(
 
     check.opt_mapping_param(loggers, "loggers", key_type=str, value_type=LoggerDefinition)
 
-    jobs_with_resources = (
-        [
-            job.with_top_level_resources(resource_defs) if isinstance(job, JobDefinition) else job
-            for job in jobs
-        ]
-        if jobs
-        else []
-    )
+    (
+        jobs_with_resources,
+        schedules_with_resources,
+        sensors_with_resources,
+    ) = _attach_resources_to_jobs(jobs, schedules, sensors, resource_defs)
 
     @repository(
         name=name,
@@ -151,8 +217,8 @@ def _create_repository_using_definitions_args(
     def created_repo():
         return [
             *with_resources(assets or [], resource_defs),
-            *(schedules or []),
-            *(sensors or []),
+            *(schedules_with_resources),
+            *(sensors_with_resources),
             *(jobs_with_resources),
         ]
 
--- a/python_modules/dagster/dagster/_core/definitions/definitions_class.py
+++ b/python_modules/dagster/dagster/_core/definitions/definitions_class.py
@@ -101,7 +101,7 @@ class _AttachedObjects(NamedTuple):
     sensors: Iterable[SensorDefinition]
 
 
-def _attach_resources_to_jobs(
+def _attach_resources_to_jobs_and_instigators(
     jobs: Optional[Iterable[Union[JobDefinition, UnresolvedAssetJobDefinition]]],
     schedules: Optional[
         Iterable[Union[ScheduleDefinition, UnresolvedPartitionedAssetScheduleDefinition]]
@@ -210,7 +210,7 @@ def _create_repository_using_definitions_args(
         jobs_with_resources,
         schedules_with_resources,
         sensors_with_resources,
-    ) = _attach_resources_to_jobs(jobs, schedules, sensors, resource_defs)
+    ) = _attach_resources_to_jobs_and_instigators(jobs, schedules, sensors, resource_defs)
 
     @repository(
         name=name,
--- a/python_modules/dagster/dagster/_core/definitions/definitions_class.py
+++ b/python_modules/dagster/dagster/_core/definitions/definitions_class.py
@@ -133,6 +133,16 @@ def _create_repository_using_definitions_args(
 
     check.opt_mapping_param(loggers, "loggers", key_type=str, value_type=LoggerDefinition)
 
+    jobs_with_resources = (
+        [
+            job.with_top_level_resources(resource_defs)
+            for job in jobs
+            if isinstance(job, JobDefinition)
+        ]
+        if jobs
+        else []
+    )
+
     @repository(
         name=name,
         default_executor_def=executor_def,
@@ -144,7 +154,7 @@ def _create_repository_using_definitions_args(
             *with_resources(assets or [], resource_defs),
             *(schedules or []),
             *(sensors or []),
-            *(jobs or []),
+            *(jobs_with_resources),
         ]
 
     return created_repo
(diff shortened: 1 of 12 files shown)


benpankow:
* master
  * **PR #12430** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12430" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12430?utm_source=stack-comment).

python_modules/dagster/dagster_tests/definitions_tests/test_repository_definition.py
```diff
@@ -376,7 +376,10 @@ def needy(context):
     def bare():
         needy()
 
-    with pytest.raises(DagsterInvalidDefinitionError, match="Failed attempting to coerce Graph"):
+    with pytest.raises(
```
benpankow, line 379: these error messages change because the `Failed attempting to coerce Graph` is thrown in the except of a try/except guard around job creation - since the `DagsterInvalidDefinitionError` w/ missing resources only fires at repo init time, the Graph error isn't shown

python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -85,6 +95,65 @@ def create_repository_using_definitions_args(
     )
 
 
+class _AttachedObjects(NamedTuple):
+    jobs: Iterable[Union[JobDefinition, UnresolvedAssetJobDefinition]]
+    schedules: Iterable[Union[ScheduleDefinition, UnresolvedPartitionedAssetScheduleDefinition]]
+    sensors: Iterable[SensorDefinition]
+
+
+def _attach_resources_to_jobs(
```
dpeng817, line 104: probably rename this to `_attach_resources_to_defs` or something

dpeng817 (commented):
Would be nice to see a few more cases under test just to have behavior defined:
- what happens when resource keys conflict between job and definitions object?
- what happens if two instigators (schedule / sensor) previously targeted the same job, but now have resources bound? (I suspect this will cause an unwanted error)

benpankow:
> * what happens when resource keys conflict between job and definitions object?

If I'm interpreting this right, the job takes precedence, see https://github.com/dagster-io/dagster/pull/12430/files#diff-5fc8121949b7c3fde1c820d9deb32c2a01b4b852cf0091fab3b9c8fc8cf86fbaR1124 for a test case (one job has an override, one doesn't)

> * what happens if two instigators (schedule / sensor) previously targeted the same job, but now have resources bound? (I suspect this will cause an unwanted error)

I think this will work actually since we call `with_top_level_resources` only once per job object (e.g. this mapping https://github.com/dagster-io/dagster/pull/12430/files#diff-b9703e2b8757e40cf8a844844e89ed0b9f7a338dae712c9773113841df79008eR127). This test has a schedule + sensor bound to the same job: https://github.com/dagster-io/dagster/pull/12430/files#diff-5fc8121949b7c3fde1c820d9deb32c2a01b4b852cf0091fab3b9c8fc8cf86fbaR1191 (again I may be misinterpreting the q)

dpeng817:
Ah nice - not sure how I missed those tests but looks reasonable!

python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -85,6 +95,69 @@ def create_repository_using_definitions_args(
...
+    # Find unsatisfied jobs
+    unsatisfied_jobs = [
+        job
+        for job in jobs
+        if isinstance(job, JobDefinition) and job.is_missing_required_resources()
+    ]
+
+    # Create a mapping of job id to a version of the job with the resource defs bound
+    unsatisfied_job_to_resource_bound_job = {
+        id(job): job.with_top_level_resources(resource_defs)
```
dpeng817, line 129: i like this honestly - it's direct about the fact that we use reference equality for jobs.

dpeng817 (commented):
One other test case I'm now thinking of - instigator uses name targeting schema, target doesn't have all resources bound, ensuring that the definition that ends up on the schedule / sensor has all resources bound

dpeng817 (commented):
Implementation here looks reasonable to me - my only question is what the strategy for rolling something like this out should be? It's not exactly a breaking change, but it is a pretty big behavioral change that I'm not sure we want to silently roll into a minor release

benpankow:
> my only question is what the strategy for rolling something like this out should be? It's not exactly a breaking change, but it is a pretty big behavioral change that I'm not sure we want to silently roll into a minor release

Yeah, there are cases where this adjusts how errors appear. The core loading flow of existing jobs should be unaltered since `is_missing_required_resources` (which gates the new  `_attach_resources_to_jobs` load flow) only returns `True` on jobs missing resource definitions which would error at construction-time presently. It's still a bit scary to touch and I'd be curious for others' input.

schrockn:
@benpankow thanks for taking this one.

In terms of implementation, I think there is a fairly fundamental question to answer here (and I'm sure @sryza) has opinions on this: whether or not to represent a job in an non-executable state as the same type or not. I prototyped this a while back, and my approach was to add a `PendingJobDefinition` type (see https://github.com/dagster-io/dagster/pull/11217) and then opt _in_ the APIs where we accept the pending variant (e.g. in `Definitions` and as targets of `ScheduleDefinition` and `SensorDefinition`). This matches the mental of `UnresolvedAssetJobDefinition` which is another instance of a thing that turns into a fully executable job later with information only available at the Definitions/repository layer.

The advantages of that approach is that there is a fairly wide surface area of APIs in the system (both public and internal) that take `JobDefinition` as argument, and currently it is assumed that they are directly executable. Whether or not an API can take an unresolved job is encoded in types, rather than implicit assumptions about the state of `JobDefinition`.

I'd like Sandy's take on this, but my inclination is to instead add an `UnresolvedJobDefinition`. The jobs argument to `Definitions` and the target argument to schedules and sensors _already_ takes a mix of real JobDefinitions and the unresolved jobs that are only resolved later. This PR adds an entire different strategy for unresolved jobs which I think will be confusing. 

We'll also need a content rollout strategy for this and whether or not we will drive all our content and examples towards late-binding jobs. My inclination is towards mirroring our mental model to assets where the common case is binding at the Definitions layer but allowing advanced APIs to bind specific jobs as needed. This also calls into question how we will treat and interrelate `@job` and `@graph` as with this change they become ever closer to eachother, to the point where we should consider allowing jobs to be nested in other jobs and make graph an internal implementation detail of the system.

So to summarize, three top-level questions to answer here before proceeding here is

1) Adding a new type or having a dual state JobDefinition
2) What our rollout strategy and recommendation for usage will be in the 1.3 timeframe.
3) What is the future of `@job` and `@graph` with this change.

We'll need at least loose alignment on those questions. Ideally we would get at least loose alignment on those three questions, and then ship this experimentally in 1.2.

schrockn (dismissed):
Requesting changes for queue mgmt and to do get @sryza feedback

sryza:
Oy yeah this is a question that has plagued my mind for years, but I don't have a strong stance.  A couple questions:
- Assuming we went the `UnresolvedJobDefinition` route, what are the user-facing APIs that would accept an `JobDefinition` but not an `UnresolvedJobDefinition`? Is it basically just `execute_in_process` and `execute_job`, or are there a bunch?
- Assuming we went the `UnresolvedJobDefinition` route, would we have an `@unresolved_job` decorator? Or would `@job` spit out different types depending on whether all the resource defs were satisfied? If the latter, it seems like a big chunk of the user-facing value of having separate types gets diluted.

dpeng817:
I remember having conversations about this year(s) ago and one thing that often came up was how rough the user experience would be to see `Union[JobDefinition, PendingJobDefinition]` in the API docs - our decorator types are already hard to manage, that could make it worse arguably

dpeng817:
> If the latter, it seems like a big chunk of the user-facing value of having separate types gets diluted.
@sryza +1 on this - if the point of adding `PendingJobDefinition` is to not change behavior for `@job` / `JobDefinition` users then I think we have to create a separate entrypoint unless I'm misunderstanding, and that experience seems rough

schrockn:
@sryza 

> Assuming we went the UnresolvedJobDefinition route, what are the user-facing APIs that would accept an JobDefinition but not an UnresolvedJobDefinition? Is it basically just execute_in_process and execute_job, or are there a bunch?

There are certainly tons of _internal_ callsites that require a bound job. That is definitely a consideration IMO and it leaks out to public APIs, both present and future.

> Assuming we went the UnresolvedJobDefinition route, would we have an @unresolved_job decorator? Or would @job spit out different types depending on whether all the resource defs were satisfied? If the latter, it seems like a big chunk of the user-facing value of having separate types gets diluted.

In my mind this was definitely `@job` returning a Union rather than an additional decorator.

> I remember having conversations about this year(s) ago and one thing that often came up was how rough the user experience would be to see Union[JobDefinition, PendingJobDefinition] in the API docs 

@dpeng817   `Union[JobDefinition, PendingJobDefinition]` super rough and `Union[RepositoryDefinition, PendingRepositoryDefinition, _Repository]` (return type of `@repository`) is even worse. However the "toothpaste is out of the tube" on this so I think adding another type is just marginally worse.

The other thing that we have discussed here is the implications on `@graph` and `@job` going forward as this narrows the distance between those two APIs considerably. Has anyone thought deeply about that issue?

sryza:
One more thought on the `UnresolvedJobDefinition` stuff: I think it would be inconvenient if users had to use `cast` to get code like the following to pass type checks.
```python
@job
def my_job_that_requires_no_resources():
    ...

my_job_that_requires_no_resources.execute_in_process()
```

> The other thing that we have discussed here is the implications on `@graph` and `@job` going forward as this narrows the distance between those two APIs considerably. Has anyone thought deeply about that issue?

@benpankow and I were chatting a little about this at lunch last week. Definitely somewhat gnarly, and I haven't fully formulated my thoughts.  If there was no such thing as an asset job, I would say that we should seriously think about whether jobs and graphs can be consolidated.  With the existence of assets though, I think jobs are a useful abstraction layer that provides a single interface to two different things that can be run, scheduled, and monitored: op graphs and fixed asset selections.

schrockn:
> One more thought on the UnresolvedJobDefinition stuff: I think it would be inconvenient if users had to use cast to get code like the following to pass type checks.

Yeah also inconvenient that if the job is missing resources it will pass type checks and then error at runtime 🤷. I'm a bit conflicted here because this class of type errors with `Union[RepositoryDefinition, PendingRepository]` was extremely annoying when coding against that layer.

In terms of a path forward here I am fine with not introducing another type. I just wanted to chat it through to ensure that we are all aligned here on the tradeoffs.

Now the question is: how hard do we want to push people to this in our content and examples. Is this the "default" way to use the system?

dpeng817:
> I would say that we should seriously think about whether jobs and graphs can be consolidated
@sryza what about the graph-backed assets case? I feel like it would feel pretty weird to package an asset over a job as opposed to a graph.

sryza:
> @sryza what about the graph-backed assets case? I feel like it would feel pretty weird to package an asset over a job as opposed to a graph.

I agree that it feels weird. There's also a bunch of other stuff that feels weird - like that jobs have a partitions_def, executor_def, and loggers.  However, it's not clear to me that it's soo weird that it merits a separate concept.

Btw "If there was no such thing as an asset job, I would say that we should seriously think about whether jobs and graphs can be consolidated." - emphasis on **seriously think**, not saying we would definitely want to do this.

benpankow:
Per discussion earlier today, going to get this ready to merge.

python_modules/dagster/dagster/_core/definitions/repository_definition/repository_data_builder.py
```diff
@@ -367,6 +370,10 @@ def build_caching_repository_data_from_dict(
                 f"Object mapped to {key} is not an instance of JobDefinition or GraphDefinition."
             )
 
+    for pipeline_or_job in repository_definitions["jobs"].values():
+        if isinstance(pipeline_or_job, PipelineDefinition):
+            pipeline_or_job.validate_resource_requirements_satisfied()
+
```
schrockn, line 376: what is going on here

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
```diff
@@ -429,6 +429,25 @@ class ScheduleDefinition:
         required_resource_keys (Optional[Set[str]]): The set of resource keys required by the schedule.
     """
 
+    def with_updated_job(self, new_job: ExecutableDefinition) -> "ScheduleDefinition":
+        """Returns a copy of this schedule with the job replaced.
+
+        Args:
+            job (Union[GraphDefinition, JobDefinition]): The job that should execute when this
```
schrockn, line 436: just say `ExecutableDefinition`?

schrockn (commented):
couple comments

schrockn:
hopping into meeting but will look more later

python_modules/dagster/dagster/_core/definitions/repository_definition/repository_data_builder.py
benpankow: This is running the job-resource validation late, since it can't be run at job-creation time. Will add a comment.

schrockn (approved):
Alrighty here we go

PR #12431, opened 2023-02-17
Improve multipartition performance for `get_partition`

clairelin135:
Fixes https://github.com/dagster-io/dagster/issues/12441

User with 90K partitions in a multipartitions definition was attempting to yield 50 multipartitioned run requests within a schedule, causing timeouts in sensor evaluation.
slow_schedule_eval

Within `run_request_for_partition`, calling `partition_set.get_partition` invokes `get_partitions` to fetch the matching partition. Calling `get_partitions` this many times on a multipartitions definition becomes expensive with 90K partitions and causes the timeout in the flamegraph above.

This PR adds a performance improvement for this by adding a `get_partition` method on `PartitionsDefinition`, with `MultiPartitionsDefinition` subclassing this method and fetching the partition by dimension to avoid fetching all partitions.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
@@ -19,6 +19,7 @@ from dagster._annotations import experimental
 from dagster._core.errors import (
     DagsterInvalidDefinitionError,
     DagsterInvalidInvocationError,
+    DagsterUnknownPartitionError,
 )
 from dagster._core.instance import DynamicPartitionsStore
 from dagster._core.storage.tags import (
@@ -32,6 +33,7 @@ from .partition import (
     PartitionsDefinition,
     PartitionsSubset,
     StaticPartitionsDefinition,
+    Partition,
 )
 from .time_window_partitions import TimeWindowPartitionsDefinition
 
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
```diff
@@ -210,6 +212,37 @@ def partition_dimension_names(self) -> List[str]:
...
+        partitions_by_dimension: Dict[str, Partition] = {}
+        for dimension in self.partitions_defs:
+            partition = dimension.partitions_def.get_partition(
+                partition_key.keys_by_dimension[dimension.name],
+                current_time=current_time,
+                dynamic_partitions_store=dynamic_partitions_store,
+            )
+            if not partition:
+                check.failed(
+                    "Invalid partition key {partition_key}. The partition key does not exist in the"
```
sryza, line 239: Missing an f-string?

sryza (approved):
I think we should ultimately get rid of `Partition` and `get_partition` entirely, because `Partition` is part of the pre-CRAG partitions world

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
clairelin135: good catch

PR #12432, opened 2023-02-17
[dagster-dbt] Add ability to use the "state:" selector

OwenKephart:
### Summary & Motivation

The "state:modified" selector method is a fairly powerful dbt construction, which allows you to execute models based on if they have been modified since the last time the project has been compiled.

However, it relies on the previous manifest being available. This pr allows the user to specify a state_path, which will point to wear that previous state should live.

It also moves towards a world where we just supply manifest filepaths, rather than parsed json objects. This is nice because we can let dbt handle the reading/parsing of the manifest file into structured objects, rather than trying to do it ourselves.

Factored out the select_unique_ids_from_manifest into its own utility function.


sryza (commented):
A concern here, if I understand correctly how this is working:
- dbt typically resolves the state selector at the time that the dbt run is being executed
- We resolve the asset selection at the time the code location is loaded

It seems like this could lead to some unexpected / undesirable behavior?

OwenKephart:
@sryza in order for there to be a difference in the resolved selection between those two points in time (code location loading and `dbt run` being executed), there'd need to be a meaningful difference in the manifest.json file between those two points in time, as state is purely a function of the manifest.json file (plus some other compiled artifacts).

Barring some weird exceptions (i.e. generating your model's sql based on queries to the database or something like that), there's no reason to expect that the manifest.json that gets compiled when executing `dbt run` will differ from the file that was generated at code deployment time (if there were, then `load_assets_from_dbt_manifest` would not be a reliable way to load dbt assets).

sryza:
Doesn't `dbt run` overwrite the manifest?

OwenKephart:
It overwrites it, but with an essentially identical artifact if the project hasn't changed (some timestamps and stuff may be different)

sryza (approved):

PR #12433, opened 2023-02-17
[asset-reconciliation][perf] Do not get/set known_used_data in asset reconciliation loop

OwenKephart:
### Summary & Motivation

Even though it seems like this would save time, the per-query overhead is a lot. If you have many freshness policies, reading / writing the asset_event_tags table a bunch of times is not really all that efficient, you'd rather just do the query from scratch.

It's still nice to do this in the frontend codepath, as that's evaluated one node at a time.


sryza (approved):
Mind adding documentation for this parameter? Also, if I understand correctly, the parameter impacts whether we do writes? But the name doesn't exactly communicate that? Is there a name that would communicate that better?

PR #12434, opened 2023-02-19
[dagit] Use range-based asset health for asset partitions / job partitions pages

bengotow:
### Summary & Motivation

This PR moves the remaining bits of Dagit to use range-based asset health data in the most direct form possible.

This is likely to be a big performance win, especially for the asset job partitions page which loaded and very-inefficiently-queried the health data of all the assets in the job.  ( It was using a translation layer of sorts mapping the ranges to a structure that was then queried per-partition to count partitions, etc. )

One caveat of this move to range-based health is that we had to assemble a utility library for slicing / subsetting ranges, counting keys in a set of ranges with a given status, etc. This PR adds a few more (hopefully the last of them), and I'm writing exhaustive test coverage for them because the logic can be complex.

I think the range-based implementation is much more efficient, but it's harder to grok two dimensional ranges windowed to a two dimensional selection than it was to understand + visualize a 2D matrix of booleans.  I think there's a chance this could get /real/ complicated if/when we layer "stale" into asset health, but at least we'll have strong test coverage.

15 new tests and a bunch of updated ones! The tests and storybooks share mocks. This PR adds:

- Storybooks for AssetPartitions

- Basic tests that the AssetPartitions page renders and interactions work:
   + Slicing the partition space via the top text input when a time partition is present
   + Filtering the partitions based on status and seeing the counts / list change
   + Clicking partitions to set focus

- Exhaustive tests for the new utility methods

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/AssetPartitions.test.tsx
+++ b/js_modules/dagit/packages/core/src/assets/AssetPartitions.test.tsx
@@ -1,0 +1,158 @@
+import {MockedProvider} from '@apollo/client/testing';
+import {act, render, screen, waitFor} from '@testing-library/react';
+import userEvent, {specialChars} from '@testing-library/user-event';
+import React from 'react';
+import {Route} from 'react-router-dom';
+
+import {AssetKeyInput} from '../graphql/types';
+import {TestProvider} from '../testing/TestProvider';
+
+import {AssetPartitionListProps} from './AssetPartitionList';
+import {AssetPartitions} from './AssetPartitions';
+import {AssetViewParams} from './AssetView';
+import {
+  SingleDimensionStaticPartitionHealthQuery,
+  SingleDimensionTimePartitionHealthQuery,
+} from './PartitionHealthSummary.mocks';
+
+// This file must be mocked because useVirtualizer tries to create a ResizeObserver,
+// and the component tree fails to mount. We still want to test whether certain partitions
+// are shown, so we print the keys in a simple "list".
+jest.mock('./AssetPartitionList', () => ({
+  AssetPartitionList: (props: AssetPartitionListProps) => (
+    <div>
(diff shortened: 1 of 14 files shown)


js_modules/dagit/packages/core/src/assets/AssetPartitions.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+    <div>
+      <div data-testid="focused-partition">{props.focusedDimensionKey}</div>
+      {props.partitions.slice(0, 50).map((p) => (
+        <div key={p} onClick={() => props.setFocusedDimensionKey?.(p)}>
+          {p}
+        </div>
+      ))}
+    </div>
+  ),
+}));
```
hellendag, line 32: 👍

js_modules/dagit/packages/core/src/assets/AssetPartitions.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+const SingleDimensionAssetPartitions: React.FC<{assetKey: AssetKeyInput}> = ({assetKey}) => {
+  const [params, setParams] = React.useState<AssetViewParams>({});
+  return (
+    <TestProvider>
```
hellendag, line 37: Is the full `TestProvider` definitely needed for your tests? I've been trying to avoid it where possible when using `MockedProvider`, since `TestProvider` includes the full Apollo mocked schema and everything else. No big deal either way though.

hellendag (approved):

js_modules/dagit/packages/core/src/assets/AssetPartitions.test.tsx
bengotow: Ahh I think I can do without, will change these! I think it's really just the workspace context that ends up weirdly required in these.

PR #12444, opened 2023-02-21
[CustomConfirmationDialog] Allow overriding the button text

salazarm:
### Summary & Motivation

I'd like to override the button text for my usecase.

used the feature:
Screen Shot 2023-02-21 at 12 24 45 PM

### DIFF
--- a/js_modules/dagit/packages/core/src/app/CustomConfirmationProvider.tsx
+++ b/js_modules/dagit/packages/core/src/app/CustomConfirmationProvider.tsx
@@ -7,7 +7,7 @@ interface ConfirmationOptions {
   description?: React.ReactNode;
   icon?: ConfirmationDialogProps['icon'];
   intent?: ConfirmationDialogProps['intent'];
-  buttonText?: string;
+  buttonText?: React.ReactNode;
 }
 
 interface ConfirmationDialogProps extends ConfirmationOptions {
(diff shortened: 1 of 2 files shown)


js_modules/dagit/packages/ui/src/components/TextInput.tsx
```diff
@@ -56,7 +56,7 @@ export const TextInputContainerStyles = css`
   position: relative;
 `;
 
-const Container = styled.div<{$disabled: boolean}>`
+export const TextInputContainer = styled.div<{$disabled: boolean}>`
```
salazarm, line 59: Also exporting this because I've found myself needing to customize the styling of this input (eg. to make the input 100% width)

js_modules/dagit/packages/core/src/app/CustomConfirmationProvider.tsx
```diff
@@ -7,6 +7,7 @@ interface ConfirmationOptions {
   description?: React.ReactNode;
   icon?: ConfirmationDialogProps['icon'];
   intent?: ConfirmationDialogProps['intent'];
+  buttonText?: string;
```
hellendag, line 10: `React.ReactNode` instead of string, since that's the `Button` child type?

hellendag (approved):

PR #12445, opened 2023-02-21
[pythonic resources] Clean up initialization of env vars, treat resource objects as immutable

benpankow:
## Summary

Restructures the way that we initialize `ConfigurableResource` objects at runtime. Now, they are immutable, and we construct new versions of the resource when we resolve environment variables and initialize resources.

This makes the code cleaner and safer, and ensures that we handle `EnvVar` resolution properly, including in nested data structures (lists, dicts, config objects). Previously, we'd evaluate `EnvVar`s at load time, meaning we'd error on loading the code location. This was a bad experience and differed from current behavior.

## Test Plan

Add a whole bunch of unit tests to ensure env vars work in various places & are only checked at runtime.


benpankow:
* master
  * **PR #12445** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12445" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12445?utm_source=stack-comment).

python_modules/dagster/dagster/_config/field_utils.py
```diff
@@ -459,20 +459,31 @@ def _convert_potential_field(
...
-        if value is None:
-            continue
-
-        if isinstance(value, EnvVar):
-            new_values[key] = {"env": str(value)}
-        else:
-            new_values[key] = value
-    return new_values
+
+    return cast(Dict[str, Any], _config_dictionary_from_values_inner(values))
```
schrockn, line 486: This is definitely a case where I think we should do a runtime invariant check that pairs with the cast to prevent coding errors, both now and in the future.

python_modules/dagster/dagster/_config/structured_config/__init__.py
```diff
@@ -402,10 +401,38 @@ def configure_at_launch(cls: "Type[Self]", **kwargs) -> "PartialResource[Self]":
...
-    def initialize_and_run(self, context: InitResourceContext) -> TResValue:
-        # If we have any partially configured resources, we need to update them
-        # with the fully configured resources from the context
+    def _with_updated_values(self, values: Mapping[str, Any]) -> "ConfigurableResource[TResValue]":
+        """
+        Returns a new instance of the resource with the given values.
+        Used when initializing a resource at runtime.
+        """
+        return self.__class__(
+            **{**{k: v for k, v in self.__dict__.items() if not k.startswith("_")}, **values}
```
schrockn, line 410: now I feel especially strongly wrapping this since `startswith` appears twice in this PR

python_modules/dagster/dagster/_config/field_utils.py
```diff
@@ -459,20 +459,31 @@ def _convert_potential_field(
...
+    elif isinstance(obj, list):
+        return [_config_dictionary_from_values_inner(v) for v in obj]
+    elif isinstance(obj, EnvVar):
+        return {"env": str(obj)}
+    elif isinstance(obj, Config):
+        return {
+            k: _config_dictionary_from_values_inner(v)
+            for k, v in obj.dict().items()
+            if not k.startswith("_")
+        }
```
schrockn, line 476: also to be clear this change is sort of separate from the title (refactoring resources to be immutable), as it expands support of the coercion to include lists and strongly typed config objects, correct?

python_modules/dagster/dagster/_config/structured_config/__init__.py
```diff
@@ -402,10 +401,38 @@ def configure_at_launch(cls: "Type[Self]", **kwargs) -> "PartialResource[Self]":
...
 
-    def initialize_and_run(self, context: InitResourceContext) -> TResValue:
-        # If we have any partially configured resources, we need to update them
-        # with the fully configured resources from the context
+    def _with_updated_values(self, values: Mapping[str, Any]) -> "ConfigurableResource[TResValue]":
+        """
+        Returns a new instance of the resource with the given values.
+        Used when initializing a resource at runtime.
+        """
+        return self.__class__(
```
schrockn, line 409: interesting I've never seen this pattern (construction using `self.__class__`). It's a bit odd because it imposes an unencoded constraint on the class that the __init__ function must be this signature. Right now I *believe* this is essentially enforced by pydantic (in my experimentation it was very difficult to override `__init__`) but we don't know if that will be true forever.

Stack overflow says that behavior is different between `type(self)` and `self.__class__` (https://stackoverflow.com/questions/1060499/difference-between-typeobj-and-obj-class/10633356#10633356) and it _seems_ like `type` is what you want here.

At a minimum, this seems worth embellish with comments.

schrockn (changes requested):

python_modules/dagster/dagster/_config/field_utils.py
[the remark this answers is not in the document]
benpankow: Moved to a new helper method 👍

benpankow: Yes, exactly. Added a comment to this effect, and updated PR description just to make it more clear.

python_modules/dagster/dagster/_config/structured_config/__init__.py
benpankow: My slight bias is towards using `__class__` since subclasses shouldn't be overriding the `__class__` descriptor and if they are, they should have good reason for doing so. We use that pattern elsewhere, [here](https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_core/definitions/assets.py#L687) and [here](https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_core/definitions/output.py#L183) but not the `type(self)` pattern.

Will add some comments.

benpankow: Just to hopefully assuage some of the concern about the `__init__` function - since `BaseModel` is a Python datamodel, I believe the signature of the `__init__` method is fixed to be reflective of its fields. Users can override it to add in their own behavior, but the signature basically needs to remain the same.

python_modules/dagster/dagster/_config/field_utils.py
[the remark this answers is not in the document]
benpankow: done 👍

python_modules/dagster/dagster/_config/structured_config/__init__.py
```diff
@@ -101,6 +99,14 @@ class Config(MakeConfigCacheable):
...
     """
 
+    @property
+    def config_dictionary(self) -> Mapping[str, Any]:
+        """
+        Returns a dictionary representation of this config object,
+        ignoring any private fields.
+        """
+        return {k: v for k, v in self.__dict__.items() if not k.startswith("_")}
+
```
schrockn, line 109: seems like a dangerous precedent to make this a public without prefix. Recommend a naming convention similar to `_asdict()` on named tuples. How about `_as_config_dict`?

schrockn (approved):
Please heed final comment re: `config_dictionary`, but otherwise looks great!

image

PR #12446, opened 2023-02-21
[pythonic config] Special rendering for env vars in UI

benpankow:
## Summary

Adds a basic special renderer for env vars in the Dagit resources page.

Screen Shot 2023-02-21 at 11 11 55 AM

## Test Plan

Add unit test, test locally.

### DIFF
--- a/js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
+++ b/js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
@@ -101,7 +101,10 @@ export const ResourceRoot: React.FC<Props> = (props) => {
           }
 
           const configuredValues = Object.fromEntries(
-            topLevelResourceDetailsOrError.configuredValues.map((cv) => [cv.key, cv.value]),
+            topLevelResourceDetailsOrError.configuredValues.map((cv) => [
+              cv.key,
+              {value: cv.value, type: cv.type},
+            ]),
           );
 
           return (
@@ -123,12 +126,16 @@ export const ResourceRoot: React.FC<Props> = (props) => {
                       <tbody>
                         {topLevelResourceDetailsOrError.configFields.map((field) => {
                           const defaultValue = field.defaultValueAsJson;
+                          const type =
+                            field.name in configuredValues
+                              ? configuredValues[field.name].type
+                              : null;
                           const actualValue =
                             field.name in configuredValues
-                              ? configuredValues[field.name]
+                              ? configuredValues[field.name].value
                               : defaultValue;
 
-                          const isDefault = defaultValue === actualValue;
+                          const isDefault = type === 'VALUE' && defaultValue === actualValue;
 
                           return (
                             <tr key={field.name}>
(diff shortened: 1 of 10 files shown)


benpankow:
* master
  * **PR #12445** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12445" target="_blank">Graphite</a> 
    * **PR #12446** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12446" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12446?utm_source=stack-comment).

js_modules/dagit/packages/core/src/resources/ResourceRoot.tsx
```diff
@@ -146,9 +151,10 @@ export const ResourceRoot: React.FC<Props> = (props) => {
                                   content={<>Default: {defaultValue}</>}
                                   canShow={!isDefault}
                                 >
-                                  {actualValue}
+                                  {type === 'ENV_VAR' ? <Tag>{actualValue}</Tag> : actualValue}
                                 </Tooltip>
                                 {isDefault && <Tag>Default</Tag>}
+                                {type === 'ENV_VAR' && <Tag intent="success">Env var</Tag>}
```
salazarm, line 157: Is this intentional?

salazarm (commented):
js lgtm

python_modules/dagster/dagster/_core/host_representation/external_data.py
```diff
@@ -912,7 +920,10 @@ def __new__(
             ),
             configured_values=dict(
                 check.mapping_param(
-                    configured_values, "configured_values", key_type=str, value_type=str
+                    configured_values,
+                    "configured_values",
+                    key_type=str,
+                    value_type=(str, ExternalResourceConfigEnvVar),
```
schrockn, line 926: If I am reading this correct, we typehint as `Dict[str, ExternalResourceValue]` but check `value_type=(str, ExternalResourceConfigEnvVar)` for backwards compat purposes.

Definitely threw me for a second because I thought you were committing a breaking change here. I'd prefer an approach with an explicit serializer here.

schrockn (changes requested):
Let's use a serializer.

python_modules/dagster/dagster/_core/host_representation/external_data.py
benpankow: Hmm, could you elaborate on what you are looking for here? Right now the `configured_values` dict can take either strings (representing normal values) or `ExternalResourceConfigEnvVar` (representing env vars) with `ExternalResourceValue` as an alias for the union of the two.

Would you prefer that we don't use raw `str` for normal values?

benpankow:
re-request for clarification

python_modules/dagster/dagster/_core/host_representation/external_data.py
schrockn: Oh I see I was confused. Also my bad on the slow turnaround here.

schrockn (approved):

PR #12456, opened 2023-02-21
clarify error when op is missing argument for In

sryza:
### Summary & Motivation

The second sentence makes it sound like the source of the error is that the user included something they shouldn't have, when it's actually that they left something out that they should have included.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/op_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/op_decorator.py
@@ -416,10 +416,10 @@ def resolve_checked_solid_fn_inputs(
     if not has_kwargs and undeclared_inputs:
         undeclared_inputs_printed = ", '".join(undeclared_inputs)
         raise DagsterInvalidDefinitionError(
-            f"{decorator_name} '{fn_name}' decorated function does not have parameter(s) "
-            f"'{undeclared_inputs_printed}', which are in provided input_defs. {decorator_name} "
-            "decorated functions should only have keyword arguments that match input names and, if "
-            "system information is required, a first positional parameter named 'context'."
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster/_core/definitions/decorators/op_decorator.py
```diff
@@ -417,9 +417,9 @@ def resolve_checked_solid_fn_inputs(
         undeclared_inputs_printed = ", '".join(undeclared_inputs)
         raise DagsterInvalidDefinitionError(
             f"{decorator_name} '{fn_name}' decorated function does not have parameter(s) "
-            f"'{undeclared_inputs_printed}', which are in provided input_defs. {decorator_name} "
-            "decorated functions should only have keyword arguments that match input names and, if "
-            "system information is required, a first positional parameter named 'context'."
+            f"'{undeclared_inputs_printed}', which are in provided ins. {decorator_name}-"
```
dpeng817, line 420: `in provided ins` hurts a bit ngl

python_modules/dagster/dagster/_core/definitions/decorators/op_decorator.py
```diff
@@ -417,9 +417,9 @@ def resolve_checked_solid_fn_inputs(
...
+            "decorated functions should have a keyword argument for each In name, except for Ins "
```
dpeng817, line 421: sidenote but it's not true that they should have a keyword argument for each `In` right? It could be a non-kwarg too

dpeng817 (approved):
Fine with this but comments for your consideration

sryza:
Thanks for the suggestions @dpeng817 . I incorporated them.

alangenfeld:
Just to make sure you caught it - this is only the direct invocation code path - the composition evaluation is separate so you may want to ensure the error message for that is updated as well.

PR #12460, opened 2023-02-22
Reimplement more helpful asset key mismatch errors using rapidfuzz

benpankow:
## Summary

Reimplements #12008 (which was reverted in #12459 after fossa failure) using `rapidfuzz`, which is a MIT licensed CPP reimpl & faster.

Unfortunately,

> On Windows the Visual C++ 2019 redistributable is required

Which might be a deal breaker to using this lib.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/resolved_asset_deps.py
+++ b/python_modules/dagster/dagster/_core/definitions/resolved_asset_deps.py
@@ -1,5 +1,7 @@
 from collections import defaultdict
-from typing import AbstractSet, Dict, Iterable, List, Mapping, Tuple, cast
+from typing import AbstractSet, Dict, Iterable, List, Mapping, Sequence, Tuple, cast
+
+from rapidfuzz import fuzz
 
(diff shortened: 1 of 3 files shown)


benpankow:
* master
  * **PR #12460** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12460" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12460?utm_source=stack-comment).

sryza:
> Which might be a deal breaker to using this lib.

What's the verdict on this? Not something I'm super familiar with this. If it adds a huge dependency on Windows, it's probably not worth it.

shalabhc:
It looks like they do [publish](https://pypi.org/project/rapidfuzz/#files) Windows wheels on PyPI. However I'd feel better if we could test this on a real Windows machine. In fact we should also probably test on an M1 Mac, given this is not a pure python dependency.

PR #12462, opened 2023-02-22
[graphql] empty numPartitions and partitionNames for cross-partitioning backfills

sryza:
### Summary & Motivation

This PR makes a handful of backfill fields optional, because things like `numPartitions` and `partitionNames` are not always available for pure asset backfills with partition mapping. The number of partitions is /usually/ available (unless we were unable to determine the relationships between the partition mappings), but the partition names are never available.

@bengotow UI Changes: On the front-end this required making a few things in the UI optional - most notably you cannot click the partition count badge to see the partitions requested, and the run status column is blank. 

To make sure this functionality doesn't get broken in the future and ensure we preview what these cases look like when changing the backfill table, I added two new cases to the storybook. (Bottom two in screenshot!)

image

### DIFF
--- a/js_modules/dagit/packages/core/src/instance/BackfillRow.tsx
+++ b/js_modules/dagit/packages/core/src/instance/BackfillRow.tsx
@@ -78,17 +78,21 @@ export const BackfillRow = ({
   const statusUnsupported = backfill.numPartitions === null || backfill.partitionNames === null;
 
   // Note: We switch queries based on how many partitions there are to display,
-  // because the detail is nice for small backfills but breaks for 100k+ partitions
-  const [queryStatus, queryResult] = statusUnsupported
+  // because the detail is nice for small backfills but breaks for 100k+ partitions.
+  //
+  // If the number of partitions or partition names are missing, we use a mock to
+  // avoid executing any query at all. This is a bit awkward, but seems cleaner than
+  // making the hooks below support an optional query function / result.
+  const [statusQueryFn, statusQueryResult] = statusUnsupported
     ? NoBackfillStatusQuery
     : (backfill.numPartitions || 0) > BACKFILL_PARTITIONS_COUNTS_THRESHOLD
     ? statusCounts
     : statusDetails;
 
-  useDelayedRowQuery(queryStatus);
-  useQueryRefreshAtInterval(queryResult, FIFTEEN_SECONDS);
+  useDelayedRowQuery(statusQueryFn);
+  useQueryRefreshAtInterval(statusQueryResult, FIFTEEN_SECONDS);
 
-  const {data} = queryResult;
+  const {data} = statusQueryResult;
   const {counts, statuses} = React.useMemo(() => {
     if (data?.partitionBackfillOrError.__typename !== 'PartitionBackfill') {
       return {counts: null, statuses: null};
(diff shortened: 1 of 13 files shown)


bengotow:
Hey @sryza thanks for making these changes! Going to comandeer this and push the front-end updates to your branch so the build will pass 👋

js_modules/dagit/packages/core/src/instance/BackfillRow.tsx
```diff
@@ -70,10 +75,15 @@ export const BackfillRow = ({
     },
   );
 
+  const statusUnsupported = backfill.numPartitions === null || backfill.partitionNames === null;
+
   // Note: We switch queries based on how many partitions there are to display,
   // because the detail is nice for small backfills but breaks for 100k+ partitions
-  const [queryStatus, queryResult] =
-    backfill.numPartitions > BACKFILL_PARTITIONS_COUNTS_THRESHOLD ? statusCounts : statusDetails;
+  const [queryStatus, queryResult] = statusUnsupported
```
hellendag, line 82: This is confusing -- `queryStatus` is the _function_, and `queryResult` is the data? These variable names are awfully similar. Could they be renamed to clarify this? Since the values are separated from the `useLazyQuery` calls themselves, it's a little tough to understand what these values actually are.

js_modules/dagit/packages/core/src/instance/BackfillRow.tsx
```diff
@@ -37,6 +37,11 @@ import {
 } from './types/BackfillRow.types';
 import {BackfillTableFragment} from './types/BackfillTable.types';
 
+const NoBackfillStatusQuery = [
+  () => Promise.resolve({data: undefined} as QueryResult<undefined>),
+  {data: undefined, called: true, loading: false} as QueryResult<undefined>,
+] as const;
```
hellendag, line 43: This is kind of unusual.

js_modules/dagit/packages/core/src/instance/BackfillRow.tsx
```diff
@@ -124,18 +134,18 @@ export const BackfillRow = ({
         />
       </td>
       <td style={{width: 140}}>
-        {counts ? (
+        {counts || statusUnsupported ? (
           <BackfillStatusTag backfill={backfill} counts={counts} />
         ) : (
-          <LoadingOrNone queryResult={queryResult} />
+          <LoadingOrNone queryResult={queryResult} noneString="" />
```
hellendag, line 140: Could there be some text instead of an empty string? Maybe a `–`, or something that indicates what's missing?

hellendag (approved):

js_modules/dagit/packages/core/src/instance/BackfillRow.tsx
bengotow: Yeah I'm not a fan of this :( I want there to be a case where we never call the per-row function at all, but it didn't seem worth going through and making the `useDelayedRowQuery` and `useQueryRefreshAtInterval` accept an optional query object, so I passed in a no-op query instead.

I'll add a detailed comment here for now :-/

js_modules/dagit/packages/core/src/instance/BackfillRow.tsx
bengotow: Yeah good call - I renamed this `[statusQueryFn, statusQueryResult]` which seems better!

js_modules/dagit/packages/core/src/instance/BackfillRow.tsx
bengotow: Ahh I like that - it looks like we use {'\u2013'} elsewhere for a dash, will use that here!

bengotow:
Test failures are unrelated, going to merge this

PR #12464, opened 2023-02-22
Speed up pyright BK with pins for venv

smackesey:
### Summary & Motivation

Changes to how pyright runs on BK to speed up build time:

- Add a `--pinned-deps` mode to the `run-pyright.py` script that builds a venv using `requirements-pinned.txt` instead of the looser `requirements.txt`.
- Add `make pinned_pyright` that runs the script in `--pinned-deps` mode.
- Change BK config to invoke `make pinned_pyright`.

The `requirements-pinned.txt` files will need to be manually maintained for each env, but it's pretty simple and should only be needed infrequently-- just build the pyright venv normally, `pip freeze > requirements-pinned.txt` in the env, and change the `git` requirements from the editable installs into plain relative paths (in the style of `requirements.txt`).

Looked at pyright BK step times-- reduction from ~14 min to ~8 min.


smackesey:
* master
  * **PR #12464** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12464" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12464?utm_source=stack-comment).

alangenfeld:
In what conditions are we going to need to rebuild the lock file / pinned deps? 

I think at a minimum we should have a cli entry & documentation somewhere (in the CLI help output maybe?) for updating the lock file / pinned deps. 

I think we should probably default to using the pinned deps when local users are doing an install too to save them time. 

Would it make sense to switch a python environment manager tool that has direct support for lock files? Ideally we can offload some complexity and undiscovered cornercases to a mature system if possible.

smackesey:
>In what conditions are we going to need to rebuild the lock file / pinned deps?

If dependencies of one of our packages change in a "significant" way-- that would manifest as runtime tests of functionality using some new dep API passing, but pyright giving a false positive error (since it's using an old version of the dep).

> I think at a minimum we should have a cli entry & documentation somewhere (in the CLI help output maybe?) for updating the lock file / pinned deps.

Excellent idea-- added.

>I think we should probably default to using the pinned deps when local users are doing an install too to save them time.

Another good idea. Changed it so that it uses pinned by default, but if you pass `--update-pins` then it will rebuild the venv with the looser requirements and update `requirements-pinned.txt`.

>Would it make sense to switch a python environment manager tool that has direct support for lock files? Ideally we can offload some complexity and undiscovered cornercases to a mature system if possible.

It might. I have limited experience with both [Poetry](https://python-poetry.org/) and [Pipenv](https://pipenv.pypa.io/), and found them both buggy and unpolished (pipenv especially) at the time. I suspect it would be more trouble than its worth to introduce one of these (or another option?) just for pyright, but I am open and think it might make sense as part of a broader CI overhaul at some point.

alangenfeld:
What about adding entirely new dependencies? Just trying to understand how often any given eng is going to have to know about this script and run it.

smackesey:
>What about adding entirely new dependencies?

That would count as a significant dependency change. Basically if a call to a new third-party function is made, or, for a `py.typed` dep, if a call is made using  new parameters that aren't in the older versions type signature.

alangenfeld (approved):
I am worried about how people will know that they need to take this action based on how the error manifests. If you feel good about this I am willing to see how it plays out in practice.

smackesey:
>I am worried about how people will know that they need to take this action based on how the error manifests

That's reasonable. I added some instructions that will show up if pyright fails and there is at least one missing import error (which is going to be the error 95% of the time) when the pyright env is using an old version of deps:

```

At least one error was caused by a missing import. This is often caused by changing package depnedencies.
If you have added dependencies to an existing package, run `make rebuild_pyright_pins` to rebuild and update the dependencies of the pyright venv.
If you have added an entirely new package, add it to pyright/master/requirements.txt and then run `make rebuild_pyright_pins`.

PR #12466, opened 2023-02-22
[For 1.2] Allow both protobuf 3 and 4 in dagster

gibsondan:
Summary:
This removes the protobuf3 pin in dagster and re-generates our protobuf api with protobuf 4 (which is mostly backwards compatible with protobuf3, see below)

Two downsides of this:
- we need to bump our min grpcio version to 1.44 since that's the first version that is protobuf3 and protobuf4 compatible, and our min version of protobuf a bit higher. This leaves our grpcio pin fairly narrow until we can get the upper bound removed once they fix the upstream issues.
- it nukes our pyright typehints for the generated modules

But users have been clamoring for us to remove this protobuf pin so I think it's worth it


python_modules/dagster/dagster/_grpc/client.py
```diff
@@ -189,12 +189,12 @@ def _streaming_query(
 
     def ping(self, echo: str):
         check.str_param(echo, "echo")
-        res = self._query("Ping", api_pb2.PingRequest, echo=echo)
+        res = self._query("Ping", api_pb2.PingRequest, echo=echo)  # type: ignore
```
alangenfeld, line 192: what exactly are the type errors being suppressed here?

alangenfeld (commented):
hard for me to forecast what impact narrowing grpcio from 1.32.0 (Sep 9, 2020) to 1.44.0 (Feb 17, 2022) would have. Min version over a year old seems reasonable I think.

gibsondan:
I'll definitely feel better about bumping the min grpcio pin if we can get the upper bound removed.. 1.44-1.47 is a bit narrow

python_modules/dagster/setup.py
```diff
@@ -74,12 +74,12 @@ def get_version() -> str:
         "alembic>=1.2.1,!=1.6.3,!=1.7.0",
         "croniter>=0.3.34",
         # grpcio>=1.48.1 has hanging/crashing issues: https://github.com/grpc/grpc/issues/30843 and https://github.com/grpc/grpc/issues/31885
-        # ensure version we require is >= that with which we generated the grpc code (set in dev-requirements)
-        "grpcio>=1.32.0,<1.48.1",
-        "grpcio-health-checking>=1.32.0,<1.44.0",
+        # grpcio 1.44.0 is the min version compatible with both protobuf 3 and 4
+        "grpcio>=1.44.0,<1.48.1",
+        "grpcio-health-checking>=1.44.0",
```
gibsondan, line 79: re - the pins here, one thing i'm not sure about: this new min version is what's needed assuming you want things to keep working if you upgrade to protobuf4, but if you leave the installed protobuf version at 3, you can keep it back at grpcio 1.32.0

i.e. these combos are valid after this PR:
grpcio 1.32.0 / protobuf 3.20.0
grpcio 1.48.0 / protobuf 3.20.0
grpcio 1.48.0 / protobuf 4.22.0

So we could technically leave the >=1.32.0 but that could still result in a combo of libraries that satisfies our constraints but doesn't load correctly

python_modules/dagster/dagster/_grpc/client.py
alangenfeld: bump

python_modules/dagster/dagster/_grpc/__generated__/api_pb2.py
```diff
@@ -10,2446 +10,145 @@
...
-PingRequest = _reflection.GeneratedProtocolMessageType(
-    "PingRequest",
-    (_message.Message,),
-    {
-        "DESCRIPTOR": _PINGREQUEST,
-        "__module__": "api_pb2"
-        # @@protoc_insertion_point(class_scope:api.PingRequest)
-    },
-)
-_sym_db.RegisterMessage(PingRequest)
```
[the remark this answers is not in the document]
gibsondan: @alangenfeld ^^ this was the response to your bump

python_modules/dagster/dagster/_grpc/client.py
gibsondan: See below

alangenfeld: brutal - wonder what else we can do to manage this better - maybe our own layer of indirection if the mentioned external libraries are not usable.

alangenfeld (approved):

PR #12473, opened 2023-02-22
[op] error on multiple values for argument

alangenfeld:
Via debugging a user report we didnt catch this correctly.
In normal python we get
```
def foo(x, y):
    print(x, y)


bar = partial(foo, x=4)
bar(5)

====
TypeError: foo() got multiple values for argument 'x'
```

so emulate that error message

added test


alangenfeld:
* master
  * **PR #12473** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12473" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12473?utm_source=stack-comment).

OwenKephart (approved):

PR #12474, opened 2023-02-22
[dagster-fivetran]: Add op_tags to build_fivetran_assets

Sedosa:
### Summary & Motivation
Dagster supports the addition of custom k8s configurations to ops and jobs via tags. In an environment where a custom configuration is required to be attached e.g. Kubernetes Pods, the current `build_fivetran_assets` can't support attaching these tags to the underlying Op, causing the job to fail.

In this PR I'm adding an optional `op_tags` parameter to support use case

Adding a unit test to asset op_tags are as expected


benpankow (approved):
sweet!

PR #12475, opened 2023-02-22
[typing] fix typing in daemon tests

dpeng817:
Adds typing to fixtures used in daemon tests.

### DIFF
--- a/python_modules/dagster/dagster_tests/daemon_tests/conftest.py
+++ b/python_modules/dagster/dagster_tests/daemon_tests/conftest.py
@@ -1,18 +1,25 @@
 import os
 import sys
+from typing import Generator, Optional, cast
 
 import pytest
-from dagster._core.host_representation import InProcessRepositoryLocationOrigin
+from dagster import DagsterInstance
+from dagster._core.host_representation import (
+    ExternalRepository,
+    InProcessRepositoryLocationOrigin,
+    RepositoryLocation,
+)
 from dagster._core.test_utils import (
     InProcessTestWorkspaceLoadTarget,
     create_test_daemon_workspace_context,
     instance_for_test,
 )
 from dagster._core.types.loadable_target_origin import LoadableTargetOrigin
+from dagster._core.workspace.context import WorkspaceProcessContext
 
 
 @pytest.fixture(name="instance_module_scoped", scope="module")
-def instance_module_scoped_fixture():
+def instance_module_scoped_fixture() -> Generator[DagsterInstance, None, None]:
     with instance_for_test(
         overrides={
             "run_launcher": {
@@ -25,7 +32,7 @@ def instance_module_scoped_fixture():
 
 
 @pytest.fixture(name="instance", scope="function")
-def instance_fixture(instance_module_scoped):
+def instance_fixture(instance_module_scoped) -> Generator[DagsterInstance, None, None]:
     instance_module_scoped.wipe()
     instance_module_scoped.wipe_all_schedules()
     yield instance_module_scoped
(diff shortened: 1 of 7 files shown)


python_modules/dagster/dagster_tests/daemon_tests/test_monitoring_daemon.py
```diff
@@ -135,34 +141,50 @@ def test_monitor_starting(instance, logger):
...
+    assert run
+    assert run.status == DagsterRunStatus.FAILURE
 
 
-def test_monitor_started(instance, workspace, logger):
+def test_monitor_started(
+    instance: DagsterInstance, workspace_context: WorkspaceProcessContext, logger: Logger
+):
     run = create_run_for_test(instance, pipeline_name="foo", status=DagsterRunStatus.STARTED)
+    workspace = workspace_context.create_request_context()
```
dpeng817, line 153: seems like an `IWorkspaceContext` instead of an `IWorkspace` was being used incorrectly here.

python_modules/dagster/dagster_tests/daemon_tests/conftest.py
```diff
@@ -1,18 +1,25 @@
...
     create_test_daemon_workspace_context,
     instance_for_test,
 )
 from dagster._core.types.loadable_target_origin import LoadableTargetOrigin
+from dagster._core.workspace.context import WorkspaceProcessContext
 
 
 @pytest.fixture(name="instance_module_scoped", scope="module")
-def instance_module_scoped_fixture():
+def instance_module_scoped_fixture() -> Generator[DagsterInstance, None, None]:
```
smackesey, line 22: Prefer `Iterator[Foo]` to `Generator[Foo, None, None]`-- also applies to other instances in this PR

python_modules/dagster/dagster_tests/daemon_tests/conftest.py
```diff
@@ -41,21 +48,26 @@ def workspace_load_target(attribute=None):
...
 
 
 @pytest.fixture(name="external_repo", scope="module")
-def external_repo_fixture(workspace_context):
-    return next(
-        iter(workspace_context.create_request_context().get_workspace_snapshot().values())
-    ).repository_location.get_repository("the_repo")
+def external_repo_fixture(
+    workspace_context: WorkspaceProcessContext,
+) -> Generator[ExternalRepository, None, None]:
```
smackesey, line 61: Looks like the type you are casting to and the return type don't match?

python_modules/dagster/dagster_tests/daemon_tests/test_backfill.py
```diff
@@ -701,25 +743,29 @@ def test_pure_asset_backfill(instance, workspace_context, external_repo):
...
+    runs = reversed(list(instance.get_runs()))
     for run in runs:
         assert run.tags[BACKFILL_ID_TAG] == "backfill_with_asset_selection"
         assert step_succeeded(instance, run, "foo")
         assert step_succeeded(instance, run, "reusable")
         assert step_succeeded(instance, run, "bar")
     # selected
     for asset_key in asset_selection:
-        assert len(instance.run_ids_for_asset_key(asset_key)) == 3
+        assert len(list(instance.run_ids_for_asset_key(asset_key))) == 3
```
smackesey, line 754: Prob negligible performance change but there is `len_iter` for these cases:

https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_utils/__init__.py#L731-L734

python_modules/dagster/dagster_tests/daemon_tests/conftest.py
dpeng817: I'm returning an attribute of the thing I'm casting, not the thing I'm casting

python_modules/dagster/dagster_tests/daemon_tests/test_backfill.py
dpeng817: oo nice

python_modules/dagster/dagster_tests/daemon_tests/conftest.py
smackesey: ah I see now, was confused by too-narrow diff window

PR #12489, opened 2023-02-23
make setting the log level in grpc servers apply to all loggers, not just dagster/dagit

gibsondan:
Summary:
There's a "--log-level" argument to 'dagster api grpc', but right now it only applies to a set of handpicked dagster loggers. This PR changes it so that it applies to all loggers, which I think is more in line with what somebody would expect when they set that flag.

Test Plan:
Load a repo that makes a logging.info() call
Before, it would do nothing even if you ran it with "dagster dev --code-server-log-level INFO" - now it does
Verify that dagster log messages are unchanged in dagit


alangenfeld (approved):
k

PR #12382, opened 2023-02-15
refactor(databricks): lift polling methods up to the client

rexledesma:
### Summary & Motivation
Remove the unnecessary duplication.

### How I Tested These Changes
pytest

PR #12377, opened 2023-02-15
refactor(databricks): use databricks_cli's raw api client

rexledesma:
### Summary & Motivation
Use `XXXService` when possible to start removing the trivial uses of `DatabricksApi`. Also, remove unnecessary indrection in `DatabricksClient`.

### How I Tested These Changes
pytest

PR #12366, opened 2023-02-15
refactor(databricks): consolidate types

rexledesma:
### Summary & Motivation
We shouldn't need to export our internal representations about the Databricks API.

### How I Tested These Changes
pytest

PR #12364, opened 2023-02-15
refactor(databricks): add types to databricks.py

rexledesma:
### Summary & Motivation
Wrangle some of the complexity into types.

### How I Tested These Changes
pyright

PR #12492, opened 2023-02-23
feat(databricks): add basic op implementations

rexledesma:
### Summary & Motivation
There are two standard ways of triggering jobs on Databricks: either you can submit API calls to trigger an existing job, or you can construct a job definition on the fly. Here, we add support for both methods.

pytest, local


rexledesma:
* master
  * **PR #12364** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12364" target="_blank">Graphite</a> 
    * **PR #12366** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12366" target="_blank">Graphite</a> 
      * **PR #12377** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12377" target="_blank">Graphite</a> 
        * **PR #12382** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12382" target="_blank">Graphite</a> 
          * **PR #12492** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12492" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12492?utm_source=stack-comment).

sryza (commented):
What do you think about making these op factories to make them consistent with `create_databricks_job_op`?

Aside from consistency, it also makes it easier to customize the name and number of inputs.

rexledesma:
> What do you think about making these op factories to make them consistent with `create_databricks_job_op`?

I actually do not like this pattern: I believe asset/op factories should be an abstraction that the user defines, since the amount of customization that they want here could be unbounded. 

> Aside from consistency, it also makes it easier to customize the name and number of inputs.

- For the name, could we direct the user to use `.alias`? 
- For the number of inputs, can't `start_after` take in a list of inputs? Since this is a `Nothing` input anyways, one input will suffice in order to define dependency order from `N` inputs.

rexledesma:
@sryza After some more thought, I'll wrap these ops in a factory pattern. But I'll keep the arguments of the factory to be arbitrary kwargs that are passed down to the op. This accomplishes two things. This way, the factory pattern is infinitely customizable out of the box, but retains the sensible defaults of the underlying op.

Out-of-the-box factories like this are a little complicated since there's another layer of indirection that the user needs to understand, but this plays nicely for interpreting arbitrary assets from the job using `AssetsDefintion.from_op`.

```
def factory(**kwargs):
    @op(**{**sensible_defaults, **kwargs})
    def _impl():
        pass
    
    return _impl
```

rexledesma:
Changed a couple of things here:

- refactored to op factory notation
- lifted up required configuration to be parameters of the op factory, making `.configured` optional

python_modules/libraries/dagster-databricks/dagster_databricks/ops.py
```diff
@@ -135,3 +136,239 @@ def create_ui_url(databricks_client, op_config):
...
+    """
+    Creates an op that launches an existing databricks job.
+
+    As config, the op accepts a blob of the form described in Databricks' Job API:
+    https://docs.databricks.com/api-explorer/workspace/jobs/runnow. The only required field is
+    ``job_id``, which is the ID of the job to be executed. Additional fields can be used to specify
+    override parameters for the Databricks Job.
+
+    Arguments:
+        databricks_job_id (int): The ID of the Databricks Job to be executed.
```
sryza, line 151: I know you mentioned it above, but I think it's worth mentioning the kwargs here and being specific about what kinds of parameters are accepted (all the parameters to the op decorator?)

python_modules/libraries/dagster-databricks/dagster_databricks/ops.py
```diff
@@ -135,3 +136,239 @@ def create_ui_url(databricks_client, op_config):
         return "https://{host}/?o={workspace_id}#joblist".format(
             host=host, workspace_id=workspace_id
         )
+
+
+def create_databricks_run_now_op(databricks_job_id: int, **kwargs) -> OpDefinition:
```
sryza, line 141: I wonder if this should take a `databricks_job_configuration` like the other function? Seems nice to not need to use `configured` to specify `python_params`?

sryza (commented):
This looks close to me - just left one comment and one question

rexledesma:
@sryza Great - that feedback should be addressed in https://github.com/dagster-io/dagster/pull/12492/commits/aa7ba1752efc4c65ba678c548a9092e619c46b01. The user shouldn't have to use `configured` at all now, and will still get the benefits of runtime configuration via the launchpad.

schrockn:
My general impression is that this API offers too many "degrees of freedom" here by passing through all the keyword args.

sryza (approved):
Nice nice nice

python_modules/libraries/dagster-databricks/dagster_databricks/ops.py
```diff
@@ -135,3 +138,247 @@ def create_ui_url(databricks_client, op_config):
         return "https://{host}/?o={workspace_id}#joblist".format(
             host=host, workspace_id=workspace_id
         )
+
+
+def create_databricks_run_now_op(
```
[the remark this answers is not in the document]
sryza: Relevant discussion here btw: https://elementl-workspace.slack.com/archives/C04JNJCSS13/p1677526031428709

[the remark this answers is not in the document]
sryza: Maybe worth a chat? I agree that this is an API that's hard to walk back and I'm a little nervous about it, but I also think it's a promising pattern.  

Maybe best to be conservative and leave out the runtime config for now?

Having two separate top-level APIs seems costly to me - I think many users will find the distinction immaterial and be deterred by needing to make a choice.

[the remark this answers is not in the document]
rexledesma: I think the confusion comes from the underlying Databricks API call.

When using the `Run Now` API, you can specify a custom list of params to pass into the underlying Databricks Job. Otherwise, the default params of the Databricks Job will be used.

Similarly, when using the `Submit Run` Databricks API, you can specify the list of Databricks tasks you want to run, as well as the configuration of the cluster on which the tasks run. It is this cluster configuration that could be useful to configure at run time, but from talking to some users, cluster configuration is rarely manipulated at run time - if any changes are made to it, it would probably be a permanent change, meaning a code change would be required.

I can get rid of its config for now then, and wait until a user explicitly asks for this run-time configuration of their jobs.

python_modules/libraries/dagster-databricks/dagster_databricks/ops.py
```diff
@@ -135,3 +138,224 @@ def create_ui_url(databricks_client, op_config):
...
+                float,
+                description=(
+                    "If the Databricks Job is not complete after this length of time, raise an"
+                    " error."
+                ),
+                default_value=_DEFAULT_RUN_MAX_WAIT_TIME_SEC,
+            ),
+        },
+        required_resource_keys={"databricks"},
+        tags={"kind": "databricks"},
```
schrockn, line 222: in terms of parameterizing config, what I think *would* make sense is to make these default values parameterizable, which is right way to get "best of both worlds" here.

schrockn: And I think the approach is _preferable_ to calling `configured`

rexledesma: Addressed in https://github.com/dagster-io/dagster/pull/12492/commits/9c42bfdc217cc32e0755af7eca2accb921765470

schrockn:
Last question in this one is about parameterizing `ins` and `outs` here. This is a case where the asset variant of this makes more sense (upstream asset keys and the asset_keys materialized by the databricks job) but the op mental model is less great.

rexledesma:
> Last question in this one is about parameterizing `ins` and `outs` here. This is a case where the asset variant of this makes more sense (upstream asset keys and the asset_keys materialized by the databricks job) but the op mental model is less great.

Alright, we'll also divest from this functionality and only expose it in the `multi_asset` factory version of these APIs. There was some thinking here that if we wanted the user to migrate from the op factory to its asset version, they would use `AssetDefinition.from_op`. But instead, we'll assume that they just migrate wholesale to the `multi_asset` factory version.

PR #12494, opened 2023-02-23
add dynamic partitions example

sryza:
### Summary & Motivation

Adds an example to demonstrate how to use dynamic partitions. Take a look at the README for details on how the pipeline functions.

Ran the full pipeline locally.


examples/assets_dynamic_partitions/assets_dynamic_partitions/assets.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+@asset(
+    partitions_def=releases_partitions_def,
+    io_manager_key="warehouse",
+    metadata={"partition_expr": "release_tag"},
+)
+def releases_metadata(context) -> DataFrame:
+    """Table with a single row per release"""
+    release_tag = context.partition_key
```
slopp, line 21: Throughout... should we use `context.asset_partition_key_for_output()` or `context.partition_key`? I prefer the later but the docs use the former: https://docs.dagster.io/concepts/partitions-schedules-sensors/partitions#defining-partitioned-assets

examples/assets_dynamic_partitions/assets_dynamic_partitions/release_sensor.py
```diff
@@ -0,0 +1,50 @@
...
+    if latest_tracked_release is None:
+        new_releases = all_releases
+    else:
+        new_releases = [
+            release
+            for release in all_releases
+            if semver_tuple(release) > semver_tuple(latest_tracked_release)
+        ]
+
+    partitions_to_add = sorted(new_releases, key=semver_tuple)
```
slopp, line 43: Sorting is only important here because of the line 50 "only get latest release" right?

examples/assets_dynamic_partitions/assets_dynamic_partitions/release_sensor.py
```diff
@@ -0,0 +1,50 @@
...
+    context.log.info(f"Adding partitions: {partitions_to_add}")
+    assets.releases_partitions_def.add_partitions(partitions_to_add, context.instance)
+
+    # We only launch a run for the latest release, to avoid unexpected large numbers of runs the
```
slopp, line 47: This is an interesting point that might be worth calling out in the README or documenting as a pattern in docs more generally. In my experiments  / examples, I didn't do the subsetting and just had "turn on sensor" = "run backfill".

examples/assets_dynamic_partitions/assets_dynamic_partitions/assets.py
```diff
@@ -0,0 +1,125 @@
...
+
+    response = requests.get(
+        f"https://api.github.com/repos/dagster-io/dagster/releases/tags/{release_tag}",
+        auth=("sryza", os.environ["GITHUB_API_KEY"]),
```
slopp, line 25: Maybe move the username to an env var too?

examples/assets_dynamic_partitions/README.md
```diff
@@ -0,0 +1,26 @@
...
+First, install the project as a Python package. By using the --editable flag, pip will install your Python package in ["editable mode"](https://pip.pypa.io/en/latest/topics/local-project-installs/#editable-installs) so that as you develop, local code changes will automatically apply.
+
+```bash
+pip install -e ".[dev]"
+```
+
+Then, start the Dagster web server:
+
+```bash
+dagster dev
```
slopp, line 23: I would document the required env vars in the readme to be consistent with other quick starts

slopp (approved):
I only looked at the example, not the tweaks to the IO manager, but the example is cool! I left a few comments, none blocking IMO.

One minor point, not a deal breaker but might be worth calling out in the readme, the direct use of the file system means that on Dagster Cloud this example might be a bit wonky because the fs is ephemeral per run:

1. The summary asset will only show the result for the upstream partition that happened to be materialized in the same run. Materializing the summary asset on its own will always fail

2. Partial re-runs will not work as expected

examples/assets_dynamic_partitions/assets_dynamic_partitions/release_sensor.py
sryza: It also determined the order that the partitions show up in in the UI

examples/assets_dynamic_partitions/assets_dynamic_partitions/assets.py
sryza: oops good call

examples/assets_dynamic_partitions/assets_dynamic_partitions/assets.py
sryza: Hmm good question. The former is theoretically more principled, but I'm just going to switch the docs to the latter.

examples/assets_dynamic_partitions/README.md
sryza: great point

examples/assets_dynamic_partitions/assets_dynamic_partitions/release_sensor.py
sryza: 👁️ 👁️

PR #12500, opened 2023-02-23
[rename] LogicalVersion -> DataVersion

smackesey:
### Summary & Motivation

Rename `LogicalVersion` -> `DataVersion` everywhere except for the GQL schema and docs (this is done upstack).

The only remaing use of "logical" version is in the (internal) function `compute_logical_data_version`, which is a rename of `compute_logical_version`. "logical" was retained here as a qualifier to "data version" to indicate the method of computation (`compute_data_version` is too unspecified).

Note that this also updates the tags used to store logical versions and input logical versions from:

- `dagster/logical_version` -> `dagster/data_version`
- `dagster/input_logical_version` -> `dagster/input_data_version`

This is important since we display the actual tags in dagit. All reading of these tags should be done through `extract_data_version_from_entry` and `extract_data_provenance_from_entry`, which will try to read both the old and the new tags.

Existing test suite + added a new unit test that accesses legacy tags.


python_modules/dagster/dagster/_core/definitions/events.py
```diff
@@ -241,8 +241,12 @@ def __init__(
         output_name: Optional[str] = DEFAULT_OUTPUT,
         metadata_entries: Optional[Sequence[Union[MetadataEntry, PartitionMetadataEntry]]] = None,
         metadata: Optional[Mapping[str, RawMetadataValue]] = None,
-        logical_version: Optional[LogicalVersion] = None,
+        data_version: Optional[DataVersion] = None,
+        logical_version: Optional[DataVersion] = None,
```
schrockn, line 245: unless we are exporting LogicalVersion, seems like we should just cut it over in one shot.

schrockn (changes requested):
With the tag names changes, what is going to happen when we push this to Cloud? Are we going to have bugs around staleness until people start doing more materializations?

Seems like we need an intermediate state where we read from one OR the other.

python_modules/dagster/dagster/_core/definitions/data_version.py
```diff
@@ -44,158 +45,177 @@ def foo(x):
...
-        code_version = tags.get(CODE_VERSION_TAG_KEY)
+        code_version = tags.get(CODE_VERSION_TAG)
         if code_version is None:
             return None
-        start_index = len(INPUT_LOGICAL_VERSION_TAG_KEY_PREFIX) + 1
-        input_logical_versions = {
-            AssetKey.from_user_string(k[start_index:]): LogicalVersion(tags[k])
+        input_data_versions = {
+            # Everything after the 2nd slash is the asset key
+            AssetKey.from_user_string(k.split("/", maxsplit=2)[-1]): DataVersion(tags[k])
```
schrockn, line 119: why the change in logic here?

python_modules/dagster/dagster/_core/definitions/data_version.py
```diff
@@ -44,158 +45,177 @@ def foo(x):
...
+INPUT_EVENT_POINTER_TAG_PREFIX: Final[str] = "dagster/input_event_pointer"
+
+
+def read_input_data_version_from_tags(
+    tags: Mapping[str, str], input_key: "AssetKey"
+) -> Optional[DataVersion]:
+    value = tags.get(
+        get_input_data_version_tag(input_key, prefix=INPUT_DATA_VERSION_TAG_PREFIX)
+    ) or tags.get(get_input_data_version_tag(input_key, prefix=_OLD_INPUT_DATA_VERSION_TAG_PREFIX))
+    return DataVersion(value) if value is not None else None
```
schrockn, line 150: Ah I see we are doing that here. I guess I just need to see that this is validated in the test plan.

python_modules/dagster/dagster/_core/definitions/data_version.py
smackesey: Because there are now two possible tag prefixes so we can't use the length.

python_modules/dagster/dagster/_core/definitions/events.py
smackesey: whoops yeah that was leftover from when this was originally intended to merge pre 1.2

python_modules/dagster/dagster/_core/definitions/data_version.py
schrockn: ah i see

python_modules/dagster/dagster/_core/definitions/data_version.py
smackesey: I added a test that accesses legacy tags

schrockn (approved):
Ok.

I think it is important to do some manual QA on this.

PR #12504, opened 2023-02-23
Modify `get_first_partition_window` to account for offset

clairelin135:
Causes a graphQL error reported by a user here: https://dagster.slack.com/archives/C01U954MEER/p1677064207587029

Previously, `get_first_partition_window` did not account for the end offset.

This led to errors such as `get_first_partition_window` returning `None` when the start time window is the current time window and offset > 0.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
@@ -370,17 +370,41 @@ class TimeWindowPartitionsDefinition(
     def get_first_partition_window(
         self, current_time: Optional[datetime] = None
     ) -> Optional[TimeWindow]:
-        current_timestamp = (
+        current_time = cast(
+            datetime,
             pendulum.instance(current_time, tz=self.timezone)
             if current_time
-            else pendulum.now(self.timezone)
-        ).timestamp()
+            else pendulum.now(self.timezone),
+        )
+        current_timestamp = current_time.timestamp()
 
         time_window = next(iter(self._iterate_time_windows(self.start)))
-        if time_window.end.timestamp() <= current_timestamp:
-            return time_window
+
+        if self.end_offset == 0:
+            return time_window if time_window.end.timestamp() <= current_timestamp else None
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -370,17 +370,41 @@ def get_prev_partition_window(self, start_dt: datetime) -> Optional[TimeWindow]:
...
 
         time_window = next(iter(self._iterate_time_windows(self.start)))
-        if time_window.end.timestamp() <= current_timestamp:
-            return time_window
+
+        if self.end_offset == 0:
+            return time_window if time_window.end.timestamp() <= current_timestamp else None
         else:
-            return None
+            if self.end_offset > 0:
```
sryza, line 386: Nitpick: making this an `elif` would save a level of indentation and make this a little easier to read

python_modules/dagster/dagster_tests/definitions_tests/test_time_window_partitions.py
```diff
@@ -652,3 +652,77 @@ def my_partitioned_config_2(_start, _end):
...
+
+    assert DailyPartitionsDefinition(
+        start_date="2023-01-01", end_offset=1
+    ).get_first_partition_window(
+        current_time=datetime.strptime("2023-01-01", "%Y-%m-%d")
+    ) == time_window(
+        "2023-01-01", "2023-01-02"
+    )
+
+    assert (
```
sryza, line 670: Curious why this isn't line-wrapping

sryza (commented):
I think it would be worthwhile to include a test where the current_time doesn't lie exactly on a partition boundary.

clairelin135:
@sryza There are a couple of tests in there with `MonthlyPartitionsDefinitions` where `current_time` is in the middle of the month

sryza (approved):
To be extra careful, I think it's worth having one in the middle of the day or hour. I won't block this though.

Thanks for fixing btw.

Issue #12426, opened 2023-02-17
Dynamic type checking fails when downstream asset refers to a group of upstream partitions


plaflamme:
### Dagster version

1.1.18

### What's the issue?

Assuming a custom IO manager that can provide different representations of an asset, i.e.:

```python
class MyIOManager(UPathIOManager):
  # returns an pandas.DataFrame or pyarrow.Table based on dynamic type checking
  def load_from_path(self, context: InputContext, path: UPath) -> Any:
    tpe = context.dagster_type.typing_type
    if tpe == pandas.DataFrame:
      # returns a pandas.DataFrame
      return ...
    elif tpe == pyarrow.Table:
      # returns a pyarrow.Table
      return ...
```

Then assuming an hourly asset exists:

```python
@asset(partitions_def=HourlyPartitionDefinition(...))
def upstream() -> pandas.DataFrame:
  pass
```

Downstream assets can use either:

```python
@asset(partitions_def=HourlyPartitionDefinition(...))
def use_pandas(upstream: pandas.DataFrame):
  pass

@asset(partitions_def=HourlyPartitionDefinition(...))
def use_arrow(upstream: pyarrow.Table):
  pass
```

But if a downstream asset wishes to combine the hourly partitions in to a daily partition, type checking fails:

```python
@asset(partitions_def=DailyPartitionDefinition(...))
def use_pandas(upstream: Dict[str, pandas.DataFrame]):
  pass

@asset(partitions_def=DailyPartitionDefinition(...))
def use_arrow(upstream: Dict[str, pyarrow.Table]):
  pass
```

### What did you expect to happen?

Dynamic type checking should allow the situation where a downstream asset takes a group of upstream partitions and an IO manager is able to read different "flavours" of an asset.

### Deployment type

Local

### Additional information

Relevant slack thread: https://dagster.slack.com/archives/C01U954MEER/p1676585117044459

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

sryza:
@plaflamme I'm trying to triangulate this a little bit. Which of the following is the case for this?
- Dagster is raising an error in a situation where it shouldn't be raising an error. If it didn't raise that error, then you'd be able to get things working with custom logic your code.
- Dagster should be exposing information that it currently doesn't expose, to allow you to accomplish what you're trying to do.
- A third option I haven't thought of

plaflamme:
[...]

PR #12508, opened 2023-02-23
pare down multi-partition runtime type checking in upath IO manager

sryza:
### Summary & Motivation

The `UPathIOManager` includes some type-checking to help users catch situations where they're mishandling the types for input partitions.  This type-checking is currently pretty aggressive and can lead to false positives, e.g. in the situation outlined here: https://github.com/dagster-io/dagster/issues/12426.

It's better to be too lax than too aggressive, because users have no workaround for the false positives.

This PR pares down this type-checking to only check a single situation, for what I believe is the most common error: handling unexpected cases of multiple input partitions.

updated test

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/upath_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/upath_io_manager.py
@@ -1,8 +1,5 @@
-from __future__ import annotations
-
-import inspect
 from abc import abstractmethod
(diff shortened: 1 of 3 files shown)


jamiedemaria (approved):
looks good to me! could add a test that replicates the issue in #12426, but i think it's also fine without it

PR #12512, opened 2023-02-24
Fetch multipartition key from context methods

clairelin135:
Reported by a user here: https://dagster.slack.com/archives/C01U954MEER/p1677064207587029

In Dagit, we enabled launching a range of partitions as a single run. This PR updates context methods to handle partition keys correctly following this:
- `context.partition_key` would through an "unpartitioned run" error in this case. This PR updates the error message to be more informative when a partition range is provided
- partition keys fetched from the `ASSET_PARTITION_RANGE_START` and `ASSET_PARTITION_RANGE_END` tags were the formatted complete multipartition tag (with the "|"). But these values were returned as strings and not as `MultiPartitionKeys`, so users were unable to use `key.keys_by_dimension`. This PR fixes this behavior, by fetching the partitions definition to create a `MultiPartitionKey` instead.

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/context/system.py
+++ b/python_modules/dagster/dagster/_core/execution/context/system.py
@@ -355,7 +355,7 @@ class PlanExecutionContext(IPlanContext):
         return partitions_def
 
     @property
-    def is_partitioned_run(self) -> bool:
+    def has_partitions(self) -> bool:
         tags = self._plan_data.dagster_run.tags
         return bool(
             PARTITION_NAME_TAG in tags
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/execution/context/system.py
```diff
@@ -342,59 +342,96 @@ def scoped_resources_builder(self) -> ScopedResourcesBuilder:
...
+        pipeline_def = self._execution_data.pipeline_def
+        if not isinstance(pipeline_def, JobDefinition):
+            check.failed(
+                "Can only call 'partitions_def', when using jobs, not legacy pipelines",
+            )
+        partitions_def = pipeline_def.partitions_def
+        return partitions_def
+
+    @property
+    def is_partitioned_run(self) -> bool:
```
sryza, line 358: what do you think about calling this `has_partitions` for consistency with `has_asset_partitions`?

sryza (approved):

python_modules/dagster/dagster/_core/execution/context/system.py
clairelin135: sounds good! changed this

PR #12517, opened 2023-02-24
handle multiple non-time-window partitions in db io managers

sryza:


sryza:
@jamiedemaria - thanks for pointing that out. I updated the reference query for static partitions. I don't think the time-partitions query needs updating.

jamiedemaria (approved):

PR #12519, opened 2023-02-24
[dagit] Remove unnecessary usage of <TestProvider>

bengotow:
### Summary & Motivation

I went through the tests that used `<TestProvider>` and refactored or removed it in places where it seemed straightforward. It's still in use in 15 files, but those pass mocks or other props and probably need more careful removal.

Tests still pass! I also altered the ReexcutionDialog test "failure" case to be a bit more interesting since we have more control over the mocks with the MockedProvider.


salazarm (approved):

bengotow:
Test failures in the oss-compatibility task are in JS but look unrelated, going to go ahead and merge this!

PR #12528, opened 2023-02-24
[bugfix] Make assets downstream of partitions never stale in dagit

smackesey:
### Summary & Motivation

Adjusts GQL resolver to return `None` for the projected logical version of any asset downstream of a partitioned asset. This will cause the asset to never be marked stale in dagit (the correct stale status is currently undefined for this case).

Also removes a chunk of leftover unreachable code.

Modified unit test

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/logical_version.py
+++ b/python_modules/dagster/dagster/_core/definitions/logical_version.py
@@ -252,6 +252,9 @@ class CachingStaleStatusResolver:
     def get_projected_logical_version(self, key: AssetKey) -> LogicalVersion:
         return self._get_projected_logical_version(key=key)
 
+    def is_partitioned_or_downstream(self, key: AssetKey) -> bool:
+        return self._is_partitioned_or_downstream(key=key)
+
     @cached_method
     def _get_status(self, key: AssetKey) -> StaleStatus:
         if self.asset_graph.get_partitions_def(key):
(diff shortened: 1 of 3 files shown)


smackesey:
* master
  * **PR #12528** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12528" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12528?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/logical_version.py
```diff
@@ -260,15 +263,6 @@ def _get_status(self, key: AssetKey) -> StaleStatus:
             causes = self._get_status_causes(key=key)
             return StaleStatus.FRESH if len(causes) == 0 else causes[0].status
 
-        current_version = self._get_current_logical_version(key=key)
```
smackesey, line 263: This code is unreachable and should've been deleted previously.

sryza (commented):
Do we have a Python test that includes an asset downstream of a partitioned asset?

smackesey:
>Do we have a Python test that includes an asset downstream of a partitioned asset?

Yes, the test is modified in this PR (changed from expecting `UNKNOWN` to expecting `None` for the projected logical version)

sryza (approved):

PR #12531, opened 2023-02-24
[fix] Correctly resolve asset jobs with empty selections

OwenKephart:
### Summary & Motivation

Previously, when an AssetSelection that resolved to an empty set was passed into define_asset_job, the generated job would select all assets, rather than none of them. Classic case of checking for `if foo` instead of `if foo is None`.

The test added to test_assets_job.py failed before, and passes now.


bengotow (approved):
Thanks for fixing this!

PR #12534, opened 2023-02-24
[dagit] Upgrade to Jest 29, allow more time for coverage collection

bengotow:
### Summary & Motivation

The goal of this PR is to resolve sporadic test failures due to Jest running ouf out-of-memory or running too slowly.

- Upgraded to Jest 29 and configured it to restart it's worker if it ever uses more then 1GB of RAM

- Increased the test timeout to 10s. The tests run considerably more slowly with `--coverage` enabled, especially tests that "cover" a large amount of code. My AssetPartitions tests render the whole page, which covers dozens of files, and adding --coverage pushes that test over the 5s timeout limit randomly. We could also disable coverage, but it seems nice to have? (Have we been viewing it somewhere?)


salazarm:
I removed coverage from the cloud build. I'd be in favor of removing it in OSS too.

hellendag (approved):

PR #12536, opened 2023-02-24
[pythonic config] Add test showcasing use of Pydantic validators

benpankow:
## Summary

Introduces test coverage validating that Pydantic [validators](https://docs.pydantic.dev/usage/validators/) work as expected.

## Test Plan

Unit test. Tested manually in launchpad - the error here shows up when the run starts. Eventually, it would be good if this validation would show up in the launchpad itself.

Screen Shot 2023-03-06 at 10 01 32 AM

### DIFF
--- a/python_modules/dagster/dagster_tests/core_tests/structured_config_tests/test_validation.py
+++ b/python_modules/dagster/dagster_tests/core_tests/structured_config_tests/test_validation.py
@@ -1,7 +1,7 @@
-from dagster import op, job
+import pytest
+from dagster import job, op
 from dagster._config.structured_config import Config
 from pydantic import ValidationError, validator
-import pytest
 
 
 def test_validators_basic() -> None:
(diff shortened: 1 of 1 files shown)


benpankow:
* master
  * **PR #12536** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12536" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12536?utm_source=stack-comment).

schrockn (changes requested):
Can we add to the test plan in  his PR to see how a violation of a validation error is surfaced in the config editor?

benpankow:
Yes, updated w/ a screenshot. Right now the error only surfaces on run start (when the config class is populated). Will look into if we can get this to work as part of launchpad configuration.

schrockn:
I think this is right behavior in the launchpad btw. We used to have server-side evaluation in the critical path in the launchpad and it caused lots of perf issues and lag. We should never try to interpret pythonic rules client-side so I actually believe the behavior you are demonstrating is the correct tradeoff.

schrockn (approved):

PR #12550, opened 2023-02-25
Fix master

clairelin135:
Looks like this `dirs_exist_ok` I added to a test util isn't available in python 3.7.
https://github.com/dagster-io/dagster/blob/31022e03a816e01027fcd88e9795b8fdaacdac87/python_modules/dagster/dagster/_utils/test/__init__.py#L402

BK did not catch since it only runs tests in 3.9


rexledesma (approved):

clairelin135:
🙌 thanks rex!

Issue #11439, opened 2022-12-30
Support loading Definitions from multiple modules in the pyproject.toml file


bendnorman:
### What's the use case?

 I’m able to load multiple definitions from modules using the -m argument:
```
dagit -m module_1 -m module_2
```
It would be nice to be able to specify multiple modules in the pyproject.toml file so multiple modules can be loaded by just using the dagit command.

### Ideas of implementation

I figure the end result could look like this:

```
[tool.dagster]
module_name = ["module_1", "module_2]
```

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

bendnorman:
I'm happy to take a stab at this if a maintainer can point me in the right direction. Thanks!

bendnorman:
This was closed by #22327!

PR #12551, opened 2023-02-25
Add support for loading multiple modules from pyproject.toml

bendnorman:
### Summary & Motivation
It would be nice to load multiple modules using `pyproject.toml` so users don't have to type out all the modules they want to load using the CLI command. See issue #11439 and discussion #11167.

I added a new unit test in `dagster_tests/cli_tests/test_toml_loading.py`

### DIFF
--- a/docs/next/components/mdx/includes/dagster/DagsterDevTabs.mdx
+++ b/docs/next/components/mdx/includes/dagster/DagsterDevTabs.mdx
@@ -60,6 +60,13 @@ Instead of this:
 dagster dev -m your_module_name
 ```
 
+You can also include multiple modules at a time using the `pyproject.toml` file, where each module will be loaded as a code location:
+
+```toml
+[tool.dagster]
+module_name = ["your_module_name", "you_second_module_name"]  ## names of project's Python modules
+```
+
 ---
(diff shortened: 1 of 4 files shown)


gibsondan:
This seems totally reasonable to me. @schrockn do you feel OK about the syntax here? The alternative would be to have separate 'module_name' (when its singular) and 'module_names' (when it's plural) but a single key is probably better for consistency.

schrockn:
My bad on the tardiness to the review here.

One alternative to consider here is a different more "advanced" entry to allow for future extensions.

`modules = [{module_name: foo}, {module_name: bar}]`

and also also the specification of attribute and location_name here.

The most generic and future-proof being: 
'code_locations = [{type:module, module_name: foo}, {type:module, module_name: bar}]`

My recommendation is a new `modules` toml entry.

gibsondan (changes requested):
Back to you for any thoughts about incorporating @schrockn 's suggestion. Thanks again for sending this out!

PedramNavid:
Closing stale PR.

bendnorman:
This was resolved by #22327 👍

PR #12562, opened 2023-02-27
GQL Add dynamic partition mutation

clairelin135:
In Dagit, we want to be able to add a new partition to a dynamic partitions definition through the UI.

This PR adds a mutation that allows for adding this new partition.


salazarm (approved):
lgtm!

python_modules/dagster/dagster/_core/workspace/permissions.py
```diff
@@ -16,6 +16,7 @@ class Permissions(str, Enum):
     WIPE_ASSETS = "wipe_assets"
     LAUNCH_PARTITION_BACKFILL = "launch_partition_backfill"
     CANCEL_PARTITION_BACKFILL = "cancel_partition_backfill"
+    ADD_DYNAMIC_PARTITION = "add_dynamic_partition"
```
gibsondan, line 19: suggestion: EDIT_DYNAMIC_PARTITIONS? i bet you will be able to update and remove them too? unless those permissions could be separate

gibsondan (changes requested):
can we get a test that fails with an unauthorized error in read-only mode? grep for that error + graphql tests for examples

python_modules/dagster/dagster/_core/workspace/permissions.py
clairelin135: This is a good suggestion--changed it to this. I've also added the test as you suggested

salazarm:
Can we add an error for if the partition name is not unique? 

It's one of the error states we want to show on the client.

python_modules/dagster-graphql/dagster_graphql/schema/roots/mutation.py
```diff
@@ -329,6 +331,24 @@ def mutate(self, graphene_info: ResolveInfo, backfillId: str):
...
+        partitionsDefName = graphene.NonNull(graphene.String)
+        partitionKey = graphene.NonNull(graphene.String)
+
+    class Meta:
+        name = "AddDynamicPartitionMutation"
+
+    @capture_error
+    @require_permission_check(Permissions.EDIT_DYNAMIC_PARTITIONS)
+    def mutate(self, graphene_info: ResolveInfo, partitionsDefName: str, partitionKey: str):
+        return add_dynamic_partition(graphene_info, partitionsDefName, partitionKey)
```
bengotow, line 349: Hey gang! Hmm @clairelin135  I'm not sure that we expose the partition definition name to the UI -- it looks like this is a property of `DynamicPartitionsDefinition`, maybe we can expose it on `GraphenePartitionDefinition`? I don't think we have a way to retrieve this param currently!

Separately it'd actually be nice to have an ID of some sort on these -- I think the UI has to compare partition definitions for equality by comparing their typename and dimension types + names.

clairelin135:
Hey @salazarm @bengotow I've gone ahead and added:
- an error message when the partition already exists (there's a gql test for this case)
- a `name` field on partitionDefinition that exists when the partitions definition is dynamic

Would the ID field be something on the partitions definition? I can go ahead and add that if it's helpful for this PR, otherwise I can break it out and add it in a future PR

clairelin135:
ah @salazarm I see that you've added the partitions def name functionality in https://github.com/dagster-io/dagster/pull/12585#pullrequestreview-1318184516 -- I'll go ahead and remove it in this PR

python_modules/dagster-graphql/dagster_graphql/implementation/execution/dynamic_partitions.py
```diff
@@ -0,0 +1,28 @@
...
+@capture_error
+def add_dynamic_partition(
+    graphene_info, partitions_def_name: str, partition_key: str
+) -> "GrapheneAddDynamicPartitionSuccess":
+    from ...schema.partition_sets import GrapheneAddDynamicPartitionSuccess
+
+    assert_permission(graphene_info, Permissions.EDIT_DYNAMIC_PARTITIONS)
+
+    if graphene_info.context.instance.has_dynamic_partition(partitions_def_name, partition_key):
+        raise DagsterInvalidInvocationError(
```
salazarm, line 21: `UserFacingGraphqlError`

clairelin135: nice, thank you!

salazarm (approved):
lgtm

python_modules/dagster-graphql/dagster_graphql/schema/errors.py
```diff
@@ -495,6 +495,24 @@ def __init__(self, message=None):
...
+        super().__init__()
+        self.partitions_def_name = check.str_param(partitions_def_name, "partitions_def_name")
+        self.partition_name = check.str_param(partition_name, "partition_name")
+        self.message = (
+            f"Partition {self.partition_name} already exists in dynamic partitions definition"
+            f" {self.partitions_def_name}."
+        )
+
+
 types = [
```
salazarm, line 516: Does it need to be included in this array

clairelin135:
@gibsondan -- mind taking a second look at this or dismissing your review? merging is blocked until your review is amended it seems, and I'd like to unblock @salazarm as soon as possible

python_modules/dagster-graphql/dagster_graphql/schema/errors.py
clairelin135: most likely--i've added it now

Issue #10240, opened 2022-10-27
Deprecate legacy asset keys


clairelin135:
We've been using the new asset key version since `0.9.19`. We should deprecate the legacy asset keys in the next major release to reduce compute time for our event log queries

plaflamme:
This is particularly painful on relatively large `event_log` tables (3.1M rows). Here's an example of the latency difference it can cause for a query that is executed thousands of times per hour under normal operation for my instance (I'm not sure what triggers this query, but it's quite frequent during backfilling):

```
EXPLAIN ANALYZE SELECT event_logs.id, event_logs.event 
FROM event_logs 
WHERE event_logs.dagster_event_type = 'ASSET_MATERIALIZATION' AND (event_logs.asset_key = '["not", "the "real" "key"]' OR event_logs.asset_key = 'not.the.real.key') ORDER BY event_logs.id DESC 
 LIMIT 1;
                                                                      QUERY PLAN                                                                       
-------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.43..280.12 rows=1 width=831) (actual time=246.699..246.701 rows=1 loops=1)
   ->  Index Scan Backward using idx_event_type on event_logs  (cost=0.43..129776.03 rows=464 width=831) (actual time=246.698..246.699 rows=1 loops=1)
         Index Cond: (dagster_event_type = 'ASSET_MATERIALIZATION'::text)
         Filter: ((asset_key = '["not", "the "real" "key"]'::text) OR (asset_key = 'not.the.real.key'::text))
         Rows Removed by Filter: 21649
 Planning Time: 0.151 ms
 Execution Time: 246.728 ms
(7 rows)
```

vs

```
EXPLAIN ANALYZE SELECT event_logs.id, event_logs.event 
FROM event_logs 
WHERE event_logs.dagster_event_type = 'ASSET_MATERIALIZATION' AND (event_logs.asset_key = '["not", "the "real" "key"]') ORDER BY event_logs.id DESC 
 LIMIT 1;
                                                                     QUERY PLAN                                                                      
-----------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.55..1.59 rows=1 width=831) (actual time=1.655..1.657 rows=1 loops=1)
   ->  Index Scan Backward using idx_events_by_asset on event_logs  (cost=0.55..481.68 rows=462 width=831) (actual time=1.654..1.655 rows=1 loops=1)
[...]

PR #12563, opened 2023-02-27
[1.2] Remove legacy asset key format from queries

clairelin135:
Fixes https://github.com/dagster-io/dagster/issues/10240

This PR removes support for the legacy asset key serialization (path joined by `'.'`). Users who were using assets prior to `0.10.0` may be affected. For these users, we'll direct them to reach out to us for a data migration.


sryza (approved):
Nice - let's wait until the branch cut to merge this and make sure to include a note in the migration guide.

prha (approved):

PR #12565, opened 2023-02-27
Fraser/rework readme

frasermarlow:
### Summary & Motivation

To make the README.md file more attractive and inviting to new users.

All changes are in the MARKDOWN.md file and were tested visually in the browser.  The Asset graph was created with the code displayed in the README.md file and matches 1:1.


python_modules/dagster/README.md
```diff
@@ -26,55 +16,59 @@
...
-An asset graph defined in Python:
+Here is an example of a graph of three assets defined in Python:
 
 ```python
 from dagster import asset
 from pandas import DataFrame, read_html, get_dummies
 from sklearn.linear_model import LinearRegression
 
-@asset
+@asset(compute_kind="Python")
```
sryza, line 43: IMO we should keep the `compute_kind`s out, because it looks like extra boilerplate

python_modules/dagster/README.md
```diff
@@ -26,55 +16,59 @@
...
+  <a target="_blank" href="https://dagster.io/slack" style="background:none">
+    <img src="https://img.shields.io/badge/slack-dagster-blue.svg?labelColor=4F43DD&color=163B36&logo=slack" />
+  </a>
+  <a target="_blank" href="https://linkedin.com/showcase/dagster" style="background:none">
+    <img src="https://img.shields.io/badge/linkedin-dagster-blue.svg?labelColor=4F43DD&color=163B36&logo=linkedin" />
+  </a>
+</div>
 
-Dagster is an orchestrator that's designed for developing and maintaining data assets, such as tables, data sets, machine learning models, and reports.
+__Dagster is a cloud-native orchestrator for the whole development lifecycle, with integrated lineage and observability, a declarative programming model, and best-in-class testability.__
```
sryza, line 30: Is it a little weird to have "Dagster is a" twice in a row?

frasermarlow: Agreed - I will fix that.

python_modules/dagster/README.md
frasermarlow: I appreciate the value of pairing this right down.  I took out the descriptions for the assets as they added extra lines.  I would favor keeping the `compute_kind` because they do make the asset graph a lot more engaging.

schrockn: For a README I share @sryza's instinct here that we should keep the code example minimal and omit the `compute_kind`

frasermarlow: OK, changes are reverted to the minimal code and basic asset graph.

sryza (approved):
Looks good! Sorry for the delay here.

PR #12567, opened 2023-02-27
Storybook for LaunchAssetChoosePartitionsDialog

salazarm:
### Summary & Motivation

Adds a storybook with 2 examples. An empty state, and one with "ordinal" partitions. I need to add an example for a timeseries based one once I'm able to load one.

Storybook

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/__stories__/LaunchAssetChoosePartitionsDialog.stories.tsx
+++ b/js_modules/dagit/packages/core/src/assets/__stories__/LaunchAssetChoosePartitionsDialog.stories.tsx
@@ -2,7 +2,6 @@ import {MockedProvider} from '@apollo/client/testing';
 import {Meta} from '@storybook/react/types-6-0';
 import * as React from 'react';
 
-import {StorybookProvider} from '../../testing/StorybookProvider';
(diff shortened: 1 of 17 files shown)


js_modules/dagit/packages/core/src/assets/__stories__/LaunchAssetChoosePartitionsDialog.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+// eslint-disable-next-line import/no-default-export
+export default {
+  title: 'LaunchAssetChoosePartitionsDialog',
+  component: LaunchAssetChoosePartitionsDialog,
+} as Meta;
+
+export const Empty = () => {
+  return (
+    <StorybookProvider>
```
hellendag, line 25: Does it need the whole `StorybookProvider`? It's a bit like `TestProvider` in that it provides the full mocked schema version of the Apollo provider, which it looks like you probably shouldn't need here.

hellendag (approved):

js_modules/dagit/packages/core/src/assets/__stories__/LaunchAssetChoosePartitionsDialog.stories.tsx
salazarm: Ah okay, I wasn't sure if I needed it tbh

PR #12573, opened 2023-02-27
add gitlab setup docs

prha:
### Summary & Motivation
Need to acknowledge that both Github and Gitlab are supported, even though the copy is not very different between the two.

Inspection


prha:
Ah great, will make those changes and update this PR

prha:
Still need to write up a section for branch deployments, but will do in a follow-up diff, since it's pretty involved.

erinkcochran87 (approved):

PR #12576, opened 2023-02-28
Fix ProgrammingError import in snowflake_io_manager 

herbert-allium:
### Summary & Motivation
I noticed that the `ProgrammingError` class is imported from Snowflake. It should have been from SQLAlchemy instead, as we are using the SQLAlchemy connector here as specified in lines 157-161
```python
            dict(
                schema=table_slice.schema,
                connector="sqlalchemy",
                **cast(Mapping[str, str], no_schema_config),
            ),
```

I modified the file locally and tested with Dagit. It works in this way.

### DIFF
--- a/python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
+++ b/python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
@@ -10,7 +10,7 @@ from dagster._core.storage.db_io_manager import (
     TablePartitionDimension,
     TableSlice,
 )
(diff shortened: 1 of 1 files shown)


sryza:
@jamiedemaria mind taking a look?

jamiedemaria (approved):
looks good! thanks for the contribution!

herbert-allium:
Thank you, @jamiedemaria! May we get a new release on PyPI? So that I can use this fix :)

jamiedemaria:
This will be included in this week's release. it will come out Thursday!

PR #12578, opened 2023-02-28
Move useViewport to `ui` form `core`

salazarm:
### Summary & Motivation

`useViewport` is a useful hook outside of just core. I'm moving it to `UI` so I can use it in the tagSelector component

 buildkite


bengotow (approved):
Yep this seems like a good change!

PR #12584, opened 2023-02-28
Update stale reasons to remove UNKNOWN, add MISSING

smackesey:
### Summary & Motivation

Wrap up a few changes to stale status resolution. This should enable dagit to remove any remaining staleness logic and just display what it receives from the server.

- `MISSING` has been added as a stale status for assets that have never been materialized.
- `UNKNOWN` is no longer a defined stale reason-- all assets are `STALE`, `FRESH` or `MISSING`. Some nodes that would have previously been `UNKNOWN` are now `STALE` (non-code-versioned node with upstream dep changed, node with no provenance), others are `FRESH` (non-code-versioned node with upstream not changed)
- Partitioned assets or assets downstream of partitioned assets are only ever `MISSING` or `FRESH`. (This will change in the future.)
- `StaleStatusReason` now has an optional `dependency` field (null for code-version-related reasons), and the `reason` field is simpler. That allows the reason to be straightforwardly displayed in dagit.

Added some new staleness tests

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
@@ -87,6 +87,7 @@ class GrapheneAssetStaleStatusCause(graphene.ObjectType):
     status = graphene.NonNull(GrapheneAssetStaleStatus)
     key = graphene.NonNull(GrapheneAssetKey)
     reason = graphene.NonNull(graphene.String)
+    dependency = graphene.Field(GrapheneAssetKey)
 
     class Meta:
         name = "StaleStatusCause"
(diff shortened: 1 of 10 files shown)


smackesey:
* master
  * **PR #12584** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12584" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12584?utm_source=stack-comment).

python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
```diff
@@ -87,6 +87,7 @@ class GrapheneAssetStaleStatusCause(graphene.ObjectType):
     status = graphene.NonNull(GrapheneAssetStaleStatus)
     key = graphene.NonNull(GrapheneAssetKey)
     reason = graphene.NonNull(graphene.String)
+    dependency = GrapheneAssetKey
```
bengotow, line 90: This might have to be `graphene.Field(GrapheneAssetKey)`? Didn't know this could be a bare property actually 🙌

bengotow (approved):
The GraphQL changes here look good to me

sryza (approved):
Nice

Issue #12575, opened 2023-02-28
to_source_asset doesn't provide partitions to IOManager


spenczar:
### Dagster version

1.1.20

### What's the issue?

SourceAssets created from partitioned assets via `asset.to_source_asset()` do not seem to bring their partition information along with them when used in a Graph.

Here is a minimal reproducer. I make an asset with a static partition definition (bug is observed with a DailyPartitionDefinition too, this is just simpler). I then make an Op and Job; the job uses `to_source_asset()` to provide the asset as an input to the op.

```python
from dagster import op, asset, StaticPartitionsDefinition
import datetime

@asset(partitions_def=StaticPartitionsDefinition(["1", "2", "3"]))
def partitioned_asset(context):
    key = context.asset_partition_key_for_output()
    return int(key) * 10


@op
def double_it(partitioned_asset):
    return src_asset * 2


@job
def demo_inconsistency():
    asset_val = partitioned_asset.to_source_asset()
    double_it(asset_val)
```

If I execute this, I get a surprising error: the IOManager can't load the input, because the key doesn't exist, because it looks at an un-partitioned path:
```
dagster._core.errors.DagsterExecutionLoadInputError: Error occurred while loading input "partitioned_asset" of step "double_it":

  File "/usr/local/lib/python3.10/dist-packages/dagster/_core/execution/plan/execute_plan.py", line 269, in dagster_event_sequence_for_step
    for step_event in check.generator(step_events):
  File "/usr/local/lib/python3.10/dist-packages/dagster/_core/execution/plan/execute_step.py", line 346, in core_dagster_event_sequence_for_step
    for event_or_input_value in step_input.source.load_input_object(step_context, input_def):
  File "/usr/local/lib/python3.10/dist-packages/dagster/_core/execution/plan/inputs.py", line 190, in load_input_object
    yield from _load_input_with_input_manager(loader, load_input_context)
  File "/usr/local/lib/python3.10/dist-packages/dagster/_core/execution/plan/inputs.py", line 828, in _load_input_with_input_manager
    with op_execution_error_boundary(
  File "/usr/lib/python3.10/contextlib.py", line 153, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/usr/local/lib/python3.10/dist-packages/dagster/_core/execution/plan/utils.py", line 85, in op_execution_error_boundary
    raise error_cls(

The above exception was caused by the following exception:
[...]

PR #12586, opened 2023-02-28
fix loading partitioned source assets in non-partitioned non-asset jobs

sryza:
### Summary & Motivation

Fixes https://github.com/dagster-io/dagster/issues/12575

The context passed to `load_input` will look like the context passed to `load_input` when a non-partitioned asset depends on a partitioned asset.  I.e., by default, it will load all partitions of the partitioned asset.

In the future, we might want to enable specifying a PartitionMapping in some way, but that's out of the scope of this PR


OwenKephart (approved):
makes sense

PR #12589, opened 2023-02-28
divest permissions for editing sensor from updating sensor cursor

dpeng817:
Creates separate perms set for updating a sensor than from editing a sensor.


gibsondan (dismissed):
this looks good, can we get a test case that hits an unauthorizederror when its in read only mode and tries to use this mutation? grep for UnauthorizedError for similar tests on the other endpoints (guess one didn't exist before b/c we figured there was already one covering EDIT_SENSOR in some fashion)

dpeng817:
@gibsondan I think that already exists: https://github.com/dagster-io/dagster/blob/bc130b863b141c4b8f216318ce436bf34b0a1254/python_modules/dagster-graphql/dagster_graphql_tests/graphql/test_sensors.py#L560

prha (approved):

PR #12593, opened 2023-02-28
Expose image on WorkspaceLocationEntry

gibsondan:
Summary:
Useful to know for various things what the image is for a given location in the workspace.

Test Plan: BK

### DIFF
--- a/python_modules/dagster/dagster/_core/workspace/context.py
+++ b/python_modules/dagster/dagster/_core/workspace/context.py
@@ -610,10 +610,11 @@ class WorkspaceProcessContext(IWorkspaceProcessContext):
             repository_location=location,
             load_error=error,
             load_status=WorkspaceLocationLoadStatus.LOADED,
-            display_metadata=location.get_display_metadata()
-            if location
-            else origin.get_display_metadata(),
+            display_metadata=(
+                location.get_display_metadata() if location else origin.get_display_metadata()
+            ),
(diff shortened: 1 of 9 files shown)


python_modules/dagster/dagster/_core/workspace/context.py
```diff
@@ -610,10 +610,11 @@ def _load_location(self, origin: RepositoryLocationOrigin) -> WorkspaceLocationE
             repository_location=location,
             load_error=error,
             load_status=WorkspaceLocationLoadStatus.LOADED,
-            display_metadata=location.get_display_metadata()
+            display_metadata=(location.get_display_metadata()
             if location
-            else origin.get_display_metadata(),
+            else origin.get_display_metadata()),
             update_timestamp=time.time(),
+            image=location.container_image if location else None,
```
alangenfeld, line 617: This information is already in display metadata right? Whats the motivation for repeating it? Just that we want to respect it as not-just-for-display?

alangenfeld (approved):
small concern that if this information is duplicated and not consistently sourced it could get weird / be confusing in the future but hard to imagine `image` specifically being problematic like that

python_modules/dagster/dagster/_core/workspace/context.py
gibsondan: It's not in display metadata in serverless, and I am plotting to use it for something other than just display

PR #12597, opened 2023-02-28
docs for graphs that depend on assets

sryza:
### Summary & Motivation

Motivated by this feedback: https://dagster.slack.com/archives/C01U5LFUZJS/p1677548018200809

### DIFF
--- a/docs/content/concepts/ops-jobs-graphs/graphs.mdx
+++ b/docs/content/concepts/ops-jobs-graphs/graphs.mdx
@@ -223,6 +223,7 @@ Using dynamic outputs, you can duplicate portions of a graph at runtime. Refer t
 ## Defining and constructing dependencies
 
 - [Defining nothing dependencies](#defining-nothing-dependencies)
+- [Loading an asset as an input](#loading-an-asset-as-an-input)
 - [Constructing dependencies](#constructing-dependencies)
 
 ### Defining nothing dependencies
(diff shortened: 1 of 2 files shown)


docs/content/concepts/ops-jobs-graphs/graphs.mdx
```diff
@@ -262,6 +263,36 @@ Note that in most cases, it is usually possible to pass some data dependency. In
 
 Dagster also provides more advanced abstractions to handle dependencies and IO. If you find that you are finding it difficult to model data dependencies when using external storage, check out [IO managers](/concepts/io-management/io-managers).
 
+### Loading an asset as an input
+
+You can supply an asset as an input to one of the ops in a graph. Dagster can then use the [IO manager](/concepts/io-management/io-managers) on the asset to load the input value for the op.
```
erinkcochran87, line 268:
```suggestion
You can supply an asset as an input to one of the ops in a graph. Dagster can then use the [I/O manager](/concepts/io-management/io-managers) on the asset to load the input value for the op.
```

docs/content/concepts/ops-jobs-graphs/graphs.mdx
```diff
@@ -262,6 +263,36 @@ Note that in most cases, it is usually possible to pass some data dependency. In
...
+def send_emails(emails) -> None:
+    ...
+
+
+@job
+def send_emails_job():
+    send_emails(emails_to_send.to_source_asset())
+```
+
+We must use the <PyObject object="AssetsDefinition" method="to_source_asset" />, because <PyObject object="SourceAsset" plural /> are used to represent assets that other assets or jobs depend on, in settings where they won't be materialized themselves.
```
erinkcochran87, line 289: `SourceAsset` isn't showing up as plural here

docs/content/guides/dagster/how-assets-relate-to-ops-and-graphs.mdx
```diff
@@ -153,3 +153,5 @@ def send_emails_job():
 ```
 
 In this case, the asset - specifically, the table the job reads from - is only used as a data source for the job. It’s not materialized when the graph is run.
+
+You can read more about how this works [here](/concepts/ops-jobs-graphs/graphs#loading-an-asset-as-an-input).
```
erinkcochran87, line 157:
```suggestion
Refer to the [Graph documentation](/concepts/ops-jobs-graphs/graphs#loading-an-asset-as-an-input) for more info on how this works.
```

Generally, I think we should avoid vague link titles like `here` - it's helpful to provide context so the user has some expectation of where they're going when they click the link.

erinkcochran87 (commented):
Left a few comments, but looks good!

docs/content/concepts/ops-jobs-graphs/graphs.mdx
spenczar, line 268: How do I override the IOManager used by the asset? I can do that with `@asset(Ins={"key": AssetIn(..., input_manager_key: "overriding_io_mgr")})` for assets, how do I do it with ops?

docs/content/concepts/ops-jobs-graphs/graphs.mdx
```diff
@@ -262,6 +263,36 @@ Note that in most cases, it is usually possible to pass some data dependency. In
...
+
+If the asset is partitioned, then:
+
+- If the job is partitioned, the corresponding partition of the asset will be loaded.
+- If the job is not partitioned, then all partitions of the asset will be loaded.
```
spenczar, line 294: What does "all partitions will be loaded" mean for the shape of the value? Is it a list, or a dictionary, or a generator, or something else? I'm wondering how I ought to write my op to handle that.

docs/content/concepts/ops-jobs-graphs/graphs.mdx
sryza: Here are some docs: https://docs.dagster.io/concepts/io-management/unconnected-inputs#providing-an-input-manager-for-an-unconnected-input.

docs/content/concepts/ops-jobs-graphs/graphs.mdx
sryza: The type depends on the I/O manager implementation:
- The Pandas and PySpark type handlers of the DB IO managers (Snowflake, DuckDB, BigQuery) always return a single DataFrame, which can includes values from all the partitions.
- When loading an input that corresponds to multiple partitions, the UPathIOManager returns a dictionary that maps each input partition key to the input value for that partition key.

This needs better docs, but I don't think this is the right place to put them.

docs/content/concepts/ops-jobs-graphs/graphs.mdx
spenczar: Oh, I didn't put together that this is an "unconnected input" but I guess that makes sense, OK.

erinkcochran87 (approved):

PR #12605, opened 2023-02-28
Misc dynamic partitions updates

clairelin135:
This PR adds two things:
- Telemetry for dynamic partitions instance methods. For now I've only added telemetry for the `add_dynamic_partitions` and `get_dynamic_partitions` methods, logging the number of dynamic partitions as metadata. Thoughts on other entry points to capture? I also considered adding one for `delete_dynamic_partition`, though I figured we'd keep this entry point regardless of what telemetry tells us.
- Changes the `get_dynamic_partitions` method to return keys in the order that they exist in the database. Previously the index would cause the keys to be returned alphabetically.


sryza:
@clairelin135 not the end of the world if you're scrambling to get stuff in for the release, but it would be preferable for these two changes to be two separate PRs.

@dpeng817 do we have any guidelines on whether we should avoid adding telemetry to stuff that gets invoked too often? I'm worried that `get_partitions` will get invoked a lot.

python_modules/dagster/dagster_tests/storage_tests/utils/event_log_storage.py
```diff
@@ -2906,18 +2906,18 @@ def test_add_dynamic_partitions(self, storage):
         )
         partitions = storage.get_dynamic_partitions("foo")
         assert len(partitions) == 3
-        assert set(partitions) == {"foo", "bar", "baz"}
+        assert partitions == ["foo", "bar", "baz"]
```
dpeng817, line 2909: is this necessary?

dpeng817:
As far as frequency of calls go - I think it could potentially cause issues. I'm worried about the upload in terms of number of events - sometimes the telemetry server has issues if there's a ton of events, sounds like this might make that significantly worse. One thing we could do is stagger the logging of these events using kvs to set a cursor on how many calls have been made, and only upload every x times. 
@sryza @clairelin135

clairelin135:
@sryza @dpeng817 I've split this PR, keeping the telemetry changes here and adding the dynamic partitions ordering changes to https://github.com/dagster-io/dagster/pull/12641

I took Chris's suggestion of using kvs to only write the `get_partition_keys` logs 1 in 10 times.

sryza:
> I took Chris's suggestion of using kvs to only write the get_partition_keys logs 1 in 10 times.

Can we try to put some rough numbers on this? How many telemetry events per hour do we think is too many, and how many do we think we'll end up in like the 80th percentile case, with this scheme?

dpeng817:
we run into errors at around 4 mb of compressed data - assuming dagit is running an upload should be happening every hour - so if we take that information into account and think about a super big deployment, how often do we expect the instance method to be called per hour.

clairelin135:
I think these calls happen frequently:
- once per dagit request that loads partition information
- once per step or more, via partition mappings or when executing across a range of partitions
- could be called within sensor(s) every 30 seconds via `context.instance.get_dynamic_partitions`, which is not a recommended pattern but not out of the question for a user to do

Shot in the dark guess, but conservatively let's say 50 requests / hr + 50 steps per hour + 120 calls within sensors = > 200 calls?

Thinking more about this, our intention behind adding telemetry is mostly just to discern that there are users using dynamic partitions and to hopefully inform when we can mark the feature as stable. Since overloading the telemetry server is a possibility, I think we can get away with just tracking adding dynamic partitions or relying on #dagster-support for more qualitative feedback

dpeng817:
@clairelin135 this information can be determined from the `Definitions` object right? Wondering if this can just exist as metadata when we do an `update_repo_stats` call

clairelin135:
@dpeng817 sorry about the late response, your suggestion about fetching the info off the definitions object makes a lot of sense. I updated the PR to follow your suggestion

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1789,6 +1789,7 @@ def get_materialization_count_by_partition(
     @traced
     def get_dynamic_partitions(self, partitions_def_name: str) -> Sequence[str]:
         check.str_param(partitions_def_name, "partitions_def_name")
+
```
sryza, line 1792: spurious change?

python_modules/dagster/dagster/_core/telemetry.py
```diff
@@ -460,12 +460,26 @@ def log_external_repo_stats(
...
+        ExternalDynamicPartitionsDefinitionData,
+    )
 
     check.inst_param(instance, "instance", DagsterInstance)
     check.str_param(source, "source")
     check.inst_param(external_repo, "external_repo", ExternalRepository)
     check.opt_inst_param(external_pipeline, "external_pipeline", ExternalPipeline)
 
+    def _get_num_dynamic_partitioned_assets(external_asset_nodes) -> int:
+        return sum(
```
sryza, line 473: I think this could be written slightly more simply as

```python
    sum(
                1
                if asset.partitions_def_data
                and isinstance(asset.partitions_def_data, ExternalDynamicPartitionsDefinitionData)
                for asset in external_asset_nodes
        )

File: python_modules/dagster/dagster/_core/telemetry.py [hunk 4]
```diff
@@ -460,12 +460,26 @@ def log_external_repo_stats(
...
+    from dagster._core.host_representation.external_data import (
+        ExternalDynamicPartitionsDefinitionData,
+    )
 
     check.inst_param(instance, "instance", DagsterInstance)
     check.str_param(source, "source")
     check.inst_param(external_repo, "external_repo", ExternalRepository)
     check.opt_inst_param(external_pipeline, "external_pipeline", ExternalPipeline)
 
+    def _get_num_dynamic_partitioned_assets(external_asset_nodes) -> int:
```
sryza, line 472: input type annotation plz

sryza (commented):
Small comment, otherwise looks good - will leave it up to @dpeng817 for final approve because he understands this code better than me

dpeng817 (approved):
lgtm!

PR #12607, opened 2023-03-01
Add resource support for more complex sensor types

benpankow:
## Summary

Stacks on #12401, adding resource support for all of our various complex sensor types (`@asset_sensor`, `@multi_asset_sensor`, `@freshness_policy_sensor`, and `@run_status_sensor`).

For most cases, this is straightforward replication of what's done in #12401. There are a few interesting cases, e.g. asset sensors currently take two inputs (context and event log entry). This is now set up so that the first two non-resource params are treated as context and event log entry, respectively.

## Test Plan

New unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_sensor_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_sensor_definition.py
@@ -1,8 +1,10 @@
 import inspect
-from typing import TYPE_CHECKING, Callable, Optional, Sequence
+from typing import Any, Callable, NamedTuple, Optional, Sequence, Set
 
 import dagster._check as check
 from dagster._annotations import public
+from dagster._core.decorator_utils import get_function_params
+from dagster._core.definitions.resource_annotation import get_resource_args
 
 from .events import AssetKey
 from .run_request import RunRequest, SkipReason
@@ -10,14 +12,29 @@ from .sensor_definition import (
     DefaultSensorStatus,
     RawSensorEvaluationFunctionReturn,
     SensorDefinition,
-    SensorEvaluationContext,
     SensorType,
+    validate_and_get_resource_dict,
 )
 from .target import ExecutableDefinition
 from .utils import check_valid_name
 
-if TYPE_CHECKING:
-    from dagster._core.events.log import EventLogEntry
+
+class ContextAndEventLogEntryParamNames(NamedTuple):
+    context_param_name: Optional[str]
+    event_log_entry_param_name: Optional[str]
+
+
+def get_context_and_event_log_entry_param_names(fn: Callable) -> ContextAndEventLogEntryParamNames:
+    """
+    Determines the names of the context and event log entry parameters for an asset sensor function.
+    These are assumed to be the first two non-resource params, in order (context param before event log entry).
+    """
+    resource_params = {param.name for param in get_resource_args(fn)}
+
+    non_resource_params = [
+        param.name for param in get_function_params(fn) if param.name not in resource_params
+    ] + [None, None]
+    return ContextAndEventLogEntryParamNames(*non_resource_params[:2])
 
 
 class AssetSensorDefinition(SensorDefinition):
@@ -51,7 +68,7 @@ class AssetSensorDefinition(SensorDefinition):
         asset_key: AssetKey,
         job_name: Optional[str],
         asset_materialization_fn: Callable[
-            [SensorEvaluationContext, "EventLogEntry"],
+            ...,
             RawSensorEvaluationFunctionReturn,
         ],
         minimum_interval_seconds: Optional[int] = None,
@@ -59,14 +76,24 @@ class AssetSensorDefinition(SensorDefinition):
         job: Optional[ExecutableDefinition] = None,
         jobs: Optional[Sequence[ExecutableDefinition]] = None,
         default_status: DefaultSensorStatus = DefaultSensorStatus.STOPPED,
+        required_resource_keys: Optional[Set[str]] = None,
     ):
         self._asset_key = check.inst_param(asset_key, "asset_key", AssetKey)
 
         from dagster._core.events import DagsterEventType
         from dagster._core.storage.event_log.base import EventRecordsFilter
 
-        def _wrap_asset_fn(materialization_fn):
-            def _fn(context):
+        resource_arg_names: Set[str] = {
+            arg.name for arg in get_resource_args(asset_materialization_fn)
+        }
+
+        combined_required_resource_keys = (
+            check.opt_set_param(required_resource_keys, "required_resource_keys", of_type=str)
+            | resource_arg_names
+        )
+
+        def _wrap_asset_fn(materialization_fn) -> Any:
+            def _fn(context) -> Any:
                 after_cursor = None
                 if context.cursor:
                     try:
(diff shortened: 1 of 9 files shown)


benpankow:
* master
  * **PR #12401** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12401" target="_blank">Graphite</a> 
    * **PR #12607** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12607" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12607?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/asset_sensor_definition.py
```diff
@@ -1,23 +1,43 @@
...
-    SensorEvaluationContext,
     SensorType,
+    validate_and_get_resource_dict,
 )
 from .target import ExecutableDefinition
 from .utils import check_valid_name
 
 if TYPE_CHECKING:
-    from dagster._core.events.log import EventLogEntry
+    pass
```
OwenKephart, line 22: should be removed it seems?

python_modules/dagster/dagster/_core/definitions/asset_sensor_definition.py
```diff
@@ -91,7 +121,29 @@ def _fn(context):
...
+
+                resource_args_populated = validate_and_get_resource_dict(
+                    context.resources, name, resource_arg_names
+                )
+
+                # Build asset sensor function args, which can include any subset of
+                # context arg, event log entry arg, and any resource args
+                args = {
+                    **resource_args_populated,
+                    **({context_param_name: context} if context_param_name else {}),
```
OwenKephart, line 138: matter of taste (so feel free to ignore), but I think this might be easier to read as:

```
args = resource_args_populated
if context_param_name:
    args[context_param_name] = context
if event_log_entry_param_name:
    args[event_log_entry_param_name] = event_record.event_log_entry
```

python_modules/dagster/dagster_tests/definitions_tests/test_sensor_invocation.py
```diff
@@ -158,6 +159,123 @@ def my_cm_resource(_) -> Iterator[str]:
...
+    context = build_run_status_sensor_context(
+        sensor_name="status_sensor",
+        dagster_instance=instance,
+        dagster_run=dagster_run,
+        dagster_event=dagster_event,
+        context=build_sensor_context(resource_defs={"my_resource": MyResource(a_str="bar")}),
+    )
+
+    status_sensor(context)
+    status_sensor_no_context(context)
```
OwenKephart, line 267: this feels off to me -- why do we invoke it like this instead of:

```
status_sensor_no_context(my_resource=MyResource(a_str="bar"))
```

As a user, it would be very hard to figure out that I had to invoke build_run_status_sensor_context and build_sensor_context in order to pass my resources through to the underlying function

OwenKephart (commented):
For the most part, this makes sense, but I'm confused about the direct invocation behavior. It seems like we should allow users to pass resource instances as arguments rather than threading them through the context

python_modules/dagster/dagster_tests/definitions_tests/test_sensor_invocation.py
benpankow: Simplified to just require `build_run_status_sensor_context` (nixing the weird nested context requirement). See my other comment about directly passing resources - I think that's the right impulse and something I can handle in a follow up.

benpankow:
>  It seems like we should allow users to pass resource instances as arguments rather than threading them through the context

Yeah, Nick had this feedback too for the stacked change. I think it's worth adding support for doing so in another PR - the reason I didn't enable that here was just to keep the surface area small. It's easier to funnel everything through the context vs trying to meld multiple inputs (e.g. just context, context plus resources etc).

Did a pass on this PR simplifying a lot of the changes based on #12401 helper methods.

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -616,21 +684,33 @@ def _wrapped_fn(
...
+
                         sensor_return = run_status_sensor_fn(
-                            RunStatusSensorContext(  # type: ignore
-                                sensor_name=name,
-                                dagster_run=pipeline_run,
-                                dagster_event=event_log_entry.dagster_event,
-                                instance=context.instance,
-                                context=context,
-                            )
+                            **contest_param,
```
OwenKephart, line 710: context_param

python_modules/dagster/dagster_tests/definitions_tests/test_sensor_invocation.py
OwenKephart: when that change lands, would we remove the ability to invoke sensors this way? my inclination would be that we shouldn't let people invoke sensors that have no context param with `build_..._context` in that future world, which makes me tempted to either not release this without the companion change, or live with these sensors not being directly-invokable until that change lands.

schrockn: @benpankow is the change to allow for direct resource passing particularly challenging. I'm inclined to agree wiht @OwenKephart that we should include that if possible.

python_modules/dagster/dagster/_core/definitions/asset_sensor_definition.py
```diff
@@ -1,23 +1,39 @@
...
     SensorType,
+    validate_and_get_resource_dict,
 )
 from .target import ExecutableDefinition
 from .utils import check_valid_name
 
-if TYPE_CHECKING:
-    from dagster._core.events.log import EventLogEntry
+
+class ContextAndEventLogEntryParamNames(NamedTuple):
```
schrockn, line 22: `AssetSensorParamNames` seems a little more accurate/future-proof

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -794,6 +762,67 @@ def build_sensor_context(
...
 
+T = TypeVar("T")
+
+
+def get_or_create_sensor_context_base(
+    fn: Callable,
+    *args: Any,
+    context_type: Type[T],
+    **kwargs: Any,
+) -> Optional[T]:
```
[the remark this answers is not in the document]
benpankow: Changed to `get_sensor_context_from_args_or_kwargs` 👍

python_modules/dagster/dagster/_core/definitions/freshness_policy_sensor_definition.py
```diff
@@ -128,6 +133,7 @@ def __new__(
             if previous_minutes_late is not None
             else None,
             instance=check.inst_param(instance, "instance", DagsterInstance),
+            resources=resources or ScopedResourcesBuilder().build_empty(),
```
schrockn, line 136: nit: `build_empty` makes more sense as a static method

python_modules/dagster/dagster/_core/definitions/scoped_resources_builder.py
```diff
@@ -103,3 +103,7 @@ class _ScopedResources(
                 ...
 
             return _ScopedResources(**resource_instance_dict)  # type: ignore[call-arg]
+
+    def build_empty(self) -> Resources:
+        """Returns an empty Resources object, equivalent to ScopedResourcesBuilder.build(None)."""
```
schrockn, line 108: equivalent to `ScopedResourcesBuilder().build(None)` no?

schrockn (approved):
👍🏻

please heed final comment re: ScopedResourcesBuilder

PR #12610, opened 2023-03-01
Allow sensors that don't target a particular job (or jobs) to launch runs against a particular job in the repository

gibsondan:
### Summary & Motivation
This extends sensor functionality to allow untargeted jobs - given that we already allow you to have run requests that target arbitrary assets I don't see any particular reason to not allow this as well.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/sensor_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/sensor_definition.py
@@ -465,7 +465,7 @@ class SensorDefinition:
                 else:
                     check.failed("Expected a single SkipReason: received multiple SkipReasons")
 
-        self.check_valid_run_requests(run_requests)
+        self.check_valid_run_requests(run_requests, context)
 
         if self._asset_selection:
             run_requests = [
@@ -495,17 +495,11 @@ class SensorDefinition:
                 targets.append(target.load())
         return targets
 
-    def check_valid_run_requests(self, run_requests: Sequence[RunRequest]):
+    def check_valid_run_requests(
+        self, run_requests: Sequence[RunRequest], context: SensorEvaluationContext
+    ):
         has_multiple_targets = len(self._targets) > 1
         target_names = [target.pipeline_name for target in self._targets]
-
-        if run_requests and not self._targets and not self._asset_selection:
-            raise Exception(
-                f"Error in sensor {self._name}: Sensor evaluation function returned a RunRequest "
-                "for a sensor lacking a specified target (job_name, job, or jobs). Targets "
-                "can be specified by providing job, jobs, or job_name to the @sensor "
-                "decorator."
-            )
 
         for run_request in run_requests:
             if run_request.job_name is None and has_multiple_targets:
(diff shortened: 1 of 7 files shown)


python_modules/dagster/dagster/_daemon/sensor.py
```diff
@@ -638,8 +639,12 @@ def _evaluate_sensor(
...
                 )
 
-        target_data: ExternalTargetData = check.not_none(
-            external_sensor.get_target_data(run_request.job_name)
+        target_data: Optional[ExternalTargetData] = external_sensor.get_target_data(
+            run_request.job_name
+        ) or ExternalTargetData(
+            pipeline_name=check.not_none(run_request.job_name),
+            mode=DEFAULT_MODE_NAME,
+            solid_selection=None,
```
alangenfeld, line 647: this is a standout bit of weirdness - especially given `asset_selection` being on `RunRequest`

python_modules/dagster/dagster_tests/definitions_tests/test_sensor_invocation.py
```diff
@@ -121,27 +121,6 @@ def test_instance_access_with_mock():
...
-    ):
-        no_job_sensor.check_valid_run_requests(
-            [
-                RunRequest(
-                    run_key=None,
-                    run_config=None,
-                    tags=None,
-                )
-            ]
-        )
```
alangenfeld, line 142: this should still fail albeit in a different way - with no singular target and `job_name` not specified the `RunRequest` is invalid. Is this covered somewhere else? if not worth modifying this test instead of deleting.

python_modules/dagster/dagster/_daemon/sensor.py
```diff
@@ -638,8 +639,12 @@ def _evaluate_sensor(
                     asset_selection=stale_assets, stale_assets_only=False
                 )
 
-        target_data: ExternalTargetData = check.not_none(
-            external_sensor.get_target_data(run_request.job_name)
+        target_data: Optional[ExternalTargetData] = external_sensor.get_target_data(
+            run_request.job_name
+        ) or ExternalTargetData(
```
alangenfeld, line 644: nits:
prefer explicit if / else-ing w/ comments here to make it clear what the assumptions we are operating against are

might be clearer to just move away from `ExternalTargetData` and move this handling in to `PipelineSelector` construction

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -495,29 +495,33 @@ def load_targets(
...
+            elif run_request.job_name and target_names and run_request.job_name not in target_names:
                 raise Exception(
                     f"Error in sensor {self._name}: Sensor returned a RunRequest with job_name "
                     f"{run_request.job_name}. Expected one of: {target_names}"
                 )
+            elif (
+                run_request.job_name
+                and context.repository_def
+                and not context.repository_def.has_job(run_request.job_name)
+                and not context.repository_def.has_pipeline(run_request.job_name)
```
alangenfeld, line 519: nit: you can drop pipeline since we no longer support them

alangenfeld (commented):
Do we need to update any doc blocks? 

The downside to this is that we don't know the sensor will kick off these jobs which prevents us from associating them in the UI. For the use case that inspired this of a sensor launching any from a large set of jobs I think thats fine, but theres nothing in the API preventing someone from using this pattern for a sensor that always fires the same job but having a degraded UI experience of those entities no longer being associated. Docs & examples steering people in the right direction might be all thats needed, but i do think it warrants some consideration. A more aggressive approach would be to have an explicit "any job"  opt-in on the sensor, which feels like a lot to me.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
gibsondan: Heh my test failed if I left this out - I will update the test to use a job and take this out

gibsondan: I left this in b/c it seems like you can still write code that hits it?

alangenfeld (approved):

alangenfeld (changes requested):
this has been idle for a while - bouncing to your queue to decide how to proceed

PR #12611, opened 2023-03-01
[dagit] Switch to staleStatus for stale tags, display causes in tooltips

bengotow:
### Summary & Motivation

This PR switches Dagit from a client-side implementation of the "stale" keyword, where it compared current and projected logical versions on it's own, to a version where it receives and displays a server-provided staleStatus.

This PR also adds support for staleStatusCauses, which gives the user a bit more context about why the stale tag is being displayed. The UI does not expect that causes are always available and does not use causes when choosing to display the stale state -- it's strictly for the tooltips.

### Pictures

image

image

image

image

- I updated the AssetNode storybooks and checked them. I also verified the appearance of the AssetTable, because status tags also appear there.

- I verified visually that my fairly complex local asset graph did not change at all (states or colors) between `main` and this branch, which is a good check that the whole switch to staleStatus had no user-facing impact.  NOTE: There is one small discrepancy, assets that have not been materialized since the introduction of this feature now appear stale. (I think we're on board with that though?)

- I verified visually that "never materialized" / "MISSING" assets render the same way. 

- I verified that at least with my test repo of ~30 assets, there is not a noticeable performance regression. I do think we should keep an eye on performance though or clip staleStatusCauses to a fixed length.

- I verified that this looks ok in Safari as well as Firefox.

image

image


smackesey:
Nice work!

Looking at the reasons displayed in the prototype tooltip, IMO it is potentially confusing in a few places.

### Unclear path structure of reasons

It says "updated input (inp2)" despite the fact that "inp2" isn't a (first-order) input of the target. I see two possible solutions:

1. List the asset key (this is a field on the reason) alongside the reason text. So:

```
- [oof4] stale input (oof3)
- [oof3] updated input (inp2)
```

Optionally this could use indentation-based nesting:

```
- [oof4] stale input (oof3)
    - [oof3] updated input (inp2)
```

Particularly with the nesting, I think this is considerably clearer than the prototype. But I recognize that in a large asset graph this display could grow overly large.

2. Display only root causes.

In this case, we'd want to change the descriptor for the root cause so that it's relative to the target node-- so, "input" to "ancestor" for 2nd-order or higher dependencies. We could compute this server-side.

```
- updated ancestor (inp2)
```

I prefer solution (2), with the full staleness "backtrace" available only as debug-level info.

### Non-obvious reason descriptors

The difference between "stale input" vs "updated input" isn't immediately obvious. I don't think we have a solution for that in-tooltip, but it would be nice if an explanation was accessible somewhere in the UI. This problem could be sort-of solved by using solution (2) above (root causes only). Then maybe we could provide an explanatory info modal when viewing the full backtrace in the "debug" view.

bengotow:
Hey @smackesey  will let @sryza and others weigh in on this. I'd also vote for #2 -- I don't see much value in listing out a whole indented tree of stalness because most likely you just want to know the "cause" of staleness and that is the earliest child.  I think the "which reason" part of this work is more interesting than the "which assets" part, which is already sorta evident in the graph.

Re: reasons, I wonder if we should just make a docs page that highlights all the supported reasons. It'd be cool to see how complete the system is but I'm not sure how we could present that in the UI.

smackesey:
> I particularly struggle with what updated input means. I think it's just telling me that the upstream asset is not stale? Or maybe that the code version has changed?

Yeah there is a fair amount of complexity in the reasons. Suppose we have two assets `A -> B`. Then `B` could be stale because `A` is stale, _or_ `B` could be stale because it simply hasn't yet been materialized with the latest (non-stale) value of `A`. That's what "updated input" means.

>Or maybe that the code version has changed?

It will say "updated code version" if code version changed.

>Do we display the full asset key in the tooltip? In most cases that is much longer than the what is shown in the screenshot which might look awkward here.

Particularly with root causes, we're going to want some representation of the asset key.  For example, you might have an asset 4 steps upstream that had its code version changed-- you want to see which asset it is at a glance.

Maybe a width limited badge that shows the full key on hover?

bengotow:
I've updated the front-end code here to show the key along with each staleness reason. Per discussion with @smackesey it sounds like this is the best thing for now, and we can iterate on the reasons on the backend.

image

salazarm (approved):
lgtm!

PR #12615, opened 2023-03-01
[UI] Dynamic partition creation and selection

salazarm:
### Summary & Motivation

Adds the ability to create dynamic partitions
Screen Shot 2023-03-01 at 7 23 20 AM

Screen Shot 2023-03-01 at 7 24 22 AM

Screen Shot 2023-03-01 at 7 23 24 AM

I tested this with `dagit dev` and also in `storybook`. I need to spend some time digging through this some more to setup tests


js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
```diff
@@ -364,6 +384,11 @@ const LaunchAssetChoosePartitionsDialogBody: React.FC<Props> = ({
                   selections.length === 2 ? selections[1 - idx].selectedRanges : undefined,
                 ),
               }}
+              isDynamic={assets.some(
```
salazarm, line 387: I'm not toooo sure about this logic. @bengotow

js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
```diff
@@ -372,6 +397,11 @@ const LaunchAssetChoosePartitionsDialogBody: React.FC<Props> = ({
                   ),
                 )
               }
+              partitionDefinitionName={
```
salazarm, line 400: Also, not sure about this since after we spoke you said there could be multiple partition definition names?

js_modules/dagit/packages/core/src/assets/usePartitionHealthData.tsx
```diff
@@ -469,14 +467,47 @@ export function usePartitionHealthData(assetKeys: AssetKey[], assetLastMateriali
       ]);
     };
     run();
-  }, [client, missingKeyJSON, assetLastMaterializedAt]);
+  }, [client, missingKey, assetLastMaterializedAt]);
```
bengotow, line 470: I'm not sure this is memoizable because it's an object containing a path and not a string, maybe let's leave this as-is for now and make a second PR with the JSON => no-JSON change when we have more time to test that the memoization is still working properly.

I think there are a few places in the UI where the asset keys are actually built into this array so they might be different objects between renders?

js_modules/dagit/packages/core/src/assets/usePartitionHealthData.tsx
```diff
@@ -469,14 +467,47 @@ export function usePartitionHealthData(assetKeys: AssetKey[], assetLastMateriali
...
     const assetKeyJSONs = JSON.parse(assetKeyJSON);
     return result.filter((r) => assetKeyJSONs.includes(JSON.stringify(r.assetKey)));
   }, [assetKeyJSON, result]);
 }
 
+/**
+ *  Like usePartitionHealthData, but fetches queries in parallel and allows refetching them all in parallel
+ * @returns
+ */
+export function usePartitionHealthData2(assetKeys: AssetKey[], cacheKey: string) {
```
bengotow, line 482: I think we could make the main `usePartitionHealthData` run more queries in parallel but I don't think we want to run them ALL in parallel because the user can actually select hundreds of assets when opening the launch backfill modal. (If you don't make any selection at all, the modal is opened for all your assets), and I think this will flood the backend with pretty expensive requests for asset health data. 

It is a bummer that the serial loading is slow, but we might have to fix that separately.

I think if you want, you can pass your `cacheKey` to the original implementation as a fake `assetLastMaterializedAt` and that will bust the cache the same way. (Could also rename that param to cacheKey because I think that explains what it does nicely!)

js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
```diff
@@ -317,6 +317,18 @@ export const useMaterializationAction = (preferredJobName?: string) => {
...
+            const result = await client.query<
+              LaunchAssetLoaderQuery,
+              LaunchAssetLoaderQueryVariables
+            >({
+              query: LAUNCH_ASSET_LOADER_QUERY,
+              variables: {assetKeys: state.assets.map(({assetKey}) => ({path: assetKey.path}))},
+            });
+            const assets = result.data.assetNodes;
+            const next = await stateForLaunchingAssets(client, assets, false, preferredJobName);
+            setState(next);
```
bengotow, line 330: Not urgent, but it might be nice to handle the next.state === 'error' scenario here. I wonder if we could pull lines 239-262 out of  `onClick` into a separate helper, and have both that onClick method and this one call that to avoid having two places where we need to make the query.

js_modules/dagit/packages/core/src/asset-graph/useLiveDataForAssetKeys.tsx
```diff
@@ -187,6 +187,9 @@ const ASSETS_GRAPH_LIVE_QUERY = gql`
   query AssetGraphLiveQuery($assetKeys: [AssetKeyInput!]!) {
     assetNodes(assetKeys: $assetKeys, loadMaterializations: true) {
       id
+      type {
+        name
+      }
```
bengotow, line 192: I'm not totally sure this ended up being used in this PR?

js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
bengotow: I think you can simplify this a bit -- up above in this file, we determine a `displayedPartitionDefinition` that is the "primary" one we're going to let you interact with. (Typically there can only be one, but if the selection contains partition definitions that are mapped to each other, it can be valid to have more than one)

I think this can just be `displayedPartitionDefinition?.type === PartitionDefinitionType.DYNAMIC`

js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
bengotow: I think this can be `displayedPartitionDefinition?.name` actually 🙌

bengotow (approved):
This looks great! I left some comments inline but it seems like this is almost there 🙌

js_modules/dagit/packages/core/src/asset-graph/useLiveDataForAssetKeys.tsx
salazarm: This is how we determine that it's dynamic

js_modules/dagit/packages/core/src/assets/usePartitionHealthData.tsx
salazarm: Ah I see, sure. I made these changes when trying to adapt this hook but ended up abandoning it because having some stale health data and some not stale causes `mergeAssetHealth` to raise an exception.

js_modules/dagit/packages/core/src/asset-graph/useLiveDataForAssetKeys.tsx
salazarm: oh wait, you're right, not using this one.

PR #12618, opened 2023-03-01
Use to_python_identifier for naming objects derived from Assets/AssetKeys

simonvanderveldt:
### Summary & Motivation
Whilst looking into the newly added `@asset_graph` decorator I noticed that there was some duplication of the logic to derive names from an Asset. Given the logic that was there the already existing [`to_python_identifier()` method](https://github.com/dagster-io/dagster/blob/e403d3974c44254a9a500492e12230f2d6d326cc/python_modules/dagster/dagster/_core/definitions/events.py#L129) available on `AssetKey` objects.

Ran these changes locally with a `@graph_asset`, the Graph name is the same as it was before.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
@@ -732,9 +732,9 @@ class _GraphBackedAsset:
             if asset_in.partition_mapping
         }
 
-        op_graph = graph(
-            name="__".join(out_asset_key.path).replace("-", "_"), description=self.description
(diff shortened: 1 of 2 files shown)


sryza:
Thanks for the contribution @simonvanderveldt! Do you mind running `make black` to format the code? Otherwise, this looks good to me.

simonvanderveldt:
> Thanks for the contribution @simonvanderveldt! Do you mind running `make black` to format the code? Otherwise, this looks good to me.

@sryza Done!

sryza (approved):
Looks great - merging

PR #12619, opened 2023-03-01
Add root stale causes for assets

smackesey:
### Summary & Motivation

- Change `stale_causes` returned from `CachingStaleStatusResolver` to return a tree.
- Add `stale_root_causes` that returns a flat list of causes representing root causes.
- Change the `staleStatusCauses` GQL resolver to return just the root causes, which are more suitabel for display in dagit.
- Add "missing input" staleness reason.

New test testing for root causes


smackesey:
* master
  * **PR #12619** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12619" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12619?utm_source=stack-comment).

bengotow (approved):
Will leave the python changes to other folks, but I think this is a good move from the UI side of things 👍

PR #12599, opened 2023-02-28
Add failures to partition status cache

johannkm:

PR #12570, opened 2023-02-27
asset partitions numFailed

johannkm:
gql for counting failed partitions.

Update: to make it so that numMaterialized and numFailed can't total to greater than numPartitions, we now filter out failed partitions from materialized.

PR #12405, opened 2023-02-16
asset partitions numFailed

johannkm:

PR #12620, opened 2023-03-01
Gql failed partitions subset

johannkm:
New resolver

```
        assetNodes(pipeline: $pipelineSelector) {
            assetPartitions {
                ... on TimePartitions {
                    ranges {
                        startTime
                        endTime
                        startKey
                        endKey
                        status <~ we consolidate materialized and failed to a set of ranges
                    }
                }
                ... on DefaultPartitions {
                    materializedPartitions
                    failedPartitions
                    unmaterializedPartitions
                }
                ... on MultiPartitions {
                    ranges {
                        primaryDimStartKey
                        primaryDimEndKey
                        primaryDimStartTime
                        primaryDimEndTime
                        secondaryDim {
                            ... on TimePartitions {
                                ranges {
                                    startTime
                                    endTime
                                    startKey
                                    endKey
                                    status
                                }
                            }
                            ... on DefaultPartitions {
                                materializedPartitions
                                unmaterializedPartitions
                                failedPartitions
                            }
                        }
                    }
                }
            }
        }

```

We filter out materialized partitions that are also failed.

Currently leaving the `materializedPartitions` resolver for backcompat with the current frontend (otherwise we'd need to land both changes at once). It can be removed once the front end is updated


johannkm:
* master
  * **PR #12333** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12333" target="_blank">Graphite</a> 
    * **PR #12404** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12404" target="_blank">Graphite</a> 
      * **PR #12405** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12405" target="_blank">Graphite</a> 
      * **PR #12599** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12599" target="_blank">Graphite</a> 
        * **PR #12570** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12570" target="_blank">Graphite</a> 
          * **PR #12620** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12620" target="_blank">Graphite</a>  👈
    * **PR #12310** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12310" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12620?utm_source=stack-comment).

python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
```diff
@@ -215,6 +217,7 @@ class GrapheneAssetNode(graphene.ObjectType):
         partitions=graphene.List(graphene.String),
     )
     materializedPartitions = graphene.NonNull(GrapheneMaterializedPartitions)
+    assetPartitions = graphene.NonNull(GrapheneAssetPartitions)
```
johannkm, line 220: this vs assetPartitionStatuses?

clairelin135 (commented):
If I understand correctly it seems like now we're adding a new resolver that basically provides data that combines the failed + materialized statuses, but it seems like a lot of the existing materialization code is duplicated to handle the failed case as well.

I'm wondering if we can just remove the original "materializedPartitions" resolver and rely purely on the resolver created in this PR? Since it seems like in the places we want to view materialization statuses, we also want to view failed statuses.

@bengotow do you know if we need just the `materializedPartitions` resolver? One instance I can think of is to display materialization events per partition on the asset details page, but we do have a different resolver for that

johannkm:
Correct this replaces it

> Currently leaving the materializedPartitions resolver for backcompat with the current frontend (otherwise we'd need to land both changes at once). It can be removed once the front end is updated

is the thought behind leaving it for now, but we don't have to- would just have to wait to land until the front end is ready to go

clairelin135:
Ah I see. I can't see the diffs between the new behavior and the old behavior which makes this harder to review... if you guys dont feel strongly about this, is it okay if we remove the backcompat and just transition directly to using the new resolver? That means @bengotow would have to stack on top of this diff and merge the changes into this diff.

johannkm:
@clairelin135 updated to remove `materlializedPartitions`. Much easier review

python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
clairelin135: prefer assetPartitionStatuses as this currently sounds like a list of the partition keys

clairelin135 (approved):
This is looking good to me, though I think it's worth @bengotow looking at it too.

I also think it's worth testing that the gql queries return what's expected for failed partition ranges. Especially the multidimensional case, to confirm that we're doing the run length encoding correctly.

bengotow (approved):
This looks awesome! I like the renames as well. I will get a UI PR stacked on this branch tomorrow morning and we can merge them together since the resolver has been changed.

PR #12623, opened 2023-03-01
[1.2] make ScheduleEvaluationContext.scheduled_execution_time non-optional

sryza:
### Summary & Motivation

Motivation: doing basic schedule things raises a TypeError. E.g. look at the example that's changed in this PR.

In normal operation, this property will always be set, but it's possible for someone to construct a test context using `build_schedule_context` and leave it out.

This makes `ScheduleEvaluationContext` follow the same pattern that we use in other contexts when users omit fields when building a test context.

I would consider this a bug, but it's arguably a breaking change, so I'm going to merge it with 1.2 instead of in a patch release.


gibsondan (approved):
ah yes the only reason this was originally non-optional was for the long-lost cron-based scheduler

PR #12624, opened 2023-03-01
[dagit] Use GraphQL object based permissions where possible

hellendag:
### Summary & Motivation

Where possible, use newly-added permission fields on GraphQL objects, instead of relying on global permission values. This means cleaning up some existing `usePermissionsDEPRECATED` callsites. I'll try to knock out the rest of them in a followup.

- Add permission fields to existing fragments, rerun GraphQL generation
- Where deprecated global perms are used, replace them with references to these fields

I also updated the `Unloadables` test to use `MockedProvider` with fixtures instead of a `TestProvider`.

TS, lint, jest.

Load Dagit, verify that affected features (sensors, schedules, runs) are working as expected for OSS. We'll test this more in Cloud as well, with various permission states in place.


js_modules/dagit/packages/core/src/instigation/Unloadable.test.tsx
```diff
@@ -1,140 +1,62 @@
...
-          >
-            <Test />
-          </TestProvider>,
-        );
-      });
+      render(
+        <MockedProvider>
+          <UnloadableSensors sensorStates={[unloadableSensor]} />
+        </MockedProvider>,
+      );
```
hellendag, line 17: This is sooo much nicer

bengotow: This looks great! Does this UnloadableSensors component include a useQuery though? It seems odd that these tests all work without passing any mocks to the MockedProvider. I wonder if we could actually not wrap this in any Apollo provider at all?

js_modules/dagit/packages/core/src/runs/RunActionsMenu.tsx
```diff
@@ -303,30 +313,32 @@ export const RunBulkActionsMenu: React.FC<{
       <Popover
         content={
           <Menu>
-            {canTerminatePipelineExecution.enabled ? (
+            {canTerminateAny ? (
```
bengotow, line 316: Not strictly related to this PR, but I'd be in favor of showing these menu items grayed out with a tooltip rather than hiding them if it's possible -- just so people know that they can / should ask for permission

bengotow (approved):
This all looks great! Left a few inline comments but nothing blocking. 🙌

js_modules/dagit/packages/core/src/instigation/Unloadable.test.tsx
hellendag: There are lower-level components within the tree that do require it, unfortunately.

hellendag:
OSS compat test failure is not related. Merging.

PR #12628, opened 2023-03-01
Change format of staleness root causes and remove old resolvers

smackesey:
### Summary & Motivation

- Enhance clarity of GQL stale status reasons
- Drop `status` field from `GrapheneAssetStaleCause`
- Remove `projectedLogicalVersion` field on GQL assetNode
- Completely remove `projectedLogicalVersion` from code base

Updated logical versioning test suite

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/logical_version.py
+++ b/python_modules/dagster/dagster/_core/definitions/logical_version.py
@@ -236,6 +236,14 @@ class StaleStatusCause(NamedTuple):
     children: Optional[Sequence["StaleStatusCause"]] = None
 
 
+# Root reasons for staleness. Thes differ from `StaleStatusCause`in that there is no status
+# associated with them-- rather they are causes for the staleness of some downstream node.
+class StaleStatusRootCause(NamedTuple):
+    key: AssetKey
+    reason: str
+    dependency: Optional[AssetKey] = None
+
+
 class CachingStaleStatusResolver:
     """
     Used to resolve logical version information. Avoids redundant database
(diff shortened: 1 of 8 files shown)


smackesey:
* master
  * **PR #12628** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12628" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12628?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/logical_version.py
```diff
@@ -390,6 +384,22 @@ def _gather_leaves(
             for child in cause.children:
                 yield from self._gather_leaves(child, level=level + 1)
 
+    def _convert_to_root_cause(self, cause: StaleStatusCause) -> StaleStatusRootCause:
+        if cause.reason == "updated dependency logical version":
+            assert cause.dependency, "[updated input] cause must have a dependency"
+            return StaleStatusRootCause(
+                cause.dependency,
+                "updated logical version",
```
bengotow, line 392: If we're making the shift to `data version` soon maybe we go ahead and change it here

python_modules/dagster/dagster/_core/definitions/logical_version.py
```diff
@@ -390,6 +384,22 @@ def _gather_leaves(
...
+            )
+        elif cause.reason == "updated dependency timestamp":
+            assert cause.dependency, "[updated input] cause must have a dependency"
+            return StaleStatusRootCause(
+                cause.dependency,
+                "updated timestamp",
```
bengotow, line 398: This seems fine but I wonder if text like "dependency materialization is newer" would be more  clear. The tooltip will look weird if these get super long, but i think it'd be fine to make them a bit longer.

bengotow (approved):
The description changes here look fine to me, I threw in some ideas but overall I expect this will be iterated on a lot. Getting rid of those resolvers is nice!

python_modules/dagster/dagster/_core/definitions/logical_version.py
smackesey: The key of this is no longer the asset that is stale, but the asset whose timestamp was updated-- I chose "updated timestamp" because it is essentially a substitute for "updated logical version" for the case where logical versions aren't available.

python_modules/dagster/dagster/_core/definitions/logical_version.py
smackesey: I have a stack of PRs that makes the "data version" rename but Nick hasn't given final blessing, there is still a question whether we'll use another name

Issue #12604, opened 2023-02-28
FutureWarning in dagster_snowflake_pandas


sethkimmel3:
### Dagster version

1.1.20

### What's the issue?

Small warning from [this line](https://github.com/dagster-io/dagster/blob/116ec378b77a195182b9b89f3d44ac9e59e54505/python_modules/libraries/dagster-snowflake-pandas/dagster_snowflake_pandas/snowflake_pandas_type_handler.py#L80):

`FutureWarning: iteritems is deprecated and will be removed in a future version. Use .items instead.`

### What did you expect to happen?

Should probably be replaced.

### How to reproduce?

Run the code!

### Deployment type

Local

### Additional information

Happy to submit a PR, would love to see how Dagster tests changes.

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

sryza:
> Happy to submit a PR, would love to see how Dagster tests changes.

If you'd be up for submitting a PR, that would be great. Before changing this, we should figure out if this will break compatibility with earlier versions of Pandas that users might still be relying on.

sethkimmel3:
Done! See https://github.com/dagster-io/dagster/pull/12629

PR #12629, opened 2023-03-01
iteritems -> items

sethkimmel3:
### Summary & Motivation

Fixes https://github.com/dagster-io/dagster/issues/12604

`.iteritems` is [deprecated since 1.50 ](https://pandas.pydata.org/docs/whatsnew/v1.5.0.html#other-deprecations) and will be removed in a future version. Changing to `.items`. (https://github.com/pandas-dev/pandas/pull/45321).

I did not, but https://github.com/pandas-dev/pandas/pull/45321 appears to indicate it's a safe change to make. If I can somehow install this version of dagster on my machine, I'm happy to test using the code that surfaced the issue, but I'm not sure how to do that.

### DIFF
--- a/python_modules/libraries/dagster-snowflake-pandas/dagster_snowflake_pandas/snowflake_pandas_type_handler.py
+++ b/python_modules/libraries/dagster-snowflake-pandas/dagster_snowflake_pandas/snowflake_pandas_type_handler.py
@@ -77,7 +77,7 @@ class SnowflakePandasTypeHandler(DbTypeHandler[pd.DataFrame]):
                 TableSchema(
                     columns=[
                         TableColumn(name=name, type=str(dtype))
-                        for name, dtype in obj.dtypes.iteritems()
+                        for name, dtype in obj.dtypes.items()
                     ]
                 )
             ),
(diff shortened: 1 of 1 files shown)


sryza:
@sethkimmel3 do you know when `items` was added? Just want to make sure we're not breaking people who are on older-but-still-used Pandas versions

sethkimmel3:
@sryza looks like it's been around since 0.5.0 (or earlier): https://github.com/pandas-dev/pandas/blob/8f79f7c6e01874bbf7f3b9bc38af01551c90121e/pandas/core/series.py#L447 (pending python3 compatibility).

sryza:
Great - that's a long time. This LGTM. Kicking off a build.

sryza:
pyright has this complaint:

```
[2023-03-01T20:56:02Z] /workdir/python_modules/libraries/dagster-snowflake-pandas/dagster_snowflake_pandas/snowflake_pandas_type_handler.py:
[2023-03-01T20:56:02Z]   79:41: Argument of type "Hashable" cannot be assigned to parameter "name" of type "str" in function "__new__"
[2023-03-01T20:56:02Z]   "Hashable" is incompatible with "str" (reportGeneralTypeIssues)
```

To avoid this, I think probably just casting to a string is best?

sethkimmel3:
@sryza that seems sensible, but I'm also not sure why it would be suddenly flag this as an issue, as it appears `.itemitems` has just been an alias for `.items` and calls that function directly. What version of pandas is dagster using under the hood?

sryza:
> What version of pandas is dagster currently using?

I don't believe we pin any particular version. In our CI, I believe we use the latest version (just `pip install pandas`).

sethkimmel3:
I'm assuming this is what you meant by casting to str. Let's see if the build likes this!

sryza (approved):
Looks good!

PR #12632, opened 2023-03-01
Flatten overlapping time window subsets in order of priority

clairelin135:
This PR provides a helper method that accepts a list of `TimeWindowPartitionsSubset`s and their statuses (failed or materialized). Flattens these subsets into a list of range statuses based on the priority of the statuses.

This helper function:
- calculates the overlap between the two subset's ranges
- removes the overlap from the lower priority subset's ranges

The hope is that this function in the future can be adapted to accept `n` subsets and can flatten based on the priority of each subset.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
@@ -2,6 +2,7 @@ import hashlib
 import json
 import re
 from datetime import datetime
+from enum import Enum
 from typing import (
     AbstractSet,
     Any,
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -1494,3 +1495,181 @@ def __contains__(self, partition_key: str) -> bool:
...
+                curr_time_window_i += 1
+
+    return new_time_windows
+
+
+def fetch_flattened_time_window_ranges(
+    subset_statuses: Sequence[Tuple[TimeWindowPartitionsSubset, PartitionRangeStatus]],
+) -> Sequence[PartitionTimeWindowStatus]:
+    """
+    Given a list of TimeWindowPartitionsSubsets that are ordered by priority (highest priority first),
```
johannkm, line 1640: nit, looks like we sort by priority regardless of the order of the input

johannkm (approved):
Thanks for putting this up!

PR #12633, opened 2023-03-01
[dagster-snowflake] fix inconsistencies in snowflake resource

### LINKED ISSUE #12084: dagster-snowflake `execute_queries` appends results to a DataFrame instead of a list when `use_pandas_result` is True
### Dagster version

1.1.15

### What's the issue?

[`execute_queries()` includes this code](https://docs.dagster.io/_modules/dagster_snowflake/resources#SnowflakeConnection.execute_queries):

```
        if use_pandas_result:
            import pandas as pd

            results = pd.DataFrame()
        else:
            results = []
        with self.get_connection() as conn:
            with closing(conn.cursor()) as cursor:
                for sql in sql_queries:
                    if sys.version_info[0] < 3:
                        sql = sql.encode("utf-8")
                    self.log.info("Executing query: " + sql)
                    parameters = dict(parameters) if isinstance(parameters, Mapping) else parameters
                    cursor.execute(sql, parameters)
                    if fetch_results:
                        if use_pandas_result:
                            results = results.append(cursor.fetch_pandas_all())  # type: ignore
                        else:
                            results.append(cursor.fetchall())  # type: ignore

        return results if fetch_results else None
```

If `use_pandas_result` is `True`, then each queries is retrieved with `cursor.fetch_pandas_all()`, which returns a DataFrame. However, at the top of the code block, you'll see that `results` is initialized as a DataFrame. What this means is that the code attempts to append each query's results to the `results` DataFrame. This causes failure when the queries return significantly different results. 

I expect that `execute_queries()` will return a list of DataFrames when the `use_pandas_result` flag is True. This would fix it:

```
        if use_pandas_result:
            import pandas as pd

        results = []

        with self.get_connection() as conn:
            with closing(conn.cursor()) as cursor:
                for sql in sql_queries:
                    if sys.version_info[0] < 3:
                        sql = sql.encode("utf-8")
                    self.log.info("Executing query: " + sql)
                    parameters = dict(parameters) if isinstance(parameters, Mapping) else parameters
                    cursor.execute(sql, parameters)
                    if fetch_results:
                        if use_pandas_result:
                            results = results.append(cursor.fetch_pandas_all())  # type: ignore
                        else:
                            results.append(cursor.fetchall())  # type: ignore

        return results if fetch_results else None
```

### How to reproduce?

```
queries = [
    "USE SECONDARY ROLES ALL",
    "SELECT * FROM SOME_TABLE_WITH_MANY_COLUMNS LIMIT 5"
]

results = context.resources.snowflake.execute_queries(
    sql_queries=queries,
    fetch_results=True,
    use_pandas_result=True
)
```

### Deployment type

Docker Compose

### Deployment details

Deployment is not relevant to this problem

### Additional information

pandas.DataFrame.append is deprecated anyhow: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.append.html

Also note that the code example for `execute_queries()` contains an error. The first parameter is `sql_queries` and not `sql` as shown in the example.

jamiedemaria:
### Summary & Motivation
Fixes #12084

A user pointed out that the `execute_query` and `execute_queries` methods are inconsistent with each other:
* in `execute_query`, if `fetch_result` is True and `use_pandas_result` is True, the non pandas result is returned
* in `execute_queries`, if `fetch_result` is True and `use_pandas_result` is True, the pandas result is used
* 
* in `execute_query`, if `fetch_result` is False and `use_pandas_result` is True, the pandas result is returned
* in `execute_queries`, if `fetch_result` is False and `use_pandas_result` is True, nothing is returned

This PR makes the two methods consistent according to this chart

`fetch_result` |  `use_pandas_result`   |  outcome
True                 |     True                             | pandas result
True                 |     False                            | non-pandas result
False               |      True                             | ~pandas result~ no result + error (see comments below)
False               |      False                            | no results

 Open to other options for determining which result to return -  for example, we could make the two options mutually exclusive and error if both are True. Just needed to pick one to implement as a starting point for discussion

--- 
Additionally this fixes an issue in `execute_queries` when `use_pandas_result` is True. Before, all of the results of the queries were being put into a single dataframe, which would cause issues if the queries where for different tables. This updates the method to append the dataframes to a list

### DIFF
--- a/python_modules/libraries/dagster-snowflake-pandas/dagster_snowflake_pandas_tests/test_snowflake_pandas_type_handler.py
+++ b/python_modules/libraries/dagster-snowflake-pandas/dagster_snowflake_pandas_tests/test_snowflake_pandas_type_handler.py
@@ -311,7 +311,7 @@ def test_time_window_partitioned_asset():
         )
 
         out_df = snowflake_conn.execute_query(
-            f"SELECT * FROM {snowflake_table_path}", use_pandas_result=True
+            f"SELECT * FROM {snowflake_table_path}", use_pandas_result=True, fetch_results=True
         )
         assert out_df["A"].tolist() == ["1", "1", "1"]
 
@@ -323,7 +323,7 @@ def test_time_window_partitioned_asset():
         )
 
         out_df = snowflake_conn.execute_query(
-            f"SELECT * FROM {snowflake_table_path}", use_pandas_result=True
+            f"SELECT * FROM {snowflake_table_path}", use_pandas_result=True, fetch_results=True
         )
         assert sorted(out_df["A"].tolist()) == ["1", "1", "1", "2", "2", "2"]
 
@@ -335,7 +335,7 @@ def test_time_window_partitioned_asset():
         )
 
         out_df = snowflake_conn.execute_query(
-            f"SELECT * FROM {snowflake_table_path}", use_pandas_result=True
+            f"SELECT * FROM {snowflake_table_path}", use_pandas_result=True, fetch_results=True
         )
         assert sorted(out_df["A"].tolist()) == ["2", "2", "2", "3", "3", "3"]
 
@@ -398,7 +398,7 @@ def test_static_partitioned_asset():
         )
 
         out_df = snowflake_conn.execute_query(
-            f"SELECT * FROM {snowflake_table_path}", use_pandas_result=True
+            f"SELECT * FROM {snowflake_table_path}", use_pandas_result=True, fetch_results=True
         )
         assert out_df["A"].tolist() == ["1", "1", "1"]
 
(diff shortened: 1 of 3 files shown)


jamiedemaria:
* master
  * **PR #12633** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12633" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12633?utm_source=stack-comment).

sryza:
> False | True | pandas result

If fetch_result is False, seems like we shouldn't return a result? And probably specifying it as False and use_pandas_result as True should be an error?

jamiedemaria:
> If fetch_result is False, seems like we shouldn't return a result? And probably specifying it as False and use_pandas_result as True should be an error?

seems reasonable to me. that's the current behavior of execute_query, which i agree is confusing

jamiedemaria:
@sryza @benpankow  could i get another pass on this? if possible, it'd be great to get this in with the 1.2 release since it's a breaking change. otherwise we can wait until 1.3

benpankow (approved):
lgtm 👍

PR #12635, opened 2023-03-01
[dagster-gcp-pandas] Use bigquery methods instead of pandas_gbq

### LINKED ISSUE #12542: [dagster-bigquery-pandas] investigate usage of load_table_from_dataframe and to_dataframe instead of pandas_gbq
### What's the use case?

bigquery provides an out of box `load_table_from_dataframe`  and `to_dataframe` methods that we could use instead of pandas_gbq. Would need to investigate that these methods will be drop-in replacements. For example, based on api docs it is unclear is load_table_from_dataframe will append or overwrite an existing table 

https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.client.Client#google_cloud_bigquery_client_Client_load_table_from_dataframe

https://cloud.google.com/bigquery/docs/samples/bigquery-load-table-dataframe

https://cloud.google.com/bigquery/docs/samples/bigquery-list-rows-dataframe

this would also allow custom  LoadJobConfig

https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.job.LoadJobConfig

### Ideas of implementation

_No response_

### Additional information

_No response_

jamiedemaria:
### Summary & Motivation
resolves #12542 

Bigquery has built in methods that allow for working with pandas dataframes. these are ~mysteriously hidden~ not hidden, just not immediately obvious, in their documentation, but a user pointed them out to me! Switching to using these methods will allow us to configure things like retries and timeouts, and maybe allow the user to set their own job config (although i'm not sure how that would work yet). I didn't need to update the test suite to make this change so i think it's pretty safe to say this would be a safe/non-breaking change

### DIFF
--- a/pyright/master/requirements.txt
+++ b/pyright/master/requirements.txt
@@ -51,7 +51,7 @@
 -e python_modules/libraries/dagster-duckdb-polars/
 -e python_modules/libraries/dagster-fivetran/
 -e python_modules/libraries/dagster-gcp/
--e python_modules/libraries/dagster-gcp-pandas/
+-e python_modules/libraries/dagster-gcp-pandas[test]/
 -e python_modules/libraries/dagster-gcp-pyspark/
 -e python_modules/libraries/dagster-ge/
 -e python_modules/libraries/dagster-github/
--- a/pyright/master/requirements.txt
+++ b/pyright/master/requirements.txt
@@ -51,7 +51,7 @@
 -e python_modules/libraries/dagster-duckdb-polars/
 -e python_modules/libraries/dagster-fivetran/
 -e python_modules/libraries/dagster-gcp/
--e python_modules/libraries/dagster-gcp-pandas[test]/
+-e python_modules/libraries/dagster-gcp-pandas[test]
 -e python_modules/libraries/dagster-gcp-pyspark/
 -e python_modules/libraries/dagster-ge/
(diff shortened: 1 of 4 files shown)


jamiedemaria:
* master
  * **PR #12635** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12635" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12635?utm_source=stack-comment).

benpankow (approved):
> these are mysteriously hidden in their documentation

is there any concern that these are non-public or might have their signatures changed? change looks good to me otherwise, nice to eliminate a dep

jamiedemaria:
I don't think so - the functions are documented, they just were hard to find when i was doing the proof of concept last month. after a tiny bit more digging i have a new theory, which is that these methods might be relatively new. There's now a migration guide page telling you how to move away from `pandas_gbq` that i don't remember seeing a couple weeks ago. 

For posterity - here's the docs for [load_table_from_dataframe](https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.client.Client#google_cloud_bigquery_client_Client_load_table_from_dataframe)

PR #12641, opened 2023-03-02
Return dynamic partitions in order

clairelin135:
Changes the `get_dynamic_partitions` method to return keys in the order that they exist in the database. Previously the index would cause the keys to be returned alphabetically.


sryza (approved):
Nice

PR #12645, opened 2023-03-02
Refactor flatten_time_window_ranges

johannkm:
The current implementation fails at least a couple edge cases. 

This replaces it with a simplified version that knows about materialized and failed statuses, rather than trying to be generic. We can revisit this when we also factor in stale etc.

Also added more tests


johannkm:
* master
  * **PR #12645** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12645" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12645?utm_source=stack-comment).

clairelin135 (approved):
thanks for making these updates

PR #12646, opened 2023-03-02
add metadata support to graph_asset

askvinni:
### Summary & Motivation
Noticed that the new decorators don't support passing metadata to the underlying assets. Inspired by #12630 
Wrote tests, spun up local dagster instance to ensure that metadata is correctly displayed in dagit.


sryza:
Thanks for the contribution @askvinni ! Mind running `make black` on this to format it?

askvinni:
@sryza oddly, black didn't detect anything when I ran it yesterday or today. I _think_ the problem was on line 674 that was 103 characters long, looks like it's skipping docstrings in my system.

Also realized I could have tested the case within the test `test_graph_asset_with_args` so I deleted the new test I created and just added the argument and assertion to the existing one.

sryza (approved):
LGTM!

PR #12650, opened 2023-03-02
[ui] Disable repo location reload based on permissions

hellendag:
### Summary & Motivation

Clean up permissioning on code location reload. Pass a `hasReloadPermission` to the rendered `children` prop of `ReloadRepositoryLocationButton`, and use that value to show tooltips and disabled states accordingly.

Screenshot 2023-03-02 at 10 16 03 AM
Screenshot 2023-03-02 at 10 16 11 AM
Screenshot 2023-03-02 at 10 16 23 AM

Force reload permission to be disabled, navigate to code location reload buttons throughout the app, verify disabled state and tooltip display.

### DIFF
--- a/js_modules/dagit/packages/core/src/nav/ReloadRepositoryLocationButton.tsx
+++ b/js_modules/dagit/packages/core/src/nav/ReloadRepositoryLocationButton.tsx
@@ -1,25 +1,33 @@
 import * as React from 'react';
 
 import {AppContext} from '../app/AppContext';
+import {usePermissionsForLocation} from '../app/Permissions';
 import {RepositoryLocationErrorDialog} from '../workspace/RepositoryLocationErrorDialog';
 
 import {buildReloadFnForLocation, useRepositoryLocationReload} from './useRepositoryLocationReload';
 
-type ChildProps = {
+export type ChildProps = {
+  codeLocation: string;
   tryReload: () => void;
   reloading: boolean;
+  hasReloadPermission: boolean;
 };
 
 interface Props {
-  children: (childProps: ChildProps) => React.ReactNode;
+  ChildComponent: React.FC<ChildProps>;
   location: string;
 }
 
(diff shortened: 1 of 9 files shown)


hellendag:
* master
  * **PR #12650** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12650" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12650?utm_source=stack-comment).

js_modules/dagit/packages/core/src/nav/ReloadRepositoryLocationButton.tsx
```diff
@@ -31,7 +38,7 @@ export const ReloadRepositoryLocationButton: React.FC<Props> = (props) => {
 
   return (
     <>
-      {children({tryReload, reloading})}
+      {children({tryReload, reloading, hasReloadPermission})}
```
salazarm, line 41: If you're expecting exactly 1 child with this interface then why not mark `Children` as it's own Component?

eg:
```
interface Props {
  ChildComponent: React.FC<ChildProps>;
  location: string;
}
```

then here you can do:
`<ChildComponent {...{tryReload, reloading, hasReloadPermission}}>`

This would allow the children to be a true component which would let them use React hooks

salazarm: I guess this pattern makes it a little bit easier to use the parent state so you don't need to pass that down.

salazarm (approved):
Should we add a test case for this?

js_modules/dagit/packages/core/src/nav/ReloadRepositoryLocationButton.tsx
hellendag: Actually I don't think I need any parent state at all for these. I'll give this a try.

PR #12658, opened 2023-03-02
Dynamic partioned asset creation feedback

salazarm:
### Summary & Motivation

- Automatically select partition after its added
- Make enter save the partition
- Create Partition -> Add Partition
- "Click and drag to select a range" only for timeseries based partitions

storybook + manually confirmed this worked

image


sryza (approved):
Nice. It would be good to get this in ahead of the dogfooding session in 40 minutes.

PR #12660, opened 2023-03-02
improve execution plan snapshot use in execute_run

alangenfeld:
Fixes dynamic graph execution bug caused in asset usages by not threading `inintial_known_state` through from the execution plan snapshot on the run when we go through the create_run -> execute_run path.

`asset_selection` is always set in asset executions so we were always rebuilding from scratch for assets.

Here we:
* pass along `initial_known_state` when we rebuild
* rebuild from snapshot when subselections match instead of only when they are `None` 

existing coverage. Can add further tests

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/api.py
+++ b/python_modules/dagster/dagster/_core/execution/api.py
@@ -885,29 +885,28 @@ def _check_pipeline(pipeline: Union[PipelineDefinition, IPipeline]) -> IPipeline
 
 
 def _get_execution_plan_from_run(
-    pipeline: IPipeline, dagster_run: DagsterRun, instance: DagsterInstance
+    pipeline: IPipeline,
+    dagster_run: DagsterRun,
+    instance: DagsterInstance,
 ) -> ExecutionPlan:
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster/_core/execution/api.py
```diff
@@ -885,29 +885,28 @@ def _check_pipeline(pipeline: Union[PipelineDefinition, IPipeline]) -> IPipeline
...
+
+    # Rebuild from snapshot if able and selection has not changed
     if (
-        pipeline.solids_to_execute is None
-        and pipeline.asset_selection is None
-        and dagster_run.execution_plan_snapshot_id
+        execution_plan_snapshot is not None
+        and execution_plan_snapshot.can_reconstruct_plan
+        and pipeline.solids_to_execute == dagster_run.solids_to_execute
+        and pipeline.asset_selection == dagster_run.asset_selection
```
alangenfeld, line 903: most recent change was https://github.com/dagster-io/dagster/pull/8994

OwenKephart (approved):
🙏

PR #12670, opened 2023-03-02
Call logging.shutdown at the end of Dagster runs and step processes

gibsondan:
Summary:
This is a shot-in-the-dark attempt to address a user report that I am unable to reproduce myself, where process output from the multiprocess executor is sporadically included in the process output when calling execute_job: https://dagster.slack.com/archives/C01U954MEER/p1677705961810909

logging.shutdown is supposed to be called at the end of a Python application to flush any loggers. This adds it at the end of each of the CLI calls that emit events during a run. This seems at worst harmless and at best is a plausible fix for the above issue.


johannkm (approved):
Add to `resume_run_command` for consistency?

gibsondan:
alex seems to have diagnosed the actual problem here

PR #12672, opened 2023-03-02
Don't double-write log events from the execute_run and execute_step CLI calls

gibsondan:
Summary:
Right now in these API calls we write each event twice;
- one from the dagster logging handler
- one from this structured JSON write at the end

The latter is only needed for celery, so let's leave it for just celery. No reason that I can see to double-write in the execute_run call.
Defaulting the argument to True to backcompat to match old behavior, but setting to to False in the step delegating executor.

Test Plan:
existing e2e tests of executors


johannkm (approved):
lgtm once tests pass

PR #12674, opened 2023-03-02
small updates to automating pipelines guide

sryza:
### Summary & Motivation
- Change URL to end in automating-pipelines instead of automated_pipelines, because we prefer hyphens over dashes and to match the title more closely
- Small tweaks to table at the bottom
  - "Basic and Cron Scheduling" could imply that Basic and Cron are two different options, which they are not
  - "Freshness policy sensors" are actually a different thing than what's being discussed here (they allow users to run arbitrary code when a freshness policy is violated, which is different than materializing assets based on their freshness policies)

### DIFF
--- a/docs/content/_navigation.json
+++ b/docs/content/_navigation.json
@@ -727,7 +727,7 @@
         "children": [
           {
             "title": "Automating your data pipelines",
-            "path": "/guides/dagster/automated_pipelines"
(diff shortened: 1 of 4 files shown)


odette-elementl:
Makes sense - thanks for the edits!

odette-elementl (approved):

PR #12677, opened 2023-03-02
Add AssetIn input_manager_key to docs

bendnorman:
### Summary & Motivation
The `AssetIn` doc string was missing a description of the `input_manager_key` argument.

I wasn't sure how to build the API docs locally.


sryza (commented):
Thanks for the contribution @bendnorman ! Would you mind specifying that, if this isn't set, the IO manager on the upstream asset will be used? This is also still technically experimental functionality, so we should mention that as well.

cmpadden:

Issue #12569, opened 2023-02-27
specifying the description on `@graph_asset` has no effect


sryza:
Reported:
- https://dagster.slack.com/archives/C01U954MEER/p1677531902396549?thread_ts=1677531677.245319&cid=C01U954MEER
- https://github.com/dagster-io/dagster/pull/10152#issuecomment-1446917951

sryza:
Surprisingly difficult fix.

@OwenKephart I'm curious if you have thoughts on what to do here, because I'm guessing you added the `_metadata_by_key` property on `AssetsDefinition`?

I.e. should we maintain duplicate metadata and description on both the `AssetsDefinition` and the op Outs?

simonvanderveldt:
> Surprisingly difficult fix.

Yeah, I look into the changes in #10152 to see if I could make a small PR to fix the missing description but it ended up being quiet a bit more involved than I expected. Also noticed your observation:

> I.e. should we maintain duplicate metadata and description on both the `AssetsDefinition` and the op Outs?

We've had this as a TODO to create an issue/feature request for. We're currently setting the metadata and description on an Op (the last Op in our graph backed Asset) Out, which always seemed incorrect to us.
We expected to be able to set both the description and the metadata on the Asset directly since they are Asset attributes.

OwenKephart:
@sryza I think you're correct that this is similar in nature to the `_metadata_by_key` stuff. In my head, an asset can have a description which is completely independent of ops/graphs. We just pull from the op/output description at the moment because the most common ways of creating an asset (the `@asset` / `@multi_asset` decorators) happens to make it look like you're setting a description on the asset when you're really setting a description on either the output or the op.

The ideal situation in my mind would be for the AssetsDefinition object to have a `_descriptions_by_key` field which can be (optionally) populated directly. If there's no description set for a particular key, then we can try to scoop one off of the underlying op/output definition inside the AssetsDefinition constructor.

This all lines up with the `_metadata_by_key` approach for allowing users to set metadata for graph-backed assets.

sryza:
Here is a fix, that uses that approach: https://github.com/dagster-io/dagster/pull/12680

PR #12680, opened 2023-03-03
Fix setting descriptions with graph_asset and graph_multi_asset decor…

sryza:
…ators

### Summary & Motivation

Fixes https://github.com/dagster-io/dagster/issues/12569


OwenKephart (approved):
🥇

PR #12691, opened 2023-03-03
asset materialize CLI

sryza:
### Summary & Motivation

Usage:
```
dagster asset materialize -f my_assets.py --select asset1

dagster asset materialize -f my_assets.py --select asset1,asset2

dagster asset materialize -f my_assets.py --select asset1*

dagster asset materialize -f my_assets.py --select asset1 --partition 2022-02-01
```

Stuff to think about:
- For jobs, we have `execute` vs. `launch`. Do we want a similar dichotomy here? Probably yes? If so, a couple options:
  - A flag that lets you launch instead of execute
  - A separate `launch-materializations` command


shalabhc:
Very nice! Is it also possible to add an `asset list` command? I find that very useful when inspecting a new project.

The difference between `execute` and `launch` is not very clear to me. However a flag sounds better to me because both options "materialize" the asset. Also in the future if there is another variation of materialization, it would just be another flag rather than another command. For consistency it may make sense to port the flag to `job execute` and deprecate `job launch`.

gibsondan (commented):
How would we explain the difference between this and a hypothetical future command that: materializes an asset over graphql or over the real run launcher? (for jobs there is 'dagster job execute' and 'dagster job launch')

gibsondan:
er I should try actually reading the PR summary, sorry

sryza:
@shalabhc - added in a separate PR: https://github.com/dagster-io/dagster/pull/12790

python_modules/dagster/dagster/_cli/asset.py
```diff
@@ -11,6 +30,57 @@ def asset_cli():
...
+        assets_defs=list(repo_def._assets_defs_by_key.values()),
+        source_assets=list(repo_def.source_assets_by_key.values()),
+        asset_selection=kwargs["select"].split(","),
+    )
+
+    implicit_job_def = repo_def.get_implicit_job_def_for_assets(asset_keys)
+    if implicit_job_def is None:
+        raise DagsterInvalidSubsetError(
+            "All selected assets must share the same PartitionsDefinition or have no"
+            " PartitionsDefinition"
```
gibsondan, line 60: pardon my ignorance, but does this error message map the condition being tested here?

gibsondan (approved):

python_modules/dagster/dagster/_cli/asset.py
sryza: Interestingly, yes.

All the assets get smushed into one implicit job - unless they have different PartitionsDefinitions, in which case they can't be.  So if we can't find an implicit job with all the given assets, it's because they couldn't be put into the same implicit job, because of their conflicting PartitionsDefinitions.

sryza: I'll add a comment

PR #12693, opened 2023-03-03
Fix selecting partition after creating it when "Select all" is not checked

salazarm:
### Summary & Motivation

If you create a new partition from the partition creation dialog then we want to automatically select that partition.

Currently this doesn't happen if you don't have "Select all" checked.

The problem is that we're trying to select the partition before the partition health data has refreshed so our partition key validation filters it out.

This PR adds a prop to allow skipping that validation. 

I added a partition key and saw it was selected despite "Select all" not being checked.


bengotow (approved):
This looks great to me! I think this is definitely the simplest solution a pretty complicated state update 😅

salazarm:
the build failures look unrelated

PR #12695, opened 2023-03-03
[dagster-duckdb] convert to new resource system

jamiedemaria:
### Summary & Motivation
Converts the DuckDB IO Manager and the pandas type handler to the new pythonic resources system

unit tests cover both versions of the io manager

### DIFF
--- a/python_modules/libraries/dagster-duckdb-pandas/dagster_duckdb_pandas/duckdb_pandas_type_handler.py
+++ b/python_modules/libraries/dagster-duckdb-pandas/dagster_duckdb_pandas/duckdb_pandas_type_handler.py
@@ -4,8 +4,8 @@ import pandas as pd
 from dagster import InputContext, MetadataValue, OutputContext, TableColumn, TableSchema
 from dagster._core.storage.db_io_manager import DbTypeHandler, TableSlice
 from dagster_duckdb.io_manager import (
-    ConfigurableDuckDBIOManager,
     DuckDbClient,
+    DuckDBIOManager,
     build_duckdb_io_manager,
 )
 
@@ -136,7 +136,58 @@ Examples:
 """
 
 
-class ConfigurableDuckDBPandasIOManager(ConfigurableDuckDBIOManager):
+class DuckDBPandasIOManager(DuckDBIOManager):
+    """An IO manager definition that reads inputs from and writes Pandas DataFrames to DuckDB. When
+    using the duckdb_pandas_io_manager, any inputs and outputs without type annotations will be loaded
+    as Pandas DataFrames.
+
+    Returns:
+        IOManagerDefinition
+
+    Examples:
+        .. code-block:: python
+
+            from dagster_duckdb_pandas import DuckDBPandasIOManager
+
+            @asset(
+                key_prefix=["my_schema"]  # will be used as the schema in DuckDB
+            )
+            def my_table() -> pd.DataFrame:  # the name of the asset will be the table name
+                ...
+
+            defs = Definitions(
+                assets=[my_table],
+                resources={"io_manager": DuckDBPandasIOManager(database="my_db.duckdb")}
+            )
+
+        If you do not provide a schema, Dagster will determine a schema based on the assets and ops using
+        the IO Manager. For assets, the schema will be determined from the asset key, as in the above example.
+        For ops, the schema can be specified by including a "schema" entry in output metadata. If "schema" is not provided
+        via config or on the asset/op, "public" will be used for the schema.
+
+        .. code-block:: python
+
+            @op(
+                out={"my_table": Out(metadata={"schema": "my_schema"})}
+            )
+            def make_my_table() -> pd.DataFrame:
+                # the returned value will be stored at my_schema.my_table
+                ...
+
+        To only use specific columns of a table as input to a downstream op or asset, add the metadata "columns" to the
+        In or AssetIn.
+
+        .. code-block:: python
+
+            @asset(
+                ins={"my_table": AssetIn("my_table", metadata={"columns": ["a"]})}
+            )
+            def my_table_a(my_table: pd.DataFrame) -> pd.DataFrame:
+                # my_table will just contain the data from column "a"
+                ...
+
+    """
+
     @staticmethod
     def type_handlers() -> Sequence[DbTypeHandler]:
         return [DuckDBPandasTypeHandler()]
--- a/python_modules/libraries/dagster-duckdb-pandas/dagster_duckdb_pandas/duckdb_pandas_type_handler.py
+++ b/python_modules/libraries/dagster-duckdb-pandas/dagster_duckdb_pandas/duckdb_pandas_type_handler.py
@@ -1,7 +1,11 @@
 import pandas as pd
 from dagster import InputContext, MetadataValue, OutputContext, TableColumn, TableSchema
 from dagster._core.storage.db_io_manager import DbTypeHandler, TableSlice
-from dagster_duckdb.io_manager import DuckDbClient, build_duckdb_io_manager
+from dagster_duckdb.io_manager import (
+    DuckDbClient,
+    build_configurable_duckdb_io_manager,
+    build_duckdb_io_manager,
+)
 
 
 class DuckDBPandasTypeHandler(DbTypeHandler[pd.DataFrame]):
@@ -128,3 +132,7 @@ Examples:
             ...
 
 """
+
+configurable_duckdb_pandas_io_manager = build_configurable_duckdb_io_manager(
+    [DuckDBPandasTypeHandler()], default_load_type=pd.DataFrame
+)
--- a/python_modules/libraries/dagster-duckdb-pandas/dagster_duckdb_pandas/duckdb_pandas_type_handler.py
+++ b/python_modules/libraries/dagster-duckdb-pandas/dagster_duckdb_pandas/duckdb_pandas_type_handler.py
@@ -1,9 +1,11 @@
+from typing import Optional, Sequence, Type
+
 import pandas as pd
 from dagster import InputContext, MetadataValue, OutputContext, TableColumn, TableSchema
 from dagster._core.storage.db_io_manager import DbTypeHandler, TableSlice
 from dagster_duckdb.io_manager import (
+    ConfigurableDuckDBIOManager,
     DuckDbClient,
-    build_configurable_duckdb_io_manager,
     build_duckdb_io_manager,
 )
 
@@ -133,6 +135,12 @@ Examples:
 
 """
 
-configurable_duckdb_pandas_io_manager = build_configurable_duckdb_io_manager(
-    [DuckDBPandasTypeHandler()], default_load_type=pd.DataFrame
-)
+
+class ConfigurableDuckDBPandasIOManager(ConfigurableDuckDBIOManager):
+    @staticmethod
+    def type_handlers() -> Sequence[DbTypeHandler]:
+        return [DuckDBPandasTypeHandler()]
+
+    @staticmethod
+    def default_load_type() -> Optional[Type]:
+        return pd.DataFrame
--- a/python_modules/libraries/dagster-duckdb-pandas/dagster_duckdb_pandas/duckdb_pandas_type_handler.py
+++ b/python_modules/libraries/dagster-duckdb-pandas/dagster_duckdb_pandas/duckdb_pandas_type_handler.py
@@ -13,26 +13,29 @@ from dagster_duckdb.io_manager import (
 class DuckDBPandasTypeHandler(DbTypeHandler[pd.DataFrame]):
     """Stores and loads Pandas DataFrames in DuckDB.
 
-    To use this type handler, pass it to ``build_duckdb_io_manager``
+    To use this type handler, return it from the ``type_handlers` method of an I/O manager that inherits from ``DuckDBIOManager``.
 
     Example:
         .. code-block:: python
 
-            from dagster_duckdb import build_duckdb_io_manager
+            from dagster_duckdb import DuckDBIOManager
             from dagster_duckdb_pandas import DuckDBPandasTypeHandler
 
-            @asset
-            def my_table():
+            class MyDuckDBIOManager(DuckDBIOManager):
+                @staticmethod
+                def type_handlers() -> Sequence[DbTypeHandler]:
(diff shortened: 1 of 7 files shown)


jamiedemaria:
* master
  * **PR #12695** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12695" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12695?utm_source=stack-comment).

python_modules/libraries/dagster-duckdb/dagster_duckdb/io_manager.py
```diff
@@ -107,6 +108,26 @@ def duckdb_io_manager(init_context):
     return duckdb_io_manager
 
 
+def build_configurable_duckdb_io_manager(
+    type_handlers: Sequence[DbTypeHandler], default_load_type: Optional[Type] = None
+):
+    class ConfigurableDuckDBIOManager(ConfigurableIOManagerFactory):
+        database: str
```
jamiedemaria, line 115: todo - dedupe defining the config schema

python_modules/libraries/dagster-duckdb/dagster_duckdb/io_manager.py
```diff
@@ -107,6 +110,26 @@ def duckdb_io_manager(init_context):
     return duckdb_io_manager
 
 
+def build_configurable_duckdb_io_manager(
+    type_handlers: Sequence[DbTypeHandler], default_load_type: Optional[Type] = None
+):
```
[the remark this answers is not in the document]
schrockn: I strongly recommend either accepting some copy paste here OR using a more OOP pattern rather than this class factory approach, which I think it both difficult to understand and difficult to type.


Option 1: OOP

Tell the use to write code like this and have a new base class that constructs the DBIOManager on the user's behalf

```
class UserDefinedDuckDbIOManager(CustomizedDuckDbIOManager):
   def type_handlers(self):
       # abstract method in base class
       return [UserDefinedTypeHandler()]
       
   def default_load_type(self):   
      return ...
```

Option 2: Copy paste

Just tell the user to copy and paste code and customize it to their needs.

```
class ExampleConfigurableIOManager(ConfigurableIOManagerFactory):
    database: str

    def create_io_manager(self, context) -> DbIOManager:
            return DbIOManager(
                db_client=DuckDbClient(),
                database=self.database,
                schema=self.schema_,
                type_handlers=... # put your type handlers here,
                default_load_type=# put your default load type here,
                io_manager_name="DuckDBIOManager",
        )
```

This makes DbIOManager a public API we have to support and feels like exposing internals. But at least it is obvious what is going to the typechecker and the user.

python_modules/libraries/dagster-duckdb/dagster_duckdb/io_manager.py
```diff
@@ -107,6 +110,26 @@ def duckdb_io_manager(init_context):
...
+    class ConfigurableDuckDBIOManager(ConfigurableIOManagerFactory):
+        database: str
+        schema_: Optional[str] = None  # schema is a reserved word for pydantic
```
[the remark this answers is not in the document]
jamiedemaria: i tried that and it doesn't seem to work as expected. this is the type-ahead i get 
Screenshot 2023-03-09 at 3 38 54 PM

jamiedemaria:
Updated so the implementations of the io manager are classes too, instead of functions. Still getting used to resources as classes instead of functions so lmk if there's any other places i messed that up!

python_modules/libraries/dagster-duckdb/dagster_duckdb/io_manager.py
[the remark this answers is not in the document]
schrockn: I think we want do when @benpankow said but I am also seeing some unexpected behavior.


For example I have the code:

```py
from dagster import Config, asset, Resource
from dagster._config.structured_config import infer_schema_from_config_class
from pydantic import Field


class ConfigWithSchema(Config):
    schema_: str = Field(
        alias="schema",
        description="A string",
    )


config_object = ConfigWithSchema(schema="foo")
print(config_object.dict())
print(config_object._as_config_dict())  # same behavior as .dict()
print(config_object.dict(by_alias=True))
print(config_object.schema_)
print(infer_schema_from_config_class(ConfigWithSchema).config_type.fields)


@asset
def an_asset(config: ConfigWithSchema):
    print(config.schema_)
```

Running this prints out

```
{'schema_': 'foo'}
{'schema_': 'foo'}
{'schema': 'foo'}
foo
{'schema': Field(<dagster._config.source.StringSourceType object at 0x10e45ace0>, default=@, is_required=True)}
```

Which I *think* should be right. Good question if `_as_config_dict()` should respect aliases or not. Seems like it should since it generates a config schema with `schema`.

The config editor also shows `schema`. 

Screen Shot 2023-03-09 at 4 58 00 PM

I think this is fairly important as users will copy paste. @jamiedemaria it looks like you are seeing different behavior here. Is it different for Resource classes and Config classes?

schrockn (changes requested):
Let's figure out our final answer for fields named "schema" and then document it. This is not the first time this will come up

python_modules/libraries/dagster-duckdb/dagster_duckdb/io_manager.py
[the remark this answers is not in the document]
jamiedemaria: When I set up my io manager like this
```
resources = {
    "io_manager": ConfigurableDuckDBPandasIOManager(database="my_db.duckdb", schema="my_schema")
}
```
and materialize an asset, i also get this error
```
dagster._core.errors.DagsterSubprocessError: During multiprocess execution errors occurred in child processes:
In process 16033: dagster._core.errors.DagsterInvalidConfigError: Incorrect values passed to .configured
Error 1: Received unexpected config entry "schema_" at the root. Expected: "{ database: (String | { env: String }) schema?: (String | { env: String }) }".

Stack Trace:
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_core/executor/child_process_executor.py", line 79, in _execute_command_in_child_process
for step_event in command.execute():
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_core/executor/multiprocess.py", line 93, in execute
yield from execute_plan_iterator(
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_core/execution/api.py", line 1094, in __iter__
yield from self.execution_context_manager.prepare_context()
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_utils/__init__.py", line 536, in generate_setup_events
obj = next(self.generator)
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_core/execution/context_creation_pipeline.py", line 261, in execution_context_event_generator
yield from resources_manager.generate_setup_events()
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_utils/__init__.py", line 536, in generate_setup_events
obj = next(self.generator)
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_core/execution/resources_init.py", line 260, in resource_initialization_event_generator
yield from _core_resource_initialization_event_generator(
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_core/execution/resources_init.py", line 189, in _core_resource_initialization_event_generator
for event in manager.generate_setup_events():
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_utils/__init__.py", line 536, in generate_setup_events
obj = next(self.generator)
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_core/execution/resources_init.py", line 328, in single_resource_event_generator
resource_def.resource_fn(context)
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_config/structured_config/__init__.py", line 340, in setup_context_resources_and_call
return self._resource.resource_fn(context_with_key_mapping)
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_config/structured_config/__init__.py", line 503, in initialize_and_run
with_nested_resources = self._resolve_and_update_nested_resources(context)
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_config/structured_config/__init__.py", line 500, in _resolve_and_update_nested_resources
return self._with_updated_values(to_update)
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_config/structured_config/__init__.py", line 447, in _with_updated_values
return self.__class__(**{**self._as_config_dict(), **values})
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_annotations.py", line 121, in __init__
undecorated_init(self, *args, **kwargs)
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_config/structured_config/__init__.py", line 646, in __init__
ConfigurableResource.__init__(self, **data)
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_annotations.py", line 121, in __init__
undecorated_init(self, *args, **kwargs)
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_config/structured_config/__init__.py", line 412, in __init__
curried_schema = _curry_config_schema(schema, resolved_config_dict)
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_config/structured_config/__init__.py", line 210, in _curry_config_schema
return DefinitionConfigSchema(_apply_defaults_to_schema_field(schema_field, data))
File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_config/structured_config/__init__.py", line 158, in _apply_defaults_to_schema_field
raise DagsterInvalidConfigError(


  File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_core/execution/api.py", line 988, in pipeline_execution_iterator
    for event in pipeline_context.executor.execute(pipeline_context, execution_plan):
  File "/Users/jamie/dev/dagster/python_modules/dagster/dagster/_core/executor/multiprocess.py", line 306, in execute
    raise DagsterSubprocessError(
```

python_modules/libraries/dagster-duckdb/dagster_duckdb/io_manager.py
```diff
@@ -106,6 +105,30 @@ def duckdb_io_manager(init_context):
...
+    database: str
+    schema_: Optional[str] = Field(None, alias="schema")  # schema is a reserved word for pydantic
+
+    @staticmethod
+    @abstractmethod
+    def type_handlers() -> Sequence[DbTypeHandler]:
+        ...
+
+    @staticmethod
+    def default_load_type() -> Optional[Type]:
```
benpankow, line 118: This will continue to follow the defaulting logic if the user doesn't override it (pulling from the type handler if there's only one), right?

python_modules/libraries/dagster-duckdb-pandas/dagster_duckdb_pandas/duckdb_pandas_type_handler.py
```diff
@@ -128,3 +134,13 @@ def my_table_a(my_table: pd.DataFrame) -> pd.DataFrame:
             ...
 
 """
+
+
+class ConfigurableDuckDBPandasIOManager(ConfigurableDuckDBIOManager):
```
benpankow, line 139: Just to make sure, this will be the user-implemented pattern for e.g. if I make `MyTypeHandler`:

```python
class MyCustomDuckDBIOManager(ConfigurableDuckDBIOManager):
    @staticmethod
    def type_handlers() -> Sequence[DbTypeHandler]:
        return [MyTypeHandler()]
    
    # not strictly necessary, since it will be pulled from MyTypeHandler(?)
    @staticmethod
        return MyType
```

benpankow (approved):
This seems like a pretty good end state for the user-facing APIs. For most users who are using a built-in TypeHandler, they can just directly construct an IO manager with type hinting etc. The interface for wrapping and applying your own type handlers also seems pretty minimal.

python_modules/libraries/dagster-duckdb/dagster_duckdb/io_manager.py
jamiedemaria: yeah, the default return of None will make the DB IO manager pull from the type handler

python_modules/libraries/dagster-duckdb-pandas/dagster_duckdb_pandas/duckdb_pandas_type_handler.py
jamiedemaria: yep that's it! and yes `default_load_type` will only be required if there are multiple type handlers AND you want a default type

jamiedemaria:
what do you think about renaming the classes to `DuckDBIOManager` and `DuckDBPandasIOManager` (rather than `ConfigurableDuckDBIOManager` and `ConfigurableDuckDBPandasIOManager`)? I stuck the `Configurable` when i first wrote this to help distinguish them during development, but I think keeping the `Configurable` 1. makes it longer to type, 2. make the new resource system seem like a special case of the original system. But removing it could make it easier to confuse this version of the io manager with the older version? What are your thoughts?

benpankow:
> what do you think about renaming the classes to `DuckDBIOManager` and `DuckDBPandasIOManager` (rather than `ConfigurableDuckDBIOManager` and `ConfigurableDuckDBPandasIOManager`)? I stuck the `Configurable` when i first wrote this to help distinguish them during development and because the base classes are `ConfigurableIOManager`, but I think keeping the `Configurable` 1. makes it longer to type, 2.could make it seem like the new resource system is a special case of the original system. But removing it could make it easier to confuse this version of the io manager with the older version? What are your thoughts?

Definitely support removing `Configurable` prefix, makes sense to me.

python_modules/libraries/dagster-duckdb-pandas/dagster_duckdb_pandas/duckdb_pandas_type_handler.py
```diff
@@ -128,3 +134,64 @@ def my_table_a(my_table: pd.DataFrame) -> pd.DataFrame:
             ...
 
 """
+
+
+class DuckDBPandasIOManager(DuckDBIOManager):
+    """An IO manager definition that reads inputs from and writes Pandas DataFrames to DuckDB. When
+    using the duckdb_pandas_io_manager, any inputs and outputs without type annotations will be loaded
```
jamiedemaria, line 141:
```suggestion
    using the DuckDBPandasIOManager, any inputs and outputs without type annotations will be loaded
```

schrockn (approved):
👍🏻

PR #12698, opened 2023-03-03
[dagit] Support displaying a “failed” partition status

bengotow:
### Summary & Motivation

This PR adds support for @johannkm's new failed states to Dagit. This is part 1 of 2, with updated designs for asset nodes still to come.

Here's a loom video of the changes, demoed on a variety of assets:
https://www.loom.com/share/cfd03a2cbf6d4846a4d8d6a332b6b5c4?from_recorder=1&focus_title=1
(There is a bug in the video where it shows "Failed(0)" in one spot, fixed and updated in the screenshot below!)

There are two awkward bits that are worthy of follow-up:

- We don't yet have a "partially failed, partially materialized" or "partially failed, partially missing" state for the bars, which will be helpful for multi-dimensional assets. It introduces some fan out the way it's currently built, because we store the state in an enum and it now needs combinatorial values. I think this requires a larger refactor because when we add "STALE" we'll have even more possible combinations.

- The "Only missing" feature in the materialize dialog doesn't play super nicely with this because we actually don't know which of the failed are missing + failed and which of the failed partitions were previously materialized. __One immediate option would be to change this to "Only missing and failed"__

I edited the mock data by hand to embed some interesting failure states and then updated the tests to match.

I also updated the mock data used by our Storybooks for asset health so they render with some red bars:

image

image

image

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
@@ -194,8 +194,9 @@ export function buildAssetNodeStatusRow({
   }
 
   if (liveData.partitionStats) {
-    const {numPartitions, numMaterialized} = liveData.partitionStats;
-    const numMissing = numPartitions - numMaterialized;
+    const {numPartitions, numMaterialized, numFailed} = liveData.partitionStats;
+    const numMissing = numPartitions - numMaterialized - numFailed;
+
     return {
       background: late ? Colors.Red50 : numMissing ? Colors.Yellow50 : Colors.Green50,
       border: late ? Colors.Red500 : numMissing ? Colors.Yellow500 : Colors.Green500,
@@ -212,7 +213,7 @@ export function buildAssetNodeStatusRow({
             >
               {late
                 ? humanizedLateString(liveData.freshnessInfo.currentMinutesLate)
-                : `${numMissing.toLocaleString()} missing`}
+                : `${numMissing.toLocaleString()} missing / ${numFailed.toLocaleString()} failed`}
             </Link>
           </Caption>
         </>
@@ -366,6 +367,7 @@ export const ASSET_NODE_LIVE_FRAGMENT = gql`
     partitionStats {
       numMaterialized
       numPartitions
+      numFailed
     }
   }
 
--- a/js_modules/dagit/packages/core/src/assets/usePartitionHealthData.tsx
+++ b/js_modules/dagit/packages/core/src/assets/usePartitionHealthData.tsx
@@ -3,9 +3,11 @@ import isEqual from 'lodash/isEqual';
 import React from 'react';
 
 import {assertUnreachable} from '../app/Util';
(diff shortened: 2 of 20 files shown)


johannkm:
> The "Only missing" feature in the materialize dialog doesn't play super nicely with this because we actually don't know which of the failed are missing + failed and which of the failed partitions were previously materialized. One immediate option would be to change this to "Only missing and failed"

Good callout. We do know this on the backend if we wanted to filter to only truly missing, but I think `Only missing and failed` is probably what people more often want?

js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
```diff
@@ -194,8 +194,9 @@ export function buildAssetNodeStatusRow({
   }
 
   if (liveData.partitionStats) {
-    const {numPartitions, numMaterialized} = liveData.partitionStats;
-    const numMissing = numPartitions - numMaterialized;
+    const {numPartitions, numMaterialized, numFailed} = liveData.partitionStats;
+    const numMissing = numPartitions - numMaterialized - numFailed;
```
salazarm, line 198: I think I've heard there's a "Failed Materialized" state? Or was that "Success, not materialized" ?

js_modules/dagit/packages/core/src/assets/usePartitionHealthData.tsx
```diff
@@ -286,16 +295,21 @@ export function keyCountInSelection(selections: PartitionDimensionSelectionRange
...
     return {
       [PartitionState.MISSING]: 0,
       [PartitionState.SUCCESS]: 0,
+      [PartitionState.FAILURE]: 0,
     };
   }
 
+  // Make sure that the provided selections are in the same order as the /underlying/
+  // range data, which may be reversed if the time series is the second axis.
+  const selections = assetHealth?.isRangeDataInverted ? [..._selections].reverse() : _selections;
```
salazarm, line 311: we should probably memoize this ?

salazarm (approved):

js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
bengotow: It's a bit interesting -- a single partition key can be materialized, and then a backfill of it can fail, so the same key can be in both states. We decided to implement these numbers on the backend so that each key can only be in one bucket though, and failed trumps materailized. 

I think it might be worth a call-out in this code though 👍

js_modules/dagit/packages/core/src/assets/usePartitionHealthData.tsx
bengotow: Ahh this is actually one of those hooks you were talking about where the hook returns an object with methods for interacting with the data. In this case selections is actually a param passed to a method it provides, so we can't memoize it here :-( I think it's only a 2 element array though so not too bad.

Issue #11845, opened 2023-01-23
DBIOManager doesn't work with self-dependent assets


mjclarke94:
### Dagster version

dagster, version 1.1.13

### What's the issue?

When a partitioned asset is first materialised using a DBIOManager, a corresponding table is created in the target database. 

This behaviour breaks down when the asset depends on a past partition of itself. I think the IOManager isn't respecting the `start_time` parameter of the `partition_def`, and so instead of `None` being passed in as an input to the first instance of the asset (as there are no prior partitions), the IOManager tries and fails to access a table which does not yet exist.

### What did you expect to happen?

DBIOManagers should respect the `start_date` parameter of a `TimeWindowPartitionDefinition`, allowing the example given [here](https://gist.github.com/slopp/c4eee86c55c5d5635ccba196672e5ea2) to work with a DBIOManager set.

### How to reproduce?

```
from dagster import (
    asset,
    AssetIn,
    TimeWindowPartitionMapping,
    DailyPartitionsDefinition,
    Definitions,
)
import pandas as pd
from pendulum import datetime
from dagster_duckdb import build_duckdb_io_manager
from dagster_duckdb_pandas import DuckDBPandasTypeHandler

# Create IOManager
duckdb_io_manager = build_duckdb_io_manager([DuckDBPandasTypeHandler()])
resources = {"io_manager": duckdb_io_manager.configured({"database": "test.duckdb"})}

# Create partitions
daily_partitions = DailyPartitionsDefinition(start_date=datetime(2023, 1, 1))


@asset(
    partitions_def=daily_partitions,
    io_manager_key="io_manager",
    ins={
        "broken_asset": AssetIn(
            partition_mapping=TimeWindowPartitionMapping(
                start_offset=-1, end_offset=-1
            ),
        ),
    },
    metadata={
        "partition_expr": "time",
    },
)
def broken_asset(context, broken_asset: pd.DataFrame) -> pd.DataFrame:
    context.log.info("Made it to here.")
    start, end = context.output_asset_partitions_time_window()
    return pd.DataFrame({"time": [start, end], "data": [0.1, 0.2]})


@asset(
    partitions_def=daily_partitions,
    io_manager_key="io_manager",
    metadata={
        "partition_expr": "time",
    },
)
def working_asset(context) -> pd.DataFrame:
    start, end = context.output_asset_partitions_time_window()
    return pd.DataFrame({"time": [start, end], "data": [0.1, 0.2]})


[...]

PR #12700, opened 2023-03-03
[db io managers] support self dependent assets in db io managers

jamiedemaria:
### Summary & Motivation

This PR adds support for self-dependent assets. For a self-dependent asset, the "base case" partition doesn't depend on previous partitions, so the `InputContext.asset_partition_keys` will be an empty list. We check to see if this is the case and if so we return an empty DataFrame. Otherwise we perform the SELECT query as normal

----
Original comment:

Starting to think about how we'll manage self-dependent assets (see #11845). This is one option where we attempt to select from the table, catch the exception if the table doesn't exist, and then confirm that the asset depends on itself. We then make the assumption that that means it's the "base case" of the self dependent asset and return an empty dataframe. I'm not 100% sure what the expectations are around self-dependent asset behavior, so would like some feedback!

I implemented this for the duckdb io manager and pandas type handler as an example. if this approach makes sense i'll implement for the remaining io managers/type handlers

One issue with this approach: imagine we have a self-dependent asset with daily partition starting on `2023-01-01`
where each partition is dependent on the one before it. partition `2023-01-01` is the "base case". if we materialize partition `2023-01-15` first, the io manager will return an empty dataframe (since `2023-01-14` hasn't been materialized yet), which i think (?) would be unexpected behavior. 

If in the case above, returning an empty dataframe is expected, then this approach would work. However, if we want to only return the empty dataframe for the first partition in the partition definition i'll need to do some more work to make that determination in the io manager. 

I think comparing the partition key of the asset to the start of the partitions definition could work in all cases except for static partitions. For static partitions i'm not sure how we know what partition is considered the "first"

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/db_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/db_io_manager.py
@@ -45,6 +45,7 @@ class TableSlice(NamedTuple):
     database: Optional[str] = None
     columns: Optional[Sequence[str]] = None
     partition_dimensions: Optional[Sequence[TablePartitionDimension]] = None
+    is_self_dependent: bool = False
 
 
 class DbTypeHandler(ABC, Generic[T]):
@@ -223,9 +224,11 @@ class DbIOManager(IOManager):
                                 " column of the database contains data for the"
                                 f" {part.name} partition."
                             )
+
                         partition_dimensions.append(
                             TablePartitionDimension(
-                                partition_expr=cast(str, partition_expr_str), partitions=partitions
+                                partition_expr=cast(str, partition_expr_str),
+                                partitions=partitions,
                             )
                         )
                 elif isinstance(context.asset_partitions_def, TimeWindowPartitionsDefinition):
@@ -258,12 +261,17 @@ class DbIOManager(IOManager):
             else:
                 schema = "public"
 
+        is_self_dependent = False
+        if isinstance(context, InputContext) and context.asset_key == output_context.asset_key:
+            is_self_dependent = True
+
         return TableSlice(
             table=table,
             schema=schema,
             database=self._database,
             partition_dimensions=partition_dimensions,
             columns=(context.metadata or {}).get("columns"),  # type: ignore  # (mypy bug)
+            is_self_dependent=is_self_dependent,
         )
 
     def _check_supported_type(self, obj_type):
--- a/python_modules/dagster/dagster/_core/storage/db_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/db_io_manager.py
@@ -223,11 +223,9 @@ class DbIOManager(IOManager):
                                 " column of the database contains data for the"
                                 f" {part.name} partition."
                             )
-
                         partition_dimensions.append(
                             TablePartitionDimension(
-                                partition_expr=cast(str, partition_expr_str),
-                                partitions=partitions,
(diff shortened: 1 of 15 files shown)


jamiedemaria:
* master
  * **PR #12700** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12700" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12700?utm_source=stack-comment).

python_modules/dagster/dagster/_core/storage/db_io_manager.py
```diff
@@ -45,6 +45,7 @@ class TableSlice(NamedTuple):
     database: Optional[str] = None
     columns: Optional[Sequence[str]] = None
     partition_dimensions: Optional[Sequence[TablePartitionDimension]] = None
+    is_self_dependent: bool = False
```
jamiedemaria, line 48: not super happy with `is_self_dependent` yet. i'll be thinking on it a bit more

python_modules/libraries/dagster-duckdb-pandas/dagster_duckdb_pandas/duckdb_pandas_type_handler.py
```diff
@@ -62,7 +63,16 @@ def load_input(
         self, context: InputContext, table_slice: TableSlice, connection
     ) -> pd.DataFrame:
         """Loads the input as a Pandas DataFrame."""
-        return connection.execute(DuckDbClient.get_select_statement(table_slice)).fetchdf()
+        try:
```
jamiedemaria, line 66: alternative logic here would be 
```
if table_slice.is_self_dependent and is_first_partition:
    return pd.DataFrame()
return connection.execute(...)
```

but i'm not sure how to figure out the `is_first_partition` part yet. I could compare the partition key with the start of the partition definition, but static partitions are a bit more complicated

jamiedemaria:
oh awesome, i didn't know that! that should make this a lot easier - ty!

python_modules/dagster/dagster/_core/storage/db_io_manager.py
```diff
@@ -223,9 +223,11 @@ def _get_table_slice(
                                 " column of the database contains data for the"
                                 f" {part.name} partition."
                             )
+
                         partition_dimensions.append(
                             TablePartitionDimension(
-                                partition_expr=cast(str, partition_expr_str), partitions=partitions
+                                partition_expr=cast(str, partition_expr_str),
```
sryza, line 229: Is this formatting change on purpose?

sryza (commented):
Is there some chance we'd want to return `None` in this case instead of an empty dataframe? I think it's probably worse, but curious about the pros/cons.

jamiedemaria:
If we return None it'll get really messy with how we use typing to determine the TypeHandler to use. If we allow the `load_input` method to return None, we then have to change the asset function signature to look like this 
```
def self_dependent_asset(context, self_dependent_asset: Optional[pd.DataFrame]) -> pd.DataFrame:
```

And when the DB io manager tries to load the input, it can't pick the right type handler because the type of the input is `Optional[pd.DataFrame]` instead of `pd.DataFrame`. We could modify the DB IO manager so it can handle this case, but i don't see a benefit to returning None instead of an empty DataFrame that justifies it. If you think of a reason returning `None` makes more sense let me know though!

sryza (approved):
Nice

PR #12702, opened 2023-03-03
Add failing multipartitions to toys

johannkm:
Add a multipartition asset that fails randomly, and one that fails in defined patterns to exercise different bars in the UI.

Screen Shot 2023-03-03 at 4 09 38 PM


bengotow (approved):
Ahhhh this is great, thank you! I think this will make it a lot easier to get marco + dish spun up with proper asset test cases too 🙌

PR #12704, opened 2023-03-03
Fix some upcoming lint errors in docstrings

charliermarsh:
### Summary & Motivation

Previously, if Ruff saw a docstring like:

```py
def get_gas_price(data: EasyEnergyData, hours: int) -> float | None:
    """Get the gas price for a given hour.

    Args:
        data: The data object.
        hours: The number of hours to add to the current time.

    Returns:
        The gas market price value.
    """
```

It treated it as a NumPy-style docstring, since `Returns` is a valid section header in both NumPy- and Google-style conventions (despite `Args` being a Google-only header).

This matches `pydocstyle`'s own behavior, but I think we can improve it a bit. Specifically, in https://github.com/charliermarsh/ruff/pull/3325, we now use those convention-specific headers to improve the docstring 
convention inference.

I ran `main` over Dagster, and it surfaced a couple issues, since it now (correctly) treats some docstrings that it _used_ to consider NumPy-style as Google-style.

As an aside, you can forgo this convention inference altogether via:

```toml
[tool.ruff.pydocstyle]
convention = "google"
```

This is recommended, but there's at least one NumPy style docstring in `dagster/python_modules/dagster/dagster_tests/general_tests/py3_tests/test_inference.py`, so I omitted it for now in case that's intentional.

`ruff .`


charliermarsh:
If those few NumPy-style docstrings are unintentional, also happy to just convert those and enable the explicit Google-style setting.

sryza:
Thanks @charliermarsh. @smackesey - mind taking a look at this one?

smackesey:
@charliermarsh Very nice-- thanks especially for the pointer on `convention = "google"`, I'll be adding that to our config.

smackesey (approved):

PR #12705, opened 2023-03-03
Validation for partition name

salazarm:
### Summary & Motivation

As titled adding validation for the partition creation dialog.

Screen Shot 2023-03-03 at 4 53 38 PM

jest


js_modules/dagit/packages/core/src/partitions/__tests__/CreatePartitionDialog.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+    });
+  });
+
+  it('Shows error state', async () => {
+    await act(async () => {
+      render(<Test mocks={[]} />);
+    });
+    const partitionInput = screen.getByTestId('partition-input');
+    userEvent.type(partitionInput, 'invalidname\n\r\t');
+    expect(screen.getByTestId('warning-icon')).toBeVisible();
```
bengotow, line 60: Small nit but it'd be nice to verify that the warning icon isn't visible before the typing, and have a second bit here that puts in a valid name and then verifies it's no longer visible  (since there's a useMemo involved in the state change)

bengotow (approved):

PR #12709, opened 2023-03-03
[pythonic resources] Add ability to remap resource key -> op/asset arg name

benpankow:
## Summary

Adds the ability to optionally pass a `resource_key_argument_mapping` to your `@op` or `@asset` decorator. This overrides the mapping of resource key -> resource argument name when passing resources as arguments.

For example:

```python

    a_str: str
    
@op(resource_key_argument_mapping={"foo": "my_resource"})
def an_op(my_resource: MyResource) -> None:
    pass

@job(resource_defs={"foo": MyResource(a_str="foo")})
def my_job() -> None:
    an_op()


```

This came up during yesterday's dogfooding session as useful in cases with many similar resources or instances of the same resource type.

## Test Plan

New unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
@@ -65,6 +65,7 @@ def asset(
     config_schema: Optional[UserConfigSchema] = None,
     required_resource_keys: Optional[Set[str]] = ...,
     resource_defs: Optional[Mapping[str, object]] = ...,
+    resource_key_argument_mapping: Optional[Dict[str, str]] = ...,
     io_manager_def: Optional[object] = ...,
     io_manager_key: Optional[str] = ...,
     compute_kind: Optional[str] = ...,
@@ -94,6 +95,7 @@ def asset(
     config_schema: Optional[UserConfigSchema] = None,
     required_resource_keys: Optional[Set[str]] = None,
     resource_defs: Optional[Mapping[str, object]] = None,
+    resource_key_argument_mapping: Optional[Dict[str, str]] = None,
     io_manager_def: Optional[object] = None,
     io_manager_key: Optional[str] = None,
     compute_kind: Optional[str] = None,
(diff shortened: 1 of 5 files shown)


benpankow:
* master
  * **PR #12709** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12709" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12709?utm_source=stack-comment).

sryza:
The intended usage for this is inside a factory pattern? Do you have an example of what that looks like?

benpankow:
> The intended usage for this is inside a factory pattern? Do you have an example of what that looks like?

Factory pattern or cases where e.g. an input name overlaps with a resource key the user would like to use

I could imagine factory pattern being used like:

```python
class MyDBResource(ConfigurableResource):
  connection_uri: str

  def load(self):
    pass
 
def create_load_from_db_asset(db_resource_name: str) -> AssetsDefinition:
  @asset(resource_key_argument_mapping={db_resource_name: "my_db_resource"})
  def load_from_db_asset(my_db_resource: MyResource):
    return my_db_resource.load()

defs = Definitions(
  assets = [
    create_load_from_db_asset("foo_db"),
    create_load_from_db_asset("bar_db")
  ],
  resources={
    "foo_db": MyDBResource(connection_uri=...),
    "bar_db": MyDBResource(connection_uri=...)
  }
)

```

sryza (commented):
I'm curious to hear @schrockn's take on this (even though I know it will take a couple weeks). Part of me feels like this is pretty lightweight so not a big deal. Part of me feels like they should just use `context.resources` in this case so we can avoid adding a new argument and layer of indirection.

schrockn (changes requested):
This is an old PR (clearing out my q) but this seems useful but also I want to avoid adding a top-level argument. 

Can we add this to `ResourceParam` or by some other mechanism to avoid adding a top-level argument?

Issue #12447, opened 2023-02-21
Dagster compatbility with older versions of mysql for storing runs


jcauserjt:
### Dagster version

1.1.19

### What's the issue?

I am trying to migrate my storage to mysql. One issue I have is that my mysql instance requires a primary key to be defined at the time of creating every table but the sql statement it's running to create table `daemon_heartbeats` does not have a primary key (despite it having a unique column. Is there any way to tweak the storage config in `dagster.yaml` to resolve this?

The issue is that a primary key is not specified in some of the metadata tables. Mainly:
`daemon_heartbeats`  and `instance_info` tables in `dagster._core.storage.runs.schema`
The recommended settings for a mysql implementation is for every table to have a primary key. Even if it is an auto increment identity column. I think this would be an incremental benefit to the library.

### What did you expect to happen?

All metadata tables to have a primary key so that initializing tables in older versions of mysql works.

### How to reproduce?

Need mysql version 8.0.21 or earlier

Afaik, this is fixed in 8.0.30 because it auto-creates missing primary keys

### Deployment type

None

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

prha:
@jcauserjt Can you provide any more info on this?  I was just able to run our run storage test suite on a mysql server running `8.0.21`.

prha:
This should be resolved with the `1.2.0` release.

PR #12711, opened 2023-03-04
add primary keys to all run tables

prha:
### Summary & Motivation
We have a handful of tables that don't have primary keys.  This is causing issues with some MySQL instances that enforce that tables have primary keys.  It would also most likely cause replication issues with postgres instances.

Initially reported here: https://github.com/dagster-io/dagster/issues/12447

This PR will set up the metadata such that new tables will be set up correctly, and tries a dialect-fragmented approach for the migration.

For sqlite, we can force the batch-recreate mode, which will under the hood create a new table with the correct schema and copy all the data from the old table to the new table.

For mysql, we split the behavior between instances configured to automatically create an invisible primary key (See: https://dev.mysql.com/doc/refman/8.0/en/create-table-gipks.html) and those that haven't.  This is done by manually running SQL migration commands to rename the generated primary key in those situations.

For postgres / others, we create the primary key using standard alembic commands.

BK

This includes changing our pinned MySQL test instance to use a flag that requires primary keys for all tables.  This should guard against new tables being created without primary keys.  We have to split out our MySQL test instances because these instances are incompatible with our current suite of backcompat tests (which all obviously have some tables without primary keys).

In addition to these tests, I also manually tested our backcompat tests against a MySQL instance with the `--sql_generate_invisible_primary_key` flag turned on, to ensure that the primary keys were created correctly.


prha:
Matching internal PR is here: https://github.com/dagster-io/internal/pull/5081

python_modules/libraries/dagster-mysql/dagster_mysql_tests/docker-compose.yml
```diff
@@ -15,10 +15,10 @@ services:
     networks:
       - mysql
 
-  test-mysql-db-backcompat:
-    image: mysql:8.0.30
-    command: mysqld --default-authentication-plugin=mysql_native_password
-    container_name: test-mysql-db-backcompat
+  test-mysql-db-pinned:
+    image: mysql:8.0.21
+    command: mysqld --default-authentication-plugin=mysql_native_password --sql-require-primary-key=ON
```
alangenfeld, line 20: comments explaining these  - especially the backcompat doesnt work with flag bit

alangenfeld (approved):

PR #12716, opened 2023-03-04
[refactor] ConfigurableClass APIs

smackesey:
### Summary & Motivation

Update `ConfigurableClass` and related APIs for better typing.

- `ConfigurableClassData.rehydrate` now takes as an `as_type` kwarg which enforces return of a typed value (matching API of `serdes.deserialize_value`. Callsites accordingly (mostly in the `InstanceRef`) updated.
- `ConfigurableClass.from_config_value` and `config_type` are now class methods instead of static methods. This makes sense since the return type of `from_config_value` depends on the class the method is attached to (it should return an instance of the class). Implementations of from_config_value accordingly updated.
- Add `ConfigurableClass` as a superclass of abstract component classes, remove from subclasses.
- Improve some related docstrings for clarity.
- Add some type annotations

Internal Companion PR: https://github.com/dagster-io/internal/pull/5067

Existing test suite

### DIFF
--- a/python_modules/dagster/dagster/_core/instance/ref.py
+++ b/python_modules/dagster/dagster/_core/instance/ref.py
@@ -1,5 +1,5 @@
 import os
-from typing import Any, Mapping, NamedTuple, Optional, Sequence
+from typing import TYPE_CHECKING, Any, Mapping, NamedTuple, Optional, Sequence
 
 import yaml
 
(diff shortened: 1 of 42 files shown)


python_modules/dagster/dagster/_core/instance/ref.py
```diff
@@ -455,54 +467,93 @@ def value_for_ref_item(k, v):
...
         return (
-            self.secrets_loader_data.rehydrate()
+            self.secrets_loader_data.rehydrate(as_type=SecretsLoader)
             if self.secrets_loader_data
             else ConfigurableClassData(
                 "dagster._core.secrets.env_file",
                 "EnvFileLoader",
                 yaml.dump({}),
-            ).rehydrate()
+            ).rehydrate(as_type=EnvFileLoader)
```
alangenfeld, line 556: Some vague fear over these runtime checks causing problems in the future for weird back compat cases since we pass the instance-ref to potentially old user code servers. Hard to imagine how exactly it would be a problem though since the target must exist in the old code and should be a child of the correct class.

alangenfeld (approved):

python_modules/dagster/dagster/_core/instance/ref.py
smackesey: >Hard to imagine how exactly it would be a problem though since the target must exist in the old code

Yeah because of this not sure what would go wrong, but possibly not thinking everything through

PR #12722, opened 2023-03-05
feat(dbt): add support for `--debug`

rexledesma:
### Summary & Motivation
As the title. 

In the future, we should support reading this from environment variables/yaml config (as documented in https://docs.getdbt.com/reference/profiles.yml). Passing these as explicit command line flags can get unwieldy, and makes us subject to needing to support new configuration every time a new command line flag is added. We should excise ourselves from this dependency by natively supporting dbt's existing precedence hierachy.

pytest


rexledesma:
* master
  * **PR #12722** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12722" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12722?utm_source=stack-comment).

OwenKephart (approved):
looks good, and yeah I agree that replicating all these arguments is a pain

PR #12723, opened 2023-03-05
Add Python 3.11 to dagster buildkite

gibsondan:
This depends on https://github.com/dagster-io/dagster/pull/12917 which makes it possible to build dagster on python 3.11 by relaxing our grpcio pin for that version only. To keep things green despite hangs and segfaults I kept the specific suites where the hangs and crashes usually happen from running on python 3.11 - otherwise this is mostly just building a new buildkite image and blocking out any packages or examples that have upstream dependencies that aren't 3.11 ready yet.


.buildkite/dagster-buildkite/dagster_buildkite/package_spec.py
```diff
@@ -60,6 +60,7 @@ def _infer_package_type(directory: str) -> str:
 
 PytestExtraCommandsFunction = Callable[[AvailablePythonVersion, Optional[str]], List[str]]
 PytestDependenciesFunction = Callable[[AvailablePythonVersion, Optional[str]], List[str]]
+UnsupportedVersionsFunction = Callable[[Optional[str]], List[AvailablePythonVersion]]
```
smackesey, line 63: Should tag this (and the ones above with `TypeAlias`):

```
UnsupportedVersionFunction: TypeAlias = ...
```

smackesey (approved):
Very nice

alangenfeld:
should sync package metadata with whats under test for each package 
https://github.com/dagster-io/dagster/blame/master/python_modules/dagster/setup.py#L54-L58

gibsondan:
will tackle the setup.py changes next

PR #12724, opened 2023-03-05
Remove deprecated MetadataEntry constructors, update `entry_data` internal refs

smackesey:
### Summary & Motivation

- Remove static `MetadataEntry` APIs originally deprecated for `0.15.0` that should have been removed on past releases.
- Updates internal references to `MetadataEntry` to use `value` instead of `entry_data` (which is deprecated).

Internal Companion PR: https://github.com/dagster-io/internal/pull/5066

Existing BK suite w/ new unit test for backcompat.

### DIFF
--- a/examples/assets_pandas_type_metadata/assets_pandas_type_metadata/resources/csv_io_manager.py
+++ b/examples/assets_pandas_type_metadata/assets_pandas_type_metadata/resources/csv_io_manager.py
@@ -9,6 +9,7 @@ from dagster import (
     TableSchemaMetadataValue,
     io_manager,
 )
+from dagster._core.definitions.metadata import MetadataValue
 
 
 class LocalCsvIOManager(MemoizableIOManager):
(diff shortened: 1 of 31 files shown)


smackesey:
* master
  * **PR #12724** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12724" target="_blank">Graphite</a>  👈
    * **PR #12725** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12725" target="_blank">Graphite</a> 
      * **PR #12726** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12726" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12724?utm_source=stack-comment).

sryza:
As much as I'd like to make this change, I think it's too risky to include in the 1.2 release, this close to release.

Removing the static constructors seems surface-level enough to be safe.

smackesey:
>As much as I'd like to make this change, I think it's too risky to include in the 1.2 release, this close to release.

OK, I updated this PR so that it just removes the static constructors and updates our internal references to the deprecated `entry_data` to `value` instead (which was majority of the changes). I broke the _removal_ of `entry_data` and `description` into an upstack PR: https://github.com/dagster-io/dagster/pull/12769

The scope of that change is a lot clearer to see in that separate PR. FWIW we have been emitting warnings for use of `MetadataEntry.entry_data` and `MetadataEntry.description` forever. There will be a few who will break regardless of when we do this because they've been using positional args `MetadataEntry("label", None, value)` instead of kwargs or passing metadata as a dict (but they've been receiving deprecation warnings). IMO we might as well get it over with as it's confusing to leave this non-functional `description` field in the API indefinitely. That said I respect your conservative instinct here.

examples/assets_pandas_type_metadata/assets_pandas_type_metadata/resources/csv_io_manager.py
```diff
@@ -29,26 +30,26 @@ def handle_output(self, context, obj: pd.DataFrame):
         with open(fpath + ".version", "w", encoding="utf8") as f:
             f.write(context.version if context.version else "None")
 
-        yield MetadataEntry.int(obj.shape[0], "Rows")
-        yield MetadataEntry.path(fpath, "Path")
-        yield MetadataEntry.md(obj.head(5).to_markdown(), "Sample")
-        yield MetadataEntry.text(context.version, "Resolved version")
-        yield MetadataEntry.table_schema(
-            self.get_schema(context.dagster_type),
+        yield MetadataEntry("Rows", value=MetadataValue.int(obj.shape[0]))
```
sryza, line 33: Can we switch this to a dictionary? I think we generally don't want users to be constructing `MetadataEntry`s at all, right?

smackesey: done

sryza (approved):

PR #12725, opened 2023-03-05
MetadataEntryUnion

smackesey:
### Summary & Motivation

Pure typing PR:

- Create a type alias `MetadataEntryUnion` for the common `Union[MetadataEntry, PartitionMetadataEntry]`.
- Make `MetadataValue` and `MetadataEntry` generic.

### How I Tested These Changes

BK

PR #12726, opened 2023-03-05
Remove PartitionMetadataEntry

smackesey:
### Summary & Motivation

Remove `PartitionMetadataEntry`. Consider this an RFC because I don’t have full context here, but here is my understanding from a study of the code:

- `PartitionMetadataEntry` is an experimental class that was added in the past to support creation of metadata entries qualified by a single partition.
- `PartitionMetadataEntry` is not a top-level export— it’s not part of the public API. Unclear whether this is inadvertent.
- Dagit has no concept of a `PartitionMetadataEntry`.
- `PartitionMetadataEntry` is incompatible with our recommended dictionary-based metadata API, since it does not allow the same metadata label (key in the dict) to be used for multiple partititons.
- `PartitionMetadataEntry` is only used in a single test, which is skipped.
- The only place the codebase has `PartitionMetadataEntry`-specific logic is during the generation of `AssetMaterializations` associated with an output. Here is what happens:
    - Metadata entries yielded from the IO manager’s `handle_output` (this runs before output-derived `AssetMaterializations` are generated) and metadata entries passed on the `Output` object are combined.
    - If any of these are `PartitionMetadataEntries`, they are resolved to plain `MetadataEntries` and stored in a dict that maps partitions to `MetadataEntry`.
    - This table is consulted to obtain the `MetadataEntries` that will be placed on each `AssetMaterialization` to be generated from an `Output`.
    - The only way an output generates multiple `AssetMaterializations` is if the run tags specify a `PartitionKeyRange`. In this case, for each asset-associated-output in the run, one `AssetMaterialization` will be generated for each partition in the range.
    - Therefore in theory one could either (a) yield `PartitionMetadataEntry` events from the IO manager while writing an asset-associated output spanning multiple partititons; (b) return an `Output` object in a materialization function with a number of `PartitionMetadataEntry` objects associated to different partitions of the output.

So, IIUC PartitionMetadataEntry was created to solve the problem of providing individual-partitioned-scoped metadata entries for the case whene a single run materializes multiple asset partitions. But:

It’s unclear to me whether materializing multiple partitions in one run is something we even support now or if it’s an old abandoned API. The [Partitions concept page](https://docs.dagster.io/concepts/partitions-schedules-sensors/partitions#partitioned-asset-jobs) doesn’t appear to mention it.

If my understanding above is right, I propose removing PartitionMetadataEntry. It is incompatible with our recommended metadata API and complicates the docstrings we present to users. In the event we want to allow users to provide individual-partition-scoped metadata on an Output object, we can allow passing a Mapping[str, MetadataDict] on the output. We could introduce a similar API for yielding partition-scoped metadata entries from handle_output, though it’s unclear if that’s even something we want to support.

BK

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/implementation/events.py
+++ b/python_modules/dagster-graphql/dagster_graphql/implementation/events.py
@@ -22,7 +22,6 @@ from dagster import (
 )
 from dagster._core.definitions.metadata import (
     DagsterRunMetadataValue,
-    MetadataEntryUnion,
 )
 from dagster._core.events import (
     DagsterEventType,
@@ -38,7 +37,7 @@ MAX_INT = 2147483647
 MIN_INT = -2147483648
 
 
-def iterate_metadata_entries(metadata_entries: Sequence[MetadataEntryUnion]) -> Iterator[Any]:
+def iterate_metadata_entries(metadata_entries: Sequence[MetadataEntry]) -> Iterator[Any]:
     from ..schema.metadata import (
         GrapheneAssetMetadataEntry,
         GrapheneBoolMetadataEntry,
--- a/python_modules/dagster/dagster/_core/definitions/decorators/source_asset_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/source_asset_decorator.py
@@ -3,7 +3,7 @@ from typing import Mapping, Optional, Sequence, Union, overload
 from dagster._annotations import experimental
 from dagster._core.definitions.events import AssetKey, CoercibleToAssetKeyPrefix
 from dagster._core.definitions.metadata import (
-    MetadataEntryUnion,
+    MetadataEntry,
     MetadataUserInput,
 )
 from dagster._core.definitions.partition import PartitionsDefinition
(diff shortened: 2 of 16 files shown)


smackesey:
* master
  * **PR #12724** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12724" target="_blank">Graphite</a> 
    * **PR #12725** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12725" target="_blank">Graphite</a> 
      * **PR #12726** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12726" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12726?utm_source=stack-comment).

smackesey:
* master
  * **PR #12724** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12724" target="_blank">Graphite</a> 
    * **PR #12725** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12725" target="_blank">Graphite</a> 
      * **PR #12726** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12726" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12726?utm_source=stack-comment).

OwenKephart (approved):
Your analysis is correct here. If we need to support partition-specific metadata there are likely more elegant ways to re-introduce that feature in the future. For now, this is not a high priority thing to support, so removing it seems good to me.

sryza (changes requested):
Sorry to be a pain here, but I think we should leave this in a little longer. I think it's likely that we'll want to support partition-level metadata soon. E.g. geoHeil recently asked for it.

OwenKephart:
@sryza I'm not convinced the existence of this API helps all that much -- setting `metadata_entries` on an OutputDefinition is a deprecated pattern and diverges significantly from the accepted way of applying metadata to an output. Realistically, we'd want to have some sort of mapping from partition key to metadata dictionary, which this API does not map well to.

sryza (approved):
I am convinced

PR #12731, opened 2023-03-06
Fix dagster-k8s job execution when a job has more than 1 pod

Taadas:
###  Summary & Motivation

Fix dagster-k8s job execution when a job has more than 1 pod (parallelism). There were two scenarios when job parallelism is set more than 1:
- Dagster would count a job as successful prematurely. It happens because by default `wait_for_running_job_to_succeed` method waits for only one pod in the job to succeed. So, even when there is more than 1 pod in the job, if at least 1 pod succeeds, Dagster would count a job as successful and won't wait for other pods to finish.
- Dagster would wait for a job to succeed infinitely. It happens because by default `wait_for_running_job_to_succeed` method waits for one pod in the job to succeed. If by the time check is executed more than 1 in the job succeeded, the loop will execute infinitely.

 ###  How I Tested These Changes

New integration test `test_k8s_job_op_with_paralellism`.


gibsondan:
Thank you! this looks great. Any chance you could rebase it past a merge conflict? (another test that was added, should be quick)

gibsondan:
Thank you! this looks great. Any chance you could rebase it past a merge conflict? (another test that was added, should be quick)

Taadas:
Sure, rebased. Also, I've changed how `num_pods_to_wait_for` is being set, to handle cases when `job_spec_config`exists but `parallelism` isn't provided in it.

gibsondan (approved):

Issue #12420, opened 2023-02-17
on the asset details page, clicking the "Materialize" button when you've selected a partition should default to that partition


sryza:

salazarm:
just found out it doesn't work with multi partition selection. fix inc

PR #12734, opened 2023-03-06
[Asset Details Page] Auto-select partition in materialization dialog.

salazarm:
### Summary & Motivation

Fixes https://github.com/dagster-io/dagster/issues/12420

We rely on the "partition" query string parameter that the asset detail page uses to track the selected partition.

Tested this with a statically partitioned and dynamically partitioned asset. 
1. Visit the asset detail page
2. Select a partition
3. Click Materialize
4. See dialog has automatically selected the partition that was selected on the details page.

### DIFF
--- a/js_modules/dagit/packages/core/src/partitions/DimensionRangeWizard.tsx
+++ b/js_modules/dagit/packages/core/src/partitions/DimensionRangeWizard.tsx
@@ -47,12 +47,10 @@ export const DimensionRangeWizard: React.FC<{
 
   const didSetInitialPartition = React.useRef(false);
   React.useEffect(() => {
-    if (didSetInitialPartition.current) {
+    if (didSetInitialPartition.current || !partitionKeys.length) {
       return;
     }
-    if (partitionKeys.length) {
-      didSetInitialPartition.current = true;
-    }
+    didSetInitialPartition.current = true;
     const query = qs.parse(window.location.search, {ignoreQueryPrefix: true});
     const partition = query.partition as string;
     if (partition && partitionKeys.includes(partition)) {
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/core/src/partitions/DimensionRangeWizard.tsx
```diff
@@ -44,6 +45,22 @@ export const DimensionRangeWizard: React.FC<{
...
+    if (partitionKeys.length) {
+      didSetInitialPartition.current = true;
+    }
+    const query = qs.parse(window.location.search, {ignoreQueryPrefix: true});
+    const partition = query.partition as string;
+    if (partition && partitionKeys.includes(partition)) {
+      setSelected([partition]);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [partitionKeys]);
```
hellendag, line 62: Why is the lint override necessary here? I'm not spotting what's missing.

salazarm: oh yeah not necessary anymore. Originally I was leaving partitionKeys out but I have to add it since it has to loads first.

salazarm: oh, it's `setSelected` which unfortunately isn't memoized by the callers so it would cause an infinite loop if included.

hellendag: Ah, I mistook it for a regular `useState` setter.

That's really too bad. The infinite loop issue isn't resolved by the `didSetInitialPartition` ref?

hellendag (approved):

js_modules/dagit/packages/core/src/partitions/DimensionRangeWizard.tsx
salazarm: I guess it would but we'd still be needlessly entering this useEffect every time the parent component renders

PR #12735, opened 2023-03-06
[Sensor Testing] Add "Test again" button 

salazarm:
### Summary & Motivation

We want to allow users to easily test their sensor again without closing the dialog and having to click the test sensor button again. The "test again" button resets the dialog back to its initial state

jest

### DIFF
--- a/js_modules/dagit/packages/core/src/ticks/tests/SensorDryRunDialog.test.tsx
+++ b/js_modules/dagit/packages/core/src/ticks/tests/SensorDryRunDialog.test.tsx
@@ -73,6 +73,23 @@ describe('SensorDryRunTest', () => {
     });
   });
 
+  it('allows you to test again', async () => {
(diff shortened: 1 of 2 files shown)


js_modules/dagit/packages/core/src/ticks/tests/SensorDryRunDialog.test.tsx
```diff
@@ -73,6 +73,23 @@ describe('SensorDryRunTest', () => {
     });
   });
 
+  it('allows you to test again', async () => {
+    await act(async () => {
+      render(<Test mocks={[SensorDryRunMutationError]} />);
+    });
```
hellendag, line 79: I think since you're using `MockedProvider`, you don't need the render to be async. (At least, I haven't been using it and it hasn't been a problem.)

hellendag (approved):

PR #12744, opened 2023-03-06
Remove DynamicPartitionsDefinitions API methods

clairelin135:
Removes `DynamicPartitionsDefinitions` API methods.

Updates the tests and the docs to instead use calls from the instance (e.g. `instance.add_dynamic_partitions`).


python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1758,12 +1762,19 @@ def add_dynamic_partitions(
...
+        Delete a partition for the specified dynamic partitions definition.
+        If the partition does not exist, exits silently.
+        """
         check.str_param(partitions_def_name, "partitions_def_name")
         check.sequence_param(partition_key, "partition_key", of_type=str)
         self._event_storage.delete_dynamic_partition(partitions_def_name, partition_key)
 
     @traced
     def has_dynamic_partition(self, partitions_def_name: str, partition_key: str) -> bool:
+        """
```
sryza, line 1775: Can we mark these `@public` so they show up in docs?

sryza (approved):
LGTM after addressing the one comment I left

PR #12759, opened 2023-03-06
Make late-binding of resources to jobs opt-in, with future behavior warning

benpankow:
## Summary

Possible implementation of gating late-binding resources (#12430) to jobs behind a job wrapper.

This ensures that behavior doesn't change for existing useres, see https://github.com/dagster-io/internal/discussions/5072 for more details.

With this setup, users need to explicitly opt-in to have resources from `Definitions` bound to their jobs:

```python

defs = Definitions(
    jobs=BindResourcesToJobs([hello_world_job]),
    resources={
        "io_manager": MyCustomIOManager(),
    },
)
```

Users who will hit the breaking change in 1.3 are shown a warning, as well:

```
You have overridden the default `io_manager` on `Definitions`. One or more jobs utilize the default filesystem I/O manager. In the future, these jobs will use the `Definitions`-level override instead. To silence this warning, explicitly set the `io_manager` on the jobs in question:

@job(resource_defs={{io_manager': fs_io_manager}})
def my_job():
    ...

Alternatively, wrap the jobs input to `Definitions` with `BindResourceToJobs`:

defs = Definitions(
    jobs=BindResourcesToJobs([my_job, my_other_job, ...])
)

The following jobs are affected: `my_job`
```

## Test Plan

Unit tests for warning and for opt-in, opt-out behavior.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/definitions_class.py
+++ b/python_modules/dagster/dagster/_core/definitions/definitions_class.py
@@ -131,7 +131,7 @@ def _jobs_which_will_have_io_manager_replaced(
     ]
 
 
-def _attach_resources_to_jobs_and_instigators(
+def _attach_resources_to_jobs_and_instigator_jobs(
     jobs: Optional[Iterable[Union[JobDefinition, UnresolvedAssetJobDefinition]]],
     schedules: Optional[
         Iterable[Union[ScheduleDefinition, UnresolvedPartitionedAssetScheduleDefinition]]
@@ -240,11 +240,12 @@ def _create_repository_using_definitions_args(
     check.opt_mapping_param(loggers, "loggers", key_type=str, value_type=LoggerDefinition)
 
     if isinstance(jobs, BindResourcesToJobs):
+        # Binds top-level resources to jobs and any jobs attached to schedules or sensors
         (
             jobs_with_resources,
             schedules_with_resources,
             sensors_with_resources,
-        ) = _attach_resources_to_jobs_and_instigators(jobs, schedules, sensors, resource_defs)
+        ) = _attach_resources_to_jobs_and_instigator_jobs(jobs, schedules, sensors, resource_defs)
     else:
         jobs_to_warn = _jobs_which_will_have_io_manager_replaced(jobs, resource_defs)
         if jobs_to_warn:
--- a/python_modules/dagster/dagster/_core/definitions/definitions_class.py
+++ b/python_modules/dagster/dagster/_core/definitions/definitions_class.py
@@ -1,7 +1,9 @@
+import warnings
 from typing import (
     TYPE_CHECKING,
     Any,
     Iterable,
+    List,
     Mapping,
     NamedTuple,
     Optional,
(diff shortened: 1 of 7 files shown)


benpankow:
* master
  * **PR #12430** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12430" target="_blank">Graphite</a> 
    * **PR #12759** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12759" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12759?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -220,11 +239,37 @@ def _create_repository_using_definitions_args(
...
+    else:
+        jobs_to_warn = _jobs_which_will_have_io_manager_replaced(jobs, resource_defs)
+        if jobs_to_warn:
+            jobs_text = ", ".join([f"`{job.name}`" for job in jobs_to_warn[:10]])
+            if len(jobs_to_warn) > 10:
+                jobs_text += f", and {len(jobs_to_warn) - 10} others"
+            warnings.warn(
+                f"""You have overridden the default `io_manager` on `Definitions`. One or more jobs utilize the default filesystem I/O manager. In the future, these jobs will use the `Definitions`-level override instead. To silence this warning, explicitly set the `io_manager` on the jobs in question:
+
+@job(resource_defs={{io_manager': fs_io_manager}})
```
yuhan, line 257:
```suggestion
@job(resource_defs={{'io_manager': fs_io_manager}})
```

yuhan (commented):
i like this option

python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -220,11 +239,37 @@ def _create_repository_using_definitions_args(
...
-        jobs_with_resources,
-        schedules_with_resources,
-        sensors_with_resources,
-    ) = _attach_resources_to_jobs_and_instigators(jobs, schedules, sensors, resource_defs)
+    if isinstance(jobs, BindResourcesToJobs):
+        (
+            jobs_with_resources,
+            schedules_with_resources,
+            sensors_with_resources,
+        ) = _attach_resources_to_jobs_and_instigators(jobs, schedules, sensors, resource_defs)
```
schrockn, line 247: so right now if `BindResourcesToJobs` wraps the jobs then it will also attach the resources to sensors and schedules. So in effect `BindResourcesToJobs` acts as a definitions-wide global even though it appears to be scope to jobs only. I think I am reading this right.

schrockn (changes requested):
Per comment this is a Definitions-wide behavior change that has a bit of a misleading API.

python_modules/dagster/dagster/_core/definitions/definitions_class.py
benpankow: Maybe the helper method here could be clarified - what this does is updates all the jobs and the instigators' jobs (if the instigators have `JobDefinitions` directly attached, they also need to be updated). This is separate from binding resources to a schedule or sensor function.

schrockn:
Per in-person discussion, I think the use case we have to think about is user that upgrades directly from 1.1.x to 1.3. They wouldn't see this warning and would immediately run headlong into the problematic behavior.

python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -220,11 +239,38 @@ def _create_repository_using_definitions_args(
...
+@job(resource_defs={{'io_manager': fs_io_manager}})
+def my_job():
+    ...
+    
+Alternatively, wrap the jobs input to `Definitions` with `BindResourceToJobs`:
+
+defs = Definitions(
+    jobs=BindResourcesToJobs([my_job, my_other_job, ...])
+)
+
```
schrockn, line 267: Maybe note that in later releases `BindResourcesToJobs` will not be required

schrockn (approved):
Ok let's do it

PR #12761, opened 2023-03-06
cron_schedule validation for time window partitions

prha:
### Summary & Motivation
Saw some 500s where an invalid cron schedule bubbled up when generating some asset event metadata.

We have to have some `_legacy` validation flag, so that we can decouple the dagster user code version from the dagster host cloud version.

BK

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
@@ -25,9 +25,13 @@ import dagster._check as check
 from dagster._annotations import PublicAttr, public
 from dagster._core.instance import DynamicPartitionsStore
 from dagster._utils.partitions import DEFAULT_HOURLY_FORMAT_WITHOUT_TIMEZONE
-from dagster._utils.schedules import cron_string_iterator, reverse_cron_string_iterator
+from dagster._utils.schedules import (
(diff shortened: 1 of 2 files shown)


prha:
The `TimeWindowPartitionsDefinition` is originally defined in user code, but it gets reconstructed in host code to populate some graphql entries:
https://github.com/dagster-io/dagster/blob/master/python_modules/dagster-graphql/dagster_graphql/schema/partition_sets.py#L396

When we update cloud, we will have the new constructor, which will include the cron string validation, but it could be constructing it from an external data representation of user code definitions using an older version of dagster (without the cron string validation).  We'd then get a 500, which we're hoping to avoid.

The user that hit the 500 has updated their definition to fix the cron string, so we could potentially not include the flag if we don't care about patching this version inconsistency (cc @gibsondan )

sryza:
> The user that hit the 500 has updated their definition to fix the cron string, so we could potentially not include the flag if we don't care about patching this version inconsistency

If this is basically the first time we've hit it, my instinct would be to leave it out, to avoid complicating the code

sryza:
I don't feel super strongly though - feel free to overrule me on that. If we do keep it though, I think it's worth a comment explaining what it's for.

sryza (approved):
looks good!

PR #12767, opened 2023-03-07
Set default memory/CPU for new ECS tasks based on runtime platform

gibsondan:
Summary:
Windows task definitions need to be registered with at least 1024 CPU and 2048 memory, even if the cpu/memory are going to be later overridden by a task override / container override. Instead of hard-coding 256/512 for all tasks, check the runtime platform and change the default CPU/memory accordingly.

Additionally, if run_resources have been explicitly set, send those in as the task definition CPU/memory (This is harmless and is another fix for the problem above - it also probably makes it easier to understand what the actual resource limits are if you're inspecting the task).

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws_tests/ecs_tests/launcher_tests/test_launching.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws_tests/ecs_tests/launcher_tests/test_launching.py
@@ -14,7 +14,12 @@ from dagster._core.launcher import LaunchRunContext
 from dagster._core.launcher.base import WorkerStatus
 from dagster._core.origin import PipelinePythonOrigin, RepositoryPythonOrigin
 from dagster_aws.ecs import EcsEventualConsistencyTimeout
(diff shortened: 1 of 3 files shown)


python_modules/libraries/dagster-aws/dagster_aws_tests/ecs_tests/launcher_tests/test_launching.py
```diff
@@ -469,6 +469,112 @@ def test_reuse_task_definition(instance, ecs):
...
+
+        task = ecs.describe_tasks(tasks=[task_arn])["tasks"][0]
+
+        task_definition_arn = task["taskDefinitionArn"]
+
+        task_definition = ecs.describe_task_definition(taskDefinition=task_definition_arn)[
+            "taskDefinition"
+        ]
+
+        # Since this is not a windows task def, the default cpu/memory are 256/512
```
jmsanders, line 573: This comment is wrong

jmsanders (approved):
Now that we've discovered this behavior, it'd be nice if our stubbed ECS implemented it so we don't accidentally re-introduce it later.

PR #12775, opened 2023-03-07
update some guide titles and descriptions

sryza:
### Summary & Motivation

My motivation for this was to make the difference between /guides/dagster/how-assets-relate-to-ops-and-graphs and /guides/dagster/enriching-with-software-defined-assets more clear.  I made a few other changes as well while there.

### DIFF
--- a/docs/content/_navigation.json
+++ b/docs/content/_navigation.json
@@ -673,7 +673,7 @@
             "path": "/guides/dagster/how-assets-relate-to-ops-and-graphs"
           },
           {
-            "title": "Enriching with Software-defined assets",
+            "title": "Moving to Software-defined assets",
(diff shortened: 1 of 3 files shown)


erinkcochran87 (approved):
Ty!

One other thing I'd check before you merge this is if the **Moving to SDAs** guide is referenced by title anywhere. Best way to look for this is probably to search for its file path (`/guides/dagster/enriching-with-software-defined-assets`) and looking at the display text.

sryza:
> Before you merge this, I'd also check if the Moving to SDAs guide is referenced by title anywhere. Best way to look for this is probably to search for its file path (/guides/dagster/enriching-with-software-defined-assets) and looking at the display text.

Just did a search and didn't come across any

Issue #12751, opened 2023-03-06
after clicking "Add partition" in the materialize dialog, put focus on text box


sryza:
we should minimize the number of clicks that users have to do in order to add a partition

PR #12777, opened 2023-03-07
[Create Partition Dialog] Autofocus input

salazarm:
### Summary & Motivation

Auto focus the input so that users can start typing immediately.

Addresses https://github.com/dagster-io/dagster/issues/12751

Storybook
Screen Shot 2023-03-07 at 2 26 11 PM


hellendag (approved):
Does it not work to use the `autoFocus` prop on `Dialog`? I would have expected that to be enabled by default, but maybe it's not working for some reason...

PR #12782, opened 2023-03-07
Maintain a separate cursor in the asset partition cache pointing back to the earliest in progress run

johannkm:
The current implementation of failed partition tracking looses track of in progress runs, and doesn't update if they fail later. 

Get around this by halting a new cursor when we find an in progress materialization and re-evaluating it in future updates until it finishes.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql_tests/graphql/snapshots/snap_test_assets.py
+++ b/python_modules/dagster-graphql/dagster_graphql_tests/graphql/snapshots/snap_test_assets.py
@@ -1327,6 +1327,994 @@ snapshots['TestAssetAwareEventLog.test_all_asset_keys[sqlite_with_default_run_la
     }
 }
 
+snapshots['TestAssetAwareEventLog.test_asset_op[postgres_with_default_run_launcher_deployed_grpc_env] 1'] = {
+    'assetOrError': {
+        'definition': {
+            'op': {
+                'description': None,
+                'inputDefinitions': [
+                    {
+                        'name': 'asset_one'
+                    }
+                ],
+                'name': 'asset_two',
+                'outputDefinitions': [
+                    {
+                        'name': 'result'
+                    }
+                ]
+            }
+        }
+    }
+}
+
+snapshots['TestAssetAwareEventLog.test_asset_op[postgres_with_default_run_launcher_managed_grpc_env] 1'] = {
+    'assetOrError': {
+        'definition': {
+            'op': {
+                'description': None,
+                'inputDefinitions': [
+                    {
+                        'name': 'asset_one'
+                    }
+                ],
+                'name': 'asset_two',
+                'outputDefinitions': [
+                    {
+                        'name': 'result'
+                    }
+                ]
+            }
+        }
+    }
+}
+
+snapshots['TestAssetAwareEventLog.test_asset_op[sqlite_with_default_run_launcher_deployed_grpc_env] 1'] = {
+    'assetOrError': {
+        'definition': {
+            'op': {
+                'description': None,
+                'inputDefinitions': [
+                    {
+                        'name': 'asset_one'
+                    }
+                ],
+                'name': 'asset_two',
+                'outputDefinitions': [
+                    {
+                        'name': 'result'
+                    }
+                ]
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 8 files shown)


johannkm:
* master
  * **PR #12782** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12782" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12782?utm_source=stack-comment).

clairelin135 (commented):
I think this makes sense to me.

I left one comment in the PR. One other thing is that now it seems like we're writing a partially resolved state in the cache--if a partition has failed after the cursor, we're writing it into the failed partitions subset. Similarly materialized partitions after the cursor will also exist in the cache. I think it will be less confusing and easier to handle if the objects written into the cache reflect purely the state up until the storage ID.

What do you think about only persisting the state in the cache up until the latest storage ID? but returning the object with the up to date contents?

python_modules/dagster/dagster/_core/storage/partition_status_cache.py
```diff
@@ -269,32 +295,53 @@ def _get_updated_failed_partition_subset(
...
+        new_failed_partitions = set()
+        if incomplete_materialization_records:
+            failed_run_ids = {
+                r.run_id
+                for r in instance.get_runs(
+                    filters=RunsFilter(
+                        run_ids=[
+                            record.run_id for record in incomplete_materialization_records.values()
+                        ],
+                        statuses=[DagsterRunStatus.FAILURE],
```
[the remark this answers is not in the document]
johannkm: yes that's an important catch

johannkm:
> What do you think about only persisting the state in the cache up until the latest storage ID? but returning the object with the up to date contents?

That's a great idea, reduces the oddity

python_modules/dagster/dagster/_core/storage/partition_status_cache.py
```diff
@@ -269,32 +297,55 @@ def _get_updated_failed_partition_subset(
...
+    incomplete_materialization_records: Dict[str, EventLogRecord] = {}
     for record in sorted(unevaluated_event_records, key=lambda r: r.storage_id):
         event = check.not_none(record.event_log_entry.dagster_event)
         if event.is_asset_materialization_planned:
             if event.partition:
                 # If we have a planned materilization for a partition, keep track of it to see if we
                 # also find a materialization. If not, we'll check if the run failed and add it to
                 # the failed partitions.
-                incomplete_materialization_runs[event.partition] = record.run_id
+                incomplete_materialization_records[event.partition] = record
```
johannkm, line 312: this implementation only keeps track of 1 in progress materialization at a time- the one that started most recently. I think that's fairly reasonable. If 
- run 1 starts before run 2
- run 1 finishes after run 2
- run 1 fails to materialize and run 2 succeeds

what state should the partition be in? This implementation will say materialized, which I think is reasonable. Especially given that we don't know when the actual step that tried to materialize in run 1 failed, it could have happened immediately before run 2 even started.

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1788,7 +1791,7 @@ def get_latest_asset_partition_materialization_attempts_without_materializations
 
         with self.index_connection() as conn:
             rows = conn.execute(query).fetchall()
-            return {row["partition"]: row["run_id"] for row in rows}
+            return {row["partition"]: (row["run_id"], row["id"]) for row in rows}
```
johannkm, line 1794: will need to update cloud query

python_modules/dagster-graphql/dagster_graphql_tests/graphql/snapshots/snap_test_assets.py
```diff
@@ -1327,6 +1327,994 @@
     }
 }
 
+snapshots['TestAssetAwareEventLog.test_asset_op[postgres_with_default_run_launcher_deployed_grpc_env] 1'] = {
```
johannkm, line 1330: confused about these snapshots

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
johannkm: https://github.com/dagster-io/internal/pull/5102

clairelin135 (approved):

PR #12781, opened 2023-03-07
[pythonic resources][fix] Fix inheriting attributes when extending Pythonic resources

benpankow:
## Summary

Fixes an error where extending a `ConfigurableResource` class would drop any defaults associated with the base class, e.g.

```python

class BaseResource(ConfigurableResource):
    a_str: str = "bar"
    an_int: int = 1

class ExtendingResource(BaseResource):
    a_float: float = 1.0

# errors because `a_str` is unset, since default is discarded
ExtendingResource()

```

This was caused by unnecessarily merging attributes of upstream classes in the metaclass.

## Test Plan

Added unit test to verify that this now works (& added unit test to ensure this pattern works on `Config` too).

PR #12664, opened 2023-03-02
[dagster-airbyte] Migrate to Pythonic config & resources

benpankow:
## Summary

Converts the `dagster-airbyte` internals to use the new config and resource systems:

- Assets and ops now use Pythonic config rather than `config_schema`
- Assets and ops retrieve resource dependencies through params rather than the context / `required_resource_keys`
- `AirbyteResource` is now a `ConfigurableResource`, holding its own config schema. The `airbyte_resource` function resource now wraps this class rather than holding the config.

## Test Plan

Existing unit tests.

PR #12783, opened 2023-03-07
Add quiet_experimental decorator to ignore nested uses of experimental APIs

benpankow:
## Summary

Adds a new `@quiet_experimental` internal API which we can use to utilize experimental classes and methods without triggering user-facing warnings, useful for e.g. the experimental config system.

Usage:

```python
@experimental
def my_experimental_function(my_arg) -> None:
    pass

@quiet_experimental
def my_quiet_wrapper(my_arg) -> None:
    my_experimental_function(my_arg)
```

## Test Plan

Added unit test.


benpankow:
* master
  * **PR #12781** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12781" target="_blank">Graphite</a> 
    * **PR #12664** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12664" target="_blank">Graphite</a> 
      * **PR #12783** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12783" target="_blank">Graphite</a>  👈
        * **PR #12784** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12784" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12783?utm_source=stack-comment).

sryza:
Should basically every Dagster function have this decorator on it? If so, maybe we should modify the experimental decorator to check whether it's being called by a Dagster function?

python_modules/dagster/dagster/_annotations.py
```diff
@@ -136,3 +138,59 @@ def is_experimental(obj: Annotatable, attr: Optional[str] = None) -> bool:
 def _get_target(obj: Annotatable, attr: Optional[str] = None):
     lookup_obj = obj.__dict__[attr] if attr else obj
     return lookup_obj.fget if isinstance(lookup_obj, property) else lookup_obj
+
+
+def quiet_experimental(obj: T_Annotatable) -> T_Annotatable:
+    """
+    Mark a class/method/function as ignoring experimental warnings. This quiets any "experimental" warnings
+    emitted inside the passed callable is called. Useful when we want to use experimental features internally
```
smackesey, line 146: grammar error

smackesey (dismissed):
This seems useful, two comments:

- the name sounds to me like the thing being decorated is "quietly experimental" . Recommend changing the name to `@quiet_experimental_warnings`.
- Also, I'm not sure it makes sense to mark a class with this and then only silence warnings in the `__init__`. I realize you are following the pattern of the existing `@experimental` decorator, but in that case we are marking the target itself as experimental, which is not really analogous to what we're doing here. Why not drop the class case and just apply the function version to a class' `__init__`? IMO the effect would be more self explanatory this way.

benpankow:
👍  updated name and removed the class-wrapping functionality in favor of applying manually to `__init__`. Added a test for this case.

smackesey (approved):
Sweet 🙏

PR #12790, opened 2023-03-08
asset list command

sryza:

### DIFF
--- a/python_modules/dagster/dagster/_cli/asset.py
+++ b/python_modules/dagster/dagster/_cli/asset.py
@@ -83,6 +83,34 @@ def execute_materialize_command(instance: DagsterInstance, kwargs: Mapping[str, 
     )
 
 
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster/_cli/asset.py
```diff
@@ -81,6 +81,34 @@ def execute_materialize_command(instance: DagsterInstance, kwargs: Mapping[str,
     )
 
 
+@asset_cli.command(name="list", help="List assets")
+@python_origin_target_argument
+@click.option("--select", help="Asset selection to target", required=False)
```
OwenKephart, line 86: "Asset selection" is a bit of a fraught term but I think it works here. Should we at least add some examples (e.g. "prefixed/asset*")?

OwenKephart (approved):
nice

PR #12798, opened 2023-03-08
[dagit] Update the asset partitions / events view after run failures

bengotow:
### Summary & Motivation

We use a "hint" value to trigger a refresh of the asset event and asset partition views when the asset's live data has changed (using lightweight data to trigger a refresh of "heavy" requests). Now that we have a partition failure state, we need to refresh the views when a new run(s) have failed.

I also updated the name of the prop that implements this to be more explicit.

I manually tested this change by having a run fail and observing that the partitions view refreshes properly to show the red stripe and failed partition dot.


johannkm (approved):
nice

PR #12799, opened 2023-03-08
fix execution plan step output collisions

### LINKED ISSUE #12073: op result is overwritten when op is run inside graph
### Dagster version

dagster, version 1.1.6

### What's the issue?

Consider the following repro:
```python
import numpy as np
from dagster import job, graph, op, get_dagster_logger


@op
def random():
    return np.random.randint(0, 100)


@op
def log(value):
    get_dagster_logger().warning(f"VALUE: {value}")


@graph
def iteration(value):
    log(value)
    return random()


@job
def my_job():
    starting_value = random()
    iteration(starting_value)
    iteration(starting_value)


if __name__ == "__main__":
    my_job.execute_in_process()
```

When this script is executed, I see the following output:
```
2023-02-03 07:01:09 -0800 - dagster - WARNING - my_job - beb4d1fd-e709-475e-8d87-670ef1f0e031 - iteration.log - VALUE: 20
2023-02-03 07:01:09 -0800 - dagster - WARNING - my_job - beb4d1fd-e709-475e-8d87-670ef1f0e031 - iteration_2.log - VALUE: 82
```
You can see that the value being logged has changed in between different iterations, even though the job above pretty clearly shows that the value should *not* change. It seems like the run of `random` inside of the graph is actually affecting the return value of the run *outside* of the graph, which is not the behavior I would expect.

One pretty interesting thing I found is that if I change my graph to:
```python
@graph
def iteration(value):
    log(value)
    return random.alias("inside_graph")()
```
... then the problem goes away (ie: the "starting_value" won't change due to the ops called inside the graph). This leads me to suspect that dagster is using the op's invocation alias to store the results, and that is causing the initial results to be overridden by the call inside the graph--could that be the case?

I would expect that the ops that run inside of the graph should have no effect on ops that run outside of the grpah. I would not expect my initial value to be magically updated.

### How to reproduce?

See the section above.

### Deployment type

None

### Deployment details

Not relevant for the reproducibility. 

### Additional information

_No response_

alangenfeld:
We we were using a plan scoped data structure instead of a composition layer scoped one for mapping nodes to step outputs which led to collisions if you happen to re-use names at different scopes. 

resolves #12073

added test


alangenfeld:
* master
  * **PR #12799** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12799" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12799?utm_source=stack-comment).

OwenKephart (approved):
took some head scratching but I get it now -- nice 👍

PR #12809, opened 2023-03-08
[freshness-refactor][1/n] Separate part of CachingInstanceQueryer into CachingDataTimeResolver

OwenKephart:
### Summary & Motivation

The CachingInstanceQueryer was getting quite gnarly. Its functionality is useful in many contexts, not just calculating data times. This first PR is basically just a pure cut/paste of all the data-time-related functions into a new CachingDataTimeResolver class.

There are some rough edges on the API where we have to call `data_time_resolver.instance_queryer`, but that sort of stuff will be smoothed out later on in the stack.


sryza (approved):
Very good idea

PR #12812, opened 2023-03-08
[RFC] Enable returning SensorTickResult containing DynamicPartitionsRequests

clairelin135:
This PR introduces a more fault tolerant way of mutating dynamic partitions definitions through sensors, by introducing a `SensorTickResult` object that can be returned from a sensor. This object when returned will contain everything intended to be returned from the sensor (run requests, new cursor, skip reason, dynamic partitions requests, pipeline run reactions). 

Run requests are evaluated using the state of the partitions after dynamic partitions requests are applied, meaning that the following is valid:
```
quux = DynamicPartitionsDefinition(name="quux")

SensorTickResult(
        dynamic_partitions_requests=[
            quux.request_for_adding_partitions(["1"]),
        ],
        run_requests=[quux_asset_job.run_request_for_partition("1")],
    )
```
Within `evaluate_tick`, run requests and dynamic partitions requests are validated, throwing an error for invalid requests. In this situation, no changes are applied--so the partitions def is not mutated and the run request is not created.

```
return SensorTickResult(
        dynamic_partitions_requests=[
            quux.request_for_deleting_partitions(["2"]),
        ],
        run_requests=[quux_asset_job.run_request_for_partition("2")],
    )
```

These dynamic partitions requests are only applied (added/deleted from the instance) within `_evaluate_sensor`, where we do other things like submitting runs. This enables us to have logging for partitions that are added and deleted. In the future, we could also persist the mutations in the tick so that we have a record of changes and can display them in Dagit.

### DIFF
--- a/examples/docs_snippets/docs_snippets/concepts/partitions_schedules_sensors/dynamic_partitioned_asset.py
+++ b/examples/docs_snippets/docs_snippets/concepts/partitions_schedules_sensors/dynamic_partitioned_asset.py
@@ -41,7 +41,7 @@ def image_sensor(context):
     context.instance.add_dynamic_partitions(images_partitions_def.name, new_images)
 
     run_requests = [
-        images_job.run_request_for_partition(img_filename, instance=context.instance)
+        images_job.run_request_for_partition(img_filename)
         for img_filename in new_images
     ]
     return run_requests
--- a/python_modules/dagster/dagster/_core/definitions/job_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/job_definition.py
@@ -71,7 +71,7 @@ from .partition import PartitionedConfig, PartitionsDefinition, PartitionSetDefi
 from .pipeline_definition import PipelineDefinition
 from .preset import PresetDefinition
 from .resource_definition import ResourceDefinition
-from .run_request import RunRequest
+from .run_request import PendingPartitionedRunRequest, RunRequest
 from .utils import DEFAULT_IO_MANAGER_KEY
 from .version_strategy import VersionStrategy
 
@@ -596,7 +596,7 @@ class JobDefinition(PipelineDefinition):
         run_config: Optional[Mapping[str, Any]] = None,
         instance: Optional["DagsterInstance"] = None,
         current_time: Optional[datetime] = None,
-    ) -> RunRequest:
+    ) -> Union[RunRequest, PendingPartitionedRunRequest]:
         """
         Creates a RunRequest object for a run that processes the given partition.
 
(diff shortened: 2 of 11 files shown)


sryza:
Thanks for putting this together @clairelin135. I added @prha @gibsondan @OwenKephart and @schrockn, because there are a bunch of subjective API aesthetics questions here that would be good to get a diversity of opinions on.

A few things to discuss here:

#### Name of the result object.

We have an object called a `SensorEvaluationContext`, so `SensorEvaluationResult` would be the most consistent? But it's a lot to type. Would we go back in time and change `SensorEvaluationContext` to `SensorTickContext` if we could? How would we feel about just `SensorResult`?

#### A method on `DynamicPartitionsDefinition` to produce the `DynamicPartitionsRequest`, vs. instantiating it directly

@schrockn had concerns about `DynamicPartitionsDefinition.add_partitions`. @schrockn - do those concerns extend here?

#### `DynamicPartitionsRequest` with a `DynamicPartitionsAction` enum, vs. something like `AddDynamicPartitionsRequest` and `DeleteDynamicPartitionsRequest`.

An advantage of the latter is that we could potentially support different data structures in the future? E.g. what if we wanted to support the ability to rename a set of dynamic partitions, and we wanted to accept a dictionary that mapped old names to new names?

sryza:
Btw it's useful to have these together as an RFC, but when we get ready to merge, I think there should be separate PRs for:
- Adding the result class
- Adding dynamic partitions requests

sryza:
I wanted to raise one more thing, which isn't part of this PR, but has some interaction with it. What do you think about replacing

```python
my_job.run_request_for_partition(partition_key="abc", ...)
```

with a new `partition_key` argument on `RunRequest`:

```python
RunRequest(partition_key="abc", ...)
```

Aside from being less verbose, it has the advantage that it works for sensors that target asset selections and don't have a job at all. Then we might not need the `PendingPartitionRunRequest` that's in this PR?

A consequence would be that validation of the partition key happens a little bit later, but I think that's fine and consistent with other parts of how `RunRequest` works. E.g. if you specify an unknown job name for the `job_name` parameter, you don't get an error until you exit the sensor evaluation function.

schrockn:
* Re naming I think `SensorEvaluationResult` is a fine name and mirrors `SensorEvaluationContext` nicely.

* > @schrockn had concerns about DynamicPartitionsDefinition.add_partitions. @schrockn - do those concerns extend here?

Nope. I don't think this is a big mental model shift. It's just a helper for creating another immutable in-memory object.

My question is what is our API for adding partitions in an observable manner _outside_ the context of a sensor. During the execution of an op, for example. I think one risk here is that we introduce a sensor-specific API here, then add a more generic one for ops, then people want to use the generic one in the sensor context, and then we are left with supporting two way of doing things.

clairelin135:
One way that we could support mutating partitions within ops:
```
@op
def my_op():
    yield DynamicPartitionsRequest(DynamicPartitionsAction.ADD, partition_keys)
    yield Output(...)
```
Similarly for ops with dynamic outputs:
```
@op
def my_dynamic_op():
     for i in range(...):
           yield DynamicPartitionsRequest(DynamicPartitionsAction.ADD, [str(i)])
           yield DynamicOutput(i, mapping_key=str(i)
```
I thought about adding the dynamic partitions request on the output itself too, but that felt awkward as you might not want to yield an output, or you might yield outputs that don't map cleanly to the dynamic partition requests.

prha:
How are dynamic partitions resolved in terms of the execution lifecycle?  Can an upstream op and a downstream op as part of the same run have a different set of partitions for the same partitions definition?

clairelin135:
@prha I think we should not allow for mutating a dynamic partitions definition within a dynamically-partitioned run, and I can't think of any clear use cases for why someone would want to do this. So it shouldn't be possible for you to execute a run for a dynamic partition and then delete the partition mid run.

I think the cases above where you mutate a dynamic partitions definition within an op output/dynamic output are for backfill cases and should only occur within unpartitioned runs. I.e. bulk creating a dynamic partition for each file in a bucket, or deleting all dynamic partitions that are older than 1 yr.

In the dynamic output case, a downstream step that accepts a dynamic output as an input would have a partition key based on the mapping key, so you can do something like creating a partition for each file in a bucket, and logging metadata per file within a downstream mapped op. But ops that are not mapped within this run should not have a partition key.

sryza:
> Similarly for ops with dynamic outputs:

My instinct would be that, for the case where we want both dynamic and outputs and dynamic partitions, that these should be tightly coupled. I.e. every DynamicOutput should correspond to exactly one dynamic partition.  This is because:
- It's important that Dagster understands which dynamic outputs correspond to which partitions
- It seems very weird to launch downstream steps for some partitions but not other ones

clairelin135:
@sryza that makes sense. In the dynamic output case, doing something like this also makes sense to me:
@op
def my_dynamic_op():
     for i in range(...):
           yield DynamicPartitionsRequest(DynamicPartitionsAction.ADD, [str(i)])
           yield DynamicOutput(i, mapping_key=str(i)

clairelin135:
@sryza that makes sense. In the dynamic output case, we could instead do something like this:
```
@op
def my_dynamic_op():
     for i in range(...):
           yield DynamicOutput(i, dynamic_partition_request=dynamic_partitions_def.request_for_adding_partitions(["1"]))
```
and enforce that each dynamic output is mapped to only one requested partition.

Though this makes me wonder if we should instead create `AddDynamicPartitionsRequest` instead of a request w/ enum, so it's clearer that for a dynamic output the only action you can request is an "add".

clairelin135:
@schrockn, thoughts on `SensorResult`? I agree that `SensorEvaluationResult` is more consistent, but it does feel long

schrockn:
Fine with it

clairelin135:
If there is no other feedback, I'll move forward with this:
- `SensorResult` instead of `SensorTick`/`SensorEvaluationResult`
- `partitions_def.request_for_adding_partitions` returns `AddDynamicPartitionsRequest`, similar for deleting partitions. We don't use the enum pattern anywhere else in our codebase, so having an explicit class should be more consistent.

I'll break this PRs out into two parts:
- Adding the result class, refactoring run requests to validate partitions within sensor evaluation
- Adding dynamic partitions requests

PR #12822, opened 2023-03-08
Add a way to cause partition health data to refetch without passing around refetch callbacks

salazarm:
### Summary & Motivation

In the AssetView page we show a sidebar with all of our partitions for an asset which currently doesn't update when you create a new partition after clicking the "Materialize" button. 
Instead of threading through refetch callbacks through the MaterializeButton all the way down to the CreatePartitionDialog I'm adding this new module which allows CreatePartitionDialog to notify any subscribers of partition health data that they need to refetch.

Normally we could just add the query to `refetchQueries` however, in this case, the queries are being made without using `useQuery` so `refetchQueries` doesn't work.

https://www.loom.com/share/7c020e6034d94013828eb83d7dd6c51a

On the asset view page I added a new partition and saw the sidebar reflect the new partition.

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/PartitionSubscribers.ts
+++ b/js_modules/dagit/packages/core/src/assets/PartitionSubscribers.ts
@@ -1,0 +1,24 @@
+import React from 'react';
+
+const subscriptions: Array<() => void> = [];
+
+export function usePartitionDataSubscriber(onInvalidate: () => void) {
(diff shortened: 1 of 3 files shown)


bengotow (approved):
We talked about this on Slack - I agree this is one of those things best accomplished with a little global event. Adding a partition is somewhat rare and happens very far from these hooks and state.

I was actually wondering if maybe adding a partition should just totally reload the page, but that’s kinda lame!

Issue #12727, opened 2023-03-06
[dagster-dbt] - support a custom asset metadata function


leozqin:
### What's the use case?

Because of its user-friendly interface (and for Dagster Cloud, its pricing model), one of Dagster's compelling use cases is as a data catalog. `dagster-dbt` already uses the manifest to populate a table schema for each model that has one configured. However, there are other use pieces of information in the manifest that are not surfaced, such as the `config` and `meta` blocks. Providing a way to surface this information would increase Dagster's utility as a data catalog for dbt users.

### Ideas of implementation

I think that the best way to implement this functionality would be to support defining a custom function in `load_assets_from_dbt_project` / `load_assets_from_dbt_manifest` that maps node information to metadata values. This pattern already exists in `dagster-dbt` for asset keys, asset groups, and freshness policy. 

Some care would be needed to maintain an ergonomic interface for the existing functionality that yields a table schema, since that is also a case of an "asset-level metadata value derived from the dbt manifest".

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

OwenKephart:
Hi! This sounds like a great idea, and it looks like several people would find this useful. I threw up a PR to implement this behavior above. To be transparent, I think these function signatures are straining a bit underneath the load of all of these arguments, and in an ideal world there might be a more ergonomic way to expose more flexibility to this integration without having to continually tack on additional arguments. This is definitely something we're thinking about as we work to improve the dagster-dbt integration.

Would love to hear feedback on how annoying this `node_info_to..._fn` pattern is in practice (if it all!)

linssendane:
Hey @OwenKephart ! Thanks for the insight.

ATM I'm actually a big fan of the `node_info_to..._fn` pattern as it's quite powerful. I struggled a bit with it initially (the docs could be improved with concrete usage examples), but when I found out how to use it through your slack channel I was able to play around with it and perfectly tailor it to our situation! :)

leozqin:
@OwenKephart - appreciate your taking a look at this! I do also find the pattern to be useful because it lets you do pretty much anything in python-world with the node info. 

One alternative I could think of would be to expose an event interface (sort of like an `OpExecutionContext`) where we can emit metadata during asset definition... an `AssetDefinitionContext`, if you will. It would be more than just a dbt thing - rather, any scenario where you enumerate objects into Dagster assets could take advantage.

alex-orlovskyi:
[...]

PR #12831, opened 2023-03-09
[dagster-dbt] add support for node_info_to_definition_metadata_fn

OwenKephart:
### Summary & Motivation

Resolves: https://github.com/dagster-io/dagster/issues/12727

It hurts a bit to add yet another `node_info_to...` parameter to these signatures, and I think in the future we should investigate a nicer way to expose this sort of flexibility. However, in the short term this addresses an issue that many users are running into, and does not significantly diverge from the existing pattern.

Open to suggestions on the name. My goal was to disambiguate this from the `runtime_metadata_fn` parameter, and make it clear that this metadata will be wrapped up into the definition of the asset.


rexledesma (approved):

linssendane:
Thanks guys!!! You're the best! 🎉

Issue #12786, opened 2023-03-07
making materializing an asset with config less hidden


sryza:
Right now, you need to know to shift-click the Materialize button, which is not very discoverable.

Relevant feedback:
- https://dagster.slack.com/archives/C01U954MEER/p1678230002022199?thread_ts=1678127682.574859&cid=C01U954MEER

rahuldave:
For op based joba, there is an explicit launchpad button, and I can access run history. If i have previous runs of an asset based job, i can also access run history and get config from there. The shift-click materialize instead opens up in a modal, and has no access to previous runs, if in a job context. If not in a job context, access to previous materializations are still useful! My 2c: replace the modal on shift-click by a proper window on a launchpad tab...

PR #12852, opened 2023-03-09
Add dropdown for materializing using launchpad

salazarm:
### Summary & Motivation

Addresses https://github.com/dagster-io/dagster/issues/12786

Screen Shot 2023-03-09 at 2 05 06 PM

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
+++ b/js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
@@ -196,61 +196,54 @@ export const LaunchAssetExecutionButton: React.FC<{
           position="bottom-right"
           useDisabledButtonTooltipFix
         >
-          <JoinedButtons>
-            <MaterializeButton
-              intent={intent}
-              data-testid={testId('materialize-button')}
-              onClick={(e) => onClick(firstOption.assetKeys, e)}
-              style={
-                options.length > 1
-                  ? {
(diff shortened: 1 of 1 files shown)


braunjj:
What happens for versioned assets? I think currently we show a "Materialize stale and missing" in the dropdown. Does this conflict with that?

js_modules/dagit/packages/core/src/assets/LaunchAssetExecutionButton.tsx
```diff
@@ -186,24 +196,33 @@ export const LaunchAssetExecutionButton: React.FC<{
...
+              }
+              disabled={!firstOption.assetKeys.length}
+              icon={loading ? <Spinner purpose="body-text" /> : <Icon name="materialization" />}
+            >
+              {firstOption.label}
+            </MaterializeButton>
+            <MaterialzeButtonOptions
+              openWithLaunchpad={(e: React.MouseEvent<any>) => {
+                e.shiftKey = true;
+                onClick(firstOption.assetKeys, e);
```
bengotow, line 221: Weird but seems fine to me since it's a fairly obvious thing. Might be good to double check that shiftKey is writable in all browser though.

bengotow (approved):

salazarm:
> What happens for versioned assets? I think currently we show a "Materialize stale and missing" in the dropdown. Does this conflict with that?

Good catch. Just found that codepath and realized I should be re-using that logic

Issue #12602, opened 2023-02-28
Add a tooltip on asset nodes in the graph


braunjj:
Currently, assets names get truncated in the asset graph, this makes it hard to distinguish many similarly named assets from each other. 

We should display the full asset key on hover. 

Requested in Slack: https://elementl-workspace.slack.com/archives/C04P1BJH88M/p1677594654236979

PR #12853, opened 2023-03-09
Add a tooltip on asset nodes in the graph

salazarm:
### Summary & Motivation

Resolves https://github.com/dagster-io/dagster/issues/12602

Screen Shot 2023-03-09 at 2 39 52 PM
Screen Shot 2023-03-09 at 2 39 49 PM


bengotow (approved):
This looks good to me but I'm a little wary of putting immediate-on-hover tooltips on the asset graph. (Though we did recently do it for partitioned asset counts). When you're zoomed out a lot, it could just sorta be tooltips-galore 😅 but let's try it and see what people think!

braunjj:
sweet. Thanks for tackling this @salazarm!

PR #12855, opened 2023-03-09
Add volatility property to CachingStaleStatusResolver

smackesey:
### Summary & Motivation

- Add `_is_volatile` on `CachingStaleStatusResolver`.
    - This required exposing observability on `AssetGraph`.
- Remove `StaleRootCause`, which is dead code that should've been previously removed.

BK

No new tests since this property is at present internal to the `CachingStaleStatusResolver` and not used for anything. Testing will be done on the outputs of `CachingStaleStatusResolver` that volatility informs.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/data_version.py
+++ b/python_modules/dagster/dagster/_core/definitions/data_version.py
@@ -255,14 +255,6 @@ class StaleCause(NamedTuple):
     children: Optional[Sequence["StaleCause"]] = None
 
 
-# Root reasons for staleness. Thes differ from `StaleStatusCause`in that there is no status
-# associated with them-- rather they are causes for the staleness of some downstream node.
-class StaleStatusRootCause(NamedTuple):
-    key: AssetKey
-    reason: str
-    dependency: Optional[AssetKey] = None
-
-
 class CachingStaleStatusResolver:
     """
     Used to resolve data version information. Avoids redundant database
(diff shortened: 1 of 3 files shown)


smackesey:
* master
  * **PR #12855** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12855" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12855?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/data_version.py
```diff
@@ -458,3 +450,14 @@ def _is_partitioned_or_downstream(self, *, key: AssetKey) -> bool:
                 self._is_partitioned_or_downstream(key=dep_key)
                 for dep_key in self.asset_graph.get_parents(key)
             )
+
+    # Volatility means that an asset is assumed to be constantly changing. We assume that observable
+    # source assets are non-volatile, since the primary purpose of the observation function is to
+    # determine if a source asset has changed. We assume that regular assets are volatile if they
+    # are at the root of the graph (have no dependencies).
+    @cached_method
```
OwenKephart, line 458: Originally, I was thinking (incorrectly) that this volatility thing would be recursively defined, but that's not the case. So maybe it makes more sense for this to just live on the asset graph? The operation is cheap so I don't see any benefit to it being cached

OwenKephart (approved):
would be good to put some common situations under test but the logic looks good

smackesey:
>would be good to put some common situations under tes

Prefer to avoid this atm since right now `_is_volatile` is just an implementation detail of `CachingStaleStatusResolver`. If we expose it, then I would support testing it directly the way the public API is tested

PR #12858, opened 2023-03-09
Enable dynamic partitions dimensions on multipartitions defs

clairelin135:
This PR enables `DynamicPartitionsDefinition`s to be dimensions of `MultiPartitionsDefinition`s.

Since we allow caching multipartitions statuses in the asset status cache, this PR enables dynamic partitions statuses to also be cached to bring this up to parity with multipartitions. This means that the `serializable_unique_identifier` we use to check whether a partitions definition has remained the same or not must be changed to a function that accepts a `DynamicPartitionsStore`, in order to build an identifier containing the dynamic partition keys.

Tested via BK and locally on Dagit.


sryza:
> Since we allow caching multipartitions statuses in the asset status cache, this PR enables dynamic partitions statuses to also be cached to bring this up to parity with multipartitions.

It would be helpful to see this broken out into its own change, with a brief discussion of the perf implications - i.e. whether there are any situations we'd expect to be slower with that change.

mgierada:
Can't wait for that! Great job 🎉

clairelin135:
Per @sryza's suggestion above, this PR is now broken into two chunks:
- Part 1 enables dynamic dimensions on multipartitions: https://github.com/dagster-io/dagster/pull/12894
- Part 2 enables caching dynamic partitions statuses: https://github.com/dagster-io/dagster/pull/12897

PR #12871, opened 2023-03-10
add docs page on asset selection syntax

sryza:


clairelin135 (approved):
The content here looks great to me. Maybe it's worth adding some <info> bar that emphasizes that this selection is not an `AssetSelection`, and linking to the `AssetSelection` API docs? I think this is a current point of confusion for users that these two entities are different

sryza:
Good point. I added a mention.

Issue #12854, opened 2023-03-09
AssetValueLoader.load_asset_value() doesn't handle indirect required resources, causing a 'KeyError'


krywicki:
### Dagster version

1.1.21

### What's the issue?

When loading an asset via `AssetValueLoader.load_asset_value()`, an exception is raised if the asset's `io_manager` requires a resource that has required resources. 

The `AssetValueLoader.load_asset_value` only creates the required resources needed by it's io_manager, but it only looks at directly required resource keys and not indirect ones. 

### What did you expect to happen?

I expect `AssetValueLoader.load_asset_value()` to create all directly _and_ indirectly required resources needed by an io_manager.

### How to reproduce?

The following code produces a `KeyError: 'resource_b'` during `loader.load_asset_value(AssetKey("my_asset"))` call.

```python
from dagster import AssetKey, Definitions, IOManager, asset, io_manager, resource


@resource(required_resource_keys={"resource_b"})
def resource_a():
    return None

@resource
def resource_b():
    return None

@asset
def my_asset() -> None:
    pass

class DummyIOManager(IOManager):
    def handle_output(self, context, obj):
        raise NotImplementedError()

    def load_input(self, context):
        return "some value"

@io_manager(required_resource_keys={"resource_a"})
def dummy_io_manager():
    return DummyIOManager()

defs = Definitions(
    assets=[my_asset],
    resources={"resource_a": resource_a, "resource_b": resource_b, "io_manager": dummy_io_manager},
)

with defs.get_asset_value_loader() as loader:
    loader.load_asset_value(AssetKey("my_asset"))
```

### Deployment type

Local

### Additional information

Adding indirect required resource keys to the `required_resource_keys` variable [here](https://github.com/dagster-io/dagster/blob/0c36c230ba570ef7bd2e59ab7ec87804a5c52a71/python_modules/dagster/dagster/_core/storage/asset_value_loader.py#L101) should address this problem. 

I modified the function locally in such a way, and it seemed to address the issue. 

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

PR #12872, opened 2023-03-10
Fix transitive resource deps for asset value loader

dpeng817:
Fixes https://github.com/dagster-io/dagster/issues/12854


python_modules/dagster/dagster/_core/storage/asset_value_loader.py
```diff
@@ -98,7 +99,10 @@ def load_asset_value(
         )
         io_manager_key = assets_def.get_io_manager_key_for_asset_key(asset_key)
         io_manager_def = resource_defs[io_manager_key]
-        required_resource_keys = io_manager_def.required_resource_keys | {io_manager_key}
+        get_transitive_required_resource_keys(io_manager_def.required_resource_keys, resource_defs)
```
sryza, line 102: Is this line on purpose?

dpeng817: yes - this is what enables the transitive resource dependencies. They didn't work before this.

dpeng817: nope

sryza (approved):
nice

PR #7538, opened 2022-04-22
[RFC] lighter weight pull request template

alangenfeld:
The existing template was created when we were internally using phabricator, so the target audience was only external contributors.

Now that we are using GH, I feel the template should strike a better balance between having enough information for external contributors but also being stream lined for us internally.

Personally, this was motiviated by seeing internal PRs that submit without filling out the template (or editing it away).

Very open to feedback on this, just wanted to start a discussion.

PR #12875, opened 2023-03-10
Move docs style guidelines from pull request template to contribution guide

sryza:
## Summary & Motivation

I am a huge fan of our docs style guidelines and making them publicly available.

However, I'd also like to advocate for keeping our pull request template as lightweight as possible, to minimize cognitive load when submitting a PR. It used to have a lot more in it, and @alangenfeld made a change to trim it down to the bare essentials: https://github.com/dagster-io/dagster/pull/7538. That change improved my quality of life.

This change adds the style guidelines to the top of the docs section of the contribution and removes it from the PR template.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,7 +1,3 @@
 ## Summary & Motivation
 
-> **Does this pull request update the docs?** If so, please verify that the updates follow the [docs style checklist](https://github.com/dagster-io/dagster/blob/master/docs/DOC_CHECKLIST.md).
-
----
-
 ## How I Tested These Changes
--- a/docs/content/community/contributing.mdx
+++ b/docs/content/community/contributing.mdx
@@ -96,6 +96,8 @@ To run all of them together, run `yarn test`.
 
 ## Developing Docs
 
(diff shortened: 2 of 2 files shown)


erinkcochran87:
Qq- this wasn't mentioned on my PR earlier today that merged this in. Why now?

sryza:
@erinkcochran87 I liked the idea of the change when I saw your PR!

Then, after it was merged, I made some PRs of my own – non-docs changes where I had to delete that line – and remembered my general feelings about our PR template.

erinkcochran87:
Gotcha. Fair enough! I'm just concerned about folks not finding that info is all, but I don't want to create headaches or anything.

alangenfeld:
Are there any features around how github templates work where we could only trigger this in docs directory changes? 

If not and we can only have one global one for the repository - i think it makes sense to optimize for the more common case of regular development

erinkcochran87:
> Are there any features around how github templates work where we could only trigger this in docs directory changes?

It looks like we can have multiple templates ([link](https://github.blog/2018-01-25-multiple-issue-and-pull-request-templates/)) but switching between them would be manual. There is probably something we can do with GitHub actions to get this to work, but I don't think it's necessary to spend a ton of time on it at the moment.

sryza:
@alangenfeld @yuhan or @erinkcochran87 – mind giving this an "Approved" or "Request Changes"?

alangenfeld (approved):

PR #12903, opened 2023-03-10
[ui] Refactor permissions objects

hellendag:
## Summary & Motivation

Work toward repairing the Permissions footgun, in which it's too easy to check the `PermissionResult` object for truthiness instead of looking at its `enabled` value.

Return a `PermissionsState` object from `useUnscopedPermissions` and `usePermissionsForLocation`. This includes a `permissions` map (booleans), a `disabledReasons` map (strings), and a `loading` boolean.

This means that for a given permission, the "enabled" state is separate from its `disabledReason`, and truthiness checks are simple boolean checks.

I have a temporary shim for Cloud types that I'll clean up right away.

## How I Tested These Changes

TS, lint, jest. Load Cloud and view a job with the "Viewer" role, verify that the launchpad is disabled.


hellendag:
* master
  * **PR #12903** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12903" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12903?utm_source=stack-comment).

js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
```diff
@@ -370,9 +371,9 @@ const LaunchAssetChoosePartitionsDialogBody: React.FC<Props> = ({
   };
 
   const launchButton = () => {
-    if (launchAsBackfill && !canLaunchPartitionBackfill.enabled) {
+    if (launchAsBackfill && !canLaunchPartitionBackfill) {
       return (
-        <Tooltip content={canLaunchPartitionBackfill.disabledReason}>
+        <Tooltip content={disabledReasons.canLaunchPartitionBackfill}>
```
bengotow, line 376: This all looks great! I'm a big fan of removing this foot cannon... I wish Prettier would keep the destructuring to one line, but oh well.

One other thought -- I like the structure in this PR because we could potentially have other attributes on each permission in other dicts like `disabledReasons` BUT if it's just going to be these two, we could also flatten it all into a single field by flipping from enabled to disabled, and using the disabled data as the truthy values. 

(eg:  `disableLaunchPipelineExecution: {message: string} || false`)

I think I like the way you have it better though? At least, switching from "enabled = true" to "disabled = false" would be a more significant change!  😬

bengotow (approved):

js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
hellendag: Interesting thought! I think I'd be a little wary of representing the affirmative ability to take an action as a double-negative -- "I can launch this job, so it's `!disabled`" feels a little trickier to think through. (There are also other reasons why a control might be "disabled", so the negative version of this might be better represented as `cannotLaunchPipelineExecution`.)

I'll go ahead with this PR as is, and we can evaluate. :)

Issue #12867, opened 2023-03-09
Make schedule page respect the set timezone for the schedule


dpeng817:
schedules have a `cron_schedule_timezone` argument - but the page for a schedule still always defaults to UTC. This can end up being quite confusing for users. It should default to the set timezone for the schedule (if provided). 

Relevant thread: https://dagster.slack.com/archives/C01U954MEER/p1678395603160099

bengotow:
Hey @OwenKephart, I created an asset with the config in the screenshot below:

```
@asset(
    partitions_def=hourly_partitions_def2,
    freshness_policy=FreshnessPolicy(maximum_lag_minutes=60, cron_schedule="0 0,6,12,18 * * *", cron_schedule_timezone="America/Los_Angeles"),
)
def test_hourly_march8_2(context):
    time.sleep(1)

```

But it renders correctly for me in the UI:

<img width="510" alt="image" src="https://user-images.githubusercontent.com/1037212/224437685-6b6e459e-3ee6-4172-b002-84a27fecf864.png">

Maybe this is specific to the adapter that passes `dagster_freshness_policy` through? It seems like the field is being ignored.

PR #12904, opened 2023-03-10
[dagster-dbt] add cron_schedule_timezone to set of properties parsed from dbt config

OwenKephart:
## Summary & Motivation

Resolves: https://github.com/dagster-io/dagster/issues/12867

---


rexledesma (changes requested):
would like to see a test to assert the functionality here

python_modules/libraries/dagster-dbt/dagster_dbt_tests/utils.py
```diff
@@ -45,7 +45,9 @@ def assert_assets_match_project(dbt_assets, prefix=None, has_non_argument_deps=F
             maximum_lag_minutes=123
         ),
         AssetKey(prefix + ["cold_schema", "sort_cold_cereals_by_calories"]): FreshnessPolicy(
-            maximum_lag_minutes=123, cron_schedule="0 9 * * *"
+            maximum_lag_minutes=123,
+            cron_schedule="0 9 * * *",
+            cron_schedule_timezone="America/New_York",
```
OwenKephart, line 50: note: this test did not originally start failing because the manifest.json file that was being read from was not updated to contain the new value present in the model sql file.

rexledesma (approved):

PR #12906, opened 2023-03-10
[freshness-refactor][3/n] Update methods on the CachingDataTimeResolver to work with scalar data time

OwenKephart:
## Summary & Motivation

A few things happening here:
- some renaming of functions on the CachingDataTimeResolver
- added some functionality for common patterns (such as "how many minutes late is this key"). Previously multiple callsites would do repeated work of (get freshness policy for the key, get most recent record for that key, get the data times for that record, pass those into the freshness policy). Now it's just a single function call.
- made the AssetGraph a property of the resolver, to prevent it from needing to be passed into every function call

And, the largest change:
- refactor most internal functions to work with a single data time, rather than a mapping of asset key to upstream data time. previously, we represented data time as a mapping from root asset key to timestamp. This added a ton of complexity, and this high-fidelity information is not used anywhere except on the AssetDetails page. The core reconciliation logic has been simplified a ton by virtue of not having to unpack dictionaries all the time.

As an added bonus, for dense asset graphs, this dictionary business was actually causing noticeable impacts on perf. An unscientific test indicates that this is actually a fairly sizable (~15-30%) performance win  for very large asset graphs. 

---


python_modules/dagster-graphql/dagster_graphql/implementation/fetch_assets.py
```diff
@@ -602,42 +600,17 @@ def get_2d_run_length_encoded_partitions(
...
-        used_data_times=used_data_times,
-    )
-
     return GrapheneAssetFreshnessInfo(
-        currentMinutesLate=current_minutes_late,
-        latestMaterializationMinutesLate=latest_materialization_minutes_late,
+        currentMinutesLate=data_time_resolver.get_current_minutes_late(
+            asset_key, evaluation_time=current_time
+        ),
+        latestMaterializationMinutesLate=None,
```
OwenKephart, line 613: fyi: @bengotow -- we were not using this value anywhere so I set it to None instead of maintaining the code to populate it. I assume it'd be fine to just remove this field from the GrapheneAssetFreshnessInfo entirely? Just didn't want to make this PR larger than necessary

PR #12907, opened 2023-03-10
Build schedule from multipartitioned job

clairelin135:
Requested by multiple users. Issue here: https://github.com/dagster-io/dagster/issues/10800

### DIFF
--- a/python_modules/dagster/dagster_tests/definitions_tests/test_multi_partitions.py
+++ b/python_modules/dagster/dagster_tests/definitions_tests/test_multi_partitions.py
@@ -350,3 +350,42 @@ def test_asset_partition_key_is_multipartition_key():
         resources={"my_io_manager": MyIOManager()},
         partition_key="a|b",
     )
+
+
+def test_keys_with_dimension_value():
+    static_keys = ["a", "b", "c", "d"]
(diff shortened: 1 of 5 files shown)


python_modules/dagster/dagster_tests/definitions_tests/test_multi_partitions.py
```diff
@@ -350,3 +350,42 @@ def asset2(context, my_asset):
...
+    ) == [
+        MultiPartitionKey({"static": val[0], "date": val[1]})
+        for val in [
+            ("a", "2015-01-01"),
+            ("a", "2015-01-02"),
+            ("a", "2015-01-03"),
+            ("a", "2015-01-04"),
+        ]
+    ]
+    print(
```
sryza, line 376: get rid of print statement?

sryza (approved):
This is awesome.

I think it's worth adding a sentence to the docstring

python_modules/dagster/dagster_tests/definitions_tests/test_multi_partitions.py
clairelin135: thanks for the catch

PR #12917, opened 2023-03-13
Unpin grpcio on dagster for python 3.11

gibsondan:
Summary:
We are stuck between a rock and a hard place on Python 3.11 due to two issues:
a) https://github.com/grpc/grpc/issues/31885 is keeping us pinned <1.48 due to issues related to grpc subprocess forking (likely only an issue while running dagster locally, and potentially only in some edge cases involving multiple gRPC servers - it's hard to be certain
b) 3.11 wheels are only available on later versions, so dagster effectively can't build on python 3.11

This PR tries to have our cake and eat it too, by unpinning us, but only on python 3.11 - being able to install the package at all seems better than it behaving without any hangs or crashes during local dev.

Test Plan:

Build dagster locally in a python 3.11 venv - now succeeds
Upcoming PR will add full 3.11 BK coverage

## Summary & Motivation

> **Does this pull request update the docs?** If so, please verify that the updates follow the [docs style checklist](https://github.com/dagster-io/dagster/blob/master/docs/DOC_CHECKLIST.md).

---

### DIFF
--- a/python_modules/dagster/setup.py
+++ b/python_modules/dagster/setup.py
@@ -73,9 +73,11 @@ setup(
         # pin around issues in specific versions of alembic that broke our migrations
         "alembic>=1.2.1,!=1.6.3,!=1.7.0",
         "croniter>=0.3.34",
-        # grpcio>=1.48.1 has hanging/crashing issues: https://github.com/grpc/grpc/issues/30843 and https://github.com/grpc/grpc/issues/31885
         # grpcio 1.44.0 is the min version compatible with both protobuf 3 and 4
-        "grpcio>=1.44.0,<1.48.1",
+        # Also pinned <1.48.0 until the resolution of https://github.com/grpc/grpc/issues/31885
+        # (except on python 3.11, where newer versions are required just to install the grpcio package)
+        "grpcio>=1.44.0,<1.48.0; python_version<'3.11'",
+        "grpcio>=1.44.0; python_version>='3.11'",
         "grpcio-health-checking>=1.44.0",
         "packaging>=20.9",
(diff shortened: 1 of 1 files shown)


alangenfeld:
have we seen unpinned grpc cause problems in the test suite on all other python versions? Wondering if this unpin should for example be for 3.10 +

gibsondan:
Yeah, it's consistent for 3.7 - 3.10 (interestingly early data suggests it may be gone on 3.11? still testing)

gibsondan:
Nah they still happen on 3.11 too, that was optimistic

alangenfeld (approved):

emirkmo:
I think the grpcIO issue should be included in the changelog for dagster being tested on 3.11. I was hoping it was fixed and jumped the gun upgrading :P

gibsondan:
Have you run into any problems after upgrading? The problems we’ve been
seeing with grpcio seem to mostly be happening when lots of code locations
are updating at once under stress testing, but if you’re seeing problems
during normal usage that would be very useful information.

emirkmo:
Well.. did not update prod. I will keep a branch on 3.11 and report if I find issues. I assumed the problem would come when we deploy on K8s.

gibsondan:
So actually if there's a problem I would expect it to be the opposite and only happen during local dev (where multiple grpc servers are being managed in a single parent process). In k8s every grpc server has its own totally isolated pod/container so I would not expect the upstream issue with subprocess forks to come into play.

gibsondan:
but the point that we could have been more detailed about this in the changelog is well taken

PR #12919, opened 2023-03-13
[Partitions] Use OrdinalPartitionSelector for STATIC partitioned assets

salazarm:
## Summary & Motivation

> **Does this pull request update the docs?** If so, please verify that the updates follow the [docs style checklist](https://github.com/dagster-io/dagster/blob/master/docs/DOC_CHECKLIST.md).

We added the "OrdinalPartitionSelector" component for dynamic partitions to be able to select partitions using a dropdown.
This PR updates static definitions to also use this component too (with an extra isDynamic flag used to enable/disable adding partitions, Static definitions cannot have new partitions added). Previously these components were using the drag/drop range selector component which makes less sense for static partitions.

This PR is a bit of a stepping stone to the modal redesign https://www.figma.com/file/RFJ1i5WlEq3CHIlY6UyyYX/Asset-Partition-Dialog?node-id=6%3A3685&t=kikWzIfGSDMQ2REG-0

---

## How I Tested These Changes

I tested this with 2 static definitions, and a static mixed with a time window definition.

I tried mixing a dynamic with static/timeseries but that's not supported. We also only support up 2 partitions atm.

See screenshot:
Screen Shot 2023-03-13 at 10 35 38 AM


bengotow (approved):
This looks good to me and I think this is the way it should work 👍

PR #12920, opened 2023-03-13
[long-running runs] long-running run termination

dpeng817:
Adds ability to terminate long-running runs.

**User experience changes**
- A new `MAX_RUNTIME_TAG` for a user to set to a seconds-timestamp value of how long to wait before termination. 
- run monitoring daemon becomes available on every run launcher. Instead, we gate specific run monitoring behavior per run launcher (IE resume run does not happen)
- When a run is terminated due to maximum runtime, a run cancellation event is sent with a message describing that this occurred.

**Testing**
- Added testing for the too long, not too long, and no tag set evaluation cases.
- Manually tested by setting a sleepy job:
```python
import time

from dagster import job, op
from dagster._core.storage.tags import MAX_RUNTIME_TAG


@job(tags={MAX_RUNTIME_TAG: "60"})
def the_job():
    @op
    def sleep_op():
        time.sleep(1000)

    sleep_op()
```
Ensured that after launching a run (using default run launcher) upon hitting approx. 60 seconds (bc you have to wait for daemon to evaluate, timing is not exact) the run terminated:
Screen Shot 2023-03-13 at 9 27 30 AM

**Additional considerations**
- Changed a bunch of monitoring code to utilize run record instead of run, so that we can get the start_time of the run easily.


dpeng817:
* master
  * **PR #12920** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12920" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12920?utm_source=stack-comment).

johannkm:
> RUN_TERMINATED` event, instead of just RUN_CANCELED. Another option would be to provide a cancellation reason / message that we can render in dagit. 

strong support for the latter

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1510,6 +1506,26 @@ def get_run():
         except DagsterRunAlreadyExists:
             return get_run()
 
+    def handle_started_run(self, run_record: RunRecord, logger: logging.Logger) -> None:
+        max_time_str = run_record.dagster_run.tags.get(MAX_RUNTIME_TAG)
+
+        # If MAX_RUNTIME_TAG has not been set, nothing to do
+        if not max_time_str:
+            return
```
gibsondan, line 1514: falling back to an optional system-wide default seems good here? So you can just say "kill any run that goes over 12 hours that should never happen"

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1510,6 +1506,26 @@ def get_run():
         except DagsterRunAlreadyExists:
             return get_run()
 
+    def handle_started_run(self, run_record: RunRecord, logger: logging.Logger) -> None:
```
gibsondan, line 1509: I'm not super convinced this should be on the instance? The instance is already very out of control

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1510,6 +1506,26 @@ def get_run():
...
+        max_time_str = run_record.dagster_run.tags.get(MAX_RUNTIME_TAG)
+
+        # If MAX_RUNTIME_TAG has not been set, nothing to do
+        if not max_time_str:
+            return
+        max_time = float(max_time_str)
+        if (
+            run_record.start_time is not None
+            and pendulum.now().timestamp() - run_record.start_time > max_time
```
gibsondan, line 1518: is this just hipster `time.time()`

gibsondan: if you use pendulum i believe you want pendulum.now("UTC") (not positive if that matters if you then call .timestamp() on it)

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -1510,6 +1506,26 @@ def get_run():
...
+        ):
+            logger.info(
+                f"Run {run_record.dagster_run.run_id} has exceeded maximum runtime of"
+                f" {max_time} seconds: terminating run."
+            )
+            self.run_launcher.terminate(
+                run_id=run_record.dagster_run.run_id,
+                message="Run has exceeded maximum allowed runtime.",
+            )
```
gibsondan, line 1527: i think we may want something like this where it moves it into a CANCELED state no matter what but still tries to clean it up (instead of maybe possibly terminating it if everything works out): https://github.com/dagster-io/dagster/blob/master/python_modules/dagster-graphql/dagster_graphql/implementation/execution/__init__.py#L104-L121

Arguably all run terminations should work this way

gibsondan (commented):
Overall I am a fan - johann should take a look too. Some thoughts inline

python_modules/dagster/dagster/_core/instance/__init__.py
dpeng817: Originally had it on instance in case we wanted different behavior on different instance types, but I don't think it's actually needed.

python_modules/dagster/dagster/_core/instance/__init__.py
dpeng817: pendulum is useful for testing bc you can use `pendulum.test` to freeze `now`

python_modules/dagster/dagster/_daemon/monitoring/monitoring_daemon.py
```diff
@@ -3,35 +3,43 @@
...
+    run = run_record.dagster_run
     check.invariant(run.status == DagsterRunStatus.STARTING)
     run_stats = instance.get_run_stats(run.run_id)
 
     launch_time = check.not_none(
         run_stats.launch_time, "Run in status STARTING doesn't have a launch time."
     )
-    if time.time() - launch_time >= instance.run_monitoring_start_timeout_seconds:
+    if (
+        instance.run_launcher.supports_check_run_worker_health
```
johannkm, line 40: this check isn't needed is is?

python_modules/dagster/dagster/_daemon/monitoring/monitoring_daemon.py
```diff
@@ -3,35 +3,43 @@
...
     DagsterRunStatus,
+    RunRecord,
     RunsFilter,
 )
+from dagster._core.storage.tags import MAX_RUNTIME_TAG
 from dagster._core.workspace.context import IWorkspace, IWorkspaceProcessContext
 from dagster._utils import DebugCrashFlags
 from dagster._utils.error import SerializableErrorInfo, serializable_error_info_from_exc_info
 
+DEFAULT_MAX_RUNTIME = 60 * 60 * 12
```
johannkm, line 25: what is this? I don't think we should have a default limit

python_modules/dagster/dagster/_core/storage/tags.py
```diff
@@ -57,6 +57,8 @@
 RETRY_NUMBER_TAG = "{prefix}retry_number".format(prefix=SYSTEM_TAG_PREFIX)
 RETRY_STRATEGY_TAG = "{prefix}retry_strategy".format(prefix=SYSTEM_TAG_PREFIX)
 
+MAX_RUNTIME_TAG = "{prefix}max_runtime".format(prefix=SYSTEM_TAG_PREFIX)
+
 USER_EDITABLE_SYSTEM_TAGS = [PRIORITY_TAG, MAX_RETRIES_TAG, RETRY_STRATEGY_TAG]
```
johannkm, line 62: needs to be in this list or else users can't set it in code, only via dagit

python_modules/dagster/dagster/_core/launcher/base.py
```diff
@@ -69,7 +69,7 @@ def launch_run(self, context: LaunchRunContext) -> None:
         """
 
     @abstractmethod
-    def terminate(self, run_id: str) -> bool:
+    def terminate(self, run_id: str, message=None) -> bool:
```
johannkm, line 72: do we write this to the event log somewhere?

johannkm (changes requested):

python_modules/dagster/dagster/_daemon/monitoring/monitoring_daemon.py
dpeng817: https://github.com/dagster-io/dagster/pull/12920#discussion_r1134645178 @gibsondan disagrees

python_modules/dagster/dagster/_core/launcher/base.py
dpeng817: artifact from a bygone implementation

python_modules/dagster/dagster/_daemon/monitoring/monitoring_daemon.py
dpeng817: I believe it is needed - since the run termination stuff isn't bound by the same constraints as run worker health checks, I got rid of the check for having the run monitoring daemon only enabled when certain run launchers are usable, and instead moved the check here.

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -404,11 +403,6 @@ def __init__(
         self._subscribers: Dict[str, List[Callable]] = defaultdict(list)
 
         run_monitoring_enabled = self.run_monitoring_settings.get("enabled", False)
-        if run_monitoring_enabled and not self.run_launcher.supports_check_run_worker_health:
-            run_monitoring_enabled = False
-            warnings.warn(
-                "The configured run launcher does not support run monitoring, disabling it.",
-            )
```
dpeng817, line 411: [1]

python_modules/dagster/dagster/_daemon/monitoring/monitoring_daemon.py
johannkm: I think it could be reasonable to offer that as an optional configuration alongside the start timeout https://github.com/dagster-io/dagster/blob/master/helm/dagster/values.yaml#L1095 though I think it will either be very long or cause unintended interrupts. I definitely don't think we should have a hard coded max run time for all jobs unless you specifically override it with a tag. Users should be able to opt in one job to a timeout without affecting others.

python_modules/dagster/dagster/_daemon/monitoring/monitoring_daemon.py
johannkm: That's to guard the calls to `check_run_worker_health`, which we aren't calling here. That may change soon with @gibsondan 's debug improvements but then we could just gate that

There's no particular reason why starting timeouts weren't supported for all run launchers so if we're making the change for max run time then may as well support it for starting too

python_modules/dagster/dagster/_daemon/monitoring/monitoring_daemon.py
dpeng817: Maybe we just leave the default bit out for now - and address later based on actual feedback, because it's unclear to me how the tag ergonomics would work there

python_modules/dagster/dagster/_core/launcher/sync_in_memory_run_launcher.py
```diff
@@ -42,5 +42,5 @@ def launch_run(self, context: LaunchRunContext) -> None:
         recon_pipeline = recon_pipeline_from_origin(context.pipeline_code_origin)  # type: ignore
         execute_run(recon_pipeline, context.dagster_run, self._instance)
 
-    def terminate(self, run_id):
+    def terminate(self, run_id, message=None):
```
gibsondan, line 45: unintentional right?

python_modules/dagster/dagster/_core/storage/tags.py
```diff
@@ -57,7 +57,9 @@
 RETRY_NUMBER_TAG = f"{SYSTEM_TAG_PREFIX}retry_number"
 RETRY_STRATEGY_TAG = f"{SYSTEM_TAG_PREFIX}retry_strategy"
 
-USER_EDITABLE_SYSTEM_TAGS = [PRIORITY_TAG, MAX_RETRIES_TAG, RETRY_STRATEGY_TAG]
+MAX_RUNTIME_TAG = f"{SYSTEM_TAG_PREFIX}max_runtime"
```
gibsondan, line 60: this is seconds right? Will people assume that? max_runtime_seconds is kind of a mouthful but is an option

python_modules/dagster/dagster/_daemon/monitoring/monitoring_daemon.py
```diff
@@ -102,34 +110,83 @@ def execute_monitoring_iteration(
...
+            f" {max_time} seconds: terminating run."
+        )
+
+        instance.report_run_canceling(
+            run_record.dagster_run,
+            message=f"Canceling due to exceeding maximum runtime of {max_time} seconds.",
+        )
+        try:
+            instance.run_launcher.terminate(run_id=run_record.dagster_run.run_id)
+        except:
```
gibsondan, line 173: i think we should make this work more like the FORCE_TERMINATE_IMMEDIATELY case in graphql and always move it into a CANCELED state no matter what (while still trying to terminate it cleanly if possible) - we shouldn't ever leave runs in CANCELING when this happens

python_modules/dagster/dagster_tests/daemon_tests/test_monitoring_daemon.py
```diff
@@ -151,42 +176,156 @@ def test_monitor_starting(instance: DagsterInstance, logger: Logger):
...
+        # termination event should be triggered.
+        monitor_started_run(instance, workspace, too_long_record, logger)
+        run = instance.get_run_by_id(too_long_record.dagster_run.run_id)
+        assert run
+        assert len(run_launcher.termination_calls) == 1
+
+
+def test_long_running_termination_failure(
+    instance: DagsterInstance, workspace_context: WorkspaceProcessContext, logger: Logger
+):
```
gibsondan, line 294: similar-ish test case that verifies that it still terminates even if the terminate doesn't happen cleanly would be nice (as per suggestion above)

gibsondan (changes requested):
nice - the API here might actually be worth a quick product-review post?

the big suggestion is to always move it into CANCELED when this timeout is hit, no matter what

python_modules/dagster/dagster/_core/launcher/sync_in_memory_run_launcher.py
dpeng817: argh thought I caught all of these

python_modules/dagster/dagster/_core/storage/tags.py
dpeng817: yea I kind of like the explicit-ness of it (and yes it is in seconds)

python_modules/dagster/dagster_tests/daemon_tests/test_monitoring_daemon.py
dpeng817: by terminate cleanly - you mean forcing that it enters failure state (check new tests I think they handle this case as you would expect now)

python_modules/dagster/dagster/_daemon/monitoring/monitoring_daemon.py
```diff
@@ -102,34 +114,93 @@ def execute_monitoring_iteration(
...
+        try:
+            if instance.run_launcher.terminate(run_id=run_record.dagster_run.run_id):
+                instance.report_run_failed(
+                    run_record.dagster_run, f"Exceeded maximum runtime of {max_time} seconds."
+                )
+        except:
+            instance.report_engine_event(
+                (
+                    "Exception while attempting to force-terminate run. Run will still be marked as"
+                    " failed."
```
gibsondan, line 192: i don't think we need to call it force-terminate here

python_modules/dagster/dagster/_daemon/monitoring/monitoring_daemon.py
```diff
@@ -102,34 +114,93 @@ def execute_monitoring_iteration(
...
+
+        instance.report_run_canceling(
+            run_record.dagster_run,
+            message=f"Canceling due to exceeding maximum runtime of {max_time} seconds.",
+        )
+        try:
+            if instance.run_launcher.terminate(run_id=run_record.dagster_run.run_id):
+                instance.report_run_failed(
+                    run_record.dagster_run, f"Exceeded maximum runtime of {max_time} seconds."
+                )
```
gibsondan, line 187: can you show an event log for a run that goes through this? This is the first time that i'm aware of that we knowingly terminate a run while its in a FAILURE state and i'm curious what the last event that the run worker emits looks like

python_modules/dagster/dagster_tests/core_tests/instance_tests/test_instance.py
```diff
@@ -348,7 +348,7 @@ def launch_run(self, context):
     def join(self, timeout=30):
         raise NotImplementedError()
 
-    def terminate(self, run_id):
+    def terminate(self, run_id, message=None):
```
gibsondan, line 351: sounds like this is not intentional

python_modules/dagster/dagster_tests/daemon_tests/test_monitoring_daemon.py
```diff
@@ -151,42 +180,187 @@ def test_monitor_starting(instance: DagsterInstance, logger: Logger):
...
+            assert run
+            assert run.status == DagsterRunStatus.FAILURE
+
+            run_failure_events = instance.all_logs(
+                too_long_record.dagster_run.run_id, of_type=DagsterEventType.RUN_FAILURE
+            )
+            assert len(run_failure_events) == 1
+            event = run_failure_events[0].dagster_event
+            assert event
+            assert event.message == "Exceeded maximum runtime of 500.0 seconds."
```
gibsondan, line 305: we can probably int() it, i don't think we need this to multiple decimal places

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -471,6 +471,9 @@ def terminate(self, run_id):
             return False
 
         self.ecs.stop_task(task=tags.arn, cluster=tags.cluster)
+
+        run = self._instance.get_run_by_id(run_id)
+        self._instance.report_run_canceling(run, message=message)
         return True
```
gibsondan, line 477: rm

gibsondan (approved):

python_modules/dagster/dagster/_daemon/monitoring/monitoring_daemon.py
dpeng817: Here's an example of the following job going through the termination flow:
```python
import time

from dagster import job, op, resource


@resource
def some_resource_with_spindown(context):
    try:
        yield "foo"
    finally:
        context.log.info("Spinning down resource")


@op(required_resource_keys={"foo"})
def some_op():
    time.sleep(60)
    return 1



@job(resource_defs={"foo": some_resource_with_spindown})
def the_job():
    some_op()
    some_op()
```
I induced this by enabling run monitoring with a 5 second polling interval, and setting the tag with 5 seconds as the maximum time. The following was the event stream post cancellation:
Screen Shot 2023-04-05 at 1 50 53 PM
Basically, you end up getting the actual step-related events + spindown stuff after the run is marked as a failure.

dpeng817: Ended up changing this to avoid the two `run_failure` events above; now the second event doesn't produce a run failure, but instead an engine event:
Screen Shot 2023-04-05 at 2 15 08 PM

dpeng817:
To illustrate how this works in the asset case, take the following asset job:
```python
import time

from dagster import Definitions, asset, define_asset_job, resource
from dagster._core.storage.tags import MAX_RUNTIME_SECONDS_TAG


@resource
def some_resource_with_spindown(context):
    try:
        yield "foo"
    finally:
        context.log.info("Spinning down resource")

@asset
def normal():
    return 1

@asset(required_resource_keys={"foo"})
def sleepy_asset(normal):
    time.sleep(60)
    return 1


@asset(required_resource_keys={"foo"})
def other_sleepy_asset(normal):
    time.sleep(60)
    return 1

@asset
def downstream_asset(sleepy_asset, other_sleepy_asset):
    return 1

defs = Definitions([normal, sleepy_asset, other_sleepy_asset, downstream_asset], jobs=[define_asset_job("all_assets", selection=[normal, sleepy_asset, other_sleepy_asset, downstream_asset], tags={MAX_RUNTIME_SECONDS_TAG: 30})], resources={"foo": some_resource_with_spindown})
```
When this job runs, asset normal succeeds, while assets `sleepy` and `other_sleepy` are interrupted during execution. In the asset graph, `normal` is marked as succeeded, while `sleepy`, `other_sleepy`, and `downstream` are all marked as failed:
Screen Shot 2023-04-06 at 11 05 06 AM

Issue #12885, opened 2023-03-10
avoid reconciling any asset partitions that are targeted by active backfills


sryza:

sryza:
internal implementation note: here's how we can find the list of asset partitions that are targeted by active backfills:
```python
from dagster._core.definitions.asset_graph_subset import AssetGraphSubset


def _get_active_backfill_targets(
    instance: DagsterInstance, asset_graph: AssetGraph
) -> AssetGraphSubset:
    from dagster._core.execution.asset_backfill import AssetBackfillData
    from dagster._core.execution.backfill import BulkActionStatus

    asset_backfills = [
        backfill
        for backfill in instance.get_backfills(status=BulkActionStatus.REQUESTED)
        if backfill.is_asset_backfill
    ]

    result = AssetGraphSubset(asset_graph)
    for asset_backfill in asset_backfills:
        if asset_backfill.serialized_asset_backfill_data is None:
            check.failed("Asset backfill missing serialized_asset_backfill_data")

        asset_backfill_data = AssetBackfillData.from_serialized(
            asset_backfill.serialized_asset_backfill_data, asset_graph
        )

        result |= asset_backfill_data.target_subset

    return result
```

PR #12926, opened 2023-03-13
[asset-reconciliation] Do not reconcile partitions involved in active asset backfills

OwenKephart:
## Summary & Motivation

resolves: https://github.com/dagster-io/dagster/issues/12885

## How I Tested These Changes

added a property to the AssetReconciliationScenario object which makes it possible to simulate the presence of asset backfills.

confirmed that two of the tests ("...with_backfill_intersecting" and "...with_backfill_superceding") failed before the changes, and now succeed. "..._with_backfill_independent" passes regardless of the changes by design.


sryza (approved):
Nice!

PR #12927, opened 2023-03-13
Fix num static partitions to count unique partitions

clairelin135:
User reported an issue where partition counts were incorrect: https://elementl-workspace.slack.com/archives/C047XLJ0BRT/p1678716536609149

This was caused by duplicate static partition keys. In the next major release, we will raise an error for this. In the meantime, we'll change our partition counts logic to find the unique static partition keys.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/partition.py
+++ b/python_modules/dagster/dagster/_core/definitions/partition.py
@@ -3,7 +3,6 @@ import hashlib
 import inspect
 import json
 from abc import ABC, abstractmethod
-from collections import Counter
 from datetime import (
     datetime,
     time,
@@ -402,12 +401,7 @@ class StaticPartitionsDefinition(
     def __init__(self, partition_keys: Sequence[str]):
         check.sequence_param(partition_keys, "partition_keys", of_type=str)
 
-        if len(partition_keys) != len(set(partition_keys)):
-            raise DagsterInvalidDefinitionError(
-                "Partition keys must be unique. Found duplicate keys: "
-                f"{[k for k, v in Counter(partition_keys).items() if v > 1]}"
-            )
-
+        # TODO 1.3.0 enforce that partition keys are unique
         raise_error_on_invalid_partition_key_substring(partition_keys)
 
         self._partitions = [Partition(key) for key in partition_keys]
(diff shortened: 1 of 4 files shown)


clairelin135:
This technically is a breaking change, but I think it can be classified more as a bugfix since this behavior is unsupported

sryza (commented):
Is there some situation we need to worry about where old user code sends bad StaticPartitionsDefinitions over the wire to a host process that has this check and disaster ensues?

clairelin135:
@sryza one possible situation that might happen is some external static partitions definition object contains duplicate keys and then graphQL requests fail when attempting to build a static partitions definition. The likely consequence is that the user won't be able to load the asset graph without fixing the issue and upgrading their user code.

I remember we still let the change to make escaped characters invalid go out in a dot release. I'm happy to wait until the next major release if we think it's very likely for cloud users to end up in this situation.

johannkm:
It’d be a little tricky but I think we could test in advance if this would
affect cloud users and give a heads up. Probably will come in handy to
write a script for downloading the caches anyway. Happy to help

clairelin135:
On second thought, I think it's probably a non-trivial number of users who might have duplicate static partition keys, given that someone reported this error. It's probably best to wait until the next major release to enforce uniqueness. For now, this PR has been amended to only count the number of unique partitions instead to display the correct information in Dagit.

python_modules/dagster/dagster/_core/definitions/partition.py
```diff
@@ -423,6 +423,13 @@ def __eq__(self, other) -> bool:
     def __repr__(self) -> str:
         return f"{type(self).__name__}(partition_keys={[p.name for p in self._partitions]})"
 
+    def get_num_partitions(
+        self,
+        current_time: Optional[datetime] = None,
+        dynamic_partitions_store: Optional[DynamicPartitionsStore] = None,
+    ) -> int:
+        return len(set(self.get_partition_keys(current_time, dynamic_partitions_store)))
```
johannkm, line 431: could add a comment about why we need a set

johannkm (approved):
Does this leave the default implementation unused? could remove

clairelin135:
The default implementation is still used for multipartitions

clairelin135:
Hm.. the counts could also be off for the multipartitioned case. I'll add a similar function for the multipartitioned case to return unique counts, but I still think we should retain the original default function in case if anybody has a custom partitions definition.

PR #12933, opened 2023-03-14
[pythonic config] Fix using RunConfig with optional config fields

benpankow:
## Summary

Updates the Pythonic config class to config dict conversion code to drop optional entries with `None` values, to satisfy Dagster's internal type checking.

Previously, this lead to errors with optional fields using `RunConfig`:

```python
class AssetConfig(Config):
    baz: Optional[str] = None


@asset
def foo(config: AssetConfig) -> str:
    ...

# would error, since underlying Dagster config is passed `{"baz": None}` rather than `{}`
materialize([foo], run_config=RunConfig({"foo": AssetConfig()}))
```

Surfaced here: https://dagster.slack.com/archives/C01U5LFUZJS/p1678621110892759

Also adds some new documentation to describe how to set defaults & optional values.

## Test Plan

new unit tests

### DIFF
--- a/docs/content/guides/dagster/pythonic-config.mdx
+++ b/docs/content/guides/dagster/pythonic-config.mdx
@@ -126,6 +126,31 @@ class MyMetadataConfig(Config):
 MyMetadataConfig(person_name="Alice", age=200)
 ```
 
+### Defaults and optional config fields
+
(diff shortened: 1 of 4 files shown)


benpankow:
* master
  * **PR #12933** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12933" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12933?utm_source=stack-comment).

schrockn (approved):

PR #12944, opened 2023-03-14
Fix valid start time check for time partitions

clairelin135:
Reported by a user: https://dagster.slack.com/archives/C01U954MEER/p1678304642514719

Fixes an error where the first partition in the partition health bar always displays as unmaterialized despite a materialization existing for this partition. This error occurs when the running OS timezone is ahead of the partitions def timezone (by default UTC).

The cause of this issue was a partition key validation check that relied on a `datetime.strptime` object which is not timezone aware (so defaults to the running OS timezone). Because the external time window partitions def creates a start time object that is timezone aware, certain valid partition keys would be considered invalid as they were set to an earlier timezone and thus were considered to be "before" the valid start time.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
@@ -109,8 +109,11 @@ class TimeWindowPartitionsDefinition(
         day_offset: Optional[int] = None,
         cron_schedule: Optional[str] = None,
     ):
+        check.opt_str_param(timezone, "timezone")
+        timezone = timezone or "UTC"
+
         if isinstance(start, datetime):
-            start_dt = start
+            start_dt = pendulum.instance(start, tz=timezone)
         else:
             start_dt = pendulum.instance(datetime.strptime(start, fmt), tz=timezone)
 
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -679,8 +679,10 @@ def empty_subset(self) -> "PartitionsSubset":
 
     def is_valid_partition_key(self, partition_key: str) -> bool:
         try:
-            time_obj = datetime.strptime(partition_key, self.fmt)
-            return time_obj.timestamp() >= self.start.timestamp()
+            partition_time = pendulum.instance(
+                datetime.strptime(partition_key, self.fmt), tz=self.timezone
+            )
+            return partition_time.timestamp() >= self.start.timestamp()
```
OwenKephart, line 685: now that everything is tz-aware, can we just do partition_time >= self.start?

OwenKephart (approved):
see comment, otherwise looks good

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
clairelin135: yep! makes sense

PR #12946, opened 2023-03-14
Resolve transitive resource deps for schedules and sensors

benpankow:
## Summary

Ensures that when a resource requirement is specified for a schedule or sensor that we resolve all transitive resource deps when building the set of resources to pass to the schedule/sensor fn, e.g.

```python
@resource
def the_inner() -> str:
    return "oo"


@resource(required_resource_keys={"the_inner"})
def the_outer(init_context) -> str:
    return "f" + init_context.resources.the_inner


@sensor(job=the_job, required_resource_keys={"the_outer"})
def sensor_resource_deps(context):
    return RunRequest(context.resources.the_outer, run_config={}, tags={})
```

## Test Plan

New unit tests

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/sensor_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/sensor_definition.py
@@ -782,22 +782,6 @@ def build_sensor_context(
     check.opt_str_param(cursor, "cursor")
     check.opt_str_param(repository_name, "repository_name")
 
-    # Determine the set of resources to pass by
-    # 1. Trying to pull the required resource keys from the repository, if available
-    # 2. Using the resources explicitly passed in
-    required_resource_keys = None
-    if repository_def and sensor_name:
-        sensor_def = repository_def.get_sensor_def(sensor_name)
-        required_resource_keys = sensor_def.required_resource_keys
-
-    top_level_resources = repository_def.get_top_level_resources() if repository_def else {}
-    all_resources = {**top_level_resources, **(resources or {})}
-
-    resources_to_build = (
-        {k: v for k, v in all_resources.items() if k in required_resource_keys}
-        if required_resource_keys
-        else all_resources
-    )
     return SensorEvaluationContext(
         instance_ref=None,
         last_completion_time=None,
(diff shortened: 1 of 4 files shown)


benpankow:
* master
  * **PR #12946** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12946" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12946?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -782,22 +782,6 @@ def build_sensor_context(
     check.opt_str_param(cursor, "cursor")
     check.opt_str_param(repository_name, "repository_name")
 
-    # Determine the set of resources to pass by
```
benpankow, line 785: This codepath is not used right now, removing so that the resources-to-build logic is only computed in one place (in the GRPC server)

schrockn: I'm still a bit confused here.

Can you provide a concrete example of a code example that would have worked before this change but that will no longer work? Is it just a case where we pass in a partial set of resources and the rest are filled in from the repository level?

schrockn (changes requested):
Code looks good, I just have the one question about what capabilities we are removing.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
benpankow:
```python
build_sensor_context(
    instance=instance,
    sensor_name="my_sensor_name"
)
```

This would (1) pull all resources from the instance and (2) figure out which to bind based on the sensor name, by looking up the sensor in the instance. However, we don't use this pattern internally & I think the number of users who would use it is pretty low (relative to `build_sensor_context(..., resources={...})`). I ended up implementing this pattern pretty early on as an alternative to `resources={...}` but found it not very useful (& couldn't implement a corresponding behavior for schedules anyway since schedules don't take the optional `instance`, `schedule_name` inputs).

schrockn: Cool thanks for the explanation.

schrockn (approved):

PR #12948, opened 2023-03-14
[dagit] Use pure-asset backfills instead of hidden asset job backfills

bengotow:
## Summary & Motivation

This change is based on this slack thread:
https://elementl-workspace.slack.com/archives/C03CA4TVCAW/p1678813493408879

The materialize button's behavior is as follows:

- on the asset job view, single runs and backfills should use the visible asset job
- on the asset group view, single runs should use the hidden asset jobs
- on the asset group view, all backfills should be pure asset backfills

Previously, we were using the hidden asset job for all backfills that did not REQUIRE the pure asset backfill machinery. Following this update, hidden asset jobs are only used for single runs.

## How I Tested These Changes

I manually tested and verified the GraphQL result of:

- launching a single partitioned asset from an asset group and from an asset job
- launching two same-partitioned assets from an asset group and from an asset job
- launching two partition-mapped assets from an asset group and from an asset job
- launching two unpartitioned assets from an asset group and from an asset job

I'm working on more automated tests for this in a small follow-up PR, but we may want to ship this tonight.

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
+++ b/js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
@@ -25,7 +25,11 @@ import {showCustomAlert} from '../app/CustomAlertProvider';
 import {PipelineRunTag} from '../app/ExecutionSessionStorage';
 import {usePermissionsForLocation} from '../app/Permissions';
 import {PythonErrorInfo} from '../app/PythonErrorInfo';
-import {displayNameForAssetKey, itemWithAssetKey} from '../asset-graph/Utils';
+import {
+  displayNameForAssetKey,
+  isHiddenAssetGroupJob,
+  itemWithAssetKey,
+} from '../asset-graph/Utils';
 import {AssetKey} from '../assets/types';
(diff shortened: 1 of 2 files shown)


js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
```diff
@@ -412,13 +412,20 @@ const LaunchAssetChoosePartitionsDialogBody: React.FC<Props> = ({
...
         <ToggleableSection
-          title={<Subheading>Partition selection</Subheading>}
+          title={
+            <Box flex={{direction: 'row', justifyContent: 'space-between'}}>
+              <Subheading>Partition selection</Subheading>
+              <span>
+                {keysFiltered.length} {keysFiltered.length === 1 ? 'partition' : 'partitions'}
+              </span>
+            </Box>
+          }
```
bengotow, line 428: Noticed this was missing @salazarm, I think maybe we forgot to put the count in?

salazarm: Oh I had taken the count after talking adhoc with @braunjj but I'm fine with adding it back

salazarm (approved):

Issue #12905, opened 2023-03-10
Add support for multi-to-multi partition mapping with TimeWindowPartitions


ShahBinoy:
## Original Slack Thread
https://dagster.slack.com/archives/C01U954MEER/p1677679275163299

### What's the use case?

I want to achieve a Fan-In functionality using Partitions and Assets ?
I want to scan S3 files using Daily and Hourly Partition Scheme, but then Merge them on a Daily Partition basis.
I need to leverage multiple partitioning schemes to process data and combine/merge according to day granularity

```python
user_id_ranges: list[str] = ["0..4999", "5000..9999"]
user_partitions = StaticPartitionsDefinition(partition_keys=user_id_ranges)
daily_partition = DailyPartitionsDefinition(start_date="2023-02-10")
hourly_partition = HourlyPartitionsDefinition(start_date=datetime(2023, 2, 10))

def dynamic_customers():
    ...


@asset(
    partitions_def=MultiPartitionsDefinition(partitions_defs={
        "users": user_partitions,
        "hour": hourly_partition})
)
def hourly_dump(context):
    val_output = context.asset_partition_key_for_output()
    context.log.info(f"Processing RAW Hourly Data {val_output}")
    return val_output


@asset(
    partitions_def=MultiPartitionsDefinition(partitions_defs={
        "users": user_partitions,
        "day": daily_partition})
)
def compressed_data(context, hourly_dump):
    val_output = context.asset_partition_key_for_output()
    context.log.info(f"Received RAW Data f{hourly_dump}")
    context.log.info(f"Processing Hourly Data {val_output}")

    return val_output

```

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

clairelin135:
This is now enabled via the `MultiPartitionMapping` object: https://github.com/dagster-io/dagster/pull/12950

PR #12950, opened 2023-03-14
MultiPartitionMapping

clairelin135:
Requested by multiple users, primarily for the use case of defining multipartitions dependencies for different time intervals: https://github.com/dagster-io/dagster/issues/12905

Introduces a new `MultiPartitionsMapping` partition mapping that allows for defining partition mapping relationships per-dimension. For example, the below partition mapping would map `a|2023-01-01` in `weekly_abc` would map to `1|2023-01-01`... `1|2023-01-07`.
```
weekly_abc = MultiPartitionsDefinition(
    {
        "abc": StaticPartitionsDefinition(["a", "b", "c"]),
        "weekly": WeeklyPartitionsDefinition("2023-01-01"),
    }
)
daily_123 = MultiPartitionsDefinition(
    {
        "123": StaticPartitionsDefinition(["1", "2", "3"]),
        "daily": DailyPartitionsDefinition("2023-01-01"),
    }
)

MultiPartitionsMapping(
    {
      "abc": DimensionMapping(
          downstream_dimension_name="123",
          partition_mapping=StaticPartitionMapping({"a": "1", "b": "2", "c": "3"}),
      ),
      "weekly": DimensionMapping(
            downstream_dimension_name="daily",
            partition_mapping=TimeWindowPartitionMapping(),
       )
    }
)
```
The `MultiPartitionsMapping` object checks that the selected dimensions in the dimension mapping objects are existent, relying on the partition mapping per-dimension to perform type checks and calculate the mapped partitions. Each dimension in the upstream partitions definition must be mapped 1:1 with a dimension in the downstream partitions definition.

In the situation where a user only wants to define a dependency relationship between one dimension of the upstream and downstream multipartitions defs, they could do something like the below example. In this case, `a|2023-01-01` would map to `1|2023-01-01`, `2|2023-01-01`, and `3|2023-01-01`.
```
MultiPartitionsDefinition(
    {
        "abc": StaticPartitionsDefinition(["a", "b", "c"]),
        "daily": DailyPartitionsDefinition("2023-01-01"),
    }
),
MultiPartitionsDefinition(
    {
        "123": StaticPartitionsDefinition(["1", "2", "3"]),
        "daily": DailyPartitionsDefinition("2023-01-01"),
    }
),
MultiPartitionsMapping(
    {
        "abc": DimensionMapping(
            "123",
            AllPartitionMapping(),
        ),
        "daily": DimensionMapping(
            "daily",
            IdentityPartitionMapping(),
        ),
    }
)
```
I'm very open to API changes. My intent here was to make this partition mapping generic, so it can handle any number of dimensions in the future.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
@@ -224,6 +224,12 @@ class MultiPartitionsDefinition(PartitionsDefinition):
     def partitions_defs(self) -> Sequence[PartitionDimensionDefinition]:
         return self._partitions_defs
 
+    def get_partitions_def_for_dimension(self, dimension_name: str) -> PartitionsDefinition:
+        for dim_def in self._partitions_defs:
+            if dim_def.name == dimension_name:
+                return dim_def.partitions_def
+        check.failed(f"Invalid dimension name {dimension_name}")
+
     def get_partition(
         self,
         partition_key: str,
--- a/python_modules/dagster/dagster/_core/definitions/partition_mapping.py
+++ b/python_modules/dagster/dagster/_core/definitions/partition_mapping.py
@@ -1,9 +1,12 @@
 import collections.abc
+import itertools
 import warnings
 from abc import ABC, abstractmethod
 from collections import defaultdict
 from typing import (
     Collection,
+    Dict,
+    List,
     Mapping,
     NamedTuple,
     Optional,
@@ -12,11 +15,8 @@ from typing import (
     Tuple,
     Type,
     Union,
-    Dict,
     cast,
-    List,
 )
-import itertools
 
 import dagster._check as check
 from dagster._annotations import PublicAttr, experimental, public
@@ -31,6 +31,7 @@ from dagster._core.definitions.partition import (
 )
 from dagster._core.definitions.partition_key_range import PartitionKeyRange
 from dagster._core.definitions.time_window_partitions import TimeWindowPartitionsDefinition
+from dagster._core.errors import DagsterInvalidDefinitionError
 from dagster._core.instance import DynamicPartitionsStore
 from dagster._serdes import whitelist_for_serdes
 from dagster._utils.backcompat import ExperimentalWarning
@@ -564,6 +565,17 @@ class DimensionMapping(
         ],
     )
 ):
+    """
+    Defines an explicit correspondence between one dimension of a MultiPartitionsDefinition
+    and one dimension of another MultiPartitionsDefinition.
+
+    Args:
+        upstream_dimension_name (str): The name of the dimension in the upstream MultiPartitionsDefinition.
+        downstream_dimension_name (str): The name of the dimension in the downstream MultiPartitionsDefinition.
+        partition_mapping (PartitionMapping): The partition mapping object used to calculate
+            the downstream dimension partitions from the upstream dimension partitions and vice versa.
+    """
+
     def __new__(
         cls,
         upstream_dimension_name: str,
@@ -591,7 +603,19 @@ class MultiPartitionsMapping(
     NamedTuple("_MultiPartitionMapping", [("dimension_mappings", Sequence[DimensionMapping])]),
 ):
     """
-    TODO add docstrings
+    Defines a correspondence between two MultiPartitionsDefinitions.
+
+    Accepts a list of DimensionMappings, each of which defines an explicit correspondence
+    between one dimension of the upstream MultiPartitionsDefinition and one dimension of the
+    downstream MultiPartitionsDefinition. Each dimension of an upstream MultiPartitionsDefinition
+    must be explicitly mapped to a dimension of the downstream MultiPartitionsDefinition, and vice
+    versa.
+
+
+    Args:
+        dimension_mappings (Sequence[DimensionMapping]): A list of DimensionMappings, each of which
+            defines an explicit correspondence between one dimension of the upstream MultiPartitionsDefinition
+            and one dimension of the downstream MultiPartitionsDefinition.
     """
 
     def __new__(cls, dimension_mappings: Sequence[DimensionMapping]):
@@ -646,27 +670,43 @@ class MultiPartitionsMapping(
                 " multi-partitioned"
             )
 
-        upstream_dimension_names = set(
-            [
-                dim.name
-                for dim in cast(MultiPartitionsDefinition, upstream_partitions_def).partitions_defs
-            ]
-        )
-        downstream_dimension_names = set(
-            [
-                dim.name
-                for dim in cast(
-                    MultiPartitionsDefinition, downstream_partitions_def
-                ).partitions_defs
-            ]
-        )
+        upstream_dimension_names = [
+            dim.name
+            for dim in cast(MultiPartitionsDefinition, upstream_partitions_def).partitions_defs
+        ]
+        downstream_dimension_names = [
+            dim.name
+            for dim in cast(MultiPartitionsDefinition, downstream_partitions_def).partitions_defs
+        ]
+        if len(upstream_dimension_names) != len(set(upstream_dimension_names)) or len(
+            downstream_dimension_names
+        ) != len(set(downstream_dimension_names)):
+            raise DagsterInvalidDefinitionError(
+                "A dimension name cannot be specified in more than one DimensionMapping in a"
+                " MultiPartitionsMapping."
+            )
 
-        if len(upstream_dimension_names) != len(downstream_dimension_names):
-            check.failed(
-                "The number of dimensions in the upstream and downstream partitions defs must match"
+        if len(upstream_dimension_names) != len(
+            cast(MultiPartitionsDefinition, upstream_partitions_def).partitions_defs
+        ) or len(downstream_dimension_names) != len(
+            cast(MultiPartitionsDefinition, downstream_partitions_def).partitions_defs
+        ):
+            raise DagsterInvalidDefinitionError(
+                "Each dimension in the upstream and downstream partitions defs must be mapped to "
+                "exactly one dimension in the other partitions def."
             )
 
         for dimension_mapping in self.dimension_mappings:
+            if dimension_mapping.upstream_dimension_name not in upstream_dimension_names:
+                check.failed(
+                    "Dimension mapping has an upstream dimension name that is not in the upstream "
+                    "partitions def"
+                )
+            if dimension_mapping.downstream_dimension_name not in downstream_dimension_names:
+                check.failed(
+                    "Dimension mapping has a downstream dimension name that is not in the"
+                    " downstream partitions def"
+                )
             upstream_dimension_names.remove(dimension_mapping.upstream_dimension_name)
             downstream_dimension_names.remove(dimension_mapping.downstream_dimension_name)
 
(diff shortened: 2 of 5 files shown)


python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -551,6 +555,338 @@ def get_downstream_partitions_for_partitions(
...
+    downstream MultiPartitionsDefinition. Each dimension of an upstream MultiPartitionsDefinition
+    must be explicitly mapped to a dimension of the downstream MultiPartitionsDefinition, and vice
+    versa.
+
+
+    Args:
+        dimension_mappings (Sequence[DimensionMapping]): A list of DimensionMappings, each of which
+            defines an explicit correspondence between one dimension of the upstream MultiPartitionsDefinition
+            and one dimension of the downstream MultiPartitionsDefinition.
+    """
```
sryza, line 619: I think it would be worth including the example you put together in the commit as an example in this docstring.

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -551,6 +555,338 @@ def get_downstream_partitions_for_partitions(
...
+            ),
+            partition_mapping=check.inst_param(
+                partition_mapping, "partition_mapping", PartitionMapping
+            ),
+        )
+
+
+@experimental
+@whitelist_for_serdes
+class MultiPartitionsMapping(
```
sryza, line 601: Since this is a subclass of `PartitionMapping`, I would expect it to end in `PartitionMapping` like the other subclasses (instead of `PartitionsMapping`). Let me know if you think otherwise.

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -551,6 +555,338 @@ def get_downstream_partitions_for_partitions(
...
+            ),
+        )
+
+    def _get_dependency_keys(
+        self,
+        a_partitions_def: MultiPartitionsDefinition,
+        a_partition_keys: Sequence[str],
+        b_partitions_def: MultiPartitionsDefinition,
+        a_upstream_of_b: bool,
+    ):
```
sryza, line 732: Return type plz

sryza (commented):
In the cases where the upstream and downstream assets have the same MultiPartitionsDefinition, it seems like this could be simpler - e.g. a dictionary that maps dimension names to PartitionMappings.  I suspect these are the most common cases, so worth providing a simpler API for them?

In the other direction, it also seems like we'll likely ultimately want to supply something that can handle mappings between multi-partitioned assets where the upstream and downstream assets have different numbers of dimensions.

Thoughts on how we'd want to handle those? Asking here because it's worth considering which of these cases should be handled under the same PartitionMapping, and how their names should affect the names of the other ones.

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
clairelin135: I'm fine with naming it `MultiPartitionMapping`

sryza:
> Upstream and downstream multipartitions are both the same. They should be able to use the identity partition mapping instead.

What about if they're the same but the user wants to set a custom PartitionMapping for one of the dimensions? E.g. they both have a static dimension and a time dimension, but they want to set a `start_offset` on the mapping between the two time dimensions?

> As long as one mapping between an upstream and downstream dimension exists, all of the remaining downstream dimension partitions basically depend on all of the remaining upstream dimension partitions.

Extending this to its logical conclusion, do mappings need to be provided? Would it be sensible to say that, if no mappings are provided, we essentially do an AllPartitionMapping?

clairelin135:
> What about if they're the same but the user wants to set a custom PartitionMapping for one of the dimensions? E.g. they both have a static dimension and a time dimension, but they want to set a start_offset on the mapping between the two time dimensions?

They'd still have to define that `TimeWindowPartitionMapping` between those two time dimensions. But the other dimensions we can infer--i.e. if the partitions defs are the same, then we automatically assume an identity partition mapping. If partitions defs are not the same, we assume an `AllPartitionMapping`.

I think this is something that is possible for us to implement now, but I think only a very small subset of users want this partition mapping. Those broad use cases where auto inference is helpful likely won't come into play until we support multiple dimensions in `MultiPartitionsDefinition`s. We can always loosen restrictions and do more inference in the future, but in that case I think users would still be defining the class in the same fashion, i.e.:
```
abc = StaticPartitionsDefinition(["a", "b", "c"])
weekly = WeeklyPartitionsDefinition("2023-01-01")
weekly_abc = MultiPartitionsDefinition(
    {
        "abc": abc,
        "weekly": weekly,
    }
)
downstream_weekly_abc = MultiPartitionsDefinition(
    {
        "abc": abc,
        "weekly": weekly,
    }
)

MultiPartitionsMapping(
    [
       DimensionMapping(
            "weekly", "weekly, TimeWindowPartitionMapping(start_offset=1),
        ),
    ]
)
```
so we could just loosen these restrictions without breaking existing users. Since the use case is currently so narrow, I think the current implementation of this class as it stands should meet the needs of the users who are asking for it.

sryza:
Part of the reason I'm asking these questions is that accepting a list of dimension mappings feels a little sketchy, because it makes it possible to include the same dimension twice.  The more constrained we can make the API, the easier it will be for users to understand how to use it.

If we were to instead accept a dictionary that's keyed by dimension name, then it would constrain it so that each key can only have one value. However, I'm trying to parse whether this makes sense by probing the wider set of cases. And to understand: would we want it to be keyed by the upstream dimension name or the downstream dimension name? Do either of these make sense? Do they make sense equally?

Thoughts?

clairelin135:
Dictionary keyed by dimension name makes sense to me.

I can think of situations where multiple upstream dimensions would map to a downstream dimension, i.e:
- Upstream asset representing student test scores per school `MultiPartitionsDef(student_id, test_num, school_num)`
- Downstream asset representing average scores for 10 random students per school `MultiPartitionsDef(school_10_students_sample, avg_test_score)`

But I can't think of situations where an upstream dimension would be mapped to multiple downstream dimensions.

This makes me lean toward keying by upstream dimension name, so that a downstream dimension can depend on multiple upstream dimensions. I think this also has a nice consistency with `StaticPartitionMapping`, which keys by upstream partition keys.

sryza:
> This makes me lean toward keying by upstream dimension name, so that a downstream dimension can depend on multiple upstream dimensions. I think this also has a nice consistency with StaticPartitionMapping, which keys by upstream partition keys.

Cool, that makes sense to me

python_modules/dagster/dagster_tests/asset_defs_tests/partition_mapping_tests/test_multipartition_partition_mapping.py
```diff
@@ -203,3 +218,376 @@ def test_error_thrown_when_no_partition_dimension_name_provided():
             ),
             multipartitions_def,
         )
+
+
+upstream_and_downstream_tests = [
```
sryza, line 223: I like this big set of tests

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -551,6 +555,309 @@ def get_downstream_partitions_for_partitions(
...
+            # b_partitions_def
+            a_dim_to_dependent_b_dim = {
+                upstream_dim: (
+                    dimension_mapping.downstream_dimension_name,
+                    dimension_mapping,
+                )
+                for upstream_dim, dimension_mapping in self.downstream_mappings_by_upstream_dimension.items()
+            }
+
+            for dimension_name, keys in a_partition_keys_by_dimension.items():
```
sryza, line 786: Nitpick: maybe `dimension_name` -> `a_dim_name` for consistency with `b_dim_name`?

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -551,6 +555,309 @@ def get_downstream_partitions_for_partitions(
...
+
+        return b_partitions_def.empty_subset().with_partition_keys(b_partition_keys)
+
+    def get_upstream_partitions_for_partitions(
+        self,
+        downstream_partitions_subset: Optional[PartitionsSubset],
+        upstream_partitions_def: PartitionsDefinition,
+        dynamic_partitions_store: Optional[DynamicPartitionsStore] = None,
+    ) -> PartitionsSubset:
+        check.opt_inst_param(
```
sryza, line 876: Because this is an internal method, I think we can rely on the type annotations instead of these checks

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -551,6 +555,309 @@ def get_downstream_partitions_for_partitions(
...
+
+        check.invariant(
+            not upstream_dimension_names,
+            "Upstream partitions def has dimensions that are not mapped: {}".format(
+                upstream_dimension_names
+            ),
+        )
+        check.invariant(
+            not downstream_dimension_names,
+            "Downstream partitions def has dimensions that are not mapped: {}".format(
```
sryza: f-strings instead of format plz

sryza (commented):
This is looking close

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -551,6 +555,309 @@ def get_downstream_partitions_for_partitions(
             )
 
 
+class DimensionMapping(
```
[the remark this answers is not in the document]
clairelin135: Makes sense, though I think `DimensionPartitionMapping` is clearer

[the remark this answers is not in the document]
sryza: Oo you are right

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -551,6 +555,309 @@ def get_downstream_partitions_for_partitions(
...
+            ("downstream_dimension_name", str),
+            ("partition_mapping", PartitionMapping),
+        ],
+    )
+):
+    """
+    TODO update docstring
+
+    Args:
+        downstream_dimension_name (str): The name of the dimension in the downstream MultiPartitionsDefinition.
```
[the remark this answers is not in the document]
clairelin135: That makes sense to me

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -551,6 +555,309 @@ def get_downstream_partitions_for_partitions(
...
+                    dimension_mapping,
+                ) = a_dim_to_dependency_b_dim[dimension_name]
+                a_dimension_partitions_def = a_dimension_partitions_defs_by_name[dimension_name]
+                b_dimension_partitions_def = b_dimension_partitions_def_by_name[b_dim_name]
+                for key in keys:
+                    dep_b_keys_by_a_dim_and_key[dimension_name][key] = list(
+                        dimension_mapping.partition_mapping.get_upstream_partitions_for_partitions(
+                            a_dimension_partitions_def.empty_subset().with_partition_keys([key]),
+                            b_dimension_partitions_def,
+                            # Add dynamic partitions store
```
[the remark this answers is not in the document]
clairelin135: yep, resolved now--needed to add the dynamic partition store to make this class work with dynamic partitions defs

clairelin135:
Thanks for the feedback @sryza! I updated per your feedback, I think this is ready for another look

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
```diff
@@ -224,6 +224,12 @@ def partition_dimension_names(self) -> List[str]:
     def partitions_defs(self) -> Sequence[PartitionDimensionDefinition]:
         return self._partitions_defs
 
+    def get_partitions_def_for_dimension(self, dimension_name: str) -> PartitionsDefinition:
+        for dim_def in self._partitions_defs:
+            if dim_def.name == dimension_name:
+                return dim_def.partitions_def
+        check.failed(f"Invalid dimension name {dimension_name}")
```
sryza, line 231:
```suggestion
        check.failed(f"Invalid dimension name '{dimension_name}'")
```

PR #12963, opened 2023-03-15
[Dynamic partition requests 3/3] Dynamic partitions requests

clairelin135:
Part 1: https://github.com/dagster-io/dagster/pull/13011
Part 2: https://github.com/dagster-io/dagster/pull/12962

This PR introduces the `AddDynamicPartitionsRequest` and `DeleteDynamicPartitionsRequest` objects, to be added to a `SensorResult`. When partitioned run requests are resolved with tags and config, we validate that the partition will exist at the end of the tick as it could be added or deleted by a dynamic partitions request.

In the daemon, we mutate the partitions definitions and write logs for the changes. In the future, we can persist the mutations that occurred in the tick, allowing Dagit to populate logs of changes per tick.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/partition.py
+++ b/python_modules/dagster/dagster/_core/definitions/partition.py
@@ -714,19 +714,12 @@ class DynamicPartitionsDefinition(
             partitions_def_name=self._validated_name(), partition_key=partition_key
         )
 
-    def get_pending_partition(self, partition_key: str) -> Partition:
-        return Partition(partition_key, partition_key)
-
-    def request_for_adding_partitions(
-        self, partition_keys: Sequence[str]
-    ) -> AddDynamicPartitionsRequest:
+    def build_add_request(self, partition_keys: Sequence[str]) -> AddDynamicPartitionsRequest:
         check.sequence_param(partition_keys, "partition_keys", of_type=str)
         validated_name = self._validated_name()
         return AddDynamicPartitionsRequest(validated_name, partition_keys)
 
-    def request_for_deleting_partitions(
-        self, partition_keys: Sequence[str]
-    ) -> DeleteDynamicPartitionsRequest:
+    def build_delete_request(self, partition_keys: Sequence[str]) -> DeleteDynamicPartitionsRequest:
         check.sequence_param(partition_keys, "partition_keys", of_type=str)
         validated_name = self._validated_name()
         return DeleteDynamicPartitionsRequest(validated_name, partition_keys)
--- a/python_modules/dagster/dagster/_core/definitions/run_request.py
+++ b/python_modules/dagster/dagster/_core/definitions/run_request.py
@@ -171,6 +171,7 @@ class RunRequest(
         from dagster._core.definitions.job_definition import JobDefinition
         from dagster._core.definitions.partition import (
             DynamicPartitionsDefinition,
+            Partition,
             PartitionedConfig,
             PartitionsDefinition,
         )
@@ -233,7 +234,7 @@ class RunRequest(
                     " applied, it does not exist in the set of valid partition keys."
                 )
 
-            partition = partitions_def.get_pending_partition(self.partition_key)
+            partition = Partition(self.partition_key, self.partition_key)
 
         else:
             # Relies on the partitions def to throw an error if the partition does not exist
--- a/python_modules/dagster/dagster_tests/daemon_sensor_tests/test_sensor_run.py
+++ b/python_modules/dagster/dagster_tests/daemon_sensor_tests/test_sensor_run.py
@@ -555,7 +555,7 @@ quux_asset_job = define_asset_job("quux_asset_job", [quux_asset], partitions_def
 def add_dynamic_partitions_sensor(context):
     return SensorResult(
         dynamic_partitions_requests=[
-            quux.request_for_adding_partitions(["baz"]),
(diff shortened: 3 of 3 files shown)


python_modules/dagster/dagster/_core/definitions/partition.py
```diff
@@ -732,13 +750,20 @@ def get_run_config_for_partition_key(
         partition_key: str,
         dynamic_partitions_store: Optional[DynamicPartitionsStore] = None,
         current_time: Optional[datetime] = None,
+        check_valid_partition_key=True,
```
clairelin135, line 753: I don't love this pattern...

The alternative is making the `get_run_config_for_partition_key` and `get_tags_for_partition_key` accept a partition object. Then we'd create a `PendingPartition` to pass into those functions. I think this is more performant since we won't be refetch the same partition object multiple times (as we are now), but I don't want to enable this if we plan on removing `Partition` in the near future.

sryza:
@clairelin135 it's a little difficult to review this PR in its current state because it has the changes from the upstream PRs mixed in. Want to rebase and then I'll take a look?

clairelin135:
yep, sorry about that! this should be good to review now

python_modules/dagster/dagster/_core/definitions/partition.py
sryza: Oh wow, I forgot we were doing this O(N) operation every time `get_run_config_for_partition_key` is called.

I don't love this pattern either.  I think it would be cool to remove or at least sequester `Partition`, but probably not best to wait on that for this PR.

A couple different ideas here:
- Add a boolean property to `PartitionsDefinition` that lets implementations declare whether there's any difference between the values inside their `Partition` objects and their `partition_key`. It would be false for everything except for `TimeWindowPartitionsDefinition` I think? Although what does the `Partition` look like for a `MultiPartitionsDefinition` that contains a `TimeWindowPartitionsDefinition`?
- Add a method to `PartitionsDefinition` that accepts a `partition_key` and returns a `Partition` object.

clairelin135: before 1.2.4 we were not--this function just accepted a `Partition` object, which ensured that the partition was valid. But after the removal of `PartitionSet`, these functions now check for the existence of the partition within each call.

> Add a method to PartitionsDefinition that accepts a partition_key and returns a Partition object.

This option sounds good to me, since we can ensure that the partition exists before creating the Partition object. In the future when we remove `Partition`, maybe we can create some internal `ValidatedPartitionKey` object that we pass around to these methods to avoid revalidating on each function call.

python_modules/dagster/dagster/_core/definitions/partition.py
```diff
@@ -90,6 +94,10 @@ def __eq__(self, other: object) -> bool:
             return self.value == other.value and self.name == other.name
 
 
+class PendingPartition(Partition):
```
sryza, line 97: I'm not fully grokking the need for this class - why not just construct a `Partition` object?

clairelin135:
@sryza I refactored `PartitionedConfig` to also have `get_run_config_for_partition` and `get_tags_for_partition` methods, instead of just the methods that accept the partition key (`get_run_config_for_partition_key`, `get_tags_for_partition_key`) and check if it exists.

We can refactor the callsites to rely on the `Partition` methods in a future PR. In a world where `Partition` no longer exists, we can instead pass in a `ValidatedPartitionKey` object so we aren't checking for the existence of the partition key in each call.

python_modules/dagster/dagster/_core/definitions/partition.py
clairelin135: we can, but it feels weird to create a `Partition` object where the key currently doesn't exist. If you're fine with that, I'm open to doing that too.

sryza: That doesn't seem weird to me - an object is just an object. It doesn't have any context about the environment it's in.

clairelin135: ok, got it. I'll remove this class

python_modules/dagster/dagster/_core/definitions/partition.py
```diff
@@ -710,6 +714,23 @@ def has_partition_key(
             partitions_def_name=self._validated_name(), partition_key=partition_key
         )
 
+    def get_pending_partition(self, partition_key: str) -> Partition:
+        return Partition(partition_key, partition_key)
+
+    def request_for_adding_partitions(
```
sryza, line 720: What do you think about `build_add_request`? Briefer and helps avoid the possible misunderstanding that "request" is a verb in this context.

sryza (commented):
I left one last comment on naming. Otherwise, this looks great.

clairelin135:
@sryza thanks for the suggestions, agree that `build_add_request` is clearer. I've gone ahead and renamed to `build_add_request` and `build_delete_request`. I think this is ready for another look!

sryza:
I'm still seeing `request_for_adding_partitions` - might you need to push up your latest changes?

clairelin135:
I think there was something weird going on with git... switching base branch to something else then back to master seemed to fix it

sryza (approved):
Nice! I think it's worth giving people a heads up in #product-review.

PR #12968, opened 2023-03-15
[toy] user-computed data versions

smackesey:
## Summary & Motivation

Adds a toy modeling a scenario where:

- An asset graph resides in an external system
- An asset factory is used to build a representation of the external system's graph
- Dagster forwards materialization requests to the external system, passing an asset key and provenance
- The external system decides whether to actually recompute the asset based on the provenance
- The external system reports back a data version and whether it represents a memoized value
- This information is fed into Dagster by being attached to a Nothing `Output`

## How I Tested These Changes

- `dagit -m external_system_assets` (from `python_modules/dagster_test/toys`
- materialize `alpha`
- materialize `beta` -- it should take ~10 seconds
- materialize `beta` again-- it should be fast because `beta` is memoized
- observe `delta`
- materialize `epsilon`-- it should take ~10 seconds
- materialize `epsilon` again-- it should be fast because `epsilon` is memoized

### DIFF
--- a/python_modules/dagster-test/dagster_test/toys/user_computed_data_versions/__init__.py
+++ b/python_modules/dagster-test/dagster_test/toys/user_computed_data_versions/__init__.py
@@ -1,0 +1,164 @@
+"""Models a case where Dagster is used purely for orchestration and not data transformation.
+
+In this scenario, both the structure of the asset graph and the compute logic for assets are
+defined externally to Dagster. The asset graph is specified in a schema, and an external system
+knows how to compute each asset listed in this schema.
+
+We integrate the system with Dagster by using an asset factory to construct Dagster
+`AssetsDefinition` objects for each node in the asset graph. Dagster forwards materialization
+requests to the external system. We pass an asset specification together with provenance information
+for the last materialization of the asset on record. The external system is responsible for
+correctly processing this information-- if the provenance of the last materialization is stale
+(different from the current state of the asset), then it should recompute the asset and overwrite
+the existing value. Otherwise, it should simply use the existing stored value of the asset.
+
+The materialization endpoint of the external system returns a data version string and a flag
+indicating whether a memoized value was used. This information is passed to the Dagster framework by
+returning a `Nothing` `Output`.
+"""
+import warnings
+from typing import Sequence, cast
+
+from dagster import (
+    AssetKey,
+    AssetsDefinition,
+    AssetSelection,
+    DataProvenance,
+    DataVersion,
+    Definitions,
+    ExperimentalWarning,
+    In,
+    Nothing,
+    OpDefinition,
+    OpExecutionContext,
+    Out,
+    Output,
+    SourceAsset,
+    define_asset_job,
+)
+from typing_extensions import TypedDict
+
+from .external_system import AssetSpec, ExternalSystem, ProvenanceSpec, SourceAssetSpec
+
+warnings.filterwarnings("ignore", category=ExperimentalWarning)
+
+
+def external_asset(asset_spec: AssetSpec):
+    """Factory to build an `AssetsDefinition` object to represent an asset externally defined by an
+    `AssetSpec`.
+
+    The op attached to the `AssetsDefinition` forwards materialization requests to the external
+    system, sending an asset key and provenance info for the last recorded materialization of the
+    asset. The external system responds with a data version and a flag indicating whether a memoized
+    value was used for the materialization (i.e. whether recomputation occurred). This is passed to
+    the dagster framework by packaging it on a returned `Output` object. The `Output` returns
+    `Nothing` because Dagster is not handling passing data in between ops in this scenario.
+
+    Args:
+        asset_spec (AssetSpec):
+            A dictionary containing the metadata that defines the asset.
+
+    Returns (AssetsDefinition):
+        An `AssetsDefinition` instance representing the asset in the external system.
+    """
+    key = AssetKey([asset_spec["key"]])
+    code_version = asset_spec["code_version"]
+    dependencies = {AssetKey([dep]) for dep in asset_spec["dependencies"]}
+
+    def fn(context: OpExecutionContext, *args):
+        external_system = ExternalSystem(context.instance.storage_directory())
+
+        provenance = context.get_asset_provenance(key)
+        provenance_spec = provenance_to_dict(provenance) if provenance else None
+        result = external_system.materialize(
+            asset_spec,
+            provenance_spec,
+        )
+        if result["is_memoized"]:
+            context.log.info(f"Used memoized value for {key.to_user_string()}")
+        yield Output(
+            None,
+            data_version=DataVersion(result["data_version"]),
+            metadata={"is_memoized": result["is_memoized"]},
+        )
+
+    keys_by_input_name = {k.path[-1]: k for k in dependencies}
+    keys_by_output_name = {"result": key}
+    op_def = OpDefinition(
+        name=key.path[-1],
+        ins={k.to_user_string(): In(cast(type, Nothing)) for k in dependencies},
+        outs={"result": Out(Nothing)},
+        code_version=code_version,
+        compute_fn=fn,
+    )
+    return AssetsDefinition(
+        keys_by_input_name=keys_by_input_name,
+        keys_by_output_name=keys_by_output_name,
+        node_def=op_def,
+    )
+
+
+def external_source_asset(source_asset_spec: SourceAssetSpec) -> SourceAsset:
+    """Factory to build a `SourceAsset` object to represent a source asset externally defined by a `SourceAssetSpec`.
+
+    The op attached to the `SourceAsset` forwards observation requests to the external system. The
+    external system responds with the current data version of the asset.
+
+    Args:
+        source_asset_spec (SourceAssetSpec):
+            A dictionary containing the metadata that defines the source asset.
+
+    Returns (SourceAsset):
+        A `SourceAsset` instance representing the source asset in the external system.
+    """
+    key = AssetKey([source_asset_spec["key"]])
+
+    def fn(context) -> DataVersion:
+        external_system = ExternalSystem(context.instance.storage_directory())
+        result = external_system.observe(source_asset_spec)
+        return DataVersion(result["data_version"])
+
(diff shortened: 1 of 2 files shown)


smackesey:
* master
  * **PR #12968** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12968" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12968?utm_source=stack-comment).

sryza (commented):
I suspect most elementl devs associate "backend" with "the dagster web server", so I think it's worth finding a different name

What distinguishes this this from the normal way we expect people to use versioning? Is it that the materializations are provided by the user instead of computed by Dagster? If so, that could be a good way to name this.

I think it's also worth adding a module-level comment block explaining this usage pattern.

python_modules/dagster-test/dagster_test/toys/external_system_assets/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+from dagster._utils.backcompat import ExperimentalWarning
+
+from .external_system import ExternalSystem
+
+warnings.filterwarnings("ignore", category=ExperimentalWarning)
+
+EXTERNAL_SYSTEM = ExternalSystem()
+
+
+def external_asset(key: AssetKey):
```
rexledesma, line 40: Any reason why this doesn't just use `multi_asset`?

python_modules/dagster-test/dagster_test/toys/external_system_assets/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+from dagster import OpExecutionContext, observable_source_asset
+from dagster._core.definitions.assets import AssetsDefinition
+from dagster._core.definitions.data_version import DataProvenance, DataVersion
+from dagster._core.definitions.definitions_class import Definitions
+from dagster._core.definitions.events import AssetKey, Output
+from dagster._core.definitions.input import In
+from dagster._core.definitions.op_definition import OpDefinition
+from dagster._core.definitions.output import Out
+from dagster._core.types.dagster_type import Nothing
+from dagster._utils.backcompat import ExperimentalWarning
```
rexledesma, line 31: nit: prefer to import from `dagster`

python_modules/dagster-test/dagster_test/toys/external_system_assets/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+
+@observable_source_asset
+def delta(context) -> DataVersion:
+    EXTERNAL_SYSTEM.set_storage_directory(context.instance.storage_directory())
+    return DataVersion(EXTERNAL_SYSTEM.get_data_version("delta"))
+
+
+epsilon = external_asset(AssetKey(["epsilon"]))
+
+defs = Definitions(assets=[alpha, beta, delta, epsilon])
```
rexledesma, line 102: could we include two jobs that materialize `alpha` and downstream, as well as `delta` and downstream?

python_modules/dagster-test/dagster_test/toys/external_system_assets/external_system.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def get_data_version(self, key: str) -> str:
+        return self._db.get(key).data_version
+
+    def get_dependencies(self, key: str) -> AbstractSet[str]:
+        return _DEPENDENCIES[key]
+
+    def materialize(
+        self, key: str, provenance_json: Optional[str]
+    ) -> Tuple[str, bool]:  # (data_version, is_memoized)
```
rexledesma, line 60: [1]

rexledesma (changes requested):

python_modules/dagster-test/dagster_test/toys/external_system_assets/__init__.py
smackesey: That's a good question. I haven't fully thought it through, but a constraint on a multi-asset is that the same op is used for all assets. I would need to add additional logic into this master op to handle the selection and yield appropriate outputs. Seems like additional complexity without benefit?

If this is what our stakeholders are using or what you'd like to recommend to them then I can change it, but to me atm it seems cleaner to keep it as is.

smackesey:
>If I understand correctly, what's most unique about this example is not that it uses an external system (which is quite common - e.g. anybody using Dagster to run queries against a data warehouse), but that the user code is computing the data versions.

That's correct, and I do think it's important to get the name right, so I'm going to rename it.

python_modules/dagster-test/dagster_test/toys/user_computed_data_versions/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+        )
+
+    deps = [AssetKey(k) for k in EXTERNAL_SYSTEM.get_dependencies(key.to_user_string())]
+    keys_by_input_name = {k.path[-1]: k for k in deps}
+    keys_by_output_name = {"result": key}
+    op_def = OpDefinition(
+        name=key.path[-1],
+        ins={k.to_user_string(): In(cast(type, Nothing)) for k in deps},
+        outs={"result": Out(Nothing)},
+        code_version=EXTERNAL_SYSTEM.get_code_version(key.to_user_string()),
```
schrockn, line 88: also do not think the fetching code versions should be the responsibility of "external system"

python_modules/dagster-test/dagster_test/toys/user_computed_data_versions/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+        key (AssetKey):
+            The asset key for the represented asset.
+
+    Returns (AssetsDefinition):
+        An `AssetsDefinition` object representing the asset in the external system.
+    """
+
+    def fn(context: OpExecutionContext, *args):
+        # Keep the temp storage directory used by the backend to the current instance.
+        EXTERNAL_SYSTEM.set_storage_directory(context.instance.storage_directory())
```
schrockn, line 67: Recommend just directly constructing ExternalSystem here rather than doing this setter

python_modules/dagster-test/dagster_test/toys/user_computed_data_versions/external_system.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def get_data_version(self, key: str) -> str:
+        return self._db.get(key).data_version
+
+    def get_dependencies(self, key: str) -> AbstractSet[str]:
+        return _DEPENDENCIES[key]
+
+    def materialize(
+        self, key: str, provenance_json: Optional[str]
+    ) -> Tuple[str, bool]:  # (data_version, is_memoized)
```
schrockn, line 60: named tuple instead of comment?

python_modules/dagster-test/dagster_test/toys/user_computed_data_versions/external_system.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+@dataclass
+class _DatabaseRecord:
+    value: Any
+    data_version: str
+    code_version: Optional[str]
+
+
+class _Database:
```
schrockn, line 119: why the underscores?

python_modules/dagster-test/dagster_test/toys/user_computed_data_versions/external_system.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def ensure_exists(self) -> None:
+        if not self._initialized:
+            assert self.root_path is not None
+            if not os.path.exists(self.root_path):
+                os.mkdir(self.root_path)
+            for k, v in _SOURCE_ASSETS.items():
+                with open(self.asset_path(k), "w") as fd:  # source asset
+                    record = _DatabaseRecord(v, _get_hash(v), None)
+                    fd.write(yaml.dump(asdict(record)))
```
schrockn, line 131: doesn't matter that much, but seems kind of goofy to use yaml as a serialization format. json more natural

schrockn (changes requested):
IMO the "ExternalSystem" should only handle the responsibilities that a databricks client (or similar) would handle. That means code version and the dependency graph are not their responsibility.

python_modules/dagster-test/dagster_test/toys/user_computed_data_versions/external_system.py
smackesey: Because these are implementation details of this module (`ExternalSystem` is the public export)

python_modules/dagster-test/dagster_test/toys/user_computed_data_versions/external_system.py
smackesey: Trying to make the interface between the Dagster code and the `ExternalSystem` only use basic collections to more realistically model the data structures used to communicate with an external system (that's why the provenance comes in as json rather than the Dagster `DataProvenance` object.

smackesey:
>IMO the "ExternalSystem" should only handle the responsibilities that a databricks client (or similar) would handle. That means code version and the dependency graph are not their responsibility.

@schrockn I'm not 100% clear on this. If the computation is being done in the external system, then Dagster doesn't contain the code used to actually compute assets. Shouldn't the code version then come from an external source (wherever the transform logic actually is)? Ditto for dependencies, they need to be defined where the transform logic actually lives right?

python_modules/dagster-test/dagster_test/toys/user_computed_data_versions/external_system.py
```diff
@@ -0,0 +1,144 @@
...
+
+from typing_extensions import TypedDict
+
+# ########################
+# ##### SYSTEM
+# ########################
+
+_SOURCE_ASSETS: Mapping[str, Any] = {
+    "delta": 100,
+}
```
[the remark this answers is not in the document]
schrockn: Some comments in general about what source assets represent within this file also would be helpful.

"ExternalSystem simulates source assets outside of Dagster's control by idempotently creating a set of them from dummy values whenever it is constructed"

schrockn (changes requested):
almost there

schrockn (approved):
yay

smackesey:
@rexledesma any further requests?

PR #12981, opened 2023-03-16
[asset-reconciliation][perf] Cache common properties on the TimeWindowPartitionsDefinition

OwenKephart:
## Summary & Motivation

When doing asset reconciliation, it's highly likely that you will have multiple (usually lots!) of the same PartitionsDefinition. The modified methods are expensive to calculate, so let's just do it once and save the result.

I went with this two-method approach as the `@cached_method` decorator enforces that args are passed in as keywords (not positionally), while the PartitionsDefinition base class does not have this restriction.

## How I Tested These Changes

Dramatic performance improvements for large partitioned asset graphs -- for an asset graph with 1000 hourly-partitioned assets:

**Before:** 45s
**After:** 18s


sryza (commented):
I think that this could lead to some unexpected behavior?

If I call `my_hourly_partitions_def.get_last_partition_window(current_time=None)` and then call it again an hour later, would I get the same result?

OwenKephart:
@sryza yeah reasonable point -- rewired things so that it will get an explicit evaluation time on this path. also switched to lru cache to prevent this from growing in an unbounded manner for long-living processes

clairelin135 (commented):
Is there a way to avoid caching when the `current_time=None`? I think we're still caching in this case, which will lead to unexpected behavior i.e. `get_last_partition_key` continuing to return the same value

clairelin135 (approved):
LGTM, thanks for breaking out these methods into an internal caching method and an external user-facing method

PR #12987, opened 2023-03-16
[ui] Allow customizing hour cycle format

### LINKED ISSUE #12921: add 24 hour time format option
Requested in Slack [here](https://dagster.slack.com/archives/C01U5LFUZJS/p1678505617422639?thread_ts=1678397824.635719&cid=C01U5LFUZJS).

Can we offer an option to display times in 24hr format instead of 12hr?

I imaging this being a toggle that is configurable in user settings.

hellendag:
## Summary & Motivation

Resolves #12921.

Make hour cycle format (12-hour, 24-hour) configurable. This adds a `Select` to the User Settings dialog to allow choosing from:

- "Automatic", which is whatever the user's browser language would use in default formatting (en-US is 12-hour, de-DE is 24-hour, etc)
- 12-hour, e.g. 1:00 PM
- 24-hour, e.g. 13:00

Note that "24-hour" is actually "h23" in JS configuration, since there also exists an "h24" that treats midnight as 24:00. I'm not sure this is something we really ever want, but if there are locales that are expecting it, we can always add it.

Additionally, rename `TimezoneContext` and `TimezoneProvider` to `TimeContext` and `TimeProvider`, respectfully, since this context now contains both timezone and hour cycle.

https://user-images.githubusercontent.com/2823852/225655132-0e323cb0-fe8f-48ef-85be-98e7ae7f7161.mov

## How I Tested These Changes

See attached video. Verified that "Automatic" values are respected, including when I change browser language.

Changed hour cycle format to non-automatic values and navigated throughout the app. Verified that the correct hour cycle format is used.


hellendag:
* master
  * **PR #12987** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12987" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/12987?utm_source=stack-comment).

salazarm (approved):
Reselect would help us cut down to just 1 provider for state.

PR #13004, opened 2023-03-17
[ui] Use a multi-select component for log filters

### LINKED ISSUE #12741: Have an easy way to filter to just errors in logs view in dagit
Right now, when viewing logs in dagit, by default all logs are shown for the run. There are toggles for every type of event that is displayed:
Screen Shot 2023-03-06 at 9 35 06 AM
Often when debugging, you want to drill down to specifically the failure events. But in order to do that, you have to switch off the other types of events. The toggle system is actually very counterintuitive to how most people would expect to be filtering events in a debugging scenario. 

I am suggesting adding a shortcut button that appears when there are failure logs, that will automatically filter down to just failure logs.

hellendag:
## Summary & Motivation

Resolves #12741.

On the Run page:

- Replace the log level "switch-tags" with a single select component that allows choosing multiple log levels. In the menu of the component, show the number of logs for that level, as we've done with the switch-tags.
- When changing log levels, the URL querystring will update as before. However, we will also now persist the enabled log levels so that these will be the defaults the next time you visit a Run.

## How I Tested These Changes

Storybook example for LogFilterSelect.

View a run in the app. Change log levels, verify that the log table and querystring update accordingly. Reload the page with no query parameters, verify that my log level selections persist. Navigate to another Run, verify same.


js_modules/dagit/packages/core/src/runs/useQueryPersistedLogFilter.ts
```diff
@@ -74,10 +76,27 @@ export function encodeRunPageFilters(filter: LogFilter) {
...
+  return json.filter((level) => validLevels.has(level));
+};
+
+export function useQueryPersistedLogFilter(): [LogFilter, (updates: LogFilter) => void] {
+  const [storedLogLevels] = useStateWithStorage(EnabledRunLogLevelsKey, validateLogLevels);
+
+  const defaults = React.useMemo(() => {
+    return {...DefaultQuerystring, levels: levelsToQuery(storedLogLevels || DefaultLogLevels)};
+  }, [storedLogLevels]);
+
```
bengotow, line 96: I think this hook is only consumed in the LogsToolbar so it probably doesn't matter, but I could see moving the setter that copies the selection to local storage into this hook (and exposing it in place of the setter returned by on line 97) so that the `useStateWithStorage(EnabledRunLogLevelsKey)` can be private to this file.

bengotow (approved):
This looks great and I like that it takes up so much less space in the toolbar!

js_modules/dagit/packages/core/src/runs/useQueryPersistedLogFilter.ts
hellendag: This was my initial approach. The problem is that it can end up writing levels to storage based on the _querystring_, which I don't think we actually want.

That is, if I navigate to a run with levels set in the querystring, I want those levels to be applied to the filter because that's what I'm being permalinked to, but I *don't* want them written to storage -- they aren't my choices. I think we only want to write to storage when the user explicitly chooses filters.

So within this hook, we only want to read the persisted values and treat them as defaults, and we leave the setter behavior to the component usage itself. I'll add a comment to that effect.

bengotow: Ahh ok that makes sense!

PR #13011, opened 2023-03-17
[Dynamic partition requests 1/3] Validate run request partition in sensor eval

clairelin135:
The proposed dynamic partitions sensor behavior requires users being able to yield run requests for partitions that do not exist yet in sensors/schedules. This PR bumps the validation into sensor/schedule evaluation, where we also raise errors for invalid job names / asset selections.

This enables yielding partitioned run requests from sensors/schedules that have invalid partition keys, via `RunRequest(partition_key=...)`.

Consequences of bumping the validation of partition keys into sensor evaluation:
- In order to evaluate directly instantiated partition run requests (`RunRequest(partition_key=...)`) methods like `build_sensor_context` and `build_schedule_context` have to accept a repository def so we can fetch the partition set of the target job. 
- Users who have written custom tests that evaluate run request objects will receive a different run request object from direct instantiation versus `run_request_for_partition`. The latter will return back a resolved run request with partition tags/config.

In order to prevent breaking user code in the situations above above, `run_request_for_partition` retains its current behavior in returning a resolved run request. We can optionally deprecate this method a designate direct instantiation as the new way forward.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -785,9 +785,10 @@ def build_run_requests(
                 check.failed("Partition key provided for unpartitioned asset")
             tags.update({**partitions_def.get_tags_for_partition_key(partition_key)})
 
-        # Normally the run request would have to be resolved with the partitions def, but
-        # we have already confirmed that the partition key exists. Performing another check
-        # in the run request resolution would be redundant and costly to fetch all partition keys.
+        # Do not call run_request.with_resolved_tags_and_config as the partition key is
+        # valid and there is no config.
+        # Calling with_resolved_tags_and_config is costly in asset reconciliation as it
+        # checks for valid partition keys.
         run_requests.append(
             RunRequest(
                 asset_selection=list(asset_keys),
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -785,9 +785,13 @@ def build_run_requests(
                 check.failed("Partition key provided for unpartitioned asset")
             tags.update({**partitions_def.get_tags_for_partition_key(partition_key)})
 
+        # Normally the run request would have to be resolved with the partitions def, but
+        # we have already confirmed that the partition key exists. Performing another check
+        # in the run request resolution would be redundant and costly to fetch all partition keys.
         run_requests.append(
             RunRequest(
                 asset_selection=list(asset_keys),
+                partition_key=partition_key,
                 tags=tags,
             )
         )
--- a/python_modules/dagster/dagster/_core/definitions/job_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/job_definition.py
@@ -585,6 +585,7 @@ class JobDefinition(PipelineDefinition):
         instance: Optional["DagsterInstance"] = None,
         current_time: Optional[datetime] = None,
     ) -> RunRequest:
+        # TODO deprecate this function?
         """Creates a RunRequest object for a run that processes the given partition.
 
         Args:
@@ -620,13 +621,16 @@ class JobDefinition(PipelineDefinition):
             run_config
             if run_config is not None
             else self.partitioned_config.get_run_config_for_partition_key(
-                partition.name, instance=instance, current_time=current_time
+                partition.name, dynamic_partitions_store=instance, current_time=current_time
             )
         )
         run_request_tags = {
             **(tags or {}),
             **self.partitioned_config.get_tags_for_partition_key(
-                partition_key, instance=instance, current_time=current_time, job_name=self.name
+                partition_key,
+                dynamic_partitions_store=instance,
+                current_time=current_time,
+                job_name=self.name,
             ),
         }
 
@@ -636,6 +640,7 @@ class JobDefinition(PipelineDefinition):
             tags=run_request_tags,
             job_name=self.name,
             asset_selection=asset_selection,
+            partition_key=partition_key,
         )
 
     @public
--- a/python_modules/dagster/dagster/_core/definitions/job_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/job_definition.py
@@ -56,6 +56,7 @@ from dagster._core.selector.subset_selector import (
 )
 from dagster._core.storage.io_manager import IOManagerDefinition, io_manager
 from dagster._core.utils import str_format_set
+from dagster._utils.backcompat import deprecation_warning
 from dagster._utils.merger import merge_dicts
 
 from .asset_layer import AssetLayer, build_asset_selection_job
@@ -585,7 +586,6 @@ class JobDefinition(PipelineDefinition):
         instance: Optional["DagsterInstance"] = None,
         current_time: Optional[datetime] = None,
     ) -> RunRequest:
-        # TODO deprecate this function?
         """Creates a RunRequest object for a run that processes the given partition.
 
         Args:
@@ -605,6 +605,12 @@ class JobDefinition(PipelineDefinition):
         Returns:
             RunRequest: an object that requests a run to process the given partition.
         """
+        deprecation_warning(
+            "JobDefinition.run_request_for_partition",
+            "2.0.0",
+            additional_warn_txt="Directly instantiate `RunRequest(partition_key=...)` instead.",
+        )
+
         if not (self.partitions_def and self.partitioned_config):
             check.failed("Called run_request_for_partition on a non-partitioned job")
 
--- a/python_modules/dagster/dagster/_core/definitions/run_request.py
+++ b/python_modules/dagster/dagster/_core/definitions/run_request.py
@@ -110,7 +110,7 @@ class RunRequest(
                 fields[k] = kwargs[k]
         return RunRequest(**fields)
 
-    def with_resolved_partition(
+    def with_resolved_tags_and_config(
         self,
         target_definition: Union["JobDefinition", "UnresolvedAssetJobDefinition"],
         current_time: Optional[datetime] = None,
@@ -126,7 +126,10 @@ class RunRequest(
         partitions_def = target_definition.partitions_def
         if partitions_def is None:
             check.failed(
-                "Cannot resolve partition for run request without partitions_def",
+                (
+                    "Cannot resolve partition for run request when target job"
+                    f" '{target_definition.name}' is unpartitioned."
+                ),
             )
         partitions_def = cast(PartitionsDefinition, partitions_def)
 
@@ -177,7 +180,7 @@ class RunRequest(
             tags=get_run_request_tags(partition.name),
         )
 
-    def has_resolved_partition_run_request(self) -> bool:
+    def has_resolved_partition(self) -> bool:
         # Backcompat run requests yielded via `run_request_for_partition` already have resolved
         # partitioning
         return self.tags.get(PARTITION_NAME_TAG) is not None if self.partition_key else True
--- a/python_modules/dagster/dagster/_core/definitions/run_request.py
+++ b/python_modules/dagster/dagster/_core/definitions/run_request.py
@@ -1,13 +1,22 @@
+from datetime import datetime
 from enum import Enum
-from typing import Any, Mapping, NamedTuple, Optional, Sequence
+from typing import TYPE_CHECKING, Any, Mapping, NamedTuple, Optional, Sequence, Union, cast
 
 import dagster._check as check
 from dagster._annotations import PublicAttr
 from dagster._core.definitions.events import AssetKey
+from dagster._core.definitions.partition import PartitionedConfig, PartitionsDefinition
+from dagster._core.instance import DynamicPartitionsStore
 from dagster._core.storage.pipeline_run import DagsterRun, DagsterRunStatus
 from dagster._core.storage.tags import PARTITION_NAME_TAG
 from dagster._serdes.serdes import whitelist_for_serdes
 from dagster._utils.error import SerializableErrorInfo
+
+if TYPE_CHECKING:
+    from dagster._core.definitions.job_definition import JobDefinition
+    from dagster._core.definitions.unresolved_asset_job_definition import (
+        UnresolvedAssetJobDefinition,
+    )
 
 
 @whitelist_for_serdes(old_storage_names={"JobType"})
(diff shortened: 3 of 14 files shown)


python_modules/dagster/dagster/_core/definitions/schedule_definition.py
```diff
@@ -736,12 +753,65 @@ def evaluate_tick(self, context: "ScheduleEvaluationContext") -> ScheduleExecuti
...
+                )
+
+                if isinstance(scheduled_target, GraphDefinition):
+                    raise Exception(
+                        f"Error in schedule {self._name}: Schedule returned a partitioned"
+                        " RunRequest, but the target is a graph with no partitions definition."
+                    )
+
+                if isinstance(scheduled_target, PipelineDefinition):
+                    if scheduled_target.is_job:
```
gibsondan, line 779: shouldn't this be

```
if not scheduled_target.is_job:
```

but then i don't understand why tests wouldn't be failing

gibsondan (changes requested):

python_modules/dagster/dagster_tests/definitions_tests/test_sensor_invocation.py
```diff
@@ -173,22 +176,110 @@ def test_instance_access_with_mock():
...
+            invalid_job_str_target_sensor,
+        ]
+
+    with build_sensor_context(repository_def=my_repo) as context:
+        for valid_sensor in [valid_req_sensor, job_str_target_sensor]:
+            run_requests = valid_sensor.evaluate_tick(context).run_requests
+            assert len(run_requests) == 1
+            run_request = run_requests[0]
+            assert run_request.run_config == partition_fn("a")
+            assert run_request.tags.get(PARTITION_NAME_TAG) == "a"
```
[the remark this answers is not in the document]
clairelin135: Ah, so if you do direct invocation
```
valid_sensor(context)
```
you'll just get back the run request returned by the user.

`evaluate_tick` is where we do things like resolving the asset selection, checking for valid target job, etc. so that's where a partitioned run request is resolved and the tags are added.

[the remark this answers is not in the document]
sryza: do we modify the RunRequest in evaluate_tick in other ways? do we add the dagster/sensor_name tag?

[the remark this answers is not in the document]
clairelin135: No, we add the sensor name tag in the daemon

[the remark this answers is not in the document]
clairelin135: hm... actually it all seems pretty fragmented at the moment. for schedules we add the schedule tag in `evaluate_tick`, but in sensors we do it in the daemon. 

I'm happy to move all of this logic into the daemon to unify this

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
clairelin135: ah good catch, yes the `not` is supposed to be in there. the tests didn't fail because this code path is only hit when `partition_key` is specified on the run request, and there wasn't an existing test for that. I've added a test now

python_modules/dagster/dagster_tests/definitions_tests/test_sensor_invocation.py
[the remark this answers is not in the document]
clairelin135: On second thought, resolving the run request requires calling out to user code to fetch the run config per partition. If we did this in the daemon it'd require a gRPC call, I think I'd prefer resolving the run request in `evaluate_tick` instead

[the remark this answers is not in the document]
sryza: Got it - makes sense

python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -583,6 +583,7 @@ def run_request_for_partition(
         instance: Optional["DagsterInstance"] = None,
         current_time: Optional[datetime] = None,
     ) -> RunRequest:
+        # TODO deprecate this function?
```
sryza, line 586: Hmm yes, I think we can deprecate this. Any downside you can think of? If we deprecate, we should explain the alternative in the docstring.

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -783,9 +783,13 @@ def build_run_requests(
                 check.failed("Partition key provided for unpartitioned asset")
             tags.update({**partitions_def.get_tags_for_partition_key(partition_key)})
 
+        # Normally the run request would have to be resolved with the partitions def, but
```
sryza, line 786: I find this comment a little confusing - it seems like it's trying to explain the context for why we're doing something weird in the code, but the code below looks perfectly expected.

Is this explaining why we have:
```python
        if partition_key is not None:
            if partitions_def is None:
                check.failed("Partition key provided for unpartitioned asset")
            tags.update({**partitions_def.get_tags_for_partition_key(partition_key)})
```

python_modules/dagster/dagster/_core/definitions/run_request.py
```diff
@@ -90,19 +100,88 @@ def __new__(
...
-        return self.tags.get(PARTITION_NAME_TAG)
-
     def with_replaced_attrs(self, **kwargs: Any) -> "RunRequest":
         fields = self._asdict()
         for k in fields.keys():
             if k in kwargs:
                 fields[k] = kwargs[k]
         return RunRequest(**fields)
 
+    def with_resolved_partition(
```
sryza, line 113: Would `with_tags_and_config_from_partition_key` be accurate? If so, I think it would make it more clear what's happening here.

python_modules/dagster/dagster/_core/definitions/run_request.py
```diff
@@ -90,19 +100,88 @@ def __new__(
...
+            check.failed(
+                "Cannot resolve partition for run request without partition key",
+            )
+
+        partitions_def = target_definition.partitions_def
+        if partitions_def is None:
+            check.failed(
+                "Cannot resolve partition for run request without partitions_def",
+            )
+        partitions_def = cast(PartitionsDefinition, partitions_def)
```
sryza, line 131: It seems like we already have the `partitions_def` in all the places this method is called? If so, maybe we should just pass it in? You can ignore this comment.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -615,37 +618,104 @@ def has_loadable_targets(self) -> bool:
...
                 )
             elif run_request.job_name and run_request.job_name not in target_names:
                 raise Exception(
                     f"Error in sensor {self._name}: Sensor returned a RunRequest with job_name "
                     f"{run_request.job_name}. Expected one of: {target_names}"
                 )
 
+            if run_request.partition_key and not run_request.has_resolved_partition_run_request():
+                selected_job = _get_repo_job_by_name(
+                    context, run_request.job_name if run_request.job_name else target_names[0]
```
sryza, line 675: Does this handle the case where `@sensor` isn't provided any job names and is instead just provided an asset selection, and there are multiple asset jobs?

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
clairelin135: Ah the comment is referring to how we don't do `run_request.with_resolved_partition(...)` like we do with all other run requests with a partition key.

`with_resolved_partition` is responsible for confirming that the partition exists and applying the necessary tags to the run request, but this causes the asset reconciliation sensor to perform significantly slower, so we directly instantiate here instead.

I'll update this comment to clarify

sryza: How do we avoid doing `run_request.with_resolved_partition`? Doesn't it end up getting called by the sensor machinery once this function returns?

python_modules/dagster/dagster/_core/definitions/job_definition.py
clairelin135: I can't think of any downsides for doing this, I think it's generally much clearer to directly instantiate `RunRequest(partition_key=...)` and provides fewer entry points for us to manage

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
clairelin135: we check if run requests have resolved partitioning via `run_request.has_resolved_partition_run_request()`, which checks if the partition tag exists. This allows run requests generated here and via `run_request_for_partition` to bypass the `with_resolved_partition` function.

If we didn't add that, existing user tests may fail when `repository_def` isn't provided to the sensor/schedule context, since the sensor will assume that run requests created via `run_request_for_partition` require resolution and thus will try to load the repository.

python_modules/dagster/dagster/_core/definitions/run_request.py
clairelin135: yep I think this is a better name

python_modules/dagster/dagster/_core/definitions/run_request.py
clairelin135: It is duplicative to check partitions def again here, but I prefer it as it ensures the partitions def is bound to the target

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
clairelin135: ah good catch--ive updated the PR to handle this

python_modules/dagster/dagster/_core/definitions/run_request.py
```diff
@@ -90,19 +100,88 @@ def __new__(
...
+            if self.run_config
+            else partitioned_config.get_run_config_for_partition_key(
+                partition.name,
+                dynamic_partitions_store=dynamic_partitions_store,
+                current_time=current_time,
+            ),
+            tags=get_run_request_tags(partition.name),
+        )
+
+    def has_resolved_partition_run_request(self) -> bool:
```
sryza, line 180: This name makes it sound like "this RunRequest contains a(nother) RunRequest with a resolved partition". `has_resolved_partition` would be clearer I think.

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
```diff
@@ -291,6 +299,10 @@ def has_captured_logs(self):
     def log_key(self) -> Optional[List[str]]:
         return self._log_key
 
+    @property
+    def repository_def(self) -> Optional["RepositoryDefinition"]:
```
sryza, line 303: The pattern we generally use elsewhere with context methods is to make the return types non-optional - e.g. with the `instance` and `scheduled_execution_time` properties. The advantage of this is that, when someone is using them in a normal way, they don't need to put up null guards to make their code work with type checking.

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
```diff
@@ -745,12 +763,65 @@ def evaluate_tick(self, context: "ScheduleEvaluationContext") -> ScheduleExecuti
             skip_message = None
 
         # clone all the run requests with the required schedule tags
-        run_requests_with_schedule_tags = [
-            request.with_replaced_attrs(
-                tags=merge_dicts(request.tags, DagsterRun.tags_for_schedule(self))
+        run_requests_with_schedule_tags = []
```
sryza, line 766: `resolved_run_requests` would be a clearer name now that we're doing more than just adding schedule tags

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
```diff
@@ -745,12 +763,65 @@ def evaluate_tick(self, context: "ScheduleEvaluationContext") -> ScheduleExecuti
...
+                partitions_def = scheduled_target.partitions_def
+                if not partitions_def:
+                    raise Exception(
+                        f"Error in schedule {self._name}: Schedule returned a partitioned"
+                        " RunRequest, but the target job has no partitions definition."
+                    )
+
+                dynamic_partitions_store = None
+                if isinstance(partitions_def, DynamicPartitionsDefinition):
+                    # Instance also needs to be provided when partitions def is a multipartitions
```
sryza, line 803: Is this a TODO?

sryza: Also, feel free to ignore, but it could make sense to push this (and maybe the check above) down into `with_resolved_tags_and_config` so this code doesn't need to be duplicated twice.

sryza: Thinking a little more on this - it's not clear to me that we need to go out of our way to provide this nice error message, because `evaluate_tick` is not a public function - we wouldn't expect anyone to use it with `build_schedule_context`.

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
```diff
@@ -770,8 +841,9 @@ def targets_unresolved_asset_job(self) -> bool:
     def load_target(
         self,
     ) -> Union[GraphDefinition, PipelineDefinition, UnresolvedAssetJobDefinition]:
-        if isinstance(self._target, DirectTarget):
-            return self._target.load()
+        target = self._target
```
sryza, line 844: This is necessary for type-checking?

python_modules/dagster/dagster_tests/definitions_tests/test_sensor_invocation.py
```diff
@@ -194,22 +197,149 @@ def test_instance_access_with_mock():
...
+    @sensor(asset_selection=AssetSelection.keys("a"))
+    def valid_req_sensor():
+        return RunRequest(partition_key="a")
+
+    @sensor(asset_selection=AssetSelection.keys("a"))
+    def invalid_req_sensor():
+        return RunRequest(partition_key="b")
+
+    @asset(partitions_def=StaticPartitionsDefinition(["a"]))
+    def a():
```
sryza, line 316: nitpick: a little confusing to have the partition name and the asset name be the same

python_modules/dagster/dagster_tests/definitions_tests/test_sensor_invocation.py
```diff
@@ -194,22 +197,149 @@ def test_instance_access_with_mock():
...
+            assert run_request.run_config == partition_fn("a")
+            assert run_request.tags.get(PARTITION_NAME_TAG) == "a"
+            assert run_request.tags.get("yay") == "yay!"
+
+        for invalid_sensor in [invalid_req_sensor, invalid_job_str_target_sensor]:
+            with pytest.raises(DagsterUnknownPartitionError, match="Could not find a partition"):
+                run_requests = invalid_sensor.evaluate_tick(context).run_requests
+
+
+def test_asset_selection_run_request_partition_key():
```
sryza, line 306: Because this example only has a single `PartitionsDefinition`, it could only result in a single implicit asset job. I think it's worth including a test that has multiple implicit asset jobs.

sryza (commented):
A few last small things. Looks very close.

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
clairelin135: good call

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
clairelin135: ah yeah makes sense--I think i just got confused since it appears in one of our api docs examples. I'll remove it https://docs.dagster.io/_apidocs/schedules-sensors#dagster.build_schedule_context

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
clairelin135: nope, I think this was just a remnant of me experimenting with the return types. will remove

sryza (approved):
Nice

PR #13019, opened 2023-03-17
[freshness-refactor][4/n] Simplify scheduling algorithm

OwenKephart:
## Summary & Motivation

further simplifies the freshness-based scheduling logic.

completely removes the concept of "constraint propagation", in favor of just informing nodes what their downstream freshness policies are, then using those freshness policies to create time windows where it's valid to execute the asset.


python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/test_asset_reconciliation_sensor.py
```diff
@@ -1319,6 +1319,7 @@ def test_reconciliation(scenario):
     instance = DagsterInstance.ephemeral()
     run_requests, _ = scenario.do_scenario(instance)
 
+    print(run_requests)
```
sryza, line 1322: rogue print

sryza (approved):
I'm going to trust you on this one because:
- I'm not super familiar with these internals
- The number of lines of code has gone down
- The tests still pass

Does this invalidate any of what's in that Notion doc that you wrote?

Issue #12984, opened 2023-03-16
Regression for "remove need to provide partitions_def to asset job targeted by build_schedule_from_partitioned_job"


gregleongf:
### Dagster version

dagster, version 1.2.1

### What's the issue?

There seems to be a regression related to #11844 task.

My code is pretty similar to the one from documentation:
```
converted_recommendations_job = define_asset_job("converted_recommendations_job", selection=[converted_recommendations])

converted_recommendations_asset_partition_schedule = build_schedule_from_partitioned_job(
    converted_recommendations_job,
    default_status=DefaultScheduleStatus.RUNNING
)
```

And the asset has partition_ref configured
```
@asset(partitions_def=DailyPartitionsDefinition(start_date="2022-11-01"))
```

Yet when I try to run the job on schedule I get this error:
```
dagster._core.errors.ScheduleExecutionError: Error occurred during the execution function for schedule converted_recommendations_job_schedule
  File "/usr/local/lib/python3.7/site-packages/dagster/_grpc/impl.py", line 291, in get_external_schedule_execution
    return schedule_def.evaluate_tick(schedule_context)
  File "/usr/local/lib/python3.7/contextlib.py", line 130, in __exit__
    self.gen.throw(type, value, traceback)
  File "/usr/local/lib/python3.7/site-packages/dagster/_core/errors.py", line 215, in user_code_error_boundary
    ) from e
The above exception was caused by the following exception:
dagster._check.CheckError: Failure condition: Called run_request_for_partition on a non-partitioned job
  File "/usr/local/lib/python3.7/site-packages/dagster/_core/errors.py", line 206, in user_code_error_boundary
    yield
  File "/usr/local/lib/python3.7/site-packages/dagster/_grpc/impl.py", line 291, in get_external_schedule_execution
    return schedule_def.evaluate_tick(schedule_context)
  File "/usr/local/lib/python3.7/site-packages/dagster/_core/definitions/schedule_definition.py", line 582, in evaluate_tick
    result = list(ensure_gen(execution_fn(context)))
  File "/usr/local/lib/python3.7/site-packages/dagster/_core/definitions/partitioned_schedule.py", line 185, in schedule_fn
    current_time=context.scheduled_execution_time,
  File "/usr/local/lib/python3.7/site-packages/dagster/_core/definitions/unresolved_asset_job_definition.py", line 140, in run_request_for_partition
    check.failed("Called run_request_for_partition on a non-partitioned job")
[...]

PR #13021, opened 2023-03-18
Auto-infer partitions def for unresolved asset job

clairelin135:
Fixes https://github.com/dagster-io/dagster/issues/12984

The cause of the error was that the unresolved partitioned schedule definition targeted an unresolved asset job that did not have a partitions definition, thus the call to `run_request_for_partition` threw an error.

In order to fix the issue above, this PR modifies unresolved asset jobs to determine their own partitions def based on the selected assets. Resolving an `UnresolvedPartitionedAssetScheduleDefinition` is now split into three steps:
- validating the target job
- resolving the target job into a `JobDefinition` and inferring its partitions def
- replacing the schedule's target job with the resolved `JobDefinition`

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/partitioned_schedule.py
+++ b/python_modules/dagster/dagster/_core/definitions/partitioned_schedule.py
@@ -34,27 +34,15 @@ class UnresolvedPartitionedAssetScheduleDefinition(NamedTuple):
     day_of_month: Optional[int]
     tags: Optional[Mapping[str, str]]
 
-    def resolve(
-        self, asset_graph: InternalAssetGraph, resolved_job: JobDefinition
-    ) -> ScheduleDefinition:
-        selected_assets = self.job.selection.resolve(asset_graph)
-        partitions_defs = {
-            cast(PartitionsDefinition, asset_graph.get_partitions_def(asset_key))
-            for asset_key in selected_assets
-            if asset_graph.get_partitions_def(asset_key) is not None
-        }
-        if len(partitions_defs) == 0:
-            raise DagsterInvalidDefinitionError(
-                "Tried to build a partitioned schedule from an asset job, but none of the assets"
-                " are partitioned."
-            )
-        if len(partitions_defs) > 1:
-            raise DagsterInvalidDefinitionError(
-                "Tried to build a partitioned schedule from an asset job, but some of the assets"
-                " have different PartitionsDefinitions."
+    def resolve(self, resolved_job: JobDefinition) -> ScheduleDefinition:
+        partitions_def = resolved_job.partitions_def
+        if partitions_def is None:
+            check.failed(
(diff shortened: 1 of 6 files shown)


sryza:
@clairelin135 is this a Draft PR on purpose?

sryza:
Wrapping my head around some of what's going on here:
- Even prior to your change that added support for resolving multi-dimensionally partitioned asset, there was a bug in the code for resolving partitioned schedules.
- The reason we didn't catch this bug is that there was also a bug in `test_unresolved_partitioned_asset_schedule`, which was that we never actually tested with an implicit job in the `Definitions`.

Is that right?

sryza (commented):
Gotcha - thanks for fixing this issue that was my fault in the first place.

Working through the implications of this change:
- If I understand correctly, this change affects jobs that don't even have partitioned schedules pointing to them. Is that right?  I.e. now, in general, you don't need to attach a PartitionsDefinition to your job that targets partitioned assets?
- If that's the case, is there any need to retain the code inside `UnresolvedPartitionedAssetScheduleDefinition.resolve` that resolves gets the `PartitionsDefinition`. Can we just pick it up off of the resolved job?
- If someone previously had a set of assets with different partitions and they targeted them all with the same job, but didn't supply a `PartitionsDefinition` on that job, they'll now hit an error, when previously they might not have? I think this is OK, because it doesn't make sense to do that. But I want to understand what we're getting into here.

clairelin135:
> If I understand correctly, this change affects jobs that don't even have partitioned schedules pointing to them. Is that right? I.e. now, in general, you don't need to attach a PartitionsDefinition to your job that targets partitioned assets?

Yes, that's right. This change will affect any job defined via `define_asset_job` that has partitioned assets but no `partitions_def` specified on the job--now we infer the partitions def based on the assets, and throw an error if the inferred partitions def is not the partitions def explicitly defined on the job.

> If that's the case, is there any need to retain the code inside UnresolvedPartitionedAssetScheduleDefinition.resolve that resolves gets the PartitionsDefinition. Can we just pick it up off of the resolved job?

Yes, we can delete that code now.

> If someone previously had a set of assets with different partitions and they targeted them all with the same job, but didn't supply a PartitionsDefinition on that job, they'll now hit an error, when previously they might not have? I think this is OK, because it doesn't make sense to do that. But I want to understand what we're getting into here.

No, they already see that error now, even if they didn't define a partitions def on the job. They also currently see an error if the specified partitions def is different than any asset in the job.

The only new error that this PR might populate is if you've defined a job with only unpartitioned assets, but the job has a partitions def. Previously there was no error. With this PR, now you see an error.

sryza:
Cool - after that code is deleted, this looks good to go. Also worth updating the title to represent the larger scope of what's going on here.

sryza (approved):
Nice

PR #13006, opened 2023-03-17
[refactor] delete hourly/daily/weekly/monthly schedule decorators, PartitionScheduleDefinition, build_schedule_from_partition

smackesey:
## Summary & Motivation

Internal companion PR: https://github.com/dagster-io/internal/pull/5208

Delete legacy schedule decorators,`PartitionScheduleDefinition`, `build_schedule_from_partitions`.

Mostly this involved converting uses of these decorators into regular `@schedule` calls. Needed to adjust a lot of tests for compatibility with the newer APIs. Some tests targeting `PartitionScheduleDefinition` were removed entirely.

## How I Tested These Changes

Existing test suite.

PR #13027, opened 2023-03-19
AssetSelection allow source assets, error for mixed

smackesey:
## Summary & Motivation

Adjustments to `AssetSelection` and `define_asset_job`:

- `AssetSelection`
    - Allow source assets to be selected by `AssetSelection.keys`
    - Prevent `AssetSelection.upstream` from selecting source assets (previously it did, this was a bug)
    - Throw an error on resolution if selection includes both regular and mixed source assets
    - Rename `AssetSelection.sources` -> `roots`, with `sources` deprecated until 2.0. This is necessary because `sources` doesn't correspond to source assets.
- `UnresolvedAssetJob`
    - Add deprecation warning for 1.3 for passing `assets`/`source_assets` to `resolve` of `unresolved_asset_job`
    - Update internal uses of resolve to pass an `AssetGraph`

## How I Tested These Changes

- Updated some asset selection tests to operate in a context with `SourceAsset`
- Add test for `roots`

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -86,7 +86,7 @@ class AssetGraph:
 
     @staticmethod
     def from_assets(
-        all_assets: Sequence[Union[AssetsDefinition, SourceAsset]]
+        all_assets: Iterable[Union[AssetsDefinition, SourceAsset]]
     ) -> "InternalAssetGraph":
         assets_defs: List[AssetsDefinition] = []
         source_assets: List[SourceAsset] = []
--- a/python_modules/dagster/dagster/_core/definitions/asset_selection.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_selection.py
@@ -1,12 +1,13 @@
 import operator
 from abc import ABC, abstractmethod
 from functools import reduce
-from typing import AbstractSet, Optional, Sequence, Union
+from typing import AbstractSet, Iterable, Optional, Union
 
 import dagster._check as check
-from dagster._annotations import public
+from dagster._annotations import deprecated, public
 from dagster._core.errors import DagsterInvalidSubsetError
 from dagster._core.selector.subset_selector import fetch_connected, fetch_sinks, fetch_sources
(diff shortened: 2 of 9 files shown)


smackesey:
* master
  * **PR #13027** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13027" target="_blank">Graphite</a>  👈
    * **PR #12018** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12018" target="_blank">Graphite</a> 
      * **PR #13006** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13006" target="_blank">Graphite</a> 
      * **PR #11993** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/11993" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13027?utm_source=stack-comment).

sryza (approved):
I suspect that we're going to want an add an `include_sources` argument to `upstream` that defaults to `False`, so that people can easily perform selections of upstream assets without accidentally selecting observable sources.

smackesey:
>I suspect that we're going to want an add an include_sources argument to upstream that defaults to False, so that people can easily perform selections of upstream assets without accidentally selecting observable sources.

Makes sense, we can do that once it becomes possible to run observations and materializations simultaneously.

smackesey:
@will-holley this doesn't affect whether source assets can be used _as dependencies_ together with regular assets-- it only affects whether an `AssetSelection` can include both types (dependencies don't need to be inside the `AssetSelection` of a n asset job).

The reason for this limitation is that the only reason to include source assets in an `AssetSelection` is to run their observation functions, but currently it does not work to launch a run that performs both source asset observations and regular asset materializations-- they need to be in separate runs. This is a temporary state of affairs, and we can allow `AssetSelection` to include both kinds of assets once we can both kinds of operations in the same run.

will-holley:
Thanks for the info, @smackesey! For context I was encountering the issue running an asset materialization job, thus the confusion.

sryza:
@will-holley - sounds like that might be a bug? If so, would you be able to file a github issue for it?

PR #13029, opened 2023-03-19
Support FIPs environments

dmosesson:
Set the flag that says that this is not used for security purposes (as I don't believe that it is). This will allow this code to run in FIPs environments (currently it fails saying that MD5 is not allowed to be used)

Summary & Motivation
Currently, this code fails in FIPs environments because md5 is not allowed to be used for security purposes there, and by default, that is what python assumes that it is being used for. This changes makes it clear that it is not the case so the code will work

How I Tested These Changes
I made the changes on our systems and it works.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/assets.py
@@ -975,7 +975,7 @@ class AssetsDefinition(ResourceAddable):
     @property
     def unique_id(self) -> str:
         """A unique identifier for the AssetsDefinition that's stable across processes."""
-        return hashlib.md5((json.dumps(sorted(self.keys))).encode("utf-8")).hexdigest()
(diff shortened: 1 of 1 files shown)


dmosesson:
Tentatively I committed a change that resolved it. Let me know if that didn't work

sryza:
Thanks for these updates. Unfortunately, this is still failing our type-checking step. Also, our CI runs on Python 3.10, so it might fail with earlier Python versions.  Some more googling has led me to believe that this option is only available on Python 3.9 and later versions.

I think the best course of action here is to:
- Only pass this option if `sys.version_info[:2] >= (3, 9)`
- Use `# type: ignore to avoid the pyright issue`

One more thing: the build reported some issues with formatting, which you can fix by running `make black`.

cmpadden:
This was resolved in #17866

PR #10496, opened 2022-11-11
update mypy, assorted type annotations

smackesey:
### Summary & Motivation

- bump mypy to `0.990`
- many type annotations
- some improvements to `dagster._check`

### How I Tested These Changes

BK

PR #13030, opened 2023-03-19
[refactor] delete Materialization

smackesey:
## Summary & Motivation

Delete `Materialization` from `dagster._legacy`.

This turned out to be extremely simple-- just needed to add `Materialization` to `old_storage_names` of `AssetMaterialization`. This works because `Materialization` has one extra field (`label`) that is dropped on deserialization, and `AssetMaterialization` has an extra field `tags` that gets a default on deserialization. Otherwise the two classes are the same.

## How I Tested These Changes

Added test `test_load_old_materialization` to ensure an old serialized `Materialization` loads as an `AssetMaterialization`.


smackesey:
* master
  * **PR #13030** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13030" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13030?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/events.py
```diff
@@ -548,100 +547,6 @@ def metadata(self) -> MetadataMapping:
...
 
-@whitelist_for_serdes
-class Materialization(
-    NamedTuple(
-        "_Materialization",
-        [
-            ("label", str),
-            ("description", Optional[str]),
-            ("metadata_entries", Sequence[MetadataEntry]),
-            ("asset_key", AssetKey),
```
prha, line 559: Context: 
- `Materialization` which did not require asset key was deprecated in `0.10.0` (https://dagster.phacility.com/D3866)
- `asset_key` was switched from optional to non-optional in `0.12.10` (https://dagster.phacility.com/D9139)

This might be a moot point since it's been this way for so long, but I know that the `Materialization` event preceded the concept of `AssetKey`, so I'm wondering if this will change behavior for deserialization of materialization events persisted without asset keys (pre-`0.10.0`).  

The way I'm reading it is that we've already broken everyone who has stored legacy Materialization events, in which case I'm fine with this change, but could use another brain to think through this, since I don't want our backcompat handling to change with this PR.

smackesey: >The way I'm reading it is that we've already broken everyone who has stored legacy Materialization events, in which case I'm fine with this change, but could use another brain to think through this, since I don't want our backcompat handling to change with this PR.

If `AssetKey` became non-optional on `Materialization` in the past, then yes, it would've already broken anyone storing a `Materialization` without an `AssetKey`, given that there is no custom serializer logic for `Materialization`. So I don't think this PR changes anything there... requesting @alangenfeld for second take.

python_modules/dagster/dagster/_core/definitions/events.py
```diff
@@ -548,100 +547,6 @@ def metadata(self) -> MetadataMapping:
...
-        return super(Materialization, cls).__new__(
-            cls,
-            label=check.str_param(label, "label"),
-            description=check.opt_str_param(description, "description"),
-            metadata_entries=check.opt_sequence_param(
-                metadata_entries, "metadata_entries", of_type=MetadataEntry
-            ),
-            asset_key=asset_key,
-            partition=check.opt_str_param(partition, "partition"),
-        )
```
alangenfeld, line 619: [1]

python_modules/dagster/dagster/_core/definitions/events.py
alangenfeld: while the static typing was off  - it appears to me looking at `[1]` that we would deserialize a `Materialization` with `asset_key` `None` at runtime just fine


through it turns out `AssetMaterialization` does not runtime check against `None` currently either (looks like an accident from https://github.com/dagster-io/dagster/pull/10496) 
```
>>> AssetMaterialization(asset_key=None)
AssetMaterialization(asset_key=None, description=None, metadata_entries=[], partition=None, tags=None)
```

Not sure exactly where that leaves us. I think if we don't fix `AssetMaterialization` then its fine - but if we do then its not? Not sure we can fix `AssetMaterialization` without it being a breaking change unless something else downstream fails on `None` `asset_key`

alangenfeld: > Since it was operational before then, any old Materializations with null asset keys would've been breaking a few months ago. So we're not going to be newly breaking old materialization events if we reintroduce that check now.

I don't follow this - do we coerce `Materialization` in to `AssetMaterialization` somewhere?

smackesey: My mistake, you are right. Re-introducing null-check on `AssetMaterialization` + this PR (which introduces the coercion that was absent before) would break old asset-key-less `Materialization` objects that deserialized fine before.

But IIUC:

- This PR as is does not introduce behavior change because `AssetMaterialization` in its current state does _not_ null-check `AssetKey`
- A reasonable solution to the problem above is to introduce a default dummy asset key

alangenfeld: got it - lets put it under test [2]

python_modules/dagster/dagster_tests/general_tests/compat_tests/test_back_compat.py
```diff
@@ -1163,3 +1166,53 @@ def test_add_primary_keys():
...
+            "asset_lineage": [
+                {
+                    "__class__": "AssetLineageInfo",
+                    "asset_key": {"__class__": "AssetKey", "path": ["foo", "bar"]},
+                    "partitions": {"__set__": ["alpha"]},
+                }
+            ],
+        }
+    )
+
```
alangenfeld, line 1202: [2] add a None asset_key case here as well

smackesey: done

python_modules/dagster/dagster_tests/general_tests/compat_tests/test_back_compat.py
```diff
@@ -1163,3 +1166,53 @@ def test_add_primary_keys():
...
+        old_materialization, StepMaterializationData
+    ) == StepMaterializationData(
+        materialization=AssetMaterialization(
+            description="bar",
+            metadata_entries=[
+                MetadataEntry("baz", description="qux", entry_data=MetadataValue.text("quux"))
+            ],
+            # metadata={
+            #     "baz": MetadataValue.text("quux")
+            # },
```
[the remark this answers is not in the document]
smackesey: oops, removed

alangenfeld (approved):

PR #13033, opened 2023-03-20
move tests away from AssetGroup

sryza:


smackesey (approved):
Good to go when failing test passes

sryza:
This PR should fix the remaining test failure: https://github.com/dagster-io/dagster/pull/13847 (we're moving a test to a test file that doesn't allow warnings).

PR #13039, opened 2023-03-20
Expand scope of try/catch blocks in grpc impl functions

gibsondan:
Summary:
Two problems here we observed:
- the try/catch doesn't catch issues that come up while cleaning up the Sensor/Schedule execution context
- There is 'user code' when creating a schedule/sensor context (creating resources) that is not wrapped in a user code error boundary. This is less important since it mostly determines which exception gets raised, but is still worth fixing?

This fixes those issues by moving the try/catch to the very beginning of the gRPC execution function for all grpc calls, so errors will always bubble up as structured error messages instead of cryptic grpc errors with no stack traces.

## How I Tested These Changes
Existing tests cover gRPC errors in general - testing a failure in the specific new callsites that are newly inside a try/catch block here may not be worth the additional effort? There's also a test flake that hopefully now includes a clearer exception

### DIFF
--- a/python_modules/dagster/dagster/_grpc/impl.py
+++ b/python_modules/dagster/dagster/_grpc/impl.py
@@ -2,7 +2,6 @@
 
 import os
 import sys
-from contextlib import ExitStack
 from typing import Generator, Optional, Sequence, Union
 
 import pendulum
@@ -257,12 +256,14 @@ def get_external_pipeline_subset_result(
             op_selection=solid_selection,
             asset_selection=frozenset(asset_selection) if asset_selection else None,
         )
+        external_pipeline_data = external_pipeline_data_from_def(definition)
+        return ExternalPipelineSubsetResult(
+            success=True, external_pipeline_data=external_pipeline_data
+        )
     except Exception:
         return ExternalPipelineSubsetResult(
             success=False, error=serializable_error_info_from_exc_info(sys.exc_info())
         )
-    external_pipeline_data = external_pipeline_data_from_def(definition)
-    return ExternalPipelineSubsetResult(success=True, external_pipeline_data=external_pipeline_data)
 
 
 def get_external_schedule_execution(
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster/_grpc/impl.py
```diff
@@ -274,42 +275,44 @@ def get_external_schedule_execution(
...
+            if k in required_resource_keys
+        }
+
+        with user_code_error_boundary(
+            ScheduleExecutionError,
+            lambda: "Error occurred during the execution function for schedule {schedule_name}".format(
+                schedule_name=schedule_def.name
+            ),
+        ):
+            with ScheduleEvaluationContext(
```
alangenfeld, line 304: [1] we should leave a note here about why this is in the user code boundary since its not immediately obvious

`# user code to build resources potentially run when creating/entering this context`

python_modules/dagster/dagster/_grpc/impl.py
```diff
@@ -333,8 +336,14 @@ def get_external_sensor_execution(
...
-        sensor_context = stack.enter_context(
-            SensorEvaluationContext(
+
+        with user_code_error_boundary(
+            SensorExecutionError,
+            lambda: "Error occurred during the execution of evaluation_fn for sensor {sensor_name}".format(
+                sensor_name=sensor_def.name
+            ),
+        ):
+            with SensorEvaluationContext(
```
alangenfeld, line 346: [1]

alangenfeld (commented):
do you want to fix the `instance` in this PR too or a follow up?

benpankow:
* master
  * **PR #13039** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13039" target="_blank">Graphite</a>  👈
    * **PR #13041** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13041" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13039?utm_source=stack-comment).

alangenfeld (approved):

PR #13041, opened 2023-03-20
Use instance from sensor/schedule context to instantiate resources, delay until accessed

benpankow:


benpankow:
* master
  * **PR #13039** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13039" target="_blank">Graphite</a> 
    * **PR #13041** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13041" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13041?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -171,21 +151,56 @@ def __enter__(self) -> "SensorEvaluationContext":
...
+
+            instance = self.instance if self._instance or self._instance_ref else None
+            self._resources_cm = build_resources(
+                resources=self._resource_defs or {}, instance=instance
+            )
+
+            self._resources = self._resources_cm.__enter__()
+
+            self._resources_contain_cm = isinstance(self._resources, IContainsGenerator)
+
```
alangenfeld, line 198: I think we can tidy stuff up here a bit - messing with `__del__` is cursed and its not clear to me that we need to.

The cases we have are:
* resource are context mangers
  - we have gone through `__enter__` and have an `ExitStack` we can append to we know will close on `__exit__`
  - we have not gone through `__enter__` and are about to throw `DagsterInvariantViolationError`, nothing to clean up since we are blowing up
* resources are not context managers
  - nothing to clean up 
  
So I think we can get rid of `__del__` and some of these other state vars and just use the exit stack and locals

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
```diff
@@ -198,28 +194,48 @@ def __init__(
...
+                IContainsGenerator,
+            )
+            from dagster._core.execution.build_resources import build_resources
+
+            instance = self.instance if self._instance or self._instance_ref else None
+            self._resources_cm = build_resources(
+                resources=self._resource_defs or {}, instance=instance
+            )
+
+            self._resources = self._resources_cm.__enter__()
```
gibsondan, line 235: would using the ExitStack object give us more options here

gibsondan (changes requested):
agree with alex re: __del__ removal - since we control the object lifecycle here by making it a contextmanager

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
gibsondan: @alangenfeld would an argument for using __del__ be if we expect users to create their own context objects and can't neccesarily trust them to do so in contextmanagers?

benpankow:
Updated to remove the `__del__` logic (since we will call `__exit__` if there are any CM resources, if not no need to clean up). Use the Exit Stack to track the resources CM rather than manipulating it directly.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
benpankow: I think we can get rid of `__del__` since the two cases are:

1. User has no context manager resources. They are able to use the context directly as `my_sensor(build_sensor_context_manager(resources={"foo": bar}))`. There's no resource to clean up here.
2. User has context manager resources. Then, cleanup in del isn't needed since we'll call `__exit__`.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -171,21 +150,44 @@ def __enter__(self) -> "SensorEvaluationContext":
         return self
 
     def __exit__(self, *exc) -> None:
-        self._resources_cm.__exit__(*exc)  # pylint: disable=no-member
```
gibsondan, line 174: i think it is fine that we lose this right? The idea before was that an exception might get raised while the context was in scope and we might want to give the resource contextmanager a chance to catch it?

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
```diff
@@ -198,28 +194,35 @@ def __init__(
...
+                IContainsGenerator,
+            )
+            from dagster._core.execution.build_resources import build_resources
+
+            instance = self.instance if self._instance or self._instance_ref else None
+
+            resources_cm = build_resources(resources=self._resource_defs or {}, instance=instance)
+            self._resources = self._exit_stack.enter_context(resources_cm)
+
+            self._resources_contain_cm = isinstance(self._resources, IContainsGenerator)
```
gibsondan, line 224: does this still need to be state / do we use this anywhere outside this function now?

gibsondan (commented):
ok this makes sense to me, the exception handling piece seems subtle but does my comment here match your understanding @benpankow and @alangenfeld ? seems fine to remove that unless I am missing something?

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
benpankow: Is this not handled by the `ExitStack`'s `__exit__`? maybe we are losing behavior here

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
benpankow: We use it in this function, but in cases where the `if` is not triggered, e.g. on subsequent calls to `.resources`

benpankow: ah wait, yeah, can totally move it to a local

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -171,21 +149,44 @@ def __enter__(self) -> "SensorEvaluationContext":
...
+                my_sensor(context)
+            """
+
+            instance = self.instance if self._instance or self._instance_ref else None
+
+            resources_cm = build_resources(resources=self._resource_defs or {}, instance=instance)
+            self._resources = self._exit_stack.enter_context(resources_cm)
+
+        resources_contain_cm = self._resources and isinstance(self._resources, IContainsGenerator)
+        if resources_contain_cm and not self._cm_scope_entered:
```
alangenfeld, line 185: [1]
* indent this in and just handle it on initial access
* close the exit_stack right before we raise when `self._cm_scope_entered is False` since we don't expect to hit `__exit__` during raise since we did not `__enter__`

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
alangenfeld: the condition we need to handle at [1] is if we did not `__enter__` (`self._cm_scope_entered is False`) then we should run `__exit__` right before we raise the exception

alangenfeld (approved):

PR #13042, opened 2023-03-20
CLI for wiping asset cached status data

clairelin135:
Second part for fixing https://github.com/dagster-io/dagster/pull/12944

After an incorrect value is serialized into the asset status cache, we want to create an escape hatch for users to clear and rebuild their caches. This PR introduces a CLI command that will clear the cache for a selection of assets, or all assets.

Corresponding internal PR: https://github.com/dagster-io/internal/pull/5227

### DIFF
--- a/python_modules/dagster/dagster/_cli/asset.py
+++ b/python_modules/dagster/dagster/_cli/asset.py
@@ -158,3 +158,61 @@ def asset_wipe_command(key, **cli_args):
             click.echo("Removed asset indexes from event logs")
         else:
             click.echo("Exiting without removing asset indexes")
+
+
+@asset_cli.command(name="wipe-cache")
+@click.argument("key", nargs=-1)
+@click.option("--all", is_flag=True, help="Wipe partitions status cache of all asset keys")
+@click.option("--noprompt", is_flag=True)
+def asset_wipe_cache_command(key, **cli_args):
+    r"""Clears the asset partition status cache.
+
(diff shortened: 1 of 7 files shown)


python_modules/dagster/dagster/_cli/asset.py
```diff
@@ -158,3 +158,61 @@ def asset_wipe_command(key, **cli_args):
             click.echo("Removed asset indexes from event logs")
         else:
             click.echo("Exiting without removing asset indexes")
+
+
+@asset_cli.command(name="wipe-cache")
```
sryza, line 163: I think it's worth being verbose and specific with the name - e.g. `wipe-partition-status-cache`.

python_modules/dagster/dagster/_cli/asset.py
```diff
@@ -158,3 +158,61 @@ def asset_wipe_command(key, **cli_args):
...
+@click.argument("key", nargs=-1)
+@click.option("--all", is_flag=True, help="Wipe partitions status cache of all asset keys")
+@click.option("--noprompt", is_flag=True)
+def asset_wipe_cache_command(key, **cli_args):
+    r"""Clears the asset partition status cache.
+
+    Warning: Cannot be undone.
```
sryza, line 170: I don't think we need to warn about this, because the consequence is just that queries will take a little longer.

sryza (approved):
Looks good to go after addressing comments

PR #13043, opened 2023-03-20
[instance] make local artifact directory scheme thread safe

alangenfeld:
The current shared process global temp directory is not thread safe, as multiple threads may simultanously enter the `is None` condition and construct new `TemporaryDirectory` instances that get overwritten and GC-ed causing stale references to become problematic.

One approach would be to just lock the instantiation of the global shared `TemporaryDirectory`

Instead here I revisited the notion that a shared directory was a good idea in the first place and
* made `epehmeral` instances backed by a `TemporaryLocalArtifactStorage` that doesn't even create the directory until we access it.
* made `local_temp` instances use `TemporaryDirectory` instances that we hold a reference to in a weak key dict, using the vended instance as the weak ref key. 

## How I Tested These Changes

added threading tests that fail


python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -2398,8 +2398,6 @@ def __exit__(
         traceback: Optional[TracebackType],
     ) -> None:
         self.dispose()
-        if DagsterInstance._EXIT_STACK:
-            DagsterInstance._EXIT_STACK.close()
```
gibsondan, line 2402: I am very perplexed by how this worked before - the first time you dispose any instance it cleans up all the tempdirs? Is that not very broken? seems like a bigger problem than just thread safety

python_modules/dagster/dagster/_core/storage/root.py
```diff
@@ -49,3 +50,25 @@ def from_config_value(
...
+
+    @property
+    def base_dir(self):
+        if self._temp_dir is None:
+            self._temp_dir = TemporaryDirectory()
+        return self._temp_dir.name
+
+    def dispose(self):
+        if self._temp_dir:
+            self._temp_dir.cleanup()
```
gibsondan, line 74: this could use an ExitStack too but cleanup() maps to __exit__ i imagine?

gibsondan (changes requested):
question about lifecycle / need for weakrefs

python_modules/dagster/dagster/_core/instance/__init__.py
alangenfeld: yea the environment override thing which was the only thing hooked up to the ExitStack had very busted behavior

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -503,10 +508,16 @@ def get() -> "DagsterInstance":
...
+        tempdir: Optional[str] = None,
+        overrides: Optional[DagsterInstanceOverrides] = None,
     ) -> "DagsterInstance":
         if tempdir is None:
-            tempdir = DagsterInstance.temp_storage()
+            created_dir = TemporaryDirectory()
+            i = DagsterInstance.from_ref(
+                InstanceRef.from_dir(created_dir.name, overrides=overrides)
+            )
+            DagsterInstance._TEMP_DIRS[i] = created_dir
```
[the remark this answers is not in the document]
alangenfeld: I briefly went  down this path for the reasoning you laid out but due to way the `local_temp` instance is currently constructed the refactor to accomplish that would be larger than i thought worth it.

python_modules/dagster/dagster/_core/storage/root.py
alangenfeld: yep - this is a more direct variant

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -858,6 +858,10 @@ def dispose(self) -> None:
         if self._secrets_loader:
             self._secrets_loader.dispose()
 
+        if self in DagsterInstance._TEMP_DIRS:
+            DagsterInstance._TEMP_DIRS[self].cleanup()
+            del DagsterInstance._TEMP_DIRS[self]
```
gibsondan, line 863: are there thread safety issues here if multiple DagsterInstances are getting cleaned up at the same time, does this need a lock?

python_modules/dagster/dagster/_core/instance/__init__.py
gibsondan, line 519: Does this also need a lock? Apparently not given your test case but I don't understand why

gibsondan (changes requested):
Seems like we should need a lock here but I don't understand why that wouldn't be reflected in the new test case

python_modules/dagster/dagster/_core/instance/__init__.py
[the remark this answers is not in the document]
alangenfeld: A version that wouldn't be a big of a refactor is making it so that `DagsterInstance` has an `ExitStack` and exposes tacking things on to it, `instance.add_context` or something. Would you prefer that?

gibsondan (approved):

PR #13044, opened 2023-03-20
Deprecate environment_vars argument to `ScheduleDefinition`, `@schedule`

smackesey:
## Summary & Motivation

Deprecate `environment_vars` argument to `ScheduleDefinition` for removal in 1.3. It is a no-op and should've been removed in the past.

## How I Tested These Changes

Existing test suite.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/schedule_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/schedule_decorator.py
@@ -106,8 +106,6 @@ def schedule(
             schedule execution time to determine whether a schedule should execute or skip. Takes a
             :py:class:`~dagster.ScheduleEvaluationContext` and returns a boolean (``True`` if the
             schedule should execute). Defaults to a function that always returns ``True``.
-        environment_vars (Optional[Dict[str, str]]): Any environment variables to set when executing
-            the schedule.
         execution_timezone (Optional[str]): Timezone in which the schedule should run.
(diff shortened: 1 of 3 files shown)


smackesey:
* master
  * **PR #13044** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13044" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13044?utm_source=stack-comment).

gibsondan (commented):
er wait, does the environment_vars arg do anything?

gibsondan (dismissed):
i think the answer is no, and we shouldn't add it in more places (but rather should deprecate / remove it from ScheduleDefinition)

gibsondan:
long ago these schedules executed in crontab, and I think this field set env vars that only ran in the cron process or something?

smackesey:
>er wait, does the environment_vars arg do anything?

Ah, if you're right that makes my life easier. I'm just trying to convert some old `@daily_schedule` uses so that that legacy API can be deleted, and needed this arg to do the conversion. But IIUC those env vars are going to be set anyway even without specifying them here?

gibsondan:
since it's a no-op i have to assume so, yeah

smackesey:
@gibsondan Updated to deprecate in `ScheduleDefinition` instead of add to newer APIs.

dpeng817 (approved):

PR #13047, opened 2023-03-20
[rfc] Add graphql-codegen-typescript-mock-data to generate mock data builders

salazarm:
## Summary & Motivation

Writing mock data can be tedious and hard to maintain. 
By codegen-ing mock data builders we can ease some of that burden.

- When we add a new field or change a schema the underlying builders will take it into account so we won't need to update a bunch of mocks
- Reduces the amount of boiler plate when writing fixtures
- Lets us get back some of the "default" behavior we had before. This might be a little controversial because the previous default behavior was leading to breaking tests, but I think that was because the behavior was random. The codegen defines the default values statically so they never change. Though it does have an option to generate the values dynamically using a static seed so that tests are repeatable.

If there's interest we could codegen a function to create the MockedResponse objects themselves: 
eg:
```
export function buildCloudContextQueryMockResponse(variables: CloudContextQuery, responseData: CloudContextQuery['data']) {
  return {
    request: {
      query: CLOUD_CONTEXT_QUERY,
      variables,
    },
    result: jest.fn(() => ({
      data: {
         __typename: 'DagitCloudQuery',
         ...responseData,
      }
    }))
  }
}
```


js_modules/dagit/packages/core/src/ticks/tests/SensorDryRunDialog.mocks.tsx
```diff
@@ -195,19 +164,16 @@ export const PersistCursorValueMock: MockedResponse<SetSensorCursorMutation> = {
   result: {
     data: {
       __typename: 'DagitMutation',
```
salazarm, line 166: I found that using `buildDagitMutation` created an extremely huge mock of everything so I had to create the outer `__typename: 'DagitMutation'` by hand.

hellendag (approved):

PR #13050, opened 2023-03-20
ExitStack.pop_all -> close

alangenfeld:
The current call sites appear to believe this is equivalent to close even though it is not quite. Clearer to use the correct method.
 
> pop_all()
> Transfers the callback stack to a fresh ExitStack instance and returns it. No callbacks are invoked by this operation - instead, they will now be invoked when the new stack is closed (either explicitly or implicitly at the end of a with statement).

## How I Tested These Changes

bk


alangenfeld:
* master
  * **PR #13050** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13050" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13050?utm_source=stack-comment).

gibsondan (approved):
ah these are quite different actually right? since pop_all does no cleanup? not a huge impact given the location of the callsite, but still

PR #13054, opened 2023-03-21
add docs example for multi-asset conditional materialization

sryza:
## Summary & Motivation

Also adds imports and small editorial changes to the single asset conditional materialization section, for consistency.

### DIFF
--- a/docs/content/concepts/assets/multi-assets.mdx
+++ b/docs/content/concepts/assets/multi-assets.mdx
@@ -46,6 +46,41 @@ def my_function():
 
 By default, the names of the outputs will be used to form the asset keys of the multi-asset. The decorated function will be used to create the op for these assets and must emit an output for each of them. In this case, we can emit multiple outputs by returning a tuple of values, one for each asset.
 
+### Conditional materialization
+
+In some cases, an asset may not need to be updated in storage each time the decorated function is executed. You can use the `is_required` parameter along with `yield` syntax to implement this behavior.
+
+If the `is_required` parameter is set to `False` on an output, and your function does not `yield` an <PyObject object="Output" /> object for that output, then no asset materialization event will be created, the I/O manager will not be invoked, downstream assets will not be materialized, and asset sensors monitoring the asset will not trigger.
+
+```python file=/concepts/assets/multi_asset_conditional_materialization.py
+import random
+
+from dagster import AssetOut, Output, asset, multi_asset
+
+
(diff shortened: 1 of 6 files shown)


docs/content/concepts/assets/multi-assets.mdx
```diff
@@ -46,6 +46,39 @@ def my_function():
 
 By default, the names of the outputs will be used to form the asset keys of the multi-asset. The decorated function will be used to create the op for these assets and must emit an output for each of them. In this case, we can emit multiple outputs by returning a tuple of values, one for each asset.
 
+### Conditional materialization
+
+In some cases, an asset may not need to be updated in storage each time the decorated function is executed. You can use the `is_required` parameter along with `yield` syntax to invoke this behavior. If the `is_required` parameter is set to `False` on an output, you may indicate to the Dagster framework that no data should be persisted to storage by not yielding an output from your computation function. If an output is not emitted during computation, no asset materialization event will be created, the I/O manager will not be invoked, downstream assets will not be materialized, and asset sensors monitoring the asset will not trigger.
```
erinkcochran87, line 51: Thoughts on breaking this paragraph up? It's a bit visually dense - ex:

> In some cases, an asset may not need to be updated in storage each time the decorated function is executed. You can use the `is_required` parameter along with `yield` syntax to invoke this behavior.
>
> If the `is_required` parameter...

erinkcochran87, line 51: > If the `is_required` parameter is set to `False` on an output, you may indicate to the Dagster framework

This feels a bit odd to me - a few thoughts:

- Do we 'indicate' things to Dagster? 
- Is it Dagster or the Dagster framework?

Would something like this still get the point across?

> Setting the `is_required` parameter to `False` on an output tells Dagster that no data should be persisted to storage by not yielding an output from the computation function.

sryza: FWIW this is copy/pasted from the analogous section in the assets page. I agree that there are opportunities to tighten it up though.

> Would something like this still get the point across?

Not 100%, because "not yielding an output from your computation function" is what tells Dagster that no data should be persisted to storage, not just "Setting the is_required parameter to False".

I made a similar change based on your recommendation. Let me know what you think.

erinkcochran87 (approved):
Thanks for taking another pass at this :) I think your changes look great!

Issue #13022, opened 2023-03-18
BindResourcesToJobs is great but if hardcoding a config to job we still need to give the job resource_defs


plaanv:
### Dagster version

dagster, version 1.2.2

### What's the issue?

We should be able to give the job a config without worrying about resources as it will be handled when initializing Definitions.

Doing so results in : 
```
UserWarning: Error loading repository location main.py:dagster._core.errors.DagsterInvalidDefinitionError: resource with key 'foo' required by op 'my_op' was not provided. Please provide a <class 'dagster._core.definitions.resource_definition.ResourceDefinition'> to key 'foo', or change the required key to one of the following keys which points to an <class 'dagster._core.definitions.resource_definition.ResourceDefinition'>: ['io_manager']
```

### What did you expect to happen?

It should not throw an error.

### How to reproduce?

This works : 
```
from dagster import Definitions, op, job, resource, BindResourcesToJobs

@resource()
def foo_resource():
    return "Im a string resource"


@op(required_resource_keys={"foo"})
def my_op(context):
    print(context.foo)


@job
def my_job():
    my_op()


defs = Definitions(
    jobs=BindResourcesToJobs([my_job]),
    resources={"foo": foo_resource}
)

```

But this fails (see the config passed to job) raising an InvalidDefinitionError even before Definitions are loaded (the job decorator is the one involved) : 
```
from dagster import Definitions, op, job, resource, BindResourcesToJobs

@resource()
def foo_resource():
    return "Im a string resource"


@op(required_resource_keys={"foo"})
def my_op(context):
    print(context.foo)

#giving the job a config makes BindResourcesToJobs useless as this fails . Config can even be an empty dict to make it fail:
@job(config={})
def my_job():
    my_op()


defs = Definitions(
    jobs=BindResourcesToJobs([my_job]),
    resources={"foo": foo_resource}
)
```

### Deployment type

Local

### Deployment details

It would fail on any kind of deployment

### Additional information

this has to do with in dagster._core.definitions.job_definition.py 's handling of configs, providing default resources (l.205) and verifying them, but it gets beyond my understanding of the old API with ModeDefinition and all.

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

PR #13057, opened 2023-03-21
[fix] Fix behavior of BindResourcesToJobs when jobs have config specified

benpankow:
## Summary

Addresses #13022. Jobs which have config bound would cause issues when trying to late-bind resources because the job would be validated early, since we construct new `JobDefinition` objects as part of processing this config.

This PR passes through resource-binding status to these nested `JobDefinitions` to ensure that their resources are resolved late.

## Test Plan

new unit test


python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -173,7 +174,11 @@ def __init__(
             _preset_defs, "preset_defs", of_type=PresetDefinition
         )
 
-        did_user_provide_resources = bool(resource_defs)
+        did_user_provide_resources = (
```
dpeng817, line 177: maybe `does_provide_resources` - feels weird to use the term `user` in the code itself

dpeng817 (approved):

PR #13064, opened 2023-03-21
Show unauthorized error graphql error message

salazarm:
## Summary & Motivation

Use the message from the graphql error if it's provided. 

## How I Tested These Changes

jest


js_modules/dagit/packages/core/src/partitions/__tests__/CreatePartitionDialog.test.tsx
```diff
@@ -63,4 +75,33 @@ describe('CreatePartitionDialog', () => {
...
+          repositoryName: 'testing',
+        },
+      },
+      data: buildUnauthorizedError({
+        message: 'test message 123',
+      }),
+    });
+    await act(async () => {
+      render(<Test mocks={[createFixture]} />);
+    });
```
hellendag, line 95: You shouldn't need the `await act` when using `MockedProvider` (instead of `TestProvider`). At least, I haven't needed it so far.

hellendag (approved):

bengotow (approved):

PR #13066, opened 2023-03-21
add toy for eager asset reconciliation

sryza:


bengotow (approved):
I don't have context on this particular feature and the toy assets look simple enough!  I'm a bit confused that there's nothing in the code (aside from the group name) defining "eager" vs "non-eager" behavior. I guess all asset reconciliation sensors are eager? Or it would take some additional code to make it not eager?

sryza:
> I guess all asset reconciliation sensors are eager? Or it would take some additional code to make it not eager?

@bengotow exactly. If you add in `FreshnessPolicy`s, they become not eager.

PR #13070, opened 2023-03-21
Add dynamic partitions name resolver to dimension type

clairelin135:
Currently the Dagit view to add a partition to a multipartitions def containing a dynamic dimension is broken and throws a "you do not have permission" error. The cause of this issue is that the name of the multipartitions def (which is `null`) is being passed into the add partitions mutation, so no matching partitions defs are found.

This PR adds a `dynamicPartitionsName` resolver to the partition dimension gql object, to fetch the name of nested dynamic partitions defs within multipartitions defs. This resolver is not to be confused with the `name` resolver, which represents the dimension name (though longer term we should refactor these objects to be cleaner). 

In addition, modifies the permission check to also detect when a matching partitions def is found that is a dimension a multipartitions def.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql_tests/graphql/test_assets.py
+++ b/python_modules/dagster-graphql/dagster_graphql_tests/graphql/test_assets.py
@@ -355,6 +355,13 @@ GET_2D_ASSET_PARTITIONS = """
                     }
                 }
             }
+            partitionDefinition {
+                name
+                dimensionTypes {
+                    dynamicPartitionsDefinitionName
+                    name
+                }
+            }
         }
     }
 """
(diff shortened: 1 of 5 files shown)


python_modules/dagster-graphql/dagster_graphql_tests/graphql/test_assets.py
```diff
@@ -355,6 +355,13 @@
                     }
                 }
             }
+            partitionDefinition {
+                name
+                dimensionTypes {
+                    dynamicPartitionsDefinitionName
```
salazarm, line 361: so always use `dynamicPartitionsDefinitionName` if available otherwise `name` ?

salazarm (approved):

python_modules/dagster-graphql/dagster_graphql_tests/graphql/test_assets.py
clairelin135: `dynamicPartitionsDefinitionName` should only be used in the add partitions mutation, `name` should be used everywhere where we display the dimension names.

Since it's possible that somebody could do something like this:
```
MultiPartitionsDefinition({"a": DynamicPartitionsDefinition("foo"), "b": DynamicPartitionsDefinition("foo")})
```

PR #13072, opened 2023-03-21
[rfc] CancelationSource

dpeng817:
Adds the notion of a `CancelationSource` - which intends to provide context when a run moves into a CANCELING state, of where the cancelation request is coming from. The idea here is to distinguish between a run that is canceled by the user (IE clicking the `terminate run` button in dagit) and a run that is canceled by the system itself (long-running runs being terminated)

Right now there are only two ways to trigger cancelation:
- Via graphql (including clicking the `terminate run` button in the dagster UI under this). We would consider this a user cancelation
- (Once https://github.com/dagster-io/dagster/pull/12920 lands) Via the monitoring daemon. We would consider this a system cancelation.

The CancelationSource is only added to the `CANCELING` event mostly to avoid redundancy. Would potentially need to hit the db if we wanted to also add the CancelationReason to the `CANCELED` event, but could be convinced to add that for completeness.

### DIFF
--- a/python_modules/dagster/dagster/_core/run_coordinator/base.py
+++ b/python_modules/dagster/dagster/_core/run_coordinator/base.py
@@ -1,7 +1,6 @@
 from abc import ABC, abstractmethod
 from typing import NamedTuple, Optional
 
-from dagster._core.events import CancellationReason
 from dagster._core.instance import MayHaveInstanceWeakref, T_DagsterInstance
 from dagster._core.storage.pipeline_run import DagsterRun
 from dagster._core.workspace.context import IWorkspace, WorkspaceRequestContext
@@ -41,7 +40,7 @@ class RunCoordinator(ABC, MayHaveInstanceWeakref[T_DagsterInstance]):
         """
 
     @abstractmethod
-    def cancel_run(self, run_id: str, cancellation_reason: CancellationReason) -> bool:
+    def cancel_run(self, run_id):
         """Cancels a run. The run may be queued in the coordinator, or it may have been launched.
 
         Returns False is the process was already canceled. Returns true if the cancellation was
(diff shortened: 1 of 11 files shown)


dpeng817:
* master
  * **PR #12920** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12920" target="_blank">Graphite</a> 
    * **PR #13072** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13072" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13072?utm_source=stack-comment).

python_modules/dagster/dagster/_core/run_coordinator/base.py
```diff
@@ -40,7 +40,7 @@ def submit_run(self, context: SubmitRunContext) -> DagsterRun:
         """
 
     @abstractmethod
-    def cancel_run(self, run_id: str) -> bool:
+    def cancel_run(self, run_id, cancelation_source):
```
gibsondan, line 43: technically run coordinators are user-pluggable so we will want to default this to None to make this not a breaking change

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -472,8 +472,6 @@ def terminate(self, run_id, message=None):
 
         self.ecs.stop_task(task=tags.arn, cluster=tags.cluster)
 
-        run = self._instance.get_run_by_id(run_id)
-        self._instance.report_run_canceling(run, message=message)
```
gibsondan, line 476: does this belong on the other PR?

gibsondan (changes requested):
what if we kept moving things up the chain and moved the report_run_canceling call into the callsite here as well? I think there is exactly one callsite in graphql, could we emit it in that callsite?

dpeng817:
Removing this in favor of having long-running run termination just mark the run as failed

Issue #11808, opened 2023-01-19
Make `AssetsDefinition.to_source_assets()` method public


bendnorman:
### What's the use case?

The `AssetsDefinition.to_source_assets()` method makes it easier to connect assets between `Definitions`.

### Ideas of implementation

I think this is just a matter of adding a `@public` decorator to the `to_source_assets()` method and updating the documentation.

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

PR #13073, opened 2023-03-22
make AssetsDefinition.to_source_assets public

sryza:
## Summary & Motivation

Fixes https://github.com/dagster-io/dagster/issues/11808

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/assets.py
@@ -885,7 +885,13 @@ class AssetsDefinition(ResourceAddable):
                 freshness_policies_by_key=self.freshness_policies_by_key,
             )
 
+    @public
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster/_core/definitions/assets.py
```diff
@@ -885,7 +885,13 @@ def subset_for(
                 freshness_policies_by_key=self.freshness_policies_by_key,
             )
 
+    @public
     def to_source_assets(self) -> Sequence[SourceAsset]:
+        """Returns a SourceAsset for each asset in this definition.
+
+        Each SourceAsset will have the same key, metadata, io_manager_key, etc. as one of the assets
```
clairelin135, line 892: nit: I think the following is clearer:
```suggestion
        Each produced SourceAsset will have the same key, metadata, io_manager_key, etc. as the corresponding asset
```

clairelin135 (approved):

PR #13080, opened 2023-03-22
feat(helm): add path-prefix to dagit command

### LINKED ISSUE #3637: [helm] dagit command overrides
## Use Case

To be able to for example, set `--path-prefix` in the `dagit` invocation to allow it to be hosted on a non `/` path

mpicard:
## Summary & Motivation

partially fixes #3637 by allowing the --path-prefix argument to be supplied via the Helm chart.

## How I Tested These Changes

These changes have been unit tested using the helm schema tests


alangenfeld:
thanks!

PR #13085, opened 2023-03-22
Fix issue where backfill fails when gRPC server is replaced mid-backfill

gibsondan:
Summary:
When the daemon is running its own gRPC servers locally, the backfill can take long enough that the old gRPC server no longer exists. To fix this, generate a new CodeLocation object for each run in the backfill.

Test Plan: BK, run a 2500-run backfill locally in dagster dev, no more failure ~1000 runs in.


python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -250,7 +257,10 @@ def execute_asset_backfill_iteration(
     for run_request in result.run_requests:
         yield None
         submit_run_request(
-            run_request=run_request, asset_graph=asset_graph, workspace=workspace, instance=instance
+            run_request=run_request,
+            asset_graph=asset_graph,
+            workspace=workspace_process_context.create_request_context(),
```
alangenfeld, line 262: leave a comment to prevent future optimization from undoing

alangenfeld (approved):
lgtm

PR #13086, opened 2023-03-22
Greatly reduce the number of gRPC calls when doing large asset backfills

gibsondan:
Summary:
By caching the externalpipeline and externalexecutionplan when they are identical, we can greatly reduce the amount of back and forth that we have to do with the gRPC server. When I run a 2500-run backill of hourly_partitioned_asset, this moves it from 5000 gRPC calls to 2.

Test Plan: Run the above backfill locally

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/asset_backfill.py
+++ b/python_modules/dagster/dagster/_core/execution/asset_backfill.py
@@ -2,6 +2,7 @@ import json
 from typing import (
     TYPE_CHECKING,
     AbstractSet,
+    Dict,
     Iterable,
     List,
     Mapping,
@@ -9,6 +10,7 @@ from typing import (
     Optional,
     Sequence,
     Set,
+    Tuple,
     cast,
 )
 
@@ -28,6 +30,10 @@ from dagster._core.definitions.run_request import RunRequest
 from dagster._core.definitions.selector import PipelineSelector
 from dagster._core.errors import DagsterBackfillFailedError
 from dagster._core.events import DagsterEventType
+from dagster._core.host_representation import (
+    ExternalExecutionPlan,
+    ExternalPipeline,
+)
 from dagster._core.instance import DagsterInstance, DynamicPartitionsStore
 from dagster._core.storage.pipeline_run import DagsterRunStatus, RunsFilter
 from dagster._core.storage.tags import BACKFILL_ID_TAG, PARTITION_NAME_TAG
@@ -35,6 +41,7 @@ from dagster._core.workspace.context import (
     BaseWorkspaceRequestContext,
     IWorkspaceProcessContext,
 )
+from dagster._utils import hash_collection
 from dagster._utils.caching_instance_queryer import CachingInstanceQueryer
 
 if TYPE_CHECKING:
@@ -254,6 +261,10 @@ def execute_asset_backfill_iteration(
     if result.backfill_data.is_complete():
         updated_backfill = updated_backfill.with_status(BulkActionStatus.COMPLETED)
 
+    pipeline_and_execution_plan_cache: Dict[
+        int, Tuple[ExternalPipeline, ExternalExecutionPlan]
+    ] = {}
+
     for run_request in result.run_requests:
         yield None
         submit_run_request(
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -286,26 +297,39 @@ def submit_run_request(
...
+    if not run_request.asset_selection:
+        check.failed("Expected RunRequest to have an asset selection")
+
     pipeline_selector = PipelineSelector(
         location_name=location_name,
         repository_name=repo_handle.repository_name,
         pipeline_name=job_name,
         asset_selection=run_request.asset_selection,
         solid_selection=None,
     )
```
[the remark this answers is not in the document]
sryza: The tests inside test_asset_backfill are just testing the side-effect-less inner function.

Here's an example scenario that would result in runs for different asset selections within the same the same backfill iteration:
- Asset A -> Asset B
- Asset A -> Asset C
- A, B, and C all have different `PartitionsDefinition`s
- Partition "foo_b" of B depends on partition "foo_a" of A
- Partition "foo_c" of C depends on partition "foo_a" of A
- The user launches backfill that covers "foo_a" and all its descendant asset partitions
- In iteration 1, the backfill daemon launches a run for "foo_a" with asset selection [A]
- The run for "foo_a" of A completes
- In iteration 2, the backfill daemon launches two runs
  - "foo_b" with asset selection [B]
  - "foo_c" with asset selection [C]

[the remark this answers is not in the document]
gibsondan:
```
@asset(partitions_def=partitions_a)
def asset_a():
    pass


@asset(
    partitions_def=partitions_b,
    ins={"asset_a": AssetIn(partition_mapping=StaticPartitionMapping({"foo_a": "foo_b"}))},
)
def asset_b(asset_a):
    pass


@asset(
    partitions_def=partitions_c,
    ins={"asset_a": AssetIn(partition_mapping=StaticPartitionMapping({"foo_a": "foo_c"}))},
)
def asset_c(asset_a):
    pass
```

is my attempt to map that onto code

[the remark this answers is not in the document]
sryza: @gibsondan that looks right to me

gibsondan:
Added test that fails if I make the cache misbehave (for example always returning the same key)

python_modules/dagster/dagster/_core/definitions/selector.py
```diff
@@ -60,6 +62,9 @@ def with_solid_selection(self, solid_selection):
             self.location_name, self.repository_name, self.pipeline_name, solid_selection
         )
 
+    def get_id(self) -> str:
+        return hash_str(json.dumps(self._asdict()))
```
gibsondan, line 66: @alangenfeld this is inspired from serdes (this is not officially a serdes class) - should be consistent though? Unless serdes does some additional ordering more than json.dumps does

alangenfeld: that should be fine but some relevant code in https://github.com/dagster-io/dagster/pull/12293/files if you defer til thats landed

sryza: What would be the drawbacks of naming this `__hash__` and use these objects directly as the dictionary key?

python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -254,6 +260,10 @@ def execute_asset_backfill_iteration(
     if result.backfill_data.is_complete():
         updated_backfill = updated_backfill.with_status(BulkActionStatus.COMPLETED)
 
+    pipeline_and_execution_plan_cache: Dict[
```
sryza, line 263: Idea here that you can choose to ignore:

```python
class PipelineAndExecutionPlanCache:
    def __init__(workspace):
        ...

    @cached_method
    def get_pipeline_and_execution_plan(self, selector):
        code_location = workspace.get_code_location(repo_handle.code_location_origin.location_name)
        ...
```

gibsondan: not ideologically opposed but the specific I am choosing to ignore is that I don't think we can pass the workspace into the constructor - since it may change between calls after https://github.com/dagster-io/dagster/pull/13085

Issue #12039, opened 2023-02-02
Can't test time windowed partitioned asset that uses `asset_partitions_time_window_for_output`


MathieuLaksh:
### Dagster version

1.1.7

### What's the issue?

```
@asset(
    partitions_def=DailyPartitionsDefinition(start_date=datetime(2023, 1, 1)),
)
def partitioned_asset(context):
    start, end = context.asset_partitions_time_window_for_output()
    ...
```

Called in the test:
```
context = build_op_context(
      partition_key="2023-02-02",
  )
partitioned_asset(context)
```

Fails with 
```

self = <dagster._core.execution.context.invocation.BoundOpExecutionContext object at 0x40327af540>
output_name = 'result'

    @public
    def asset_partitions_time_window_for_output(self, output_name: str = "result") -> TimeWindow:
        """The time window for the partitions of the output asset.
    
        Raises an error if either of the following are true:
        - The output asset has no partitioning.
        - The output asset is not partitioned with a TimeWindowPartitionsDefinition.
        """
>       return self._step_execution_context.asset_partitions_time_window_for_output(output_name)
E       AttributeError: 'BoundOpExecutionContext' object has no attribute '_step_execution_context'. Did you mean: 'get_step_execution_context'?

/venv/lib/python3.10/site-packages/dagster/_core/execution/context/compute.py:392: AttributeError
```

This seems related to https://github.com/dagster-io/dagster/issues/10586, but maybe I'm missing something when using `build_op_context()`

### How to reproduce?

Create a partitioned asset that attempts to access the partition time window e.g. `start, end = context. asset_partitions_time_window_for_output()`
Attempt to call the asset function with a context created with `build_op_context`

### Deployment type

None

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

prha:
@MathieuLaksh thanks for the report!  We'll take a look and get back to you.

MathieuLaksh:
Hi! Any updates on this?

clairelin135:
Hi Mathieu. I have a fix in progress for this: https://github.com/dagster-io/dagster/pull/13088

MathieuLaksh:
Thanks a lot!

PR #13088, opened 2023-03-22
Add assets def to op context

clairelin135:
Fixes https://github.com/dagster-io/dagster/issues/12039

Assets that are directly invoked for testing can't access fields on the assets def, i.e. the partitions def. This PR adds the assets def to the context to enable accessing these fields.

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/context/compute.py
+++ b/python_modules/dagster/dagster/_core/execution/context/compute.py
@@ -5,6 +5,7 @@ from typing_extensions import TypeAlias
 
 import dagster._check as check
 from dagster._annotations import experimental, public
+from dagster._core.definitions.assets import AssetsDefinition
 from dagster._core.definitions.data_version import (
     DataProvenance,
     extract_data_provenance_from_entry,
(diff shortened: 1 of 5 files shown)


sryza (commented):
I think it might be useful in general to expose the assets def as a public attribute on `OpExecutionContext`

Would that allow us to avoid the underscore at the beginning of `_assets_def` here?

clairelin135:
This PR is now updated to add `assets_def` as a property to the context

python_modules/dagster/dagster/_core/execution/context/compute.py
```diff
@@ -267,6 +272,16 @@ def op_def(self) -> OpDefinition:
         """OpDefinition: The current op definition."""
         return cast(OpDefinition, self.op.definition)
 
+    @public
+    @property
+    def assets_def(self) -> AssetsDefinition:
+        assets_def = self.job_def.asset_layer.assets_def_for_node(self.solid_handle)
+        if assets_def is None:
+            raise DagsterInvalidInvocationError(
```
dpeng817, line 280: Maybe `DagsterInvalidPropertyError` instead

clairelin135: makes sense, switched over

PR #13090, opened 2023-03-22
Use dynamic partition definition name for dimension of multipartition definition

salazarm:
## Summary & Motivation

Multipartition definitions don't have names, instead the partitions they compose have names, so we should be using those in the create partition dialog.

## How I Tested These Changes

Definition:
```
MultiPartitionsDefinition(
    {
        "abc": abc_def,
        "123": DynamicPartitionsDefinition(name="testing123"),
    }
)
```

You can see in the dialog the label is still "123".
In the CreatePartitionDialog we see the "testing123"

I successfully added the partition "hello1234"

Screen Shot 2023-03-22 at 1 18 49 PM


clairelin135 (approved):
It's working for me! 🙏

PR #10371, opened 2022-11-04
Add syntax highlighting to metadata markdown in dagit

smackesey:
### Summary & Motivation

When a language-specified code block is inside a `MarkdownMetadataValue`, the code block is now rendered with syntax highlighting.

A user requested this functionality [here](https://dagster.slack.com/archives/C01U954MEER/p1666916513841459).

This required adding the `rehype-highlight` and `highlight.js` package, as well as updating some other markdown-related packages.

### How I Tested These Changes

Loaded this repo and observed syntax highlighting:

```python
from dagster import MetadataValue, asset, repository

CODE = """
print("Hello, world!")
"""

@asset(
    metadata={
        "python": MetadataValue.md(f'Testing testing testing\n\n```python\n{CODE}\n```\n\nTesting testing testing'),
    }
)
def alpha():
    return 100

@repository
def repo():
    return [alpha]
```

image

PR #13092, opened 2023-03-22
[ui] Upgrade react-markdown

hellendag:
## Summary & Motivation

Upgrade react-markdown.

- Move it out of `dagster-io/ui`, which does not need to provide it. This makes it much easier to mock locally for Jest in `packages/core`, which is necessary because `react-markdown` is ESM-only, and Jest can't handle that.
- Add syntax highlighting as done in https://github.com/dagster-io/dagster/pull/10371.

## How I Tested These Changes

TS, jest, lint.

@smackesey, can you try to apply this on your branch and verify that it does what you're aiming to do?


hellendag:
* master
  * **PR #13092** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13092" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13092?utm_source=stack-comment).

smackesey (approved):
Tried it out with the test code in https://github.com/dagster-io/dagster/pull/10371

It appears to work, so I think we're good to go

PR #13096, opened 2023-03-22
Include much more k8s debug info when run monitoring fails

gibsondan:
Summary:
Like we already have with the EcsRunLauncher, when a k8s job crashes or fails to start, include:
- the last 10 logs from the pod
- the status of the containers in the pod
- any warning events in the pod


johannkm (approved):
🙌

PR #13097, opened 2023-03-22
[dagster-snowflake] config to determine how to handle timestamp data

jamiedemaria:
## Summary & Motivation
There is an issue with storing pandas Timestamp values in snowflake where the year will get converted to a non-valid year (example 48399). In pr https://github.com/dagster-io/dagster/pull/8760 i got around this by storing pandas timestamps as strings, but i finally found this https://github.com/snowflakedb/snowflake-connector-python/issues/319 that indicates that the real fix is to include timezone information in your pandas timestamps https://github.com/snowflakedb/snowflake-connector-python/issues/319#issuecomment-764145625.

This PR adds a new configuration value to the snowflake io manager that allows the user to specify if they want to convert timestamp data to strings, or add a timezone to it. We also check the column types of the table to ensure that the chosen conversion aligns with the type of the corresponding table.

Right now I have deprecation warnings in place so that we can remove the string conversion at some point. I'm not sure what version to choose as the deprecation version, or if committing to deprecating makes sense right now. 

I also added documentation explaining the behavior of the configuration value and providing SQL commands that will allow a user to migrate a column from string to timestamp if they wish to move from converting timestamp data to strings to using timezones.

See https://github.com/dagster-io/dagster/pull/12190 for additional context

### DIFF
--- a/docs/content/integrations/snowflake/reference.mdx
+++ b/docs/content/integrations/snowflake/reference.mdx
@@ -332,34 +332,14 @@ In this example, the `iris_dataset` asset will be stored in the `IRIS` schema, a
 ---
 
 ## Storing timestamp data in Pandas DataFrames
-Due to a longstanding bug in the Snowflake Pandas connector, loading timestamp data from a Pandas DataFrame to Snowflake sometimes causes the data to be corrupted. Prior to `dagster-snowflake` version `0.19.0` we solved this issue by converting all timestamp data to strings before loading the data in Snowflake, and doing the opposite conversion when fetching a DataFrame from Snowflake. However, we can also avoid this issue by ensuring that all timestamp data has a timezone. This allows us to store the data as TIMESTAMP_NTZ(9) type in Snowflake.
-
-To specify how you would like timestamp data to be handled, use the `time_data_to_string` configuration value for the Snowflake I/O manager. If `True`, the I/O manager will convert timestamp data to a string before loading it into Snowflake. If `False` the I/O manager will ensure the data has a timezone (attaching the UTC timezone if necessary) before loading it into Snowflake.
-
-If you would like to migrate a table created prior to `0.19.0` to one with a TIMESTAMP_NTZ(9) type, you can run the follow SQL queries in Snowflake. In the example, our table is located at `database.schema.table` and the column we want to migrate is called `time`:
-
-```sql
-
-// Add a column of type TIMESTAMP_NTZ(9)
-ALTER TABLE database.schema.table
-ADD COLUMN time_copy TIMESTAMP_NTZ(9)
-
-// copy the data from time and convert to timestamp data
-UPDATE database.schema.table
-SET time_copy = to_timestamp_ntz(time)
-
-// drop the time column
-ALTER TABLE database.schema.table
-DROP COLUMN time
-
-// rename the time_copy column to time
-ALTER TABLER database.schema.table
-RENAME COLUMN time_copy TO time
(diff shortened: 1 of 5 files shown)


jamiedemaria:
* master
  * **PR #13097** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13097" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13097?utm_source=stack-comment).

python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
```diff
@@ -128,9 +129,29 @@ def my_table_a(my_table: pd.DataFrame) -> pd.DataFrame:
                 ),
                 is_required=False,
             ),
+            "time_data_to_string": Field(
```
sryza, line 132: `store_datetimes_without_timezones_as_strings` might be clearer? also, I think worth mentioning in the docs that the reason we offer this is to offer compatibility with earlier versions.

python_modules/libraries/dagster-snowflake-pandas/dagster_snowflake_pandas/snowflake_pandas_type_handler.py
```diff
@@ -36,6 +59,20 @@ def _convert_string_to_timestamp(s: pd.Series) -> pd.Series:
         return s
 
 
+def _add_missing_timezone(s: pd.Series, column_types) -> pd.Series:
```
sryza, line 62:
```suggestion
def _add_missing_timezone(s: pd.Series, column_types: Mapping[str, str]) -> pd.Series:
```

docs/content/integrations/snowflake/reference.mdx
```diff
@@ -330,6 +331,39 @@ In this example, the `iris_dataset` asset will be stored in the `IRIS` schema, a
 
 ---
 
+## Storing timestamp data in Pandas DataFrames
```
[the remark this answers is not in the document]
erinkcochran87: I like Jamie's approach. The migration guide is a lot to take in, and I agree that it would likely get lost.

erinkcochran87 (commented):
Content-wise, this looks good to me :+1:

python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
sryza: `store_timestamps_as_strings` sounds good

erinkcochran87 (approved):

jamiedemaria:
@benpankow @sryza  the last thing to decide for this PR is what version to list as the deprecation version for converting to strings. some options:

dagster-snowflake 0.20.0 (or 0.21.0 etc)
dagster-snowflake 1.0.0
dagster 2.0.0

jamiedemaria:
@benpankow @sryza I'd like to get this in for the 1.3 code freeze (so ideally merge tomorrow afternoon). I'm getting buildkite passing after some nasty merge conflicts, could you do a final review pass and let me know if you have opinions on a deprecation date? I lean toward punting or doing `dagster-snowflake=1.0.0` but that's more of a gut feeling, so i'm happy to go another direction

sryza:
I think we can punt on specifying a deprecation version

PR #13099, opened 2023-03-22
Set stdin to DEVNULL when opening dagster subprocesses

gibsondan:
Summary:
To prevent terminal signals from hanging dagster code.

Test Plan: New test case that failed before

### DIFF
--- a/python_modules/dagster/dagster/_serdes/ipc.py
+++ b/python_modules/dagster/dagster/_serdes/ipc.py
@@ -211,7 +211,13 @@ def open_ipc_subprocess(parts: Sequence[str], **kwargs: object) -> Popen[bytes]:
     else:
         # Works on all UNIX systems (but not WASI, see: https://docs.python.org/3/library/os.html#os.setpgrp)
         preexec_fn = os.setpgrp
-    return subprocess.Popen(parts, creationflags=creationflags, preexec_fn=preexec_fn, **kwargs)
+    return subprocess.Popen(
+        parts,
+        creationflags=creationflags,
+        preexec_fn=preexec_fn,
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster_tests/core_tests/serdes_tests/import_readline.py
```diff
@@ -0,0 +1,2 @@
+# This will hang in OSX if it is imported in a subprocess with stdin not equal to DEVNULL
```
alangenfeld, line 1: not sure this is OSX specific - that line of thinking was incorrectly drawn from "local (osx) fails, remote (linux) works" but  the other difference between "local" and "remote" was going through `open_ipc_subprocess` at all

python_modules/dagster/dagster/_serdes/ipc.py
```diff
@@ -211,7 +211,13 @@ def open_ipc_subprocess(parts: Sequence[str], **kwargs: object) -> Popen[bytes]:
     else:
         # Works on all UNIX systems (but not WASI, see: https://docs.python.org/3/library/os.html#os.setpgrp)
         preexec_fn = os.setpgrp
-    return subprocess.Popen(parts, creationflags=creationflags, preexec_fn=preexec_fn, **kwargs)
+    return subprocess.Popen(
+        parts,
+        creationflags=creationflags,
+        preexec_fn=preexec_fn,
+        stdin=subprocess.DEVNULL,  # Prevent terminal signals from hanging the subprocess
```
alangenfeld, line 218: i think this is safe

alangenfeld (approved):
i think we can drop the OSX specific language but i think this is legit

PR #13104, opened 2023-03-22
[resource verification][1/n] Add ConfigVerifiable interface for resources

benpankow:
## Summary

Adds a new `ConfigVerifiable` interface which resources can implement so that they can be validated in the Dagster UI (and perhaps at code location load time). Subsequent PRs will interface with this.

This is meant as an experimental proof of concept API, something we can ship and let users toy with to proof out the usefulness of verification.

## Test Plan

n/a

PR #13105, opened 2023-03-22
[resource verification][2/n] Generate verification jobs, GRPC call for ConfigVerifiable checks

benpankow:
## Summary

Generates a set of config verification jobs for any top-level resources which implement `ConfigVerifiable` from #13104. These jobs will be hidden from the UI (in a similar vein to the all assets job).

These jobs are a thin wrapper around the `verify_config` method on the resource. They're executed in-process on the GRPC server as a result of a GRPC call. The outputs from the job are stored in the event log, which is used as the source-of-truth and log of past verification checks.

## Test Plan

Added tests for verification jobs, GRPC calls.


python_modules/dagster/dagster/_grpc/server.py
```diff
@@ -820,16 +829,54 @@ def GetCurrentImage(self, request, _context):
...
+        )
+
+        try:
+            resource_verification_request = deserialize_value(
+                request.serialized_resource_verification_request, ResourceVerificationRequest
+            )
+
+            definition = self._get_repo_for_origin(resource_verification_request.repository_origin)
+
+            pipeline = cast(
```
dpeng817, line 855: nit: better name for this variable pls

python_modules/dagster/dagster_tests/core_tests/resource_tests/verifiable_resources/test_verifiable_resource.py
```diff
@@ -0,0 +1,119 @@
...
+from dagster._config.structured_config.resource_verification import (
+    resource_verification_job_name,
+    resource_verification_op_name,
+)
+from dagster._core.test_utils import environ
+
+
+def test_config_verifiable_job_basic() -> None:
+    class MyVerifiableResource(ConfigurableResource, ConfigVerifiable):
+        a_str: str
```
dpeng817, line 20: nit: variables that describe function in test a bit better such as `failure_switch`

python_modules/dagster/dagster_tests/core_tests/resource_tests/verifiable_resources/test_grpc.py
```diff
@@ -0,0 +1,174 @@
...
+    jobs=[the_job],
+    resources={
+        "success_resource": MyVerifiableResource(a_str="foo"),
+        "failure_resource": MyVerifiableResource(a_str="bar"),
+        "exception_resource": MyUnverifiableResource(),
+    },
+)
+
+
+def test_resources(
```
dpeng817, line 125: Not sure if this is a reasonable scenario given the likely callsite: attempting to launch resource verification for a resource that can't be found

python_modules/dagster/dagster_tests/core_tests/resource_tests/verifiable_resources/test_grpc.py
```diff
@@ -0,0 +1,174 @@
...
+        yield workspace
+
+
+@pytest.fixture(name="external_repo", scope="module")
+def external_repo_fixture(workspace_context: WorkspaceProcessContext):
+    repo_loc = next(
+        iter(workspace_context.create_request_context().get_workspace_snapshot().values())
+    ).code_location
+    assert repo_loc
+    return repo_loc.get_repository(SINGLETON_REPOSITORY_NAME)
```
dpeng817, line 78: we should really have a shared fixture directory bc there's a bajillion of the same fixture everywhere

dpeng817 (approved):
Approach seems reasonable to me, at least for a first pass. A few comments for your consideration. 

It would also be good to detail other approaches for storing this information, since rn we're basically abusing the run storage

python_modules/dagster/dagster_tests/core_tests/resource_tests/verifiable_resources/test_grpc.py
dpeng817: (to be clear not in this PR - just a general comment)

benpankow:
Based on [this Slack thread](https://elementl-workspace.slack.com/archives/C04S3DKJDDG/p1679598465443999), moved to a generic GRPC endpoint that represents a generic user code execution which we can re-use for future instigator types etc. This is to prevent users from needing to upgrade their agent every time we add a new user-code-execution-on-GRPC-server type.

Still uses runs under the hood, but toying with using tick storage instead.

dpeng817 (approved):

benpankow:
> - With this implementation, should we be showing resource verification runs in the run history? I guess there's no harm in it.

Yes; I could put up a stacked PR that will hide it from the main runs view (if we want to go that route).

> - The intended feature will only ever be introspecting back a single verification right? So the cost of completely switching the implementation is low - you just re-request verification?

Yeah, the only other cost to switching would be the leftover verification runs reminaing in the run log, which I think is ok for an experimental feature.

python_modules/dagster/dagster/_grpc/types.py
```diff
@@ -732,3 +734,151 @@ def __new__(
...
+        check.inst(self.data, unpack_class)
+        check.invariant(
+            unpack_class in REQUEST_CLASS_TO_EXECUTION_TYPE,
+            f"Unpack class {unpack_class} does not match any known request type",
+        )
+        check.invariant(
+            REQUEST_CLASS_TO_EXECUTION_TYPE.get(unpack_class) == self.execution_type,
+            f"Unpack class {unpack_class} does not match request type {self.execution_type}",
+        )
+        return self.data
```
alangenfeld, line 848: i'm not sure we need this defensive unpack at this layer. I also am skeptical it would work as desired given its next to a formal enum that the future entries would be missing from causing a failure.

python_modules/dagster/dagster/_grpc/types.py
```diff
@@ -732,3 +734,151 @@ def __new__(
...
+
+@whitelist_for_serdes
+class UserCodeExecutionType(Enum):
+    """Represents the type of user code execution request.
+
+    This is used to determine and verify specific request and result types when unpacking the generic
+    UserCodeExecutionRequest and UserCodeExecutionResult types.
+    """
+
+    RESOURCE_VERIFICATION = "resource_verification"
```
alangenfeld, line 796: I am not confident that our enum forward compat story is great - so we may want to use a string or GRPC level enum

python_modules/dagster/dagster/_grpc/server.py
```diff
@@ -819,16 +827,50 @@ def GetCurrentImage(self, request, _context):
...
+                "Unsupported user code execution type:"
+                f" {user_code_execution_request.execution_type}"
+            )
+        return api_pb2.UserCodeExecutionResult(
+            serialized_user_code_execution_result=serialize_value(
+                UserCodeExecutionResult(
+                    execution_type=user_code_execution_request.execution_type, data=result
+                )
+            )
+        )
```
alangenfeld, line 872: some ideas to consider for this set-up 
 * Just unpack the request in a try/catch and if it fails send raise "unvailable on version ..." exception chained with the deser exception. This would handle both top level missing request types and any problems on inner contents (though those should be unlikely). 
 * Make these serdes tuples implement an interface like `execute(self, ...)` and just call that here. Need to figure out what exactly to pass in - maybe `LoadedRepostiories` instead of the whole server. That part can change over time so not super critical to get right. 
 
 Its possible we can get nice behavior by actually trying to the appropriate grpc / protobuf primitives but we have talked about moving away from GRPC at somepoint which remains quite feasible given our low investment in their abstractions.

alangenfeld (commented):
nice, thanks for taking the extra detour to try to create some future flexibility here 

We should add a test emulating future `UserCodeExecutionRequest` to ensure we have the reasonable behavior we want. 

Agent wise, I don't think we need much any additional defense at this layer. We will end up with 
```
request_api: DagsterCloudApi.USER_CODE_EXECUTION
request_args: 
  serialized_user_code_execution_result: ... # agent does not touch
```

and then passing back `serialized_user_code_execution_result` without touching it so we should be good.

gibsondan (commented):
I have no context on the all assets job and still digesting this PR, but it is surprising to me that this is implemented as a hidden job.

The main benefit there is taking advantage of run storage and event log storage right? I feel like this has a pretty different use case / query pattern / potential storage optimization for this use case.

What kind of volume do you expect these jobs will be hit by? Do we need to be really careful to e.g. exclude the runtime from these jobs from cost calculations in cloud? These are the types of things that worry me

gibsondan:
What's a good way to get resolution on whether building the data here on top of runs/event logs is the right direction even for an MVP? Is that still the plan of record here?

benpankow:
> What's a good way to get resolution on whether building the data here on top of runs/event logs is the right direction even for an MVP? Is that still the plan of record here?

@schrockn and I spoke offline and we feel better about implementing this using the tick model. I'm working on porting it over right now.

schrockn (changes requested):
needs to be rebased

PR #13109, opened 2023-03-23
[ui] Use DefaultLogLevels when there is no level state stored

hellendag:
## Summary & Motivation

In the recent change to `useQueryPersistedLogFilter`, I didn't correctly check the empty state of the persisted log levels value. If it's undefined, we shouldn't treat that as an empty array -- we should treat it as falling back to the current `DefaultLogLevels` array.

## How I Tested These Changes

Clear the `EnabledRunLogLevels` localStorage key. Load a run, verify that the default five levels are selected.

Make a change in the level selector show only `event` logs, reload the run with the `levels` query parameter removed (to make sure it uses the persisted value). Verify that my change is reflected in the selector, and sanity check that the localStorage value has updated with my change as well.

Manually modify the `localStorage` value to an empty array, reload the run without the query parameter. Verify that the selector has zero levels selected, correctly matching the empty array.

Manually modify the `localStorage` value to `["EVENT"]`, reload the run. Verify that the selector loads with just `event` selected.

### DIFF
--- a/js_modules/dagit/packages/core/src/runs/useQueryPersistedLogFilter.test.tsx
+++ b/js_modules/dagit/packages/core/src/runs/useQueryPersistedLogFilter.test.tsx
@@ -1,8 +1,19 @@
+import {render} from '@testing-library/react';
+import * as React from 'react';
+
+import {useQueryPersistedState} from '../hooks/useQueryPersistedState';
+
 import {
   DefaultQuerystring,
   decodeRunPageFilters,
   encodeRunPageFilters,
+  useQueryPersistedLogFilter,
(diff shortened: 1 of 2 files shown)


hellendag:
* master
  * **PR #13109** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13109" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13109?utm_source=stack-comment).

bengotow (approved):
This looks good to me! Nice reminder that we should check the cleared-localstorage-state when we test :(

PR #13110, opened 2023-03-23
fix: AttributeError: 'coroutine' object has no attribute 'token'

mpicard:
## Summary & Motivation

Fixes a runtime error:  `"AttributeError: 'coroutine' object has no attribute 'token' "` caused by using the async DefaultAzureCredential instead of the synchronous version, which the DataLakeServiceClient and BlobServiceClient expect.

## How I Tested These Changes

- unit tested
- e2e tested in a deployed dagster instance in Azure


jamiedemaria:
is there anyway to write a unit test for this? if not, that's fine!

mpicard:
I'm currently looking into the unit testing, and I found an issue with the resource config and I need an opinion from you folks.

```python
ADLS2_CLIENT_CONFIG = {
    "storage_account": Field(StringSource, description="The storage account name."),
    "credential": Field(
        Selector(
            {
                "sas": Field(StringSource, description="SAS token for the account."),
                "key": Field(
                    StringSource, description="Shared Access Key for the account."
                ),
                "DefaultAzureCredential": Field(
                    Permissive(description="kwargs passed to DefaultAzureCredential()"),
                    description="Uses DefaultAzureCredential to authenticate and can pass keyword arguments",
                    default_value={},
                ),
            }
        ),
        description="The credentials with which to authenticate.",
    ),
}
```
The issue is that the Selector only allows for one option, but `sas` and `key` need to be together, so either I nest them in a Permissive or refactor the config (breaking change?) into something else entirely, maybe make the `sas`, `key` and `DefaultAzureCredential` all optional and handle the checks in the resource, but I don't know if this is desirable from a Dagster point of view, i.e. is it better to use Selector or not? What are your opinions on this?

mpicard:
I spend several hours trying to find a way to unit test it but it's very difficult to mock the azure libraries, the `get_token` for the `DefaultAzureCredential` is buried really deep in the azure libraries call stack and it seems very difficult to mock/stop the HTTP requests from actually happening but not mock away too much so that the `get_token` gets called.

PR #13118, opened 2023-03-23
[dagit] reload/redeploy buton UX

alangenfeld:
* expose `mutating` bool
* fetch `loadStatus` and remove `no-cache` setting to allow mutation to update apollo cache

related internal change https://github.com/dagster-io/internal/pull/5275

## How I Tested These Changes

use reload button via dev_webapp - no meaningful change since mutation blocks on reload


js_modules/dagit/packages/core/src/nav/useRepositoryLocationReload.tsx
```diff
@@ -314,7 +315,6 @@ export const buildReloadFnForLocation = (location: string) => {
     >({
       mutation: RELOAD_REPOSITORY_LOCATION_MUTATION,
       variables: {location},
-      fetchPolicy: 'no-cache',
```
alangenfeld, line 317: couldn't figure out the motivation for this from the blame rev @bengotow

hellendag: This seems like a reasonable change to make now that there is a value to be cached, since before it was just `id`. But I also don't know/remember the motivation for `no-cache` here.

js_modules/dagit/packages/core/src/nav/useRepositoryLocationReload.tsx
```diff
@@ -208,11 +208,12 @@ export const useRepositoryLocationReload = ({
   const {mutating, pollStartTime, error, errorLocationId} = state;
   const reloading = mutating || pollStartTime !== null;
 
-  return React.useMemo(() => ({reloading, error, errorLocationId, tryReload}), [
+  return React.useMemo(() => ({reloading, error, errorLocationId, tryReload, mutating}), [
```
hellendag, line 211: Is the `reloading` value insufficient?

alangenfeld: basically i tried that and didn't get the behavior i expeced, it seems like `pollStartTime` doesnt clear when I would expect it to. I think there are corner cases where it may be valid to submit another redeploy request while we are in the polling state so i didnt chase it down further and opted to just expose when the mutation request is inflight.

js_modules/dagit/packages/core/src/nav/useRepositoryLocationReload.tsx
bengotow: Hey gang, I'm afraid I don't remember this change either. Based on my understanding of how this fetchPolicy stuff works, this was preventing apollo from putting the result data into the cache, but doing that should also be fine. 👍

bengotow (approved):

PR #13120, opened 2023-03-23
enable setting freshness policies when loading assets

sryza:
## Summary & Motivation

I was working on moving project_fully_featured to declarative scheduling and found myself wanting this.

In service of this:
- Renames `AssetsDefinition. with_prefix_or_group` to `AssetsDefinition.with_attributes` to accommodate more attributes
- Refactors some common code in `load_assets_from_modules` into a shared function.


sryza:
ping for review @bengotow @OwenKephart

OwenKephart (approved):
nice, we'll want similar stuff for AutoMaterializationPolicies, so swapping the name now is great

johannkm:
Do we also need a with_auto_materialize_policy? I don't see an equivalent for freshness

PR #13121, opened 2023-03-23
Restore asset key null check

smackesey:
## Summary & Motivation

`AssetMaterialization` has inadvertently been allowing `AssetKey` to be set to `None` for several months. This PR restores the runtime check to prevent this while maintaining compatibility with stored `AssetMaterialization` (or older `Materialization`) objects that lack an `AssetKey`.

See: https://github.com/dagster-io/dagster/pull/13030 for further context.

## How I Tested These Changes

Modified `test_load_old_materialization` unit test.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/events.py
+++ b/python_modules/dagster/dagster/_core/definitions/events.py
@@ -22,6 +22,7 @@ from dagster._annotations import PublicAttr, public
 from dagster._core.definitions.data_version import DataVersion
 from dagster._core.storage.tags import MULTIDIMENSIONAL_PARTITION_PREFIX, SYSTEM_TAG_PREFIX
 from dagster._serdes import whitelist_for_serdes
+from dagster._serdes.serdes import NamedTupleSerializer
 from dagster._utils.backcompat import experimental_class_param_warning
 
 from .metadata import (
@@ -378,7 +379,7 @@ class AssetObservation(
 
     def __new__(
         cls,
-        asset_key: Union[Sequence[str], AssetKey, str],
+        asset_key: CoercibleToAssetKey,
         description: Optional[str] = None,
         metadata_entries: Optional[Sequence[MetadataEntry]] = None,
         partition: Optional[str] = None,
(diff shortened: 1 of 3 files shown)


smackesey:
* master
  * **PR #13121** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13121" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13121?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/events.py
```diff
@@ -421,7 +422,23 @@ def label(self) -> str:
         return " ".join(self.asset_key.path)
 
 
-@whitelist_for_serdes(old_storage_names={"Materialization"})
+UNDEFINED_ASSET_KEY = AssetKey(["undefined"])
```
alangenfeld, line 425: should this be more obnoxious ie `_undefined_` / `__undefined__` / `<<undefined>>` ? 
 
unclear how many of these flowed through the system so not sure this is worth fretting about at all

alangenfeld (commented):
add a null asset_key fails test

python_modules/dagster/dagster/_core/definitions/events.py
smackesey: I'm partial to  `__undefined__`, why not

python_modules/dagster/dagster/_core/definitions/events.py
```diff
@@ -421,7 +416,23 @@ def label(self) -> str:
...
+
+
+class AssetMaterializationSerializer(NamedTupleSerializer):
+    # There are old `Materialization` objects in storage. We set the default value for asset key to be
+    # `AssetKey(["undefined"])` to ensure that we can load these objects, without needing to allow for
+    # the construction of new `AssetMaterialization` objects with no defined AssetKey.
+    def before_unpack(self, **raw_dict: Any) -> Any:
+        # cover both the case where "asset_key" is not present at all and where it is None
+        if raw_dict.get("asset_key") is None:
+            raw_dict["asset_key"] = {"__class__": "AssetKey", "path": ["undefined"]}
```
alangenfeld, line 429: this is out of sync now right?  maybe const the string instead of the obj

PR #13130, opened 2023-03-23
[asset-reconciliation] Fix issue with duplicate runs for partitions with in-progress materializations

OwenKephart:
## Summary & Motivation

When we notice a new materialization of an (asset/partition), this often means that we'll need to propagate this change to its children (asset/partitions). However, if its child is set to be materialized in the same run that just materialized the parent, then there's no need for the sensor to kick off an additional run.

In the unpartitioned case, we were already doing this check. However, for the partitioned case, we were simply getting the set of partitions that had been materialized since the last tick (since we can do that with a single query), then adding the children of all those partitions to our set without checking if a run was already set to materialize them.

It's rather unfortunate, but this change means that in the case that there are many partitions for a given asset that are materialized in between two ticks of the sensor, then many individual queries will need to be fired to get the most recent materialization record for each of these newly-materialized partitions.

I was able to filter out some cases where we know that the asset can't be in the same run (such as when the partition key doesn't match), which should help a fair amount. Because the sensor will generally run fairly regularly, it seems fairly unlikely that more than a few partitions for a given asset will land in the ~30 seconds between sensor ticks, so in most cases this will not be a huge deal.

The scariest cases are the ones where many partitions can accumulate, namely:

a) when the sensor is first turned on -- I hack around this by disabling this check for the first tick of the sensor (i.e. when latest_record_id is None). I think this is an acceptable compromise
b) if the sensor is turned off for awhile, then turned back on. there's no easy way around this, other than perhaps keeping track of the last tick time in the cursor, and skipping this logic if it's been more than, say, 10 minutes since the last tick.

A future ability to query for the most recent materialization record for a set of (asset_key, partition_keys) would also get around this issue, as we could get the set of records we care about with a single query, rather than a bunch.

## How I Tested These Changes

Added test failed before, passes now


sryza (commented):
`latest_storage_id is not None` seems a little sketchy to me - there could be ongoing runs when this is switched on, right?

What would be the consequence of deferring to this check to later. Presumably the stuff we do inside `is_reconciled` is at least as expensive? But it's not as much of a concern because we've already done some filtering?

OwenKephart:
> What would be the consequence of deferring to this check to later?

This is a good idea, we can do this inside of [should_reconcile](https://sourcegraph.com/github.com/dagster-io/dagster/-/blob/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py?L457:5), which currently can filter out a good chunk of cases for us. However, in the (eventual) case that you don't have any partition filtering set up (i.e. you're not limiting yourself to just the last partition or last N partitions), then none of these partitions will be filtered.


> Presumably the stuff we do inside is_reconciled is at least as expensive? But it's not as much of a concern because we've already done some filtering?

This was true up until recently, but no longer is. `is_reconciled` no longer needs to make a query for the most recent materialization record for each examined partition, as it can rely on a shared query of `get_materialization_count_by_partition`.

> latest_storage_id is not None seems a little sketchy to me - there could be ongoing runs when this is switched on, right?

absolutely true, this was an explicit choice to have slightly incorrect behavior, rather than make it impossible for the reconciliation sensor to start for some cases. I think for now, pushing this into should_reconcile function should duck the immediate issue, but once we allow users to target all partitions of their assets, we'll need to be a little fancier to make this performant for the first tick.

sryza (approved):

PR #13132, opened 2023-03-23
CoercibleToAssetSelection

sryza:
## Summary & Motivation

`define_asset_job` has some fancy code for coercing a variety of arguments into `AssetSelection`.

Factor this out so it can be used elsewhere.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_selection.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_selection.py
@@ -1,18 +1,33 @@
 import operator
 from abc import ABC, abstractmethod
 from functools import reduce
-from typing import AbstractSet, Iterable, Optional, Union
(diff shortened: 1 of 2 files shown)


OwenKephart (approved):
nice

python_modules/dagster/dagster/_core/definitions/asset_selection.py
```diff
@@ -1,19 +1,32 @@
...
+    parse_clause,
+)
 from dagster._utils.backcompat import deprecation_warning
 
 from .asset_graph import AssetGraph
 from .assets import AssetsDefinition
 from .events import AssetKey, CoercibleToAssetKey
 from .source_asset import SourceAsset
 
+CoercibleToAssetSelection = Union[
```
smackesey, line 22: `TypeAlias`

smackesey (approved):
one comment but looks good

PR #13133, opened 2023-03-23
add selection argument to materialize and materialize_to_memory

sryza:
## Summary & Motivation

Before:
```python
materialize([*upstream.to_source_assets(), downstream])
```

After:
```python
materialize([upstream, downstream], selection=[downstream])
```

While the after state doesn't have fewer characters, I think it's much easier to understand.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/materialize.py
+++ b/python_modules/dagster/dagster/_core/definitions/materialize.py
@@ -13,6 +13,8 @@ from .assets import AssetsDefinition
 from .source_asset import SourceAsset
 
 if TYPE_CHECKING:
+    from dagster._core.definitions.asset_selection import CoercibleToAssetSelection
+
     from ..execution.execute_in_process_result import ExecuteInProcessResult
 
 
@@ -24,6 +26,7 @@ def materialize(
     partition_key: Optional[str] = None,
     raise_on_error: bool = True,
     tags: Optional[Mapping[str, str]] = None,
+    selection: Optional["CoercibleToAssetSelection"] = None,
 ) -> "ExecuteInProcessResult":
     """Executes a single-threaded, in-process run which materializes provided assets.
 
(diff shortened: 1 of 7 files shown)


python_modules/dagster/dagster/_core/definitions/materialize.py
```diff
@@ -41,6 +44,17 @@ def materialize(
...
+            A sub-selection of assets to materialize.
+
+            The selected assets must all be included in the assets that are passed to the assets
+            argument.
+
+            If not provided, then all assets will be materialized.
+
+            The string "my_asset*" selects my_asset and all downstream assets within the code
+            location. A list of strings represents the union of all assets selected by strings
+            within the list.
```
smackesey, line 57: This is user-facing and IIUC solves the annoying problem of needing to provide source asset versions of assets you don't want to materialize in `assets`. Think it would benefit from an example to be crystal clear.

sryza: > Seems odd to reference just this one part of the selection syntax. Can we provide a link to a comprehensive description of it?

This is copy/pasted from the selection argument of `define_asset_job`. But yeah this is kind of weird.

I pushed a change that adds an example and fixes up the docs.

sryza:
ping for review @dpeng817 @smackesey

python_modules/dagster/dagster/_core/definitions/materialize.py
```diff
@@ -59,7 +95,11 @@ def materialize(
 
     JOB_NAME = "__ephemeral_asset_job__"
 
-    defs = Definitions(jobs=[define_asset_job(name=JOB_NAME)], assets=assets, resources=resources)
+    defs = Definitions(
+        jobs=[define_asset_job(name=JOB_NAME, selection=selection)],
```
smackesey, line 99: Do we need the `define_asset_job` given `Definitions.get_implicit_global_asset_job_def`?

sryza: We could find a way to work around it, but it makes things easier, because otherwise we would need to:
- Find the right implicit asset job (there can be multiple if different assets have different partitions)
- Coerce the asset selection selection

Open to changing it, but seems orthogonal to what's going on in this PR.

smackesey: true, I just noticed bc the line was green but now I realize it was already there

smackesey (approved):

Issue #13077, opened 2023-03-22
`context.partition_time_window` when MultiPartitionsDefinition


rubenbriones:
### Dagster version

1.2.1

### What's the issue?

The property `context.partition_time_window` is not working when `MultiPartitionsDefinition` is used, even though one of the partitions is of type `DailyPartitionsDefinition`.

### What did you expect to happen?

I would expect that the properties `partition_time_window.start` and `partition_time_window.end` can be accessed for the Daily partition.

It would be great if dagster is able to automatically recognize the date partition, and return its _time_window_. Or if not, that something like that could be done: `context.get_partition_by_key('date').partition_time_window`

### How to reproduce?

```py
@asset(
    partitions_def=MultiPartitionsDefinition(
        {
            "country": StaticPartitionsDefinition(["USA", "ESP"]),
            "date": DailyPartitionsDefinition(start_date="20230317", fmt='%Y%m%d'),
        }
    ),
)
def foo(context):
    date_requested = context.partition_time_window.start.strftime("%Y%m%d")
    print(date_requested)
```

### Deployment type

None

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

jamiedemaria:
thanks for the report! @clairelin135  would you or sandy be the right person to take a look at this?

clairelin135:
I'll take a look at this

PR #13137, opened 2023-03-23
Enable context partition time window methods for multipartitioned runs

clairelin135:
Fixes https://github.com/dagster-io/dagster/issues/13077

Makes the following context methods work for multipartitioned runs:
- context.partition_time_window
- context.asset_partitions_time_window_for_output

Not sure why we have both of these methods considering that they do the same thing (until we support per-output partitioning). For now, this PR continues to support both methods.

Also enables these methods to work on direct invocation too.

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/context/compute.py
+++ b/python_modules/dagster/dagster/_core/execution/context/compute.py
@@ -398,7 +398,8 @@ class OpExecutionContext(AbstractComputeExecutionContext):
 
         Raises an error if either of the following are true:
         - The output asset has no partitioning.
-        - The output asset is not partitioned with a TimeWindowPartitionsDefinition.
+        - The output asset is not partitioned with a TimeWindowPartitionsDefinition or a
+          MultiPartitionsDefinition with one time-partitioned dimension.
         """
         return self._step_execution_context.asset_partitions_time_window_for_output(output_name)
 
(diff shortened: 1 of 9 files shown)


python_modules/dagster/dagster/_core/execution/context/compute.py
```diff
@@ -398,7 +398,7 @@ def asset_partitions_time_window_for_output(self, output_name: str = "result") -
 
         Raises an error if either of the following are true:
         - The output asset has no partitioning.
-        - The output asset is not partitioned with a TimeWindowPartitionsDefinition.
+        - The output asset is not partitioned with a TimeWindowPartitionsDefinition or a MultiPartitionsDefinition with one time-partitioned dimension.
```
sryza, line 401: Line length

python_modules/dagster/dagster/_core/execution/context/invocation.py
```diff
@@ -562,17 +564,24 @@ def partition_key(self) -> str:
...
         return self.partition_key
 
-    def asset_partitions_time_window_for_output(
-        self, output_name: str = "result"
-    ) -> Optional[TimeWindow]:
+    def asset_partitions_time_window_for_output(self, output_name: str = "result") -> TimeWindow:
+        return self.partition_time_window
+
+    @property
+    def partition_time_window(self) -> TimeWindow:
```
sryza, line 571: This name makes it seem like this would work when there are no assets. Is that intended? Is that necessary for this change?

clairelin135: ah... so in the execution context it does work for non-asset jobs. But in the direct invocation context it doesn't work for non-asset ops, since partitioned config is applied on the job level. I think we can remove this function from direct invocation and just ask users to use `asset_partitions_time_window_for_output` for the asset use case

sryza (approved):

PR #13143, opened 2023-03-24
celery-k8s executor: handle stdout that's valid json but not a dagster event

johannkm:
Celery-k8s step workers write json serialized events to stdout. The celery workers parse them. It seems like if you write a line to stdout that parses as json but isn't an event, we fail with a DeserializationError: https://dagster.slack.com/archives/C01U954MEER/p1679635251354799

https://github.com/dagster-io/dagster/pull/7665 added 

```
except JSONDecodeError:
            pass
```

in order to skip lines that don't parse as valid json. We can add DeserializationError to the errors to skip, so we just ignore lines that aren't events.

Of course ideally we should switch to reading events from the db but that's a larger change. I don't think there's too much risk of true errors getting swallowed by this?

---

Side note:

I'm not sure what the 

```
        except check.CheckError:
            pass
```

catches. Is it possible that before the `deserialize_json_to_dagster_namedtuple`, this would raise a CheckError instead of a DeserializationError? Doesn't look like it from some quick browsing but I am a bit surprised this hasn't come up before. Maybe bc it's rare to write directly to stdout?

---

Testing the test here: https://github.com/dagster-io/dagster/pull/13144

### DIFF
--- a/python_modules/dagster-test/dagster_test/test_project/test_pipelines/repo.py
+++ b/python_modules/dagster-test/dagster_test/test_project/test_pipelines/repo.py
@@ -273,6 +273,7 @@ def define_docker_celery_pipeline():
     @resource
     def resource_with_output():
         print("writing to stdout")  # noqa: T201
+        print("{}")  # noqa: T201
         return 42
 
     @op(required_resource_keys={"resource_with_output"})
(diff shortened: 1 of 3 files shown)


johannkm:
* master
  * **PR #13143** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13143" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13143?utm_source=stack-comment).

alangenfeld (approved):

PR #13146, opened 2023-03-24
Make sure we always wait for all grpc server processes to spin down in tests

gibsondan:
Summary:
Some of the grpc flakiness we are seeing on upgrade seems to stem from lingering grpc subprocesses still running after their test is finished. We previously only spun these down for launched runs due to issues with filesystem contention - this PR takes a step beyond that and ensures that any grpc server process that is opened during a test is closed before that test ends.

### DIFF
--- a/python_modules/dagster/dagster/_core/instance/config.py
+++ b/python_modules/dagster/dagster/_core/instance/config.py
@@ -295,7 +295,11 @@ def dagster_instance_config_schema() -> Mapping[str, Field]:
             }
         ),
         "code_servers": Field(
-            {"local_startup_timeout": Field(int, is_required=False)}, is_required=False
+            {
+                "local_startup_timeout": Field(int, is_required=False),
+                "wait_for_local_processes_on_shutdown": Field(bool, is_required=False),
+            },
+            is_required=False,
         ),
         "secrets": secrets_loader_config_schema(),
         "retention": retention_config_schema(),
--- a/python_modules/dagster/dagster/_core/instance_for_test.py
+++ b/python_modules/dagster/dagster/_core/instance_for_test.py
@@ -68,18 +68,14 @@ def instance_for_test(
         if not temp_dir:
             temp_dir = stack.enter_context(tempfile.TemporaryDirectory())
 
-        # If using the default run launcher, wait for any grpc processes that created runs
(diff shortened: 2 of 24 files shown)


python_modules/dagster/dagster/_grpc/server.py
```diff
@@ -1254,17 +1256,61 @@ def __init__(
...
+            warnings.warn(
+                "GrpcServerProcess is being destroyed without waiting for the process to "
+                + "fully terminate. This can cause test instability."
+            )
+
+            # Just for pr testing, do not land
+            traceback.print_stack()
+            from dagster._utils import segfault
+
+            segfault()
```
gibsondan, line 1311: i will be sure to take this out

alangenfeld: is warning right here or should we raise?

python_modules/dagster/dagster/_core/instance/config.py
```diff
@@ -301,4 +301,5 @@ def dagster_instance_config_schema() -> Mapping[str, Field]:
         "retention": retention_config_schema(),
         "sensors": sensors_daemon_config(),
         "schedules": schedules_daemon_config(),
+        "wait_for_processes": Field(bool, is_required=False, default_value=False),
```
alangenfeld, line 304: more specific name since its an instance level setting

python_modules/dagster/dagster/_core/instance_for_test.py
```diff
@@ -40,6 +40,7 @@ def instance_for_test(
     overrides: Optional[Mapping[str, Any]] = None,
     set_dagster_home: bool = True,
     temp_dir: Optional[str] = None,
+    wait_for_processes=True,
```
alangenfeld, line 43: is this arg ever set to False? should we even expose setting it?

alangenfeld (changes requested):

python_modules/dagster/dagster/_grpc/server.py
gibsondan: I don't think raising in _del_ really does anything - it essentially just warns anyway

alangenfeld (approved):

PR #13151, opened 2023-03-24
[caching-refactor] Remove use of get_and_update_asset_status_cache.

OwenKephart:
## Summary & Motivation

Previously, we only attempted to use this in cases where this method was available, now we don't use it at all. Also updated the method to actually check which keys were valid.

The updated test wasn't running before because it was in a folder instead of a module. So it actually needed to be updated to pass as it was executing refactored functions.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/data_time.py
+++ b/python_modules/dagster/dagster/_core/definitions/data_time.py
@@ -56,7 +56,6 @@ class CachingDataTimeResolver:
         beginning of that time window, or all data up until the end of the first filled time window
         in the total set, whichever is less.
         """
-
         # the total set of materialized partitions
         partition_subset = partitions_def.empty_subset().with_partition_keys(
             partition_key
--- a/python_modules/dagster/dagster/_core/definitions/data_time.py
+++ b/python_modules/dagster/dagster/_core/definitions/data_time.py
@@ -45,8 +45,7 @@ class CachingDataTimeResolver:
         At a high level, this algorithm works as follows:
 
         First, calculate the subset of partitions that have been materialized up until this point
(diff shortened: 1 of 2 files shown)


clairelin135 (approved):
Yay! This means that the updated cached status data endpoint is no longer needed in the graphQL event log storage endpoint, right? Would be great to remove that in a followup

python_modules/dagster/dagster/_core/definitions/data_time.py
```diff
@@ -57,31 +56,13 @@ def _calculate_data_time_partitioned(
         beginning of that time window, or all data up until the end of the first filled time window
         in the total set, whichever is less.
         """
-        from dagster._core.storage.partition_status_cache import (
-            get_and_update_asset_status_cache_value,
+        # the total set of materialized partitions
+        partition_subset = partitions_def.empty_subset().with_partition_keys(
```
clairelin135, line 60: In the unpartitioned asset case, this could be called once for each upstream partition right? Can we cache the data in this case to avoid recalculating?

OwenKephart: Are you thinking about the case where an unpartitioned is downstream of a time-partitioned asset? In this codepath, that will still only result in a single call to `_calculate_data_time_partitioned`, as the data time of a partitioned asset is a property of the entire asset, rather than a property of each partition of that asset.

So in general, in this part of the world, the number of queries does not scale with the number of partitions in the graph

OwenKephart:
Performance hit should be fine, although we could think about being a bit more efficient in here to reduce the amount of time spent in cron string iterators.

PR #13156, opened 2023-03-24
Fix case where a pipeline load takes long enough that run monitoring kills it

gibsondan:
Summary:
Fixes two issues, one big and one small:
- pipeline_run was incorrectly renamed to dagster_run in https://github.com/dagster-io/dagster/pull/9095 without changing the rest of the callsites, negating the point of reloading it (also unclear why we were fetching it twice)
- Swap around the ordering of failure messages when a run comes in that's already in a failure state - the "giving up" message was incorrectly overly tailored to the case where a running job is restarted, but the case where the run is already finished and we should just give up should take precedence

Test Plan:
launch a run locally that sleeps for 15 seconds on code load, with run monitoring set to 5 seconds
Before: run would keep going once it started up, ignoring that it's now in a failure state
Now: Error message that says "Ignoring a run worker that started after the run had already finished."

Looking into an automated test now as well, given that this is regressed once already likely worth investing in

### DIFF
--- a/python_modules/dagster/dagster_tests/execution_tests/test_api_iterators.py
+++ b/python_modules/dagster/dagster_tests/execution_tests/test_api_iterators.py
@@ -1,11 +1,17 @@
 import pytest
 from dagster import (
+    DagsterEventType,
     _check as check,
+    job,
+    reconstructable,
     resource,
 )
 from dagster._core.definitions import op
 from dagster._core.definitions.pipeline_base import InMemoryPipeline
 from dagster._core.errors import DagsterInvariantViolationError
+from dagster._core.events import (
+    DagsterEvent,
+)
 from dagster._core.events.log import EventLogEntry, construct_event_logger
 from dagster._core.execution.api import (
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster_tests/execution_tests/test_api_iterators.py
```diff
@@ -145,6 +145,12 @@ def event_callback(record):
                 "run_monitoring": {"enabled": True},
             }
         ) as run_monitoring_instance:
+            pipeline_run = instance.create_run_for_pipeline(
+                pipeline_def=pipeline_def,
+                run_config={"loggers": {"callback": {}}},
+                mode="default",
+            ).with_status(DagsterRunStatus.CANCELING)
```
alangenfeld, line 152: this test change would have prevented the accidental rename?

gibsondan:
No I’ll add a better one now that would have

alangenfeld (approved):

PR #13162, opened 2023-03-24
[long-running runs] Move run canceling reporting into callsites of run termination

dpeng817:
Previously, it was up to the run launcher to report a run as canceling when terminated. This makes it technically possible to terminate a run without reporting it as canceled, if the run launcher implementation does not do so. Moving the cancelation reporting to the callsites makes this less likely, and easier to fix. 

One potentially controversial result of this change is that when `terminate` is called, we now expect the state of the run to already be `CANCELING`. So needed to change the implementation of `can_terminate`. `can_terminate` isn't on the base run launcher class, and is only used within the context of the `terminate` method of K8s and CeleryK8s run launchers, which seem reasonable.

### DIFF
--- a/python_modules/dagster/dagster/_core/run_coordinator/default_run_coordinator.py
+++ b/python_modules/dagster/dagster/_core/run_coordinator/default_run_coordinator.py
@@ -53,7 +53,5 @@ class DefaultRunCoordinator(RunCoordinator, ConfigurableClass):
         run = self._instance.get_run_by_id(run_id)
         if run is None:
             check.failed(f"Failed to load run {run_id}")
-        self._instance.report_run_canceling(
-            run, message="Received cancellation request from default run coordinator"
-        )
+        self._instance.report_run_canceling(run, message="Received cancellation request.")
         return self._instance.run_launcher.terminate(run_id)
--- a/python_modules/dagster/dagster/_core/run_coordinator/default_run_coordinator.py
+++ b/python_modules/dagster/dagster/_core/run_coordinator/default_run_coordinator.py
@@ -50,4 +50,10 @@ class DefaultRunCoordinator(RunCoordinator, ConfigurableClass):
         return run
 
     def cancel_run(self, run_id: str) -> bool:
+        run = self._instance.get_run_by_id(run_id)
+        if run is None:
+            check.failed(f"Failed to load run {run_id}")
+        self._instance.report_run_canceling(
+            run, message="Received cancellation request from default run coordinator"
+        )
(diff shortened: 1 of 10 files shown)


dpeng817:
* master
  * **PR #13162** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13162" target="_blank">Graphite</a>  👈
    * **PR #12920** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12920" target="_blank">Graphite</a> 
      * **PR #13072** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13072" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13162?utm_source=stack-comment).

smackesey (approved):
I'm not particularly knowledgable on run launchers/coordinators, so I wouldn't be able to identify any subtler issues, but this idea makes sense and the code looks good.

dpeng817:
@smackesey I think I added you by accident - since this is important behavior change I definitely want @gibsondan or @johannkm 's eyes on this, since they know this code intimately

python_modules/dagster/dagster/_core/run_coordinator/default_run_coordinator.py
```diff
@@ -50,4 +50,10 @@ def submit_run(self, context: SubmitRunContext) -> DagsterRun:
         return run
 
     def cancel_run(self, run_id: str) -> bool:
+        run = self._instance.get_run_by_id(run_id)
+        if run is None:
+            check.failed(f"Failed to load run {run_id}")
+        self._instance.report_run_canceling(
+            run, message="Received cancellation request from default run coordinator"
```
gibsondan, line 57: I would suggest dropping "from default run coordinator" - don't think a user needs to understand what that is

python_modules/dagster/dagster/_core/run_coordinator/queued_run_coordinator.py
```diff
@@ -258,4 +258,7 @@ def cancel_run(self, run_id: str) -> bool:
             self._instance.report_run_canceled(run)
             return True
         else:
+            self._instance.report_run_canceling(
+                run, message="Received cancellation request from queued run coordinator."
```
gibsondan, line 262: same. How about just 'Received cancellation request."

gibsondan (approved):

dpeng817:
As it turns out - this introduces a breaking change by altering the state before termination. Closing.

PR #13169, opened 2023-03-26
[refactor] Misc composition and config refactors/renames

smackesey:
## Summary & Motivation

Type annotations and some light refactoring/fixing of type errors related to composition and config.

- Make `PendingNodeInvocation` generic in `node_def`
- Eliminate `ensure_graph_def` and `ensure_node_def` (these functions are fulfilled by previously added `Node` subclasses `OpNode` and `GraphNode` coupled with generic `PendingNodeInvocation`)
- Condense some logic used to generate modified `PendingNodeInvocation` in `NodeDefinition`.
- Type annotations for many composition-related functions
- Type annotations for config-mapping/composite-descent functions
- Light refactor some composite descent logic to fix surfaced type errors (use dummy root `NodeHandle`)
- Local variable/arg renames (solid -> op/node/graph) throughout composition and composite descent code

## How I Tested These Changes

Existing test suite

PR #13168, opened 2023-03-26
[ruff] Enable UP032 (f-string)

smackesey:
## Summary & Motivation

Enable ruff `UP032` (`f-string`). This is a nice rule that auto-converts `string.format()` to f-strings conservatively (it won't flag convert string format calls where the args are long, i.e. where `string.format()` actually makes sense).

## How I Tested These Changes

Existing test suite.

PR #13145, opened 2023-03-24
[refactor] Remove PartitionSetDefinition

smackesey:
## Summary & Motivation

Delete `PartitionSetDefinition`.

- Refactor `JobDefinition` internals so that it no longer relies on an internally generated `PartitionSetDefinition` to launch partitioned runs.
- Still (temporarily) use `ExternalPartitionSetDefinition` in the external representation of a repo, but construct the `ExternalPartitionSetDefinition` objects directly from job definitions instead of needing to go through `PartitionSetDefinition`.

## How I Tested These Changes

Existing test suite.

PR #13108, opened 2023-03-23
[refactor] *Solid*Snap renames

smackesey:
## Summary & Motivation

Rename (mostly) `*Snap` classes and fields to eliminate "solid".  Where classes are serdes-whitelisted, use `storage_name`/`storage_field_names` to keep the serialized name the same for backcompat. Also includes local variable renames.

- assorted local variables
- DependencyStructureIndex.solid_names -> node_names
- DependencyStructureIndex.solid_invocation_names -> node_invocation_names
- DependencyStructureIndex.solid_invocations -> node_invocations
- DependencyStructureSnapshot.solid_invocation_snaps -> node_invocation_snaps
- PipelineSnapshot.solid_defs_snapshot -> node_defs_snapshot
- PipelineSnapshot.solid_names_in_topological_order -> node_names_in_topological_order
- PipelineLineageSnapshot.solid_selection -> node_selection
- PipelineLineageSnapshot.solids_to_execute -> nodes_to_execute
- PipelineSnaphot.solid_names -> node_names
- PipelineSnaphot.solid_definitions_snapshot -> node_defs_snapshot
- build_solid_invocation_snap -> build_node_invocation_snap
- InputHandle.solid_def_name -> node_def_name
- InputHandle.solid_name -> node_name
- OutputHandleSnap.solid_name -> node_name
- OutputHandleSnap.solid_name -> node_name
- CompleteCompositionContext.solid_defs -> node_defs
- DynamicFanIn.solid_name -> node_name
- InvokedNodeOutputHandle.solid_name -> node_name
- InvokedNodeDynamicOutputWrapper.solid_name -> node_name
- _core.snap.solid -> node (module)
- _core.definitions.solid_invocation -> op_invocation (module)
- build_dep_structure_snapshot_from_icontains_solids -> build_dep_structure_snapshot_from_graph_def
- OutputMappingSnap.mapped_solid_name -> mapped_node_name
- InputMappingSnap.mapped_solid_name -> mapped_node_name
- build_core_solid_def_snap -> build_op_def_snap
- build_composite_solid_def_snap -> build_graph_def_snap
- build_solid_definitions_snapshot -> build_node_defs_snap
- SolidDefinitionsSnapshot -> NodeDefsSnapshot
- CompositeSolidDefSnap -> GraphDefSnap
- SolidDefSnap -> OpDefSnap
- NodeInvocationSnap.solid_def_name -> node_def_name
- NodeInvocationSnap.solid_name -> node_name
[...]

PR #13171, opened 2023-03-26
[refactor] rename X.solid_handle -> node_handle

smackesey:


smackesey:
* master
  * **PR #13145** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13145" target="_blank">Graphite</a> 
    * **PR #13108** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13108" target="_blank">Graphite</a> 
      * **PR #13168** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13168" target="_blank">Graphite</a> 
        * **PR #13169** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13169" target="_blank">Graphite</a> 
          * **PR #13081** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13081" target="_blank">Graphite</a> 
            * **PR #13083** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13083" target="_blank">Graphite</a> 
            * **PR #12770** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12770" target="_blank">Graphite</a> 
              * **PR #13170** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13170" target="_blank">Graphite</a> 
                * **PR #13171** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13171" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13171?utm_source=stack-comment).

python_modules/dagster-graphql/dagster_graphql/implementation/loader.py
```diff
@@ -428,7 +428,7 @@ def get_sink_asset(self, asset_key: AssetKey) -> ExternalAssetNode:
 
     def get_cross_repo_dependent_assets(
         self, repository_location_name: str, repository_name: str, asset_key: AssetKey
-    ) -> List[ExternalAssetDependedBy]:
+    ) -> Sequence[ExternalAssetDependedBy]:
```
dpeng817, line 431: I feel like a lot of this stuff is unrelated?

smackesey: Sorry you caught it when I was in the middle of fixing my screwed up graphite state so it was reflecting downstack PRs-- now updated

python_modules/dagster/dagster/_core/execution/plan/inputs.py
```diff
@@ -391,15 +391,15 @@ def required_resource_keys(self, pipeline_def: PipelineDefinition) -> Set[str]:
...
+@whitelist_for_serdes(storage_field_names={"node_handle": "solid_handle"})
 class FromStepOutput(
     NamedTuple(
         "_FromStepOutput",
         [
             ("step_output_handle", StepOutputHandle),
             ("fan_in", bool),
             # deprecated, preserved for back-compat
-            ("solid_handle", NodeHandle),
+            ("node_handle", NodeHandle),
```
dpeng817, line 402: do we move this out of the deprecated column? Unclear

dpeng817 (commented):
This looks reasonable to me, only question would be putting some of these serdes renames under test - @alangenfeld are there any tests we should add regarding these backcompat serdes objects?

smackesey:
>only question would be putting some of these serdes renames under test

Some significant subset of them (possibly all) are already under test via snapshots

smackesey:
I investigated the deprecated `soild_handle` (or `node_handle` in this PR) fields in `StepInputSource` and found that they are required for backcompat (specifically when user code is running old dagster version). I don't fully understand what's going on, since they appear to be filled with dummy that aren't used in the current codebase, but I think what may be happening is that older versions of Dagster supply this value from user code and that same value gets sent back-- so if we insert a dummy field at the serdes layer the user code value is lost in the roundtrip.

Regardless going to need to leave those in in this PR.

PR #13175, opened 2023-03-27
Add a reusable filtering component

salazarm:
## Summary & Motivation

Motivation is this Cloud issue: https://linear.app/elementl/issue/DSN-75/cloud-audit-logs-ui-follow-ups

This PR implements a reusable filtering component. The API exposed to users of the API is a hook: 

```
interface UseFiltersProps {
  filters: Filter<any, any>[];
  activeFilters: Filter<any, any>[];
  setActiveFilters: (next: React.SetStateAction<Filter<any, any>[]>) => void;
}
function useFilters({filters, activeFilters, setActiveFilters}: UseFiltersProps): { button: JSX.Element, activeFiltersJsx: JSX.Eelment }

```

This PR also implements 2 types of filters: TimeRangeFilter and SetFilter. 
The idea here is that we can extend the base Filter class to implement new types of filtering and support different behaviors (eg. supporting typeahead results from the server).

The reason for a hook  is because we need to return 2 separately renderable pieces of UI that share state.

## How I Tested These Changes

Storybook + I used this in the audit log filtering UI https://www.loom.com/share/b271abf4a2bf48309029bdd205d5338e

Screen Shot 2023-03-27 at 10 41 35 AM

[INTERNAL_BRANCH=auditLogFiltering]


js_modules/dagit/packages/core/src/ui/Filters/StaticSetFilter.tsx
```diff
@@ -0,0 +1,192 @@
...
+  state: Set<any>;
+  getStringValue: (value: any) => string;
+  onRemove: () => void;
+  renderLabel: (value: any) => JSX.Element;
+}) {
+  const arr = React.useMemo(() => Array.from(state), [state]);
+  const label = React.useMemo(() => {
+    if (arr.length === 0) {
+      return null;
+    } else if (arr.length <= 3) {
```
hellendag, line 100: Track `3` in a const?

js_modules/dagit/packages/core/src/ui/Filters/StaticSetFilter.tsx
```diff
@@ -0,0 +1,192 @@
...
+      return (
+        <>
+          is&nbsp;{arr.length === 1 ? '' : <>any of&nbsp;</>}
+          {arr.map((value, index) => (
+            <>
+              <FilterTagHighlightedText>{getStringValue(value)}</FilterTagHighlightedText>
+              {index < arr.length - 1 ? <>,&nbsp;</> : ''}
+            </>
```
hellendag, line 108: Looks like this can run out of horizontal space. Worth clamping the width and doing a `text-overflow: ellipsis`?

js_modules/dagit/packages/core/src/ui/Filters/StaticSetFilter.tsx
```diff
@@ -0,0 +1,192 @@
...
+                style={{maxHeight: '300px', overflow: 'auto'}}
+              >
+                {arr.map((value) => (
+                  <div key={value}>{renderLabel({value, isActive: true})}</div>
+                ))}
+              </Box>
+            }
+          >
+            <FilterTagHighlightedText>
+              {arr.length} {name.toLowerCase()}s
```
hellendag, line 132: This plural construction looks a little fragile, but we can go ahead with it and see if it breaks anywhere.

js_modules/dagit/packages/core/src/ui/Filters/StaticSetFilter.tsx
```diff
@@ -0,0 +1,192 @@
...
+          is any of{' '}
+          <Popover
+            interactionKind="hover"
+            position="bottom"
+            content={
+              <Box
+                padding={{vertical: 8, horizontal: 12}}
+                flex={{direction: 'column', gap: 4}}
+                style={{maxHeight: '300px', overflow: 'auto'}}
+              >
```
hellendag, line 124: Does this box need a max width? Curious what this looks like with super long strings.

js_modules/dagit/packages/core/src/ui/Filters/TimeRangeFilter.tsx
```diff
@@ -0,0 +1,291 @@
...
+import {Filter, FilterTag, FilterTagHighlightedText} from './Filter';
+
+import 'react-dates/initialize';
+import 'react-dates/lib/css/_datepicker.css';
+
+function calculateTimeRanges() {
+  return {
+    TODAY: {
+      label: 'Today',
+      range: [dayjs().startOf('day').toDate().getTime(), null] as TimeRangeState,
```
hellendag, line 19: See earlier comment re: moment-tz. If a user is using a different timezone in our app than the default browser timezone, these ranges might be incorrect.

js_modules/dagit/packages/core/src/ui/Filters/TimeRangeFilter.tsx
```diff
@@ -0,0 +1,291 @@
...
+
+  onSelect(
+    key: TimeRangeKey,
+    close: () => void,
+    createPortal: (element: JSX.Element) => () => void,
+  ) {
+    if (key === 'CUSTOM') {
+      const closeRef = {
+        current: () => {
+          debugger;
```
hellendag, line 96: eep

js_modules/dagit/packages/core/src/ui/Filters/TimeRangeFilter.tsx
```diff
@@ -0,0 +1,291 @@
...
+      {range}
+    </Box>
+  );
+}
+
+const L_FORMAT = new Intl.DateTimeFormat(navigator.language, {
+  year: 'numeric',
+  month: 'numeric',
+  day: 'numeric',
+  timeZone: 'UTC',
```
hellendag, line 125: Timezone concern here as well

js_modules/dagit/packages/core/src/ui/Filters/TimeRangeFilter.tsx
```diff
@@ -0,0 +1,291 @@
...
+        </>
+      );
+    } else if (isEqual(state, TimeRanges.LAST_30_DAYS.range)) {
+      return (
+        <>
+          is within <FilterTagHighlightedText>Last 30 days</FilterTagHighlightedText>
+        </>
+      );
+    } else {
+      if (!state[0]) {
```
hellendag, line 155: Worth using keys instead of a tuple for tracking ranges? I think it's not too bad as is, but it might be a little easier to read.

js_modules/dagit/packages/ui/src/components/Group.stories.tsx
```diff
@@ -108,10 +108,7 @@ export const EmptyChildren = () => {
 export const PointerEventsTest = () => {
   return (
     <div>
-      <ButtonLink
-        color={{link: Colors.Blue500, hover: Colors.Yellow500}}
-        onClick={() => console.log('Clicked!')}
-      >
```
hellendag, line 114: Undo?

hellendag (approved):
Looks cool! I have some comments throughout, mostly fairly minor.

I do find it a bit hard to follow in some places, so I think it would be good to get a bunch of tests on these that really put everything through its paces.

I'll go ahead and approve so that we can keep things moving. Would be great to get @bengotow's eyes on this too.

js_modules/dagit/packages/ui/src/components/Group.stories.tsx
salazarm: Oops, I was cleaning up console.log's and thought I added this one

js_modules/dagit/packages/core/src/ui/Filters/StaticSetFilter.tsx
salazarm: Yeah, I mean the right way to do this would probably be an internationalization framework

js_modules/dagit/packages/core/src/ui/Filters/Filter.tsx
```diff
@@ -0,0 +1,81 @@
...
+  <BaseTag
+    icon={<Icon name={iconName} color={Colors.Link} />}
+    rightIcon={
+      <div onClick={onRemove} style={{cursor: 'pointer'}}>
+        <Icon name="close" color={Colors.Link} />
+      </div>
+    }
+    label={label}
+    fillColor={Colors.Blue50}
+    textColor={Colors.Link}
```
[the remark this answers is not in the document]
salazarm: I think you could just create your own BaseTag if you want it to look different

js_modules/dagit/packages/core/src/hooks/useQueryPersistedState.tsx
```diff
@@ -96,7 +97,6 @@ export function useQueryPersistedState<T extends QueryPersistedDataType>(
         }
       }
 
-      currentQueryString = next;
```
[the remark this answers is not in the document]
salazarm: It was never necessary since this is the last reference to `currentQueryString` in this closure and no child closures are referencing it, so it was never doing anything or being used by anything.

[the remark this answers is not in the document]
salazarm: oh wait, it's being set up by the parent closure. Yeah I can add this back. Mistakenly thought it wasn't necessary

js_modules/dagit/packages/core/src/ui/Filters/Filter.tsx
```diff
@@ -0,0 +1,81 @@
...
+  label: JSX.Element;
+  iconName: IconName;
+  onRemove: () => void;
+}) => (
+  <BaseTag
+    icon={<Icon name={iconName} color={Colors.Link} />}
+    rightIcon={
+      <div onClick={onRemove} style={{cursor: 'pointer'}}>
+        <Icon name="close" color={Colors.Link} />
+      </div>
```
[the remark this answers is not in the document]
salazarm: How about I add `tabIndex={0}`? It's always a drag to figure out what are all the styles that need to be overriden especially when you take into account all the psuedo states (active, focused, hover).

js_modules/dagit/packages/core/src/ui/Filters/Filter.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  public abstract isActive(): boolean;
+
+  public abstract renderActiveFilterState(): JSX.Element | null;
+  public abstract getResults(query: string): {label: JSX.Element; value: TValue}[];
+
+  public abstract onSelect(
+    value: TValue,
+    close: () => void,
+    createPortal: (element: JSX.Element) => () => void,
```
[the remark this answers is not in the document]
salazarm: Yeah let me do that

js_modules/dagit/packages/core/package.json
```diff
@@ -50,9 +50,11 @@
     "graphql": "^15.5.0",
     "lodash": "^4.17.15",
     "lru-cache": "^6.0.0",
+    "moment": "^2.29.4",
```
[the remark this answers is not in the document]
salazarm: Good call out.
For now I'm passing time in using the locale time. I think as long as the rendered timestamp in the table and the input timestamp in the filtering use the same locale then there shouldn't be anything confusing. Let me take another look at this

js_modules/dagit/packages/core/src/ui/Filters/Filter.tsx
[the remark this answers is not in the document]
salazarm: We should probably have a lint rule... any base html element with an onClick that isn't an interactive element should have `tabIndex={0}`

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,272 @@
...
+    [setIsOpen, setPortaledElements],
+  );
+
+  const allResultsJsx = React.useMemo(() => {
+    if (selectedFilter) {
+      return selectedFilter
+        .getResults(search)
+        .map((result, index) => (
+          <MenuItem
+            key={index}
```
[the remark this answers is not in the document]
salazarm: yeah good point

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,272 @@
...
+        keyRef.current++;
+      }
+      prevOpenRef.current = isOpen;
+    }, []),
+  );
+
+  return (
+    <ShortcutHandler
+      shortcutLabel="⌥F"
+      shortcutFilter={(e) => e.code === 'KeyF'}
```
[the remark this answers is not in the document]
salazarm: I think in this case we want to trigger it without the alt key

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
```diff
@@ -0,0 +1,272 @@
...
+      ) {
+        return;
+      }
+      setIsOpen(false);
+    };
+    document.body.addEventListener('click', listener);
+    return () => {
+      document.body.removeEventListener('click', listener);
+    };
+  }, []);
```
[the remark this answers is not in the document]
salazarm: A controlled Popover component doesn't seem to trigger the popover to close when you click outside, so I added this logic. `canEscapeKeyClose` also doesn't seem to be working :\

js_modules/dagit/packages/core/src/ui/Filters/StaticSetFilter.tsx
```diff
@@ -0,0 +1,192 @@
...
+        label: <SetFilterLabel value={value} renderLabel={this.renderLabel} filter={this} />,
+        value,
+      }));
+    }
+    return this.allValues
+      .filter(({match}) => match.some((value) => value.toLowerCase().includes(query.toLowerCase())))
+      .map(({value}) => ({
+        label: <SetFilterLabel value={value} renderLabel={this.renderLabel} filter={this} />,
+        value,
+      }));
```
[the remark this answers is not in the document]
salazarm: sure, though sometimes not consolidating is easier to read, in this case probably not to difficult to read though when consolidated

js_modules/dagit/packages/core/src/ui/Filters/StaticSetFilter.tsx
salazarm: Looks like `BaseTag` takes care of that

PR #13177, opened 2023-03-27
[structured config] respect field aliases for resources

jamiedemaria:
## Summary & Motivation
Writing a resource with an aliased config field would error because the alias was not applied when constructing the configuration as a dictionary. 

based off the convo [here](https://github.com/dagster-io/dagster/pull/12695#discussion_r1129763364) it seems like  _as_config_dict should already respect aliases, but in my testing that wasn't the case with resources. Maybe there's a different place that this fix should go that's more resource specific? let me know. I'm still pretty unfamiliar with the code for structured config so idk if i'll be able to find it myself

## How I Tested These Changes
new unit test - failed before the fix and passes now

### DIFF
--- a/python_modules/dagster/dagster/_config/structured_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/structured_config/__init__.py
@@ -134,7 +134,10 @@ class Config(MakeConfigCacheable):
             field = self.__fields__.get(key)
             if field and value is None and not _is_pydantic_field_required(field):
                 continue
-            output[key] = value
+            if field:
+                output[field.alias] = value
+            else:
+                output[key] = value
         return output
 
 
--- a/python_modules/dagster/dagster_tests/core_tests/resource_tests/test_struct_config_resources.py
+++ b/python_modules/dagster/dagster_tests/core_tests/resource_tests/test_struct_config_resources.py
@@ -1812,3 +1812,14 @@ def test_aliased_field_structured_resource():
 
     assert prefix_job.execute_in_process().success
     assert out_txt == ["greeting: hello, world!"]
+
+    out_txt.clear()
+
+    @job(resource_defs={"writer": WriterResource.configure_at_launch()})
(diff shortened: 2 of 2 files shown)


jamiedemaria:
* master
  * **PR #13177** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13177" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13177?utm_source=stack-comment).

python_modules/dagster/dagster/_config/structured_config/__init__.py
```diff
@@ -134,7 +134,10 @@ def _as_config_dict(self) -> Mapping[str, Any]:
             field = self.__fields__.get(key)
             if field and value is None and not _is_pydantic_field_required(field):
                 continue
-            output[key] = value
+            if field:
+                output[field.alias] = value
```
jamiedemaria, line 138: if the config doesn't have an alias, the `alias` attribute is just the name of the config

python_modules/dagster/dagster_tests/core_tests/resource_tests/test_struct_config_resources.py
```diff
@@ -1781,3 +1782,33 @@ def hello_world_job() -> None:
...
+
+    assert no_prefix_job.execute_in_process().success
+    assert out_txt == ["hello, world!"]
+
+    out_txt.clear()
+
+    @job(resource_defs={"writer": WriterResource(prefix="greeting: ")})
+    def prefix_job():
+        hello_world_op()
+
```
benpankow, line 1812: Would be good to test runtime config too (e.g. `WriterResource.configure_at_launch()` and then pass the `{"prefix": "foo"}` to `execute_in_process`.

benpankow (approved):
:pray: thanks for the fix!

PR #13181, opened 2023-03-27
[dagit] redirect trailing slash on GraphQL route

benpankow:
## Summary

Adds a redirect from the graphql URL with a trailing slash: `localhost:3000/graphql/`->`localhost:3000/graphql`. A user recently ran into this: https://elementl-workspace.slack.com/archives/C046ZK3KQ5D/p1679944131264839?thread_ts=1679942695.590929&cid=C046ZK3KQ5D

### DIFF
--- a/python_modules/dagit/dagit/webserver.py
+++ b/python_modules/dagit/dagit/webserver.py
@@ -286,6 +286,11 @@ class DagitWebserver(GraphQLServer, Generic[T_IWorkspaceProcessContext]):
                     self.graphql_ws_endpoint,
                     name="graphql-ws",
                 ),
+                Route(
+                    "/graphql/",
+                    lambda _: RedirectResponse(url="/graphql"),
+                    methods=["GET", "POST"],
+                ),
             ]
             + self.build_static_routes()
             + [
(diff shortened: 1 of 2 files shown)


benpankow:
* master
  * **PR #13181** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13181" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13181?utm_source=stack-comment).

brad-alexander:
A compelling reason to omit the redirect (or make it optional) would be to prevent accidental performance issues.  It looks like even if you use sessions in `requests`, it does not remember 307 or even 308 responses, and will always make two roundtrips for every request.  This probably wouldn't be much of a hit if the server is on the same network, but if the server is under significant load, or over slow WAN/VPN it could be significant overhead - especially if it is a large POST request.

alangenfeld:
should we take an approach similar to what we did here and just dupe the routes instead 
https://github.com/dagster-io/internal/pull/876/files

alangenfeld (changes requested):
to your queue

alangenfeld:
related https://github.com/encode/starlette/issues/869 though that doesn't necessarily help much

PR #13182, opened 2023-03-27
Use specific events to trigger a "Preparing" state instead of assuming that any logged event means that the step has started

gibsondan:
Summary:
This is intended to address a problem where the Dagit UI incorrectly thinks that a step is being prepared after its upstream event fails so we log a "Dependencies for step XXX failed. Not executing." events.

Since https://github.com/dagster-io/dagster/commit/e163de1e7d1927199242b87f25b1839b818908d7 we have added additional structured events that explicitly indicate that a step has started. Use those to determine that the step is being prepared - until then, leave the state null.

Furthermore, source the set of steps that haven't executed from the graph, rather than purely the event log - since a step that didn't execute may not have logged an event at all.

Test Plan:
Launch this run:
```
import time

from dagster import asset

@asset
def a():
    time.sleep(45)

@asset
def c(b):
    pass

@asset
def b():
    raise Exception("Oops")

```

Before this change, c is shown in a Preparing / Not executed state in the right hand bar. After this change, it is not shown in the right hand bar at all.

Launch a normal run where every step succeeds or fails, no changes

Are there tests related to dynamic orchestration that I should check, can we count on the nodes in the graph matching up with the keys in the metadata?

### DIFF
--- a/js_modules/dagit/packages/core/src/gantt/GanttStatusPanel.tsx
+++ b/js_modules/dagit/packages/core/src/gantt/GanttStatusPanel.tsx
@@ -2,6 +2,7 @@ import {Colors, Spinner, Tooltip} from '@dagster-io/ui';
 import * as React from 'react';
 import styled from 'styled-components/macro';
 
+import {GraphQueryItem} from '../app/GraphQueryImpl';
 import {formatElapsedTime} from '../app/Util';
 import {SidebarSection} from '../pipelines/SidebarComponents';
 import {IRunMetadataDict, IStepState} from '../runs/RunMetadataProvider';
@@ -12,6 +13,7 @@ import {boxStyleFor} from './GanttChartLayout';
 import {RunGroupPanel} from './RunGroupPanel';
 
 interface GanttStatusPanelProps {
+  graph: GraphQueryItem[];
   metadata: IRunMetadataDict;
   selection: StepSelection;
   runId: string;
@@ -25,18 +27,20 @@ interface GanttStatusPanelProps {
 export const GanttStatusPanel: React.FC<GanttStatusPanelProps> = ({
   runId,
   nowMs,
+  graph,
   metadata,
   selection,
   onClickStep,
   onDoubleClickStep,
   onHighlightStep,
 }) => {
-  const {preparing, executing, errored, succeeded} = React.useMemo(() => {
+  const {preparing, executing, errored, succeeded, notExecuted} = React.useMemo(() => {
     const keys = Object.keys(metadata.steps);
     const preparing = [];
     const executing = [];
     const errored = [];
     const succeeded = [];
+    const notExecuted = [];
     for (const key of keys) {
       const state = metadata.steps[key].state;
       switch (state) {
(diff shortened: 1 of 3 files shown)


js_modules/dagit/packages/core/src/runs/RunMetadataProvider.tsx
```diff
@@ -285,7 +280,12 @@ export function extractMetadataFromLogs(
         }
       }
 
-      if (log.__typename === 'ExecutionStepStartEvent') {
+      if (
+        log.__typename === 'ResourceInitStartedEvent' ||
+        log.__typename === 'StepWorkerStartingEvent'
+      ) {
+        upsertState(step, timestamp, IStepState.PREPARING);
```
alangenfeld, line 287: did you test in process execution? I think that would just skip `Preparing` state entirely which i think is fine behavior but worth double checking if you did not

js_modules/dagit/packages/core/src/gantt/GanttStatusPanel.tsx
```diff
@@ -52,10 +56,18 @@ export const GanttStatusPanel: React.FC<GanttStatusPanelProps> = ({
...
           succeeded.push(key);
+          break;
       }
     }
-    return {preparing, executing, errored, succeeded};
-  }, [metadata]);
+
+    for (const node of graph) {
+      const name = node.name;
+      if (!(name in metadata.steps) || !metadata.steps[name].state) {
```
hellendag, line 65: Looks like you could probably just do:

```
if (!metadata.steps[name]?.state) {
```

Or if you want to be explicit about checking for the key, it's slightly safer to use `Object.hasOwn` instead of the `in` operator.

```
if (!Object.hasOwn(metadata.steps, 'name') || ...) {
``

hellendag (commented):
Seems okay to me but @bengotow might have a bit more context on it.

hellendag (approved):

js_modules/dagit/packages/core/src/runs/RunMetadataProvider.tsx
gibsondan: ResourceInitStartedEvent still fires in in process

alangenfeld: but only once for the entire in process run not associated with a specific step

gibsondan: ah true... that was somewhat before though this PR I think? (in the case where it actually executes, we weren't showing a Preparing line at all since I don't believe there were any logs before the step startsed?)

I will be sure to test that nothing too crazy happens

bengotow (approved):
This looks ok to me 👍  I think this used to work because we would explicitly emit a `STEP_SKIPPED` / `ExecutionStepSkippedEvent` event and our state machine would switch the step into a skipped state that had it's own visual styling. It seems like the problem is that some new / other / different events are now emitted for skipped steps?

I like the idea of moving to the preparing state only on specific events, seems like it'll prevent this sort of problem from cropping back up in the future too 👍

js_modules/dagit/packages/core/src/runs/RunMetadataProvider.tsx
alangenfeld: ya i highly suspect its fine / no change, but its worth including in your test plan in my opinion

PR #13185, opened 2023-03-27
RFC: Add a timeout to the multiprocess executor for shutting down the subprocess for each step

gibsondan:
Summary:
Rather than hanging forever, if an issue makes the join() hang after a subprocess call completes, timeout, fail the step, and move on.

Sending out for early feedback, will add an executor-specific test if it seems reasonable

### DIFF
--- a/python_modules/dagster/dagster/_core/executor/multiprocess.py
+++ b/python_modules/dagster/dagster/_core/executor/multiprocess.py
@@ -35,6 +35,7 @@ from .child_process_executor import (
     ChildProcessCommand,
     ChildProcessCrashException,
     ChildProcessEvent,
+    ChildProcessHangOnShutdownException,
     ChildProcessSystemErrorEvent,
     execute_child_process_command,
 )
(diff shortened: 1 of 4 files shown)


python_modules/dagster/dagster/_core/executor/multiprocess.py
```diff
@@ -244,6 +248,32 @@ def execute(
...
+                                EngineEventData.engine_error(serializable_error),
+                            )
+                            step_failure_event = DagsterEvent.step_failure_event(
+                                step_context=plan_context.for_step(
+                                    active_execution.get_step_by_key(key)
+                                ),
+                                step_failure_data=StepFailureData(
+                                    error=serializable_error, user_failure_data=None
+                                ),
+                            )
```
gibsondan, line 273: not sure about failing here - technically this could fire after a STEP_SUCCESS event is already yielded right? (the same is theoretically possible with ChildProcessCrashException if it fires after the STEP_SUCCESS event)

johannkm (changes requested):
for @alangenfeld 's comment

PR #13186, opened 2023-03-28
[asset selection] Added required_neighbors asset selection option

benpankow:
## Summary

Adds a new `required_neighbors()` [name not final] asset selection option, which is used to select the minumum set of assets necessary to materialize the selected assets.

This arises from a user who is having trouble materializing the upstream Airbyte assets for their dbt project. The Airbyte sync op cannot be subset.

This is useful for multi-assets which cannot be subselected. For example, if the same op computes `foo`, `bar`, and `baz`, `AssetSelection.keys("foo").required_neighbors()` will return all three assets.

## Test Plan

New unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_selection.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_selection.py
@@ -57,7 +57,7 @@ class AssetSelection(ABC):
             AssetSelection.all() - AssetSelection.groups("marketing")
 
             # Select all assets which are materialized by the same op as "projections":
-            AssetSelection.keys("projections").required_neighbors()
+            AssetSelection.keys("projections").required_multi_asset_neighbors()
     """
 
     @public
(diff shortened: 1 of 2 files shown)


benpankow:
* master
  * **PR #13186** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13186" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13186?utm_source=stack-comment).

OwenKephart (approved):
I like this functionality, and I think the name is reasonable. cc @sryza for a second opinion

sryza:
Definitely seems like useful functionality. I'm not thrilled on the name, but having trouble coming up with concise alternatives.

Reasons not thrilled:
- "required" - required for what?
- "neighbors" - new term
- It sounds like this would include the neighbors, but not the assets included in the original selection

sryza:
`required_multi_asset_neighbors`?

That makes it clear what "neighbors" is referring to.

benpankow:
I like `required_multi_asset_neighbors`. Updated!

sryza (approved):

PR #13189, opened 2023-03-28
feat(dbt): follow dbt Core's version support constraints

rexledesma:
## Summary & Motivation
Follow https://docs.getdbt.com/docs/dbt-versions/core, and explicitly prevent older versions of `dbt-core` from being installed.

## How I Tested These Changes
bk

PR #13190, opened 2023-03-28
feat(dbt): remove unnecessary checks for dbt Core installation

rexledesma:
## Summary & Motivation
The installation of `dbt-core` is already enforced by the package requirements. Adding a check for it is superfluous.

## How I Tested These Changes
pytest

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cli/utils.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cli/utils.py
@@ -116,13 +116,6 @@ def _core_execute_cli(
     json_log_format: bool,
 ) -> Iterator[Union[DbtCliEvent, int]]:
     """Runs a dbt command in a subprocess and yields parsed output line by line."""
-    try:
-        import dbt  # noqa: F401
-    except ImportError as e:
-        raise check.CheckError(
-            "You must have `dbt-core` installed in order to execute dbt CLI commands."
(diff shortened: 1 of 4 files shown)


rexledesma:
* master
  * **PR #13189** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13189" target="_blank">Graphite</a> 
    * **PR #13190** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13190" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13190?utm_source=stack-comment).

yuhan (commented):
i found this one https://github.com/dagster-io/dagster/pull/8666 from a while back. curious why we added the checks in the first place?

rexledesma:
> i found this one #8666 from a while back. curious why we added the checks in the first place?

My thought process back then was that if users somehow managed to upgrade `dagster-dbt` to the patch requiring `dbt-core`, at least they would run into an error message.

But considering that the `dbt-core` requirement has been baked into `dagster-dbt` since `0.15.3` now, I think it's safe to say that this error message is not needed anymore. We can just safely assume that `dbt-core` is properly installed.

yuhan (approved):
sounds good to me. also worth auditing / cleaning up the oss examples where we have dbt as a dependency.

PR #13197, opened 2023-03-28
don't run mlflow tests on python 3.11

gibsondan:
Summary:
Later versions of mlflow don't install on this version:


alangenfeld (approved):
setup.py need update too or is 3.11 not encoded there yet?

gibsondan:
it seems to be specific versions of mlflow that don't install on python 3.11 - i don't expect it to be permanent and i dunno that we need to add a pin ourselves around a temporary upstream issue? But not strongly held

alangenfeld:
oh i didnt mean pin i just meant what versions of python we expect `dagster-mlflow` to work on based on what its tested on https://github.com/dagster-io/dagster/blame/master/python_modules/libraries/dagster-mlflow/setup.py#L28

I think it makes sense for our advertised version functionality to match whats under test but I dont think its a huge deal.

gibsondan (commented):
ah got it

PR #13200, opened 2023-03-28
[dagit] Collapse large table-schema metadata values

### LINKED ISSUE #12075: `TableSchema` metadata should be collapsible in Dagit
### Dagster version

1.1.15

### What's the issue?

The `TableSchema` metadata takes too much screen space as it's not collapsible. I have tables with hundreds of columns and it breaks the asset metadata view.

Collapse long TableSchemas in Dagit

### How to reproduce?

Log a `TableSchema` with a lot of columns

### Deployment type

Docker Compose

### Deployment details

_No response_

### Additional information

_No response_

bengotow:
## Summary & Motivation

Previously table schema metadata always rendered inline, even if it was very long. Now we collapse it the same way we collapse JSON and markdown behind a "Show Table Schema" link that opens a modal. Fixes #12075. 

If there are <5 columns on the schema and "expandSmallValues" is set, the data is still shown inline.

image

image

## How I Tested These Changes

I verified this fix by creating a new storybook test case and verifying both the small and large table schema scenarios.

### DIFF
--- a/js_modules/dagit/packages/core/src/metadata/MetadataEntry.stories.tsx
+++ b/js_modules/dagit/packages/core/src/metadata/MetadataEntry.stories.tsx
@@ -250,6 +250,22 @@ const MetadataEntryMocks = [
     label: 'my_short_json',
     jsonString: '{"short_value": 12}', // Very short JSON is inlined
   } as JsonMetadataEntry,
+
+  {
+    __typename: 'TableSchemaMetadataEntry',
+    description: 'This is the description',
+    label: 'my_table_schema',
+    schema: {
+      ...MetadataTableSchema,
+      columns: [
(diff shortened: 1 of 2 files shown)


js_modules/dagit/packages/core/src/metadata/MetadataEntry.stories.tsx
```diff
@@ -250,6 +250,22 @@ const MetadataEntryMocks = [
     label: 'my_short_json',
     jsonString: '{"short_value": 12}', // Very short JSON is inlined
   } as JsonMetadataEntry,
+
+  {
+    __typename: 'TableSchemaMetadataEntry',
```
salazarm, line 255: `buildTableSchemaMetadataEntry`

Issue #12422, opened 2023-02-17
sidebar on asset graph shows outdated number of partitions


sryza:
E.g. here, where the asset node indicates that 3 partitions are filled, but the bar only shows that two are filled:
<img width="787" alt="image" src="https://user-images.githubusercontent.com/654855/219717872-d635c732-fb8b-4cb3-a305-76dfbd5a0f0c.png">

This happens immediately after a manual asset partition materialization. Refreshing the page or closing and then reopening the sidebar fixes the issue.

bengotow:
Merging this into https://github.com/dagster-io/dagster/issues/13023 - the issue here is that `usePartitionHealthData` is pulling the full health data from separate GraphQL queries and doesn't know to refetch when the live query on the asset graph sees the count change.

PR #13201, opened 2023-03-28
[dagit] Refresh health in the asset sidebar if live data updates

### LINKED ISSUE #13023: Reloading definitions in Dagit / changes to materialization counts on the graph don't refresh the currently viewed asset partitions
### Dagster version

1.2.2

### What's the issue?

The title (looking at the asset tab on the right). 
Perhaps something else is not refreshed too? 

The partitions should refresh, for example if the code has been changed.

### How to reproduce?

1. Define a partitioned asset
2. Open it in Dagit
3. Change the partitions def (I gave it an offset)

### Deployment type

Local

### Deployment details

_No response_

### Additional information

_No response_

bengotow:
## Summary & Motivation

This PR resolves #13023. If you're viewing the asset graph and have an asset selected, it's partition health bar is shown in the sidebar. Previously, the health bar would not refresh when the asset's partition counts updated on the graph because the health bar uses a separate GraphQL query that is not refreshed periodically. 

The `usePartitionHealthData` hook already accepts a "cache hint" that allows us to use lightweight polling to trigger a heavy refresh of the health bar (on the asset details page). This PR wires it up on the asset graph as well, and moves some of the code to a shared helper.

## How I Tested These Changes

We don't have tests for the asset graph sidebar yet. To test this I launched runs of an asset from the asset graph and observed that the counts on the graph and the counts in the sidebar are now in sync. 

I also changed the partition definition in code and verified that both the asset graph and the asset sidebar updated to reflect the new number of partitions in the asset. (The case in #12422)

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/SidebarAssetInfo.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/SidebarAssetInfo.tsx
@@ -16,7 +16,10 @@ import {DependsOnSelfBanner} from '../assets/DependsOnSelfBanner';
 import {PartitionHealthSummary} from '../assets/PartitionHealthSummary';
 import {assetDetailsPathForKey} from '../assets/assetDetailsPathForKey';
 import {AssetKey} from '../assets/types';
-import {usePartitionHealthData} from '../assets/usePartitionHealthData';
+import {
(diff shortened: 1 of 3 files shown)


salazarm (approved):
looks good to me!

PR #13203, opened 2023-03-28
Speed up failed partition materialization query

clairelin135:
Internal PR: https://github.com/dagster-io/internal/pull/5309

The existing query that rebuilds the failed materialization partitions subset causes timeouts: https://app.datadoghq.com/apm/trace/7937935115983931247?spanID=6629004549093613704&spanViewType=errors

The outer join in the existing query attempts to find a matching materialization for each materialization planned event, but PG's query planner causes the query to loop through each materialization event many times.

This PR amends this approach to just fetch the latest materialization / materialization planned events per partition, and iterate in python to return the planned events without a corresponding materialization event.


python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1756,31 +1756,21 @@ def get_latest_asset_partition_materialization_attempts_without_materializations
...
+
+        materialization_planned_rows_by_partition = {
+            row["partition"]: (row["run_id"], row["id"]) for row in materialization_planned_rows
+        }
+        for row in materialization_rows:
+            if (
+                row["partition"] in materialization_planned_rows_by_partition
+                and materialization_planned_rows_by_partition[row["partition"]][0] == row["run_id"]
+            ):
+                materialization_planned_rows_by_partition.pop(row["partition"], None)
```
johannkm, line 1771: nit, the None isn't required I think. Missed that in cloud but doesn't really matter

johannkm (approved):

PR #13204, opened 2023-03-28
Enhance error output for partition check

CodeMySky:
### Summary & Motivation
Previously this invariant check doesn't provide the parent asset name which don't have the partition, it could be confusing  and hard to debug when the dependency graph is large. I added the `parent_asset_key` in the error to add more information about which asset is expected to have partitions. 

```
make black
python -m pytest python_modules/dagster/dagster_tests
```


smackesey:
Thanks @CodeMySky, we always like to improve our error messages. I'm running our CI against your PR and we can merge if it passes.

PR #13205, opened 2023-03-28
[RFC] Initial AutoMaterializationPolicy

OwenKephart:
## Summary & Motivation

Strawman implementation for ReconciliationPolicy [name TBD].

One consideration is that we don't want to instantly break people who are using the existing `build_asset_reconciliation_sensor` functionality. Instead, we detect if there are any ReconciliationPolicies defined in the selection for the sensor, and if there aren't we revert to the legacy way of determining which assets are eager and which assets are lazy (basically, you're lazy if you're upstream of a FreshnessPolicy, otherwise you're eager).

This also does not currently implement a serialization scheme for these policies, so this only works for InternalAssetGraphs.


sryza:
@OwenKephart did you git add `auto_materialization_policy.py`?

PR #13211, opened 2023-03-28
[docs] Update build docs action to link to changed pages

benpankow:
## Summary

Updates the "build docs" GitHub action to assemble a list of changed pages and link to them in the comment, rather than the generic message. This makes it easier to jump directly to the new/changed page rather than navigating in the PR docs instance manually.

## Test Plan

Tested on this PR - the message is a bit updated from the last test but the GH actions outage is preventing it from rerunning 😄

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -1,5 +1,14 @@
 name: Deploy Docs
-on: push
+on:
+  push:
+    branches:
+      - master
+      - 'release-*'
+    paths:
+      - docs/**
+  pull_request:
+    paths:
+      - docs/**
 jobs:
   deploy:
     runs-on: ubuntu-latest
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -1,21 +1,35 @@
 name: Deploy Docs
-on:
-  push:
-    branches:
-      - master
-      - 'release-*'
-    paths:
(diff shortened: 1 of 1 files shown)


benpankow:
* master
  * **PR #13211** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13211" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13211?utm_source=stack-comment).

prha:
Can you check this against non-mdx docs changes?

.github/workflows/build-docs.yml
```diff
@@ -1,21 +1,31 @@
 name: Deploy Docs
-on:
-  push:
```
prha, line 3: I think we still want to build on landing to master... I believe this is important with how the docs push is run.

Why are we changing the path filter?

prha (changes requested):
make sure to

prha (changes requested):
I think we still want to build on landing to master... I believe this is important with how the docs push is run.

Why are we changing the path filter?

benpankow:
> I think we still want to build on landing to master... I believe this is important with how the docs push is run.
> 
> Why are we changing the path filter?

This was just to get it to run for the sake of testing. Will switch back before merging 👍

PR #13215, opened 2023-03-29
[dagit] Partition status cleanup + support for materializing and mixed failed/missing statuses, etc.

bengotow:
## Summary & Motivation

Partial failed / other states!
image

New blue "materializing" status!
image

Partial states supported in dots:
image

This PR implements the new GraphQL endpoint for `numMaterializing` and also shows the new "materializing" state on the asset health partition bar.

This unfortunately required a lot of changes -- our previous implementation of the health bar used an enum called PartitionState for each color band. This was bad because:

- Each combinatorial state (eg: partially materializing, partially failed) would be a new enum value. Until now we have not supported partial states with the new "failed" red color because there just weren't enum values for it.

- Both assets and op backfills used this PartitionState enum. This was nice in a few places but led to a lot of code supporting just a subset of the enum values, mapping materialized = success, etc. Asset and op backfill status are getting more different and it's preferable to separate them now.

- The enum was called `PartitionState` even though the GraphQL types are all "status"

Most of the changes in this PR fold this enum away in favor of:

- PartitionState => AssetPartitionStatus (materialized, materializing, failed, missing, and soon stale)

- PartitionState => RunStatus for op backfills (we were previously taking RunStatus and converting it to a PartitionState client-side)

- Multiple AssetPartitionStatus values allowed for each partition key. This is useful for the first dimension of multi-partitioned assets but ALSO surfaces in Dagit if you select two assets with the same partitioning and open the backfill modal. Their bars are folded into one with "partial red / partial green", etc.

## How I Tested These Changes

We have extensive test coverage of the asset portion of this stuff! The tests are all passing and I'm fairly confident the changes there are safe. I manually reviewed the storybooks and the asset pages for correctness as well.

There are a few changes in this PR that impact the Op backfill UI, and I am testing the following manually:
- Viewing the op job partitions page
- Viewing the op job backfill history to see "Requested runs" and "Run status" elements are unchanged
- Clicking "Launch backfill" and confirming that the partition range selector renders properly and that the checkboxes below function correctly.

- Open our Backfill table storybooks and make sure they are unchanged

image

### DIFF
--- a/js_modules/dagit/packages/core/src/app/Util.tsx
+++ b/js_modules/dagit/packages/core/src/app/Util.tsx
@@ -167,8 +167,8 @@ export function weakmapMemoize<T extends object, R>(
   };
 }
 
-export function assertUnreachable(_: never): never {
-  throw new Error("Didn't expect to get here");
+export function assertUnreachable(value: never): never {
+  throw new Error(`Didn't expect to get here with value: ${value}`);
 }
 
 export function debugLog(...args: any[]) {
--- a/js_modules/dagit/packages/core/src/app/Util.tsx
+++ b/js_modules/dagit/packages/core/src/app/Util.tsx
@@ -168,7 +168,7 @@ export function weakmapMemoize<T extends object, R>(
 }
 
 export function assertUnreachable(value: never): never {
-  throw new Error(`Didn't expect to get here with value: ${value}`);
+  throw new Error(`Didn't expect to get here with value: ${JSON.stringify(value)}`);
 }
 
 export function debugLog(...args: any[]) {
--- a/js_modules/dagit/packages/core/src/asset-graph/AssetNode.mocks.ts
+++ b/js_modules/dagit/packages/core/src/asset-graph/AssetNode.mocks.ts
@@ -350,6 +350,7 @@ export const LiveDataForNodePartitionedSomeMissing: LiveDataForNode = {
   freshnessPolicy: null,
   partitionStats: {
     numMaterialized: 6,
+    numMaterializing: 0,
     numPartitions: 1500,
     numFailed: 0,
   },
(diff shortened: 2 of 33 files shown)


js_modules/dagit/packages/core/src/asset-graph/AssetNode.mocks.ts
```diff
@@ -543,6 +551,7 @@ export const LiveDataForNodePartitionedLatestRunFailed: LiveDataForNode = {
   freshnessPolicy: null,
   partitionStats: {
     numMaterialized: 1495,
+    numMaterializing: 0,
```
salazarm, line 554: If you use the `build` functions then when the schema changes like this you dont need to update any mocks!

js_modules/dagit/packages/core/src/app/Util.tsx
```diff
@@ -167,8 +167,8 @@ export function weakmapMemoize<T extends object, R>(
   };
 }
 
-export function assertUnreachable(_: never): never {
-  throw new Error("Didn't expect to get here");
+export function assertUnreachable(value: never): never {
+  throw new Error(`Didn't expect to get here with value: ${value}`);
```
hellendag, line 171: `JSON.stringify()` to avoid `[object Object]` noise?

js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
```diff
@@ -291,7 +291,9 @@ function buildAssetNodeStatusRow({
         <>
           <AssetLatestRunSpinner liveData={liveData} />
           <Caption style={{flex: 1}} color={Colors.Gray800}>
-            Materializing...
+            {liveData.partitionStats?.numMaterializing
+              ? `Materializing ${liveData.partitionStats.numMaterializing} partitions...`
```
hellendag, line 295: Plural is incorrect if it's one partition

hellendag (approved):

js_modules/dagit/packages/core/src/asset-graph/AssetNode.mocks.ts
bengotow: Ahh yeah, these ones all have specific values plugged in to test certain cases so I think they need to be made manually - going to try to use the build functions elsewhere though.

PR #13223, opened 2023-03-29
Add backfill resolvers for per-asset status counts

clairelin135:
This PR adds resolvers to `GraphenePartitionBackfill` to enable querying for partition status counts on a per asset basis, i.e. for backfilled asset `A` one run is requested, one is complete, one is failed.

One thing I came across is that the asset backfill iteration considers a backfill as "complete" when all runs have been requested. Subsequently, the `AssetBackfillData` object contains only the complete and failed partitions at the time that the last set of runs were requested, not including the results from the newly requested set of runs.

In this PR I edited the "complete" logic to instead return true when all requested partitions have either materialized/failed/canceled, so the `AssetBackfillData` object has the full status counts after completion. This has the consequence of instigating asset backfill iterations until all runs have concluded.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/schema/partition_sets.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/partition_sets.py
@@ -14,7 +14,6 @@ from dagster._core.host_representation.external_data import (
 from dagster._core.storage.pipeline_run import RunsFilter
 from dagster._core.storage.tags import PARTITION_NAME_TAG, PARTITION_SET_TAG
 from dagster._utils.merger import merge_dicts
-from .asset_key import GrapheneAssetKey
 
 from dagster_graphql.implementation.fetch_partition_sets import (
     get_partition_by_name,
@@ -26,7 +25,8 @@ from dagster_graphql.implementation.fetch_partition_sets import (
 )
 from dagster_graphql.implementation.fetch_runs import get_runs
 
-from .backfill import GraphenePartitionBackfill, GrapheneBulkActionStatus
+from .asset_key import GrapheneAssetKey
+from .backfill import GraphenePartitionBackfill
 from .errors import (
     GrapheneDuplicateDynamicPartitionError,
     GraphenePartitionSetNotFoundError,
@@ -122,21 +122,15 @@ class GraphenePartitionStatusesOrError(graphene.Union):
         name = "PartitionStatusesOrError"
 
 
-class GrapheneBulkActionStatusPartitionCounts(graphene.ObjectType):
-    bulkActionStatus = graphene.NonNull(GrapheneBulkActionStatus)
-    count = graphene.NonNull(graphene.Int)
-
+class GrapheneAssetPartitionsStatusCounts(graphene.ObjectType):
     class Meta:
-        name = "PartitionBulkActionStatusCounts"
-
-
-class GrapheneAssetPartitionStatusCounts(graphene.ObjectType):
-    class Meta:
-        name = "AssetPartitionStatusCounts"
+        name = "AssetPartitionsStatusCounts"
 
     assetKey = graphene.NonNull(GrapheneAssetKey)
     numPartitionsTargeted = graphene.NonNull(graphene.Int)
-    partitionStatusCounts = non_null_list(GrapheneBulkActionStatusPartitionCounts)
+    numPartitionsRequested = graphene.NonNull(graphene.Int)
+    numPartitionsCompleted = graphene.NonNull(graphene.Int)
+    numPartitionsFailed = graphene.NonNull(graphene.Int)
 
 
 class GraphenePartitionTagsOrError(graphene.Union):
--- a/python_modules/dagster-graphql/dagster_graphql/schema/partition_sets.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/partition_sets.py
@@ -14,6 +14,7 @@ from dagster._core.host_representation.external_data import (
 from dagster._core.storage.pipeline_run import RunsFilter
 from dagster._core.storage.tags import PARTITION_NAME_TAG, PARTITION_SET_TAG
 from dagster._utils.merger import merge_dicts
+from .asset_key import GrapheneAssetKey
 
 from dagster_graphql.implementation.fetch_partition_sets import (
     get_partition_by_name,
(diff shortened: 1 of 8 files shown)


python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -56,7 +56,7 @@ class AssetBackfillData(NamedTuple):
     def is_complete(self) -> bool:
         return (
             (
-                self.requested_subset | self.failed_and_downstream_subset
+                self.materialized_subset | self.failed_and_downstream_subset
```
sryza, line 59: If I understand correctly, a consequence of this change is that the record for this backfill in the bulk actions table will no longer be marked as COMPLETE until all the targeted asset partitions have either been materialized, failed, or skipped.

This means that the meaning of COMPLETE in the bulk actions table will be different for asset backfills than for job backfills.

@prha do you have thoughts on whether this will have consequences? The alternative would be to introduce a new bulk action status that encodes that all the work requested on behalf of the bulk action is complete. I'm having trouble coming up with a good name for this, but haven't thought it too hard about it yet.

prha: Backfill status is an overloaded term.

The original intent of this column is purely to track state on the behalf of the backfill daemon.  The backfill daemon has completed its work for a given backfill job if it has requested a run for every partition.  It can then filter by this column to find more work that it should do.

This is completely separate from the user-facing concept of backfill status (i.e. have all the runs completed successfully for a backfill).  We make sure the UI reflects the individual run status rather than just reading from the backfill status column.

For asset backfills, even though the particulars are different, we should still maintain this split between the concept of "complete" from the daemon perspective (all runs requested), and from the user perspective (backfill is done).

I'm not as up to speed on the asset backfill particulars (the logic is quite complicated and involves the cache queryer).  If a run fails, will the backfill retry?  If not, then I think "complete" for the daemon means that all runs are requested.  That ensures that we won't request duplicate runs on the next backfill iteration.

sryza: > If not, then I think "complete" for the daemon means that all runs are requested. 

The potential issue here is that this PR introduces a distinction between:
- "We no longer need to request any more runs for this backfill"
- "We no longer need to process this backfill within backfill iterations"

I.e., with this PR, we want to keep loading asset backfills after we're done requesting runs, in order to update the `AssetBackfillData` object. This object already needs to keep track of the final status (materialized/failed) of some of the asset partitions in the backfill, so if we can make `AssetBackfillData` keep track of _all_ of them, it makes it simpler and cheaper to query for info about the backfill's progress for the UI.

This means that we need some status that can allow us to distinguish between:
- Backfills where all the requested runs have completed, and thus we should never even load them in the backfill daemon
- Backfills that we want to load in the backfill daemon, either to request new runs for or just to update the data we're tracking for them

> I'm not as up to speed on the asset backfill particulars (the logic is quite complicated and involves the cache queryer). If a run fails, will the backfill retry? 

No

prha: > Backfills where all the requested runs have completed, and thus we should never even load them in the backfill daemon
> Backfills that we want to load in the backfill daemon, either to request new runs for or just to update the data we're tracking for them

As long as the backfill daemon can distinguish between the two (and avoid rerequesting runs unnecessarily), I'm fine with `BulkActionStatus.COMPLETE` to represent the latter.  It's not clear to me that having another state to keep track of will help avoid future confusion.

Hopefully we can entomb some of this discussion in comments in this PR.

python_modules/dagster/dagster/_core/execution/backfill.py
```diff
@@ -126,6 +126,44 @@ def is_valid_serialization(self, workspace: IWorkspace) -> bool:
         else:
             return True
 
+    def get_num_partitions_targeted_by_asset(self, workspace: IWorkspace) -> Mapping[AssetKey, int]:
```
sryza, line 129: Maybe worth combining these two new methods into a single one that returns a single object with all the data, so that we don't need to deserialize the backfill data and build the asset graph both times?

python_modules/dagster-graphql/dagster_graphql/schema/partition_sets.py
```diff
@@ -121,6 +122,32 @@ class Meta:
         name = "PartitionStatusesOrError"
 
 
+class GraphenePartitionsStatus(graphene.Enum):
```
sryza, line 125: Is this built from `BackfillPartitionsStatus`? If so, should it be `GrapheneBackfillPartitionsStatus`?

python_modules/dagster-graphql/dagster_graphql/schema/partition_sets.py
```diff
@@ -121,6 +122,32 @@ class Meta:
...
+        name = "PartitionBulkActionStatusCounts"
+
+
+class GrapheneAssetPartitionsStatusCounts(graphene.ObjectType):
+    class Meta:
+        name = "assetPartitionsStatusCounts"
+
+    assetKey = graphene.NonNull(GrapheneAssetKey)
+    numPartitionsTargeted = graphene.NonNull(graphene.Int)
+    partitionsStatusCounts = non_null_list(GraphenePartitionsStatusCounts)
```
sryza, line 148: Rather than introduce `GraphenePartitionsStatusCounts` and `GraphenePartitionsStatus`, are there drawbacks to just doing:
```python
numPartitionsRequested = graphene.NonNull(graphene.Int)
numPartitionsCompleted = graphene.NonNull(graphene.Int)
numPartitionsFailed = graphene.NonNull(graphene.Int)
```

clairelin135: i don't think so, probably cleaner to have it in individual resolvers as you suggested

sryza (approved):
Nice

clairelin135:
> One idea would be to have a separate mutation in cloud that adds a tag with the user email/id.
We'll also have a separate cloud query that uses the tag to resolve the user. That way the UI can fetch the picture / name in cloud.
@salazarm this makes sense! I'm happy to break these changes out into a separate PR, since they seem pretty separate from the changes in this PR.

PR #13229, opened 2023-03-29
[cached_method] performance optimizations

alangenfeld:
Some tweaks based on profiling of code that frequently invokes cached_method - notably on a method with no args. 

## How I Tested These Changes

captures from speedscope samples out of py-spy, looking at `helper` (the cached_method fn) in the sandwich view

https://gist.github.com/alangenfeld/74a5c77fddab62c871bc80161e087d22

### DIFF
--- a/python_modules/dagster/dagster/_utils/cached_method.py
+++ b/python_modules/dagster/dagster/_utils/cached_method.py
@@ -16,6 +16,7 @@ class _Sentinel:
 
 
 NO_VALUE_IN_CACHE_SENTINEL: Final = _Sentinel()
+NO_ARGS_HASH_VALUE = 0
 
 
 def cached_method(method: Callable[Concatenate[S, P], T]) -> Callable[Concatenate[S, P], T]:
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_utils/cached_method.py
```diff
@@ -107,18 +108,23 @@ def _make_key(
...
-            "@cached_method does not support non-keyword arguments, because doing so would enable "
-            "functionally identical sets of arguments to correpond to different cache keys."
-        ),
-    )
-    key = tuple()
-    if kwds:
-        key = tuple(sorted(kwds.items()))
-    else:
-        key = tuple()
-    if len(key) == 1 and type(key[0]) in fasttypes:
```
alangenfeld, line 122: `type(key[0])` was always `tuple` so this never fired

python_modules/dagster/dagster/_utils/cached_method.py
```diff
@@ -16,6 +16,7 @@ class _Sentinel:
 
 
 NO_VALUE_IN_CACHE_SENTINEL: Final = _Sentinel()
+NO_ARGS_HASH_VALUE = 0
```
alangenfeld, line 19: can do something else but i think it doesn't matter

PR #13240, opened 2023-03-29
Allow sensor context constructors to take Definitions

### LINKED ISSUE #13207: Testing sensors is not compatible with Definitions
### Dagster version

1.2.3

### What's the issue?

I am writing tests of a multi-asset sensor. I need to give it a context. [The docs](https://docs.dagster.io/concepts/partitions-schedules-sensors/asset-sensors#multi-asset-sensors) tell me to use [build_multi_asset_sensor_context](https://docs.dagster.io/_apidocs/schedules-sensors#dagster.build_multi_asset_sensor_context).

That function has a required argument of 'repository_def: RepositoryDefinition`. So to use it, I *must* provide a repository_def.

But I don't have a repository definition, since I'm using Definitions. A different page in the docs (the [one for Repositories](https://docs.dagster.io/concepts/repositories-workspaces/repositories#repositories)) has an admonition on the top:
> In 1.1.6, we introduced [Definitions](https://docs.dagster.io/_apidocs/definitions#dagster.Definitions), which replaces repositories. While repositories will continue to work, we recommend migrating to Definitions

If Definitions are preferred, I think build_multi_asset_sensor_context should accept them.

Maybe the docs should be consistent in their recommendation, or explain how to work with either system by providing conversion tools, or maybe the build_multi_asset_sensor_context function should accept a Definitions.

### How to reproduce?

n/a

### Deployment type

None

### Deployment details

_No response_

### Additional information

_No response_

smackesey:
## Summary & Motivation

Allows sensor context constructors to take a `Definitions` object instead of a `RepositoryDefinition`. Resolves #13207.

## How I Tested These Changes

Modified existing sensor context tests.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py
@@ -32,6 +32,7 @@ from dagster._core.errors import (
 )
 from dagster._core.instance import DagsterInstance
 from dagster._core.instance.ref import InstanceRef
+from dagster._utils import normalize_to_repository
 
 from .events import AssetKey
 from .run_request import RunRequest, SensorResult, SkipReason
@@ -48,6 +49,7 @@ from .target import ExecutableDefinition
 from .utils import check_valid_name
 
 if TYPE_CHECKING:
+    from dagster._core.definitions.definitions_class import Definitions
     from dagster._core.definitions.repository_definition import RepositoryDefinition
     from dagster._core.storage.event_log.base import EventLogRecord
 
@@ -185,7 +187,9 @@ class MultiAssetSensorEvaluationContext(SensorEvaluationContext):
         monitored_assets (Union[Sequence[AssetKey], AssetSelection]): The assets monitored
             by the sensor. If an AssetSelection object is provided, it will only apply to assets
             within the Definitions that this sensor is part of.
-        repository_def (RepositoryDefinition): The repository that the sensor belongs to.
+        repository_def (Optional[RepositoryDefinition]): The repository that the sensor belongs to.
+            If needed by the sensor top-level resource definitions will be pulled from this repository.
+            You can provide either this or `definitions`.
         instance_ref (Optional[InstanceRef]): The serialized instance configured to run the schedule
         cursor (Optional[str]): The cursor, passed back from the last sensor evaluation via
             the cursor attribute of SkipReason and RunRequest. Must be a dictionary of asset key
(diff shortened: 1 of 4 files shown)


smackesey:
* master
  * **PR #13240** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13240" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13240?utm_source=stack-comment).

smackesey:
>We're moving away from the name "repository" entirely, so I think that we should expose an argument that doesn't require users to use that work. Probably just definitions?

Experimented with this but it causes problems due to argument ordering. In `build_multi_asset_sensor_context`, `monitored_assets` is required second arg. So we can't make `repository_def`, first arg, optional, which we would need to do if we were exposing a second `definitions` arg.

I think the way to go here is just to make `repository_def` optionally take `Definitions` wherever it is publicly exposed, then have unified warnings/messaging and maybe change that arg name to `definitions` once `Repository` is deprecated.

sryza:
`build_multi_asset_sensor_context` is experimental - does that solve the problem?

python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py
```diff
@@ -182,7 +184,8 @@ class MultiAssetSensorEvaluationContext(SensorEvaluationContext):
         monitored_assets (Union[Sequence[AssetKey], AssetSelection]): The assets monitored
             by the sensor. If an AssetSelection object is provided, it will only apply to assets
             within the Definitions that this sensor is part of.
-        repository_def (RepositoryDefinition): The repository that the sensor belongs to.
+        repository_def (Union[Definitions, RepositoryDefinition]): The repository or Definitions
```
yuhan, line 187: i think this should be a separate arg, maybe just `definitions`

smackesey:
>build_multi_asset_sensor_context is experimental - does that solve the problem?

I don't think so, because we want the API to match that of `build_sensor_context`, which is not experimental.

python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py
smackesey: see existing convo w sandy

smackesey:
@sryza ok, updated

sryza (commented):
Does MultiAssetSensorEvaluationContext need to support both?

Also, not a blocker for this PR, but might be worth adding a wildcard argument to `build_multi_asset_sensor_context` so we don't find ourselves in a similar situation in the future?

smackesey:
>Does MultiAssetSensorEvaluationContext need to support both?

I think we should support both everywhere that our public API accepts a `RepositoryDefinition` argument.

>might be worth adding a wildcard argument to build_multi_asset_sensor_context so we don't find ourselves in a similar situation in the future?

Do you mean bare `*` to enforce keyword only?

sryza:
> I think we should support both everywhere that our public API accepts a RepositoryDefinition argument.

In that case, worth updating to take a `definitions` argument like you already updated the functions to do so.

> Do you mean bare * to enforce keyword only?

Yeah, exactly.

sryza (approved):

PR #13244, opened 2023-03-29
[pythonic config] Add ConfigurableResourceFactory base class

benpankow:
## Summary

Adds a new `ConfigurableResourceFactory` class based on offline discussion with @schrockn. This doesn't lead to a functional change, but makes the new resource system a bit more self-documenting for new users.

Now, there's two base classes, one you use (`ConfigurableResource`) for the case where the resource class itself is passed to user code:

```python
    a_str: str


@op
def my_op(foo_resource: MyResource):
    print(foo_resource.a_str)
```

And one you use (`ConfigurableResourceFactory`) when the class produces a plain Python value which is passed to user code:

```python
class MyResourceFactory(ConfigurableResourceFactory[str]):
    a_str: str

    def construct_resource(self):
        return self.a_str
    
@op
def my_op(foo_resource: str, context):
    print(foo_resource)
```

(Previously, users would just use `ConfigurableResource` and override `construct_resource`, but having these as two separate base classes makes explaining it in docs nicer.)

## Test Plan

Existing unit tests.

### DIFF
--- a/python_modules/dagster/dagster/__init__.py
+++ b/python_modules/dagster/dagster/__init__.py
@@ -105,6 +105,7 @@ from dagster._config.structured_config import (
     ConfigurableLegacyIOManagerAdapter as ConfigurableLegacyIOManagerAdapter,
     ConfigurableLegacyResourceAdapter as ConfigurableLegacyResourceAdapter,
     ConfigurableResource as ConfigurableResource,
+    ConfigurableResourceFactory as ConfigurableResourceFactory,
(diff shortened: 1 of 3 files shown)


benpankow:
* master
  * **PR #13244** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13244" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13244?utm_source=stack-comment).

Issue #12470, opened 2023-02-22
Global op concurrency limits across all runs (without needing Celery)


gibsondan:
### What's the use case?

Setting a global limit across all runs for the number of ops with a given tag that can run at a given time, for example to protect access to a shared resources. Right now you can apply run-level concurrency in Dagster, and op-level concurrency within a single run, but you can't apply a global limit for an op (or asset) across all runs.

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

yeachan153:
Hi @gibsondan, this would be indeed nice to have as a feature. Do you have an idea how you'd implement this? I thought of a couple of approaches and wanted to get your feedback. The first one was implemented, and it seems to work as expected.

**Approach 1**:
Create a new table in the DB that contains the name of the pool, occupied number of slots for that pool and the total number of slots for that pool as columns. The executor increments the occupied number of slots after checking the current figure is below the total number of allowed slots for that pool. If satisfied, the OP is allowed to start. During this check and increment, we use a lock that prevents reading + updating the number of occupied slots on stale values (think the FOR UPDATE lock in Postgres would do this). Once the task is no longer running (i.e. terminated, success, failure), decrement the number of occupied slots.

We tried this implementation and it seemed to work. We made changes to this function by adding the incrementing functionality: https://github.com/dagster-io/dagster/blob/b8c20ca5a86afd901e82bba325aa58fe09f3157c/python_modules/dagster/dagster/_core/execution/plan/active.py#L277, and decremented the number of occupied slots in this function conditionally: https://github.com/dagster-io/dagster/blob/b8c20ca5a86afd901e82bba325aa58fe09f3157c/python_modules/dagster/dagster/_core/execution/plan/active.py#L456.

It would have been nice to pass the pool config as an OP level config, however that would have been a more intrusive change so for now we extended the executor configuration to allow setting which pool to use for which OP.

Pros:
+ Seems like quite simple unintrusive change
+ We just decrement/increment values atomically in one column (occupied slots) for a slot, so doesn't add too much IO operations on the DB.

Cons:
- Slightly worried what would happen in case there's a connectivity issue (not only due to networking, but maybe the DB itself is down for instance) and we fail to decrement the number of occupied slots.
- Session per run, which is quite a few more connections to the DB

**Approach 2:**
[...]

PR #13250, opened 2023-03-30
Pool implementation for multiprocess executor

yeachan153:
## Summary & Motivation
Not meant to be a mergeable PR, more as a response to https://github.com/dagster-io/dagster/issues/12470#issuecomment-1487125740.

An implementation we're trying for OP concurrency control across jobs using OP tags. We're using the multi-process executor and deploying via Kubernetes.

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/plan/active.py
+++ b/python_modules/dagster/dagster/_core/execution/plan/active.py
@@ -17,7 +17,6 @@ from typing import (
 
 from typing_extensions import Self
 
-from pools.pools import decrement_pool, engine, increment_pool
 from psycopg2.errors import LockNotAvailable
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/execution/plan/active.py
```diff
@@ -305,6 +317,19 @@ def get_steps_to_execute(
             if limit is not None and len(batch) >= limit:
                 break
 
+            if "pool" in step.tags and increment:
```
yeachan153, line 320: Could just move this part out to a separate function for now, but put it here since all the checks for concurrency seem to be done here..

smackesey:
cc @gibsondan

cmpadden:

PR #13254, opened 2023-03-30
[serdes] bottom-up deserialize

alangenfeld:
Use `json.loads` `object_hook` to unpack `serdes` entries as we decode the json instead of recursive descent over a fully parsed object tree. 

https://docs.python.org/3/library/json.html#json.load

## How I Tested These Changes

for a benchmark of loading varying serdes payloads 10 times
CPU time:  14s -> 7s
Peak Mem: 217.6MB -> 142.8MB
Allocations: 425461 -> 325478

Before and After:
Screen Shot 2023-03-30 at 9 25 50 AM
Screen Shot 2023-03-30 at 9 26 34 AM

Before
```
Target 0 size 826.7KiB:
deserialize_value (25 samples): mean 28.857, stdev 1.431, min 27.118
serialize_value   (25 samples): mean 30.708, stdev 4.840, min 27.711

Target 1 size 4.2KiB:
deserialize_value (25 samples): mean 0.449, stdev 0.316, min 0.304
serialize_value   (25 samples): mean 0.294, stdev 0.093, min 0.255

Target 2 size 3.6KiB:
deserialize_value (25 samples): mean 0.330, stdev 0.102, min 0.277
serialize_value   (25 samples): mean 0.260, stdev 0.013, min 0.239

Target 3 size 8.8KiB:
deserialize_value (25 samples): mean 0.520, stdev 0.171, min 0.423
serialize_value   (25 samples): mean 0.486, stdev 0.184, min 0.417

Target 4 size 1.8MiB:
deserialize_value (25 samples): mean 78.040, stdev 8.390, min 72.277
serialize_value   (25 samples): mean 80.270, stdev 6.421, min 74.635

Target 5 size 23.8MiB:
deserialize_value (25 samples): mean 1076.835, stdev 20.284, min 984.325
serialize_value   (25 samples): mean 1071.841, stdev 9.865, min 1060.849
```

After
```
Target 0 size 826.7KiB:
deserialize_value (25 samples): mean 15.307, stdev 0.267, min 15.024
serialize_value   (25 samples): mean 26.967, stdev 0.406, min 26.309

Target 1 size 4.2KiB:
deserialize_value (25 samples): mean 0.227, stdev 0.076, min 0.190
serialize_value   (25 samples): mean 0.234, stdev 0.013, min 0.218

Target 2 size 3.6KiB:
deserialize_value (25 samples): mean 0.201, stdev 0.013, min 0.185
serialize_value   (25 samples): mean 0.216, stdev 0.012, min 0.202

Target 3 size 8.8KiB:
deserialize_value (25 samples): mean 0.325, stdev 0.104, min 0.261
serialize_value   (25 samples): mean 0.433, stdev 0.028, min 0.385

Target 4 size 1.8MiB:
deserialize_value (25 samples): mean 45.223, stdev 2.878, min 41.869
serialize_value   (25 samples): mean 76.347, stdev 4.487, min 71.088

Target 5 size 23.8MiB:
deserialize_value (25 samples): mean 562.411, stdev 24.907, min 458.196
serialize_value   (25 samples): mean 1038.115, stdev 9.249, min 1028.231
```

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/events.py
+++ b/python_modules/dagster/dagster/_core/definitions/events.py
@@ -457,11 +457,11 @@ class AssetMaterializationSerializer(NamedTupleSerializer):
     # There are old `Materialization` objects in storage. We set the default value for asset key to
     # be `AssetKey(["__undefined__"])` to ensure that we can load these objects, without needing to
     # allow for the construction of new `AssetMaterialization` objects with no defined AssetKey.
-    def before_unpack(self, **raw_dict: Any) -> Any:
+    def before_unpack(self, context, unpacked_dict: Any) -> Any:
         # cover both the case where "asset_key" is not present at all and where it is None
-        if raw_dict.get("asset_key") is None:
-            raw_dict["asset_key"] = {"__class__": "AssetKey", "path": UNDEFINED_ASSET_KEY_PATH}
-        return raw_dict
+        if unpacked_dict.get("asset_key") is None:
+            unpacked_dict["asset_key"] = AssetKey(UNDEFINED_ASSET_KEY_PATH)
+        return unpacked_dict
 
 
 @whitelist_for_serdes(
--- a/python_modules/dagster/dagster/_core/definitions/metadata/__init__.py
+++ b/python_modules/dagster/dagster/_core/definitions/metadata/__init__.py
@@ -25,9 +25,9 @@ from dagster._serdes import whitelist_for_serdes
 from dagster._serdes.serdes import (
     FieldSerializer,
     PackableValue,
+    UnpackContext,
     WhitelistMap,
     pack_value,
-    unpack_value,
 )
 from dagster._utils.backcompat import (
     canonicalize_backcompat_args,
@@ -961,21 +961,11 @@ class MetadataFieldSerializer(FieldSerializer):
 
     def unpack(
         self,
-        metadata_entries: List[Mapping[str, Any]],
+        metadata_entries: List["MetadataEntry"],
         whitelist_map: WhitelistMap,
-        descent_path: str,
+        context: UnpackContext,
     ) -> Mapping[str, MetadataValue]:
-        return {
-            e["label"]: unpack_value(
-                e["entry_data"],
-                # MetadataValue itself can't inherit from NamedTuple and so isn't a PackableValue,
-                # but one of its subclasses will always be returned here.
-                as_type=MetadataValue,  # type: ignore
-                whitelist_map=whitelist_map,
-                descent_path=descent_path,
-            )
-            for e in metadata_entries
-        }
+        return {e.label: e.entry_data for e in metadata_entries}
 
 
 T_MetadataValue = TypeVar("T_MetadataValue", bound=MetadataValue, covariant=True)
--- a/python_modules/dagster/dagster/_daemon/types.py
+++ b/python_modules/dagster/dagster/_daemon/types.py
@@ -1,21 +1,23 @@
-from typing import Any, NamedTuple, Optional, Sequence
+from typing import NamedTuple, Optional, Sequence
 
 import dagster._check as check
 from dagster._serdes import whitelist_for_serdes
-from dagster._serdes.serdes import NamedTupleSerializer, is_packed_enum
+from dagster._serdes.serdes import NamedTupleSerializer, UnknownSerdesValue
 from dagster._utils.error import SerializableErrorInfo
 
 
 class DaemonHeartbeatSerializer(NamedTupleSerializer["DaemonHeartbeat"]):
-    def before_unpack(self, **packed_dict: Any):
+    def before_unpack(self, context, unpacked_dict):
         # Previously daemon types were enums, now they are strings. If we find a packed enum,
         # just extract the name, which is the string we want.
-        if is_packed_enum(packed_dict.get("daemon_type")):
-            packed_dict["daemon_type"] = packed_dict["daemon_type"]["__enum__"].split(".")[-1]
-        if packed_dict.get("error"):
-            packed_dict["errors"] = [packed_dict["error"]]
-            del packed_dict["error"]
-        return packed_dict
+        if isinstance(unpacked_dict.get("daemon_type"), UnknownSerdesValue):
+            unknown = unpacked_dict["daemon_type"]
+            unpacked_dict["daemon_type"] = unknown.value["__enum__"].split(".")[-1]
+            context.clear_ignored_unknown_values(unknown)
+        if unpacked_dict.get("error"):
+            unpacked_dict["errors"] = [unpacked_dict["error"]]
+            del unpacked_dict["error"]
+        return unpacked_dict
 
 
 @whitelist_for_serdes(serializer=DaemonHeartbeatSerializer)
--- a/python_modules/dagster/dagster/_serdes/serdes.py
+++ b/python_modules/dagster/dagster/_serdes/serdes.py
@@ -12,11 +12,11 @@ Why not pickle?
 * This isn't meant to replace pickle in the conditions that pickle is reasonable to use
   (in memory, not human readable, etc) just handle the json case effectively.
 """
-
 import collections.abc
 import warnings
 from abc import ABC, abstractmethod
 from enum import Enum
+from functools import partial
 from inspect import Parameter, signature
 from typing import (
     AbstractSet,
@@ -28,7 +28,6 @@ from typing import (
     List,
     Mapping,
     NamedTuple,
-    NoReturn,
     Optional,
     Sequence,
     Set,
@@ -39,7 +38,7 @@ from typing import (
     overload,
 )
 
-from typing_extensions import Final, Literal, Self, TypeAlias, TypeGuard, TypeVar
+from typing_extensions import Final, Self, TypeAlias, TypeVar
 
 import dagster._check as check
 import dagster._seven as seven
@@ -74,6 +73,21 @@ PackableValue: TypeAlias = Union[
     Set["PackableValue"],
     FrozenSet["PackableValue"],
     Enum,
+]
+
+UnpackedValue: TypeAlias = Union[
+    Sequence["UnpackedValue"],
+    Mapping[str, "UnpackedValue"],
+    str,
+    int,
+    float,
+    bool,
+    None,
+    NamedTuple,
+    Set["PackableValue"],
+    FrozenSet["PackableValue"],
+    Enum,
+    "UnknownSerdesValue",
 ]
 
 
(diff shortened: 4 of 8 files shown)


alangenfeld:
* master
  * **PR #13254** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13254" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13254?utm_source=stack-comment).

alangenfeld:
notes on handling unknown types when bottom-up unpacking 
* returning `UnknownSerdesData` 
  * effects `before_unpack` implementations 
* returning the packed dict value and tracking by id 
  * causes issues when building up
  * cleverly allows before_unpack to work but not in a clear way

python_modules/dagster/dagster/_core/definitions/events.py
```diff
@@ -442,11 +442,11 @@ class AssetMaterializationSerializer(NamedTupleSerializer):
...
     # be `AssetKey(["__undefined__"])` to ensure that we can load these objects, without needing to
     # allow for the construction of new `AssetMaterialization` objects with no defined AssetKey.
-    def before_unpack(self, **raw_dict: Any) -> Any:
+    def before_unpack(self, context, storage_dict: Any) -> Any:
         # cover both the case where "asset_key" is not present at all and where it is None
-        if raw_dict.get("asset_key") is None:
-            raw_dict["asset_key"] = {"__class__": "AssetKey", "path": UNDEFINED_ASSET_KEY_PATH}
-        return raw_dict
+        if storage_dict.get("asset_key") is None:
+            storage_dict["asset_key"] = AssetKey(UNDEFINED_ASSET_KEY_PATH)
```
alangenfeld, line 448: eagerly unpacking to real types is fine as they flow through unpack untouched

python_modules/dagster/dagster/_core/definitions/metadata/__init__.py
```diff
@@ -945,37 +948,44 @@ def pack(
...
+                    whitelist_map=whitelist_map,
+                    context=context,
+                )
+                for e in metadata_entries
+            }
+        elif isinstance(context, BottomUpUnpackContext):
+            metadata_entries = cast(List[MetadataEntry], metadata_entries)
+            return {e.label: e.entry_data for e in metadata_entries}
+        else:
+            check.failed(f"unexpected serdes context {context}")
```
alangenfeld, line 988: Complexity unleashed by this change is having two modes of unpacking. One bottom up when we are deserializing from a string and the other recursive descent when we are unpacking an already inflated structure. 

We can handle this forked behavior in different ways, such as having two `unpack` methods if we think thats better.

python_modules/dagster/dagster/_daemon/types.py
```diff
@@ -1,21 +1,34 @@
...
-        return packed_dict
+
+        if is_packed_enum(storage_dict.get("daemon_type")):
+            storage_dict["daemon_type"] = storage_dict["daemon_type"]["__enum__"].split(".")[-1]
+        elif isinstance(storage_dict.get("daemon_type"), UnknownSerdesValue):
+            if not isinstance(context, BottomUpUnpackContext):
+                check.failed(f"Unexpected state, got UnknownSerdesValue with context {context}")
+            unknown = storage_dict["daemon_type"]
+            storage_dict["daemon_type"] = unknown.value["__enum__"].split(".")[-1]
+            context.clear_ignored_unknown_values(unknown)
```
alangenfeld, line 26: gnarly to have to do this clearing in this custom unpack but not too gnarly

python_modules/dagster/dagster/_serdes/serdes.py
```diff
@@ -314,6 +313,79 @@ def __whitelist_for_serdes(klass: T_Type) -> T_Type:
...
+            self._observed_unknown_serdes_values.discard(obj)
+        elif isinstance(obj, (list, set, frozenset)):
+            for inner in obj:
+                self.clear_ignored_unknown_values(inner)
+        elif isinstance(obj, dict):
+            for k, v in obj.items():
+                self.clear_ignored_unknown_values(k)
+                self.clear_ignored_unknown_values(v)
+
+        return obj
```
alangenfeld, line 375: This somewhat complex pattern of observe/clear/assert was the only way I could figure out to 
* catch things in nested structures 
* fire the `handle_unpack_error` at the right time
* be performant

python_modules/dagster/dagster/_core/definitions/metadata/__init__.py
gibsondan: what's the type of the 'inflated structure' being referred to here? Some internal json object representation?

I think i would find this clearer fwiw if they're not sharing a lot of code - unpack_string vs. unpack_<answer to question above>

gibsondan:
I see why memory is much faster (assuming the above is correct) but i'm surprised that there are big CPU wins too

python_modules/dagster/dagster/_core/definitions/metadata/__init__.py
alangenfeld: > what's the type of the 'inflated structure' being referred to here? Some internal json object representation?

The default `json.loads` output for JSON object is `dict` - so that tree of `dict` / `list` containers that we recurse over in search of the special `dict` entries containing the `__class__` / `__enum__` keys that trigger special unpack behavior 

> I think i would find this clearer fwiw if they're not sharing a lot of code - unpack_string vs. unpack_

👍  curious to hear what others think too- can take a pass at harder bifurcation of paths

alangenfeld:
The current top-down recursive implementation eagerly builds new `list` / `dict` containers as it descends, and since we support serdes object anywhere in the tree, we recurse all the way to the leaves.

In the new bottom-up approach, `list` are untouched and only the special `dict` containers are discarded and replaced with tuples/enums.

python_modules/dagster/dagster/_daemon/types.py
alangenfeld: `[1]`

python_modules/dagster/dagster/_core/definitions/metadata/__init__.py
alangenfeld: > Is there a way to prevent deserialization into MetadataEntry by this bottom-up process?

I believe the way this would be handled once `MetadataEntry` is removed is that you would get a `UnknownSerdesValue` which you would then process the `.value` of like we do in `[1]`

prha: > recursive descent when we are unpacking an already inflated structure

I'm still not clear why we would need the `RecursiveDescentContext` in an unpack... when do we have the inflated structure?

smackesey: It's exclusively for when you are calling `unpack_value` directly (i.e. not going through `json.loads`). @alangenfeld and I talked offline about it and think it's possible to change `unpack_value` to just go bottom up as well, which eliminates the need for two kinds of contexts.

gibsondan (changes requested):
Bubbling up sean's comment: "it's possible to change unpack_value to just go bottom up as well, which eliminates the need for two kinds of contexts."

Is this the plan of record for this PR now or would that be a followup? Certainly sounds nice to have a single algorithm, to your queue for that question

smackesey:
>Is this the plan of record for this PR now or would that be a followup? Certainly sounds nice to have a single algorithm, to your queue for that question

This is implemented in a PR I stacked on this which alex will squash in

python_modules/dagster/dagster/_serdes/serdes.py
```diff
@@ -314,6 +328,75 @@ def __whitelist_for_serdes(klass: T_Type) -> T_Type:
     return __whitelist_for_serdes
 
 
+###################################################################################################
+# Serializers
+###################################################################################################
+
+
+class PackContext:
+    """values are packed/unpacked by recursing from the top."""
```
gibsondan, line 337: is packed/unpacked accurate?

gibsondan (approved):

python_modules/dagster/dagster/_serdes/serdes.py
```diff
@@ -362,47 +450,70 @@ def __init__(
...
         except Exception as exc:
-            return self.handle_unpack_error(exc, **storage_dict)
+            value = self.handle_unpack_error(exc, context, unpacked_dict)
+            if isinstance(context, UnpackContext):
+                context.assert_no_unknown_values(value)
+                context.clear_ignored_unknown_values(unpacked_dict)
+            return value
 
-    # Hook: Modify the contents of the loaded dict before it is unpacked into domain objects during
+    # Hook: Modify the contents of the unpacked dict before domain object construction during
```
smackesey, line 493: "unpack" has meant object construction up until now-- is there something that distiguishes the "unpacked dict" you're referring to above from the raw deserialized dict?

smackesey: oh nm I guess it's that the children have been constructed already

python_modules/dagster/dagster/_serdes/serdes.py
```diff
@@ -362,47 +450,70 @@ def __init__(
...
                 # Naively implements backwards compatibility by filtering arguments that aren't present in
                 # the constructor. If a property is present in the serialized object, but doesn't exist in
                 # the version of the class loaded into memory, that property will be completely ignored.
                 if loaded_name in self.constructor_param_names:
-                    unpack_fn = self.get_field_unpack_fn(field=loaded_name)
-                    unpacked[loaded_name] = unpack_fn(
-                        value, whitelist_map=whitelist_map, descent_path=descent_path
-                    )
+                    # custom unpack regardless of hook vs recursive descent
+                    custom = self.field_serializers.get(loaded_name)
```
smackesey, line 467: I'm not following the reason for the change/expansion in complexity here-- why doesn't resolving one "unpack" function work and then performing a single assignment with its result work?

smackesey:
This looks like its about there, but one thing I'm a little hesitant about is the expansion in terminology. Prior to this PR, there is:

- serialize/deserialize: domain objects <-> strings
- pack/unpack: domain objects <-> basic python collections

The serializer hooks and logic reflects this terminology, with `before_unpack`, `after_pack`. IMO this makes it easy to follow.

This PR is now introducing two new verbs into the serdes logic-- `process` and `construct`, while still retaining "unpack" which now means something else. Why not keep the old meaning of "unpack", which AFAICT still encompasses the meaning of both of these new verbs? It means the same as before except that the children/elements of collections have already themselves been unpacked.

python_modules/dagster/dagster/_serdes/serdes.py
alangenfeld: this code path is very hot, `get_field_unpack_fn` was showing up as material in profiler results so i inlined it

alangenfeld:
> Why not keep the old meaning of "unpack", which AFAICT can still encompass the meaning of both new verbs? It means the same as before except that the children/elements of collections have already themselves been unpacked.

Ya it seemed important to distinguish them while i was changing it but agree with your assessment. Reverted back to the previous names, but did keep the update from `storage_dict` to `unpacked_dict` since i think that still makes sense.

python_modules/dagster/dagster/_serdes/serdes.py
alangenfeld: I did some additional inlining in this vein above and was able to drop cpu further 
```
Target 0 size 826.7KiB:
mean time from 10 samples: 17.147ms

Target 1 size 4.2KiB:
mean time from 10 samples: 0.149ms

Target 2 size 3.6KiB:
mean time from 10 samples: 0.139ms

Target 3 size 8.8KiB:
mean time from 10 samples: 0.195ms

Target 4 size 1.8MiB:
mean time from 10 samples: 47.278ms

Target 5 size 23.8MiB:
mean time from 10 samples: 474.157ms
```

smackesey: Interesting, ok

smackesey (approved):
very nice

Issue #13249, opened 2023-03-30
define_asset_job does not accept graph-backed assets as selection


aksestok:
### Dagster version

dagster, version 1.2.3

### What's the issue?

From version 1.2.3, `define_asset_job` no longer accepts assets defined using `graph_multi_asset` in the `selection`.

Error from log:
```
2023-03-30 10:57:04 +0200 - dagster - INFO - Launching Dagster services...
<env-path>/lib/python3.11/site-packages/dagster/_core/workspace/context.py:591: UserWarning: Error loading repository location repro.py:dagster._check.CheckError: Invariant failed. Description: Tried to retrieve asset key from an assets definition with multiple asset keys: ["a"], ["b"]

Stack Trace:
  File "<env-path>/lib/python3.11/site-packages/dagster/_grpc/server.py", line 266, in __init__
    self._loaded_repositories: Optional[LoadedRepositories] = LoadedRepositories(
                                                              ^^^^^^^^^^^^^^^^^^^
  File "<env-path>/lib/python3.11/site-packages/dagster/_grpc/server.py", line 115, in __init__
    loadable_targets = get_loadable_targets(
                       ^^^^^^^^^^^^^^^^^^^^^
  File "<env-path>/lib/python3.11/site-packages/dagster/_grpc/utils.py", line 37, in get_loadable_targets
    else loadable_targets_from_python_file(python_file, working_directory)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<env-path>/lib/python3.11/site-packages/dagster/_core/workspace/autodiscovery.py", line 27, in loadable_targets_from_python_file
    loaded_module = load_python_file(python_file, working_directory)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<env-path>/lib/python3.11/site-packages/dagster/_core/code_pointer.py", line 83, in load_python_file
    return import_module_from_path(module_name, python_file)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<env-path>/lib/python3.11/site-packages/dagster/_seven/__init__.py", line 49, in import_module_from_path
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 940, in exec_module
[...]

PR #13259, opened 2023-03-30
fix asset selection coercion for multi-assets

sryza:
## Summary & Motivation

Fixes https://github.com/dagster-io/dagster/issues/13249


smackesey (approved):
Good. This should fix the dbt users' issues too right? (in https://dagster.slack.com/archives/C04CW71AGBW/p1680129158765899).

sryza:
@smackesey ah yes it looks like it should fix that issue reported on slack too

PR #13262, opened 2023-03-30
link to dynamic partitions example from concept page

sryza:


erinkcochran87 (approved):
lgtm! BK is unhappy with something code snippet-related, but otherwise you can :shipit:

Issue #13005, opened 2023-03-17
Table metadata type is not being displayed


danielgafni:
### Dagster version

1.2.2

### What's the issue?

I logged a `TableMetadataValue`, but it's empty in Dagit. 
It has bothed `TableRecord`s and a `TableSchema`. 
![image](https://user-images.githubusercontent.com/49863538/225954345-42aacde5-2b84-4b09-961c-6b6596d4ef4a.png)

### What did you expect to happen?

Expected to see a table

### How to reproduce?

log a `TableMetadataValue`

### Deployment type

Local

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

tacastillo:
Hey hey! Thanks for reporting this! I've attached to a GitHub project so the right people are aware of it.

PR #13267, opened 2023-03-30
Render TableMetadata type in asset definition sidebar

salazarm:
## Summary & Motivation

Fixes https://github.com/dagster-io/dagster/issues/13005

## How I Tested These Changes

Screen Shot 2023-03-30 at 3 32 39 PM

Dialog opens when you click on "Show schema":
Screen Shot 2023-03-30 at 3 40 05 PM

### DIFF
--- a/js_modules/dagit/packages/core/src/metadata/MetadataEntry.tsx
+++ b/js_modules/dagit/packages/core/src/metadata/MetadataEntry.tsx
@@ -171,7 +171,13 @@ export const MetadataEntry: React.FC<{
         </MetadataEntryLink>
       );
     case 'TableMetadataEntry':
-      return null;
+      console.log(entry);
+      return (
+        <Box>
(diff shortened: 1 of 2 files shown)


smackesey:
Good to see progress here-- that's the `TableSchema`, but what about the actual table values? For v1 of this  I actually don't think we should render the schema attached to a `TableMetadataValue` at all (you can use `TableSchemaMetadataValue`) for that.

salazarm:
> Good to see progress here-- that's the TableSchema, but what about the actual table values? For v1 of this I actually don't think we should render the schema attached to a TableMetadataValue at all (you can use TableSchemaMetadataValue) for that.

The graphql object on the client didnt have the values so I assumed we didn't want to show them? Do we need to update the resolvers for this too then? Do we know what we want to show here exactly? cc @braunjj

smackesey (approved):
lgtm!

Issue #13256, opened 2023-03-30
[1.2.4] Crash when the DynamicPartition of a MultiPartition is empty


DrawnWren:
### Dagster version

1.2.4

### What's the issue?

When starting Dagit with an asset whose `partitions_def=MultiPartition(TimeWindowPartition,DynamicPartition)` and the dynamic partition is populated entirely during another asset's materialization (so empty @ initialization): 
```Operation name: AssetGraphLiveQuery

Message: Invariant failed. Description: Expected 2 partition keys in partition key string 2020-01-07, but got 1

Path: ["assetNodes",2,"partitionStats"]

Locations: [{"line":53,"column":3}]

Stack Trace:
  File "/home/wing/.cache/pypoetry/virtualenvs/syncdag-NSDSaiv4-py3.11/lib/python3.11/site-packages/graphql/execution/execute.py", line 521, in execute_field
    result = resolve_fn(source, info, **args)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wing/.cache/pypoetry/virtualenvs/syncdag-NSDSaiv4-py3.11/lib/python3.11/site-packages/dagster_graphql/schema/asset_graph.py", line 810, in resolve_partitionStats
    ) = get_partition_subsets(
        ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wing/.cache/pypoetry/virtualenvs/syncdag-NSDSaiv4-py3.11/lib/python3.11/site-packages/dagster_graphql/implementation/fetch_assets.py", line 340, in get_partition_subsets
    updated_cache_value = get_and_update_asset_status_cache_value(
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wing/.cache/pypoetry/virtualenvs/syncdag-NSDSaiv4-py3.11/lib/python3.11/site-packages/dagster/_core/storage/partition_status_cache.py", line 576, in get_and_update_asset_status_cache_value
    updated_cache_value = _get_fresh_asset_status_cache_value(
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wing/.cache/pypoetry/virtualenvs/syncdag-NSDSaiv4-py3.11/lib/python3.11/site-packages/dagster/_core/storage/partition_status_cache.py", line 551, in _get_fresh_asset_status_cache_value
    updated_cache_value = _build_status_cache(
[...]

PR #13268, opened 2023-03-30
Fix cache error when single-dimension partition fails and asset becomes multipartitioned

clairelin135:
Fixes https://github.com/dagster-io/dagster/issues/13256

A graphQL error occurs as described in the issue when:
- an asset with single-dimension partitions def fails for one or more partitions
- the partitions def is changed to a multipartitions def
- the cache attempts to build multipartition keys from the failed partition keys, throwing an error as the failed partition keys are single-dimensioned

This PR fixes this issue.

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/partition_status_cache.py
+++ b/python_modules/dagster/dagster/_core/storage/partition_status_cache.py
@@ -1,4 +1,4 @@
-from typing import Dict, List, NamedTuple, Optional, Sequence, Set, Tuple, cast
+from typing import Dict, List, NamedTuple, Optional, Sequence, Set, Tuple
 
 from dagster import (
     AssetKey,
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster/_core/storage/partition_status_cache.py
```diff
@@ -203,7 +204,11 @@ def get_validated_partition_keys(
         }
         validated_partitions = set()
         for partition_key in partition_keys:
-            multipartition_key = partitions_def.get_partition_key_from_str(partition_key)
+            try:
+                multipartition_key = partitions_def.get_partition_key_from_str(partition_key)
+            except CheckError:
```
johannkm, line 209: this except is a little scary... maybe an is_multipartition_key error?

johannkm (commented):
yikes, I haven't thought much about changing partition definition types. This is a problem with the materialized query too right? Or did it already handle this somewhere?

clairelin135:
actually, it's not an issue with materialized partitions. With materialized partitions we query for the asset event tags table, so we don't find single-dimension partition keys there

python_modules/dagster/dagster/_core/storage/partition_status_cache.py
clairelin135: yeah i'm happy to change this part

johannkm (approved):

PR #13270, opened 2023-03-30
[asset-reconciliation][perf] improve perf on all partition mapping

OwenKephart:
## Summary & Motivation

Improving performance for the case that you have a mapping from [lots of partitions] -> [unpartitioned].

Previously, if you were turning on a sensor for the first time, and there was a large set of materialized partitions upstream of an unpartitioned asset, we would call `get_children_partitions` once for each of those materialized partitions. This could be multiple thousands of times, which would time out the sensor.

Now, we group all the newly-materialized partitions into a PartitionsSubset, making it a lot cheaper to compute the downstream partitions subset in the common case (IdentityPartitionMapping), and also skipping the mapping entirely for AllPartitionMapping.


clairelin135 (approved):
LGTM

PR #13272, opened 2023-03-30
Fix spring DST schedule transitions

gibsondan:
Summary:
- Workarounds for upstream issues with croniter where it sometimes has trouble around DST transitions
- Better handle a spring DST transition in our own day incrementing code that increments things by a day but didn't realize it needed to go from 3AM back to 2AM

### DIFF
--- a/python_modules/dagster/dagster/_utils/schedules.py
+++ b/python_modules/dagster/dagster/_utils/schedules.py
@@ -62,6 +62,7 @@ def cron_string_iterator(
 
     delta_fn = None
     should_hour_change = False
+    expected_hour = None
 
     # Special-case common intervals (hourly/daily/weekly/monthly) since croniter iteration can be
     # much slower than adding a fixed interval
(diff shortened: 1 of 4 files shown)


python_modules/dagster/dagster/_utils/schedules.py
```diff
@@ -44,6 +44,9 @@ def cron_string_iterator(
     # and matches the cron schedule
     next_date = date_iter.get_prev(datetime.datetime)
 
+    if not croniter.match(cron_string, next_date):
+        next_date = date_iter.get_next(datetime.datetime)
```
alangenfeld, line 48: whats this part about?

gibsondan:
added an explanation to @alangenfeld's question plus more comprehensive tests

alangenfeld (approved):
brutal complexity but good comments and tests

the spirit of your phabricator badge lives on 
Screen Shot 2023-03-31 at 11 25 22 AM

PR #13273, opened 2023-03-30
[docs] Add generated documentation for all public methods on InputContext, OutputContext, OpExecutionContext

spenczar:
## Summary & Motivation

I was trying to write an IOManager and couldn't figure out how to get the asset keys associated with the input. The API docs showed no way to do this. I had to read source for other IOManagers to figure it out.

This is because API documentation isn't generated for properties or methods which are lacking docstrings, so I added docstrings and rebuilt docs.

## How I Tested These Changes

Built docs locally and looked around at them.


smackesey:
Thanks @spenczar, I'll review this week

python_modules/dagster/dagster/_core/execution/context/compute.py
```diff
@@ -318,6 +326,12 @@ def partition_time_window(self) -> TimeWindow:
     @public
     @property
     def selected_asset_keys(self) -> AbstractSet[AssetKey]:
+        """AbstractSet[AssetKey]: The set of asset keys which are
```
jamiedemaria, line 329: Typically we format our docstrings like this

```
"""Description of the method/function/property

Args:
      param1 (type): description of param1
     ....

Returns (type): description of the returned value

Raises (type): description of the error that could be raised
"""
```

So for example, this docstring would be 

```
"""
Returns (AbstractSet[AssetKey]): The set of asset keys which are selected for materialization for the current run. If the op does not have an assets definition, returns an empty set.
"""
```
I think since this is a property, you just need the `Returns`. If it were a function, you'd want to split the description up in a way that the initial description contains most of the info, and the `Returns` section just has the type and description of what's returned.

jamiedemaria (changes requested):
Thanks for doing this work! I have a couple comments on formatting to bring your new docstrings more in line with our existing docstrings. I haven't had time to go through each docstring and check its accuracy, but i will do that tomorrow

python_modules/dagster/dagster/_core/execution/context/compute.py
spenczar: I agree that format is common elsewhere, but I am following the pattern in this file. See, for example: https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_core/execution/context/compute.py#L171-L192:
 ```python
    @public
    @property
    def resources(self) -> Any:
        """Resources: The currently available resources."""
        return self._step_execution_context.resources

    @property
    def step_launcher(self) -> Optional[StepLauncher]:
        """Optional[StepLauncher]: The current step launcher, if any."""
        return self._step_execution_context.step_launcher

    @public
    @property
    def run_id(self) -> str:
        """str: The id of the current execution's run."""
        return self._step_execution_context.run_id

    @public
    @property
    def run_config(self) -> Mapping[str, object]:
        """dict: The run config for the current execution."""
        return self._step_execution_context.run_config
```

Note that this abbreviated format is the one that the `napoleon` Sphinx extension (which dagster uses) seems to recommend for public properties (https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). Just to say one is not necessarily better than the other - but I'm happy to change it to be consistent.

Would you like me to bring the rest of the docstrings in this file in line with that format, with a `Returns` key? Or just the ones that I've written?

Comment (jamiedemaria) [reply python_modules/dagster/dagster/_core/execution/context/compute.py:329]: ah i see. in that case it's fine to leave like this. ty for teaching me about our docs system

File: python_modules/dagster/dagster/_core/execution/context/compute.py [hunk 2]
```diff
@@ -347,6 +366,9 @@ def selected_output_names(self) -> AbstractSet[str]:
 
     @public
     def asset_key_for_output(self, output_name: str = "result") -> AssetKey:
+        """
+        AssetKey: The asset key for the given output, referenced by name.
```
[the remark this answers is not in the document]
jamiedemaria: if you want to do it, go ahead, but we can also handle it on our end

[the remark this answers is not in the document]
spenczar: Okay, I'll leave it to just what I've done, for now.

PedramNavid:
Closing stale PR.

Issue #13174, opened 2023-03-27
"RuntimeError: generator raised StopIteration" when backfilling assets with partitions or cross code location dependency


boenshao:
### Dagster version

1.2.3

### What's the issue?

"RuntimeError: generator raised StopIteration" when backfilling assets with cross code location dependency.

When I use "Materialize All" to backfill assets that have cross code location dependency.
![image](https://user-images.githubusercontent.com/65853605/227904328-2ecfa46f-74a3-4337-aea0-09c51bf4c8d8.png)
![image](https://user-images.githubusercontent.com/65853605/227904203-6e9a0887-11cf-4a47-b99c-6f98b100cc1e.png)

Below error occured.
![image](https://user-images.githubusercontent.com/65853605/227903335-0ba8d15c-36b5-4daf-ada6-e7623317cb65.png)

### How to reproduce?

You can find the minimal reproducible example here

- https://bitbucket.org/berry-ai/dagster-asset-backfill-error-report/src/master/

### Deployment type

Docker Compose

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

smackesey:
@clairelin135 @sryza is this a known issue?

boenshao:
I just found out that backfilling an asset with multiple partitions can also trigger the same error.

![image](https://user-images.githubusercontent.com/65853605/228729601-275170ec-d876-47e9-bdf8-3805839cb232.png)

The steps to reproduce the issue have been added to the repo below:
- https://bitbucket.org/berry-ai/dagster-asset-backfill-error-report/src/master/

I'll update the issue's title to reflect this finding.

sryza:
@boenshao thanks for reporting this and for the detailed reproduction steps. I was able to reproduce this.

Here's a PR that fixes the issue: https://github.com/dagster-io/dagster/pull/13275.

PR #13275, opened 2023-03-30
fix ExternalAssetGraph.get_job_names

sryza:
## Summary & Motivation

Fixes https://github.com/dagster-io/dagster/issues/13174

## How I Tested These Changes

Successfully ran a cross-repo backfill on a test code location with two repos. Before the change, I hit the `StopIteration` error mentioned in the bug report.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -78,7 +78,7 @@ class AssetGraph:
         """Non-source asset keys that have no non-source parents."""
         from .asset_selection import AssetSelection
 
-        return AssetSelection.keys(*self.all_asset_keys).sources().resolve(self)
+        return AssetSelection.keys(*self.non_source_asset_keys).roots().resolve(self)
 
     @property
     def freshness_policies_by_key(self) -> Mapping[AssetKey, Optional[FreshnessPolicy]]:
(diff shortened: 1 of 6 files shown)


OwenKephart (approved):
makes sense

python_modules/dagster/dagster/_core/definitions/external_asset_graph.py
```diff
@@ -108,7 +108,9 @@ def from_repository_handles_and_external_asset_nodes(
             if not node.is_source
         }
         job_names_by_key = {
-            node.asset_key: node.job_names for _, node in repo_handle_external_asset_nodes
+            node.asset_key: node.job_names
+            for _, node in repo_handle_external_asset_nodes
+            if not node.is_source
```
smackesey, line 113: IIRC this is reversing a change I made that was part of allowing observable source assets to run in asset jobs, which is probably why the stale scheduler test is failing.

python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -137,6 +137,10 @@ def from_assets(
     def all_asset_keys(self) -> AbstractSet[AssetKey]:
         return self._asset_dep_graph["upstream"].keys()
 
+    @property
+    def non_source_asset_keys(self) -> AbstractSet[AssetKey]:
```
OwenKephart, line 141: we have such an awful ugly trinary thing going on (SourceAsset, AssetsDef, thing that is referenced by an AssetsDef but has no SourceAsset) but this works

PR #13276, opened 2023-03-31
[pythonic config] Allow ConfigMappings to take, return Pythonic config

benpankow:
## Summary

Adds some wrapper code which makes `ConfigMappings` receptive to the new Pythonic config classes - in particular, the input schema can be defined by a Pythonic `Config` and the output can be a `RunConfig` (in addition to the typical config dict). For example:

**Before:**
```python
class DoSomethingConfig(Config):
    config_param: str

@op
def do_something(config: DoSomethingConfig) -> str:
    return config.config_param


@config_mapping(config_schema={"simplified_param": str})
def simplified_config(val):
    return {
        "ops": {
            "do_something": {
                 "config": {
                     "config_param": val["simplified_param"]
                 }
            }
        }
    }

@job(config=simplified_config)
def do_it_all_with_simplified_config() -> None:
    do_something()

result = do_it_all_with_simplified_config.execute_in_process(
    run_config={"simplified_param": "foo"}
)
```

**After:**
```python
class DoSomethingConfig(Config):
    config_param: str

@op
def do_something(config: DoSomethingConfig) -> str:
    return config.config_param

class ConfigMappingConfig(Config):
    simplified_param: str

# New, fancy config mapping takes in a Pythonic config object and returns a RunConfig
@config_mapping
def simplified_config(config_in: ConfigMappingConfig) -> RunConfig:
    return RunConfig(
        ops={"do_something": DoSomethingConfig(config_param=config_in.simplified_param)}
    )

@job(config=simplified_config)
def do_it_all_with_simplified_config() -> None:
    do_something()

result = do_it_all_with_simplified_config.execute_in_process(
    run_config={"simplified_param": "foo"}
)
```

## Test Plan

New set of unit tests, existing tests to test existing `ConfigMapping` functionality.

PR #13277, opened 2023-03-31
[pythonic config] Allow direct binding of RunConfig to @job decorators

benpankow:
## Summary

Adds support for passing `RunConfig` to `@job` decorators instead of config dicts. This is nice because it lets users pass `Config` objects for the ops in question rather than having to expand them to config dicts.

**Before:**
```python
class DoSomethingConfig(Config):
    config_param: str

@op
def do_something(config: DoSomethingConfig) -> str:
    return config.config_param

@job(config={
    "ops": {
        "do_something": {
            "config": {
                "config_param": "foo",
            }
        }
    }
})
def do_it_all_with_baked_in_config() -> None:
    do_something()
```

**After:**
```python
class DoSomethingConfig(Config):
    config_param: str

@op
def do_something(config: DoSomethingConfig) -> str:
    return config.config_param

@job(config=RunConfig(
    ops={
        "do_something": DoSomethingConfig(config_param="foo")
    }
)
def do_it_all_with_baked_in_config() -> None:
    do_something()
```

## Test Plan

Small unit test.


benpankow:
* master
  * **PR #13276** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13276" target="_blank">Graphite</a> 
    * **PR #13277** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13277" target="_blank">Graphite</a>  👈
      * **PR #13278** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13278" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13277?utm_source=stack-comment).

simonvanderveldt:
This is definitely a nicer interface when using the Pythonic configuration/`Config` subclasses because the user no longer has to do the conversion themselves.
Unfortunately because under the hood the `Config` instance is still being converted to a dict I expect validation of the config will still happen twice, once when instantiating the `Config` instance and once when actually running the Job/Op when the Op picks up the dict version of the config. This can/will cause issues in case users include some logic in the `Config` subclasses and might be a bit more confusing since the signature now seems to suggest the Op config natively takes `Config` subclasses.
Ops being able to natively handle `Config` instances internally would not have this issue, but that is probably a lot more work (also because the config can AFAIK also be provided at runtime).

I'm not sure what the long term plans are with regards to config? IMHO the pydantic/pythonic run configuration is quiet a bit nicer than using `config_schema` so I expect the pythonic config to become the default or only way to define config, do you have any idea if that's the plan? I can also understand this might take a while, in that case it might still make sense to merge a change like this to improve things as much as possible without fundamental changes.

schrockn:
@simonvanderveldt 

Yes the double validation will be the behavior and in fact I would argue that it is desirable. Generally the Config instance will be instantiated when booting up the code location. Then the config will be instantiated when run and step processes are booted up as well. This is the behavior we want as lots of things can happen in between config persistence and when it is reloaded for execution. Code can be updated. Someone could manually or accidentally modify persisted config.

In terms of long-term plans, we also have high conviction that this is a superior API but we must remain backwards compatible until our next major version bump, which has not been scheduled. We'll be converting our examples and documentation to "lead" with the new resources API starting with our next release on April 19th.

schrockn (changes requested):
I think this a good change. Please update the docblock for `@job`

simonvanderveldt:
@schrockn Thanks for the clarifications, that definitely helps to clear things up.

> Yes the double validation will be the behavior and in fact I would argue that it is desirable. Generally the Config instance will be instantiated when booting up the code location. Then the config will be instantiated when run and step processes are booted up as well. This is the behavior we want as lots of things can happen in between config persistence and when it is reloaded for execution. Code can be updated. Someone could manually or accidentally modify persisted config.

What would be the suggested approach to provide configuration to programmatically (from config files) created graphs that consist of reusable Ops where some of the config for the Ops is being derived from the incoming config?

For example an Op has a `MyOpConfig` with field "a" and it derives field "b" from that. This "b" field is and should not not settable directly. This currently passes the validations during the instantiation of `MyOpConfig` but fails when passing the dict version of this `Config` instance to the Op because it now has a field "b" which isn't allowed by the `MyOpConfig`.

P.S. This might be more something for a discussion I guess :P


> In terms of long-term plans, we also have high conviction that this is a superior API but we must remain backwards compatible until our next major version bump, which has not been scheduled. We'll be converting our examples and documentation to "lead" with the new resources API starting with our next release on April 19th.

:+1:

simonvanderveldt:
@schrockn here's a simplified example, hope this helps clarify it :)
```
from dagster import AssetsDefinition, Config, ConfigMapping, GraphDefinition, OutputMapping, op, repository
from pydantic import Extra, root_validator


class MyOpConfig(Config, extra=Extra.forbid):
    a: str
    b: str

    @root_validator(pre=True)
    def validate_and_set(cls, values):
        assert "b" not in values, "b can not be set in config"
        values["b"] = f"{values['a']}-baz"
        return values

@op
def my_op(context, config: MyOpConfig):
    return f"{config.a} {config.b}"

# This will pass validation
my_op_config = MyOpConfig(a="foo")
# a and b are both defined now
print(my_op_config.dict())

# But materializing this asset will fail validation
# because my_op_config.b is now set and the config gets validated again
graph = GraphDefinition(
    name="foo_graph",
    node_defs=[my_op],
    config=ConfigMapping(config_fn=lambda _: {"my_op": {"config": my_op_config.dict()}}, config_schema=None),
    output_mappings=[
        OutputMapping(
            graph_output_name="result",
            mapped_node_name="my_op",
            mapped_node_output_name="result"
        )
    ],
)

asset = AssetsDefinition.from_graph(graph)

@repository
def repo():
    return [asset]
```

schrockn:
Got it

@simonvanderveldt in this case I think the right answer will be to use `RunConfig` in the `ConfigMapping` case rather than the python dic, in which case we could avoid the double validation. The PR to add that is here: https://github.com/dagster-io/dagster/pull/13276

simonvanderveldt:
> @simonvanderveldt in this case I think the right answer will be to use `RunConfig` in the `ConfigMapping` case rather than the python dic, in which case we could avoid the double validation. The PR to add that is here: #13276

@schrockn OK, I missed that #13276 would make this possible, will give it a try and report back in that PR.

python_modules/dagster/dagster/_core/definitions/decorators/job_decorator.py
```diff
@@ -140,7 +143,9 @@ def job(
     name: Optional[str] = None,
     description: Optional[str] = None,
     resource_defs: Optional[Mapping[str, ResourceDefinition]] = None,
-    config: Optional[Union[ConfigMapping, Mapping[str, Any], "PartitionedConfig[object]"]] = None,
+    config: Optional[
+        Union[ConfigMapping, Mapping[str, Any], "RunConfig", "PartitionedConfig[object]"]
+    ] = None,
```
schrockn, line 148: Minor point here, but I think we should order the types in this Union by probability of desired usage. Meaning I think we want people using RunConfig the most frequently, then the dict, then PartitionedConfig, then ConfigMapping. 

So:
```
config: Optional[
        Union["RunConfig", Mapping[str, Any], "PartitionedConfig[object]", ConfigMapping]
    ] = None,

```

schrockn (approved):
Please heed final comment. Also cc: @sryza on the point of the ordering of types within a Union.

schrockn (approved):

PR #13278, opened 2023-03-31
[docs][pythonic config] Update jobs docs

benpankow:
## Summary

Brief update to the Jobs docs page, including new `ConfigMapping` pattern, resource examples, and config example.

### DIFF
--- a/docs/content/concepts/ops-jobs-graphs/jobs.mdx
+++ b/docs/content/concepts/ops-jobs-graphs/jobs.mdx
@@ -112,12 +112,17 @@ You can model this by building multiple jobs that use the same underlying graph 
 To do this, you first define a graph with the <PyObject object="graph" decorator /> decorator.
 
 ```python file=/concepts/ops_jobs_graphs/jobs_from_graphs.py startafter=start_define_graph endbefore=end_define_graph
-from dagster import graph, op
+from dagster import graph, op, Resource
 
 
-@op(required_resource_keys={"server"})
-def interact_with_server(context):
-    context.resources.server.ping_server()
+class Server:
+    def ping_server(self):
+        ...
+
+
+@op
+def interact_with_server(server: Resource[Server]):
+    server.ping_server()
 
 
 @graph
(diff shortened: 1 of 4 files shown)


benpankow:
* master
  * **PR #13276** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13276" target="_blank">Graphite</a> 
    * **PR #13277** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13277" target="_blank">Graphite</a> 
      * **PR #13278** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13278" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13278?utm_source=stack-comment).

docs/content/concepts/ops-jobs-graphs/jobs.mdx
```diff
@@ -155,18 +160,24 @@ When constructing a job, you can customize how that configuration will be satisf
 
 ### Hardcoded configuration
 
-You can supply a config dictionary. The supplied dictionary will be used to configure the job whenever the job is launched. It will show up in the Dagit Launchpad and can be overridden.
+You can supply a config dictionary or <PyObject object="RunConfig"/> object. The supplied config will be used to configure the job whenever the job is launched. It will show up in the Dagit Launchpad and can be overridden.
```
schrockn, line 163: This begs the question of how hard we want to push/promote the use of `RunConfig` over the config dictionary.

e.g. we could say

"You can supply a `RunConfig` (or a legacy config dictionary)."

docs/content/concepts/ops-jobs-graphs/jobs.mdx
```diff
@@ -190,19 +201,27 @@ Refer to the [Partitions documentation](/concepts/partitions-schedules-sensors/p
...
-    }
+class SimplifiedConfig(Config):
+    simplified_param: str
+
+
+@config_mapping
+def simplified_config(val: SimplifiedConfig) -> RunConfig:
+    return RunConfig(
+        ops={"do_something": DoSomethingConfig(config_param=val.simplified_param)}
+    )
```
schrockn, line 224: Do we want to keep the old examples around as well? cc: @erinkcochran87

schrockn (changes requested):

docs/content/concepts/ops-jobs-graphs/jobs.mdx
erinkcochran87: A downside to keeping the old examples around is that we need to continue to maintain the code and tests. If we think there's value in that, we could put these examples into tabs so folks can toggle between the new and the old, ex:

```
| Pydantic config | Legacy config |
```

We could remove the tabs and old examples in a future version, so we're not totally taking things away or making them difficult to find.

Alternatively, we didn't keep many (if any?) examples of `@repository` around when we did the `Definitions` migration, except for the dedicated Repository page. If we want to be consistent in our doc updates - and for this to be The Way users do this thing - it may be best to remove them.

Thoughts @schrockn? I realize this isn't a direct answer to your question, but there's some product strategy to consider here outside of just doc updates.

schrockn (approved):

docs/content/concepts/ops-jobs-graphs/jobs.mdx
erinkcochran87: Understood! Then lets go with having new examples only.

FWIW, I think it'd be good to validate a theory like that at some point. We can probably do some work in Google Analytics at some point for custom events, or even implement event capture like Snowplow. Definitely not a near-term need, but it'd be great to have more interaction data like this to validate design choices.

PR #13282, opened 2023-03-31
[Checkbox] stopPropagation for `onClick`

salazarm:
## Summary & Motivation

The Checkbox component wraps the checkbox input and its label in a `<label>` element. This means that when the checkbox is clicked, any parent `onClick` handlers will fire twice (see this codesandbox for a demonstration: https://codesandbox.io/s/muddy-https-6zypxg?file=/src/index.js) .

To fix this I move the `onClick` handler onto the `label` element, and I call `stopPropagation` when clicking on the checkbox input directly. see for demonstration of the fix: https://codesandbox.io/s/nice-star-ppf0hk?file=/src/index.js

### DIFF
--- a/js_modules/dagit/packages/core/src/ui/Filters/useStaticSetFilter.tsx
+++ b/js_modules/dagit/packages/core/src/ui/Filters/useStaticSetFilter.tsx
@@ -217,13 +217,7 @@ function SetFilterLabel(props: SetFilterLabelProps) {
       margin={{left: 4}}
       style={{maxWidth: '500px'}}
     >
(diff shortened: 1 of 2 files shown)


hellendag (approved):
Discussed in a huddle. It seems fair to me to limit our `Checkbox` component to producing a single `onClick` event in this manner, since it seems like it could be a point of confusion for anyone not expecting the `label-input` behavior to produce two click events.

It's worth double-checking that this doesn't adversely impact any vanilla `onChange` usage.

PR #13284, opened 2023-03-31
Allow a top-level "run_queue" config field in dagster.yaml instead of requiring users to grok RunCoordinators

gibsondan:
This makes configuring a run queue in OSS and cloud dramatically more similar and lets you set up queuing functionality without needing to understand a whole separate "run coordinator" concept.

@erinkcochran87 is updating docs separately as part of the new concurrency docs.

### DIFF
--- a/python_modules/dagster/dagster/_core/instance/ref.py
+++ b/python_modules/dagster/dagster/_core/instance/ref.py
@@ -418,11 +418,20 @@ class InstanceRef(
             config_value, "scheduler", defaults["scheduler"]
         )
 
-        run_coordinator_data = configurable_class_data_or_default(
-            config_value,
-            "run_coordinator",
-            defaults["run_coordinator"],
-        )
+        if config_value.get("run_queue"):
+            run_coordinator_data = configurable_class_data(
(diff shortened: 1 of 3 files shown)


johannkm (approved):
excellent

python_modules/dagster/dagster/_core/instance/ref.py
```diff
@@ -418,11 +418,20 @@ def from_dir(
             config_value, "scheduler", defaults["scheduler"]
         )
 
-        run_coordinator_data = configurable_class_data_or_default(
-            config_value,
-            "run_coordinator",
-            defaults["run_coordinator"],
-        )
+        if config_value.get("run_queue"):
```
johannkm, line 421: could there be some confusing case since

```
run_queue: {}
```

doesn't enable the daemon but 

```
run_queue:
  tag_concurrency_limits: []
```

does? I think that seems fine

gibsondan: I think its probably fine to have empty mean disabled, yeah

PR #13289, opened 2023-03-31
[docs] - Improve concurrency documentation

erinkcochran87:
## Summary & Motivation

This PR:

- Adds a dedicated guide for configuring concurrency limits
- Adds a dedicated guide for customizing run queue priority
- Updates the run coordinator documentation to be more high-level and act as a jumping-off point for more detailed content

TODO:

- [x] Update test for unique tag limits (`/docs_snippets_tests/deploying_tests/test_concurrency_limits.py`)
- [x] Get new examples into `/docs_snippets` and under tests
- [x] Add new guides to sidenav
- [x] Add new guides to `/guides` category page
- [x] Add GH discussion info into concurrency guide: https://github.com/dagster-io/dagster/discussions/8414, https://github.com/dagster-io/dagster/discussions/10075
- [x] Update Job execution concept page
- [x] **TBD**: Update `op concurrency` term


docs/content/deployment/run-coordinator.mdx
```diff
@@ -1,164 +1,140 @@
...
+        headers to inject run tags for{" "}
+        <a href="/guides/dagster/run-attribution">run attribution</a>.
+      </td>
+    </tr>
+  </tbody>
+</table>
+
+---
+
+## Configuring run coordinators
```
erinkcochran87, line 105: This section is the only content that has really changed in this guide. Everything else is cosmetic.

I'm not sure it's 100% accurate, so eyeballs here would be great.

docs/content/guides.mdx
```diff
@@ -10,8 +10,12 @@ Learn to apply [Dagster concepts](/concepts) to your work, explore experimental
 
 ## General
```
erinkcochran87, line 11: This page only contains cosmetic changes. No content review needed.

docs/content/guides/customizing-run-queue-priority.mdx
```diff
@@ -0,0 +1,138 @@
+---
```
erinkcochran87, line 1: This guide is sourced entirely from the current Run coordinator concept page: https://docs.dagster.io/deployment/run-coordinator

The motivation for breaking this out is to make the content more findable, as it was pretty buried. The new RC page links to this new page.

docs/content/guides/experimental-features.mdx
```diff
@@ -1,11 +1,29 @@
 ---
-title: Experimental Feature Guides | Dagster Docs
+title: Advanced and experimental feature guides | Dagster Docs
```
erinkcochran87, line 2: This page only contains cosmetic changes. No content review needed.

docs/content/guides/general.mdx
```diff
@@ -4,8 +4,16 @@ title: General Dagster Guides | Dagster Docs
 
 # General guides
```
erinkcochran87, line 5: This page only contains cosmetic changes. No content review needed.

docs/content/guides/limiting-concurrency-in-data-pipelines.mdx
```diff
@@ -0,0 +1,471 @@
...
+```
+
+With this configuration, each unique value of the `use-case` tag will be limited to three concurrent runs. For example, `team: marketing` and `team: sales` will both be limited to three concurrent runs each.
+
+</TabItem>
+</TabGroup>
+
+---
+
+## Configuring op/asset-level concurrency
```
erinkcochran87, line 199: @gibsondan is it fair to say that this type of concurrency is configured (at least somewhat under the hood) using a job executor?

erinkcochran87:
@gibsondan Over to you for another review! These are the pages that require content review - everything else is cosmetic:

- `/deploying/run-coordinator` - I took the concurrency and run queue priority content out and made them their own guides. I left a comment on the section in here I'd like your eyeballs on.
- `/guides/limiting-concurrency-in-data-pipelines` - The new concurrency guide
- `/guides/customizing-run-queue-priority` - This is new, but the content is sourced 100% from the current Run coordinator page. All I did was apply our style standards and update some of the copy for clarity.

I could also use your help getting the tests for the `docs_snippets` examples to pass - would you mind taking a look?

docs/content/guides/limiting-concurrency-in-data-pipelines.mdx
```diff
@@ -0,0 +1,471 @@
...
+## Configuring run-level concurrency
+
+Run concurrency can be configured by:
+
+- Limiting the overall number of runs in a deployment
+- Specifying limits using tags, which is useful for applying limits to specific jobs. For example, jobs that share a resource could all have the same tag.
+
+These methods of limiting concurrency can be used individually or together. For each queued run, Dagster will check that launching the run will satisfy all conditions. For example, if launching a run would exceed the maximum number of concurrent runs, Dagster will queue the run until another run finishes.
+
+**Note**: Setting maximum concurrent runs for the same job isn't currently supported. If you want to prevent a new run from launching if a current run is in progress, we recommend writing some custom logic in a job schedule. Check out [this GitHub discussion for an example](https://github.com/dagster-io/dagster/discussions/8414).
```
gibsondan, line 127: I don't think "Setting maximum concurrent runs for the same job isn't currently supported." is quite right (you can use a tag to set maximum concurrent runs for a job) - not sure about the exact language here, but what I believe that github discussion is giving an example of is more like "Run-level concurrency rules can't currently be used to prevent a run from being submitted to the queue - they only control limits for the number of runs on the queue that will be launched at once. If you want to prevent runs from ever being launched if another job is already being, running, check out this Github discussion"

docs/content/guides/limiting-concurrency-in-data-pipelines.mdx
```diff
@@ -0,0 +1,471 @@
...
+
+<TabGroup>
+<TabItem name="Limiting overall runs">
+
+### Limiting overall runs
+
+How you limit the overall number of runs in a deployment depends on whether you're using Dagster Cloud or Dagster Open Source:
```
gibsondan, line 134: nitpick but maybe "Limiting the overall number of *in-progress* runs or *concurrent* runs?

docs/content/guides/limiting-concurrency-in-data-pipelines.mdx
```diff
@@ -0,0 +1,471 @@
...
+
+```yaml file=/deploying/concurrency_limits/dagster.yaml startafter=start_overall_example endbefore=end_overall_example
+run_queue:
+  max_concurrent_runs: 15
+```
+
+When defining a value for `max_concurrent_runs`, keep the following in mind:
+
+- This setting defaults to `10`
+- Disable the setting with a `-1` value. **Note:** All other negative numbers are invalid.
```
gibsondan, line 149: disabling the setting means there will be no limit (and isn't an option in cloud - there has to be some limit there)

docs/content/guides/limiting-concurrency-in-data-pipelines.mdx
```diff
@@ -0,0 +1,471 @@
...
+
+</TabItem>
+<TabItem name="Limiting specific runs using tags">
+
+### Limiting specific runs using tags
+
+How you limit specific runs based on tags depends on whether you're using Dagster Cloud or Dagster Open Source:
+
+- **Dagster Cloud**: Use the [Dagster Cloud UI or the dagster-cloud CLI][cloud-deployment-settings]
+- **Dagster Open Source**: Use your instance's `dagster.yaml`
```
gibsondan, line 160: not strongly held but we could also put this part once at the top rather than repeating, then go into the different settings

docs/content/guides/limiting-concurrency-in-data-pipelines.mdx
gibsondan: Extremely fair and entirely accurate - I think we should include that and link to the executor docs

docs/content/guides/limiting-concurrency-in-data-pipelines.mdx
```diff
@@ -0,0 +1,471 @@
...
+Utilizing op/asset-level concurrency provides fine-grained control for the maximum number of ops/assets that can be executed at once within a single run, ensuring that shared resources aren't overwhelmed.
+
+Op and asset concurrency can be configured by:
+
+- Limiting the number of ops/assets that can execute in a single run at once
+- Specifying limits using tags
+
+### Limiting overall concurrency in a job
+
+To limit concurrency for ops and assets in jobs, use `max_concurrent` in the job’s config, either in Python or using the Launchpad in the Dagster UI.
```
gibsondan, line 218: this is where I think we should reference the executor (that's what's being configured in the "execution" key) - and will help explain why the key differs depending on what executor you're using

docs/content/guides/limiting-concurrency-in-data-pipelines.mdx
```diff
@@ -0,0 +1,471 @@
...
+
+<TabGroup>
+<TabItem name="In Python">
+
+Whiel `max_concurrent` is used to limit concurrency in both asset and op-based jobs, how jobs are defined differs between assets and ops. Click the tabs below for examples.
```
gibsondan, line 223: Whiel => While

docs/content/guides/limiting-concurrency-in-data-pipelines.mdx
gibsondan: I dunno if it's worth noting in an example that you can also apply this configuration directly on the executor itself via configured? That would let you supply a default across multiple jobs. Maybe we are de-emphasizing configured though since it's complicated? but it would look like this:

```
my_custom_executor = multiprocess_executor.configured({"max_concurrent": 2})
@job(executor=my_custom_executor)
...
```

Or even as a default in the Definitions object (which we don't currently show in the docs anywhere that you can do i don't think)

```
Definitions(
  ...
  default_executor_def=my_custom_executor
)
```

docs/content/guides/limiting-concurrency-in-data-pipelines.mdx
```diff
@@ -0,0 +1,471 @@
...
+    }
+)
+```
+
+</TabItem>
+<TabItem name="Op-based jobs">
+
+#### Op-based jobs
+
+Op-based jobs are defined using the <PyObject object="job" decorator /> decorator. In this example, using the <PyObject module="dagster" object="multiprocess_executor" />, the job will execute up to four ops at once:
```
gibsondan, line 252: or via `graph.to_job()` but that's probably not needed here

gibsondan, line 252: it also has a tag concurrency_limit should we mention that here?

docs/content/guides/limiting-concurrency-in-data-pipelines.mdx
```diff
@@ -0,0 +1,471 @@
...
+</TabItem>
+</TabGroup>
+
+---
+
+## Troubleshooting
+
+When limiting concurrency, you might run into some issues until you get the configuration right.
+
+### Runs going to STARTED status and skipping QUEUED
```
gibsondan, line 391: this will never happen in cloud, there's always a run queue

docs/content/guides/limiting-concurrency-in-data-pipelines.mdx
```diff
@@ -0,0 +1,471 @@
...
+
+The possible causes for runs remaining in `QUEUED` status depend on whether you're using Dagster Cloud or Dagster Open Source.
+
+<TabGroup>
+<TabItem name="Dagster Cloud">
+
+If runs aren't being dequeued in Dagster Cloud, the root causes could be:
+
+- **If using a [Hybrid deployment](/dagster-cloud/deployment/hybrid)**, the agent serving the deployment may be down. In this situation, runs will be paused.
+- **Dagster Cloud is experiencing downtime**. Check the [status page](http://dagstercloud.statuspage.io/) for the latest on potential outages.
```
gibsondan, line 408: this might go without saying (any problem could be caused by a bug or downtime) - not sure if we need to specifically mention it here, but not strongly held

docs/content/guides/limiting-concurrency-in-data-pipelines.mdx
```diff
@@ -0,0 +1,471 @@
...
+
+If runs aren’t being dequeued in Dagster Open Source, the root cause is likely an issue with the Dagster daemon or the run queue configuration.
+
+- **Troubleshoot the Dagster daemon**:
+
+  1. **Verify the Dagster daemon is set up and running**. In the Dagster UI, navigate to **Deployment > Daemons** and verify that the daemon is running. The **Run queue** should also be running. If you used [`dagster dev`](/guides/running-dagster-locally) to start the Dagster UI, the daemon should have been started for you.
+
+     If the daemon isn’t running, proceed to step 2.
+
+  2. **Verify the Dagster daemon can access the same storage as the Dagit process**. Both the Dagit (Dagster UI) process and Dagster daemon should access the same storage, meaning they should use the same `dagster.yaml`. Locally, this means both processes should have the same set `DAGSTER_HOME` environment variable. Refer to the [Dagster Instance docs](/deployment/dagster-instance) for more information.
```
gibsondan, line 421: If you're using `dagster dev` they will always be using the same storage (one of the reasons we built it)

gibsondan (approved):
the only comment here that i would describe as blocking is the part about "Setting maximum concurrent runs for the same job isn't currently supported." - everything else is just a suggestion or optional. Happy to talk through that part offline but I don't think this requires another full pass once that's resolved. Thanks for putting this together, I think it will be a huge help

docs/content/deployment/run-coordinator.mdx
```diff
@@ -1,164 +1,140 @@
...
+          object="DefaultRunCoordinator"
+        />{" "}
+        calls <code>launch_run</code> on the instance’s run launcher immediately
+        in the same process, without applying any limits or prioritization
+        rules.
+        <br />
+        <br />
+        When this coordinator is set, clicking <strong>Launch Run</strong> in the
+        Dagster UI will immediately launch the run from the Dagit process. Similarly,
+        scheduled runs will immediately launch from the scheduler process.
```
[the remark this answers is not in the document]
erinkcochran87: will fix!

docs/content/deployment/run-coordinator.mdx
```diff
@@ -1,164 +1,140 @@
...
+      <td>
+        <strong>Custom run coordinator</strong>
+      </td>
+      <td>
+        Using a subclass of one of the <code>RunCoordinator</code>s, you can
+        define a custom queueing strategy when submitting runs to the{" "}
+        <a href="/deployment/dagster-daemon">Dagster daemon</a> or executing
+        custom hooks before runs are submitted. For example, you can read HTTP
+        headers to inject run tags for{" "}
+        <a href="/guides/dagster/run-attribution">run attribution</a>.
```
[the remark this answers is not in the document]
erinkcochran87: It is indeed old text. Happy to update it though

docs/content/guides/limiting-concurrency-in-data-pipelines.mdx
erinkcochran87: I can call that out here, that it only applies to OSS

docs/content/guides/customizing-run-queue-priority.mdx
```diff
@@ -0,0 +1,138 @@
...
+title: "Customizing run queue priority"
+description: "Define custom prioritization rules for your Dagster instance's run queue."
+---
+
+# Customizing run queue priority
+
+Using the `QueuedRunCoordinator` or a custom run coordinator, you can define custom prioritization rules for your Dagster instance’s run queue.
```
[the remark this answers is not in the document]
erinkcochran87: I agree!

docs/content/guides/limiting-concurrency-in-data-pipelines.mdx
erinkcochran87: ty!

docs/content/guides/limiting-concurrency-in-data-pipelines.mdx
erinkcochran87: I think this is good info, but I want to get these out sooner rather than later. I'm going to do a follow-up PR that adds this in.

PR #13310, opened 2023-04-01
fixes #13309

### LINKED ISSUE #13309: [1.2.4] Dynamic Partitions broken with remote grpc
### Dagster version

1.2.4

### What's the issue?

with the removal of PartitionSetDefinition #13145 , a side effect is that we are unable to launch a dynamically partitionned asset materialization (at least from dagit).

the job should launch normally

### How to reproduce?

create a simple asset with a dynamic partition a `main.py`script for example:

```python
from dagster import asset, Definitions, DynamicPartitionsDefinition

dynamic_partitions = DynamicPartitionsDefinition(name="testpartition")


@asset(partitions_def=dynamic_partitions)
def myasset(context):
    return 1


defs = Definitions(
    assets=[myasset],
)

```

launch a grpc server : `dagster api grpc -f main.py -h localhost -p 3030`

create a remote code location with a `workspace.yml`file : 

```yaml
load_from:
  - grpc_server:
      host: localhost
      port: 3030
      location_name: brokenpartitions
```

launch dagit : `dagit -w workspace.yml`

try to materialize the asset in dagit (create a partition if needed)

get the error message : 

image

### Deployment type

Dagster Helm chart

### Deployment details

should fail as long as the code location is a grpc server

### Additional information

_No response_

planvin:
## Summary & Motivation
An attempt to fix #13309
## How I Tested These Changes

patched impl.py locally

Launched a grpc instance as described in issue #13309
Launched a dagit instance as described in issue #13309

Materialization launched normally


gibsondan:
Thanks so much for sending this out! This looks right to me, we’ll take a closer look on Monday.

python_modules/dagster/dagster/_grpc/impl.py
```diff
@@ -156,6 +156,18 @@ def core_execute_run(
         raise
 
 
+def _instance_from_ref(instance_ref: InstanceRef, partitions_def: PartitionsDefinition):
+    # Certain gRPC servers do not have access to the instance, so we only attempt to instantiate
```
smackesey, line 160: - A helper method here is reasonable, but let's call it `_instance_from_ref_for_dynamic_partitions` to communicate that it's not always yielding an instance.
- This needs a `@contextmanager` decorator and the result yielded instead of returned to work as intended
- Annotate return vale as `Iterator[Optional[DagsterInstance]]`
- This should never be called with a `partitions_def=None`, so no need to check

python_modules/dagster/dagster/_grpc/impl.py
```diff
@@ -525,14 +531,7 @@ def get_partition_set_execution_param_data(
...
     try:
-        # Certain gRPC servers do not have access to the instance, so we only attempt to instantiate
-        # the instance when necessary for dynamic partitions: https://github.com/dagster-io/dagster/issues/12440
-        instance = (
-            DagsterInstance.from_ref(instance_ref)
-            if (_partitions_def_contains_dynamic_partitions_def(partitions_def) and instance_ref)
-            else None
-        )
-        with user_code_error_boundary(
+        with _instance_from_ref(instance_ref, partitions_def) as instance, user_code_error_boundary(
```
smackesey, line 534: Let's put the `user_code_error_boundary` nested on a separate line

smackesey (changes requested):

python_modules/dagster/dagster/_grpc/impl.py
```diff
@@ -401,24 +413,19 @@ def get_partition_config(
...
-            with DagsterInstance.from_ref(
-                instance_ref
-            ) if instance_ref else nullcontext() as instance:
-                partition = partitions_def.get_partition(
-                    partition_key, dynamic_partitions_store=instance
-                )
-        else:
-            partition = partitions_def.get_partition(partition_key, dynamic_partitions_store=None)
+        with _instance_from_ref(instance_ref, partitions_def) as instance:
+            partition = partitions_def.get_partition(
```
clairelin135, line 417: I think the `get_partition` call is unnecessary, cc @smackesey to confirm

I think we fetch the partition object within `get_run_config_for_partition_key` to validate that it is a valid partition

python_modules/dagster/dagster/_grpc/impl.py
```diff
@@ -462,20 +469,19 @@ def get_partition_tags(
...
-        partition = partitions_def.get_partition(partition_name, dynamic_partitions_store=instance)
-        with user_code_error_boundary(
-            PartitionExecutionError,
-            lambda: f"Error occurred during the evaluation of the `tags_for_partition` function for partitioned config on job '{job_def.name}'",
-        ):
-            tags = partitioned_config.get_tags_for_partition_key(
-                partition.name, job_name=job_def.name
+
+        with _instance_from_ref(instance_ref, partitions_def) as instance:
+            partition = partitions_def.get_partition(
```
clairelin135, line 474: same here

python_modules/dagster/dagster/_grpc/impl.py
smackesey: I think you're right, but it's not returned, while we need the partition name out here for return in the `ExternalPartitionConfigData`.

clairelin135: I think the partition key is always partition.name

smackesey: You're probably right but I think you should just remove it in your followup PR that already exists to avoid adding complexity and potential back-and-forth with @planvin, since these `get_partition` calls pre-existed his PR anyway.

python_modules/dagster/dagster/_grpc/impl.py
planvin: thanks, applied these changes in last commit

smackesey:
@planvin thx, running our CI against this now

python_modules/dagster/dagster/_grpc/impl.py
clairelin135: 👍

python_modules/dagster/dagster/_grpc/impl.py
```diff
@@ -401,24 +415,19 @@ def get_partition_config(
...
-            lambda: f"Error occurred during the evaluation of the `run_config_for_partition` function for partition set {partition_set_name}",
-        ):
-            run_config = partitioned_config.get_run_config_for_partition_key(partition_key)
-            return ExternalPartitionConfigData(name=partition.name, run_config=run_config)
+            with user_code_error_boundary(
+                PartitionExecutionError,
+                lambda: f"Error occurred during the evaluation of the `run_config_for_partition` function for partition set {partition_set_name}",
+            ):
+                run_config = partitioned_config.get_run_config_for_partition_key(
+                    partition_key, instance=instance
```
smackesey, line 428: `instance=instance` -> `dynamic_partitions_store=instance`

smackesey:
@planvin CI shows some failures:

- instance=instance -> dynamic_partitions_store=instance (see my inline comment)
- You need to change the annotation of `instance_ref` in your helper to `Optional[InstanceRef]`, and yield null context if `instance_ref` is None
- Run `make ruff` and make sure it passes
- Rebase on master before repush

Hopefully those will get this to green

planvin:
> @planvin CI shows some failures:
> 
>     * instance=instance -> dynamic_partitions_store=instance (see my inline comment)
fixed
> 
>     * You need to change the annotation of `instance_ref` in your helper to `Optional[InstanceRef]`, and yield null context if `instance_ref` is None
fixed the typing. not sure about the yielding of nullcontext, if instance_ref is none we will yield None so I think we are ok here right ?
> 
>     * Run `make ruff` and make sure it passes
> 
ok
>     * Rebase on master before repush
> 
ok
> 
> Hopefully those will get this to green

smackesey:
Sweet, thx @planvin, merging this now!

PR #12697, opened 2023-03-03
Support requiring resources in schedules, experimentally

benpankow:
## Summary

Implements very basic resource support for schedules. Effectively just calls `build_resouces` and passes the results to user code.

Resources can be specified through `required_resource_keys` or through annoated params on the schedule fn. They can be accessed through context or through these params:

```python

class MyResource(ConfigurableResource):
    a_str: str
    
@schedule(cron_schedule="* * * * *", job_name="the_job", required_resource_keys={"my_resource"})
def schedule_from_context(context: ScheduleEvaluationContext):
    return RunRequest(context.resources.my_resource.a_str, run_config={}, tags={})


@schedule(cron_schedule="* * * * *", job_name="the_job")
def schedule_from_fn_arg(context: ScheduleEvaluationContext, my_resource: MyResource):
    return RunRequest(my_resource.a_str, run_config={}, tags={})

```

Directly invoking schedule requires that a user only provide the context:

```python
# Here, `resource_defs` is exploded into the various resource args which `schedule_from_fn_arg` expects 
schedule_from_fn_arg(
    build_schedule_context(
        resource_defs={"my_resource": MyResource(a_str="foo")}
    )
)
```

## Test Plan

unit test

PR #13323, opened 2023-04-03
[pythonic resources] Better ergonomics for direct invocation of sensors/schedules

benpankow:
## Summary

Implements similar functionality to #13002 to make passing resources to sensors and schedules easier in the direct invocation case. This makes testing in particular easier.

The new API changes enable the following (schedule behavior is identical):

```python
    a_str: str

@sensor(job_name="foo_pipeline")
def basic_sensor_resource_req(my_resource: MyResource):
    return RunRequest(run_key=None, run_config={"foo": my_resource.a_str}, tags={})


# Can pass resource through context
assert cast(
    RunRequest,
    basic_sensor_resource_req(
        context=build_sensor_context(resources={"my_resource": MyResource(a_str="foo")})
    ),
).run_config == {"foo": "foo"}

# Can pass resource directly
assert cast(
    RunRequest,
    basic_sensor_resource_req(my_resource=MyResource(a_str="foo")),
).run_config == {"foo": "foo"}


# Can pass resource directly with context
assert cast(
    RunRequest,
    basic_sensor_resource_req(build_sensor_context(), my_resource=MyResource(a_str="foo")),
).run_config == {"foo": "foo"}

```

## Test Plan

New unit tests; existing unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/schedule_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/schedule_definition.py
@@ -94,14 +94,22 @@ def get_or_create_schedule_context(
     Raises an exception if the user passes more than one argument or if the user-provided
     function requires a context parameter but none is passed.
     """
+    from dagster._config.structured_config import ResourceDefinition
     from dagster._core.definitions.sensor_definition import get_context_param_name
 
     context_param_name = get_context_param_name(fn)
 
-    if len(args) + len(kwargs) > 1:
+    kwarg_keys_non_resource = set(kwargs.keys()) - {param.name for param in get_resource_args(fn)}
+    if len(args) + len(kwarg_keys_non_resource) > 1:
         raise DagsterInvalidInvocationError(
-            "Schedule invocation received multiple arguments. Only a first "
-            "positional context parameter should be provided."
+            "Schedule invocation received multiple non-resource arguments. Only a first "
+            "positional context parameter should be provided when invoking."
+        )
+
+    if any(isinstance(arg, ResourceDefinition) for arg in args):
+        raise DagsterInvalidInvocationError(
+            "If directly invoking a schedule, you may not provide resources as"
+            " positional arguments, only as keyword arguments."
         )
 
     context: Optional[ScheduleEvaluationContext] = None
@@ -113,8 +121,11 @@ def get_or_create_schedule_context(
             raise DagsterInvalidInvocationError(
                 f"Schedule invocation expected argument '{context_param_name}'."
             )
-        context_param_name = context_param_name or list(kwargs.keys())[0]
-        context = check.opt_inst(kwargs.get(context_param_name), ScheduleEvaluationContext)
+        context = (
+            check.opt_inst(kwargs.get(context_param_name), ScheduleEvaluationContext)
+            if context_param_name
+            else check.opt_inst(kwargs.get("context"), ScheduleEvaluationContext)
+        )
     elif context_param_name:
         # If the context parameter is present but no value was provided, we error
         raise DagsterInvalidInvocationError(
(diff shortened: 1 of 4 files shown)


benpankow:
* master
  * **PR #12401** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12401" target="_blank">Graphite</a> 
    * **PR #12607** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12607" target="_blank">Graphite</a> 
      * **PR #13323** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13323" target="_blank">Graphite</a>  👈
    * **PR #12697** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/12697" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13323?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
```diff
@@ -113,16 +121,35 @@ def get_or_create_schedule_context(
             raise DagsterInvalidInvocationError(
                 f"Schedule invocation expected argument '{context_param_name}'."
             )
-        context_param_name = context_param_name or list(kwargs.keys())[0]
-        context = check.opt_inst(kwargs.get(context_param_name), ScheduleEvaluationContext)
+        context = (
```
schrockn, line 124: Nit:

`check.opt_inst(kwargs.get(context_param_name or "context"), ScheduleEvaluationContext)` 

A little more terse/readable iMO

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
```diff
@@ -113,16 +121,35 @@ def get_or_create_schedule_context(
...
 
-    return context or build_schedule_context()
+    context = context or build_schedule_context()
+    resource_args_from_kwargs = {}
+
+    resource_args = {param.name for param in get_resource_args(fn)}
+    for resource_arg in resource_args:
+        if resource_arg in kwargs:
+            resource_args_from_kwargs[resource_arg] = kwargs[resource_arg]
+            del kwargs[resource_arg]
```
schrockn, line 143: why are you deleting here if it is not used in the rest of the function?

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
```diff
@@ -113,16 +121,35 @@ def get_or_create_schedule_context(
...
+
+    resources_provided_in_multiple_places = (resource_args_from_kwargs) and (context.resource_defs)
+    if resources_provided_in_multiple_places:
```
schrockn, line 146: I'm not convinced this should be illegal. In fact in production contexts resources will always be in both the context and available for direct passing, so we are making illegal a perfectly legal state for the program to be in.

schrockn (changes requested):

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
benpankow: This is reasonable, changed to support passing both. I am a little confused by

> in production contexts resources will always be in both the context and available for direct passing

I'm trying to conceptualize an instance where a user would would directly invoke a schedule with resources in context & pass directly, in production or otherwise

schrockn: If the code called other code that relies on the resources being attached to the context.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -169,6 +169,29 @@ def __exit__(self, *exc) -> None:
     def resource_defs(self) -> Optional[Mapping[str, "ResourceDefinition"]]:
         return self._resource_defs
 
+    def merge_resources(self, resources_dict: Mapping[str, Any]) -> "SensorEvaluationContext":
```
schrockn, line 172: ah great. almost commented on previous PR that this should be the behavior (merge vs replace)

schrockn (approved):

PR #13324, opened 2023-04-03
Add dbt tests as bona fide assets

emirkmo:
## Summary & Motivation

Dbt tests are also "just" sql with a data asset, like models, but with the specific caveat that they should not produce any output.
This works well with the mental model of a Dagster asset. However, DBT tests are currently not supported as bona-fide assets in dagster_dbt.

In fact, they are hard coded in various functions to be explicitly excluded or handled separately. However, we feel that our assets fit the mental model of a software defined asset really well. We want to see them fail when they produce output, and have this be reflected in the dagit UI. we want to visualize upstream and downstream dependencies based on the tests as assets, set freshness policy, and have all of this be handled by the reconciliation sensor.

This PR is a proposal to investigate adding in dbt tests as proper assets. There are various implementations possible, but we currently prefer making it an optional tweak, as projects with thousands or more of tests may not want them as assets. 

However, as assets, tests can still act as batch jobs with define asset job, show up as metadata on model assets, etc. So We do not lose anything, only gain. 

## How I Tested These Changes

Tested using dagster test, but as this is a proposal, no new tests were written yet, as the behavior of DBT nodes as assets is already tested, and this PR just Optionally removes the hard-coded dbt test exclusions from the code. meaning that if you pass is `tests_as_assets=True`, they are treated as they should be, a proper DBT node.

We make a few concessions to be able to run `dbt run` for models, optionally `dbt build` for both models and tests, and `dbt test` for tests otherwise. We also opt to keep the code architecture as is, and extend the functionality of the existing get_dbt_op factory and _dbt_nodes_to_assets methods (There were other possibilities..). We maintain the batch run capability.  

**Note: I was not able to run most tests that rely on test docker posgres due to a peculiar `FileNotFoundError` with subprocess and docker-compose. Tests error instead of skipping, see low below.** 
<details>
  <summary>Pytest Error</summary>

```Python

E               FileNotFoundError: [Errno 2] No such file or directory: 'docker-compose'

args       = ['docker-compose',
 '-f',
 '/home/emirk/Software/dagster/python_modules/libraries/dagster-dbt/dagster_dbt_tests/docker-compose.yml',
 'stop',
 'test-postgres-db-dbt']
c2pread    = 12
c2pwrite   = 13
child_exception_type = <class 'OSError'>
child_exec_never_called = False
close_fds  = True
creationflags = 0
cwd        = None
env        = None
env_list   = None
err_filename = 'docker-compose'
err_msg    = 'No such file or directory'
errno_num  = 2
errpipe_data = bytearray(b'OSError:2:')
errpipe_read = 14
errpipe_write = 15
errread    = -1
errwrite   = -1
exception_name = bytearray(b'OSError')
executable = b'docker-compose'
executable_list = (b'/home/emirk/mambaforge/envs/dagsterOS310/bin/docker-compose',
 b'/home/emirk/Software/dagster/.micromamba/docker-compose',
 b'/home/emirk/mambaforge/condabin/docker-compose',
 b'/usr/local/sbin/docker-compose',
 b'/usr/local/bin/docker-compose',
 b'/usr/sbin/docker-compose',
 b'/usr/bin/docker-compose',
 b'/sbin/docker-compose',
 b'/bin/docker-compose',
 b'/usr/games/docker-compose',
 b'/usr/local/games/docker-compose',
 b'/snap/bin/docker-compose',
 b'/var/cfengine/bin/docker-compose',
 b'/home/emirk/bin/docker-compose',
 b'/home/emirk/bin.linux/docker-compose')
fds_to_keep = {15}
gid        = None
gids       = None
hex_errno  = bytearray(b'2')
low_fds_to_close = []
orig_executable = 'docker-compose'
p2cread    = -1
p2cwrite   = -1
part       = b''
pass_fds   = ()
pid        = 386422
preexec_fn = None
restore_signals = True
self       = <Popen: returncode: 255 args: ['docker-compose', '-f', '/home/emirk/Software...>
shell      = False
start_new_session = False
startupinfo = None
sts        = 65280
uid        = None
umask      = -1

../../mambaforge/envs/dagsterOS310/lib/python3.10/subprocess.py:1847: FileNotFoundError
================================================================================================================================================================================================== short test summary info ===================================================================================================================================================================================================
ERROR python_modules/libraries/dagster-dbt/dagster_dbt_tests/test_asset_defs.py::test_custom_resource_key_asset_load - FileNotFoundError: [Errno 2] No such file or directory: 'docker-compose'
```

</details>

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
@@ -343,6 +343,7 @@ def _batch_event_iterator(
     runtime_metadata_fn: Optional[
         Callable[[OpExecutionContext, Mapping[str, Any]], Mapping[str, RawMetadataValue]]
     ],
+    tests_as_assets: bool,
     kwargs: Dict[str, Any],
 ) -> Iterator[Union[AssetObservation, AssetMaterialization, Output]]:
     """Yields events for a dbt cli invocation. Waits until the entire command has completed before
@@ -381,6 +382,7 @@ def _batch_event_iterator(
                 manifest_json=manifest_json,
                 extra_metadata=extra_metadata,
                 generate_asset_outputs=True,
+                tests_as_assets=tests_as_assets,
             )
 
 
@@ -463,11 +465,11 @@ def _stream_event_iterator(
     context: OpExecutionContext,
     dbt_resource: DbtCliResource,
     command: _DbtOpCommands,
-    tests_as_assets: bool,
     node_info_to_asset_key: Callable[[Mapping[str, Any]], AssetKey],
     runtime_metadata_fn: Optional[
         Callable[[OpExecutionContext, Mapping[str, Any]], Mapping[str, RawMetadataValue]]
     ],
+    tests_as_assets: bool,
     kwargs: Dict[str, Any],
 ) -> Iterator[Union[AssetObservation, Output]]:
     """Yields events for a dbt cli invocation. Emits outputs as soon as the relevant dbt logs are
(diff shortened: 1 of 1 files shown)


sryza:
Thanks for submitting this @emirkmo . The person who is best suited to review this – @OwenKephart – is currently on vacation. He's going to be back on April 17 and will be able to take a look then.

Another way that we had thought about modeling this is with new asset expectations functionality: https://github.com/dagster-io/dagster/discussions/9543.

emirkmo:
Closed since asset checks are so much more elegant.

PR #13341, opened 2023-04-03
Asset backfills page

salazarm:
## Summary & Motivation

Screen Shot 2023-04-03 at 4 34 18 PM

A couple of things are still missing.
1. The creator of the backfill
2. The partition selection that was made in the dialog.

## How I Tested These Changes

Storybook (check github action)


js_modules/dagit/packages/core/src/instance/backfill/BackfillPage.tsx
```diff
@@ -0,0 +1,310 @@
...
+    function getRunsUrl(status: 'requested' | 'complete' | 'failed' | 'targeted') {
+      const filters: RunFilterToken[] = [
+        {
+          token: 'tag',
+          value: `dagster/backfill=${backfillId}`,
+        },
+      ];
+      switch (status) {
+        case 'requested':
+          filters.push(
```
salazarm, line 62: I'm not sure about how to map these statuses tbh @bengotow.

python_modules/dagster-graphql/dagster_graphql/schema/backfill.py
```diff
@@ -213,6 +218,16 @@ def resolve_runs(self, graphene_info: ResolveInfo) -> "Sequence[GrapheneRun]":
         records = self._get_records(graphene_info)
         return [GrapheneRun(record) for record in records]
 
+    def resolve_endTimestamp(self, graphene_info: ResolveInfo) -> Optional[float]:
```
salazarm, line 221: @clairelin135  added this field so we wouldn't need to query runs on the client

clairelin135: 👍 👍

clairelin135:
https://github.com/dagster-io/dagster/pull/13350 adds a resolver for the initial partition selection

js_modules/dagit/packages/core/src/instance/backfill/BackfillPage.tsx
```diff
@@ -0,0 +1,310 @@
...
+  useDocumentTitle(`Backfill | ${backfillId}`);
+
+  const {data, loading} = useQuery<
+    BackfillStatusesByAssetQuery,
+    BackfillStatusesByAssetQueryVariables
+  >(BACKFILL_DETAILS_QUERY, {
+    variables: {backfillId},
+  });
+
+  console.log({backfillId, data, loading});
```
bengotow, line 37: Hmm want to rm this?

js_modules/dagit/packages/core/src/instance/backfill/BackfillPage.tsx
bengotow: I think this looks good!

js_modules/dagit/packages/core/src/instance/backfill/BackfillPage.tsx
```diff
@@ -0,0 +1,310 @@
...
+              <Duration
+                start={backfill.timestamp * 1000}
+                end={backfill.endTimestamp ? backfill.endTimestamp * 1000 : null}
+              />
+            }
+          />
+          {/* <Detail label="Partition Selection" detail="test" /> */}
+          <Detail label="Status" detail={<StatusLabel status={backfill.status} />} />
+        </Box>
+        <Table $monospaceFont={false}>
```
bengotow, line 121: This all looks great, but it seems asset specific and there isn't a fallback for non-asset backfills. I think that the updates to BackfillMessaging.tsx and BackfillRow make this page reachable for both asset backfills AND op job backfills, and the content here might not make sense for the latter. Maybe we could make this table conditional? Would be good to see what happens when you make the GraphQL query for `assetPartitionsStatusCounts` on a non-asset backfill

bengotow (approved):
This looks good to me -- my only concern is around the intersection of this new page with non-asset backfills 👍

js_modules/dagit/packages/core/src/instance/backfill/BackfillPage.tsx
salazarm: Good call out.

How do I create a non-asset backfill? Is there a toy I can use?

PR #13342, opened 2023-04-03
Remove spurious warning when a local grpc server process shuts down in the daemon

gibsondan:
Summary:
This warning is getting triggered every time the daemon shuts down a local grpc server process, but it's not relevant because the heartbeat will spin it down. Change the error to only fire when there's no heartbeat.

Test Plan:
Run dagster-daemon locally, no more spurious warning every couple of minutes

### DIFF
--- a/python_modules/dagster/dagster/_grpc/server.py
+++ b/python_modules/dagster/dagster/_grpc/server.py
@@ -1200,7 +1200,7 @@ class GrpcServerProcess:
         check.bool_param(force_port, "force_port")
         check.int_param(max_retries, "max_retries")
         check.opt_int_param(max_workers, "max_workers")
-        check.bool_param(heartbeat, "heartbeat")
+        self._heartbeat = check.bool_param(heartbeat, "heartbeat")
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster/_grpc/server.py
```diff
@@ -1285,12 +1285,12 @@ def shutdown_server(self):
...
     def __del__(self):
-        if not self._shutdown:
+        if not self._shutdown and not self._heartbeat:
             warnings.warn(
-                "GrpcServerProcess is being destroyed without signalling to server that "
-                "it should shut down. This may result in server processes living longer than "
-                "they need to. To fix this, wrap the GrpcServerProcess in a contextmanager or "
-                "call shutdown_server on it."
+                "GrpcServerProcess without a heartbeat is being destroyed without signalling to"
+                " server that it should shut down. This may result in server processes living"
```
prha, line 1291: "to the server"

prha (approved):

PR #13235, opened 2023-03-29
[snowflake-pyspark] SnowflakePySparkIOManager

jamiedemaria:
## Summary & Motivation
converts the snowflake pyspark io manager to the pythonic resource system

## How I Tested These Changes
unit tests now test both versions of the io manager

PR #13347, opened 2023-04-03
[dagster-snowflake] pydantic SnowflakeResource

jamiedemaria:
## Summary & Motivation
converts the snowflake resource to pydantic config


jamiedemaria:
* master
  * **PR #13232** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13232" target="_blank">Graphite</a> 
    * **PR #13233** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13233" target="_blank">Graphite</a> 
    * **PR #13347** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13347" target="_blank">Graphite</a>  👈
    * **PR #13235** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13235" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13347?utm_source=stack-comment).

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -335,11 +344,249 @@ def write_parquet_file(context):
...
+
+            my_snowflake_job.execute_in_process()
+    """
+
+    account: Optional[str] = Field(
+        default=None,
+        description="Your Snowflake account name. For more details, see  https://bit.ly/2FBL320.",
+    )
+
+    user: str = Field(..., description="User login name.")
```
jamiedemaria, line 384: remove ...

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -335,11 +344,249 @@ def write_parquet_file(context):
...
+    @validator("connector")
+    def validate_connector(cls, v):
+        if v is not None and v != "sqlalchemy":
+            raise DagsterInvalidConfigError(
+                "Snowflake Resource: 'connector' configuration value must be None or sqlalchemy"
+            )
+        return v
+
+    @overload
+    def __init__(self, schema: str, *args, **kwargs):
```
benpankow, line 571: Sadly you may have to explicitly list out the other arguments here for the autocomplete in-editor to look right. Are you able to check that e.g. you get hints for one of the other args?

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -335,11 +344,249 @@ def write_parquet_file(context):
         self.execute_queries(sql_queries)
 
 
+class SnowflakeResource(ConfigurableResourceFactory):
```
benpankow, line 347: `ConfigurableResourceFactory[SnowflakeConnection]` will help type hinting in some places

benpankow (commented):
This looks good to me - open question about the in-editor type hints for constructor parameters. Also would like to get @schrockn's opinion on a solution like this for `_schema`.

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
jamiedemaria: yeah you're right

jamiedemaria: that's kind of a pain to have to copy all of the config names to the init along with their types. seems like something that could very easily get out of sync

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -335,11 +344,249 @@ def write_parquet_file(context):
...
+                    password=EnvVar("SNOWFLAKE_PASSWORD")
+                    database="MY_DATABASE",
+                    schema="MY_SCHEMA",
+                    warehouse="MY_WAREHOUSE"
+                )}
+            )
+            def my_snowflake_job():
+                get_one()
+
+            my_snowflake_job.execute_in_process()
```
schrockn, line 376: We need to make a decision about how strongly we will push for the using resource_defs on  `Definitions` versus arguments on `job` versus passing to `execute_in_process`. We are kind of all over the place.

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -335,11 +344,249 @@ def write_parquet_file(context):
...
+    def __init__(self, schema: str, *args, **kwargs):
+        ...
+
+    @overload
+    def __init__(self, *args, **kwargs):
+        ...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
```
schrockn, line 580: This is super painful and I don't think we can recommend this. What are the other options here?

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -335,11 +344,249 @@ def write_parquet_file(context):
...
+    def create_resource(self, context) -> SnowflakeConnection:
+        return SnowflakeConnection(context.resource_config, context.log)
+
+
 @resource(
-    config_schema=define_snowflake_config(),
+    config_schema=infer_schema_from_config_class(SnowflakeResource),
```
schrockn, line 586: `SnowflakeResource.to_config_schema`

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
schrockn: For a factory I also think our naming convention should be:

`UnderlyingClassResource(ConfigurableResourceFactory[UnderlingClass]`

so in this case

`SnowflakeConnectionResource(ConfigurableResourceFactory[SnowflakeConnection]`

schrockn (changes requested):
Don't think the overload of init is tenable as a way to deal with schema

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -335,11 +340,238 @@ def write_parquet_file(context):
...
+            raise DagsterError(
+                "Snowflake Resource: 'paramstyle' configuration value must be one of:"
+                f" {','.join(valid_config)}"
+            )
+        return v
+
+    @validator("connector")
+    def validate_connector(cls, v: Optional[str]) -> Optional[str]:
+        if v is not None and v != "sqlalchemy":
+            raise DagsterError(
```
jamiedemaria, line 561: Is there a standard error class we want to raise for config validation?

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
jamiedemaria: moved to the `ConfigurableResource` with a `get_client` method pattern

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
jamiedemaria: has there been any discussion on this front?

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -33,6 +33,11 @@ class SnowflakeConnection:
     """
 
     def __init__(self, config: Mapping[str, str], log):
+        config = cast(Dict[str, str], config)
+        # with pydantic resources the schema config field will be named schema_
+        if config.get("schema_") is not None:
+            config["schema"] = config["schema_"]
+
```
schrockn, line 40: I don't think this should be necessary.

```
from dagster import Config
from pydantic import Field


class ConfigWithSchema(Config):
    schema_: str = Field(alias="schema")


assert ConfigWithSchema(schema="foo")._as_config_dict() == {"schema": "foo"}
```

schrockn (changes requested):
Question about `schema` versus `schema_`

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
jamiedemaria: it was failing before, but that was before ben got the alias piped all the way through, so it's probably good now. thanks for the catch!

schrockn (changes requested):
Just for exception type convo. cc: @benpankow @yuhan we should come to agreement here and then document in a norms page. (Search for "Pythonic config &  resources: Common questions, issues, & patterns" in Notion)

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -335,11 +334,238 @@ def write_parquet_file(context):
         self.execute_queries(sql_queries)
 
 
+class SnowflakeResource(ConfigurableResource):
+    """A resource for connecting to the Snowflake data warehouse. The returned resource object is an
+    instance of :py:class:`SnowflakeConnection`.
```
jamiedemaria, line 339:
```suggestion
    
```
no longer true

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -335,11 +334,237 @@ def write_parquet_file(context):
...
+                    password=EnvVar("SNOWFLAKE_PASSWORD")
+                    database="MY_DATABASE",
+                    schema="MY_SCHEMA",
+                    warehouse="MY_WAREHOUSE"
+                )}
+            )
+            def my_snowflake_job():
+                get_one()
+
+            my_snowflake_job.execute_in_process()
```
schrockn, line 365: let's put the resources on `execute_in_process`

schrockn (approved):

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -335,11 +334,239 @@ def write_parquet_file(context):
...
+
+    @validator("connector")
+    def validate_connector(cls, v: Optional[str]) -> Optional[str]:
+        if v is not None and v != "sqlalchemy":
+            raise ValueError(
+                "Snowflake Resource: 'connector' configuration value must be None or sqlalchemy."
+            )
+        return v
+
+    def get_client(self) -> SnowflakeConnection:
```
jamiedemaria, line 561: my final question/concern is what should this method be called? Based on previous convos, it seems like it should be `get_connection` since it returns a `SnowflakeConnection` object. However, the snowflake library also has a `SnowflakeConnection` class (that we use in our SnowflakeConnection class). 

My concern is that `get_connection` might be confusing/misleading as users will think they are getting a snowflake library `SnowflakeConnection`, when in fact they are getting a dagster `SnowflakeConnection`

Another kind of annoying thing caused by our own `SnowflakeConnector` class is that using the snowflake resource is a bit different than other db resources. For example 

```
@asset 
def my_asset(snowflake: SnowflakeResource):
    snowflake.get_client() # Returns dagster_snowflake.SnowflakeConnection
    with snowflake.get_client().get_connection() as conn:
           # conn is a snowflake.SnowflakeConnection
          conn.execute("SELECT *")
    # but we also have this method
    snowflake.get_client().execute_query("SELECT *") 
```
as an example, BigQuery is much more bare bones
```
@asset 
def my_asset(bigquery: BigQueryResource):
     with bigquery.get_client() as client:
        # client is a bigquery.Client
       client.query("SELECT *")
```

cc @yuhan  @benpankow  @schrockn  for your thoughts

schrockn: Sounds like the core problem is pre-existing: That both we and snowflake have a `SnowflakeConnection` class. Being forced to write a `get_connection` method poke at the existing problem.

Does `dagster_snowflake.SnowflakeConnection` add sufficient value over `snowflake.SnowflakeConnection`?

* If not we could use this opportunity to divorce ourselves from it and instead just have `get_connection` return `snowflake.SnowflakeConnection`
* If it does, we can name it something new and deprecate the old class name by having it be a trivial subclass for backwards compat.

jamiedemaria: Our SnowflakeConnector provides some helper methods (execute_query, execute_queries, and a method to load a parquet file into snowflake) and, but the most complicated/helpful thing it does is handle authentication. I know of one user who is using the `execute_queries` helper methods, but in my opinion it's no more challenging to use the snowflake connection object to execute some queries, and it's easy enough to switch to using the snowflake connector directly

I'm putting up a commit with a potential solution where we dont use the Dagster SnowflakeConnection anymore in the new resource. It's a bit rough but let me know what you think. it'll be easy to revert if it doesn't make sense as a path forward

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -64,64 +275,76 @@ def __init__(self, config: Mapping[str, str], log):
...
-                self.conn_args["private_key"] = self.__snowflake_private_key(config)
-
-        self.autocommit = self.conn_args.get("autocommit", False)
-        self.log = log
-
-    def __snowflake_private_key(self, config) -> bytes:
+        return v
+
+    @property
+    @cached_method
```
jamiedemaria, line 281: alternate implementation - we factor out `_connection_args` `_sqlalchemy_connection_args` `_sqlalcehmy_engine_args` and `get_connection` to a third class and then wrap the `get_connection` call in the `SnowflakeResource` and `SnowflakeConnection`

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
jamiedemaria: here's some code snippets outlining how to do the same things with the old resource and the new resource if the new resource just returns a snowflake SnowflakeConnection 

```
@asset(
    required_resource_keys={"snowflake"}
)
def run_query_old(context):
    snowflake = context.resources.snowflake
    snowflake.execute_query("SELECT 1", fetch_results=True, use_pandas_result=True)

@asset
def run_query_new(snowflake: SnowflakeResource):
    with snowflake.get_connection() as conn:
        conn.cursor().execute("SELECT 1").fetch_pandas_all()


@asset(
    required_resource_keys={"snowflake"}
)
def run_queries_old(context):
    snowflake = context.resources.snowflake
    results = snowflake.execute_queries(["SELECT 1", "SELECT 2", "SELECT 3"], fetch_results=True, use_pandas_result=True)

@asset
def run_queries_new(snowflake: SnowflakeResource):
    queries = ["SELECT 1", "SELECT 2", "SELECT 3"]
    results = []
    with snowflake.get_connection() as conn:
        for q in queries:
            results.append(conn.cursor().execute(q).fetch_pandas_all())
```

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -26,25 +26,236 @@
     raise
 
 
-class SnowflakeConnection:
-    """A connection to Snowflake that can execute queries. In general this class should not be
-    directly instantiated, but rather used as a resource in an op or asset via the
-    :py:func:`snowflake_resource`.
+class SnowflakeResource(ConfigurableResource):
```
jamiedemaria, line 29: TODO - implement IAttach... interface

python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
```diff
@@ -173,20 +403,73 @@ def get_query_status(context, query_id):
...
             yield conn
             if not self.autocommit:
                 conn.commit()
             conn.close()
 
+    def get_object_to_set_on_execution_context(self) -> Any:
+        return SnowflakeConnection(
+            config=self._resolved_config_dict,
+            log=get_dagster_logger(),
+            snowflake_connection_resource=self,
```
benpankow, line 426: Just to make sure - `SnowflakeConnection` here is only used for backcompat reasons, e.g. for the old resource and for injecting the old resource into old-style ops?

jamiedemaria: yes!

yuhan: can we leave a comment about the backcompat here?

yuhan (approved):

PR #13348, opened 2023-04-03
[pythonic resources][dagit] Display resources in search

benpankow:
## Summary

Adds resources in Dagit's search results, if the flag is enabled:

Screen Shot 2023-04-03 at 3 23 58 PM

## Test Plan

Tested locally.

### DIFF
--- a/js_modules/dagit/packages/core/src/search/useRepoSearch.tsx
+++ b/js_modules/dagit/packages/core/src/search/useRepoSearch.tsx
@@ -2,6 +2,7 @@ import {gql, useLazyQuery} from '@apollo/client';
 import Fuse from 'fuse.js';
 import * as React from 'react';
 
+import {useFeatureFlags} from '../app/Flags';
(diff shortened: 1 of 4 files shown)


js_modules/dagit/packages/core/src/search/useRepoSearch.tsx
```diff
@@ -10,6 +10,7 @@ import {workspacePath} from '../workspace/workspacePath';
 
 import {SearchResult, SearchResultType} from './types';
 import {SearchBootstrapQuery, SearchSecondaryQuery} from './types/useRepoSearch.types';
+import { useFeatureFlags } from '../app/Flags';
```
hellendag, line 13: Lint should fix this.

hellendag (approved):
Seems ok to me. Does it need to be feature flagged? If the flag or other gating prevents people from creating resources elsewhere, then there won't be any returned from the backend at all, and you'd have the same rendered result without adding the flag here. Or is it the case that people can add resources, but the flag prevents them from appearing in UI?

PR #13349, opened 2023-04-03
in asset reconciliation, only take action on partitions that are part of PartitionsDefinition

sryza:
## Summary & Motivation

Before this change, if we found an asset materialization with a poorly formatted or out-of-range partition, we would try to use it, which could result in parse errors or other oddities.


clairelin135 (approved):
LGTM

PR #13350, opened 2023-04-03
Asset backfill GQL targeted ranges resolver

clairelin135:
Adds a resolver to `PartitionBackfill` that returns back the targeted root asset partitions in an asset backfill.
- for time window assets, returns back a list of ranges
- for other assets, return back a set of selected partition keys

Additionally refactors the asset partition status counts and the targeted partitions to be inside a `assetBackfillData` resolver, since these fields only apply to asset backfills.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/schema/backfill.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/backfill.py
@@ -2,6 +2,9 @@ from typing import TYPE_CHECKING, Optional, Sequence
 
 import dagster._check as check
 import graphene
+from dagster._core.definitions.time_window_partitions import (
+    TimeWindowPartitionsSubset,
+)
 from dagster._core.execution.backfill import BackfillPartitionsStatus, PartitionBackfill
 from dagster._core.host_representation.external import ExternalPartitionSet
 from dagster._core.storage.pipeline_run import RunPartitionData, RunRecord, RunsFilter
@@ -29,11 +32,12 @@ from .util import ResolveInfo, non_null_list
 
 if TYPE_CHECKING:
     from dagster_graphql.schema.partition_sets import (
-        GrapheneAssetPartitionsStatusCounts,
         GraphenePartitionStatusCounts,
     )
 
-    from ..schema.partition_sets import GraphenePartitionSet
+    from ..schema.partition_sets import (
+        GraphenePartitionSet,
+    )
     from .pipelines.pipeline import GrapheneRun
 
 pipeline_execution_error_types = (
@@ -101,6 +105,19 @@ class GrapheneBulkActionStatus(graphene.Enum):
         name = "BulkActionStatus"
 
 
+class GrapheneAssetBackfillData(graphene.ObjectType):
+    class Meta:
+        name = "AssetBackfillData"
+
+    assetPartitionsStatusCounts = non_null_list(
+        "dagster_graphql.schema.partition_sets.GrapheneAssetPartitionsStatusCounts"
+    )
+    rootAssetTargetedRanges = graphene.List(
+        graphene.NonNull("dagster_graphql.schema.partition_sets.GraphenePartitionKeyRange")
+    )
+    rootAssetTargetedPartitions = graphene.List(graphene.NonNull(graphene.String))
+
+
 class GraphenePartitionBackfill(graphene.ObjectType):
     class Meta:
         name = "PartitionBackfill"
@@ -132,9 +149,8 @@ class GraphenePartitionBackfill(graphene.ObjectType):
     partitionStatusCounts = non_null_list(
         "dagster_graphql.schema.partition_sets.GraphenePartitionStatusCounts"
     )
-    assetPartitionsStatusCounts = non_null_list(
-        "dagster_graphql.schema.partition_sets.GrapheneAssetPartitionsStatusCounts"
-    )
+    isAssetBackfill = graphene.NonNull(graphene.Boolean)
+    assetBackfillData = graphene.Field(GrapheneAssetBackfillData)
 
     hasCancelPermission = graphene.NonNull(graphene.Boolean)
     hasResumePermission = graphene.NonNull(graphene.Boolean)
@@ -260,11 +276,13 @@ class GraphenePartitionBackfill(graphene.ObjectType):
             check.not_none(self._backfill_job.get_partition_names(graphene_info.context)),
         )
 
-    def resolve_assetPartitionsStatusCounts(
-        self, graphene_info: ResolveInfo
-    ) -> Sequence["GrapheneAssetPartitionsStatusCounts"]:
+    def resolve_isAssetBackfill(self, _graphene_info: ResolveInfo) -> bool:
+        return self._backfill_job.is_asset_backfill
+
+    def resolve_assetBackfillData(self, graphene_info: ResolveInfo) -> GrapheneAssetBackfillData:
         from dagster_graphql.schema.partition_sets import (
             GrapheneAssetPartitionsStatusCounts,
+            GraphenePartitionKeyRange,
         )
 
         status_counts_by_asset = (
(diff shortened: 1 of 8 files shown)


python_modules/dagster-graphql/dagster_graphql/schema/backfill.py
```diff
@@ -132,9 +149,7 @@ class Meta:
     partitionStatusCounts = non_null_list(
         "dagster_graphql.schema.partition_sets.GraphenePartitionStatusCounts"
     )
-    assetPartitionsStatusCounts = non_null_list(
-        "dagster_graphql.schema.partition_sets.GrapheneAssetPartitionsStatusCounts"
-    )
+    assetBackfillData = graphene.Field(GrapheneAssetBackfillData)
```
salazarm, line 152: Should this be non nullable?

python_modules/dagster-graphql/dagster_graphql/schema/backfill.py
```diff
@@ -101,6 +105,19 @@ class Meta:
...
+    class Meta:
+        name = "AssetBackfillData"
+
+    assetPartitionsStatusCounts = non_null_list(
+        "dagster_graphql.schema.partition_sets.GrapheneAssetPartitionsStatusCounts"
+    )
+    rootAssetTargetedRanges = graphene.List(
+        graphene.NonNull("dagster_graphql.schema.partition_sets.GraphenePartitionKeyRange")
+    )
+    rootAssetTargetedPartitions = graphene.List(graphene.String)
```
salazarm, line 118: `graphene.List(graphene.NonNull(graphene.String))`

python_modules/dagster-graphql/dagster_graphql/schema/partition_sets.py
```diff
@@ -133,6 +133,14 @@ class Meta:
     numPartitionsFailed = graphene.NonNull(graphene.Int)
 
 
+class GraphenePartitionKeyRange(graphene.ObjectType):
+    class Meta:
+        name = "PartitionKeyRange"
+
+    start = graphene.NonNull(graphene.String)
```
salazarm, line 140: instead of a string can this just be the timestamp?

salazarm: fine if not

python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -90,6 +91,24 @@ def get_target_root_asset_partitions(self) -> Iterable[AssetKeyPartitionKey]:
...
+        if (
+            not root_asset_keys
+            or self.target_subset.asset_graph.get_partitions_def(next(iter(root_asset_keys)))
+            is None
+        ):
+            return None
+
+        # An asset backfill must target the same partitions for the root assets
+        # Return the partitions subset for the first root asset
+        return self.target_subset.get_partitions_subset(next(iter(root_asset_keys)))
```
sryza, line 110: What if one of the roots isn't partitioned but the others are?

python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -90,6 +91,24 @@ def get_target_root_asset_partitions(self) -> Iterable[AssetKeyPartitionKey]:
...
 
+    def get_target_root_partitions_subset(self) -> Optional[PartitionsSubset]:
+        root_asset_keys = (
+            AssetSelection.keys(*self.target_subset.asset_keys)
+            .sources()
+            .resolve(self.target_subset.asset_graph)
+        )
+
+        if (
+            not root_asset_keys
```
sryza, line 102: I don't think this should ever be the case - a backfill targeting no assets?

python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -90,6 +91,24 @@ def get_target_root_asset_partitions(self) -> Iterable[AssetKeyPartitionKey]:
...
+            or self.target_subset.asset_graph.get_partitions_def(next(iter(root_asset_keys)))
```
sryza, line 103: I don't think this is allowed. I guess it could happen if someone removes the partitions defs from an asset that was previously partitioned?

python_modules/dagster-graphql/dagster_graphql/schema/backfill.py
clairelin135: I don't think it always exists though, since the asset backfill data will only exist on asset backfills, not job backfills

python_modules/dagster/dagster/_core/execution/asset_backfill.py
clairelin135: I thought we only allow creating an asset backfill when the roots share the same partitioning?

https://github.com/dagster-io/dagster/blob/607478333d44788b4063de708a9a0ea01a2a6ec9/python_modules/dagster/dagster/_core/execution/asset_backfill.py#L217

python_modules/dagster-graphql/dagster_graphql/schema/backfill.py
clairelin135: 👍

python_modules/dagster-graphql/dagster_graphql/schema/partition_sets.py
clairelin135: I think i'd prefer to keep it as a string, so we can generalize the partition key range for other types of partitions defs too

python_modules/dagster-graphql/dagster_graphql/schema/backfill.py
clairelin135: ah actually, the intent was to only populate the ranges for time window partitions, and for other partition types to populate the set of specific partitions (since in static cases, it doesn't make as much sense to display ranges)

clairelin135: let me know if you think we should populate in all cases

python_modules/dagster-graphql/dagster_graphql/schema/backfill.py
salazarm: Ah okay makes sense thanks! Can we also export a flag `isAssetBackfill` that will effectively check that `assetBackfillData` exists? This is so that on the backfill overview page I can link to the new asset backfill page only if it's a asset backfill.

python_modules/dagster-graphql/dagster_graphql/schema/backfill.py
salazarm: The list itself would still be nullable. My suggestion is to mark the contents of the list as non nullable.

it's kind of silly but by default lists in graphene are nullable. That means the typescript type generated for this looks like:

```
rootAssetTargetedPartitions: Array<string | null> | null;
```
Notice how the contents of the array itself are nullable.

After my suggestion it will look like this:
```
rootAssetTargetedPartitions: Array<string> | null;
```

The array itself is still nullable but not the contents of the array

clairelin135: makes sense!

python_modules/dagster/dagster/_core/execution/asset_backfill.py
sryza: That's after excluding non-partitioned assets (look at the lines directly above that link)

salazarm (approved):
This is working for me, thanks!

python_modules/dagster/dagster/_core/execution/asset_backfill.py
clairelin135: Ok, I've updated this.

I think it's possible that a backfill can still be created when the upstream asset is unpartitioned and the downstream assets are partitioned. I think in that case, we want to display the partition ranges of the most upstream partitioned asset, which is what I've amended this PR to do. 

I think in all cases there should be at least one partitioned asset in the targeted backfill, let me know if this is incorrect @sryza

clairelin135:
@sryza would you mind taking a last look at the logic to determine the target root asset partitions?

sryza (approved):
That logic looks right to me!

PR #13354, opened 2023-04-04
docs infra: markdoc 1/ mvp

yuhan:
### Why?
Although MDX is very flexible, it has limitations that hold us back in developing the docs, e.g.:
- MDX doesn't support importing other mdx snippets in a mdx file, which prevents us from reusing our docs snippets across pages.
- MDX conflicts with our formatting a lot, which hurts the docs authoring experience and sometimes breaks master bk.

### What?
This PR sets up the docs infra to use Markdoc as the default content framework and prep the migration from mdx to Markdoc.

This PR does the V1:
- Existing capabilities to keep:
  * redirects ✅ 
  * re-use layout/UI components
  * PyObject
  * code snapshot
  * link test
  * emoji support ✅ 
  * right sidebar for toc ✅ 
- Partials (so we can reuse md content across pages)

To do:
- [ ] Figure out ways to migrate/reuse components
- [ ] Tests for broken links

### How's the migration path gonna look like?
It's opt-in so that we don't have to migrate everything all at once.
- One can opt-in to migrate content such as [docs/content/getting-started/install.mdx → docs/next/pages/getting-started/install.md](https://github.com/dagster-io/dagster/pull/13354/files#diff-c3d719309d2ad233d3f4c0f700fa98737a62859b91f3da2395c91861e27ee1f0) and add a `"nonMdx": true` in [docs/content/_navigation.json](https://github.com/dagster-io/dagster/pull/13354/files#diff-aad58bfb7b150647e2256c5de24ccb2efc2d48c2c764bbbb3bdf0065c2216827) so renderer knows to not use mdx renderer

    Implementation details: the docs infra routes to `docs/next/pages/path-to-file` file if exists, then fall back to `docs/content/path-to-file`. nextjs routes url to page based on the folder structure in `pages` ([docs](https://nextjs.org/docs/basic-features/pages)), so when `docs.dagster.io/X` where `docs/next/pages/X` exists, it would directly route to that file, such as [pages/changelog.tsx](https://github.com/dagster-io/dagster/blob/master/docs/next/pages/changelog.tsx); when `docs/next/pages/X` doesn't exist, it goes to [[...page].tsx](https://github.com/dagster-io/dagster/blob/master/docs/next/pages/%5B...page%5D.tsx) where we conditionally route to `content/` + versioning using custom render.

- No longer `mdx`: Markdoc uses `.md` or `.mdoc`
- If templated, add the templates to `docs/next/markdoc/partials`, e.g. [docs/next/markdoc/partials/DagsterVersion.mdx](https://github.com/dagster-io/dagster/pull/13354/files#diff-6604396c31df36407a4e30e38f1c26e279a6da2022da01cb1958523f904ffad2)

### How do we handle versions going forward?

As the docs may no longer be in `docs/content/` folder (which is the versioned folder), the current version scheme won't work. Here's a proposal: https://github.com/dagster-io/internal/discussions/5767

### Next steps in V2:
- Loops: generate a list of general guides based on e.g. `tags: general`
- Text transformation/customization: display text based on page’s frontmatter


yuhan:
* master
  * **PR #13354** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13354" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13354?utm_source=stack-comment).

erinkcochran87:
I'm so so excited about this! Some thoughts/questions after tinkering around with it a bit this afternoon - there's more I'd like to test with, but here's an initial braindump:

- Why is `markdoc` needed in Frontmatter access (ex: `{% markdoc.frontmatter.title %}`, [but the docs don't use it](https://markdoc.dev/docs/frontmatter#access-frontmatter-values)? Is this something to do with this test implementation?
- I could use some help figuring out how to implement loops and switch statements ([docs](https://markdoc.dev/docs/examples#loops)) - my React/JS knowledge is pretty limited. I'd love to pair on this so I can learn if ya don't mind.
- Is there an out-of-the-box way to transform text, or would it require building a [custom function](https://markdoc.dev/docs/functions#creating-a-custom-function)?

   Ex: In the past with Liquid, I could do this:
   
   ```
   {{ "Install Dagster | Dagster Docs" | remove: " | Dagster Docs" }}
   
   Output:
   Install Dagster
   ```

   Or:
   
   ```
   {{ "a great title" | capitalize | append: "!" }}

   Output:
   A great title!
   ```
   
   This is really handy for templates and manipulating data based on context.

yuhan:
this stack is too old to pick up back, so retiring in favor of new stack here: https://github.com/dagster-io/dagster/pull/19093

PR #13361, opened 2023-04-04
[docs] Fix example & mention target in DBT reference

NicolasPA:
## Summary & Motivation

Questions about how to load DBT assets with a different DBT target are fairly frequent. https://dagster.slack.com/archives/C04CW71AGBW/p1680514153349179

So I suggest adding to an existing example the setting of the DBT target depending on the execution environment.

I'm also suggesting a fix to the `load_assets_from_dbt_manifest` example because it is not functional.


sryza:
@rexledesma - mind taking a look?

PR #13359, opened 2023-04-04
Revert 13323 and 12607

schrockn:
## Summary & Motivation

Reverting sensor diffs (#13323 and #12607) that caused test failures

See https://buildkite.com/dagster/dagster/builds/54620#01874c58-b659-4072-aeea-06ad3b305201

## How I Tested These Changes

Locally tested failing tests and buildkite

Issue #2607, opened 2020-06-16
Consider slack/phabricator integration


schrockn:
Maybe https://github.com/etcinit/phabulous?

It's been suggested that it would be good for folks to be able to opt-in to see all organizational activity in order to get a sense of how we work and to get context about what folks are working on.

alangenfeld:
@rexledesma did this

PR #13367, opened 2023-04-04
Re-implement schedule/sensor changes

benpankow:
## Summary

Undoes the revert (#13359) of some schedule/sensor changes (#13359 and #2607).

See this commit for the actual changes (first commit is the revert-of-the-revert):
https://github.com/dagster-io/dagster/pull/13367/commits/649b9e7f4cd2156701cb98122fd6cf83c55c7c53

## Change 1

Implements similar functionality to #13002 to make passing resources to sensors and schedules easier in the direct invocation case. This makes testing in particular easier.

The new API changes enable the following (schedule behavior is identical):

```python
    a_str: str

@sensor(job_name="foo_pipeline")
def basic_sensor_resource_req(my_resource: MyResource):
    return RunRequest(run_key=None, run_config={"foo": my_resource.a_str}, tags={})


# Can pass resource through context
assert cast(
    RunRequest,
    basic_sensor_resource_req(
        context=build_sensor_context(resources={"my_resource": MyResource(a_str="foo")})
    ),
).run_config == {"foo": "foo"}

# Can pass resource directly
assert cast(
    RunRequest,
    basic_sensor_resource_req(my_resource=MyResource(a_str="foo")),
).run_config == {"foo": "foo"}


# Can pass resource directly with context
assert cast(
    RunRequest,
    basic_sensor_resource_req(build_sensor_context(), my_resource=MyResource(a_str="foo")),
).run_config == {"foo": "foo"}

```

## Change 2

Adding resource support for all of our various complex sensor types (`@asset_sensor`, `@multi_asset_sensor`, `@freshness_policy_sensor`, and `@run_status_sensor`).

For most cases, this is straightforward replication of what's done in #12401. There are a few interesting cases, e.g. asset sensors currently take two inputs (context and event log entry). This is now set up so that the first two non-resource params are treated as context and event log entry, respectively.

## Test Plan

Ran previously broken run status sensor tests locally.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py
@@ -225,7 +225,7 @@ class MultiAssetSensorEvaluationContext(SensorEvaluationContext):
         repository_def: Optional["RepositoryDefinition"],
         monitored_assets: Union[Sequence[AssetKey], AssetSelection],
         instance: Optional[DagsterInstance] = None,
-        resources: Optional[Mapping[str, ResourceDefinition]] = None,
+        resource_defs: Optional[Mapping[str, ResourceDefinition]] = None,
         definitions: Optional["Definitions"] = None,
     ):
         from dagster._core.definitions.definitions_class import Definitions
@@ -274,7 +274,7 @@ class MultiAssetSensorEvaluationContext(SensorEvaluationContext):
             repository_name=repository_name,
             instance=instance,
             repository_def=repository_def,
-            resources=resources,
+            resources=resource_defs,
         )
 
     def _cache_initial_unconsumed_events(self) -> None:
@@ -1025,7 +1025,7 @@ def build_multi_asset_sensor_context(
         instance=instance,
         monitored_assets=monitored_assets,
         repository_def=repository_def,
-        resources=resources,
+        resource_defs=resources,
     )
 
 
@@ -1111,7 +1111,11 @@ class MultiAssetSensorDefinition(SensorDefinition):
                             " based on the latest materialization for each monitored asset."
                         )
 
-                multi_asset_sensor_context = MultiAssetSensorEvaluationContext(
+                resource_args_populated = validate_and_get_resource_dict(
+                    context.resources, name, resource_arg_names
+                )
+
+                with MultiAssetSensorEvaluationContext(
                     instance_ref=context.instance_ref,
                     last_completion_time=context.last_completion_time,
                     last_run_key=context.last_run_key,
@@ -1121,12 +1125,7 @@ class MultiAssetSensorDefinition(SensorDefinition):
                     monitored_assets=monitored_assets,
                     instance=context.instance,
                     resources=context.resource_defs,
-                )
-                resource_args_populated = validate_and_get_resource_dict(
-                    context.resources, name, resource_arg_names
-                )
-
-                with multi_asset_sensor_context:
+                ) as multi_asset_sensor_context:
                     context_param_name = get_context_param_name(materialization_fn)
                     context_param = (
                         {context_param_name: multi_asset_sensor_context}
--- a/python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py
@@ -19,10 +19,12 @@ from typing import (
 
 import dagster._check as check
 from dagster._annotations import experimental, public
-from dagster._core.decorator_utils import has_at_least_one_parameter
 from dagster._core.definitions.asset_selection import AssetSelection
 from dagster._core.definitions.assets import AssetsDefinition
 from dagster._core.definitions.partition import PartitionsDefinition
+from dagster._core.definitions.resource_annotation import get_resource_args
+from dagster._core.definitions.resource_definition import ResourceDefinition
+from dagster._core.definitions.scoped_resources_builder import ScopedResourcesBuilder
 from dagster._core.errors import (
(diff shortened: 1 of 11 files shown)


benpankow:
* master
  * **PR #13367** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13367" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13367?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -134,6 +162,49 @@ def for_run_failure(self):
...
+
+            if isinstance(self._resources, IContainsGenerator) and not self._cm_scope_entered:
+                self._exit_stack.close()
+                raise DagsterInvariantViolationError(
+                    "At least one provided resource is a generator, but attempting to access"
+                    " resources outside of context manager scope. You can use the following syntax"
+                    " to open a context manager: `with build_schedule_context(...) as context:`"
+                )
+
+        return self._resources
```
schrockn, line 206: Not your fault as you are copying existing pattern in the codebase, but I'm pretty scandalized that we decided that this sort of stateful/complicated/fragile sort of manual management of context managers intermixed in the object instance is a desirable pattern.

schrockn (dismissed):
Ok.

I have one request here which is to add a test case the demonstrates the purported value of making the context itself a contextmanager. I assume this is facilitate unit testing. If that is not true and this is an internal structure IMO we should separate the contextmanager and the context itself.

benpankow:
> a test case the demonstrates the purported value of making the context itself a contextmanager. I assume this is facilitate unit testing

Here's an example of where that's useful:

https://github.com/dagster-io/dagster/blob/649b9e7f4cd2156701cb98122fd6cf83c55c7c53/python_modules/dagster/dagster_tests/definitions_tests/test_sensor_invocation.py#L255

```python
 with build_sensor_context(resources={"my_resource": my_cm_resource}) as context:
   basic_sensor_str_resource_req(context))
```

Where `my_cm_resource` is a context manager. I think there are a few other ways we could handle this. One is to make `build_sensor_context` itself be a context manager, but this makes unit testing the non-CM case more annoying:

```python
@contextmanager
def build_sensor_context() -> Generator[SensorContext, None, None]:
    with build_resources() as resources:
        yield SensorContext(resources)


with build_sensor_context() as context:
    my_sensor(context)

# no longer works
my_sensor(build_sensor_context())
```

A way to potentially remedy this would be to let a sensor function take either (1) a Context object or (2) a generator for a Context object. In the latter case it'd just open the context around its internals. What do you think?

python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py
```diff
@@ -935,6 +940,7 @@ def build_multi_asset_sensor_context(
     cursor: Optional[str] = None,
     repository_name: Optional[str] = None,
     cursor_from_latest_materializations: bool = False,
+    resources: Optional[Mapping[str, ResourceDefinition]] = None,
```
schrockn, line 943: resource_defs

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -116,14 +122,36 @@ class RunStatusSensorContext:
...
 
-    def __init__(self, sensor_name, dagster_run, dagster_event, instance, context=None):
+    def __init__(
+        self,
+        sensor_name,
+        dagster_run,
+        dagster_event,
+        instance,
+        context=None,
+        resources: Optional[Mapping[str, "ResourceDefinition"]] = None,
```
schrockn, line 132: resource_defs

schrockn (dismissed):

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -116,14 +122,36 @@ class RunStatusSensorContext:
...
     """
 
-    def __init__(self, sensor_name, dagster_run, dagster_event, instance, context=None):
+    def __init__(
+        self,
+        sensor_name,
+        dagster_run,
+        dagster_event,
+        instance,
+        context=None,
```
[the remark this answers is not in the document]
benpankow: This is pre-existing, my guess is to pass along the logger. Updated to take a logger param explicitly and ignore a context param.

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -610,24 +699,31 @@ def _wrapped_fn(
...
-                        if has_at_least_one_parameter(run_status_sensor_fn):
-                            # one user code invocation maps to one failure event
-                            sensor_return = run_status_sensor_fn(
-                                RunStatusSensorContext(
-                                    sensor_name=name,
-                                    dagster_run=pipeline_run,
-                                    dagster_event=event_log_entry.dagster_event,
-                                    instance=context.instance,
-                                    context=context,
-                                )
```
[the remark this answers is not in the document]
benpankow: It wasn't - the context management is introduced to support resources as part of this PR.

schrockn (approved):
Alright let's get this in. I may take a stab at refactoring these contexts to not be context managers themselves later.

benpankow:
> Ok so should it take a context at all now if the only purpose was to pass along the logger?

`RunStatusSensorContext` is exported, so I didn't want to adjust the signature.  I'll add a comment - we still will pull the logger of the context if provided, but there's a dedicated arg our internal codepaths prefer now.

PR #13373, opened 2023-04-04
[dagster] remove required positional argument to __dir__

brad-alexander:
## Summary & Motivation
fixes:

```py
>>> import dagster
>>> dir(dagster)
TypeError: __dir__() missing 1 required positional argument: '_self'
```
according to [pep562](https://peps.python.org/pep-0562/#:~:text=The%20__dir__%20function%20should%20accept%20no%20arguments%2C%20and%20return%20a%20list%20of%20strings%20that%20represents%20the%20names%20accessible%20on%20module%3A) for `__dir__` on modules:
> The `__dir__` function should accept no arguments, and return a list of strings that represents the names accessible on module:
## How I Tested These Changes
bk

### DIFF
--- a/python_modules/dagster/dagster/__init__.py
+++ b/python_modules/dagster/dagster/__init__.py
@@ -579,5 +579,5 @@ def __getattr__(name: str) -> TypingAny:
         raise AttributeError(f"module '{__name__}' has no attribute '{name}'")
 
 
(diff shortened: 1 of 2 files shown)


smackesey (approved):
thx for the fix

PR #13375, opened 2023-04-04
[pythonic config] Fix deep nesting of data structures in Pythonic config

benpankow:
## Summary

Fixes Pydantic -> Dagster Type conversion when using several-level-deep data structures, e.g.

```python
class AnOpConfig(Config):
    a_complex_thing: Mapping[int, List[Mapping[str, Optional[int]]]]
```

Previously, we weren't applying the "data structure processing logic" on the inner type of Pydantic data structure shapes (lists, dicts etc).

## Test Plan

New unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_config/structured_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/structured_config/__init__.py
@@ -783,6 +783,27 @@ def _wrap_config_type(
         raise NotImplementedError(f"Pydantic shape type {shape_type} not supported.")
 
 
+def _get_inner_field_if_exists(
+    shape_type: PydanticShapeType, field: ModelField
+) -> Optional[ModelField]:
+    """Grabs the inner Pydantic field type for a data structure such as a list or dictionary.
+
+    Returns None for types which have no inner field.
+    """
+    # See https://github.com/pydantic/pydantic/blob/v1.10.3/pydantic/fields.py#L758 for
+    # where sub_fields is set.
+    if shape_type == SHAPE_SINGLETON:
+        return None
+    elif shape_type == SHAPE_LIST:
+        # List has a single subfield, which is the type of the list elements.
+        return check.not_none(field.sub_fields)[0]
+    elif shape_type in MAPPING_TYPES:
+        # Mapping has a single subfield, which is the type of the mapping values.
+        return check.not_none(field.sub_fields)[0]
+    else:
+        raise NotImplementedError(f"Pydantic shape type {shape_type} not supported.")
+
+
 def _convert_pydantic_field(pydantic_field: ModelField) -> Field:
     """Transforms a Pydantic field into a corresponding Dagster config field."""
     key_type = (
@@ -806,10 +827,13 @@ def _convert_pydantic_field(pydantic_field: ModelField) -> Field:
 
         return Field(config=wrapped_config_type, description=inferred_field.description)
     else:
-        if pydantic_field.sub_fields:
-            config_type = _convert_pydantic_field(pydantic_field.sub_fields[0]).config_type
+        # For certain data structure types, we need to grab the inner Pydantic field (e.g. List type)
+        inner_field = _get_inner_field_if_exists(pydantic_field.shape, pydantic_field)
+        if inner_field:
+            config_type = _convert_pydantic_field(inner_field).config_type
         else:
             config_type = _config_type_for_pydantic_field(pydantic_field)
+
         wrapped_config_type = _wrap_config_type(
             shape_type=pydantic_field.shape, config_type=config_type, key_type=key_type
         )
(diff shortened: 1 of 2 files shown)


benpankow:
* master
  * **PR #13375** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13375" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13375?utm_source=stack-comment).

python_modules/dagster/dagster/_config/structured_config/__init__.py
```diff
@@ -806,7 +806,10 @@ def _convert_pydantic_field(pydantic_field: ModelField) -> Field:
 
         return Field(config=wrapped_config_type, description=inferred_field.description)
     else:
-        config_type = _config_type_for_pydantic_field(pydantic_field)
+        if pydantic_field.sub_fields:
+            config_type = _convert_pydantic_field(pydantic_field.sub_fields[0]).config_type
```
schrockn, line 810: does this only work if len(pydantic_field.sub_fields) is 1? If not, what happens to sub_fields[1:]?

schrockn (changes requested):

python_modules/dagster/dagster/_config/structured_config/__init__.py
benpankow: In the case of the Pydantic types we support (list, dict) there is only every a single subfield.

Updated our code to be more explicit about grabbing the subfield from each type.
https://github.com/pydantic/pydantic/blob/v1.10.3/pydantic/fields.py#L758

python_modules/dagster/dagster_tests/core_tests/structured_config_tests/test_structured_config_types.py
```diff
@@ -202,6 +202,60 @@ def a_job():
     assert executed["yes"]
 
 
+def test_struct_config_mapping_list() -> None:
+    class AnOpConfig(Config):
+        a_list_of_string_to_int_mapping: List[Mapping[str, int]]
+
```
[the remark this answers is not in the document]
benpankow: see above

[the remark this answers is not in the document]
schrockn: I'm still confused here.

What I'm asking for is

```
    class AnOpConfigWithMultipleFields(Config):
        a_list_of_string_to_int_mapping: List[Mapping[str, int]]
        a_list_of_list_of_strings: List[List[str]]
 ```
 
 What is the definition of subfield here?

File: python_modules/dagster/dagster/_config/structured_config/__init__.py [hunk 3]
```diff
@@ -783,6 +783,27 @@ def _wrap_config_type(
...
+    # See https://github.com/pydantic/pydantic/blob/v1.10.3/pydantic/fields.py#L758 for
+    # where sub_fields is set.
+    if shape_type == SHAPE_SINGLETON:
+        return None
+    elif shape_type == SHAPE_LIST:
+        # List has a single subfield, which is the type of the list elements.
+        return check.not_none(field.sub_fields)[0]
+    elif shape_type in MAPPING_TYPES:
+        # Mapping has a single subfield, which is the type of the mapping values.
+        return check.not_none(field.sub_fields)[0]
```
schrockn, line 802: is the single sub_field typed as as tuple?

schrockn (changes requested):

python_modules/dagster/dagster/_config/structured_config/__init__.py
benpankow: No, it's typed to whatever the value type is:

```python
>>> class MyModel(BaseModel):
...     foo: Dict[str, int]
...
>>> MyModel.__fields__
{'foo': ModelField(name='foo', type=Mapping[str, int], required=True)}
>>> MyModel.__fields__['foo'].sub_fields
[ModelField(name='_foo', type=int, required=True)]
```

schrockn: Ah how strange

schrockn (approved):

python_modules/dagster/dagster_tests/core_tests/structured_config_tests/test_structured_config_types.py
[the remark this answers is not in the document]
benpankow: Ah yeah, a subfield is the inner type for a generic data type like a `List` or `Map`

`a_list_of_string_to_int_mapping` has a subfield which is of type `Mapping[str, int]`, this in turn has a subfield `int`

`a_list_of_list_of_strings` has a subfield of `List[str]` which has subfield `str`

PR #13376, opened 2023-04-05
[pythonic config] Add better error message for unsupported config types

benpankow:
## Summary

Introduces a new error type which fires when the Pydantic -> Dagster Type code encounters an unconvertible config type. This is a bit more actionable than the typical `DagsterInvalidConfigDefinitionError` since it points to the config class & field name. It also includes more specific details about what config types are allowed and disallowed.

## Test Plan

New unit tests.

PR #13377, opened 2023-04-05
[pythonic config] Improve error message for bad config types directly on ops/assets

benpankow:
## Summary

Extends the better error messaging on the stacked PR to also affect bad annotations directly on an op or asset's `config` parame, e.g.

```python
@op
def my_op(config: Tuple[str, str]):
    pass
```

## Test Plan

New unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_config/structured_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/structured_config/__init__.py
@@ -34,6 +34,7 @@ from dagster._core.definitions.definition_config_schema import (
 from dagster._core.errors import (
     DagsterInvalidConfigDefinitionError,
     DagsterInvalidConfigError,
+    DagsterInvalidDefinitionError,
     DagsterInvalidPythonicConfigDefinitionError,
 )
 from dagster._core.execution.context.init import InitResourceContext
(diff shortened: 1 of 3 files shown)


benpankow:
* master
  * **PR #13376** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13376" target="_blank">Graphite</a> 
    * **PR #13377** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13377" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13377?utm_source=stack-comment).

python_modules/dagster/dagster/_config/structured_config/__init__.py
```diff
@@ -867,6 +868,14 @@ def _config_type_for_type_on_pydantic_field(
...
         try:
             return convert_potential_field(potential_dagster_type).config_type
+        except DagsterInvalidDefinitionError:
+            raise DagsterInvalidPythonicConfigDefinitionError(
+                config_class=model_cls,
+                field_name=field_name,
+                invalid_type=potential_dagster_type,
+                is_resource=model_cls is not None
+                and safe_is_subclass(model_cls, ConfigurableResourceFactory),
+            )
```
schrockn, line 878: similar feedback as previous PR. We should explicilty check that this is instead the `model_cls` case rather than assuming the fallthrough is this case

schrockn (changes requested):

python_modules/dagster/dagster/_config/structured_config/__init__.py
benpankow: extracted check to just wrap annotation-checking segment of the code

schrockn (approved):
great

PR #13390, opened 2023-04-05
improved keyboard navigation for filtering

salazarm:
## Summary & Motivation

Allowing arrow keys / escape to better control navigating the filter menu

## How I Tested These Changes

https://www.loom.com/share/a3a29d801d8745df8c605311f6281c4c


js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
```diff
@@ -65,51 +88,111 @@ export const FilterDropdown = ({filters, setIsOpen, setPortaledElements}: Filter
...
-  }, [filteredFilters, filters, results, search, selectValue, selectedFilter]);
+  }, [
+    selectedFilter,
+    filters,
+    search,
+    menuKey,
+    selectValue,
+    filteredFilters,
+    results,
+    setFocusedItemIndex,
```
hellendag, line 158: With this many dependencies, is the `useMemo` still useful?

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
```diff
@@ -261,6 +348,27 @@ const TextInputWrapper = styled.div`
...
+const MenuItem = React.memo(({menuKey, index, ...rest}: MenuItemProps) => {
+  const divRef = React.useRef<HTMLDivElement | null>(null);
+  const isFocused = useRecoilValue(
+    isFocusedSelector(React.useMemo(() => ({key: menuKey, index}), [index, menuKey])),
+  );
+  React.useLayoutEffect(() => {
+    if (isFocused) {
+      divRef.current?.querySelector('a')?.focus();
+    }
+  }, [isFocused]);
```
hellendag, line 364: Is it possible that there could be a search input that should retain focus even as the keyboard is used to navigate up/down to highlight results?

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
```diff
@@ -23,7 +31,21 @@ interface FilterDropdownProps {
...
+  key: 'FilterDropdown:isFocused',
+  get: ({key, index}) => ({get}) => get(focusedIndexAtom(key)) === index,
+});
+
 export const FilterDropdown = ({filters, setIsOpen, setPortaledElements}: FilterDropdownProps) => {
+  const [menuKey, _] = React.useState(() => uuidv4());
+  const setFocusedItemIndex = useSetRecoilState(focusedIndexAtom(menuKey));
+  const isSearchInputFocused = useRecoilValue(
+    isFocusedSelector(React.useMemo(() => ({key: menuKey, index: -1}), [menuKey])),
+  );
```
hellendag, line 48: Is recoil necessary to manage this state? It seems fairly straightforward, and it doesn't look like any props need to be passed far down the tree.

Also see comment re: highlighting menu items while retaining focus on the input.

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
```diff
@@ -65,51 +88,111 @@ export const FilterDropdown = ({filters, setIsOpen, setPortaledElements}: Filter
...
+    results,
+    setFocusedItemIndex,
+  ]);
+
+  const inputRef = useRef<HTMLInputElement | null>(null);
+  const dropdownRef = useRef<HTMLDivElement | null>(null);
+
+  useEffect(() => {
+    inputRef.current?.focus();
+  }, []);
```
hellendag, line 166: Consolidate with the `useLayoutEffect` above?

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
```diff
@@ -118,10 +201,10 @@ export const FilterDropdown = ({filters, setIsOpen, setPortaledElements}: Filter
           type="text"
           value={search}
           onChange={(e) => setSearch(e.target.value)}
+          onKeyUp={handleKeyUp}
```
hellendag: Keydown instead of keyup, so that I could potentially hold the key down to go through a long list of options?

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
```diff
@@ -261,6 +348,27 @@ const TextInputWrapper = styled.div`
   }
 `;
 
+type MenuItemProps = React.ComponentProps<typeof _MenuItem> & {
+  menuKey: string;
+  index: number;
+};
+const MenuItem = React.memo(({menuKey, index, ...rest}: MenuItemProps) => {
```
hellendag, line 355: Rename to something that doesn't conflict with the `ui` component? I could see this being pretty confusing in the Components dev tools, or for users who are new to this code.

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
salazarm: Mainly for the case where you're tabbing away from the search bar.

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
salazarm: I wanted focus to move with the active element. Otherwise the element that's active isn't the element that's focused and I don't think screen readers would pick that up.

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
salazarm: Yes Recoil allowed me to remove `focusedIndex` from the `useMemo` that re-renders the results. This means that when we change focus we don't re-render the entire list. This makes keyboard navigation a lot smoother.

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
salazarm: Forgot to remove this one

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
salazarm: Specifically when `isSearchInputFocused` changes, since it's not a dep here we won't re-render. I didn't want that to feel sluggish. Also it's possible down the line that some of the filters are async we don't want to call their search function more than once.

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
```diff
@@ -261,6 +349,45 @@ const TextInputWrapper = styled.div`
...
+  ({menuKey, index, ...rest}: FilterDropdownMenuItemProps) => {
+    const divRef = React.useRef<HTMLDivElement | null>(null);
+    const isFocused = useRecoilValue(
+      isFocusedSelector(React.useMemo(() => ({key: menuKey, index}), [index, menuKey])),
+    );
+    React.useLayoutEffect(() => {
+      if (isFocused) {
+        divRef.current?.querySelector('a')?.focus();
+      }
+    }, [isFocused]);
```
hellendag, line 366: Aria roles can/should be used for the "active" item state in a combobox -- you'll want to avoid taking DOM focus from the input while it's there. 

https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/

Using aria roles and the default menu item hover style, I think there's a good chance you don't need the `FilterDropdownMenuItemProps` wrapper component.

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
hellendag: Why is it so slow to rerender the results? It doesn't seem like it should be too expensive just rendering a list of keyed items, but I may be missing something.

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
```diff
@@ -261,6 +349,45 @@ const TextInputWrapper = styled.div`
...
+    const [hovered, setHovered] = React.useState(false);
+    return (
+      <div
+        ref={divRef}
+        onMouseEnter={() => {
+          setHovered(true);
+        }}
+        onMouseLeave={() => {
+          setHovered(false);
+        }}
```
hellendag, line 376: This visually merges the "selected" state and the "hover" state, which we don't do on menu items elsewhere and its a bit confusing. (https://dagit-storybook.vercel.app/?path=/story/suggest--default)

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
salazarm: I was looking at that page the other day and didn't find an example that matched this specific case. All of the examples on that page assume that the user can only make a single selection. In this case they can sometimes make more than one selection.

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
salazarm: > Why is it so slow to rerender the results? 

It's extra render code that is running which makes it slower. Since this is supposed to be a "core" component I think it's important to optimized renders. Especially since I'm planning on making some of these filters async down the line and I wouldn't want an async request to be fired more than necessary

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
salazarm: Weird, I wonder if @braunjj finds it confusing too. I actually liked it better which is why I added it but I can remove it if nobody else agrees.

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
salazarm: Okay there this example: https://ariakit.org/examples/combobox-multiple
This makes sense. Though our example won't work the same because I'm using MenuItem, which means tabbing will put you into the list.

salazarm: That fact about MenuItem^ is why I resorted to managing focus this way because otherwise it's weird that `tab` will move you in the list but looking at the example again I see they used negative tabIndex to disable that behavior so let me try that

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
```diff
@@ -23,7 +31,21 @@ interface FilterDropdownProps {
...
+  default: -1,
+});
+const isFocusedSelector = selectorFamily<boolean, {key: string; index: number}>({
+  key: 'FilterDropdown:isFocused',
+  get: ({key, index}) => ({get}) => get(focusedIndexAtom(key)) === index,
+});
+
 export const FilterDropdown = ({filters, setIsOpen, setPortaledElements}: FilterDropdownProps) => {
+  const [menuKey, _] = React.useState(() => uuidv4());
+  const setFocusedItemIndex = useSetRecoilState(focusedIndexAtom(menuKey));
```
bengotow, line 45: It seems like this is a code smell. Making up UUIDs so we can shove component-local state into a global state controller? I guess I'm sorta with @hellendag that this seems like it should be basic local state.

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
bengotow: Just a drive-by, but I think the solution here might be to use a virtualized list for the dropdown so that "re-render the entire list" is just 6 items? Otherwise to dish's point it still takes a while for the dropdown to appear because the initial render is still large?

salazarm: A Virtualized list will certainly help, but it's still an unnecessary extra render that can be skipped

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
salazarm: Optimizing renders in a list with a "focused" / "selected" state that changes on hover / interaction is actually a classic scenario where you don't want to use vanilla React state since it will cause more components to re-render than the ones that changed. Couple that with our extremely light usage of `useMemo` and `useCallback` and it's a recipe for re-rendering everything.

The other way to solve this (this = not re-rendering the whole list and only re-render the item losing focus and the item gaining focus) is to use context like I did here:  https://github.com/dagster-io/dagster/pull/10239

But that's uglier imo.

salazarm: The reason I want to prevent that render is to help callers of the component fall into the "Pit of success". Otherwise it's a bit like giving them a foot gun.

salazarm: That would have the extra re-renders issue mentioned before.  Solving it with just React requires Context that defines a custom hook for components to use to subscribe to only the state updates they care about (see this PR for an example: https://github.com/dagster-io/dagster/pull/10239) but I prefer this tbh.

salazarm: If you took this PR: https://github.com/dagster-io/dagster/pull/10239/files

And tried using basic React state instead, you'll get EXTREMELY sluggish and bad performance when hovering elements. And those components are not even complex. It's just the YAML config renderer.

salazarm:
>  Otherwise to dish's point it still takes a while for the dropdown to appear because the initial render is still large?

Also to be clear. I'm not saying that this component in particular is slow. I'm saying that the components being passed in to be rendered could be slow so if/when we do render slow components, I don't want this component to compound the issue.

PR #13391, opened 2023-04-05
Don't include the full set of pipeline tags in all logged events

gibsondan:
Summary:
These tags can sometimes be quite large (for example, k8s job and op config can be serialized json, or a run might have a large step selection) so including them as metadata in every logged event seems inadvisable when they're already available on the run.

In one (perhaps somewhat unsusual edge case, since it has 7.5k steps in the dagster/step_selection tag) this changed the total size of the serialized events in the run debug payload from 1.5GB to 7MB.

Test Plan: BK

### DIFF
--- a/python_modules/dagster/dagster/_core/events/__init__.py
+++ b/python_modules/dagster/dagster/_core/events/__init__.py
@@ -394,7 +394,7 @@ class DagsterEvent(
             step_handle=step_context.step.handle,
             node_handle=step_context.step.node_handle,
             step_kind_value=step_context.step.kind.value,
-            logging_tags=step_context.logging_tags,
+            logging_tags=step_context.event_tags,
             event_specific_data=_validate_event_specific_data(event_type, event_specific_data),
             message=check.opt_str_param(message, "message"),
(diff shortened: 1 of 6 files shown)


prha:
I could see users who've implemented their own logging handlers making use of this metadata (e.g. datadog log ingestion), even though our internal logging handler would want to strip this out of the persistent store.

gibsondan:
on every event though? The data doesn't change between events

prha:
Yeah, I'm mostly just wary of this being a breaking change, in case some OSS users are making use of the tags somehow.  But especially if we're not displaying the tags anywhere, there's no need for us to store them.  We can also strip them out of the log metadata altogether, but may need to think harder about about whether that's the right decision and that should probably be in a `1.3.0` release.

gibsondan:
Addressed @prha 's feedback - all logging tags are still available, but only some are persisted

prha (approved):

prha:
wow, https://github.com/dagster-io/dagster/pull/458/files is very old

PR #13394, opened 2023-04-05
[dagit] Stop populating the dagster/step_selection tag except for selection re-execution

bengotow:
## Summary & Motivation

Currently Dagit launches all asset runs with a `dagster/step_selection` tag. On very large graphs, the tag can be huge because it's a comma separated list of every asset's op name(s).

It appears that this tag is no longer necessary to scope execution in any way -- we pass `selector.assetSelection` as well, and I think the step selection tag predates that.

I believe there is exactly one remaining use case for `dagster/step_selection` in the front-end. If you run assets A,B,C,D and then view the run details page and type "C++" in the step selector and choose "Re-execute Selected", we launch a run with:

```
 assetSelection: <<same as the root run>>
 stepKeys: [C, D]
 tags: {
    dagster/step_selection: "C++"
}
```

In this narrow use case, sending the step selection query text allows us to show you the subset you chose to execute concisely -- we can show "C++" in the UI below instead of "C, D, E, F, G....."

image

## How I Tested These Changes

I updated our test coverage for the materialization button. I also launched a variety of runs and backfills for assets and then re-executed the runs to verify everything works fine.


gibsondan:
amazing, thanks for the fast turnaround here ben - just confirming, in the narrow remaining case you mentioned, there's no risk of the gigantic step selection value happening right?

gibsondan (approved):
given that i see zero usage of this tag on the backend, this looks right to me

bengotow:
> amazing, thanks for the fast turnaround here ben - just confirming, in the narrow remaining case you mentioned, there's no risk of the gigantic step selection value happening right?

I think that's 99% true. Technically someone could individually select every node in the original run and "re-execute selected"... but I doubt it'll ever happen because there's no "Select All" option in the UI. Should be small values only!

bengotow:
Failure is unrelated, merging this!

PR #13395, opened 2023-04-05
[pythonic resources] Error early if a user annotates a param with a ConfigResourceFactory or bare ResourceDef

benpankow:
## Summary

Throws an error early if the user annotates an op, asset, schedule, or sensor function param with a `ConfigurableResourceFactory` or bare `ResourceDefinition` that is not a `ConfigurableResource`, e.g.

```python
class MyFactory(ConfigurableResourceFactory[str]):
    ...


# should be my_resource: Resource[str]!

@op
def my_op(my_resource: MyFactory):
    ...
```

## Test PLan

unit tests

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/resource_annotation.py
+++ b/python_modules/dagster/dagster/_core/definitions/resource_annotation.py
@@ -1,50 +1,10 @@
 from inspect import Parameter
 from typing import Sequence, TypeVar
 
-from typing_extensions import Annotated, get_args
+from typing_extensions import Annotated
 
 from dagster._core.decorator_utils import get_function_params
 from dagster._core.definitions.resource_definition import ResourceDefinition
-
-
-def validate_resource_annotated_function(fn) -> None:
-    """Validates any parameters on the decorated function that are annotated with
-    :py:class:`dagster.ResourceDefinition`, raising a :py:class:`dagster.DagsterInvalidDefinitionError`
-    if any are not also instances of :py:class:`dagster.ConfigurableResource` (these resources should
-    instead be wrapped in the :py:func:`dagster.Resource` Annotation).
-    """
-    from dagster import DagsterInvalidDefinitionError
-    from dagster._config.structured_config import (
-        ConfigurableResource,
-        ConfigurableResourceFactory,
-        TResValue,
-    )
-    from dagster._config.structured_config.utils import safe_is_subclass
-
-    malformed_params = [
-        param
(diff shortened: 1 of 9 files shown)


benpankow:
* master
  * **PR #13376** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13376" target="_blank">Graphite</a> 
    * **PR #13377** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13377" target="_blank">Graphite</a> 
      * **PR #13395** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13395" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13395?utm_source=stack-comment).

schrockn (approved):
cool

schrockn (changes requested):
wait a sec there was an change submitted while i reviewed

schrockn (changes requested):
So I generally think that boolean flags like this should be avoided. This leads future users of `get_resource_args` to ask: should I error aggressively? Why would I not check errors? Etc etc. In this case my interpretation is that you can only setting `err_aggressively=True` near entry points.

Instead I would recommend having a separate function (`validate_resource_annotatable_function` ?)  that does this type of error checking, and do it ASAP (e.g in the relevant `__init__` or immediately after the user passes the `fn` to the framework) and leave `get_resource_args` unmodified  Alternatively a separate function which does this more extensive error checking and then in turn calls `get_resource_args` that is named appropriately.

That you have to do all those dynamic imports is another "smell" that indicates that you might be jamming higher level logic into a lower layer piece of code.

schrockn (approved):

python_modules/dagster/dagster/_core/definitions/resource_annotation.py
```diff
@@ -1,12 +1,52 @@
...
+                param_name=malformed_param.name,
+                annotation_type=malformed_param.annotation,
+                value_message=f"a '{output_type}'" if output_type else "an unknown",
+                annotation_suggestion=f"'Resource[{output_type.__name__}]'"
+                if output_type
+                else "'Resource[Any]' or 'Resource[<output type>]'",
+            )
+        )
+
+
```
schrockn, line 49: consider moving this function to `structured_config` given the all the dynamic imports

schrockn (approved):

PR #13401, opened 2023-04-05
Better handle ecsrunlauncher with use_current_ecs_task_config=False but no task_definition set

gibsondan:
Summary:
We were implicitly assuming that if you set use_current_ecs_task_config to False, you must want to configure the task definition in some way, but that is not always the case. Make it possible to launch a run using a 'default' task definition that pulls nothing from the run.

We actually had a test for this case, but it stil had access to the env var that is set when you are inside the ECS task, so the test passed incorrectly.

Test Plan: BK test that would fail before

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws_tests/ecs_tests/launcher_tests/conftest.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws_tests/ecs_tests/launcher_tests/conftest.py
@@ -174,7 +174,7 @@ def instance_with_resources(instance_cm):
 
 
 @pytest.fixture
-def instance_dont_use_current_task(instance_cm, subnet):
+def instance_dont_use_current_task(instance_cm, subnet, monkeypatch):
     with instance_cm(
         config={
             "use_current_ecs_task_config": False,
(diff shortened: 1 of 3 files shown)


python_modules/libraries/dagster-aws/dagster_aws_tests/ecs_tests/launcher_tests/conftest.py
```diff
@@ -189,6 +189,8 @@ def instance_dont_use_current_task(instance_cm, subnet):
             },
         }
     ) as dagster_instance:
+        # Not running in an ECS task
+        monkeypatch.setenv("ECS_CONTAINER_METADATA_URI_V4", None)
```
jmsanders, line 193: So have we severed the need to launch tasks _from_ ECS. As in you can use the ECS launcher (when properly configured) when otherwise deployed to K8s, Docker, etc.?

gibsondan: That's right - that happened a while ago

jmsanders (approved):

PR #13409, opened 2023-04-06
Add volumes and mountPoints to EcsRunLauncher and EcsContainerContext

gibsondan:
Summary:
Lets you set EFS volumes and other volumes in your ECS tasks.


jmsanders (approved):
Presumably these often (always?) need to be provided together - do we want to do any validation of that relationship or just let the ECS task fail if you provide one but not the other?

gibsondan:
I think particularly with sidecars potentially in the mix its hard to conclusively say that setting volumes without setting mountPaths in the main container is invalid - i'm ok letting ECS do the heavy lifting there

Issue #13098, opened 2023-03-22
[DB IO managers] support when a new column is added to an asset


jamiedemaria:
Right now if you have a partitioned asset and you change the dataframe returned by the asset to have a new column, the IO manager will fail (specifically the snowflake pandas io manager will fail, still need to test the other io managers). This is because, for partitioned assets, we don't fully delete and replace the table for the asset materialization. Instead we drop the table (but retain the column schema) or delete the rows corresponding to the partition and then append the new dataframe. Since the new dataframe has a column that doesn't exist in the snowflake table, we see an error

Ideas of implementation
Two options (potentially more as well)

a) introspect on the dataframe, find the new column, map it to the DB column type, and manually create the column. We'd need to find a way to keep our mapping of python type to database type in sync.

b) copy the full table into memory, append the new partition data with the new column, drop the table, the load the new table into the DB. This would be super memory intensive for large tables, which is likely a deal breaker

Additional information
OG slack thread https://dagster.slack.com/archives/C01U954MEER/p1679514144698129

Message from the maintainers
Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

sryza:
I agree that (b) is a no-go, for the reason you listed.

sryza:
(a) seems promising

jamiedemaria:
This issue also exists with non-partitioned assets. Replication case

```
@asset
def a_plain_table() -> pd.DataFrame:
    return pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6],})
```
and materialize. Then update asset def to 
```
@asset
def a_plain_table() -> pd.DataFrame:
    return pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6], "c": [7, 8, 9]})
```
and materialization will fail

DanMeyer:
I've been doing (b) manually. I have smaller tables than most (the largest one has 2-3 million records), so this hasn't been a problem performance-wise. However, I have a few Snowflake roles (some for user-facing apps) that only have access to a few specific tables (so I can't just grant select on all future tables in a schema), and I lose those permission grants when I drop the tables.

My best idea so far is include a statement granting those permissions as part of any such assets.

malow106:
Hi, any plan for this issue ? Thx !

geoHeil:
What about a not fully automatic mode but a configuration-driven approach where one can decide what should happen on a per-asset (metadata) level?

milesgranger:
[...]

PR #13410, opened 2023-04-06
Improve performance of table slice cleanup in dagster-snowflake

mjclarke94:
## Summary & Motivation

`DELETE FROM` is a particularly expensive command to run in snowflake, especially when there is no partitioning key applied to the table. When rematerialising an asset, an initial cleanup command is run, regardless of whether or not there are any impacted rows, which is fairly time consuming.

These changes improve performance here in two ways:
1) For unpartitioned assets, the `TRUNCATE` command is used for the cleanup statement, which has the same overall effect but is far cheaper to run (even on empty tables). This also does not require a snowflake warehouse to run which is a bonus.

2) For partitioned assets, a `COUNT(*)` type command is first run to ensure that the delete command is necessary (i.e. that there are some rows affected), and only executes if needed. There is a slight trade off here in that if the delete is necessary, you've added some small overhead, but IMO it is a worthwhile tradeoff.

## How I Tested These Changes
Tested locally on a partitioned and unpartitioned table. Ran pytest.

### DIFF
--- a/python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
+++ b/python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
@@ -284,7 +284,12 @@ class SnowflakeDbClient(DbClient):
     @staticmethod
     def delete_table_slice(context: OutputContext, table_slice: TableSlice, connection) -> None:
         try:
-            connection.execute(_get_cleanup_statement(table_slice))
+            if table_slice.partition_dimensions and len(table_slice.partition_dimensions) > 0:
+                count = connection.execute(_get_count_statement(table_slice))
+                if count.fetchall()[0][0] > 0:
+                    connection.execute(_get_cleanup_statement(table_slice))
+            else:
+                connection.execute(_get_cleanup_statement(table_slice))
         except ProgrammingError:
             # table doesn't exist yet, so ignore the error
             pass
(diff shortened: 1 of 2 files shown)


python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
```diff
@@ -312,7 +317,21 @@ def _get_cleanup_statement(table_slice: TableSlice) -> str:
         )
         return query + _partition_where_clause(table_slice.partition_dimensions)
     else:
-        return f"DELETE FROM {table_slice.database}.{table_slice.schema}.{table_slice.table}"
+        return f"TRUNCATE {table_slice.database}.{table_slice.schema}.{table_slice.table}"
```
jamiedemaria, line 320: I don't think we can support a TRUNCATE call here right now. There would be an issue if a user wrote an asset, materialized it, then changed the code to add a new column. When we re-materialize the asset, if we just truncate the table, the underlying snowflake table will still have the columns of the original materialization. When we try to load the new dataframe with the new column, we will get an error because the new dataframe column doesn't have a corresponding column in the snowflake table. 

Unfortunately, the library we are using to load dataframes (pandas.to_sql with the snowflake pandas writer) doesn't handle adding new columns. We have an issue open [here](https://github.com/dagster-io/dagster/issues/13098) to support adding new columns for partitions (where we have the same issue). Once we implement this, then we could switch to using TRUNCATE here as well

mjclarke94: My understanding is that `DELETE FROM` also preserves column names/schema?

jamiedemaria: ah yeah, you are totally right,  i had bad info. Tested it and it does keep the columns. truncate seems like it should be good then. 

The issue with adding a new column to the dataframe in an asset exists in the current dagster version, so i'll update the gh issue. I'll give this another review pass, but it generally looks good

cmpadden:

PR #13413, opened 2023-04-06
Remove parentPipelineSnapshotId field from runs queries

gibsondan:
Summary:
This field requires fetching the pipeline snapshot from storage for each run, which can be expensive if the job has many assets.

The one place it was being used appears to be to detect a case that seems pretty edge-casey to me and not worth slowing down the whole UI - the pipeline snapshot but the repository origin not matching (maybe after doing a repo rename?) Curious what reviewers think about just taking that out.

Test Plan: Load runs page on a deployment with large pipeline snapshots - runs page now loads in under a second instead of 5-15 seconds

### DIFF
--- a/js_modules/dagit/packages/core/src/instance/types/JobMenu.types.ts
+++ b/js_modules/dagit/packages/core/src/instance/types/JobMenu.types.ts
@@ -26,9 +26,9 @@ export type RunReExecutionQuery = {
         pipelineName: string;
         solidSelection: Array<string> | null;
         pipelineSnapshotId: string | null;
-        parentPipelineSnapshotId: string | null;
         stepKeysToExecute: Array<string> | null;
         updateTime: number | null;
+        parentPipelineSnapshotId: string | null;
         startTime: number | null;
         endTime: number | null;
         tags: Array<{__typename: 'PipelineTag'; key: string; value: string}>;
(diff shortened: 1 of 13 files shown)


gibsondan:
oh wait hm, this is used here: https://github.com/dagster-io/dagster/blob/master/js_modules/dagit/packages/core/src/workspace/useRepositoryForRun.ts#L51-L54

Might have to get a bit trickier

js_modules/dagit/packages/core/src/workspace/useRepositoryForRun.ts
```diff
@@ -7,7 +7,7 @@ import {findRepoContainingPipeline, repoContainsPipeline} from './findRepoContai
 
 type MatchType = {
   match: DagsterRepoOption;
-  type: 'origin-and-snapshot' | 'origin-only' | 'snapshot-only' | 'pipeline-name-only';
+  type: 'origin-only' | 'pipeline-name-only';
```
hellendag, line 10: Change this `'origin' | 'pipeline-name-only'`? Since the implication is now that `origin` means that it's a "complete" match, including the job name.

js_modules/dagit/packages/core/src/runs/RunActionButtons.tsx
```diff
@@ -286,35 +286,6 @@ function usePipelineAvailabilityErrorForRun(
       };
     }
 
-    if (matchType === 'snapshot-only') {
```
hellendag, line 289: The `matchType === 'origin-and-snapshot'` "best case" should now be used for the `'origin'` match type.

hellendag (approved):

gibsondan:
Closing in favor of https://github.com/dagster-io/dagster/pull/13431 which does the same thing, but better

PR #13415, opened 2023-04-06
[ui] Clear old Launchpad sessions when localStorage quota is reached

hellendag:
## Summary & Motivation

Make an attempt to clean up old launchpad sessions when a user encounters a localStorage write error, since this most likely indicates that the localStorage quota has been reached. Sort the existing stored sessions by timestamp, and delete them until the write succeeds.

## How I Tested These Changes

Jest.

Load launchpad, verify that sessions are correctly written to localStorage (and updated correctly) in the normal case.

### DIFF
--- a/js_modules/dagit/packages/core/src/app/ExecutionSessionStorage.tsx
+++ b/js_modules/dagit/packages/core/src/app/ExecutionSessionStorage.tsx
@@ -1,3 +1,4 @@
+import memoize from 'lodash/memoize';
 import * as React from 'react';
 
 import {AssetKeyInput} from '../graphql/types';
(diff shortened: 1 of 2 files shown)


hellendag:
* master
  * **PR #13415** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13415" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13415?utm_source=stack-comment).

js_modules/dagit/packages/core/src/app/ExecutionSessionStorage.tsx
```diff
@@ -160,23 +156,11 @@ export function useExecutionSessionStorage(
     () => buildValidator(JSON.parse(initialAsJSON) as Partial<IExecutionSession>),
     [initialAsJSON],
   );
-  const [newState, setNewState] = useStateWithStorage(key, validator);
-
-  // Read stored value for the old key. If a value is present, we will use it
-  // for the new key, then remove the old key entirely.
-  const oldValidator = React.useCallback((json: any) => json, []);
-  const [oldState, setOldState] = useStateWithStorage(oldKey, oldValidator);
```
hellendag, line 168: Also cleaning this up, since it's about a year old at this point.

salazarm (approved):
It would be nice if this happened on any localStorage write error (that we make room to accommodate it). Although if we do reach quota this is likely the cause since its storing the biggest objects atm?

hellendag:
> It would be nice if this happened on any localStorage write error (that we make room to accommodate it). Although if we do reach quota this is likely the cause since its storing the biggest objects atm?

Yeah, that's the idea. I don't think we store anything nearly as big -- these configs can be *very* big and there can be tons of them.

PR #13419, opened 2023-04-06
[1/n][dagster-aws] Begin migrating dagster-aws to new resource system

benpankow:
## Summary

Migrates Athena, ECR, & Redshift resources to Pythonic resource system.

## Test Plan

New unit tests

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws/athena/resources.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/athena/resources.py
@@ -14,7 +14,7 @@ from dagster import (
     resource,
 )
 from dagster._core.execution.context.init import InitResourceContext
-from pydantic import Field as PyField
+from pydantic import Field
 
 
 class AthenaError(Exception):
@@ -199,33 +199,33 @@ class FakeAthenaClient(AthenaClient):
 
 
 class BaseAthenaResource(ConfigurableResourceFactory[AthenaClient]):
-    workgroup: str = PyField(
-        "primary",
+    workgroup: str = Field(
+        default="primary",
         description=(
             "The Athena WorkGroup to use."
             " https://docs.aws.amazon.com/athena/latest/ug/manage-queries-control-costs-with-workgroups.html"
         ),
     )
-    polling_interval: int = PyField(
-        5,
+    polling_interval: int = Field(
+        default=5,
         description=(
             "Time in seconds between checks to see if a query execution is finished. 5 seconds"
             " by default. Must be non-negative."
         ),
     )
-    max_polls: int = PyField(
-        120,
+    max_polls: int = Field(
+        default=120,
         description=(
             "Number of times to poll before timing out. 120 attempts by default. When coupled"
             " with the default polling_interval, queries will timeout after 10 minutes (120 * 5"
             " seconds). Must be greater than 0."
         ),
     )
-    aws_access_key_id: Optional[str] = PyField(
-        None, description="AWS access key ID for authentication purposes."
+    aws_access_key_id: Optional[str] = Field(
+        default=None, description="AWS access key ID for authentication purposes."
     )
-    aws_secret_access_key: Optional[str] = PyField(
-        None, description="AWS secret access key for authentication purposes."
+    aws_secret_access_key: Optional[str] = Field(
+        default=None, description="AWS secret access key for authentication purposes."
     )
 
 
--- a/python_modules/libraries/dagster-aws/dagster_aws/athena/resources.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/athena/resources.py
@@ -3,16 +3,18 @@ import io
 import os
 import time
 import uuid
+from typing import Optional
 from urllib.parse import urlparse
 
 import boto3
 from botocore.stub import Stubber
 from dagster import (
-    Field,
-    StringSource,
+    ConfigurableResourceFactory,
     _check as check,
     resource,
 )
+from dagster._core.execution.context.init import InitResourceContext
+from pydantic import Field as PyField
 
 
 class AthenaError(Exception):
@@ -23,7 +25,7 @@ class AthenaTimeout(AthenaError):
     pass
 
 
-class AthenaResource:
+class AthenaClient:
     def __init__(self, client, workgroup="primary", polling_interval=5, max_polls=120):
         check.invariant(
             polling_interval >= 0, "polling_interval must be greater than or equal to 0"
@@ -97,7 +99,7 @@ class AthenaResource:
         return results
 
 
-class FakeAthenaResource(AthenaResource):
+class FakeAthenaClient(AthenaClient):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.polling_interval = 0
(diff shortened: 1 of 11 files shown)


python_modules/libraries/dagster-aws/dagster_aws/athena/resources.py
```diff
@@ -23,7 +25,7 @@ class AthenaTimeout(AthenaError):
     pass
 
 
-class AthenaResource:
+class AthenaClient:
```
schrockn, line 28: Hmmm is this not a breaking change? `AthenaResource` was public and now methods like `execute_query` don't exist

python_modules/libraries/dagster-aws/dagster_aws/redshift/resources.py
```diff
@@ -272,67 +275,69 @@ def execute_queries(
...
-            description="Connection timeout in seconds. 5 seconds by default",
-            is_required=False,
-            default_value=5,
+
+    host: str = PyField(..., description="Redshift host")
+    port: int = PyField(5439, description="Redshift port")
+    user: Optional[str] = PyField(None, description="Username for Redshift connection")
+    password: Optional[str] = PyField(None, description="Password for Redshift connection")
+    database: Optional[str] = PyField(
+        None,
```
schrockn, line 309: just say no to PyField

python_modules/libraries/dagster-aws/dagster_aws/redshift/resources.py
```diff
@@ -272,67 +275,69 @@ def execute_queries(
...
-        "sslmode": Field(
-            str,
-            description=(
-                "SSL mode to use. See the Redshift documentation for more information on usage:"
-                " https://docs.aws.amazon.com/redshift/latest/mgmt/connecting-ssl-support.html"
-            ),
-            is_required=False,
-            default_value="require",
+    )
+    autocommit: Optional[bool] = PyField(None, description="Whether to autocommit queries")
```
schrockn, line 315: default=None

python_modules/libraries/dagster-aws/dagster_aws/redshift/resources.py
```diff
@@ -272,67 +275,69 @@ def execute_queries(
...
-    }
+    )
+
+    def create_resource(self, context: InitResourceContext) -> RedshiftClient:
+        return RedshiftClient(context)
+
+
+class FakeRedshiftResource(RedshiftResource):
+    def create_resource(self, context: InitResourceContext) -> FakeRedshiftClient:
+        return FakeRedshiftClient(context)
```
schrockn, line 333: IMO this should not inherit from `RedshiftResource` having config on the fake doesn't make too much sense

python_modules/libraries/dagster-aws/dagster_aws/redshift/resources.py
```diff
@@ -1,23 +1,26 @@
...
 
 
 class RedshiftError(Exception):
     pass
 
 
-class _BaseRedshiftResource(abc.ABC):
-    def __init__(self, context):
+class BaseRedshiftClient(abc.ABC):
+    def __init__(self, context: InitResourceContext):
```
schrockn, line 23: So this predates you, but I don't think it makes sense for the Fake to inherit from a base where the constructor does real work like this. IMO we should heavily bias towards pure-interface-style ABCs.

schrockn (changes requested):
I think this is an important PR as it will serve as a showcase for how to structure and name testable resources. It's important to dig in and get the details right and then we can document our norms. 

Also we need to make this non-breaking.

I made some comments but might be worth digging in in a call.

python_modules/libraries/dagster-aws/dagster_aws/athena/resources.py
benpankow: I think this is worth discussing in a call  - I'm not sure what the best path forward is for cases like this to preserve compat while adding a factory wrapper.

python_modules/libraries/dagster-aws/dagster_aws/athena/resources.py
```diff
@@ -97,7 +100,12 @@ def _results(self, execution_id):
         return results
 
 
-class FakeAthenaResource(AthenaResource):
+@deprecated
+class AthenaResource(AthenaClient):
+    pass
```
schrockn, line 105: comment-worthy

python_modules/libraries/dagster-aws/dagster_aws/athena/resources.py
```diff
@@ -196,47 +204,89 @@ def _fake_results(self, execution_id, expected_results):
...
+
+
+class FakeAthenaClientResource(ResourceWithAthenaConfig):
+    def provide_object_for_execution(self, context: InitResourceContext) -> FakeAthenaClient:
+        return FakeAthenaClient(
+            client=boto3.client("athena", region_name="us-east-1"),
+            workgroup=self.workgroup,
+            polling_interval=self.polling_interval,
+            max_polls=self.max_polls,
+        )
```
schrockn, line 282: This predates you, but I don't see how this is a fake. This is an Athena client hardcoded to talk to us-east-1

python_modules/libraries/dagster-aws/dagster_aws/redshift/resources.py
```diff
@@ -203,9 +207,33 @@ def _get_cursor(self, conn, cursor_factory=None):
...
+                "database",
+                "connect_timeout",
+                "sslmode",
+            )
+            if context.resource_config.get(k) is not None
+        }
+
+        self.autocommit = context.resource_config.get("autocommit")
+        self.log = context.log
+
```
schrockn, line 236: Yeah I didn't mean to move you to do a big copy paste job.

I think the core issue here is that the stuff you are porting aren't really "fakes".

python_modules/libraries/dagster-aws/dagster_aws/redshift/resources.py
```diff
@@ -272,67 +300,74 @@ def execute_queries(
...
+        .. code-block:: python
+
+            from dagster import build_op_context, op
+            from dagster_aws.redshift import RedshiftResource
+
+            @op(required_resource_keys={'redshift'})
+            def example_redshift_op(context):
+                return context.resources.redshift.execute_query('SELECT 1', fetch_results=True)
+
+            redshift_configured = RedshiftResource(
```
schrockn, line 322: update docblock with new class name

python_modules/libraries/dagster-aws/dagster_aws/redshift/resources.py
```diff
@@ -272,67 +300,74 @@ def execute_queries(
...
 
 
-def define_redshift_config():
-    """Redshift configuration.
+@deprecated
+class FakeRedshiftResource(FakeRedshiftClient):
+    pass
+
+
+class RedshiftClientResource(FactoryResource[RedshiftClient]):
```
schrockn, line 308: this naming feels a lot better 👍🏻

python_modules/libraries/dagster-aws/dagster_aws/redshift/resources.py
schrockn: I believe you can just delete `self.conn_args` and `self.autocommit`?

schrockn (changes requested):
* Comment all the deprecated classes and point them to what should be used in the comments
* Remove the copy-pasta between fake redshift client and the redshift client. I think you can just delete stuff
* Make sure to update docblocks with new names

python_modules/libraries/dagster-aws/dagster_aws_tests/ecr_tests/test_resources.py
```diff
@@ -14,3 +14,16 @@ def ecr_public_solid(context):
     )
 
     assert result.output_value() == "token"
+
+
+def test_ecr_public_get_login_password_pythonic() -> None:
+    @op
+    def ecr_public_solid(ecr_public: FakeECRPublicResource):
```
schrockn, line 21: ecr_public_op

schrockn (commented):
k this feels reasonable

python_modules/libraries/dagster-aws/dagster_aws/athena/resources.py
benpankow: Changed name to `TestHardcodedAthenaClientResource` & moved to test files for now. Not sure why this lived in the public package.

benpankow: It was Fake but only if you tuned on `moto` like we did in the tests, meaning it's not very useful out of the box for users.

python_modules/libraries/dagster-aws/dagster_aws/redshift/resources.py
```diff
@@ -47,7 +33,14 @@ def execute_queries(
         pass
 
 
-class RedshiftResource(_BaseRedshiftResource):
+class RedshiftClient(BaseRedshiftClient):
+    def __init__(self, conn_args: Dict[str, Any], autocommit: Optional[bool], log: Logger):
```
benpankow, line 37: this change is breaking because the init signature changes - if we think that users are constructing `RedshiftClient` themselves I can hack around this to keep the `context`-requiring signature, but it's pretty painful to build a context when we don't have one

schrockn: we don't export it. I think it is fine.

python_modules/libraries/dagster-aws/dagster_aws/athena/resources.py
```diff
@@ -196,47 +206,82 @@ def _fake_results(self, execution_id, expected_results):
...
+
+                from dagster import build_op_context, op
+                from dagster_aws.athena import AthenaClientResource
+
+                @op
+                def example_athena_op(athena: AthenaClientResource):
+                    return athena.create_athena_client().execute_query("SELECT 1", fetch_results=True)
+
+                context = build_op_context(resources={"athena": AthenaClientResource()})
+                assert example_athena_op(context) == [("1",)]
```
schrockn, line 261: docblock example should directly pass resource instead of context

python_modules/libraries/dagster-aws/dagster_aws/redshift/resources.py
```diff
@@ -203,9 +196,19 @@ def _get_cursor(self, conn, cursor_factory=None):
                 conn.commit()
 
 
-class FakeRedshiftResource(_BaseRedshiftResource):
+@deprecated
+class RedshiftResource(RedshiftClient):
+    pass
```
schrockn, line 201: comment as to deprecation reason

python_modules/libraries/dagster-aws/dagster_aws/redshift/resources.py
```diff
@@ -272,67 +275,88 @@ def execute_queries(
             return [self.QUERY_RESULT] * 3
 
 
-def define_redshift_config():
-    """Redshift configuration.
+@deprecated
+class FakeRedshiftResource(FakeRedshiftClient):
```
schrockn, line 279: comment re: deprecation reason

python_modules/libraries/dagster-aws/dagster_aws/redshift/resources.py
```diff
@@ -272,67 +275,88 @@ def execute_queries(
...
+
+            redshift_configured = RedshiftClientResource(
+                host='my-redshift-cluster.us-east-1.redshift.amazonaws.com',
+                port=5439,
+                user='dagster',
+                password='dagster',
+                database='dev',
+            )
+            context = build_op_context(resources={'redshift': redshift_configured})
+            assert example_redshift_op(context) == [(1,)]
```
schrockn, line 306: directly pass resource rather than context

python_modules/libraries/dagster-aws/dagster_aws_tests/athena_tests/test_resources.py
```diff
@@ -45,29 +61,39 @@ def test_execute_query_state_transitions(mock_athena_client, expected_states):
...
+
+def test_op_pythonic_resource(mock_athena_client) -> None:
+    from dagster import build_op_context, op
+
+    @op
+    def example_athena_op(athena: TestAthenaClientResource):
+        return athena.create_athena_client().execute_query("SELECT 1", fetch_results=True)
+
+    context = build_op_context(resources={"athena": TestAthenaClientResource.configure_at_launch()})
+    assert example_athena_op(context) == [("1",)]
```
schrockn, line 99: Let's prefer direct resource passing

python_modules/libraries/dagster-aws/dagster_aws_tests/ecr_tests/test_resources.py
```diff
@@ -1,16 +1,29 @@
...
+
+
+def test_ecr_public_get_login_password_pythonic() -> None:
+    @op
+    def ecr_public_op(ecr_public: FakeECRPublicResource):
+        return ecr_public.create_ecr_client().get_login_password()
+
+    result = wrap_op_in_graph_and_execute(
+        ecr_public_op,
+        resources={"ecr_public": FakeECRPublicResource.configure_at_launch()},
```
schrockn, line 26: I think we can just construct an empty once rather than use `configure_at_launch` no?

schrockn (approved):
good stuff. please heed final comments

python_modules/libraries/dagster-aws/dagster_aws/redshift/resources.py
```diff
@@ -203,9 +196,21 @@ def _get_cursor(self, conn, cursor_factory=None):
...
+    """Deprecated. This class was used by the function-style Redshift resource. New code should instead
+    use the Pythonic, class-style RedshiftClientResource.
+    """
+
+
+class FakeRedshiftClient(BaseRedshiftClient):
     QUERY_RESULT = [(1,)]
 
+    def __init__(self, log: Logger):
+        # Extract parameters from resource config
```
jamiedemaria, line 210: nit: remove comment

python_modules/libraries/dagster-aws/dagster_aws/redshift/resources.py
```diff
@@ -272,67 +277,93 @@ def execute_queries(
...
+    """This resource enables connecting to a Redshift cluster and issuing queries against that
+    cluster.
+
+    Example:
+        .. code-block:: python
+
+            from dagster import Definitions, asset, EnvVar
+            from dagster_aws.redshift import RedshiftClientResource
+
+            @asset(required_resource_keys={'redshift'})
```
jamiedemaria, line 297: example should use pythonic resources syntax

PR #13425, opened 2023-04-06
[dagster_gcp] BigQuery resource upgraded to pydantic config

jamiedemaria:
## Summary & Motivation
Updates the BigQuery resource to use pydantic config.

One main issue with this resource is handling authentication via config values. In order to support serverless users, we need to allow them to pass GCP creds as configuration. In the "old" version i could setup credentials on resource init, but that doesn't work in the pydantic resource. Instead I've made a context manager that the users can call to set up the credentials (if there is no credential config passed, the context manager does nothing. So if in local dev they have GOOGLE_APPLICATION_CREDENTIALS set, the asset will still work). See the test file for how this will look in an asset. IMO it's a fine pattern, but not great. However, I'm not sure what our other options are. if you have ideas, let me know!

I believe this is a pattern we will need to include for all GCP resources (GCS, Dataproc) so we should make sure we like it.

### DIFF
--- a/examples/docs_snippets/docs_snippets/integrations/bigquery/resource.py
+++ b/examples/docs_snippets/docs_snippets/integrations/bigquery/resource.py
@@ -9,15 +9,12 @@ from dagster import Definitions, asset
 @asset
 def small_petals(bigquery: BigQueryResource):
     with bigquery.get_client() as client:
-        return (
-            client.query(
-                (
-                    'SELECT * FROM IRIS.IRIS_DATA WHERE "petal_length_cm" < 1 AND'
-                    ' "petal_width_cm" < 1'
-                ),
-            )
-            .result()
-        )
+        return client.query(
+            (
+                'SELECT * FROM IRIS.IRIS_DATA WHERE "petal_length_cm" < 1 AND'
+                ' "petal_width_cm" < 1'
+            ),
+        ).result()
 
 
 defs = Definitions(
--- a/examples/docs_snippets/docs_snippets/integrations/bigquery/resource.py
+++ b/examples/docs_snippets/docs_snippets/integrations/bigquery/resource.py
@@ -6,7 +6,7 @@ from dagster import Definitions, asset
 # Using Dagster with BigQuery tutorial
 
 
-@asset()
+@asset
 def small_petals(bigquery: BigQueryResource):
     with bigquery.get_client() as client:
         return (
--- a/examples/docs_snippets/docs_snippets/integrations/bigquery/resource.py
+++ b/examples/docs_snippets/docs_snippets/integrations/bigquery/resource.py
@@ -1,4 +1,4 @@
-from dagster_gcp import bigquery_resource
+from dagster_gcp import BigQueryResource
 
 from dagster import Definitions, asset
 
@@ -6,24 +6,26 @@ from dagster import Definitions, asset
 # Using Dagster with BigQuery tutorial
 
 
-@asset(required_resource_keys={"bigquery"})
-def small_petals(context):
-    return context.resources.bigquery.query(
-        (
-            'SELECT * FROM IRIS.IRIS_DATA WHERE "petal_length_cm" < 1 AND'
-            ' "petal_width_cm" < 1'
-        ),
-    ).result()
+@asset()
+def small_petals(bigquery: BigQueryResource):
+    with bigquery.get_client() as client:
+        return (
+            client.query(
+                (
+                    'SELECT * FROM IRIS.IRIS_DATA WHERE "petal_length_cm" < 1 AND'
+                    ' "petal_width_cm" < 1'
+                ),
+            )
+            .result()
+        )
 
 
 defs = Definitions(
     assets=[small_petals],
     resources={
-        "bigquery": bigquery_resource.configured(
-            {
-                "project": "my-gcp-project",
-                "location": "us-east5",
-            }
+        "bigquery": BigQueryResource(
+            project="my-gcp-project",
+            location="us-east5",
         )
     },
 )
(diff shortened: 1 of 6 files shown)


jamiedemaria:
* master
  * **PR #13425** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13425" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13425?utm_source=stack-comment).

examples/docs_snippets/docs_snippets/integrations/bigquery/resource.py
```diff
@@ -1,29 +1,31 @@
...
 
-@asset(required_resource_keys={"bigquery"})
-def small_petals(context):
-    return context.resources.bigquery.query(
-        (
-            'SELECT * FROM IRIS.IRIS_DATA WHERE "Petal length (cm)" < 1 AND "Petal'
-            ' width (cm)" < 1'
-        ),
-    ).result()
+@asset()
```
schrockn, line 9: nit: rm `()`

python_modules/libraries/dagster-gcp/dagster_gcp/bigquery/resources.py
```diff
@@ -1,20 +1,54 @@
...
+    def setup_credentials(self):
+        if self.gcp_credentials:
+            with setup_gcp_creds(self.gcp_credentials):
+                yield
+        else:
+            yield
+
+    def get_client(self) -> bigquery.Client:
+        return bigquery.Client(project=self.project, location=self.location)
+
```
schrockn, line 45: Stepping back for a second, why do we need this pattern in the first place? It seems pretty contorted to have to provided a base64-encoded key and write it out to a temp file like this.

schrockn: Assuming we need this, IMO we should have the `get_client` be a context manager. Like so:

```
with bq_resource.client() as client:
   client.whatever()
```

Prevents class of programming errors as right now you could construct a client in a way that will break in non-obvious ways if you forgot to call `setup_credentials`

schrockn (changes requested):
See inline comments

python_modules/libraries/dagster-gcp/dagster_gcp/bigquery/resources.py
jamiedemaria: mainly it's to support serverless. to auth with gcp you either need to run `gcloud auth login` or set an env var `GOOGLE_APPLICATION_CREDENTIALS=creds_file.json` where `creds_file.json` has your auth key. For serverless, users can't upload a creds file, or run the interactive `gcloud auth login` command so we do this sort of work around to allow them to pass the raw credentials string and then we write it to a file and set the env var. 

The base64 encoding is to solve an issue where newlines in the creds file will cause authentication to fail. by base64 encoding it, we dont have to worry about newlines

as for having `get_client` be a context manager - are you thinking like this instead?
```
@contextmanager
    def get_client(self):
        if self.gcp_credentials:
            with setup_gcp_creds(self.gcp_credentials):
                yield bigquery.Client(project=self.project, location=self.location)
        else:
            yield bigquery.Client(project=self.project, location=self.location)

@asset 
def my_asset(bigquery: BigQueryResource):
    with bigquery.get_client() as client:
        client.query(....)
```

schrockn: Yeah that's what I'm thinking. It unfortunately imposes some overhead on people who wouldn't have any use for the `gcp_credentials` feature with a strictly unnecessary `with` block.

I think having a `setup_credentials` is a bit odd since if you have the env var set it does work. `setup_credentials` sounds like something you are required to call in all circumstances.

jamiedemaria: yeah i think moving to `get_client` being the context manager makes sense. I was playing around with a duckdb resource and i think the `get_client` method there will also need to be a context manager so we have handle closing the connection. This way everything is consistent with `get_client` being a context manager for some resources

schrockn (dismissed):
Cool sounds good. Sending back to your q.

python_modules/libraries/dagster-gcp/dagster_gcp/bigquery/resources.py
```diff
@@ -1,20 +1,86 @@
...
+from dagster import ConfigurableResource, resource
 from google.cloud import bigquery
+from pydantic import Field
 
-from .configs import bq_resource_config
 from .utils import setup_gcp_creds
 
 
+class BigQueryResource(ConfigurableResource):
```
jamiedemaria, line 11: TODO - implement IAttach... interface

python_modules/libraries/dagster-gcp/dagster_gcp/bigquery/resources.py
```diff
@@ -1,20 +1,90 @@
...
+                    "bigquery": BigQueryResource(project="my-project")
+                }
+            )
+    """
+
+    project: Optional[str] = Field(
+        default=None,
+        description="""Project ID for the project which the client acts on behalf of. Will be passed
+        when creating a dataset / job. If not passed, falls back to the default inferred from the
+        environment.""",
```
benpankow, line 37: nit:  I think using the triple quote string will add the indents after the newlines here

```suggestion
        default=None,
        description=(
            "Project ID for the project which the client acts on behalf of. Will be passed"
            " when creating a dataset / job. If not passed, falls back to the default inferred from the"
            " environment."
        ),
```

benpankow (approved):

PR #13430, opened 2023-04-06
[pythonic config][fix] Fix conversion of optional nested Config classes to schema

benpankow:
## Summary

The `if` clause which deals with optional Pydantic fields was only triggering for non-`Config`-extending fields. Fixes this behavior and puts it under test.

## Test Plan

New unit test.

### DIFF
--- a/python_modules/dagster/dagster/_config/structured_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/structured_config/__init__.py
@@ -911,8 +911,14 @@ def _convert_pydantic_field(pydantic_field: ModelField, model_cls: Optional[Type
             config_type=inferred_field.config_type,
             key_type=key_type,
         )
+        if pydantic_field.allow_none:
+            wrapped_config_type = Noneable(wrapped_config_type)
 
-        return Field(config=wrapped_config_type, description=inferred_field.description)
+        return Field(
+            config=wrapped_config_type,
+            description=inferred_field.description,
+            is_required=_is_pydantic_field_required(pydantic_field),
+        )
     else:
         # For certain data structure types, we need to grab the inner Pydantic field (e.g. List type)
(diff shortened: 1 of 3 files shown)


benpankow:
* master
  * **PR #13430** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13430" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13430?utm_source=stack-comment).

python_modules/dagster/dagster/_config/structured_config/__init__.py
```diff
@@ -900,8 +900,14 @@ def _convert_pydantic_field(pydantic_field: ModelField) -> Field:
             config_type=inferred_field.config_type,
             key_type=key_type,
         )
+        if pydantic_field.allow_none:
+            wrapped_config_type = Noneable(wrapped_config_type)
 
-        return Field(config=wrapped_config_type, description=inferred_field.description)
+        return Field(
+            config=wrapped_config_type,
```
schrockn, line 907: minor nit: if think we convenient we should avoid reassignment (oh to have `const` in Python) so doing this as one-liner would be nice

python_modules/dagster/dagster_tests/core_tests/structured_config_tests/test_structured_config_types.py
```diff
@@ -310,6 +310,48 @@ def a_job():
...
+        {
+            "ops": {
+                "a_struct_config_op": {
+                    "config": {"an_optional_nested": [{"a_str": "foo"}, {"a_str": "bar"}]}
+                }
+            }
+        }
+    ).success
+    assert executed["yes"]
+    executed.clear()
```
schrockn, line 343: `executed = {}` might be more obvious

schrockn (approved):
👍🏻

Issue #13382, opened 2023-04-05
Asset reconciliation sensor is triggering all partition of MultiPartitionsDefinition asset instead on only the latest


mgierada:
### What's the use case?

My asset uses  `MultiPartitionsDefinition` partitions. I noticed that **all** of them are triggered by `asset_reconciliation_sensor` My understanding was it should behave similarly to `DailyPartitionsDefinition` so only the latest partition triggered. Is it a bug or a feature? Is there a way to setup `asset_reconciliation_sensor` to trigger only latest partitions of `MultiPartitionsDefinition` asset?

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

sryza:
As mentioned on Slack, @clairelin135 is going to look into this when she has the chance

PR #13432, opened 2023-04-06
[asset-reconciliation] Limit time-dimension multipartitions that can be reconciled with the sensor

clairelin135:
Unifies asset reconciliation behavior for multipartitioned and time-partitioned assets, limiting the multipartitions that can be reconciled to the same time window range (1 day).

For multipartitioned assets that do not contain a time dimension, follows existing behavior and reconciles all partitions.

Resolves https://github.com/dagster-io/dagster/issues/13382


mgierada (commented):
LGTM 🚀  I waited for that! Thanks for a quick pick up.

sryza (approved):
Nice

sryza:
Sorry it took me a while to get to this

PR #13421, opened 2023-04-06
[2/n][dagster-aws] Migrate SecretsManager, SSM resources to new APIs

benpankow:
## Summary

Migrates SecretsManager & SSM resources to Pythonic resource system.

## Test Plan

Existing + new unit tests

PR #13433, opened 2023-04-06
[3/n][dagster-aws] Port S3Resource, S3IOManager to Pythonic resources

benpankow:
## Summary

Finishes off the `dagster-aws` migration by migrating the `S3IOManager` and `S3Resource` over.

## Test Plan

Add new resources to existing unit tests.

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
@@ -157,18 +157,18 @@ class ConfigurablePickledObjectS3IOManager(ConfigurableIOManager):
     )
 
     @cached_method
-    def inner_db_io_manager(self) -> PickledObjectS3IOManager:
+    def inner_io_manager(self) -> PickledObjectS3IOManager:
         return PickledObjectS3IOManager(
             s3_bucket=self.s3_bucket,
-            s3_session=self.s3_resource,
+            s3_session=self.s3_resource.create_client(),
             s3_prefix=self.s3_prefix,
         )
 
     def load_input(self, context: InputContext) -> Any:
-        return self.inner_db_io_manager().load_input(context)
+        return self.inner_io_manager().load_input(context)
 
     def handle_output(self, context: OutputContext, obj: Any) -> None:
-        return self.inner_db_io_manager().handle_output(context, obj)
+        return self.inner_io_manager().handle_output(context, obj)
 
 
 @io_manager(
--- a/python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
@@ -104,7 +104,7 @@ class PickledObjectS3IOManager(MemoizableIOManager):
         context.add_output_metadata({"uri": MetadataValue.path(path)})
 
 
-class S3IOManagerResource(ConfigurableIOManager):
+class PickledObjectS3IOManagerResource(ConfigurableIOManager):
     """Persistent IO manager using S3 for storage.
 
     Serializes objects via pickling. Suitable for objects storage for distributed executors, so long
@@ -125,7 +125,7 @@ class S3IOManagerResource(ConfigurableIOManager):
     .. code-block:: python
 
         from dagster import asset, Definitions
-        from dagster_aws.s3 import S3PickleIOManager, S3Resource
+        from dagster_aws.s3 import PickledObjectS3IOManagerResource, S3Resource
 
 
         @asset
@@ -140,7 +140,7 @@ class S3IOManagerResource(ConfigurableIOManager):
         defs = Definitions(
             assets=[asset1, asset2],
             resources={
-                "io_manager": S3PickleIOManager(
+                "io_manager": PickledObjectS3IOManagerResource(
                     s3_resource=S3Resource(),
                     s3_bucket="my-cool-bucket",
                     s3_prefix="my-cool-prefix",
@@ -172,7 +172,7 @@ class S3IOManagerResource(ConfigurableIOManager):
 
 
 @io_manager(
-    config_schema=S3IOManagerResource.to_config_schema(),
+    config_schema=PickledObjectS3IOManagerResource.to_config_schema(),
     required_resource_keys={"s3"},
 )
 def s3_pickle_io_manager(init_context):
--- a/python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
@@ -3,8 +3,8 @@ import pickle
 from typing import Sequence, Union
 
 from dagster import (
-    ConfigurableIOManagerFactory,
     InputContext,
+    IOManagerFactoryResource,
     MemoizableIOManager,
     MetadataValue,
     OutputContext,
@@ -103,7 +103,7 @@ class PickledObjectS3IOManager(MemoizableIOManager):
         context.add_output_metadata({"uri": MetadataValue.path(path)})
 
 
-class S3IOManagerResource(ConfigurableIOManagerFactory):
+class S3IOManagerResource(IOManagerFactoryResource[PickledObjectS3IOManager]):
     """Persistent IO manager using S3 for storage.
 
     Serializes objects via pickling. Suitable for objects storage for distributed executors, so long
@@ -155,7 +155,7 @@ class S3IOManagerResource(ConfigurableIOManagerFactory):
         default="dagster", description="Prefix to use for the S3 bucket for this file manager."
     )
 
-    def create_io_manager(self, context) -> PickledObjectS3IOManager:
+    def provide_object_for_execution(self, context) -> PickledObjectS3IOManager:
         return PickledObjectS3IOManager(
             s3_bucket=self.s3_bucket,
             s3_session=self.s3_resource,
--- a/python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
@@ -103,7 +103,7 @@ class PickledObjectS3IOManager(MemoizableIOManager):
         context.add_output_metadata({"uri": MetadataValue.path(path)})
 
 
-class S3PickleIOManager(ConfigurableIOManagerFactory):
+class S3IOManagerResource(ConfigurableIOManagerFactory):
     """Persistent IO manager using S3 for storage.
 
     Serializes objects via pickling. Suitable for objects storage for distributed executors, so long
@@ -164,7 +164,7 @@ class S3PickleIOManager(ConfigurableIOManagerFactory):
 
 
 @io_manager(
-    config_schema=S3PickleIOManager.to_config_schema(),
+    config_schema=S3IOManagerResource.to_config_schema(),
     required_resource_keys={"s3"},
 )
 def s3_pickle_io_manager(init_context):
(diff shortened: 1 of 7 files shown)


benpankow:
* master
  * **PR #13430** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13430" target="_blank">Graphite</a> 
    * **PR #13419** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13419" target="_blank">Graphite</a> 
      * **PR #13421** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13421" target="_blank">Graphite</a> 
        * **PR #13433** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13433" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13433?utm_source=stack-comment).

python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
```diff
@@ -100,11 +101,71 @@ def handle_output(self, context, obj):
...
+                    "io_manager": S3PickleIOManager(
+                        s3_resource=S3Resource(),
+                        s3_bucket="my-cool-bucket",
+                        s3_prefix="my-cool-prefix",
+                    )
+                },
+            )
+
+
+    """
```
schrockn, line 151: update docblock ot `Definitions`, etc

python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
```diff
@@ -100,11 +101,71 @@ def handle_output(self, context, obj):
...
+
+    s3_resource: ResourceDependency[Any]
```
schrockn, line 153: `ResourceDependency[S3Resource]` ?

python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
```diff
@@ -100,11 +101,71 @@ def handle_output(self, context, obj):
...
+    s3_prefix: str = Field(
+        default="dagster", description="Prefix to use for the S3 bucket for this file manager."
+    )
+
+    def create_io_manager(self, context) -> PickledObjectS3IOManager:
+        return PickledObjectS3IOManager(
+            s3_bucket=self.s3_bucket,
+            s3_session=self.s3_resource,
+            s3_prefix=self.s3_prefix,
+        )
```
schrockn, line 164: seems like we have another naming problem here in terms of `S3PickleIOManager` versus `PickledObjectS3IOManager`

python_modules/libraries/dagster-aws/dagster_aws/s3/resources.py
```diff
@@ -1,77 +1,107 @@
...
-        StringSource,
-        description="Specifies a custom endpoint for the S3 session",
-        is_required=False,
-    ),
-    "max_attempts": Field(
-        int,
+T = TypeVar("T")
+
+
+class S3ResourceBase(ConfigurableResourceFactory[T]):
```
schrockn, line 16: same discussion to be had as the upstream diff about these "resource base" classes

python_modules/libraries/dagster-aws/dagster_aws/s3/resources.py
```diff
@@ -1,77 +1,107 @@
...
+                    Bucket='my-bucket',
+                    Prefix='some-key'
+                )
+
+            @job(resource_defs={
+                's3': S3Resource(region_name='us-west-1')
+            })
+            def example_job():
+                example_s3_op()
+
```
schrockn, line 83: update docblock

python_modules/libraries/dagster-aws/dagster_aws/s3/resources.py
```diff
@@ -140,47 +170,40 @@ def example_job():
...
+    return S3Resource.from_resource_context(context)
+
+
+class S3FileManagerResource(S3ResourceBase[S3FileManager]):
+    s3_bucket: str = Field(description="S3 bucket to use for the file manager.")
+    s3_prefix: str = Field(
+        default="dagster", description="Prefix to use for the S3 bucket for this file manager."
     )
 
+    def create_resource(self, context: InitResourceContext) -> S3FileManager:
```
schrockn, line 182: the more I see `create_resource` the more I cringe here.

I think we should consider `create_object_for_user_code` or something which will also solve the same io_manager problem and will allow us to consolidate those two methods in the base class

schrockn (dismissed):

python_modules/libraries/dagster-aws/dagster_aws/s3/resources.py
benpankow: Updated to `ResourceWithS3Configuration`, for now

python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
benpankow: Yeah, this is tricky since we need to maintain backcompat. This might be worth bikeshedding on a call.

python_modules/libraries/dagster-aws/dagster_aws/s3/resources.py
benpankow: I can put up a PR for this - are we OK breaking the `create_resource` path for existing resources (e.g. users who are trying the experimental APIs out?). 

We could also do some magic to find an impl of `create_resource` and defer to that if `create_object_for_user_code` isn't overridden.

schrockn: Yes it is experimental so that we have the latitude to do exactly that.

python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
```diff
@@ -100,11 +103,68 @@ def handle_output(self, context, obj):
         context.add_output_metadata({"uri": MetadataValue.path(path)})
 
 
+class S3PickleIOManager(ConfigurableIOManagerFactory):
```
schrockn, line 106: i/o is just the abc

```
class S3IOManagerResource(FactoryResource[PickledObjectS3IOManager]):
  pass
```

according to our discussed naming convention

schrockn (changes requested):
Popping back in your queue. Do rename as discusses and restack on new proposed naming scheme.

python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
```diff
@@ -100,11 +103,68 @@ def handle_output(self, context, obj):
         context.add_output_metadata({"uri": MetadataValue.path(path)})
 
 
+class S3IOManagerResource(IOManagerFactoryResource[PickledObjectS3IOManager]):
```
schrockn, line 106: even thought it is more verbose, I do think that it is more consistent and clear to call this `PickledObjectS3IOManagerResource`

schrockn (changes requested):
makes me a bit sad, but I think we use the long name OR rename PickledObjectS3IOManager and have a consistent naming convention

python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
```diff
@@ -100,11 +104,75 @@ def handle_output(self, context, obj):
         context.add_output_metadata({"uri": MetadataValue.path(path)})
 
 
+class PickledObjectS3IOManagerResource(ConfigurableIOManager):
```
schrockn, line 107: ConfigurablePickledObjectS3IOManager?

python_modules/libraries/dagster-aws/dagster_aws/s3/ops.py
```diff
@@ -66,12 +66,25 @@ def last_key(key: str) -> str:
...
+    file_manager_resource = context.resources.file_manager
+    s3_resource = context.resources.s3
+
+    # Extract clients from new Pythonic resource types, if needed
+    file_manager: S3FileManager = file_manager_resource
+    if isinstance(file_manager_resource, S3FileManagerResource):
+        file_manager = file_manager_resource.create_client()
+    s3 = s3_resource
+    if isinstance(s3_resource, S3Resource):
+        s3 = s3_resource.create_client()
```
schrockn, line 84: I think we should hold off on merging this until https://github.com/dagster-io/dagster/pull/13495 is merged so we can demonstrate this as an example

python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
```diff
@@ -100,11 +104,75 @@ def handle_output(self, context, obj):
...
+    """
+
+    s3_resource: ResourceDependency[S3Resource]
+    s3_bucket: str = Field(description="S3 bucket to use for the file manager.")
+    s3_prefix: str = Field(
+        default="dagster", description="Prefix to use for the S3 bucket for this file manager."
+    )
+
+    @cached_method
+    def inner_db_io_manager(self) -> PickledObjectS3IOManager:
```
schrockn, line 160: `inner_io_manager`

schrockn (changes requested):
Some naming stuff.

python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
```diff
@@ -100,11 +104,75 @@ def handle_output(self, context, obj):
...
+            s3_prefix=self.s3_prefix,
+        )
+
+    def load_input(self, context: InputContext) -> Any:
+        return self.inner_db_io_manager().load_input(context)
+
+    def handle_output(self, context: OutputContext, obj: Any) -> None:
+        return self.inner_db_io_manager().handle_output(context, obj)
+
+
```
schrockn, line 173: @sryza this is still definitely making me wince. I think this will be copied and pasted throughout the land and it's quite a bit of boilerplate, in the end it's extremely similar behavior.

Definitely still think we should support people making IOManagers directly configurable but forcing us to do these delegation shenanigans rather than just support factory directly seems like a lot.

schrockn (changes requested):
Def should not name it PickledObjectS3IOManagerResource

python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
schrockn: Regardless of the wisdom of invoking list_objects as a health check, its there and removing it is a large behavior change that will potentially break users if we remove it.

python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
```diff
@@ -100,11 +104,75 @@ def handle_output(self, context, obj):
...
+    s3_bucket: str = Field(description="S3 bucket to use for the file manager.")
+    s3_prefix: str = Field(
+        default="dagster", description="Prefix to use for the S3 bucket for this file manager."
+    )
+
+    @cached_method
+    def inner_db_io_manager(self) -> PickledObjectS3IOManager:
+        return PickledObjectS3IOManager(
+            s3_bucket=self.s3_bucket,
+            s3_session=self.s3_resource,
```
schrockn, line 163: wait what is the behavior here? I think this will not work? we need to call `self.s3_resource.create_client()` no?

schrockn (changes requested):
much better with `IAttachDifferentObjectToOpContext`. However I'm not convinced that that resource dependency works as advertised.

python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py
benpankow: You're right. The test doesn't catch this since it was passing in an old-style S3 resource. Updated tests + this code accordingly.

schrockn (approved):

PR #13440, opened 2023-04-07
[configurable resources] Update dataclass_transform annotation to support aliases

benpankow:
## Summary

Updates the [`__dataclass_transform__`](https://peps.python.org/pep-0681/) decorator on our custom `ConfigurableResource` metaclass to point to Pydantic Field classes as field specifiers. This helps Pyright and other type checking tools recognize aliases.

## Test Plan

Tested in editor manually to verify that alias was respected.

Added pyright type signature unit test.

### DIFF
--- a/python_modules/dagster/dagster/_config/structured_config/typing_utils.py
+++ b/python_modules/dagster/dagster/_config/structured_config/typing_utils.py
@@ -1,6 +1,8 @@
 from typing import TYPE_CHECKING, Any, Generic, Optional, Type, TypeVar, Union, cast
 
 import pydantic
+from pydantic import Field
+from pydantic.fields import FieldInfo
 from typing_extensions import dataclass_transform, get_origin
 
 from .utils import safe_is_subclass
@@ -49,7 +51,7 @@ class LateBoundTypesForResourceTypeChecking:
         )
 
 
-@dataclass_transform()
+@dataclass_transform(kw_only_default=True, field_specifiers=(Field, FieldInfo))
 class BaseResourceMeta(pydantic.main.ModelMetaclass):
     """Custom metaclass for Resource and PartialResource. This metaclass is responsible for
     transforming the type annotations on the class so that Pydantic constructor-time validation
--- a/python_modules/dagster/dagster/_config/structured_config/typing_utils.py
+++ b/python_modules/dagster/dagster/_config/structured_config/typing_utils.py
@@ -2,7 +2,6 @@ from typing import TYPE_CHECKING, Any, Generic, Optional, Type, TypeVar, Union, 
 
 import pydantic
 from pydantic import Field
-from pydantic.fields import FieldInfo
 from typing_extensions import dataclass_transform, get_origin
 
 from .utils import safe_is_subclass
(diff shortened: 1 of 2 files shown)


benpankow:
* master
  * **PR #13440** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13440" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13440?utm_source=stack-comment).

python_modules/dagster/dagster/_config/structured_config/typing_utils.py
```diff
@@ -49,7 +51,7 @@ def set_actual_types_for_type_checking(
         )
 
 
-@dataclass_transform()
+@dataclass_transform(kw_only_default=True, field_specifiers=(Field, FieldInfo))
```
benpankow, line 54: https://github.com/pydantic/pydantic/blob/main/pydantic/main.py#L72

benpankow: `kw_only_default` helps with type annotation here, already the new config/resources will break if you feed them non-kwargs (hence the change in tests below)

jamiedemaria: more out of curiosity, what does the field_specifiers do?

benpankow: They basically tell Python & tools like the type checkers what classes to look for for metadata on each field. In this case, it's telling Pyright to look at Pydantic `Field` for metadata like the alias

https://peps.python.org/pep-0681/#field-specifiers has more info:
> `alias` is an optional str parameter that provides an alternative name for the field. This alternative name is used in the synthesized `__init__ `method.

schrockn: FYI this is the declaration in pydantic itself: 

https://github.com/pydantic/pydantic/blob/251c22c7d224d9ed9aea67777a2494674c5d0886/pydantic/main.py#L73

```
@typing_extensions.dataclass_transform(kw_only_default=True, field_specifiers=(Field,))
class ModelMetaclass(ABCMeta):
    def __new__(
        mcs,
        cls_name: str,
        bases: tuple[type[Any], ...],
        namespace: dict[str, Any],
        __pydantic_generic_metadata__: _generics.PydanticGenericMetadata | None = None,
        __pydantic_reset_parent_namespace__: bool = True,
        **kwargs: Any,
    ) -> type:
```

benpankow: Hmm, yeah, looks like they got rid of `FieldInfo`. I think we can remove it as well since it's an internal API rather than something users would use.

jamiedemaria (commented):
this looks good to me, but will defer to Nick who knows more about the internals of the config system. I did a quick test and the pyright issues (both the vscode warning and the actual pyright failure) go away with this change

schrockn (approved):

PR #13441, opened 2023-04-07
[ui] Dead code cleanup

hellendag:
## Summary & Motivation

Delete dead code, as identified by `ts-prune`.

## How I Tested These Changes

Buildkite.

In terminal tabs, run `yarn ts --watch` for `dagit/packages/core` and Cloud. Run `yarn find-dead-code` in `packages/core`. Remove exports, values, and files, and verify that TS continues to pass in both repos.


js_modules/dagit/packages/core/src/assets/groupByPartition.tsx
```diff
@@ -23,10 +23,7 @@ const sortByEventTimestamp = (a: Event, b: Event) => Number(b?.timestamp) - Numb
  * A hook that can bucket a list of materializations by partition, if any, with the `latest`
  * materialization separated from predecessor materializations.
  */
-export const groupByPartition = (
-  events: Event[],
-  definedPartitionKeys: string[],
-): AssetEventGroup[] => {
+const groupByPartition = (events: Event[], definedPartitionKeys: string[]): AssetEventGroup[] => {
```
bengotow, line 26: Kinda weird this is the namesake function of the file and isn't exported 👀 I wonder if even more is unused here?

js_modules/dagit/packages/core/src/partitions/PartitionGraphSet.tsx
```diff
@@ -1,176 +0,0 @@
-import {gql} from '@apollo/client';
-import * as React from 'react';
-import styled from 'styled-components/macro';
-
-import {PartitionGraph} from './PartitionGraph';
```
bengotow, line 5: Hmm @salazarm  maybe these were after your latest PR? I'm surprised these are unused...

Edit: no, looks like they've been unused for a while - nice!

bengotow (approved):
Reviewed the asset and partition related changes carefully and this all looks great! Thanks for running this, curious if it'll actually make autocompletion a bit better to have so many fewer things exported.

PR #13448, opened 2023-04-07
support removing assets and PartitionsDefinitions between reconciliat…

sryza:
…ion ticks

### DIFF
--- a/python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/definition_change_scenarios.py
+++ b/python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/definition_change_scenarios.py
@@ -1,7 +1,7 @@
 """Scenarios where the set of asset definitions changes between ticks.
 """
 
-from .asset_reconciliation_scenario import AssetReconciliationScenario, run_request
+from .asset_reconciliation_scenario import AssetReconciliationScenario
 from .basic_scenarios import one_asset, two_assets_in_sequence
 from .partition_scenarios import one_asset_one_partition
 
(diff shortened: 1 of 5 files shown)


python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/definition_change_scenarios.py
```diff
@@ -0,0 +1,45 @@
...
+    "asset_removed": AssetReconciliationScenario(
+        assets=[],
+        unevaluated_runs=[],
+        expected_run_requests=[],
+        cursor_from=AssetReconciliationScenario(
+            assets=one_asset,
+            unevaluated_runs=[],
+        ),
+    ),
+    "downstream_asset_removed": AssetReconciliationScenario(
```
clairelin135, line 18: can we also add a test for when the upstream asset is removed?

python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/definition_change_scenarios.py
```diff
@@ -0,0 +1,45 @@
...
+        expected_run_requests=[],
+        cursor_from=AssetReconciliationScenario(
+            assets=one_asset_one_partition,
+            unevaluated_runs=[],
+        ),
+    ),
+    "partitions_def_removed": AssetReconciliationScenario(
+        assets=one_asset,
+        unevaluated_runs=[],
+        expected_run_requests=[run_request(asset_keys=["asset1"])],
```
clairelin135, line 39: what happens when a partitions def is added to a previously unpartitioned asset that was materialized by the reconciliation sensor? do we kick off partitioned runs?

sryza: yes - I just added a test for this

clairelin135 (approved):
This looks good to me, thanks for adding this

sryza:
The `partitions_def_added` and `partitions_def_removed` tests are now failing because of details of how the test harness works with the auto-materialization test cases (we don't use the prior repo for `cursor_from`).

It will be easier to fix these after Johann merges his asset daemon PR. However that might not make it in for today's release, and I want this to make it in for today's release, so I took those test cases out for now.

Issue #13450, opened 2023-04-07
Freshness Policy doesn't work with AssetsDefinition.from_op with key_prefix set


tghanken:
### Dagster version

dagster, version 1.2.6

### What's the issue?

Freshness policies don't apply properly on Asset Definitions that use a key prefix.

### What did you expect to happen?

I'd expect to be able to define a `FreshnessPolicy` on an asset defined using `AssetsDefinition.from_op()` while using the `key_prefix` parameter.

### How to reproduce?

Here's the minimal code I was able to use to reproduce the issue:

```python
from dagster import AssetsDefinition, FreshnessPolicy, Out, op


@op(
    out={"test_table_w_freshness": Out()},
)
def test_table_w_freshness():
    return ""


@op(
    out={"test_table_wo_freshness": Out()},
)
def test_table_wo_freshness():
    return ""


# Freshness policy correctly applies to this asset
asset_w_freshness_policy = AssetsDefinition.from_op(
    op_def=test_table_w_freshness,
    freshness_policies_by_output_name={
        "test_table_w_freshness": FreshnessPolicy(maximum_lag_minutes=60)
    },
)

# This asset doesn't have a freshness policy in dagit
asset_wo_freshness_policy = AssetsDefinition.from_op(
    op_def=test_table_wo_freshness,
    key_prefix=["test"],
    freshness_policies_by_output_name={
        "test_table_wo_freshness": FreshnessPolicy(maximum_lag_minutes=60)
    },
)
```

### Deployment type

Local

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

clairelin135:
Thanks for reporting and contributing a PR to fix. Looks great pending one comment that I left on the PR!

cmpadden:
Closing as Dagster has moved on from the exact API shape. Thank you.

PR #13455, opened 2023-04-07
Change `AssetsDefinition._from_node()` to use output prefixes globally

tghanken:
## Summary & Motivation

We are trying to use Freshness Policies to make our pipelines easier to manage, but were running into issues where they were not being applied to specific assets.  After investigating, I narrowed down the issue to the `key_prefix` parameter in `AssetsDefinition.from_op()` and opened up issue #13450.

I looked into the underlying code a bit, and it seems like this is a very straightforward fix, so I threw together a PR.

## How I Tested These Changes

Using the following test file, I verified that `test_table_wo_freshness` doesn't show a freshness policy with Dagster v1.2.6, but does with this patch.

```python
from dagster import AssetsDefinition, FreshnessPolicy, Out, op

@op(out={"test_table_w_freshness": Out()})
def test_table_w_freshness():
    return ""

@op(out={"test_table_wo_freshness": Out()})
def test_table_wo_freshness():
    return ""

# Freshness policy correctly applies to this asset
asset_w_freshness_policy = AssetsDefinition.from_op(
    op_def=test_table_w_freshness,
    freshness_policies_by_output_name={
        "test_table_w_freshness": FreshnessPolicy(maximum_lag_minutes=60)
    },
)

# This asset doesn't have a freshness policy in dagit
asset_wo_freshness_policy = AssetsDefinition.from_op(
    op_def=test_table_wo_freshness,
    key_prefix=["test"],
    freshness_policies_by_output_name={
        "test_table_wo_freshness": FreshnessPolicy(maximum_lag_minutes=60)
    },
)
```

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/assets.py
@@ -327,7 +327,7 @@ class AssetsDefinition(ResourceAddable):
                 be associated with each of the output assets for this node. Keys are names of the
                 outputs, and values are dictionaries of metadata to be associated with the related
                 asset.
-            freshness_policies_by_output_name_ouptut_name (Optional[Mapping[str, Optional[FreshnessPolicy]]]): Defines a
+            freshness_policies_by_output_name (Optional[Mapping[str, Optional[FreshnessPolicy]]]): Defines a
                 FreshnessPolicy to be associated with some or all of the output assets for this node.
                 Keys are the names of the outputs, and values are the FreshnessPolicies to be attached
                 to the associated asset.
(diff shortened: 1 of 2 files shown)


clairelin135:
Hi @tghanken, thanks for reporting and pushing out a fix for this!

This looks great to me. Would you mind adding a test for this to ensure that this behavior is preserved? This file is probably a good place: https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster_tests/asset_defs_tests/test_assets.py

I think we just need to make sure that `assets_def.freshness_policies_by_key` fetches the freshness policy when the asset key contains a prefix.

tghanken:
Of course!  Do you prefer that I add assertions to the existing `test_from_op_w_key_prefix()` and `test_from_graph_w_key_prefix()` tests, or add new test functions?

clairelin135:
Adding additional assertions to `test_from_op_w_key_prefix` sounds great

clairelin135 (approved):
Looks great to me! Thanks for making this change

PR #13456, opened 2023-04-07
[pythonic resources] Better direct invocation experience with resources

benpankow:
## Summary

Now, ops and assets which use resources as params can specify those resources as `args` (previously just through `kwargs` or the `context`).

For example, the following are all valid ways to invoke this op:

```
    z: int

@op
def my_op(context, x: int, y: int, resource: MyResource):
    return x + y + resource.z


# pass resource as arg
my_op(build_op_context(), 1, 2, MyResource(z=3))

# pass resource as kwarg
my_op(build_op_context(), 1, 2, resource=MyResource(z=3))

# pass resource through context (no arg/kwarg)
my_op(build_op_context(resources={"resource": MyResource(z=3)}), 1, 2)
```

## Test Plan

Uncomment xfail tests, a few new test cases.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/op_invocation.py
+++ b/python_modules/dagster/dagster/_core/definitions/op_invocation.py
@@ -39,21 +39,21 @@ if TYPE_CHECKING:
 T = TypeVar("T")
 
 
-class ExtractedOutputs(NamedTuple):
-    args: Tuple[Any, ...]
-    kwargs: Dict[str, Any]
-    resource_inputs: Dict[str, Any]
+class InputsAndResources(NamedTuple):
+    input_args: Tuple[Any, ...]
+    input_kwargs: Dict[str, Any]
+    resources_by_param_name: Dict[str, Any]
 
 
-def _extract_resource_args(
+def _separate_args_and_kwargs(
     compute_fn: "DecoratedOpFunction",
     args: Tuple[Any, ...],
     kwargs: Dict[str, Any],
     resource_arg_mapping: Dict[str, Any],
-) -> ExtractedOutputs:
+) -> InputsAndResources:
     """Given a decorated compute function, a set of args and kwargs, and set of resource param names,
-    extracts any resource inputs from the args and kwargs and returns a tuple of the remaining
-    args and kwargs, along with a dict of the extracted resource inputs.
+    separates the set of resource inputs from op/asset inputs returns a tuple of the categorized
+    args and kwargs.
 
     We use the remaining args and kwargs to cleanly invoke the compute function, and we use the
     extracted resource inputs to populate the execution context.
@@ -84,10 +84,10 @@ def _extract_resource_args(
         k: v for k, v in kwargs.items() if k not in resource_inputs_from_args_and_kwargs
     }
 
-    return ExtractedOutputs(
-        args=tuple(adjusted_args),
-        kwargs=adjusted_kwargs,
-        resource_inputs=resource_inputs_from_args_and_kwargs,
+    return InputsAndResources(
+        input_args=tuple(adjusted_args),
+        input_kwargs=adjusted_kwargs,
+        resources_by_param_name=resource_inputs_from_args_and_kwargs,
     )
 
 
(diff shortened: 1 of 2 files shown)


benpankow:
* master
  * **PR #13456** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13456" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13456?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/op_invocation.py
```diff
@@ -59,31 +121,24 @@ def op_invocation_result(
...
-                " positional"
-                " arguments, only as keyword arguments."
-            ) from e
-        raise
-    input_dict = _resolve_inputs(op_def, args, kwargs, bound_context)
+    bound_context = context.bind(op_def_or_invocation)
+
+    input_dict = _resolve_inputs(op_def, adjusted_args, adjusted_kwargs, bound_context)
 
     result = invoke_compute_fn(
```
schrockn, line 143: I think this codepath would be a lot clearer if we renamed the `kwargs` argument in `invoke_compute_fn` to "input_by_param_name" or something. Because what is actually happening here is that post this workstream `invoke_compute_fn` is constructing the `kwargs` rather than getting them passed to.

python_modules/dagster/dagster/_core/definitions/op_invocation.py
```diff
@@ -59,31 +121,24 @@ def op_invocation_result(
...
-            resource_args_from_kwargs[resource_arg] = kwargs[resource_arg]
-            del kwargs[resource_arg]
 
-    resources_provided_in_multiple_places = (resource_args_from_kwargs) and (context.resource_keys)
+    # Extract any resource inputs from the args and kwargs
+    extracted = _extract_resource_args(compute_fn, args, kwargs, resource_arg_mapping)
+
+    adjusted_args = extracted.args
+    adjusted_kwargs = extracted.kwargs
+    resource_inputs_from_args = extracted.resource_inputs
```
schrockn, line 130: resource_inputs_from_args is a very weird variable name here. Combining "resource" and "input" in this way is add since input and resource are both terms of art in internal dagster ontology but they are combined here differently.

'resource_object_from_param_names`

python_modules/dagster/dagster/_core/definitions/op_invocation.py
```diff
@@ -59,31 +121,24 @@ def op_invocation_result(
...
-        if resource_arg in kwargs:
-            resource_args_from_kwargs[resource_arg] = kwargs[resource_arg]
-            del kwargs[resource_arg]
 
-    resources_provided_in_multiple_places = (resource_args_from_kwargs) and (context.resource_keys)
+    # Extract any resource inputs from the args and kwargs
+    extracted = _extract_resource_args(compute_fn, args, kwargs, resource_arg_mapping)
+
+    adjusted_args = extracted.args
+    adjusted_kwargs = extracted.kwargs
```
schrockn, line 129: should be say `input_args` and `input_kwargs`? I think that is what these actually are.

python_modules/dagster/dagster/_core/definitions/op_invocation.py
```diff
@@ -29,6 +39,58 @@
 T = TypeVar("T")
 
 
+class ExtractedOutputs(NamedTuple):
+    args: Tuple[Any, ...]
+    kwargs: Dict[str, Any]
+    resource_inputs: Dict[str, Any]
```
schrockn, line 45: suggest

`input_args`
`input_kwargs`
`resources_by_param_name`

python_modules/dagster/dagster/_core/definitions/op_invocation.py
```diff
@@ -59,31 +121,24 @@ def op_invocation_result(
...
 
     resource_arg_mapping = {arg.name: arg.name for arg in compute_fn.get_resource_args()}
-    resource_args_from_kwargs = {}
-    for resource_arg in resource_arg_mapping:
-        if resource_arg in kwargs:
-            resource_args_from_kwargs[resource_arg] = kwargs[resource_arg]
-            del kwargs[resource_arg]
 
-    resources_provided_in_multiple_places = (resource_args_from_kwargs) and (context.resource_keys)
+    # Extract any resource inputs from the args and kwargs
```
schrockn, line 125: A preamble comment to explain the overall process here might be useful, given the complexity of the codepath.

```
The user is allowed to invoke an op with an arbitrary mix of args and kwargs. The framework must ensure that all of these are correctly categorized as inputs, config, or resource objects and then validated.
 
 * Any resources passes as parameters are also made available to user-defined code in the passed context.
 * Provide high-quality error messages (e.g. if something tried to pass a value to an input typed Nothing)
 * Apply default values appropriately
 * Typechecks of inputs
 
Then it recollects them all into a dictionary and invokes the user-defined function with kwargs only.
```

schrockn (changes requested):
Had to think very very hard to figure out what was going on there. I think we can make this eaiser to read with some better naming and some commentary.

schrockn (approved):

PR #13458, opened 2023-04-08
[pythonic resources] Rename resource-related classes

benpankow:
## Summary & Motivation

Based on offline discussion with @schrockn, we're hoping to make the naming scheme for Pythonic resources more approachable. The new names are:

- `ConfigurableResource` -> `Resource`. This rename is simpler, and gives us a good grounding point for what a resource is. Now, a `Resource` subclass is clearly a resource, with no adjective necessary.
- `ConfigurableResourceFactory` -> `FactoryResource`. Previously, this name muddled what a resource is. A `ConfigurableResourceFactory` sounds like it builds a resource, wheras a `FactoryResource` more clearly is a resource which builds something else.
- `Resource[]` -> `FromResources[]`. With this terminology change, resource usage in ops and assets becomes more self-documenting. A user unfamiliar with Dagster can read code and understand that a config input comes from 'resources'.
  ```python
  @op
  def my_op(github: FromResources[GitHubClient]):
        github.get_users()
  ```
- `Resource.create_resource` -> `Resource.provide_object_for_execution`. Now a resource does not define a `create_resource` method on itself, but rather produces some object for user code.

## How I Tested These Changes

Existing tests

PR #13460, opened 2023-04-10
[pythonic resources] Update IO-manager class names

benpankow:
## Summary & Motivation

Based on offline discussion with @schrockn and stacking on #13458 , we're hoping to make IO Manager Pythonic classes make more sense by creating a better dividing line between the `IOManager` ABC and the resource system.

- `ConfigurableIOManager` -> `IOManagerResource`. This rename is simpler and clearly communicates that this class represents that something is both an IO Manager and a Resource. This is not necessarily true in all cases - some resources produce IO managers which are separate, but subclasses of `IOManagerResource` are both.
- `ConfigurableIOManagerFactory` -> `IOManagerFactoryResource`. This one is ugly, but some behind-the-scenes changes could eliminate the need for it (using `FactoryResource[MyIOManager]` instead). See stacked PR for more.

## How I Tested These Changes

Existing tests

### DIFF
--- a/docs/content/guides/dagster/pythonic-resources.mdx
+++ b/docs/content/guides/dagster/pythonic-resources.mdx
@@ -287,10 +287,10 @@ from dagster import (
     AssetKey,
     OutputContext,
     InputContext,
-    ConfigurableIOManager,
+    IOManagerResource,
 )
 
-class MyIOManager(ConfigurableIOManager):
+class MyIOManager(IOManagerResource):
     root_path: str
 
     def _get_path(self, asset_key: AssetKey) -> str:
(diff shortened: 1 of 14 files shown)


benpankow:
* master
  * **PR #13458** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13458" target="_blank">Graphite</a> 
    * **PR #13460** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13460" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13460?utm_source=stack-comment).

PR #13462, opened 2023-04-10
[dagit] Split Stale tag into several more specific tags, update catalog status column

bengotow:
## Summary & Motivation

This PR is linked to the discussion here: https://github.com/dagster-io/dagster/discussions/13102, https://github.com/dagster-io/internal/discussions/5381

- Replaces the word "Late" with "Overdue"

- Replaces generalized uses of the concept "is stale" with "has unpropagated changes"

- Splits the single "Stale" tag into three. They can appear above the asset node if they refer to it's inputs, or inside the asset node if the referenced assetkey is the current one.

- The asset catalog now has a "Status" column instead of "Last Materialized" which matches the asset graph. To keep things sane and always consistent, the catalog and assetNode share a bit of the rendering code.

image

image

## Testing Approach

I used our storybooks to do most of the required AssetNode changes and update the asset catalog table. Because these are just new rendering states for largely existing states, i think this is a good way to verifying correctness.

I verified with Sean that the upstream tags are working as expected on a test DAG.


js_modules/dagit/packages/core/src/assets/Stale.tsx
```diff
@@ -0,0 +1,127 @@
...
+  onClick?: () => void;
+}> = ({liveData, include, assetKey, onClick}) => {
+  if (!isAssetStale(liveData) || !liveData?.staleCauses.length) {
+    return null;
+  }
+
+  return (
+    <Body color={Colors.Yellow700}>
+      <Tooltip position="top" content={<StaleCausesSummary causes={liveData.staleCauses} />}>
+        <Box onClick={onClick} flex={{gap: 4}}>
```
hellendag, line 43: I don't love `onClick`s on divs. Can this be a `ButtonLink` or something instead? Also it has a `gap` applied but there's just the one string child.

js_modules/dagit/packages/core/src/assets/Stale.tsx
```diff
@@ -0,0 +1,127 @@
...
+}> = ({liveData, include, assetKey, onClick}) => {
+  if (!isAssetStale(liveData) || !liveData?.staleCauses.length) {
+    return null;
+  }
+
+  return (
+    <>
+      {Object.entries(groupedCauses(assetKey, include, liveData)).map(([label, causes]) => (
+        <Tooltip key={label} position="top" content={<StaleCausesSummary causes={causes} />}>
+          <Box onClick={onClick} flex={{gap: 4}}>
```
hellendag, line 65: Same comments as above.

hellendag (approved):

js_modules/dagit/packages/core/src/assets/__fixtures__/AssetsCatalogTable.mocks.ts
```diff
@@ -355,6 +358,7 @@ export const AssetCatalogTableMock: MockedResponse<AssetCatalogTableQuery> = {
               id: 'test.py.repo.["good_asset"]',
               groupName: 'GROUP2',
               isSource: false,
+              isObservable: false,
```
salazarm, line 361: Not sure if you need this for your tests or if its just for typescript but If its the latter then using the `build` functions will make it so you don't need to update the mocks again in the future.

bengotow: Ahh let me switch these out to the build functions that would be nice here actually!

bengotow:
Test failures are unrelated, merging this!

PR #13482, opened 2023-04-10
Enable observations on partitioned source assets

smackesey:
## Summary & Motivation

This fixes a bug with performing on observations on source assets. The bug arose from the fact that defining "job_names" for source assets was done via a different code path than the per-partitions-def "base asset job" calculation.

## How I Tested These Changes

New unit test

### DIFF
--- a/examples/docs_snippets/docs_snippets_tests/concepts_tests/assets_tests/test_observable_source_asset.py
+++ b/examples/docs_snippets/docs_snippets_tests/concepts_tests/assets_tests/test_observable_source_asset.py
@@ -1,6 +1,8 @@
 from dagster import repository
 from dagster._core.definitions.assets_job import build_assets_job
-from dagster._core.definitions.data_version import extract_data_version_from_entry
+from dagster._core.definitions.data_version import (
+    extract_data_version_from_entry,
+)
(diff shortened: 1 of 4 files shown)


smackesey:
* master
  * **PR #13482** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13482" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13482?utm_source=stack-comment).

sryza:
Or maybe a different way of asking the question: what happens when they click the Observe button?

sryza (commented):
Do they have access to the partition key from inside their observation function? I assume that they would need that to know what physical address to observe.

smackesey:
>How does the user select what partitions to observe?

This needs some UI work. I am coordinating with @bengotow here, but this PR should land regardless of if that gets done because it fixes a bug that prevents any observation from occurring (even on non-partitioned observables) if there are partitioned assets in the graph.

>Do they have access to the partition key from inside their observation function? I assume that they would need that to know what physical address to observe.

Yes, the source asset observation context is just `OpExecutionContext`. I updated a test to use the `partition_key`.

PR #13492, opened 2023-04-10
[docs] Update dynamic partitions sensor examples to use `SensorResult`

clairelin135:
Change docs references instead to use `SensorResult` instead of `context.instance.add_dynamic_partitions(...)`


python_modules/dagster/dagster/_core/definitions/partition.py
```diff
@@ -607,8 +607,10 @@ class DynamicPartitionsDefinition(
 
             @sensor(job=my_job)
             def my_sensor(context):
-                context.instance.add_dynamic_partitions(fruits.name, [partition_key])
-                return RunRequest(partition_key=partition_key)
+                return SensorResult(
+                    run_requests=[RunRequest(partition_key="apple")],
+                    dynamic_partitions_requests=[fruits.build_add_request([partition_key])]
```
sryza, line 612: Where is `partition_key` defined?

sryza (commented):
Do `SensorResult` and the new request classes show up in the api doc?

python_modules/dagster/dagster/_core/definitions/partition.py
clairelin135: good catch, i've changed this

clairelin135:
`SensorResult` and the request classes now show up in the sensor api docs

sryza (approved):

PR #13494, opened 2023-04-11
GQL resolver for backfill creator

clairelin135:
In cloud, [we add a "user" tag to runs](https://github.com/dagster-io/internal/blob/0b753a1c0999c35ddd61a1c6c4292c4500c67aa3/dagster-cloud/python_modules/dagster-cloud-backend/dagster_cloud_backend/run_coordinator/dagster_cloud_run_coordinator.py#L49) containing the user's email as a value. This enables displaying who launched a run in Dagit.

We want to enable a similar functionality for backfills--the "user" tag can be added in the launch backfill mutation to be persisted as part of the backfill. This PR adds a resolver to return the corresponding value.

I wish we'd prefixed "user" with the `dagster/` prefix... for now this PR continues using `"user"` for consistency, but I'm open to also using a new `dagster/user` tag moving forward.


salazarm (approved):
this looks good to me!

PR #13493, opened 2023-04-11
[pythonic resources] Rename Resource[] annotation to ResouceParam[]

benpankow:
## Summary

Based on a discussion with @schrockn and @sryza, we are going to rename the `Resource` annotation to `ResourceParam`, which is a bit more descriptive and avoids defining the `Resource` keyword (which could prompt users to accidentally subclass it).

## Test Plan

Existing unit tests.

PR #13495, opened 2023-04-11
[pythonic resources] Implement IAttachOtherObjectToContext migration aide

benpankow:
## Summary

Introduces the new `IAttachOtherObjectToOpContext` migration aide interface, which `ConfigurableResource`s can implement to make it easier to migrate existing resources.

## Test Plan

Invocation tests for ops.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/scoped_resources_builder.py
+++ b/python_modules/dagster/dagster/_core/definitions/scoped_resources_builder.py
@@ -29,7 +29,7 @@ class Resources:
         raise DagsterUnknownResourceError(name)
 
     @property
-    def _unmodified_resource_dict(self) -> Mapping[str, object]:
+    def _original_resource_dict(self) -> Mapping[str, object]:
         raise NotImplementedError()
 
 
@@ -59,7 +59,7 @@ class ScopedResourcesBuilder(
         )
 
     def build(self, required_resource_keys: Optional[AbstractSet[str]]) -> Resources:
-        from dagster._config.structured_config import IAttachBareObjectToContext
+        from dagster._config.structured_config import IAttachDifferentObjectToContext
 
         """We dynamically create a type that has the resource keys as properties, to enable dotting into
         the resources from a context.
@@ -84,10 +84,10 @@ class ScopedResourcesBuilder(
             for key in required_resource_keys
             if key in self.resource_instance_dict
         }
-        processed_resource_instance_dict = {
+        resources_to_attach_to_context = {
             k: (
                 v.get_object_to_set_on_execution_context()
-                if isinstance(v, IAttachBareObjectToContext)
+                if isinstance(v, IAttachDifferentObjectToContext)
                 else v
             )
             for k, v in resource_instance_dict.items()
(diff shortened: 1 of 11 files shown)


benpankow:
* master
  * **PR #13493** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13493" target="_blank">Graphite</a> 
    * **PR #13495** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13495" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13495?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/scoped_resources_builder.py
```diff
@@ -78,28 +84,41 @@ def build(self, required_resource_keys: Optional[AbstractSet[str]]) -> Resources
             for key in required_resource_keys
             if key in self.resource_instance_dict
         }
+        processed_resource_instance_dict = {
```
benpankow, line 87: Here we get all the bare values from `IAttachBareObjectToContext` resources, but we keep around the raw resource objects so that the op/sensor/schedule invocation code can splat them into kwargs.

schrockn: I think we should use the terms "original resource" objects to describe the ones passed into Definitions and then call `processed_resource_instance_dict` `resources_to_attach_to_context`. Just in these cases be as specific as possible.

python_modules/dagster/dagster/_core/execution/context/compute.py
```diff
@@ -109,7 +109,11 @@ def hello_world(context: OpExecutionContext):
 
     __slots__ = ["_step_execution_context"]
 
-    def __init__(self, step_execution_context: StepExecutionContext):
+    def __init__(
+        self,
+        step_execution_context: StepExecutionContext,
+        _resources_override: Optional[Any] = None,
+    ):
         self._step_execution_context = check.inst_param(
```
schrockn, line 117: I don't think this is necessary

python_modules/dagster/dagster/_config/structured_config/__init__.py
```diff
@@ -1240,3 +1241,61 @@ def validate_resource_annotated_function(fn) -> None:
...
+    return resource_obj.__class__(
+        **{
+            k: (
+                v.get_object_to_set_on_execution_context()
+                if isinstance(v, IAttachBareObjectToContext)
+                else v
+            )
+            for k, v in resource_obj._asdict().items()
+        }
+    )
```
schrockn, line 1301: and this is also no longer necessary

python_modules/dagster/dagster/_config/structured_config/__init__.py
```diff
@@ -1240,3 +1241,61 @@ def validate_resource_annotated_function(fn) -> None:
                 else "'ResourceParam[Any]' or 'ResourceParam[<output type>]'",
             )
         )
+
+
+class IAttachBareObjectToContext:
```
schrockn, line 1246: IAttachDifferentObjectOnContext?

"Bare" isn't really a meaningful adjective here IMO

schrockn: We could also consider not having the interface at all and instead just having the method `get_object_to_set_on_execution_context` to avoid exporting another public symbol.

However I do like the marker interface and it makes it clear that this is a temporary, compartmentalized measure. (We should move it to its own file)

schrockn (dismissed):
This is much more under control. Nice work.

schrockn (dismissed):
Given that the feature is new, I don't think we need to support schedules and sensors.  We can name it `IAttachDifferentObjectToOpContext` to make it clear that we only support it there. This will narrow the surface area of this.

schrockn (approved):
Great. Ended up being pretty minimally invasive.

PR #13510, opened 2023-04-11
[ui] Fix 60s+ countdown display

hellendag:
## Summary & Motivation

Repair countdown display in situations where the time is greater than 60s. Right now we're just blindly assuming the countdown will be less than a minute, which isn't true in cases where we adjust for long queries. In those situations, we end up with stuff like `0:167`.

Break the seconds amount up into a hh:mm:ss format.

## How I Tested These Changes

Jest


hellendag:
* master
  * **PR #13510** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13510" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13510?utm_source=stack-comment).

PR #13511, opened 2023-04-11
Forward assetSelection from RunRequests created when testing Sensors

salazarm:
## Summary & Motivation

Sensors support the concept of "assetSelection". When testing a sensor we render a table with a launchpad link to run the job. This PR makes that link include the `assetSelection` from the `RunRequest` object.
 
## How I Tested These Changes

https://www.loom.com/share/c39520a3db3542468dd4ca1db8bb38b2


dpeng817 (approved):
Backend changes lgtm - to my untrained eye frontend looks reasonable as well, but not really familiar with those codepaths

js_modules/dagit/packages/core/src/launchpad/LaunchpadRoot.tsx
```diff
@@ -71,6 +71,16 @@ export const JobLaunchpad: React.FC<{repoAddress: RepoAddress}> = (props) => {
     return <Redirect to={`/locations/${repoPath}/pipeline_or_job/${pipelinePath}`} />;
   }
 
+  if (pipelinePath.includes('__ASSET_JOB')) {
```
bengotow, line 74: Hmm we have a constant for this value - ASSET_JOB_PREFIX I think

bengotow (approved):
This looks good to me, just added one inline comment.

js_modules/dagit/packages/core/src/launchpad/LaunchpadRoot.tsx
salazarm: TIL, thanks!

PR #13525, opened 2023-04-11
Revert opening each dagster subprocess in its own process group

gibsondan:
Summary:
This reverts open_ipc_subprocess to its original state before I messed with its subprocess lifecycle - I thought that it would be desirable in general to run grpc servers in their own process groups as well, but that seems to have opened up a whole host of sketchy issues related to stdin.

Instead, add a hook to dagit so that even if it spins down a bit messily, it doesn't log an unsightly CancelledError to the terminal

### DIFF
--- a/python_modules/dagster/dagster/_cli/api.py
+++ b/python_modules/dagster/dagster/_cli/api.py
@@ -739,6 +739,9 @@ def grpc_command(
 
         try:
             server.serve()
+        except KeyboardInterrupt:
+            # Terminate cleanly on interrupt
+            logger.info("Code server was interrupted")
         finally:
             logger.info("Shutting down %s", server_desc)
 
--- a/python_modules/dagit/dagit/app.py
+++ b/python_modules/dagit/dagit/app.py
@@ -15,6 +15,7 @@ from .webserver import DagitWebserver
 def create_app_from_workspace_process_context(
     workspace_process_context: WorkspaceProcessContext,
     path_prefix: str = "",
+    **kwargs,
 ) -> Starlette:
     check.inst_param(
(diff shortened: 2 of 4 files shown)


python_modules/dagster/dagster/_cli/dev.py
```diff
@@ -118,14 +118,17 @@ def dev_command(
         if kwargs.get("working_directory"):
             args.extend(["--working-directory", check.str_elem(kwargs, "working_directory")])
 
+        # open each subprocess in its own process group so that terminal CTRL-C only goes
+        # to this process, allowing us to cleanly terminate the subprocesses
```
alangenfeld, line 122: rm?

alangenfeld (approved):

python_modules/dagster/dagster/_cli/api.py
```diff
@@ -739,6 +739,9 @@ def grpc_command(
 
         try:
             server.serve()
+        except KeyboardInterrupt:
+            # Terminate cleanly on interrupt
+            pass
```
gibsondan, line 744: @alangenfeld added this but seems uncontroversial

python_modules/dagster/dagster/_cli/api.py
```diff
@@ -739,6 +739,9 @@ def grpc_command(
 
         try:
             server.serve()
+        except KeyboardInterrupt:
+            # Terminate cleanly on interrupt
+            logger.info("Code server was interrupted")
```
alangenfeld, line 744: i like the log line - we should maybe do that for the dagit supression as well

PR #13527, opened 2023-04-11
add gcs compute log manager to docs site

prha:
## Summary & Motivation
GCS compute log manager is undocumented

## How I Tested These Changes
Generated docs site locally, saw both API and the instance guide updated.


erinkcochran87 (commented):
A few small things in the API docs. Adding the screenshot for reference, since the deploy preview is broken (if you pull updates from master to this branch and then push again, it should fix it) and I can't comment directly on the docstrings.

Screen Shot 2023-04-12 at 3 10 40 PM

- The Dagster instance doc updates demonstrate how users can use env vars to supply values for these fields. Do we need to mention it in the API docs as well? Or link to the example in the Dagster instance doc?
- Can you please update `gcs` > `GCS` in the parameter descriptions? Anywhere it's being used as an acronym and not a literal value / in code
- In the description for `json_credentials_envvar`, could you please update it to the following, to address a few typos?:

  > Environment variable that contains the JSON
- In `inst_data`, what does `newed up` mean?

erinkcochran87 (approved):
looks good, ty!

PR #13529, opened 2023-04-11
Update stale reasons for descriptiveness

smackesey:
## Summary & Motivation

This updates the "stale reasons" that are shown in the dagit tooltip:

- "removed dependency" -> "removed dependency on foo/bar"
- "new dependency" -> "has a new dependency on foo/bar"
- "updated code version" -> "has a new code version"
- "updated data version" ->
    - "has a new data version" IF the asset has a defined code version OR the upstream data version is user-generated
    - "has a new materialization" IF the asset has no defined version
- "updated timestamp" -> "has a new materialization"

The reason for the split in "updated data version" is that, if an asset has no defined code version, it will automatically get a new data version on every materialization, and the user likely hasn't thought about data versions, so we just say "new materialization".

In order to enable the correct split behavior, I also needed to add a tag to mark user-generated data versions on materialization events.

## How I Tested These Changes

Updated existing test suite (sufficient tests existed to cover both the new tag and the reasons themselves).


smackesey:
* master
  * **PR #13529** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13529" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13529?utm_source=stack-comment).

sryza:
@smackesey are you able to include a screenshot of one of these? Trying to understand what these look like in context. My initial impression is that these name changes are good.

sryza:
Found a screenshot:

sryza:
I found the square bracket format a little difficult to parse. I think that a terse sentence might make it less open to different interpretations? E.g.
- "upstream_asset_x has a new data version"
- "asset_x has a new code version"

sryza:
@bengotow @braunjj I think it could also be helpful to apply some visual treatment to the asset names inside the tooltip - e.g. highlight them or use a monospace font - to make it clear that they're special values and not just part of the explanatory text.

smackesey:
Yes I think instead of square brackets we should have keys distinguished with styling

bengotow (approved):
These seem like nice improvements to me 👍

smackesey:
>Connective words make the relationship clear between the asset keys and descriptors like "new upstream data". Without them, the user needs to make a small logical jump.

My opinion isn't too strong so happy to go along if you feel strongly, but I can't really imagine an alternative interpretation of the terse text (without connective words) and prefer to bias towards density in stuff the user is likely to see frequently. I think with the extra words it becomes slightly harder to parse for the user who is used to these messages and adds some friction to debugging/navigating the graph.

It's really an instance of a more general trade-off in dagit between the user who is frequently looking at the graph and the more unfamiliar user.

braunjj:
Chiming in late here, but I tend to agree with @sryza. I like the clarity of ""upstream_asset_x has a new data version" -- it's less open to interpretation and since this is in a tooltip, IMO, it's ok to be more verbose.

smackesey:
OK, I've updated this to use "has a" and to also to add the tag I'd mentioned. Should be good to go now.

sryza (approved):
Nice. I think it's worth doing a little bit of QA (e.g. clicking around the data versions toy) to make sure this doesn't break anything.

PR #13534, opened 2023-04-11
update header text in staleness tooltip

sryza:
## Summary & Motivation

I was playing around with this a little bit, and thought the wording "This asset has unpropagated changes" could be a little confusing, because it implies that this asset has changed, while it's actually usually that other assets have changed.

I'm suggesting "Changes since last materialization". Another option could be "Unpropagated changes".  I also considered "Unpropagated changes since last materialization", but thought that "unpropagated" and "since last materialization" might be redundant.


bengotow (approved):
Oh I like this text, it's much more direct and removing "propagation" is a nice simplification.

PR #13536, opened 2023-04-11
tweak partition_expr error

sryza:
## Summary & Motivation

I've seen missing `partition_expr` be a stumbling block for a few people who are trying to use a DB IO manager with partitions.

I think the error message is already pretty helpful, but this is an attempt at making the error message as helpful as possible.  I took out "filter it on", because I thought it might not be clear to people why we would need to do filtering at the time that we're writing out data.

Let me know if you thinks this makes it less helpful in some way.

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/db_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/db_io_manager.py
@@ -196,9 +196,9 @@ class DbIOManager(IOManager):
                 if partition_expr is None:
                     raise ValueError(
                         f"Asset '{context.asset_key}' has partitions, but no 'partition_expr'"
-                        " metadata value, so we don't know what column to filter it on. Specify"
-                        " which column(s) of the database contains partitioned data as the"
-                        " 'partition_expr' metadata."
+                        " metadata value, so we don't know what column it's partitioned on. To"
+                        " specify a column, set this metadata value. E.g."
+                        ' @asset(metadata={"partition_expr": "your_partition_column"}).'
                     )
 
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster/_core/storage/db_io_manager.py
```diff
@@ -196,9 +196,9 @@ def _get_table_slice(
                 if partition_expr is None:
                     raise ValueError(
                         f"Asset '{context.asset_key}' has partitions, but no 'partition_expr'"
-                        " metadata value, so we don't know what column to filter it on. Specify"
-                        " which column(s) of the database contains partitioned data as the"
-                        " 'partition_expr' metadata."
+                        " metadata value, so we don't know what column it's partitioned on. To"
+                        " specify a column, set this metadata value. E.g."
+                        ' @asset(metadata={"partition_expr": "your_partition_column"}).'
```
jamiedemaria, line 201: maybe we could also add something like "You can also do type conversions in the partition_expr, like @asset(metadata={"partition_expr": "TO_TIMESTAMP(your_partition_column)"})"

jamiedemaria (approved):
nice 🚀

python_modules/dagster/dagster/_core/storage/db_io_manager.py
sryza: I'm on the fence. On one hand, this makes the error more comprehensive. On the other hand, makes the error more verbose and therefore maybe less likely to the reader focuses on the most relevant parts.  I'm going to merge the current version for expediency, but open to that being added in the future.

PR #13537, opened 2023-04-12
Move kvs access to DaemonCursorStorage

johannkm:
I need something like this to support the AssetDaemon in cloud where the kvs is on cloud storage instead of run storage, and we've talked about moving in this direction anyway

https://github.com/dagster-io/internal/pull/5427

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/daemon_cursor.py
+++ b/python_modules/dagster/dagster/_core/storage/daemon_cursor.py
@@ -1,0 +1,12 @@
+from abc import abstractmethod
+from typing import Mapping, Set
+
+
+class DaemonCursorStorage:
+    @abstractmethod
+    def kvs_get(self, keys: Set[str]) -> Mapping[str, str]:
+        """Retrieve the value for a given key in the current deployment."""
+
+    @abstractmethod
+    def kvs_set(self, pairs: Mapping[str, str]) -> None:
+        """Set the value for a given key in the current deployment."""
--- a/python_modules/dagster/dagster/_core/storage/daemon_cursor.py
+++ b/python_modules/dagster/dagster/_core/storage/daemon_cursor.py
@@ -4,9 +4,9 @@ from typing import Mapping, Set
 
 class DaemonCursorStorage:
     @abstractmethod
-    def kvs_get(self, keys: Set[str]) -> Mapping[str, str]:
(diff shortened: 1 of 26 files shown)


johannkm:
* master
  * **PR #13537** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13537" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13537?utm_source=stack-comment).

prha:
This is more about creating a strict interface / tests for cloud than it is about cleaning up OSS, right?  Because we still need the `supports_kvs` flags and the run storage implementations for everything?

Are you thinking about splitting out the daemon cursor storage tests from the run storage tests?

johannkm:
@prha yeah it's just to unify the interface for oss and cloud. The name is the only part that's getting stricter, vs having `instance.kvs` which I think would be even more appealing for arbitrary use

prha:
While we're at it, can we extract kv tests out into its own abstract test class?  In OSS, we would just run it on `run_storage` instance.  But we could ensure the same behavior / same test suite between OSS and cloud and get rid of the standalone `test_kv.py` in cloud.

We might even be able to get rid of the `supports_kvs` flag, which I think is just to disable the run storage tests.

johannkm:
separated the tests and removed supports_kvs

python_modules/dagster/dagster/_core/storage/daemon_cursor.py
```diff
@@ -0,0 +1,12 @@
+from abc import abstractmethod
+from typing import Mapping, Set
+
+
+class DaemonCursorStorage:
+    @abstractmethod
+    def kvs_get(self, keys: Set[str]) -> Mapping[str, str]:
```
gibsondan, line 7: get_cursor_values?

python_modules/dagster/dagster/_core/storage/daemon_cursor.py
```diff
@@ -0,0 +1,12 @@
...
+        """Retrieve the value for a given key in the current deployment."""
+
+    @abstractmethod
```
gibsondan, line 10: set_cursor_values

gibsondan (approved):
excellent

prha (approved):
nicely done

PR #13539, opened 2023-04-12
Avoid rebuilding multipartitions subset when time partitions are added

clairelin135:
Currently the multipartitions subset serializable unique ID uses the [default implementation](https://github.com/dagster-io/dagster/blob/a26d6c6616451701ad88462b0e497a5e212b92f8/python_modules/dagster/dagster/_core/definitions/partition.py#L322-L329), which hashes a list of all the partition keys.

When the multipartitions def has a time dimension, this means that the serializable unique ID changes whenever new time partition keys exist, which for a `DailyPartitionsDefinition` dimension occurs every day. This causes a rebuild of the partitions status cache every time the serializable unique ID changes.

This PR amends the serializable unique ID for multipartitions defs to be hashed from a mapping of dimension name to serializable unique ID for that dimension definition. This behavior ensures that the serializable ID remains the same when a time dimension partitions def is the same, independent of the existent partition keys. When static partition keys / dynamic partition keys are added and removed from within a dimension, their corresponding serializable IDs will change, causing a rebuild of the cache as desired.

Updating this behavior will speed up page load for users especially cloud customers.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
@@ -1,3 +1,4 @@
+import hashlib
 import itertools
 from datetime import datetime
 from functools import reduce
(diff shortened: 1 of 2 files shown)


sryza (approved):
Nice

PR #13548, opened 2023-04-12
[dagit] Expose run.updateTime in the correct timezone

bengotow:
## Summary & Motivation

Slack discussion: https://elementl-workspace.slack.com/archives/C03CCE471E0/p1681241621839859

When a run is "Starting..." we pull the run.updateTime instead of the run.startTime, which is still null. It appears that this field is applying a timezone offset twice and is four hours early for a user running Dagit locally and in EDT.

This field was the subject of a bug in 0.11.5:

- The `update_timestamp` column in the runs table is now updated with a UTC timezone, making it consistent with the `create_timestamp` column.

I believe that we are correctly transforming a time in the same format to a timestamp here, and I copied this to fix:
https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_core/definitions/data_time.py#L430

## How I Tested These Changes

Before:

image

After:

image

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/schema/pipelines/pipeline.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/pipelines/pipeline.py
@@ -1,3 +1,4 @@
+import datetime
 from typing import List, Optional, Sequence
 
 import dagster._check as check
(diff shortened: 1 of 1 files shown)


bengotow:
The build failures here are unrelated 👍

PR #13550, opened 2023-04-12
Revert "Use specific events to trigger a "Preparing" state instead of assuming that any logged event means that the step has started (#13182)"

gibsondan:
This reverts commit 1ecd600325479e96f720bfd19418582ee42b5662. It seems to be struggling with dynamic outputs and incorrectly including them as "Not executed".

Test Plan: BK

### DIFF
--- a/js_modules/dagit/packages/core/src/gantt/GanttStatusPanel.tsx
+++ b/js_modules/dagit/packages/core/src/gantt/GanttStatusPanel.tsx
@@ -2,7 +2,6 @@ import {Colors, Spinner, Tooltip} from '@dagster-io/ui';
 import * as React from 'react';
 import styled from 'styled-components/macro';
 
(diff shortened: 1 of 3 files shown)


js_modules/dagit/packages/core/src/gantt/GanttStatusPanel.tsx
```diff
@@ -56,18 +52,10 @@ export const GanttStatusPanel: React.FC<GanttStatusPanelProps> = ({
...
-          break;
       }
     }
-
-    for (const node of graph) {
-      const name = node.name;
-      if (!metadata.steps[name]?.state) {
-        notExecuted.push(name);
-      }
-    }
```
gibsondan, line 68: this seems to be the source of the problem - there are nodes in the graph that don't have a state but still shouldn't be included in the list

hellendag (approved):

PR #13554, opened 2023-04-12
[ui] Consoldiate rendering of graph / op name on details page + sidebar

bengotow:
## Summary & Motivation

I noticed a slight bug in the "View graph" link in the asset sidebar showing when an `opName` was present and not specifically a `graphName`. 

The logic is kinda complicated so I pulled the rendering out of the AssetNodeDefinition into a shared component and put that in the asset sidebar.

## How I Tested These Changes

I added a few new storybooks for previewing these different scenarios.

image
image

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/__stories__/SidebarAssetInfo.stories.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/__stories__/SidebarAssetInfo.stories.tsx
@@ -1,0 +1,261 @@
+import {MockedProvider, MockedResponse} from '@apollo/client/testing';
+import {Box} from '@dagster-io/ui';
+import React from 'react';
+
+import {createAppCache} from '../../app/AppCache';
+import {buildPartitionHealthMock} from '../../assets/__fixtures__/PartitionHealthQuery.fixtures';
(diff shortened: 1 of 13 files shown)


js_modules/dagit/packages/core/src/asset-graph/__stories__/SidebarAssetInfo.stories.tsx
```diff
@@ -0,0 +1,261 @@
...
+                id: 'test.py',
+                repositoryLocationName: 'repo',
+                repositoryName: 'test.py',
+              },
+            },
+          },
+        ],
+        assetObservations: [
+          {
+            __typename: 'ObservationEvent',
```
salazarm, line 180: buildObservationEvent?

salazarm (approved):

js_modules/dagit/packages/core/src/asset-graph/__stories__/SidebarAssetInfo.stories.tsx
bengotow: Ahh man, I tried but I'm having trouble with the typescript unions for MetadataEntries conflicting. Will keep digging into that separately and see if i can figure out why, but it seems like the generators are returning instances of types that are not /quite/ what is specified by the fragment type somehow.

PR #12229, opened 2023-02-09
disable check_same_thread on in-memory sqlite storage

alangenfeld:
I tried using `dagit-debug` for the first time in a while and ran in to this check causing all web requests to fail. I think its safe enough to disable this.

### How I Tested These Changes

run `dagit-debug <debug file`>` and look at a run

PR #13555, opened 2023-04-12
in memory sqlite storage fixes

alangenfeld:
https://github.com/dagster-io/dagster/pull/12229 disabled a warning about same thread use that I thought was fine but turns out with further `dagit-debug` use it was causing `segfault`s.

To address this we
* use `cache=shared` mode on the connection string to allow multiple connections to the same in memory DB
* vend connections on demand, so that they are propery thread scoped like in the context of `dagit-debug`
* since the in memory DB deletes on last connection close, we hold on to one connection for life of instance
* we use the object id of the storage component to scope the in memory DB so concurrent ephemeral instances do not collide

## How I Tested These Changes

bk tests
open the same run many many times in `dagit-debug` which was previously causing `segfault`s

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/event_log/in_memory.py
+++ b/python_modules/dagster/dagster/_core/storage/event_log/in_memory.py
@@ -1,9 +1,8 @@
 import logging
 from collections import defaultdict
 from contextlib import contextmanager
-from typing import Callable, Optional, cast
+from typing import Callable, Optional
 
-from sqlalchemy.engine import Engine
 from sqlalchemy.pool import NullPool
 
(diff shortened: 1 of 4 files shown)


alangenfeld:
* master
  * **PR #13555** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13555" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13555?utm_source=stack-comment).

prha (approved):
the held connection is clever.

PR #13558, opened 2023-04-12
[docs] - Re-organize Cloud docs

erinkcochran87:
## Summary & Motivation

This PR splits the Cloud-focused IA changes from #12823 off into its own branch. We've received a lot of feedback about the structure of the Cloud docs in the sidenav, namely that it's too nested. We've also heard that the categories don't fit with the 'mental model' of the UI, which leads to confusion and frustration.

This PR updates the sidenav to be more in line with the UI and accomplishing specific tasks. Note that while it does address some of the nesting concerns, there are likely some design changes we could make to improve the nesting problem overall, not just here.

## How I Tested These Changes

local, 👀


slopp (approved):
🤩

PR #13567, opened 2023-04-12
[pythonic resources] Separate out resource definitions from ConfigurableResourceFactory/PartialResource

benpankow:
## Summary & Motivation

Another story for the annals of preferring composition to inheritance.

Prior to this change `ConfigurableResourceFactory`, `ConfigurableResource` et al all inherited from `ResourceDefinition`. While this made backwards compatibility convenient (you could pass configurable resources anywhere the accepted a ResourceDefinition) it was a serious error.

The largest one is that it would forever constrain us to never be able to add a public property in the ResourceDefinition, as it could potentially conflict with any public property on any resource object defined by the user. Indeed, before this PR, we already were making certain perfectly legitimate config keys illegal or not work.

For example the following caused an error:

```
class RequiredResourceKeyTestResource(ConfigurableResource):
    required_resource_keys: int


resource = IRequiredResourceKeyTestResource(required_resource_keys=3)
print(resource.required_resource_keys)
```

```NameError: Field name "required_resource_keys" shadows a BaseModel attribute; use a different field name with "alias='required_resource_keys'".```

This decouples the ConfigurableResource hierarchy from the ResourceDefinition entirely. Instead we are changing all of our entry points to mimic Definitions and coerce the resource dictionary, allowing for bare objects to be passed. Then in this diff we add another case to the coercion logic, which creates a ResourceDefinition from the ConfigurableResource instance.

### DIFF
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -405,17 +405,58 @@ def attach_resource_id_to_key_mapping(
     resource_def: Any, resource_id_to_key_mapping: Dict[ResourceId, str]
 ) -> Any:
     if isinstance(resource_def, (ConfigurableResourceFactory, PartialResource)):
+        defn = resource_def.get_resource_definition()
         return (
-            IOManagerWithKeyMapping(resource_def, resource_id_to_key_mapping)
-            if isinstance(resource_def, IOManagerDefinition)
-            else ResourceWithKeyMapping(resource_def, resource_id_to_key_mapping)
+            IOManagerWithKeyMapping(defn, resource_id_to_key_mapping)
+            if isinstance(defn, IOManagerDefinition)
+            else ResourceWithKeyMapping(defn, resource_id_to_key_mapping)
         )
     return resource_def
 
 
+from typing import Callable
+
+
+class ConfigurableResourceFactoryResourceDefinition(ResourceDefinition, AllowDelayedDependencies):
+    def __init__(
+        self,
+        resource_fn: ResourceFunction,
+        config_schema: Any,
+        description: Optional[str],
+        resolve_resource_keys: Callable[[Mapping[int, str]], AbstractSet[str]],
+    ):
+        super().__init__(
+            resource_fn=resource_fn, config_schema=config_schema, description=description
+        )
+        self._resolve_resource_keys = resolve_resource_keys
+
+    def _resolve_required_resource_keys(
+        self, resource_mapping: Mapping[int, str]
+    ) -> AbstractSet[str]:
+        return self._resolve_resource_keys(resource_mapping)
+
+
+class ConfigurableIOManagerFactoryResourceDefinition(IOManagerDefinition, AllowDelayedDependencies):
+    def __init__(
+        self,
+        resource_fn: ResourceFunction,
+        config_schema: Any,
+        description: Optional[str],
+        resolve_resource_keys: Callable[[Mapping[int, str]], AbstractSet[str]],
+    ):
+        super().__init__(
+            resource_fn=resource_fn, config_schema=config_schema, description=description
+        )
+        self._resolve_resource_keys = resolve_resource_keys
+
+    def _resolve_required_resource_keys(
+        self, resource_mapping: Mapping[int, str]
+    ) -> AbstractSet[str]:
+        return self._resolve_resource_keys(resource_mapping)
+
+
 class ConfigurableResourceFactory(
     Generic[TResValue],
-    ResourceDefinition,
     Config,
     TypecheckAllowPartialResourceInitParams,
     AllowDelayedDependencies,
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

**Comment by benpankow** [comment]
* master
  * **PR #13562** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13562" target="_blank">Graphite</a> 
    * **PR #13563** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13563" target="_blank">Graphite</a> 
      * **PR #13567** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13567" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13567?utm_source=stack-comment).

File: python_modules/dagster/dagster/_config/pythonic_config/__init__.py [hunk 1]
```diff
@@ -565,15 +638,22 @@ def _resolve_and_update_nested_resources(
...
         # Also evaluate any resources that are not partial
         resources_to_update, _ = separate_resource_params(self.__dict__)
         resources_to_update = {
-            attr_name: _call_resource_fn_with_default(resource_def, context)
-            for attr_name, resource_def in resources_to_update.items()
+            attr_name: _call_resource_fn_with_default(
+                (
+                    resource.get_resource_definition()
+                    if isinstance(resource, (ConfigurableResourceFactory, PartialResource))
+                    else resource
```
schrockn: A single helper to do this resolution step would be good

python_modules/libraries/dagster-airbyte/dagster_airbyte/asset_defs.py
```diff
@@ -566,9 +567,10 @@ def __init__(
             connection_to_freshness_policy_fn=connection_to_freshness_policy_fn,
         )
         self._workspace_id = workspace_id
-        self._airbyte_resource_def = airbyte_resource_def
-        self._airbyte_instance: AirbyteResource = airbyte_resource_def(
-            build_init_resource_context()
+        self._airbyte_instance: AirbyteResource = (
```
benpankow, line 570: change needed because we were relying on `AirbyteResource` being a `ResourceDefinition` (being able to invoke it like a fn)

python_modules/dagster/dagster/_core/host_representation/external_data.py
```diff
@@ -1545,7 +1545,13 @@ def _get_nested_resources(
             )
             for k, nested_resource in resource_def.nested_resources.items()
         }
-        if isinstance(resource_def, (ConfigurableResource, PartialResource))
+        if isinstance(
```
benpankow, line 1548: handles anon. nested resources (only in play with configurableresources)

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
```diff
@@ -106,7 +107,10 @@ def get_or_create_schedule_context(
             "positional context parameter should be provided when invoking."
         )
 
-    if any(isinstance(arg, ResourceDefinition) for arg in args):
+    if any(
+        isinstance(arg, (ResourceDefinition, ConfigurableResourceFactory, PartialResource))
+        for arg in args
+    ):
```
schrockn, line 113: is_coercible_to_resource?

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
```diff
@@ -272,12 +276,15 @@ def merge_resources(self, resources_dict: Mapping[str, Any]) -> "ScheduleEvaluat
         check.invariant(
             self._resources is None, "Cannot merge resources in context that has been initialized."
         )
+        from dagster._core.execution.build_resources import wrap_resources_for_execution
+
+        wrapped_resources_defs = wrap_resources_for_execution(resources_dict)
```
schrockn, line 281: actually having the `wrapped_resources_defs` makes the line below more confusing. I would just inline

schrockn (commented):
Ok this didn't turn out as nasty as I would have thought.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -201,6 +201,9 @@ def merge_resources(self, resources_dict: Mapping[str, Any]) -> "SensorEvaluatio
         check.invariant(
             self._resources is None, "Cannot merge resources in context that has been initialized."
         )
+        from dagster._core.execution.build_resources import wrap_resources_for_execution
+
+        wrapped_resources_defs = wrap_resources_for_execution(resources_dict)
```
schrockn, line 206: same comment here that having the variable `wrapped_resources_defs` actually obscures things

python_modules/dagster/dagster/_core/host_representation/external_data.py
schrockn: can you make a comment in the code

schrockn (approved):
Ok let's land this. Please heed final comments. Wait for https://app.graphite.dev/github/pr/dagster-io/dagster/13576/Disable-test_logging_run_status_sensor-in-preparation-for-a-fix-forward to merge before merging this

schrockn:
https://github.com/dagster-io/dagster/pull/13576 has landed

PR #13574, opened 2023-04-12
[graphql] Replace “PartitionBackfill.backfillId” with “id”

bengotow:
## Summary & Motivation

This PR replaces `PartitionBackfill.backfillId` with `PartitionBackfill.id`, both in Dagit and in our GraphQL layer.

Mutations like LaunchBackfill still return a `backfillId` for clarity.

Using a non-standard field name for an object's unique ID causes problems in our Apollo GraphQL library and requires manually telling it what it can use to uniquely key the objects. It created such a big problem for runs that we actually have both `runId` and `id` and usually request both.

Our goal is to remove all of the non-standard IDs, but it might take a handful of PRs :-) 

## How I Tested These Changes

- Run GraphQL tests and confirm they still pass
- Launch a backfill in Dagit, view the backfills page

### DIFF
--- a/js_modules/dagit/packages/core/src/app/AppCache.tsx
+++ b/js_modules/dagit/packages/core/src/app/AppCache.tsx
@@ -22,8 +22,6 @@ export const createAppCache = () =>
         return 'Instance';
       } else if (object.__typename === 'Workspace') {
         return 'Workspace';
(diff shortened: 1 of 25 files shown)


hellendag (approved):
Needs some lint fixes, but otherwise looks good to me!

PR #13578, opened 2023-04-12
[dagit] Use `Run.id` instead of `Run.runId` everywhere in Dagit

bengotow:
## Summary & Motivation

This PR replaces all uses of Run.runId with Run.id in Dagit. I'm assembling a follow-up PR with the GraphQL changes to fully remove `Run.runId`, but doing that separately because I think there could be reasons to leave it in place for longer.

Mutations results ala `LaunchRunSuccess` still return a runId for clarity.

Using a non-standard field name for an object's unique ID causes problems in our Apollo GraphQL library and requires manually telling it what it can use to uniquely key the objects. We've had both `run.runId` and `run.id` for a long time, and we usually request both.

Our goal is to remove all of the non-standard IDs, but it might take a handful of PRs :-)

## How I Tested These Changes

- Run GraphQL tests and confirm they still pass
- View runs 
- View run details view
- Launch runs in Dagit

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/AssetRunLinking.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/AssetRunLinking.tsx
@@ -33,10 +33,10 @@ export const AssetRunLink: React.FC<{
   event?: Parameters<typeof linkToRunEvent>[1];
 }> = ({runId, children, event}) => (
   <Link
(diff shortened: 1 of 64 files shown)


bengotow:
Looks like this one requires a small internal PR too

hellendag (approved):
Yes please

PR #13581, opened 2023-04-12
Make rules in MakeConfigCacheable much more strict

schrockn:
## Summary & Motivation

MakeConfigCacheable is a class that allows us attach variables directly to Pydantic classes that are otherwise immutable. Before we had a fairly blanket policy of allowing any variable starting with "_" to be set on these classes. That was very broad and it would be easier for our users to get themselves into trouble.

Here we make this much more strict and only fields ending in `__internal__` can be set on these classes. For the classes the depend on this behavior I extracted out a separate class so that only these variables exist per class.

## How I Tested These Changes

BK

### DIFF
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -113,7 +113,7 @@ class MakeConfigCacheable(BaseModel):
         return super().__setattr__(name, value)
 
     def _is_field_internal(self, name: str) -> bool:
-        return name.startswith("_") or name.endswith(CACHED_METHOD_FIELD_SUFFIX)
+        return name.endswith("__internal__") or name.endswith(CACHED_METHOD_FIELD_SUFFIX)
 
 
 class Config(MakeConfigCacheable):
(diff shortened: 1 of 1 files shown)


schrockn:
* master
  * **PR #13581** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13581" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13581?utm_source=stack-comment).

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -558,18 +571,39 @@ def __init__(self, **data: Any):
...
         }
         resolved_config_dict = config_dictionary_from_values(casted_data_without_resources, schema)
-        self._config_schema = _curry_config_schema(schema, resolved_config_dict)
 
-        # We keep track of any resources we depend on which are not fully configured
-        # so that we can retrieve them at runtime
-        self._nested_partial_resources: Mapping[str, CoercibleToResource] = {
-            k: v for k, v in resource_pointers.items() if (not _is_fully_configured(v))
-        }
+        self._state__internal__ = ConfigurableResourceFactoryState(
```
benpankow, line 575: this is pretty nice, is the goal down the line to remove the `@property` impls and reference everything through the state? or keep the property access pattern for cleanliness

benpankow (approved):

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
schrockn: We should probably get rid of them to minimize namespace pollution with user-defined properties, but I don't think it is critical.

schrockn: I don't have a super intentional goal here. This in my mind was the most expedient way to bang this out.

PR #13598, opened 2023-04-13
concept page for auto asset-materialization

sryza:
## Summary & Motivation

There's a lot more we can put in here, but I think this is a good start.

### DIFF
--- a/docs/content/concepts/assets/asset-auto-execution.mdx
+++ b/docs/content/concepts/assets/asset-auto-execution.mdx
@@ -1,0 +1,203 @@
+---
+title: Auto-Materializing Assets | Dagster
+description: You can set up Dagster to automatically materialize assets when criteria are met.
+---
+
+# Auto-Materializing Assets <Experimental />
+
+You can set up Dagster to automatically materialize assets when criteria are met. This enables a declarative approach to asset scheduling – instead of defining imperative workflows to materialize your assets, you just describe the conditions under which they should be materialized.
+
+At a high-level, there are two factors that can be used to determine when an asset is auto-materialized:
+
+1. Whether upstream data has changed
+2. Any <PyObject object="FreshnessPolicy" pluralize /> set on the asset or downstream assets
+
+Assets can be auto-materialized "eagerly" – i.e. immediately after upstream changes occur. Or they can be auto-materialized "lazily" – i.e. by waiting until downstream <PyObject object="FreshnessPolicy" pluralize /> dictate that they need to be fresh. Or a mixture of both.
+
+### Auto-materialize policies
+
+You can set up an asset to be auto-materialized by assigning it an <PyObject object="AutoMaterializePolicy" />. In this example, we use <PyObject object="AutoMaterializePolicy" method="eager" /> to indicate that, any time that `asset1` is materialized, `asset2` should be automatically materialized right after:
+
+```python file=/concepts/assets/auto_materialize_eager.py
+from dagster import AutoMaterializePolicy, asset
+
+
+@asset
+def asset1():
+    ...
+
+
+@asset(auto_materialize_policy=AutoMaterializePolicy.eager())
+def asset2(asset1):
+    ...
+```
+
+This example assumes that `asset1` will be materialized in some other way - e.g. manually, via a [sensor](/concepts/partitions-schedules-sensors/sensors), or via a [schedule](/concepts/partitions-schedules-sensors/schedules).
+
+Note that the [Dagster Daemon](/deployment/dagster-daemon) needs to be running for auto-materialization to function. If you use the `dagster dev` command to run Dagster, the daemon will automatically run as part of it, so assets will immediately start to be auto-materialized when Dagster starts.
+
+### Adding an auto-materialize policy to multiple assets at once
+
+If you want to apply the same <PyObject object="AutoMaterializePolicy" /> to a set of assets, you can use the `auto_materialize_policy` when loading them with functions like <PyObject object="load_assets_from_current_module" /> and <PyObject object="load_assets_from_package_module" />.
+
+```python file=/concepts/assets/auto_materialize_multiple.py
+from dagster import (
+    AutoMaterializePolicy,
+    Definitions,
+    asset,
+    load_assets_from_current_module,
+)
+
+
+@asset
+def asset1():
+    ...
+
+
+@asset
+def asset2(asset1):
+    ...
+
+
+defs = Definitions(
+    assets=load_assets_from_current_module(
+        auto_materialize_policy=AutoMaterializePolicy.eager(),
+    )
+)
+```
+
+### Auto-materialize policies and freshness policies
+
+Instead of auto-materializing downstream assets immediately after new upstream data arrives, you can use <PyObject object="AutoMaterializePolicy" method="lazy" /> to materialize assets only when needed to meet an asset's <PyObject object="FreshnessPolicy" />. This allows avoiding unnecessary materializations.
+
+In this example, even if `asset1` is materialized every hour, `asset2` will only be materialized roughly once per day:
+
+```python file=/concepts/assets/auto_materialize_lazy.py
+from dagster import AutoMaterializePolicy, FreshnessPolicy, asset
+
+
+@asset
+def asset1():
+    ...
+
+
+@asset(
+    auto_materialize_policy=AutoMaterializePolicy.lazy(),
+    freshness_policy=FreshnessPolicy(maximum_lag_minutes=24 * 60),
+)
+def asset2(asset1):
+    ...
+```
+
+Freshness policies express how fresh an asset needs to be relative to data at the root of the graph. This means that, for an asset to meet it's freshness policy, both it and upstream assets need to be materialized in time.
+
+Setting a lazy auto-materialize policy on an asset allows it to be auto-materialized to help downstream assets meet their freshness policies. In this example, both `asset2` and `asset3` will be auto-materialized up to once per day, to help `asset3` meet its freshness policy. Conversely, if `asset2`, did not have an auto-materialize policy, then `asset3` would never become fresh, unless `asset2` were materialized in some other way.
+
+```python file=/concepts/assets/auto_materialize_lazy_transitive.py
+from dagster import AutoMaterializePolicy, FreshnessPolicy, asset
+
+
+@asset
+def asset1():
+    ...
+
+
+@asset(auto_materialize_policy=AutoMaterializePolicy.lazy())
+def asset2(asset1):
+    ...
+
+
+@asset(
+    auto_materialize_policy=AutoMaterializePolicy.lazy(),
+    freshness_policy=FreshnessPolicy(maximum_lag_minutes=24 * 60),
+)
+def asset3(asset2):
+    ...
+```
+
+### Auto-materialize policies and data versions
+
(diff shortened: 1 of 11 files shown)


python_modules/dagster/dagster/_core/definitions/auto_materialize_policy.py
```diff
@@ -17,7 +17,7 @@ class AutoMaterializePolicy(NamedTuple):
 
     - one or more of its parents is missing
     - the partition is a time window partition with a start time that is more than
-      time_window_partition_scope older than the latest available time window partition. By
+      time_window_partition_scope_minutes older than the latest available time window partition. By
```
johannkm, line 20: do you have a sense whether `time_window_partition_scope_minutes: float` (what we store in the NT) or `time_window_partition_scope: datetime.timedelta` (the arg to lazy and eager) is better? It seems unnecessary to have them be different

johannkm: I think I like minutes to match freshness

docs/content/concepts/assets/asset-auto-execution.mdx
```diff
@@ -0,0 +1,169 @@
...
+    auto_materialize_policy=AutoMaterializePolicy.lazy(),
+    freshness_policy=FreshnessPolicy(maximum_lag_minutes=24 * 60),
+)
+def asset3(asset2):
+    ...
+```
+
+### Auto-materialize policies and data versions
+
+If you set an <PyObject object="AutoMaterializePolicy" /> on an asset that's downstream of an [observable source assets](/concepts/assets/asset-observations#observable-source-assets), then changes to the source asset will be treated as new upstream data that can cause the downstream asset to be auto-materialized.
```
johannkm, line 93: nit: singular observable source asset

docs/content/concepts/assets/asset-auto-execution.mdx
```diff
@@ -0,0 +1,169 @@
+---
+title: Auto-Materializing Assets | Dagster
+description: You can set up Dagster to automatically materialize assets when criteria are met.
+---
+
+# Auto-Materializing Assets <Experimental />
+
+You can set up Dagster to automatically materialize assets when criteria are met. This enables a declarative approach to asset scheduling – instead of defining imperative workflows to materialize your assets, you just describe the conditions under which they should be materialized.
```
slopp, line 8: I think another benefit worth mentioning here is that auto-materialization makes it easier to add new things. You don't have to worry about building one mega dag, or linking different tags together via complex schedules. Dagster automatically figures all of this out and all you have to do is write the new asset.

docs/content/concepts/assets/asset-auto-execution.mdx
```diff
@@ -0,0 +1,169 @@
...
+    auto_materialize_policy=AutoMaterializePolicy.lazy(),
+    freshness_policy=FreshnessPolicy(maximum_lag_minutes=24 * 60),
+)
+def asset2(asset1):
+    ...
+```
+
+Freshness policies express how fresh an asset needs to be relative to data at the root of the graph. This means that, for an asset to meet it's freshness policy, both it and upstream assets need to be materialized in time.
+
+Setting a lazy auto-materialize policy on an asset allows it to be auto-materialized to help downstream assets meet their freshness policies. In this example, both `asset2` and `asset3` will be auto-materialized up to once per day, to help `asset3` meet its freshness policy.
```
slopp, line 67: I think a natural question for a reader going through this guide is "do I need to add these policies to every asset?"

It might be worth making the counter point explicit here: what happens if asset 3 has a freshness policy but asset 1 or 2 does not have an auto materialization policy?

docs/content/concepts/assets/asset-auto-execution.mdx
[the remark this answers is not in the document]
slopp: Agreed - I also think it is worth clarifying why, in the example below, you need a job to observe the observable source asset. I think a reader could naturally assume the asset daemon observes assets, or maybe they would think there is some sort of AutoObservePolicy. We should call out the job for the source asset is still currently required.

docs/content/concepts/assets/asset-auto-execution.mdx
```diff
@@ -0,0 +1,169 @@
...
+
+### Auto-materialization and partitions
+
+Partitioned assets can have <PyObject object="AutoMaterializePolicy" pluralize />. Partitions are eligible for auto-materialization when either:
+
+- They are missing - i.e. have never been materialized before
+- New data has arrived in the partitions that are upstream of them
+- They fall within the `time_window_partition_scope_minutes` (more on this below)
+
+Here's a pipeline with two daily-partitioned assets that have eager auto-materialize policies. At the end of each day, a partition for that day will be added to the set of partitions for each of the assets. Dagster will notice that the new partitions exist, but have no materializations, and then auto-materialize them.
```
slopp, line 145: It may also be worth having an example of different time partitions (eg if you have hourly partitions the daemon waits till they are all available before launching the daily partition... link to some thing about partition mapping)

docs/content/concepts/assets/asset-auto-execution.mdx
```diff
@@ -0,0 +1,169 @@
...
+    partitions_def=DailyPartitionsDefinition(start_date="2020-10-10"),
+    auto_materialize_policy=AutoMaterializePolicy.eager(),
+)
+def asset2(asset1):
+    ...
+```
+
+If a partition of `asset1` is re-materialized, e.g. manually from the UI, then the corresponding partition of `asset2` will be auto-materialized after.
+
+To avoid accidentally auto-materializing enormous numbers of partitions, <PyObject object="AutoMaterializePolicy" pluralize /> have a `time_window_partition_scope_minutes`, which limits how far back in time partitions will be auto-materalized. It defaults to 24 hours. To materialize earlier partitions, launch a backfill or set `time_window_partition_scope_minutes` to a larger value.
```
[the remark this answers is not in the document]
slopp: What is the behavior for non-time partitions?

docs/content/concepts/assets/asset-auto-execution.mdx
```diff
@@ -0,0 +1,169 @@
...
+def asset1():
+    ...
+
+
+@asset(auto_materialize_policy=AutoMaterializePolicy.eager())
+def asset2(asset1):
+    ...
+```
+
+Note that the [Dagster Daemon](/some/link) needs to be running for auto-materialization to function. If you use the `dagster dev` command to run Dagster, the daemon will automatically run as part of it.
```
slopp, line 40: This daemon is pretty different from sensors and schedules which default to off and require a UI change to turn on. I think it is worth calling out that these policies will take effect _as soon as you start dagster_ and perhaps explain how you would pause them (if thats possible? I guess in prod just kill that daemon or in OSS kill dagster or push a change to the code... seems scary that we don't have a way to quickly shut down a system that is going crazy without a code push)

slopp (commented):
Looks great! I left a few minor suggestions.

My one other request would be that we add pointers to these policies from the existing reconciliation sensor page (as we deprecate it) and from the automation guide. These cross-doc breadcrumbs are really useful, but they could be added in follow on PRs.

docs/content/concepts/assets/asset-auto-execution.mdx
sryza: Is there wording you have in mind? "instead of defining imperative workflows" arguably captures this.

docs/content/concepts/assets/asset-auto-execution.mdx
sryza: I'll update the language to point out that it takes effect as soon as your start Dagster.

We're going to add a way to pausing them, but it's not in the codebase yet.

docs/content/concepts/assets/asset-auto-execution.mdx
sryza: thanks for catching

docs/content/concepts/assets/asset-auto-execution.mdx
sryza: > I think a natural question for a reader going through this guide is "do I need to add these policies to every asset?"

I'm going to add a section up here on how to add them to multiple assets at once.

I'll also incorporate your suggestion and make the counter-point more explicit.

docs/content/concepts/assets/asset-auto-execution.mdx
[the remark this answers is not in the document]
sryza: Adding both

docs/content/concepts/assets/asset-auto-execution.mdx
sryza: I'm going to leave this out for now. I think it would be useful, but doesn't need to be part of the MVP of this guide.

sryza: and I'm a little worried about overwhelming readers with a lot of examples

docs/content/concepts/assets/asset-auto-execution.mdx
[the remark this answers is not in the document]
sryza: Updated to clarify on both of these.

sryza:
@johannkm @slopp - mind taking another look? I'd like to get this in in time for 1.3.

docs/content/concepts/assets/asset-auto-execution.mdx
slopp: In the interest of brevity I rescind this recommendation

docs/content/concepts/assets/asset-auto-execution.mdx
```diff
@@ -0,0 +1,201 @@
...
+    ...
+```
+
+This example assumes that `asset1` will be materialized in some other way - e.g. manually, via a [sensor](/content/concepts/partitions-scheduled-sensors/sensors), or via a [schedule](/content/concepts/partitions-scheduled-sensors/schedules).
+
+Note that the [Dagster Daemon](/deployment/dagster-daemon) needs to be running for auto-materialization to function. If you use the `dagster dev` command to run Dagster, the daemon will automatically run as part of it, so assets will immediately start to be auto-materialized when Dagster starts.
+
+### Adding an auto-materialize policy to multiple assets at once
+
+If you want to apply the same <PyObject object="AutoMaterializePolicy" /> to a set of assets, you can use the `auto_materialize_policy` when loading them with functions like <PyObject object="load_assets_from_current_module" /> and <PyObject object="load_assets_from_package_module" />.
```
slopp, line 42: Great addition

johannkm (approved):
This looks great. There's a few things that aren't explicitly covered but I think we can add details in follow ups

- missing is only mentioned re: partitions
- it's not explicit that eager reconciliation also respects freshness policies

slopp (approved):
This looks great, thanks for flushing this out!

docs/content/concepts/assets/asset-auto-execution.mdx
```diff
@@ -0,0 +1,201 @@
...
+
+@asset(
+    auto_materialize_policy=AutoMaterializePolicy.lazy(),
+    freshness_policy=FreshnessPolicy(maximum_lag_minutes=24 * 60),
+)
+def asset2(asset1):
+    ...
+```
+
+Freshness policies express how fresh an asset needs to be relative to data at the root of the graph. This means that, for an asset to meet it's freshness policy, both it and upstream assets need to be materialized in time.
```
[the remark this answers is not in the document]
sryza: 👍 added a link

PR #13613, opened 2023-04-13
[pythonic resources] Make init context available in Pythonic resources

benpankow:
## Summary

Makes the `InitResourceContext` available to Pythonic resources through a `get_context()` getter. This can be used to retrieve e.g. instance information in the body of a resource.

```python
 class OutputDirResource(ConfigurableResource):
    output_dir: Optional[str] = None

    def get_effective_output_dir(self) -> str:
        if self.output_dir:
            return self.output_dir

        context = self.get_resource_context()
        assert context.instance
        return context.instance.storage_directory()
```

## Test Plan

Added unit test, todo a few more

### DIFF
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -753,7 +753,7 @@ class ConfigurableResourceFactory(
     def _create_object_fn(self, context: InitResourceContext) -> TResValue:
         return self.create_resource(context)
 
-    def get_context(self) -> InitResourceContext:
+    def get_resource_init_context(self) -> InitResourceContext:
         """Returns the context that this resource was initialized with."""
         return check.not_none(
             self._state__internal__.context,
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -743,17 +743,25 @@ class ConfigurableResourceFactory(
         to_update = {**resources_to_update, **partial_resources_to_update}
         return self._with_updated_values(to_update)
 
+    def with_resource_context(
+        self, context: InitResourceContext
+    ) -> "ConfigurableResourceFactory[TResValue]":
+        """Returns a new instance of the resource with the given resource init context bound."""
+        copy = self._with_updated_values({})
+        copy._state__internal__ = copy._state__internal__._replace(context=context)  # noqa: SLF001
+        return copy
+
     def initialize_and_run(self, context: InitResourceContext) -> TResValue:
-        self._state__internal__ = self._state__internal__._replace(context=context)
         with_nested_resources = self._resolve_and_update_nested_resources(context)
-        with_env_vars = with_nested_resources._resolve_and_update_env_vars()  # noqa: SLF001
+        with_context = with_nested_resources.with_resource_context(context)
+        with_env_vars = with_context._resolve_and_update_env_vars()  # noqa: SLF001
 
         return with_env_vars._create_object_fn(context)  # noqa: SLF001
 
     def _create_object_fn(self, context: InitResourceContext) -> TResValue:
         return self.create_resource(context)
 
-    def get_resource_init_context(self) -> InitResourceContext:
+    def get_resource_context(self) -> InitResourceContext:
         """Returns the context that this resource was initialized with."""
         return check.not_none(
             self._state__internal__.context,
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -649,7 +649,7 @@ class ConfigurableResourceFactory(
 
     def get_resource_definition(self) -> ConfigurableResourceFactoryResourceDefinition:
         return ConfigurableResourceFactoryResourceDefinition(
-            resource_fn=self.initialize_and_run,
+            resource_fn=self._initialize_and_run,
             config_schema=self._config_schema,
             description=self.__doc__,
             resolve_resource_keys=self._resolve_required_resource_keys,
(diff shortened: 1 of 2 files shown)


benpankow:
* master
  * **PR #13613** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13613" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13613?utm_source=stack-comment).

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -706,6 +713,13 @@ def initialize_and_run(self, context: InitResourceContext) -> TResValue:
     def _create_object_fn(self, context: InitResourceContext) -> TResValue:
         return self.create_resource(context)
 
+    def get_context(self) -> InitResourceContext:
```
schrockn, line 716: So one point of potential confusion here is that there will be a get_context() method in contexts where an OpExecutionContext is also available

e.g.

```
def an_asset(context: OpExecutionContext, a_resource: AResource):
   a_resource.get_context() != context
```

Maybe we should call it `get_resource_context` or `get_context_for_resource`

schrockn: There are definitely problems with those names too

schrockn (changes requested):
Back to you to discuss `get_context` name

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
jamiedemaria: are there cases when you would want/need to use the resource's context in an asset/op/etc? If not, then it could just be a private `_get_context` (or `_get_resource_context` etc) method right?

schrockn: I don't think that users will actually call `get_context` in the body of the op. It's more about the conceptual confusion, which I still think would exist with a private `_get_context` method. Meaning that if I type that code, which context am I getting? Is it the same as the one passed to the op? It's already confusion and this risks making it more confusing.

jamiedemaria: cool, that all makes sense to me!

schrockn: `get_resource_init_context` potentially confusing as it is no longer access in an initializing context. It is accessed during execution

benpankow:
Updated to simply `get_resource_context`. Added a `with_resource_context` utility for testing purposes to let a user bind a context to a direct-instantiation resource as well.

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -583,6 +584,7 @@ def __init__(self, **data: Any):
             config_schema=_curry_config_schema(schema, resolved_config_dict),
             schema=schema,
             nested_resources={k: v for k, v in resource_pointers.items()},
+            context=None,
```
schrockn, line 587: `resource_context` to be consistent with this?

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -697,15 +703,31 @@ def _resolve_and_update_nested_resources(
         to_update = {**resources_to_update, **partial_resources_to_update}
         return self._with_updated_values(to_update)
 
+    def with_resource_context(
+        self, context: InitResourceContext
```
schrockn, line 707: resource_context

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -697,15 +703,31 @@ def _resolve_and_update_nested_resources(
...
+        """Returns a new instance of the resource with the given resource init context bound."""
+        copy = self._with_updated_values({})
+        copy._state__internal__ = copy._state__internal__._replace(context=context)  # noqa: SLF001
+        return copy
+
     def initialize_and_run(self, context: InitResourceContext) -> TResValue:
         with_nested_resources = self._resolve_and_update_nested_resources(context)
-        with_env_vars = with_nested_resources._resolve_and_update_env_vars()  # noqa: SLF001
+        with_context = with_nested_resources.with_resource_context(context)
+        with_env_vars = with_context._resolve_and_update_env_vars()  # noqa: SLF001
```
schrockn, line 717: seems like this is a spot where we can do a one liner and it would be better

schrockn (changes requested):
being anal here, but let's change context to resource_context everywhere where appropriate, even in internals

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -737,15 +743,35 @@ def _resolve_and_update_nested_resources(
         to_update = {**resources_to_update, **partial_resources_to_update}
         return self._with_updated_values(to_update)
 
+    def with_resource_context(
+        self, resource_context: InitResourceContext
+    ) -> "ConfigurableResourceFactory[TResValue]":
+        """Returns a new instance of the resource with the given resource init context bound."""
+        copy = self._with_updated_values({})
```
schrockn, line 750: this looks odd enough to be comment worthy

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -737,15 +743,35 @@ def _resolve_and_update_nested_resources(
...
+        copy._state__internal__ = copy._state__internal__._replace(  # noqa: SLF001
+            resource_context=resource_context
+        )
+        return copy
+
     def initialize_and_run(self, context: InitResourceContext) -> TResValue:
```
schrockn, line 756: not your diff, but should we make this private since we don't want users to call this directly?

schrockn (approved):

PR #13614, opened 2023-04-13
Backfill page dogfooding feedback

salazarm:
## Summary & Motivation

- Make toast to "View" backfill into an href so that it can be command click opened in a separate tab
- Link to the asset detail page from the asset name in the table
- Add the query refresh countdown component to the right side of the page header
- Change "In Progress" -> "Requested" to keep terminology consistent
- Moved links to view completed/failed/targeted/canceled runs into the table header cells because I don't know of a way that we can filter runs by the assets they materialize 
- Fix partition names being cut off
- Fixed the hardcoded start time
- 

## How I Tested These Changes
Screen Shot 2023-04-13 at 2 53 06 PM


bengotow (approved):
This looks great!

sryza:
Awesome

PR #13620, opened 2023-04-13
Revert "Revert "Use specific events to trigger a "Preparing" state instead of assuming that any logged event means that the step has started (#13182)" (#13550)"

gibsondan:
This reverts commit b6ae36f5b002ac6f79af42a2b630f6e0cd45e55d. To fix the issue we were seeing where [?] steps were sticking around in the RHC, filter them out from the list of unexecuted steps (they never execute, so this should be fine).

Test Plan:
Run the following graph - see the ops on the right appear but no lingering [?] ops remain at the end:

```
@op(
    config_schema={
        "first": Field(IntSource, is_required=False, default_value=1),
        "second": Field(IntSource, is_required=False, default_value=1),
        "how_many": Field(IntSource, is_required=False, default_value=10),
    },
    out=DynamicOut(int),
)
def chase_giver(context):
    chase_duration = context.op_config["first"]

    for i in range(context.op_config["how_many"]):
        yield DynamicOutput(chase_duration, mapping_key=str(i))

@op
def reducer(_, nums):
    return 16

@op
def map_me(_, num):
    return num

@graph(tags={"test": "please"})
def hammer_graph():
    outs = chase_giver().map(map_me)
    reducer(outs.collect())
```


gibsondan:
@bengotow can i request a look at this when you have a chance? I don't think it needs to be cherry-picked or anything

bengotow (approved):
This change looks good to me! The addition of the isPlannedDynamicStep check makes sense there I think 👍

PR #13623, opened 2023-04-13
[docs] Add section on Serverless page about I/O Managers

jming422:
## Summary & Motivation

When I setup a Serverless deployment it never occurred to me to set up my own I/O manager, nor that I wouldn't be able to download asset outputs without doing so. After chatting about it with @tacastillo in [this thread in #dagster-support](https://dagster.slack.com/archives/C01U954MEER/p1681336470898429), I learned that setting up an I/O manager is recommended in general for production, and is the best way to ensure I could access the data produced by my assets and/or ops.

I also noticed in the docs for `s3_pickle_io_manager` that the way to configure the bucket it should use is to use its `.configured()` method, but I felt like I had to do more searching than I wanted to discover that fact, so I added calls to `.configured()` with a custom bucket to the example config snippets on the IO Managers page, since I imagine this is the most common use case for the S3 I/O manager.

## How I Tested These Changes

I configured my own Serverless deployment to use the S3 pickle I/O manager, then took what I learned on Slack and from experience to write this note. Then I ran it locally to edit/reword, and now here we are!

### DIFF
--- a/docs/content/concepts/io-management/io-managers.mdx
+++ b/docs/content/concepts/io-management/io-managers.mdx
@@ -100,7 +100,12 @@ def downstream_asset(upstream_asset):
 
 defs = Definitions(
     assets=[upstream_asset, downstream_asset],
-    resources={"io_manager": s3_pickle_io_manager, "s3": s3_resource},
+    resources={
+       "io_manager": s3_pickle_io_manager.configured(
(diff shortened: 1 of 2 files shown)


tacastillo:
Thanks for giving back to the community and contributing this! I (and/or Erin, and maybe someone from the Cloud team) will take a look at this.

PedramNavid:
Closing this stale PR.

PR #13629, opened 2023-04-13
Add name field to databricks ops

clairelin135:
User reports a `Detected conflicting node definitions with the same name` error when multiple databricks ops exist in the same repository:
https://dagster.slack.com/archives/C01U954MEER/p1681412625636049

This PR threads the `name` arg through to enable this use case.


clairelin135:
Aliasing is insufficient in this case--aliasing is only allowed when the node definition is the same. If the node definition is not the same, we assert that the node definition names must be unique: https://github.com/dagster-io/dagster/blob/a32938afc4ec1058d78de2ba06ee3381cf9bb48c/python_modules/dagster/dagster/_core/definitions/composition.py#L315-L318

sryza (approved):
It's worth adding this to the docstring. Otherwise, looks great.

PR #13634, opened 2023-04-14
[ui] Organize js_modules src

hellendag:
## Summary & Motivation

Try to establish consistency about where we put fixtures, stories, and tests: directories alongside implementation files, named `__fixtures__`, `__stories__`, and `__tests__`.

I had been kind of resistant to this for a while, but now that we're trying to ramp up our test and story coverage, it has begun to feel much cleaner to separate these pieces out into their own spaces.

Open to thoughts on this.

## How I Tested These Changes

Buildkite


js_modules/dagit/packages/core/jest.config.js
```diff
@@ -148,10 +148,7 @@ module.exports = {
   // testLocationInResults: false,
 
   // The glob patterns Jest uses to detect test files
-  // testMatch: [
-  //   "**/__tests__/**/*.[jt]s?(x)",
-  //   "**/?(*.)+(spec|test).[tj]s?(x)"
-  // ],
+  testMatch: ['**/__tests__/**/*.test.[jt]s?(x)'],
```
hellendag, line 151: This enforces that only tests in `__tests__` directories will be run, which hopefully serves as a stick to ensure that we're consistent about where we put test files and how we name them.

bengotow (approved):

js_modules/dagit/packages/core/jest.config.js
hellendag: Yeah, same for stories -- let's do this in a followup

PR #13547, opened 2023-04-12
Supporting files guide

odette-elementl:

PR #13635, opened 2023-04-14
[docs] - Dagster project file reference

erinkcochran87:
## Summary & Motivation

This PR adds a reference for files in Dagster projects, including standard scaffolded files and optional, manually-added files like `dagster.yaml`. It builds on some of the work Odette did in #13692 and #13547.

There are some scattered pieces of content that touch on this topic, but nothing that's a one-stop-shop. The goal of this reference is to explain the files in a project and provide a jumping-off point for more detailed content, like how to structure your project or understand the settings in a configuration file.

Preview: https://dagster-p5slcysnx-elementl.vercel.app/getting-started/project-file-reference

**Note**: Ideally the `dagster_cloud.yaml` PR would land before this one, so I can add links in appropriate places. They're not present here to prevent angering Buildkite tests.

To-do:

- [x] Update the **Create new project** guide
- [x] Update the **Structuring your project** guide

## How I Tested These Changes

:eyes:, local


docs/content/getting-started/project-file-reference.mdx
```diff
@@ -0,0 +1,458 @@
+---
+title: Dagster project file reference | Dagster Docs
+description: ""
+---
+
+# Dagster project file reference
+
+Following [our recommended best practices for project structure](/guides/dagster/recommended-project-structure), a small Dagster project might look like this:
```
odette-elementl, line 8: We have the best practices for project structure doc, i worry that duplicating it here might get confused with you only need these files and the point of this doc is to say when you need different stuff

docs/content/getting-started/project-file-reference.mdx
```diff
@@ -0,0 +1,458 @@
...
+│   │   └── slack_on_failure_sensor.py
+│   └── utils
+├── my_dagster_project_tests
+├── pyproject.toml
+├── setup.cfg
+├── setup.py
+└── tox.ini
+```
+
+Let's take a look at what each of these files and directories does:
```
odette-elementl, line 42: seems to be what the recommended project structure aimed to do as well

docs/content/getting-started/project-file-reference.mdx
```diff
@@ -0,0 +1,458 @@
...
+        An ini file that contains option defaults for <code>setup.py</code>{" "}
+        commands.
+      </td>
+    </tr>
+  </tbody>
+</table>
+
+---
+
+## Configuration files
```
odette-elementl, line 122: i would jump directly to this for this specific guide

docs/content/getting-started/project-file-reference.mdx
```diff
@@ -0,0 +1,438 @@
...
+      </td>
+      <td>Optional</td>
+      <td>n/a</td>
+    </tr>
+  </tbody>
+</table>
+
+---
+
+## Example project structures
```
erinkcochran87, line 278: Reviewers - I'm not sure about keeping this section **in this guide**, but I do think it's useful info to have somewhere. Would love your input!

docs/content/getting-started/project-file-reference.mdx
```diff
@@ -0,0 +1,438 @@
...
+        <a href="/concepts/code-locations">Code locations documentation</a> to
+        learn more.
+        <br />
+        <br /> <strong>Note:</strong> <code>pyproject.toml</code> was introduced
+        in{" "}
+        <a href="https://peps.python.org/pep-0518/https://peps.python.org/pep-0518/">
+          PEP-518
+        </a>{" "}
+        and meant to replace <code>setup.py</code>, but we may still include a{" "}
+        <code>setup.py</code> for compatibility with tools that do not use this
```
slopp, line 144: Our current examples all use setup.py for dependencies, and deploying to cloud does not support specifying dependencies using pyproject.toml.

Maybe we could change the language a bit to replace this note and change say setup.py is used to specify dependencies?

docs/content/getting-started/project-file-reference.mdx
```diff
@@ -0,0 +1,438 @@
...
+  </tbody>
+</table>
+
+---
+
+## Configuration files
+
+<Note>
+  <strong>Not sure where to put these files?</strong> Unless otherwise noted,
+  all configuration files should be placed in the root of your Dagster project
```
slopp, line 173: See comment, but I don't think this is generally true for most non-local dagster deployments. The recommendation would probably be to place these files (that control global dagster settings) alongside your infrastructure-as-code definitions. Though `dagster_cloud.yaml` would normally live in the project root.

docs/content/getting-started/project-file-reference.mdx
```diff
@@ -0,0 +1,438 @@
...
+      >
+        Cloud
+      </th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>dagster.yaml</td>
+      <td>
+        Configures your Dagster instance, including defining storage locations,
```
slopp, line 215: Most hybrid agents are not configured directly by a dagster.yaml file. Instead, the k8s agent is configured through a helm chart that, in turn, creates a dagster.yaml file (via a ConfigMap) and ECS is configured through a CloudFormation template that generated a dagster.yaml file as part of the agent task definition.

docs/content/getting-started/project-file-reference.mdx
```diff
@@ -0,0 +1,438 @@
...
+          customize the Hybrid agent
+        </a>.
+      </td>
+      <td>Optional</td>
+      <td>Optional</td>
+    </tr>
+    <tr>
+      <td>dagster_cloud.yaml</td>
+      <td>
+        Defines multiple code locations for Dagster Cloud. For Hybrid
```
slopp, line 233: Strike "For Hybrid deployments this file can be used to manage environment variables/secrets". While partially true, I think that line is a confusing way to describe what `dagster_cloud.yaml` is for. Instead I would just leave it at "Defines code locations for Dagster Cloud"

docs/content/getting-started/project-file-reference.mdx
```diff
@@ -0,0 +1,438 @@
...
+        deployments, this file can be used{" "}
+        <a href="/dagster-cloud/managing-deployments/setting-environment-variables-dagster-cloud-agents">
+          to manage environment variables/secrets
+        </a>
+        .
+      </td>
+      <td>n/a</td>
+      <td>Optional</td>
```
slopp, line 241: This file should be marked as Required. There are cases where the file is not strictly required (if you configure a code location through the UI or if you run `dagster-cloud` CLI commands with no file), but in practice I think every Cloud deployment uses this file and listing it as optional is confusing. If Required is too strong, than I'd favor Recommended over Optional.

slopp (changes requested):

docs/content/getting-started/project-file-reference.mdx
slopp: Keep it!

slopp:
I really like the structure of this guide and the use of tables  + tabs for examples keeps it organized and easy to read without being too nested.

I have two recommended changes:

(1) We should standardize on whether we want to refer to Python packages as modules or packages. This is most clear in the section describing `pyproject.toml` where we refer to python _package_ in the docs but _module_ in the setting. I had thought our general plan was to refer to everything as a Python module, though  our functions `load_assets_from_package_module` is  confusing in light of that choice.

(2) This guide seems to suggest that some global configuration files would live inside a project directory. I'm not actually sure if that is true in most cases?  I like that the doc highlights the idea that you develop locally with one set of concerns,  and then when you deploy you have additional config files to possibly be concerned with, but I'd guess most of the time `dagster.yaml`, `deployment_settings.yaml`, and sometimes even `workspace.yaml` would be managed globally outside of a project dir.

I left a few other minor nits. Back to you!

docs/content/getting-started/project-file-reference.mdx
```diff
@@ -0,0 +1,438 @@
...
+    </tr>
+    <tr>
+      <td>pyproject.toml</td>
+      <td>
+        A file that specifies package core metadata in a static, tool-agnostic
+        way.
+        <br />
+        <br />
+        This file includes a <code>tool.dagster</code> section which references the
+        Python package with your Dagster definitions defined and discoverable at
```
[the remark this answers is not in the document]
erinkcochran87: Thanks for the catch! Your comment about module vs package is spot on - I'll make the changes throughout this doc

docs/content/getting-started/project-file-reference.mdx
erinkcochran87: Is your comment referring to "Configures your Dagster instance", or the bit about "Using this file to customize a Hybrid agent"?

erinkcochran87:
> I'd guess most of the time dagster.yaml, deployment_settings.yaml, and sometimes even workspace.yaml would be managed globally outside of a project dir.

@slopp Do you have recommendations for how to convey this? I don't know what this would look like, or how to frame it for folks.

docs/content/getting-started/project-file-reference.mdx
erinkcochran87: Not updated yet - see other comment: https://github.com/dagster-io/dagster/pull/13635#issuecomment-1522066388

docs/content/getting-started/project-file-reference.mdx
slopp: I meant that "Using this file" feels incorrect, since most people do not literally use dagster.yaml. Perhaps we can just add a note that this file is often configured indirectly through helm or cloud formation?

slopp:
@erinkcochran87  I think I mostly would just get rid of the "Not sure where to put this file?" note in the Configuration files section, and then in the example project section I would maybe do something like

```
# your project location
|_ project_stuff
# your infrastructure-as-code 
|_dagster_deployment
|__dagster.yaml
```

Just so that it is clear those global configuration files might (typically even) live in a different part of the folder tree

erinkcochran87:
@slopp this is back to you!

slopp (approved):
LGTM thanks!

PR #13641, opened 2023-04-14
[serdes] pack optimizations

alangenfeld:
* inline some functions on hot paths based on profiler inspection 
* re-order the if checks in `_pack_value`

## How I Tested These Changes
before
```
Target 4 size 1.8MiB:
serialize_value   (25 samples): mean 78.470, stdev 5.006, min 71.353

Target 5 size 23.8MiB:
serialize_value   (25 samples): mean 1037.819, stdev 8.261, min 1028.644
```
after
```
Target 4 size 1.8MiB:
serialize_value   (25 samples): mean 53.055, stdev 1.820, min 51.226

Target 5 size 23.8MiB:
serialize_value   (25 samples): mean 822.777, stdev 16.007, min 808.087
```


alangenfeld:
An average end user of dagit is unlikely to notice. Some users who have large enough entities or volumes of entities maybe. 

Users who are monitoring where cpu cycles are being spent in aggregate (like us and our hosted multitenant service) will definitely notice.

PR #13645, opened 2023-04-14
[wip][pythonic config] Add support for pydantic.SecretStr

benpankow:
## Summary

Adds support for Pydantic's [`SecretStr`](https://docs.pydantic.dev/usage/types/#secret-types) data type, which when directly accessed or converted to JSON appears as asterisks. 

The config system treats it as a normal string source for purposes of accepting configuration - the only difference here is accessing the value from the config or resource object, in which case the user needs to call `.get_secret_value()` to access the underlying value.

Once we have a generic solution up for "secret" config values, we can make it so `SecretStr` is automatically treated as such.

## Test Plan

New unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -18,7 +18,7 @@ from typing import (
     cast,
 )
 
-from pydantic import ConstrainedFloat, ConstrainedInt, ConstrainedStr
+from pydantic import ConstrainedFloat, ConstrainedInt, ConstrainedStr, SecretField, SecretStr
 from typing_extensions import TypeAlias, TypeGuard, get_args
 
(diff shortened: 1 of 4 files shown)


benpankow:
* master
  * **PR #13645** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13645" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13645?utm_source=stack-comment).

schrockn (changes requested):
So I'm a bit skeptical that we should support this until we have our plan settled for secrets. Otherwise this feels like supporting additional surface area without much benefit. Indeed it might actually be a bit misleading since this wouldn't lead to any in-product support (e.g. obscuring persisted configuration values or behavior in the launchpad)

PR #13648, opened 2023-04-14
improvements to automated docs screenshotting

sryza:
## Summary & Motivation

- Use `dagster dev` instead of `dagit` (otherwise runs don't ever get launched, because Dagster uses the queued run coordinator by default)
- Make the docs a little clearer
- Auto-format `dagit_screenshot/utils`
- Ignore DAGSTER_HOME in the environment

## How I Tested These Changes

Updated some screenshots on https://docs.dagster.io/guides/dagster/asset-versioning-and-caching#asset-versioning-and-caching

### DIFF
--- a/docs/dagit-screenshot/README.md
+++ b/docs/dagit-screenshot/README.md
@@ -27,7 +27,7 @@ screenshot specs.
 
 Specs have the following fields. Only the `id` field requires explicit specification for all specs:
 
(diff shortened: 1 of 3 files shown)


smackesey:
very nice

>Auto-format dagit_screenshot/utils

Any idea why this isn't being handled by black already

smackesey (approved):

sryza:
> Any idea why this isn't being handled by black already

It looks like `make black` doesn't run on the `docs` directory

PR #13649, opened 2023-04-14
[pythonic config] More explicit error message for setting fields on Config/ConfigurableResource

benpankow:
## Summary

Adds a more explicit/actionable error message for trying to set fields on a `ConfigurableResource`. Here, we prompt users to build a separate, stateful client class.

Previously, users would see the Pydantic error:
```
E   TypeError: "MyResource" is immutable and does not support item assignment
```

Now, they see

```
E   DagsterInvalidInvocationError: 'MyResource' is a Pythonic resource and does not support item assignment. 
    If trying to set state on this resource, consider building a separate, stateful client class.
```

### DIFF
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -118,19 +118,23 @@ class MakeConfigCacheable(BaseModel):
 
         try:
             return super().__setattr__(name, value)
-        except TypeError as e:
-            if "is immutable and does not support item assignment" in str(e):
+        except (TypeError, ValueError) as e:
+            if "is immutable and does not support item assignment" in str(
+                e
+            ) or "object has no field" in str(e):
                 clsname = self.__class__.__name__
                 if isinstance(self, ConfigurableResourceFactory):
                     raise DagsterInvalidInvocationError(
-                        f"'{clsname}' is a Pythonic resource and does not support item assignment."
-                        " If trying to set state on this resource, consider building a separate,"
-                        " stateful client class."
+                        f"'{clsname}' is a Pythonic resource and does not support item assignment,"
+                        " as it inherits from 'pydantic.BaseModel' with frozen=True. If trying to"
+                        " maintain state on this resource, consider building a separate, stateful"
+                        " client class, and provide a method on the resource to construct and"
+                        " return the stateful client."
                     ) from e
                 else:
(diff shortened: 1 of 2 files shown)


benpankow:
* master
  * **PR #13649** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13649" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13649?utm_source=stack-comment).

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -115,7 +116,24 @@ def __setattr__(self, name: str, value: Any):
...
+        try:
+            return super().__setattr__(name, value)
+        except TypeError as e:
+            if "is immutable and does not support item assignment" in str(e):
+                clsname = self.__class__.__name__
+                if isinstance(self, ConfigurableResourceFactory):
+                    raise DagsterInvalidInvocationError(
+                        f"'{clsname}' is a Pythonic resource and does not support item assignment."
+                        " If trying to set state on this resource, consider building a separate,"
+                        " stateful client class."
```
schrockn, line 128: I would mention:

1) I would mention how this relates to Pydantic.(e.g. this inherits from BaseModel with frozen=true)
2) I would be a little more prescriptive. Explain how the pattern is have a method on the ConfigurableResource that constructs the other object. 

Also consider catching ValueError here to detect the assignment of undeclared attributes:

```
class TestModel(Config):
    def set_something(self):
        self._something = "something"


model = TestModel()
model.set_something()
```

```ValueError: "TestModel" object has no field "_something"```

schrockn (changes requested):

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
benpankow: updated error message + added valueerror catching

schrockn (changes requested):
So one minor change requested.

The two cases (TypeError and ValueError) are different enough IMO to warrant two different error messages.

In one case, we disallowing mutation of _declared_ attributes. This is the one that is because we made `frozen=True`.

In the other case, assignment is failing on an undeclared attribute. This would happen whether it is frozen or not.

Pydantic is a well-known library and if we are not precise we may confuse certain users and reduce our credibility.

PR #13656, opened 2023-04-14
[ui] Fix half-clipped headers on repo-scoped asset table

bengotow:
## Summary & Motivation

This PR renames the asset catalog table stories and adds a story for the repo-scoped asset table, which has slightly different columns and folding behavior. It fixes an issue with the repo-scoped asset table defining one too many columns and adjusts the column widths slightly.

Thankfully these use the same virtualized rows and the storybooks are almost the same.

Note:  I also added an "id" value to some Workspace mocks, because the storybook was throwing exceptions without an id. I'm not 100% sure all of the places I added it are required, but I think specifying a constant value is correct and we shouldn't let our old faker-based Apollo mocks randomize the value.

## How I Tested These Changes

image
image

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/__fixtures__/AssetTables.fixtures.ts
+++ b/js_modules/dagit/packages/core/src/assets/__fixtures__/AssetTables.fixtures.ts
@@ -7,6 +7,7 @@ import {
   buildAssetNode,
   buildRepositoryLocation,
   buildRepository,
+  buildAssetKey,
 } from '../../graphql/types';
 import {SINGLE_ASSET_QUERY} from '../../workspace/VirtualizedAssetRow';
 import {SingleAssetQuery} from '../../workspace/types/VirtualizedAssetRow.types';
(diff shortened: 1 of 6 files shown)


js_modules/dagit/packages/core/src/assets/__fixtures__/AssetTables.fixtures.ts
```diff
@@ -309,6 +309,97 @@ export const SingleAssetQueryLastRunFailed: MockedResponse<SingleAssetQuery> = {
...
+      hasMaterializePermission: true,
+      description:
+        'This is a super long description that could involve some level of SQL and is just generally very long',
+      repository,
+      __typename: 'AssetNode',
+    }),
+  },
+  {
+    id: 'test.py.repo.["late_asset"]',
+    __typename: 'Asset',
```
hellendag, line 354: You could probably use `buildAsset` for these, then you won't need the `__typename`

hellendag (approved):

js_modules/dagit/packages/core/src/assets/__fixtures__/AssetTables.fixtures.ts
bengotow: Oh good call I'll upgrade these while I'm here

js_modules/dagit/packages/core/src/assets/__fixtures__/AssetTables.fixtures.ts
```diff
@@ -309,6 +309,97 @@ export const SingleAssetQueryLastRunFailed: MockedResponse<SingleAssetQuery> = {
...
+      __typename: 'AssetKey',
+    },
+    definition: buildAssetNode({
+      id: 'test.py.repo.["run_failing_asset"]',
+      groupName: 'GROUP4',
+      partitionDefinition: null,
+      description: null,
+      hasMaterializePermission: true,
+      repository,
+      __typename: 'AssetNode',
```
salazarm, line 382: you don't need to the __typename

js_modules/dagit/packages/core/src/assets/__fixtures__/AssetTables.fixtures.ts
```diff
@@ -309,6 +309,97 @@ export const SingleAssetQueryLastRunFailed: MockedResponse<SingleAssetQuery> = {
...
+    }),
+  },
+  {
+    id: 'test.py.repo.["asset_with_a_very_long_key_that_will_require_truncation"]',
+    __typename: 'Asset',
+    key: {
+      path: ['asset_with_a_very_long_key_that_will_require_truncation'],
+      __typename: 'AssetKey',
```
salazarm: buildAssetKey({

js_modules/dagit/packages/core/src/assets/__fixtures__/AssetTables.fixtures.ts
salazarm: in vscode if you use regex find:
```
\s\{[\s\r]*(__typename:\s*'([^']*)')(,)?
```
replace:
```
 build$2({
```

PR #13657, opened 2023-04-14
RFC [postgres] move to SqlPollingEventWatcher

alangenfeld:


alangenfeld:
* master
  * **PR #13657** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13657" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13657?utm_source=stack-comment).

ankit-gautam23 (commented):
Could you plz add something how you tested it? It would be easy for us people to review.

python_modules/libraries/dagster-postgres/dagster_postgres/event_log/event_log.py
```diff
@@ -181,6 +180,8 @@ def store_event(self, event: EventLogEntry) -> None:
             )
             res = result.fetchone()
             result.close()
+
+            # LISTEN/NOTIFY no longer used for pg event watch - preserved here to support version skew
```
gibsondan, line 184: ah I see - old dagit version reading from the websocket, new user code writing to the DB? fun, sad that we need to still do this for the forseeable future. Is it at all expensive if nothing is listening?

gibsondan (approved):

python_modules/libraries/dagster-postgres/dagster_postgres/event_log/event_log.py
alangenfeld: from what ive come to understand from the docs the cost is ~ zero for notify with no listeners

prha (approved):

PR #13673, opened 2023-04-17
[docs][pythonic config] Update IO management page

benpankow:
## Summary

Updates the IO management concepts page to (mostly) use configurable IO managers. Provides some guidance on using ConfigurableIOManager vs the factory.

Doesn't cut over the input/output schema examples, not sure the best way to handle those.

### DIFF
--- a/docs/content/concepts/io-management/io-managers.mdx
+++ b/docs/content/concepts/io-management/io-managers.mdx
@@ -18,12 +18,13 @@ IO Managers are user-provided objects that store asset and op outputs and load t
 
 ## Relevant APIs
 
-| Name                                                        | Description                                                                                                                                                                                                 |
-| ----------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| <PyObject module="dagster" object="io_manager" decorator /> | A decorator used to define IO managers.                                                                                                                                                                     |
-| <PyObject module="dagster" object="IOManager" />            | Base class for user-provided IO managers.                                                                                                                                                                   |
-| <PyObject object="build_input_context"/>                    | Function for directly constructing a <PyObject object="InputContext"/>, to be passed to the <PyObject object="IOManager" method="load_input"/> method. This is designed primarily for testing purposes.     |
-| <PyObject object="build_output_context"/>                   | Function for directly constructing a <PyObject object="OutputContext"/>, to be passed to the <PyObject object="IOManager" method="handle_output"/> method. This is designed primarily for testing purposes. |
+| Name                                                                          | Description                                                                                                                                                                                                 |
+| ----------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| <PyObject module="dagster" object="ConfigurableIOManager" decorator />        | A base class used to define configurable IO managers which are also resources.                                                                                                                              |
+| <PyObject module="dagster" object="ConfigurableIOManagerFactory" decorator /> | A base class used to specify configuration for more advanced I/O managers.                                                                                                                                  |
+| <PyObject module="dagster" object="IOManager" />                              | Base class for user-provided IO managers.                                                                                                                                                                   |
+| <PyObject object="build_input_context"/>                                      | Function for directly constructing a <PyObject object="InputContext"/>, to be passed to the <PyObject object="IOManager" method="load_input"/> method. This is designed primarily for testing purposes.     |
+| <PyObject object="build_output_context"/>                                     | Function for directly constructing a <PyObject object="OutputContext"/>, to be passed to the <PyObject object="IOManager" method="handle_output"/> method. This is designed primarily for testing purposes. |
 
 ## Overview
 
@@ -320,31 +321,59 @@ def a_job():
 
 If you have specific requirements for where and how your outputs should be stored and retrieved, you can define your own IO manager. This boils down to implementing two functions: one that stores outputs and one that loads inputs.
 
-To define an IO manager, use the <PyObject module="dagster" object="io_manager" displayText="@io_manager" /> decorator.
+To define an IO manager, you will want to extend the <PyObject module="dagster" object="IOManager" /> class. Often times, you will want to extend the <PyObject module="dagster" object="ConfigurableIOManager"/> class (which subclasses `IOManager`) to attach a config schema to your IO manager.
+
+Here, we define a simple IO manager that reads and writes CSV values to the filesystem. It takes an optional prefix path through config.
 
 ```python file=/concepts/io_management/custom_io_manager.py startafter=start_io_manager_marker endbefore=end_io_manager_marker
-from dagster import IOManager, io_manager
+from dagster import ConfigurableIOManager
 
 
-class MyIOManager(IOManager):
+class MyIOManager(ConfigurableIOManager):
+    path_prefix: List[str] = []
+
     def _get_path(self, context) -> str:
-        return "/".join(context.asset_key.path)
+        return "/".join(self.path_prefix + context.asset_key.path)
 
     def handle_output(self, context, obj):
         write_csv(self._get_path(context), obj)
 
     def load_input(self, context):
         return read_csv(self._get_path(context))
-
-
-@io_manager
-def my_io_manager():
-    return MyIOManager()
 ```
 
-The <PyObject module="dagster" object="io_manager" /> decorator behaves nearly identically to the <PyObject module="dagster" object="resource" /> decorator. It yields an <PyObject module="dagster" object="IOManagerDefinition" />, which is a subclass of <PyObject module="dagster" object="ResourceDefinition" /> that will produce an <PyObject module="dagster" object="IOManager" />.
-
 The provided `context` argument for `handle_output` is an <PyObject module="dagster" object="OutputContext" />. The provided `context` argument for `load_input` is an <PyObject module="dagster" object="InputContext" />. The linked API documentation lists all the fields that are available on these objects.
+
+### Using an IO manager factory
+
+If your IO manager is more complex, or needs to manage internal state, it may make sense to split out the IO manager definition from its configuration. In this case, you can use `ConfigurableIOManagerFactory`, which specifies config schema and implements a factory function that takes the config and returns an IO manager.
+
+```python file=/concepts/io_management/custom_io_manager.py startafter=start_io_manager_factory_marker endbefore=end_io_manager_factory_marker
+from dagster import IOManager, ConfigurableIOManagerFactory
+import requests
+
+
+class ExternalIOManager(IOManager):
+    def __init__(self, api_token):
+        self._api_token = api_token
+
+    def handle_output(self, context, obj):
+        ...
+
+    def load_input(self, context):
+        ...
+
+
+class ConfigurableExternalIOManager(ConfigurableIOManagerFactory):
+    username: str
+    password: str
+
+    def create_io_manager(self, context) -> ExternalIOManager:
+        api_token = requests.get(
+            "https://my-api.com/token", auth=(self.username, self.password)
+        ).json()
+        return ExternalIOManager(api_token)
+```
 
 ### Handling partitioned assets
 
@@ -440,17 +469,12 @@ class MyNumpyLoader(PandasIOManager):
         return array
 
 
-@io_manager
-def numpy_io_manager():
-    return MyNumpyLoader()
-
-
 @op(ins={"np_array_input": In(input_manager_key="numpy_manager")})
 def analyze_as_numpy(np_array_input: np.ndarray):
     assert isinstance(np_array_input, np.ndarray)
 
 
-@job(resource_defs={"numpy_manager": numpy_io_manager, "io_manager": pandas_io_manager})
+@job(resource_defs={"numpy_manager": MyNumpyLoader(), "io_manager": PandasIOManager()})
 def my_job():
     df = produce_pandas_output()
     analyze_as_numpy(df)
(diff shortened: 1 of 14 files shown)


benpankow:
* master
  * **PR #13673** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13673" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13673?utm_source=stack-comment).

docs/content/concepts/io-management/io-managers.mdx
```diff
@@ -98,13 +98,18 @@ def downstream_asset(upstream_asset):
     return upstream_asset + [4]
 
 
+s3_resource = S3Resource()
```
yuhan, line 101: since it's just one line, should we move this to inline?

docs/content/concepts/io-management/io-managers.mdx
```diff
@@ -575,32 +587,29 @@ The extension attribute defines the suffix all the file paths generated by the I
...
-    }
-)
-def s3_parquet_io_manager(init_context: InitResourceContext) -> PandasParquetIOManager:
-    # `UPath` will read boto env vars.
-    # The credentials can also be taken from the config and passed to `UPath` directly.
-    base_path = UPath(init_context.resource_config.get("base_path"))
-    assert str(base_path).startswith("s3://"), base_path
-    return PandasParquetIOManager(base_path=base_path)
+
+class LocalPandasParquetIOManager(ConfigurableIOManagerFactory):
```
yuhan, line 595: do we plan to explain `ConfigurableIOManagerFactory`?

docs/content/concepts/io-management/io-managers.mdx
benpankow: yes 👍

docs/content/concepts/io-management/io-managers.mdx
```diff
@@ -100,11 +101,15 @@ def downstream_asset(upstream_asset):
...
+    resources={
+        "io_manager": ConfigurablePickledObjectS3IOManager(
+            s3_resource=S3Resource(), s3_bucket="my-bucket"
+        ),
+    },
 )
 ```
 
-This example also includes `"s3": s3_resource` because the <PyObject module="dagster_aws.s3" object="s3_pickle_io_manager" /> depends on an S3 resource.
+This example also constructs a <PyObject module="dagster_aws.s3" object="S3Resource" /> because the <PyObject module="dagster_aws.s3" object="ConfigurablePickledObjectS3IOManager" /> depends on the S3 resource.
```
yuhan, line 112: hmm was thinking maybe worth linking to resource-resource dependency doc section for more details (https://docs.dagster.io/master/concepts/resources#resources-which-depend-on-other-resources) but then was debating myself because we no longer say "io managers are resources" so linking might be actually more confusing 🤔

docs/content/concepts/io-management/io-managers.mdx
```diff
@@ -310,31 +321,59 @@ def a_job():
...
+
     def _get_path(self, context) -> str:
-        return "/".join(context.asset_key.path)
+        return "/".join(self.path_prefix + context.asset_key.path)
 
     def handle_output(self, context, obj):
         write_csv(self._get_path(context), obj)
 
     def load_input(self, context):
         return read_csv(self._get_path(context))
```
yuhan, line 342: can we type the context here?

docs/content/concepts/io-management/io-managers.mdx
```diff
@@ -310,31 +321,59 @@ def a_job():
...
+The provided `context` argument for `handle_output` is an <PyObject module="dagster" object="OutputContext" />. The provided `context` argument for `load_input` is an <PyObject module="dagster" object="InputContext" />. The linked API documentation lists all the fields that are available on these objects.
 
-@io_manager
-def my_io_manager():
-    return MyIOManager()
-```
+### Using an IO manager factory
 
-The <PyObject module="dagster" object="io_manager" /> decorator behaves nearly identically to the <PyObject module="dagster" object="resource" /> decorator. It yields an <PyObject module="dagster" object="IOManagerDefinition" />, which is a subclass of <PyObject module="dagster" object="ResourceDefinition" /> that will produce an <PyObject module="dagster" object="IOManager" />.
+If your IO manager is more complex, or needs to manage internal state, it may make sense to split out the IO manager definition from its configuration. In this case, you can use `ConfigurableIOManagerFactory`, which specifies config schema and implements a factory function that takes the config and returns an IO manager.
```
yuhan, line 349: > you can use `ConfigurableIOManagerFactory`

do we have apidoc for `ConfigurableIOManagerFactory`? maybe use PyObject instead of plain code syntax here so users can click to see details

docs/content/concepts/io-management/io-managers.mdx
```diff
@@ -310,31 +321,59 @@ def a_job():
...
+class ConfigurableExternalIOManager(ConfigurableIOManagerFactory):
+    username: str
+    password: str
+
+    def create_io_manager(self, context) -> ExternalIOManager:
+        api_token = requests.get(
+            "https://my-api.com/token", auth=(self.username, self.password)
+        ).json()
+        return ExternalIOManager(api_token)
+```
```
yuhan, line 376: i found this example a bit abstract and hard to grasp why i'd bother with the factor pattern. maybe worth switching to a more realistic io manager example, or keep it but also link to concrete examples like "Custom filesystem-based IO Manager" down below.

docs/content/concepts/io-management/io-managers.mdx
schrockn: Agree on this, which is why IMO we should stick with the "IOManagers are resources" language.

docs/content/concepts/io-management/io-managers.mdx
schrockn: Could be as simple as adding some sort of cache as private variable to the `ExternalIOManager` and note that one could not add that to a `ConfigurableIOManager` without erroring.

schrockn (changes requested):
Req'ing changes to switch to "IO managers are resources"

This would be late breaking codemod, but would the language make more sense if we called `ConfigurableIOManagerFactory` a `DelegatingIOManager`?

Also @benpankow to support input and output config schema, we will need to add class methods to `ConfigurableResourceFactory` and thread those values to `ConfigurableIOManagerFactoryResourceDefinition` construction sites. There as already feedback about this in Slack. See [thread](https://dagster.slack.com/archives/C01U954MEER/p1681264993096749).

docs/content/concepts/io-management/io-managers.mdx
```diff
@@ -18,12 +18,13 @@ IO Managers are user-provided objects that store asset and op outputs and load t
...
+| ----------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| <PyObject module="dagster" object="ConfigurableIOManager" decorator />        | A base class used to define configurable IO managers which are also resources.                                                                                                                              |
+| <PyObject module="dagster" object="ConfigurableIOManagerFactory" decorator /> | A base class used to specify configuration for more advanced I/O managers.                                                                                                                                  |
+| <PyObject module="dagster" object="IOManager" />                              | Base class for user-provided IO managers.                                                                                                                                                                   |
```
jamiedemaria, line 25: i know this is what it said before, but if i were new to IO managers i think this description "Base class for user-provided IOmanagers" meant any io manager i wrote should be off of this class

docs/content/concepts/io-management/io-managers.mdx
```diff
@@ -18,12 +18,13 @@ IO Managers are user-provided objects that store asset and op outputs and load t
...
+| <PyObject object="build_input_context"/>                                      | Function for directly constructing a <PyObject object="InputContext"/>, to be passed to the <PyObject object="IOManager" method="load_input"/> method. This is designed primarily for testing purposes.     |
```
erinkcochran87, line 26:
```suggestion
| <PyObject object="build_input_context"/>                                      | Function for directly constructing an <PyObject object="InputContext"/>, to be passed to the <PyObject object="IOManager" method="load_input"/> method. This is designed primarily for testing purposes.     |
```

docs/content/concepts/io-management/io-managers.mdx
```diff
@@ -18,12 +18,13 @@ IO Managers are user-provided objects that store asset and op outputs and load t
...
+| <PyObject object="build_output_context"/>                                     | Function for directly constructing a <PyObject object="OutputContext"/>, to be passed to the <PyObject object="IOManager" method="handle_output"/> method. This is designed primarily for testing purposes. |
```
erinkcochran87, line 27:
```suggestion
| <PyObject object="build_output_context"/>                                     | Function for directly constructing an <PyObject object="OutputContext"/>, to be passed to the <PyObject object="IOManager" method="handle_output"/> method. This is designed primarily for testing purposes. |
```

docs/content/concepts/io-management/io-managers.mdx
```diff
@@ -66,9 +67,9 @@ The default IO manager, <PyObject module="dagster" object="fs_io_manager" />, st
 
 Dagster provides out-of-the-box IO managers for popular storage systems: AWS S3 (<PyObject module="dagster_aws.s3" object="s3_pickle_io_manager" />), Azure Blob Storage (<PyObject module="dagster_azure.adls2" object="adls2_pickle_io_manager" />), GCS (<PyObject module="dagster_gcp.gcs" object="gcs_pickle_io_manager" />), and Snowflake (<PyObject module="dagster_snowflake" object="build_snowflake_io_manager" />) - or you can write your own: either from scratch or by extending the `UPathIOManager` if you want to store data in an `fsspec`-supported filesystem.
 
-### IO managers are resources
+### IO managers are provided by resources
 
-IO managers are [resources](/concepts/resources), which means users can supply different IO managers for the same assets or ops in different situations. For example, you might use an in-memory IO manager for unit-testing and an S3 IO manager in production.
+IO managers are provided through the [resources](/concepts/resources) system which means users can supply different IO managers for the same assets or ops in different situations. For example, you might use an in-memory IO manager for unit-testing and an S3 IO manager in production.
```
erinkcochran87, line 72:
```suggestion
IO managers are provided through the [resources](/concepts/resources) system which means you can supply different IO managers for the same assets or ops in different situations. For example, you might use an in-memory IO manager for unit-testing and an Amazon S3 IO manager in production.
```

Switching `user` to `you` to be consistent with the rest of this page

docs/content/concepts/io-management/io-managers.mdx
```diff
@@ -193,7 +203,7 @@ In some cases you may need to load the input to an asset with different logic th
 
 To set an IO manager for a particular input, use the `input_manager_key` argument on <PyObject module="dagster" object="AssetIn" />.
 
-In this example,`first_asset` and `second_asset` will be stored using the default IO manager, but will be loaded as inputs to `third_asset` using the logic defined in the `pandas_series_io_manager` (in this case loading as Pandas Series rather than python lists).
+In this example,`first_asset` and `second_asset` will be stored using the default IO manager, but will be loaded as inputs to `third_asset` using the logic defined in the `PandasSeriesIOManager` (in this case loading as Pandas Series rather than python lists).
```
erinkcochran87, line 206:
```suggestion
In this example,`first_asset` and `second_asset` will be stored using the default IO manager, but will be loaded as inputs to `third_asset` using the logic defined in the `PandasSeriesIOManager` (in this case loading as Pandas Series rather than Python lists).
```

docs/content/concepts/io-management/io-managers.mdx
```diff
@@ -310,31 +321,59 @@ def a_job():
 
 If you have specific requirements for where and how your outputs should be stored and retrieved, you can define your own IO manager. This boils down to implementing two functions: one that stores outputs and one that loads inputs.
 
-To define an IO manager, use the <PyObject module="dagster" object="io_manager" displayText="@io_manager" /> decorator.
+To define an IO manager, you will want to extend the <PyObject module="dagster" object="IOManager" /> class. Often times, you will want to extend the <PyObject module="dagster" object="ConfigurableIOManager"/> class (which subclasses `IOManager`) to attach a config schema to your IO manager.
```
erinkcochran87, line 324:
```suggestion
To define an IO manager, extend the <PyObject module="dagster" object="IOManager" /> class. Oftentimes, you will want to extend the <PyObject module="dagster" object="ConfigurableIOManager"/> class (which subclasses `IOManager`) to attach a config schema to your IO manager.
```

docs/content/concepts/io-management/io-managers.mdx
erinkcochran87: +1 ^

We should use this whenever possible. Better experience for users, but it'll also make it easy for us to keep things updated/correct.

erinkcochran87 (commented):
- Can you update `IO manager` to `I/O manager`? Some of the new copy uses both, and this is a good opportunity to standardize it in here.

docs/content/concepts/io-management/io-managers.mdx
benpankow: Added a reference to the resource-resource dep docs

docs/content/concepts/io-management/io-managers.mdx
benpankow: Updated example to use a cache.

benpankow:
> Also @benpankow to support input and output config schema, we will need to add class methods to ConfigurableResourceFactory and thread those values to ConfigurableIOManagerFactoryResourceDefinition construction sites. There as already feedback about this in Slack. See [thread](https://dagster.slack.com/archives/C01U954MEER/p1681264993096749).

https://github.com/dagster-io/dagster/pull/13740

benpankow:
> This would be late breaking codemod, but would the language make more sense if we called ConfigurableIOManagerFactory a DelegatingIOManager?

We could leave `ConfigurableIOManagerFactory` exported if we're worried about this change. It's a bit late in the game, but I'm not opposed to doing so.

benpankow:
Did a general pass, removed the output config section for now (but leaving the old legacy IO manager guide as a separate page, which includes it).

PR #13679, opened 2023-04-17
add `MultiPartitionKey` to `OpExecutionContext.partition_key` return type hint

askvinni:
## Summary & Motivation
Noticed the type hint for `OpExecutionContext.partition_key` is set to only `str`, but it can also return a `MultiPartitionKey`.

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/context/compute.py
+++ b/python_modules/dagster/dagster/_core/execution/context/compute.py
@@ -1,5 +1,17 @@
 from abc import ABC, abstractmethod
-from typing import AbstractSet, Any, Dict, Iterator, List, Mapping, Optional, Sequence, Set, Union, cast
+from typing import (
+    AbstractSet,
+    Any,
+    Dict,
+    Iterator,
+    List,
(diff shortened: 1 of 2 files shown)


tacastillo:
I'm not actually sure who would own this module, so we're going to play some hot potato for it to find the right reviewer. I'm gonna throw it over to @clairelin135 for once she gets back from PTO tomorrow.

clairelin135:
Hi @askvinni, thanks for filing. Makes a lot of sense to change the return type, but unfortunately I think this would constitute a breaking change as mypy will complain for any user who is relying on the return type to be a string.

I think in order to make this change we'd probably have to do it in the next major release. And ideally for full coverage we'd change the return types of all the partition key methods (context.partition_key_range, context.asset_partition_keys_for_output, etc).

askvinni:
@clairelin135 gotcha, makes sense. Probably best to close the PR then?

clairelin135:
Yep, I think so. Appreciate you contributing nonetheless, hopefully we can get this in in 1.4.0

PR #13681, opened 2023-04-17
[pythonic config] Add ability to create int EnvVars

benpankow:
## Summary

Introduces a way to use `EnvVar` values with int fields:

```python
class MyIntConfig(Config):
  an_int: int


MyIntConfig(an_int=EnvVar.int("AN_INT_ENV"))
```

There is not an immediate analogue for float (since we have no `FloatSource`) or bool (since `bool` is final and cannot be extended - needs some cleverness).

## Test Plan

New unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_config/field_utils.py
+++ b/python_modules/dagster/dagster/_config/field_utils.py
@@ -469,6 +469,10 @@ def _config_dictionary_from_values_inner(obj: Any):
         return [_config_dictionary_from_values_inner(v) for v in obj]
     elif isinstance(obj, EnvVar):
         return {"env": str(obj)}
+    elif isinstance(obj, IntEnvVar):
+        return {"env": obj.name}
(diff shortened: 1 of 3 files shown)


benpankow:
* master
  * **PR #13681** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13681" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13681?utm_source=stack-comment).

schrockn (changes requested):
float in sep diff?

benpankow:
ah yup, left that in by mistake - updated

schrockn (approved):
👍🏻

schrockn:
FWIW adding float source should be relatively straightforward. You can mimic the other implementations in python_modules/dagster/dagster/_config/source.py

PR #13684, opened 2023-04-17
[dagit] Show the partition dialog for partitioned observable source assets

bengotow:


bengotow:
I think we've decided not to pursue this for the time being

PR #13693, opened 2023-04-17
Guard against cursor deserialization issue

OwenKephart:
## Summary & Motivation

Previously, if you changed from (e.g.) an HourlyPartitionsDef to a DailyPartitionsDef, you'd hit an error that you could not recover from, as we would try to deserialize the old partition keys in the context of the new partitions definition. Now, we catch those errors and instead just use an empty partitions definition when this occurs.

Unit testing would require a fair bit of reworking of the daemon part of the test suite as it currently is assumed that the repo will be the same between different iterations, so the same WorkspaceRequestContext thing is used, even if you supply a different set of assets definitions in the `cursor_from` argument.

Manually tested instead.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -280,7 +280,9 @@ class AssetReconciliationCursor(NamedTuple):
                     key
                 ] = partitions_def.deserialize_subset(serialized_subset)
             except:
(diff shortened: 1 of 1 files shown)


sryza:
Can you add a test to `definition_change_scenarios.py`?

OwenKephart:
@sryza the definition changed scenarios are not properly handled by the test framework. it just so happens that the existing tests pass, because the sensors are evaluated against the outer AssetReconciliationScenario workspace, and all of the existing tests assume that no runs will be kicked off (which is correct behavior regardless of what the previous workspace looked like)

sryza (approved):

PR #13699, opened 2023-04-17
[ui] Cache code location reload errors after mutation

hellendag:
## Summary & Motivation

When reloading a code location, cache the mutation response. This should update the UI to show the correct error.

The goal here is to resolve an issue where a code location has multiple errors surfaced when loading it. After fixing one error and reloading, the fixed error doesn't go away. This is because the error is tracked via the top-level workspace query, but not updated by the mutation. By writing to the cache, the error dialog updates with the new (as yet unfixed) error right away.

https://user-images.githubusercontent.com/2823852/232617176-c9515797-7871-49cf-abbd-791e6555525b.mov

## How I Tested These Changes

Raise exceptions from two `@repository` objects in `toys/repo.py`, load UI. Navigate to code locations, open error dialog. Comment out one exception, click "Reload". Verify that the next exception shows up in the dialog. Comment it out, click Reload. Verify that the code location loads successfully.

### DIFF
--- a/js_modules/dagit/packages/core/src/nav/types/useRepositoryLocationReload.types.ts
+++ b/js_modules/dagit/packages/core/src/nav/types/useRepositoryLocationReload.types.ts
@@ -124,5 +124,18 @@ export type ReloadRepositoryLocationMutation = {
         __typename: 'WorkspaceLocationEntry';
         id: string;
         loadStatus: Types.RepositoryLocationLoadStatus;
+        locationOrLoadError:
+          | {
+              __typename: 'PythonError';
+              message: string;
+              stack: Array<string>;
+              errorChain: Array<{
+                __typename: 'ErrorChainLink';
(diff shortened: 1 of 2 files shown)


hellendag:
* master
  * **PR #13699** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13699" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13699?utm_source=stack-comment).

bengotow (approved):
This looks great and the test plan seems thorough 👍  I wish there was some way to comment the GraphQL to note the reason for the extra fields so we can be sure they don't get trimmed in the future, but I don't think we really can.

hellendag:
Yeah, that's a good point. I think we just always need to be hesitant to remove `id` fields...

PR #13705, opened 2023-04-17
fix logic for new_version_exists

OwenKephart:
## Summary & Motivation

Previously, if there was no asset observation for a given asset before the latest_storage_id, then we would always return False for new_version_exists (because previous_version_record would be None). However, this meant that if an asset was observed, then the next tick of the sensor would ignore that new observation, as the only observation of the asset would exist *after* the latest storage id.


johannkm (approved):
A test for this would be excellent

PR #13722, opened 2023-04-18
Change auto materialize tag to auto_materialize:true

johannkm:


prha:
@sryza have you considered having the values instead be `schedule: my_schedule`, `sensor: my_sensor`, `ui`, `auto_materialize`? 

Trying to figure out what views we would want filtering by.

alangenfeld:
It seems like we started off trying to solve
1. what to tag `AutoMaterializePolicy` instigated runs 

and trying to be forward looking are toying with 

2. a generalize tagging scheme for what instigated a run 

Trying to do 2 for just this case and not actually apply it to all other instigators seems dubious to me. 

`policy` as an instigator type conceptually sibling to sensor/schedule meaning "a fully dagster managed process respecting declarative attributes provided by user" makes sense to me. I think if we want to solve just 1, we could follow the existing sensor / schedule pattern and do `policy_name` or similar. Whatever we decide to do to solve 2 later would then just apply the same here as it does for sensor / schedule. If i am alone on that theres other ways to solve for just 1 including "just leave it". 

If we want to solve for 2 and not actually apply it consistently, you should make a case for why thats preferable. 

If we want to solve for 2 for real, phra's questions above are good. I think we want a scheme where the value could be a column in the runs table, which i think points towards a `<instigation type>:<instigator name>`. I personally like `policy:auto_materielize` / `policy:run_retry` more than `auto_materialize` / `run_retry` given my mental model described above.

alangenfeld (changes requested):
to your queue to clarify what exactly we are trying to solve in this PR

sryza:
> we could follow the existing sensor / schedule pattern and do `policy_name` or similar.

Policies don't have names - there's multiple schedules / sensors, but a single auto-materialize daemon. Each asset has its own policy.

> If i am alone on that theres other ways to solve for just 1 including "just leave it".

I think "just leave it" is effectively solving for 2, because the tag key we have right now, `created_by` is not specific to auto-materializing.

If we wanted to purely solve for 1, then I think we might want a tag with key "auto_materialize" and value "true"?

johannkm:
I'm not opposed to `auto_materialize: true`, I think it's just a little less clear than `created_by: auto_materialize`.

Overall goal is that looking at the runs page quickly clarifies where a run came from. Currently I think the possibilities are
- manual
- sensor
- schedule
- backfill
- run retries
- auto materialize

prha:
I think all of this makes sense.  I'm glad @alangenfeld listed out the different goals.

My thoughts are mainly that for (2), tags are probably the wrong storage mechanism for indicating "created by".  But more than that, we'd have to inventory all of our run creation flows to make sure we're not lying.  We'd have to put in some special logic for this specific tag so they don't get copied over from the UI.  For example, can I manually retry an auto-materialization run?  What is listed as the created by value there?

In effect, there's already an implicit ordering to figure out whether the tag presence indicates the true source.  I think we can (and should) get to a place where a `created_by` tag is the definitive truth, but I don't think this PR is comprehensive enough for that to be the case.

Short of that, I'd just as soon treat the tag as an implementation detail, use `auto-materialize: true` for the tag key/value and then change the front-end to make it look however we want.  I'm pretty sure that @salazarm / @braunjj are already doing that (translating tags into a created column on the runs table).

sryza (approved):

PR #13725, opened 2023-04-18
further improvemnts for test_threaded_ephemeral_instance

alangenfeld:
My best guess for the remaining flakes https://buildkite.com/dagster/dagster/builds/56393#01879468-5c83-4838-819e-24ff92adf248
is that the connection fairy is firing via GC for main thread connections left over from other tests when another thread is running. 

* cleans up calls that were causing GC based connection finalization that i could find by breaking in sqlalchemy/base/pool.py in the debugger (even though they are sqlite)
* pause GC during this test to prevent it from running in a worker thread

## How I Tested These Changes

could never reproduce locally, ran full suite repeatedly as well as the specific test with `--count 1000`

### DIFF
--- a/python_modules/dagster/dagster_tests/storage_tests/test_local_instance.py
+++ b/python_modules/dagster/dagster_tests/storage_tests/test_local_instance.py
@@ -251,44 +251,51 @@ def test_run_step_stats_with_retries():
 
 
 def test_threaded_ephemeral_instance(caplog):
-    n = 5
+    # prevent GC from firing during this to prevent false positives on thread warnings
+    gc.disable()
+    try:
+        n = 5
 
-    with DagsterInstance.ephemeral() as shared_instance:
+        with DagsterInstance.ephemeral() as shared_instance:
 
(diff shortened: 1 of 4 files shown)


alangenfeld:
* master
  * **PR #13725** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13725" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13725?utm_source=stack-comment).

python_modules/dagster/dagster_tests/storage_tests/test_local_instance.py
```diff
@@ -251,44 +251,51 @@ def should_not_execute(_, x):
 
 
 def test_threaded_ephemeral_instance(caplog):
-    n = 5
+    # prevent GC from firing during this to prevent false positives on thread warnings
+    gc.disable()
```
johannkm, line 255: woah

johannkm (approved):

PR #13737, opened 2023-04-18
Changing pandas Timestamp check in snowflake_pandas_type_handler

ezpzbz:
## Summary & Motivation
Currently, in `snowflake_pandas_type_handler` the `_add_missing_timezone` method uses `is_datetime_or_timedelta_dtype` to check for `pd.Timestamp` to decide adding the missing timezone. However, this fails if we have `timedelta` datatype columns in the dataframe.
This PR should fix this issue by changing the check to `is_datetime64_any_dtype`.

However, the issue with handling `timedelta` still exists for which I kept this PR as draft in case further discussion would help resolving that issue too.
I have following suggestions for the `timedelta` columns:

* convert them to `total_seconds` so they can be stored in efficient format.
* changing the `_add_missing_timezone` method to something like `_handle_datetime_data` and add the `timedelta` handling there.

Once we decide on a proper solution, I can have it added to this PR and make it ready for review.

@jamiedemaria Hi Jamie, I would be glad if you let me know your thoughts about this.

## How I Tested These Changes
Ran tests locally.
It still requires adding proper tests for cases discussed above.


jamiedemaria:
Hey @pzarabadip  thanks for the contribution! I'll take a closer look later this week! 

What specifically is the issue with storing timedelta data? is it the same/similar as normal timestamp data (ie snowflake garbles it and stores it as nonsensical values) or something else?

jamiedemaria:
ok great. the snowflake behavior will determine if there's anything we need to do on our end to store it properly. If snowflake does store the data as expected, then we can probably just go ahead and change this method so that we don't try to modify timedelta data

pzarabadip:
Hey @jamiedemaria I ran some tests for this. I created the following dataframe:
```
from datetime import datetime
import pandas as pd
import numpy as np


days = np.random.randint(1,28,12)
months = np.random.randint(1,12,12)
years = np.random.randint(2010,2020,12)

dataset = []
for y,m,d in zip(years,months,days):
    
    data = {}
    data['year'] = y
    data['month'] = m
    data['day'] = d
    data['time_string'] = f'{y}-{m}-{d}'
    data['timestamp'] = datetime(y, m, d)
    data['timedalta'] = datetime.now() - datetime(y, m, d)
    dataset.append(data)

df_pd = pd.DataFrame(dataset)
```
Before applying the `_add_missing_timezone`, we get the expected compilation error for the `timestamp` column:
```
ProgrammingError: (snowflake.connector.errors.ProgrammingError) 002023 (22000): SQL compilation error:
Expression type does not match column data type, expecting TIMESTAMP_NTZ(9) but got NUMBER(38,0) for column timestamp
[SQL: INSERT INTO "test_timedelta" ("year", "month", "day", "time_string", "timestamp", "timedalta") SELECT t0."year", t0."month", t0."day", t0."time_string", t0."timestamp", t0."timedalta" 
FROM "_ibis_pandas_memtable_btu4dx499ejyj35isnhs1idv7" AS t0]
(Background on this error at: https://sqlalche.me/e/14/f405)
```
After applying the `_add_missing_timezone`, as expected the above error goes away and we get a similar error for `timedelta` column:
```
ProgrammingError: (snowflake.connector.errors.ProgrammingError) 002023 (22000): SQL compilation error:
Expression type does not match column data type, expecting TIMESTAMP_NTZ(9) but got NUMBER(38,0) for column timedalta
[SQL: INSERT INTO "test_timedelta" ("year", "month", "day", "time_string", "timestamp", "timedalta") SELECT t0."year", t0."month", t0."day", t0."time_string", t0."timestamp", t0."timedalta" 
FROM "_ibis_pandas_memtable_23r0o5ih2bu2z5qkxcumge9tq" AS t0]
(Background on this error at: https://sqlalche.me/e/14/f405)
```
To get rid of this error, the best solution in my opinion is converting the `timedelta` type data to `total_seconds` which I tested to be okay.
The question would be whether we want to change them back to duration when the data is retrieved from the database.

**NOTE** BTW, my suggested change in this PR, still can result in failure if the timestamp column already has the timezone info. My suggestion would be doing something like below:
```
def _add_missing_timezone(s: pd.Series) -> pd.Series:

    if pd_core_dtypes_common.is_datetime64_any_dtype(s):  # type: ignore  # (bad stubs)
        if not pd_core_dtypes_common.is_datetime64tz_dtype(s):  # type: ignore  # (bad stubs)
            return s.dt.tz_localize("UTC")
    return s
```

jamiedemaria:
Just wanted to give you an update on this. The issue solved in this PR is unfortunately part of a larger set of issues around how snowflake handles timestamp/timedelta data. I'm still working on understanding all of the issues at play, so i'd like to hold off on merging this until we have a better idea about what's going on.

python_modules/libraries/dagster-snowflake-pandas/dagster_snowflake_pandas/snowflake_pandas_type_handler.py
```diff
@@ -68,7 +68,7 @@ def _add_missing_timezone(
     s: pd.Series, column_types: Optional[Mapping[str, str]], table_name: str
 ) -> pd.Series:
     column_name = str(s.name)
-    if pd_core_dtypes_common.is_datetime_or_timedelta_dtype(s):  # type: ignore  # (bad stubs)
+    if pd_core_dtypes_common.is_datetime64_any_dtype(s):  # type: ignore  # (bad stubs)
```
jamiedemaria, line 71: can remove the inline comment here! the function is no longer unrecognized

jamiedemaria:
Hey @pzarabadip can you rebase your PR on master? the unit tests for your changes are green, but i think some other test failures are happening because you are based on a older version of the code base.

Additionally, do you know if the function you are using was added in a recent version of pandas? We may need to include a version pin if so

pzarabadip:
Hi @jamiedemaria, thanks for the review and comments. I'll do the rebase and corrections tomorrow.

pzarabadip:
Hey @jamiedemaria, I just merged the latest commit of `master` into this branch and also updated the function.
As of `pandas`, `pandas` in my dev env is `1.5.3`. I need to check if these still exists in `pandas` >2.0.

jamiedemaria:
hey - so sorry I dropped off on this. It fell out of my review queue and into the void. good news is that I believe we have removed the need for adding timezone at all with https://github.com/dagster-io/dagster/pull/18410. Closing this, but feel free to open an issue if you run into any bugs once #18410 is released

PR #11346, opened 2022-12-22
[prototype] Pass typed config args to IO manager save/load fns

benpankow:

PR #13740, opened 2023-04-18
[pythonic resources] Add support for input, output config in ConfigurableIOManagers

benpankow:
## Summary

Adds support for specifying input and output config types for `ConfigurableIOManager`s and `ConfigurableIOManagerFactory`s, loosely based on #11346.

These schemas can be either old-style config dicts or `Config` classes. The raw config data passed to input/output functions is a config dict, but we could instead support passing the config directly as in #11346.

```python
class OutputConfigSchema(Config):
    path: str

class InputConfigSchema(Config):
    format: str

class MyIOManager(ConfigurableIOManager):
    base_dir: str

    @classmethod
    def input_config_schema(cls) -> Type[Config]:
        return InputConfigSchema

    @classmethod
    def output_config_schema(cls) -> Type[Config]:
        return OutputConfigSchema

        ...

        ...
```

## Test Plan

New tests.

### DIFF
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -33,6 +33,7 @@ from dagster._config.source import BoolSource, IntSource, StringSource
 from dagster._config.validate import process_config, validate_config
 from dagster._core.decorator_utils import get_function_params
 from dagster._core.definitions.definition_config_schema import (
+    CoercableToConfigSchema,
     ConfiguredDefinitionConfigSchema,
     DefinitionConfigSchema,
 )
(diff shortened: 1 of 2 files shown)


benpankow:
* master
  * **PR #13740** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13740" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13740?utm_source=stack-comment).

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -515,9 +516,26 @@ def __init__(
...
+            cast(Type[Config], input_config_schema).to_config_schema()
+            if safe_is_subclass(input_config_schema, Config)
+            else cast(CoercableToConfigSchema, input_config_schema)
+        )
+        output_config_schema_resolved: CoercableToConfigSchema = (
+            cast(Type[Config], output_config_schema).to_config_schema()
+            if safe_is_subclass(output_config_schema, Config)
+            else cast(CoercableToConfigSchema, output_config_schema)
+        )
+        print(input_config_schema_resolved)
```
schrockn, line 532: print

schrockn (changes requested):

schrockn (approved):

benpankow:
> is this functionality still planned? I'm trying to figure out if now is the right time to migrate our IOManagers to the new style.

Yes, this should be shipping with next week's release!

> Will there be some sort of support for input/ouput config using `RunConfig`, or will we have to use plain dicts?

For now you can construct the config objects in the input/output function bodies. It's a bit tricky for us to adjust the signature of these methods without incurring backwards compat issues, but we may revisit this down the line

```python
class OutputConfigSchema(Config):
    path: str

class InputConfigSchema(Config):
    format: str

class MyIOManager(ConfigurableIOManager):
    base_dir: str

    @classmethod
    def input_config_schema(cls) -> Type[Config]:
        return InputConfigSchema

    @classmethod
    def output_config_schema(cls) -> Type[Config]:
        return OutputConfigSchema

        output_config = OutputConfigSchema(**context.config)
        ...

        input_config = InputConfigSchema(**context.config)
        ...
```

erinov1:
Thanks! My question was more about how to supply the input/output config to a job. I have a bunch of assets in a job, some of which specify optional input/output parameters to an IO manager. Currently I pass a plain dictionary to the `config` argument in `define_asset_job`. My understanding is that `RunConfig` is the new preferred way to pass configuration to the `config` argument, but this doesn't seem possible if I need to specify per-asset config for the IO manager? For example, from the tests for this PR:

```python
check_input_managers.execute_in_process(
    run_config={
        "ops": {
            "first_op": {"outputs": {"first_op": {"postfix_output": "post"}}},
            "second_op": {
                "inputs": {"first_op": {"prefix_input": "bar"}},
                "outputs": {"second_op": {"postfix_output": "post"}},
            },
        }
    }
)
```

PR #13743, opened 2023-04-18
[auto-materialization] Fix detection of in-progress runs

OwenKephart:
## Summary & Motivation

We try to be a bit clever when checking if a newly-detected AssetMaterialization is going to be propagated by a currently in-progress run. The naive solution would be to always just query if the run that created that materialization also plans to materialize the child, but this would result in lots and lots of queries, which would be very inefficient.

Instead, we filter out cases where we know for a fact that the child cannot exist in the same run (i.e. they have different partitions definitions or different partition keys).

The old solution skipped over the fact that TimeWindowPartitionsDefinitions <> TimeWindowPartitionsDefinitions will have a TimeWindowPartitionMapping, not an IdentityPartitionMapping.


johannkm (changes requested):
could we get a new test scenario

johannkm (approved):

PR #13758, opened 2023-04-19
update runs filtering to use new component

salazarm:
## Summary & Motivation

We're updating the Runs page to the new filtering component.

There is still a quirk due to the `Loading` component recreating its children from scratch in a way that they lose their internal state. This causes the dropdown to close whenever the page refreshes. I need to dig into this further but it doesn't look like a quick fix for now.

There are also further visual changes to make but I will do those in a follow-up.

## How I Tested These Changes

Jest  + manual testing

https://www.loom.com/share/7bf0cfd841284b02a27e73511b95b92d

### DIFF
--- a/js_modules/dagit/packages/core/src/ui/Filters/useFilter.tsx
+++ b/js_modules/dagit/packages/core/src/ui/Filters/useFilter.tsx
@@ -4,14 +4,7 @@ import styled from 'styled-components/macro';
 
 import {TruncatedTextWithFullTextOnHover} from '../../nav/getLeftNavItemsForOption';
 
-export type FilterListenerCallback<TState> = (value: {
-  state: TState;
-  previousActive: boolean;
-  active: boolean;
-  name: string;
-}) => void;
-
-export type FilterObject<TState> = {
+export type FilterObject = {
   isActive: boolean;
   activeJSX: JSX.Element;
   icon: IconName;
@@ -22,8 +15,8 @@ export type FilterObject<TState> = {
     close: () => void;
     createPortal: (element: JSX.Element) => () => void;
   }) => void;
-  state: TState;
-  setState: (state: TState) => void;
+  onUnselected?: () => void;
+  isLoadingFilters?: boolean;
 };
 
 export const FilterTag = ({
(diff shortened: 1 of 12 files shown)


js_modules/dagit/packages/ui/src/components/SubwayDot.tsx
```diff
@@ -0,0 +1,95 @@
+// eslint-disable-next-line no-restricted-imports
```
salazarm, line 1: This is a copy paste job of the SubwayDot component from cloud

braunjj:
this looks great. one nit, can we make sure the filter tags are the same height as the rest of our tags in the app?  These looks like they are scaling up to fit the height of the container. 

They should have 24px height and be centered vertically

image

js_modules/dagit/packages/core/src/ui/Filters/useFilter.tsx
```diff
@@ -35,17 +28,19 @@ export const FilterTag = ({
...
-    rightIcon={
-      <div onClick={onRemove} style={{cursor: 'pointer'}} tabIndex={0}>
-        <Icon name="close" color={Colors.Link} />
-      </div>
-    }
-    label={label}
-    fillColor={Colors.Blue50}
-    textColor={Colors.Link}
-  />
+  <div style={{display: 'inline-block', height: '24px'}}>
```
hellendag, line 31: If this is to avoid the vertical stretching on the tag, it might be sufficient just to have an unstyled `div` wrapper, since then the `BaseTag` won't be affected by the flex properties of the container.

hellendag (approved):

js_modules/dagit/packages/core/src/ui/Filters/useSuggestionFilter.tsx
```diff
@@ -0,0 +1,180 @@
...
+      renderActiveStateLabel,
+      nextSuggestions,
+      nextSuggestionsLoading,
+      initialSuggestions,
+      getKey,
+      isMatch,
+      setState,
+      onSuggestionClicked,
+      isOR,
+    ],
```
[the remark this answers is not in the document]
salazarm: yeah I think so because this a hook. It can be used with a lot of other hooks that could be causing re-rendering.

js_modules/dagit/packages/core/src/ui/Filters/useFilter.tsx
salazarm: ah let me try that

PR #13764, opened 2023-04-19
Added get_gcs_keys helper function

ajohnson5:
## Summary & Motivation
Added `get_gcs_keys` helper function to the dagster-gcp library for use in detecting when new files have been added to gcs buckets.

In the [Dagster docs](https://docs.dagster.io/concepts/partitions-schedules-sensors/sensors#s3-sensors) there is an example of using sensors for detecting when new files are uploaded into an s3 bucket which makes reference to a get_s3_keys helper function. In my own personal project I needed to use a similar design but for Google Cloud Storage so I have refactored the `get_s3_keys` function found here: [get_s3_keys github](https://github.com/dagster-io/dagster/blob/master/python_modules/libraries/dagster-aws/dagster_aws/s3/sensor.py). Hopefully this can help other people with similar sensors in the future. 

## How I Tested These Changes
Tested locally to match behaviour of `get_s3_keys` function.

### DIFF
--- a/python_modules/libraries/dagster-gcp/dagster_gcp/gcs/sensor.py
+++ b/python_modules/libraries/dagster-gcp/dagster_gcp/gcs/sensor.py
@@ -1,0 +1,45 @@
+import dagster._check as check
+from google.cloud import storage
+
+MAX_KEYS = 1000
+
+
+def get_gcs_keys(bucket, prefix="", since_key=None, gcs_session=None):
+    check.str_param(bucket, "bucket")
+    check.str_param(prefix, "prefix")
+    check.opt_str_param(since_key, "since_key")
+
+    if not gcs_session:
+        gcs_session = storage.Client()
+
+    cursor = ""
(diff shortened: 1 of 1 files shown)


tacastillo:
Hi! I'm going to try to find the right person to help work with you on this PR.

In the meantime, have you run it through `black` and `ruff`, thought about tests, etc.?

mlarose:
hi @ajohnson5 - Better late than never, but I wrote a [test under review for the get_s3_keys](https://github.com/dagster-io/dagster/pull/19840). 

I used moto to mock the boto3 SDK that makes the call, the goal being to test the logic of the sensor, not the client library or S3 itself. I am not sure if there is a off the shelf mocking solution like moto for the Google Cloud Storage Python SDK but one could replicate with unittest.mock with a bit of effort.

mlarose:
replaced by https://github.com/dagster-io/dagster/pull/20191

PR #13766, opened 2023-04-19
[ui] Visual asset graph tweaks, fix for erroneous “Failed, Overdue” text

bengotow:
## Summary & Motivation

- On mini asset nodes, the spinner should be vertically centered

- When an asset is late, show "Failed, Overdue" only if there is also a failed run ID

- Make the "materializing" spinner the same size as the green / red dots and undo it's padding to align it perfectly with the dots.

- Use a relative font size increase (1.2em) to make the monospace text in the asset status the right size. That way, the surrounding font size is not overridden. 

## How I Tested These Changes

- I created a new storybook case

- Visual tests below

image

image

image

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
@@ -164,7 +164,7 @@ export function buildAssetNodeStatusContent({
         border: Colors.Gray300,
         content: (
           <>
-            <AssetLatestRunSpinner liveData={liveData} />
+            <AssetLatestRunSpinner liveData={liveData} purpose="caption-text" />
             <span style={{flex: 1}} color={Colors.Gray800}>
               Observing...
             </span>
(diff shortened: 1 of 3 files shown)


js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
```diff
@@ -407,7 +416,14 @@ export const AssetNodeMinimal: React.FC<{
             $background={background}
             $border={border}
           >
-            <div style={{position: 'absolute', bottom: 6, left: 6}}>
+            <div
+              style={{
+                position: 'absolute',
+                top: '50%',
+                transform: 'translateY(-16px)',
+                left: 8,
```
salazarm, line 424:
```suggestion
                transform: 'translate(8px, -16px)',
```

js_modules/dagit/packages/core/src/asset-graph/AssetRunLinking.tsx
```diff
@@ -37,6 +37,10 @@ export const AssetRunLink: React.FC<{
     target="_blank"
     rel="noreferrer"
   >
-    {children || <CaptionMono>{titleForRun({id: runId})}</CaptionMono>}
+    {children || (
+      <span style={{fontSize: '1.2em', fontFamily: FontFamily.monospace}}>
```
salazarm, line 41: Should we be doing this more generally across the app?

salazarm (approved):
,

js_modules/dagit/packages/core/src/asset-graph/AssetRunLinking.tsx
bengotow: I think we should definitely consider it!

PR #13769, opened 2023-04-19
Add storage methods for managing op concurrency

prha:
## Summary & Motivation
Storage methods for allocating/claiming/freeing concurrency slots.

At the instance-level, users must first allocate a set number of free slots per concurrency key.  Each slot is represented by a single row in the database.  During execution, the executor should try to claim a slot per concurrency key that applies to the step to be executed.

We will need to put some guards in place so that storage cannot be allocated in an unbounded fashion.

step lifecycle

concurrency table state

## How I Tested These Changes
BK

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/event_log/base.py
+++ b/python_modules/dagster/dagster/_core/storage/event_log/base.py
@@ -412,6 +412,11 @@ class EventLogStorage(ABC, MayHaveInstanceWeakref[T_DagsterInstance]):
         """Indicates that the EventLogStoarge is sharded."""
         return False
 
+    @property
+    def supports_global_concurrency_limits(self) -> bool:
+        """Indicates that the EventLogStorage supports global concurrency limits."""
+        return False
+
     @abstractmethod
     def allocate_concurrency_slots(self, concurrency_key: str, num_int: int) -> None:
         """Allocate concurrency slots for the given concurrency key."""
--- a/python_modules/dagster/dagster/_core/storage/event_log/base.py
+++ b/python_modules/dagster/dagster/_core/storage/event_log/base.py
@@ -433,7 +433,9 @@ class EventLogStorage(ABC, MayHaveInstanceWeakref[T_DagsterInstance]):
         raise NotImplementedError()
 
     @abstractmethod
-    def claim_concurrency_slots(self, concurrency_keys: Set[str], run_id: str, step_key: str):
+    def claim_concurrency_slot(
+        self, concurrency_key: str, run_id: str, step_key: str, priority: Optional[int] = None
+    ):
         """Claim concurrency slots for step."""
         raise NotImplementedError()
 
--- a/python_modules/dagster/dagster/_core/storage/event_log/base.py
+++ b/python_modules/dagster/dagster/_core/storage/event_log/base.py
@@ -28,6 +28,7 @@ from dagster._core.storage.dagster_run import DagsterRunStatsSnapshot
 from dagster._core.storage.sql import AlembicVersion
 from dagster._seven import json
 from dagster._utils import PrintFn
+from dagster._utils.concurrency import ConcurrencyClaimStatus, ConcurrencyKeyInfo
 
 if TYPE_CHECKING:
     from dagster._core.storage.partition_status_cache import AssetStatusCacheValue
@@ -418,41 +419,40 @@ class EventLogStorage(ABC, MayHaveInstanceWeakref[T_DagsterInstance]):
         return False
 
     @abstractmethod
-    def allocate_concurrency_slots(self, concurrency_key: str, num_int: int) -> None:
+    def set_concurrency_slots(self, concurrency_key: str, num: int) -> None:
         """Allocate concurrency slots for the given concurrency key."""
         raise NotImplementedError()
 
     @abstractmethod
-    def get_concurrency_limited_keys(self) -> Set[str]:
+    def get_concurrency_keys(self) -> Set[str]:
         """Get the set of concurrency limited keys."""
         raise NotImplementedError()
 
     @abstractmethod
-    def get_concurrency_info(self, concurrency_key: str, include_deleted: bool = False):
+    def get_concurrency_info(self, concurrency_key: str) -> ConcurrencyKeyInfo:
         """Get concurrency info for key."""
         raise NotImplementedError()
 
     @abstractmethod
     def claim_concurrency_slot(
         self, concurrency_key: str, run_id: str, step_key: str, priority: Optional[int] = None
-    ):
+    ) -> ConcurrencyClaimStatus:
         """Claim concurrency slots for step."""
         raise NotImplementedError()
 
     @abstractmethod
-    def free_concurrency_slots(self, run_id: str, step_key: Optional[str] = None):
+    def check_concurrency_claim(
+        self, concurrency_key: str, run_id: str, step_key: str
+    ) -> ConcurrencyClaimStatus:
+        """Claim concurrency slots for step."""
+        raise NotImplementedError()
+
+    @abstractmethod
+    def free_concurrency_slots_for_run(self, run_id: str) -> None:
+        """Frees concurrency slots for a given run."""
+        raise NotImplementedError()
+
+    @abstractmethod
+    def free_concurrency_slot_for_step(self, run_id: str, step_key: str) -> None:
         """Frees concurrency slots for a given run/step."""
-        raise NotImplementedError()
-
-    @abstractmethod
-    def can_claim_from_pending(self, concurrency_key: str, run_id: str, step_key: str):
-        """Indicates that a given step can now claim a concurrency, given its priority relative to
-        other pending steps
-        .
-        """
-        raise NotImplementedError()
-
-    @abstractmethod
-    def remove_pending_step(self, run_id: str, step_key: Optional[str] = None):
-        """Removes the pending step for a given run, step."""
         raise NotImplementedError()
--- a/python_modules/dagster/dagster/_core/storage/event_log/base.py
+++ b/python_modules/dagster/dagster/_core/storage/event_log/base.py
@@ -448,6 +448,11 @@ class EventLogStorage(ABC, MayHaveInstanceWeakref[T_DagsterInstance]):
         raise NotImplementedError()
 
     @abstractmethod
+    def get_concurrency_run_ids(self) -> Set[str]:
+        """Get a list of run_ids that are occupying or waiting for a concurrency key slot."""
+        raise NotImplementedError()
+
+    @abstractmethod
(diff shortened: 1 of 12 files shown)


prha:
* master
  * **PR #13769** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13769" target="_blank">Graphite</a>  👈
    * **PR #13770** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13770" target="_blank">Graphite</a> 
      * **PR #13771** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13771" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13769?utm_source=stack-comment).

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1860,6 +1863,364 @@ def delete_dynamic_partition(self, partitions_def_name: str, partition_key: str)
                 )
             )
 
+    @property
+    def supports_global_concurrency_limits(self) -> bool:
+        return self.has_table(ConcurrencySlotsTable.name)
+
+    def allocate_concurrency_slots(self, concurrency_key: str, num_int: int) -> None:
```
alangenfeld, line 1870: this removes and adds right? so maybe `set_` / `resize_` or something instead of allocate. `num_int` -> `num_desired` 

thoughts on starting with an initial tightish limit that we can relax later instead of starting unchecked?

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1860,6 +1863,364 @@ def delete_dynamic_partition(self, partitions_def_name: str, partition_key: str)
...
+                        "step_key": None,
+                        "deleted": False,
+                    }
+                    for _ in range(count, num_int)
+                ]
+                conn.execute(ConcurrencySlotsTable.insert().values(rows))
+                keys_to_assign = [concurrency_key for _ in range(count, num_int)]
+
+        if keys_to_assign:
+            self.assign_pending_steps(keys_to_assign)
```
alangenfeld, line 1941: this took me a bit to discern, add comments

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1860,6 +1863,364 @@ def delete_dynamic_partition(self, partitions_def_name: str, partition_key: str)
...
+                        ConcurrencySlotsTable.c.concurrency_key == concurrency_key,
+                        ConcurrencySlotsTable.c.deleted == False,  # noqa: E712
+                    )
+                )
+            ).fetchone()
+        pending_count = cast(int, pending_row[0]) if pending_row else 0
+        slots_count = cast(int, slots[0]) if slots else 0
+        return slots_count > pending_count
+
+    def can_claim_from_pending(self, concurrency_key: str, run_id: str, step_key: str):
```
alangenfeld, line 1969: nit: return types on many of these fns

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1860,6 +1863,364 @@ def delete_dynamic_partition(self, partitions_def_name: str, partition_key: str)
...
+                                concurrency_key=concurrency_key,
+                                priority=priority or 0,
+                                assigned_timestamp=db.func.now() if should_assign else None,
+                            )
+                        ]
+                    )
+                )
+            except db_exc.IntegrityError:
+                # do nothing
+                pass
```
alangenfeld, line 2049: this is unique contraint guard + no better way to do it across supported DBs?

python_modules/dagster/dagster/_core/storage/event_log/base.py
```diff
@@ -411,3 +411,48 @@ def alembic_version(self) -> Optional[AlembicVersion]:
...
+
+    @abstractmethod
+    def claim_concurrency_slot(
+        self, concurrency_key: str, run_id: str, step_key: str, priority: Optional[int] = None
+    ):
+        """Claim concurrency slots for step."""
+        raise NotImplementedError()
+
+    @abstractmethod
+    def free_concurrency_slots(self, run_id: str, step_key: Optional[str] = None):
```
alangenfeld, line 443: might be better to have one method for freeing a slot and one for clearing a whole run, just to reduce chance of accidental missing `step_key` clearing all bug

python_modules/dagster/dagster/_core/storage/event_log/base.py
```diff
@@ -411,3 +411,48 @@ def alembic_version(self) -> Optional[AlembicVersion]:
...
+    @abstractmethod
+    def can_claim_from_pending(self, concurrency_key: str, run_id: str, step_key: str):
+        """Indicates that a given step can now claim a concurrency, given its priority relative to
+        other pending steps
+        .
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def remove_pending_step(self, run_id: str, step_key: Optional[str] = None):
```
alangenfeld, line 456: similar to above, does this deserve its own entry point? This should only happen in 
1. succesful transfer to concurrency slot
2. abandon everything for run 
?

python_modules/dagster/dagster/_core/storage/event_log/base.py
```diff
@@ -411,3 +411,48 @@ def alembic_version(self) -> Optional[AlembicVersion]:
...
+        """Claim concurrency slots for step."""
+        raise NotImplementedError()
+
+    @abstractmethod
+    def free_concurrency_slots(self, run_id: str, step_key: Optional[str] = None):
+        """Frees concurrency slots for a given run/step."""
+        raise NotImplementedError()
+
+    @abstractmethod
+    def can_claim_from_pending(self, concurrency_key: str, run_id: str, step_key: str):
```
alangenfeld, line 448: feel like theres a better name here though not sure what yet

alangenfeld (changes requested):
overall looking very solid

can you add an excalidraw or other visualization of the state machine to this PR? 

think we need some more tests for resizing up and down & with slots in use 

should there be any specific tests for the "worst case" timings like two of X calls at the same time ? Maybe a thread pool test which stresses the locks? 

to your queue

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
prha: yeah, +1 on the limit.

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
prha: The pattern I've seen is to override with the postgres/mysql variants.  But I think this is the generic way to do it.

python_modules/dagster/dagster/_core/storage/event_log/base.py
prha: +1

python_modules/dagster/dagster/_core/storage/event_log/base.py
prha: Initially, I had it the way you described, where claiming removed a step from `pending_steps`.  But the math / locking logic got too complicated across tables, so now there is a row in `pending_steps` for all in flight steps including ones that are currently occupying a slot.  This means that every executing step should occupy a row in the slots table and the pending steps table.

This is the state progression for steps that want to claim a slot:
1.  Step is waiting for a slot to open up (unassigned).  This state is skipped if there are more open slots than pending steps (e.g. the pending step is assigned immediately on write).
2.  A slot has opened up and the highest priority step is tapped to claim it but has not yet claimed it (assigned).
3.  The assigned step has woken up from sleep in the executor and has claimed an open slot.
4. The step has completed so the slot row is cleared and the pending step row is removed.

prha: But yeah, maybe we don't need this entry point...  It's called with:
1. free up slot
2. abandon everything

I initially exposed this to clean up issues where the executor hits an error, but probably we can just abandon everything and not care that we already know it has only added a pending step (not a slot).

alangenfeld (approved):
i think this is good to go 

may want to wait til whole stack is accepted to avoid any further  DB iterations being a headache

schrockn:
My biggest concern around this is preventing the user from getting themselves into a state where all slots are occupied that they don't realize what has happened. This also connects to your comment in the summary for https://github.com/dagster-io/dagster/pull/13772 about failed runs and my request for changes there.

We are relying on the interaction with the monitoring daemon to make sure that things get cleaned up if I am understanding correctly. Namely if a run worker process dies in an unexpected fashion, we are relying on the monitoring daemon to resume or clean up that run. 

A few things come to mind:

* When the run launcher does not support `check_run_worker_health`, ignore the concurrency key completely, and warn the user that this is not respected in that case. This will prevent users from getting themselves into a state where zombie runs occupy slots forever and are never cleaned up.
* We need to be disciplined about understanding the relationship between run status and status of the concurrency slots of that run status. Namely we should just make it a rule that you can never enter a finished status (defined in `FINISHED_STATUSES`) without having free'ed all concurrency slots first, and enforce that at a very low level and make the operation to free slots idempotent. We need to think about cases where both the run worker process dies at any point, and where the monitoring daemon itself dies at any point, and that we always eventually get to a good state. I don't think this stack does this right now as `free_concurrency_slots_for_run` is only called in `monitor_canceled_run? I might be missing something here.

In the case that I have missed something here about the state of runs, then I think that means that we have to document it better.

And definitely as an editorial comment/self criticism I wish we would have been able to consolidate on a single step delegating executor before this was implemented, as it would have made implementation more straightforward.

prha:
Yeah, I agree that the observability/management tools are pretty important here.

In [IRL] discussion with @alangenfeld, we determined that for V0 (e.g. experimental, soliciting feedback from beta partners), bailing to the user to manually intervene via Dagit is acceptable.  Some fast follow work includes:
* Adding daemon configuration to free slots for runs after some amount timeout.  I've more or less punted on https://github.com/dagster-io/dagster/pull/13772 as a non-blocker, but if you think that is a requirement for v0, I'm happy to start iterating on it again.
* Injecting custom event log engine events into runs blocked by these concurrency limits, pointing them to the configuration page
* More iteration on the UI for fine-grained visibility into slots.

re: step delegating executor - I treated this as basically a non-issue.  The vast majority of the work was done within `ActiveExecution` precisely so that the amount of unique executor integration would be minimal.

schrockn:
I think nailing the run status state machine/slots interaction is very important and will save lots of potential heartache with our beta partners. They are using the system are pretty serious scale and anyone using spot/interruptible instances will run into this.

PR #13770, opened 2023-04-19
Hook up storage concurrency methods to active execution

prha:
## Summary & Motivation
Hook up concurrency storage methods to the active execution.

Should note that the locks are claimed and released in the executor (as opposed to the step workers). 

TBD:  Still need to add support to the dask executor

## How I Tested These Changes
BK


prha:
* master
  * **PR #13769** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13769" target="_blank">Graphite</a> 
    * **PR #13770** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13770" target="_blank">Graphite</a>  👈
      * **PR #13771** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13771" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13770?utm_source=stack-comment).

python_modules/dagster/dagster/_core/storage/tags.py
```diff
@@ -68,6 +68,8 @@
     MAX_RUNTIME_SECONDS_TAG,
 ]
 
+GLOBAL_CONCURRENCY_TAG = f"{SYSTEM_TAG_PREFIX}pool"
```
alangenfeld, line 71: is this a locked name?

python_modules/dagster/dagster/_core/execution/plan/active.py
```diff
@@ -457,6 +514,8 @@ def handle_event(self, dagster_event: DagsterEvent) -> None:
         step_key = cast(str, dagster_event.step_key)
         if dagster_event.is_step_failure:
             self.mark_failed(step_key)
+            if self._global_concurrency_keys and self._run_id:
+                self._instance.event_log_storage.free_concurrency_slots(self._run_id, step_key)
```
alangenfeld, line 518: does this just eagerly call it for every step instead only for steps with claims ?

python_modules/dagster/dagster/_core/execution/plan/active.py
```diff
@@ -44,17 +46,23 @@ def _default_sort_key(step: ExecutionStep) -> float:
...
 
     def __init__(
         self,
+        instance: DagsterInstance,
         execution_plan: ExecutionPlan,
         retry_mode: RetryMode,
         sort_key_fn: Optional[Callable[[ExecutionStep], float]] = None,
         max_concurrent: Optional[int] = None,
         tag_concurrency_limits: Optional[List[Dict[str, Any]]] = None,
+        run_id: Optional[str] = None,
```
alangenfeld, line 63: Might be good to make A Thing we pass in here that handles the global concurrency interactions and have a different variant you can pass in when things are disabled or in the delegated case where the claims were already acquired. That would allow us to put The Thing under test as well as pass a mock in to ActiveExecution to verify its interactions with it.

alangenfeld (commented):
`register_steps` is a doozy, similar pattern to how we deal with retries. I think its worth some thought on how to deal with it instead of just a defaulting bool. 

should we emit engine events about things being blocked on global concurrency at some frequency? should we get "position" information back from the can claim check to be able to show?

python_modules/dagster/dagster/_core/execution/plan/active.py
```diff
@@ -247,26 +270,39 @@ def _update(self) -> None:
...
+
     def sleep_til_ready(self) -> None:
         now = time.time()
-        sleep_amt = min([ready_at - now for ready_at in self._waiting_to_retry.values()])
+        ready_at_times = []
+        if self._waiting_to_retry:
+            ready_at_times.extend(self._waiting_to_retry.values())
+        if self._waiting_to_claim:
+            ready_at_times.extend(self._waiting_to_claim.values())
+        sleep_amt = min([ready_at - now for ready_at in ready_at_times])
```
alangenfeld, line 288: i think this looks right but I think its worth trying to explicitly have a test with a step in both retry delay and claim delay

alangenfeld (changes requested):
to you for now - might need to talk this one over in sync

python_modules/dagster/dagster/_core/storage/tags.py
prha: not a locked name... I think I probably prefer having an explicit `concurrency_key` arg on the op definition, rather than using a special tag key.

And I like using `concurrency_key` over `pool`, but it can get confusing with both run-scoped op concurrency and run-level concurrency.

python_modules/dagster/dagster/_core/execution/plan/global_concurrency_context.py
```diff
@@ -0,0 +1,102 @@
...
+)
+
+from typing_extensions import Self
+
+from dagster._core.instance import DagsterInstance
+
+CONCURRENCY_CLAIM_BLOCKED_INTERVAL = 1
+
+
+class GlobalConcurrencyContext:
```
alangenfeld, line 17: this doc block would be a good place to overview the setup for how this all works

python_modules/dagster/dagster/_core/executor/multiprocess.py
```diff
@@ -187,137 +188,143 @@ def execute(
...
-                            serializable_error = serializable_error_info_from_exc_info(
-                                sys.exc_info()
-                            )
-                            step_context = plan_context.for_step(
-                                active_execution.get_step_by_key(key)
-                            )
+            with GlobalConcurrencyContext(
+                plan_context.instance, plan_context.run_id
+            ) as global_concurrency_context:
+                with ActiveExecution(
```
alangenfeld, line 194: do these all exit at the same time? might be able to remove some indentation with an ExitStack

python_modules/dagster/dagster/_core/storage/tags.py
alangenfeld: `global_concurrency_key` ? what happens if we change this - nothing right since its just how the user process interacts with user definitions so its isolated ?

alangenfeld (commented):
looking good

python_modules/dagster/dagster/_core/storage/tags.py
alangenfeld: "deployment scoped" or "instance scoped" might be more appropriate names than "global"

alangenfeld (approved):

PR #13777, opened 2023-04-19
Optimize `get_partition` method for time window partitions

clairelin135:
Cloud users are reporting sensors timing out when yielding many partitioned run requests.

Currently for each run request, we call `partitions_def.get_partitions()` to check if the partition exists. This becomes an expensive operation when this call occurs for many run requests.

This PR updates the run request resolution to make an optimized call to `get_partition()`, which allows us to bypass iterating through all partitions.

Also adjusts `run_request_for_partition` to utilize the `get_partition` code path to speed up performance.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/partition.py
+++ b/python_modules/dagster/dagster/_core/definitions/partition.py
@@ -11,7 +11,6 @@ from enum import Enum
 from typing import (
     Any,
     Callable,
-    Dict,
     Generic,
     Iterable,
     List,
@@ -1202,35 +1201,3 @@ class DefaultPartitionsSubset(PartitionsSubset[T_cov]):
     @classmethod
     def empty_subset(cls, partitions_def: PartitionsDefinition[T_cov]) -> "PartitionsSubset[T_cov]":
         return cls(partitions_def=partitions_def)
-
-
-class CachingPartitionsLoader:
-    """Caches the partitions returned from a partition definition's get_partitions method.
-    Meant to be used within a tick, when the partitions do not change over the course of the tick.
-    """
-
-    def __init__(
-        self,
-        current_time: Optional[datetime] = None,
-        dynamic_partitions_store: Optional[DynamicPartitionsStore] = None,
-    ):
-        self._current_time = current_time
-        self._dynamic_partitions_store = dynamic_partitions_store
-        self._partitions_by_def: Dict[PartitionsDefinition, Sequence[Partition]] = {}
-
-    def get_partition(
-        self,
-        partitions_def: PartitionsDefinition,
-        partition_key: str,
-    ) -> Partition:
-        if partitions_def not in self._partitions_by_def:
-            self._partitions_by_def[partitions_def] = partitions_def.get_partitions(
-                current_time=self._current_time,
-                dynamic_partitions_store=self._dynamic_partitions_store,
-            )
-
-        for partition in self._partitions_by_def[partitions_def]:
-            if partition.name == partition_key:
-                return partition
-
-        raise DagsterUnknownPartitionError(f"Could not find a partition with key `{partition_key}`")
--- a/python_modules/dagster/dagster/_core/definitions/partition.py
+++ b/python_modules/dagster/dagster/_core/definitions/partition.py
@@ -837,16 +837,13 @@ class PartitionedConfig(Generic[T_cov]):
         current_time: Optional[datetime],
         dynamic_partitions_store: Optional[DynamicPartitionsStore],
     ) -> Partition[T_cov]:
-        matches = [
-            p
-            for p in self.partitions_def.get_partitions(
-                current_time=current_time, dynamic_partitions_store=dynamic_partitions_store
-            )
-            if p.name == partition_key
-        ]
-        if len(matches) == 0:
-            raise DagsterUnknownPartitionError(f"No partition for partition key `{partition_key}`.")
-        return matches[0]
+        partition = self.partitions_def.get_partition(
+            partition_key,
+            current_time=current_time,
+            dynamic_partitions_store=dynamic_partitions_store,
+        )
+
+        return partition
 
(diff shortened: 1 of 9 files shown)


python_modules/dagster/dagster/_core/definitions/partition.py
```diff
@@ -1201,3 +1202,26 @@ def __repr__(self) -> str:
...
+
+class CachingPartitionsLoader:
+    def __init__(self):
+        self._partitions_by_def: Dict[PartitionsDefinition, Sequence[Partition]] = {}
+
+    def get_partition(
+        self,
+        partitions_def: PartitionsDefinition,
+        partition_key: str,
+        current_time: Optional[datetime] = None,
```
gibsondan, line 1215: This eseentially ignores current_time if the partitions_def is already in the cache, right? is that bad? Maybe not in practice? Not sure what exactly current_time is used for. Maybe it should go in the constructor of the object though?

OwenKephart: current_time is used so that time-based partitions know when their latest partition is (partitions only pop into existence after their time window is over).

within the context of tick, this shouldn't change (of course it's possible to have this happen, but it's rare and probably not worth worrying about, as if a partition exists at T_0, then it will exist at T_0+1).

not against putting it in the constructor, but don't think it makes a big difference for now, as in general the current_time that's passed in here seems to be static throughout the duration of the tick anyway

clairelin135: `current_time` is used to determine which time window partitions are existent at the given time. Within the context of a tick, this won't change.

But i've gone ahead and moved these args to the constructor. Sensors will still pass in `current_time=None` since they don't have a notion about current time, but I added comments to clarify that the partition loader is intended to be used within a tick (because the partitions don't change during the course of a tick).

OwenKephart (changes requested):
I think at a high level, there's a better way to approach this. This solution still requires that we resolve all of the partition keys once, which seems needlessly expensive considering all we want to do is determine if this is a valid partition key or not.

In fact, in the case of TimeWindowPartitionsDefinitions (the impetus for this change), we already have an `is_valid_partition_key` method implemented, which does not require any partitions to be resolved.

My proposal would be:

1. make `is_valid_partition_key` a method on the base PartitionDefinition class (it can have a default implementation that just fetches all the partitions)
2. just call this `is_valid_partition_key` method to validate that we have a valid partition key, rather than doing partition = partitions_def.get_partition(...)

because this leaves you with a partition key (which is a string) instead of a `Partition` object, you'll need to use `partitioned_config.get_tags_for_partition_key` instead of `partitioned_config.get_tags_for_partition`. however, that implementation actually [just calls get_partition ](https://sourcegraph.com/github.com/dagster-io/dagster/-/blob/python_modules/dagster/dagster/_core/definitions/partition.py?L831)under the hood, so you'll need to rearrange those functions (really get_tags_for_partition should call get_tags_for_partition_key, as get_tags_for_partition only uses the key (partition.name) off of the partition anyway)

EDIT: Oops this is slightly wrong, as technically get_tags_for_partition does actually supply the full Partition object to self._tags_for_partition_fn. So an alternative approach (even simpler!) would be to just supply an efficient `get_partition` to override the default (slow) implementation on the TimeWindowPartitionsDefinition class.

This could basically just 1) check that it's a valid partition key, then 2) return a `Partition(partition_key, time_window_for_partition_key(partition_key))`

clairelin135:
Hey @OwenKephart, thanks for taking a look.

I also thought about this too. One problem with `is_valid_partition_key` is that it actually isn't accurate--it checks that the format is correct and that the date is after the start time, but it doesn't actually check that the partition key falls on the start of a valid time window. E.g. `weekly_partitions_def.is_valid_partition_key("2023-04-19")` will return true even though it should be false, since today is a wednesday.

We're currently getting away with this for time window partitions subsets because we convert partition keys into valid time windows via `time_windows_for_partition_keys` through the cron string iterator object, which is the expensive operation.

In order to make `is_valid_partition_key` actually check that the key falls on a partition time window, we'd similarly have to create the expensive cron string iterator object, potentially for each partitioned run request yielded. I think the current solution of fetching the partition keys once per partitions def and caching that information may be cheaper.

clairelin135:
Chatted with @OwenKephart offline about this, he suggested I do some perf testing to see how an optimized get partition method would perform.

Evaluation time for 200 time partitioned run requests:
- Current implementation takes 30s
- Caching implementation takes 1.5s
- Optimized get partition method takes 2s

I think this implementation is better than the previous caching version since we won't need to do additional work to maintain it over time, and the execution time is pretty much the same.

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -244,6 +248,46 @@ def get_partitions(
 
         return partitions
 
+    def _get_validated_time_window_for_partition_key(
+        self, partition_key: str, current_time: Optional[datetime] = None
+    ) -> Optional[TimeWindow]:
+        """Returns a TimeWindow for the given partition key if it is valid, otherwise returns None.
+        """
+        try:
```
sryza, line 256: Could we invoke `has_partition_key(...)` here instead of duplicating the implementation?

clairelin135: We could. I originally structured it this way so we wouldn't have to fetch the time window twice. If we called has_partition_key, we'd still then have to refetch the time window if has_partition_key returned true

sryza: would it make sense to factor our the implementation to a private function that's used by both?

sryza: Ohhh I misunderstood

sryza (approved):

PR #13778, opened 2023-04-19
Remove timeWindowMetadata from the GraphenePartitionDefinition class

gibsondan:
Summary:
This has no callsites and may be expensive since it is initialized in the constructor of the graphql object. Try removing it.

Test Plan: BK


clairelin135 (approved):
LGTM

PR #13783, opened 2023-04-19
[dagit] If there is no asset selection, don’t fetch any asset partition definitions

bengotow:
## Summary & Motivation

Discussion: https://elementl-workspace.slack.com/archives/C03CA4TVCAW/p1681931733988229

If you open the launchpad with no assetSelection (eg: for a standard non-asset run), we pass `undefined` to the query, which omits the query parameter on `assetNodes(assetKeys: $assetKeys)`, loading ALL asset nodes in the repo.

Note: This is just the front-end fix -- we also need to update the backend to discern between `undefined` and `empty array` and return nothing in the empty array case.

## How I Tested These Changes

I checked in Dagit that opening the launchpad before this change passed no params to `assetNodes()`, and that after this change, it passes `assetNodes(assetKeys: [])`

### DIFF
--- a/js_modules/dagit/packages/core/src/launchpad/ConfigEditorConfigPicker.tsx
+++ b/js_modules/dagit/packages/core/src/launchpad/ConfigEditorConfigPicker.tsx
@@ -165,7 +165,9 @@ const ConfigEditorPartitionPicker: React.FC<ConfigEditorPartitionPickerProps> = 
       variables: {
         repositorySelector,
         partitionSetName,
-        assetKeys: assetSelection?.map((selection) => ({path: selection.assetKey.path})),
+        assetKeys: assetSelection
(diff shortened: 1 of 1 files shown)


gibsondan:
https://github.com/dagster-io/dagster/pull/13784/files will make this actually have an effect on the server - we will need to make sure that there aren't any callsites that are doing the reverse and sending in an empty list expecting to get all asset nodes (probably unlikely but you never know)

salazarm (approved):
thanks for fixing this!

bengotow:
Good call @gibsondan - I'll do a quick run through and make sure we don't pass an empty array anywhere expecting all results. I don't think we do but can't hurt to check!

bengotow:
Errors in CI are not related, going to go ahead and merge this 👍

PR #13784, opened 2023-04-19
Distinguish between an empty list of asset keys and passing None to the assetKeys parameter

gibsondan:
Summary:
I don't love that this requires callsites to distinguish between an empty list and None, but this makes it so that if you have a query that fetches assetNodes and explicitly passes in [] to the list of asset keys, it will return onone of them instead of all of them.

Test Plan: BK

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/schema/roots/query.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/roots/query.py
@@ -742,10 +742,16 @@ class GrapheneDagitQuery(graphene.ObjectType):
         pipeline: Optional[GraphenePipelineSelector] = None,
         assetKeys: Optional[Sequence[GrapheneAssetKeyInput]] = None,
     ) -> Sequence[GrapheneAssetNode]:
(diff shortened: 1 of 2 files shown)


python_modules/dagster-graphql/dagster_graphql/schema/roots/query.py
```diff
@@ -742,10 +742,16 @@ def resolve_assetNodes(
...
-        use_all_asset_keys = len(resolved_asset_keys) == 0
+        if assetKeys == []:
+            return []
+        elif not assetKeys:
+            use_all_asset_keys = True
+            resolved_asset_keys = None
+        else:
+            use_all_asset_keys = False
+            resolved_asset_keys = set(
+                AssetKey.from_graphql_input(asset_key_input) for asset_key_input in assetKeys or []
```
bengotow, line 753: Seems fine to leave it, but I think the ` or []` here can be removed now!

bengotow (approved):
LGTM!

PR #13789, opened 2023-04-19
add fetch depth 1000 so that the mdx preview link does not fail

prha:
## Summary & Motivation
We compare the head to the branch base to get a list of links to generate.

If there are more than one commits in the branch, this command will fail (see https://github.com/dagster-io/dagster/actions/runs/4734422024).

This PR sets the depth to 1000, which I think is the standard with the original checkout anyway ( how github actions powers the `on: paths:` directive.

## How I Tested These Changes
This very PR

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -19,6 +19,8 @@ jobs:
     steps:
       - name: Checkout PR branch
         uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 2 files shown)


benpankow (approved):
:no

prha:
Seems to add ~10-20s of latency... we can adjust this back down to around 100 if that's a problem

PR #13794, opened 2023-04-20
Add `node_info_to_auto_materialize_policy_fn` to `load_assets_from_dbt_*`

askvinni:
## Summary & Motivation
With the launch of `AutoMaterializePolicy`s, I had an immediate use case to implement them into my dbt assets. The proposed solution follows the same pattern as the `node_info_to_freshness_policy_fn`, where users can specify the auto materialization policy by default by adding it to the dbt model config under `dagster_auto_materialize_policy: {"type": "eager" | "lazy"}`.

## How I Tested These Changes
Added unit test and tested locally, run was successfully kicked off. Granted, I'm not sure this small test will be enough, but happy to add more if necessary.

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
@@ -31,6 +31,7 @@ from dagster import (
 )
 from dagster._config.field import Field
 from dagster._config.field_utils import Permissive
+from dagster._core.definitions.auto_materialize_policy import AutoMaterializePolicy
 from dagster._core.definitions.events import (
     AssetMaterialization,
     AssetObservation,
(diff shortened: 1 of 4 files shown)


askvinni:
Just noticed the `AutoMaterializePolicy` argument in `AssetsDefinition` is marked as experimental. Let me know if I should add a similar flag here.

OwenKephart:
hey @askvinni ! This looks great, thanks for putting this out :)

Just one small issue, this new argument needs to be added here: https://sourcegraph.com/github.com/dagster-io/dagster/-/blob/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/asset_defs.py?L321 as well (it's fine if it's just locked to the default function)

askvinni:
@OwenKephart Thanks for catching! Though since `AssetsDefinitionCacheableData` doesn't take `AutoMaterializePolicy`s, I took the liberty to take this offline so we can discuss a possible extension to this functionality, otherwise dbt cloud users would be missing out. I'm happy to take that as well (maybe in a separate PR) since I'm already on it.

OwenKephart (approved):
thanks again for implementing this!

PR #13801, opened 2023-04-20
[dagster-gcp] Dataproc resource upgrade to pythonic resource

jamiedemaria:
## Summary & Motivation
Rough first pass at converting the Dataproc resource to begin gathering feedback. This PR:
- renames `DataprocResource` class to `DataprocClient` (this class was not exported, so i think this should be fine)
- writes a new `DataProcResource` configurable resource
- The new resource can accept cluster config as a yaml file, json file, or python dictionary (we could trim this down to two options if we want)
- the new resource does not do any validation of the cluster config (we could try to add this if it seems valuable)
- the old resource is not updated to use the config of the new resource for 2 reasons: 1 - the config schema names are in camelCase and so a change to snake_case would be breaking, 2- the resource expects cluster config directly in the config schema. We could make the old resource accept the new resource config in addition to the old style config, and put deprecation warnings on old style config 

Lots still to do in this PR, most of which is testing, but i'd like to get a first pass of feedback on the general approach!

### DIFF
--- a/python_modules/libraries/dagster-gcp/dagster_gcp/dataproc/resources.py
+++ b/python_modules/libraries/dagster-gcp/dagster_gcp/dataproc/resources.py
@@ -1,10 +1,11 @@
 import json
 import time
 from contextlib import contextmanager
-from typing import Any, Mapping, Optional
+from typing import Any, Dict, Mapping, Optional
 
+import dagster._check as check
 import yaml
-from dagster import ConfigurableResource, resource
+from dagster import ConfigurableResource, IAttachDifferentObjectToOpContext, resource
 from googleapiclient.discovery import build
 from oauth2client.client import GoogleCredentials
 from pydantic import Field
@@ -155,7 +156,7 @@ class DataprocClient:
             self.delete_cluster()
 
 
-class DataprocResource(ConfigurableResource):
+class DataprocResource(ConfigurableResource, IAttachDifferentObjectToOpContext):
     project_id: str = Field(
         description=(
             "Required. Project ID for the project which the client acts on behalf of. Will be"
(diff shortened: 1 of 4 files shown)


jamiedemaria:
* master
  * **PR #13801** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13801" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13801?utm_source=stack-comment).

python_modules/libraries/dagster-gcp/dagster_gcp/dataproc/resources.py
```diff
@@ -149,9 +155,75 @@ def cluster_context_manager(self):
...
+        with open(path, "r", encoding="utf8") as f:
+            return yaml.safe_load(f)
+
+    def _read_json_config(self, path: str) -> Mapping[str, Any]:
+        with open(path, "r", encoding="utf8") as f:
+            return json.load(f)
+
+    def get_client(self) -> DataprocClient:
+        cluster_config = None
+        if self.cluster_config_json_path:
```
benpankow, line 207: nit: raise an error if more than one config type is specified, rather than silently preferring one

benpankow (commented):
this seems like a very reasonable approach to me

I think no validation is fine for now, presumably Dataproc would complain at some point in the execution if it was passed bad config. The burden on our end of maintaining a config schema seems like a pain.

yuhan (dismissed):
i think it's a set of reasonable new apis. moving it back to your queue for getting the tests green.

> renames DataprocResource class to DataprocClient (this class was not exported, so i think this should be fine)

i did a quick grep and didn't see any mentions either so i think it's ok to rename

> the new resource does not do any validation of the cluster config (we could try to add this if it seems valuable)

i might have missed this in the status quo. what's the validation of cluster config are we talking about?

jamiedemaria:
> i might have missed this in the status quo. what's the validation of cluster config are we talking about?
just the built in config_schema type validation. Since the dataproc cluster config is built out in https://github.com/dagster-io/dagster/blob/master/python_modules/libraries/dagster-gcp/dagster_gcp/dataproc/configs_dataproc_cluster.py there would be the type checking that all config gets

benpankow (approved):
this looks good to me

PR #13802, opened 2023-04-20
[4/n][Runs page] Update columns to Runs Table

salazarm:
## Summary & Motivation

From Linear issue:

- Remove the big pile of tags and move them into a dialog that can be opened from the actions menu on each run.
- Some high-priority tags are still shown inline (snapshot and partition).
- Adds a new "Created by" column that shows either a user tag, sensor tag, or schedule tag. In OSS there may not be a user tag available so we'd just leave this blank.

Changes from figma: https://www.figma.com/file/28bQQedKzttjEyfciPtMQf/Runs-Filtering?node-id=101-5683&t=5JQXy8jKwA5n4PNC-0

## How I Tested These Changes

https://www.loom.com/share/593e6f840251483497cf7d9027c2e9f9


js_modules/dagit/packages/core/src/runs/RunActionsMenu.tsx
```diff
@@ -112,108 +118,119 @@ export const RunActionsMenu: React.FC<{
...
-            </>
-            <MenuExternalLink
-              text="Download debug file"
-              icon="download_for_offline"
-              download
-              href={`${rootServerURI}/download_debug/${run.id}`}
-            />
-            {run.hasDeletePermission ? (
+      <JoinedButtons>
+        <AnchorButton to={`/runs/${run.id}`}>View Run</AnchorButton>
```
hellendag, line 122: Nit: `View run`

js_modules/dagit/packages/core/src/runs/RunCreatedByCell.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export function RunCreatedByCell(props: Props) {
+  const tags = props.run.tags || [];
+
+  const backfillTag = tags.find((tag) => tag.key === DagsterTag.Backfill);
+  const scheduleTag = tags.find((tag) => tag.key === DagsterTag.ScheduleName);
+  const sensorTag = tags.find((tag) => tag.key === DagsterTag.SensorName);
+  const user = tags.find((tag) => tag.key === DagsterTag.User);
+
+  const ret = [];
```
hellendag, line 21: What is `ret`?

hellendag (approved):

js_modules/dagit/packages/core/src/runs/RunCreatedByCell.tsx
salazarm: its just the `ret`urn array

salazarm: could just call it $jsx

PR #13803, opened 2023-04-20
[1/n][Runs Table v2] Duplicate RunsRoot and RunsFilterInput so that we can gate new changes

salazarm:
## Summary & Motivation

We want to gate the new Runs changes being made in https://github.com/dagster-io/dagster/pull/13758/ and https://github.com/dagster-io/dagster/pull/13758/ 

To that end, since all of the changes are in RunsRoot, RunsFilterInput and RunsTable, I'm duplicating these files and will make the changes to these files to make it easier to review what actually changed. 

This is a lot simpler to maintain than adding a ton of conditional branching in the components but it does mean any changes in the mean time need to be made to both. I don't reckon we'll have both versions for long though or that we'll be making other changes to these components soon.

## How I Tested These Changes

Nothing to test these are just carbon copies and I will base my changes on top of these copies to make it easier to review

### DIFF
--- a/js_modules/dagit/packages/core/src/runs/RunsRootNew.tsx
+++ b/js_modules/dagit/packages/core/src/runs/RunsRootNew.tsx
@@ -1,0 +1,290 @@
+import {ApolloError, gql, useQuery} from '@apollo/client';
+import {
+  Alert,
+  Box,
+  Colors,
+  CursorHistoryControls,
+  NonIdealState,
+  Page,
+  Tag,
+  tokenToString,
+} from '@dagster-io/ui';
+import partition from 'lodash/partition';
+import * as React from 'react';
+import {Link} from 'react-router-dom';
+
+import {PYTHON_ERROR_FRAGMENT} from '../app/PythonErrorFragment';
+import {FIFTEEN_SECONDS, useQueryRefreshAtInterval} from '../app/QueryRefresh';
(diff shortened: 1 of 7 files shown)


js_modules/dagit/packages/core/src/runs/RunsRootNew.tsx
```diff
@@ -0,0 +1,290 @@
...
+      ... on InvalidPipelineRunsFilterError {
+        message
+      }
+      ...PythonErrorFragment
+    }
+  }
+
+  ${RUN_TABLE_RUN_FRAGMENT}
+  ${PYTHON_ERROR_FRAGMENT}
+`;
```
hellendag, line 251: Move the gql docs out into their own files that both old and new component files can import from? I'm not sure what weird things might happen if a query is defined identically in multiple places...

hellendag (approved):

js_modules/dagit/packages/core/src/runs/RunsRootNew.tsx
salazarm: I ended up just renaming the queries by appending "New" that way I don't need to shuffle around a ton of imports since the new components are required by the old components and the new components would then need to import the queries from the old components which would cause a circular dependency. So yeah renaming them unique is simplest for now and we'll get rid of the duplication once we merge in the new table/filters

PR #13795, opened 2023-04-20
Allow setting DAGSTER_CHECKOUT_DEPTH for internal compat build

smackesey:
## Summary & Motivation

Allow setting `DAGSTER_CHECKOUT_DEPTH` as defined in https://github.com/dagster-io/internal/pull/5488 for the internal compatibility build.

## How I Tested These Changes

BK, upstack PR that uses `DAGSTER_CHECKOUT_DEPTH`.

PR #13793, opened 2023-04-20
Clean up is_job fields and references

smackesey:
## Summary & Motivation

Internal companion PR: https://github.com/dagster-io/internal/pull/5496

Clean up some lingering job/pipeline branching code after removal of `PipelineDefinition`. In the process, fix a GQL bug where the now non-existent property `is_job` is accessed.

## How I Tested These Changes

Existing test suite, verifying dagit works for executing basic jobs.

PR #13755, opened 2023-04-19
[refactor] Convert graphene to strawberry with libcst

smackesey:
## Summary & Motivation

Convert GQL API from graphene to strawberry using a libcst codemod.

## How I Tested These Changes

Diffed GQL schema output from Strawberry against previous (should be no differences, plus existing GQL tests.

PR #13497, opened 2023-04-11
pipeline -> job symbol renames

smackesey:
Internal companion PR: https://github.com/dagster-io/internal/pull/5417

This is a very large PR that renames `pipeline` to `job` almost everywhere it is used except for module names (these are done in a separate upstack PR) and the GQL schema: in class names, function/method names, arguments, local variables, and docstrings. This is appropriate after folding `PipelineDefinition` into `JobDefinition` in #13327.

While the volume of changes is large, the engineering complexity is low. The changes were made by grepping for "pipeline" and using Pylance "Rename symbol" on any instance I came across. In some cases I also needed to manually follow up to convert untyped references and references in docstrings. In cases where I discovered an untyped reference, I sometimes added type annotations to the containing function as well. When renaming a serdes class or field, I added the appropriate `storage_name` or `storage_field_names` to`@whitelist_for_serdes` to keep the serialization format constant.

I have kept some structure in the commits so that this can be broken up if necessary, though I would like to avoid that due to this having an internal companion PR in which every split will need to be duplicated. Also, though there is structure to the commits (I tried to scope them to individual classes and functions-- see below), fixing errors on rebase has introduced some deviations to that structure, (e.g. if a rebase introduced a new callsite to some `*pipeline*` method, I needed to rename it in a different commit than the initial one commit that renamed all prior instances), so any split will require some additional work in interactive rebase slice-and-dice of commits where I try to corral all renames into the "structured" correct commit.

Aside from standard symbol renames, there was one "non-standard" change where `"pipeline_name"` was replaced with `"job_name"` in the logging tag keys.

Below is a list of all the renames of non-local symbols (the PR also included many local symbol renames). The list uses the following scheme:

- `Foo -> Bar`: a class or top-level function was renamed
- `Foo.*pipeline* -> *job*`: attributes, methods, and method arguments of Foo containing "pipeline" were renamed with "job" equivalent
- `foo(*pipeline*) -> foo(*job*)`: arguments of foo containing "pipeline" were renamed with "job" equivalent
- `Foo.*pipeline* -> Bar.*pipeline*`: class/top-level function was renamed AND attributes/methods/arguments renamed
- `locals`: assorted local variable and docstring renames.

Here is the list:

1. `PipelineRunStatsSnapshot -> DagsterRunStatsSnapshot`
1. `PipelinePythonOrigin -> JobPythonOrigin`
1. `DagsterRun.*pipeline* -> *job*`
[...]

PR #13805, opened 2023-04-20
[refactor] deprecate Partition

smackesey:
## Summary & Motivation

This deprecates `Partition` and replaces uses of `Partition` internally wherever possible with partition keys instead. `Partition` itself needs to be retained because it was not deprecated until now and is still usable in user-defined functions `tags_for_partition`, `run_config_for_partition`, and `partitions_fn`-- but all of these have alternatives now and are deprecated.

In more detail:

- `tags_for_partition_fn` deprecated on `@static_partitioned_config` and `@dynamic_partitioned_config`, replacement arg `tags_for_partition_key_fn`
- Deprecate `run_config_for_partition_fn` and `tags_for_partition_fn` on `PartitionedConfig`
- Add `run_config_for_partition_key_fn` and `tags_for_partition_key_fn` on `PartitioendConfig`
- Convert all internal calls to `get_partitions` and similar methods to partition key counterparts (e.g. `get_partition_keys`)
- Delete `get_partition`, `get_partitions` on all partition classes
- Make `PartitionedConfig` generic in the associated `PartitionsDefinition`
- Make `PartitionsDefinition` generic in the key type instead of `Partition` type

## How I Tested These Changes

Existing test suite (with some tests updated to use partition keys instead of `Partition`).


smackesey:
* master
  * **PR #13795** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13795" target="_blank">Graphite</a> 
    * **PR #13793** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13793" target="_blank">Graphite</a> 
      * **PR #13505** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13505" target="_blank">Graphite</a> 
        * **PR #13497** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13497" target="_blank">Graphite</a> 
          * **PR #13755** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13755" target="_blank">Graphite</a> 
          * **PR #13805** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13805" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13805?utm_source=stack-comment).

clairelin135 (changes requested):
Thanks for removing partition! Happy to see it go.

I left a couple comments, mostly about perf. I think the main piece is switching `get_partition` to `has_partition` so checking if a partition exists doesn't have to be a o(n) operation.

python_modules/dagster/dagster/_core/definitions/partition.py
```diff
@@ -781,69 +815,65 @@ def get_run_config_for_partition_key(
...
         """
-        partition = self._key_to_partition(partition_key, current_time, dynamic_partitions_store)
-        return self.get_run_config_for_partition(partition)
-
-    def get_run_config_for_partition(
-        self,
-        partition: Partition,
-    ) -> Mapping[str, Any]:
-        """Generates the run config corresponding to a partition.
+        if not self.partitions_def.has_partition_key(
```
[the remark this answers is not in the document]
smackesey: >I think everywhere this function is called the partition key is already confirmed to exist via has_partition_key right?

I don't think that's currently true-- see for instance `JobDefinition.execute_in_process`. But I support adding the checks around those calls and moving them out of this function, so I'll do that. In general I think we need _less_ runtime checking of all kinds in internal functions.

>And optionally renaming the function to be get_run_config_for_validated_partition_key

I don't support this for consistency reasons-- all over the codebase we make assumptions that the data passed in is valid, and we typically don't name those functions specially. Validity should be the default assumption, with verification of that validity concentrated at points where we receive user or out-of-process input.

smackesey:
OK, made some changes:

- Add `has_partition_key` to `MultiPartitionDefinition` and `TimeWindowPartitionsDefinition`
- Remove partition key checks and (the arguments only used for this check) from `PartitionedConfig.get_tags_for_partition_key` and `get_run_config_for_partition_key`
- Add new method `PartitionsDefinition.validate_partition_key` which calls `has_partition_key` and throws an error if False. Make sure this is called (if key not already validated) before calls to `PartitionedConfig` functions above.

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
```diff
@@ -243,71 +243,45 @@ def get_partitions_def_for_dimension(self, dimension_name: str) -> PartitionsDef
...
-                " not the dimensions of the partitions definition."
-            )
-
-        partitions_by_dimension: Dict[str, Partition] = {}
+    ) -> bool:
+        partition_key = (
+            partition_key
+            if isinstance(partition_key, MultiPartitionKey)
+            else self.get_partition_key_from_str(partition_key)
+        )
```
clairelin135, line 257: I think we still need the removed check:
```
if partition_key.keys_by_dimension.keys() != set(self.partition_dimension_names):
            return False
```
in case of the partition key's dimension names don't match

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -271,23 +267,6 @@ def _get_validated_time_window_for_partition_key(
 
         return time_window
 
-    def get_partition(
```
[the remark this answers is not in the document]
clairelin135: I think we still need this as `has_partition_key`

python_modules/dagster/dagster/_core/definitions/partition.py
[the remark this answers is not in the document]
clairelin135: that makes sense--this looks good to me

__MACOSX/._cereal-ratings.csv
clairelin135, line 1: probably an accidental file add?

clairelin135 (commented):
Added a couple more comments, otherwise this is looking good to me

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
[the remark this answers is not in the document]
smackesey: It's there (see line 733). It was actually already defined.

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
smackesey: ah right

__MACOSX/._cereal-ratings.csv
smackesey: good catch

python_modules/dagster/dagster/_core/definitions/partition.py
```diff
@@ -693,27 +688,29 @@ def get_partitions(
...
         current_time: Optional[datetime] = None,
         dynamic_partitions_store: Optional[DynamicPartitionsStore] = None,
     ) -> bool:
-        if dynamic_partitions_store is None:
-            check.failed(
-                "The instance is not available to load partitions. You may be seeing this error"
-                " when using dynamic partitions with a version of dagit or dagster-cloud that"
-                " is older than 1.1.18."
-            )
+        if self.partition_fn:
```
clairelin135, line 701: thanks for fixing this!

clairelin135 (approved):
LGTM! Goodbye partition 👋

Issue #13759, opened 2023-04-19
On asset backfill page, sort assets in topological order


sryza:

sryza:
@clairelin135 - @salazarm asked if we could do this on the backend - what do you think?

clairelin135:
definitely possible--I'll file a PR for this

PR #13808, opened 2023-04-20
Topologically sort assets in asset backfill resolver

clairelin135:
Resolves https://github.com/dagster-io/dagster/issues/13759

Fetches the topological ordering of assets in the asset graph, and returns the targeted partitioned assets in topological order.

Added a test and some additional asset nodes in the graphql repo, not sure how it adds thousands of new lines in the snapshots, but I did at least double check that nothing was unintentionally deleted.

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/asset_backfill.py
+++ b/python_modules/dagster/dagster/_core/execution/asset_backfill.py
@@ -129,6 +129,25 @@ class AssetBackfillData(NamedTuple):
             for asset_key, subset in self.target_subset.partitions_subsets_by_asset_key.items()
         }
 
+    def get_targeted_partitioned_asset_keys_topological_order(self) -> List[AssetKey]:
+        """
+        Returns a topological ordering of partitioned asset keys targeted by the backfill.
+
+        For keys in the same topological level, the order is arbitrary.
+        """
+        toposorted_keys = self.target_subset.asset_graph.toposort_asset_keys()
+
+        targeted_toposorted_keys = []
+        for level_keys in toposorted_keys:
+            for key in level_keys:
+                if (
+                    key in self.target_subset.asset_keys
+                    and self.target_subset.asset_graph.get_partitions_def(key) is not None
+                ):
+                    targeted_toposorted_keys.append(key)
+
+        return targeted_toposorted_keys
+
     def get_partitions_status_counts_by_asset_key(
         self,
     ) -> Mapping[AssetKey, Mapping[BackfillPartitionsStatus, int]]:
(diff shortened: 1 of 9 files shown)


python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -129,6 +129,25 @@ def get_num_targeted_partitions_by_asset_key(self) -> Mapping[AssetKey, int]:
             for asset_key, subset in self.target_subset.partitions_subsets_by_asset_key.items()
         }
 
+    def get_targeted_partitioned_asset_keys_topological_order(self) -> List[AssetKey]:
```
sryza, line 132: Nitpick: prefer `Sequence` over `List`

python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -129,6 +129,25 @@ def get_num_targeted_partitions_by_asset_key(self) -> Mapping[AssetKey, int]:
...
+        """
+        Returns a topological ordering of partitioned asset keys targeted by the backfill.
+
+        For keys in the same topological level, the order is arbitrary.
```
sryza, line 136: Will it be deterministic? Otherwise, things might jump around on refreshes?

python_modules/dagster/dagster/_core/execution/backfill.py
```diff
@@ -127,11 +127,15 @@ def is_valid_serialization(self, workspace: IWorkspace) -> bool:
         else:
             return True
 
-    def get_partitions_status_counts_and_totals_by_asset(
+    def get_partitions_status_counts_and_totals_per_asset(
         self, workspace: IWorkspace
-    ) -> Mapping[AssetKey, Tuple[Mapping[BackfillPartitionsStatus, int], int]]:
+    ) -> Sequence[Tuple[AssetKey, Mapping[BackfillPartitionsStatus, int], int]]:
```
sryza, line 132: In Python 3, dictionaries are actually ordered, so I don't think we need to switch this to a sequence.

python_modules/dagster/dagster/_core/execution/asset_backfill.py
clairelin135: Updated this to sort in alphabetical order for keys on the same level

python_modules/dagster/dagster/_core/execution/backfill.py
clairelin135: ooo this is cool

clairelin135:
@sryza updated per your feedback, I think this is ready for another look!

sryza (approved):

PR #13812, opened 2023-04-20
Pin great_expectations !0.16.8

jmsanders:
The latest great expectations release is causing circular import errors in one of our tests on py3.11. I'm going to pin around it while I run down the source of the issue.


benpankow (approved):
📌

PR #13821, opened 2023-04-20
support allPartitions for pure asset backfills

sryza:
## Summary & Motivation

The backfill graphql mutation has long supported an allPartitions argument that specifies that all partitions of the target job should be included of the backfill.

This argument was not supported for pure asset backfills. This PR adds support for it.

The argument allows requesting a kind of backfill that is not requestable via other parameters: a backfill where root assets have different PartitionsDefinitions.

This is not yet exposed from the UI, but a user wanted to be able to launch graphql requests from a sensor.

## How I Tested These Changes
- GraphQL tests that check that correct backfill object (correct target subset) is created when allPartitions=True is supplied
- Backfill tests that check that backfills with root assets that have different PartitionsDefinitiona can be executed successfully.


prha (approved):
implementation seems sound... can you flesh out the summary / test plan?

sryza:
👍 added details

PR #13830, opened 2023-04-21
Fix multi-asset subsetting logic

OwenKephart:
## Summary & Motivation

Imagine you have a single subsettable multi-asset which produces `top`, `a`, and `b`.  you also have a regular `python` asset, which goes between `a` and `b`:

Screenshot 2023-04-24 at 3 29 22 PM

In order to execute these, you'll need to do one of the following:

- First, run just `a`, then run `python`, then run `top` and `b`
- First, run `a` and `top`, then run `python`, then run just `b`

Either of these are equally valid solutions to the subsetting problem, and in this case our framework happens to choose the second option. However, if we think about the final step (run just  `b`), it should not depend solely on the `python` step. If `top` is being executed in the same run, then `b` needs to wait for both `python` AND `top` to complete.

However, the underlying node that produces `top`, `a`, and `b` does not have an input for `top` for us to plug into. This means that our operational dependency graph was incorrect prior to this PR, as `b` would only depend on `python`.

This PR updates the subsetting logic to create a new node definition with the required inputs for us to plug into. In order to do so, we needed a few spicy changes:

- Previously, if multiple nodes had the same name, we would do a reference equality check when creating the repository data. Now, multiple copies of the same node definition might be produced in different code paths that create these new subset nodes. This required implementing an __eq__ method on OpDefinition.

- Previously, because the underlying node definitions were the same for all subsets of an asset, they could all share the same name. Now, we need to create new names for these subsetted nodes (I went with the ol' md5 hash approach). This in turn required updating tests that happened to rely on the old names.


python_modules/dagster/dagster/_core/definitions/assets.py
```diff
@@ -887,6 +890,37 @@ def with_attributes(
...
+        }.difference(asset_subselection)
+        ins = {}
+
+        input_names_by_key = {v: k for k, v in self.keys_by_input_name.items()}
+        output_names_by_key = {v: k for k, v in self.keys_by_output_name.items()}
+        for input_key in input_keys:
+            input_name = input_names_by_key.get(input_key)
+            if input_name is None:
+                # there is no input existing for this key, meaning this is something that is
+                # traditionally produced within the op, so we create a new input
```
sryza, line 910: "traditionally" -> "produced in the op if it's run without a subset selection"?

PR #13831, opened 2023-04-21
Remove plural `AutoMaterializePolicy`s

johannkm:
Probably not worth a backfill, but I think that 

"  `AutoMaterializePolicy`s  "

looks like a typo and we should just refrain from trying to make it plural

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,7 +18,7 @@
 
 ### Core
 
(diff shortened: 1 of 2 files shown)


johannkm:
* master
  * **PR #13831** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13831" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13831?utm_source=stack-comment).

sryza:
I also used this a bunch of places in the docs - do you think we should remove it there too?

erinkcochran87 (approved):

johannkm:
> I also used this a bunch of places in the docs - do you think we should remove it there too?

Do you remember where? I see https://docs.dagster.io/concepts/assets/asset-auto-execution#auto-materialization-and-partitions but the s is inside the code block, I don't mind that as much haha

PR #13833, opened 2023-04-21
[ui] Performance improvements for global search

hellendag:
## Summary & Motivation

Introduce a couple of performance improvements on global search.

- Debounce queries while typing.
  - This delays actually executing the search so that we avoid expensive searches that we don't need.
- Move Fuse.js searching into Web Workers.
  - For very large data collections, Fuse.js is pretty slow. By pushing the cost to a worker thread, we avoid letting Fuse block the main thread.
  - Use two of these Web Workers for global search, one for the "primary" query (most of the defined objects: jobs, schedules, etc.) and one for the "secondary" query (currently assets).
    - Since most definitions should be available soon after the page loads, the primary query should (hopefully) be ready soon after search is opened.
    - With separate threads for these data collections, we can start showing primary results even before the secondary GraphQL query is complete, or before the secondary Fuse query is complete.

Each of these improvements helps a bit on their own, but they work best in conjunction, so I'm adding both in this PR.

When the user opens search:

- Two lazy GraphQL queries are started, primary and secondary.
- When a GraphQL query completes, create a search worker and populate it with the results. The GraphQL query results may be cached initially, in which case the worker should already have a collection to search on.
- While either GraphQL query is loading, a loading spinner appears in the search input.
- When the user types, perform a debounced search. Results arrive async from the worker. If results arrive that don't match the user's current querystring, discard them.

I also made some tweaks to the Fuse options to improve surfacing exact matches.

Since we have other searchable pages throughout the app, we can use search workers elsewhere as well and get Fuse fuzzy searching in a bunch of places.

## How I Tested These Changes

Created Storybook stories to simulate:

- Basic search with not-too-many objects
- Fast primary query, slow secondary query
- Normal primary query, seconcary query with tens of thousands of objects

In each case, open search and verify (visually and with Performance profiling in dev tools) that querying and rendering are snappy.

Repeat search in app to sanity check.

### DIFF
--- a/js_modules/dagit/packages/core/src/search/__stories__/SearchDialog.stories.tsx
+++ b/js_modules/dagit/packages/core/src/search/__stories__/SearchDialog.stories.tsx
@@ -1,8 +1,15 @@
-import {Story, Meta} from '@storybook/react/types-6-0';
+import {MockedProvider} from '@apollo/client/testing';
+import {Meta} from '@storybook/react/types-6-0';
 import * as React from 'react';
 
-import {StorybookProvider} from '../../testing/StorybookProvider';
+import {AnalyticsContext, dummyAnalytics} from '../../app/analytics';
 import {SearchDialog} from '../SearchDialog';
+import {
+  buildPrimarySearch,
+  buildPrimarySearchStatic,
+  buildSecondarySearch,
+  buildSecondarySearchStatic,
+} from '../__fixtures__/Search.fixtures';
 
 // eslint-disable-next-line import/no-default-export
 export default {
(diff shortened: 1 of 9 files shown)


hellendag:
* master
  * **PR #13833** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13833" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13833?utm_source=stack-comment).

js_modules/dagit/packages/core/src/search/__stories__/SearchDialog.stories.tsx
```diff
@@ -1,19 +1,42 @@
...
+  <AnalyticsContext.Provider value={dummyAnalytics()}>
+    <MockedProvider mocks={[buildPrimarySearchStatic(), buildSecondarySearchStatic(10000)]}>
+      <SearchDialog searchPlaceholder="" />
+    </MockedProvider>
+  </AnalyticsContext.Provider>
+);
+
+export const LotsOfAssetsSearch = () => (
+  <AnalyticsContext.Provider value={dummyAnalytics()}>
+    <MockedProvider mocks={[buildPrimarySearch(), buildSecondarySearch(10000)]}>
```
bengotow, line 38: Oh this is awesome to have 🙌 

One thing I've noticed recently is that switching tabs in storybook is taking longer and longer - I wonder if it's actually running all this code every time you click something in the left sidebar and move between stories. Will do a bit more testing and see next week...

js_modules/dagit/packages/core/src/workers/fuseSearch.worker.ts
```diff
@@ -0,0 +1,36 @@
...
+
+        // Consider the empty string as returning no results.
+        const results = queryString ? fuseObject.search(queryString) : [];
+        self.postMessage({type: 'results', queryString, results});
+      }
+    }
+  }
+});
+
+export {};
```
bengotow, line 36: This looks good! It's sorta surprising that Fuse is so slow but I guess it does a LOT more than basic substring matching 😅

bengotow (approved):

PR #13839, opened 2023-04-21
Fix issue with Runs Filter dropdown closing whenever filters change  / data refreshes.

salazarm:
## Summary & Motivation

The root cause of the issue here is that `<RunsFilterInput />` is being rendered in two different spots depending on if we have data or if we're loading. The problem here is that by moving the component to a different spot in the tree, the component loses all of its state. This PR moves it outside of the Loading component so that it's position doesn't change. One issue with this approach was the styling. I had to use a right margin and negative margin in order to get the "Actions" button aligned with the filters properly. This means if we do make more styling changes in that are then the margins may need to be manually adjusted rather than relying on Flex.

## How I Tested These Changes

https://www.loom.com/share/a5601573043246708b8a387865cb82eb


js_modules/dagit/packages/core/src/runs/RunsRootNew.tsx
```diff
@@ -192,29 +192,27 @@ export const RunsRoot = () => {
...
+                  {currentTab !== 'all' ? (
+                    <Box flex={{direction: 'row', gap: 8}}>
+                      {filterTokens
+                        .filter((token) => token.token === 'status')
+                        .map(({token, value}) => (
+                          <Tag key={`${token}:${value}`}>{`${token}:${value}`}</Tag>
+                        ))}
+                    </Box>
+                  ) : null}
+                  <RunsFilterInput
```
salazarm, line 205: I moved this out so that both the `renderError` and this path have the same component structure:
```
<Box>
  <RunsFilterInput>
```
which means React will be able to keep the component's state when swapping between loading and not loading.

js_modules/dagit/packages/core/src/ui/Loading.tsx
```diff
@@ -14,7 +14,17 @@ const BLANK_LOADING_DELAY_MSEC = 500;
 
 export const Loading = <TData extends Record<string, any>>(props: ILoadingProps<TData>) => {
   const {children, purpose, allowStaleData = false, renderError} = props;
-  const {error, data, loading} = props.queryResult;
+  const {error, data: _data, loading} = props.queryResult;
```
salazarm, line 17: I think the issue might actually be due to `useCursorPaginatedQuery` not returning stale data. so maybe this logic should actually live inside `useCursorPaginatedQuery`. Basically because `useCursorPaginatedQuery` never returns stale data when the variables change (might be intentional? I'm not sure) then the `allowStaleData` flag here doesn't do anything since we never give it stale data.

salazarm:
I made this codesandbox to demo the crux of the issue: https://codesandbox.io/s/intelligent-benz-9x6zql?file=/src/App.js

salazarm:
Got some feedback from Dan, I think I'm going to switch to the "slot" approach he recommended here: https://twitter.com/dan_abramov/status/1649464857425485824

js_modules/dagit/packages/core/src/runs/RunsFilterInputNew.tsx
```diff
@@ -253,36 +253,125 @@ export const RunsFilterInput: React.FC<RunsFilterInputProps> = ({
     };
   }, [options]);
 
+  const jobFilter = useStaticSetFilter({
```
salazarm, line 256: Now that the state is preserved I need to remove the conditional hook usage. The hooks are pretty lightweight, all the work happens when they're actually used so its not big overhead. I should probably not fetch values for filters that aren't being used though

bengotow (approved):
Left an inline comment - this seems like a nice solution, but it also seems like we'll likely need this on all pages that are structured in this manner and it'd be nice to see if we could avoid the early `loading` return in RunTable so that actionBarComponents are rendered in the same place consistently.

js_modules/dagit/packages/core/src/runs/RunsRootNew.tsx
```diff
@@ -208,11 +214,7 @@ export const RunsRoot = () => {
                               ))}
                           </Box>
                         ) : null}
-                        <RunsFilterInput
-                          tokens={mutableTokens}
-                          onChange={setFilterTokensWithStatus}
-                          enabledFilters={enabledFilters}
-                        />
+                        {filtersSlot}
```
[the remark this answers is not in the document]
salazarm: @bengotow  I resolved the moving the component issue inside `RunTable` but I think the `usePortalSlot` should stay in the root since the `renderError` path moves the component too.

PR #13841, opened 2023-04-21
[dagster-pyspark] Rough migration to Pythonic resources

benpankow:
## Summary

Very barebones port of `dagster-pyspark` to Pythonic resources.

Renames the non-exported `PySparkResource` to `PySparkClient`, creates a new `PySparkResource` resource which wraps the client and takes an arbitrary config dict.

This dict is type-checked under the hood when the client is constructed, rather than as part of the config schema itself.

## Test Plan

unit tests.

### DIFF
--- a/python_modules/libraries/dagster-pyspark/dagster_pyspark/resources.py
+++ b/python_modules/libraries/dagster-pyspark/dagster_pyspark/resources.py
@@ -1,5 +1,10 @@
+from typing import Any, Dict
+
 import dagster._check as check
-from dagster import resource
+from dagster import ConfigurableResource, IAttachDifferentObjectToOpContext, resource
+from dagster._config.validate import validate_config
+from dagster._core.errors import DagsterInvalidConfigError
+from dagster._utils.cached_method import cached_method
 from dagster_spark.configs_spark import spark_config
 from dagster_spark.utils import flatten_dict
 from pyspark.sql import SparkSession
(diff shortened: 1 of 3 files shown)


benpankow:
* master
  * **PR #13841** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13841" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13841?utm_source=stack-comment).

python_modules/libraries/dagster-pyspark/dagster_pyspark/resources.py
```diff
@@ -28,6 +33,48 @@ def spark_context(self):
...
+                "Error when processing PySpark resource config ",
+                result.errors,
+                self.spark_config,
+            )
+        return PySparkClient(self.spark_config)
+
+    def get_object_to_set_on_execution_context(self) -> Any:
+        return self.get_client()
+
+
```
schrockn, line 77: should we be using lifecycle methods in lieu of IAttachDifferentObjectToOpContext at this point?

schrockn (changes requested):
Should we be using the lifecycle methods here?

python_modules/libraries/dagster-pyspark/dagster_pyspark/resources.py
benpankow: Yup, updated.

schrockn (approved):

jamiedemaria:
I know this just merged, but today i caught a couple pythonic resource upgrades that I forgot to add to the API docs. Looks like this one didn't get added either

Issue #13791, opened 2023-04-20
negative number of requested runs on backfill page


sryza:
Notice the value under "Requested" for daily_asset:

<img width="1186" alt="image" src="https://user-images.githubusercontent.com/654855/233227122-b0798cdf-97d5-4d93-968f-e063a87be90b.png">

How I got here:
* Launched `dagster dev` from the repo root
* Launched a backfill of daily_asset, hourly_asset1, and hourly_asset2, with this selection: "2023-03-24-04:00...2023-03-26-13:00"
* Waited

clairelin135:
Was able to repro this--looks like negative counts are only displaying when certain partitions fail.

Looking into this and will put out a fix

PR #13846, opened 2023-04-21
Fix negative count on asset backfill page error

clairelin135:
Resolves https://github.com/dagster-io/dagster/issues/13791

For asset backfills, if an upstream partition fails, the downstream partition is automatically marked as failed without requesting a run for it.

Previously, we calculated the number of "in progress" partitions as:
```
# requested - (# materialized + # failed)
```
but this resulted in negative counts for in progress partitions since some of the failed partitions were never requested.

This PR fixes the issue by only subtracting failed partitions if they existed as part of the requested subset.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/schema/backfill.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/backfill.py
@@ -315,8 +315,8 @@ class GraphenePartitionBackfill(graphene.ObjectType):
                 GrapheneAssetPartitionsStatusCounts(
                     assetKey=asset_key,
                     numPartitionsTargeted=total_num_partitions,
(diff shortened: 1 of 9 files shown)


sryza (approved):
You don't need to change it at this point, but my instinct would be to test this kind of change using a more tightly scoped test.

I.e. it seems like the error is a logic error inside `AssetBackfillData`, so we shouldn't need all the graphql scaffolding to produce a test that ensures the correct behavior. Instead, a test could construct an `AssetBackfillData` object and then ensure that `get_partitions_status_counts_by_asset_key` returns the expected result.

clairelin135:
@sryza that makes a lot of sense. In the future, I'll do something like the pattern you suggested.

For now, I've added a unit test with the `AssetBackfillData.get_partitions_status_count....` call. I've retained the graphql test for now since it's already been implemented

Issue #13702, opened 2023-04-17
on backfill page, include targeted assets that are not partitioned


sryza:
Currently, if you launch a backfill that includes some partitioned assets and some non-partitioned assets, only the partitioned assets will be shown on the backfill page

The value in the "Partitions targeted" column these assets should be "-"

clairelin135:
@salazarm I have a PR out to update the gql queries to also return unpartitioned assets, do you think you could add the front end changes to display rows for these assets?

PR: https://github.com/dagster-io/dagster/pull/13853

PR #13853, opened 2023-04-21
Return unpartitioned assets in graphQL asset backfill queries

clairelin135:
Backend portion of resolving https://github.com/dagster-io/dagster/issues/13702
 
Amends the asset backfill partition status code paths to also return unpartitioned asset statuses: 
```
type UnpartitionedAssetStatus {
  assetKey: AssetKey!
  inProgress: Boolean!
  materialized: Boolean!
  failed: Boolean!
}
```
When all of the booleans are false, then the asset has not begun execution yet.

The resolver now returns unpartitioned and partitioned asset statuses together, topologically sorted.

Also adds some renames to match what we plan to display in the UI
- requested -> in progress
- complete -> materialized


js_modules/dagit/packages/core/src/instance/backfill/BackfillPage.tsx
```diff
@@ -182,31 +182,33 @@ export const BackfillPage = () => {
...
-                    </div>
-                  </Box>
-                </td>
-                <td>{asset.numPartitionsTargeted}</td>
-                <td>{asset.numPartitionsRequested}</td>
-                <td>{asset.numPartitionsCompleted}</td>
-                <td>{asset.numPartitionsFailed}</td>
-              </tr>
-            ))}
+            {backfill.assetBackfillData?.assetBackfillStatuses.map((asset) => {
```
salazarm, line 185: curious if `assetBackfillData` is actually nullable?

clairelin135: ah yeah good catch--I think before we were still returning an asset backfill data object even with the backfill wasn't an asset backfill. I've updated the resolver to return None for non-asset backfills.

salazarm (commented):
js looks good to me, will defer to Sandy for python

python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -138,45 +133,78 @@ def get_targeted_partitioned_asset_keys_topological_order(self) -> Sequence[Asse
...
-                    asset_key
-                ).get_partition_keys()
-            )
-            requested_partitions = set(
-                self.requested_subset.get_partitions_subset(asset_key).get_partition_keys()
-            )
+        def _get_status_for_asset_key(
+            asset_key: AssetKey,
+        ) -> Union[
+            Tuple[Mapping[BackfillPartitionsStatus, int], int],
```
sryza, line 162: Someone reading this code is likely to need to take some time to decipher the meaning of this data structure. Replacing it with a NamedTuple would make it clearer what's going on.

sryza (approved):
Nice

PR #13854, opened 2023-04-22
[Suggestion Filter] Reset search bar entry when selecting an entry

salazarm:
## Summary & Motivation

As titled, this clears the search bar whenever an entry is selected.

## How I Tested These Changes

Jest


bengotow (approved):
This looks good to me!

PR #13866, opened 2023-04-24
Retry ECS concurrency test on failure

jmsanders:
Our use of concurrency here makes this occassionally flaky. I think this is pretty rare - I've had it running in a tight loop on my local machine for the past 30 min and I haven't seen it flake yet.

For now, let's just retry on failure using:

https://pypi.org/project/pytest-rerunfailures/

And if it ever re-emerges, we can disable the test or rethink our approach to detect regressions in:

https://github.com/dagster-io/dagster/pull/11192

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws_tests/ecs_tests/test_stubbed_ecs.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws_tests/ecs_tests/test_stubbed_ecs.py
@@ -275,6 +275,13 @@ def test_register_task_definition(ecs):
                 cpu="256",
             )
 
+        # Infrequently, our concurrent futures don't fire fast enough to hit
+        # our stubbed ECS's lock. We can force this flaky behavior by firing
+        # thousands of futures; by the time the later futures launch, the
+        # earlier futures have already completed and released their lock.
+        #
+        # We've marked this test as flaky and retry it once on failure to
+        # try to mitigate this.
         futures = [executor.submit(task) for i in range(2)]
 
     # We successfully registered only 1 task definition revision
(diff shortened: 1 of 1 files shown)


schrockn:
Isn't the entire premise of a concurrency test like this that it will only occasionally fail if there is an underlying bug? I don't see how this test has any value.

jmsanders:
In this case, I think it failed because buildkite's runner didn't actually call the `register_task_definition` method concurrently - the time between the futures being launched was longer than our fake ECS test fixture holds its lock.

I can brute force the failure by running a very large _n_ of futures - my computer can't launch each fast enough and the lock from the method call of the first future is released by the time some of the later futures start.

We run tests against our own fake ECS test fixture because the alternatives aren't great: mocking, recording, third party fake implementations, or integration tests. This test in particular is actually a test of the fixture itself.

So I think our best options are to:
- retry because it happens so infrequently
- change our fake to lock for longer (slowing down other tests)
- remove the test of our fixture and incur the risk of our fixture regressing

 I think it's pretty low cost to retry and if it remains a problem, we can rip it out.

jmsanders:
Probably not without some additional work on our end. They have rolled out a bunch of new [test analytics features](https://buildkite.com/test-analytics) in the last year or so that we're not really leveraging, but I suspect anything manages behavior across builds like that would need to be a little bit custom.

This particular change feels like a decent poor man's version though: the build will only fail if the flaky test fails twice in a row within the same build.

schrockn (changes requested):
Ok.

One thing I'd like added to this is an extensive comment next to the assertion that typically flakes/fails so that this is documented.

As an engineer without more context, all I see is flake to hide a concurrency bug.

I'd like to see a comment with the following points:

* Here's a test that tests concurrency.
* Concurrency tests often fail periodically, and therefore flakes are actually a real underlying bug.
* This is not true in this case because XXX. (This explains why a bug will result in non-deterministic failure)
* Because of this, we are going to retry to mitigate harmless flakes.

schrockn (approved):

PR #13869, opened 2023-04-24
[dagster-prometheus] Update to Pythonic resources

benpankow:
## Summary

Small migration update to provide a new `PrometheusResource` which provides a prometheus client.

## Test Plan

Updated unit tests.


schrockn (changes requested):
I'd like to use this a test case for loosening the restrictions on the resource system and consolidating this to one class with mutable internal state. Wdyt? It would require framework changes. We can do it in one PR as an RFC and then split it out later.

benpankow:
Here's a stab at revamping this using a more relaxed resource setup: https://github.com/dagster-io/dagster/pull/13894

benpankow:
Closed in favor of #13894

Issue #13875, opened 2023-04-24
Union types config example is broken in docs


dpeng817:
Union config types are broken in docs. Context: https://docs.dagster.io/concepts/configuration/advanced-config-types#union-types
Error: 
```
../dagster/python_modules/dagster/dagster_tests/core_tests/graph_tests/test_blah.py:30: in <module>
    "pet_stats": ConfigWithUnion(
../dagster/python_modules/dagster/dagster/_config/pythonic_config/__init__.py:176: in __init__
    nested_items = list(check.is_dict(value).items())
../dagster/python_modules/dagster/dagster/_check/__init__.py:434: in is_dict
    raise _type_mismatch_error(obj, dict, additional_message)
E   dagster._check.CheckError: Object Cat(pet_type='cat', meows=10) is not a dict. Got Cat(pet_type='cat', meows=10) with type <class 'dagster_tests.core_tests.graph_tests.test_blah.Cat'>.
```

PR #13876, opened 2023-04-24
[pythonic config][fix] Fix directly constructing config w/ discriminated union

benpankow:
## Summary

Addresses #13875.

Fixes direct construction of discriminated unions, which currently errors:

```python

class Cat(Config):
    pet_type: Literal["cat"] = "cat"
    meows: int

class Dog(Config):
    pet_type: Literal["dog"] = "dog"
    barks: float

class ConfigWithUnion(Config):
    pet: Union[Cat, Dog] = Field(discriminator="pet_type")
    
config = ConfigWithUnion(pet=Cat(meows=10))
```

This behavior was only tested in the "unstructured config" case, so it wasn't caught.

## Test Plan

A few new unit tests

### DIFF
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -205,12 +205,9 @@ class Config(MakeConfigCacheable):
 
                 discriminator_key = check.not_none(field.discriminator_key)
                 if isinstance(value, Config):
-                    value_dict = value._as_config_dict()  # noqa: SLF001
-                    value_discriminator = check.not_none(value_dict.get(discriminator_key))
-                    values_without_key = {
-                        k: v for k, v in value_dict.items() if k != discriminator_key
-                    }
-                    nested_dict = {value_discriminator: values_without_key}
+                    nested_dict = value._as_config_dict_inner(  # noqa: SLF001
+                        discriminator_key=field.discriminator_key, deep=False
+                    )
 
                 nested_items = list(check.is_dict(nested_dict).items())
                 check.invariant(
@@ -226,7 +223,15 @@ class Config(MakeConfigCacheable):
                 modified_data[key] = value
         super().__init__(**modified_data)
 
-    def _as_config_dict(self) -> Mapping[str, Any]:
+    def _as_config_dict_deep(self) -> Mapping[str, Any]:
+        return self._as_config_dict_inner(None, True)
+
+    def _as_config_dict_shallow(self) -> Mapping[str, Any]:
+        return self._as_config_dict_inner(None, False)
+
+    def _as_config_dict_inner(
+        self, discriminator_key: Optional[str] = None, deep: Optional[bool] = False
+    ) -> Mapping[str, Any]:
         """Returns a dictionary representation of this config object,
         ignoring any private fields.
         """
@@ -237,10 +242,18 @@ class Config(MakeConfigCacheable):
             field = self.__fields__.get(key)
             if field and value is None and not _is_pydantic_field_required(field):
                 continue
+
+            output_value = (
+                _config_value_to_dict_representation(self, field, value) if deep else value
+            )
+
             if field:
-                output[field.alias] = value
+                output[field.alias] = output_value
             else:
-                output[key] = value
+                output[key] = output_value
+        if discriminator_key:
+            discriminator_value = output.pop(discriminator_key)
+            output = {discriminator_value: output}
         return output
 
     @classmethod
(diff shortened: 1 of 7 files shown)


benpankow:
* master
  * **PR #13876** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13876" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13876?utm_source=stack-comment).

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -200,15 +200,26 @@ def __init__(self, **config_dict) -> None:
...
+                discriminator_key = check.not_none(field.discriminator_key)
+                if isinstance(value, Config):
+                    value_dict = value._as_config_dict()  # noqa: SLF001
+                    value_discriminator = check.not_none(value_dict.get(discriminator_key))
+                    values_without_key = {
+                        k: v for k, v in value_dict.items() if k != discriminator_key
+                    }
+                    nested_dict = {value_discriminator: values_without_key}
+
+                nested_items = list(check.is_dict(nested_dict).items())
```
schrockn, line 214: shouldn't this logic reside inside of `_as_config_dict`?

schrockn (changes requested):

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
benpankow: The current name doesn't adequately describe what the method does (a shallow flattening of a config class to a dict, a user might expect a deep conversion).  I'm moving the logic into a method in the config class and doing some slight restructuring/renaming to match.

python_modules/dagster/dagster/_core/definitions/run_config.py
```diff
@@ -617,7 +617,7 @@ def _convert_config_classes(configs: Dict[str, Any]) -> Dict[str, Any]:
     return {
         k: {
             "config": config_dictionary_from_values(
-                v._as_config_dict(), v.to_config_schema().as_field()  # noqa: SLF001
+                v._as_config_dict_shallow(), v.to_config_schema().as_field()  # noqa: SLF001
             )
```
schrockn, line 621: So if this call site is calling `config_dictionary_from_values` as well, could this just be a "deep" copy?

schrockn (changes requested):
At this point I think it is appropriate to step back and thing about what we are modeling here. The terminology here around deep and shallow is quite subtle and I think at this point misleading and confusing. I don't know when it is appropriate to call one version versus another.

`_as_config_dict_shallow`: This is enforcing the following rules:

* Skipping internal fields (This step should be gone soon once we land lifecycle methods)
* Skipping optional fields that are none
* Remapping aliases.
* All first-level values in the dictionary are preserved

So this is used in the context where one is only manipulating `keys` or the structure of the dictionary. One could argue it should be a separate function that maps old key names to new key names and return that data structure.

As it is, this function returns a dictionary that is in a weird intermediate state where the _structure_ of dictionary has changed, but the values are not serializable/directly usable. It is up to the programmer to understand this, which is asking a lot.

`_as_config_dict_deep`

Does the dictionary restructuring _and_ then processes it so that it can be consumed by the underlying configuration system. This also returns the representation that is persisted with runs and passed to the UI, etc.

Step 1: structure the keys
Step 2: process the restructured tree

So broadly I think we need to have cleaner terminology here and then structure the code to match that terminology so that our future selves understand what is going on.

benpankow:
With fresh eyes I'm realizing that beyond some very basic shared code the use cases for the so-called 'shallow' and 'deep' methods are pretty different.

I've gone ahead and done a naming pass along with splitting out the shared code to make the code paths a bit easier to trace.

The shallow method (which is really only used in one place) is renamed to `_get_non_none_public_field_values`, which returns a mapping of field names to values for the `Config` object, dropping `None` optional values and private fields (to be dropped later)

The deep method, which is what we're using everywhere else, is renamed to `_convert_to_config_dictionary`, which more accurately represents that it returns the identical "config dict" representation for the `Config` object.

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -201,23 +201,44 @@ def __init__(self, **config_dict) -> None:
...
 
-    def _as_config_dict(self) -> Mapping[str, Any]:
+    def _convert_to_config_dictionary(self) -> Mapping[str, Any]:
+        """Converts this Config object to a Dagster config dictionary, in the same format as the dictionary
+        accepted as run config or as YAML in the launchpad.
+
+        Inner fields are recursively converted to dictionaries, meaning nested config objects
+        or EnvVars will be converted to the appropriate dictionary representation.
+        """
+        return self._convert_to_config_dictionary_inner()
```
schrockn, line 234: Why is this "inner" function separate?

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -226,12 +247,32 @@ def _as_config_dict(self) -> Mapping[str, Any]:
...
                 continue
+
             if field:
                 output[field.alias] = value
             else:
                 output[key] = value
         return output
 
+    def _convert_to_config_dictionary_inner(
+        self,
```
schrockn, line 258: I think we can avoid the code duplication with _get_non_none_public_field_values by calling it in this implementation here.

```
public_fields = self._get_non_none_public_field_values()
return { k: _config_value_to_dict_representation(v) for k, v in public_fields.items() }
```

schrockn (changes requested):
one more cleanup pass

schrockn (approved):
Great this got to a much better spot. 👍🏻

Issue #13834, opened 2023-04-21
MultiAssetSensorEvaluationContext.latest_materialization_records_by_partition_and_asset fails with source assets


sryza:
Reproduction:

```python
from dagster import (
    AssetKey,
    SourceAsset,
    DailyPartitionsDefinition,
    multi_asset_sensor,
    asset,
    repository,
    define_asset_job,
)


@asset(partitions_def=DailyPartitionsDefinition(start_date="2023-03-01"))
def upstream_asset1():
    ...


@asset(partitions_def=DailyPartitionsDefinition(start_date="2023-03-01"))
def upstream_asset2():
    ...


upstream_asset1_source = SourceAsset(
    key=AssetKey("upstream_asset1"),
    partitions_def=DailyPartitionsDefinition(start_date="2023-03-01"),
)

upstream_asset2_source = SourceAsset(
    key=AssetKey("upstream_asset2"),
    partitions_def=DailyPartitionsDefinition(start_date="2023-03-01"),
)


@asset()
def downstream_asset(upstream_asset1, upstream_asset2):
    ...


@multi_asset_sensor(
    monitored_assets=[
        AssetKey("upstream_asset1"),
        AssetKey("upstream_asset2"),
    ],
    job=define_asset_job("foo", selection=[downstream_asset]),
)
def my_sensor(context):
    context.log.info(str(context.latest_materialization_records_by_partition_and_asset()))


@repository
def repo1():
    return [upstream_asset1, upstream_asset2]


@repository
def repo2():
    return [upstream_asset1_source, upstream_asset2_source, downstream_asset, my_sensor]
```

`my_sensor` should fail. I think this can be fixed by including source assets in MultiAssetSensorEvaluationContext._partitions_def_by_asset_key.

PR #13877, opened 2023-04-24
Enables multi asset sensor partitions methods to work for source assets

clairelin135:
Fixes https://github.com/dagster-io/dagster/issues/13834

The multi asset sensor context partition methods currently don't work for source assets since we only fetch existent partitions defs for derived assets. This PR updates this functionality to also fetch partitions definitions for source assets.


sryza (approved):
Nice

PR #13881, opened 2023-04-25
[2/2] Scheduled runs tab update

salazarm:
## Summary & Motivation

Follow up to https://github.com/dagster-io/dagster/pull/13872

The "Scheduled" tab on the Runs page has its own root: ScheduledRunListRoot. This PR updates this root to look similar to RunsRootNew.tsx.

## How I Tested These Changes
Screen Shot 2023-04-25 at 9 50 25 AM


js_modules/dagit/packages/core/src/runs/ScheduledRunsListRoot.new.tsx
```diff
@@ -35,9 +41,35 @@ export const ScheduledRunListRoot = () => {
...
+  const {data: countData} = runCountResult;
+  const {queuedCount, inProgressCount} = React.useMemo(() => {
+    return {
+      queuedCount:
+        countData?.queuedCount?.__typename === 'Runs' ? countData.queuedCount.count : null,
+      inProgressCount:
+        countData?.inProgressCount?.__typename === 'Runs' ? countData.inProgressCount.count : null,
+    };
+  }, [countData]);
+
```
bengotow, line 63: Hmm if we revisit this, it might be nice to break these ~15 lines out into a hook just to make sure any future changes to it are applied in both places

bengotow (approved):

js_modules/dagit/packages/core/src/runs/ScheduledRunsListRoot.new.tsx
salazarm: Yeah good call, let me do that

salazarm: Ah I see, I need to convert the component into a hook because of the combined query refresh state call that happens within `RunRoot`

braunjj (approved):

PR #13883, opened 2023-04-25
Fix undefined array index access

salazarm:
Noticed this error locally for an asset that was never materialized:

Screen Shot 2023-04-25 at 10 04 09 AM


bengotow:
Ahh man this makes sense -- good catch!

bengotow (approved):

js_modules/dagit/packages/core/src/runs/UserDisplay.tsx
```diff
@@ -0,0 +1,23 @@
+import {BaseTag, Box, SubwayDot} from '@dagster-io/ui';
```
salazarm, line 1: oops i'll rm this

salazarm:
Hmm actually, I think it might be due to a different error that this array is `null` in the first place because according to the schema, the array shouldn't be nullable 🤔

PR #13887, opened 2023-04-25
Initialize variables used in __del__ right away in GrpcServerProcess

gibsondan:
Summary:
Without this, if there's an exception constructing the object, there will be another exception in __del__ when it tries to access an attribute that does not exist

## How I Tested These Changes
Buildkite

### DIFF
--- a/python_modules/dagster/dagster/_grpc/server.py
+++ b/python_modules/dagster/dagster/_grpc/server.py
@@ -1193,6 +1193,10 @@ class GrpcServerProcess:
     ):
         self.port = None
         self.socket = None
+        self._waited = False
+        self._shutdown = False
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_grpc/server.py
```diff
@@ -1193,14 +1193,17 @@ def __init__(
     ):
         self.port = None
         self.socket = None
+        self._waited = False
+        self._shutdown = False
+        self._heartbeat = heartbeat
+        self.server_process = None
```
alangenfeld, line 1199: leave a comment near these explaining the situation

alangenfeld (approved):

PR #13889, opened 2023-04-25
Use tan stack virtualizer with TagSelector

salazarm:
## Summary & Motivation

I used a virtualization approach that we stopped using (potentially due to the cross browser issues?).
Updating TagSelector to use the new updated approach.

## How I Tested These Changes
Starting up my windows PC to checkout this PR and see if it fixes the issue.

### DIFF
--- a/js_modules/dagit/packages/ui/src/components/VirtualizedTable.tsx
+++ b/js_modules/dagit/packages/ui/src/components/VirtualizedTable.tsx
@@ -1,6 +1,8 @@
-import {Box, Colors} from '@dagster-io/ui';
 import * as React from 'react';
 import styled from 'styled-components/macro';
+
+import {Box} from './Box';
+import {Colors} from './Colors';
 
 export const HeaderCell = ({children}: {children?: React.ReactNode}) => (
   <CellBox


js_modules/dagit/packages/ui/src/components/VirtualizedTable.tsx
```diff
@@ -0,0 +1,67 @@
+import {Box, Colors} from '@dagster-io/ui';
```
salazarm, line 1: copied this from `core`.

bengotow: Hmm can we remove this file from core now? I think having two copies that export the same things could get confusing

js_modules/dagit/packages/ui/src/components/TagSelector.tsx
```diff
@@ -130,11 +145,12 @@ export const TagSelector = ({
   }, [
     allTags,
     dropdownStyles,
+    items,
     renderDropdown,
     renderDropdownItem,
-    rowHeight,
     selectedTags,
     setSelectedTags,
+    totalHeight,
```
bengotow, line 153: This looks good to me - I think @hellendag has the context on why `react-virtualized` vs `@tanstack/react-virtual`. I wonder if an alternative to the bugs we've been seeing is upgrading to a newer version of react-virtualized (maybe?)

bengotow (approved):

js_modules/dagit/packages/ui/src/components/VirtualizedTable.tsx
salazarm: I'll do this in a follow up

salazarm: (I have to update a bunch of imports)

PR #13893, opened 2023-04-25
Make mode optional on GrapheneRun

johannkm:


johannkm:
* master
  * **PR #13893** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13893" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13893?utm_source=stack-comment).

python_modules/dagster-graphql/dagster_graphql/schema/pipelines/pipeline.py
```diff
@@ -349,7 +349,7 @@ class GrapheneRun(graphene.ObjectType):
     stepKeysToExecute = graphene.List(graphene.NonNull(graphene.String))
     runConfigYaml = graphene.NonNull(graphene.String)
     runConfig = graphene.NonNull(GrapheneRunConfigData)
-    mode = graphene.NonNull(graphene.String)
+    mode = graphene.Field(graphene.String)
```
alangenfeld, line 352: if you look below we have `mode=DEFAULT_MODE_NAME` so im not sure how it could be null

smackesey (approved):

python_modules/dagster-graphql/dagster_graphql/schema/pipelines/pipeline.py
alangenfeld: also have to deal with the        ` interfaces = (GraphenePipelineRun,)` bit think these fields conflict now.

smackesey: I suspect this weirdness has something to do with `GraphenePipelineRun`. Do we even need it? The only place it's used is in `GrapheneRun`. Perhaps we can just delete it and fold any fields it has that aren't also in `GrapheneRun` in?

johannkm:
will be fixed in release-1.3.2, release-1.3.1 didn't have the mode default https://github.com/dagster-io/dagster/blob/release-1.3.1/python_modules/dagster-graphql/dagster_graphql/schema/pipelines/pipeline.py#L379

PR #13894, opened 2023-04-25
[dagster-prometheus] Update dagster-prometheus using Pythonic resources w/ lifecycle hooks

benpankow:
## Sumamry

Reimplementation of #13869 which migrates the `dagster-prometheus` resource as a configurable-resource w/ setup lifecycle hook that sets up some state on the class.


benpankow:
* master
  * **PR #13894** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13894" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13894?utm_source=stack-comment).

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -681,6 +684,10 @@ def __init__(self, **data: Any):
             nested_resources={k: v for k, v in resource_pointers.items()},
             resource_context=None,
         )
+        self.__post_init__()
```
benpankow, line 687: One immediate question is where to invoke this - right now it's after every `__init__` call, but this means it's called both at instantiation time (e.g. in Definitons) and at run time when the resource is fully setup (e.g. env vars populated, nested resources populated).

We may want to only call it as part of the resource function, e.g. after the resource is set up at run time

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -160,6 +160,9 @@ def __setattr__(self, name: str, value: Any):
     def _is_field_internal(self, name: str) -> bool:
         return name.endswith(INTERNAL_MARKER)
 
+    def _is_field_private(self, name: str) -> bool:
+        return name.startswith("_")
```
schrockn, line 164: Specifically I don't think we should do this and instead we should just change the base class definition to tell Pydantic to allow this.

benpankow: That seems reasonable.

Right now, the model doesn't support assignment because it is `frozen = True`. If we disable that, we'll still hit the `ValueError` for trying to initialize unspecified fields, which I think will still require some custom `__setattr__` handling to get around

benpankow: Unless we use `PrivateAttr` - that seems to work ok, based on [this discussion](https://github.com/pydantic/pydantic/issues/655#issuecomment-1289635676):

```python
class PrometheusResource(ConfigurableResource):
  gateway: str = Field(
      description=(
          "The url for your push gateway. Either of the"
          " form 'http://pushgateway.local', or 'pushgateway.local'."
          " Scheme defaults to 'http' if none is provided"
      )
  )
  timeout: int = Field(
      default=30,
      description="is how long delete will attempt to connect before giving up. Defaults to 30s.",
  )
  _registry: prometheus_client.CollectorRegistry = PrivateAttr(default=None)
  
  def __post_init__(self) -> None:
      self._registry = prometheus_client.CollectorRegistry()
```

benpankow:
Updated to use new lifecycle hooks.

python_modules/libraries/dagster-prometheus/dagster_prometheus/resources.py
```diff
@@ -1,21 +1,62 @@
...
+
+    Example:
+    .. code-block:: python
+
+        from dagster import job, op, Definitions
+        from dagster_prometheus import PrometheusResource
+
+        @op
+        def example_prometheus_op(prometheus: PrometheusResource):
+            prometheus.get_client().push_to_gateway(job="my_job")
```
schrockn, line 26: this docblock is now incorrect

schrockn: eliminate `get_client`

python_modules/libraries/dagster-prometheus/dagster_prometheus_tests/test_resources.py
```diff
@@ -30,6 +30,25 @@ def prometheus_solid(context):
     ).success
 
 
+def test_prometheus_counter_pythonic_res() -> None:
+    @op
+    def prometheus_solid(prometheus: PrometheusResource) -> None:
```
schrockn, line 35: s/solid/op

python_modules/libraries/dagster-prometheus/dagster_prometheus_tests/test_resources.py
```diff
@@ -30,6 +30,25 @@ def prometheus_solid(context):
...
+        )
+        c.inc()
+        c.inc(1.6)
+        recorded = prometheus.registry.get_sample_value("some_counter_seconds_total")
+        assert recorded
+        assert abs(2.6 - recorded) < EPS
+
+    assert wrap_op_in_graph_and_execute(
+        prometheus_solid, run_config=ENV, resources=RESOURCES
+    ).success
```
schrockn, line 49: does this test even use the configuration at all?

schrockn (changes requested):
Have some minor suggestions

This is also a bit of a goofy resource in that it take configuration but then that configuration does not impact the collector registry as far as I can tell.  So the configuration only effects the methods we wrap, but then the user is also expected to use the prometheus APIs directly.

For example in our test we configure the resource but I don't think we use it all?

Just requesting changes to make sure you see this question.

PR #13895, opened 2023-04-25
[fix][pythonic config] Fix direct-invocation config behavior with nested Pythonic config

benpankow:
## Summary

Fixes direct op/asset invocation behavior when specifying a config class which is more complex (e.g. nested config). Under the hood, this wasn't properly flatting the passed config into a config dictionary, only the top level of config.

The following now works:

```python
class MyConfig(Config):
    num: int

class MyOuterConfig(Config):
    inner: MyConfig
    string: str

executed = {}

@op
def an_op(config: MyOuterConfig) -> None:
    assert config.inner.num == 1
    assert config.string == "foo"
    executed["yes"] = True

an_op(config=MyOuterConfig(inner=MyConfig(num=1), string="foo"))
```

## Test Plan

New unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/op_invocation.py
+++ b/python_modules/dagster/dagster/_core/definitions/op_invocation.py
@@ -158,7 +158,7 @@ def op_invocation_result(
         from dagster._config.pythonic_config import Config
 
         context = (context or build_op_context()).replace_config(
-            config_input._get_non_none_public_field_values()  # noqa: SLF001
+            config_input._as_config_dict_deep()  # noqa: SLF001
             if isinstance(config_input, Config)
             else config_input
         )
--- a/python_modules/dagster/dagster/_core/definitions/op_invocation.py
+++ b/python_modules/dagster/dagster/_core/definitions/op_invocation.py
@@ -158,7 +158,7 @@ def op_invocation_result(
         from dagster._config.pythonic_config import Config
 
         context = (context or build_op_context()).replace_config(
-            config_input._as_config_dict_deep()  # noqa: SLF001
+            config_input._convert_to_config_dictionary()  # noqa: SLF001
             if isinstance(config_input, Config)
             else config_input
         )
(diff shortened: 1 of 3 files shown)


benpankow:
* master
  * **PR #13895** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13895" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13895?utm_source=stack-comment).

python_modules/dagster/dagster_tests/core_tests/pythonic_config_tests/test_basic_pythonic_config.py
```diff
@@ -894,6 +895,85 @@ def an_op(config: MyConfig) -> None:
...
+    executed = {}
+
+    @op
+    def an_op(config: MyOutermostConfig) -> None:
+        assert config.inner.inner.num == 2
+        assert config.inner.string == "foo"
+        assert config.boolean is False
+        executed["yes"] = True
+
+    os.environ["ENV_VARIABLE_FOR_TEST_INT"] = "2"
```
schrockn, line 961: could use `from dagster._core.test_utils import environ` here

python_modules/dagster/dagster/_core/definitions/op_invocation.py
```diff
@@ -158,7 +159,10 @@ def op_invocation_result(
         from dagster._config.pythonic_config import Config
 
         context = (context or build_op_context()).replace_config(
-            config_input._as_config_dict()  # noqa: SLF001
+            config_dictionary_from_values(
+                config_input._as_config_dict(),  # noqa: SLF001
+                config_input.to_config_schema().as_field(),
+            )
```
schrockn, line 165: similar to feedback on other diff. should this not live in `_as_config_dict`?

schrockn (changes requested):

python_modules/dagster/dagster/_core/definitions/op_invocation.py
benpankow: stacked on #13876 and updated

schrockn (approved):

PR #13897, opened 2023-04-25
[dagster-mlflow] Update to use Pythonic resources

benpankow:
## Summary

Introduces a Pythonic MLflow resource, which users can use to acquire and use a tracking client:

```python
 @op
 def op1(_, mlflow: MLFlowTrackingResource):
    with mlflow.get_tracking_client() as client:
      client.log_params(params)
      run_id_holder["op1_run_id"] = client.active_run().info.run_id
```

## Test Plan

New unit test.


benpankow:
* master
  * **PR #13897** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13897" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13897?utm_source=stack-comment).

python_modules/libraries/dagster-mlflow/dagster_mlflow/resources.py
```diff
@@ -4,13 +4,22 @@
...
+from dagster import (
+    ConfigurableResource,
+    Field,
+    Noneable,
+    Permissive,
+    StringSource,
+    resource,
+)
 from mlflow.entities.run_status import RunStatus
+from pydantic import Field as PyField
```
schrockn, line 22: I think we not alias pydantic.Field and alias dagster.Field if we need both

python_modules/libraries/dagster-mlflow/dagster_mlflow/resources.py
```diff
@@ -218,6 +227,30 @@ def chunks(params: dict, size: int = 100):
...
+    extra_tags: Dict[str, str] = PyField(
+        default={}, description="Any extra key-value tags to log in MlFlow."
+    )
+
+    @contextmanager
+    def get_tracking_client(self) -> Generator[MlFlowTrackingClient, None, None]:
+        context = self.get_resource_context()
+        client = MlFlowTrackingClient(context)
+        yield client
+        client.cleanup_on_error()
```
schrockn, line 251: `cleanup_on_error` should be in a finally block.

Also in terms of naming convention, what would you think about `open_tracking_client` to indicate that it is a context manager?

schrockn (changes requested):
Inline changes

cameronUJ:
Is there any update on the progress of updating this integration @benpankow? I can see this MR was closed, is this because you are taking a different approach?

PR #13910, opened 2023-04-26
Improve performance of `cron_string_iterator` in cases where you're starting on a cron tick boundary.

OwenKephart:
## Summary & Motivation

We've gotten bitten by the fact that creating new `cron_string_iterator`s is surprisingly expensive, such that doing it `O(1000)` times can time out a sensor tick.

This has two main optimizations:

1. In the case that we're on a cron tick boundary (calculated using the relatively cheap `croniter.match()`), we skip this `get_prev` / `get_next` stuff, which is significantly slower. This means that in the case of calculating partition windows, we'll never even hit croniter unless the user has a custom (and devious) TimeWindowPartitionsDefinition.
2. A lightweight shim over the static croniter class, which allows us to cache the result of the `expand` call. This gets called on every iteration of croniter, as well as in the matching codepath. It calls regex stuff, which makes it fairly slow. This is admittedly a more hacky solution than I'd like.

On a local repo, constructed a case where the auto materialization logic would require the calculation of ~8000 time windows (in turn requiring this method to be called ~8000 times).

- pre-changes: ~18s
- without hacky shim: ~10s
- with hacky shim: ~4.5s

Note that these times are including a lot of other stuff that the auto materialization logic is doing, but the vast majority of these runtimes (~90%) are associated specifically with this cron string iteration.

### DIFF
--- a/python_modules/dagster/dagster/_utils/schedules.py
+++ b/python_modules/dagster/dagster/_utils/schedules.py
@@ -76,8 +76,9 @@ def cron_string_iterator(
     if is_numeric[1]:
         expected_hour = int(cron_parts[1][0])
 
-    date_iter = None
     if delta_fn is not None and start_offset == 0 and croniter.match(cron_string, start_timestamp):
+        # In simple cases, where you're already on a cron boundary, the below logic is unnecessary
+        # and slow
         next_date = start_datetime
     else:
         date_iter = croniter(cron_string, start_datetime)
@@ -136,7 +137,7 @@ def cron_string_iterator(
             yield next_date
     else:
         # Otherwise fall back to croniter
-        date_iter = date_iter or croniter(cron_string, next_date)
+        date_iter = croniter(cron_string, next_date)
         while True:
             next_date = to_timezone(
                 pendulum.instance(date_iter.get_next(datetime.datetime)), timezone_str
--- a/python_modules/dagster/dagster/_utils/schedules.py
+++ b/python_modules/dagster/dagster/_utils/schedules.py
@@ -1,12 +1,22 @@
 import datetime
+import functools
 from typing import Iterator, Optional, Sequence, Union
 
 import pendulum
 import pytz
-from croniter import croniter
+from croniter import croniter as _croniter
 
 import dagster._check as check
 from dagster._seven.compat.pendulum import to_timezone
+
+
+class croniter(_croniter):
+    """Lightweight shim to enable caching certain values that may be calculated many times."""
+
+    @classmethod
+    @functools.lru_cache(maxsize=128)
+    def expand(cls, *args, **kwargs):
+        return super().expand(*args, **kwargs)
 
 
 def is_valid_cron_string(cron_string: str) -> bool:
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_utils/schedules.py
```diff
@@ -1,14 +1,24 @@
...
 import pytz
-from croniter import croniter
+from croniter import croniter as _croniter
 
 import dagster._check as check
 from dagster._seven.compat.pendulum import to_timezone
 
 
+class croniter(_croniter):
+    """Lightweight shim to enable caching certain values that may be calculated many times."""
```
gibsondan, line 14: this feels a little magical to me - consider something like CroniterShim or something (not blocking)

python_modules/dagster/dagster/_utils/schedules.py
```diff
@@ -83,6 +76,28 @@ def cron_string_iterator(
     if is_numeric[1]:
         expected_hour = int(cron_parts[1][0])
 
+    if delta_fn is not None and start_offset == 0 and croniter.match(cron_string, start_timestamp):
```
gibsondan, line 79: here's my confusion - croniter.match literally calls get_prev and get_current! https://github.com/kiorky/croniter/blob/master/src/croniter/croniter.py#L813-L820

But that corresponds to your "without hacky shim" case above right?

gibsondan (changes requested):
I am mostly curious about why moving from get_prev() to match() helps when the latter calls the former under the hood? But if we have a plausible explanation for that then I would be thrilled to accept this

python_modules/dagster/dagster/_utils/schedules.py
OwenKephart: good callout -- dug more into the profiling, and there's two things going on.

first, in the original code, we were calling `croniter.match` anyway, so removing extra calls to get_prev/get_next will always result in a speed improvement.

second (perhaps more importantly), the time taken to get the previous tick is correlated to the distance between those ticks. Inside of `match`, they're calculating the previous tick of the schedule, relative to 1 microsecond after the input time (so the previous tick will only be one microsecond away. Inside the other codepath, we're getting the true previous tick, which is, in my example, 15 minutes away. You can see the difference in time spent in this profile:

Screenshot 2023-04-26 at 10 25 57 AM

The left hand side is the time spent on "get_prev" in the old codepath, whereas the right hand side is the time spent on get_prev in the "match" codepath.

If you dig in even further, the fast codepath calls `proc_second` inside of `_calc`, whereas the slow codepath calls `proc_minute`.

python_modules/dagster/dagster/_utils/schedules.py
```diff
@@ -122,6 +137,7 @@ def cron_string_iterator(
             yield next_date
     else:
         # Otherwise fall back to croniter
+        date_iter = croniter(cron_string, next_date)
```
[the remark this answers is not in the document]
OwenKephart: good callout -- originally, this was indeed expensive, but it turns out that the expensive part of it was calling `expand()` inside the init  function for the class, so the caching shim solves this problem

python_modules/dagster/dagster/_utils/schedules.py
```diff
@@ -1,18 +1,28 @@
...
 
 import dagster._check as check
 from dagster._seven.compat.pendulum import to_timezone
 
 
+class CroniterShim(_croniter):
+    """Lightweight shim to enable caching certain values that may be calculated many times."""
+
+    @classmethod
+    @functools.lru_cache(maxsize=128)
```
gibsondan, line 17: 128 appears to be the default - how scientifically was that chosen here / should we be tempted at all to make this None so that it caches everything? The space of all possible cron strings seems pretty low in practice. If we're unlikely to ever hit past 128 in a single session then it probably does not matter a ton

python_modules/dagster/dagster/_utils/schedules.py
gibsondan: ahhh very interesting. thanks for digging into that

gibsondan (approved):
excellent - wonder if the caching thing could make a good/quick upstream contribution to croniter? They seem fairly receptive to PRs, another cloud user reported a bug and it got fixed fairly recently

python_modules/dagster/dagster/_utils/schedules.py
OwenKephart: zero science -- was just a combination of it being the default, seeming much larger than the number of cron strings I'd expect to see in a single session, and unbounded caching felt scary

OwenKephart:
@gibsondan re-requesting review due to some moderate changes. The big one being that I needed to implement `exact_match`. Basically, if you're a few seconds past a cron schedule tick, we were counting that as "still on the tick", because of how croniter does its calculation.

python_modules/dagster/dagster/_utils/schedules.py
```diff
@@ -1,18 +1,38 @@
...
+    def expand(cls, *args, **kwargs):
+        return super().expand(*args, **kwargs)
+
+
+def exact_match(cron_expression: str, dt: datetime.datetime) -> bool:
```
gibsondan, line 22: `_exact_match` since its private?

gibsondan (approved):
lgtm - do we need any of this stuff in reverse_cron_string_iterator? Seems like that one is more out of the critical path?

OwenKephart:
@gibsondan yeah the reverse_cron_string_iterator is not on the critical path

Issue #13474, opened 2023-04-10
Experiment: Change scroll behavior in Asset graph


braunjj:
I'd like to explore a change in the scroll behavior in the asset graph. Today, scrolling zooms in and out of the graph and in order to move around the graph, you need to click and drag to pan.

This is not ideal for really large asset graphs. Moving around the graph requires lots of clicking and dragging. I think we should prototype using the scroll behavior to pan around the graph and then move the zoom behavior to a keyboard shortcut like (Shift + and Shift -)

**Some specific user feedback regaurding this**:
<img width="403" alt="image" src="https://user-images.githubusercontent.com/2798333/230950298-644b469a-ccaa-4472-8a0f-d731dcc51fdf.png">
<img width="420" alt="image" src="https://user-images.githubusercontent.com/2798333/230950212-12b5413d-64cc-43f3-b0ea-8f3c1242b532.png">

PR #13913, opened 2023-04-26
[ui] Add feature flag for panning vs. zooming DAG on scroll

bengotow:
## Summary & Motivation

Fixes https://github.com/dagster-io/dagster/issues/13474

This PR adds a new feature flag: `Pan graph visualizations when scrolling`

When enabled, the scroll wheel and the trackpad pan the asset and op graphs rather than zooming them. To zoom, you hold the Meta key while scrolling.

This mirrors Figma with one small exception -- in Figma with a mouse scroll wheel, holding shift scrolls -X instead of +X. I can't figure out how they do this on scroll wheel scrolling and not on trackpad scrolling, and it may not be worth the squeeze unless we really like this interaction.

I think this feels much more natural and treats the viz more like a "canvas" in other apps!

## How I Tested These Changes

I toggled the feature flag on and off and verified that the interactions change as expected.

### DIFF
--- a/js_modules/dagit/packages/core/src/graph/SVGViewport.tsx
+++ b/js_modules/dagit/packages/core/src/graph/SVGViewport.tsx
@@ -5,6 +5,8 @@ import ReactDOM from 'react-dom';
 import {MemoryRouter} from 'react-router-dom';
 import styled from 'styled-components/macro';
 
+import {FeatureFlag, featureEnabled} from '../app/Flags';
+
 import {IBounds} from './common';
 import {makeSVGPortable} from './makeSVGPortable';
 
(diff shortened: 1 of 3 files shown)


js_modules/dagit/packages/core/src/graph/SVGViewport.tsx
```diff
@@ -115,16 +117,32 @@ const PanAndZoomInteractor: SVGViewportInteractor = {
...
+    // convert wheel event units to a better scroll speed
+    const speed = 0.7;
+
     if (!cursorPosition) {
       return;
     }
-
-    if (event.altKey || event.shiftKey) {
-      viewport.shiftXY(-event.deltaX, -event.deltaY);
+    if (flagDAGPanWithScrollWheel) {
```
salazarm, line 128: Shouldn't this flag only be around the new behavior? Looks like if the flag is off then `onWheel` is a no-op. is that correct?

salazarm (approved):
left a question, would be nice if we could write a test for this `onWheel` behavior (maybe just rendering a dag and firing a wheel event and making sure the right methods are called with  the flag on/off)

js_modules/dagit/packages/core/src/graph/SVGViewport.tsx
bengotow: Ahh the last  `} else {` is for this flag so we should be good on that front!

bengotow:
Ahh let me see if we can write a test -- I think that the viewport currently doesn't render in `jest` because the DOM mocks aren't good enough but let me see why

js_modules/dagit/packages/core/src/graph/SVGViewport.tsx
salazarm: oh derp, I can't count closing/openning braces 😅

bengotow:
Bah, going to do a follow-up where I try to test this, I played around with it a bunch this morning and I can't seem to get this to work in jest.

Issue #13717, opened 2023-04-18
make it obvious when auto-materializing is off, but an asset has an auto-materialize policy


sryza:

bengotow:
We're going to add a tag similar to the schedule + sensor tag shown at the top of asset details / asset group -- green or yellow,  on or off tag icon, text: "Auto-materialize". Tooltip that explains that the daemon is paused. Click to go to daemons page and change setting.

PR #13914, opened 2023-04-26
[ui] Show a tag with pause state when auto-materialization policy is present

bengotow:
## Summary & Motivation

Fixes https://github.com/dagster-io/dagster/issues/13717

Loom video: https://www.loom.com/share/68907579d31342a090780abc7307d48b

This PR adds a small tag in the page header: green if an auto-materializing asset is present and the daemon is un-paused, and yellow if it is paused. If no auto-materializing assets are present, it is hidden. You can click the tag to visit the health page where you can toggle the status of the daemon.

image

## How I Tested These Changes

I tested this by viewing asset jobs, asset groups, and asset details pages with and without a materialization policy present. I updated the existing daemon page tests to reflect the changes.


salazarm (approved):
sweet

PR #13920, opened 2023-04-26
enable prefixing source assets when loading

sryza:
## Motivation

From a user:

> Is there a reason why source assets aren't able to be prefixed when loading from a module?
>
> Context here, is we are trying to load into assets with a client prefix from the same module per client. We can prepend the client prefix to normal assets, but it explicitly isn't working for source assets. Found a comment in the code that seems to indicate this is intentional?

## Summary

This adds a `source_key_prefix` boolean to the core `load_assets_from_*` functions. This makes it consistent with `load_assets_from_dbt_project`.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/load_assets_from_modules.py
+++ b/python_modules/dagster/dagster/_core/definitions/load_assets_from_modules.py
@@ -102,6 +102,7 @@ def load_assets_from_modules(
     *,
     freshness_policy: Optional[FreshnessPolicy] = None,
     auto_materialize_policy: Optional[AutoMaterializePolicy] = None,
+    prefix_sources: bool = False,
 ) -> Sequence[Union[AssetsDefinition, SourceAsset, CacheableAssetsDefinition]]:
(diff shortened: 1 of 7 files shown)


smackesey (approved):
lgtm

sryza:
@OwenKephart - mind taking a quick look at the API? Thoughts on `prefix_sources` vs. `prefix_source_assets` vs. some other parameter entirely?

OwenKephart:
@sryza I'm generally happy with `prefix_sources`, although an alternative implementation would be to just have a `source_key_prefix` parameter (this is what dagster-dbt does).

This would a) make the signature a bit more consistent and b) add some additional flexibility.

The downside is that it requires a bit of extra typing on the user's end, as they have to duplicate the prefix in the common case, rather than just typing `True`.

sryza:
@OwenKephart I think `source_key_prefix` is better, for consistency and flexibility. I posted a new version of the PR with it. Mind taking another look

OwenKephart (approved):
nice

PR #13924, opened 2023-04-26
[2/2] Show dynamic partition requests in "Test Sensor" dialog

salazarm:
## Summary & Motivation

image

## How I Tested These Changes
`dagster dev -f python_modules/dagster-test/dagster_test/toys/partitioned_assets/dynamic_asset_partitions.py -p 3333`

jest


js_modules/dagit/packages/core/src/ticks/DynamicPartitionRequests.tsx
```diff
@@ -0,0 +1,72 @@
...
+  return (
+    <Box flex={{direction: 'column', gap: 12, grow: 1}} margin={{top: 24}}>
+      <Box flex={{direction: 'row', alignItems: 'center', gap: 8}}>
+        <Icon name="partition" />
+        <Subheading>Dynamic Partition Requests</Subheading>
+      </Box>
+      <Table style={{borderRight: `1px solid ${Colors.KeylineGray}`}}>
+        <thead>
+          <tr>
+            <th>Partition</th>
```
clairelin135, line 43: What do you think about changing the order of the columns to first group by partitions def name, and then partition?

I think it might be a bit easier to parse if multiple partitions defs requests are returned in one sensor tick

clairelin135: We could also make the "partition" column a "partitions" column that contains a list of partitions that have been requested, so that each row matches a request object.

I don't feel strongly about this though--up to you.

clairelin135 (commented):
Added a couple small comments, will leave it to others to review the tsx changes

js_modules/dagit/packages/core/src/ticks/DynamicPartitionRequests.tsx
salazarm: @braunjj any thoughts re the above?

js_modules/dagit/packages/core/src/instigation/InstigationTick.tsx
```diff
@@ -223,6 +223,14 @@ export const TICK_TAG_FRAGMENT = gql`
   ${PYTHON_ERROR_FRAGMENT}
 `;
 
+export const DYNAMIC_PARTITION_REQUESTS_FRAGMENT = gql`
```
bengotow, line 226: Technically I feel like this fragment should be colocated with the `DynamicPartitionRequests` component which utilizes the fragment type. Want to move it there maybe?

js_modules/dagit/packages/core/src/ticks/SensorDryRunDialog.tsx
```diff
@@ -368,11 +374,21 @@ export const EVALUATE_SENSOR_MUTATION = gql`
...
       }
       ...PythonErrorFragment
     }
   }
+
+  fragment DynamicPartitionRequestFragment on DynamicPartitionRequest {
+    partitionKeys
+    partitionsDefName
+    type
+  }
```
bengotow, line 390: Not sure why this is redefined here - I think this should be a reference to `${DYNAMIC_PARTITION_REQUESTS_FRAGMENT}`?

bengotow (approved):

PR #13928, opened 2023-04-26
Remove ScheduleTimeBasedPartitionsDefinition class

clairelin135:
This class previously existed to support `@hourly_schedule`/`@daily_schedule`/`@weekly_schedule`/`@monthly_schedule`, but those are gone now. Removing this class since it is unused.


clairelin135:
Vinnie from Mushlabs is adding new time window partitions definitions  (`YearlyPartitionsDefinition` and `QuarterlyPartitionsDefinition`) and had some confusion about implementations in this class.

I noticed that this class isn't being used, so trying to delete it now. I've deleted all of the test cases for this too, let me know if we should be modifying the test cases in some way instead.

rexledesma (approved):
If these tests are used in the `TimeWindowPartitionsDefinition`, then I would have no objection to deleting!

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -143,6 +138,12 @@ def __new__(
                 day_offset=day_offset or 0,
             )
 
+        if not is_valid_cron_schedule(cron_schedule):
```
clairelin135, line 141: before, if an invalid value was provided to partitioned config (i.e. day offset = 8  in a weekly partitioned config), Dagster would throw an error when Dagit loads the partition page or the daemon loads a schedule that contains the partitioned config. 

This bumps this error to instead be thrown at definition time.

clairelin135:
These tests previously did not exist in `TimeWindowPartitionsDefinition`. The implementation is pretty different between the `ScheduleTimeBasedPartitionsDefinition` class and the current `TimeWindowPartitionsDefinition`:
- `ScheduleTimeBasedPartitionsDefinition` removes the last N partitions where N = offset, and offset = 0 retains the current day. Whereas `TimeWindowPartitionsDefinition` adds N partitions where N = offset, and offset = 0 does not include the current day.
- `ScheduleTimeBasedPartitionsDefinition` contains an end time, whereas `TimeWindowPartitionsDefinition` does not.

I've updated the existing tests to match the current implementation of the time window partitions def.

PR #13929, opened 2023-04-26
Add auto materialize run tags to docs

johannkm:


sryza (commented):
I think we should resolve the discussion in https://github.com/dagster-io/dagster/pull/13722 before adding this.

sryza (approved):
nice

PR #13905, opened 2023-04-25
[docs] Update state-in-resources docs to include lifecycle hooks

benpankow:
## Summary

Updates the resources docs to utilize the hooks in #13938, explaining a basic case using `PrivateAttr` & `setup_for_execution` alongside a more complex yield case.

Corresponds to the approach taken in #13894

PR #13938, opened 2023-04-26
[rfc][pythonic resources] Introduce Pythonic resource lifecycle hooks

benpankow:
## Motivation

As part of the initial 1.3 release of Pythonic resources, the decision was made to make `ConfigurableResources` immutable and to encourage users to defer state management to a separate class.

This created a clean separation of concerns, leaving the resource class as a lightweight class holding only config schema. However, this decision was a heavy-handed prescription for how users needed to model their code:

```python

class MyDatabaseClient:
    def __init__(self, conn_str: str):
        self._connection = connection.open(conn_str)

    def query(self, query: str):
        return self._connection.query(query)

class MyDatabaseResource(ConfigurableResource):
    username: str
    password: str
    uri: str
    
    def get_client(self) -> MyDatabaseClient:
        return MyDatabaseClient(conn_str=...)

@asset
def my_asset(db: MyDatabaseResource):
    db.get_client().query(...)
```

Initial feedback from users suggests that we should give more freedom to creating stateful resource classes which both model config and hold some state. Many resources, from database connections to external API clients require initial setup that is verbose to build with the current restrictions.

## Summary

This PR introduces (1) a blessed pattern for specifying private state within a `ConfigurableResource` and (2) a set of Dagster-specific lifecycle hooks which users can use to specify setup, teardown, and context behavior for their resources.

### Specifying private state

The Pythonic resource system is built on top of Pydantic, which already supports a notion of [private model attributes](https://docs.pydantic.dev/usage/models/#private-model-attributes). These attributes can be specified by assigning attributes intended to be private an instance of `PrivateAttr`. As part of this PR, `PrivateAttr` will become an endorsed method of managing state on a resource class:

```python
class MyDatabaseResource(ConfigurableResource):
    username: str
    password: str
    uri: str

    _connection: Connection = PrivateAttr()
```

These private attributes can be initialized before runtime as part of lifecycle hooks (see below) and reassigned at any time during execution:

```python
class MyAPIClientResource(ConfigurableResource):
    api_key: str

    _cache: Dict[str, Any] = PrivateAttr(default_factory=dict)

    def request(self, input: str):
        if input in self._cache:
            return self._cache["input"]
        result = ...
        self._cache[input] = result
        return result
```

### Lifecycle hooks

To make setting up and tearing down connections and state easier, this PR introduces a set of resource lifecycle hooks that can optionally be implemented. 

#### `setup_for_execution` and `teardown_after_execution` hooks

The simplest set of hooks is `setup_for_execution` and `teardown_after_execution`. `setup_for_execution` is invoked at run execution time, after all resource config is populated, env vars resolved, and nested resources set up. `teardown_after_execution` is invoked after a run execution completes to allow for cleanup.

```python
class MyAPIClientResource(ConfigurableResource):
    api_key: str
    _internal_client: MyAPIClient = PrivateAttr()

    def setup_for_execution(self, context):
        self._internal_client = MyAPIClient(self.api_key)

    def get_all_items(self):
        return self._internal_client.items.get()

@asset
def all_items(api_client: MyAPIClientResource):
    return api_client.get_all_items()
```

#### `yield_for_execution`

More advanced use cases may instead want to open contexts instead of separate setup and teardown steps. These resources can override `yield_for_execution`, which by default invokes `setup_for_execution`, yields the resource, and invokes `teardown_after_execution` afterwards.

```python
class MyDatabaseResource(ConfigurableResource):
    username: str
    password: str
    uri: str

    _connection: Connection = PrivateAttr()
   
    @contextlib.contextmanager
    def yield_for_execution(self, context):
        with open_connection(self.uri, self.username, self.password) as conn:
            self._connection = conn
            yield self
```

#### Dagster lifecycle methods v.s. dataclass lifecycle methods

Here we choose explicitly to introduce Dagster-specific lifecycle methods rather than using established methods such as `dataclass`' `__post_init__` or Pydantic [root validators](https://docs.pydantic.dev/usage/validators/#root-validators). This is to delay resource setup and teardown to Dagster execution time, as traditional init lifecycle methods would be called every time a resource class is initialized on code location refresh or run load.

These Dagster lifecycle methods can coexist with lightweight post-init methods such as root validators.

### Example patterns

This change allows users to follow more palatable patterns for common resource use cases. For example, `ConfigurableIOManagers` which wrap a plain `IOManager`:

```python
class InternalIOManager(IOManager):
    ...

class AConfigurableIOManager(ConfigurableIOManager):
    api_key: str

    _internal_io_manager: InternalIOManager = PrivateAttr()

    def setup_for_execution(self, context):
        self._internal_io_manager = InternalIOManager(self.api_key)

        return self._internal_io_manager.load_input(context)

    def handle_output(self, context):
        return self._internal_io_manager.handle_output(context)
``` 

## Test Plan

New suite of unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -6,6 +6,7 @@ from typing import (
     Any,
     Callable,
     Dict,
+    Generator,
     Generic,
     Iterable,
     Mapping,
@@ -903,17 +904,44 @@ class ConfigurableResourceFactory(
         )
         return copy
 
-    def _initialize_and_run(self, context: InitResourceContext) -> TResValue:
+    def _initialize_and_run(self, context: InitResourceContext) -> Generator[TResValue, None, None]:
         updated_resource = (
             self._resolve_and_update_nested_resources(context)  # noqa: SLF001
             .with_resource_context(context)
             ._resolve_and_update_env_vars()
         )
 
-        return updated_resource._create_object_fn(context)  # noqa: SLF001
+        yield from updated_resource.yield_for_execution(context)
 
-    def _create_object_fn(self, context: InitResourceContext) -> TResValue:
-        return self.create_resource(context)
+    def pre_execute(self, context: InitResourceContext) -> None:
+        """Optionally override this method to perform any pre-execution steps
+        needed before the resource is used in execution.
+        """
+        pass
+
+    def post_execute(self, context: InitResourceContext) -> None:
+        """Optionally override this method to perform any post-execution steps
+        needed after the resource is used in execution.
+
+        post_execute will be called even if the resource fails to initialize in the
+        pre_execute step, and if any part of the run fails.
+        """
+        pass
+
+    def yield_for_execution(self, context: InitResourceContext) -> Generator[TResValue, None, None]:
+        """Optionally override this method to perform any lifecycle steps
+        before or after the resource is used in execution. By default, calls
+        pre_execute before yielding, and post_execute after yielding.
+
+        Note that if you override this method and want pre_execute or
+        post_execute to be called, you must invoke them yourself.
+        """
+        try:
+            self.pre_execute(context)
+            yield self.create_resource(context)
+        finally:
+            # what to do if post_execute throws
+            self.post_execute(context)
 
     def get_resource_context(self) -> InitResourceContext:
         """Returns the context that this resource was initialized with."""
@@ -940,7 +968,7 @@ class ConfigurableResourceFactory(
                 return MyResource.from_resource_context(context)
 
         """
-        return cls(**context.resource_config or {})._create_object_fn(context)  # noqa: SLF001
+        return cls(**context.resource_config or {}).create_resource(context)
 
 
 class ConfigurableResource(ConfigurableResourceFactory[TResValue]):
@@ -1184,13 +1212,8 @@ class ConfigurableIOManagerFactory(ConfigurableResourceFactory[TIOManagerValue])
         """Implement as one would implement a @io_manager decorator function."""
         raise NotImplementedError()
 
-    def _create_object_fn(self, context: InitResourceContext) -> TIOManagerValue:
-        return self.create_io_manager(context)
-
     def create_resource(self, context: InitResourceContext) -> TIOManagerValue:
-        # I/O manager factories execute a different code path that does not
-        # call create_resource
-        raise NotImplementedError()
+        return self.create_io_manager(context)
 
     @classmethod
     def configure_at_launch(cls: "Type[Self]", **kwargs) -> "PartialIOManager[Self]":
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -798,8 +798,12 @@ class ConfigurableResourceFactory(
 
     @cached_method
     def get_resource_definition(self) -> ConfigurableResourceFactoryResourceDefinition:
+        is_cm_resource = (
+            self.__class__.yield_for_execution != ConfigurableResourceFactory.yield_for_execution
+            or self.__class__.post_execute != ConfigurableResourceFactory.post_execute
+        )
         return ConfigurableResourceFactoryResourceDefinition(
-            resource_fn=self._initialize_and_run,
+            resource_fn=self._initialize_and_run_cm if is_cm_resource else self._initialize_and_run,
             config_schema=self._config_schema,
             description=self.__doc__,
             resolve_resource_keys=self._resolve_required_resource_keys,
@@ -887,8 +891,8 @@ class ConfigurableResourceFactory(
         with contextlib.ExitStack() as stack:
             resources_to_update, _ = separate_resource_params(self.__dict__)
             resources_to_update = {
-                attr_name: stack.enter_context(
-                    _call_resource_fn_with_default(coerce_to_resource(resource), context)
+                attr_name: _call_resource_fn_with_default(
+                    stack, coerce_to_resource(resource), context
                 )
                 for attr_name, resource in resources_to_update.items()
                 if attr_name not in partial_resources_to_update
@@ -909,8 +913,19 @@ class ConfigurableResourceFactory(
         )
         return copy
 
+    def _initialize_and_run(self, context: InitResourceContext) -> TResValue:
+        with self._resolve_and_update_nested_resources(context) as has_nested_resource:
+            updated_resource = has_nested_resource.with_resource_context(  # noqa: SLF001
+                context
+            )._resolve_and_update_env_vars()
+
+            updated_resource.pre_execute(context)
+            return updated_resource.create_resource(context)
+
     @contextlib.contextmanager
-    def _initialize_and_run(self, context: InitResourceContext) -> Generator[TResValue, None, None]:
+    def _initialize_and_run_cm(
+        self, context: InitResourceContext
+    ) -> Generator[TResValue, None, None]:
         with self._resolve_and_update_nested_resources(context) as has_nested_resource:
             updated_resource = has_nested_resource.with_resource_context(  # noqa: SLF001
                 context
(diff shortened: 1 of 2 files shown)


benpankow:
* master
  * **PR #13938** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13938" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13938?utm_source=stack-comment).

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -812,16 +825,56 @@ def with_resource_context(
...
+    def yield_for_execution(self, context: InitResourceContext) -> Generator[TResValue, None, None]:
+        """Optionally override this method to perform any lifecycle steps
+        before or after the resource is used in execution. By default, calls
+        pre_execute before yielding, and post_execute after yielding.
+
+        Note that if you override this method and want pre_execute or
+        post_execute to be called, you must invoke them yourself.
+        """
+        try:
+            self.pre_execute(context)
```
schrockn, line 873: Need to talk through the decision about whether `pre_execute` is in the `try` or not.

If this is the right decision, then we should change the name of the methods as they would now be misleading. Namely we would have to write a sentence like "if pre_execute throws, then post_execute will be called, even though no execution has occured".

We also need to think ahead in case we develop a feature to execute code for a resource on run start, rather than for step start.

Most explicit and accuarate:

`pre_execute` --> `setup_for_step_execution`
`post_execute` --> `teardown_after_step_execution`

In this naming, we should say: "If setup_for_step_execution fails, then step execution is failed. As a result `teardown_after_step_execution` will be called."

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -812,16 +825,56 @@ def with_resource_context(
...
+    def _initialize_and_run_cm(
+        self, context: InitResourceContext
+    ) -> Generator[TResValue, None, None]:
+        with self._resolve_and_update_nested_resources(context) as has_nested_resource:
+            updated_resource = has_nested_resource.with_resource_context(  # noqa: SLF001
+                context
+            )._resolve_and_update_env_vars()
+
+            with updated_resource.yield_for_execution(context) as value:
+                yield value
```
schrockn, line 846: Having these dual code paths is confusing. Seems like `_initialize_and_run_cm` should be the canonical implementation here

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -702,10 +704,17 @@ def _nested_resources(self):
     def _resolved_config_dict(self):
         return self._state__internal__.resolved_config_dict
 
+    def _get_initialize_and_run_fn(self) -> Callable:
+        is_cm_resource = (
+            self.__class__.yield_for_execution != ConfigurableResourceFactory.yield_for_execution
+            or self.__class__.post_execute != ConfigurableResourceFactory.post_execute
+        )
+        return self._initialize_and_run_cm if is_cm_resource else self._initialize_and_run
+
```
schrockn, line 713: This shouldn't be necessary IMO

schrockn (changes requested):
Cool this is good start. I think we can simplify this implementation as noted line, and we need to iterate on naming, but this is the right direction.

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
schrockn: I don't think `setup_for_execution_process` as the word "process" itself is ambiguous between the "operating system process" and "general process".

`setup_for_execution` and `teardown_after_execution` seem good to me

sryza: On the face of it, those names sound good to me. I think it's worth considering how they'd sound if we eventually decide to also add run-level setup / teardown methods (e.g. for setting up / tearing down a cluster per run).

schrockn: Yeah agree sandy. I commented similar above:

> We also need to think ahead in case we develop a feature to execute code for a resource on run start, rather than for step start.

I'm also pretty confident we won't want to make to them methods, lest the model for state in these classes gets even _more_ complicated, but instead some other lifecycle mechanism that accepts the resource as an argument much more akin to an `op`. Otherwise the user will have to thing of the object in a bunch in a huge number of states. (e.g. I am in process where the  `setup_for_execution` method has been called, but not the `on_run_start` call, since this is a process that is in the middle of run). 

We can dig into that in another PR/discussion.

sryza: > I'm also pretty confident we won't want to make to them methods, lest the model for state in these classes gets even more complicated

I agree that it shouldn't be able to modify the state of the object. Maybe a static method?

danielgafni:
I think this is great. I need exactly something like this. Great design!

Just two questions about the naming: 
 - why 'teardown_for_execution' and not 'teardown_after_execution'? As I understand, executing a job for Dagster includes using the resource, but it might be confusing for new users. 
 - why are the names so verbose? Why not something more simple like 'setup' and 'teardown' (for example, PyTorch-lightning uses these names for their datamodule hooks).

schrockn:
Hey @danielgafni thanks for the feedback.

Re the `setup_for_execution` versus `setup`. One of my biggest concern is confusion around when these lifecycle methods are called. 

Specifically, these methods are _not_ called when the resource objects are passed to the `Definitions` in process. This is so that potentially expensive/mutative setup and teardown actions are not called on every code location spinup (whether it be the grpc server or an unrelated step.) They are only called if this resource is actually involved in the execution of a run (i.e. passed to an op) The names are meant to communicate that these are calls only in preparation for execution in an op, rather than after ever time the object is constructed.

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -904,16 +924,56 @@ def with_resource_context(
...
+        Note that if you override this method and want setup_for_execution or
+        teardown_after_execution to be called, you must invoke them yourself.
+        """
+        try:
+            self.setup_for_execution(context)
+            yield self.create_resource(context)
+        finally:
+            # what to do if teardown_after_execution throws
+            self.teardown_after_execution(context)
```
schrockn, line 976: So I'm going to go ahead and argue the case that `setup_for_execution` should be invoked outside of the `try`. That would mean that `teardown_after_execution` would _not_ be called if `setup_for_execution` throws. 

In effect this would say that if `setup_for_execution` fails, the resource is in an invalid, untrusted state and will not be invoked by the framework.

On a practical basis, this prevents one from having to write a bunch of boilerplate in teardown methods to prevent crashes if the setup fails.

Imagine the following class:
```
class ResourceWithComplexSetup(ConfigurableResource):
    # ...
    def setup_for_execution(self):
        self._thing_one = get_thing_one()
        self._thing_two = get_thing_two()
```

Now imagine that and `get_thing_one`, and `get_thing_two` can both throw.

Now the user has to deal with many states of the object. Namely that both `_thing_one` and `_thing_two` are None (`get_thing_one` threw), `_thing_two` is None ('get_thing_two' threw), or neither is None

In order to do thing safely at minimum you have to ensure that each object is not None to account for the three possible states of the object.
```
class ResourceWithComplexSetup(ConfigurableResource):
    # ...
    def teardown_after_execution(self):
        if self._thing_one:
            self._thing_one.cleanup()
        if self._thing_two:
           self._thing_two.cleanup()
```
This obviously can be much more complex. There are more subtle cases in this when the code in the setup methods are modifying some global state in some way. You have to think about the combination of all the possible codepaths and think about all the possible states the object might be in to handle teardown safely 

It's worth noting that we have an escape hatch in _either_ case, in that we can instruct the user to override `yield_for_execution` if the default we choose does not suit them.  They can put setup code inside or outside of the `try` block they write themselves. (The purist in me wants to only have this single plug-ability point, to simplify the relationship between framework code and user code [avoiding the conversation about the semantics of teardown entirely], but it would make the simple cases a bit annoying).

schrockn (changes requested):
Requesting changes to surface last discussion in your queue. Let's resolve quickly and then get this merged.

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
benpankow: This seems reasonable to me. Updated.

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -904,16 +946,55 @@ def with_resource_context(
...
+        """Optionally override this method to perform any pre-execution steps
+        needed before the resource is used in execution.
+        """
+        pass
+
+    def teardown_after_execution(self, context: InitResourceContext) -> None:
+        """Optionally override this method to perform any post-execution steps
+        needed after the resource is used in execution.
+
+        teardown_after_execution will be called even if any part of the run fails.
```
schrockn, line 979: Add comment on how it will not be called if `setup_for_execution` fails

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -904,16 +946,55 @@ def with_resource_context(
...
+        setup_for_execution before yielding, and teardown_after_execution after yielding.
+
+        Note that if you override this method and want setup_for_execution or
+        teardown_after_execution to be called, you must invoke them yourself.
+        """
+        self.setup_for_execution(context)
+        try:
+            yield self.create_resource(context)
+        finally:
+            # what to do if teardown_after_execution throws
```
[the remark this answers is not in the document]
schrockn: Instead add test case that does this throw in the middle of execute_in_process, confirm behavior

benpankow:
Added more test cases based on feedback 👍

python_modules/dagster/dagster_tests/core_tests/resource_tests/pythonic_resources/test_lifecycle_methods.py
```diff
@@ -0,0 +1,669 @@
...
+    def hello_world_job() -> None:
+        my_hello_world_op()
+
+    defs = Definitions(
+        jobs=[hello_world_job], resources={"first": MyResource(), "second": MySecondResource()}
+    )
+
+    assert defs.get_job_def("hello_world_job").execute_in_process().success
+
+    # order is not deterministic
```
schrockn, line 412: oh how interesting. I suppose we iterate through an unordered dictionary somewhere which causes this?

schrockn (approved):
Alrighty let's do this.

We also need to update content to recommend this. Is there are PR out for that or is that a TODO?

benpankow:
I have some WIP docs here: #13905

PR #13944, opened 2023-04-27
Use new lifecycle methods

schrockn:

### DIFF
--- a/python_modules/libraries/dagster-datadog/dagster_datadog/resources.py
+++ b/python_modules/libraries/dagster-datadog/dagster_datadog/resources.py
@@ -5,7 +5,7 @@ from datadog import DogStatsd, initialize, statsd
 from pydantic import Field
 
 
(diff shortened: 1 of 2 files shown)


schrockn:
Stacked on https://github.com/dagster-io/dagster/pull/13938 to kicktest the new API

python_modules/libraries/dagster-datadog/dagster_datadog/resources.py
```diff
@@ -136,4 +141,6 @@ def dd_job():
             )
 
     """
-    return DatadogResource.from_resource_context(context).get_client()
+    resource = DatadogResource.from_resource_context(context)
+    resource.pre_execute(context)
+    return resource
```
schrockn, line 146: @benpankow This PR feels pretty good except for this. For the backwards compat case the user has to handle lifecycle hooks, which doesn't feel great.

PR #13957, opened 2023-04-27
[dagster-dbt] relax pin

OwenKephart:


rexledesma (approved):
Need to relax some dbt pins in `project-fully-featured` and `assets-smoke-test` as well

PR #13706, opened 2023-04-18
[docs][pythonic resources] Update branch deployment guide

benpankow:
## Summary

Updates the branch deployment guide to use the new Snowflake IO manager + Pythonic config system. Does not update the Snowflake resource since that's still pending.

PR #13961, opened 2023-04-27
[dagster-snowflake] Fix issues with newlines in private key authentication

jamiedemaria:
## Summary & Motivation
User reported that if you use private key authentication for snowflake and pass the private key as an environment variable, the environment variable can't handle the newlines within the key correctly, causing the authentication to fail. 

This is similar to what happens with GCP, so we can solve the problem the same way, by base64 encoding the key before passing it as configuration, and decoding it within the resource/io manager 

One thing i'm a little concerned about is that this "private_key" config option has been available for a while. The fact that no one has complained about passing a key via env var not working either means that there is another workaround, or that no one is using pk authentication with an env var. If it's the later, then this change is fine, but if it's the former then this change is breaking. Maybe instead we add a new config value called `b64_private_key` or something along those lines?

To Do:
- [x] fix unit tests that mock out a private key

## How I Tested These Changes
tested locally

### DIFF
--- a/python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
+++ b/python_modules/libraries/dagster-snowflake/dagster_snowflake/resources.py
@@ -1,3 +1,4 @@
+import base64
 import sys
 import warnings
 from contextlib import closing, contextmanager
@@ -120,7 +121,9 @@ class SnowflakeResource(ConfigurableResource, IAttachDifferentObjectToOpContext)
         description=(
             "Raw private key password to use. See"
             " https://docs.snowflake.com/en/user-guide/key-pair-auth.html for details. Required for"
-            " both private_key and private_key_path."
+            " both private_key and private_key_path. To avoid issues with newlines in the keys, you"
+            " must base64 encode the key. You can retrieve the base64 encoded key with this shell"
+            " command: cat rsa_key.pub | base64"
         ),
     )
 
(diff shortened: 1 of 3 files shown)


jamiedemaria:
* master
  * **PR #13232** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13232" target="_blank">Graphite</a> 
    * **PR #13347** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13347" target="_blank">Graphite</a> 
      * **PR #13961** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13961" target="_blank">Graphite</a>  👈
      * **PR #13706** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13706" target="_blank">Graphite</a> 
    * **PR #13235** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13235" target="_blank">Graphite</a> 
    * **PR #13233** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13233" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13961?utm_source=stack-comment).

benpankow:
> One thing i'm a little concerned about is that this "private_key" config option has been available for a while. The fact that no one has complained about passing a key via env var not working either means that there is another workaround, or that no one is using pk authentication with an env var. If it's the later, then this change is fine, but if it's the former then this change is breaking. Maybe instead we add a new config value called b64_private_key or something along those lines?

Do all keys have newlines? If so, we might be able to detect the presence of a newline (if the user is using a workaround) and either not decode it or loudly error to change the env var to be base64 encoded. That way we either don't break existing users (if any) or give them an actionable fix.

jamiedemaria:
@benpankow  we could wrap the function that fails with newlines in a try block and if it fails try the b64 decode and if that fails then raise a more helpful error message. that seems reasonable to me. i'll try implementing it

jamiedemaria:
updated this so that it is backwards compatible. Another review pass would be great!

yuhan (approved):

Issue #9076, opened 2022-07-28
Make Output metadata available on the OutputContext object


OwenKephart:
### What's the use case?

When you yield an Output from the body of an op, you have the opportunity to add metadata to that event in one of two ways:

1. `yield Output(..., metadata={"some": "metadata"})`
2. `context.add_output_metadata(...)`

Inside the IOManager, you might want to use this runtime metadata to modulate how you process that output. If metadata is supplied the first way, this is essentially impossible, as you don't have access to the Output object (perhaps you could do some gross thing to get it, but I didn't feel like digging into that). If metadata is supplied in the second way, you can do `context.step_context.get_output_metadata(context.name)` to retrieve it, which works ok, but is a little confusing as nowhere else in dagster does it matter which way the output metadata was supplied.

### Ideas of implementation

When building the OutputContext, we could supply a new field "runtime_metadata" (name tbd)

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

### Requests
- https://dagster.slack.com/archives/C01U5LFUZJS/p1659619871992709
- Sean Lindo - https://dagster.slack.com/archives/C01U954MEER/p1690141594194749
  - I’m returning an asset like so: return Output(df, metadata={“key”:”value}). How do I access this metadata in my IOManager? context.metadata.get(“key”) returns None

sryza:
Related to https://github.com/dagster-io/dagster/issues/8521

dmosesson:
Did not even realize the second option here. How does that work in the context of multiple Outputs?

FlightOfStairs:
We were impacted by this. Our use-case is that we want to report all numerical metadata for each asset to DataDog.

```
        numerical_metadata: dict[str, IntMetadataValue | FloatMetadataValue] = {
            k: v
            for k, v in context.get_logged_metadata().items()
            if isinstance(v, IntMetadataValue) or isinstance(v, FloatMetadataValue)
        }

        ...

        for metadata_key, metadata_value in numerical_metadata.items():
            statsd.distribution(f"asset.{metadata_key}", metadata_value.value, tags=[f"asset:{asset_name}"])
```

[...]

Issue #8521, opened 2022-06-21
Make runtime output metadata available to downstream ops


sryza:
Requests for this:
* https://github.com/dagster-io/dagster/discussions/6913
* https://dagster.slack.com/archives/C01U954MEER/p1655806363783739
* https://dagster.slack.com/archives/C01U954MEER/p1655606789374799

dpeng817:
This doesn't work if you have partitioned assets - no way to specify which partition key you're searching for.

dpeng817:
If you want to figure out the latest materialization event for a particular partition key:
```python
@asset(non_argument_deps={AssetKey("some_upstream_asset")})
def downstream_asset(context):
    event_log_record = context.instance.get_event_records(EventRecordsFilter(asset_key=AssetKey("some_upstream_asset"), asset_partitions=["some_partition"]), ascending=False)
    event_log_entry = event_log_record.event_log_entry
    metadata = event_log_entry.dagster_event.event_specific_data.materialization.metadata
    foo_value = metadata["foo"].value
    assert foo_entry.text == "bar"
```

danielgafni:
For documentation purposes here is a snipped for getting the mapped upstream asset partitions (if using partition mappings): 

```python
partitions_def = context.asset_partitions_def
assert isinstance(partitions_def, TimeWindowPartitionsDefinition)
partitions = partitions_def.get_partition_keys_in_range(context.asset_partition_key_range)
```

danielgafni:
@sryza maybe @dpeng817 's snippet deserves to be a dagster utility function?

sryza:
@danielgafni - I agree that we should make this more straightforward.

A potential path:
- Accept a `partition_key` argument on `get_latest_materialization_event`.
- Include a property on `EventLogRecord` that reaches down and returns the materialization.

So then @dpeng817's example would turn into:

```python
@asset(non_argument_deps={AssetKey("some_upstream_asset")})
def downstream_asset(context):
    materialization = context.instance.get_latest_materialization_event(AssetKey("some_upstream_asset"), partition_key="xyz)).materialization
    foo_value = materialization.metadata["foo"].value
    assert foo_entry.text == "bar"
```

danielgafni:
Probably (1) looks nicer

sowusu-ba:
Any workaround for getting metadata passed via `context.add_input_metadata`?

sryza:
[...]

PR #13968, opened 2023-04-27
[proof of concept] Fix output metadata

dpeng817:
Related issues: 
https://github.com/dagster-io/dagster/issues/8521
https://github.com/dagster-io/dagster/discussions/6913
https://github.com/dagster-io/dagster/issues/9076

Shows what code changes could look like that enable retrieval of output metadata from `Output(...)` objects. A few things to consider from a design perspective:
- putting this on the metadata object requires overriding the metadata provided on the `Out()` object, which could be undesirable.
- requires a read every time an input is loaded, probably prohibitive from a perf standpoint.
- doesn't handle memoization case / how do we retrieve custom metadata for memoization? I guess we need to read event log in that case as well.

The first two points could be solved by some sort of gated, lazy access of the event log a la `OutputContext.get_runtime_output_metadata`, but the memoization case is challenging. At least for assets, maybe we could just retrieve the last materialization event and use the metadata provided there?

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/events.py
+++ b/python_modules/dagster/dagster/_core/definitions/events.py
@@ -5,6 +5,7 @@ from typing import (
     AbstractSet,
     Any,
     Callable,
+    Dict,
     Generic,
     List,
     Mapping,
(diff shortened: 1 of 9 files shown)


dpeng817:
* master
  * **PR #13968** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13968" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13968?utm_source=stack-comment).

PR #13956, opened 2023-04-27
[fix] Fix sensor resource binding error when late-binding to sensor, job

benpankow:
## Summary

Fixes a bug where sensors attached to jobs that rely on late-bound resources do not have the required resource definition:

https://dagster.slack.com/archives/C01U954MEER/p1682588150393499

This was a casualty of the two features (late-binding resources to jobs and resources in sensors) being developed in parallel - the codepath to reconstruct a sensor with an updated job fails to take into account the resource requirements.

## Test Plan

Test case of job w/ late-bound resource & sensor w/ late-bound resource.

PR #13970, opened 2023-04-27
Introduce dagster_internal_init pattern w/ interface

benpankow:
## Summary

Implements a pattern of a separate, internal-facing `dagster_internal_init` method for public classes. This acts as a variant of `__init__` but which which requires all inputs (e.g. no defaults). The goal of this change is to prevent bugs like #13956 which are introduced by new, optional inputs being added to the constructor. Patterned after #13960.

This change requires us to explicitly specify all inputs/defaults in internal construction cases, while still allowing users to use the more flexible constructor w/ defaults.

## Test Plan

Added a unit test which finds all subclasses of the marker interface `IHasInternalInit` and validates that the `dagster_internal_init` method has the right properties (correct return value, all arguments same as constructor, no defaults etc)

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/assets.py
@@ -239,7 +239,7 @@ class AssetsDefinition(ResourceAddable, IHasInternalInit):
         )
 
     @staticmethod
-    def internal_init(
+    def dagster_internal_init(
         *,
         keys_by_input_name: Mapping[str, AssetKey],
         keys_by_output_name: Mapping[str, AssetKey],
@@ -562,7 +562,7 @@ class AssetsDefinition(ResourceAddable, IHasInternalInit):
         else:
             group_names_by_key = None
 
-        return AssetsDefinition.internal_init(
+        return AssetsDefinition.dagster_internal_init(
             keys_by_input_name=keys_by_input_name,
             keys_by_output_name=keys_by_output_name_with_prefix,
             node_def=node_def,
@@ -1080,7 +1080,7 @@ class AssetsDefinition(ResourceAddable, IHasInternalInit):
                 for out_asset_key in subsetted_keys_by_output_name.values()
             }
 
-            return AssetsDefinition.internal_init(
+            return AssetsDefinition.dagster_internal_init(
                 keys_by_input_name=subsetted_keys_by_input_name,
                 keys_by_output_name=subsetted_keys_by_output_name,
                 node_def=subsetted_node,
@@ -1228,7 +1228,7 @@ class AssetsDefinition(ResourceAddable, IHasInternalInit):
             if key in relevant_keys
         }
 
-        return AssetsDefinition.internal_init(
+        return AssetsDefinition.dagster_internal_init(
             keys_by_input_name=self._keys_by_input_name,
             keys_by_output_name=self._keys_by_output_name,
             node_def=self.node_def,
--- a/python_modules/dagster/dagster/_core/definitions/assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/assets.py
@@ -609,7 +609,7 @@ class AssetsDefinition(ResourceAddable, IHasInternalInit):
             if descriptions_by_output_name
             else None,
             can_subset=can_subset,
-            selected_asset_keys=None,
+            selected_asset_keys=None,  # node has no subselection info
         )
 
     @public
--- a/python_modules/dagster/dagster/_core/definitions/assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/assets.py
@@ -29,6 +29,7 @@ from dagster._core.definitions.time_window_partitions import TimeWindowPartition
 from dagster._core.errors import DagsterInvalidDefinitionError, DagsterInvalidInvocationError
 from dagster._core.selector.subset_selector import SelectionTree
 from dagster._core.types.dagster_type import Nothing
+from dagster._utils import IHasInternalInit
 from dagster._utils.backcompat import (
     ExperimentalWarning,
     deprecation_warning,
@@ -62,7 +63,7 @@ if TYPE_CHECKING:
     from .graph_definition import GraphDefinition
 
 
-class AssetsDefinition(ResourceAddable):
+class AssetsDefinition(ResourceAddable, IHasInternalInit):
     """Defines a set of assets that are produced by the same op or graph.
 
     AssetsDefinitions are typically not instantiated directly, but rather produced using the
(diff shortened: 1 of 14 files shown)


benpankow:
* master
  * **PR #13970** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13970" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/13970?utm_source=stack-comment).

python_modules/dagster/dagster_tests/general_tests/test_internal_init_implementations.py
```diff
@@ -0,0 +1,33 @@
+import inspect
+from typing import Type
+
+import pytest
+from dagster._utils import IHasInternalInit
+
+
+@pytest.mark.parametrize("cls", IHasInternalInit.__subclasses__())
```
schrockn, line 8: Use of `__subclasses__` highly suspect. I didn't know that this even existed, but almost certainly such a construct is highly dependent on import order. A quick test bears this out:

```
class ISomeInterface:
    pass

print(ISomeInterface.__subclasses__())

class Foo(ISomeInterface):
    pass

print(ISomeInterface.__subclasses__())
```

```
(dagster-x86-3.10.3) elementl-13in-m1-mbppro-2021:dagster schrockn$ python subclasses.py
[]
[<class '__main__.Foo'>]
```

The right approach here is to either (a) explicitly whitelist definition classes or (b) import from top-level dagster and check all imports to see if they subclass from IHasInternalInit.

schrockn (dismissed):
So looks like this caught some real bugs (description and metadata shuffling in the assets code paths)

schrockn:
Also @sryza should be up to speed on this

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -369,6 +369,11 @@ def __call__(self, fn: Callable) -> AssetsDefinition:
             auto_materialize_policies_by_key={out_asset_key: self.auto_materialize_policy}
             if self.auto_materialize_policy
             else None,
+            asset_deps=None,
+            selected_asset_keys=None,
+            can_subset=False,
+            metadata_by_key=None,
+            descriptions_by_key=None,
```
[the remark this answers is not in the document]
sryza: The story here is:
- For a long time, these fields did not exist on `AssetsDefinition`.  We instead used the `metadata` and `description` fields on `Out` for this metadata.
- When we added graph-backed assets, this was no longer sufficient, so we added these fields to `AssetsDefinitions`, but we couldn't take them off `Out`.
- So currently our code paths that read them look in both places.

I filed an issue to track adding this info to the `AssetsDefinition` as well: https://github.com/dagster-io/dagster/issues/13988.

python_modules/dagster/dagster/_core/definitions/assets.py
```diff
@@ -241,6 +242,41 @@ def __init__(
             partition_mappings=self._partition_mappings,
         )
 
+    @staticmethod
+    def internal_init(
```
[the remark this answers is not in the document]
schrockn: Internal means "any point where you construct a definition inside production codepaths in the core dagster library". This isn't just an AssetsDefinitions problem. E.g. `copy_for_configured` codepaths, as well as all construction points that implement corresponding decorators.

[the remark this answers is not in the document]
sryza: Got it. `dagster_internal_init` or similar might make this more clear?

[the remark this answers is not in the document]
schrockn: Good idea cc: @benpankow

python_modules/dagster/dagster_tests/general_tests/test_internal_init_implementations.py
benpankow: I'll go ahead an add a top-level `import dagster as dagster` import. I think that should ensure that all user-exposed/top-level classes are loaded, e.g. equivalent to (b)?

schrockn: I feel pretty strongly that we shouldn't use something as fragile as `__subclasses__`. What if some lazy initialization is introduced or something?

Let's directly iterate over the top-level symbols in dagster and test for `isinstance`

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
```diff
@@ -512,7 +512,13 @@ def with_updated_job(self, new_job: ExecutableDefinition) -> "ScheduleDefinition
             description=self.description,
             job=new_job,
             default_status=self.default_status,
+            run_config=None,
+            run_config_fn=self._run_config_fn,
+            tags=self._tags,
+            tags_fn=self._tags_fn,
+            environment_vars=self._environment_vars,
             required_resource_keys=self.required_resource_keys,
+            should_execute=None,
```
schrockn, line 521: please add comments on the `None` arguments so that we understand fully why this is not a bug

python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -893,6 +941,8 @@ def _copy(self, **kwargs: Any) -> "JobDefinition":
             _subset_selection_data=self._subset_selection_data,
             asset_layer=self.asset_layer,
             input_values=self.input_values,
+            partitions_def=self.partitions_def,
+            _was_explicitly_provided_resources=None,
         )
         resolved_kwargs = {**base_kwargs, **kwargs}  # base kwargs overwritten for conflicts
         job_def = JobDefinition(**resolved_kwargs)
```
schrockn, line 948: call dagster_internal_init here

python_modules/dagster/dagster/_core/definitions/decorators/schedule_decorator.py
```diff
@@ -182,6 +182,11 @@ def _wrapped_fn(context: ScheduleEvaluationContext) -> RunRequestIterator:
             job=job,
             default_status=default_status,
             required_resource_keys=required_resource_keys,
+            run_config=None,
+            run_config_fn=None,
+            tags=None,
+            tags_fn=None,
+            should_execute=None,
```
schrockn, line 189: comments here about why all there are null

python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -608,6 +651,11 @@ def execute_in_process(
             version_strategy=self.version_strategy,
             asset_layer=self.asset_layer,
             input_values=input_values,
+            description=self.description,
+            partitions_def=None,
+            metadata=None,
+            _subset_selection_data=None,
+            _was_explicitly_provided_resources=None,
```
schrockn, line 658: comments about Nones

schrockn (dismissed):

schrockn (changes requested):
One last thing. Can you also invoke this in all the `def with_attributes` methods. e.g. assets.py:846. __init__ is dynamically invoked and it should use dagster_internal_init

schrockn (approved):

PR #13983, opened 2023-04-28
[Runs Filtering] Add auto materialize as a "Created by" filter option

salazarm:
## Summary & Motivation

Screen Shot 2023-04-28 at 9 59 16 AM


js_modules/dagit/packages/core/src/runs/RunsFilterInputNew.tsx
```diff
@@ -157,12 +162,7 @@ interface RunsFilterInputProps {
 }
 
 // Exclude these tags from the "tag" filter because theyre already being fetched by other filters.
-const tagsToExclude = [
-  DagsterTag.User,
-  DagsterTag.ScheduleName,
-  DagsterTag.SensorName,
-  DagsterTag.Backfill,
-];
+const tagsToExclude = [...CREATED_BY_TAGS, DagsterTag.Backfill];
```
bengotow, line 165: Ahh nice this is good cleanup 👍  I wonder if these should actually be part of the file with DagsterTag? Might be good to try grepping for `DagsterTag.SensorName` and see if there are other places we're excluding this same set at the component level

bengotow (approved):

js_modules/dagit/packages/core/src/runs/RunsFilterInputNew.tsx
salazarm: That's a good point. I think as we expand filtering to other screens we'll probably end up doing that.

PR #13994, opened 2023-04-28
Remove "solids" alias for "ops" in run config

smackesey:
## Summary & Motivation

Remove the "solids" alias for "ops" in run config. Update the execution plan snapshot backcompat test to use a snapshot from 1.0 using "ops" instead of "solids" (w/out this change the test cannot pass).

## How I Tested These Changes

Existing test suite. Note that many snapshots had to be updated since this alias was part of the default run config schema.


alangenfeld:
The relevant test comes from https://dagster.phacility.com/D7129

I think it makes sense to rerecord it with `1.0.0` and swap out the old `0.11.1` snapshot with that.

smackesey:
>I think it makes sense to rerecord it with 1.0.0 and swap out the old 0.11.1 snapshot with that.

Done

alangenfeld:
I have some hesitancy to land this due to changing all job snapshots for users who update to a version including this change.

The most material consequence is cloud users having to re-upload all their job snapshots for their workspace. I am not sure this is material enough to be that meaningful yet but I do think its something that we probably need to be more mindful of.

alangenfeld (approved):
i think its likely fine

PR #13997, opened 2023-04-28
[dynamic partitions requests in tick 2/3] Serialize dynamic partitions requests in tick

clairelin135:


python_modules/dagster/dagster/_core/scheduler/instigation.py
```diff
@@ -360,6 +428,11 @@ class TickData(
         origin_run_ids (List[str]): The runs originated from the schedule/sensor.
         failure_count (int): The number of times this tick has failed. If the status is not
             FAILED, this is the number of previous failures before it reached the current state.
+        dynamic_partitions_request_results (Sequence[Union[AddDynamicPartitionsRequestResult,
+            DeleteDynamicPartitionsRequestResult]]): The results of the dynamic partitions requests
+            evaluated within the tick.
+
```
sryza, line 434: Nitpick: extra whitespace

sryza (approved):

python_modules/dagster/dagster/_core/scheduler/instigation.py
```diff
@@ -241,6 +291,18 @@ def with_origin_run(self, origin_run_id: str) -> "InstigatorTick":
     def with_log_key(self, log_key: Sequence[str]) -> "InstigatorTick":
         return self._replace(tick_data=self.tick_data.with_log_key(log_key))
 
+    def with_dynamic_partitions_request_result(
+        self,
+        dynamic_partitions_request_result: Union[
+            AddDynamicPartitionsRequestResult, DeleteDynamicPartitionsRequestResult
```
prha, line 297: consider combining the two into just a `DynamicPartitionsRequestResult` to reduce boiler plate / unnecessary type checking?  It looks like it's just for display purposes...  We could check that there are either additions or deletions, not both.

clairelin135: good call

Issue #13762, opened 2023-04-19
Display added/removed dynamic partitions in tick page


clairelin135:
### What's the use case?

We now can display the information now that the daemon is responsible for mutating the partitions def

PR #13999, opened 2023-04-28
[dynamic partitions requests in tick 3/3] Add dynamic partitions request results to GQL

clairelin135:
Part 1: https://github.com/dagster-io/dagster/pull/13996
Part 2: https://github.com/dagster-io/dagster/pull/13997

This PR adds a resolver to `InstigationTick` that returns the evaluated dynamic partitions requests in the sensor. This enables displaying the evaluated dynamic partitions requests in Dagit.

Addresses the backend portion of https://github.com/dagster-io/dagster/issues/13762

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/schema/instigation.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/instigation.py
@@ -14,6 +14,8 @@ from dagster._core.definitions.schedule_definition import ScheduleExecutionData
 from dagster._core.definitions.selector import ScheduleSelector, SensorSelector
 from dagster._core.definitions.sensor_definition import SensorExecutionData
 from dagster._core.scheduler.instigation import (
+    AddDynamicPartitionsRequestResult,
+    DeleteDynamicPartitionsRequestResult,
     InstigatorState,
     InstigatorTick,
     InstigatorType,
@@ -140,6 +142,8 @@ class GrapheneDynamicPartitionsRequestType(graphene.Enum):
 
 
 class GrapheneDynamicPartitionsRequest(graphene.ObjectType):
+    # TODO maybe unify this object?
+    # Or at least use some sort of mixin?
     partitionKeys = graphene.List(graphene.NonNull(graphene.String))
     partitionsDefName = graphene.NonNull(graphene.String)
     type = graphene.NonNull(GrapheneDynamicPartitionsRequestType)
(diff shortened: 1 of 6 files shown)


python_modules/dagster-graphql/dagster_graphql_tests/graphql/test_sensors.py
```diff
@@ -397,6 +397,33 @@
...
+    ... on Sensor {
+      id
+      sensorState {
+        id
+        ticks {
+          id
+          dynamicPartitionsRequestResults {
+            partitionsDefName
+            partitionKeys
+            skippedPartitionKeys
```
salazarm, line 417: Should we add this field to the Dry run version of this too?

https://github.com/dagster-io/dagster/blob/0de69c34b0378cb795922be270fd76b5ae6a89e6/python_modules/dagster-graphql/dagster_graphql/schema/instigation.py#L378

clairelin135:
@salazarm @sryza would appreciate your eyes on this PR so we can close out this feature. Let me know what you think!

python_modules/dagster-graphql/dagster_graphql/schema/instigation.py
```diff
@@ -139,11 +140,34 @@ class Meta:
...
-class GrapheneDynamicPartitionsRequest(graphene.ObjectType):
+class DynamicPartitionsRequestMixin:
     partitionKeys = graphene.List(graphene.NonNull(graphene.String))
     partitionsDefName = graphene.NonNull(graphene.String)
     type = graphene.NonNull(GrapheneDynamicPartitionsRequestType)
 
+    class Meta:
+        name = "DynamicPartitionRequestMixin"
+
+    def get_dynamic_partitions_request(
```
sryza, line 151: I'm not super familiar with this mixin pattern, but this strikes me as a little weird. Why would a request be expected to have a request inside it?

clairelin135: A mixin is like a superclass, and subclasses inherit the methods and resolvers of the mixin. I was mostly following the pattern here: https://github.com/dagster-io/dagster/blob/master/python_modules/dagster-graphql/dagster_graphql/schema/pipelines/pipeline.py#L578-L584

The alternative would be to pass in the request object when the mixin is initialized instead.

sryza: Got it. I think it's worth adding a similar comment here.

sryza (approved):

PR #14000, opened 2023-04-28
Expose `QueuedRunCoordinator`, `SubmitRunContext`, `script_relative_path` as public and clean up docs_snippets imports

smackesey:
## Summary & Motivation

Primary goal of this PR was to make sure any dagster imports exposed in our docs snippets are top-level. Ideally we would have a lint-rule enforcing this but I don't think that's possible until ruff supports custom rules.

- Clean up `docs_snippets` imports. Ruff clean up of unused imports is disabled for `docs_snippets`, and import sorting is disabled in many individual files, so there was room for some manual improvement here.
- Make sure any imports appearing in an actual snippet are from top-level `dagster`. This required making some previous private symbols public.
- Make `script_relative_path`, `QueuedRunCoordinator`, and `SubmitRunContext` public (exported from top-level dagster). These were used in our docs snippets.
- Massage some imports in dagster core to make sure that the import changes above didn't lead to importing grpc.

## How I Tested These Changes

Existing test suite.

### DIFF
--- a/examples/docs_snippets/docs_snippets/legacy/dagster_pandas_guide/core_trip.py
+++ b/examples/docs_snippets/docs_snippets/legacy/dagster_pandas_guide/core_trip.py
@@ -3,8 +3,7 @@ from datetime import datetime
 from dagster_pandas import PandasColumn, create_dagster_pandas_dataframe_type
 from pandas import DataFrame, read_csv
 
-from dagster import Out, job, op
-from dagster._utils import script_relative_path
+from dagster import Out, file_relative_path, job, op
 
 # start_core_trip_marker_0
 TripDataFrame = create_dagster_pandas_dataframe_type(
@@ -30,7 +29,7 @@ TripDataFrame = create_dagster_pandas_dataframe_type(
 @op(out=Out(TripDataFrame))
 def load_trip_dataframe() -> DataFrame:
     return read_csv(
-        script_relative_path("./ebike_trips.csv"),
+        file_relative_path(__file__, "./ebike_trips.csv"),
         parse_dates=["start_time", "end_time"],
         date_parser=lambda x: datetime.strptime(x, "%Y-%m-%d %H:%M:%S.%f"),
         dtype={"color": "category"},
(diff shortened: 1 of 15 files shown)


examples/docs_snippets/docs_snippets/legacy/dagster_pandas_guide/core_trip.py
```diff
@@ -3,8 +3,7 @@
 from dagster_pandas import PandasColumn, create_dagster_pandas_dataframe_type
 from pandas import DataFrame, read_csv
 
-from dagster import Out, job, op
-from dagster._utils import script_relative_path
+from dagster import Out, job, op, script_relative_path
```
alangenfeld, line 6: lets just use `file_relative_path(__file__,`

python_modules/dagster/dagster/__init__.py
```diff
@@ -520,7 +524,10 @@
     deserialize_value as deserialize_value,
     serialize_value as serialize_value,
 )
-from dagster._utils import file_relative_path as file_relative_path
+from dagster._utils import (
+    file_relative_path as file_relative_path,
+    script_relative_path as script_relative_path,
```
alangenfeld, line 529: ild rather not export this, it does some cute `inspect.stack` shenanigans and `file_relative_path(__file__` isn't that much worse

smackesey: good call

python_modules/dagster/dagster/__init__.py
```diff
@@ -454,6 +454,10 @@
 from dagster._core.instance_for_test import instance_for_test as instance_for_test
 from dagster._core.launcher.default_run_launcher import DefaultRunLauncher as DefaultRunLauncher
 from dagster._core.log_manager import DagsterLogManager as DagsterLogManager
+from dagster._core.run_coordinator.queued_run_coordinator import (
+    QueuedRunCoordinator as QueuedRunCoordinator,
+    SubmitRunContext as SubmitRunContext,
+)
```
alangenfeld, line 460: some concern on import cost if there are any heavy things contained within. Theres a test for specific known heavy libs so that should prevent anything too egregious

alangenfeld (approved):
I think this makes sense - these objects don't have much "public" on them but i think if we have a documented example it makes sense to promote them like this

python_modules/dagster/dagster/__init__.py
smackesey: Yeah that test was failing until I made some minor changes

PR #14003, opened 2023-04-29
feature: implement asset selection by regex to asset key.

takeru911:
## Summary & Motivation

- AssetSelection, which selects AssetKey by regular expression, is implemented.
- We wanted to make our selection by AssetKey prefix.
  - Similar selection could be done by using all, upstream or downstream, etc., but it was difficult to understand.
  - Therefore, we wanted to use the hierarchical structure of AssetKey to do the selection.
    - I think this PR would also solve the problem he is citing
      -  https://dagster.slack.com/archives/C01U954MEER/p1681845883730989

## How I Tested These Changes

- Added test for newly implemented selections
  - Tests if the correct AssetKey is selected for a given regular expression

I am new to contribute to dagster, so if there are any problems, please point them out!


cmpadden:

PR #14004, opened 2023-04-29
more targeted checking for warnings in asset tests

sryza:
## Summary & Motivation

In some of our asset test files, we check to make sure that none of the tests spit out warnings, except where expected for experimental APIs. This change makes a few small improvements to that:
- Ignores warnings on a per-test level, instead of blanket
- Prints out the line that triggered the warning (before, you would just see the warning and not know where it came from)
- Eliminates some warning exceptions that we were no longer using


smackesey (approved):
Definitely an improvement.

I think that ge failure will go away if you rebase.

PR #14007, opened 2023-04-30
fix worker_start_command for windows when use --config-yaml

rubikplanet:
## Summary & Motivation

--config-yaml option is not working on windows because it uses a colon as the path seperator.

use os.pathsep to join the path make it work both on linux and windows.

## How I Tested These Changes

Yes, I tested manually on windows.


yuhan:
@alangenfeld @gibsondan mind taking a look? (tagging you because you two showed up in git blame, but plz feel free to re-route!)

alangenfeld:
looks reasonable, @rubikplanet  did you test this manually on windows? Could you update the `How I Tested These Changes` section?

alangenfeld (dismissed):
for previous comment

rubikplanet:
fix: f-string instead of `format`

rubikplanet:
I don't know how to split them. Instead, create another PR https://github.com/dagster-io/dagster/pull/14317.
And please close this one.

PR #14009, opened 2023-04-30
re-generate pyright pins

gibsondan:
Summary:
Should resolve several dependabot CVEs - particularly due to upgrading apache-airflow to 2.6

However I think we should probably find a way to exclude the pyright folder from dependabot?


examples/assets_pandas_pyspark/assets_pandas_pyspark/local_spark_filesystem_io_manager.py
```diff
@@ -70,7 +70,7 @@ def load_input(self, context) -> Union[PandasDF, SparkDF]:
             return pd.concat(map(pd.read_csv, paths))
         elif context.dagster_type.typing_type == SparkDF:
             return (
-                SparkSession.builder.getOrCreate()
+                SparkSession.builder.getOrCreate()  # type: ignore
```
gibsondan, line 73: pyright doesn't like something about the new classproperty here in the latest version: https://github.com/apache/spark/blob/master/python/pyspark/sql/session.py#L495-L496

I would be lying if I said I understood why some SparkSession.builder calls hit the type issue and others do not though

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -654,7 +654,7 @@ def delete_events_for_run(self, conn: Connection, run_id: str) -> None:
         )
 
         removed_asset_keys = [
-            AssetKey.from_db_string(row[0])
+            AssetKey.from_db_string(row[0])  # type: ignore
```
gibsondan, line 657: latest version of sqlalchemy types does not like indexing into rows (it thinks they are some kind of fancy tuple, even though indexing works fine still)

python_modules/dagster/dagster/_core/storage/migration/utils.py
```diff
@@ -202,8 +202,8 @@ def extract_asset_keys_idx_columns() -> None:
         return
 
     # add timestamp, tags columns to avoid event deserialization
-    op.add_column("asset_keys", db.Column("wipe_timestamp", db.types.TIMESTAMP))
-    op.add_column("asset_keys", db.Column("last_materialization_timestamp", db.types.TIMESTAMP))
+    op.add_column("asset_keys", db.Column("wipe_timestamp", db.types.TIMESTAMP))  # type: ignore
```
gibsondan, line 205: this one is very mystifying to me - the sqlalchemy.types module resolves in some situations but not in others

smackesey: Errors appear here because `sqlalchemy.types` is not imported in this module, and the `sqlalchemy/__init__.py` doesn't explictly bind `types`-- instead access to `types` is predicated on the side effect of binding the types attr on`sqlalchemy` at the time that `sqlalchemy.types` is actually imported. It's an annoying aspect of python import system.

Normally we could solve this by just making sure we have `import sqlalchemy.types` at the top, but in this case that won't work because we are going through the `db` alias. The solution I've used elsewhere in our codebase is to do:

```
import sqlalchemy.types as db_types
```

Then access the types as `db_types` instead of `db.types`.

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
smackesey: Wow that seems like a bug in the types. Hopefully it goes away with a future update.

examples/assets_pandas_pyspark/assets_pandas_pyspark/local_spark_filesystem_io_manager.py
smackesey: >I would be lying if I said I understood why some SparkSession.builder calls hit the type issue and others do not though

Would need to see an example of one that doesn't hit but my guess is that it's inside an unannotated function.

smackesey (approved):
Made suggestion re `db.types`, the rest looks good. Thanks for doing this, annoying that the pyspark and sqlalchemy bumps generated all of these false positives.

PR #14010, opened 2023-05-01
Include public members in api docs even if they don't have a docblock

gibsondan:
Summary:
User expressed surprise today that the `asset_key` property on DagsterEvent wasn't in the docs - it's a public field, so it seems like it should be in the docs, but it wasn't, because the type wasn't annotated with :undoc-members.

Now that we are filtering all api docs to things with the @public decorater, we might as well include all properties by default, rather than only including the ones with docblocks.

Separately, we should consider requiring that all public properties have docblocks and return types? But indicating their existence without docblocks seems preferable to excluding them.

To do this without confusing sphinx, we had to move some places where we had an attribute defined in the constructor but not in the method to the method instead. (Sphinx was raising warnings about duplicate entries otherwise).

Test Plan: View api docs locally, see many properties that were not there before


smackesey:
Thanks for doing this-- been on my list but haven't gotten around to it. There are some other issues with apidoc generation, like non-public attrs showing up for `NamedTuple`, but this can be addressed in a different PR.

>Separately, we should consider requiring that all public properties have docblocks and return types? But indicating their existence without docblocks seems preferable to excluding them.

Yes-- need a custom lint rule to enforce, the easiest way would be with a custom lint rule, which we can't do right now, but we can also trigger a failure in the docs build.

>To do this without confusing sphinx, we had to move some places where we had an attribute defined in the constructor but not in the method to the method instead. (Sphinx was raising warnings about duplicate entries otherwise). Another item for a follow-up that I can look into.

This is an improvement.

smackesey (approved):

Issue #13694, opened 2023-04-17
backfill error when hourly asset overhangs daily asset


sryza:
The scenario:
- An hourly-partitioned asset is upstream of a daily-partitioned asset
- There are hour partitions that don't yet have downstream date partitions - e.g. it's 5 pm on Thursday, so the 4 pm Thursday partition exists in the hourly asset, but the Thursday partition doesn't yet exist in the daily asset
- You launch a backfill over both assets that includes the latest hourly partition

What happens:
- Dagster tries to launch a run for Thursday in the downstream asset, because it launched a run for Thursday / 4 pm in the upstream asset
- The fact that there's no Thursday partition causes an error when trying to actually execute the run.

sryza:
I think one way to fix this would be to make `PartitionMapping`s aware of `current_time`, and then make it so that `TimeWindowPartitionMapping` never returns partitions that aren't part of the `PartitionsDefinition` being mapped to.

clairelin135:
I agree we should make `TimeWindowPartitionMapping`s aware of the current time. This is probably something we should've been doing before this anyway.

One consequence is that if `get_upstream_partitions_for_partitions` also respects the current time, an hourly partition would successfully execute when it doesn't yet have an upstream daily partition, which might be surprising behavior. One way to deal with this is to not pass in current time during runs, so that within a run all dependency partitions will be returned, rather than just the ones that exist at that time.

sryza:
> One consequence is that if get_upstream_partitions_for_partitions also respects the current time, an hourly partition would successfully execute when it doesn't yet have an upstream daily partition, which might be surprising behavior.

@clairelin135 are you able to explain this scenario in more detail?

clairelin135:
So say we have:
- upstream daily partitions def `DailyPartitionsDefinition("2023-01-01")`
- downstream hourly partitions def `HourlyPartitionsDefinition("2023-01-01-00:00")`

Calling `get_upstream_partitions_for_partitions(hourly_subset("2023-05-01-00:00"), downstream_partitions_def, current_time="2023-05-01")` returns an empty subset since the 2023-05-01 partition of the upstream daily partitions def does not exist yet. Subsequently a run that materializes just the "2023-05-01-00:00" partition will succeed even though "2023-05-01" is not materialized, since dagster thinks there are no upstream partition dependencies.

[...]

PR #14013, opened 2023-05-01
Make time window partition mapping current-time aware

clairelin135:
Fixes https://github.com/dagster-io/dagster/issues/13694

This PR makes time window partition mappings current-time aware. Previously, calling `get_upstream_partitions`/`get_downstream_partitions` would just return all overlapping partitions in the interval regardless of whether they exist during at the current time or not. This resolves several bugs that occur when nonexistent partitions are returned:
- When kicking off a backfill with an upstream hourly partitions def and a downstream daily partitions def, selecting an hourly partition for the current day (downstream daily partition does not exist yet) will not select any downstream daily partition. Previously the backfill would error.
- If an hourly partitions def is downstream of a daily partitions def and a run/backfill targets hourly partitions for the current day (upstream daily partition does not exist yet) an error is raised. Previously the error was raised within the IO manager when an upstream materialization was not found.

The implementation involves threading a `current_time` argument throughout affected systems:
- Backfills: passes the backfill creation time to the partition mapping as the current time. This means that only partitions that are existent when the backfill was created can be targeted.
- Asset reconciliation: Threads the evaluation time throughout the asset graph and adds it as a property to the caching instance queryer, which are both responsible for fetching upstream/downstream partitions. 
- If the current time is not provided to the partition mapping methods, fetches the current time from pendulum. This ensures that this behavior propagates to other callsites i.e. ad-hoc materializations/runs

Tested on BK and locally by launching backfills
- Daily upstream of hourly: if hourly partitions are selected that do not yet have existent daily partitions, the backfill fails on creation with the error raised from `get_upstream_partitions_for_partitions`: `Partition window TimeWindow(...) does not exist for partitions definition Daily, starting 2023-01-01 UTC.`
image

- Hourly upstream of daily: if hourly partitions are selected that do not yet have existent daily partitions, the backfill page will display 0 targeted partitions for the daily partitioned asset. I plan on fixing this in a follow-up.
image


python_modules/dagster/dagster_tests/core_tests/execution_tests/test_asset_backfill.py
```diff
@@ -649,3 +662,38 @@ def downstream_weekly_partitioned_asset(
...
+            upstream_hourly_partitioned_asset.key,
+            downstream_daily_partitioned_asset.key,
+        ],
+        dynamic_partitions_store=MagicMock(),
+        all_partitions=False,
+        evaluation_time=pendulum.datetime(2023, 1, 9, 0, 0, 0),
+    )
+
+    target_subset = backfill_data.target_subset
+    assert len(target_subset.get_partitions_subset(downstream_daily_partitioned_asset.key)) == 0
```
OwenKephart, line 699: if I'm understanding correctly, we should still have the single partition for `upstream_hourly_partitioned_asset` in the target subset, so we should check that this is the case here (in addition to this check)

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -54,6 +56,8 @@ def __init__(self, instance: DagsterInstance):
 
         self._dynamic_partitions_cache: Dict[str, Sequence[str]] = {}
 
+        self._evaluation_time = evaluation_time
```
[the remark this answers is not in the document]
OwenKephart: alternatively, we could just make this a required argument

python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -268,6 +272,7 @@ def empty(cls, target_subset: AssetGraphSubset) -> "AssetBackfillData":
             materialized_subset=AssetGraphSubset(asset_graph),
             failed_and_downstream_subset=AssetGraphSubset(asset_graph),
             latest_storage_id=None,
+            evaluation_time=evaluation_time,
```
OwenKephart, line 275: maybe this field could be called something like "backfill_start_time" or "backfill_time", as "evaluation_time" is a bit confusing.

sryza (commented):
I think it's worth adding a new scenario to execution_tests/test_asset_backfill.py.

Also, have you had a chance to play around this manually and see what happens when you launch backfills? It would be helpful to document that in the Test Plan.

python_modules/dagster/dagster/_core/definitions/time_window_partition_mapping.py
```diff
@@ -97,17 +99,72 @@ def get_upstream_partitions_for_partitions(
...
+
+        Raises an error if upstream partitions do not exist at the given current_time, fetching the
+        current time if not provided.
+        """
         if not isinstance(downstream_partitions_subset, TimeWindowPartitionsSubset):
             check.failed("downstream_partitions_subset must be a TimeWindowPartitionsSubset")
 
-        return self._map_partitions(
-            downstream_partitions_subset.partitions_def,
+        if not isinstance(upstream_partitions_def, TimeWindowPartitionsDefinition):
```
OwenKephart, line 113: it's not clear to me why the get_upstream_partitions_for_partitions and get_downstream_partitions_for_partitions implementations need to diverge from _map_partitions here. It seems like they're still doing essentially the same thing as each other

OwenKephart (commented):
overall, this approach makes sense

clairelin135:
@sryza Updated to include the backfills I ran locally in the PR description

python_modules/dagster/dagster/_core/definitions/time_window_partition_mapping.py
clairelin135: It's a small difference -- `get_upstream_partitions_for_partitions` raises errors if mapped time windows don't exist at the current time, while `get_downstream_partitions_for_partitions` just filters the time windows to only contain existent time windows.

I forked the implementations here because it felt better than passing a flag into `_map_partitions`, but I'm open to reworking this if you have recommendations

python_modules/dagster/dagster_tests/core_tests/execution_tests/test_asset_backfill.py
clairelin135: good call

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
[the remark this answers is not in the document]
clairelin135: that sounds good to me. Originally I didn't set it because there's a couple other places the caching instance queryer is used without eval time, but I think we can safely assume in these callsites that the partitions in each partitions def shouldn't change

python_modules/dagster/dagster/_core/execution/asset_backfill.py
clairelin135: good call--I'll switch to `backfill_start_time`

python_modules/dagster/dagster/_core/definitions/time_window_partition_mapping.py
OwenKephart: I think passing a flag for "raise_on_non_existent" would be better/more maintainable here. As it stands, if we ever want to update the logic for _map_partitions to be (e.g.) more efficient, we'd have to update both of these places, which runs a pretty high risk of divergence

clairelin135:
@OwenKephart @sryza thanks for reviewing!

I've updated this per the feedback above, I think it's ready for another look. Another missing piece I added is that the time partition mapping now filters out partitions that are before the partition def's start time. I didn't think this was a common use case initially, but after talking to a user today with this situation I've added it in.

python_modules/dagster/dagster/_core/definitions/time_window_partition_mapping.py
```diff
@@ -207,6 +223,45 @@ def _map_partitions(
 
                 time_windows.append(TimeWindow(window_start, window_end))
 
+        current_time = cast(
```
OwenKephart, line 226: nitpick, but shouldn't get_first_partition_window/get_last_partition_window handle this logic for us? i.e. if we pass in current_time = current_time (be it a real datetime or `None`), the partitions definition will generate a current time inside those method calls if necessary.

python_modules/dagster/dagster/_core/definitions/time_window_partition_mapping.py
```diff
@@ -207,6 +223,45 @@ def _map_partitions(
...
+            datetime,
+            pendulum.instance(current_time, tz=to_partitions_def.timezone)
+            if current_time
+            else pendulum.now(to_partitions_def.timezone),
+        )
+
+        first_window = to_partitions_def.get_first_partition_window(current_time=current_time)
+        last_window = to_partitions_def.get_last_partition_window(current_time=current_time)
+        if raise_on_non_existent_partition:
```
OwenKephart, line 235: also a nitpick, but could we save some lines of code / logic by just generating the filtered set of time windows regardless of this flag, then only raising the error if this flag is set and the filtered time windows don't equal the non-filtered time windows?

OwenKephart (commented):
I think this is close! Just a couple remaining comments

python_modules/dagster/dagster/_core/definitions/time_window_partition_mapping.py
clairelin135: oo yes i forgot about this

python_modules/dagster/dagster/_core/definitions/time_window_partition_mapping.py
clairelin135: makes sense!

python_modules/dagster/dagster/_core/storage/db_io_manager.py
```diff
@@ -232,7 +232,9 @@ def _get_table_slice(
                     partition_dimensions.append(
                         TablePartitionDimension(
                             partition_expr=cast(str, partition_expr),
-                            partitions=context.asset_partitions_time_window,
+                            partitions=context.asset_partitions_time_window
```
clairelin135, line 235: Fixes a small bug that I encountered.

Previously, the time window partition mapping would return time windows with no partitions (the start == the end). This PR fixes this behavior, but then `context_asset_partitions_time_window` would throw an error in this case. This check passes in an empty list of partitions if no partitions exist.

OwenKephart (approved):
great!

PR #14015, opened 2023-05-01
Fix GQL source asset jobNames

smackesey:
## Summary & Motivation

Fixes observable source asset showing up where it does not belong as reported here and here:

- https://elementl-workspace.slack.com/archives/C03C3E81TNE/p1682709168331509
- https://elementl-workspace.slack.com/archives/C03CCE471E0/p1682613364691559

## How I Tested These Changes

Adjust existing test to make sure source asset not included where it shouldn't be.

### DIFF
--- a/python_modules/dagster/dagster/_core/host_representation/external_data.py
+++ b/python_modules/dagster/dagster/_core/host_representation/external_data.py
@@ -1392,8 +1392,16 @@ def external_asset_graph_from_defs(
                     job_def.name
                     for job_def in job_defs
                     if source_asset.key in job_def.asset_layer.source_assets_by_key
-                    and source_asset.partitions_def is None
-                    or source_asset.partitions_def == job_def.partitions_def
+                    and (
+                        # explicit source-asset observation job
+                        not job_def.asset_layer.has_assets_defs
+                        # "base asset job" will have both source and materializable assets
+                        or is_base_asset_job_name(job_def.name)
+                        and (
+                            source_asset.partitions_def is None
+                            or source_asset.partitions_def == job_def.partitions_def
+                        )
+                    )
                 ]
                 if source_asset.node_def is not None
                 else []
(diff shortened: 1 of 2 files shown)


smackesey:
* master
  * **PR #14015** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14015" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14015?utm_source=stack-comment).

python_modules/dagster/dagster/_core/host_representation/external_data.py
```diff
@@ -1392,8 +1392,16 @@ def external_asset_graph_from_defs(
                     job_def.name
                     for job_def in job_defs
                     if source_asset.key in job_def.asset_layer.source_assets_by_key
-                    and source_asset.partitions_def is None
-                    or source_asset.partitions_def == job_def.partitions_def
+                    and (
+                        # explicit source-asset observation job
+                        not job_def.asset_layer.has_assets_defs
+                        # "base asset job" will have both source and materializable assets
+                        or is_base_asset_job_name(job_def.name)
```
OwenKephart, line 1399: if I'm understanding correctly, the code here is just to ensure that source assets will reference the "base asset job" in the list of jobs they are a part of. This doesn't seem necessary, as we don't display this job to the user (it's just a trick to let us do ad-hoc materializations). In the interest of simplicity, would prefer to just remove this

smackesey: This _is_ necessary for the "Observe" button to work correctly when you select an observable in the Asset Graph Explorer. Observations launched in this way are executed as a selection on one of these base asset jobs.

bengotow (approved):
This looks good to me!

python_modules/dagster/dagster/_core/host_representation/external_data.py
OwenKephart: ah makes sense

PR #14019, opened 2023-05-01
split external url for compute logs into stdout/stderr, put under test

prha:
## Summary & Motivation
Allows the compute log manager to specify different external urls for stdout/stderr to be displayed in the compute log panel in Dagit.  Also adds a reference S3 implementation for the `dagster_aws` library.

## How I Tested These Changes
BK, manually saw stdout/stderr urls displayed in compute log panel

### DIFF
--- a/js_modules/dagit/packages/core/src/runs/LogsRow.tsx
+++ b/js_modules/dagit/packages/core/src/runs/LogsRow.tsx
@@ -185,6 +185,8 @@ export const LOGS_ROW_STRUCTURED_FRAGMENT = gql`
       fileKey
       stepKeys
       externalUrl
+      externalStdoutUrl
+      externalStderrUrl
     }
   }
 
--- a/python_modules/dagster-graphql/dagster_graphql/implementation/events.py
+++ b/python_modules/dagster-graphql/dagster_graphql/implementation/events.py
@@ -369,6 +369,8 @@ def from_dagster_event_record(event_record: EventLogEntry, pipeline_name: str) -
             logKey=data.file_key,
             stepKeys=data.step_keys,
             externalUrl=data.external_url,
+            externalStdoutUrl=data.external_stdout_url or data.external_url,
+            externalStderrUrl=data.external_stderr_url or data.external_url,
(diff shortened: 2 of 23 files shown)


js_modules/dagit/packages/core/src/runs/LogsRow.tsx
```diff
@@ -185,6 +185,8 @@ export const LOGS_ROW_STRUCTURED_FRAGMENT = gql`
       fileKey
       stepKeys
       externalUrl
+      externalStdoutUrl
+      externalStderrUrl
```
gibsondan, line 189: could also make it a filter on externalUrl with an enum key? like ```externalUrl(STDOUT)``` do you think we'll ever support additional stream types?

gibsondan: Do we still need externalUrl?

python_modules/dagster-graphql/dagster_graphql/implementation/events.py
```diff
@@ -369,6 +369,8 @@ def from_dagster_event_record(event_record: EventLogEntry, pipeline_name: str) -
             logKey=data.file_key,
             stepKeys=data.step_keys,
             externalUrl=data.external_url,
+            externalStdoutUrl=data.external_stdout_url or data.external_url,
+            externalStderrUrl=data.external_stderr_url or data.external_url,
```
gibsondan, line 373: maybe we should pick one as the Default one rather than using the same URL for both STDOUT and STDERR? maybe just send it to STDOUT in that case?

gibsondan (approved):
few optional thoughts inline

js_modules/dagit/packages/core/src/runs/LogsRow.tsx
prha: I don't think we need `externalUrl` anymore... you probably want both types because they correspond to the different panel views.

We do need to keep supporting the serialized event metadata field `external_url` though (although, I'm fairly confident that no user OSS or cloud has ever populated it besides me).

PR #14023, opened 2023-05-01
[oss] React 17 -> React 18

salazarm:
## Summary & Motivation

Updating React version to 18 to unlock React concurrent mode. Still need to update the root 

The only not so ideal thing I found was that react-virtualized needs to update its typed for react-18 so I had to `any` type it.

[INTERNAL_BRANCH=react18]

## How I Tested These Changes

`yarn ts` `yarn lint` `yarn jest`


js_modules/dagit/packages/ui/src/components/Suggest.tsx
```diff
@@ -4,13 +4,16 @@ import {InputGroupProps2, IPopoverProps} from '@blueprintjs/core';
 import {isCreateNewItem, Suggest as BlueprintSuggest, SuggestProps} from '@blueprintjs/select';
 import deepmerge from 'deepmerge';
 import * as React from 'react';
-import {List} from 'react-virtualized';
+import {List as _List} from 'react-virtualized';
```
bengotow, line 7: I wonder if we should go ahead and switch this back to the tan-stack one soon. What were the problems you were having with the react-virtualized library on Windows?

bengotow (approved):
This looks good to me!

js_modules/dagit/tox.ini
```diff
@@ -22,15 +22,16 @@ commands =
     yarn install
   yarn workspace @dagster-io/dagit-core generate-graphql
   yarn workspace @dagster-io/dagit-core generate-perms
-  yarn workspace @dagster-io/dagit-app lint
   yarn workspace @dagster-io/dagit-app ts
+  yarn workspace @dagster-io/dagit-app lint
   yarn workspace @dagster-io/dagit-core ts
+  yarn workspace @dagster-io/dagit-core lint
+  yarn workspace @dagster-io/ui ts
+  yarn workspace @dagster-io/ui lint
```
salazarm, line 30: added missing `lint` call for @dagster-io/core and also moved all the ts/lint checks to happen before we run any tests.

PR #13984, opened 2023-04-28
solid renames

smackesey:
## Summary & Motivation

- Rename a bunch of local variables containing "solid" to op equivalents
- Change "solid"/"composite solid" in docstrings/comments to "op"/"node"/"graph"

Non-local changes

- `resolve_checked_solid_fn_inputs` -> `resolved_checked_op_fn_inputs`
- delete `OpExecutionContext.solid` (.`op` already exists)
- `StepExecutionContext.solid` -> `op`
- `StepExecutionContext.solid_retry_policy` -> `op_retry_policy`
- `ExternalJob.solid_names_in_topological_order` -> `node_names_in_topological_order`
- `ExternalJob.solid_names` -> `node_names`
- `ExternalJob.has_solid_invocation` -> `has_node_invocation`
- `JobIndex.has_solid_invocation` -> `has_node_invocation`
- `DagsterLoggingMetadata.solid_name` -> `op_name`

## How I Tested These Changes

Existing test suite

PR #13982, opened 2023-04-28
rename test_*pipeline/solid* -> test_*job/op*

smackesey:
## Summary & Motivation

Internal companion PR: https://github.com/dagster-io/internal/pull/5554

Rename test functions containing "pipeline" or "solid" to "job"/"op" equivalent.

## How I Tested These Changes

Existing test suite.

PR #13981, opened 2023-04-28
pipeline local variable renames

smackesey:
## Summary & Motivation

- Rename many local variable and docstring instances of "pipeline"/"solid" to "job"/"op"
- Rename a few non-local (but not public API) "pipeline" arg names in `dagster_k8s`, `dagster_celery_k8s`, `dagster_docker`

## How I Tested These Changes

Existing test suite

PR #13496, opened 2023-04-11
Module renames (pipeline/soild -> job/op)

smackesey:
## Summary & Motivation

Internal companion PR: https://github.com/dagster-io/internal/pull/5553

Rename python modules containing "pipeline" or "solid" to "op/job" equivalents. Probably 75% of the changes in this PR  are just from renaming `dagster._core.storage.pipeline_run` -> `dagster_run`. Here is the full list:

- rename assorted test files
- fold decorators_tests.test_solid -> test_op
- rm dagster_aws.s3.solids (.ops already exists)
- rm dagster_gcp.dataproc.solids (.ops already exists)
- rm dagster_gcp.bigquery.solids (.ops already exists)
- dagster_shell.solids -> ops
- dagster_snowflake.solids -> ops
- dagster._core.utility_solids -> utility_ops
- dagster._core.snap.pipeline_snapshot -> job_snapshot
- dagster._core.host_representation.pipeline_index -> job_index
- dagster._core.execution.context_creation_pipeline -> context_creation_job
- dagster._core.definitions.pipeline_base -> job_base
- dagster._api.snapshot_pipeline -> snapshot_job
- dagster._core.storage.pipeline_run -> dagster_run

## How I Tested These Changes

Existing test suite.

PR #14024, opened 2023-05-01
Refactor job subset specification

smackesey:
## Summary & Motivation

The current handling of job subsetting is a bit of a mess due to inherited technical debt from `PipelineDefinition`. Now that `PipelineDefinition` is gone, there is a lot of room for improvement. This PR is the first in a series that will aim to refactor the subsetting and execution APIs for simplicity, clarity, and consistency. In this PR:

- Consolidate two subsetting methods `subset_for_execution` and `subset_for_execution_from_existing_job` on `IJob` class into new `get_subset` method.
- Replace `IJob.solids_to_execute` with `op_selection`
- Change fields on `ReconstructableJob` to accommodate new interface
- Move op selection resolution to new `op_selection` module. Introduce `OpSelection` class (currently an implementation detail). This parallels the existing `AssetSelection` and  is part of unifying our approach to `asset_selection` and `op_selection`.
- `JobDefinition.is_subset_job` -> `is_subset`

## How I Tested These Changes

Existing test suite, with some tests modified to account for changed APIs.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/dependency.py
+++ b/python_modules/dagster/dagster/_core/definitions/dependency.py
@@ -586,6 +586,10 @@ class NodeOutput(NamedTuple("_NodeOutput", [("node", Node), ("output_def", Outpu
     def is_dynamic(self) -> bool:
         return self.output_def.is_dynamic
 
+    @property
+    def output_name(self) -> str:
+        return self.output_def.name
+
 
 class DependencyType(Enum):
     DIRECT = "DIRECT"
--- a/python_modules/dagster/dagster/_core/definitions/job_base.py
+++ b/python_modules/dagster/dagster/_core/definitions/job_base.py
@@ -1,9 +1,8 @@
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, AbstractSet, Iterable, Optional, Sequence
+from typing import TYPE_CHECKING, AbstractSet, Iterable, Optional
 
 from typing_extensions import Self
 
-import dagster._check as check
 from dagster._core.definitions.events import AssetKey
 
 if TYPE_CHECKING:
@@ -24,6 +23,7 @@ class IJob(ABC):
     @abstractmethod
     def get_subset(
         self,
+        *,
         op_selection: Optional[Iterable[str]] = None,
         asset_selection: Optional[AbstractSet[AssetKey]] = None,
     ) -> "IJob":
@@ -31,7 +31,7 @@ class IJob(ABC):
 
     @property
     @abstractmethod
-    def op_selection(self) -> Optional[Sequence[str]]:
+    def op_selection(self) -> Optional[AbstractSet[str]]:
         pass
 
     @property
@@ -48,41 +48,27 @@ class InMemoryJob(IJob):
     def __init__(
         self,
         job_def: "JobDefinition",
-        solid_selection: Optional[Iterable[str]] = None,
-        asset_selection: Optional[AbstractSet[AssetKey]] = None,
     ):
         self._job_def = job_def
-        self._solid_selection = list(solid_selection) if solid_selection else None
-        self._asset_selection = asset_selection
 
     def get_definition(self) -> "JobDefinition":
         return self._job_def
 
     def get_subset(
         self,
+        *,
         op_selection: Optional[Iterable[str]] = None,
         asset_selection: Optional[AbstractSet[AssetKey]] = None,
     ) -> Self:
-        if op_selection and asset_selection:
-            check.failed(
-                "solid_selection and asset_selection cannot both be provided as arguments",
-            )
-        elif op_selection:
-            op_selection = list(op_selection)
-            return InMemoryJob(self._job_def.get_subset(op_selection))
-        elif asset_selection:
-            return InMemoryJob(
-                self._job_def.get_subset(asset_selection=asset_selection),
-                asset_selection=asset_selection,
-            )
-        else:
-            check.failed("Must provide solid_selection or asset_selection")
+        op_selection = set(op_selection) if op_selection else None
+        return InMemoryJob(
+            self._job_def.get_subset(op_selection=op_selection, asset_selection=asset_selection)
+        )
 
     @property
-    def op_selection(self) -> Sequence[str]:
-        # a list of solid queries provided by the user
-        return self._solid_selection  # type: ignore  # (possible none)
+    def op_selection(self) -> Optional[AbstractSet[str]]:
+        return self._job_def.op_selection
 
     @property
     def asset_selection(self) -> Optional[AbstractSet[AssetKey]]:
-        return self._asset_selection
+        return self._job_def.asset_selection
--- a/python_modules/dagster/dagster/_core/definitions/job_base.py
+++ b/python_modules/dagster/dagster/_core/definitions/job_base.py
@@ -1,12 +1,10 @@
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, AbstractSet, Optional, Sequence
+from typing import TYPE_CHECKING, AbstractSet, Iterable, Optional, Sequence
 
 from typing_extensions import Self
 
 import dagster._check as check
 from dagster._core.definitions.events import AssetKey
-from dagster._core.errors import DagsterInvalidSubsetError
-from dagster._core.selector import parse_solid_selection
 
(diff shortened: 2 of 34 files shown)


smackesey:
* master
  * **PR #13507** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13507" target="_blank">Graphite</a> 
    * **PR #13482** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13482" target="_blank">Graphite</a> 
    * **PR #13327** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13327" target="_blank">Graphite</a> 
      * **PR #13496** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13496" target="_blank">Graphite</a> 
        * **PR #13981** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13981" target="_blank">Graphite</a> 
          * **PR #13982** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13982" target="_blank">Graphite</a> 
            * **PR #13984** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13984" target="_blank">Graphite</a> 
              * **PR #13994** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/13994" target="_blank">Graphite</a> 
                * **PR #14024** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14024" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14024?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/dependency.py
```diff
@@ -586,6 +586,10 @@ def node_name(self) -> str:
     def is_dynamic(self) -> bool:
         return self.output_def.is_dynamic
 
+    @property
+    def output_name(self) -> str:
+        return self.output_def.name
```
smackesey, line 591: There is a corresponding `input_name` on `NodeInputHandle`, so I added this for symmetry.

python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -918,6 +906,17 @@ def with_executor_def(self, executor_def: ExecutorDefinition) -> "JobDefinition"
...
+        return self
+
+    @property
+    def op_selection(self) -> Optional[AbstractSet[str]]:
+        return self.op_selection_data.resolved_op_selection if self.op_selection_data else None
+
+    @property
+    def asset_selection(self) -> Optional[AbstractSet[AssetKey]]:
+        return self.asset_selection_data.asset_selection if self.asset_selection_data else None
+
```
smackesey, line 919: These three methods implement `IJob` on `JobDefinition`

python_modules/dagster/dagster/_core/definitions/job_base.py
```diff
@@ -24,127 +19,23 @@ def get_definition(self) -> "JobDefinition":
...
-    @abstractmethod
-    def subset_for_execution_from_existing_job(
-        self,
-        solids_to_execute: Optional[AbstractSet[str]] = None,
-        asset_selection: Optional[AbstractSet[AssetKey]] = None,
-    ) -> "IJob":
-        pass
-
-
-class InMemoryJob(IJob, object):
```
alangenfeld, line 53: Having `JobDefinition` implement `IJob` directly makes it possible to pass the "unwrapped" `JobDefinition` from a  `ReconstructableJob` down the callstack and lose its reconstruction context. Worried that blurring the container type and contained type together may be more confusing.

python_modules/dagster/dagster/_core/definitions/job_base.py
```diff
@@ -24,127 +19,23 @@ def get_definition(self) -> "JobDefinition":
         pass
 
     @abstractmethod
-    def subset_for_execution(
+    def get_subset(
         self,
-        solid_selection: Optional[Sequence[str]] = None,
+        op_selection: Optional[Iterable[str]] = None,
```
alangenfeld, line 24: should we do the `*` trick to enforce kwargs on call sites here? Might be nice

python_modules/dagster/dagster/_core/definitions/job_base.py
smackesey: Currently it is possible to pass an `InMemoryJob` down the callstack with no reconstruction context-- in cases where that context is required, the type can be limited to `ReconstructableJob` rather than `IJob`.

>Worried that blurring the container type and contained type together may be more confusing.

IMO the status quo is over-complex because `JobDefinition` was already implementing subsetting methods, the seeming reason for existence of `IJob`, just with different names.

With this change there is no longer really a container/contained dichotomy-- instead `IJob` just means "some subsettable representation of a job.

@dpeng817 what do you think?

alangenfeld: > implementing subsetting methods, the seeming reason for existence of IJob

The reason for existence of `IJob` is in the doc block for the class 

> enables these execution APIs to operate on both in memory job
    definitions to be executed in the current process (InMemoryJob) as well as definitions that
    can be reconstructed and executed in a different process (ReconstructableJob)
    
> in cases where that context is required, the type can be limited to ReconstructableJob rather than IJob

This isn't really viable, at least with out further changes, due to how the APIs are layered/shared. Most of the execution APIs in question that operate on `IJob` are hit in both in process and multi process execution. Only once the executor definition is resolved lower in the callstack do you know whether you need reconstruction information. 
    
The subsetting APIs operating on `IJob` / `ReconstructableJob` were to be able to encode the subsetting information on the pointer object that gets passed across process boundaries. I agree this part is messy - and i believe there is inconsistency around callsites expecting to unpack an already subsetted job vs calling subset on the job definition post unpack. 


Some history on this https://dagster.phacility.com/D2739 - this current sub-optimal state was to escape an even worse previous state.

smackesey: I think `InMemoryJob` has left a bad taste in my mouth from trying to parse callsites like we find in `dagster/_utils/dagster_type.py`, where we are alternatively passing these two representations of a job to different functions, but it turns out that the `InMemoryJob` doesn't actually offer any additional functionality on top of `JobDefinition`, and on the other end of the call site there is just an `isinstance` check for `ReconstructableJob` and then the rest of the logic is accessing the underlying `JobDefinition` of the passed `InMemoryJob` (this is what happens inside `create_execution_plan`).

```
    job = InMemoryJob(GraphDefinition(node_defs=[], name="empty").to_job())
    job_def = job.get_definition()

    instance = DagsterInstance.ephemeral()
    execution_plan = create_execution_plan(job)
    dagster_run = instance.create_run_for_job(job_def)
    with scoped_job_context(execution_plan, job, {}, dagster_run, instance) as context:
```

It's possible I haven't fully grasped things, but having spent a lot of time looking at these codepaths recently, I have a strong opinion that "net confusion" is reduced by simplifying sites like the above by eliminating `InMemoryJob`. I think it would be different if `InMemoryJob` was used more like an "over-the-wire" representation, where we wrapped a `JobDefinition` in it, passed it through some code that only accessed limited `IJob` APIs, and then "unwrapped it" into a `JobDefinition` at the end. But instead there's a lot of intermittent access to the definition, which makes `InMemoryJob` feel like an unnecessary wrapper-- an abstraction barrier that isn't really respected.

That said I think you know these APIs better than I do-- if you'd prefer to keep `InMemoryJob` I can revert that part of this PR and just do the subset stuff.

python_modules/dagster/dagster/_core/definitions/job_base.py
smackesey: good idea

python_modules/dagster/dagster/_core/definitions/job_base.py
alangenfeld: I think reverting it out of this PR is right since the rest of the changes are not contended and good.

alangenfeld: I agree with your assessment that `InMemoryJob` sucks. Its only reason for existence is to make it harder to mess up with `ReconstructableJob` in the core execution APIs. I am sure there is a better setup, but I'm not yet sold on this proposed approach.

alangenfeld (approved):
nice

PR #14025, opened 2023-05-01
resources and io manager telemetry

### LINKED ISSUE #13901: add telemetry for use of IO managers
Some of these might be more difficult to determine than others, but it would be helpful to understand the adoption of:
* Each of the built-in integration IO managers, i.e. s3, snowflake, gcs, duckdb, etc.
* Custom user-authored IO managers

jamiedemaria:
## Summary & Motivation
Adds a `_dagster_maintained` property for pythonic Resources and I/O managers and two decorators `dagster_maintained_resource` and `dagster_maintained_io_manager` that can be used to indicate that we maintain a resource or io manager. These values are piped through to telemetry. This allows us to see which resources are being used by users and if users are creating custom resources. 

See https://github.com/dagster-io/dagster/pull/14467 for an example telemetry output

#### For pythonic resources 

The base `ConfigurableResourceFactory` class has a `_dagster_maintained` property that defaults to `False`. All of our resources and IO managers will need to set this property to `True`

```python
    foo: str

    @classmethod
    @property
    def _dagster_maintained(cls) -> bool:
        return True 
```

This property is then piped through to the `ExternalResourceData` which we can read when collection telemetry 

#### For functional resources:
A decorator `@dagster_maintained_resource` and `@dagster_maintained_io_manager` will wrap all `@resource` and `@io_manager`. This decorator reaches into the `ResourceDefinition` or `IOManagerDefinition` and updates a `_dagster_maintained` attribute 

```python
@dagster_maintained_io_manager
@io_manager
def my_io_manager():
    ....

@dagster_maintained_resource
@resource
def my_resource():
    ....
```
The `_dagster_maintained` attribute is then piped through to the `ExternalResourceData` which we can read when collection telemetry 

### Questions
* I made the `_dagster_maintained` attribute/property with the _ prefix so that it indicates to users that it's a private property that they shouldn't mess with. But this causes a lot of lint errors since I have to read + modify the property in certain situations. I can put all of the lint ignore directives in, but wanted to check that there isn't another route we should go down instead. 

To Do:
- [X] resource to resource dependencies are not being captured. Need to figure out how to do that - decision: going to hold off on this for now since it looks like a big lift. Want to get the first version in a tested and then can address nested resources.

resolves #13901

--- 
Old description
I've been starting to look at adding telemetry for resources and io managers. I put together this really rough example of one of the ways we could collect this data.

It's got a few drawbacks, mainly that we need to keep a list of all of the dagster supported io managers and resources so that we can match on them in the telemetry file. Is there a better way to do this? Just looking for `dagster` in the resource_type field would return false positives if the user code package is named something like `my_dagster_project` or anything else with dagster in it.

Another issue that i'll potentially need to figure out is how to tell the difference between user authored resources and user authored io managers. I'm treating them as the same right now, but may need to find ways to introspect on the classes to distinguish resources from io managers if that is useful info to collect

In what's implemented in the rfc right now, project_fully_featured  (with some additional io managers and resources) would give the telemetry dictionary:

--- 
5/11 version

```
'dagster_resources':
     "[{'module_name': 'dagster_dbt', 'class_name': 'DbtCliClientResource'}, {'module_name': 'dagster_snowflake', 'class_name': 'SnowflakeResource'}, {'module_name': 'dagster_snowflake', 'class_name': 'snowflake_resource'}]"
'dagster_io_managers':
     "[{'module_name': 'dagster_snowflake_pandas', 'class_name': 'SnowflakePandasIOManager'}, {'module_name': 'dagster_snowflake', 'class_name': 'snowflake_io_manager'}]"
'has_custom_resources':
     'True'
```
depending on how we determine if a resource/io manager is dagster's vs custom we might collapse the `dagster_resources` and `dagster_io_managers` into a single dictionary and split them out in post-processing

---- 
Original version

```
'dagster_resources':
    "['dagster_dbt.cli.resources.DbtCliClientResource', 'dagster_snowflake.resources.SnowflakeResource', 'dagster_snowflake.resources.snowflake_resource']"
'dagster_io_managers':
    "['dagster_snowflake_pandas.snowflake_pandas_type_handler.SnowflakePandasIOManager', 'dagster_snowflake.snowflake_io_manager.snowflake_io_manager']"
'custom_resources':
    "['project_fully_featured.resources.hn_resource.HNAPIClient', 'project_fully_featured.resources.parquet_io_manager.LocalPartitionedParquetIOManager', 'project_fully_featured.resources.duckdb_parquet_io_manager.DuckDBPartitionedParquetIOManager']"

```

----
One potential alternate solution could be to add a private attr to the resource/io manager base classes called something like `_resource_maintianed_by_dagster` or `_io_manager_maintained_by_dagster` which defaults to False and then we manually set it to True in all of our resources/io managers. Then we'd pipe that through to the snapshot and use that in the telemetry

### DIFF
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -1173,7 +1173,7 @@ class PartialResource(Generic[TResValue], AllowDelayedDependencies, MakeConfigCa
             description=self._state__internal__.description,
             resolve_resource_keys=self._resolve_required_resource_keys,
             nested_resources=self.nested_resources,
-            dagster_maintained=self.resource_cls._dagster_maintained,
+            dagster_maintained=self.resource_cls._dagster_maintained,  # noqa: SLF001
         )
 
 
@@ -1368,7 +1368,7 @@ class PartialIOManager(Generic[TResValue], PartialResource[TResValue]):
             nested_resources=self._state__internal__.nested_resources,
             input_config_schema=input_config_schema,
             output_config_schema=output_config_schema,
-            dagster_maintained=self.resource_cls._dagster_maintained,
+            dagster_maintained=self.resource_cls._dagster_maintained,  # noqa: SLF001
         )
 
 
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -1165,7 +1165,7 @@ class PartialResource(Generic[TResValue], AllowDelayedDependencies, MakeConfigCa
     @cached_method
     def get_resource_definition(self) -> ConfigurableResourceFactoryResourceDefinition:
         return ConfigurableResourceFactoryResourceDefinition(
-            self.__class__,
+            self.resource_cls,
             resource_fn=self._state__internal__.resource_fn,
             config_schema=self._state__internal__.config_schema,
             description=self._state__internal__.description,
@@ -1358,7 +1358,7 @@ class PartialIOManager(Generic[TResValue], PartialResource[TResValue]):
             output_config_schema = factory_cls.output_config_schema()
 
         return ConfigurableIOManagerFactoryResourceDefinition(
-            self.__class__,
+            self.resource_cls,
             resource_fn=self._state__internal__.resource_fn,
             config_schema=self._state__internal__.config_schema,
             description=self._state__internal__.description,
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -18,6 +18,7 @@ from typing import (
     Union,
     cast,
 )
+import dagster
 
 from pydantic import ConstrainedFloat, ConstrainedInt, ConstrainedStr
 from typing_extensions import TypeAlias, TypeGuard, get_args
@@ -620,6 +621,7 @@ class ConfigurableResourceFactoryResourceDefinition(ResourceDefinition, AllowDel
         description: Optional[str],
         resolve_resource_keys: Callable[[Mapping[int, str]], AbstractSet[str]],
         nested_resources: Mapping[str, CoercibleToResource],
+        dagster_maintained: bool = False,
     ):
         super().__init__(
             resource_fn=resource_fn,
@@ -629,6 +631,7 @@ class ConfigurableResourceFactoryResourceDefinition(ResourceDefinition, AllowDel
         self._configurable_resource_cls = configurable_resource_cls
         self._resolve_resource_keys = resolve_resource_keys
         self._nested_resources = nested_resources
+        self._dagster_maintained = dagster_maintained
 
     @property
     def configurable_resource_cls(self) -> Type:
@@ -657,6 +660,7 @@ class ConfigurableIOManagerFactoryResourceDefinition(IOManagerDefinition, AllowD
         nested_resources: Mapping[str, CoercibleToResource],
         input_config_schema: Optional[Union[CoercableToConfigSchema, Type[Config]]] = None,
         output_config_schema: Optional[Union[CoercableToConfigSchema, Type[Config]]] = None,
+        dagster_maintained: bool = False,
     ):
         input_config_schema_resolved: CoercableToConfigSchema = (
             cast(Type[Config], input_config_schema).to_config_schema()
@@ -678,6 +682,7 @@ class ConfigurableIOManagerFactoryResourceDefinition(IOManagerDefinition, AllowD
         self._resolve_resource_keys = resolve_resource_keys
         self._nested_resources = nested_resources
         self._configurable_resource_cls = configurable_resource_cls
+        self._dagster_maintained = dagster_maintained
 
     @property
     def configurable_resource_cls(self) -> Type:
@@ -811,6 +816,11 @@ class ConfigurableResourceFactory(
         return self._state__internal__.resolved_config_dict
 
     @classmethod
+    @property
+    def _dagster_maintained(cls) -> bool:
+        return False
+
+    @classmethod
     def _is_cm_resource_cls(cls: Type["ConfigurableResourceFactory"]) -> bool:
         return (
             cls.yield_for_execution != ConfigurableResourceFactory.yield_for_execution
@@ -833,6 +843,7 @@ class ConfigurableResourceFactory(
             description=self.__doc__,
             resolve_resource_keys=self._resolve_required_resource_keys,
             nested_resources=self.nested_resources,
+            dagster_maintained=self._dagster_maintained,
         )
 
     @abstractmethod
@@ -1161,6 +1172,7 @@ class PartialResource(Generic[TResValue], AllowDelayedDependencies, MakeConfigCa
             description=self._state__internal__.description,
             resolve_resource_keys=self._resolve_required_resource_keys,
             nested_resources=self.nested_resources,
+            dagster_maintained=self.resource_cls._dagster_maintained,
         )
 
 
@@ -1230,6 +1242,7 @@ class ConfigurableLegacyResourceAdapter(ConfigurableResource, ABC):
             description=self.__doc__,
             resolve_resource_keys=self._resolve_required_resource_keys,
             nested_resources=self.nested_resources,
+            dagster_maintained=self._dagster_maintained,
         )
 
     def __call__(self, *args, **kwargs):
@@ -1310,6 +1323,7 @@ class ConfigurableIOManagerFactory(ConfigurableResourceFactory[TIOManagerValue])
             nested_resources=self.nested_resources,
             input_config_schema=self.__class__.input_config_schema(),
             output_config_schema=self.__class__.output_config_schema(),
+            dagster_maintained=self._dagster_maintained,
         )
 
     @classmethod
@@ -1353,6 +1367,7 @@ class PartialIOManager(Generic[TResValue], PartialResource[TResValue]):
             nested_resources=self._state__internal__.nested_resources,
             input_config_schema=input_config_schema,
             output_config_schema=output_config_schema,
+            dagster_maintained=self.resource_cls._dagster_maintained,
         )
 
 
@@ -1608,6 +1623,7 @@ class ConfigurableLegacyIOManagerAdapter(ConfigurableIOManagerFactory):
             description=self.__doc__,
             resolve_resource_keys=self._resolve_required_resource_keys,
             nested_resources=self.nested_resources,
+            dagster_maintained=self._dagster_maintained,
         )
 
 
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -18,7 +18,6 @@ from typing import (
     Union,
     cast,
 )
-import dagster
 
 from pydantic import ConstrainedFloat, ConstrainedInt, ConstrainedStr
 from typing_extensions import TypeAlias, TypeGuard, get_args
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -817,6 +817,8 @@ class ConfigurableResourceFactory(
     @classmethod
     @property
     def _dagster_maintained(cls) -> bool:
+        """This property should be overridden to return True by all dagster maintained resources and IO managers.
+        """
         return False
 
     @classmethod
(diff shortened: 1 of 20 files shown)


jamiedemaria:
* master
  * **PR #14025** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14025" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14025?utm_source=stack-comment).

python_modules/dagster/dagster/_core/telemetry.py
```diff
@@ -538,6 +545,49 @@ def get_stats_from_external_repo(external_repo: "ExternalRepository") -> Mapping
...
+            used_dagster_io_managers.append(resource_type)
+        else:
+            used_custom_resources.append(resource_type)
+
+    d = {
+        "dagster_resources": str(used_dagster_resources),
+        "dagster_io_managers": str(used_dagster_io_managers),
+        "custom_resources": str(used_custom_resources),
+        # "custom_io_managers": str(used_custom_io_managers),
+    }
```
yuhan, line 586: i wonder if we could just send raw blob per line like `{"resource_name": "dagster_snowflake_pandas.snowflake_pandas_type_handler.SnowflakePandasIOManager"}` and aggregate it in post-processing in a purina job. so that if anything we can post-process or backfill as opposed to updating telemetry.

python_modules/dagster/dagster/_core/telemetry.py
```diff
@@ -538,6 +545,49 @@ def get_stats_from_external_repo(external_repo: "ExternalRepository") -> Mapping
...
+def get_resource_stats(external_resources: Sequence["ExternalResource"]) -> Mapping[str, str]:
+    dagster_resources = [
+        "dagster_snowflake.resources.SnowflakeResource",
+        "dagster_dbt.cli.resources.DbtCliClientResource",
+        "dagster_snowflake.resources.snowflake_resource"
+        # fill out with rest of resources
+    ]
+
+    dagster_io_managers = [
+        "dagster_snowflake_pandas.snowflake_pandas_type_handler.SnowflakePandasIOManager",
```
yuhan, line 557: would it be possible to separate module name and class name, so something like:
```
{"module_name": "dagster_snowflake_pandas", "class_name": "dagster_snowflake_pandas.snowflake_pandas_type_handler.SnowflakePandasIOManager"}
```
so we can easily tell whether it's a custom or builtin resource by the module name in post processing

python_modules/dagster/dagster/_core/telemetry.py
jamiedemaria: yeah that makes a lot of sense

sryza: I like the idea of nesting the resources so that the top level in the hierarchy doesn't get to crowded

if we do this, we should avoid including custom IO managers in the blob, because I think that would be an invasion of users' privacy

I also think we should keep this resilient to refactors of our integration libraries. i.e. I think the names should use what gets imported by the user, not include internal packages in the path.

jamiedemaria: > if we do this, we should avoid including custom IO managers in the blob, because I think that would be an invasion of users' privacy

makes sense, do you think having a boolean indicating the presence of custom resources or io managers would be ok? that seems like useful info to know 

> I also think we should keep this resilient to refactors of our integration libraries. i.e. I think the names should use what gets imported by the user, not include internal packages in the path.

that also makes sense. Right now the info we get in telemetry is the full class path, but i could cut out the middle bit and just keep the module (ie `dagster-snowflake-pandas`) and the class name (ie `SnowflakePandasIOManager`)

jamiedemaria: ok so in sum it seems like we want to:
1. record the module and class name of all dagster resources and IO managers that are used
2. potentially have a boolean indicating if there are custom resources/io managers 

The main difficulty with 1 will be determining which resources are dagster's vs which ones are the users

sryza: > makes sense, do you think having a boolean indicating the presence of custom resources or io managers would be ok? that seems like useful info to know

yeah, definitely

sryza:
> 5/11 version
```
'dagster_resources':
     "[{'module_name': 'dagster_dbt', 'class_name': 'DbtCliClientResource'}, {'module_name': 'dagster_snowflake', 'class_name': 'SnowflakeResource'}, {'module_name': 'dagster_snowflake', 'class_name': 'snowflake_resource'}]"
'dagster_io_managers':
     "[{'module_name': 'dagster_snowflake_pandas', 'class_name': 'SnowflakePandasIOManager'}, {'module_name': 'dagster_snowflake', 'class_name': 'snowflake_io_manager'}]"
'has_custom_resources':
     'True'
```

should there be a double-quote before the square bracket on the second line?

jamiedemaria:
@sryza  thanks for the catch. meant to ask about this but forgot! all the other values in the telemetry dictionary in `get_stats_from_external_repo` are strings so i converted the sub-dictionaries here to strings. I don't know what happens with the telemetry dictionary after it gets returned and if everything needs to be strings, or if other values are also ok. maybe @yuhan  or @dpeng817 knows?

dpeng817:
A few comments:
- Regarding string-ifying dictionaries - you're correct Jamie in that TelemetryEvent objects essentially get serialized into the log file, before being spun out to our telemetry server. I think this is fine as long as this dictionary ends up getting serialized into the `metadata` field of the telemetry event - we can essentially process metadata however we want in the purina job.
- I _strongly_ suggest we don't keep this written record of dagster-created IO managers - feels like something that will very easily fall out of date (our track record with stuff like this is not good). My suggestion here would be a default-`false` property on the IO manager that indicates whether it is a dagster-created IO manager - and then a test in our CI/CD that ensures that every IO manager imported from the top level of a dagster module returns `True` for that property - then you could just slurp every `dagster_*` package while not worrying about false positives - then you just don't include the path, and dedupe.
- I suggest doing some manual testing here to see what the edge cases are like - and to make sure that telemetry machinery can process everything effectively - should be able to use https://www.notion.so/dagster/Telemetry-Playbook-1291847fe6184c1d83b8189c2adbf94d?pvs=4 - I think essentially you just need to make sure that your added data ends up in the metadata of the log lines.

sryza:
> I strongly suggest we don't keep this written record of dagster-created IO managers - feels like something that will very easily fall out of date (our track record with stuff like this is not good)

From a user research perspective, it's much more useful to understand the particular built-in IO managers that people are using than the fact that they're using at least one built-in IO manager.  I.e. I would be very curious to know how often our DuckDB integration is used.

How exactly would these fall out of date?

dpeng817:
@jamiedemaria that's right - @sryza to be clear I'm suggesting that we programmatically figure out the class names from dagster modules - not that we should completely anonymize IO managers/resources in this telemetry

jamiedemaria:
Now that I'm off support, I'm circling back to this. Seems like the consensus is to add a property that defaults to False for IO managers, then set it to True for all of our IO managers. 

I'm a little unsure how to write a test that ensures that all dagster IO managers have the property set to True. In particular, how to import all IO managers from dagster libraries.  If you have ideas or know of a test that does something similar, lmk (cc @dpeng817 ).

yuhan:
re: tests, i feel it depends on how restrict we want the telemetry enforcement to be.
if we want it to be "opt-in" / loose, we can test just a couple ones that we feel important (snowflake io manager, dbt resource, etc). but if we want to enforce that "every built-in resource should have telemetry set up", we could maybe grab resources from each lib's the top-level export, and "opt-out" when not applicable.

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -18,6 +18,7 @@
     Union,
     cast,
 )
+import dagster
```
jamiedemaria, line 21: auto import from vscode - remove

dpeng817:
Apologies, thought I commented on this already but looks like it didn't go through.

Essentially I think this would amount to a new buildkite step, that imports every library and walks through the top level module imports searching for IO managers / resources. Then just a very simple unit test that, for each of those found IO managers, ensures that it has the property set.

Resources on iterating through module: https://stackoverflow.com/questions/7584418/iterate-the-classes-defined-in-a-module-imported-dynamically
Resources on making a new buildkite step: https://github.com/dagster-io/internal/blob/master/.buildkite/pipeline.py

python_modules/dagster/dagster/_core/host_representation/external_data.py
```diff
@@ -1633,8 +1637,10 @@ def external_resource_data_from_def(
         ),
     ):
         resource_type = _get_class_name(resource_type_def.configurable_resource_cls)
+        dagster_maintained = resource_type_def._dagster_maintained
```
benpankow, line 1640: this can probably be pulled out of the if/else and simplified to

`dagster_maintained = resource_type_def._dagster_maintained._dagster_maintained`

since I think that `resource_type_def` is guaranteed to be a ResourceDefinition (and is already the inner, non-wrapped resource in the case of ResourceWithKeyMapping)

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -1155,12 +1165,13 @@ def nested_resources(
     @cached_method
     def get_resource_definition(self) -> ConfigurableResourceFactoryResourceDefinition:
         return ConfigurableResourceFactoryResourceDefinition(
-            self.__class__,
+            self.resource_cls,
```
jamiedemaria, line 1168: cc @benpankow  I had to change this to `self.resource_cls` so that the class type of the resource thats turned into the PartialResource would be passed through, instead of the `PartialResource` type. I did a quick look through the other places `PartialResource` is used and didn't see anything that would make this change breaking, but want to confirm with you that this is an ok change to make

python_modules/dagster/dagster/_core/definitions/resource_definition.py
```diff
@@ -107,6 +107,8 @@ def __init__(
         if version:
             experimental_arg_warning("version", "ResourceDefinition.__init__")
 
+        self._dagster_maintained = False
```
jamiedemaria, line 110: TODO - add documenting comment that this attribute is modified by the dagster_maintained decorators

python_modules/dagster/dagster/_core/storage/io_manager.py
```diff
@@ -92,7 +92,7 @@ def copy_for_configured(
         description: Optional[str],
         config_schema: CoercableToConfigSchema,
     ) -> "IOManagerDefinition":
-        return IOManagerDefinition(
+        io_def = IOManagerDefinition(
```
jamiedemaria, line 95: todo - directly return this object

jamiedemaria:
@yuhan @benpankow @dpeng817 @sryza  this is ready for review now!

Please let me know if anything looks off with how the attributes are getting piped through to the telemetry system. I was still figuring out how everything worked (and am still not fully confident that I know how all the pieces interact) when writing this and could definitely have made some mistakes or have redundant code

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
benpankow: This seems ok to me 👍

python_modules/dagster/dagster/_core/definitions/resource_definition.py
```diff
@@ -265,6 +275,13 @@ def get_resource_requirements(
             yield ResourceDependencyRequirement(key=resource_key, source_key=source_key)
 
 
+def dagster_maintained_resource(
+    resource_def: ResourceDefinition,
+) -> Callable[[ResourceDefinition], ResourceDefinition]:
+    resource_def._dagster_maintained = True
```
benpankow, line 282: mutating the field here feels not ideal, but this is tricky since `resource_def` could be a user-specified subclass of `ResourceDefinition`, so no way we could reliably reconstruct e.g. a copy, so it's probably fine?

benpankow (commented):
looks good to me! this interface feels pretty good, it isn't user-visible and is pretty non invasive for us

python_modules/dagster/dagster_tests/storage_tests/test_io_manager.py
```diff
@@ -1086,3 +1086,34 @@ def another_asset(an_asset: int) -> int:
         defs.load_asset_value("another_asset", instance=instance)
 
         assert executed["yes"]
```
benpankow, line 1088: It would be neat if we could write a test to gather publicly exported resources from our packages and check that they all have the flag set

python_modules/dagster/dagster/_core/telemetry.py
```diff
@@ -538,6 +545,27 @@ def get_stats_from_external_repo(external_repo: "ExternalRepository") -> Mapping
     }
 
 
+def get_resource_stats(external_resources: Sequence["ExternalResource"]) -> Mapping[str, str]:
```
benpankow, line 548: would be good to put this under test as well

python_modules/dagster/dagster_tests/storage_tests/test_io_manager.py
jamiedemaria: yes! i'm working on this in another branch

yuhan (approved):
amazing! and thanks for the thorough test coverage

PR #14027, opened 2023-05-01
remove asset materializations concept page

sryza:
## Summary & Motivation

This was inspired by this blog post: https://www.flypenguin.de/2023/04/30/getting-started-with-dagster/.

Our docs currently present assets in two different ways – the asset materializations page that's removed here, and the software-defined assets page.  I believe this causes significant confusion for users.

I believe that asset materializations without SDAs are an edge case, and that giving them this much weight unnecessarily causes users to question whether they're following the right approach. I can't think of any use case I've seen that requires this level of dynamic-ness.  The main reason to use asset materializations is if you're [anchored on task-based workflows](https://docs.dagster.io/guides/dagster/how-assets-relate-to-ops-and-graphs#situation-3-youre-anchored-in-task-based-workflows). However, I think the [asset materialization section of the op events page](https://docs.dagster.io/concepts/ops-jobs-graphs/op-events#asset-materializations
) provides enough docs coverage for that use case.


schrockn (changes requested):
Agree with deleting the page. Just looking through the deleted page versus the one you are pointing to, looks like we are removing some valuable content. e.g. partitioning and asset materializations.

sryza:
> Just looking through the deleted page versus the one you are pointing to, looks like we are removing some valuable content. e.g. partitioning and asset materializations.

I added a sentence on this.

I generally think our concept docs are a little bit too comprehensive right now on stuff like this, and it makes them feel overwhelming. If we're documenting capabilities (i.e. parameters) that can be learned by reading apidoc, I think it's generally best to point users to the apidoc.

schrockn (changes requested):
Hmm do we have that in API docs? I don't see it in `AssetMaterialization`'s API doc

sryza:
> Hmm do we have that in API docs? I don't see it in AssetMaterialization's API doc

It's one of the documented parameters: https://docs.dagster.io/_apidocs/ops#dagster.AssetMaterialization. Is your thinking that we should have an example that includes it?

schrockn:
Yeah I was thinking example code. Anyways don't feel too strongly here.

schrockn (approved):

PR #14038, opened 2023-05-01
[docs][tutorial] Adding "managing your own i/o" to the tutorial

tacastillo:
## Summary & Motivation

A good chunk of Dagster users get confused about how to use assets without i/o managers. This section offers using `non_argument_deps` to chain things together. It's essentially the OG tutorial section/assets w/o i/o re-written.

## How I Tested These Changes

Haven't added tests yet in case of pivots during review, but it uses doc snippets and went through black/ruff


docs/content/tutorial/managing-your-own-io.mdx
```diff
@@ -0,0 +1,106 @@
+---
+title: "Tutorial, part seven: Managing your own I/O | Dagster Docs"
+description: Learn how to use I/O managers to save your data.
+---
+
+# Tutorial, part 7: Managing your own I/O
+
+When building the dictionary or word counts, the `most_frequent_words` asset currently uses a static list of stopwords to decide what words to ignore. You’ll notice that despite the length of the list, many unhelpful words and prepositions are still not filtered out. You’d like to ignore those words. To scale this properly, you’ll use a CSV of stopwords found on the internet, use it for `most_frequent_words`, and update it before the refresh to ensure that your list of stopwords is up-to-date.
+
+While doing so, you'll take a step back and learn how to use more of Dagster's versatility without using every feature provided to you. Throughout this tutorial, you've learned multiple Dagster concepts that explain how Dagster can manage functionality on your behalf, ex. I/O managers to read and write from storage.
```
sryza, line 10:
```suggestion
While doing so, you'll take a step back and learn how to use more of Dagster's versatility without using every feature provided to you. Throughout this tutorial, you've learned multiple Dagster concepts that explain how Dagster can manage functionality on your behalf, e.g. I/O managers to read and write from storage.
```

docs/content/tutorial/managing-your-own-io.mdx
```diff
@@ -0,0 +1,106 @@
+---
+title: "Tutorial, part seven: Managing your own I/O | Dagster Docs"
+description: Learn how to use I/O managers to save your data.
+---
+
+# Tutorial, part 7: Managing your own I/O
+
+When building the dictionary or word counts, the `most_frequent_words` asset currently uses a static list of stopwords to decide what words to ignore. You’ll notice that despite the length of the list, many unhelpful words and prepositions are still not filtered out. You’d like to ignore those words. To scale this properly, you’ll use a CSV of stopwords found on the internet, use it for `most_frequent_words`, and update it before the refresh to ensure that your list of stopwords is up-to-date.
```
sryza, line 8: Feel free to push back on this, but my instinct would be to basically put the second paragraph first.  I.e. start by motivating what this section of the tutorial is about before getting into the nitty gritty.

docs/content/tutorial/managing-your-own-io.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+While doing so, you'll take a step back and learn how to use more of Dagster's versatility without using every feature provided to you. Throughout this tutorial, you've learned multiple Dagster concepts that explain how Dagster can manage functionality on your behalf, ex. I/O managers to read and write from storage.
+
+You'll create assets to manage reading and writing from storage on your own. This is useful for multiple purposes, such as:
+
+- Your pipeline manages I/O on its own
```
sryza, line 14: Nit: add period for consistency

docs/content/tutorial/managing-your-own-io.mdx
```diff
@@ -0,0 +1,106 @@
...
+- You want to run a SQL query that creates or updates a table on the database.
+- Adjacent to the above point, your data is too large to fit into memory.
```
sryza, line 16: I'm not sure about this point. Both Spark and Polars allow processing data in a streaming fashion that doesn't load it all into memory at once, and they both work with IO managers.

docs/content/tutorial/managing-your-own-io.mdx
```diff
@@ -0,0 +1,106 @@
...
+- You don't want to use I/O managers.
```
sryza, line 17: It's not clear to me how this is different than "your pipeline manages I/O on its own". Is it something like distinguishing between these two different cases?
- You're using libraries or command-line utilities that read/write to files and tables, and thus you don't have the choice of whether to use I/O managers.
- It's possible to write your pipeline using I/O managers, but you don't want to deal with this Dagster-specific concept. Perhaps because you already have working code, and you don't want to spend time refactoring it.

docs/content/tutorial/managing-your-own-io.mdx
```diff
@@ -0,0 +1,106 @@
...
+```python file=/tutorial/managing_your_own_io/assets.py startafter=start_stopwords_zip endbefore=end_stopwords_zip
+import urllib.request  # Addition: new import at the top of `assets.py`
+
+from dagster import asset
+
+
+@asset
+def stopwords_zip() -> None:
+    urllib.request.urlretrieve(
+        "https://docs.dagster.io/assets/stopwords.zip,data/stopwords.zip",
```
sryza, line 36: should this be two separate strings?

docs/content/tutorial/managing-your-own-io.mdx
```diff
@@ -0,0 +1,106 @@
...
+- Your pipeline manages I/O on its own
+- You want to run a SQL query that creates or updates a table on the database.
+- Adjacent to the above point, your data is too large to fit into memory.
+- You don't want to use I/O managers.
+
+By the end of this section, you'll also have assets that depend on other assets without loading the dependent asset's data into memory.
+
+## Step 1: Creating an asset without an I/O Manager
+
+Many valid situations exist where you would want to avoid using an I/O manager to write an asset back to memory. One is when the asset is already saved in the asset function definition without having to return the data.
```
sryza, line 23: "One is when the asset is already saved in the asset function definition without having to return the data." - this feels a bit circular - i.e. we're not using an I/O manager because we're not using an I/O manager.

I would say something like "we're invoking an API that writes data to disk, rather than returning it to us in memory. Therefore, we don't need an I/O manager to handle writing the data to disk."

docs/content/tutorial/managing-your-own-io.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+These are the only two changes required to have an asset not use an I/O manager.
+
+Refresh your code location, observe the new asset, and materialize it as you have with the previous assets. You can confirm that this works by finding the `stopwords.zip` in the `/data` directory made in the **Saving your data** section of the tutorial.
+
+## Step 2: Depend on an asset without loading it
+
+You’ll now unzip the zip file asset you made and extract the CSV from it. This CSV will be a new asset that depends on the zip file asset you downloaded.
+
+By default, an asset will load its dependencies into memory when it materializes. However, there are situations where you would want to avoid doing this. As previously mentioned, if your data is too large to fit into memory, or if you didn’t use an I/O manager for the dependency, you would not want to try loading the dependency.
```
sryza, line 53: I think the main reason that we want to avoid doing this is that we're invoking a library method – `extractall` – that operates on data that's already on disk. So there's no need for us to read/write to disk in an IO manager.

sryza (commented):
I don't particularly think we should do this, but better to consider it now then later: is there some chance we should start the tutorial without IO managers and then have them refactor their code to use IO managers after, to see the benefit?

docs/content/tutorial/managing-your-own-io.mdx
tacastillo: Hmm, you're right. I'm in between whether streaming/lazyloading are edge cases or common enough that.

How about:

```suggestion
- Adjacent to the above point, your I/O manager is batch-oriented and the data loaded would not fit into memory.
```

?

docs/content/tutorial/managing-your-own-io.mdx
tacastillo: Astute of you. I was going back and forth on this during drafting, so I'm glad that the value of one over the other wasn't just in my head.

tacastillo:
> I don't particularly think we should do this, but better to consider it now then later: is there some chance we should start the tutorial without IO managers and then have them refactor their code to use IO managers after, to see the benefit?

@sryza Definitely a chance, but the lift could use some more data in order to help make a more informed call on the proper sequence of concepts. So push, collect feedback, and iterate?

docs/content/tutorial/managing-your-own-io.mdx
sryza: I think whether something is batch-oriented has less to do with the IO manager and more to do with the compute framework (Pandas, Spark, etc). But I think we're already covering that when we talk about what tools they chose. So my instinct would be to leave this out.

docs/content/tutorial/managing-your-own-io.mdx
```diff
@@ -0,0 +1,106 @@
+---
+title: "Tutorial, part seven: Managing your own I/O | Dagster Docs"
+description: Learn how to use I/O managers to save your data.
+---
+
+# Tutorial, part 7: Managing your own I/O
+
+In this section, you'll take a step back and learn how to use more of Dagster's versatility without using every feature provided to you. Throughout this tutorial, you've learned multiple Dagster concepts that explain how Dagster can manage functionality on your behalf, ex. I/O managers to read and write from storage.
+
+When building the dictionary or word counts, the `most_frequent_words` asset currently uses a static list of stopwords to decide what words to ignore. You’ll notice that despite the length of the list, many unhelpful words and prepositions are still not filtered out. You’d like to ignore those words. To scale this properly, you’ll use a CSV of stopwords found on the internet, use it for `most_frequent_words`, and update it before the refresh to ensure that your list of stopwords is up-to-date.
```
sryza, line 10: Nitpick: "stop words"

docs/content/tutorial/managing-your-own-io.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+These are the only two changes required to have an asset not use an I/O manager.
+
+Refresh your code location, observe the new asset, and materialize it as you have with the previous assets. You can confirm that this works by finding the `stopwords.zip` in the `/data` directory made in the **Saving your data** section of the tutorial.
+
+## Step 2: Depend on an asset without loading it
+
+You’ll now unzip the zip file asset you made and extract the CSV from it. This CSV will be a new asset that depends on the zip file asset you downloaded.
+
+By default, an asset will load its dependencies into memory when it materializes. However, there are situations where you would want to avoid doing this. If you're making changes directly on data in storage, such as unzipping a zip file, you don't need to load the data into memory.
```
sryza, line 53:
```suggestion
By default, an asset will load its dependencies into memory when it materializes. However, there are situations where you would want to avoid doing this. If you're operating directly on data in storage, such as unzipping a zip file, you don't need to load the data into memory.
```

docs/content/tutorial/managing-your-own-io.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+To tell an asset not to load a dependency into memory when materializing, you can use the `@asset` decorator’s `non_argument_deps` parameter to establish the dependency instead of passing it in as an argument of the asset’s function definition.
```
sryza, line 55:
```suggestion
To tell an asset not to load a dependency into memory when materializing, you can use the `@asset` decorator’s `non_argument_deps` parameter to establish the dependency, instead of including it as an argument of the asset’s function definition.
```

docs/content/tutorial/managing-your-own-io.mdx
```diff
@@ -0,0 +1,106 @@
...
+import zipfile  # Note: remember to add imports to the top of the file
+
+
+@asset(non_argument_deps={"stopwords_zip"})
+def stopwords_csv() -> None:
+    with zipfile.ZipFile("data/stopwords.zip", "r") as zip_ref:
+        zip_ref.extractall("data")
+```
+
+Reload your code location, observe the new `stopwords_csv` asset, and notice how it depends on the `stopwords_zip` asset.
```
sryza, line 69:
```suggestion
In the UI, reload your code location, observe the new `stopwords_csv` asset, and notice how it depends on the `stopwords_zip` asset.
```

docs/content/tutorial/managing-your-own-io.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+## Step 3: Combining assets with and without I/O managers
+
+Now that you’ve made two assets that don’t use I/O managers, you’ll connect the `stopwords_csv` asset to your `most_frequent_words` asset to use a more comprehensive list of stopwords.
+
+When defining an asset, you can combine assets that use I/O managers with assets that don’t. Update your `most_frequent_words` asset by replacing the static list of stopwords. Replace it with `stopwords_csv` by updating your code to match the below:
```
sryza, line 75:
```suggestion
When defining an asset, you can combine assets that use I/O managers with assets that don’t. Update your `most_frequent_words` asset by replacing the static list of stopwords. Replace it with `stopwords_csv` by updating your code to match what's below:
```

sryza (commented):
I left a few small remaining comments. Also, the code snippets would benefit from a test. 
 Here's the test from the old tutorial section, if that helps: https://github.com/dagster-io/dagster/blob/release-1.1.0/examples/docs_snippets/docs_snippets_tests/guides_tests/asset_tutorial_tests/test_non_argument_deps.py. `AssetGroup` is deprecated, but you can just use `materialize` directly on the assets instead.

Aside from those few things, this looks great.

docs/content/tutorial/managing-your-own-io.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+
+# ...
+
+# Keep the rest of the function the same
+```
+
+To validate that this worked, materialize your entire asset graph to see `most_frequent_words` depend on the `stopwords_csv` asset
+
+## Next Steps
```
erinkcochran87, line 100:
```suggestion
---

## Next steps
```

docs/content/tutorial/saving-your-data.mdx
```diff
@@ -169,14 +169,10 @@ Despite the changes in where the data is stored, the developer experience hasn
...
 
-Assets don’t need to use I/O managers. Sometimes it’s impossible to separate business logic from I/O, or you’ll never want to load the data into memory. In situations like these, you’re able to not return any data, and dependencies can be defined without loading data in the dependent asset’s function. Refer to the [Assets without I/O](/guides/dagster/non-argument-deps) for more information.
-
 We also briefly mentioned resources. A property of resources is that they can be configured depending on the use case, such as using DuckDB as your data warehouse during development and Databricks during production.
 
-Now that you've completed the tutorial, find out what else you can do in with our [Next steps](/tutorial/next-steps).
+Assets don’t need to use I/O managers. Sometimes it’s impossible to separate business logic from I/O, or you’ll never want to load the data into memory. In situations like these, you’re able to not return any data, and dependencies can be defined without loading data in the dependent asset’s function. In the next section, you'll learn how to create assets and dependencies without using I/O managers.
```
erinkcochran87, line 178:
```suggestion
Assets don’t need to use I/O managers. Sometimes it’s impossible to separate business logic from I/O, or you’ll never want to load the data into memory. In situations like these, you’re able to not return any data, and dependencies can be defined without loading data in the dependent asset’s function. [In the next section](/tutorial/managing-your-own-io), you'll learn how to create assets and dependencies without using I/O managers.
```

Adding a link to the next section

docs/content/_navigation.json
```diff
@@ -56,6 +56,10 @@
         "title": "Saving your data",
         "path": "/tutorial/saving-your-data"
       },
+      {
+        "title": "Managing your own I/O",
```
erinkcochran87, line 60: This will be `Part 7: Managing your own I/O` once you pull in master

erinkcochran87 (commented):
Could you pull in the latest from master? I think there are some changes to make to the `fs_io_manager` (Ben recently replaced a bunch of references to this elsewhere) and the way the tutorial sections are labeled in the sidenav.

sryza (approved):
Approving because the test looks good! I haven't looked at the content (which I was already satisfied with) since my last review.

Still worth paying attention to any outstanding feedback from @erinkcochran87 of course.

PR #14042, opened 2023-05-02
Expand config on launchpad load for jobs without preset definitions

benpankow:
## Summary

When loading the launchpad for a new pipeline, presents the user with the default, fully expanded config for jobs without any mode preset.

## Test Plan

Tested locally.

PR #14040, opened 2023-05-01
Add 'scaffold all default config' button to launchpad

benpankow:
## Summary

Introduces a new button to the launchpad which allows users to expand all default config values.

Resolves #11828 

Screen Shot 2023-05-01 at 4 31 41 PM

Screen Shot 2023-05-01 at 4 31 47 PM

One possible improvement to stack on this would be checkboxes in the dialog to toggle on/off which defaults to expand (ops, resources, execution, logging).

## Test Plan

Tested locally.

PR #14039, opened 2023-05-01
Snapshot default run config & expose via GraphQL

benpankow:
## Summary

Update of #12282. Serializes the default run config for jobs, making it available via GraphQL to the frontend. Used in stacked PR to enable "expand all defaults" button in launchpad.

## Test Plan

Snapshot test; tested manually.

PR #14044, opened 2023-05-02
[pythonic resources] Support setting fixed/secret values at Definitions-time

benpankow:
## Summary

Introduces a pattern for setting secret values on a ConfigurableResource. The new `FixedConfig` function wraps a config value (or env var) and ensures that the value will not show up in the resources UI and cannot be overridden in the launchpad:

```python
    username: str
    password: str

# retrieve password from secret store etc
password = ...

defs = Definitions(
    ...
    resources={
        "my_resource": MyResource(
            username="foo",
            password=FixedConfig(password),
        )
    )
)
```

## Test Plan

A few unit tests, snapshot test ensuring resource snap doesn't include the value.

### DIFF
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -1852,5 +1852,8 @@ class _FixedConfig:
         return self._value
 
 
-def FixedConfig(value: Any) -> Any:
-    return _FixedConfig(value)
+C = TypeVar("C")
+
+
+def FixedConfig(value: C) -> C:
+    return _FixedConfig(value)  # type: ignore
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -314,9 +314,7 @@ def _config_value_to_dict_representation(field: Optional[ModelField], value: Any
     from dagster._config.field_utils import EnvVar, IntEnvVar
 
     if isinstance(value, dict):
-        return {
-            k: _config_value_to_dict_representation(None, v) for k, v in value.items()
-        }
+        return {k: _config_value_to_dict_representation(None, v) for k, v in value.items()}
     elif isinstance(value, list):
         return [_config_value_to_dict_representation(None, v) for v in value]
     elif isinstance(value, EnvVar):
@@ -333,9 +331,7 @@ def _config_value_to_dict_representation(field: Optional[ModelField], value: Any
                 ).items()
             }
         else:
-            return {
-                k: v for k, v in value._convert_to_config_dictionary().items()
-            }
+            return {k: v for k, v in value._convert_to_config_dictionary().items()}  # noqa: SLF001
     elif isinstance(value, Enum):
         return value.name
 
@@ -382,9 +378,7 @@ class PermissiveConfig(Config):
 
 
 # This is from https://github.com/dagster-io/dagster/pull/11470
-def _apply_defaults_to_schema_field(
-    field: Field, additional_default_values: Any
-) -> Field:
+def _apply_defaults_to_schema_field(field: Field, additional_default_values: Any) -> Field:
     # This work by validating the top-level config and then
     # just setting it at that top-level field. Config fields
     # can actually take nested values so we only need to set it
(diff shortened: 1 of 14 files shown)


benpankow:
* master
  * **PR #14039** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14039" target="_blank">Graphite</a> 
    * **PR #14040** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14040" target="_blank">Graphite</a> 
      * **PR #14042** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14042" target="_blank">Graphite</a> 
        * **PR #14044** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14044" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14044?utm_source=stack-comment).

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -1699,3 +1741,18 @@ def validate_resource_annotated_function(fn) -> None:
...
+        self._value = value
+
+    def get_value(self) -> Any:
+        return self._value
+
+
+C = TypeVar("C")
+
+
+def FixedConfig(value: C) -> C:
```
benpankow, line 1757: This is pretty hacky and I'm pretty hesitant to ship something like this. There may be a better way to handle this (or just return `Any`) that will not cause type issues.

schrockn:
Is there a solution that does not involve us adding another concept? Could we instead direct users to override init and set a `PrivateAttr` or something?

schrockn (changes requested):
Asking question about whether we need this at all. See comment.

benpankow:
> Is there a solution that does not involve us adding another concept? Could we instead direct users to override init and set a `PrivateAttr` or something?

I think we could support this on the resource definition side by allowing users to somehow annotate a parameter as secret.

I do think this PR handles a slightly different case - because the fixed/secret values are specified by the passed config rather than the resource definition I think this is a bit more broadly useful - e.g. defining a SnowflakeResource which has fixed config that can't be changed in the launchpad except for the schema. 


```python

defs = Definitions(
    ...
    resources={
        "snowflake": SnowflakeResource(
            schema="my_default_overridable_schema",
            username=FixedConfig("user"),
            password=FixedConfig(password),
            account=FixedConfig("na1234"),
        )
    )
)

**Comment by schrockn** [comment]
I see. `FixedConfig` definitely seems a bit odd to me. And I don't think we should use the word "secret" here anywhere because that is subtley. Maybe "HiddenConfig" or "ErasedConfig"

**Comment by benpankow** [comment]
Here `Fixed` was meant to indicate that the config can't be changed or overridden in the launchpad. Maybe `ImmutableConfig`?

Stepping back I think there are a few possible patterns we could use to address (secret/fixed config), which are related but separate concerns:

1. Wrapping config value in an indicator, which is encompassed in this PR. The benefits of this approach is that we can decide on a subset of values to fix & hide form the UI. This does bunch together "hiding config" and "fixing config" when a user might only want to do one or the other. There's a world in which we ship two indicators.

  ```python
    username: str
    password: str = Field(is_secret=True)
  ```

2. Applying a secret indicator to the resource definition itself, as discussed above. This is useful for cases where we know particular config values will be secret and want to systemically hide them in the UI. This arises in #14114. This does not handle the case of users wanting to make secret values fixed/unable to be overridden.

  ```python
    username: str
    password: str
  
  MyResource(username=ImmutableConfig("foo"), password=HiddenConfig("bar"))
  ```

4. Using a special method to configure fixed or secret config, rather than the constructor. This loses out on the ergonomics of the typed constructor for our config/resource classes.
  ```python
    username: str
    password: str
  
  # one way to model the method
  MyResource(...).with_immutable("password")```

There might be some combination of the above that feels best (e.g. wrapper for fixed, definition-sided indicator for secrets?)

**Comment by schrockn** [comment]
```
class AResource2(ConfigurableResource):
    number: int
    _private_thing: str = PrivateAttr("private")
a_resource_2 = AResource2(_private_thing="foo", number=1)
print(a_resource_2._private_thing)
print(a_resource_2._convert_to_config_dictionary())
```

Let's see if we use vanilla `PrivateAttr`

PR #14045, opened 2023-05-02
Only run dependabot on certain folders

gibsondan:
Summary:
Running on pyright is leading to a bunch of false positives with transitive dependencies of airflow. Since dependabot doesn't let you exclude folders, enumerate the different top-level folders we do want to run it on.

I also updated the reviewers, although it didn't seem that it was respecting the old setting to make it sandy - hopefully this file is being used at all? Tricky to test without waiting for the next dependabot update.


rexledesma (approved):
Looks like there's validation in the GitHub action that happens after this is merged: https://github.com/dependabot/dependabot-core/issues/4605#issuecomment-1021155926

I would check that after we land this just to make sure we're not missing anything. We could consider landing a validator, but feels like overkill since we don't update this configuration often.

gibsondan:
Do you have any ideas for why the sryza reviewer setting that was set
before was being ignored?

rexledesma (changes requested):
I think it makes sense here to have security updates on the following packages/applications:

- Dagster
- Dagster GraphQL
- Dagit (including `js_modules/dagit`)

gibsondan:
Hmmm so these were firing on other folders already then despite the existing rules being misconfigured - does that mean there's no way to exclude /pyright/master from dependabot altogether? That's the main goal here

rexledesma:
> Hmmm so these were firing on other folders already then despite the existing rules being misconfigured - does that mean there's no way to exclude /pyright/master from dependabot altogether? That's the main goal here

Sounds like the answer to that is no: https://github.com/dependabot/dependabot-core/issues/4364

rexledesma:
Looks like no, which is a bane in monorepo setups: https://github.com/dependabot/dependabot-core/issues/4364

rexledesma (approved):
🙏 thoughts and prayers! 🙏

smackesey:
After getting pinged on multiple pyright dependabot PRs and seeing that a zillion of them are open, I decided to look at this, and now I'm confused. IIUC our `dependabot.yaml` whitelists a few directories we want to watch, and `pyright/*` isn't part of this list. Anyone know why we are still getting alerts for it?

rexledesma:
@smackesey From our internal investigation, summarized by @gibsondan:

> not sure where to put this so documenting it here. These Github vulnerabilities are usually in our pyright folder and are basically false positives (the only reason that folder is checked in is for development and it contains a ton of transitive dependencies - if airflow has some vulnerability and we happen to depend on airflow in the dagster-airflow, that's not really worth of our attention)
>
> Also dependabot doesn't seem to have a way to exclude specific folders(!) despite me and `@rex` scouring the internet looking for a way to do it. So what i've been doing is checking to see if the dependabox issue is in the pyright folder, and it if is, marking it as irrelevant

Unfortunately, seems like dependabot configuration is just not adequate for a monorepo setup.

PR #14050, opened 2023-05-02
[docs] - Sidenav improvements, part 3

erinkcochran87:
## Summary & Motivation

This PR builds on #12823 and #13558, making some changes and improvements to the sidenav:

- Standardizes capitalization (sentence casing) where needed
- Corrects integration name formatting (ex: `Github` > `GitHub`) where needed
- Moves the **Community** section under **About**
- Updates the tutorial:
  - In the sidenav, adds `Part X:` to each part of the tutorial, to make it clear the progression is sequential
  - Re-names the tutorial to `Building your first pipeline using Software-defined Assets`. While this doesn't get `Pipeline` into the actual sidenav, it is now in an H1/page title and will be more prominent in search results for this term.
  
     I made this pivot after adding the `Part X:` bits to the sidenav. I think removing `Tutorial` from the sidenav but retaining these bits would be an odd experience.
- Adds `Part X:` to the sections for the dbt tutorial to keep things consistent / clear

## How I Tested These Changes

👀 , local

### DIFF
--- a/docs/content/_navigation.json
+++ b/docs/content/_navigation.json
@@ -1,6 +1,6 @@
 [
   {
-    "title": "Getting Started",
+    "title": "Getting started",
     "icon": "Flag",
     "path": "/getting-started",
     "isDefaultOpen": true,
@@ -72,11 +72,11 @@
         "path": "/concepts/assets/software-defined-assets",
         "children": [
           {
-            "title": "Software-Defined Assets",
+            "title": "Software-defined Assets",
             "path": "/concepts/assets/software-defined-assets"
           },
           {
-            "title": "Asset Materializations",
+            "title": "Asset materializations",
             "path": "/concepts/assets/asset-materializations"
           },
           {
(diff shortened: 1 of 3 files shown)


docs/content/_navigation.json
```diff
@@ -572,7 +572,7 @@
             "path": "/integrations/airflow/from-airflow-to-dagster"
           },
           {
-            "title": "Airflow Migration Guide",
+            "title": "Migrating from Airflow",
```
tacastillo, line 575: Didn't realize how much I'd like this.

tacastillo (commented):
Any particular reason why some of the sections didn't get sentence-cased?

image

Other than that, these are all reasonable, consistent and sane changes. just lmk about above and I'll hit approve afterwards.

erinkcochran87:
> Other than that, these are all reasonable, consistent and sane changes. just lmk about above and I'll hit approve afterwards.

@tacastillo Good question! Some of these I probably just missed, like `Metadata & Tags`

As for `Graphs` - tbh I'm not sure if we should be capitalizing those, now that I think about it. I'll ask Fraser and report back

tacastillo (approved):
let's do iiiiiit

Issue #14057, opened 2023-05-02
dbt test_python_interleaving[pythonic] is flaky


jmsanders:
This has failed the last few builds:
https://buildkite.com/dagster/dagster/builds/57813#0187dddd-c84e-4e52-863f-334e9d3654c6  
https://buildkite.com/dagster/dagster/builds/57778#0187dd32-c3f0-4b97-a1cb-5929e4ee044e

https://github.com/dagster-io/dagster/pull/14052

sryza:
@OwenKephart mind taking a look?

OwenKephart:
looking

OwenKephart:
My best guess here is that this is actually a bug on the dbt-core or dbt-duckdb side of things, but it's unclear why we haven't seen this before. This error purely happens in the test fixture which calls `dbt seed`, so no Dagster code is called before or after this error. I was not able to replicate this on my laptop after 10 attempts, so it's also possible that this only happens in certain environments.

The closest I could get to someone else encountering this error was: /Users/owen/src/dagster/python_modules/libraries/dagster-dbt/dagster_dbt_tests/test_asset_defs.py, which seems to indicate that this can happen when you import numpy at the same time in multiple different threads.

In this specific test, we have a project with two seeds, rather than one seed as in the default implementation. This would cause two threads to be spun up, potentially resulting in this race condition sort of behavior. I just put up a PR that sets the concurrency of this seed process to 1, to eliminate that possibility:

rexledesma:
Resolved by https://github.com/dagster-io/dagster/pull/14058.

PR #14058, opened 2023-05-02
fix flaky dbt test

OwenKephart:
## Summary & Motivation

https://github.com/dagster-io/dagster/issues/14057


jmsanders (approved):
Seems reasonable to me. And if it regresses again, we'll keep digging.

PR #14063, opened 2023-05-03
Clean up executor with contextmanager

jmsanders:
Sometimes, this test suite completes but hangs until Buildkite kills the job after its 25 minute timeout.

These tests are parameterized to run both synchronously and with a threadpool. My suspicion is that sometimes, we have a leaky thread that prevents pytest from "finishing."

This splits the threadpool tests out from the synchronous tests and outputs more verbose PyTest logging in hopes that we can pinpoint the cause of these occasional timeouts.


gibsondan:
if we think that's it i bet there's a good chance that wrapping this object in a with() would also help - is that worth trying?

https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster_tests/daemon_sensor_tests/test_sensor_run.py#L855-L859

gibsondan (approved):
do we still need the separate test suites after this change now? seems like they're both about 6 minutes now so not the critical tentpole

jmsanders:
The split was more to isolate the root cause than it was to circumvent timeouts (since they were already succeeding in ~12 minutes). I'll re-combine them for now and we can always split them back out later if our fix is insufficient.

Issue #13059, opened 2023-03-21
Add retries with backoff to work around etcdserver timeouts in k8s executor


markfickett:
### Dagster version

Agent v1.1.20

### What's the issue?

We are using the step-per-pod k8s executor in a hybrid cloud deployment with an EKS cluster. A few times a week (for a 4x-daily job that lasts about 4.5h) we are seeing an etcdserver timeout. AWS support says that this is because of etcdserver database defragmentation processes, which last about 20s, and runs every 6h (adjusted to 24h when I asked for other changes).

Context on [slack thread one](https://dagster.slack.com/archives/C014N0PK37E/p1677770120520539) and [two](https://dagster.slack.com/archives/C014N0PK37E/p1678979927825229).

### What did you expect to happen?

I would like the Dagster job to execute successfully even when there are underlying transient k8s errors.

However, when a k8s job can't be scheduled due to this timeout, the full Dagster job fails.

### Deployment type

Dagster Cloud

### Deployment details

Hybrid Dagster Cloud on AWS EKS, cluster with two nodes.

### Additional information

Here's one example error:

```
kubernetes.client.exceptions.ApiException: (500)
Reason: Internal Server ErrorHTTP response headers: HTTPHeaderDict({'Audit-Id': 'c27783d1-149c-4c14-aa13-39b704e351c3', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'X-Kubernetes-Pf-Flowschema-Uid': 'be963a72-9c7c-4ba6-8b96-08dd1b67b88b', 'X-Kubernetes-Pf-Prioritylevel-Uid': 'd762e8f5-544e-4b13-8031-f7f6909643ff', 'Date': 'Thu, 02 Mar 2023 03:34:23 GMT', 'Content-Length': '122'})HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"etcdserver: request timed out","code":500}
  File "/usr/local/lib/python3.10/site-packages/dagster/_core/execution/api.py", line 991, in pipeline_execution_iterator
    for event in pipeline_context.executor.execute(pipeline_context, execution_plan):
  File "/usr/local/lib/python3.10/site-packages/dagster/_core/executor/step_delegating/step_delegating_executor.py", line 305, in execute
    list(
  File "/usr/local/lib/python3.10/site-packages/dagster_k8s/executor.py", line 260, in launch_step
    self._api_client.batch_api.create_namespaced_job(
[...]

PR #14081, opened 2023-05-04
Add retries to k8s create_namespaced_job

johannkm:
Per https://github.com/dagster-io/dagster/issues/13059, this api call can result in transient 500s that AWS claims will be avoided if you retry at most 30 seconds.

The retry function is limited to transient error codes (including 500), and will try, wait 15 seconds, try, wait another 15 seconds, try a final time.

In the worst case where we use all these tries, the method will be blocking for 30 seconds. This feels bad but may actually be ok: the two places it's called is from the run queue daemon and run workers. Blocking will stop them from launching other runs/steps, but if the api requests are failing then that's fine. It'd only be problematic if there was something about the specific job causing the failure, and usually that wouldn't be a transient error code.

### DIFF
--- a/python_modules/libraries/dagster-k8s/dagster_k8s/client.py
+++ b/python_modules/libraries/dagster-k8s/dagster_k8s/client.py
@@ -11,7 +11,7 @@ from dagster import (
     _check as check,
 )
 from dagster._core.storage.dagster_run import DagsterRunStatus
-from kubernetes.client.models import V1JobStatus
+from kubernetes.client.models import V1Job, V1JobStatus
 
 try:
     from kubernetes.client.models import EventsV1Event  # noqa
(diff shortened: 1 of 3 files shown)


johannkm:
* master
  * **PR #14081** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14081" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14081?utm_source=stack-comment).

python_modules/libraries/dagster-k8s/dagster_k8s/client.py
```diff
@@ -749,3 +749,10 @@ def get_pod_debug_info(self, pod_name, namespace, container_name: Optional[str]
             + (f"\n\n{log_str}" if log_str else "")
             + f"\n\n{warning_str}"
         )
+
+    def create_namespaced_job_with_retries(self, body: V1Job, namespace: str):
+        k8s_api_retry(
+            lambda: self.batch_api.create_namespaced_job(body=body, namespace=namespace),
+            max_retries=3,
+            timeout=15.0,
```
johannkm, line 757: it'd be nice to make this configurable, I wonder where that should live though

gibsondan: how about as arguments to this function with defaults set?

gibsondan: 15 seconds pretty high for an initial retry to me - should we get fancy and throw some exponential backoff in here?

gibsondan (changes requested):
thanks for picking this up

python_modules/libraries/dagster-k8s/dagster_k8s/client.py
johannkm: reduced to 10 seconds. I miscounted- the default that we use for all the other apis (4 attempts, 10 seconds pause) will get us the 30 seconds we need here

exponential backoff sounds good, I don't think it will make much difference in this case though

gibsondan (approved):

PR #14083, opened 2023-05-04
[SqlPollingEventWatcher] dont error on unwatch after dispose

alangenfeld:
if `end_watch` is called after the context manager exit we currently get
`'SqlPollingEventWatcher' object has no attribute '_run_id_to_watcher_dict'"`

since we delete the property. Just reset it to an empty dict on dispose instead.

## How I Tested These Changes

added test that was previously failing

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/event_log/polling_event_watcher.py
+++ b/python_modules/dagster/dagster/_core/storage/event_log/polling_event_watcher.py
@@ -82,7 +82,7 @@ class SqlPollingEventWatcher:
                         watcher_thread.should_thread_exit.set()
                 for run_id in self._run_id_to_watcher_dict:
                     self._run_id_to_watcher_dict[run_id].join()
-                del self._run_id_to_watcher_dict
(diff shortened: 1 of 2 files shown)


gibsondan (dismissed):
This seems like a reasonable change to me but the situation you described (calling a method after it has exited/been disposed) seems bad - is there a callsite problem as well that we need to fix, or are there extenuating circumstances that make that a reasonable thing to do?

alangenfeld:
> are there extenuating circumstances that make that a reasonable thing to do?

We call unwatch in this `finally` block 

https://github.com/dagster-io/dagster/blame/master/python_modules/dagster-graphql/dagster_graphql/implementation/execution/__init__.py#L256-L257

which apparently is firing after the middleware closes that exits the instance context manger - likely related to (mis)use of low level async task primitives in the server impl

https://github.com/dagster-io/dagster/blame/master/python_modules/dagit/dagit/graphql.py

gibsondan (approved):
gotcha - thems the breaks i guess

PR #14084, opened 2023-05-04
Add a timeout for runs stuck in canceling

gibsondan:
Summary:
Just like for STARTING. only question I have is whether or not the point query on the event log to determine the CANCELING time is likely to run into perf issues, or if the fact that it's filtering on a specific event type will help.

Test Plan: BK, simulate a real hanging cancel in `dagster dev` by commenting out the line that cancels the run and waiting 3 minutes

### DIFF
--- a/helm/dagster/values.yaml
+++ b/helm/dagster/values.yaml
@@ -1093,6 +1093,8 @@ dagsterDaemon:
     enabled: true
     # Timeout for runs to start (avoids runs hanging in STARTED)
     startTimeoutSeconds: 300
+    # Timeout for runs to cancel (avoids runs hanging in CANCELING)
+    # cancelTimeoutSeconds: 300
     # How often to check on in progress runs
     pollIntervalSeconds: 120
     # [Experimental] Max number of times to attempt to resume a run with a new run worker instead
--- a/python_modules/dagster/dagster/_daemon/monitoring/monitoring_daemon.py
+++ b/python_modules/dagster/dagster/_daemon/monitoring/monitoring_daemon.py
@@ -19,7 +19,7 @@ from dagster._core.storage.dagster_run import (
 )
 from dagster._core.storage.tags import MAX_RUNTIME_SECONDS_TAG
 from dagster._core.workspace.context import IWorkspace, IWorkspaceProcessContext
-from dagster._utils import DebugCrashFlags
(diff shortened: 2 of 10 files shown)


helm/dagster/values.yaml
```diff
@@ -1093,6 +1093,8 @@ dagsterDaemon:
     enabled: true
     # Timeout for runs to start (avoids runs hanging in STARTED)
     startTimeoutSeconds: 300
+    # Timeout for runs to cancel (avoids runs hanging in CANCELING)
+    # startTimeoutSeconds: 300
```
alangenfeld, line 1097: wrong val

python_modules/dagster/dagster/_daemon/monitoring/monitoring_daemon.py
```diff
@@ -55,6 +55,47 @@ def monitor_starting_run(
         instance.report_run_failed(run, msg)
 
 
+def monitor_canceling_run(
+    instance: DagsterInstance, run_record: RunRecord, logger: logging.Logger
+) -> None:
+    run = run_record.dagster_run
```
alangenfeld, line 61: i'm curious in what cases the `update_timestamp` on the run record doesnt match what we calculate via the event lookup

python_modules/dagster/dagster/_daemon/monitoring/monitoring_daemon.py
```diff
@@ -137,12 +178,18 @@ def execute_monitoring_iteration(
...
                 monitor_starting_run(instance, run_record, logger)
             elif run_record.dagster_run.status == DagsterRunStatus.STARTED:
                 monitor_started_run(instance, workspace, run_record, logger)
-            elif run_record.dagster_run.status == DagsterRunStatus.CANCELING:
-                # TODO: implement canceling timeouts
+            elif (
+                instance.run_monitoring_cancel_timeout_seconds > 0
+                and run_record.dagster_run.status == DagsterRunStatus.CANCELING
+            ):
+                monitor_canceling_run(instance, run_record, logger)
```
alangenfeld, line 192: Seems we could restructure this to not recalculate "when X entered state Y" every tick and only when its first observed. Looks like we suggest a high poll interval so not clearly a high priority.

alangenfeld (approved):

PR #14097, opened 2023-05-04
RFC: Add a new LambdaRepositoryData that lets you reload definitions from dagit w/o any caching

gibsondan:
Summary:
This is the start of an idea of how we might address the frequent ask to make reloading a code location in dagit do a full reload of all definitions, without any caching and without needing to restart the Python process (to handle the case where your jobs and ops are dynamically generated, and something has changed). This would be for advanced users only who know what they're doing.

The idea would be:
- A new flag on the repo that indicates whether the definitions should be cached (exact nomenclature unclear)
- If that flag is set, the RepositoryData passes through the repository data function down into the RepositoryData object instead of evaluating it
- Triggering a reload could call a new gRPC method that re-runs the lambda and re-generates the definitions, but is a no-op in the common case where this flag is not set (right now I jammed it into the LoadRepositories call, which actually mostly works since that's called on every reload)

An alternative would be to make reloading in Dagit trigger a full restart of the Python process, but that would require significant deployment architecture changes (and a nice thing about this solution is you don't incur process or task startup times)

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/repository_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/repository_decorator.py
@@ -28,6 +28,7 @@ from ..partitioned_schedule import UnresolvedPartitionedAssetScheduleDefinition
 from ..repository_definition import (
     VALID_REPOSITORY_DATA_DICT_KEYS,
     CachingRepositoryData,
+    LambdaRepositoryData,
     PendingRepositoryDefinition,
     PendingRepositoryListDefinition,
     RepositoryData,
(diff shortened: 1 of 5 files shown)


python_modules/dagster/dagster/_core/definitions/decorators/repository_decorator.py
```diff
@@ -52,6 +53,99 @@ def _flatten(items: Iterable[Union[T, List[T]]]) -> Iterator[T]:
             yield x
 
 
+def create_repository_data_from_fn(
```
gibsondan, line 56: pulled more or less wholesale from the existing repo creation logic (so that it can be deferred later in LambdaRepositoryData)

can_defer_repository_data is new - for now i disabled that functionality for these

gibsondan:
Issue that this would help with: https://github.com/dagster-io/dagster/issues/12581

schrockn:
I find this pretty terrifying and I think it will be the source of lots of difficult-to-debug bugs as we have no control over code the definitions will invoke.

And are we even sure this works?  The function that collects all the definitions might get re-invoked, but does the python module correctly identify that it needs to reload code?

gibsondan:
Having sat with this for a few hours i think it is probably not worth the additional complexity - you'd have to substantially change your code to take advantage of it (e.g. declare your jobs within the repository function if they are likely to change) and most people who want this probably want to just reload the python module entirely (which almost certainly requires restarting the code process)

gibsondan:
@schrockn i'm not sure what you mean by 'does the python module correctly identify that it needs to reload code?' - no code is re-loaded, but the repository function is re-invoked and a new set of definitions is available to anything that calls methods on the RepositoryDefinition object.

schrockn:
> @schrockn i'm not sure what you mean by 'does the python module correctly identify that it needs to reload code?' - no code is re-loaded, but the repository function is re-invoked and a new set of definitions is available to anything that calls methods on the RepositoryDefinition object.

@gibsondan A lot of times definitions at created at python module import time (e.g. a top-level `@asset` decorator does this) so the repository just returns a reference to that rather than reconstructing it. The only way it would be reconstructed is if the python module were reloaded.

```python
# foo.py

from dagster import asset

@asset
def foo_asset():
   ...
```

```python
# bar.py

from dagster import Definitions
# do we know caching behavior of the below at all times?
from .foo import foo_asset

defs = Definitions([foo_asset])
```

gibsondan:
Right - so the pattern being tried out here would not work with Definitions (which is probably reason enough to not pursue it). You would need to a) define an old-style repository decorator function b) set it up so that everything you need to reload is defined within that repository decorator function, since that's the only thing that re-executes.

So i think this is a bust - a better fix probably involves either moving a) OSS to an agent architecture so that we have a separate process that is responsible for spinning up and reloading code servers, OR b) instead of having `dagster api grpc` spin up code in-process, spin it up in a subprocess that can be reloaded without restarting the whole process

gibsondan:
@alangenfeld return of the Cli api?

I had been imagining spinning up a second gRPC server as a subprocess and
then proxying requests through, but that could work too.

schrockn:
Yeah basically return of the cli api. I think critical difference is that all comms would be between stdin/stdout with streaming json rather than jamming into shell args.

alangenfeld:
> all comms would be between stdin/stdout with streaming json rather than jamming into shell args

or just use python `multiprocessing` primitives for communication

Issue #13843, opened 2023-04-21
Negative missing partition counts on asset graph


johannkm:
![image](https://user-images.githubusercontent.com/22018973/233722146-db405021-830d-4baa-924a-47e4681ac913.png)

Partition subsets aren't getting properly flattened when we report counts

sryza:
Also, it looks like the text is escaping off the left side. Maybe that needs its own separate issue?

PR #14101, opened 2023-05-04
Fix failed partition counts with in progress overlap

johannkm:
Closes https://github.com/dagster-io/dagster/issues/13843

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
@@ -846,13 +846,10 @@ class GrapheneAssetNode(graphene.ObjectType):
             ):
                 check.failed("Expected partitions subset for a partitioned asset")
 
-            num_materialized = len(materialized_partition_subset)
-            num_materialized_and_not_failed = num_materialized - len(
-                [
-                    k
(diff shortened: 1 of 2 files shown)


johannkm:
* master
  * **PR #14101** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14101" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14101?utm_source=stack-comment).

python_modules/dagster-graphql/dagster_graphql/schema/asset_graph.py
```diff
@@ -862,12 +862,20 @@ def resolve_partitionStats(
                 ]
             )
 
+            num_failed_and_not_in_progress = len(
```
johannkm, line 865: if we add many more statuses we'll want a smarter way to do this, akin to the flatten method

clairelin135: nit: seems like this would call `in_progress_subset.get_partition_keys()` k times? can we pull that out so we only fetch it once

clairelin135 (approved):

PR #14102, opened 2023-05-04
Update io-managers.mdx, improper class called

out-running-27:
When looking at the class MyBetterNumpyLoader. It appears that the `_get_path` method being called should be from the BetterPandasIOManager. Further, I would then expect the BetterPandasIOManager class to be part of the resource definition.

I'm new to dagster - so apologies if I'm not understanding the example properly. But this seems like it may be a straightforward fix that can make it easier for the next individual reading through the docs. 😄


jamiedemaria (approved):
yep you are totally correct! thank you for the catch! I need to run CI and then this can merge

seanglynn-thrive (commented):
🦅 👁️

PR #14106, opened 2023-05-05
Make ExecutionPlan build from JobDefinition

smackesey:
## Summary & Motivation

`ExecutionPlan` construction does not require access to reconstruction context, nor does it pass on its job argument to other functions or structures that need this context. Therefore it makes sense for construction to operate directly on `JobDefinition`.

This also reroutes all calls to `ExecutionPlan.build` through `create_execution_plan` (the vast majority of existing calls were already routed this way but there were a few stragglers). This allows `ExecutionPlan.build` to be written directly against `JobDefinition`, as its only callsite is in `create_execution_plan` which should be considered the "public" function for creating an execution plan.

Prior to this PR, `ExecutionPlan` internal construction operated on an `IJob`, but the only thing it did was perform a check on `ReconstructableJob` to make sure its repository load data was the same as any `repository_load_data` passed in. That's no longer necessary if we operate on `JobDefinition` directly, and this check is satisfied in any case by going through `create_execution_plan`.

## How I Tested These Changes

Existing test suite.

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/api.py
+++ b/python_modules/dagster/dagster/_core/execution/api.py
@@ -651,15 +651,6 @@ def execute_plan(
     )
 
 
-def _check_job(job_arg: Union[JobDefinition, IJob]) -> IJob:
-    # backcompat
-    if isinstance(job_arg, JobDefinition):
-        job_arg = InMemoryJob(job_arg)
-
-    check.inst_param(job_arg, "job_arg", IJob)
-    return job_arg
-
-
 def _get_execution_plan_from_run(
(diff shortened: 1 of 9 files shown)


smackesey:
* master
  * **PR #14106** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14106" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14106?utm_source=stack-comment).

alangenfeld (approved):
nice

PR #14108, opened 2023-05-05
Fix LocalFileSystemIOManager example software-defined-assets.mdx

sisidra:
The success file should be written only when all the data has been successfully written.

## Summary & Motivation
Nit-pick semantical fix for example code.


PedramNavid:
Closed by #15908

PR #14109, opened 2023-05-05
Fix table name in example enriching-with-software-defined-assets.mdx

sisidra:
`write_products_table` example function should write to `products` table instead of `categories` table.

## Summary & Motivation
Fix what seems to be a copy/paste mistake in example.


erinkcochran87:
Hey @sisidra! I actually just opened a PR that fixes this and a few other issues in the guide, so I'm gonna close this out. Thanks for reporting it!

PR #14114, opened 2023-05-05
[dagster-airflow] switch uri to string source

Ramshackle-Jamathon:
## Summary & Motivation
to allow passing the uri as an environment variable and obfuscate the value from the UI. @benpankow is there a way to mark config values as secrets in the new config system?


benpankow:
> @benpankow is there a way to mark config values as secrets in the new config system?

No, but this is something that Nick and I have been chatting about a bit. https://github.com/dagster-io/dagster/pull/14044 is one approach (but puts the onus on the caller/user vs the resource provider). There are a couple alternate options in the comments there.

benpankow (approved):

PR #14118, opened 2023-05-05
Refactor AssetNode status logic

salazarm:
## Summary & Motivation

The AssetNode coloring logic is currently entrenched with a bunch of rendering logic specific to the asset graph.
This PR refactors the core logic out into a hook `getAssetNodeStatusCase` which maps each case of the logic to a specific "StatusCase". That function returns an object with a StatusCase and any relevant variables to the status case. The reason for also returning the relevant variables is to "reuse" the type checking done by `getAssetNodeStatusCase`.

The idea is that now I can re-use that core logic that maps an asset to a "status" and I can group these cases into buckets for the new Asset Factory Floor page.

## How I Tested These Changes

before:
screencapture-localhost-6006-iframe-html-2023-05-05-13_28_21

After:
screencapture-localhost-6006-iframe-html-2023-05-05-13_30_02

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
@@ -187,14 +187,12 @@ export type StatusCaseObject =
       liveData: LiveDataForNode;
     } & Pick<GenericStatusProps, 'liveData' | 'currentMinutesLate' | 'late'> &
       Partial<Pick<GenericStatusProps, 'runWhichFailedToMaterialize'>>)
-  | {
+  | ({
       case: StatusCase.NEVER_MATERIALIZED;
-      liveData: LiveDataForNode;
-    }
-  | {
+    } & Pick<GenericStatusProps, 'liveData'>)
+  | ({
       case: StatusCase.MATERIALIZED;
-      liveData: LiveDataForNode;
-    }
+    } & Pick<GenericStatusProps, 'liveData'>)
   | ({
       case: StatusCase.PARTITIONS_FAILED;
     } & Pick<
@@ -570,19 +568,6 @@ export function buildAssetNodeStatusContent({
         ),
       };
   }
-
-  return {
-    background: Colors.Green50,
-    border: Colors.Green500,
-    content: (
-      <>
-        {expanded && <AssetPartitionStatusDot status={[AssetPartitionStatus.MATERIALIZED]} />}
-        <span style={{color: Colors.Green700}}>Materialized</span>
-        {expanded && <SpacerDot />}
-        {lastMaterializationLink}
-      </>
-    ),
-  };
 }
 
 export const AssetNodeMinimal: React.FC<{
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
```diff
@@ -129,28 +134,117 @@ function getStepKey(definition: {opNames: string[]}) {
...
+  MATERIALIZING = 'MATERIALIZING',
+  LATE_OR_FAILED = 'LATE_OR_FAILED',
+  NEVER_MATERIALIZED = 'NEVER_MATERIALIZED',
+  MATERIALIZED = 'MATERIALIZED',
+  PARTITIONS_FAILED = 'PARTITIONS_FAILED',
+  PARTITIONS_MISSING = 'PARTITIONS_MISSING',
+  PARTITIONS_MATERIALIZED = 'PARTITIONS_MATERIALIZED',
+}
+
+type GenericStatusProps = {
```
salazarm, line 152: This exists so I can reference these types without manually typing them every time below

js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
```diff
@@ -162,25 +250,126 @@ export function buildAssetNodeStatusContent({
...
 
   if (definition.isSource) {
     if (materializingRunId) {
+      return {case: StatusCase.SOURCE_OBSERVING, materializingRunId, liveData};
+    }
+    if (liveData?.lastObservation) {
+      return {
+        case: StatusCase.SOURCE_OBSERVED,
+        liveData,
+        lastObservation: liveData.lastObservation,
```
bengotow, line 263: I'm not sure slicing this down to a subset of the properties is that helpful unless we do some meaningful summarization or transformation. Maybe just having the status cases and then rendering them using the available data would be more straightforward? Don't feel too strongly about it though!

bengotow (approved):

bengotow:
I think re: testing, there are pretty extensive AssetNode storybooks - if you want to go to that storybook, screenshot it, switch to this branch, and then visually compare the two that'd probably verify nothing has changed for any state.

js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
salazarm: I can only return the ones that have been calculated so far. For example for loading, liveData would be null, so if I were to share the type between all of the cases it would have to be nullable. Basically this is a way to reuse any type refinements available for the case

js_modules/dagit/packages/core/src/asset-graph/AssetNode.tsx
```diff
@@ -162,25 +250,126 @@ export function buildAssetNodeStatusContent({
...
+  definition,
+  liveData,
+  expanded,
+}: {
+  assetKey: AssetKeyInput;
+  definition: {opNames: string[]; isSource: boolean; isObservable: boolean};
+  liveData: LiveDataForNode | null | undefined;
+  expanded?: boolean;
+}) {
+  const statusCase = getAssetNodeStatusCase({definition, liveData});
```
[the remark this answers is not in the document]
salazarm: I think you're write we should colocate the rendering. Maybe we can make the rendering lazy by converting it into a function we can call.

PR #14120, opened 2023-05-05
solid_selection, solids_to_execute -> op_selection, resolved_op_selection

smackesey:


python_modules/dagster/dagster/_grpc/types.py
```diff
@@ -468,33 +467,33 @@ def __new__(
...
             job_origin=check.inst_param(job_origin, "job_origin", ExternalJobOrigin),
-            solid_selection=check.sequence_param(solid_selection, "solid_selection", of_type=str)
-            if solid_selection
-            else None,
-            asset_selection=check.opt_sequence_param(
-                asset_selection, "asset_selection", of_type=AssetKey
+            op_selection=check.opt_nullable_sequence_param(
+                op_selection, "op_selection", of_type=str
             ),
+            asset_selection=asset_selection,
```
alangenfeld, line 496: this + the type changes above are a bit scary to me - can you give some context.

smackesey: `asset_selection` is a set everywhere else it is used-- I think it was just an oversight that it is a `Sequence` here. Where this payload was consumed (e.g. `get_subset_external_job_result` -> `get_external_pipeline_subset_result`) it was being converted to a set anyway (along with response to this I'm pushing a commit that removes those now-unnecessary conversions, which I'd forgotten to remove).

alangenfeld: was that the case for all time? worried about pulling old stuff out of the DB. Is there a reason to omit `check` here?

smackesey: I'm not sure if it was the case for all time, but as best I can tell we were converting to/from a set on the grpc client/server respectively.

But I think you identified an issue here, which is that older grpc clients may not handle the set if they're expecting a sequence-- I guess we don't have test coverage here.

I added a field serializer to keep the serialization format constant (this is in a new commit). I also added back the `check` call (it was removed because I was doing a superfluous set-to-set conversion to set right above it, which I then removed and forgot to add back the `check`).

alangenfeld (changes requested):
to your queue for now til thats all sorted out, and ideally under some test if there is indeed a clear gap

smackesey:
@alangenfeld Issues fixed, just forgot to pack the elements in the custom set-to-sequence serializer.

I'm not sure there's a test gap, and I think if there is the right place would be in the backcompat test suite, which I'd really rather not get into-- with the custom field serializer (which is tested) the serialization format is staying constant, so I think that's enough.

alangenfeld (approved):

PR #14126, opened 2023-05-05
Split pytest logs into their own Buildkite section

jmsanders:
Similar to where I've done this for some of our internal repos. I assume we're comfortable with the top level conftest?


alangenfeld (approved):
worth it

PR #14127, opened 2023-05-05
lazy load instance run coordinator

prha:
## Summary & Motivation
Make it easier to build the instance from ref, without necessarily having the dependencies loaded.

Simplifies / combines the lazy load tests.

## How I Tested These Changes
BK


johannkm (approved):
What dependencies does the run coordinator have?

prha:
I guess none... do users write their own run coordinators?  or do we just have default and queued?

johannkm:
we have one we inject in cloud, and I know one customer that wrote a custom one to inject a user run tag in oss

johannkm:
I don't mind this for consistency though

gibsondan:
What this would let somebody do is write a custom run coordinator without
needing to make sure that the code is also included in their user code
images as well

PR #14131, opened 2023-05-05
[ui] Fix infinite recursion on asset job partitions page when using TimeWindowPartitionMapping

bengotow:
## Summary & Motivation

Discussion: https://elementl-workspace.slack.com/archives/C03CCE471E0/p1682953217787039

## How I Tested These Changes

I created a repro case (code below), added a failing test case, and then fixed the layout logic.

```

@asset()
def depends_on_nothing():
    return 2


@asset(
    ins={
        "depends_on_yesterday": AssetIn(
            partition_mapping=TimeWindowPartitionMapping(start_offset=-1, end_offset=-1)
        )
    },
)
def depends_on_yesterday(depends_on_nothing, depends_on_yesterday):
    return 2


job = define_asset_job("depends_on_yesterday_job", selection=AssetSelection.keys("depends_on_nothing", "depends_on_yesterday")),
````

### DIFF
--- a/js_modules/dagit/packages/core/src/gantt/GanttChartLayout.ts
+++ b/js_modules/dagit/packages/core/src/gantt/GanttChartLayout.ts
@@ -56,7 +56,11 @@ export const buildLayout = (params: BuildLayoutParams) => {
       return;
     }
     box.x = x;
-    box.children.forEach((child) => deepen(child, box.x + box.width + BOX_SPACING_X));
+    box.children.forEach((child) => {
+      if (child.key !== box.key) {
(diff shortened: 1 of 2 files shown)


salazarm (approved):
I think we need to track the keys seen in the recursion to avoid an infinite loop longer than length 1

bengotow:
Ahh you're totally right that we would -- thankfully for now we're just looking to patch the specific case where an asset depends on itself. Still no plans to support longer-path / indirect loops in DAGs! (Thank goodness haha)

Issue #14123, opened 2023-05-05
[dagster-airbyte] TypeError with Airbyte config as code


joel-olazagasti:
### Dagster version

1.3.3

### What's the issue?

As of Dagster 1.3, the old ResourceDefinition pattern of defining an airbyte resource causes a duplicate resource key error. The easy solution would be to refactor to use the new `AirbyteResource` pythonic configuration, per the new docs, but `load_assets_from_connections` is still expecting a `ResourceDefinition` type, and fails on receiving an `AirbyteResource.` 

### What did you expect to happen?

I believe the ingestion-as-code portion of Dagster-Airbyte just wasn't updated during the 1.3 pythonic resources update. 

### How to reproduce?

Try to use the ingestion as code functionality with an `AirbyteManagedElementReconciler` in Dagster 1.3.3. Both legacy and new pythonic resources will fail, for different reasons.

### Deployment type

Dagster Cloud

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

joel-olazagasti:
PR that fixes the issue [here](https://github.com/dagster-io/dagster/pull/14137)

joel-olazagasti:
Fixed in https://github.com/dagster-io/dagster/pull/14137

PR #14137, opened 2023-05-05
Airbyte resource bugfix

joel-olazagasti:
## Summary & Motivation
As of `1.3.0` The dagster-managed-elements library, specifically the Airbyte ingestion as code functionality, was no longer compatible with the Airbyte resource. This PR changes the parameter type passed to `AirbyteManagedReconciler` and upstream dependencies to `AirbyteResource` instead of the generic `ResourceDefinition`. 
## How I Tested These Changes
I ran the test suite for the Airbyte-dagster library, including the underlying managed elements. All tests passed except for those in `test-managed-elements.py`. Those won't run no matter what, at least for me locally, because the Airbyte docker compose instance fails to spin up in time, and none of the subsequent tests are run. I did, however, make identical changes in my company's implementation of dagster, and used them to successfully spin up `dagster dev` and used the managed elements cli resource to deploy airbyte config to a test airbyte instance. So I can confirm that with these changes, one can successfully use the Airbyte ingestion as code in `1.3.3` by simply reconfiguring their airbyte_instance definition to use the new pythonic `AirbyteResource`

Note: 
The first commit here: c539fa9b054761a611ae4b233dd396c0acbeb0ba
was from a previous PR of mine. Not 100% sure why it's showing up again, as that PR was already merged, but it doesn't actually change the file on master.

Raised in this [issue](https://github.com/dagster-io/dagster/issues/14123)


benpankow:
This looks great, thanks for the PR @joel-olazagasti! One ask - this change will be backwards incompatible for folks using the old resource. I've put up [a commit here stacked on the PR](https://github.com/dagster-io/dagster/commit/bcb17a8a4fc63fc00e5cd41cff1dfdb088b1fe72) which keeps backcompat (keeping the integration test working) + fixes some of the minor lint quibbles. Otherwise, seems good to go.

joel-olazagasti:
Awesome, thanks @benpankow! The backwards compatibility is definitely the way to go. Weird that `make ruff` didn't give me a warning and  `make black` didn't take. Must be something with nvim settings that overwrote it with my formatter still. Do you need me to take any action with that commit, or will you handle pulling in and merging?

benpankow:
I'll cherry-pick the commit and get it merged. Thanks again!

benpankow (approved):

Issue #3734, opened 2021-02-24
grpc SSL support


dehume-drizly:
## Use Case
For our production setup to run properly using grpc workspaces we need dagster grpc to support SSL and be able to be set in the workspace yaml

`workspace.yaml`
```
load_from:
  - grpc_server:
      host: grpc.drizly.com
      port: 4000
      ssl: True
      location_name: "bi"
```

## Ideas of Implementation
Update `DagsterGrpcClient` with an optional parameter for ssl and then have channels use `secure_channel` if ssl is set. Something like:
```
    def _channel(self):
        if self.ssl:
            credential = grpc.ssl_channel_credentials()
            return grpc.secure_channel(self._server_address, credential)
        return grpc.insecure_channel(self._server_address)
```

## Additional Info

---
#### Message from the maintainers:
Excited about this feature? Give it a :thumbsup:. We factor engagement into prioritization.

gibsondan:
(This was added successfully and pushed in an 0.10.x release)

nicolas-luchetti:
HI @dehume-drizly  if I want to have more than one workspace, how should I leave the .yml? Thanks!

jrouly:
Added support for passing `ssl` in the Helm chart values for the workspace configmap: https://github.com/dagster-io/dagster/pull/14146/files

PR #14146, opened 2023-05-07
Support resolving remote servers over SSL

jrouly:
## Summary & Motivation

Dagster already supports resolving remote servers that require SSL being specified in `workspace.yaml` with `ssl: true` ([issue](https://github.com/dagster-io/dagster/issues/3734)), but the Helm chart doesn't allow you to pass `ssl`.

This PR changes that, allowing you to specify whether or not a remote server should be resolved over SSL.

### unit tests

I added two unit tests: one without `ssl`, and one with it.

### practical test

First I tried deploying with `ssl: true` in the values, and noticed that it wasn't making it into my workspace configmap.

I manually edited the configmap to add `ssl: true`, and Dagit was able to resolve my user code location (yay).

Then after making the change in this PR, I performed the following steps to verify the change.
- removed my manual changes to the configmap
- changed the Helm chart version to `1.3.1` to correspond with the version of Dagit we have been running
- ran `helm package helm/dagster` to package up a Helm chart
- updated my deployment configuration to point at the packaged Helm chart (which includes my changes in this PR)
- deployed it
- verified that `ssl: true` shows up in the appropriate server in the workspace configmap
- verified that Dagit can load my code location over SSL (yay)

### DIFF
--- a/helm/dagster/schema/schema_tests/test_dagit.py
+++ b/helm/dagster/schema/schema_tests/test_dagit.py
@@ -1,4 +1,5 @@
 import pytest
+import yaml
 from dagster_k8s.models import k8s_model_from_dict, k8s_snake_case_dict
 from kubernetes import client as k8s_client
 from kubernetes.client import models
@@ -351,11 +352,16 @@ def test_dagit_workspace_servers(workspace_configmap_template: HelmTemplate):
         ),
     )
 
+    expected = {
+        "load_from": [
+            {"grpc_server": {"host": "example.com", "port": 443, "location_name": "Example"}}
+        ]
+    }
+
     [dagit_workspace_configmap] = workspace_configmap_template.render(helm_values)
-    assert (
-        dagit_workspace_configmap.data["workspace.yaml"]
-        == """load_from:\n  - grpc_server:\n      host: example.com\n      port: 443\n      location_name: Example\n"""
-    )
+    actual = yaml.safe_load(dagit_workspace_configmap.data["workspace.yaml"])
+
+    assert actual == expected
 
 
 def test_dagit_workspace_servers_ssl(workspace_configmap_template: HelmTemplate):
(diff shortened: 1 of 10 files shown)


jrouly:
@rexledesma you may be interested in this PR if you're still doing Helm things with Dagster :)

helm/dagster/values.yaml
```diff
@@ -64,6 +64,7 @@ dagit:
       - host: "k8s-example-user-code-1"
         port: 3030
         name: "user-code-example"
+        ssl: true
```
rexledesma, line 67: Let's remove this from the defaults to make this backwards compatible.
```suggestion
```

helm/dagster/schema/schema_tests/test_dagit.py
```diff
@@ -324,6 +334,55 @@ def test_dagit_workspace_external_configmap(deployment_template: HelmTemplate):
...
+                externalConfigmap=None,
+            )
+        ),
+    )
+
+    [dagit_workspace_configmap] = workspace_configmap_template.render(helm_values)
+    assert (
+        dagit_workspace_configmap.data["workspace.yaml"]
+        == """load_from:\n  - grpc_server:\n      host: example.com\n      port: 443\n      location_name: Example\n"""
+    )
```
rexledesma, line 358: can we parse this yaml and assert it against a dict to make this more legible?

rexledesma (changes requested):
Thanks for the contribution! This looks fine to merge, just some minor comments

helm/dagster/schema/schema_tests/test_dagit.py
jrouly: Addressed.

jrouly:
@rexledesma is there a command I should be running locally before pushing? I ran `make` and it looks like everything is passing, but I see the `ruff` commit you pushed and the buildkite failure (which unfortunately looks like I can't see the results of).

jrouly:
I ran `pre-commit run -a` which produced [2241fe8](https://github.com/dagster-io/dagster/pull/14146/commits/2241fe8050c029b41c21b4ad1f69a7d7cd92e646). These changes are unrelated to my PR, so not sure how you feel about including them here.

jrouly:
I've reverted [`2241fe8`](https://github.com/dagster-io/dagster/pull/14146/commits/2241fe8050c029b41c21b4ad1f69a7d7cd92e646) which was `pre-commit` updating a bunch of doc files. Hopefully the PR is good to go as-is now? Unless there are other scripts I should be running.

PR #14147, opened 2023-05-08
WIP: Add a grpc proxy server that allows you to reload code within dagit without restarting the whole process

gibsondan:
Alternate solution to https://github.com/dagster-io/dagster/pull/14097/files . grpc servers can now be spun up as system code processes that don't actually load user code.

The simplest way to implement this was to have them both implement the same grpc api, but have the proxy forward it on to a 'real' grpc server running as a subprocess.

that part was actually relatively straightforward and seems to work (although lots of cleanup still required, and unclear if we'll end up using GrpcServerRegistry here vs. interacting with the subprocess directly)

The thing i'm now stuck on is making it thread-safe in dagit :/ I put a comment inline with the specific question i'm struggling with in case you all have any ideas.

### DIFF
--- a/python_modules/dagster/dagster/_core/workspace/context.py
+++ b/python_modules/dagster/dagster/_core/workspace/context.py
@@ -33,7 +33,10 @@ from dagster._core.host_representation.grpc_server_state_subscriber import (
     LocationStateChangeEventType,
     LocationStateSubscriber,
 )
-from dagster._core.host_representation.origin import GrpcServerCodeLocationOrigin
+from dagster._core.host_representation.origin import (
+    GrpcServerCodeLocationOrigin,
+    ManagedGrpcPythonEnvCodeLocationOrigin,
+)
 from dagster._core.instance import DagsterInstance
 from dagster._utils.error import SerializableErrorInfo, serializable_error_info_from_exc_info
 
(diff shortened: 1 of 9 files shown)


python_modules/dagster/dagster/_core/workspace/context.py
```diff
@@ -501,15 +512,47 @@ def rm_state_subscriber(self, token: int) -> None:
...
+            if event:
+                event.set()
+                thread.join()
+
+            try:
+                new_location = origin.reload_location()
+            finally:
+                self._start_watch_thread(origin)
+
+            return new_location
```
gibsondan, line 549: This is super gross and the main thing keeping me from pushing this PR forward right now. The code here is not particularly correct, but the problem that it's intending to solve is:

- User initiates a reload from a graphql request, that happens synchronously during the request
- There are also these background threads that watch for updates to the gRPC server and prompt you to update if the server ID changes, they will pick up on the change
- We don't want that - the user is initiating the request, so there's no need to pop something up telling them to reload

So this was intended to keep that from happening - can't come up with a solution I like to deal with that

alangenfeld: Other users observing the same dagit instance *should* get the prompt to reload their stale copy of the workspace when this fires though right? If the only problem is the dagit ux, it may require a dagit solution of ignoring/disabling the subscription when the mutation is firing.

gibsondan: oh hm, that is a good point. I kinda feel like the whole experience around that prompt is not so good, and fixing that might resolve this - why doesn't it just update things in place when it detects there has been a change?

alangenfeld: i think its to avoid the risk of blowing things away when a user "is in the middle of something" in the not so happy paths like where whole jobs are disappearing  / moving

gibsondan: OK yeah fair enough - pursuing the frontend 'disable the prompt while this mutation is happening' angle with @hellendag , will see how hard that is

gibsondan:
that was helpful, will return with something more fully baked

Issue #14103, opened 2023-05-04
auto-observe source assets


sryza:
Allow specifying automatic source asset observations. Something like:

```python
@observable_source_asset(auto_observe_frequency_minutes=30)
def raw_data():
    ...
```

### What we've heard

- Prratek Ramchandani - https://dagster.slack.com/archives/C01U5LFUZJS/p1683032402854509
  - hey, as we refactor more of our dagster code to use declarative scheduling patterns i’ve been using observable source assets a fair bit and have some feedback. previously, if we had tables in our data warehouse that were populated by some process outside of dagster, we'd create custom sensors to check when those tables were updated and kick off downstream jobs. instead we're now representing them as observable source assets, which seems like the recommended practice:
  - observation functions being run as jobs feels a little unintuitive - i'm using them to tell me if the asset in question has changed and think of it as a background process that evaluates every few seconds along with typical sensor evaluations
  - i like the benefit of then being able to use auto materializations for downstream assets but would like the process of "observing" a source asset to feel more lightweight

PR #14148, opened 2023-05-08
auto-observe source assets: API addition

sryza:
## Background

(Much of the following language stolen from an earlier discussion by Owen):

Observing source assets is critical to the functionality of auto-materialization, as we need to know when sources are updated in order to know when to update our downstream assets.

We never know what the "best time" is to run an observation function, as it's always valuable to know the specific time that a source changes, and this can be very unpredictable. However, the downside to running observation functions more often is that this computation may be expensive / noisy. Different source assets may have different properties. For something that you don't expect to change very often, running it once every few hours might be fine. For something you expect to change very frequently, you'll want to run the observation function much more often.

Currently, users can create a scheduled job to observe their source assets, as documented at the bottom of [this docs section](https://docs.dagster.io/concepts/assets/asset-observations#observable-source-assets). However, this requires them to use and understand concepts like schedules and jobs that they otherwise wouldn't need to when using declarative scheduling.

### Proposal

Proposal: Observable source assets get an `auto_observe_frequency_minutes` argument. The same daemon that launches auto-materialization runs launches auto-observation runs, at the specified frequency.

I.e., if the asset daemon is switched on, a run will be launched to observe this asset every 30 minutes:

```python
@observable_source_asset(auto_observe_frequency_minutes=30)
def raw_events():
    ...
```

Some implications:
- Because observations happens as part of a run that's launched by the daemon, separate ticks are required to observe an asset and then auto-materialize downstream assets based on that observation.
The UI will need to be updated to reflect the fact that the asset daemon can launch observations as well as materializations. I.e. this is no longer comprehensive:
  - image

Along with downstream PRs, resolves https://github.com/dagster-io/dagster/issues/14103. Next PR in the stack: https://github.com/dagster-io/dagster/pull/14149.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -71,6 +71,7 @@ class AssetGraph:
         required_multi_asset_sets_by_key: Optional[Mapping[AssetKey, AbstractSet[AssetKey]]],
         code_versions_by_key: Mapping[AssetKey, Optional[str]],
         is_observable_by_key: Mapping[AssetKey, bool],
+        auto_observe_interval_minutes_by_key: Mapping[AssetKey, Optional[float]],
     ):
         self._asset_dep_graph = asset_dep_graph
         self._source_asset_keys = source_asset_keys
@@ -82,6 +83,7 @@ class AssetGraph:
         self._required_multi_asset_sets_by_key = required_multi_asset_sets_by_key
         self._code_versions_by_key = code_versions_by_key
         self._is_observable_by_key = is_observable_by_key
+        self._auto_observe_interval_minutes_by_key = auto_observe_interval_minutes_by_key
 
     @property
     def asset_dep_graph(self) -> DependencyGraph[AssetKey]:
@@ -112,6 +114,9 @@ class AssetGraph:
     ) -> Mapping[AssetKey, Optional[AutoMaterializePolicy]]:
         return self._auto_materialize_policies_by_key
 
+    def get_auto_observe_interval_minutes(self, asset_key: AssetKey) -> Optional[float]:
+        return self._auto_observe_interval_minutes_by_key.get(asset_key)
+
     @staticmethod
     def from_assets(
         all_assets: Iterable[Union[AssetsDefinition, SourceAsset]]
@@ -128,6 +133,7 @@ class AssetGraph:
         required_multi_asset_sets_by_key: Dict[AssetKey, AbstractSet[AssetKey]] = {}
         code_versions_by_key: Dict[AssetKey, Optional[str]] = {}
         is_observable_by_key: Dict[AssetKey, bool] = {}
+        auto_observe_interval_minutes_by_key: Dict[AssetKey, Optional[float]] = {}
 
         for asset in all_assets:
             if isinstance(asset, SourceAsset):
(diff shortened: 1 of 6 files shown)


OwenKephart:
I largely supportive of this, but I want to document my own internal debate in case other people have/had similar qualms.

First, at a high level, making this auto-observation bit part of the AutoMaterializeDaemon caps a maximum frequency that we can observe things. In general, launching runs to auto-observe assets is basically a free operation, meaning if we had a completely separate daemon, then it would be fairly easy to support launching observations at very high frequency (i.e. every few seconds).

However, the logic for auto materialization is much more expensive, and this sort of sub-second evaluation is not possible in the general case. This means that this design places an "unfair" limit on the observation frequency. If your materialization logic takes 20 minutes to complete, then your assets won't get observed for 20 minutes, even if it would have been easy to do so.

In the end, I feel like this is ok. In general, the reason you're auto-observing your source assets is so that you can auto-materialize downstream assets. So it doesn't really matter if you observe your source assets a bunch while your materialization logic is chugging along, as it wouldn't be able to react to that new information anyway.

Second, the naming. I think it does make sense to have `minutes` be the primary unit here (rather than seconds), due to the above issue (we can't really promise that if you set seconds=10, that will do anything different than seconds=30, but minutes=10 vs minutes=30 will certainly have different behaviors). Also it'd be nice to encourage people not to run these things a silly number of times per day.

However, it'd be nice to somehow signal that this is more of a suggestion than a precise timing. I.e. we don't want someone to be confused if they set `auto_observe_frequency_minutes=10` and their observations are 10, 11, or 12 minutes apart (rather than always exactly 10).

We have a similar constraint in sensor evaluation world, where we've called the sensor frequency `minimum_interval_seconds`, to represent the fact that if a sensor evaluation takes longer than X amount of time, then it's possible the next tick will be more than X seconds after the previous one.

So maybe this should be called `minimum_auto_observe_interval_minutes` or something like that?

sryza:
> First, at a high level, making this auto-observation bit part of the AutoMaterializeDaemon caps a maximum frequency that we can observe things. In general, launching runs to auto-observe assets is basically a free operation, meaning if we had a completely separate daemon, then it would be fairly easy to support launching observations at very high frequency (i.e. every few seconds).

Yeah, I can see the advantage here of separating it out. I think we have the freedom to do this in the future. I.e. nothing about the API forces it to be in the same daemon.

> Also it'd be nice to encourage people not to run these things a silly number of times per day.

Yeah, this is the main reason to use minutes IMO. It's technically a float, so they can ask for sub-minute (although, as you say, it will be hard to actually provide this).

> We have a similar constraint in sensor evaluation world, where we've called the sensor frequency `minimum_interval_seconds`

FWIW this parameter always rubs me a little weird. It makes me feel like the parameter's worldview is that there's some reason I wouldn't want the sensor to execute too often and I'm OK with any value that's higher than the minimum. When, in reality, what I'm saying is "do your best to make it happen this often".

"interval" might be better than "frequency" though.  Frequency implies a volume of executions - i.e. if my frequency is every 30 minutes and I've gone a day without executing, I might try to observe 48 times in the last hour so that my frequency is 48 observations/day = 1/30 minutes. (This is obviously undesirable behavior).

OwenKephart (approved):
@sryza based on that discussion, I think if `auto_observe_frequency_minutes` -> `auto_observe_interval_mintues`, then I'm on board. Agree that "minimum" is a bit weird in that context, although it's a bit sad to be inconsistent with the sensor.

salazarm:
> In the end, I feel like this is ok. In general, the reason you're auto-observing your source assets is so that you can auto-materialize downstream assets. So it doesn't really matter if you observe your source assets a bunch while your materialization logic is chugging along, as it wouldn't be able to react to that new information anyway.

I  can imagine a world where in order to further scale customers could setup more than 1 automaterialization daemon. But even in that world it wouldn't make sense for observations to happen in a separate daemon for similar reasons. Though it does depend a bit on the resource allocation strategy being used in this multi asset daemon scenario (eg. it could be possible in this world for resources to be allocated within organizational boundaries.

PR #14149, opened 2023-05-08
auto-observe assets: asset daemon

sryza:
## Summary & Motivation

Augments the asset daemon to auto-observe source assets that have an `auto_observe_interval_minutes` set.


python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/test_auto_observe.py
```diff
@@ -0,0 +1,121 @@
+"""fdsjkldf.
```
OwenKephart: 👁️ 👄 👁️

python_modules/dagster/dagster/_daemon/asset_daemon.py
```diff
@@ -111,63 +119,14 @@ def run_iteration(
...
-                run_request.run_config,
-                step_keys_to_execute=None,
-                known_state=None,
-                instance=instance,
+            run = submit_asset_run(
+                run_request,
+                instance,
+                workspace,
+                asset_graph,
+                {AUTO_MATERIALIZE_TAG: "true", **instance.auto_materialize_run_tags},
```
OwenKephart, line 128: No need to handle for this PR, but theoretically the auto-observe assets might desire a different tag for their runs (likely all auto-kicked off runs should have the same tag, but that tag shouldn't be called "auto_materialize")

python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/auto_observe_scenarios.py
```diff
@@ -0,0 +1,62 @@
...
+import pendulum
+from dagster import DataVersion, observable_source_asset
+
+from .asset_reconciliation_scenario import (
+    AssetReconciliationScenario,
+    run_request,
+)
+
+
+@observable_source_asset(auto_observe_interval_minutes=30)
```
OwenKephart, line 12: should we have a test with a partitioned observable source asset?

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -1102,9 +1166,36 @@ def _sensor(context):
             instance=context.instance,
             cursor=cursor,
             run_tags=run_tags,
+            auto_observe=False,
```
OwenKephart, line 1169: can we just remove this boolean and just have this always be True? this sensor will be removed imminently anyway, and I don't see a huge benefit to turning off the behavior for one release

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -344,20 +363,35 @@ def empty(cls) -> "AssetReconciliationCursor":
...
+            (
+                latest_storage_id,
+                serialized_materialized_or_requested_root_asset_keys,
+                serialized_materialized_or_requested_root_partitions_by_asset_key,
+            ) = data[:3]
 
-        evaluation_id = data[3] if len(data) == 4 else 0
+            evaluation_id = data[3] if len(data) == 4 else 0
+            serialized_last_observe_request_timestamp_by_asset_key = {}
+        else:
```
OwenKephart, line 383: this is getting pretty nasty -- might be nice to swap over to the `whitelist_for_serdes` serializer

OwenKephart (approved):
some notes, otherwise I'm on board

python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/test_asset_reconciliation_sensor.py
```diff
@@ -183,3 +190,43 @@ def repo():
         match=r"build_asset_reconciliation_sensor: Asset '.*' has an AutoMaterializePolicy set",
     ):
         reconciliation_sensor(context)
+
+
+def test_deserialize_asset_reconciliation_cursor_backcompat():
```
johannkm, line 195: glad to have this test- it could move to this file with the former backcompat test (which probably needs updating) https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/test_asset_daemon_cursor.py

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -404,12 +442,18 @@ def serialize(self) -> str:
             for key, subset in self.materialized_or_requested_root_partitions_by_asset_key.items()
         }
         serialized = json.dumps(
-            (
-                self.latest_storage_id,
-                [key.to_user_string() for key in self.materialized_or_requested_root_asset_keys],
-                serializable_materialized_or_requested_root_partitions_by_asset_key,
-                self.evaluation_id,
-            )
+            {
```
johannkm, line 445: a good change. I wonder how hard it would be to go all the way to using a proper NamedTuple with a custom deserializer for backcompat.

johannkm: ah yeah just echoing owen

sryza: The difficulty here, and the reason we went with this approach originally, is that some of the fields are `PartitionsSubset`s , and `PartitionsSubset`s require the matching `PartitionsDefinition` to be deserialized.

To make that work, we'd need to either modify the deserialization machinery, or modify `PartitionsSubset`s to not require `PartitionsDefinition`s to be deserialized. Neither of these are super straightforward though.

cc @OwenKephart

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
sryza: asset selections involving source assets are kinda gnarly, and I think that unless we're really careful – aka add some tests and stuff – the asset reconciliation sensor will end up accidentally automatically observing some source assets that the user didn't expect to include.

if it's purely about saving effort, ripping out this parameter feels easier than adding those tests

sryza: * by "ripping out this parameter" I mean "submitting a PR next week to rip out this parameter"

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
sryza: agree that this is nasty, but alas I don't think there's a straightforward way to avoid the nastiness: https://github.com/dagster-io/dagster/pull/14149/files#r1255004744

python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/test_asset_reconciliation_sensor.py
sryza: good point - moving

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
OwenKephart: ah forgot about that, oh well

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
OwenKephart: sounds good

python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/auto_observe_scenarios.py
sryza: good idea - adding. shouldn't actually behave any differently than the single asset scenario, but good to encode that.

python_modules/dagster/dagster/_daemon/asset_daemon.py
sryza: Hmm good point. I'm going to actually add an auto-observe tag, because this is weird otherwise.

PR #14155, opened 2023-05-08
Enable to set an `end_dt` in TimeWindowPartitionsDefinition to stop generating new partitions

renzhe-brian:
## Summary
Add an optional argument `end` date for `TimeWindowPartitionsDefinition`,  so that it stops generating new partitions from `end` date. Eg.
```
HourlyPartitionsDefinition(start_date="2022-01-01-00:00", end_date="2022-02-01-00:00")
DailyPartitionsDefinition(start_date="2022-01-01", end_date="2022-02-01")
```
## Motivation
We have a use case, which we need to keep the asset with TimeWindowPartitionsDefinition, but we must stop running new partitions from an end date.

A simple work around is to remove the scheduler for that job, but it will keep generating new partitions from UI. It doesn't run these partitions though, since the scheduler is removed. But it's confusing for users. Therefore I make this PR

## How I Tested These Changes
* Unittests are included in the PR
* Test in our environment:
image
image

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
@@ -123,7 +123,7 @@ class TimeWindowPartitionsDefinition(
             start_dt = pendulum.instance(datetime.strptime(start, fmt), tz=timezone)
 
         if not end:
-            end_dt = pendulum.instance(datetime.strptime("9999-12-31", "%Y-%m-%d"), tz=timezone)
+            end_dt = None
         elif isinstance(end, datetime):
             end_dt = pendulum.instance(end, tz=timezone)
         else:
--- a/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
@@ -479,6 +479,9 @@ class TimeWindowPartitionsDefinition(
             else pendulum.now(self.timezone)
         )
 
+        if self.end and self.end < current_time:
+            current_time = self.end
+
         if self.end_offset == 0:
             return next(iter(self._reverse_iterate_time_windows(current_time)))
         else:
--- a/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
@@ -66,7 +66,7 @@ class TimeWindowPartitionsDefinition(
         [
             ("start", PublicAttr[datetime]),
             ("timezone", PublicAttr[str]),
-            ("end", PublicAttr[datetime]),
+            ("end", PublicAttr[Optional[datetime]]),
             ("fmt", PublicAttr[str]),
             ("end_offset", PublicAttr[int]),
             ("cron_schedule", PublicAttr[str]),
--- a/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
@@ -66,6 +66,7 @@ class TimeWindowPartitionsDefinition(
         [
             ("start", PublicAttr[datetime]),
             ("timezone", PublicAttr[str]),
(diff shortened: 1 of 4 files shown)


sryza (commented):
Thanks for submitting this @renzhe-brian !

For this to go in, I think the main thing is that we'd need unit tests that the following methods work as expected:
- get_partition_keys
- get_last_partition_window
- get_next_partition_key
- get_next_partition_window
- get_last_partition_key
- has_partition_key

renzhe-brian:
Hi @sryza , Thank you for your feedback. Should the unittests be added here https://github.com/dagster-io/dagster/blob/0b1ee06b187e3febe13487f23963d892cbfee594/python_modules/dagster/dagster_tests/asset_defs_tests/partition_mapping_tests/test_time_window_partition_mapping.py?

Should maintainers add unittests or shall I add some unittests in current PR?

sryza:
> Should the unittests be added here https://github.com/dagster-io/dagster/blob/0b1ee06b187e3febe13487f23963d892cbfee594/python_modules/dagster/dagster_tests/asset_defs_tests/partition_mapping_tests/test_time_window_partition_mapping.py?

This would be the file for the tests: https://github.com/dagster-io/dagster/blob/0b1ee06b187e3febe13487f23963d892cbfee594/python_modules/dagster/dagster_tests/definitions_tests/test_time_window_partitions.py.

> Should maintainers add unittests or shall I add some unittests in current PR?

Unit tests usually go in the same PR that the functionality is added. I would try to help you out with this if I had time, but I think it might be a while until I have bandwidth, so I'd suggest adding them yourself if this is at all time-sensitive for you.

renzhe-brian:
Hi @sryza , could you take a look at the unittests I created? And please let me know what you think.

sryza:
These tests look great, and this PR looks very close.

A few outstanding things that came up with the build:
1. Are you able to run `make ruff` and `make black` in the root of the repo to auto-format the code?
2. There was a test failure:
```
[2023-05-09T16:00:24Z] dagster_tests/core_tests/host_representation_tests/test_external_data.py::test_back_compat_external_time_window_partitions_def FAILED [ 54%]
```

I think the best way to fix this would be for `TimeWindowPartitionsDefinition.end` to be `None` in the case where the `None`  value is provided to the constructor. Would this work?

sryza:
Tests are passing now, but the build is reporting one type error:

```
[2023-05-09T20:37:38Z]   158:37: Argument of type "Any | None" cannot be assigned to parameter "end" of type "PublicAttr[datetime]" in function "__new__"
[2023-05-09T20:37:38Z]   Type "Any | None" cannot be assigned to type "PublicAttr[datetime]"
[2023-05-09T20:37:38Z]     Type "None" cannot be assigned to type "PublicAttr[datetime]" (reportGeneralTypeIssues)
```

You can run type checks by running `make pyright`.

renzhe-brian:
@sryza , oh, this is caused by last commit which sets `end` to None if None is provided in the constructor

python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -66,6 +66,7 @@ class TimeWindowPartitionsDefinition(
         [
             ("start", PublicAttr[datetime]),
             ("timezone", PublicAttr[str]),
+            ("end", PublicAttr[datetime]),
```
sryza, line 69:
```suggestion
            ("end", PublicAttr[Optional[datetime]]),
```

should fix the type error

renzhe-brian: I'm running `make pyright` locally. But it takes very long time. Is that normal?

sryza: It takes a very long time the first time. After that, I believe, it should take about a minute.

sryza (approved):
This looks good to go. Thanks for adding this @renzhe-brian ! I'm going to merge.

renzhe-brian:
Thank you @sryza. It's a very pleasant experience.

Issue #7963, opened 2022-05-18
way to see all the runs that targeted an asset, even if they didn't materialize it


sryza:
Reasons why a run might not materialize an asset that it targets:
* The run fails
* The code inside the run determines that the asset is up-to-date and doesn't need to be rematerialized

bengotow:
I think we looked at doing this back when we started adding live data to the asset graph and there was a technical reason that it was challenging at the GraphQL layer. We landed on exposing just assetNode.latestRun, which leverages a `last_run_id` that is cached on to each `AssetEntry`. I think scanning the whole runs table for runs targeting a specific asset required a big refactor at the time? Maybe it'd be more feasible now!

sryza:
I think it's likely that we added ASSET_MATERIALIZATION_PLANNED since then, which I believe should make this trivial. Thoughts @clairelin135 ?

clairelin135:
Yep, I think this should be pretty trivial using materialization planned events. We can just query for materialization planned events for a given asset key and fetch the corresponding run IDs.

PR #14165, opened 2023-05-08
Allow filtering the new runs table by Asset

bengotow:
## Summary & Motivation

Resolves https://github.com/dagster-io/dagster/issues/7963

This PR adds a new GraphQL `RunsFilter.assetKeys` option for filtering for runs involving a particular asset and implements it in the front-end filter dropdown.

On the backend, it's difficult to translate this into a filter within run storage. Instead, we parse the `GrapheneRunsFilter` into both a `RunsFilter` and the passed asset keys, resolve the asset keys to run IDs, and build a final `RunsFilter` with the run id selection.

This is admittedly a bit of a hack, but allows us to re-use the existing `instance.run_ids_for_asset_key` helper method which respects requests to wipe asset materializations and keeps run and event storage separate. Very open to other implementations!

image

## How I Tested These Changes

I added a new test that checks a variety of different combinations of this new filter param and the existing run_ids filter param for the correct results.

### DIFF
--- a/js_modules/dagit/packages/core/src/graphql/schema.graphql
+++ b/js_modules/dagit/packages/core/src/graphql/schema.graphql
@@ -2056,6 +2056,7 @@ input RunsFilter {
   updatedAfter: Float
   createdBefore: Float
   mode: String
+  assetKeys: [AssetKeyInput]
 }
 
 input PipelineSelector {
--- a/js_modules/dagit/packages/core/src/graphql/types.ts
+++ b/js_modules/dagit/packages/core/src/graphql/types.ts
@@ -3335,6 +3335,7 @@ export type Runs = PipelineRuns & {
 };
 
 export type RunsFilter = {
+  assetKeys?: InputMaybe<Array<InputMaybe<AssetKeyInput>>>;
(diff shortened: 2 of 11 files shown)


prha:
I'm a little worried about this implementation... we're hitting the event log storage with some pretty expensive queries to resolve all run_ids for all time (by asset).  Does this work in `shadow-dagit` against real data?  My guess is it'd timeout a lot.

This implementation is using `get_asset_run_ids`, but I don't think that's being used anywhere else...

I think ideally, we write some of this data on the runs table itself, using something like a json column (it might be that we can only support this with certain versions of postgres/mysql storage backends)

bengotow:
Thanks @prha that sounds good to me! I may ask around and see if anyone would have bandwidth to help with a more significant backend component of this. I agree it's a little sketchy, the lack of any bounding on the number of run_ids scares me about the current solution too.

Adding a new column or converting run_body to JSONB with a GIN index would let us do something like this more directly on the runs table and seems much better:

```
SELECT * FROM runs WHERE run_body::jsonb @> '{"asset_selection": {"__frozenset__": [{"path": ["collect_latest_data"]}]}}';
```

PR #14167, opened 2023-05-08
Re-generate grpc api proto

gibsondan:
Summary:
The latest version of grpcio+protobuf that we're using seems to generate much more pycharm-friendly stubs. Use those instead.

Test Plan: BK (includes protobuf3 and 4)
Make extra sure that protobuf4, protobuf3, and the earliest version of grpcio that we support still work by running 'dagster dev' locally


smackesey (approved):
Sweet-- goodbye `type: ignore`

PR #14174, opened 2023-05-09
fix(dagster_airbyte): Add new normalization config check

### LINKED ISSUE #13524: dagster-airbyte misidentifies destinations which can support normalization
### Dagster version

version 1.2.6

### What's the issue?

https://github.com/airbytehq/airbyte/pull/21005

I believe due to the change above in the Airbyte API (which, as far as I can tell is not reflected in the openapi spec) the dagster-airbyte library can no longer correctly configure a connection to `normalize_data=True`. Last I understood, the dagster Ingestion as Code sources/destinations were code-genned from Airbyte documentation, so it seems the two may have gotten out of sync. 

When I set the connection parameter `normalize_data = True` on a destination that supports normalization, such as Snowflake, I do not receive the exception:
`Exception: Destination $destination_name does not support normalization`
and my connection configuration is a success.

### How to reproduce?

Use Airbyte version `0.43.1` and try to configure a connection with `normalize_data  = True` in dagster-airbyte Ingestion as Code.

### Deployment type

Dagster Cloud

### Deployment details

Right now I'm running this command locally to test, before it hits our CICD. 

### Additional information

There's an open thread in #airbyte-help in the Airbyte Slack about the openapi spec being out of sync, ~2 weeks old. I've asked for an update on that status, and I'll update here when I receive a response

nina-j:
## Summary & Motivation
The Airbyte API changed source of truth for normalization availability in https://github.com/airbytehq/airbyte/pull/21005. This commit adds an additional check for normalization configuration in `does_dest_support_normalization`. The current check is kept to ensure backward compatibility - thus, if any of the checks are true, the function returns true.

Fixes #13524.

## How I Tested These Changes
I ran the test suite for the `dagster_airbyte` module and added tests that cover my changes.

### DIFF
--- a/python_modules/libraries/dagster-airbyte/dagster_airbyte/resources.py
+++ b/python_modules/libraries/dagster-airbyte/dagster_airbyte/resources.py
@@ -514,8 +514,10 @@ class AirbyteResource(BaseAirbyteResource):
 
     def does_dest_support_normalization(
         self, destination_definition_id: str, workspace_id: str
-    ) -> Dict[str, Any]:
-        return cast(
+    ) -> bool:
+        # Airbyte API changed source of truth for normalization in PR
+        # https://github.com/airbytehq/airbyte/pull/21005
+        norm_dest_def_spec: bool = cast(
             Dict[str, Any],
             check.not_none(
                 self.make_request_cached(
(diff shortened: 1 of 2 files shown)


smackesey:
Thanks for the PR, this looks reasonable and I'm running our CI on it.

smackesey:
Alright this looks great, thanks and merging!

PR #14175, opened 2023-05-09
[Runs Table] Show run tags dialog when pressing "t" hotkey while hovering a row 

salazarm:
## Summary & Motivation

1. We want to be able to quickly open/close the tags dialog by clicking "t" on the row that's being hovered
2. We want to make it possible to reach the dialog with 1 click instead of 2 so we added a new "tag" to be shown that has a link to the dialog.

## How I Tested These Changes

Screen Shot 2023-05-09 at 7 28 23 AM
Screen Shot 2023-05-09 at 7 28 10 AM
Screen Shot 2023-05-09 at 7 28 06 AM


js_modules/dagit/packages/core/src/runs/RunTableNew.tsx
```diff
@@ -240,8 +256,26 @@ const RunRow: React.FC<{
...
+  const containerRef = React.useRef<HTMLTableRowElement | null>(null);
+  React.useLayoutEffect(() => {
+    const mousemoveListener = (ev: MouseEvent) => {
+      if (!containerRef.current?.contains(ev.target as HTMLElement)) {
+        setShowRunTagsIcon(false);
+      } else {
+        setShowRunTagsIcon(true);
+      }
+    };
+    document.addEventListener('mousemove', mousemoveListener);
```
hellendag, line 271: `mousemove` seems a bit more aggressive than necessary -- can you use `onMouseEnter` on the row instead?

js_modules/dagit/packages/core/src/runs/RunTableNew.tsx
```diff
@@ -291,25 +325,55 @@ const RunRow: React.FC<{
...
+            key="runtabletags"
+            onShortcut={() => {
+              setShowRunTags((showRunTags) => !showRunTags);
+            }}
+            shortcutLabel="t"
+            shortcutFilter={(e) => e.code === 'KeyT'}
+          >
+            {null}
+          </ShortcutHandler>
+        ) : null}
```
hellendag, line 376: What does this end up looking like?

js_modules/dagit/packages/core/src/runs/RunTableNew.tsx
```diff
@@ -291,25 +325,55 @@ const RunRow: React.FC<{
...
-                run.tags.find((tag) => tag.key === DagsterTag.Partition),
-              ].filter((x) => x) as TagType[]
-            }
-            mode={isJob ? (run.mode !== 'default' ? run.mode : null) : run.mode}
-            onAddTag={onAddTag}
-          />
+          <Box flex={{direction: 'row', gap: 8, wrap: 'wrap'}}>
+            <Box>
+              <BaseTag
+                fillColor={Colors.Gray100}
```
hellendag, line 331: Is this very different from a regular `Tag` with default intent?

js_modules/dagit/packages/core/src/runs/RunActionsMenu.tsx
```diff
@@ -132,7 +135,21 @@ export const RunActionsMenu: React.FC<{
...
+                      flex={{
+                        justifyContent: 'center',
+                        alignItems: 'center',
+                        display: 'inline-flex',
+                      }}
+                      padding={{vertical: 4, horizontal: 8}}
+                    >
+                      <SlashShortcut>t</SlashShortcut>
+                    </Box>
+                  </div>
```
hellendag, line 151: Could this just be a `<Box flex={{direction: 'row', ...}}>` that contains the text and the `SlashShortcut`?

js_modules/dagit/packages/core/src/runs/RunActionsMenu.tsx
```diff
@@ -51,6 +53,7 @@ import {
   LaunchPipelineReexecutionMutationVariables,
 } from './types/RunUtils.types';
 import {useJobAvailabilityErrorForRun} from './useJobAvailabilityErrorForRun';
+import styled from 'styled-components';
```
hellendag, line 56: Lint should insist that this use `styled-components/macro`

js_modules/dagit/packages/core/src/runs/RunTableNew.tsx
```diff
@@ -326,6 +390,35 @@ const RunRow: React.FC<{
       <td>
         <RunActionsMenu run={run} onAddTag={onAddTag} />
       </td>
+      {showRunTags ? (
```
hellendag, line 393: Shouldn't use this, since `showRunTags` is already the `isOpen` value, and rendering the dialog to null will kill the hide animation.

hellendag (changes requested):

js_modules/dagit/packages/core/src/runs/RunTableNew.tsx
salazarm: Oh it does render the "t" if you hold command

js_modules/dagit/packages/core/src/runs/RunTableNew.tsx
salazarm: I needed a mouseout listener anyways so I used mousemove. I did originally use onMouseEnter on the `tr` but for some reason it wasn't firing so I switched to mousemove. I can try it again maybe I was doing something wrong before?

js_modules/dagit/packages/core/src/runs/RunTableNew.tsx
salazarm: I wanted to stick to the figma colors

js_modules/dagit/packages/core/src/runs/RunActionsMenu.tsx
salazarm: This box is being used to `center` the T specifically.

js_modules/dagit/packages/core/src/runs/RunActionsMenu.tsx
salazarm: I used lint's "import all" and it did this, will fix.

js_modules/dagit/packages/core/src/runs/RunTableNew.tsx
salazarm: can we use Josh's latest colors or should we find what's closest? it changes the aesthetic a bit so I tend to stick to the figma colors

PR #14180, opened 2023-05-09
Remove invalid "trace" log level from dagster dev

gibsondan:
Summary:
I foolishly copied this from the dagit --log-level arg orginally, not realizing that "trace" is a uvicorn-specific log level.

Test Plan:
dagster dev -h
dagit -h
dagster-daemon run -h


smackesey (approved):
👍

PR #14183, opened 2023-05-09
[ui] Hide upstream data missing warning for partition-mapped backfills

bengotow:
## Summary & Motivation

This PR fixes a bug in the "upstream data missing" warning - it should only consider upstream assets that share the same partitioning as the root assets in your selection. Previously, selecting an asset to materialize that had partition-mapped upstream assets would result in this warning always appearing, since the upstream assets did not have the partition keys being selected.

## How I Tested These Changes

I tested this using the `mapped` asset group in the toys repo.
- Select asset with an upstream of the same partitioning, verify that the warning appears
- Select asset with an upstream of different partitioning, verify that the warning does not appear

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
+++ b/js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
@@ -50,7 +50,7 @@ import {
   showBackfillErrorToast,
   showBackfillSuccessToast,
   UsingDefaultLauncherAlert,
-  USING_DEFAULT_LAUNCH_ERALERT_INSTANCE_FRAGMENT,
+  USING_DEFAULT_LAUNCHER_ALERT_INSTANCE_FRAGMENT,
 } from '../partitions/BackfillMessaging';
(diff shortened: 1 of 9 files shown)


js_modules/dagit/packages/core/src/assets/LaunchAssetChoosePartitionsDialog.tsx
```diff
@@ -727,29 +742,45 @@ export const LAUNCH_ASSET_CHOOSE_PARTITIONS_QUERY = gql`
...
+  const instance = warningsResult.data?.instance;
+  const upstreamAssets = warningsResult.data?.assetNodes;
+  const upstreamAssetKeysSamePartitioning = React.useMemo(
+    () =>
+      (upstreamAssets || [])
+        .filter(
+          (a) =>
+            a.partitionDefinition &&
+            displayedPartitionDefinition &&
+            partitionDefinitionsEqual(a.partitionDefinition, displayedPartitionDefinition),
```
bengotow, line 775: Did a bit of renaming in this PR, this is the key change

salazarm (approved):

PR #14185, opened 2023-05-09
Asset policy evaluations table

johannkm:
Storage for asset daemon evaluations scoped to individual assets. Each row has

- id (just a primary key)
- evaluation id (the asset daemon will use consecutive ids for each iteration. We can use this to show the most recent iteration, and to know in which iterations this asset wasn't considered)
- asset key
- num materialized (pulling this out of the serialized object so we can optimize queries to find evaluations that resulted in runs

Check that index gets used for the planned read query:
```
# explain select * from asset_daemon_evaluations where asset_key='foo' order by evaluation_id;
                                                       QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=11.31..11.31 rows=3 width=92)
   Sort Key: evaluation_id
   ->  Bitmap Heap Scan on asset_daemon_evaluations  (cost=4.17..11.28 rows=3 width=92)
         Recheck Cond: (asset_key = 'foo'::text)
         ->  Bitmap Index Scan on idx_asset_daemon_evaluations_asset_key_evaluation_id  (cost=0.00..4.17 rows=3 width=0)
               Index Cond: (asset_key = 'foo'::text)
```

Test plan:

`dagster instance migrate` succeeded

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/alembic/versions/039_d3a4c9e87af3_add_asset_policy_evaluations_table.py
+++ b/python_modules/dagster/dagster/_core/storage/alembic/versions/039_d3a4c9e87af3_add_asset_policy_evaluations_table.py
@@ -1,0 +1,61 @@
+"""Add asset policy evaluations table
+
+Revision ID: d3a4c9e87af3
+Revises: 701913684cb4
+Create Date: 2023-05-09 11:50:38.931820
+
+"""
+
+import sqlalchemy as db
+from alembic import op
+from dagster._core.storage.migration.utils import has_index, has_table
+from dagster._core.storage.sql import get_current_timestamp
+from sqlalchemy.dialects import sqlite
+
+# revision identifiers, used by Alembic.
+revision = "d3a4c9e87af3"
+down_revision = "701913684cb4"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    if not has_table("asset_daemon_evaluations"):
+        op.create_table(
+            "asset_daemon_evaluations",
+            db.Column(
+                "id",
+                db.BigInteger().with_variant(sqlite.INTEGER(), "sqlite"),
+                primary_key=True,
+                autoincrement=True,
+            ),
+            db.Column(
+                "evaluation_id",
+                db.BigInteger().with_variant(sqlite.INTEGER(), "sqlite"),
+                index=True,
+            ),
+            db.Column("asset_key", db.Text),
+            db.Column("asset_evaluation_body", db.Text),
+            db.Column("num_requested", db.Integer),
+            db.Column("num_skipped", db.Integer),
+            db.Column("num_discarded", db.Integer),
+            db.Column("create_timestamp", db.DateTime, server_default=get_current_timestamp()),
+        )
+        op.create_index(
+            "idx_asset_daemon_evaluations_asset_key_evaluation_id",
+            "asset_daemon_evaluations",
+            ["asset_key", "evaluation_id"],
+            mysql_length={"asset_key": 64},
+            unique=True,
+        )
+
+
+def downgrade():
+    if has_index(
+        "asset_daemon_evaluations", "idx_asset_daemon_evaluations_asset_key_evaluation_id"
+    ):
+        op.drop_index(
+            "idx_asset_daemon_evaluations_asset_key_evaluation_id", "asset_daemon_evaluations"
+        )
+    if has_table("asset_daemon_evaluations"):
+        op.drop_table("asset_daemon_evaluations")
--- a/python_modules/dagster/dagster/_core/storage/schedules/schema.py
+++ b/python_modules/dagster/dagster/_core/storage/schedules/schema.py
@@ -1,4 +1,5 @@
 import sqlalchemy as db
+from sqlalchemy.dialects import sqlite
 
 from ..sql import MySQLCompatabilityTypes, get_current_timestamp
 
@@ -45,6 +46,27 @@ JobTickTable = db.Table(
     db.Column("update_timestamp", db.DateTime, server_default=get_current_timestamp()),
 )
 
+AssetPolicyEvaluationsTable = db.Table(
+    "asset_daemon_evaluations",
+    ScheduleStorageSqlMetadata,
+    db.Column(
+        "id",
+        db.BigInteger().with_variant(sqlite.INTEGER(), "sqlite"),
+        primary_key=True,
+        autoincrement=True,
+    ),
+    db.Column(
+        "evaluation_id", db.BigInteger().with_variant(sqlite.INTEGER(), "sqlite"), index=True
+    ),
+    db.Column("asset_key", db.Text),
+    db.Column("asset_evaluation_body", db.Text),
+    db.Column("num_requested", db.Integer),
+    db.Column("num_skipped", db.Integer),
+    db.Column("num_discarded", db.Integer),
+    db.Column("create_timestamp", db.DateTime, server_default=get_current_timestamp()),
+)
+
+
 # Secondary Index migration table, used to track data migrations, event_logs and runs.
 # This schema should match the schema in the event_log storage, run schema
 SecondaryIndexMigrationTable = db.Table(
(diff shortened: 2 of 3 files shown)


python_modules/dagster/dagster/_core/storage/schedules/schema.py
```diff
@@ -45,6 +47,24 @@
     db.Column("update_timestamp", db.DateTime, server_default=get_current_timestamp()),
 )
 
+AssetPolicyEvaluationsTable = db.Table(
+    "asset_policy_evaulations",
+    ScheduleStorageSqlMetadata,
+    db.Column(
```
OwenKephart, line 53: what's the difference between id and evaluation_id?

python_modules/dagster/dagster/_core/storage/schedules/schema.py
```diff
@@ -45,6 +47,24 @@
...
+        "id",
+        db.BigInteger().with_variant(sqlite.INTEGER(), "sqlite"),
+        primary_key=True,
+        autoincrement=True,
+    ),
+    db.Column("evaluation_id", db.BigInteger().with_variant(sqlite.INTEGER(), "sqlite")),
+    db.Column("asset_key", db.Text, index=True),
+    db.Column("evaluation_body", db.Text),
+    db.Column("policy_type", db.String(63)),
```
OwenKephart, line 62: do we need policy_type?

OwenKephart: ah I see why we might want to keep track of what the policy was at the time of given evaluation. however, I think to do this completely correctly, we'd need to serialize the whole AutoMaterializePolicy NamedTuple.

this is probably a good idea though -- means that users won't be confused if they update their policy and look at a historical tick that would have made a different decision with the new policy

python_modules/dagster/dagster/_core/storage/alembic/versions/039_d3a4c9e87af3_add_asset_policy_evaluations_table.py
```diff
@@ -0,0 +1,42 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "d3a4c9e87af3"
+down_revision = "701913684cb4"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    if not has_table("asset_policy_evaulations"):
```
sryza, line 23: evaulations

sryza: did you consider `asset_auto_materialize_evaluations`?

freshness policy is another kind of asset policy

python_modules/dagster/dagster/_core/storage/alembic/versions/039_d3a4c9e87af3_add_asset_policy_evaluations_table.py
```diff
@@ -0,0 +1,42 @@
...
+            "asset_policy_evaulations",
+            db.Column(
+                "id",
+                db.BigInteger().with_variant(sqlite.INTEGER(), "sqlite"),
+                primary_key=True,
+                autoincrement=True,
+            ),
+            db.Column("evaluation_id", db.Integer),
+            db.Column("asset_key", db.Text, index=True),
+            db.Column("evaluation_body", db.Text),
```
sryza, line 34: Having the name of this column match the type of the Python objected that's serialized there would make it easier to find one from the other.

python_modules/dagster/dagster/_core/storage/alembic/versions/039_d3a4c9e87af3_add_asset_policy_evaluations_table.py
```diff
@@ -0,0 +1,42 @@
...
+            db.Column("policy_type", db.String(63)),
```
sryza, line 35: A comment for what values this can accept would be helpful.

python_modules/dagster/dagster/_core/storage/alembic/versions/039_d3a4c9e87af3_add_asset_policy_evaluations_table.py
johannkm: Do you think we'll ever have more complex freshness policies that would warrant storing the evaluation info? What about source asset observation policies? Or mayybe future asset expectations? I was thinking that this table could store future things that the asset daemon polls

sryza: One way to think about it is: imagine a world where there was no such thing as an `AutoMaterializePolicy` – all assets were just auto-materialized and users had no control over how. I think we'd still want this table.  I.e. the "policy" is a little bit orthogonal.

I can't think of a reason we'd need this for asset observations, but I suppose we can't rule it out.  The logic for whether to trigger an asset expectation is likely to be more complex than the logic for whether to trigger an asset observation, so I could imagine wanting to record that. We could call it `asset_daemon_evaluations` to be safe? Although then what do we call the table that contains a row per evaluation (not a row per asset)?

python_modules/dagster/dagster/_core/storage/schedules/schema.py
johannkm: id is a primary key assigned by the db
evaluation id is the number for the daemon iteration. It's non-unique

python_modules/dagster/dagster/_core/storage/schedules/schema.py
johannkm: That's actually not what I was thinking of, but you raise a good point. I wonder if we'd need to serialize the policy each time though- we could probably use a snapshot id. I think we could add that in the future though

python_modules/dagster/dagster/_core/storage/alembic/versions/039_d3a4c9e87af3_add_asset_policy_evaluations_table.py
johannkm: I'm ok with asset_daemon_evaluations. @alangenfeld convinced me that for the global record for the daemon, the access pattern matches regular sensor/schedules so we could make it another tick type inside the existing ticks table

python_modules/dagster/dagster/_core/storage/schedules/schema.py
johannkm: My thought was this would store AUTO_MATERIALIZE for each row to future proof the table in the event that we added asset expectations etc. to it in the future. Per discussion here https://github.com/dagster-io/dagster/pull/14185#discussion_r1189245250 maybe it'd be `evaluation_type`?

python_modules/dagster/dagster/_core/storage/alembic/versions/039_d3a4c9e87af3_add_asset_policy_evaluations_table.py
johannkm: discussion about this column here https://github.com/dagster-io/dagster/pull/14185#discussion_r1189173341

python_modules/dagster/dagster/_core/storage/schedules/schema.py
sryza: What precisely is the meaning of "Evaluation" in this Ontology? CC @OwenKephart

johannkm: In Josh's designs the table is labeled evaluations, so I think of an evaluation as equal to one row in the per asset table. Once instance of evaluating whether an asset should materialize or not. Asset daemon evaluations does sound more like there'd be 1 per loop though

sryza (approved):
lgtm!x

sryza (commented):
last thing here: did we make a decision about whether we're going to use the existing ticks table or add a new daemon evaluations table? if the latter, wondering if that changes what we'd want to name this table.

PR #14186, opened 2023-05-09
feat(dbt-assets)[1/N]: add dbt_asset decorator

rexledesma:
## Summary & Motivation
Begin the scaffold for a `dbt_asset` decorator. The decorator takes in a dbt manifest and dbt select statement syntax and creates the corresponding asset definitions for the dbt models. In later diffs in the stack, we provide support for the user to populate the compute function for the decorator. 

## How I Tested These Changes
pytest

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
@@ -1,0 +1,76 @@
+import json
+from dataclasses import dataclass
+from typing import Any, Dict, Mapping
+
+from dagster import AssetKey
+
+from ..asset_utils import default_asset_key_fn, output_name_fn
+
+
+@dataclass
+class DbtManifest:
+    """Helper class for dbt manifest operations."""
+
+    raw_manifest: Dict[str, Any]
+
+    @classmethod
+    def read(cls, path: str) -> "DbtManifest":
+        """Read the file path to a dbt manifest and create a DbtManifest object.
+
+        Args:
+            path(str): The path to the dbt manifest.json file.
+
+        Returns:
+            DbtManifest: A DbtManifest object.
+        """
+        with open(path, "r") as handle:
+            raw_manifest: Dict[str, Any] = json.loads(handle.read())
+
+        return cls(raw_manifest=raw_manifest)
+
+    @property
+    def node_info_by_dbt_unique_id(self) -> Mapping[str, Mapping[str, Any]]:
+        """A mapping of a dbt node's unique id to the node's dictionary representation in the manifest.
+        """
+        return {
+            **self.raw_manifest["nodes"],
+            **self.raw_manifest["sources"],
+            **self.raw_manifest["exposures"],
+            **self.raw_manifest["metrics"],
+        }
+
+    @classmethod
+    def node_info_to_asset_key(cls, node_info: Mapping[str, Any]) -> AssetKey:
+        return default_asset_key_fn(node_info)
+
+    @property
+    def node_info_by_asset_key(self) -> Mapping[AssetKey, Mapping[str, Any]]:
+        """A mapping of the default asset key for a dbt node to the node's dictionary representation in the manifest.
+        """
+        return {
+            self.node_info_to_asset_key(node): node
+            for node in self.node_info_by_dbt_unique_id.values()
+        }
+
+    @property
+    def node_info_by_output_name(self) -> Mapping[str, Mapping[str, Any]]:
+        """A mapping of the default output name for a dbt node to the node's dictionary representation in the manifest.
+        """
+        return {output_name_fn(node): node for node in self.node_info_by_dbt_unique_id.values()}
+
+    @property
+    def output_asset_key_replacements(self) -> Mapping[AssetKey, AssetKey]:
+        """A mapping of replacement asset keys for a dbt node to the node's dictionary representation in the manifest.
+        """
+        return {
+            DbtManifest.node_info_to_asset_key(node_info): self.node_info_to_asset_key(node_info)
+            for node_info in self.node_info_by_dbt_unique_id.values()
+        }
+
+    def get_node_info_by_output_name(self, output_name: str) -> Mapping[str, Any]:
+        """Get a dbt node's dictionary representation in the manifest by its Dagster output name."""
+        return self.node_info_by_output_name[output_name]
+
+    def get_node_info_by_asset_key(self, asset_key: AssetKey) -> Mapping[str, Any]:
+        """Get a dbt node's dictionary representation in the manifest by its Dagster output name."""
+        return self.node_info_by_asset_key[asset_key]
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorator.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorator.py
@@ -1,0 +1,62 @@
+from typing import Callable, Optional
+
+from dagster import AssetsDefinition, multi_asset
+from dagster._annotations import experimental
+
+from .asset_utils import get_dbt_multi_asset_args, get_deps
+from .cli.resources_v2 import DbtManifest
+from .utils import ASSET_RESOURCE_TYPES, select_unique_ids_from_manifest
+
+
+@experimental
+def dbt_assets(
+    *,
+    manifest: DbtManifest,
+    select: str = "fqn:*",
+    exclude: Optional[str] = None,
+) -> Callable[..., AssetsDefinition]:
+    """Create a definition for how to compute a set of dbt resources, described by a manifest.json.
+
+    Args:
+        manifest (DbtManifest): The wrapper of a dbt manifest.json.
+        select (Optional[str]): A dbt selection string for the models in a project that you want
+            to include. Defaults to "*".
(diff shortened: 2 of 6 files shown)


rexledesma:
* master
  * **PR #14186** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14186" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14186?utm_source=stack-comment).

python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorators.py
```diff
@@ -0,0 +1,86 @@
...
+) -> Callable[..., AssetsDefinition]:
+    unique_ids = select_unique_ids_from_manifest(
+        select=select, exclude=exclude or "", manifest_json=manifest
+    )
+    dbt_nodes = DbtManifest(manifest).dbt_unique_id_to_nodes
+
+    deps = get_deps(
+        dbt_nodes,
+        selected_unique_ids=unique_ids,
+        asset_resource_types=ASSET_RESOURCE_TYPES,
```
rexledesma, line 64: I don't feel great about having this here, since we're making a decision for the user on what deserves to be an `AssetKey`. I think ideally, this is a choice that the user should make.

python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorators.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+@dataclass
+class DbtManifest:
+    """Helper class for dbt manifest operations."""
+
+    manifest: dict
+
+    @cached_property
+    def dbt_unique_id_to_nodes(self) -> Mapping[str, dict]:
```
sryza, line 19: Naming: we typically name dictionaries with the convention `nodes_by_dbt_unique_id`.

Also, is this an improper use of "node" in dbt nomenclature, given that only a subset of these nodes are `**self.manifest["nodes"]`?

Also also, in the past we've used "node_info" instead of "node". Thoughts on one vs. the other? I think there's value in consistency.

python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorators.py
```diff
@@ -0,0 +1,86 @@
...
+
+from dagster import AssetsDefinition, multi_asset
+from dagster._core.definitions.events import AssetKey
+
+from .asset_utils import default_asset_key_fn, get_dbt_multi_asset_args, get_deps
+from .utils import ASSET_RESOURCE_TYPES, output_name_fn, select_unique_ids_from_manifest
+
+
+@dataclass
+class DbtManifest:
```
OwenKephart, line 13: Is this class foreshadowing future utility functions that would take a DbtManifest as input? If so, what are patterns you'd imagine using it in?

It seems like it's not necessary for this PR (just because only the `dbt_unique_id_to_nodes` property is used, so it might be better to punt on this thing until it's needed.

python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorators.py
OwenKephart: yeah this is a good callout. my thinking is that the desired `asset_resource_types` are highly correlated to the commands that you're running inside the compute_fn.

this is how it works right now (if you're running `dbt build`, then asset resource types = the set of things that dbt build runs, if you're running `dbt build`, then asset resource types = just models).

I also think that the command you're running is highly correlated to what actually exists in your project. That is, if you have seeds and snapshots, you're much more likely to use `dbt build` than `dbt run`, or in this new world, potentially run all of `dbt run` ,  `dbt seed`, and `dbt snapshot`.

In the (theoretically somewhat rare) case that people do have (e.g.) snapshots, but don't want them to be loaded as Dagster assets, it's easy enough to have `exclude="resource:snapshot"`, so I think this would be an acceptable way to configure this behavior.

OwenKephart: the one potential exception here would be dbt tests, which are executed, but are not represented as assets. In theory, someone might want to alter that behavior, and modifying select/exclude would not help them there. I don't think we need to worry about this in the short term though.

OwenKephart (commented):
overall, I think this looks solid

python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorators.py
rexledesma: I'm imagining that the user will use this class for customization:

- When customizing the asset keys of the `AssetsDefinition` after invoking `@dbt_assets`. Here, we provide them a way to access the node info so they can map it however they please (e.g. to generate a custom asset key).
- Also, this allows the user to have access to the manifest if they want to customize the body of their computation (e.g. adding arbitrary metadata to their `AssetMaterialization`/`AssetObservation`/`Output`), using either the asset key or output.
- I am also envisioning a similar class for dbt's Run Results artifact.

sryza (commented):
Is my understanding correct that `yield from dbt.cli(["run"])` doesn't yet work?

My only remaining reservation about this API is about that, and I'm wondering whether this is the right place to give feedback or somewhere else is better.

TLDR my reservation is that I think that users might find the behavior of a generator returned by `dbt.cli` unintuitive:
- The first time that `next` is invoked on the generator, it will invoke the dbt CLI to run all the dbt models, but just return an `AssetMaterialization` that corresponds to one of the models.
- Subsequent invocations of `next` won't impact what dbt models get run, they'll just return `AssetMaterialization`s.

I think that an API that somehow distinguishes between "kicking off computations" and "streaming back the results" might be easier to understand and harder to misuse.  Haven't thought about the names much, but maybe something like:

```python
yield from dbt.cli(...).stream_results()
```

Everything else in this PR looks good to me, although I will admit to not being an expert on the guts of the dbt integration.

schrockn:
It doesn't stream the results back? I thought we parsed logs as they come in and immediately yield events rather than waiting for the run command to complete

sryza:
> It doesn't stream the results back? I thought we parsed logs as they come in and immediately yield events rather than waiting for the run command to complete

The assets returned by `load_assets_from_dbt_project` do this, but `DbtCliClient.cli` returns a `DbtCliOutput`, not a generator.

rexledesma:
> It doesn't stream the results back? I thought we parsed logs as they come in and immediately yield events rather than waiting for the run command to complete

This is new that we support as of `dbt-core==1.4`. We automatically determine if we can do this [on behalf of the user](https://sourcegraph.com/github.com/dagster-io/dagster/-/blob/python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py?L304-321).

In the resource though, they are currently two entry points that the user must explicitly choose between if they want this behavior: `.cli` (batch) and `.cli_stream_json` (streaming).

If we the user to explicitly to determine the streaming behavior from one entrypoint, we'll need it to return some object that supports something like `.wait()` and `.stream_results()`.

sryza:
To be clear, what I'm worried about is users doing this and being surprised that `do_stuff_that_should_happen_after_dbt_run` runs before `dbt run` is invoked:

```python
@dbt_multi_asset
def my_dbt_multi_asset(dbt: DbtCliClient):
    result = dbt.cli(["run"])
    do_stuff_that_should_happen_after_dbt_run()
    yield from result
```

python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorators.py
OwenKephart: My feeling is that if the main usecase of this class is to help users translate between dbt representations and dagster representations of an asset, we'd want to bundle it together with the asset key generating function. there are a few independent places where we'd like to be able to do this translation:

- doing the `with_attributes` call after `@dbt_assets`
- inside the compute_fn, to potentially do fancy things
- when generating AssetObservations for tests (we need to covert the upstream unique ids for the test into upstream asset keys for the observations)
- when generating an asset selection via `DbtManifestAssetSelection`

There's also a particularly rough edge when you involve key prefixes, as these can get applied post-hoc, meaning there's no automatic way of ensuring these callsites are in sync with whatever key prefix has been added. Not sure if there's any perfect solution for that bit, but it's something to be aware of

Anyway, I think it'd be an ok outcome if the user was able to create their own subclass of this object to supply their own definition for `get_asset_key_for_unqiue_id` or something, I just feel like we probably wouldn't want to call this object `DbtManifest` in that case, as it's more of a translation layer

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -0,0 +1,61 @@
...
+from functools import cached_property
+from typing import Any, Dict, Mapping
+
+from dagster import AssetKey
+
+from ..asset_utils import default_asset_key_fn, output_name_fn
+
+
+@dataclass
+class DbtManifest:
```
OwenKephart, line 11: I think we should provide a static constructor for this object, so that people can do `DbtManifest.from_path`, or `DbtManifest.from_dict`. This saves them from having to do the whole with open(...): json.loads(...) stuff on their own, and also allows us (in the future) to take advantage of the `WriteableManifest.from_file` or whatever it's called in dbt land

OwenKephart (approved):

PR #14189, opened 2023-05-09
Fix freshness calculation comparison error

OwenKephart:
## Summary & Motivation

Adding some typing to the relevant dictionary reveals the issue. This retains the desired behavior of having `None` win out in cases where you want to take the minimum of a datetime and None


alangenfeld:
how hard would it be to add a test that would have caught this?

sryza (approved):

Issue #13873, opened 2023-04-24
Custom name for Op created by load_assets_from_dbt_* methods


wkeifenheim:
### What's the use case?

Some patterns of using `dbt` fail to load in Dagster due to non-unique Op names generated by the `dagster_dbt` library.

I have many different `dbt` profiles, each targeting a distinct client domain (i.e., database) to run the _same_ model code for each client. When supplying configuration to invocations of `load_assets_from_dbt_*` method, all result in the same underlying Op name _unless_ you are applying a select/exclude filter that differs between each.

![image](https://user-images.githubusercontent.com/639361/234098651-874eb55f-c2bd-4d68-8429-ed9867632391.png)

These Op names are [auto-generated](https://github.com/dagster-io/dagster/blob/1.3.1/python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py#L382-L385) using a hash of the `select` and `exclude` filters, but not `target` or `profile`.

My runs/builds of `dbt` primarily differ in which `target` is selected, and these fail to load because we cannot have non-unique Op names in a `Definitions` object.

The current suggestion is to pre-compile all the needed `dbt manifests` and load using `load_assets_from_dbt_manifest`, however we still need a way to make each Op (generated from each manifest) always have a unique name.

### Ideas of implementation

Two options, not necessarily mutually exclusive and both may have enough value to pursue.

1. Add a `name` parameter to the `load_assets_from_dbt_*` methods. This would allow a user to create Op names with more semantic value (i.e. "run_dbt_for_some_target" vs. "run_dbt_ab15d").
2. Include additional CLI parameter fields [`profile`, `target`, `vars`] in the `load_assets_from_dbt_*` methods and include it in the hash method that auto-generates the Op name.

### Additional information

This request stems from a conversation I had with @OwenKephart in Slack.

I am working from a modified version of the `assets_dbt_python` example, happy to share that code if it's helpful.

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

rexledesma:
Resolved with https://github.com/dagster-io/dagster/pull/14190

PR #14190, opened 2023-05-09
Custom Op name for load_assets_from_dbt_*

wkeifenheim:
## Summary & Motivation
See issue  #13873

This PR does not include capturing other CLI parameters  [`profile`, `target`, `vars`] towards dynamically generating op names.

## How I Tested These Changes
The `test_op_custom_name` test builds two asset collections from dbt, each with a different target, and no other differences (of dbt config). Loading from the different manifests while (distinctly) setting the new `op_name` parameter yields `op` names that do not collide.

_Example of one custom op name from the dbt load, within the `AssetsDefintion` object_
image

I have also manually tested this by modifying the `assets_dbt_python` example to load multiple asset groups of the dbt project. Each group targets a different database destination, and in this example consume from a common source. The multiple instances of the dbt project worked as intended. Without the custom `op_name` field we would fail once the different dbt asset groups are added to `Definitions`.
image


smackesey:
cc @OwenKephart

python_modules/libraries/dagster-dbt/dagster_dbt_tests/test_asset_defs.py
```diff
@@ -655,6 +656,37 @@ def test_op_config(
     assert all_keys == expected_keys
 
 
+@pytest.mark.parametrize("instances", [[{"target": "target_a"}, {"target": "target_b"}]])
+def test_op_custom_name(instances, test_project_dir, dbt_config_dir, dbt_target_dir):
```
OwenKephart, line 660: I think this test could be simplified a ton, just by having all of the load_assets_from_dbt_manifest calls share the same manifest.json file (sample_manifest.json would work).

as long as the keys are distinct (which is handled by the key_prefix already), the actual contents of the manifest shouldn't matter for the purposes of this test.

also, as there's technically only one set of instances, we can probably skip parameterizing the test and just hard-code those in.

python_modules/libraries/dagster-dbt/dagster_dbt_tests/test_asset_defs.py
```diff
@@ -655,6 +656,37 @@ def test_op_config(
...
+                    manifest_json=json.load(manifest_json),
+                    key_prefix=[instance['target'], "duckdb", "test-schema"],
+                    source_key_prefix=["duckdb"],
+                    op_name=f"{instance['target']}_dbt_op",
+                    node_info_to_group_fn=lambda ni: f"{instance['target']}",
+                    dbt_resource_key=f"dbt_{instance['target']}",
+                )
+            )
+    op_names = [asset_group.op.name for asset_group in dbt_assets]
+    assert len(op_names) == len(set(op_names)),\
```
OwenKephart, line 684: a bit of a nitpick, but would prefer multi-line strings to be in this format:

```
assert len(op_names) == len(set(op_names)), (
    f"Multiple instances of a dbt project cannot have the same op name.\n"
    f"dbt targets were: {instances}\n"
    f"op names generated were: {op_names}"
)
```

OwenKephart (changes requested):
thanks for making this change! everything code-wise seems good to me, but had some comments around simplifying the test code / formatting.

also, make sure to run `make black` at the root of the repo before committing.

OwenKephart (approved):
Thanks again for the contribution! Merging this in now, and it'll be in next week's release

PR #14192, opened 2023-05-09
Fix sensor dry run dynamic partitions requests error

clairelin135:
Error from datadog: https://app.datadoghq.com/apm/trace/316573255363055180?spanID=1393001358924365529&spanViewType=errors

The likely cause of this error is that the user code version is older than the dagster cloud version, so the returned sensor execution data object doesn't contain a dynamic partitions requests field. Subsequently we need to add a check in graphQL to handle this field being `None`.


clairelin135:
@sryza Its tricky, I can't think of a straightforward way to test this. I think we would have to mock the response from the sensor within graphQL, which would probably take a good chunk of time to figure out.

sryza (approved):

PR #14193, opened 2023-05-09
Asset policy evaluations storage methods

johannkm:
Insert 1 evaluation id at at time for many assets (for the daemon)
Read 1 asset at a time, with many evaluation ids (for the ui)

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -75,6 +75,9 @@ class AutoMaterializeAssetEvaluation(NamedTuple):
         Sequence[AutoMaterializeCondition],
         Sequence[Tuple[AutoMaterializeCondition, PartitionsSubset]],
     ]
+    num_requested: int
+    num_skipped: int
+    num_discarded: int
 
     @staticmethod
     def from_conditions(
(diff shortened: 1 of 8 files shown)


python_modules/dagster/dagster/_core/storage/schedules/sql_schedule_storage.py
```diff
@@ -446,6 +455,55 @@ def purge_ticks(
...
+        with self.connect() as conn:
+            bulk_insert = AssetPolicyEvaluationsTable.insert().values(
+                [
+                    {
+                        "evaluation_id": evaluation_id,
+                        "asset_key": decision.asset_key.to_string(),
+                        "asset_evaluation_body": serialize_value(decision),
+                        "num_requested": decision.num_requested,
+                    }
+                    for decision in asset_evaluations
```
sryza, line 473: Nitpick: "decision" isn't a term we're using elsewhere, right?

python_modules/dagster/dagster/_core/storage/schedules/sql_schedule_storage.py
```diff
@@ -446,6 +455,55 @@ def purge_ticks(
...
+                ]
+            )
+            conn.execute(bulk_insert)
+
+    def get_auto_materialize_asset_evaluations(
```
sryza, line 478: At some point in the nearish future, we're probably going to want to support a view that aggregates evaluations across multiple assets. Would it make sense to plan ahead for that by accepting multiple asset keys here?

johannkm: Offline discussion:
For the aggregate view, we wouldn't want to deserialize the evaluation bodies for each included asset. In a followup, we'll add a separate method that does a groupby evaluation id and sums num_requested, num_skipped, and num_discarded

OwenKephart (dismissed):
I know it's a pain, but this should be rebased off of https://github.com/dagster-io/dagster/pull/14245.

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -75,6 +75,9 @@ class AutoMaterializeAssetEvaluation(NamedTuple):
         Sequence[AutoMaterializeCondition],
         Sequence[Tuple[AutoMaterializeCondition, PartitionsSubset]],
     ]
+    num_requested: int
```
OwenKephart, line 78: ok last naming thing (I promise), but because these things will be annoying to rename once we start writing them to the db, I'm wondering if we should rename `AutoMaterializeDecisionType.MATERIALIZE` to `AutoMaterializeDecisionType.REQUEST`.

This aligns with the naming here, and also indicates what we actually plan to do. If the request fails to materialize the asset (failure or otherwise), then we continue on, we don't continue trying to materialize it.

Minor thing, as this ultimately isn't user facing, so feel free to discard

johannkm: I think AutoMaterializeDecisionType.MATERIALIZE is clearer than AutoMaterializeDecisionType.REQUEST. `num_requested` is just shorthand for `num_materializations_requested`

OwenKephart (approved):
this looks great to me

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
OwenKephart: I buy it

PR #14205, opened 2023-05-10
[wip][dagster-snowflake] case sensitive config

jamiedemaria:

### DIFF
--- a/python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
+++ b/python_modules/libraries/dagster-snowflake/dagster_snowflake/snowflake_io_manager.py
@@ -3,9 +3,7 @@ from contextlib import contextmanager
 from typing import Optional, Sequence, Type, cast
 
 from dagster import IOManagerDefinition, OutputContext, io_manager
-from dagster._config.pythonic_config import (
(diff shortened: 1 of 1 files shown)


mjclarke94:
Hi @jamiedemaria. Just looking through this and it looks relevant to something we were working on recently. I was wondering about the need for pandas dataframe column names to be excluded from the case sensitivity config?

We had a workflow where we loaded in some data using an IOManager, then used that data to inform what other data we needed to fetch from the same table using Ibis to compose some relatively complex queries. The inconsistency between needing upper and lower case column names in different contexts was impacting readability (not to mention the source of about a million bugs!)

jamiedemaria:
Hey @mjclarke94 - i had some higher priority things come up and kinda lost track of this PR. I'm not sure I'll have time to pick it back up in the next couple weeks. If you'd like to take a stab at your own PR for this issue, please feel free! I'm more than happy to review at any stage of the process and provide any help!

mjclarke94:
Hi @jamiedemaria.

Sorry, probably isn't something I'll have the bandwidth for in the next few weeks either. Not having this isn't a blocker for us by any stretch, just thought I'd drop the above discussion point in for whenever you revisit this next.

PR #14208, opened 2023-05-10
[ui] Reimplement Gantt animation loop

hellendag:
## Summary & Motivation

Reimplement Gantt chart animation to use `requestAnimationFrame`. Hopefully this helps resolve some of the weird animation loop issues we currently see.

## How I Tested These Changes

Kick off some runs, verify that the Gantt chart animates smoothly. Use the React profiler to ensure that it's not too expensive to use rAF for this.

I should check this on some heavy jobs with lots of concurrent steps, and I could use some help hammering away at this to make sure everything works as expected even in edge cases.

### DIFF
--- a/js_modules/dagit/packages/core/src/gantt/GanttChart.tsx
+++ b/js_modules/dagit/packages/core/src/gantt/GanttChart.tsx
@@ -229,6 +229,7 @@ const GanttChartInner = (props: GanttChartInnerProps) => {
   const [hoveredTime, setHoveredTime] = React.useState<number | null>(null);
   const [nowMs, setNowMs] = React.useState<number>(() => props.overrideNowTime || Date.now());
   const {options, metadata, selection} = props;
+  const animationRequest = React.useRef<number | null>(null);
 
   const {rootServerURI} = React.useContext(AppContext);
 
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/core/src/gantt/GanttChart.tsx
```diff
@@ -249,32 +250,34 @@ const GanttChartInner = (props: GanttChartInnerProps) => {
...
 
-  // Because renders can happen "out of band" of our update interval, we set a timer for
-  // "time until the next interval after the current nowMs".
+  // When the run is complete, stop the animation. We also do this when the WebSocket is lost,
+  // since we would just be animating endlessly with no new logs.
   React.useEffect(() => {
-    if (scale === 0 || lostWebsocket) {
-      return;
+    if (scale === 0 || lostWebsocket || exitedAt) {
+      animationRequest.current && cancelAnimationFrame(animationRequest.current);
```
bengotow, line 264: Wow I actually didn't know cancelAnimationFrame existed!

bengotow (approved):
This looks good to me! I think that for long-running runs, I was trying to be fancy here and only re-render when the graph moved perceptibly (which could be every 5 seconds or so if the job has been running for 30min). Rendering on every frame might be a bit overkill in that case, but I like that this is simpler (and probably not buggy... 🤦‍♂️)

hellendag:
Tested on `megadags.py` and it seems to animate nicely.

PR #14215, opened 2023-05-10
[Cancel asset backfill 1/2] Enable cancel backfill buttons in UI

clairelin135:
This PR is part 1 of 2 for enabling cancellations for asset backfills. The following buttons are now displayed in the backfill menu, bringing asset backfill cancellation functionality up to parity with job backfill cancellation:
- `Cancel backfill submission`, which marks the backfill as canceled and prevents more runs in the backfill from being created.
- `Terminate unfinished runs`, which marks the unfinished runs in the backfill as canceled.

This involves making changes to two resolvers, adding renames to clarify behavior:
- `numCancelablePartitions`: This resolver now returns back the number of partitions for which runs have not yet been created. This information is used to display a "There are X partitions yet to be queued or launched" confirmation message.
- `backfillRunStatuses`: This resolver now returns back the status for each partitioned run launched from the backfill.

image

When the same partitions are targeted in each asset, the UI displays the bar with missing/in progress/materialized partitions. When different partitions are targeted in each asset, the UI displays counts for in progress/successful/failed partitions.

Test plan: Tested locally and on BK.

Next steps:
- One major gap following this PR is that after the asset backfill is marked as canceled the backfill is not evaluated by the daemon again. This means that asset backfill page will continue to display the state of the backfill at the moment of cancellation, even if the runs have since finished. Part 2 of this stack resolves this gap.
- The status counts assume that each partition name in a backfill is unique. This is to handle run retries, so only the status of the latest run per partition key is displayed. For asset backfills it's possible but unlikely that multiple runs with the same partition key will be generated i.e. a hourly -> daily -> hourly asset graph will kick off multiple runs with the same hourly partition. This PR currently doesn't handle this case, so the status counts will be off. I plan on resolving this in a future TODO.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/implementation/fetch_partition_sets.py
+++ b/python_modules/dagster-graphql/dagster_graphql/implementation/fetch_partition_sets.py
@@ -259,7 +259,7 @@ def get_partition_set_partition_statuses(
 def partition_statuses_from_run_partition_data(
     partition_set_name: Optional[str],
     run_partition_data: Sequence[RunPartitionData],
-    partition_names: Sequence[str],
+    partition_names: Optional[Sequence[str]],
     backfill_id: Optional[str] = None,
 ) -> Sequence["GraphenePartitionStatus"]:
     from ..schema.partition_sets import GraphenePartitionStatus, GraphenePartitionStatuses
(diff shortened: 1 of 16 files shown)


python_modules/dagster-graphql/dagster_graphql/implementation/fetch_partition_sets.py
```diff
@@ -252,49 +253,58 @@ def get_partition_set_partition_statuses(
...
-                id=partition_id,
-                partitionName=name,
-                runId=partition_data.run_id,
-                runStatus=partition_data.status.value,
-                runDuration=partition_data.end_time - partition_data.start_time
-                if partition_data.end_time and partition_data.start_time
-                else None,
-            )
+    status_by_partition = {}
+    for name, partition_data in partition_data_by_name.items():
```
clairelin135, line 282: Is it possible for an asset backfill to request multiple runs with the same partition key?

One situation I can think of is an hourly -> daily -> hourly asset graph kicking off multiple runs for the same hour partition, but I think that these situations are generally pretty rare. Happy to break this out into another PR to resolve in the future.

sryza:
@clairelin135 am I understanding correctly that this change also fills out the "Run status" column for asset backfills? Would it make sense to split that out into its own change?

clairelin135:
@sryza sure, I can split this PR

PR #14219, opened 2023-05-10
Always build API docs

jmsanders:
Previously, we only ran the api doc build if something in the docs dir changed. But the API docs depend on Python changes.

Let's run them on every build since it's otherwise pretty difficult to know where a docstring may exist and where it may not.

### DIFF
--- a/.buildkite/dagster-buildkite/dagster_buildkite/steps/docs.py
+++ b/.buildkite/dagster-buildkite/dagster_buildkite/steps/docs.py
@@ -44,7 +44,6 @@ def build_docs_steps() -> List[BuildkiteStep]:
             # "git diff --ignore-all-space --stat",
             # "git diff --exit-code --ignore-all-space --no-patch",
         )
-        .with_skip(skip_if_no_docs_changes())
         .on_test_image(AvailablePythonVersion.get_default())
         .build(),
         # Verify screenshot integrity.


rexledesma (approved):
any reason why the git diff above is commented out? `make apidoc-build` produces some artifacts, and I am always unsure whether that should be checked in or not

jmsanders:
Probably a @smackesey question - looks like it was added but already commented out in https://github.com/dagster-io/dagster/pull/8960. I'm guessing it's an artifact of the development process.

smackesey:
>any reason why the git diff above is commented out?

IIRC I was trying to make this work so that you would get exit code 0 only if there was no diff-- I think this was in a time where you had to build the docs and include them in your diff to get changes to show in the docs site (now they are automatically built right?). But I couldn't get it to work, I think there was some weird whitespace thing where there would _always_ be a diff between the built docs included in the diff and the docs built on BK. So I commented it out intending to get it to work later but never did.

But now I think even the concept is moot, right? The only thing we care about it in BK is that the docs build with no errors.

rexledesma:
@smackesey Gotcha: if we're automatically building the docs, I don't think we should be including the build artifacts in our repo then. Perhaps we should add them to `.gitignore`?

smackesey:
>if we're automatically building the docs, I don't think we should be including the build artifacts in our repo then. Perhaps we should add them to .gitignore?

That sounds reasonable to me-- cc @yuhan for confirmation that this would be ok

yuhan:
> I was trying to make this work so that you would get exit code 0 only if there was no diff

This was the original intent where we want [the json files](https://github.com/dagster-io/dagster/tree/master/docs/content/api) to be checked in and always up-to-date. But it caused too much trouble in build as Sean mentioned. So we decided to flex that. So it doesn't really matter whether to check in for now.

However, the only dependency is: the master docs (https://docs.dagster.io/master/getting-started) still renders api docs using these files. So we just can't delete them right now. But I'm working on a new and simplified way to handle docs versioning (will share proposal soon) which should result in us not needing to check in the json files at all soon.

All that being said, enabling this test makes total sense to me cos we do want to make sure the apidocs are buildable all the time (not just on docs only changes)

PR #14223, opened 2023-05-11
[WIP][auto-materialize] add max_materializations_per_minute

OwenKephart:
## Summary & Motivation

Adds a `max_materializations_per_minute` parameter to AutoMaterializePolicy, and a corresponding MAX_MATERIALIZATIONS_EXCEEDED `SkipReason`.

To keep behavior consistent with what we'd expect, we add a partition sort key for all time window partitions (not just self dependencies). This makes it so if a ton of time partitions are missing, we'll fill the newest partitions first.

I figured people doing backfills might expect the opposite behavior (fill in the oldest partitions first), rather than the random ordering they get now or the newest -> oldest behavior , so I made this configurable in the ToposortedPriorityQueue. That's probably not necessary though, so I'll likely remove that bit.

This still needs more tests and some polish

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -19,6 +19,7 @@ from typing import (
     cast,
 )
 
+import pendulum
 import toposort
 
 import dagster._check as check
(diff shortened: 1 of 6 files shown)


sryza (commented):
Based on our discussion today, should this not be part of the skip reason?

OwenKephart:
made obsolete by https://github.com/dagster-io/dagster/pull/14254

PR #14224, opened 2023-05-11
[Cancel asset backfill 2/2] Cancel runs within daemon and update asset backfill data until complete

clairelin135:
Following https://github.com/dagster-io/dagster/pull/14244:
- Unfinished runs continue to execute
- The asset backfill page does not update after a backfill is canceled by a user. This is because the daemon stops evaluating backfills with a canceled status, so the backfill page continues to display the status of the materialized/failed asset partitions at the moment of cancellation.

This PR bridges the two gaps above by introducing a new `CANCELING` bulk action status. This status represents asset backfills that have been requested for cancellation by a user, but whose runs have not yet successfully canceled. In this `CANCELING` state, the daemon:
- Does not create or queue new runs
- Marks existing runs as canceled when possible
- Updates asset backfill data with materialized/failed partitions until all runs have concluded
- Marks the backfill is as `CANCELED` when all requested runs have concluded. This allows the asset backfill page to continue to update until all runs have finished.

The reason why the daemon is responsible for canceling the runs is because runs with certain statuses cannot be canceled; only runs in the `STARTED` or `QUEUED` state can be canceled. The other unfinished states (`STARTING`, `NOT_STARTED`) will eventually move to a `STARTED` or `QUEUED` state, allowing them to be canceled. Handling this within the daemon allows us to reevaluate un-cancellable runs at a later time and cancel them when possible.


python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -414,51 +426,137 @@ def execute_asset_backfill_iteration(
...
         # failure, or cancellation). Since the AssetBackfillData object stores materialization states
         # per asset partition, the daemon continues to update the backfill data until all runs have
         # finished in order to display the final partition statuses in the UI.
-        updated_backfill = updated_backfill.with_status(BulkActionStatus.COMPLETED)
 
-    pipeline_and_execution_plan_cache: Dict[int, Tuple[ExternalJob, ExternalExecutionPlan]] = {}
+        # We check if the all runs are finished before updating the asset backfill data.
+        # If we did this the other way around, a race condition would occur where runs finish
+        # between the asset backfill data update and the backfill cancellation.
+        mark_backfill_as_canceled = asset_backfill_data.requested_runs_finished(
```
sryza, line 486: Why do we need to check here for whether the requested runs have finished if we're also monitoring runs inside `get_canceling_asset_backfill_iteration_data`? It seems like we might be able to find out if we're done by checking whether materialized_subset | failed_and_downstream_subset == requested_subset ?

python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -549,11 +647,69 @@ class AssetBackfillIterationResult(NamedTuple):
...
+    yield updated_materialized_subset
+
+
+def get_asset_backfill_iteration_failed_and_downstream_partitions(
+    backfill_id: str,
+    asset_backfill_data: AssetBackfillData,
+    asset_graph: ExternalAssetGraph,
+    instance_queryer: CachingInstanceQueryer,
+    backfill_start_time: datetime,
+):
```
sryza, line 692: Return type plz

python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -549,11 +647,69 @@ class AssetBackfillIterationResult(NamedTuple):
...
+        yield None
+
+    updated_materialized_subset = (
+        asset_backfill_data.materialized_subset | recently_materialized_asset_partitions
+    )
+
+    yield updated_materialized_subset
+
+
+def get_asset_backfill_iteration_failed_and_downstream_partitions(
```
sryza, line 686: Calling this `_get_failed_and_downstream_asset_partitions` would make it more consistent with `_get_failed_asset_partitions`. Or we could change the name of that function. Either way, I don't think we need the word "iteration" in the title - in theory this function could be used outside the context of an iteration.

python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -414,51 +426,137 @@ def execute_asset_backfill_iteration(
...
+                run_request=run_request,
+                asset_graph=asset_graph,
+                # create a new request context for each run in case the code location server
+                # is swapped out in the middle of the backfill
+                workspace=workspace_process_context.create_request_context(),
+                instance=instance,
+                pipeline_and_execution_plan_cache=pipeline_and_execution_plan_cache,
+            )
+
+    elif backfill.status == BulkActionStatus.CANCELING:
```
sryza, line 477: Thoughts on pushing this logic down into `execute_asset_backfill_iteration_inner`?

The way I conceived the difference between `execute_asset_backfill_iteration_inner` and `execute_asset_backfill_iteration` is that `execute_asset_backfill_iteration_inner` includes all the "smarts", without doing any side effects, and `execute_asset_backfill_iteration` carries out the side effects, i.e. submitting runs, updating the backfill record in the database.

Maintaining this separation makes it possible to test the smarts without dealing with the side effects.

It seems like the cancellation logic doesn't have any side effects, so could go in the inner function?

sryza (commented):
The approach here makes sense. The naming of these statuses hurts my brain a little bit. CANCELING feels like an appropriate name if we're actively trying to cancel the runs within the backfill. If we're just hanging around and letting them complete, I think a different name would make more sense.

python_modules/dagster/dagster/_core/execution/asset_backfill.py
clairelin135: good call. for some reason the fact that we had a requested subset as well slipped my mind

clairelin135:
Per discussion in slack, I will update this PR to instead terminate the existing runs when the backfill is marked as canceled.

python_modules/dagster/dagster/_core/execution/asset_backfill.py
clairelin135: There's been some changes to this function since you last reviewed, I think now there are side effects--runs are canceled, and the bulk action status is updated.

Let me know if you'd still prefer a refactor here

clairelin135:
@sryza this PR has been updated to handle run cancellation within the daemon (more details in the PR description). If you have time to look at this before you leave for PTO that would be great, otherwise no worries.

I'm still working on putting together more tests for this--it's surprisingly tricky since apparently pytests run in a single thread, so it's hard to emulate a run cancellation occurring during a run.

clairelin135:
@prha @johannkm I think this PR is ready for you to take a look. This now adds a `CANCELING` bulk action status where the daemon cancels existing runs in the backfill iteration (more details in the PR description)

python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -414,49 +439,136 @@ def execute_asset_backfill_iteration(
...
+
+        updated_backfill = backfill.with_asset_backfill_data(
+            updated_asset_backfill_data, dynamic_partitions_store=instance
+        )
+        # The asset backfill is successfully canceled when all requested runs have finished (success,
         # failure, or cancellation). Since the AssetBackfillData object stores materialization states
         # per asset partition, the daemon continues to update the backfill data until all runs have
         # finished in order to display the final partition statuses in the UI.
-        updated_backfill = updated_backfill.with_status(BulkActionStatus.COMPLETED)
+        if updated_asset_backfill_data.have_all_requested_runs_finished():
```
johannkm, line 524: what is the ui like while we're waiting? Can I see which runs haven't finished and go force terminate them myself if necessary?

johannkm (approved):

PR #14240, opened 2023-05-11
Eliminate places where we are calling capture_error multiple times in a single request

gibsondan:
Summary:
Prevent nested capture_error calls in a single graphql resolver by moving all capture_error decorators to the top-level resolver instead of helper functions

Test Plan: Existing graphql coverage should cover all endpoints that have a capture_error in them today

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/implementation/execution/__init__.py
+++ b/python_modules/dagster-graphql/dagster_graphql/implementation/execution/__init__.py
@@ -23,7 +23,6 @@ if TYPE_CHECKING:
 from ..utils import (
     assert_permission,
     assert_permission_for_location,
-    capture_error,
 )
 from .backfill import (
     cancel_partition_backfill as cancel_partition_backfill,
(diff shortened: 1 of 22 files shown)


python_modules/dagster-graphql/dagster_graphql/implementation/utils.py
```diff
@@ -117,18 +117,25 @@ def watch(fn: Callable[[Exception], None]) -> Iterator[None]:
...
+    capturing_error = False
+
 
 def capture_error(
     fn: Callable[P, T]
 ) -> Callable[P, Union[T, "GrapheneError", "GraphenePythonError"]]:
     def _fn(*args: P.args, **kwargs: P.kwargs) -> T:
         try:
+            if ErrorCapture.capturing_error:
+                raise Exception("capture_error invoked more than once within a single graphql call")
```
gibsondan, line 129: another option would be skipping the middleware if its already been enabled? maybe with a warning?

gibsondan (commented):
ended up just moving each of these up to the resolver declaration - which makes it much less likely that you would end up with two of them within a single resolver, so i dunno that we need to do anything fancy to detect it now

alangenfeld (approved):
im fine with this - update the PR title & summary before landing

PR #14241, opened 2023-05-11
[Cancel asset backfill 1/3] Display asset backfill run statuses

clairelin135:
This PR enables displaying:
- The run statuses for runs requested within an asset backfill
- A "terminate unfinished runs" button in the backfill menu, which marks all unfinished runs as canceled

In order to support this, the `backfillRunStatuses` resolver (renamed from `partitionStatuses` for clarity) now returns back the status for each partitioned run launched from an asset backfill.

One assumption here is that all runs with the same partition key target the same assets. This is to handle run retries, so only the status of the latest run per partition key is displayed. 

For asset backfills it's possible that multiple runs will target the same partition key but different assets i.e. for a hourly -> daily -> hourly asset graph a run for partition X will be generated for each hourly asset. Subsequently, only the latest run for partition X would be displayed in the counts. I think that this case is fairly unlikely so I haven't handled this case yet, but if we wanted to handle it it would involve deduping by the asset selection per partitioned run.

When each asset in the backfill does not target the same partitions, we display the counts:

image

When each asset targets the same partitions, we display the row with missing/materialized/in progress partitions:

image

### DIFF
--- a/js_modules/dagit/packages/core/src/graphql/schema.graphql
+++ b/js_modules/dagit/packages/core/src/graphql/schema.graphql
@@ -2430,7 +2430,7 @@ type PartitionBackfill {
   runs(limit: Int): [Run!]!
   unfinishedRuns(limit: Int): [Run!]!
   error: PythonError
-  partitionStatuses: PartitionStatuses!
+  backfillRunStatuses: PartitionStatuses!
   partitionStatusCounts: [PartitionStatusCounts!]!
   isAssetBackfill: Boolean!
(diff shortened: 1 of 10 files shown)


clairelin135:
Thinking about this more, I think it might be more common than I initially thought that multiple runs target the same partition key but different assets. I.e. a daily -> weekly asset graph would generate runs for both the daily and weekly asset with partition key "2023-05-07". Going to rethink through this PR and re-request when it's ready for review.

PR #14244, opened 2023-05-11
[Cancel asset backfill 1/2] Enable cancel backfill button

clairelin135:
This PR adds support for canceling an asset backfill:
- Updates the `numCancelable` resolver to return the number of partitions for which runs have not yet been created. This is used to display a "There are X partitions yet to be queued or launched" confirmation message.
- Validates permissions and updates the backfill's bulk action status to be `CANCELED`.

One gap following this PR is that after the asset backfill is marked as canceled the backfill is not evaluated by the daemon again. This means that asset backfill page will continue to display the state of the backfill at the moment of cancellation, even if the runs have since finished. Part 3 of this stack resolves this gap: https://github.com/dagster-io/dagster/pull/14224

image

### DIFF
--- a/js_modules/dagit/packages/core/src/graphql/schema.graphql
+++ b/js_modules/dagit/packages/core/src/graphql/schema.graphql
@@ -2433,7 +2433,7 @@ type PartitionBackfill {
   runs(limit: Int): [Run!]!
   unfinishedRuns(limit: Int): [Run!]!
   error: PythonError
-  partitionStatuses: PartitionStatuses!
+  partitionStatuses: PartitionStatuses
   partitionStatusCounts: [PartitionStatusCounts!]!
   isAssetBackfill: Boolean!
   assetBackfillData: AssetBackfillData
--- a/js_modules/dagit/packages/core/src/graphql/types.ts
+++ b/js_modules/dagit/packages/core/src/graphql/types.ts
@@ -2382,7 +2382,7 @@ export type PartitionBackfill = {
   partitionSet: Maybe<PartitionSet>;
   partitionSetName: Maybe<Scalars['String']>;
   partitionStatusCounts: Array<PartitionStatusCounts>;
-  partitionStatuses: PartitionStatuses;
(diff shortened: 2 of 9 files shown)


sryza:
> Updates the numCancelablePartitions resolver (renamed from numCancelable for clarity)

In the case that the backfill targets a set of assets with the same partitions, I think it's still unclear what this means – i.e. does it refer to the number of partitions selected by the person launching the backfill or the number of partitions times the number of assets? My initial guess was actually the former, although I see the logic for the latter.

I think that, if we wanted to be maximally clear, we would have two separate fields - `numCancelablePartitions` and `numCancelableAssetPartitions`, and the message for asset backfills would say something like "There are X asset partitions and un-partitioned assets yet to be queued or launched".

That said, I'm not necessarily saying that it's worth doing this. What would be the consequence of just omitting the "There are X partitions yet to be queued or launched" message in the case that we're dealing with an asset backfill? That information is already accessible in the asset backfill UI.

clairelin135:
I don't think it's worth it to create multiple resolvers just to handle this case. We're pretty much just checking that if `numCancelable > 0` we enable canceling the backfill and display the modal.

I'd prefer to just update the message in the modal for asset backfills to be something like what you wrote above: "There are X asset partitions and un-partitioned assets yet to be queued or launched". And I can update the docstring in the resolver if it isn't clear.

sryza:
> I don't think it's worth it to create multiple resolvers just to handle this case. We're pretty much just checking that if numCancelable > 0 we enable canceling the backfill and display the modal.

Cool, will defer to you on that. In that case, I think `numCancelable` might actually be a better name than `numCancelablePartitions`. The ambiguity of the former more flexibly handles the different meanings that we're assigning it in different circumstances.  E.g. omitting the word "partitions" makes it more appropriate to include non-partitioned assets in the count.

clairelin135:
ok, will revert the rename

sryza (approved):

PR #14245, opened 2023-05-11
[auto-materialize] refactor reasons

OwenKephart:
## Summary & Motivation

Now that we have 3 different states that a condition may belong to, the "separate enum and namedtuple and field per state" thing was getting pretty ugly, as we were using the result ("MATERIALIZE", "SKIP", or "DISCARD") as a top level concept which determined class names.

This refactors thing to make everything built on top of these results, rather than vice versa. So we have an enum indicating the type of outputs we can get for a decision, then an enum indicating the types of conditions that can lead to those results, then finally a NamedTuple that can bundle this condition with extra metadata.

In general, this results in way less code than the alternative solution, and also makes it easier to add new results / conditions.


python_modules/dagster/dagster/_core/definitions/auto_materialize_reason.py
```diff
@@ -0,0 +1,75 @@
+from enum import Enum
+from typing import NamedTuple
+
+
+class AutoMaterializeResult(Enum):
```
sryza, line 5: What do you think about calling this `AutoMaterializeResultType`? I did a grep for Enum in our codebase, and it seems like they basically all end in either "Type" or "Status".

python_modules/dagster/dagster/_core/definitions/auto_materialize_reason.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class AutoMaterializeCondition(Enum):
+    """Represents the set of conditions that can impact auto-materialization for an asset."""
+
+    def __init__(self, id: str, result: AutoMaterializeResult):
+        self.id = id
+        self.result = result
+
+    # Materialization Reasons
```
sryza, line 27: Is this use of "Reason" congruous with the name `AutoMaterializeConditionReason`?

python_modules/dagster/dagster/_core/definitions/auto_materialize_reason.py
```diff
@@ -0,0 +1,75 @@
...
+    MISSING = ("MISSING", AutoMaterializeResult.MATERIALIZE)
+
+    # Skip Reasons
+    PARENT_OUTDATED = ("PARENT_OUTDATED", AutoMaterializeResult.SKIP)
+
+    # Discard Reasons
+
+
+class AutoMaterializeConditionReason(NamedTuple):
+    """Denotes the reason that the auto-materialize logic decided that an asset should be materialized.
```
sryza, line 40: I'm finding this a little difficult to parse. Is this class only relevant in the case that the `AutoMaterializeResult` of the `AutoMaterializeCondition` is MATERIALIZE? From the name, I would expect it to apply to any `AutoMaterializeCondition`.

python_modules/dagster/dagster/_core/definitions/auto_materialize_reason.py
OwenKephart: yeah makes sense

python_modules/dagster/dagster/_core/definitions/auto_materialize_reason.py
OwenKephart: should be `"Materialize Conditions"`

python_modules/dagster/dagster/_core/definitions/auto_materialize_reason.py
OwenKephart: I just forgot to update this docstring -- this is just a wrapper class that allows (in the future) more specific information to be attached to the underlying condition

OwenKephart: e.g. the "condition" is "DOWNSTREAM_FRESHNESS", and the specific data is "Assets X, Y, and Z needed fresher data", or the "condition" is "PARENT_NEVER_MATERIALIZED", and the specific data is "Upstream asset A has never been materialized"

OwenKephart:
I think "reason" is bad here, as "AutoMaterializeReason" implies "the reason that this asset was auto materialized", whereas a) some of these reasons imply that the asset will NOT be auto materialized and b) multiple of these objects may apply to each asset-partition.

Some alternatives:

AutoMaterializeFactor
AutoMaterializeCriterion
AutoMaterializeInput
AutoMaterializeCondition

Went with Condition for now.

sryza (approved):
Nice

Maybe `AutoMaterializeDecisionCondition`? But is more wordy and might not be better.

PR #14246, opened 2023-05-11
Fixing ConfigurableResource run_config application code path

schrockn:

### DIFF
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -891,17 +891,24 @@ class ConfigurableResourceFactory(
         )
         return out
 
-    def _resolve_and_update_env_vars(self) -> "ConfigurableResourceFactory[TResValue]":
+    def _resolve_and_update_env_vars(self, context) -> "ConfigurableResourceFactory[TResValue]":
         """Processes the config dictionary to resolve any EnvVar values. This is called at runtime
         when the resource is initialized, so the user is only shown the error if they attempt to
         kick off a run relying on this resource.
 
         Returns a new instance of the resource.
         """
+        # This is the configuration passed via run_config (e.g. launchpad or Python API)
+        incoming_resource_config = process_config(self._schema.config_type, context.resource_config)
+
+        # this is the data that was passed to the resource instance
         post_processed_data = _process_config_values(
             self._schema, self._resolved_config_dict, self.__class__.__name__
         )
-        return self._with_updated_values(post_processed_data)
+
+        return self._with_updated_values(
+            {**post_processed_data, **(incoming_resource_config.value or {})}
+        )
 
     @contextlib.contextmanager
     def _resolve_and_update_nested_resources(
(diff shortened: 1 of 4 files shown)


schrockn:
There is still a little cleaning up to do this in this PR, but I wanted to make sure we agreed that the semantic here are correct. This is also extremely confusing behavior so I think a potential hotfix should be in the conversation here.

schrockn:
> If no default is supplied, but the value is required eg Does this still fail at code location load time?. 

Yes that will still fail at code location load time

> Not sure I follow the comment in your test, are you just saying we shouldn't be able to pass resources to execute_in_process but only run_config?

You can still pass resources to `execute_in_process`. This does not change that.

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -891,17 +891,29 @@ def _with_updated_values(
...
+
+        This is called at runtime when the resource is initialized, so the user is only shown
+        the errors involving environment variable resolution if they attempt to kick off a
+        run relying on this resource.
 
         Returns a new instance of the resource.
         """
+        # This is the configuration passed via run_config (e.g. launchpad or Python API)
+        incoming_resource_config = process_config(self._schema.config_type, context.resource_config)
+
```
schrockn, line 908: This poses a real problem in the case of enums. This gets post-processed with actually converts the enum to be the value, rather than the name, of the enum value. When that in turn is passed back to the ConfigurableResourceFactory constructor it does another roundtrip through the config validator, which fails the validation.

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -891,17 +891,46 @@ def _with_updated_values(
...
+        # enums break # that assumption. (Environment variable resolution for example works
+        # fine because the initial evaluation resolves the env var to its value, and subsequent
+        # evaluations are idempotent)
+
+        incoming_resource_config_evr = process_config(
+            self._schema.config_type, context.resource_config
+        )
+
+        # this is the data that was passed to the resource instance
         post_processed_data = _process_config_values(
```
schrockn, line 927: This call is a problem too because the "default" environment variables get evaluated. 

```
MyResource(foo=EnvVar("SHOULD_NOT_BE_ACCESS")
```

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -891,17 +891,46 @@ def _with_updated_values(
...
             self._schema, self._resolved_config_dict, self.__class__.__name__
         )
-        return self._with_updated_values(post_processed_data)
+
+        return self._with_updated_values(
+            {**post_processed_data, **(incoming_resource_config_evr.value or {})}
+        )
```
schrockn, line 933:
```
return self._with_update_values(translate_enums(context.resource_config))
```

schrockn:
@dpeng817 https://gist.github.com/schrockn/aaffb0d6df7895d9292790be9a90046a gist of some slimmed down enum test cases which minimally expose the problem cases

schrockn:
Information about this PR placed here: https://github.com/dagster-io/dagster/issues/14309.

Superceded by https://github.com/dagster-io/dagster/pull/14306

PR #14247, opened 2023-05-11
Don't assume run group lists are ordered

jmsanders:
We recently had a test flake because this comparison assumes the order of runs in run groups are deterministic. But that doesn't appear to always be the case:

https://buildkite.com/dagster/internal/builds/43597#01880c16-9d42-4124-8eb2-c39a066c80d0

The alternative is to force run groups to return in a deterministic order. I don't know enough about how they're used to know if that's the behavior we want.

### DIFF
--- a/python_modules/dagster/dagster_tests/storage_tests/utils/run_storage.py
+++ b/python_modules/dagster/dagster_tests/storage_tests/utils/run_storage.py
@@ -1,6 +1,7 @@
 import sys
 import tempfile
 import time
+import unittest
 from datetime import datetime, timedelta
 
 import pendulum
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster_tests/storage_tests/utils/run_storage.py
```diff
@@ -1036,8 +1037,9 @@ def test_fetch_run_group(self, storage: RunStorage):
 
         assert len(run_group_two[1]) == 7
 
-        assert run_group_one[0] == run_group_two[0]
-        assert run_group_one[1] == run_group_two[1]
+        # The order of runs in each run run group is not deterministic
```
jmsanders, line 1040: I'm just commenting what I observed - if we feel the order should be sorted, I can enforce that behavior instead (or dig deeper into why it's not happening if it's already expected).

alangenfeld: maybe include something like  "at this time `get_run_group` does not apply an ordering"

alangenfeld (approved):

PR #14249, opened 2023-05-11
clarify lazy auto-materialize reasons in doc and UI

sryza:
## Summary & Motivation

From Sean:

> I think I just mentally assumed each bullet point was one condition
> And my eye skipped ahead to "children freshness policy" and then I assumed "oh the condition for this bullet is about children freshness"


slopp (approved):
TY!

PR #14253, opened 2023-05-11
Store postgres timestamps in UTC, even if the postgres DB is not in UTC

gibsondan:
Summary:
Fixes an issue where the current timestamp was stored in non-UTC if the postgres DB was using a non-UTC timezone

Test Plan:
Run deploy_docker/from_source and check the postgres DB - the time is now stored with a default value using this functino instead of the old CURRENT_TIMESTAMP
Change the timezone of the DB to a non-UTC timezone, launch a run, check the starting timestamp

### DIFF
--- a/examples/deploy_docker/from_source/psql.sh
+++ b/examples/deploy_docker/from_source/psql.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+PGPASSWORD=postgres_password psql -h localhost -p 5432 -U postgres_user postgres_db
--- a/examples/deploy_docker/from_source/docker-compose.yml
+++ b/examples/deploy_docker/from_source/docker-compose.yml
@@ -12,6 +12,10 @@ services:
       POSTGRES_USER: "postgres_user"
       POSTGRES_PASSWORD: "postgres_password"
       POSTGRES_DB: "postgres_db"
+    expose:
+      - "5432"
+    ports:
+      - "5432:5432"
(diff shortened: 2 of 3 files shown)


examples/deploy_docker/from_source/psql.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+PGPASSWORD=postgres_password psql -h localhost -p 5432 -U postgres_user postgres_db
```
gibsondan, line 2: left this helper script in for future postgres spelunking

prha:
Ah... this makes sense to me.

It used to be that we didn't trust db-created timestamps at all, because of weird timezone issues, and relied solely on the Python-generated UTC `timestamp` column in the event log.  But for perf reasons, we decided to use the create_timestamp/update_timestamp values from the runs row into the run record, so we didn't have to query the event log to calculate duration.  So now the actual system timestamp has meaning (unfortunately).

prha (approved):

gibsondan:
hmmm I don't think this quite works actually - this gets it stored using the correct value, but it still comes out as though its in the wrong timezone :/ I think maybe we need to say that you need your postgres to be in UTC for now, seems like a big hassle to get it always in UTC in all circumstances

gibsondan:
No more appetite for this particular rabbit hole - postgres DBs need to use UTC for now i think

PR #14254, opened 2023-05-11
[auto-materialize] Add max materializations per minute

OwenKephart:
## Summary & Motivation

Adds a max_materializations_per_minute parameter to AutoMaterializePolicy, and a corresponding MAX_MATERIALIZATIONS_EXCEEDED discard reason.


python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -723,19 +726,42 @@ def should_reconcile(
...
+                            auto_materialize_policy
+                            and auto_materialize_policy.max_materializations_per_minute
+                            and materialization_counts_by_asset_key[candidate.asset_key]
+                            > auto_materialize_policy.max_materializations_per_minute
+                        ):
+                            max_materializations_exceeded = True
+                            break
+                    # ... then all assets in the unit should have the limit exceeded condition
+                    if max_materializations_exceeded:
+                        for candidate in candidates_unit:
```
sryza, line 759: Feel free to decide this is aesthetically worse,  but all this logic could go inside the `if` statement above, right? And then just `return False` instead of breaking?

python_modules/dagster/dagster/_core/definitions/auto_materialize_policy.py
```diff
@@ -53,6 +54,17 @@ class AutoMaterializePolicy(
...
+
+    Args:
+        on_missing (bool): Whether to materialize an asset / partition when it is missing.
+        on_new_parent_data (bool): Whether to materialize an asset / partition when any of its
+        parents have been updated more recently than it has.
+        for_freshness (bool): Whether to materialize an asset / partition when it, or any of its
+            downstream assets require more up-to-date data to satisfy their freshness policies.
+        time_window_partition_scope_minutes (Optional[float]): [DEPRECATED] The maximum age of
+            a time-window partition to consider materializing.
+        max_materializations_per_minute (Optional[int]): The maximum number of materializations
```
sryza, line 66: I think it's worth mentioning what happens if the limit is exceeded.

python_modules/dagster/dagster/_core/definitions/auto_materialize_policy.py
```diff
@@ -53,6 +54,17 @@ class AutoMaterializePolicy(
 
     Constructing an AutoMaterializePolicy directly is not recommended as the API is subject to change.
     AutoMaterializePolicy.eager() and AutoMaterializePolicy.lazy() are the recommended API.
+
+    Args:
+        on_missing (bool): Whether to materialize an asset / partition when it is missing.
```
sryza, line 59: Based on the earlier discussion where we decided to make `eager`/`lazy` the public-facing mental model, I think we should keep some of these args hidden.

sryza (commented):
We'll also want to mention this behavior in the docs concept page

python_modules/dagster/dagster/_core/definitions/auto_materialize_policy.py
OwenKephart: yeah I'll remove this docstring and then just add this as a parameter to `eager` and `lazy`, with associated docstring on those methods

docs/content/concepts/assets/asset-auto-execution.mdx
```diff
@@ -211,7 +211,10 @@ pluralize
 ### Rules of auto-materialization
 
 - Assets will not be auto-materialized if any of their ancestors are currently being auto-materialized.
+- Assets will not be auto-materialized if any of their ancestors have not yet incorporated the newest materialization of their upstream assets.
 - If the run to auto-materialize an asset fails, it can be retried if [run retries](/deployment/run-retries) are configured. Otherwise, Dagster won't try to auto-materialize that asset again until it would auto-materialize it if the failed run had succeeded. I.e., if an asset has a daily freshness policy, and it fails, Dagster won't auto-materialize it again until the next day.
+- By default, no more than one materialization of a given asset will be kicked off per mintue. Further materialization requests will be discarded, and will require manual backfilling to complete. This can be configured using the `max_materializations_per_minute` arugment to `AutoMaterializePolicy.eager()` and `AutoMaterializePolicy.lazy()`.
```
sryza, line 216:
```suggestion
- By default, no more than one materialization of a given asset will be kicked off per minute. Further materialization requests will be discarded, and will require manual backfilling to complete. This can be configured using the `max_materializations_per_minute` argument to `AutoMaterializePolicy.eager()` and `AutoMaterializePolicy.lazy()`.
```

sryza: Best to use `<PyObject object="AutoMaterializePolicy" method="eager" />`

python_modules/dagster/dagster/_core/definitions/auto_materialize_policy.py
```diff
@@ -83,22 +87,40 @@ def time_window_partition_scope(self) -> Optional[datetime.timedelta]:
...
+            max_materializations_per_minute=check.opt_int_param(
+                max_materializations_per_minute, "max_materializations_per_minute"
+            ),
         )
 
     @public
     @staticmethod
-    def lazy() -> "AutoMaterializePolicy":
+    def lazy(max_materializations_per_minute: Optional[int] = 1) -> "AutoMaterializePolicy":
+        """Args:
```
sryza, line 110: I think this will look nicer if you make the first line "Constructs a lazy AutoMaterializePolicy"

sryza (approved):

PR #14256, opened 2023-05-11
safer timestamp handling in data_time calculation

OwenKephart:
## Summary & Motivation

re: https://github.com/dagster-io/dagster/pull/14251

This part of the conditional should essentially never be called, as any run that fails should have an end timestamp, so this codepath was mostly there to keep type checkers happy.

Nevertheless, we should make sure that we're calling timestamp() in safe ways.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/data_time.py
+++ b/python_modules/dagster/dagster/_core/definitions/data_time.py
@@ -429,7 +429,8 @@ class CachingDataTimeResolver:
             return current_data_time
 
         run_failure_time = datetime.datetime.utcfromtimestamp(
-            latest_run_record.end_time or latest_run_record.create_timestamp.timestamp()
+            latest_run_record.end_time
+            or latest_run_record.create_timestamp.replace(tzinfo=datetime.timezone.utc).timestamp()
         ).replace(tzinfo=datetime.timezone.utc)
         return self._get_in_progress_data_time_in_run(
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster/_core/definitions/data_time.py
```diff
@@ -429,7 +429,8 @@ def get_ignored_failure_data_time(
             return current_data_time
 
         run_failure_time = datetime.datetime.utcfromtimestamp(
-            latest_run_record.end_time or latest_run_record.create_timestamp.timestamp()
+            latest_run_record.end_time
+            or latest_run_record.create_timestamp.replace(tzinfo=datetime.timezone.utc).timestamp()
```
gibsondan, line 433: can I humbly suggest replacing this with `datetime_as_float(latest_run_record.create_timestamp)` to outsource the creative timezone thinking to that function?

gibsondan (changes requested):

python_modules/dagster/dagster/_core/definitions/data_time.py
gibsondan: (but thx for the quick fix)

OwenKephart: ah fair enough

gibsondan (approved):
thanks!

PR #14259, opened 2023-05-12
feedback

salazarm:
## Summary & Motivation

1. Don't assume re-executions are made by users -- they can also be made by retry policies
2. Add back the gray page header
3. Convert Backfill into a tag so you can easily "Add tag to filter"

## How I Tested These Changes
Screen Shot 2023-05-12 at 10 24 44 AM


slopp (commented):
Not close enough to front end code to approve the changes, but the new UX looks great, thank you!!

bengotow (approved):
Code changes here look good to me!

Before you merge, could you also try clicking the "Snapshot" tag to filter? I just remembered when I was adding the asset filter to the runs page (tbd), clicking the Snapshot tags didn't seem to work and I haven't followed up yet. I think we're faking it as a tag and it may need to have a slightly different onClick behavior...

PR #14265, opened 2023-05-12
Partition stale status resolution

smackesey:
## Summary & Motivation

Add stale status and cause resolution at partition granularity. The implementation here is somewhat complicated:

- change the identifier used in the `CachingStaleStatusResolver` to `AssetKeyPartitionKey` instead of the plain `AssetKey`
- update `DataProvenance` to contain input storage ids (already stored on `AssetMaterialization` events) in addition to data versions
- given assets A -> B, with A partitioned, the `DataVersion` for A stored in the provenance tags of a B materialization now uses _the combined hash of all partitions of A on which B depends_. This is also used when auto-computing the `DataVersion` of B. This change allows us to avoid an explosion of tags stored on B materializations in the event that are many dependency partitions of A.
- Exception to above: If A has > `SKIP_PARTITION_DATA_VERSION_DEPENDENCY_THRESHOLD` (set to 100 in this PR) partition dependencies of B, then we revert to old behavior (opt out of accurate data version computation and stale status reporting for the A->B edge). This is a temporary limitation that eventually we should remove.

Given A -> B, with A partitioned, when resolving stale status of B we use the storage ID of A stored on the last materialization of B as a cursor:

- check if _any_ new data version record (materialization or observation) for A has been created after the cursor
    - if no, then the dependency hasn't changed, there is no new upstream data from A, since any dependency partition that was newly materialized would have to exist post-cursor
    - if yes, then for each dependency partition, retrieve two records (1) the most recent materialization; (2) the most recent materialization before the cursor. Record (2) corresponds to the materialization last used as input for B. Compare data versions on these two records-- if they are the same for at least one dependency partition, then B is stale (i.e. has new upstream data from A).

Current issues:

- This _does not work_ for observable source assets, since the optimized method used to fetch the data version tag for a large set of partitions requires filtering by a single event type (so we can't look for both `AssetObservation` and `AssetMaterialization` at once).
- The `CachingStaleStatusResolver` currently performs an optimized batch fetch of the most recent materialization/observation for regular asset keys, but it implements no such optimization for `AssetKeyPartitionKey`
- We should probably change the provenance tag used for the data version of a partitioned dependency-- currently the `DATA_VERSION_TAG` is used, but this is misleading since we are now storing a hash of the data versions of all dependency assets rather than an actual data version.
- There is no record on the materialization of exactly which upstream partitions are used to calculate the "combined data version"-- this is determined at runtime by applying the partition mapping. We might want to store this information.

## How I Tested These Changes

- New unit tests
- Profiling done with this script: https://github.com/dagster-io/dagster/pull/15610

Profiling gives a rough estimate of the effect of the dependency limit. In the benchmark script, resolving stale status for ~2000 partitions took approximately 8 seconds with no limit and 4 seconds with a limit (see full results below). This benchmark tests a dead simple scenario for resolving partition status, and more varied scenarios need to be explored to better understand performance characteristics here.

```
Experiment settings
┏━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━┓
┃                      Key ┃ Value ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━┩
│           num_partitions │  2000 │
│ override_partition_limit │  True │
└──────────────────────────┴───────┘

Execution times
┏━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━┓
┃ Index ┃                                      Step ┃    Time ┃
┡━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━┩
│     0 │         Resolve StaleStatus of all assets │  0.0443 │
│     1 │ Materialize all 2000 partitions of `root` │ 75.1230 │
│     2 │                 Materialize `downstream1` │  0.1839 │
│     3 │         Resolve StaleStatus of all assets │  7.7065 │
│     4 │                 Materialize `downstream2` │  0.1750 │
│     5 │         Resolve StaleStatus of all assets │  7.5177 │
│     6 │    Materialize single partition of `root` │  0.1642 │
│     7 │         Resolve StaleStatus of all assets │  7.5286 │
└───────┴───────────────────────────────────────────┴─────────┘

Experiment settings
┏━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━┓
┃                      Key ┃ Value ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━┩
│           num_partitions │  2000 │
│ override_partition_limit │ False │
└──────────────────────────┴───────┘

Execution times
┏━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━┓
┃ Index ┃                                      Step ┃    Time ┃
┡━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━┩
│     0 │         Resolve StaleStatus of all assets │  0.0440 │
│     1 │ Materialize all 2000 partitions of `root` │ 74.4319 │
│     2 │                 Materialize `downstream1` │  0.2226 │
│     3 │         Resolve StaleStatus of all assets │  4.0985 │
│     4 │                 Materialize `downstream2` │  0.1963 │
│     5 │         Resolve StaleStatus of all assets │  4.1553 │
│     6 │    Materialize single partition of `root` │  0.1618 │
│     7 │         Resolve StaleStatus of all assets │  4.2651 │
└───────┴───────────────────────────────────────────┴─────────┘
```

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/data_version.py
+++ b/python_modules/dagster/dagster/_core/definitions/data_version.py
@@ -23,7 +23,12 @@ from dagster._utils.cached_method import cached_method
 
 if TYPE_CHECKING:
     from dagster._core.definitions.asset_graph import AssetGraph
-    from dagster._core.definitions.events import AssetKey, AssetMaterialization, AssetObservation
+    from dagster._core.definitions.events import (
+        AssetKey,
+        AssetKeyPartitionKey,
+        AssetMaterialization,
+        AssetObservation,
+    )
     from dagster._core.event_api import EventLogRecord
     from dagster._core.events.log import EventLogEntry
     from dagster._core.instance import DagsterInstance
@@ -32,10 +37,6 @@ if TYPE_CHECKING:
 
 class UnknownValue:
     pass
-
-
-def foo(x):
-    return False
 
 
 UNKNOWN_VALUE: Final[UnknownValue] = UnknownValue()
@@ -91,6 +92,7 @@ class DataVersionsByPartition(
 DEFAULT_DATA_VERSION: Final[DataVersion] = DataVersion("INITIAL")
 NULL_DATA_VERSION: Final[DataVersion] = DataVersion("NULL")
 UNKNOWN_DATA_VERSION: Final[DataVersion] = DataVersion("UNKNOWN")
+NULL_EVENT_POINTER: Final[str] = "NULL"
 
 
 class DataProvenance(
@@ -99,6 +101,7 @@ class DataProvenance(
         [
             ("code_version", str),
             ("input_data_versions", Mapping["AssetKey", DataVersion]),
+            ("input_storage_ids", Mapping["AssetKey", Optional[int]]),
             ("is_user_provided", bool),
         ],
     )
@@ -117,6 +120,7 @@ class DataProvenance(
         cls,
         code_version: str,
         input_data_versions: Mapping["AssetKey", DataVersion],
+        input_storage_ids: Mapping["AssetKey", Optional[int]],
         is_user_provided: bool,
     ):
         from dagster._core.definitions.events import AssetKey
@@ -130,30 +134,46 @@ class DataProvenance(
                 key_type=AssetKey,
                 value_type=DataVersion,
             ),
+            input_storage_ids=check.mapping_param(
+                input_storage_ids,
+                "input_storage_ids",
+                key_type=AssetKey,
+            ),
             is_user_provided=check.bool_param(is_user_provided, "is_user_provided"),
         )
 
     @staticmethod
     def from_tags(tags: Mapping[str, str]) -> Optional["DataProvenance"]:
-        from dagster._core.definitions.events import AssetKey
-
         code_version = tags.get(CODE_VERSION_TAG)
         if code_version is None:
             return None
         input_data_versions = {
-            # Everything after the 2nd slash is the asset key
-            AssetKey.from_user_string(k.split("/", maxsplit=2)[-1]): DataVersion(tags[k])
+            _asset_key_from_tag(k): DataVersion(v)
             for k, v in tags.items()
             if k.startswith(INPUT_DATA_VERSION_TAG_PREFIX)
             or k.startswith(_OLD_INPUT_DATA_VERSION_TAG_PREFIX)
         }
+        input_storage_ids = {
+            _asset_key_from_tag(k): int(v) if v != NULL_EVENT_POINTER else None
+            for k, v in tags.items()
+            if k.startswith(INPUT_EVENT_POINTER_TAG_PREFIX)
+        }
         is_user_provided = tags.get(DATA_VERSION_IS_USER_PROVIDED_TAG) == "true"
-        return DataProvenance(code_version, input_data_versions, is_user_provided)
+        return DataProvenance(
+            code_version, input_data_versions, input_storage_ids, is_user_provided
+        )
 
     @property
     @deprecated(breaking_version="2.0", additional_warn_text="Use `input_data_versions` instead.")
     def input_logical_versions(self) -> Mapping["AssetKey", DataVersion]:
         return self.input_data_versions
+
+
+def _asset_key_from_tag(tag: str) -> "AssetKey":
+    from dagster._core.definitions.events import AssetKey
+
+    # Everything after the 2nd slash is the asset key
+    return AssetKey.from_user_string(tag.split("/", maxsplit=2)[-1])
 
 
 # ########################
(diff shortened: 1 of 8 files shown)


smackesey:
* master
  * **PR #14265** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14265" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14265?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/data_version.py
```diff
@@ -503,21 +621,29 @@ def _is_partitioned_or_downstream(self, *, key: "AssetKey") -> bool:
                 for dep_key in self.asset_graph.get_parents(key)
             )
 
+    @cached_method
+    def _get_materialization_count_by_partition(self, *, key: "AssetKey", cursor: int):
```
OwenKephart, line 625: is this necessary? there's a similar method on the CachingInstanceQueryer

python_modules/dagster/dagster/_core/definitions/data_version.py
```diff
@@ -503,21 +621,29 @@ def _is_partitioned_or_downstream(self, *, key: "AssetKey") -> bool:
...
-        if self.asset_graph.is_source(key):
-            return self.asset_graph.is_observable(key)
+    def _is_volatile(self, *, key: "AssetKeyPartitionKey") -> bool:
+        from dagster._core.definitions.events import AssetKeyPartitionKey
+
+        if self.asset_graph.is_source(key.asset_key):
+            return self.asset_graph.is_observable(key.asset_key)
         else:
-            deps = self.asset_graph.get_parents(key)
+            deps = [
```
OwenKephart, line 639: I don't think modifying this method is necessary -- volatility is still linked to the asset key, and so we shouldn't need to traverse the full-resolution asset graph.

python_modules/dagster/dagster/_core/definitions/data_version.py
```diff
@@ -527,15 +653,30 @@ def _get_latest_data_version_event(
...
+        )
+
+    @cached_method
+    def _get_partition_dependencies(
+        self, *, key: "AssetKeyPartitionKey"
+    ) -> Sequence["AssetKeyPartitionKey"]:
+        return sorted(
+            self.asset_graph.get_parents_partitions(
+                dynamic_partitions_store=self._instance,
+                current_time=datetime.datetime.today(),
```
OwenKephart, line 678: should set current_time to self.instance_queryer.evaluation_time

python_modules/dagster/dagster/_core/execution/context/system.py
```diff
@@ -888,35 +903,80 @@ def fetch_external_input_asset_records(self) -> None:
...
-            if event_data_version == self._data_version_cache[key]:
-                self._input_asset_records[key] = event
+        event = self._get_input_asset_event(key)
+        if event is None:
+            self._input_asset_version_info[key] = None
+        else:
+            storage_id = event.storage_id
+            # Input name will be none if this is an internal dep
+            input_name = self.job_def.asset_layer.input_for_asset_key(self.node_handle, key)
+            if input_name and self.has_asset_partitions_for_input(input_name):
```
OwenKephart, line 923: my main concern here is just situations in which we have a `<huge number of partitions>` -> `<unpartitioned asset>` connection, which would require us to fire off a massive query in order to get the latest data versions per each one of those partitions.

it's definitely possible that this is kinda fine in practice, but for this PR I think we can still claim an objective improvement over status quo if we filter out cases where there's an AllPartitionMapping. Essentially, `if input_name and self.has_asset_partitions_for_input and not self._has_all_partition_mapping(input): ...` where has_all_partition_mapping basically just grabs the partition_mapping in the same way as asset_partitions_subset_for_input and makes sure it's not an AllPartitionMapping()

This change is less principled but I think shaves off a bunch of risk from this implementation and is easy to remove at a later point.

OwenKephart (commented):
looking close!

python_modules/dagster/dagster/_core/definitions/data_version.py
smackesey: There was no real change to the logic here, I had changed it just so that all the cached methods use `AssetKeyPartitionKey` to keep things simple, but your comment has made me realize that there is a performance cost for multiple partitions. Reverted.

python_modules/dagster/dagster_tests/execution_tests/test_data_versions.py
```diff
@@ -628,13 +643,99 @@ def asset2(asset1):
         assert status_resolver.get_status(asset2.key) == StaleStatus.FRESH
 
 
-def test_stale_status_partitioned() -> None:
-    partitions_def = StaticPartitionsDefinition(["foo"])
+# import pytest
+# @pytest.mark.skip
```
OwenKephart, line 647: nit: remove these comments

python_modules/dagster/dagster/_core/definitions/data_version.py
```diff
@@ -476,30 +598,32 @@ def _get_current_data_version(self, *, key: "AssetKey") -> DataVersion:
...
         record = self._get_latest_data_version_record(key=key)
         if record is None:
             return None
         else:
             return extract_data_provenance_from_entry(record.event_log_entry)
 
     @cached_method
-    def _is_partitioned_or_downstream(self, *, key: "AssetKey") -> bool:
+    def _is_downstream_of_all_partitions_mapping(self, *, key: "AssetKey") -> bool:
         if self.asset_graph.get_partitions_def(key):
```
OwenKephart, line 620: seems like this function is misnamed

OwenKephart (approved):
great!

python_modules/dagster/dagster/_core/execution/context/system.py
```diff
@@ -880,35 +903,117 @@ def fetch_external_input_asset_records(self) -> None:
...
+        if event is None and is_step_internal_input and retries <= 5:
+            return self._get_input_asset_event(key, retries + 1)
+        elif event is None:
+            return None
+        else:
+            # The cache gets populated for inputs generated during the current step. We need to keep
+            # reloading the latest event until the data version matches the one in the cache to ensure
+            # the storage id reflects the input actually used.
+            data_version = extract_data_version_from_entry(event.event_log_entry)
+            if is_step_internal_input and data_version != self._data_version_cache[key]:
```
alangenfeld, line 982: X

alangenfeld: https://github.com/dagster-io/dagster/issues/15822

PR #14267, opened 2023-05-12
feat(dbt-assets)[2/N]: simplify dbt cli resource

rexledesma:
## Summary & Motivation
The current implementation of the dbt resource that allows the user to invoke dbt CLI commands in Dagster needs to be simplified. Here, we propose a simplification.

- Simplify the interface of the dbt resource to match the underlying invocation of the dbt CLI.
- Provide composability and sensible defaults while configuring and executing the dbt CLI.
    - We should maximize the configuration of the dbt CLI using dbt [Global Configs](https://docs.getdbt.com/reference/global-configs) (i.e. environment variables over YAML configuration).
    - We should provide sensible defaults to convert dbt concepts to Dagster concepts, but the user should have sufficient control over outputs to effectively do whatever they want given the dbt framework in Dagster.

### Nota Bene

- In the initial implementation to support this simplification, we should drop support for dbt-core<1.4.

## How I Tested These Changes
pytest


rexledesma:
* master
    * **PR #14267** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14267" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14267?utm_source=stack-comment).

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -0,0 +1,132 @@
...
+        # https://docs.getdbt.com/reference/exit-codes
+        return_code = self._process.wait()
+
+        return return_code
+
+
+class DbtClientV2:
+    def __init__(self) -> None:
+        # TODO: We should probably check that the dbt-core version here is above 1.4.0.
+        check.invariant(
```
OwenKephart, line 61: commented along these lines on the notion doc, but I don't think we should rely solely on environment variables here, as this is generally going to bewilder users who aren't familiar w/ setting env vars, and likely cause headaches when people try to deploy their local code to prod.

instead, we can have a permissive config schema -- any flags set therein would be passed into CLI invocations (any of these global config env vars can also be passed in as cli flags). we can do some proactive error checking for the user (like if they don't configure a `project_dir`, and their `DBT_PROJECT_DIR` env var is not set, then we can error early), but in general I don't think we should requires these to be set for the integration to work.

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -0,0 +1,132 @@
...
+import os
+import subprocess
+from dataclasses import dataclass
+from typing import Any, Dict, Iterator, List, Sequence
+
+from dagster import _check as check
+
+
+@dataclass
+class DbtCliEventV2:
```
OwenKephart, line 11: I worry that this class is not long for this world, as once we start interfacing with the dbt python API, we'll likely just want to yield raw EventMsg objects from dbt execution.

Is there a way we can avoid having three objects to represent "dbt message"? (those would be `DbtCliEvent`, `DbtCliEvent2`, and `EventMsg`)

OwenKephart: I suppose we could just rename the current `DbtCliEvent` (which we plan to deprecate) `DeprecatedDbtCliEvent` or something, as it's not part of the public interface

python_modules/libraries/dagster-dbt/dagster_dbt_tests/conftest.py
```diff
@@ -8,7 +8,7 @@
 # ======= CONFIG ========
 DBT_EXECUTABLE = "dbt"
 TEST_PROJECT_DIR = file_relative_path(__file__, "dagster_dbt_test_project")
-DBT_CONFIG_DIR = os.path.join(TEST_PROJECT_DIR, "dbt_config")
```
rexledesma, line 11: `profiles.yml` should live in the project directory anyways: https://docs.getdbt.com/docs/core/connection-profiles#advanced-customizing-a-profile-directory.

After this stack lands, we should clean up our test suite to follow dbt best practices, since some of this configuration is from dbt<1.0 days.

python_modules/libraries/dagster-dbt/dagster_dbt/__init__.py
```diff
@@ -12,7 +12,12 @@
     DbtCliResource as DbtCliResource,
     dbt_cli_resource as dbt_cli_resource,
 )
-from .cli.resources_v2 import DbtManifest as DbtManifest
+from .cli.resources_v2 import (
+    DbtClientV2 as DbtClientV2,
+    DbtCliEventV2 as DbtCliEventV2,
```
OwenKephart, line 17: I don't think we should export these as top level exports until after things become a bit less experimental (and we'll want to rename these regardless)

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -74,3 +83,118 @@ def get_node_info_by_output_name(self, output_name: str) -> Mapping[str, Any]:
...
+            yield event
+
+        # TODO: handle the return codes here!
+        # https://docs.getdbt.com/reference/exit-codes
+        return_code = self.process.wait()
+
+        return return_code
+
+
+class DbtClientV2(ConfigurableResource):
```
OwenKephart, line 139: Maybe just call this DbtClient, as the base DbtClient that DbtCliClient subclasses is not really doing anything (and can be renamed BaseDbtCliClient, _DbtClient, or just removed as its only child classes are DbtRpcClient and DbtCliClient, with the former being deprecated soon)

In the interest of minimal work, I think:

DbtClient -> _DbtClient
DbtClientV2 -> DbtClient

would be the easiest

OwenKephart: Ah and make sure to mark this as experimental

OwenKephart (approved):
nice -- had some comments to address but other than those minor things this is good to go

PR #14273, opened 2023-05-12
[wip] gcp auth resource to support mulitple ways of authenticating

jamiedemaria:
## Summary & Motivation

issue right now is that we need to call the get_credentials function of auth resource and i'm not sure where to do that... could give the BigQueryClient an init and call get_credentials and pass the creds to the Client.


jamiedemaria:
* master
  * **PR #14273** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14273" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14273?utm_source=stack-comment).

PR #14274, opened 2023-05-12
[ui] Remove jest warnings about `act` in core

bengotow:
## Summary & Motivation

This PR removes all occurrences of "change must be wrapped in act" from the `core` tests. Removing these mostly came down to:

- Do not `await` click commands, add a waitFor afterwards that waits for the impact to be visible.

- Do await `type` commands -- otherwise subsequent commands (like a click) will run before the typing is complete. You can achieve this by using `act` or by adding a `waitsFor value to be <what you just typed>`

- Do not `await` act if the body of the act function does not return a promise.

Fixing all these revealed some /actual/ warnings coming from the tests, and a few tests which were not actually testing things as intended because they were not awaiting the result properly! We may need some follow-ups to fix the real warnings.

## How I Tested These Changes

Getting these out of the tests was tricky because the log stack traces often don't reference the test that triggered the act warning because the trace is a bunch of async react stuff. To cause tests with act warnings to fail, I opened the react-testing node module and patched it to throw instead of warn. That made it easy to fix a test + re-run over and over again.

### DIFF
--- a/js_modules/dagit/packages/core/src/app/__tests__/CustomConfirmationProvider.test.tsx
+++ b/js_modules/dagit/packages/core/src/app/__tests__/CustomConfirmationProvider.test.tsx
@@ -42,9 +42,11 @@ describe('CustomConfirmationProvider', () => {
     );
 
     const button = screen.queryByRole('button', {name: /terminate/i}) as HTMLButtonElement;
-    await userEvent.click(button);
+    userEvent.click(button);
 
-    expect(screen.queryByText(/r u sure about this/i)).toBeVisible();
+    await waitFor(() => {
+      expect(screen.queryByText(/r u sure about this/i)).toBeVisible();
+    });
   });
 
   it('must perform confirmation correctly', async () => {
@@ -55,22 +57,22 @@ describe('CustomConfirmationProvider', () => {
     );
 
     const button = screen.queryByRole('button', {name: /terminate/i}) as HTMLButtonElement;
-    await userEvent.click(button);
+    userEvent.click(button);
 
-    const confirmationButton = screen.queryByRole('button', {
-      name: /confirm/i,
-    }) as HTMLButtonElement;
+    await waitFor(async () => {
+      const confirmationButton = await screen.findByRole('button', {name: /confirm/i});
+      userEvent.click(confirmationButton);
+    });
 
-    // Resolves the promise.
-    await userEvent.click(confirmationButton);
-
-    expect(screen.queryByText(/confirmed\? true/i)).toBeVisible();
+    await waitFor(() => {
+      expect(screen.queryByText(/confirmed\? true/i)).toBeVisible();
+    });
   });
 
   it('must remove the confirmation Dialog from the DOM afterward', async () => {
     jest.useFakeTimers();
 
-    await act(() => {
+    act(() => {
       render(
         <CustomConfirmationProvider>
           <Test />
@@ -79,7 +81,7 @@ describe('CustomConfirmationProvider', () => {
     });
 
     const button = screen.queryByRole('button', {name: /terminate/i}) as HTMLButtonElement;
-    await act(() => {
+    act(() => {
       button.click();
     });
 
@@ -88,7 +90,7 @@ describe('CustomConfirmationProvider', () => {
     }) as HTMLButtonElement;
 
     // Resolves the promise.
-    await act(() => {
+    act(() => {
       confirmationButton.click();
     });
 
--- a/js_modules/dagit/packages/core/src/assets/__tests__/AssetPartitions.test.tsx
+++ b/js_modules/dagit/packages/core/src/assets/__tests__/AssetPartitions.test.tsx
@@ -70,28 +70,26 @@ const SingleDimensionAssetPartitions: React.FC<{
 
 describe('AssetPartitions', () => {
   it('should support filtering a time-partitioned asset to a time range using the top bar', async () => {
-    await act(async () => {
-      render(<SingleDimensionAssetPartitions assetKey={{path: ['single_dimension_time']}} />);
-    });
+    render(<SingleDimensionAssetPartitions assetKey={{path: ['single_dimension_time']}} />);
 
     await waitFor(() => {
       expect(screen.getByTestId('partitions-selected')).toHaveTextContent('6,000 Partitions');
     });
 
-    const partitionInput = screen.getByTestId('dimension-range-input');
-    await userEvent.clear(partitionInput);
-    await userEvent.type(partitionInput, '{[}2022-11-28-20:00...2022-12-05-01:00{]}');
-    await userEvent.tab();
+    await act(async () => {
+      const partitionInput = screen.getByTestId('dimension-range-input');
+      await userEvent.clear(partitionInput);
+      await userEvent.type(partitionInput, '{[}2022-11-28-20:00...2022-12-05-01:00{]}');
+      await userEvent.tab();
+    });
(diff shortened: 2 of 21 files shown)


js_modules/dagit/packages/core/src/app/__tests__/CustomConfirmationProvider.test.tsx
```diff
@@ -55,22 +57,22 @@ describe('CustomConfirmationProvider', () => {
...
 
-    const confirmationButton = screen.queryByRole('button', {
-      name: /confirm/i,
-    }) as HTMLButtonElement;
-
-    // Resolves the promise.
-    await userEvent.click(confirmationButton);
+    await waitFor(async () => {
+      const confirmationButton = await screen.findByRole('button', {name: /confirm/i});
+      userEvent.click(confirmationButton);
```
salazarm, line 64: does it matter that the click is inside the `waitFor` ?

bengotow: No it doesn't -- we could also write it this way, it just repeats the find clause twice. I think I put it inside the block so that I didn't have to copy the regexp:

```

    await waitFor(async () => {
      await screen.findByRole('button', {name: /confirm/i});
    });
    userEvent.click(screen.getByRole('button', {name: /confirm/i}));

 
```

salazarm: you can also write 
```
    const button = await waitFor(async () => screen.findByRole('button', {name: /confirm/i});
    userEvent.click(button);
```

js_modules/dagit/packages/core/src/assets/__tests__/AssetPartitions.test.tsx
```diff
@@ -70,45 +70,43 @@ const SingleDimensionAssetPartitions: React.FC<{
...
 
     await waitFor(() => {
       expect(screen.getByTestId('partitions-selected')).toHaveTextContent('6,000 Partitions');
     });
 
-    const partitionInput = screen.getByTestId('dimension-range-input');
-    await userEvent.clear(partitionInput);
-    await userEvent.type(partitionInput, '{[}2022-11-28-20:00...2022-12-05-01:00{]}');
-    await userEvent.tab();
+    await act(async () => {
```
salazarm, line 79: do you need this act even though there's an waitFor below?

js_modules/dagit/packages/core/src/ticks/__tests__/EvaluateScheduleDialog.test.tsx
```diff
@@ -38,59 +38,58 @@ function Test({mocks, resolvers}: {mocks?: MockedResponse[]; resolvers?: Resolve
...
-    await act(async () => {
-      render(<Test mocks={[GetScheduleQueryMock, ScheduleDryRunMutationRunRequests]} />);
-    });
+    render(<Test mocks={[GetScheduleQueryMock, ScheduleDryRunMutationRunRequests]} />);
+
     await waitFor(async () => {
       const selectButton = screen.getByTestId('tick-selection');
       expect(selectButton).toBeVisible();
-      await userEvent.click(selectButton);
+      userEvent.click(selectButton);
```
salazarm, line 48: actually I think these need to be awaited now.

js_modules/dagit/packages/core/src/schedules/__tests__/ScheduleStateChangeDialog.test.tsx
```diff
@@ -116,7 +116,7 @@ describe('ScheduleStateChangeDialog', () => {
       );
 
       const confirmButton = screen.getByRole('button', {name: /start/i});
-      await userEvent.click(confirmButton);
+      userEvent.click(confirmButton);
```
salazarm, line 119: same pretty sure these should be awaited no?

salazarm: https://testing-library.com/docs/user-event/intro/

js_modules/dagit/packages/core/src/nav/__tests__/useRepositoryLocationReload.test.tsx
```diff
@@ -171,15 +177,13 @@ describe('useRepositoryReloadLocation', () => {
...
-    rerender(
-      <TestProvider apolloProps={{mocks: defaultMocks}}>
-        <Test />
-      </TestProvider>,
-    );
-
-    jest.runAllTicks();
-
     await waitFor(() => {
+      rerender(
```
salazarm, line 181: why is the `rerender` in the `waitFor` ?

js_modules/dagit/packages/core/src/nav/__tests__/useRepositoryLocationReload.test.tsx
```diff
@@ -171,15 +177,13 @@ describe('useRepositoryReloadLocation', () => {
...
+        <TestProvider apolloProps={{mocks: defaultMocks}}>
+          <Test />
+        </TestProvider>,
+      );
+      jest.runAllTicks();
```
salazarm, line 186: if the re-render is doing async things then we should wrap it in `act` rather than `waitFor` imo.

js_modules/dagit/packages/core/src/partitions/__tests__/CreatePartitionDialog.test.tsx
```diff
@@ -63,17 +66,27 @@ describe('CreatePartitionDialog', () => {
...
+    });
     await act(async () => {
-      render(<Test mocks={[]} />);
+      userEvent.clear(partitionInput);
+    });
+    await waitFor(() => {
+      expect(screen.queryByTestId('warning-icon')).toBeNull();
+    });
+    await act(async () => {
+      await userEvent.type(partitionInput, 'validname');
```
salazarm, line 85: do we need `act` around `userEvent.type` ? seems odd.

js_modules/dagit/packages/core/src/app/__tests__/CustomConfirmationProvider.test.tsx
bengotow: Nice! Will switch it over to this format

js_modules/dagit/packages/core/src/nav/__tests__/useRepositoryLocationReload.test.tsx
bengotow: Ahh we can't do this because for whatever reason, `rerender` does not return a promise but IS asynchornous, so you can't `await act(async () => await rerender())`

js_modules/dagit/packages/core/src/partitions/__tests__/CreatePartitionDialog.test.tsx
bengotow: Yeah oddly enough we do, these tests do some pretty wild stuff if we remove this await. You can actually get into a situation where it's still typing /from the previous test/ if you don't wait for it to complete here.

js_modules/dagit/packages/core/src/assets/__tests__/AssetPartitions.test.tsx
bengotow: Yes because it's important that we run these user events /in order/, and they are generating react re-renders as they run so they must be in an act block.

PR #14275, opened 2023-05-12
Add a `dagster code-server start` CLI entrypoint with the ability to reload code without restarting the process

gibsondan:
## Summary & Motivation
A new entry point that adds a layer of indirection between the entrypoint process, and a subprocess that actually runs user code. The main benefit this gives is the ability to reload code without restarting the process (or re-deploying the k8s pod / task / etc.) - resolving https://github.com/dagster-io/dagster/issues/12581.

The main entry point implements the same gRPC api, but then redirects requests to a subprocess that it spins up.

The parent process uses a GrpcServerRegistry under the hood, which handles things like cleaning up subprocesses after they die and are no longer running any executions.

Unfortunately since gRPC servers still don't always have access to the instance, had to make the instance argument to GrpcServerRegistry optional (and add a few more config toggles since they get passed through).

Also added a new graphql test suite which should exercise the actual underlying gRPC api pretty comprehensively, and added a test that verifies that reloading works the way you would expect.

This is a large PR, I could potentially split it into two - one that adds the entrypoint, the other that implements the reload logic in dagster-graphql.

### DIFF
--- a/python_modules/dagit/dagit/debug.py
+++ b/python_modules/dagit/dagit/debug.py
@@ -1,11 +1,16 @@
 from gzip import GzipFile
+from typing import Any, Optional
 
 import click
 from dagster import (
     DagsterInstance,
 )
 from dagster._core.debug import DebugRunPayload
-from dagster._core.workspace.context import WorkspaceProcessContext
+from dagster._core.workspace.context import (
+    BaseWorkspaceRequestContext,
+    IWorkspaceProcessContext,
+    WorkspaceRequestContext,
+)
 from dagster._serdes.serdes import deserialize_value
 
 from .cli import (
@@ -14,6 +19,41 @@ from .cli import (
     host_dagit_ui_with_workspace_process_context,
 )
 from .version import __version__
+
+
+class DagitDebugWorkspaceProcessContext(IWorkspaceProcessContext):
+    def __init__(
+        self,
+        instance: DagsterInstance,
+    ):
+        self._instance = instance
+
+    def create_request_context(self, source: Optional[Any] = None) -> BaseWorkspaceRequestContext:
+        return WorkspaceRequestContext(
+            instance=self._instance,
+            workspace_snapshot={},
+            process_context=self,
+            version=__version__,
+            source=source,
+            read_only=False,
+        )
+
+    @property
+    def version(self) -> str:
+        return __version__
+
+    def refresh_code_location(self, name: str) -> None:
+        raise NotImplementedError
+
+    def reload_code_location(self, name: str) -> None:
+        raise NotImplementedError
+
+    def reload_workspace(self) -> None:
+        pass
+
+    @property
+    def instance(self) -> DagsterInstance:
+        return self._instance
 
 
 @click.command(
@@ -44,7 +84,7 @@ def dagit_debug_command(input_files, port):
             debug_payloads.append(debug_payload)
 
     instance = DagsterInstance.ephemeral(preload=debug_payloads)
-    with WorkspaceProcessContext(instance, None, version=__version__) as workspace_process_context:
+    with DagitDebugWorkspaceProcessContext(instance) as workspace_process_context:
         host_dagit_ui_with_workspace_process_context(
             workspace_process_context=workspace_process_context,
             port=port,
--- a/python_modules/dagit/dagit/debug.py
+++ b/python_modules/dagit/dagit/debug.py
@@ -22,6 +22,11 @@ from .version import __version__
 
 
 class DagitDebugWorkspaceProcessContext(IWorkspaceProcessContext):
+    """IWorkspaceProcessContext that works with an ephemeral instance, which is needed
+    for dagit-debug to work (a regular WorkspaceProcessContext will fail when it tries
+    to call .get_ref() on the instance when spinning up a code server).
+    """
+
     def __init__(
         self,
         instance: DagsterInstance,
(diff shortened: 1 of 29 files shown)


python_modules/dagster/dagster/_cli/code_server.py
```diff
@@ -0,0 +1,239 @@
...
+    "--max-workers",
+    "--max_workers",  # for backwards compatibility
+    "-n",
+    type=click.INT,
+    required=False,
+    default=None,
+    help="Maximum number of (threaded) workers to use in the code server",
+)
+@click.option(
+    "--lazy-load-user-code",
```
gibsondan, line 65: this name is bad (we don't lazy load it, we just don't fail the process) and it shouldn't say 'user' code - should keep the old name for back-compat but include a new name

python_modules/dagster/dagster/_core/host_representation/grpc_server_registry.py
```diff
@@ -268,11 +281,23 @@ def __exit__(self, exception_type, exception_value, traceback):
...
+        if self._wait_for_processes_on_shutdown:
+            self.wait_for_processes()
+
+    def shutdown_all_processes(self):
         for process in self._all_processes:
             process.shutdown_server()
 
-        if self.instance.code_server_settings.get("wait_for_local_processes_on_shutdown", False):
-            self.wait_for_processes()
+    def have_all_servers_shut_down(self) -> bool:
```
alangenfeld, line 293: nit: i got confused by the name interpreting it as a command "have all servers shut down" instead of a question. Consider `all_servers_have_shutdown`, `check_...` or something

python_modules/dagster/dagster/_grpc/proxy_server.py
```diff
@@ -0,0 +1,315 @@
...
+    CancelExecutionResult,
+    ExecuteExternalJobArgs,
+    ShutdownServerResult,
+    StartRunResult,
+)
+
+CLEANUP_TICK = 1
+
+
+class DagsterProxyApiServer(DagsterApiServicer):
```
alangenfeld, line 33: + docblock

python_modules/dagster-graphql/dagster_graphql_tests/graphql/graphql_context_test_suite.py
```diff
@@ -377,6 +423,8 @@ class Marks:
     multi_location = pytest.mark.multi_location
     managed_grpc_env = pytest.mark.managed_grpc_env
     deployed_grpc_env = pytest.mark.deployed_grpc_env
+    code_server_cli_grpc_env = pytest.mark.code_server_cli_grpc_env
```
alangenfeld, line 426: im not sure the best place to do it -but we should write down exactly what the difference is between the three conditions here since i dont think it would be easy for someone to re-derive the underlying assumed behavior differences that are the reason we test each

alangenfeld (commented):
What are the biggest risks with this? Errors getting weird with the extra proxy hop? 

How does the start-up sequence work relative to direct launch? In the case of super slow start-up times, do we gracefully handle that in the proxy til the real server comes up?

gibsondan:
Re: startup time there's no timeout by default (it doesn't make a blocking gRPC call or anything, it just waits until it's ready). There's an optional --startup-timeout arg you can add that will make it raise an error if takes longer than that (which the old command did not have, it would just sit forever)

If there's a crash or a timeout on startup it should show you the error when you call ListRepositories just like the regular gRPC server does.

Other risks... more memory usage? Weird cases when there are two subprocesses running at the same time? But overall not too worried.

python_modules/dagster-graphql/dagster_graphql_tests/graphql/graphql_context_test_suite.py
gibsondan: The labels of the marks in the pytest.ini do a decent summary i think, i'll add it to the implementation as well

gibsondan:
- added a test case for the --startup-timeout arg
- dockblock for DagsterApiProxyServicer
- comments around the various dagster-graphql test suite cases

python_modules/dagit/dagit/debug.py
```diff
@@ -16,6 +21,41 @@
...
 
 
+class DagitDebugWorkspaceProcessContext(IWorkspaceProcessContext):
+    def __init__(
+        self,
+        instance: DagsterInstance,
+    ):
+        self._instance = instance
+
+    def create_request_context(self, source: Optional[Any] = None) -> BaseWorkspaceRequestContext:
```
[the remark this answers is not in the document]
alangenfeld: maybe leave this as a comment on the class to explain why it exists since its subtle

python_modules/dagster/dagster/_cli/code_server.py
```diff
@@ -0,0 +1,237 @@
...
+    envvar="DAGSTER_CODE_SERVER_STARTUP_TIMEOUT",
+)
+@click.option(
+    "--instance-ref",
+    type=click.STRING,
+    required=False,
+    help="[INTERNAL] Serialized InstanceRef to use for accessing the instance",
+    envvar="DAGSTER_INSTANCE_REF",
+)
+def start_command(
```
alangenfeld, line 144: some concern over ensuring all these settings get passed over to the proxy correctly

alangenfeld (approved):

PR #14278, opened 2023-05-14
Allow specification of target within load_assets_from_dbt_project

bsalerno:
## Summary & Motivation

Currently, using the `load_assets_from_dbt_project` function to load dbt assets uses the default target for the specified profile. This does not affect *runs* of the assets, since that is handled by the dbt cli resource (which is using a specified `target`), but it does affect metadata that is target dependent (i.e. [this example](https://docs.getdbt.com/reference/dbt-jinja-functions/target#use-targetname-to-change-your-source-database) from the dbt documentation). 

We can add this functionality by allowing the function within `load_assets_from_dbt_project` that creates the manifest (`_load_manifest_for_project`) to take a `target` argument and pass it to the `execute_cli` function (which can take `target` within its `flags_dict`. With this change, users can pass through the target (ideally the one that they are using for the dbt cli resource definition) so that the metadata within the manifest matches up with the target environment that commands will be run in. 

Note that if no target is passed to `load_assets_from_dbt_project`, the existing behavior (default target for specified profile) will remain, so this should be backwards compatible.

## How I Tested These Changes

Installed this version of `dagster-dbt` within local deployment and checked the "Definition" metadata fields of assets that had target-dependent metadata fields.

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
@@ -70,6 +70,7 @@ def _load_manifest_for_project(
     target_dir: str,
     select: str,
     exclude: str,
+    target: str,
 ) -> Tuple[Mapping[str, Any], DbtCliOutput]:
     # running "dbt ls" regenerates the manifest.json, which includes a superset of the actual
     # "dbt ls" output
@@ -82,6 +83,7 @@ def _load_manifest_for_project(
             "profiles-dir": profiles_dir,
             "select": select,
             "exclude": exclude,
+            "target": target,
             "output": "json",
         },
         warn_error=False,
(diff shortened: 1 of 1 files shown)


OwenKephart:
Hi @bsalerno ! Thanks for reaching out. In the past, we've taken the stance that this is mostly expected behavior -- in other (non-dbt) cases, the metadata attached to the asset definitions generally does not change from environment to environment.

At the same time, we'd ideally like to deprecate the `load_assets_from_dbt_project` behavior, in favor of loading from a pre-existing manifest. In general, having to wait for dbt to compile the dbt project whenever we spin up a subprocess causes a variety of issues (it can cause timeouts, if multiple processes are executing the compile at once you can get race conditions, and it means the state of the project when you launch a run may differ from the state of the project once you start a run, if the dbt project changes between those points in time).

If you want this behavior specifically, another option would be to do something like:

```
def my_get_dbt_assets(...):
    # invoke dbt compile --target my target
    ...
    with open("path/to/target/manifest.json") as f:
        manifest = json.load(f)
    return load_assets_from_dbt_manifest(manifest, ...)
```

With that said, my current inclination would be to not move forward on getting this merged in (so I'm closing this for now), although I'm definitely open to discussing this more!

PR #14279, opened 2023-05-15
Docs: Removed unneeded imports

jerome3o:
## Summary & Motivation

I've just finished the latest tutorial for Dagster (amazing tutorial by the way!) and noticed in part 4 (Building an asset graph) there are some unneeded imports - I got through the rest of the tutorial and they remained unused. Unless there is an additional section I've missed I think these can be removed.

## How I Tested These Changes

I finished off the rest of the tutorial and had a working pipeline.


jamiedemaria:
@tacastillo  mind taking a look?

PR #14300, opened 2023-05-15
[ui] Add filter button to Schedules and Sensors

hellendag:
## Summary & Motivation

Use the new filter component to allow code location and running state filtering on the Overview schedules and sensors pages, as well as the code location schedules and sensors pages.

When no code locations are selected, all code locations are visible. Changing the selected code locations will also update the left nav, as it changes global-level code location filtering, as tracked in localStorage.

https://github.com/dagster-io/dagster/assets/2823852/88089699-9cae-4243-bdc8-de42449ffcde

## How I Tested These Changes

View Overview > Schedules, verify that the filter behavior is correct:

- Add/remove code locations in filter, verify that this updates the "hidden" code location state correctly, including the left nav.
- Add running/stopped in filter, verify that schedules are filtered correctly.

Repeat on Overview sensors and code location schedules/sensors.

### DIFF
--- a/js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
+++ b/js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
@@ -247,7 +247,14 @@ export const FilterDropdown = ({filters, setIsOpen, setPortaledElements}: Filter
               <Spinner purpose="section" />
             </Box>
           ) : allResultsJsx.length ? (
-            <Container ref={parentRef} style={{maxHeight: '500px', overflowY: 'auto'}}>
+            <Container
+              ref={parentRef}
+              style={{
+                maxHeight: '500px',
+                overflowY: 'auto',
+                width: selectedFilter?.menuWidth || 'auto',
+              }}
+            >
               <Inner $totalHeight={totalHeight}>
                 {items.map(({index, end, start}) => {
                   return (
@@ -332,7 +339,7 @@ export const FilterDropdownButton = React.memo(({filters}: FilterDropdownButtonP
         canEscapeKeyClose
         popoverClassName="filter-dropdown"
         isOpen={isOpen}
-        position="bottom"
+        placement="bottom-start"
         onClosing={() => {
           prevOpenRef.current = false;
         }}
@@ -427,7 +434,6 @@ const SlashShortcut = styled.div`
 
 const PopoverStyle = createGlobalStyle`
   .filter-dropdown.filter-dropdown.filter-dropdown.filter-dropdown {
-    margin-left: 16px !important;
     border-radius: 8px;
     max-width: 100%;
     overflow: hidden;
--- a/js_modules/dagit/packages/core/src/ui/Filters/useCodeLocationFilter.tsx
+++ b/js_modules/dagit/packages/core/src/ui/Filters/useCodeLocationFilter.tsx
@@ -1,0 +1,52 @@
+import * as React from 'react';
+
+import {TruncatedTextWithFullTextOnHover} from '../../nav/getLeftNavItemsForOption';
+import {WorkspaceContext} from '../../workspace/WorkspaceContext';
+import {buildRepoAddress} from '../../workspace/buildRepoAddress';
+import {repoAddressAsHumanString} from '../../workspace/repoAddressAsString';
+import {RepoAddress} from '../../workspace/types';
+
+import {useStaticSetFilter} from './useStaticSetFilter';
+
+export const useCodeLocationFilter = () => {
+  const {allRepos, visibleRepos, setVisible, setHidden} = React.useContext(WorkspaceContext);
+
+  const allRepoAddresses = React.useMemo(() => {
+    return allRepos.map((repo) =>
+      buildRepoAddress(repo.repository.name, repo.repositoryLocation.name),
+    );
+  }, [allRepos]);
+
+  const visibleRepoAddresses = React.useMemo(() => {
+    return visibleRepos.length === allRepos.length
+      ? []
+      : visibleRepos.map((repo) =>
+          buildRepoAddress(repo.repository.name, repo.repositoryLocation.name),
+        );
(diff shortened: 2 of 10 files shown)


hellendag:
* master
  * **PR #14300** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14300" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14300?utm_source=stack-comment).

js_modules/dagit/packages/core/src/ui/Filters/useFilter.tsx
```diff
@@ -29,7 +29,7 @@ export const FilterTag = ({
   iconName: IconName;
   onRemove: () => void;
 }) => (
-  <div style={{display: 'inline-block', height: '24px'}}>
+  <div>
```
hellendag, line 32: It seems like we can just use a flex row to render tags, wdyt? Wrapping the tag in a `div` should prevent it from stretching weirdly.

js_modules/dagit/packages/core/src/workspace/WorkspaceContext.tsx
```diff
@@ -273,7 +284,32 @@ const useVisibleRepos = (
...
+      const keysToHide = new Set(
+        repoAddresses.map((repoAddress) => `${repoAddress.name}:${repoAddress.location}`),
+      );
+      setHiddenKeys((current) => {
+        const updatedSet = new Set([...(current || []), ...keysToHide]);
+        return Array.from(updatedSet);
+      });
+    },
+    [setHiddenKeys],
+  );
```
hellendag, line 310: Added these to `useVisibleRepos` because they allow better control of multiple repos vs. `toggleVisible`.

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
```diff
@@ -332,7 +339,7 @@ export const FilterDropdownButton = React.memo(({filters}: FilterDropdownButtonP
         canEscapeKeyClose
         popoverClassName="filter-dropdown"
         isOpen={isOpen}
-        position="bottom"
+        placement="bottom-start"
```
hellendag, line 342: Shifted the menu to be left-aligned flush with the button, instead of centered.

salazarm: oh sweet, thanks!

js_modules/dagit/packages/core/src/ui/Filters/FilterDropdown.tsx
```diff
@@ -247,7 +247,14 @@ export const FilterDropdown = ({filters, setIsOpen, setPortaledElements}: Filter
               <Spinner purpose="section" />
             </Box>
           ) : allResultsJsx.length ? (
-            <Container ref={parentRef} style={{maxHeight: '500px', overflowY: 'auto'}}>
+            <Container
+              ref={parentRef}
+              style={{
+                maxHeight: '500px',
+                overflowY: 'auto',
+                width: selectedFilter?.menuWidth || 'auto',
```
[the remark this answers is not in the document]
salazarm: Looks good to me!

js_modules/dagit/packages/core/src/ui/Filters/useCodeLocationFilter.tsx
```diff
@@ -0,0 +1,51 @@
...
+  return useStaticSetFilter<RepoAddress>({
+    name: 'Code location',
+    icon: 'folder',
+    initialState: visibleRepoAddresses,
+    allValues: allRepoAddresses.map((repoAddress) => {
+      return {value: repoAddress, match: [repoAddressAsHumanString(repoAddress)]};
+    }),
+    getKey: (repoAddress) => repoAddressAsHumanString(repoAddress),
+    renderLabel: ({value}) => (
+      <div style={{overflow: 'hidden'}}>{repoAddressAsHumanString(value)}</div>
```
salazarm, line 36: `TruncatedTextWithFullTextOnHover`

js_modules/dagit/packages/core/src/ui/Filters/useFilter.tsx
salazarm: seems fine to me.

salazarm:
We could probably filter on the last run timestamp

salazarm (approved):

js_modules/dagit/packages/core/src/workspace/WorkspaceContext.tsx
```diff
@@ -273,16 +286,41 @@ const useVisibleRepos = (
...
     // If there's only one repo, skip the local storage check -- we have to show this one.
     if (allRepos.length === 1) {
       return allRepos;
-    } else {
-      return allRepos.filter((o) => !hiddenKeys.includes(getRepositoryOptionHash(o)));
     }
-  }, [allRepos, hiddenKeys]);
+    const hiddenKeys = new Set(JSON.parse(hiddenKeysJSON));
+    return allRepos.filter((o) => !hiddenKeys.has(getRepositoryOptionHash(o)));
+  }, [allRepos, hiddenKeysJSON]);
```
hellendag, line 321: This fixes a bug I noticed where `hiddenKeys` is being updated in a loop with the same values. It appears to be happening because of the `onStateChanged`, which pushes updates to the hidden and visible keys, which results in an update to context, which pushes a new state to the filter component, which triggers `onStateChanged`, etc.

I think there's always a bit of risk with setting the global hidden/visible key state, as any caller might set the hidden repo keys at any time and at any cadence, and there's nothing stopping the entire workspace context from updating at that cadence.

We can at least guard against this flavor of loop (where the same list is set in state repeatedly) by stringifying the array and using that for `useMemo` comparison. This way, if the list of hidden and visible keys is the same, we won't update the `visibleRepos` array at all, and the filter component won't push a new state value.

PR #14289, opened 2023-05-15
feat(dbt-assets)[3/N]: add method to convert dbt events to Dagster events

rexledesma:
## Summary & Motivation
We basically take `_events_for_structured_json_line` and add it as a default implementation for converting dbt CLI events to Dagster events.

## How I Tested These Changes
pytest

PR #14301, opened 2023-05-15
feat(dbt-assets)[6/N]: add examples for `@dbt_assets`

rexledesma:
## Summary & Motivation
We add examples to show off the new APIs for interfacing with dbt within Dagster. This will be presented in an associated GitHub discussion.

Some use cases for your dbt project in Dagster:
- Running `dbt build`
- Customizing the corresponding Dagster asset keys for your dbt resources
- Running your dbt project with Dagster partitions
- Adding customized metadata to your Dagster events
- Manipulating dbt artifacts after running a dbt command

## How I Tested These Changes
local

### DIFF
--- a/examples/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/assets/use_artifacts.py
+++ b/examples/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/assets/use_artifacts.py
@@ -1,0 +1,17 @@
+import pprint
+import sys
+
+from dagster import OpExecutionContext
+from dagster_dbt.asset_decorator import dbt_assets
+from dagster_dbt.cli import DbtCli
+
+from tutorial_dbt_dagster_v2.assets import manifest
+
+
+@dbt_assets(manifest=manifest)
+def my_dbt_assets(context: OpExecutionContext, dbt: DbtCli):
+    dbt_build = dbt.cli(["build"], manifest=manifest, context=context)
+
+    yield from dbt_build.stream()
+
+    sys.stdout.write(pprint.pformat(dbt_build.get_artifact("run_results.json")))
--- a/examples/tutorial_dbt_dagster_v2/README.md
+++ b/examples/tutorial_dbt_dagster_v2/README.md
@@ -1,0 +1,28 @@
+# (Experimental) dbt with software-defined assets tutorial example
+
+This example demonstrates how to integrate dbt with Dagster using dbt's example [jaffle shop project](https://github.com/dbt-labs/jaffle_shop), the [dagster-dbt library](/\_apidocs/libraries/dagster-dbt), and a [DuckDB database](https://duckdb.org/).
+
+In this example, we use the new `dagster-dbt` APIs that require the usage of `dbt-core>=1.4.0`.
+
+---
+
+## Getting started
+
+To download this example, run:
+
+```shell
+dagster project from-example --name my-dagster-project --example tutorial_dbt_dagster_v2
+```
+
+To install this example and its dependencies, run:
+
+```shell
+cd my-dagster-project
+pip install -e ".[dev]"
+```
+
+At this point, you can view the **completed** project in the Dagster UI by running:
+
+```shell
+dagster dev
+```
--- a/examples/tutorial_dbt_dagster_v2/jaffle_shop/models/staging/stg_orders.sql
+++ b/examples/tutorial_dbt_dagster_v2/jaffle_shop/models/staging/stg_orders.sql
@@ -1,3 +1,5 @@
+{{ config(alias='staging_orders') }}
+
 with source as (
 
     {#-
--- a/examples/tutorial_dbt_dagster_v2/pyproject.toml
+++ b/examples/tutorial_dbt_dagster_v2/pyproject.toml
@@ -1,0 +1,9 @@
+[project]
+name = "tutorial_dbt_dagster_v2"
+
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[tool.dagster]
+module_name = "tutorial_dbt_dagster_v2"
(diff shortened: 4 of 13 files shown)


rexledesma:
* master
    * **PR #14267** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14267" target="_blank">Graphite</a> 
      * **PR #14289** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14289" target="_blank">Graphite</a> 
        * **PR #14301** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14301" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14301?utm_source=stack-comment).

examples/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/assets/case_8.py
```diff
@@ -0,0 +1,33 @@
...
+    dbt_subset_selection = DbtManifest(manifest=manifest).get_subset_selection_for_context(
+        context=context
+    )
+    dbt_commands = (
+        [
+            ["seed", "--select", dbt_subset_selection],
+            ["run", "--select", dbt_subset_selection],
+            ["test", "--select", dbt_subset_selection],
+        ]
+        if dbt_subset_selection
```
OwenKephart, line 20: I think we could just have the user use `dbt_subset_selection` in all cases, which would simplify this a bit. Maybe just calling it `dbt_selection_for_context` would be clearer in that case. An added bonus of that approach is that if we wire the raw select string from the decorator into the op metadata / tags, then we avoid making the use duplicate their raw select string from the decorator into the body of the asset

examples/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/assets/case_1b.py
```diff
@@ -0,0 +1,39 @@
...
+
+from tutorial_dbt_dagster_v2.assets import manifest
+
+STAGING_SELECT = "fqn:staging.*"
+MATERIALIZE_SELECT = "fqn:customers fqn:orders"
+
+
+## Case 1(b): run dbt seed, run, and test as separate commands and as separate assets.
+@dbt_multi_asset(manifest=manifest, select="resource_type:seed")
+def dbt_seed_assets(dbt: ResourceParam[DbtClientV2]):
```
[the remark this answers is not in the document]
rexledesma: Made it configurable in https://github.com/dagster-io/dagster/pull/14267!

examples/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/assets/dbt_seed_run_test.py
```diff
@@ -0,0 +1,20 @@
+from dagster_dbt.asset_decorators import dbt_multi_asset
+from dagster_dbt.cli.resources_v2 import DbtClientV2, DbtManifest
+
+from tutorial_dbt_dagster_v2.assets import raw_manifest
+
+manifest = DbtManifest(raw_manifest=raw_manifest)
+
+
+## Case 1(c): run dbt seed, run, and test as separate commands.
```
sryza, line 9: I think we should take out these numbers and letters, because they'll be difficult to maintain

examples/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/assets/dbt_add_partitions.py
```diff
@@ -0,0 +1,34 @@
...
+
+manifest = DbtManifest(raw_manifest=raw_manifest)
+
+
+## Case 5: Partition different sections of my dbt graph. This replaces:
+##   - partitions_def
+##   - partition_key_to_vars_fn
+@dbt_multi_asset(manifest=manifest, select=DBT_SELECT_SEED)
+def dbt_seed_assets(context: OpExecutionContext, dbt: DbtClientV2):
+    for event in dbt.cli(["seed"], context=context, manifest=manifest).stream():
```
OwenKephart, line 19: this applies to all of these examples, but

```
yield from dbt.cli(["seed"], context=context, manifest=manifest).steam_dagster_events()
```

feels a lot smoother. we can keep around `stream()` for users that want even more flexibility, but reduce boilerplate for the average user. Potentially we just call `stream_dagster_events` "`stream`", then rename the existing behavior to `stream_raw_events()` or something of the sort (just to spare some keystrokes)

rexledesma: This has been added in https://github.com/dagster-io/dagster/pull/14289! Thanks for the suggestion, this improvement saves a line in the compute body implementation in the simple case 🙌

examples/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/assets/use_artifacts.py
```diff
@@ -0,0 +1,17 @@
...
+from dagster import OpExecutionContext
+from dagster_dbt.asset_decorator import dbt_assets
+from dagster_dbt.cli import DbtCli
+
+from tutorial_dbt_dagster_v2.assets import manifest
+
+
+@dbt_assets(manifest=manifest)
+def my_dbt_assets(context: OpExecutionContext, dbt: DbtCli):
+    dbt_build = dbt.cli(["build"], manifest=manifest, context=context)
```
OwenKephart, line 13: in other places, we just have `yield from dbt.cli(...).stream()`

OwenKephart (approved):
note comments

examples/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/assets/use_artifacts.py
rexledesma: This is an example to show how to retrieve artifacts from the dbt task after it is completed. We need a reference to the dbt task to retrieve the artifacts, which is why this deviates from the regular `yield from dbt.cli(...).stream()` syntax.

PR #14306, opened 2023-05-16
Actually apply runtime config to instantiated ConfigurableResource objects

dpeng817:
Fix for https://github.com/dagster-io/dagster/issues/14309#event-9257838645.

This PR untangles and fixes a few separate bugs, which were all obfuscated by the fact that were calling `process_config` on the config passed to a `ConfigurableResource` at instantiation:
- We were previously ignoring unstructured run config passed into a configurable resource. This PR fixes to dump config values into said configurable resource.
- We were erroneously allowing `EnvVar` and `IntEnvVar` to be passed in unstructured run config - and then promptly treating it as just a string. This PR adds an error message to point users in correct directions.
- Enums were being incorrectly converted from pythonic space to legacy config space - meaning you would receive different config values depending on where an enum value was set. This PR unifies to always use `dagster.Enum.from_python_enum` to perform pythonic enum to dagster enum conversion. Furthermore, it changes `from_python_enum` to set the dagster enum's values to be the values of the pythonic enum, rather than the _symbolic_ values. This avoids the weird multiple enums in same python execution bug I discuss later.
- The resolution of env vars in `_resolve_and_update_env_vars` was actually the only way that nested resource config was ever getting post-processed - this explicitly adds a post-processing step for that configuration (at resource initialiation time)

I also uncovered something wacky going on with having multiple enums of the same name in the same file - it seems to cause pydantic models to not be able to track equality between those enums after config translation, and causes tests to fail with a pydantic validation error. Because of the fact that we were passing only the enum _values_ and not the symbolic value, we weren't hitting this weird equality thing. Instead forcing everything use enum values only, means we never hit this.

### DIFF
--- a/python_modules/dagster/dagster/_config/config_type.py
+++ b/python_modules/dagster/dagster/_config/config_type.py
@@ -343,7 +343,37 @@ class Enum(ConfigType):
                 config_schema={"color": Field(Enum.from_python_enum(Color))}
             )
             def select_color(context):
-                # ...
+                assert context.op_config["color"] == Color.RED
+        """
+        if name is None:
+            name = enum.__name__
+        return cls(name, [EnumValue(v.name, python_value=v) for v in enum])
+
+    @classmethod
+    def from_python_enum_direct_values(cls, enum, name=None):
+        """Create a Dagster enum corresponding to an existing Python enum, where the direct values are passed instead of symbolic values (IE, enum.symbol.value as opposed to enum.symbol).
+
+        This is necessary for internal usage, as the symbolic values are not serializable.
+
+        Args:
+            enum (enum.EnumMeta):
+                The class representing the enum.
+            name (Optional[str]):
+                The name for the enum. If not present, `enum.__name__` will be used.
+
+        Example:
+        .. code-block:: python
+
+            class Color(enum.Enum):
+                RED = enum.auto()
+                GREEN = enum.auto()
+                BLUE = enum.auto()
+
+            @op(
+                config_schema={"color": Field(Enum.from_python_enum(Color))}
+            )
+            def select_color(context):
+                assert context.op_config["color"] == Color.RED.value
         """
         if name is None:
             name = enum.__name__
--- a/python_modules/dagster/dagster/_config/config_type.py
+++ b/python_modules/dagster/dagster/_config/config_type.py
@@ -328,9 +328,11 @@ class Enum(ConfigType):
         )
 
     def post_process_translate_enum(self, value):
-        """Pass through enum value as is (assumes that config processing has already occurred once).
-        """
-        return value
+        """Pass through enum value as is (treats config processing as idempotent)."""
+        for ev in self.enum_values:
+            if ev.python_value == value:
+                return ev.python_value
+        return self.post_process(value)
 
     @classmethod
     def from_python_enum(cls, enum, name=None):
--- a/python_modules/dagster/dagster/_config/config_type.py
+++ b/python_modules/dagster/dagster/_config/config_type.py
@@ -97,9 +97,6 @@ class ConfigType:
         """
         return value
 
-    def translate_enum_value(self, value):
-        return self.post_process(value)
-
     def get_snapshot(self) -> "ConfigTypeSnap":
         from .snap import snap_from_config_type
 
@@ -169,9 +166,6 @@ class Float(BuiltinConfigScalar):
 
     def post_process(self, value):
         return float(value)
-
-    def translate_enum_value(self, value):
-        return self.post_process(value)
 
 
 class Any(ConfigType):
@@ -326,13 +320,6 @@ class Enum(ConfigType):
                 "Should never reach this. config_value should be pre-validated. Got {config_value}"
             ).format(config_value=value)
         )
-
-    def translate_enum_value(self, value):
-        """Pass through enum value as is (treats config processing as idempotent)."""
-        for ev in self.enum_values:
-            if ev.python_value == value:
-                return ev.python_value
-        return self.post_process(value)
 
     @classmethod
     def from_python_enum(cls, enum, name=None):
--- a/python_modules/dagster/dagster/_config/config_type.py
+++ b/python_modules/dagster/dagster/_config/config_type.py
@@ -347,7 +347,7 @@ class Enum(ConfigType):
         """
         if name is None:
             name = enum.__name__
-        return cls(name, [EnumValue(v.name, python_value=v) for v in enum])
+        return cls(name, [EnumValue(v.name, python_value=v.value) for v in enum])
 
 
 class ScalarUnion(ConfigType):
--- a/python_modules/dagster/dagster/_config/config_type.py
+++ b/python_modules/dagster/dagster/_config/config_type.py
@@ -97,7 +97,7 @@ class ConfigType:
         """
         return value
 
-    def post_process_translate_enum(self, value):
+    def translate_enum_value(self, value):
         return self.post_process(value)
 
     def get_snapshot(self) -> "ConfigTypeSnap":
@@ -170,7 +170,7 @@ class Float(BuiltinConfigScalar):
     def post_process(self, value):
         return float(value)
 
-    def post_process_translate_enum(self, value):
+    def translate_enum_value(self, value):
         return self.post_process(value)
 
 
@@ -327,7 +327,7 @@ class Enum(ConfigType):
             ).format(config_value=value)
         )
 
-    def post_process_translate_enum(self, value):
+    def translate_enum_value(self, value):
         """Pass through enum value as is (treats config processing as idempotent)."""
         for ev in self.enum_values:
             if ev.python_value == value:
--- a/python_modules/dagster/dagster/_config/config_type.py
+++ b/python_modules/dagster/dagster/_config/config_type.py
@@ -97,6 +97,9 @@ class ConfigType:
         """
         return value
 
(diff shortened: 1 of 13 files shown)


dpeng817:
* master
  * **PR #14306** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14306" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14306?utm_source=stack-comment).

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -891,17 +873,30 @@ def _with_updated_values(
...
+        enum_translation = translate_enums(self._schema.config_type, config_dict)
+        if not enum_translation.success:
+            raise DagsterInvalidConfigError(
+                f"Error while processing {self.__class__.__name__} config ",
+                enum_translation.errors,
+                context.resource_config,
+            )
+        check.not_none(enum_translation.value)
+
+        return self._with_updated_values(enum_translation.value)  # type: ignore (must be a dict)
```
schrockn, line 899: consider doing a check and then a `cast` rather than a `type: ignore`

python_modules/dagster/dagster_tests/core_tests/resource_tests/pythonic_resources/test_env_vars.py
```diff
@@ -108,10 +262,13 @@ def hello_world_asset(writer: WriterResource):
...
+            # I do not think this should work. Worth discussing in PR.
+            # .execute_in_process(
+            #     {"resources": {"writer": {"config": {"prefix": EnvVar("MY_PREFIX_FOR_TEST")}}}}
+            # )
             .execute_in_process(
-                {"resources": {"writer": {"config": {"prefix": EnvVar("MY_PREFIX_FOR_TEST")}}}}
-            )
-            .success
+                {"resources": {"writer": {"config": {"prefix": {"env": "MY_PREFIX_FOR_TEST"}}}}}
+            ).success
```
schrockn, line 271: Undo this to preserve existing behavior, but we should discuss changing this behavior when ben gets back

schrockn (dismissed):

schrockn:
Also make sure to write up nice PR summary prior to commit

python_modules/dagster/dagster_tests/core_tests/resource_tests/pythonic_resources/test_env_vars.py
schrockn: @dpeng817 This is the case we were discussing on the phone

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -891,17 +873,26 @@ def _with_updated_values(
...
+        )  # Converts any EnvVars to their proper config representation (this is not handled when passing EnvVar to a raw run config dictionary)
+        # Perform post-processing step which also resolves any environment variables
+        pydantic_values = translate_to_pydantic(self._schema.config_type, config_dict)
+        if not pydantic_values.success:
+            raise DagsterInvalidConfigError(
+                f"Error while processing {self.__class__.__name__} config ",
+                pydantic_values.errors,
+                context.resource_config,
+            )
+        check.not_none(pydantic_values.value)
```
schrockn, line 893: `not_none` is redundant

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -891,17 +873,26 @@ def _with_updated_values(
         )
         return out
 
-    def _resolve_and_update_env_vars(self) -> "ConfigurableResourceFactory[TResValue]":
-        """Processes the config dictionary to resolve any EnvVar values. This is called at runtime
-        when the resource is initialized, so the user is only shown the error if they attempt to
-        kick off a run relying on this resource.
-
+    def _translate_processed_run_config_to_pydantic(
```
schrockn, line 876: s don't think this method name is accurate. We're creating a new copy of the pythonic resource sourced from the incoming run config.

`def _with_resource_config(self, resource_config)`

python_modules/dagster/dagster/_config/traversal_context.py
```diff
@@ -12,6 +12,7 @@ class TraversalType(Enum):
     VALIDATE = "VALIDATE"
     RESOLVE_DEFAULTS = "RESOLVE_DEFAULTS"
     RESOLVE_DEFAULTS_AND_POSTPROCESS = "RESOLVE_DEFAULTS_AND_POSTPROCESS"
+    TRANSLATE_TO_PYDANTIC = "TRANSLATE_TO_PYDANTIC"
```
schrockn, line 15: `MAKE_PYDANTIC_COMPATIBLE` maybe?

schrockn: translate a bit of a weird word

python_modules/dagster/dagster/_config/config_type.py
```diff
@@ -321,6 +327,13 @@ def post_process(self, value: typing.Any) -> typing.Any:
             ).format(config_value=value)
         )
 
+    def translate_enum_value(self, value):
+        """Pass through enum value as is (treats config processing as idempotent)."""
+        for ev in self.enum_values:
+            if ev.python_value == value:
+                return ev.python_value
+        return self.post_process(value)
```
schrockn, line 335: Wait I'm a bit confused. Is this even right?

Should we not be converting back to `ev.config_value`?

schrockn (changes requested):

schrockn:
Per our in-person discussion, we have figured out what is going on here, in that we were incorrectly mapping enums from Pythonic space into the legacy config space. Additionally we were allowing `EnvVar` into the raw `run_config` argument which was complicating the entire situation. We are going call that a bug and fix it.

- Fix enums in pythonic space so that they project the enum symbolic value into the config dictionary.
    - When you do can do that, you can pass the enum symbolic value directly into the pythonic resource
    - Fix this in `_config_type_for_type_on_pydantic_field`
    - `DagsterEnum.from_python_enum`
- We make case where the EnvVar object passed directly into run_config ***not*** work. That means that we don’t have to consider that case in the `resource_fn` that constructs the pythonic resource instance.
    - Then we can just call `with_updated_values(context.resource_config)` instead of `_resolve_and_update_env_vars`

schrockn (dismissed):
Per comment I just left

python_modules/dagster/dagster/_config/config_type.py
dpeng817: Just for historical sake, this is not right - result of a mismatched understanding due to the enum bug nick discusses in a later comment.

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -1784,11 +1749,23 @@ def separate_resource_params(data: Dict[str, Any]) -> SeparatedResourceParams:
...
     stack: contextlib.ExitStack, obj: ResourceDefinition, context: InitResourceContext
 ) -> Any:
+    from dagster._config.validate import process_config
+
     if isinstance(obj.config_schema, ConfiguredDefinitionConfigSchema):
         value = cast(Dict[str, Any], obj.config_schema.resolve_config({}).value)
         context = context.replace_config(value["config"])
     elif obj.config_schema.default_provided:
-        context = context.replace_config(obj.config_schema.default_value)
+        unprocessed_config = obj.config_schema.default_value
```
schrockn, line 1758: Oh man good catch.

This is worth an extended comment explaining how by the time we get here:

*  `context.resource_config` has _already_ been post processed.
*  The nested resource's config has _also_ been post processed. However it is not available at this point in the stack. The nested resources's post processed config is in the broader run_config dictionary. (e.g. it would live somewhere in process as `run_config['resources']['nested_resource']` if the nested resource had key `nested_resource`
* The only way we have access to the information here is through the unprocessed default value, so in the current state of things we have to get the default value, process it, and jam it into the resource config.

python_modules/dagster/dagster/_config/config_type.py
```diff
@@ -347,7 +347,7 @@ def select_color(context):
         """
         if name is None:
             name = enum.__name__
-        return cls(name, [EnumValue(v.name, python_value=v) for v in enum])
+        return cls(name, [EnumValue(v.name, python_value=v.value) for v in enum])
```
schrockn, line 350: Let's make a new method instead of changing this one. There very well might be someone somewhere calling this. No reason to risk it. Just add a new method.

schrockn (approved):
Please heed final comment. Thanks for digging in here.

dpeng817:
Tracked down the python enum weirdness:
- In python_modules/dagster/dagster/_config/field_utils.py we have a module scoped registry FIELD_HASH_CACHE that maps a “config type” key to an instance of a particular config type
- For the Enum config type, we set the key to be the “name” of the enum.
- For Enum.for_python_enum, name is set to the passed-in enum’s class name, and the values are set to the symbolic values for those enums. So if two different enums are passed in with the same class name, they will collide.
- Because we use the symbolic values as the enum values, the collision behavior is not memoizable.

The solution here is to pass the direct values instead of the enum symbolic values through to the dagster enum type. In order to not break people who may be using `from_python_enum`, I added a new method `from_python_enum_direct_value` which handles things in this different way.

PR #14313, opened 2023-05-16
Add link to GraphQL examples in tests

jmsanders:
There are a lot of useful GraphQL snippets in our test suite. Let's link to them from the documentation.


gibsondan:
Hmm not sure about the UX here, I think some of those examples are fairly old and janky? I see a "test_solids.py" in there, for example

erinkcochran87 (approved):

gibsondan (changes requested):
not clear to me that this is net good, but open to discussing

I wonder if its worth suggesting that you check the request your browser makes in the network tab when Dagit is doing what you want to do? that is more likely to be up to date i think

PR #14314, opened 2023-05-16
[ui] Show an instructional tooltip about zoom modifier keys

bengotow:
## Summary & Motivation

This is a small tooltip that appears beside the zoom control when you start using the scroll wheel (to attempt to zoom) and disappears for the remainder of the time the component is mounted once you use a modifier key. Our goal here is to compensate for the lack of discoverability of the zoom functionality for users who do not have trackpads (pinch to zoom) and are not familiar with Figma, Photoshop, et. al.

https://www.loom.com/share/e206ec32d0724607ac45703ef0fc1d12

## How I Tested These Changes

I verified this manually in Chrome, Firefox, Edge and Safari. I used user-agent overrides in Chrome to verify that the key changes from "Cmd" to "Ctrl" for non-Mac users.

### DIFF
--- a/js_modules/dagit/packages/core/src/graph/SVGViewport.tsx
+++ b/js_modules/dagit/packages/core/src/graph/SVGViewport.tsx
@@ -1,4 +1,4 @@
-import {Box, Colors, Icon, IconWrapper, Slider, Tag} from '@dagster-io/ui';
+import {Box, Colors, FontFamily, Icon, IconWrapper, Slider} from '@dagster-io/ui';
 import animate from 'amator';
 import * as React from 'react';
 import ReactDOM from 'react-dom';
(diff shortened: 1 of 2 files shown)


js_modules/dagit/packages/core/src/graph/SVGViewport.tsx
```diff
@@ -596,3 +597,46 @@ const ZoomSliderContainer = styled.div`
   padding-bottom: 0;
   background: rgba(245, 248, 250, 0.4);
 `;
+
+const WheelInstructionTooltip: React.FC = () => {
```
hellendag, line 601: Can skip the `React.FC` annotation, it doesn't get us anything anymore

js_modules/dagit/packages/core/src/graph/SVGViewport.tsx
```diff
@@ -596,3 +597,46 @@ const ZoomSliderContainer = styled.div`
...
+      clearTimeout(timeout.current);
+    };
+  }, []);
+
+  const zoomKey = navigator.userAgent.includes('Mac') ? '⌘' : 'Ctrl';
+  const visible = wheeling && !usedMeta;
+
+  return (
+    <WheelInstructionTooltipContainer style={{opacity: visible ? 1 : 0}}>
+      <Tag>{`Hold ${zoomKey} to Zoom`}</Tag>
```
hellendag, line 631: s/Zoom/zoom

js_modules/dagit/packages/core/src/graph/SVGViewport.tsx
```diff
@@ -596,3 +597,46 @@ const ZoomSliderContainer = styled.div`
...
+const WheelInstructionTooltip: React.FC = () => {
+  const [usedMeta, setUsedMeta] = React.useState(false);
+  const [wheeling, setWheeling] = React.useState(false);
+  const timeout = React.useRef<NodeJS.Timeout>();
+
+  React.useEffect(() => {
+    const listener = (e: WheelEvent) => {
+      clearTimeout(timeout.current);
+      if (e.metaKey || e.shiftKey || e.ctrlKey) {
+        setUsedMeta(true);
```
hellendag, line 610: It seems like `usedMeta` can never be set to `false` -- do you definitely need both `usedMeta` and `wheeling` state variables? Or could it just be a single `visible` state value?

hellendag (approved):

PR #14304, opened 2023-05-15
feat(dbt-assets)[5/N]: add jaffle shop dbt project

rexledesma:
## Summary & Motivation
Essentially adds the raw `jaffle_shop` dbt project as a demo for showing off the new dbt assets APIs.

Although we mention `jaffle_shop` in other examples, we subtly modify the example. Here, we don't do that. We just copy over https://github.com/dbt-labs/jaffle_shop_duckdb wholesale.

## How I Tested These Changes
N/A

PR #14316, opened 2023-05-16
feat(dbt-assets)[4/N]: add subsetting implementation

rexledesma:
## Summary & Motivation
Within Dagster, one computation can model the materialization of multiple assets. However, there are instances where this computation is flexible enough that, when specified by the user, it can optionally materialize a subselection of its assets.

This behavior is called [Subsetting](https://docs.dagster.io/concepts/assets/multi-assets#subsetting-multi-assets).

Historically, because we constrained the compute function in our dbt integration to essentially be either `dbt run` or `dbt build`, it was easy for us to also implement subsetting on the user's behalf. That way, they could achieve this feature out of the box.

Now, we have decided to liberalize the compute function in the `dbt` integration, allowing the user to specify this compute function. Consequentially, the user will also need to implement subsetting in the body of their computation in order to achieve this feature.

Here, we provide a utility to accomplish this use case. When the user passes in the Dagster execution context to their dbt resource that controls CLI invocation, we can:
- Automatically construct the proper dbt selection arguments, and
- Pass the selection arguments into the CLI invocation on behalf of the user.

## How I Tested These Changes
pytest

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt_tests/cli/test_resources_v2.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt_tests/cli/test_resources_v2.py
@@ -1,8 +1,9 @@
-from typing import List
+from typing import List, Optional
 
 import pytest
 from dagster import AssetObservation, Output
 from dagster_dbt.cli import DbtCli, DbtCliEventMessage, DbtManifest
+from pytest_mock import MockerFixture
 
 from ..conftest import TEST_PROJECT_DIR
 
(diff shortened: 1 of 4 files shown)


rexledesma:
* master
    * **PR #14267** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14267" target="_blank">Graphite</a> 
      * **PR #14289** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14289" target="_blank">Graphite</a> 
        * **PR #14316** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14316" target="_blank">Graphite</a>  👈
        * **PR #14304** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14304" target="_blank">Graphite</a> 
          * **PR #14301** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14301" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14316?utm_source=stack-comment).

python_modules/libraries/dagster-dbt/dagster_dbt_tests/cli/test_resources_v2.py
```diff
@@ -25,6 +26,58 @@ def test_dbt_cli(global_config: List[str], command: str) -> None:
...
+
+
+def test_dbt_cli_default_selection(mocker: MockerFixture) -> None:
+    mock_context = mocker.MagicMock()
+    mock_selected_output_names = ["least_caloric", "sort_by_calories"]
+
+    type(mock_context.op).tags = mocker.PropertyMock(
+        return_value={
+            "dagster-dbt/select": "fqn:*",
+            "dagster-dbt/exclude": "fqn:*",
```
OwenKephart, line 62: this bit is confusing to me -- isn't the default exclude "None" (or more accurately, no entry in the dictionary)?

it's not really relevant to this test as we're not actually running the command, but might be confusing for someone coming across this test, maybe swap out the exclude to some fancy selection string

python_modules/libraries/dagster-dbt/dagster_dbt_tests/cli/test_resources_v2.py
```diff
@@ -25,6 +26,58 @@ def test_dbt_cli(global_config: List[str], command: str) -> None:
     assert dbt_cli_task.process.returncode == 0
 
 
+def test_dbt_cli_subsetted_execution(mocker: MockerFixture) -> None:
+    mock_context = mocker.MagicMock()
+    mock_selected_output_names = ["least_caloric", "sort_by_calories"]
+
+    type(mock_context.op).tags = mocker.PropertyMock(return_value={})
```
OwenKephart, line 33: would be nice to test this in a context where select / exclude were set by the user, just to verify that those don't override the subselection (not that I think they do)

OwenKephart (approved):
some minor comments on the tests but the code itself looks good

PR #5429, opened 2021-10-28
dagster-k8s run worker fault tolerance

johannkm:
Add support for the monitoring daemon on dagster-k8s. This entails:
- support resume_run and check_run_worker_health on the launcher
- helm support
- integration tests

This diff has exposed a gap in the check_run_worker_health API that can be fixed in a later diff- that api should include the resume attempt number. Here, we recalculate that unnecessarily.

PR #14323, opened 2023-05-16
Make default max_resume_run_attempts value 0, not 3

gibsondan:
Summary:
This should only come into play if the maxResumeRunAttempts value in the helm chart is explicitly set to None or nil somehow - but if it does, we should default to 0, not 3.

Test Plan: Bk

### DIFF
--- a/helm/dagster/schema/schema_tests/test_dagster_daemon.py
+++ b/helm/dagster/schema/schema_tests/test_dagster_daemon.py
@@ -247,6 +247,23 @@ def test_run_monitoring_disabled(
     assert "run_monitoring" not in instance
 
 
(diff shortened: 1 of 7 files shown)


alangenfeld (approved):
i was confused how we got here but looks like just something that got missed going from https://github.com/dagster-io/dagster/pull/5429 to https://github.com/dagster-io/dagster/pull/11512

gibsondan:
I think the winding road is:
- run monitoring was originally introduced only for this experimental feature, so setting a non-zero default made sense
- run monitoring later expanded to less esoteric features, but the default value for the experimental value remained
- run monitoring turned on by default, default value for the resume run functionality became more of an issue if it wasn't explicitly set

PR #14324, opened 2023-05-16
Revert "Revert "Change AutoMaterializeAssetEvaluation to store serialized subsets (#14295)""

johannkm:

PR #14276, opened 2023-05-12
Auto materialize asset evaluations gql

johannkm:
Feels a bit overengineered but matches what we're doing with the python namedtuples and will support us adding more metadata to different conditions

PR #14329, opened 2023-05-17
Auto materialize asset evaluations gql with partitions

johannkm:
Naive implementation of surfacing partition subsets for the various conditions. Uses a raw list of partition keys. Followup will add partition key ranges for time windows


johannkm:
* master
  * **PR #14324** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14324" target="_blank">Graphite</a> 
    * **PR #14319** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14319" target="_blank">Graphite</a> 
      * **PR #14276** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14276" target="_blank">Graphite</a> 
        * **PR #14329** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14329" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14329?utm_source=stack-comment).

python_modules/dagster-graphql/dagster_graphql/schema/auto_materialize_asset_evaluations.py
```diff
@@ -65,23 +70,43 @@ class Meta:
...
+    ParentOutdatedAutoMaterializeCondition: GrapheneParentOutdatedAutoMaterializeCondition,
+}
+
+
+def create_graphene_auto_materialize_condition(
+    condition: Union[AutoMaterializeCondition, Tuple[AutoMaterializeCondition, str]],
+    partitions_def: Optional[PartitionsDefinition],
+):
+    if partitions_def is not None:
+        if not len(condition) == 2 and isinstance(condition[1], str):
```
OwenKephart, line 87: In the event that we add a new field to one of these `AutoMaterializeCondition`s whose type is `str`, then this check will incorrectly identify that as a the second case.

Personally, I prefer the:

`isinstance(condition, AutoMaterializeCondition.__args__)`

thing that you had previously, as it's guaranteed to stay in sync with the underlying AutoMaterializeCondition type. Any reservations with that approach?

python_modules/dagster-graphql/dagster_graphql/schema/auto_materialize_asset_evaluations.py
```diff
@@ -65,23 +70,43 @@ class Meta:
...
-        return GrapheneParentOutdatedAutoMaterializeCondition(decisionType=condition.decision_type)
+        subset = partitions_def.deserialize_subset(serialized_subset)
+        partition_keys = subset.get_partition_keys()
     else:
-        check.failed(f"Unexpected condition type {type(condition)}")
+        auto_materialize_condition = cast(AutoMaterializeCondition, condition)
+        partition_keys = None
+
+    check.invariant(
+        type(auto_materialize_condition) in DAGSTER_NAMEDTUPLE_TO_GRAPHENE_MAP,
```
OwenKephart, line 102: sort of combining the above two comments, but I think we should get rid of this map

python_modules/dagster-graphql/dagster_graphql/schema/auto_materialize_asset_evaluations.py
```diff
@@ -65,23 +70,43 @@ class Meta:
...
+        auto_materialize_condition, serialized_subset = cast(
+            Tuple[AutoMaterializeCondition, str], condition
         )
-    elif isinstance(condition, MissingAutoMaterializeCondition):
-        return GrapheneMissingAutoMaterializeCondition(decisionType=condition.decision_type)
-    elif isinstance(condition, ParentOutdatedAutoMaterializeCondition):
-        return GrapheneParentOutdatedAutoMaterializeCondition(decisionType=condition.decision_type)
+        subset = partitions_def.deserialize_subset(serialized_subset)
+        partition_keys = subset.get_partition_keys()
     else:
```
OwenKephart, line 97: Imagine the sequence of events:

1. partitioned asset is created
2. auto materializations are kicked off for it, resulting in serialized evaluations
3. that asset's partitions definition is removed

In this codepath, we're using the asset's current partitions definitions to determine what format we're expecting of the serialized data, so we would end up erroring, rather than more gracefully degrading in some way. I leave it to the designers of the world to figure out what exactly we should do there, but ideally in the meantime, users wouldn't see graphql errors in this case. Maybe this function returns None in the case that we detect partitioned data for an asset that does not currently have a partitions definition.

A similar thing should be done for the case that `partitions_def.can_deserialize_subset` is false.

python_modules/dagster-graphql/dagster_graphql/schema/auto_materialize_asset_evaluations.py
```diff
@@ -65,23 +70,43 @@ class Meta:
         )
 
 
-def create_graphene_auto_materialize_condition(condition: AutoMaterializeCondition):
-    if isinstance(condition, FreshnessAutoMaterializeCondition):
```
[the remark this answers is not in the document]
johannkm: Yeah I was just being lazy and didn't feel like passing a new argument to each of them lol

python_modules/dagster-graphql/dagster_graphql/schema/auto_materialize_asset_evaluations.py
johannkm: With the silly type detection challenges as they are we should just persist an is_partitioned bool on each evaluation

And yeah it should handle cases where it can't deserialize

python_modules/dagster-graphql/dagster_graphql/implementation/fetch_auto_materialize_asset_evaluations.py
```diff
@@ -0,0 +1,46 @@
...
+    asset_node: Optional[ExternalAssetNode] = None
+    for _, _, external_asset_node in asset_node_iter(graphene_info):
+        if external_asset_node.asset_key == asset_key:
+            asset_node = external_asset_node
+            break
+
+    if not asset_node:
+        raise DagsterError(f"Asset {asset_key} not found in workspace")
+
+    partitions_def = partitions_def = (
```
clairelin135, line 35: nit: `partitions_def = partitions_def = (` -> `partitions_def = (`

python_modules/dagster-graphql/dagster_graphql/schema/auto_materialize_asset_evaluations.py
clairelin135: I think we ran into a similar issue with deserializing asset backfills in the past.

The solution we landed on was to serialize (a) the partitions def serializable unique ID and (b) class name for the partitions subset for each partitions subset. This enables:
- detecting when a partitions def has been removed.
- for time window partitions subsets, checking the serializable unique ID to see if the partitions def is still the same. If it is not, we can no longer deserialize.
- for other partitions subsets (static/multipartitioned), if the class name is equivalent, we can deserialize since it just stores a set of partition keys.

This is all logic that already exists in `partitions_def.can_deserialize_subset`. I suspect we'll likely want to implement something similar for these cases, though it's up to you on if you want to implement this now or in the future.

johannkm: Added this via the SerializedPartitionsSubset

python_modules/dagster-graphql/dagster_graphql/implementation/fetch_auto_materialize_asset_evaluations.py
```diff
@@ -1,39 +1,60 @@
...
             message=(
                 "Auto materialize evaluations are not getting logged. Run `dagster instance"
                 " migrate` to enable."
             )
         )
 
+    asset_key = AssetKey.from_graphql_input(graphene_asset_key)
+
+    asset_node: Optional[ExternalAssetNode] = None
+    for _, _, external_asset_node in asset_node_iter(graphene_info):
```
clairelin135, line 42: I think we should use `get_asset_nodes_by_asset_key` instead.

`asset_node_iter` iterates through all code locations, repositories, and asset in order, which means you might get a source asset with the same key rather than the actual asset node. This is probably fine with the current implementation, but may lead to unexpected behaviors if we do more things with the external asset node in the future

python_modules/dagster-graphql/dagster_graphql/schema/auto_materialize_asset_evaluations.py
```diff
@@ -79,26 +81,40 @@ class Meta:
...
 def create_graphene_auto_materialize_condition(
-    condition_tuple: Tuple[AutoMaterializeCondition, Optional[SerializedPartitionsSubset]]
+    condition_tuple: Tuple[AutoMaterializeCondition, Optional[SerializedPartitionsSubset]],
+    partitions_def: Optional[PartitionsDefinition],
 ):
-    condition, _ = condition_tuple
+    condition, serialized_partition_subset = condition_tuple
+
+    if serialized_partition_subset and serialized_partition_subset.can_deserialize(partitions_def):
+        subset = serialized_partition_subset.deserialize(partitions_def)
```
clairelin135, line 90: what's the expected behavior when the partitions def has changed and the partitions subset can no longer be deserialized?

clairelin135: IMO, we should display a different message, something like "the partitions def has changed and the auto materialize condition can no longer be displayed".

johannkm: added an error message

python_modules/dagster-graphql/dagster_graphql/schema/auto_materialize_asset_evaluations.py
```diff
@@ -22,8 +23,30 @@
...
+
+class GraphenePartitionSubsetDeserializationError(graphene.ObjectType):
+    message = graphene.NonNull(graphene.String)
+
+    class Meta:
+        interfaces = (GrapheneError,)
+        name = "PartitionSubsetDeserializationError"
+
+
+class GrapheneParititonKeysOrError(graphene.Union):
```
clairelin135, line 41: nit: `Parititon`

clairelin135 (approved):
LGTM!

PR #14335, opened 2023-05-17
Do not initialize instance in cases where schedules and sensors have no resources defined

schrockn:
## Summary & Motivation

A user reported that during an upgrade from 1.1.9 to 1.3.4 they were required to access an environment variable they had not been required to access before. The issue is the change to support resources in schedules and sensors added a code path that unconditionally constructed the instance where it did not before. This PR reverts to the old behavior.

However I do not think this is the right fix. Allowing users to deploy code locations that do not have access to the instance is extremely fragile. Any access to `context.instance` can trigger a fatal, and if we add any feature at any time.

The core problem here is the wildly complicated state of all our initialization pathways, where everything is "Optional" when in fact they are not. It seems that these evolved this way to enable convenient test pathways, or just out of sheer laziness, but we are now left in a state where there is a combinatorial explosion of possible states in all of our context objects, and further more on top of that a combinatorial explosion of initialization codepaths, as we seem to lazily create things whenever it pleases us. This will result in a continual stream of difficult-to-predict bugs as it is super fragile, as this bug demonstrates.

## How I Tested These Changes

Included unit tests

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/schedule_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/schedule_definition.py
@@ -378,6 +378,7 @@ def build_schedule_context(
     scheduled_execution_time: Optional[datetime] = None,
     resources: Optional[Mapping[str, object]] = None,
     repository_def: Optional["RepositoryDefinition"] = None,
+    instance_ref: Optional["InstanceRef"] = None,
 ) -> ScheduleEvaluationContext:
     """Builds schedule execution context using the provided parameters.
 
@@ -401,7 +402,11 @@ def build_schedule_context(
     check.opt_inst_param(instance, "instance", DagsterInstance)
 
     return ScheduleEvaluationContext(
-        instance_ref=instance.get_ref() if instance and instance.is_persistent else None,
+        instance_ref=instance_ref
+        if instance_ref
+        else instance.get_ref()
+        if instance and instance.is_persistent
+        else None,
         scheduled_execution_time=check.opt_inst_param(
             scheduled_execution_time, "scheduled_execution_time", datetime
         ),
--- a/python_modules/dagster/dagster/_core/definitions/schedule_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/schedule_definition.py
@@ -243,9 +243,11 @@ class ScheduleEvaluationContext:
 
         if not self._resources:
             # Early exit if no resources are defined. This skips unnecessary initialization
-            # entirely. This also fixes an issue where a user had a deployment that depended
-            # on not setting environment variables necessary to access the instance, and then
-            # an upgrade broke them.
+            # entirely. This allows users to run user code servers in cases where they
+            # do not have access to the instance if they use a subset of features do
+            # that do not require instance access. In this case, if they do not use
+            # resources on schedules they do not require the instance, so we do not
+            # instantiate it
             if not self._resource_defs:
                 self._resources = ScopedResourcesBuilder.build_empty()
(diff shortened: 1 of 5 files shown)


sryza:
Added @clairelin135 to this review because she ran up against a similar issue with dynamic partitions - I believe we initially required instance access in the code servers and then walked this back when users complained. There was an issue filed by a user, but I'm having trouble finding it.

python_modules/dagster/dagster/_core/definitions/schedule_definition.py
```diff
@@ -242,9 +242,17 @@ def resources(self) -> Resources:
         from dagster._core.execution.build_resources import build_resources
 
         if not self._resources:
+            # Early exit if no resources are defined. This skips unnecessary initialization
+            # entirely. This also fixes an issue where a user had a deployment that depended
+            # on not setting environment variables necessary to access the instance, and then
+            # an upgrade broke them.
```
gibsondan, line 248: comment nit: i think this could mention the situation we are hoping to guard against (users running gRPC servers that can't load the instance) rather than the specific upgrade bug

gibsondan (approved):
Thoughts:

- I agree the current state of things potentially breaking as soon as something calls context.instance is not great.

- Requiring the user code server to be able to load the instance seems like a reasonable requirement to me - I think we used to have that, and removed it when we split the helm chart out into a user code helm chart and a system helm chart. With the benefit of hindsight I think allowing user code to work without an instance there was probably an error since you need to be able to load the instance anyway in order to do lots of things. Loading the instance once on user code server startup and passing it through / always being able to assume that it exists and has loaded already would be a lot simpler then all these lazy-loaded instance ref shenanigans (and probably more performant too - right now I believe we are creating a separate DagsterInstance object on every tick).

- it would be easier to guarantee that the instance will always load if we moved to a UserCodeLauncher model where dagster is responsible for spinning up the user code servers and setting all the needed secrets on them, this is how cloud works already today.

- I don't actually think optional args is the core problem here, that may be a problem for other reasons but the arg was set in the case where it broke for the user, it just didn't load when it instantiated because a secret was missing.

- the other angle you could take here is threading the lazy-loading even further down into the resource generation code, and have it wait to load the instance until there's a resource that actually needs to access the instance in its initialization function (I would guess that these are fairly rare). But that is probably doubling down on a bad setup.

- Overall I think we should land this and have a plan for a future where the instance is loaded once always on gRPC server startup

schrockn:
@gibsondan do you have any suggestions on how to add an automated test to catch this. With the way this codepath is structured it's a challenging one

gibsondan:
What i would suggest is making a pluggable instance component that raises an Exception in its constructor, constructing an InstanceRef that uses that component, making a schedule context passing in that instance_ref, and calling context.resources on it

gibsondan:
alternately creating an instance ref that uses an existing component but references an env var that does not exist in its config could also work (the tricky part is creating the instance ref without it throwing the exception)

schrockn:
> I don't actually think optional args is the core problem here, that may be a problem for other reasons but the arg was set in the case where it broke for the user, it just didn't load when it instantiated because a secret was missing.

Completely disagree. If we were disciplined about the lifecycle of instance ref and instance and had only one init path rather than a bunch of parallel lazy initialization scattered throughout codebase this all would have been avoided.

gibsondan:
ok I think we're on pretty similar pages actually - I was just saying that "If we were disciplined about the lifecycle of instance ref and instance" was the core problem, and the optional arg is a follow-on symptom of that.

PR #14339, opened 2023-05-17
[ui] Fix Run timeline resizing

hellendag:
## Summary & Motivation

When I refactored the run timeline to be virtualized, I broke the resizing behavior that expands/shrinks the rows to fit the window. Restore this. ~and apply a debounce to the width setter so that the resize doesn't risk a crunchy animation.~

## How I Tested These Changes

View run timeline in app and in Storybook, resize window. Verify that the timeline rows and chunks re-render properly.

### DIFF
--- a/js_modules/dagit/packages/core/src/runs/RunTimeline.tsx
+++ b/js_modules/dagit/packages/core/src/runs/RunTimeline.tsx
@@ -9,6 +9,7 @@ import {
   Icon,
   Spinner,
   MiddleTruncate,
+  useViewport,
 } from '@dagster-io/ui';
 import {useVirtualizer} from '@tanstack/react-virtual';
 import * as React from 'react';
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/core/src/runs/RunTimeline.tsx
```diff
@@ -127,10 +128,12 @@ export const RunTimeline = (props: Props) => {
...
-      observer.current = new ResizeObserver((entries) => {
-        const entry = entries[0];
-        setWidth(entry.contentRect.width);
-      });
+      observer.current = new ResizeObserver(
+        debounce((entries) => {
+          const entry = entries[0];
+          setWidth(entry.contentRect.width);
+        }, 100),
+      );
```
salazarm, line 136: Use `useViewport` rather than rolling your own resize observer?

hellendag:
Switched to `useViewport`. It means not having the debounce, but I think it's probably fine without that anyway.

js_modules/dagit/packages/core/src/runs/RunTimeline.tsx
```diff
@@ -73,8 +74,12 @@ interface Props {
 
 export const RunTimeline = (props: Props) => {
   const {loading = false, jobs, range} = props;
-  const [width, setWidth] = React.useState<number | null>(null);
+  // const [width, setWidth] = React.useState<number | null>(null);
```
bengotow, line 77: Rm :-)

bengotow (approved):
This looks good to me! Definitely a fan of re-using that useViewport hook in as many places as we can 👍

PR #14346, opened 2023-05-17
Fix multi asset sensor unconsumed events caching

clairelin135:
Cloud user reported that a multi asset sensor that used `context.latest_materialization_records_by_partition_and_asset` was timing out.

The reason that this occurred was because we implemented filtering event log records by storage IDs in OSS in order to cache unconsumed events: https://github.com/dagster-io/dagster/blob/dbad48a80318130305cc0d197c3ef38251b639f5/python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py#L294-L303
But this functionality was not mirrored in cloud, so in cloud we just did not do any filtering which cause the `get_event_records` method to fetch an unbounded number of records.

This PR accompanies the cloud fix https://github.com/dagster-io/internal/pull/5707 and adds tests for this functionality.

**Test Plan**
- Adds tests to the multi asset sensor context to assert that only unconsumed events are cached
- Adds tests to event log storage to assert that filtering by storage ID will occur if a non-empty list of storage IDs is provided

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py
@@ -291,19 +291,19 @@ class MultiAssetSensorEvaluationContext(SensorEvaluationContext):
             return
 
         for asset_key in self._monitored_asset_keys:
-            event_records = self.instance.get_event_records(
-                EventRecordsFilter(
-                    event_type=DagsterEventType.ASSET_MATERIALIZATION,
-                    storage_ids=list(
-                        self._get_cursor(
-                            asset_key
-                        ).trailing_unconsumed_partitioned_event_ids.values()
-                    ),
+            unconsumed_event_ids = list(
+                self._get_cursor(asset_key).trailing_unconsumed_partitioned_event_ids.values()
+            )
+            if unconsumed_event_ids:
+                event_records = self.instance.get_event_records(
+                    EventRecordsFilter(
+                        event_type=DagsterEventType.ASSET_MATERIALIZATION,
+                        storage_ids=unconsumed_event_ids,
+                    )
                 )
-            )
-            self._initial_unconsumed_events_by_id.update(
-                {event_record.storage_id: event_record for event_record in event_records}
-            )
+                self._initial_unconsumed_events_by_id.update(
+                    {event_record.storage_id: event_record for event_record in event_records}
+                )
 
         self._fetched_initial_unconsumed_events = True
 
--- a/python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py
@@ -369,6 +369,7 @@ class MultiAssetSensorEvaluationContext(SensorEvaluationContext):
             self._cursor = new_cursor
             self._unpacked_cursor = MultiAssetSensorContextCursor(new_cursor, self)
             self._cursor_advance_state_mutation = MultiAssetSensorCursorAdvances()
+            self._fetched_initial_unconsumed_events = False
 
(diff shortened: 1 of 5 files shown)


python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py
```diff
@@ -369,6 +369,7 @@ def update_cursor_after_evaluation(self) -> None:
             self._cursor = new_cursor
             self._unpacked_cursor = MultiAssetSensorContextCursor(new_cursor, self)
             self._cursor_advance_state_mutation = MultiAssetSensorCursorAdvances()
+            self._fetched_initial_unconsumed_events = False
```
clairelin135, line 372: The other changes in this PR surfaced an existing bug--this line change fixes the bug.

Before the other changes, we would fetch all events and then cache them, so tests that fetched unconsumed events would pass:
https://github.com/dagster-io/dagster/blob/cfdea8e3e898ef30e4c7f38a65fb23a1a889b05c/python_modules/dagster/dagster/_core/definitions/multi_asset_sensor_definition.py#L284-L291

But now that we fetch only the desired events, this line must be added to basically refetch unconsumed events on every separate sensor invocation.

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -820,7 +820,7 @@ def _apply_filter_to_query(
                 > datetime.utcfromtimestamp(event_records_filter.after_timestamp)
             )
 
-        if event_records_filter.storage_ids:
+        if event_records_filter.storage_ids is not None:
```
alangenfeld, line 823: I think + the change above is risky - version skew will cause any old `EventRecordsFilter` to trigger this clause and I assume get 0 results back.

Whats the context on this? Am I misunderstanding that this would cause the query to return 0 results? Can we not detect this condition at the callsite and just skip asking the DB to give us back an empty list?

clairelin135: Yeah, this value in an old event records filter would be an empty list and thus return 0 results. My motivation for making this change was to fix the existing behavior that passing in an empty list results in no filtering, since I think the expected behavior is to return no results.

I'm a bit confused about how version skew would occur. In OSS it has to be imported from the dagster library, which should ensure that in this version and onwards the `get_event_records` call handles the None vs empty list distinction correctly. Or is it possible version skew can occur on cloud?

clairelin135: Regardless, I'm fine with just detecting the condition at the callsite

python_modules/dagster/dagster_tests/definitions_tests/test_sensor_invocation.py
```diff
@@ -1169,6 +1169,7 @@ def test_multi_asset_sensor_latest_materialization_records_by_partition_and_asse
     @multi_asset_sensor(monitored_assets=[july_asset.key, july_asset_2.key])
     def my_sensor(context):
         events = context.latest_materialization_records_by_partition_and_asset()
+        assert context._initial_unconsumed_events_by_id == {}  # noqa: SLF001
```
clairelin135, line 1172: Adds this test to assert that we are not fetching all events here

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
clairelin135: ahhh I see, thanks for explaining. So it seems like in order to support these cases, in cloud we would still have to do no filtering when an empty list is passed in

alangenfeld (approved):
nit: think the PR title should focus on the material changes here with test updates covered in test plan section

PR #14347, opened 2023-05-17
AssetSelection.source_assets + key_prefixes bugfix

smackesey:
## Summary & Motivation

Internal companion PR: https://github.com/dagster-io/internal/pull/5722

- Add `AssetSelection.upstream_source_assets` method to allow selecting source assets upstream of the current selection. There is no current way to do this. See: https://elementl-workspace.slack.com/archives/C03C3E81TNE/p1683647912931259
- Add optional `include_sources` flag to `AssetSelection.key_prefixes` and `.groups` that toggles inclusion of source assets.
- Update docstrings of several `AssetSelection` methods to clarify that they _do not_ select source assets.
- Rename `AssetSelection.non_source_assets` to `materializable_assets`.
- Fix `AssetGraph.all_assets` (used in selection resolution) so that it returns _all assets_, including source assets. It was not intended that it did not, since there was also `_non_source_assets`, which returned the same thing with a different implementation. Update existing use of `all_asset_keys` that depend on _not_ selecting source assets to instead call `materializable_assets` (new name for `non_source_assets`)

## How I Tested These Changes

- New unit test for `AssetSelection.source_assets`
- Add to unit test for `AssetSelection.key_prefixes`
- Add to unit test for `AssetSelection.groups`

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -102,7 +102,7 @@ class AssetGraph:
         """Non-source asset keys that have no non-source parents."""
         from .asset_selection import AssetSelection
 
-        return AssetSelection.keys(*self.non_source_asset_keys).roots().resolve(self)
+        return AssetSelection.keys(*self.materializable_asset_keys).roots().resolve(self)
 
     @property
     def freshness_policies_by_key(self) -> Mapping[AssetKey, Optional[FreshnessPolicy]]:
@@ -176,11 +176,12 @@ class AssetGraph:
 
     @property
     def all_asset_keys(self) -> AbstractSet[AssetKey]:
-        return self._asset_dep_graph["upstream"].keys()
+        return self.materializable_asset_keys | self.source_asset_keys
 
     @property
-    def non_source_asset_keys(self) -> AbstractSet[AssetKey]:
-        return self._asset_dep_graph["upstream"].keys() - self._source_asset_keys
+    def materializable_asset_keys(self) -> AbstractSet[AssetKey]:
+        # Source assets keys can sometimes appear in the upstream dict
+        return self._asset_dep_graph["upstream"].keys() - self.source_asset_keys
 
     def get_partitions_def(self, asset_key: AssetKey) -> Optional[PartitionsDefinition]:
         return self._partitions_defs_by_key.get(asset_key)
@@ -397,7 +398,9 @@ class AssetGraph:
         )
 
     def is_source(self, asset_key: AssetKey) -> bool:
-        return asset_key in self.source_asset_keys or asset_key not in self.all_asset_keys
+        return (
+            asset_key in self.source_asset_keys or asset_key not in self.materializable_asset_keys
+        )
 
     def has_non_source_parents(self, asset_key: AssetKey) -> bool:
         """Determines if an asset has any parents which are not source assets."""
--- a/python_modules/dagster/dagster/_core/definitions/asset_selection.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_selection.py
@@ -105,8 +105,14 @@ class AssetSelection(ABC):
 
     @public
     @staticmethod
-    def key_prefixes(*key_prefixes: CoercibleToAssetKeyPrefix) -> "KeyPrefixesAssetSelection":
+    def key_prefixes(
+        *key_prefixes: CoercibleToAssetKeyPrefix, include_sources: bool = False
+    ) -> "KeyPrefixesAssetSelection":
         """Returns a selection that includes assets that match any of the provided key prefixes.
+
+        Args:
+            include_sources (bool): If True, then include source assets matching the key prefix(es)
+                in the selection.
 
         Examples:
             .. code-block:: python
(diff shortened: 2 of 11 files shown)


smackesey:
* master
  * **PR #14347** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14347" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14347?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/asset_selection.py
```diff
@@ -213,6 +227,13 @@ def sources(self) -> "RootAssetSelection":
         )
         return self.roots()
 
+    @public
+    def source_assets(self) -> "SourceAssetSelection":
```
sryza, line 231: I think we should call this `upstream_source_assets`. `source_assets` implies that it's filtering down the current selection to only include the entries that are source assets, but this is actually returning entries that aren't in the current selection, if I'm understanding correctly.

python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_selection.py
```diff
@@ -141,6 +141,10 @@ def test_asset_selection_key_prefixes(all_assets: _AssetList):
     sel = AssetSelection.key_prefixes("plants")
     assert sel.resolve(all_assets) == _asset_keys_of(set())
 
+    # works for source assets
+    sel = AssetSelection.key_prefixes("celestial")
```
sryza, line 145: If I understand correctly including source assets with `key_prefixes` is inconsistent with the behavior of `groups`? I would expect them to have the same behavior with respect to source assets.

python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -149,11 +149,11 @@ def from_assets(
 
     @property
     def all_asset_keys(self) -> AbstractSet[AssetKey]:
-        return self._asset_dep_graph["upstream"].keys()
+        return self.materializable_asset_keys | self.source_asset_keys
 
     @property
-    def non_source_asset_keys(self) -> AbstractSet[AssetKey]:
-        return self._asset_dep_graph["upstream"].keys() - self._source_asset_keys
+    def materializable_asset_keys(self) -> AbstractSet[AssetKey]:
```
sryza, line 155: I like the name change.

python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_selection.py
smackesey: I think it's a wash between similarity to `groups` (no source assets) or `keys` (yes source assets). Erring on the side of `keys` since it's better to have more ways to select source assets (the reason `groups` doesn't is just for backcompat). That said, I can change if you feel strongly.

python_modules/dagster/dagster/_core/definitions/asset_selection.py
smackesey: Ok, good point

python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_selection.py
smackesey: >Isn't it also because we don't want to allow people to select both source and non-source assets at once? Or does that fall under backcompat?

Well, at the time `AssetSelection` was designed, there was no user code associated with a source asset, no it didn't make sense for `AssetSelection` to capture them-- that's why none of the older `AssetSelection` methods select source assets.

But in a (future) world where source asset observations can run together with materializing assets, then it makes sense to be able to select them together. IMO the only reason we wouldn't do so for `groups` immediately when this functionality is available is that it would change existing jobs defined with this method-- so, yes, purely backcompat.

sryza: > But in a (future) world where source asset observations can run together with materializing assets, then it makes sense to be able to select them together.

Got it, this makes sense.

In this future, do you have thoughts on how we'd want users to be able to specify "only the observable assets within this group" or "only the materializable assets within this group" or vice versa?

> IMO the only reason we wouldn't do so for groups immediately when this functionality is available is that it would change existing jobs defined with this method-- so, yes, purely backcompat.

FWIW, I think we're technically allowed to do this because observable source assets are experimental? Or are there situations where this would affect how non-observable source assets are handled?  E.g. it could affect the behavior of the `monitored_assets` argument of `multi_asset_sensor`. Although `multi_asset_sensor` is experimental too.

smackesey: >In this future, do you have thoughts on how we'd want users to be able to specify "only the observable assets within this group" or "only the materializable assets within this group" or vice versa?

I've added `include_sources` kwargs that default to `False` for both `key_prefixes` and `groups`-- that makes them behave similarly as filters and preserves backcompat.

>FWIW, I think we're technically allowed to do this because observable source assets are experimental? Or are there situations where this would affect how non-observable source assets are handled?

This will cause job definitions that currently don't execute observable source assets and include nothing experimental to start executing them, so it doesn't seem like it falls under the "we can do it because it's experimental" rule.

smackesey:
@sryza Sorry for letting this one drop, I've addressed your last comment and would like to get this over the line now.

sryza (approved):

PR #14358, opened 2023-05-18
Get asset evaluation by evalution id

johannkm:
Currently we only have a method to get these via a list. A singular fetch by id lets @salazarm give a link to a particular evaluation

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/implementation/fetch_auto_materialize_asset_evaluations.py
+++ b/python_modules/dagster-graphql/dagster_graphql/implementation/fetch_auto_materialize_asset_evaluations.py
@@ -9,6 +9,25 @@ from dagster_graphql.schema.auto_materialize_asset_evaluations import (
 from dagster_graphql.schema.inputs import GrapheneAssetKeyInput
 
 
+def fetch_auto_materialize_asset_evaluation(
+    instance: DagsterInstance, asset_key: GrapheneAssetKeyInput, evaluation_id: int
+) -> Optional[GrapheneAutoMaterializeAssetEvaluationRecord]:
+    if instance.schedule_storage is None:
+        raise DagsterError(
+            "Instance does not have schedule storage configured, cannot fetch evaluations."
+        )
+    if not instance.schedule_storage.supports_auto_materialize_asset_evaluations:
+        raise DagsterError(
+            "Auto materialize evaluations are not getting logged. Run `dagster instance"
+            " migrate` to enable."
+        )
+
+    record = instance.schedule_storage.get_auto_materialize_asset_evaluation_by_evaluation_id(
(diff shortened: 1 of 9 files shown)


python_modules/dagster-graphql/dagster_graphql_tests/graphql/test_auto_materialize_asset_evaluations.py
```diff
@@ -11,7 +11,23 @@
     ExecutingGraphQLContextTestMatrix,
 )
 
-QUERY = """
+SINGLE_QUERY = """
+query GetEvaluationsQuery($assetKey: AssetKeyInput!, $evaluationId: Int!) {
```
salazarm, line 15: this has the same name as the multi query

johannkm: fixed

python_modules/dagster-graphql/dagster_graphql/implementation/fetch_auto_materialize_asset_evaluations.py
```diff
@@ -9,6 +9,25 @@
...
+        raise DagsterError(
+            "Instance does not have schedule storage configured, cannot fetch evaluations."
+        )
+    if not instance.schedule_storage.supports_auto_materialize_asset_evaluations:
+        raise DagsterError(
+            "Auto materialize evaluations are not getting logged. Run `dagster instance"
+            " migrate` to enable."
+        )
+
+    record = instance.schedule_storage.get_auto_materialize_asset_evaluation_by_evaluation_id(
```
OwenKephart, line 25: I'd prefer not to add a new method to the schedule storage if we can avoid it, as this incurs maintenance burden and complexity (also have to implement this in cloud) etc.

It feels like this new method is not fundamentally different from:

```
record = next(
    get_auto_materialize_asset_evaluations(asset_key=..., cursor=evaluation_id+1, limit=1),
    None,
)
```

so I would strongly prefer if we went that route rather than adding a new query. there is a potential argument that the new query is more performant, but I don't think this will be an issue in the short term and we can cross that bridge when we come to it.

OwenKephart (changes requested):

python_modules/dagster-graphql/dagster_graphql/implementation/fetch_auto_materialize_asset_evaluations.py
johannkm: Good point... @salazarm do you want to do this query instead? If you'd prefer we can still add the gql resolver

salazarm: fine with it exposed with the existing query

PR #14360, opened 2023-05-18
Don't prompt to use '-h' for 'dagster dev' help since its already in use

gibsondan:
Summary:
Override the default help names for this command specifically since -h means something else for `dagster dev`.

Test Plan
Run `dagster dev` locally, help message changed


petehunt (approved):
i approve!

PR #14363, opened 2023-05-18
[ui] Fix flashing timeline empty state

hellendag:
## Summary & Motivation

I accidentally introduced a measurement loop in #14339, where rendering the empty state leads to zero width because there is no longer a container ref to measure. I needed to have a measurement div always present, so I relocated the ref to a div that will always appear, even in the empty state.

I also removed a styled component that we don't really need.

## How I Tested These Changes

View run timeline, verify that the empty state doesn't have rendering flashes. Page forward and back, find a time range with some runs. Resize the page, verify that the timeline resizes.

### DIFF
--- a/js_modules/dagit/packages/core/src/runs/RunTimeline.tsx
+++ b/js_modules/dagit/packages/core/src/runs/RunTimeline.tsx
@@ -77,7 +77,7 @@ export const RunTimeline = (props: Props) => {
   const parentRef = React.useRef<HTMLDivElement | null>(null);
   const {
     viewport: {width},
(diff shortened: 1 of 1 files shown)


hellendag:
* master
  * **PR #14363** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14363" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14363?utm_source=stack-comment).

bengotow (approved):

Issue #14336, opened 2023-05-17
dagster-dbt source key prefix include refs


bmarcj:
### What's the issue or suggestion?

The [documentation](https://docs.dagster.io/integrations/dbt/reference#adding-a-prefix-to-asset-keys) says:

> Note: The key_prefix argument only applies to models. If you want to apply a prefix to the source keys that Dagster generates, pass in a source_key_prefix argument:

This seems to not just matter for source keys, but also for any refs that are not loaded as models (by include/exclude flags on the dbt loading). Either the documentation should change to mention this fact, or all models (not just those included in the loading) need to have the key appended. 

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

jamiedemaria:
cc @OwenKephart  or @rexledesma  mind taking a look?

OwenKephart:
Hi @bmarcj ! Thanks for pointing this out. The behavior works as you're expecting (refs will get the key prefix as well), it's the documentation that's wrong. Just created a PR to address this!

OwenKephart:
@bmarcj sorry about that -- you're totally correct (mistyped in the above answer, just edited)

PR #14364, opened 2023-05-18
[dagster-dbt] update docs

OwenKephart:
## Summary & Motivation

Addresses: https://github.com/dagster-io/dagster/issues/14336

### DIFF
--- a/docs/content/integrations/dbt/reference.mdx
+++ b/docs/content/integrations/dbt/reference.mdx
@@ -185,7 +185,7 @@ dbt_assets = load_assets_from_dbt_project(
 )
 ```
 
-**Note**: The `key_prefix` argument only applies to models. If you want to apply a prefix to the source keys that Dagster generates, pass in a `source_key_prefix` argument:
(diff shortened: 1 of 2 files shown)


rexledesma (approved):

docs/content/integrations/dbt/reference.mdx
```diff
@@ -185,7 +185,7 @@ dbt_assets = load_assets_from_dbt_project(
 )
 ```
 
-**Note**: The `key_prefix` argument only applies to models. If you want to apply a prefix to the source keys that Dagster generates, pass in a `source_key_prefix` argument:
+**Note**: The `key_prefix` argument only applies to assets loaded from this command. If you want to apply a prefix to the input keys that Dagster generates, pass in a `source_key_prefix` argument:
```
erinkcochran87, line 188: ❓ Is this actually a command? I typically think of command as something you run like on a CLI

OwenKephart: good point -- updated

PR #14365, opened 2023-05-19
Add optional threadpool for enqueueing/launching sensor ticks in parallel

gibsondan:
Summary:
Added another separate threadpool for enqueueing and/or launching runs simultaneously. Keeping the two threadpools separate ensures that they can't starve each other out and can be tuned separately.

Test Plan:
New BK test case
Manually: Run a sensor that creates 20 ticks locally, benchmark the difference in execution times - ticks go by much faster with threads enabled.

### DIFF
--- a/python_modules/dagster/dagster/_core/instance/config.py
+++ b/python_modules/dagster/dagster/_core/instance/config.py
@@ -251,6 +251,7 @@ def sensors_daemon_config() -> Field:
         {
             "use_threads": Field(Bool, is_required=False, default_value=False),
             "num_workers": Field(int, is_required=False),
+            "num_enqueue_workers": Field(int, is_required=False),
         },
         is_required=False,
     )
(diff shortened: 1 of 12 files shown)


python_modules/dagster/dagster/_core/instance/config.py
```diff
@@ -251,6 +251,7 @@ def sensors_daemon_config() -> Field:
         {
             "use_threads": Field(Bool, is_required=False, default_value=False),
             "num_workers": Field(int, is_required=False),
+            "num_enqueue_workers": Field(int, is_required=False),
```
alangenfeld, line 254: should have a description to discern the difference between these two workers, doubt we can capture via name 

nit: is there something better than enqueue? submit?

gibsondan:
I think once this lands a "how can I measure and improve Dagster run throughput" Github discussion would be useful

gibsondan:
- enqueue => submit
- added blurb in docs alongside the existing threading section

alangenfeld (approved):
🙏

PR #14377, opened 2023-05-19
Add 'dagster code-server start' support as an option to the Helm chart

gibsondan:
Summary:
- make 'codeServerArgs' an option alongside 'dagsterApiGrpcArgs' that spins up the new entrypoint.

### DIFF
--- a/helm/dagster/charts/dagster-user-deployments/values.yaml
+++ b/helm/dagster/charts/dagster-user-deployments/values.yaml
@@ -39,6 +39,10 @@ deployments:
     #   - "dagster_test.test_project.test_jobs.repo"
     #   - "-a"
     #   - "define_demo_execution_repo"
+    # The `dagsterApiGrpcArgs` key can also be replaced with `codeServerArgs` to use a new
+    # experimental `dagster code-server start` command instead of `dagster api grpc`, which takes
+    # identical arguments but can reload its definitions from within the Dagster UI without
+    # needing to restart the user code deployment pod.
     dagsterApiGrpcArgs:
       - "-f"
       - "/example_project/example_repo/repo.py"
--- a/helm/dagster/values.yaml
+++ b/helm/dagster/values.yaml
@@ -300,6 +300,11 @@ dagster-user-deployments:
       #   - "dagster_test.test_project.test_jobs.repo"
       #   - "-a"
       #   - "define_demo_execution_repo"
+      #
+      # The `dagsterApiGrpcArgs` key can also be replaced with `codeServerArgs` to use a new
+      # experimental `dagster code-server start` command instead of `dagster api grpc`, which takes
+      # identical arguments but can reload its definitions from within the Dagster UI without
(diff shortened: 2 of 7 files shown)


alangenfeld:
what makes you confident enough to flip the default immediately? I would assume we want to have observed the proxy server in production environments a bit first

helm/dagster/values.yaml
```diff
@@ -292,15 +292,13 @@ dagster-user-deployments:
...
+      # Ex: "dagster code-server start -m dagster_test.test_project.test_jobs.repo"
       # would translate to:
-      # dagsterApiGrpcArgs:
+      # codeServerArgs:
       #   - "-m"
       #   - "dagster_test.test_project.test_jobs.repo"
-      #   - "-a"
-      #   - "define_demo_execution_repo"
-      dagsterApiGrpcArgs:
+      codeServerArgs:
```
alangenfeld, line 301: [1]

helm/dagster/charts/dagster-user-deployments/values.yaml
```diff
@@ -31,15 +31,13 @@ deployments:
...
+    # Ex: "dagster code-server start -m dagster_test.test_project.test_jobs.repo"
     # would translate to:
-    # dagsterApiGrpcArgs:
+    # codeServerArgs:
     #   - "-m"
     #   - "dagster_test.test_project.test_jobs.repo"
-    #   - "-a"
-    #   - "define_demo_execution_repo"
-    dagsterApiGrpcArgs:
+    codeServerArgs:
```
alangenfeld, line 40: [1]

alangenfeld (commented):
[1] I *think* its worth communicating the difference between `codeServerArgs` (reloadable vis proxy) and `dagsterApiGrpcArgs` (direct server). I can imagine someones managing a deployment may choose to save memory if they don't care about reload-ability for example.

alangenfeld:
> I think having two ways to do things is confusing and the memory footprint should be pretty minimal (particularly since it's not loading any user code anymore).

I buy the argument against the added complexity. In my past explorations without optimizing for it we pull in a lot on import leading to a surprisingly high memory foot print before any user code is loaded (hundreds of MB).

I think some measurements or observations from running it in production would be convincing on what path to take on complete deprecation

gibsondan:
"flipping the default" is slightly stronger language than what's actually happening, it wouldn't affect anybody that already has dagsterApiGrpcArgs set - it's more like 'flip the recommendation for new users coming to the Helm chart for the first time'. I think you're still right though, i'll do a less invasive change that makes it an option without it being the default recommendation

gibsondan:
@alangenfeld i added a less aggro version that makes it an option but not the default - I was thinking i'd soft-launch it in the github issue with some instructions for how to turn it on before referencing it in the values.yaml at all to avoid confusion

alangenfeld (approved):
I think adding comments to values.yaml about it would be fine too but you call the ball

PR #14383, opened 2023-05-22
Update launcher.py

ArnoudItility:
## Summary & Motivation

Include 3 different spot instances strategies:
1. No spot instances (near real time deliveries)
2. Spot instance as capacity provider, but spot is not required (deliveries in the near future)
3. Spot instance as capacity provider and spot instances required (deliveries in the future, no time pressure at all)

## How I Tested These Changes
**Within this PR the run_retry does not work yet!**

**Todo:** 
Find a way to requeue the Dagster run when Dagster reserved tag ecs/spot_only is set to True. As discussed with Daniel on slack (https://app.slack.com/client/TCDGQDUKF/C014UDS8LAV/thread/C014UDS8LAV-1684242078.712329), this PR is made to show my changes to see if there is a solution for the run_retry function. 

**Note:** 
No documentation and unit tests are included

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
@@ -42,6 +42,12 @@ from .tasks import (
 )
 from .utils import get_task_logs, sanitize_family, task_definitions_match
 
+# Imports for retry run
+from dagster._core.workspace.context import IWorkspaceProcessContext
+from dagster._core.definitions.selector import PipelineSelector
+from dagster._core.execution.plan.resume_retry import ReexecutionStrategy
+from dagster._core.definitions.metadata import MetadataValue
+
 Tags = namedtuple("Tags", ["arn", "cluster", "cpu", "memory"])
 
 RUNNING_STATUSES = [
@@ -59,6 +65,9 @@ DEFAULT_WINDOWS_RESOURCES = {"cpu": "1024", "memory": "2048"}
 
 DEFAULT_LINUX_RESOURCES = {"cpu": "256", "memory": "512"}
 
+class SpotInstanceException(Exception):
+    def __init__(self, message='No spot instance available'):
+        super(SpotInstanceException, self).__init__(message)
 
 class EcsRunLauncher(RunLauncher[T_DagsterInstance], ConfigurableClass):
     """RunLauncher that starts a task in ECS for each Dagster job run."""
@@ -335,6 +344,83 @@ class EcsRunLauncher(RunLauncher[T_DagsterInstance], ConfigurableClass):
 
         return Tags(arn, cluster, cpu, memory)
 
+    def retry_run(
+        failed_run: DagsterRun,
+        workspace_context: IWorkspaceProcessContext,
+    ) -> None:
+        """Submit a retry as a re-execute from failure."""
+        instance = workspace_context.instance
+        workspace = workspace_context.create_request_context()
+        if not failed_run.external_pipeline_origin:
+            instance.report_engine_event(
+                "Run does not have an external pipeline origin, unable to retry the run.",
+                failed_run,
+            )
+            return
+
+        origin = failed_run.external_pipeline_origin.external_repository_origin
+        code_location = workspace.get_code_location(origin.code_location_origin.location_name)
+        repo_name = origin.repository_name
+
+        if not code_location.has_repository(repo_name):
+            instance.report_engine_event(
+                (
+                    f"Could not find repository {repo_name} in location {code_location.name}, unable to"
+                    " retry the run. It was likely renamed or deleted."
+                ),
+                failed_run,
+            )
+            return
+
+        external_repo = code_location.get_repository(repo_name)
+
+        if not external_repo.has_external_job(failed_run.pipeline_name):
+            instance.report_engine_event(
+                (
+                    f"Could not find job {failed_run.pipeline_name} in repository {repo_name}, unable"
+                    " to retry the run. It was likely renamed or deleted."
+                ),
+                failed_run,
+            )
+            return
+
+        external_pipeline = code_location.get_external_pipeline(
+            PipelineSelector(
+                location_name=origin.code_location_origin.location_name,
+                repository_name=repo_name,
+                pipeline_name=failed_run.pipeline_name,
+                solid_selection=failed_run.solid_selection,
+                asset_selection=None
+                if failed_run.asset_selection is None
+                else list(failed_run.asset_selection),
+            )
+        )
+
+        strategy = ReexecutionStrategy.ALL_STEPS
+
+        new_run = instance.create_reexecuted_run(
+            parent_run=failed_run,
+            code_location=code_location,
+            external_pipeline=external_pipeline,
+            strategy=strategy,
+            use_parent_run_tags=True,
+        )
+
+        instance.report_engine_event(
+            "Retrying the run",
+            failed_run,
+            engine_event_data=EngineEventData({"new run": MetadataValue.dagster_run(new_run.run_id)}),
+        )
+        instance.report_engine_event(
+            "Launched as an automatic retry",
+            new_run,
+            engine_event_data=EngineEventData(
+                {"failed run": MetadataValue.dagster_run(failed_run.run_id)}
+            ),
+        )
+
+        instance.submit_run(new_run.run_id, workspace)
+
     def launch_run(self, context: LaunchRunContext) -> None:
         """Launch a run in an ECS task."""
         run = context.dagster_run
@@ -398,20 +484,64 @@ class EcsRunLauncher(RunLauncher[T_DagsterInstance], ConfigurableClass):
         if "launchType" in run_task_kwargs and run_task_kwargs.get("capacityProviderStrategy"):
             del run_task_kwargs["launchType"]
 
-        # Run a task using the same network configuration as this processes's task.
-        response = self.ecs.run_task(**run_task_kwargs)
+        # FARGATE_SPOT safety feature 
+        # -> Set capacityProviderStrategy to FARGATE if no spot instances available and spot_only is not True
+        # -> Requeue job if no spot instances available and spot_only is True
+        if run_task_kwargs.get("capacityProviderStrategy") == [{'capacityProvider': 'FARGATE_SPOT'}]:
+            try:
+                response = self.ecs.run_task(**run_task_kwargs)
+                tasks = response["tasks"]
+                if not tasks:
+                    failures = response["failures"]
+                    exceptions = []
+                    for failure in failures:
+                        reason = failure.get("reason")
+                        exceptions.append(f"Failure reason: {reason}")
+                        if reason == "Capacity is unavailable at this time. Please try again later or in a different availability zone":
+                            raise SpotInstanceException()
+                    # Raise regular error if error not due to spot integration
+                    raise Exception(exceptions)
+            except SpotInstanceException:
+                spot_only = run.tags.get(
+                    "ecs/spot_only", container_context.run_resources.get("spot_only")
+                )
+                if spot_only == True:
+                    # Re-enqueue job
+                    try:
+                        # TODO: verify if works!
+                        self.retry_run(run, IWorkspaceProcessContext)
+                    except Exception as e:
+                        raise e
 
-        tasks = response["tasks"]
+                else:
+                    # Set capacityProviderStrategy to FARGATE
+                    run_task_kwargs["capacityProviderStrategy"] = [{'capacityProvider': 'FARGATE'}]
+                    response = self.ecs.run_task(**run_task_kwargs)
+                    tasks = response["tasks"]
 
-        if not tasks:
-            failures = response["failures"]
-            exceptions = []
-            for failure in failures:
-                arn = failure.get("arn")
-                reason = failure.get("reason")
-                detail = failure.get("detail")
-                exceptions.append(Exception(f"Task {arn} failed because {reason}: {detail}"))
-            raise Exception(exceptions)
+                    if not tasks:
+                        failures = response["failures"]
+                        exceptions = []
+                        for failure in failures:
+                            arn = failure.get("arn")
+                            reason = failure.get("reason")
+                            detail = failure.get("detail")
+                            exceptions.append(Exception(f"Task {arn} failed because {reason}: {detail}"))
+                        raise Exception(exceptions)                    
+        else:
+            # if capacityProviderStrategy is not FARGATE_SPOT, business as usual
+            response = self.ecs.run_task(**run_task_kwargs)
+            tasks = response["tasks"]
+
+            if not tasks:
+                failures = response["failures"]
+                exceptions = []
+                for failure in failures:
+                    arn = failure.get("arn")
+                    reason = failure.get("reason")
+                    detail = failure.get("detail")
+                    exceptions.append(Exception(f"Task {arn} failed because {reason}: {detail}"))
+                raise Exception(exceptions)
 
         arn = tasks[0]["taskArn"]
         cluster_arn = tasks[0]["clusterArn"]


ArnoudItility:
Please beware of the following requirements:

1. We want to requeue runs that have the Dagster reserved tag "ecs/spot_only" set to True, rather than having a for/while loop with backoff, since new runs with higher priority should not be stuck due to this retry mechanism. 
2. There is no limit on number of retries in case no spot instances are available.
3. Retry strategy should be "ALL_STEPS". No persistent storage should be required in order to retry the runs when no spot instances are available.
4. Retrying all Dagster runs that fail is not a valid solution (current retry policy already integrated in Dagster), as we are dealing with a high volume of batched Dagster runs using DynamicOut, of which it frequently occurs that at least one out of the entire batch fails. Retrying the failures would therefore mean that the vast majority would be re-executed for no reason!

gibsondan (commented):
Re-submitting to a non-spot instance if a spot instance is unavailable seems like a reasonable change to me. I'm a little confused why it would need to call retry_run - can it try calling run_task for the same run again instead, similar to how it re-submits the same run with FARGATE if ecs/spot_only is False?

ArnoudItility:
See requirement 1 in the comment above! 

Correct me if I am wrong, but it seems that if we add a retry with backoff on the spot_only runs, the Daemon remains occupied with this run and does not start other runs with higher priority? 

Also, what happens if you have a max run concurrency of 100 and 100 jobs are retrying until spot instances become available, whereas on the side you are also running jobs that do not require spot only that are activated slightly later than the spot only runs. Then it seems to me that these runs are not activated due to the fact that the Daemon is "stuck" at the spot only runs!

ArnoudItility:
I have created a possible solution in case it is not feasible to requeue the job. I replaced the "spot_only" tag with a "spot_retries" tag, in which you can denote how many retries it can perform before eventually switching to an on demand instance.


        # FARGATE_SPOT safety feature 
        if run_task_kwargs.get("capacityProviderStrategy") == [{'capacityProvider': 'FARGATE_SPOT'}]:
            # if capacityProviderStrategy is FARGATE_SPOT 
            #   try to execute 'spot_retries' times to use spot instance 
            #   else switch to on demand instance

            spot_retries = run.tags.get("ecs/spot_retries", container_context.run_resources.get("spot_retries"))
            if spot_retries == None:
                spot_retries = 0

            for i in range(spot_retries + 1):
                if i < (spot_retries):
                    try:
                        tasks = run_task()
                        break
                    except SpotInstanceException:
                        time.sleep(5)
                        continue
                else:
                    # Set capacityProviderStrategy to FARGATE
                    run_task_kwargs["capacityProviderStrategy"] = [{'capacityProvider': 'FARGATE'}]
                    tasks = run_task()
        else:
            # if capacityProviderStrategy is not FARGATE_SPOT
            #   normal flow
            tasks = run_task()

        arn = tasks[0]["taskArn"]
        cluster_arn = tasks[0]["clusterArn"]
        self._set_run_tags(run.run_id, cluster=cluster_arn, task_arn=arn)
        self.report_launch_events(run, arn, cluster_arn)

gibsondan:
Overall I think this is cool but maybe a little too tailored to your specific use case to merge into the main EcsRunLauncher as is - the piece of it that I can most imagine merging in is the 'fall back to non-spot instance if a spot instance is not available' piece, presumably with a flag to toggle it on and off. I'd have to think a bit about the best place to express that option though since right now we expose the ability to do spot instances by letting you customize the run_task_kwargs, without really understanding the semantics of those args

I know I was the one who advocated for a PR - are you planning to keep this yourself as a forked custom run launcher going forward? Or are you hoping to merge some of this into the EcsRunLauncher in master?

ArnoudItility:
Sounds fair. 
I was hoping to get at least the following integrated in the EcsRunLauncher in master:

- Spot instance with and option to add retries following a Dagster reserved tag.
- Automatic fall back to on demand instance when no spot instances available (after given number of retries)

I will create a new PR which include these changes after these have been thoroughly tested from our side. 
Hopefully this is in line with the Dagster team!

Cheers

ArnoudItility:
Can you guys let me know if I need to make some additional changes in order to make sure that the changes will get included in the master branch? I am willing to make it a bit more generic, like including a toggle tag to (de-)activate the automated fall back feature. 

Thank you in advance!

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -396,22 +401,59 @@ def launch_run(self, context: LaunchRunContext) -> None:
...
+            tasks = response["tasks"]
+            if not tasks:
+                failures = response["failures"]
+                exceptions = []
+                for failure in failures:
+                    arn = failure.get("arn")
+                    reason = failure.get("reason")
+                    detail = failure.get("detail")
+                    exceptions.append(Exception(f"Task {arn} failed because {reason}: {detail}"))
+                    if reason == "Capacity is unavailable at this time. Please try again later or in a different availability zone":
```
gibsondan, line 418: this error can be hit even if you are not using fargate spot instances, and you might want to retry. Can we configure the number of retries to use if this happens separately from the capacity provider strategy?

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -396,22 +401,59 @@ def launch_run(self, context: LaunchRunContext) -> None:
...
+            else:
+                spot_retries = 0
+
+            for i in range(spot_retries + 2):  # if spot_retries = 0, first try spot [i = 0] and then use on demand [i = 1]
+                if i < (spot_retries + 1):  # if spot_retries = 0, first evaluation is True [i=0 < 1] and second evaluation is False [i=1 !< 1]
+                    try:
+                        tasks = run_task()
+                        break
+                    except SpotInstanceException:
+                        asyncio.sleep(5)
```
gibsondan, line 447: the interval here should be something configurable as well - you probably would want a configuration dictionary just for capacity retries with the number of retries and the interval between retries

gibsondan (changes requested):
ok, I did a more detailed pass on this. Some thoughts:

- i think we would want to decouple the "retry if there's a capacity failure" behavior (which can happen in both FARGATE and FARGATE_SPOT) into a separate setting.

- We would definitely want the "fall back to a different capacity provider on failure" thing to be opt-in vs. the default (maybe worth splitting that out into a separate PR once the first thing is done?). What do you think about making that a fallbackCapacityProviderStrategy tag so that they can set the whole strategy dictionary to use if we exhaust all the retries? 

We'd also want to add some unit testing in dagster_aws_tests that verify the new behavior.

ArnoudItility:
Thanks for the inputs! 

We will start working on these changes timely, separating them into two different PR's (will post them in this chat to keep track of the overall status)

ArnoudItility:
Hi Daniel,

Would the following pseudo-code be in line with the expectations?

### PATCHES
# 1. capacityFailureStrategy
    # capacity_failure_strategy = run.tags.get("ecs/capacityFailureStrategy")
        # where capacityFailureStrategy = {'maxRetries': '<INSERT_MAX_NUMBER_OF_RETRIES>', 'retryInterval': '<INSERT_RETRY_INTERVAL_IN_SECONDS>'}
    # loop maxRetries times if capacityFailureException is raised with interval set to retryInterval, raise Exception afterwards
        # default maxRetries = 0

# 2. fallbackCapacityProviderStrategy
    # If maxRetries is reached and run.tags.get("ecs/fallbackCapacityProviderStrategy") is set:
        # run_task_kwargs["capacityProviderStrategy"] = run.tags.get("ecs/fallbackCapacityProviderStrategy")
            # where fallbackCapacityProviderStrategy = [{'capacityProvider': '<INSERT_FALLBACK_CAPACITY_PROVIDER>'}]

gibsondan:
@ArnoudItility what do you think about starting with the 'retry on capacity failure' settings as instance-level settings rather than tags though? That way people don't have to set it for each individual job in code, and it seems less likely to me that users would want to vary the number of times to retry on capacity failures per-job

ArnoudItility:
Sounds reasonable. Where can I find an example to add this on instance level? Or is there no documentation for this?

gibsondan:
Take a look at "run_task_kwargs" here: https://github.com/dagster-io/dagster/blob/master/python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py#LL300C14-L300C29

If you add it to that map in the config_type method it will be automatically passed into the constructor of the class as well

ArnoudItility:
Hello Daniel, 

would you also like to see the 'fallback capacity providerStrategy' on instance level rather than job level (or both)?

gibsondan:
Starting with the instance level makes sense to me, yeah.

ArnoudItility:
Hi Daniel, I have created both strategies on instance level now and also added tag level for the capacity failure strategy. I have tested it locally and on AWS and behavior is conform expectations

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -335,6 +385,62 @@ def _get_run_tags(self, run_id):
...
+                        for failure in failures:
+                            arn = failure.get("arn")
+                            reason = failure.get("reason")
+                            detail = failure.get("detail")
+                            exceptions.append(Exception(f"Task {arn} failed because {reason}: {detail}"))
+                        raise Exception(exceptions)
+
+                return tasks
+            
+            except CapacityFailureException:
```
gibsondan, line 421: I don't think we should use an exception for expected business logic like this. Can we increment a counter directly instead?

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -306,6 +316,46 @@ def config_type(cls):
                     " always be set by the run launcher."
                 ),
             ),
+            "capacity_failure_strategy": Field(
+                Permissive(
+                    {
+                        "maxRetries": Field(
```
gibsondan, line 322: maxRetries => max_retries

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -306,6 +316,46 @@ def config_type(cls):
...
+                            int,
+                            is_required=False,
+                            default_value=5,
+                            description="Number of retries in case a capacity failure takes place"
+                        ),
+                        "retryInterval": Field(
```
gibsondan, line 328: retryInterval => retry_interval

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -306,6 +316,46 @@ def config_type(cls):
...
+                                str,
+                                is_required=True,
+                                description="Fallback capacity provider"
+                            )
+                        }
+                    ),
+                ),
+                is_required=False,
+                description=(
+                    "An array of the fallback capacity provider strategy (List[Dict[str, Any]])"
```
gibsondan, line 356: can we explain what it is and what it does vs. repeating its name

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -335,6 +385,62 @@ def _get_run_tags(self, run_id):
 
         return Tags(arn, cluster, cpu, memory)
 
+    def _run_task(self, run_task_kwargs, capacity_failure_strategy: dict, fallback_capacity_provider_strategy: list):
```
gibsondan, line 388: `dict` and `list` don't work in the versions of Python that dagster needs to support - I think you want Dict and List

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -335,6 +385,62 @@ def _get_run_tags(self, run_id):
...
+                    if any(failure.get("reason") == "Capacity is unavailable at this time. Please try again later or in a different availability zone" for failure in failures):
+                        raise CapacityFailureException()
+                    else:
+                        exceptions = []
+                        for failure in failures:
+                            arn = failure.get("arn")
+                            reason = failure.get("reason")
+                            detail = failure.get("detail")
+                            exceptions.append(Exception(f"Task {arn} failed because {reason}: {detail}"))
+                        raise Exception(exceptions)
```
gibsondan, line 417: this logic has changed if you rebase - it no longer raises an Exception with a list of exceptions

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -335,6 +385,62 @@ def _get_run_tags(self, run_id):
...
+    
+    def _get_capacity_failure_strategy(self, run):
+        capacity_failure_strategy = {}
+
+        capacity_failure_strategy_from_instance = self.capacity_failure_strategy
+        capacity_failure_strategy_from_tag = run.tags.get("ecs/capacity_failure_strategy")
+
+        # tag has highest priority
+        if capacity_failure_strategy_from_tag:
+            capacity_failure_strategy = ast.literal_eval(capacity_failure_strategy_from_tag)
```
gibsondan, line 438: this is a json-encoded string right? i think json.loads is the standard way the dagster codebase parses those

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -335,6 +385,62 @@ def _get_run_tags(self, run_id):
...
+        fallback_iteration = 1 if fallback_capacity_provider_strategy else 0
+        retry_interval = int(capacity_failure_strategy["retryInterval"]) if capacity_failure_strategy else 0
+
+        # fire off runs
+        total_iterations = default_iteration + retry_iterations + fallback_iteration
+        for i in range(total_iterations):
+
+            if fallback_capacity_provider_strategy and i == (total_iterations - 1):
+                run_task_kwargs["capacityProviderStrategy"] = fallback_capacity_provider_strategy
+
```
gibsondan, line 402: this is clever, but i think we should consider doing the fallback iteration separately to improve readability here

gibsondan (changes requested):
OK i think this is getting closer. You'll want to:
- rebase
- add some automated unit testing that exercises the new functionality with the mocked out ECS api, see the other run launcher tests in test_launching.py for examples

ArnoudItility:
Hi Daniel, 

I got a question related to the unit testing of the adjusted run launcher functionalities:

In all existing unit tests you simply launch a run where you can verify whether the behavior is in line with expectations (i.e. for the spot unit test you check if the capacityProvider equals FARGATE_SPOT). However, the newly added logic is within the "launch_run" function. To be more precise, the only way we can unit test the new functionalities is to perform the launch_run and within this function it should throw some capacity failures. But since you do not raise an error until the max_retries are exceeded, how do you test something like this? Same goes for the fallback functionality, where you first need to have a failure and then you can check if the new capacityProvider equals the fallbackCapacityProvider.

Maybe we simply need to test the values passed in the test_instance.py or the tags?

gibsondan:
Maybe take a look at this test for inspiration: https://github.com/dagster-io/dagster/blob/master/python_modules/libraries/dagster-aws/dagster_aws_tests/ecs_tests/launcher_tests/test_failures.py#L4

ArnoudItility:
Hi Daniel, 

I have created two tests in order to test both functionalities. I have not yet tested them locally, but I was curious whether this is in line with expectations. I do see that I used "master" branch to create my new unit tests, whereas I uploaded it to 1.3.2 (hence the many differences). This does make rebasing easier for integrating them into master...

ArnoudItility:
Hi Daniel, 

I am running into issue when trying to do the pytest locally as I am not using a Linux OS and it is taking me quite some time to reinstall everything accordingly. Would it be possible to handover the checking of the unit tests to you guys, I think there is not a lot of development work left there. 

If so, I will rebase my changes into master (launcher.py). The test_launching.py is already rebased according to master.

Sorry for the inconvenience...

gibsondan (commented):
@ArnoudItility i'm going to be out on leave for the next several weeks but you're in good hands with @jmsanders and/or others on the Dagster team on pushing this forward.

The test you have there looks broadly reasonable to me - the main thing I think you want to do next before getting this for review is to rebase this on top of master rather than the 1.3.2 release branch, as its quite difficult to separate out your change from the other changes that have happened in the repo since that release.

ArnoudItility:
Thank you Daniel!

We have had some busy weeks so there was not a lot of time for rebasing. @jmsanders we will pick this up this week and hopefully have a rebased version to current master by the end of the week!

Kind regards,
Arnoud

ArnoudItility:
Hello @jmsanders, 

please find the rebased files below:
https://github.com/dagster-io/dagster/pull/15088

ArnoudItility:
Hi @gibsondan, 

I just got back from holidays as well, however, I see the latest rebase to master has not been picked up unfortunately :(
Before I will rebase it again to master, can I get a confirmation that it will be picked up timely?

Kind regards,
Arnoud

cmpadden:

PR #14388, opened 2023-05-22
[Runs Page] More runs tweaks

salazarm:
## Summary & Motivation

- Change “Created by” to “Launched by”
    - Change the table column header and the filter key
- In OSS, if no user is specified, lets change the tag label from “Launchpad” to “Manually launched”
    - I think it would be cool to add a tooltip on the “Manually launched” tag that says “No ‘user’ tag specified” … that will give users an indication of how to populate this tag in the future.
- Remove the “Snapshot” tag from the Target column since everyone agreed it’s not that useful.
- For backfill runs:
    - Let’s show the user that triggered the backfill in the “Launched by column”
    - Lets move the backfill tag to the “Target” column
    - It looks like for single run backfills we append two extra tags. asset_partition_range_start and asset_partition_range_end. I think we should just show those two tags in the “Target” column so it’s easy to see the range of targeted partitions. 

## How I Tested These Changes
Screen Shot 2023-05-22 at 10 41 22 AM
Screen Shot 2023-05-22 at 10 40 05 AM
Screen Shot 2023-05-22 at 10 39 16 AM

### DIFF
--- a/js_modules/dagit/packages/core/src/runs/RunsFilterInputNew.tsx
+++ b/js_modules/dagit/packages/core/src/runs/RunsFilterInputNew.tsx
@@ -399,7 +399,7 @@ export const useRunsFilterInput = ({tokens, onChange, enabledFilters}: RunsFilte
     filters: [
       !enabledFilters || enabledFilters?.includes('status') ? statusFilter : null,
       useStaticSetFilter({
-        name: 'Created By',
+        name: 'Launched By',
         icon: 'add_circle',
         allValues: createdByValues,
         renderLabel: ({value}) => {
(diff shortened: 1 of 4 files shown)


js_modules/dagit/packages/core/src/runs/RunsFilterInputNew.tsx
```diff
@@ -399,7 +399,7 @@ export const useRunsFilterInput = ({tokens, onChange, enabledFilters}: RunsFilte
     filters: [
       !enabledFilters || enabledFilters?.includes('status') ? statusFilter : null,
       useStaticSetFilter({
-        name: 'Created By',
+        name: 'Launched By',
```
hellendag, line 402: `Launched by`

hellendag (approved):

PR #14389, opened 2023-05-22
Keep going if you get a 409 Conflict after retrying creating a k8s job

gibsondan:
Summary:
Retrying and getting a 409 after getting a 500 after creating a k8s job indicates that the job was actually created after all - so allow it through.

Test Plan: BK

### DIFF
--- a/python_modules/libraries/dagster-k8s/dagster_k8s/client.py
+++ b/python_modules/libraries/dagster-k8s/dagster_k8s/client.py
@@ -133,6 +133,61 @@ def k8s_api_retry(
     check.failed("Unreachable.")
 
 
+def k8s_api_retry_creation_mutation(
+    fn: Callable[..., None],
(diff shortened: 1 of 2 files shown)


python_modules/libraries/dagster-k8s/dagster_k8s/client.py
```diff
@@ -133,6 +133,55 @@ def k8s_api_retry(
...
+
+        try:
+            fn()
+            return
+        except kubernetes.client.rest.ApiException as e:
+            retry_count = retry_count + 1
+            # Only catch whitelisted ApiExceptions
+            status = e.status
+
+            if retry_count > 1 and status == 409:
```
alangenfeld, line 159: doc block / extra comment on this 409 line to make it stand out what the difference is between this and the other

alangenfeld (approved):

Issue #14381, opened 2023-05-21
dagster-dbt assets can't be materialized via jobs when the whole lineage of asset is not specified


bvujanic:
### Dagster version

dagster, version 1.3.5

### What's the issue?

After upgrading dagster to 1.3.x and dagster-dbt to 0.19.x the ability to materialize assets via job trough UI that don't have the whole lineage is lost.

The  example that demonstrates the issue uses the official dbt example.
https://github.com/dagster-io/dagster/tree/master/examples/assets_dbt_python

I've just added two jobs as definitions

```code
materialize_one_asset = define_asset_job("example_job_materialize_one_asset", selection=AssetSelection.keys(AssetKey(["duckdb", "dbt_schema", "company_stats"])))
materialize_asset_upstream = define_asset_job("example_job_materialize_one_asset_with_upstream", selection=AssetSelection.keys(AssetKey(["duckdb", "dbt_schema", "company_stats"])).upstream())
```
where the `materialize_one_asset` exhibits the problem and `materialize_asset_upstream` shows the case when dbt job behaves properly.

See:

![Screenshot 2023-05-21 at 20 59 00](https://github.com/dagster-io/dagster/assets/31449262/2cdb8148-b608-4526-9079-24140f87c088)

The image shows that the job itself doesn't have the `Materialize` capability as if it lacks config

### What did you expect to happen?

Instead of the above the expected behavior is that the job has materialize button and can be ran via launchpad.

![Screenshot 2023-05-21 at 21 00 01](https://github.com/dagster-io/dagster/assets/31449262/10a5449f-f18e-423a-a4ac-04da97ebc8b9)

To achieve that 

### How to reproduce?

Create a job that relies on a dbt asset without specifying all of the upstream assets.

### Deployment type

Dagster Helm chart

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

OwenKephart:
Hi! Thanks for the report (and sorry you ran into this!) I tracked down the cause to a PR from a few weeks back. Long story short, a new op name was generated in some cases when this wasn't necessary, which confused the UI and made it unable to associate the new op with any assets.

I was able to confirm that this was fixed after the change. This fix will go out in this week's release.

PR #14390, opened 2023-05-22
Fix subsetting bug

OwenKephart:
## Summary & Motivation

We have some special handling in place for the rare case that you have a subsettable multi asset which has some nodes upstream of another asset (`A`) and other nodes downstream of `A`.

In these cases, we need to separate out the asset into two separate op invocations in order for it to be possible to execute (as otherwise the op for the multi-asset would need to execute both before and after the op for `A`).

In some cases, this requires just invoking the original op two separate times, but in other cases, we actually need to generate a new op. See: https://github.com/dagster-io/dagster/pull/13830 for more context.

However, we were overly zealous in the original PR, generating new ops in situations where it was not actually necessary (as the broader job would not actually need to plug in any output into the new input we created).

This PR fixes this, meaning we will only create the new input in cases where it will actually be used. 

Further work should be done to investigate if there are other similar issues that are caused by this behavior, but this PR should quarantine the impact to the rare situation described in the above PR.

## How I Tested These Changes

Modified test failed before this change, passes now.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/assets.py
@@ -926,12 +926,18 @@ class AssetsDefinition(ResourceAddable, IHasInternalInit):
         )
 
     def _subset_op_backed_asset(
-        self, asset_subselection: AbstractSet[AssetKey]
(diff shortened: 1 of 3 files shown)


clairelin135 (approved):

OwenKephart:
Resolves: https://github.com/dagster-io/dagster/issues/14381

PR #14392, opened 2023-05-22
Prefer run config over config mapping when reconstructing Jobs

### LINKED ISSUE #14371: Hardcoded run config does not work with default in_process_executor
### Dagster version

1.3.4

### What's the issue?

When using **in_process_executor** as default in the **Definitions** for a code-location, hardcoded job run-configs fail with:

> dagster._core.errors.DagsterInvalidConfigError: Error in config for job XXX
    Error 1: Received unexpected config entry "multiprocess" at path root:execution:config. Expected: "{ marker_to_close?: String retries?: { disabled?: { } enabled?: { } } }"

It even seems to ignore any execution run-config and always pushes "multiprocess" downstream to the executor which then fails config-validation in the **in_process_executor**.

Hardcoded run-config should not assume **multiprocess_executor** or **multi_or_in_process_executor** as default and not unintentionally specify "multiprocess" config to the default executor under the hood.

### How to reproduce?

I used the [Hardcoded configuration](https://docs.dagster.io/concepts/ops-jobs-graphs/jobs#hardcoded-configuration) sample from the docs and just added a **Definitions** with **in_process_executor**. Run it with dagster dev -f sample.py

```python
from dagster import (
    Config,
    Definitions,
    RunConfig,
    in_process_executor,
    job,
    op
)


class DoSomethingConfig(Config):
    config_param: str


@op
def do_something(context, config: DoSomethingConfig):
    context.log.info("config_param: " + config.config_param)


default_config = RunConfig(
    ops={"do_something": DoSomethingConfig(config_param="stuff")}
)


@ job(config=default_config)
def do_it_all_with_default_config():
    do_something()


defs = Definitions(jobs=[do_it_all_with_default_config], executor=in_process_executor)
```

### Deployment type

Local

### Deployment details

_No response_

### Additional information

_No response_

benpankow:
## Summary

When a job is supplied a run config dict, we use `_config_mapping_with_default_value` to create a corresponding config mapping which produces this default config. This config mapping also bundles in information related to e.g. the executor.
The job object then has `_run_config` and `_config_mapping` set.

When copying a job (to attach a default executor, for example), we forward the config by passing `job._config_mapping or job._run_config`.

In the case where the original job object was passed a run config dict, it has `_run_config` and `_config_mapping` set. The or precedence means the copied job has `_config_mapping`, which does not take into account any changes in the copy (such as a new executor).

This PR reverses the `or` precedence so that `_run_config` is passed and a new config mapping is generated in this scenario.

Should resolve #14371

## Test Plan

Reproduced #14371 locally and ensured this fixed. Added a small test case (previously failing).

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/job_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/job_definition.py
@@ -209,6 +209,7 @@ class JobDefinition(IHasInternalInit):
         self._partitioned_config = None
         self._run_config = None
         self._run_config_schema = None
+        self._original_config_argument = config
 
         if partitions_def:
             self._partitioned_config = PartitionedConfig.from_flexible_config(
@@ -918,13 +919,7 @@ class JobDefinition(IHasInternalInit):
             resource_defs=dict(self.resource_defs),
             executor_def=self._executor_def,
             logger_defs=self._loggers,
-            # One of the following is true:
-            #   _config_mapping is set, and _partitioned_config and _run_config are None
-            #   _partitioned_config is set, and _config_mapping and _run_config are None
-            #   _run_config is set and _config_mapping is built from it using _config_mapping_with_default_value (and _partitioned_config is None)
-            # In the last case, we want to copy the _run_config, not the _config_mapping, since a new _config_mapping will be built
-            # from the _run_config in the constructor
-            config=self._run_config or self._config_mapping or self._partitioned_config,
+            config=self._original_config_argument,
             name=self._name,
             description=self.description,
             tags=self.tags,
--- a/python_modules/dagster/dagster/_core/definitions/job_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/job_definition.py
@@ -918,7 +918,7 @@ class JobDefinition(IHasInternalInit):
             resource_defs=dict(self.resource_defs),
             executor_def=self._executor_def,
             logger_defs=self._loggers,
-            config=self._config_mapping or self._partitioned_config or self._run_config,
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -918,7 +918,13 @@ def _copy(self, **kwargs: Any) -> "JobDefinition":
...
             executor_def=self._executor_def,
             logger_defs=self._loggers,
-            config=self._config_mapping or self._partitioned_config or self._run_config,
+            # One of the following is true:
+            #   _config_mapping is set, and _partitioned_config and _run_config are None
+            #   _partitioned_config is set, and _config_mapping and _run_config are None
+            #   _run_config is set and _config_mapping is built from it using _config_mapping_with_default_value (and _partitioned_config is None)
+            # In the last case, we want to copy the _run_config, not the _config_mapping, since a new _config_mapping will be built
+            # from the _run_config in the constructor
+            config=self._run_config or self._config_mapping or self._partitioned_config,
```
schrockn, line 927: So this seems very hard to get right.

Why don't we instead just add another variable `self._config_to_copy` that we set alongside the rest of this on line 218. Then there is one place where we have to set the logic, and it will be self-evident from the codepath what the cases are rather than encoding it into a comment like this which is likely go out of date at some point.

schrockn (changes requested):

python_modules/dagster/dagster/_core/definitions/job_definition.py
benpankow: seems reasonable - updated

PR #14394, opened 2023-05-22
[1.4] feat(dbt)!: remove dbt-rpc integration

rexledesma:
## Summary & Motivation
`dbt-rpc` is in maintenance mode, in preparation to be replaced by the more modern [`dbt-server`](https://github.com/dbt-labs/dbt-server). 

Here, we follow dbt's lead and deprecate our support for this portion of the dbt integration.

## How I Tested These Changes
pytest


rexledesma:
* master
  * **PR #14394** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14394" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14394?utm_source=stack-comment).

OwenKephart (commented):
I'm totally in support of removing this, and I don't think the impact of removing it would be too large, but I think it would be a nice gesture to explicitly run this through a deprecation cycle (targeting true removal for the 1.4 release).

yuhan:
agree with owen on deprecation cycle + removal in 1.4 - can we retitle this pr to include something like "merge until 1.4"

sryza (approved):
love to see this kind of diff

image

PR #14396, opened 2023-05-22
Add env vars in all our CLI entrypoints if there's a .env file in the local working directory

gibsondan:
Summary:
This PR intends to address a point of user confusion that comes up frequently in support - we tell users in our docs to put env vars in a .env file and they'll be automatically injected into the environment, but it doesn't apply to our own dagster env vars, just the ones that are used in their code. So users still hit the same friction / confusion that the .env file is intended to address.

So instead of using a pluggable component and defaulting it to a SecretLoader that pulls from a local .env file but is only called in certain CLIs, make it called in every CLI / Dagster entrypoint.

I thought about putting this directly in DagsterInstance.get() but it seemed bad to add a permanent side effect there that wasn't there before, and there are many places where we create a DagsterInstance outside of a process entrypoint that wouldn't need to inject the env vars.

Will add testing after any initial thoughts on this direction.

### DIFF
--- a/python_modules/dagit/dagit/app.py
+++ b/python_modules/dagit/dagit/app.py
@@ -1,14 +1,11 @@
 from dagster import (
-    DagsterInstance,
     _check as check,
 )
-from dagster._cli.workspace.cli_target import get_workspace_process_context_from_kwargs
 from dagster._core.execution.compute_logs import warn_if_compute_logs_disabled
 from dagster._core.telemetry import log_workspace_stats
 from dagster._core.workspace.context import IWorkspaceProcessContext
 from starlette.applications import Starlette
 
-from .version import __version__
 from .webserver import DagitWebserver
 
 
@@ -38,21 +35,3 @@ def create_app_from_workspace_process_context(
         workspace_process_context,
         path_prefix,
     ).create_asgi_app(**kwargs)
-
-
-def default_app(debug=False):
-    instance = DagsterInstance.get()
-    process_context = get_workspace_process_context_from_kwargs(
-        instance=instance,
-        version=__version__,
-        read_only=False,
-        kwargs={},
-    )
-
-    return DagitWebserver(
-        process_context,
-    ).create_asgi_app(debug=debug)
-
-
-def debug_app():
-    return default_app(debug=True)
--- a/python_modules/dagster/dagster/_cli/utils.py
+++ b/python_modules/dagster/dagster/_cli/utils.py
@@ -26,8 +26,38 @@ def _load_env_var_dict(logger: logging.Logger) -> None:
 
 
 @contextmanager
-def get_possibly_ephemeral_instance_for_cli(
-    cli_command: Optional[str] = "this command",
+def _get_temporary_instance(cli_command: str, logger: logging.Logger) -> Iterator[DagsterInstance]:
+    # make the temp dir in the cwd since default temp dir roots
+    # have issues with FS notify-based event log watching
+    with tempfile.TemporaryDirectory(dir=os.getcwd()) as tempdir:
+        logger.info(
+            f"Using temporary directory {tempdir} for storage. This will be removed when"
+            f" {cli_command} exits."
+        )
+        logger.info(
+            "To persist information across sessions, set the environment variable DAGSTER_HOME"
+            " to a directory to use."
+        )
+
+        with DagsterInstance.from_ref(
+            InstanceRef.from_dir(tempdir, config_dir=os.getcwd())
+        ) as instance:
+            yield instance
+
+
+@contextmanager
+def get_temporary_instance_for_cli(
+    cli_command: str = "this command",
+    logger: logging.Logger = logging.getLogger("dagster"),
+) -> Iterator[DagsterInstance]:
+    _load_env_var_dict(logger)
+    with _get_temporary_instance(cli_command, logger) as instance:
+        yield instance
+
+
+@contextmanager
+def get_possibly_temporary_instance_for_cli(
+    cli_command: str = "this command",
     instance_ref: Optional[InstanceRef] = None,
     logger: logging.Logger = logging.getLogger("dagster"),
 ) -> Iterator[DagsterInstance]:
@@ -50,22 +80,8 @@ def get_possibly_ephemeral_instance_for_cli(
         with DagsterInstance.get() as instance:
             yield instance
     else:
-        # make the temp dir in the cwd since default temp dir roots
-        # have issues with FS notify-based event log watching
-        with tempfile.TemporaryDirectory(dir=os.getcwd()) as tempdir:
-            logger.info(
-                f"Using temporary directory {tempdir} for storage. This will be removed when"
-                f" {cli_command} exits."
-            )
-            logger.info(
-                "To persist information across sessions, set the environment variable DAGSTER_HOME"
-                " to a directory to use."
-            )
-
-            with DagsterInstance.from_ref(
-                InstanceRef.from_dir(tempdir, config_dir=os.getcwd())
-            ) as instance:
-                yield instance
+        with _get_temporary_instance(cli_command, logger) as instance:
+            yield instance
 
 
 @contextmanager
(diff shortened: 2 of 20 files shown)


schrockn (dismissed):
q mgmt

gibsondan:
I thought only methods with @public were considered part of our public API now, but I'm happy to move that method elsewhere

schrockn:
I mean even public in an internal sense. I don't think it should exist, pretty much.

gibsondan:
@schrockn addressed the biggest piece of your feedback which I think has made this a lot cleaner:
- moved the method out of DagsterInstance into a separate get_instance_for_cli method (repurposed a somewhat similar get_instance_for_service method that was already in use for some entrypoints)

The list of dagster env vars is DAGSTER_HOME plus anything that is referenced in your dagster.yaml with an "env" key (people reasonably thought they could put, say, the postgres password in the .env file and then do `env:POSTGRES_PASSWORD` in their dagster.yaml

python_modules/dagit/dagit/app.py
```diff
@@ -38,21 +35,3 @@ def create_app_from_workspace_process_context(
...
-        kwargs={},
-    )
-
-    return DagitWebserver(
-        process_context,
-    ).create_asgi_app(debug=debug)
-
-
-def debug_app():
-    return default_app(debug=True)
```
gibsondan, line 58: unused

python_modules/dagster/dagster_tests/core_tests/instance_tests/test_instance.py
```diff
@@ -569,6 +572,44 @@ def test_dagster_home_not_dir():
             DagsterInstance.get()
 
 
+def test_dagster_env_vars_from_dotenv_file():
```
gibsondan, line 575: this test is an example of both kinds of Dagster env vars: DAGSTER_HOME itself, plus SQLITE_STORAGE_BASE_DIR , referenced in the dagster.yaml file

python_modules/dagster/dagster/_cli/utils.py
```diff
@@ -1,35 +1,68 @@
...
 
 
 @contextmanager
-def get_instance_for_service(
-    service_name: str,
+def get_instance_for_cli(
+    cli_command: Optional[str] = "this command",
     instance_ref: Optional[InstanceRef] = None,
-    logger_fn: Callable[[str], None] = click.echo,
+    fallback_to_tempdir: bool = False,
```
schrockn, line 16: Can we push this fall back logic to the single callsite that uses it rather than adding another parameter here? I think adding these one-off booleans is q code smell and we should avoid it if possible

python_modules/dagster/dagster/_cli/utils.py
```diff
@@ -1,35 +1,68 @@
...
+    cli_command: The name of the command - optional, used for logging messages.
+    instance_ref: For commands that delegate to other commands, they can pass in
+        a serialized instance_ref object rather than loading it from a dagster.yaml file.
+    fallback_to_tempdir: What the command should do if DAGSTER_HOME is not set and no InstanceRef
+        is passed in. If true, will create a temporary instance using the local filesystem that is
+        removed when the process ends.
+    logger: Customize any log messages emitted while creating the instance.
+    """
+    dotenv_file_path = os.getcwd()
+    env_var_dict = get_env_var_dict(dotenv_file_path)
```
schrockn, line 33: why are we doing this twice?

python_modules/dagster/dagster/_cli/utils.py
```diff
@@ -1,35 +1,68 @@
...
+
+    try:
+        env_var_dict = get_env_var_dict(dotenv_file_path)
+    except:
+        logger.exception("Failure loading environment variables from .env file")
+        env_var_dict = {}
```
schrockn, line 39: not a big fan of eating all exceptions here. At a minimum would should log the inner exception. I'd argue that we should just hard fail if there is a syntax error that makes the file unparseable, to avoid confusion.

python_modules/dagster/dagster_tests/core_tests/instance_tests/test_instance.py
```diff
@@ -569,6 +572,43 @@ def test_dagster_home_not_dir():
...
+
+        with new_cwd(working_dir):
+            with environ({"DAGSTER_HOME": None}):
+                # without .env file with a DAGSTER_HOME, loading fails
+                with pytest.raises(DagsterHomeNotSetError):
+                    with get_instance_for_cli():
+                        pass
+
+                storage_dir = os.path.join(dagster_home, "my_storage")
+                # with DAGSTER_HOME and SQLITE_STORAGE_BASE_DIR set in a .env file, DagsterInstacne succeeds
```
schrockn, line 600: s/DagsterInstacne/DagsterInstance

schrockn (changes requested):

schrockn (changes requested):
Some inline comments

python_modules/dagster/dagster/_cli/utils.py
gibsondan: I count 7 callsites (basically everywhere that was using this old method that I have now repurposed - for CLI entrypoints where we are comfortable creating an ephemeral instance if no DAGSTER_HOME is set) - I could add a separate method instead? get_instance_for_cli (default) / get_possibly_ephemeral_instance_for_cli?

python_modules/dagster/dagster/_cli/utils.py
gibsondan: Actually i think if that was an issue it would be an issue before this PR too, so it should be fine to not try/catch here

python_modules/dagster/dagster/_cli/utils.py
```diff
@@ -1,40 +1,109 @@
...
+
+    # Load any env vars
+    for k, v in env_var_dict.items():
+        os.environ[k] = v
+
+    if len(env_var_dict):
+        logger.info(
+            "Loaded environment variables from .env file: "
+            + ",".join([env_var for env_var in env_var_dict]),
+        )
```
schrockn, line 25: IMO this should be a context manager that unsets the environment variables on exit for tests etc

schrockn (dismissed):
Final comment about making env var manipulation happen within a context manager

schrockn (approved):

python_modules/dagster/dagster_tests/core_tests/instance_tests/test_instance.py
```diff
@@ -569,6 +572,43 @@ def test_dagster_home_not_dir():
...
+                with open(os.path.join(working_dir, ".env"), "w", encoding="utf8") as fd:
+                    fd.write(f"DAGSTER_HOME={dagster_home}\n")
+                    fd.write(f"SQLITE_STORAGE_BASE_DIR={storage_dir}\n")
+
+                with get_instance_for_cli() as instance:
+                    assert (
+                        _runs_directory(str(storage_dir))
+                        in instance.run_storage._conn_string  # noqa: SLF001
+                    )
```
alangenfeld, line 609: failing on windows
```
E                       AssertionError: assert 'C:\\Users\\VSSADM~1\\AppData\\Local\\Temp\\tmplqngc2je\\my_storage\\history\\' in 'sqlite:///C:/Users/VSSADM~1/AppData/Local/Temp/tmplqngc2je/my_storage/history/runs.db'
E                        +  where 'C:\\Users\\VSSADM~1\\AppData\\Local\\Temp\\tmplqngc2je\\my_storage\\history\\' = _runs_directory('C:\\Users\\VSSADM~1\\AppData\\Local\\Temp\\tmplqngc2je\\my_storage')
E                        +    where 'C:\\Users\\VSSADM~1\\AppData\\Local\\Temp\\tmplqngc2je\\my_storage' = str('C:\\Users\\VSSADM~1\\AppData\\Local\\Temp\\tmplqngc2je\\my_storage')
E                        +  and   'sqlite:///C:/Users/VSSADM~1/AppData/Local/Temp/tmplqngc2je/my_storage/history/runs.db' = <dagster._core.storage.runs.sqlite.sqlite_run_storage.SqliteRunStorage object at 0x000002E699425A30>._conn_string
E                        +    where <dagster._core.storage.runs.sqlite.sqlite_run_storage.SqliteRunStorage object at 0x000002E699425A30> = <dagster._core.instance.DagsterInstance object at 0x000002E697B5C8E0>.run_storage
```
https://dev.azure.com/elementl/dagster/_build/results?buildId=19844&view=logs&jobId=5ac23fe9-6406-5dbc-6e1a-7ff483cf56a2&j=5ac23fe9-6406-5dbc-6e1a-7ff483cf56a2&t=32caa6cd-e8d3-51cf-ac93-4302b79532bc

PR #14404, opened 2023-05-23
feat(dbt): capture logs by default when streaming events

rexledesma:
## Summary & Motivation
Re-emit the dbt logs into Dagster.

## How I Tested These Changes
local


rexledesma:
* master
  * **PR #14404** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14404" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14404?utm_source=stack-comment).

schrockn (changes requested):
per inline comment. Not sure we should be spamming structured logs as opposed to just forwarding to stderr or stdout

rexledesma:
> per inline comment. Not sure we should be spamming structured logs as opposed to just forwarding to stderr or stdout

Ok, forwarding it to stdout. I believe we could get more granular with the behavior and differentiate `stdout` and `stderr`, but this adds some unneeded complexity, so I will defer this behavior until there is a crowd asking for it.

Previously, this setting was controlled by configuration, but I'm not too fond of that option. In the ideal scenario, this should be re-emitted to `stdout`/`stderr` using `logging`, and users could silence these logs by configuring the correct logger. They could even re-emit their own format of these logs by logging the events directly in the body of their computation. 

Again, for that question, I will defer its configuration until we start the logging configuration revamp. In most cases, there is no need to customize the output from these `stdout` logs.

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -254,6 +254,9 @@ def stream_raw_events(self) -> Iterator[DbtCliEventMessage]:
             log: str = raw_line.decode().strip()
             event = DbtCliEventMessage.from_log(log=log)
 
+            # Re-emit the logs from dbt CLI process into stdout.
+            print(str(event))  # noqa: T201
```
OwenKephart, line 258: nit: `sys.stdout.write()` would save us from having to have a noqa here.

OwenKephart (approved):
small comment, otherwise 👍

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
rexledesma: 🙏

PR #14349, opened 2023-05-17
[RFC] feat(dbt-assets)[7/N]: automatically invoke with_attributes

rexledesma:

PR #14411, opened 2023-05-23
feat(dbt): add argument to modify partitions def on dbt_assets

rexledesma:
## Summary & Motivation
Ideally, this should be on the `.with_attributes` decorator (https://github.com/dagster-io/dagster/pull/14349). But for now, at least give a hook into defining this setting in one place, rather than mutating the `AssetsDefinition` after the fact using the `.with_attributes` method.

## How I Tested These Changes
local, pytest

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorator.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorator.py
@@ -1,6 +1,6 @@
 from typing import Callable, Optional
 
-from dagster import AssetsDefinition, multi_asset
+from dagster import AssetsDefinition, PartitionsDefinition, multi_asset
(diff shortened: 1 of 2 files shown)


python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorator.py
```diff
@@ -14,6 +14,7 @@ def dbt_assets(
     manifest: DbtManifest,
     select: str = "fqn:*",
     exclude: Optional[str] = None,
+    partition_def: Optional[PartitionsDefinition] = None,
```
OwenKephart, line 17: partition_def -> partitions_def

OwenKephart (approved):
note comment, otherwise 👍

PR #14415, opened 2023-05-23
Fix cleanup issue with OpExecutionContext

gibsondan:
Summary:
We were always entering this contextmanager but not always leaving it, causing surprising error spew even when you use the op context as a contextmanager. Fix that.

Test Plan:

BK, and

The following script:
```
@asset
def my_asset(context):
    pass

def test_asset() -> None:
    with build_op_context() as context:
        actual = my_asset(context)

    assert False
```

No longer produces weird error spew.

### DIFF
--- a/python_modules/dagster/dagster_tests/core_tests/test_op_invocation.py
+++ b/python_modules/dagster/dagster_tests/core_tests/test_op_invocation.py
@@ -85,6 +85,18 @@ def test_op_invocation_none_arg():
     assert result == 5
 
 
+def test_op_invocation_lifecycle():
+    @op
+    def basic_op(context):
+        return 5
+
+    with build_op_context() as context:
+        pass
+
+    # Verify dispose was called on the instance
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster_tests/core_tests/test_op_invocation.py
```diff
@@ -85,6 +85,18 @@ def basic_op(_):
...
+def test_op_invocation_lifecycle():
+    @op
+    def basic_op(context):
+        return 5
+
+    with build_op_context() as context:
+        pass
+
+    # Verify dispose was called on the instance
+    assert context.instance.run_storage._held_conn.closed  # noqa
```
alangenfeld, line 97: unfortunate but understandable

alangenfeld (approved):
theres gotta be a better way to manage these with a bigger re-work but this seems like a fine iterative change

Issue #13248, opened 2023-03-30
Self-dependencies working with MultiPartitioned assets


rubenbriones:
### What's the use case?

I have a MultiPartitioned asset, where one dimension is a DailyPartition and another dimension is a static set of items (countries for example), and I do not want to materialize any partition before their prior day partition/country have succeed.

I would be nice to define something like it is [here](https://github.com/dagster-io/dagster/discussions/11829), but for multipartitions:

```py
multi_partitions_def = MultiPartitionsDefinition(
    {
        "country": StaticPartitionsDefinition(["USA", "ESP"]),
        "date": DailyPartitionsDefinition(start_date="2023-03-17"),
    }
)

@asset(
    ins={
        'asset1': AssetIn(
            partition_mapping=TimeWindowPartitionMapping(start_offset=-1, end_offset=-1),
            dagster_type=Nothing,
        )
    },
    partitions_def=multi_partitions_def,
)
def asset1():
    pass
```

But nowadays it raises the next exception: `dagster._check.CheckError: Failure condition: Assets with self-dependencies must have time-window partitions`

Thanks in advance!

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

sryza:
@clairelin135 are you able to take a look at this one when you get a chance?

clairelin135:
Would be great to enable this in the multidimensional case too.

I imagine that we'd constrain this behavior to multipartitions defs that contain a single time dimension, where each multipartition depends on earlier time-dimension partitions with the same value in other dimensions.

I'll track it in our backlog and get around to this when I get a chance.

hendrikserruys:
Hi there! Just wanted to say I'm dealing with the same issue.

chaosGuppy:
Also interested in this, working around it with the following monkeypatch for now
```python
def _validate_self_deps(
    input_keys: Iterable[AssetKey],
    output_keys: Iterable[AssetKey],
    partition_mappings: Mapping[AssetKey, PartitionMapping],
) -> None:
    output_keys_set = set(output_keys)
    for input_key in input_keys:
        if input_key in output_keys_set:
            is_valid = False
            if input_key in partition_mappings:
                partition_mapping = partition_mappings[input_key]
                if (
                    isinstance(partition_mapping, TimeWindowPartitionMapping)
                    and (partition_mapping.start_offset or 0) < 0
                    and (partition_mapping.end_offset or 0) < 0
                ):
[...]

PR #14418, opened 2023-05-23
Enable multipartitioned assets to contain self dependencies

clairelin135:
Resolves https://github.com/dagster-io/dagster/issues/13248

**Test Plan**
- Tested locally by launching runs and backfills
- BK tests for asset backfills, asset reconciliation, unit tests

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/assets.py
@@ -24,7 +24,9 @@ from dagster._core.definitions.auto_materialize_policy import AutoMaterializePol
 from dagster._core.definitions.freshness_policy import FreshnessPolicy
 from dagster._core.definitions.input import In
 from dagster._core.definitions.metadata import ArbitraryMetadataMapping
+from dagster._core.definitions.multi_dimensional_partitions import MultiPartitionsDefinition
 from dagster._core.definitions.op_selection import get_graph_subset
+from dagster._core.definitions.partition_mapping import MultiPartitionMapping
 from dagster._core.definitions.time_window_partition_mapping import TimeWindowPartitionMapping
 from dagster._core.definitions.time_window_partitions import TimeWindowPartitionsDefinition
 from dagster._core.errors import DagsterInvalidDefinitionError, DagsterInvalidInvocationError
@@ -236,6 +238,7 @@ class AssetsDefinition(ResourceAddable, IHasInternalInit):
             input_keys=self._keys_by_input_name.values(),
             output_keys=self._selected_asset_keys,
             partition_mappings=self._partition_mappings,
+            partitions_def=self._partitions_def,
         )
 
     @staticmethod
@@ -1375,20 +1378,39 @@ def _validate_self_deps(
     input_keys: Iterable[AssetKey],
     output_keys: Iterable[AssetKey],
     partition_mappings: Mapping[AssetKey, PartitionMapping],
+    partitions_def: Optional[PartitionsDefinition],
 ) -> None:
+    is_valid_self_dep_mapping = (
+        lambda partition_mapping: isinstance(partition_mapping, TimeWindowPartitionMapping)
+        and (partition_mapping.start_offset or 0) < 0
+        and (partition_mapping.end_offset or 0) < 0
+    )
     output_keys_set = set(output_keys)
     for input_key in input_keys:
         if input_key in output_keys_set:
             if input_key in partition_mappings:
                 partition_mapping = partition_mappings[input_key]
                 if (
-                    isinstance(partition_mapping, TimeWindowPartitionMapping)
-                    and (partition_mapping.start_offset or 0) < 0
-                    and (partition_mapping.end_offset or 0) < 0
+                    isinstance(partitions_def, TimeWindowPartitionsDefinition)
+                    and is_valid_self_dep_mapping(partition_mapping)
+                ) or (
+                    (
+                        isinstance(partitions_def, MultiPartitionsDefinition)
+                        and partitions_def.has_one_time_window_dimension
+                        and isinstance(partition_mapping, MultiPartitionMapping)
+                        and partitions_def.time_window_dimension.name
+                        in partition_mapping.downstream_mappings_by_upstream_dimension
+                        and is_valid_self_dep_mapping(
+                            partition_mapping.downstream_mappings_by_upstream_dimension[
+                                partitions_def.time_window_dimension.name
+                            ].partition_mapping
+                        )
+                    )
                 ):
                     continue
 
             raise DagsterInvalidDefinitionError(
-                "Assets can only depend on themselves if they are time-partitioned and each"
-                " partition depends on earlier partitions"
+                "Assets can only depend on themselves if they are:\n(a) time-partitioned and each"
+                " partition depends on earlier partitions\n(b) multipartitioned, with one time"
+                " dimension that depends on earlier time partitions"
             )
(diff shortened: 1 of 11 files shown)


python_modules/dagster/dagster/_core/definitions/assets.py
```diff
@@ -1375,20 +1378,39 @@ def _validate_self_deps(
     input_keys: Iterable[AssetKey],
     output_keys: Iterable[AssetKey],
     partition_mappings: Mapping[AssetKey, PartitionMapping],
+    partitions_def: Optional[PartitionsDefinition],
 ) -> None:
+    is_valid_self_dep_mapping = (
+        lambda partition_mapping: isinstance(partition_mapping, TimeWindowPartitionMapping)
+        and (partition_mapping.start_offset or 0) < 0
+        and (partition_mapping.end_offset or 0) < 0
+    )
```
OwenKephart, line 1387: I'm thinking we can clean this up with something like:

```
time_partitions_def: Optional[TimeWindowPartitionsDefinition] = get_time_partitions_def(partitions_def)
```

... then

python_modules/dagster/dagster/_core/definitions/assets.py
```diff
@@ -1375,20 +1378,39 @@ def _validate_self_deps(
...
     output_keys_set = set(output_keys)
     for input_key in input_keys:
         if input_key in output_keys_set:
             if input_key in partition_mappings:
                 partition_mapping = partition_mappings[input_key]
                 if (
-                    isinstance(partition_mapping, TimeWindowPartitionMapping)
-                    and (partition_mapping.start_offset or 0) < 0
-                    and (partition_mapping.end_offset or 0) < 0
+                    isinstance(partitions_def, TimeWindowPartitionsDefinition)
```
OwenKephart, line 1394:
```
if time_window_partitions_def is not None:
   time_window_partition_mapping = get_time_window_partition_mapping(partition_mapping)
   if (
       time_window_partition_mapping is not None
       and (partition_mapping.start_offset or 0) < 0
       and (partition_mapping.end_offset or 0) < 0
   ):
       continue
 ```
 
 where those proposed functions handle the mess of sorting out the different cases (partitions_def is None, partitions_def is TimeWindow, partitions_def is Multi with no time dimension, etc.)
 
 this way we don't have to duplicate this work in a bunch of callsites

File: python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py [hunk 3]
```diff
@@ -426,7 +435,11 @@ def get_cron_schedule(
         ).get_cron_schedule(minute_of_hour, hour_of_day, day_of_week, day_of_month)
 
     @property
-    def timezone(self) -> Optional[str]:
+    def fmt(self) -> str:
+        return cast(TimeWindowPartitionsDefinition, self.time_window_dimension.partitions_def).fmt
+
+    @property
+    def timezone(self) -> str:
```
OwenKephart, line 442: I have a similar comment about the pre-existing timezone property (which I think is just used in a couple places, where we can refactor)

In general, I think we should be consistent in the pattern that we expect callsites to have. Either we should treat a MultiPartitionsDefinition exactly like a TimeWindowPartitionsDefinition, and access the same properties (I'd rather not), or we should have a standard method to extract the time window dimension stuff from a multi partitions definition, as proposed above

note that this would somewhat obsolete `has_one_dimension_time_window_partitioning`, which I think is ok, as in a few places we call this method and then immediately do the work to parse out the TimeWindowPartitionsDefinition

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
```diff
@@ -414,6 +414,15 @@ def time_window_dimension(self) -> PartitionDimensionDefinition:
         )
         return next(iter(time_window_dims))
 
+    @property
+    def has_one_time_window_dimension(self) -> bool:
```
OwenKephart, line 418: would prefer not to add this function in light of other comments

OwenKephart (changes requested):
note comments -- I know I'm being annoying and requesting things that are somewhat outside the bounds of this review, but I think it's better we adjust this pattern sooner rather than later as it'll just get more difficult as time goes on

python_modules/dagster/dagster/_core/definitions/multi_dimensional_partitions.py
clairelin135: Initially my plan was to treat a MultiPartitionsDefinition like a TimeWindowPartitionsDefinition, since there likely is a maximum of 1 time dimension and all of these functions can just call the underlying `TimeWindowPartitionsDefinition`. method. 

I'm fine with refactoring this, but I'm wondering why you don't want to have a pattern like this?

OwenKephart: Yeah I should have explained my reasoning a bit more here. The basic idea is that in most callsites, we have an input of type `Optional[PartitionsDefinition]`. There are many categories this `partitions_def` can be in. Either:

- None
- TimeWindowPartitionsDefinition
- MultiPartitionsDefinition (exactly one time window dimension)
- MultiPartitionsDefinition (not exactly one time window dimension)
- Other PartitionsDefinition (static/dynamic)

Now, in any callsite that accepts a parameter of type PartitionsDefinition, where we want to treat "timey" partitions differently, we'll need to have two separate conditionals, one to check if it's a regular TimeWindowPartitionsDefinition, and another to check if it's a MultiPartitionsDefinition with exactly one time dimension.

Once we have validated if we're in a "timey" case, then we often want to pass that "timey" partitions def into another function. In this case, we're doing that for `_sort_key_for_time_window_partition`, but we also do a similar thing in [this function](https://sourcegraph.com/github.com/dagster-io/dagster/-/blob/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py?L165), for example. In each of these cases, we need to include extra logic to disambiguate between the TimeWindowPartitionsDefinition and MultiPartitionsDefinitions, as even if they share some properties, there are some parts which are different (e.g. in the sort key case, we have to pull out the partition key in a slightly different way, even if we get the `fmt` in the same way)

So this would all be a bit simpler if we could just have `_sort_key_for_time_window_partition` just take in a TimeWindowPartitionsDefinition object directly, regardless of if it came from a MultiPartitionsDef or not.

This also helps prevent mistakes where we accidentally pass in a MultiPartitionsDefinition that does not have exactly one time dimension, by letting us rely on the static type checker.

clairelin135: Thanks for explaining.

I'll change the code around to make `_sort_key_for_time_window_partition` accept a time partitions def, though at this callsite (and other related callsites like the one you mentioned) there has to be a point where the code diverges to distinguish between time partitioned and multipartitioned keys.

OwenKephart: that's a good point about having to distinguish between time partition and multi partition keys -- in that case, could we have a `get_time_partition_key(partitions_def, partition_key)` to handle that (then pass in both the partitions_def and the time_partition_key into that function)?

my basic feeling is that if we want to have a function that handles something with time partitions, it'd be great to be able to only pass it things that are already in the language of time partitions, so that it can be completely concerned with handling time-based stuff

clairelin135:
@OwenKephart thanks for helping me think through this! I amended the PR to add the `get_time_partitions_def` and `get_time_partition_key` methods, but I didn't add the `get_time_partition_mapping` one yet since there aren't other places where this same logic is reused.

I think this is ready for another look

OwenKephart (approved):
nice

PR #14421, opened 2023-05-23
[ui] Repair path-prefix

hellendag:
## Summary & Motivation

With the upgrade to Storybook, `@types/webpack` was eliminated, which removed the global type declaration for `__webpack_public_path__`. In an effort to repair the resulting type error, I added a `(window as any)` to the variable in `publicPath.ts`. It appears that this is what led to the breakage of path prefixing.

By manually adding a type declaration for this Webpack global in our own typedef file, I can remove the `(window as any)` annotation, and the variable goes back to working properly.

## How I Tested These Changes

`yarn build`, load dagit with a path-prefix. Verify that the app loads, and that all JS assets are retrieved properly.

### DIFF
--- a/js_modules/dagit/packages/app/src/publicPath.ts
+++ b/js_modules/dagit/packages/app/src/publicPath.ts
@@ -10,7 +10,7 @@ const extracted = extractInitializationData();
 // Set the webpack path prefix based on DOM value. This will be used
 // for dynamically loaded bundles.
 if (typeof extracted.pathPrefix === 'string') {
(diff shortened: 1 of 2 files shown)


hellendag:
* master
  * **PR #14421** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14421" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14421?utm_source=stack-comment).

PR #14424, opened 2023-05-23
feat(dbt): add method to retrieve artifacts from dbt invocation

rexledesma:
## Summary & Motivation
Here, we provide an API for the user to retrieve the dbt artifacts associated with their dbt CLI invocation. We take inspiration from two sources:

- [In the dbt Cloud API](https://docs.getdbt.com/dbt-cloud/api-v2#tag/Runs/operation/getArtifactsByRunId), artifacts are retrieved by path.
- Multiple dbt tasks could be happening at the same time. However, we want each task's artifacts to be independent: a newer task's artifacts should not overwrite a previous task's artifacts. To do this, we just manipulate the [`DBT_TARGET_PATH` variable](https://docs.getdbt.com/reference/project-configs/target-path), so that it is unique for each task. [This has been recommended before](https://discourse.getdbt.com/t/multiple-run-results-json-and-manifest-json-files/7555).

## How I Tested These Changes
pytest, local

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
@@ -2,6 +2,7 @@ import json
 import os
 import subprocess
 import sys
+import time
 from dataclasses import dataclass
 from typing import Any, Dict, Iterator, List, Mapping, Optional, Sequence, Union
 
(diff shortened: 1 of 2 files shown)


rexledesma:
* master
  * **PR #14424** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14424" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14424?utm_source=stack-comment).

OwenKephart (approved):
nice

PR #14428, opened 2023-05-23
feat(dbt): automatically apply asset key replacements using the manifest

rexledesma:
## Summary & Motivation
Because we have access to the manifest, the source of truth of mappings, and the underlying `AssetsDefinition`, just apply the new mappings onto the created definition. We do not need to wait for the user to do it.

Originally, this was going to live in the `.with_attributes` decorator, but this seems more ergonomic.

## How I Tested These Changes
pytest

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorator.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorator.py
@@ -63,6 +63,9 @@ def dbt_assets(
             },
         )(fn)
 
-        return asset_definition
+        return asset_definition.with_attributes(
+            input_asset_key_replacements=manifest.asset_key_replacements,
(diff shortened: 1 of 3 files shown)


rexledesma:
* master
  * **PR #14411** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14411" target="_blank">Graphite</a> 
    * **PR #14428** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14428" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14428?utm_source=stack-comment).

OwenKephart (approved):
I think it's possible that we'll want to rename `node_info_to_asset_key` to something like `get_asset_key`, as the former is a little ugly, but I think this is fine for now

PR #14435, opened 2023-05-23
When building dagit url paths from python change from instance/runs to runs

OwenKephart:
## Summary & Motivation

We removed the `/instance` prefix here: https://github.com/dagster-io/dagster/commit/c35644392244b646efe22c8117c3b233023b6d72, but we have a bunch of hooks that still point to the old path.

Just did a find/replace to change all relevant locations.


ruotianwang (commented):
Thanks for the fix!

PR #14436, opened 2023-05-24
Add configurable resource key to create_databricks_run_now_op

srggrs:
## Summary & Motivation
When having different workspaces/accounts in your pipeline, you need to create different databricks resources to access them (ie using different credentials)

## How I Tested These Changes
This change is so small that I am not sure how to test it. Also a default value is provided (previously hardcoded), so default implementation should run with no worries.

### DIFF
--- a/python_modules/libraries/dagster-databricks/dagster_databricks/ops.py
+++ b/python_modules/libraries/dagster-databricks/dagster_databricks/ops.py
@@ -46,7 +46,7 @@ def create_databricks_run_now_op(
         name (Optional[str]): The name of the op. If not provided, the name will be
             _databricks_run_now_op.
         databricks_resource_key (Optional[str]): The name of the resource key used by this op. If not
-        provided, the resource key will be "databricks".
+            provided, the resource key will be "databricks".
 
     Returns:
         OpDefinition: An op definition to run the Databricks Job.
@@ -138,6 +138,7 @@ def create_databricks_submit_run_op(
     poll_interval_seconds: float = DEFAULT_POLL_INTERVAL_SECONDS,
     max_wait_time_seconds: float = DEFAULT_MAX_WAIT_TIME_SECONDS,
     name: Optional[str] = None,
+    databricks_resource_key: Optional[str] = "databricks",
 ) -> OpDefinition:
     """Creates an op that submits a one-time run of a set of tasks on Databricks.
 
@@ -154,6 +155,8 @@ def create_databricks_submit_run_op(
             before raising an error.
         name (Optional[str]): The name of the op. If not provided, the name will be
             _databricks_submit_run_op.
+        databricks_resource_key (Optional[str]): The name of the resource key used by this op. If not
+            provided, the resource key will be "databricks".
 
     Returns:
         OpDefinition: An op definition to submit a one-time run of a set of tasks on Databricks.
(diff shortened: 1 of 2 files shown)


python_modules/libraries/dagster-databricks/dagster_databricks/ops.py
```diff
@@ -98,7 +101,7 @@ class DatabricksRunNowOpConfig(Config):
 
     @op(
         ins={"start_after": In(Nothing)},
-        required_resource_keys={"databricks"},
+        required_resource_keys={databricks_resource_key},
```
OwenKephart, line 104: github won't let me comment on lines that aren't modified by this PR, but on line 111, you won't be able to access `context.resources.databricks` in the case that `databricks_resource_key` is not `"databricks"`. Instead, you can follow the pattern seen here: https://sourcegraph.com/github.com/dagster-io/dagster/-/blob/python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py?L282 to access the resource.

OwenKephart (changes requested):
Hi @srggrs ! Thanks for the contribution. I think this change looks reasonable, but you'll want to update the code as described in the comment.

For testing, you can parameterize this test: https://sourcegraph.com/github.com/dagster-io/dagster/-/blob/python_modules/libraries/dagster-databricks/dagster_databricks_tests/test_ops.py?L74:10 to execute with a different databricks_resource_key.

For completeness, it'd be great if this parameter could be added to `create_databricks_submit_run_op` as well, as it might be confusing if it were available on one of these methods but not another.

python_modules/libraries/dagster-databricks/dagster_databricks/ops.py
srggrs: yeah make sense! Will add those changes as well

srggrs:
Added test and changes for `databricks_submit_run_op` as well. See local test evidence
image

python_modules/libraries/dagster-databricks/dagster_databricks/ops.py
```diff
@@ -25,6 +25,7 @@ def create_databricks_run_now_op(
     poll_interval_seconds: float = DEFAULT_POLL_INTERVAL_SECONDS,
     max_wait_time_seconds: float = DEFAULT_MAX_WAIT_TIME_SECONDS,
     name: Optional[str] = None,
+    databricks_resource_key: Optional[str] = "databricks",
```
OwenKephart, line 28: This type should just be `str`, not `Optional[str]`, as `None` is not a valid key to the required_resource_keys dictionary.

python_modules/libraries/dagster-databricks/dagster_databricks_tests/test_ops.py
```diff
@@ -52,27 +52,33 @@ def databricks_client_factory(request):
...
+        "empty Databricks job configuration, and default databricks resource key",
+        "Databricks job configuration with python params and custom databricks resource key",
     ],
 )
 def test_databricks_run_now_op(
-    databricks_client_factory, mocker: MockerFixture, databricks_job_configuration: Optional[dict]
+    databricks_client_factory,
+    mocker: MockerFixture,
+    databricks_job_configuration: Optional[dict],
+    databricks_resource_key: Optional[dict],
```
OwenKephart, line 81: This should be `Optional[str]`

python_modules/libraries/dagster-databricks/dagster_databricks_tests/test_ops.py
```diff
@@ -107,7 +125,20 @@ def test_databricks_job() -> None:
...
+    ids=[
+        "no databricks resource key",
+        "default databricks resource key",
+        "custom databricks resource key",
+    ],
+)
+def test_databricks_submit_run_op(
+    databricks_client_factory,
+    mocker: MockerFixture,
+    databricks_resource_key: Optional[dict],
```
OwenKephart, line 140: `Optional[str]`

python_modules/libraries/dagster-databricks/dagster_databricks/ops.py
```diff
@@ -135,6 +138,7 @@ def create_databricks_submit_run_op(
...
```
OwenKephart, line 141: `str`

OwenKephart (changes requested):
This is looking really close! There are a few issues w/ the type annotations that will need to be fixed -- if you run `make pyright` at the top level of the dagster repo that will let you know if you have any remaining type issues (I might have missed some in my comments).

Thanks for working on this!

OwenKephart (approved):
Thanks for the contribution! Merging now

Issue #14305, opened 2023-05-16
PartitionMapping for last partition of asset depends unpartitioned asset and earlier partitions are uncomputable


sryza:
### Motivation

The situation:
- Upstream asset A is overwritten daily.
- Every day, as snapshot of A is added to downstream asset B.
- A is represented with a non-partitioned asset and B with a daily-partitioned asset.

The problem:
- Dagster assumes that every partition of B depends on A.
- When A is updated, this means that Dagster thinks every partition of B is stale.
- However, it would be impossible to materialize those partitions to freshen them, because they depend on data that's no longer available in A.
- Auto-materialize policies won't materialize anything that's downstream of a stale asset.
- Hence, nothing downstream of B can ever be auto-materialized.

### The potential solution

Introduce some kind of PartitionMapping that says:
- The last partition of B depends on A
- Earlier partitions of B depend on data that no longer exist, and thus are un-computeable

clairelin135:
An idea for how this could look:
```
UnpartitionedToTimeDimensionPartitionMapping(num_dependent_time_partitions_from_end: int = 1)
```
This would allow the user to define that the N latest time partitions depend on the upstream asset, and that prior partitions are not dependent. By default, we assume that only the last time partition is dependent.

I also considered `UnpartitionedToTimePartitionedPartitionMapping`, but thought that it would be worth supporting unpartitioned -> multipartitioned relationships with this partition mapping as well. In this multipartitioned case, each 2nd dimension partition with the latest time partition key would be a dependent partition.

In the future, we could also support unpartitioned -> static relationships. This could be implemented through some shared `UnpartitionedToPartitionedPartitionMapping` base class.

sryza:
`UnpartitionedToTimePartitionedPartitionMapping` sounds like a solid name, and I think it might be fair to use it even in the case where the downstream is a multi-partitioned asset with a time dimension.

`num_dependent_time_partitions_from_end` doesn't seem as clear to me. It's also not clear to me that it makes sense to support a value > 1 here. Maybe we could replace it with a boolean? Still not sure what the name would be though.

There's also the question of what the default behavior should be. I think it might be a breaking change for the default behavior to be that only the last partition of the partitioned asset depends on the unpartitioned asset.

[...]

PR #14449, opened 2023-05-24
Enable different start/end partitions defs in TimeWindowPartitionMapping

clairelin135:
Requested by a user: https://dagster.slack.com/archives/C01U954MEER/p1682972599494859

Sometimes, a downstream time-partitioned asset Z might be dependent on an upstream time-partitioned asset Y that starts later than B. For example:
- Assets Z and X are partitioned daily, starting 2022-01-01
- Logic was changed recently to make Z dependent on X _and_ Y, but Y only exists starting on 2022-06-01
- Z should be able to execute on partitions before 2022-06-01, with Y being passed in as a null value

Selecting one of these partitions with a nonexistent upstream errors within a call to `get_upstream_partitions_for_partitions`.

This PR enables this behavior by adding an additional `raise_error_on_nonexistent_upstream_partition` arg to `TimeWindowPartitionsDefinition`, which when set to True (the default) raises an error when upstream partitions fall outside of the start-end time range of their partitions def. 

For the users above in this special case, they can set this bool to False. In this case, no error will be raised and the method will just filter for existent partitions. There are no changes to behavior for when `get_downstream_partitions_for_partitions` is called.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
+++ b/python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
@@ -298,17 +298,6 @@ class TimeWindowPartitionsDefinition(
             )
         return partition_def_str
 
-    def __eq__(self, other):
-        return (
-            isinstance(other, TimeWindowPartitionsDefinition)
-            and pendulum.instance(self.start, tz=self.timezone).timestamp()
-            == pendulum.instance(other.start, tz=other.timezone).timestamp()
-            and self.timezone == other.timezone
-            and self.fmt == other.fmt
-            and self.end_offset == other.end_offset
-            and self.cron_schedule == other.cron_schedule
-        )
-
     def __repr__(self):
         # Between python 3.8 and 3.9 the repr of a datetime object changed.
         # Replaces start time with timestamp as a workaround to make sure the repr is consistent across versions.
--- a/python_modules/dagster/dagster/_core/definitions/time_window_partition_mapping.py
+++ b/python_modules/dagster/dagster/_core/definitions/time_window_partition_mapping.py
@@ -24,7 +24,7 @@ class TimeWindowPartitionMapping(
         [
             ("start_offset", PublicAttr[int]),
             ("end_offset", PublicAttr[int]),
-            ("raise_error_on_nonexistent_upstream_partition", PublicAttr[bool]),
+            ("allow_nonexistent_upstream_partitions", PublicAttr[bool]),
         ],
     ),
 ):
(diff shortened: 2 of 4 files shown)


python_modules/dagster/dagster/_core/definitions/time_window_partitions.py
```diff
@@ -298,17 +298,6 @@ def __str__(self) -> str:
             )
         return partition_def_str
 
-    def __eq__(self, other):
```
clairelin135, line 301: Tests in this PR exposed a bug with this function. We recently added an `end` param to time window partitions defs, and since `end` is not included in this function, this function falsely evaluated defs with different end values as equivalent.

I don't think this function is necessary though, since this class is a named tuple so it should already have an `__eq__` function that we don't need to maintain like this one.

sryza:
This reminds me of the discussion we've been having about a partitioned asset that snapshots an unpartitioned asset.

As I mentioned on https://github.com/dagster-io/dagster/issues/14305#issuecomment-1552161013, I think we should try to come up with some vocabulary to talk about the situation where a partition isn't materializable because the upstream data that would be required to compute it isn't present.

In the future, instead of just raising an error, the UI could actually give an indication that this is the case.  E.g. if you tried to select those partitions in the backfill dialog, we could prevent you and pop up a tooltip explaining the situation.

I don't think we need to do that now, but I think we should try to find a name that would still make sense if/when we do do it.

sryza (approved):

sryza:
Btw should we mark this parameter as experimental just to be safe?

clairelin135:
> There is one edge case where this PR does not respect the bool setting. This is when a partitions def of smaller granularity is downstream of a partitions def with greater granularity, i.e. if a hourly partitions def is downstream of a daily partitions def. If current time = 2023-01-01-12:00, hourly partitions exist up until 2023-01-01-12:00, but no daily partition exists yet for 2023-01-01. It would be surprising if one of those hourly partitions could run when the upstream partition does not yet exist, so we still raise an error in this case, since the upstream partition will exist in the future. This case should be pretty unlikely since I can't think of a good reason why someone would want a smaller time partition granularity downstream of a time partitions def with larger granularity. 

@sryza would appreciate a second look at this. 

Previously this PR raised an error when encountering the edge case above regardless of if `allow_nonexistent_upstream_partitions` was set, but after reading https://github.com/dagster-io/dagster/discussions/14693 I decided to fully respect the `allow_nonexistent_upstream_partitions` bool and not raise an error. I think the situation detailed in the discussion is a much more common situation where you wouldn't want an error to be yielded. If you are in the weird edge case above where an hourly is downstream of a daily, if `allow_nonexistent_upstream_partitions` is set, you'll just have to set custom logic in your asset to prevent execution for "hanging partitions" if desired.

sryza (approved):

PR #14456, opened 2023-05-25
[ui] Clean up loading and error states on op job partitions view

hellendag:
## Summary & Motivation

We're currently showing `null` on the op job partitions view when the query response has errors, which doesn't give the user much information about what's going on.

Instead, surface a few error states, including Python errors. I'm also removing the `<Loading>` component usage here, because I think we need to rethink that component.

I don't think this will help us track down the weird JS error spotted in Cloud this morning, but it should at least give the user a bit more context about what's going on on the page.

## How I Tested These Changes

View a partition tab in a case where Python errors are being returned in the response, verify correct loading state, and verify that the red box appears on the page post-response.

View partition tabs on successful pages, verify that they render correctly.

### DIFF
--- a/js_modules/dagit/packages/core/src/partitions/OpJobPartitionsView.tsx
+++ b/js_modules/dagit/packages/core/src/partitions/OpJobPartitionsView.tsx
@@ -1,12 +1,23 @@
 import {gql, useQuery} from '@apollo/client';
-import {Box, Button, Dialog, Icon, Tooltip, Colors, Subheading, useViewport} from '@dagster-io/ui';
+import {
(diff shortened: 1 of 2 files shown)


hellendag:
* master
  * **PR #14456** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14456" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14456?utm_source=stack-comment).

salazarm (approved):

PR #14462, opened 2023-05-25
Purge asset daemon evaluations

johannkm:
Purge evaluations like ticks after 1 week. Maybe we'll make this configurable in the future but I don't think that's necessary off the bat


python_modules/dagster/dagster/_core/storage/schedules/sql_schedule_storage.py
```diff
@@ -511,6 +511,17 @@ def get_auto_materialize_asset_evaluations(
             rows = conn.execute(query)
             return [AutoMaterializeAssetEvaluationRecord.from_db_row(row) for row in rows]
 
+    def purge_evaluations(self, before: float):
```
OwenKephart, line 514: should this be called purge_asset_evaluations?

johannkm: sure

OwenKephart (approved):

PR #14463, opened 2023-05-25
[telemetry] mark functional resources as dagster maintained

jamiedemaria:
## Summary & Motivation
marks all function resources with `dagster_maintained_resource` decorator for telemetry


benpankow (approved):
sweet

PR #14479, opened 2023-05-26
[ui] Guard against partition set status error

hellendag:
## Summary & Motivation

An error can occur on the op job partitions view when the partition status field returns with an error. We have no way to map the partitions back to the statuses, and an unreachable assert is hit because we try to evaulate a status string on `undefined`.

Work around this by allowing `undefined` to be treated as a "status", rendering as dark gray and using the label "Unknown". This should ideally not happen often, and seems better than erroring out the whole page because of the faulty error handling.

## How I Tested These Changes

Added a jest spec with a fragment matching the error state.

Raise an error from `resolve_partitionStatusesOrError` in Python and load an op job partitions view. Verify that the partition bar loads entirely dark gray, but that the page doesn't crash.

Screenshot 2023-05-25 at 8 58 45 PM

### DIFF
--- a/js_modules/dagit/packages/core/src/partitions/__tests__/OpJobPartitionsViewContent.test.tsx
+++ b/js_modules/dagit/packages/core/src/partitions/__tests__/OpJobPartitionsViewContent.test.tsx
@@ -1,0 +1,35 @@
+import {MockedProvider} from '@apollo/client/testing';
+import {render, screen} from '@testing-library/react';
+import * as React from 'react';
+
+import {buildRepoAddress} from '../../workspace/buildRepoAddress';
+import {OpJobPartitionsViewContent} from '../OpJobPartitionsView';
+import {buildOpJobPartitionSetFragmentWithError} from '../__fixtures__/OpJobPartitionsViewContent.fixtures';
+
+jest.mock('../usePartitionStepQuery', () => ({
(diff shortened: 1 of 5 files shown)


hellendag:
* master
  * **PR #14479** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14479" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14479?utm_source=stack-comment).

salazarm (approved):
lgtm💰

js_modules/dagit/packages/core/src/partitions/__tests__/OpJobPartitionsViewContent.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+}));
+
+jest.mock('../JobBackfillsTable', () => ({
+  JobBackfillsTable: () => <div />,
+}));
+
+// This file must be mocked because Jest can't handle `import.meta.url`.
+jest.mock('../../graph/asyncGraphLayout', () => ({}));
+
+describe('OpJobPartitionsViewContent', () => {
```
bengotow, line 20: Ahh thanks for starting tests for this!

PR #14480, opened 2023-05-26
change db.select calls to wrapped db_select call

prha:

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/event_log/migration.py
+++ b/python_modules/dagster/dagster/_core/storage/event_log/migration.py
@@ -5,6 +5,7 @@ from tqdm import tqdm
 
 from dagster._core.assets import AssetDetails
 from dagster._core.events.log import EventLogEntry
+from dagster._core.storage.sql import db_select
 from dagster._serdes.serdes import deserialize_value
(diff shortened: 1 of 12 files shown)


prha:
* master
  * **PR #14480** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14480" target="_blank">Graphite</a>  👈
    * **PR #14481** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14481" target="_blank">Graphite</a> 
      * **PR #14482** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14482" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14480?utm_source=stack-comment).

gibsondan (approved):
for these will we have a test suite running with 1.3/1.4/2.0, similar to what we do with pendulum? seems easy to miss a spot here in the future and accidentally break an older version

this is all for airflow 1 compat in dagster-airflow right? We double-checked that still need to support that?

prha:
I didn't check that we needed to support that... @Ramshackle-Jamathon is still OOO, but wanted to line up the work in case we do.

Issue #14452, opened 2023-05-25
Make EnvVar functional even outside Dagster's config system


chriszs:
### What's the use case?

I've seen some confusion about when and how to use `EnvVar` in the new Python config system. In particular, a colleague noted that it returned the name of the environment variable and proceeded to use it like this: `EnvVar(os.getenv("EXAMPLE"))`. This was in part because he was using a resource in a way that accessed it before launch time, but I searched Slack to try to better understand how to use it and found several people confused about this. And I didn't quite understand it at first when I looked up the `EnvVar` class. It appears to be just a string. That something else happens to it later in the resource system to access the value is a little opaque.

So, what if it also functioned as a regular, but lazy way to access the variable? This would allow you to use it as drop in replacement for `os.getenv` while still allowing the config system to model it and fill it in at launch time. It would allow you to remove references to `os.getenv` in the docs while relying somewhat less on other parts of the system to do magic.

### Ideas of implementation

Here's my idea:

```python
from typing import Optional
import os
from dataclasses import dataclass

@dataclass
class EnvVar(str):

    name: str
    default: Optional[str] = None

    def __str__(self) -> str:
        if self.name not in os.environ and self.default is None:
            raise ValueError(f"Environment variable {self.name} not set")
        return str(os.getenv(self.name, self.default))

    def __repr__(self) -> str:
        return f"EnvVar('{self.name}')"
```

Example usage:

```python
@dataclass
class TestClass:
    test: str

    def create_resource(self):
        print(self.test)


# variable not accessed
test = TestClass(test=EnvVar("TEST"))

# variable accessed
test.create_resource()
```

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

PR #14488, opened 2023-05-26
[pythonic resources][rfc] Adjust EnvVar behavior to allow for direct usage

benpankow:
## Summary

Small RFC in response to #14452. Keeps `EnvVar` as a direct subclass of `str` but implements a custom override of `__str__` to fetch the value from the environment in the case that the user attempts to use the `EnvVar` anywhere.

In the case that it's passed into a resource or config, the resolution will still happen in a lazy manner.

A potentially compelling alternative could be to error on `__str__` to explicitly prevent users from using `EnvVar` outside of resource/config contexts. Right now we're sort of in the worst of both worlds where a user can use an `EnvVar` in inapproporiate places and they get an odd result (see user comment in the above issue). (See #14490 for an example of this)

## Test Plan

Small unit test, existing unit tests. Will expand suite if this is a direction we want to go.

### DIFF
--- a/python_modules/dagster/dagster/_config/field_utils.py
+++ b/python_modules/dagster/dagster/_config/field_utils.py
@@ -1,5 +1,6 @@
 # encoding: utf-8
 import hashlib
+import os
 from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Mapping, Sequence
 
 import dagster._check as check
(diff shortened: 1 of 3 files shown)


benpankow:
* master
  * **PR #14488** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14488" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14488?utm_source=stack-comment).

schrockn:
I think this is too magical and will end up confusing a different set of users for different reason.. I agree with having the conversion to the underlying environment name being an explicit method for our own internal purposes. I prefer https://github.com/dagster-io/dagster/pull/14490

schrockn (changes requested):
Definitely seems too magical. Seem comment in thread.

PR #14490, opened 2023-05-26
[pythonic resources][rfc] Adjust EnvVar behavior to explicitly disallow direct usage

benpankow:
## Summary

Alternative RFC to #14488. Keeps `EnvVar` as a direct subclass of `str` but errors on resolving the value via `__str__` to explicitly prevent users from using `EnvVar` outside of resource/config contexts. Right now we're sort of in the worst of both worlds where a user can use an `EnvVar` in inapproporiate places and they get an odd result.

1. Users who try to use an `EnvVar` value as a string will hit an error:

```python
my_env_var = EnvVar("MY_ENV_VAR")

# errors!
print(f"Hello, {my_env_var}!")
```

2. Users can directly access the `EnvVar` name or value:

```python
my_env_var = EnvVar("MY_ENV_VAR")

print(f"Env var {my_env_var.env_var_name} has value {my_env_var.get_value()}")
```

## Test Plan

Set of unit tests, existing unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_config/field_utils.py
+++ b/python_modules/dagster/dagster/_config/field_utils.py
@@ -1,7 +1,7 @@
 # encoding: utf-8
 import hashlib
 import os
-from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Mapping, Optional, Sequence
+from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Mapping, Optional, Sequence, Type
 
 import dagster._check as check
 from dagster._annotations import public
@@ -475,6 +475,16 @@ def config_dictionary_from_values(
     return check.is_dict(_config_value_to_dict_representation(None, values))
 
 
+def _create_direct_access_exception(cls: Type, env_var_name: str) -> Exception:
+    return RuntimeError(
+        f'Attempted to directly retrieve environment variable {cls.__name__}("{env_var_name}").'
+        f" {cls.__name__} defers resolution of the environment variable value until run time, and"
+        " should only be used as input to Dagster config or resources.\n\nTo access the"
+        f" environment variable value, call `get_value` on the {cls.__name__}, or use os.getenv"
+        " directly."
+    )
+
+
 class IntEnvVar(int):
     """Class used to represent an environment variable in the Dagster config system.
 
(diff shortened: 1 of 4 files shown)


benpankow:
* master
  * **PR #14490** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14490" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14490?utm_source=stack-comment).

schrockn:
This one seems better to me. We could also offer a method to access the underlying value so that users won't have to import `os` if they find that annoying.

python_modules/dagster/dagster/_config/field_utils.py
```diff
@@ -489,14 +490,61 @@ def create(cls, name: str) -> "IntEnvVar":
         var.name = name
         return var
 
+    def __int__(self) -> int:
+        """Raises an exception of the EnvVar value is directly accessed. Users should instead use
+        the `get_value` method, or use the EnvVar as an input to Dagster config or resources.
+        """
+        raise RuntimeError(
+            f'Attempted to directly retrieve environment variable IntEnvVar("{self.name}").'
+            " IntEnvVar should only be used as input to Dagster config or resources."
```
schrockn, line 499: Add:

"If you want to access the environment variable directly, you can call `get_value` on this object, or `os.getenv` directly"

python_modules/dagster/dagster/_config/field_utils.py
```diff
@@ -489,14 +490,61 @@ def create(cls, name: str) -> "IntEnvVar":
...
         return IntEnvVar.create(name=name)
+
+    def __str__(self) -> str:
+        """Raises an exception of the EnvVar value is directly accessed. Users should instead use
+        the `get_value` method, or use the EnvVar as an input to Dagster config or resources.
+        """
+        raise RuntimeError(
+            f'Attempted to directly retrieve environment variable EnvVar("{self.env_var_name}").'
+            " EnvVar should only be used as input to Dagster config or resources."
+        )
```
schrockn, line 539: Same here.

python_modules/dagster/dagster/_config/field_utils.py
schrockn: We could also save.

"The purpose of `EnvVar` is to defer retrieval to right before the step is executed, and to be viewable in the UI"

schrockn (changes requested):
Let's do a turn on the error messaging, which is the most important part of this PR.

benpankow:
Updated copy 👍

python_modules/dagster/dagster/_config/field_utils.py
```diff
@@ -489,14 +490,67 @@ def create(cls, name: str) -> "IntEnvVar":
...
     def int(cls, name: str) -> "IntEnvVar":
         return IntEnvVar.create(name=name)
+
+    def __str__(self) -> str:
+        """Raises an exception of the EnvVar value is directly accessed. Users should instead use
+        the `get_value` method, or use the EnvVar as an input to Dagster config or resources.
+        """
+        raise RuntimeError(
+            f'Attempted to directly retrieve environment variable EnvVar("{self.env_var_name}").'
+            " EnvVar defers resolution of the env var value until run time, and should only be"
```
schrockn, line 541: spell out "environment variable" instead of "env var"

schrockn (approved):
Great. Consider a common function to share between two callsites to generate error string.

PR #14491, opened 2023-05-26
[pythonic resources] Better error message for attempting to use DagsterTypes with Config/ConfigurableResource

benpankow:


benpankow:
* master
  * **PR #14491** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14491" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14491?utm_source=stack-comment).

python_modules/dagster/dagster/_config/pythonic_config/typing_utils.py
```diff
@@ -51,7 +55,28 @@ def set_actual_types_for_type_checking(
...
+        for field in annotations:
+            if not field.startswith("__"):
+                # Need try/catch because DagsterType may not be loaded when some of the base Config classes are
+                # being created
+                # Any user-created Config class will have DagsterType loaded by the time it's created, so this
+                # will only affect the base Config classes (where this error won't be an issue)
+                try:
+                    from dagster._core.types.dagster_type import DagsterType
+
+                    if isinstance(annotations[field], DagsterType):
```
schrockn, line 70: what does `annotations[field]` contain in the case where DagsterType is not loaded?

schrockn (changes requested):

python_modules/dagster/dagster/_config/pythonic_config/typing_utils.py
benpankow: `annotations[field]` is the type annotation of the field on the `Config` class. 

```python
class MyOpConfig(Config):
    foo: str
    bar: List[int]
```

In this case, `str` or `List[int]`.

In the case that `DagsterType` is not yet available, it means whatever annotation is in use must not be a `DagsterType` (e.g. a primitive `str`, `int`, another `Config` class etc).

python_modules/dagster/dagster/_core/errors.py
```diff
@@ -136,6 +136,31 @@ def __init__(
...
+        self,
+        config_class_name: str,
+        field_name: Optional[str],
+        **kwargs,
+    ):
+        self.field_name = field_name
+        super(DagsterInvalidDagsterTypeInPythonicConfigDefinitionError, self).__init__(
+            f"""Error defining Dagster config class '{config_class_name}' on field '{field_name}'. DagsterTypes cannot be used to annotate a config type. DagsterType is meant only for type checking and coercion in op and asset inputs and outputs.
+
+This config type can be a:
```
schrockn, line 154: Please reuse `PYTHONIC_CONFIG_ERROR_VERBIAGE` instead of copy paste

schrockn (changes requested):

schrockn (approved):

PR #14499, opened 2023-05-26
[RFC] feat: deprecate `file_relative_path`

rexledesma:
## Summary & Motivation
This utility function is used in the `dagster-dbt` integration a lot, and I'm not too happy about this import.

- This is subsumed by https://docs.python.org/3/library/pathlib.html, which is in the standard library.
- Finding a relative path is not the purpose of the Dagster framework, and so this function shouldn't be in the public API.
- At its very best, this could stay as an internal utility function, but I don't really see how it's more clear than just constructing the `Path` directly.

## Deprecation Plan
1. Migrate examples away from `file_relative_path`.
2. [RFC] Migrate internal utilities away from `file_relative_path`. 
3. Remove the function from the public API in `1.4.0`

## How I Tested These Changes
N/A

### DIFF
--- a/python_modules/dagster/dagster/__init__.py
+++ b/python_modules/dagster/dagster/__init__.py
@@ -579,6 +579,11 @@ _DEPRECATED: Final[Mapping[str, TypingTuple[str, str, str]]] = {
     #     "1.1.0",  # breaking version
     #     "Use Bar instead.",
     # ),
+    "file_relative_path": (
+        "dagster._utils",
+        "1.4.0",
+        "Construct the relative path using `pathlib.Path(__file__)` instead.",
+    )
 }
 
 _DEPRECATED_RENAMED: Final[Mapping[str, TypingTuple[Callable, str]]] = {


rexledesma:
* master
  * **PR #14499** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14499" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14499?utm_source=stack-comment).

schrockn:
You're correct that this probably should have never been publicly exported. 

However I will certainly continue to use this function in our own code as I think the alternative API in pathlib is wacky (overloading the `/` operator, really? Now I have to think about operator resolution order and all sorts of chicanery, it is much more difficult to jump to implementations of such things) and less amenable to autocomplete etc.

Does this result in warnings if our own integrations use this function?

rexledesma:
> You're correct that this probably should have never been publicly exported.
> 
> However I will certainly continue to use this function in our own code as I think the alternative API in pathlib is wacky (overloading the `/` operator, really? Now I have to think about operator resolution order and all sorts of chicanery, it is much more difficult to jump to implementations of such things) and less amenable to autocomplete etc.

That's fair. Implementing `__truediv__` for this behavior is cute, but pretty unusual. But we can just substitute it for `.joinpath`, which is the method equivalent.


> Does this result in warnings if our own integrations use this function?

Yup. I believe the warnings would only occur if the import happens from the top-level `dagster` module (e.g. `from dagster import file_relative_path`). If we import from the internal submodule, then no warning should occur.

So here's an approach to get this done:

1. Refactor our external integration usage to use `pathlib`.
2. Refactor internal usage of `from dagster import file_relative_path` to `from dagster._utils import file_relative_path`.
3. Deprecate `file_relative_path` (land this PR).
4. Wait until 1.4.0 release.
5. Remove `file_relative_path` from top-level exports.

Issue #14492, opened 2023-05-26
UPathIOManager strips kwargs from base_path


raetzka:
### Dagster version

dagster, version 1.3.6

### What's the issue?

When using a subclassed UPathIOManager, any kwargs including credentials are stripped when adding the file extension in [_with_extension](https://github.com/dagster-io/dagster/blob/14730fe2df1c540a752fdb0d6e689396abb82415/python_modules/dagster/dagster/_core/storage/upath_io_manager.py#LL78C19-L78C19) as a new UPath is instantiated. 

### What did you expect to happen?

The kwargs from `base_path` and derivative `path` variables are persisted similar to

```python   
def _with_extension(self, path: UPath) -> UPath:
        return UPath(f"{path}{self.extension}", **path._kwargs) if self.extension else path
```

### How to reproduce?

The following example

```python
from dagster import (
    UPathIOManager,
    OutputContext,
    InputContext,
    InitResourceContext,
    io_manager,
    materialize,
    asset
)
from upath import UPath
import pandas as pd


class TestUPathIOManager(UPathIOManager):
    extension: str = ".parquet"

    def dump_to_path(
        self, context: OutputContext, obj: pd.DataFrame, path: UPath
    ):
        with path.open("wb") as file:
            obj.to_parquet(file)

    def load_from_path(self, context: InputContext, path: UPath) -> None:
        return None


@io_manager()
def az_parquet_io_manager() -> TestUPathIOManager:
    base_path = UPath("az://data/", account_name="testazaccount")
    return TestUPathIOManager(base_path=base_path)

@asset
def test_asset() -> pd.DataFrame:
    return pd.DataFrame.from_dict({"test": [1]})

materialize(
    [test_asset],
    resources={"upath_io": az_parquet_io_manager()},
)
```

should lead to the ValueError:

```python

dagster._core.errors.DagsterExecutionHandleOutputError: Error occurred while handling output "result" of step "test_asset"::

ValueError: unable to connect to account for Must provide either a connection_string or account_name with credentials!!

```

### Deployment type

Local

### Deployment details

Ubuntu 22.04.2 LTS with Python 3.11 and

```shell
 name         : dagster
 version      : 1.3.6
 description  : The data orchestration platform built for productivity.

dependencies
 - alembic >=1.2.1,<1.6.3 || >1.6.3,<1.7.0 || >1.7.0,<1.11.0
 - click >=5.0
 - coloredlogs >=6.1,<=14.0
[...]

PR #14501, opened 2023-05-26
use with_suffix in UPathIOManager

OwenKephart:
## Summary & Motivation

resolves: https://github.com/dagster-io/dagster/issues/14492

UPath.with_suffix _replaces_ the existing suffix. So, if you have a file named `foo.bar.0`, and want to give it the extension `.txt`, directly calling UPath.with_suffix(".txt") would result in `foo.bar.txt`. This code adds the existing suffix (if any) to the extension, giving us the desired `foo.bar.0.txt`.

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/upath_io_manager.py
+++ b/python_modules/dagster/dagster/_core/storage/upath_io_manager.py
@@ -78,7 +78,7 @@ class UPathIOManager(MemoizableIOManager):
         return self.path_exists(self._get_path(context))
 
     def _with_extension(self, path: "UPath") -> "UPath":
-        return UPath(f"{path}{self.extension}") if self.extension else path
+        return path.with_suffix(path.suffix + self.extension) if self.extension else path
 
     def _get_path_without_extension(self, context: Union[InputContext, OutputContext]) -> "UPath":
         if context.has_asset_key:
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster/_core/storage/upath_io_manager.py
```diff
@@ -75,7 +75,7 @@ def has_output(self, context: OutputContext) -> bool:
         return self.path_exists(self._get_path(context))
 
     def _with_extension(self, path: UPath) -> UPath:
-        return UPath(f"{path}{self.extension}") if self.extension else path
```
jamiedemaria, line 78: did `UPath(f"{path}{self.extension}")` also truncate the suffix?

jamiedemaria (approved):
one question! but otherwise looks good. 

I guess a second question - is this worth adding a test? seems pretty small to me, but if we want to ensure suffixes don't get dropped in the future, could be worthwhile

OwenKephart:
@jamiedemaria there actually is a test for this behavior already ([here](https://sourcegraph.com/github.com/dagster-io/dagster/-/blob/python_modules/dagster/dagster_tests/storage_tests/test_upath_io_manager.py?L246))! this retains the current (good) suffix handling behavior, but does so in a way where additional properties of the original class are retained (there's some metadata that gets dropped when you create a new UPath from scratch, but does not get dropped if you use `.with_suffix`

PR #14503, opened 2023-05-26
Fix resource binding when a job is only provided by a schedule/sensor and isn't included in the top-level list of jobs

gibsondan:
Summary:
Test before was failing to map resources onto the job, now it is succeeding

Test Plan: BK

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/definitions_class.py
+++ b/python_modules/dagster/dagster/_core/definitions/definitions_class.py
@@ -142,6 +142,27 @@ def _attach_resources_to_jobs_and_instigator_jobs(
     schedules = schedules or []
     sensors = sensors or []
 
+    # Add jobs in schedules and sensors as well
+    jobs = [
+        *jobs,
+        *[
+            schedule.job
+            for schedule in schedules
+            if isinstance(schedule, ScheduleDefinition)
+            and schedule.has_loadable_target()
+            and isinstance(schedule.job, (JobDefinition, UnresolvedAssetJobDefinition))
+        ],
+        *[
+            job
+            for sensor in sensors
+            if sensor.has_loadable_targets()
+            for job in sensor.jobs
+            if isinstance(job, (JobDefinition, UnresolvedAssetJobDefinition))
+        ],
+    ]
+    # Dedupe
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -142,6 +142,27 @@ def _attach_resources_to_jobs_and_instigator_jobs(
...
+                for schedule in schedules
+                if isinstance(schedule, ScheduleDefinition)
+                and schedule.has_loadable_target()
+                and isinstance(schedule.job, (JobDefinition, UnresolvedAssetJobDefinition))
+            ],
+            *[
+                job
+                for sensor in sensors
+                if sensor.has_loadable_targets()
+                for job in sensor.jobs
```
gibsondan, line 160: i may have pythoned a little too close to the sun here

python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -142,6 +142,27 @@ def _attach_resources_to_jobs_and_instigator_jobs(
...
+    # Add jobs in schedules and sensors (dedupe if they appear in both)
+    jobs = [
+        *{
+            *jobs,
+            *[
+                schedule.job
+                for schedule in schedules
+                if isinstance(schedule, ScheduleDefinition)
+                and schedule.has_loadable_target()
+                and isinstance(schedule.job, (JobDefinition, UnresolvedAssetJobDefinition))
```
gibsondan, line 154: this is needed to satisfy pyright because schedules and sensors can apparently have a GraphDefinition in the job field? Does that ever actually happen?

schrockn:
@benpankow I think it is important that you take a look at this.

I especially want to understand the relationship of this to the code in `repository_data_builder.py` that cals `_process_and_validate_target` which is what does this lower in the stack. It feels like a code smell to have this sort of thing at two layers in the stack.

benpankow (approved):

gibsondan:
I think this is worth getting out a fix for quickly, yeah - not an uncommon thing for users to do

Issue #14073, opened 2023-05-03
UI to backfill all partitions of a set of assets with different root `PartitionsDefinition`s


sryza:
The backfill UI allows launching backfills over assets with different `PartitionsDefinition`s, but all the root assets of the backfill must have the same `PartitionsDefinition`.

The GraphQL mutation for launching backfills is a little more flexible than the UI. If the `allPartitions` flag is set to true, then the backend can handle selections of assets that have different `PartitionsDefinition`s at the root.

We could update the UI to support this.

### What we've heard:
- Rui Zhang - https://dagster.slack.com/archives/C04KC6XRU8N/p1683070203032799?thread_ts=1683050533.679829&cid=C04KC6XRU8N
  -  It would be nice if a user can select a bunch of assets and toggle a "backfill all partition" option without needing to specify the partitions when launching a backfill.

bengotow:
Hey @sryza! This seems pretty reasonable - I talked about it with @braunjj and I'm wondering if we should hold off on this until we have a preview endpoint that will let us show the number of runs that will be launched and/or a way to implement the "only missing" checkbox for partition mapped asset backfills? It seems like right now, this option would really only be useful once, and then it'd do a lot of extra work each time it was clicked after that.

sryza:
> It seems like right now, this option would really only be useful once, and then it'd do a lot of extra work each time it was clicked after that.

Do you mind expanding on what you mean by this a little bit? Any time that people change the code that derives their assets, they might want to run a backfill over all the partitions.

bengotow:
Ahh sorry, I should have clarified a bit. I was thinking of a graph with a set of mapped time partitions. If I put them all in a group and view it, the UI will let me materialize all for all partitions without making an asset selection. If I come back tomorrow and just need partitions for the new day, I'm sorta stuck - I have to individually select subsets of the graph or "materialize all" again and rebuild everything.

Maybe for users with a schedule or auto-materialize policy this would be ok, since they're rarely going to try to fill in missing materializations? It's just tricky to explain that if you just want the missing day, you need to select subsets with a common root.

sryza:
I think it's fairly rare that users would want to manually materialize the latest partitions of all their assets every day. It's more likely that they'd have some sort of automated process, either via schedules or auto-materialize policies, for doing so. If there are failures in parts of the asset graph, they might select those subsets and manually materialize them.

[...]

PR #14514, opened 2023-05-29
[ui] Offer to backfill all partitions if selection has no varying root parition definitions

bengotow:
Discussion: https://github.com/dagster-io/dagster/issues/14073

This PR adds a new scenario to the "launch backfill" modal. If your assets have varying partition definitions, rather than show an error, we offer to launch a backfill with the "allPartitions" flag. A notice in the modal lets you know that if you'd like to choose a specific range of partitions you need to select a subset of the assets.

## Summary & Motivation

image

## How I Tested These Changes

Updated test coverage!


salazarm (approved):
Nice!

sryza (commented):
I think we should somehow communicate directly that the backfill will fill in all the partitions.  Maybe "Partition selection: all" or something.

bengotow:
Ahhh good call Sandy, I think that we can put an "All" in the section header (similar to how it says `Tags          0 tags`). Will update this!

Issue #14260, opened 2023-05-12
Scale up duration units in Run Gantt Chart as time elapses


braunjj:
Reported by a user in Slack [here](https://dagster.slack.com/archives/C01U5LFUZJS/p1683824881355989).

Showing run duration as "360 minutes" is not as intuitive as "6 hours"

> In Dagit's "Timed view," I find the display of large numbers to be unintuitive. It might be more user-friendly to increase the units earlier. I've provided some examples in the comments

![image](https://github.com/dagster-io/dagster/assets/2798333/048d7ecf-3e94-44c2-8bd4-9e7dcac88530)
![image](https://github.com/dagster-io/dagster/assets/2798333/c56f1ea0-b8b9-40ff-9dde-c947bd774510)

PR #14515, opened 2023-05-30
[ui] Adjust Gantt chart timescales for long runs

bengotow:
## Summary & Motivation

Loom: https://www.loom.com/share/6375e913a3a44fff864af5f43edb5cbd
Fixes: https://github.com/dagster-io/dagster/issues/14260

This PR adds a few more tick mark configurations for very long runs (it'll switch to showing "days" as tick marks now), and also fixes an issue where zooming in to a long run could show "4850s" because it chooses the unit based on the duration of each tick, not the size of the number. In these cases, we now fall back to the "1:20:50" format used elsewhere in Dagit.

image

## How I Tested These Changes

I fixed and then used the Gantt Chart storybook. It looks like our storybook upgrade broke the "prop controls" that were allowing you to fly through time on this page, and the updated animation loop was missing a reference to the  "fake time" prop that allows the storybook to render a specific moment in time. 

(I tried to use Storybook's new "argTypes" to do this, but it seems they really aren't intended for this use case and don't render on the individual story page anymore. Seems much easier to just put state inside the story.)

I also upgraded the tests to use MockedProvider rather than ApolloProvider, and fixed a bug that could cause the tick marks to get off due to rounding error and show `1:59:59` instead of `2:00:00` (unfortunately showcased in Loom video!)

### DIFF
--- a/js_modules/dagit/packages/core/src/gantt/__stories__/GanttChart.stories.tsx
+++ b/js_modules/dagit/packages/core/src/gantt/__stories__/GanttChart.stories.tsx
@@ -1,60 +1,67 @@
-import {CustomTooltipProvider} from '@dagster-io/ui';
+import {MockedProvider, MockedResponse} from '@apollo/client/testing';
+import {Box, Button, Colors, CustomTooltipProvider} from '@dagster-io/ui';
 import {Meta} from '@storybook/react';
 import React, {useState} from 'react';
 
-import {RunStatus} from '../../graphql/types';
+import {RunStatus, buildRun, buildRunGroup, buildRunStatsSnapshot} from '../../graphql/types';
 import {extractMetadataFromLogs} from '../../runs/RunMetadataProvider';
 import {RunMetadataProviderMessageFragment} from '../../runs/types/RunMetadataProvider.types';
-import {StorybookProvider} from '../../testing/StorybookProvider';
+import {WorkspaceProvider} from '../../workspace/WorkspaceContext';
 import {IGanttNode} from '../Constants';
 import {GanttChart, GanttChartLoadingState} from '../GanttChart';
+import {RUN_GROUP_PANEL_QUERY} from '../RunGroupPanel';
+import {RunGroupPanelQuery, RunGroupPanelQueryVariables} from '../types/RunGroupPanel.types';
 
 const R1_START = 1619468000;
 const R2_START = 1619468000 + 30000;
 
-const APOLLO_MOCKS = {
-  RunGroupOrError: () => ({
-    __typename: 'RunGroup',
-    rootRunId: 'r1',
-    runs: [
-      {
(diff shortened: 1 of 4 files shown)


js_modules/dagit/packages/core/src/gantt/__stories__/GanttChart.stories.tsx
```diff
@@ -1,60 +1,67 @@
...
-          id: 'r2',
-          enqueuedTime: R2_START,
-          launchTime: R2_START + 12,
-          startTime: null,
-          endTime: null,
-        },
-      },
-    ],
-  }),
+const runGroupMock: MockedResponse<RunGroupPanelQuery, RunGroupPanelQueryVariables> = {
```
salazarm, line 18: I actually added a new hook for this in internal:
https://github.com/dagster-io/internal/blob/356395a1617046d164bb434982bfaf7f64a73089/dagster-cloud/js_modules/dagit-cloud/src/testing/mocking.ts#L5

salazarm: so you could rewrite this as:

```
const runGroupMock: MockedResponse<RunGroupPanelQuery, RunGroupPanelQueryVariables> = buildQueryMock({
  query: RUN_GROUP_PANEL_QUERY,
  variables: {runId: 'r2'},
  data: {
     runGroupOrError: buildRunGroup({
        rootRunId: 'r1',
        runs: [
          buildRun({
            id: 'r1',
            runId: 'r1',
            parentRunId: null,
            status: RunStatus.FAILURE,
            stepKeysToExecute: [],
            pipelineName: 'Test',
            tags: [],
            stats: buildRunStatsSnapshot({
              id: 'r1',
              enqueuedTime: R1_START,
              launchTime: R1_START + 12,
              startTime: R1_START + 24,
              endTime: R1_START + 400,
            }),
          }),
          buildRun({
            id: 'r2',
            runId: 'r2',
            parentRunId: 'r1',
            status: RunStatus.STARTING,
            stepKeysToExecute: [],
            pipelineName: 'Test',
            tags: [],
            stats: buildRunStatsSnapshot({
              id: 'r2',
              enqueuedTime: R2_START,
              launchTime: R2_START + 12,
              startTime: null,
              endTime: null,
            }),
          }),
   }
});
```

Eliminates a little bit of boilerplate

salazarm (approved):

js_modules/dagit/packages/core/src/gantt/__stories__/GanttChart.stories.tsx
bengotow: Oh nice! It looks like this is still in a automaterialize policy page file in OSS, I'll make a follow-up PR to adopt usage of it in OSS

Issue #14408, opened 2023-05-23
"Materialize selected" button doesn't work when you're >1 level deep in the Folder view of the asset catalog


gibsondan:
### Dagster version

1.3.5

### What's the issue?

Can't materialize assets using the button on the assets catalog in certain situations, outlined below

### How to reproduce?

Navigate to assets catalog
Switch to folder view
Go 1 level deep
select an asset
you can materialize it from the menu (so its not a permissions problem) but the "Materalize selected" button stays unselected

### Deployment type

None

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

PR #14518, opened 2023-05-30
[ui] Fix “Materialize” button when >1 level deep in asset catalog

bengotow:
## Summary & Motivation

Fixes https://github.com/dagster-io/dagster/issues/14408

I cleaned up this code a bit, but the core problem was that we were selecting rows with the current path prefix added (eg: `folder/asset`), but later comparing / filtering them without the path prefix added. 

## How I Tested These Changes

I tested this by creating assets 1, 2, and 3 layers deep, and verifying that I could launch them from all three layers (eg: selecting the folder containing them vs them individually), in both the "hierarchical" view and the "flat" view.

It turns out this also repro'd in our existing `Global Catalog With Prefix` Asset Table storybook example. We don't currently have unit tests for this component, but I will write one to cover this later this week.

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/AssetTable.tsx
+++ b/js_modules/dagit/packages/core/src/assets/AssetTable.tsx
@@ -11,6 +11,7 @@ import {
   Checkbox,
   NonIdealState,
 } from '@dagster-io/ui';
+import groupBy from 'lodash/groupBy';
 import * as React from 'react';
 
 import {useUnscopedPermissions} from '../app/Permissions';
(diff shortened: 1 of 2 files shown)


bengotow:
Test failure in `[2023-05-30T03:31:50Z] FAIL src/settings/__tests__/DeploymentsRoot.test.tsx (7.88 s)` looks unrelated

salazarm (approved):
Can we add a test case for this?

salazarm (approved):

bengotow:
Test failures are unrelated, added a new test case for this behavior!

Issue #14519, opened 2023-05-30
Store Databricks stdout logs in .out file in Azure container


loerinczy:
### Dagster version

1.3.4

### What's the issue?

When materializing a step remotely on Databricks with the help of databricks_pyspark_step_launcher with dagster.yaml configured to store the compute logs in an Azure storage account container (as described here: https://docs.dagster.io/deployment/dagster-instance#compute-log-storage), the logs written to stdout on the Databricks cluster (or simply anything written to stdout with sys.stdout.write(...)) are not stored in the .out file in the storage account, however, they are shown in Dagit as captured stdout from the asset.
It is noticed, however, that the whole stdout is stored once in the .err file in Azure container.

### What did you expect to happen?

The logs written to stdout on Databricks should be stored in the .out file in the Azure container.

### How to reproduce?

* Set up logging into Azure storage as per guide: https://docs.dagster.io/deployment/dagster-instance#compute-log-storage
* Have an asset where python logging is directed into stdout (or simply call sys.stdout.write(...))
* Materialize an asset on Databricks (with the databricks_pyspark_step_launcher)

### Deployment type

Local

### Additional information

Slack conversation: https://dagster.slack.com/archives/C01U954MEER/p1684936380171939
Suggested solution in Slack: change line 273 in dagster_databricks/databricks_pyspark_step_launcher.py to sys.stdout.write(stdout). This removes the stdout log in Dagit after the execution of the asset (undesirable behavior) and has no effect on the aforementioned issue. However, adding this line after line 273 in the file solves the issue, the logs appear in the .out file in the Azure container.

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

cmpadden:
This specific stdout-to-.err behavior appears to have been fixed. Thank you.

PR #14521, opened 2023-05-30
added sys.stdout.write(stdout)

loerinczy:
## Summary & Motivation
Details are described in issue #14519.
Added call sys.stdout.write(stdout) in dagster_databricks/databricks_pyspark_step_launcher.py in order to write stdout collected from Databricks cluster into local stdout, so that it will be stored in Azure.

## How I Tested These Changes
Not yet tested, please suggest a way for proper testing.

### DIFF
--- a/python_modules/libraries/dagster-databricks/dagster_databricks/databricks_pyspark_step_launcher.py
+++ b/python_modules/libraries/dagster-databricks/dagster_databricks/databricks_pyspark_step_launcher.py
@@ -2,6 +2,7 @@ import gzip
 import io
 import os.path
 import pickle
+import sys
 import tempfile
 import time
 import zlib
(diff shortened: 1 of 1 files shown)


loerinczy:
I have tested this sys.stderr.write(stderr) scenario, and it actually leads to the duplication of logs in stderr, so the logs are already in stderr. Thus, I thought having this function call there would be redundant. Anyways, I added the line.

OwenKephart:
@loerinczy gotcha, makes sense -- the system logs from executing in Databricks will be emitted to stderr, so it makes sense that they'll show up there. However, it's possible that additional log messages (not directly from Dagster) may be emitted, so I think it's fine to have some level of duplication. Not a huge deal either way.

Thanks for the contribution! I'll merge this in for next week's release.

OwenKephart (approved):
Thanks again!

PR #14527, opened 2023-05-30
[fix][dagster-dbt] Fix initialization of DBT cloud Pythonic resource when loading from instance

benpankow:
## Summary

Fixes an unfortunate invocation of `with_resource_context` (which attaches a resource init context to a Pythonic resource) instead of `from_resource_context` (which initializes the resource using that context). This meant that, when using the Pythonic DBT Cloud resource to load assets from the DBT Cloud instance, we weren't fully resolving the config. This meant that things like env vars were not getting resolved properly.

Because of the changes in #14306, Pythonic resources now expect fully processed config when initialized. Accordingly, introduces a new helper method `_process_config_and_initialize` which processes the Pythonic resource's config (resolving env vars, etc) and then initializes it.

## Test Plan

Unit test using env vars and `load_assets_from_dbt_cloud_job`

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/asset_defs.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cloud/asset_defs.py
@@ -75,10 +75,9 @@ class DbtCloudCacheableAssetsDefinition(CacheableAssetsDefinition):
             if isinstance(dbt_cloud_resource_def, DbtCloudClientResource)
             else dbt_cloud_resource_def
         )
+
         self._dbt_cloud: DbtCloudClient = (
-            dbt_cloud_resource_def.from_resource_context(
-                build_init_resource_context()
-            ).get_dbt_client()  # type: ignore
+            dbt_cloud_resource_def._process_config_and_initialize().get_dbt_client()  # noqa: SLF001
             if isinstance(dbt_cloud_resource_def, DbtCloudClientResource)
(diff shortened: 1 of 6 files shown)


rexledesma (changes requested):
Could we validate the change with a test, which uses `monkeypatch` + `EnvVar`?

python_modules/libraries/dagster-dbt/dagster_dbt/cloud/asset_defs.py
```diff
@@ -75,10 +75,9 @@ def __init__(
             if isinstance(dbt_cloud_resource_def, DbtCloudClientResource)
             else dbt_cloud_resource_def
         )
+
         self._dbt_cloud: DbtCloudClient = (
-            dbt_cloud_resource_def.with_resource_context(
-                build_init_resource_context()
-            ).get_dbt_client()  # type: ignore
+            dbt_cloud_resource_def._process_config_and_initialize().get_dbt_client()  # noqa: SLF001
```
rexledesma, line 80: I am surprised that the user would need to use a private method here to get the underlying object from a resource  definition.

1. Can this method be public?
2. Looks like the `.with_resource_context` and `.from_resource_context` methods are confusing right now, given that we've made this mistake. Could we potentially consolidate them or rename them?

rexledesma (approved):
Great. Do we still need the `noqa:` comments?

PR #14534, opened 2023-05-30
Expose 'raw instance config' as a helm chart field

gibsondan:
Summary:
Right now every time we add a new instance config field, we require it to also be added to the helm chart schema. This causes two problems:
- It's a pain for Dagster engineers to add every feature in the Helm chart. It's easy to forget fields, and users invariably want to use them. (for example, 'auto_materialize', 'local_artifact_storage', and 'code_servers' could all plausibly be asked for by users (the last two have) and we have to scramble to get out a helm PR).
- The docs are no longer useful if you are using the helm chart, since the helm chart is in camelCase and doesn't include all fields.

It makes sense to stub in things that are actually determined by other properties of the helm chart, like storage and potentially the run launcher - but for the long tail of instance configuration fields that we offer, I think this translation layer does more harm than good. (THe main good that we're losing is the Helm-level schema validation - if you enter a key that doesn't exist here, it will fail at deploy time rather than violate a schema)

This PR addresses this problem by adding a new field that stubs in raw instance yaml and is included in the dagster.yaml.

### DIFF
--- a/helm/dagster/schema/schema/charts/dagster/values.py
+++ b/helm/dagster/schema/schema/charts/dagster/values.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import Any, List, Mapping, Optional
 
 from pydantic import BaseModel, Field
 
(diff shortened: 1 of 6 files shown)


rexledesma (approved):
Makes sense

prha (approved):

Issue #14002, opened 2023-04-29
Document potential consequences of turning on auto-materialization with lots of assets


sryza:
Maybe something like "We’ve tested this at scale but still recommend introducing incrementally, because XYZ. Here's what could happen if you do this."

PR #14536, opened 2023-05-31
in auto-materialize docs, warn about starting with too many assets at…

sryza:
… once

## Summary & Motivation

Fixes https://github.com/dagster-io/dagster/issues/14002


erinkcochran87 (commented):
Visually this is really dense, especially for a callout:

Screen Shot 2023-06-01 at 3 01 29 PM

Ideally callouts should be concise and start with a (bolded) short summary of whatever the point the callout is intending to make. In this case, that could be something like:

> **New to auto-materialization? Start small.** If it's your first time using auto-materialize...

Instead of a callout, what do you think about making this the first paragraph in the **Turning on auto-materializing** section? This is ideally something the user would need top know before they start, but here it's following instructions for telling them how to turn it on.

sryza:
@erinkcochran87 yeah you're right that it doesn't look great as a call-out.  The warning only applies to users with large numbers of partitioned assets (I updated the text to clarify this). I don't love putting it at the front, because it distracts from the information that's relevant to all users.

I tried taking it out of the call-out, leaving it at the end, and adding a short bold phrase at the beginning the draw attention to us.  Let me know what you think.

erinkcochran87 (approved):
I dig this - it's a good compromise between the callout styling and the callout visually being a bit much. :shipit: !

Issue #14542, opened 2023-05-31
Update to Census Management API results in buggy ops that run syncs


ldnicolasmay:
### Dagster version

1.2.7

### What's the issue?

The Census integration no longer works as expected because Census recently changed its "Management API".

[Here](https://docs.getcensus.com/basics/developers/api/sync-runs#get-sync_runs-id) are the docs where you can see that, in the response `data.status` field, there's a new possible value `"preparing"`.

### What did you expect to happen?

When a Census sync is run, Dagster should poll throughout the `"preparing"` _**and**_ `"working"` status.

However, this new value `"preparing"` is now always returned before the `"working"` value when the API is called. So, the Dagster integration code currently treats it as a finished status and breaks the `while` loop [here](https://github.com/dagster-io/dagster/blob/master/python_modules/libraries/dagster-census/dagster_census/resources.py#L151-L176) in the `CensusResource` class's `poll_sync_run` method.

### How to reproduce?

Run any op that starts a valid Census sync run.

### Deployment type

Docker Compose

### Deployment details

N/A

### Additional information

N/A

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

clairelin135:
Responded on the PR, thanks for adding a fix!

PR #13315, opened 2023-04-01
[fix][dagster-census] CensusResource poll_sync_run method bug

ldnicolasmay:
## Summary & Motivation

There is a simple bug in the `CensusResource` class's `poll_sync_run` method.

The `response_dict` that's returned by the `self.get_sync_run(sync_run_id)` on [L153](https://github.com/dagster-io/dagster/blob/master/python_modules/libraries/dagster-census/dagster_census/resources.py#L153) is structured as detailed in the Census API docs [here](https://docs.getcensus.com/basics/developers/api/sync-runs#get-sync_runs-id).

On [L162](https://github.com/dagster-io/dagster/blob/master/python_modules/libraries/dagster-census/dagster_census/resources.py#L162), the `poll_sync_run` method gets the `status` of the API request with `response_dict["status"]`. This isn't what's intended. What's intended is the `data.status` of the Census sync run. This bug results in the method waiting 10 seconds and then finding the API request status as `"success"` or `"failure"`, never `"working"`, which is required for the while-loop to continue polling the sync run, as seen on [LL162-166](https://github.com/ldnicolasmay/dagster/blob/master/python_modules/libraries/dagster-census/dagster_census/resources.py#L162-L166).

In order to get the `status` of the Census sync run, the method needs to check `response_dict["data"]["status"]`.

## How I Tested These Changes

I ran `pytest .` from the `python_modules/libraries/dagster-census` module, and all test succeeded:
```
======================================== test session starts =========================================
platform linux -- Python 3.10.6, pytest-7.0.1, pluggy-1.0.0
rootdir: /home/hynso/GitHub/ldnicolasmay/dagster, configfile: pyproject.toml
plugins: anyio-3.6.2, forked-1.6.0, cov-2.10.1, rerunfailures-10.0, buildkite-test-collector-0.1.7, mock-3.3.1, dependency-0.5.1, requests-mock-1.10.0, xdist-2.1.0, snapshottest-0.6.0
collected 10 items                                                                                   

dagster_census_tests/test_op.py .                                                              [ 10%]
dagster_census_tests/test_resources.py ........                                                [ 90%]
dagster_census_tests/test_version.py .                                                         [100%]

========================================= 10 passed in 2.39s =========================================
```

[...]

PR #14543, opened 2023-05-31
[fix][dagster-census] CensusResource poll_sync_run new 'preparing' st…

ldnicolasmay:
…atus

## Summary & Motivation

Fixes [this issue](https://github.com/dagster-io/dagster/issues/14542) where any op that runs a Census sync prematurely breaks the `while` loop in the `CensusResource` class's `poll_sync_run` method.

## How I Tested These Changes

I have the Census integration in prod work pipelines and have been using a locally patched package with this small PR change. With this change, the `poll_sync_run` method behaves as expected, meaning it successfully polls a Census sync during both its `"preparing"` and `"working"` statuses until the sync is finished.

### DIFF
--- a/python_modules/libraries/dagster-census/dagster_census/resources.py
+++ b/python_modules/libraries/dagster-census/dagster_census/resources.py
@@ -159,7 +159,7 @@ class CensusResource:
 
             sync_id = response_dict["data"]["sync_id"]
 
-            if response_dict["data"]["status"] == "working":
+            if response_dict["data"]["status"] in {"preparing", "working"}:
                 self._log.debug(
                     f"Sync {sync_id} still running after {datetime.datetime.now() - poll_start}."
                 )


clairelin135:
@ldnicolasmay thanks for adding a fix for this issue! This fix looks great. I was wondering how difficult it would be to add a test for when the state is 'preparing'? Not familiar with census, but it is surprising that none of the existing tests caught this issue when the preparing state was added.

ldnicolasmay:
Thanks for the feedback @clairelin135 ! The question of better testing came up in [a previous, related PR I opened](https://github.com/dagster-io/dagster/pull/13315), but ultimately [Sean Mackesey suggested](https://github.com/dagster-io/dagster/pull/13315#issuecomment-1495986386) that "figuring out how to make our tests better [for this integration] is a larger project". 

That's why I didn't touch testing in this PR. It seems it's not robust enough to make a small improvement, then overhaul later. Also, as you can see from the previous PR convo, I'm not really sure how to improve the testing for this particular method. I _am_ happy to get feedback here though and put in more work.

Thx again!

clairelin135:
@ldnicolasmay that makes sense. I'm happy to land this PR as is--I agree with Sean that improving testing for this package is a larger project.

clairelin135 (approved):

PR #14551, opened 2023-05-31
Store asset daemon evaluations after cursor

johannkm:
I'm trying to figure out causes of https://elementl-workspace.slack.com/archives/C04UD72HHQX/p1685543991696809. One potential cause is that if the daemon crashes before it stores the cursor, then it will keep trying the previous evaluation_id.

I'm suspicious that this isn't the only cause bc it seems unlikely that multiple org's daemons would crash there, but good to squash it anyway

### DIFF
--- a/python_modules/dagster/dagster/_daemon/asset_daemon.py
+++ b/python_modules/dagster/dagster/_daemon/asset_daemon.py
@@ -89,11 +89,6 @@ class AssetDaemon(IntervalDaemon):
             run_tags=None,
         )
 
-        if schedule_storage.supports_auto_materialize_asset_evaluations:
-            schedule_storage.add_auto_materialize_asset_evaluations(
-                check.not_none(new_cursor.evaluation_id), evaluations
-            )
-
         for run_request in run_requests:
             yield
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster/_daemon/asset_daemon.py
```diff
@@ -157,3 +152,10 @@ def run_iteration(
             instance.submit_run(run.run_id, workspace)
 
         instance.daemon_cursor_storage.set_cursor_values({CURSOR_KEY: new_cursor.serialize()})
+
+        # We enforce uniqueness per (asset key, evaluation id). Store the evaluations after the cursor,
+        # so that if the daemon crashes and doesn't update the cursor we don't try to write duplicates.
+        if schedule_storage.supports_auto_materialize_asset_evaluations:
+            schedule_storage.add_auto_materialize_asset_evaluations(
+                check.not_none(new_cursor.evaluation_id), evaluations
```
OwenKephart, line 160: nit: don't need the check.not_none anymore

PR #14554, opened 2023-05-31
Use a different task definition family for each job in a code location

gibsondan:
Summary:
Since we set a different env var for each job, the current behavior causes lots of thrash if you are launching multiple jobs in the same code location in sequence. Use a separate one for each job instead.

Test Plan: BK

### DIFF
--- a/python_modules/libraries/dagster-aws/dagster_aws/ecs/utils.py
+++ b/python_modules/libraries/dagster-aws/dagster_aws/ecs/utils.py
@@ -1,5 +1,8 @@
+import hashlib
 import re
 from typing import Any, Mapping
+
+from dagster._core.host_representation.origin import ExternalJobOrigin
 
 from .tasks import DagsterEcsTaskDefinitionConfig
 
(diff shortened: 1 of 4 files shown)


python_modules/libraries/dagster-aws/dagster_aws/ecs/utils.py
```diff
@@ -9,6 +12,34 @@ def sanitize_family(family):
...
+    m.update(name.encode("utf-8"))
+    name_hash = m.hexdigest()[:8]
+    return f"{name[:55]}_{name_hash}"
+
+
+def get_task_definition_family(
+    prefix: str,
+    job_origin: ExternalJobOrigin,
+) -> str:
+    job_name = job_origin.job_name
```
jmsanders, line 26: Kind of pathological, but in theory, if you have two different dagster deployments (let's say you run a dev and a prod) and they have matching job names, repo names, and location names, I think they'd still have collisions here.

Probably an edge case not worth worrying about for now since this is strictly better than what we're coming from. But just a heads up.

jmsanders (approved):
Nice

python_modules/libraries/dagster-aws/dagster_aws/ecs/utils.py
gibsondan: true but also an issue before right? we'd need some kind of way to namespace the deployment to be able to get around that (in cloud we include the deployment name in this calculation for this exact reason)

andy-hdai: not sure if this will be true for most people hosting their own dagster clusters on ECS, but we have dev/qa/prod clusters separated into different AWS accounts, so there wouldn't be any collisions.

PR #14555, opened 2023-05-31
makes the schedule execution independent from tick history

prha:
## Summary & Motivation
With the monthly schedules and a 30-day purge window, we can run into an issue where schedules will repeatedly try to kick off duplicate runs because they are trying to read from the tick history.

This diff adds the last tick into the schedule state, to prevent from this dependency on the tick history being maintained.

(keeping a sensible tick history for observability on these sparse schedules is a separate issue to be tackled in a different PR).

## How I Tested These Changes
BK

### DIFF
--- a/python_modules/dagster/dagster/_core/scheduler/instigation.py
+++ b/python_modules/dagster/dagster/_core/scheduler/instigation.py
@@ -36,6 +36,14 @@ class InstigatorStatus(Enum):
     STOPPED = "STOPPED"
 
 
+@whitelist_for_serdes(old_storage_names={"JobTickStatus"})
+class TickStatus(Enum):
+    STARTED = "STARTED"
+    SKIPPED = "SKIPPED"
+    SUCCESS = "SUCCESS"
+    FAILURE = "FAILURE"
+
+
 @whitelist_for_serdes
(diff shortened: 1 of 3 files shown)


prha:
* master
  * **PR #14555** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14555" target="_blank">Graphite</a>  👈
    * **PR #14561** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14561" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14555?utm_source=stack-comment).

alangenfeld (commented):
i think this is fine - would like @gibsondan to view as well

gibsondan (commented):
"(keeping a sensible tick history for observability on these sparse schedules is a separate issue to be tackled in a different PR)."

If the solution (which as i understand it would be to always keep the last N ticks) fixes both problems, are we sure it's a separate issue? Wouldn't that fix this problem too with no additional changes required?

gibsondan (changes requested):
to you for that question - just making sure we actually need this if we think we are likely to do the other thing

Issue #14271, opened 2023-05-12
UI not displaying custom skip_message of SkipReason when sensor has both yield and return


ruizh22:
### Dagster version

1.3.2

### What's the issue?

UI is displaying a generic message "Sensor function returned an empty result" when the code actually specified a custom skip_message like `SkipReason(skip_message='No assets to backfill')`.

![image](https://github.com/dagster-io/dagster/assets/109117963/5a6a8cbc-3e2e-44c9-8692-d23520c266f8)
![image](https://github.com/dagster-io/dagster/assets/109117963/68a73f39-e178-48c0-b457-3c620eaf85c7)

### What did you expect to happen?

UI to display the custom skip message when hovering over the "Skipped" status and also reflect the same message on the "test sensor" result page.

### How to reproduce?

Define a sensor to return a SkipReason with custom skip_message, the custom skip_message will not show up on UI or anywhere in the log. 

### Deployment type

None

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

sryza:
Hey @ruizh22 - do you have a code example that reproduces this?

I just gave this a shot with this code example

```python
from dagster import sensor, SkipReason, Definitions


@sensor()
def sensor1():
    return SkipReason("hello")


defs = Definitions(sensors=[sensor1])
```

and got the expected behavior:

<img width="106" alt="image" src="https://github.com/dagster-io/dagster/assets/654855/a6033460-b547-401d-b470-fc8c31a79c6f">

-

<img width="220" alt="image" src="https://github.com/dagster-io/dagster/assets/654855/d946eb94-1ca7-4f5e-b558-4d3c656199e2">

ruizh22:
hi @sryza , we are doing something like this:

```
@sensor(name='a_sensor', job_name='a_job')
    def code_versions_sensor(context: SensorEvaluationContext) -> Union[RunRequest, SkipReason]:
        ...
        assets_to_backfill = {...}
        if len(backfill_asset_versions) == 0:
            skip_message: str = 'No assets to backfill'
            context.log.info(skip_message)
            return SkipReason(skip_message=skip_message)
        yield backfill_run_request(assets_to_backfill)
        context.update_cursor(...)
```

sryza:
Can you verify that you're making it to the inside of that `if` statement? E.g. by raising an exception there and seeing if it shows up? Or writing out to a file that you can inspect?

ruizh22:
[...]

PR #14558, opened 2023-05-31
capture return value when sensor also has yield

sryza:
## Summary & Motivation

If a sensor has both a `yield` and a `return` statement, we were ignoring the returned value. Now, we no longer ignore it.

Generators hurt my brain.

Fixes: https://github.com/dagster-io/dagster/issues/14271

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/sensor_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/sensor_definition.py
@@ -378,7 +378,7 @@ RawSensorEvaluationFunctionReturn = Union[
 ]
 RawSensorEvaluationFunction: TypeAlias = Callable[..., RawSensorEvaluationFunctionReturn]
 
-SensorEvaluationFunction: TypeAlias = Callable[..., Iterator[Union[SkipReason, RunRequest]]]
+SensorEvaluationFunction: TypeAlias = Callable[..., Sequence[Union[SkipReason, RunRequest]]]
 
 
 def get_context_param_name(fn: Callable) -> Optional[str]:
@@ -486,7 +486,7 @@ class SensorDefinition(IHasInternalInit):
         """
         return SensorDefinition.dagster_internal_init(
             name=self.name,
-            evaluation_fn=self._evaluation_fn,
+            evaluation_fn=self._raw_fn,
             minimum_interval_seconds=self.minimum_interval_seconds,
             description=self.description,
             job_name=None,  # if original init was passed job name, was resolved to a job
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -978,20 +978,24 @@ def _wrapped_fn(context: SensorEvaluationContext):
...
+                )
+
+        if inspect.isgenerator(result) or isinstance(result, list):
+            result_value = yield from result
+            # we reach this case if the function has both yield statements and a return statement
+            # that returns a value
+            if result_value is not None:
+                yield check_returned_scalar(result_value)
+        else:
+            yield check_returned_scalar(result)
```
gibsondan, line 998: this is a little odd.. so we're basically transforming the generator into a different generator that yields the return value along with the other things that are yielded instead of returning it? (which would normally get raised as a StopIteration as I understand it - which a yield from can accept if it wants to, like we are doing here).

Maybe it would be slightly less weird if this SensorEvaluationFunction type was a boring regular function that returned a list instead of still cosplaying as a generator? It seems like it's only callsite calls list() on it right away anyways. Something about that seems slightly less odd than making it still a generator with subtly different semantics

gibsondan (changes requested):
a not strongly held idea to make this slightly less weird inline, without changing the user-facing piece

gibsondan (approved):
Hm, my suggestion might not be feasible though since there doesn't seem to be a great way to pull the return value out of a generator if you aren't yourself a generator (other than catching a StopIteration in a while loop or something, gross). So it's probably fine... this wrapped generator function is internal to the class anyway

prha:
Yeah, I think this is also fine... but I think we should strongly consider changing our documentation as much as possible to just use `return` instead of `yield`.

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
sryza: I implemented this change. Mind taking another look?

python_modules/dagster/dagster/_core/definitions/sensor_definition.py
```diff
@@ -978,20 +978,30 @@ def _wrapped_fn(context: SensorEvaluationContext):
...
-            )
+                )
+
+        if inspect.isgenerator(result):
+            return_value = []
+            try:
+                while True:
+                    return_value.append(next(result))
+            except StopIteration as e:
+                if e.value is not None:
```
gibsondan, line 997: this is maybe worth a comment for future non-generator experts

gibsondan (approved):
ah ok this works too! Thx for taking my suggestion after all :)

Issue #13597, opened 2023-04-13
Dagster scheduler sometimes re-runs ticks when cron schedule changes


gibsondan:
### Dagster version

1.2.6

### What's the issue?

Example feedback : "I had a flow scheduled to run at 1:15am. It is working fine. Due to the timing of an external dependency, I needed to change the schedule to 2:15am. So, all I did was cron_schedule="15 2 * * 1-5" in the schedule definition (changing the 1 to to a 2). When I pushed this change, the flow automatically started to run again, even though it just ran successfully a few hours ago

### What did you expect to happen?

To recognize that the cronstring has changed and reset the timer rather than going back and backfilling the previous day

### Deployment type

None

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

PR #14561, opened 2023-06-01
handle cron schedule changes

prha:
## Summary & Motivation
https://github.com/dagster-io/dagster/issues/13597

## How I Tested These Changes
BK

### DIFF
--- a/python_modules/dagster/dagster/_scheduler/scheduler.py
+++ b/python_modules/dagster/dagster/_scheduler/scheduler.py
@@ -52,7 +52,7 @@ if TYPE_CHECKING:
 # how often do we update the job row in the database with the last iteration timestamp.  This
 # creates a checkpoint so that if the cron schedule changes, we don't try to backfill schedule ticks
 # from the start of the schedule, just since the last recorded iteration interval.
-LAST_RECORDED_ITERATION_INTERVAL = 3600
+LAST_RECORDED_ITERATION_INTERVAL_SECONDS = 3600
 
 
 class _ScheduleLaunchContext:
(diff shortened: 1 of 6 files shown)


prha:
* master
  * **PR #14555** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14555" target="_blank">Graphite</a> 
    * **PR #14561** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14561" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14561?utm_source=stack-comment).

gibsondan (changes requested):
see inline perf question

python_modules/dagster/dagster/_scheduler/scheduler.py
```diff
@@ -49,6 +49,12 @@
     from dagster._daemon.daemon import DaemonIterator
 
 
+# how often do we update the job row in the database with the last iteration timestamp.  This
+# creates a checkpoint so that if the cron schedule changes, we don't try to backfill schedule ticks
+# from the start of the schedule, just since the last recorded iteration interval.
+LAST_RECORDED_ITERATION_INTERVAL = 3600
```
alangenfeld, line 55: what unit is this / how long is this (comment or name)

python_modules/dagster/dagster/_scheduler/scheduler.py
```diff
@@ -483,6 +490,24 @@ def launch_scheduled_runs_for_schedule_iterator(
     if not tick_times:
         if log_verbose_checks:
             logger.info(f"No new tick times to evaluate for {schedule_name}")
+
+        # insert last considered timestamp here
```
alangenfeld, line 494: why is this in the `if not tick_times` condition?

python_modules/dagster/dagster/_scheduler/scheduler.py
```diff
@@ -483,6 +490,24 @@ def launch_scheduled_runs_for_schedule_iterator(
...
+        if (
+            not instigator_data.last_iteration_timestamp
+            or instigator_data.last_iteration_timestamp + LAST_RECORDED_ITERATION_INTERVAL
```
alangenfeld, line 497: seems like theres an assumption being taken here 

instead of keeping this totally up to date, we keep it up to date within the tolerance of `LAST_RECORDED_ITERATION_INTERVAL` which we assume any deletes will be greater than that time

we could consider making it invalid to set the bit of instance config that cleans up ticks to less than this value (or some relative math) 

short of that we should leave some comments explaining whatsup

Does this window have an implication on schedules changes that are faster than this ?

prha: > Does this window have an implication on schedules changes that are faster than this ?
Let's say I have a schedule that fires on Sundays at 12:00 AM.  On Wednesday at 2:00 AM, I change this schedule to fire on Wednesday at 1:30 AM and deploy it, we may kick off a run for the 1:30 AM tick.  It's somewhat non-deterministic because it depends on the logged iteration of the scheduler, but also it's not that crazy unexpected to see a run for the previous tick if it lands within the hour.  

This is a consideration independent of tick deletions, since we can have the above scenario, without any ticks logged (let's say that I turned the schedule on Monday morning).

prha: Re: tick deletion, initially had this stacked on https://github.com/dagster-io/dagster/pull/14555, but decoupled these two so that we could merge this fix and solve the tick deletion issue a different way.

alangenfeld (approved):
alright alright alright

PR #14565, opened 2023-06-01
[ui] Handle network error on repository location reload

hellendag:
## Summary & Motivation

When reloading a code location, a network error on the mutation will be ignored by the UI and the button remains disabled and spinning.

Resolve this by catching the rejected promise on the `client.mutate` call, which is how the network error is surfaced in direct Apollo API access like this. (We use `client.mutate` here to avoid needing to use a React hook to set up the mutation.)

Currently, the rejected promise just falls through and the UI is none the wiser. By catching the rejection and returning an error payload to the caller, the UI can handle it and update state accordingly.

## How I Tested These Changes

Add an error to the reload mutation in `schema/roots/mutation.py`, try to reload the code location. Verify that an error dialog appears with information about the 500 error.

Repeat with the error removed, verify that reloading the code location is successful.

Screenshot 2023-06-01 at 10 18 47 AM

### DIFF
--- a/js_modules/dagit/packages/core/src/nav/useRepositoryLocationReload.tsx
+++ b/js_modules/dagit/packages/core/src/nav/useRepositoryLocationReload.tsx
@@ -1,4 +1,4 @@
-import {ApolloClient, gql, useApolloClient, useQuery} from '@apollo/client';
+import {ApolloClient, ApolloError, gql, useApolloClient, useQuery} from '@apollo/client';
 // eslint-disable-next-line no-restricted-imports
 import {Intent} from '@blueprintjs/core';
 import * as React from 'react';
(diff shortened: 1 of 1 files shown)


hellendag:
* master
  * **PR #14565** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14565" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14565?utm_source=stack-comment).

bengotow (approved):
This looks good to me! I always forget that GraphQL requests can throw network errors like this (beyond returning GraphQL error shapes). At some point we may need to audit some of our important mutations like launchBackfill and try to handle this case there :-/

gibsondan:
this PR is great, and I think our standard pattern of returning a PythonError for errors that aren't "expected" user code errors is a bit odd - should this be more of the default pattern?

hellendag:
> this PR is great, and I think our standard pattern of returning a PythonError for errors that aren't "expected" user code errors is a bit odd - should this be more of the default pattern?

Hm, we could make more use of the `error` field on `useQuery`/`useMutation` than we currently do.

I think there's value currently in having fields be able to be `PythonError` and not fail the entire response, so that we can use the successful data supplied in the response. As long as we appropriately handle those error cases (which TypeScript should somewhat force us to do) then I think we're okay with what we have. But I don't think I feel strongly about that, and am open to discussion about it.

PR #14570, opened 2023-06-01
Fix rendering MetadataEntryTable when their are `nulls` in the table.

salazarm:
## Summary & Motivation

We're trying to call `toString` on `null`, use optional chaining to call `toString` instead to avoid calling it on `null`.

## How I Tested These Changes

👀


bengotow (approved):
Looks good to me! May not be worth it now, but we do have storybooks for all of these metadata entry types -- maybe we should throw a `null` value in the example to make sure we don't break this in the future

PR #14576, opened 2023-06-01
Asset job metadata

Elliot2718:
## Summary & Motivation
Job metadata can be added when defining jobs using the `@asset` decorator ([docs](https://docs.dagster.io/concepts/ops-jobs-graphs/metadata-tags#specifying-metadata)). However, they cannot currently be added when using the `define_asset_job` function. I would like to be to do this, so we can specify job owners, links to external documentation, and other metadata for asset jobs.

This would resolve issue #10639.
## How I Tested These Changes
I added two metadata fields to the `basic_assets_job` in the `basic_assets_repository` and verified that the metadata section is populated with these fields:
image

## Caveats
- I was not able to get pyright running, so I would appreciate if someone could run type checking on this.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/unresolved_asset_job_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/unresolved_asset_job_definition.py
@@ -10,7 +10,7 @@ from dagster._utils.backcompat import deprecation_warning
 
 from .asset_layer import build_asset_selection_job
 from .config import ConfigMapping
-from .metadata import MetadataValue
+from .metadata import RawMetadataValue
 
 if TYPE_CHECKING:
     from dagster._core.definitions import (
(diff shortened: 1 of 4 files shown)


python_modules/dagster/dagster/_core/definitions/unresolved_asset_job_definition.py
```diff
@@ -289,6 +295,8 @@ def define_asset_job(
             Values that are not strings will be json encoded and must meet the criteria that
             `json.loads(json.dumps(value)) == value`.  These tag values may be overwritten by tag
             values provided at invocation time.
+        metadata (Optional[Mapping[str, RawMetadataValue]]):
```
clairelin135, line 298: nit: completing the docstring here would be nice. Copy pasted from elsewhere in the codebase:
```
Arbitrary metadata about the job.  Keys are displayed string labels, and values are one of the following: string, float, int, JSON-serializable dict, JSON-serializable list, and one of the data classes returned by a MetadataValue static method.
```

clairelin135 (commented):
This looks good to me! Kicking off tests for it now

python_modules/dagster/dagster/_core/definitions/unresolved_asset_job_definition.py
clairelin135: I think you need to push up a change that adds this snippet, but aside from that this PR is good to go! :)

Elliot2718: Pushed the change with the docstring. Thanks!

clairelin135 (approved):
Looks great, thanks for adding!

PR #14592, opened 2023-06-02
Include package data in dagster-graphql setup.py

Tanguy-LeFloch:
## Summary & Motivation

Currently the `py.typed` is not included in the installed package of `dagster_graphql`. To do so I added the `include_package_data=True` in the `setup.py`.

N.B: In many other libraries the `include_package_data=True` is not included, but it seems to work because the `MANIFEST.in` itself is fine to pick up non-empty files, and many of the `py.typed` are actually filled with `partial` inside. It is not the case for the `py.typed` within `dagster_graphql` which is empty and thus not copied to the installation location.
I'm fine doing it like that also but I'm not sure if adding the "partial" had a meaning for the dagster team.

## How I Tested These Changes
- Locally by performing an install and checking that the file is there. Latest `dagster_graphql` and `pip` versions

### DIFF
--- a/python_modules/dagster-graphql/setup.py
+++ b/python_modules/dagster-graphql/setup.py
@@ -33,6 +33,7 @@ setup(
         "Operating System :: OS Independent",
     ],
     packages=find_packages(exclude=["dagster_graphql_tests*"]),
+    include_package_data=True,
     install_requires=[
         f"dagster{pin}",
         "graphene>=3",


clairelin135:
Hi @Tanguy-LeFloch. Taking a look--if tests pass I'll go ahead and merge this.

Tanguy-LeFloch:
Hey @clairelin135 , thanks.
It seems to fail unfortunately. Is there any way I could see what happened?

I'm sorry if I missed it but couldn't find anything in the contribution doc about buildkite.

clairelin135:
@Tanguy-LeFloch Ah, our buildkite is currently only viewable internally. But it seems to be an unrelated failure with the wandb integration. I will go ahead and merge this

clairelin135 (approved):

PR #14603, opened 2023-06-02
Add a way to install dagster with no grpcio pin

gibsondan:
Summary:
We're still seeing enough flakiness with newer versions of grpcio that we can't confidently remove the grpcio pin for everybody as the default experience - but for folks who need to use dagster with dependencies that are pinned to a higher version, this PR provides an escape hatch.

Test Plan:
pip install dagster with and without this env var set

### DIFF
--- a/docs/content/getting-started/install.mdx
+++ b/docs/content/getting-started/install.mdx
@@ -40,6 +40,14 @@ pip install dagster dagit
 pip install dagster dagit --find-links=https://github.com/dagster-io/build-grpcio/wiki/Wheels
 ```
 
+**Need to install Dagster using a more recent version of `grpcio`**? On Python 3.7-3.10, `dagster` currently has a `grpcio<1.48` pin due to reports of sporadic crashes and hangs with more recent versions. If you see a version incompatibility error when installing `dagster` due to this pin, you can set the DAGSTER_NO_GRPCIO_PIN environment variable when installing to allow more recent versions of `grpcio` to be installed.
+
+For example:
+
+```bash
+DAGSTER_NO_GRPCIO_PIN=1 pip install dagster dagit
+```
+
 ---
 
 ## Installing Dagster from source
@@ -61,6 +69,14 @@ poetry add dagster dagit
 ```bash
 poetry source add grpcio https://github.com/dagster-io/build-grpcio/wiki/Wheels
 poetry add dagster dagit
+```
+
+**Need to install Dagster using a more recent version of `grpcio`**? On Python 3.7-3.10, `dagster` currently has a `grpcio<1.48` pin due to reports of sporadic crashes and hangs with more recent versions. If you see a version incompatibility error when installing `dagster` due to this pin, you can set the DAGSTER_NO_GRPCIO_PIN environment variable when installing to allow more recent versions of `grpcio` to be installed.
+
+For example:
+
+```bash
+DAGSTER_NO_GRPCIO_PIN=1 poetry add dagster dagit
 ```
 
 ---
--- a/python_modules/dagster/setup.py
+++ b/python_modules/dagster/setup.py
@@ -1,3 +1,4 @@
+import os
 from pathlib import Path
 from typing import Dict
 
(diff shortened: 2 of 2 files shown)


gibsondan:
I think I could probably also be convinced to take out the pin entirely (or make it higher, like <1.55), verify that it works, but leave us pinned to <1.48 in CI until we get at least one report of the issues that we see in CI from a user... not clear that they're actual issues vs. what happens if you create hundreds of instances and grpc server at once in a CI setting

docs/content/getting-started/install.mdx
```diff
@@ -40,6 +40,14 @@ pip install dagster dagit
 pip install dagster dagit --find-links=https://github.com/dagster-io/build-grpcio/wiki/Wheels
 ```
 
+**Need to install Dagster using a more recent version of `grpcio`**? On Python 3.7-3.10, `dagster` currently has a `grpcio<1.48` pin due to reports of sporadic crashes and hangs with more recent versions. If you see a version incompatibility error when installing `dagster` due to this pin, you can set the DAGSTER_NO_GRPCIO_PIN environment variable when installing to allow more recent versions of `grpcio` to be installed.
```
erinkcochran87, line 43:
```suggestion
**Need to install Dagster using a more recent version of `grpcio`**? On Python 3.7-3.10, `dagster` currently has a `grpcio<1.48` pin due to reports of sporadic crashes and hangs with more recent versions. If you see a version incompatibility error when installing `dagster` due to this pin, you can set the `DAGSTER_NO_GRPCIO_PIN` environment variable when installing to allow more recent versions of `grpcio` to be installed.
```

docs/content/getting-started/install.mdx
```diff
@@ -63,6 +71,14 @@ poetry source add grpcio https://github.com/dagster-io/build-grpcio/wiki/Wheels
 poetry add dagster dagit
 ```
 
+**Need to install Dagster using a more recent version of `grpcio`**? On Python 3.7-3.10, `dagster` currently has a `grpcio<1.48` pin due to reports of sporadic crashes and hangs with more recent versions. If you see a version incompatibility error when installing `dagster` due to this pin, you can set the DAGSTER_NO_GRPCIO_PIN environment variable when installing to allow more recent versions of `grpcio` to be installed.
```
erinkcochran87, line 74:
```suggestion
**Need to install Dagster using a more recent version of `grpcio`**? On Python 3.7-3.10, `dagster` currently has a `grpcio<1.48` pin due to reports of sporadic crashes and hangs with more recent versions. If you see a version incompatibility error when installing `dagster` due to this pin, you can set the `DAGSTER_NO_GRPCIO_PIN` environment variable when installing to allow more recent versions of `grpcio` to be installed.
```

erinkcochran87 (approved):
Content-wise I'm good with this. I left two comments about small formatting stuff, but otherwise lgtm

python_modules/dagster/setup.py
```diff
@@ -81,10 +98,9 @@ def get_version() -> str:
         # pin around issues in specific versions of alembic that broke our migrations
         "alembic>=1.2.1,!=1.6.3,!=1.7.0,<1.11.0",
         "croniter>=0.3.34",
-        f"grpcio>={GRPC_VERSION_FLOOR},<{GRPC_VERSION_CAP}; python_version<'3.11'",
-        f"grpcio>={GRPC_VERSION_FLOOR}; python_version>='3.11'",
-        f"grpcio-health-checking>={GRPC_VERSION_FLOOR},<{GRPC_VERSION_CAP}; python_version<'3.11'",
-        f"grpcio-health-checking>={GRPC_VERSION_FLOOR}; python_version>='3.11'",
+    ]
+    + get_grpcio_requires()
```
shalabhc, line 102: @gibsondan I don't think this will work.

This will work look at the env var when we build the wheel, eg [here](https://github.com/dagster-io/internal/blob/e846c750f30c86bc9851f5bdfc91846d53d921c2/python_modules/automation-internal/automation_internal/release/cli.py#L271).

But not during `pip install` which will just use the wheel and not run setup.py.

We can only do this if we ask users to build from the source (eg run `python setup.py install`)

gibsondan:
gooooood point. thanks shalabh.

PR #14607, opened 2023-06-02
Revert "resources and io manager telemetry"

jmsanders:
Reverts dagster-io/dagster#14025

Tests are breaking on Python 3.7 and 3.8 (which we don't explicitly test on the branch - only on main).

I'm reverting for now to get builds passing again.


jamiedemaria:
I think the build is failing because 6fcdbcaea3fbb521fcd2fb38692f9257a1925b9f didn't get reverted

gibsondan (approved):

PR #14468, opened 2023-05-25
[telemetry] do not merge - example telemetry

jamiedemaria:
## Summary & Motivation
Adds a bunch of types of resources and io managers to Project Fully Featured to show how telemetry is collected. Will not actually merge this PR it is just a proof of concept for the other PRs in the stack 

The  relevant telemetry received is 

``` python
{"dagster_resources": "[
  {'module_name': 'dagster_dbt', 'class_name': 'DbtCliClientResource'}, 
  {'module_name': 'dagster_snowflake', 'class_name': 'SnowflakeResource'}, 
  {'module_name': 'dagster_duckdb_pandas', 'class_name': 'DuckDBPandasIOManager'}, 
  {'module_name': 'dagster_snowflake_pandas', 'class_name': 'SnowflakePandasIOManager'}, 
  {'module_name': 'dagster_snowflake', 'class_name': 'SnowflakeResource'}, 
  {'module_name': 'dagster_snowflake', 'class_name': 'snowflake_io_manager'}, 
  {'module_name': 'dagster_duckdb', 'class_name': 'duckdb_io_manager'}, 
  {'module_name': 'dagster_snowflake', 'class_name': 'snowflake_resource'}, 
  {'module_name': 'dagster_gcp', 'class_name': 'bigquery_resource'}
]", 
"has_custom_resources": "True",
}

```

PR #14615, opened 2023-06-05
Telemetry for resources and io managers

jamiedemaria:
"Reverts" dagster-io/dagster#14607 which reverted https://github.com/dagster-io/dagster/pull/14025 and  https://github.com/dagster-io/dagster/pull/14463

This should fix the bug/python error in the original PRs. Basically only python 3.9 and 3.10 support stacked `@classmethod` and `@property` decorators. https://docs.python.org/3.11/library/functions.html?highlight=classmethod#classmethod

Since the `@property` decorator isn't strictly required for telemetry to function, we remove it in this version and update the naming so that there aren't name collisions

Ran this version of telemetry on project fully featured like in https://github.com/dagster-io/dagster/pull/14468 and confirmed that the correct telemetry logs were produced

### DIFF
--- a/examples/project_fully_featured/project_fully_featured/resources/__init__.py
+++ b/examples/project_fully_featured/project_fully_featured/resources/__init__.py
@@ -4,11 +4,7 @@ from dagster._utils import file_relative_path
 from dagster_aws.s3 import S3Resource
 from dagster_aws.s3.io_manager import ConfigurablePickledObjectS3IOManager
 from dagster_dbt import DbtCliClientResource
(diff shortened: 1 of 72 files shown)


examples/project_fully_featured/project_fully_featured/resources/__init__.py
```diff
@@ -4,7 +4,11 @@
 from dagster_aws.s3 import S3Resource
```
jamiedemaria, line 4: TODO - restore this file

benpankow (approved):

PR #14616, opened 2023-06-05
feat(dbt): add method to retrieve success of dbt CLI process

rexledesma:
## Summary & Motivation

- According to https://docs.getdbt.com/reference/exit-codes, the only contract with dbt exit codes is that a zero exit code is a success.
- We should not raise an exception on behalf of the user. This behavior should be explicitly handled by the user. This is similar to the behavior in https://docs.getdbt.com/reference/programmatic-invocations#dbtrunnerresult.

## How I Tested These Changes
pytest

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
@@ -238,6 +238,14 @@ class DbtCliTask:
         """
         return list(self.stream_raw_events())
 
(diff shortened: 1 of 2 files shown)


rexledesma:
* master
  * **PR #14616** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14616" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14616?utm_source=stack-comment).

OwenKephart (approved):

PR #14617, opened 2023-06-05
disable alembic quiet calls for sensor logging test

prha:
## Summary & Motivation
The stamp alembic call temporarily disables the python logging at the root logger.  This is interfering with some sensor logging tests which checks to see if some python logs are being captured.  This was not an issue seen before, but is being exposed by the switch to sqlalchemy 2, which is causing some race conditions between the logging module reset and the call to log from the sensor.

## How I Tested These Changes
BK, local tests w/ sqlalchemy 2 installed.  Should only affect tests.  Will be extra tested with a revert of https://github.com/dagster-io/dagster/pull/14586

### DIFF
--- a/python_modules/dagster/dagster_tests/daemon_sensor_tests/test_run_status_sensors.py
+++ b/python_modules/dagster/dagster_tests/daemon_sensor_tests/test_run_status_sensors.py
@@ -22,7 +22,6 @@ from dagster._core.storage.event_log.base import EventRecordsFilter
 from dagster._core.test_utils import create_test_daemon_workspace_context, instance_for_test
 from dagster._core.workspace.context import WorkspaceProcessContext
 from dagster._core.workspace.load_target import WorkspaceFileTarget, WorkspaceLoadTarget
-from dagster._utils.env import environ
 
 from .conftest import create_workspace_load_target
 from .test_sensor_run import (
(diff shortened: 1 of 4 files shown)


prha:
* master
  * **PR #14617** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14617" target="_blank">Graphite</a>  👈
    * **PR #14618** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14618" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14617?utm_source=stack-comment).

python_modules/dagster/dagster_tests/daemon_sensor_tests/test_run_status_sensors.py
```diff
@@ -1239,58 +1240,59 @@ def test_logging_run_status_sensor(
     workspace_context: WorkspaceProcessContext,
     external_repo: ExternalRepository,
 ):
-    freeze_datetime = pendulum.now()
-    with pendulum.test(freeze_datetime):
-        success_sensor = external_repo.get_external_sensor("logging_status_sensor")
-        instance.start_sensor(success_sensor)
+    with environ({"OVERRIDE_ALEMBIC_QUIET": "False"}):
```
jmsanders, line 1243: Maybe a comment about why we're setting this here? Although I still think we should just never quiet - seems like it'd be very hard to diagnose a similar flaky test in the future to know this is the right fix.

jmsanders (dismissed):

rexledesma (approved):
we can also remove `quieten`, since this was the only place it was used. our use of dynamic logging configuration is truly a sin 🥲

PR #14619, opened 2023-06-05
[auto-materialize] Fix issue with double-firing freshness policies

OwenKephart:
## Summary & Motivation

Imagine you have a -> b -> c -> d, and all are up to date. `d` has a freshness policy, and `c` and `d` have lazy AMPs.

Now, between two ticks, there are two runs that complete: one run containing a and b, and another run containing just a, which finishes after the first one. First, we do the freshness calculation, which determines that in order to make d fresh again, we need newer data. `b` has quite new data at the moment (but not the newest possible), so executing `c` and `d` right now will pull that new data in, satisfying the freshness policy. So we exit the freshness part of the logic intending to materialize both `c` and `d`.

Now, we enter the eager part of the logic, which tells us we must analyze any assets downstream of something that was materialized between the last two ticks. `b` was materialized, so we look at `c`, which requires that all of its parents have the newest possible data, which they don't, so we add a ParentOutdated skip reason to `c`. However, this doesn't impact `d`, so `d` gets materialized (doing nothing), and `c` is skipped. This is bad!

On the next tick, we redo the freshness calculation, realize we didn't do anything, then materialize both `c` and `d` again. This time, there's no new materializations from `b`, so both go through, giving us our double-firing behavior.

This sort of situation is much more likely to show up when a/b are partitioned, as b can be out of date without multiple materializations of a happening in a single tick.

In theory, we should skip both `c` and `d` in these scenarios, but that would be much more involved, and I want to address that in a followup.


python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -718,6 +718,12 @@ def can_reconcile_candidate(candidate: AssetKeyPartitionKey) -> bool:
         instance=instance_queryer.instance,
     )
 
+    def will_be_materialized_for_freshness(asset_partition: AssetKeyPartitionKey) -> bool:
+        return asset_partition in conditions_by_asset_partition_for_freshness and all(
+            condition.decision_type == AutoMaterializeDecisionType.MATERIALIZE
```
johannkm, line 723: Maybe should included DISCARD? Let's say C is partitioned and hits the max partitions limit, D would still double fire right?

johannkm (approved):
That's a complex scenario! This fix seems ok, but it also makes me feel like there are likely more issues lurking with mixing lazy and eager

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
OwenKephart: technically, a partitioned asset will never have a ForFreshness reason, as only unpartitioned assets are handled by that logic, so that specific scenario isn't possible.

agree with the high level feedback that there's weird stuff that can happen at these edges

PR #14621, opened 2023-06-05
UPath IO Manager

smackesey:


sryza (commented):
Awesome. This has come up a bunch of times.

What do you think about requiring users to return an `OutputsByPartition` object in this situation (which contains a dictionary of partition keys to values)? That would eliminate the possibility of confusing edge cases where an individual output value is a dictionary.

E.g. the outputs are JSON blobs. The user code only handles a single partition, but they try to materialize a range of partitions. They get a confusing error message saying '"some_field_in_json_blog" is not a valid partition key'. We could provide a more clear error message: "IOManager operating on a range of partitions, but returned output value only corresponds to a single partition. Return OutputByPartitions instead."

I kind of regret not doing that in the input case: https://github.com/dagster-io/internal/discussions/4968.

smackesey:
>What do you think about requiring users to return an OutputsByPartition object in this situation (which contains a dictionary of partition keys to values)?

I agree with this idea as I believe we discussed before-- but do you imagine this being general across IO managers, or specific to the `UPathIOManager`? IMO it should be general-- that would mean `OutputsByPartition` is a generic with a type parameter generic to specify the type of the multi-partition container value (with the default being a dict).

sryza:
> IMO it should be general-- that would mean OutputsByPartition is a generic with a type parameter generic to specify the type of the multi-partition container value (with the default being a dict).

What's the advantage of this?

smackesey:
@sandy So I've been doing some work on `PartitionedOutput` but this ends up being somewhat complex. I will post an internal GH discussion about it.

In the course of that work, I've also come around to the opinion that `PartitionedOutput` should be optional. This has a few advantages:

- Maintains a consistent interface across IO managers (which can either return the plain value or `PartitionedOutput`, rather than having an abstraction that is only required for our builtins)
- Maintains consistency with the un-partitioned or one-partition case (`Output` is optional)

Therefore I think this should be landed as is with `PartitionedOutput` landed in a followup. This will also allow me to get the stale partitions PR on top of this in.

sryza:
No matter what approach we go with here, DB IO managers will need to accept different kinds of output objects than object store IO managers, right?

Given that, I think a special type is better than a dictionary, for the reasons discussed earlier.

someaveragepunter:
Hello,
one of my workstreams is currently blocked on this, do you have an ETA on this?
My actual requirement is to be able to run multiple partitions within the same process to reduce overhead.
image

smackesey:
@sryza, returning to this:

The `PartitionedOutput` question should be considered separately from IO managers, because in the status quo, the IO manager `handle_output` method does not receive an `Output` or `DynamicOutput` object-- instead the wrapped value is passed to the IO manager. And `PartitionedOutput` properly belongs at the same layer as `Output` and `DynamicOutput`, so the IO manager should always receive whatever is wrapped by `PartitionedOutput` instead of `PartitionedOutput` itself.

If we want to enforce use of a `PartitionedOutput` for returning an output value representing multiple partitions, that needs to be done at the level of the output checking on the op rather than IO manager. Obviously we can't start requiring that immediately but it's something we should consider for 2.0.

In the meantime, we can introduce `PartitionedOutput` as a recommendation return value for multi-partition outputs. There is currently an implementation in this PR, but it should go in a separate PR since it is a separate problem than simply allowing `UPathIOManager` to support multi-partition formats. IMO the clear choice for the type that `UPathIOManager` uses to represent a multi-partition output is `Dict[str, Any]` (which is what this PR implements), since that is the type it uses for multi-partition input.

@someaveragepunter Hopefully we get this over the line shortly, but you can always use a different IO manager or implement a custom IO manager that supports multiple partitions.

sryza:
> since that is the type it uses for multi-partition input.

I believe this was a mistake, and I wouldn't want to repeat it if we can avoid it.  Here's an example of the consequences: https://github.com/dagster-io/dagster/issues/15538.

smackesey:
My response to above point is on your proposal: https://github.com/dagster-io/dagster/discussions/15557#discussioncomment-6607538

jamiedemaria:
hey @smackesey  the need for this feature has come up again https://dagster.slack.com/archives/C01U954MEER/p1693816005927989 - what's the status of this pr? anything blocking it?

cmpadden:

PR #14625, opened 2023-06-05
[Auto-materialize Policy page] Group skips

salazarm:
## Summary & Motivation

As titled

## How I Tested These Changes

Storybook + local testing + jest test
Screen Shot 2023-06-05 at 7 50 06 PM
Screen Shot 2023-06-05 at 7 49 58 PM
Screen Shot 2023-06-05 at 7 49 42 PM
Screen Shot 2023-06-05 at 7 49 39 PM


bengotow (approved):
Code here looks good to me, but I'm a little unclear about the screenshots -- each item seems to contain a timestamp that is 1 minute earlier and there isn't anything illustrating grouping / collapsing? I wonder if we could show the start + end timestamp when things are collapsed rather than just the one timestamp? 

Sort of "no evaluations met -- 8:04PM -> 10:12PM"

salazarm:
@bengotow yeah, I'll update the PR to match the figma. I thought I couldn't do it for some reason but realized I just need to special case the first "conditions not met" evaluation (which wouldn't have a start timestamp)

PR #14627, opened 2023-06-05
feat(dbt): allow customization profile and target on the resource

rexledesma:
## Summary & Motivation
We add Dagster configuration for `profile` and `target`, then concatenate the associated command line flags to the dbt command that the user specifies in `DbtCli.cli(...)`. 

Arguably, the command line flags to specify the profile (`--profile`) and associated target (`--target`) should be configurable at the dbt global config level (https://docs.getdbt.com/reference/global-configs/about-global-configs). They are options in every single dbt command.

## How I Tested These Changes
pytest

### DIFF
--- a/examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/__init__.py
+++ b/examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/__init__.py
@@ -7,6 +7,11 @@ from .assets.build import my_dbt_assets
 defs = Definitions(
     assets=[my_dbt_assets],
     resources={
-        "dbt": DbtCli(project_dir=DBT_PROJECT_DIR, global_config=["--no-use-colors"]),
+        "dbt": DbtCli(
+            project_dir=DBT_PROJECT_DIR,
+            global_config=["--no-use-colors"],
+            profile="jaffle_shop",
+            target="dev",
+        ),
     },
 )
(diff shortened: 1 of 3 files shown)


rexledesma:
* master
  * **PR #14627** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14627" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14627?utm_source=stack-comment).

OwenKephart (commented):
Is the expected pattern that users should manually set values via `os.environ`? or is this more an example demonstrating "imagine you had these env vars set through some other means, here's how things would work"?

I think manually setting things in os.environ is a bit scary, as it has global side effects (imagine you have multiple dbt projects that you want to load).

Also, is  setting `profile:` and `target:` to env vars an established pattern in normal dbt world? If not, I think this would be a tough sell to existing users.

In general, I think it will be more natural for the user to just pass in a value for target / profile into `DbtCli`, rather than have to worry about environment variables. I share your disappointment that those aren't `global_config`, as that would be way cleaner for this implementation, but at the very least I think `target` is an important enough thing to earn itself a configuration param alongside `project_dir`.

rexledesma:
Yeah, I was being needlessly adversarial on this. Updated to add `profile` and `target` as Dagster configuration.

geoHeil:
The old way of the experimental API (manually setting the arguments) will continue to work right?

rexledesma:
> The old way of the experimental API (manually setting the arguments) will continue to work right?

`profile` and `target` are optional. If you want to work in the old case where you explicitly invoke `--target` and `--profile` in `DbtCli.cli(...)`, then do not configure `profile` and `target` on the `DbtCli` resource.

OwenKephart (approved):

PR #14633, opened 2023-06-06
Unfeature gate new Runs table

salazarm:
## Summary & Motivation

1. Copy the `.new` files over to the original files
2. Delete the `.new` files
3. remove `new` everywhere
4. Also remove the dagster/user tooltip

## How I Tested These Changes

I loaded the Runs page and everything was working


hellendag (approved):
startreknod

PR #14493, opened 2023-05-26
[docs] show snowflake resource w update api docs

jamiedemaria:
## Summary & Motivation
Noticed that pythonic snowflake resource never got added to the docs (oops). Also updates the doc strings to be more explicit about what connection objects are being returned in what circumstances

PR #14638, opened 2023-06-06
[docs] snowflake api docs

jamiedemaria:
## Summary & Motivation
re-open of #14493 for the docs preview 

Noticed that pythonic snowflake resource never got added to the docs (oops). Also updates the doc strings to be more explicit about what connection objects are being returned in what circumstances


jamiedemaria:
graphite mad at me now. closing this one too

PR #14643, opened 2023-06-06
Show evaluation counts instead of timestamps

johannkm:
Per offline discussion. Showing the timestamps doesn't really provide any additional information. Showing the number of evaluations can help the user be confident that the daemon is running

image

For the last item, we don't show an evaluation count because 
- it'd be a huge number
- we don't know when this asset was added. If it's a new asset it's wrong to say it was considered for all the previous evaluations

Screen Shot 2023-06-06 at 1 29 38 PM

If there's no evaluations besides the `no materialization conditions met`, we say `Before now` for similar reasons

Screen Shot 2023-06-06 at 1 33 39 PM


sryza (approved):

salazarm:
You could also update the stories for storybook https://dagit-core-storybook-qu8mmeyt6-elementl.vercel.app/?path=/story/assets-automaterializepolicypage-stories-automaterializepolicypage--empty-state. They're a little out of date.

salazarm (approved):

PR #14645, opened 2023-06-06
Revert changes from "Fix multi-asset subsetting logic"

OwenKephart:
## Summary & Motivation

Reverts changes from https://github.com/dagster-io/dagster/pull/13830.

It seems like this "automatically generating ops" business is indeed causing problems in some edge cases, and the risk/reward of leaving this behavior in is not very promising. We will want some strategy to address the issue brought up in original issue, but that solution will need to be more carefully considered, to avoid interfering with users who are not currently encountering that issue.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/assets.py
@@ -1028,7 +1028,7 @@ class AssetsDefinition(ResourceAddable, IHasInternalInit):
             )
         else:
             # multi_asset subsetting
-            return AssetsDefinition(
+            return AssetsDefinition.dagster_internal_init(
                 # keep track of the original mapping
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster/_core/definitions/assets.py
```diff
@@ -1099,7 +1028,22 @@ def subset_for(
             )
         else:
             # multi_asset subsetting
-            return self._subset_op_backed_asset(asset_subselection, selected_asset_keys)
+            return AssetsDefinition(
```
sryza, line 1031: We should use `dagster_internal_init` here.

sryza (approved):
Ah, too bad. Make sure to reopen the original issue if there is one.

PR #14649, opened 2023-06-06
Lower heartbeat timeout for daemon grpc servers

gibsondan:
summary:
noticed while testing these refresh changes that we have 3-4 standing gRPC servers for each code location, most of which are likely just waiting to time out in the steady state. Heartbeats are frequent so 20 seconds should be plenty here instead of 2x the average lifetime of the grpc server.

Test Plan: Daemon integration tests
run daemon locally, `ps aux | grep grpc ` now maxes out at 2-3 instead of 4-5

### DIFF
--- a/python_modules/dagster/dagster/_daemon/controller.py
+++ b/python_modules/dagster/dagster/_daemon/controller.py
@@ -48,8 +48,10 @@ HEARTBEAT_CHECK_INTERVAL = 15
 
 RELOAD_WORKSPACE_INTERVAL = 60
 
-DAEMON_GRPC_SERVER_RELOAD_INTERVAL = 60
-DAEMON_GRPC_SERVER_HEARTBEAT_TTL = 120
+# Amount of time that a local code server spun up by the daemon will keep running
+# after it is no longer receiving any heartbeat pings - for this duration there may be
+# multiple code server processes running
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster/_daemon/controller.py
```diff
@@ -48,8 +48,7 @@
 
 RELOAD_WORKSPACE_INTERVAL = 60
 
-DAEMON_GRPC_SERVER_RELOAD_INTERVAL = 60
-DAEMON_GRPC_SERVER_HEARTBEAT_TTL = 120
+DAEMON_GRPC_SERVER_HEARTBEAT_TTL = 20
```
alangenfeld, line 51: maybe leave a comment here with why 20 is good

alangenfeld (approved):

gibsondan:
I have tested this pretty thoroughly (including code servers that take a long time to start up - making sure that the daemon waits for any new servers to be ready before relinquishing heartbeats on the old server, so the lower timeout value should not have any adverse side effects - if anything it could probably be lower and shut down right away after nothing is heartbeating anymore)

But if I somehow missed something and we see issues related to how the daemon refreshes code every 60 seconds, we could always temporarily revert this

PR #14650, opened 2023-06-06
[fix] Fix error from field without subfields while assembling rootDefaultYaml

benpankow:
## Summary

Fixes an error which arose when the root `type_snap` had no subfields. We would try to access `field_names`, which has a `check` assertion that `fields` is not empty.


benpankow:
* master
  * **PR #14650** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14650" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14650?utm_source=stack-comment).

alangenfeld (approved):
test would be great

PR #14673, opened 2023-06-07
Migrate to official Databricks Python SDK

judahrand:
## Summary & Motivation

Databricks has released an official Python SDK: https://github.com/databricks/databricks-sdk-py

Dagster should probably be using it as it is the future of Databricks Python interaction.

I believe that the only user-facing change that this causes is that the User Agent has some additional information appended to it.

### DIFF
--- a/python_modules/libraries/dagster-databricks/dagster_databricks/databricks.py
+++ b/python_modules/libraries/dagster-databricks/dagster_databricks/databricks.py
@@ -1,22 +1,22 @@
 import base64
 import logging
 import time
-from typing import IO, Any, Mapping, Optional, Tuple
+from typing import IO, Any, Mapping, Optional, Tuple, cast
 
 import dagster
 import dagster._check as check
 import dagster_pyspark
+import databricks_api
+import databricks_cli.sdk
 import requests.exceptions
-from dagster._annotations import public
-from databricks_api import DatabricksAPI
-from databricks_cli.sdk import ApiClient, ClusterService, DbfsService, JobsService
-from typing_extensions import Final
+from dagster._annotations import deprecated, public
+from dagster._utils.backcompat import deprecation_warning
+from databricks.sdk import WorkspaceClient
+from databricks.sdk.service import compute, jobs
 
 import dagster_databricks
 
 from .types import (
-    DatabricksRunLifeCycleState,
-    DatabricksRunResultState,
     DatabricksRunState,
 )
 from .version import __version__
@@ -36,23 +36,48 @@ class DatabricksClient:
         self.host = host
         self.workspace_id = workspace_id
 
+        self._workspace_client = WorkspaceClient(
+            host=host,
+            token=token,
+            product="dagster-databricks",
+            product_version=__version__,
+        )
+
         # TODO: This is the old shim client that we were previously using. Arguably this is
         # confusing for users to use since this is an unofficial wrapper around the documented
-        # Databricks REST API. We should consider removing this in the future.
-        self.client = DatabricksAPI(host=host, token=token)
-        self.__setup_user_agent(self.client.client)
+        # Databricks REST API. We should consider removing this in the next minor release.
+        self._client = databricks_api.DatabricksAPI(host=host, token=token)
+        self.__setup_user_agent(self._client.client)
 
-        # Expose an interface directly to the official Databricks API client.
-        self._api_client = ApiClient(host=host, token=token)
+        # TODO: This is the old `databricks_cli` client that was previous recommended by Databricks.
+        # It is no longer supported and should be removed in favour of `databricks-sdk` in the next
+        # minor release.
+        self._api_client = databricks_cli.sdk.ApiClient(host=host, token=token)
         self.__setup_user_agent(self._api_client)
 
-    def __setup_user_agent(self, client: ApiClient) -> None:
+    def __setup_user_agent(self, client: databricks_cli.sdk.ApiClient) -> None:
         """Overrides the user agent for the Databricks API client."""
         client.default_headers["user-agent"] = f"dagster-databricks/{__version__}"
 
+    @deprecated
     @public
     @property
-    def api_client(self) -> ApiClient:
+    def client(self) -> databricks_api.DatabricksAPI:
+        deprecation_warning(
+            "`client` property on DatabricksClient",
+            "0.20",
+            "Use `workspace_client` property instead.",
+        )
+        return self._client
+
+    @client.setter
+    def client(self, value: databricks_api.DatabricksAPI) -> None:
+        self._client = value
+
+    @deprecated
+    @public
+    @property
+    def api_client(self) -> databricks_cli.sdk.ApiClient:
         """Retrieve a reference to the underlying Databricks API client. For more information,
         see the `Databricks Python API <https://docs.databricks.com/dev-tools/python-api.html>`_.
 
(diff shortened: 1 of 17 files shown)


judahrand:
> Is there a backwards compatibility concern here? I.e. would this not work with older versions of databricks? Or not an issue because Databricks is in the Cloud and doesn't have old versions?

Afaik, this is using the same underlying API ([Jobs 2.1](https://docs.databricks.com/api/workspace/jobs/runnow)). So I can't imagine that there would be backwards compatibility issues at the Databricks level.

> I noticed that the Databricks SDK is marked as experimental. That might mean it will be more likely to change under our feet than the current APIs we're using?

This is possible I suppose. From what I can tell of the SDK it is mostly autogenerated from their OpenAPI definitions internally so unless they change the actual API I imagine most things will remain somewhat static. Additionally, the [docs](https://docs.databricks.com/dev-tools/cli/index.html) for `databricks_cli` state that it is also 'Experimental' and also unsupported...  And other [docs](https://docs.databricks.com/archive/dev-tools/python-api.html) now encourage users to use the new SDK instead of `databricks_cli.

I think I may have also been wrong in saying that this would not introduce user facing changes. I believe it would introduce breaking changes in that the 'api_client' that the 'Dagster Resource' provides is now a different class. This should therefore result in a minor (or even major) version bump. 

I still think this is a good step forwards. Using `databricks_cli` isn't best practice anymore and it was never a very ergonomic library anyway... The new SDK looks *much* more promising.

judahrand:
> To maintain backwards compatibility, we could have the internal implementation use the new databricks.sdk API. But the underlying client that we expose to the user (.api_client) should still be the one from databricks_cli.

`databricks_cli` is no longer supported so Dagster should probably move away from supporting it. Is there a deprecation policy for this sort of case that Dagster uses? 

> There are related docs changes associated with this as well.

Sure. Didn't want to embark on those if the code changes were unlikely to be accepted.

> This would break users, as the underlying API client would have changed. For example, triggering a one-time run of a Databricks workload has changed from .submit_run to .submit.

No. This is calling the same API. See: https://github.com/databricks/databricks-cli/blob/dd47ce4b42cd3ad17fc74022651aa0dbbbaf1d98/databricks_cli/runs/api.py#L31-L33

judahrand:
Or more correctly see: https://github.com/databricks/databricks-cli/blob/dd47ce4b42cd3ad17fc74022651aa0dbbbaf1d98/databricks_cli/sdk/service.py#L272-L274

judahrand:
I've absolutely no idea why the method is called `.submit_run` in `databricks_cli` when it is actually calling `/jobs/submit`... but Databricks be Databricks (pretty terrible considering how much it costs...)

rexledesma:
I know `.submit_run` and `.submit` are calling the same underlying Databricks API. That is not the issue. The point is the signature for calling the underlying API has changed for users. So this change needs to be released in a minor release.

If we want to release this in a patch release, then users who are still calling `.submit_run` should not be broken. In this case, `.api_client` should continue to return the client from `databricks_cli`. Then, we can fully migrate over during a minor release.

judahrand:
(Github are having issues with PRs [it seems](https://www.githubstatus.com/) but closing and reopening the PR is resolving some of it for me.)

judahrand:
@rexledesma Is this a more acceptable set of changes?

python_modules/libraries/dagster-databricks/dagster_databricks/databricks.py
```diff
@@ -31,27 +33,55 @@ class DatabricksError(Exception):
...
 
-    def __setup_user_agent(self, client: ApiClient) -> None:
+    def __setup_user_agent(self, client: databricks_cli.sdk.ApiClient) -> None:
         """Overrides the user agent for the Databricks API client."""
         client.default_headers["user-agent"] = f"dagster-databricks/{__version__}"
 
     @public
     @property
-    def api_client(self) -> ApiClient:
+    def client(self) -> databricks_api.DatabricksAPI:
```
rexledesma, line 68: Use `@deprecated` https://github.com/dagster-io/dagster/blob/1ac5c5934646f5b606872b34b1becc6054cc5d9b/python_modules/dagster/dagster/_annotations.py#L55

python_modules/libraries/dagster-databricks/dagster_databricks_tests/test_pyspark.py
```diff
@@ -225,6 +225,7 @@ def test_pyspark_databricks(
                 "execution": {"config": {"in_process": {}}},
             },
         )
+        print(result)
```
rexledesma, line 228: Remove

rexledesma (changes requested):
We shouldn't use `legacy_api` as a toggle. Just expose the `WorkspaceClient` in a new property.

python_modules/libraries/dagster-databricks/dagster_databricks/databricks.py
```diff
@@ -35,23 +37,50 @@ def __init__(self, host: str, token: str, workspace_id: Optional[str] = None):
...
+                "The `databricks_api.DatabricksAPI` client is deprecated and will be removed in a"
+                " future release. Please use `DatabricksClient.workspace_client`."
+            ),
+            DeprecationWarning,
+        )
+        return self._client
+
+    @client.setter
+    def client(self, value: databricks_api.DatabricksAPI) -> None:
+        self._client = value
```
rexledesma, line 78:
```suggestion
```

rexledesma (changes requested):
Almost there! Thanks for being patient on this. Just need to remove some unused stuff here, then we can run the test suite on this.

python_modules/libraries/dagster-databricks/dagster_databricks/databricks.py
judahrand: - Why do we want to remove the warnings? The [docstring](https://github.com/dagster-io/dagster/blob/3f9ea13096ec551a69cae5a542781a2c5a433ecc/python_modules/dagster/dagster/_annotations.py#L59) for the `deprecated` decorator specifically says that warnings should be added separately.
- If we don't have the setter for the `client` property we've introduced a breaking change since the user used to be able to assign to `.client` (even if they shouldn't have!).

python_modules/libraries/dagster-databricks/dagster_databricks/databricks.py
```diff
@@ -62,45 +91,93 @@ def api_client(self) -> ApiClient:
...
         Returns:
             ApiClient: The authenticated Databricks API client.
         """
+        warnings.warn(
+            (
+                "The `databricks_cli.sdk.ApiClient` client is deprecated and will be removed in"
+                " a future release. Please use `DatabricksClient.workspace_client`."
+            ),
+            DeprecationWarning,
+        )
```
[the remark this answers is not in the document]
rexledesma: Ah you're right, I misremembered. There's a utility for these deprecation warnings: mind using `from dagster._utils.backcompat import deprecation_warning`?

judahrand:
Realized that a local test failure that I thought was my local setup in fact wasn't. But I think I've fixed it now.

judahrand:
@rexledesma could you please re-trigger the tests?

judahrand:
@rexledesma bump

rexledesma:
@judahrand This is still breaking in our test suite. And since our CI processes are not accessible to the public, collaboration for this is a bit difficult. 

We're currently prioritizing other work, since our community has yet to explicitly ask for access to the new Databricks API. I will revisit this later in July and commit to the PR myself to fix the tests.

judahrand:
> And since our CI processes are not accessible to the public, collaboration for this is a bit difficult.

Are there any plans to fix this? One of the big advantages we thought we'd get from choosing to use Dagster was being able to contribute back to the project with our own fixes/improvements. It's really disappointing if that isn't really practical.

sryza:
@judahrand we are looking into fixing this. I definitely hear your pain on this one, and it's something I've been harping on quite a bit internally.

srggrs:
At my company we are waiting for this to be merged as the old databricks cli runs into these [silly click context issues](https://github.com/databricks/databricks-cli/issues/563) when running an op created by `create_databricks_run_now_op`. And that at the moment is breaking all our pipelines :disappointed: 

Thanks guys for working on this. LMK if I can help cheers

smackesey:
I am going to pull this down, adjust it to make it pass our CI, and merge it. @judahrand thanks for getting this going, it's sorely needed. You will still be recorded as contributor when it lands.

smackesey (approved):
Thanks @judahrand!

PR #14674, opened 2023-06-07
[ruff] Update 0.0.255 -> 0.0.277

smackesey:
## Summary & Motivation

Internal companion PR: https://github.com/dagster-io/internal/pull/5846

Updates ruff from 0.0.255 to 0.0.277.

This introduces autofix for the TCH rule, which allows auto-sorting type-checking only imports into `TYPE_CHECKING` blocks. Because the changes resulting from enabling this rule are large, they are in an upstack PR.

This PR contains only the ruff version update, a few fixes for issues surfaced by ruff behavior changes to already-enabled rules, and the new rules RUF009 and RUF010 (which are enabled because we enable the whole category of RUF* rules).

- RUF009 flags function calls in field defaults for dataclasses
- RUF010 converts e.g. this: `f"foo {repr(some_var)}"` to the more idiomatic `f"foo {some_var!r}"` using [f-string conversion flags](https://peps.python.org/pep-0498/#specification).

## How I Tested These Changes

ruff

Issue #14571, opened 2023-06-01
Importing annotations from __future__ breaks Config


mjclarke94:
### Dagster version

1.3.6

### What's the issue?

We are attempting to create classes which define their own Config structure internally. As such, the model needs to be able to self-reference, which requires that we use `from __future__ import annotations`, but dagster is very much not happy about this!

### What did you expect to happen?

It should be possible to enable forward references (which work fine with Pydantic models) and have them work with `dagster.Config`.

### How to reproduce?

Consider the following:
```
from pydantic import BaseModel
from __future__ import annotations

class Foo(BaseModel):
    num: int
    config: Foo.ModelConfig
    
    class ModelConfig(BaseModel):
        mult: int

    def calculate(self):
        return self.num * self.config.mult
```

This is a valid Pydantic model, where we include the code required to configure the model `Foo` within the definition for Foo. Ideally we'd be able to use the `Foo.BaseModel` as a config directly (maybe a feature request needed!) so external libraries don't need Dagster as a dependency, but we'll adjust accordingly:

```
from dagster import Config, asset
from pydantic import BaseModel
from __future__ import annotations


class Foo(BaseModel):
    num: int
    config: Foo.ModelConfig
    
    class ModelConfig(Config):
        mult: int

    def calculate(self, config):
        return self.num * config.mult
```

Again, all good so far. However when we use this in an asset, we get an error.

```
from dagster import Config, asset
from pydantic import BaseModel
from __future__ import annotations


class Foo(BaseModel):
    num: int
    config: Foo.ModelConfig
    
    class ModelConfig(Config):
        mult: int

    def calculate(self, config):
        return self.num * config.mult
    
@asset
def my_asset(config: Foo.ModelConfig) -> int:
    foo = Foo()
    return foo.calculate(config = config)

___

DagsterInvalidPythonicConfigDefinitionError: 
Error defining Dagster config class.
Unable to resolve config type 'Foo.ModelConfig' to a supported Dagster config type.
```

Ok, let's strip it back to this:

```
from dagster import Config, asset

class ModelConfig(Config):
    mult: int

@asset
def my_asset(config: ModelConfig) -> int:
    foo = Foo()
    return foo.calculate(config = config)
```

This works, but once we enable annotations, we get the same error as before:

```
from dagster import Config, asset
from __future__ import annotations


class ModelConfig(Config):
    mult: int

@asset
def my_asset(config: ModelConfig) -> int:
    foo = Foo()
[...]

PR #14675, opened 2023-06-07
[ruff] Ban `from __future__ import annotations` and enable TCH (typing import autosort)

smackesey:
## Summary & Motivation

Internal companion PR: https://github.com/dagster-io/internal/pull/5847

- Turn on TID251 (banned imports) and configure it to ban `from __future__ import annotations`.
- Turn on TCH suite of rules, which allows auto-sort to automatically place typing-only imports in a `TYPE_CHECKING` block.

This PR was a journey. The goal was to get automatic sorting of imports only used in type annotations into `TYPE_CHECKING` blocks. Ruff offers the `TCH` set of rules for this, for which autofix was recently added. Unfortunately, it turned out to be more complicated than just turning on these rules, because the way things currently work, ruff will only autosort the imports if the annotations are already all strings, which most of the time means `from __future__ import annotations` is imported.

The initial plan was actually opposite the current state of this PR-- the idea was to _inject_ `from __future__ import annotations` (using `I002`) into all Python modules for maximum `TCH` power. However, for reasons explained below, `from __future__ import annotations` (a) does not work well with Dagster in particular; (b) is not a good thing to add everywhere because it's not part of the future of Python.

Therefore, I think the best thing to do for simplicity and future-proofing of the codebase is to ban use of `from __future__ import annotations`. It was imported in ~15 of our 2500 source files-- this PR removes them, adds needed quotes to affected type annotations, and turns on `TID251` to flag any future use of `from __future__ import annotations`.

Without `from __future__ import annotations`, typing-only import auto-sort (`TCH`) is relatively limited. However, I have submitted a [ruff feature request](https://github.com/astral-sh/ruff/issues/5559) to get a new ruff setting that will make it much more powerful for codebases not using `from __future__ import annotations`.

---

The issues with string annotations, `TYPE_CHECKING`, and `from __future__ import annotations` are complex and merit a thorough explanation.

Suppose I have the following module:

```
from dagster._core.definitions.events import AssetKey

def do_something_to_asset(key: AssetKey):
    ...
```

`AssetKey` above is only used in a type annotation. However, if we move it to a `TYPE_CHECKING` block with no other changes, then our code will crash. That's because, by default, Python evaluates type annotations (in function sigs and module-scope variables, but _not_ local var annotations) as standard expressions at runtime. If the `AssetKey` import is in a `TYPE_CHECKING` block, then it won't execute at runtime, and when the `AssetKey` annotation is encountered, Python will be unable to evaluate the symbol `AssetKey` and will error. Therefore we can only move the `AssetKey` import to a typing block under two conditions (1) we turn it into a string annotation; (2) the evaluated expression never needs to be accessed. Let's consider both of these separately:

## 1. Turning annotation into a string

There are two ways to do this:

1. Quote the `AssetKey` annotation.
2. Add `from __future__ import annotations` to the file. This will cause _all_ annotations in the file to be interpreted it as strings. It is equivalent to quoting all annotations in the file.

Option (2) is smoother and eliminates the extra visual complexity of having some arguments quoted and others not. It also improves import time. (Improvement for Dagster is 10-20%; I performed a rough experiment timing `import dagster` on this branch vs master) This was one of the motivations for PEP 562, which introduced `from __future__ import annotations` and _was_ scheduled to become the standard way annotations are evaluated in Python 3.11-- but that plan has been changed, see below.

Turning _all_ annotations into strings does not play nicely with most libs that do runtime processing of annotations. This includes Pydantic and... Dagster. The reason is that these libs need access to the result of the evaluated expression, not just a string:

```
from dagster import op
from pandas import DataFrame

@op
def foo(bar: "DataFrame"):  # This is currently an error
    ...
```

This problem can be somewhat mitigated by using `typing.get_type_hints` on functions that might have string annotations-- `get_type_hints` evaluates the annotations and returns them, providing the necessary runtime values. The problem is that it doesn't always work-- if string annotations refer to symbols in local scope, `get_type_hints` can't resolve them:

```
from typing import get_type_hints

class Foo:
    ...

def foo(x: "Foo"):
     ...

# works to dereference "Foo" since referent `Foo` is in module-level scope
get_type_hints(foo)  # => {'x': <class '__main__.Foo'>}

def make_bar():
    class Bar:
        ...

    def bar(x: "Bar"):
        ...
    return bar

bar = make_bar()

# does not work to dereference "Bar" since referent `Bar` vanished with local scope
get_type_hints(bar)  # => NameError: name 'Bar' is not defined.
```

So string annotations, either explicit or via `from __future__ import annotations`, are flatly incompatible with any situation where (1) annotations are being created in a local scope; (2) an annotation references a symbol only available in that local scope; (3) the annotation must be operated on at runtime (and therefore evaluated with `get_type_hints`).

For the above reasons (and because of the popularity of Pydantic), the Python Steering Committee nixed plans to adopt string annotations in Python 3.11. The new plan is described in [PEP 649](https://peps.python.org/pep-0649/) and will instead package all annotations in a lazily evaluated function. This both enables forward references _and_ ensures that the evaluated value of an annotation can always be accessed.

Unfortunately, PEP 649 won't become part of Python until 3.12, and it's unclear to me whether it will be backportable via another `from __future__` import (but I suspect no). Since we need to support Python <3.12 for many years, that means we probably won't have access to this solution.

## 2. Evaluated expression never needs to be accessed

Even if we had access to PEP 649 the problems discussed in the preceding section were solved, the ability to move "annotation-only" imports to a `TYPE_CHECKING` block is still limited by whether the evaluated value _ever_ needs to be accessed. If it does, you can't move an import to a `TYPE_CHECKING` block.

For us, this mostly means that any type used in an annotation for a Dagster decorator should never be moved to a `TYPE_CHECKING` block:

```
# This is only used in an annotation but it will break dagster if it is moved to a `TYPE_CHECKING` class
from foo import SomeDagsterConfigClass

@asset
def an_asset(config: SomeDagsterConfigClass):
    ...
```

Fortunately ruff allows configuring `TCH` rules to detect special decorators like `@asset`, so I don't think this limitation should block us.

## How I Tested These Changes

ruff

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -230,6 +230,10 @@ select = [
   # (private member access) Flag access to `_`-prefixed symbols. By default the various special
   # methods on `NamedTuple` are ignored (e.g. `_replace`).
   "SLF001",
+  
+  # (flake8-type-checking) Auto-sort imports into TYPE_CHECKING blocks depending on whether
+  # they are runtime or type-only imports.
+  "TCH",
 
   # (disallow print statements) keep debugging statements out of the codebase
   "T20",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -233,7 +233,10 @@ select = [
   
   # (flake8-type-checking) Auto-sort imports into TYPE_CHECKING blocks depending on whether
   # they are runtime or type-only imports.
-  "TCH",
+  # "TCH",
+
+  # (banned-api) Flag use of banned APIs. See tool.ruff.flake8-tidy-imports.banned-api for details.
+  "TID251",
 
   # (disallow print statements) keep debugging statements out of the codebase
   "T20",
@@ -252,6 +255,11 @@ required-version = "0.0.277"
 
 # We use `id` in many places and almost never want to use the python builtin.
 builtins-ignorelist = ["id"]
+
+[tool.ruff.flake8-tidy-imports.banned-api]
+
+"__future__.annotations".msg = "Test"
+
 
 [tool.ruff.isort]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -233,7 +233,7 @@ select = [
   
   # (flake8-type-checking) Auto-sort imports into TYPE_CHECKING blocks depending on whether
   # they are runtime or type-only imports.
-  # "TCH",
+  "TCH",
 
   # (banned-api) Flag use of banned APIs. See tool.ruff.flake8-tidy-imports.banned-api for details.
   "TID251",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -258,8 +258,7 @@ builtins-ignorelist = ["id"]
 
 [tool.ruff.flake8-tidy-imports.banned-api]
 
-"__future__.annotations".msg = "Test"
-
+"__future__.annotations".msg = "Directly quote annotations instead."
 
 [tool.ruff.isort]
 
--- a/examples/docs_snippets/pyproject.toml
+++ b/examples/docs_snippets/pyproject.toml
@@ -12,7 +12,7 @@ target-version = ['py38', 'py39', 'py310', 'py311']
 
 [tool.ruff]
 
-# Extend root configuration.
+# Extend example root configuration.
 extend = "../pyproject.toml"
 
 # Match black. Note that this also checks comment line length, but black does not format comments.
@@ -34,6 +34,12 @@ extend-ignore = [
   # purposes.
   "F841",
 
+  # (isort): No need to automatically insert imports in docs snippets.
+  "I002",
+
+  # (flake8-type-checking) We can't use this in docs_snippets because we don't auto-insert `from
+  # __future__ import annotations`. But we also don't need it becasse we don't use `TYPE_CHECKING`.
+  "TCH",
 ]
 
 [tool.ruff.isort]
--- a/examples/docs_snippets/pyproject.toml
+++ b/examples/docs_snippets/pyproject.toml
@@ -34,11 +34,7 @@ extend-ignore = [
   # purposes.
   "F841",
 
-  # (isort): No need to automatically insert imports in docs snippets.
-  "I002",
-
-  # (flake8-type-checking) We can't use this in docs_snippets because we don't auto-insert `from
-  # __future__ import annotations`. But we also don't need it becasse we don't use `TYPE_CHECKING`.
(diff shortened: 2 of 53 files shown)


smackesey:
* master
  * **PR #14674** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14674" target="_blank">Graphite</a> 
    * **PR #14675** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14675" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14675?utm_source=stack-comment).

mjclarke94:
Hey @smackesey, with regards to 2), have you seen #14571? There are some contexts where importing `__future__.annotations` seems cause issues.

smackesey:
@mjclarke94 

>with regards to 2), have you seen https://github.com/dagster-io/dagster/issues/14571? There are some contexts where importing __future__.annotations seems cause issues.

Thanks for linking that-- I independently discovered this issue while working on this PR. I believe I have a fix for that bug in _most_ scenarios here: https://github.com/dagster-io/dagster/pull/14707#event-9723240402

Turns out this is a pretty complicated topic, and the future of this PR is uncertain.

smackesey:
@benpankow This PR summary now contains an explanation of the situation with runtime introspection of annotations and `from __future__ import annotations`-- tagging you because you probably want to understand this (somewhat convoluted) situation as the main developer of Pythonic config.

pyproject.toml
```diff
@@ -249,6 +256,10 @@ required-version = "0.0.277"
 # We use `id` in many places and almost never want to use the python builtin.
 builtins-ignorelist = ["id"]
 
+[tool.ruff.flake8-tidy-imports.banned-api]
+
+"__future__.annotations".msg = "Test"
```
alangenfeld, line 261: Test?

alangenfeld (commented):
I see this in the summary

> Fortunately ruff allows configuring TCH rules to detect special decorators like @asset, so I don't think this limitation should block us.

but no code related to it unless im missing something

alangenfeld (approved):
this seems reasonable enough from what im able to understand - if you are confident i think this is fine but re-request review if theres is something risky in here that merits extra attention

pyproject.toml
smackesey: good catch

PR #14677, opened 2023-06-07
Remove the okta early access banner (to land once we are live in the okta store)

gibsondan:
Summary:
Just so I don't lose track. I'll land this once it's true.

Test Plan: View docs preview


erinkcochran87:
@gibsondan There's also a banner here: `/dagster-cloud/account/authentication/utilizing-scim-provisioning`

Should this guy be removed as well?

gibsondan:
good catch, removed the other one as well

erinkcochran87 (approved):

Issue #13325, opened 2023-04-03
Quick access to current run compute logs from the Dagit UI asset sidebar


mkleinbort-ic:
### What's the use case?

I have some long running asset materialization, and it'd be good to see the logs within the asset lineage page. 

So, my feature request is:

"As someone monitoring asset materializations, I'd like to be able to have quick access to an asset materialization's compute logs from within the asset lineage view so that I can quickly see the progress of my assets"

### Ideas of implementation

I envision this as adding these views:

![image](https://user-images.githubusercontent.com/112868935/229468439-f7c5f307-bd4f-4ea0-b4c4-5cdd0ff13e99.png)

From the current run to the side panel of that opens when you click on an asset in the Dagit Global Asset Lineage.

This should point to the current run's logs or the latest run's logs. 

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

sryza:
Thanks for the idea @mkleinbort-ic - I'm forwarding this to the Dagster UI team.

PR #14682, opened 2023-06-07
[ui] Add step compute logs to “Last Materialization” in the asset sidebar

bengotow:
## Summary & Motivation

Fixes: https://github.com/dagster-io/dagster/issues/13325
Loom: https://www.loom.com/share/0a43f36d16644d71a7f6729f7fe1d4ea

Note: Right now, we only show the compute logs /after/ the step has completed -- I believe we need the materialization event with the stepKey to know what stepKey to pull logs for. I don't know that there's a way to get from the asset node => op / op names to a step key.  For multi-op assets (graph backed assets) this shows the run logs of the op that emitted the materialization event, for the same reason.

image

## How I Tested These Changes

We don't currently have tests for the asset sidebar but I'll see if I can get some going later this week! This was tested manually by viewing compute logs both on the runs page and from the new sidebar button.

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/LastMaterializationMetadata.tsx
+++ b/js_modules/dagit/packages/core/src/assets/LastMaterializationMetadata.tsx
@@ -1,4 +1,4 @@
-import {Box, Colors, Group, Icon, Mono, NonIdealState, Table} from '@dagster-io/ui';
+import {Box, Button, Colors, Group, Icon, Mono, NonIdealState, Table} from '@dagster-io/ui';
 import * as React from 'react';
 import {Link} from 'react-router-dom';
 import styled from 'styled-components/macro';
@@ -9,6 +9,7 @@ import {AssetKeyInput} from '../graphql/types';
 import {MetadataEntry} from '../metadata/MetadataEntry';
 import {PipelineReference} from '../pipelines/PipelineReference';
 import {linkToRunEvent, titleForRun} from '../runs/RunUtils';
+import {StepComputeLogsDialog} from '../runs/StepComputeLogsDialog';
 import {isThisThingAJob, useRepository} from '../workspace/WorkspaceContext';
 import {buildRepoAddress} from '../workspace/buildRepoAddress';
 
@@ -24,6 +25,7 @@ export const LatestMaterializationMetadata: React.FC<{
   latest: AssetObservationFragment | AssetMaterializationFragment | undefined;
   liveData: LiveDataForNode | undefined;
 }> = ({assetKey, latest, liveData}) => {
+  const [showingComputeLogs, setShowingComputeLogs] = React.useState(false);
   const latestRun = latest?.runOrError.__typename === 'Run' ? latest?.runOrError : null;
   const repositoryOrigin = latestRun?.repositoryOrigin;
   const repoAddress = repositoryOrigin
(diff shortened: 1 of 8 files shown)


js_modules/dagit/packages/core/src/runs/StepComputeLogsDialog.tsx
```diff
@@ -0,0 +1,143 @@
...
+      });
+      setComputeLogFileKey(selectedLogKey || logFileKeys[0]);
+    } else if (!stepKeys.includes(computeLogFileKey)) {
+      const matching = matchingComputeLogKeyFromStepKey(metadata.logCaptureSteps, stepKeys[0]);
+      matching && setComputeLogFileKey(matching);
+    } else if (stepKeys.length === 1 && computeLogFileKey !== stepKeys[0]) {
+      const matching = matchingComputeLogKeyFromStepKey(metadata.logCaptureSteps, stepKeys[0]);
+      matching && setComputeLogFileKey(matching);
+    }
+  }, [stepKeys, computeLogFileKey, metadata.logCaptureSteps, setComputeLogFileKey]);
```
salazarm, line 99: I'm always wary of calling `setState` inside of `useEffect` but I trust its probably necessary here

salazarm (approved):

bengotow:
Ok I've updated this PR to include the full set of toolbar options and the structured log viewer. It supports filtering down the structured logs in the normal ways 🙌 

image
image

js_modules/dagit/packages/core/src/assets/LastMaterializationMetadata.tsx
```diff
@@ -55,11 +57,30 @@ export const LatestMaterializationMetadata: React.FC<{
           <td>
             {latestRun ? (
               <div>
-                <Box>
-                  {'Run '}
-                  <Link to={`/runs/${latestEvent.runId}?timestamp=${latestEvent.timestamp}`}>
-                    <Mono>{titleForRun({id: latestEvent.runId})}</Mono>
-                  </Link>
+                <Box flex={{direction: 'row', justifyContent: 'space-between'}}>
```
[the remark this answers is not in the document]
bengotow: Talked to josh about this one and I think we're going to leave it as is -- it's weird looking, but it feels more weird for these two text blocks to be unaligned.
 
image

js_modules/dagit/packages/core/src/runs/useComputeLogFileKeyForSelection.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+import {useQueryPersistedState} from '../hooks/useQueryPersistedState';
+
+import {
+  IRunMetadataDict,
+  matchingComputeLogKeyFromStepKey,
+  ILogCaptureInfo,
+} from './RunMetadataProvider';
+
+export function useComputeLogFileKeyForSelection({
```
bengotow, line 16: Just moved this logic out to a hook so that it didn't need to be replicated to show the logs in a second place

js_modules/dagit/packages/core/src/runs/LogsToolbar.tsx
```diff
@@ -95,6 +90,11 @@ export const LogsToolbar: React.FC<ILogsToolbarProps> = (props) => {
     [onSetLogType, setInitialComputeLogType],
   );
 
+  const activeItems = React.useMemo(
+    () => new Set([logType === LogType.structured ? logType : initialComputeLogType]),
```
bengotow, line 94: This is a small bug fix that prevented the "stderr" tab from showing an active item in the structured vs. raw picker

bengotow:
Yep @sryza! Here's a screenshot - right now it's a full size button in the sidebar next to the Run ID. We could also make it smaller but Josh and I were thinking it could be revisited along with the rest of the sidebar and was cool enough to justify a big button vs a small icon.

image

sryza:
I like the button. I wonder if we could get away with just "View logs" to avoid introducing the term "step" here?

PR #14684, opened 2023-06-07
Attempt to run graphql tests in parallel

jmsanders:
Some of our graphql test suites have crept up to their 25 minute limit. The test orchestration here is...complicated.

I'm going to just try the easiest possible solution which is to give pytest 2 parallel workers and see what happens.


jmsanders:
The suite ran in 15 minutes with 2 threads (and like 9 minutes on my machine with more...but let's start with 2 and see how much that buys us)

PR #14695, opened 2023-06-08
Only check per-location permissions for the specific asset keys being backfilled

gibsondan:
Summary:
Before we were checking every code location with an asset on it during a backfill - now we check the specific locations that are associated with the asset selection

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql_tests/graphql/test_asset_backfill.py
+++ b/python_modules/dagster-graphql/dagster_graphql_tests/graphql/test_asset_backfill.py
@@ -20,6 +20,7 @@ from dagster_graphql.test.utils import (
     GqlResult,
     define_out_of_process_context,
     execute_dagster_graphql,
+    main_repo_location_name,
 )
 
 GET_PARTITION_BACKFILLS_QUERY = """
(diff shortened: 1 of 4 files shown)


python_modules/dagster-graphql/dagster_graphql_tests/graphql/test_asset_backfill.py
```diff
@@ -150,6 +151,36 @@ def test_launch_asset_backfill_read_only_context():
...
+        ) as read_only_context:
+            assert read_only_context.read_only
+            # launchPartitionBackfill
+            launch_backfill_result = execute_dagster_graphql(
+                read_only_context,
+                LAUNCH_PARTITION_BACKFILL_MUTATION,
+                variables={
+                    "backfillParams": {
+                        "partitionNames": ["a", "b"],
+                        "assetSelection": [key.to_graphql_input() for key in all_asset_keys],
```
benpankow, line 172: just to make sure, `all_asset_keys` here is the set of all asset keys in `location_name`?

benpankow: are we de-selecting any assets with the subselection (e.g. assets in a read-only location?)

benpankow (approved):
Looks good to me, thanks for investigating

PR #14696, opened 2023-06-08
[ui] Turn on noUncheckedIndexedAccess for TS, Part 1

hellendag:
## Summary & Motivation

The first wave of changes made to allow us to enable `noUncheckedIndexedAccess` in TypeScript in `packages/core`.

This generally consists of a few different approaches:

- Use the `!` assertion operator on array/object key access. I used this when it looked like a previous check had validated the presence of the key, e.g. checking `.length` on an array and subsequently looking up `[0]`.
- Use `.forEach` instead of `for (let ii = ...)` style loops where possible, since then we don't need to key into the array.
- Similarly, use `Object.entries()` for loops that currently use `Object.keys()` and then index into the object based on the key.

It is possible that this will surface new errors, especially in situations where the `!` assertion is inaccurate. Fortunately, the error should now be very obvious at that point in code, instead of surfacing downstream.

There will be one or two more PRs to finish this for OSS.

## How I Tested These Changes

ts, lint, jest.

### DIFF
--- a/js_modules/dagit/packages/core/src/workspace/useRepositoryForRun.ts
+++ b/js_modules/dagit/packages/core/src/workspace/useRepositoryForRun.ts
@@ -103,7 +103,7 @@ export const useRepositoryForRunWithoutSnapshot = (
   }
   const jobNameMatches = jobNameMatchesForRun(options, run);
   if (jobNameMatches && jobNameMatches.length) {
-    return {match: jobNameMatches[0], type: 'pipeline-name-only'};
+    return {match: jobNameMatches[0]!, type: 'pipeline-name-only'};
   }
   return null;
 };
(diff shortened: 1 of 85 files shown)


js_modules/dagit/packages/core/src/workspace/useRepositoryForRun.ts
```diff
@@ -141,11 +141,11 @@ export const useRepositoryForRunWithParentSnapshot = (
   }
 
   if (snapshotMatches && snapshotMatches.length) {
```
salazarm, line 143: Instead of using `!` can we just explicitly check `snapshotMatches[0]` instead of `.length` ?

salazarm (approved):
Makes sense to me! Though I wonder if instead of using `!` everywhere we instead just use a comment to disable the rule. That'll make it more obvious that the code maybe not be properly guarding against nulls? This is generally the strategy we used at FB where we put // HH_FIX_ME or // $FlowFixMe's whenever new rules were introduced

js_modules/dagit/packages/core/src/workspace/useRepositoryForRun.ts
hellendag: Yep, that seems reasonable for these.

hellendag: I accidentally applied this change to the last branch in the stack, so that's where it'll show up.

PR #14706, opened 2023-06-08
Handle partitions with invalid parents in auto-materialize

clairelin135:
Background here: https://www.notion.so/dagster/Handle-invalid-parents-in-auto-materialize-5fe6dbdfba0f43a6beae88fb61ccc07d?pvs=4

This PR builds upon https://github.com/dagster-io/dagster/pull/14449 and introduces graceful handling for auto-materialize cases where an invalid parent exists. Errors are currently raised in `PartitionMapping.get_upstream_partitions_for_partitions` when mapped upstream partitions were nonexistent. It is good that an error occurs when launching backfills and runs in these cases, but undesirable for the error to occur in auto-materialization. This PR introduces logic to gracefully handle invalid parents without erroring in auto-materialize.

**Modifies the auto-materialize logic:**
- Only considers partitions that can be materialized. If a partition has invalid upstreams, it should not be possible to auto-materialize it.
- Modifies `is_reconciled` to recursively check only valid upstream partitions. If all of a partition's parents are invalid, then returns `True` if the partition is materialized, `False` if not. This enables reconciling downstream assets.

**Adds new methods to `PartitionMapping`:**
- `downstream_partition_has_valid_upstream_partitions`: Users should be able to define on the partition mapping whether some dependency relationship is allowed or not. This method enables gracefully checking if a particular partition has valid parents (rather than raising an error within `get_upstream_partitions_for_partitions`) 
- `get_valid_upstream_partitions_for_partitions`:  a particular partition can have multiple upstream assets, each of which can contain valid or invalid mapped partitions. In our reconciliation checks, we want to still check that the valid upstream parents are reconciled, and skip checking the invalid upstream parents. This necessitates defining a function that filters for valid parents.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -733,7 +733,11 @@ def determine_asset_partitions_to_auto_materialize(
         for parent in asset_graph.get_parents_partitions(
             instance_queryer, evaluation_time, candidate.asset_key, candidate.partition_key
         ):
-            if instance_queryer.is_reconciled(asset_partition=parent, asset_graph=asset_graph):
+            if instance_queryer.is_reconciled(
+                asset_partition=parent,
+                asset_graph=asset_graph,
+                dynamic_partitions_store=instance_queryer,
+            ):
                 continue
 
             if not (
@@ -776,7 +780,9 @@ def determine_asset_partitions_to_auto_materialize(
         ):
             conditions.add(MissingAutoMaterializeCondition())
         elif auto_materialize_policy.on_new_parent_data and not instance_queryer.is_reconciled(
-            asset_partition=candidate, asset_graph=asset_graph
+            asset_partition=candidate,
+            asset_graph=asset_graph,
+            dynamic_partitions_store=instance_queryer,
         ):
             conditions.add(ParentMaterializedAutoMaterializeCondition())
 
@@ -815,8 +821,13 @@ def determine_asset_partitions_to_auto_materialize(
             return False
 
         if all(
-            will_be_materialized_for_freshness(candidate)
-            or parents_will_be_reconciled(asset_graph, candidate)
+            asset_graph.partition_maps_to_valid_parents(
+                instance_queryer, evaluation_time, candidate
+            )
+            and (
+                will_be_materialized_for_freshness(candidate)
+                or parents_will_be_reconciled(asset_graph, candidate)
+            )
             for candidate in candidates_unit
         ):
             unit_conditions = set().union(
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -701,6 +701,10 @@ def determine_asset_partitions_to_auto_materialize(
             candidate
         ):
             return False
+        elif not asset_graph.partition_maps_to_valid_parents(
+            instance_queryer, evaluation_time, candidate
+        ):
+            return False
 
         return True
 
@@ -736,7 +740,6 @@ def determine_asset_partitions_to_auto_materialize(
             if instance_queryer.is_reconciled(
                 asset_partition=parent,
                 asset_graph=asset_graph,
-                dynamic_partitions_store=instance_queryer,
             ):
                 continue
 
@@ -782,7 +785,6 @@ def determine_asset_partitions_to_auto_materialize(
         elif auto_materialize_policy.on_new_parent_data and not instance_queryer.is_reconciled(
             asset_partition=candidate,
             asset_graph=asset_graph,
-            dynamic_partitions_store=instance_queryer,
         ):
             conditions.add(ParentMaterializedAutoMaterializeCondition())
 
@@ -821,13 +823,8 @@ def determine_asset_partitions_to_auto_materialize(
             return False
 
         if all(
-            asset_graph.partition_maps_to_valid_parents(
-                instance_queryer, evaluation_time, candidate
-            )
-            and (
-                will_be_materialized_for_freshness(candidate)
-                or parents_will_be_reconciled(asset_graph, candidate)
-            )
+            will_be_materialized_for_freshness(candidate)
+            or parents_will_be_reconciled(asset_graph, candidate)
             for candidate in candidates_unit
         ):
             unit_conditions = set().union(
--- a/python_modules/dagster/dagster/_core/definitions/time_window_partition_mapping.py
+++ b/python_modules/dagster/dagster/_core/definitions/time_window_partition_mapping.py
@@ -19,15 +19,7 @@ from dagster._utils.backcompat import (
 )
 
 
-class MappedTimeWindowPartitionsResult(
-    NamedTuple(
-        "_MappedTimeWindowPartitionsResult",
-        [
-            ("partitions_subset", TimeWindowPartitionsSubset),
-            ("mapped_to_invalid_time_windows", bool),
-        ],
-    )
-):
+class MappedTimeWindowPartitionsResult(NamedTuple):
     """Represents the result of mapping a TimeWindowPartitionsSubset to the corresponding
     partitions in another TimeWindowPartitionsDefinition.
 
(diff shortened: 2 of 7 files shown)


python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -586,11 +590,9 @@ def is_reconciled(
...
             (TimeWindowPartitionsDefinition, DynamicPartitionsDefinition),
         )
-        for parent in asset_graph.get_parents_partitions(
-            self,
-            self._evaluation_time,
-            asset_partition.asset_key,
-            asset_partition.partition_key,
+
+        for parent in asset_graph.get_valid_parent_partitions(
+            dynamic_partitions_store, self._evaluation_time, asset_partition
```
OwenKephart, line 595: a quick nit before a more thorough review is that we can just pass in `self` here, rather than re-routing the caching instance queryer into itself at all the call sites.

python_modules/dagster/dagster/_core/definitions/time_window_partition_mapping.py
```diff
@@ -16,6 +16,25 @@
 from dagster._serdes import whitelist_for_serdes
 
 
+class MappedTimeWindowPartitionsResult(
```
OwenKephart, line 19: A more compact notation here is:

```
class MappedTimeWindowPartitionsResult(NamedTuple):
    """Docstring"""
    partitions_subset: TimeWindowPartitionsSubset
    mapped_to_invalid_time_windows: bool
```

I think it also would be reasonable to keep the returned tuple "unnamed", i.e. _map_partitions returns a regular tuple (subset, mapped_to_invalid), and the callsites of _map_partitions would just have:

partitions_subset, mapped_to_invalid_time_windows = _map_partitions(...)

I'll leave that bit up to you though

python_modules/dagster/dagster/_core/definitions/time_window_partition_mapping.py
```diff
@@ -112,6 +131,58 @@ def get_upstream_partitions_for_partition_range(
...
 
+    def downstream_partition_has_valid_upstream_partitions(
+        self,
+        downstream_partitions_def: PartitionsDefinition,
+        downstream_partition_key: str,
+        upstream_partitions_def: PartitionsDefinition,
+        current_time: Optional[datetime] = None,
+        dynamic_partitions_store: Optional[DynamicPartitionsStore] = None,
+    ) -> bool:
+        if not isinstance(downstream_partitions_def, TimeWindowPartitionsDefinition):
```
OwenKephart, line 142: if all the callsites need to do this check, should we just put these checks inside _map_partitions?

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -815,8 +821,13 @@ def should_reconcile(
             return False
 
         if all(
-            will_be_materialized_for_freshness(candidate)
-            or parents_will_be_reconciled(asset_graph, candidate)
+            asset_graph.partition_maps_to_valid_parents(
```
OwenKephart, line 824: This should be in the `can_reconcile_candidate` above instead. As is, if we're in this situation, we'll hit the "else" case, which means that the reason that this asset is skipped will be registered as "ParentOutdated", which would be inaccurate.

It's arguable that we should carve out a separate reason for this case, but it's rare enough that for now it's ok to just leave blank.

OwenKephart (changes requested):
Had some minor comments, but overall this looks great -- very into this direction

sryza (commented):
On board with this approach.  My main question: this ends up with a situation where TimeWindowPartitionMapping has a bunch of methods with similar implementations.

Might it make sense to change the signature of `get_upstream_partitions_for_partitions` so that it returns something like a `MappedTimeWindowPartitionsResult`? (But more generic).

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
clairelin135: good call

python_modules/dagster/dagster/_core/definitions/time_window_partition_mapping.py
clairelin135: good call, I've updated to the suggested notation. I prefer to still use the named tuple to keep the return objects unambiguous

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
clairelin135: oo good catch

clairelin135:
@sryza I think long term `get_upstream_partitions_for_partitions` should return a MappedTimeWindowPartitionsResult with the following objects, and not raise an error internally:
- resultant partitions subset that contains the filtered existent partition keys
- a list of nonexistent partition keys that were mapped to

This way, each method that calls `get_upstream_partitions_for_partitions` can choose to raise a more descriptive error message if nonexistent partition keys were mapped to
- for backfills and launching manual runs, we'll be able to raise an error with the targeted asset keys and invalid partition keys
- reconciliation methods can ignore the nonexistent partition keys

I'd prefer to break it out into a separate refactor PR though in the interest of ensuring we can get the fix out to discord. No other `get_upstream_partitions_for_partitions` methods currently map to invalid partitions, so updating the return type of the method doesn't seem imminently necessary

sryza:
> I'd prefer to break it out into a separate refactor PR though in the interest of ensuring we can get the fix out to discord

Ok, I agree with wanting to unblock users. I think we should make sure to follow up on this shortly afterwards so we don't leave the codebase in a more complicated state.

A possible middle-ground, which you can push back on if you think it will prevent us from being able to merge this in time:
1. In this PR, add the method to PartitionMapping that has the target signature you described in your last comment, and use that in the auto-materialize code paths already touched by this PR.
2. In a followup PR, update other call sites of the `get_upstream_partitions_for_partitions` to use the new method and get rid of the old one.

clairelin135:
@sryza I followed the approach you suggested above. This PR now uses a `PartitionMapping.get_upstream_mapped_partitions_result_for_partitions` method in the reconciliation logic. I will start working on a follow up to replace calls to `get_upstream_partitions_for_partitions` with this method instead.

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
sryza: Should this actually be in `parents_will_be_reconciled`? It seems like, from a logical perspective, finding out whether the parents are valid is part of finding out whether they're reconciled.

I'm imagining that eventually we'd want to get to a point where we just invoke the PartitionMapping once to find out what's up with the parent partitions: i.e. both what the parent partitions are whether any of the ones we need are non-existent, and then using that to answer the question of `parents_will_be_reconciled`.

clairelin135: I actually think that it makes sense for this check to exist in `can_reconcile_candidate`. It's checking whether the partition itself has the capacity to be materialized, not the parents

So I think in this case, it doesn't matter that the parents will be reconciled if the candidate itself can't be materialized

sryza: > It's checking whether the partition itself has the capacity to be materialized, not the parents

Whether the partition itself has the capacity to be materialized depends on the parents though, right?

The purpose of the `parents_will_be_reconciled` function is basically to make sure that the parents are in the state that's required to be able to materialize the child.

We might need to amend the name of `parents_will_be_reconciled` to something a little more general for this to make full sense.

clairelin135: I agree that there is benefit in being able to only fetch upstream info from the partition mapping once. 

This is tricky in this case though since the `parents_will_be_reconciled` function is used to determine if parents are outdated, displaying a corresponding message in the UI if so. Adding this check in the `parents_will_be_reconciled` function would then display a "parent outdated" message in the UI when the upstreams are invalid, which would be inaccurate

clairelin135: I'm happy to refactor the existing code to fetch the upstream info from the partition mapping once, and then propagate that information to `parents_will_be_reconciled`, but I would prefer to do that in the next PR in this stack since this change is slightly more involved (requires changing return types for methods on the asset graph)

OwenKephart (approved):
small nit, otherwise looks good!

PR #14707, opened 2023-06-08
Use typing.get_type_hints for input/output and resource/config type inference

### LINKED ISSUE #14571: Importing annotations from __future__ breaks Config

smackesey:
## Summary & Motivation

Fixes #14571 

Currently, `Parameter.annotation` is used when inferring the dagster type from an op compute function signature. This causes an error if either: (a) a string is used directly by the user for the parameter annotation; (b) `from __future__ import annotations` is used. This is because both cases cause the value of `Parameter.annotation` to be a string, and our dagster type resolution logic does not handle strings.

This PR replaces direct access to `Parameter.annotation` with `typing.get_type_hints`, which handles evaluation of a string annotation into an actual class which can be handled by our dagster type resolution logic. Note that this only works if the referents of a string annotation is module-level scope-- it will fail for a local scope.

Still this is an upgrade and will fix the problem in most cases. It will:

- Allows users to use string annotations or `from __future__ import annotations` together with dagster type signature inference
- Allows us to use `from __future__ import annotations` in our own code, which is necessary to allow Ruff to auto-sort typing-only imports-- see https://github.com/dagster-io/dagster/pull/14675. Not clear whether this is a good idea.

## How I Tested These Changes

New unit test using string annotations.

### DIFF
--- a/python_modules/dagster/dagster/_core/decorator_utils.py
+++ b/python_modules/dagster/dagster/_core/decorator_utils.py
@@ -1,8 +1,23 @@
+import functools
 import textwrap
 from inspect import Parameter, signature
-from typing import Any, Callable, Optional, Sequence, Set, TypeVar, Union
+from typing import (
+    Any,
+    Callable,
+    Mapping,
+    Optional,
+    Sequence,
+    Set,
+    TypeVar,
+    Union,
+)
 
-from typing_extensions import Concatenate, ParamSpec, TypeGuard
+from typing_extensions import (
+    Concatenate,
+    ParamSpec,
+    TypeGuard,
+    get_type_hints as typing_get_type_hints,
+)
 
 R = TypeVar("R")
 T = TypeVar("T")
@@ -34,6 +49,11 @@ def _is_param_valid(param: Parameter, expected_positional: str) -> bool:
 
 def get_function_params(fn: Callable[..., Any]) -> Sequence[Parameter]:
     return list(signature(fn).parameters.values())
+
+
+def get_type_hints(fn: Callable) -> Mapping[str, Any]:
+    target = fn.func if isinstance(fn, functools.partial) else fn
+    return typing_get_type_hints(target, include_extras=True)
 
 
 def validate_expected_params(
--- a/python_modules/dagster/dagster/_core/decorator_utils.py
+++ b/python_modules/dagster/dagster/_core/decorator_utils.py
@@ -1,4 +1,5 @@
 import functools
+import re
 import textwrap
 from inspect import Parameter, signature
 from typing import (
@@ -18,6 +19,8 @@ from typing_extensions import (
     TypeGuard,
     get_type_hints as typing_get_type_hints,
 )
+
+from dagster._core.errors import DagsterInvalidDefinitionError
 
 R = TypeVar("R")
 T = TypeVar("T")
@@ -53,7 +56,22 @@ def get_function_params(fn: Callable[..., Any]) -> Sequence[Parameter]:
 
 def get_type_hints(fn: Callable) -> Mapping[str, Any]:
     target = fn.func if isinstance(fn, functools.partial) else fn
-    return typing_get_type_hints(target, include_extras=True)
+
+    try:
+        return typing_get_type_hints(target, include_extras=True)
+    except NameError as e:
+        match = re.search(r"'(\w+)'", str(e))
+        assert match
+        annotation = match[1]
+        raise DagsterInvalidDefinitionError(
+            f'Failed to resolve type annotation "{annotation}" in function {target.__name__}. This'
+            " can occur when the parameter has a string annotation that references either: (1) a"
+            " type defined in a local scope (2) a type that is defined or imported in an `if"
+            " TYPE_CHECKING` block. Note that if you are including `from __future__ import"
+            " annotations`, all annotations in that module are stored as strings. Suggested"
+            " solutions include: (1) convert the annotation to a non-string annotation; (2) move"
+            " the type referenced by the annotation out of local scope or a `TYPE_CHECKING` block."
+        )
 
 
 def validate_expected_params(
--- a/python_modules/dagster/dagster/_core/definitions/inference.py
+++ b/python_modules/dagster/dagster/_core/definitions/inference.py
@@ -1,6 +1,14 @@
 from inspect import Parameter, Signature, isgeneratorfunction, signature
-from typing import Any, Callable, Mapping, NamedTuple, Optional, Sequence
+from typing import (
+    Any,
+    Callable,
+    Mapping,
+    NamedTuple,
+    Optional,
+    Sequence,
+)
 
+from dagster._core.decorator_utils import get_type_hints
 from dagster._seven import is_module_available
 
 from .utils import NoValueSentinel
@@ -55,11 +63,12 @@ def _infer_output_description_from_docstring(fn: Callable) -> Optional[str]:
 
 
 def infer_output_props(fn: Callable) -> InferredOutputProps:
-    sig = signature(fn)
-
-    annotation = Parameter.empty
-    if not isgeneratorfunction(fn):
-        annotation = sig.return_annotation
+    type_hints = get_type_hints(fn)
+    annotation = (
+        type_hints["return"]
+        if not isgeneratorfunction(fn) and "return" in type_hints
+        else Parameter.empty
+    )
 
     return InferredOutputProps(
         annotation=annotation,
@@ -74,6 +83,7 @@ def has_explicit_return_type(fn: Callable) -> bool:
 
 def _infer_inputs_from_params(
     params: Sequence[Parameter],
+    type_hints: Mapping[str, object],
     descriptions: Optional[Mapping[str, Optional[str]]] = None,
 ) -> Sequence[InferredInputProps]:
     _descriptions: Mapping[str, Optional[str]] = descriptions or {}
@@ -82,14 +92,14 @@ def _infer_inputs_from_params(
         if param.default is not Parameter.empty:
             input_def = InferredInputProps(
                 param.name,
-                param.annotation,
+                type_hints.get(param.name, param.annotation),
                 default_value=param.default,
                 description=_descriptions.get(param.name),
             )
         else:
             input_def = InferredInputProps(
                 param.name,
-                param.annotation,
+                type_hints.get(param.name, param.annotation),
                 description=_descriptions.get(param.name),
             )
 
@@ -101,7 +111,8 @@ def _infer_inputs_from_params(
 def infer_input_props(fn: Callable, context_arg_provided: bool) -> Sequence[InferredInputProps]:
     sig = signature(fn)
     params = list(sig.parameters.values())
+    type_hints = get_type_hints(fn)
     descriptions = _infer_input_description_from_docstring(fn)
     params_to_infer = params[1:] if context_arg_provided else params
-    defs = _infer_inputs_from_params(params_to_infer, descriptions=descriptions)
+    defs = _infer_inputs_from_params(params_to_infer, type_hints, descriptions=descriptions)
     return defs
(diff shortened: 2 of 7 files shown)


smackesey:
* master
  * **PR #14707** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14707" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14707?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/inference.py
```diff
@@ -55,11 +63,12 @@ def _infer_output_description_from_docstring(fn: Callable) -> Optional[str]:
...
-    sig = signature(fn)
-
-    annotation = Parameter.empty
-    if not isgeneratorfunction(fn):
-        annotation = sig.return_annotation
+    type_hints = get_type_hints(fn)
+    annotation = (
+        type_hints["return"]
+        if not isgeneratorfunction(fn) and "return" in type_hints
+        else Parameter.empty
```
alangenfeld, line 70: whats going on here?

python_modules/dagster/dagster/_core/decorator_utils.py
```diff
@@ -36,6 +51,11 @@ def get_function_params(fn: Callable[..., Any]) -> Sequence[Parameter]:
     return list(signature(fn).parameters.values())
 
 
+def get_type_hints(fn: Callable) -> Mapping[str, Any]:
+    target = fn.func if isinstance(fn, functools.partial) else fn
```
alangenfeld, line 55: is this just to deal with a difference in behavior between `typing_get_type_hints` and  `inspect`?

python_modules/dagster/dagster/_core/definitions/source_asset.py
```diff
@@ -187,6 +178,9 @@ def _get_op_def_compute_fn(self, observe_fn: SourceAssetObserveFunction):
             DecoratedOpFunction,
             is_context_provided,
         )
+        from dagster._core.execution.context.compute import (
```
alangenfeld, line 181: is this file an unrelated change?

python_modules/dagster/dagster_tests/core_tests/resource_tests/pythonic_resources/test_binding.py
```diff
@@ -585,3 +585,28 @@ def my_asset(context, my_resource: MyResourceWithDefault):
...
+
+
+class MyModuleLevelResource(ConfigurableResource):
+    str_field: str
+
+
+# Note that an explicit string annotation has the same effect as defining a resource in a module
+# using `from __future__ import annotations`. This test will only work against a module-scoped
+# resource-- this is a hard limitation of string annotations in Python as of 2023-07-06 and Python
+# 3.11.
```
alangenfeld, line 597: what does the error look like when you use `from __future__ import annotations` with non module scope type? or i guess just a typo-ed string ?

python_modules/dagster/dagster/_core/definitions/inference.py
```diff
@@ -101,7 +111,8 @@ def _infer_inputs_from_params(
 def infer_input_props(fn: Callable, context_arg_provided: bool) -> Sequence[InferredInputProps]:
     sig = signature(fn)
     params = list(sig.parameters.values())
+    type_hints = get_type_hints(fn)
     descriptions = _infer_input_description_from_docstring(fn)
     params_to_infer = params[1:] if context_arg_provided else params
-    defs = _infer_inputs_from_params(params_to_infer, descriptions=descriptions)
+    defs = _infer_inputs_from_params(params_to_infer, type_hints, descriptions=descriptions)
```
alangenfeld, line 117: unfortunate how this similar but slightly different information flows through from inspect and typing

alangenfeld (approved):
seems strictly better than current state, unless `typing_get_type_hints` throws in cases not under test that currently work for users which seems unlikely

python_modules/dagster/dagster/_core/definitions/inference.py
smackesey: This is the same logic as before but pulling the annotation using `get_type_hints` instead of `inspect.signature`.

python_modules/dagster/dagster/_core/definitions/inference.py
smackesey: indeed, the python APIs around type annotations are a bit of a mess

python_modules/dagster/dagster/_core/definitions/source_asset.py
smackesey: No. The old code fails with the new method for subtle reasons. When `OpExecutionContext` is defined in a `TYPE_CHECKING` block, it does not exist at runtime. When accessing this annotation through `inspect.signature`, we get a string `"OpExecutionContext"` with no attempt to evaluate because `OpExecutionContext` does not correspond to an input/output/config arg, so it does not fail.

`typing.get_type_hints` attempts to evaluate `OpExecutionContext`, and raises a `NameError`. We avoid this by making it an actual type.

This _is_ technically breaking for users who have done something similar (annotated a context arg with a string referring to a type that is only imported in a `TYPE_CHECKING` block), although it seems unlikely people are doing this since string annotations with ops/asset params will already cause breakage most of the time.

That said I could add some code to catch this error and fall back to old behavior.

python_modules/dagster/dagster_tests/core_tests/resource_tests/pythonic_resources/test_binding.py
smackesey: I tried to explain the situation in detail here: https://github.com/dagster-io/dagster/pull/14675

But here is an excerpt (you can mentally sub out the quoted annotations `"Foo"` and `"Bar"` with `from __future__ import annotations`-- they have the same effect):

```
from typing import get_type_hints

class Foo:
    ...

def foo(x: "Foo"):
     ...

# works to dereference "Foo" since referent `Foo` is in module-level scope
get_type_hints(foo)  # => {'x': <class '__main__.Foo'>}

def make_bar():
    class Bar:
        ...

    def bar(x: "Bar"):
        ...
    return bar

bar = make_bar()

# does not work to dereference "Bar" since referent `Bar` vanished with local scope
get_type_hints(bar)  # => NameError: name 'Bar' is not defined.
```

python_modules/dagster/dagster/_core/definitions/source_asset.py
alangenfeld: Even if its unlikely its seems like a very terrible error experience so seems worth avoiding. Lets get this case explicitly under test to demonstrate how it errors. 

Should we just try to do the string lookup in globals for the type when we end up with a string from `inspect` instead of eagerly running `get_type_hints`?

python_modules/dagster/dagster_tests/core_tests/resource_tests/pythonic_resources/test_binding.py
alangenfeld: I was just asking to figure out what the end user experience would be for someone hitting this corner

python_modules/dagster/dagster/_core/definitions/source_asset.py
alangenfeld: Its hard to concoct scenarios where an end user would hit the `OpExecutionContext` problem specifically, but it does seem worth avoiding introducing a known breakage

smackesey: >Should we just try to do the string lookup in globals for the type when we end up with a string from inspect instead of eagerly running get_type_hints?

I think `globals()` would be insufficient because it returns the dict for the _current_ module namespace. We need access to the dict for the module namespace where the signature we're processing is written-- `get_type_hints` must access this under the hood somehow.

I think `try`-`except` block that catches `NameError` and falls back to the old behavior if `get_type_hints` fails is the way to go. Note that this will still make code like this (which crashes now) crash:

```
from __future__ import annotations

from dagster import op
from some_lib import SomeType

from typing import TYPE_CHECKING:
     from dagster import OpExecutionContext

@op
def foo(context: OpExecutionContext, bar: SomeType):
    ...
```

This will fail because, when we fall back to the old behavior, `SomeType` will no longer be dereferenced and it will be fed to Dagster machinery as a string.

alangenfeld: got it - if its possible to throw a better error that would guide users in a helpful direction that would be great

python_modules/dagster/dagster/_core/decorator_utils.py
smackesey: yes, `get_type_hints` chokes on `partial`, `inspect.signature` doesn't

python_modules/dagster/dagster/_core/definitions/source_asset.py
smackesey: OK, I pushed a commit that catches `NameError` in `get_type_hints` and gives a thorough error message. Given this error message and the likely extreme rarity of the situation that fallback logic would fix, I don't think it's worth adding that logic.

To spell it out, for the fallback logic to save the user:

1. they would have to  have _only_ the context argument be a string and no other input/output config (which is impossible using `from __future__ import annotations`)
2. have the context type import in a `TYPE_CHECKING` block or local scope

This is very unlikely because the user is almost always going to have other runtime imports from dagster like `op`/`asset`.

PR #14578, opened 2023-06-01
feat(dbt-assets): automatically apply description replacements using the manifest

rexledesma:
## Summary & Motivation
Some users want to modify the description of their dbt assets to include information like the dbt model's compiled code, which was supported previously using the `display_raw_sql` argument.

Add it back.

## How I Tested These Changes
pytest, local

PR #14577, opened 2023-06-01
feat: allow description overrides by asset key

rexledesma:
## Summary & Motivation
Overriding `descriptions_by_key` was missing in `.with_attributes`. Add it.

## How I Tested These Changes
pytest

PR #14718, opened 2023-06-08
feat(dbt-assets): automatically apply metadata replacements using the manifest

slopp:
## Summary & Motivation

The original dbt load functions accepted metadata from two places, the dbt manifest (how we show column schemas) and from a user supplied function `node_info_to_definition_metadata_fn`. This latter was particularly useful  in order for partitioned dbt assets to specify a `partition_expr` metadata attribute for attached io managers. 

This PR unblocks this use case for the new asset decorator, but instead of a helper function, it prefers setting metadata directly on the asset decorator and then merging that metadata with any found in the dbt manifest.

## How I Tested These Changes

I tested this on hooli and added unit tests

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorator.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorator.py
@@ -74,6 +74,7 @@ def dbt_assets(
             input_asset_key_replacements=manifest.asset_key_replacements,
             output_asset_key_replacements=manifest.asset_key_replacements,
             descriptions_by_key=manifest.descriptions_by_asset_key,
+            metadata_by_key=manifest.metadata_by_asset_key,
         )
 
     return inner
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
@@ -23,7 +23,12 @@ from dbt.node_types import NodeType
 from pydantic import Field
 from typing_extensions import Literal
 
-from ..asset_utils import default_asset_key_fn, default_description_fn, output_name_fn
+from ..asset_utils import (
+    default_asset_key_fn,
+    default_description_fn,
+    default_metadata_fn,
+    output_name_fn,
+)
 
 logger = get_dagster_logger()
 
@@ -71,6 +76,10 @@ class DbtManifest:
     def node_info_to_description(cls, node_info: Mapping[str, Any]) -> str:
         return default_description_fn(node_info)
 
+    @classmethod
+    def node_info_to_metadata(cls, node_info: Mapping[str, Any]) -> Mapping[str, Any]:
+        return default_metadata_fn(node_info)
+
     @property
     def node_info_by_asset_key(self) -> Mapping[AssetKey, Mapping[str, Any]]:
         """A mapping of the default asset key for a dbt node to the node's dictionary representation in the manifest.
@@ -101,6 +110,15 @@ class DbtManifest:
         """
         return {
             self.node_info_to_asset_key(node): self.node_info_to_description(node)
+            for node in self.node_info_by_dbt_unique_id.values()
+        }
+
+    @property
+    def metadata_by_asset_key(self) -> Mapping[AssetKey, Mapping[Any, str]]:
+        """A mapping of the default asset key for a dbt node to the node's metadata in the manifest.
+        """
+        return {
+            self.node_info_to_asset_key(node): self.node_info_to_metadata(node)
             for node in self.node_info_by_dbt_unique_id.values()
         }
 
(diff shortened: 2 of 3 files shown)


python_modules/libraries/dagster-dbt/dagster_dbt/asset_utils.py
```diff
@@ -193,9 +195,9 @@ def get_dbt_multi_asset_args(
             key=asset_key,
             dagster_type=Nothing,
             io_manager_key=io_manager_key,
+            metadata=metadata,
             description=default_description_fn(node_info, display_raw_sql=False),
             is_required=False,
-            metadata=default_metadata_fn(node_info),
```
slopp, line 198: Originally metadata was sourced from dbt manifest, but I think this metadata was actually dropped from the final asset and not passed through.

We could change the approach in this PR and instead of using metadata from the decorator source it from the manifest. The original dbt decorator included both options, essentially, by providing a function `node_info_to_definition_metadata_fn`. I'd prefer not to add that here, but we could I guess combine metadata from the decorator with metadata from the manifest automatically.

rexledesma (changes requested):
From our DMs, I think this can look more like:

https://github.com/dagster-io/dagster/pull/14577
https://github.com/dagster-io/dagster/pull/14578

I can make this change.

The user should have one place to set this sort of metadata. Also, it should be up to the user whether the behavior should to merge the metadata with the default metadata, or completely override it.

rexledesma:
* master
  * **PR #14718** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14718" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14718?utm_source=stack-comment).

slopp:
The latest build applied to a hooli branch using asset decorators:

```python
class CustomizedDbtManifest(DbtManifest):

    @classmethod
    def node_info_to_metadata(cls, node_info: Mapping[str, Any]) -> Mapping[str, Any]:
            metadata = {"partition_expr": "orders_date"}
           
            if node_info['name'] == 'orders_cleaned':
                metadata = {"partition_expr": "dt"}
            
            if node_info['name'] == 'users_cleaned':
                metadata = {"partition_expr": "created_at"}

            return metadata   

    @classmethod
    def node_info_to_asset_key(cls, node_info: Mapping[str, Any]) -> AssetKey:
        
        node_path = node_info['path']
        prefix = node_path.split('/')[0]
        
        if node_path == 'models/sources.yml':
            prefix = "RAW_DATA"
        
        return AssetKey([prefix, node_info['name']])
    
    @classmethod
    def node_info_to_description(cls, node_info: Mapping[str, Any]) -> str:
        description_sections = []

        description = node_info.get("description", "")
        if description:
            description_sections.append(description)

        raw_code = node_info.get("raw_code", "")
        if raw_code:
            description_sections.append(f"#### Raw SQL:\n```\n{raw_code}\n```")

        return "\n\n".join(description_sections)

manifest = CustomizedDbtManifest.read(path=DBT_MANIFEST)

def _process_partitioned_dbt_assets(context: OpExecutionContext, dbt2: DbtCli):
     # map partition key range to dbt vars
    first_partition, last_partition = context.asset_partitions_time_window_for_output(list(context.selected_output_names)[0])
    dbt_vars = {"min_date": str(first_partition), "max_date": str(last_partition)}
    dbt_args = ["run", "--vars", json.dumps(dbt_vars)]

    dbt_cli_task = dbt2.cli(dbt_args, manifest=manifest, context=context)

    dbt_events = list(dbt_cli_task.stream_raw_events())

    for event in dbt_events:
        # add custom metadata to the asset materialization event 
        context.log.info(event)
        for dagster_event in event.to_default_asset_events(manifest=manifest):
            
            if isinstance(dagster_event, Output):
                event_node_info = event.event["data"]["node_info"]

                started_at = parser.isoparse(event_node_info["node_started_at"])
                completed_at = parser.isoparse(event_node_info["node_finished_at"])

                metadata = {
                    "Execution Started At": started_at.isoformat(timespec="seconds"),
                    "Execution Completed At": completed_at.isoformat(timespec="seconds"),
                    "Execution Duration": (completed_at - started_at).total_seconds(),
                    "Owner": "data@hooli.com"
                }

                context.add_output_metadata(
                    metadata=metadata,
                    output_name=dagster_event.output_name,
                )

            yield dagster_event
    
    if not dbt_cli_task.is_successful():
        raise Exception('dbt command failed, see preceding events')
    

@dbt_assets(
        manifest=manifest,
        select="orders_cleaned users_cleaned orders_augmented", 
        partitions_def=daily_partitions,
)
def daily_dbt_assets(context: OpExecutionContext, dbt2: DbtCli):
    yield from _process_partitioned_dbt_assets(context=context, dbt2=dbt2)
````

Screen Shot 2023-06-13 at 8 58 33 AM

Interestingly, the default metadata function is still be applied to the dbt assets - see how `users_cleaned` has both `partition_expr` and `table_schema`?  I think this is fine, but not quite what I expected.

OwenKephart (approved):

PR #14725, opened 2023-06-09
Rename dagit to dagster-webserver

smackesey:
## Summary & Motivation

- New `dagster-webserver` Python package.
    - Package renames all uses of `dagit` to either "webserver" or "dagster-webserver", depending on context.
    - A few options in the CLI had the word "dagit"-- these were renamed to use "webserver", but the "dagit" variants were preserved as aliases for backcompat (removal in 2.0)
    - A few webserver routes (`/dagit_info` and `/dagit/notebook`) also included "dagit"-- these renamed to "webserver" equivalents, with the "dagit" variants preserved as aliases.
- `dagit` package is now a stub package.
    - Includes `dagster-webserver` as a dependency.
    - Installing the `dagit` package still installs a `dagit` executable, but the executable points to the CLI defined in `dagster-webserver`.
    - All tests transferred to `dagster-webserver` except one that invokes the `dagit` executable (and checks that `dagster-webserver` executable is invoked.
    - Invoking the `dagit` executable will cause a deprecation warning to be printed advising use of `dagster-webserver` or `dagster dev`.
- Updates to `dagster`
	- Rename `dagster dev` options `--dagit-port` and `--dagit-host` to `--webserver-port` and `--webserver-host`, preserving the old names as deprecated aliases
- Updates to `dagster-graphql`
    - Root query type renamed from "DagitQuery" -> "WebserverQuery"
- Other
    - Packages and test suites that depended on `dagit` now depend on `dagster-webserver`
    - Change new library template to include `dagster-webserver` as dep instead of `dagit`
    - I preserved the telemetry code `"start_dagit_webserver"` (but renamed the constant referencing it) to avoid disrupting data collection.
    - Makefile: `rebuild_dagit` -> `rebuild_ui`

## How I Tested These Changes

- Existing test suite
- New unit tests to ensure URL and CLI redirects work
- Fresh `make dev_install` on this branch, ensured the following work (don't crash and you can visit localhost:
    - `dagster dev -f ...`
    - `dagit -f ...`
    - `dagster-webserver -f ...`


smackesey:
* master
  * **PR #14725** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14725" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14725?utm_source=stack-comment).

python_modules/dagster-graphql/dagster_graphql/schema/roots/mutation.py
```diff
@@ -748,11 +748,11 @@ def mutate(self, graphene_info, runId: str):
         return True
 
 
-class GrapheneDagitMutation(graphene.ObjectType):
+class GrapheneWebserverMutation(graphene.ObjectType):
     """The root for all mutations to modify data in your Dagster instance."""
 
     class Meta:
-        name = "DagitMutation"
+        name = "WebserverMutation"
```
alangenfeld, line 755: lets pull these GraphQL schema renames out to reduce risk - they are likely fine but im not 100% certain. We probably just want to call them `Mutation`, `Query`, etc - can discuss that in its own PR

python_modules/dagster-webserver/dagster_webserver/cli.py
```diff
@@ -26,35 +27,32 @@
...
-def create_dagit_cli():
-    return dagit
+def create_dagster_webserver_cli():
+    return dagster_webserver
 
 
-DEFAULT_DAGIT_HOST = "127.0.0.1"
-DEFAULT_DAGIT_PORT = 3000
+DEFAULT_WEBSERVER_HOST = "127.0.0.1"
+DEFAULT_WEBSERVER_PORT = 3000
```
alangenfeld, line 35: nit: "webserver" feels redundant here

python_modules/dagster-webserver/dagster_webserver/cli.py
```diff
@@ -80,7 +78,7 @@ def create_dagit_cli():
     "-l",
     type=click.STRING,
     default="",
-    help="The path prefix where Dagit will be hosted (eg: /dagit)",
+    help="The path prefix where webserver will be hosted (eg: /dagster-webserver)",
```
alangenfeld, line 81: [1]

python_modules/dagster-webserver/dagster_webserver/cli.py
```diff
@@ -106,14 +104,14 @@ def create_dagit_cli():
 @click.option(
     "--read-only",
     help=(
-        "Start Dagit in read-only mode, where all mutations such as launching runs and "
+        "Start webserver in read-only mode, where all mutations such as launching runs and "
```
alangenfeld, line 107: [1]

python_modules/dagster-webserver/dagster_webserver/cli.py
```diff
@@ -106,14 +104,14 @@ def create_dagit_cli():
...
         "turning schedules on/off are turned off."
     ),
     is_flag=True,
 )
 @click.option(
     "--suppress-warnings",
-    help="Filter all warnings when hosting Dagit.",
+    help="Filter all warnings when hosting webserver.",
```
alangenfeld, line 114: [1]

python_modules/dagster-webserver/dagster_webserver/cli.py
```diff
@@ -127,7 +125,7 @@ def create_dagit_cli():
 )
 @click.option(
     "--code-server-log-level",
-    help="Set the log level for any code servers spun up by dagit.",
+    help="Set the log level for any code servers spun up by the webserver.",
```
alangenfeld, line 128: [1] "the webserver" feels fines but the others above do not

python_modules/dagster-webserver/dagster_webserver/cli.py
```diff
@@ -230,9 +234,9 @@ def host_dagit_ui_with_workspace_process_context(
         )
 
 
-cli = create_dagit_cli()
+cli = create_dagster_webserver_cli()
 
 
 def main():
     # click magic
-    cli(auto_envvar_prefix="DAGIT")  # pylint:disable=E1120
```
alangenfeld, line 238: [2]

python_modules/dagit/dagit_tests/test_cli.py
```diff
@@ -1,27 +1,10 @@
...
-
-
-def test_invoke_cli_wrapper_with_invalid_option():
-    process = subprocess.Popen(["dagit", "-d", "."], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    _, stderr = process.communicate()
-    assert process.returncode != 0
-    assert (
-        b"Error: Invalid set of CLI arguments for loading repository/job. See --help for details.\n"
-        in stderr
+def test_invoke_cli():
```
alangenfeld, line 4: [2] we should test for the click auto env var thing with `DAGIT_` which i believe this PR currently breaks

python_modules/dagster-webserver/dagster_webserver/cli.py
```diff
@@ -26,35 +27,32 @@
...
-            "\n\nOptions can also provide arguments via environment variables prefixed with DAGIT"
-            "\n\nFor example, DAGIT_PORT=3333 dagit"
+            "\n\nExamples:\n\n1. dagster-webserver (works if .{default_filename} exists)\n\n2."
+            " dagster-webserver -w path/to/{default_filename}\n\n3. dagster-webserver -f"
+            " path/to/file.py\n\n4. dagster-webserver -f path/to/file.py -d"
+            " path/to/working_directory\n\n5. dagster-webserver -m some_module\n\n6."
+            " dagster-webserver -f path/to/file.py -a define_repo\n\n7. dagster-webserver -m"
+            " some_module -a define_repo\n\n8. dagster-webserver -p 3333\n\nOptions can also"
+            " provide arguments via environment variables prefixed with DAGSTER_WEBSERVER\n\nFor"
+            " example, DAGSTER_WEBSERVER_PORT=3333 dagster-webserver"
```
alangenfeld, line 55: nit: this seems much harder to edit now. We should probably drop the "repo" language

python_modules/dagster-webserver/dagster_webserver/webserver.py
```diff
@@ -91,16 +91,16 @@ def make_csp_header(self, nonce: str) -> str:
             raise Exception(
                 """
                 CSP configuration file could not be found.
-                If you are using dagit, then probably it's a corrupted installation or a bug.
-                However, if you are developing dagit locally, your problem can be fixed by running
-                "make rebuild_dagit" in the project root.
+                If you are using dagster-webserver, then probably it's a corrupted installation or a bug.
+                However, if you are developing dagster-webserver locally, your problem can be fixed by running
+                "make rebuild_dagster_ui" in the project root.
```
alangenfeld, line 96: just `rebuild_ui` right

python_modules/dagster-webserver/dagster_webserver/webserver.py
```diff
@@ -274,7 +274,9 @@ def build_static_routes(self):
     def build_routes(self):
         routes = (
             [
-                Route("/dagit_info", self.dagit_info_endpoint),
+                Route("/dagster-webserver-info", self.webserver_info_endpoint),
```
alangenfeld, line 277: nit: maybe `server_info` ? this feels like a lot

python_modules/dagster-webserver/dagster_webserver/webserver.py
```diff
@@ -298,6 +300,11 @@ def build_routes(self):
                     "/logs/{path:path}",
                     self.download_captured_logs_endpoint,
                 ),
+                Route(
+                    "/dagster-webserver/notebook",
```
alangenfeld, line 304: dagster-webserver prefix feels odd / redundant here

python_modules/dagster/dagster/_cli/dev.py
```diff
@@ -57,22 +57,34 @@ def dev_command_options(f):
...
-@click.option("--dagit-host", "-h", help="Host to use for the Dagit UI.", required=False)
+@click.option(
+    "--webserver-port",
+    "--dagit-port",
+    "-p",
+    help="Port to use for the Dagster webserver.",
+    required=False,
+)
+@click.option(
+    "--webserver-host",
```
alangenfeld, line 68: should just be `port` and `host` i think

alangenfeld (changes requested):

python_modules/dagster-graphql/dagster_graphql/schema/roots/mutation.py
smackesey: done

python_modules/dagster-webserver/dagster_webserver/cli.py
smackesey: I would agree if these were module-local but they are imported elsewhere so I think "WEBSERVER" is a useful qualifier.

python_modules/dagster-webserver/dagster_webserver/cli.py
smackesey: Yeah I'm actually not even sure how this reformat happened. I've now changed it to use a """ string with `textwrap.dedent`, which improves readability/editability of original.

python_modules/dagster-webserver/dagster_webserver/webserver.py
smackesey: good catch

python_modules/dagster-webserver/dagster_webserver/webserver.py
smackesey: yeah good idea

python_modules/dagster-webserver/dagster_webserver/webserver.py
smackesey: I agree. I was just cargo-culting the older format (/dagit/notebook), not knowing if it was like that for a reason. But if you don't think so, I'll drop it.

python_modules/dagster/dagster/_cli/dev.py
smackesey: ye

python_modules/dagit/dagit_tests/test_cli.py
smackesey: Ah yeah my bad I forgot to handle this. Added backcompat and test now.

python_modules/dagit/setup.py
```diff
@@ -44,15 +44,18 @@ def get_version():
...
         f"dagster{pin}",
+        f"dagster-webserver{pin}",
         f"dagster-graphql{pin}",
-        "starlette",
-        "uvicorn[standard]",
     ],
     extras_require={
         "notebook": ["nbconvert"],  # notebooks support
         "test": ["starlette[full]"],  # TestClient deps in full
     },
```
alangenfeld, line 54: should this only depend on  `dagster-webserver{pin}` ? 

and 
below just point at `dagster-webserver[full]` / `dagster-webserver[test]`

alangenfeld (commented):
whats the roll-out plan for this change? I assume we want to do some vetting ahead of the normal release?

python_modules/dagster-webserver/dagster_webserver/cli.py
```diff
@@ -156,10 +169,16 @@ def dagit(
         os.environ["PYTHONWARNINGS"] = "ignore"
 
     configure_loggers()
-    logger = logging.getLogger("dagit")
+    logger = logging.getLogger("dagster-webserver")
```
alangenfeld, line 172: hmm, we may want to keep this logger name locked with the entry ie `dagit`, may break users log monitoring if we change it implicitly

python_modules/dagit/setup.py
smackesey: Yes-- except `dagster-webserver[notebook]` and `dagster-webserver[test]` (there is no `dagster-webserver[full]`.

python_modules/dagster-webserver/dagster_webserver/cli.py
smackesey: >keep this logger name locked with the entry ie dagit

What exactly do you mean by "name locked with the entry"?

If it's important to maintain the dagit logger, I see a couple options:

- Define both `dagster-webserver` and `dagit` loggers. Forward messages from `dagster-webserver` to `dagit`, and suppress `dagit` output when running `dagster dev`. This will create duplicate log messages in other channels.
- Define only `dagster-webserver` logger, but introduce custom formatting so that the log messages include "dagit" also-- something like:

```
2023-07-03 08:27:36 -0400 - dagster-webserver (dagit) - INFO - Serving dagster-webserver on http://127.0.0.1:3000 in process 78370
```

Just plain keeping it as "dagit" I think would be confusing to users who install "dagster-webserver" but then see "dagit" log messages.

alangenfeld: I just meant if a user invokes `dagit`  we name the logger `dagit` , similar to how we support `DAGIT_` env vars for that entry point

smackesey: OK, I just pushed a commit that modulates the logger name based on CLI command

.gitignore
```diff
@@ -118,8 +118,9 @@ dagit_run_logs
 
 python_modules/libraries/dagster-aws/dagster_aws/ecs/config.yaml
 
-python_modules/dagit/node_modules/
-python_modules/dagit/yarn.lock
```
alangenfeld, line 122: any harm in keeping these for now? I imagine that having these large dirty directories become no longer ignored may result some annoyance / accidents

alangenfeld (approved):
alright lets go

.gitignore
smackesey: good point

PR #14738, opened 2023-06-09
Use new upstream partition mapping method

clairelin135:
This PR follows up from https://github.com/dagster-io/dagster/pull/14706 and switches internal callsites that call `get_upstream_partitions_from_partitions` to the new `get_upstream_mapped_partitions_result_for_partitions` method.

The new `get_upstream_mapped_partitions_result_for_partitions` method returns an object with two fields: the valid parent partitions subset and invalid parent partition keys. This object must be handled at callsites:
- In reconciliation, do not materialize candidates if invalid parent partitions exist. 
- In backfills and manual runs, raise an error with the invalid parent partition keys. Previously the error was raised within the `get_upstream_partitions_from_partitions` method.

In the next PR, will remove these old functions from partition mappings:
- `get_upstream_partitions_from_partitions`
- `get_upstream_partitions_for_partition_range`

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -33,7 +33,7 @@ from .assets import AssetsDefinition
 from .events import AssetKey, AssetKeyPartitionKey
 from .freshness_policy import FreshnessPolicy
 from .partition import PartitionsDefinition, PartitionsSubset
-from .partition_mapping import MappedPartitionsResult, PartitionMapping, infer_partition_mapping
+from .partition_mapping import PartitionMapping, UpstreamPartitionsResult, infer_partition_mapping
 from .source_asset import SourceAsset
 from .time_window_partitions import (
     TimeWindowPartitionsDefinition,
@@ -46,9 +46,17 @@ if TYPE_CHECKING:
     from dagster._core.definitions.asset_graph_subset import AssetGraphSubset
 
 
-class ParentPartitionsResult(NamedTuple):
-    valid_parent_partitions: Set[AssetKeyPartitionKey]
-    invalid_parent_partitions: Set[AssetKeyPartitionKey]
+class ParentsPartitionsResult(NamedTuple):
+    """Represents the result of mapping an asset partition to its upstream parent partitions.
+
+    parent_partitions (AbstractSet[AssetKeyPartitionKey]): The existent parent partitions that are
+        upstream of the asset partition. Filters out nonexistent partitions.
+    required_but_nonexistent_parents_partitions (AbstractSet[AssetKeyPartitionKey]): The required
+        upstream asset partitions that were mapped to but do not exist.
+    """
+
+    parent_partitions: AbstractSet[AssetKeyPartitionKey]
+    required_but_nonexistent_parents_partitions: AbstractSet[AssetKeyPartitionKey]
 
 
 class AssetGraph:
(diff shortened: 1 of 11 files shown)


python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -46,6 +46,11 @@
     from dagster._core.definitions.asset_graph_subset import AssetGraphSubset
 
 
+class ParentPartitionsResult(NamedTuple):
+    valid_parent_partitions: Set[AssetKeyPartitionKey]
```
sryza, line 50: Can these be `AbstractSet`?

python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -46,6 +46,11 @@
     from dagster._core.definitions.asset_graph_subset import AssetGraphSubset
 
 
+class ParentPartitionsResult(NamedTuple):
```
sryza, line 49: This needs docstring

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -673,7 +674,9 @@ def determine_asset_partitions_to_auto_materialize(
     ] = defaultdict(set, conditions_by_asset_partition_for_freshness)
 
     # a filter for eliminating candidates
-    def can_reconcile_candidate(candidate: AssetKeyPartitionKey) -> bool:
+    def can_reconcile_candidate(
+        candidate: AssetKeyPartitionKey,
```
sryza, line 678: This change seems spurious?

sryza (commented):
I'm giving the concepts introduced in the earlier PR a more thorough review this time.  

Looking at `MappedPartitionsResult`, I wonder if we can change the names in a way that avoids introducing new terminology:
- `invalid_partitions_mapped_to` -> `required_but_nonexistent_partition_keys`.
- `MappedPartitionsResult` -> `UpstreamPartitionsResult`, unless we think there's some notion of nonexistent / invalid that's meaningful in the downstream case?

Then for the constructs in this PR:
- `ParentPartitionsResult` -> `ParentsPartitionsResult` (parents is plural to match the naming of `get_parents_partitions`, where parents is plural to indicate that there are multiple assets)
- `valid_parent_partitions` -> `parents_partitions`
- `invalid_parent_partitions` -> `required_but_nonexistent_parents_partitions`

Combined, these would avoid introducing a new concept of a "valid / invalid" partition.

clairelin135:
@sryza I've updated the naming following your suggestions.

Initially I named the `MappedPartitionsResult` to be more general and not upstream/downstream specific in case if we did want to raise an error in the downstream case in the future. But there aren't instances where this would happen currently--for now I've renamed to `UpstreamPartitionsResult` and we can rename later if needed

python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -408,17 +365,21 @@ def get_parent_partition_keys_for_child(
...
-            if partition_key
-            else None,
-            upstream_partitions_def=parent_partitions_def,
-            dynamic_partitions_store=dynamic_partitions_store,
-            current_time=current_time,
+
+        parent_partitions_result = (
+            partition_mapping.get_upstream_mapped_partitions_result_for_partitions(
+                cast(PartitionsDefinition, child_partitions_def)
+                .empty_subset()
```
sryza, line 372: nitpick: can use `subset_with_partition_keys`

python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -408,17 +365,21 @@ def get_parent_partition_keys_for_child(
...
-            cast(PartitionsDefinition, child_partitions_def)
-            .empty_subset()
-            .with_partition_keys([partition_key])
-            if partition_key
-            else None,
-            upstream_partitions_def=parent_partitions_def,
-            dynamic_partitions_store=dynamic_partitions_store,
-            current_time=current_time,
+
+        parent_partitions_result = (
```
sryza, line 369: nitpick: could just put the `return` statement here

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -807,25 +803,41 @@ def conditions_for_candidate(
 
         return conditions
 
+    def can_materialize_candidate(
+        candidate: AssetKeyPartitionKey, parent_partitions_result: ParentsPartitionsResult
+    ) -> bool:
+        return len(parent_partitions_result.required_but_nonexistent_parents_partitions) == 0
```
sryza, line 809: can we just inline this? if we want it to be a little shorter, we could put an `all_required_partitions_exist` method `ParentsPartitionsResult`

sryza (approved):
Looks good aside from a few small remaining comments!

PR #13615, opened 2023-04-13
Skip assets_dbt_python and assets_modern_data_stack in bk

johannkm:
https://elementl-workspace.slack.com/archives/C03A0D72A6T/p1681416178005189?thread_ts=1681399769.236979&cid=C03A0D72A6T

We can move these tests to internal if they need editable dagster-cloud, or wait for the dagster-cloud package to move to oss. But let's disable for now to get builds green

PR #14740, opened 2023-06-09
[example] fix quickstart_aws post resource migration

yuhan:

PR #14741, opened 2023-06-09
RFC: [example] make sure tests for quickstart-* run

yuhan:
## Summary & Motivation
add `tox.ini` files to `quickstart-*` to make sure unit tests are running in bk, because we only include packages with tox.ini in bk runs (https://github.com/dagster-io/dagster/blob/master/.buildkite/dagster-buildkite/dagster_buildkite/steps/packages.py#L85)

this pr currently only tests for one example. will do it for all quickstart examples once i have a path forward for the following situation:

related to https://github.com/dagster-io/dagster/pull/13615, because these examples require `dagster-cloud` which cannot be installed as `dev+0` so it causes version conflict in tox. the change in PR can pass tox build but isn't ideal. a couple ideas:
1. move these tests to internal, where editable dagster-cloud is available
2. move dagster-cloud to oss
3. find a way to skip dagster-cloud install in tox (help needed. i can't figure out tox magic to do this)

### DIFF
--- a/examples/quickstart_aws/tox.ini
+++ b/examples/quickstart_aws/tox.ini
@@ -1,0 +1,22 @@
+[tox]
+envlist = py{38, 37}
+skipsdist = true
+
+[testenv]
+download = true
+passenv = CI_* COVERALLS_REPO_TOKEN BUILDKITE*
+deps =
+  -e ../../python_modules/dagster[test]
+  -e ../../python_modules/dagit
+  -e ../../python_modules/libraries/dagster-aws/
+  pandas
+  matplotlib
+  textblob
+  tweepy
+  wordcloud
+  ; -e .
+allowlist_externals =
+  /bin/bash
+commands =
+  !windows: /bin/bash -c '! pip list --exclude-editable | grep -e dagster -e dagit'
+  pytest -c ../../pyproject.toml -vv
--- a/examples/quickstart_aws/tox.ini
+++ b/examples/quickstart_aws/tox.ini
@@ -18,5 +18,5 @@ deps =
 allowlist_externals =
   /bin/bash
 commands =
-  !windows: /bin/bash -c '! pip list --exclude-editable | grep -e dagster -e dagit'
+  !windows: /bin/bash -c '! pip list --exclude-editable | grep -e dagster -e dagit -v dagster-graphql'
   pytest -c ../../pyproject.toml -vv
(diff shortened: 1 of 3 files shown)


yuhan:
* master
  * **PR #14740** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14740" target="_blank">Graphite</a> 
    * **PR #14741** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14741" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14741?utm_source=stack-comment).

examples/quickstart_aws/tox.ini
```diff
@@ -0,0 +1,22 @@
...
+deps =
+  -e ../../python_modules/dagster[test]
+  -e ../../python_modules/dagit
+  -e ../../python_modules/libraries/dagster-aws/
+  pandas
+  matplotlib
+  textblob
+  tweepy
+  wordcloud
+  ; -e .
```
yuhan, line 17: this is pretty bad i think because it can't really test the setup.py of this package

alangenfeld:
some thoughts:
* for testing `master` we can use do an install from https://github.com/dagster-io/dagster-cloud to get the dagster-cloud package
* we could setup tests that use published packages for dagster / dagster-cloud that verify the examples work. Could reference the back compat tests for how to set something up like that.

yuhan:
i think we'd wanna do both right?
> for testing master we can use do an install from https://github.com/dagster-io/dagster-cloud to get the dagster-cloud package

when in the middle of a release, we want to make sure examples work on the latest dagster-cloud and oss packages. guess then the question is is moving dagster-cloud to oss mono repo happening any time soon? if not, we could do git-based pip install and also point master to "dev+0" version to be consistent with other packages. does that sound right?

> we could setup tests that use published packages for dagster / dagster-cloud that verify the examples work. Could reference the back compat tests for how to set something up like that.

this i view as an end-to-end cloud nux test. maybe we can do this as an internal test?

alangenfeld:
> do git-based pip install 
this is what i was thinking, in conjunction with editable installs of other packages, but https://github.com/dagster-io/dagster-cloud/blob/main/dagster-cloud/dagster_cloud/version.py :/ makes that a no go unless we change things 

> this i view as an end-to-end cloud nux test. maybe we can do this as an internal test?

I think we can do the "do these examples work with the currently published package" test in oss, even for the ones that install dagster-cloud. 

For the "will the examples work once we publish whats in master" test, that i think could be done in `internal`

alangenfeld:
trying an approach here https://github.com/dagster-io/dagster/pull/14770

PR #14754, opened 2023-06-12
dagster project from-example use current version

alangenfeld:
to reduce risk of landed changes breaking examples for users, have the download script grab the version sycned with whats installed in `dagster`

## How I Tested These Changes

added test, manually run from-example

### DIFF
--- a/python_modules/dagster/dagster/_generate/download.py
+++ b/python_modules/dagster/dagster/_generate/download.py
@@ -6,10 +6,10 @@ from io import BytesIO
 import click
 import requests
 
+from dagster.version import __version__ as dagster_version
+
 from .generate import _should_skip_file
 
-# Currently we only download from 'master' branch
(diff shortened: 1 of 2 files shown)


alangenfeld:
I could potentially set-up a more complex test that actually downloads and loads up the example project using installed versions

python_modules/dagster/dagster/_generate/download.py
```diff
@@ -6,10 +6,10 @@
 import click
 import requests
 
+from dagster.version import __version__ as dagster_version
```
yuhan, line 9: can we instead get version from the current python env?

alangenfeld: can you elaborate? by my current interpretation that is already true - we are importing `dagster` in the current python environment and getting what version it is

yuhan: nvm - i misunderstood `dagster.version`.

yuhan (approved):

PR #14762, opened 2023-06-12
feat(dbt): propagate partial parsing to dynamic target

rexledesma:
## Summary & Motivation
Because we are dynamically setting the target path, dbt does not take advantage of partial parsing because it is unable to find a previously parsed artifact in the new target path.

This is unfortunate, because:

1. This artifact already exists, since we require a reference to the manifest when invoking the CLI.
2. In [dbt's experimental programmatic API](https://docs.getdbt.com/reference/programmatic-invocations#reusing-objects), there is an explicit example to reuse pre-constructed objects.

Here, we emulate that behavior. We assume the following:

- Given a dbt project directory (`DBT_PROJECT_DIR`), defined by the user when instantiating `DbtCli`, we inspect if `{DBT_PROJECT_DIR}/target/partial_parse.msgpack` exists.
- If this file exists, then copy it over to the dynamic target path before invoking `DbtCli.cli(...)`. This way, partial parsing is enabled. 

## How I Tested These Changes
pytest

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
@@ -1,9 +1,11 @@
 import json
 import os
+import shutil
 import subprocess
 import sys
 import time
 from dataclasses import dataclass
+from pathlib import Path
 from typing import Any, Dict, Iterator, List, Mapping, Optional, Sequence, Union
 
 from dagster import (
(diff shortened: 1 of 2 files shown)


rexledesma:
* master
  * **PR #14762** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14762" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14762?utm_source=stack-comment).

geoHeil:
Looks really interesting!

OwenKephart (approved):

PR #14767, opened 2023-06-13
feat: allow metadata overrides by asset key

rexledesma:
## Summary & Motivation
Overriding `metadata_by_key` was missing in `.with_attributes`. Add it.

## How I Tested These Changes
pytest

### DIFF
--- a/python_modules/dagster/dagster_tests/asset_defs_tests/test_assets.py
+++ b/python_modules/dagster/dagster_tests/asset_defs_tests/test_assets.py
@@ -156,6 +156,40 @@ def test_with_replaced_description() -> None:
     }
 
 
+def test_with_replaced_metadata() -> None:
+    @multi_asset(
+        outs={
+            "foo": AssetOut(metadata={"foo": "foo"}),
+            "bar": AssetOut(metadata={"bar": "bar"}),
+            "baz": AssetOut(metadata={"baz": "baz"}),
+        }
+    )
+    def abc():
+        ...
+
+    assert abc.metadata_by_key == {
+        AssetKey("foo"): {"foo": "foo"},
+        AssetKey("bar"): {"bar": "bar"},
+        AssetKey("baz"): {"baz": "baz"},
+    }
+
+    # If there's no replacement metadata for the asset key, the original metadata is retained.
+    replaced = abc.with_attributes(metadata_by_key={})
+    assert replaced.metadata_by_key == {
+        AssetKey("foo"): {"foo": "foo"},
+        AssetKey("bar"): {"bar": "bar"},
+        AssetKey("baz"): {"baz": "baz"},
+    }
+
+    # Otherwise, use the replaced description.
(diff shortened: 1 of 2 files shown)


rexledesma:
* master
  * **PR #14767** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14767" target="_blank">Graphite</a>  👈
    * **PR #14718** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14718" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14767?utm_source=stack-comment).

rexledesma:
@slopp [noticed a peculiar interaction here](https://github.com/dagster-io/dagster/pull/14718#issuecomment-1589493721): we're not actually replacing the metadata, but merging it. 

https://github.com/dagster-io/dagster/blob/a8b1023262545d745d8a57850d09d9ced8ff6a9b/python_modules/dagster/dagster/_core/definitions/assets.py#L197-L200

This implicit merging behavior doesn't feel in line with `.with_attributes`. Changing this default behavior is out of the scope of this change. I just wanted to surface this for the team (cc @jamiedemaria @schrockn)

python_modules/dagster/dagster_tests/asset_defs_tests/test_assets.py
```diff
@@ -156,6 +156,40 @@ def abc():
...
+    def abc():
+        ...
+
+    assert abc.metadata_by_key == {
+        AssetKey("foo"): {"foo": "foo"},
+        AssetKey("bar"): {"bar": "bar"},
+        AssetKey("baz"): {"baz": "baz"},
+    }
+
+    # If there's no replacement metadata for the asset key, the original metadata is retained.
```
OwenKephart, line 176: This logic differs from (e.g.) descriptions_by_key, which on closer inspection seems to wipe out existing descriptions when this method is invoked without a value there. I also think we should distinguish between an empty dictionary and `None`, as passing in an empty dictionary seems like "there is no metadata" rather than "just keep whatever metadata you have". High level I think we should do something like:

```
thing_by_key = check.dict_param(...) if thing_by_key is not None else self.thing_by_key

replaced_thing_by_key = {output_key_replacements.get(key, key): value for key, value in thing_by_key.items()}
```

OwenKephart (changes requested):
quick note on adjusting the logic for both this and descriptions_by_key

python_modules/dagster/dagster_tests/asset_defs_tests/test_assets.py
rexledesma: This is actually the same behavior as `descriptions_by_key`.

https://github.com/dagster-io/dagster/blob/d876bde018919fbbb252f8078a6a57627d22b2d8/python_modules/dagster/dagster_tests/asset_defs_tests/test_assets.py#L142-L156

OwenKephart (approved):

PR #14768, opened 2023-06-13
[tox] remove envlist

alangenfeld:
Why I think we should remove these
* the vast majority are out of sync with the versions actively under test
* all this does is provide a set of environments to test by default when invoking `tox` which i dont believe is a workflow anyone does  ( https://tox.wiki/en/latest/config.html#envlist )

## How I Tested These Changes

bk

### DIFF
--- a/python_modules/libraries/dagster-k8s/tox.ini
+++ b/python_modules/libraries/dagster-k8s/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{39,38,37,36}-{unix,windows}-{default,old_kubernetes}
+envlist = {default,old_kubernetes}
 skipsdist = true
 
 [testenv]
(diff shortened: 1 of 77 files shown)


alangenfeld:
* master
  * **PR #14768** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14768" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14768?utm_source=stack-comment).

python_modules/libraries/dagstermill/tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{39,38,37,36}-{papermill1,papermill2}-{unix,windows}
+envlist = {papermill1,papermill2}
```
alangenfeld, line 2: [1]

python_modules/libraries/dagster-k8s/tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{39,38,37,36}-{unix,windows}-{default,old_kubernetes}
+envlist = {default,old_kubernetes}
```
alangenfeld, line 2: [1] here we have to list the factors out since `default` here and `papermill2` below are not referenced directly in the tox file

gibsondan (approved):
good riddance

PR #14773, opened 2023-06-13
[fix][pythonic resources] Fix requesting Pythonic resources in run failure sensor function

### LINKED ISSUE #14556: Run failure sensors don't inject pythonic resources
### Dagster version

1.3.6

### What's the issue?

When creating a sensor, I can declare constructor parameters to have pythonic resources injected when that sensor runs.

However when creating a run failure sensor, I get an error trying to do the same.

For example, I have a pythonic resource class called ```DbConnection```.

If I declare this run failure sensor:

```python
@run_failure_sensor(
    name="project_processing_failed_sensor"
)
def project_processing_failed_sensor(
    context: RunFailureSensorContext,
    db_connection: DbConnectionResource):
```

I can enable the sensor, however when a job fails and initiates it - I get this error:

_TypeError: project_processing_failed_sensor() missing 1 required positional argument: 'db_connection_

I would expect a run failure sensor to support pythonic resources in the constructor.

### How to reproduce?

1. Create any job that will throw an error and fail when started.
1. Create a pythonic resource class.
1. Create a run failure sensor that takes that pythonic resource class as a parameter.
1. Start dagster (I was testing in dev though my prod environment is on k8).
1. Start the new run failure sensor in the "sensors" area of the UI so it is enabled.
1. Start the job that will fail and wait for it to fail.
1. You should see an error in the logs: 

_TypeError: project_processing_failed_sensor() missing 1 required positional argument: [name of your pythonic resource variable]_

### Deployment type

Local

### Deployment details

_No response_

### Additional information

_No response_

benpankow:
## Summary

Fixes requesting Pythonic resources with a run failure sensor. This was broken because the wrapper function which turns the run failure sensor into a generic run status sensor didn't forward the resource requirements.

Resolves #14556

## Test Plan

New unit tests for direct invocation + testing daemon sensor execution.


benpankow:
* master
  * **PR #14773** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14773" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14773?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/run_status_sensor_definition.py
```diff
@@ -461,8 +468,17 @@ def inner(
             request_job=request_job,
             request_jobs=request_jobs,
         )
-        def _run_failure_sensor(context: RunStatusSensorContext):
-            return fn(context.for_run_failure())  # fmt: skip
+        @mirror_signature_of(fn)
```
alangenfeld, line 471: does `wraps` not cover this? 

https://docs.python.org/3/library/functools.html#functools.wraps

benpankow: so it does

OwenKephart (approved):
nice

PR #14776, opened 2023-06-13
feat(dbt-assets): standardize auto materialize and freshness policies under `+meta`

rexledesma:
## Summary & Motivation
Previously, freshness policies and auto materialize policies were configured using the `dagster_freshness_policy` and `dagster_auto_materialize_policy` fields.

These should not be top level fields. Instead, this should be [nested under `+meta`](https://docs.getdbt.com/reference/resource-configs/meta). As an integration, we can own `dagster` as a field in `+meta`. In this `dagster` field, we can stick existing configuration as well as future configuration that should be associated with a dbt model.

This makes features like [config inheritance](https://docs.getdbt.com/reference/configs-and-properties#config-inheritance) more intuitive for the user to use when configuring their project for Dagster consumption. 

## How I Tested These Changes
pytest


rexledesma:
* master
  * **PR #14776** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14776" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14776?utm_source=stack-comment).

OwenKephart (approved):

PR #14780, opened 2023-06-13
bump dagster-graphql timeout

gibsondan:
Summary:
This is a sad thing, but bumping the timeout is better than repeatedly timing out as a short-term workaround

Test Plan: BK


.buildkite/dagster-buildkite/dagster_buildkite/steps/packages.py
```diff
@@ -408,6 +408,7 @@ def k8s_extra_cmds(version: str, _) -> List[str]:
                 else []
             )
         ),
+        timeout_in_minutes=30,
```
alangenfeld, line 411: :(

alangenfeld (approved):

PR #14782, opened 2023-06-13
Make graphql context test suite fixture class-scoped

gibsondan:
Summary:
Sharing the context between the same tests in a class (wiping the storage between classes) makes these test suites take ~15 minutes instead of ~25 minutes.

To make the wiping between tests not destructive I had to make the sqlite event log storage wipe semantically similar to the other wipe() methods - deleting the rows in the table but not dropping the schema entirely.

### DIFF
--- a/python_modules/dagster/dagster/_core/storage/event_log/sqlite/sqlite_event_log.py
+++ b/python_modules/dagster/dagster/_core/storage/event_log/sqlite/sqlite_event_log.py
@@ -368,15 +368,21 @@ class SqliteEventLogStorage(SqlEventLogStorage, ConfigurableClass):
             self.delete_events_for_run(conn, run_id)
 
     def wipe(self) -> None:
-        # should delete all the run-sharded dbs as well as the index db
+        # should delete all the run-sharded db files and drop the contents of the index
         for filename in (
             glob.glob(os.path.join(self._base_dir, "*.db"))
             + glob.glob(os.path.join(self._base_dir, "*.db-wal"))
             + glob.glob(os.path.join(self._base_dir, "*.db-shm"))
         ):
-            os.unlink(filename)
+            if (
+                not filename.endswith(f"{INDEX_SHARD_NAME}.db")
+                and not filename.endswith(f"{INDEX_SHARD_NAME}.db-wal")
+                and not filename.endswith(f"{INDEX_SHARD_NAME}.db-shm")
+            ):
+                os.unlink(filename)
 
         self._initialized_dbs = set()
+        self._wipe_index()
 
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster/_core/storage/event_log/sqlite/sqlite_event_log.py
```diff
@@ -368,15 +368,17 @@ def delete_events(self, run_id: str) -> None:
...
     def wipe(self) -> None:
-        # should delete all the run-sharded dbs as well as the index db
+        # should delete all the run-sharded db files and drop the contents of the index
         for filename in (
             glob.glob(os.path.join(self._base_dir, "*.db"))
             + glob.glob(os.path.join(self._base_dir, "*.db-wal"))
             + glob.glob(os.path.join(self._base_dir, "*.db-shm"))
         ):
-            os.unlink(filename)
+            if f"{INDEX_SHARD_NAME}.db" not in filename:
```
gibsondan, line 377: probably a better way to do this part...

alangenfeld: maybe slightly cleaner with an `endswith` ? i think this is fine to commit but leave a comment here in the code

alangenfeld (approved):
nice, how did you find this, profiler ?

gibsondan:
Just eyeballed a lot of starting and stopping gRPC servers - run processes starting up is still a low hanging fruit to speed this up further (using the SyncMemoryRunLauncher or calling execute_job is probably the right move in many of these tests that aren't specifically testing run launching)

python_modules/dagster/dagster/_core/storage/event_log/sqlite/sqlite_event_log.py
```diff
@@ -368,15 +368,21 @@ def delete_events(self, run_id: str) -> None:
...
         for filename in (
             glob.glob(os.path.join(self._base_dir, "*.db"))
             + glob.glob(os.path.join(self._base_dir, "*.db-wal"))
             + glob.glob(os.path.join(self._base_dir, "*.db-shm"))
         ):
-            os.unlink(filename)
+            if (
+                not filename.endswith(f"{INDEX_SHARD_NAME}.db")
+                and not filename.endswith(f"{INDEX_SHARD_NAME}.db-wal")
+                and not filename.endswith(f"{INDEX_SHARD_NAME}.db-shm")
```
alangenfeld, line 380: ah - i didnt catch the `in`check  was for handling the `wal` and `shm` files as well

gibsondan: all the more reason to be explicit then!

PR #14784, opened 2023-06-13
Merge job resource_definitions with late-bound resources instead of replacing

### LINKED ISSUE #14727: Adding io_manager to Definitions conflicts with job resource_defs
### Dagster version

1.3.9

### What's the issue?

When there is an `io_manager` in the list of resources provided to the `Definitions` objects, Dagster ignores any resources provided to a job with `resource_defs`. 

Example error:
```
2023-06-09 09:46:11 -0700 - dagster.daemon - INFO - Instance is configured with the following daemons: ['AssetDaemon', 'BackfillDaemon', 'SchedulerDaemon', 'SensorDaemon']
2023-06-09 09:46:11 -0700 - dagster.daemon.SensorDaemon - WARNING - Could not load location repository.py to check for sensors due to the following error: dagster._core.errors.DagsterInvalidDefinitionError: resource with key 'database' required by op 'op_requires_resources' was not provided. Please provide a <class 'dagster._core.definitions.resource_definition.ResourceDefinition'> to key 'database', or change the required key to one of the following keys which points to an <class 'dagster._core.definitions.resource_definition.ResourceDefinition'>: ['io_manager']
```

I know this is not a common case, but developers might experience this behavior when they start migrating their code to use the `Definitions` object.

I would expect the job to ignore the `io_manager` define in `Definitions` if I provide `resource_defs`, just like it ignores other resources.

### How to reproduce?

Here's a simple example based on the resources example. Running `dagster dev` fails.

```python
from dagster import op, job, resource, Definitions
from dagster import FilesystemIOManager

CREATE_TABLE_1_QUERY = "create table_1 as select * from table_0"


class DummyDB:
    def execute_query(self, query):
        print("executing query: ", query)


@resource
def dummy_database_resource(init_context):
    return DummyDB()


@op(required_resource_keys={"database"})
def op_requires_resources(context):
    context.resources.database.execute_query(CREATE_TABLE_1_QUERY)


@job(resource_defs={"database": dummy_database_resource})
def do_database_stuff():
    op_requires_resources()


@op
def simple_op():
    print("simple op")


@job()
def simple_job():
    simple_op()


resources = {"io_manager": FilesystemIOManager()}

defs = Definitions(
    jobs=[do_database_stuff, simple_job],
    resources=resources,
)
```

If you remove `simple_job` and provide an empty dict of resources to the `Definitions` object, the code works:
```
defs = Definitions(
    jobs=[do_database_stuff, simple_job],
    resources=resources,
)
```

### Deployment type

None

### Deployment details

_No response_

### Additional information

The workaround is to add an `io_manager` to the job's `resource_defs`.

benpankow:
## Summary

Resolves #14727.

Explicitly merges top-level resource definitions with job-level resource definitions. This caused issues (such as in the linked report above) when a job specified `resource_defs` but the user supplied e.g. a top-level IO manager which got bound to the job.

## Test Plan

New, failing test based on user report, existing unit tests.

### DIFF
--- a/python_modules/dagster/dagster_tests/core_tests/resource_tests/pythonic_resources/test_binding.py
+++ b/python_modules/dagster/dagster_tests/core_tests/resource_tests/pythonic_resources/test_binding.py
@@ -633,8 +633,8 @@ def test_late_binding_with_resource_defs() -> None:
 
     @op
     def simple_op():
-        print("simple op")
-
+        pass
+    
     @job()
     def simple_job():
         simple_op()
--- a/python_modules/dagster/dagster_tests/core_tests/resource_tests/pythonic_resources/test_binding.py
+++ b/python_modules/dagster/dagster_tests/core_tests/resource_tests/pythonic_resources/test_binding.py
@@ -615,9 +615,11 @@ def test_bind_with_string_annotation():
 
 
 def test_late_binding_with_resource_defs() -> None:
+    queries = []
+
     class DummyDB:
         def execute_query(self, query):
-            pass
+            queries.append(query)
 
     @resource
     def dummy_database_resource(init_context):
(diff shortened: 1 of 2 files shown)


benpankow:
* master
  * **PR #14784** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14784" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14784?utm_source=stack-comment).

python_modules/dagster/dagster_tests/core_tests/resource_tests/pythonic_resources/test_binding.py
```diff
@@ -585,3 +587,36 @@ def my_asset(context, my_resource: MyResourceWithDefault):
...
+    @job()
+    def simple_job():
+        simple_op()
+
+    # io_manager here will be bound to both jobs
+    # we need to make sure this doesn't invalidate the database resource
+    Definitions(
+        jobs=[do_database_stuff, simple_job],
+        resources={"io_manager": FilesystemIOManager()},
+    )
```
dpeng817, line 622: could we also test the behavior when you provide the key at both the job level and the definitions level?

dpeng817 (approved):
jolly good

PR #14790, opened 2023-06-14
docs(dbt): add example to yield materializations in a job

rexledesma:
## Summary & Motivation
Show an example of using `DbtCli` if you are still using ops and jobs.

## How I Tested These Changes
local

### DIFF
--- a/examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/jobs/yield_materializations.py
+++ b/examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/jobs/yield_materializations.py
@@ -1,0 +1,29 @@
+from dagster import AssetMaterialization, Output, job, op
+from dagster_dbt.cli import DbtCli, DbtManifest
+
+from ..constants import MANIFEST_PATH
+
+manifest = DbtManifest.read(path=MANIFEST_PATH)
+asset_key_by_output_name = {
+    output_name: manifest.node_info_to_asset_key(node_info)
+    for output_name, node_info in manifest.node_info_by_output_name.items()
+}
+
+
+@op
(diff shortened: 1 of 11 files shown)


rexledesma:
* master
  * **PR #14790** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14790" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14790?utm_source=stack-comment).

examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/jobs/yield_materializations.py
```diff
@@ -0,0 +1,29 @@
...
+manifest = DbtManifest.read(path=MANIFEST_PATH)
+asset_key_by_output_name = {
+    output_name: manifest.node_info_to_asset_key(node_info)
+    for output_name, node_info in manifest.node_info_by_output_name.items()
+}
+
+
+@op
+def my_dbt_build_op(dbt: DbtCli):
+    for dagster_event in dbt.cli(["build"], manifest=manifest).stream():
```
OwenKephart, line 15: At the very least, we should have a comment here explaining why we're converting Output events into AssetMaterialization events, as it's a bit weird. Some alternatives:

- just use stream_raw_events() and manually construct the Output events
- use stream_raw_events() in conjunction with event.to_default_asset_events(for_op=True) or something of the sort (alternatively, `event.to_op_events()`)
- have some boolean flag on `.stream()` that indicates that we want AssetMaterializations instead of Outputs

I don't think this is super high priority in the short term (we can always layer on APIs after the fact), so for now just a comment is fine, just food for thought

OwenKephart (approved):

PR #14794, opened 2023-06-14
feat(dbt): parse asset key from `+meta`

rexledesma:
## Summary & Motivation
Allow users to configure their Dagster asset keys as part of their model configuration.

This gives them a durable reference for an asset key when their project is ingested by Dagster, rather than relying on dbt project/model hierarchies.

## How I Tested These Changes
pytest


sryza (commented):
This looks good to me. Going to leave it to @OwenKephart for final approve because of my lack of dbt knowledge

OwenKephart (approved):
one question -- for things like freshness policy / auto materialize policies, it makes sense to configure this for multiple assets at a time, whereas for asset keys that's of course not the case.

If I configure an asset key inside the "dagster" meta field for a specific model, but then configure an auto-materialize policy for all models somewhere else, do those meta dictionaries merge or will the global setting override the local one?

If those dictionaries merge, then all good 👍

rexledesma:
> one question -- for things like freshness policy / auto materialize policies, it makes sense to configure this for multiple assets at a time, whereas for asset keys that's of course not the case.
> 
> If I configure an asset key inside the "dagster" meta field for a specific model, but then configure an auto-materialize policy for all models somewhere else, do those meta dictionaries merge or will the global setting override the local one?
> 
> If those dictionaries merge, then all good 👍

They merge! See https://docs.getdbt.com/reference/configs-and-properties#combining-configs.

PR #14795, opened 2023-06-14
Odette/machine_learning_basic

odette-elementl:


odette-elementl:
Leaving the mlflow stuff for another PR

sryza (commented):
What's the relationship between this code and the code in the guide? Is it identical?

odette-elementl:
> 

yeah it is

sryza:
I'm noticing some differences between it and ml_pipeline.py. E.g.
- it has `make_plot`
- it doesn't have `comments_model_test_set_r_squared`

Including / omitting those on their own seems reasonable either way, but I think this will be easier to maintain if it actually has the exact same assets and implementations, because then changes can be applied to both without thinking, and users can expect that what they saw in the guide is exactly what they'll get in the example.

odette-elementl:
> I'm noticing some differences between it and ml_pipeline.py. E.g.
> 
> * it has `make_plot`
> * it doesn't have `comments_model_test_set_r_squared`
> 
> Including / omitting those on their own seems reasonable either way, but I think this will be easier to maintain if it actually has the exact same assets and implementations, because then changes can be applied to both without thinking, and users can expect that what they saw in the guide is exactly what they'll get in the example.

Aligned directly with the guide now

sryza (approved):
lgtm!

cmpadden:

PR #14798, opened 2023-06-14
Initial commit on guide on computing with Dask on Dagster

odette-elementl:


docs/content/guides/dagster/parallel-computing-dask.mdx
```diff
@@ -0,0 +1,270 @@
...
+
+- Set up a Dask resources
+- Use Dask to speed up data assets
+- Use Dask’s machine learning tools to build a model
+
+---
+
+## Using Dagster’s resources to set up a Dask Cluster
+
+When using Dask with Dagster, you can use the `DaskResource` to set up a Dask client and simplify a pipeline that uses Dask across multiple assets. In this example, a local`dask.distributed` client is set up as a `ConfigurableResource`. Resources allow you to centralize your configuration in a single location, reducing redundant code in your assets.
```
sryza, line 20: Add space between "local" and "dask.distributed" I think.

docs/content/guides/dagster/parallel-computing-dask.mdx
```diff
@@ -0,0 +1,270 @@
...
+
+```python file=/guides/dagster/parallel-computing-dask/resources.py startafter=first_resource_start endbefore=first_resource_end
+from dask.distributed import Client
+
+# isort: skip_file
```
sryza, line 25: I don't think we want this isort directive to show up in the example code. Do we need to skip import sorting?

docs/content/guides/dagster/parallel-computing-dask.mdx
sryza: > you can use the `DaskResource`

This makes it sound like Dagster provides its own `DaskResource`. I think something like "You can define a `DaskResource` to share Dask configuration across multiple assets" would be more accurate.

docs/content/guides/dagster/parallel-computing-dask.mdx
```diff
@@ -0,0 +1,270 @@
...
+from dagster import ConfigurableResource
+
+
+class DaskResource(ConfigurableResource):
+    def make_dask_cluster(self) -> Client:
+        client = Client()
+        return client
+```
+
+The `dask.distributed.LocalCluster` API allows customization of the cluster, which can be leveraged through Dagster’s `resource`. In the following example, configuration is setup at the resource level, which is then leveraged by Definitions and applied to all uses of the resource.
```
sryza, line 35: I don't think "resource" should be monospaced,  because it doesn't correspond to a particular code symbol.

docs/content/guides/dagster/parallel-computing-dask.mdx
```diff
@@ -0,0 +1,270 @@
...
+## Using the Dask Resource
+
+Now that the Dask resource is set up, an asset can use it to parallelize computations. This asset demonstrates using the `client` within Dagster using the example from the [Dask distributed quickstart](https://distributed.dask.org/en/stable/quickstart.html).
+
+```python file=/guides/dagster/parallel-computing-dask/use_dask.py startafter=first_ex_start endbefore=first_ex_end
+from dagster import asset, MetadataValue, Definitions
+
+
+@asset
+def simple_dask_asset(context, MyDaskResource: DaskResource):
```
sryza, line 80: variable names should be lower case, maybe just `dask_resource` or `dask`

sryza: it looks like `context` isn't used, so best to just omit it

sryza: including a return type annotation would help make it clear what this function is doing

also, does our default IO manager work with dask dataframes?

sryza (commented):
I left a few comments - otherwise I think this looks good!

Can we make sure the code is tested?

yuhan (changes requested):
queue management - we rolled out https://dagster.io/blog/dagster-dask, so maybe close this one?

PR #14802, opened 2023-06-14
[1.4] feat(dbt)!: parse asset group from `+group`

rexledesma:
## Summary & Motivation
Rather than infering the group using the subdirectory structure, leverage dbt's 1.5 concept for a group: https://docs.getdbt.com/docs/build/groups.

If a dbt group (`+group`) and Dagster group (`+meta`, using `dagster.group`) are configured for a dbt model, then defer to using the Dagster group.

## How I Tested These Changes
pytest


rexledesma:
* master
  * **PR #14794** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14794" target="_blank">Graphite</a> 
    * **PR #14802** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14802" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14802?utm_source=stack-comment).

rexledesma:
This is potentially breaking for users. Might have to wait to release this as part of `1.4.0`?

sryza:
> This is potentially breaking for users. Might have to wait to release this as part of 1.4.0?

This only affects users who opt in to the dbt 1.5 feature, right? And it mainly affects how assets are organized in the UI, it won't cause Python errors.

If that's the case, I think it could be safe to merge pre-1.4.

rexledesma:
> > This is potentially breaking for users. Might have to wait to release this as part of 1.4.0?
> 
> This only affects users who opt in to the dbt 1.5 feature, right? And it mainly affects how assets are organized in the UI, it won't cause Python errors.
> 
> If that's the case, I think it could be safe to merge pre-1.4.

Yes, this is a dbt 1.5 feature. I'm thinking about the case if a user uses `AssetSelection.groups(...)`. Arguably this could be treated as a bug fix though.

sryza:
> Yes, this is a dbt 1.5 feature. I'm thinking about the case if a user uses AssetSelection.groups(...). Arguably this could be treated as a bug fix, since this is probably the intended behavior?

Hmm good point. Ok, maybe better to wait.

sryza (approved):

PR #14803, opened 2023-06-14
Remove old partition mapping methods

clairelin135:
This PR refactors built in partition mappings to implement:
- `get_upstream_mapped_partitions_result_for_partitions`
- `get_downstream_partitions_for_partitions`

Removes the following unused APIs:
- `get_upstream_partitions_for_partition_range`
- `get_upstream_partitions_for_partitions`
- `get_downstream_partitions_for_partition_range`

Updates tests:
- Uses the new methods
- Removes some tests for custom partition mappings, since this is no longer supported

This will be a breaking change for anyone that has implemented their own partition mapping. Defining a custom partition mapping has been unsupported since 1.1.7.

Happy to break this down into 2 different PRs if preferred

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/partition_mapping.py
+++ b/python_modules/dagster/dagster/_core/definitions/partition_mapping.py
@@ -223,10 +223,33 @@ class IdentityPartitionMapping(PartitionMapping, NamedTuple("_IdentityPartitionM
         downstream_partitions_def: Optional[PartitionsDefinition],
         upstream_partitions_def: PartitionsDefinition,
     ) -> PartitionKeyRange:
-        if downstream_partitions_def is None or downstream_partition_key_range is None:
+        raise NotImplementedError()
+
+    def get_upstream_mapped_partitions_result_for_partitions(
+        self,
+        downstream_partitions_subset: Optional[PartitionsSubset],
+        upstream_partitions_def: PartitionsDefinition,
+        current_time: Optional[datetime] = None,
+        dynamic_partitions_store: Optional[DynamicPartitionsStore] = None,
+    ) -> UpstreamPartitionsResult:
+        if downstream_partitions_subset is None:
             check.failed("downstream asset is not partitioned")
 
-        return downstream_partition_key_range
+        if downstream_partitions_subset.partitions_def == upstream_partitions_def:
+            return UpstreamPartitionsResult(
+                downstream_partitions_subset,
+                [],
+            )
+
+        upstream_partition_keys = set(upstream_partitions_def.get_partition_keys())
+        downstream_partition_keys = set(downstream_partitions_subset.get_partition_keys())
+
+        return UpstreamPartitionsResult(
+            upstream_partitions_def.empty_subset().with_partition_keys(
+                list(upstream_partition_keys & downstream_partition_keys)
+            ),
+            list(downstream_partition_keys - upstream_partition_keys),
+        )
 
     def get_downstream_partitions_for_partition_range(
         self,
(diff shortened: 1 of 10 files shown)


python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -217,24 +107,51 @@ class IdentityPartitionMapping(PartitionMapping, NamedTuple("_IdentityPartitionM
...
-    ) -> PartitionKeyRange:
-        if downstream_partitions_def is None or downstream_partition_key_range is None:
+        current_time: Optional[datetime] = None,
+        dynamic_partitions_store: Optional[DynamicPartitionsStore] = None,
+    ) -> UpstreamPartitionsResult:
+        if downstream_partitions_subset is None:
             check.failed("downstream asset is not partitioned")
 
-        return downstream_partition_key_range
+        if downstream_partitions_subset.partitions_def == upstream_partitions_def:
```
clairelin135, line 120: Previously the `IdentityPartitionMapping` would raise an error when partition keys in `downstream_partitions_subset` did not exist in `upstream_partitions_def`. This change switches this behavior to return the list of nonexistent but required parent keys instead.

A good future change would be to add more definition-time checking, i.e. to ensure that when the IdentityPartitionMapping is applied or inferred, the partitions defs are the same.

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -217,24 +107,51 @@ class IdentityPartitionMapping(PartitionMapping, NamedTuple("_IdentityPartitionM
...
+            return UpstreamPartitionsResult(
```
sryza, line 121: Nitpick: this could go on one line

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -217,24 +107,51 @@ class IdentityPartitionMapping(PartitionMapping, NamedTuple("_IdentityPartitionM
...
+                downstream_partitions_subset,
+                [],
+            )
+
+        upstream_partition_keys = set(upstream_partitions_def.get_partition_keys())
+        downstream_partition_keys = set(downstream_partitions_subset.get_partition_keys())
+
+        return UpstreamPartitionsResult(
+            upstream_partitions_def.empty_subset().with_partition_keys(
```
sryza, line 130: Nitpick: use `subset_with_partition_key`

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -217,24 +107,51 @@ class IdentityPartitionMapping(PartitionMapping, NamedTuple("_IdentityPartitionM
...
             check.failed("downstream asset is not partitioned")
 
-        return downstream_partition_key_range
+        if downstream_partitions_subset.partitions_def == upstream_partitions_def:
+            return UpstreamPartitionsResult(
+                downstream_partitions_subset,
+                [],
+            )
+
+        upstream_partition_keys = set(upstream_partitions_def.get_partition_keys())
```
sryza, line 126: Do we need to be passing in the dynamic partitions store?

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
```diff
@@ -217,24 +107,51 @@ class IdentityPartitionMapping(PartitionMapping, NamedTuple("_IdentityPartitionM
...
+        downstream_partition_keys = set(downstream_partitions_subset.get_partition_keys())
+
+        return UpstreamPartitionsResult(
+            upstream_partitions_def.empty_subset().with_partition_keys(
+                list(upstream_partition_keys & downstream_partition_keys)
```
sryza, line 131: Not critical right now, but at some point we might want to add an `__and__` method to `PartitionsSubset` so we can do these set operations without needing to convert back and forth between different collection types.

sryza (commented):
Love to see this:

image


For the tests we're getting rid of, is the functionality that they're testing covered by other tests?

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
clairelin135: No, dynamic partitions use the regular default partitions subset which serializes the partition keys in the subset. We only need to pass in the dynamic partitions store for operations that require fetching partition keys outside of the set (i.e. `get_partition_keys_not_in_subset`)

sryza: What about for `PartitionsDefinition.get_partition_keys` though, which we're calling for the upstream?

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
clairelin135: Agree--we do lots of "and" operations in the asset backfill and partition status cache logic, so this would be helpful. Would prefer to add this in a separate PR though

sryza: That's fine

clairelin135:
@sryza most of the removed tests were testing custom partition mappings. Since we only support built-in partition mappings now, I think we have sufficient test coverage for these these partition mappings

sryza:
> @sryza most of the removed tests were testing custom partition mappings. Since we only support built-in partition mappings now, I think we have sufficient test coverage for these these partition mappings

Gotcha. And it seems like a lot of the functionality that we were testing is now covered by actual PartitionMappings that we expose.

python_modules/dagster/dagster/_core/definitions/partition_mapping.py
clairelin135: oooo good catch, yes that should be there

sryza (approved):

PR #14172, opened 2023-05-09
bumped pandas version to greater than 2

pzarabadip:
## Summary & Motivation
This PR aims to address #13339.
It's gonna be ready for review when following libraries are migrated/adapted to work with both pandas 1.0 and 2.0.

- [ ] assets-pandas-type-metadata
- [ ] assets-pandas-pyspark
- [ ] dagster-duckdb-pandas
- [ ] dagster-duckdb-pyspark
- [ ] dagster-gcp-pandas

## How I Tested These Changes
Tests run successfully on my local machine.
- [ ] assets-pandas-type-metadata
- [ ] assets-pandas-pyspark
- [ ] dagster-duckdb-pandas
- [ ] dagster-duckdb-pyspark
- [ ] dagster-gcp-pandas

pinging @smackesey

PR #14804, opened 2023-06-14
Use syrupy instead of snapshottest for snapshot tests

gibsondan:
Summary:
This other snapshot library has the same API but doesn't delete all the other snapshots when you run a test suite with -k, and raises an exception if the snapshots are missing. Plus, it allows you to customize the name of the snapshot so we don't need to generate a unique one for each and every graphql test suite.


examples/docs_snippets/docs_snippets/legacy/dagster_pandas_guide/core_trip.py
```diff
@@ -31,7 +31,7 @@ def load_trip_dataframe() -> DataFrame:
     return read_csv(
         file_relative_path(__file__, "./ebike_trips.csv"),
         parse_dates=["start_time", "end_time"],
-        date_parser=lambda x: datetime.strptime(x, "%Y-%m-%d %H:%M:%S.%f"),
+        date_format="%Y-%m-%d %H:%M:%S.%f",
```
gibsondan, line 34: unrelated result of bumping the pyright deps i think

examples/docs_snippets/docs_snippets/legacy/dagster_pandas_guide/core_trip.py
```diff
@@ -1,3 +1,5 @@
+# pyright uses pandas 2 but docs_snippets is stuck on pandas 1 due to pins in dagster-duckdb-pandas
```
gibsondan, line 1: this is a side effect of the pyright pins updating, not directly related other than that i needed to bump the pins to add the syrupy dep and remove the snapshottest dep

alangenfeld: thoughts on this @smackesey ?

smackesey: I see, I thought we'd successfully moved to pandas 2 everywhere. @jamiedemaria I think I remember talking about this transition with you?

In any case putting an ignore here is no biggie, but we should take care of the dagster-duckdb-pandas transition ASAP so we can get everything on pandas 2. Then this ignore will get auto-flagged as unnecessary and can be removed.

jamiedemaria: https://github.com/dagster-io/dagster/pull/14172 this is what i found

smackesey: OK, here is a PR that updates the whole repo to support pandas 2 (and updates pins): https://github.com/dagster-io/dagster/pull/14956

@gibsondan This will merge before the current PR so you should regen the pyright pins on rebase and shouldn't need to have any special handling for pandas.

alangenfeld:
* master
  * **PR #14804** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14804" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14804?utm_source=stack-comment).

PR #14751, opened 2023-06-12
call setup_for_execution for inner IOManagers of the BranchingIOManager

danielgafni:
## Summary & Motivation

The `BranchingIOManager` was drafted by @schrockn before the introduction of lifecycle hooks. 

This PR adds `setup_for_execution` and `teardown_after_execution` calls to the inner IOManagers.

## How I Tested These Changes
Not yet

PR #14807, opened 2023-06-14
[pythonic resources] Improve nested resource evaluation logic

benpankow:
## Summary

This PR attempts to address a couple shortcomings with the Pythonic resource system which are exposed in #14751. 

First, the logic to determine whether a field on a resource represents a nested resource or a config field is not a good heuristic currently. Right now, the check only assesses whether the value is a resource type - but this fails in the case that a user wants to provide a raw value (ie an instance of `IOManager`). This PR updates this check to inspect the annotation on the field, and treat it as a resource only if the annotation indicates as much.

Second, the initialization logic for nested resources (`coerce_to_resource`) doesn't properly handle raw values such as a plain `IOManager`. This PR updates the logic in `coerce_to_resource` to call out to `wrap_resources_for_execution`, which can successfully deal with these cases.

## Example

The following resource previously would not accept a raw `IOManager` as input, only a resource which produces one:
```python
class MyResourceNeedsIOManager(ConfigurableIOManager):
    base_io_manager: ResourceDependency[IOManager]


@io_manager
def my_io_manager(context) -> IOManager:
    pass

# previously OK, since my_io_manager is a resource
MyResourceNeedsIOManager(base_io_manager=my_io_manager)


class MyIOManager(IOManager):
    pass

# previously errors, since MyIOManager is not a resource
MyResourceNeedsIOManager(base_io_manager=MyIOManager())
```

The new code will coerce `MyIOManager` to a resource by wrapping it, since it is annotated as a `ResourceDependency`. This gives the user more flexibility when specifying resource-resource dependencies to input raw values or mocks.

## Test Plan

Existing unit tests succeed. New unit tests of nesting behavior with raw values (such as an IOManager).

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/build_resources.py
+++ b/python_modules/dagster/dagster/_core/execution/build_resources.py
@@ -115,22 +115,26 @@ def build_resources(
 def wrap_resources_for_execution(
     resources: Optional[Mapping[str, Any]] = None
 ) -> Dict[str, ResourceDefinition]:
+    return (
+        {
+            resource_key: wrap_resource_for_execution(resource)
+            for resource_key, resource in resources.items()
+        }
+        if resources
+        else {}
+    )
+
+
+def wrap_resource_for_execution(resource: Any) -> ResourceDefinition:
     from dagster._config.pythonic_config import ConfigurableResourceFactory, PartialResource
 
-    resources = check.opt_mapping_param(resources, "resources", key_type=str)
-    resource_defs = {}
     # Wrap instantiated resource values in a resource definition.
     # If an instantiated IO manager is provided, wrap it in an IO manager definition.
-    for resource_key, resource in resources.items():
-        # Wrap instantiated resource values in a resource definition.
-        # If an instantiated IO manager is provided, wrap it in an IO manager definition.
-        if isinstance(resource, (ConfigurableResourceFactory, PartialResource)):
-            resource_defs[resource_key] = resource.get_resource_definition()
-        elif isinstance(resource, ResourceDefinition):
-            resource_defs[resource_key] = resource
-        elif isinstance(resource, IOManager):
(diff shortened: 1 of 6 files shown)


benpankow:
* master
  * **PR #14807** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14807" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14807?utm_source=stack-comment).

python_modules/dagster/dagster/_core/execution/build_resources.py
```diff
@@ -115,22 +115,26 @@ def the_resource():
...
-
-    return resource_defs
+    if isinstance(resource, (ConfigurableResourceFactory, PartialResource)):
+        return resource.get_resource_definition()
+    elif isinstance(resource, ResourceDefinition):
+        return resource
+    elif isinstance(resource, IOManager):
+        return IOManagerDefinition.hardcoded_io_manager(resource)
+    else:
+        return ResourceDefinition.hardcoded_resource(resource)
```
schrockn, line 140: is this a behavior change or a refactor?

schrockn (changes requested):
Mostly about documentation. Would love to see tests changed in place so that we can see behavior changes in action, and put in example in the PR summary.

python_modules/dagster/dagster/_core/execution/build_resources.py
benpankow: Refactor, so that we're able to wrap just a single resource in some cases.

benpankow:
None of the existing unit tests were modified, just moved to the new file - the main behavior changes are highlighted in new tests. 

Updated the PR description with an example - this is the case that #14751 ran into, since it was relying on `IOManagers` being supplied which weren't necessarily resources.

PR #14811, opened 2023-06-15
Unpack abritrary number of values

jmsanders:
Croniter < 1.4 returns 2 items
Croniter >= 1.4 returns 3 items

Continue to unpack what we need and throw away the rest. This will break if they ever change their API to return items in a different order.

https://github.com/kiorky/croniter/commit/09e51e13aae39640bc200ddb8a1dbb4a7423d29f

PR #14812, opened 2023-06-15
Consolidate temporary pins for example projects

jmsanders:
Suggested in:

https://github.com/dagster-io/dagster/pull/14811#pullrequestreview-1481827412

This gives us a common file with a brief explanation where we can introduce temporary pins to keep our example tests that install from PyPI working until the next version of Dagster is released.


jmsanders:
Per our discussion, I also opened this GitHub discussion to track these kinds of suggested pins. https://github.com/dagster-io/dagster/discussions/14817

examples/user_pins.txt
```diff
@@ -0,0 +1,14 @@
...
+# typically don't patch previously released versions of Dagster to
+# introduce a pin. Instead, introduce a pin or a backward compatible fix
+# in the next release of Dagster.
+
+# We run automated tests to install the latest version Dagster from PyPI
+# and verify that these example projects work. We use this file to
+# temporarily introduce pins for these tests until the next version of
+# Dagster is published to PyPI.
+
+croniter<1.4.0
```
alangenfeld, line 14: can already stack remove this

alangenfeld (approved):
i like this

alangenfeld:
`user_pins.txt` is sort of odd name i think, maybe `workaround_pins`, `temp_pins` or something?

PR #6304, opened 2022-01-21
[context logging 1/n] Log user events via context

dpeng817:
https://github.com/dagster-io/dagster/issues/6394
Proposed API: 
log events like so
```python
@op
def my_op(context):
    context.log_event(AssetMaterialization(...))
```

**Things to do**
- [x] Basic implementation
- [x] Input lineage information
- [x] expand test cases

**Considerations**
- The current approach maintains ordering in both the event log and among yielded user events. 

**Problems**
- We now go through step inputs list twice for each step. This is an artifact from the chosen API, and could be changed.

PR #14814, opened 2023-06-15
make AssetExecutionContext subclass of OpExecutionContext

alangenfeld:
Refine `AssetExecutionContext` to extend `OpExecutionContext` and set-up methods to be moved 

## How I Tested These Changes

existing tests

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/context/compute.py
+++ b/python_modules/dagster/dagster/_core/execution/context/compute.py
@@ -1,10 +1,20 @@
 from abc import ABC, abstractmethod
-from typing import AbstractSet, Any, Dict, Iterator, List, Mapping, Optional, Sequence, Set, cast
-
-from typing_extensions import TypeAlias
+from typing import (
+    AbstractSet,
+    Any,
+    Dict,
+    Iterator,
+    List,
+    Mapping,
+    Optional,
+    Sequence,
+    Set,
+    Union,
+    cast,
+)
 
 import dagster._check as check
-from dagster._annotations import experimental, public
+from dagster._annotations import deprecated, experimental, public
 from dagster._core.definitions.assets import AssetsDefinition
 from dagster._core.definitions.data_version import (
     DataProvenance,
@@ -239,16 +249,6 @@ class OpExecutionContext(AbstractComputeExecutionContext):
 
     @public
     @property
-    def assets_def(self) -> AssetsDefinition:
-        assets_def = self.job_def.asset_layer.assets_def_for_node(self.node_handle)
-        if assets_def is None:
-            raise DagsterInvalidPropertyError(
-                f"Op '{self.op.name}' does not have an assets definition."
-            )
-        return assets_def
-
-    @public
-    @property
     def has_partition_key(self) -> bool:
         """Whether the current run is a partitioned run."""
         return self._step_execution_context.has_partition_key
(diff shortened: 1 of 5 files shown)


alangenfeld:
* master
  * **PR #14814** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14814" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14814?utm_source=stack-comment).

python_modules/dagster/dagster/_core/execution/context/compute.py
```diff
@@ -281,17 +281,181 @@ def partition_time_window(self) -> TimeWindow:
...
+
+    def add_output_metadata(
+        self,
+        metadata: Mapping[str, Any],
+        output_name: Optional[str] = None,
+        mapping_key: Optional[str] = None,
+    ) -> None:
+        """Add metadata to one of the outputs of an op.
+
+        This can only be used once per output in the body of an op. Using this method with the same output_name more than once within an op will result in an error.
```
schrockn, line 360: Btw I think this comment is out of date. I think sandy changed this so that you can call it multiple times and it merges/builds it as appropriate.

python_modules/dagster/dagster/_core/execution/context/compute.py
```diff
@@ -601,11 +731,12 @@ def get_asset_provenance(self, asset_key: AssetKey) -> Optional[DataProvenance]:
...
-            None if record is None else extract_data_provenance_from_entry(record.event_log_entry)
-        )
+        return super().get_asset_provenance(asset_key)
 
 
-AssetExecutionContext: TypeAlias = OpExecutionContext
+def build_execution_context(
+    step_context: StepExecutionContext,
+) -> Union[OpExecutionContext, AssetExecutionContext]:
+    if step_context.step_materializes_assets:
```
schrockn, line 740: name predates you, but I actually think this name is quite deceiving, as ops can yield `AssetMaterialization` objects.

python_modules/dagster/dagster/_core/execution/context/compute.py
```diff
@@ -399,9 +570,10 @@ def asset_partitions_def_for_output(self, output_name: str = "result") -> Partit
 
         return result
 
+    @deprecated
```
schrockn, line 573: really wish this was named `deprecated_in_docs` or something because I had to check that this didn't emit runtime warnings.

schrockn (approved):

python_modules/dagster/dagster/_core/execution/context/compute.py
alangenfeld: ya i had a similar reaction - what do you think would be better? this is internal so we can change it

python_modules/dagster/dagster/_core/execution/context/compute.py
alangenfeld: checked the impl - this comment is still accurate

python_modules/dagster/dagster/_core/execution/context/system.py
```diff
@@ -833,15 +833,18 @@ def op_config(self) -> Any:
...
-    def step_materializes_assets(self) -> bool:
-        step_outputs = self.step.step_outputs
-        if len(step_outputs) == 0:
-            return False
-        else:
+    def is_sda_step(self) -> bool:
+        """Whether this step corresponds to a software define asset, inferred by presence of asset info on outputs.
+
+        note: ops can materialize assets as well.
+        """
```
alangenfeld, line 840: `is_sda_step` ?

python_modules/dagster/dagster/_core/execution/context/compute.py
```diff
@@ -281,17 +281,181 @@ def partition_time_window(self) -> TimeWindow:
...
+        """Pops and yields all user-generated events that have been recorded from this context.
+
+        If consume_events has not yet been called, this will yield all logged events since the beginning of the op's computation. If consume_events has been called, it will yield all events since the last time consume_events was called. Designed for internal use. Users should never need to invoke this method.
+        """
+        events = self._events
+        self._events = []
+        yield from events
+
+    @public
+    def log_event(self, event: UserEvent) -> None:
```
schrockn, line 321: @alangenfeld revisiting the naming here. I think `log_event` is misleading since it implies that it *immediately* writes which it does not.

```python
context.log.info("foo")
context.log_event(...)
context.log.info("bar)
```

prints "foo", then "bar", then the event at at end, which I think it very counterintuitive.

alangenfeld: this PR is just moving things around, the blame rev for `log_event` is https://github.com/dagster-io/dagster/pull/6304/

i've been opting for `report_`  so think `report_event` would make sense here if want to go through the deprecation dance

PR #14819, opened 2023-06-15
Actually document the GCSComputeLogManager

PeterJCLaw:
## Summary & Motivation

Previously the configuration snippet for the `AzureBlobComputeLogManager` was incorrectly duplicated under the section for the `GCSComputeLogManager`. Replace that snippet with one derived from the docstring for the `GCSComputeLogManager`.

## How I Tested These Changes

Untested.


erinkcochran87:
👋 This was addressed in https://github.com/dagster-io/dagster/pull/15803, so I'm gonna close this out. Thanks for reporting it!

PR #14822, opened 2023-06-15
1.4 remove asset methods from OpExecutionContext

alangenfeld:
complete the migration of asset related methods off of `OpExecutionContext` on to `AssetExecutionContext`

## How I Tested These Changes

existing coverage

### DIFF
--- a/docs/sphinx/sections/api/apidocs/execution.rst
+++ b/docs/sphinx/sections/api/apidocs/execution.rst
@@ -73,6 +73,10 @@ Contexts
   :members:
   :inherited-members:
 
+.. autoclass:: AssetExecutionContext
+  :members:
+  :inherited-members:
+
 .. autofunction:: build_op_context
 
 .. autoclass:: TypeCheckContext
--- a/python_modules/dagster/dagster/_core/execution/context/compute.py
+++ b/python_modules/dagster/dagster/_core/execution/context/compute.py
@@ -1,10 +1,20 @@
 from abc import ABC, abstractmethod
-from typing import AbstractSet, Any, Dict, Iterator, List, Mapping, Optional, Sequence, Set, cast
-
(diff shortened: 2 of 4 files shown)


alangenfeld:
* master
  * **PR #14814** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14814" target="_blank">Graphite</a> 
    * **PR #14822** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14822" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14822?utm_source=stack-comment).

sryza:
Is this not a breaking change? At least for type warnings?

alangenfeld:
> Is this not a breaking change? At least for type warnings?

ya thats why this is waiting til `1.4.0` , do you think requiring users using asset related context methods to have to update to `AssetExecutionContext` to satisfy their type checkers should wait til 2.0?

sryza:
> do you think requiring users using asset related context methods to have to update typehint to AssetExecutionContext to satisfy their type checkers should wait til 2.0?

Hmm yeah this is an interesting question. I do think there's benefit from a user perspective of having that guarantee, though I don't think we strictly have to do this. What do you see as the main advantage of removing them?

alangenfeld:
> What do you see as the main advantage of removing them?

I don't think theres a particularly strong case:
* cleaning things up 
* encouraging adoption of `AssetExecutionContext` which maybe someday we want to break away from being a child of `OpExecutionContext` 

we could certainly defer this

sryza:
did we decide not to go this direction?

alangenfeld:
yea this is invalid after moving back towards just having the same object with a type alias

PR #14823, opened 2023-06-15
fix: Handle multi-dimensional partitions in W&B integration

chrishiste:
## Summary & Motivation
A client who uses the W&B integration contacted our customer support regarding an issue with Multi-Partitions.

Dagster, by default, adds a pipe (|) between partitions. However, W&B doesn't permit the usage of pipes in an Artifact name. To resolve this, the present Pull Request replaces all pipe characters with dashes.

## How I Tested These Changes
The modifications were tested and verified locally.

### DIFF
--- a/python_modules/libraries/dagster-wandb/dagster_wandb/io_manager.py
+++ b/python_modules/libraries/dagster-wandb/dagster_wandb/io_manager.py
@@ -273,8 +273,8 @@ class ArtifactsIOManager(IOManager):
 
                 # We replace the | character with - because it is not allowed in artifact names
                 # The | character is used in multi-dimensional partition keys
-                artifact_name = artifact_name.replace("|", "-" )
-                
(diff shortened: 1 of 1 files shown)


OwenKephart (changes requested):
Thanks for looking into this! The change seems reasonable, mind just running `make black` at the top level to fix a formatting issue?

Edit: ah an also you'll need to cast the `artifact_name` to a str type to avoid any typing issues, as in theory the type system thinks it can be None at this point (make pyright)

chrishiste:
@OwenKephart Sorry about that and thanks for the tip. Should be all good now :)

OwenKephart (approved):
Thanks for the contribution!

OwenKephart:
Just merged this in, it'll go out in next week's release :)

chrishiste:
@OwenKephart Thanks a lot for the help :)

PR #14825, opened 2023-06-15
Storage method for `get_latest_storage_id_by_partition`

OwenKephart:
## Summary & Motivation

Originally, this was going to be `get_latest_materialization_storage_id_by_partition`, but then I realized that there are futures in which we'd want to call this for AssetObservation and AssetMaterializationPlanned events, so I figured why limit it.

Factors out a pre-existing subquery into its own thing that is now used for both this new method and `get_latest_asset_partition_materialization_attempts_without_materializations`.

I'd also like to add another method like this, but joining on the asset_event_tags table


clairelin135 (approved):
LGTM

PR #14837, opened 2023-06-16
Test ParentOutdatedAutoMaterializeCondition

johannkm:
First pass. I think how we're treating multi assets seems a little shaky- both before and after this diff. Should an asset with no dependencies in a multi asset ever say 'waiting on parent'? I think ideally the evaluations UI would show that the assets are grouped in a single evaluation...

---

Update: just adding a test case

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -729,7 +729,7 @@ def determine_asset_partitions_to_auto_materialize(
         asset_graph: AssetGraph,
         candidate: AssetKeyPartitionKey,
         parent_partitions_result: ParentsPartitionsResult,
-    ) -> bool:
+    ) -> Tuple[bool, Optional[ParentOutdatedAutoMaterializeCondition]]:
         from dagster._core.definitions.external_asset_graph import ExternalAssetGraph
 
         for parent in parent_partitions_result.parent_partitions:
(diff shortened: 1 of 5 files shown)


johannkm:
* master
  * **PR #14837** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14837" target="_blank">Graphite</a>  👈
    * **PR #14838** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14838" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14837?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -732,7 +732,7 @@ def will_be_materialized_for_freshness(asset_partition: AssetKeyPartitionKey) ->
     def parents_will_be_reconciled(
         asset_graph: AssetGraph,
         candidate: AssetKeyPartitionKey,
-    ) -> bool:
+    ) -> Tuple[bool, Optional[ParentOutdatedAutoMaterializeCondition]]:
```
johannkm, line 735: i don't love this pattern, taking suggestions. We'll probably need it on most of the helper methods eventually

python_modules/dagster/dagster/_core/definitions/auto_materialize_condition.py
```diff
@@ -56,9 +59,15 @@ class MissingAutoMaterializeCondition(NamedTuple):
...
 class ParentOutdatedAutoMaterializeCondition(NamedTuple):
     """Indicates that this asset should be skipped because one or more of its parents are outdated.
+    This could be because the parent isn't going to run, or it could be that the parent is going to be materialized
+    but in a separate run due to different partitions or repositories.
     """
 
     decision_type: AutoMaterializeDecisionType = AutoMaterializeDecisionType.SKIP
+    parent_asset_key: Optional["AssetKey"] = None
+    parent_will_materialize: Optional[bool] = None
+    different_partitions: Optional[bool] = None
```
sryza, line 69: From the name, I expected the type of this would be a collection. Something like `parent_has_different_partitions_def` would be a clearer name,

sryza (commented):
A couple questions:
1. What if we're waiting on multiple parents?
2. How difficult would it be to traverse up through the graph and find the ancestor that we're actually waiting on?

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -745,30 +745,42 @@ def parents_will_be_reconciled(
...
                 ) is not asset_graph.get_repository_handle(parent.asset_key):
-                    return False
+                    return False, ParentOutdatedAutoMaterializeCondition(
+                        parent_asset_key=parent.asset_key,
+                        parent_will_materialize=True,
+                        different_repositories=True,
+                    )
 
-        return True
+        return True, None
```
OwenKephart, line 783: what if we just had this return the AutoMaterializeCondition? then we could have a separate function to determine if that should be treated as a True or False

OwenKephart (commented):
Echoing Sandy's question, I think we'd want similar logic for the ParentUpdatedAutoMaterializeCondition (basically, the positive case). This makes me think the logic for determining the parent asset key should go in the is_reconciled logic (or thereabouts), as this is what's actually determining what we're waiting on

python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/auto_materialize_policy_scenarios.py
```diff
@@ -136,6 +138,28 @@ def with_auto_materialize_policy(
...
+            **{
+                ("hourly", p): {MissingAutoMaterializeCondition()}
+                for p in hourly_partitions_def.get_partition_keys_in_range(
+                    PartitionKeyRange(start="2013-01-05-04:00", end="2013-01-07-03:00")
+                )
+            },
+            ("daily", "2013-01-05"): {
+                ParentOutdatedAutoMaterializeCondition(
+                    parent_asset_key=AssetKey("hourly"),
+                    parent_will_materialize=True,
```
johannkm, line 151: There's a strange non deterministic error here, where sometimes parent_will_materialize is True and sometimes False for this partition https://www.diffchecker.com/ENWRhbex/

Haven't been able to figure it out yet

sryza:
> Could you describe the scenario you're thinking of? We've chatted about the case where A->B->C and B skips, so C doesn't show anything, but that's slightly separate from this diff since that'd be expanding when we emit skip reasons

Asset graph:

```
A ---> B ---> E
 \-> C -> D -/
```

A and B are re-materialized. E has new data for its parent B, but is waiting on C.

OwenKephart (approved):

PR #14838, opened 2023-06-16
ParentOutdatedAutoMaterializeCondition gql

johannkm:
Add field to gql for which assets we're waiting on

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql/schema/auto_materialize_asset_evaluations.py
+++ b/python_modules/dagster-graphql/dagster_graphql/schema/auto_materialize_asset_evaluations.py
@@ -18,6 +18,7 @@ from dagster._core.scheduler.instigation import AutoMaterializeAssetEvaluationRe
 
 from dagster_graphql.schema.errors import GrapheneError
 
+from .asset_key import GrapheneAssetKey
 from .util import non_null_list
 
 GrapheneAutoMaterializeDecisionType = graphene.Enum.from_enum(AutoMaterializeDecisionType)
@@ -77,6 +78,8 @@ class GrapheneMissingAutoMaterializeCondition(graphene.ObjectType):
 
 
 class GrapheneParentOutdatedAutoMaterializeCondition(graphene.ObjectType):
+    waitingOnAssetKeys = graphene.List(graphene.NonNull(GrapheneAssetKey))
+
     class Meta:
         name = "ParentOutdatedAutoMaterializeCondition"
         interfaces = (GrapheneAutoMaterializeConditionWithDecisionType,)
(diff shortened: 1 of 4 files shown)


johannkm:
* master
  * **PR #14837** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14837" target="_blank">Graphite</a> 
    * **PR #14838** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14838" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14838?utm_source=stack-comment).

python_modules/dagster-graphql/dagster_graphql/schema/auto_materialize_asset_evaluations.py
```diff
@@ -142,7 +149,11 @@ def create_graphene_auto_materialize_condition(
         )
     elif isinstance(condition, ParentOutdatedAutoMaterializeCondition):
         return GrapheneParentOutdatedAutoMaterializeCondition(
-            decisionType=condition.decision_type, partitionKeysOrError=partition_keys_or_error
+            decisionType=condition.decision_type,
+            partitionKeysOrError=partition_keys_or_error,
+            parentAssetKey=condition.parent_asset_key,
+            differentPartitions=condition.different_partitions,
+            differntRepositories=condition.different_repositories,
```
OwenKephart, line 156: differnt

OwenKephart (commented):
Not sure if this PR or the other one is better for this discussion, but how did we arrive at this set of information to display to the user? In particular, do we need differentPartitions and differentRepositories?

johannkm:
Maybe not, it seemed like we easily had that information so might as well surface more details about why we're waiting

OwenKephart:
@johannkm my feeling was that it might make sense to just link out to the asset definition (from the key), and then users can discover that info for themselves.

not durable if the partitions def or repository changes, but I think that's not a huge deal

johannkm:
I think there's value in being explicit but I don't feel strongly. It's kind of a confusing thing to explain either way ("waiting on upstream instead of materializing in same run as upstream because it has a different partitions def than at least one asset that was selected to materialize with the upstream")

OwenKephart:
@johannkm yeah definitely reasonable -- I guess the way that I look at it, it's actually the _other_ behavior (grouping downstream assets into the same run) that's doing the "special" thing, and that's where "ParentUpdatedMaterializationCondition" doesn't quite make sense (the downstream asset's parent hasn't been updated yet, it just will be by the time that it's executed in the run we kicked off).

So I think it still feels fairly natural to just tell them "we're waiting for X asset to update"

johannkm:
Removed the additional details about why we're waiting, only added the set of asset keys

sryza (approved):
~~This looks good. Merging. Thanks @renzhe-brian!~~

Oops, wrong PR.

sryza:
However, this PR also LGTM @johannkm

Issue #14690, opened 2023-06-08
Enable box/lasso selection of assets in the DAG view


combiz:
### What's the use case?

Allow arbitrary sets of assets to be materialized in dagit without defining an arbitrary job.

### Ideas of implementation

Add a box or lasso select tool for the DAG view in dagit.
Allow any asset to be shift clicked (currently this seems hit-and-miss when assets are in different groups).

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

sryza:
Thanks for filing @combiz - I've added this to the queue for our UI team to look at

bengotow:
Taking a look at this today -- I'm able to reproduce some funkiness with shift-clicking and I think we can get that fixed. The idea is that shift-clicking select the assets in the lineage between your last-selected asset and the one you just selected, but I found an example below where it skips one asset for some reason.

I think we'll probably hold off on adding lasso / box selection to the DAG for now, but we've been talking about adding multiple pointer modes (zoom, select, pan) and I think it dovetails nicely into that discussion.

Thanks for flagging this!

![image](https://github.com/dagster-io/dagster/assets/1037212/0dc1e49f-09de-4670-94e4-1d147275f37f)

PR #14839, opened 2023-06-16
[ui] Fix issue shift-selecting assets with multi-component asset keys

bengotow:
## Summary & Motivation

This is a small fix for a bug raised in https://github.com/dagster-io/dagster/issues/14690. The remaining ideas in the issue will likely get resolved when we add different cursor modes to the asset graph.

The issue was caused by the fact that the "opsInRange" helper was a holdover from an old op graph feature and was using a different stringification method. If your path from asset A -> B included an asset with a multi-component asset key, it was not selected.

## How I Tested These Changes

I think it'd be good to bring more of this under test but right now this page won't mount in jest's test env. I think we could refactor this code so that the "Selection" is managed by a selection object which can be tested separately from the component UI, but I'm not sure it's worth the lift at the moment.

### DIFF
--- a/js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
+++ b/js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
@@ -175,15 +175,20 @@ const AssetGraphExplorerWithData: React.FC<WithDataProps> = ({
       let nextOpsNameSelection = token;
 
       if (e.shiftKey || e.metaKey) {
+        // Meta key adds the node you clicked to your existing selection
         let tokensToAdd = [token];
-        if (e.shiftKey && lastSelectedNode && node) {
-          const tokensInRange = opsInRange({
-            graph: assetGraphData,
-            from: lastSelectedNode,
-            to: node,
-          });
-          if (tokensInRange.length) {
-            tokensToAdd = tokensInRange;
+
+        // Shift key adds the nodes between the node you clicked and your existing selection.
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
```diff
@@ -175,15 +175,20 @@ const AssetGraphExplorerWithData: React.FC<WithDataProps> = ({
...
+        // To better support clicking a bunch of leaves and extending selection, we try to reach
+        // the new node from each node in your current selection until we find a path.
+        if (e.shiftKey && selectedGraphNodes.length && node) {
+          for (let ii = selectedGraphNodes.length - 1; ii >= 0; ii--) {
+            const from = selectedGraphNodes[ii]!;
+            const tokensInRange = assetKeyTokensInRange({from, to: node, graph: assetGraphData});
+            if (tokensInRange.length) {
+              tokensToAdd = tokensInRange;
+              break;
+            }
```
hellendag, line 191: Could do this to avoid the non-null assertion:

```
const reversed = [...selectedGraphNodes].reverse();
for (const from of reversed) {
  ...
}
```

hellendag (approved):

js_modules/dagit/packages/core/src/asset-graph/AssetGraphExplorer.tsx
bengotow: oh good call! will switch it out for this 👍

PR #14840, opened 2023-06-16
[ui] Fix asset stale causes to a popover, stop truncating the list

### LINKED ISSUE #14654: Display all provenance information. Do not hide behind "and N more..."
### What's the use case?

We have useful provenance information but we hide it sometimes in the UI, with no way of viewing it.

Here's an example from our own system:

Screen Shot 2023-06-06 at 3 53 24 PM

This is frustrating because it is often that missing piece of information that is necessary for debugging and understandability. Furthermore there is no way to expand this.

We should display all this information somewhere if someone needs it.

### Ideas of implementation

_No response_

### Additional information

_No response_

bengotow:
## Summary & Motivation

Resolves #14654

This PR switches the on-hover display of stale causes from a Tooltip to a Popover, which allows you to move the mouse into the overlay without it closing. We already use this on the run timeline to show a list of runs on hover.

This new rendering style means that the list can scroll when there are many stale causes and we can also linkify the references to other assets (they now link to those asset's detail pages). 

image

Note: These new ones are slightly larger, primarily because I made the styling match the padding, etc. of the run timeline popover:
image
image

## How I Tested These Changes

I tested these changes manually by viewing stale causes overlaid on the asset DAG and overlaid via the Stale tag in the asset catalog / asset details pages. I used a hack to give myself many stale causes to ensure that the max-height and scrolling behavior works as expected.

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/Stale.tsx
+++ b/js_modules/dagit/packages/core/src/assets/Stale.tsx
@@ -16,6 +16,7 @@ import React from 'react';
 import {Link} from 'react-router-dom';
 
 import {displayNameForAssetKey, LiveDataForNode} from '../asset-graph/Utils';
+import {AssetNodeKeyFragment} from '../asset-graph/types/AssetNode.types';
 import {AssetKeyInput, StaleCauseCategory, StaleStatus} from '../graphql/types';
 
 import {assetDetailsPathForKey} from './assetDetailsPathForKey';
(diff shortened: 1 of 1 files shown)


salazarm (approved):
lgtm!

bengotow:
> image Why does the string "whatever3" show up twice here?

@sryza Hmm yeah it's not ideal - I can do some string replacement and strip out the first reference to the asset key client-side. I'm not sure why the backend implementation bakes in the asset key as well as providing it as the dependency.

bengotow:
@sryza - I added a bit of client-side processing to these, so that if the stale `reason` ends in the dependency key, that one is linkified and it's not repeated in parenthesis. From the backend impl, it looks like it just applies to these two but should clean up that case a bit: 

image

sryza:
Ah got it. That looks good!

Issue #14588, opened 2023-06-01
When displaying auto-materialize policies in the UI, mention skip reasons


OwenKephart:
### What's the use case?

![Screenshot 2023-06-01 at 4 14 46 PM](https://github.com/dagster-io/dagster/assets/22457492/c3c998cd-b52e-41f9-aac7-a9307a7975fe)

Even if one of those things happens, we won't always auto-materialize the asset. This can be confusing to people new to the system.

### Ideas of implementation

Add skip reasons to the copy

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

PR #14848, opened 2023-06-16
When displaying auto-materialize policies in the UI, mention skip rea…

sryza:
…sons

## Summary & Motivation

Fixes https://github.com/dagster-io/dagster/issues/14588


OwenKephart (approved):
"Have newer data" is easier to parse, but less accurate (at least right now). However, as we plan to incorporate data versions into this logic, I think this is ok.

Issue #14744, opened 2023-06-11
AutoMaterializePolicy.lazy should wait for Freshness policy trigger for assets that are missing materialization


nixent:
### What's the use case?

When Auto-materialize daemon starts for assets with lazy policy, it picks missing assets and starts materialization. For Prod environment it results in full load of all assets initiated in same time, while in normal operations materializations would be spread in time according to their Freshness policies. In Dev/Test environments developers often need to test handful of models and don't need full set of models, deferring missing assets materialization until their Freshness policy would allow tests based on Freshness policy without having materializations for all missing assets.

It is possible to define custom `AutoMaterializationPolicy` with `on_missing = false`, however for dbt scenario, it would be nice to control it in the dbt config, which supports two types: lazy and eager

### Ideas of implementation

Freshness policy evaluation can be based on Daemon start time. If asset is not materialized yet, Daemon start time is the reference point for calculation of the asset freshness.
Example:
DAG: A->B->C
Asset A is materialized manually, B and C are missing. C has freshness policy of 1h.
When daemon starts it should wait for 1h and trigger materializations of A->B->C

Maybe making additional "super-lazy" type for dagster-dbt with `on_missing = false` is the easiest way to achieve it

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

sryza:
If I understand correctly, the request is to be able to have dbt assets with lazy auto-materialize policies and avoid materializing missing assets?

It sounds like there are a couple gaps here in what we have:
- In Python, we don't have a blessed way of setting `on_missing=False`. You can do it with the `AutoMaterializePolicy` constructor, but we currently discourage the use of that constructor.
  - So we should probably just add it as a parameter to the `lazy` and `eager` methods.
- In dbt, we don't have an easy way of setting `on_missing=False`. You can do it with the `node_info_to_auto_materialize_policy_fn` on `load_assets_from_dbt_manifest`, but that's a pain.
  - We should probably have a way of specifying it in dbt config that uses the same terms (e.g. "on_missing") that are used with the Python API.

cc @OwenKephart

nixent:
It boils down to definition of "**lazy**" AutoMaterializationPolicy, my understanding of "lazy" is "do as little as possible, do nothing until it is really necessary", with `on_missing=True`, that is not the case.
`lazy()` is defined as 
```
AutoMaterializePolicy(
            on_missing=True,
            on_new_parent_data=False,
            for_freshness=True,
[...]

PR #14849, opened 2023-06-16
for lazy auto-materialize policies, set on_missing=False

sryza:
## Summary & Motivation

Lazy auto-materialize policies should only materialize assets if needed for freshness.

Fixes https://github.com/dagster-io/dagster/issues/14744

## How I Tested These Changes

Added tests to verify that:
- Lazy asset isn't materialized when there's no freshness policy
- Lazy asset is materialized when there is a freshness policy

### DIFF
--- a/python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/auto_materialize_policy_scenarios.py
+++ b/python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/auto_materialize_policy_scenarios.py
@@ -1,4 +1,3 @@
-import datetime
 from typing import Sequence
 
 from dagster import (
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/auto_materialize_policy_scenarios.py
```diff
@@ -340,4 +349,14 @@ def with_auto_materialize_policy(
...
     ),
+    "auto_materialize_policy_lazy_missing": AssetReconciliationScenario(
+        assets=single_lazy_asset,
+        unevaluated_runs=[],
+        expected_run_requests=[],
+    ),
+    "auto_materialize_policy_lazy_freshness_missing": AssetReconciliationScenario(
+        assets=single_lazy_asset_with_freshness_policy,
+        unevaluated_runs=[],
+        expected_run_requests=[run_request(asset_keys=["asset1"])],
```
OwenKephart, line 360: maybe set the expected_conditions thing here as well?

OwenKephart (approved):

PR #14864, opened 2023-06-20
[1.5] feat(dbt)!: do not parse asset group from dbt project structure

rexledesma:
## Summary & Motivation
We want users to explicitly set group name on their dbt models, using either `+group` or `+meta` to set `dagster.group`.

## How I Tested These Changes
bk


OwenKephart (commented):
This is a pretty big breaking change for existing users, especially as most people won't have migrated to setting group properties in their dbt project yet.

Seems like it wouldn't harm much to keep this as the default, no?

rexledesma:
> This is a pretty big breaking change for existing users, especially as most people won't have migrated to setting group properties in their dbt project yet.
> 
> Seems like it wouldn't harm much to keep this as the default, no?

Yeah, this is a breaking change. But this behavior inevitably needs to be deprecated since this sort of implicit configuration is unintuitive for the user. The confusion will be more apparent when more users are on dbt `1.5.0` and are starting to use dbt native groups: the implicit and explicit groups sound like a to reason about.

I am fine to budge on the timeline: we can wait to deprecate this in Dagster `1.5.0`, and include a warning to indicate the deprecation of behavior. We can also do the same for the previous configuration path for `dagster_freshness_policy` and `dagster_auto_materialize_policy`.

OwenKephart:
@rexledesma There's a reasonable migration path here (just set up your project configs such that groups or the dagster meta are applied to subdirectories), so that timeline sounds more reasonable.

rexledesma:
Superseded by https://github.com/dagster-io/dagster/pull/15346.

PR #14869, opened 2023-06-20
Utility to fetch asset key from dbt id

clairelin135:
Adds a utility methods to `DbtManifest` to support defining python dependencies upstream and downstream of a dbt model:
- `get_asset_keys_for_dbt_source`
- `get_asset_key_for_dbt_model`

Would be great to get feedback on these API methods. I wasn't sure that users are aware of what the dbt unique ID is, so I thought it would be better to have specific methods based on the type of node.

How this would look if you wanted to define upstream/downstream SDA dependencies:
```python


@multi_asset(
    outs={
        name: AssetOut(key=asset_key)
        for name, asset_key in manifest.get_asset_keys_by_output_name_for_source("raw_data").items()
    },
)
def upstream_python_asset():
    ...

@dbt_assets(manifest=manifest)
def my_dbt_assets():
    ...


@asset(
    non_argument_deps={
        manifest.get_asset_key_for_model(
           "daily_order_summary", key_prefix=["duckdb", "dbt_schema"]
        )
    }
)
def downstream_python_asset():
    ...

```

Adds tests for these methods and locally viewed Dagit to confirm that the dependency graph displayed as intended.


python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -130,6 +131,46 @@ def get_asset_key_by_output_name(self, output_name: str) -> AssetKey:
...
+        if not node_info:
+            raise DagsterInvalidInvocationError(
+                f"Could not find a dbt node with unique_id: {unique_id}. A unique ID consists of"
+                " the node type (model, source, seed, etc.), project name, and node name in a"
+                " dot-separated string. For example: model.my_project.my_model"
+            )
+
+        return self.node_info_to_asset_key(node_info)
+
+    def get_asset_keys_for_dbt_source(self, source_name: str) -> Sequence[AssetKey]:
```
sryza, line 146: how can a source end up with multiple asset keys?

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -130,6 +131,46 @@ def get_asset_key_by_output_name(self, output_name: str) -> AssetKey:
...
+        }
+
+        if len(sources) == 0:
+            raise DagsterInvalidInvocationError(
+                f"Could not find a dbt source with name: {source_name}"
+            )
+
+        return sorted([self.get_asset_key_for_dbt_unique_id(key) for key in sources])
+
+    def get_asset_key_for_dbt_model(self, model_name: str) -> AssetKey:
```
sryza, line 160: I wonder if we can get away with leaving `dbt` out of the name, given that the object already pertains to dbt?

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -130,6 +131,46 @@ def get_asset_key_by_output_name(self, output_name: str) -> AssetKey:
...
         return self.node_info_to_asset_key(self.get_node_info_by_output_name(output_name))
 
+    def get_asset_key_for_dbt_unique_id(self, unique_id: str) -> AssetKey:
+        node_info = self.node_info_by_dbt_unique_id.get(unique_id)
+
+        if not node_info:
+            raise DagsterInvalidInvocationError(
+                f"Could not find a dbt node with unique_id: {unique_id}. A unique ID consists of"
+                " the node type (model, source, seed, etc.), project name, and node name in a"
+                " dot-separated string. For example: model.my_project.my_model"
```
rexledesma, line 141: Let's link https://docs.getdbt.com/reference/artifacts/manifest-json as well, since this page describes the structure of the `unique_id` (`<resource_type>.<package>.<resource_name>`)

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
rexledesma: A dbt source is referenced by a `source_name` and a `table_name`: https://docs.getdbt.com/reference/dbt-jinja-functions/source. A source can have multiple tables.

python_modules/libraries/dagster-dbt/dagster_dbt_tests/test_dbt_python_dependencies.py
```diff
@@ -0,0 +1,79 @@
...
+def test_asset_downstream_of_dbt_asset() -> None:
+    upstream_asset_key = AssetKey(["orders"])
+
+    @dbt_assets(manifest=test_dagster_metadata_manifest)
+    def my_dbt_assets():
+        ...
+
+    @asset(
+        non_argument_deps={
+            test_dagster_metadata_manifest.get_asset_key_for_dbt_unique_id(
```
rexledesma, line 37: I assume users won't be calling `get_asset_key_for_dbt_unique_id` directly, so let's use `get_asset_key_for_dbt_model` here instead.

python_modules/libraries/dagster-dbt/dagster_dbt_tests/test_dbt_python_dependencies.py
rexledesma, line 1: `test_dependencies.py`? `dbt` and `python` are implicit

rexledesma (changes requested):
Adding examples of upstream/downstream deps in the experimental example would be great, just to get a real sense of how this API shapes up against the sample project.

This can be done in a separate PR.

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
sryza: Ah I see. What's the scenario where we'd imagine someone using this method? Also, this reminds me that the new methods should have docstring.

rexledesma: A source (e.g. `fivetran`) could have multiple tables that are synced by one API call. So this would require the user to define a `@multi_asset` for the source.

rexledesma, line 146: Probably want to return a mapping of `table_name: str -> asset_key: AssetKey`, just to give some clarity to the user about what is producing the `AssetKey`
```suggestion
    def get_asset_keys_for_dbt_source(self, source_name: str) -> Mapping[str, AssetKey]:
```

sryza: I'm still not entirely following. When would someone be defining dependencies between non-dbt Dagster assets and dbt sources? I thought the whole point of dbt sources was to make source data available to dbt models.

That said, it seems like including this method is fairly low cost, so not a big deal if we don't have a super concrete use case for it. Mostly just curious.

rexledesma: This method has a concrete use case and it is to define software-defined assets for the dbt sources.

Like you said, the point of dbt sources is to define the source data for the dbt models. In dbt, this defines the upstream dependencies (i.e. we map the `(source_name, table_name)` to an `AssetKey`) of the dbt models that are external to the system.

So, the dependencies are already defined when the dbt project is read by the `dagster-dbt` integration. However, to materialize these dbt sources in Dagster, the user needs to define the logic associated with these upstream `AssetKey`'s.

This function, `get_asset_keys_for_dbt_source`, gives you the `AssetKey`'s that you need to define the software-defined assets for the dbt sources.

clairelin135: good call

examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/assets/dependencies.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@multi_asset(
+    outs={
+        name: AssetOut(key=asset_key)
+        for name, asset_key in manifest.get_asset_keys_by_output_name_for_source("raw_data").items()
+    },
+)
+def raw_inventory() -> Any:
+    return pd.DataFrame()
```
rexledesma, line 20: Can we handle the case where:
- A dbt `source_name` just consists of one `table_name`, and
- The user uses `@asset` instead of using `@multi_asset`?

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -130,6 +136,67 @@ def get_asset_key_by_output_name(self, output_name: str) -> AssetKey:
...
+            raise DagsterInvalidInvocationError(
+                f"Could not find a dbt source with name: {source_name}"
+            )
+
+        return {
+            output_name_fn(value): self.node_info_to_asset_key(value) for value in matching_nodes
+        }
+
+    def get_asset_key_for_model(
+        self, model_name: str, key_prefix: Optional[CoercibleToAssetKeyPrefix] = None
```
rexledesma, line 175: Not a fan of this `key_prefix` argument... but its existence precedes this PR

rexledesma (changes requested):
I belive we need to handle the case where the user reaches for `@asset` to define their dbt source.

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -130,6 +136,67 @@ def get_asset_key_by_output_name(self, output_name: str) -> AssetKey:
...
+                f"Could not find a dbt node with unique_id: {unique_id}. A unique ID consists of"
+                " the node type (model, source, seed, etc.), project name, and node name in a"
+                " dot-separated string. For example: model.my_project.my_model\n For more"
+                " information on the unique ID structure:"
+                " https://docs.getdbt.com/reference/artifacts/manifest-json"
+            )
+
+        return self.node_info_to_asset_key(node_info)
+
+    def get_asset_keys_by_output_name_for_source(self, source_name: str) -> Mapping[str, AssetKey]:
```
sryza, line 153: If I understand correctly, this method is here to facilitate a very specific pattern and wouldn't really be useful in other situations? If that's the case, I think we should:
- Make sure that this is explicit in the docstring and include a code example of that pattern
- Consider having it return a `Mapping[str, AssetOut]`

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -130,6 +136,67 @@ def get_asset_key_by_output_name(self, output_name: str) -> AssetKey:
...
+        }
+
+    def get_asset_key_for_model(
+        self, model_name: str, key_prefix: Optional[CoercibleToAssetKeyPrefix] = None
+    ) -> AssetKey:
+        """For a dbt model, return the dagster asset key for the model.
+
+        Args:
+            model_name (str): The name of the dbt model.
+            key_prefix (Optional[CoercibleToAssetKeyPrefix]): An optional prefix to prepend to the
```
sryza, line 181: In what situation would someone want to use this argument? Could they just use `with_prefix` instead?

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
sryza: "This is a convenience method that makes it easy to define a multi-asset that generates all the tables for a given dbt source".

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
clairelin135: This argument can be used when the upstream dbt assets have been loaded via `load_assets_from_modules(..., key_prefix=["blah"])`. In the downstream python asset, the `AssetIn` accepts either a key or a key prefix, but not both. 

I added the argument as a way to directly specify `AssetIn(key=manifest.get_asset_key_for_model(..., key_prefix=["blah"])` but I think it's probably better to do `AssetIn(key=manifest.get_asset_key_for_model(...).with_prefix(...))` instead.

rexledesma: Agree with using `manifest.get_asset_key_for_model(...).with_prefix(...)`. 

Again, this is separate from this PR, but we should definitely revisit the ergonomics of applying `key_prefix` via `load_assets_from_modules`: it feels odd that assets should be aware about how they are loaded after they have been defined.

sryza: (Not a blocker for this though – `with_prefix` is fine too)

sryza: but yes - `with_prefix`

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
clairelin135: It is more ergonomic for most cases for the method to return an AssetOut, but it would be hard for users to customize other properties on the asset out (i.e. make outputs non-required). For that reason I think we should still return a mapping of str to asset key

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
clairelin135: We can remove the `key_prefix` arg. Instead, users can just use `AssetKey(...).with_prefix(..)` instead

examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/assets/dependencies.py
```diff
@@ -0,0 +1,41 @@
...
+@dbt_assets(manifest=manifest)
+def my_dbt_assets(context: OpExecutionContext, dbt: DbtCli):
+    yield from dbt.cli(["build"], manifest=manifest, context=context).stream()
+
+
+@asset(
+    ins={"customers": AssetIn(key=manifest.get_asset_key_for_model("customers"))},
+)
+def cleaned_customers(customers: pd.DataFrame) -> Any:
+    return customers
```
rexledesma, line 41: Would this work without an IO manager? Can we just use `non_argument_deps` instead?

rexledesma (changes requested):
Almost there, just some nits.

I would also prefer that the introduction of `key` and `with_prefix` are in separate PR's since that is altering other framework-level APIs. We should loop in the essentials team on those separate changes.

clairelin135:
Breaking this PR out into:
- https://github.com/dagster-io/dagster/pull/14933
- https://github.com/dagster-io/dagster/pull/14932

PR #14843, opened 2023-06-16
[ui] Improve Runs list empty state for job

hellendag:
## Summary & Motivation

Resolves #14835.

When viewing the Runs tab on a job page, the empty state calls to action just redirect the user to generic Jobs and Assets pages.

Instead, use information about the job to show appropriate calls to action: "Launch a run" if it's a non-asset job and link to the Launchpad, "Materialize an asset" if it's an asset job and link to the Overview.

Screenshot 2023-06-16 at 3 58 55 PM
Screenshot 2023-06-16 at 3 58 39 PM

## How I Tested These Changes

View a never-run asset job, and a never-run regular job. Verify new empty states on Runs page for both, and verify that the buttons lead to useful places on those job pages.

PR #14885, opened 2023-06-21
[ui] Add optional EmptyState to RunTable

hellendag:
## Summary & Motivation

In https://github.com/dagster-io/dagster/pull/14843, I didn't handle the case where the user has set filters on the view. While fixing this, I noticed that the empty state could prevent the filters from rendering at all, since they're part of `RunTable`.

To resolve this, add an optional `emptyState` prop to `RunTable` so that an empty state can be defined by the component that uses `RunTable`.

Then, check for the presence of any filters to determine exactly what copy to show the user in the empty state case.

Screenshot 2023-06-21 at 10 52 31 AM

## How I Tested These Changes

View asset jobs and op jobs with and without filters. Verify that the correct empty states appear, and that their buttons navigate to the correct place.

View global Runs view, use filters to get to an empty state. Verify that it renders correctly, even though this RunTable has no custom empty state defined.

### DIFF
--- a/js_modules/dagit/packages/core/src/pipelines/PipelineRunsRoot.tsx
+++ b/js_modules/dagit/packages/core/src/pipelines/PipelineRunsRoot.tsx
@@ -62,7 +62,6 @@ export const PipelineRunsRoot: React.FC<Props> = (props) => {
 
   const repo = useRepository(repoAddress);
   const isJob = isThisThingAJob(repo, pipelineName);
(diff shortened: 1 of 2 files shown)


hellendag:
* master
  * **PR #14885** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14885" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14885?utm_source=stack-comment).

PR #14886, opened 2023-06-21
non_arg_deps rename 1/n - new deps arg for asset decorator

### LINKED ISSUE #15196: `non_argument_deps` successor API
### What's the use case?

make `non_argument_deps` a more first class concept by updating the API to be better named, and more ergonomic 

### Ideas of implementation

_No response_

### Additional information

_No response_

jamiedemaria:
## Summary & Motivation
Base PR for non_argument_deps successor. This PR adds a `deps` parameter to `@asset` and `@multi_asset` with the type `Sequence[Union[AssetsDefinition, SourceAsset, CoercibleToAssetKey]` and logs a deprecation warning if `non_argument_deps` is used. Other PRs in the stack handle internal parameter renames, call-site updates, etc.

resolves #15196 

New API

```python
@asset 
def this_asset_makes_a_table(snowflake: SnowflakeResource):
    with snowflake.get_connection() as conn:
        create_snowflake_table_1(conn)

@asset 
def this_asset_makes_another_table(snowflake: SnowflakeResource):
    with snowflake.get_connection() as conn:
        create_snowflake_table_2(conn)

@asset(
    deps=[this_asset_makes_a_table, this_asset_makes_another_table]
)
def this_asset_joins_tables(snowflake: SnowflakeResource):
    with snowflake.get_connection() as conn:
        join_tables(conn)
```

## How I Tested These Changes
new unit tests


jamiedemaria:
Discussion: What should the new parameter be named? 
Some options:
* `upstream_assets`
* `deps`

jamiedemaria:
Discussion: What should the type signature of the new parameter be?
Some options:
* `Optional[Set[CoercibleToAssetKey]]`
* `Optional[Set[Union[AssetKey, str]]]`
* `Optional[Set[Union[CoercibleToAssetKey, UpstreamAsset]]]` Where `UpstreamAsset` (name tbd) is a new class allowing you to specify asset key, partition mapping, etc
* `Optional[Set[Union[AssetKey, str, UpstreamAsset]]]`

related question - `Set` vs `Sequence` with some runtime checking for duplicated

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -60,6 +60,7 @@ def asset(
     key_prefix: Optional[CoercibleToAssetKeyPrefix] = None,
     ins: Optional[Mapping[str, AssetIn]] = ...,
     non_argument_deps: Optional[Union[Set[AssetKey], Set[str]]] = ...,
+    upstream_assets: Optional[Set[CoercibleToAssetKey]] = ...,
```
jamiedemaria, line 63: Discussion: What should the new parameter be named? 
Some options:
* `upstream_assets`
* `deps`

jamiedemaria, line 63: Discussion: What should the type signature of the new parameter be?
Some options:
* `Optional[Set[CoercibleToAssetKey]]`
* `Optional[Set[Union[AssetKey, str]]]`
* `Optional[Set[Union[CoercibleToAssetKey, UpstreamAsset]]]` Where `UpstreamAsset` (name tbd) is a new class allowing you to specify asset key, partition mapping, etc
* `Optional[Set[Union[AssetKey, str, UpstreamAsset]]]`

related question - `Set` vs `Sequence` with some runtime checking for duplicated

jamiedemaria: Some options we've discussed:
* `upstream_assets`
* `deps`

jamiedemaria: We've discussed a `Set` or `Sequence` containing some combination of the following:
* `CoercibleToAssetKey`
* `str`
* `List[str]`
* `AssetsDefinition`
* `UpstreamAsset` (name tbd) - a new class allowing you to specify asset key, partition mapping, etc

jamiedemaria: I am going to write up a PR that explores passing `AssetDefinitions` around

alangenfeld: * `materialize_after`

jamiedemaria: * `depends_on`

jamiedemaria, line 63: Discussion: How should we handle partition mappings?

jamiedemaria: Some ideas:
* Introduce `UpstreamAsset` class with a partition mapping attribute the user can set - ie `upstream_assets={UpstreamAsset(key="my_upstream_asset", partition_mapping=TimeWindowPartitionMapping(...))}`
* Some kind of class method on `AssetsDefinition` that returns an internal type - ie `upstream_assets={my_upstream_asset.with_partition_mapping(TimeWindowPartitionMapping(...))}`

jamiedemaria: Also wrote up a rough PR exploring the `UpstreamAsset` class https://github.com/dagster-io/dagster/pull/14908

jamiedemaria: Some reflections after writing those two PRs:
* passing asset definitions is very ergonomic. When I wrote the `UpstreamAsset` example, i missed using the definition python symbol
* If we think that passing around the full definition object is a bad pattern, we could instead have users pass `my_asset.key` which still lets you work with symbols 
* maybe a bit out there, but what are the consequences of allowing `AssetsDefinition` as a `CoercibleToAssetKey`? We'd need to decide how to handle multi_assets (either make a dependency on all assets w/in the multi asset or explicitly disallow it)

jamiedemaria: see https://github.com/dagster-io/dagster/pull/14908 for `UpstreamAsset` class that allows for partition mappings

schrockn: If we allow `AssetsDefinition` (which definitely has a certain appeal) I think the usage of `Set` versus `Sequence` becomes increasingly problematic. I think it is more reflective in reality to say we take a sequence of things and then do runtime checks/processing to unique-ify. 

The previous `non_argument_deps` as typed the way it was for good reason, since strings and asset keys don't play together nicely in a set:

```
from dagster import AssetKey

string = "foo"
key = AssetKey("foo")
set([string, key])
# prints {AssetKey(['foo']), 'foo'} in interpreter
```

So if we want to mix and match these things it needs to be a `Sequence` IMO

schrockn: The other behavior to consider if `AssetsDefinition` is provided is to only add dependencies to the "leaf" assets (those with no downstream dependences within that AssetsDefinition) to avoid cluttering the dependency graph.

jamiedemaria: another potential partitions option https://github.com/dagster-io/dagster/pull/14912

alangenfeld: supporting `AssetIn` i think is worth considering, from its API doc block it `Defines an asset dependency.`

schrockn:
Logging zoom discussion here with @jamiedemaria and @alangenfeld. We have soured on the name 'upstream_assets'. It heavily guides us to introduce a `UpstreamAsset` class which doesn't feel great for an "edge" type. It feels like a new thing, which we want to avoid.

We considered the following name:

- upstream_deps
    - “Deps” makes more sense since it is the “edge”. The proposed “upstream assets” is weird because in prototypes it contains partition mapping because that is an “edge” property
- deps
    - Very terse (+1) but does not indicate directionality (possibly confusing)
    - Think of it a Union[Node, Edge]
    - Terseness could cause confusion. “What on earth is a dep”
- asset_deps
    - Same issue with upstream assets where “edge” use cases are strange.
- upstreams: Not seriously considered
- parents: Not seriously considered
- ancestors: Not seriously considered
- depends_on
    - Union of types feels good because name is agnostic to target
    - Union[Node, NodeWithEdgeDetails]
    - Might be a bit strange if we have an “edge” object
    - Depends on is tough if we introduce an edge object.
- materialize_after
    - Collision with auto-materialization policies is disqualifying
    
We are leaning toward `deps`. It is terse and it provides a lot of option value to have both "nodes" and "edges" as arguments. `deps` is also used in other contexts throughout the system.

The other proposal we have is changing type to `Sequence[CoercibleToAssetKey, AssetsDefinition]` to allow for direct references.

```python

from dagster import asset

@asset
def parent_asset():
   ...
   
@asset(deps=[parent_asset])
def child_asset():
   ...

```

We also ruled out taking `AssetIn` as valid to `deps` as it pollutes the mental model by allowing i/o manager related things in this API.

sryza:
The concern about `upstream_assets` vs. it being an edge makes sense to me.

My only concern with `deps` is that there are other kinds of dependencies, e.g. package dependencies: imagine some sort of future world where each asset executes in its own container and we want to allow each asset to specify the packages it depends on.

I'm not entirely following the concern about `asset_deps`. A "dep" is an edge, and "asset" is a modifier that adds specificity about what kind of edge it is.

That said, I do love the terseness of `deps`, and I think it's easy to make the argument that asset dependencies are the main kind of dependency in this realm, so any other kind of dependency, like a package dependency, is the one that needs the modifier.

cc @schrockn @alangenfeld

schrockn:
@sryza I had similar concerns about `deps`, but in a slightly different domain. That was the potential confusion with execution-based dependencies, especially if we get cute with data quality execution (or anything else) and interleave those with assets in interesting ways. However I was persuaded that the asset-graph has primacy and therefore `deps` works. I think the same argument applies to "packages" and other future use cases.

alangenfeld:
I would also expect that if we had package dependencies, it would be a property on a nested or separate object from the asset itself, to support varying between "compute by running this function defined next to me",  "compute by creating an environment with these packages and running X", and "compute in this container ...".

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -130,8 +134,8 @@ def asset(
             contains letters, numbers, and _) and may not contain python reserved keywords.
         ins (Optional[Mapping[str, AssetIn]]): A dictionary that maps input names to information
             about the input.
-        non_argument_deps (Optional[Union[Set[AssetKey], Set[str]]]): Set of asset keys that are
-            upstream dependencies, but do not pass an input to the asset.
+        deps (Optional[Set[Union[AssetKey, str, Sequence[str]]]]): Set of asset keys that are upstream
```
jamiedemaria, line 137: can do a better job with this doc string

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -180,11 +186,21 @@ def my_asset(my_upstream_asset: int) -> int:
...
 
     def create_asset():
+        if non_argument_deps is not None and deps is not None:
+            raise DagsterInvalidDefinitionError(
+                "Cannot specify both deps and non_argument_deps. Use only deps instead."
+            )
+
+        upstream_asset_deps: Optional[Set[CoercibleToAssetKey]] = deps
+        if non_argument_deps is not None:
+            deprecation_warning("non_argument_deps", "X.X.X", "use parameter deps instead")
```
jamiedemaria, line 196: TODO - set deprecation version

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -130,8 +135,8 @@ def asset(
             contains letters, numbers, and _) and may not contain python reserved keywords.
         ins (Optional[Mapping[str, AssetIn]]): A dictionary that maps input names to information
             about the input.
-        non_argument_deps (Optional[Union[Set[AssetKey], Set[str]]]): Set of asset keys that are
-            upstream dependencies, but do not pass an input to the asset.
+        deps (Optional[Sequence[Union[AssetsDefinition, AssetKey, str, Sequence[str]]]]): Set of asset keys that are upstream
```
alangenfeld, line 138: align this with typehint above

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
alangenfeld: 2.0.0 i would guess - no reason to hard remove ahead of that i think

examples/non_argument_deps_successor/non_argument_deps_successor/assets.py
```diff
@@ -0,0 +1,23 @@
...
+
+# multi asset case, passing by AssetsDefinitions does not work, so pass by CoercibleToAssetKey instead
+@multi_asset(outs={"asset_1": AssetOut(), "asset_2": AssetOut()})
+def a_multi_asset():
+    return None, None
+
+
+@asset(deps=["asset_1", AssetKey("asset_2")])
+def downstream_of_multi_asset():
+    return None
```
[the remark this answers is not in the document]
jamiedemaria: ah should have clarified - I added this as an example for us as part of the review so we could quickly see what it looks like in code while test are being written. Was planning to delete the folder before merge

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -977,12 +1010,20 @@ def build_asset_outs(asset_outs: Mapping[str, AssetOut]) -> Mapping[AssetKey, Tu
...
         return deps
 
-    deps_asset_keys = {
-        AssetKey.from_user_string(dep) if isinstance(dep, str) else dep for dep in deps
-    }
+    deps_asset_keys: Set[AssetKey] = set()
+    for dep in deps:
+        if isinstance(dep, AssetsDefinition):
+            # this will error if the AssetsDefinition is a multi_asset
+            deps_asset_keys.add(dep.key)
```
jamiedemaria, line 1025: If a user tries to pass a multi asset AssetsDefinition, they will get an error like this
```
dagster._check.CheckError: Invariant failed. Description: Tried to retrieve asset key from an assets definition with multiple asset keys: ["asset_1"], ["asset_2"]
```

This doesn't give them much direction about what to fix - maybe we should catch and reraise a more useful message?

python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_deps.py
```diff
@@ -0,0 +1,245 @@
...
+        CheckError,
+        match="Tried to retrieve asset key from an assets definition with multiple asset keys",
+    ):
+
+        @asset(deps=[a_multi_asset])
+        def depends_on_both_sub_assets():
+            return None
+
+
+def test_source_asset_deps_via_assets_definition():
```
jamiedemaria, line 206: This test fails - in order for this to work I think we need to update the `deps` type signature to 
```python
deps: Optional[Sequence[Union[AssetsDefinition, SourceAsset, CoercibleToAssetKey]]]
```
and then handle SourceAssets within the `_make_asset_keys` function

what do you think @alangenfeld  @schrockn ?

schrockn: Yup makes sense

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
schrockn: We detect the error right at the entry point at throw immediately. Always good to catch errors as close to the user code as possile.

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -180,11 +188,36 @@ def my_asset(my_upstream_asset: int) -> int:
...
+
+        upstream_asset_deps: Optional[
+            Sequence[Union[CoercibleToAssetKey, AssetsDefinition, SourceAsset]]
+        ] = None
+        if deps is not None:
+            for dep in deps:
+                if isinstance(dep, AssetsDefinition):
+                    if len(dep.keys) > 1:
+                        raise DagsterInvalidDefinitionError(
+                            "Cannot pass a multi asset AssetsDefinition as an argument to deps."
```
jamiedemaria, line 204: TODO - better error message

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -130,8 +136,8 @@ def asset(
             contains letters, numbers, and _) and may not contain python reserved keywords.
         ins (Optional[Mapping[str, AssetIn]]): A dictionary that maps input names to information
             about the input.
-        non_argument_deps (Optional[Union[Set[AssetKey], Set[str]]]): Set of asset keys that are
-            upstream dependencies, but do not pass an input to the asset.
+        deps (Optional[Sequence[Union[AssetsDefinition, SourceAsset, CoercibleToAssetKey]]]):
+             Set of asset keys that are upstream dependencies, but do not pass an input to the asset.
```
alangenfeld, line 140: alright - i didnt understand when i first commented, but I now understand not using `CoercibleToAssetKey` in the API docs since its an internal-ish type alias

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -182,11 +190,36 @@ def my_asset(my_upstream_asset: int) -> int:
...
+        if deps is not None:
+            for dep in deps:
+                if isinstance(dep, AssetsDefinition) and len(dep.keys) > 1:
+                    raise DagsterInvalidDefinitionError(
+                        "Cannot pass a multi_asset AssetsDefinition as an argument to deps."
+                        " Instead, specify dependencies on the assets created by the multi_asset"
+                        f" via AssetKeys or strings. For the multi_asset {dep.node_def.name}, the"
+                        f" available keys are: {dep.keys}."
+                    )
+            upstream_asset_deps = deps
```
jamiedemaria, line 210: type checks on deps? 

Need to make sure a CacheableAssetsDefinitiion isn't passed and if it is that we give a good error message

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -993,12 +1054,23 @@ def build_asset_outs(asset_outs: Mapping[str, AssetOut]) -> Mapping[AssetKey, Tu
     return outs_by_asset_key
 
 
-def _make_asset_keys(deps: Optional[Union[Set[AssetKey], Set[str]]]) -> Optional[Set[AssetKey]]:
-    """Convert all str items to AssetKey in the set."""
+def _make_asset_keys(
```
alangenfeld, line 1057: nit: might be nice to move all the deps coercing & errors out to this / a helper to keep the constructor reasonable

alangenfeld (commented):
im not seeing any major blockers here, lets drop the RFC and get sign-off from folks

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -993,12 +1012,63 @@ def build_asset_outs(asset_outs: Mapping[str, AssetOut]) -> Mapping[AssetKey, Tu
...
+    if non_argument_deps is not None and deps is not None:
+        raise DagsterInvalidDefinitionError(
+            "Cannot specify both deps and non_argument_deps to @asset. Use only deps instead."
+        )
+
+    upstream_asset_deps: Optional[
+        Sequence[Union[CoercibleToAssetKey, AssetsDefinition, SourceAsset]]
+    ] = None
+    if deps is not None:
+        check.list_param(deps, "deps", of_type=(CoercibleToAssetKey, AssetsDefinition, SourceAsset))
```
jamiedemaria, line 1034: adding this check in makes everything fail with ` TypeError: Subscripted generics cannot be used with class and instance checks` errors. I think this is because `CoercibleToAssetKey` contains `List[str]`. I'm not sure what to do about this though

alangenfeld: i think it is understandable that you'll need to write some manual runtime checks instead of  leaning on `check.list_` , should combine reasonably well with the other checks you are doing

jamiedemaria: yup - ended up modifying into what it is right now. The error message is pretty minimal - I could try to break out the `CacheableAssetsDefinition` to a separate error message, but need to figure out what the error message would be first

python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_deps.py
```diff
@@ -0,0 +1,399 @@
...
+def test_non_existent_asset_key():
+    @asset(deps=["not_real"])
+    def my_asset():
+        return None
+
+    res = materialize([my_asset], resources={"io_manager": TestingIOManager()})
+
+    assert res.success
+
+
```
alangenfeld, line 380: how do we handle when `deps` is set in addition to passing the same asset as an input to the function? reasonable error? I could see us potentially relaxing the error since the fn input trumps `deps` pretty clearly

jamiedemaria: good call out - i will add a test

jamiedemaria: you get the error 
```
E dagster._core.errors.DagsterInvalidDefinitionError: @op 'depends_on_upstream_asset' decorated function has parameter 'the_upstream_asset' that is one of the input_defs of type 'Nothing' which should not be included since no data will be passed for it.
```

sryza:
Haven't had a chance to look through this yet, but I think it's worth updating the summary to reflect what the PR is doing.

sryza:
Would it be accurate to say that this PR is deprecating support for having a mix of argument and non-argument deps? If so, what would we say to users who are currently doing this?

jamiedemaria:
> Would it be accurate to say that this PR is deprecating support for having a mix of argument and non-argument deps? If so, what would we say to users who are currently doing this?

Mixing `deps` and function argument deps is still supported

https://github.com/dagster-io/dagster/pull/14886/files#diff-1f3e66d53b99f54128e5ecb94df197c7a943b983fc957e9ff04cdc8ecc7df607R345

python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_deps.py
alangenfeld: thats not great imo - we should at least have a better error and arguably permit it

jamiedemaria: yeah this is the existing behavior though, so maybe we split it out as a separate PR?

python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_deps.py
```diff
@@ -0,0 +1,420 @@
...
+    @asset
+    def asset_1():
+        return None
+
+    @asset(deps=[asset_1])
+    def asset_2():
+        return None
+
+    assert len(asset_2.input_names) == 1
+    assert asset_2.op.ins["asset_1"].dagster_type.kind == DagsterTypeKind.NOTHING
```
alangenfeld, line 40: nit: theres an `is_nothing` method on DagsterType to simplify this

python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_deps.py
```diff
@@ -0,0 +1,420 @@
...
+            self.foo = "bar"
+
+    not_an_asset = NotAnAsset()
+
+    with pytest.raises(
+        DagsterInvalidDefinitionError,
+        match=(
+            "Cannot pass an instance of type <class"
+            " 'dagster_tests.asset_defs_tests.test_asset_deps.test_bad_types.<locals>.NotAnAsset'>"
+            " to deps parameter of @asset"
```
alangenfeld, line 395: nit: reduce what text you are matching against to avoid test thrash on things like python version differences in how local classes are named

python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_deps.py
```diff
@@ -0,0 +1,420 @@
...
+    @asset
+    def the_upstream_asset():
+        return 1
+
+    with pytest.raises(
+        DagsterInvalidDefinitionError,
+        match=(
+            "@op 'depends_on_upstream_asset' decorated function has parameter 'the_upstream_asset'"
+            " that is one of the input_defs of type 'Nothing' which should not be included since no"
+            " data will be passed for it."
```
alangenfeld, line 414: ya we should definitely stack a PR to improve this I think 

maybe leave a comment here indicating this is known sub optimal

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -132,8 +138,8 @@ def asset(
             contains letters, numbers, and _) and may not contain python reserved keywords.
         ins (Optional[Mapping[str, AssetIn]]): A dictionary that maps input names to information
             about the input.
-        non_argument_deps (Optional[Union[Set[AssetKey], Set[str]]]): Set of asset keys that are
-            upstream dependencies, but do not pass an input to the asset.
+        deps (Optional[Sequence[Union[AssetsDefinition, SourceAsset, AssetKey, str]]]):
+            The assets that are upstream dependencies, but do not pass an input value to the asset.
```
sryza, line 142: I think that "input" is a little bit ambiguous. Maybe "but do not correspond to a parameter of the decorated function" would be more clear.

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -88,7 +91,9 @@ def asset(
     name: Optional[str] = None,
     key_prefix: Optional[CoercibleToAssetKeyPrefix] = None,
     ins: Optional[Mapping[str, AssetIn]] = None,
-    non_argument_deps: Optional[Union[Set[AssetKey], Set[str]]] = None,
+    deps: Optional[
+        Sequence[Union[CoercibleToAssetKey, AssetsDefinition, SourceAsset]]
+    ] = None,  # if deps will become the default experience, should it go above ins?
```
sryza, line 96: I think we can omit this comment. The ordering doesn't really matter because of the wildcard.

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -993,12 +1012,73 @@ def build_asset_outs(asset_outs: Mapping[str, AssetOut]) -> Mapping[AssetKey, Tu
...
+        raise DagsterInvalidDefinitionError(
+            "Cannot specify both deps and non_argument_deps to @asset. Use only deps instead."
+        )
+
+    upstream_asset_deps: Optional[
+        Sequence[Union[CoercibleToAssetKey, AssetsDefinition, SourceAsset]]
+    ] = None
+    if deps is not None:
+        for dep in deps:
+            # this gnarly type check is because we can't do isinstance(CoercibleToAssetKey) because CoercibleToAssetKey contains a List[str] subtype
```
sryza, line 1035: Could we use `AssetKey.from_coercible` here?

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -993,12 +1012,73 @@ def build_asset_outs(asset_outs: Mapping[str, AssetOut]) -> Mapping[AssetKey, Tu
...
+        check.set_param(non_argument_deps, "non_argument_deps", of_type=(AssetKey, str))
+        upstream_asset_deps = list(non_argument_deps)
+
+    return upstream_asset_deps
+
+
+def _make_asset_keys(
+    deps: Optional[Sequence[Union[CoercibleToAssetKey, AssetsDefinition, SourceAsset]]]
+) -> Optional[Set[AssetKey]]:
+    """Convert all items to AssetKey in a set. By putting all of the AsestKeys in a set, it will also deduplicate them.
```
sryza, line 1068: Typo

python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_deps.py
```diff
@@ -0,0 +1,420 @@
+# TODO - for ease of initial review, putting all of these tests in a single file. After a review pass,
```
sryza, line 1: Make sure to remove this before merging

sryza (commented):
A few comments but otherwise looks great!

alangenfeld (approved):
think this should be good to go after resolving final set of comments

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
jamiedemaria: where? i can't see it, sigh

alangenfeld: `AsestKeys`

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
jamiedemaria: can you elaborate a bit? use `AssetKey.from_coercible` to do the type check?

python_modules/dagster/dagster_tests/asset_defs_tests/test_asset_deps.py
jamiedemaria: opened an issue too https://github.com/dagster-io/dagster/issues/15167

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
sryza: Yeah or maybe that exact function isn't right. My more general thought here is that, if I understand correctly, we're trying to test whether it will successfully pass `AssetKey.from_coercible`? If that's the case, it seems like this probably isn't the only place that we'll want to be able to check this, and it would be better to put that logic into a method on `AssetKey` than silo it here

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
jamiedemaria: ty!

python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
```diff
@@ -132,8 +136,9 @@ def asset(
             contains letters, numbers, and _) and may not contain python reserved keywords.
         ins (Optional[Mapping[str, AssetIn]]): A dictionary that maps input names to information
             about the input.
-        non_argument_deps (Optional[Union[Set[AssetKey], Set[str]]]): Set of asset keys that are
-            upstream dependencies, but do not pass an input to the asset.
+        deps (Optional[Sequence[Union[AssetsDefinition, SourceAsset, AssetKey, str]]]):
+            The assets that are upstream dependencies, but do not correspond to a parameter of the
+            decorated function
```
jamiedemaria, line 141: add . after sentence

PR #14892, opened 2023-06-21
[auto-materialize] fix source observation issue

OwenKephart:
## Summary & Motivation

Added tests which failed before, and pass now.

Note: This `_new_version_of_source_exists_after_partition` method is kinda gross at the moment, but once the new storage method makes its way in it can get cleaned up a bunch.

### DIFF
--- a/python_modules/dagster/dagster/_utils/caching_instance_queryer.py
+++ b/python_modules/dagster/dagster/_utils/caching_instance_queryer.py
@@ -17,6 +17,7 @@ import pendulum
 from dagster._core.definitions.asset_graph import AssetGraph
 from dagster._core.definitions.data_version import (
     DataVersion,
+    extract_data_provenance_from_entry,
     extract_data_version_from_entry,
 )
 from dagster._core.definitions.events import AssetKey, AssetKeyPartitionKey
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -374,12 +375,21 @@ def new_version_storage_id(
             else None
         )
 
-        next_version_record = self.next_version_record(
+        latest_version_record = self.get_observation_record(
```
OwenKephart, line 378: note: previously, this would return the next asset observation after the cursor with a different version. So if you had (storage_id, version) of:

`[(1, a), (2, b), (3, b), (4, b), (5, c), (6, d), (7, d)]`

Then `f(2)` would previously return `5` (the next record with a different version), and now will return `7` (the latest record id, if it has a different version).

If you added `(8, b)` to the end of the list, then `f(2)` would return `None`, as the version is the same as it was at `2`.

python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/observable_source_asset_scenarios.py
```diff
@@ -124,4 +129,29 @@
             run_request(asset_keys=["asset1", "asset3"]),
         ],
     ),
+    "unchanging_observable": AssetReconciliationScenario(
+        assets=downstream_of_unchanging_observable_source,
+        unevaluated_runs=[
+            run(["source_asset1"], is_observation=True),
+            run(["asset1"]),
+            run(["source_asset1"], is_observation=True),
```
OwenKephart, line 137: previously, this would result in a new run request, as we weren't looking at the specific versions of each observation.

sryza (approved):

PR #14899, opened 2023-06-21
[dagster-airflow] remove < 2.6.2 airflow contraint, ignore dynamic task mapping examples

Ramshackle-Jamathon:
## Summary & Motivation

airflow 2.6.2 added the `example_dynamic_task_mapping_with_no_taskflow_operators` example dag which relies on dynamic task mapping functionality that is not currently supported by dagster-airflow. This PR disables using those examples so that dagster-airflow can still be used with 2.6.2 (although without support of dynamic task mapping)


alangenfeld:
can you update the summary with context / thinking behind the change?

alangenfeld (approved):

PR #14608, opened 2023-06-02
Add server_ecs_tags and run_ecs_tags to EcsContainerContext

gibsondan:
Summary:
This PR has two purposes:
- add the ability to set ECS tags on both runs and code servers
- document the general process of adding additional fields to EcsContainerContext so that they can be configured per code location in OSS and cloud.

I'll add comments in the PR that walk through what the various things do and answer any questions, so that while I am out hopefully the process of adding these is distributed across the team.

PR #14903, opened 2023-06-22
Expose repositoryCredentials field for AWS ECS

armandobelardo:
## Summary & Motivation
This allows passing in an ARN for the repositoryCredentials field, which is required by ECS to access container images stored in private repositories (like Github Container Registry).

The relevant AWS docs are here (and linked in the code): https://docs.aws.amazon.com/AmazonECS/latest/userguide/private-auth.html

Additional detail in Slack: https://dagster.slack.com/archives/C01U954MEER/p1687396168338119

## How I Tested These Changes
I'm not sure how to test this, but happy to give it a go with some pointers!


python_modules/libraries/dagster-aws/dagster_aws/ecs/tasks.py
```diff
@@ -62,6 +64,7 @@ def __new__(
             check.str_param(container_name, "container_name"),
             check.opt_sequence_param(command, "command"),
             check.opt_mapping_param(log_configuration, "log_configuration"),
+            check.opt_str_param(repository_secret, "repository_credentials"),
```
jmsanders, line 67:
```suggestion
            check.opt_str_param(repository_credentials, "repository_credentials"),
```

python_modules/libraries/dagster-aws/dagster_aws/ecs/tasks.py
```diff
@@ -96,6 +99,7 @@ def task_definition_dict(self):
                     ({"secrets": self.secrets} if self.secrets else {}),
                     ({"environment": self.environment} if self.environment else {}),
                     ({"mountPoints": self.mount_points} if self.mount_points else {}),
+                    ({"repositoryCredentials": { "credentialsParameter": repository_credentials }} if self.repository_credentials else {}),
```
jmsanders, line 102:
```suggestion
                    ({"repositoryCredentials": { "credentialsParameter": self.repository_credentials }} if self.repository_credentials else {}),
```

python_modules/libraries/dagster-aws/dagster_aws/ecs/launcher.py
```diff
@@ -257,6 +260,16 @@ def config_type(cls):
                     "The container name to use when launching new tasks. Defaults to 'run'."
                 ),
             ),
+            "repository_credentials": Field(
+                StringSource,
+                is_required=False,
+                default_value=None,
```
jmsanders, line 266:
```suggestion
```

I think the type system is having trouble coercing `None` to a string.

python_modules/libraries/dagster-aws/dagster_aws/ecs/tasks.py
```diff
@@ -42,6 +43,7 @@ def __new__(
         container_name: str,
         command: Optional[Sequence[str]],
         log_configuration: Optional[Mapping[str, Any]],
+        repository_credentials: Optional[str],
```
jmsanders, line 46:
```suggestion
        repository_credentials: Optional[str]=None,
```

Also, you might want to rearrange the order (maybe sticking it at the bottom?) throughout the call stack.

What's happening here is that named tuple's parameter is being type annotated as an optional string, but no default is ever set.

```
>>> from typing import Optional
>>> def foo(bar: Optional[str]):
...     return bar
... 
>>> foo()
TypeError: foo() missing 1 required positional argument: 'bar'
>>> def foo(bar: Optional[str] = None):
...     return bar
... 
>>> foo()
>>> 
```

python_modules/libraries/dagster-aws/dagster_aws/ecs/tasks.py
```diff
@@ -16,6 +16,7 @@ class DagsterEcsTaskDefinitionConfig(
             ("container_name", str),
             ("command", Optional[Sequence[str]]),
             ("log_configuration", Optional[Mapping[str, Any]]),
+            ("repository_credentials", Optional[str]),
```
jmsanders, line 19: These changes will allow you to set the repository credentials at the instance level (for example, in your dagster.yaml). That means every code location will receive the same repository credentials.

If you also thread these through to the "container context," you'll be able to see these at a code location level as well. [This PR is probably our best documentation at the moment for how to make such a change.](https://github.com/dagster-io/dagster/pull/14608)

jmsanders (changes requested):
Thanks! This generally looks good. I commented with a few fixes that I found by running our existing test suite and a suggestion for how to extend this further to allow varying these credentials per code location.

https://docs.dagster.io/community/contributing includes directions for getting your environment set up for testing. But because this is a narrowly scoped change, you might find that setup to be a bit heavy handed. Instead, I think the following should be sufficient to run the relevant tests:

```
pip install 'tox==3.25.0'
python_modules/libraries/dagster-aws/
tox -- -k ecs_tests
```

These tests run against a local mock of ECS. That mock uses [botocore's stubber](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/stubber.html) to ensure that calls to the actual ECS API are valid.

Historically, when we've extended our ECS code in similar ways ([like when we added volumes and mount points](https://github.com/dagster-io/dagster/pull/13409/files#diff-83428dfde55d54583a7e6470d4f59378558e53d2ddce1143f98d02f7f7fd0b67)), we've also extended the existing tests to pass in the new parameter and assert that it gets set as expected.

We don't run any integration tests on this particular code but I'd recommend once you feel like your changes are sufficient, you run it on one of your own deployments and confirm that you can pull from your third-party repository.

armandobelardo:
Changes made and tests pass, thank you for the detailed review and tips, they were extremely extremely helpful!
Going to deploy and make sure it works pulling my credentials for real now

jmsanders:
This is looking pretty good AFAICT. I'm going to make the accompanying changes for dagster cloud and some integration tests we have there should flex this behavior against a hybrid ECS agent.

jmsanders:
Thanks!

PR #14910, opened 2023-06-22
[instigators] refresh request_context during submissions

alangenfeld:
## How I Tested These Changes

added tests for sensors and schedules that failed before these changes

### DIFF
--- a/python_modules/dagster/dagster/_daemon/sensor.py
+++ b/python_modules/dagster/dagster/_daemon/sensor.py
@@ -589,7 +589,6 @@ def _submit_run_request(
     run_request: RunRequest,
     workspace_process_context: IWorkspaceProcessContext,
     external_sensor: ExternalSensor,
-    code_location: CodeLocation,
     existing_runs_by_key,
     logger,
     sensor_debug_crash_flags,
@@ -601,6 +600,10 @@ def _submit_run_request(
         external_sensor.get_target_data(run_request.job_name)
     )
 
+    # reload the code_location on each submission, request_context derived data can become out date
+    # * non-threaded: if number of serial submissions is too many
+    # * threaded: if thread sits pending in pool too long
+    code_location = _get_code_location_for_sensor(workspace_process_context, external_sensor)
     job_subset_selector = JobSubsetSelector(
         location_name=code_location.name,
         repository_name=sensor_origin.external_repository_origin.repository_name,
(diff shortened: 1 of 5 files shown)


alangenfeld:
* master
  * **PR #14910** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14910" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14910?utm_source=stack-comment).

alangenfeld:
context: https://discuss.dagster.io/t/12651848/hello-we-have-a-sensor-that-yields-runrequests-for-a-job-on-#1ef6141b-1037-4e9c-8f41-cb42429870de

python_modules/dagster/dagster/_daemon/sensor.py
```diff
@@ -601,6 +600,10 @@ def _submit_run_request(
         external_sensor.get_target_data(run_request.job_name)
     )
 
+    # reload the code_location on each submission, request_context derived data can become out date
+    # * non-threaded: if number of serial submissions is too many
+    # * threaded: if thread sits pending in pool too long
+    code_location = _get_code_location_for_sensor(workspace_process_context, external_sensor)
```
prha, line 606: What's the cost of recreating the request context?  Will it be overly burdensome for sensors/schedules that emit multiple run requests per ticks?

prha (approved):
I think this makes sense.  The batching by tick probably doesn't make sense if it's going to be distributed across the thread pool anyway.

python_modules/dagster/dagster/_daemon/sensor.py
alangenfeld: i don't think its meaningful cost, some dict copies (shallow?)

alangenfeld:
@prha any ideas on how to test this?

prha (commented):
Probably depends a lot on how deep you want to test...

The very surface level would be to make sure the `log_action` call has a different repo hash.  You'd get this if we have a custom workspace context that generates code location entries with an updated timestamp each time.  A sensor that issues two identical run requests per tick would have two log calls with different code location hashes.

To get two different runs, you'd have to have custom code locations that generate different execution plan snapshot ids.

prha (approved):
reminder typo: loocation

PR #14917, opened 2023-06-22
[auto-materialize][refactor] Use new instance methods

OwenKephart:
## Summary & Motivation

Ok this is a massive set of changes, but please be heartened by the fact that all tests continue to pass, and there is a net removal of ~100LoC.

High level, we previously treated observable source assets and regular assets quite differently, leading to us having to fork down different code paths for each. This also lead to this weird situation where:

- For observable source assets, we factored in data versions, but did not support partitioned source assets
- For regular assets, we supported partitioned assets, but did not factor in data versions

Now, we support partitioned observable sources (factoring in data versions), and the only thing stopping us from factoring in data versions for regular assets is a single conditional (which I left in there until we do an evaluation of the perf impact this might incur).

Previously, we had code strewn about in serval different places to deal with the differences between observable source assets (which require grabbing asset observation events) and regular assets (where we get materializations). Now, both go through the same functions (the big ones being `get_latest_record()` and `get_new_asset_partitions()`). This lets us avoid a bunch of junk in `find_parent_materialized_asset_partitions()`.

This will also be hugely more performant in cases where we're evaluating a large number of partitions of the same asset in a single tick (the classic example would be unpartitioned -> static partitioned). This takes advantage of our new `get_latest_materialization_id_by_partition` functionality.

With a single db call, we can know what the most recent storage id for each of the static partitions is. Similarly, we'll know the latest storage id of the unpartitioned asset. To find all the partitions that have been updated after the unpartitioned asset, we just need to compare the relevant numbers (so total of 2 db calls). Compare this to the previous world in which we'd need a separate db call for each partition to get the latest storage id.


python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -572,82 +410,131 @@ def has_dynamic_partition(self, partitions_def_name: str, partition_key: str) ->
...
+                else None,
+            )
+            return {
+                AssetKeyPartitionKey(asset_key, partition_key): DataVersion(tags[DATA_VERSION_TAG])
+                if tags.get(DATA_VERSION_TAG)
+                else None
+                for partition_key, tags in query_result.items()
+            }
+
+    def get_new_asset_partitions(
```
sryza, line 452: This method name makes me expect that it's returning asset partitions that previously didn't exist, independently of what has been materialized. I think it should have a more specific name, as well as include docstring that explains what precisely it's aiming to return.

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -154,172 +155,93 @@ def _get_latest_materialization_record(
...
-            return not self.materialization_exists(b, after_cursor=latest_a.storage_id)
-        else:
-            latest_b = self.get_latest_materialization_record(b)
-            if latest_b is None:
-                return False
-            return self.materialization_exists(a, after_cursor=latest_b.storage_id)
+        return (self.get_latest_storage_id(asset_partition) or 0) > (after_cursor or 0)
 
-    def get_latest_materialization_record(
+    def get_latest_record(
```
sryza, line 210: I think this needs a more specific name

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -154,172 +155,93 @@ def _get_latest_materialization_record(
...
+        materialized, returns None.
+
+        Args:
+            asset_partition (AssetKeyPartitionKey): The asset partition to query.
+        """
+        return self._get_latest_storage_ids(asset_key=asset_partition.asset_key).get(
+            asset_partition
+        )
+
+    def record_exists(
```
sryza, line 195: I think this needs a more specific name. `asset_partition_has_observation_or_materialization`?

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -154,172 +155,93 @@ def _get_latest_materialization_record(
         )
         return next(iter(records), None)
 
-    def materialization_exists(
+    @cached_method
+    def _get_latest_storage_ids(
```
sryza, line 159: I think this needs a more specific name

python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -747,9 +752,20 @@ def execute_asset_backfill_iteration_inner(
     if request_roots:
         initial_candidates.update(asset_backfill_data.get_target_root_asset_partitions())
 
-        next_latest_storage_id = instance_queryer.get_latest_storage_id(
-            DagsterEventType.ASSET_MATERIALIZATION
+        next_latest_storage_id = None
```
sryza, line 755: I think it's worth making sure the `CachingInstanceQueryer` has some method that does this, to avoid enlarging this already-large function.

sryza (approved):

PR #14920, opened 2023-06-22
[bug] Fix issue with offsetted TimeWindowPartitionMappings

OwenKephart:
## Summary & Motivation

Added test, and it failed. Now it passes

Likely avoided detection as we rarely call `get_downstream_partitions_for_partitions` (basically only in the AutoMaterialize world as far as I'm aware)


clairelin135 (approved):
🙌

PR #14926, opened 2023-06-23
[dagster-dbt] fix failing tests

OwenKephart:
## Summary & Motivation

Followup to the pinning required for dbt-core==1.5.2

Three things:

0. Unpin the dependencies
1. Deprecate the use of the state parameter in DbtManifestAssetSelection (this is experimental, and we don't have high confidence that supporting state like this is a good idea)
2. Update the test_ls* tests to be less brittle. We don't have a consistent structured format to compare against (and different versions would end up producing different numbers of lines), so instead we just do a spot check that all the things we expect to be in the output are, and that nothing we don't expect to be in the output is.


jmsanders (approved):
Love it. Thanks so much!

PR #14929, opened 2023-06-23
Gracefully handle unloadable code locations in asset backfill logic

clairelin135:
Certain users are sometimes seeing a mysterious asset backfill deserialization error:
```
2023-06-23 14:45:18 -0700 - dagster.daemon.BackfillDaemon - ERROR - Backfill failed for dummy_backfill: dagster._core.errors.DagsterDefinitionChangedDeserializationError: Asset asset_a had a PartitionsDefinition at storage-time, but no longer does
```

This error can occur in a variety of situations:
- the code location is no longer loadable (i.e. the location now has a definition-time error, or some gRPC error)
- the asset no longer exists in the workspace
- the asset's partitions def has been removed

This PR clarifies the error messaging to be more informative. Now we:
- raise warnings when some code locations cannot load, like we do in the scheduler and sensor daemons. By raising a warning instead of an exception, we allow backfill with loadable code locations to proceed.
- raise a more informative error when a targeted asset does not exist in the workspace

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph_subset.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph_subset.py
@@ -6,7 +6,10 @@ from dagster._core.definitions.partition import (
     PartitionsDefinition,
     PartitionsSubset,
 )
-from dagster._core.errors import DagsterDefinitionChangedDeserializationError
+from dagster._core.errors import (
+    DagsterAssetBackfillDataLoadError,
+    DagsterDefinitionChangedDeserializationError,
+)
 from dagster._core.instance import DynamicPartitionsStore
 
 from .asset_graph import AssetGraph
@@ -205,6 +208,14 @@ class AssetGraphSubset:
         partitions_subsets_by_asset_key: Dict[AssetKey, PartitionsSubset] = {}
         for key, value in serialized_dict["partitions_subsets_by_asset_key"].items():
             asset_key = AssetKey.from_user_string(key)
+
+            if asset_key not in asset_graph.all_asset_keys:
+                raise DagsterAssetBackfillDataLoadError(
+                    f"Asset {key} does not exist. This error may occur when (1) the asset's code"
+                    " location is unloadable or (2) the asset has been removed. When (1) occurs,"
+                    " review the backfill daemon logs to see which code locations are unloadable."
+                )
+
             partitions_def = asset_graph.get_partitions_def(asset_key)
 
             if partitions_def is None:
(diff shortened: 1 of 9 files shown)


python_modules/dagster/dagster/_core/execution/asset_backfill.py
```diff
@@ -417,8 +419,23 @@ def fetch_cancelable_run_ids_for_asset_backfill(instance: DagsterInstance, backf
...
 
+def _check_workspace_is_loadable(context: IWorkspace, logger: logging.Logger) -> None:
+    location_entries_by_name = {
+        location_entry.origin.location_name: location_entry
+        for location_entry in context.get_workspace_snapshot().values()
+    }
+
+    for location_name, location_entry in location_entries_by_name.items():
+        if location_entry.load_error:
+            logger.warning(
```
clairelin135, line 430: Here we just throw a warning instead of erroring. This is because if just one location in the workspace is unloadable, we don't want to terminate unaffected backfills

python_modules/dagster/dagster/_core/definitions/asset_graph_subset.py
```diff
@@ -205,6 +208,14 @@ def from_storage_dict(
         partitions_subsets_by_asset_key: Dict[AssetKey, PartitionsSubset] = {}
         for key, value in serialized_dict["partitions_subsets_by_asset_key"].items():
             asset_key = AssetKey.from_user_string(key)
+
+            if asset_key not in asset_graph.all_asset_keys:
+                raise DagsterAssetBackfillDataLoadError(
```
sryza, line 213: It's weird to have a backfill-specific error in `AssetGraphSubset`, which can be used in places other than backfills.

sryza (commented):
I think the most helpful way to error would be something like "Definition for asset {key} could not be found. This could be because it's inside a code location that's failing to load:  {unloadable_code_locations}." A lot of users won't have access to their daemon logs (especially in Cloud I think?).

Thoughts? To do that, we might need to catch the error in `AssetBackfillData.from_serialized` and re-throw.

clairelin135:
@sryza That sounds like a good call to me. I've amended this PR to just raise the `DefinitionChangedDeserializationError` from asset graph subset, and to catch this error and raise a more specific error in the asset backfill.

The new error that occurs: 
```python
dagster._core.errors.DagsterAssetBackfillDataLoadError: Asset asset_a existed at storage-time, but no longer does. This could be because it's inside a code location that's failing to load: ['unloadable']
```

PR #14931, opened 2023-06-23
[prototype] return AssetMaterialization

alangenfeld:
explorations for returning AssetMaterializtion directly

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/events.py
+++ b/python_modules/dagster/dagster/_core/definitions/events.py
@@ -20,6 +20,7 @@ import dagster._check as check
 import dagster._seven as seven
 from dagster._annotations import PublicAttr, experimental_param, public
 from dagster._core.definitions.data_version import DataVersion
+from dagster._core.errors import DagsterInvariantViolationError
 from dagster._core.storage.tags import MULTIDIMENSIONAL_PARTITION_PREFIX, SYSTEM_TAG_PREFIX
 from dagster._serdes import whitelist_for_serdes
 from dagster._serdes.serdes import NamedTupleSerializer
@@ -484,7 +485,7 @@ class AssetMaterialization(
 
     Args:
         asset_key (Union[str, List[str], AssetKey]): A key to identify the materialized asset across
-            job runs
+            job runs. Optional in cases when the key can be inferred from the current context.
         description (Optional[str]): A longer human-readable description of the materialized value.
         partition (Optional[str]): The name of the partition
             that was materialized.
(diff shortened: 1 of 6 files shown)


alangenfeld:
* master
  * **PR #14931** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14931" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14931?utm_source=stack-comment).

sryza:
Have you considered a new `MaterializationMetadata` type?

alangenfeld:
>  Have you considered a new MaterializationMetadata type?

Ya thats the other option, in previous discussions no one was hot on the idea of adding a new noun. Whats your take?

sryza:
My initial lean would be towards `MaterializationMetadata` I think. allowing `AssetMaterialization` to be in two different states (with asset key or without) feels weird.

It's also a breaking change, right? For anyone depending on the asset key attribute to be non-None.

And it means that anyone using type-checking needs to guard against None whenever they access `AssetMaterialization.asset_key`.

alangenfeld:
> allowing AssetMaterialization to be in two different states (with asset key or without) feels weird.

Agree, and that is not the case in the current updated state of this prototype PR. We allow omitting from the constructor in cases we can infer it and error otherwise [2].

python_modules/dagster/dagster/_core/definitions/events.py
```diff
@@ -490,18 +491,32 @@ class AssetMaterialization(
...
+                raise DagsterInvariantViolationError(
+                    "Could not infer asset_key, not currently in the context of an execution."
+                )
+            keys = current_ctx.selected_asset_keys
+            if len(keys) != 1:
+                raise DagsterInvariantViolationError(
+                    f"Could not infer asset_key, there are {len(keys)} in the current execution"
+                    " context. Specify the appropriate asset_key."
+                )
+            asset_key = next(iter(keys))
```
alangenfeld, line 519: [2]

alangenfeld:
How do you feel about [2] and with that does that change your stance on `MaterializationMetadata`?

sryza:
> How do you feel about [2] and with that does that change your stance on MaterializationMetadata?

Ah, I missed this. I'm having trouble putting my finger on any concrete negatives, but this kind of pattern gives me some generalized discomfort – I think it's difficult for users to look at the code and guess how it's working.

alangenfeld:
>  this kind of pattern gives me some generalized discomfort – I think it's difficult for users to look at the code and guess how it's working.

Makes sense, this is definitely a "may be lesser of two evils" type of proposal. 

Probably need to flesh this prototype out to include `multi_asset` as I think the full picture including how things work there will influence the final call.

python_modules/dagster/dagster/_core/execution/plan/execute_step.py
```diff
@@ -165,6 +171,15 @@ def _step_output_error_checked_user_event_sequence(
                     f'Emitting implicit Nothing for output "{step_output_def.name}" on {op_label}'
                 )
                 yield Output(output_name=step_output_def.name, value=None)
+
+            if step_context.is_sda_step and step_context.get_observed_user_asset_mat(
```
sryza, line 175: Not sure if this has any consequence with the way that execution currently works, but imagine a world where we could kick off downstream steps as soon as an upstream output is produced, even if the upstream step hasn't finished yet.

In that world, we would ideally yield the `Output` as soon the materialization is reported/yielded/returned, right?

PR #14935, opened 2023-06-24
[NO_SKIP][test-all] Unpin grpcio

gibsondan:
[NO_SKIP][test-all]

Some combination of the grpcio 1.56.0 release and possibly some unintentional changes on our side (could be the dagster-graphql fixture changes so we spin fewer servers up and down on each test run?) makes removing this pin no longer destabilize our BK pipeline.

## How I Tested These Changes
BK (in the past - at least one dagster or dagster-graphql run would consistently hang and time out, now it never happens across multiple runs)

### DIFF
--- a/python_modules/dagster/setup.py
+++ b/python_modules/dagster/setup.py
@@ -27,9 +27,6 @@ def get_version() -> str:
 
 # grpcio 1.44.0 is the min version compatible with both protobuf 3 and 4
 GRPC_VERSION_FLOOR = "1.44.0"
-# Also pinned <1.48.0 until the resolution of https://github.com/grpc/grpc/issues/31885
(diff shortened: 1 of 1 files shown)


jmsanders:
I'm liking the 4 straight builds without a flake 🙌 

I like it - seems worth trying to get into this week's release.

jmsanders (approved):

alangenfeld:
for historical tracking purposes, messages from the pat leave void:
```
- 1.56.0 came out last week: https://pypi.org/project/grpcio/#history

- somebody reported on that issue that it fixed the hang for them: https://github.com/grpc/grpc/issues/31885#issuecomment-1604681629

- so I really quickly rebased an unpin PR and for the first time in forever BK seems to consistently stay green after the unpin: https://github.com/dagster-io/dagster/pull/14935
```

PR #14942, opened 2023-06-26
make --name argument optional when scaffolding example

sryza:
## Summary & Motivation

One less decision for new users to make


yuhan (approved):
sure why not!

PR #14954, opened 2023-06-26
add indirect execution context access

alangenfeld:
adds a means to get the current op/asset execution context via a function call, setup using a `ContextVar`. Exposed as `OpExecutionContext.get()` & `AssetExecutionContext.get()`

## How I Tested These Changes

added test

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/context/compute.py
+++ b/python_modules/dagster/dagster/_core/execution/context/compute.py
@@ -1,4 +1,6 @@
 from abc import ABC, ABCMeta, abstractmethod
+from contextlib import contextmanager
+from contextvars import ContextVar
 from inspect import _empty as EmptyAnnotation
 from typing import (
     AbstractSet,
@@ -48,6 +50,7 @@ from dagster._core.events import DagsterEvent
 from dagster._core.instance import DagsterInstance
 from dagster._core.log_manager import DagsterLogManager
 from dagster._core.storage.dagster_run import DagsterRun
+from dagster._utils.cached_method import cached_method
 from dagster._utils.forked_pdb import ForkedPdb
 from dagster._utils.warnings import (
     deprecation_warning,
@@ -1334,15 +1337,34 @@ class OpExecutionContext(AbstractComputeExecutionContext, metaclass=OpExecutionC
     def set_requires_typed_event_stream(self, *, error_message: Optional[str] = None) -> None:
         self._step_execution_context.set_requires_typed_event_stream(error_message=error_message)
 
+    @staticmethod
+    def get() -> "OpExecutionContext":
+        ctx = _current_asset_execution_context.get()
+        if ctx is None:
+            raise DagsterInvariantViolationError("No current OpExecutionContext in scope.")
+        return ctx.get_op_execution_context()
+
 
 class AssetExecutionContext(OpExecutionContext):
     def __init__(self, step_execution_context: StepExecutionContext):
         super().__init__(step_execution_context=step_execution_context)
 
+    @staticmethod
+    def get() -> "AssetExecutionContext":
+        ctx = _current_asset_execution_context.get()
+        if ctx is None:
+            raise DagsterInvariantViolationError("No current AssetExecutionContext in scope.")
+        return ctx
 
-def build_execution_context(
+    @cached_method
+    def get_op_execution_context(self) -> "OpExecutionContext":
+        return OpExecutionContext(self._step_execution_context)
+
+
+@contextmanager
+def enter_execution_context(
     step_context: StepExecutionContext,
-) -> Union[OpExecutionContext, AssetExecutionContext]:
+) -> Iterator[Union[OpExecutionContext, AssetExecutionContext]]:
     """Get the correct context based on the type of step (op or asset) and the user provided context
     type annotation. Follows these rules.
 
@@ -1393,16 +1415,30 @@ def build_execution_context(
             " OpExecutionContext, or left blank."
         )
 
-    if context_annotation is EmptyAnnotation:
-        # if no type hint has been given, default to:
-        # * AssetExecutionContext for sda steps not in graph-backed assets, and asset_checks
-        # * OpExecutionContext for non sda steps
-        # * OpExecutionContext for ops in graph-backed assets
-        if is_asset_check:
-            return AssetExecutionContext(step_context)
-        if is_op_in_graph_asset or not is_sda_step:
-            return OpExecutionContext(step_context)
-        return AssetExecutionContext(step_context)
-    if context_annotation is AssetExecutionContext:
-        return AssetExecutionContext(step_context)
-    return OpExecutionContext(step_context)
+    # Structured assuming upcoming changes to make AssetExecutionContext contain an OpExecutionContext
+    asset_ctx = AssetExecutionContext(step_context)
+    asset_token = _current_asset_execution_context.set(asset_ctx)
+
+    try:
+        if context_annotation is EmptyAnnotation:
+            # if no type hint has been given, default to:
+            # * AssetExecutionContext for sda steps not in graph-backed assets, and asset_checks
+            # * OpExecutionContext for non sda steps
+            # * OpExecutionContext for ops in graph-backed assets
+            if is_asset_check:
+                yield asset_ctx
+            elif is_op_in_graph_asset or not is_sda_step:
+                yield asset_ctx.get_op_execution_context()
+            else:
+                yield asset_ctx
+        elif context_annotation is AssetExecutionContext:
+            yield asset_ctx
+        else:
+            yield asset_ctx.get_op_execution_context()
+    finally:
+        _current_asset_execution_context.reset(asset_token)
+
+
+_current_asset_execution_context: ContextVar[Optional[AssetExecutionContext]] = ContextVar(
+    "_current_asset_execution_context", default=None
+)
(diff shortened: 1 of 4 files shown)


alangenfeld:
* master
  * **PR #14954** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14954" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14954?utm_source=stack-comment).

schrockn:
If we do this, I think it should be a getter of the class (e.g. AssetExecutionContext.get()) so that the type of the context is clear.

python_modules/dagster/dagster/_core/execution/context/compute.py
```diff
@@ -1300,15 +1305,24 @@ def typed_event_stream_error_message(self) -> Optional[str]:
     def set_requires_typed_event_stream(self, *, error_message: Optional[str] = None) -> None:
         self._step_execution_context.set_requires_typed_event_stream(error_message=error_message)
 
+    @staticmethod
+    def get_current() -> Optional["OpExecutionContext"]:
```
alangenfeld, line 1309: should these raise or return None when there is no current context?

python_modules/dagster/dagster/_core/execution/plan/execute_step.py
```diff
@@ -462,13 +463,14 @@ def core_dagster_event_sequence_for_step(
...
-    with time_execution_scope() as timer_result:
-        user_event_sequence = check.generator(
-            execute_core_compute(
-                step_context,
-                inputs,
-                core_gen,
-            )
+    with time_execution_scope() as timer_result, enter_execution_context(
+        step_context
+    ) as compute_context:
```
alangenfeld, line 468: this is hoisted up here to the top of the iterator call stack since context managers + active generators get goofy and if you raise an exception based on a yielded value in a frame above where the context manager is opened it does not get closed

python_modules/dagster/dagster/_core/execution/context/compute.py
schrockn: IMO we should raise

schrockn:
Interested @slopp 's feedback on naming, etc

python_modules/dagster/dagster_tests/core_tests/execution_tests/test_context.py
```diff
@@ -405,3 +405,18 @@ def the_op(context: int):
         @asset
         def the_asset(context: int):
             pass
+
+
+def test_get_context():
```
benpankow, line 410: might be worth testing thread behavior? I guess ContextVar should handle it fine

alangenfeld: ya feel pretty good about thread & coroutine concurrency 
https://docs.python.org/3/library/contextvars.html

schrockn (dismissed):
Req'ing change based on the in-memory instantiations issue (comment inline)

slopp:
I'm not sure I understand the use case enough to have an opinion on the name 

I think the only thing that is potentially confusing as a user is whether this method has to be called... eg 

```python

@asset
def my_asset(context: AssetExecutionContext): 
     context.get_current() # what? is this necessary; why or why not?

```

Does re-usijg the standard Python `get` make this more or less clear? 🤷 Context injection is just fairly dagster magical.

Otherwise fine with `get_current` or `get_context` or even the fairly verbose `get_current_context`

python_modules/dagster/dagster/_core/execution/context/compute.py
```diff
@@ -1359,16 +1373,37 @@ def build_execution_context(
...
-        if is_asset_check:
-            return AssetExecutionContext(step_context)
-        if is_op_in_graph_asset or not is_sda_step:
-            return OpExecutionContext(step_context)
-        return AssetExecutionContext(step_context)
-    if context_annotation is AssetExecutionContext:
-        return AssetExecutionContext(step_context)
-    return OpExecutionContext(step_context)
+    asset_ctx = AssetExecutionContext(step_context)
+    op_ctx = OpExecutionContext(step_context)
```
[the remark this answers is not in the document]
alangenfeld: The thing that led me this current approach was ensuring that the passed down `context` was `==` to the result of `_ExecutionContext.get()`. It will requires some shenaigans to maintain that equality if we do the proposed shifting.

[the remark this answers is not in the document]
jamiedemaria: I'm not sure I'm quite following the issue/proposed solution here. can you add some elaboration?

schrockn:
> I'm not sure I understand the use case enough to have an opinion on the name
> 
> I think the only thing that is potentially confusing as a user is whether this method has to be called... eg
> 
> ```python
> @asset
> def my_asset(context: AssetExecutionContext): 
>      context.get_current() # what? is this necessary; why or why not?
> ```
> 
> Does re-using the standard Python `get` make this more or less clear? 🤷 Context injection is just fairly dagster magical.
> 
> Otherwise fine with `get_current` or `get_context` or even the fairly verbose `get_current_context`

@slopp the use case here is avoid the need to pass contexts around entirely and instead provide a global accessor.

You would instead be able to write code like the following:

 ```python
@asset
def my_asset(): 
      context = AssetExecutionContext.get_current()
```

There are tradeoffs to this approach, but we have certainly gotten this feature request in the past.

File: python_modules/dagster/dagster/_core/execution/context/compute.py [hunk 5]
```diff
@@ -1359,16 +1384,30 @@ def build_execution_context(
...
+            elif is_op_in_graph_asset or not is_sda_step:
+                yield asset_ctx.get_op_execution_context()
+            else:
+                yield asset_ctx
+        elif context_annotation is AssetExecutionContext:
+            yield asset_ctx
+        else:
+            yield asset_ctx.get_op_execution_context()
+    finally:
+        _current_asset_execution_context.reset(asset_token)
```
schrockn, line 1408: 👍🏻 thanks much more comfortable with this

python_modules/dagster/dagster/_core/execution/context/compute.py
```diff
@@ -1300,15 +1306,34 @@ def typed_event_stream_error_message(self) -> Optional[str]:
     def set_requires_typed_event_stream(self, *, error_message: Optional[str] = None) -> None:
         self._step_execution_context.set_requires_typed_event_stream(error_message=error_message)
 
+    @staticmethod
+    def get() -> "OpExecutionContext":
```
alangenfeld, line 1310: i updated these to just `get` now that they raise + slopps feedback

benpankow:
looks good on my end, will leave to others w/ feedback to approve

slopp:
> > I'm not sure I understand the use case enough to have an opinion on the name
> > I think the only thing that is potentially confusing as a user is whether this method has to be called... eg
> > ```python
> > @asset
> > def my_asset(context: AssetExecutionContext): 
> >      context.get_current() # what? is this necessary; why or why not?
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Does re-using the standard Python `get` make this more or less clear? 🤷 Context injection is just fairly dagster magical.
> > Otherwise fine with `get_current` or `get_context` or even the fairly verbose `get_current_context`
> 
> @slopp the use case here is avoid the need to pass contexts around entirely and instead provide a global accessor.
> 
> You would instead be able to write code like the following:
> 
> ```python
> @asset
> def my_asset(): 
>      context = AssetExecutionContext.get_current()
> ```
> 
> There are tradeoffs to this approach, but we have certainly gotten this feature request in the past.

Thanks for clarifying. Yes this new capability makes sense to me. I've seen a few users get very tripped up by the use of `context` (eg trying to put it as the second argument or trying to use a different name). I think this addition is beneficial. Would we go so far as to say this is the best practice?

Understanding the use case, I still don't have super strong opinions on the names.

schrockn:
Also would like @PedramNavid and @tacastillo's thoughts on this. It solves a problem that some users have but at the cost of yet-another-way of doing things.

alangenfeld:
From what I understand @benpankow has an internal use case for this functionality (being able to swap in a insights compatible resource without changing the API). 

In the current form of the PR the new static `get` is not yet marked `@public`. 

If we feel like we are close to resolution I am fine getting the buy in to mark this `@public`, but otherwise we may want to separate out the end user facing exposure of this functionality to prevent blocking progress of work.

tacastillo:
Oh wow this is _beautiful_

1. ZocDoc

PedramNavid:
Generally supportive! I assume same experience with regards to type-ahead/completions? So long as we maintain support for existing context magic for a while I think it's fine. 

How would this work with functions called from an asset? You would still need to pass context right?

Or would this work?

```python
def my_wrapper_func(x): 
    if AssetExecutionContext.some_method():
        do_something_with(x)

@asset
def my_asset():
    x = 123
    my_wrapper_func(x)
```

schrockn:
@PedramNavid @tacastillo The tradeoff I'd like you to consider is 1) should we allow both methods and accept that there are "two ways of doing things" 2) should we drive heavy towards to global accessor and incur switching costs and 3) what is the price we pay for having code depend on global state.

@tacastillo in terms of "my vote would be to appoint this as the best practice and to move away from magic arguments unless needed (sounds like there's a use case for keeping them?)" consider the cases where we have resource initialization or sensors or anywhere else where there is a subtly different context. Should we have N global accessors and error when users call the wrong one? Or consolidate into a single `ExecutionContext` that covers all cases?

That all being said as @alangenfeld said this solves an immediate problem for Ben and does not make this public, so we can go ahead and land this. Seems like whether or not to make this public is a sep discussion.

schrockn (approved):

PR #14956, opened 2023-06-26
Update pandas calls to support pandas 2.x

### LINKED ISSUE #13339: Update integration libs for compatibility with pandas 2.0
### What's the use case?

Pandas 2.0 was released on 4/3/2023, the following libs exampled have been pinned at `pandas<2` and need to be updated for compatibility:

- assets-pandas-type-metadata
- assets-pandas-pyspark
- dagster-duckdb-pandas
- dagster-duckdb-pyspark
- dagster-gcp-pandas

### Ideas of implementation

_No response_

### Additional information

_No response_

smackesey:
## Summary & Motivation

Updates calls to pandas to support 2.x. Also maintains compatibility with later pandas 1.x releases. List of changes:

- Calls to `iteritems` now use `items` instead.
- Calls to `append` now use `concat` instead.

Both `iteritems` and `append` had been deprecated in Pandas 1.x so changing this call should maintain compatibility with at least more recent Pandas 1.x releases.

This PR also necessarily bumps our pyright pins to account for pandas 2. This introduces a number of new type errors that are due to stub errors. These errors are type-ignored in this PR, they should go away (and the unnecessary type-ignores flagged) when the stubs are fixed.

Resolves #13339 

## How I Tested These Changes

Existing test suite. Manually tested changed codepaths using venv with pandas 1.x.


smackesey:
* master
  * **PR #14956** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14956" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14956?utm_source=stack-comment).

python_modules/libraries/dagster-duckdb-pyspark/dagster_duckdb_pyspark/duckdb_pyspark_type_handler.py
```diff
@@ -211,3 +213,12 @@ def type_handlers() -> Sequence[DbTypeHandler]:
     @staticmethod
     def default_load_type() -> Optional[Type]:
         return pyspark.sql.DataFrame
+
+
+# This is temporarily needed until pyspark supports converting timestamps properly to pandas 2.x.
+# See: https://stackoverflow.com/questions/76072664/convert-pyspark-dataframe-to-pandas-dataframe-fails-on-timestamp-column
+def _pyspark_df_to_pandas_df(df: pyspark.sql.DataFrame) -> pd.DataFrame:
```
smackesey, line 220: @sryza Pinging you for confirmation that this is OK and non-destructive, since I don't know pyspark at all.

Basically the call `df.toPandas()` fails with pandas 2 if the spark df contains a timestamp column, because of either a bug or just failure of pyspark to fully support pandas 2-- this function is a workaround that changes any pyspark df timestamp columns before the call to `df.toPandas` so that they are properly autoconverted.

jamiedemaria: duckdb also has support for direct ingestion from polars and pyarrow now (they didn't when this lib was written) so we could convert to a different table type if that makes more sense https://duckdb.org/docs/api/python/overview#dataframes

smackesey: I DMed sandy about this and he didn't know the answer. Then I investigated a bit more and I'm pretty sure this is not a clean conversion, it is ending up in duckdb as a string column instead of a timestamp.

> duckdb also has support for direct ingestion from polars and pyarrow now

Do you know if we can easily convert to one of these intermediate formats from pyspark (equivalent of `.toPandas()`)?

smackesey: @jamiedemaria thanks for your help, I got it to work by converting to arrow instead of pandas

PR #14957, opened 2023-06-26
[auto-materialize] Fix off-by-one error

OwenKephart:
## Summary & Motivation

This method returns the storage id of the latest record that has a different data version from what we saw last tick (if such a record exists).

Previously, we would compare the latest available record to the latest record that happened _before_ after_cursor (which will be the last storage id from last tick). However, this misses the correct record in cases where the last storage id from last tick is the record that we saw last tick.

This can happen if no materializations or observations occur in between successive observations of the same version.


sryza (approved):
is there a test we can write for this?

OwenKephart:
@sryza yep, just added

Issue #14384, opened 2023-05-22
Partition performance issue due to size of StaticPartitionsDefinition


yuuuxt:
### Dagster version

1.3.5

### What's the issue?

When the size of `StaticPartitionsDefinition` is large, the time cost also increases and it is too much. 

### What did you expect to happen?

I'm expecting the time consumed for materialization does not increase when `StaticPartitionsDefinition` increases.

### How to reproduce?

Please check the [minimal example repo](https://github.com/yuuuxt/dagster_performance_check), which includes 2 assets of `return [2,3,4]`. On my laptop:

- when the asset is `3 * time-dimension`, materialization of one partition takes ~0.3s:
![image](https://github.com/dagster-io/dagster/assets/3153534/4f80308a-5364-4780-804d-aef74963b8ab)
![image](https://github.com/dagster-io/dagster/assets/3153534/6a015f98-f6e3-40c3-aea2-de63d7ca14be)

- when the asset is `5000 * time-dimension`, materialization of one partition takes ~ 1s:
![image](https://github.com/dagster-io/dagster/assets/3153534/4893f2bf-28c7-4b90-9c69-ba85a80862ae)
![image](https://github.com/dagster-io/dagster/assets/3153534/a2531f6a-c0d5-49c3-b272-4388d7be4def)

Besides, in my another test (in docker compose), it's like for a 1.3kb pickle materialization: `3 * time-dimension` takes 0.1s, while `~5000 * time-dimension` takes 3s.

### Deployment type

Local

### Deployment details

the result of example repo is from a win 10 laptop, and the "another test" is on a linux machine using docker compose.

### Additional information

https://dagster.slack.com/archives/C01U954MEER/p1684429605641439

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

OwenKephart:
Hi @yuuuxt ! Thanks for the report, and the reproduction (very helpful!).

This is happening due to a quirk in the IOManager logic which is making Dagster generate all partition keys for that MultiPartitionsDefinition (so creating N*M string objects in memory) even though execution is just targeting a single one of those N*M partitions.

We're working on a series of performances improvements around this area, and so this should be fixed within the next couple of weeks.

yuuuxt:
Checked that it hasn't been fixed in 1.3.11. Any plan for the foreseeable future? Thx.

OwenKephart:
[...]

PR #14965, opened 2023-06-26
Improve perf for get_partition_keys_in_range

OwenKephart:
## Summary & Motivation

Resolves: https://github.com/dagster-io/dagster/issues/14384

Most of the time that we call this, there is only a single partition key in range. In these cases, calling get_partition_keys() is quite wasteful, so we avoid it.

## How I Tested These Changes

Confirmed that get_partition_keys() is no longer called in the materialization flow.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/partition.py
+++ b/python_modules/dagster/dagster/_core/definitions/partition.py
@@ -176,11 +176,9 @@ class PartitionsDefinition(ABC, Generic[T_str]):
         partition_key_range: PartitionKeyRange,
         dynamic_partitions_store: Optional[DynamicPartitionsStore] = None,
     ) -> Sequence[T_str]:
-        partition_keys = self.get_partition_keys(dynamic_partitions_store=dynamic_partitions_store)
-
         keys_exist = {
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster/_core/definitions/partition.py
```diff
@@ -189,6 +191,11 @@ def get_partition_keys_in_range(
                 {list(key for key in keys_exist if keys_exist[key] is False)}"""
             )
 
+        # in the simple case, simply return the single key in the range
+        if partition_key_range.start == partition_key_range.end:
+            return [cast(T_str, partition_key_range.start)]
+
+        partition_keys = self.get_partition_keys(dynamic_partitions_store=dynamic_partitions_store)
```
clairelin135, line 198: can we add a comment here detailing why we defer calling `get_partition_keys`? Could see someone refactoring and then undo-ing the changes here

clairelin135 (approved):
Thanks for adding this!

python_modules/dagster/dagster/_core/definitions/partition.py
OwenKephart: good call, added !

PR #14972, opened 2023-06-27
Add `Title`

salazarm:
## Summary & Motivation

Adds a new heading type "Title". Ideally we'd rename our headings to be like Heading1, Heading2, etc, but it's something we can do down the line with a regex rename when we have spare cycles.

## How I Tested These Changes

Figma spec for the new header text:

Screenshot 2023-06-27 at 11 03 28 AM

### DIFF
--- a/js_modules/dagit/packages/ui/src/components/Text.tsx
+++ b/js_modules/dagit/packages/ui/src/components/Text.tsx
@@ -7,7 +7,7 @@ interface TextProps {
   color?: string;
 }
 
-export const BigHeading = styled.span<TextProps>`
+export const Title = styled.span<TextProps>`
   ${({color}) => (color ? `color: ${color};` : null)}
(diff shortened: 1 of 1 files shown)


js_modules/dagit/packages/ui/src/components/Text.tsx
```diff
@@ -7,6 +7,14 @@ interface TextProps {
   color?: string;
 }
 
+export const Title = styled.span<TextProps>`
+  ${({color}) => (color ? `color: ${color};` : null)}
+  font-size: 24px;
+  font-weight: 600;
+  line-height: 29px;
```
hellendag, line 14: @braunjj the odd pixel value here is a bit unusual -- should this maybe be 30px or 32px?

hellendag (approved):

PR #14985, opened 2023-06-27
by->for in DbtManifest methods

sryza:
## Summary & Motivation

We generally use `by` when we're returning a dictionary and `for` when we're returning a single element from a dictionary.


rexledesma (approved):
Need to also change the reference [in the experimental examples](https://sourcegraph.com/github.com/dagster-io/dagster/-/blob/examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/jobs/yield_materializations.py?L14:36).

PR #14993, opened 2023-06-28
Suppress flaky fixture cleanup

jmsanders:
We sometimes see flakes in these tests presumably because of a race condition on setup:

https://buildkite.com/dagster/dagster/builds/60833

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql_tests/graphql/graphql_context_test_suite.py
+++ b/python_modules/dagster-graphql/dagster_graphql_tests/graphql/graphql_context_test_suite.py
@@ -1,3 +1,4 @@
+import contextlib
 import sys
 import tempfile
 from abc import ABC, abstractmethod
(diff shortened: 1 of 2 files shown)


python_modules/dagster-graphql/dagster_graphql_tests/graphql/graphql_context_test_suite.py
```diff
@@ -794,8 +795,9 @@ def yield_class_scoped_graphql_context(self, request):
         @pytest.fixture(name="graphql_context")
         def yield_graphql_context(self, class_scoped_graphql_context):
             instance = class_scoped_graphql_context.instance
-            instance.wipe()
-            instance.wipe_all_schedules()
+            with contextlib.suppress(FileNotFoundError):
+                instance.wipe()
```
jmsanders, line 799: I'm a little surprised to see the wiping happening _before_ we yield rather than _after_ and considered moving it. But figured I'd start by seeing if this fixes our issue.

Given that this is happening on setup and not cleanup, I'm a little afraid we might be in a weird state where we yield a sqlite db that doesn't exist.

jmsanders: It felt a bit odd to me to suppress this during "normal" wipe operations instead of only in tests. But I guess realistically, wipe is really only there for testing purposes anyway.

alangenfeld: I think it makes sense in the context of that unlink call to continue if its already gone since the purpose is clean it up and remove it

alangenfeld (approved):

PR #14992, opened 2023-06-28
docs(dbt): add example to schedule a materialization of a dbt selection

rexledesma:
## Summary & Motivation
Add a simple example of using `ScheduleDefinition` and `define_asset_job` to schedule a selection of dbt assets.

We are planning on adding an API in `DbtManifest` that streamlines this experience. This PR serves as a baseline for the experience that what we are optimizing against.

## How I Tested These Changes
local

PR #14982, opened 2023-06-27
refactor(dbt): move `DbtManifestAssetSelection` to `resources_v2`

rexledesma:
## Summary & Motivation
`DbtManifestAssetSelection` is experimental, so just group it with the other related experimental APIs. We can figure out the optimal folder structure and separate them out into their own files closer to the 1.4 release date.

Also `DbtManifestAssetSelection` was being exported from the top level API. Don't do that since this is experimental.

## How I Tested These Changes
existing pytest

PR #14980, opened 2023-06-27
feat(dbt): add method to create asset selections from dbt selections

rexledesma:
## Summary & Motivation
`AssetSelection` is a Dagster specific concept that generalizes dbt selection syntax. Here, we allow the user to create an `AssetSelection` using the native dbt selection syntax.

`DbtManifest` is our source of truth for dbt operations, as it wraps the dbt manifest, the artifact representation of the dbt project. Now, users can invoke a method on this object to create an `AssetSelection`.

We already have `DbtManifestAssetSelection` representing an `AssetSelection` in dbt space. We continue to use this object for this purpose, but we make it simpler.

We keep the following arguments:

| Argument | Notes |
|--------|--------|
| `manifest` | renaming from `manifest_json`, and this is now required to be a `DbtManifest` |
| `select` | corresponds to dbt `--select` criteria |
| `exclude` | corresponds to dbt `--exclude` criteria |

And remove other arguments:

| Argument | Notes |
|--------|--------|
| `manifest_json_path` | subsumed by `DbtManifest.read(...)` |
| `resource_types` | subsumed under `select` |
| `node_info_to_asset_key` | subsumed by overriding `DbtManifest.node_info_to_asset_key(...)` |
| `state_path` | deprecated | 

## How I Tested These Changes
pytest

PR #14997, opened 2023-06-28
feat(dbt): add method to build schedules from dbt selections

rexledesma:
## Summary & Motivation

We introduce a `.build_schedule(...)` method on `DbtManifest` to create a schedule given a selection of dbt assets. This way, users are shielded from directly interacting with `define_asset_job` and `ScheduleDefinition`.

The API that we introduce makes the construction of `ScheduleDefinition` for simple cases. But, we should not seek to recreate a parallel API that contains all the `ScheduleDefinition` and `define_asset_job` arguments.

For example, if the user wants to use partition config or yield `RunRequest` and `SkipReason` in their schedule, then they must use the underlying Dagster framework API's (`ScheduleDefinition`, `@schedule`, `define_asset_job`) to accomplish their use cases.

## How I Tested These Changes
pytest

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
@@ -6,7 +6,18 @@ import sys
 import time
 from dataclasses import dataclass
 from pathlib import Path
-from typing import AbstractSet, Any, Dict, Iterator, List, Mapping, Optional, Sequence, Union
+from typing import (
+    TYPE_CHECKING,
+    AbstractSet,
+    Any,
+    Dict,
+    Iterator,
+    List,
+    Mapping,
+    Optional,
+    Sequence,
+    Union,
+)
 
 import dateutil.parser
 from dagster import (
@@ -16,7 +27,9 @@ from dagster import (
     ConfigurableResource,
     OpExecutionContext,
     Output,
+    ScheduleDefinition,
     _check as check,
+    define_asset_job,
     get_dagster_logger,
 )
 from dagster._annotations import experimental
@@ -35,6 +48,9 @@ from ..asset_utils import (
     output_name_fn,
 )
 from ..utils import ASSET_RESOURCE_TYPES, select_unique_ids_from_manifest
+
+if TYPE_CHECKING:
+    from dagster import RunConfig
 
 logger = get_dagster_logger()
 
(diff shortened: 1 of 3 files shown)


rexledesma:
* master
  * **PR #14980** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14980" target="_blank">Graphite</a> 
    * **PR #14982** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14982" target="_blank">Graphite</a> 
      * **PR #14992** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14992" target="_blank">Graphite</a> 
        * **PR #14997** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14997" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/14997?utm_source=stack-comment).

sryza (commented):
Nice. Only question here is if we should call it `build_scheduled_job`. The advantage would be that it might make it more clear what to look for in the UI when they want to find what they've built.

rexledesma:
> Nice. Only question here is if we should call it `build_scheduled_job`. The advantage would be that it might make it more clear what to look for in the UI when they want to find what they've built.

This makes sense from the UI perspective, but from the perspective of organizing the Dagster definitions in code, it might be confusing.

When I assemble my `Definitions`, `build_scheduled_job` would suggest putting the definition under `jobs` imo.

```python

defs = Definitions(jobs=[manifest.build_scheduled_job(...)]
```

Alternatives?
- `build_schedule_for_job`
- `build_schedule_for_selection`

sryza:
Ah, good point. In light of that, I think you're right that `build_schedule` is best.

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -258,6 +274,49 @@ def build_asset_selection(
...
+        self,
+        job_name: str,
+        cron_schedule: str,
+        dbt_select: str = "fqn:*",
+        dbt_exclude: Optional[str] = None,
+        tags: Optional[Mapping[str, str]] = None,
+        config: Optional["RunConfig"] = None,
+        execution_timezone: Optional[str] = None,
+    ) -> ScheduleDefinition:
+        """Build a schedule to materialize a specified set of dbt resources from a dbt selection string.
```
sryza, line 287: The docstring should mention that this builds a job and wraps it in a schedule.

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -258,6 +274,49 @@ def build_asset_selection(
...
+
+        See https://docs.getdbt.com/reference/node-selection/syntax#how-does-selection-work.
+
+        Args:
+            job_name (str): The name of the job to materialize the dbt resources.
+            cron_schedule (str): The cron schedule to define the schedule.
+            dbt_select (str): A dbt selection string to specify a set of dbt resources.
```
sryza, line 294: This should mention the default selection value.

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -258,6 +274,49 @@ def build_asset_selection(
...
+            dbt_exclude (Optional[str]): A dbt selection string to exclude a set of dbt resources.
+            tags (Optional[Mapping[str, str]]): A dictionary of tags (string key-value pairs) to attach
+                to the scheduled runs.
+            config (Optional[RunConfig]): The config that parameterizes the execution of this schedule.
```
sryza, line 298: Not blocking, but it could be helpful to add some docstring here to help users understand what this "config" is and how it relates to the kind of dbt config they'd want to set.

sryza (approved):

PR #14998, opened 2023-06-28
[ui] Show button to view asset step logs in failed / current run banners

bengotow:
## Summary & Motivation

This PR adds a button to view step logs to both the "Run is currently materializing this asset" and "Run failed to materialize this asset" banners. 

This ended up changing quite a bit of code because we are using the asset.opNames to pre-filter the step logs modal, and the modal must not be conditionally rendered based on whether the banners are visible (to avoid the modal closing as soon as the run finishes.) I improved a few loading states in the asset sidebar - the event loading spinner is shown inside the event section, and the event loading spinner doesn't appear while it's background-refreshing to show a newer event (see last screenshot)

The modal is no longer "locked" to just logs for the selected asset - instead, it scrolls to the selected asset's logs and highlights them in yellow, but shows everything. This is important for "failed to materialize" since the failure may have been upstream.

image

image

image

image

image

## How I Tested These Changes

I tested these buttons manually, and verified that if you sit in each modal for a long time / as the run finishes, the modal does not disappear.

### DIFF
--- a/js_modules/dagit/packages/core/src/runs/StepLogsDialog.tsx
+++ b/js_modules/dagit/packages/core/src/runs/StepLogsDialog.tsx
@@ -1,8 +1,9 @@
-import {Box, Button, Colors, Dialog, DialogFooter, Icon, Mono} from '@dagster-io/ui';
+import {Box, Button, Colors, Dialog, DialogFooter, Icon, Mono, Spinner} from '@dagster-io/ui';
 import React, {useState} from 'react';
 import {Link} from 'react-router-dom';
 import styled from 'styled-components/macro';
 
+import {DagsterEventType} from '../graphql/types';
 import {useSupportsCapturedLogs} from '../instance/useSupportsCapturedLogs';
 
 import {CapturedOrExternalLogPanel} from './CapturedLogPanel';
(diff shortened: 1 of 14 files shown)


js_modules/dagit/packages/core/src/runs/StepLogsDialog.tsx
```diff
@@ -15,21 +16,50 @@ import {IRunMetadataDict, RunMetadataProvider} from './RunMetadataProvider';
...
+  const [showingLogs, setShowingLogs] = React.useState<{runId: string; stepKeys: string[]} | null>(
+    null,
+  );
+
+  // Note: This hook captures the runId + stepKeys in state when you click the button and then passes
+  // those values to the modal. This ensures that the modal is "stable" while it's open, even if
+  // the runId passed to the hook changes or becomes undefined. (eg: "Currently materializing" banner)
+
+  return {
+    dialog: (
```
salazarm, line 29: this is a good approach I tend to use often. Having a hook that returns components gives the caller more flexibility in terms of where to put things 👍🏾

salazarm (approved):

Issue #14987, opened 2023-06-28
DynamicPartitions via pyspark_step_launcher - Nonexistent partition keys (DagsterInvalidInvocationError)


weberdavid:
### Dagster version

1.3.10

### What's the issue?

We discussed this issue in detail in Slack: https://dagster.slack.com/archives/C01U954MEER/p1687185900524019

Here is a summary:

1. We use file partitions based on DynamicPartitionsDefinition
2. Asset1: Is executed locally, uses the partitions via the `partitions_def` in the `@asset` decorator
3. Asset2: Executed in Databricks using the `pyspark_step_launcher`, also using `partitions_def` in the `@asset` decorator; this asset depends on asset1, by having it as an input parameter in the function.

This is the exact error message:
```
dagster._core.errors.DagsterInvalidInvocationError: Partition range <our_partition_key> to <our_partition_key> is
                not a valid range. Nonexistent partition keys:
                [<our_partition_key>]
```

Apparently, dagster on Databricks is not aware of the partition that we are passing to asset2.

### What did you expect to happen?

The error message is unexpected, as I expected Dagster on Databricks to have access to the DynamicPartition that is used throughout a job.

Here is a potential solution on how to fix it, suggested by @OwenKephart (thanks!):

1. StepRunRef contains all of the information that is sent to the external instance, but doesnt include dynamic partitions
2. Updating [run_step_from_ref](https://sourcegraph.com/github.com/dagster-io/dagster/-/blob/python_modules/dagster/dagster/_core/execution/plan/external_step.py?L237) would be a lightweight solution, to add a partition to the external Databricks instance.
3. First there should be a check, whether we are dealing with a DynamicPartitionsDef

I am currently testing @OwenKephart suggestion in our setup and will have a PR available soon.

### How to reproduce?

Here is a minimal code example that reproduces the issue (thanks @loerinczy):
```
from pathlib import Path
from dagster import (
    asset, 
    sensor, 
    SensorResult, 
    RunRequest,
    DynamicPartitionsDefinition,
    Definitions,
    AssetSelection,
    define_asset_job
)
from dagster_databricks import databricks_pyspark_step_launcher

partition = DynamicPartitionsDefinition(name='test_partition')

@asset(
    partitions_def=partition,
    required_resource_keys={'pyspark_step_launcher'}
)
def test_asset():
    return 1

test_job = define_asset_job(name='test_job', selection=AssetSelection.keys('test_asset'))

@sensor(
    name='test_sensor',
[...]

PR #15000, opened 2023-06-28
Databricks dynamic partitions fix

weberdavid:
Proposed fix by @OwenKephart, to fix this issue: #14987 

Additionally, I added a wrapper method in python_modules/dagster/dagster/_core/execution/context/system.py so that we do not access the private method directly. If you have a better naming for it, or any other proposed ideas feel free to add / change them.

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/plan/external_step.py
+++ b/python_modules/dagster/dagster/_core/execution/plan/external_step.py
@@ -237,6 +237,10 @@ def run_step_from_ref(
     check.inst_param(instance, "instance", DagsterInstance)
     step_context = step_run_ref_to_step_context(step_run_ref, instance)
 
+    # Note: This is a patch that enables using DynamicPartitionsDefinitions with step launchers in the specific case where:
+    # 1. The external step operates on a single dynamic partition.
+    # 2. No dynamic partitions are added in this external step.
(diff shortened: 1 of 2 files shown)


smackesey:
cc @OwenKephart

python_modules/dagster/dagster/_core/execution/plan/external_step.py
```diff
@@ -236,6 +237,21 @@ def run_step_from_ref(
     check.inst_param(instance, "instance", DagsterInstance)
     step_context = step_run_ref_to_step_context(step_run_ref, instance)
 
+    if step_context.has_partition_key:
```
OwenKephart, line 240: The only bit of feedback I have is to leave a note here along the lines of:

```
Note: This is a patch that enables using DynamicPartitionsDefinitions with step launchers in the specific case where...

1. The external step operates on a single dynamic partition.
2. No dynamic partitions are added in this external step.

A more complete solution would require including all dynamic partitions on the StepRunRef object.
```

Happy to merge this in once added!

OwenKephart (approved):

python_modules/dagster/dagster/_core/execution/plan/external_step.py
weberdavid: @OwenKephart thanks for the feedback - its done! 👍

PR #15008, opened 2023-06-28
[refactor] Clean up should_reconcile logic

OwenKephart:
## Summary & Motivation

This was becoming quite a mess, with different bits of logic placed somewhat at random throughout these functions.

For the most part, this is a no-op in terms of functionality (just clearing the way to make adjustments easier in the future), but the one small change is that now when an asset is missing and one of its parents are updated, it gets both of those conditions associated with it, rather than one.


sryza (approved):
<3

PR #15011, opened 2023-06-28
[Runs] Pass timezone to timerange filter

salazarm:
## Summary & Motivation
As titled

## How I Tested These Changes

Loaded the Runs page and checked the timestamp when selecting "Today" that its correctly the start of "today" in my timezone (eg: the time is 00:00:00):
Screenshot 2023-06-28 at 1 44 20 PM

### DIFF
--- a/js_modules/dagit/packages/core/src/ui/Filters/useTimeRangeFilter.tsx
+++ b/js_modules/dagit/packages/core/src/ui/Filters/useTimeRangeFilter.tsx
@@ -67,10 +67,12 @@ type Args = {
 export function useTimeRangeFilter({
   name,
   icon,
-  timezone,
+  timezone: _timezone,
   initialState,
   onStateChanged,
 }: Args): TimeRangeFilter {
+  const timezone =
+    _timezone === 'Automatic' ? Intl.DateTimeFormat().resolvedOptions().timeZone : _timezone;
   const [state, setState] = React.useState<TimeRangeState>(initialState || [null, null]);
   React.useEffect(() => {
     onStateChanged?.(state);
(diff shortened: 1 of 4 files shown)


js_modules/dagit/packages/core/src/ui/Filters/useTimeRangeFilter.tsx
```diff
@@ -67,10 +67,12 @@ type Args = {
 export function useTimeRangeFilter({
   name,
   icon,
-  timezone,
+  timezone: _timezone,
   initialState,
   onStateChanged,
 }: Args): TimeRangeFilter {
+  const timezone =
+    _timezone === 'Automatic' ? Intl.DateTimeFormat().resolvedOptions().timeZone : _timezone;
```
bengotow, line 75: If it's all the same I'd vote to put the useContext(TimeContext) in here to eliminate the possibility of another callsite doing this incorrectly in the future. I'm not sure there's a use case where the timezone should /not/ be the one from the time context?

bengotow (approved):

js_modules/dagit/packages/core/src/ui/Filters/useTimeRangeFilter.tsx
hellendag: If it's `Automatic`, it should use `browserTimezone()`

salazarm: Ah in the Audit log we explicitly use UTC, I think its intentional but I'm in favor of being consistent here so lets change it and see if anyone complains

PR #15016, opened 2023-06-28
[Tabs component] Set background color on icons

salazarm:
## Summary & Motivation

Our Icons use a combination of background color + mask so this should be background color.
## How I Tested These Changes
storybook:
Screenshot 2023-06-28 at 3 19 10 PM


hellendag (approved):
Yep, checked to see where this line originated, and it's from when we were briefly using the Material Icons font, at which time `color` was appropriate.

PR #15017, opened 2023-06-28
:zap: add async support to UPathIOManager for concurrent partitions loads

danielgafni:
## Summary & Motivation

This PR brings  `async` support to the `UPathIOManager`. 
This should massively speedup loading multiple partitions (with partition mappings) from remote filesystems like S3. 
The `UPathIOManager` automatically detects if the `load_from_path` method is `async`, and uses `asyncio.gather` to execute the concurrent loading.

@sryza 

## How I Tested These Changes

Added tests for `async def load_from_path`

--- 

I actually only touched `upath_io_manager.py` , `test_upath_io_manager.py` and `setup.py`, but running `pre-commit` caused some unrelated files to change too. I can `git stash` them if needed. 

Seems like `pre-commit` has to be enforced by CI to prevent situations like this. 

---

I can add more docs if this feature is welcome


tacastillo:
Thanks for raising this. Let me send this over to the team to figure out who can best help you land this.

danielgafni:
We could also have parallel load without asyncio (just with threading).
This could instantly speedup loads for the existing user code without any changes on the user side.

python_modules/dagster/dagster/_core/storage/upath_io_manager.py
```diff
@@ -191,33 +201,82 @@ def _load_multiple_inputs(self, context: InputContext) -> Dict[str, Any]:
...
+        paths = self._get_paths_for_partitions(context)  # paths for normal partitions
+        backcompat_paths = self._get_multipartition_backcompat_paths(
+            context
+        )  # paths for multipartitions
 
         context.log.debug(f"Loading {len(paths)} partitions...")
 
-        for partition_key, path in paths.items():
-            context.log.debug(f"Loading partition from {path} using {self.__class__.__name__}")
+        def load_partition(partition_key: Union[str, MultiPartitionKey]):
```
sryza, line 211: With all the additions here, `_load_multiple_inputs` is starting to get a little large and unwieldy. I think moving `load_partition` to be its own method on `UPathIOManager` would make this easier to follow.

python_modules/dagster/dagster/_core/storage/upath_io_manager.py
```diff
@@ -191,33 +201,82 @@ def _load_multiple_inputs(self, context: InputContext) -> Dict[str, Any]:
...
+            path = paths[partition_key]
+            backcompat_path = backcompat_paths.get(partition_key)
+
+            # 1. We will first try to load the partition from the normal path.
```
sryza, line 215: This comment would make sense as the docstring for this function

python_modules/dagster/dagster/_core/storage/upath_io_manager.py
```diff
@@ -191,33 +201,82 @@ def _load_multiple_inputs(self, context: InputContext) -> Dict[str, Any]:
...
+            # 2. If it is not found, we will try to load it from the backcompat path.
+            # 3. If allow_missing_partitions is True, we will skip the partition if it was not found in any of the paths.
+            # otherwise, we will raise an error.
+
+            missing_partition_message = f"Couldn't load partition {partition_key} and skipped it "
+            "because the input metadata includes allow_missing_partitions=True"
+
             try:
+                context.log.debug(f"Loading partition from {path}...")
```
sryza, line 224: Do we not want to include `self.__class__.__name__` in the debug message anymore?

python_modules/dagster/dagster/_core/storage/upath_io_manager.py
```diff
@@ -191,33 +201,82 @@ def _load_multiple_inputs(self, context: InputContext) -> Dict[str, Any]:
...
                 obj = self.load_from_path(context=context, path=path)
-                objs[partition_key] = obj
+                return obj
             except FileNotFoundError as e:
-                backcompat_path = backcompat_paths.get(partition_key)
                 if backcompat_path is not None:
-                    obj = self.load_from_path(context=context, path=backcompat_path)
+                    try:
+                        context.log.debug(
```
sryza, line 230: Asymptotically, the cases where the backcompat path will have data will tend towards 0, so the added confusion ("what's a backcompat path?") will eventually exceed the value.

It might make sense to only log the backcompat path in the case that we successfully load from it.

sryza (commented):
Thanks for submitting this @danielgafni 

When you say "running pre-commit", what do you mean by this exactly? Is this a command that we supply? Would you be up for submitting these changes in a separate PR and tagging @smackesey ?

A few thoughts on the implementation:
- We should log an experimental warning for this behavior for now (we typically do this whenever we add new behavior).
- Is this able to include some tests for failure cases?
- Adding the morefs dependency makes me nervous, because it seems like a pretty fledgling package. If this is just for local filesystems, and most of the benefit comes for non-local filesystems, I'd lean towards keeping this out.

danielgafni:
1. Yes, you have a .pre-commit-config.yaml file in the repo :) running 'pre-commit install' installs git hooks which run black, ruff and the others (described in the file) before every commit. The problem is - because these checks are not enforced in CI, some contributors might not be aware of it, and commit code which haven't been run through the checks locally. This causes situation like this one (I had the hooks installed and they fixed some code unrelated to my PR). 
2. Sure, I'll add a warning 
3. I can add tests for failures
4. I'm not sure what to do about morefs either. Note how I didn't add it to the actual Dagster dependencies l, but only to the ones required to run the tests. It would only be needed if the user decided to use the helper method I provided (get_async_filesystem). In fact, this is not mandatory, the user may chose to use something different to implement async load_from_path. Ok the other hand, it has an (almost) unified API with the official fsspec async filesystems, which is nice.

sryza:
> I'm not sure what to do about morefs either. Note how I didn't add it to the actual Dagster dependencies l, but only to the ones required to run the tests. It would only be needed if the user decided to use the helper method I provided (get_async_filesystem). In fact, this is not mandatory, the user may chose to use something different to implement async load_from_path. Ok the other hand, it has an (almost) unified API with the official fsspec async filesystems, which is nice. Yes, the benefit would mostly be for remote filesystems. However this is a little tricky - for example, the user might be mounting an S3 bucket, which would be recognized as local filesystem by the current code but would actually benefit from async. I would like to write async code even for the local filesystem for this reason.

Thinking about this a little more, I think we should avoid exposing this method and just leave it up to the user to get an async filesystem.  This functionality has more to do with user-space code and the specifics of the UPath system than Dagster.  We can keep `morefs` in the test dependencies if we need to use that implementation in the test to test things out.

smackesey:
>Yes, you have a .pre-commit-config.yaml file in the repo :) running 'pre-commit install' installs git hooks which run black, ruff and the others (described in the file) before every commit. The problem is - because these checks are not enforced in CI, some contributors might not be aware of it, and commit code which haven't been run through the checks locally. This causes situation like this one (I had the hooks installed and they fixed some code unrelated to my PR).

Hmm-- so the extra files changed in this PR are a result of running `black` on the `docs` folder. Our pre-commit hooks are generally enforced in CI, though `docs` is not checked for with `black` (this is an oversight, I think it's because there just didn't used to be any python code in there).

However, the black pre-commit hook would not have changed these files unless they were already part of the diff-- so the pre-commit is not responsible here. I'm guessing what happened is that you ran `black` from the root at some point @danielgafni -- pretty sure that just recursively picks up all python files, so it would've picked up those in `docs`. The reason these weren't already black-formatted is because we (and our CI) use `make black`, which excludes the docs folder.

Regardless I put up a PR that adds `docs` to `make black` and formats those files, so if you rebase this after that merges the extra changes should go away: https://github.com/dagster-io/dagster/pull/15063

danielgafni:
@sryza Done with the refactoring. I also made a separate method `get_path_for_partition`. It's useful to be able to override it, I had to do it in `dagster-polars` to [implement](https://github.com/danielgafni/dagster-polars/blob/a81d2e8b7c1dd16acd580c3ad05fac7edd0f84c7/dagster_polars/io_managers/delta.py#L60C6-L60C6) native `deltalake` partitioning. Sometimes we don't want Dagster to have separate paths for different partitions (same for writing partitioned parquet datasets with e.g. pyspark).

On a side note: how could we be *reading* individual partitions in this case? We need to somehow set `partition_filters` for pyarrow, but I'm not sure how to do it properly, as it involves both custom user metadata and the partition key. Which can be a `MultiPartitionKey` :) I decided to ignore this for now and leave partition filtering out of scope of the IOManager. As I remember, the `DBIOManager` handles this with a SQL `partition_expr`, it might be possible to do the same with `polars` (as it supports SQL). 

2. So what should we do with 

```python
@staticmethod
    def get_async_filesystem(path: "Path") -> AsyncFileSystem:
        """A helper method for the `UPathIOManager` end-user, is useful inside an async `load_from_path`.
        The returned `fsspec` FileSystem will have async IO methods.
        https://filesystem-spec.readthedocs.io/en/latest/async.html.
        """
        if isinstance(path, Path) and not isinstance(path, UPath):
            try:
                from morefs.asyn_local import AsyncLocalFileSystem

                return AsyncLocalFileSystem()
            except ImportError as e:
                raise RuntimeError(
                    "Install 'morefs[asynclocal]' to use `get_async_filesystem` with a local"
                    " filesystem"
                ) from e
        elif isinstance(path, UPath):
            kwargs = path._kwargs.copy()  # noqa
            kwargs["asynchronous"] = True
            return path._default_accessor(path._url, **kwargs)._fs  # noqa
        else:
            raise DagsterInvariantViolationError(
                f"Path type {type(path)} is not supported by the UPathIOManager"
            )
```
?

 Are you sure we don't need to expose it and just move it to tests completely? It's quite some boilerplate to write. 
Perhaps, it could be referenced somewhere in async examples?

danielgafni:
Also, are there any plans for loading data from multiple Ins concurrently?

smackesey:
@danielgafni Can you try rebasing on master? That should eliminate the unrelated `docs` changes.

danielgafni:
ok cool

sryza:
> Are you sure we don't need to expose it and just move it to tests completely? It's quite some boilerplate to write.

It will be less boilerplate for any individual implementation, right? E.g. if someone was writing an S3 async UPathIOManager, would they need any of that code?  And if someone was writing a local async UPathIOManager, would they need the code in the `elif`?

sryza:
> On a side note: how should we implement reading individual partitions in this case? We need to somehow set partition_filters for pyarrow, but I'm not sure how to do it properly, as it involves both custom user metadata and the partition key. Which can be a MultiPartitionKey :) I decided to ignore this for now and leave partition filtering out of scope of the IOManager

Each partition is in its own file, so partition filtering shouldn't be needed, right? `partition_expr` is only needed with the DB IO manager because all the partitions are lumped together in a single table.

danielgafni:
> Each partition is in its own file, so partition filtering shouldn't be needed, right? partition_expr is only needed with the DB IO manager because all the partitions are lumped together in a single table.

All partitions [end up](https://github.com/danielgafni/dagster-polars/blob/master/dagster_polars/io_managers/delta.py#L67) in the same directory. We don't separate them anymore and let the the `DeltaLake` writer handle separating the partitions. Just like you would have with Spark: 

```python
df.partitionBy("date").write.mode("overwrite").parquet("/partitioned.parquet")
```

So same situation as with the DBIOManager.

> It will be less boilerplate for any individual implementation, right? E.g. if someone was writing an S3 async UPathIOManager, would they need any of that code? And if someone was writing a local async UPathIOManager, would they need the code in the elif?

The point of the `UPathIOManager` is to be filesystem agnostic. There shouldn't be an `S3` implementation or a `GCP` implementation (but of course it's possible). The reason we are dealing with `morefs` here is because `fsspec` doesn't (yet?) implement the `AsyncLocalFilesystem`. However, all the normal cloud filesystems have an async version, so this is rather an exception. 

But I understand the hesitation to introduce this small library in the top-level code. Lets keep this code as an implementation example. Where should I put it?

sryza:
Got it. Regarding `AsyncLocalFilesystem`, I think the best place to put it would be in the API docs for `load_from_path`.

Here are the issues that buildkite is reporting:
- There are some type checker that you should be able to uncover with `make pyright`
- dagster_tests/general_tests/test_import.py::test_import_perf is failing. It looks like we specifically try to make sure that `upath` doesn't get imported when you import Dagster, because it can be slow.
- I'm seeing this error in the `with_great_expectations` example:
```
[2023-07-03T18:49:21Z] INFO: pip is looking at multiple versions of dagster[test] to determine which version is compatible with other requirements. This could take a while.
[2023-07-03T18:49:21Z] ERROR: Ignored the following versions that require a different python version: 0.0.1 Requires-Python >=3.8; 0.0.2 Requires-Python >=3.8; 0.0.3 Requires-Python >=3.8; 0.1.0 Requires-Python >=3.8; 0.1.1 Requires-Python >=3.8; 0.1.2 Requires-Python >=3.8; 0.10.6 Requires-Python >=2.7, <3.6; 0.10.7 Requires-Python >=2.7, <3.6; 1.0.0 Requires-Python >=3.8
[2023-07-03T18:49:21Z] ERROR: Could not find a version that satisfies the requirement morefs[asynclocal] (from dagster[test]) (from versions: none)
[2023-07-03T18:49:21Z] ERROR: No matching distribution found for morefs[asynclocal]
[2023-07-03T18:49:21Z] ERROR: invocation failed (exit code 1)
[2023-07-03T18:49:21Z] ERROR: could not install deps [-e../../python_modules/dagster[test], -e../../python_modules/dagit, -e../../python_modules/dagster-graphql, -e../../python_modules/libraries/dagster-pandas, -e../../python_modules/libraries/dagster-ge, -e.]; v = InvocationError("/workdir/examples/with_great_expectations/.tox/py37/bin/python -m pip install '-e../../python_modules/dagster[test]' -e../../python_modules/dagit -e../../python_modules/dagster-graphql -e../../python_modules/libraries/dagster-pandas -e../../python_modules/libraries/dagster-ge -e.", 1)
```

sryza:
The code changes look good! I'm broadly seeing build errors like the following. Do you know how to fix? If not, I can investigate how to help.

```
[2023-07-05T22:44:47Z]
--
  | [2023-07-05T22:44:47Z]   ignore::dagster.ExperimentalWarning
  | [2023-07-05T22:44:47Z]
  | [2023-07-05T22:44:47Z] This error occurred:
  | [2023-07-05T22:44:47Z]
  | [2023-07-05T22:44:47Z] Traceback (most recent call last):
  | [2023-07-05T22:44:47Z]   File "/workdir/python_modules/dagster/.tox/py310-api_tests/lib/python3.10/site-packages/_pytest/config/__init__.py", line 1642, in parse_warning_filter
  | [2023-07-05T22:44:47Z]     category: Type[Warning] = _resolve_warning_category(category_)
  | [2023-07-05T22:44:47Z]   File "/workdir/python_modules/dagster/.tox/py310-api_tests/lib/python3.10/site-packages/_pytest/config/__init__.py", line 1680, in _resolve_warning_category
  | [2023-07-05T22:44:47Z]     m = __import__(module, None, None, [klass])
  | [2023-07-05T22:44:47Z]   File "/workdir/python_modules/dagster/dagster/__init__.py", line 477, in <module>
  | [2023-07-05T22:44:47Z]     from dagster._core.storage.fs_io_manager import (
  | [2023-07-05T22:44:47Z]   File "/workdir/python_modules/dagster/dagster/_core/storage/fs_io_manager.py", line 21, in <module>
  | [2023-07-05T22:44:47Z]     from dagster._core.storage.upath_io_manager import UPathIOManager
  | [2023-07-05T22:44:47Z]   File "/workdir/python_modules/dagster/dagster/_core/storage/upath_io_manager.py", line 28, in <module>
  | [2023-07-05T22:44:47Z]     class UPathIOManager(MemoizableIOManager):
  | [2023-07-05T22:44:47Z]   File "/workdir/python_modules/dagster/dagster/_core/storage/upath_io_manager.py", line 162, in UPathIOManager
  | [2023-07-05T22:44:47Z]     ) -> UPath:
  | [2023-07-05T22:44:47Z] NameError: name 'UPath' is not defined
```

Also, are you able to run `make black` and `make ruff`?

smackesey:
The issue with the build errors is that `UPath` is currently in a `TYPE_CHECKING` block, move it out of there and the errors will resolve.

danielgafni:
Yeah I had to move it there because of the increase import time for dagster. 

It's actually only used in the new properties I've added, I'll place individual imports inside them.

danielgafni:
@sryza hey, I fixed pyright errors and added some tests for failures with asyncio

sryza:
Thanks for the update @danielgafni. Here are the build issues I'm seeing:
- Are you able to run `make ruff`?
- I'm seeing errors like this when running tests for the `with_airflow` example, the `dagster-airflow` library, and the `with_great_expectations` example:

```
[2023-07-11T21:11:57Z] ERROR: Ignored the following versions that require a different python version: 0.0.1 Requires-Python >=3.8; 0.0.2 Requires-Python >=3.8; 0.0.3 Requires-Python >=3.8; 0.1.0 Requires-Python >=3.8; 0.1.1 Requires-Python >=3.8; 0.1.2 Requires-Python >=3.8; 0.10.6 Requires-Python >=2.7, <3.6; 0.10.7 Requires-Python >=2.7, <3.6; 1.0.0 Requires-Python >=3.8
[2023-07-11T21:11:57Z] ERROR: Could not find a version that satisfies the requirement morefs[asynclocal] (from dagster[test]) (from versions: none)
[2023-07-11T21:11:57Z] ERROR: No matching distribution found for morefs[asynclocal]
[2023-07-11T21:11:58Z] ERROR: invocation failed (exit code 1)
[2023-07-11T21:11:58Z] ERROR: could not install deps [-e../../python_modules/dagster[test], -e../../python_modules/dagit, -e../../python_modules/dagster-graphql, -e../../python_modules/libraries/dagster-airflow[test], -e.]; v = InvocationError("/workdir/examples/with_airflow/.tox/py37/bin/python -m pip install '-e../../python_modules/dagster[test]' -e../../python_modules/dagit -e../../python_modules/dagster-graphql '-e../../python_modules/libraries/dagster-airflow[test]' -e.", 1)
```

- I'm seeing this error:

```
[2023-07-11T21:18:27Z]
[2023-07-11T21:18:27Z] =================================== FAILURES ===================================
[2023-07-11T21:18:27Z] ____________ test_upath_io_manager_async_fail_on_missing_partitions ____________
[2023-07-11T21:18:27Z]
[2023-07-11T21:18:27Z] tmp_path = PosixPath('/tmp/pytest-of-root/pytest-0/test_upath_io_manager_async_fa0')
[2023-07-11T21:18:27Z] daily = TimeWindowPartitionsDefinition(start=1640995200.0, timezone='UTC', fmt='%Y-%m-%d', end_offset=0, cron_schedule='0 0 * * *')
[2023-07-11T21:18:27Z] start = datetime.datetime(2022, 1, 1, 0, 0)
[2023-07-11T21:18:27Z]
[2023-07-11T21:18:27Z]     def test_upath_io_manager_async_fail_on_missing_partitions(
[2023-07-11T21:18:27Z]         tmp_path: Path,
[2023-07-11T21:18:27Z]         daily: DailyPartitionsDefinition,
[2023-07-11T21:18:27Z]         start: datetime,
[2023-07-11T21:18:27Z]     ):
[2023-07-11T21:18:27Z]         manager = AsyncJSONIOManager(base_dir=str(tmp_path))
[2023-07-11T21:18:27Z]
[2023-07-11T21:18:27Z]         @asset(partitions_def=daily, io_manager_def=manager)
[2023-07-11T21:18:27Z]         def upstream_asset(context: AssetExecutionContext) -> str:
[2023-07-11T21:18:27Z]             return context.partition_key
[2023-07-11T21:18:27Z]
[2023-07-11T21:18:27Z]         @asset(
[2023-07-11T21:18:27Z]             partitions_def=daily,
[2023-07-11T21:18:27Z]             io_manager_def=manager,
[2023-07-11T21:18:27Z]             ins={
[2023-07-11T21:18:27Z]                 "upstream_asset": AssetIn(partition_mapping=TimeWindowPartitionMapping(start_offset=-1))
[2023-07-11T21:18:27Z]             },
[2023-07-11T21:18:27Z]         )
[2023-07-11T21:18:27Z]         def downstream_asset(upstream_asset: Dict[str, str]):
[2023-07-11T21:18:27Z]             return upstream_asset
[2023-07-11T21:18:27Z]
[2023-07-11T21:18:27Z]         materialize(
[2023-07-11T21:18:27Z]             [upstream_asset],
[2023-07-11T21:18:27Z]             partition_key=start.strftime(daily.fmt),
[2023-07-11T21:18:27Z]         )
[2023-07-11T21:18:27Z]
[2023-07-11T21:18:27Z]         with pytest.raises(FileNotFoundError):
[2023-07-11T21:18:27Z] >           materialize(
[2023-07-11T21:18:27Z]                 [upstream_asset.to_source_asset(), downstream_asset],
[2023-07-11T21:18:27Z]                 partition_key=(start + timedelta(days=4)).strftime(daily.fmt),
[2023-07-11T21:18:27Z]             )
[2023-07-11T21:18:27Z]
[2023-07-11T21:18:27Z] dagster_tests/storage_tests/test_upath_io_manager.py:565:
[2023-07-11T21:18:27Z] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[2023-07-11T21:18:27Z] dagster/_core/definitions/materialize.py:103: in materialize
[2023-07-11T21:18:27Z]     return check.not_none(
[2023-07-11T21:18:27Z] dagster/_core/definitions/job_definition.py:679: in execute_in_process
[2023-07-11T21:18:27Z]     return core_execute_in_process(
[2023-07-11T21:18:27Z] dagster/_core/execution/execute_in_process.py:74: in core_execute_in_process
[2023-07-11T21:18:27Z]     event_list = list(execute_run_iterable)
[2023-07-11T21:18:27Z] dagster/_core/execution/api.py:871: in __iter__
[2023-07-11T21:18:27Z]     yield from self.iterator(
[2023-07-11T21:18:27Z] dagster/_core/execution/api.py:754: in job_execution_iterator
[2023-07-11T21:18:27Z]     for event in job_context.executor.execute(job_context, execution_plan):
[2023-07-11T21:18:27Z] dagster/_core/executor/in_process.py:55: in execute
[2023-07-11T21:18:27Z]     yield from iter(
[2023-07-11T21:18:27Z] dagster/_core/execution/api.py:871: in __iter__
[2023-07-11T21:18:27Z]     yield from self.iterator(
[2023-07-11T21:18:27Z] dagster/_core/executor/in_process.py:26: in inprocess_execution_iterator
[2023-07-11T21:18:27Z]     yield from inner_plan_execution_iterator(
[2023-07-11T21:18:27Z] dagster/_core/execution/plan/execute_plan.py:119: in inner_plan_execution_iterator
[2023-07-11T21:18:27Z]     for step_event in check.generator(
[2023-07-11T21:18:27Z] dagster/_core/execution/plan/execute_plan.py:371: in dagster_event_sequence_for_step
[2023-07-11T21:18:27Z]     raise error
[2023-07-11T21:18:27Z] dagster/_core/execution/plan/execute_plan.py:273: in dagster_event_sequence_for_step
[2023-07-11T21:18:27Z]     for step_event in check.generator(step_events):
[2023-07-11T21:18:27Z] dagster/_core/execution/plan/execute_step.py:343: in core_dagster_event_sequence_for_step
[2023-07-11T21:18:27Z]     for evt in check.generator(
[2023-07-11T21:18:27Z] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[2023-07-11T21:18:27Z]
[2023-07-11T21:18:27Z] step_context = <dagster._core.execution.context.system.StepExecutionContext object at 0x7fcd525504c0>
[2023-07-11T21:18:27Z] input_name = 'upstream_asset'
[2023-07-11T21:18:27Z] input_value = {'2022-01-04': FileNotFoundError(2, 'No such file or directory'), '2022-01-05': FileNotFoundError(2, 'No such file or directory')}
[2023-07-11T21:18:27Z]
[2023-07-11T21:18:27Z]     def _type_checked_event_sequence_for_input(
[2023-07-11T21:18:27Z]         step_context: StepExecutionContext,
[2023-07-11T21:18:27Z]         input_name: str,
[2023-07-11T21:18:27Z]         input_value: Any,
[2023-07-11T21:18:27Z]     ) -> Iterator[DagsterEvent]:
[2023-07-11T21:18:27Z]         check.inst_param(step_context, "step_context", StepExecutionContext)
[2023-07-11T21:18:27Z]         check.str_param(input_name, "input_name")
[2023-07-11T21:18:27Z]
[2023-07-11T21:18:27Z]         step_input = step_context.step.step_input_named(input_name)
[2023-07-11T21:18:27Z]         input_def = step_context.op_def.input_def_named(step_input.name)
[2023-07-11T21:18:27Z]
[2023-07-11T21:18:27Z]         check.invariant(
[2023-07-11T21:18:27Z]             input_def.name == input_name,
[2023-07-11T21:18:27Z]             f"InputDefinition name does not match, expected {input_name} got {input_def.name}",
[2023-07-11T21:18:27Z]         )
[2023-07-11T21:18:27Z]
[2023-07-11T21:18:27Z]         dagster_type = input_def.dagster_type
[2023-07-11T21:18:27Z]         type_check_context = step_context.for_type(dagster_type)
[2023-07-11T21:18:27Z]         input_type = type(input_value)
[2023-07-11T21:18:27Z]         op_label = step_context.describe_op()
[2023-07-11T21:18:27Z]
[2023-07-11T21:18:27Z]         with user_code_error_boundary(
[2023-07-11T21:18:27Z]             DagsterTypeCheckError,
[2023-07-11T21:18:27Z]             lambda: f'Error occurred while type-checking input "{input_name}" of {op_label}, with Python type {input_type} and Dagster type {dagster_type.display_name}',
[2023-07-11T21:18:27Z]             log_manager=type_check_context.log,
[2023-07-11T21:18:27Z]         ):
[2023-07-11T21:18:27Z]             type_check = do_type_check(type_check_context, dagster_type, input_value)
[2023-07-11T21:18:27Z]
[2023-07-11T21:18:27Z]         yield _create_step_input_event(
[2023-07-11T21:18:27Z]             step_context, input_name, type_check=type_check, success=type_check.success
[2023-07-11T21:18:27Z]         )
[2023-07-11T21:18:27Z]
[2023-07-11T21:18:27Z]         if not type_check.success:
[2023-07-11T21:18:27Z] >           raise DagsterTypeCheckDidNotPass(
[2023-07-11T21:18:27Z]                 description=(
[2023-07-11T21:18:27Z]                     f'Type check failed for step input "{input_name}" - '
[2023-07-11T21:18:27Z]                     f'expected type "{dagster_type.display_name}". '
[2023-07-11T21:18:27Z]                     f"Description: {type_check.description}"
[2023-07-11T21:18:27Z]                 ),
[2023-07-11T21:18:27Z]                 metadata=type_check.metadata,
[2023-07-11T21:18:27Z]                 dagster_type=dagster_type,
[2023-07-11T21:18:27Z]             )
[2023-07-11T21:18:27Z] E           dagster._core.errors.DagsterTypeCheckDidNotPass: Type check failed for step input "upstream_asset" - expected type "Dict[String,String]". Description: Value "[Errno 2] No such file or directory: '/tmp/pytest-of-root/pytest-0/test_upath_io_manager_async_fa0/upstream_asset/2022-01-04'" of python type "FileNotFoundError" must be a string.
```

sryza:
It looks like the only remaining issues are the ones that come up with the with_airflow example, the dagster-airflow library, and the with_great_expectations example.

danielgafni:
I'm not sure how to proceed with these. 
What is the Python version in these environments?

sryza:
I did some investigation here. I believe the issue is that morefs only supports Python 3.8+, but we test on Python 3.7 and above.

How critical is including morefs in the tests? There is likely a way to have those tests only run on Python 3.8+, but I don't know off the top of my head. If you grep the codebase, you might be able to find examples where we've done this.

smackesey:
Use this:

```
@pytest.mark.skipif(sys.version_info < (3, 8), reason="requires python3.8")
```

danielgafni:
should be working now

sryza (approved):
This looks good. Merging!

smackesey:
@danielgafni `universal-pathlib` 0.1.0 was released and it breaks some tests in our suite-- see specifically:

```
python_modules/dagster/dagster_tests/storage_tests/test_upath_io_manager.py::test_upath_io_manager_async_allow_missing_partitions
```

I played around a little bit trying to fix it, but ended up having to get into the internals of `universal-pathlib` due to the private API access (`_kwargs`) added in this PR. I was unable to figure it out-- do you have any insight into what changed in universal/pathlib, and/or do you know if there is a solution that does not involve private API access to UPath? I found this which seems relevant: https://github.com/fsspec/universal_pathlib/issues/129

ap--:
Hello everyone,

I just released a new bugfix release for universal_pathlib, that restores the regression in the private API when accessing `._kwargs` and `._url`. I will open another issue to plan how dagster can move to a new public API for its use case without any other interruptions.

Cheers,
Andreas

smackesey:
Hey @ap--, thanks for your efforts.

Is the new release you're referring to `0.1.1`? Because I am seeing the same test failures on this version, and it looks like from the same cause, which is that `<UPath_instance>._kwargs` is throwing an error.

I believe the problem must be specific to `UPath` subclasses, because the class in the test is `upath.implementations.local.PosixUPath` but this works: `python -c 'import upath; print(upath.UPath(".")._kwargs)`.

Any suggestions?

ap--:
I can't reproduce this locally. Can you point me to the failing test?

```python
>>> import upath
>>> upath.__version__
'0.1.1'
>>> upath.UPath(".")._kwargs
{}
>>> from upath.implementations.local import PosixUPath
>>> PosixUPath(".")._kwargs
{}
```

Note, that for local paths (without 'file:' prefix, let's call them non-URI paths), `UPath(pth)` returns a `PosixUPath` or `WindowsUPath` instance.

smackesey:
@sryza The private API is also accessed in `UPathIOManager` itself: https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_core/storage/upath_io_manager.py#L73-L88

Sounds like the issue might lie with the lack of a public option in the library though? But sounds like @ap-- is going to help us out:

>I will open another issue to plan how dagster can move to a new public API for its use case without any other interruptions.

danielgafni:
I don't think so. As I understand, @ap-- has plans for exposing storage_options with a public attribute, buts this is not implemented yet.

ap--:
Hmm, I see. It's a bit more involved than I thought.

So if I understand correctly, what you want is:

"given a UPath instance, return an corresponding fsspec filesystem, but return it with asynchronous=True"

Currently, the public API to do this would be:
```python
if isinstance(pth, UPath):
    so = pth.fs.storage_options.copy()
    cls = type(pth.fs)
    so["asynchronous"] = True
    fs = cls(**so)

```
But, this will instantiate the filesystem class with the old storage options, just to retrieve them, AND it only works if the filesystem class already inherits from AsyncFileSystem‎.

I believe, that your previous implementation doesn't do this for `UPath("file:///some/file.txt")` (so uri-style local paths).


For now, until `UPath(...).storage_options` is available, you should be able to fix your code by changing:

https://github.com/dagster-io/dagster/blob/1b6adb3e2579fe269d731682a73a63ce204659ce/python_modules/dagster/dagster_tests/storage_tests/test_upath_io_manager.py#L529-L542

It depends a bit if you still want to support `pathlib` paths.

```python
        if isinstance(path, UPath):
            so = pth.fs.storage_options.copy()
            cls = type(pth.fs)
            if cls is fsspec.implementations.local.LocalFileSystem:
                cls = morefs.asyn_local.AsyncLocalFileSystem
            so["asynchronous"] = True
            return cls(**so)
        elif isinstance(path, Path):
            try:
                from morefs.asyn_local import AsyncLocalFileSystem

                return AsyncLocalFileSystem()
            except ImportError as e:
                raise RuntimeError(
                    "Install 'morefs[asynclocal]' to use `get_async_filesystem` with a local"
                    " filesystem"
                ) from e
        else:
            raise DagsterInvariantViolationError(
                f"Path type {type(path)} is not supported by the UPathIOManager"
            )
```

I haven't tested it, but I think this should work now across upath versions. 
And it should do the correct thing for local paths `UPath('/abc/file.txt')` and `UPath('file:///abc/file.txt')` and `pathlib.Path('/abc/file.txt')`


Let me know if this helps,
I'll try collect some of this in an issue in the universal_pathlib repository, later today or tomorrow.

smackesey:
Thanks @ap-- your suggestion worked! https://github.com/dagster-io/dagster/pull/15732

PR #15025, opened 2023-06-29
Vary the target when downloading example projects

jmsanders:
We recently changed where the download script downloads from:

https://github.com/dagster-io/dagster/pull/14754

Varying the target between "master" for untagged versions and the version for tagged versions.

But we didn't make the same update to our path replacement logic so when running `dagster project from-example --example tutorial` from 1.3.11 or 1.3.12, no files in the downloaded set of files (all begining with `dagster-1.3.11/`) would match our filter for files beginning with `dagster-master`. Consequently, the command would no-op.

This makes a few small tweaks to the script:
1. It varies target between master and the version in both places
2. It updates our filter to include a trailing slash. Otherwise, we get weird interactions if you try to build an example like `tutorial` that is a subset of another example you already have downloaded (like `tutotial_dbt_dagster`.

I tested this by running locally with both a versioned and unversioned dagster.

### DIFF
--- a/python_modules/dagster/dagster/_generate/download.py
+++ b/python_modules/dagster/dagster/_generate/download.py
@@ -43,12 +43,18 @@ AVAILABLE_EXAMPLES = [
 ]
 
 
-def _get_url_for_version(version: str) -> str:
+def _get_target_for_version(version: str) -> str:
     if version == "1!0+dev":
         target = "master"
     else:
         target = version
-    return f"https://codeload.github.com/dagster-io/dagster/tar.gz/{target}"
+    return target
+
+
+def _get_url_for_version(version: str) -> str:
+    return (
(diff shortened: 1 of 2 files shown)


smackesey (approved):
Excellent

python_modules/dagster/dagster/_generate/download.py
```diff
@@ -43,12 +43,18 @@
...
 
 
-def _get_url_for_version(version: str) -> str:
+def _get_target_for_version(version: str) -> str:
     if version == "1!0+dev":
         target = "master"
     else:
         target = version
-    return f"https://codeload.github.com/dagster-io/dagster/tar.gz/{target}"
+    return target
```
smackesey, line 51: nit: `return "master" if version == "1!0+dev" else version`

PR #15030, opened 2023-06-29
[from-example] add cli flag for version

alangenfeld:
make the versioned code path more accessible and testable, then put under test

## How I Tested These Changes

ensured new test fails when #15025 is reverted

### DIFF
--- a/python_modules/dagster/dagster_tests/cli_tests/test_project_commands.py
+++ b/python_modules/dagster/dagster_tests/cli_tests/test_project_commands.py
@@ -90,6 +90,28 @@ def test_from_example_command_succeeds():
         assert not os.path.exists("my_dagster_project/tox.ini")
 
 
+def test_from_example_command_versioned_succeeds():
+    runner = CliRunner()
+    with runner.isolated_filesystem():
+        result = runner.invoke(
+            from_example_command,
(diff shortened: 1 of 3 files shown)


alangenfeld:
* master
  * **PR #15030** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15030" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15030?utm_source=stack-comment).

python_modules/dagster/dagster_tests/cli_tests/test_project_commands.py
```diff
@@ -90,6 +90,28 @@ def test_from_example_command_succeeds():
         assert not os.path.exists("my_dagster_project/tox.ini")
 
 
+def test_from_example_command_versioned_succeeds():
```
jmsanders, line 93: Nice

A couple other useful regression tests might be:
- testing to make sure it fails correctly if the path already exists
- testing to make sure it doesn't fail if a path that starts with the same prefix already exists (foo succeeds even if foobar exists)

jmsanders (approved):

PR #15034, opened 2023-06-29
Fetch AssetRecords all at once in CachingStaleStatusResolver

smackesey:
## Summary & Motivation

See https://elementl-workspace.slack.com/archives/C03CCE471E0/p1687988463503799

This changes the `CachingStaleStatusResolver` to fetch and cache all `AssetRecord` objects for the asset graph in one shot. This prevents one-query-per-asset, which can be prohibitively expensive in large asset graphs.

I am not sure if it is best to fetch _all_ the records in every case though-- one alternative possibility is to use a counter in the resolver, and do solo queries until some threshold N has been crossed, at which point we fetch them all.

The major wrinkle in the implementation here is that source assets can't be handled with `AssetRecord`, because the `AssetRecord` contains the last materialization but not the last observation. In theory this could also pose a problem for materializabele assets if someone is manually yielding observations for them with `DataVersion`, but I doubt this is happening anywhere. Regardless we should start including the most recent observation on the `AssetRecord`.

## How I Tested These Changes

Existing test suite.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/data_version.py
+++ b/python_modules/dagster/dagster/_core/definitions/data_version.py
@@ -25,9 +25,11 @@ from dagster._utils.cached_method import cached_method
 
 if TYPE_CHECKING:
     from dagster._core.definitions.asset_graph import AssetGraph
-    from dagster._core.definitions.events import AssetKey
+    from dagster._core.definitions.events import AssetKey, AssetMaterialization, AssetObservation
+    from dagster._core.event_api import EventLogRecord
     from dagster._core.events.log import EventLogEntry
     from dagster._core.instance import DagsterInstance
+    from dagster._utils.caching_instance_queryer import CachingInstanceQueryer
 
 
 class UnknownValue:
@@ -282,6 +284,7 @@ class CachingStaleStatusResolver:
     """
 
     _instance: "DagsterInstance"
+    _instance_queryer: Optional["CachingInstanceQueryer"]
     _asset_graph: Optional["AssetGraph"]
     _asset_graph_load_fn: Optional[Callable[[], "AssetGraph"]]
 
@@ -293,6 +296,7 @@ class CachingStaleStatusResolver:
         from dagster._core.definitions.asset_graph import AssetGraph
 
         self._instance = instance
+        self._instance_queryer = None
         if isinstance(asset_graph, AssetGraph):
             self._asset_graph = asset_graph
             self._asset_graph_load_fn = None
@@ -341,7 +345,7 @@ class CachingStaleStatusResolver:
         dependency_keys = self.asset_graph.get_parents(key)
 
         # only used if no provenance available
-        materialization = check.not_none(self._get_latest_materialization_event(key=key))
+        materialization = check.not_none(self._get_latest_data_version_record(key=key))
         materialization_time = materialization.timestamp
 
         if provenance:
@@ -405,7 +409,7 @@ class CachingStaleStatusResolver:
             # provenance is on all newer materializations. If dep is a source, then we'll never
             # provide a stale reason here.
             elif not self.asset_graph.is_source(dep_key):
-                dep_materialization = self._get_latest_materialization_event(key=dep_key)
+                dep_materialization = self._get_latest_data_version_record(key=dep_key)
                 if dep_materialization is None:
                     # The input must be new if it has no materialization
                     yield StaleCause(key, StaleCauseCategory.DATA, "has a new input", dep_key)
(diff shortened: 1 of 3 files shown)


smackesey:
* master
  * **PR #15034** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15034" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15034?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/data_version.py
```diff
@@ -508,3 +529,11 @@ def _is_volatile(self, *, key: AssetKey) -> bool:
         else:
             deps = self.asset_graph.get_parents(key)
             return len(deps) == 0 or any(self._is_volatile(key=dep_key) for dep_key in deps)
+
+    def _get_asset_record(self, *, key: AssetKey) -> Optional["AssetRecord"]:
+        # When targeting an asset, fetch records for all of its ancestors in
+        # one shot as a performance optimization. The records for all ancestors
+        # may be needed in determining the status.
+        ancestors = self.asset_graph.get_ancestors(key)
```
OwenKephart, line 537: This is O(n_ancestors^2), which can cause problems in larger asset graphs (we end up calling _get_asset_record for all the ancestors, so each of those ancestors will need to get their set of ancestors).

Two possible fixes:
- a public CachingInstanceQueryer.has_cached_asset_record() method, which we can call before this to avoid the get_ancestors call if we already have the record (and therefore can skip all this other stuff)
- turning get_ancestors into a cached method

I think the second option is still O(N^2) though as the prefetch_asset_records thing will need to iterate through the list of ancestors, so I think the first option is the way to go

python_modules/dagster/dagster/_core/definitions/data_version.py
```diff
@@ -450,24 +454,41 @@ def asset_graph(self) -> "AssetGraph":
...
-            is_source,
-        )
-        if event is None and is_source:
-            return DEFAULT_DATA_VERSION
-        elif event is None:
-            return NULL_DATA_VERSION
+        # Currently we can only use asset records, which are fetched in one shot, for non-source
+        # assets. This is because the most recent AssetObservation is not stored on the AssetRecord.
+        if self.asset_graph.is_source(key):
+            event = self._instance.get_latest_data_version_record(
```
OwenKephart, line 472: this can be replaced with a `self.instance_queryer.get_latest_materialization_or_observation_record()`, which handles reading from the asset record cache for you. The benefit here is that if we plan on storing observations in the record cache, this code can immediately take advantage of that performance boost

python_modules/dagster/dagster/_core/definitions/data_version.py
```diff
@@ -508,3 +529,11 @@ def _is_volatile(self, *, key: AssetKey) -> bool:
         else:
             deps = self.asset_graph.get_parents(key)
             return len(deps) == 0 or any(self._is_volatile(key=dep_key) for dep_key in deps)
+
+    def _get_asset_record(self, *, key: AssetKey) -> Optional["AssetRecord"]:
```
OwenKephart, line 533: I think we can avoid this _get_asset_record method entirely (it's too specific, you just need the latest materialization event), and instead have something like:

```
def get_latest_materialization_or_observation_record(self, key):
    if not self.asset_graph.is_source(key) and not self.instance_queryer.has_cached_asset_record(key):
        ancestors = self.instance_queryer.prefetch_asset_records(ancestors)
        self.instance_queryer.prefetch_asset_records(ancestors)
    return self.instance_queryer.get_latest_materialization_or_observation_record(key)
```

this seems like the cleanest way to interface with the existing functionality imo

OwenKephart (changes requested):

sryza:
is there a risk that this causes the asset details page to load much slower than it currently does?

OwenKephart:
@sryza very little risk (post-my suggestions) -- we'd need to pull in this amount of information anyway so this just batches it all together

sryza:
@OwenKephart it doesn't make it O(# of assets) when it could have been O(1)?

OwenKephart:
@sryza the staleness calculation for a single asset is already O(n) as we need to walk up the tree to find any upstream issues. I believe if the asset is not stale, then we will end up walking up the entire set of ancestors every time, but in cases where the asset is stale, then we only have to walk up to the "point of failure", so there are cases where this would overfetch data. However, this isn't asymptotically more data than we'd have to fetch otherwise, and we expect the common case to be "all assets non-stale", so I think this is a reasonable tradeoff.

We could do something along the lines of what @smackesey suggested on the original thread, where we're allowed to do N individual fetches before we give up and decide to fetch everything upstream of the asset, but I'm not sure how much of a benefit that would yield, compared to the cost of having to do those extra N individual fetches in the common case

python_modules/dagster/dagster/_core/definitions/data_version.py
smackesey: Good suggestion, I've updated the PR to incorporate this.

python_modules/dagster/dagster/_core/definitions/data_version.py
smackesey: >a public CachingInstanceQueryer.has_cached_asset_record() method, which we can call before this to avoid the get_ancestors call if we already have the record (and therefore can skip all this other stuff)

I implemented this.

smackesey:
>the staleness calculation for a single asset is already O(n) as we need to walk up the tree to find any upstream issues. I believe if the asset is not stale, then we will end up walking up the entire set of ancestors every time, but in cases where the asset is stale, then we only have to walk up to the "point of failure", so there are cases where this would overfetch data. However, this isn't asymptotically more data than we'd have to fetch otherwise, and we expect the common case to be "all assets non-stale", so I think this is a reasonable tradeoff.

I concur

>We could do something along the lines of what @smackesey suggested on the original thread, where we're allowed to do N individual fetches before we give up and decide to fetch everything upstream of the asset, but I'm not sure how much of a benefit that would yield, compared to the cost of having to do those extra N individual fetches in the common case

I think there's a good chance there is no significant negative perf impact from this so I think we should hold off unless we see an issue.

sryza:
Ahh, the following led me to believe we'd be fetching AssetRecords for every asset in the graph.

> This changes the CachingStaleStatusResolver to fetch and cache all AssetRecord objects for the asset graph in one shot. This prevents one-query-per-asset, which can be prohibitively expensive in large asset graphs.

However, if I understand correctly, we'll only be fetching ancestors? If so, that seems fine to me.

In large deployments, # ancestors will usually be << # total assets in deployment.

smackesey:
@sryza Sorry, the initial version of this PR _did_ fetch for all ancestors at once, then it was refined at suggestion of @OwenKephart .

PR #15036, opened 2023-06-29
move examples from DbtCliClientResource to DbtCli

sryza:

PR #15038, opened 2023-06-29
change default profiles dir in dbt examples

sryza:
This makes the examples cleaner, and also lets people run the dbt projects by just going into the project directories and running `dbt build` etc.

DbtCli will automatically look for the profiles.yml inside the root of the dbt projects.


examples/assets_dbt_python/dbt_project/profiles.yml
rexledesma, line 1: `profiles.yml` was move to the top level, but we should also just remove the `config/` directory in the dbt project. The user id shouldn't be checked in anyways.

rexledesma (changes requested):
This will not work out of the box since `load_assets_from_dbt_project` needs to be compatible with `DbtCli`.

We will need to branch on the underlying op logic, depending on what resource was passed in.

examples/assets_dbt_python/dbt_project/profiles.yml
sryza: oops I missed that - thanks for catching

rexledesma (commented):
This change should just be merged with the previous PR since `DbtCli` doesn't accept `profiles_dir` in the first place

sryza:
closing and smushing into https://github.com/dagster-io/dagster/pull/15036

sryza:
because they can't be merged independently

PR #15040, opened 2023-06-29
dagster-dbt docs, move to DbtCli and move profiles.yml to root of project

sryza:
## Summary & Motivation

This makes the dbt docs and tutorial a little cleaner

Depends on unexperimentalizing the DbtCli: https://github.com/dagster-io/dagster/pull/15068

### DIFF
--- a/docs/content/integrations/dbt/using-dbt-with-dagster/load-dbt-models.mdx
+++ b/docs/content/integrations/dbt/using-dbt-with-dagster/load-dbt-models.mdx
@@ -36,11 +36,8 @@ from dagster import file_relative_path
 
 
 DBT_PROJECT_PATH = file_relative_path(__file__, "../../jaffle_shop")
-DBT_PROFILES = file_relative_path(__file__, "../../jaffle_shop/config")
 
-dbt_assets = load_assets_from_dbt_project(
-    project_dir=DBT_PROJECT_PATH, profiles_dir=DBT_PROFILES, key_prefix=["jaffle_shop"]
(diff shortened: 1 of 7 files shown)


docs/content/integrations/dbt/using-dbt-with-dagster/load-dbt-models.mdx
```diff
@@ -73,32 +69,27 @@ Let's take a look at the arguments we've supplied:
...
-}
+resources = {"dbt": DbtCli(project_dir=DBT_PROJECT_PATH)}
 
 defs = Definitions(assets=load_assets_from_modules([assets]), resources=resources)
 ```
 
 Let's take a look at what's happening here:
 
-- In the `resources` key, we've provided configuration info for the <PyObject module="dagster_dbt" object="DbtCliClientResource" /> resource.
+- In the `resources` key, we've provided configuration info for the <PyObject module="dagster_dbt" object="DbtCli" /> resource.
```
rexledesma, line 92: I don't believe `DbtCli` is in our API docs yet, should double check that this link works.

rexledesma (approved):

docs/content/integrations/dbt/using-dbt-with-dagster/load-dbt-models.mdx
sryza: Ah good point. I updated this PR to make that change: https://github.com/dagster-io/dagster/pull/15068.

PR #15043, opened 2023-06-29
remove with_resources from dbt reference

sryza:


erinkcochran87 (approved):
Good catch - thanks Sandy!

PR #15046, opened 2023-06-29
[AutoMaterialize] Add asset keys to ParentOutdatedAutoMaterializeCondition

OwenKephart:
## Summary & Motivation

Tell people which assets we're waiting on!

I had a terrible time trying to think of a good name for the function that returns the things that we're waiting on, very open to suggestions.

TODO: more in-depth tests

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -680,9 +680,9 @@ def determine_asset_partitions_to_auto_materialize(
         can_reconcile_fn=can_reconcile_candidate,
     )
 
-    def get_unresolved_parent_asset_keys(candidate: AssetKeyPartitionKey) -> FrozenSet[AssetKey]:
-        """Returns the set of parent asset keys which would be unresolved if this candidate were
-        materialize this tick.
+    def get_waiting_on_asset_keys(candidate: AssetKeyPartitionKey) -> FrozenSet[AssetKey]:
+        """Returns the set of ancestor asset keys that must be materialized before this asset can be
+        materialized.
         """
         from dagster._core.definitions.external_asset_graph import ExternalAssetGraph
 
@@ -707,11 +707,9 @@ def determine_asset_partitions_to_auto_materialize(
                     == asset_graph.get_repository_handle(parent.asset_key)
                 )
             ):
-                print("PARENT", parent)
                 unresolved_parents.update(
-                    instance_queryer.unreconciled_parent_keys(asset_partition=parent)
+                    instance_queryer.to_be_materialized_before_reconciled(asset_partition=parent)
                 )
-                print(unresolved_parents)
         return frozenset(unresolved_parents)
 
     def conditions_for_candidate(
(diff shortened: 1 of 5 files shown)


python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -562,32 +563,34 @@ def get_asset_partitions_updated_after_cursor(
...
+        }
 
     @cached_method
-    def is_reconciled(self, *, asset_partition: AssetKeyPartitionKey) -> bool:
-        """Returns a boolean representing if the given `asset_partition` is currently reconciled.
-        An asset (partition) is considered unreconciled if any of:
-        - It has never been materialized
-        - One of its parents has been updated more recently than it has
-        - One of its parents is unreconciled.
+    def to_be_materialized_before_reconciled(
```
sryza, line 573: Would it make sense to call this `get_unreconciled_ancestors`?

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -679,40 +680,37 @@ def can_reconcile_candidate(candidate: AssetKeyPartitionKey) -> bool:
...
-                    # alongside the candidate
-                    and asset_graph.have_same_partitioning(parent.asset_key, candidate.asset_key)
-                    and parent.partition_key == candidate.partition_key
-                    # the parent must be in the same repository to be materialized alongside the candidate
-                    and (
-                        not isinstance(asset_graph, ExternalAssetGraph)
-                        or asset_graph.get_repository_handle(candidate.asset_key)
-                        == asset_graph.get_repository_handle(parent.asset_key)
-                    )
+        unresolved_parents = set()
```
sryza, line 689: This should be "ancestors", not parents, right? Also, would `unreconciled` make sense instead of `unresolved`? Wary of introducing a new term here.

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -741,8 +739,11 @@ def conditions_for_candidate(
             conditions.add(ParentMaterializedAutoMaterializeCondition())
 
         # if the parents will not be resolved this tick
-        if get_unresolved_parent_asset_keys(candidate):
-            conditions.add(ParentOutdatedAutoMaterializeCondition())
+        waiting_on_asset_keys = get_waiting_on_asset_keys(candidate)
+        if waiting_on_asset_keys:
+            conditions.add(
+                ParentOutdatedAutoMaterializeCondition(waiting_on_asset_keys=waiting_on_asset_keys)
```
sryza, line 745: If I understand correctly, this will include all the ancestors that we're waiting on, not just the ones at the root of the ancestor graph?

Is that what we want to show on the frontend?

I generally think the frontend here should be pretty dumb, and if we're going to filter the intermediate assets out, we should do it here, not there.

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
OwenKephart: ah yes, this was me trying to play around with the potential new terminology but putting it back.

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
OwenKephart: ah so the function on the caching instance queryer actually does that filtering already -- so if you have:

`A -> B -> C -> D -> E`, and `A` has just been updated, the set of keys that we're waiting on would be identical for `C`, `D`, or `E` (always just `B`). I added a test confirming this behavior.

Updated the name of the function to be a bit clearer here

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
OwenKephart: went w/ get_unreconciled_root_ancestors

sryza (approved):
The implementation looks good to me.  Will leave it up to your judgment on what/whether additional tests are needed.

PR #15047, opened 2023-06-30
enable using the DbtCli resource with load_assets_from_dbt_x functions

sryza:
## Summary & Motivation

Thank you Owen for helping me understand some of this code.


rexledesma (changes requested):
not so bad

python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
```diff
@@ -222,26 +232,47 @@ def _events_for_structured_json_line(
...
+                manifest_json,
+            )
+    else:
+        if runtime_metadata_fn is not None:
+            raise DagsterDbtError(
+                "The runtime_metadata_fn argument on the load_assets_from_dbt_... functions is not"
+                " supported when using the DbtCli resource. Use the @dbt_assets decorator instead"
+                " of load_assets_from_dbt_... if you want control over what metadata is yielded at"
+                " runtime."
+            )
```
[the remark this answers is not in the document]
sryza: > Can we raise it at definition time?

I'm not sure where we'd do this. I agree that it's possible to know this before executing the run. However, at the time `load_assets_from_dbt...` is invoked, we don't know what resources will be provided to it.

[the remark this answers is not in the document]
sryza: > We could support runtime_metadata_fn if we want to using DbtCli.

Yeah I figured it would just be a waste if we're planning to deprecate it anyway.

python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
```diff
@@ -711,3 +758,22 @@ def load_assets_from_dbt_manifest(
...
         dbt_assets, _ = prefix_assets(dbt_assets, key_prefix, [], source_key_prefix)
     return dbt_assets
+
+
+def build_manifest_from_fn_args(
+    manifest_json: Mapping[str, Any],
+    node_info_to_asset_key: Callable[[Mapping[str, Any]], AssetKey] = default_asset_key_fn,
+    node_info_to_definition_metadata_fn: Callable[
+        [Mapping[str, Any]], Mapping[str, MetadataUserInput]
+    ] = default_metadata_fn,
```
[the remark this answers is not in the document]
sryza: Ok yeah sounds reasonable

python_modules/libraries/dagster-dbt/dagster_dbt_tests/conftest.py
```diff
@@ -51,12 +53,15 @@ def dbt_python_config_dir():
...
+        "DbtCliClientResource",
+    ]
+    + (["DbtCli"] if version.parse(dbt.version.__version__) >= version.parse("1.4.0") else []),
 )
 def dbt_cli_resource_factory(request):
-    if request.param == "pythonic":
+    if request.param == "DbtCliClientResource":
         return DbtCliClientResource
+    elif request.param == "DbtCli":
+        return DbtCli
```
rexledesma, line 64: We should do the version check here, and then call if `pytest.skip()` if it fails. Just so that it's explicit in the test suite that we are omitting tests: https://docs.pytest.org/en/7.1.x/how-to/skipping.html#skipping-test-functions

python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
```diff
@@ -222,26 +232,49 @@ def _events_for_structured_json_line(
...
+                runtime_metadata_fn,
+                manifest_json,
+            )
+    else:
+        if runtime_metadata_fn is not None:
+            raise DagsterDbtError(
+                "The runtime_metadata_fn argument on the load_assets_from_dbt_... functions is not"
+                " supported when using the DbtCli resource. Use the @dbt_assets decorator instead"
+                " of load_assets_from_dbt_... if you want control over what metadata is yielded at"
+                " runtime."
```
rexledesma, line 265: There's two `load_assets_from_dbt_...` functions, so let's just be clear and write both of them out?

```suggestion
                "The runtime_metadata_fn argument on the `load_assets_from_dbt_project` and `load_assets_from_dbt_manifest` functions is not"
                " supported when using the DbtCli resource. To control the metadata yielded at runtime with the DbtCli resource, use the @dbt_assets decorator"
```

rexledesma (approved):

python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
sryza: 👍

python_modules/libraries/dagster-dbt/dagster_dbt_tests/conftest.py
sryza: 👍

PR #15055, opened 2023-06-30
feat(dbt): raise exception by default when dbt command is not successful

rexledesma:
## Summary & Motivation
The user should have control on what to do when a dbt command fails. By default, we should expose the failure to the user. Then, the user can either handle it by:

- suppressing the failure, and setting `raise_on_failure=False`, or
- catching the failure and handling the exception.

Before, the failure was not raised, which caused the underlying execution in Dagster to not display as failure.

## How I Tested These Changes
pytest

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
@@ -47,6 +47,7 @@ from ..asset_utils import (
     is_non_asset_node,
     output_name_fn,
 )
+from ..errors import DagsterDbtCliRuntimeError
 from ..utils import ASSET_RESOURCE_TYPES, select_unique_ids_from_manifest
 
 if TYPE_CHECKING:
@@ -690,12 +691,14 @@ class DbtCliTask:
         manifest (DbtManifest): The dbt manifest wrapper.
         project_dir (Path): The path to the dbt project.
         target_path (Path): The path to the dbt target folder.
+        raise_on_error (bool): Whether to raise an exception if the dbt command fails.
     """
 
     process: subprocess.Popen
     manifest: DbtManifest
     project_dir: Path
     target_path: Path
+    raise_on_error: bool
 
     @classmethod
     def run(
@@ -705,6 +708,7 @@ class DbtCliTask:
         manifest: DbtManifest,
         project_dir: Path,
         target_path: Path,
+        raise_on_error: bool,
     ) -> "DbtCliTask":
         # Attempt to take advantage of partial parsing. If there is a `partial_parse.msgpack` in
         # in the target folder, then copy it to the dynamic target path.
@@ -739,6 +743,7 @@ class DbtCliTask:
             manifest=manifest,
             project_dir=project_dir,
             target_path=target_path,
+            raise_on_error=raise_on_error,
         )
 
     def wait(self) -> Sequence[DbtCliEventMessage]:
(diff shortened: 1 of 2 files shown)


rexledesma:
* master
  * **PR #15055** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15055" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15055?utm_source=stack-comment).

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -827,7 +832,9 @@ def stream_raw_events(self) -> Iterator[DbtCliEventMessage]:
                 sys.stdout.flush()
 
         # Ensure that the dbt CLI process has completed.
-        self.process.wait()
+        self.is_successful()
```
sryza, line 835: We don't need this standalone call to `is_successful`, right?

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -827,7 +832,9 @@ def stream_raw_events(self) -> Iterator[DbtCliEventMessage]:
...
+        if not self.is_successful() and self.raise_on_error:
+            raise Failure
```
sryza, line 837: Based on [this thread](https://elementl-workspace.slack.com/archives/C03A0D72A6T/p1688142008101359) with @alangenfeld , it seems like the main reason to use `Failure` is if we want to add metadata and control the retries, but those don't seem to be the case here. It would be better to have a more specific error type.  Then users could catch it if they want to do their own special handling.

The existing error-handling code appears to distinguish between errors that cause the dbt CLI to fail unexpectedly, and errors that the dbt CLI handles but also reports. I don't know the background enough to be able to say whether it's important that we distinguish between these, so can probably leave that out for now, but just mentioning here for posterity.

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -705,6 +708,7 @@ def run(
         manifest: DbtManifest,
         project_dir: Path,
         target_path: Path,
+        raise_on_error: bool,
```
sryza, line 711: I think we should default this to `True`, like we do for `execute_in_process` and `materialize`.

sryza (commented):
Thanks for the quick turnaround. It would be good to include a test.

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
rexledesma: We default to true in `.cli(...)`

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
rexledesma: I will just use `DagsterDbtCliRuntimeError` since that already exists.

python_modules/libraries/dagster-dbt/dagster_dbt_tests/cli/test_resources_v2.py
```diff
@@ -38,6 +38,9 @@ def test_dbt_cli_failure() -> None:
     dbt = DbtCli(project_dir=TEST_PROJECT_DIR)
     dbt_cli_task = dbt.cli(["run", "--profiles-dir", "nonexistent"], manifest=manifest)
 
+    with pytest.raises(Exception):
```
sryza, line 41: This could be `pytest.raises(DagsterDbtCliRuntimeError)`, right?

sryza (approved):
Left one last comment - otherwise this looks great

PR #15056, opened 2023-06-30
non_argument_deps rename - update dagstermill to use deps

jamiedemaria:
## Summary & Motivation
adds `deps` param and just passes through `non_argument_deps` and `deps` to `@asset` to be handled by the decorator
## How I Tested These Changes
bk

PR #14959, opened 2023-06-26
[wip] update airbyte integration to use deps

jamiedemaria:

PR #14958, opened 2023-06-26
non_argument_deps rename -  update dbt integration to use deps

jamiedemaria:
## Summary & Motivation
updates dbt integration to use `deps` instead of `non_argument_deps`

PR #14896, opened 2023-06-21
[RFC] non_argument_deps rename exploration - passing AssetsDefinitions

jamiedemaria:
## Summary & Motivation
Play around with passing AssetsDefinition objects instead of magic strings

PR #15059, opened 2023-06-30
[docs] non_argument_deps rename - update docs snippets to use deps

jamiedemaria:
## Summary & Motivation
* updates the tutorial to introduce `deps` first and then add on I/O managers as a later step.
* updates the SDA concept page to introduce `deps` first and argument based dependencies next
* updates docs snippets to use `deps` instead of `non_argument_deps`


jamiedemaria:
* master
  * **PR #14886** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14886" target="_blank">Graphite</a> 
    * **PR #14888** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14888" target="_blank">Graphite</a> 
      * **PR #14958** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14958" target="_blank">Graphite</a> 
        * **PR #14959** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14959" target="_blank">Graphite</a> 
          * **PR #15052** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15052" target="_blank">Graphite</a> 
            * **PR #15056** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15056" target="_blank">Graphite</a> 
              * **PR #15059** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15059" target="_blank">Graphite</a>  👈
      * **PR #14887** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14887" target="_blank">Graphite</a> 
        * **PR #14912** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14912" target="_blank">Graphite</a> 
        * **PR #14908** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14908" target="_blank">Graphite</a> 
        * **PR #14896** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14896" target="_blank">Graphite</a> 
        * **PR #14889** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/14889" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15059?utm_source=stack-comment).

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -188,9 +188,9 @@ Using source assets has a few advantages over having the code inside of an asset
 - **Dagster can use data-loading code factored into an <PyObject object="IOManager" /> to load the contents of the source asset**.
 - **Asset dependencies can be written in a consistent way,** independent of whether they're downstream from a source asset or a derived asset. This makes it easy to swap out a source asset for a derived asset and vice versa.
 
-#### Non-argument dependencies
+#### Defining dependencies that don't load data
```
jamiedemaria, line 191: wasn't sure what to rename this - open to suggestions!

erinkcochran87: Chatted with Tim about this heading/section - some initial thoughts:

- Should this copy explicitly state that this has to do with I/O managers? Or no, since we're trying to simplify concepts?
- One thing this section is missing is when you'd use this instead of just normal dependencies. Ex: Setting up the order of assets, data is too big to load, etc.
- Title may actually be fine if we add some of the why/when to this

Thoughts?

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -67,15 +67,38 @@ By default, the name of the decorated function, `my_asset`, is used as the asset
...
+- [Basic data-loading asset dependencies](#defining-explicit-dependencies)
+- [Explicit data-loading asset dependencies](#defining-explicit-dependencies)
 - [External asset dependencies](#defining-external-asset-dependencies)
 
 #### Defining basic dependencies
 
-The easiest way to define an asset dependency is to include an upstream asset name as an argument to the decorated function.
+You can define a dependency between two assets by passing the upstream asset to the `deps` parameter in the downstream asset's `@asset` decorator.
+
+In this example, `upstream_asset` creates a new table (`sugary_cereals`) by selecting records from the `cereals` table. Then `downstream_asset` creates a new table (`shopping_list`) by selecting records from `sugary_cereals`
```
erinkcochran87, line 83:
```suggestion
In this example, `upstream_asset` creates a new table (`sugary_cereals`) by selecting records from the `cereals` table. Then `downstream_asset` creates a new table (`shopping_list`) by selecting records from `sugary_cereals`:
```

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -140,8 +165,16 @@ from dagster import AssetKey, SourceAsset, asset
 my_source_asset = SourceAsset(key=AssetKey("a_source_asset"))
 
 
+@asset(deps=[my_source_asset])
+def my_derived_asset():
+    return execute_query("SELECT * from a_source_asset").as_list() + [4]
+```
+
+You can also define dependencies on `SourceAsset`s that will load the data of the asset:
```
erinkcochran87, line 173:
```suggestion
You can also define dependencies on a `SourceAsset` that will load the data of the asset:
```

This feels nitpicky - sorry!

erinkcochran87 (commented):
This is looking really good Jamie! The main ? I had was in the **Defining basic dependencies** section, do we need to explicitly say that using this approach means data won't be loaded? It's implied in the code and in the other sections (because they specifically say data *will* be loaded), but I wonder if calling it out is a good idea. That being said, I think the section headings are fine - I'd put this info in the section copy somewhere, if we want to do that.

jamiedemaria:
> do we need to explicitly say that using this approach means data won't be loaded?

good question. I feel like i could be swayed either way, but in general I think that since the default workflow has been to load the assets into memory, it's good to specify that this API doesn't load the asset into memory. I also feel like if i were writing an asset and i wanted to understand how `deps` works I would want that information to be in the section about `deps`. I wouldn't want to have to read the section about the argument/data loading dependencies and then draw the conclusion that `deps` does the opposite

Maybe once `deps` graduates into being the default experience we could remove the specification

jamiedemaria:
TODO - should also update the hello dagster example?? https://docs.dagster.io/getting-started/hello-dagster

docs/content/tutorial/writing-your-first-asset.mdx
```diff
@@ -22,45 +22,65 @@ By the end of this section, you will:
...
-def topstory_ids(): # turn it into a function
+
+def topstory_ids() -> None: # turn it into a function
     newstories_url = "https://hacker-news.firebaseio.com/v0/topstories.json"
     top_new_story_ids = requests.get(newstories_url).json()[:100]
-    return top_new_story_ids # return the data
+
+    os.makedirs("data", exist_ok=True)
+    with open("data/topstory_ids.json", "w") as f:
+        f.write(json.dumps((top_new_story_ids)))
```
sryza, line 64: Are all these parentheses needed?

sryza: Also, I think you can just do `json.dump` to avoid needing to invoke `f.write`. Same on the read side too.

docs/content/tutorial/building-an-asset-graph.mdx
```diff
@@ -46,20 +52,13 @@ def topstories(topstory_ids):  # this asset is dependent on topstory_ids
             print(f"Got {len(results)} items so far.")  # noqa: T201
 
     df = pd.DataFrame(results)
-
-    return df
+    df.to_csv("data/topstories.csv")
 ```
 
-Dependencies between assets are defined by adding the asset name as a parameter to the dependent asset's function. In this case, `topstory_ids` (the list of IDs) is a dependency of `topstories` (the DataFrame).
+Dependencies between assets are defined using the `deps` parameter of the `@asset` decorator. In this case, `topstory_ids` (the list of IDs) is a dependency of `topstories` (the DataFrame).
```
sryza, line 58: Does it still make sense to say "the DataFrame" here, given that we're not returning a DataFrame?

docs/content/tutorial/introduction.mdx
```diff
@@ -25,19 +25,29 @@ If you have an existing data pipeline, you likely already have assets.
 
 _Software-defined assets_ are a Dagster concept that allows you to write data pipelines in terms of the assets that they produce. How they work: you write code that describes an asset that you want to exist, along with any other assets that the asset is derived from, and a function that can be run to compute the contents of the asset.
 
-Here's an example. The asset is a dataset called `topstories`, and it's computed by starting with an asset called `topstory_ids` and then fetching data for each of those IDs.
+Here's an example. The asset is a dataset called `topstories`, and it depends on another asset called `topstory_ids`. `topstories` gets the IDs computed in `topstory_ids`, then fetches data for each of those IDs.
 
 ```python
-@asset
-def topstories(topstory_ids):
+# TODO - ensure updated with the final version of this asset
```
sryza, line 31: Make sure to address the TODOs and `noqa` before merging

sryza (commented):
Left a few comments but overall looking good.

Should we delete the "managing-your-own-io" section?

jamiedemaria:
> Should we delete the "managing-your-own-io" section? 

yeah i was thinking we could replace it with a section about using dagster's io managers

docs/content/tutorial/building-an-asset-graph.mdx
jamiedemaria: yeah i suppose it's really the csv file that's the asset

docs/content/tutorial/writing-your-first-asset.mdx
jamiedemaria: 🤦

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -67,15 +67,38 @@ By default, the name of the decorated function, `my_asset`, is used as the asset
...
+def upstream_asset() -> None:
+    execute_query("CREATE TABLE sugary_cereals AS SELECT * FROM cereals")
+
+
+@asset(deps=[upstream_asset])
+def downstream_asset() -> None:
+    execute_query("CREATE TABLE shopping_list AS SELECT * FROM sugary_cereals")
+```
+
+#### Defining basic data-loading dependencies
```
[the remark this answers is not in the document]
jamiedemaria: to me, i don't know what "managed-loading" means off the top of my head. we could definitely give it meaning, but then i wonder if people skimming through the concepts to find the answer to a question will miss this section because they don't know what the header means

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -67,15 +67,38 @@ By default, the name of the decorated function, `my_asset`, is used as the asset
...
+
+The easiest way to define an asset dependency that will load the contents of an upstream asset is to include the upstream asset name as an argument to the decorated function.
```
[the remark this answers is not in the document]
jamiedemaria: Yeah this elaboration is good, one qq thought

> it's expected that the code you include inside your @asset--decorated function will load the data from the upstream asset.  

that's not quite true though right? because you might just be running SQL queries on the table created by the upstream asset, or launching some external process that then uses the upstream asset

jamiedemaria:
TODO - some of the screenshots in the first two sections have the `path` metadata filled in by the IO manager. since they aren't using the IO manager anymore we should recapture the screenshots

docs/content/tutorial/saving-your-data.mdx
```diff
@@ -3,33 +3,44 @@ title: "Tutorial, part six: Saving your data | Dagster Docs"
...
-They manage **input** by reading an asset from where it’s stored and loading it into memory to be used by a dependent asset. For example, `most_frequent_words` needs `topstories`. Therefore, the I/O manager that `topstories` uses will load the data into memory so the `most_frequent_words` asset can use it.
+They manage **input** by reading an asset from where it’s stored and loading it into memory to be used by a dependent asset. For example, `most_frequent_words` needs `topstories`. Therefore, you could use an I/O manager to load the `topstories` data into memory so that the `most_frequent_words` asset can use it.
+
+I/O managers control **output** by writing the asset's to the location configured. By the end of this section, you’ll configure an I/O manager to write the `topstories` DataFrame into a database as a table.
+
+---
+
+## When should you use I/O managers?
+
+### TODO WRITE THIS SECTION
```
jamiedemaria, line 31: resolve this TODO

docs/content/tutorial/saving-your-data.mdx
```diff
@@ -76,13 +87,115 @@ defs = Definitions(
...
+    # remove manually saving top_words
+
+    context.add_output_metadata(metadata={"plot": MetadataValue.md(md_content)})
+
+    return top_words  # return top_words and the I/O manager will save it
+```
+
+To see the effect of this change, reload your code location in the Dagster UI and materialize your assets. When you click on an asset, you’ll see the `path` metadata pointing to the `data` directory on your computer.
+
+### TODO - some kind of closing statement that make sense
```
jamiedemaria, line 198: resolve this TODO

docs/content/tutorial/saving-your-data.mdx
```diff
@@ -150,15 +263,18 @@ def topstories(topstory_ids):
         if len(results) % 20 == 0:
             logger.info(f"Got {len(results)} items so far.")
 
-    df = pd.DataFrame(results).drop(["kids"], axis=1)
+    df = pd.DataFrame(results).drop(
+        ["kids"], axis=1
+    )  # TODO - why are we dropping kids column here, but not in the other versions of this asset?
```
jamiedemaria, line 268: resolve this TODO - cc @erinkcochran87  maybe you have some context as to why`.drop(["kids"], axis=1) is added in here?

examples/docs_snippets/docs_snippets/tutorial/saving/assets.py
```diff
@@ -22,15 +24,18 @@ def topstories(topstory_ids):
         if len(results) % 20 == 0:
             logger.info(f"Got {len(results)} items so far.")
 
-    df = pd.DataFrame(results).drop(["kids"], axis=1)
+    df = pd.DataFrame(results).drop(
+        ["kids"], axis=1
+    )  # TODO - why are we dropping kids column here, but not in the other versions of this asset?
```
jamiedemaria, line 29: resolve this todo

docs/content/tutorial/saving-your-data.mdx
```diff
@@ -76,13 +87,115 @@ defs = Definitions(
   information.
 </Note>
 
-To see the effect of this change, reload your code location in the Dagster UI and materialize your assets. When you click on an asset, you’ll see the `path` metadata pointing to the `data` directory on your computer.
+Now, you will modify your three assets `topstory_ids`, `topstories`, and `most_frequent_words` to use the I/O manager to read and write data rather than direct calls to the `json` and `pandas` libraries. To do this, you will need to modify how the dependencies between assets are defined, and remove manual reading and writing of data.
```
jamiedemaria, line 90: this paragraph and the following paragraph feel repetetive

docs/content/tutorial/saving-your-data.mdx
```diff
@@ -76,11 +88,111 @@ defs = Definitions(
   information.
 </Note>
 
+Now, you will modify your three assets `topstory_ids`, `topstories`, and `most_frequent_words` to use the I/O manager to read and write data rather than direct calls to the `json` and `pandas` libraries. To do this, you will:
+* Return your data from each asset, rather than manually writing it to the file system. The I/O manager will now save the data.
+* Modify how you define dependencies between assets. To use the I/O manager to load the data associated with an upstream asset into memory, you use the upstream asset's name as the name of one of the arguments to the downstream asset's function.
+
+
+Modify `topstory_ids` to use an I/O manager:
```
jamiedemaria, line 96: this set of modifications ends up being like a huge wall of code for the user to read through, but i'm not sure there's a better way to do this. let me know if you've got any thoughts, or if this seems fine

docs/content/tutorial/saving-your-data.mdx
```diff
@@ -3,33 +3,45 @@ title: "Tutorial, part six: Saving your data | Dagster Docs"
...
 
-They manage **input** by reading an asset from where it’s stored and loading it into memory to be used by a dependent asset. For example, `most_frequent_words` needs `topstories`. Therefore, the I/O manager that `topstories` uses will load the data into memory so the `most_frequent_words` asset can use it.
+They manage **input** by reading an asset from where it’s stored and loading it into memory to be used by a dependent asset. For example, `most_frequent_words` needs `topstories`. Therefore, you could use an I/O manager to load the `topstories` data into memory so that the `most_frequent_words` asset can use it.
 
-I/O managers control **output** by writing the asset's to the location configured. By the end of this section, you’ll configure an I/O manager to write the `hackernews_topstories` DataFrame into a database as a table.
+I/O managers control **output** by writing the asset's to the location configured. By the end of this section, you’ll configure an I/O manager to write the `topstories` DataFrame into a database as a table.
+
+---
+
+## When should you use I/O managers?
```
jamiedemaria, line 29: this section is still WIP. I've iterated on it a couple times, but am still not very happy with it. Maybe it would be better at the end of the section when the user is a bit more familiar with I/O managers? Basically I just want the user to feel empowered to decide when to use I/O managers

docs/content/tutorial/saving-your-data.mdx
```diff
@@ -76,11 +88,111 @@ defs = Definitions(
   information.
 </Note>
 
+Now, you will modify your three assets `topstory_ids`, `topstories`, and `most_frequent_words` to use the I/O manager to read and write data rather than direct calls to the `json` and `pandas` libraries. To do this, you will:
```
jamiedemaria, line 91: this section is still a WIP, and also not super happy with it yet. I'm struggling a bit to concisely explain the changes that need to be made and why

docs/content/concepts/assets/software-defined-assets.mdx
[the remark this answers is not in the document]
sryza: My concern about "data-loading" dependencies is that, in many cases, `deps` dependencies will involve loading data, it's just that it's happening inside the `@asset`-decorated function.

To anchor, do you have an example in mind of a question that they might be skimming to answer?

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -70,12 +70,33 @@ By default, the name of the decorated function, `my_asset`, is used as the asset
...
+
+
+@asset(deps=[sugary_cereals])
+def shopping_list() -> None:
+    execute_query("CREATE TABLE shopping_list AS SELECT * FROM sugary_cereals")
+```
+
+#### Defining basic data-loading dependencies
+
+When using basic dependencies, as above, it's expected that if you need direct access to the contents of the asset, the code you include inside your `@asset`-decorated function will load the data from the upstream asset. Dagster alternatively allows you to delegate loading data to an [I/O manager](\(/concepts/io-management/io-managers\)). To do this, you express the dependency by using the upstream asset name as the name of one of the arguments on the decorated function.
```
sryza, line 99: 👍 👍

docs/content/tutorial/saving-your-data.mdx
```diff
@@ -3,33 +3,45 @@ title: "Tutorial, part six: Saving your data | Dagster Docs"
 description: Learn how to use I/O managers to save your data.
 ---
 
-# Tutorial, part six: Saving your data
+# Tutorial, part six: Using Dagster to save and load your data
 
-You've written a data pipeline using assets that runs every hour. Currently, your assets are stored in a temporary folder that is deleted after closing the `dagster` process. **Note**: If you set up Dagster without the tutorial, you may have set a `DAGSTER_HOME` environment variable. If so, then your assets will be stored in that directory.
-
-You likely want to store your assets in places such as tables in your database or files in cloud storage. To do this, Dagster has I/O managers that enable reading and writing data to storage systems.
+You've written a data pipeline using assets that runs every hour. Currently, you are manually storing and loading your data using the `json` and `pandas` libraries at the beginning and end of each asset. Dagster provides I/O managers that will handle reading and writing data to storage for you. I/O managers allow you to reduce boilerplate reading and writing code and easily change where your data is stored.
```
sryza: Instead of leading with "Dagster provides I/O managers that will handle reading and writing data to storage for you." I think the main way I would message the point of using IO managers is that they help you separate business logic from IO.

As you mention, this makes code easier to test and easier to understand, and allows you to plug in generic built-in IO managers for various storage systems.

docs/content/tutorial/saving-your-data.mdx
```diff
@@ -3,33 +3,45 @@ title: "Tutorial, part six: Saving your data | Dagster Docs"
...
 
-They manage **input** by reading an asset from where it’s stored and loading it into memory to be used by a dependent asset. For example, `most_frequent_words` needs `topstories`. Therefore, the I/O manager that `topstories` uses will load the data into memory so the `most_frequent_words` asset can use it.
+They manage **input** by reading an asset from where it’s stored and loading it into memory to be used by a dependent asset. For example, `most_frequent_words` needs `topstories`. Therefore, you could use an I/O manager to load the `topstories` data into memory so that the `most_frequent_words` asset can use it.
 
-I/O managers control **output** by writing the asset's to the location configured. By the end of this section, you’ll configure an I/O manager to write the `hackernews_topstories` DataFrame into a database as a table.
+I/O managers control **output** by writing the asset's to the location configured. By the end of this section, you’ll configure an I/O manager to write the `topstories` DataFrame into a database as a table.
```
sryza, line 25: Not your fault, but could be an opportunity to fix "asset's" -> "assets"

docs/content/tutorial/saving-your-data.mdx
sryza: I would also lead with the general idea that IO managers are very powerful and beneficial for some situations, but are a poor fit for other situations. And if you find yourself getting tripped up trying to write your pipeline to use IO managers, often the easiest way to make progress is to not use them.

docs/content/tutorial/saving-your-data.mdx
sryza: I agree that this would work better at the end of the section.

Worth looking at some of the content here to maybe copy/paste:
- https://docs.dagster.io/guides/dagster/non-argument-deps#when-to-use-assets-without-arguments-and-return-values
- https://docs.dagster.io/tutorial/managing-your-own-io#tutorial-part-7-managing-your-own-io

sryza (commented):
Not necessarily critical right now, but in the future I'd recommend splitting up a change like this into three separate PRs:
- A find/replace for non_argument_deps in docs_snippets
- Updates to the SDA concepts page
- Updates to the tutorial

docs/content/tutorial/saving-your-data.mdx
sryza: I think this works

docs/content/concepts/assets/software-defined-assets.mdx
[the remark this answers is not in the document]
jamiedemaria: yeah that's a good point, i hadn't considered that. 

I guess the question would be like "I've read the tutorial but forgot some syntax, how do i define the dependency where dagster loads the data for me?" - which i guess `managed-loading` would match up with that.

---

To do a little exercise:

I think if I were to write a super verbose heading, I'd end up with something like "defining dependencies where dagster will load the data from the upstream asset into memory for you".

Shortening that - "defining dependencies where dagster manages loading the upstream data" 

could be shorter still, but i'm not finding the right way to do that atm. i think bringing in "dagster manages" in the second version indicates that "managed-loading" might be where this ends up anyway. i'll change it and can always revisit later if needed

docs/content/tutorial/saving-your-data.mdx
```diff
@@ -167,12 +268,25 @@ Despite the changes in where the data is stored, the developer experience hasn
...
+
+```python file=/tutorial/saving/no_io_assets_examples.py startafter=start_snowflake_example endbefore=end_snowflake_example
+@asset(deps=[orders])
+def returns():
+    conn = get_snowflake_connection()
+    conn.execute(
+        "CREATE TABLE returns AS SELECT * from orders WHERE status = 'RETURNED'"
+    )
+```
+
```
erinkcochran87, line 285:
````suggestion
```

---

````

Missing a rule before the next section

File: docs/content/tutorial/saving-your-data.mdx [hunk 21]
```diff
@@ -167,12 +268,25 @@ Despite the changes in where the data is stored, the developer experience hasn
 
 ---
 
+## When should you use I/O managers?
+
+I/O managers are a powerful tool that can simplify your code and allow you to easily modify where your assets are stored. In this tutorial, using an I/O manager made sense because the contents of each asset are used in the downstream assets. For example, you needed to load the `topstory_ids` asset in order to compute the `topstories` asset. Using an I/O manager allowed up to remove the repetitive reading and writing code.
+
+However, some assets do not need an I/O manager. For example, you may write an asset that executes a Snowflake query that creates a new tables based on the data of another table. This asset would depend on the the existing table, but does not need to load that table in memory in order to execute the query:
```
erinkcochran87, line 275:
```suggestion
However, some assets do not need an I/O manager. For example, you may write an asset that executes a Snowflake query that creates a new table based on the data of another table. This asset would depend on the existing table, but does not need to load that table in memory in order to execute the query:
```

Fixes some typos.

Also, should we link out to the docs for `deps` here or nah?

docs/content/concepts/assets/software-defined-assets.mdx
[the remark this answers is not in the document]
sryza: Yeah I guess I'm being a little loose here. When you write "select ...", that does cause data to be loaded, it just happens in the database instead of the process executing your Python code. With Spark, it's even more of a gray area: you actually write `spark.read` inside Python, but the read is usually executed remotely.  There's probably some way of being more precise, it's not coming to me immediately though.

docs/content/tutorial/saving-your-data.mdx
```diff
@@ -167,12 +268,25 @@ Despite the changes in where the data is stored, the developer experience hasn
 
 ---
 
+## When should you use I/O managers?
+
+I/O managers are a powerful tool that can simplify your code and allow you to easily modify where your assets are stored. In this tutorial, using an I/O manager made sense because the contents of each asset are used in the downstream assets. For example, you needed to load the `topstory_ids` asset in order to compute the `topstories` asset. Using an I/O manager allowed up to remove the repetitive reading and writing code.
```
erinkcochran87, line 273:
```suggestion
I/O managers are a powerful tool that can simplify your code and allow you to easily modify where your assets are stored. In this tutorial, using an I/O manager made sense because the contents of each asset were used in the downstream assets. For example, you needed to load the `topstory_ids` asset in order to compute the `topstories` asset. Using an I/O manager allowed you to remove repetitive reading and writing code.
```

Fixes a few typos + grammar tweaks

docs/content/tutorial/saving-your-data.mdx
erinkcochran87: Unfortunately I don't - I honestly think this was very likely an oversight and not intentional.

jamiedemaria:
> in the future I'd recommend splitting up a change like this into three separate PRs

yeah this PR definitely spiraled more than i expected. I don't want to lose the comments, but any other sections that need changes will get their own PRs

docs/content/tutorial/saving-your-data.mdx
erinkcochran87: I do too, especially since you added comments highlighting and explaining the changes

docs/content/tutorial/saving-your-data.mdx
jamiedemaria: did a bit of work on this section and incorporated things from those two docs

docs/content/tutorial/saving-your-data.mdx
```diff
@@ -76,11 +75,110 @@ defs = Definitions(
   information.
 </Note>
 
-To see the effect of this change, reload your code location in the Dagster UI and materialize your assets. When you click on an asset, you’ll see the `path` metadata pointing to the `data` directory on your computer.
+Now, you will modify your three assets `topstory_ids`, `topstories`, and `most_frequent_words` to use the I/O manager to read and write data rather than direct calls to the `json` and `pandas` libraries. To do this, you will:
+
+- Return your data from each asset, rather than manually writing it to the file system. The I/O manager will now save the data.
+- Modify how you define dependencies between assets. To use the I/O manager to load the data associated with an upstream asset into memory, you use the upstream asset's name as the name of one of the arguments to the downstream asset's function.
```
erinkcochran87, line 81: A few thoughts on this bit:

- Should the dependencies bullet come first, since that's what comes first in the code examples?
- Perhaps bold the first sentences of each bullet to make them stand out, ex: "**Modify how you define dependencies between assets.** To use..."
- I think this does a good job of concisely explaining the changes, tbh. What do you feel is missing or falling short? We can chat about it later if Zoomin' is easier too

erinkcochran87 (commented):
The tutorial is looking good! Thank you so much for doing the lord's work on this while Tim is out 🙏

docs/content/tutorial/saving-your-data.mdx
jamiedemaria: re ordering: I've been going back and forth on this, and eventually went with the current order bc chronologically data needs to be saved before it can be loaded. but also the dependency setup happens at the top of the asset and return happens at the bottom so it also makes sense the other way around. idk idk 

re what i thought was falling short: I didn't have any specifics, i just had to rewrite this section several times and didn't have an "ah ha i got it" moment. also had tired friday brain so wasn't confident what i wrote made any sense. After reading it again today i think it isn't so bad

docs/content/concepts/assets/software-defined-assets.mdx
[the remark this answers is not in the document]
sryza: > it's expected that the code you include inside your @asset-decorated function will access the data from the upstream asset

I like this phrasing

> For example, you may run a SELECT query against the upstream asset, or a spark.read that executes remotely. If you need direct access to the contents of the upstream asset in memory, then you are also responsible for loading the contents of the asset within the @asset-decorated function.

"or a spark.read that executes remotely" makes it sound like you wouldn't do it for the case where a spark.read executes locally. I would maybe just present a couple values equally: "For example, you may run a SQL statement that selects from the upstream table, load the upstream CSV file into a Pandas DataFrame".

docs/content/tutorial/saving-your-data.mdx
```diff
@@ -167,12 +266,40 @@ Despite the changes in where the data is stored, the developer experience hasn
 
 ---
 
+## When should you use I/O managers?
```
yuhan, line 269: i love this section. thank you!

yuhan (approved):

docs/content/concepts/assets/software-defined-assets.mdx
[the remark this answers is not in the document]
jamiedemaria: what if we left it more generic and said "for example, you may want to fetch the data into memory, or you may want to execute a command that works with the data in an external process"

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -70,12 +70,33 @@ By default, the name of the decorated function, `my_asset`, is used as the asset
...
+
+
+@asset(deps=[sugary_cereals])
+def shopping_list() -> None:
+    execute_query("CREATE TABLE shopping_list AS SELECT * FROM sugary_cereals")
+```
+
+#### Defining basic managed-loading dependencies
+
+When using basic dependencies, as above, it's expected that if you need direct access to the contents of the asset, the code you include inside your `@asset`-decorated function will load the data from the upstream asset. Dagster alternatively allows you to delegate loading data to an [I/O manager](/concepts/io-management/io-managers). To do this, you express the dependency by using the upstream asset name as the name of one of the arguments on the decorated function.
```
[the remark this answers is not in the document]
jamiedemaria: this is in the concepts section though so i think linking to other concepts pages makes sense. I do think there are a couple places in the tutorial where we make this link though, i'll fix those

docs/content/tutorial/saving-your-data.mdx
erinkcochran87: The ordering makes sense; thanks for explaining it. I think this is fine as-is, and agreed that the wording in here is also fine. Let's plan to revisit if we get feedback that something needs clarification.

erinkcochran87 (approved):

docs/content/_navigation.json
```diff
@@ -57,11 +57,7 @@
         "path": "/tutorial/saving-your-data"
```
jamiedemaria, line 57: rename to Using Dagster to save your data

PR #15066, opened 2023-06-30
DbtCliTask -> DbtCliInvocation

sryza:
## Summary & Motivation

This PR renames `DbtCliTask` to `DbtCliInvocation`.  The thinking here is that "task" is a loaded term in orchestration, and it often refers to "something that can be executed" rather than "the result of a particular invocation".

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/__init__.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/__init__.py
@@ -50,7 +50,7 @@ from .asset_decorator import dbt_assets as dbt_assets
 from .cli import (
     DbtCli as DbtCli,
     DbtCliEventMessage as DbtCliEventMessage,
-    DbtCliTask as DbtCliTask,
+    DbtCliInvocation as DbtCliInvocation,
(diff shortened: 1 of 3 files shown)


rexledesma:
I agree that `Task` is an overloaded term. In fact, I picked `Task` for its [connotations in async/await](https://docs.python.org/3/library/asyncio-future.html), since we don't know the contents of the dbt CLI execution until we `.wait()` or `.stream()` the execution.

I don't particularly like `Result` since it implies that the underlying execution is already finished once the object has been instantiated.

sryza:
@rexledesma perhaps `DagsterCliInvocation`? I wasn't familiar with asyncio tasks, but I agree that this usage does match it.  However, I think that people encountering this API are more likely to be familiar with the Airflow notion of a task than the Python asyncio notion of a task.

rexledesma:
@sryza `DbtCliInvocation` sounds good to me.

sryza:
> @sryza DbtCliInvocation sounds good to me.

Cool - respun with this name.

rexledesma (approved):

PR #15067, opened 2023-06-30
DbtCli.global_config -> global_config_flags

sryza:
## Summary & Motivation

I'm aiming to remove the experimental label from DbtCli, which will allow me to start switching over the tutorial. As a precursor to this, I did a final review of the API. The only thing that came up was that I found the name `global_config` to be a tiny bit confusing - I thought it make accept a dictionary of config key value pairs, when in fact it accepts a list of command line flags.  If I understand correctly from the [dbt documentation](https://docs.getdbt.com/reference/global-configs/command-line-flags), flags are one way to set dbt global config, and this particular API sets flags.

### DIFF
--- a/examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/__init__.py
+++ b/examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/__init__.py
@@ -13,7 +13,7 @@ defs = Definitions(
     resources={
         "dbt": DbtCli(
             project_dir=DBT_PROJECT_DIR,
-            global_config=["--no-use-colors"],
+            global_config_flags=["--no-use-colors"],
             profile="jaffle_shop",
             target="dev",
         ),
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
@@ -887,7 +887,7 @@ class DbtCli(ConfigurableResource):
         project_dir (Path): The path to the dbt project directory. This directory should contain a
             `dbt_project.yml`. See https://docs.getdbt.com/reference/dbt_project.yml for more
             information.
-        global_config (List[str]): A list of global flags configuration to pass to the dbt CLI
+        global_config_flags (List[str]): A list of global flags configuration to pass to the dbt CLI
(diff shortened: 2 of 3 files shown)


python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -887,7 +887,7 @@ class DbtCli(ConfigurableResource):
         project_dir (Path): The path to the dbt project directory. This directory should contain a
             `dbt_project.yml`. See https://docs.getdbt.com/reference/dbt_project.yml for more
             information.
-        global_config (List[str]): A list of global flags configuration to pass to the dbt CLI
+        global_flags (List[str]): A list of global flags configuration to pass to the dbt CLI
```
rexledesma, line 890:
```suggestion
        global_flags (List[str]): A list of CLI flags to edit the global configuration of the dbt CLI
```

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources_v2.py
```diff
@@ -918,7 +918,7 @@ class DbtCli(ConfigurableResource):
             " information."
         ),
     )
-    global_config: List[str] = Field(
+    global_flags: List[str] = Field(
         default=[],
         description=(
             "A list of global flags configuration to pass to the dbt CLI invocation. See"
```
rexledesma, line 924:
```suggestion
            "A list of CLI flags to edit the global configuration of the dbt CLI invocation. See"
```

examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/__init__.py
```diff
@@ -13,7 +13,7 @@
     resources={
         "dbt": DbtCli(
             project_dir=DBT_PROJECT_DIR,
-            global_config=["--no-use-colors"],
+            global_flags=["--no-use-colors"],
```
rexledesma, line 16: Perhaps `global_config_flags`?
```suggestion
            global_config_flags=["--no-use-colors"],
```

rexledesma (approved):
> flags are one way to set dbt global config, and this particular API sets flags.

Yeah, this framing makes sense. I believe `global_config_flags` would make this more precise.

sryza:
> I believe global_config_flags would make this more precise.

Makes sense. I will change.

PR #15068, opened 2023-06-30
unexperimentalize the DbtCli resource and deprecate DbtCliResource

sryza:
## Summary & Motivation

I'm aiming to remove the experimental label from DbtCli, which will allow me to start switching over the tutorial. 

Depends on https://github.com/dagster-io/dagster/pull/15067 and https://github.com/dagster-io/dagster/pull/15066.

### DIFF
--- a/docs/sphinx/sections/api/apidocs/libraries/dagster-dbt.rst
+++ b/docs/sphinx/sections/api/apidocs/libraries/dagster-dbt.rst
@@ -51,8 +51,13 @@ more advanced use cases, we suggest building your own ops which directly interac
 Resources (dbt Core)
 ====================
 
-CLI Resources
--------------
+CLI Resource
+------------
+
+.. autoclass:: DbtCli
+
+Deprecated CLI Resource
+-----------------------
 
 .. autoclass:: DbtCliResource
 
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/core/resources.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/core/resources.py
@@ -2,9 +2,10 @@ from typing import Any, Iterator, Mapping, Optional, Sequence, Set
 
 import dagster._check as check
 from dagster import resource
-from dagster._annotations import public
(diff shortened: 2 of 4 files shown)


rexledesma (approved):
I believe this depends on https://github.com/dagster-io/dagster/pull/15047 as well.

`DbtCli.cli(...)` requires a `DbtManifest`, which is still experimental. So I think the only non-experimental usage for this resource right now is for the existing `load_assets_from_XXX` APIs.

sryza:
> I believe this depends on https://github.com/dagster-io/dagster/pull/15047 as well.

Ah yes, definitely

sryza:
@rexledesma I updated this PR to also deprecate DbtCliResource. Mind taking another look?

python_modules/libraries/dagster-dbt/dagster_dbt/cli/types.py
```diff
@@ -1,14 +1,19 @@
...
 
 from ..types import DbtOutput
 
 
+@deprecated
 class DbtCliOutput(DbtOutput):
     """The results of executing a dbt command, along with additional metadata about the dbt CLI
     process that was run.
 
+    This class it's deprecated, because it's only produced by methods of the DbtCliResource class,
```
rexledesma, line 14:
```suggestion
    This class is deprecated because it's only produced by methods of the DbtCliResource class,
```

python_modules/libraries/dagster-dbt/dagster_dbt/cli/resources.py
```diff
@@ -468,7 +468,11 @@ def get_manifest_json(self, **kwargs) -> Optional[Mapping[str, Any]]:
         return parse_manifest(project_dir, target_path)
 
 
+@deprecated
```
rexledesma, line 471: Let's make this emit `deprecation_warning` as well? I don't think `@deprecated` actually does anything.

rexledesma (approved):

erinkcochran87 (approved):

PR #15069, opened 2023-06-30
[pythonic resources] Error if extra params passed to Config or ConfigurableResources

benpankow:
## Summary

Introduces the pydantic extra param "forbidden" flag and ensures that extra values are passed through to Pydantic so an error is thrown in the case that extra params are passed to the constructor of a Config or ConfigurableResouce class

## Test Plan

Unit test.

### DIFF
--- a/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
+++ b/python_modules/dagster/dagster/_config/pythonic_config/__init__.py
@@ -125,6 +125,7 @@ class MakeConfigCacheable(BaseModel):
         keep_untouched = (cached_property,)
         # Ensure the class is serializable, for caching purposes
         frozen = True
+        extra = "forbid"
 
     def __setattr__(self, name: str, value: Any):
         # This is a hack to allow us to set attributes on the class that are not part of the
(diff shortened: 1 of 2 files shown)


benpankow:
* master
  * **PR #15069** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15069" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15069?utm_source=stack-comment).

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
```diff
@@ -1795,7 +1799,7 @@ def separate_resource_params(cls: Type[BaseModel], data: Dict[str, Any]) -> Sepa
     """
     keys_by_alias = {field.alias: field for field in cls.__fields__.values()}
     data_with_annotation: List[Tuple[str, Any, Type]] = [
-        (k, v, keys_by_alias[k].annotation) for k, v in data.items() if k in keys_by_alias
+        (k, v, keys_by_alias[k].annotation if k in keys_by_alias else None) for k, v in data.items()
```
jamiedemaria, line 1802: is this related to the extra params fix?

jamiedemaria (approved):
cool

python_modules/dagster/dagster/_config/pythonic_config/__init__.py
benpankow: Yeah - this would actually filter out the extra params (because of the if clause) before it was being passed to Pydantic. This would mean the user-facing error wouldn't show up (it would just silently remove the extra params).

jamiedemaria: ok cool - i just couldn't figure out where it fit in with the rest of the changes, but that makes sense!

schrockn (changes requested):
Isn't this a breaking change?

Issue #14564, opened 2023-06-01
Allow any valid Pydantic field type to be used in Config


ndellosa95:
### What's the use case?

Currently Dagster only allows primitive, collection, `Config` or discriminated union types in Pydantic configs. However, Pydantic provides a bunch of field types aside from these, as well as the ability to [define a custom field type](https://docs.pydantic.dev/latest/usage/types/#custom-data-types). Ideally anything that is valid in Pydantic should be valid for dagster configs. 

Alternatively, there should be a dagster specific way to create custom field types from the currently allowed field types. In dagster 1.3.6, this is like half-implemented if you subclass `Config` and yield a validator in `__get_validators__` that accepts both Mapping and non-Mapping input (say a string as an example) you can pass the value of the field in your preferred format directly to the `Config` / `ConfiguredResource` class. However, the subclass still needs to accept a Mapping input and if specifying the config via run config, it needs to be in the mapping format.

### Ideas of implementation

I'm not sure what internals (if any) make the primitive, collection, `Config` or discriminated union types restriction necessary, but ideally Dagster should just offload the validation of config fields off to Pydantic. Anything that Pydantic accepts Dagster should also accept.

### Message from the maintainers

Impacted by this issue? Give it a 👍! We factor engagement into prioritization.

ndellosa95:
@benpankow I hacked your type system 😅 

By creating a custom metaclass that inherits from `ConfigScalar` for my custom datatype and overriding `_get_non_none_public_field_values` in my resource, I was able to get my custom datatype to serialize and validate as a string but it appears as my custom type in the resource object itself. 

So the simplest implementation of this might be to just have an abc to setup a custom type that just needs to know how to serialize the class in a config dict and we can rely on Pydantic to validate that the input is correct.

mjclarke94:
We ran in to this recently. @pzarabadip, am I remembering correctly that `datetime.timedelta` instances weren't working? (Sorry for the ping, not at my laptop so can't check!)

pzarabadip:
Hey @ndellosa95. As @mjclarke94 mentioned, we've got similar issue with `timedelta`. Following code snipped results in validation failure:
```
from dagster import asset, Config, OpExecutionContext
from datetime import timedelta

class MyConfig(Config):
    input: timedelta


@asset
def test_timedelta(context: OpExecutionContext, config: MyConfig):
    return context.log.info("test")
```
with following (trimmed) log:
```
[...]

PR #15071, opened 2023-06-30
feat(dbt): add Dagster code location scaffold for dbt project

rexledesma:
## Summary & Motivation
The initial scaffold for a `dagster-dbt` project. In later PRs in the stack, we'll parameterize this directory to be configurable via CLI settings in `dagster-dbt project scaffold`.

As a baseline, we'll put all the definitions in one file.

## How I Tested These Changes
N/A

### DIFF
--- a/python_modules/libraries/dagster-dbt/MANIFEST.in
+++ b/python_modules/libraries/dagster-dbt/MANIFEST.in
@@ -1,2 +1,3 @@
 include LICENSE
 include dagster_dbt/py.typed
+recursive-include dagster_dbt/include *
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/include/pyproject.toml.jinja
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/include/pyproject.toml.jinja
@@ -1,0 +1,6 @@
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[tool.dagster]
+module_name = "scaffold.definitions"
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/include/scaffold/definitions.py.jinja
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/include/scaffold/definitions.py.jinja
@@ -1,0 +1,32 @@
+from pathlib import Path
+
+from dagster import Definitions
+
+from dagster_dbt import DbtCli, DbtManifest, dbt_assets
+
+dbt_project_dir = Path(__file__).parent.joinpath("..", "jaffle_shop")
+dbt_manifest_path = dbt_project_dir.joinpath("target", "manifest.json")
+dbt_manifest = DbtManifest.read(path=dbt_manifest_path)
+
+
+@dbt_assets(manifest=dbt_manifest)
+def build_dbt_project(dbt: DbtCli):
+    yield from dbt.cli(["build"], manifest=dbt_manifest).stream()
+
+
+schedules = [
+    dbt_manifest.build_schedule(
+        job_name="materialize_dbt_models",
+        cron_schedule="0 0 * * *",
+    )
+]
+
+defs = Definitions(
+    assets=[build_dbt_project],
+    schedules=schedules,
+    resources={
+        "dbt": DbtCli(
+            project_dir=dbt_project_dir.as_posix(),
+        ),
+    },
+)
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/include/setup.py.jinja
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/include/setup.py.jinja
@@ -1,0 +1,18 @@
+from setuptools import find_packages, setup
+
+setup(
+    name="dbt_scaffold",
+    version="0.0.1",
+    packages=find_packages(),
+    install_requires=[
+        "dagster",
+        "dagster-cloud",
+        "dagster-dbt",
+        "dbt-core>=1.4.0",
+    ],
+    extras_require={
+        "dev": [
+            "dagster-webserver",
+        ]
+    },
+)


rexledesma:
* master
  * **PR #15071** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15071" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15071?utm_source=stack-comment).

python_modules/libraries/dagster-dbt/dagster_dbt/include/dbt_scaffold/definitions.py
```diff
@@ -0,0 +1,32 @@
...
+        cron_schedule="0 0 0 * *",
+    )
+]
+
+defs = Definitions(
+    assets=[my_dbt_assets],
+    schedules=schedules,
+    resources={
+        "dbt": DbtCli(
+            project_dir=DBT_PROJECT_DIR.as_posix(),
```
sryza, line 29: Is there some way we can avoid the need for `as_posix` here?

python_modules/libraries/dagster-dbt/dagster_dbt/include/dbt_scaffold/definitions.py
```diff
@@ -0,0 +1,32 @@
...
+DBT_MANIFEST_PATH = DBT_PROJECT_DIR.joinpath("target", "manifest.json")
+DBT_MANIFEST = DbtManifest.read(path=DBT_MANIFEST_PATH)
+
+
+@dbt_assets(manifest=DBT_MANIFEST)
+def my_dbt_assets(dbt: DbtCli):
+    yield from dbt.cli(["build"], manifest=DBT_MANIFEST).stream()
+
+
+schedules = [
```
sryza, line 17: My takeaway from our earlier conversations had been that we would put this in a separate file, so that users don't need to be confronted with all this other intimidating stuff in order to mess with their basic schedules.  It has some downsides in terms of having imports between files, but my intuition is that it's worth it to make things more friendly.

python_modules/libraries/dagster-dbt/dagster_dbt/include/dbt_scaffold/definitions.py
```diff
@@ -0,0 +1,32 @@
...
+
+from dagster_dbt import DbtCli, DbtManifest, dbt_assets
+
+DBT_PROJECT_DIR = Path(__file__).parent.joinpath("..", "jaffle_shop")
+DBT_MANIFEST_PATH = DBT_PROJECT_DIR.joinpath("target", "manifest.json")
+DBT_MANIFEST = DbtManifest.read(path=DBT_MANIFEST_PATH)
+
+
+@dbt_assets(manifest=DBT_MANIFEST)
+def my_dbt_assets(dbt: DbtCli):
```
sryza, line 13: Is `my_dbt_assets` the name that we would expect people to use in a "realistic" project setup? If not, is there a different one?

python_modules/libraries/dagster-dbt/dagster_dbt/include/dbt_scaffold/definitions.py
```diff
@@ -0,0 +1,32 @@
...
+from dagster import Definitions
+
+from dagster_dbt import DbtCli, DbtManifest, dbt_assets
+
+DBT_PROJECT_DIR = Path(__file__).parent.joinpath("..", "jaffle_shop")
+DBT_MANIFEST_PATH = DBT_PROJECT_DIR.joinpath("target", "manifest.json")
+DBT_MANIFEST = DbtManifest.read(path=DBT_MANIFEST_PATH)
+
+
+@dbt_assets(manifest=DBT_MANIFEST)
```
sryza, line 12: Our current dbt quickstarts use `load_assets_from_dbt_project`, which enables an easy local development flow. Is there some way we can preserve that?

python_modules/libraries/dagster-dbt/dagster_dbt/include/setup.py
```diff
@@ -0,0 +1,14 @@
...
+
+setup(
+    name="dbt_scaffold",
+    packages=find_packages(),
+    install_requires=[
+        "dagster",
+        "dagster-cloud",
+        "dagster-dbt",
+        "dbt-core>=1.4.0",
+        "dbt-duckdb",
```
sryza, line 11: Not everyone will be using duckdb, so probably best to leave this out?

python_modules/libraries/dagster-dbt/dagster_dbt/include/dbt_scaffold/definitions.py
```diff
@@ -0,0 +1,32 @@
+from pathlib import Path
+
+from dagster import Definitions
+
+from dagster_dbt import DbtCli, DbtManifest, dbt_assets
+
+DBT_PROJECT_DIR = Path(__file__).parent.joinpath("..", "jaffle_shop")
+DBT_MANIFEST_PATH = DBT_PROJECT_DIR.joinpath("target", "manifest.json")
+DBT_MANIFEST = DbtManifest.read(path=DBT_MANIFEST_PATH)
```
sryza, line 9: Uppercasing this while not uppercasing `my_dbt_assets`, `schedules`, and `defs` feels weird. My instinct would be to make it lowercase, even if that means `DBT_PROJECT_DIR` and `DBT_MANIFEST_PATH` need to be lowercase as well for consistency (though could go either way on those two - feels different because they're strings I think). Thoughts?

python_modules/libraries/dagster-dbt/dagster_dbt/include/dbt_scaffold/definitions.py
```diff
@@ -0,0 +1,32 @@
+from pathlib import Path
+
+from dagster import Definitions
+
+from dagster_dbt import DbtCli, DbtManifest, dbt_assets
+
+DBT_PROJECT_DIR = Path(__file__).parent.joinpath("..", "jaffle_shop")
```
sryza, line 7: This `__file__` business will likely be black magic to many readers. I think it's worth including a comment here to explain what's going on and the assumptions we're making.

sryza (commented):
Very excited to see this

python_modules/libraries/dagster-dbt/dagster_dbt/include/dbt_scaffold/definitions.py
rexledesma: Your takeaway is right.

However, I've put everything in one file for this PR to have a baseline, similar to what I did for the `.build_schedule` API (https://github.com/dagster-io/dagster/pull/14992 + https://github.com/dagster-io/dagster/pull/14997).

Like last time, I'll put the changes in folder structure in another PR, so we can visually see the changes in comparison to this single-file option.

python_modules/libraries/dagster-dbt/dagster_dbt/include/dbt_scaffold/definitions.py
rexledesma: The name of the decorated function here corresponds to the underlying op definition. I assume that we're trying to name that accurately.

From the user's perspective, this corresponds to [running a dbt project](https://docs.getdbt.com/docs/running-a-dbt-project/run-your-dbt-projects). So perhaps:

- `build_dbt_project`
- `run_dbt_project`

sryza: For `@asset`- and `@multi_asset`- decorated functions, the name applies to the op, but it also the variable name of the `AssetsDefinition` object.  So our general convention is for the name to be a name that's appropriate for a reference to that object, i.e. a noun that describes that collection of assets.

I think it's worth following that convention here.  It does result in a weird name for the op, but no weirder than for other non-dbt ops that are used in similar contexts.

python_modules/libraries/dagster-dbt/dagster_dbt/include/dbt_scaffold/definitions.py
sryza: Got it - sounds good

python_modules/libraries/dagster-dbt/dagster_dbt/include/dbt_scaffold/definitions.py
rexledesma: Because we are using `pathlib`, this is not a primitive type that can be expressed in our Dagster config system. So currently, this is blocked by https://github.com/dagster-io/dagster/issues/14564.

If we want to continue using this as a string, then we need to use `file_relative_path`. I avoided this since I want to deprecate this function (https://github.com/dagster-io/dagster/pull/14499), but we can use this if we want.

python_modules/libraries/dagster-dbt/dagster_dbt/include/dbt_scaffold/definitions.py
rexledesma: We could use the name of the dbt project in which we are scaffolding. We can get this name from `dbt_project.yml`. For example, the jaffle shop project has the name `jaffle_shop`.

python_modules/libraries/dagster-dbt/dagster_dbt/include/dbt_scaffold/definitions.py
rexledesma: We could run a subprocess in this file to recreate manifest using `dbt parse`.

python_modules/libraries/dagster-dbt/dagster_dbt/include/dbt_scaffold/definitions.py
rexledesma: We can't avoid this in the `file_relative_path` case or `pathlib` case. I was planning on describing this in the CLI invocation, but we can add a comment in the code itself as well.

python_modules/libraries/dagster-dbt/dagster_dbt/include/dbt_scaffold/definitions.py
sryza: That makes sense to me. Something like `all_dbt_assets` could be fine too if easier.

python_modules/libraries/dagster-dbt/dagster_dbt/include/dbt_scaffold/definitions.py
sryza: Yeah, I agree this isn't something that was any clearer with `file_relative_path`.

python_modules/libraries/dagster-dbt/dagster_dbt/include/dbt_scaffold/definitions.py
sryza: Got it - ok not the end of the world for now to use `as_posix`

sryza (approved):

PR #15077, opened 2023-07-01
Render @deprecated in API docs

### LINKED ISSUE #15003: `@deprecated` doesn't render on properties in the API docs
### What's the issue or suggestion?

Making this render a deprecated badge should require a small tweak to `autodoc_dagster`.

### Additional information

_No response_

smackesey:
## Summary & Motivation

Closes #15003 

Throughout our code we use `@deprecated` as a decorator on deprecated classes/methods. This was not being rendered in our API docs.

This PR extends our custom sphinx extension to render `@deprecated` as a special badge at the top of the API doc entry. It works as follows:

- Use a docstring processing hook to inject `.. deprecated:: <placeholder>` at the start of the docstring for any object having a `@deprecated` tag. This is a builtin sphinx deprecated directive that is ultimately rendered like this:

```
<div class="deprecated">
  <p>
    <span class="deprecated">
      Deprecated in version <placeholder>.
    </span>
  </p>
</div>
```

- Because there are a number of existing `@deprecated` in the codebase and it  is not clear at what version the target was deprecated, I used a placeholder  here that we remove later. The placeholder is necessary because `.. deprecated::` requires an argument.
- After the sphinx doctree is built, look up all `versionmodified` nodes (the  docutils node type corresponding to sphinx `deprecated` directive) and replace the contained text with just `Deprecated`.
- The result in the final HTML:

```
<div class="deprecated">
  <p>
    <span class="deprecated">
      Deprecated
    </span>
  </p>
</div>
```

I added CSS to the NextJS site for `span.deprecated` to style this as a badge. I am sure it doesn't quite match what we want here, so cc @bengotow for CSS tweaks:

image

## How I Tested These Changes

Rendered the docs.


smackesey:
* master
  * **PR #15077** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15077" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15077?utm_source=stack-comment).

sryza (commented):
I like the idea and think it looks good visually. Will defer to Ben on the CSS.

erinkcochran87 (commented):
This is great! Also deferring to Ben on CSS.

bengotow (approved):
This looks good to me!

PR #15081, opened 2023-07-02
Update run_request.py tags validation

### LINKED ISSUE #15080: Tags require 'str' value in RunRequest but 'Any' in the JobDefinition
### Dagster version

1.2.6

### What's the issue?

In order to define k8s specs we have to give the run a tag with a Dict value.
This goes fine in the JobDefinition because the validation there is: 
`tags = check.opt_mapping_param(tags, "tags", key_type=str)`

But when we try to set it in a RunRequest we fail with this error:
`dagster._check.CheckError: Value in Mapping mismatches expected type for key dagster-k8s/config. Expected value of type <class 'str'>. Got value {'container_config': {'resources': {'limits': {'cpu': '42', 'memory': '243G'}, 'requests': {'cpu': '42', 'memory': '243G'}}}} of type <class 'dict'>.`

Becuase the validation is of such:
`tags=check.opt_mapping_param(tags, "tags", key_type=str, value_type=str),`

I expected to be able to give a tag that does not have a 'str' as a value to be able to set k8s specs for my run.

### How to reproduce?

Create a RunRequest with a String: Dict in the tags

### Deployment type

Dagster Helm chart

### Deployment details

irrelevant

### Additional information

-

yuvalgimmunai:
## Summary & Motivation

Fixes #15080 

## How I Tested These Changes

Triggered a RunRequest from sensor

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/run_request.py
+++ b/python_modules/dagster/dagster/_core/definitions/run_request.py
@@ -150,9 +150,7 @@ class RunRequest(
             run_config=check.opt_mapping_param(
                 convert_config_input(run_config), "run_config", key_type=str
             ),
-            tags=validate_tags(
-                check.opt_mapping_param(tags, "tags", key_type=str), allow_reserved_tags=False
-            ),
+            tags=validate_tags(check.opt_mapping_param(tags, "tags", key_type=str)),
             job_name=check.opt_str_param(job_name, "job_name"),
             asset_selection=check.opt_nullable_sequence_param(
                 asset_selection, "asset_selection", of_type=AssetKey
--- a/python_modules/dagster/dagster/_core/definitions/run_request.py
+++ b/python_modules/dagster/dagster/_core/definitions/run_request.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING, Any, Mapping, NamedTuple, Optional, Sequence, 
 import dagster._check as check
 from dagster._annotations import PublicAttr, experimental
 from dagster._core.definitions.events import AssetKey
+from dagster._core.definitions.utils import validate_tags
 from dagster._core.instance import DynamicPartitionsStore
 from dagster._core.storage.dagster_run import DagsterRun, DagsterRunStatus
 from dagster._core.storage.tags import PARTITION_NAME_TAG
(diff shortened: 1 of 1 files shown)


jamiedemaria:
Thanks for the contribution! cc @sryza  and @smackesey  do you mind taking a look? This seems like a sensical change to me, but I'm not sure if there are broader implications

python_modules/dagster/dagster/_core/definitions/run_request.py
```diff
@@ -149,7 +149,7 @@ def __new__(
             run_config=check.opt_mapping_param(
                 convert_config_input(run_config), "run_config", key_type=str
             ),
-            tags=check.opt_mapping_param(tags, "tags", key_type=str, value_type=str),
+            tags=check.opt_mapping_param(tags, "tags", key_type=str),
```
smackesey, line 152: This needs to be wrapped in a call to `validate_tags` (from `dagster._core.definitions.utils`:

```
tags=validate_tags(check.opt_mapping_param(tags, "tags", key_type=str), allow_reserved=False)
```

Also the `tags` type annotation in the `__new__` signature should be changed to `Mapping[str, Any]`

smackesey (changes requested):
Thanks @danielgafni. If you can address my comment I'll run CI on this and then we should be good to go.

yuvalgimmunai:
Hey guys thanks for the quick response, please review the new commit

python_modules/dagster/dagster/_core/definitions/run_request.py
```diff
@@ -100,7 +101,7 @@ class RunRequest(
         [
             ("run_key", PublicAttr[Optional[str]]),
             ("run_config", PublicAttr[Mapping[str, Any]]),
-            ("tags", PublicAttr[Mapping[str, str]]),
+            ("tags", PublicAttr[Mapping[str, Any]]),
```
smackesey, line 104: This one should still be `str` since validate tags normalizes everything to strings (your change to the `__new__` signature is correct though).

smackesey (changes requested):

python_modules/dagster/dagster/_core/definitions/run_request.py
yuvalgimmunai: The perks of trying to make python into a strictly typed lang 🙃

smackesey:
Two things-- the keyword is actually `allow_reserved_tags`, not `allow_reserved` (sorry).

You need to run `make black` on this to pass CI.

smackesey:
Hmm ok so according to CI, this pathway is actually being used internally and is now failing because of `allow_reserved_tags=False` (my bad). Pls just remove the `allow_reserved_tags` keyword arg altogether (it will default to True) and make sure you run black again in case it changes the formatting. Almost there...

smackesey (approved):
@yuvalgimmunai we made it! thanks, merging

PR #15082, opened 2023-07-02
Support partition_key_range for build_op_context, build_asset_context

### LINKED ISSUE #14990: Testing with asset partition ranges
### What's the use case?

Currently we can test a software defined asset that makes use of partition ranges, by creating an op context:

`context = dagster.build_op_context(partition_key="2023-01-05")`

And then we can do `time_window = context.asset_partitions_time_window_for_output()` and it will work for the single day (because we can derive a time window from the key).

However, there are some issues:
- `asset_partitions_time_window_for_output()` is marked as deprecated on `OpExecutionContext` (but not `AssetExecutionContext`) - possibly related to https://github.com/dagster-io/dagster/issues/14871 and a missing `build_asset_context` function?
- testing like this does not work with `context.partition_time_window` (` AttributeError: 'BoundOpExecutionContext' object has no attribute '_step_execution_context'. Did you mean: 'get_step_execution_context'?`) 
- Most importantly, we cannot test multiple days with this approach

### Ideas of implementation

Being able to specify a date/time range in the `build_op_context` would be ideal.  This might even need to be part of the upcoming `build_asset_context` (see  https://github.com/dagster-io/dagster/issues/14871 again).

### Additional information

_No response_

smackesey:
## Summary & Motivation

Fixes #14990.

- Adds `OpExecutionContext.partition_key_range` and deprecates `OpExecutionContext.asset_partition_key_range`
- Fix docstring of `OpExecutionContext.asset_partition_key_range`
- Allows passing `partition_key_range` directly to `build_op_context` and `build_asset_context` (which sets it directly on `UnboundExecutionContext`)-- this allows testing of ops with `partition_key_range`.

I can revert the rename of `asset_partition_key_range` if necessary, but IMO it should be done for consistency with the other partition key methods (`.partition_key` and `.partition_time_window`).

## How I Tested These Changes

Add new unit test.

### DIFF
--- a/python_modules/dagster/dagster/_core/execution/context/compute.py
+++ b/python_modules/dagster/dagster/_core/execution/context/compute.py
@@ -263,12 +263,24 @@ class OpExecutionContext(AbstractComputeExecutionContext):
         """
         return self._step_execution_context.partition_key
 
+    @deprecated
     @public
     @property
     def asset_partition_key_range(self) -> PartitionKeyRange:
-        """The asset partition key for the current run.
+        """The range of partition keys for the current run. (DEPRECATED: Use `partition_key_range` instead).
 
-        Raises an error if the current run is not a partitioned run.
+        If run is for a single partition key, return a `PartitionKeyRange` with the same start and
+        end. Raises an error if the current run is not a partitioned run.
+        """
+        return self.partition_key_range
+
+    @public
+    @property
+    def partition_key_range(self) -> PartitionKeyRange:
+        """The range of partition keys for the current run.
+
+        If run is for a single partition key, return a `PartitionKeyRange` with the same start and
+        end. Raises an error if the current run is not a partitioned run.
         """
         return self._step_execution_context.asset_partition_key_range
 
--- a/python_modules/dagster/dagster/_core/execution/context/invocation.py
+++ b/python_modules/dagster/dagster/_core/execution/context/invocation.py
@@ -27,6 +27,7 @@ from dagster._core.definitions.hook_definition import HookDefinition
 from dagster._core.definitions.job_definition import JobDefinition
 from dagster._core.definitions.multi_dimensional_partitions import MultiPartitionsDefinition
 from dagster._core.definitions.op_definition import OpDefinition
+from dagster._core.definitions.partition_key_range import PartitionKeyRange
 from dagster._core.definitions.resource_definition import (
     IContainsGenerator,
     ResourceDefinition,
@@ -75,6 +76,7 @@ class UnboundOpExecutionContext(OpExecutionContext):
         resources_config: Mapping[str, Any],
         instance: Optional[DagsterInstance],
         partition_key: Optional[str],
+        partition_key_range: Optional[PartitionKeyRange],
         mapping_key: Optional[str],
         assets_def: Optional[AssetsDefinition],
     ):
@@ -108,7 +110,12 @@ class UnboundOpExecutionContext(OpExecutionContext):
         self._log = initialize_console_manager(None)
         self._pdb: Optional[ForkedPdb] = None
         self._cm_scope_entered = False
+        check.invariant(
+            not (partition_key and partition_key_range),
+            "Must supply at most one of partition_key or partition_key_range",
+        )
         self._partition_key = partition_key
+        self._partition_key_range = partition_key_range
         self._user_events: List[UserEvent] = []
         self._output_metadata: Dict[str, Any] = {}
 
@@ -223,6 +230,20 @@ class UnboundOpExecutionContext(OpExecutionContext):
             return self._partition_key
         check.failed("Tried to access partition_key for a non-partitioned run")
 
+    @property
+    def partition_key_range(self) -> PartitionKeyRange:
+        """The range of partition keys for the current run.
+
+        If run is for a single partition key, return a `PartitionKeyRange` with the same start and
+        end. Raises an error if the current run is not a partitioned run.
+        """
+        if self._partition_key_range:
+            return self._partition_key_range
+        elif self._partition_key:
+            return PartitionKeyRange(self._partition_key, self._partition_key)
+        else:
+            check.failed("Tried to access partition_key range for a non-partitioned run")
+
     def asset_partition_key_for_output(self, output_name: str = "result") -> str:
         return self.partition_key
 
@@ -272,6 +293,7 @@ class UnboundOpExecutionContext(OpExecutionContext):
             output_metadata=self._output_metadata,
             mapping_key=self._mapping_key,
             partition_key=self._partition_key,
+            partition_key_range=self._partition_key_range,
             assets_def=self._assets_def,
         )
 
@@ -334,6 +356,7 @@ class UnboundOpExecutionContext(OpExecutionContext):
             resources_config=self._resources_config,
             instance=self._instance,
             partition_key=self._partition_key,
+            partition_key_range=self._partition_key_range,
             mapping_key=self._mapping_key,
             assets_def=self._assets_def,
         )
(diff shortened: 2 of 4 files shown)


smackesey:
* master
  * **PR #15082** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15082" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15082?utm_source=stack-comment).

sryza:
Adding @alangenfeld because I think we're undeprecating `asset_partition_key_range` on `OpExecutionContext`.

smackesey:
>I think we're undeprecating asset_partition_key_range on OpExecutionContext.

Confused by this-- it's not currently deprecated, this PR deprecates it.

sryza:
@smackesey ah I misunderstood, I was thinking this was related to Alex's work that deprecated all the asset-related properties on `OpExecutionContext`.

The reason I used "asset_" in the first place is that we don't currently support, or have the intention of supporting partition key ranges for non-asset runs.

The tags that set this have "asset" in their name:
```python
ASSET_PARTITION_RANGE_START_TAG = "{prefix}asset_partition_range_start".format(
    prefix=SYSTEM_TAG_PREFIX
)

ASSET_PARTITION_RANGE_END_TAG = f"{SYSTEM_TAG_PREFIX}asset_partition_range_end"
```

On the other hand, brevity is nice. What's your thinking here?

smackesey:
>What's your thinking here?

Strong opinion that leaving off `asset_` is definitely better, because (even in the context of asset materialization) you don't access the partition key with `asset_partition_key` or time window with `asset_partition_time_window`. IMO this consistency trumps any added information from the "asset_" prefix.

> we don't currently support, or have the intention of supporting partition key ranges for non-asset runs.

I've barely thought about it but my intuition says why not? Is there some extra complexity I'm missing? Partition ranges for non-asset runs seem like a perfectly coherent idea that as a naive user I would expect to work-- nothing about partition ranges belongs conceptually to the asset layer, so it is just extra arbitrary complexity that it doesn't work for ops. And it doesn't seem like much extra work since it's the same machinery used for assets anyway.

sryza:
> Strong opinion that leaving off asset_ is better, because (even in the context of asset materialization) you don't access the partition key with asset_partition_key or time window with asset_partition_time_window. IMO this consistency trumps any added information from the "asset_" prefix.

Most of our code examples use `asset_partition_time_window` I believe.  This is also present in `OutputContext` and `InputContext` btw.  We would need to change those too.

That said, yes, I think I am convinced by your consistency argument.

> I've barely thought about it but my intuition says why not? Is there some extra complexity I'm missing? Partition ranges for non-asset runs seem like a perfectly coherent idea that as a naive user I would expect to work-- nothing about partition ranges belongs conceptually to the asset layer, so it is just extra arbitrary complexity that it doesn't work for ops. And it doesn't seem like much extra work since it's the same machinery used for assets anyway.

Thinking about this a little more, I think it might be possible theoretically, but I'm not sure how we'd implement it.

For a non-asset job, "is this partition green?" means "was there a successful run that targeted this partition"?  Currently, our way of answering that question is checking to see whether there's a run with a tag with "key=dagster/partition` and "value=<that partition>". Not sure how we'd solve this with run tags.

We don't have this problem with assets because we have a separate materialization event for each partition, even when a run targets multiple partitions.

smackesey:
>Most of our code examples use `asset_partition_time_window` I believe.

This name exactly isn't defined-- I see:

- `asset_partitions_time_window_for_output` (on `OpExecutionContext`)
- `asset_partitions_time_window_for_input` (on `OpExecutionContext`)
- `asset_partitions_time_window` (on `OutputContext` and `InputContext`)

Seeing this I actually think that we should rename the "for_output/input" methods to use the singular "partition" since that matches all the other methods, and just use `partition_time_window` inside `OutputContext`/`InputContext`. Here's the full list of methods in the codebase starting with "asset_partition...":

image

smackesey:
>For a non-asset job, "is this partition green?" means "was there a successful run that targeted this partition"? Currently, our way of answering that question is checking to see whether there's a run with a tag with "key=dagster/partition` and "value=". Not sure how we'd solve this with run tags.

So IIUC it's a DB indexing issue? (Since if you had the full list of runs you could just check for ones with range run tags and check if the target is in range, but the range logic isn't implemented in SQL)

sryza:
> So IIUC it's a DB indexing issue? (Since if you had the full list of runs you could just check for ones with range run tags and check if the target is in range, but the range logic isn't implemented in SQL)

Right

sryza:
I parse `partition_key_range` as "range where the start is a partition key and the end is a partition key" and `partitions_time_window` as "time window that covers all the partitions in the partition key range", so I wouldn't expect the to match. I also have a bias towards staying closer to what we currently have.

smackesey:
>I parse partition_key_range as "range where the start is a partition key and the end is a partition key" and partitions_time_window as "time window that covers all the partitions in the partition key range", so I wouldn't expect the to match. I also have a bias towards staying closer to what we currently have.

The currently existing uses without the `asset_` prefix in the codebase use the singular (`OpExecutionContext` and `StepOrchestrationContext`). Would you agree that regardless of plural/singular we should be consistent?

smackesey:
>I parse partition_key_range as "range where the start is a partition key and the end is a partition key" and partitions_time_window as "time window that covers all the partitions in the partition key range", so I wouldn't expect the to match. I also have a bias towards staying closer to what we currently have.

`OpExecutionContext` and `StepOrchestrationContext` currently define `partition_time_window` (i.e. singular)-- would you agree regardless of which way, we should be consistent on plural/singular here?

python_modules/dagster/dagster/_core/execution/context/compute.py
```diff
@@ -262,12 +262,25 @@ def partition_key(self) -> str:
...
-        """The asset partition key for the current run.
+        """The range of partition keys for the current run.
 
-        Raises an error if the current run is not a partitioned run.
+        If run is for a single partition key, return a `PartitionKeyRange` with the same start and
+        end. Raises an error if the current run is not a partitioned run.
+        """
+        return self.partition_key_range
+
+    @deprecated
```
alangenfeld, line 276: accident?

python_modules/dagster/dagster/_core/execution/context/invocation.py
```diff
@@ -671,6 +712,7 @@ def build_op_context(
     instance: Optional[DagsterInstance] = None,
     config: Any = None,
     partition_key: Optional[str] = None,
+    partition_key_range: Optional[PartitionKeyRange] = None,
```
alangenfeld, line 715: i just landed `build_asset_context` so will need to update that to support this argument as well

smackesey: done

smackesey:
@alangenfeld do you have a take on the above naming discussion

smackesey:
@sryza Just pinging on the above thread so we can reach a decision while it's still fresh

alangenfeld:
I don't have a take on the naming - I can take time to form an opinion if necessary

sryza:
Chatted with @smackesey offline and we decided to go with `partition_time_window` over `partitions_time_window`

smackesey:
@sryza is there anything else to discuss in this PR or are we good to go? I addressed @alangenfeld comments

python_modules/dagster/dagster/_core/execution/context/compute.py
```diff
@@ -262,12 +262,24 @@ def partition_key(self) -> str:
         """
         return self._step_execution_context.partition_key
 
+    @deprecated
```
sryza, line 265: We should say what we expect people to use instead.

sryza (approved):
One comment to address, otherwise looks good.

PR #15083, opened 2023-07-02
[refactor] Merge UnboundOpExecutionContext/BoundOpExecutionContext -> OpInvocationContext

smackesey:
## Summary & Motivation

Motivation:

Currently there are two classes `UnboundOpExecutionContext` and `BoundOpExecutionContext`. Both classes inherit from `OpExecutionContext`. `UnboundOpExecutionContext` is returned by `build_op_context`, i.e. it is used for direct op invocation. This is converted to a `BoundOpExecutionContext` in the `__call__` method of `OpDefinition` when an op is invoked.

The status quo is problematic for a few reasons:

- The names `(Un)BoundOpExecutionContext` are confusing. It makes it sound like
  these classes are used for all op executions, when in fact they are never
  used as part of a run-- only for what is called "invocation".
- There is a large amount of code duplication between the two classes. This is
  hard to maintain and probably a source of bugs.

## How I Tested These Changes

Existing BK tests


smackesey:
* master
  * **PR #15082** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15082" target="_blank">Graphite</a> 
    * **PR #15083** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15083" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15083?utm_source=stack-comment).

sryza (commented):
I'm going to defer to @dpeng817 on this one, at least initially. I agree that the duplication between these classes is undesirable, and it would be great to find a way to reduce it.

dpeng817 (commented):
I like this general direction, I think this makes sense. The previous state was definitely not user-friendly, this seems a lot more so, especially subclassing `OpExecutionContext` directly.

Some things that come to mind to consider:
- Having a reasonable experience on the "unbound" state - IE what happens when I call resource based methods
- Export OpInvocationContext at top level
- We never actually exported the bound / unbound contexts, so I don't think we need to worry about getting rid of them.

python_modules/dagster/dagster/_core/execution/context/invocation.py
```diff
@@ -206,19 +241,43 @@ def log(self) -> DagsterLogManager:
...
-    def solid(self) -> Node:
-        raise DagsterInvalidPropertyError(_property_msg("solid", "property"))
+    def op(self) -> Node:
+        raise DagsterInvalidPropertyError(_property_msg("op", "property"))
 
     @property
     def op_def(self) -> OpDefinition:
-        raise DagsterInvalidPropertyError(_property_msg("op_def", "property"))
+        self.validate_bound_property_access("op_def")
+        assert self._op_def is not None
```
dpeng817, line 253: I think we shy away from asserts in production code

smackesey:
>Having a reasonable experience on the "unbound" state - IE what happens when I call resource based methods

Can you elaborate here? In the unbound state properties that aren't set throw an error:

```def validate_bound_property_access(self, property_name: str) -> None:
        if not self.is_bound:
            raise DagsterInvalidPropertyError(
                f"Attempted to access {property_name} property of OpInvocationContext before "
                "context was bound."
            )
```

Resources can be provided to `build_op_context` so access works, as in the existing `test_op_invocation_with_resources`.

python_modules/dagster/dagster/_core/execution/context/invocation.py
dpeng817: makes sense, we can still use the assert to send a cleaner message though I believe right with the tuple / comma syntax?

dpeng817:
@smackesey I think I meant testing here, apologies. I quite like the `is_bound` property, just ensuring consistency around the experience there is all.

dpeng817:
> In the course of doing this I noticed that run_id returns the dummy value EPHEMERAL while other properties normally sourced from the StepExecutionContext throw an error. I think run_id should also throw an error for consistency-- what say you?

IIRC I added this as a stop-gap for a user quite a long time ago... I don't think we can just break that experience now unfortunately, but I do agree either an error, or allowing it to be set explicitly would be good, but I think that it's too late tbh

alangenfeld:
#15042 is related / tangential clean-up

alangenfeld (changes requested):
to your queue to rebase past merge conflicts at least, not sure if there was any additional reconciliation that needs to happen

PR #15088, opened 2023-07-03
Rebase for PR#14383

ArnoudItility:
## Summary & Motivation
Implemented and rebased to master the following features: 
- fallback capacity strategy: if no capacity available on initial capacity provider, then the fallback will be used
- capacity failure strategy: number of retries and retry interval when capacity is not available

## How I Tested These Changes
I was not able to locally test the unit tests. Tested the functionalities locally and we are currently running them on AWS using a patch file.


cmpadden:

PR #15078, opened 2023-07-01
Require docstring for @public methods

smackesey:
## Summary & Motivation

- Resolves #14996.
- Resolves #15076.

This updates our custom sphinx extension so that docstrings are required for `@public` methods. The docs build will fail if a `@public` method with no docstring is encountered.

This will enhance the quality of our docs and prevent strange behaviors that can occur in sphinx due to differential handling of objects with and without docstrings.

(NOTE: I also tacked the last few straggler docstring additions into this PR after the below list was completed)

---

Before this can land, there are about 200 `@public` methods that need docstrings. If we split the work it shouldn't be too bad. I have split the missing docstrings into ~10-ish groups below. If you want to help with this project, please comment and volunteer to take on 1-2 groups-- I'll check off the ones that are taken. I'll start by taking `DagsterInstance` and `Misc`.

@sryza @clairelin135 @OwenKephart @dpeng817 @benpankow @jamiedemaria @yuhan @tacastillo @alangenfeld 

(just names that came to mind that might be interested, no obligation)

- [x] Assets & Ops (@OwenKephart)
    - AssetsDefinition.can_subset
    - AssetsDefinition.dependency_keys
    - AssetsDefinition.descriptions_by_key
    - AssetsDefinition.get_partition_mapping
    - AssetsDefinition.group_names_by_key
    - AssetsDefinition.key
    - AssetsDefinition.keys
    - AssetsDefinition.node_def
    - AssetsDefinition.op
    - AssetsDefinition.partitions_def
    - AssetsDefinition.required_resource_keys
    - AssetsDefinition.resource_defs
    - OpDefinition.alias
    - OpDefinition.config_schema
    - OpDefinition.ins
    - OpDefinition.name
    - OpDefinition.outs
    - OpDefinition.required_resource_keys
    - OpDefinition.retry_policy
    - OpDefinition.tag
    - OpDefinition.tags
    - OpDefinition.version
    - OpDefinition.with_hooks
    - OpDefinition.with_retry_policy

- [x] Config & Resources (@benpankow)
    - Field.default_value
    - Field.description
    - Field.is_required
    - Map.key_label_name
    - Array.description
    - ResourceDefinition.description
    - ResourceDefinition.required_resource_keys
    - ResourceDefinition.string_resource
    - ResourceDefinition.version
    - InitResourceContext.instance
    - InitResourceContext.log
    - InitResourceContext.resources

- [x] DagsterInstance (@smackesey)
    - DagsterInstance.add_dynamic_partitions
[...]

PR #15096, opened 2023-07-03
Add docstrings for public job methods

dpeng817:
To help through https://github.com/dagster-io/dagster/pull/15078

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/job_definition.py
+++ b/python_modules/dagster/dagster/_core/definitions/job_definition.py
@@ -328,31 +328,53 @@ class JobDefinition(IHasInternalInit):
     @public
     @property
     def executor_def(self) -> ExecutorDefinition:
+        """Returns the default :py:class:`ExecutorDefinition` for the job.
+
+        If the user has not specified an executor definition, then this will default to the :py:func:`multi_or_in_process_executor`. If a default is specified on the :py:class:`Definitions` object the job was provided to, then that will be used instead.
+        """
         return self._executor_def or DEFAULT_EXECUTOR_DEF
 
     @public
     @property
     def has_specified_executor(self) -> bool:
+        """Returns True if this job has explicitly specified an executor, and False if the executor was inherited through defaults or the :py:class:`Definitions` object the job was provided to.
+        """
         return self._executor_def is not None
 
     @public
     @property
     def resource_defs(self) -> Mapping[str, ResourceDefinition]:
+        """Returns the set of ResourceDefinition objects specified on the job.
+
+        This may not be the complete set of resources required by the job, since those can also be provided on the :py:class:`Definitions` object the job may be provided to.
+        """
         return self._resource_defs
 
     @public
     @property
     def partitioned_config(self) -> Optional[PartitionedConfig]:
+        """The partitioned config for the job, if it has one.
+
+        A partitioned config defines a way to map partition keys to run config for the job.
+        """
         return self._partitioned_config
 
     @public
     @property
     def config_mapping(self) -> Optional[ConfigMapping]:
+        """The config mapping for the job, if it has one.
+
+        A config mapping defines a way to map a top-level config schema to run config for the job.
+        """
         return self._config_mapping
 
     @public
     @property
     def loggers(self) -> Mapping[str, LoggerDefinition]:
+        """Returns the set of LoggerDefinition objects specified on the job.
+
+        If the user has not specified a mapping of :py:class:`LoggerDefinition` objects, then this will default to the :py:func:`colored_console_logger` under the key `console`. If a default is specified on the :py:class:`Definitions` object the job was provided to, then that will be used instead.
+        """
         from dagster._loggers import default_loggers
 
         return self._loggers or default_loggers()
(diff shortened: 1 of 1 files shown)


dpeng817:
* master
  * **PR #15096** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15096" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15096?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -328,38 +328,62 @@ def dependencies(self) -> DependencyMapping[NodeInvocation]:
...
+        """Returns True if this job has explicitly specified an executor, and False if the executor was inherited through defaults or the :py:class:`Definitions` object the job was provided to.
+        """
         return self._executor_def is not None
 
     @public
     @property
     def resource_defs(self) -> Mapping[str, ResourceDefinition]:
+        """Returns the set of ResourceDefinition objects specified on the job.
+
+        This may not be the complete set of resources required by the job, since those can also be provided on the :py:class:`Definitions` object the job may be provided to.
```
dpeng817, line 349: would like verification from @benpankow that this is the case

python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -328,38 +328,62 @@ def dependencies(self) -> DependencyMapping[NodeInvocation]:
     @public
     @property
     def executor_def(self) -> ExecutorDefinition:
+        """Returns the default :py:class:`ExecutorDefinition` for the job.
```
smackesey, line 331: IMO this would benefit from a note on overriding the "default", because the next paragraph makes it sound like the"default" is just `multi_or_in_process_executor`. Maybe:

>Returns the default :py:class:`ExecutorDefinition` for the job. This is the `ExecutorDefinition` that will be used if one is not specified at runtime via config.

python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -328,38 +328,62 @@ def dependencies(self) -> DependencyMapping[NodeInvocation]:
...
+
+        If the user has not specified an executor definition, this will be the :py:func:`multi_or_in_process_executor`, which defaults to multiprocess mode.
+        """
         return self._executor_def or DEFAULT_EXECUTOR_DEF
 
     @public
     @property
     def has_specified_executor(self) -> bool:
+        """Returns True if this job has explicitly specified an executor, and False if the executor was inherited through defaults or the :py:class:`Definitions` object the job was provided to.
```
smackesey, line 340: Given the above confusion around "default", would avoid "inherited through defaults" here and just say "inherited from the Definitions object..."

python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -328,38 +328,62 @@ def dependencies(self) -> DependencyMapping[NodeInvocation]:
...
+        """
         return self._config_mapping
 
     @public
     @property
     def loggers(self) -> Mapping[str, LoggerDefinition]:
+        """Returns the set of LoggerDefinition objects specified on the job.
+
+        If no loggers were provided on the job, this will be the :py:func:`colored_console_logger`.
+        """
```
smackesey, line 377: Aren't these inherited from `Definitions` just like the executor? If so, I would use similar language to my executor comment.

python_modules/dagster/dagster/_core/definitions/job_definition.py
```diff
@@ -328,38 +328,62 @@ def dependencies(self) -> DependencyMapping[NodeInvocation]:
...
         from dagster._loggers import default_loggers
 
         return self._loggers or default_loggers()
 
     @public
     @property
     def has_specified_loggers(self) -> bool:
+        """Returns true if the job explicitly set loggers, and False if loggers were inherited through defaults or the :py:class:`Definitions` object the job was provided to.
```
smackesey, line 385: [1]

smackesey (approved):
Some suggestions for clarity around executors/loggers, but looks great otherwise

python_modules/dagster/dagster/_core/definitions/job_definition.py
dpeng817: Good point regarding inherited from Definitions - although if nothing is specified on Definitions, a default is still set.

smackesey: We should probably change it so that `Definitions` just has a default, that simplifies the model and makes it so that the job inherits from the `Definitions` object with no caveat.

python_modules/dagster/dagster/_core/definitions/job_definition.py
dpeng817: I agree with the sentiment, although not entirely accurate to say that the default executor `multi_or_in_process_executor` comes from `Definitions` object, since it will be set on the job if none is provided (just overwritten)

python_modules/dagster/dagster/_core/definitions/job_definition.py
dpeng817: I think that change will be hairier than we anticipate (as in, a few hours of work rather than a one liner). For now, I think we should describe the situation as accurately as possible.

python_modules/dagster/dagster/_core/definitions/job_definition.py
dpeng817: From re-reading docs, pretty positive that this is the case.

PR #15097, opened 2023-07-03
Add ParentMaterialized details

johannkm:
Surface which parent materialized, causing the eager materialize condition

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -756,7 +756,7 @@ def determine_asset_partitions_to_auto_materialize(
                 if parents:
                     conditions.add(
                         ParentMaterializedAutoMaterializeCondition(
-                            materialized_asset_keys=frozenset(parents)
+                            updated_asset_keys=frozenset(parents)
                         )
                     )
 
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -747,16 +747,30 @@ def determine_asset_partitions_to_auto_materialize(
             )
         ):
             conditions.add(MissingAutoMaterializeCondition())
-        else:
-            # if the parent has been updated
-            if auto_materialize_policy.on_new_parent_data:
-                parents = instance_queryer.get_missing_or_updated_or_going_to_update_parents(
-                    asset_partition=candidate
-                )
-                if parents:
+        # if parent data has been updated or will update
+        elif auto_materialize_policy.on_new_parent_data:
+            if not asset_graph.is_source(candidate.asset_key):
+                parent_asset_partitions = asset_graph.get_parents_partitions(
+                    dynamic_partitions_store=instance_queryer,
+                    current_time=evaluation_time,
+                    asset_key=candidate.asset_key,
+                    partition_key=candidate.partition_key,
+                ).parent_partitions
+                updated_parents = {
+                    parent.asset_key
+                    for parent in instance_queryer.get_updated_parent_asset_partitions(
+                        candidate, parent_asset_partitions
+                    )
+                }
+                will_update_parents = set()
+                for parent in parent_asset_partitions:
+                    if _will_materialize_for_conditions(conditions_by_asset_partition.get(parent)):
+                        will_update_parents.add(parent.asset_key)
+                if updated_parents or will_update_parents:
                     conditions.add(
                         ParentMaterializedAutoMaterializeCondition(
-                            updated_asset_keys=frozenset(parents)
+                            updated_asset_keys=frozenset(updated_parents),
+                            will_update_asset_keys=frozenset(will_update_parents),
                         )
                     )
 
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -750,13 +750,13 @@ def determine_asset_partitions_to_auto_materialize(
         else:
             # if the parent has been updated
             if auto_materialize_policy.on_new_parent_data:
-                ancestors = instance_queryer.get_ancestors_missing_or_with_unreconciled_children(
-                    candidate
+                parents = instance_queryer.get_missing_or_updated_or_going_to_update_parents(
+                    asset_partition=candidate
                 )
-                if ancestors:
+                if parents:
                     conditions.add(
                         ParentMaterializedAutoMaterializeCondition(
-                            materialized_asset_keys=frozenset(ancestors)
+                            materialized_asset_keys=frozenset(parents)
                         )
                     )
 
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -747,6 +747,7 @@ def determine_asset_partitions_to_auto_materialize(
             )
         ):
             conditions.add(MissingAutoMaterializeCondition())
+
         # if parent data has been updated or will update
         elif auto_materialize_policy.on_new_parent_data:
             if not asset_graph.is_source(candidate.asset_key):
@@ -756,16 +757,20 @@ def determine_asset_partitions_to_auto_materialize(
                     asset_key=candidate.asset_key,
                     partition_key=candidate.partition_key,
                 ).parent_partitions
-                updated_parents = {
-                    parent.asset_key
-                    for parent in instance_queryer.get_updated_parent_asset_partitions(
-                        candidate, parent_asset_partitions
-                    )
-                }
+
+                (
+                    updated_parent_asset_partitions,
+                    _,
+                ) = instance_queryer.get_updated_and_missing_parent_asset_partitions(
+                    candidate, parent_asset_partitions
+                )
+                updated_parents = {parent.asset_key for parent in updated_parent_asset_partitions}
+
                 will_update_parents = set()
                 for parent in parent_asset_partitions:
                     if _will_materialize_for_conditions(conditions_by_asset_partition.get(parent)):
                         will_update_parents.add(parent.asset_key)
+
                 if updated_parents or will_update_parents:
                     conditions.add(
                         ParentMaterializedAutoMaterializeCondition(
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -747,12 +747,18 @@ def determine_asset_partitions_to_auto_materialize(
             )
         ):
             conditions.add(MissingAutoMaterializeCondition())
-
-        # if the parent has been updated
-        if auto_materialize_policy.on_new_parent_data and not instance_queryer.is_reconciled(
-            asset_partition=candidate
-        ):
-            conditions.add(ParentMaterializedAutoMaterializeCondition())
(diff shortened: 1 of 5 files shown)


johannkm:
* master
  * **PR #15097** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15097" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15097?utm_source=stack-comment).

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -731,12 +731,18 @@ def conditions_for_candidate(
...
-            conditions.add(ParentMaterializedAutoMaterializeCondition())
+        else:
+            # if the parent has been updated
+            if auto_materialize_policy.on_new_parent_data:
+                ancestors = instance_queryer.get_ancestors_missing_or_with_unreconciled_children(
+                    candidate
+                )
+                if ancestors:
+                    conditions.add(
+                        ParentMaterializedAutoMaterializeCondition(
```
johannkm, line 742: Currently there's two cases we're conflating: parent materialized, and parent is missing and will be materialized (e.g. if the parent is time partitioned and got a new partition), so we're materializing this asset with it. I don't think it's urgent but we can separate those cases eventually

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -636,11 +636,66 @@ def get_root_unreconciled_ancestors(
...
-    def is_reconciled(self, asset_partition: AssetKeyPartitionKey) -> bool:
-        """Returns a boolean representing if the given `asset_partition` is currently reconciled.
-        An asset (partition) is considered unreconciled if any of:
-        - It has never been materialized
-        - One of its parents has been updated more recently than it has
-        - One of its parents is unreconciled.
-        """
-        return len(self.get_root_unreconciled_ancestors(asset_partition=asset_partition)) == 0
+    @cached_method
+    def get_ancestors_missing_or_with_unreconciled_children(
```
johannkm, line 640: similar logic to the above method but gets the parents that changed instead of the unreconciled assets

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -636,11 +636,69 @@ def get_root_unreconciled_ancestors(
...
         return root_unreconciled_ancestors
 
-    def is_reconciled(self, asset_partition: AssetKeyPartitionKey) -> bool:
-        """Returns a boolean representing if the given `asset_partition` is currently reconciled.
-        An asset (partition) is considered unreconciled if any of:
-        - It has never been materialized
-        - One of its parents has been updated more recently than it has
-        - One of its parents is unreconciled.
+    @cached_method
+    def get_ancestors_missing_or_with_unreconciled_children(
```
OwenKephart, line 640: re: our conversation yesterday, we'd want to replace this with something that returns the immediate parents that have been updated more recently. In terms of factoring things out, I think the body of the `get_missing_or_updated_parents` function could be shared between this and the recursive function. I was toying around with it and it turned out to be a bit tougher than I thought it'd be, but I believe this should basically do the trick, although it'd need to be cleaned up a bit:

```python
    def _get_updated_or_missing_parent_asset_partitions(
        self,
        asset_partition,
        parent_asset_partitions,
    ):
        parent_asset_partitions_by_key: Dict[AssetKey, List[AssetKeyPartitionKey]] = defaultdict(
            list
        )
        for parent in parent_asset_partitions:
            parent_asset_partitions_by_key[parent.asset_key].append(parent)

        time_or_dynamic_partitioned = isinstance(
            self.asset_graph.get_partitions_def(asset_partition.asset_key),
            (TimeWindowPartitionsDefinition, DynamicPartitionsDefinition),
        )
        updated_or_missing_parents = set()
        for parent_key, parent_asset_partitions in parent_asset_partitions_by_key.items():
            # ignore non-observable source parents
            if self.asset_graph.is_source(parent_key) and not self.asset_graph.is_observable(
                parent_key
            ):
                continue

            # find missing parents
            for parent in parent_asset_partitions:
                if not self.asset_partition_has_materialization_or_observation(parent):
                    updated_or_missing_parents.add(parent)

            # when mapping from time or dynamic downstream to unpartitioned upstream, only check
            # for existence of upstream materialization, do not worry about updates
            if time_or_dynamic_partitioned and not self.asset_graph.is_partitioned(parent_key):
                continue

            updated_or_missing_parents.update(self.get_asset_partitions_updated_after_cursor(
                asset_key=parent_key,
                asset_partitions=parent_asset_partitions,
                after_cursor=self.get_latest_materialization_or_observation_storage_id(
                    asset_partition
                ),
            ))
        return updated_or_missing_parents

    @cached_method
    def get_root_unreconciled_ancestors(
        self, *, asset_partition: AssetKeyPartitionKey
    ) -> AbstractSet[AssetKey]:
        """Return the set of root unreconciled ancestors of the given asset partition, i.e. the set
        of ancestors of this asset partition which are unreconciled for a reason other than that
        one of their ancestors is unreconciled.
        """
        # always treat source assets as reconciled
        if self.asset_graph.is_source(asset_partition.asset_key):
            return set()
        elif not self.asset_partition_has_materialization_or_observation(asset_partition):
            return {asset_partition.asset_key}

        parent_asset_partitions = self.asset_graph.get_parents_partitions(
            dynamic_partitions_store=self,
            current_time=self._evaluation_time,
            asset_key=asset_partition.asset_key,
            partition_key=asset_partition.partition_key,
        ).parent_partitions

        updated_or_missing_parent_asset_partitions = self._get_updated_or_missing_parent_asset_partitions(
            asset_partition,
            parent_asset_partitions
        )

        root_unreconciled_ancestors = { asset_partition.asset_key } if updated_or_missing_parent_asset_partitions else set()

        # recurse over parents
        for parent in set(parent_asset_partitions) - updated_or_missing_parent_asset_partitions:
            root_unreconciled_ancestors.update(
                self.get_root_unreconciled_ancestors(asset_partition=parent)
            )
        return root_unreconciled_ancestors

```

One small note is that I think technically the missing / updated asset partitions should be split out into two different sets, as we should only count an asset_partition as a "root unreconciled asset partition" if its parent has been updated more recently than it. If its parent is missing, we should instead count the missing parent as the "root unreconciled asset partition". Minor change in code, just was too lazy to do it here

OwenKephart: Might also want to refactor how we're passing in the parent asset partitions, but a big thing here is that

```python
self.asset_graph.get_parents_partitions(...)
```

Can be a fairly expensive call so ideally we'd only do that once per asset partition

johannkm:
Refactored to share logic and only refer to immediate parents. A future diff can filter out assets with no AMP (will be a little easier to do once I fully remove the reconciliation sensor)

python_modules/dagster/dagster/_core/definitions/auto_materialize_condition.py
```diff
@@ -45,6 +45,7 @@ class ParentMaterializedAutoMaterializeCondition(NamedTuple):
     """
 
     decision_type: AutoMaterializeDecisionType = AutoMaterializeDecisionType.MATERIALIZE
+    materialized_asset_keys: Optional[FrozenSet[AssetKey]] = None
```
OwenKephart, line 48: If we're being pedantic, this whole class should be called ParentUpdatedAutoMaterializeCondition, as an observable source asset upstream of this asset would also cause this condition to be applied.

It's easier to retroactively change the name of this class than change the name of the class and the field, so maybe we should call this `updated_asset_keys`.

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -604,6 +604,50 @@ def get_asset_partitions_updated_after_cursor(
             if previous_versions.get(asset_partition) != version
         }
 
+    def _get_updated_or_missing_parent_asset_partitions(
+        self,
+        asset_partition,
```
OwenKephart, line 609: types would be good to add on this method

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -618,64 +662,59 @@ def get_root_unreconciled_ancestors(
...
 
-        for parent in self.asset_graph.get_parents_partitions(
+        # recurse over parents
+        for parent in set(parent_asset_partitions) - updated_or_missing_parent_asset_partitions:
+            root_unreconciled_ancestors.update(
+                self.get_root_unreconciled_ancestors(asset_partition=parent)
+            )
+        return root_unreconciled_ancestors
+
+    def get_missing_or_updated_or_going_to_update_parents(
```
OwenKephart, line 689: I think this is a bit of a confusing formulation ("going to update parents" only makes sense if you understand the implementation details of the auto materialize logic, and is not actually accurate in cases where some of the parents don't have auto materialize policies.

I think it'd be better to just have this function focus on the missing / updated parents, then separately use the information we have available to us in the `conditions_for_candidate` area to determine which parents are going to be materialized. Then we can just union those sets (set of parents that have been updated more recently than X, as well as set of parents that are going to materialize on this tick).

OwenKephart (dismissed):

sryza:
Is this PR still live?

johannkm:
yep, about to update

python_modules/dagster/dagster/_core/definitions/auto_materialize_condition.py
```diff
@@ -45,6 +45,8 @@ class ParentMaterializedAutoMaterializeCondition(NamedTuple):
     """
 
     decision_type: AutoMaterializeDecisionType = AutoMaterializeDecisionType.MATERIALIZE
+    updated_asset_keys: Optional[FrozenSet[AssetKey]] = None
+    will_update_asset_keys: Optional[FrozenSet[AssetKey]] = None
```
johannkm, line 49: we might as well track these separately, can always decide not to distinguish in the ui

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -748,11 +748,36 @@ def conditions_for_candidate(
...
+                    updated_parent_asset_partitions,
+                    _,
+                ) = instance_queryer.get_updated_and_missing_parent_asset_partitions(
+                    candidate, parent_asset_partitions
+                )
+                updated_parents = {parent.asset_key for parent in updated_parent_asset_partitions}
+
+                will_update_parents = set()
+                for parent in parent_asset_partitions:
+                    if _will_materialize_for_conditions(conditions_by_asset_partition.get(parent)):
```
johannkm, line 771: I think we can refactor the waiting conditions to do something similar instead of traversing back up the graph

johannkm (commented):
Updated to separate parent updated and parent will update cases. The latter now just uses the parent's conditions, which are more accurate (e.g. we used to return a ParentMaterialized condition as if an upstream was going to materialize, even if it didn't have an AMP).

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
OwenKephart: probably correct on that one yeah

OwenKephart (approved):
this looks solid -- out of curiosity, what would happen in the UI if we had e.g. an asset where partition 1 had upstream asset A update and partition 2 had upstream asset B update? is that all handled as expected?

this isn't a new issue in this PR (same issue exists for the "waiting on asset keys" stuff, just wondering)

johannkm:
We list the waiting on (and will do the same for this) per partition. In the future we could record the specific partition as well, that just prevents grouping of conditions in more cases

PR #15109, opened 2023-07-05
[ez] Add dot icon

salazarm:
## Summary & Motivation

New icon just dropped
Screenshot 2023-07-05 at 2 14 32 PM

## How I Tested These Changes
👀


bengotow (approved):
This looks good to me! `partition_missing` is also a hollow circle, but I think it's good to have two separate icons for the two separate use cases

PR #15114, opened 2023-07-05
Avoid creation of local directories by the GCS IO Manager

PeterJCLaw:
## Summary & Motivation

Prior to this change the underlying behaviour from the base `UPathIOManager` would be invoked when the user had configured the use of a GCS IO Manager. This would mostly result in the (perhaps innocuous) creation of local directories, however when running on a filesystem where the user does not have write access it would cause `IOError`s that would cause job runs to fail.

This change copies the approach taken in the AWS S3 IO Manager, as (just like S3) GCS doesn't really have the concept of directories and there is no need to try to create them (either locally or in the remote store).

## How I Tested These Changes

Validated by applying this fix on top of 1.3.12 and running with that in our cluster.

~Are there any unit tests I should be adding to?~

### DIFF
--- a/python_modules/libraries/dagster-gcp/dagster_gcp/gcs/io_manager.py
+++ b/python_modules/libraries/dagster-gcp/dagster_gcp/gcs/io_manager.py
@@ -58,6 +58,10 @@ class PickledObjectGCSIOManager(UPathIOManager):
     def _uri_for_path(self, path: UPath) -> str:
         return f"gs://{self.bucket}/{path}"
 
+    def make_directory(self, path: UPath) -> None:
+        # It is not necessary to create directories in GCP
+        return None
+
     def load_from_path(self, context: InputContext, path: UPath) -> Any:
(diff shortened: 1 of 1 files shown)


jamiedemaria:
thanks for the spot and for the contribution! taking a look now

jamiedemaria (commented):
change looks good. If you want to add a unit test, the right file would be https://github.com/dagster-io/dagster/blob/master/python_modules/libraries/dagster-gcp/dagster_gcp_tests/gcs_tests/test_io_manager.py 

I don't think it requires a unit test (we don't have one for the same feature for the s3 io manager) but if you want to add one that ensures that no new directory was created, that would also be fine!

Once you've tested on your end, I'll run the CI pipeline and once we get that passing I'll approve

PeterJCLaw:
@jamiedemaria we've confirmed that the GCS IO manager does indeed work with this fix.

jamiedemaria (approved):

PR #15117, opened 2023-07-05
[poc] Show waiting on assets in evaluations ui

johannkm:
- now we surface waitingOnAssetKeys for the `Waiting on upstream data` condition over gql
- in a related change, the list of conditions returned over gql now may have more than 1 condition with type `ParentOutdatedAutoMaterializeCondition`, since different partition subsets can have different assets they're waiting on. 

The current front end doesn't handle the above well since it makes a map from type -> partition subset. That should be changed to type -> partition subset[].

image


johannkm:
* master
  * **PR #15117** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15117" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15117?utm_source=stack-comment).

PR #15118, opened 2023-07-05
remove dbt Cloud callout from top of dbt integration page

sryza:
## Summary & Motivation

Context: https://elementl-workspace.slack.com/archives/C05BF0CSC07/p1688596540672729

### DIFF
--- a/docs/content/integrations/dbt.mdx
+++ b/docs/content/integrations/dbt.mdx
@@ -5,11 +5,6 @@ description: Dagster can orchestrate dbt alongside other technologies.
 
 # dbt + Dagster
 
-<Note>
-  Using dbt Cloud? Check out the{" "}
-  <a href="/integrations/dbt-cloud">dbt Cloud with Dagster guide</a>!
-</Note>
-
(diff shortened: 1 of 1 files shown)


erinkcochran87:
Lookin' at that Slack message you linked to, I get the reasoning behind removing this callout. I'm cool with that, but should we also fold the dbt Cloud guide under the dbt heading in the sidenav? That can be part of another PR, but bringing it up here to see if there's related, additional work before closing this guy out.

Screen Shot 2023-07-05 at 8 49 04 PM

sryza:
> should we also fold the dbt Cloud guide under the dbt heading in the sidenav?

I think this is an important question to raise, but my instinct would be to keep them separate, because Dagster integrates with dbt core in a fundamentally different way than it integrates with dbt Cloud.  I could be overruled on this though.

erinkcochran87:
@sryza That's fine - I don't want to hold this up for the event this week. We can come back to this afterwards. There are probably some small tweaks we can make to house everything together/standardize integration docs that follow a similar pattern without it being confusing. I'll stamp this guy for now!

PR #15121, opened 2023-07-05
unexperimentalize DynamicPartitionsDefinition and SensorResult

sryza:
## Summary & Motivation

I actually forgot these were experimental until I noticed a user say they're not using them because they're marked as such: https://dagster.slack.com/archives/C01U5LFUZJS/p1685986062633069

A fair number of users rely on this functionality, and it receives few complaints.  Here are the areas where I could imagine changes happening in this realm:
- DynamicPartitionsDefinition names are currently deployment-scoped, like asset keys. This allows dynamically-partitioned pipelines to span code locations. I could also imagine situations where users would want to minimize the opportunity for collisions. Users have not brought this up yet.
- They don't yet work with dynamic mapping. Here's the issue where we're tracking this. I don't anticipate that this would require breaking changes to the above APIs.
- Some users have requested dynamic partition mappings. I don't anticipate that this would require breaking changes to the above APIs.

### DIFF
--- a/docs/content/concepts/partitions-schedules-sensors/partitions.mdx
+++ b/docs/content/concepts/partitions-schedules-sensors/partitions.mdx
@@ -132,14 +132,6 @@ result = materialize(
 
 #### Dynamically partitioned assets
 
-<Experimental />
-
-<Note>
-  This feature is <strong>experimental</strong>. Until this feature is marked as
-  stable, further migrations may be required and partition history is not
-  guaranteed to be retained.
-</Note>
-
 Sometimes you don't know the set of partitions ahead of time when you're defining your assets. For example, maybe you want to add a new partition every time a new data file lands in a directory, or every time you want to experiment with a new set of hyperparameters. In these cases, you can use a <PyObject object="DynamicPartitionsDefinition"/>.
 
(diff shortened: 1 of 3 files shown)


erinkcochran87 (commented):
Doc changes look fine to me; I'll leave it to Claire to :+1: on the issues you raised in the description

clairelin135 (approved):
I'm ok with unexperimentalizing.

Wanted to surface another possible change though--in product review we previously discussed being able to view a history of dynamic partitions changes (i.e. displaying a history of when keys were added or deleted). This would likely involve running a schema migration that adds additional columns to the dynamic partitions table.

I think it's fine to unexperimentalize regardless of this because (a) nobody's asked for this (outside of product review) and (b) even if we did un-experimentalize and later decided we wanted to support this behavior, we can support backcompat for unmigrated versions of the db.

sryza:
@clairelin135 good call out on that change.

> I think it's fine to unexperimentalize regardless of this because (a) nobody's asked for this (outside of product review) and (b) even if we did un-experimentalize and later decided we wanted to support this behavior, we can support backcompat for unmigrated versions of the db.

I agree.

PR #15122, opened 2023-07-06
[ui] Update the treatment of "canceling" and "canceled" backfills and runs

### LINKED ISSUE #14646: Cancelled backfill runs cannot be re-run from the backfill page
### Dagster version

dagster, version 1.3.7

### What's the issue?

If a pending backfill run was cancelled, it can no longer be resubmitted from the backfill page.  In the example below, I should be able to submit 4 runs, but the UI only let's me submit one.  The difference is that the one run failed and the others were cancelled.

#### Backfill Page
image

#### Jobs from the cancelled backfill
image

I should be able to submit backfills for cancelled jobs.  Either with a separate checkbox or as part of the "Failed" group.

### How to reproduce?

1. Setup a simple partitioned job
2. Start a backfill, but cancel the runs it before they all complete
3. Try to backfill again, but notice that the cancelled jobs cannot be selected

### Deployment type

Dagster Helm chart

### Deployment details

I've seen this in local development as well as deployed via a Helm chart on EKS.

### Additional information

_No response_

bengotow:
## Summary & Motivation

Resolves #14646

Summary of changes:
https://www.loom.com/share/4b1a2806311a4ab1bdff16ef14da4673

- Canceled runs and backfills have a gray status instead of a red status

- Canceling runs are shown in blue with a spinner instead of red.

- The run page favicon now uses the in progress daggy for canceling and the plain daggy for canceled instead of the red daggy.

- Partition status no longer returns a "canceled" status for some partitions -- it ignores canceled runs and looks at the previous runs (This is a backend change to `get_run_partition_data`). This means that cancellation is essentially invisible on the partition health bar - per the slack discussion last month, this seems like the best thing to do. (https://elementl-workspace.slack.com/archives/C03CA4TVCAW/p1685652714171209?thread_ts=1685572176.294529&cid=C03CA4TVCAW)

- Partition status in the expanded view no longer leaves in-progress / canceling / starting steps "gray" and copies the blue styling that is used on the overall partition health bar.

There were a couple other places we were special-casing canceled / canceling and I tried to revisit them and address them consistently. 

Before:
image

After:

image

Before:

image

After: 
image

Before:

image

After:
image

## How I Tested These Changes

I added test runs to cover the backend change and tested the front-end changes by starting + canceling two backfills, and then additionally modifying my postgres runs table so that I had perma-`Canceling` and `Starting` runs. Those cases seem rarer and hard to catch organically.

I also opened storybook and checked the RunStatus and Timeline stories, and added a "Canceling" state to the BackfillTable story.

### DIFF
--- a/js_modules/dagit/packages/core/src/partitions/PartitionStepStatus.tsx
+++ b/js_modules/dagit/packages/core/src/partitions/PartitionStepStatus.tsx
@@ -440,10 +440,13 @@ const PartitionSquare: React.FC<{
   } else if (runs.length === 0) {
     squareStatus = 'empty';
   } else {
-    const runStatus = runs[runs.length - 1]!.status;
-    squareStatus = runStatus === RunStatus.CANCELED ? 'failure' : runStatus.toLowerCase();
+    const runStatus = [...runs].reverse().find((r) => r.status !== RunStatus.CANCELED)?.status;
+    if (runStatus) {
+      squareStatus = runStatus.toLowerCase();
+    } else {
+      squareStatus = 'empty';
+    }
   }
-
   const content = (
     <div
       className={`square ${squareStatus}`}
--- a/js_modules/dagit/packages/core/src/runs/RunStatusDots.tsx
+++ b/js_modules/dagit/packages/core/src/runs/RunStatusDots.tsx
@@ -27,7 +27,7 @@ interface RunStatusProps {
 }
 
 export const RunStatusIndicator: React.FC<RunStatusProps> = React.memo(({status, size}) => {
-  if (status === 'STARTED') {
+  if (status === 'STARTED' || status === 'CANCELING') {
     return <Spinner purpose="caption-text" />;
   }
   if (status === 'SCHEDULED') {
--- a/python_modules/dagster-graphql/dagster_graphql/implementation/fetch_partition_sets.py
+++ b/python_modules/dagster-graphql/dagster_graphql/implementation/fetch_partition_sets.py
@@ -12,7 +12,7 @@ from dagster._core.host_representation.external_data import (
     ExternalPartitionExecutionErrorData,
     ExternalPartitionNamesData,
 )
-from dagster._core.storage.dagster_run import RunPartitionData, RunsFilter
+from dagster._core.storage.dagster_run import DagsterRunStatus, RunPartitionData, RunsFilter
 from dagster._core.storage.tags import (
     PARTITION_NAME_TAG,
     PARTITION_SET_TAG,
@@ -229,6 +229,7 @@ def get_partition_set_partition_statuses(
 
     run_partition_data = graphene_info.context.instance.run_storage.get_run_partition_data(
         runs_filter=RunsFilter(
+            statuses=[status for status in DagsterRunStatus if status != DagsterRunStatus.CANCELED],
             tags={
                 PARTITION_SET_TAG: partition_set_name,
                 REPOSITORY_LABEL_TAG: repository_handle.get_external_origin().get_label(),
--- a/python_modules/dagster/dagster/_core/storage/runs/sql_run_storage.py
+++ b/python_modules/dagster/dagster/_core/storage/runs/sql_run_storage.py
@@ -903,6 +903,8 @@ class SqlRunStorage(RunStorage):
             for row in rows:
                 if not row["partition"] or row["partition"] in _partition_data_by_partition:
                     continue
+                if DagsterRunStatus[row["status"]] == DagsterRunStatus.CANCELED:
+                    continue
 
                 _partition_data_by_partition[row["partition"]] = RunPartitionData(
                     run_id=row["run_id"],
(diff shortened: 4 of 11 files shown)


python_modules/dagster/dagster/_core/storage/runs/sql_run_storage.py
```diff
@@ -903,6 +903,8 @@ def get_run_partition_data(self, runs_filter: RunsFilter) -> Sequence[RunPartiti
             for row in rows:
                 if not row["partition"] or row["partition"] in _partition_data_by_partition:
                     continue
+                if DagsterRunStatus[row["status"]] == DagsterRunStatus.CANCELED:
```
clairelin135, line 906: I think this method is used to display run statuses for backfills which means that canceled backfill runs may not display as expected in the UI. In this case I believe we try to display a run status "bar" for each partition, and since the canceled partitions no longer have runs they'd display as not started?

I think we probably instead edit the partition set runs filter to not select canceling runs here: https://github.com/dagster-io/dagster/blob/429661fed123d71a43309a1f98e3e23145098e03/python_modules/dagster-graphql/dagster_graphql/implementation/fetch_partition_sets.py#L231-L237

clairelin135 (commented):
Thanks for making these changes, all of the color changes make sense. I left one comment about the changes to avoid behavior changes to the backfill run status bar

python_modules/dagster/dagster/_core/storage/runs/sql_run_storage.py
bengotow: Ahh it looks like you're right, good catch! I think moving this up into the filter on the original query makes sense, will change this. Glad you noticed this!

js_modules/dagit/packages/core/src/partitions/PartitionStepStatus.tsx
```diff
@@ -440,10 +440,13 @@ const PartitionSquare: React.FC<{
   } else if (runs.length === 0) {
     squareStatus = 'empty';
   } else {
-    const runStatus = runs[runs.length - 1]!.status;
-    squareStatus = runStatus === RunStatus.CANCELED ? 'failure' : runStatus.toLowerCase();
+    const runStatus = [...runs].reverse().find((r) => r.status !== RunStatus.CANCELED)?.status;
```
hellendag, line 443:
```
const runStatus = runs.findLast(...);
```

js_modules/dagit/packages/core/src/runs/RunStatusDots.tsx
```diff
@@ -27,7 +27,7 @@ interface RunStatusProps {
 }
 
 export const RunStatusIndicator: React.FC<RunStatusProps> = React.memo(({status, size}) => {
-  if (status === 'STARTED') {
+  if (status === 'STARTED' || status === 'CANCELING') {
     return <Spinner purpose="caption-text" />;
```
hellendag, line 31: We could make the spinner red in the canceling case, though I don't. know if it might look terrible or if @braunjj might hate it. 😆

hellendag (approved):

js_modules/dagit/packages/core/src/partitions/PartitionStepStatus.tsx
salazarm: https://caniuse.com/?search=findLast

salazarm: Screenshot 2023-07-10 at 12 16 39 PM
very recent feature, might be best to avoid it?

python_modules/dagster-graphql/dagster_graphql/implementation/fetch_partition_sets.py
```diff
@@ -229,6 +229,7 @@ def get_partition_set_partition_statuses(
 
     run_partition_data = graphene_info.context.instance.run_storage.get_run_partition_data(
         runs_filter=RunsFilter(
+            statuses=[status for status in DagsterRunStatus if status != DagsterRunStatus.CANCELED],
```
clairelin135, line 232: maybe we should also include "canceling" runs?

js_modules/dagit/packages/core/src/partitions/PartitionStepStatus.tsx
bengotow: Wow I didn't know about findLast! Super cool. I think it's probably unlikely that someone is still using a version of Chrome that is 1+ year old... but I wonder if this would be the first breaking API use? It doesn't seem like we use `findLast` anywhere else yet. 

We use `flatMap`, but that's since 2018, and `Object.fromEntries` which is since 2019 (Chrome 73). Maybe I'll hold off on this for now :/

python_modules/dagster-graphql/dagster_graphql/implementation/fetch_partition_sets.py
bengotow: Ahh I was debating - I think that makes sense as well, will change this!

js_modules/dagit/packages/core/src/partitions/PartitionStepStatus.tsx
salazarm: I know we had a user report a broken page due to using a recent `Intl` feature. I believe they were on Chrome 80ish.

PR #15124, opened 2023-07-06
Avoid trying to create local directory unnecessarily

judahrand:
## Summary & Motivation

Without overriding the `make_directory` method to be a noop the `PickledObjectGCSIOManager` creates a local directory which is never used or cleaned up. The same override is present in the AWS equivalent: https://github.com/dagster-io/dagster/blob/37ccaadad044038846ce0c1c890507e50e116ce8/python_modules/libraries/dagster-aws/dagster_aws/s3/io_manager.py#L66-L68


jamiedemaria:
This is being addressed in https://github.com/dagster-io/dagster/pull/15114#pullrequestreview-1515434881

PR #15132, opened 2023-07-06
[ui] Fix auto-materialize conditions for non-partitioned asset

hellendag:
## Summary & Motivation

When determining the set of materialization conditions, I accidentally changed things only to regard conditions that affected some set of partition keys. This broke non-partitioned assets. Fix the issue by using all conditions in the non-partitioned asset.

Screenshot 2023-07-06 at 10 42 09 AM
Screenshot 2023-07-06 at 10 41 57 AM

## How I Tested These Changes

View non-partitioned asset, verify that when the missing materialization condition is present, it is correctly checked in the UI. Verify that partitioned assets are the same as before.


johannkm (approved):

salazarm (approved):
Can we add a test for this?

hellendag:
> Can we add a test for this?

Yeah, I'll do that after the PR that splits the middle panel. The partitions and non-partitions cases should get their own tests.

PR #15135, opened 2023-07-06
Remove `time_window_partition_scope_minutes`

OwenKephart:
## Summary & Motivation

This has been superseded by the max_materializations_per_minute parameter, and requires a bunch of extra code to maintain.

## How I Tested These Changes

Note that several tests were updated here. At a high level, there are some things that cannot fully be replicated just by swapping between the partition scope and the max materialization limits. This mostly is impactful on the first tick, but most of our tests start with a fresh cursor, so these differences are magnified.

On the first tick, imagine you have a daily-partitioned asset which has only today's data filled in.

Old world: Nothing happens, as there are no time partitions within the scope

New world: The day before yesterday's data will get filled in, as we have a limit of one materialization, and so we'll try to fill in the most recent unfilled partition.

Tests that have been updated will be annotated with the reason for the update.

Note that after the first tick, no further historical partitions should be filled in, as all older partitions will have been discarded. This is not currently how it works (hence the skipped test), but this will be remedied in the next PR in the stack.


python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/partition_scenarios.py
```diff
@@ -71,7 +72,13 @@
 ]
 
 
-time_multipartitioned_asset = [asset_def("asset1", partitions_def=time_multipartitions_def)]
+time_multipartitioned_asset = [
+    asset_def(
+        "asset1",
+        partitions_def=time_multipartitions_def,
+        auto_materialize_policy=AutoMaterializePolicy.eager(max_materializations_per_minute=2),
```
OwenKephart, line 79: similar to above, this MultiPartitionsDefinition has a static dimension of size 2, so setting the limit to 2 allows it to fill in one day's worth of data, aligning with our tests

python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/partition_scenarios.py
```diff
@@ -195,15 +202,15 @@
         unevaluated_runs=[
             run(["hourly"], partition_key=partition_key)
             for partition_key in hourly_partitions_def.get_partition_keys_in_range(
-                PartitionKeyRange(start="2013-01-06-00:00", end="2013-01-06-23:00")
+                PartitionKeyRange(start="2013-01-05-00:00", end="2013-01-05-23:00")
```
OwenKephart, line 205: same thing as before, just backing up the time to the start of the partitions def

python_modules/dagster/dagster_tests/definitions_tests/asset_reconciliation_tests/test_asset_reconciliation_sensor.py
```diff
@@ -109,6 +109,7 @@ def sort_run_request_key_fn(run_request):
         assert run_request.partition_key == expected_run_request.partition_key
 
 
+@pytest.mark.skip("Temporarily skip until #14984 is closed")
```
OwenKephart, line 112: this test asserts that, if you do a tick of a certain scenario, then nothing happens before the next tick, then nothing should happen.

This is a correct assertion, but because of the bug in the linked issue (where we don't fully discard old time partitions), this test fails. Fixing this later on in the stack

sryza (approved):
Nice!

Issue #14984, opened 2023-06-27
when time_window_partitions_scope=None, root asset partitions are requested after being discarded


sryza:
On each tick, the Auto-materialize history tab will say that partitions were discarded, but each tick will also request a run.

```python
from dagster import asset, DailyPartitionsDefinition, AutoMaterializePolicy

@asset(
    partitions_def=DailyPartitionsDefinition(start_date="2020-01-01")
    auto_materialize_policy=AutoMaterializePolicy(
        on_missing=True,
        on_new_parent_data=True,
        for_freshness=True,
        time_window_partition_scope_minutes=None,
        max_materializations_per_minute=1,
    )
)
def asset1():
    ...
```

sryza:
@OwenKephart is this on your radar already?

OwenKephart:
@sryza Yep on the radar -- the reason is that we don't currently merge the discarded partitions back into the materialized_or_requested_root_asset_partitions bit of the cursor. should be a simple change

PR #15136, opened 2023-07-06
Properly discard root asset partitions

OwenKephart:
## Summary & Motivation

Resolves: https://github.com/dagster-io/dagster/issues/14984

Depends on: https://github.com/dagster-io/dagster/pull/15135

Renamed a bunch of stuff, using "handled" to represent "materialized_or_requested_or_discarded".

## How I Tested These Changes

The test that used to fail now passes, also added a new test for good measure.


python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -179,10 +176,10 @@ class AssetReconciliationCursor(NamedTuple):
         latest_storage_id:
             The latest observed storage ID across all assets. Useful for finding out what has
             happened since the last tick.
-        materialized_or_requested_root_asset_keys:
+        handled_root_asset_keys:
```
sryza, line 179: Do we need to update this docstring to include discarded?

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -316,30 +319,28 @@ def from_serialized(cls, cursor: str, asset_graph: AssetGraph) -> "AssetReconcil
...
             evaluation_id = data[3] if len(data) == 4 else 0
             serialized_last_observe_request_timestamp_by_asset_key = {}
         else:
             latest_storage_id = data["latest_storage_id"]
-            serialized_materialized_or_requested_root_asset_keys = data[
-                "materialized_or_requested_root_asset_keys"
-            ]
-            serialized_materialized_or_requested_root_partitions_by_asset_key = data[
-                "materialized_or_requested_root_partitions_by_asset_key"
+            serialized_handled_root_asset_keys = data["handled_root_asset_keys"]
```
sryza, line 330: is the idea that we don't need to handle backcompat for the old names because we haven't released since we added them?

sryza (approved):

python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
OwenKephart: exactly yep

PR #15142, opened 2023-07-06
[docs] small io manager docs fixes

jamiedemaria:
## Summary & Motivation
A collection of small io manager docs fixes that i ran across in OSS support this week. the changes are:
* fix link to the table of built in io managers so that it works (previously there were two sections with the same name, causing the link to direct to the wrong heading)
* add the pythonic IO managers to the API docs page 
* add a comment to the InMemoryIOManager that clarifies that materialized values are garbage-collected after each run


jamiedemaria:
* master
  * **PR #15142** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15142" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15142?utm_source=stack-comment).

erinkcochran87 (commented):
LGTM, but do we need to clarify what `garbage-collected` means?

jamiedemaria:
i feel like garbage-collected is pretty common coding terminology, but we could also say something like 
"after execution completes, the in-memory data structure storing the materializes values will be deleted" if that makes more sense

erinkcochran87 (approved):

PR #15130, opened 2023-07-06
[ui] Auto-materialization: Repair "Before" item in left list

hellendag:
## Summary & Motivation

When no auto-materialize evaluations have taken place yet, make sure to show the empty state item in the left list.

Additionally, on the last page of left list items, ensure that the "Before [date]" list item always appears at the end.

Screenshot 2023-07-06 at 9 57 57 AM
Screenshot 2023-07-06 at 9 43 42 AM

## How I Tested These Changes

Jest. Wipe out Dagster data, view auto-materialize page for a handful of assets. Verify that the "Before now" item appears. Enable auto-materialization, then verify that the "Before [...]" item appears at the end of the list.

PR #15144, opened 2023-07-06
[ui] Split AutomaterializeMiddlePanel

hellendag:
## Summary & Motivation

Split the auto-materialize middle panel component for the partitioned and non-partitioned cases. There are some shared pieces, but enough differences at this point that it's a bit simpler to treat them separately.

## How I Tested These Changes

View partitioned and non-partitioned assets' auto-materialization pages. Verify that they render correctly.

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/AutoMaterializePolicyPage/AutomaterializeMiddlePanel.tsx
+++ b/js_modules/dagit/packages/core/src/assets/AutoMaterializePolicyPage/AutomaterializeMiddlePanel.tsx
@@ -1,108 +1,44 @@
 import {useQuery} from '@apollo/client';
-import {Box, Colors, Subheading, Icon, NonIdealState} from '@dagster-io/ui';
+import {Box, Colors, NonIdealState, Subheading} from '@dagster-io/ui';
 import * as React from 'react';
-import styled from 'styled-components/macro';
 
 import {ErrorWrapper} from '../../app/PythonErrorInfo';
 import {AssetKey} from '../types';
 
-import {AutomaterializeRequestedPartitionsLink} from './AutomaterializeRequestedPartitionsLink';
-import {AutomaterializeRunTag} from './AutomaterializeRunTag';
+import {AutomaterializeMiddlePanelNoPartitions} from './AutomaterializeMiddlePanelNoPartitions';
+import {AutomaterializeMiddlePanelWithPartitions} from './AutomaterializeMiddlePanelWithPartitions';
 import {GET_EVALUATIONS_QUERY} from './GetEvaluationsQuery';
 import {EvaluationOrEmpty} from './types';
-import {
-  AutoMateralizeWithConditionFragment,
-  GetEvaluationsQuery,
-  GetEvaluationsQueryVariables,
-} from './types/GetEvaluationsQuery.types';
-
-const isRequestCondition = (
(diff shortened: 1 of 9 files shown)


hellendag:
* master
  * **PR #15130** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15130" target="_blank">Graphite</a> 
    * **PR #15132** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15132" target="_blank">Graphite</a> 
      * **PR #15144** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15144" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15144?utm_source=stack-comment).

js_modules/dagit/packages/core/src/assets/AutoMaterializePolicyPage/AutomaterializeMiddlePanel.tsx
```diff
@@ -1,352 +0,0 @@
...
-    {
-      variables: {
-        assetKey,
-        cursor: selectedEvaluation?.evaluationId
-          ? (selectedEvaluation.evaluationId + 1).toString()
-          : undefined,
-        limit: 2,
-      },
-    },
-  );
```
hellendag, line 80: It's not clear to me that we actually need this query -- the page already retrieves this data, and the selected evaluation is passed to the middle panel. Perhaps I'm missing something?

salazarm: Originally this was needed because we wanted the ability for you to go to the next page without changing the selected evaluation. In that case, since the page is changed then the data for the selected evaluation won't be present. I believe Johann changed the logic to just pick the first item (so switching pages also switches your selected evaluation which maybe is not a big deal).

salazarm (approved):

js_modules/dagit/packages/core/src/assets/AutoMaterializePolicyPage/AutomaterializeMiddlePanel.tsx
johannkm: I think the initial thought for a second query was that we could query the minimum amount for the left panel (just numRequested etc.) and all the conditions info for the main panel. Probably premature optimization

salazarm: @johannkm It's also for linking directly to an evaluation. Right now if the evaluation isn't in the left panel (because it's no longer on the first page) and you give someone a link to the evaluation, then when the evaluation falls outside of the first page (there's no cursor in the URL for the first page), then the link stops working.

salazarm: If we include the cursor even on page 1 then the URL linking would still work, its not super straight forward to do without causing a refetch though.

hellendag:
I restored the query in the middle panel based on the pagination use case.

- I'm now passing along the evaluation ID and querying for it.
- At the page level, only select the first item in the list by default if we're on the first page. Otherwise, allow the current selection to be `undefined` so that we don't blow away the evaluation shown in the middle panel if we paginate.

PR #15145, opened 2023-07-06
[auto-materialize] Fix error when calculating data time of in progress runs

OwenKephart:
## Summary & Motivation

https://dagster.slack.com/archives/C02LJ7G0LAZ/p1688653467290869

High level, we were originally assuming that we always start our recursive calls from an asset materialization event, even though this is not the case (and the inner recursion handles asset observations just fine)

## How I Tested These Changes

Reproduced the error with the added test, now it succeeds.


johannkm (approved):

johannkm:
Quick fix!

PR #15156, opened 2023-07-07
add defs property to SensorEvaluationContext

sryza:
## Summary & Motivation

`SensorEvaluationContext` currently exposes a `repository_def` property, but no property that feels relevant in the world of `Definitions`.

This property will make it possible to use definition-level information to make decisions about what computations to request. For example, request all assets whose code versions have changed since their latest materialization.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/definitions_class.py
+++ b/python_modules/dagster/dagster/_core/definitions/definitions_class.py
@@ -225,7 +225,7 @@ def _create_repository_using_definitions_args(
     resources: Optional[Mapping[str, Any]] = None,
     executor: Optional[Union[ExecutorDefinition, Executor]] = None,
     loggers: Optional[Mapping[str, LoggerDefinition]] = None,
-):
+) -> RepositoryDefinition:
     check.opt_iterable_param(
         assets, "assets", (AssetsDefinition, SourceAsset, CacheableAssetsDefinition)
     )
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster/_core/definitions/definitions_class.py
```diff
@@ -402,6 +402,12 @@ def __init__(
             loggers=loggers,
         )
 
+    @staticmethod
+    def from_repository(repo: RepositoryDefinition) -> "Definitions":
+        defs = Definitions()
+        defs._created_pending_or_normal_repo = repo  # noqa: SLF001
+        return defs
+
```
schrockn, line 410: This is super gross.

schrockn (changes requested):
How about an approach where a RepositoryDefinition gets a reference to the `Definitions` object that created it, and then return that out of `defs`. This would also make it so the object is actually the same object, rather than a reconstructed one, which I think is more clear.

Also what about all the other context types? We should be consistent.

sryza:
> Also what about all the other context types? We should be consistent.

The only other context with a `repository_def` argument is `ScheduleEvaluationContext`. I'll add it there.  `OpExecutionContext` etc. don't have one in part because there isn't always a repository / definitions when you're executing one.  I tried to add it once and was not super simple.

prha (changes requested):
queue mgmt

cmpadden:

PR #15157, opened 2023-07-07
DagsterInstance.get_latest_materialization_code_versions

sryza:
## Summary & Motivation

Adds a `DagsterInstance` method that finds the code version assigned to the latest materialization of each of a set of assets.

This will enable users to e.g. write sensors that only request runs for assets whose definition code versions have changed since their latest materialization.

### DIFF
--- a/python_modules/dagster/dagster/_core/instance/__init__.py
+++ b/python_modules/dagster/dagster/_core/instance/__init__.py
@@ -18,6 +18,7 @@ from typing import (
     Callable,
     Dict,
     Generic,
+    Iterable,
     List,
     Mapping,
     Optional,
@@ -34,6 +35,7 @@ from typing_extensions import Protocol, Self, TypeAlias, TypeVar, runtime_checka
 
 import dagster._check as check
 from dagster._annotations import public
+from dagster._core.definitions.data_version import extract_data_provenance_from_entry
 from dagster._core.definitions.events import AssetKey
 from dagster._core.errors import (
     DagsterHomeNotSetError,
@@ -1751,7 +1753,7 @@ class DagsterInstance(DynamicPartitionsStore):
 
     @traced
     def get_latest_materialization_events(
-        self, asset_keys: Sequence[AssetKey]
+        self, asset_keys: Iterable[AssetKey]
     ) -> Mapping[AssetKey, Optional["EventLogEntry"]]:
         return self._event_storage.get_latest_materialization_events(asset_keys)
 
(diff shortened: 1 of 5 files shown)


python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -2645,3 +2647,31 @@ def get_latest_data_version_record(
...
+
+        Args:
+            asset_keys (Iterable[AssetKey]): The asset keys to find latest materialization code
+                versions for.
+
+        Returns:
+            Mapping[AssetKey, Optional[str]]: A dictionary with a key for each of the provided asset
+                keys. The values will be None if the asset has no materializations. If an asset does
+                not have a code version explicitly assigned to its definitions but was
+                materialization, Dagster assigns a randomly-generated string as its code version.
```
smackesey, line 2666: "was materialization" -> "was materialized'

smackesey: Dagster uses the run id as the code version if none is on the def, might as well say that

python_modules/dagster/dagster/_core/storage/event_log/sql_event_log.py
```diff
@@ -1202,9 +1202,9 @@ def get_asset_keys(
         return [asset_key for asset_key in asset_keys if asset_key]
 
     def get_latest_materialization_events(
-        self, asset_keys: Sequence[AssetKey]
+        self, asset_keys: Iterable[AssetKey]
```
smackesey, line 1205: Good idea to put in `Iterables` here

python_modules/dagster/dagster/_core/instance/__init__.py
```diff
@@ -2645,3 +2647,31 @@ def get_latest_data_version_record(
...
+        """
+        result: Dict[AssetKey, Optional[str]] = {}
+        latest_materialization_events = self.get_latest_materialization_events(asset_keys)
+        for asset_key in asset_keys:
+            event = latest_materialization_events.get(asset_key)
+            if event is None:
+                result[asset_key] = None
+            else:
+                result[asset_key] = (event.tags or {}).get(CODE_VERSION_TAG)
```
smackesey, line 2675: Rather than directly accessing the `CODE_VERSION_TAG`, I suggest using `get_data_provenance_from_entry(event).code_version`. More future-proof against changes to code version persistence.

python_modules/dagster/dagster/_core/instance/__init__.py
sryza: Got it - made these changes

python_modules/dagster/dagster/_core/instance/__init__.py
sryza: Got it - made this change

smackesey (approved):

PR #15168, opened 2023-07-07
[auto-materialize] Use partition status cache to filter out common case on startup

OwenKephart:
## Summary & Motivation

When we get a huge number of new materialization events between two successive ticks, we want to investigate all of those children as potential candidates to be kicked off, except in the case that those children were intended to be materialized in the same run as that materialized parent.

This is to prevent us from double-firing in the following case:

Imagine we have asset graph `A -> B -> C`. A is materialized, causing us to kick off a run consisting of `B` and `C`. While this run executes, `B` is materialized. Before `C` is materialized (either because it's still in progress or it failed), we evaluate another tick, which notices that `B` has been materialized. Now, we need to check to see if `C` was intended to be materialized in that run in order to determine if we should consider materializing `B`.

We filter out some cases where `C` and `B` could not possibly exist in the same run, but there are plenty where they could, and previously we would resolve this by a) fetching the latest materialization event for the asset partition of `B` b) grabbing the run record for that run, and c) checking if `C` was planned for that run. This can be pretty expensive, as things go, as if we have thousands of new materialization events, then we need to potentially grab thousands of event log records / dagster runs.

This PR optimizes this significantly, with the observation that we only run into problems if the child is currently in progress or failed. If the child has been successfully materialized, it doesn't matter if that happened in the same run as its parent or not. Most partitions of an asset will not be in progress or failed at any given time, so this significantly cuts down on the number of event log records we need to fetch.

To help realize the full potential of these performance gains, I made some edits to the partition status cache logic to speed it up a bunch, as we'll end up calling this a fair amount. Those changes are:

1. Make it possible to feed in an existing AssetRecord if we have one handy -- this is nice because in the AutoMaterialize logic, we already have fetched all the asset records we need, so grabbing it again is wasteful
2. Use the last_materialization_storage_id on the AssetRecord to avoid having to query the instance to get this -- we already have this info, so again, grabbing it is wasteful
3. Only write the updated cache value to the instance if it differs from the one that was stored -- no reason to incur extra database calls if we can avoid it.

## How I Tested These Changes

Passes all unit tests, and added a few more in the "run still in progress" realm just to be safe


johannkm (approved):
lgtm

PR #15176, opened 2023-07-07
remove code that handles Out args to multi_asset

sryza:
## Summary & Motivation

We previously removed the ability to provide `Out`s to the `outs` argument of `@multi_asset`, but we left around some vestigial code for handling it in some places.


OwenKephart (approved):
nice

PR #15177, opened 2023-07-07
PythonObjectMetadataValue

sryza:
## Summary & Motivation

This adds the ability to include arbitrary Python objects as metadata on asset and job definitions. This metadata rides along with the definition in the same process, but *does not get serialized when the metadata is serialized*. This means it does not show up in the UI and does not end up in the event log.

The initial use case for this is to allow `DbtManifest` objects to live on the asset definitions that are derived from them. `DbtCli.cli` will be able to access the `DbtManifest` from the context object and thus avoid requiring the user to re-specify it.

### DIFF
--- a/python_modules/dagster/dagster/__init__.py
+++ b/python_modules/dagster/dagster/__init__.py
@@ -233,6 +233,7 @@ from dagster._core.definitions.metadata import (
     NullMetadataValue as NullMetadataValue,
     PathMetadataValue as PathMetadataValue,
     PythonArtifactMetadataValue as PythonArtifactMetadataValue,
+    PythonObjectMetadataValue as PythonObjectMetadataValue,
     TableMetadataValue as TableMetadataValue,
     TableSchemaMetadataValue as TableSchemaMetadataValue,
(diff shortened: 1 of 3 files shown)


OwenKephart (approved):
I like it

smackesey:
Hate to rain on the parade, but input and output definitions already allow for arbitrary metadata values-- they just aren't wrapped with anything. Seems like this PR introduces a new way of packaging an arbitrary metadata value? I think we should stay consistent, which would mean we have to keep the old way for backcompat-- are there advantages to introducing a new wrapper?

sryza:
> Hate to rain on the parade, but input and output definitions already allow for arbitrary metadata values-- they just aren't wrapped with anything. Seems like this PR introduces a new way of packaging an arbitrary metadata value?

Ah didn't realize the prior way still worked. Not raining on the parade, this makes my life simpler. Going to retire this, at least for now.

sryza:
@smackesey eh let's hold off for now. I'm still not sure that this is a good pattern, and requests for it haven't come up elsewhere as far as I know.

PR #15184, opened 2023-07-08
Fix perf regression

OwenKephart:
## Summary & Motivation

We were calling `materializiable_asset_keys` a bunch, causing us to do a set union which seems to be much less efficient on older versions of python. Just calculate this upfront instead

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -84,6 +84,10 @@ class AssetGraph:
         self._code_versions_by_key = code_versions_by_key
         self._is_observable_by_key = is_observable_by_key
         self._auto_observe_interval_minutes_by_key = auto_observe_interval_minutes_by_key
+        # source assets keys can sometimes appear in the upstream dict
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/definitions/asset_graph.py
```diff
@@ -176,12 +180,11 @@ def from_assets(
 
     @property
     def all_asset_keys(self) -> AbstractSet[AssetKey]:
-        return self.materializable_asset_keys | self.source_asset_keys
+        return self._asset_dep_graph["upstream"].keys() | self.source_asset_keys
```
smackesey, line 183: Same result but I think this reads cleaner if it's `self.materializable_asset_keys | self.source_asset_keys`

OwenKephart: yeah good point re: readability

as for precomputing all_asset_keys as well, I think it's fine to wait and see if we hit any issues (this seems like a niche thing that only hits older python versions in a subset of cases, so would rather not take any additional memory hit if possible)

smackesey (approved):
my bad I thought I approved this before

PR #15185, opened 2023-07-08
with @dbt_assets, pass manifest through context

sryza:
## Summary & Motivation

The user no longer needs to explicitly pass the manifest when invoking `DbtCli.cli` from inside `@dbt_assets`. The aim is to reduce boilerplate and help avoid errors that would be caused by passing different manifests to the `@dbt_assets` and `DbtCli.cli`.

Previous:

```python
    @dbt_assets(manifest=manifest)
    def all_dbt_assets(context, dbt: DbtCli):
        yield from dbt.cli(["build"], manifest=manifest, context=context).stream()
```

With this PR:
```python
    @dbt_assets(manifest=manifest)
    def all_dbt_assets(context, dbt: DbtCli):
        yield from dbt.cli(["build"], context=context).stream()
```

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorator.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorator.py
@@ -42,7 +42,7 @@ def dbt_assets(
 
             @dbt_assets(manifest=manifest)
             def my_dbt_assets(context: OpExecutionContext, dbt: DbtCli):
-                yield from dbt.cli(["build"], manifest=manifest, context=context).stream()
+                yield from dbt.cli(["build"], context=context).stream()
     """
     unique_ids = select_unique_ids_from_manifest(
         select=select, exclude=exclude or "", manifest_json=manifest.raw_manifest
@@ -60,6 +60,7 @@ def dbt_assets(
         dbt_nodes=manifest.node_info_by_dbt_unique_id,
         deps=deps,
         io_manager_key=io_manager_key,
+        manifest=manifest,
     )
 
     def inner(fn) -> AssetsDefinition:
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/asset_utils.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/asset_utils.py
@@ -1,6 +1,17 @@
 import hashlib
 import textwrap
-from typing import AbstractSet, Any, Dict, FrozenSet, List, Mapping, Optional, Set, Tuple
+from typing import (
+    TYPE_CHECKING,
+    AbstractSet,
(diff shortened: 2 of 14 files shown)


python_modules/libraries/dagster-dbt/dagster_dbt/asset_utils.py
```diff
@@ -248,7 +265,10 @@ def get_dbt_multi_asset_args(
             io_manager_key=io_manager_key,
             description=default_description_fn(node_info, display_raw_sql=False),
             is_required=False,
-            metadata=default_metadata_fn(node_info),
+            metadata={
+                **default_metadata_fn(node_info),
+                MANIFEST_METADATA_KEY: manifest,
+            },
```
rexledesma, line 271: Could we add this as metadata on the asset definition instead of the asset definition's outputs?

Located in [1], where we do it for `dagster-dbt/select` and `dagster-dbt/exclude`.

python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorator.py
```diff
@@ -60,6 +60,7 @@ def my_dbt_assets(context: OpExecutionContext, dbt: DbtCli):
         dbt_nodes=manifest.node_info_by_dbt_unique_id,
         deps=deps,
         io_manager_key=io_manager_key,
+        manifest=manifest,
     )
 
     def inner(fn) -> AssetsDefinition:
```
rexledesma, line 66: [1]

rexledesma (commented):
neat

python_modules/libraries/dagster-dbt/dagster_dbt/asset_utils.py
sryza: I agree that would be preferable, however `op_tags` doesn't accept arbitrary Python objects.

If we went with the proposal raised [here](https://github.com/dagster-io/internal/discussions/6143), then we could add it as a field on `DbtAssetsDefinition`.

python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorator.py
```diff
@@ -42,7 +42,7 @@ def dbt_assets(
 
             @dbt_assets(manifest=manifest)
             def my_dbt_assets(context: OpExecutionContext, dbt: DbtCli):
-                yield from dbt.cli(["build"], manifest=manifest, context=context).stream()
+                yield from dbt.cli(["build"], context=context).stream()
```
rexledesma, line 45: Should make sure that all other API docs follow this new convention

rexledesma (approved):

PR #15186, opened 2023-07-08
accept Path and dict args for manifest argument of `@dbt_assets`

sryza:
## Summary & Motivation

The aim here is to make it so that users don't need to understand dagster-dbt's notion of a DbtManifest until they need it.

The current `DbtManifest` class is kind of two things:
1. A simple wrapper around a raw manifest
2. A set of override-able methods that act as a translation layer between dbt concepts and Dagster concepts

It's probably going to make sense to rename `DbtManifest` to something like `DagsterDbtManifest` or `DbtManifestAdapter` in order to disambiguate from the dbt library's `Manifest` class and to make it clear that the manifest is responsible for (2).

Allowing a Path or JSON blob allows users who don't need to think about (2) to avoid dealing with this Dagster concept, as well as removing a small amount of boilerplate.

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorator.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorator.py
@@ -1,5 +1,7 @@
-from typing import Callable, Optional
+from pathlib import Path
+from typing import Any, Callable, Mapping, Optional, Union
 
+import dagster._check as check
 from dagster import AssetsDefinition, PartitionsDefinition, multi_asset
 from dagster._annotations import experimental
 
@@ -11,7 +13,7 @@ from .utils import ASSET_RESOURCE_TYPES, select_unique_ids_from_manifest
 @experimental
 def dbt_assets(
     *,
-    manifest: DbtManifest,
+    manifest: Union[Mapping[str, Any], DbtManifest, Path],
     select: str = "fqn:*",
     exclude: Optional[str] = None,
     io_manager_key: Optional[str] = None,
@@ -20,7 +22,10 @@ def dbt_assets(
     """Create a definition for how to compute a set of dbt resources, described by a manifest.json.
 
     Args:
-        manifest (DbtManifest): The wrapper of a dbt manifest.json.
+        manifest (Union[Mapping[str, Any], DbtManifest, Path]): The contents of a manifest.json file
+            or the path to a manifest.json file. A manifest.json contains a representation of a
+            dbt project (models, tests, macros, etc). We use this representation to create
+            corresponding Dagster assets.
         select (Optional[str]): A dbt selection string for the models in a project that you want
             to include. Defaults to "*".
         exclude (Optional[str]): A dbt selection string for the models in a project that you want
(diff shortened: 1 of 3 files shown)


python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorator.py
```diff
@@ -20,7 +22,9 @@ def dbt_assets(
     """Create a definition for how to compute a set of dbt resources, described by a manifest.json.
 
     Args:
-        manifest (DbtManifest): The wrapper of a dbt manifest.json.
+        manifest (Union[Mapping[str, Any], DbtManifest, Path]): A path to or contents of a dbt
+            manifest.json file that contains the dbt models, seeds, snapshots, etc. that we want to
+            create Dagster asset definitions for.
```
rexledesma, line 27: This is an odd sentence.

```suggestion
        manifest (Union[Mapping[str, Any], DbtManifest, Path]): The contents of a dbt
            manifest.json file that contains the representation of a dbt project (models, tests, macros, etc). We use this representation to create corresponding Dagster assets. Alternatively, the path to a manifest.json can be specified.
```

python_modules/libraries/dagster-dbt/dagster_dbt/asset_decorator.py
```diff
@@ -34,16 +38,21 @@ def dbt_assets(
...
-            from dagster_dbt import DbtCli, DbtManifest, dbt_assets
-
-            manifest = DbtManifest.read(path="target/manifest.json")
+            from pathlib import Path
 
+            from dagster import OpExecutionContext
+            from dagster_dbt import DbtCli, dbt_assets
 
-            @dbt_assets(manifest=manifest)
+            @dbt_assets(manifest=Path("target", "manifest.json"))
```
rexledesma, line 46: Might have to check the other docstrings in `resources_v2` to make sure its now following this new convention

python_modules/libraries/dagster-dbt/dagster_dbt_tests/test_asset_decorator.py
```diff
@@ -34,6 +35,45 @@ def all_dbt_assets(context, dbt: DbtCli):
     ).success
 
 
+def test_json_manifest():
+    with open(manifest_path, "r") as f:
+        manifest_json = json.load(f)
```
rexledesma, line 40: nit: use `pathlib` for opening the file
```suggestion
    with manifest_path.open() as f:
        manifest_json = json.load(f)
```

python_modules/libraries/dagster-dbt/dagster_dbt_tests/test_asset_decorator.py
```diff
@@ -34,6 +35,45 @@ def all_dbt_assets(context, dbt: DbtCli):
     ).success
 
 
+def test_json_manifest():
```
rexledesma, line 38: Can we just combine make this into one test, and just parameterize `manifest`?

rexledesma, line 38: Can we just combine make this into one test, and just parameterize `manifest`?

```suggestion
@pytest.mark.parametrize("manifest", ...)
def test_dbt_assets_from_manifest():
```

rexledesma (changes requested):

slopp (commented):
I like this change, I've seen users of `load_assets_from_dbt_manifest` be confused as to why that function doesn't accept a path but requires users to write the `json.loads` themselves. 

My only other thought is that sometimes in large dbt projects even loading that json file can be very slow. I'd hope that maybe with the new dbt python interface they might optimize that serialization and, ideally, we could take advantage of whatever manifest-to-python reader they create and maintain.

rexledesma (approved):

PR #15191, opened 2023-07-10
use auto_observe_interval_minutes in auto-materialize concept docs

sryza:


johannkm (approved):
wow I hadn't realized how much code this removes

PR #15195, opened 2023-07-10
tweaks to argument order, types, and docs for get_status_by_partition

sryza:
## Summary & Motivation

- Add a bit more docs.
- Use `Sequence` instead of `List`, to indicate that we won't mutate it.
- Move `partitions_def` argument to the end. I think that, in the future, it will be possible to improve the implementation so that in some cases this argument isn't required. Better to put non-required arguments at the end.


clairelin135 (approved):

sryza:
@renzhe-brian just wanted to give you a heads up on this PR

PR #15203, opened 2023-07-10
[docs] update op events output page

jamiedemaria:
## Summary & Motivation
A user was confused by the op events page discussing outputs and thought that all returned values from an op needed to be wrapping in `Output`. I can see why this is the case since we don't showcase the reasons you use `Output` in the doc examples.  https://dagster.slack.com/archives/C01U954MEER/p1688763698975829

I put together these updates to add the use of `metadata` to `Output` to give some indication that `Output` is being used because we are attaching other information. I also adding the non Output version in the little introduction paragraph. Let me know if you think these help!

### DIFF
--- a/docs/content/concepts/ops-jobs-graphs/op-events.mdx
+++ b/docs/content/concepts/ops-jobs-graphs/op-events.mdx
@@ -38,7 +38,17 @@ Yielding events from within the body of an op is a useful way of communicating w
 
 Because returning a value from an op is such a fundamental part of creating a data pipeline, we have a few different interfaces for this functionality.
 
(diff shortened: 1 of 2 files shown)


jamiedemaria:
* master
  * **PR #15203** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15203" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15203?utm_source=stack-comment).

tacastillo (approved):
lgtm 👍🏽

PR #15206, opened 2023-07-10
adding backfill policy to asset definitions and related API objects

ruizh22:
## Summary & Motivation
RFC: https://github.com/dagster-io/dagster/discussions/14829

This PR only introduces BackfillPolicy to the core API model layer, it should have no effect on the system that is not yet using the properties added. 

## How I Tested These Changes
While I am still working on adding more unit tests to make sure those added properties are working as expected, I ran the existing unit tests and they were all passing without issues. 

Also, please advise on the best way to do the unit testing mentioned above.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_graph.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_graph.py
@@ -165,7 +165,7 @@ class AssetGraph:
                 group_names_by_key.update(asset.group_names_by_key)
                 freshness_policies_by_key.update(asset.freshness_policies_by_key)
                 auto_materialize_policies_by_key.update(asset.auto_materialize_policies_by_key)
-                backfill_policies_by_key.update(asset.backfill_policies_by_key)
+                backfill_policies_by_key.update({key: asset.backfill_policy for key in asset.keys})
                 if len(asset.keys) > 1 and not asset.can_subset:
                     for key in asset.keys:
                         required_multi_asset_sets_by_key[key] = asset.keys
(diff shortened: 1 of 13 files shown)


ruizh22:
@sryza Can you offer some advice on the best places to add tests for those changes?

sryza (commented):
Hey @ruizh22 - thanks for submitting this.  I haven't had a chance to give a detailed review yet, but a few high-level comments:
- Instead of a backfill policy per key, I think the backfill policy should be per-`AssetsDefinition`, because, if multiple assets are generated by the same function, I don't think it's possible to parallelize that function in different ways.
- On `BackfillPolicy`, we could leave out the `is_single_run` boolean? It seems like the code makes the assumption that `max_partitions_per_run=None` corresponds to a single run.
- Here are a few files where it's worth adding tests:
  - test_external_asset_graph.py for testing the changes to `ExternalAssetGraph`
  - test_external_data.py for testing the changes to `ExternalAssetNode`

sryza (commented):
This looks great and has passed all the build checks.  Two small comments on the docs for `BackfillPolicy`:
- "todo: add more docs." in the class docstring. We generally try to avoid having TODOs in the codebase, especially in docs, so it would be best to either fill this in now or just take out that line if you plan to fill it out in a later PR.
- `BackfillPolicy.single_run`- I think the docs here could be a little more clear, e.g. "Creates a BackfillPolicy that executes the entire backfill in a single run"

Other than that, this looks good to merge.

ruizh22:
Hi @sryza, I addressed your last comment and can you help me to run it with your CI again? I did run the `black`, `ruff`, and `pyright`. Everything looks good except pyright is complaining about the below error, which I think is unrelated to my changes.
```
.../dagster/python_modules/dagster/dagster_tests/storage_tests/test_upath_io_manager.py:
  465:70: Unnecessary "# type: ignore" comment
```

sryza (approved):
This looks good! Merging.

ruizh22:
@sryza tysm!

PR #15212, opened 2023-07-11
refactor: AssetsDefinition.get_attributes_dict

sryza:
## Summary & Motivation

Modifies the implementations of the `subset_for`, `with_attributes`, and `with_resources` methods of `AssetsDefinition` to rely on a new `get_attributes_dict` method.

This will allow subclasses of `AssetsDefinition` (i.e. `DbtAssetsDefinition`) to override this method to provide additional attributes, instead of having these subclasses override `subset_for`, `with_attributes`, and `with_resources`.

This reduces the number of locations that we need to enumerate all the attributes of `AssetsDefinition`.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/assets.py
@@ -4,6 +4,7 @@ import warnings
 from typing import (
     TYPE_CHECKING,
     AbstractSet,
+    Any,
     Dict,
     Iterable,
     Iterator,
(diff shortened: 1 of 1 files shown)


OwenKephart (approved):
much better -- I wonder if we could just have this subclass NamedTuple and just use the `._replace()` method on it, but we're doing enough funky stuff elsewhere in this class (defining __call__, dagster_internal_init, etc.) that I'm not confident it wouldn't cause problems

PR #15213, opened 2023-07-11
DbtAssetsDefinition and DagsterDbtTranslator

sryza:
## Summary & Motivation

Adds a `DbtAssetsDefinition` class, as described in https://github.com/dagster-io/internal/discussions/6143.

A `DbtAssetsDefinition` is an `AssetsDefinition`, plus:
- A manifest blob
- A `DagsterDbtTranslator`
- A set of convenience methods

Some of the convenience methods have different names than the names on `DbtManifest`. This avoids taking up namespace that could eventually go on `AssetsDefinition` itself, and makes it clear that they're dbt-specific:
- `build_schedule` -> `build_schedule_from_dbt_select`
- `build_asset_selection` -> `build_dbt_asset_selection`

Depends on https://github.com/dagster-io/dagster/pull/15212.

### DIFF
--- a/examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/assets/build.py
+++ b/examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/assets/build.py
@@ -1,11 +1,9 @@
 from dagster import OpExecutionContext
-from dagster_dbt import DbtCli, DbtManifest, dbt_assets
+from dagster_dbt import DbtCli, dbt_assets
 
 from ..constants import MANIFEST_PATH
 
-manifest = DbtManifest.read(path=MANIFEST_PATH)
 
-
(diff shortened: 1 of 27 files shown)


OwenKephart (commented):
Reviving a comment I had re: naming on a PR awhile back, while we're in the "refactor" mode, I'm not a big fan of this `node_info_to_{x}` naming convention. My main gripe is that "node info" is not really a concept in the dbt landscape, so it does not really add anything other than noise to the name of the function.

I think we could instead rename all references of "node info" to something like "dbt node metadata", but `dbt_node_metadata_to_asset_key` would obviously be quite ugly for functions that users would directly interact with, so I'd probably lean more towards something like `get_asset_key(cls, dbt_node_metadata: Mapping[str, Any])`.

OwenKephart (approved):

PR #15218, opened 2023-07-11
Change 'requested' to 'launched' in evaluations ui

johannkm:
"Requested" is a confusing word in the evaluations UI. When we display `1 requested / 1 skipped / 0 discarded` it seems like a partition was requested, but then skipped. It makes sense to think that way because the materialize conditions can be overridden by skips- e.g. I could say that in an evaluation where a materialization condition was met, we "requested" a run but then had that run discarded or skipped.

The two alternatives we've discussed are "materialized" and "launched". The problem with "materialized" is that while we kicked off a materialization, it may fail. The issue with launched is we don't usually use it with assets- on the other hand we still display `Launched run xyz` once you click materialize. I think it's an intuitive enough term and clearer than requested.

Screenshot 2023-07-11 at 8 32 29 AM
Screenshot 2023-07-11 at 8 32 06 AM

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/AutoMaterializePolicyPage/AutomaterializeRequestedPartitionsLink.tsx
+++ b/js_modules/dagit/packages/core/src/assets/AutoMaterializePolicyPage/AutomaterializeRequestedPartitionsLink.tsx
@@ -52,7 +52,7 @@ export const AutomaterializeRequestedPartitionsLink = ({runIds, partitionKeys}: 
 
   const label = React.useMemo(() => {
     if (runIds) {
-      return count === 1 ? '1 partition requested' : `${count} partitions requested`;
+      return count === 1 ? '1 partition launched' : `${count} partitions launched`;
     }
     return count === 1 ? '1 partition' : `${count} partitions`;
   }, [count, runIds]);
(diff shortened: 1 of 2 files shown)


johannkm:
* master
  * **PR #15218** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15218" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15218?utm_source=stack-comment).

braunjj (commented):
LGTM

hellendag (approved):

johannkm:
> Is a run considered "launched" if it's in the queue? If not, should we say "submitted"?

This has been an issue ever since I added the run coordinator. Overall I think 'submitted' was a mistake in that a `launch` button is more clear than a `submit` button, so we've kept it as `launch` when under the hood it actually does a submit. We haven't come up with another way to differentiate the two. Maybe one day submitted will turn in to launched and current day launched (making some api/cli call to start a run worker) will be called 'starting' (to match the run status) or something.

Long story short, we already use `launched` in this context in a few other places and I think we'll continue to do so

sryza:
@johannkm ok sounds good. Easy to change in the future also if for some reason we need to.

PR #15230, opened 2023-07-11
feat(dbt): deprecate `DbtCliClientResource` and associated classes

rexledesma:
## Summary & Motivation
Deprecates the following classes, which were used by the previous dbt resource. We deprecate these classes and utilities in favor of the new `DbtCli` resource.

```python
    from .core import (
        DbtCliClientResource as DbtCliClientResource,
        DbtCliOutput as DbtCliOutput,
        DbtCliResource as DbtCliResource,
        dbt_cli_resource as dbt_cli_resource,
    )
    from .dbt_resource import DbtResource as DbtResource
    from .errors import (
        DagsterDbtCliFatalRuntimeError as DagsterDbtCliFatalRuntimeError,
        DagsterDbtCliHandledRuntimeError as DagsterDbtCliHandledRuntimeError,
        DagsterDbtCliOutputsNotFoundError as DagsterDbtCliOutputsNotFoundError,
    )
    from .types import DbtOutput as DbtOutput
```

We plan to do full deprecation in `0.21.0`.

## How I Tested These Changes
N/A

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/__init__.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/__init__.py
@@ -11,17 +11,7 @@ from .cloud import (
     dbt_cloud_run_op as dbt_cloud_run_op,
     load_assets_from_dbt_cloud_job as load_assets_from_dbt_cloud_job,
 )
-from .core import (
-    DbtCliClientResource as DbtCliClientResource,
-    DbtCliOutput as DbtCliOutput,
-    DbtCliResource as DbtCliResource,
-    dbt_cli_resource as dbt_cli_resource,
-)
-from .dbt_resource import DbtResource as DbtResource
 from .errors import (
-    DagsterDbtCliFatalRuntimeError as DagsterDbtCliFatalRuntimeError,
-    DagsterDbtCliHandledRuntimeError as DagsterDbtCliHandledRuntimeError,
-    DagsterDbtCliOutputsNotFoundError as DagsterDbtCliOutputsNotFoundError,
     DagsterDbtCliRuntimeError as DagsterDbtCliRuntimeError,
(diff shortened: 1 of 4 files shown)


rexledesma:
* master
  * **PR #15230** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15230" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15230?utm_source=stack-comment).

sryza (approved):
My only question is about DagsterDbtCliFatalRuntimeError and DagsterDbtCliHandledRuntimeError. If it was useful to distinguish between these kinds of CLI errors in the past, it seems like it probably will be in the future?

rexledesma:
> My only question is about DagsterDbtCliFatalRuntimeError and DagsterDbtCliHandledRuntimeError. If it was useful to distinguish between these kinds of CLI errors in the past, it seems like it probably will be in the future?

@sryza I believe this goes back to https://github.com/dagster-io/dagster/pull/15055.

Currently, we raise a generic error in either case. If the user wants to handle this failure in a customized way, then they can decide themselves using the underlying `.process.returncode`. 

From https://docs.getdbt.com/reference/exit-codes:

> While these exit codes may change in the future, a zero exit code will always imply success whereas a nonzero exit code will always imply failure.

Because these exit codes aren't contracted, I don't want to create specific objects (e.g. `FatalRuntimeError`/`HandledRuntimeError` for them. We should leave the specifics up to the user after they should know an exception occurred.

sryza:
> Because these exit codes aren't contracted, I don't want to create specific objects (e.g. FatalRuntimeError/HandledRuntimeError for them. We should leave the specifics up to the user after they should know an exception occurred.

Got it - this makes sense to me.

PR #15261, opened 2023-07-13
Fix RunRequest typehint in docstring

Jinior:
## Summary & Motivation

The docstring of the RunRequest class showed the wrong type for the `stale_assets_only` parameter. According to the typehint of the class this should be `bool`. This made the documentation very confusion for me. I changes the type in the documentation to `bool`.


dpeng817:
ah nice catch. Thanks!

dpeng817 (approved):

Issue #14844, opened 2023-06-16
When sensor targets `asset_selection`, don't show `__ASSET_JOB` in UI


sryza:
Here's a sensor that targets an `asset_selection`:

```python
from dagster import sensor, AssetSelection, Definitions, asset


@sensor(asset_selection=AssetSelection.keys("asset1"))
def sensor1():
    ...


@asset
def asset1():
    ...


defs = Definitions(sensors=[sensor1], assets=[asset1])
```

When you look at this in the UI, you see:

<img width="604" alt="image" src="https://github.com/dagster-io/dagster/assets/654855/da4fc709-d23d-446d-a757-4c30c2fa79d6">

and

<img width="373" alt="image" src="https://github.com/dagster-io/dagster/assets/654855/3bccb8f2-38f5-4ac2-95c1-341bbda0b037">

We would ideally show the selection in the UI. If that information isn't available to the frontend, just showing the text "selection of assets" would be an improvement over the status quo.

sryza:
@yuhan - FYI, because you're currently working on extending this to schedules.

bengotow:
Hmm, I was thinking we might be able to do something about this because Sensors have a `metadata.assetKeys`, and we render that in the UI already if it's present (as `Monitored assets`),  but it looks like that's for a separate asset sensor feature where the assets are under observation and not the sensor target. This may not be something we can tackle in the UI/UX bucket until @yuhan has taken a pass at it!

sryza:
@yuhan - is this something that you think you might get to soon?

If not, then I think it would still be an improvement to say "A selection of assets", instead of the current obtuse `__ASSET_JOB`

danielgafni:
It would be great to get rid of `__ASSET_JOB` in general. 
I wonder if we could have unique job names based on the asset selection or something like this.

yuhan:
hey sorry totally missed the pings. here's the pr for the schedule implementation: https://github.com/dagster-io/dagster/pull/15348

yuhan:
hey sorry totally missed the pings. here's the pr for the schedule implementation: https://github.com/dagster-io/dagster/pull/15348 @bengotow @sryza - are tracking the schedule UI changes separately?

PR #15263, opened 2023-07-13
[ui] Hide __ASSET_JOB references on the sensor pages

bengotow:
## Summary & Motivation

This is a quick fix for the issue raised in https://github.com/dagster-io/dagster/issues/14844. We don't have the information about the exact assets in your selection, but we can see that the targets are asset jobs and show "A selection of assets". (Fixes https://github.com/dagster-io/dagster/issues/14844)

I think there are a few other things we should consider changing:

- Backfills have "targets" and sensors have "Asset / Job". I'm tempted to change the labeling in the table + details page to say "Target" to match the backfill terminology.

- We are currently showing `sensor.metadata.assetKeys` in the Asset / Job column. Confusingly these are not the targeted asset keys, they're the monitored asset keys (if the sensor is an @asset_sensor). To make this more clear in the code I wrapped the rendering into a component called `<SensorMonitoredAssets />`, but the placement of them in the table (beneath the "A selection of assets" or job target info) might be confusing to users. I don't have a better placement idea - perhaps we just remove them from the list view. 

image

Loom of this: https://www.loom.com/share/056baeb022fe4c95874656b2a576267b

## How I Tested These Changes

I tested these with a few asset_sensor, op job and asset job sensors

### DIFF
--- a/js_modules/dagit/packages/core/src/sensors/SensorDetails.tsx
+++ b/js_modules/dagit/packages/core/src/sensors/SensorDetails.tsx
@@ -1,7 +1,6 @@
 import {
   Box,
   Button,
-  Group,
   MetadataTableWIP,
   PageHeader,
   Tag,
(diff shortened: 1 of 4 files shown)


js_modules/dagit/packages/core/src/sensors/SensorDetails.tsx
```diff
@@ -70,30 +67,8 @@ export const SensorDetails: React.FC<{
...
-      (pipelineOrJob) => !pipelineOrJob.isJob && targetNames.has(pipelineOrJob.name),
-    );
-  }, [pipelinesAndJobs, targetNames]);
-
-  const pipelineOrJobLabel = React.useMemo(() => {
-    if (anyPipelines) {
-      return targetCount > 1 ? 'Jobs / Pipelines' : 'Pipeline';
-    }
-    return targetCount > 1 ? 'Jobs' : 'Job';
-  }, [anyPipelines, targetCount]);
```
bengotow, line 95: All of this code was determining whether to show Jobs / Pipelines, Pipelines or Job. In the list view we just say `Asset / Job` and I think doing that in the detail view is better for consistency + much simpler 😅

sryza:
+1 to these changes. And I like "Target" over "Asset / Job".

> We are currently showing sensor.metadata.assetKeys in the Asset / Job column. Confusingly these are not the targeted asset keys, they're the monitored asset keys (if the sensor is an @asset_sensor). To make this more clear in the code I wrapped the rendering into a component called <SensorMonitoredAssets />, but the placement of them in the table (beneath the "A selection of assets" or job target info) might be confusing to users. I don't have a better placement idea - perhaps we just remove them from the list view.

I agree that this is confusing. I think we should take that information out of the Asset / Job column, even if it means we just don't show it in the table view at all.

bengotow:
Thanks for the feedback @sryza! I've updated this to use Target instead of Asset / Job and removed the "Monitored Assets" display from the list view. I think it was definitely confusing in the target column and doesn't seem worth adding another column and crunching down all the existing data

hellendag (approved):

PR #15268, opened 2023-07-13
Remove old deprecated asset_key args/methods

smackesey:
## Summary & Motivation

Removes various deprecated APIs before 1.4:

- `ExecuteJobResult` alias for `JobExecutionResult`
- `asset_key(s)` argument/property on `AssetIn`, `AssetsDefinition`
- `output_asset_partition_key` and `output_asset_partitions_time_window` on `OpExecutionContext`

Also bumps breaking version on two older metadata APIs to 2.0.0.

## How I Tested These Changes

Existing test suite.

PR #15269, opened 2023-07-13
remove deprecated UnresolvedAssetJob.resolve args

smackesey:
## Summary & Motivation

The `assets` and `source_assets` args to `UnresolvedAssetJob.resolve` were deprecated for removal in 1.3. This PR removes them.

It also adds support for passing `Definitions` to `UnresolvedAssetJob.resolve`-- this opens the door to making the method public.

## How I Tested These Changes

New unit test for passing `Definitions` to `UnresolvedAssetJob.resolve`.

### DIFF
--- a/examples/assets_pandas_type_metadata/assets_pandas_type_metadata_tests/test_bollinger_analysis.py
+++ b/examples/assets_pandas_type_metadata/assets_pandas_type_metadata_tests/test_bollinger_analysis.py
@@ -1,4 +1,5 @@
-from dagster import AssetSelection, define_asset_job, with_resources
+from dagster import AssetSelection, define_asset_job
+from dagster._core.definitions.definitions_class import Definitions
 
 from assets_pandas_type_metadata.assets.bollinger_analysis import (
     sp500_anomalous_events,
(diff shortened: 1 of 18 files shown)


smackesey:
* master
  * **PR #15268** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15268" target="_blank">Graphite</a> 
    * **PR #15269** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15269" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15269?utm_source=stack-comment).

sryza (commented):
I think it's definitely within our rights to get rid of this. However, I think there are a set of users who rely on it, and we don't really have a good alternative.

One way to provide an alternative would be to allow the `asset_graph` argument to accept a `Definitions` argument.

smackesey:
>we don't really have a good alternative.

Is the issue that `AssetGraph` isn't public?

sryza:
> Is the issue that AssetGraph isn't public?

Right.

smackesey:
I added support for passing `Definitions`.

sryza:
Hey so I got the chance to think about this a little bit more. I think the end state that we want to end up with is that `resolve` accepts only an `AssetGraph`, and `AssetGraph` is public. To get there, we need to make `AssetGraph`, which will require a product review, but otherwise I think should happen soon.

I don't think that accepting `Definitions` is a great final state for this API. What makes it weird is that `Definitions` can contain jobs, so resolving jobs with an object that also contains jobs (maybe the same job?) feels weird.

If `Definitions` has an `asset_graph` property, then I think we don't need `resolve` to accept a `Definitions` object.  I think we'd leave that property non-public for now, but informally encourage people to use it if they were using `resolve`.

Sorry for the thrash on this one.

smackesey:
So the PR in it's current state added `get_asset_graph` to the `Definitions` object (not a property since all the other methods are `get_*` on `Definitions`). If I remove the added `Definitions` support from `.resolve` will this then be good to merge in your mind?

sryza:
> If I remove the added Definitions support from .resolve will this then be good to merge in your mind?

Right

smackesey:
done

sryza (approved):

PR #15277, opened 2023-07-13
fix(dbt): add tests for loading scaffolded Dagster project

rexledesma:
## Summary & Motivation
After scaffolding the project:
- Compile the manifest associated with the dbt project
- Load the Dagster definitions from the scaffolded project 
- Assert that the desired definitions are present and run successfully

Along the way, fix errors in the scaffold resulting from API names changing around.

## How I Tested These Changes
pytest

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt_tests/cli/test_app.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt_tests/cli/test_app.py
@@ -1,7 +1,11 @@
+import importlib
 import shutil
+import subprocess
+import sys
 from pathlib import Path
 
 import pytest
+from dagster import Definitions
 from dagster_dbt.cli.app import app
 from typer.testing import CliRunner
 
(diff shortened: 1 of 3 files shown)


rexledesma:
* master
  * **PR #15277** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15277" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15277?utm_source=stack-comment).

python_modules/libraries/dagster-dbt/dagster_dbt_tests/cli/test_app.py
```diff
@@ -47,6 +57,20 @@ def test_project_scaffold_command(monkeypatch: pytest.MonkeyPatch, tmp_project_d
...
     assert not any(path.suffix == ".jinja" for path in dagster_project_dir.glob("**/*"))
 
+    subprocess.run(["dbt", "compile"], cwd=dbt_project_dir, check=True)
+
+    assert dbt_project_dir.joinpath("target", "manifest.json").exists()
+
+    monkeypatch.chdir(tmp_path)
+    sys.path.append(tmp_path.as_posix())
+
+    defs: "Definitions" = getattr(
```
sryza, line 67: I'm surprised this type annotation would need quotes

sryza (approved):

python_modules/libraries/dagster-dbt/dagster_dbt_tests/cli/test_app.py
rexledesma: This is because `Definitions` is a conditional import since a `Definitions` object isn't being constructed and is only used for type hints.

sryza: why does it need to be conditional though? I think it's fairly common to import something just to use for type hints, and the main reason for conditional is to avoid circular deps

btw this is non-blocking feedback

rexledesma: Ah, my rule was that if the object isn't being used in the file, I usually enclose it in the `TYPE_CHECKING` guard, even if it isn't an expensive import.

I guess the loss in legibility isn't worth it. will change

rexledesma: @sryza Funny enough, @smackesey is working on lint checks to automate this: https://github.com/dagster-io/dagster/pull/14675. Imports only used for type checks are now automatically under a `TYPE_CHECKING` conditional.

Issue #15073, opened 2023-07-01
visualize freshness policies in the UI


sryza:
An asset is considered overdue when it's derived from data that's older than the required time for it to be fresh. An asset's data time is based on materialization times of the ancestor materializations of its latest materialization.

All of this information is currently opaque to the user, which makes it difficult to understand the semantics of freshness.  We should expose it.  I.e. you should be able to click or hover or something over a freshness policy and see:
- Its "data time", i.e. how old Dagster thinks its data is
- What its data time needs to be for it not to be considered overdue
- How its data time was computed

One fancy way to represent how its data time was computed would be a DAG of the materializations that the asset is based on.  Each materialization would show its data time, and materializations with data times prior to the expected data time would be colored red.

This materialization lineage graph might also be generally useful for debugging, even when no freshness policy is set.

This would help to resolve confusion like this: https://github.com/dagster-io/internal/discussions/5928#discussioncomment-6328060.

bengotow:
Hey @sryza  - this seems like a great idea. I believe we /sorta/ have a way to inspect this already, in the form of this "Source Data"  rendering that pulls from `AssetNode.assetMaterializationUsedData(timestamp)`:

<img width="1041" alt="image" src="https://github.com/dagster-io/dagster/assets/1037212/096543cf-0c46-4a13-9bfc-9c79500a7fd0">

Right now we collapse it (only showing the furthest ancestors) and the times are relative and not absolute, but I think the data it's fetching is the data we want for this.

I'll take a stab at presenting this in an overlay that you can open from the "Late" tag and colorizing it to highlight the source of lateness.

sryza:
@bengotow awesome - excited to see what you come up with. Let me know if it would be helpful to bounce around ideas.

PR #15279, opened 2023-07-13
[ui] Show popover context to help users debug “overdue” frehness policies

bengotow:
## Summary & Motivation

Fixes: https://github.com/dagster-io/dagster/issues/15073
Loom: https://www.loom.com/share/8552264fadc64d3abf234bef144886f9

image

- The "overdue" tags and "overdue" labels on the asset graph now have a large, informative popover that includes both the last materialization and the materializations upstream that were incorporated into that materialization.

- The relative times in the popover help you understand how / which pieces are in violation of the maximumLagMinutes on your freshness policy, and turn red when they're too old.

Other cleanup:

- I renamed CurrentMinutesLateTag to OverdueTag for consistency with the new terminology

- I added a small set-check in the render function of `AssetMaterializationUpstreamData` because we had a production issue caused by the backend sending bad data and causing an infinitely recursive render.

## How I Tested These Changes

I added a file of stories covering all the states of this tag (largely using our existing fixtures!) I think it may be worth adding some jest tests, but it's also seems likely this will change a bit more this quarter.

image


js_modules/dagit/packages/core/src/assets/__stories__/OverdueTag.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+              assetKey: buildAssetKey({
+                path: ['inp2'],
+              }),
+              downstreamAssetKey: buildAssetKey({
+                path: ['inp8'],
+              }),
+              __typename: 'MaterializationUpstreamDataVersion',
+            },
+          ],
+          __typename: 'AssetNode',
```
salazarm, line 64: I think you remove the __typename here since you're using the builder

hellendag (approved):

bengotow:
There was a discussion of this UI in https://elementl-workspace.slack.com/archives/C03CA4TVCAW/p1689274462013299 and I'm making two follow-up changes:

- Update language: "The latest materialization is derived from source data that's X minutes older than 2:30 AM UTC, but the asset's freshness policy requires it to be derived from data that's no more than Y minutes older." 

- For freshness policies with a cron schedule, the relative times shown in red need to be relative to the last tick time, and we need a GraphQL API of some sort for retrieving the timestamp of that tick. Will investigate this.

js_modules/dagit/packages/core/src/assets/__stories__/OverdueTag.stories.tsx
bengotow: Nice! Good call

js_modules/dagit/packages/core/src/assets/OverdueTag.tsx
```diff
@@ -0,0 +1,227 @@
...
+  const lagDesc =
+    maximumLagMinutes % 30 === 0
+      ? `${maximumLagMinutes / 60} hour${maximumLagMinutes / 60 !== 1 ? 's' : ''}`
+      : `${maximumLagMinutes} min`;
+
+  if (format === 'short') {
+    if (cronDesc) {
+      return `no more than ${lagDesc} old by${nbsp}${cronDesc}.`;
+    } else {
+      return `no more than ${lagDesc} old at any${nbsp}time.`;
```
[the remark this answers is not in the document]
bengotow: Ahh this is a bit awkward but it's to make sure that the popover never wraps just the word `time` to the last line since the description varies in length. There might be a more generic way we could do this for all tooltips actually 🤔

OwenKephart (approved):

sryza:
Awesome to have this in.

@bengotow a couple of pieces of feedback:
- Visually, it feels like the popup could have a little extra margin on the bottom, and maybe something else to help visually distinguish it from the UI behind it?
image
- Thoughts on displaying something like this in the Fresh case as well?
- I think it might be clearer to leave out the latest materialization timestamp, because it's not actually used in the calculation

bengotow:
Thanks @sryza! I agree re: margin and general shadowing to make the popover "pop" a bit more. Let me ping @braunjj and see what he thinks because I think we'd want to make the styling change other places as well.

I think we could display this in the fresh case actually, let me see if there's anything blocking that and i'll put up a small follow-up!

PR #15284, opened 2023-07-13
[dagster-dbt] rename translator methods 

OwenKephart:
## Summary & Motivation

Renamed some symbols

### DIFF
--- a/examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/assets/customize_asset_keys.py
+++ b/examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/assets/customize_asset_keys.py
@@ -8,8 +8,8 @@ from ..constants import MANIFEST_PATH
 
 class CustomDagsterDbtTranslator(DagsterDbtTranslator):
     @classmethod
-    def get_asset_key(cls, dbt_repsource_props: Mapping[str, Any]) -> AssetKey:
-        return AssetKey(["prefix", dbt_repsource_props["alias"]])
+    def get_asset_key(cls, dbt_resource_props: Mapping[str, Any]) -> AssetKey:
+        return AssetKey(["prefix", dbt_resource_props["alias"]])
 
 
 @dbt_assets(manifest=MANIFEST_PATH, dagster_dbt_translator=CustomDagsterDbtTranslator())
--- a/examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/assets/customize_asset_keys.py
+++ b/examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/assets/customize_asset_keys.py
@@ -8,8 +8,8 @@ from ..constants import MANIFEST_PATH
 
 class CustomDagsterDbtTranslator(DagsterDbtTranslator):
     @classmethod
-    def get_asset_key(cls, dbt_resource_info: Mapping[str, Any]) -> AssetKey:
-        return AssetKey(["prefix", dbt_resource_info["alias"]])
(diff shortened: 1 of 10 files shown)


python_modules/libraries/dagster-dbt/dagster_dbt/dagster_dbt_translator.py
```diff
@@ -11,17 +11,17 @@
 
 class DagsterDbtTranslator:
     @classmethod
-    def node_info_to_asset_key(cls, node_info: Mapping[str, Any]) -> AssetKey:
-        """A function that takes a dictionary representing the dbt node and returns the
-        Dagster asset key the represents the dbt node.
+    def get_asset_key(cls, dbt_resource_info: Mapping[str, Any]) -> AssetKey:
```
sryza, line 14: I think it's worth adding a sentence that explicitly calls out that a dbt resource is different than a Dagster resource.

rexledesma: yeah, we should update the API docstrings here.

also maybe consider dbt_resource_props, since dbt resource info in google yields therapy as the first search result, and i don't want to compete against that when we have documentation on these concepts soon

python_modules/libraries/dagster-dbt/dagster_dbt/dagster_dbt_translator.py
```diff
@@ -34,23 +34,23 @@ def node_info_to_asset_key(cls, node_info: Mapping[str, Any]) -> AssetKey:
 
                 class CustomDagsterDbtTranslator(DagsterDbtTranslator):
                     @classmethod
-                    def node_info_to_asset_key(cls, node_info: Mapping[str, Any]) -> AssetKey:
+                    def get_asset_key(cls, dbt_resource_info: Mapping[str, Any]) -> AssetKey:
```
rexledesma, line 37: @clairelin135 introduced a with_prefix method on AssetKey that simplifies things here. we should use it

rexledesma (approved):
🙏

examples/experimental/tutorial_dbt_dagster_v2/tutorial_dbt_dagster_v2/assets/customize_asset_keys.py
```diff
@@ -8,8 +8,8 @@
 
 class CustomDagsterDbtTranslator(DagsterDbtTranslator):
     @classmethod
-    def node_info_to_asset_key(cls, node_info: Mapping[str, Any]) -> AssetKey:
-        return AssetKey(["prefix", node_info["alias"]])
+    def get_asset_key(cls, dbt_resource_info: Mapping[str, Any]) -> AssetKey:
```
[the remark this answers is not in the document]
OwenKephart: Yeah, I looked into it and `properties` seems to be the dbt-native word for this set of things. Updating

python_modules/libraries/dagster-dbt/dagster_dbt/dagster_dbt_translator.py
```diff
@@ -11,7 +11,7 @@
 
 class DagsterDbtTranslator:
     @classmethod
-    def get_asset_key(cls, dbt_resource_info: Mapping[str, Any]) -> AssetKey:
+    def get_asset_key(cls, dbt_repsource_props: Mapping[str, Any]) -> AssetKey:
```
rexledesma, line 14: Assume this is a typo - could we also update the docstrings to use `properties` instead of `information` and also give a durable link to the dbt manifest API (https://docs.getdbt.com/reference/artifacts/manifest-json) when documenting the argument?

rexledesma (approved):

PR #15287, opened 2023-07-13
[docs] API docs for AssetsDefinition and OpDefinition

OwenKephart:
## Summary & Motivation

https://github.com/dagster-io/dagster/pull/15078

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/assets.py
+++ b/python_modules/dagster/dagster/_core/definitions/assets.py
@@ -626,7 +626,8 @@ class AssetsDefinition(ResourceAddable, IHasInternalInit):
     @property
     def group_names_by_key(self) -> Mapping[AssetKey, str]:
         """Mapping[AssetKey, str]: Returns a mapping from the asset keys in this AssetsDefinition
-        to the group names assigned to them (if any).
+        to the group names assigned to them. If there is no assigned group name for a given AssetKey,
+        it will not be present in this dictionary.
         """
         return self._group_names_by_key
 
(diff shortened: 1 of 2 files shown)


python_modules/dagster/dagster/_core/definitions/assets.py
```diff
@@ -616,21 +616,34 @@ def _from_node(
...
+        """bool: If True, indicates that this AssetsDefinition may materialize any subset of its
+        asset keys in a given computation (as opposed to being required to materialize all asset
+        keys).
+        """
         return self._can_subset
 
     @public
     @property
     def group_names_by_key(self) -> Mapping[AssetKey, str]:
+        """Mapping[AssetKey, str]: Returns a mapping from the asset keys in this AssetsDefinition
```
smackesey, line 628: [1] This docstring suggests the value type here should be `Optional[str]`-- should it be?

python_modules/dagster/dagster/_core/definitions/assets.py
```diff
@@ -616,21 +616,34 @@ def _from_node(
...
+        to the group names assigned to them (if any).
+        """
         return self._group_names_by_key
 
     @public
     @property
     def descriptions_by_key(self) -> Mapping[AssetKey, str]:
+        """Mapping[AssetKey, str]: Returns a mapping from the asset keys in this AssetsDefinition
```
smackesey, line 636: [1]

python_modules/dagster/dagster/_core/definitions/assets.py
```diff
@@ -670,16 +692,23 @@ def key(self) -> AssetKey:
...
     def keys(self) -> AbstractSet[AssetKey]:
+        """AbstractSet[AssetKey]: The asset keys associated with this AssetsDefinition."""
         return self._selected_asset_keys
 
     @public
     @property
     def dependency_keys(self) -> Iterable[AssetKey]:
+        """Iterable[AssetKey]: The asset keys which are upstream of assets associated with this
+        AssetsDefinition.
+        """
```
smackesey, line 711: nit: I think it could be a bit clearer-- maybe "upstream of any asset included in this `AssetsDefinition`"

smackesey (approved):
Few comments but otherwise nice

python_modules/dagster/dagster/_core/definitions/assets.py
OwenKephart: Yeah it's a bit weird (if there's no group name, then there's no entry in this dictionary)

PR #15297, opened 2023-07-14
accept Path in load_assets_from_dbt_manifest

sryza:

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
@@ -1,6 +1,7 @@
 import hashlib
 import json
 import os
+from pathlib import Path
 from typing import (
     AbstractSet,
     Any,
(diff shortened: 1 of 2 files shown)


python_modules/libraries/dagster-dbt/dagster_dbt/asset_defs.py
```diff
@@ -639,7 +640,7 @@ def load_assets_from_dbt_project(
 
 
 def load_assets_from_dbt_manifest(
-    manifest: Optional[Mapping[str, Any]] = None,
+    manifest: Optional[Union[Path, Mapping[str, Any]]] = None,
```
rexledesma, line 643: Use https://docs.python.org/3/library/os.html#os.PathLike so we can support string paths as well
```suggestion
    manifest: Optional[Union[os.PathLike, Mapping[str, Any]]] = None,
```

rexledesma (approved):
Could also take in `str` if we use `os.PathLike`

sryza:
> Could also take in str if we use os.PathLike

I'm seeing `isinstance("fdshkl", os.PathLike)` returns False. Is that what you'd expect?

sryza:
Going to merge this as-is and we can do a followup to expand to PathLike for both this method and dbt_assets later if we choose

PR #15302, opened 2023-07-15
Incorporate warnings in @deprecated decorator

smackesey:


smackesey:
* master
  * **PR #15302** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15302" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15302?utm_source=stack-comment).

python_modules/dagster/dagster/_core/decorator_utils.py
```diff
@@ -128,3 +147,151 @@ def has_at_least_one_parameter(
...
+    new_fn = _wrap_with_pre_call_fn(target, pre_call_fn)
+    return _update_decoratable(obj, new_fn)
+
+
+def _wrap_with_pre_call_fn(
+    fn: T_Callable,
+    pre_call_fn: Callable[[], None],
+) -> T_Callable:
+    @functools.wraps(fn)
+    def new_fn(*args, **kwargs):
```
alangenfeld, line 193: does `new_fn` show up in stack traces? may want a slightly more useful name then - its nice to be able to quickly at a glance understand what some random shit in the calls stack is there for 

`wrapped_with_pre_call_fn` or something maybe

python_modules/dagster/dagster/_core/decorator_utils.py
```diff
@@ -128,3 +147,151 @@ def has_at_least_one_parameter(
...
+    we stored on the `fget` (since we can't and therefore don't store them on the `property`
+     object).
+    """
+    from dagster._annotations import copy_annotations
+
+    curr_fn = get_decorator_target(decoratable)
+    new_fn = functools.update_wrapper(new_fn, curr_fn)
+    # from types import FunctionType
+    # if inspect.isfunction(decoratable) or isinstance(decoratable, FunctionType):
+    #     val = new_fn
```
alangenfeld, line 235: ?

python_modules/dagster/dagster/_core/decorator_utils.py
```diff
@@ -128,3 +147,151 @@ def has_at_least_one_parameter(
...
+        val = decoratable
+    elif is_resource_def(decoratable):
+        setattr(decoratable, "_resource_fn", new_fn)
+        val = decoratable
+    elif is_op_def(decoratable):
+        if is_decorated_op_function(decoratable.compute_fn):
+            setattr(decoratable.compute_fn, "decorated_fn", new_fn)
+        else:
+            setattr(decoratable, "compute_fn", new_fn)
+        val = decoratable
```
alangenfeld, line 257: yikes, why do we need special casing for op and resource?

python_modules/dagster/dagster/_core/decorator_utils.py
```diff
@@ -128,3 +147,151 @@ def has_at_least_one_parameter(
...
+    return hasattr(obj, "_compute_fn")
+
+
+def is_decorated_op_function(
+    obj: Union[Callable, "DecoratedOpFunction"]
+) -> TypeGuard["DecoratedOpFunction"]:
+    """Use this in place of `isinstance` when `DecoratedOpFunction`
+    can't be imported yet (due to circular import issues).
+    """
+    return hasattr(obj, "decorated_fn")
```
alangenfeld, line 297: more yikes, these are quite spicy

python_modules/dagster/dagster/_core/decorator_utils.py
smackesey: yes it does, good call

python_modules/dagster/dagster/_core/decorator_utils.py
smackesey: removed

python_modules/dagster/dagster/_core/decorator_utils.py
smackesey: Because these are objects that store the function they wrap under a special attribute. In this they are similar to `property` (which uses `fget`, and `staticmethod`/`classmethod` (which use `__func__`). If you want to be able to do e.g. this:

```
@deprecated(breaking_version="0.21", additional_warn_text="Use DbtCli instead.")
@dagster_maintained_resource
@resource(config_schema=DbtCliClientResource.to_config_schema())
def dbt_cli_resource(context) -> DbtCliClient:
    """This resource issues dbt CLI commands against a configured dbt project."""
```

You need to know how to get at the function the `ResourceDefinition` is wrapping.

alangenfeld: are these covered in the test suite at all?

python_modules/dagster/dagster/_core/decorator_utils.py
alangenfeld: can we confirm against the `__name__` or something too? Should at least be under direct explicit test since someone refactoring implementations can break this

smackesey: Yeah these are rough but this is a tricky situation. All the annotation code is being executed at import time, which means that if we used e.g `isinstance(obj, ResourceDefinition)`, it would get executed the first time an annotation gets encountered, which is before `ResourceDefinition`. I tried a local import of `ResourceDefinition` but it led to a tangled mess of circular import errors.

While I recognize this is ugly and not an approach I would use in a more general context, these functions are _only_ called when we've decorated something with one of these special decorators, and IMO it is OK in this limited context.

python_modules/dagster/dagster/_core/decorator_utils.py
smackesey: Yes, see `test_deprecated_resource` and `test_experimental_resource`.

python_modules/dagster/dagster/_core/decorator_utils.py
smackesey: I updated the implementation to check the `mro` of `obj.__class__` for `"ResourceDefinition"`

alangenfeld (approved):
alright alright alright

PR #15307, opened 2023-07-16
Fix description of DuckDB

chodera:
## Summary & Motivation

DuckDB is not an in-memory database, but an in-process database.


clairelin135:
@chodera this fix is looking good to me, thanks for adding. Will merge assuming tests pass!

PR #15317, opened 2023-07-17
[auto-materialize][refactor] Move "asset_reconciliation_sensor" logic

OwenKephart:
## Summary & Motivation

High level, this creates a single object (AssetDaemonContext) which is responsible for managing the lifecycle of a single tick of the AssetDaemon. This has some benefits:

- Previously, we needed to pass around a bunch of values (newly_materialized_root_asset_partitions, etc.) through a bunch of different functions in order to make sure that we properly updated the AssetDaemonCursor (freshly renamed) at the end of the tick. Now, we just update the cursor as soon as we get a relevant value.
- Previously, we needed to define a big stack of functions within another function in order to make sure they had access to relevant local variables. Now we can just access properties of the class (e.g. building up the conditions_by_asset_partitions dictionary as we go)

Lower level changes:
- Move the logic to get asset partitions with newly updated parents to the CachingInstanceQueryer (as it's used by both the AssetDaemon and asset backfills, it makes more sense to be in that shared spot)
  - Goes along with small refactor to break it up into two functions 
- Move AutoMaterializeAssetEvaluation class to auto_materialize_conditions.py, as it's really just a collection of conditions so it makes more sense there.
- Rename AssetReconciliationCursor to AssetDaemonCursor, and move it to the new asset_daemon_context.py file
- Copy-paste all the logical stuff from the asset_reconciliation_sensor.py file to asset_daemon_context.py
- Update some of those functions to take advantage of the simplicity offered by doing things in a class (i.e. accessing class variables rather than local variables, updating the cursor at relevant times).

This lays the groundwork for a further pass of simplification and refactoring, centered around making the auto materialize logic get evaluated in a single pass through the asset graph. However, I wanted to decouple that larger conceptual change from this more banal stuff.


OwenKephart:
"AssetDaemonContext" is very probably not the right name, as we already use "Context" a lot elsewhere and this is somewhat different. Some other options:
- AssetDaemonEvaluation
- AssetDaemonIteration
- AssetDaemonTick
- AssetDaemonTickEvaluator
- AssetDaemonEvaluator

sryza (commented):
I like the AssetDaemonCursor rename. What do you think about putting it into its own file?

A reaction I have here is that AssetDaemonContext is a bit of a [god object](https://en.wikipedia.org/wiki/God_object), and that this approach is more stateful than the earlier code, which can make it more difficult to reason about.  It's annoying to pass the never_materialized stuff through a bunch of places, but it makes it very clear what data is flowing where. It seems like refactor is kind of saying "there's enough cross-talk that we're just going to allow anything within an asset daemon tick to access anything else", and I think that's a dangerous pattern.

I wonder if there's a way to consolidate some of the stuff that's getting passed around without needing to have a class with mutable state?

OwenKephart:
@sryza Yeah great point -- definitely reworkable

python_modules/dagster/dagster/_core/definitions/asset_daemon_context.py
```diff
@@ -0,0 +1,450 @@
...
+    )
+
+
+class AssetDaemonContext:
+    def __init__(
+        self,
+        instance: "DagsterInstance",
+        asset_graph: AssetGraph,
+        stored_cursor: AssetDaemonCursor,
+        target_asset_keys: Optional[AbstractSet[AssetKey]] = None,
```
johannkm, line 67: I can do the diff to remove this and migrate the tests

OwenKephart: that would be awesome, ty 🙏

OwenKephart:
closing this and breaking it up into smaller more reviewable chunks

Issue #15233, opened 2023-07-11
Don't require manifest argument to `DbtCli.cli`, even if context is not provided


sryza:
I believe that in most cases the cli will produce a manifest, which we can use if the argument is not provided

rexledesma:
Resolved with https://github.com/dagster-io/dagster/pull/15185

PR #15318, opened 2023-07-17
feat(dbt): allow runtime compilation of manifest in Dagster scaffold

rexledesma:
## Summary & Motivation
We essentially provide a workflow that emulates the behavior of `load_assets_from_dbt_project`. In this workflow, the dbt manifest is created at runtime, and is piped into `@dbt_assets`.

Users can invoke this flow when developing by doing:

```bash
DAGSTER_DBT_PARSE_PROJECT_ON_LOAD=1 dagster dev
```

## How I Tested These Changes
pytest, local

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/include/scaffold/definitions.py.jinja
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/include/scaffold/definitions.py.jinja
@@ -5,8 +5,16 @@ from dagster import Definitions, OpExecutionContext
 from dagster_dbt import DbtCliResource, dbt_assets
 
 dbt_project_dir = Path(__file__).parent.joinpath({{ dbt_project_dir_relative_path_parts | join(', ')}})
-dbt_manifest_path = dbt_project_dir.joinpath("target", "manifest.json")
+dbt = DbtCliResource(project_dir=os.fspath(dbt_project_dir))
 
+# If DAGSTER_DBT_PARSE_PROJECT_ON_LOAD is set, a manifest will be created at runtime.
+# Otherwise, we expect a manifest to be present in the project's target directory.
+if os.getenv("DAGSTER_DBT_PARSE_PROJECT_ON_LOAD"):
+    dbt_parse_invocation = dbt.cli([{{ dbt_parse_command | join(', ')}}], manifest={})
+    dbt_parse_invocation.wait()
+    dbt_manifest_path = dbt_parse_invocation.target_path.joinpath("manifest.json")
+else:
+    dbt_manifest_path = dbt_project_dir.joinpath("target", "manifest.json")
 
 @dbt_assets(manifest=dbt_manifest_path)
 def {{ dbt_assets_name }}(context: OpExecutionContext, dbt: DbtCliResource):
@@ -25,8 +33,6 @@ defs = Definitions(
     assets=[{{ dbt_assets_name }}],
     schedules=schedules,
     resources={
-        "dbt": DbtCliResource(
-            project_dir=os.fspath(dbt_project_dir),
-        ),
+        "dbt": dbt,
     },
 )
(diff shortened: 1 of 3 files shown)


rexledesma:
* master
  * **PR #15318** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15318" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15318?utm_source=stack-comment).

python_modules/libraries/dagster-dbt/dagster_dbt/include/scaffold/definitions.py.jinja
```diff
@@ -7,6 +7,14 @@ from dagster_dbt import DbtCli, dbt_assets
 dbt_project_dir = Path(__file__).parent.joinpath({{ dbt_project_dir_relative_path_parts | join(', ')}})
 dbt_manifest_path = dbt_project_dir.joinpath("target", "manifest.json")
 
+dbt = DbtCli(project_dir=os.fspath(dbt_project_dir))
+
+if os.getenv("DAGSTER_DBT_BUILD_PROJECT"):
+    dbt_parse_invocation = dbt.cli(["parse"], manifest={})
+    dbt_parse_invocation.wait()
+    dbt_parse_invocation.target_path.joinpath("manifest.json").replace(
```
sryza, line 15: What exactly is this doing / for?

python_modules/libraries/dagster-dbt/dagster_dbt/include/scaffold/definitions.py.jinja
```diff
@@ -7,6 +7,14 @@ from dagster_dbt import DbtCli, dbt_assets
 dbt_project_dir = Path(__file__).parent.joinpath({{ dbt_project_dir_relative_path_parts | join(', ')}})
 dbt_manifest_path = dbt_project_dir.joinpath("target", "manifest.json")
 
+dbt = DbtCli(project_dir=os.fspath(dbt_project_dir))
+
+if os.getenv("DAGSTER_DBT_BUILD_PROJECT"):
+    dbt_parse_invocation = dbt.cli(["parse"], manifest={})
```
sryza, line 13: We theoretically don't need the manifest argument here, right? I filed https://github.com/dagster-io/dagster/issues/15233 to track.

python_modules/libraries/dagster-dbt/dagster_dbt_tests/cli/test_app.py
```diff
@@ -77,6 +77,54 @@ def test_project_scaffold_command(
     assert materialize_dbt_models_schedule.cron_schedule == "0 0 * * *"
 
 
+def test_project_scaffold_command_with_runtime_manifest(
```
sryza, line 80: Nice test

python_modules/libraries/dagster-dbt/dagster_dbt/include/scaffold/definitions.py.jinja
rexledesma: We are moving the manifest that's generated in the dynamic target path of a `DbtCli.cli` invocation (e.g. `target-123213`) to the static target path of the dbt project `target/`.

sryza: Gotcha. That sounds correct, but the dataflow is a little hard to follow, and it overwrites the manifest at the static target path which may or may not be desirable. What if we made something like this possible?

```python
if os.getenv("DAGSTER_DBT_BUILD_PROJECT"):
    dbt_manifest_path = dbt.cli(["parse"]).wait().get_output_manifest_path()
else:
    dbt_manifest_path = dbt_project_dir.joinpath("target", "manifest.json")
```

PR #15320, opened 2023-07-17
[dagit] Frontload "ops" and "resources" keys when autoloading config

benpankow:
## Summary

Reorder the autoloaded default config keys to put ops and resources at the top, since users most often would care about these.

## Test Plan

Tested locally.

### DIFF
--- a/python_modules/dagster-graphql/dagster_graphql_tests/graphql/__snapshots__/test_environment_schema.ambr
+++ b/python_modules/dagster-graphql/dagster_graphql_tests/graphql/__snapshots__/test_environment_schema.ambr
@@ -47,17 +47,17 @@
     'runConfigSchemaOrError': dict({
       '__typename': 'RunConfigSchema',
       'rootDefaultYaml': '''
-        execution:
-          config:
-            retries:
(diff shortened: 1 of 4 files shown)


benpankow:
* master
  * **PR #15320** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15320" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15320?utm_source=stack-comment).

yuhan:
where is this idea coming from? implementation wise it's pretty straightforward, but i'm a little worried whether it's worth the user value while this could add internal complexity as it hardcodes `ops` and `resources` - e.g. if we later introduce an `assets`, how do we make sure the future developer knows to update this logic here.

yuhan (approved):
got it. the huge execution default block does seem overwhelming. this looks good.

benpankow:
@benpankow started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15320).

PR #15321, opened 2023-07-17
[rfc] accept list of lists in deps

jamiedemaria:
## Summary & Motivation

@slopp  mentioned that users sometimes struggle with list concatenation and that the `deps` param may cause some issues with this. Specifically

```python
my_other_assets = load_assets_from_modules("my_other_module") # my_other_assets is a list

@asset(
   deps=[my_other_assets, some_other_asset]
)
def depends_on_stuff():
   ...
```

will fail, because `my_other_assets` is a list. Instead you need to do 

```python
my_other_assets = load_assets_from_modules("my_other_module") # my_other_assets is a list

@asset(
   deps=[*my_other_assets, some_other_asset]
)
def depends_on_stuff():
   ...
```

This RFC proposes allowing `deps` to accept a list as an element of the outer list. We then flatten this list internally.

This presents a funky edge case because one of the `CoercibleToAssetKey` types is a list of strings, so that means we need to support this case

```python
@asset
def asset_1():
    ...

@asset
def asset_2():
   ....

my_assets = [asset_1, asset_2]

@asset(
   deps=[my_assets, ["key_prefix", "another_asset"]]  # needs to be flattened to [asset_1, asset_2, ["key_prefix", "another_asset"]]
)
def downstream_asset():
    ...
```

This definitely adds a significant amount of internal complexity for us to handle for a small-ish (imo) benefit for the user. Interested to get other folk's takes on if this is worth it.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
+++ b/python_modules/dagster/dagster/_core/definitions/decorators/asset_decorator.py
@@ -5,11 +5,14 @@ from typing import (
     Any,
     Callable,
     Dict,
+    Iterator,
+    List,
     Mapping,
     Optional,
     Sequence,
     Set,
     Tuple,
+    TypeVar,
     Union,
     cast,
     overload,
@@ -46,6 +49,8 @@ from ..partition import PartitionsDefinition
 from ..policy import RetryPolicy
 from ..resource_definition import ResourceDefinition
 from ..utils import DEFAULT_IO_MANAGER_KEY, NoValueSentinel
+
+T = TypeVar("T")
 
 
 @overload
@@ -1022,11 +1027,23 @@ def _type_check_deps_and_non_argument_deps(
             "Cannot specify both deps and non_argument_deps to @asset. Use only deps instead."
         )
 
+    def _flatten(items: Sequence) -> Iterator:
+        # this is to add support for a list of AssetsDefinitions or SourceAssets to the deps parameter
+        for x in items:
+            if isinstance(x, List):
+                if isinstance(x[0], str):
+                    # ["my_prefix", "my_asset_key"] needs to stay a list
+                    yield x
+                yield from x
+            else:
+                yield x
+
     upstream_asset_deps: Optional[
         Sequence[Union[CoercibleToAssetKey, AssetsDefinition, SourceAsset]]
-    ] = None
+    ] = []
     if deps is not None:
-        for dep in deps:
+        flattened_deps = _flatten(deps)
+        for dep in flattened_deps:
             if isinstance(dep, AssetsDefinition):
                 # Only AssetsDefinition with a single asset can be passed
                 if len(dep.keys) > 1:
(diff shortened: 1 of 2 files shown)


jamiedemaria:
* master
  * **PR #15321** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15321" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15321?utm_source=stack-comment).

yuhan:
I'm personally not in favor of this change because it loosens the typing constraint, and therefore would add internal complexity and could also make our educational content hard to explain.

My main reservation is in tutorial/guide/apidoc, it becomes unintuitive to explain what exactly `deps` would take, and it could confuse users (especially beginners) in terms of api parity -- if `deps` can take list of list, what about ins, outs, and other plural names?

All being said, would love to hear the supportive arguments.

schrockn:
Agree that this adds a lot of complexity. I wonder how far we can get with better error messages and better naming conventions instead.

This problem was inevitable the moment we decided to make `AssetsDefinition` the core abstraction, rather than something that represents a singular object. Anything that refers to `assets` will be ambiguous since the name makes logical sense for either `AssetsDefinition` or `Sequence[AssetsDefinition]` .  We could adopt a naming convention where things that return a `Sequence[AssetsDefinition]` we name `load_assets_defs`. However since they are all `Union[AssetsDefinition, SourceAsset, CacheableAssetsDefinition]` that is _also_ confusing, in addition to being verbose. Tl;dr We are in a tough spot where no naming convention will be clear because the core abstraction is pluralized, and it will just be a set of unsatisfactory tradeoffs.

Given that I bias towards not changing things since the new style will just be unsatisfactory in different ways. In particular the shenanigans around `Sequence[str]` are brutal here.

One issue is that we are inconsistent with it comes to functions that  `load_assets_from`. Some (e.g. airbyte, fivetran, dbt cloud) return a singular `CacheableAssetsDefinition`.  I believe it was a mistake to have those load as single definition object, and they should return a sequence.

If we have integrations return a singular `AssetsDefinition` then they will be out of sync with the our core autoloading libraries, `load_assets_from_modules`, since those *have* to return a Sequence.

My proposed course of action:

* For sure add a special error handling pass prior to the hard runtime type check that detects a `List[AssetsDefinition]` as one of the list entries and provides a hyper specific error message for that case.
* Optionally come up with some migration plan to make `load_assets_from_fivetran_instance`, `load_assets_from_dbt_cloud_job` and `load_assets_from_connections`, `load_assets_from_airbyte_project`, and `load_assets_from_airbyte_instance` return a sequence.

schrockn:
cc: @sryza tl;dr for above comment is that having `load_assets_from_dbt_project` return just `AssetsDefinition` would be out of sync with `load_assets_from_modules` which have to return a `Sequence` and I think that is nonstarter.

sryza:
@schrockn by that logic, should `@dbt_assets` return a `Sequence[AssetsDefinition]`?

schrockn:
I think `dbt_multi_asset` is probably more consistent on a technical level, but that doesn't feel great. 

IMO the decorator is sufficiently different (it naturally feels more like a singular definition object as it is a single python symbol prefixed with `def`) that it is ok to be slightly inconsistent in that case.

sryza:
> IMO the decorator is sufficiently different (it naturally feels more like a singular definition object as it is a single python symbol prefixed with def) that it is ok to be slightly inconsistent in that case.

One impact of this is that it will be easily to accidentally use `get_asset_key_for_model` incorrectly:

```python
my_dbt_assets = load_assets_from_dbt_project(...)

get_asset_key_for_model(my_dbt_assets, "my_model")
```

and

```python
@dbt_assets
def my_dbt_assets():
    ...

get_asset_key_for_model([my_dbt_assets], "my_model")
```

jamiedemaria:
closing since we seem aligned on not supporting this

PR #15330, opened 2023-07-17
Enable gRPC readiness probes by default

johannkm:
Currently users don't have readiness probes on their grpc servers unless they explicitly opt in. Change this to opt out. Would be good timing if we get it in for 1.4, though I don't envision it causing issues. [We just bumped the timeout](https://github.com/dagster-io/dagster/pull/9073) from 3 seconds to 10 so it's very forgiving

### DIFF
--- a/helm/dagster/charts/dagster-user-deployments/templates/deployment-user.yaml
+++ b/helm/dagster/charts/dagster-user-deployments/templates/deployment-user.yaml
@@ -87,7 +87,16 @@ spec:
                 optional: true
             {{- end }}
           resources: {{ $deployment.resources | toYaml | nindent 12 }}
-        {{- if $deployment.readinessProbe }}
+        # Only disable readiness if explicitly set to false
+        {{- if not $deployment.readinessProbe }}
+          readinessProbe:
+            exec:
+              command: ["dagster", "api", "grpc-health-check", "-p", "{{ $deployment.port }}"]
+            periodSeconds: 20
+            timeoutSeconds: 3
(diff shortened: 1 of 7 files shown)


helm/dagster/charts/dagster-user-deployments/templates/deployment-user.yaml
```diff
@@ -87,7 +87,16 @@ spec:
...
-        {{- if $deployment.readinessProbe }}
+        # Only disable readiness if explicitly set to false
+        {{- if not $deployment.readinessProbe }}
+          readinessProbe:
+            exec:
+              command: ["dagster", "api", "grpc-health-check", "-p", "{{ $deployment.port }}"]
+            periodSeconds: 20
+            timeoutSeconds: 10
+            successThreshold: 1
+            failureThreshold: 15
```
alangenfeld, line 98: i'm not sure how representative it is, but there a number of users who have generated definitions who have to extend the server startup timeout setting in cloud deployments

johannkm: I'd still expect that to be ok unless the generation time is more than 5 minutes- these settings are really lenient, but there doesn't seem to be too big a drawback to that.

15 allowed failures * one per 20 seconds = 300 seconds. Not counting time waiting on each invocation of `grpc-health-check`, which is allowed to take up to 10 seconds

alangenfeld (approved):
alright trust your call

PR #15331, opened 2023-07-17
Retry .watch() tests

johannkm:
For flakes like https://buildkite.com/dagster/dagster/builds/61889#018964eb-68ce-48ff-a180-c632da10a2d6

At least for sqlite, .watch() is async (it uses a filewatcher). Every once in a while, we'll get to an assert before the event handler has run. We could:
- put time.sleeps in the tests
- retry and hope things pass, since they almost always do
- join the watcher somehow? I don't see a way to do this

Sprinkling sleeps around felt worse so I went with the latter, but it's not strongly held opinion

### DIFF
--- a/python_modules/dagster/dagster_tests/storage_tests/utils/event_log_storage.py
+++ b/python_modules/dagster/dagster_tests/storage_tests/utils/event_log_storage.py
@@ -473,6 +473,8 @@ class TestEventLogStorage:
             for run in runs:
                 instance.delete_run(run)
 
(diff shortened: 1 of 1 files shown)


johannkm:
* master
  * **PR #15331** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15331" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15331?utm_source=stack-comment).

jmsanders (approved):

PR #15343, opened 2023-07-18
have custom dbt metadata override default metadata instead of appending

sryza:
## Summary & Motivation

This PR includes a minor refactor of how customization functions are passed through the dbt internals. We now construct a `DagsterDbtTranslator` instead of passing all the functions around as their own arguments.


OwenKephart (approved):
very nice

PR #15346, opened 2023-07-18
don't fall back to model subdirectory by default when assigning groups to dbt assets

sryza:
## Summary & Motivation

dbt now has a way to specify [groups](https://docs.getdbt.com/docs/build/groups), so we should get out of the way and defer to that.

The `default_group_fn_fallback_to_directory` function is available if people want to revert to the previous behavior.


python_modules/libraries/dagster-dbt/dagster_dbt/asset_utils.py
```diff
@@ -380,6 +378,30 @@ def default_group_fn(node_info: Mapping[str, Any]) -> Optional[str]:
     if dbt_group:
         return dbt_group
 
+    return None
+
+
+def default_group_fn_fallback_to_directory(node_info: Mapping[str, Any]) -> Optional[str]:
```
rexledesma, line 384: need API string for args
```suggestion
def default_group_fn_fallback_to_directory(dbt_resource_props: Mapping[str, Any]) -> Optional[str]:
```

rexledesma (changes requested):
needs a test 🙏

rexledesma:
Feel free to take from https://github.com/dagster-io/dagster/pull/14864 -- will close that PR

docs/sphinx/sections/api/apidocs/libraries/dagster-dbt.rst
```diff
@@ -121,3 +121,7 @@ Utils
 .. currentmodule:: dagster_dbt.utils
 
 .. autofunction:: generate_materializations
+
+.. autofunction:: default_group_fn
```
rexledesma, line 125: Do we have a better name for this? Don't like this `_fn` prefix. Maybe `_from_dbt_resource_props`?

- `default_group_from_dbt_resource_props`
- `folder_group_from_dbt_resource_props`
- `directory_group_from_dbt_resource_props`
- `subdirectory_group_from_dbt_resource_props`

rexledesma (changes requested):

rexledesma (approved):

PR #15347, opened 2023-07-18
api docs for DagsterDbtTranslator and dbt_assets

sryza:

### DIFF
--- a/docs/sphinx/sections/api/apidocs/libraries/dagster-dbt.rst
+++ b/docs/sphinx/sections/api/apidocs/libraries/dagster-dbt.rst
@@ -26,6 +26,10 @@ Assets (dbt Core)
 
 .. autofunction:: load_assets_from_dbt_manifest
 
+.. autodecorator:: dbt_assets
+
+.. autoclass:: DagsterDbtTranslator
+
 Ops (dbt Core)
 ==============
 
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/dagster_dbt_translator.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/dagster_dbt_translator.py
@@ -10,6 +10,13 @@ from .asset_utils import default_asset_key_fn, default_description_fn, default_m
 
 
 class DagsterDbtTranslator:
+    """Holds a set of methods that derive Dagster asset definition metadata given a representation
+    of a dbt resource (models, tests, sources, etc).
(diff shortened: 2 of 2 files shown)


docs/sphinx/sections/api/apidocs/libraries/dagster-dbt.rst
```diff
@@ -121,3 +121,5 @@ Utils
 .. currentmodule:: dagster_dbt.utils
 
 .. autofunction:: generate_materializations
+
+.. autoclass:: DagsterDbtTranslator
```
rexledesma, line 125: This should be in the dbt Core section of the API docs

python_modules/libraries/dagster-dbt/dagster_dbt/dagster_dbt_translator.py
```diff
@@ -10,6 +10,12 @@
 
 
 class DagsterDbtTranslator:
+    """Holds a set of methods that derive Dagster metadata from dbt metadata and configuration.
```
rexledesma, line 13:
```suggestion
    """Holds a set of methods that derive Dagster asset definition metadata given a representation of a dbt resource (models, tests, sources, etc).
```

rexledesma (commented):
We should move `DagsterDbtTranslator` higher up so that there's no ambiguity with which part of `dagster-dbt` it should be used with.

sryza:
I incorporated comments and also expanded the PR to add `@dbt_assets` to the api docs.

python_modules/libraries/dagster-dbt/dagster_dbt/dagster_dbt_translator.py
```diff
@@ -10,6 +10,13 @@
 
 
 class DagsterDbtTranslator:
+    """Holds a set of methods that derive Dagster asset definition metadata given a representation
+    of a dbt resource (models, tests, sources, etc).
+
+    This class is exposed so that methods can be overriden to customize how Dagster metadata is
+    derived from dbt metadata and configuration.
```
rexledesma, line 17: Typo + more concise
```suggestion
    This class is exposed so that methods can be overridden to customize how Dagster asset definition metadata is derived.
```

rexledesma (approved):
Approving, but there is a typo

erinkcochran87 (approved):

PR #15351, opened 2023-07-18
deprecate compute log manager

prha:
## Summary & Motivation
For cleanliness... we could also choose to bump it further into the future (`2.0`?)

## How I Tested These Changes
BK


smackesey:
There is an existing deprecation warning on instance initialization that should be updated with the same breaking version and message: https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_core/instance/__init__.py#L377-L387

> we could also choose to bump it further into the future (2.0?)

If this was introduced after 1.0, then yeah 2.0 should be the breaking version since `CapturedLogManager` was not experimental.

PR #15335, opened 2023-07-17
Update dagit -> dagster-webserver for helm server entry point

smackesey:
## Summary & Motivation

This PR changes the `dagster-helm` chart so that it launches `dagster-webserver` instead of `dagit`, thus not printing a deprecation warning.

The bulk of webserver renames are upstack at #15356 , the entry point change is separated into this PR so that it can be merged separately if necessary.

## How I Tested These Changes

Existing test suite.

PR #15356, opened 2023-07-18
dagster-helm rename dagit -> dagster-webserver

smackesey:
## Summary & Motivation

Update the Dagster Helm chart with the new "webserver" terminology instead of "dagit."

- Rename all fields using "dagit" in `values.yaml`:
    - `.dagit` -> `.dagsterWebserver`
    - `.ingress.dagit` -> `.ingress.dagsterWebserver`
    - `.ingress.readOnlyDagit` -> `.ingress.readOnlyDagsterWebserver`
- Default pod name now starts with"dagster-webserver" instead of "dagster-dagit"
- To manage backcompatibility in templates, a new named template "dagster.backcompat" was added. This defines a dictionary that mounts merges of the old and new names at all renamed paths. It is intended to be "imported" into the variable `$_` and used like so:

```
{{ $_ := include "dagster.backcompat" . | mustFromJson }}
dagster-webserver -h 0.0.0.0 -p {{ $_.Values.dagsterWebserver.service.port }}
```

- In the schema, remove the `dagit` fields entirely. This does not prevent any values passed in under the old names from being accessed during template rendering. I am not sure if this is the best approach, but it keeps our schema simple. The alternative would be to clone the subschema used for `dagsterWebserver` but mark all the fields optional-- this would be necessary because there are no longer defaults defined for the old fields in `values.yaml`, so validation fails if partial values are provided under the old names (which in practice does not matter because they are merged into the values under the new names).
- Update relevant docs with the new terminology.
- Update integration test suites with the new terminology.

## How I Tested These Changes

I modified some of the schema tests we have to parametrize on old vs new field names. This verifies that using either name works. I confirmed that changing the parametrization to use a name that is neither new name or old name fails.


smackesey:
* master
  * **PR #15298** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15298" target="_blank">Graphite</a> 
    * **PR #15335** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15335" target="_blank">Graphite</a> 
      * **PR #15356** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15356" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15356?utm_source=stack-comment).

docs/content/deployment/guides/kubernetes/how-to-migrate-your-instance.mdx
```diff
@@ -18,23 +18,23 @@ For simplicity, we assume that you have only installed one release of the Dagste
 1. Before running the migration, check the [migration document](https://github.com/dagster-io/dagster/blob/master/MIGRATION.md) for any additional steps you may need to take as part of the migration.
 2. Make sure you backup the data in your PostgreSQL database!
 3. Run `helm upgrade` with your desired Dagster chart version and Helm values.
-4. Scale down the Dagit and Daemon Deployments. We do not want any ongoing job runs to be writing to the database as the migration is happening. As we scale down the Deployments, we note each Deployment's replica count to scale each back up after the migration is complete.
+4. Scale down the webserver and daemon deployments. We do not want any ongoing job runs to be writing to the database as the migration is happening. As we scale down the Deployments, we note each Deployment's replica count to scale each back up after the migration is complete.
```
erinkcochran87, line 21:
```suggestion
4. Scale down the Dagster webserver and daemon deployments. We do not want any ongoing job runs to be writing to the database as the migration is happening. As we scale down the Deployments, we note each Deployment's replica count to scale each back up after the migration is complete.
```

This looks like the first reference to this on this page, so adding `Dagster` to be consistent with other content

helm/dagster/README.md
```diff
@@ -24,10 +24,10 @@ helm install my-release . \
 
 ### Installing a specific Chart version
```
erinkcochran87, line 25: This isn't you, but since you changed `Chart` > `chart` in the copy that follows, could you also change it in the heading so it's consistent?

erinkcochran87 (commented):
Two small things for docs, so that lgtm. I'll leave it to the others to stamp for code.

docs/content/deployment/guides/kubernetes/how-to-migrate-your-instance.mdx
smackesey: good catch

helm/dagster/values.yaml
```diff
@@ -929,11 +929,11 @@ ingress:
   # See https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/#specifying-the-class-of-an-ingress
   ingressClassName: ~
 
-  dagit:
-    # Ingress hostname for Dagit e.g. dagit.mycompany.com
+  dagsterWebserver:
```
alangenfeld, line 932: should this just be `webserver` we just call the daemon `daemon` right?

alangenfeld (commented):
which tests ensure the `dagit` backcompat works?

helm/dagster/values.yaml
alangenfeld: ah my bad i see its `dagsterDaemon`

smackesey:
>which tests ensure the dagit backcompat works?

- `test_webserver_port` in `schema_tests/test_dagit.py`
- `test_ingress` and `test_ingress_read_only` in `schema_tests/test_ingress.py`

helm/dagster/templates/helpers/_helpers.tpl
```diff
@@ -1,4 +1,9 @@
 {{/* vim: set filetype=mustache: */}}
+
+{{/*
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
```
alangenfeld, line 5: ?

helm/dagster/values.yaml
```diff
@@ -964,11 +964,11 @@ ingress:
...
     succeedingPaths: []
 
-  readOnlyDagit:
-    # Ingress hostname for read only Dagit e.g. viewer.dagit.mycompany.com
+  readOnlyDagsterWebserver:
+    # Ingress hostname for read only webserver e.g. viewer.dagster.mycompany.com
     # This variable allows customizing the route pattern in the ingress. Some
     # ingress controllers may only support "/", whereas some may need "/*".
-    # NOTE: Dagit doesn't yet support hosting on a path, e.g. mycompany.com/dagit.
+    # NOTE: The Dagster webserver doesn't yet support hosting on a path, e.g. mycompany.com/dagster.
```
alangenfeld, line 971: [1] I think this note can be removed - we support `pathPrefix` for this

helm/dagster/values.yaml
```diff
@@ -929,11 +929,11 @@ ingress:
...
   ingressClassName: ~
 
-  dagit:
-    # Ingress hostname for Dagit e.g. dagit.mycompany.com
+  dagsterWebserver:
+    # Ingress hostname for the webserver e.g. dagster.mycompany.com
     # This variable allows customizing the route pattern in the ingress. Some
     # ingress controllers may only support "/", whereas some may need "/*".
-    # NOTE: Dagit doesn't yet support hosting on a path, e.g. mycompany.com/dagit.
+    # NOTE: The Dagster webserver doesn't yet support hosting on a path, e.g. mycompany.com/dagster.
```
alangenfeld, line 936: [1]

alangenfeld (approved):
I hope those tests are sufficient - no way did i catch small mistakes / typos in the yaml files

helm/dagster/schema/schema_tests/test_dagit.py
```diff
@@ -57,230 +60,250 @@ def workspace_configmap_helm_template() -> HelmTemplate:
...
 
     assert deployment_name == f"{deployment_template.release_name}-{name_override}"
 
 
-@pytest.mark.parametrize("path_prefix", ["dagit", "some-path"])
-def test_dagit_path_prefix(deployment_template, path_prefix):
-    helm_values = DagsterHelmValues.construct(dagit=Dagit.construct(pathPrefix=path_prefix))
-    dagit_deployments = deployment_template.render(helm_values)
+@pytest.mark.parametrize("path_prefix", ["webserver", "some-path"])
+def test_webserver_path_prefix(deployment_template, path_prefix):
```
alangenfeld, line 209: [2]

helm/dagster/templates/helpers/_helpers.tpl
smackesey: good catch just some leftover detritus

helm/dagster/schema/schema_tests/test_dagit.py
```diff
@@ -49,6 +49,9 @@ def workspace_configmap_helm_template() -> HelmTemplate:
     )
 
 
+# Parametrizing "webserver_field" here tests backcompat. `dagit` was the old field name.
+# This parametrization can be removed in 2.0.
+@pytest.mark.parametrize("webserver_field", ["dagsterWebserver", "dagit"])
```
alangenfeld, line 54: just this one? maybe slap this parametrize on a few more like [2]

smackesey: done

PR #15357, opened 2023-07-18
add get_group_name to DagsterDbtTranslator

sryza:

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/dagster_dbt_translator.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/dagster_dbt_translator.py
@@ -6,7 +6,12 @@ from dagster._core.definitions.events import (
     check_opt_coercible_to_asset_key_prefix_param,
 )
 
-from .asset_utils import default_asset_key_fn, default_description_fn, default_metadata_fn
+from .asset_utils import (
+    default_asset_key_fn,
+    default_description_fn,
+    default_group_fn,
+    default_metadata_fn,
+)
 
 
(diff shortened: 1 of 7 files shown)


python_modules/libraries/dagster-dbt/dagster_dbt/dagster_dbt_translator.py
```diff
@@ -110,6 +115,39 @@ def get_metadata(cls, dbt_resource_props: Mapping[str, Any]) -> Mapping[str, Any
         """
         return default_metadata_fn(dbt_resource_props)
 
+    @classmethod
+    def get_group_name(cls, dbt_resource_props: Mapping[str, Any]) -> Optional[str]:
```
rexledesma, line 119:
```suggestion
    def get_group(cls, dbt_resource_props: Mapping[str, Any]) -> Optional[str]:
```

rexledesma (commented):
We should add `get_freshness_policy` and `get_auto_materialize_policy` while we are at it

python_modules/libraries/dagster-dbt/dagster_dbt/dagster_dbt_translator.py
sryza: The property on `@asset` is `group_name`, so I thought it would be best to be consistent.

sryza:
> We should add get_freshness_policy and get_auto_materialize_policy while we are at it

I'd like to hold off on these for the sake of my time and sprinting towards the release

rexledesma (approved):

PR #15366, opened 2023-07-18
update deprecation warning messages post-translator renames

sryza:


OwenKephart (approved):
good catch -- one day, vscode will be smart enough to do this for me

PR #15368, opened 2023-07-18
[auto-materialize] Fix mapping from unpartitioned to time partitioned assets

OwenKephart:
## Summary & Motivation

This is a temporary fix for a perf regression caused by removing the time window partition scope minutes. In the specific case where an unpartitioned asset is upstream of a time partitioned assets, the general interpretation that we take is that this update is only relevant to the latest partition of the time partitioned asset. However, we don't currently represent this in the 

The net result is that the update to the unpartitioned upstream would result in us evaluating every single downstream, then creating a bunch of discard reasons for all of the partitions that we skipped (due to the rate limit). Beyond being slow, this would be confusing to a user who would not understand why those partitions would need to be discarded in the first place.

### DIFF
--- a/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
+++ b/python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
@@ -661,6 +661,17 @@ def determine_asset_partitions_to_auto_materialize(
             or candidate.asset_key not in target_asset_keys
             # must not be currently backfilled
             or candidate in instance_queryer.get_active_backfill_target_asset_graph_subset()
+            or (
+                # is not a root asset
+                candidate.asset_key not in asset_graph.root_asset_keys
+                # has a rate limit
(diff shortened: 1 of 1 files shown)


python_modules/dagster/dagster/_core/definitions/asset_reconciliation_sensor.py
```diff
@@ -445,8 +446,18 @@ def find_parent_materialized_asset_partitions(
                 current_time=instance_queryer.evaluation_time,
                 asset_key=asset_key,
             ):
+                child_partitions_def = asset_graph.get_partitions_def(child.asset_key)
                 if (
                     child.asset_key in target_asset_keys
+                    # when mapping from unpartitioned assets to time partitioned assets, we ignore
```
johannkm, line 452: is the plan to eventually encode this in a partition mapping object?

johannkm (approved):

PR #15369, opened 2023-07-18
feat(dbt): recommend runtime project reloading in scaffold command

rexledesma:
## Summary & Motivation
Recommend `DAGSTER_DBT_PARSE_PROJECT_ON_LOAD=1 dagster dev` after running the scaffold.

## How I Tested These Changes
local

### DIFF
--- a/python_modules/libraries/dagster-dbt/dagster_dbt/cli/app.py
+++ b/python_modules/libraries/dagster-dbt/dagster_dbt/cli/app.py
@@ -161,10 +161,8 @@ def project_scaffold_command(
         Syntax(
             code="\n".join(
                 [
-                    f"cd '{dbt_project_dir}' \\",
-                    "  && dbt parse --target-path target \\",
-                    f"  && cd '{dagster_project_dir}' \\",
-                    "  && dagster dev",
+                    f"cd '{dagster_project_dir}' \\",
+                    "  && (DAGSTER_DBT_PARSE_PROJECT_ON_LOAD=1 dagster dev)",
                 ]
             ),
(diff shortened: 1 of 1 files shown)


python_modules/libraries/dagster-dbt/dagster_dbt/cli/app.py
```diff
@@ -161,10 +161,8 @@ def project_scaffold_command(
         Syntax(
             code="\n".join(
                 [
-                    f"cd '{dbt_project_dir}' \\",
-                    "  && dbt parse --target-path target \\",
-                    f"  && cd '{dagster_project_dir}' \\",
-                    "  && dagster dev",
+                    f"cd '{dagster_project_dir}' \\",
```
rexledesma, line 164: Could potentially remove this `cd` if we switch into the newly created directory in this CLI. Would be pretty clean to just display one command.

sryza: I think it's best how it is. I think cd-ing within a command that does other things is unexpected and makes it harder to script against.

sryza (approved):

sryza (commented):
Something I realized while working on the tutorial: if they set `DAGSTER_DBT_PARSE_PROJECT_ON_LOAD=0` on the command line, then `os.getenv("DAGSTER_DBT_PARSE_PROJECT_ON_LOAD")` will still evaluate to true

Also, `DAGSTER_DBT_PARSE_PROJECT_ON_LOAD=true` might be clearer to users who aren't familiar with binary.

PR #15372, opened 2023-07-18
add .user.yml to .gitignore

sryza:


rexledesma (approved):
Can we remove all instances of `.user.yml` in our codebase as well?

PR #15377, opened 2023-07-18
Filter out default executor config when splatting defaults

benpankow:
## Summary

Special case the default multi-process executor config when splatting out run config yaml, since it can be confusing for users:

```yaml
execution:
  config:
    multiprocess:
      max_concurrent: 0
```

Here, `max_concurrent: 0` is a sentinel value that adjusts concurrency to match the number of cores on the executing machine. The value of including this entry in the default config is lower than ops/resources since most users won't need to touch it - it's more liable to cause confusion.

## Test Plan

Update unit tests.

### DIFF
--- a/python_modules/dagster/dagster/_core/snap/snap_to_yaml.py
+++ b/python_modules/dagster/dagster/_core/snap/snap_to_yaml.py
@@ -27,17 +27,23 @@ def _cleanup_run_config_dict(run_config_dict: Any) -> Any:
     """Performs cleanup of the run config dict to remove empty dicts and strip the default executor
     config if it has not been overridden, to make the output more readable.
     """
-    filtered_dict = _filter_empty_dicts(run_config_dict)
-
     # We manually strip the default executor config because
     # the max_concurrent default value is a sentinel and can be confusing
-    # if presented to the user
-    if filtered_dict.get("execution") == {
-        "config": {"multiprocess": {"max_concurrent": 0}},
-    }:
-        del filtered_dict["execution"]
+    # if presented to the user:
+    # execution:
+    #   config:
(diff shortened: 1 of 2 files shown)


benpankow:
* master
  * **PR #15377** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15377" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15377?utm_source=stack-comment).

yuhan (approved):

yuhan (commented):
stamped to unblock. but would be good to do a follow up pr to make this filter logic more future future proof - as in thinking how to prevent mistake if we were to change the default executor config. -- this and other similar prs almost make me think we should have a way to standardize our default config values and let other parts of the codebase easily reference the defaults.

python_modules/dagster/dagster/_core/snap/snap_to_yaml.py
```diff
@@ -23,12 +23,35 @@ def _filter_empty_dicts(to_filter: Any) -> Any:
...
+    # execution:
+    #   config:
+    #     multiprocess:
+    #       max_concurrent: 0
+    if (
+        run_config_dict.get("execution", {})
+        .get("config", {})
+        .get("multiprocess", {})
+        .get("max_concurrent", None)
+        == 0
```
alangenfeld, line 42: this is p gross - did you consider changing the schema so that it handles unset instead of default value 0 ?

benpankow: #15389

PR #15384, opened 2023-07-19
Add `config` argument to graph_asset

### LINKED ISSUE #12647: Allow providing config to @graph_asset decorator
### What's the use case?

`@graph_asset` (added in https://github.com/dagster-io/dagster/pull/10152) is missing the `config` attribute that allows one to pass/define config for Ops that are part of the graph/graph backed Asset.

The `@graph` annotation does support this
https://github.com/dagster-io/dagster/blob/fe8ed527ee9c4acae0568a6c3466346be544973f/python_modules/dagster/dagster/_core/definitions/decorators/graph_decorator.py#L135
as does the `GraphDefinition` class
https://github.com/dagster-io/dagster/blob/fe8ed527ee9c4acae0568a6c3466346be544973f/python_modules/dagster/dagster/_core/definitions/graph_definition.py#L205
so I'd expect that the `@graph_asset` should also be able to take a `config` argument and pass this on to the graph it creates for the Asset.

### Ideas of implementation

_No response_

### Additional information

_No response_

### LINKED ISSUE #14944: `graph_asset` is missing a `config_mapping` argument
### Dagster version

1.3.11

### What's the issue?

The title. 

While a workaround with using `AssetsDefinition.from_graph` exists, it adds some unnecessary boilerplate. 

[Slack thread](https://dagster.slack.com/archives/C01U954MEER/p1686765536456289) 

_No response_

### How to reproduce?

_No response_

### Deployment type

None

### Deployment details

_No response_

### Additional information

_No response_

smackesey:
## Summary & Motivation

Resolves #14944.
Resolves #12647.

- Add `config` arg to `@graph_asset`
- Add `config` docstring to `@graph`
- Light refactor of `@graph_asset` higher-order decorator style to eliminate duplication of long argument list.

`@graph_asset` is currently missing `config`. This has been reported by multiple users. IMO it should clearly be present since it's part of the standard functionality of `@graph`.

@sryza in #12647 you wrote:

>@asset doesn't currently have a config attribute, but we've been discussing adding one. Arguably, @graph_asset and @asset should be consistent on this.

`@graph_asset` combines arguments of both `@graph` and `@asset`, so I think it is important to add this regardless of whether it is added to `@asset` as well, as we need to be consistent with `@graph`.

## How I Tested These Changes

New unit test.

### DIFF
--- a/python_modules/dagster/dagster_tests/asset_defs_tests/test_decorators.py
+++ b/python_modules/dagster/dagster_tests/asset_defs_tests/test_decorators.py
@@ -31,6 +31,7 @@ from dagster import (
     resource,
 )
 from dagster._check import CheckError
+from dagster._config.pythonic_config import Config
 from dagster._core.definitions import (
     AssetIn,
     AssetsDefinition,
@@ -39,6 +40,7 @@ from dagster._core.definitions import (
     multi_asset,
 )
 from dagster._core.definitions.auto_materialize_policy import AutoMaterializePolicy
+from dagster._core.definitions.decorators.config_mapping_decorator import config_mapping
 from dagster._core.definitions.policy import RetryPolicy
 from dagster._core.definitions.resource_requirement import ensure_requirements_satisfied
 from dagster._core.errors import DagsterInvalidConfigError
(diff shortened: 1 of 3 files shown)


smackesey:
* master
  * **PR #15384** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15384" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15384?utm_source=stack-comment).

python_modules/dagster/dagster_tests/asset_defs_tests/test_decorators.py
```diff
@@ -988,6 +990,45 @@ def str_prefix():
...
+    class FooConfig(Config):
+        val: int
+
+    @op
+    def foo_op(config: FooConfig):
+        return config.val
+
+    @config_mapping(config_schema={"val": int})
+    def foo_config_mapping(val: Any) -> Any:
+        return {"foo_op": {"config": {"val": val}}}
```
alangenfeld, line 1021:
```suggestion
    @config_mapping(config_schema={"val": int})
    def foo_config_mapping(config: Any) -> Any:
        return {"foo_op": {"config": config}}
```

smackesey: yeah this was indeed the problem

alangenfeld (approved):

smackesey:
@sryza I know in the linked issues and Slack you were uncertain whether we should add this so pinging you.

PR #15385, opened 2023-07-19
Coerce asset_partitions to a set

jmsanders:
This comparison hasn't been as quick as the comment indicates. Profiling suggests that a lot of time is being spent in this particular set comparsion.

It looks like `asset_partitions` is being provided as an `Optional[Sequence[AssetPartitionKey]]` but being treated as if it's a set. This naively coerces the sequence to a set for our comparison but I think a better implementation might involve updating our call sites to make sure we're passing a set in the first place.

### DIFF
--- a/python_modules/dagster/dagster/_utils/caching_instance_queryer.py
+++ b/python_modules/dagster/dagster/_utils/caching_instance_queryer.py
@@ -581,9 +581,12 @@ class CachingInstanceQueryer(DynamicPartitionsStore):
             for asset_partition, latest_storage_id in self._get_latest_materialization_or_observation_storage_ids_by_asset_partition(
                 asset_key=asset_key
             ).items()
-            if (asset_partitions is None or asset_partition in set(asset_partitions))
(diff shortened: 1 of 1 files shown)


shalabhc (approved):

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -581,9 +581,12 @@ def get_asset_partitions_updated_after_cursor(
             for asset_partition, latest_storage_id in self._get_latest_materialization_or_observation_storage_ids_by_asset_partition(
                 asset_key=asset_key
             ).items()
-            if (asset_partitions is None or asset_partition in asset_partitions)
-            and (latest_storage_id or 0) > (after_cursor or 0)
+            if (latest_storage_id or 0) > (after_cursor or 0)
         }
+
+        if asset_partitions:
```
OwenKephart, line 587: we should never be passing in an empty list here but I always prefer to be specific with these comparisons (if asset_partitions is not None)

PR #15396, opened 2023-07-19
[docs] various asset concept docs changes

jamiedemaria:
[low priority]

## Summary & Motivation

I was reading through the assets concepts pages to determine what examples need to be changed to use the `deps` API, and I ran across a couple random places that I think could use a little improvement (not related to `deps`). Opening them in a PR here to propose some changes. I'm leaving my reasoning for each change in comments, and please feel free to accept/reject individual changes. I erred on the side of making bigger/more dramatic changes with the thinking that it's easier to say no or scale back changes

### DIFF
--- a/docs/content/concepts/assets/asset-observations.mdx
+++ b/docs/content/concepts/assets/asset-observations.mdx
@@ -5,7 +5,7 @@ description: Dagster provides functionality to record metadata about assets.
 
 # Asset Observations
 
-An asset observation is an event that records metadata about a given asset. Unlike asset materializations, asset observations do not signify that an asset has been mutated.
+You may wish to record metadata about an asset outside of that asset's materialization function. To do this, you can write another function and log <PyObject object="AssetObservation" /> events. An asset observation is an event that records metadata about a given asset. Unlike asset materializations, asset observations do not signify that an asset has been mutated.
 
 ## Relevant APIs
 
@@ -33,7 +33,7 @@ from dagster import AssetObservation, op
 
 
 @op
-def observation_op(context):
+def observation_op(context: OpExecutionContext):
     df = read_df()
     context.log_event(
         AssetObservation(asset_key="observation_asset", metadata={"num_rows": len(df)})
@@ -59,7 +59,7 @@ from dagster import AssetMaterialization, AssetObservation, MetadataValue, op
 
 
 @op
-def observes_dataset_op(context):
+def observes_dataset_op(context: OpExecutionContext):
     df = read_df()
     remote_storage_path = persist_to_storage(df)
     context.log_event(
@@ -101,7 +101,7 @@ class MyOpConfig(Config):
 
 
 @op
-def partitioned_dataset_op(context, config: MyOpConfig):
+def partitioned_dataset_op(context: OpExecutionContext, config: MyOpConfig):
     partition_date = config.date
     df = read_df_for_date(partition_date)
     context.log_event(
--- a/docs/content/concepts/assets/graph-backed-assets.mdx
+++ b/docs/content/concepts/assets/graph-backed-assets.mdx
@@ -5,6 +5,16 @@ description: Defining a software-defined asset with multiple discrete computatio
 
 # Graph-Backed Assets
 
+<Note>
+  This page assumes you have some familiarity with <code>ops</code> and{" "}
+  <code>graphs</code>. To learn more about <code>ops</code> and{" "}
+  <code>graphs</code>, see the documentation for <code>ops</code>
+  <a href="https://docs.dagster.io/concepts/ops-jobs-graphs/ops">here</a> and <code>
+    graphs
+  </code>
+  <a href="https://docs.dagster.io/concepts/ops-jobs-graphs/graphs">here</a> .
+</Note>
+
 [Basic software-defined assets](/concepts/assets/software-defined-assets#a-basic-software-defined-asset) are computed using a single op. If generating an asset involves multiple discrete computations, you can use graph-backed assets by separating each computation into an op and building a graph to combine your computations. This allows you to launch re-executions of runs at the op boundaries, but doesn't require you to link each intermediate value to an asset in persistent storage.
 
 ---
@@ -28,11 +38,12 @@ In this case, when you tell Dagster to materialize the `slack_files_table` asset
 ```python file=/concepts/assets/graph_backed_asset.py startafter=start example endbefore=end example
 import pandas as pd
 from dagster import OpExecutionContext, graph_asset, op
+from dagster_slack import SlackResource
 
 
-@op(required_resource_keys={"slack"})
-def fetch_files_from_slack(context: OpExecutionContext) -> pd.DataFrame:
-    files = context.resources.slack.files_list(channel="#random")
+@op
+def fetch_files_from_slack(slack: SlackResource) -> pd.DataFrame:
+    files = slack.get_client().files_list(channel="#random")
     return pd.DataFrame(
         [
             {
@@ -93,12 +104,22 @@ def downstream_asset(middle_asset):
 
 ### Graph-backed multi-assets
 
-If you want to write a graph-backed asset that results in the materialization of multiple assets, you can use the <PyObject object="graph_multi_asset" decorator />.
+If you want to write a graph-backed asset that results in the materialization of multiple assets, you can use the <PyObject object="graph_multi_asset" decorator /> decorator.
 
-In the below example, `two_assets` depends on `upstream_asset` (from the above example). `two_assets` is a graph-backed multi-asset that is composed of two ops, `multiply_by_two` (defined in the above example) and `divide_by_three`:
+In the below example, `two_assets` depends on the asset `ten`. `two_assets` is a graph-backed multi-asset that is composed of two ops, `multiply_by_two` and `divide_by_three`:
 
 ```python file=/concepts/assets/graph_backed_asset.py startafter=start_basic_dependencies_2 endbefore=end_basic_dependencies_2
 from dagster import AssetOut, graph_multi_asset
+
+
+@asset
+def ten():
+    return 10
+
+
+@op
+def multiply_by_two(input_num):
+    return input_num * 2
 
 
 @op
@@ -110,8 +131,8 @@ def divide_by_three(number: int) -> Tuple[int, int]:
 
 
 @graph_multi_asset(outs={"quotient": AssetOut(), "remainder": AssetOut()})
-def two_assets(upstream_asset):
-    q, r = divide_by_three(multiply_by_two(upstream_asset))
+def two_assets(ten):
+    q, r = divide_by_three(multiply_by_two(ten))
     return {"quotient": q, "remainder": r}
 ```
 
--- a/docs/content/concepts/assets/graph-backed-assets.mdx
+++ b/docs/content/concepts/assets/graph-backed-assets.mdx
@@ -60,7 +60,7 @@ def slack_files_table():
 
 Similar to software-defined assets, Dagster infers the upstream assets from the names of the arguments to the decorated function. Dagster will then delegate loading the data to an [I/O manager](/concepts/io-management/io-managers).
 
-The example below includes an asset named `middle_asset`. `middle_asset` depends on `upstream_asset`, and `downstream_asset` depends on `middle_asset`:
+The example below includes a graph-backed asset named `middle_asset`. `middle_asset` depends on `upstream_asset`, and `downstream_asset` depends on `middle_asset`:
 
 ```python file=/concepts/assets/graph_backed_asset.py startafter=start_basic_dependencies endbefore=end_basic_dependencies
 from dagster import asset, graph_asset, op
@@ -93,18 +93,26 @@ def downstream_asset(middle_asset):
 
 ### Graph-backed multi-assets
 
-Using the <PyObject object="graph_multi_asset" decorator />, you can create a combined definition of multiple assets that are computed using the same graph of ops and same upstream assets.
+If you want to write a graph-backed asset that results in the materialization of multiple assets, you can use the <PyObject object="graph_multi_asset" decorator />.
 
-In the below example, `two_assets` accepts `upstream_asset` and outputs two assets, `first_asset` and `second_asset`:
+In the below example, `two_assets` depends on `upstream_asset` (from the above example). `two_assets` is a graph-backed multi-asset that is composed of two ops, `multiply_by_two` (defined in the above example) and `divide_by_three`:
 
 ```python file=/concepts/assets/graph_backed_asset.py startafter=start_basic_dependencies_2 endbefore=end_basic_dependencies_2
 from dagster import AssetOut, graph_multi_asset
 
 
-@graph_multi_asset(outs={"first_asset": AssetOut(), "second_asset": AssetOut()})
+@op
+def divide_by_three(number: int) -> Tuple[int, int]:
+    quotient = number // 3
+    remainder = number % 3
+
+    return quotient, remainder
+
+
+@graph_multi_asset(outs={"quotient": AssetOut(), "remainder": AssetOut()})
 def two_assets(upstream_asset):
-    one, two = two_outputs(upstream_asset)
-    return {"first_asset": one, "second_asset": two}
+    q, r = divide_by_three(multiply_by_two(upstream_asset))
+    return {"quotient": q, "remainder": r}
 ```
 
 ### Advanced: Subsetting graph-backed assets
--- a/docs/content/concepts/assets/graph-backed-assets.mdx
+++ b/docs/content/concepts/assets/graph-backed-assets.mdx
@@ -27,11 +27,11 @@ In this case, when you tell Dagster to materialize the `slack_files_table` asset
 
 ```python file=/concepts/assets/graph_backed_asset.py startafter=start example endbefore=end example
 import pandas as pd
-from dagster import graph_asset, op
+from dagster import OpExecutionContext, graph_asset, op
 
 
 @op(required_resource_keys={"slack"})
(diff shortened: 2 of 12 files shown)


jamiedemaria:
* master
  * **PR #15396** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15396" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15396?utm_source=stack-comment).

docs/content/concepts/assets/graph-backed-assets.mdx
```diff
@@ -58,9 +58,9 @@ def slack_files_table():
 
 ### Defining basic dependencies for graph-backed assets
 
-Similar to with basic software-defined assets, Dagster infers the upstream assets from the names of the arguments to the decorated function.
+Similar to software-defined assets, Dagster infers the upstream assets from the names of the arguments to the decorated function.
```
jamiedemaria, line 61: small grammar fix

docs/content/concepts/assets/graph-backed-assets.mdx
```diff
@@ -58,9 +58,9 @@ def slack_files_table():
...
 
-The example below includes an asset named `middle_asset`. `middle_asset` depends on `upstream_asset`, and `downstream_asset` depends on `middle_asset`:
+The example below includes a graph-backed asset named `middle_asset`. `middle_asset` depends on `upstream_asset`, and `downstream_asset` depends on `middle_asset`:
```
jamiedemaria, line 63: i felt like being more specific that `middle_asset` was the graph backed asset made it easier to find the important part of this example

docs/content/concepts/assets/graph-backed-assets.mdx
```diff
@@ -93,18 +93,26 @@ def downstream_asset(middle_asset):
 
 ### Graph-backed multi-assets
 
-Using the <PyObject object="graph_multi_asset" decorator />, you can create a combined definition of multiple assets that are computed using the same graph of ops and same upstream assets.
+If you want to write a graph-backed asset that results in the materialization of multiple assets, you can use the <PyObject object="graph_multi_asset" decorator />.
```
jamiedemaria, line 96: I found this section pretty confusing/hard to parse. I boiled it down to two reasons:
* the intro sentence was quite complicated, lots of noun soup with `combined definition of multiple assets that are computed using the same graph of ops and same upstream assets` 
* the example didn't explain what `two_outputs` did, nor did it really need to be a graph-backed multi asset since it was a single op (could have just been a plain multi asset)

So i played with the intro sentence a bit. This read easier to me, but that could just be a personal reading style thing, so very open to feedback.

I also updated the example to do some computation (even though that computation is trivial). This is a personal nit with some of our docs examples, but i feel like we often force the user to imagine/infer what the functions do, which can take away from understanding how the pieces fit together

docs/content/concepts/assets/multi-assets.mdx
```diff
@@ -132,3 +134,23 @@ def my_complex_assets(a, b):
     # d only depends on b
     yield Output(value=b + 1, output_name="d")
 ```
+
+---
+
+## Multi-asset code versions
```
jamiedemaria, line 140: I moved the multi-asset specific code version docs to the multi-asset concept page because I felt that they didn't fit in the assets concept page. The asset concept page is probably one of the first ones a user looks at after the tutorial, and this particular section introduces two new concepts (multi assets and ops) that are not strictly necessary yet

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -25,9 +25,9 @@ A software-defined asset includes the following:
 
 - An <PyObject object="AssetKey" />, which is a handle for referring to the asset.
 - A set of upstream asset keys, which refer to assets that the contents of the software-defined asset are derived from.
-- An [op](/concepts/ops-jobs-graphs/ops), which is a function responsible for computing the contents of the asset from its upstream dependencies.
+- A Python function which is responsible for computing the contents of the asset from its upstream dependencies.
```
jamiedemaria, line 28: This is maybe a more controversial change, so if you don't think it makes sense that's totally cool. putting myself in the shoes of a brand new user, I envision my path through the docs to be 1) tutorial, 2) concept pages. Specifically, I would start with the assets concept page. and being presented with the concept of ops right away would be pretty intimidating (thoughts like "ok so i need to go read that concept page first?", "maybe i won't understand assets if i dont understand ops" etc)

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -544,22 +528,22 @@ def test_uses_resource() -> None:
     assert result == {"foo": "bar"}
 ```
 
-If you use a context object in your function, you can use <PyObject object="build_op_context" /> to generate the `context` object, because under the hood the function decorated by `@asset` is an op.
+If you use a context object in your function, you can use <PyObject object="build_asset_context" /> to generate the `context` object.
```
jamiedemaria, line 531: just updating this section to use the new `build_asset_context` api

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -629,15 +613,31 @@ def table1() -> Output[DataFrame]:
...
-from dagster import Output, asset
+from dagster import AssetExecutionContext, asset
 
 
 @asset
-def table1() -> Output[None]:
+def table1(context: AssetExecutionContext) -> None:
     ...  # write out some data to table1
-    return Output(None, metadata={"num_rows": 25})
+    context.add_output_metadata(metadata={"num_rows": 25})
```
jamiedemaria, line 622: not 100% sure this is what we want to recommend, but ti's the approach we are taking in the tutorial for when an asset had no return value, but you want to attach metadata. We should update this to whatever the outcome of this [discussion](https://github.com/dagster-io/dagster/discussions/15392) is

docs/content/concepts/assets/asset-observations.mdx
```diff
@@ -5,7 +5,7 @@ description: Dagster provides functionality to record metadata about assets.
 
 # Asset Observations
 
-An asset observation is an event that records metadata about a given asset. Unlike asset materializations, asset observations do not signify that an asset has been mutated.
+You may wish to record metadata about an asset outside of that asset's materialization function. To do this, you can write another function and log <PyObject object="AssetObservation" /> events. An asset observation is an event that records metadata about a given asset. Unlike asset materializations, asset observations do not signify that an asset has been mutated.
```
jamiedemaria, line 8: i felt like the intro was missing some motivation on why you would use an AssetObservation (rather than just log metadata in the asset materialization function). I still sort of feel that way, but this felt like a small improvement

docs/content/concepts/assets/graph-backed-assets.mdx
```diff
@@ -27,11 +27,11 @@ In this case, when you tell Dagster to materialize the `slack_files_table` asset
 
 ```python file=/concepts/assets/graph_backed_asset.py startafter=start example endbefore=end example
 import pandas as pd
-from dagster import graph_asset, op
+from dagster import OpExecutionContext, graph_asset, op
 
 
 @op(required_resource_keys={"slack"})
```
jamiedemaria, line 33: make this a Pythonic resource

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -86,12 +86,17 @@ from dagster import asset
 
 @asset
 def sugary_cereals() -> None:
-    execute_query("CREATE TABLE sugary_cereals AS SELECT * FROM cereals")
+    execute_query(
```
jamiedemaria, line 89: updated these queries to be a bit more interesting than copying the whole table

docs/content/concepts/assets/multi-assets.mdx
```diff
@@ -22,6 +22,8 @@ Multi-assets may be useful in the following scenarios:
 - A single call to an API results in multiple tables being updated (e.g. Airbyte, Fivetran, dbt).
 - The same in-memory object is used to compute multiple assets.
 
+---
```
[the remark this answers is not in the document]
erinkcochran87: I like this a lot. I think we could do a better job of telling folks there are recommended prerequisites for concepts that go beyond surface level

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -25,11 +25,11 @@ A software-defined asset includes the following:
...
 
-**Materializing** an asset is the act of running its op and saving the results to persistent storage. You can initiate materializations from [the Dagster UI](/concepts/webserver/ui) or by invoking Python APIs. By default, assets are materialized to pickle files on your local filesystem, but materialization behavior is fully customizable using [I/O managers](/concepts/io-management/io-managers#applying-io-managers-to-assets). It's possible to materialize an asset in multiple storage environments, such as production and staging.
+**Materializing** an asset is the act of running its function and saving the results to persistent storage. You can initiate materializations from [the Dagster UI](/concepts/webserver/ui) or by invoking Python APIs. By default, assets are materialized to pickle files on your local filesystem, but materialization behavior is fully customizable using [I/O managers](/concepts/io-management/io-managers#applying-io-managers-to-assets). It's possible to materialize an asset in multiple storage environments, such as production and staging.
```
jamiedemaria, line 32: i did a pass where i tried to remove the word `op` as much as possible. Will mark them as such (with a [1]), and as before, feel free to accept/reject individual changes or the entire `op` replacement

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -38,7 +38,7 @@ A software-defined asset includes the following:
...
 | ------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
 | <PyObject object="asset" decorator /> | A decorator used to define assets.                                                                                                                                                                                                   |
-| <PyObject object="SourceAsset" />     | A class that describes an asset, but doesn't define how to compute it. <PyObject object="SourceAsset" />s are used to represent assets that other assets or jobs depend on, in settings where they can't be materialized themselves. |
+| <PyObject object="SourceAsset" />     | A class that describes an asset, but doesn't define how to compute it. <PyObject object="SourceAsset" />s are used to represent assets that other assets depend on, in settings where they can't be materialized themselves. |
```
jamiedemaria, line 41: removed "or jobs" since it didn't seem necessary and references a concept the user may not have familiarity with

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -63,7 +63,7 @@ def my_asset():
     return [1, 2, 3]
 ```
 
-By default, the name of the decorated function, `my_asset`, is used as the asset key. The decorated function forms the asset's op: it's responsible for producing the asset's contents. The asset in this example doesn't depend on any other assets.
+By default, the name of the decorated function, `my_asset`, is used as the asset key. The decorated function is responsible for producing the asset's contents. The asset in this example doesn't depend on any other assets.
```
jamiedemaria, line 66: [1]

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -213,22 +218,22 @@ defs = Definitions(
 )
 ```
 
-Using source assets has a few advantages over having the code inside of an asset's op load the data:
+Using source assets has a few advantages over having the code inside of an asset's function load the data without specifying any dependencies:
```
jamiedemaria, line 221: [1]

reworked this sentence in general since "having the code inside an assets' op load the data" kinda undermines the `deps` workflow where the asset's op might do just that. I leaned more into the specifying dependencies vs not specifying, but i'm not sure i totally got the idea across quite right. TODO - return to this sentence and re-work

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -213,22 +218,22 @@ defs = Definitions(
...
 
 - **The UI can show asset lineage that includes the source assets**. If different asset definitions in different code locations have the same asset key as a <PyObject object="SourceAsset" /> and both code locations are loaded into the underlying webserver, the UI can represent the asset lineage across the code locations. This can be accomplished using [workspace files](/concepts/code-locations/workspace-files).
 - **Dagster can use data-loading code factored into an <PyObject object="IOManager" /> to load the contents of the source asset**.
-- **Asset dependencies can be written in a consistent way,** independent of whether they're downstream from a source asset or a derived asset. This makes it easy to swap out a source asset for a derived asset and vice versa.
+- **Asset dependencies can be written in a consistent way,** independent of whether they're downstream from a source asset or a software-defined asset. This makes it easy to swap out a source asset for a software-defined asset and vice versa.
```
jamiedemaria, line 225: the term "derived asset" isn't used anywhere else in this page and it's kinda confusing what we mean here. Swapped for "software-defined asset" but i'm not sure that's totally accurate either

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -213,22 +218,22 @@ defs = Definitions(
...
 
-### Graph-backed assets and multi-assets
+### Writing more computationally complex assets
```
jamiedemaria, line 227: did a rework of this section to avoid `ops` and `graphs` so that, ideally, a new user would be able to understand what `multi_asset` and `graph backed asset` do without needing to first understand other concepts

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -306,23 +311,7 @@ def asset_with_version():
...
-        "b": AssetOut(code_version="2"),
-    }
-)
-def multi_asset_with_versions():
-    yield Output(100, "a")
-    yield Output(200, "b")
-```
-
-Just as with regular assets, these versions are attached to the `AssetMaterialization` objects for each of the constituent assets and represented in the UI.
+When an asset with a code version is materialized, it is tagged with the version. The UI will indicate when an asset has a different code version than the code version used for its most recent materialization.
```
jamiedemaria, line 314: removed the `AssetMaterialization` here since this is the first mention of it and i could see that being confusing. Additionally, i don't think it's that important that the user knows we tag the `AssetMaterialization` event (unless their like a power user maybe) but open to push-back on that conclusion

docs/content/concepts/assets/software-defined-assets.mdx
```diff
@@ -615,7 +604,7 @@ def downstream_asset(upstream_asset):
 
 ### Recording materialization metadata
 
-Dagster supports attaching arbitrary [metadata](/\_apidocs/ops#dagster.MetadataValue) to asset materializations. This metadata will be displayed on the "Activity" tab of the "Asset Details" page in the UI. If it's numeric, it will be plotted. To attach metadata, your asset's op can return an <PyObject object="Output" /> object that contains the output value and a dictionary of metadata:
+Dagster supports attaching arbitrary [metadata](/\_apidocs/ops#dagster.MetadataValue) to asset materializations. This metadata will be displayed on the "Activity" tab of the "Asset Details" page in the UI. If it's numeric, it will be plotted. To attach metadata, your asset can return an <PyObject object="Output" /> object that contains the output value and a dictionary of metadata:
```
jamiedemaria, line 607: [1]

jamiedemaria (commented):
did another pass with some more updates

As a side note, I definitely feel like I might be over-indexing on the experience of a first time reader, potentially at the expense of an experienced dagster user who is looking for more in-depth information. let me know if you also think that is the case and i can backtrack a bit to find a happy medium. But i do think it's an interesting exercise to re-work this page for a first time user, it's really opened my eyes to the vast number of concepts we toss out at the get-go

docs/content/concepts/assets/multi-assets.mdx
```diff
@@ -109,7 +111,7 @@ def split_actions(context):
 
 Because our outputs are now optional, we can no longer rely on returning a tuple of values, as we don't know in advance which values will be computed. Instead, we explicitly `yield` each output that we're expected to create.
 
-### Dependencies inside multi-assets
+## Dependencies inside multi-assets
```
erinkcochran87, line 114:
```suggestion
---

## Dependencies inside multi-assets
```

docs/content/concepts/assets/multi-assets.mdx
```diff
@@ -132,3 +134,23 @@ def my_complex_assets(a, b):
     # d only depends on b
     yield Output(value=b + 1, output_name="d")
 ```
+
+---
+
+## Multi-asset code versions
+
+Like regular assets, multi-assets may assign different code versions for each of their outputs. Versions let you help Dagster track what assets haven't been re-materialized since their code has changed, and avoid performing redundant computation.
```
erinkcochran87, line 142: D'you think we should link to the versioning/caching guide here? https://docs.dagster.io/guides/dagster/asset-versioning-and-caching

I feel like I want *something* more about versioning, but not sure if this is the right page - or if we should maybe create something new that's not as complicated as that guide.

docs/content/concepts/assets/software-defined-assets.mdx
jamiedemaria: that's a good callout. i will think on what this could be re-titled

docs/content/concepts/assets/asset-observations.mdx
jamiedemaria: tbh me too. I think it's mostly helpful for reporting metadata about an asset after it's been materialized, which can be helpful if your asset calls out to an external tool that handles storing data for you (rather that doing IO managers). However, it looks like a lot of the content in the page is pretty old (like from right around when assets were first introduced), so it might be due for a refresh to account for all the changes since then

jamiedemaria:
I'm in the process of splitting this PR out into several PRs. closing this to help clear up review queues

PR #15404, opened 2023-07-20
[docs] Enforce use of Args: over Attrs: in class docstrings

smackesey:
## Summary & Motivation

Currently in the codebase, some classes use `Attributes:` and others use `Args:` to list class parameters in class docstrings. The two cases render quite differently (see screenshots). This PR:

- converts all existing uses of `Attributes` to `Args` (which renders more compactly and consistently with functions)
- adds a check to the API doc build that causes a failure if use of `Attributes:` is detected.

## How I Tested These Changes

Built docs.


smackesey:
* master
  * **PR #15404** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15404" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15404?utm_source=stack-comment).

sryza:
I agree that the way that "Args" gets rendered looks a lot nicer. However, I think that "Attrs" is more representative of what we're trying to communicate, i.e. that these are the attributes of this class.  I wonder if there's some way to address the rendering issues.

Also see: https://stackoverflow.com/questions/46569163/sphinx-document-an-attribute-that-is-same-as-a-parameter

smackesey:
>I agree that the way that "Args" gets rendered looks a lot nicer. However, I think that "Attrs" is more representative of what we're trying to communicate, i.e. that these are the attributes of this class. I wonder if there's some way to address the rendering issues.

Disagree that "Attrs" is more representative of what we're communicating, because the way most of our public classes are written we don't expose attributes-- we expose properties, which are documented separately. At the level of the class docstring itself, the public API we are documenting is the constructor parameters, not the attributes, which are typically private. Also, often the constructor parameters are going to differ from the attributes-- this happens whenever there is a deprecated param, but also in other cases where there is some logic in the constructor. It's clear that we need to document such deprecated constructor params but it would be incorrect to call them "attributes".

Now for `NamedTuple` classes, we actually _do_ expose attributes via `PublicAttr`. Most of the time these are identical to the constructor params, but not always. To be technically correct we would have separate documentation of the constructor params and the attributes, but that seems like overkill when they are the same 95% of the time. There is probably something we can do in sphinx to solve this, I have to think about it.

sryza:
In a google search on "python properties vs attributes", most results claim that properties are a special kind of attribute.

In many cases, the types that a constructor accepts will be more permissive than the class's attributes, or constructors will include deprecated parameters that get converted to canonical attributes.  It's useful to have documentation for the canonical attributes because they describe what the class _is_, rather than just what arguments are required to construct it.

sryza:
Might be worth collecting other opinions from the team? Not a hill that I'm going to die on.

erinkcochran87 (commented):
I'll leave it to others to stamp on code, but stylistically this is a great improvement. Thank you Sean!

smackesey:
>In a google search on "python properties vs attributes", most results claim that properties are a special kind of attribute.
>
>In many cases, the types that a constructor accepts will be more permissive than the class's attributes, or constructors will include deprecated parameters that get converted to canonical attributes. It's useful to have documentation for the canonical attributes because they describe what the class is, rather than just what arguments are required to construct it.

We might be talking past each other here. As I see it it is immaterial whether properties are a subclass of attributes or not, because properties _have their own docstrings_ and are not documented in the class-level docstring. The two candidates for documentation in the class-level docstring are:

- Non-property attributes ("Attributes" case)
- Constructor parameters ("Args" case)

For non-NamedTuple classes, it seems 100% clear to me that our class-level docstrings should be covering the constructor parameters. That's because: (a) we don't use public non-property attributes; (b) constructor parameters of public classes must be documented somewhere; (c) the content of these docstrings right now actually reflects constructor parameters rather than non-property attributes.

The tricky case is `NamedTuple` descendants, because for these classes we do in fact expose public attributes without using properties. The attributes tend to closely mirror the constructor parameters but not perfectly-- for example, [`AssetMaterialization`](https://github.com/dagster-io/dagster/blob/b2f50e1ba63316bac7339f7a796db31830717edd/python_modules/dagster/dagster/_core/definitions/events.py#L465) takes `CoercibleToAssetKey` in the constructor but has `AssetKey` for the type of the corresponding attribute.

I think the solution here is to render both the (public) attributes and constructor params in the docs (this would match our non-`NamedTuple` classes, which have both rendered). The class level docstring should primarily document "Args", as for other classes.  We can have sphinx by default copy the docstring from the "Args" for each matching parameter, but you can also override this for individual attributes in a separate "Attributes" section.

smackesey:
>Might be worth collecting other opinions from the team? Not a hill that I'm going to die on.

pinging @dpeng817

smackesey:
@sryza ping for response to [above](https://github.com/dagster-io/dagster/pull/15404#issuecomment-1654060404)

AFAICT this is straightforwardly correct for non-NamedTuple classes for the reasons I wrote.

yuhan:
im on team `Args` but im not a super python expert. my reason is mostly the api docs generated via args look a lot better.

however, i did a quick research in popular python libs:
- https://scikit-learn.org/stable/modules/generated/sklearn.covariance.EmpiricalCovariance.html#sklearn.covariance.EmpiricalCovariance
- https://pandas.pydata.org/docs/reference/frame.html
- https://docs.python.org/3/library/collections.html#collections.OrderedDict
- https://boto3.amazonaws.com/v1/documentation/api/latest/reference/customizations/s3.html#s3-transfers

where some list those as "parameters" and some are "attributes". wonder if we could leave things as "parameters" and not differentiate them?

im fine either way. if we want to be *right*, it's worth checking with other folks who have more core python understanding.

dpeng817:
What's the status here? Requesting changes to get this out of my queue.

FWIW, I'm in favor of the args approach. Formatting is clearly a step function improvement over attrs.

dpeng817 (changes requested):
back to q

PedramNavid:
Looks like this is still relevant even with the new docs beta release
image
image

I will make a strong case for Args: style too @smackesey. Are you still motivated to get this PR in? Colton or myself can also make these changes in a new PR and close this one instead of worrying about merge conflicts.

smackesey:
@PedramNavid @cmpadden I am motivated to get this in now, I think everyone agrees on "Args"? I can rebase and spruce this up if you guys didn't already do this:

>Colton or myself can also make these changes in a new PR and close this one instead of worrying about merge conflicts.

cmpadden:
> @PedramNavid @cmpadden I am motivated to get this in now, I think everyone agrees on "Args"? I can rebase and spruce this up if you guys didn't already do this:
> 
> 
> 
> >Colton or myself can also make these changes in a new PR and close this one instead of worrying about merge conflicts.

Args sounds good to me @smackesey. Feel free to tag me once you've rebased and I can give it a stamp.

smackesey:
@cmpadden I rebased and this is good to go

docs/sphinx/_ext/dagster-sphinx/dagster_sphinx/__init__.py
```diff
@@ -83,6 +83,18 @@ def check_public_method_has_docstring(env: BuildEnvironment, name: str, obj: obj
         record_error(message)
 
 
+# Note that in our codebase docstrings with attributes will usually be written as:
+#
+#     Attributes:
```
cmpadden, line 88: Is this backwards? Should we say we use `Args` in our codebase?

cmpadden (approved):
Approved with one non-blocking comment.

docs/sphinx/_ext/dagster-sphinx/dagster_sphinx/__init__.py
smackesey: Nah this is the code that checks for attributes and throws an error if it gets detected

cmpadden: Ah okay, cool. 👍

PR #15411, opened 2023-07-20
[docs] remove guide about non argument deps

jamiedemaria:
## Summary & Motivation
removes an out of date guide on non-argument-deps


sryza (approved):
RIP

PR #15422, opened 2023-07-20
[docs] update sda docs to use deps more

jamiedemaria:
## Summary & Motivation
In an effort to make `deps` a first class concept, I'm updating the concept code snippets where it makes sense to use `deps` rather than other ways to defining dependencies. 

This PR updates the Assets section of the concepts.

I chose to update whatever examples I could to use deps. We could alternatively move some of those back to arg-based deps to show them as equally usable (specifically thinking about the auto materialization page)

Several examples could not be updated because they are incompatible with `deps`. I either left them as is, or opened issues for ones that we should address with API changes

### DIFF
--- a/docs/content/concepts/assets/asset-auto-execution.mdx
+++ b/docs/content/concepts/assets/asset-auto-execution.mdx
@@ -36,8 +36,8 @@ def asset1():
     ...
 
 
-@asset(auto_materialize_policy=AutoMaterializePolicy.eager())
-def asset2(asset1):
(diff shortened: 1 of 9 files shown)


jamiedemaria:
* master
  * **PR #15422** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15422" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15422?utm_source=stack-comment).

erinkcochran87 (approved):

PR #15424, opened 2023-07-20
[auto-materialize][refactor] Rename AssetDaemonCursor and moved namedtuples to more appropriate files

OwenKephart:
## Summary & Motivation

Breaking up the previous massive PR (https://github.com/dagster-io/dagster/pull/15317) into smaller chunks. This one just renames the AssetReconciliation cursor to the AssetDaemonCursor, puts it in a separate file, and moves the AutoMaterializeAssetEvaluation tuple into the auto_materialize_conditions.py file


sryza (approved):
So this isn't used anywhere in storage?

OwenKephart:
@sryza thankfully the special serdes stuff saves us here -- the name is not persisted into storage

PR #15425, opened 2023-07-20
[auto-materialize][refactor] Move "find_parent_materialized_asset_partitions" to the caching instance queryer

OwenKephart:
## Summary & Motivation

It makes more sense for this logic to live here, as it's used by both the backfill daemon and the asset daemon, and it mostly just calls a bunch of other caching instance queryer methods.

### DIFF
--- a/python_modules/dagster/dagster/_utils/caching_instance_queryer.py
+++ b/python_modules/dagster/dagster/_utils/caching_instance_queryer.py
@@ -3,12 +3,15 @@ from datetime import datetime
 from typing import (
     TYPE_CHECKING,
     AbstractSet,
+    Callable,
     Dict,
+    FrozenSet,
     Iterable,
     List,
     Mapping,
     Optional,
     Sequence,
+    Set,
     Tuple,
     Union,
     cast,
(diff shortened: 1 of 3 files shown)


python_modules/dagster/dagster/_utils/caching_instance_queryer.py
```diff
@@ -515,6 +518,147 @@ def get_dynamic_partitions(self, partitions_def_name: str) -> Sequence[str]:
...
         return partition_key in self.get_dynamic_partitions(partitions_def_name)
 
+    @cached_method
+    def asset_partitions_with_newly_updated_parents_and_new_latest_storage_id(
+        self,
+        *,
+        latest_storage_id: Optional[int],
+        target_asset_keys: FrozenSet[AssetKey],
+        target_asset_keys_and_parents: FrozenSet[AssetKey],
+        can_reconcile_fn: Callable[[AssetKeyPartitionKey], bool] = lambda _: True,
```
sryza, line 528: Will this go away in some way in the future? A little weird to be caching on this.

sryza (approved):

python_modules/dagster/dagster/_utils/caching_instance_queryer.py
OwenKephart: yeah this can be removed at any time (just didn't want to comingle that mild implementation change with this)

PR #15427, opened 2023-07-20
[auto-materialize][refactor] Move "asset reconciliation" logic into an AssetDaemonContext object

OwenKephart:
## Summary & Motivation

This moves all of the asset reconciliation logic into a more manageable AssetDaemonContext -- basically a pure copy/paste of the methods, but with the added benefit of removing a bunch of the arguments to some of these functions, as they can just become properties of the class.

This is just making room for a more fundamental change to the reconciliation logic in the next PR.


johannkm (approved):
i share sandy's concerns over a magic context object but it's in pursuit of a good goal

PR #15439, opened 2023-07-20
Pin marshmallow < 3.20.0

jmsanders:
https://github.com/marshmallow-code/marshmallow/issues/2152

I think this release breaks part of sqlalchemy that airflow is using. Let's pin until they have a fix so our builds stay green.


lafrech:
Fix released in marshmallow 3.20.1.

PR #15440, opened 2023-07-20
[ui] Use a regular RR link for historical alert on AssetView

### LINKED ISSUE #15410: No browser history entry created when showing latest materialisations from banner on historic view
### Dagster version

1.3.14

### What's the issue?

When looking at an older materialisation there's a warning shown that the materialisation is old and what appears to be a link to the latest values:

image

However that link appears to instead be implemented as a `<button>` and despite updating the url of the page (from e.g: `/assets/items?asOf=1689862000173` to `/assets/items?time=1689862000173`) no browser history entry is created. This means that the user cannot easily return to the previous view -- pressing "Back" takes the user to the previous view, from where they must remember which link they had clicked to get to the historic materialisation view.

Clicking a link should act like a link and create a browser history entry. Ideally changes in url should be implemented using anchor elements so that the browser does this natively (along with all the accessibility and power-user benefits from using browser-native behaviours rather than trying to reimplement them).

I also couldn't actually see any differences in the content of the page after having clicked the link, though that may be due to there having been few differences between my runs.

### How to reproduce?

Clicking the `[View Asset]` link on a historic run of an asset will take the user to a view which has the relevant message overlay.
My asset was a partitioned asset, though I'm not expecting that to be important.

### Deployment type

Local

### Deployment details

_No response_

### Additional information

_No response_

hellendag:
## Summary & Motivation

Resolves #15410

The link in the `HistoricalViewAlert` on `AssetView` is really for navigation, not for in-place query param updates. Instead of using query persisted state, just build a link to the appropriate `time` value.

## How I Tested These Changes

Navigate from a run to a historical view of an asset. Verify that when I click the link in the alert, I navigate via normal link behavior, and that I can navigate back with the back button, as expected.

### DIFF
--- a/js_modules/dagit/packages/core/src/assets/AssetView.tsx
+++ b/js_modules/dagit/packages/core/src/assets/AssetView.tsx
@@ -2,7 +2,6 @@ import {gql, useQuery} from '@apollo/client';
 import {
   Alert,
   Box,
-  ButtonLink,
   Colors,
   NonIdealState,
   Spinner,
@@ -12,7 +11,7 @@ import {
   ErrorBoundary,
 } from '@dagster-io/ui';
 import * as React from 'react';
-import {Link} from 'react-router-dom';
+import {Link, useLocation} from 'react-router-dom';
 
 import {
   FIFTEEN_SECONDS,
@@ -296,12 +295,8 @@ export const AssetView: React.FC<Props> = ({assetKey}) => {
           </Box>
         }
       />
-      {!viewingMostRecent && (
-        <HistoricalViewAlert
-          asOf={params.asOf}
-          onClick={() => setParams({asOf: undefined, time: params.asOf})}
-          hasDefinition={!!definition}
-        />
+      {!viewingMostRecent && params.asOf && (
+        <HistoricalViewAlert asOf={params.asOf} hasDefinition={!!definition} />
       )}
       <ErrorBoundary region="page" resetErrorOnChange={[assetKey, params]}>
         {selectedTab === 'definition' ? (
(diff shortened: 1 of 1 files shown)


hellendag:
* master
  * **PR #15440** <a href="https://app.graphite.dev/github/pr/dagster-io/dagster/15440" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/dagster-io/dagster/15440?utm_source=stack-comment).

js_modules/dagit/packages/core/src/assets/AssetView.tsx
```diff
@@ -458,39 +453,44 @@ export const ASSET_VIEW_DEFINITION_QUERY = gql`
...
-            ? 'Show definition and latest materializations'
-            : 'Show latest materializations'}
-        </ButtonLink>
-      }
-    />
-  </Box>
-);
+        }
+        description={
+          <Link to={`${pathname}?time=${asOf}`}>
```
bengotow, line 484: Ahh thanks for jumping on this! You beat me to it 🙌  

I think this will work but this alert appears outside the tab area of the AssetView and I'm worried that making this just pathname + time could strip the `?view=partitions` param and jump you to the events tab from the partitions tab. 

It's kind of a pain but maybe this to update just the one param?

``` 
return `${pathname}?${new URLSearchParams({...Object.fromEntries(new URLSearchParams(search)), asOf}).toString()}`
```

bengotow (approved):

js_modules/dagit/packages/core/src/assets/AssetView.tsx
hellendag: Oh, yeah, I wasn't sure if it was necessary to preserve those. Will update.

